PR #2580, opened 2016-08-26
WIP: Abstract list

nop33:
It is rebased on `wip/abstract-models` because I needed the models to work on the list.
Just before I make the PR I rebased on `wip/abstract-models`, but I didn't have the time to fix my db after the latest changes in the models so I didn't test it after the rebase.

PR #2581, opened 2016-08-29
Abstract list

ThiefMaster:
Supersedes #2580

* [x] Mark contribution type as mandatory (apply abstract settings)
* [x] Remove `vertical-aligner` from abstract list template
* [x] Unlink contribution - abstract when abstract is soft-deleted
* [x] Ask if contribution should also be deleted when the abstract is soft-deleted

### DIFF
--- a/indico/MaKaC/webinterface/rh/conferenceModif.py
+++ b/indico/MaKaC/webinterface/rh/conferenceModif.py
@@ -1512,7 +1512,7 @@ class RHMaterialPackage(RHConferenceModifBase, AttachmentPackageGeneratorMixin):
         if not attachments:
             flash(_('The selected contributions do not have any materials.'), 'warning')
             return redirect(url_for('event_mgmt.confModifContribList', self._conf))
-        return self._generate_zip_file(attachments)
+        return self._generate_zip_file(attachments, name_suffix=self.event_new.id)
 
 
 class RHAbstractBook( RHConfModifCFABase ):
--- a/indico/htdocs/js/custom/dropzone.js
+++ b/indico/htdocs/js/custom/dropzone.js
@@ -63,18 +63,20 @@
                     }
                 }
 
-                $button.on('click', function(e) {
-                    e.preventDefault();
-                    e.stopPropagation();
-                    $button.prop('disabled', true);
-                    $.each(self.getRejectedFiles(), function(index, file) {
-                        self.removeFile(file);
-                    });
-                    if (self.getQueuedFiles().length) {
-                        $dz.find('.dz-progress').show();
-                        self.processQueue();
-                    } else if (options.submitIfEmpty) {
-                        $form.submit();
+                $form.on('submit', function(e) {
+                    var evt = $.Event('ajaxDialog:validateBeforeSubmit');
+                    $(this).trigger(evt);
+                    if (!evt.isDefaultPrevented()) {
+                        $button.prop('disabled', true);
+                        $.each(self.getRejectedFiles(), function(index, file) {
+                            self.removeFile(file);
+                        });
+                        if (self.getQueuedFiles().length) {
+                            e.preventDefault();
+                            e.stopPropagation();
+                            $dz.find('.dz-progress').show();
+                            self.processQueue();
+                        }
                     }
                 });
 
--- a/indico/htdocs/js/indico/jquery/ajaxdialog.js
+++ b/indico/htdocs/js/indico/jquery/ajaxdialog.js
@@ -308,6 +308,11 @@
                     dataType: 'json',
                     data: options.getExtraData.call(this, options.trigger),
                     beforeSubmit: function() {
+                        var evt = $.Event('ajaxDialog:validateBeforeSubmit');
+                        $this.trigger(evt);
+                        if (evt.isDefaultPrevented()) {
+                            return false;
+                        }
                         $this.trigger('ajaxDialog:beforeSubmit');
                     },
                     error: function(xhr) {
--- a/indico/htdocs/js/indico/widgets/person_link_widget.js
+++ b/indico/htdocs/js/indico/widgets/person_link_widget.js
@@ -317,7 +317,7 @@
                 people.forEach(function(person) {
                     if (person.authorType === undefined) {
                         var maxOrder = _.max(people, _.iteratee('displayOrder')).displayOrder;
-                        person.authorType = options.defaults.authorType;
+                        setPersonDefaults(person);
                         person.displayOrder = customOrder ? (maxOrder + 1) : 0;
                     }
                 });
--- a/indico/htdocs/js/indico/widgets/person_link_widget.js
+++ b/indico/htdocs/js/indico/widgets/person_link_widget.js
@@ -35,6 +35,12 @@
                 authorType: 0,
                 isSpeaker: false,
                 isSubmitter: false
+            },
+            require: {
+                primaryAuthor: false,
+                secondaryAuthor: false,
+                submitter: false,
+                speaker: false
             }
         }, options);
 
@@ -345,6 +351,39 @@
                 });
             }
             $field.principalfield('refresh');
+        });
+
+        $field.closest('form').on('ajaxDialog:validateBeforeSubmit', function(evt) {
+            var $this = $(this);
+            var req = options.require;
+            if (req.primaryAuthor || req.secondaryAuthor || req.submitter || req.speaker) {
+                var hiddenData = JSON.parse($field.val());
+                var $formField = $field.closest('.form-field');
+                var $formGroup = $field.closest('.form-group');
+                var hasError = false;
+
+                if (req.speaker && _.indexOf(_.pluck(hiddenData, 'isSpeaker'), true) === -1) {
+                    hasError = true;
+                    $formField.data('error', $T.gettext('You must add at least one speaker'));
+                } else if (req.submitter && _.indexOf(_.pluck(hiddenData, 'isSubmitter'), true) === -1) {
+                    hasError = true;
+                    $formField.data('error', $T.gettext('You must add at least one submitter'));
+                } else if (req.secondaryAuthor && _.indexOf(_.pluck(hiddenData, 'authorType'), 2) === -1) {
+                    hasError = true;
+                    $formField.data('error', $T.gettext('You must add at least one co-author'));
+                } else if (req.primaryAuthor && _.indexOf(_.pluck(hiddenData, 'authorType'), 1) === -1) {
+                    hasError = true;
+                    $formField.data('error', $T.gettext('You must add at least one author'));
+                }
+                if (hasError) {
+                    evt.preventDefault();
+                    $formGroup.addClass('has-error');
+                    showFormErrors($this.parent());
+                } else {
+                    $formGroup.removeClass('has-error');
+                    $formField.removeData('error');
+                }
+            }
         });
     };
 })(window);
--- a/indico/htdocs/js/indico/widgets/person_link_widget.js
+++ b/indico/htdocs/js/indico/widgets/person_link_widget.js
@@ -29,7 +29,8 @@
                 authors: false,
                 submitters: false,
                 emptyEmail: false,
-                externalUsers: true
+                externalUsers: true,
+                speakers: true
             },
             defaults: {
                 authorType: 0,
@@ -170,12 +171,12 @@
                 $personRoles.append($submitterLabel.toggleClass('selected', person.isSubmitter));
             }
 
-            if (options.allow.authors) {
+            if (options.allow.authors && options.allow.speakers) {
                 $personRoles.prepend($speakerLabel.toggleClass('selected', person.isSpeaker));
             }
 
             if (entryType !== null && entryType === options.authorTypes.none) {
-                if (options.allow.authors) {
+                if (options.allow.authors && options.allow.speakers) {
                     $speakerLabel.addClass('other');
                 }
             }
@@ -233,33 +234,34 @@
                     $buttons.append(actionButton($T.gettext("Move to co-authors"), $coauthorList, {authorType: 2}));
                 }
 
-                if (person.authorType !== options.authorTypes.none) {
-                    $buttons.append(actionButton($T.gettext("Move to others"), $otherList,
-                                                 {authorType: 0, isSpeaker: true}));
+                if (options.allow.speakers) {
+                    if (person.authorType !== options.authorTypes.none) {
+                        $buttons.append(actionButton($T.gettext("Move to others"), $otherList,
+                                                     {authorType: 0, isSpeaker: true}));
+                        var text = person.isSpeaker ? $T.gettext("Not a speaker anymore")
+                                                    : $T.gettext("Make a speaker");
 
-                    var text = person.isSpeaker ? $T.gettext("Not a speaker anymore")
-                                                : $T.gettext("Make a speaker");
+                        $buttons.append($buttonsSeparator);
+                        $buttons.append(actionButton(text, null, {isSpeaker: !person.isSpeaker}));
 
-                    $buttons.append($buttonsSeparator);
-                    $buttons.append(actionButton(text, null, {isSpeaker: !person.isSpeaker}));
-
-                    $speakerLabel.on('click', function() {
-                        $field.principalfield('set', person.id, {isSpeaker: !person.isSpeaker});
-                    });
-                } else {
-                    $buttons.append($buttonsSeparator);
-                    $speakerLabel.qtip({
-                        content: {
-                            text: $T.gettext("People in the 'Others' section <strong>must</strong> be speakers.")
-                        },
-                        show: {
-                            event: 'click',
-                            solo: true
-                        },
-                        hide: {
-                            event: 'unfocus click'
-                        }
-                    });
+                        $speakerLabel.on('click', function() {
+                            $field.principalfield('set', person.id, {isSpeaker: !person.isSpeaker});
+                        });
+                    } else {
+                        $buttons.append($buttonsSeparator);
+                        $speakerLabel.qtip({
+                            content: {
+                                text: $T.gettext("People in the 'Others' section <strong>must</strong> be speakers.")
+                            },
+                            show: {
+                                event: 'click',
+                                solo: true
+                            },
+                            hide: {
+                                event: 'unfocus click'
+                            }
+                        });
+                    }
                 }
             }
 
@@ -356,10 +358,11 @@
         $field.closest('form').on('ajaxDialog:validateBeforeSubmit', function(evt) {
             var $this = $(this);
             var req = options.require;
-            if (req.primaryAuthor || req.secondaryAuthor || req.submitter || req.speaker) {
-                var hiddenData = JSON.parse($field.val());
+
+            if (req.primaryAuthor || req.secondaryAuthor || req.submitter || req.speaker || !options.allow.speakers) {
                 var $formField = $field.closest('.form-field');
                 var $formGroup = $field.closest('.form-group');
+                var hiddenData = JSON.parse($field.val());
                 var hasError = false;
 
                 if (req.speaker && _.indexOf(_.pluck(hiddenData, 'isSpeaker'), true) === -1) {
@@ -374,7 +377,12 @@
                 } else if (req.primaryAuthor && _.indexOf(_.pluck(hiddenData, 'authorType'), 1) === -1) {
                     hasError = true;
                     $formField.data('error', $T.gettext('You must add at least one author'));
+                } else if (!options.allow.speakers && _.indexOf(_.pluck(hiddenData, 'authorType'), 0) !== -1) {
+                    hasError = true;
+                    $formField.data('error', $T.gettext('You cannot have users in the Other section, please make ' +
+                                                        'them authors or co-authors'));
                 }
+
                 if (hasError) {
                     evt.preventDefault();
                     $formGroup.addClass('has-error');
--- a/indico/modules/attachments/forms.py
+++ b/indico/modules/attachments/forms.py
@@ -30,9 +30,10 @@ from indico.modules.attachments.models.folders import AttachmentFolder
 from indico.modules.attachments.util import get_default_folder_names
 from indico.util.i18n import _
 from indico.web.forms.base import IndicoForm, generated_data
-from indico.web.forms.fields import IndicoSelectMultipleCheckboxField, IndicoRadioField, AccessControlListField
+from indico.web.forms.fields import (IndicoSelectMultipleCheckboxField, IndicoRadioField, AccessControlListField,
+                                     FileField)
 from indico.web.forms.validators import UsedIf, HiddenUnless
-from indico.web.forms.widgets import SwitchWidget, TypeaheadWidget, DropzoneWidget
+from indico.web.forms.widgets import SwitchWidget, TypeaheadWidget
 
 
 class AttachmentFormBase(IndicoForm):
@@ -62,12 +63,12 @@ class EditAttachmentFormBase(AttachmentFormBase):
 
 
 class AddAttachmentFilesForm(AttachmentFormBase):
-    files = HiddenField(_("Files"), widget=DropzoneWidget(submit_form=False))
+    files = FileField(_("Files"))
 
 
 class EditAttachmentFileForm(EditAttachmentFormBase):
-    file = HiddenField(_("File"), widget=DropzoneWidget(max_files=1, submit_form=False, add_remove_links=False),
-                       description=_("Already uploaded file. Replace it by adding a new file."))
+    file = FileField(_("File"), max_files=1, add_remove_links=False,
+                     description=_("Already uploaded file. Replace it by adding a new file."))
 
 
 class AttachmentLinkFormMixin(object):
--- a/indico/modules/attachments/forms.py
+++ b/indico/modules/attachments/forms.py
@@ -63,11 +63,11 @@ class EditAttachmentFormBase(AttachmentFormBase):
 
 
 class AddAttachmentFilesForm(AttachmentFormBase):
-    files = FileField(_("Files"))
+    files = FileField(_("Files"), multiple_files=True)
 
 
 class EditAttachmentFileForm(EditAttachmentFormBase):
-    file = FileField(_("File"), max_files=1, add_remove_links=False,
+    file = FileField(_("File"), add_remove_links=False,
                      description=_("Already uploaded file. Replace it by adding a new file."))
 
 
--- a/indico/modules/events/abstracts/controllers/management.py
+++ b/indico/modules/events/abstracts/controllers/management.py
@@ -188,7 +188,7 @@ class RHCreateAbstract(RHAbstractListBase):
             files = request.files.getlist('file')
             abstract = create_abstract(self.event_new, *get_field_values(data), files=files)
             flash(_("Abstract '{}' created successfully").format(abstract.title), 'success')
-            tpl_components = self.list_generator.render_list()
+            tpl_components = self.list_generator.render_list(abstract)
             if tpl_components.get('hide_abstract'):
                 self.list_generator.flash_info_message(abstract)
             return jsonify_data(**tpl_components)
--- a/indico/modules/events/abstracts/controllers/management.py
+++ b/indico/modules/events/abstracts/controllers/management.py
@@ -182,11 +182,9 @@ class RHCreateAbstract(RHAbstractListBase):
         form = abstract_form_class(event=self.event_new)
         if form.validate_on_submit():
             data = form.data
-            data.pop('attachments', None)
             if isinstance(data['submitted_for_tracks'], Track):
                 data['submitted_for_tracks'] = {data['submitted_for_tracks']}
-            files = request.files.getlist('file')
-            abstract = create_abstract(self.event_new, *get_field_values(data), files=files)
+            abstract = create_abstract(self.event_new, *get_field_values(data))
             flash(_("Abstract '{}' created successfully").format(abstract.title), 'success')
             tpl_components = self.list_generator.render_list(abstract)
             if tpl_components.get('hide_abstract'):
--- a/indico/modules/events/abstracts/controllers/management.py
+++ b/indico/modules/events/abstracts/controllers/management.py
@@ -33,6 +33,7 @@ from indico.modules.events.abstracts.settings import boa_settings, abstracts_set
 from indico.modules.events.abstracts.util import AbstractListGenerator, make_abstract_form
 from indico.modules.events.abstracts.views import WPManageAbstracts
 from indico.modules.events.contributions.models.persons import AuthorType
+from indico.modules.events.tracks.models.tracks import Track
 from indico.modules.events.util import get_field_values, ZipGeneratorMixin
 from indico.util.fs import secure_filename
 from indico.util.i18n import _, ngettext
@@ -181,7 +182,9 @@ class RHCreateAbstract(RHAbstractListBase):
         form = abstract_form_class(event=self.event_new)
         if form.validate_on_submit():
             data = form.data
-            del data['attachments']
+            data.pop('attachments', None)
+            if isinstance(data['submitted_for_tracks'], Track):
+                data['submitted_for_tracks'] = {data['submitted_for_tracks']}
             files = request.files.getlist('file')
             abstract = create_abstract(self.event_new, *get_field_values(data), files=files)
             flash(_("Abstract '{}' created successfully").format(abstract.title), 'success')
--- a/indico/modules/events/abstracts/controllers/management.py
+++ b/indico/modules/events/abstracts/controllers/management.py
@@ -197,12 +197,20 @@ class RHCreateAbstract(RHAbstractListBase):
 
 class RHDeleteAbstracts(RHManageAbstractsActionsBase):
     def _process(self):
+        delete_contribs = request.values.get('delete_contribs') == '1'
+        deleted_contrib_count = 0
         for abstract in self.abstracts:
-            delete_abstract(abstract)
-        deleted_count = len(self.abstracts)
+            if delete_contribs and abstract.contribution:
+                deleted_contrib_count += 1
+            delete_abstract(abstract, delete_contribs)
+        deleted_abstract_count = len(self.abstracts)
         flash(ngettext("The abstract has been deleted.",
-                       "{count} abstracts have been deleted.", deleted_count)
-              .format(count=deleted_count), 'success')
+                       "{count} abstracts have been deleted.", deleted_abstract_count)
+              .format(count=deleted_abstract_count), 'success')
+        if deleted_contrib_count:
+            flash(ngettext("The linked contribution has been deleted.",
+                           "{count} linked contributions have been deleted.", deleted_contrib_count)
+                  .format(count=deleted_contrib_count), 'success')
         return jsonify_data(**self.list_generator.render_list())
 
 
--- a/indico/modules/events/abstracts/controllers/management.py
+++ b/indico/modules/events/abstracts/controllers/management.py
@@ -178,13 +178,16 @@ class RHCreateAbstract(RHAbstractListBase):
         abstract_form_class = make_abstract_form(self.event_new)
         form = abstract_form_class(event=self.event_new)
         if form.validate_on_submit():
-            abstract = create_abstract(self.event_new, *get_field_values(form.data))
+            data = form.data
+            del data['attachments']
+            files = request.files.getlist('file')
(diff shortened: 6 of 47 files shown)


indico/modules/events/contributions/models/fields.py
```diff
@@ -115,6 +115,10 @@ def field(self):
     def mgmt_field(self):
         return self._get_field(management=True)
 
+    @property
+    def filter_choices(self):
+        return {x['id']: x['option'] for x in self.field_data['options']} if self.field_data.get('options') else []
```
ThiefMaster: `dict` vs `list`

indico/htdocs/js/indico/jquery/ajaxdialog.js
```diff
@@ -308,6 +308,11 @@
                     dataType: 'json',
                     data: options.getExtraData.call(this, options.trigger),
                     beforeSubmit: function() {
+                        var evt = $.Event('ajaxDialog:validateBeforeSubmit');
```
ThiefMaster: I guess you needed to use a new event to prevent the ajaxDialog's own beforeSubmit handler from showing the progress dialog?

nop33: exactly

indico/modules/events/contributions/models/fields.py
```diff
@@ -115,6 +115,10 @@ def field(self):
     def mgmt_field(self):
         return self._get_field(management=True)
 
+    @property
+    def filter_choices(self):
+        return {x['id']: x['option'] for x in self.field_data['options']} if self.field_data.get('options') else {}
```
ThiefMaster: In case `self.field_data['options']` is always a dict (i.e. not `None` etc) you can simplify this:

    return {x['id']: x['option'] for x in self.field_data.get('options', {})}

indico/modules/events/management/controllers.py
```diff
@@ -126,6 +127,33 @@ def _process(self):
                                 event_persons=contribution_persons_dict, event=self.event_new)
 
 
+class RHAbstractPersonListMixin:
```
ThiefMaster: Why a mixin? It's only used in one place unlike the one for sessions/contributions (which is why that one was a mixin)

indico/modules/events/management/controllers.py
```diff
@@ -126,6 +127,33 @@ def _process(self):
...
+    """List of persons somehow related to abstracts (co-authors, speakers...)"""
+
+    @property
+    def _membership_filter(self):
+        raise NotImplementedError
+
+    def _process(self):
+        submitters = {abstract.submitter for abstract in self.abstracts}
+        abstract_persons = (AbstractPersonLink
+                            .find(AbstractPersonLink.abstract.has(self._membership_filter))
```
ThiefMaster: you can use `find_all()` here and it probably fits in one line

indico/modules/events/util.py
```diff
@@ -310,12 +311,19 @@ def _filter_list_entries(self):
 
     def _get_filters_from_request(self):
         """Get the new filters after the filter form is submitted."""
+        def get_selected_options(item_id, item):
```
ThiefMaster: I'd underscore it

OmeGak: Why? It's local scoped.

indico/htdocs/js/indico/widgets/person_link_widget.js
```diff
@@ -346,5 +352,42 @@
...
+                var hiddenData = JSON.parse($field.val());
+                var $formField = $field.closest('.form-field');
+                var $formGroup = $field.closest('.form-group');
+                var hasError = false;
+
+                if (options.require.speaker && _.indexOf(_.pluck(hiddenData, 'isSpeaker'), true) === -1) {
+                    hasError = true;
+                    $formField.data('error', $T.gettext('You must add at least 1 speaker'));
+                }
+                if (options.require.submitter && _.indexOf(_.pluck(hiddenData, 'isSubmitter'), true) === -1) {
```
OmeGak: You check all requirements but then you show only the error message corresponding to the last missing requirement. Why not make this an `if .. else if` chain? That way we can stop checking after the first one that doesn't apply.

indico/htdocs/sass/modules/_abstracts.scss
```diff
@@ -16,3 +16,7 @@
  */
 
 @import 'abstracts/abstract_page';
+
+.submitter-column {
```
OmeGak: I don't know if we want to mix imports and CSS rules. Maybe you could create an `abstract/_abstract_list.scss` file? Not a strong opinion.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -352,6 +353,18 @@ def reviewing_state(self):
         else:
             return AbstractReviewingState.mixed
 
+    @property
+    def data_by_field(self):
+        return {value.contribution_field_id: value for value in self.field_values}
+
+    @property
+    def submitted_for_tracks_titles(self):
```
OmeGak: Why do we need to add a property for this?

indico/modules/events/abstracts/util.py
```diff
@@ -136,3 +193,41 @@ def render_list(self):
             'filtering_enabled': filtering_enabled,
             'filter_statistics': filter_statistics
         }
+
+
+def serialize_abstract_person_link(person_link):
```
OmeGak: Maybe move it to `abstracts/serialize.py`?

indico/modules/events/abstracts/templates/forms/abstract.html
```diff
@@ -0,0 +1,10 @@
+{% from 'forms/_form.html' import form_header, form_footer, form_rows, form_fieldset %}
```
nop33: I need to pass the `action` attribute to the `<form>` for the dropzone to work.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -352,6 +353,18 @@ def reviewing_state(self):
...
+        return [x.title for x in self.submitted_for_tracks]
+
+    @property
+    def proposed_for_tracks_titles(self):
+        return [x.title for x in self.proposed_for_tracks]
```
nop33: Shouldn't the `_proposed_for_tracks` table have been translated as well?

indico/modules/events/abstracts/util.py
```diff
@@ -63,42 +79,82 @@ def _get_static_columns(self, ids):
...
+                .order_by(ContributionField.position)
+                .all())
+
+    def _get_filters_from_request(self):
+        filters = super(AbstractListGenerator, self)._get_filters_from_request()
+        for field in self.list_event.contribution_fields:
+            if field.field_type == 'single_choice':
+                options = request.form.getlist('field_{}'.format(field.id))
+                if options:
+                    filters['fields'][str(field.id)] = options
```
nop33: Why is this necessary? The `field.id` is always an `int` (it's the `id` column)

ThiefMaster: most likely stored as json where all dict keys become strings. unicode instead of str is just nicer since we use unicode everywhere anyway

indico/modules/events/abstracts/models/abstracts.py
ThiefMaster: didn't i? i'll check tomorrow (ping me in case i forget)

ThiefMaster: The new `proposed_for_tracks` table is something different (reviews that propose "other tracks", i.e. a many-to-many relationship between AbstractReview and Track)

indico/modules/events/abstracts/templates/forms/abstract.html
ThiefMaster: `jsonify_form` has `form_header_kwargs`

indico/modules/events/abstracts/models/abstracts.py
ThiefMaster: we don't, iirc i already suggested using `join(.., attr=..)` in jinja in the places where the property is used atm

indico/modules/events/abstracts/util.py
nop33: Hm, is it really worth it to create a new file for just one function? I would totally be in to create the file if we had a couple functions, but it's just one. If the majority thinks it's cleaner though, I'll gladly do it!

indico/modules/events/abstracts/templates/forms/abstract.html
nop33: Are we sure we don't want fieldsets though? The abstract creation form is growing pretty long: http://imgur.com/11y9qPW
Maybe a fieldset for the attachments and/or the contribution fields?

indico/modules/events/abstracts/util.py
```diff
@@ -63,42 +79,82 @@ def _get_static_columns(self, ids):
...
-            if None in ids:
-                column_criteria.append(column.is_(None))
-            if ids - {None}:
-                column_criteria.append(column.in_(ids - {None}))
-            criteria.append(db.or_(*column_criteria))
+        field_filters = filters.get('fields')
+        item_filters = filters.get('items')
+        extra_filters = filters.get('extra')
+
+        if not (field_filters or item_filters or extra_filters):
```
nop33: Well, I read it "if there is no X or Y or Z". Not so sure if it's grammatically correct tho! Perhaps "if there is no X nor Y nor Z".

indico/modules/events/abstracts/util.py
ThiefMaster: No, not really.

indico/modules/events/management/templates/contribution_person_list.html
```diff
@@ -60,6 +67,18 @@
                                 -
                             {% endif %}
                         </td>
+                        {% if include_submitters %}
+                            {% set is_submitter = person_roles.submitter %}
+                            <td class="i-table submitter-column">
+                                {%- if is_submitter %}
+                                    {% set submitter_title = _('This person is a submitter') %}
```
ThiefMaster: "This person submitted an abstract"? "is a submitter" sounds more like people with some kind of submission privs

indico/htdocs/sass/modules/_abstracts.scss
ThiefMaster: Agreed, I'd also rather add a new file in `abstracts/` instead of mixing both in `_abstract.scss`

indico/modules/events/abstracts/forms.py
```diff
@@ -61,3 +66,27 @@ class AbstractSubmissionSettingsForm(IndicoForm):
...
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event')
+        self.abstract = kwargs.pop('abstract', None)
+        super(AbstractForm, self).__init__(*args, **kwargs)
+        self.submitted_contrib_type.query = self.event.contribution_types
+        if not self.submitted_contrib_type.query.count():
+            del self.submitted_contrib_type
+        self.submitted_for_tracks.query = Track.query.with_parent(self.event).order_by(Track.title)
+        tpl = get_template_module('forms/_dropzone_themes.html')
+        self.attachments.widget.options['previewTemplate'] = tpl.thin_preview_template()
```
ThiefMaster: I don't think this really belongs here. A custom Field class would be more appropriate

indico/modules/events/abstracts/templates/forms/abstract.html
ThiefMaster: For the custom fields it might be a good idea

indico/modules/events/abstracts/forms.py
```diff
@@ -61,3 +66,27 @@ class AbstractSubmissionSettingsForm(IndicoForm):
...
+class AbstractForm(IndicoForm):
+    title = StringField(_("Title"), [DataRequired()])
+    description = IndicoMarkdownField(_('Content'), [DataRequired()], editor=True, mathjax=True)
+    submitted_contrib_type = QuerySelectField(_("Type"), get_label='name', allow_blank=True,
+                                              blank_text=_("No type selected"))
+    person_links = AbstractPersonLinkListField(_("People"), [DataRequired()])
+    submitted_for_tracks = IndicoQuerySelectMultipleCheckboxField(_("Tracks"), get_label=lambda x: x.title,
+                                                                  collection_class=set)
+    submission_comment = TextAreaField(_("Comments"))
+    attachments = HiddenField(_('Attachments'), widget=DropzoneWidget(style='thin'))
```
ThiefMaster: this should only be visible if attachments are enabled (`del` it in `__init__` otherwise)

`abstract_settings.get('allow_attachments')`

indico/modules/events/abstracts/util.py
```diff
@@ -136,3 +194,32 @@ def render_list(self):
...
+    return data
+
+
+def make_abstract_form(event):
+    """Extends the abstract WTForm to add the extra fields.
+
+    Each extra field will use a field named ``custom_ID``.
+
+    :param event: The `Event` for which to create the abstract form.
+    :return: A `AbstractForm` subclass.
```
ThiefMaster: *An*

indico/web/templates/forms/_dropzone_themes.html
```diff
@@ -0,0 +1,13 @@
...
+            <div class="dz-filename icon-file"><span data-dz-name></span></div>
+        </div>
+        <div class="dz-progress"><span class="dz-upload" data-dz-uploadprogress></span></div>
+    </div>
+{% endmacro %}
+
+{% macro remove_icon() %}
+    <i class="icon-remove"></i>
+{% endmacro %}
+
```
ThiefMaster: extra blank line

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -123,3 +140,63 @@ class RHAbstractListStaticURL(RHAbstractListBase):
...
         return jsonify(url=self.list_generator.generate_static_url())
+
+
+class RHCreateAbstract(RHAbstractListBase):
+    def _process(self):
+        abstract_form_class = make_abstract_form(self.event_new)
+        form = abstract_form_class(event=self.event_new)
+        if form.validate_on_submit():
+            data = form.data
+            del data['attachments']
```
nop33: Do we have such a field already, or would it need to be implemented?

ThiefMaster: The latter, but it's pretty straightforward. You'd use a custom Field that uses the DropzoneWidget and then just gets the files from `valuelist` in `process_formdata` and puts them in `self.data`.

indico/htdocs/js/custom/dropzone.js
```diff
@@ -39,7 +39,7 @@
                 var self = this;
                 var file = $dz.data('value');
 
-                if (file) {
+                if (file && typeof file === 'object') {
```
ThiefMaster: what other type did you get here?

nop33: I think now that I fixed the problem with the form embedded dropzone, this is not necessary anymore

indico/modules/events/util.py
```diff
@@ -509,3 +519,74 @@ def serialize_event_for_ical(event, detail_level):
...
+                fname = fname[:-cut]
+            result.append(fname + ext)
+
+        return reversed(result)
+
+    def _iter_items(self, files_holder):
+        for f in files_holder:
+            yield f
+
+    def _generate_zip_file(self, files_holder, name_prefix='material', name_suffix=None):
```
ThiefMaster: why not `files_list`?

indico/MaKaC/webinterface/rh/conferenceModif.py
```diff
@@ -1512,7 +1512,7 @@ def _process(self):
         if not attachments:
             flash(_('The selected contributions do not have any materials.'), 'warning')
             return redirect(url_for('event_mgmt.confModifContribList', self._conf))
-        return self._generate_zip_file(attachments)
+        return self._generate_zip_file(files_holder=attachments, name_suffix=self.event_new.id)
```
ThiefMaster: files_holder is the first argument and required so i'd pass it as a posarg, not a kwarg

indico/htdocs/js/custom/dropzone.js
```diff
@@ -63,18 +63,22 @@
...
-                    $button.prop('disabled', true);
-                    $.each(self.getRejectedFiles(), function(index, file) {
-                        self.removeFile(file);
-                    });
-                    if (self.getQueuedFiles().length) {
-                        $dz.find('.dz-progress').show();
-                        self.processQueue();
-                    } else if (options.submitIfEmpty) {
-                        $form.submit();
+                $form.on('submit', function(e) {
```
ThiefMaster: Make sure to test all the places where we use a dropzone to ensure it still works fine

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -123,3 +142,79 @@ class RHAbstractListStaticURL(RHAbstractListBase):
...
+        abstract_form_class = make_abstract_form(self.event_new)
+        form = abstract_form_class(event=self.event_new)
+        if form.validate_on_submit():
+            data = form.data
+            if 'attachments' in data:
+                del data['attachments']
+            if not isinstance(data['submitted_for_tracks'], set):
+                submitted_for_track = set()
+                submitted_for_track.add(data['submitted_for_tracks'])
+                data['submitted_for_tracks'] = submitted_for_track
```
ThiefMaster: You can replace those three lines with this:

    data['submitted_for_tracks'] = {data['submitted_for_tracks']}

Also, I'd rather check if you already got a `Track` object and put it in a set instead of the other way round (what if we ever changed it to use a list - it's unlikely but would anyone think of changing the `isinstance` check here in that case?)

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -123,3 +142,79 @@ class RHAbstractListStaticURL(RHAbstractListBase):
...
+            files = request.files.getlist('file')
```
ThiefMaster: I'd get this through WTForms instead of passing it separately

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -16,16 +16,25 @@
 
 from __future__ import unicode_literals
 
-from flask import redirect, flash, jsonify
+from collections import defaultdict
+from os import path
```
ThiefMaster: Don't import it like this; `import os` and then later use `os.path.join` - `path` is a way too common name to have it as a global.

indico/web/forms/widgets.py
```diff
@@ -294,6 +294,7 @@ def __init__(self, post_url=None, max_file_size=None, max_files=10, add_remove_l
             'parallelUploads': max_files,
             'handleFlashes': handle_flashes
         }
+        self.style = style
```
ThiefMaster: Do we really want a freeform `style` argument instead of something like `lightweight=True`?

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -123,3 +142,79 @@ class RHAbstractListStaticURL(RHAbstractListBase):
...
+            if not isinstance(data['submitted_for_tracks'], set):
+                submitted_for_track = set()
+                submitted_for_track.add(data['submitted_for_tracks'])
+                data['submitted_for_tracks'] = submitted_for_track
+            files = request.files.getlist('file')
+            abstract = create_abstract(self.event_new, *get_field_values(data), files=files)
+            flash(_("Abstract '{}' created successfully").format(abstract.title), 'success')
+            tpl_components = self.list_generator.render_list()
+            if tpl_components.get('hide_abstract'):
+                self.list_generator.flash_info_message(abstract)
```
ThiefMaster: This method doesn't exist (it's only in the contribution list generator).

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -123,3 +142,79 @@ class RHAbstractListStaticURL(RHAbstractListBase):
...
         return jsonify(url=self.list_generator.generate_static_url())
+
+
+class RHCreateAbstract(RHAbstractListBase):
+    def _process(self):
+        abstract_form_class = make_abstract_form(self.event_new)
+        form = abstract_form_class(event=self.event_new)
+        if form.validate_on_submit():
+            data = form.data
+            if 'attachments' in data:
```
ThiefMaster: fyi, you can `data.pop('attachments', None)` to remove it if it's present and not do anything if not

indico/modules/events/abstracts/forms.py
```diff
@@ -113,3 +123,53 @@ class CreateEmailTemplateForm(EditEmailTemplateRuleForm):
...
+            del self.attachments
+        self.person_links.require_speaker_author = abstracts_settings.get(self.event, 'speakers_required')
+        self.person_links.allow_speakers = abstracts_settings.get(self.event, 'allow_speakers')
+
+
+class SingleTrackMixin(object):
+    submitted_for_tracks = QuerySelectField(_("Tracks"), get_label=lambda x: x.title)
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event')
```
ThiefMaster: I don't think this belongs in the mixin. I'd rather keep it in AbstractForm - you can still use it here since this methods runs when you call super in `AbstractForm.__init__`

indico/modules/events/abstracts/forms.py
```diff
@@ -113,3 +123,53 @@ class CreateEmailTemplateForm(EditEmailTemplateRuleForm):
...
+        if abstracts_settings.get(self.event, 'tracks_required'):
+            inject_validators(self, 'submitted_for_tracks', [DataRequired()])
+        super(SingleTrackMixin, self).__init__(*args, **kwargs)
+        if not abstracts_settings.get(self.event, 'tracks_required'):
+            self.submitted_for_tracks.allow_blank = True
```
ThiefMaster: What about always setting this and just using different blank texts ("No track selected" / "Select a track")? That way people have to select a track explicitly instead of just having the first one preselected.

indico/modules/events/abstracts/forms.py
```diff
@@ -113,3 +123,53 @@ class CreateEmailTemplateForm(EditEmailTemplateRuleForm):
...
+    submission_comment = TextAreaField(_("Comments"))
+    attachments = HiddenField(_('Attachments'), widget=DropzoneWidget(style='thin'))
+
+    def __init__(self, *args, **kwargs):
+        self.abstract = kwargs.pop('abstract', None)
+        super(AbstractForm, self).__init__(*args, **kwargs)
+        self.submitted_contrib_type.query = self.event.contribution_types
+        if not self.submitted_contrib_type.query.count():
+            del self.submitted_contrib_type
+        self.submitted_for_tracks.query = Track.query.with_parent(self.event).order_by(Track.title)
```
ThiefMaster: I'd move this to the mixins

indico/modules/events/abstracts/forms.py
```diff
@@ -113,3 +123,53 @@ class CreateEmailTemplateForm(EditEmailTemplateRuleForm):
...
+        if abstracts_settings.get(self.event, 'allow_attachments'):
+            tpl = get_template_module('forms/_dropzone_themes.html')
+            self.attachments.widget.options['previewTemplate'] = tpl.thin_preview_template()
```
ThiefMaster: Modifying this is not a good idea. When we move to a proper `DropzoneField` (which I think we should, see some of my other comments regarding `request.files`) this would modiy the widget instance which is usually set on the Field class level, i.e. shared between instances. So here you modify it for all instances of the field, not just this one.

indico/htdocs/js/custom/dropzone.js
nop33: yep, this'll go away

indico/htdocs/js/custom/dropzone.js
nop33: I tested it in the materials dialog and the conference images, works fine

indico/modules/events/abstracts/controllers/management.py
nop33: OK, I'll change it in `events/util.py` too then

indico/modules/events/abstracts/controllers/management.py
nop33: ...and discard the return value. Sure, good idea, thx.

indico/modules/events/abstracts/controllers/management.py
nop33: I only wanted to avoid an extra import, but sure, that was my idea too initially

ThiefMaster: `db.m.Track` if you don't want to import it ;)

indico/modules/events/util.py
nop33: It's not always a list, it can be any object that contains an object that cont.... that contains a list of files :stuck_out_tongue:

indico/web/forms/widgets.py
nop33: Well I was thinking that we might want to add further styles in the future.

ThiefMaster: do we? in that case i'd rather go for an enum - but i'd do that once we do end up adding 3+ styles which i find somewhat unlikely

indico/modules/events/abstracts/forms.py
nop33: How do you validate then if the user chooses the blank option and the field is mandatory?

ThiefMaster: `DataRequired()` validator as usual

indico/modules/events/abstracts/forms.py
nop33: `'_AbstractForm' object has no attribute 'event'` when I move it in the `AbstractForm`.

indico/web/forms/fields/files.py
```diff
@@ -0,0 +1,65 @@
...
+from indico.core.config import Config
+from indico.web.forms.widgets import JinjaWidget
+
+
+class FileField(Field):
+    """A dropzone field"""
+
+    widget = JinjaWidget('forms/dropzone_widget.html')
+
+    default_options = {
```
ThiefMaster: Why not just put the values right in the `pop()` calls?

indico/web/forms/fields/files.py
```diff
@@ -0,0 +1,65 @@
...
+        'max_files': 10,
+        'add_remove_links': True,
+        'param_name': 'file',
```
ThiefMaster: can we please get rid of this once and for all and use `self.name` instead? :)

indico/web/forms/fields/files.py
```diff
@@ -0,0 +1,65 @@
...
+from indico.web.forms.widgets import JinjaWidget
+
+
+class FileField(Field):
+    """A dropzone field"""
+
+    widget = JinjaWidget('forms/dropzone_widget.html')
+
+    default_options = {
+        'max_files': 10,
```
ThiefMaster: default to 1 and only set it in case you want to allow multiple files somewhere?

indico/modules/events/abstracts/forms.py
```diff
@@ -161,3 +173,54 @@ class CreateEmailTemplateForm(EditEmailTemplateRuleForm):
...
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event')
+        self.abstract = kwargs.pop('abstract', None)
+        super(AbstractForm, self).__init__(*args, **kwargs)
+        self.submitted_contrib_type.query = self.event.contribution_types
+        if not self.submitted_contrib_type.query.count():
+            del self.submitted_contrib_type
+        if abstracts_settings.get(self.event, 'allow_attachments'):
+            tpl = get_template_module('forms/_dropzone_themes.html')
+            self.attachments.widget_options['previewTemplate'] = tpl.thin_preview_template()
```
ThiefMaster: This shouldn't be done here but in the `FieldField` based on the `lightweight` option

indico/web/forms/fields/files.py
nop33: Sure, since it's a `FileField`, it kinda makes sense to default it to 1.

indico/web/forms/fields/files.py
nop33: It gives you a better overview of the default values (that are not `None`)

indico/web/forms/widgets.py
```diff
@@ -272,7 +272,7 @@ class DropzoneWidget(JinjaWidget):
     """Renders a dropzone file upload field"""
 
     def __init__(self, post_url=None, max_file_size=None, max_files=10, add_remove_links=True, accepted_file_types=None,
```
ThiefMaster: is this class still used anywhere?

nop33: Nope, I was meant to get rid of it, thx.

indico/web/forms/fields/files.py
nop33: Actually, how about having the boolean option `multipleFiles=False` as default? I don't like the hardcoded value `10` everywhere we want multiple file upload.

ThiefMaster: Do we ever specify a file limit that's not the default of 10 or 1? If not I'd get rid of `max_files` and just add a `multiple` boolean setting...

nop33: I would keep the `max_files` defaulted to 10. It's nice in case you want to configure it. But right now we can just use the new `multiple` option in all the `FileField` cases.

indico/web/forms/fields/files.py
nop33: Are you sure this is the right thing to do atm? It will require changes in all the controllers where the `FileField` is used, many of them use mixins and base classes to handle the file uploads

ThiefMaster: ok then let's do it another time

indico/modules/attachments/forms.py
```diff
@@ -62,12 +63,12 @@ class EditAttachmentFormBase(AttachmentFormBase):
 
 
 class AddAttachmentFilesForm(AttachmentFormBase):
-    files = HiddenField(_("Files"), widget=DropzoneWidget(submit_form=False))
+    files = FileField(_("Files"), multipleFiles=True)
```
ThiefMaster: Why is this kwarg camelcased? This is python, it shouldn't use camelcase ;)

indico/modules/events/abstracts/operations.py
```diff
@@ -0,0 +1,64 @@
...
+        abstract_file.save(f.file)
+        db.session.flush()
+    signals.event.abstract_created.send(abstract)
+    logger.info('Abstract %s created by %s', abstract, session.user)
+    abstract.event_new.log(EventLogRealm.management, EventLogKind.positive, 'Abstracts',
+                           'Abstract "{}" has been created'.format(abstract.title), session.user)
+    return abstract
+
+
+def delete_abstract(abstract, include_contrib):
```
ThiefMaster: default to `False` and rename to `delete_contrib`

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -155,3 +174,84 @@ class RHAbstractListStaticURL(RHAbstractListBase):
...
+            if tpl_components.get('hide_abstract'):
+                self.list_generator.flash_info_message(abstract)
+            return jsonify_data(**tpl_components)
+        return jsonify_template('events/abstracts/forms/abstract.html', event=self.event_new, form=form)
+
+
+class RHDeleteAbstracts(RHManageAbstractsActionsBase):
+    def _process(self):
+        include_contribs = request.values.get('include_contribs') == '1'
+        contrib_counter = 0
```
ThiefMaster: `num_contribs`? Also, why not something like this?

    num_contribs = sum(1 for a in self.abstracts if a.contribution) if include_contribs else 0

indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -34,68 +34,108 @@
...
                                 data-friendly-id="{{ abstract.friendly_id }}"
-                                data-title="{{ abstract.as_legacy.getTitle() }}">
+                                data-title="{{ abstract.title }}">
                                 <td class="i-table id-column">
-                                    <span class="vertical-aligner">
-                                        <input type="checkbox" class="select-row" name="abstract_id"
-                                               value="{{ abstract.id }}">
-                                    </span>
+                                    <input type="checkbox" class="select-row" name="abstract_id"
+                                           value="{{ abstract.id }}">
```
ThiefMaster: fits in the previous line ;)

indico/modules/attachments/forms.py
nop33: Scumbag brain, messed up the JS arguments with the python ones

indico/modules/events/abstracts/templates/management/_abstract_list.html
nop33: Well, what's the convention in HTML? We don't have a rule and in any case, it's above 180 chars.

ThiefMaster: ah.. looked shorter. we don't have a strict limit as we do in python but we usually try to keep around <=120 too. for 124 i would decide case by case whether wrapping makes it more or less readable. in this case i don't have a strong opinion so i'd probably just keep it as it is now

indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -34,68 +34,108 @@
...
                                 </td>
                                 <td class="i-table id-column">
-                                    <span class="vertical-aligner">{{ abstract.friendly_id }}</span>
+                                    {{ abstract.friendly_id }}
                                 </td>
                                 <td class="i-table">
-                                    <span class="vertical-aligner js-mathjax">{{ abstract.as_legacy.getTitle() }}</span>
+                                    <a class="js-mathjax"
+                                       href="{{ url_for('abstracts.manage_abstract', abstract) }}">
```
ThiefMaster: this one does fit in the previous line! ;)

indico/modules/events/abstracts/templates/management/abstract_list_filter.html
```diff
@@ -31,6 +32,19 @@
...
 
+{% macro _render_boolean_filter(id=none, name=none, label=none) %}
+    <div id="form-group-tracks_required" class="form-group">
+        {% if label %}
+            <div class="form-label">
+                <label for="{{ id }}">{{ label }}</label>
+            </div>
+        {% endif %}
+        <div class="form-field">
+            {{ switch(id=id, name=name, checked=filters['extra'][id] | bool) }}
```
ThiefMaster: i think without spaces around the `|` it's more readable here (or `()` around the whole `filters['extra'][id] | bool` but i'd go for removing these two spaces instead)

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -463,9 +432,24 @@ def _process(self):
...
 
+    def _prepare_folder_structure(self, attachment):
+        registration = attachment.registration
+        regform_title = secure_filename(attachment.registration.registration_form.title, 'registration_form')
+        registrant_name = secure_filename("{}_{}".format(registration.get_full_name(),
+                                          unicode(registration.friendly_id)), registration.friendly_id)
+        file_name = secure_filename("{}_{}_{}".format(attachment.field_data.field.title, attachment.field_data.field_id,
+                                                      attachment.filename), attachment.filename)
+        path = os.path.join(*self._adjust_path_length([regform_title, registrant_name, file_name]))
+        return path
```
ThiefMaster: i'd just return it in the previous line. no need in assigning it just to return it and not do anything else with it

indico/modules/events/surveys/forms.py
```diff
@@ -112,8 +112,6 @@ class TextForm(IndicoForm):
...
 
 class ImportQuestionnaireForm(IndicoForm):
-    json_file = HiddenField(_('File'),
-                            description=_("Choose a previously exported survey content to import."
-                                          " Existing sections will be preserved."),
-                            widget=DropzoneWidget(max_files=1, submit_form=False, submit_if_empty=False,
-                                                  accepted_file_types="application/json,.json"))
+    json_file = FileField(_('File'), accepted_file_types="application/json,.json",
+                          description=_("Choose a previously exported survey content to import."
+                                        " Existing sections will be preserved."))
```
ThiefMaster: usually we put the space in the previous line (`import. "`)

indico/modules/events/util.py
```diff
@@ -346,6 +356,9 @@ def store_configuration(self):
             self.list_config['items'] = sorted(visible_items)
         session.modified = True
 
+    def flash_info_message(self, contrib):
```
ThiefMaster: why is the argument called `contrib` here? it's the base class

PR #2600, opened 2016-09-15
Fix search filter in logs

DavidAndreev:


indico/modules/events/logs/templates/entry_email.html
```diff
@@ -53,7 +53,7 @@
         <td class="i-table caption log-caption">
             {% trans %}Body{% endtrans %}
         </td>
-        <td class="i-table value searchable">
+        <td class="i-table value ">
```
ThiefMaster: extra space

ThiefMaster:
Merged 2b69e89

PR #2601, opened 2016-09-16
Remove method for auto sending emails for users with no Indico account

DavidAndreev:


indico/modules/events/timetable/controllers/legacy.py
```diff
@@ -194,12 +195,14 @@ def _process(self):
...
                                          day=tt_entry_dt.date(), session_block=parent_session_block)
+            non_users_before = find_unregistered_users(contrib.speakers)
             if form.validate_on_submit():
                 with track_time_changes(auto_extend=True, user=session.user) as changes:
                     update_contribution(contrib, *_get_field_values(form.data))
+                flash_if_unregistered(non_users_before, contrib.speakers, self.event_new)
                 notifications = get_time_changes_notifications(changes, tzinfo=self.event_new.tzinfo, entry=self.entry)
                 return jsonify_data(update=serialize_entry_update(self.entry, session_=self.session),
-                                    notifications=notifications, flash=False)
+                                    notifications=notifications, flash=True)
```
ThiefMaster: `flash=True` is the default so you can just omit the argument

indico/modules/events/management/util.py
```diff
@@ -182,3 +183,38 @@ def _process_attachments(folders):
...
+    """Get a set of users which don't have an Indico account
+
+        :param person_list: List of persons to check
+
+    """
+    return set(filter(lambda p: p.person.user is None, person_list))
+
+
+def flash_if_unregistered(former_person_list, new_person_list, event):
+    """ Flash a message when user(s) with no Indico account is added for a contribution or as a chairperson
```
ThiefMaster: this line is way too long for a docstring. try to limit it to 72 chars, if it needs to be longer start it in the next line (only start it in the `"""` line if it fits there).

also, too much whitespace

indico/modules/events/management/util.py
```diff
@@ -182,3 +183,38 @@ def _process_attachments(folders):
...
+
+        :param former_person_list: List of persons before new users were added
```
ThiefMaster: this should not be indented (check the style of existing docstrings in new code)

https://www.python.org/dev/peps/pep-0257/

indico/modules/events/management/util.py
```diff
@@ -182,3 +183,38 @@ def _process_attachments(folders):
...
         raise TypeError('Unexpected object of type {}: {}'.format(type(root).__name__, root))
+
+
+def find_unregistered_users(person_list):
+    """Get a set of users which don't have an Indico account
+
+        :param person_list: List of persons to check
+
+    """
+    return set(filter(lambda p: p.person.user is None, person_list))
```
ThiefMaster: don't use filter with lambda, use a list comprehension instead. actually, here you want a set comprehension:

`{pl for pl in person_list if pl.person.user is None}`

I don't think we really need a function for it...

indico/modules/events/util.py
```diff
@@ -142,27 +142,6 @@ def get_random_color(event):
...
-    elif acl_entry.has_management_role('submit', explicit=True):
-        template_name = 'events/emails/pending_submitter.txt'
-        endpoint = 'event.conferenceDisplay'
-    else:
-        return
-    event = acl_entry.event_new
-    email = acl_entry.principal.email
-    template = get_template_module(template_name, event=event, email=email,
-                                   url=url_for_register(url_for(endpoint, event), email=email))
-    send_email(make_email(to_list={email}, template=template), event.as_legacy, module='Protection')
```
ThiefMaster: I think these imports are unused now. Remove them if they are.

indico/modules/events/management/util.py
DavidAndreev: Well I use it in 4 other places and in each you pass different list so in each it would have to be different, I think it makes sense to have a function and I think its also easier to understand what it does by looking at function name rather then just looking at the code

indico/modules/events/management/util.py
```diff
@@ -182,3 +183,38 @@ def _process_attachments(folders):
...
+def flash_if_unregistered(former_person_list, new_person_list, event):
+    """ Flash a message when user(s) with no Indico account is added for a contribution or as a chairperson
+
+        :param former_person_list: List of persons before new users were added
+        :param new_person_list: List of persons with new users
+        :param event: Current event
+
+    """
+    added_non_users = len(find_unregistered_users(new_person_list) - former_person_list)
+    if added_non_users > 0:
```
DavidAndreev: Im making a set from difference of the 2 sets (new_person_list and former_person_list), to see what are the new people with no account, and then I need length of that set to display the number, so I dont know how to do it without len()

indico/modules/events/management/util.py
```diff
@@ -182,3 +183,38 @@ def _process_attachments(folders):
...
+        msg = _(' An Indico account may be needed to upload materials and/or manage contents. ')
+        if event.can_access(session.user):
+            continue_msg = ngettext('To send an invitation email for this user, please go to the Roles section',
+                                    'To send invitation emails for these users, please go to the Roles section',
+                                    added_non_users)
+        else:
+            continue_msg = ngettext('Contact the manager if you think this user should be able to access these '
```
DavidAndreev: Yeah I agree that its not the best but I cant think of anything better, any suggestions ?

indico/modules/events/contributions/controllers/management.py
```diff
@@ -200,14 +201,16 @@ def _process(self):
...
         if form.validate_on_submit():
             with track_time_changes():
                 update_contribution(self.contrib, *_get_field_values(form.data))
+            flash_if_unregistered(non_users_before, self.contrib.speakers, self.event_new)
             flash(_("Contribution '{}' successfully updated").format(self.contrib.title), 'success')
             tpl_components = self.list_generator.render_list(self.contrib)
             if tpl_components['hide_contrib']:
                 self.list_generator.flash_info_message(self.contrib)
-            return jsonify_data(**tpl_components)
+            return jsonify_data(flash='flash' in request.args, **tpl_components)
```
ThiefMaster: I'd rather do `flash=(request.args.get('flash') == '1')` - even if we don't do it, I always find it weird that `flash=0` would be the exact same as `flash=1` in the query string.

indico/modules/events/contributions/templates/management/_contribution_list.html
```diff
@@ -184,7 +184,7 @@
                             <td class="i-table actions-column">
                                 <div class="group right entry-action-buttons vertical-aligner">
                                     <a href="#" class="icon-edit i-button-icon js-dialog-action"
-                                       data-href="{{ url_for('.manage_update_contrib', contrib) }}"
+                                       data-href="{{ url_for('.manage_update_contrib', contrib, flash = True) }}"
```
ThiefMaster: `flash=true` (kwargs have no space around the `=` and jinja uses/prefers lowercase true/false/none, even though uppercase also works)

indico/modules/events/management/util.py
```diff
@@ -182,3 +183,35 @@ def _process_attachments(folders):
...
+def find_unregistered_users(person_list):
+    """Get a set of users which don't have an Indico account
+
+    :param person_list: List of persons to check
+    """
+    return {pl for pl in person_list if pl.person.user is None}
+
+
+def flash_if_unregistered(former_person_list, new_person_list, event):
+    """Flash a message when user(s) with no Indico account is added
```
ThiefMaster: `"""Flash message when adding users with no indico account`

More concise and sounds less weird

indico/modules/events/management/controllers.py
```diff
@@ -265,8 +265,10 @@ class RHManageEventPersonLinks(RHConferenceModifBase):
 
     def _process(self):
         form = EventPersonLinkForm(obj=self.event_new, event=self.event_new, event_type=self.event_new.type)
+        non_users_before = find_unregistered_users(self.event_new.person_links)
         if form.validate_on_submit():
             update_event(self.event_new, form.data)
+            flash_if_unregistered(non_users_before, self.event_new.person_links, self.event_new)
```
ThiefMaster: I would change this function to act as a [context manager](https://docs.python.org/2/library/contextlib.html#contextlib.contextmanager):

```python
with flash_if_unregistered(self.event_new, self.event_new.person_links):
    update_event(self.event_new, form.data)
```

Implementation-wise, you'd do something like this:

```python
@contextmanager
def flash_if_unregistered(...):
    old_non_users = ...
    yield
    new_non_users = ...
    # now do what you currently do (compare the two lists and flash if needed)
```

indico/modules/events/management/util.py
```diff
@@ -182,3 +183,35 @@ def _process_attachments(folders):
 
     else:
         raise TypeError('Unexpected object of type {}: {}'.format(type(root).__name__, root))
+
+
+def find_unregistered_users(person_list):
```
ThiefMaster: `person_links` since that's what you actually pass here

indico/modules/events/management/util.py
```diff
@@ -182,3 +183,35 @@ def _process_attachments(folders):
...
+
+def flash_if_unregistered(former_person_list, new_person_list, event):
+    """Flash a message when user(s) with no Indico account is added
+
+    :param former_person_list: List of persons before new users were added
+    :param new_person_list: List of persons with new users
+    :param event: Current event
+    """
+    added_non_users = len(find_unregistered_users(new_person_list) - former_person_list)
+    if added_non_users:
```
ThiefMaster:
```
if not added_non_users:
    return
```

That way you have less indentation on the whole rest of this function

indico/modules/events/management/util.py
```diff
@@ -182,3 +183,35 @@ def _process_attachments(folders):
...
+        msg = _(' An Indico account may be needed to upload materials and/or manage contents. ')
```
ThiefMaster: Don't put surrounding whitespace here. It makes translation ugly and translators are likely to remove it in their translations.

indico/modules/events/management/util.py
```diff
@@ -182,3 +183,35 @@ def _process_attachments(folders):
...
+        if event.can_manage(session.user):
+            continue_msg = ngettext('To send an invitation email for this user, please go to the Roles section',
+                                    'To send invitation emails for these users, please go to the Roles section',
+                                    added_non_users)
+        else:
+            continue_msg = ngettext('Please contact the manager if you think this user should be able to access these '
+                                    'features.',
+                                    'Please contact the manager if you think these users should be able to access '
+                                    'these features.', added_non_users)
+        flash(ngettext('You have added a user with no Indico account. ',
```
ThiefMaster: `flash(' '.join([xxx, msg, continue_message]))` (find a decent name for `xxx`)

indico/modules/events/persons/__init__.py
```diff
@@ -32,3 +32,16 @@
 def _sidemenu_items(sender, event, **kwargs):
     if event.can_manage(session.user):
         return SideMenuItem('persons', _('Roles'), url_for('persons.person_list', event), section='reports')
+
+
+@signals.get_placeholders.connect_via('roles-email')
```
ThiefMaster: "roles" is just the "displayed" name of this feature, internally it's "event persons", so i'd call this `event-persons-email`

indico/modules/events/persons/controllers.py
```diff
@@ -152,21 +148,32 @@ def _checkParams(self, params):
     def _process(self):
         person_ids = request.form.getlist('person_id')
         recipients = {p.email for p in self._find_event_persons(person_ids) if p.email}
-        form = EmailEventPersonsForm(person_id=person_ids, recipients=', '.join(recipients))
+        if 'no_account' in request.args:
```
ThiefMaster: `request.args.get('no_account') == '1'`

indico/modules/events/persons/controllers.py
```diff
@@ -152,21 +148,32 @@ def _checkParams(self, params):
...
+            tpl = get_template_module('events/registration/emails/pending_submitter.html', event=self.event_new)
+            default_body = tpl.get_html_body()
+            form = EmailEventPersonsForm(person_id=person_ids, recipients=', '.join(recipients), body=default_body)
+        else:
+            form = EmailEventPersonsForm(person_id=person_ids, recipients=', '.join(recipients))
```
ThiefMaster: no need to repeat this, just set `body = ''` before the `if` and always pass it to this constructor

indico/modules/events/persons/controllers.py
```diff
@@ -152,21 +148,32 @@ def _checkParams(self, params):
...
+                                                 event=self.event_new)
+            email = make_email(to_list=recipient.email, from_address=form.from_address.data,
+                               body=email_body, subject=email_subject, html=True)
+            send_email(email, self.event_new, 'Event Persons')
+
     def _find_event_persons(self, person_ids):
         return (self.event_new.persons
-                .filter(EventPerson.id.in_(person_ids), EventPerson.email != '')
-                .all())
+                .filter(EventPerson.id.in_(person_ids), EventPerson.email != '').all())
```
ThiefMaster: Please keep this in separate lines, it's more readable to have no more than one chain element in each line when wrapping queries over multiple lines.

indico/modules/events/persons/controllers.py
```diff
@@ -152,21 +148,32 @@ def _checkParams(self, params):
     def _process(self):
         person_ids = request.form.getlist('person_id')
         recipients = {p.email for p in self._find_event_persons(person_ids) if p.email}
-        form = EmailEventPersonsForm(person_id=person_ids, recipients=', '.join(recipients))
+        if 'no_account' in request.args:
+            tpl = get_template_module('events/registration/emails/pending_submitter.html', event=self.event_new)
+            default_body = tpl.get_html_body()
```
ThiefMaster: I would also provide a default subject

indico/modules/events/management/util.py
```diff
@@ -182,3 +186,40 @@ def _process_attachments(folders):
...
+                             check
+    :param event: Current event
+    """
+    old_non_users = find_unregistered_users(get_person_links())
+    yield
+    new_non_users = find_unregistered_users(get_person_links())
+    added_non_users = len(new_non_users - old_non_users)
+    if not added_non_users:
+        return
+    warning = ngettext('You have added a user with no Indico account. ',
```
OmeGak: Trailing whitespace.

indico/modules/events/management/util.py
```diff
@@ -182,3 +186,40 @@ def _process_attachments(folders):
 
     else:
         raise TypeError('Unexpected object of type {}: {}'.format(type(root).__name__, root))
+
+
+def find_unregistered_users(person_links):
```
OmeGak: `get_unregistered_users(...)`

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -88,6 +88,12 @@
                 }
             }
         });
+        if ($('#person-list [data-no-account]').length > 0){
+            $('#no-account-dropdown').toggleClass('disabled', false);
```
ThiefMaster: `.removeClass('disabled')`

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -88,6 +88,12 @@
                 }
             }
         });
+        if ($('#person-list [data-no-account]').length > 0){
```
ThiefMaster: space before `{`

indico/modules/events/management/controllers.py
```diff
@@ -30,12 +30,12 @@
 from indico.modules.events.contributions.models.subcontributions import SubContribution
 from indico.modules.events.forms import EventReferencesForm, EventLocationForm, EventPersonLinkForm, EventKeywordsForm
 from indico.modules.events.management.forms import EventProtectionForm
-from indico.modules.events.management.util import can_lock
```
ThiefMaster: Why did you remove this import? It's used in this file. :bomb:

indico/modules/events/management/util.py
```diff
@@ -182,3 +186,40 @@ def _process_attachments(folders):
 
     else:
         raise TypeError('Unexpected object of type {}: {}'.format(type(root).__name__, root))
+
+
+def get_unregistered_users(person_links):
+    """Get a set of users which don't have an Indico account
+
+    :param person_links: List of persons to check
```
ThiefMaster: List of *person links* to check

indico/modules/events/management/util.py
```diff
@@ -182,3 +186,40 @@ def _process_attachments(folders):
 
     else:
         raise TypeError('Unexpected object of type {}: {}'.format(type(root).__name__, root))
+
+
+def get_unregistered_users(person_links):
```
ThiefMaster: This is only used in this file. I'd underscore it and remove the docstring. It's not really useful elsewhere...

Considering how simple (one line) it is you could even just get rid of that function altogether. It's used only twice anyway.

indico/modules/events/management/util.py
```diff
@@ -182,3 +186,40 @@ def _process_attachments(folders):
...
+    return {pl for pl in person_links if pl.person.user is None}
+
+
+@contextmanager
+def flash_if_unregistered(event, get_person_links):
+    """Flash message when adding users with no indico account
+
+    :param get_person_links: Callable returning list of person links to
+                             check
+    :param event: Current event
```
ThiefMaster: move this up to match the param order of the function

indico/modules/events/persons/__init__.py
```diff
@@ -32,3 +32,17 @@
 def _sidemenu_items(sender, event, **kwargs):
     if event.can_manage(session.user):
         return SideMenuItem('persons', _('Roles'), url_for('persons.person_list', event), section='reports')
+
+
+@signals.get_placeholders.connect_via('event-persons-email')
+def _get_placeholders(sender, person, event, **kwargs):
```
ThiefMaster: `def _get_placeholders(sender, person, event, register_link=False, **kwargs):` (and adapt the check below etc accordingly)

We only accept `**kwargs` in signals so they don't break if someone adds extra arguments when calling them - but arguments that are actually used in a signal handler should always be explicitly specified.

indico/modules/events/persons/controllers.py
```diff
@@ -151,17 +148,38 @@ def _checkParams(self, params):
 
     def _process(self):
         person_ids = request.form.getlist('person_id')
-        recipients = {p.email for p in self._find_event_persons(person_ids) if p.email}
-        form = EmailEventPersonsForm(person_id=person_ids, recipients=', '.join(recipients))
+        recipients = self._find_event_persons(person_ids)
+        if request.args.get('not_invited_only') == '1':
+            recipients = {p for p in recipients if p.invited_dt is None}
```
ThiefMaster: Why not add an argument for this to `_find_event_persons`? You can use `EventPerson.invited_dt.is_(None)` in a SQLAlchemy filter so you don't even query those persons from the DB

indico/modules/events/persons/controllers.py
```diff
@@ -151,17 +148,38 @@ def _checkParams(self, params):
...
+        default_body = ''
+        default_subject = ''
+        disable_until_change = True
```
ThiefMaster: `disabled_until_change`

indico/modules/events/persons/controllers.py
```diff
@@ -151,17 +148,38 @@ def _checkParams(self, params):
...
+        if request.args.get('no_account') == '1':
```
ThiefMaster: `no_account = request.args.get('no_account') == '1'` and use it here and also below. Right now you use two different checks for the same argument which aren't even equal in behavior

indico/modules/events/persons/controllers.py
```diff
@@ -151,17 +148,38 @@ def _checkParams(self, params):
...
             num = len(recipients)
             flash(ngettext('Your email has been sent.', '{} emails have been sent.', num).format(num))
-            return jsonify_data()
-        return jsonify_form(form, submit=_('Send'))
+            return jsonify_data(flash=False)
+        return jsonify_form(form, submit=_('Send'), disabled_until_change=disable_until_change)
+
+    def _send_emails(self, form, recipients):
+        for recipient in recipients:
+            if request.args.get('no_account') == '1':
```
ThiefMaster: oh, another place where this is used. Maybe it'd be a good idea to override `_checkParams` (don't forget to call the parent one) and assign `self.no_account = ...` so you can use it here too and not just in `_process`

indico/modules/events/persons/controllers.py
```diff
@@ -151,17 +148,38 @@ def _checkParams(self, params):
...
+                recipient.invited_dt = now_utc()
+            email_body = replace_placeholders('event-persons-email', form.body.data, person=recipient,
+                                              event=self.event_new)
```
ThiefMaster: you are not specifying whether to allow the register link placeholder or not here

indico/modules/events/persons/forms.py
```diff
@@ -40,6 +41,8 @@ def __init__(self, *args, **kwargs):
         from_addresses = ['{} <{}>'.format(session.user.full_name, email)
                           for email in sorted(session.user.all_emails, key=lambda x: x != session.user.email)]
         self.from_address.choices = zip(from_addresses, from_addresses)
+        self.body.description = render_placeholder_info('event-persons-email', event=None, person=None,
+                                                        no_register_link=kwargs.get('no_register_link'))
```
ThiefMaster: pop it from kwargs before calling super to avoid passing the kwarg to the superclass which doesn't even know about it (it won't break anything to pass it but it's cleaner)

indico/modules/events/persons/templates/management/_person_list_base.html
```diff
@@ -72,6 +72,37 @@
...
+                               title="{% trans %}Send an email to all users with no Indico account who have not
+                                     received an invitation email yet{% endtrans %}"
+                               data-href="{{ url_for('persons.email_event_persons', event, no_account = True,
+                                                      not_invited_only = True) }}"
+                               data-params-selector="#persons-list [data-no-account] input[name=person_id]:checkbox"
+                               data-title="{% trans %}Send email{% endtrans %}"
+                               data-method="POST"
+                               data-ajax-dialog
+                               data-reload-after>
+                                {% trans %}Email not invited yet{% endtrans %}</a>
```
ThiefMaster: that text sounds weird. anyone has better ideas on how to label this button? cc @OmeGak  @nop33  @pferreir

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -54,9 +54,16 @@
...
-                {% trans %}No Indico account{% endtrans %}
+                   account. If they do have an account, they should add {{ email }} to their profile{%- endtrans %}">
+                   {% trans %}No Indico account{% endtrans %}
+            </span>
+        {% endif %}
+    </td>
+    <td class="i-table">
+        {% if person_data.person.invited_dt %}
+            <span class="i-icon icon-mail"
+                   title="{% trans email=person.email %}Email for registration was sent{%- endtrans %}
```
ThiefMaster: you don't use `email` inside the translated string

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -54,9 +54,16 @@
...
+                    {{ person_data.person.invited_dt| format_date(timezone=person_data.person.event_new.tzinfo) }}">
```
ThiefMaster: put a space on both sides of the `|` or don't put one at all

indico/modules/events/persons/templates/management/_person_list_row.html
ThiefMaster: I think this would be better:
```
{% trans date=person_data.person.invited_dt|format_date(timezone=person_data.person.event_new.tzinfo) -%}
Invited on {{ date }}
{%- endtrans %}
```

indico/modules/events/persons/templates/management/person_list.html
```diff
@@ -61,7 +61,14 @@
         </thead>
         <tbody>
             {% for person_data in persons -%}
-                <tr id="person-{{ person_data.person.id }}" class="i-table" data-person-roles="{{ person_data.roles | tojson | forceescape }}">
+                <tr id="person-{{ person_data.person.id }}" class="i-table"
+                    {% if person_data.roles.no_account %}
+                        data-no-account
+                    {% endif %}
+                    {% if person_data.person.invited_dt %}
+                        data-invited-already
```
ThiefMaster: `data-already-invited`?

indico/modules/events/persons/templates/management/person_list.html
```diff
@@ -61,7 +61,14 @@
...
+                    {% endif %}
+                    data-person-roles="{{ person_data.roles | tojson | forceescape }}" >
```
ThiefMaster: move this up before the conditional ones. also, get rid of that extra space at the end ;)

indico/modules/events/persons/controllers.py
```diff
@@ -151,17 +148,38 @@ def _checkParams(self, params):
...
-        recipients = {p.email for p in self._find_event_persons(person_ids) if p.email}
-        form = EmailEventPersonsForm(person_id=person_ids, recipients=', '.join(recipients))
+        recipients = self._find_event_persons(person_ids)
+        if request.args.get('not_invited_only') == '1':
+            recipients = {p for p in recipients if p.invited_dt is None}
+        default_body = ''
+        default_subject = ''
+        disable_until_change = True
+        if request.args.get('no_account') == '1':
+            tpl = get_template_module('events/registration/emails/pending_submitter.html', event=self.event_new)
```
ThiefMaster: This file doesn't belong in the registration module - "registration" has nothing to do with registering indico accounts but with registering for an event. While user registration is handled by the "auth" module, this email is part of "persons" so it should be in there.

Not sure about the naming of the file either. What about `events/persons/emails/invitation.html`?

indico/modules/events/persons/templates/management/_person_list_base.html
```diff
@@ -72,6 +72,37 @@
...
+                               data-href="{{ url_for('persons.email_event_persons', event, no_account = True) }}"
+                               data-params-selector="#persons-list [data-no-account] input[name=person_id]:checkbox"
+                               data-title="{% trans %}Send email{% endtrans %}"
+                               data-method="POST"
+                               data-ajax-dialog
+                               data-reload-after>
+                                {% trans %}Email all{% endtrans %}</a>
+                        </li>
+                        <li>
+                            <a id="not-invited-only" href="#"
```
ThiefMaster: I'd keep `href` as the very first attribute

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -88,6 +88,12 @@
...
+            $('#no-account-dropdown').toggleClass('disabled', false);
+        }
+        if ($('#person-list [data-no-account]').length - $('#person-list [data-invited-already]').length === 0){
```
ThiefMaster: Why not just check if the lengths are equal instead of subtracting them?

indico/modules/events/timetable/controllers/legacy.py
```diff
@@ -113,13 +114,15 @@ def _process(self):
         defaults = self._get_form_defaults(location_parent=self.session_block)
         form = ContributionEntryForm(obj=defaults, to_schedule=True, **self._get_form_params())
         if form.validate_on_submit():
-            with track_time_changes(auto_extend=True, user=session.user) as changes:
+            contrib = Contribution()
+            with flash_if_unregistered(self.event_new, lambda: contrib.speakers),\
```
ThiefMaster: I think using two nested `with` blocks is more readable here

indico/modules/events/management/controllers.py
DavidAndreev: Sorry, it must have been shown as unused

indico/modules/events/persons/templates/management/_person_list_base.html
```diff
@@ -72,6 +72,37 @@
...
                 {% endif %}
+                <div class="group">
+                    <a id="no-account-dropdown" href="#" class="i-button arrow highlight disabled js-dropdown" data-toggle="dropdown">
+                        {% trans %}Users with no account{% endtrans %}
+                    </a>
+                    <ul class="dropdown">
+                        <li>
+                            <a href="#"
+                               title="{% trans %}Send an email to all users with no Indico account{% endtrans %}"
+                               data-href="{{ url_for('persons.email_event_persons', event, no_account = True) }}"
```
ThiefMaster: `no_account=true` - bools/none are lowercase in jinja and kwargs have no space around the `=`

indico/modules/events/persons/templates/management/_person_list_base.html
```diff
@@ -72,6 +72,37 @@
...
+                               data-method="POST"
+                               data-ajax-dialog
+                               data-reload-after>
+                                {% trans %}Email all{% endtrans %}</a>
+                        </li>
+                        <li>
+                            <a id="not-invited-only" href="#"
+                               title="{% trans %}Send an email to all users with no Indico account who have not
+                                     received an invitation email yet{% endtrans %}"
+                               data-href="{{ url_for('persons.email_event_persons', event, no_account = True,
```
ThiefMaster: https://github.com/indico/indico/pull/2601/files#r80866472

indico/modules/events/management/controllers.py
```diff
@@ -42,6 +42,7 @@
 from indico.web.util import jsonify_template, jsonify_data, jsonify_form
 from MaKaC.webinterface.rh.base import RH
 from MaKaC.webinterface.rh.conferenceModif import RHConferenceModifBase
+from indico.modules.events.management.util import can_lock
```
ThiefMaster: Now the imports are not sorted properly ;x

indico/modules/events/models/acls_test.py
```diff
@@ -147,31 +147,17 @@ def _signal_fn(sender, **kwargs):
...
 @pytest.mark.usefixtures('request_context')
-def test_update_principal_resolve_email(create_event, create_user, smtp):
+def test_update_principal_resolve_email(create_event, create_user):
     event = create_event()
     user = create_user(123, email='user@example.com')
     # add email that belongs to a user
     entry = event.update_principal(EmailPrincipal('user@example.com'), full_access=True)
     assert entry.principal == user
     assert entry.type == PrincipalType.user
-    extract_emails(smtp, required=False, count=0)
```
ThiefMaster: You can also remove the import of `extract_emails` from this file

indico/modules/events/persons/forms.py
```diff
@@ -40,6 +41,9 @@ def __init__(self, *args, **kwargs):
         from_addresses = ['{} <{}>'.format(session.user.full_name, email)
                           for email in sorted(session.user.all_emails, key=lambda x: x != session.user.email)]
         self.from_address.choices = zip(from_addresses, from_addresses)
+        register_link = kwargs.pop('register_link')
```
ThiefMaster: pop it *before* the `super()` call

indico/modules/events/persons/__init__.py
```diff
@@ -32,3 +32,17 @@
 def _sidemenu_items(sender, event, **kwargs):
     if event.can_manage(session.user):
         return SideMenuItem('persons', _('Roles'), url_for('persons.person_list', event), section='reports')
+
+
+@signals.get_placeholders.connect_via('event-persons-email')
+def _get_placeholders(sender, person, event, register_link=True, **kwargs):
```
ThiefMaster: I'd really default to `False` and only set it to `True` in cases where you need it.

indico/modules/events/persons/controllers.py
```diff
@@ -147,23 +144,47 @@ class RHEmailEventPersons(RHConferenceModifBase):
...
-    def _find_event_persons(self, person_ids):
+            return jsonify_data(flash=False)
+        return jsonify_form(form, submit=_('Send'), disabled_until_change=disabled_until_change)
+
+    def _send_emails(self, form, recipients):
+        for recipient in recipients:
+            if self.no_account:
+                recipient.invited_dt = now_utc()
+            email_body = replace_placeholders('event-persons-email', form.body.data, person=recipient,
+                                              event=self.event_new)
```
ThiefMaster: enable the `register_link` only if necessary

indico/modules/events/persons/templates/management/_person_list_base.html
```diff
@@ -72,6 +73,37 @@
...
+                               data-ajax-dialog
+                               data-reload-after>
+                                {% trans %}Email all{% endtrans %}</a>
+                        </li>
+                        <li>
+                            <a href="#" id="not-invited-only"
+                               title="{% trans %}Send an email to all users with no Indico account who have not
+                                     received an invitation email yet{% endtrans %}"
+                               data-href="{{ url_for('persons.email_event_persons', event, no_account=true,
+                                                      not_invited_only = True) }}"
```
ThiefMaster: `not_invited_only=true`

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -54,9 +54,17 @@
...
-                   account. If they do have an account, they should add {{ email }} to their profile
+                   account. If they do have an account, they should add {{ email }} to their profile{%- endtrans %}">
+                   {% trans %}No Indico account{% endtrans %}
+            </span>
+        {% endif %}
+    </td>
+    <td class="i-table">
+        {% if person_data.person.invited_dt %}
+            <span class="i-icon icon-mail"
+                   title="{% trans date=person_data.person.invited_dt | format_date(timezone=person_data.person.event_new.tzinfo) -%}
```
ThiefMaster: too much indentation

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -54,9 +54,17 @@
...
             <span class="i-label warning icon-warning"
                    title="{% trans email=person.email %}This user seems to have no Indico
-                   account. If they do have an account, they should add {{ email }} to their profile
+                   account. If they do have an account, they should add {{ email }} to their profile{%- endtrans %}">
+                   {% trans %}No Indico account{% endtrans %}
+            </span>
+        {% endif %}
+    </td>
+    <td class="i-table">
+        {% if person_data.person.invited_dt %}
```
ThiefMaster: in line 2 we assign `person` so you can replace `person_data.person` with just `person` here and two lines below

indico/modules/events/persons/templates/management/person_list.html
```diff
@@ -61,7 +61,15 @@
         </thead>
         <tbody>
             {% for person_data in persons -%}
-                <tr id="person-{{ person_data.person.id }}" class="i-table" data-person-roles="{{ person_data.roles | tojson | forceescape }}">
+                <tr id="person-{{ person_data.person.id }}" class="i-table"
+
```
ThiefMaster: why that empty line here?

indico/modules/events/management/controllers.py
OmeGak: Maybe you could start using a decent text editor. :P

indico/modules/events/persons/templates/management/_person_list_base.html
OmeGak: "Invite all" / "Invite pending" with an `icon-mail`?

indico/modules/events/persons/__init__.py
DavidAndreev: Well I tried it with false more times but then there was some problem with render method or something (it was complaining about arguments), and in 2 cases we need it true and in only one its false so I think this logic makes sense ;)

indico/modules/events/persons/templates/management/_person_list_base.html
ThiefMaster: "Invite pending" is not very clear. I wouldn't associate "pending" with "people not emailed yet"...

OmeGak: Maybe having a screenshot to understand the UI could help getting some ideas. cc @DavidAndreev

indico/modules/events/contributions/controllers/management.py
```diff
@@ -182,7 +183,9 @@ def _process(self):
         contrib_form_class = make_contribution_form(self.event_new)
         form = contrib_form_class(obj=FormDefaults(location_data=inherited_location), event=self.event_new)
         if form.validate_on_submit():
-            contrib = create_contribution(self.event_new, *_get_field_values(form.data))
+            contrib = Contribution()
```
ThiefMaster: why?

ThiefMaster: ok, now i see. you need a contribution from which you get the empty speakers list. it's a bit ugly. maybe add a comment so it's clear why you need this?

    contrib = Contribution()  # so we have an empty initial person list in flash_if_unregistered

indico/modules/events/contributions/controllers/management.py
```diff
@@ -201,13 +204,13 @@ def _process(self):
                                                    **custom_field_values),
                                   event=self.event_new, contrib=self.contrib, session_block=parent_session_block)
         if form.validate_on_submit():
-            with track_time_changes():
+            with track_time_changes(), flash_if_unregistered(self.event_new, lambda: self.contrib.speakers):
```
ThiefMaster: `self.contrib.person_links` (in any place you use `speakers` atm)

indico/modules/events/persons/templates/management/_person_list_base.html
DavidAndreev: screenshot from 2016-09-29 13-33-02

OmeGak: Does emailing them imply inviting them as well?

ThiefMaster: Only when using the "users with no account" button

indico/modules/events/contributions/controllers/management.py
```diff
@@ -182,7 +183,10 @@ def _process(self):
         contrib_form_class = make_contribution_form(self.event_new)
         form = contrib_form_class(obj=FormDefaults(location_data=inherited_location), event=self.event_new)
         if form.validate_on_submit():
-            contrib = create_contribution(self.event_new, *_get_field_values(form.data))
+            # Creating empty contribution so it can be compared to the created one in flash_if_unregistered
```
ThiefMaster: "the created one" sounds weird. I'd rather go for "the new one" and "Create" instead of "Creating"

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -54,9 +54,18 @@
         {% if person_data.roles.no_account %}
             <span class="i-label warning icon-warning"
                    title="{% trans email=person.email %}This user seems to have no Indico
-                   account. If they do have an account, they should add {{ email }} to their profile
-                   {%- endtrans %}">
-                {% trans %}No Indico account{% endtrans %}
+                   account. If they do have an account, they should add {{ email }} to their profile{%- endtrans %}">
+                   {% trans %}No Indico account{% endtrans %}
```
ThiefMaster: too much indentation

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -54,9 +54,18 @@
...
+            </span>
+        {% endif %}
+    </td>
+    <td class="i-table">
+        {% if person.invited_dt %}
+            <span class="i-label icon-mail"
+                  title="This person has been invited to create an Indico account">
+                  {% trans date=person.invited_dt | format_date(timezone=person_data.person.event_new.tzinfo) -%}
```
ThiefMaster: also too much indentation

Issue #1968, opened 2015-07-10
Can't upload files with non-ascii names (like Chinese filename)


xfuren:
At least I can't upload in material management page

ASCII file name -> successfully uploaded
UTF8 file name -> nothing happened

ThiefMaster:
Please provide some more information...

- Which version of Indico are you using?
- Are you getting any errors? (indico.log, browser console)

xfuren:
1. I encountered this issue both on your demo site (indico-sandbox.cern.ch)
and my own 1.2 setup.
2. I saw nothing on my setup, but the following error in the browser
console on tour demo site.

POST
https://indico-sandbox.cern.ch/event/102/manage/material/add?conference=102&parentProtected=false
500 (INTERNAL SERVER ERROR)doSubmit @ jquery_code_67c9365f.min.js:2053

ThiefMaster:
I think the issue is gone in our development version (1.9.x) - can you check in the test category on http://indico.cern.ch if that works fine for you?

pferreir:
Or rather http://indico-sandbox.cern.ch.

ThiefMaster:
sandbox is still running 1.9.. we should probably update it at some point

pferreir:
Yeah, but the issue shouldn't exist there. Should it?

ThiefMaster:
This is the exception that caused it to fail:

```
2015-07-13 08:26:44,558  ERROR    7f69cc6a5ed0  indico                    Request <IndicoRequest 'https://indico-sandbox.cern.ch/event/102/manage/material/add?conference=102&parentProtected=false' [POST]> finished with UnicodeDecodeError: 'ascii' codec can't decode byte 0xe8 in position 0: ordinal not in range(128)
Traceback (most recent call last):
  File "/opt/indico/env/lib/python2.7/site-packages/indico-1.9-py2.7.egg/MaKaC/webinterface/rh/base.py", line 581, in process
    profile_name, res = self._process_retry(params, i, profile, forced_conflicts)
  File "/opt/indico/env/lib/python2.7/site-packages/indico-1.9-py2.7.egg/MaKaC/webinterface/rh/base.py", line 530, in _process_retry
    return self._process_retry_do(profile)
  File "/opt/indico/env/lib/python2.7/site-packages/indico-1.9-py2.7.egg/MaKaC/webinterface/rh/base.py", line 500, in _process_retry_do
    self._checkParams(self._reqParams)
[...]

PR #2608, opened 2016-09-21
v1.2 Handle UnicodeDecodeError at encoding uploaded filename

wtakase:
This patch may fix https://github.com/indico/indico/issues/1968.


pferreir:
Hi!
Thanks for the patch, but this kind of `try...except` solution tends to cause more harm thank good. It's fine as a quick way to solve your issue, but not as a permanent fix.

We should instead try to find out how we end up with a UTF8 string in code that expects a unicode object (or, otherwise, why are we even expecting a unicode object?)

This problem is fixed anyway in future Indico versions (namely 2.0). So, if you've got a fix that works for you and no time to provide a more complete solution, it may be better to wait for it.

PR #2611, opened 2016-09-29
Abstract page

OmeGak:


indico/modules/events/abstracts/models/reviews.py
```diff
@@ -137,3 +139,9 @@ class AbstractReview(db.Model):
     @return_ascii
     def __repr__(self):
         return format_repr(self, 'id', 'abstract_id', 'user_id', proposed_action=None)
+
+    @property
+    def score(self):
+        if not self.ratings:
+            return None
+        return sum((x.value for x in self.ratings)) / len(self.ratings)
```
ThiefMaster: Not sure if this logic is correct. We may need to take the scale into account. If you have a scale from 1 to 5 we want the score in that range.

What we store in the rating is not the value on the scale (that's often an ugly float depending on the scale) but the index of the selected option (since you could change the scale after there are already reviews and we don't want to have to re-calculate all those values in the DB).

indico/modules/events/abstracts/templates/_abstract.html
```diff
@@ -99,6 +111,24 @@
     </div>
 {% endmacro %}
 
+{% macro _render_summary_state(abstract) %}
+    {% if abstract.public_state.name == 'under_review' %}
+        {% set tag_type = 'highlight' %}
```
ThiefMaster: Why not use a mapping dict?

indico/modules/events/abstracts/templates/_abstract.html
```diff
@@ -303,3 +331,35 @@ <h3 class="abstract-title">
...
+                            <div class="stretch">
+                                {% trans name=comment.user.name, date=comment.created_dt|format_human_date -%}
+                                    <strong>{{ name }}</strong> commented on {{ date }}
+                                {%- endtrans %}
+                                {% if comment.modified_dt %}
+                                    {{ render_edited_hint(comment.modified_dt, comment.modified_by) }}
+                                {% endif %}
+                            </div>
+                            <div class="abstract-comment-badges">
+                                {% if session.user == comment.user or abstract.event_new.can_manage(session.user) %}
```
ThiefMaster: Add a `comment.can_edit(user)` method instead that performs this check. We need it when editing on the server too so we'd replicate protection logic in two places. Guess how likely it is that someone only modifies one of them in the future ;)

indico/modules/events/abstracts/templates/management/abstract.html
```diff
@@ -1,6 +1,143 @@
...
+        {{ _render_abstract_reviews(abstract) }}
+        <div class="i-timeline-separator"></div>
+        <div class="i-timetable-connect-up"></div>
+        {{ _render_decision_box(abstract) }}
+    </div>
 {% endblock %}
+
+{% macro _render_abstract_reviews(abstract) %}
+    {% for track in abstract.proposed_for_tracks %}
+        {% set reviews = track.abstract_reviews|selectattr('abstract', 'equalto', abstract)|list %}
```
ThiefMaster: Wouldn't it be cleaner to get the reviews of the abstract and filter by track? There are most likely less reviews on an abstract than there are for a whole track.

indico/modules/events/abstracts/models/reviews.py
OmeGak: I was assuming we store the value within the scale. If that's not the case we have to adapt this, for sure.

indico/modules/events/abstracts/templates/management/abstract.html
OmeGak: Another problem, I just realized, is that reviews without track won't be listed.

indico/modules/events/abstracts/templates/management/abstract.html
```diff
@@ -1,6 +1,144 @@
...
+{% endmacro %}
+
+{% macro _render_review_state(review) %}
+    {% set mapping = {'accept': 'success',
+                      'reject': 'error',
+                      'merge': 'visited',
+                      'mark_as_dupe': 'strong'} %}
+    <span class="i-tag outline {{ mapping[review.proposed_action.name] }}">
+        {{ review.proposed_action.title }}
+        {#- -#}
```
ThiefMaster: why? you can do it as `-}}` in the previous line

indico/htdocs/sass/base/_theme.scss
```diff
@@ -14,3 +14,67 @@
...
+
+    &.error {
+        background: $red;
+    }
+
+    &.warning {
+        background: $yellow;
+    }
+
+    &.visited {
```
ThiefMaster: Is that really what we had decided on last week? I thought we'd picked something else...

indico/modules/events/abstracts/templates/management/abstract.html
OmeGak: To make it explicit. Otherwise a small padding gets added. Maybe I should explain it in the comment itself.

ThiefMaster: But we don't do it explicitly anywhere else except if it's needed because there's no jinja stuff where the `-` can be added

indico/modules/events/abstracts/templates/management/abstract.html
```diff
@@ -1,6 +1,180 @@
...
+        <div class="i-timeline-item-box content-indicator">
+            <div class="i-box-content">
+                <div class="flex">
+                    <div class="stretch">
+                        <strong>
+                            {% trans %}Review summary{% endtrans %}
+                        </strong>
+                    </div>
+                    <div class="abstract-comment-badges">
+                        {% trans %}Reviews are{% endtrans %}
```
ThiefMaster: `{% trans state=_render_abstract_reviewing_state(abstract) %}Reviews are {{ state }}{% endtrans %}`

Not super pretty either but less awful to translate if it's not "<Reviews are> <xxx>" in that order in some other language. If you change it make sure to test it, I don't know if you need `|safe` here. Probably not but just in case better test it

indico/modules/events/abstracts/templates/_abstract.html
```diff
@@ -303,3 +324,35 @@ <h3 class="abstract-title">
...
+                    <div class="i-timeline-item-box header-indicator">
+                        <div class="i-box-header flex">
+                            <div class="stretch">
+                                {% trans name=comment.user.name, date=comment.created_dt|format_human_date -%}
+                                    <strong>{{ name }}</strong> commented on {{ date }}
+                                {%- endtrans %}
+                                {% if comment.modified_dt %}
+                                    {{ render_edited_hint(comment.modified_dt, comment.modified_by) }}
+                                {% endif %}
+                                {% if comment.can_edit(user) %}
```
ThiefMaster: yuck! why not close the `div` here and then start the new one in the if can_edit block?

indico/modules/events/abstracts/templates/management/abstract.html
```diff
@@ -1,6 +1,180 @@
...
+        <div class="i-timetable-connect-up"></div>
+        {{ _render_decision_box(abstract) }}
+    </div>
 {% endblock %}
+
+{% macro _render_abstract_reviews(abstract) %}
+    <div id="abstract-reviews" class="i-timeline">
+        {% set reviews = abstract.reviews|selectattr('track', 'equalto', none)|list %}
+        {% if reviews %}
+            <h3>No track</h3>
```
ThiefMaster: missing i18n

indico/modules/events/abstracts/templates/management/abstract.html
```diff
@@ -1,6 +1,180 @@
...
+                    {% trans name=review.user.name, date=review.created_dt|format_human_date -%}
+                        <strong>{{ name }}</strong> reviewed on {{ date }}
+                    {%- endtrans %}
+                    {% if review.modified_dt %}
+                        {{ render_edited_hint(comment.modified_dt) }}
+                    {% endif %}
+                </div>
+                <div class="abstract-comment-badges">
+                    <span>
+                        {% trans %}Proposes to{% endtrans %}
```
ThiefMaster: similar solution as in https://github.com/indico/indico/pull/2611/files#r81586082 maybe?

indico/modules/events/abstracts/templates/_abstract.html
OmeGak: Ew! How the heck did I end up with this :hankey:?

indico/modules/events/abstracts/templates/management/abstract.html
```diff
@@ -1,6 +1,182 @@
...
+                <h3>
+                    {% trans title=track.title -%}
+                        Track: {{ title }}
+                    {%- endtrans %}
+                </h3>
+                {% for review in reviews %}
+                    {{ _render_review(review) }}
+                {% endfor %}
+            {% endif %}
+        {% else %}
```
ThiefMaster: :monkey:

PR #2612, opened 2016-09-29
Add settings dialog for abstract reviewing

ThiefMaster:


indico/modules/events/abstracts/forms.py
```diff
@@ -74,6 +74,29 @@ def validate_contrib_type_required(self, field):
             raise ValidationError(_('The event has no contribution types defined.'))
 
 
+class AbstractReviewingSettingsForm(IndicoForm):
+    """Settings form for abstract reviewing"""
+
+    num_answers = IntegerField(_('Number of choices'), [NumberRange(min=2), DataRequired()],
```
OmeGak: I think `num_answers` is wrong naming (inherited from legacy IIRC). It's not how many answers you can give to a question, but how many choices are given as an answer (as in the label and description).

indico/modules/events/abstracts/forms.py
```diff
@@ -74,6 +74,29 @@ def validate_contrib_type_required(self, field):
...
+                               description=_('The number of choices reviewers have for each question.'))
+    scale_lower = IntegerField(_('Scale (from)'), [InputRequired()])
+    scale_upper = IntegerField(_('Scale (to)'), [InputRequired()])
+    conveners_final_judgment = BooleanField(_('Allow track conveners to judge'), widget=SwitchWidget(),
```
OmeGak: `conveners_judment`. There's only one type of judgment. I would even go for `allow_conveners_to_judge`.

indico/modules/events/abstracts/forms.py
```diff
@@ -74,6 +74,29 @@ def validate_contrib_type_required(self, field):
...
+                                            description=_('Enabling this allows track conveners to make a final '
```
OmeGak: Same here. No need for **final**, I'd say.

indico/modules/events/abstracts/forms.py
```diff
@@ -74,6 +74,29 @@ def validate_contrib_type_required(self, field):
             raise ValidationError(_('The event has no contribution types defined.'))
 
 
+class AbstractReviewingSettingsForm(IndicoForm):
+    """Settings form for abstract reviewing"""
+
+    num_answers = IntegerField(_('Number of choices'), [NumberRange(min=2), DataRequired()],
+                               description=_('The number of choices reviewers have for each question.'))
+    scale_lower = IntegerField(_('Scale (from)'), [InputRequired()])
```
OmeGak: How often are these used? Fiddling with the number of choices and scale will cause the values of the choices to be fractional unless carefully adjusted. Maybe we can hide this complexity with a `custom scale` checkbox. If not checked, the scale adjusts so that the values of the choices start at **1**, go up to `num_choices`, and have increments of **1**.

PR #2613, opened 2016-09-30
Name format setting

DavidAndreev:


indico/modules/users/__init__.py
```diff
@@ -41,6 +41,7 @@
     'timezone': None,
     'force_timezone': False,  # always use the user's timezone instead of an event's timezone
     'show_past_events': False,
+    'name_ordering': NameOrdering.first_last,
```
ThiefMaster: Right now retrieving this setting will return just a plain integer instead of the actual enum value. This often works fine but we now support properly handling enums in a SettingsProxy using the [EnumConverter](https://github.com/indico/indico/blob/devel/indico/core/settings/converters.py#L77). Just search for it and you'll find an example how it's used.

indico/modules/users/forms.py
```diff
@@ -59,6 +59,9 @@ class UserPreferencesForm(IndicoForm):
         widget=SwitchWidget(),
         description=_('Show past events by default.'))
 
+    name_ordering = IndicoEnumSelectField(_('Name ordering'), enum=NameOrdering,
```
ThiefMaster: "Name format" might be clearer

ThiefMaster:
You commited lots of changes related to the setting itself in the second commit so having two separate commits like this is not very useful anymore. Two ways to fix this:

- squash them into a single one (interactive rebase, mark the second one as fixup or squash)
- (better IMO) remove the commits but keep the changes (`git reset devel`), then `git add -p` the changes and commit them with proper separation (i.e. first commit for the setting etc, second one to use the new property)

indico/modules/events/templates/display/indico/_common.html
```diff
@@ -73,6 +72,7 @@
 {% endmacro %}
 
 {% macro render_users(user_list, span_class='', title=true, italic_affiliation=false, separator=', ') -%}
+    {{ name_format }}
```
ThiefMaster: :monkey: this looks like test code you forgot to remove ;)

indico/modules/users/models/users.py
```diff
@@ -122,6 +140,20 @@ def full_name(self):
...
+    upper = name_format in (NameFormat.first_last_upper, NameFormat.f_last_upper, NameFormat.last_f_upper,
+                            NameFormat.last_first_upper)
+    if name_format in (NameFormat.first_last, NameFormat.first_last_upper):
+        return obj.get_full_name(last_name_first=False, last_name_upper=upper, abbrev_first_name=False)
+    elif name_format in (NameFormat.last_first, NameFormat.last_first_upper):
+        return obj.get_full_name(last_name_first=True, last_name_upper=upper, abbrev_first_name=False)
+    elif name_format in (NameFormat.last_f, NameFormat.last_f_upper):
+        return obj.get_full_name(last_name_first=True, last_name_upper=upper, abbrev_first_name=True)
+    elif name_format in (NameFormat.f_last, NameFormat.f_last_upper):
+        return obj.get_full_name(last_name_first=False, last_name_upper=upper, abbrev_first_name=True)
```
ThiefMaster:
```
else:
    raise ValueError('Invalid name format: {}'.format(name_format))
```

Otherwise you'd just return `None` in such a case (which should never happen anyway)

indico/modules/users/__init__.py
```diff
@@ -44,6 +45,8 @@
     'use_previewer_pdf': True,
     'synced_fields': None,  # None to synchronise all fields, empty set to not synchronize
     'suggest_categories': False  # whether the user should receive category suggestions
+}, converters={
+    'name_format': EnumConverter(NameFormat.first_last)
```
ThiefMaster: Why `.first_last`? you need to specify the enum here, not a (default) value...

indico/modules/users/models/users.py
```diff
@@ -107,6 +120,11 @@ def full_name(self):
         """Return the person's name in 'Firstname Lastname' notation."""
         return self.get_full_name(last_name_first=False, last_name_upper=False, abbrev_first_name=False)
 
+    @property
+    def display_full_name(self):
+        """Return the person's name in format specified in user settings"""
```
ThiefMaster: """Return the full name using the user's preferred name format."""

indico/modules/users/models/users.py
```diff
@@ -56,6 +56,19 @@ def from_legacy(cls, text):
         return next((x for x in cls if unicode(x.title) == text), None)
 
 
+class NameFormat(TitledIntEnum):
+    __titles__ = (_('John Doe'), _('Doe, John'), _('Doe, J.'), _('J. Doe'),
+                  _('John DOE'), _('DOE, John'), _('DOE, J.'), _('J. DOE'))
+    first_last = 0
```
OmeGak: Am I the only one finding `first_last` quite confusing?

ThiefMaster: I don't think it's really possible to find great names for these options (`first_name_last_name` etc quickly becomes *very* long and verbose) - and it's not used in many places so you can always check the title/implementation to see how a format works.

DavidAndreev: I agree it can be confusing at first but if you look at others it should start making sense

OmeGak: `¯\_(ツ)_/¯`

ThiefMaster:
FYI I rebased this PR to the latest devel and force-pushed your branch. If you change something in it make sure to update your local branch after checking it out using `git fetch origin` and then `git reset --hard origin/name-ordering-setting`

indico/modules/users/models/users.py
OmeGak: Not a strong opinion, but I would rather have long variables than not clear variables that end up requiring comments.

PR #2623, opened 2016-10-05
Fix events at hand - long ongoing events

DavidAndreev:


indico/modules/users/controllers.py
```diff
@@ -128,7 +128,7 @@ def _process(self):
         linked_events = [(event, {'management': bool(roles & self.management_roles),
                                   'reviewing': bool(roles & self.reviewer_roles),
                                   'attendance': bool(roles & self.attendance_roles)})
-                         for event, roles in get_linked_events(self.user, from_dt, None, 10).iteritems()]
+                         for event, roles in get_linked_events(self.user, from_dt, now_utc(False), 10).iteritems()]
```
ThiefMaster: Now you won't get any events that will happen in the future...

Two ways of fixing it:

1) not specifying and end time and changing the `happens_between` logic if only a `start_dt` is present
2) specifying an `end_dt` far enough in the future. Maybe 3 or 6 months? Not sure if it makes much sense to show an event that starts e.g. a year from now anyway.

I'd ping others on IRC for an opinion on this.

PR #2626, opened 2016-10-05
Abstract list PDF generation

nop33:


ThiefMaster:
You probably need to add `.encode('utf-8')` in quite a few places. Try creating an abstract with non-ascii data everywhere and chances are good things go :bomb:

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -240,55 +231,46 @@ def __init__(self, abstract, tz=None):
...
+            st = ''
+            track_review_state = abstract.get_track_reviewing_state(track)
+            if track_review_state == AbstractReviewingState.positive:
+                track_reviews = abstract.get_track_reviews(track)
+                # TODO: Handle different contribution type proposals
+                contrib_type = track_reviews[0].proposed_contribution_type
+                contrib_type_title = contrib_type.name
+                if not all(track_review.proposed_contribution_type == contrib_type for track_review in track_reviews):
+                    contrib_type_title = _("Multiple contribution types")
+                st = _("Proposed to accept: {}").format(contrib_type_title)
```
OmeGak: Shouldn't we use the same naming as in the UI? That would be using `AbstractReviewingState.positive.title`.

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -240,55 +231,46 @@ def __init__(self, abstract, tz=None):
...
+            elif track_review_state == AbstractReviewingState.negative:
                 st = _("Proposed to reject")
-
-            elif isinstance(status, review.AbstractInConflict):
+            elif track_review_state == AbstractReviewingState.mixed:
                 st = _("Conflict")
```
OmeGak: This is not correct. Mixed is just mixed. In conflict is restricted to when a contribution is accepted in different tracks and does not apply to track reviewing state.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -419,6 +419,9 @@ def get_track_reviewing_state(self, track):
         else:
             return AbstractReviewingState.mixed
 
+    def get_track_reviews(self, track):
```
OmeGak: Do we really need this? It doesn't add much and I don't think we want to fill the models with trivial methods.

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -240,55 +231,46 @@ def __init__(self, abstract, tz=None):
...
+        if abstract.state == AbstractState.duplicate:
+            return _("{} ({}: {})").format(state_title, abstract.duplicate_of.friendly_id, abstract.duplicate_of.title)
+        elif abstract.state == AbstractState.merged:
+            return _("{} ({}: {})").format(state_title, abstract.merged_into.friendly_id, abstract.merged_into.title)
         else:
-            return AbstractStatusList.getInstance().getCaption(status.__class__).upper()
+            return abstract.state.title.upper()
 
     @staticmethod
     def _get_track_judgements(abstract):
```
nop33: well, if I start renaming legacy methods and classes, this will never end :stuck_out_tongue: 
I just went for adapting the necessary code for it to work. It's gonna be rewritten after 2.0.

indico/MaKaC/PDFinterface/conference.py
nop33: @ThiefMaster , @pferreir what do you think?

indico/modules/events/abstracts/models/abstracts.py
nop33: Well, I need it for the PDF generation. And I don't think there's a place more appropriate than the model.

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -240,55 +231,46 @@ def __init__(self, abstract, tz=None):
...
+        for track in abstract.reviewed_for_tracks:
+            st = ''
+            track_review_state = abstract.get_track_reviewing_state(track)
+            if track_review_state == AbstractReviewingState.positive:
+                track_reviews = abstract.get_track_reviews(track)
+                # TODO: Handle different contribution type proposals
+                contrib_type = track_reviews[0].proposed_contribution_type
+                contrib_type_title = contrib_type.name
+                if not all(track_review.proposed_contribution_type == contrib_type for track_review in track_reviews):
+                    contrib_type_title = _("Multiple contribution types")
```
OmeGak: Let's discuss this IRL.

indico/modules/events/abstracts/models/abstracts.py
OmeGak: But it's just a simple list comprehension.

indico/MaKaC/PDFinterface/conference.py
ThiefMaster: I don't think backward compatibility matters that much here - it's not an API after all :)

indico/MaKaC/PDFinterface/conference.py
OmeGak: Only to prevent confusion since now we refer to a different thing with judgment.

nop33: Yeah, you're right, it's just one function

indico/modules/events/abstracts/models/abstracts.py
nop33: We have the same list comprehension in `get_track_reviewing_state` though. So when deciding whether I want to have 2 lines of code repeated in 2 different places, I preferred to introduce the new method. If the majority believes it's better with 2 list comprehensions, fine by me!

ThiefMaster: no strong opinion, but for a simple list comprehension in 2 places i'd probably keep the list comprehensions.

nop33: If there won't be any other use of this method, I'll gladly remove it. Are you sure there won't be any other place where we'll need to get the list of abstract reviews for a specific track?

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -182,48 +183,37 @@ def __init__(self, abstract, tz=None):
 
     @staticmethod
     def _get_track_classification(abstract):
-        status = abstract.getCurrentStatus()
-        if isinstance(status, review.AbstractStatusAccepted):
-            if status.getTrack() is not None:
-                return escape(status.getTrack().getTitle())
+        if abstract.state == AbstractState.accepted:
+            return escape(abstract.accepted_track.title)
```
ThiefMaster: fails if there is no track (which is perfectly valid). i guess the check someone said we can remove yesterday wasn't so useless after all ;)

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -182,48 +183,38 @@ def __init__(self, abstract, tz=None):
 
     @staticmethod
     def _get_track_classification(abstract):
-        status = abstract.getCurrentStatus()
-        if isinstance(status, review.AbstractStatusAccepted):
-            if status.getTrack() is not None:
-                return escape(status.getTrack().getTitle())
+        if abstract.state == AbstractState.accepted:
+            if abstract.accepted_track:
```
ThiefMaster: Except that this check was the other way round -- you *can* have accepted abstracts with no track

nop33:
Yesterday we agreed that in the "Track Reviews" section of the PDF export we will take the value from the AbstractState (Positive, Negative, Mixed, ...). This lead to showing also the "Not started" state. I think we should hide that. What do you think @ThiefMaster @OmeGak @pferreir ?
image

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -182,48 +183,38 @@ def __init__(self, abstract, tz=None):
...
+                return escape(abstract.accepted_track.title)
         else:
-            listTrack = []
-            for track in abstract.getTrackListSorted():
-                listTrack.append(escape(track.getTitle()))
-            return "; ".join(listTrack)
+            tracks = abstract.submitted_for_tracks | abstract.reviewed_for_tracks
+            return '; '.join(escape(t.title) for t in tracks)
```
ThiefMaster: `u'; '` for consistency (the titles are unicode strings)

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -235,65 +226,48 @@ def __init__(self, abstract, tz=None):
...
-        #style.leftIndent = 90
-        if isinstance(status, review.AbstractStatusDuplicated):
-            original = status.getOriginal()
-            return _("DUPLICATED ({0}: {1})").format(original.getId(), original.getTitle())
-        elif isinstance(status, review.AbstractStatusMerged):
-            target = status.getTargetAbstract()
-            return _("MERGED ({0}: {1})").format(target.getId(), target.getTitle())
+        state_title = abstract.state.title.upper()
+        if abstract.state == AbstractState.duplicate:
+            return _("{} ({}: {})").format(state_title, abstract.duplicate_of.friendly_id, abstract.duplicate_of.title)
```
ThiefMaster: `_(u'...')`

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -235,65 +226,48 @@ def __init__(self, abstract, tz=None):
...
+        elif abstract.state == AbstractState.merged:
+            return _("{} ({}: {})").format(state_title, abstract.merged_into.friendly_id, abstract.merged_into.title)
```
ThiefMaster: `_(u'...')`

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -235,65 +226,48 @@ def __init__(self, abstract, tz=None):
...
+                track_reviews = abstract.get_track_reviews(track)
+                proposed_contrib_types = {r.proposed_contribution_type.name for r in track_reviews}
+                contrib_types = ', '.join(proposed_contrib_types)
+                review_state = '{}: {}'.format(review_state, contrib_types)
+            elif track_review_state == AbstractReviewingState.mixed:
+                track_reviews = abstract.get_track_reviews(track)
+                other_tracks = {x for r in track_reviews for x in r.proposed_tracks}
+                if other_tracks:
+                    review_state = _("{}: Proposed for other tracks: {}").format(review_state, ', '.join(other_tracks))
+            judgements.append((track.title, review_state))
```
ThiefMaster: rename to `judgments`? it's just a local var so it's a quick change

nop33: well, I should rename it to `reviews`...

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -182,48 +183,38 @@ def __init__(self, abstract, tz=None):
...
+        if abstract.state == AbstractState.accepted:
+            if abstract.accepted_track:
+                return escape(abstract.accepted_track.title)
         else:
-            listTrack = []
-            for track in abstract.getTrackListSorted():
-                listTrack.append(escape(track.getTitle()))
-            return "; ".join(listTrack)
+            tracks = sorted(abstract.submitted_for_tracks | abstract.reviewed_for_tracks, key=attrgetter('position'))
+            return '; '.join(escape(t.title) for t in tracks)
```
ThiefMaster: `u'; '`

PR #2627, opened 2016-10-05
Fix broken layout after undoing category sorting

DavidAndreev:


ThiefMaster:
Cherrypicked it into v1.9.8 (8eb66af) and merged that into master.

PR #2628, opened 2016-10-05
Add abstract picker field

ThiefMaster:
Here's a patch that adds the field to the reviewing settings form just for testing it:

```patch
diff --git a/indico/modules/events/abstracts/controllers/management.py b/indico/modules/events/abstracts/controllers/management.py
index b318f93..860f5aa 100644
--- a/indico/modules/events/abstracts/controllers/management.py
+++ b/indico/modules/events/abstracts/controllers/management.py
@@ -127,8 +127,11 @@ class RHManageAbstractReviewing(RHManageAbstractsBase):
         defaults = FormDefaults(abstract_review_questions=self.event_new.abstract_review_questions,
                                 **abstracts_reviewing_settings.get_all(self.event_new))
         form = AbstractReviewingSettingsForm(event=self.event_new, obj=defaults, has_ratings=has_ratings)
-        if form.validate_on_submit():
+        if form.process_ajax():
+            return form.ajax_response
+        elif form.validate_on_submit():
             data = form.data
+            flash('Selected abstract: {}'.format(data.pop('test')))
             # XXX: we need to do this assignment for new questions,
             # but editing or deleting existing questions changes an
             # object that is already in the session so it's updated
diff --git a/indico/modules/events/abstracts/forms.py b/indico/modules/events/abstracts/forms.py
index c5bfc22..ee14f6a 100644
--- a/indico/modules/events/abstracts/forms.py
+++ b/indico/modules/events/abstracts/forms.py
@@ -96,6 +96,7 @@ class AbstractReviewingSettingsForm(IndicoForm):
                                            description=_('Enabling this allows track conveners to make a judgment '
                                                          'such as accepting or rejecting an abstract.'))
     abstract_review_questions = AbstractReviewQuestionsField(_('Review questions'))
+    test = AbstractField('XXX')

```

### DIFF
--- a/indico/htdocs/js/indico/widgets/selectize_widget.js
+++ b/indico/htdocs/js/indico/widgets/selectize_widget.js
@@ -56,8 +56,9 @@
             score: function(query) {
                 var data = getSearchData(query);
                 if (data && data.id !== undefined) {
+                    // when searching by ID ensure we don't get other results from selectize's internal cache
                     return function(item) {
-                        return item.friendly_id === +data.id;
+                        return +(item.friendly_id === +data.id);
                     };
                 } else {
                     var scoreFunc = this.getScoreFunction(query);
--- a/indico/htdocs/js/indico/widgets/selectize_widget.js
+++ b/indico/htdocs/js/indico/widgets/selectize_widget.js
@@ -23,26 +23,58 @@
             fieldId: null,
             minTriggerLength: 0,
             searchUrl: null,
-            selectizeOptions: null
+            selectizeOptions: null,
+            allowById: false
         }, options);
+
+        function getSearchData(query) {
+            var m;
+            if (options.allowById && (m = query.match(/^#(\d+)$/))) {
+                return {id: m[1]};
+            } else if (query.length >= options.minTriggerLength) {
+                return {q: query};
+            } else {
+                return null;
+            }
+        }
 
         var params = {
             load: function(query, callback) {
-                if (query.length < options.minTriggerLength) {
+                var data = getSearchData(query);
+                if (!data) {
                     return callback();
                 }
                 $.ajax(options.searchUrl, {
-                    data: {q: query},
+                    data: data,
                     cache: false
                 }).fail(function() {
                     callback();
                 }).done(function(res) {
                     callback(res);
                 });
+            },
+            score: function(query) {
+                var data = getSearchData(query);
+                if (data && data.id !== undefined) {
+                    return function(item) {
+                        return item.friendly_id === +data.id;
+                    };
+                } else {
+                    var scoreFunc = this.getScoreFunction(query);
+                    return function(item) {
+                        return scoreFunc(item);
+                    };
+                }
             }
         };
 
         _.extend(params, options.selectizeOptions);
-        $('#' + options.fieldId).selectize(params);
+        var $field = $('#' + options.fieldId);
+        $field.selectize(params);
+
+        // disallow removing options
+        $field[0].selectize.removeOption = function() {
+            return false;
+        };
     };
 })(window);
--- a/indico/modules/events/abstracts/fields.py
+++ b/indico/modules/events/abstracts/fields.py
@@ -16,18 +16,24 @@
 
 from __future__ import unicode_literals
 
+from flask import jsonify, request, session
+from werkzeug.exceptions import BadRequest
+from wtforms.ext.sqlalchemy.fields import QuerySelectField
+
+from indico.core.db.sqlalchemy.util.queries import escape_like
 from indico.core.db.sqlalchemy.util.session import no_autoflush
+from indico.modules.events.abstracts.models.abstracts import Abstract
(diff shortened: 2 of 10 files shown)


indico/htdocs/js/indico/widgets/selectize_widget.js
```diff
@@ -23,26 +23,57 @@
...
+                    return function(item) {
+                        return scoreFunc(item);
+                    };
+                }
             }
         };
 
         _.extend(params, options.selectizeOptions);
-        $('#' + options.fieldId).selectize(params);
+        var field = $('#' + options.fieldId);
```
nop33: We usually write `$field` when the variable is a jQuery object.

ThiefMaster: I think that's mostly a matter of preference :) But yeah, I've seen it quite often lately

indico/modules/events/abstracts/fields.py
```diff
@@ -135,3 +141,54 @@ def pre_validate(self, form):
...
+            query = query.filter(Abstract.id != self.excluded_abstract.id)
+        return query.order_by(Abstract.friendly_id)
+
+    def _get_object_list(self):
+        return [(key, abstract)
+                for key, abstract in super(AbstractField, self)._get_object_list()
+                if abstract.can_access(session.user)]
+
+    @property
+    def event(self):
```
OmeGak: Just out of curiosity. Why not simply call `self.get_form().event` where needed?

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -386,7 +386,22 @@ def can_access(self, user):
             return True
         if any(x.person.user == user for x in self.person_links):
             return True
-        return False
+        return self.can_review(user)
+
+    def can_review(self, user):
+        if not user:
+            return False
+        elif self.event_new.can_manage(user):
```
OmeGak: Is this so? I think we agreed that only people listed as reviewer would be able to review.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -386,7 +386,22 @@ def can_access(self, user):
             return True
         if any(x.person.user == user for x in self.person_links):
             return True
-        return False
+        return self.can_review(user)
+
+    def can_review(self, user):
```
OmeGak: I think we are gonna need `can_review(user, track)` as well.

indico/modules/events/abstracts/views.py
```diff
@@ -28,11 +28,13 @@ class WPManageAbstracts(WPJinjaMixin, WPConferenceModifBase):
     def getJSFiles(self):
         return (WPConferenceModifBase.getJSFiles(self) +
                 self._asset_env['markdown_js'].urls() +
+                self._asset_env['selectize_js'].urls() +
```
OmeGak: Why don't we have field-related JS files always included?

indico/htdocs/js/indico/widgets/selectize_widget.js
```diff
@@ -23,26 +23,57 @@
...
-                    data: {q: query},
+                    data: data,
                     cache: false
                 }).fail(function() {
                     callback();
                 }).done(function(res) {
                     callback(res);
                 });
+            },
+            score: function(query) {
```
OmeGak: What's this for?

indico/modules/events/abstracts/models/abstracts.py
ThiefMaster: wasn't completely sure about that, but makes sense

indico/htdocs/js/indico/widgets/selectize_widget.js
ThiefMaster: Otherwise selectize also searches its cache, so if you search for `#10` first and then for `#1` you also see `#10` even though we don't want partial matches when specifying an ID.

indico/modules/events/abstracts/views.py
ThiefMaster: we do that for some other fields, i just kept it as it was done before for this one

indico/modules/events/abstracts/models/abstracts.py
ThiefMaster: let's add it when we need it? in this case we don't really care about the track and it'd be easy to extend this method once we need it somewhere

indico/modules/events/abstracts/fields.py
ThiefMaster: No special reason, just found it nicer to go through a property. Also, like this it's clear why we don't do `self.event = self.get_form().event` in `__init__`

indico/modules/events/abstracts/views.py
ThiefMaster: I still hope to move away from separate js bundles in favor of one big indico.js/indico.css bundle for 2.0 so maybe it's something we can keep in mind for then.

PR #2629, opened 2016-10-06
Add calendar export to category list

DavidAndreev:


ThiefMaster:
Ugh, I just noticed that we do not specify any `from` argument in the ical URL... so it's super slow for the root category. And even with e.g. -7d..90d this takes ages (`/categ/0.ics?from=-7d&to=90d [4941 queries (11.116833s) | 165.991234s]`)

I'd either defer this until we have something decent for ical export (it makes no sense to even query notes etc when you are generating ical data) or add some extra code to the category exporter to return very lightweight data when it'll be serialized as ical.

indico/modules/categories/serialize.py
```diff
@@ -31,21 +31,23 @@
 from indico.web.flask.util import url_for
 
 
-def serialize_category_ical(category, user, event_filter):
+def serialize_category_ical(category_ids, user, event_filter=True, event_filter_fn=None, update_query=None):
```
OmeGak: Since we are now accepting several IDs the method should reflect, somehow, this multiplicity. What about `serialize_categories_ical`?

indico/modules/categories/serialize.py
```diff
@@ -31,21 +31,23 @@
...
     """Export the events in a category to iCal
 
-    :param category: The category to export
+    :param category_ids: Category IDs to export
     :param user: The user who needs to be able to access the events
     :param event_filter: A SQLalchemy criterion to restrict which
                          events will be returned.  Usually something
                          involving the start/end date of the event.
+    :param event_filter_fn: Event filter function to use when called from api.py
+    :param update_query: A query to update with when called from api.py
```
OmeGak: It looks to me `update_query` is a callable, not a query. Also, what exactly are we updating here?

indico/modules/events/api.py
```diff
@@ -111,6 +112,7 @@ class CategoryEventHook(HTTPAPIHook):
         'subcontributions': 500,
         'sessions': 100,
     }
+    NO_CACHE = True
```
OmeGak: Is this a debugging statement you forgot or are we stopping caching altogether for this hook?

indico/modules/categories/serialize.py
ThiefMaster: That docstring is not very good. You are explaining how you currently use it, not what the parameter does.

> A callable that can update the query used to retrieve the events. Must return the updated query object.

That'd be clearer

indico/modules/categories/controllers/display.py
```diff
@@ -400,7 +400,8 @@ def _process_PUT(self):
 class RHExportCategoryICAL(RHDisplayCategoryBase):
     def _process(self):
         filename = '{}-category.ics'.format(secure_filename(self.category.title, str(self.category.id)))
-        buf = serialize_category_ical(self.category, session.user, Event.end_dt >= (now_utc() - timedelta(weeks=4)))
+        buf = serialize_categories_ical([self.category.id], session.user,
+                                      Event.end_dt >= (now_utc() - timedelta(weeks=4)))
```
ThiefMaster: indentation

indico/modules/categories/serialize.py
```diff
@@ -55,7 +58,12 @@ def serialize_category_ical(category, user, event_filter):
                       own_room_strategy,
                       own_venue_strategy)
              .order_by(Event.start_dt))
-    events = [e for e in query if e.can_access(user)]
+    if update_query:
+        query = update_query(query)
+    events = query.all()
```
pferreir: Do we need `query.all()` here? From what I see we are iterating over it below.

pferreir:
Otherwise LGTM 👍

indico/modules/categories/serialize.py
ThiefMaster: Doesn't really make a difference - all rows are retrieved either way, with `.all()` the difference is that it puts them in a list immediately, otherwise the mapping of rows to objects happens once by one while iterating.

IRC we went for it since otherwise naming would be tricky since one of the iterations is conditional and reassigning query with something not a query would be ugly. `events = query` is slightly ugly too.

pferreir: In my understanding, by doing it the way I said the list would only be built once. `.all()` builds it once and then the list comprehension below does it over again. Not to mention that if there's an `event_filter_fn` you get an extra list coming form `filter` (as opposed to `ifilter` that would defer the check).
As for the naming, couldn't we just use `query` until we finally do `events = [...]`?
Anyway, the performance penalty is probably zero, not important, just wondering.

PR #2631, opened 2016-10-06
Small fixes and UI improvements

nop33:


indico/htdocs/sass/partials/_core.scss
```diff
@@ -64,7 +64,7 @@
 }
 
 .weak-hidden {
-    display: none;
+    display: none !important;
```
ThiefMaster: :bomb: This is going to break stuff! We have `.hidden` for this. weak-hidden is explicitly not important, hence "weak".

ThiefMaster: FWIW, I think using hidden or weak-hidden in combination with JS widgets is a bad idea. You can't `.show()`, `.hide()`, `.toggle()` those elements while you can do that if you hide them with the good old `style="display: none;""

indico/htdocs/js/indico/modules/list_generator.js
```diff
@@ -60,7 +60,8 @@
                 0: {
                     sorter: false
                 }
-            }
+            },
+            sortReset: true
```
ThiefMaster: more readable if you put it above the headers options (i.e. have the multiline dict ones last)

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -87,7 +88,9 @@ class RHAbstracts(RHManageAbstractsBase):
     """Display abstracts management page"""
 
     def _process(self):
-        return WPManageAbstracts.render_template('management/abstracts.html', self._conf, event=self.event_new)
+        abstracts_count = Abstract.query.with_parent(self.event_new).with_entities(db.func.count()).scalar()
```
ThiefMaster: why so complicated and not just `Abstract.query.with_parent(self.event_new).count()`? in registration we needed the more complex query since you had multiple regforms

indico/htdocs/sass/partials/_core.scss
ThiefMaster: if you don't need to change it from JS simply use `hidden`, otherwise inline CSS

PR #2633, opened 2016-10-06
Make author list look good

mvidalgarcia:
Adds ellipsis to long Affiliation names that were breaking the layout and adds scrolling when there are many authors (long lists in large events).

### DIFF
--- a/indico/htdocs/sass/modules/_event_management.scss
+++ b/indico/htdocs/sass/modules/_event_management.scss
@@ -490,14 +490,54 @@ th.i-table {
 }
 
 #person-list {
-    .affiliation-column, .name-column {
-        max-width: 20em;
-    }
-
     .no-role {
         opacity: 0.25;
         .badge {
             background-color: $dark-gray;
+        }
+    }
+
+    thead {
+        @include border-bottom();
+        display: block;
+        position: relative;
+        background-color: $light-gray;
+    }
+
+    tbody {
+        display: block;
+        max-height: 50em;
+        overflow: auto;
+    }
+
+    .checkbox-column {
+        min-width: 2.5em;
+    }
+
+    .name-column, .affiliation-column, .email-column {
+        max-width: 20em;
+        min-width: 20em;
+    }
+
+    .author-type-column {
+        width: 11em;
+    }
+
+    .speaker-column {
+        width: 2.1em;
+    }
+
+    .form-group-mail {
+        display: block;
+    }
+
+    tbody > tr:not(.hidden) {
+        td {
+            border-top-width: 0;
+        }
+
+        ~ tr td {
+            border-top-width: 1px;
         }
     }
 }
--- a/indico/htdocs/sass/modules/_sessions.scss
+++ b/indico/htdocs/sass/modules/_sessions.scss
@@ -27,7 +27,7 @@
     }
 }
 
-#persons-list {
+#person-list {
     .send-mails-btn {
         margin-top: 5px;
     }
--- a/indico/modules/events/management/templates/contribution_person_list.html
+++ b/indico/modules/events/management/templates/contribution_person_list.html
@@ -1,4 +1,4 @@
-<div id="persons-list" class="i-form no-max-width">
+<div id="person-list" class="i-form no-max-width">
     <div class="flashed-messages"></div>
     <div>
         <div class="toolbar form-group">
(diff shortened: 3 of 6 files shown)


nop33:
Add a screenshot! :smile:

indico/modules/events/management/templates/contribution_person_list.html
```diff
@@ -28,7 +28,7 @@
             </div>
         </div>
     </div>
-    <div class="form-group">
+    <div class="form-group table-content">
```
nop33: This way the `<thead>` is also scrolled, meaning that you cannot see the header after you scroll. In the particular case you can easily guess what each column represents, but in principal the header shouldn't scroll.

image

nop33: Also, the scroll bars shouldn't be visible when it's not necessary. Try `overflow: auto`

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -500,6 +500,11 @@ th.i-table {
             background-color: $dark-gray;
         }
     }
+
+    .table-content {
+        max-height: 50em;
+        overflow: scroll;
```
nop33: The scroll bars shouldn't be visible when it's not necessary. Try overflow: auto
(sorry for the other comment, wrong place!)

indico/htdocs/sass/modules/_sessions.scss
```diff
@@ -27,7 +27,7 @@
     }
 }
 
-#persons-list {
+#person-list {
```
OmeGak: Grammar nazi! :+1:

mvidalgarcia:
How it looks:

image

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -501,9 +501,38 @@ th.i-table {
         }
     }
 
-    .table-content {
+    thead {
+        display: block;
+        position: relative;
+        background-color: $light-gray;
+        border-bottom: 1px solid $gray;
```
OmeGak: Simply use `@include border-bottom();` like we do everywhere else with `border-all()`. It will apply by default `border-bottom: 1px solid $gray` and won't need any manual update if we ever change how borders should look Indico-wide.

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -501,9 +501,38 @@ th.i-table {
...
+    }
+
+    tbody {
+        display: block;
         max-height: 50em;
-        overflow: scroll;
+        overflow: auto;
+    }
+
+    td:nth-child(1), th:nth-child(1) { min-width: 2.5em; }  // checkbox
```
OmeGak: Better than having comments that will become obsolete, why not simply adding per-column classes? Otherwise if we at some point we decide to change the order of the columns we'll need to update this set of rules.

ThiefMaster: Is it an i-table? I think we already have a `.checkbox-column` class there...

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -490,16 +490,60 @@ th.i-table {
...
+        overflow: auto;
+    }
+
+    .checkbox-column {
+        min-width: 2.5em;
+    }
+
+    .name-column, .affiliation-column {
+        max-width: 20em;
+        min-width: 20em;
```
nop33: This is the same as adding `width: 20em;`.
Why do we need to specify the width of the columns?

nop33: For example, the `.i-table-widget` in the [payment settings](https://ilias.cern.ch/admin/payment/) page doesn't need a min/max column width and it works just fine. What is the reason behind this need?

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -505,7 +501,7 @@ th.i-table {
         display: block;
         position: relative;
         background-color: $light-gray;
-        border-bottom: 1px solid $gray;
+        @include border-bottom();
```
OmeGak: We usually follow this rule of thumb:

```scss
.foo {
    // @extend
    // @include
    // rules
}
```

nop33: I think is necessary ;) Just created a wiki page, similar to the JS one: https://github.com/indico/indico/wiki/SASS-Guidelines

pferreir:
Merged in f3f6a36.

PR #2634, opened 2016-10-07
Adapt single abstract PDF generation

nop33:


indico/modules/events/abstracts/controllers/management.py
```diff
@@ -57,6 +57,18 @@ def _process(self):
...
+class RHManageAbstractBase(AbstractMixin, RHManageAbstractsBase):
+    """Base class for all abstract management RHs"""
+
+    def _checkParams(self, params):
+        RHManageAbstractsBase._checkParams(self, params)
+        AbstractMixin._checkParams(self)
+
+    def _checkProtection(self):
+        RHManageAbstractsBase._checkProtection(self)
+        AbstractMixin._checkProtection(self)
```
ThiefMaster: This only checks for read access, not for management access. We need a `can_manage` method that checks whether the guy is an event manager or `can_review()` the abstract.

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -74,17 +86,17 @@ def _checkParams(self, params):
...
 
-    def _checkProtection(self):
-        RHManageAbstractsBase._checkProtection(self)
-        AbstractMixin._checkProtection(self)
 
+class RHAbstractExportPDF(RHManageAbstractBase):
     def _process(self):
-        return WPManageAbstracts.render_template('management/abstract.html', self._conf, abstract=self.abstract)
+        pdf = ConfManagerAbstractToPDF(self.abstract)
+        file_name = secure_filename('{}-Abstract.pdf'.format(self.abstract.title), 'abstract.pdf')
```
ThiefMaster: I'd just use `abstract-<friendly_id>.pdf` as the filename

indico/modules/events/abstracts/controllers/management.py
ThiefMaster: no, editing is not possible once an abstract has reached a final state for example

nop33: So basically, only the event managers can access the abstract management pages, which is satisfied by the `_checkProtection` of `RHModificationBaseProtected`, meaning that I'll have to remove the `AbstractMixin._checkProtection(self)` line.

OmeGak: It doesn't harm to leave it there in case we add some extra check in `AbstractMixin._checkProtection`.

indico/modules/events/abstracts/blueprint.py
```diff
@@ -70,3 +71,4 @@
 
 # Abstract-specific management
 _bp.add_url_rule('/manage/abstracts/<int:abstract_id>/', 'manage_abstract', RHManageAbstract)
+_bp.add_url_rule('/manage/abstracts/<int:abstract_id>/abstract.pdf', 'abstract_pdf_export', RHAbstractExportPDF)
```
OmeGak: If this is the PDF export for the managers I would make that obvious both in the RH and the endpoint name.

PR #2638, opened 2016-10-10
Display day of week in event header date

DavidAndreev:


nop33:
Can you also please add a screenshot so we don't have to check it out manually? Cheers

DavidAndreev:
header

DavidAndreev:
Or should I add coma after the day of week ?

DavidAndreev:
It looks the same, after the arrow it says the day of the week and then the date and time, just like starting date, sorry but I already rebased to devel so I cant take screenshot

indico/modules/events/templates/display/indico/_common.html
```diff
@@ -89,6 +89,7 @@
 
 {% macro render_datetime(prop_name, dt, timezone, format='medium') -%}
     <time itemprop="{{ prop_name }}" datetime="{{ dt.isoformat() }}">
+        {{ dt.strftime('%A') }}
```
ThiefMaster: I'd rather `| format_date(...)` this so we keep the correct timezone and locale. Also make sure to test it with another language setting to make sure you get the weekday in that language.

ThiefMaster: trans is used to translate static text; but if you use `format_date` it'll properly translate weekdays etc

DavidAndreev: dates
Im not sure what you mean, i get this when using format='full' , its good but that timezone name is really long

ThiefMaster: You can put custom format strings there: http://babel.pocoo.org/en/latest/dates.html#date-fields

DavidAndreev: Yeah that works and also gets translated, thanks

ThiefMaster:
Do we want to show this for multi-day events at all?

pferreir:
No, this is not needed for multi-day events. Those should have the weekday visible in the header of each day.

indico/modules/events/templates/display/indico/_common.html
```diff
@@ -87,8 +87,11 @@
     {% endfor %}
 {%- endmacro %}
 
-{% macro render_datetime(prop_name, dt, timezone, format='medium') -%}
+{% macro render_datetime(prop_name, dt, timezone, format='medium', show_week_day=false) -%}
```
ThiefMaster: `show_weekday`

PR #2639, opened 2016-10-10
Adapt abstracts CSV and XLSX export

nop33:


indico/modules/events/abstracts/controllers/management.py
```diff
@@ -279,3 +281,29 @@ def _process(self):
...
+        self.headers, self.rows = generate_spreadsheet_from_abstracts(self.abstracts,
+                                                                      self.export_config['static_item_ids'],
+                                                                      self.export_config['dynamic_items'])
+
+
+class RHAbstractsExportCSV(RHAbstractsExportBase):
+    """Export list of abstracts to CSV"""
+
+    def _process(self):
+        RHAbstractsExportBase._process(self)
```
ThiefMaster: Wouldn't it be cleaner to have a differently named methods for this which returns those headers and rows?

`return send_csv('abstracts.csv', *self._generate_spreadsheet())`

indico/modules/events/abstracts/util.py
```diff
@@ -227,6 +229,52 @@ def build_default_email_template(event, tpl_type):
...
+def generate_spreadsheet_from_abstracts(abstracts, static_item_ids, dynamic_items):
+    """Generates a spreadsheet data from a given abstract list.
+
+    :param abstracts: The list of abstracts to include in the file
+    :param static_item_ids: The abstract properties to be used as columns
+    :param dynamic_items: Contribution fields as extra columns
+    """
+    field_names = ['ID', 'Title']
+    static_item_mapping = OrderedDict([
+        ('state', ('State', lambda x: x.state.title)),
```
OmeGak: `attrgetter('state.title')`

indico/modules/events/abstracts/util.py
```diff
@@ -227,6 +229,52 @@ def build_default_email_template(event, tpl_type):
...
+                                   lambda x: x.accepted_contrib_type.name if x.accepted_contrib_type else None)),
+        ('submitted_contrib_type', ('Submitted type',
+                                    lambda x: x.submitted_contrib_type.name if x.submitted_contrib_type else None)),
+        ('score', ('Score', lambda x: round(x.score, 1))),
+        ('submitted_dt', ('Submission date', lambda x: to_unicode(format_datetime(x.submitted_dt)))),
+        ('modified_dt', ('Modification date', lambda x: (to_unicode(format_datetime(x.modified_dt)) if x.modified_dt
+                                                         else '')))
+    ])
+    for item in dynamic_items:
+        field_names.append(unique_col(item.title, item.id))
```
pferreir:
```python
field_names += [unique_col(item.title, item.id) for item in dynamic_items]
```
?

indico/modules/events/abstracts/util.py
nop33: Is there any benefit over using this instead of a lambda?

OmeGak: Performance, as `attrgetter` is implemented in C. Not that it matters much in this particular case.

nop33: I see. I'd like to keep it like this, makes the code more consistent in some sense. But not a strong argument.

PR #2641, opened 2016-10-12
Handle call for abstracts scheduling

nop33:


indico/modules/events/abstracts/forms.py
```diff
@@ -221,3 +224,18 @@ def __init__(self, *args, **kwargs):
...
         super(MultiTrackMixin, self).__init__(*args, **kwargs)
         self.submitted_for_tracks.query = Track.query.with_parent(event).order_by(Track.title)
+
+
+class AbstractsScheduleForm(IndicoForm):
+    start_dt = IndicoDateTimeField(_("Start"), default_time=time(0, 0),
+                                   description=_("The moment users can start submitting abstracts"))
+    end_dt = IndicoDateTimeField(_("End"), [LinkedDateTime('start_dt')], default_time=time(23, 59),
+                                 description=_("The moment the submission process will end"))
+    modification_end_dt = IndicoDateTimeField(_("Modification end"), [Optional(), LinkedDateTime('end_dt')],
```
pferreir: Maybe "Modification deadline"? "Modification" end doesn't sound that good...

indico/modules/events/abstracts/templates/management/abstracts_setup_actions/ended.html
```diff
@@ -0,0 +1,25 @@
+{% extends 'events/abstracts/management/abstracts_setup_actions/_base.html' %}
```
OmeGak: Maybe `.../cfa_actions/...` better than `.../abstracts_setup_actions/...`?

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -98,13 +100,68 @@ def _process(self):
         return send_file(file_name, pdf.generate(), 'application/pdf')
 
 
-class RHAbstracts(RHManageAbstractsBase):
+class RHAbstractsCallScheduleBase(RHManageAbstractsBase):
```
nop33: What looks better, `RHAbstractsCFASchedule` or `RHCFASchedule`? I would say the first.

indico/modules/events/abstracts/templates/management/abstracts_setup_actions/ended.html
nop33: Although I don't like using acronyms, I think your proposal is better since it represents exactly what the folder is used for.

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -98,13 +100,68 @@ def _process(self):
...
+    """Opens the call for abstracts"""
+
+    def _process(self):
+        if self.call_has_ended:
+            abstracts_settings.set_multi(self.event_new, {
+                'end_dt': None,
+                'modification_end_dt': None
+            })
+        else:
+            abstracts_settings.set(self.event_new, 'start_dt', now_utc())
```
ThiefMaster: I'd use `now_utc(False)`; we don't really need secs/usecs

indico/modules/events/abstracts/templates/management/abstracts_setup_actions/ended.html
```diff
@@ -0,0 +1,25 @@
...
+        Closed on <strong>{{ date }}</strong> at <strong>{{ time }}</strong> ({{ tz }})
+    {%- endtrans %}
+{% endblock %}
+
+{% block buttons %}
+    <div class="group">
+        <a href="#" class="i-button"
+           data-href="{{ url_for('.open_abstracts_call', event) }}"
+           data-method="POST"
+           data-confirm="{% trans %}The call for abstracts is finished already. Do you want to open it again?{% endtrans %}">
```
ThiefMaster: `has already finished`?

indico/modules/events/abstracts/controllers/management.py
ThiefMaster: I think we have more RH with "RH do xyz"-like names than "RH xyz do"

indico/modules/events/abstracts/forms.py
```diff
@@ -221,3 +224,18 @@ def __init__(self, *args, **kwargs):
...
+                                              default_time=time(23, 59),
+                                              description=_("The submitted abstracts cannot be modified after this "
```
ThiefMaster: Yes, it's technically the same since the date has to be after the `end_dt` (we do check that, right?). But the idea behind the feature is not so much to disallow editing after date X but rather to allow editing until date X even though the submission period already ended at date Y (<X)

nop33: Yes, we do check it. Do you think a message like *"Deadline until which the submitted abstracts can be modified"* expresses it better?

indico/modules/events/abstracts/templates/management/abstracts_setup_actions/ended.html
nop33: Don't know which one is more correct. In the registration form we use `is`. With the `has` I would expect a `been` as well.

indico/modules/events/abstracts/operations.py
```diff
@@ -62,3 +62,29 @@ def delete_abstract(abstract, delete_contrib=False):
...
     abstract.event_new.log(EventLogRealm.management, EventLogKind.negative, 'Abstracts',
                            'Abstract "{}" has been deleted'.format(abstract.title), session.user)
+
+
+def schedule_cfa(event, data):
+    event.cfa.schedule(data)
+    logger.info("Call for abstracts for %s scheduled by %s", event, session.user)
+    date_format = '%d/%m/%Y %H:%M'
+    log_data = {
+        'Start': data['start_dt'].strftime(date_format),
```
OmeGak: You probably want to use `format_datetime(..., format='code')` instead of `strftime`.

ThiefMaster: While you are right that it should not be done with strftime, this is data shown to humans so I don't think "code" format is the right choice. Maybe check if we put a date in log data anywhere else and see which format we use there?

Or maybe we could just use `.isoformat()`. That's human-readable but more important we could easily add some extra code in the future when displaying log data that checks with a whether it's an isoformat date string and then formats it in a pretty way.

nop33: We use `format_date`:

* https://github.com/indico/indico/blob/master/indico/modules/events/timetable/operations.py#L85
* https://github.com/indico/indico/blob/master/indico/modules/events/timetable/operations.py#L111
* https://github.com/indico/indico/blob/master/indico/modules/events/timetable/operations.py#L123
* https://github.com/indico/indico/blob/master/indico/modules/events/timetable/reschedule.py#L97

Should I use `format_date` too, or just `.isoformat()`?

ThiefMaster:
Just isoformat() is fine

ThiefMaster:
Second one is awful to reformat while isoformat could be easily formatted 
nicely when displaying (few lines of code, could be something for one of my 
officemates)

indico/modules/events/abstracts/blueprint.py
```diff
@@ -40,6 +41,9 @@
 
 # Management
 _bp.add_url_rule('/manage/abstracts/', 'manage_abstracts', RHAbstracts)
+_bp.add_url_rule('/manage/abstracts/schedule', 'schedule_abstracts_call', RHScheduleCFA, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/abstracts/open', 'open_abstracts_call', RHOpenCFA, methods=('GET', 'POST'))
```
ThiefMaster: :warning: should be POST-only

indico/modules/events/abstracts/models/call_for_abstracts.py
```diff
@@ -0,0 +1,64 @@
...
+from indico.modules.events.abstracts.settings import abstracts_settings
+from indico.util.date_time import now_utc
+
+
+class CallForAbstracts(object):
+    """Proxy class to facilitate access to the call for abstracts settings"""
+
+    def __init__(self, event):
+        self.event = event
+        self.settings = abstracts_settings.get_all(self.event)
```
ThiefMaster: There's lots of request-level caching on settings, so you could just `get()` them when needed instead of storing the dict here.

indico/modules/events/abstracts/models/call_for_abstracts.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @property
+    def has_started(self):
+        start_dt = self.settings['start_dt']
+        return start_dt is not None and start_dt <= now_utc()
+
+    @property
+    def has_ended(self):
+        call_end_dt = self.settings['end_dt']
```
ThiefMaster: `end_dt`

indico/modules/events/abstracts/models/call_for_abstracts.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @property
+    def start_dt(self):
+        return self.settings['start_dt']
+
+    @property
+    def end_dt(self):
+        return self.settings['end_dt']
+
+    def schedule(self, data):
```
ThiefMaster: I'd take explicit args here instead of `**kwargs`.

indico/modules/events/abstracts/models/call_for_abstracts.py
```diff
@@ -0,0 +1,64 @@
...
+        if self.has_ended:
+            abstracts_settings.set_multi(self.event, {
+                'end_dt': None,
+                'modification_end_dt': None
+            })
+        else:
+            abstracts_settings.set(self.event, 'start_dt', now_utc(False))
+
+    def close(self):
+        current_ts = now_utc(False)
```
ThiefMaster: I think usually we call this `now`

indico/modules/events/abstracts/templates/management/cfa_actions/ended.html
```diff
@@ -0,0 +1,25 @@
...
+        Closed on <strong>{{ date }}</strong> at <strong>{{ time }}</strong> ({{ tz }})
+    {%- endtrans %}
+{% endblock %}
+
+{% block buttons %}
+    <div class="group">
+        <a href="#" class="i-button"
+           data-href="{{ url_for('.open_abstracts_call', event) }}"
+           data-method="POST"
+           data-confirm="{% trans %}The call for abstracts is finished already. Do you want to open it again?{% endtrans %}">
```
ThiefMaster: "is already finished"? "is finished already" sounds wrong to me (but better ask a native speaker)

indico/modules/events/abstracts/operations.py
```diff
@@ -62,3 +62,28 @@ def delete_abstract(abstract, delete_contrib=False):
     logger.info('Abstract %s deleted by %s', abstract, session.user)
     abstract.event_new.log(EventLogRealm.management, EventLogKind.negative, 'Abstracts',
                            'Abstract "{}" has been deleted'.format(abstract.title), session.user)
+
+
+def schedule_cfa(event, data):
```
OmeGak: I would still leave `start_dt, end_dt, modification_end_dt` instead of `data`. Otherwise it's quite cryptic to use from the shell.

PR #2642, opened 2016-10-12
Add "Roles" option in CFA setup

pferreir:


indico/htdocs/js/indico/jquery/trackrolewidget.js
```diff
@@ -0,0 +1,94 @@
...
+            var self = this;
+            $trackConfig.find('.js-role-container').each(function(__, container) {
+                self._setUpRoleContainer($(container));
+            });
+        },
+
+        _create: function() {
+            var self = this;
+            var $container = this.options.containerElement;
+            this.roleData = JSON.parse(this.element.val());
```
nop33: Use `self` instead of `this` since you assigned it in the first row.

indico/htdocs/js/indico/jquery/trackrolewidget.js
```diff
@@ -0,0 +1,94 @@
...
+
+                onAdd: function(people) {
+                    people.forEach(function(person) {
+                        roleData.push(person.id);
+                        self._update();
+                    });
+                }
+            });
+
+            $container.on('click', '.remove-user', function(e) {
```
pferreir: Really? When? Where?

nop33: Well, it was @ThiefMaster who told me I should use `evt` and if you search in our codebase for `(evt)` you'll find quite many occurrences!

indico/htdocs/js/indico/jquery/trackrolewidget.js
nop33: What's the point of assigning `self` then?!

indico/htdocs/js/indico/jquery/trackrolewidget.js
pferreir: Well, we should document all this stuff. "Person said" is not good enough.

indico/htdocs/js/indico/jquery/trackrolewidget.js
pferreir: How do you refer to `this` inside a closure otherwise?

indico/htdocs/js/indico/jquery/trackrolewidget.js
pferreir: OK, what about this? https://github.com/indico/indico/wiki/JavaScript-Guidelines

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -279,3 +282,46 @@ def _process(self):
...
+                        track.conveners = users
+                        conveners |= users
+                    if role_id == 'reviewer':
+                        track.abstract_reviewers = users
+                        reviewers |= users
+
+            # Update actual ACLs
+            update_object_principals(self.event_new, conveners, role="track_convener")
+            update_object_principals(self.event_new, reviewers, role="abstract_reviewer")
+            flash(_("Abstract reviewing roles have been updated."), 'success')
```
OmeGak: What about some logging?

indico/modules/events/abstracts/fields.py
```diff
@@ -30,13 +29,28 @@
...
 from indico.modules.events.fields import PersonLinkListFieldBase
+from indico.modules.users.models.users import User
 from indico.util.decorators import classproperty
 from indico.util.i18n import _
 from indico.web.forms.base import AjaxFieldMixin
 from indico.web.forms.fields import MultipleItemsField, JSONField
 from indico.web.forms.widgets import JinjaWidget, SelectizeWidget
 
 
+def serialize_user(user):
```
OmeGak: I would move this inside `TrackRoleField` since it's only used there.

indico/htdocs/sass/modules/_users.scss
```diff
@@ -70,6 +70,28 @@
     }
 }
 
+.event-user {
+    @include border-radius();
+    @include border-all($pastel-gray);
```
pferreir: Ah yeah. That's a copy-paste, but I'll fix that. Thanks.

indico/htdocs/js/indico/jquery/trackrolewidget.js
```diff
@@ -0,0 +1,94 @@
+/* This file is part of Indico.
```
ThiefMaster: `track_role_widget.js`? Those filenames without any separators are awful, especially now that many editors have smart search where `trwi` would find `track_role_widget` but not `trackrolewidget`...

I know it's consistent with the ones we already have but I'd also consider renaming these at some point...

indico/htdocs/js/indico/jquery/trackrolewidget.js
```diff
@@ -0,0 +1,94 @@
...
+                self._setUpRoleContainer($(container));
+            });
+        },
+
+        _create: function() {
+            var self = this;
+            var $container = self.options.containerElement;
+            self.roleData = JSON.parse(self.element.val());
+            $container.find('.js-track-config').each(function(__, elem) {
+                self._setUp($(elem));
```
ThiefMaster: `$(this)` and you don't need any arguments in the callback

indico/htdocs/sass/modules/abstracts/roles.scss
```diff
@@ -0,0 +1,85 @@
...
+ * Indico is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+ @import 'base';
```
ThiefMaster: leading whitespace

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -279,3 +282,46 @@ def _process(self):
...
+
+    def _process(self):
+        roles = get_roles_for_event(self.event_new)
+        form = AbstractReviewingRolesForm(event=self.event_new, obj=FormDefaults(roles=roles))
+        conveners = set()
+        reviewers = set()
+
+        if form.validate_on_submit():
+            # Handle global reviewers/conveners
+            if '' in form.roles.data:
```
ThiefMaster: Wouldn't a key like `*` be cleaner than an empty string?

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -279,3 +282,46 @@ def _process(self):
...
+                role_data = form.roles.data.pop('')
+                self.event_new.global_conveners = set(User.find(User.id.in_(role_data['convener'])))
```
ThiefMaster: `is_deleted=False`

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -279,3 +282,46 @@ def _process(self):
...
+                self.event_new.abstract_global_reviewers = set(User.find(User.id.in_(role_data['reviewer'])))
+                conveners |= self.event_new.global_conveners
+                reviewers |= self.event_new.abstract_global_reviewers
+
+            track_dict = {track.id: track for track in Track.find(Track.id.in_(form.roles.data))}
```
ThiefMaster: Why not query only tracks for the event? Then you don't need the check whether it's from the same event below...

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -279,3 +282,46 @@ def _process(self):
...
+            user_dict = dict(get_users_in_roles(form.roles.data))
+            # Update track-specific reviewers/conveners
+            for track_id, roles in form.roles.data.viewitems():
+                track = track_dict[int(track_id)]
+                if track.event_new != self.event_new:
+                    raise Forbidden('Cannot manage track {} from event {}'.format(track, self.event_new))
```
ThiefMaster: I have the feeling most of the logic in here actually belongs in the Field, not the Form or RH.

The field could return a tuple with the two reviewer lists and the necessary ACL role updates and then we'd just set/apply them here.

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -279,3 +282,46 @@ def _process(self):
...
+                    users = {user_dict[user_id] for user_id in user_ids}
+                    if role_id == 'convener':
+                        track.conveners = users
+                        conveners |= users
+                    if role_id == 'reviewer':
+                        track.abstract_reviewers = users
+                        reviewers |= users
+
+            # Update actual ACLs
+            update_object_principals(self.event_new, conveners, role="track_convener")
```
ThiefMaster: `'`

indico/modules/events/abstracts/fields.py
ThiefMaster: Or at least underscore it.

indico/modules/events/abstracts/fields.py
```diff
@@ -30,13 +29,28 @@
...
+    return {
+        'id': user.id,
+        'name': user.name
+    }
+
+
+def get_users_in_roles(data):
+    user_ids = {user_id for user_roles in data.viewvalues()
```
ThiefMaster: I'd wrap before `for` to have each loop on its own line. At least I usually try to wrap everything as soon as it doesn't fit in a single line and IMO it's more readable. Especially with the ridiculous order Python uses for multiple-for comprehensions I think it's not a good idea to have a somewhat unrelated loop and the `user_id` in the same line.

indico/modules/events/models/events.py
```diff
@@ -338,6 +338,32 @@ def __table_args__(cls):
         )
     )
 
+    #: Users who can review on all tracks
+    abstract_global_reviewers = db.relationship(
```
ThiefMaster: `global_abstract_reviwers`?

migrations/versions/201610111427_44ef3c219621_allow_event_wide_track_conveners_.py
```diff
@@ -0,0 +1,47 @@
+"""Allow event-wide track conveners/reviewers
```
ThiefMaster: I think it'd be better to merge the changes in the revision where those tables are initially created.

indico/modules/events/abstracts/controllers/management.py
pferreir: Good idea.

indico/htdocs/js/indico/jquery/trackrolewidget.js
OmeGak: Now that you mention it. This should actually go into `htdocs/js/indico/jquery/widgets/`.

indico/modules/events/abstracts/fields.py
pferreir: I'm pretty much against static methods in Python (with very few exceptions). I'd underscore it.

indico/htdocs/js/indico/jquery/trackrolewidget.js
pferreir: > Now that you mention it. This should actually go into `htdocs/js/indico/jquery/widgets/`.

Should it? I considered that option, but this is a self-contained jQueryUI widget, not supporting code for a widget (event if it ends up working as such).

OmeGak: I've been throwing stuff in `.../indico/widget` when the code was strictly related to a WTForm widget. Even if it's a self-contained jQueryUI widget, it's still quite specific and strictly related to the field (expecting certain HTML structure and data available in `data-` attributes), so to me, it feels like it belongs to `.../indico/widget`.

indico/modules/events/abstracts/controllers/management.py
pferreir: Done. Please have a look.

migrations/versions/201610111427_44ef3c219621_allow_event_wide_track_conveners_.py
pferreir: Yeah, just wanted to have it reviewable.

ThiefMaster: dbdiff is better at reviewing alembic revision changes than any human ;)

indico/htdocs/js/indico/jquery/trackrolewidget.js
ThiefMaster: wtforms widgets are in `htdocs/js/indico/widgets/` but so far we don't have anything else in there (i.e. no jquery ui widgets) so I think this one does belong into `indico/htdocs/js/indico/jquery/`

indico/htdocs/sass/modules/abstracts/roles.scss
```diff
@@ -0,0 +1,85 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+@import 'base';
+
+#form-group-roles .form-field {
```
ThiefMaster: could we add a custom class in the field's html instead of hardcoding an ID here?

indico/modules/events/models/events.py
```diff
@@ -338,6 +338,32 @@ def __table_args__(cls):
...
+    )
+
+    #: Users who are conveners on all tracks
+    global_conveners = db.relationship(
+        'User',
+        secondary='events.track_conveners',
+        collection_class=set,
+        lazy=True,
+        backref=db.backref(
+            'global_convener_for_events',
```
ThiefMaster: conveners

indico/modules/events/tracks/models/abstract_reviewers.py
```diff
@@ -23,20 +23,31 @@
     'track_abstract_reviewers',
     db.metadata,
     db.Column(
+        'id',
+        db.Integer,
+        primary_key=True,
+        autoincrement=True,
```
ThiefMaster: that's implied for "simple" single-column PKs; i was only setting it to false because the behavior for composite PKs is broken/unexpected in sqlalchemy

indico/htdocs/sass/modules/abstracts/roles.scss
pferreir: Why? There's a single one of those in the whole application.

indico/modules/events/models/events.py
pferreir: What?

indico/htdocs/sass/modules/abstracts/roles.scss
ThiefMaster: just don't like the idea very much to have css referencing a specific *name* of a form field. this would break e.g. when specifying a `prefix` when instantiating the form containing it

PR #2644, opened 2016-10-13
Fix rules layout and improve UX

nop33:
### Fix layout

The layout was broken when any of the `<select>` fiends inside a rule had a very long `<option>`.

### Improve UX

The **description** and the **position** of the _Stop on match_ field is very misleading. The text _"Do not evaluate any other rules once this one matches"_ feels wrong because what the user will think is that it refers to the rules of this email template (since it is positioned inside the **Rules** fieldset) and not to the rules of the following email templates.

I had to check the code to understand the logic behind _Stop on match_ and a user won't be able to do that :wink: 

This is my (easy-fix) solution:

### Future improvements

The best solution imo would be to add a button in the list of email templates which will trigger an AJAX request. Here's a quick mockup:

### DIFF
--- a/indico/modules/events/abstracts/forms.py
+++ b/indico/modules/events/abstracts/forms.py
@@ -130,7 +130,8 @@ class EditEmailTemplateRuleForm(IndicoForm):
     title = StringField(_("Title"), [DataRequired()])
     rules = EmailRuleListField(_("Rules"), [DataRequired()])
     stop_on_match = BooleanField(_("Stop on match"), [DataRequired()], widget=SwitchWidget(), default=True,
-                                 description=_("Do not evaluate any other rules once this one matches."))
+                                 description=_("If any of the rules from this email template matches, do not "
+                                               "send emails from any following email template."))
 
     def __init__(self, *args, **kwargs):
         self.event = kwargs.pop('event')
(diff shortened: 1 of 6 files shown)


indico/modules/events/abstracts/forms.py
```diff
@@ -148,7 +149,7 @@ class EditEmailTemplateTextForm(IndicoForm):
 
     reply_to_address = SelectField(_('"Reply to" address'), [DataRequired()])
     include_submitter = BooleanField(_('Send to submitter'), widget=SwitchWidget())
-    include_authors = BooleanField(_('Send to primary authorized_submittershors'), widget=SwitchWidget())
```
ThiefMaster: :monkey:

indico/modules/events/abstracts/templates/management/notification_tpl_form.html
```diff
@@ -2,9 +2,9 @@
 
 {{ form_header(form) }}
     {{ form_row(form.title) }}
+    {{ form_row(form.stop_on_match) }}
```
ThiefMaster: `{{ form_rows(form, fields=('title', 'stop_on_match')) }}`

indico/modules/events/abstracts/forms.py
nop33: Some review process before merging didn't go as expected here, ha? :stuck_out_tongue:

nop33:
@mvidalgarcia will create a new PR based on this one. Then, this PR will be closed.

ThiefMaster:
@mvidalgarcia your PR fully supersedes this one, right? if it's the case let's close this one

mvidalgarcia:
Yes

OmeGak:
Superseded by #2654

PR #2646, opened 2016-10-14
List of tracks

DavidAndreev:


indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -519,3 +519,43 @@ th.i-table {
...
+    }
+
+    .i-box-content {
+        border-top: 1px solid $gray;
+        padding: 20px 10px 10px 10px;
+        margin-left: -10px;
+        margin-right: -10px;
+    }
+
+    .gradient{
```
ThiefMaster: missing space

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -519,3 +519,43 @@ th.i-table {
...
     }
 }
+
+#track-list {
+    list-style-type: none;
+    padding-left: 0;
+    width: 800px;
+
+    .i-box {
+        padding: 10px 10px 0 10px ;
```
ThiefMaster: too much space

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -519,3 +519,43 @@ th.i-table {
...
+        margin-bottom: 10px;
+    }
+
+    .i-box-header{
```
ThiefMaster: missing space

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -519,3 +519,43 @@ th.i-table {
...
+    .i-box-content {
+        border-top: 1px solid $gray;
+        padding: 20px 10px 10px 10px;
+        margin-left: -10px;
+        margin-right: -10px;
+    }
+
+    .gradient{
+        position: relative;
+            &:before {
```
ThiefMaster: too much indentation - also, put this at the end of the block instead of having it between attributes

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+            cursor: 'move',
+            distance: 2,
+            handle: '.ui-i-box-sortable-handle',
+            items: '> li.track-row',
+            tolerance: 'pointer',
+            forcePlaceholderSize: true,
+            update: function(e) {
+                var $elem = $('#track-list');
+                var sortedList = $elem.find('li.track-row').map(function(i, elem) {
+                    return $(elem).data('id');
```
ThiefMaster: `$(this)`

indico/modules/events/tracks/blueprint.py
```diff
@@ -16,13 +16,17 @@
...
 _bp = IndicoBlueprint('tracks', __name__, template_folder='templates', virtual_template_folder='events/tracks',
-                      url_prefix='/event/<confId>')
+                      url_prefix='/event/<confId>/manage')
 
-_bp.add_url_rule('/manage/tracks/', 'manage', RHManageTracks)
-_bp.add_url_rule('/manage/tracks/create', 'create', RHCreateTrack, methods=('GET', 'POST'))
+_bp.add_url_rule('/tracks/', 'manage', RHManageTracks)
+_bp.add_url_rule('/tracks/create', 'create', RHCreateTrack, methods=('GET', 'POST'))
+_bp.add_url_rule('/tracks/<int:id>/edit', 'edit_track', RHEditTrack, methods=('GET', 'POST'))
+_bp.add_url_rule('/tracks/<int:id>/delete', 'delete_track', RHDeleteTrack, methods=('DELETE', ))
```
ThiefMaster: don't use the DELETE method if you have a `/delete` in the URL. either use /delete or use DELETE on `/tracks/<int:id>/`

indico/modules/events/tracks/blueprint.py
```diff
@@ -16,13 +16,17 @@
...
 
 _bp = IndicoBlueprint('tracks', __name__, template_folder='templates', virtual_template_folder='events/tracks',
-                      url_prefix='/event/<confId>')
+                      url_prefix='/event/<confId>/manage')
 
-_bp.add_url_rule('/manage/tracks/', 'manage', RHManageTracks)
-_bp.add_url_rule('/manage/tracks/create', 'create', RHCreateTrack, methods=('GET', 'POST'))
+_bp.add_url_rule('/tracks/', 'manage', RHManageTracks)
+_bp.add_url_rule('/tracks/create', 'create', RHCreateTrack, methods=('GET', 'POST'))
+_bp.add_url_rule('/tracks/<int:id>/edit', 'edit_track', RHEditTrack, methods=('GET', 'POST'))
```
ThiefMaster: could be just `/tracks/<int:track_id>` without `/edit` unless we add more stuff "within" `/tracks/<int:track_id>/...` later

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +68,39 @@ def _process(self):
                                 tracks=[{'id': int(t.getId()), 'title': to_unicode(t.getTitle())}
                                         for t in self._conf.getTrackList()])
         return jsonify_form(form)
+
+
+class RHEditTrack(RHManageTracksBase):
+
```
ThiefMaster: you only put an empty line here if there's a docstring

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +68,39 @@ def _process(self):
...
+    def _checkParams(self, params):
+        RHConferenceModifBase._checkParams(self, params)
+        self.track = Track.find_one(id=request.view_args['id'])
```
ThiefMaster: Youdo this in two RHs -> add a `RHManageTrackBase` that contains this method and inherit from that one for edit/delete

ThiefMaster: :warning: Security issue - you can put any track ID from another event here. Set an URL normalization rule on the RH to prevent this!

```python
    normalize_url_spec = {
        'locators': {
            lambda self: self.track
        }
}
```

That way any request to a track not matching its locator data (which should contain track id and event id) will automatically redirect to the proper location where the access check is performed for the correct event.

indico/modules/events/tracks/templates/management.html
```diff
@@ -1,20 +1,25 @@
 {% extends 'layout/full_content_base.html' %}
```
OmeGak: Just extend from `base.html`.

```jinja
{% extends 'layout/base.html' %}
```

This will make sure the page is page content is limited to `800px` and you can remove that CSS rule from `#track-list`.

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +68,39 @@ def _process(self):
...
+
+    def _process(self):
+        form = TrackForm(obj=FormDefaults(self.track))
```
ThiefMaster: `obj=self.track` - no need to wrap it in FormDefaults if you just pass-through to an object without any of the extra FormDefaults functionality

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -519,3 +519,43 @@ th.i-table {
         }
     }
 }
+
+#track-list {
+    list-style-type: none;
+    padding-left: 0;
+    width: 800px;
```
OmeGak: You can remove this rule if you extend from the correct base html template (comment down below).

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -519,3 +519,43 @@ th.i-table {
...
+
+    .i-box {
+        padding: 10px 10px 0 10px ;
+        margin-bottom: 10px;
```
OmeGak: You don't need to specify `margin-bottom` here. Simply wrap the `.i-box` elements in a `div.i-box-group.vert` and that will take care of spacing them conveniently.

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +68,39 @@ def _process(self):
...
+        if form.validate_on_submit():
+            update_track(self.track, form.data)
+            return jsonify_data(flash=False)
+        return jsonify_form(form)
+
+
+class RHSortTracks(RHManageTracksBase):
+
+    def _process(self):
+        sort_order = request.json['sort_order']
```
ThiefMaster: just fyi, you could also use  `request.form.getlist('sort_order')` like we do in [many other](https://github.com/indico/indico/blob/master/indico/modules/events/contributions/controllers/management.py#L382-L388) sorting RHs

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -519,3 +519,43 @@ th.i-table {
         }
     }
 }
+
+#track-list {
+    list-style-type: none;
+    padding-left: 0;
+    width: 800px;
+
+    .i-box {
```
OmeGak: Out of curiosity. What's the need for customizing the `.i-box`, `.i-box-header`, and `.i-box-content`?

indico/modules/events/tracks/models/tracks.py
```diff
@@ -97,6 +97,11 @@ class Track(DescriptionMixin, db.Model):
     # - contributions (Contribution.track)
     # - proposed_abstract_reviews (AbstractReview.proposed_tracks)
 
+    @property
+    def locator(self):
+        return {'confId': self.event_id,
```
ThiefMaster: @locator_property
    def locator(self):
        return dict(self.event_new.locator, track_id=self.id)

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -519,3 +519,43 @@ th.i-table {
...
+        margin-bottom: 10px;
+    }
+
+    .i-box-header{
+        border-bottom: None;
+        margin-bottom: 0;
+    }
+
+    .i-box-content {
+        border-top: 1px solid $gray;
```
OmeGak: You can do this by using the SASS mixin `@include border-top();`.

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -519,3 +519,43 @@ th.i-table {
...
+    .gradient{
+        position: relative;
+            &:before {
+                content: '';
+                width: 100%;
+                height: 100%;
+                left: 0;
+                top: 20px;
+                position: absolute;
+                background: linear-gradient(transparent 0.4em, white 4.2em)
```
OmeGak:
```sass
@include gradient(salmon, tomato);
```

indico/modules/events/tracks/operations.py
```diff
@@ -31,4 +32,21 @@ def create_track(event, data):
...
+    logger.info('Track %r modified by %s', track, session.user)
+    track.event_new.log(EventLogRealm.management, EventLogKind.change, 'Tracks',
+                        'Track "{}" has been modified.'.format(track.title), session.user)
+    flash(_('Track "{}" has been modified.').format(track.title), 'success')
+    return track
+
+
+def delete_track(track):
+    db.session.delete(track)
+    logger.info('Track deleted by %s: %s', session.user, track)
```
ThiefMaster: `%r` for both (makes no difference but it's cleaner since we want the repr anyway)

indico/modules/events/tracks/operations.py
```diff
@@ -31,4 +32,21 @@ def create_track(event, data):
...
     event.log(EventLogRealm.management, EventLogKind.positive, 'Tracks',
               'Track "{}" has been created.'.format(track.title), session.user)
+    flash(_('Track "{}" has been created.').format(track.title), 'success')
     return track
+
+
+def update_track(track, data):
+    track.populate_from_dict(data)
+    db.session.flush()
+    logger.info('Track %r modified by %s', track, session.user)
```
ThiefMaster: `%r` for both (makes no difference but it's cleaner since we want the repr anyway)

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
+{% from 'message_box.html' import message_box %}
+
+{% macro render_track_list(event) %}
+    {% if not event.tracks %}
+        {% call message_box('info', classes='no-rules-placeholder') %}
+            {% trans %}No tracks have been added yet.{% endtrans%}
+        {% endcall %}
+    {% else %}
+        <ul id="track-list" data-url="{{ url_for('.sort_tracks', event) }}">
+            {% for track in event.tracks|sort(attribute='position') %}
```
ThiefMaster: I'm quite sure `event.tracks` is already sorted on the relationship level

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+                <li class="track-row i-box" data-id="{{ track.id }}">
+                    <div class="ui-i-box-sortable-handle"></div>
+                    <div class="i-box-header">
+                        <span class="i-box-title">{{ track.title }}</span>
+                        <div class="i-box-buttons toolbar right">
+                            <div class="group">
+                                <a href="#" class="i-button icon-edit" title="{% trans %}Edit track{% endtrans %}"
+                                   data-href="{{ url_for('tracks.edit_track', track) }}"
+                                   data-title="{% trans title=track.title %}Edit {{ title }}{% endtrans %}"
```
ThiefMaster: quote the title, it looks really weird otherwise

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+                            {{ track.description }}
+                        </div>
+                    {% endif %}
+                </li>
+            {% endfor %}
+        </ul>
+    {% endif %}
+
+    <script>
+        var heightLimit = 50;
```
ThiefMaster: Where does this number come from? Also you are leaking a global. Wrap the whole thing in an IIFE like we do in other inline scripts that define variables/functions. Or just move it to a JS file and just call the function containing it here.

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+        $(document).ready(function() {
```
ThiefMaster: this is most likely not needed if everything you care about is already defined in the DOM

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+            $('.i-box-content').each(function() {
+            var innerHeight =$(this).innerHeight()-$(this).css('paddingTop').replace('px','')- $(this).css('paddingBottom').replace('px','');
```
ThiefMaster: that whole JS block needs some reformatting :/

also, I think this line is rather awful. There must be a cleaner way to do this, no?

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +68,39 @@ def _process(self):
...
+        for position, track_id in enumerate(sort_order, 1):
+            if track_id in tracks:
+                tracks[track_id].position = position
+
+
+class RHDeleteTrack(RHManageTracksBase):
+
+    def _checkParams(self, params):
+        RHConferenceModifBase._checkParams(self, params)
+        self.track = Track.find_one(id=request.view_args['id'])
```
OmeGak: You get the track in two different request handlers. What about adding a `RHManageTrackBase` that does this for you automatically?

```python
class RHManageTrackBase(RHManageTracksBase):
    def _checkParams(self, params):
        RHManageTracksBase._checkParams(self, params)
        self.track = Track.find_one(id=request.view_args['id'])


class RHEditTrack(RHManageTrackBase):
    def _process(self):
        # ...


class RHDeleteTrack(RHManageTrackBase):
    def _process(self):
        # ...
```

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+             if( innerHeight > heightLimit){
+                 $(this).addClass('gradient');
+                 $(this).click(function() {
```
ThiefMaster: We usually use `.on('click', ...)` and assign `var $this = $(this);` if it's used more than once, or you could just chain it with the previous line

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +68,39 @@ def _process(self):
                                 tracks=[{'id': int(t.getId()), 'title': to_unicode(t.getTitle())}
                                         for t in self._conf.getTrackList()])
         return jsonify_form(form)
+
+
+class RHEditTrack(RHManageTracksBase):
+
+    def _checkParams(self, params):
+        RHConferenceModifBase._checkParams(self, params)
```
OmeGak: Probably left out from copy&paste. It should be `RHManageTracksBase._checkParams(self, params)`.

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+                     $(this).parent().find('.icon-edit').click();
+                 });
+             }
+            });
+
+            $('.i-box').last().css('margin-bottom', '150px');
```
ThiefMaster: more hardcoded sizes which can probably be solved in a cleaner way

indico/modules/events/tracks/templates/management.html
```diff
@@ -1,20 +1,25 @@
...
 {% endblock %}
 
-{% block content %}
-    {% for track in tracks %}
-        <div>
-            {{ track.title }}
-        </div>
-    {% endfor %}
+{% block description %}
+    {%- trans %} Tracks form the programme of the event. Here you can find a list of tracks for this event.
```
ThiefMaster: linebreak after `{%- trans %}`, also use `{%- trans -%}` to avoid surrounding whitespace in the translation string (same at the beginning of `endtrans`)

indico/modules/events/tracks/operations.py
```diff
@@ -31,4 +32,21 @@ def create_track(event, data):
     logger.info('Track %r created by %s', track, session.user)
     event.log(EventLogRealm.management, EventLogKind.positive, 'Tracks',
               'Track "{}" has been created.'.format(track.title), session.user)
+    flash(_('Track "{}" has been created.').format(track.title), 'success')
```
OmeGak: `flash` belongs in the request handler, as it is strictly related with the rendering of the page. Functions in `operations.py` could be called from `indico shell` and flashing makes no sense there.

indico/modules/events/tracks/operations.py
```diff
@@ -31,4 +32,21 @@ def create_track(event, data):
...
     return track
+
+
+def update_track(track, data):
+    track.populate_from_dict(data)
+    db.session.flush()
+    logger.info('Track %r modified by %s', track, session.user)
+    track.event_new.log(EventLogRealm.management, EventLogKind.change, 'Tracks',
+                        'Track "{}" has been modified.'.format(track.title), session.user)
+    flash(_('Track "{}" has been modified.').format(track.title), 'success')
```
OmeGak: Same here. Just move it to the RH.

indico/modules/events/tracks/operations.py
```diff
@@ -31,4 +32,21 @@ def create_track(event, data):
...
+    return track
+
+
+def delete_track(track):
+    db.session.delete(track)
+    logger.info('Track deleted by %s: %s', session.user, track)
+    flash(_('Track "{}" has been deleted.').format(track.title), 'success')
```
OmeGak: And here.

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+    {% else %}
+        <ul id="track-list" data-url="{{ url_for('.sort_tracks', event) }}">
+            {% for track in event.tracks|sort(attribute='position') %}
+                <li class="track-row i-box" data-id="{{ track.id }}">
+                    <div class="ui-i-box-sortable-handle"></div>
+                    <div class="i-box-header">
+                        <span class="i-box-title">{{ track.title }}</span>
+                        <div class="i-box-buttons toolbar right">
+                            <div class="group">
+                                <a href="#" class="i-button icon-edit" title="{% trans %}Edit track{% endtrans %}"
```
OmeGak: Maybe we can follow @nop33's suggestion and keep one attribute per line for improved readability.

indico/modules/events/tracks/templates/_track_list.html
OmeGak: `Edit track "{{ title }}"` perhaps?

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+                                   data-href="{{ url_for('tracks.edit_track', track) }}"
+                                   data-title="{% trans title=track.title %}Edit {{ title }}{% endtrans %}"
+                                   data-reload-after
+                                   data-ajax-dialog></a>
+                                <a href="#" class="i-button icon-remove js-delete-item"
+                                   title="{% trans %}Delete track{% endtrans %}"
+                                   data-href="{{ url_for('tracks.delete_track', track) }}"
+                                   data-confirm="{% trans %}Are you sure you want to delete this track?{% endtrans %}"
```
OmeGak: We probably want to display the title of the track the user is about to delete. I would add `track "{{ title }}"` here.

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+                    <div class="i-box-header">
+                        <span class="i-box-title">{{ track.title }}</span>
+                        <div class="i-box-buttons toolbar right">
+                            <div class="group">
+                                <a href="#" class="i-button icon-edit" title="{% trans %}Edit track{% endtrans %}"
+                                   data-href="{{ url_for('tracks.edit_track', track) }}"
+                                   data-title="{% trans title=track.title %}Edit {{ title }}{% endtrans %}"
+                                   data-reload-after
+                                   data-ajax-dialog></a>
+                                <a href="#" class="i-button icon-remove js-delete-item"
```
OmeGak: `i-button danger text-color ...`?

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+                        <div class="i-box-content">
+                            {{ track.description }}
+                        </div>
+                    {% endif %}
+                </li>
+            {% endfor %}
+        </ul>
+    {% endif %}
+
+    <script>
```
OmeGak: We probably want to move this to a `tracks.js` file. I can help you adding it and making sure it's served on the pages that it's needed. Just let me know and I'll pass by your office.

indico/modules/events/tracks/templates/_track_list.html
OmeGak: This happens unconditionally. Why can't it be done directly in CSS? Also, why do we need to add `margin-bottom`?

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+        var heightLimit = 50;
+        $(document).ready(function() {
+            $('.i-box-content').each(function() {
+            var innerHeight =$(this).innerHeight()-$(this).css('paddingTop').replace('px','')- $(this).css('paddingBottom').replace('px','');
+             if( innerHeight > heightLimit){
```
OmeGak: Inconsistent spacing.

```js
if (...) {
    // stuff
}    
```

indico/modules/events/tracks/templates/_track_list.html
OmeGak: Wrong indentation.

OmeGak: Inconsistent spacing around operators. We leave whitespace on either side.

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+                 $(this).addClass('gradient');
```
OmeGak: I would call the class differently, as gradient sounds more like the background color of the element.

indico/modules/events/tracks/templates/_track_list.html
OmeGak:
```js
$(this).on('click', function() {
    // ...
}
```

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,91 @@
...
+                $.ajax({
+                    url: $elem.data('url'),
+                    method: 'POST',
+                    contentType: 'application/json',
+                    data: JSON.stringify({'sort_order': sortedList}),
+                    complete: IndicoUI.Dialogs.Util.progress(),
+                    error: handleAjaxError
+                });
+
+                $('.i-box').each(function() {
```
OmeGak: I'm almost positive this can be done in CSS only. We can sit together and take a look into it.

indico/modules/events/tracks/templates/management.html
OmeGak: Extra whitespace before "Tracks".

indico/modules/events/tracks/templates/management.html
```diff
@@ -1,20 +1,25 @@
...
+        You can add, edit, remove or sort these tracks.
+    {% endtrans -%}
+{% endblock %}
 
+{% block content %}
     <a class="i-button highlight icon-plus"
         data-href="{{ url_for('.create', event) }}"
+        data-title="{% trans %}Add a new track{% endtrans %}"
```
OmeGak: `Add new track`

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -519,3 +519,43 @@ th.i-table {
...
+            }
+        color: $dark-gray;
```
OmeGak: I'm not sure this class should be changing the text color...

indico/modules/events/tracks/templates/_track_list.html
ThiefMaster: I don't think splitting `href=# class=...` into two lines is really needed. I'd put the title in a new line though

indico/modules/events/tracks/templates/_track_list.html
ThiefMaster: a `data-title` would be useful too

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -519,3 +519,43 @@ th.i-table {
...
+    padding-left: 0;
+    width: 800px;
+
+    .i-box {
+        padding: 10px 10px 0 10px ;
+        margin-bottom: 10px;
+    }
+
+    .i-box-header{
+        border-bottom: None;
```
nop33: `None` is not a valid CSS property value. Use `border-bottom: 0;` instead

indico/modules/events/tracks/templates/_track_list.html
nop33: @ThiefMaster if you read my suggestion you would see that it didn't say to split it in case of `href="#"` :stuck_out_tongue:

indico/modules/events/tracks/templates/_track_list.html
nop33: I created the `.gradient-layer` a few months ago, maybe you could reuse it or at least have a look

indico/modules/events/tracks/templates/_track_list.html
DavidAndreev: Well you need to add `margin-bottom` to the last i-box of the list, because the drag and drop sorting doesn't work if you want to drag after the last one without extra margin on it. And I'm not sure how to do it only in CSS cause last i-box can be allways some different one depending on the order. That's also why I do `$('.i-box').last().css('margin-bottom', '150px');` when you sort the list.

indico/modules/events/tracks/templates/_track_list.html
DavidAndreev: It needs to get executed every time you re-order the list, so I don't know how to do it only in CSS.

indico/htdocs/sass/modules/_event_management.scss
DavidAndreev: When we have a track without a description we want to show only the i-box header with the title and nothing in the i-box-content , but on default there is 10px of white space after the header so all of that is to get rid of it and also when we do have a description so it looks nice.

indico/modules/events/tracks/templates/management.html
DavidAndreev: Thanks for the explanation :) I finally understood :D

indico/modules/events/tracks/templates/management.html
```diff
@@ -1,20 +1,25 @@
...
         data-ajax-dialog
         data-reload-after>
         {% trans %}Add track{% endtrans %}
     </a>
```
ThiefMaster: FWIW, the whitespace usually does not matter on the outside. So it's up to you whether you want to strip it or not (we don't care whether the html output is pretty or not).

However, on the inside of a trans it does matter (if there's whitespace - here there's none anyway) since those messages get extracted to the translation dict and having surrounding whitespace there is quite awful

indico/htdocs/sass/modules/_event_management.scss
OmeGak: You could simply set `padding-bottom: 0`.

indico/modules/events/tracks/templates/_track_list.html
OmeGak: You have the `:last-child` pseudoselector to do that for you without JS. Why do we need to add margin at the end of the list, anyways?

indico/modules/events/tracks/templates/_track_list.html
OmeGak: I'll pass by, then. ;)

indico/modules/events/tracks/controllers.py
```diff
@@ -34,17 +40,34 @@ def _process(self):
...
+
+    normalize_url_spec = {
+        'locators': {
+            lambda self: self.track
+        }
+    }
+
+    def _checkParams(self, params):
+        RHManageTracksBase._checkParams(self, params)
+        self.track = Track.find_one(id=request.view_args['track_id'])
```
ThiefMaster: `Track.get_one(request.view_args['track_id'])`

indico/modules/events/tracks/operations.py
```diff
@@ -28,7 +28,21 @@ def create_track(event, data):
...
     return track
+
+
+def update_track(track, data):
+    track.populate_from_dict(data)
+    db.session.flush()
+    logger.info('Track %r modified by %r', track, session.user)
+    track.event_new.log(EventLogRealm.management, EventLogKind.change, 'Tracks',
+                        'Track "{}" has been modified.'.format(track.title), session.user)
+    return track
```
ThiefMaster: no need to return it here.

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +86,36 @@ def _process(self):
...
         return jsonify_form(form)
+
+
+class RHEditTrack(RHManageTrackBase):
+    """Edit a track"""
+
+    def _process(self):
+        form = TrackForm(obj=FormDefaults(self.track))
+        if form.validate_on_submit():
+            track = update_track(self.track, form.data)
```
ThiefMaster: just use `self.track` below. no need to return anything from `update_track`

indico/htdocs/js/indico/modules/tracks.js
```diff
@@ -0,0 +1,60 @@
...
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+(function(global) {
+    'use strict';
+
+    global.setupTracksManagement = function setupTracksManagement() {
```
ThiefMaster: `setupTrackManagement`?

indico/modules/events/tracks/blueprint.py
```diff
@@ -16,13 +16,17 @@
...
 from __future__ import unicode_literals
 
-from indico.modules.events.tracks.controllers import RHManageTracks, RHCreateTrack, RHCreateTrackOld
+from indico.modules.events.tracks.controllers import (RHManageTracks, RHCreateTrack, RHCreateTrackOld, RHEditTrack,
+                                                      RHSortTracks, RHDeleteTrack)
 from indico.web.flask.wrappers import IndicoBlueprint
 
 _bp = IndicoBlueprint('tracks', __name__, template_folder='templates', virtual_template_folder='events/tracks',
-                      url_prefix='/event/<confId>')
+                      url_prefix='/event/<confId>/manage')
```
ThiefMaster: #2651 adds `/program` which is not management. So I'd keep the `/manage` explicit instead of adding it to the prefix.

indico/htdocs/js/indico/modules/tracks.js
```diff
@@ -0,0 +1,60 @@
...
+        var heightLimit = 50;
```
nop33: Maybe add a comment here like ``// height limit in pixels/em``?

ThiefMaster: I think it's clear that it's pixels since we use it with `.height()`. But I'd clarify why it's 50px...

indico/htdocs/sass/partials/_core.scss
```diff
@@ -104,3 +104,15 @@
     overflow-x: hidden;
     overflow-y: auto;
 }
+
+.transparent-overlay {
+    position: relative;
+    &:before {
```
nop33: Empty line before this line

indico/htdocs/sass/partials/_core.scss
```diff
@@ -104,3 +104,15 @@
...
+        @include gradient(rgba(0, 0, 0, 0.0), white);
```
nop33: Empty line after this line
https://github.com/indico/indico/wiki/SASS-Guidelines

indico/modules/events/tracks/controllers.py
```diff
@@ -34,17 +40,34 @@ def _process(self):
...
+
+
 class RHManageTracks(RHManageTracksBase):
     def _process(self):
         tracks = self.event_new.tracks
         return WPManageTracks.render_template('management.html', self._conf, event=self.event_new, tracks=tracks)
 
 
 class RHCreateTrack(RHManageTracksBase):
+    """Create a track"""
```
nop33: I think the class name is quite self explanatory

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +86,36 @@ def _process(self):
...
+    def _process(self):
+        sort_order = request.json['sort_order']
+        tracks = {t.id: t for t in self.event_new.tracks}
+        for position, track_id in enumerate(sort_order, 1):
+            if track_id in tracks:
+                tracks[track_id].position = position
+
+
+class RHDeleteTrack(RHManageTrackBase):
+    """Delete a track"""
```
nop33: Aaaaaand here

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,46 @@
+{% from 'message_box.html' import message_box %}
+
+{% macro render_track_list(event) %}
+    {% if not event.tracks %}
+        {% call message_box('info', classes='no-rules-placeholder') %}
+            {% trans %}No tracks have been added yet.{% endtrans%}
```
nop33: Add a space after `endtrans`

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,46 @@
...
+                <li class="track-row i-box" data-id="{{ track.id }}">
+                    <div class="ui-i-box-sortable-handle"></div>
+                    <div class="i-box-header">
+                        <span class="i-box-title">{{ track.title }}</span>
+                        <div class="i-box-buttons toolbar right">
+                            <div class="group">
+                                <a href="#" class="i-button icon-edit"
+                                   title="{% trans %}Edit track{% endtrans %}"
+                                   data-href="{{ url_for('tracks.edit_track', track) }}"
+                                   data-title="{% trans title=track.title %}Edit track &quot;{{ title }}&quot;{% endtrans %}"
```
nop33: data-title='{% trans title=track.title %}Edit track "{{ title }}"{% endtrans %}'

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,46 @@
+{% from 'message_box.html' import message_box %}
+
+{% macro render_track_list(event) %}
+    {% if not event.tracks %}
+        {% call message_box('info', classes='no-rules-placeholder') %}
+            {% trans %}No tracks have been added yet.{% endtrans%}
+        {% endcall %}
+    {% else %}
+        <ul id="track-list" class="i-box-group vert" data-url="{{ url_for('.sort_tracks', event) }}">
```
ThiefMaster: This need to be outside the macro. Currently this gets duplicated when we update via AJAX (since we set the `#track-list` content of the html from this macro.)

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -0,0 +1,46 @@
...
+                                   title="{% trans %}Edit track{% endtrans %}"
+                                   data-href="{{ url_for('tracks.edit_track', track) }}"
+                                   data-title="{% trans title=track.title %}Edit track &quot;{{ title }}&quot;{% endtrans %}"
+                                   data-reload-after
+                                   data-ajax-dialog></a>
+                                <a href="#" class="i-button danger text-color icon-remove js-delete-item"
+                                   title="{% trans %}Delete track{% endtrans %}"
+                                   data-href="{{ url_for('tracks.delete_track', track) }}"
+                                   data-confirm="{% trans %}Are you sure you want to delete this track?{% endtrans %}"
+                                   data-title="{% trans title = track.title %}Delete track &quot;{{ title }}&quot;{% endtrans %}"
```
nop33: Same here

indico/htdocs/js/indico/modules/tracks.js
OmeGak: `var heightLimitPx`? I would rather not use a comment for this.

indico/modules/events/tracks/controllers.py
```diff
@@ -16,10 +16,16 @@
...
 from __future__ import unicode_literals
 
+from flask import request, flash
+
 from indico.modules.events.tracks.forms import TrackForm
-from indico.modules.events.tracks.operations import create_track
+from indico.modules.events.tracks.models.tracks import Track
+from indico.modules.events.tracks.operations import create_track, update_track, delete_track
 from indico.modules.events.tracks.views import WPManageTracks
+from indico.web.flask.templating import get_template_module
```
OmeGak: Wrong import order.

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +86,36 @@ def _process(self):
                                 tracks=[{'id': int(t.getId()), 'title': to_unicode(t.getTitle())}
                                         for t in self._conf.getTrackList()])
         return jsonify_form(form)
+
+
+class RHEditTrack(RHManageTrackBase):
+    """Edit a track"""
+
+    def _process(self):
+        form = TrackForm(obj=FormDefaults(self.track))
```
OmeGak: I think `TrackForm(obj=self.track)` would be enough.

indico/modules/events/tracks/controllers.py
ThiefMaster: Why? modules < flask < forms

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +86,36 @@ def _process(self):
...
+        for position, track_id in enumerate(sort_order, 1):
+            if track_id in tracks:
+                tracks[track_id].position = position
+
+
+class RHDeleteTrack(RHManageTrackBase):
+    """Delete a track"""
+
+    def _process(self):
+        delete_track(self.track)
```
DavidAndreev: ` def can_delete(self):
        if self.event_new.can_manage(session.user):
            if not self.abstract_accepted:
                return True
        return False`
something like this ?

ThiefMaster:
```python
def can_delete(self, user):
    if not user:
        return False
    elif not self.event_new.can_manage(user):
        return False
    else:
        # Allow deletion only if no abstracts have been accepted for this track.
        return not self.abstracts_accepted
```

nop33: you can simply do
```
return self.event_new.can_manage(session.user) and not self.abstract_accepted
```

DavidAndreev: So if we cannot delete, should I show some dialog with a message or something like that ?

OmeGak: We should not give the option of deleting in the UI if it's not possible. In the case the request arrives it can be one of two options: 
1. In between rendering the page and clicking the button, an abstract has been accepted in that track.
2. Somebody is messing around.

The easy solution is raising an exception and the user-friendly solution is handling it so that it doesn't look like an error. I would go for the latter.

ThiefMaster: I would just silently not delete it in this case. It's way too unlikely to happen to justify any pretty error handling. If it happens to a user they'll probably try to delete again and then see the tooltip of the disabled delete icon.

indico/modules/events/tracks/templates/_track_list.html
DavidAndreev: I don't really understand, what gets duplicated? And when I take it out of the macro: when you delete a track we only update `#track-list` not the whole page, and then the JS which applies the gradient doesn't work.

indico/htdocs/js/indico/modules/tracks.js
```diff
@@ -0,0 +1,62 @@
...
+(function(global) {
+    'use strict';
+
+    global.setupTrackManagement = function setupTrackManagement() {
+        var heightLimit = 50;
+        $('#track-list').on('indico:htmlUpdated', function() {
+            $('.track-content').each(function() {
+                var $this = $(this);
+                if ($this.height() > heightLimit) {
+                    $this.addClass('transparent-overlay track-content-collapsed track-content-collapsible');
```
OmeGak: `transparent-overlay` always gets added/removed with `track-content-collapsed`. I propose completely omitting `trasparent-overlay` here and add it directly to `track-content-collapsed`. It would look like that in the SCSS file.

```scss
.track-content-collapsed {
    @extend .transparent-overlay;
    max-height: 50px;
    overflow: hidden;
}
```

indico/htdocs/js/indico/modules/tracks.js
```diff
@@ -0,0 +1,62 @@
...
+                    });
+                }
+            });
+        }).trigger('indico:htmlUpdated');
+
+        $('#track-list').sortable({
+            axis: 'y',
+            containment: 'parent',
+            cursor: 'move',
+            distance: 2,
```
OmeGak: It seems we go for `distance: 10` in other places. Any reason why you went for `2` here?

indico/htdocs/sass/partials/_core.scss
```diff
@@ -104,3 +104,17 @@
     overflow-x: hidden;
     overflow-y: auto;
 }
+
+.transparent-overlay {
+    position: relative;
+
+    &:before {
+        @include gradient(rgba(0, 0, 0, 0.0), white);
+
```
OmeGak: I'd say there's no need to add blank line after SASS declarations.

indico/modules/events/tracks/blueprint.py
```diff
@@ -16,13 +16,17 @@
...
+                                                      RHSortTracks, RHDeleteTrack)
 from indico.web.flask.wrappers import IndicoBlueprint
 
 _bp = IndicoBlueprint('tracks', __name__, template_folder='templates', virtual_template_folder='events/tracks',
                       url_prefix='/event/<confId>')
 
 _bp.add_url_rule('/manage/tracks/', 'manage', RHManageTracks)
 _bp.add_url_rule('/manage/tracks/create', 'create', RHCreateTrack, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/tracks/<int:track_id>', 'edit_track', RHEditTrack, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/tracks/<int:track_id>', 'delete_track', RHDeleteTrack, methods=['DELETE'])
```
OmeGak: `('DELETE',)`

ThiefMaster:
Also needs to be rebased

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +83,31 @@ def _process(self):
...
+            if track_id in tracks:
+                tracks[track_id].position = position
+
+
+class RHDeleteTrack(RHManageTrackBase):
+    def _process(self):
+        delete_track(self.track)
+        tpl = get_template_module('events/tracks/_track_list.html', event=self.event_new)
+        flash(_('Track "{}" has been deleted.').format(self.track.title), 'success')
+        return jsonify_data(html=tpl.render_track_list(self.event_new), flash=True)
```
ThiefMaster: `flash=True` is the default; no need to specify it

indico/modules/events/tracks/templates/management.html
```diff
@@ -1,20 +1,35 @@
...
 
+{% block content %}
     <a class="i-button highlight icon-plus"
         data-href="{{ url_for('.create', event) }}"
+        data-title="{% trans %}Add new track{% endtrans %}"
         data-ajax-dialog
         data-reload-after>
         {% trans %}Add track{% endtrans %}
     </a>
+        {% if not event.tracks %}
```
ThiefMaster: indentation...

ThiefMaster: Try adding a track to an event with no tracks. I think you need to use a container element around both this message and the UL that has the ID (you can then move both this part and the `<ul>` into the macro)

indico/htdocs/js/indico/modules/tracks.js
nop33: If we are to not leave an empty line after `@extend` and `@include`, we better update the SASS style guide, but in my opinion it's looks nicer with an empty line in between:
```scss
.track-content-collapsed {
    @extend .transparent-overlay;

    max-height: 50px;
    overflow: hidden;
}
```

indico/modules/events/tracks/templates/management.html
```diff
@@ -20,15 +21,17 @@
...
-                {% trans %}No tracks have been added yet.{% endtrans %}
-            {% endcall %}
-        {% else %}
-            <ul id="track-list" class="i-box-group vert" data-url="{{ url_for('.sort_tracks', event) }}">
-                {{ render_track_list(event) }}
-            </ul>
-        {% endif %}
+    {% if not event.tracks %}
+        <div id="no-tracks-box">
+        {% call message_box('info', classes='no-rules-placeholder') %}
```
ThiefMaster: indentation...

indico/modules/events/tracks/templates/management.html
```diff
@@ -20,15 +21,17 @@
...
-        {% if not event.tracks %}
-            {% call message_box('info', classes='no-rules-placeholder') %}
-                {% trans %}No tracks have been added yet.{% endtrans %}
-            {% endcall %}
-        {% else %}
-            <ul id="track-list" class="i-box-group vert" data-url="{{ url_for('.sort_tracks', event) }}">
-                {{ render_track_list(event) }}
-            </ul>
-        {% endif %}
+    {% if not event.tracks %}
```
ThiefMaster: the if/else part has to go inside the macro. otherwise it won't work when you delete the last / add the first track.

indico/htdocs/sass/modules/_tracks.scss
```diff
@@ -0,0 +1,55 @@
...
+        .i-box-header {
+            margin-bottom: 0;
+        }
+
+        .i-box-content {
+            word-wrap: break-word;
+            padding: 10px 0px;
+        }
+    }
+    .track-content-collapsible {
```
nop33: Empty line above.

indico/htdocs/sass/modules/_tracks.scss
```diff
@@ -0,0 +1,55 @@
...
+        cursor: pointer;
+    }
+
+    .track-content-collapsed {
+        @extend .transparent-overlay;
```
nop33: Empty line bellow.
https://github.com/indico/indico/wiki/SASS-Guidelines

indico/htdocs/sass/partials/_core.scss
```diff
@@ -104,3 +104,16 @@
     overflow-x: hidden;
     overflow-y: auto;
 }
+
+.transparent-overlay {
+    position: relative;
+
+    &:before {
+        @include gradient(rgba(0, 0, 0, 0.0), white);
```
nop33: Same here

indico/htdocs/sass/modules/_tracks.scss
OmeGak: I don't agree. Should we maybe study using this instead? https://github.com/brigade/scss-lint

ThiefMaster:
I don't think it's really worth the effort :) FWIW, those three commits could probably even be squashed into a single one.

indico/htdocs/js/indico/modules/tracks.js
```diff
@@ -0,0 +1,63 @@
...
+                var $elem = $('#track-list');
+                var sortedList = $elem.find('li.track-row').map(function() {
+                    return $(this).data('id');
+                }).get();
+
+                $.ajax({
+                    url: $elem.data('url'),
+                    method: 'POST',
+                    contentType: 'application/json',
+                    data: JSON.stringify({'sort_order': sortedList}),
```
ThiefMaster: no need for the quotes

indico/modules/events/tracks/templates/management.html
```diff
@@ -1,20 +1,29 @@
...
 
-{% block content %}
-    {% for track in tracks %}
-        <div>
-            {{ track.title }}
-        </div>
-    {% endfor %}
+{% block description %}
+    {%- trans -%}
+        Here you can find a list of tracks for this event.
```
ThiefMaster: This description is not very useful. Better explain what tracks are used for.

> Tracks represent the subject matter of the conference. They represent both the basis for the Scientific Programme and a taxonomy that can be using to classify submissions during the Call for Abstracts process.

Something similar to this (best to discuss on IRC on the wording)

indico/modules/events/tracks/templates/management.html
```diff
@@ -1,20 +1,29 @@
...
+        You can add, edit, remove or sort these tracks.
+    {%- endtrans -%}
+{% endblock %}
 
+{% block content %}
     <a class="i-button highlight icon-plus"
         data-href="{{ url_for('.create', event) }}"
+        data-title="{% trans %}Add new track{% endtrans %}"
         data-ajax-dialog
         data-reload-after>
```
ThiefMaster: Why are we still reloading the page here instead of using `data-update` like for edit/delete?

indico/modules/events/tracks/blueprint.py
```diff
@@ -16,15 +16,18 @@
...
 from indico.web.flask.wrappers import IndicoBlueprint
 
 _bp = IndicoBlueprint('tracks', __name__, template_folder='templates', virtual_template_folder='events/tracks',
                       url_prefix='/event/<confId>')
 
 _bp.add_url_rule('/manage/tracks/', 'manage', RHManageTracks)
 _bp.add_url_rule('/manage/tracks/create', 'create', RHCreateTrack, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/tracks/<int:track_id>', 'edit_track', RHEditTrack, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/tracks/<int:track_id>', 'delete_track', RHDeleteTrack, methods=('DELETE',))
+_bp.add_url_rule('/manage/tracks/sort', 'sort_tracks', RHSortTracks, methods=('POST',))
```
ThiefMaster: I'd move this up, right below the `create` endpoint (which I'd rename to `create_track` to be consistent with the others) so the not-track-specific endpoints are next to each other

indico/modules/events/tracks/templates/management.html
DavidAndreev: It doesn't work with data-update, don't know why.

indico/modules/events/tracks/controllers.py
```diff
@@ -84,3 +109,29 @@ class RHTracksPDF(RHConferenceBaseDisplay):
     def _process(self):
         pdf = ProgrammeToPDF(self.event_new)
         return send_file('program.pdf', BytesIO(pdf.getPDFBin()), 'application/pdf')
+
+
+class RHEditTrack(RHManageTrackBase):
```
ThiefMaster: Move these RHs below `RHCreateTrack` so the management stuff is together

indico/modules/events/tracks/templates/management.html
```diff
@@ -1,20 +1,30 @@
...
+{% endblock %}
 
+{% block content %}
     <a class="i-button highlight icon-plus"
-        data-href="{{ url_for('.create', event) }}"
+        data-href="{{ url_for('.create_track', event) }}"
+        data-title="{% trans %}Add new track{% endtrans %}"
         data-ajax-dialog
-        data-reload-after>
+        data-update="#track-list">
```
ThiefMaster: move one line up (so `data-ajax-dialog` is the last one)

OmeGak:
:tada:

PR #2651, opened 2016-10-17
Tracks display page

OmeGak:


indico/MaKaC/webinterface/pages/static.py
```diff
@@ -47,7 +47,7 @@ class WPStaticTimetable(WPStaticEventBase, WPDisplayTimetable):
 
 
 class WPStaticConferenceProgram(WPStaticEventBase, WPConferenceProgram):
```
ThiefMaster: I don't think this will work. This uses a legacy WP which gets the data from legacy.

https://github.com/indico/indico/blob/master/indico/MaKaC/common/offlineWebsiteCreator.py#L314

You can remove the WP handling there (program is the only remaining one) and add the new RH instead. Then create a new WP similar to the other "new" static WPs here.

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +65,11 @@ def _process(self):
                                 tracks=[{'id': int(t.getId()), 'title': to_unicode(t.getTitle())}
                                         for t in self._conf.getTrackList()])
         return jsonify_form(form)
+
+
+class RHDisplayTracks(RHConferenceBaseDisplay):
+    def _process(self):
+        page_title = get_menu_entry_by_name('program', self._conf).localized_title
+        program_description = self._conf.getProgramDescription()
```
ThiefMaster: Move to SettingsProxy?

OmeGak: Wouldn't it fit better as an `Event` property?

ThiefMaster: When moving it don't forget to add it to the event_abstracts zodbimporter

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -94,15 +95,12 @@ def enumerate_affil(person_links):
...
-    def __init__(self, conf, doc=None, story=None, tz=None):
-        self._conf = conf
-        self._event = conf.as_event
-        if not tz:
-            self._tz = self._conf.getTimezone()
-        else:
-            self._tz = tz
-        PDFBase.__init__(self, doc, story)
-        self._title = _("Conference Scientific Programme")
+    def __init__(self, event, tz=None):
```
OmeGak: Before anybody says anything. We need `tz=None` to play well with legacy stuff.

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -94,15 +95,12 @@ def enumerate_affil(person_links):
...
+        self._event = event
+        self._conf = event.as_legacy
+        self._tz = self._event.tzinfo
+        self._title = get_menu_entry_by_name('program', self._conf).localized_title
```
ThiefMaster: does this really need a Conference or does it work with an Event too?

OmeGak: It was failing with `event` for offline site generation only.

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -94,15 +95,12 @@ def enumerate_affil(person_links):
...
+        PDFBase.__init__(self, title='program.pdf')
```
ThiefMaster: Why not use `self._title` for the title here? A filename in the title seems weird.

indico/MaKaC/common/offlineWebsiteCreator.py
```diff
@@ -317,14 +318,12 @@ def __init__(self, rh, conf, event_type=""):
         # Those which are backed by a WP class get their name from that class;
         # the others are simply hardcoded.
         self._menu_offline_items = {'overview': None, 'abstracts_book': None}
-        wps = {WPStaticConferenceProgram}
-        for cls in wps:
-            self._menu_offline_items[cls.menu_entry_name] = cls(self._rh, self._conf)
```
ThiefMaster: I think this code further down can go away now too:

```
        if isinstance(obj, WPBase):
            content = obj.display()
            self._addPage(content, obj.endpoint, self._conf)
```

indico/modules/events/tracks/blueprint.py
```diff
@@ -25,4 +26,7 @@
 _bp.add_url_rule('/manage/tracks/', 'manage', RHManageTracks)
 _bp.add_url_rule('/manage/tracks/create', 'create', RHCreateTrack, methods=('GET', 'POST'))
 
+_bp.add_url_rule('/program/', 'program', RHDisplayTracks)
```
ThiefMaster: `/program` and `/program.pdf`?

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +70,17 @@ def _process(self):
...
+        page_title = get_menu_entry_by_name('program', self._conf).localized_title
+        program = track_settings.get(self.event_new, 'program')
+        return WPDisplayTracks.render_template('display.html', self._conf, event=self.event_new, page_title=page_title,
+                                               program=program)
+
+
+class RHTracksPDF(RHConferenceBaseDisplay):
+    def _process(self):
+        pdf = ProgrammeToPDF(self.event_new)
+        return send_file('programme.pdf', BytesIO(pdf.getPDFBin()), 'application/pdf')
```
ThiefMaster: `program.pdf`?

indico/modules/events/tracks/controllers.py
```diff
@@ -63,3 +70,17 @@ def _process(self):
...
+
+class RHDisplayTracks(RHConferenceBaseDisplay):
+    def _process(self):
+        page_title = get_menu_entry_by_name('program', self._conf).localized_title
+        program = track_settings.get(self.event_new, 'program')
+        return WPDisplayTracks.render_template('display.html', self._conf, event=self.event_new, page_title=page_title,
+                                               program=program)
+
+
+class RHTracksPDF(RHConferenceBaseDisplay):
```
ThiefMaster: `RHProgramPDF`

indico/modules/events/tracks/settings.py
```diff
@@ -0,0 +1,20 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from indico.modules.events.settings import EventSettingsProxy
+
+
+track_settings = EventSettingsProxy('tracks', {'program': ''})
```
ThiefMaster: I'd wrap after `{` to be consistent with all the other settings proxies

indico/modules/events/tracks/templates/display.html
```diff
@@ -0,0 +1,36 @@
...
+            <a href="{{ url_for('tracks.program_pdf', event) }}"
+               class="i-button icon-file-pdf"
+               title="{% trans %}Download PDF{% endtrans %}"></a>
+        </div>
+    </div>
+{% endblock %}
+
+{% block content %}
+    {% if program %}
+        <div class="quotation programme">
```
ThiefMaster: it's "program" internally (our code is not en_GB ;)) so the CSS class should be `program` too

indico/modules/events/tracks/views.py
```diff
@@ -17,9 +17,17 @@
...
-from MaKaC.webinterface.pages.conferences import WPConferenceModifBase
+from MaKaC.webinterface.pages.conferences import WPConferenceModifBase, WPConferenceDefaultDisplayBase
 
 
 class WPManageTracks(WPJinjaMixin, WPConferenceModifBase):
     template_prefix = 'events/tracks/'
     sidemenu_option = 'program'
+
+
+class WPDisplayTracks(WPJinjaMixin, WPConferenceDefaultDisplayBase):
```
ThiefMaster: `WPDisplayProgram`

indico/MaKaC/PDFinterface/conference.py
OmeGak: No strong opinion. It's what we do in other places.

indico/modules/events/tracks/blueprint.py
OmeGak: I followed the same pattern as in `/contributions/` and `/contributions/contributions.pdf`, but I agree I like this better.

indico/MaKaC/PDFinterface/conference.py
ThiefMaster: there's only one place (legacy program) where we pass this argument.. dunno what we use in the other cases. `self._title` maybe?

indico/modules/events/tracks/blueprint.py
ThiefMaster: probably because we have more stuff below `/contributions/..`

indico/modules/events/tracks/templates/display.html
```diff
@@ -0,0 +1,36 @@
+{% extends 'layout/conference_page_base.html' %}
+
+{% block title %}
+    {{- page_title -}}
```
OmeGak: A matter of consistency since there is no block/endblock tags in more common cases like these:

```jinja
<div>
    {{- foo -}}
</div>
```

indico/modules/events/tracks/templates/display.html
OmeGak: I'm waiting for @DavidAndreev's https://github.com/indico/indico/pull/2646 to get merged to move/rename the classes to `_tracks.scss`.

indico/MaKaC/PDFinterface/conference.py
OmeGak: I don't know where `PDFLaTeXBase` gets the PDF title from, but for `contributions/contributions.pdf` the title metadata is `contributions.pdf` (as it gets displayed in the browser tab). In `PDFBase` we were simply not passing `title` as I just added the argument yesterday, so the tab name was always `(anonymous)`.

indico/modules/events/tracks/controllers.py
OmeGak: Why not? We send `contributions.pdf`.

OmeGak: Sorry, I didn't even notice you were using a different and more British word. :P

ThiefMaster: programme is the british one, not the other way round ;)

PR #2652, opened 2016-10-17
Abstracts: rendering of submission information

pferreir:


indico/htdocs/sass/modules/abstracts/_abstract_page.scss
```diff
@@ -141,3 +140,64 @@
...
+.submitted-data {
+    .field-separator {
+        @include border-top;
+        margin: 0 -10px;
+        padding: 1em 10px;
+    }
+
+    .comment {
+        margin-bottom: 1em;
+        @extend .text-not-important;
```
nop33: I usually put the `@extend` and `@include` in the beginning of the selector's body and leave an empty line after. I've seen it in quite a few places. Maybe we could make it a "rule"? Just my opinion.

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
```diff
@@ -141,3 +140,64 @@
...
+    }
+
+    dl.custom-field-values {
+        .field-value {
+            margin-bottom: 0.5em;
+            & > * {
```
nop33: Empty line before the beginning of every selector inside the body of another maybe?
```scss
dl.custom-field-values {
    .field-value {
        margin-bottom: 0.5em;
     
        & > * {
            ....
        }
    }
}
```

OmeGak: On a different note, you don't need `&` here.

indico/modules/events/abstracts/blueprint.py
```diff
@@ -39,6 +40,8 @@
 # Display pages
 _bp.add_url_rule('/abstracts/<int:abstract_id>/', 'display_abstract', RHDisplayAbstract)
 _bp.add_url_rule('/abstracts/<int:abstract_id>/abstract.pdf', 'display_abstract_pdf_export', RHDisplayAbstractExportPDF)
+_bp.add_url_rule('/abstracts/<int:abstract_id>/attachments/<file_id>/<filename>', 'download_attachment',
```
OmeGak: Do we need `<filename>` at all?

ThiefMaster: We don't need one but it's much prettier to have one (we do the same thing for other download endpoints too)

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
```diff
@@ -141,3 +140,64 @@
...
+.submission-info {
+    .info-line {
+        margin-top: 1em;
+
+        &:first-child {
+            margin-top: 0;
+        }
+
+        .author-list {
+            font-size: 0.95em;
```
pferreir: It is perceptible. No particular reason, just looks nice.

PR #2654, opened 2016-10-18
UI improvements in email notifications layout

mvidalgarcia:
How it is currently looking:

image

Suggestions to improve the UX are welcome

### DIFF
--- a/indico/htdocs/sass/modules/abstracts/_email_templates.scss
+++ b/indico/htdocs/sass/modules/abstracts/_email_templates.scss
@@ -61,6 +61,10 @@
     .email-template-wrapper {
         padding: 0 0 1em 1em;
 
+        .explanation {
+            color: $light-black;
+        }
+
         .email-templates {
             width: 800px;
             max-height: 800px;
@@ -100,10 +104,10 @@
         .email-preview-btn {
             padding-right: 5px;
             color: $link;
-        }
 
-        .email-preview-btn:hover {
-            color: $bright-orange;
+            &:hover {
+                color: $bright-orange;
+            }
         }
     }
 }
@@ -114,6 +118,10 @@
     .form-field,
     .rule-condition[data-condition="track"] select {
         width: 100%;
+    }
+
+    .rule-list {
+        width: 700px;
     }
 }
 
--- a/indico/htdocs/sass/modules/abstracts/_email_templates.scss
+++ b/indico/htdocs/sass/modules/abstracts/_email_templates.scss
@@ -55,16 +55,65 @@
     }
 }
 
-.email-template-list {
-    list-style-type: none;
-    padding-left: 0;
-    width: 800px;
+#email-template-manager {
+    max-width: 800px;
 
-    .email-template {
-        margin-bottom: 10px;
+    .email-template-wrapper {
+        padding: 0 0 1em 1em;
+
+        .email-templates {
+            width: 800px;
+            max-height: 800px;
+            overflow: auto;
+
+            > ul {
+                list-style-type: none;
+                padding-left: 0;
+                padding-right: 20px;
+
+                > li {
+                    width: 775px;
+                }
+            }
+        }
+
+        .email-preview {
+            display: none;
+
+            .email-subject {
+                padding-bottom: 10px;
+            }
+
+            .email-body {
+                color: $light-black;
+            }
+        }
+
+        .regform-section-sortable-placeholder {
+            @extend .i-box;
+            @include border-all($style: dashed);
+            @include box-shadow(none);
+
+            background-color: $light-gray;
+        }
+
+        .email-preview-btn {
+            padding-right: 5px;
+            color: $link;
+        }
+
+        .email-preview-btn:hover {
+            color: $bright-orange;
+        }
     }
-
-    /* this padding is needed so that dragging after
-       the last element works properly. */
-    padding-bottom: 15px;
 }
+
+#form-group-rules {
+    margin-top: 0;
+
+    .form-field,
+    .rule-condition[data-condition="track"] select {
+        width: 100%;
+    }
+}
+
--- a/indico/modules/events/abstracts/controllers/email_templates.py
+++ b/indico/modules/events/abstracts/controllers/email_templates.py
@@ -16,7 +16,9 @@
 
 from __future__ import unicode_literals
 
+
 from flask import request
+from werkzeug.exceptions import BadRequest
 
 from indico.core.db import db
 from indico.modules.events.abstracts.controllers.management import RHManageAbstractsBase
@@ -105,6 +107,23 @@ class RHDeleteEmailTemplate(RHEditEmailTemplateBase):
     def _process(self):
         db.session.delete(self.email_tpl)
         return _render_notification_list(self.event_new, flash=False)
+
+
+class RHEmailTemplateREST(RHEditEmailTemplateBase):
+    """Enable or disable stop on match of a notification template"""
+
+    def _process_PATCH(self):
+        if request.json is None:
+            raise BadRequest('Expected JSON payload')
+
+        invalid_fields = request.json.viewkeys() - {'stop_on_match'}
+        if invalid_fields:
+            raise BadRequest("Invalid fields: {}".format(', '.join(invalid_fields)))
+
+        if 'stop_on_match' in request.json:
+            self.email_tpl.stop_on_match = request.json['stop_on_match']
+
+        return jsonify_data(flash=False)
 
 
 def _get_rules_fields(event_new):
--- a/indico/modules/events/abstracts/controllers/email_templates.py
+++ b/indico/modules/events/abstracts/controllers/email_templates.py
@@ -107,17 +107,6 @@ class RHDeleteEmailTemplate(RHEditEmailTemplateBase):
         return _render_notification_list(self.event_new, flash=False)
 
 
-class RHPreviewEmailTemplate(RHEditEmailTemplateBase):
-    """Preview an e-mail template."""
-
-    def _process(self):
-        self.commit = False
-        abstract = create_mock_abstract(self.event_new)
-        tpl = get_abstract_notification_tpl_module(self.email_tpl, abstract)
-        return jsonify_template('events/abstracts/management/notification_preview.html',
-                                subject=tpl.get_subject(), body=tpl.get_body())
-
-
 def _get_rules_fields(event_new):
     abstract = create_mock_abstract(event_new)
     email_tpl_dict = {et.id: get_abstract_notification_tpl_module(et, abstract)
--- a/indico/modules/events/abstracts/controllers/email_templates.py
+++ b/indico/modules/events/abstracts/controllers/email_templates.py
@@ -22,6 +22,7 @@ from indico.core.db import db
 from indico.modules.events.abstracts.controllers.management import RHManageAbstractsBase
 from indico.modules.events.abstracts.forms import (EditEmailTemplateRuleForm, EditEmailTemplateTextForm,
                                                    CreateEmailTemplateForm)
+from indico.modules.events.abstracts.models.abstracts import AbstractState
 from indico.modules.events.abstracts.models.email_templates import AbstractEmailTemplate
 from indico.modules.events.abstracts.notifications import get_abstract_notification_tpl_module
 from indico.modules.events.abstracts.util import build_default_email_template, create_mock_abstract
@@ -33,7 +34,8 @@ class RHEmailTemplateList(RHManageAbstractsBase):
     """Display list of e-mail templates."""
 
     def _process(self):
-        return jsonify_template('events/abstracts/management/notification_tpl_list.html', event=self.event_new)
+        return jsonify_template('events/abstracts/management/notification_tpl_list.html', event=self.event_new,
+                                **_get_rules_fields(self.event_new))
 
 
 class RHAddEmailTemplate(RHManageAbstractsBase):
@@ -46,9 +48,7 @@ class RHAddEmailTemplate(RHManageAbstractsBase):
             form.populate_obj(new_tpl)
             self.event_new.abstract_email_templates.append(new_tpl)
             db.session.flush()
-
-            tpl = get_template_module('events/abstracts/management/_notification_tpl_list.html')
-            return jsonify_data(html=tpl.render_notification_list(self.event_new))
+            return _render_notification_list(self.event_new)
         return jsonify_template('events/abstracts/management/notification_tpl_form.html', form=form)
 
 
@@ -73,10 +73,8 @@ class RHEditEmailTemplateRules(RHEditEmailTemplateBase):
         form = EditEmailTemplateRuleForm(obj=self.email_tpl, event=self.event_new)
         if form.validate_on_submit():
             form.populate_obj(self.email_tpl)
-            tpl = get_template_module('events/abstracts/management/_notification_tpl_list.html')
-            return jsonify_data(html=tpl.render_notification_list(self.event_new))
-        return jsonify_template('events/abstracts/management/notification_tpl_form.html',
-                                form=form, is_edit=True)
+            return _render_notification_list(self.event_new)
+        return jsonify_template('events/abstracts/management/notification_tpl_form.html', form=form, is_edit=True)
 
 
 class RHEditEmailTemplateText(RHEditEmailTemplateBase):
@@ -86,8 +84,7 @@ class RHEditEmailTemplateText(RHEditEmailTemplateBase):
         form = EditEmailTemplateTextForm(obj=self.email_tpl, event=self.event_new)
         if form.validate_on_submit():
             form.populate_obj(self.email_tpl)
-            tpl = get_template_module('events/abstracts/management/_notification_tpl_list.html')
-            return jsonify_data(html=tpl.render_notification_list(self.event_new))
+            return _render_notification_list(self.event_new)
         return jsonify_template('events/abstracts/management/notification_tpl_text_form.html', form=form, is_edit=True)
 
 
@@ -107,8 +104,7 @@ class RHDeleteEmailTemplate(RHEditEmailTemplateBase):
 
     def _process(self):
         db.session.delete(self.email_tpl)
-        tpl = get_template_module('events/abstracts/management/_notification_tpl_list.html')
-        return jsonify_data(html=tpl.render_notification_list(self.event_new), flash=False)
+        return _render_notification_list(self.event_new, flash=False)
 
 
 class RHPreviewEmailTemplate(RHEditEmailTemplateBase):
@@ -120,3 +116,17 @@ class RHPreviewEmailTemplate(RHEditEmailTemplateBase):
         tpl = get_abstract_notification_tpl_module(self.email_tpl, abstract)
         return jsonify_template('events/abstracts/management/notification_preview.html',
                                 subject=tpl.get_subject(), body=tpl.get_body())
+
+
+def _get_rules_fields(event_new):
+    abstract = create_mock_abstract(event_new)
+    email_tpl_dict = {et.id: get_abstract_notification_tpl_module(et, abstract)
+                      for et in event_new.abstract_email_templates}
+    return {'tracks': {track.id: track for track in event_new.tracks},
+            'states': AbstractState.__titles__, 'contrib_types': {ct.id: ct for ct in event_new.contribution_types},
+            'email_tpls': email_tpl_dict}
+
+
+def _render_notification_list(event, flash=True):
+    tpl = get_template_module('events/abstracts/management/_notification_tpl_list.html')
+    return jsonify_data(html=tpl.render_notification_list(event, **_get_rules_fields(event)), flash=flash)
--- a/indico/modules/events/abstracts/forms.py
+++ b/indico/modules/events/abstracts/forms.py
@@ -298,9 +298,6 @@ class EditEmailTemplateRuleForm(IndicoForm):
 
     title = StringField(_("Title"), [DataRequired()])
     rules = EmailRuleListField(_("Rules"), [DataRequired()])
-    stop_on_match = BooleanField(_("Stop on match"), [DataRequired()], widget=SwitchWidget(), default=True,
-                                 description=_("If any of the rules from this email template match, do not "
-                                               "send notifications from any following email template."))
 
     def __init__(self, *args, **kwargs):
         self.event = kwargs.pop('event')
@@ -341,7 +338,8 @@ class CreateEmailTemplateForm(EditEmailTemplateRuleForm):
         ('accept', _('Accept')),
         ('reject', _('Reject')),
         ('merge', _('Merge'))
-    ], description=_("The template that will be used as a basis for this notification. You can customize it later."))
+    ], description=_("The default template that will be used as a basis for this notification. "
+                     "You can customize it later."))
 
 
 class AbstractForm(IndicoForm):
--- a/indico/modules/events/abstracts/forms.py
+++ b/indico/modules/events/abstracts/forms.py
@@ -299,7 +299,8 @@ class EditEmailTemplateRuleForm(IndicoForm):
     title = StringField(_("Title"), [DataRequired()])
     rules = EmailRuleListField(_("Rules"), [DataRequired()])
     stop_on_match = BooleanField(_("Stop on match"), [DataRequired()], widget=SwitchWidget(), default=True,
-                                 description=_("Do not evaluate any other rules once this one matches."))
+                                 description=_("If any of the rules from this email template match, do not "
+                                               "send notifications from any following email template."))
 
     def __init__(self, *args, **kwargs):
         self.event = kwargs.pop('event')
@@ -317,7 +318,7 @@ class EditEmailTemplateTextForm(IndicoForm):
 
     reply_to_address = SelectField(_('"Reply to" address'), [DataRequired()])
     include_submitter = BooleanField(_('Send to submitter'), widget=SwitchWidget())
-    include_authors = BooleanField(_('Send to primary authorized_submittershors'), widget=SwitchWidget())
+    include_authors = BooleanField(_('Send to primary authors'), widget=SwitchWidget())
     include_coauthors = BooleanField(_('Send to co-authors'), widget=SwitchWidget())
     cc_addresses = EmailListField(_("CC"), description=_("Additional CC e-mail addresses (one per line)"))
     subject = StringField(_("Subject"), [DataRequired()])
@@ -335,7 +336,7 @@ class EditEmailTemplateTextForm(IndicoForm):
 class CreateEmailTemplateForm(EditEmailTemplateRuleForm):
     """Form for adding a new e-mail template."""
 
-    default_tpl = SelectField(_('Default template'), [DataRequired()], choices=[
+    default_tpl = SelectField(_('Email template'), [DataRequired()], choices=[
         ('submit', _('Submit')),
         ('accept', _('Accept')),
         ('reject', _('Reject')),
--- a/indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
+++ b/indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
@@ -1,86 +1,160 @@
 {% from 'message_box.html' import message_box %}
 
-{% macro render_notification_list(event) -%}
+{% macro render_notification_list(event, tracks, states, contrib_types, email_tpls) -%}
     {% if not event.abstract_email_templates %}
         {% call message_box('info', classes='no-rules-placeholder') %}
-            {% trans %}There are currently no e-mail templates{% endtrans%}
+            {% trans %}There are currently no e-mail templates{% endtrans %}
         {% endcall %}
     {% else %}
-        <ul class="email-template-list" data-url="{{ url_for('.email_tpl_sort', event) }}">
-            {% for tpl in event.abstract_email_templates|sort(attribute='position') %}
-                <li class="email-template i-box" data-id="{{ tpl.id }}">
-                    <div class="ui-i-box-sortable-handle"></div>
-                    <div class="i-box-header">
-                        <span class="i-box-title">{{ tpl.title }}</span>
-                        <div class="i-box-buttons toolbar right">
-                            <div class="group">
-                                <a class="i-button icon-eye"
-                                   data-href="{{ url_for('.email_tpl_preview', tpl) }}"
-                                   data-method="GET"
-                                   data-title="{% trans %}E-mail preview{% endtrans %}"
-                                   data-ajax-dialog>
-                                   {% trans %}Preview{% endtrans %}
-                                </a>
-                                <a class="i-button icon-edit arrow js-edit-tpl-dropdown"
-                                   data-toggle="dropdown">{% trans %}Edit{% endtrans %}</a>
-                                <ul class="dropdown">
-                                    <li>
-                                        <a data-href="{{ url_for('.email_tpl_rule_edit', tpl) }}"
-                                           title="{% trans %}Edit ruleset{% endtrans %}"
-                                           data-title="{% trans %}Edit ruleset{% endtrans %}"
+        <div class="email-template-wrapper">
+            <div class="email-templates">
+                <ul class="i-box-group vert"
+                    data-url="{{ url_for('.email_tpl_sort', event) }}">
+                    {% for tpl in event.abstract_email_templates|sort(attribute='position') %}
+                        <li class="i-box" data-id="{{ tpl.id }}">
+                            <div class="ui-i-box-sortable-handle"></div>
+                            <div class="i-box-header">
+                                <span class="i-box-title">{{ tpl.title }}</span>
+                                <div class="i-box-buttons toolbar right">
+                                    <div class="group">
+                                        <a class="i-button icon-eye"
+                                           data-href="{{ url_for('.email_tpl_preview', tpl) }}"
+                                           data-method="GET"
+                                           data-title="{% trans %}E-mail preview{% endtrans %}"
                                            data-ajax-dialog>
-                                           {% trans %}Ruleset{% endtrans %}
+                                            {% trans %}Preview{% endtrans %}
                                         </a>
-                                    </li>
-                                    <li>
-                                        <a data-href="{{ url_for('.email_tpl_text_edit', tpl) }}"
-                                           title="{% trans %}Edit e-mail text{% endtrans %}"
-                                           data-title="{% trans %}Edit e-mail text{% endtrans %}"
-                                           data-ajax-dialog>
-                                           {% trans %}E-mail text{% endtrans %}
+                                        <a class="i-button icon-edit arrow js-edit-tpl-dropdown"
+                                           data-toggle="dropdown">
+                                            {%- trans %}Edit{% endtrans -%}
                                         </a>
-                                    </li>
+                                        <ul class="dropdown">
+                                            <li>
+                                                <a data-href="{{ url_for('.email_tpl_rule_edit', tpl) }}"
+                                                   title="{% trans %}Edit ruleset{% endtrans %}"
+                                                   data-title="{% trans %}Edit ruleset{% endtrans %}"
+                                                   data-update="#email-template-manager"
+                                                   data-ajax-dialog>
+                                                   {% trans %}Ruleset{% endtrans %}
+                                                </a>
+                                            </li>
+                                            <li>
+                                                <a data-href="{{ url_for('.email_tpl_text_edit', tpl) }}"
+                                                   title="{% trans %}Edit e-mail text{% endtrans %}"
+                                                   data-title="{% trans %}Edit e-mail text{% endtrans %}"
+                                                   data-update="#email-template-manager"
+                                                   data-ajax-dialog>
+                                                   {% trans %}E-mail text{% endtrans %}
+                                                </a>
+                                            </li>
+                                        </ul>
+                                        <a href="#" class="i-button js-delete-item icon-remove icon-only"
+                                           title="{% trans %}Delete template{% endtrans %}"
+                                           data-qtip-style="danger"
+                                           data-href="{{ url_for('.email_tpl_delete', tpl) }}"
+                                           data-method="DELETE"
+                                           data-confirm="{% trans %}Are you sure you want to delete this template?{% endtrans %}"
+                                           data-title="{% trans %}Delete template{% endtrans %}"
+                                           data-update="#email-template-manager">
+                                        </a>
+                                    </div>
+                                </div>
+                            </div>
+                            <div class="i-box-content">
+                                <div class="right">
+                                    <a href="#" class="email-preview-btn"
+                                       id="email-preview-btn-{{ tpl.id }}"
+                                       data-id="{{ tpl.id }}">
+                                        {%- trans %}Show preview{% endtrans -%}
+                                    </a>
+                                </div>
+                                <div>
+                                    {%- trans %}This email will be sent when the abstract is:{% endtrans -%}
+                                </div>
+                                <ul>
+                                    {% for rule in tpl.rules if rule %}
+                                        <li class="notification-rule">
+                                            <strong>{{ states[rule.state[0]] }}</strong>
+                                            {% if 'track' in rule %}
+                                                {% if not rule.track %}
+                                                    {%- trans -%}
+                                                       in <strong>no track</strong>
+                                                    {%- endtrans -%}
+                                                {% elif rule.track[0] == '*' %}
+                                                    {%- trans -%}
+                                                       in <strong>any track</strong>
+                                                    {%- endtrans -%}
+                                                {% else %}
+                                                    {%- trans %}in{% endtrans -%}
+                                                    <strong>
+                                                        {{ tracks[rule.track[0]].title }}
+                                                    </strong>
+                                                {% endif %}
+                                            {% endif %}
+                                            {% if 'contribution_type' in rule %}
+                                                    {% if not rule.contribution_type %}
+                                                        {%- trans -%}
+                                                            as <strong>no type</strong>
+                                                        {%- endtrans -%}
+                                                    {% elif rule.contribution_type[0] == '*' %}
+                                                        {%- trans -%}
+                                                            as <strong>any type</strong>
+                                                        {%- endtrans -%}
+                                                    {% else %}
+                                                        {%- trans %}as{% endtrans -%}
+                                                        <strong>
+                                                            {{ contrib_types[rule.contribution_type[0]].name }}
+                                                        </strong>
+                                                    {% endif %}
+                                            {% endif %}
+                                            {% if not loop.last %}
+                                                {%- trans %}or{% endtrans -%}
+                                            {% endif %}
+                                        </li>
+                                    {% endfor %}
                                 </ul>
-                                <a href="#" class="i-button js-delete-item icon-remove icon-only"
-                                   title="{% trans %}Delete template{% endtrans %}"
-                                   data-qtip-style="danger"
-                                   data-href="{{ url_for('.email_tpl_delete', tpl) }}"
-                                   data-method="DELETE"
-                                   data-confirm="{% trans %}Are you sure you want to delete this template?{% endtrans %}"
-                                   data-title="{% trans %}Delete template{% endtrans %}"
-                                   data-update="#email-template-manager">
-                                </a>
-                        </div>
-                    </div>
-                </li>
-            {% endfor %}
-        </ul>
+                                <div class="email-preview"
+                                     id="email-preview-{{ tpl.id }}"
+                                     data-id="{{ tpl.id }}">
+                                    <div class="titled-rule">{% trans %}Email preview{% endtrans %}</div>
+                                    <div class="i-box-footer">
+                                        <div class="preformatted email-subject"><strong>{{ email_tpls[tpl.id].get_subject() }}</strong></div>
+                                        <div class="preformatted email-body">{{ email_tpls[tpl.id].get_body() }}</div>
+                                    </div>
+                                </div>
+                            </div>
+                        </li>
+                    {% endfor %}
+                </ul>
+            </div>
+        </div>
     {% endif %}
 
     <div class="toolbar">
         <div class="group right">
             <a href="#" class="i-button icon-plus highlight"
                data-href="{{ url_for('.email_tpl_add', event) }}"
-               data-title="{% trans %}Add a new template{% endtrans %}"
-               data-ajax-dialog
-               data-update="#email-template-manager">
+               data-title="{% trans %}Add a new notification ruleset{% endtrans %}"
+               data-update="#email-template-manager"
+               data-ajax-dialog>
                 {% trans %}Add new one{% endtrans %}
             </a>
         </div>
     </div>
     <script>
         $('.js-edit-tpl-dropdown').parent().dropdown();
-        $('.email-template-list').sortable({
+        $('.email-templates > ul').sortable({
             axis: 'y',
             containment: 'parent',
             cursor: 'move',
             distance: 2,
             handle: '.ui-i-box-sortable-handle',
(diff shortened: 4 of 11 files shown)


indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -33,7 +36,14 @@ class RHEmailTemplateList(RHManageAbstractsBase):
     """Display list of e-mail templates."""
 
     def _process(self):
-        return jsonify_template('events/abstracts/management/notification_tpl_list.html', event=self.event_new)
+        self.commit = False
+        track_dict, abstract_states, contr_type_dict, email_tpl_dict = _get_rules_fields(self.event_new)
+        return jsonify_template('events/abstracts/management/notification_tpl_list.html',
+                                event=self.event_new,
```
ThiefMaster: no need to put every argument in its own line

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -87,7 +107,12 @@ def _process(self):
         if form.validate_on_submit():
             form.populate_obj(self.email_tpl)
             tpl = get_template_module('events/abstracts/management/_notification_tpl_list.html')
-            return jsonify_data(html=tpl.render_notification_list(self.event_new))
+            track_dict, abstract_states, contr_type_dict, email_tpl_dict = _get_rules_fields(self.event_new)
+            return jsonify_data(html=tpl.render_notification_list(self.event_new,
```
ThiefMaster: you use the same args a few times. why not add a helper function that takes care of this?

`def _render_notification_list(event): ...`

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -33,7 +36,14 @@ class RHEmailTemplateList(RHManageAbstractsBase):
     """Display list of e-mail templates."""
 
     def _process(self):
-        return jsonify_template('events/abstracts/management/notification_tpl_list.html', event=self.event_new)
+        self.commit = False
+        track_dict, abstract_states, contr_type_dict, email_tpl_dict = _get_rules_fields(self.event_new)
```
ThiefMaster: we usually call it `contrib_type`, not `contr_type`

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -108,15 +133,24 @@ class RHDeleteEmailTemplate(RHEditEmailTemplateBase):
...
+                                                              contr_type=contr_type_dict,
+                                                              email_tpl=email_tpl_dict), flash=False)
+
+
+def _get_rules_fields(event_new):
+    """ Retrieve needed rules fields: Track, Abstract state and Contribution type. """
+    track_dict = {track.id: track for track in Track.query.with_parent(event_new)}
+    abstract_states = AbstractState.__titles__
+    contr_type_dict = {contr_type.id: contr_type
+                       for contr_type in ContributionType.query.with_parent(event_new)}
```
ThiefMaster: `ct`; loop variables can be short

ThiefMaster: Why not iterate over `event_new.contribution_types`?

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -108,15 +133,24 @@ class RHDeleteEmailTemplate(RHEditEmailTemplateBase):
...
+                                                              states=abstract_states,
+                                                              contr_type=contr_type_dict,
+                                                              email_tpl=email_tpl_dict), flash=False)
+
+
+def _get_rules_fields(event_new):
+    """ Retrieve needed rules fields: Track, Abstract state and Contribution type. """
+    track_dict = {track.id: track for track in Track.query.with_parent(event_new)}
+    abstract_states = AbstractState.__titles__
+    contr_type_dict = {contr_type.id: contr_type
```
ThiefMaster: `contrib_type_dict`

indico/modules/events/abstracts/forms.py
```diff
@@ -192,7 +192,8 @@ class EditEmailTemplateRuleForm(IndicoForm):
     title = StringField(_("Title"), [DataRequired()])
     rules = EmailRuleListField(_("Rules"), [DataRequired()])
     stop_on_match = BooleanField(_("Stop on match"), [DataRequired()], widget=SwitchWidget(), default=True,
-                                 description=_("Do not evaluate any other rules once this one matches."))
+                                 description=_("If any of the rules from this email template matches, do not "
```
ThiefMaster: "match", it's plural now

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,6 +1,6 @@
 {% from 'message_box.html' import message_box %}
 
-{% macro render_notification_list(event) -%}
+{% macro render_notification_list(event, track, states, contr_type, email_tpl) -%}
```
ThiefMaster: `contrib_type`

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -50,6 +43,40 @@
...
                                    data-update="#email-template-manager">
                                 </a>
+                            </div>
+                        </div>
+                    </div>
+                    <div class="i-box-content">
+                    {% trans %}This email will be sent when the abstract is:{% endtrans %}
+                        <ul>
+                            {% for rule in tpl.rules %}
+                                {% if rule %}
```
ThiefMaster: Can this ever be falsy? Anyway, you can include this in the loop:

    {% for rule in tpl.rules if rule %}

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -50,6 +43,40 @@
...
+                                            {% else %}
+                                                {{ 'any track' if rule['track'][0] == '*' else 'no track' }}
+                                            {% endif %}
+                                        {% endif %}
+                                        {% if 'contribution_type' in rule %}
+                                            <strong>as</strong>
+                                            {% if rule['contribution_type'][0] is number %}
+                                                {{ contr_type[rule['contribution_type'][0]].name }}
+                                            {% else %}
+                                                {{ 'any type' if rule['contribution_type'][0] == '*' else 'no type' }}
```
ThiefMaster: why not elif? mixing normal if/else with inline if/else is somewhat ugly

nop33:
To make use of my commits from #2644, you can use `cherry-pick`. First you need to make sure you have my remote. Check with `git remote -v` to see if you have my repository listed. If not, you can do
`git remote add ilias https://github.com/nop33/indico.git`.
Once that's done, you can `git fetch ilias` to get my latest commits and branches. Then you can use `cherry-pick` to pick my commits, like this:
`git cherry-pick [SHA hash]`
So for example, to pick the commit https://github.com/indico/indico/pull/2644/commits/d16b2beb9e6ce86225ad755d37a0673417aa0b0e you can do
`git cherry-pick d16b2beb9e6ce86225ad755d37a0673417aa0b0e` or just `git cherry-pick d16b2be`.

nop33:
Another thing you could have done is to rebase your branch on top of mine
```
git fetch ilias
git rebase ilias/fix/email-notif-layout
```
so that your commits go on top of my commits.

nop33:
Lastly, if you want to add more code to a commit (either because you forgot something or for any other reason), you can use:
```
git add -p
git commit --fixup=[SHA]
git rebase -i SHA^ --autosquash
```

If you want to add code to the latest commit, the process is a bit simpler:
```
git add -p
git commit --amend --no-edit
```

indico/htdocs/sass/modules/abstracts/_email_templates.scss
```diff
@@ -59,12 +59,44 @@
...
     padding-left: 0;
     width: 800px;
+    max-height: 800px;
+    overflow: auto !important;
 
     .email-template {
         margin-bottom: 10px;
+
+        .notification-rule:not(:last-child):after {
+            content: ' or';
```
nop33: This won't be able to be translated...

indico/htdocs/sass/modules/abstracts/_email_templates.scss
```diff
@@ -59,12 +59,44 @@
...
+
+#form-group-rules {
+    margin-top: 0;
+
+    .form-field {
+        width: 100%;
+    }
+
+    .rule-condition {
+        &[data-condition="track"] {
```
nop33: an with the rule above:
```scss
.form-field,
.rule-condition[data-condition="track"] select {
    width: 100%;
}
```

indico/modules/events/abstracts/controllers/email_templates.py
nop33: True, just keep putting arguments in the same line until it reaches the barrier of 120 chars, then go to a new line.

indico/modules/events/abstracts/forms.py
nop33: It's [not incorrect](https://www.quora.com/If-I-begin-a-sentence-with-If-any-one-of-you-should-the-following-linking-verb-be-is-or-are) to say `matches`, but in this particular case I would agree that `match` fits better

indico/htdocs/sass/modules/abstracts/_email_templates.scss
OmeGak: We should avoid using CSS `content` to add text.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -50,6 +43,40 @@
                                    data-title="{% trans %}Delete template{% endtrans %}"
                                    data-update="#email-template-manager">
                                 </a>
+                            </div>
+                        </div>
+                    </div>
+                    <div class="i-box-content">
+                    {% trans %}This email will be sent when the abstract is:{% endtrans %}
```
nop33: I would wrap it in a `<div>` (as I actually did in https://github.com/indico/indico/pull/2644/commits/3e48d57e8a92e05741cc78b9b27382a15e449f05)

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
nop33: I would love to be able to do this in python as well :stuck_out_tongue_closed_eyes:

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -50,6 +43,40 @@
...
+                                        {% if 'track' in rule %}
+                                            <strong>in</strong>
+                                            {% if rule['track'][0] is number %}
+                                                {{ track[rule['track'][0]].title }}
+                                            {% else %}
+                                                {{ 'any track' if rule['track'][0] == '*' else 'no track' }}
+                                            {% endif %}
+                                        {% endif %}
+                                        {% if 'contribution_type' in rule %}
+                                            <strong>as</strong>
```
nop33: `{% trans %}`

indico/modules/events/abstracts/templates/management/notification_tpl_form.html
```diff
@@ -1,10 +1,9 @@
 {% from 'forms/_form.html' import form_header, form_footer, form_rows, form_row, form_fieldset %}
 
 {{ form_header(form) }}
-    {{ form_row(form.title) }}
+    {{ form_rows(form, fields=('title', 'stop_on_match')) }}
```
nop33: Since the form has only 3 fields, I would use here
``` jinja
{{ form_rows(form, skip=('rules',)) }}
```

indico/htdocs/sass/modules/abstracts/_email_templates.scss
nop33: This might help you:
http://jinja.pocoo.org/docs/dev/templates/#for

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -108,15 +133,24 @@ class RHDeleteEmailTemplate(RHEditEmailTemplateBase):
...
+                       for contr_type in ContributionType.query.with_parent(event_new)}
+    email_tpl_dict = {email_template.id: email_template
+                      for email_template in event_new.abstract_email_templates}
+    abstract = create_mock_abstract(event_new)
+    for k, v in email_tpl_dict.iteritems():
+        email_tpl_dict[k] = get_abstract_notification_tpl_module(v, abstract)
+
+    return track_dict, abstract_states, contr_type_dict, email_tpl_dict
```
OmeGak: Why not returning a dictionary? It seems you build it only for it to be passed to `tpl.render_notification_list`. That way you could simply do `tpl.render_notification_list(**_get_rules_fields(self.event_new))`

OmeGak: I would even avoid this function altoguether and simply compose the dictionary in `_render_notification_list`, as @ThiefMaster proposed.

indico/modules/events/abstracts/forms.py
```diff
@@ -192,7 +192,8 @@ class EditEmailTemplateRuleForm(IndicoForm):
...
+                                               "send emails from any following email template."))
```
OmeGak: > ... send **notifications** from ...

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -108,15 +133,24 @@ class RHDeleteEmailTemplate(RHEditEmailTemplateBase):
...
+        track_dict, abstract_states, contr_type_dict, email_tpl_dict = _get_rules_fields(self.event_new)
+        return jsonify_data(html=tpl.render_notification_list(self.event_new,
+                                                              track=track_dict,
+                                                              states=abstract_states,
+                                                              contr_type=contr_type_dict,
+                                                              email_tpl=email_tpl_dict), flash=False)
+
+
+def _get_rules_fields(event_new):
+    """ Retrieve needed rules fields: Track, Abstract state and Contribution type. """
```
OmeGak: No need to state what the code does. I would keep just `Retrieve needed rules fields` if any docstring at all.

OmeGak: Too much whiespace at the beginning of the docstring, anyway.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
OmeGak: It's not a single track but a dictionary of tracks, right? I would name it `tracks`, if that's the case. The same applies for `contrib_type` and `email_tpl`, I believe.

indico/modules/events/abstracts/controllers/email_templates.py
OmeGak: Not enough indentation.

OmeGak: Not enough indentation.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
OmeGak: Not enough indentation.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
mvidalgarcia: When you add a new rule, the widget allows you to save it with "Select an option" and that generates and empty object `{}`. That's the reason why I added that check.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
mvidalgarcia: You mean this?
```python
{% if rule['contribution_type'][0] is number %}
    {{ contrib_type[rule['contribution_type'][0]].name }}
{% elif rule['contribution_type'][0] == '*' %}
    {{ 'any type' }}
{% else %}
    {{ 'no type' }}
{% endif %}
```

ThiefMaster: Yes. And I'd do the any/no type checks first and just use an `else` block for the last one (assuming you can easily detect "no type" without checking whether it's not a number, e.g. `is none`)

nop33: First of all, doing `{{ 'any type' }}` is the same as writing `any type`.
Secondly, what he means is to add `{% trans %}` around the translatable strings.

ThiefMaster: Oh yeah, didn't notice those ugly strings ;p They should indeed be translated.

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -33,7 +36,14 @@ class RHEmailTemplateList(RHManageAbstractsBase):
     """Display list of e-mail templates."""
 
     def _process(self):
-        return jsonify_template('events/abstracts/management/notification_tpl_list.html', event=self.event_new)
+        self.commit = False
```
ThiefMaster: Why is that needed here? `create_mock_abstract` doesn't create any real objects anymore so there won't be anything committed that shouldn't be committed anyway.

nop33:
What do you guys think about this:
image

image


I don't particularly like the "Open/Close" trigger, but what I want to show is the preview email into the `.i-box-footer`. This way the list by default can have all items collapsed and only when necessary the user can preview the email. It will avoid in most of the cases the usage of scrollbars. The information in the title and content of the ibox are enough to identify the email template, so the email preview shouldn't (imo) be visible by default.

ThiefMaster:
Why not keep a button on top that opens the preview? If it's closed I wouldn't even show the line in that case.

OmeGak:
I would definitely not make a `titled-rule` actionable. I would go for a simple actionable text *Show preview*/*Hide preview*. Not sure if I would put it next to the buttons or as the i-box footer.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
mvidalgarcia: Is it possible to add i18n to the state names ("Accepted", "Submitted", etc.)?
```
{{ states[rule['state'][0]] }}
```

nop33: They are already translated

indico/modules/events/abstracts/templates/management/notification_tpl_form.html
mvidalgarcia: `{{ form_rows(form, skip=('rules','default_tpl')) }}` works fine.  

Know this shouldn't it be more appropriate using `{{ form_rows(form, fields=('title','stop_on_match')) }}`?

nop33:
@mvidalgarcia how about adding the placeholder box, similar to the one in the registration form setup (`/manage/registration/XXXXX/form`) ? I am talking about the gray box that appears when dragging one of the registration form sections.

indico/modules/events/abstracts/templates/management/notification_tpl_form.html
nop33: Why do we need the field "Default template" to be at the bottom and not together with the "Title" and "Stop on match"?

mvidalgarcia: I don't know, should I move it up? And what about the "E-mail template header"?

mvidalgarcia:
Current look
image

indico/htdocs/sass/modules/abstracts/_email_templates.scss
```diff
@@ -55,16 +55,59 @@
...
+
+                .email-template {
+                    margin-bottom: 10px;
+
+                    .email-preview {
+                        display: none;
+
+                        .email-subject {
+                            padding-bottom: 10px;
+                        }
```
ThiefMaster: do we really need that much nesting? you don't need to replicate the DOM's nesting in CSS unless things would be ambiguous otherwise

indico/modules/events/abstracts/templates/management/notification_tpl_form.html
nop33: That's what I am saying, I don't see a point of having the "E-mail template" fieldset. cc @OmeGak, @ThiefMaster, @pferreir ?

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -107,16 +105,19 @@ class RHDeleteEmailTemplate(RHEditEmailTemplateBase):
...
         db.session.delete(self.email_tpl)
-        tpl = get_template_module('events/abstracts/management/_notification_tpl_list.html')
-        return jsonify_data(html=tpl.render_notification_list(self.event_new), flash=False)
+        return _render_notification_list(self.event_new, flash=False)
 
 
-class RHPreviewEmailTemplate(RHEditEmailTemplateBase):
-    """Preview an e-mail template."""
+def _get_rules_fields(event_new):
+    """Retrieve needed rules fields."""
```
ThiefMaster: No need to put docstrings on local/underscored functions

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -107,16 +105,19 @@ class RHDeleteEmailTemplate(RHEditEmailTemplateBase):
...
+    abstract = create_mock_abstract(event_new)
+    email_tpl_dict = {et.id: get_abstract_notification_tpl_module(et, abstract)
+                      for et in event_new.abstract_email_templates}
+    return {'tracks': {track.id: track for track in Track.query.with_parent(event_new)},
```
ThiefMaster: `for track in event_new.tracks`

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,60 +1,112 @@
...
+                                                   data-update="#email-template-manager">
+                                                   {% trans %}Ruleset{% endtrans %}
+                                                </a>
+                                            </li>
+                                            <li>
+                                                <a data-href="{{ url_for('.email_tpl_text_edit', tpl) }}"
+                                                   title="{% trans %}Edit e-mail text{% endtrans %}"
+                                                   data-title="{% trans %}Edit e-mail text{% endtrans %}"
+                                                   data-ajax-dialog
+                                                   data-update="#email-template-manager">
```
ThiefMaster: move up one line (it's nicer to have the value-less attributes at the end)

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,60 +1,112 @@
...
+                                    {% trans %}This email will be sent when the abstract is:{% endtrans %}
+                                </div>
+                                <ul>
+                                    {% for rule in tpl.rules if rule %}
+                                        <li class="notification-rule">
+                                            <strong>{{ states[rule['state'][0]] }}</strong>
+                                            {% if 'track' in rule %}
+                                                {% trans %}in{% endtrans %}
+                                                <strong>
+                                                    {% if rule['track'][0] == '*' %}
```
ThiefMaster: `rule.track[0]`?

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,60 +1,112 @@
...
+                                                        {% trans %}any track{% endtrans %}
+                                                    {% elif rule['track'][0] is undefined %}
```
ThiefMaster: are you sure it's not `rule.track` that's already undefined? in that case i'd move this to the first `if` and only check `if not rule.track`

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -95,5 +148,25 @@
                 });
             }
         });
+
+        $('.email-preview-btn').on('click', function() {
+            var id = $(this).data('id');
+            var $previewBtn = $('.email-preview-btn[data-id='+id+']');
+            if ($previewBtn.data('current-text') === 'show') {
```
ThiefMaster: I'd rather use `.data('visible')` and set it to a boolean. Since undefined is falsy you won't even need to clutter the HTML with a default value then.

indico/htdocs/sass/modules/abstracts/_email_templates.scss
```diff
@@ -55,16 +55,59 @@
...
-        margin-bottom: 10px;
+                        .email-body {
+                            color: $light-black;
+                        }
+                    }
+                }
+                .regform-section-sortable-placeholder {
+                    @extend .i-box;
+                    @include border-all($gray, dashed);
+                    @include box-shadow(none);
```
nop33: Empty line bellow.
https://github.com/indico/indico/wiki/SASS-Guidelines

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,60 +1,112 @@
...
-                                <a class="i-button icon-edit arrow js-edit-tpl-dropdown"
-                                   data-toggle="dropdown">{% trans %}Edit{% endtrans %}</a>
-                                <ul class="dropdown">
-                                    <li>
-                                        <a data-href="{{ url_for('.email_tpl_rule_edit', tpl) }}"
-                                           title="{% trans %}Edit ruleset{% endtrans %}"
-                                           data-title="{% trans %}Edit ruleset{% endtrans %}"
-                                           data-ajax-dialog>
-                                           {% trans %}Ruleset{% endtrans %}
+        <div class="email-template-container">
```
nop33: The semantic of the classes can be improved. Here you have `email-template-container` which suggests that inside I'll find an `email-template`. However, I find the list of email templates. Also, the `email-template-notifications` doesn't make much sense, the feature is Email Notification Templates (or Email Templates or Notification Templates). Bringing `template` before the `notifications` is a bit confusing.
I would suggest the following structure:
```
div.email-templates-wrapper
    div.email-templates
        ul
            li.email-template
```
instead of 
```
div.email-template-container
    div.email-template-notifications
        ul.email-template-list
            li.email-template
```
The `ul` can easily be selected with `$('.email-templates ul`), no need to add a class imo

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,60 +1,112 @@
...
+                    {% for tpl in event.abstract_email_templates|sort(attribute='position') %}
+                        <li class="email-template i-box" data-id="{{ tpl.id }}">
+                            <div class="ui-i-box-sortable-handle"></div>
+                            <div class="i-box-header">
+                                <span class="i-box-title">{{ tpl.title }}</span>
+                                <div class="i-box-buttons toolbar right">
+                                    <a class="i-button borderless email-preview-btn"
+                                       data-id="{{ tpl.id }}"
+                                       data-current-text="hide">
+                                        {% trans %}Show preview{% endtrans %}
```
nop33: `{%- trans %}Show preview{% endtrans -%}`

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,60 +1,112 @@
...
+                                    </a>
+                                    <div class="group">
+                                        <a class="i-button icon-edit arrow js-edit-tpl-dropdown"
+                                           data-toggle="dropdown">{% trans %}Edit{% endtrans %}</a>
```
nop33: Keep consistency as above and bring the `Edit` in a new line.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,60 +1,112 @@
...
+                                        <ul class="dropdown">
+                                            <li>
+                                                <a data-href="{{ url_for('.email_tpl_rule_edit', tpl) }}"
+                                                   title="{% trans %}Edit ruleset{% endtrans %}"
+                                                   data-title="{% trans %}Edit ruleset{% endtrans %}"
+                                                   data-ajax-dialog
```
nop33: As @ThiefMaster said, bring this in the last line

indico/modules/events/abstracts/templates/management/notification_tpl_form.html
nop33: ping @pferreir

indico/htdocs/sass/modules/abstracts/_email_templates.scss
OmeGak: Probably, keeping most of this at same level of nesting of `.email-template-container` should be fine.

indico/htdocs/sass/modules/abstracts/_email_templates.scss
```diff
@@ -55,16 +55,59 @@
...
+                    background-color: $light-gray;
+                    margin: 25px 0;
+                }
+
+                /* this padding is needed so that dragging after
+                   the last element works properly. */
+                padding-bottom: 15px;
```
OmeGak: Is this needed even with the sortable placeholder in place? Seems like a hack to me.

indico/htdocs/sass/modules/abstracts/_email_templates.scss
```diff
@@ -55,16 +55,59 @@
...
+            max-height: 800px;
+            overflow: auto;
+
+            .email-template-list {
+                list-style-type: none;
+                padding-left: 0;
+                padding-right: 20px;
+
+                .email-template {
+                    margin-bottom: 10px;
```
OmeGak: We don't need to add margin bottom to each `i-box`. Simply add the classes `.i-box-group.vert` to the parent element and it will take care of the spacing between them.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,60 +1,112 @@
...
-                                        <a data-href="{{ url_for('.email_tpl_rule_edit', tpl) }}"
-                                           title="{% trans %}Edit ruleset{% endtrans %}"
-                                           data-title="{% trans %}Edit ruleset{% endtrans %}"
-                                           data-ajax-dialog>
-                                           {% trans %}Ruleset{% endtrans %}
+        <div class="email-template-container">
+            <div class="email-template-notifications">
+                <ul class="email-template-list" data-url="{{ url_for('.email_tpl_sort', event) }}">
+                    {% for tpl in event.abstract_email_templates|sort(attribute='position') %}
+                        <li class="email-template i-box" data-id="{{ tpl.id }}">
```
OmeGak: We don't need to specify any class here once you do what I suggested in

```scss
.email-template {
    margin-bottom: 10px;
    ...
```

indico/htdocs/sass/modules/abstracts/_email_templates.scss
```diff
@@ -55,16 +55,59 @@
...
+                            color: $light-black;
+                        }
+                    }
+                }
+                .regform-section-sortable-placeholder {
+                    @extend .i-box;
+                    @include border-all($gray, dashed);
+                    @include box-shadow(none);
+                    background-color: $light-gray;
+                    margin: 25px 0;
```
OmeGak: I don't think we want to apply a different margin. It should look like it's a socket for the box, so it should have the same margin. `.i-box-group.vert` should take care of it if you add the `.i-box` class to the placeholder html element.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
mvidalgarcia: :confused: Why is this working? Is a Jinja feature? I thought dicts could only be accessed with `dict['key']`

ThiefMaster: Yes - when using `foo.bar` it first tries `foo.bar` and if that raises an AttributeError it tries `foo['bar']`. If you do `foo['bar']` it does it exactly the other way round.

So for static keys you can always uses the dot notation in Jinja; only if you either need a key that also exists as an attribute (e.g. `somedict['items']`) or with a dynamic value (`somedict[somekey]`) you need the normal bracket notation.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
mvidalgarcia: Few lines before I'm checking `{% if 'track' in rule %}`. Wouldn't it be the same?

IMHO I think there is a bug in the rules widget. When you create a rule and leave the default values for track and contribution type (any track and any contrib. type), instead of returning e.g `{u'track': [*], u'state': [3], u'contribution_type': [*]}`, it just returns `{u'state': [3]}`

ThiefMaster: That's correct behavior. If "anything" is allowed we don't add it to the rules dict at all (since the rule is effectively disabled since it would match anything).

`if 'track' in rule` also succeeds if `rule['track']` is an empty list (which I guess is what you test for when checking if the first element is undefined).

mvidalgarcia:
Current  look

image

("Show/Hide preview" is underlined when hovering)

indico/htdocs/sass/modules/abstracts/_email_templates.scss
```diff
@@ -55,16 +55,65 @@
...
+            }
+
+            .email-body {
+                color: $light-black;
+            }
+        }
+
+        .regform-section-sortable-placeholder {
+            @extend .i-box;
+            @include border-all($gray, dashed);
```
OmeGak: No need to specify the color, just the style.

```scss
@include border-all($style: dashed);
```

indico/htdocs/sass/modules/abstracts/_email_templates.scss
```diff
@@ -55,16 +55,65 @@
...
+            @include box-shadow(none);
+
+            background-color: $light-gray;
+        }
+
+        .email-preview-btn {
+            color: $link;
```
OmeGak: Not a strong opinion here. You need this because the `<a>` tag doesn't an `href="#"` attribute. If you add it you won't need to specify underline on hover either. OTOH, you will need to do `evt.preventDefault()` in the Javascript handling the click.

indico/modules/events/abstracts/forms.py
```diff
@@ -192,7 +192,8 @@ class EditEmailTemplateRuleForm(IndicoForm):
     title = StringField(_("Title"), [DataRequired()])
     rules = EmailRuleListField(_("Rules"), [DataRequired()])
     stop_on_match = BooleanField(_("Stop on match"), [DataRequired()], widget=SwitchWidget(), default=True,
-                                 description=_("Do not evaluate any other rules once this one matches."))
+                                 description=_("If any of the rules from this email template match, do not "
```
OmeGak: What about this? `If this notification is triggered do not send any other after`
/cc @ThiefMaster @pferreir

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,86 +1,145 @@
...
+                                </div>
+                                <div>
+                                    {%- trans %}This email will be sent when the abstract is:{% endtrans -%}
+                                </div>
+                                <ul>
+                                    {% for rule in tpl.rules if rule %}
+                                        <li class="notification-rule">
+                                            <strong>{{ states[rule.state[0]] }}</strong>
+                                            {% if 'track' in rule %}
+                                                {%- trans %}in{% endtrans -%}
```
OmeGak: For better i18n, I would repeat `in` within each one of the tralatable strings below. They would look like

```jinja
{%- trans -%}
    in <strong>no track</strong>
{%- endtrans -%}
```

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,86 +1,145 @@
...
+                                                        {%- trans %}no track{% endtrans -%}
+                                                    {% elif rule.track[0] == '*' %}
+                                                        {%- trans %}any track{% endtrans -%}
+                                                    {% else %}
+                                                        {{ tracks[rule.track[0]].title }}
+                                                    {% endif %}
+                                                </strong>
+                                            {% endif %}
+                                            {% if 'contribution_type' in rule %}
+                                                {%- trans %}as{% endtrans -%}
```
OmeGak: Same as above with `in`.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,86 +1,145 @@
...
+                                                        {%- trans %}no type{% endtrans -%}
+                                                    {% elif rule.contribution_type[0] == '*' %}
+                                                        {%- trans %}any type{% endtrans -%}
+                                                    {% else %}
+                                                        {{ contrib_types[rule.contribution_type[0]].name }}
+                                                    {% endif %}
+                                                </strong>
+                                            {% endif %}
+                                            {% if not loop.last %}
+                                                {%- trans %} or{% endtrans -%}
```
OmeGak: We should avoid leading whitespace in translatable strings. Was this in purpose? The translator will hardly know. If you added it for spacing you should look for other way to do it.

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -95,5 +154,25 @@
...
+                $previewBtn.text($T.gettext('Hide preview'));
+                $('.email-preview[data-id=' + id + ']').slideToggle();
+                $previewBtn.data('visible', true);
+            }
+        });
+
+        $('.ui-i-box-sortable-handle').on('mousedown', function() {
+            $('.email-preview').hide();
+            $('.email-preview-btn').text($T.gettext('Show preview'))
+                    .data('visible', false);
```
OmeGak: I think this fits just fine in one line.

indico/htdocs/sass/modules/abstracts/_email_templates.scss
ThiefMaster: ..which makes sense anyway so i'd go for `href="#"`

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -95,5 +154,25 @@
                 });
             }
         });
+
+        $('.email-preview-btn').on('click', function() {
+            var id = $(this).data('id');
+            var $previewBtn = $('#email-preview-btn-' + id);
+            if ($previewBtn.data('visible')) {
+                $previewBtn.text($T.gettext('Show preview'));
+                $('.email-preview[data-id=' + id + ']').slideToggle();
```
ThiefMaster: what about using a proper ID like you do for the button?

indico/modules/events/abstracts/forms.py
ThiefMaster: that sounds wrong/incomplete

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
ThiefMaster: It's not relevant whitespace since you get a space from the linebreak between endif and if anyway so there's no need for extra spacing

indico/htdocs/sass/modules/abstracts/_email_templates.scss
mvidalgarcia: I'm not getting the link underlined on hover doing that. It just uses [this](https://github.com/indico/indico/blob/devel/indico/htdocs/css/jquery-ui-custom.css#L57) style.

ThiefMaster:
> Make use of @nop33 PR and @ThiefMaster comments 

If easily possible it's usually a good idea to simply amend/fixup existing commits with changes suggested during reviewing. It looks like it's easily possible here.

indico/htdocs/sass/modules/abstracts/_email_templates.scss
```diff
@@ -55,16 +55,73 @@
...
+        }
+
+        .email-preview-btn {
+            padding-right: 5px;
+            color: $link;
+        }
 
-    .email-template {
-        margin-bottom: 10px;
+        .email-preview-btn:hover {
```
OmeGak: Nothing wrong with this selector. It is sometimes clearer to un-nest selectors. Just letting you know, you can also do it like this:

```scss
.email-preview-btn {
    // ...

    &:hover {
        // ...
    }
}
```

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -107,16 +105,33 @@ class RHDeleteEmailTemplate(RHEditEmailTemplateBase):
...
     def _process(self):
         db.session.delete(self.email_tpl)
-        tpl = get_template_module('events/abstracts/management/_notification_tpl_list.html')
-        return jsonify_data(html=tpl.render_notification_list(self.event_new), flash=False)
+        return _render_notification_list(self.event_new, flash=False)
 
 
-class RHPreviewEmailTemplate(RHEditEmailTemplateBase):
-    """Preview an e-mail template."""
+class RHStopOnMatchEmailTemplate(RHEditEmailTemplateBase):
```
OmeGak: `RHStopOnMatchEmailTemplate` -> `RHToggleStopOnMatchEmailTemplate`

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -107,16 +105,33 @@ class RHDeleteEmailTemplate(RHEditEmailTemplateBase):
...
+        if request.json is None:
+            raise BadRequest('Expected JSON payload')
+
+        invalid_fields = request.json.viewkeys() - {'stop_on_match'}
+        if invalid_fields:
+            raise BadRequest("Invalid fields: {}".format(', '.join(invalid_fields)))
+
+        if 'stop_on_match' in request.json:
+            self.email_tpl.stop_on_match = request.json['stop_on_match']
+
```
ThiefMaster: Not returning anything? I'd return `jsonify_data(flash=False)`?

mvidalgarcia: What does it do/mean?

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,86 +1,169 @@
...
+                        <li class="i-box" data-id="{{ tpl.id }}">
+                            <div class="ui-i-box-sortable-handle"></div>
+                            <div class="i-box-header">
+                                <span class="i-box-title">{{ tpl.title }}</span>
+                                <div class="i-box-buttons toolbar right">
+                                    <div class="group">
+                                        <a href="#"
+                                           class="i-button icon-stop icon-only js-toggle-stop-on-match {{ 'danger text-color' if tpl.stop_on_match }}"
+                                           title="{% trans %}Stop processing at this point if triggered{% endtrans %}"
+                                           data-href="{{ url_for('.email_tpl_stop_on_match', tpl) }}"
```
ThiefMaster: `.email_tpl_rest`

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -95,5 +178,47 @@
...
+                url: $this.data('href'),
+                method: $this.data('method'),
+                data: JSON.stringify({stop_on_match: stopOnMatch}),
+                dataType: 'json',
+                contentType: 'application/json',
+                error: handleAjaxError,
+                success: function() {
+                    $this.data('stop-on-match', stopOnMatch);
+                    if (stopOnMatch) {
+                        $this.addClass('danger text-color');
```
ThiefMaster: $this.toggleClass('danger text-color', stopOnMatch);

This will handle both cases so you can get rid of the if/else

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -95,5 +178,47 @@
...
+        });
+
+        $('.ui-i-box-sortable-handle').on('mousedown', function() {
+            $('.email-preview').hide();
+            $('.email-preview-btn').text($T.gettext('Show preview')).data('visible', false);
+        })
+
+        $('.js-toggle-stop-on-match').on('click', function(evt) {
+            var $this = $(this);
+            evt.preventDefault();
```
ThiefMaster: I'd move this one up or one down, but not between multiple `var` declarations

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -95,5 +178,47 @@
...
+                $('#email-preview-' + id).slideToggle();
+                $previewBtn.data('visible', false);
+            } else {
+                $previewBtn.text($T.gettext('Hide preview'));
+                $('#email-preview-' + id ).slideToggle();
+                $previewBtn.data('visible', true);
+            }
+        });
+
+        $('.ui-i-box-sortable-handle').on('mousedown', function() {
```
ThiefMaster: Don't we have a more specific selector? This would select *any* sortable on the page if there was more than one

ThiefMaster: Maybe `#email-template-manager .ui-i-box-sortable-handle`?

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -1,86 +1,169 @@
...
-                                   data-title="{% trans %}Delete template{% endtrans %}"
-                                   data-update="#email-template-manager">
-                                </a>
-                        </div>
-                    </div>
-                </li>
-            {% endfor %}
-        </ul>
+                                <div class="email-preview"
+                                     id="email-preview-{{ tpl.id }}"
```
ThiefMaster: Looks like the whole tag would fit in one line without going over 120 chars width

indico/modules/events/abstracts/controllers/email_templates.py
```diff
@@ -22,18 +22,21 @@
 from indico.modules.events.abstracts.controllers.management import RHManageAbstractsBase
 from indico.modules.events.abstracts.forms import (EditEmailTemplateRuleForm, EditEmailTemplateTextForm,
                                                    CreateEmailTemplateForm)
+from indico.modules.events.abstracts.models.abstracts import AbstractState
 from indico.modules.events.abstracts.models.email_templates import AbstractEmailTemplate
 from indico.modules.events.abstracts.notifications import get_abstract_notification_tpl_module
 from indico.modules.events.abstracts.util import build_default_email_template, create_mock_abstract
 from indico.web.flask.templating import get_template_module
 from indico.web.util import jsonify_data, jsonify_template
+from werkzeug.exceptions import BadRequest
```
ThiefMaster: This is thirdparty, not indico so it should be in the previous block of imports (below the flask one)

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -95,5 +176,43 @@
...
+                $previewBtn.text($T.gettext('Hide preview'));
+                $('#email-preview-' + id ).slideToggle();
+                $previewBtn.data('visible', true);
+            }
+        });
+
+        $('#email-template-manager .ui-i-box-sortable-handle').on('mousedown', function() {
+            $('.email-preview').hide();
+            $('.email-preview-btn').text($T.gettext('Show preview')).data('visible', false);
+        })
```
ThiefMaster: missing semicolon

indico/modules/events/abstracts/templates/management/_notification_tpl_list.html
```diff
@@ -95,5 +176,43 @@
...
+            evt.preventDefault();
+            var $this = $(this);
+            var stopOnMatch = !$this.data('stop-on-match');
+            $.ajax({
+                url: $this.data('href'),
+                method: $this.data('method'),
+                data: JSON.stringify({stop_on_match: stopOnMatch}),
+                dataType: 'json',
+                contentType: 'application/json',
+                error: handleAjaxError,
```
ThiefMaster: should we also show the progress dialog while the request is running?

If yes, add:

    complete: IndicoUI.Dialogs.Util.progress(),

PR #2655, opened 2016-10-19
Improve discoverability of contribution/abstract types

nop33:
image

image

### DIFF
--- a/indico/htdocs/sass/widgets/_sortable_list.scss
+++ b/indico/htdocs/sass/widgets/_sortable_list.scss
@@ -132,5 +132,24 @@
                }
            }
        }
-   }
+    }
+
+    &.tiles {
+        ul {
+            margin: 0;
+        }
+
+        li.i-label {
+            @include border-radius();
+
+            &:not(:last-child) {
+                margin-bottom: 5px;
+            }
+
+            &:not(.placeholder) {
+                @include border-all();
+                background-color: $light-gray;
+            }
+        }
+    }
 }
--- a/indico/modules/events/abstracts/forms.py
+++ b/indico/modules/events/abstracts/forms.py
@@ -414,7 +414,7 @@ class CreateEmailTemplateForm(EditEmailTemplateRuleForm):
 class AbstractForm(IndicoForm):
     title = StringField(_("Title"), [DataRequired()])
     description = IndicoMarkdownField(_('Content'), editor=True, mathjax=True)
-    submitted_contrib_type = QuerySelectField(_("Type"), get_label='name', allow_blank=True,
+    submitted_contrib_type = QuerySelectField(_("Contribution type"), get_label='name', allow_blank=True,
                                               blank_text=_("No type selected"))
     person_links = AbstractPersonLinkListField(_("Authors"), [DataRequired()], default_author_type=AuthorType.primary)
     submission_comment = TextAreaField(_("Comments"))
--- a/indico/modules/events/abstracts/templates/management/overview.html
+++ b/indico/modules/events/abstracts/templates/management/overview.html
@@ -14,9 +14,9 @@
             <div class="icon icon-upload"></div>
             <div class="text">
                 <div class="label">
-                    {% trans %}Submission{% endtrans %}
+                    {%- trans %}Submission{% endtrans -%}
                 </div>
-                {% trans %}Configure the submission dates, the abstract fields, etc.{% endtrans %}
+                {%- trans %}Configure general submission settings and email notification templates.{% endtrans -%}
             </div>
             <div class="toolbar">
                 <div class="group">
@@ -24,15 +24,7 @@
                        data-href="{{ url_for('.email_tpl_list', event) }}"
                        data-title="{% trans %}Configure e-mail notifications{% endtrans %}"
                        data-ajax-dialog>
-                        {% trans %}Notifications{% endtrans %}
-                    </a>
-                </div>
-                <div class="group">
-                    <a href="#" class="i-button icon-list js-dialog-action"
-                       data-title="{% trans %}Manage fields{% endtrans %}"
-                       data-href="{{ url_for('contributions.manage_fields', event) }}"
-                       data-ajax-dialog>
-                        {% trans %}Abstract fields{% endtrans %}
+                        {%- trans %}Notifications{% endtrans -%}
                     </a>
                 </div>
                 <div class="group">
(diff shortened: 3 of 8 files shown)


ThiefMaster:
We don't use the term "abstract type" anywhere; not sure if it's a good fit.

It's was called "Presentation type" in two places (legacy) and "Contribution type" everywhere else.

ThiefMaster:
I think it looks way too intrusive on the contributions page now. It's something you setup once and then don't care about it anymore when managing contributions etc.

nop33:
We might not call them `abstract types` anywhere, but we get many tickets with people looking for the `abstract types`, `presentation types`, `contribution types`... In any case, it's the `type` you choose when you submit an abstract, so personally I would go with `abstract type`. What do the rest think?

nop33:
@ThiefMaster it might be that you configure it once, but if you don't find it at all then you cannot configure it.

OmeGak:
Contribution type all the way, I would say.

nop33:
## Updated
image

ThiefMaster:
Hm.. why link to the paeg and not just show the dialog?

nop33:
So that the people are informed that that's where they need to go to configure the contribution types. If we open the dialog, they will always rely on finding the contribution types in the abstracts page.

ThiefMaster:
To me that feels like making the UX worse just to educate users - and in the end I think it's perfectly fine to configure it from there; I don't think there are that many cases where people use different contribution types while not using abstracts.

pferreir:
I agree with @thiefmaster: contribution types, despite their name, are basically useless outside the abstract reviewing mode.

nop33:
OK then! What's the reason then of having them in the contribution page? The naming?

pferreir:
The name, I believe. Anyway, they're kind of hidden which is good. The main entry point should be the CFA setup page and users who try to guess where they are based on the name should be able to get there too.

pferreir:
Why are we opening this ticket? Has there been a user complaint?

ThiefMaster:
INC1165676

nop33:
It's not the first time I get contact by Lionel or Hakan about this problem. It was a quick fix and I created a PR for it.

pferreir:
Sure, but there are 2 things here:
 1. It may be annoying but from what I see it's not urgent;
 2. We have a new release about to come out that makes that dialog accessible from what should be its main entry point.

That said, does it really make sense to go for a PR that substantially changes the way users perceive this feature?

ThiefMaster:
I'd keep the cog menu in contributions as it is and open the dialog from the abstracts menu...

nop33:
I'll add the dialog in the CFA page.

nop33:
### Update

What about this approach?
1. I added a page description and the text "Settings" next to the cog icon in the contribution page, to help users find the settings for the **Contribution types** and **Custom fields** easier.
2. In the CFA page, I created a new section with the title **Fields and types** where I added the buttons for managing the **Contribution types** and **Abstract fields**.
3. I improved the UI in the `.sortable-list` adding the option to make the items look like `.tiles`.


image

image

image

OmeGak:
Not a strong opinion. It's only that "Fields and types" feels to me like a bit out of place there.

nop33:
I think the title "Fields and types" is helping users discover exactly what they are looking for.

OmeGak:
Yes, I just said that it feels a bit out of place. Dunno how we could make it look better. Maybe not in the same group as `Submission`/`Reviewing`/`BoA`? Standalone like `List of abstracts`?

nop33:
@OmeGak IMO it should be close to the `Submission` (it should actually be part of it, but we cannot fit too many buttons there), so I wouldn't put it in a separate group. It's part of the "Settings" group, whereas the "List of abstracts" is not related to settings. Moving it outside of the group wouldn't make any semantic sense.

indico/htdocs/sass/widgets/_sortable_list.scss
```diff
@@ -132,5 +132,24 @@
                }
            }
        }
-   }
+    }
+
+    &.tiles{
```
ThiefMaster: missing space before `{`

indico/modules/events/abstracts/forms.py
```diff
@@ -413,8 +413,8 @@ class CreateEmailTemplateForm(EditEmailTemplateRuleForm):
 
 class AbstractForm(IndicoForm):
     title = StringField(_("Title"), [DataRequired()])
-    description = IndicoMarkdownField(_('Content'), editor=True, mathjax=True)
-    submitted_contrib_type = QuerySelectField(_("Type"), get_label='name', allow_blank=True,
+    description = IndicoMarkdownField(_('Content'), [DataRequired()], editor=True, mathjax=True)
```
ThiefMaster: no...

indico/modules/events/abstracts/forms.py
```diff
@@ -413,8 +413,8 @@ class CreateEmailTemplateForm(EditEmailTemplateRuleForm):
...
+    submitted_contrib_type = QuerySelectField(_("Contribution type"), get_label='name', allow_blank=True,
```
ThiefMaster: not sure if that rename is so useful. "Type: Oral/Poster" etc is pretty clear already

other opinions?

nop33: I think we made it clear when I opened the pull request that we want to rename it to "Contribution type" because in legacy it was "Presentation type". It also makes the transition from an abstract to a contribution easier for the user.

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -10,16 +10,16 @@
     {%- if event.type == 'conference' -%}
         <a id="contributions-settings-dropdown"
            class="i-button borderless icon-settings arrow js-dropdown"
-           data-toggle="dropdown"></a>
+           data-toggle="dropdown">Settings</a>
```
ThiefMaster: Looks a bit weird to me (also, missing i18n)...

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -36,6 +36,12 @@
     {%- endif -%}
 {% endblock %}
 
+{% block description %}
+    {%- trans -%}
+        Manage the contributions of the event from the list below or configure the <strong>contribution types</strong> and <strong>custom fields</strong> from the settings menu above.
```
ThiefMaster: break line after `</strong>`?

indico/modules/events/abstracts/templates/management/overview.html
```diff
@@ -14,33 +14,53 @@
...
+            <div class="toolbar">
                 <div class="group">
-                    <a href="#" class="i-button icon-settings"
-                       data-href="{{ url_for('.manage_submission_settings', event) }}"
-                       data-title="{% trans %}Configure abstract submission{% endtrans %}"
+                    <a href="#" class="i-button"
+                       data-href="{{ url_for('contributions.manage_types', event) }}"
+                       data-title="{% trans %}Manage contribution types{% endtrans %}"
+                       data-ajax-dialog>
+                        {%- trans %}Contribution types{% endtrans -%}
```
ThiefMaster: Those are the only buttons on the page with no icon now :/

indico/modules/events/contributions/templates/management/contributions.html
nop33: For me it looks better like this than with just the cog icon.

indico/modules/events/abstracts/templates/management/overview.html
nop33: Not the only ones ("Close" call for abstracts also doesn't), but I believe we're overusing icons... Any better idea?

indico/modules/events/abstracts/views.py
```diff
@@ -45,7 +45,8 @@ def getCSSFiles(self):
         return (WPConferenceModifBase.getCSSFiles(self) +
                 self._asset_env['markdown_sass'].urls() +
                 self._asset_env['selectize_css'].urls() +
-                self._asset_env['abstracts_sass'].urls())
+                self._asset_env['abstracts_sass'].urls() +
+                self._asset_env['contributions_sass'].urls())
```
nop33: It's for the "Contribution types" dialog. You really think I should move the code to `eventmanagement.scss`? We also import the file in `WPDisplayAbstractsBase`, not just `WPManageAbstracts`...

ThiefMaster: OK, nevermind then

PR #2656, opened 2016-10-20
Bulk abstract judgment actions

nop33:


indico/modules/events/abstracts/controllers/management.py
```diff
@@ -103,11 +104,47 @@ def _process(self):
...
 
+class RHManageBulkAbstractJudgment(RHManageAbstractsActionsBase):
+    """Perform bulk judgment operations on selected abstracts"""
+
+    def _process(self):
+        abstract_ids = request.form.getlist('abstract_id')
+        form = AbstractsJudgmentForm(event=self.event_new, abstract_id=[a.id for a in self.abstracts])
+        action = request.values.get('action')
+        # The following could also be achived with:
+        # form.judgment.data = action if action in {'accept', 'reject', 'merge'} else None
```
nop33: Do you guys think it's better to do it this way or the way I wrote it below?

ThiefMaster: Cleaner and more concise

nop33: Thing is that `AbstractAction.xxxx.name` is the same as the argument I pass, that's why the above line works fine.

ThiefMaster: `SomeEnum[x].name == x` is always the case with enums

nop33: So how about replacing the next few lines with
```python
allowed_actions = {AbstractAction.accept.name, AbstractAction.reject.name, AbstractAction.merge.name}
form.judgment.data = action if action in allowed_actions else None
```

ThiefMaster: What's the point of it? If we ever renamed the actions we'd have to change it here anyway, so not using the strings provides zero advantage

indico/modules/events/abstracts/forms.py
```diff
@@ -168,15 +182,24 @@ def __init__(self, *args, **kwargs):
...
         super(AbstractJudgmentForm, self).__init__(*args, **kwargs)
-        self.session.query = Session.query.with_parent(self.event).order_by(Session.title)
-        self.accepted_track.query = Track.query.with_parent(self.event).order_by(Track.title)
-        self.accepted_contrib_type.query = (ContributionType.query
-                                                            .with_parent(self.event)
-                                                            .order_by(ContributionType.name))
+
+
+class AbstractsJudgmentForm(AbstractJudgmentFormBase):
+    judgment = HiddenField()
```
ThiefMaster: How about

    judgment = IndicoEnumSelectField(only={AbstractAction.accept, AbstractAction.reject, AbstractAction.merge}, widget=HiddenInput())

nop33: `'IndicoEnumSelectField' object has no attribute '_value'` when I use your suggestion.

ThiefMaster: try this:

```python
class HiddenEnumField(IndicoEnumSelectField):
    widget = HiddenInput()
    def _value(self):
        return self.data.name if self.data is not None else ''
```

nop33: Is it worth creating a new field for this, when it can work as well with the `HiddenField()` and some extra code? :thinking:

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -103,11 +104,39 @@ def _process(self):
...
+            send_notification = abstract_data.pop('send_notifications')
+            contrib_session = abstract_data.pop('contrib_session', None)
+            merge_persons = abstract_data.pop('merge_persons', None)
+            for abstract in self._find_abstracts(self.event_new, abstract_ids):
+                judge_abstract(abstract, abstract_data, judgment=judgment, contrib_session=contrib_session,
+                               merge_persons=merge_persons, send_notification=send_notification)
+            return redirect(url_for('.manage_abstracts', self.event_new))
+        return jsonify_form(form=form, submit=_('Judge'))
+
+    def _find_abstracts(self, event, abstract_ids):
```
ThiefMaster: Why? You already have `self.abstracts` which is set in `RHManageAbstractsActionsBase`

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -103,11 +104,39 @@ def _process(self):
...
+        form.judgment.data = request.values.get('action')
+        if form.process_ajax():
+            return form.ajax_response
+        if form.validate_on_submit():
+            abstract_data = form.data
+            judgment = abstract_data.pop('judgment')
+            send_notification = abstract_data.pop('send_notifications')
+            contrib_session = abstract_data.pop('contrib_session', None)
+            merge_persons = abstract_data.pop('merge_persons', None)
+            for abstract in self._find_abstracts(self.event_new, abstract_ids):
```
ThiefMaster: `for abstract in self.abstracts`

indico/modules/events/abstracts/templates/management/abstract_list.html
```diff
@@ -73,15 +73,32 @@
                     </ul>
                 </div>
                 <div class="group">
-                    <a class="i-button icon-checkmark js-requires-selected-row js-dialog-action accept">
-                        {% trans %}Accept{% endtrans %}
+                    <a class="i-button icon-checkmark accept js-requires-selected-row"
+                       data-href="{{ url_for('.manage_judge_abstracts', event) }}"
+                       data-title="{% trans %}Accept selected abstracts{% endtrans %}"
+                       data-params-selector="#abstract-list input[name=abstract_id]:checked"
+                       data-params='{"action": "{{ actions.accept.name }}"}'
```
ThiefMaster: `"accept"` etc. no need to go through the enum and its `.name`

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -228,7 +257,7 @@ def _process(self):
             return redirect(self.list_generator.get_list_url())
         list_kwargs = self.list_generator.get_list_kwargs()
         return WPManageAbstracts.render_template('management/abstract_list.html', self._conf, event=self.event_new,
-                                                 **list_kwargs)
+                                                 actions=AbstractAction, **list_kwargs)
```
ThiefMaster: not needed after https://github.com/indico/indico/pull/2656/commits/3f76f44956daa52e443e644e8653af847734cba6#r84282337

indico/modules/events/abstracts/forms.py
```diff
@@ -130,11 +132,9 @@ def data(self):
         return data
 
 
-class AbstractJudgmentForm(IndicoForm):
-    """Form for judging an abstract"""
+class AbstractJudgmentFormBase(IndicoForm):
+    """Form base class for abstract(s) judgment operations"""
```
ThiefMaster: I think it's quite clear without the `(s)` too :)

indico/modules/events/abstracts/forms.py
```diff
@@ -168,15 +183,24 @@ def __init__(self, *args, **kwargs):
...
         super(AbstractJudgmentForm, self).__init__(*args, **kwargs)
-        self.session.query = Session.query.with_parent(self.event).order_by(Session.title)
-        self.accepted_track.query = Track.query.with_parent(self.event).order_by(Track.title)
-        self.accepted_contrib_type.query = (ContributionType.query
-                                                            .with_parent(self.event)
-                                                            .order_by(ContributionType.name))
+
+
+class AbstractsJudgmentForm(AbstractJudgmentFormBase):
+    judgment = HiddenEnumField(widget=HiddenInput(), enum=AbstractAction, skip={AbstractAction.change_tracks})
```
ThiefMaster: the widget is not needed anymore. it's already set in the field class

indico/modules/events/abstracts/templates/management/abstract_list.html
OmeGak: Just a clarification. The reason not to go through the `enum` is that if we ever change the class, we would need to change it here as well. OTOH, accessing it as `accept.name` would make our lives much easier in case of such refactoring, as searching just `accept` would return countless unrelated results.

indico/htdocs/sass/modules/_attachments.scss
```diff
@@ -332,18 +329,17 @@
 
 .material-list {
     .folder .attachment {
-        display: inline-block;
-        overflow: hidden;
+        @include ellipsis;
```
OmeGak: If I ever added line break after SASS declarations I stopped long ago. I don't think they add any extra clarity when reading SCSS code, nor they are different enough from the rest of the declarations to put them aside.

indico/htdocs/sass/modules/_contributions.scss
```diff
@@ -25,10 +25,10 @@
 
 @mixin picker-root-element-label() {
     .label {
+        @include ellipsis;
+
         float: left;
         max-width: 80%;
-        overflow: hidden !important;
```
OmeGak: Have you tested that `@include ellipsis;` correctly overrides whatever was affecting `overflow` to require `hidden !important`?

indico/modules/events/abstracts/templates/management/abstract_list.html
ThiefMaster: Yes, if we change the enum member's name/value we'd have to change it here too. But it's so unlikely to happen that I don't think it's worth making code less readable for it. Especially considering how awful it is to use string operations to build JSON.

And anyway, how likely is it that someone would actually think about searching for `accept.name`? ;)

indico/htdocs/sass/modules/_contributions.scss
```diff
@@ -64,9 +64,9 @@
         display: flex;
 
         .label {
+            @include ellipsis;
```
OmeGak: I wonder if the implicit `white-space: nowrap;` that comes with `@include ellipsis;` causes any problem. Did you test?

indico/htdocs/sass/partials/_core.scss
```diff
@@ -104,3 +104,7 @@
     overflow-x: hidden;
     overflow-y: auto;
 }
+
+.truncate-text {
```
OmeGak: As I mentioned already, `.truncate-text` doesn't convey to me that the text is going to be "ellipsisized". I would really go for `.ellipsisize`.

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -103,11 +104,39 @@ def _process(self):
...
+    """Perform bulk judgment operations on selected abstracts"""
+
+    def _process(self):
+        abstract_ids = request.form.getlist('abstract_id')
+        form = AbstractsJudgmentForm(event=self.event_new, abstract_id=[a.id for a in self.abstracts])
+        form.judgment.data = request.values.get('action')
+        if form.process_ajax():
+            return form.ajax_response
+        if form.validate_on_submit():
+            abstract_data = form.data
```
OmeGak: Maybe we could move these `pop` operations to a helper since they are exactly the same both in `RHManageAbstract` and `RHManageBulkAbstractJudgment`.

OmeGak: Would it make sense to add the helper as a member of `AbstractJudgmentFormBase`? /cc @ThiefMaster

indico/modules/events/abstracts/forms.py
```diff
@@ -151,14 +151,29 @@ class AbstractJudgmentForm(IndicoForm):
...
+                                            .with_parent(self.event)
+                                            .order_by(ContributionType.name))
+
+
+class AbstractJudgmentForm(AbstractJudgmentFormBase):
+    """Form for judging an abstract"""
+
+    judgment = IndicoEnumSelectField(_("Judgment"), [DataRequired()], enum=AbstractAction,
+                                     skip={AbstractAction.change_tracks})
+    duplicate_of = AbstractField(_("Duplicate of"),
```
OmeGak: If we allow bulk merge we should also allow bulk mark as dupe.

indico/modules/events/abstracts/controllers/management.py
ThiefMaster: maybe a split_data property returning a tuple with the special values and the remaining data dict?

indico/htdocs/sass/modules/_attachments.scss
nop33: I'll disagree, looking at `scss` code with line breaks after the `@include` and `@extend` feels less messy and more clean imo.

indico/htdocs/sass/modules/_contributions.scss
nop33: The `!important` was useless in this case, so yes, it works fine.

indico/htdocs/sass/partials/_core.scss
nop33: I would still stick with `truncate`.
https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/
What other options are there for truncating a text? You would never truncate it without using `...` at the end, it would look ugly and in any case we would never use a different class to implement this behavior.

indico/modules/events/abstracts/templates/management/abstract_list.html
ThiefMaster: But it's **extremely** unlikely that those enum entries change (especially just their values). and generating JSON using `{"foo" "{{ bar }}"}` is pretty dirty (would fail with any special chars, even though that's not going to happen here) 
And `{{ {'foo': bar}|tojson|forceescape }}` would be even uglier..

indico/modules/events/abstracts/forms.py
```diff
@@ -146,19 +146,45 @@ class AbstractJudgmentForm(IndicoForm):
...
 
     def __init__(self, *args, **kwargs):
+        super(AbstractJudgmentFormBase, self).__init__(*args, **kwargs)
+        self.session.query = Session.query.with_parent(self.event).order_by(Session.title)
+        self.accepted_track.query = Track.query.with_parent(self.event).order_by(Track.title)
+        self.accepted_contrib_type.query = (ContributionType.query
+                                            .with_parent(self.event)
+                                            .order_by(ContributionType.name))
+
+    def _split_form_data(self):
```
ThiefMaster: why underscored? you call it from outside

indico/modules/events/abstracts/forms.py
```diff
@@ -168,15 +194,24 @@ def __init__(self, *args, **kwargs):
...
             kwargs.setdefault('accepted_contrib_type', candidate_contrib_types[0])
         super(AbstractJudgmentForm, self).__init__(*args, **kwargs)
-        self.session.query = Session.query.with_parent(self.event).order_by(Session.title)
-        self.accepted_track.query = Track.query.with_parent(self.event).order_by(Track.title)
-        self.accepted_contrib_type.query = (ContributionType.query
-                                                            .with_parent(self.event)
-                                                            .order_by(ContributionType.name))
+
+
+class AbstractsJudgmentForm(AbstractJudgmentFormBase):
```
ThiefMaster: `BulkJudgmentForm` might be clearer

nop33: I'll rename it to `BulkAbstractJudgmentForm` or leave it as it is, what do you think?

ThiefMaster: fine too, just prefer having something more clear than just AbstractJudgmentForm vs AbstractsJudgmentForm

indico/modules/events/abstracts/blueprint.py
```diff
@@ -68,6 +69,8 @@
 _bp.add_url_rule('/manage/abstracts/person-list', 'person_list', RHAbstractPersonList, methods=('POST',))
 _bp.add_url_rule('/manage/abstracts/attachments', 'download_attachments', RHAbstractsDownloadAttachments,
                  methods=('POST',))
+_bp.add_url_rule('/manage/abstracts/judge', 'manage_judge_abstracts', RHManageBulkAbstractJudgment,
```
ThiefMaster: `RHBulkAbstractJudgment`

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -92,22 +93,34 @@ def _checkParams(self, params):
...
         return WPManageAbstracts.render_template('management/abstract.html', self._conf, abstract=self.abstract,
                                                  form=form)
 
 
+class RHManageBulkAbstractJudgment(RHManageAbstractsActionsBase):
+    """Perform bulk judgment operations on selected abstracts"""
+
+    def _process(self):
+        form = BulkAbstractJudgmentForm(event=self.event_new, abstract_id=[a.id for a in self.abstracts],
+                                     judgment=request.form.get('action'))
```
ThiefMaster: indentation

indico/modules/events/abstracts/forms.py
```diff
@@ -146,19 +146,45 @@ class AbstractJudgmentForm(IndicoForm):
...
 
     def __init__(self, *args, **kwargs):
+        super(AbstractJudgmentFormBase, self).__init__(*args, **kwargs)
+        self.session.query = Session.query.with_parent(self.event).order_by(Session.title)
+        self.accepted_track.query = Track.query.with_parent(self.event).order_by(Track.title)
+        self.accepted_contrib_type.query = (ContributionType.query
+                                            .with_parent(self.event)
+                                            .order_by(ContributionType.name))
+
+    def split_form_data(self):
```
ThiefMaster: `@property def split_dat(self)`?

ThiefMaster:
Do we really want to allow people to judge already-judged abstracts? If not we should disallow it in the backend (e.g. silently skip them) and handle it nicely in the frontend. We could disable the judgment buttons if any judged abstracts are selected or show a warning that some abstracts have already been judged and thus will be skipped.

cc @pferreir

pferreir:
I believe already judged abstracts should be skipped and there should be a message saying they've been ignored. I think this is probably the easiest approach and it won't be too annoying.

pferreir:
IIRC one can accept an abstract for *no track*, *no session* and with *no contribution type*. I think that's currently not handled. Or maybe that's what the default selection means? But in that case it shouldn't be "Choose a contribution type", etc... That could still be the default (but not selectable again) but there should be an "empty" option.

ThiefMaster:
The submit button should also not be disabled until change since in some cases keeping the defaults is perfectly fine.

Also, didn't we once say that we want to default to the contrib type etc. from the abstract's reviews if there's one clear result?

pferreir:
> Also, didn't we once say that we want to default to the contrib type etc. from the abstract's reviews if there's one clear result?

Yep.

OmeGak:
That's already the case. We pre-fill the form with any clear track/contrib_type.

pferreir:
@OmeGak what about having a "no track" option?

OmeGak:
You can leave the select field empty for no track. We could change the placeholder text to be `Choose a track (optional)...`.

nop33:
> Also, didn't we once say that we want to default to the contrib type etc. from the abstract's reviews if there's one clear result?
Do we really want to implement this for the bulk actions as well?

ThiefMaster:
That's actually where it's most useful. Otherwise you have to filter by type/track/etc and accept each group on its own.

nop33:
> That's actually where it's most useful. Otherwise you have to filter by type/track/etc and accept each group on its own.

I am afraid I don't understand. You really want to check whether all selected abstracts from the abstract list have the the same contribution type based on their reviews and then prefill the contribution type field in the dialog form with that? And then, they select 50+ abstracts and accidentally one has a different contrib type and they send a ticket "It doesn't work"?
I personally think it's a bit too far, but if the rest think it makes sense to invest time on it, I can do it.

ThiefMaster:
At least that's what we discussed iirc.. so you can select e.g. all abstracts with a positive review state (ie no ambiguity), bulk-accept and keep some switch in the dialog on that disables the various dropdowns and will use the data from the reviews instead.

nop33:
I would appreciate the feedback from @pferreir and @OmeGak before moving on implementing it.

pferreir:
So, what I have in mind (and from what I read @ThiefMaster agrees), is that if and only if there are non-judged abstracts proposed for different tracks/contribution types, the dialog in question consists of two options:

 * Accept everything as it is when it is consensual (abstracts in conflict are skipped, flash message shown);
 * Select what you want to accept them for (and in this case we could pre-select one of the options if all selected abstracts have it);

This could be achieved using a switch, or radio buttons for that matter.

nop33:
I think this is the optimal approach to the bulk judgments.

nop33:
There's an Asana ticket for this feature: https://app.asana.com/0/161377460377305/200877114028744
We can now merge the PR if there are no other suggestions.

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -92,19 +93,43 @@ def _checkParams(self, params):
...
+    """Perform bulk judgment operations on selected abstracts"""
+
+    def _process(self):
+        form = BulkAbstractJudgmentForm(event=self.event_new, abstract_id=[a.id for a in self.abstracts],
+                                        judgment=request.form.get('action'))
+        if form.process_ajax():
+            return form.ajax_response
+        elif form.validate_on_submit():
+            special_data, remaining_data = form.split_data
+            submitted_abstracts = [abstract for abstract in self.abstracts if abstract.state == AbstractState.submitted]
```
ThiefMaster: `{...}`

indico/modules/events/abstracts/forms.py
```diff
@@ -17,10 +17,12 @@
 from __future__ import unicode_literals
 
 from datetime import time
+from flask import request
```
ThiefMaster: flask isn't stdlib

indico/modules/events/abstracts/forms.py
```diff
@@ -17,10 +17,12 @@
...
 
 from wtforms.ext.sqlalchemy.fields import QuerySelectField
-from wtforms.fields import BooleanField, IntegerField, SelectField, StringField, TextAreaField
+from wtforms.fields import BooleanField, IntegerField, SelectField, StringField, TextAreaField, HiddenField
 from wtforms.validators import NumberRange, Optional, DataRequired, ValidationError, InputRequired
+from wtforms.widgets.core import HiddenInput
```
ThiefMaster: unused import

indico/modules/events/abstracts/forms.py
```diff
@@ -146,19 +146,46 @@ class AbstractJudgmentForm(IndicoForm):
...
+        self.accepted_contrib_type.query = (ContributionType.query
+                                            .with_parent(self.event)
+                                            .order_by(ContributionType.name))
+
+    @property
+    def split_data(self):
+        abstract_data = self.data
+        judgment = abstract_data.pop('judgment')
+        special_data = {
+            'judgment': judgment if judgment else AbstractAction[request.form['judgment']],
```
ThiefMaster: `judgment or AbstractAction[request.form['judgment']]`

Getting it through request.form is not very pretty btw. Doesn't it work if you use `self.judgment.data`?

nop33: In case we judge the abstract from the abstract management page, the judgment is in abstract_data. If we judge it from the abstract list, it's in the request.form

indico/modules/events/abstracts/forms.py
```diff
@@ -146,19 +145,46 @@ class AbstractJudgmentForm(IndicoForm):
...
+        self.accepted_contrib_type.query = (ContributionType.query
+                                            .with_parent(self.event)
+                                            .order_by(ContributionType.name))
+
+    @property
+    def split_data(self):
+        abstract_data = self.data
+        judgment = abstract_data.pop('judgment')
+        special_data = {
+            'judgment': judgment,
```
OmeGak: Why not popping `judgment` here directly?

indico/modules/events/abstracts/forms.py
```diff
@@ -146,19 +145,46 @@ class AbstractJudgmentForm(IndicoForm):
...
+            'send_notification': abstract_data.pop('send_notifications'),
```
OmeGak: Default to `True` in case it's missing.

indico/modules/events/abstracts/forms.py
```diff
@@ -146,19 +145,46 @@ class AbstractJudgmentForm(IndicoForm):
...
+        self.accepted_track.query = Track.query.with_parent(self.event).order_by(Track.title)
+        self.accepted_contrib_type.query = (ContributionType.query
+                                            .with_parent(self.event)
+                                            .order_by(ContributionType.name))
+
+    @property
+    def split_data(self):
+        abstract_data = self.data
+        judgment = abstract_data.pop('judgment')
+        special_data = {
```
OmeGak: It's not really special, just not data you can use to populate an abstract. Not convince by any, but some alternatives: `data`, `judgment_data`, `form_data`, or `other_data`. Any better idea?

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -92,19 +93,43 @@ def _checkParams(self, params):
...
-            abstract_data = form.data
-            judgment = abstract_data.pop('judgment')
-            send_notification = abstract_data.pop('send_notifications')
-            contrib_session = abstract_data.pop('contrib_session', None)
-            merge_persons = abstract_data.pop('merge_persons', None)
-            judge_abstract(self.abstract, abstract_data, judgment=judgment, contrib_session=contrib_session,
-                           merge_persons=merge_persons, send_notification=send_notification, )
-        return WPManageAbstracts.render_template('abstract.html', self._conf, abstract=self.abstract, form=form)
+        elif form.validate_on_submit():
+            special_data, remaining_data = form.split_data
```
OmeGak: These are not very descriptive names. Why is the data special? What is the remaining data?

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -92,19 +93,43 @@ def _checkParams(self, params):
...
+            for abstract in submitted_abstracts:
+                judge_abstract(abstract, remaining_data, judge=session.user, **special_data)
+            num_judged_abstracts = len(submitted_abstracts)
+            num_prejudged_abstracts = len(self.abstracts) - num_judged_abstracts
+            if num_judged_abstracts:
+                flash(ngettext("The abstract has been judged.",
+                               "{num} abstracts have been judged.",
+                               num_judged_abstracts).format(num=num_judged_abstracts), 'success')
+            if num_prejudged_abstracts:
+                flash(ngettext("The abstract has been skipped since it is already judged.",
```
OmeGak: Why do we even allow to open the judgment dialog if there's a single abstract selected and it's already judged?

indico/modules/events/abstracts/templates/management/abstract_list.html
```diff
@@ -73,17 +73,48 @@
                     </ul>
                 </div>
                 <div class="group">
-                    <a class="i-button icon-checkmark js-requires-selected-row js-dialog-action accept">
-                        {% trans %}Accept{% endtrans %}
+                    <a class="i-button icon-checkmark accept js-requires-selected-row"
```
OmeGak: Are we really adding a button for each possible action in the end? Why not simply using the same dropdown we have on the judgment box in abstract display page?

indico/modules/events/abstracts/controllers/management.py
ThiefMaster: Is it really worth to add extra javascript to handle this edge case?

indico/modules/events/abstracts/forms.py
ThiefMaster: Why would it ever be missing? It's part of this class.

indico/modules/events/abstracts/forms.py
ThiefMaster: I'd go for `other_data`

indico/modules/events/abstracts/controllers/management.py
ThiefMaster: `data, kwargs`? `judgment_data, judgment_kwargs`?

indico/modules/events/abstracts/forms.py
nop33: :+1: for `judgment_data`

indico/modules/events/abstracts/operations.py
```diff
@@ -66,8 +66,9 @@ def delete_abstract(abstract, delete_contrib=False):
                            'Abstract "{}" has been deleted'.format(abstract.title), session.user)
 
 
-def judge_abstract(abstract, abstract_data, judgment, contrib_session=None, merge_persons=False,
+def judge_abstract(abstract, abstract_data, judgment, judge, contrib_session=None, merge_persons=False,
                    send_notification=False):
+    abstract.judge = judge
```
ThiefMaster: also set judgment_dt to the current time (otherwise it fails in the check constraint)

PR #2658, opened 2016-10-21
Add link to download Book of Abstracts

DavidAndreev:


indico/modules/events/abstracts/__init__.py
```diff
@@ -108,3 +108,15 @@ def _get_notification_placeholders(sender, **kwargs):
         obj = getattr(placeholders, name)
         if issubclass(obj, Placeholder):
             yield obj
+
+
+@signals.event.sidemenu.connect
+def _extend_event_menu(sender, **kwargs):
+    from indico.modules.events.layout.util import MenuEntryData
+
+    return MenuEntryData(
```
OmeGak: We generally use alignment with opening delimiter.
https://www.python.org/dev/peps/pep-0008/#indentation

indico/modules/events/abstracts/blueprint.py
```diff
@@ -49,6 +50,7 @@
 _bp.add_url_rule('/manage/abstracts/open', 'open_abstracts_call', RHOpenCFA, methods=('POST',))
 _bp.add_url_rule('/manage/abstracts/close', 'close_abstracts_call', RHCloseCFA, methods=('POST',))
 _bp.add_url_rule('/manage/abstracts/boa', 'manage_boa', RHManageBOA, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/abstracts/boa.pdf', 'export_boa', RHExportBOA, methods=('GET', 'POST'))
```
OmeGak: We just need `GET` for this endpoint since we are not sending any form data. You can omit `methods` altogether since `GET` is the default.

indico/modules/events/abstracts/__init__.py
```diff
@@ -108,3 +108,14 @@ def _get_notification_placeholders(sender, **kwargs):
         obj = getattr(placeholders, name)
         if issubclass(obj, Placeholder):
             yield obj
+
+
+@signals.event.sidemenu.connect
+def _extend_event_menu(sender, **kwargs):
+    from indico.modules.events.layout.util import MenuEntryData
+
+    return MenuEntryData(title=_("Book of Abstracts"),
```
ThiefMaster: You don't need to put every single kwarg in its own line

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -430,7 +431,7 @@ def __init__(self, conf, aw, contribs=None, tz=None, sort_by=""):
             'fields': [f for f in event.contribution_fields if f.is_active],
             'sorted_by': sort_by,
             'aw': aw,
-            'boa_text': conf.getBOAConfig().getText()
+            'boa_text': boa_settings.get(conf.as_event, 'extra_text')
```
ThiefMaster: I would use the opportunity to change the method to accept `event` instead of `conf`, then use just `event` here and `event.as_legacy` in places where you do need the old conference object.

indico/MaKaC/PDFinterface/conference.py
```diff
@@ -443,15 +444,10 @@ class AbstractBook(ContributionBook):
...
     _table_of_contents = True
 
-    def __init__(self, conf, aw, tz=None):
-        if not tz:
-            tz = conf.getTimezone()
-
-        sort_by = conf.getBOAConfig().getSortBy()
-        if not sort_by.strip() or sort_by not in ["number", "name", "sessionTitle", "speaker", "schedule"]:
-            sort_by = "number"
+    def __init__(self, conf):
```
ThiefMaster: take the `event` here; let's use legacy conference objects as rarely as possible

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -176,6 +176,15 @@ def _process(self):
         return jsonify_form(form)
 
 
+class RHExportBOA(RHManageAbstractsBase):
+    """Export book of abstracts"""
+
+    def _process(self):
+        pdf = AbstractBook(self._target)
+        fname = pdf.generate()
+        return send_file('BookOfAbstracts.pdf', fname, 'application/pdf')
```
ThiefMaster: `book-of-abstracts.pdf` is nicer (i'd also change it in the blueprint)

indico/modules/events/abstracts/blueprint.py
```diff
@@ -49,6 +50,7 @@
 _bp.add_url_rule('/manage/abstracts/open', 'open_abstracts_call', RHOpenCFA, methods=('POST',))
 _bp.add_url_rule('/manage/abstracts/close', 'close_abstracts_call', RHCloseCFA, methods=('POST',))
 _bp.add_url_rule('/manage/abstracts/boa', 'manage_boa', RHManageBOA, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/abstracts/boa.pdf', 'export_boa', RHExportBOA)
```
ThiefMaster: This is not only used in management, so it should be `/book-of-abstracts.pdf` (I wouldn't even put `/abstracts/` in the URL)

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -176,6 +176,15 @@ def _process(self):
         return jsonify_form(form)
 
 
+class RHExportBOA(RHManageAbstractsBase):
+    """Export book of abstracts"""
+
+    def _process(self):
+        pdf = AbstractBook(self._target)
```
ThiefMaster: We don't use `self._target` in new code. The legacy Conference object is in `self._conf` (avoid using this as much as possible), the new `Event` object is in `self.event_new`

indico/modules/events/abstracts/blueprint.py
OmeGak: :snake:

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -179,6 +179,15 @@ def _process(self):
         return jsonify_form(form)
 
 
+class RHExportBOA(RHManageAbstractsBase):
+    """Export book of abstracts"""
+
+    def _process(self):
+        pdf = AbstractBook(self.event_new)
+        fname = pdf.generate()
```
ThiefMaster: I also think it's cleaner to not assign it to an intermediate variable. But FYI `generate()` actually does return the path to the temporary file containing the PDF

indico/modules/events/abstracts/templates/management/overview.html
```diff
@@ -81,7 +81,7 @@
                 <div class="label">
                     {% trans %}Book of abstracts{% endtrans %}
                 </div>
-                {% trans url='#' -%}Customize and <a href={{ url }}>download</a> the book of abstracts{%- endtrans %}
+                {% trans url=url_for('.export_boa', event) -%}Customize and <a href={{ url }}>download</a> the book of abstracts{%- endtrans %}
```
ThiefMaster: However, wrapping it would probably be more readable.

indico/modules/events/abstracts/blueprint.py
```diff
@@ -50,6 +50,7 @@
 _bp.add_url_rule('/manage/abstracts/open', 'open_abstracts_call', RHOpenCFA, methods=('POST',))
 _bp.add_url_rule('/manage/abstracts/close', 'close_abstracts_call', RHCloseCFA, methods=('POST',))
 _bp.add_url_rule('/manage/abstracts/boa', 'manage_boa', RHManageBOA, methods=('GET', 'POST'))
+_bp.add_url_rule('/book-of-abstracts.pdf', 'export_boa', RHExportBOA)
```
ThiefMaster: Move this into the "Display pages" section. It's not management. `RHExportBOA` shouldn't be in management.py either. You could move it to display.py or add a new `boa.py` and move both this RH and RHManageBOA in there.

indico/modules/events/abstracts/templates/management/overview.html
```diff
@@ -81,7 +81,9 @@
                 <div class="label">
                     {% trans %}Book of abstracts{% endtrans %}
                 </div>
-                {% trans url='#' -%}Customize and <a href={{ url }}>download</a> the book of abstracts{%- endtrans %}
+                {% trans url=url_for('.export_boa', event) -%}
+                    Customize and <a href={{ url }}>download</a> the book of abstracts
```
ThiefMaster: we quote attributes

PR #2660, opened 2016-10-21
Send notifications when abstract is judged

pferreir:


indico/modules/events/abstracts/placeholders.py
```diff
@@ -164,7 +164,7 @@ class SubmitterTitlePlaceholder(Placeholder):
 
     @classmethod
     def render(cls, abstract):
-        return abstract.submitter.title
+        return unicode(abstract.submitter.title)
```
ThiefMaster: What about getting the raw string from it instead of translating it using the user's language?

    return orig_string(abstract.submitter.title)

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -85,7 +85,8 @@
     <div class="info-line submitted-data">
         {% if abstract.field_values %}
             <dl class="custom-field-values field-separator">
-                {% for value in abstract.field_values if value.contribution_field.is_active %}
+                {% for value in abstract.field_values|sort(attribute='contribution_field.position')
+                   if value.contribution_field.is_active %}
```
ThiefMaster: `|selectattr('contribution_field.is_active')`?

indico/modules/events/abstracts/placeholders.py
pferreir: Didn't know about `orig_string`. Thanks!

indico/modules/events/abstracts/templates/abstract/public.html
pferreir: But then you cannot wrap it nicely in 2 lines :(

pferreir: well, what I don't like is a trail of dozens of characters with no single space that is > 120 chars long.

PR #2662, opened 2016-10-21
Improvements wrt the abstract reviewing interface

pferreir:
Supersedes #2660.

Includes hiding/showing of certain features depending on user role (plus `can_judge`/`is_convener`).

### DIFF
--- a/indico/modules/events/abstracts/templates/abstract/public.html
+++ b/indico/modules/events/abstracts/templates/abstract/public.html
@@ -232,6 +232,43 @@
         </div>
 {% endmacro %}
 
+{% macro _render_judgment_action(abstract) %}
+    {% set action = abstract.state.name %}
+    {% if action == 'withdrawn' %}
+        {% trans name=abstract.submitter.name -%}
+            <strong>{{ name }}</strong> withdrew this abstract
+        {%- endtrans %}
+    {% elif abstract.can_judge(session.user) %}
+        {% if action == 'accepted' %}
+            {% trans name=abstract.judge.name -%}
+                <strong>{{ name }}</strong> accepted this abstract
+            {%- endtrans %}
+        {% elif action == 'rejected' %}
+            {% trans name=abstract.judge.name -%}
+                <strong>{{ name }}</strong> rejected this abstract
+            {%- endtrans %}
+        {% elif action == 'merged' %}
+            {% trans name=abstract.judge.name -%}
+                <strong>{{ name }}</strong> merged this abstract into another
+            {%- endtrans %}
+        {% elif action == 'duplicate' %}
+            {% trans name=abstract.judge.name -%}
+                <strong>{{ name }}</strong> marked this abstract as duplicate of another
+            {%- endtrans %}
+        {% endif %}
+    {% else %}
+        {% if action == 'accepted' %}
+            {% trans %}This abstract was accepted{% endtrans %}
+        {% elif action == 'rejected' %}
+            {% trans %}This abstract was rejected{% endtrans %}
+        {% elif action == 'merged' %}
+            {% trans %}This abstract was merged{% endtrans %}
+        {% elif action == 'duplicate' %}
+            {% trans %}This abstract was marked as duplicate{% endtrans %}
+        {% endif %}
+    {% endif %}
+{% endmacro %}
+
 {% macro render_abstract_public(abstract, management=false) %}
     {% set abstract_endpoint = '.manage_abstract' if management else '.display_abstract' %}
     <div id="abstract-title">
@@ -348,43 +385,6 @@
                     <div class="i-timeline-item-box header-indicator">
                         <div class="i-box-header flex">
                             <div class="stretch">
-                                {% set action = abstract.state.name %}
-                                {% if abstract.can_judge(session.user) %}
-                                    {% set judge_name = abstract.judge.name %}
-                                    {% if action == 'withdrawn' %}
-                                        {% trans name=judge_name -%}
-                                            <strong>{{ name }}</strong> withdrew this abstract
-                                        {%- endtrans %}
-                                    {% elif action == 'accepted' %}
-                                        {% trans name=judge_name -%}
-                                            <strong>{{ name }}</strong> accepted this abstract
-                                        {%- endtrans %}
-                                    {% elif action == 'rejected' %}
-                                        {% trans name=judge_name -%}
-                                            <strong>{{ name }}</strong> rejected this abstract
-                                        {%- endtrans %}
-                                    {% elif action == 'merged' %}
-                                        {% trans name=judge_name -%}
-                                            <strong>{{ name }}</strong> merged this abstract
-                                        {%- endtrans %}
-                                    {% elif action == 'duplicate' %}
-                                        {% trans name=judge_name -%}
-                                            <strong>{{ name }}</strong> marked this abstract as duplicate
-                                        {%- endtrans %}
-                                    {% endif %}
-                                {% else %}
-                                    {% if action == 'withdrawn' %}
-                                        {% trans %}This abstract was withdrawn{% endtrans %}
-                                    {% elif action == 'accepted' %}
-                                        {% trans %}This abstract was accepted{% endtrans %}
-                                    {% elif action == 'rejected' %}
-                                        {% trans %}This abstract was rejected{% endtrans %}
-                                    {% elif action == 'merged' %}
-                                        {% trans %}This abstract was merged{% endtrans %}
-                                    {% elif action == 'duplicate' %}
-                                        {% trans %}This abstract was marked as duplicate{% endtrans %}
-                                    {% endif %}
-                                {% endif %}
                                 {{ _render_judgment_action(abstract) }}
                                 <time datetime="{{ abstract.judgment_dt.isoformat() }}">
                                     {{- abstract.judgment_dt|format_human_date -}}


indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -436,6 +437,26 @@ def can_review(self, user):
         else:
             return False
 
+    def is_convener(self, user):
```
ThiefMaster: `if not user: return False`

indico/modules/events/abstracts/templates/abstract.html
```diff
@@ -10,9 +10,13 @@
         <div class="i-timetable-connect-down"></div>
         <div class="i-timeline-separator"></div>
         {{ render_abstract_comments(abstract) }}
-        {{ render_abstract_reviews(abstract) }}
-        <div class="i-timeline-separator"></div>
-        <div class="i-timetable-connect-up"></div>
-        {{ render_decision_box(abstract, form) }}
+        {% if abstract.is_convener(session.user) or abstract.can_judge(session.user) %}
```
ThiefMaster: What about adding a `can_see_reviews` method to `Abstract`?

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -335,9 +345,54 @@ <h3 class="abstract-title">
                     <div class="i-timeline-item-box header-indicator">
                         <div class="i-box-header flex">
                             <div class="stretch">
-                                {% trans name=abstract.judge.name, action=abstract.state.title|lower, date=abstract.judgment_dt|format_human_date -%}
-                                    <strong>{{ name }}</strong> {{ action }} this abstract on {{ date }}.
-                                {%- endtrans %}
+                                {% set judgement_dt = abstract.judgment_dt|format_human_date %}
```
ThiefMaster: `judgment`, not `judgement`

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -335,9 +345,54 @@ <h3 class="abstract-title">
...
+                                {% set action = abstract.state.name %}
+                                {% if abstract.can_judge(session.user) %}
+                                    {% set judge_name = abstract.judge.name %}
+                                    {% if action == 'withdrawn' %}
+                                        {% trans name=judge_name, date=judgement_dt -%}
+                                            <strong>{{ name }}</strong> withdrew this abstract on {{ date }}.
+                                        {% endtrans %}
```
ThiefMaster: `{%-` (also in the other endtrans tags below)

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -335,9 +345,54 @@ <h3 class="abstract-title">
...
-                                {% trans name=abstract.judge.name, action=abstract.state.title|lower, date=abstract.judgment_dt|format_human_date -%}
-                                    <strong>{{ name }}</strong> {{ action }} this abstract on {{ date }}.
-                                {%- endtrans %}
+                                {% set judgement_dt = abstract.judgment_dt|format_human_date %}
+                                {% set action = abstract.state.name %}
+                                {% if abstract.can_judge(session.user) %}
+                                    {% set judge_name = abstract.judge.name %}
+                                    {% if action == 'withdrawn' %}
+                                        {% trans name=judge_name, date=judgement_dt -%}
+                                            <strong>{{ name }}</strong> withdrew this abstract on {{ date }}.
```
ThiefMaster: "on yesterday" once again ;) maybe "withdrew this abstract ({{ date }})" instead?

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -335,9 +345,54 @@ <h3 class="abstract-title">
                     <div class="i-timeline-item-box header-indicator">
                         <div class="i-box-header flex">
                             <div class="stretch">
-                                {% trans name=abstract.judge.name, action=abstract.state.title|lower, date=abstract.judgment_dt|format_human_date -%}
-                                    <strong>{{ name }}</strong> {{ action }} this abstract on {{ date }}.
-                                {%- endtrans %}
+                                {% set judgement_dt = abstract.judgment_dt|format_human_date %}
+                                {% set action = abstract.state.name %}
+                                {% if abstract.can_judge(session.user) %}
```
ThiefMaster: this block is quite huge.. move to a macro?

indico/modules/events/abstracts/templates/abstract.html
pferreir: I guess that if we have other places where we'll need it...

indico/modules/events/abstracts/templates/abstract/public.html
pferreir: Yeah. I suggest we move time information somewhere else. E.g.:

image

(following a pattern that already exists in Indico)

pferreir: ... or this

image

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -232,6 +232,46 @@
         </div>
 {% endmacro %}
 
+{% macro _render_judgment_action(abstract) %}
+    {% set action = abstract.state.name %}
+    {% if abstract.can_judge(session.user) %}
+        {% set judge_name = abstract.judge.name %}
```
ThiefMaster: You can get in here if the abstract is e.g. withdrawn (which means there is no judge) since the macro is called whenever the state is not submitted and `can_judge` doesn't check the abstract state.

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -232,6 +232,46 @@
...
+        {% if action == 'withdrawn' %}
+            {% trans name=judge_name -%}
+                <strong>{{ name }}</strong> withdrew this abstract
```
ThiefMaster: There is no judge if an abstract is withdrawn

indico/htdocs/sass/partials/_timelines.scss
```diff
@@ -50,6 +50,13 @@
         .i-box-header {
             // To match i-timeline-item-label height
             min-height: $min-height;
+
+            time {
+                &:before {
+                    content: '\00B7\00a0';
```
ThiefMaster: `// bullet+nbsp` so people don't have to google what these chars are?

indico/modules/events/abstracts/templates/abstract.html
```diff
@@ -10,9 +10,13 @@
...
+            {{ render_abstract_reviews(abstract) }}
+            <div class="i-timeline-separator"></div>
+        {% endif %}
+        {% if abstract.state.name == 'submitted' and abstract.can_judge(session.user) %}
```
ThiefMaster: What about moving the state check into `can_judge` (and maybe `can_review` too)? We do the same in `can_edit`.

I'm quite sure we'll end up the ability to re-judge without resetting first at some time (we don't check it on the server at all right now :warning:) if we need to check both all the time.

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -316,12 +366,13 @@ <h3 class="abstract-title">
...
                     <div class="i-box-header">
-                        {% trans name=abstract.submitter.name, date=abstract.submitted_dt|format_human_date -%}
-                            <strong>{{ name }}</strong> submitted this abstract on {{ date }}
+                        {% trans name=abstract.submitter.name -%}
+                            <strong>{{ name }}</strong> submitted this abstract
                         {%- endtrans %}
                         {% if abstract.modified_dt %}
                             {{ render_edited_hint(abstract.modified_dt, abstract.modified_by) }}
                         {% endif %}
+                        <time>{{ abstract.submitted_dt|format_human_date }}</time>
```
ThiefMaster: Does this still look good if there's also a `modified_dt`? Also, `render_edited_hint` could probably make use of `<time>` too

indico/modules/events/abstracts/templates/abstract/review.html
```diff
@@ -73,12 +73,13 @@
...
                 <div class="stretch">
-                    {% trans name=review.user.name, date=review.created_dt|format_human_date -%}
-                        <strong>{{ name }}</strong> reviewed {{ date }}
+                    {% trans name=review.user.name -%}
+                        <strong>{{ name }}</strong> reviewed this abstract
                     {%- endtrans %}
                     {% if review.modified_dt %}
                         {{ render_edited_hint(comment.modified_dt) }}
                     {% endif %}
+                    <time>{{ review.created_dt|format_human_date }}</time>
```
ThiefMaster: `datetime="{{ review.created_dt.isoformat() }}"`

indico/modules/events/abstracts/templates/abstract.html
OmeGak: I would rather go for a `can_see_reviews` as well.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -436,6 +437,28 @@ def can_review(self, user):
...
+        else:
+            return False
+
+    def can_judge(self, user):
+        if not user:
+            return False
+        elif self.event_new.can_manage(user):
+            return True
+        elif self.is_convener(user):
+            return abstracts_reviewing_settings.get(self.event_new, 'allow_convener_judgment')
```
OmeGak: What about using the CFA property I had added in my PR? https://github.com/OmeGak/indico/commit/b0f52a762117a5c5278f5a2f81cb84aa8c2f64af

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -436,6 +437,28 @@ def can_review(self, user):
...
+            return True
+        else:
+            return False
+
+    def can_judge(self, user):
+        if not user:
+            return False
+        elif self.event_new.can_manage(user):
+            return True
+        elif self.is_convener(user):
```
OmeGak: `is_convener` is a more expensive check than `allow_convener_judgment`, so I would flip the order.

indico/modules/events/abstracts/templates/abstract/public.html
OmeGak: Should we maybe log who withdrew the abstract? It should generally be the submitter or one of the authors and accountability would not harm here, especially after abstract merging.

indico/modules/events/abstracts/templates/abstract/public.html
pferreir: Should we just use `abstract.submitter.name` in that case?

indico/modules/events/abstracts/templates/abstract/public.html
pferreir: Ehm... yeah, indeed (see comment above).

OmeGak: Also, this information should be visible to the submitter.

indico/modules/events/abstracts/templates/abstract/public.html
ThiefMaster: Yeah. I guess even if a manager does it it's done on request from the submitter.

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -241,19 +282,28 @@ <h3 class="abstract-title">
...
                     </a>
                 </div>
-            {% elif abstract.event_new.can_manage(session.user) %}
-                <div class="group">
-                    <a href="{{ url_for('.manage_abstract', abstract) }}" class="i-button highlight icon-switchon">
-                        {% trans %}Management view{% endtrans %}
-                    </a>
-                </div>
+            {% endif %}
+            {% if abstract.event_new.can_manage(session.user) %}
```
OmeGak: I think in the end we won't need this block, since all the actions are available both in display and management. Actually, at this point we should only have the display page.

indico/modules/events/abstracts/templates/abstract/public.html
ThiefMaster: Should it? Why does it concern him who performed some action?

Regarding logging, I think logging the action in the event log is enough. We don't really need to store it in the Abstract itself..

indico/modules/events/abstracts/templates/abstract/public.html
ThiefMaster: In an ideal world I'd agree with you. But you know how users are... some prefer to do things by phone/email/etc. instead of just doing it themselves in Indico itself...

indico/modules/events/abstracts/templates/abstract/public.html
pferreir: Well, if it's been withdrawn, I agree the submitter him/herself should be able to see by whom. After all it's an action on his/her side.

ThiefMaster: hmm in then maybe we should simply use the judge/judge_id for it (in that case the `judge_if_judged` constraint and the `JUDGED_STATES` set in the migration script need to be updated)

indico/modules/events/abstracts/templates/abstract/public.html
pferreir: It does, but I'll move the "edited" mention to the end of the line, it makes more sense.

indico/modules/events/abstracts/templates/abstract/public.html
pferreir: But I don't think it's a big deal if for now they always see it's been the submitter. After all, it will be at the submitter's request.

indico/modules/events/abstracts/templates/abstract/public.html
ThiefMaster: ah... then it might not be needed after all. anyway, i'd keep it for now, it's easy to remove later if we notice it's pointless

indico/modules/events/abstracts/templates/abstract.html
pferreir: > What about moving the state check into `can_judge` (and maybe `can_review` too)? We do the same in `can_edit`.

But shouldn't we have a way to see whether the user *would be able* to judge the abstract? Imagine we want to show the button anyway, only disabled?

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -436,6 +437,30 @@ def can_review(self, user):
...
+            return True
+        elif user.convener_for_tracks & self.reviewed_for_tracks:
+            return True
+        else:
+            return False
+
+    def can_judge(self, user, check_state=True):
+        if not user:
+            return False
+        if check_state and self.state != AbstractState.submitted:
```
ThiefMaster: `elif`?

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -231,6 +232,43 @@
...
+            {% trans name=abstract.judge.name -%}
+                <strong>{{ name }}</strong> accepted this abstract
+            {%- endtrans %}
+        {% elif action == 'rejected' %}
+            {% trans name=abstract.judge.name -%}
+                <strong>{{ name }}</strong> rejected this abstract
+            {%- endtrans %}
+        {% elif action == 'merged' %}
+            {% trans name=abstract.judge.name -%}
+                <strong>{{ name }}</strong> merged this abstract
```
ThiefMaster: `into another abstract`?

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -231,6 +232,43 @@
...
+            {%- endtrans %}
+        {% elif action == 'duplicate' %}
+            {% trans name=abstract.judge.name -%}
+                <strong>{{ name }}</strong> marked this abstract as duplicate
```
ThiefMaster: `of another abstract`?

PR #2663, opened 2016-10-24
My Abstracts display page

nop33:
- [x] List abstracts
- [x] PDF export

Before you judge the CSS classes I used: The display of the "my abstracts" list is the same as the "contribution list". To avoid creating new classes (renaming `contribution-` to `abstract-`) and abstracting the term of contribution and abstract, I reused the contribution CSS classes and imported the contributions scss file.
Preview:
image

### DIFF
--- a/indico/modules/events/abstracts/__init__.py
+++ b/indico/modules/events/abstracts/__init__.py
@@ -107,7 +107,14 @@ def _get_notification_placeholders(sender, **kwargs):
 
 @signals.event.sidemenu.connect
 def _extend_event_menu(sender, **kwargs):
+    from indico.modules.events.abstracts.util import get_user_abstracts
     from indico.modules.events.layout.util import MenuEntryData
 
-    return MenuEntryData(title=_("Book of Abstracts"), name='abstracts_book', endpoint='abstracts.export_boa',
-                         position=9, visible=lambda event: event.has_feature('abstracts'), static_site=True)
+    def _is_visible(conf):
+        return (session.user and conf.has_feature('abstracts') and
+                bool(get_user_abstracts(conf.as_event, session.user)))
+
+    yield MenuEntryData(title=_("Book of Abstracts"), name='abstracts_book', endpoint='abstracts.export_boa',
+                        position=9, visible=lambda event: event.has_feature('abstracts'), static_site=True)
+    yield MenuEntryData(title=_("My Abstracts"), name='user_abstracts', visible=_is_visible,
+                        endpoint='abstracts.my_abstracts', position=0, parent='call_for_abstracts')
--- a/indico/modules/events/abstracts/blueprint.py
+++ b/indico/modules/events/abstracts/blueprint.py
@@ -17,7 +17,8 @@
 from __future__ import unicode_literals
 
 from indico.modules.events.abstracts.controllers.boa import RHManageBOA, RHExportBOA
-from indico.modules.events.abstracts.controllers.display import RHDisplayAbstract, RHDisplayAbstractExportPDF
+from indico.modules.events.abstracts.controllers.display import (RHDisplayAbstract, RHDisplayAbstractExportPDF,
+                                                                 RHMyAbstracts)
 from indico.modules.events.abstracts.controllers.email_templates import (RHAddEmailTemplate, RHEditEmailTemplateRules,
                                                                          RHEditEmailTemplateText, RHEmailTemplateList,
                                                                          RHDeleteEmailTemplate, RHPreviewEmailTemplate,
@@ -35,6 +36,7 @@ from indico.modules.events.abstracts.controllers.management import (RHAbstracts,
                                                                     RHManageReviewingRoles, RHResetAbstractJudgment,
                                                                     RHBulkAbstractJudgment, RHAbstractNotificationLog)
 from indico.modules.events.abstracts.controllers.reviewing import RHAbstractsDownloadAttachment
+from indico.web.flask.util import make_compat_redirect_func
 from indico.web.flask.wrappers import IndicoBlueprint
 
 _bp = IndicoBlueprint('abstracts', __name__, url_prefix='/event/<confId>', template_folder='templates',
@@ -46,6 +48,7 @@ _bp.add_url_rule('/abstracts/<int:abstract_id>/notifications', 'notification_log
 _bp.add_url_rule('/abstracts/<int:abstract_id>/abstract.pdf', 'display_abstract_pdf_export', RHDisplayAbstractExportPDF)
 _bp.add_url_rule('/abstracts/<int:abstract_id>/attachments/<file_id>/<filename>', 'download_attachment',
                  RHAbstractsDownloadAttachment)
+_bp.add_url_rule('/abstracts/mine', 'my_abstracts', RHMyAbstracts)
 
 # Book of Abstracts
 _bp.add_url_rule('/manage/abstracts/boa', 'manage_boa', RHManageBOA, methods=('GET', 'POST'))
@@ -98,3 +101,10 @@ _bp.add_url_rule('/manage/abstracts/<int:abstract_id>/', 'manage_abstract', RHMa
 _bp.add_url_rule('/manage/abstracts/<int:abstract_id>/reset',
                  'reset_abstract_judgment', RHResetAbstractJudgment, methods=('POST',))
 _bp.add_url_rule('/manage/abstracts/<int:abstract_id>/abstract.pdf', 'manage_abstract_pdf_export', RHAbstractExportPDF)
+
+
+# Legacy URLs
+_compat_bp = IndicoBlueprint('compat_abstracts', __name__, url_prefix='/event/<event_id>')
+
+_compat_bp.add_url_rule('/call-for-abstracts/my-abstracts', 'my_abstracts',
+                        make_compat_redirect_func(_bp, 'my_abstracts', view_args_conv={'event_id': 'confId'}))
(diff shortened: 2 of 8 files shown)


indico/modules/events/abstracts/controllers/display.py
```diff
@@ -44,3 +49,26 @@ def _process(self):
         pdf = AbstractToPDF(self.abstract)
         file_name = secure_filename('abstract-{}.pdf'.format(self.abstract.friendly_id), 'abstract.pdf')
         return send_file(file_name, pdf.generate(), 'application/pdf')
+
+
+class RHDisplayProtectionBase(RHConferenceBaseDisplay):
```
ThiefMaster: Having the page accessible even without the menu item enabled doesn't disclose anything - it only shows data *you* entered. I don't think we need this check here. We have it for the other listings because those show names of lots of people.

indico/modules/events/abstracts/controllers/display.py
```diff
@@ -44,3 +49,26 @@ def _process(self):
...
+    def _checkProtection(self):
```
OmeGak: Why not simply moving this into `RHMyAbstracts`?

indico/modules/events/abstracts/util.py
```diff
@@ -371,3 +372,10 @@ def get_roles_for_event(event):
     roles['*']['reviewer'] = [reviewer.id for reviewer in event.global_abstract_reviewers]
     roles['*']['convener'] = [reviewer.id for reviewer in event.global_conveners]
     return roles
+
+
+@memoize_request
+def get_abstracts_with_user_as_submitter(event, user):
```
ThiefMaster: Used exactly once, I don't think we need a function for thins.

indico/modules/events/layout/default.py
```diff
@@ -97,9 +97,9 @@ def get_default_menu_entries():
             visible=_visibility_call_for_abstracts,
         ),
         MenuEntryData(
-            title=_("View my Abstracts"),
-            name='user_abstracts',
-            endpoint='event.userAbstracts',
+            title=_("My Abstracts"),
+            name='my_abstracts',
```
ThiefMaster: I wouldn't rename this for now. It just adds another step (updating all menu items with the old name in the DB) to the migration.

indico/modules/events/abstracts/templates/display/user_abstract_list.html
```diff
@@ -0,0 +1,65 @@
...
+                        {{- abstract.description|truncate(400) -}}
+                    </div>
+                </a>
+            {%- endif %}
+        </div>
+    {%- endfor %}
+{% endmacro %}
+
+{% block title -%}
+    {% trans %}My Abstracts{% endtrans %}
```
OmeGak: "My Submitted Abstracts" maybe?

OmeGak: "@ThiefMaster's Thumbed-down Abstracts" better?

ThiefMaster: haha, I just think "My Abstracts" sounds better. It's quite obvious that those are the ones you submitted

indico/modules/events/layout/default.py
```diff
@@ -97,9 +97,9 @@ def get_default_menu_entries():
             visible=_visibility_call_for_abstracts,
         ),
         MenuEntryData(
```
nop33: Similar to the `_extend_event_menu` function of the `contributions/__init__.py`?

indico/modules/events/abstracts/util.py
nop33: We'll need it for the PDF export as well

indico/modules/events/layout/default.py
ThiefMaster: Yes. If you add it maybe put it in the same place (and with the same name) as this one here so we get a conflict and don't end up with two functions for the same thing when merging those PRs:

https://github.com/indico/indico/pull/2658/files#diff-377bf7cb1677018ce7c94b1f7bc116d5R114

indico/modules/events/abstracts/controllers/display.py
```diff
@@ -44,3 +50,33 @@ def _process(self):
         pdf = AbstractToPDF(self.abstract)
         file_name = secure_filename('abstract-{}.pdf'.format(self.abstract.friendly_id), 'abstract.pdf')
         return send_file(file_name, pdf.generate(), 'application/pdf')
+
+
+class RHMyAbstractsBase(RHConferenceBaseDisplay):
+    """Base RH concerning the list of current user abstracts"""
+
+    MENU_ENTRY_NAME = 'user_abstracts'
```
ThiefMaster: not needed anymore

indico/modules/events/abstracts/blueprint.py
```diff
@@ -42,6 +44,8 @@
 _bp.add_url_rule('/abstracts/<int:abstract_id>/abstract.pdf', 'display_abstract_pdf_export', RHDisplayAbstractExportPDF)
 _bp.add_url_rule('/abstracts/<int:abstract_id>/attachments/<file_id>/<filename>', 'download_attachment',
                  RHAbstractsDownloadAttachment)
+_bp.add_url_rule('/abstracts/mine', 'my_abstracts', RHMyAbstracts)
+_bp.add_url_rule('/abstracts/my-abstracts.pdf', 'my_abstracts_pdf_export', RHMyAbstractsExportPDF)
```
nop33: Wut:question: How does the file `mine.pdf` in your computer indicate that it contains abstracts? I'll stick to `my-abstracts.pdf`...

ThiefMaster: Good point. While I'd usually comment if the filenames didn't match maybe this would be a case where it's not too bad to have `/mine.pdf` in the URL to be consistent with `/mine` but `my-abstracts.pdf` in the `send_file()` call so if someone saves it he still gets it as `my-abstracts.pdf`

ThiefMaster: How about `my_abstracts_pdf` instead of `my_abstracts_pdf_export`? The `_export` seems pretty redundant, the name is clear without it too.

nop33: Sure, but we have tons of other enpoints with the `_export`. Only in contributions we have:

* `contributions_pdf_export`
* `contributions_pdf_export_book`
* `contributions_pdf_export_book_sorted`
* `export_pdf`

ThiefMaster: ugh :/ not super pretty either IMO

indico/modules/events/abstracts/__init__.py
```diff
@@ -108,3 +108,16 @@ def _get_notification_placeholders(sender, **kwargs):
...
         if issubclass(obj, Placeholder):
             yield obj
+
+
+@signals.event.sidemenu.connect
+def _extend_event_menu(sender, **kwargs):
+    from indico.modules.events.abstracts.util import get_user_abstracts
+    from indico.modules.events.layout.util import MenuEntryData
+
+    def _visible_my_abstracts(conf):
```
OmeGak: I would simply go for `_is_visible`.

indico/modules/events/abstracts/templates/display/user_abstract_list.html
```diff
@@ -0,0 +1,72 @@
...
+                </span>
+            </div>
+            <div class="speaker-list icon-user">
+                {{ render_users(abstract.speakers|sort(attribute='display_order_key'),
+                                span_class='speaker-item-inline') }}
+            </div>
+            <div class="contrib-time icon-time">
+                {% set dt = abstract.modified_dt or abstract.submitted_dt %}
+                <span>{% trans %}Last modified:{% endtrans %}</span>
+                <span>{{ dt|format_datetime(format='short', timezone=timezone) }}</span>
```
OmeGak: Since it seems we are doing this now.

```html
<time datetime="{{ dt.isoformat() }}">{{ dt|format_human_date }}</time>
```

ThiefMaster:
Just found it weird to have /mine but then /my-abstracts.pdf

indico/modules/events/abstracts/util.py
```diff
@@ -371,3 +375,12 @@ def get_roles_for_event(event):
     roles['*']['reviewer'] = [reviewer.id for reviewer in event.global_abstract_reviewers]
     roles['*']['convener'] = [reviewer.id for reviewer in event.global_conveners]
     return roles
+
+
+@memoize_request
+def get_user_abstracts(event, user):
+    event_person = user.event_persons.filter(EventPerson.event_new == event).one()
```
ThiefMaster: Also, no need for a separate query:

```python
def get_user_abstracts(event, user):
    return (Abstract.query.with_parent(event)
            .filter(db.or_(Abstract.submitter == user,
                           Abstract.person_links.any(AbstractPersonLink.person.has(user=user))))
            .all())
```

indico/modules/events/abstracts/__init__.py
```diff
@@ -108,3 +108,16 @@ def _get_notification_placeholders(sender, **kwargs):
...
+        return (session.user and bool(get_user_abstracts(conf.as_event, session.user)) and
+                conf.as_event.has_feature('abstracts'))
```
ThiefMaster: `has_feature` is exposed on `Conference` too so no need to `.as_event` here..

indico/modules/events/abstracts/templates/display/user_abstract_list.html
nop33: No need to pass the timezone in this case?

indico/modules/events/abstracts/__init__.py
```diff
@@ -111,8 +111,8 @@ def _extend_event_menu(sender, **kwargs):
     from indico.modules.events.layout.util import MenuEntryData
 
     def _is_visible(conf):
-        return (session.user and bool(get_user_abstracts(conf.as_event, session.user)) and
-                conf.has_feature('abstracts'))
+        return (conf.has_feature('abstracts') and session.user and
```
ThiefMaster: `session.user / feature / get_user_abstracts` - the first one should be the cheapest one

PR #2665, opened 2016-10-24
Improve abstracts feature handling

DavidAndreev:


indico/modules/events/abstracts/__init__.py
```diff
@@ -87,7 +86,7 @@ def is_allowed_for_event(cls, event):
 
     @classmethod
     def is_default_for_event(cls, event):
```
ThiefMaster: Just remove this method. The default one in the base class always returns False

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -55,7 +55,7 @@ class RHManageAbstractsBase(RHConferenceModifBase):
     """Base class for all abstracts management RHs"""
 
     CSRF_ENABLED = True
-    EVENT_FEATURE = 'abstracts'
+    EVENT_FEATURE = None
```
ThiefMaster: :warning: only override this in the specific RH for the page that should always be accessible.

indico/modules/events/abstracts/templates/management/overview.html
```diff
@@ -1,118 +1,155 @@
...
-            <div class="text">
-                <div class="label">
-                    {% trans %}Reviewing{% endtrans %}
+                var $this = $(this);
+                $.ajax({
+                    url: $this.data('href'),
+                    method: $this.data('method'),
+                    complete: function() {
+                        IndicoUI.Dialogs.Util.progress(),
+                        window.location.reload();
```
ThiefMaster: `location.reload();` is enough

indico/modules/events/abstracts/templates/management/overview.html
```diff
@@ -1,118 +1,155 @@
...
-            <div class="icon icon-file-check"></div>
-            <div class="text">
-                <div class="label">
-                    {% trans %}Reviewing{% endtrans %}
+                var $this = $(this);
+                $.ajax({
+                    url: $this.data('href'),
+                    method: $this.data('method'),
+                    complete: function() {
+                        IndicoUI.Dialogs.Util.progress(),
```
ThiefMaster: This doesn't work as you expect it to. `progress()` returns the function to close the progress dialog. Reload in the `success` callback instead (and call `progress()` in there again before reloading).

ThiefMaster:
```javascript
complete: IndicoUI.Dialogs.Util.progress(),
success: function() { ... }
```

indico/modules/events/abstracts/templates/management/overview.html
```diff
@@ -1,118 +1,155 @@
...
-                    </a>
+    {% if not event.has_feature('abstracts') %}
+        <div class="action-box highlight">
+            <div class="section">
+                <div class="icon icon-file-check"></div>
+                <div class="text">
+                    <div class="label">
+                        {% trans %}Call for Abstracts disabled{% endtrans %}
+                    </div>
+                    {% trans %}"Call for Abstracts" provides an integrated abstract management and reviewing module that will allow you to set up a full workflow.{% endtrans %}
```
ThiefMaster: line is way too long. split in multiple lines (don't forget whitespace trimming in the trans tags)

indico/modules/events/abstracts/templates/management/overview.html
```diff
@@ -1,118 +1,157 @@
...
-            </div>
-            <div class="toolbar">
-                <div class="group">
-                    <a href="#" class="i-button icon-mail"
-                       data-href="{{ url_for('.email_tpl_list', event) }}"
-                       data-title="{% trans %}Configure e-mail notifications{% endtrans %}"
-                       data-ajax-dialog>
-                        {% trans %}Notifications{% endtrans %}
-                    </a>
+    {% if not event.has_feature('abstracts') %}
```
ThiefMaster: Why not use a separate template? There is almost no common code between the two cases.

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -127,6 +127,8 @@ def _process(self):
 class RHAbstracts(RHManageAbstractsBase):
     """Display abstracts management page"""
 
+    EVENT_FEATURE = None
+
     def _process(self):
         abstracts_count = Abstract.query.with_parent(self.event_new).count()
```
ThiefMaster: Not needed if the feature is not enabled

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -37,4 +37,21 @@
...
+                method: $this.data('method'),
+                complete: IndicoUI.Dialogs.Util.progress(),
+                success: function() {
+                    IndicoUI.Dialogs.Util.progress();
+                    location.reload();
+                },
+                error: handleAjaxError
+            });
+        });
+    };
```
nop33: This whole block looks like something useful for `declerative.js`. What do you guys think?
cc @ThiefMaster @pferreir @OmeGak

ThiefMaster: It's already there.. `data-reload-after` - I think this whole piece of code isn't needed now that I look at it again.

ThiefMaster: Ah no, we only check it for ajaxdialogs...

nop33: sth like `data-ajax-refresh`

indico/modules/events/abstracts/templates/management/disabled.html
```diff
@@ -0,0 +1,28 @@
+{% extends 'events/abstracts/management/_base.html' %}
+
+{% block content %}
+    <div class="action-box highlight">
+        <div class="section">
+            <div class="icon icon-file-check"></div>
+            <div class="text">
+                <div class="label">
+                    {% trans %}Call for Abstracts disabled{% endtrans %}
```
nop33: I see you're using correctly the `-` below, then why not here as well? :smile: 
```
{%- trans %}Call for Abstracts disabled{% endtrans -%}
```

indico/htdocs/js/indico/modules/abstracts.js
ThiefMaster: Why? We don't do that for `data-update` either

indico/modules/events/abstracts/templates/management/disabled.html
```diff
@@ -0,0 +1,28 @@
...
+                </div>
+                {% trans -%}
+                    "Call for Abstracts" provides an integrated abstract management and reviewing module
+                    that will allow you to set up a full workflow.
+                {%- endtrans %}
+            </div>
+            <div class="toolbar">
+                <a href="#" id="enable-abstracts" class="i-button highlight"
+                   data-href="{{ url_for('event_features.switch', event, feature='abstracts')  }}"
```
nop33: There's an extra space before `}}"`

indico/htdocs/js/indico/modules/abstracts.js
nop33: Hm, right. Then maybe we should allow `data-reload-after` also without an `ajaxdialog`?

indico/modules/events/abstracts/templates/management/disabled.html
DavidAndreev: Adrian says its not really needed ;D

ThiefMaster: yep, we don't really care about pretty whitespace in HTML - it's only relevant in cases where the extra space causes issues and *inside* trans because text is extracted for translation from there

PR #2666, opened 2016-10-26
"Convener" view of abstracts inside track

mvidalgarcia:
- I agree that the names/strings/sentences are not the most suitable, but I couldn't find better ones. 
- The link to an abstract display page is not working.
- I reused the styles from the Scientific Programme because I considered them appropriate.

List of tracks the user is a reviewer or convener of:
image
image

List of abstracts within a track the user is a reviewer or convener of:
image
image

### DIFF
--- a/indico/modules/events/abstracts/__init__.py
+++ b/indico/modules/events/abstracts/__init__.py
@@ -112,7 +112,7 @@ def _extend_event_menu(sender, **kwargs):
     yield MenuEntryData(title=_("Call for Abstracts"), name='call_for_abstracts', endpoint='event.conferenceCFA',
                         position=2, visible=lambda event: event.has_feature('abstracts'))
 
-    yield MenuEntryData(title=_("View my Tracks"), name='user_tracks', endpoint='abstracts.display_reviewable_tracks',
+    yield MenuEntryData(title=_("Reviewing Area"), name='user_tracks', endpoint='abstracts.display_reviewable_tracks',
                         position=2, parent='call_for_abstracts',
                         visible=lambda event: session.user and event.has_feature('abstracts'))
     yield MenuEntryData(title=_("Book of Abstracts"), name='abstracts_book', endpoint='abstracts.export_boa',
--- a/indico/modules/events/abstracts/__init__.py
+++ b/indico/modules/events/abstracts/__init__.py
@@ -109,14 +109,11 @@ def _get_notification_placeholders(sender, **kwargs):
 def _extend_event_menu(sender, **kwargs):
     from indico.modules.events.layout.util import MenuEntryData
 
-    return MenuEntryData(title=_("Book of Abstracts"), name='abstracts_book', endpoint='abstracts.export_boa',
-                         position=9, visible=lambda event: event.has_feature('abstracts'), static_site=True)
+    yield MenuEntryData(title=_("Call for Abstracts"), name='call_for_abstracts', endpoint='event.conferenceCFA',
+                        position=2, visible=lambda event: event.has_feature('abstracts'))
 
-
-@signals.event.sidemenu.connect
-def _extend_event_menu(sender, **kwargs):
-    from indico.modules.events.layout.util import MenuEntryData
-
-    return MenuEntryData(title=_("View my Tracks"), name='user_tracks', endpoint='abstracts.display_reviewable_tracks',
-                         position=2, parent='call_for_abstracts',
-                         visible=lambda event: session.user and event.has_feature('abstracts'))
+    yield MenuEntryData(title=_("View my Tracks"), name='user_tracks', endpoint='abstracts.display_reviewable_tracks',
+                        position=2, parent='call_for_abstracts',
+                        visible=lambda event: session.user and event.has_feature('abstracts'))
+    yield MenuEntryData(title=_("Book of Abstracts"), name='abstracts_book', endpoint='abstracts.export_boa',
+                        position=9, visible=lambda event: event.has_feature('abstracts'), static_site=True)
--- a/indico/modules/events/abstracts/__init__.py
+++ b/indico/modules/events/abstracts/__init__.py
@@ -118,5 +118,5 @@ def _extend_event_menu(sender, **kwargs):
     from indico.modules.events.layout.util import MenuEntryData
 
     return MenuEntryData(title=_("View my Tracks"), name='user_tracks', endpoint='abstracts.display_reviewable_tracks',
-                         position=2, parent='call_for_abstracts', visible=lambda event: event.has_feature('abstracts'),
-                         static_site=True)
+                         position=2, parent='call_for_abstracts',
+                         visible=lambda event: session.user and event.has_feature('abstracts'))
--- a/indico/modules/events/abstracts/__init__.py
+++ b/indico/modules/events/abstracts/__init__.py
@@ -111,3 +111,12 @@ def _extend_event_menu(sender, **kwargs):
 
     return MenuEntryData(title=_("Book of Abstracts"), name='abstracts_book', endpoint='abstracts.export_boa',
                          position=9, visible=lambda event: event.has_feature('abstracts'), static_site=True)
+
+
+@signals.event.sidemenu.connect
+def _extend_event_menu(sender, **kwargs):
+    from indico.modules.events.layout.util import MenuEntryData
+
+    return MenuEntryData(title=_("View my Tracks"), name='user_tracks', endpoint='abstracts.display_reviewable_tracks',
+                         position=2, parent='call_for_abstracts', visible=lambda event: event.has_feature('abstracts'),
+                         static_site=True)
--- a/indico/modules/events/abstracts/blueprint.py
+++ b/indico/modules/events/abstracts/blueprint.py
@@ -17,15 +17,13 @@
 from __future__ import unicode_literals
 
 from indico.modules.events.abstracts.controllers.boa import RHManageBOA, RHExportBOA
-from indico.modules.events.abstracts.controllers.display import (RHDisplayAbstract, RHDisplayAbstractExportPDF,
-                                                                 RHDisplayReviewableTracks,
-                                                                 RHDisplayReviewableTrackAbstracts)
+from indico.modules.events.abstracts.controllers.display import RHDisplayAbstract, RHDisplayAbstractExportPDF
 from indico.modules.events.abstracts.controllers.email_templates import (RHAddEmailTemplate, RHEditEmailTemplateRules,
                                                                          RHEditEmailTemplateText, RHEmailTemplateList,
                                                                          RHDeleteEmailTemplate, RHPreviewEmailTemplate,
                                                                          RHSortEmailTemplates)
-from indico.modules.events.abstracts.controllers.management import (RHAbstracts, RHAbstractList,
-                                                                    RHAbstractListCustomize, RHAbstractListStaticURL,
+from indico.modules.events.abstracts.controllers.management import (RHAbstracts, RHAbstractListCustomize,
+                                                                    RHAbstractListStaticURL,
                                                                     RHManageAbstractSubmission, RHManageAbstract,
                                                                     RHManageAbstractReviewing, RHCreateAbstract,
                                                                     RHDeleteAbstracts, RHAbstractPersonList,
@@ -35,8 +33,12 @@ from indico.modules.events.abstracts.controllers.management import (RHAbstracts,
                                                                     RHAbstractsExportJSON,
                                                                     RHScheduleCFA, RHOpenCFA, RHCloseCFA,
                                                                     RHManageReviewingRoles, RHResetAbstractJudgment,
-                                                                    RHBulkAbstractJudgment, RHAbstractNotificationLog)
-from indico.modules.events.abstracts.controllers.reviewing import RHAbstractsDownloadAttachment
+                                                                    RHBulkAbstractJudgment, RHAbstractNotificationLog,
+                                                                    RHAbstractList)
+from indico.modules.events.abstracts.controllers.reviewing import (RHAbstractsDownloadAttachment,
+                                                                   RHDisplayReviewableTracks,
+                                                                   RHDisplayReviewableTrackAbstracts,
+                                                                   RHDisplayAbstractListCustomize)
 from indico.web.flask.wrappers import IndicoBlueprint
 
 _bp = IndicoBlueprint('abstracts', __name__, url_prefix='/event/<confId>', template_folder='templates',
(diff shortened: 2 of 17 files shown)


indico/modules/events/abstracts/blueprint.py
```diff
@@ -44,6 +46,9 @@
 _bp.add_url_rule('/abstracts/<int:abstract_id>/abstract.pdf', 'display_abstract_pdf_export', RHDisplayAbstractExportPDF)
 _bp.add_url_rule('/abstracts/<int:abstract_id>/attachments/<file_id>/<filename>', 'download_attachment',
                  RHAbstractsDownloadAttachment)
+_bp.add_url_rule('/call-for-abstracts/reviewing', 'display_reviewable_tracks', RHDisplayReviewableTracks)
```
ThiefMaster: trailing slash since we have stuff inside

indico/modules/events/abstracts/blueprint.py
```diff
@@ -44,6 +46,9 @@
...
+_bp.add_url_rule('/call-for-abstracts/reviewing/track/<int:track_id>', 'display_reviewable_track_abstracts',
```
ThiefMaster: I don't think the `/track` part is really needed unless we have anything else inside `/reviewing/` that's not a track

indico/modules/events/abstracts/controllers/display.py
```diff
@@ -44,3 +46,29 @@ def _process(self):
         pdf = AbstractToPDF(self.abstract)
         file_name = secure_filename('abstract-{}.pdf'.format(self.abstract.friendly_id), 'abstract.pdf')
         return send_file(file_name, pdf.generate(), 'application/pdf')
+
+
+class RHDisplayReviewableTracks(RHConferenceBaseDisplay):
+    def _process(self):
+        user = session.user
+        tracks_set, events_set = set(), set()
```
ThiefMaster: why initialize them empty first? just create them like this:

    tracks_set = user.convener_for_tracks | user.abstract_reviewer_for_tracks
    events_set = user.global_convener_for_events | user.global_abstract_reviewer_for_event

indico/modules/events/abstracts/controllers/display.py
```diff
@@ -44,3 +46,29 @@ def _process(self):
...
+        tracks_set.update(user.convener_for_tracks, user.abstract_reviewer_for_tracks)
+        events_set.update(user.global_convener_for_events, user.global_abstract_reviewer_for_events)
+        for e in events_set:
+            tracks_set.update(e.tracks)
+        tracks = {t for t in tracks_set if t.event_id == self.event_new.id}
```
ThiefMaster: Actually you can already filter it like this above:

    tracks_set = (user.convener_for_tracks | user.abstract_reviewer_for_tracks) & set(event.tracks)

indico/modules/events/abstracts/controllers/display.py
```diff
@@ -44,3 +46,29 @@ def _process(self):
...
         file_name = secure_filename('abstract-{}.pdf'.format(self.abstract.friendly_id), 'abstract.pdf')
         return send_file(file_name, pdf.generate(), 'application/pdf')
+
+
+class RHDisplayReviewableTracks(RHConferenceBaseDisplay):
+    def _process(self):
+        user = session.user
+        tracks_set, events_set = set(), set()
+        tracks_set.update(user.convener_for_tracks, user.abstract_reviewer_for_tracks)
+        events_set.update(user.global_convener_for_events, user.global_abstract_reviewer_for_events)
```
ThiefMaster: This page is only for one event, so you can simply do this:

```
if user in self.event_new.global_abstract_reviewers or user in self.event_new.global_conveners:
    tracks_set.update(self.event_new.tracks)
```

indico/modules/events/abstracts/controllers/display.py
```diff
@@ -44,3 +46,29 @@ def _process(self):
...
+                                                  tracks=tracks)
+
+
+class RHDisplayReviewableTrackAbstracts(RHConferenceBaseDisplay):
+    def _checkParams(self, params):
+        RHConferenceBaseDisplay._checkParams(self, params)
+        self.track = Track.get_one(request.view_args['track_id'])
+
+    def _process(self):
+        abstracts_set = set()
```
ThiefMaster: same as in my comment above, `abstract_set = foo | bar | ...`

indico/modules/events/abstracts/controllers/display.py
```diff
@@ -44,3 +46,29 @@ def _process(self):
...
+            tracks_set.update(e.tracks)
+        tracks = {t for t in tracks_set if t.event_id == self.event_new.id}
+        return WPDisplayAbstracts.render_template('display/tracks.html', self._conf, event=self.event_new,
+                                                  tracks=tracks)
+
+
+class RHDisplayReviewableTrackAbstracts(RHConferenceBaseDisplay):
+    def _checkParams(self, params):
+        RHConferenceBaseDisplay._checkParams(self, params)
+        self.track = Track.get_one(request.view_args['track_id'])
```
ThiefMaster: url normalization, otherwise you can use a track id from an unrelated event you might not even have access to!

simply copy the rule from `RHManageTrackBase`

indico/modules/events/abstracts/templates/display/abstracts.html
```diff
@@ -0,0 +1,25 @@
+{% extends 'layout/conference_page_base.html' %}
+
+{% block title %}
+    {% trans %} Abstracts from {% endtrans %} <span class="track-name">{{ track_title }}</span>
```
ThiefMaster: The last part should be inside the trans - otherwise it'd look awful in any language where the title isn't at the very end

indico/modules/events/abstracts/__init__.py
```diff
@@ -111,3 +111,12 @@ def _extend_event_menu(sender, **kwargs):
...
                          position=9, visible=lambda event: event.has_feature('abstracts'), static_site=True)
+
+
+@signals.event.sidemenu.connect
+def _extend_event_menu(sender, **kwargs):
+    from indico.modules.events.layout.util import MenuEntryData
+
+    return MenuEntryData(title=_("View my Tracks"), name='user_tracks', endpoint='abstracts.display_reviewable_tracks',
+                         position=2, parent='call_for_abstracts', visible=lambda event: event.has_feature('abstracts'),
+                         static_site=True)
```
ThiefMaster: This doesn't belong in a static/offline site so don't set `static_site` (it defaults to False).

indico/modules/events/abstracts/__init__.py
```diff
@@ -111,3 +111,12 @@ def _extend_event_menu(sender, **kwargs):
...
     return MenuEntryData(title=_("Book of Abstracts"), name='abstracts_book', endpoint='abstracts.export_boa',
                          position=9, visible=lambda event: event.has_feature('abstracts'), static_site=True)
+
+
+@signals.event.sidemenu.connect
+def _extend_event_menu(sender, **kwargs):
+    from indico.modules.events.layout.util import MenuEntryData
+
+    return MenuEntryData(title=_("View my Tracks"), name='user_tracks', endpoint='abstracts.display_reviewable_tracks',
+                         position=2, parent='call_for_abstracts', visible=lambda event: event.has_feature('abstracts'),
```
ThiefMaster: `session.user and ...` (the item shouldn't be visible if you are not logged in)

ThiefMaster:
I don't like the current style very much to be honest. Looks a lot like legacy to me...

OmeGak:
I think we should reuse the same style @nop33 is using for listing abstracts in https://github.com/indico/indico/pull/2663.

OmeGak:
I think the menu should be called `Reviewing area`, and it should list all abstracts per-track, maybe with the possibility of filtering.

mvidalgarcia:
Right now I'm just working on the track list, @DavidAndreev is working on the abstract list.

Current look:
image
The total of abstracts (unreviewed + reviewed) is shown in a tooltip when hovering the badge.

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -17,27 +17,38 @@
 
 from __future__ import unicode_literals
 
-from flask import request
+from flask import request, session
+from indico.core.db import db
+from indico.modules.users import User
```
nop33: Together with the line above they should be moved one block below.

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -56,3 +67,55 @@ def _checkParams(self, params):
...
+        if not session.user:
+            raise Forbidden
+
+    def _process(self):
+        query = Track.query.with_parent(self.event_new)
+        track_reviewer_abstract_count = get_track_reviewer_abstract_counts(self.event_new, session.user)
+
+        # if the user is not a global convener only show their tracks
+        if (session.user not in self.event_new.global_abstract_reviewers and
+                session.user not in self.event_new.global_conveners):
```
nop33: Wrong indentation

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -56,3 +67,55 @@ def _checkParams(self, params):
...
+            query = query.filter(db.or_(Track.conveners.any(User.id == session.user.id),
+                                        Track.abstract_reviewers.any(User.id == session.user.id)))
+
+        return WPDisplayAbstractsReviewing.render_template('display/tracks.html', self._conf, event=self.event_new,
+                                                           tracks=query.all(),
+                                                           abstract_count=track_reviewer_abstract_count)
```
nop33: It would look a bit nicer if this line was swapped with the line above, but that's just something completely OCD so you can ignore it if you want!

indico/modules/events/abstracts/templates/display/abstracts.html
```diff
@@ -0,0 +1,77 @@
+{% extends 'layout/conference_page_base.html' %}
+{% from 'events/abstracts/management/_abstract_list.html' import render_abstract_list %}
+{% from 'events/management/_lists.html' import render_displayed_entries_fragment %}
```
nop33: Do we have a rule that we first `import` and then `extends`? It would make sense to me..

indico/modules/events/abstracts/templates/display/abstracts.html
```diff
@@ -0,0 +1,77 @@
...
+    {% trans track_title=track.title -%}
+        Abstracts from <span class="track-name">{{ track_title }}</span>
+    {%- endtrans %}
+{% endblock %}
+
+{% block content %}
+    {% if not abstracts %}
+        <p>
+            {%- trans -%}
+                There are currently no abstracts within this track which you are reviewer or convener of.
```
nop33: There are currently no abstracts within this track **for** which you are **a** reviewer or **a** convener ~~of~~. Isn't this better? I'm asking for opinions.

indico/modules/events/abstracts/templates/display/tracks.html
```diff
@@ -0,0 +1,44 @@
...
+{% block title %}
+    {% trans %}Reviewing Area{% endtrans %}
+{% endblock %}
+
+{% block content %}
+    <div class="track-review-list">
+        {% if not tracks %}
+            <p>
+                {%- trans -%}
+                    There are currently no tracks which you are reviewer or convener of.
```
nop33: Same here maybe?

indico/modules/events/abstracts/templates/display/tracks.html
```diff
@@ -0,0 +1,44 @@
...
+                {%- endtrans -%}
+            </p>
+        {% else %}
+            <ul>
+                {% for track in tracks %}
+                    <div class="track-review-row">
+                        <li>
+                            <div class="title">
+                                <a href="{{ url_for('abstracts.display_reviewable_track_abstracts', track) }}">{{ track.title }}</a>
```
nop33: How about breaking the line?

indico/modules/events/abstracts/util.py
```diff
@@ -374,3 +406,29 @@ def get_roles_for_event(event):
...
+    # that have some abstract in the submitted state.
+    count_total = db.func.count(db.case({AbstractState.submitted.value: Abstract.id}, value=Abstract.state))
+    count_reviewed = db.func.count(db.case({AbstractState.submitted.value: AbstractReview.id}, value=Abstract.state))
+    query = (Track.query.with_parent(event)
+             .with_entities(Track, count_total, count_reviewed)
+             .outerjoin(Track.abstracts_reviewed)
+             .outerjoin(AbstractReview, db.and_(AbstractReview.abstract_id == Abstract.id,
+                                                AbstractReview.user_id == user.id))
+             .group_by(Track.id))
+    return {track: (total, reviewed) for track, total, reviewed in query}
```
nop33: Wouldn't it be better to return a dict with other dict-s like this?
```python
return {track: {'total: total, 'reviewed': reviewed'} for track, total, reviewed in query}
```
Then in the template you can do
```
abstract_count[track]['total'] - abstract_count[track]['reviewed']
```
instead of 
```
abstract_count[track].0 - abstract_count[track].1
```

indico/modules/events/abstracts/controllers/reviewing.py
ThiefMaster: No. Continuation indent for wrapped `if` conditions is twice the regular width.

indico/modules/events/abstracts/templates/display/abstracts.html
ThiefMaster: Usually extends should be the first thing in a template since if affects everything else (internally). While imports don't matter in that regard it's still cleaner to follow that guideline IMO - that way you always see if it's extending or not by looking at the very first line.

indico/modules/events/abstracts/util.py
ThiefMaster: Probably a good idea. Initially I wrote it like this since it's only two values but if if you don't use tuple unpacking to get actual variable names it indeed gets a bit ugly.

indico/modules/events/abstracts/controllers/reviewing.py
nop33: what? why? This looks ugly and wrong... Am I the only one who thinks that?

ThiefMaster: Probably because it's much clearer that way where the body starts:

```python
if (foo or
    bar):
    foobar()

if (foo or
        bar):
    foobar()
```

indico/modules/events/abstracts/templates/display/abstracts.html
nop33: IMO it's not that important to know exactly what it `extends` from, if you wanna know, you can look for it in the first lines. I would prefer to see the `import`s first and then the `extends`. The `import`s are not affected by the `extend`, but the rest of the template (`macro`s) are based on the `extend`, so I would keep those closer and the `import`s at the top.

indico/modules/events/abstracts/controllers/reviewing.py
nop33: All I am saying is that this:
```python
 if (session.user not in self.event_new.global_abstract_reviewers and
     session.user not in self.event_new.global_conveners):
    foo()
```
looks better than this imo:
```python
 if (session.user not in self.event_new.global_abstract_reviewers and
                session.user not in self.event_new.global_conveners):
    foo()
```

mvidalgarcia: so? :smile:

indico/modules/events/abstracts/templates/display/abstracts.html
mvidalgarcia: @DavidAndreev, you may want to have a look at this since you're working on this file

PR #2667, opened 2016-10-26
Add dialog to edit abstract in management page

DavidAndreev:


indico/modules/events/abstracts/operations.py
```diff
@@ -54,6 +54,17 @@ def create_abstract(event, abstract_data, custom_fields_data=None):
     return abstract
 
 
+def update_abstract(abstract, data):
+    abstract.person_links = []
```
ThiefMaster: still need to check if there's a better way to avoid this workaround. it's the old problem with sqlalchemy where replacing a relationship's data may result in the new data being inserted before the old has been removed. or maybe we  somehow accidentally create new abstract person links instead of reusing the existing ones...

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -314,8 +314,13 @@ <h3 class="abstract-title">
...
                     {% else %}
-                        {# TODO: open abstract edition form #}
-                        <a href="#" class="i-button icon-edit" title="{% trans %}Edit abstract{% endtrans %}"></a>
+                        <a href="#" class="i-button icon-edit"
+                           title="{% trans %}Edit abstract{% endtrans %}"
+                           data-href="{{ url_for('.edit_abstract', abstract) }}"
+                           data-title="{% trans %}Edit abstract{% endtrans %}"
+                           data-ajax-dialog
+                           data-reload-after>
+                        </a>
```
ThiefMaster: the tag has no content so i'd merge this into the previous line

indico/htdocs/sass/modules/contributions/_editor.scss
```diff
@@ -172,16 +172,6 @@ $editor-preview-bg-color: $light-gray;
     content: "\e14c";
 }
 
-.wmd-help-button {
```
ThiefMaster: legacy file, will eventually be fully removed, but it was causing a duplicate "help" button in the wmd button bar so we just removed it from there already

indico/modules/events/abstracts/controllers/common.py
```diff
@@ -0,0 +1,45 @@
...
+
+from indico.modules.events.abstracts.controllers.reviewing import RHAbstractReviewBase
+from indico.modules.events.abstracts.operations import update_abstract
+from indico.modules.events.abstracts.util import make_abstract_form
+from indico.modules.events.util import get_field_values
+from indico.util.i18n import _
+from indico.web.util import jsonify_data, jsonify_form
+
+
+class RHEditAbstract(RHAbstractReviewBase):
```
OmeGak: I don't think we want to extend from `RHAbstractReviewBase`. Edition has nothing to do with reviewing.

indico/modules/events/abstracts/forms.py
```diff
@@ -396,7 +399,9 @@ class MultiTrackMixin(object):
 
     def __init__(self, *args, **kwargs):
         event = kwargs['event']
-        if abstracts_settings.get(event, 'tracks_required'):
+        self.track_field_disabled = (kwargs.get('abstract') and
+                                     kwargs['abstract'].edit_track_mode != EditTrackMode.both)
+        if abstracts_settings.get(event, 'tracks_required') and not self.track_field_disabled:
```
OmeGak: What about moving `tracks_required` to a `CallForAbstracts` property?

OmeGak: Maybe it's worth having a cleverer way to specify which properties should be proxied in `CallForAbstracts`.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -66,6 +67,12 @@ class AbstractReviewingState(TitledIntEnum):
     mixed = 5
 
 
+class EditTrackMode(int, IndicoEnum):
```
OmeGak: `AbstractEditTrackMode`

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -367,6 +374,18 @@ def candidate_tracks(self):
         return {t for t in self.reviewed_for_tracks if self.get_track_reviewing_state(t) in states}
 
     @property
+    def edit_track_mode(self):
+        if not inspect(self).persistent:
+            return EditTrackMode.both
+        elif self.state not in {AbstractState.submitted, AbstractState.withdrawn}:
```
OmeGak: I don't think it makes sense allowing changing the tracks when the abstract is withdrawn.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -66,6 +67,12 @@ class AbstractReviewingState(TitledIntEnum):
     mixed = 5
 
 
+class EditTrackMode(int, IndicoEnum):
+    none = 0
```
OmeGak: Do we really need this much expressivity? I thought the only two cases we need is **synced** and **not-synced**.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -367,6 +374,18 @@ def candidate_tracks(self):
         return {t for t in self.reviewed_for_tracks if self.get_track_reviewing_state(t) in states}
 
     @property
+    def edit_track_mode(self):
+        if not inspect(self).persistent:
+            return EditTrackMode.both
+        elif self.state not in {AbstractState.submitted, AbstractState.withdrawn}:
+            return EditTrackMode.none
+        elif (self.public_state in (AbstractPublicState.awaiting, AbstractPublicState.withdrawn) and
+                self.reviewed_for_tracks == self.submitted_for_tracks):
```
OmeGak: So when there is a disagreement between `reviewed_for_tracks` and `submitted_for_tracks` the submitter won't be able to change it even is there is no review? IIRC, last thing we discussed was that we would allow one to override another.

indico/modules/events/abstracts/operations.py
```diff
@@ -39,15 +39,30 @@
 from indico.util.fs import secure_filename
 
 
+def _update_tracks(abstract, tracks, only_reviewed_for=False):
+    edit_track_mode = abstract.edit_track_mode
+    if edit_track_mode == EditTrackMode.none:
```
OmeGak: I don't think we should do these types of check in the operation functions. I would move it to the RH.

indico/modules/events/abstracts/operations.py
```diff
@@ -39,15 +39,30 @@
...
+        return
+
+    if tracks is None:
+        tracks = set()
+    elif isinstance(tracks, Track):
+        tracks = {tracks}
+
+    if edit_track_mode == EditTrackMode.both:
```
OmeGak: Same here, I would be expecting the operation primitives to simply do things rather than checking states.

indico/modules/events/abstracts/operations.py
```diff
@@ -64,6 +79,19 @@ def create_abstract(event, abstract_data, custom_fields_data=None):
     return abstract
 
 
+def update_abstract(abstract, abstract_data, custom_fields_data=None):
+    tracks = abstract_data.pop('submitted_for_tracks', None)
+    if abstract.edit_track_mode == EditTrackMode.both:
```
OmeGak: I don't think we need this check. Is there any case in which we should allow editing the abstract but not the tracks?

indico/modules/events/abstracts/operations.py
```diff
@@ -39,15 +39,30 @@
 from indico.util.fs import secure_filename
 
 
+def _update_tracks(abstract, tracks, only_reviewed_for=False):
```
OmeGak: I would move this to a method in `Abstract`.

indico/modules/events/abstracts/controllers/common.py
ThiefMaster: The naming of the base class is somewhat bad but it's nonetheless the right one in terms of functionality. Refactoring the messy inheritance is already on my TODO list though

indico/modules/events/abstracts/models/abstracts.py
ThiefMaster: Yes, because the `_update_track` util is now used for *all* cases where we update one of the track lists

indico/modules/events/abstracts/models/abstracts.py
ThiefMaster: At least at some point we were talking about disallowing changes by the submitter when a manager changed something

PR #2669, opened 2016-10-26
Render and add reviews

pferreir:
This is in no way near finished, but one should be already able to add their own reviews.
Judgments broken.

### DIFF
--- a/indico/htdocs/js/indico/jquery/declarative.js
+++ b/indico/htdocs/js/indico/jquery/declarative.js
@@ -16,8 +16,9 @@
  */
 
 /* eslint-disable max-len */
+/* global getParamsFromSelectors:false */
 
-(function() {
+(function(global) {
     'use strict';
 
     $(document).ready(function() {
@@ -27,6 +28,19 @@
         setupSelectAllNone();
         setupAnchorLinks();
     });
+
+    global.getParamsFromSelectors = function getParamsFromSelectors() {
+        var fieldParams = {};
+        _.each(arguments, function(selector) {
+            $(selector).each(function() {
+                if (!(this.name in fieldParams)) {
+                    fieldParams[this.name] = [];
+                }
+                fieldParams[this.name].push($(this).val());
+            });
+        });
+        return fieldParams;
+    };
 
     function setupSelectAllNone() {
         $('body').on('click', '[data-select-all]', function() {
@@ -86,14 +100,7 @@
             }
 
             if (paramsSelector) {
-                var fieldParams = {};
-                $(paramsSelector).each(function() {
-                    if (!(this.name in fieldParams)) {
-                        fieldParams[this.name] = [];
-                    }
-                    fieldParams[this.name].push($(this).val());
-                });
-                params = $.extend({}, fieldParams, params);
+                params = $.extend({}, getParamsFromSelectors(paramsSelector), params);
             }
 
             function updateHtml(selector, html, triggeredBy) {
@@ -225,4 +232,4 @@
             }).html('&para;').appendTo($elem);
         });
     }
-})();
+})(window);
--- a/indico/htdocs/js/indico/jquery/declarative.js
+++ b/indico/htdocs/js/indico/jquery/declarative.js
@@ -42,6 +42,12 @@
         return fieldParams;
     };
 
+    global.getFormParams = function getFormParams($form) {
+        return getParamsFromSelectors($form.find('input[type=text]:not(:disabled), input[type=hidden]:not(:disabled), \
+                                                  input:checked:not(:disabled), textarea:not(:disabled), \
+                                                  select:not(:disabled)'));
+    };
+
     function setupSelectAllNone() {
         $('body').on('click', '[data-select-all]', function() {
             var selector = $(this).data('select-all');
--- a/indico/htdocs/js/indico/modules/abstracts.js
+++ b/indico/htdocs/js/indico/modules/abstracts.js
@@ -15,7 +15,7 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
-/* global setupListGenerator:false */
+/* global setupListGenerator:false, initForms:false, showFormErrors:false, getFormParams:false */
 
 (function(global) {
     'use strict';
@@ -36,5 +36,67 @@
         });
 
         setupListGenerator(filterConfig);
+    };
+
+    global.setupAbstractPage = function setupAbstractPage() {
+        $('body').on('indico:confirmed', '.judge-button', function(evt) {
+            var $box = $('#abstract-decision-box');
+            var $this = $(this);
+            var $form = $box.find('form');
+
+            evt.preventDefault();
+
+            $.ajax({
+                url: $this.data('href'),
+                method: $this.data('method'),
+                complete: IndicoUI.Dialogs.Util.progress(),
+                error: handleAjaxError,
+                data: getFormParams($form),
+                success: function(data) {
+                    if (data.page_html) {
+                        $('.management-page').replaceWith(data.page_html);
+                        $box = $('#abstract-decision-box');
+                    } else {
+                        $box.html(data.box_html);
+                    }
+                    initForms($box);
+                    showFormErrors($box);
+                }
+            });
+        }).on('indico:confirmed', '.review-button', function(evt) {
+            var $this = $(this);
+            var $box = $this.closest('.abstract-review-box');
+            var $form = $box.find('form');
+            var $page = $('.management-page');
+
+            evt.preventDefault();
+
+            $.ajax({
+                url: $this.data('href'),
+                method: $this.data('method'),
+                complete: IndicoUI.Dialogs.Util.progress(),
+                error: handleAjaxError,
+                data: getFormParams($form),
+                success: function(data) {
+                    if (data.page_html) {
+                        $page.replaceWith(data.page_html);
+                    } else {
+                        $box.replaceWith(data.box_html);
+                    }
+                    initForms($page);
+                    showFormErrors($page);
+                }
+            });
+        }).on('click', '.change-review-button', function(evt) {
+            evt.preventDefault();
+
+            var $reviewDisplay = $(this).closest('.i-timeline-item');
+            var $reviewBox = $reviewDisplay.next('.abstract-review-box');
+            $reviewDisplay.hide();
+            $reviewBox.show();
+        }).on('indico:htmlUpdated', function() {
+            initForms($(this));
+            showFormErrors($(this));
+        });
     };
 })(window);
--- a/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
+++ b/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
@@ -20,6 +20,20 @@
 @import 'partials/boxes';
 @import 'partials/icons';
 
+%rating-bullet {
+    display: inline-block;
+    vertical-align: middle;
+    width: 0.8em;
+    height: 1.2em;
+    margin-right: 0.2em;
+    border-radius: 1em;
+    background-color: $pastel-gray;
+}
+
+%rating-bullet-selected {
+    background-color: $blue;
+}
+
 #abstract-title {
     display: flex;
 
@@ -218,11 +232,11 @@
 
     > li {
         label {
+            @include border-right($width: 2px);
             color: $dark-gray;
             font-weight: normal;
             font-style: italic;
 
-            border-right: 2px solid $pastel-gray;
             padding-right: 0.5em;
         }
 
@@ -233,19 +247,41 @@
                 color: $light-black;
             }
 
-            .rating {
-                .bullet {
-                    display: inline-block;
-                    vertical-align: middle;
-                    width: 0.5em;
-                    height: 1em;
-                    border-radius: 1em;
-                    background-color: $pastel-gray;
-
-                    &.full {
-                        background-color: $pastel-green;
-                    }
+            .rating .bullet {
+                @extend %rating-bullet;
+                &.full {
+                    @extend %rating-bullet-selected;
                 }
             }
         }
+    }
+}
+
+.rating-widget {
+    user-select: none;
+    display: flex;
+    flex-direction: row-reverse;
+
+    input[type="radio"] {
+        display: none;
+
+        &:checked ~ label.bullet {
+            @extend %rating-bullet-selected;
+        }
+    }
+    .bullet {
+        @extend %rating-bullet;
+        cursor: pointer;
+    }
+
+    .rating-value {
+        margin-left: 0.25em;
+        font-size: 1.2em;
+        line-height: 1.1;
+        width: 0.5em;
+    }
+}
+
+.abstract-review-box.weak-hidden {
+    display: none;
 }
--- a/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
+++ b/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
@@ -212,3 +212,40 @@
     margin-top: 5px;
     margin-right: 10px;
 }
+
+.abstract-question-list {
+    list-style-type: none;
+
+    > li {
+        label {
+            color: $dark-gray;
+            font-weight: normal;
+            font-style: italic;
+
+            border-right: 2px solid $pastel-gray;
+            padding-right: 0.5em;
+        }
+
+        .value {
+            .number {
+                vertical-align: middle;
+                margin-left: 0.5em;
+                color: $light-black;
+            }
+
+            .rating {
+                .bullet {
+                    display: inline-block;
+                    vertical-align: middle;
+                    width: 0.5em;
+                    height: 1em;
+                    border-radius: 1em;
+                    background-color: $pastel-gray;
+
+                    &.full {
+                        background-color: $pastel-green;
+                    }
+                }
+            }
+        }
+}
--- a/indico/modules/events/abstracts/blueprint.py
+++ b/indico/modules/events/abstracts/blueprint.py
@@ -33,9 +33,11 @@ from indico.modules.events.abstracts.controllers.management import (RHAbstracts,
                                                                     RHAbstractsExportCSV, RHAbstractsExportExcel,
                                                                     RHAbstractsExportJSON,
                                                                     RHScheduleCFA, RHOpenCFA, RHCloseCFA,
-                                                                    RHManageReviewingRoles, RHResetAbstractJudgment,
-                                                                    RHBulkAbstractJudgment, RHAbstractNotificationLog)
-from indico.modules.events.abstracts.controllers.reviewing import RHAbstractsDownloadAttachment
+                                                                    RHManageReviewingRoles, RHBulkAbstractJudgment,
+                                                                    RHAbstractNotificationLog)
+from indico.modules.events.abstracts.controllers.reviewing import (RHAbstractsDownloadAttachment,
+                                                                   RHReviewAbstractForTrack, RHJudgeAbstract,
+                                                                   RHResetAbstractJudgment, RHListOtherAbstracts)
 from indico.web.flask.util import make_compat_redirect_func
 from indico.web.flask.wrappers import IndicoBlueprint
 
@@ -54,6 +56,8 @@ _bp.add_url_rule('/abstracts/mine.pdf', 'my_abstracts_pdf', RHMyAbstractsExportP
 # Book of Abstracts
 _bp.add_url_rule('/manage/abstracts/boa', 'manage_boa', RHManageBOA, methods=('GET', 'POST'))
 _bp.add_url_rule('/book-of-abstracts.pdf', 'export_boa', RHExportBOA)
+
+_bp.add_url_rule('/abstracts/other-list', 'other_abstracts', RHListOtherAbstracts, methods=('POST',))
 
 # Management
 _bp.add_url_rule('/manage/abstracts/', 'manage_abstracts', RHAbstracts)
@@ -103,9 +107,19 @@ _bp.add_url_rule('/manage/abstracts/<int:abstract_id>/reset',
                  'reset_abstract_judgment', RHResetAbstractJudgment, methods=('POST',))
 _bp.add_url_rule('/manage/abstracts/<int:abstract_id>/abstract.pdf', 'manage_abstract_pdf_export', RHAbstractExportPDF)
 
+# Abstract reviewing actions
+for prefix in ('/manage/abstracts', '/abstracts'):
+    defaults = {'management': prefix != '/abstracts'}
+    _bp.add_url_rule(prefix + '/<int:abstract_id>/reset', 'reset_abstract_judgment',
+                     RHResetAbstractJudgment, methods=('POST',), defaults=defaults)
+    _bp.add_url_rule(prefix + '/<int:abstract_id>/abstract.pdf', 'manage_abstract_pdf_export',
+                     RHAbstractExportPDF, defaults=defaults)
+    _bp.add_url_rule(prefix + '/<int:abstract_id>/review/track/<int:track_id>', 'review_abstract',
+                     RHReviewAbstractForTrack, methods=('POST',), defaults=defaults)
+    _bp.add_url_rule(prefix + '/<int:abstract_id>/judge', 'judge_abstract',
+                     RHJudgeAbstract, methods=('POST',), defaults=defaults)
 
 # Legacy URLs
 _compat_bp = IndicoBlueprint('compat_abstracts', __name__, url_prefix='/event/<event_id>')
-
 _compat_bp.add_url_rule('/call-for-abstracts/my-abstracts', 'my_abstracts',
                         make_compat_redirect_func(_bp, 'my_abstracts', view_args_conv={'event_id': 'confId'}))
--- a/indico/modules/events/abstracts/controllers/base.py
+++ b/indico/modules/events/abstracts/controllers/base.py
@@ -42,7 +42,7 @@ def build_review_form(abstract, track):
 def render_abstract_page(abstract, management=False):
     review_forms = {track.id: build_review_form(abstract, track)
                     for track in abstract.reviewed_for_tracks
-                    if track.can_review(session.user)}
+                    if track.can_review_abstracts(session.user)}
     judgment_form = AbstractJudgmentForm(abstract=abstract)
 
     return render_template('events/abstracts/abstract.html', abstract=abstract, judgment_form=judgment_form,
@@ -70,7 +70,7 @@ class AbstractPageMixin(AbstractMixin):
     def _process(self):
         review_forms = {track.id: build_review_form(self.abstract, track)
                         for track in self.abstract.reviewed_for_tracks
-                        if track.can_review(session.user)}
+                        if track.can_review_abstracts(session.user)}
         judgment_form = AbstractJudgmentForm(abstract=self.abstract)
 
         return self.page_class.render_template('abstract.html', self._conf, abstract=self.abstract,
--- a/indico/modules/events/abstracts/controllers/base.py
+++ b/indico/modules/events/abstracts/controllers/base.py
@@ -16,10 +16,37 @@
 
 from __future__ import unicode_literals
 
-from flask import request, session
+from flask import render_template, request, session
 from werkzeug.exceptions import Forbidden
 
 from indico.modules.events.abstracts.models.abstracts import Abstract
+from indico.modules.events.abstracts.forms import AbstractJudgmentForm, make_review_form
+from indico.web.forms.base import FormDefaults
+
+
+def build_review_form(abstract, track):
+    review_form_class = make_review_form(abstract.event_new)
+    reviews_for_track = abstract.get_reviews(user=session.user, track=track)
+    review_for_track = reviews_for_track[0] if reviews_for_track else None
+
+    if review_for_track:
+        answers = {'question_{}'.format(rating.question.id): rating.value
+                   for rating in review_for_track.ratings}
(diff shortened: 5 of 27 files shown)


indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -51,8 +58,97 @@ class RHAbstractsDownloadAttachment(RHAbstractsReviewBase):
...
+                        for track in self.abstract.reviewed_for_tracks
+                        if track.is_user_reviewer(session.user)}
+        judgement_form = AbstractJudgmentForm(abstract=self.abstract)
+
+        return WPManageAbstracts.render_template('abstract.html', self._conf, abstract=self.abstract,
+                                                 judgement_form=judgement_form, review_forms=review_forms,
+                                                 management=self.management)
+
+
+class RHJudgeAbstract(RHAbstractReviewBase):
```
OmeGak: OTOH, why are we extending from `RHAbstractReviewBase`? IIRC, we agreed judges wouldn't be able to review unless listed as such.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -449,7 +449,7 @@ def is_convener(self, user):
         else:
             return False
 
-    def can_judge(self, user, check_state=True):
+    def can_judge(self, user, check_state=False):
```
OmeGak: YES:exclamation:

indico/modules/events/abstracts/models/review_questions.py
```diff
@@ -75,6 +75,10 @@ class AbstractReviewQuestion(db.Model):
     # relationship backrefs:
     # - ratings (AbstractReviewRating.question)
 
+    def get_review_rating(self, review):
```
OmeGak: I think we try to add the helper methods below `__repr__`.

indico/modules/events/abstracts/templates/abstract.html
```diff
@@ -14,9 +14,20 @@
             {{ render_abstract_reviews(abstract) }}
             <div class="i-timeline-separator"></div>
         {% endif %}
-        {% if abstract.can_judge(session.user) %}
+
+        {% set can_review = abstract.can_review(session.user) %}
+        {% set can_judge = abstract.can_judge(session.user, check_state=true) %}
+
+        {% if can_review or can_judge %}
```
OmeGak: We should also check if `abstract.state.name == 'submitted'`, otherwise we'll still get the connector with no other section afterwards.

indico/modules/events/abstracts/templates/abstract/review.html
```diff
@@ -18,13 +19,54 @@
...
+            {{ form_rows(form, skip=('comment', 'proposed_action')) }}
+            {{ form_rows(form, fields=('comment', 'proposed_action')) }}
+            {% set form_id = '#track-review-form-' + track.id|string %}
+            {% call form_footer(form, skip_label=skip_labels) %}
+                <button class="i-button big highlight"
+                        data-update="#abstract-review-box-{{ track.id }}"
+                        data-params-selector='{{ form_id }} select, {{ form_id }} input[type="radio"]:checked, {{ form_id }} input[type="hidden"], {{ form_id }} textarea'
+                        data-method="POST"
+                        data-href="{{ url_for('.review_abstract', abstract, track) }}">
+                    {% trans %}Review{% endtrans%}
```
OmeGak: To be exact, the reviewer has already reviewed the abstract at this point. I would maybe rename it to `Send review` or `Submit review`.

indico/modules/events/abstracts/templates/abstract/review.html
```diff
@@ -67,7 +109,15 @@
     </span>
 {% endmacro %}
 
-{% macro _render_review(review) %}
+{% macro _render_rating(cfa, value) %}
+    <span class="rating">
+    {%- for bullet in range(cfa.rating_range[0], cfa.rating_range[1]) -%}
```
OmeGak: Not enough indentation.

indico/modules/events/abstracts/util.py
```diff
@@ -371,3 +372,27 @@ def get_roles_for_event(event):
     roles['*']['reviewer'] = [reviewer.id for reviewer in event.global_abstract_reviewers]
     roles['*']['convener'] = [reviewer.id for reviewer in event.global_conveners]
     return roles
+
+
+def make_review_form(event):
```
OmeGak: Wouldn't it make more sense to have this in `abstracts/forms.py`?

ThiefMaster: So far we always had these functions in util.py but forms.py wouldn't be a bad place either

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -74,8 +74,10 @@
                 {% endfor %}
             </div>
             <div class="i-box-footer">
-                {{ simple_form(form, back_button=false, skip_labels=true, submit=_("Judge"),
-                               disabled_until_change=false, form_header_kwargs={'orientation': 'vertical'}) }}
+                {{ simple_form(form, skip_labels=true, submit=_("Judge"),
```
OmeGak: Any reason to remove `back_button=false`?

indico/modules/events/abstracts/forms.py
```diff
@@ -195,6 +195,17 @@ def __init__(self, *args, **kwargs):
         super(AbstractJudgmentForm, self).__init__(*args, **kwargs)
 
 
+class AbstractReviewForm(IndicoForm):
+    """Form for judging an abstract"""
+
+    comment = TextAreaField(_("Comments"))
+    proposed_action = IndicoEnumSelectField(_("Proposed Action"), [DataRequired()], enum=AbstractAction,
```
OmeGak: Missing `proposed_contribution_type` field.

OmeGak: Also, I would move the comment to the very bottom of the box.

indico/modules/events/abstracts/templates/abstract.html
```diff
@@ -14,9 +14,20 @@
             {{ render_abstract_reviews(abstract) }}
```
OmeGak: Shouldn't the reviewer be able to see this section with its reviews only?

OmeGak:
While I like the bullets for display, I'm not sure it's clear they are actionable. It may not be the prettiest, but maybe good ol' radio inputs wouldn't be so bad.

screenshot from 2016-10-27 11 51 06

indico/htdocs/js/indico/jquery/declarative.js
```diff
@@ -86,14 +99,7 @@
...
             if (paramsSelector) {
-                var fieldParams = {};
-                $(paramsSelector).each(function() {
-                    if (!(this.name in fieldParams)) {
-                        fieldParams[this.name] = [];
-                    }
-                    fieldParams[this.name].push($(this).val());
-                });
-                params = $.extend({}, fieldParams, params);
+                params = $.extend({}, global.getParamsFromSelectors(paramsSelector), params);
```
ThiefMaster: I'd rather declare it as global to make eslint happy and then call it without `global.` here (we don't do it anywhere else and it feels like doing `window.foo()` instead of just `foo()` for something not directly related to "window")

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -15,7 +15,7 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
-/* global setupListGenerator:false */
+/* global setupListGenerator:false, getParamsFromSelectors: false, initForms: False */
```
ThiefMaster: `s/False/false` and so far we never put spaces after the `:` in a global declaration; the [example](http://eslint.org/docs/user-guide/configuring#specifying-globals) doesn't put spaces there either

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -37,4 +37,61 @@
...
+                error: handleAjaxError,
+                data: getParamsFromSelectors($form.find('input:not(:disabled), textarea, select:not(:disabled)')),
+                success: function(data) {
+                    if (data.page_html) {
+                        $('.management-page').replaceWith(data.page_html);
+                        $box = $('#abstract-decision-box');
+                    } else {
+                        $box.replaceWith(data.box_html);
+                    }
+                    initForms($box);
```
ThiefMaster: I think you need to call `showFormErrors` explicitly to get the error tooltips in case of server-side validation failure; `initForms` doesn't call it

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -37,4 +37,61 @@
...
+            var $form = $box.find('form');
+
+            evt.preventDefault();
+
+            $.ajax({
+                url: $this.data('href'),
+                method: 'POST',
+                complete: IndicoUI.Dialogs.Util.progress(),
+                error: handleAjaxError,
+                data: getParamsFromSelectors($form.find('input:checked, input[type="hidden"], textarea, select')),
```
ThiefMaster: the quotes are optional for "simple" attribute selectors (but i think i also add them more often than not ;))

indico/modules/events/abstracts/controllers/base.py
```diff
@@ -16,10 +16,43 @@
...
+
+
+def get_user_review_for_track(user, abstract, track):
+    results = [review for review in abstract.get_track_reviews(track) if review.user == user]
+    assert len(results) <= 1
+    return results[0] if results else None
+
+
+def build_review_form(abstract, track):
+    review_form_class = make_review_form(abstract.event_new)
```
ThiefMaster: is having both `build_review_form` and `make_review_form` a good idea?

indico/modules/events/tracks/models/tracks.py
```diff
@@ -108,3 +108,11 @@ def __repr__(self):
 
     def can_delete(self, user):
         return self.event_new.can_manage(user) and not self.abstracts_accepted
+
+    def is_user_reviewer(self, user):
```
ThiefMaster: Heh, looks like we added pretty much the same methods ;)

ee30b2d213754dabe9d0d287f91863bd255a2e6f (not in devel yet; it's in the branch @DavidAndreev, @mvidalgarcia and me worked on yesterday)

I think we should decide whether we want to check the reviewer/convener ACL role too or not and then remove it in all places where we currently do or add it to the places where we don't.

I'm tending towards not checking it since being in the reviewer list should be a strong enough criterion to consider someone a reviewer.

Either way, I'd add a `if user is None: return False` here - I doubt we even end up calling this if there's no user but in case we do there's no need to even fetch the reviewer/convener lists.

For the naming... I like yours more (never liked "can_convene" much, sorry @OmeGak ;)) - OTOH, `can_xxx` is what we use everywhere else even though it's technically not the correct name (since it's not checking whether the object can do something but the user passed to the method..)

indico/modules/events/abstracts/templates/forms/rating_widget.html
```diff
@@ -0,0 +1,22 @@
+{% extends 'forms/base_widget.html' %}
+
+{% block html %}
+    <div id="{{ field.id }}" class="rating-widget">
+        <span class="rating-value">{{ '-' if field.object_data is none else field.data }}</span>
+        {% for bullet in range(cfa.rating_range[1], cfa.rating_range[0] - 1, -1) %}
+            {% set radio_id = field.id + "-" + bullet|string %}
```
ThiefMaster: `'%d-%d'|format(field.id, bullet)`?

indico/modules/events/abstracts/templates/forms/rating_widget.html
```diff
@@ -0,0 +1,22 @@
...
+            <input type="radio" id="{{ radio_id }}" name="{{ field.name }}" value="{{ bullet }}"
+                   {{ 'checked' if field.data == bullet|string else ''}}>
```
ThiefMaster: `else ''` is not needed; it defaults to `Undefined` which evaluates as an empty string

indico/modules/events/abstracts/templates/forms/rating_widget.html
```diff
@@ -0,0 +1,22 @@
...
+            <label class="bullet" title="{{ bullet }}" for="{{ radio_id }}"></label>
```
ThiefMaster: Not sure if our CSS handles it correctly, but you could probably just wrap the input inside the label instead of giving the radio buttons IDs

indico/modules/events/abstracts/templates/forms/rating_widget.html
```diff
@@ -0,0 +1,22 @@
...
+        {% endfor %}
+    </div>
+{% endblock %}
+
+{% block javascript %}
+    <script>
+        $('#{{ field.id }} input[type="radio"]').on('click', function() {
+            var $widget = $(this).closest('.rating-widget');
```
ThiefMaster: `var $this = $(this);`? or you could use `this.value` in the next line

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -51,8 +58,97 @@ class RHAbstractsDownloadAttachment(RHAbstractsReviewBase):
...
     def _checkParams(self, params):
-        RHAbstractsReviewBase._checkParams(self, params)
+        RHAbstractReviewBase._checkParams(self, params)
         self.abstract_file = AbstractFile.get_one(request.view_args['file_id'])
 
     def _process(self):
         return self.abstract_file.send()
+
+
+def get_user_review_for_track(user, abstract, track):
```
pferreir: Jolly good idea, young lad!

indico/modules/events/tracks/models/tracks.py
pferreir: I agree `can_convene` sounds awful. Anyway, I'm using your naming since it is more consistent with what we have.

indico/modules/events/abstracts/models/review_questions.py
```diff
@@ -78,3 +78,12 @@ class AbstractReviewQuestion(db.Model):
     @return_ascii
     def __repr__(self):
         return format_repr(self, 'id', 'event_id', no_score=False, is_deleted=False, _text=self.text)
+
+    def get_review_rating(self, review):
+        """
+        Get the rating given for a particular review.
```
ThiefMaster: fits nicely in the previous line

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -40,52 +40,56 @@
...
-                               disabled_until_change=false, form_header_kwargs={'orientation': 'vertical'}) }}
-            </div>
+            {% endfor %}
         </div>
+        <div class="i-box-footer">
+            {{ form_header(form, orientation='vertical') }}
+            {{ form_rows(form, fields=form._order) }}
+            {% call form_footer(form, skip_label=skip_labels) %}
+                <button class="i-button big highlight judge-button"
+                        data-href="{{ url_for('.judge_abstract', abstract, management=management) }}">
```
ThiefMaster: Wouldn't it be cleaner to put this in the `action` of the form and then bind a submit event for the form?

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -40,52 +40,56 @@
...
+                    {% trans %}Judge{% endtrans %}
+                </button>
+            {% endcall %}
+
```
ThiefMaster: extra whitespace

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -186,9 +186,11 @@
                     {% endif %}
                 </div>
                 <div>
-                    {% trans url=url_for('contributions.display_contribution', abstract.contribution) %}
-                        <a href="{{ url }}">Go to contribution</a>
-                    {% endtrans %}
+                    {% if abstract.contribution %}
+                        {% trans url=url_for('contributions.display_contribution', abstract.contribution) %}
```
ThiefMaster: The trans tag has inner whitespace that's not stripped.

Anyway, I'd move the link outside the trans block.

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -394,49 +396,56 @@ <h3 class="abstract-title">
...
                                        data-title="{% trans %}Reset judgment{% endtrans %}"
                                        data-confirm="{% trans %}Do you really want to reset the judgment? This operation is irreversible.{% endtrans %}"></a>
                                 </div>
                             {% endif %}
                         </div>
                         <div class="i-box-content">
                             {% if abstract.state.name == 'accepted' %}
-                                {% set url = url_for('contributions.display_contribution', abstract.contribution) %}
+                                {% if abstract.contribution %}
+                                    {% set link_start = '<a href="' + url_for('contributions.display_contribution', abstract.contribution) + '">' %}
```
ThiefMaster: `'<a href="%s">'|format(url_for(...))`

indico/modules/events/abstracts/templates/abstract/review.html
```diff
@@ -1,30 +1,72 @@
...
                 {% endfor %}
             {% endif %}
         {% endfor %}
     </div>
 {% endmacro %}
 
+{% macro render_review_boxes(abstract, review_forms, management=false) %}
+    {% for track in abstract.reviewed_for_tracks|sort(attribute='title') %}
+        {% set user_reviewed = abstract.get_reviews(track=track)|selectattr('user', 'equalto', session.user)|list %}
+        {% if track.can_review(session.user) and user_reviewed|length == 0 %}
```
ThiefMaster: `not user_reviewed`?

indico/modules/events/abstracts/templates/abstract/review.html
```diff
@@ -67,10 +109,19 @@
     </span>
 {% endmacro %}
 
-{% macro _render_review(review) %}
+{% macro _render_rating(cfa, value) %}
+    <span class="rating">
+    {%- for bullet in range(cfa.rating_range[0], cfa.rating_range[1] + 1) -%}
```
ThiefMaster: indentation

indico/modules/events/abstracts/templates/abstract/review.html
```diff
@@ -97,14 +148,25 @@
...
-            {% if review.ratings %}
-                {% for rating in review.ratings %}
+                {% if can_change_review %}
                     <div>
-                        {# TODO: UI #}
-                        {{ rating.question.text }}
-                        {{ rating.value }}
+                        <a class="i-button-icon icon-edit change-review-button"
+                           title="{% trans %}Change review{% endtrans %}">
+                        </a>
```
ThiefMaster: move into the previous line? (usually we do that if a tag has no content)

indico/web/forms/widgets.py
```diff
@@ -80,12 +80,13 @@ class JinjaWidget(object):
...
     """
 
-    def __init__(self, template, plugin=None, single_line=False, single_kwargs=False, **context):
+    def __init__(self, template, plugin=None, single_line=False, single_kwargs=False, inline_js=False, **context):
         self.template = template
         self.plugin = plugin
         self.context = context
         self.single_line = single_line
         self.single_kwargs = single_kwargs
+        self.inline_js = inline_js
```
ThiefMaster: Is this because you manually update the HTML of an element? In that case simply call `injectJS(data.js)` afterwards. Then the inlining here shouldn't be needed at all.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -505,8 +507,18 @@ def get_track_reviewing_state(self, track):
...
 
-    def get_track_reviews(self, track):
-        return [x for x in self.reviews if x.track == track]
+    def get_reviews(self, track=None, user=None):
+        """Get all reviews on a particular track and/or by a particular user.
+
+           :param track: will show only reviews for the given track
+           :param user: will show only review by the given user
+        """
+        reviews = self.reviews
```
pferreir: True. Thanks for spotting.

indico/modules/events/abstracts/forms.py
```diff
@@ -150,9 +175,11 @@ class AbstractJudgmentFormBase(IndicoForm):
                                description=_("The generated contribution will be allocated in this session"))
     duplicate_of = AbstractField(_("Duplicate of"),
                                  [HiddenUnless('judgment', AbstractAction.mark_as_duplicate), DataRequired()],
-                                 description=_("The current abstract will be marked as duplicate of the selected one"))
+                                 description=_("The current abstract will be marked as duplicate of the selected one"),
+                                 ajax_endpoint='abstracts.other_abstracts')
```
pferreir: Done.

indico/modules/events/abstracts/templates/abstract/review.html
pferreir: My brain was in JavaScript mode.

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -37,4 +37,67 @@
...
+        $('body').on('click', '.judge-button', function(evt) {
+            var $box = $('#abstract-decision-box');
+            var $this = $(this);
+            var $form = $box.find('form');
+
+            evt.preventDefault();
+
+            $.ajax({
+                url: $this.data('href'),
+                method: 'POST',
```
ThiefMaster: Shouldn't happen when you `preventDefault()` the `indico:confirmed` event (sorry, forgot to mention that)

indico/web/forms/widgets.py
pferreir: Yeah, but then I need to extract from `jsonify_form` the bit that gets the JS from the form. Otherwise how will I handle that server-side? Plus, at least this way the old `<script>` tags go away.

ThiefMaster: Ah, you aren't using jsonify_form.. you'd need to add `js=_pop_injected_js()` to the `jsonify()` data then

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
```diff
@@ -20,6 +20,20 @@
 @import 'partials/boxes';
 @import 'partials/icons';
 
+%rating-bullet {
```
OmeGak: Just an idea, since the file is growing quite a bit. Maybe we could move all this review-related rules to `abstracts/_abstract_reviews` to keep them better organized.

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -51,8 +58,96 @@ class RHAbstractsDownloadAttachment(RHAbstractsReviewBase):
...
     def _checkParams(self, params):
-        RHAbstractsReviewBase._checkParams(self, params)
+        RHAbstractReviewBase._checkParams(self, params)
         self.abstract_file = AbstractFile.get_one(request.view_args['file_id'])
 
     def _process(self):
         return self.abstract_file.send()
+
+
+class RHListOtherAbstracts(RHConferenceBaseDisplay):
```
OmeGak: It's not very obvious what this RH is supposed to be used for. Maybe a docstring wouldn't harm.

indico/modules/events/abstracts/blueprint.py
```diff
@@ -95,6 +99,14 @@
 
 # Abstract-specific management
 _bp.add_url_rule('/manage/abstracts/<int:abstract_id>/', 'manage_abstract', RHManageAbstract, methods=('GET', 'POST'))
-_bp.add_url_rule('/manage/abstracts/<int:abstract_id>/reset',
-                 'reset_abstract_judgment', RHResetAbstractJudgment, methods=('POST',))
-_bp.add_url_rule('/manage/abstracts/<int:abstract_id>/abstract.pdf', 'manage_abstract_pdf_export', RHAbstractExportPDF)
+
+for prefix in ('/manage/abstracts', '/abstracts'):
```
OmeGak: Can't we use the `add_prefixed_rules(...)` context like we do here https://github.com/indico/indico/blob/master/indico/modules/events/timetable/blueprint.py#L55?

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -51,8 +58,96 @@ class RHAbstractsDownloadAttachment(RHAbstractsReviewBase):
...
+        if form.validate_on_submit():
+            if self.review:
+                form.populate_obj(self.review)
+                for question in self.event_new.abstract_review_questions:
+                    rating = question.get_review_rating(self.review)
+                    if not rating:
+                        rating = AbstractReviewRating(question=question, review=self.review)
+                    rating.value = int(form.data["question_{}".format(question.id)])
+            else:
+                self.review = AbstractReview(abstract=self.abstract, track=self.track, user=session.user)
```
OmeGak: I would move all of this into a `review_abstract` function in `operations.py` and add some logging.

indico/modules/events/abstracts/forms.py
```diff
@@ -200,6 +227,34 @@ def __init__(self, *args, **kwargs):
         self.merged_into.excluded_abstract_ids = {abstract.id}
 
 
+class AbstractReviewForm(IndicoForm):
+    """Form for reviewing an abstract"""
+
+    _order = ('comment', 'proposed_action', 'proposed_contribution_type', 'proposed_related_abstract')
+
+    comment = TextAreaField(_("Comments"))
```
OmeGak: `_("Comment")`

OmeGak: OTOH, I've been using `render_kw={'placeholder': _("Leave a comment...")}` to have a placeholder in the `textarea` while hiding the form labels. I think it looks nice since we don't have that many fields in this forms and the labels make them look bloated.

indico/modules/events/abstracts/forms.py
```diff
@@ -200,6 +227,34 @@ def __init__(self, *args, **kwargs):
...
+    proposed_action = IndicoEnumSelectField(
+        _("Proposed Action"), [DataRequired()],
+        enum=AbstractAction, skip={AbstractAction.change_tracks})
```
OmeGak: Mixing indentation types. Any reason we don't go for good ol' aligning to opening delimiter here?

indico/modules/events/abstracts/forms.py
```diff
@@ -200,6 +227,34 @@ def __init__(self, *args, **kwargs):
...
+    proposed_related_abstract = AbstractField(
+        _("Target Abstract"),
+        [HiddenUnless('proposed_action', {AbstractAction.mark_as_duplicate, AbstractAction.merge}), DataRequired()],
+        description=_("The abstract should be marked as duplicate of the selected one"),
```
OmeGak: For consistency, maybe. We refer as "The current abstract" in `AbstractJudgmentFormBase` descriptions.

indico/modules/events/abstracts/templates/abstract/review.html
```diff
@@ -1,30 +1,73 @@
...
                 {% endfor %}
             {% endif %}
         {% endfor %}
     </div>
 {% endmacro %}
 
+{% macro render_review_boxes(abstract, review_forms, management=false) %}
+    {% for track in abstract.reviewed_for_tracks|sort(attribute='title') %}
+        {% set user_reviewed = abstract.get_reviews(track=track)|selectattr('user', 'equalto', session.user)|list %}
+        {% if track.can_review_abstracts(session.user) and not user_reviewed %}
```
OmeGak: I would flip the checks since `user_reviewed` is already calculated.

indico/modules/events/abstracts/templates/forms/rating_widget.html
```diff
@@ -0,0 +1,23 @@
+{% extends 'forms/base_widget.html' %}
+
+{% block html %}
+    <div id="{{ field.id }}" class="rating-widget">
+        <span class="rating-value">{{ '-' if field.object_data is none else field.data }}</span>
+        {% for bullet in range(cfa.rating_range[1], cfa.rating_range[0] - 1, -1) %}
+            {% set radio_id = '%s-%d'|format(field.id, bullet) %}
+            <input type="radio" id="{{ radio_id }}" name="{{ field.name }}" value="{{ bullet }}"
```
OmeGak: `id="bullet-{{ radio_id }}"`?

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -37,4 +37,65 @@
...
+                error: handleAjaxError,
+                data: getFormParams($form),
+                success: function(data) {
+                    if (data.page_html) {
+                        $('.management-page').replaceWith(data.page_html);
+                        $box = $('#abstract-decision-box');
+                    } else {
+                        $box.html(data.box_html);
+                    }
+                    initForms($box);
```
pferreir: We don't need to. Once the abstract is judged, there are no more forms on the page. At least for now.

indico/modules/events/abstracts/blueprint.py
ThiefMaster: No, we couldn't do the management:true|false then

pferreir: No. I looked into it, but I cannot customize `defaults` if I do it that way.

indico/modules/events/abstracts/forms.py
pferreir: > `_("Comment")`

Really? Why?

indico/modules/events/abstracts/forms.py
pferreir: Some of the text lines are just too long. It looks very ugly and broken.

indico/modules/events/abstracts/templates/forms/rating_widget.html
pferreir: And make the name even longer? ID collisions are already highly unlikely.

PR #2673, opened 2016-10-28
Abstract comments

OmeGak:
Implements adding/deleting/editing comments.

## Desirable missing features:
- [ ] Display visibility on comment metadata.
- [ ] Restrict visibility options depending on user rights.

## DB Changes

Don't forget to add this to `DB Changes`:

``` sql
ALTER TABLE event_abstracts.abstract_comments ADD COLUMN visibility smallint DEFAULT 3;
ALTER TABLE event_abstracts.abstract_comments ALTER visibility DROP DEFAULT;
ALTER TABLE event_abstracts.abstract_comments ADD CONSTRAINT ck_abstract_comments_valid_enum_visibility CHECK ((visibility = ANY (ARRAY[1, 2, 3, 4, 5])));
```

### DIFF
--- a/indico/htdocs/js/indico/modules/abstracts.js
+++ b/indico/htdocs/js/indico/modules/abstracts.js
@@ -125,6 +125,8 @@
         }).on('indico:htmlUpdated', function() {
             initForms($(this));
             showFormErrors($(this));
+        }).on('declarative:success', '.js-delete-comment', function() {
+            $(this).parents('.i-timeline-item').remove();
         }).on('click', '#edit-reviewed-for-track-list', function(evt) {
             evt.preventDefault();
             $(this).closest('.i-timeline-item').find('.i-box-footer').toggleClass('hidden');
--- a/indico/htdocs/js/indico/modules/abstracts.js
+++ b/indico/htdocs/js/indico/modules/abstracts.js
@@ -39,11 +39,11 @@
     };
 
     global.setupAbstractPage = function setupAbstractPage() {
-        $('body').on('indico:confirmed', '.comment-button', function(evt) {
+        $('body').on('indico:confirmed', '.js-comment', function(evt) {
             evt.preventDefault();
 
             var $this = $(this);
-            var $box = $('#abstract-comment-box');
+            var $box = $this.parents('.i-timeline-item');
             var $form = $box.find('form');
             var $page = $('.management-page');
 
@@ -59,13 +59,13 @@
                         $page.replaceWith($newPage);
                         $page = $newPage;
                     } else {
-                        $box.html(data.box_html);
+                        $form.replaceWith(data.form_html);
                     }
                     initForms($page);
                     showFormErrors($page);
                 }
             });
-        }).on('indico:confirmed', '.judge-button', function(evt) {
+        }).on('indico:confirmed', '.js-judge', function(evt) {
             var $box = $('#abstract-decision-box');
             var $this = $(this);
             var $form = $box.find('form');
@@ -89,7 +89,7 @@
                     showFormErrors($box);
                 }
             });
-        }).on('indico:confirmed', '.review-button, #save-reviewing-tracks-list', function(evt) {
+        }).on('indico:confirmed', '.js-review, #save-reviewing-tracks-list', function(evt) {
             var $this = $(this);
             var $box = $this.closest('.i-timeline-item');
             var $form = $box.find('form');
@@ -115,18 +115,26 @@
                     showFormErrors($page);
                 }
             });
-        }).on('click', '.change-review-button', function(evt) {
+        }).on('click', '.js-change-review', function(evt) {
             evt.preventDefault();
 
             var $reviewDisplay = $(this).closest('.i-timeline-item');
             var $reviewBox = $reviewDisplay.next('.abstract-review-box');
             $reviewDisplay.hide();
             $reviewBox.show();
+        }).on('declarative:success', '.js-delete-comment', function() {
+            $(this).parents('.i-timeline-item').remove();
+        }).on('declarative:success', '.js-edit-comment', function(evt, data) {
+            var $oldContent = $(this).parents('.i-timeline-item').find('.i-box-content');
+            var $newContent = $(data.form_html);
+            $oldContent.replaceWith($newContent);
+            $newContent.find('.js-edit-comment-cancel').on('indico:confirmed', function(evt_) {
+                evt_.preventDefault();
+                $newContent.replaceWith($oldContent);
+            });
         }).on('indico:htmlUpdated', function() {
             initForms($(this));
             showFormErrors($(this));
-        }).on('declarative:success', '.js-delete-comment', function() {
-            $(this).parents('.i-timeline-item').remove();
         }).on('click', '#edit-reviewed-for-track-list', function(evt) {
             evt.preventDefault();
             $(this).closest('.i-timeline-item').find('.i-box-footer').toggleClass('hidden');
--- a/indico/htdocs/js/indico/modules/abstracts.js
+++ b/indico/htdocs/js/indico/modules/abstracts.js
@@ -39,7 +39,33 @@
     };
 
     global.setupAbstractPage = function setupAbstractPage() {
-        $('body').on('indico:confirmed', '.judge-button', function(evt) {
+        $('body').on('indico:confirmed', '.comment-button', function(evt) {
+            evt.preventDefault();
+
+            var $this = $(this);
+            var $box = $('#abstract-comment-box');
+            var $form = $box.find('form');
+            var $page = $('.management-page');
+
+            $.ajax({
+                url: $this.data('href'),
+                method: $this.data('method'),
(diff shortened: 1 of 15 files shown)


indico/modules/events/abstracts/models/comments.py
```diff
@@ -17,9 +17,19 @@
...
 
 from indico.core.db import db
-from indico.core.db.sqlalchemy import UTCDateTime
+from indico.core.db.sqlalchemy import PyIntEnum, UTCDateTime
 from indico.util.date_time import now_utc
 from indico.util.string import format_repr, return_ascii, text_to_repr
+from indico.util.struct.enum import IndicoEnum
+
+
+class AbstractCommentVisibility(int, IndicoEnum):
```
ThiefMaster: Do we really want the visibility to be that detailed? To me a "private, only visible to judges" seems more practical...

OmeGak: As discussed with @pferreir, we want at least the 3 first ones, with potential use for the fourth. At this point I think there's no harm for having the totally public visibility as well.

pferreir: Right now there are 2 use cases that are pretty clear:
 * Messages between reviewers and conveners/judges (e.g. "This abstract sucks", which shouldn't be seen by other reviewers as those can even be authors of the abstract);
 * Messages between the conveners and the managers (e.g. "The abstract sucks, but Ome's review of it sucks even more. We shouldn't have made him a reviewer.");

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -37,4 +37,67 @@
...
+        }).on('click', '.change-review-button', function(evt) {
+            evt.preventDefault();
+
+            var $reviewDisplay = $(this).closest('.i-timeline-item');
+            var $reviewBox = $reviewDisplay.next('.abstract-review-box');
+            $reviewDisplay.hide();
+            $reviewBox.show();
+        }).on('indico:htmlUpdated', function() {
+            initForms($(this));
+            showFormErrors($(this));
```
nop33: Not a strong argument since it's only 2 lines, but on the other hand... it's 2 lines! :yum: 
```js
var $this = $(this);
```

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
```diff
@@ -212,3 +226,64 @@
...
+    input[type="radio"] {
+        display: none;
+
+        &:checked ~ label.bullet {
+            @extend %rating-bullet-selected;
+        }
+    }
+    .bullet {
+        cursor: pointer;
+        @extend %rating-bullet;
```
nop33: Don't you think it's better to always put the `@extend` and `@include` at the top of the selector's body?

indico/modules/events/abstracts/forms.py
```diff
@@ -196,8 +223,35 @@ def __init__(self, *args, **kwargs):
...
+
+
+class AbstractReviewForm(IndicoForm):
+    """Form for reviewing an abstract"""
+
+    _order = ('comment', 'proposed_action', 'proposed_contribution_type', 'proposed_related_abstract')
+
+    comment = TextAreaField(_("Comments"))
+    proposed_action = IndicoEnumSelectField(
+        _("Proposed Action"), [DataRequired()], enum=AbstractAction,
```
nop33: Y break like dis?

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -426,6 +426,11 @@ def can_access(self, user):
             return True
         return self.can_review(user)
 
+    def can_comment(self, user):
+        if not user:
+            return False
+        return self.can_judge(user) or self.can_convene(user) or self.can_review(user)
```
nop33: Isn't the whole function the same as
```python
return user is not None and self.can_judge(user) or self.can_convene(user) or self.can_review(user)
```

ThiefMaster: Yes, except that it's more readable since it's split in access checks and "not even logged in"

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -79,22 +105,33 @@
...
             var $reviewBox = $reviewDisplay.next('.abstract-review-box');
             $reviewDisplay.hide();
             $reviewBox.show();
+        }).on('declarative:success', '.js-delete-comment', function() {
+            $(this).parents('.i-timeline-item').remove();
+        }).on('declarative:success', '.js-edit-comment', function(evt, data) {
+            var $oldContent = $(this).parents('.i-timeline-item').find('.i-box-content');
+            var $newContent = $(data.form_html);
+            $oldContent.replaceWith($newContent);
+            $newContent.find('.js-edit-comment-cancel').on('indico:confirmed', function(evt_) {
```
OmeGak: I went for `evt_` since the `evt` from the scope above was somehow shadowing this one.

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -39,7 +39,33 @@
...
+                        var $newPage = $(data.page_html);
+                        $page.replaceWith($newPage);
+                        $page = $newPage;
+                    } else {
+                        $form.replaceWith(data.form_html);
+                    }
+                    initForms($page);
+                    showFormErrors($page);
+                }
+            });
```
nop33: The whole block above is the same as the one for `.js-review`, isn't it?

pferreir: We really need to find some nicer way to do this, but I guess we can discuss it after. But if it's identical to `.js-review` then it may be worth either adding it to the selector (or use the same function).

OmeGak: It's not exactly identical.

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -216,6 +218,52 @@ def _process(self):
         return jsonify_data(box_html=tpl.render_review_box(form, self.abstract, self.track, management=self.management))
 
 
+class RHLeaveComment(RHAbstractReviewBase):
+    CSRF_ENABLED = True
```
ThiefMaster: This is not set in the base class yet?

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -216,6 +218,52 @@ def _process(self):
...
+
+    def _process(self):
+        form = AbstractCommentForm()
+        if form.validate_on_submit() or False:
```
ThiefMaster: :monkey:

OmeGak: The debug was strong in me

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -216,6 +218,52 @@ def _process(self):
...
+        RHAbstractReviewBase._checkProtection(self)
+        if not self.comment.can_edit(session.user):
+            raise Forbidden
+
+
+class RHEditAbstractComment(RHAbstractCommentBase):
+    def _process(self):
+        form = AbstractCommentForm(obj=self.comment)
+        if form.validate_on_submit():
+            edit_abstract_comment(self.comment, form.data)
```
ThiefMaster: isn't it `update_...` everywhere else?

indico/modules/events/abstracts/blueprint.py
```diff
@@ -131,6 +134,13 @@
 _bp.add_url_rule('/manage/abstracts/<int:abstract_id>/', 'manage_abstract', RHManageAbstract, methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/abstracts/<int:abstract_id>/abstract.pdf', 'manage_abstract_pdf_export', RHAbstractExportPDF)
 
+# Abstract-specific
+_bp.add_url_rule('/manage/abstracts/<int:abstract_id>/comment', 'comment_abstract', RHLeaveComment, methods=('POST',))
+_bp.add_url_rule('/manage/abstracts/<int:abstract_id>/comments/<int:comment_id>/edit',
+                 'edit_abstract_comment', RHEditAbstractComment, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/abstracts/<int:abstract_id>/comments/<int:comment_id>/delete',
+                 'delete_abstract_comment', RHDeleteAbstractComment, methods=('POST',))
```
ThiefMaster: Why not `/manage/abstracts/<int:abstract_id>/comments/<int:comment_id>` for both edit and delete and the DELETE method here?

OmeGak: Not a strong opinion. How would you call the RH then?

ThiefMaster: I usually go for whatever_rest / RHWhateverREST since it's pretty much a RESTful endpoint. But in this case I wouldn't rename it. It's still only for deletion (you don't need to merge the two RHs!)

PR #2675, opened 2016-11-01
Submit abstract from event page

nop33:


indico/modules/events/abstracts/__init__.py
```diff
@@ -118,3 +118,6 @@ def _is_visible(conf):
                         position=9, visible=lambda event: event.has_feature('abstracts'), static_site=True)
     yield MenuEntryData(title=_("My Abstracts"), name='user_abstracts', visible=_is_visible,
                         endpoint='abstracts.my_abstracts', position=0, parent='call_for_abstracts')
+    yield MenuEntryData(title=_("Submit Abstract"), name='abstract_submission',
+                        visible=lambda event: event.has_feature('abstracts'),
```
ThiefMaster: Should we show this if we are outside the submission period? Either we don't or we show a message on the page that the submission is not open (yet)

ThiefMaster: Should we show this if we are outside the submission period? Either we don't or we show a message on the page that the submission is not open (yet)

indico/modules/events/abstracts/controllers/display.py
```diff
@@ -76,3 +81,26 @@ def _process(self):
...
         return send_file('my-abstracts.pdf', pdf.generate(), 'application/pdf')
+
+
+class RHSubmitAbstract(RHMyAbstractsBase):
+    def _process(self):
+        abstract_form_class = make_abstract_form(self.event_new)
+        form = abstract_form_class(event=self.event_new)
+        if form.validate_on_submit():
+            data = form.data
+            if isinstance(data['submitted_for_tracks'], Track):
```
ThiefMaster: We do this in the  management RH too. Should we move this logic into `create_abstract`?

indico/modules/events/abstracts/controllers/display.py
```diff
@@ -76,3 +81,26 @@ def _process(self):
...
+                data['submitted_for_tracks'] = {data['submitted_for_tracks']}
+            elif data['submitted_for_tracks'] is None:
+                data['submitted_for_tracks'] = set()
+            abstract = create_abstract(self.event_new, *get_field_values(data))
+            flash(_("Your abstract with title '{}' has been successfully submitted. It is registered with the number "
```
ThiefMaster: > *the* title

And maybe we can reword this a bit? The full message sounds a bit strange to me, but I don't have a better suggestion right now.

indico/modules/events/abstracts/__init__.py
nop33: I think the second approach is better, otherwise people might be wondering why they can't find how to submit an abstract.

indico/modules/events/abstracts/controllers/display.py
nop33: I only used our old message, I am open to other suggestions. Is it really "with **the** title"? It sounds fine to me to say "Your abstract with title", maybe because of Greek! :yum:

indico/modules/events/abstracts/__init__.py
nop33: We could have a message "The call for abstracts is closed. Submission period: {date1} to {date2}". This way we don't have to check whether we have to put "yet" or anything else, it covers all possible cases.

nop33: The above was implemented.

indico/modules/events/abstracts/operations.py
```diff
@@ -29,13 +29,18 @@
...
+from indico.modules.events.tracks.models.tracks import Track
 from indico.modules.events.util import set_custom_fields
 from indico.util.date_time import now_utc
 from indico.util.fs import secure_filename
 
 
 def create_abstract(event, abstract_data, custom_fields_data=None):
     abstract = Abstract(event_new=event, submitter=session.user)
+    if isinstance(abstract_data['submitted_for_tracks'], Track):
+        abstract_data['submitted_for_tracks'] = {abstract_data['submitted_for_tracks']}
```
ThiefMaster: Instead of modifying the dict passed to the function I'd rather assign `tracks = ...` here and then just set `abstract.submitted_for_tracks = tracks` after the `populate_from_dict()`

indico/modules/events/abstracts/templates/display/submit_abstract.html
```diff
@@ -0,0 +1,28 @@
...
+{% from 'forms/_form.html' import form_header, form_footer, form_rows, form_fieldset %}
+
+{% extends 'layout/conference_page_base.html' %}
+
+{% block title -%}
+    {% trans %}Submit abstract{% endtrans %}
+{%- endblock %}
+
+{%- block content %}
+    {% if cfa.is_open %}
```
ThiefMaster: I'd move this part to a separate template (and render it before even doing form stuff in the RH if the CFA is not active). Right now you could circumvent the is_open condition by simply submitting the form with httpie, curl, etc since there is no check in the RH.

ThiefMaster: Also, checking is_open is not enough - people can be authorized to submit abstracts outside that period if this check returns True: `abstracts_settings.acls.contains_user(event, 'authorized_submitters', user)`

ThiefMaster: Maybe add a `can_submit_abstracts(user)` method to the CFA proxy that checks is_open and whether the user is authorized to submit abstracts at any time

indico/modules/events/abstracts/templates/display/submit_abstract.html
```diff
@@ -0,0 +1,28 @@
...
+        {{ form_header(form, action=url_for('abstracts.submit', event), id="submit-abstract-form") }}
```
ThiefMaster: Looks like you could use `simple_form` here)

indico/modules/events/abstracts/templates/display/submit_abstract.html
```diff
@@ -0,0 +1,28 @@
...
+        <div class="info-message-box">
+            <div class="message-text">
+                {%- trans start_dt=cfa.start_dt|format_human_date, end_dt=cfa.end_dt|format_human_date -%}
+                    The call for abstracts is closed. Submission period: {{ start_dt }} to {{ end_dt }}
+                {%- endtrans -%}
+            </div>
+        </div>
+    {% endif %}
+{% endblock %}
+
```
ThiefMaster: extra empty line

indico/modules/events/abstracts/controllers/display.py
```diff
@@ -76,3 +80,22 @@ def _process(self):
...
+    def _process(self):
+        abstract_form_class = make_abstract_form(self.event_new)
+        form = abstract_form_class(event=self.event_new)
+        if form.validate_on_submit():
+            data = form.data
+            abstract = create_abstract(self.event_new, *get_field_values(data))
+            flash(_("Your abstract with title '{}' has been successfully submitted. It is registered with the number "
+                    "{}. You will be notified by email with the submission details.")
+                  .format(abstract.title, abstract.friendly_id), 'success')
+            url = url_for('abstracts.my_abstracts', self.event_new)
```
ThiefMaster: `.my_abstracts` is enough

indico/modules/events/abstracts/templates/display/submit_abstract.html
ThiefMaster: `'.submit'` is enough; it's in the same blueprint

indico/modules/events/abstracts/templates/display/submit_abstract.html
nop33: Why not something like this
```python
class RHSubmitAbstract(RHMyAbstractsBase):
    def _process(self):
        form = None
        if self.event_new.cfa.can_submit_abstracts(session.user):
            abstract_form_class = make_abstract_form(self.event_new)
            form = abstract_form_class(event=self.event_new)
            if form.validate_on_submit():
                data = form.data
                abstract = create_abstract(self.event_new, *get_field_values(data))
                flash(_("Your abstract with title '{}' has been successfully submitted. It is registered with the "
                        "number #{}. You will be notified by email with the submission details.")
                      .format(abstract.title, abstract.friendly_id), 'success')
                url = url_for('.my_abstracts', self.event_new)
                if request.is_xhr:
                    return jsonify_data(flash=False, redirect=url)
                else:
                    return redirect(url)
        return WPSubmitAbstract.render_template('display/submit_abstract.html', self._conf, event=self.event_new,
                                                form=form)
```
and in the template simply:
```
{% if event.cfa.can_submit_abstracts(session.user) %}
    {{ simple_form(form, form_header_kwargs={'action': url_for('.submit', event)}) }}
{% else %}
    ...
```

nop33: or even
```
{% if form %}
    {{ simple_form(form, form_header_kwargs={'action': url_for('.submit', event)}) }}
{% else %}
    ...
```

ThiefMaster: That would work too, but with a separate template you can just render it if the initial `if` fails and save one level of indentation of the whole form handling part. Just feels cleaner to me that way...

(of course you could also render the same template up there instead of using a separate one)

nop33: So basically what you're saying is that I would `return WPSubmitAbstract.render_template(...` twice, once `if not self.event_new.cfa.can_submit_abstracts(session.user):` and once at the very end...

nop33: imo it's better to just do 
```python
if not self.event_new.cfa.can_submit_abstracts(session.user):
    return WPSubmitAbstract.render_template('display/submit_abstract.html', self._conf, event=self.event_new,
                                            form=None)
```
at the top of the PR and just `{% if form %}` in the template..

PR #2681, opened 2016-11-02
Manage reviewed_for_tracks list

nop33:
* [x] Figure out why after a review submission the list of tracks loses its defaults
* [x] Clean up macros and extract common bits to `_common.html`

### DIFF
--- a/indico/modules/events/abstracts/blueprint.py
+++ b/indico/modules/events/abstracts/blueprint.py
@@ -48,7 +48,8 @@ from indico.modules.events.abstracts.controllers.reviewing import (RHAbstractsDo
                                                                    RHDisplayAbstractsExportPDF,
                                                                    RHDisplayAbstractsExportCSV,
                                                                    RHDisplayAbstractsExportExcel,
-                                                                   RHDisplayAbstractsDownloadAttachments)
+                                                                   RHDisplayAbstractsDownloadAttachments,
+                                                                   RHEditReviewedForTrackList)
 from indico.web.flask.util import make_compat_redirect_func
 from indico.web.flask.wrappers import IndicoBlueprint
 
@@ -131,7 +132,6 @@ _bp.add_url_rule('/manage/abstracts/email-templates/<email_tpl_id>', 'email_tpl_
 _bp.add_url_rule('/manage/abstracts/<int:abstract_id>/', 'manage_abstract', RHManageAbstract, methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/abstracts/<int:abstract_id>/abstract.pdf', 'manage_abstract_pdf_export', RHAbstractExportPDF)
 
-# Abstract reviewing actions
 for prefix in ('/manage/abstracts', '/abstracts'):
     defaults = {'management': prefix != '/abstracts'}
     _bp.add_url_rule(prefix + '/<int:abstract_id>/withdraw', 'withdraw_abstract',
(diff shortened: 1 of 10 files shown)


indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -179,3 +179,15 @@ def _process(self):
             return jsonify_data(page_html=render_abstract_page(self.abstract, management=self.management))
         tpl = get_template_module('events/abstracts/abstract/review.html')
         return jsonify_data(box_html=tpl.render_review_box(form, self.abstract, self.track, management=self.management))
+
+
+class RHEditReviewedForTrackList(RHAbstractReviewBase):
+    CSRF_ENABLED = True
```
ThiefMaster: This is (or at least should be) set in the base class

indico/modules/events/abstracts/templates/abstract/review.html
```diff
@@ -229,3 +229,49 @@
         {{ render_review_box(form, review.abstract, review.track, edit=true, management=management) }}
     {% endif %}
 {% endmacro %}
+
+{% macro render_reviewed_for_tracks_box(abstract, form, management=false) %}
+    <div id="abstract-reviewed-for-tracks-box"
+         class="i-timeline-item">
```
ThiefMaster: fits in the previous line

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -179,3 +179,15 @@ def _process(self):
...
+
+    def _process(self):
+        form = AbstractReviewedForTracksForm(event=self.event_new)
+        if form.validate_on_submit:
+            form.populate_obj(self.abstract)
```
ThiefMaster: I'd rather add an operations.py function for it so we can log it etc

indico/modules/events/abstracts/forms.py
```diff
@@ -415,3 +415,13 @@ class AbstractsScheduleForm(IndicoForm):
     def __init__(self, *args, **kwargs):
         self.event = kwargs.pop('event')
         super(AbstractsScheduleForm, self).__init__(*args, **kwargs)
+
+
+class AbstractReviewedForTracksForm(IndicoForm):
+    reviewed_for_tracks = IndicoQuerySelectMultipleCheckboxField(_("Tracks"), get_label=lambda x: x.title,
```
ThiefMaster: `get_label='title'` is enough

ThiefMaster:
#2667 could probably be assimilated into this PR. It's semi-related (and the code in there also needs to be updated to handle tracks properly, e.g. update reviewed_for_tracks and not just submitted_for_tracks when editing an unreviewed abstract)

indico/modules/events/abstracts/blueprint.py
```diff
@@ -119,6 +119,8 @@
                      RHReviewAbstractForTrack, methods=('POST',), defaults=defaults)
     _bp.add_url_rule(prefix + '/<int:abstract_id>/judge', 'judge_abstract',
                      RHJudgeAbstract, methods=('POST',), defaults=defaults)
+    _bp.add_url_rule(prefix + '/<int:abstract_id>/review/tracks', 'edit_review_tracks',
```
nop33: I thought that since it's related with the **reviewed**_for_tracks, it could use the `/review/` path.

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -179,3 +179,15 @@ def _process(self):
...
+            return jsonify_data(page_html=render_abstract_page(self.abstract, management=self.management))
+        tpl = get_template_module('events/abstracts/abstract/review.html')
+        return jsonify_data(box_html=tpl.render_reviewed_for_tracks_box(self.abstract, form=form, management=self.management))
```
pferreir: I think that after we merge everything we should try to look for a more elegant solution for this... it's a pattern that is used in at least 2 other places (same on client side).

indico/modules/events/abstracts/blueprint.py
nop33: Do we want both a management and a display endpoint, or just a management one?

ThiefMaster: Both I guess

nop33: OK. Then how about `/<int:abstract_id>/reviewing-tracks`?

ThiefMaster: `tracks` or `reviewing-tracks` - no strong opinion on which one

PR #2682, opened 2016-11-03
Merge abstract persons properly

pferreir:


indico/modules/events/abstracts/operations.py
```diff
@@ -81,6 +83,28 @@ def delete_abstract(abstract, delete_contrib=False):
                            'Abstract "{}" has been deleted'.format(abstract.title), session.user)
 
 
+def merge_person_links(target_abstract, source_abstract):
+    """Add to `target_abstract` new `AbstractPersonLink`s whose `EventPerson`
```
ThiefMaster: """
    Add to `target_abstract` new `AbstractPersonLink`s whose `EventPerson`
    exists in the `source_abstract` but is not yet in the `target_abstract`.

    :param target_abstract: The target abstract (to which the links should then be added)
    :param source_abstract: The source abstract
    """

The first line only goes in the `"""` line if it fits in one line and the other lines aren't indented according to [PEP-257](https://www.python.org/dev/peps/pep-0257/)

indico/modules/events/abstracts/operations.py
```diff
@@ -81,6 +83,28 @@ def delete_abstract(abstract, delete_contrib=False):
...
+    source_link_map = defaultdict(set)
+    for link in source_links:
+        source_link_map[link.person.id].add(link)
+    unique_persons = {link.person for link in source_links} - {link.person for link in target_abstract.person_links}
+
+    for person in unique_persons:
+        for source_link in source_link_map[person.id]:
+            link = AbstractPersonLink(person=person, author_type=source_link.author_type,
+                                      is_speaker=source_link.is_speaker)
+            target_abstract.person_links.append(link)
```
ThiefMaster: How are we dealing with the sorting in case they weren't set to use alphabetical order?

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -77,10 +77,13 @@ def _checkParams(self, params):
     def _process(self):
         query = (Abstract
                  .query.with_parent(self.event_new)
+                 .filter(Abstract.state.notin_({AbstractState.duplicate, AbstractState.merged}))
```
ThiefMaster: We can't have abstract A being a duplicate of abstract B when abstract B is a duplicate of abstract C?

pferreir: I think they should be both duplicate of C, right? Especially if B has already been marked as duplicate of it.

ThiefMaster: Makes more sense for sure, just wondering if users realize that they should do it like this

indico/modules/events/abstracts/operations.py
ThiefMaster: I'd check if the target abstract had a custom order - if yes, max(position) + 1, otherwise I'd keep 0 (ie alphabetical order) even if some of the source abstracts had a custom order there.

PR #2684, opened 2016-11-04
Fix unicode issues generating MathJax JS

pferreir:


indico/modules/events/abstracts/views.py
```diff
@@ -64,9 +69,7 @@ def _getBody(self, params):
         return WPJinjaMixin._getPageContent(self, params).encode('utf-8')
 
     def _getHeadContent(self):
-        return (WPConferenceDefaultDisplayBase._getHeadContent(self) + render('js/mathjax.config.js.tpl') +
-                '\n'.join('<script src="{0}" type="text/javascript"></script>'.format(url)
-                          for url in self._asset_env['mathjax_js'].urls()))
+        return WPConferenceDefaultDisplayBase._getHeadContent(self).decode('utf-8') + MathjaxMixin._getHeadContent(self)
```
ThiefMaster: `to_unicode()` so it doesn't break if we get unicode back in some cases?

indico/modules/events/abstracts/views.py
```diff
@@ -21,7 +21,14 @@
 from MaKaC.webinterface.pages.conferences import WPConferenceDefaultDisplayBase, WPConferenceModifBase
 
 
-class WPManageAbstracts(WPJinjaMixin, WPConferenceModifBase):
+class MathjaxMixin:
+    def _getHeadContent(self):
+        return (render('js/mathjax.config.js.tpl') +
+                u'\n'.join(u'<script src="{0}" type="text/javascript"></script>'.format(url)
```
ThiefMaster: no need for the `u` prefix, the file should have `unicode_literals` enabled

ThiefMaster:
Merged 894e265

PR #2686, opened 2016-11-04
Link accepted abstracts to contributions

nop33:


indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -156,6 +157,13 @@
                                         {{- friendly_data if friendly_data else '-' -}}
                                     </td>
                                 {% endfor %}
+                                <td class="i-table">
+                                    {% if abstract.state.name == 'accepted' and abstract.contribution%}
```
ThiefMaster: space before `%}`.

Also check if you need to add `subqueryload('contribution')` to the query options to avoid spamming one query per accepted abstract

indico/modules/events/contributions/controllers/base.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class ContributionListMixin:
+    """Display list of contributions"""
+
+    view_class = None
+    template = None
+
+    def _process(self):
+            if self.list_generator.static_link_used:
```
ThiefMaster: indentation

indico/modules/events/abstracts/operations.py
```diff
@@ -123,8 +123,7 @@ def judge_abstract(abstract, abstract_data, judgment, judge, contrib_session=Non
         abstract.accepted_track = abstract_data['accepted_track']
         abstract.accepted_contrib_type = abstract_data['accepted_contrib_type']
         if not abstract.contribution:
-            # TODO: generate contribution
-            pass
+            create_contribution_from_abstract(abstract)
```
ThiefMaster: you forgot to pass `contrib_session`

indico/modules/events/contributions/operations.py
```diff
@@ -153,3 +155,30 @@ def delete_subcontribution(subcontrib):
...
+
+
+@no_autoflush
+def create_contribution_from_abstract(abstract, contrib_session=None):
+    event = abstract.event_new
+    contrib_person_links = {}
+    person_link_attrs = {'_title', 'address', 'affiliation', 'first_name', 'last_name', 'phone', 'author_type',
+                         'person', 'is_speaker'}
+    for abstract_person_link in abstract.person_links:
+        contrib_person_link_data = {}
```
ThiefMaster: link = ContributionPersonLink()
    link.populate_from_attrs(abstract_person_link, person_link_attrs)

indico/modules/events/contributions/operations.py
```diff
@@ -153,3 +155,30 @@ def delete_subcontribution(subcontrib):
...
+                          field_value in abstract.field_values}
+    create_contribution(event, {'friendly_id': abstract.friendly_id,
+                                'title': abstract.title,
+                                'duration': duration,
+                                'description': abstract.description,
+                                'abstract': abstract,
+                                'type': abstract.accepted_contrib_type,
+                                'track': abstract.accepted_track,
+                                'session': contrib_session,
+                                'person_link_data': {l: True for l in contrib_person_links.itervalues()}},
```
ThiefMaster: lowercase-L should not be used as a variable name in python (just like uppercase-I). `link`?

indico/modules/events/contributions/operations.py
```diff
@@ -153,3 +155,30 @@ def delete_subcontribution(subcontrib):
...
+def create_contribution_from_abstract(abstract, contrib_session=None):
+    event = abstract.event_new
+    contrib_person_links = {}
+    person_link_attrs = {'_title', 'address', 'affiliation', 'first_name', 'last_name', 'phone', 'author_type',
+                         'person', 'is_speaker'}
+    for abstract_person_link in abstract.person_links:
+        contrib_person_link_data = {}
+        for attr in person_link_attrs:
+            contrib_person_link_data[attr] = getattr(abstract_person_link, attr)
+        contrib_person_links[abstract_person_link.person.id] = ContributionPersonLink(**contrib_person_link_data)
```
ThiefMaster: Why the dict by person ID? You cannot have the same person more than once in an abstract.

indico/modules/events/contributions/controllers/base.py
nop33: strangely enough, the linter didn't complain...

indico/modules/events/contributions/operations.py
nop33: oh, why is that? I named it `l` to avoid a second line

ThiefMaster: `l` and `I` look the same in some fonts and very similar on others

indico/modules/events/contributions/operations.py
```diff
@@ -153,3 +155,29 @@ def delete_subcontribution(subcontrib):
...
+
+@no_autoflush
+def create_contribution_from_abstract(abstract, contrib_session=None):
+    event = abstract.event_new
+    contrib_person_links = set()
+    person_link_attrs = {'_title', 'address', 'affiliation', 'first_name', 'last_name', 'phone', 'author_type',
+                         'person', 'is_speaker'}
+    for abstract_person_link in abstract.person_links:
+        link = ContributionPersonLink()
+        link.populate_from_attrs(abstract_person_link, person_link_attrs)
```
ThiefMaster: Set the person first (pass it in the constructor). Setting the other attributes requires a proper person association (because of the `override_attr` logic used in person links).

indico/modules/events/contributions/operations.py
```diff
@@ -153,3 +155,29 @@ def delete_subcontribution(subcontrib):
...
+        contrib_person_links.add(link)
+
+    duration = contrib_session.default_contribution_duration if contrib_session else timedelta(minutes=15)
+    custom_fields_data = {'custom_{}'.format(field_value.contribution_field.id): field_value.data for
+                          field_value in abstract.field_values}
+    create_contribution(event, {'friendly_id': abstract.friendly_id,
```
ThiefMaster: I would actually return the contribution here, not pass the `abstract` in the data dict and then assign it when calling this method:

    abstract.contribution = create_contribution_from_abstract(abstract, contrib_session)

That way the operations are less entangled with each other.

ThiefMaster:
> Create ContributionListMixin

Do we really want this at the moment? If we end up not adding a per-track contribution list we just added more complexity. I'd rather keep this commit in a separate branch in case we need it in the future but not merge it yet. If the `selected_entry` logic fixes some other issue I'd keep that part in here though.

indico/htdocs/sass/partials/_tables.scss
```diff
@@ -43,12 +43,27 @@
     }
 }
 
+@mixin i-table() {
```
ThiefMaster: In most mixins without arguments we don't have the `()`

indico/modules/events/abstracts/util.py
```diff
@@ -124,7 +124,8 @@ def _build_query(self):
                          subqueryload('submitted_for_tracks'),
                          subqueryload('reviewed_for_tracks'),
                          subqueryload('person_links'),
-                         subqueryload('reviews').joinedload('ratings'))
+                         subqueryload('reviews').joinedload('ratings'),
+                         subqueryload('contribution'))
```
ThiefMaster: I would use `joinedload('contribution').load_only('id', 'event_id')` - you don't need anything but the IDs from the contribution since all you do is building a URL, and that way 1) don't query everything and 2) don't use a seperate query (which is the case with `subqueryload`) to get two extra columns (the reason why some of the others are `subqueryload` is that they are relationships containing multiple rows - the ones where you get zero or one extra row are all `joinedload`)

indico/htdocs/sass/partials/_tables.scss
OmeGak: Any reason to use a mixin and not a placeholder class? I think the latter fits better cases like this one.

nop33:
> I'd rather keep this commit in a separate branch in case we need it in the future but not merge it yet. If the selected_entry logic fixes some other issue I'd keep that part in here though.

When implementing the `ContributionListMixin` I noticed that the `selected_entry` logic was in the wrong place. I'd rather keep the `ContributionListMixin` and maybe use it for the "My Contributions" page? It looks ugly the way it is now anyway.

nop33:
Hm, nvm my last comment, the "My Contributions" page should look more like the "My Abstracts" page I thing rather than a table list of contributions.

indico/htdocs/sass/partials/_tables.scss
```diff
@@ -43,12 +43,27 @@
     }
 }
 
+@mixin i-table {
```
OmeGak: Any reason to use a mixin and not a placeholder class? I think the latter fits better cases like this one.

indico/modules/events/contributions/operations.py
```diff
@@ -153,3 +155,28 @@ def delete_subcontribution(subcontrib):
...
+    event = abstract.event_new
+    contrib_person_links = set()
+    person_link_attrs = {'_title', 'address', 'affiliation', 'first_name', 'last_name', 'phone', 'author_type',
+                         'is_speaker'}
+    for abstract_person_link in abstract.person_links:
+        link = ContributionPersonLink(person=abstract_person_link.person)
+        link.populate_from_attrs(abstract_person_link, person_link_attrs)
+        contrib_person_links.add(link)
+
+    duration = contrib_session.default_contribution_duration if contrib_session else timedelta(minutes=15)
```
nop33: Why somewhere else?
When we create contribs from the UI, they take a duration of 20min.
I found this line of code in the new-old code for converting an abstract to a contribution.

OmeGak: Exactly my point. So now we have different defaults values depending on how the contribution gets generated. :\

indico/modules/events/contributions/controllers/management.py
```diff
@@ -28,6 +28,7 @@
 from indico.modules.events.abstracts.forms import AbstractContentSettingsForm
 from indico.modules.events.abstracts.settings import abstracts_settings
 from indico.modules.events.contributions import get_contrib_field_types
+from indico.modules.events.contributions.controllers.base import ContributionListMixin
```
ThiefMaster: `s/base/common/`? I'd rather use `base.py` for classes that are either bases or generic (not specific to one feature) mixins

indico/modules/events/contributions/operations.py
```diff
@@ -16,6 +16,7 @@
 
 from __future__ import unicode_literals
 
+from datetime import timedelta
```
ThiefMaster: stdlib -> empty line after it

indico/modules/events/contributions/operations.py
```diff
@@ -153,3 +155,28 @@ def delete_subcontribution(subcontrib):
...
     subcontrib.event_new.log(EventLogRealm.management, EventLogKind.negative, 'Subcontributions',
                              'Subcontribution "{}" has been deleted'.format(subcontrib.title), session.user)
+
+
+@no_autoflush
+def create_contribution_from_abstract(abstract, contrib_session=None):
+    event = abstract.event_new
+    contrib_person_links = set()
+    person_link_attrs = {'_title', 'address', 'affiliation', 'first_name', 'last_name', 'phone', 'author_type',
+                         'is_speaker'}
```
ThiefMaster: `display_order` is missing here

indico/modules/events/contributions/operations.py
nop33: I replicated the code I found. Do we want to keep this feature of creating contributions of 15min if they come from abstracts and 20min if they are manually created? If not, I'll update the code.

indico/modules/events/contributions/controllers/management.py
nop33: I don't understand. The `abstracts/controllers/base.py` is full of mixins, including the `DisplayAbstractListMixin` which has essentially the same functionality as the `ContributionListMixin` in `contributions/controllers/base.py`.

ThiefMaster: I moved those in my PR

indico/modules/events/contributions/operations.py
ThiefMaster: I think having a single default duration somewhere (or a setting.. maybe even be a contribution type option, since that's what often determines the duration of the contribution) would be much cleaner, but doesn't need to be done in this PR. It's more a nice-to-have thing.

indico/modules/events/contributions/controllers/management.py
nop33: and put them where? In `controllers/common.py`?

ThiefMaster: yep

PR #2687, opened 2016-11-07
Hide reviews from abstract PDF export depending on role

mvidalgarcia:
Adds a dropdown menu for users that can see the reviews to choose whether they want to export the PDF with or without reviews.

- [x] Fix bug when deleting comments in abstract review

### DIFF
--- a/indico/modules/events/abstracts/templates/abstract/public.html
+++ b/indico/modules/events/abstracts/templates/abstract/public.html
@@ -298,12 +298,32 @@
                         </a>
                     {% endif %}
                 </div>
+            {% endif %}
+            {% if abstract.can_see_reviews(session.user) %}
                 <div class="group">
-                    {% set endpoint = '.manage_abstract_pdf_export' if management else '.display_abstract_pdf_export' %}
-                    <a class="i-button icon-file-download"
-                       href="{{ url_for(endpoint, abstract) }}"
+                    <a class="i-button arrow icon-file-pdf js-dropdown-abstract-pdf" data-toggle="dropdown"></a>
+                    <ul class="dropdown">
+                        <li>
+                            <a href="{{ url_for('.display_abstract_pdf_export', abstract) }}">PDF</a>
+                        </li>
+                        <li>
+                            <a href="{{ url_for('.manage_abstract_pdf_export', abstract) }}">
+                                {%- trans %}PDF with reviews{% endtrans -%}
+                            </a>
+                        </li>
+                    </ul>
+                </div>
+                <script>
+                    $('.js-dropdown-abstract-pdf').parent().dropdown();
+                </script>
+            {% else %}
+                <div class="group">
+                    <a class="i-button icon-file-pdf"
+                       href="{{ url_for('.display_abstract_pdf_export', abstract) }}"
                        title="{% trans %}Download abstract submission in PDF{% endtrans %}"></a>
                 </div>
+            {% endif %}
+            {% if can_manage or abstract.user_owns(session.user) %}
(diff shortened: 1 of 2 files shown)


indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -301,12 +301,30 @@ <h3 class="abstract-title">
...
+                    <a class="i-button arrow icon-file-download dropdown-abstract-pdf"
+                       data-toggle="dropdown"></a>
+                    <ul class="dropdown">
+                        <li>
+                            <a href="{{ url_for('.display_abstract_pdf_export', abstract) }}"
+                               class="icon-file-pdf">PDF</a>
+                        </li>
+                        <li>
+                            <a href="{{ url_for('.manage_abstract_pdf_export', abstract) }}"
+                               class="icon-file-pdf">{% trans %}PDF with reviews{% endtrans %}</a>
```
mvidalgarcia: Not pretty sure about these names and whether they should have the PDF icon.

ThiefMaster: Maybe having no icons on the dropdown items would be the best option? All options are for PDF files anyway.

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -39,6 +39,7 @@
     };
 
     global.setupAbstractPage = function setupAbstractPage() {
+        $('.dropdown-abstract-pdf').parent().dropdown();
```
mvidalgarcia: I haven't thought about that. I've just checked again and the dropdown stops working when e.g somebody leaves a comment. Do you know how can I solve it?

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -154,6 +154,10 @@ def _process(self):
 
 
 class RHAbstractExportPDF(RHManageAbstractBase):
+    def _checkProtection(self):
```
OmeGak: I don't think we need to add this check here. We are extending rom `RHManageAbstractBase`. Doesn't that checks only managers have access? Why wouldn't a manager be able to see reviews?

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -479,4 +497,7 @@ <h3 class="abstract-title">
             </a>
         {% endif %}
     </div>
+    <script>
+        $('.dropdown-abstract-pdf').parent().dropdown();
```
OmeGak: I think we have some declarative way of making the dropdown elements automatically initialized.

mvidalgarcia: Yes, we have `.js-dropdown` but it doesn't work when there are AJAX reloads in the page as I mention in a comment before. If you know how to solve it please let me know.

ThiefMaster: Declarative doesn't run for ajax updates

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -301,12 +301,30 @@ <h3 class="abstract-title">
                         </a>
                     {% endif %}
                 </div>
+            {% endif %}
+            {% if abstract.can_see_reviews(session.user) %}
+                <div class="group">
+                    <a class="i-button arrow icon-file-download js-dropdown-abstract-pdf"
```
ThiefMaster: I'd go for `icon-file-pdf`. We also use this in contributions for example.

PR #2688, opened 2016-11-07
Remove separator before empty section in configure roles qbubble when somebody is speaker

mvidalgarcia:


indico/htdocs/js/indico/widgets/person_link_widget.js
```diff
@@ -249,7 +249,6 @@
                             $field.principalfield('set', person.id, {isSpeaker: !person.isSpeaker});
                         });
                     } else {
-                        $buttons.append($buttonsSeparator);
```
OmeGak: Doesn't this remove the separator in cases where we actually want it?

ThiefMaster: I couldn't find a case where it wasn't working fine

ThiefMaster: Actually, now that I look at it again, it seem like it could break when you have the widget with no speakers but submitters enabled.

It's an unlikely case though and I don't see why we would have such a combination...

PR #2691, opened 2016-11-07
Hide abstract submission menu when CFA is closed

mvidalgarcia:


indico/modules/events/abstracts/controllers/display.py
```diff
@@ -90,8 +90,8 @@ def _process(self):
 class RHSubmitAbstract(RHDisplayAbstractsBase):
     def _process(self):
         if not self.event_new.cfa.can_submit_abstracts(session.user):
-            return WPSubmitAbstract.render_template('display/submit_abstract.html', self._conf, event=self.event_new,
-                                                    form=None)
+            raise Forbidden(response=redirect(url_for('event.conferenceCFA', self.event_new)))
```
ThiefMaster: just `return redirect(...)`

indico/modules/events/abstracts/__init__.py
```diff
@@ -134,6 +134,11 @@ def _reviewing_area_visible(conf):
             return False
         return bool(get_user_tracks(conf.as_event, session.user))
 
+    def _submit_abstract_visible(conf):
+        if not session.user or not conf.has_feature('abstracts'):
```
OmeGak: Maybe we should redirect to the login when `not session.user`? Otherwise it may look to non-indico users like they can't submit an abstract.

indico/modules/events/abstracts/controllers/display.py
```diff
@@ -90,8 +90,8 @@ def _process(self):
 class RHSubmitAbstract(RHDisplayAbstractsBase):
     def _process(self):
         if not self.event_new.cfa.can_submit_abstracts(session.user):
-            return WPSubmitAbstract.render_template('display/submit_abstract.html', self._conf, event=self.event_new,
-                                                    form=None)
+            return redirect(url_for('event.conferenceCFA', self.event_new))
```
OmeGak: Is this the correct endpoint? Looks like legacy to me.

indico/modules/events/abstracts/__init__.py
ThiefMaster: Then I wouldn't hide the menu item at all if the submission period is active or in the future. It's just confusing if you see something while not logged in but not when logged in.

indico/modules/events/abstracts/controllers/display.py
mvidalgarcia: I think we're still using legacy for the CFA display page, but not sure.

indico/modules/events/abstracts/__init__.py
OmeGak: OTOH... We should give enough information in the general CFA page about the possibility of submitting abstracts, so maybe this logic is fine.

ThiefMaster: I wonder.. do we actually need a separate menu item to submit an abstract?

OmeGak: I would say no. We had this nested menu items for submitting a registration before and we went for a single page that does everything. Maybe we should do the same here.

pferreir: Agreed. One page should be enough,

indico/modules/events/abstracts/controllers/display.py
pferreir: @mvidalgarcia, that's what @ThiefMaster means. Maybe this is the moment to stop using legacy 😉

ThiefMaster:
Since we decided to get rid of the menu item this PR is obsolete

PR #2692, opened 2016-11-07
Refactor abstract RH inheritance

ThiefMaster:
Diff is quite awful, moved lots of things around. Better to look at (some of) the separate commits


indico/modules/events/abstracts/controllers/management.py
```diff
@@ -102,7 +79,7 @@ def _checkParams(self, params):
         self.abstracts = self._abstract_query.filter(Abstract.id.in_(ids)).all()
 
 
-class RHManageAbstract(AbstractPageMixin, RHConferenceModifBase):
+class RHManageAbstract(AbstractPageMixin, RHConferenceModifBase):  # TODO
```
OmeGak: `TODO`?

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -143,26 +120,28 @@ def _process(self):
...
         file_name = secure_filename('abstract-{}.pdf'.format(self.abstract.friendly_id), 'abstract.pdf')
         return send_file(file_name, pdf.generate(), 'application/pdf')
 
 
-class RHAbstracts(RHManageAbstractsBase):
-    """Display abstracts management page"""
+class RHAbstractsDashboard(RHManageAbstractsBase):
+    """Dashboard of the abstracts module"""
 
+    # Allow access even if the feature is disabled
```
OmeGak: Why do we need this?

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -41,35 +42,15 @@
...
-        RHAbstractsBase._checkParams(self, params)
-        AbstractMixin._checkParams(self)
-        self.management = request.view_args.get('management', False)
 
 
 class RHListOtherAbstracts(RHAbstractsBase):
     """AJAX endpoint that lists all abstracts in the event (dict representation)."""
 
-    CSRF_ENABLED = True
+    def _checkProtection(self):
```
OmeGak: I think we want a slightly less permissive protection check here. Do we really want to make a list of all abstracts (which can be perceived as confidential information) to all users that can access the event?

indico/modules/events/abstracts/util.py
```diff
@@ -471,6 +471,27 @@ def get_track_reviewer_abstract_counts(event, user):
             for track, total, reviewed, unreviewed in query}
 
 
+def render_abstract_page(abstract, view_class=None, management=False):
```
OmeGak: Shouldn't this go to `views.py`?

indico/modules/events/abstracts/controllers/base.py
```diff
@@ -152,7 +152,8 @@ class AbstractsExportPDFMixin:
 
     def _process(self):
         sorted_abstracts = sorted(self.abstracts, key=attrgetter('friendly_id'))
-        pdf = ConfManagerAbstractsToPDF(self.event_new, sorted_abstracts)
+        cls = ConfManagerAbstractsToPDF if self.management else AbstractsToPDF
```
OmeGak: I guess checking per-abstract is an overkill...

indico/modules/events/abstracts/controllers/management.py
ThiefMaster: Just to make it clear why we set it to None. And it's so we can show a nice message how to enable the feature instead of hiding it from the menu or enabling it by default in all conferences (it was there before, I just added the comment)

indico/modules/events/abstracts/controllers/base.py
ThiefMaster: Yeah... if someone has a use case to download "full" PDFs e.g. as a convener we can always make this more sophisticated. I don't think we ever need per-abstract here, if you aren't in management you only get abstracts from one track here (but we'd need to enforce this when using the management pdf).

indico/modules/events/abstracts/util.py
ThiefMaster: Good question.. did we put other util functions in `views.py` so far?

indico/modules/events/abstracts/controllers/base.py
```diff
@@ -19,57 +19,24 @@
...
-        return render_template('events/abstracts/abstract.html', no_javascript=True, **params)
-
-
-class AbstractMixin:
+from MaKaC.PDFinterface.conference import ConfManagerAbstractsToPDF, AbstractsToPDF
+from MaKaC.webinterface.rh.base import RHModificationBaseProtected
+from MaKaC.webinterface.rh.conferenceDisplay import RHConferenceBaseDisplay
+
+
+class SpecificAbstractMixin:
```
nop33: How about `SingleAbstractMixin`?

ThiefMaster: I went with the naming we used in attachments (`Specific(Attachment|Folder)Mixin`)

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -16,153 +16,30 @@
...
 from indico.modules.events.abstracts.models.review_ratings import AbstractReviewRating
 from indico.modules.events.abstracts.models.reviews import AbstractReview
-from indico.modules.events.abstracts.operations import (create_abstract, delete_abstract, schedule_cfa, open_cfa,
-                                                        close_cfa, judge_abstract)
-from indico.modules.events.abstracts.schemas import abstracts_schema
+from indico.modules.events.abstracts.operations import schedule_cfa, open_cfa, close_cfa
 from indico.modules.events.abstracts.settings import abstracts_settings, abstracts_reviewing_settings
-from indico.modules.events.abstracts.util import (make_abstract_form, get_roles_for_event,
-                                                  AbstractListGeneratorManagement)
+from indico.modules.events.abstracts.util import (get_roles_for_event)
```
nop33: Why the parenthesis?

ThiefMaster: i guess pycharm didn't remove it when i told it to remove unused imports ;)

indico/modules/events/abstracts/controllers/base.py
nop33: OK. But don't you think `Single` stands better than `Specific`?
Let's ask our naming specialist @OmeGak :stuck_out_tongue_closed_eyes:

ThiefMaster: Well it deals with a *specific* abstract. Even though that's also one single abstract ;)

OmeGak: I have no preference here.

OmeGak: Fine. Leaning towards `Single`. We don't have `Specific` in many places and would be trivial to rename.

nop33: We also have tons of `Single...Field` and `Multi...Field` or whatever, so why introduce a new way to describe a single/specific element? Just because someone put it in Folders/Attachments doesn't mean it's the way to go ;)

PR #2693, opened 2016-11-08
Fix person link widget

nop33:
### Notes

I deleted the `serialize_abstract_person_link` method and moved the code into `_serialize_person_link`, since it's the approach we use in other classes as well. I don't see any point having a class method calling a util function which is only used once.

### DIFF
--- a/indico/htdocs/js/indico/widgets/person_link_widget.js
+++ b/indico/htdocs/js/indico/widgets/person_link_widget.js
@@ -60,6 +60,12 @@
         var $buttonAlphaOrder = $fieldDisplay.find('.alpha-order-switch');
         var $form = $field.closest('form');
         var customOrder = !$buttonAlphaOrder.hasClass('active');
+        var maxNewPersonID = 0;
+
+        function setNewPersonID(person) {
+            person.id = 'new-' + maxNewPersonID;
+            maxNewPersonID++;
+        }
 
         function updatePersonOrder() {
             var people = _.map($fieldDisplay.find('.person-row'), function(e) {
@@ -205,6 +211,9 @@
             }
             if (person.isSubmitter === undefined) {
                 person.isSubmitter = options.defaults.isSubmitter;
+            }
+            if (person.id === undefined) {
+                setNewPersonID(person);
             }
         }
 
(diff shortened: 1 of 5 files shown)


indico/modules/events/abstracts/fields.py
```diff
@@ -156,8 +156,11 @@ def _get_person_link(self, data):
                       'is_speaker': data.pop('isSpeaker', self.default_is_speaker)}
         return super(AbstractPersonLinkListField, self)._get_person_link(data, extra_data)
 
-    def _serialize_person_link(self, principal, extra_data=None):
-        return serialize_abstract_person_link(principal)
+    def _serialize_person_link(self, principal, extra_data={}):
```
ThiefMaster: can we please not use mutable default arguments? while it doesn't cause problems here it's just risky since someone may accidentally modify it. `extra_data or {}` in the next line will work fine with the `None` default

nop33: What do you mean by accidentally modify it?

indico/htdocs/js/indico/widgets/person_link_widget.js
```diff
@@ -60,6 +60,12 @@
         var $buttonAlphaOrder = $fieldDisplay.find('.alpha-order-switch');
         var $form = $field.closest('form');
         var customOrder = !$buttonAlphaOrder.hasClass('active');
+        var maxManualPersonID = 0;
+
+        function setManualPersonID(person) {
+            person.id = 'manual-' + maxManualPersonID;
```
ThiefMaster: There's no code on the server side that'll try to actually query a person/user with that ID?

And I'd s/manual/new/

indico/htdocs/js/indico/widgets/person_link_widget.js
```diff
@@ -206,6 +212,9 @@
             if (person.isSubmitter === undefined) {
                 person.isSubmitter = options.defaults.isSubmitter;
             }
+            if (!person.id) {
```
ThiefMaster: what if a person has id 0 (you can have a User with ID if you have a migrated DB)?

indico/htdocs/js/indico/widgets/person_link_widget.js
nop33: It seems to be working fine, after saving the form and going back to edit, the manual entered people have received a proper person id

PR #2699, opened 2016-11-09
UI updates on abstract page

nop33:


indico/htdocs/sass/modules/abstracts/_abstract_page.scss
```diff
@@ -127,6 +127,42 @@
...
+        margin-top: 0;
+
+        textarea {
+            @include transition(height .2s);
+            height: 5em;
+            width: 100%;
+        }
+    }
+
+    &.unfocused {
```
OmeGak: Can't we use `:not(:focus)`?

ThiefMaster: Then we can't do any checks like whether something was written in the textarea

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -132,6 +132,20 @@
...
             $(this).closest('.i-timeline-item').find('.i-box-footer').toggleClass('hidden');
+        }).on('focus', '.new-comment textarea', function() {
+            var $commentForm = $(this).closest('form');
+            $commentForm.find('.form-group').show('fast');
+            $commentForm.removeClass('unfocused');
+        }).on('click', '.new-comment', function(evt) {
+            evt.stopPropagation();
+        });
+
+        $(window).on('click', function() {
```
nop33: It seems like `setupAbstractPage` runs only once and it's not re-executed when reloading with AJAX (adding a review, editing a comment, ...)

PR #2701, opened 2016-11-09
Add links to reviewing in programme if applicable

DavidAndreev:


indico/modules/events/tracks/templates/display.html
```diff
@@ -24,9 +24,15 @@
     <ul class="programme">
         {% for track in event.tracks %}
             <li>
-                <span class="title">
-                    {{- track.title -}}
-                </span>
+                {% if track.can_convene(user) or track.can_review_abstracts(session.user) %}
```
ThiefMaster: user vs session.user

ThiefMaster: If you login as someone who's not an admin, manager or global convener/reviewer I'm quite sure you are now sending at least one extra query for each track in the event.

indico/modules/events/tracks/templates/display.html
```diff
@@ -24,9 +24,15 @@
     <ul class="programme">
         {% for track in event.tracks %}
```
ThiefMaster: If you login as someone who's not an admin, manager or global convener/reviewer I'm quite sure you are now sending at least one extra query for each track in the event.

To avoid this, you can do this in the controller and then use `tracks1 instead of `event.tracks` in the loop above.
```
        tracks = (Track.query
                  .with_parent(self.event_new)
                  .options(subqueryload('conveners'), subqueryload('reviewers')).
                  all())
```

DavidAndreev: Im getting this error : 
ArgumentError: Can't find property named 'reviewers' on the mapped entity Mapper|Track|tracks in this Query.

indico/modules/events/tracks/controllers.py
```diff
@@ -111,8 +112,12 @@ class RHDisplayTracks(RHConferenceBaseDisplay):
     def _process(self):
         page_title = get_menu_entry_by_name('program', self._conf).localized_title
         program = track_settings.get(self.event_new, 'program')
+        tracks = (Track.query.with_parent(self.event_new)
+                  .options(subqueryload('conveners'),
+                           subqueryload('abstract_reviewers')).
+                  all())
```
nop33: put the dot in this line
```
...
.all())
```

PR #2705, opened 2016-11-11
Fix overlap in menu customization page when not enough width

mvidalgarcia:
Before:
screenshot from 2016-11-11 11-35-17

After:
screenshot from 2016-11-11 12-11-08

### DIFF
--- a/indico/htdocs/sass/modules/_event_management.scss
+++ b/indico/htdocs/sass/modules/_event_management.scss
@@ -161,8 +161,7 @@
 }
 
 .menu-entries {
-    width: 370px;
-    padding: 5px 10px 5px 5px;
+    padding: 5px 10px;
     margin-top: 0;
 
     list-style: none;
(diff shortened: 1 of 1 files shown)


indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -161,8 +161,8 @@
 }
 
 .menu-entries {
-    width: 370px;
-    padding: 5px 10px 5px 5px;
+    max-width: 370px;
```
nop33: I don't see any problem when removing this rule completely. Same with the one in line `325`. The more we can avoid hard-coded pixel values, the better.

OmeGak:
How responsive is this solution? I would be expecting flex boxes to get this layout behave correctly, but I don't see them in your changes.

nop33:
@OmeGak I actually checked his PR out and tried to use Flexbox instead. I was having some weird problems and didn't spend too much time on it.

Issue #2664, opened 2016-10-24
Protection: make it clear why ACL is disabled


pferreir:
![image](https://cloud.githubusercontent.com/assets/2699/19648331/0b1f05a6-9a02-11e6-8b8f-69130baaea8f.png)

Some users find it hard to figure out how to enable the ACL when the event is set as public. We could add some kind of indication of the reason why the list is disabled. This could be achieved e.g. with a tooltip.

OmeGak:
Another option is hiding the list unless it's relevant. For discoverability of the feature, instead of hiding the field altogether, we could show a link that would switch the protection mode to protected. It would look like this.

![screenshot from 2016-10-24 15 47 56](https://cloud.githubusercontent.com/assets/716307/19692508/6feecf9e-9ad8-11e6-90ed-ffcb5917a436.png)

PR #2708, opened 2016-11-14
Hide ACL when is disabled in Protection page

mvidalgarcia:
Followed @OmeGak's approach on #2664

### DIFF
--- a/indico/htdocs/js/indico/widgets/protection_widget.js
+++ b/indico/htdocs/js/indico/widgets/protection_widget.js
@@ -29,6 +29,8 @@
         }, options);
 
         var inputs = $('input[name=' + options.fieldName + '][id^=' + options.fieldId + ']');
+        var $enableAclLink = $('.enable-acl-link');
+        var $aclListField = $('.acl-list-field');
 
         inputs.on('change', function() {
             var $this = $(this);
@@ -38,7 +40,6 @@
                 $('#form-group-protected-{0} .protection-message'.format(options.fieldId)).hide();
                 $('#form-group-protected-{0} .{1}-protection-message'.format(options.fieldId, $this.val())).show();
 
-                toggleAclField($('#' + options.aclFieldId), !isProtected);
                 if (options.aclMessageUrl && options.hasInheritedAcl) {
                     $.ajax({
                         url: options.aclMessageUrl,
@@ -49,7 +50,14 @@
                         }
                     });
                 }
+                $aclListField.toggleClass('hidden', !isProtected);
+                $enableAclLink.toggleClass('hidden', isProtected);
             }
+        });
+
+        $enableAclLink.on('click', function(evt) {
+            evt.preventDefault();
+            $('input[name="protection_mode"][value="protected"]').trigger('click');
         });
 
         _.defer(function() {
--- a/indico/web/templates/forms/principal_list_widget.html
+++ b/indico/web/templates/forms/principal_list_widget.html
@@ -2,8 +2,14 @@
 
 
 {% block html %}
+    {% if acl %}
+        <a href="#" class="enable-acl-link hidden"
+           title='{% trans %}This will change the event protection mode to "Protected"{% endtrans %}'>
+            {%- trans %}Restrict access to this event{% endtrans -%}
+        </a>
+    {% endif %}
     {% set value_json = field._value() | tojson %}
-    <div class="i-form-field-fixed-width" data-tooltip-anchor>
+    <div class="i-form-field-fixed-width {{ 'acl-list-field' if acl }}" data-tooltip-anchor>
         {% if acl %}
             <div class="inheriting-acl-message"></div>
         {% endif %}


indico/web/templates/forms/principal_list_widget.html
```diff
@@ -2,8 +2,13 @@
 
 
 {% block html %}
+    {% if acl %}
+        <a href="#" class="enable-acl-link hidden" data-qtip-oldtitle="Protected mode will be enabled">
```
nop33: I think `data-qtip-oldtitle` is not needed and the `title="{% trans %}Protected mode will be enabled{% endtrans %}"` should be enough, but verify.

indico/htdocs/js/indico/widgets/protection_widget.js
```diff
@@ -29,16 +34,16 @@
         }, options);
 
         var inputs = $('input[name=' + options.fieldName + '][id^=' + options.fieldId + ']');
+        var $enableAclLink = $('.enable-acl-link');
```
mvidalgarcia: I know, it was my first intention but `ACLLink` looked weird to me because of the `LL` so a went for `Acl`. Anyway, not a big deal, I can change it.

ThiefMaster: I don't like non-uppercase acronyms but it looks pretty bad with all-uppercase too so I usually go for Html/Acl/etc even though it looks like :hankey:

indico/htdocs/js/indico/widgets/protection_widget.js
```diff
@@ -49,9 +54,19 @@
...
+                $aclListField.toggleClass('hidden', !isProtected);
+                $enableAclLink.toggleClass('hidden', isProtected);
             }
         });
 
+        $enableAclLink.on('click', function(evt) {
+            evt.preventDefault();
+            $aclListField.removeClass('hidden');
+            $enableAclLink.addClass('hidden');
+            $('input[value="protected"]').prop("checked", true);
```
ThiefMaster: `'checked'`

indico/web/templates/forms/principal_list_widget.html
mvidalgarcia: You were right :+1:

pferreir:
Let's keep this on hold until we decide this is really the solution we want?

indico/htdocs/js/indico/widgets/protection_widget.js
OmeGak: `RHAbstract` though.

ThiefMaster: I think in Python it's much more common to keep acronyms all uppercase. At least we do it pretty much everywhere in the backend

indico/web/templates/forms/principal_list_widget.html
```diff
@@ -2,8 +2,13 @@
 
 
 {% block html %}
+    {% if acl %}
+        <a href="#" class="enable-acl-link hidden" title="{% trans %}Protected mode will be enabled{% endtrans %}">
+            {%- trans %}Enable access control list{% endtrans -%}
```
mvidalgarcia: Should we find a better wording for this (and the tooltip)?

indico/htdocs/js/indico/widgets/protection_widget.js
nop33: Wow, didn't expect my comment to raise such a discussion. Let's just find a convention and stick with it! Facebook poll: :+1: for `ACL`, :-1: for `Acl` ! Haha

indico/web/templates/forms/principal_list_widget.html
nop33: I would suggest `Restrict access to this event` with a tooltip `This will change the event protection mode to "Protected"`

indico/htdocs/js/indico/widgets/protection_widget.js
nop33: Mothertrucker, don't mess with my poll.

indico/htdocs/js/indico/widgets/protection_widget.js
```diff
@@ -49,9 +50,16 @@
...
                     });
                 }
+                $aclListField.toggleClass('hidden', !isProtected);
+                $enableAclLink.toggleClass('hidden', isProtected);
             }
         });
 
+        $enableAclLink.on('click', function(evt) {
+            evt.preventDefault();
+            $('input[value="protected"]').trigger('click');
```
ThiefMaster: Quite generic. As there's no decent ID selector I'd go for `$('input[name=protection_mode][value=protected]')`

indico/web/templates/forms/principal_list_widget.html
mvidalgarcia: ping @ThiefMaster @pferreir @OmeGak @DavidAndreev ^

nop33: Seems like people agree with my comment :smile: Their silence is my answer!

mvidalgarcia: OK, I'll change it then

PR #2707, opened 2016-11-11
Add inlineAjaxForm util / data-ajax-form & use it for abstract reviewing forms

ThiefMaster:

PR #2698, opened 2016-11-09
CFA display page

mvidalgarcia:
Removes Submit Abstract side menu
Removes legacy RH and WP

PR #2712, opened 2016-11-14
AJAX form util + CFA page

ThiefMaster:
For the AJAX form stuff check #2707, the CFA page is in #2698 but modified in the latest commit in this PR.


indico/modules/events/abstracts/controllers/display.py
```diff
@@ -27,55 +26,47 @@
...
-        pdf = AbstractsToPDF(self.event_new, sorted_abstracts)
+        pdf = AbstractsToPDF(self.event_new, get_user_abstracts(self.event_new, session.user))
         return send_file('my-abstracts.pdf', pdf.generate(), 'application/pdf')
 
 
 class RHSubmitAbstract(RHAbstractsBase):
-    def _process(self):
-        if not self.event_new.cfa.can_submit_abstracts(session.user):
-            return redirect(url_for('.call_for_abstracts', self.event_new))
+    """Submit a new abstract."""
```
nop33: A bit verbose, no? :sweat_smile:

indico/modules/events/abstracts/templates/display/submit_abstract.html
```diff
@@ -1,12 +1,8 @@
...
 {% from 'forms/_form.html' import simple_form %}
 
-{% macro render_abstract_submit_form(form, event) %}
-    <div class="header">
-        <div class="header-row">
-            <h3>{%- trans %}Submit abstract{% endtrans -%}</h3>
-        </div>
+<div class="header">
+    <div class="header-row">
+        <h3>{%- trans %}Submit abstract{% endtrans -%}</h3>
```
nop33: No need for `-`

indico/modules/events/abstracts/controllers/display.py
ThiefMaster: Not having a docstring in just one RH in that file looked weird :p

indico/modules/events/abstracts/templates/display/submit_abstract.html
ThiefMaster: True, but at the same time it does no harm (just like having whitespace around is usually not an issue if it's not inside trans)

nop33: Well, if you want to put extra characters in your code that have absolutely 0 effect, sure thing :lollipop:

indico/htdocs/js/utils/misc.js
```diff
@@ -169,4 +171,33 @@
...
+        if (replace) {
+            var $html = $($.parseHTML(html, document, true));
+            $target.replaceWith($html);
+            $target = $html;
+        } else {
+            $target.html($.parseHTML(html, document, true));
+        }
+        $target.trigger('indico:htmlUpdated');
+        if (highlight) {
+            $target.effect('highlight', {color: '#5D95EA'});
```
OmeGak: Maybe we should define the palette in `vars.js`? It will be certainly easier to modify in the future.

ThiefMaster: I just kept what we had before when moving it but sounds like a good idea to me.

ThiefMaster: Since we can't easily extract it from the scss file I think I'd rather add it as a new palette.js file though instead of going through vars.js

PR #2716, opened 2016-11-16
Show affiliation of authors in abstract page

mvidalgarcia:
I don't know how to make the `background-color` to match the width of the user.
image

### DIFF
--- a/indico/htdocs/sass/modules/_users.scss
+++ b/indico/htdocs/sass/modules/_users.scss
@@ -90,6 +90,11 @@
     &:not(first-child) {
         margin-left: 0.1em;
     }
+
+    .affiliation {
+        color: $dark-gray;
+        font-size: 0.9em;
+    }
 }
 
 .user-search-results {
--- a/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
+++ b/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
@@ -256,15 +256,20 @@
             margin-top: 0;
         }
 
+        label {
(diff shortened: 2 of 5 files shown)


ThiefMaster:
Do we even need a background color there?

nop33:
I think it looks good with the bg. Doesn't `display: inline-block` fix the problem?

indico/htdocs/sass/modules/_users.scss
```diff
@@ -78,10 +78,14 @@
     display: inline-block;
     font-size: 0.95em;
     padding: 0.2em 0.5em;
-    margin-bottom: 0.1em;
+    margin-bottom: 0.5em;
```
OmeGak: We don't want to increase `margin-bottom` in all cases. We are using `.event-user` in other places, like for instance in meeting pages ([example](https://indico.cern.ch/event/588588/)).

indico/htdocs/sass/modules/_users.scss
```diff
@@ -78,10 +78,14 @@
...
     color: $black;
     background: $pastel-gray;
 
+    &.one-line {
```
OmeGak: I don't think this belongs in here. If you want to have one per line simply target `.author-list li` within the specific box in `_abstract_page.scss`.

mvidalgarcia:
@nop33 I doesn't work with `display: inline-block` since we want each user in a different line as @pferreir suggested. With `inline-block` would look like this which IMO is not that bad.
image

PR #2717, opened 2016-11-16
Several improvements (from Asana list)

pferreir:
* Author sorting
* Markdown in comments
* Fixed judgment comment placeholder
* Added CfA announcement on conference home page

### DIFF
--- a/indico/modules/events/abstracts/templates/display/_common.html
+++ b/indico/modules/events/abstracts/templates/display/_common.html
@@ -14,7 +14,7 @@
                     {% endif %}
                 </div>
                 <div>
-                    {%- trans %}Your submitted abstract will be subject to the reviewing process.{% endtrans -%}
+                    {%- trans %}You can submit an abstract for reviewing.{% endtrans -%}
                 </div>
                 <div class="toolbar right">
                     {% if session.user and not redirect_first %}


indico/modules/events/abstracts/models/comments.py
```diff
@@ -60,7 +67,7 @@ class AbstractComment(db.Model):
         index=True,
         nullable=False
     )
-    text = db.Column(
+    _text = db.Column(
```
ThiefMaster: `'text'` as the first argument so it doesn't expect the column to be named `_text`

indico/modules/events/abstracts/models/reviews.py
```diff
@@ -77,7 +84,7 @@ class AbstractReview(db.Model):
         UTCDateTime,
         nullable=True
     )
-    comment = db.Column(
+    _comment = db.Column(
```
ThiefMaster: `'comment'` for the column name

indico/modules/events/abstracts/__init__.py
```diff
@@ -141,3 +142,10 @@ def _reviewing_area_visible(conf):
     yield MenuEntryData(title=_("Reviewing Area"), name='user_tracks', endpoint='abstracts.display_reviewable_tracks',
                         position=0, parent='call_for_abstracts',
                         visible=_reviewing_area_visible)
+
+
+@template_hook('conference-home-info')
+def _inject_regform_announcement(event, **kwargs):
```
ThiefMaster: regform? :monkey:

indico/modules/events/abstracts/__init__.py
```diff
@@ -141,3 +142,10 @@ def _reviewing_area_visible(conf):
     yield MenuEntryData(title=_("Reviewing Area"), name='user_tracks', endpoint='abstracts.display_reviewable_tracks',
                         position=0, parent='call_for_abstracts',
                         visible=_reviewing_area_visible)
+
+
+@template_hook('conference-home-info')
+def _inject_cfa_announcement(event, **kwargs):
+    event = event.as_event
+    if event.cfa.is_open or (session.user and event.cfa.can_submit_abstracts(session.user)):
```
ThiefMaster: check whether the feature is enabled first. `can_submit_abstracts` always triggers a DB query to get the list of authorized users if the cfa is not open

indico/web/templates/forms/_person_link_widget_base.html
```diff
@@ -15,12 +15,13 @@
 {% endmacro %}
 
 {% block html %}
-    {% set custom_display_order = field._value()|selectattr('displayOrder')|list|length > 0 %}
+    {% set list_in_custom_order = (field._value()|selectattr('displayOrder')|list|length > 0) %}
```
OmeGak: Out of curiosity. Why is `(...)` needed here?

ThiefMaster: I don't think it is. Probably just to make it prettier (I tend to do the same when having kwargs with operators)

pferreir: It's not needed. Probably left over from a previous version where I was joining that clause with something else.

PR #2720, opened 2016-11-17
Make conference program editable

ThiefMaster:
...and convert it from html to markdown when editing an old one, with the usual warning etc

Run this in the shell so all old ones are rendered as HTML by default:

```python
from indico.core.db.sqlalchemy.descriptions import RenderMode
from indico.modules.events.tracks.settings import track_settings
for setting in EventSetting.query.filter_by(module='tracks', name='program'):
    track_settings.set(setting.event_id, 'program_render_mode', RenderMode.html)
transaction.commit()
```

### DIFF
--- a/indico/htdocs/sass/base/_pages.scss
+++ b/indico/htdocs/sass/base/_pages.scss
@@ -79,6 +79,7 @@
 
         .page-description {
             @extend %font-family-title-light;
+            @extend .fixed-width;
             color: $dark-gray;
             font-size: 1.2em;
             margin-top: $separator-margin;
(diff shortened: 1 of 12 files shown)


nop33:
I'd rather have something like this:
image


instead of the undiscoverable new pattern that we started using and I keep trying to eliminate before it lays eggs:
image

I will lose faith in "Designing user friendly interfaces" if most people actually believe that the approach of the cog icon next to a page title is better UX than the quick example I proposed above.

ThiefMaster:
I like the idea. Any good ideas for "Some text" and "Some description"? Maybe we could even indicate if there's currently a programme or not?

nop33:
What is the markdown area in the dialog that opens when clicking "Edit program" used for?

OmeGak:
I don't like either of them, but the cogwheel is one of the most confusing new elements we have added as of lately and I would avoid if possible.

ThiefMaster:
> What is the markdown area in the dialog that opens when clicking "Edit program" used for?

Preview of the program which is now written in markdown instead of HTML

nop33:
There you go then :smile:

PR #2722, opened 2016-11-18
Lots of timeline stuff

pferreir:
Just wanting to avoid merge issues.

 * Now displaying track names/codes in a logical way;
 * Reworked the way track information is displayed; 
 * Added fields for instructions:
   - Submission
   - Reviewing
   - Judgment


indico/htdocs/js/indico/jquery/declarative.js
```diff
@@ -139,6 +142,7 @@
                         url: url,
                         method: method,
                         data: params,
+                        content: $this.parent().find(content).html(),
```
OmeGak: Is going to the parent strictly necessary? If we **always** expect `content` to be descendant of `$this` we can simply use `$this.find()`. If we **never** expect `content` to be descendant of `this` we should simply use `.siblings()`. If we can't make any of these assumptions then I said nothing. :P

indico/modules/events/abstracts/templates/abstract/review.html
```diff
@@ -1,9 +1,19 @@
...
 
 {% macro render_review_form(form, abstract=none, track=none, review=none, management=false) %}
     {% set url = url_for('.review_abstract', abstract, track, management=management) if not review else
                  url_for('.edit_review', review, management=management) %}
+
+    {% set event = abstract.event_new %}
+
+    {% call(link, endlink) render_instruction_box(event.cfa.reviewing_instructions, _("Reviewing instructions")) %}
+        {% trans -%}
+            Please don't forget to read the {{ link }}reviewing instructions{{ endlink }} before submitting a review.
```
nop33: In the case of judgment you said "...before taking **your** decision", but in this case "...before submitting **a** review.
Shouldn't these 2 be the same? :thinking: 
Oh, also below "...before submitting **an** abtract"

indico/modules/events/abstracts/templates/abstract/_common.html
```diff
@@ -33,3 +35,22 @@
...
+{% macro render_instruction_box(instructions, title) %}
+    {% if instructions %}
+        {% set link %}
+            <a data-title="{{ title }}"
+               data-content=".hidden-instructions"
+               data-close-button
+               data-ajax-dialog>
+        {% endset %}
+        {% set endlink='</a>' %}
+        {% set text = caller(link|safe, endlink|safe) %}
```
OmeGak: Why do we need this sorcery here? Can't we simply render `<a ...></a>` within `{% call message_box(...) %}`?

indico/modules/events/abstracts/templates/abstract/review.html
pferreir: > Shouldn't these 2 be the same? 🤔 

:rage4:

indico/htdocs/js/indico/jquery/declarative.js
ThiefMaster: OK I see how you are using this. While it's nice for that specific usecase (not having to pass an ID to the macro) I don't think it's a good idea as it reduces reusability of this feature..

pferreir: What is the likelihood the content will be far away, in a completely different part of the DOM tree? But yeah, maybe this is too much...

indico/modules/events/abstracts/templates/abstract/_common.html
pferreir: Not with `{% trans %}`.

indico/htdocs/js/indico/jquery/declarative.js
ThiefMaster: Well all the other selectors in declarative etc aren't "restricted"

ThiefMaster:
The margins look a bit broken:

ss

ThiefMaster:
Pushed a rebased version btw

pferreir:
> Pushed a rebased version btw with the conflict in review.html solved

Oh great. Hadn't noticed that. Anyway, I solved the margin thingy too.

indico/modules/events/abstracts/templates/abstract/timeline.html
```diff
@@ -209,24 +227,23 @@
...
 
 
 {% macro _render_input(abstract, comment_form, review_form=none) %}
     <div id="abstract-timeline-input" class="i-timeline-item">
         <div class="i-timeline-item-label action"></div>
-        <div class="i-timeline-item-box footer-only header-indicator">
+        <div class="i-timeline-item-box footer-only header-indicator-left">
             <div class="i-box-footer js-form-container">
-                <div class="flex">
                     {% if abstract.can_comment(session.user) %}
```
ThiefMaster: too much indentation

PR #2725, opened 2016-11-21
Improve abstract-related logging

ThiefMaster:


indico/modules/events/abstracts/operations.py
```diff
@@ -271,22 +325,57 @@ def create_abstract_review(abstract, track, user, review_data, questions_data):
...
+        'Action': orig_string(review.proposed_action.title),
+        'Comment': review.comment
+    }
+    if review.proposed_action == AbstractAction.accept:
+        log_data['Contribution type'] = (review.proposed_contribution_type.name if review.proposed_contribution_type
+                                         else None)
+    elif review.proposed_action == AbstractAction.change_tracks:
+        log_data['Other tracks'] = sorted(t.title for t in review.proposed_tracks)
+    elif review.proposed_action in {AbstractAction.mark_as_duplicate, AbstractAction.merge}:
+        log_data['Other abstract'] = '#{0.friendly_id} ({0.title})'.format(review.proposed_related_abstract)
```
ThiefMaster: You can even do `{0['foo']}` for dicts. ;)
I usually prefer passing separate arguments but here I did it mostly to avoid having to wrap it.

FWIW, I'm considering to add a `verbose_title` property to `Abstract` returning this...

pferreir:
image

I think green and red would be easier to understand. And maybe colour the carriage return differently? (if it's easy)

ThiefMaster:
It is red/green. Looks like your have old CSS cached.

pferreir:
> It is red/green. Looks like your have old CSS cached.

Nope, event tried with another browser ™️ 
All I've done was rebasing on devel.

ThiefMaster:
Strange... https://github.com/indico/indico/pull/2725/files#diff-2389a37a89bb93ca8a3d9f9805666abfR148

I'll check when I'm in the office.

PR #2726, opened 2016-11-22
Review summary accordion

nop33:
Adding a more detailed view of the review summary:
* Listing track reviews
* Listing the average score per question per track


indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -37,43 +37,97 @@
...
+    <div class="state-icon {{ mapping_class[state.name] }} flexrow f-a-center f-j-center"
+         title="{{ mapping_title[state.name] }}">
+        <span
+            {%- if state.name in ['positive', 'conflicting'] %}
+                class="icon-checkmark"
+            {% elif state.name in ['mixed', 'negative'] %}
+                class="icon-cross"
+            {%- endif -%}
+        ></span>
+    </div>
```
nop33: In the meeting we said that we want the *Mixed* state to have a tilt icon. We'll need to add a new one in our `_icons.scss` file cause using the character `~` is not very consistent and needs special styling.
As for the *Conflict* state, we said to use a yellow color, which will be the same as *Mixed*. Is that rly what we want?

nop33: Such an icon would be more suitable for the *Mixed* state imo:
image

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -37,43 +37,97 @@
...
+                            <div class="details-content-inner">
+                                {% for review in abstract.get_reviews(track=track) %}
+                                    <div class="details-content-row flexrow f-a-center">
+                                        <div class="stretch">
+                                            {%- trans name=review.user.name, action=review.proposed_action.title -%}
+                                                <strong>{{ name }}</strong> proposes to {{ action }}
+                                            {%- endtrans %}
+                                        </div>
+                                        <a href="#abstract-review-{{ review.id }}"
+                                           class="track-review-details-link js-heighlight-review">
```
ThiefMaster: highlight

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -74,6 +74,26 @@
...
+            }
+
+            if ($this.hasClass('js-details-hidden')) {
+                $('.js-details-visible').each(function() {
+                    toggleDetails($(this));
+                });
+            }
+            toggleDetails($this);
+        }).on('click', '.js-heighlight-review', function() {
+            $($(this).attr('href') + ' .i-timeline-item-box').effect('highlight', 800);
```
ThiefMaster: `.effect('highlight', {color: Palette.highlight}, 800)`?

indico/htdocs/sass/base/_layout.scss
```diff
@@ -156,3 +156,58 @@ body > * {
 .flex-center {
     justify-content: center;
 }
+
+
+/* Flexbox layout */
+.flexrow {
```
ThiefMaster: I still think this would be clearer:

> [18:37:52] <@ThiefMaster> flexrow -> .flex.row?
> [18:38:05] <@ThiefMaster> ie .flex to set display:flex and .col/.row to set the flow

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -566,6 +567,23 @@ def get_track_reviewing_state(self, track):
         else:
             return AbstractReviewingState.mixed
 
+    def get_avg_per_question_per_track(self):
+        ratings = {}
+        abstract_ratings = (AbstractReviewRating.query
```
ThiefMaster: `.join(AbstractReviewRating.question)` and then add `~AbstractReviewQuestion.is_deleted` to the filters

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -566,6 +567,23 @@ def get_track_reviewing_state(self, track):
...
+                            .filter(AbstractReview.abstract_id == self.id,
+                                    AbstractReviewRating.review_id == AbstractReview.id)
```
ThiefMaster: joining like this is quite ugly

ThiefMaster: Actually, use this query. That way you also avoid spamming extra queries for each question/review:

```python
abstract_ratings = (AbstractReviewRating.query
                    .join(AbstractReviewRating.review)
                    .join(AbstractReviewRating.question)
                    .options(contains_eager('review'), contains_eager('question'))
                    .filter(AbstractReview.abstract == self, 
                            ~AbstractReviewQuestion.is_deleted)
                    .all())
```

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -19,13 +19,13 @@
                             'negative': _("No reviewer proposed to accept this abstract"),
                             'mixed': _("Not all reviewers proposed to accept this abstract")} %}
     {% set state = abstract.reviewing_state %}
-    <span class="i-tag {{ mapping_class[state.name] }}" title="{{ mapping_title[state.name] }}">
+    <span class="i-tag {{ mapping_class[state.name] }} reviewing-state"
```
ThiefMaster: I'd keep the dynamic class at the end and not in the middle of two static classes

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -37,43 +37,97 @@
...
     {% set state = abstract.get_track_reviewing_state(track) %}
-    <span class="i-tag outline {{ mapping_class[state.name] }}" title="{{ mapping_title[state.name] }}">
-        {{ state.title -}}
-    </span>
+    <div class="state-icon {{ mapping_class[state.name] }} flexrow f-a-center f-j-center"
+         title="{{ mapping_title[state.name] }}">
+        <span
+            {%- if state.name in ['positive', 'conflicting'] %}
+                class="icon-checkmark"
+            {% elif state.name in ['mixed', 'negative'] %}
```
ThiefMaster: same here

indico/modules/events/abstracts/templates/abstract/judge.html
ThiefMaster: that icon looks more like "swap" to me

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -37,43 +37,97 @@
...
+                <div class="track-review">
+                    <div class="track-review-summary flexrow f-a-center">
+                        {{ _render_abstract_track_reviewing_state_icon(abstract, track) }}
+                        <div class="stretch">
+                            {{- track.title -}}
+                        </div>
                         {% set score = abstract.get_track_score(track) %}
+                        {% if abstract.reviews %}
+                            <a href="#" class="track-review-details-link js-details-hidden js-toggle-details"
+                               data-target="review-list"
```
ThiefMaster: `.review-list` and remove the `.` from the JS file - currently you have no idea if that's a class or ID when seeing this.

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -74,6 +74,26 @@
...
+
+            function toggleDetails($trigger) {
+                var target = $trigger.data('target');
+                var wasVisible = $trigger.hasClass('js-details-visible');
+                $trigger.closest('.track-review').find('.track-review-details .' + target).slideToggle('fast');
+                $trigger.toggleClass('js-details-hidden js-details-visible');
+                $trigger.text(wasVisible ? $trigger.data('show-text') : $trigger.data('hide-text'));
+            }
+
+            if ($this.hasClass('js-details-hidden')) {
```
ThiefMaster: Actually, why use a class at all and not just `.data('hidden', wasVisible)`?

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -74,6 +74,26 @@
             $reviewBox.find('.js-edit-review').trigger('click');
             $('body, html').animate({scrollTop: $reviewBox.offset().top}, 'fast');
             $reviewBox.find('.i-timeline-item-box').effect('highlight', {color: Palette.highlight}, 'slow');
+        }).on('click', '.js-toggle-details', function(evt) {
+            evt.preventDefault();
+            var $this = $(this);
+
+            function toggleDetails($trigger) {
```
ThiefMaster: This functionality is so generic that this could easily go into declarative.js (bind to `[data-toggle]:not([data-toggle=dropdown])`)!

indico/modules/events/abstracts/templates/abstract/judge.html
OmeGak: Mixed as in not clear, not as in swapped.

indico/htdocs/js/indico/modules/abstracts.js
OmeGak: IngularJS

indico/htdocs/js/indico/modules/abstracts.js
OmeGak: Isn't `'fast'` not fast enough? I think it would be better to stick to standard speeds than arbitrary speed values.

indico/modules/events/abstracts/templates/abstract/judge.html
OmeGak: ermahgerd

indico/htdocs/sass/base/_layout.scss
OmeGak: I don't think we should provide every class to be orthogonal with the full expressiveness of flex-boxes. flewrow` is just a class for an opinionated layout (flex row non-wrapping).

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
```diff
@@ -223,6 +223,10 @@
     margin-left: 1em;
 }
 
+.reviewing-state {
```
OmeGak: We have several ways to display the reviewing state. I would go for nesting where it belongs or simply a more specific name (better).

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
```diff
@@ -243,8 +247,55 @@
...
+
+    .track-review-summary {
+        padding: 5px 10px;
+    }
+
+    .track-review-details {
+        background-color: $light-gray;
+
+        .details-content-row {
+            padding: 5px 10px 5px calc(20px + 1em);
```
OmeGak: `calc()` 
:thought_balloon: :boom: 
(mind blown)

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
```diff
@@ -243,8 +247,55 @@
...
+    .review-state-pretext {
+        margin-right: 5px;
+    }
+
+    .i-box-footer {
+        margin-top: 0px;
+    }
+
+    .details-content-inner {
+        box-shadow: inset 0px 15px 20px -20px $gray;
```
OmeGak: How does it look in the end?

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -566,6 +567,23 @@ def get_track_reviewing_state(self, track):
         else:
             return AbstractReviewingState.mixed
 
+    def get_avg_per_question_per_track(self):
```
OmeGak: We already refer to `avg` as `score`.

OmeGak: Do we really need to return the scores for all tracks? If it doesn't incur in massive performance drawback I would rather have `get_questions_score(self, track)`. I would compose the per-track dictionary outside.

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -37,43 +37,97 @@
...
+                                        {{ render_score(score, abstract.event_new.cfa) }}
+                                    </div>
+                                {% endfor %}
+                            </div>
+                        </div>
+                        <div class="review-list details-content weak-hidden">
+                            <div class="details-content-inner">
+                                {% for review in abstract.get_reviews(track=track) %}
+                                    <div class="details-content-row flexrow f-a-center">
+                                        <div class="stretch">
```
OmeGak: Maybe it's a good moment to rename `stretch` to `f-grow`.

indico/modules/events/abstracts/models/abstracts.py
ThiefMaster: One query vs n queries. Doing is for all tracks is most likely faster

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
nop33: Good.

nop33:
Merge this PR after #2731 is merged.

indico/modules/events/abstracts/templates/abstract/judge.html
nop33: We have the following track states: `positive`, `negative`, `conflicting`, `mixed`.

Based on the meeting we had, we decided on the following:
* `positive`: :heavy_check_mark:  (in green)
* `negative`: :x: 
* `conflicting`: :heavy_check_mark: (in purple)
* `mixed`: ~ (in yellow)

Is that the final decision?
In this case we need to find an icon for `~` (which icomoon doesn't unfortunately have), or apply some custom CSS to the character `~` to make it look like the other icons.

indico/htdocs/js/indico/modules/abstracts.js
nop33: @OmeGak I copied the line from the registration page.

indico/htdocs/js/indico/modules/abstracts.js
nop33: This particular case is not so generic since it's the implementation of an accordion, meaning that you cannot have more than 1 item open. Maybe this could be another data attribute...

ThiefMaster: Ah right, so declarative would be overkill.

nop33:
We have the following track states: `positive`, `negative`, `conflicting`, `mixed`.

Based on the meeting we had, we decided on the following:
* `positive`: :heavy_check_mark:  (in green)
* `negative`: :x: 
* `conflicting`: :heavy_check_mark: (in purple)
* `mixed`: ~ (in yellow)

Is that the final decision?
In this case we need to find an icon for `~` (which icomoon doesn't unfortunately have), or apply some custom CSS to the character `~` to make it look like the other icons.

### Current layout
image

nop33:
Comments and answers to the above comment are appreciated :smile:

indico/htdocs/js/indico/jquery/declarative.js
```diff
@@ -303,4 +304,29 @@
             }).html('&para;').appendTo($elem);
         });
     }
+
+    function setupToggleLinks() {
+        $('[data-toggle]:not([data-toggle=dropdown])').on('click', function(evt) {
```
ThiefMaster: Wait, this is not truly declarative! It runs once and binds to the elements that match the selector at that point, not ones added in the future! It actually stops working after editing e.g. a comment. You want this instead:

    $('body').on('click', '[data-toggle]:not([data-toggle=dropdown])', function(evt) {

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -77,6 +77,8 @@
             var $this = $(this);
             $this.toggleClass('open');
             $this.closest('.ratings-switch').siblings('.ratings-details').toggle();
+        }).on('click', '.js-highlight-review', function() {
+            $($(this).attr('href') + ' .i-timeline-item-box').effect('highlight');
```
ThiefMaster: use the `Palette.highlight` color

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -571,6 +574,26 @@ def get_track_reviewing_state(self, track):
...
+                            .join(AbstractReviewRating.review)
+                            .join(AbstractReviewRating.question)
+                            .options(contains_eager('review'), contains_eager('question'))
+                            .filter(AbstractReview.abstract == self,
+                                    ~AbstractReviewQuestion.is_deleted)
+                            .all())
+        for rating in abstract_ratings:
+            if rating.question.no_score:
+                continue
+            ratings.setdefault(rating.review.track_id, {})
```
ThiefMaster: `ratings = defaultdict(dict)` and you can get rid of this line

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -571,6 +574,26 @@ def get_track_reviewing_state(self, track):
...
+            question_avg_rating = ratings[rating.review.track_id].get(rating.question)
+            if question_avg_rating:
+                ratings[rating.review.track_id][rating.question] = (question_avg_rating + rating.value) / 2
```
ThiefMaster: This logic is not correct:

See this example:

```
In [6]: ratings = [10, 5, 15]

In [7]: sum(ratings) / len(ratings)
Out[7]: 10.0

In [8]: avg = 10
   ...: avg = (avg + 5) / 2
   ...: avg = (avg + 15) / 2
   ...:

In [9]: avg
Out[9]: 11.25
```

ThiefMaster: I'd rather build at list with the scores and then divide the sum of that list by the length of it.

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -37,44 +37,106 @@
                             'negative': _("No reviewer proposed to accept this abstract"),
                             'mixed': _("Not all reviewers proposed to accept this abstract")} %}
     {% set state = abstract.get_track_reviewing_state(track) %}
-    <span class="i-tag outline {{ mapping_class[state.name] }}" title="{{ mapping_title[state.name] }}">
-        {{ state.title -}}
-    </span>
+    <div class="state-icon {{ mapping_class[state.name] }} flexrow f-a-center f-j-center"
+         title="{{ mapping_title[state.name] }}">
+        <span
+            {%- if state.name in ('positive', 'conflicting') %}
```
ThiefMaster: Why not use another mapping and then just `class="{{ foo.get(state.name, 'icon-circle-small') }}"`?

indico/htdocs/js/indico/modules/abstracts.js
ThiefMaster: `.effect('highlight', {color: Palette.highlight}, 'slow')` for the sake of consistency

indico/modules/events/abstracts/models/abstracts.py
nop33: But I want to add the track id as key and an empty dict as a value.

ThiefMaster: This would replace the `ratings = {}`. As soon as you access a ratings[foo] it'll add that key with a new dict as the value. Which is exactly what you do here..

ThiefMaster:
```python
def get_track_question_scores(self):
    query = (db.session.query(AbstractReview.track_id,
                              AbstractReviewRating.question_id,
                              db.func.avg(AbstractReviewRating.value))
             .join(AbstractReviewRating.review)
             .join(AbstractReviewRating.question)
             .filter(AbstractReview.abstract == self,
                     ~AbstractReviewQuestion.is_deleted)
             .group_by(AbstractReview.track_id, AbstractReviewRating.question_id))
    scores = defaultdict(lambda: defaultdict(dict))
    for track_id, question_id, score in query:
        scores[track_id][question_id] = score
    return scores
```

ThiefMaster: If you prefer 0 as the avg if there's no data for a track/question just update the innermost lambda in the defaultdict initialization.

nop33: I think it makes more sense to not include the tracks with no reviews yet. So no 0 avg.

ThiefMaster: To me 0 sounds more like a bad score while none can be clearly shown differently as "there is no score"

nop33: yeah exactly

nop33: It'd be much better tho to have the AbstractReviewQuestion object as a key and not the ID, I need the text of the question anyway.

ThiefMaster: It might work if you simply replace `AbstractReviewRating.question_id` with `AbstractReviewRating.question` in the `query()` call

nop33:
### Updated layout
image

PR #2727, opened 2016-11-22
Add filtering in contribution list (display) and my abstracts list

DavidAndreev:


OmeGak:
Can we have screenshots?

DavidAndreev:
fl1
fl2

well there is not much to show :D

indico/modules/events/abstracts/templates/display/call_for_abstracts.html
```diff
@@ -123,11 +121,32 @@
...
+            {% if abstracts %}
+                {{ render_user_abstract_list(event, abstracts) }}
+            {% endif %}
+        </section>
     </div>
     <script>
+        (function() {
+            'use strict';
+
+            function setupMyAbstractsList() {
```
OmeGak: Let's move this to `abstracts.js`.

indico/modules/events/abstracts/templates/display/call_for_abstracts.html
```diff
@@ -123,11 +121,32 @@
         <section class="cfa-infoline-section">
             {{ render_cfa_infoline(event, cfa, show_icon=false) }}
         </section>
-        {% if abstracts %}
-            {{ render_user_abstract_list_section(event, abstracts, cfa) }}
-        {% endif %}
+        <section>
```
ThiefMaster: Why not keep this inside the macro? This avoids reindenting everything and also makes more sense - right now you have an empty section if the user has no abstracts.

indico/modules/events/abstracts/views.py
```diff
@@ -78,6 +78,14 @@ class WPDisplayAbstracts(WPDisplayAbstractsBase):
     menu_entry_name = 'call_for_abstracts'
 
 
+class WPDisplayCallForAbstracts(WPDisplayAbstracts):
+    template_prefix = WPDisplayAbstracts.template_prefix + 'display/'
+
+    def getJSFiles(self):
+        return (WPDisplayAbstractsBase.getJSFiles(self) +
+                self._asset_env['modules_event_display_js'].urls())
```
ThiefMaster: it doesn't fit in one line?

DavidAndreev: does
it does but its like this in all other classes as well so i thought its some kind of a pattern

ThiefMaster: Fine for me, no strong opinion here. I think usually we just ended up wrapping it since once you have more than 2 elements it usually doesn't fit in one line and then it makes sense to have one item per line.

indico/modules/events/contributions/templates/display/contribution_list.html
```diff
@@ -8,15 +8,13 @@
 
 {% block content -%}
     <div class="toolbar flex-right">
-        <!-- TODO: Filtering
         <div id="counter" class="group">
```
ThiefMaster: we don't need to show this stuff if we have no `total_entries`

Issue #2729, opened 2016-11-23
Show timezone-aware date in Timetable balloons


pferreir:
Timetable balloons are currently showing times in the event's timezone, regardless of the user's preference.

PR #2730, opened 2016-11-23
Fix bug in rendering of TZ-aware timetable

pferreir:
The balloons are still displaying the times in the event's timezone. I've opened #2729 for that.

### DIFF
--- a/indico/modules/events/timetable/legacy.py
+++ b/indico/modules/events/timetable/legacy.py
@@ -34,6 +34,7 @@ class TimetableSerializer(object):
         self.management = management
 
     def serialize_timetable(self, event, days=None, hide_weekends=False, strip_empty_days=False):
+        tzinfo = event.tzinfo if self.management else event.display_tzinfo
         event.preload_all_acl_entries()
         timetable = {}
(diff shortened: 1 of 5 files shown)


indico/modules/events/timetable/legacy.py
```diff
@@ -33,7 +33,9 @@ class TimetableSerializer(object):
     def __init__(self, management=False):
         self.management = management
 
-    def serialize_timetable(self, event, days=None, hide_weekends=False, strip_empty_days=False):
+    def serialize_timetable(self, event, days=None, hide_weekends=False, strip_empty_days=False, tzinfo=None):
+        if not tzinfo:
```
ThiefMaster: Hmm if we need to use it for offline site generation it indeed makes sense - in that case we want the event tz for the display page (even though it'll probably always match the display tz due to the lack of a user/session)

ThiefMaster:
https://gist.github.com/ThiefMaster/fbe614d0f965c98e3b01e3922e103345 fixes the timetable issue when the block starts on a different date than a contribution inside it - but no guarantees that the client-side JS displays the affected contributions nicely

ThiefMaster:
let's move this to master

PR #2731, opened 2016-11-23
Flex-ify!

nop33:
Create the new flexbox micro-framework and use it in existing templates.

### DIFF
--- a/indico/htdocs/sass/base/_layout.scss
+++ b/indico/htdocs/sass/base/_layout.scss
@@ -18,8 +18,7 @@
 @import "compass/css3/box-sizing";
 
 body {
-    display: flex;
-    flex-direction: column;
+    @extend .flexcol;
 }
 
 body > .main {
@@ -87,7 +86,7 @@ body > * {
 }
 
 .layout-side-menu {
-    display: flex;
+    @extend .flexrow;
     min-height: 100%;
     margin-left: 30px;
     margin-right: 30px;
@@ -136,27 +135,6 @@ body > * {
         width: 0;
     }
 }
-
-.flex {
-    display: flex;
-
-    &.space-between {
-        justify-content: space-between;
-    }
-}
-
-.stretch {
-    flex: 1;
-}
-
-.flex-right {
-    justify-content: flex-end;
-}
-
-.flex-center {
-    justify-content: center;
-}
-
 
 /* Flexbox layout */
 .flexrow {
--- a/indico/htdocs/sass/base/_layout.scss
+++ b/indico/htdocs/sass/base/_layout.scss
@@ -156,3 +156,70 @@ body > * {
 .flex-center {
     justify-content: center;
 }
+
+
+/* Flexbox layout */
+.flexrow {
+    flex-flow: row nowrap;
+}
+
+.flexcol {
+    flex-flow: column nowrap;
+}
+
+.flexrow,
+.flexcol {
+    display: flex !important;
+
+    &.f-j-start {
+        justify-content: flex-start;
+    }
+
+    &.f-j-end {
+        justify-content: flex-end;
+    }
+
+    &.f-j-center {
+        justify-content: center;
+    }
+
+    &.f-j-space-between {
+        justify-content: space-between;
+    }
+
+    &.f-j-space-around {
+        justify-content: space-around;
+    }
+
+    &.f-a-start {
+        align-items: flex-start;
+    }
+
+    &.f-a-end {
+        align-items: flex-end;
+    }
+
+    &.f-a-center {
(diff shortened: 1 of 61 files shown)


indico/web/templates/forms/linking_widget.html
```diff
@@ -5,7 +5,8 @@
 {% block html %}
     <div id="{{ field.id }}">
         {% for subfield in field %}
-            <div class="i-radio i-linking form-subfield{% if has_error[subfield.data] %} has-error{% endif %}">
+            <div class="i-radio i-linking form-subfield flexrow f-a-center
+                {%- if has_error[subfield.data] %} has-error{% endif %}">
```
ThiefMaster: I'd change this to `{{ 'has-error' if has_error[subfield.data] }}` (and not bother about the whitespace)

indico/modules/events/timetable/templates/move_entry.html
```diff
@@ -27,7 +27,8 @@
         <div class="i-box-content">
             {% for day, tt_entries in top_level_entries.iteritems() %}
                 <ul class="group-list blocks-list" data-show-for="{{ day }}">
-                    <li class="item top-level {%- if parent_entry is none and day == current_day %} selected{% endif %}"
+                    <li class="item top-level flexrow f-a-center
+                        {%- if parent_entry is none and day == current_day %} selected{% endif %}"
```
ThiefMaster: `{{ 'selected' if parent_entry is none and day == current_day }}`

indico/modules/events/logs/templates/logs.html
```diff
@@ -16,15 +16,17 @@
...
                     <span class="icon-stack-plus"></span>
                 </a>
                 <a href="#" id="collapse-all" class="i-button" title="{% trans %}Collapse all{% endtrans %}">
                     <span class="icon-stack-minus"></span>
                 </a>
             </div>
-            <div class="group right">
+        </div>
+        <div class="toolbar">
+            <div class="group ">
```
ThiefMaster: extra whitespace

indico/htdocs/sass/base/_layout.scss
```diff
@@ -137,22 +136,64 @@ body > * {
     }
 }
 
-.flex {
-    display: flex;
+/* Flexbox layout */
+.flexrow {
```
nop33: Moving conversation from #2726 :
> ThiefMaster a day ago
> I still think this would be clearer:
> [18:37:52] <@ThiefMaster> flexrow -> .flex.row?
> [18:38:05] <@ThiefMaster> ie .flex to set display:flex and .col/.row to set the flow

> OmeGak 23 hours ago 
I don't think we should provide every class to be orthogonal with the full expressiveness of flex-boxes. flewrow` is just a class for an opinionated layout (flex row non-wrapping).

indico/htdocs/sass/base/_layout.scss
```diff
@@ -137,22 +136,64 @@ body > * {
...
+    }
+
+    &.f-wrap {
+        flex-wrap: wrap;
+    }
 }
 
-.flex-center {
-    justify-content: center;
+.stretch {
```
nop33: Moving conversation:
>OmeGak 23 hours ago 
Maybe it's a good moment to rename stretch to f-grow.

nop33: How about giving the prefix `f-self-` to all children-related properties?
So it would be `f-self-grow`.

indico/modules/events/tracks/templates/management.html
```diff
@@ -38,16 +38,15 @@
         </div>
     </div>
 
-    {# TODO: Use new flex classes instead of the inline CSS once they are available #}
-    <div class="flex" style="flex-direction: row;">
-        <h3 style="flex-grow: 1; margin-bottom: 0;">{% trans %}List of tracks{% endtrans %}</h3>
-        <div style="margin: 1em 0 0 0;">
+    <div class="flexrow f-a-center">
+        <h3 class="stretch">{% trans %}List of tracks{% endtrans %}</h3>
+        <div>
```
nop33: No, but it should normally be a toolbar... However, in this case since the button is only one, we could work with just that (without a toolbar)

indico/modules/events/logs/templates/logs.html
nop33: You got a nice pair of :eyes: !

indico/htdocs/sass/base/_layout.scss
nop33: Guys, do you agree with this? Should I add it and replace all the occurrences in the templates?

PR #2732, opened 2016-11-23
Untrusted/pending event persons

ThiefMaster:
- [x] Handle pending event persons in the backend
- [x] Create pending event persons when creating abstracts
- [x] Mark them as not pending as soon as they are associated with anything else
- [x] Show persons linked only to abstracts on "roles" page (currently not shown at all)
- [x] Indicate pending event persons on that page (with a tooltip explaining it)
- [x] Make pending person indicator pretty

SQL to run:

```sql
ALTER TABLE events.persons ADD COLUMN is_untrusted boolean NOT NULL DEFAULT false;
ALTER TABLE events.persons ALTER COLUMN is_untrusted DROP DEFAULT;
```

### DIFF
--- a/indico/htdocs/sass/modules/_event_management.scss
+++ b/indico/htdocs/sass/modules/_event_management.scss
@@ -503,9 +503,7 @@ th.i-table {
         max-width: 20em;
         min-width: 20em;
     }
-}
 
-.event-person-list {
     .no-role {
         opacity: 0.25;
         .badge {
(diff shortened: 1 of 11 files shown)


indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -505,6 +505,13 @@ th.i-table {
         }
     }
 
+    .pending {
+        background-color: $light-gray;
+        td {
```
nop33: Empty line above.

PR #2733, opened 2016-11-24
Show user prefered timezone in timetable balloon

DavidAndreev:
Don't take this too seriously. I'm not sure if I'm doing it right cause I don't know what that DisplayTZ does, but it seams to fix the problem. 
works
But when I try in management timetable it shows the right time but the time rows of timetable are not right. but maybe thats that thing that Pedro was fixing.
doesnt work

### DIFF
--- a/indico/modules/events/timetable/templates/balloons/_balloon_utils.html
+++ b/indico/modules/events/timetable/templates/balloons/_balloon_utils.html
@@ -1,5 +1,7 @@
-{% macro render_time_and_location(obj, editable) %}
+{% macro render_time_and_location(obj, editable, use_event_tz) %}
     {% set tt_entry = obj.timetable_entry %}
+    {% set event = tt_entry.event_new %}
+    {% set tz = event.tzinfo if use_event_tz else event.display_tzinfo %}
     <dl class="i-data-list">
         {% if editable %}
             <dt>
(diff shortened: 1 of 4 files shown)


ThiefMaster:
Don't use `DisplayTZ` directly - there's ` display_tzinfo` property on `Event` that wraps it nicely.

ThiefMaster:
Also, in management we always want the event's timezone - only in display areas we want the "display timezone" (that takes the user's timezone preference into account)

pferreir:
Maybe this should include my related PR? Feel free to do it as you guys
wish.

DavidAndreev:
So is there a way to check if we are in management from within the template and based on that I will show the proper time zone ?

ThiefMaster:
Check when we pass `editable=true` to the macro. That might be enough to check.

indico/modules/events/timetable/util.py
```diff
@@ -19,11 +19,13 @@
 from collections import defaultdict
 from operator import attrgetter
 
+from babel.dates import get_timezone
 from flask import render_template, session
 from pytz import utc
 from sqlalchemy import Date, cast
 from sqlalchemy.orm import joinedload, subqueryload, undefer, contains_eager
 
+from MaKaC.common.timezoneUtils import DisplayTZ
```
ThiefMaster: unused

indico/modules/events/timetable/util.py
```diff
@@ -19,11 +19,13 @@
 from collections import defaultdict
 from operator import attrgetter
 
+from babel.dates import get_timezone
```
ThiefMaster: unused

ThiefMaster:
LGTM, but just like #2730 I'd rather merge this into v1.9.8 - at least if we want to hotfix it in prod.

ThiefMaster:
Added to #2730

PR #2736, opened 2016-11-25
Add sass-lint config

ThiefMaster:
Usage:
 - Install sass-lint using `npm install -g sass-lint`
 - Run `sass-lint -vq [files...]` - due to lots of errors/warning from old files it's a good idea to specify only files you modified/added as `[files...]` or to use git pre-commit hook to do that for you.

If you are using the `.git/hooks/pre-commit` hook, you can enable the check by adding the following code right before `if [[ $STATUS != 0 ]] ; then`:

```bash
FILES=$(git diff --staged --name-status HEAD '*.scss' | awk '$1 != "D" {print $2}' | xargs)
if [[ -n "$FILES" ]]; then
    RESULT=`sass-lint -vq $FILES`
    RC=$?
    if [[ $RC != 0 ]] ; then
        STATUS=1
        echo "${RED}There are SCSS errors in your code:${RESET}"
        echo "$RESULT"
    elif [[ -n "$RESULT" ]] ; then
        STATUS=1
        echo "${BRIGHTYELLOW}There are SCSS warnings in your code:${RESET}"
        echo "$RESULT"
    fi
fi
```

Full pre-commit hook: https://gist.github.com/ThiefMaster/7ca3c48ef93604be0478e30852d0c444

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -1,0 +1,130 @@
+options:
+  merge-default-rules: false
+
+files:
+  include: 'indico/htdocs/sass/**/*.scss'
+  ignore: 'indico/htdocs/sass/lib/**/*'
+
+rules:
+  attribute-quotes: 1
+  bem-depth: 0
+  border-zero:
+    - 1
+    - convention: none
+  brace-style:
+    - 2
+    - style: 1tbs
+      allow-single-line: false
+  class-name-format:
+    - 1
+    - convention: hyphenatedlowercase
+      allow-leading-underscore: false
+  clean-import-paths: 2
+  declarations-before-nesting: 1
+  empty-args:
+    - 1
+    - include: true
+  empty-line-between-blocks: 1
+  extends-before-declarations: 2
+  extends-before-mixins: 2
+  final-newline: 2
+  force-attribute-nesting: 0
+  force-element-nesting: 0
+  force-pseudo-nesting: 0
+  function-name-format:
+    - 2
+    - convention: hyphenatedlowercase
+      allow-leading-underscore: true
+  hex-length:
+    - 1
+    - style: short
+  hex-notation:
+    - 1
+    - style: lowercase
+  id-name-format:
+    - 1
+    - convention: hyphenatedlowercase
+      allow-leading-underscore: false
+  indentation:
+    - 2
+    - size: 4
+  leading-zero:
+    - 2
+    - include: true
+  max-file-line-count:
+    - 1
+    - length: 750
+  max-line-length:
+    - 2
+    - length: 120
+  mixin-name-format:
+    - 2
+    - convention: hyphenatedlowercase
+      allow-leading-underscore: true
+  mixins-before-declarations: 1
+  nesting-depth:
+    - 1
+    - max-depth: 4
+  no-attribute-selectors: 0
+  no-color-hex: 0
+  no-color-keywords: 0
+  no-color-literals: 0
+  no-combinators: 0
+  no-css-comments: 0
+  no-debug: 1
+  no-disallowed-properties: 0
+  no-duplicate-properties:
+    - 1
+    - exclude: [background-image]
+  no-empty-rulesets: 2
+  no-extends: 0
+  no-ids: 1
+  no-important: 0
+  no-invalid-hex: 2
+  no-mergeable-selectors: 1
+  no-misspelled-properties: 2
+  no-qualifying-elements:
+    - 1
+    - allow-element-with-attribute: true
+      allow-element-with-class: true
+      allow-element-with-id: false
+  no-trailing-whitespace: 2
+  no-trailing-zero: 2
+  no-transition-all: 0
+  no-universal-selectors: 0
+  no-url-domains: 2
+  no-url-protocols: 2
+  no-vendor-prefixes: 0
+  no-warn: 1
+  one-declaration-per-line: 2
+  placeholder-in-extend: 0
+  placeholder-name-format:
+    - 2
+    - convention: hyphenatedlowercase
+      allow-leading-underscore: false
+  property-sort-order: 0
+  property-units: 0
+  pseudo-element: 1
+  quotes:
+    - 1
+    - style: single
+  shorthand-values:
+    - 1
+    - allowed-shorthands: [1, 2]
+  single-line-per-selector: 0
+  space-after-bang: 2
+  space-after-colon: 2
+  space-after-comma: 2
+  space-around-operator: 0
+  space-before-bang: 2
+  space-before-brace: 2
+  space-before-colon: 2
+  space-between-parens: 2
+  trailing-semicolon: 2
+  url-quotes: 2
+  variable-for-property: 0
+  variable-name-format:
+    - 2
+    - convention: hyphenatedlowercase
+      allow-leading-underscore: false
+  zero-unit: 2
--- a/indico/htdocs/sass/partials/_badges.scss
+++ b/indico/htdocs/sass/partials/_badges.scss
@@ -27,7 +27,6 @@
 .i-badges-list {
     @extend %badges-list;
     text-align: center;
-    -webkit-justify-content: space-around;
     justify-content: space-around;
 
     > .i-badge:first-child {
@@ -37,25 +36,16 @@
 
 .i-badges-list.left {
     @extend %badges-list;
-
     float: none; // disable default .left style
-
     text-align: left;
-
-    //TODO replace with justify-content mixin once Compass is updated
-    -webkit-justify-content: flex-start;
-            justify-content: flex-start;
+    justify-content: flex-start;
 }
 
 .i-badges-list.center {
     @extend %badges-list;
-
     margin: -0.8em 0 0.8em 0; // disable default .center style
-
     text-align: center;
-
-    -webkit-justify-content: center;
-            justify-content: center;
+    justify-content: center;
 }
 
 
@@ -76,12 +66,16 @@
             height: 7em;
 
             &:first-child {
-               height: 8.8em;
+                height: 8.8em;
             }
 
-            & > .i-badge-title { font-size: 1em; }
+            & > .i-badge-title {
+                font-size: 1em;
+            }
 
-            & .i-badge-img > img { max-height: 4em; }
+            & .i-badge-img > img {
+                max-height: 4em;
+            }
         }
     }
 }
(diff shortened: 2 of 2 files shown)


ThiefMaster:
The rules are explained here btw in case anyone has better suggestions: https://github.com/sasstools/sass-lint/tree/master/docs/rules

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+    - 1
+    - convention: none
+  brace-style:
+    - 2
+    - style: 1tbs
+      allow-single-line: false
+  class-name-format:
+    - 1
+    - convention: hyphenatedlowercase
+      allow-leading-underscore: false
```
OmeGak: Won't this disallow `%_foo`?

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+  clean-import-paths: 2
+  declarations-before-nesting: 1
+  empty-args:
+    - 1
+    - include: true
+  empty-line-between-blocks: 1
+  extends-before-declarations: 2
+  extends-before-mixins: 1
```
OmeGak: I would set this to error as well.

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+    - 1
+    - length: 750
+  max-line-length:
+    - 2
+    - length: 120
+  mixin-name-format:
+    - 2
+    - convention: hyphenatedlowercase
+      allow-leading-underscore: true
+  mixins-before-declarations: 0
```
OmeGak: Why not flagging this as warning?

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+  nesting-depth:
+    - 1
```
OmeGak: No error? I don't think more than 4 levels of nesting is something we should allow.

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+    - max-depth: 4
+  no-attribute-selectors: 0
+  no-color-hex: 0
+  no-color-keywords: 0
+  no-color-literals: 0
+  no-combinators: 0
+  no-css-comments: 0
```
OmeGak: Why not avoid `/* ... */`?

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+  no-debug: 1
+  no-disallowed-properties: 0
+  no-duplicate-properties:
+    - 1
+    - exclude: [background-image, display]
```
OmeGak: Why would you allow duplicate `display`?

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+  no-empty-rulesets: 2
+  no-extends: 0
+  no-ids: 0
```
OmeGak: What about warning about this?

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+  no-important: 0
+  no-invalid-hex: 2
+  no-mergeable-selectors: 1
```
OmeGak: There are places where it's desirable not to merge selectors. One example is my recent PR #2737 that adds `.i-box.header-only`. There I add some rules to `.i-box` that are necessary only so that transitions work in `.i-box.header-only` and keeping it like this helps maintainability.

Now, maybe this use-case is not common enough to disable this rule. Thoughts?

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+  no-trailing-whitespace: 2
+  no-trailing-zero: 2
+  no-transition-all: 0
+  no-universal-selectors: 0
+  no-url-domains: 2
+  no-url-protocols: 2
+  no-vendor-prefixes: 0
+  no-warn: 1
+  one-declaration-per-line: 2
+  placeholder-in-extend: 0
```
OmeGak: I'm starting to think that extending non-placeholder classes is a mistake in most cases. Maybe we should discuss this point IRL.

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+  placeholder-name-format:
+    - 2
+    - convention: hyphenatedlowercase
+      allow-leading-underscore: false
+  property-sort-order: 0
```
OmeGak: Dunno, maybe we should consider at some point: https://github.com/sasstools/sass-lint/blob/master/docs/rules/property-sort-order.md

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+  property-units: 0
+  pseudo-element: 0
```
OmeGak: I think we want this to be a warning. https://github.com/sasstools/sass-lint/blob/master/docs/rules/pseudo-element.md

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+  quotes:
+    - 1
+    - style: single
+  shorthand-values:
+    - 1
+    - allowed-shorthands: [1, 2]
+  single-line-per-selector: 0
```
OmeGak: I would set this to warning. It's much clearer with one selector per line.

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+  space-after-bang: 2
+  space-after-colon: 2
+  space-after-comma: 2
+  space-around-operator: 0
```
OmeGak: Don't we want it?

.sass-lint.yml
```diff
@@ -0,0 +1,130 @@
...
+  space-before-brace: 2
+  space-before-colon: 2
+  space-between-parens: 2
+  trailing-semicolon: 2
+  url-quotes: 2
+  variable-for-property: 0
+  variable-name-format:
+    - 2
+    - convention: hyphenatedlowercase
+      allow-leading-underscore: false
```
OmeGak: I would allow underscored vars to indicate locality.

.sass-lint.yml
ThiefMaster: That's not a class name, and we don't have any underscored classes so far.

OmeGak:
What is the rationale for choosing between warning and error?

.sass-lint.yml
ThiefMaster: Stuff like `border-all` doesn't really make sense to be before everything else IMO

.sass-lint.yml
ThiefMaster: Mainly because of our license headers which use them

.sass-lint.yml
ThiefMaster: I've seen a few places where we seem to be using it for compatibility with older browsers. But yeah I can check again, maybe we won't need it

.sass-lint.yml
ThiefMaster: I checked the places where it was failing and it didn't seem like something problematic. Either way, to me it's more something like "check again if it's really needed and consider changing it", not "this is always bad and you shouldn't do it"

.sass-lint.yml
ThiefMaster: I was thinking about that, but we do use IDs in quite a few places. I really wouldn't mind switching this to warnings and not using IDs for styling at all...

.sass-lint.yml
ThiefMaster: Exactly my thought, and also the reason why it's just a warning. It's good at spotting accidental fuckups (I looked at the output of all warnings/errors from a rule, and saw quite a few places where we had the same selector right after itself, i.e. the warning was quite useful)

.sass-lint.yml
ThiefMaster: For very simple/short selectors  I tend to disagree that it's much clearer...

.sass-lint.yml
ThiefMaster: It's the kind of thing I'd consider changing when there's the need to use them. So far we never used any local variables in CSS

.sass-lint.yml
ThiefMaster: It's mostly for `calc()` stuff and at least to me it looked better without spaces.

ThiefMaster:
Mainly how likely it is that there is a good reason to break the rule and/or how often we are currently breaking it (even though the latter might not be a good reason to not use error)

.sass-lint.yml
nop33: I'd also like to see the mixins on the top of the selector's body, makes it clearer imo

ThiefMaster: You're right, I somehow thought we had `border-*` mixins further down

ThiefMaster: Setting it to warning. Error is too much as there could be cases where you do want a mixin later in case you want a more generic declaration first (e.g. a mixin that sets margin-bottom but you first want to zero all other margins with `margin: 0`).

.sass-lint.yml
ThiefMaster: This is why I added it:

https://github.com/indico/indico/blob/master/indico/htdocs/sass/partials/_badges.scss#L132-L136

```css
//TODO replace with flexbox mixin once Compass is updated
display: block;
text-align: center;
display: -webkit-inline-flex;
display: inline-flex;
```

.sass-lint.yml
OmeGak: Let's keep the warning, then. We can always disable it when we need something like my use case.

.sass-lint.yml
ThiefMaster:
```
$switch-handle-transition: 2*$switch-transition-step ease-out;
$switch-handle-transition: 2 * $switch-transition-step ease-out;
```

I think it's clearer what belongs together without thhe space

.sass-lint.yml
ThiefMaster: Something we could consider changing with the end-of-year header update. It's a one-line change in the header update script:

```diff
diff --git a/.sass-lint.yml b/.sass-lint.yml
index 35bcc4b48..6c9322c2f 100644
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -70,7 +70,7 @@ rules:
   no-color-keywords: 0
   no-color-literals: 0
   no-combinators: 0
-  no-css-comments: 0
+  no-css-comments: 1
   no-debug: 1
   no-disallowed-properties: 0
   no-duplicate-properties:
diff --git a/bin/maintenance/update_header.py b/bin/maintenance/update_header.py
index 666d5ce2e..d78aac731 100644
--- a/bin/maintenance/update_header.py
+++ b/bin/maintenance/update_header.py
@@ -70,8 +70,8 @@ SUPPORTED_FILES = {
         'regex': re.compile('/\*(.|[\r\n])*?\*/'),
         'format': {'comment_start': '/*', 'comment_middle': ' *', 'comment_end': ' */'}},
     'scss': {
-        'regex': re.compile('/\*(.|[\r\n])*?\*/'),
-        'format': {'comment_start': '/*', 'comment_middle': ' *', 'comment_end': ' */'}},
+        'regex': re.compile('/\*(.|[\r\n])*?\*/|((^//|[\r\n]//).*)*'),
+        'format': {'comment_start': '//', 'comment_middle': '//', 'comment_end': ''}},
     'xsl': {
         'regex': re.compile('<!--(.|[\r\n])*?-->'),
         'format': {'comment_start': '<!--\n   ', 'comment_middle': '   ', 'comment_end': '-->'}},
```

.sass-lint.yml
ThiefMaster: going for warnings

PR #2737, opened 2016-11-25
Improvements and fixes

OmeGak:
Main changes:
* Moving the judgment information to always be below the timeline
* UI of proposed action, score, show-ratings, and track information in review items.

screenshot from 2016-11-25 15 54 42

### DIFF
--- a/indico/htdocs/js/indico/modules/abstracts.js
+++ b/indico/htdocs/js/indico/modules/abstracts.js
@@ -73,10 +73,25 @@
             $reviewBox.find('.js-edit-review').trigger('click');
             $('body, html').animate({scrollTop: $reviewBox.offset().top}, 'fast');
             $reviewBox.find('.i-timeline-item-box').effect('highlight', {color: Palette.highlight}, 'slow');
-        }).on('click', '.ratings-switch .i-button', function() {
+        }).on('click', '.js-ratings-toggle', function() {
             var $this = $(this);
+            $this.find('.js-show-ratings, .js-hide-ratings').toggleClass('weak-hidden');
+            var $reviewBox = $this.closest('.i-timeline-item-box');
+            if ($reviewBox.data('no-comment') !== undefined) {
+                if ($reviewBox.hasClass('header-only')) {
+                    $reviewBox.removeClass('header-only');
+                    $reviewBox.find('.ratings-details').slideToggle('fast');
+                } else {
+                    $reviewBox.addClass('header-only-transition');
+                    $reviewBox.find('.ratings-details').slideToggle('fast', function() {
+                        $reviewBox.removeClass('header-only-transition');
+                        $reviewBox.addClass('header-only');
+                    });
+                }
+            } else {
+                $reviewBox.find('.ratings-details').slideToggle('fast');
+            }
             $this.toggleClass('open');
-            $this.closest('.ratings-switch').siblings('.ratings-details').toggle();
         }).on('click', '.js-highlight-review', function() {
             $($(this).attr('href') + ' .i-timeline-item-box').effect('highlight', {color: Palette.highlight}, 'slow');
         });
--- a/indico/htdocs/js/indico/modules/abstracts.js
+++ b/indico/htdocs/js/indico/modules/abstracts.js
@@ -93,7 +93,7 @@
             }
             $this.toggleClass('open');
         }).on('click', '.js-highlight-review', function() {
-            $($(this).attr('href') + ' .i-timeline-item-box').effect('highlight', {color: Palette.highlight}, 'slow');
+            $($(this).attr('href') + ' .i-box-header').effect('highlight', {color: Palette.highlight}, 'slow');
         });
     };
 
--- a/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
+++ b/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
@@ -102,13 +102,6 @@
     max-width: 10em;
 }
 
-.abstract-track-tag {
-    @extend .flexrow;
-    max-width: 15em;
-    margin: -0.35em 0;
-    margin-left: 1em;
-}
-
 .abstract-content {
     @extend .i-box;
     @extend .i-box.no-shadow;
@@ -205,6 +198,15 @@
 .abstract-review-badges {
     flex-grow: 1;
     color: $gray;
+
+    .js-ratings-toggle {
+        color: $gray;
+        user-select: none;
+
+        &:hover {
+            color: $dark-gray;
+        }
+    }
 }
 
 .abstract-comment-actions {
@@ -443,7 +445,7 @@
         font-weight: bold;
         background: $pastel-gray;
         color: $light-black;
-        padding: 1px 6px;
+        padding: 0 6px;
     }
 
     .question-text {
--- a/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
+++ b/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
@@ -218,6 +218,7 @@
 }
 
 #abstract-submission-details,
+#abstract-judgment-details,
 #abstract-timeline {
     .i-timeline-item-label {
         color: rgba(#fff, 0.8);
--- a/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
+++ b/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
@@ -56,9 +56,11 @@
     }
 }
 
-#abstract-title {
+.abstract-title {
+    margin-right: 0.5em;
+
     h3 {
-        color: $blue;
+        color: $blue !important;
         font-size: 1.4em;
         margin: 0;
     }
--- a/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
+++ b/indico/htdocs/sass/modules/abstracts/_abstract_page.scss
@@ -22,6 +22,7 @@
 
 .abstract-page {
     @extend %font-family-body;
+    position: relative;
 
     .i-form {
         margin-top: 0;
@@ -45,6 +46,14 @@
             margin-bottom: 0;
         }
     }
+
+    .open-abstract-notification-log {
+        margin-top: 5px;
+        margin-right: 10px;
+        position: absolute;
+        right: 0;
+        z-index: 1;
+    }
 }
 
 #abstract-title {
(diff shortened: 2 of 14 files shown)


ThiefMaster:
the "change Tracks" capitalization looks pretty weird. I'd go for "change tracks"

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
```diff
@@ -217,6 +218,7 @@
 }
 
 #abstract-submission-details,
+#abstract-judgment-details,
```
ThiefMaster: sounds like a common class for those 3 elements would be the better choice

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -427,101 +427,87 @@ <h3 class="abstract-title f-self-stretch">
...
+                            The abstract was accepted and a {{ link_start }}contribution{{ link_end }} was created in the event.
+                        {%- endtrans %}
+                    {% endif %}
+                {% elif abstract.state.name == 'rejected' %}
+                    {% trans %}The abstract was rejected.{% endtrans %}
+                {% elif abstract.state.name == 'merged' %}
+                    {% trans url=url_for('.display_abstract', abstract.merged_into), other=abstract.merged_into.title -%}
+                        The abstract was merged into {{ link_start }}{{ other }}{{ link_end }}.
+                    {%- endtrans %}
+                {% elif abstract.state.name == 'duplicate'%}
```
ThiefMaster: missing space

indico/modules/events/abstracts/templates/abstract/timeline.html
```diff
@@ -96,10 +103,11 @@
                     {% endif %}
                 </div>
                 <div>
-                    {{ render_track(review.track, classes='abstract-track-tag i-tag outline') }}
+                    {{ render_track(review.track) }}
                 </div>
             </div>
-            <div class="i-timeline-item-box header-indicator-top">
+            <div class="i-timeline-item-box header-indicator-top {% if not review.comment %}header-only{% endif %}"
+                 {% if not review.comment %}data-no-comment{% endif %}>
```
ThiefMaster: What is this used for? Can't find any other reference to `no-comment`?

indico/modules/events/abstracts/templates/abstract/timeline.html
```diff
@@ -144,12 +152,17 @@
...
                         {% if review.score is not none %}
-                            {{ render_score(review.score, review.abstract.event_new.cfa) }}
+                            · score {{ render_score(review.score, review.abstract.event_new.cfa, classes='bold semantic-text') }}
+                            (<a class="js-ratings-toggle">{#--#}
+                                <span class="js-show-ratings">{% trans %}show ratings{% endtrans %}</span>{#--#}
+                                <span class="weak-hidden js-hide-ratings">{% trans %}hide ratings{% endtrans %}</span>{#--#}
+                            </a>)
                         {% endif %}
                     </div>
+                    <div class="ratings-switch">
```
ThiefMaster: empty div? if that's correct i'd move the closing tag in the same line

indico/modules/events/abstracts/templates/abstract/timeline.html
OmeGak: https://github.com/indico/indico/pull/2737/files#diff-79e7f4b3fd8bcaaad14450267448d1beR80

indico/modules/events/abstracts/templates/abstract/timeline.html
OmeGak: Just forgot to remove it, thanks.

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
OmeGak: `_abstract_page.scss` would help a cleanup before end of the release...

indico/modules/events/abstracts/templates/abstract/timeline.html
ThiefMaster: I think in the last months we always went for not using the camelCase version in `data()` so I'd go for `data('no-comment')` instead. That way it's more consistent and people can easily search for it instead of trying both `no-comment` and `noComment`

indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -73,12 +73,27 @@
...
             $('body, html').animate({scrollTop: $reviewBox.offset().top}, 'fast');
             $reviewBox.find('.i-timeline-item-box').effect('highlight', {color: Palette.highlight}, 'slow');
-        }).on('click', '.ratings-switch .i-button', function() {
+        }).on('click', '.js-ratings-toggle', function() {
             var $this = $(this);
+            $this.find('.js-show-ratings, .js-hide-ratings').toggleClass('weak-hidden');
+            var $reviewBox = $this.closest('.i-timeline-item-box');
+            if ($reviewBox.data('noComment') !== undefined) {
+                if ($reviewBox.hasClass('header-only')) {
+                    $reviewBox.toggleClass('header-only');
```
nop33: Since you check for `hasClass` above, I'd rather have a `removeClass` here since it's essentially what happens.

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
```diff
@@ -206,6 +198,15 @@
 .abstract-review-badges {
     flex-grow: 1;
     color: $gray;
+
+    .js-ratings-toggle {
```
nop33: Is it OK to use `js-` related classes for styling? :thinking:

indico/htdocs/sass/partials/_boxes.scss
```diff
@@ -282,6 +282,36 @@ $i-box-padding: 10px;
     }
 }
 
+.i-box {
```
nop33: There are now 2 `.i-box {` in the same level...

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
ThiefMaster: No, should be without `js-` in that case.

indico/htdocs/sass/partials/_boxes.scss
OmeGak: This is intended. I think this makes easier to understand that `@include transition` is there because of the `.header-only` set of rules.

ThiefMaster: sass-lint will call you out for that! ;)

I think it'd make sense to combine the two

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
OmeGak: It's the only class I have in the element. Would you artificially add another one? I think the idea with `js-*` classes is to make clear the class is needed for JS to trigger something.

indico/modules/events/abstracts/templates/abstract/timeline.html
```diff
@@ -144,10 +152,12 @@
                                 {%- endtrans %}
                             {% endif %}
                         {% endif %}
-                    </div>
-                    <div class="abstract-comment-score">
                         {% if review.score is not none %}
-                            {{ render_score(review.score, review.abstract.event_new.cfa) }}
+                            · score {{ render_score(review.score, review.abstract.event_new.cfa, classes='bold semantic-text') }}
+                            (<a class="js-ratings-toggle">{#--#}
```
nop33: What is the `{#--#}` doing exactly? Don't you want to put the `-#}` in the next line to remove the whitespace?

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
ThiefMaster: No, I'd use the non-js class in JS. The only point of `js-` classes is to make it clear that it's not used for styling - but not the other way round, i.e. a normal class may be used for both.

indico/modules/events/abstracts/templates/abstract/timeline.html
ThiefMaster: It's less ugly that way. Since whitespace is stripped on both the left `{#-` and right `-#}` side of the jinja comment there is no need to actually comment out the linebreak it self.

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
OmeGak: It's funny how we have developed different understanding of `js-*` classes and I think we haven't been consistent about their usage. There are cases where we add an extra `js-*` class only to bind something to the element in JS. Maybe we should discuss it in the next meeting and write down the conclusion on the wiki.

pferreir: I never use them for styling either.

indico/modules/events/abstracts/templates/abstract/public.html
```diff
@@ -405,7 +405,7 @@ <h3 class="abstract-title f-self-stretch">
                 <div class="i-timeline-item-label action" style="{{ abstract.submitter.avatar_css }}">
                     {{- abstract.submitter.name[0] -}}
                 </div>
-                <div class="i-timeline-item-box header-indicator">
+                <div class="i-timeline-item-box header-indicator-left">
```
pferreir: 👍

indico/modules/events/abstracts/templates/abstract/timeline.html
pferreir: > I think in the last months we always went for not using the camelCase version in `data()`

"We" as in "I"? 😉 
Also, if it's something worth changing in the PR, let's add it to the wiki :)

ThiefMaster: I think it was actually you starting it, I always used the camelCase versions :P

indico/htdocs/sass/modules/abstracts/_abstract_page.scss
ThiefMaster: The way I always used/saw them is this:

- If there's a class that is unique and clear enough when used in JS I use that one. `toggle-something` for example.
- If there's no such class I add a `js-*` one.
- I never use "generic" style classes (like `icon-delete`) for JS bindings, as nobody would expect technical side-effects when changing/removing an icon

indico/modules/events/abstracts/templates/abstract/timeline.html
```diff
@@ -228,7 +245,7 @@
                       'merge': 'visited',
                       'mark_as_duplicate': 'strong',
                       'change_tracks': 'warning'} %}
-    <span class="i-tag outline {{ mapping[review.proposed_action.name] }}">
+    <span class="bold underline semantic-text {{ mapping[review.proposed_action.name] }}">
```
OmeGak: I was about to go for that initially, but is it worth adding the class just for `bold` and `underline`? We need to add `.semantic-text` anyway to apply the color variations from the mapping.

PR #2738, opened 2016-11-28
Update display abstract list and abstract page decision box

nop33:


indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -29,7 +29,7 @@
     </td>
 {%- endmacro %}
 
-{% macro render_abstract_list(abstracts, dynamic_columns, static_columns, total_abstracts, context_track=None) %}
+{% macro render_abstract_list(abstracts, dynamic_columns, static_columns, total_abstracts, context_track=None, reviewed_abstracts=None) %}
```
ThiefMaster: `none` (also for `context_track`) - in jinja lowercase is preferred for none/true/false

indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -56,7 +56,10 @@
                             {% for item in dynamic_columns %}
                                 <th class="i-table">{{ item.title }}</th>
                             {% endfor %}
-                            <th class="i-table thin-column"></th>
+                            {% if reviewed_abstracts %}
```
ThiefMaster: Wouldn't `is not none` make more sense here? Otherwise you don't have the column if you are in the review area but haven't reviewed anything

indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -171,6 +174,15 @@
                                         {{- friendly_data if friendly_data else '-' -}}
                                     </td>
                                 {% endfor %}
+                                {% if reviewed_abstracts %}
```
ThiefMaster: same here

indico/modules/events/abstracts/templates/management/_abstract_list.html
nop33: well, why do you need an empty column? If you haven't reviewed anything there will be nothing to show on that column. I'll post a screenshot of how it looks like.

ThiefMaster: And it wouldn't have to be empty. Could be grayed-out for unreviewed abstracts

nop33:
image

image

indico/modules/events/abstracts/lists.py
```diff
@@ -251,3 +252,16 @@ def __init__(self, event, track):
     def _build_query(self):
         return (super(AbstractListGeneratorDisplay, self)._build_query()
                 .filter(Abstract.reviewed_for_tracks.contains(self.track)))
+
+    def get_user_reviewed_abstracts_for_track(self, user, track):
+        return (db.session.query(Abstract)
```
ThiefMaster: `Abstract.query`

indico/modules/events/abstracts/templates/management/_abstract_list.html
nop33: image

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -47,10 +47,41 @@
     </div>
 {% endmacro %}
 
+{% macro render_decision_box(abstract, form, user) %}
+    {% set event = abstract.event_new %}
+    <div class="i-timeline-item-label action" style="{{ user.avatar_css }}">
+        {{- user.name[0] -}}
```
pferreir: We're probably doing this somewhere else, right? Maybe a macro?

nop33: We don't have a macro for it, but I can make one, it would make sense.

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -47,10 +47,41 @@
...
+    </div>
+    <div class="i-timeline-item-box footer-only header-indicator-left">
+        <div class="i-box-content track-reviews-wrapper">
```
OmeGak: Did you forget to remove this one?

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -139,31 +170,9 @@
...
-            {% call form_footer(form) %}
-                <button class="i-button big highlight">
-                    {% trans %}Judge{% endtrans %}
-                </button>
-            {% endcall %}
-        </div>
     </div>
 {% endmacro %}
 
-
```
OmeGak: This is starting to become another thing to write an entry in the wiki. Are we adding two blank lines before macros and blocks or not? I think having two makes it slightly clearer.

indico/modules/events/abstracts/templates/management/_abstract_list.html
OmeGak: I wouldn't gray out the icon, as if you haven't reviewed anything it may look like all of them are reviewed. OTOH, I would not hide the column.

indico/modules/events/abstracts/templates/abstract/judge.html
ThiefMaster: Around big macros/blocks it makes sense. But I don't think two blank lines between single-line blocks (`title` in many cases, or `content_class`) are really necessary

indico/modules/events/abstracts/templates/abstract.html
```diff
@@ -20,12 +21,19 @@
     {{ render_abstract_public(abstract) }}
     <div id="abstract-private" class="i-timeline">
         {{ render_timeline(abstract, comment_form, review_form) }}
+        {% if abstract.can_judge(session.user, check_state=true) and abstract.event_new.tracks %}
+            {{ render_reviewed_for_tracks_box(abstract, review_track_list_form) }}
+        {% endif %}
+
+        {% if abstract.can_see_reviews(session.user) %}
+            <div id="abstract-review-summary-box" class="i-timeline-item">
+                {{ render_review_summary_box(abstract, judgment_form) }}
```
OmeGak: We cannot do this. This assumes any convener can see reviews in **all** tracks. We need to modify `render_review_summary_box` to show only the the tracks that apply.

indico/modules/events/abstracts/templates/management/_abstract_list.html
nop33: Why have a column when there's nothing there? I find it pointless.

ThiefMaster: Consistency. Why would reviewing something suddenly cause a column to appear in that list?

nop33: Cause otherwise the column does not serve any purpose. If most of the team agrees on adding the column even if it's empty, I'll gladly do it! @pferreir ?

indico/modules/events/abstracts/templates/abstract.html
nop33: Done.

indico/modules/events/abstracts/templates/abstract/judge.html
nop33: Why remove it?

indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -47,10 +48,37 @@
...
+                <button class="i-button big highlight">
+                    {% trans %}Judge{% endtrans %}
+                </button>
+            {% endcall %}
+        </div>
+    </div>
+{% endmacro %}
 
-{% macro render_decision_box(abstract, form) %}
+{% macro render_review_summary_box(abstract, form, user, convener_tracks) %}
```
OmeGak: We don't need the `form` here anymore.

indico/modules/events/abstracts/templates/abstract.html
ThiefMaster: BTW, should we even show the by-question summary in this case? And if yes, do we want to restrict the avg there to tracks you can see reviews for?

PR #2741, opened 2016-11-28
Fix search bar style in conference pages

OmeGak:
Merge alongside with https://github.com/indico/indico-plugins/pull/38.

Examples:

---

screenshot from 2016-11-28 15 51 10

---

screenshot from 2016-11-28 15 51 22


indico/htdocs/css/Conf_Basic.css
```diff
@@ -302,40 +293,6 @@ pre {
     font-size: 11pt;
 }
 
-    /* Styles for the searchbox */
-
-.confSearchBox {
-    width: 250px;
-    float: right;
-    height: 24px;
-    background: transparent url("../images/conf/sprites_blue.png") no-repeat 0 0;
```
ThiefMaster: Is it used anywhere else?  Otherwise let's get rid of it

indico/htdocs/sass/base/_layout.scss
```diff
@@ -136,7 +136,14 @@ body > * {
     }
 }
 
-/* Flexbox layout */
+
+// -- Flexbox layout -----------------------------------------------------------
```
ThiefMaster: Maybe the presence of such a comment is a good indicator that the block below belongs into a separate file?

OmeGak: Not a strong opinion. Thoughts? @pferreir @nop33

indico/htdocs/css/Conf_Basic.css
OmeGak: It's still used in `.simpleTextAnnouncement`. :(

ThiefMaster: Maybe... just maybe... if it doesn't take too much time to clean up? ;)

indico/htdocs/sass/base/_layout.scss
pferreir: No strong opinion.

ThiefMaster: It reminds me a log of the `################ ...` lines in legacy python code

nop33: I'd keep the simple `/* */` I agree it's pretty ugly having an asymmetric line of dashes

nop33: And yes, maybe a different file wouldn't be such a bad idea. OTOH it's the `layout.scss`, so the flexbox classes fit just fine there.

ThiefMaster: FWIW I'd use the SCSS-style `//` comments. It's just one line anyway.

PR #2742, opened 2016-11-29
Update review summary header

nop33:
### Hiding the header information
image

#### Example 1
image
#### Example 2
image

### DIFF
--- a/indico/modules/events/abstracts/templates/abstract/judge.html
+++ b/indico/modules/events/abstracts/templates/abstract/judge.html
@@ -10,7 +10,6 @@
                                     "but as different contribution types") %}
     {% endif %}
     {% set mapping_class = {'positive': 'success',
-                            'conflicting': 'visited',
                             'negative': 'error',
                             'mixed': 'warning'} %}
     {% set mapping_title = {'not_started': _("No reviewer has reviewed this abstract yet"),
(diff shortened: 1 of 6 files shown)


indico/modules/events/abstracts/templates/abstract/judge.html
```diff
@@ -19,11 +18,20 @@
                             'conflicting': conflicting_msg,
                             'negative': _("No reviewer proposed to accept this abstract"),
                             'mixed': _("Not all reviewers proposed to accept this abstract")} %}
+    {% set mapping_icon = {'positive': 'icon-checkmark',
+                           'conflicting': 'icon-checkmark',
+                           'negative': 'icon-cross',
+                           'mixed': 'icon-tilde',
+                           'in_progress': 'icon-circle-small',
+                           'not_started': 'icon-circle-small'} %}
```
nop33: We have many mapping sets in this file (open the file to see the whole content). Any ideas as to how to simplify it?

OmeGak: Not sure if any simplification won't end up just moving the complexity somewhere else. I don't think it's so ugly right now.

indico/modules/events/abstracts/views.py
```diff
@@ -108,12 +108,15 @@ def render_abstract_page(abstract, view_class=None, management=False):
         review_form = build_review_form(abstract, reviewed_for_tracks[0])
     judgment_form = AbstractJudgmentForm(abstract=abstract, formdata=None)
     review_track_list_form = AbstractReviewedForTracksForm(event=abstract.event_new, obj=abstract, formdata=None)
+    convener_tracks = set(get_convener_tracks(abstract.event_new, session.user))
```
OmeGak: We use `get_convener_tracks` here only and we don't really care about the order, so maybe it's worth returning a set directly.

indico/modules/events/abstracts/views.py
```diff
@@ -108,12 +108,15 @@ def render_abstract_page(abstract, view_class=None, management=False):
...
+    visible_tracks = set(t for t in abstract.reviewed_for_tracks
```
OmeGak:
```python
if abstract.can_judge(session.user, check_state=True):
    visible_tracks = abstract.reviewed_for_tracks
else:
    ...
```

OmeGak: > Discussed IRL

PR #2748, opened 2016-12-06
Call for Papers overview page

nop33:


indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -0,0 +1,74 @@
...
+    def open(self):
+        if self.has_ended:
+            paper_reviewing_settings.set(self.event, 'end_dt', None)
+        else:
+            paper_reviewing_settings.set(self.event, 'start_dt', now_utc(False))
+
+    def close(self):
+        now = now_utc(False)
+        paper_reviewing_settings.set(self.event, 'end_dt', now)
+        if not self.has_started:
```
ThiefMaster: does this make sense? I don't think close should ever be called if it's not open. If it's just scheduled we have "start now" and "reschedule" (or "start now" / "schedule" if it's not scheduled).

So I'd rather fail or do nothing if you try to close without being `is_open`

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -0,0 +1,74 @@
...
+            paper_reviewing_settings.set(self.event_new, 'start_dt', now)
+
+    def toggle_reviewing(self, reviewing_type=None, enable=True):
```
ThiefMaster: I don't think any of these two arguments should have default values. Not specifying them doesn't make much sense

nop33: If they shouldn't have default values and by not defining default values it makes no sense, then what are you proposing exactly?!

ThiefMaster: I mean not specifying the values when calling the method makes no sense, which is why I would make both of the required by not specifying default values.

indico/modules/events/papers/blueprint.py
```diff
@@ -32,6 +32,8 @@
 _bp.add_url_rule('/manage/papers/', 'management', management.RHPapersDashboard)
 _bp.add_url_rule('/manage/papers/teams', 'manage_teams', management.RHManagePaperTeams,
                  methods=('GET', 'POST'))
+_bp.add_url_rule('mamange/papers/enable/<reviewing_type>', 'switch', management.RHSwitchReviewingType,
```
ThiefMaster: - `mamange`? 🐒 
- `.../<any(content,layout):reviewing_type>` - then you don't need to check against a list of valid types in the controller anymore

indico/modules/events/papers/controllers/management.py
```diff
@@ -51,3 +54,32 @@ def _process(self):
             logger.info("Paper managers of %r updated by %r", self.event_new, session.user)
             return jsonify_data(flash=False)
         return jsonify_form(form)
+
+
+class RHSwitchReviewingType(RHManagePapersBase):
+
+    CSRF_ENABLED = True
```
ThiefMaster: This is already inherited from `RHPapersBase` which is a parent of `RHManagePapersBase`.

indico/modules/events/papers/controllers/management.py
```diff
@@ -51,3 +54,32 @@ def _process(self):
             logger.info("Paper managers of %r updated by %r", self.event_new, session.user)
             return jsonify_data(flash=False)
         return jsonify_form(form)
+
+
+class RHSwitchReviewingType(RHManagePapersBase):
+
```
ThiefMaster: too much whitespace ;)

indico/modules/events/papers/controllers/management.py
```diff
@@ -51,3 +54,32 @@ def _process(self):
...
+    CSRF_ENABLED = True
+
+    def _process_PUT(self):
+        event = self.event_new
+        reviewing_type = request.view_args['reviewing_type']
+        if reviewing_type not in ('content', 'layout'):
```
ThiefMaster: see comment above

indico/modules/events/papers/controllers/management.py
```diff
@@ -51,3 +54,32 @@ def _process(self):
...
+            raise Forbidden('Reviewing type not available')
+        event.cfp.toggle_reviewing(reviewing_type=reviewing_type, enable=True)
```
ThiefMaster: I don't think we need kwargs style here - it's quite obvious what the arguments do :)

indico/modules/events/papers/controllers/management.py
```diff
@@ -51,3 +54,32 @@ def _process(self):
...
+        flash(_("Reviewing type enabled"), 'success')
```
ThiefMaster: do we really need a flash message for this?

indico/modules/events/papers/controllers/management.py
```diff
@@ -51,3 +54,32 @@ def _process(self):
...
+        logger.info("Reviewing type '{}' for event {} enabled by {}".format(reviewing_type, event, session.user))
```
ThiefMaster: logging always uses `%` and the arguments are passed as a `list, of, args` instead of using `.format()` so the log message passed to the method is static and still contains the actual placeholders.

For event and user  i'd use `%r` instead of `%s` since those are object reprs anyway

indico/modules/events/papers/controllers/management.py
```diff
@@ -51,3 +54,32 @@ def _process(self):
...
             return jsonify_data(flash=False)
         return jsonify_form(form)
+
+
+class RHSwitchReviewingType(RHManagePapersBase):
+
+    CSRF_ENABLED = True
+
+    def _process_PUT(self):
+        event = self.event_new
```
ThiefMaster: I'd just use `self.event_new` - eventually `event_new` will be renamed to `event` and having `event = self.event` is a bit unnecessary ;)

indico/modules/events/papers/controllers/management.py
```diff
@@ -51,3 +54,32 @@ def _process(self):
...
+        event.log(EventLogRealm.management, EventLogKind.positive, 'Papers',
+                  "Enabled reviewing type '{}'".format(reviewing_type), session.user)
+        return jsonify_data(enabled=True)
+
+    def _process_DELETE(self):
+        event = self.event_new
+        reviewing_type = request.view_args['reviewing_type']
+        if reviewing_type not in ('content', 'layout'):
+            raise Forbidden('Reviewing type not available')
+        event.cfp.toggle_reviewing(reviewing_type=reviewing_type, enable=False)
```
ThiefMaster: The code of DELETE and PUT is almost the same.

ThiefMaster: Why not add `def _toggle(self, enable): ...` and do all the stuff in there and then just call it from the PUT/DELETE methods? The only difference is in the log message which could easily be kept in the `_process_*` methods and the `logger` log could even use `toggled (%s)` and just have True/False there. It's not seen my users anyway.

nop33: I think it'd make sense to create a `toggle_review_type` method in `operations.py`, but sure, I agree with you.

indico/modules/events/papers/controllers/management.py
```diff
@@ -51,3 +52,15 @@ def _process(self):
...
             return jsonify_data(flash=False)
         return jsonify_form(form)
+
+
+class RHSwitchReviewingType(RHManagePapersBase):
+    """Enable/disable the paper reviewing types"""
+
+    def _process_PUT(self):
+        toggle_reviewing_type(self.event_new, request.view_args['reviewing_type'], True)
+        return jsonify_data(enabled=True)
```
ThiefMaster: `flash=False` since you don't flash anything and reload (so even if you did flash, you wouldn't want it to be sent in the ajax response)

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -0,0 +1,72 @@
...
+        })
+
+    def open(self):
+        if self.has_ended:
+            paper_reviewing_settings.set(self.event, 'end_dt', None)
+        else:
+            paper_reviewing_settings.set(self.event, 'start_dt', now_utc(False))
+
+    def close(self):
+        now = now_utc(False)
```
ThiefMaster: could be merged into the next line

indico/modules/events/papers/operations.py
```diff
@@ -0,0 +1,30 @@
...
+from flask import session
+
+from indico.modules.events.papers import logger
+from indico.modules.events.logs.models.entries import EventLogRealm, EventLogKind
+
+
+def toggle_reviewing_type(event, reviewing_type, enable):
+    event.cfp.toggle_reviewing(reviewing_type, enable)
+    logger.info("Reviewing type '%s' for event %r toggled (%s) by %r"
+                .format(reviewing_type, event, enable, session.user))
```
ThiefMaster: don't `.format()` here..

indico/modules/events/papers/settings.py
```diff
@@ -0,0 +1,26 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from indico.modules.events.settings import EventSettingsProxy
+
+paper_reviewing_settings = EventSettingsProxy('paper_reviewing', {
+    'start_dt': None,
```
ThiefMaster: missing converters

nop33: @DavidAndreev is adding them. The PR doesn't include the scheduling anyway.

ThiefMaster: I know, but IMO it'd still be cleaner to fix this here, no need to have a commit with known bugs that are fixed in a few lines ;)

indico/htdocs/sass/partials/_actionboxes.scss
```diff
@@ -84,11 +84,19 @@
     background: $background-color;
     color: $color;
 
-    > .section > .icon {
+    > .section {
```
OmeGak: I would take the opportunity to stop restricting action-box elements to be directly nested. (`>`)

indico/htdocs/sass/partials/_actionboxes.scss
```diff
@@ -84,11 +84,19 @@
...
+        @include action-box-section-colors($color, $emphasis-color);
+    }
+}
+
+@mixin action-box-section-colors($color, $emphasis-color) {
```
OmeGak: Just an idea, since this simply changes text color and I don't think we are using `action-box-section-colors` anywhere else. Couldn't we just simply user `.semantic-text whatever` in case we need to color the text differently.

indico/modules/events/papers/operations.py
```diff
@@ -0,0 +1,29 @@
...
+
+from flask import session
+
+from indico.modules.events.papers import logger
+from indico.modules.events.logs.models.entries import EventLogRealm, EventLogKind
+
+
+def toggle_reviewing_type(event, reviewing_type, enable):
+    event.cfp.toggle_reviewing(reviewing_type, enable)
+    logger.info("Reviewing type '%s' for event %r toggled (%s) by %r", reviewing_type, event, enable, session.user)
```
OmeGak: I think it would be cleaner to show `enabled`/`disabled` like you are doing in `event.log`.

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, event):
+        self.event = event
+
+    @return_ascii
+    def __repr__(self):
+        return '<CallForPapers({}, start_dt={}, end_dt={})>'.format(self.event.id, self.start_dt, self.end_dt)
+
+    start_dt = EventSettingProperty(paper_reviewing_settings, 'start_dt')
+    end_dt = EventSettingProperty(paper_reviewing_settings, 'end_dt')
+    content_reviewing_enabled = EventSettingProperty(paper_reviewing_settings, 'content_reviewing_enabled')
```
OmeGak: No `is_..._enabled`?

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -0,0 +1,71 @@
...
+    def open(self):
+        if self.has_ended:
+            paper_reviewing_settings.set(self.event, 'end_dt', None)
+        else:
+            paper_reviewing_settings.set(self.event, 'start_dt', now_utc(False))
+
+    def close(self):
+        paper_reviewing_settings.set(self.event, 'end_dt', now_utc(False))
+
+    def toggle_reviewing(self, reviewing_type, enable):
```
OmeGak: This doesn't look like *toggling* to me. More like **setting**.

indico/modules/events/papers/operations.py
ThiefMaster: `state = 'enabled' if enable else 'disabled'` `...%r %s by...` ;)

indico/htdocs/sass/partials/_actionboxes.scss
nop33: I agree with removing it from the `.section` class, but not from the `.icon` below. Otherwise it would affect any icon element in the `.action-box`.

indico/modules/events/papers/models/call_for_papers.py
nop33: It's toggling it's state from enabled to disabled.

nop33: Hm, `toggle_reviewing_state` would be nicer now that I think of it.

nop33: actually, `toggle_reviewing_type_state` :stuck_out_tongue_closed_eyes:

ThiefMaster: that last one sounds like 💩 ;)

nop33: Oh, but you're right, the method in the `operations.py` should be named like that, but the method of the class I do agree that it's **setting** and not **toggling**. Thanks!

OmeGak: What I mean is that toggling doesn't happen at all since passing the new enabled/disabled state is required.

nop33: @ThiefMaster why? That's essensially what it does. It `toggle`s the `state` of the `reviewing_type` content/layout.

ThiefMaster: it's quite clear from the arguments IMO, no need to be *that* verbose

nop33: This way we avoid a docstring :wink:

ThiefMaster: the arguments are kind of self-documenting ;)

PR #2754, opened 2016-12-08
Add paper reviewing teams and competences configuration

nop33:


indico/modules/events/papers/blueprint.py
```diff
@@ -30,10 +30,12 @@
 
 # Management
 _bp.add_url_rule('/manage/papers/', 'management', management.RHPapersDashboard)
-_bp.add_url_rule('/manage/papers/teams', 'manage_teams', management.RHManagePaperTeams,
+_bp.add_url_rule('/manage/papers/teams/', 'manage_teams', management.RHManagePaperTeams,
                  methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/papers/enable/<any(content,layout):reviewing_type>', 'switch',
                  management.RHSwitchReviewingType, methods=('PUT', 'DELETE'))
+_bp.add_url_rule('/manage/papers/teams/competences', 'manage_competences', management.RHManageCompetences,
```
ThiefMaster: I'd put this right below the `/manage/papers/teams/` rule

indico/modules/events/papers/controllers/management.py
```diff
@@ -51,13 +54,27 @@ class RHManagePaperTeams(RHManagePapersBase):
...
-        managers = {p.principal for p in self.event_new.acl_entries
-                    if p.has_management_role('paper_manager', explicit=True)}
-        form = PaperManagersForm(managers=managers)
+        cfp = self.event_new.cfp
+        form_data = {
+            'managers': cfp.managers,
+            'judges': cfp.judges
+        }
+        if cfp.content_reviewing_enabled:
+            form_data['content_reviewers'] = cfp.content_reviewers
```
ThiefMaster: You can always set these, no need to check whether it's enabled or not.

indico/modules/events/papers/controllers/management.py
```diff
@@ -51,13 +54,27 @@ class RHManagePaperTeams(RHManagePapersBase):
...
             update_object_principals(self.event_new, form.managers.data, role='paper_manager')
-            logger.info("Paper managers of %r updated by %r", self.event_new, session.user)
-            return jsonify_data(flash=False)
+            update_object_principals(self.event_new, form.judges.data, role='paper_judge')
+            if cfp.content_reviewing_enabled:
+                update_object_principals(self.event_new, form.content_reviewers.data, role='paper_content_reviewer')
+            if cfp.layout_reviewing_enabled:
+                update_object_principals(self.event_new, form.layout_reviewers.data, role='paper_layout_reviewer')
+            flash(_("The members of the teams were updated successfully"), 'success')
+            logger.info("Paper teams of %r updated by %r", self.event_new, session.user)
```
ThiefMaster: no event log entry?

indico/modules/events/papers/controllers/management.py
```diff
@@ -71,3 +88,23 @@ def _process_PUT(self):
...
+
+class RHManageCompetences(RHManagePapersBase):
+    """Manage the competences of the call for papers ACLs"""
+
+    def _process(self):
+        form_class = make_competences_form(self.event_new)
+        form = form_class(obj=FormDefaults(get_user_competences(self.event_new)))
+        if form.validate_on_submit():
+            user_competences = self.event_new.cfp.user_competences
+            user_competences_data = {int(key[12:]): value for key, value in form.data.iteritems()}
```
ThiefMaster: I'd add `if key.startswith('competences_')` to make it clear what the `[12:]` is about

indico/modules/events/papers/templates/management/competences.html
```diff
@@ -0,0 +1,34 @@
...
+{% block description %}
+    {%- trans -%}
+        Defining competences to the members of the teams may be helpful for the managers and judges during the paper assignment process.
+    {%- endtrans -%}
+{%- endblock %}
+
+{% block content -%}
+    <div id="competences-dialog-content">
+        {{ form_header(form) }}
+            <table class="i-table-widget">
```
ThiefMaster: too much indentation IMO

indico/modules/events/papers/templates/management/competences.html
```diff
@@ -0,0 +1,34 @@
...
+                        <tr>
+                            <td class="icon-user competences-user">{{ entry.full_name }}</td>
+                            <td>{{ form['competences_{}'.format(entry.id)] }}</td>
+                        </tr>
+                    {% endfor %}
+                </tbody>
+            </table>
+        {% call form_footer(form) %}
+            <input class="i-button big highlight" type="submit" value="{% trans %}Save{% endtrans %}">
+            <a href="{{ back_url }}" class="i-button big" data-button-back>{% trans %}Cancel{% endtrans %}</a>
```
ThiefMaster: you don't have a `back_url` - check the `simple_form` macro, i'm quite sure we omit it and/or use a `<button>` in case of no `back_url`

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -69,3 +71,41 @@ def set_reviewing_state(self, reviewing_type, enable):
             paper_reviewing_settings.set(self.event, 'content_reviewing_enabled', enable)
         elif reviewing_type == 'layout':
             paper_reviewing_settings.set(self.event, 'layout_reviewing_enabled', enable)
+
+    @property
+    def managers(self):
+        return {p.principal for p in self.event.acl_entries
+                if p.has_management_role('paper_manager', explicit=True)}
```
ThiefMaster: Do we really want `explicit=True` here? Event managers are usually equivalent with explicit managers of a module.

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -69,3 +71,41 @@ def set_reviewing_state(self, reviewing_type, enable):
...
+            entries = entries | self.layout_reviewers
+        return entries
+
+    @property
+    @memoize_request
+    def user_competences(self):
+        user_ids = {user.id for user in self.event.cfp.acl_entries}
+        user_competences = (PaperCompetence.query
+                            .with_parent(self.event)
+                            .filter(PaperCompetence.user_id.in_(user_ids)).all())
```
ThiefMaster: I'd put `.all()` into a new line

Also, usually I keep the `.with_parent(..)` in the `Foo.query` line but no strong opinion, I just see it more as something that's giving you an initial query while everything else after that modifies/executes it

indico/modules/events/papers/models/call_for_papers.py
ThiefMaster: Also, do competences for managers really make sense (since you use `acl_entries` there)?

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -69,3 +71,41 @@ def set_reviewing_state(self, reviewing_type, enable):
...
+    @property
+    def layout_reviewers(self):
+        return {p.principal for p in self.event.acl_entries
+                if p.has_management_role('paper_layout_reviewer', explicit=True)}
+
+    @property
+    def acl_entries(self):
+        entries = self.managers | self.judges
+        if self.content_reviewing_enabled:
+            entries = entries | self.content_reviewers
```
ThiefMaster: You can use `|=` here (the initial `entries` is a new set so it won't modify stuff in the DB)

indico/modules/events/papers/operations.py
```diff
@@ -28,3 +31,22 @@ def set_reviewing_state(event, reviewing_type, enable):
     logger.info("Reviewing type '%s' for event %r %s by %r", reviewing_type, event, action, session.user)
     event.log(EventLogRealm.management, EventLogKind.positive, 'Papers',
               "{} reviewing type '{}'".format("Enabled" if enable else "Disabled", reviewing_type), session.user)
+
+
+def create_competences(event, user_id, competences):
+    user_competences = PaperCompetence(user_id=user_id, event_id=event.id, competences=competences)
+    db.session.add(user_competences)
+    user = User.get_one(user_id)
```
ThiefMaster: I think this should be outside, and the function should take a `user`. That way you can also pass the user to the `PaperCompetence` constructor. Passing IDs around for relationships is a bad idea in general.

indico/modules/events/papers/operations.py
```diff
@@ -28,3 +31,22 @@ def set_reviewing_state(event, reviewing_type, enable):
     logger.info("Reviewing type '%s' for event %r %s by %r", reviewing_type, event, action, session.user)
     event.log(EventLogRealm.management, EventLogKind.positive, 'Papers',
               "{} reviewing type '{}'".format("Enabled" if enable else "Disabled", reviewing_type), session.user)
+
+
+def create_competences(event, user_id, competences):
+    user_competences = PaperCompetence(user_id=user_id, event_id=event.id, competences=competences)
+    db.session.add(user_competences)
```
ThiefMaster: not needed if you pass objects instead of ids

indico/modules/events/papers/controllers/management.py
nop33: well, what do we lose by checking?

indico/modules/events/papers/controllers/management.py
nop33: Well, it's kinda unnecessary imo cause we already have the protection log entry:

image

indico/modules/events/papers/templates/management/competences.html
nop33: Well, `{{ form_header(form) }}` will technically open an HTML tag, which will close by calling the `form_footer`.

indico/modules/events/papers/models/call_for_papers.py
nop33: Hm, it's true, doesn't make sense since you cannot assign paper managers to papers... However, we do have it available in legacy. I guess this is one of the features we don't really want to migrate.

indico/modules/events/papers/controllers/management.py
ThiefMaster: Just saying that it's code that's not actually needed :)

indico/modules/events/papers/templates/management/competences.html
ThiefMaster: Matter of opinion I guess. @pferreir also likes to do it, to me it feels weird since even though it does open/close a tag it's not actually that tag and if you made an editor/IDE/.. re-indent the file this would be undone (I don't know any tool that *can* re-indent Jinja templates though)

indico/htdocs/js/indico/jquery/ajaxdialog.js
```diff
@@ -184,6 +184,10 @@
                 _.defer(function() {
                     popup.canvas.data('ui-dialog')._focusTabbable();
                 });
+
+                // Re-apply default position after content have been loaded
```
ThiefMaster: *has* been

indico/modules/events/papers/controllers/management.py
```diff
@@ -71,3 +89,26 @@ def _process_PUT(self):
...
+        defaults = {'competences_{}'.format(user_id): competences.competences
+                    for user_id, competences in user_competences.iteritems()}
+        form = form_class(obj=FormDefaults(defaults))
+        if form.validate_on_submit():
+            form_data = {int(key[12:]): value for key, value in form.data.iteritems() if key.startswith('competences_')}
+            for user_id, competences in form_data.iteritems():
+                if user_id in user_competences:
+                    update_competences(user_competences[user_id], competences)
+                elif competences:
+                    user = User.get_one(user_id)
```
ThiefMaster: `get_one(user_id, is_deleted=False)`

indico/modules/events/papers/operations.py
```diff
@@ -28,3 +30,30 @@ def set_reviewing_state(event, reviewing_type, enable):
...
+    logger.info("Competences for user %r for event %r created by %r", user, event, session.user)
+    event.log(EventLogRealm.management, EventLogKind.positive, 'Papers',
+              "Added competences of {}".format(user.full_name), session.user,
+              data={'Competences': ', '.join(competences)})
+
+
+def update_competences(user_competences, competences):
+    event = user_competences.event_new
+    user_competences.competences = competences
+    logger.info("Competences for user %r for event %r updated by %r", user_competences.user, event, session.user)
```
ThiefMaster: *in* event?

indico/modules/events/papers/templates/management/competences.html
```diff
@@ -0,0 +1,34 @@
+{% from 'forms/_form.html' import form_header, form_field, form_footer %}
+{% extends 'layout/conference_page_base.html' %}
+
+{% block description %}
+    {%- trans -%}
+        Defining competences to the members of the teams may be helpful for the managers and judges during the paper assignment process.
```
ThiefMaster: very long line

indico/htdocs/js/indico/jquery/ajaxdialog.js
pferreir: BTW, there's a typo in the commit message.

indico/htdocs/sass/partials/_jquery-indico-taglist.scss
```diff
@@ -20,16 +20,16 @@
     @include border-all();
 
     height: auto;
-    min-height: 50px;
     cursor: text;
-    padding: 5px;
+    padding: 0.3rem;
```
OmeGak: You convince me, but maybe we should settle on a strategy for this before changing `px` to `rem` everywhere.

indico/modules/events/papers/controllers/management.py
```diff
@@ -71,3 +89,26 @@ def _process_PUT(self):
...
+    """Manage the competences of the call for papers ACLs"""
+
+    def _process(self):
+        form_class = make_competences_form(self.event_new)
+        user_competences = self.event_new.cfp.user_competences
+        defaults = {'competences_{}'.format(user_id): competences.competences
+                    for user_id, competences in user_competences.iteritems()}
+        form = form_class(obj=FormDefaults(defaults))
+        if form.validate_on_submit():
+            form_data = {int(key[12:]): value for key, value in form.data.iteritems() if key.startswith('competences_')}
```
OmeGak: No strong opinion, but here are some alternatives for `key[12:]`.

```python
key.strip('competences_')
key.replace('competences_', '')
```

indico/modules/events/papers/controllers/management.py
OmeGak: What happens here if the user got deleted in the meantime? Wouldn't it be better to check for `if user is not None`?

indico/modules/events/papers/operations.py
```diff
@@ -28,3 +30,30 @@ def set_reviewing_state(event, reviewing_type, enable):
...
+
+
+def update_team_members(event, managers, judges, content_reviewers, layout_reviewers):
+    update_object_principals(event, managers, role='paper_manager')
+    update_object_principals(event, judges, role='paper_judge')
+    if content_reviewers:
+        update_object_principals(event, content_reviewers, role='paper_content_reviewer')
+    if layout_reviewers:
+        update_object_principals(event, layout_reviewers, role='paper_layout_reviewer')
+    logger.info("Paper teams of %r updated by %r", event, session.user)
```
OmeGak: No `event.log`?

indico/modules/events/papers/templates/management/competences.html
```diff
@@ -0,0 +1,34 @@
+{% from 'forms/_form.html' import form_header, form_field, form_footer %}
+{% extends 'layout/conference_page_base.html' %}
```
OmeGak: We usually extend before of importing.

indico/modules/events/papers/templates/management/competences.html
OmeGak: I'm with @ThiefMaster here. Another point is that this requires understanding of what `form_header` internally does and will most likely confuse anybody reading the code (*"is this indentation intended?"*).

indico/web/templates/layout/base.html
```diff
@@ -1,36 +1,38 @@
 {% block page %}
     <div class="{% block page_class %}management-page{% endblock %}"
          {% if self.page_id()|trim %}id="{% filter trim %}{% block page_id %}{% endblock %}{% endfilter %}"{% endif %}>
-        {% if self.title() %}
+        {% if self.title() or self.description() or self.info() %}
```
OmeGak: Do we ever want a header without `title`?

indico/modules/events/papers/controllers/management.py
ThiefMaster: `strip` does something completely different (it takes a list of chars to strip, not a single string to strip):

```
>>> 'abcdef'.strip('foo')
'abcde'
```

Not a fan of `replace` there since it matches everywhere and a regex would be overkill IMO.

indico/modules/events/papers/operations.py
ThiefMaster: The ACL changes are already logged (I had commented the same thing ;))

indico/web/templates/layout/base.html
ThiefMaster: Probably not, but if it works fine I don't see any drawbacks in allowing it

indico/modules/events/papers/templates/management/competences.html
nop33: Had this argument with @ThiefMaster already..!

indico/web/templates/layout/base.html
nop33: I wanted it so I can have this layout (page description):
image

OmeGak: Should we really use that base layout within a dialog? This solution looks a bit like stretching `layout/base.html` too much, since it's clearly intended for pages. We could have another base template `layout/dialog.html` for dialogs and then make sure via CSS that the style of both match.

indico/modules/events/papers/templates/management/competences.html
OmeGak: And the outcome was...?

indico/modules/events/papers/controllers/management.py
OmeGak: Another way I stumbled upon yesterday and I think it's cleaner than counting yourself how many characters the string has:

```python
key[len('competences_'):]
```

indico/htdocs/js/indico/jquery/ajaxdialog.js
```diff
@@ -184,6 +184,10 @@
                 _.defer(function() {
                     popup.canvas.data('ui-dialog')._focusTabbable();
                 });
+
+                // Re-apply default position after content has been loaded
+                var initialPosition = popup.canvas.dialog('option', 'position');
+                popup.canvas.dialog('option', 'position', initialPosition);
```
ThiefMaster: It's enough to return `true` from this function: https://github.com/indico/indico/blob/master/indico/htdocs/js/indico/Core/Dialogs/Popup.js#L164-L168

indico/modules/events/papers/controllers/management.py
```diff
@@ -84,3 +87,27 @@ def _process_PUT(self):
...
+
+    def _process(self):
+        form_class = make_competences_form(self.event_new)
+        user_competences = self.event_new.cfp.user_competences
+        defaults = {'competences_{}'.format(user_id): competences.competences
+                    for user_id, competences in user_competences.iteritems()}
+        form = form_class(obj=FormDefaults(defaults))
+        if form.validate_on_submit():
+            key_prefix = 'competences_'
+            form_data = {int(key[len(key_prefix):]): value for key, value in form.data.iteritems()}
```
ThiefMaster: I'd still keep the `if key.startswith(key_prefix)` here.

indico/modules/events/papers/controllers/management.py
```diff
@@ -84,3 +87,27 @@ def _process_PUT(self):
...
+            for user_id, competences in form_data.iteritems():
+                if user_id in user_competences:
+                    update_competences(user_competences[user_id], competences)
+                elif competences:
+                    user = User.get_one(user_id, is_deleted=False)
```
ThiefMaster: It doesn't make a huge difference here since it only runs on save and you probably won't have *that* many reviewers, but you could query all of them at once:

    users = {u.id: u for u in User.query.filter(User.id.in_(form_data), ~User.is_deleted)}

indico/modules/events/papers/forms.py
```diff
@@ -40,3 +49,7 @@ def __init__(self, *args, **kwargs):
             del self.content_reviewers
         if not self.event.cfp.layout_reviewing_enabled:
             del self.layout_reviewers
+
+
+class PaperCompetencesForm(IndicoForm):
```
ThiefMaster: Just use `IndicoForm` directly in the `type()` call.

    form_class = type(b'PaperCompetencesForm', (IndicoForm,), {})

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -89,10 +91,19 @@ def layout_reviewers(self):
                 if p.has_management_role('paper_layout_reviewer', explicit=True)}
 
     @property
-    def acl_entries(self):
-        principals = self.managers | self.judges
+    def assignees(self):
+        users = self.judges
```
ThiefMaster: I'd use `set(self.judges)` just to be clear that we don't modify it. Since `judges` returns a new set every time it won't be an issue but I had to look twice to check whether we end up modifying something we don't want to modify below with the in-place changes

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -89,10 +91,19 @@ def layout_reviewers(self):
...
+        return users
+
+    @property
+    @memoize_request
+    def user_competences(self):
+        user_ids = {user.id for user in self.event.cfp.assignees}
+        user_competences = (PaperCompetence.query.with_parent(self.event)
+                            .filter(PaperCompetence.user_id.in_(user_ids))
+                            .all())
+        return {competences.user_id: competences for competences in user_competences}
```
ThiefMaster: `competence` instead of not `competences`? (better fits the model name). Or just a generic loop var like `x` or `c` ;)

indico/web/templates/layout/dialog_base.html
```diff
@@ -0,0 +1,24 @@
+{% block page %}
+    <div class="{% block dialog_class %}dialog-page{% endblock %}"
+        {% if self.dialog_id()|trim -%}
```
ThiefMaster: one space indentation missing?

indico/web/templates/layout/base.html
```diff
@@ -1,6 +1,6 @@
 {% block page %}
     <div class="{% block page_class %}management-page{% endblock %}"
-         {% if self.page_id()|trim %}id="{% filter trim %}{% block page_id %}{% endblock %}{% endfilter %}"{% endif %}>
+        {% if self.page_id()|trim %}id="{% filter trim %}{% block page_id %}{% endblock %}{% endfilter %}"{% endif %}>
```
ThiefMaster: I'd keep it aligned with the attributes... It's semi-ugly either way but more consistent IMO

indico/htdocs/js/indico/jquery/ajaxdialog.js
nop33: I think I've tried that but it didn't work. Lemme try again

indico/modules/events/papers/controllers/management.py
nop33: I liked @OmeGak suggestion better than the above.

ThiefMaster: I like the `len()` thing too, but right now this code would explode in your face if you ever had any extra field(s) in the form that don't have a `competences_*` name

indico/web/templates/layout/dialog_base.html
nop33: I would say that
``` jinja
<div class="{% block dialog_class %}dialog-page{% endblock %}"
     {% if self.dialog_id()|trim -%}
         id="{% filter trim %}{% block dialog_id %}{% endblock %}{% endfilter %}"
     {%- endif -%}
>
    {% if self.description() or self.info() %}
        <header>
``` 
looks better than
``` jinja
<div class="{% block dialog_class %}dialog-page{% endblock %}"
     {% if self.dialog_id()|trim -%}
         id="{% filter trim %}{% block dialog_id %}{% endblock %}{% endfilter %}"
     {%- endif %}>
    {% if self.description() or self.info() %}
        <header>
```
because in the first case you can clearly see that the `if` belongs to the data attributes (hence the weird indentation) whereas in the second example it feels like a indentation bug since it differs only by 1 space from the next `if` statement.

Opinions @OmeGak @pferreir ?

pferreir: I think both solutions look awful, so I have no personal preference. Anyway, I think this is the kind of thing that can be left to personal preference 😉

PR #2756, opened 2016-12-09
Paper assignment page

mvidalgarcia:
WIP:
- [x] Filter by state
- [x] Add optional columns: Contribution type, session
- [x] Filter by Track, Contribution type, session
- [ ] ~Extra filters: Material submitted~
- [x] Bulk change state

### DIFF
--- a/indico/htdocs/js/indico/modules/papers.js
+++ b/indico/htdocs/js/indico/modules/papers.js
@@ -15,7 +15,26 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
-(function() {
+/* global setupListGenerator:false, applySearchFilters:false, setupTableSorter:false */
+
+(function(global) {
     'use strict';
 
-})();
+    global.setupPaperAssignmentList = function setupPaperAssignmentList() {
+        var filterConfig = {
+            itemHandle: 'tr',
+            listItems: '#assignment-list tbody tr',
+            term: '#search-input',
+            state: '#filtering-state',
+            placeholder: '#filter-placeholder'
+        };
+
+        setupTableSorter('#assignment-list .tablesorter');
+        enableIfChecked('#assignment-list', 'input[name=contribution_id]', '.js-enable-if-checked');
+        $('#assignment-list').on('indico:htmlUpdated', function() {
+            setupTableSorter('#assignment-list .tablesorter');
+            applySearchFilters();
+        });
+        setupListGenerator(filterConfig);
+    };
+})(window);
--- a/indico/modules/events/contributions/lists.py
+++ b/indico/modules/events/contributions/lists.py
@@ -40,9 +40,9 @@ class ContributionListGenerator(ListGeneratorBase):
         super(ContributionListGenerator, self).__init__(event)
         self.default_list_config = {'filters': {'items': {}}}
 
-        session_empty = {None: 'No session'}
-        track_empty = {None: 'No track'}
-        type_empty = {None: 'No type'}
+        session_empty = {None: _('No session')}
+        track_empty = {None: _('No track')}
+        type_empty = {None: _('No type')}
         session_choices = OrderedDict((unicode(s.id), s.title) for s in sorted(self.event.sessions,
                                                                                key=attrgetter('title')))
         track_choices = OrderedDict((unicode(t.id), t.title) for t in sorted(self.event.tracks,
@@ -122,8 +122,8 @@ class ContributionListGenerator(ListGeneratorBase):
         :param contrib: Used in RHs responsible for CRUD operations on a
                         contribution.
         :return: dict containing the list's entries, the fragment of
-                 displayed entries and whether the contrib passed is displayed
-                 in the results.
+                 displayed entries and whether the contribution passed is
+                 displayed in the results.
         """
         contrib_list_kwargs = self.get_list_kwargs()
         total_entries = contrib_list_kwargs.pop('total_entries')
@@ -151,7 +151,7 @@ class ContributionDisplayListGenerator(ContributionListGenerator):
         """Render the contribution list template components.
 
         :return: dict containing the list's entries, the fragment of
-                 displayed entries and whether the contrib passed is displayed
+                 displayed entries and whether the contribution passed is displayed
                  in the results.
         """
         contrib_list_kwargs = self.get_list_kwargs()
--- a/indico/modules/events/contributions/models/contributions.py
+++ b/indico/modules/events/contributions/models/contributions.py
@@ -415,6 +415,10 @@ class Contribution(DescriptionMixin, ProtectionManagersMixin, LocationMixin, Att
     def locator(self):
         return dict(self.event_new.locator, contrib_id=self.id)
 
+    @property
+    def verbose_title(self):
+        return '#{} ({})'.format(self.friendly_id, self.title)
+
     @return_ascii
     def __repr__(self):
         return format_repr(self, 'id', is_deleted=False, _text=self.title)
--- a/indico/modules/events/contributions/models/contributions.py
+++ b/indico/modules/events/contributions/models/contributions.py
@@ -34,7 +34,7 @@ from indico.core.db.sqlalchemy.util.queries import increment_and_get
 from indico.core.db.sqlalchemy.util.session import no_autoflush
 from indico.modules.events.management.util import get_non_inheriting_objects
 from indico.modules.events.models.persons import PersonLinkDataMixin, AuthorsSpeakersMixin
-from indico.modules.events.papers.models.revisions import PaperRevisionState
+from indico.modules.events.papers.models.revisions import PaperRevisionState, PaperRevision
 from indico.modules.events.sessions.util import session_coordinator_priv_enabled
 from indico.util.locators import locator_property
 from indico.util.string import format_repr, return_ascii
@@ -331,6 +331,21 @@ class Contribution(DescriptionMixin, ProtectionManagersMixin, LocationMixin, Att
             lazy=True
         )
     )
+
+    @declared_attr
+    def paper_last_revision(cls):
+        # Incompatible with joinedload
+        return db.relationship(
+            'PaperRevision',
+            uselist=False,
+            lazy=True,
+            viewonly=True,
+            primaryjoin=((PaperRevision.contribution_id == cls.id) &
+                         (PaperRevision.submitted_dt == (db.select([db.func.max(PaperRevision.submitted_dt)])
+                                                         .where(PaperRevision.contribution_id == cls.id)
+                                                         .correlate_except(PaperRevision)
+                                                         .as_scalar())))
+        )
 
     # relationship backrefs:
     # - attachment_folders (AttachmentFolder.contribution)
--- a/indico/modules/events/contributions/models/contributions.py
+++ b/indico/modules/events/contributions/models/contributions.py
@@ -335,16 +335,16 @@ class Contribution(DescriptionMixin, ProtectionManagersMixin, LocationMixin, Att
     @declared_attr
     def paper_last_revision(cls):
         # Incompatible with joinedload
+        subquery = (db.select([db.func.max(PaperRevision.submitted_dt)])
+                    .where(PaperRevision.contribution_id == cls.id)
+                    .correlate_except(PaperRevision)
+                    .as_scalar())
         return db.relationship(
             'PaperRevision',
             uselist=False,
             lazy=True,
             viewonly=True,
-            primaryjoin=((PaperRevision.contribution_id == cls.id) &
-                         (PaperRevision.submitted_dt == (db.select([db.func.max(PaperRevision.submitted_dt)])
-                                                         .where(PaperRevision.contribution_id == cls.id)
-                                                         .correlate_except(PaperRevision)
-                                                         .as_scalar())))
+            primaryjoin=db.and_(PaperRevision.contribution_id == cls.id, PaperRevision.submitted_dt == subquery)
         )
 
     # relationship backrefs:
--- a/indico/modules/events/papers/blueprint.py
+++ b/indico/modules/events/papers/blueprint.py
@@ -36,7 +36,9 @@ _bp.add_url_rule('/manage/papers/teams/competences', 'manage_competences', manag
                  methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/papers/enable/<any(content,layout):reviewing_type>', 'switch',
                  management.RHSwitchReviewingType, methods=('PUT', 'DELETE'))
-_bp.add_url_rule('/manage/papers/assign', 'assignment', management.RHPapersAssignment)
+_bp.add_url_rule('/manage/papers/assignment-list', 'assignment', management.RHPapersAssignmentList)
+_bp.add_url_rule('/manage/papers/assign/customize', 'customize_assignment_list', management.RHAssignmentListCustomize,
+                 methods=('GET', 'POST'))
 
 # CFP scheduling
 _bp.add_url_rule('/manage/papers/schedule', 'schedule_cfp', management.RHScheduleCFP,
--- a/indico/modules/events/papers/blueprint.py
+++ b/indico/modules/events/papers/blueprint.py
@@ -36,6 +36,7 @@ _bp.add_url_rule('/manage/papers/teams/competences', 'manage_competences', manag
                  methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/papers/enable/<any(content,layout):reviewing_type>', 'switch',
                  management.RHSwitchReviewingType, methods=('PUT', 'DELETE'))
+_bp.add_url_rule('/manage/papers/assign', 'assignment', management.RHPapersAssignment)
 
 # CFP scheduling
 _bp.add_url_rule('/manage/papers/schedule', 'schedule_cfp', management.RHScheduleCFP,
--- a/indico/modules/events/papers/blueprint.py
+++ b/indico/modules/events/papers/blueprint.py
@@ -39,6 +39,8 @@ _bp.add_url_rule('/manage/papers/enable/<any(content,layout):reviewing_type>', '
 _bp.add_url_rule('/manage/papers/assignment-list/', 'assignment', management.RHPapersAssignmentList)
 _bp.add_url_rule('/manage/papers/assignment-list/customize', 'customize_assignment_list',
                  management.RHAssignmentListCustomize, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/papers/assignment-list/judge', 'manage_judge_papers',
+                 management.RHBulkPaperJudgment, methods=('POST',))
 
 # CFP scheduling
 _bp.add_url_rule('/manage/papers/schedule', 'schedule_cfp', management.RHScheduleCFP,
--- a/indico/modules/events/papers/blueprint.py
+++ b/indico/modules/events/papers/blueprint.py
@@ -36,9 +36,9 @@ _bp.add_url_rule('/manage/papers/teams/competences', 'manage_competences', manag
                  methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/papers/enable/<any(content,layout):reviewing_type>', 'switch',
                  management.RHSwitchReviewingType, methods=('PUT', 'DELETE'))
-_bp.add_url_rule('/manage/papers/assignment-list', 'assignment', management.RHPapersAssignmentList)
-_bp.add_url_rule('/manage/papers/assign/customize', 'customize_assignment_list', management.RHAssignmentListCustomize,
-                 methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/papers/assignment-list/', 'assignment', management.RHPapersAssignmentList)
+_bp.add_url_rule('/manage/papers/assignment-list/customize', 'customize_assignment_list',
+                 management.RHAssignmentListCustomize, methods=('GET', 'POST'))
 
 # CFP scheduling
 _bp.add_url_rule('/manage/papers/schedule', 'schedule_cfp', management.RHScheduleCFP,
--- a/indico/modules/events/papers/controllers/management.py
+++ b/indico/modules/events/papers/controllers/management.py
@@ -20,6 +20,7 @@ from flask import request, render_template, flash
 
 from indico.modules.events.papers.controllers.base import RHManagePapersBase
 from indico.modules.events.papers.forms import PaperTeamsForm, make_competences_form, PapersScheduleForm
+from indico.modules.events.papers.lists import PaperAssignmentListGenerator
 from indico.modules.events.papers.operations import (set_reviewing_state, update_team_members, create_competences,
                                                      update_competences)
 from indico.modules.events.papers.settings import paper_reviewing_settings
@@ -148,10 +149,29 @@ class RHCloseCFP(RHManagePapersBase):
         return jsonify_data(html=_render_paper_dashboard(self.event_new))
 
 
-class RHPapersAssignment(RHManagePapersBase):
+class RHPapersAssignmentList(RHManagePapersBase):
     """Assign contributions to reviewers and judges"""
 
     def _process(self):
-        contribs = sorted(self.event_new.contributions, key=lambda x: x.friendly_id)
+        self.list_generator = PaperAssignmentListGenerator(event=self.event_new)
         return WPManagePapers.render_template('management/assignment.html', self._conf, event=self.event_new,
-                                              contribs=contribs)
+                                              **self.list_generator.get_list_kwargs())
+
+
+class RHAssignmentListCustomize(RHManagePapersBase):
+    """Filter options and columns to display for the paper assignment list of an event"""
+
+    def _checkParams(self, params):
+        RHManagePapersBase._checkParams(self, params)
+        self.list_generator = PaperAssignmentListGenerator(event=self.event_new)
+
+    def _process_GET(self):
+        list_config = self.list_generator.list_config
+        return WPManagePapers.render_template('management/assignment_list_filter.html', self._conf,
+                                              event=self.event_new,
+                                              static_items=self.list_generator.static_items,
+                                              filters=list_config['filters'])
+
+    def _process_POST(self):
+        self.list_generator.store_configuration()
+        return jsonify_data(flash=False, **self.list_generator.render_list())
--- a/indico/modules/events/papers/controllers/management.py
+++ b/indico/modules/events/papers/controllers/management.py
@@ -16,18 +16,21 @@
 
 from __future__ import unicode_literals
 
-from flask import request, render_template, flash
+from flask import request, render_template, flash, session
 
+from indico.modules.events.contributions import Contribution
 from indico.modules.events.papers.controllers.base import RHManagePapersBase
-from indico.modules.events.papers.forms import PaperTeamsForm, make_competences_form, PapersScheduleForm
+from indico.modules.events.papers.forms import (BulkPaperJudgmentForm, make_competences_form, PapersScheduleForm,
+                                                PaperTeamsForm)
 from indico.modules.events.papers.lists import PaperAssignmentListGenerator
+from indico.modules.events.papers.models.revisions import PaperRevisionState
 from indico.modules.events.papers.operations import (set_reviewing_state, update_team_members, create_competences,
-                                                     update_competences)
+                                                     update_competences, judge_paper)
 from indico.modules.events.papers.settings import paper_reviewing_settings
 from indico.modules.events.papers.views import WPManagePapers
 from indico.modules.events.papers.operations import schedule_cfp, open_cfp, close_cfp
 from indico.modules.users.models.users import User
-from indico.util.i18n import _
+from indico.util.i18n import _, ngettext
 from indico.web.forms.base import FormDefaults
 from indico.web.util import jsonify_data, jsonify_form
 
@@ -176,3 +179,50 @@ class RHAssignmentListCustomize(RHManagePapersBase):
     def _process_POST(self):
         self.list_generator.store_configuration()
         return jsonify_data(flash=False, **self.list_generator.render_list())
+
+
+class RHManagePapersActionsBase(RHManagePapersBase):
+    """Base class for RHs performing actions on selected contributions"""
+
+    _contrib_query_options = ()
+
+    @property
+    def _contrib_query(self):
+        query = Contribution.query.with_parent(self.event_new)
+        if self._contrib_query_options:
+            query = query.options(*self._contrib_query_options)
+        return query
+
+    def _checkParams(self, params):
+        RHManagePapersBase._checkParams(self, params)
+        self.list_generator = PaperAssignmentListGenerator(event=self.event_new)
+        ids = map(int, request.form.getlist('contribution_id'))
+        self.contributions = self._contrib_query.filter(Contribution.id.in_(ids)).all()
+
+
+class RHBulkPaperJudgment(RHManagePapersActionsBase):
+    """Perform bulk change state operations on selected papers"""
+
+    def _process(self):
+        form = BulkPaperJudgmentForm(event=self.event_new, judgment=request.form.get('judgment'),
+                                     contribution_id=[c.id for c in self.contributions])
+        if form.validate_on_submit():
+            judgment_data, contrib_data = form.split_data
+            submitted_papers = [c for c in self.contributions if
+                                c.paper_last_revision and c.paper_last_revision.state == PaperRevisionState.submitted]
+            for submitted_paper in submitted_papers:
+                judge_paper(submitted_paper, contrib_data, judge=session.user, **judgment_data)
+            num_submitted_papers = len(submitted_papers)
+            num_not_submitted_papers = len(self.contributions) - num_submitted_papers
+            if num_submitted_papers:
+                flash(ngettext("One paper has been judged.",
+                               "{num} paper have been judged.",
+                               num_submitted_papers).format(num=num_submitted_papers), 'success')
+            if num_not_submitted_papers:
+                flash(ngettext("One contribution has been skipped since it has no paper submitted yet or is in "
+                               "a final state.",
+                               "{num} contributions have been skipped since they have no paper submitted yet or are in "
+                               "a final state.",
+                               num_not_submitted_papers).format(num=num_not_submitted_papers), 'warning')
+            return jsonify_data(**self.list_generator.render_list())
+        return jsonify_form(form=form, submit=_('Judge'), disabled_until_change=False)
--- a/indico/modules/events/papers/controllers/management.py
+++ b/indico/modules/events/papers/controllers/management.py
@@ -170,7 +170,8 @@ class RHAssignmentListCustomize(RHManagePapersBase):
         return WPManagePapers.render_template('management/assignment_list_filter.html', self._conf,
                                               event=self.event_new,
                                               static_items=self.list_generator.static_items,
-                                              filters=list_config['filters'])
+                                              filters=list_config['filters'],
+                                              visible_items=list_config['items'])
 
     def _process_POST(self):
         self.list_generator.store_configuration()
--- a/indico/modules/events/papers/controllers/management.py
+++ b/indico/modules/events/papers/controllers/management.py
@@ -146,3 +146,12 @@ class RHCloseCFP(RHManagePapersBase):
         close_cfp(self.event_new)
         flash(_("Call for papers is now closed"), 'success')
         return jsonify_data(html=_render_paper_dashboard(self.event_new))
+
+
+class RHPapersAssignment(RHManagePapersBase):
+    """Assign contributions to reviewers and judges"""
+
+    def _process(self):
+        contribs = sorted(self.event_new.contributions, key=lambda x: x.friendly_id)
+        return WPManagePapers.render_template('management/assignment.html', self._conf, event=self.event_new,
+                                              contribs=contribs)
--- a/indico/modules/events/papers/forms.py
+++ b/indico/modules/events/papers/forms.py
@@ -18,11 +18,15 @@ from __future__ import unicode_literals
 
 from datetime import time
 
+from flask import request
+from wtforms.fields import TextAreaField, BooleanField, HiddenField
 from wtforms.validators import Optional
 
+from indico.modules.events.papers.models.reviews import PaperAction
 from indico.util.i18n import _
 from indico.web.forms.base import IndicoForm
-from indico.web.forms.fields import PrincipalListField, IndicoDateTimeField, IndicoTagListField
+from indico.web.forms.fields import (PrincipalListField, IndicoDateTimeField, IndicoTagListField, HiddenEnumField,
+                                     HiddenFieldList)
 from indico.web.forms.validators import LinkedDateTime
 
 
@@ -65,3 +69,27 @@ class PapersScheduleForm(IndicoForm):
     def __init__(self, *args, **kwargs):
         self.event = kwargs.pop('event')
         super(PapersScheduleForm, self).__init__(*args, **kwargs)
+
+
+class BulkPaperJudgmentForm(IndicoForm):
+    judgment = HiddenEnumField(enum=PaperAction)
+    contribution_id = HiddenFieldList()
+    submitted = HiddenField()
+    judgment_comment = TextAreaField(_("Comment"), render_kw={'placeholder': _("Leave a comment for the submitter...")})
+    send_notifications = BooleanField(_("Send notifications to submitter"), default=True)
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event')
+        super(BulkPaperJudgmentForm, self).__init__(*args, **kwargs)
+
+    def is_submitted(self):
+        return super(BulkPaperJudgmentForm, self).is_submitted() and 'submitted' in request.form
+
+    @property
+    def split_data(self):
+        contrib_data = self.data
+        judgment_data = {
+            'judgment': contrib_data.pop('judgment'),
+            'send_notifications': contrib_data.pop('send_notifications'),
+        }
+        return judgment_data, contrib_data
--- a/indico/modules/events/papers/lists.py
+++ b/indico/modules/events/papers/lists.py
@@ -1,0 +1,96 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2017 European Organization for Nuclear Research (CERN).
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 3 of the
+# License, or (at your option) any later version.
+#
+# Indico is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from collections import OrderedDict
+
+from flask import request
+from sqlalchemy.orm import subqueryload
+
+from indico.core.db import db
+from indico.modules.events.contributions import Contribution
+from indico.modules.events.papers.models.revisions import PaperRevisionState, PaperRevision
+from indico.modules.events.util import ListGeneratorBase
+from indico.web.flask.templating import get_template_module
+from indico.util.i18n import _
+
+
+class PaperAssignmentListGenerator(ListGeneratorBase):
+    """Listing and filtering actions in a paper assignment list."""
+
+    endpoint = '.assignment'
+    list_link_type = 'paper_asssignment_management'
+
+    def __init__(self, event):
+        super(PaperAssignmentListGenerator, self).__init__(event)
+        self.default_list_config = {'filters': {'items': {}}}
+
+        state_not_submitted = {None: 'Not yet submitted'}
+        state_choices = {state.value: state.title for state in PaperRevisionState}
+        self.static_items = OrderedDict([
+            ('state', {'title': _('State'),
+                       'filter_choices': OrderedDict(state_not_submitted.items() + state_choices.items())})
+        ])
+        self.list_config = self._get_config()
+
+    def _build_query(self):
+        return (Contribution.query.with_parent(self.event)
+                .order_by(Contribution.friendly_id)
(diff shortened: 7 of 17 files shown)


indico/modules/events/papers/blueprint.py
```diff
@@ -34,6 +34,7 @@
                  methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/papers/enable/<any(content,layout):reviewing_type>', 'switch',
                  management.RHSwitchReviewingType, methods=('PUT', 'DELETE'))
+_bp.add_url_rule('/manage/papers/assign', 'assignment', management.RHPapersAssignment)
```
nop33: I'd use `RHPaperAssignmentList` and `/manage/papers/assignment-list/`

indico/modules/events/papers/templates/management/_paper_assignment_list.html
```diff
@@ -0,0 +1,115 @@
...
+                        </th>
+                    </tr>
+                </thead>
+                <tbody>
+                    {% for contrib in contribs %}
+                        {% set last_revision_state %}
+                            {% if contrib.paper_revisions %}
+                                {{ contrib.paper_revisions.0.state.title }}
+                            {% else %}
+                                {%- trans %}Paper not yet submitted{% endtrans -%}
```
nop33: In some places you use `-` to remove the whitespace and in others you don't. I'd suggest to stick to one of the two conventions. The best IMO is to remove the whitespace because it creates an ugly HTML structure.

indico/modules/events/papers/templates/management/_paper_assignment_list.html
```diff
@@ -0,0 +1,115 @@
...
+                        <th class="i-table layout-reviewers-column" data-sorter="false">
+                            {% trans %}Layout reviewers{% endtrans %}
+                        </th>
+                    </tr>
+                </thead>
+                <tbody>
+                    {% for contrib in contribs %}
+                        {% set last_revision_state %}
+                            {% if contrib.paper_revisions %}
+                                {{ contrib.paper_revisions.0.state.title }}
```
nop33: Does this even work? :exclamation: 

Why not `{{ contrib.paper_revisions[0].state.title }}`?

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
...
+    {%- trans -%}
+        Assign <strong>judges</strong>, <strong>content reviewers</strong> and <strong>layout reviewers</strong>
+        to contributions in order to allow assessing its paper revisions.
+    {%- endtrans -%}
+{% endblock %}
+
+{% block content %}
+    <div class="toolbars space-after">
+        <div class="toolbar">
+            <div class="group">
```
nop33: As explained in our [toolbars wiki page](https://github.com/indico/indico/wiki/TL-02:-Toolbars), single buttons in a toolbar don't need to be wrapped by the `.group` element anymore. Apply this to the rest of your templates as well! We currently have some (quite many actually) templates in Indico that still use the `.group` element even if it includes 1 button, but we'll remove them eventually. We just gotta make sure that any new code doesn't increase the amount of code we need to change!

indico/modules/events/papers/templates/management/_paper_assignment_list.html
```diff
@@ -0,0 +1,115 @@
...
+                            {% else %}
+                                {%- trans %}Paper not yet submitted{% endtrans -%}
+                            {% endif %}
+                        {% endset %}
+                        <tr id="contrib-{{ contrib.id }}" class="i-table contribution-row"
+                            data-friendly-id="{{ contrib.friendly_id }}"
+                            data-title="{{ contrib.title }}">
+                            <td class="i-table id-column">
+                                <span class="vertical-aligner">
+                                    <input type="checkbox" class="select-row" name="contribution_id"
```
nop33: Apart from the `class` and `id` attributes, i'd put the rest in a new line each, similar to the way you did it in lines 44-46. Same applies for the `data-searchable` and `data-text` below and in general anywhere where the list of attributes needs to wrap.

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
...
+                    <li><a href="#">{% trans %}Layout reviewer{% endtrans %}</a></li>
+                </ul>
+                <a class="i-button">
+                    {%- trans %}Unassign{% endtrans -%}
+                </a>
+            </div>
+            <div class="group">
+                <a href="#" class="i-button icon-checkmark js-enable-if-checked disabled"
+                   data-href="#">
+                    {%- trans %}Accept{% endtrans -%}
```
nop33: I'd group the *accept*, *reject* and *to be corrected*.

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
...
+                </a>
+            </div>
+            <div class="group">
+                <a href="#" class="i-button icon-cross js-enable-if-checked disabled"
+                   data-href="#">
+                    {%- trans %}Reject{% endtrans -%}
+                </a>
+            </div>
+            <div class="group">
+                <a href="#" class="i-button icon-quill js-enable-if-checked disabled"
```
nop33: Not sure about the `quill` icon. Maybe `edit` would be better? We use the `quill` in a complete different context on Indico...

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
...
+{% endblock %}
+
+{% block content %}
+    <div class="toolbars space-after">
+        <div class="toolbar">
+            <div class="group">
+                <a class="i-button icon-checkbox-checked arrow js-dropdown" data-toggle="dropdown"></a>
+                <ul class="dropdown">
+                    <li>
+                        <a href="#" data-select-all="#assignment-list input:checkbox">{% trans %}All{% endtrans %}</a>
```
nop33: Selecting a row doesn't highlight it. Make sure the `handleRowSelection` method is executed (is should be executed by calling `setupListGenerator` as we do in the `setupContributionList` function of the contribution list (`contributions/templates/management/contributions.html`)

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
...
+                </button>
+            </div>
+            <div class="group">
+                <a class="i-button arrow js-dropdown" data-toggle="dropdown">
+                    {%- trans %}Assign{% endtrans -%}
+                </a>
+                <ul class="dropdown">
+                    <li><a href="#">{% trans %}Judge{% endtrans %}</a></li>
+                    <li><a href="#">{% trans %}Content reviewer{% endtrans %}</a></li>
+                    <li><a href="#">{% trans %}Layout reviewer{% endtrans %}</a></li>
```
nop33: Content and layout reviewer options should be available only if the setting is enabled.

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
...
+            <div class="group">
+                <a href="#" class="i-button icon-attachment js-enable-if-checked disabled"
+                   data-href="#">
+                    {%- trans %}Download papers{% endtrans -%}
+                </a>
+            </div>
+        </div>
+        <div class="toolbar">
+            <div class="group" id="filter-statistics">
+                {{ render_displayed_entries_fragment(contribs|length, total_entries) }}
```
nop33: Where did the denominator go? :smile: 
image

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
...
+            </div>
+            <div class="group">
+                <span class="i-button label icon-search"></span>
+                <input type="text" id="search-input" placeholder="{% trans %}Enter #id or search string{% endtrans %}">
```
nop33: The live search is not functional.

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
...
+                        data-title="{% trans %}Paper assignment list configuration{% endtrans %}"
+                        data-dialog-classes="list-filter-dialog"
+                        data-update='{"html": "#assignment-list", "filter_statistics": "#filter-statistics"}'
+                        data-ajax-dialog>
+                    {% trans %}Customize list{% endtrans %}
+                </button>
+            </div>
+            <div class="group">
+                <a class="i-button arrow js-dropdown" data-toggle="dropdown">
+                    {%- trans %}Assign{% endtrans -%}
```
nop33: The *Assign* and *Unassign* buttons should be disabled unless there are checked rows. Use:
`js-enable-if-checked disabled`

indico/modules/events/papers/templates/management/_paper_assignment_list.html
nop33: I think we should have a property in the contribution model to retrieve the latest revision...

ThiefMaster: yes, please do not use `.0.` in jinja but `[0]` instead. And yes, we should have a property/relationship that takes care of it. Since the relationship is slighly tricky - but albeit very useful so we can joinedload it e.g. for the listing - I would first go for a simple property and then change it to a relationship later.

indico/htdocs/js/indico/modules/papers.js
```diff
@@ -15,7 +15,27 @@
...
+    global.setupPaperAssignmentList = function setupPaperAssignmentList() {
+        var filterConfig = {
+            itemHandle: 'tr',
+            listItems: '#assignment-list tbody tr',
+            term: '#search-input',
+            state: '#filtering-state',
+            placeholder: '#filter-placeholder'
+        };
+
+        $('.list-section [data-toggle=dropdown]').closest('.group').dropdown();
```
ThiefMaster: `js-dropdown` doesn't work?

indico/modules/events/papers/controllers/management.py
```diff
@@ -71,3 +71,12 @@ def _process_PUT(self):
     def _process_DELETE(self):
         set_reviewing_state(self.event_new, request.view_args['reviewing_type'], False)
         return jsonify_data(flash=False, html=_render_paper_dashboard(self.event_new))
+
+
+class RHPapersAssignment(RHManagePapersBase):
+    """Assign contrubutions to reviewers and judges"""
+
+    def _process(self):
+        contribs = sorted(self.event_new.contributions, key=lambda x: x.friendly_id)
```
ThiefMaster: `key=attrgetter('friendly_id')`

ThiefMaster: Also, you could do that in the Jinja template. `event.contributions|sort(attribute='friendly_id')`

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
...
+{% from 'events/papers/management/_paper_assignment_list.html' import render_paper_assignment_list %}
+{% from 'events/management/_lists.html' import render_displayed_entries_fragment %}
+
+{% block title %}
+    {% trans %}Paper Assignment{% endtrans %}
+{% endblock %}
+
+{% block description %}
+    {%- trans -%}
+        Assign <strong>judges</strong>, <strong>content reviewers</strong> and <strong>layout reviewers</strong>
```
ThiefMaster: *judges and reviewers* to avoid having to use conditionals depending on what's enabled (for such conditionals you'd need to repeat the whole sentence since you cannot have them inside trans blocks)

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
...
+                        <a href="#" data-select-all="#assignment-list input:checkbox">{% trans %}All{% endtrans %}</a>
+                    </li>
+                    <li>
+                        <a href="#" data-select-none="#assignment-list input:checkbox">{% trans %}None{% endtrans %}</a>
+                    </li>
+                </ul>
+            </div>
+            <div class="group">
+                <button class="i-button icon-settings js-dialog-action js-customize-list highlight"
+                        data-href="#"
```
ThiefMaster: I guess this is WIP?

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
...
+                        data-title="{% trans %}Paper assignment list configuration{% endtrans %}"
```
ThiefMaster: Just "Customize list" or something like that? The title doesn't need to be super verbose

indico/modules/events/papers/templates/management/assignment.html
ThiefMaster: I'd move them into a dropdown

mvidalgarcia: I thought about it but wasn't sure about the button label: _Mark as_, _Change state_? Suggestions are welcome.

indico/modules/events/papers/templates/management/assignment.html
mvidalgarcia: Forgot to mention that is WIP, my bad. I'm currently working on the list generator that will solve it.

indico/modules/events/papers/templates/management/_paper_assignment_list.html
```diff
@@ -0,0 +1,115 @@
...
+                                    {%- endfor %}
+                                </span>
+                            </td>
+                        </tr>
+                    {% endfor %}
+                </tbody>
+            </table>
+        </form>
+    {%- else %}
+        {%- call message_box('info') %}
```
OmeGak: I would move this to be the first block. This helps see more clearly the implications of the `if` block.

```jinja
{% if not contribs %}
    {%- call message_box('info') %}
    ...
{% else %}
    <form method="POST">
    ...
{% endif %}
```

indico/modules/events/papers/templates/management/_paper_assignment_list.html
OmeGak: @ThiefMaster, muy importante.

indico/modules/events/papers/templates/management/_paper_assignment_list.html
```diff
@@ -0,0 +1,115 @@
...
+                            {% trans %}Content reviewers{% endtrans %}
+                        </th>
+                        <th class="i-table layout-reviewers-column" data-sorter="false">
+                            {% trans %}Layout reviewers{% endtrans %}
+                        </th>
+                    </tr>
+                </thead>
+                <tbody>
+                    {% for contrib in contribs %}
+                        {% set last_revision_state %}
```
OmeGak: Not sure if we want to start using `set` blocks with logic inside... Opinions? 

Another way to write this would be:

```jinja
{% if contrib.paper_revisions %}
    {% set last_revision_state = contrib.paper_revisions.0.state.title %}
{% else %}
    {% set last_revision_state %}
        {%- trans %}Paper not yet submitted{% endtrans -%}
    {% endset %}
{% endif %}
```

indico/modules/events/papers/templates/management/_paper_assignment_list.html
OmeGak: Not a strong opinion but I would definitely **not** put `type` on a new line since it's the most relevant attribute on an `<input>` tag.

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,95 @@
+{% extends 'layout/full_content_base.html' %}
+{% from 'events/papers/management/_paper_assignment_list.html' import render_paper_assignment_list %}
+{% from 'events/management/_lists.html' import render_displayed_entries_fragment %}
+
+{% block title %}
+    {% trans %}Paper Assignment{% endtrans %}
```
OmeGak: The title should be the same in all paper-related pages. In this case, `Call for Papers` I believe. 
This we can move to the `subtitle` block.

indico/modules/events/papers/templates/management/assignment.html
OmeGak: I would leave `{# TODO: ... #}` commets not to forget. It is ok to commit those into `devel` in early stages of a new release when the blueprint/page is still missing.

indico/modules/events/papers/templates/management/assignment.html
OmeGak: `icon-undo` maybe? :confused:

indico/modules/events/papers/templates/management/assignment.html
OmeGak: It just tends to infinity.

indico/modules/events/papers/templates/management/assignment.html
OmeGak: Just wondering if sending papers back for corrections make sense... Shouldn't we enforce leaving some feedback for the submitter?

indico/modules/events/papers/templates/management/_paper_assignment_list.html
ThiefMaster: I'd avoid set blocks except where useful. If there's lots of logic inside it might be better suited in a macro.

indico/modules/events/papers/templates/management/assignment.html
nop33: The comment field should become mandatory in that case.

indico/modules/events/papers/blueprint.py
```diff
@@ -36,6 +36,9 @@
                  methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/papers/enable/<any(content,layout):reviewing_type>', 'switch',
                  management.RHSwitchReviewingType, methods=('PUT', 'DELETE'))
+_bp.add_url_rule('/manage/papers/assignment-list', 'assignment', management.RHPapersAssignmentList)
+_bp.add_url_rule('/manage/papers/assign/customize', 'customize_assignment_list', management.RHAssignmentListCustomize,
```
nop33: I would use `/manage/papers/assignment-list/customize` here and in the line above `/manage/papers/assignment-list/` (slash at the end)

indico/modules/events/papers/lists.py
```diff
@@ -0,0 +1,140 @@
...
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from collections import OrderedDict
+from operator import attrgetter
+from flask import request
```
nop33: flask < operator

indico/modules/events/papers/lists.py
```diff
@@ -0,0 +1,140 @@
...
+from sqlalchemy.orm import subqueryload
+
+from indico.core.db import db
+from indico.modules.events.contributions import Contribution
+from indico.modules.events.papers.models.revisions import PaperRevisionState, PaperRevision
+from indico.modules.events.util import ListGeneratorBase
+from indico.web.flask.templating import get_template_module
+from indico.util.i18n import _
```
nop33: util < web

indico/modules/events/papers/lists.py
```diff
@@ -0,0 +1,140 @@
...
+    """Listing and filtering actions in a paper assignment list."""
+
+    endpoint = '.assignment'
+    list_link_type = 'paper_asssignment_management'
+
+    def __init__(self, event):
+        super(PaperAssignmentListGenerator, self).__init__(event)
+        self.default_list_config = {
+            'items': ('state', ),
+            'filters': {'items': {}}}
```
nop33:
``` python
self.default_list_config = {
    'items': ('state', ),
    'filters': {'items': {}}
}
```

indico/modules/events/papers/lists.py
```diff
@@ -0,0 +1,140 @@
...
+
+        state_not_submitted = {None: 'Not yet submitted'}
```
nop33: Add translations everywhere: `_('Not yet submitted')`

indico/modules/events/papers/lists.py
```diff
@@ -0,0 +1,140 @@
...
+        static_item_ids, dynamic_item_ids = self._split_item_ids(list_config['items'], 'static')
+        static_columns = self._get_static_columns(static_item_ids)
+        return {'contribs': contributions, 'total_entries': total_entries, 'selected_entry': selected_entry,
+                'static_columns': static_columns}
+
+    def render_list(self):
+        """Render the contribution list template components.
+
+        :return: dict containing the list's entries, the fragment of
+                 displayed entries and whether the contrib passed is displayed
```
nop33: *"...whether the contrib**ution**..."*, it's docstring, no need for abbreviations!

nop33: btw, I'd change it also in the contribution list files as well ;)

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,90 @@
...
+{% endblock %}
+
+{% block subtitle %}
+    {% trans %}Paper Assignment{% endtrans %}
+{% endblock %}
+
+{% block description %}
+    {%- trans -%}
+        Assign <strong>judges</strong> and <strong>reviewers</strong> to contributions in order to allow assessing
+        its paper revisions.
```
nop33: *its* or *theirs*? I am not sure which one fits the best.

indico/modules/events/papers/templates/management/assignment.html
```diff
@@ -0,0 +1,90 @@
...
+                    <a href="#" data-select-none="#assignment-list input:checkbox">{% trans %}None{% endtrans %}</a>
+                </li>
+            </ul>
+            <button class="i-button icon-settings js-dialog-action js-customize-list highlight"
+                    data-href="{{ url_for('.customize_assignment_list', event) }}"
+                    data-title="{% trans %}Customize list{% endtrans %}"
+                    data-dialog-classes="list-filter-dialog"
+                    data-update='{"html": "#assignment-list", "filter_statistics": "#filter-statistics"}'
+                    data-ajax-dialog>
+                {% trans %}Customize list{% endtrans %}
```
nop33: `{%- trans %}Customize list{% endtrans -%}`

indico/modules/events/papers/templates/management/assignment_list_filter.html
```diff
@@ -0,0 +1,84 @@
...
+            </div>
+        {% endif %}
+    </div>
+{% endmacro %}
+
+<div class="action-box">
+    <div class="section">
+        <div class="icon icon-wrench"></div>
+        <div class="text">
+            <div class="label">Customize paper asssignment list</div>
```
nop33: `{% trans %}`

indico/modules/events/papers/templates/management/assignment.html
mvidalgarcia: None of them :smile:, I think _their_ fits better

indico/modules/events/papers/lists.py
```diff
@@ -0,0 +1,141 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from collections import OrderedDict
+from flask import request
```
ThiefMaster: flask and sqlalchemy are not stdlib

indico/modules/events/contributions/models/contributions.py
```diff
@@ -332,6 +332,21 @@ def __table_args__(cls):
...
 
+    @declared_attr
+    def paper_last_revision(cls):
+        # Incompatible with joinedload
+        return db.relationship(
+            'PaperRevision',
+            uselist=False,
+            lazy=True,
+            viewonly=True,
+            primaryjoin=((PaperRevision.contribution_id == cls.id) &
```
ThiefMaster: I'd use `db.and()` instead of `&` - more readable IMO when wrapped

indico/modules/events/contributions/models/contributions.py
```diff
@@ -332,6 +332,21 @@ def __table_args__(cls):
         )
     )
 
+    @declared_attr
+    def paper_last_revision(cls):
```
ThiefMaster: `latest_paper_revision` or `current_paper_revision`?

indico/modules/events/papers/lists.py
```diff
@@ -0,0 +1,141 @@
...
+                    state_criteria.append(Contribution.paper_last_revision
+                                          .has(PaperRevision.state == int(filter_state)))
+            if state_criteria:
+                criteria.append(db.or_(*state_criteria))
+
+        filter_cols = {'track': Contribution.track_id,
+                       'session': Contribution.session_id,
+                       'type': Contribution.type_id}
+        for key, column in filter_cols.iteritems():
+            ids = set(filters['items'].get(key, ()))
```
ThiefMaster: the `, ()` is not needed since you check for falsiness in the next line anyway, so the default `None` is fine

indico/modules/events/papers/templates/management/_paper_assignment_list.html
```diff
@@ -0,0 +1,157 @@
+{% from 'message_box.html' import message_box %}
+
+{% macro render_paper_assignment_list(event, total_entries, contribs, static_columns) %}
+    {% set tz = event.display_tzinfo %}
```
ThiefMaster: unused. also, management wouldn't use `display_tzinfo` but just `tzinfo`

indico/modules/events/papers/templates/management/_paper_assignment_list.html
```diff
@@ -0,0 +1,157 @@
...
+                            {% endset %}
+                        {% endif %}
+                        <tr id="contrib-{{ contrib.id }}" class="i-table contribution-row"
+                            data-friendly-id="{{ contrib.friendly_id }}"
+                            data-title="{{ contrib.title }}">
+                            <td class="i-table id-column">
+                                <span class="vertical-aligner">
+                                    <input type="checkbox" class="select-row"
+                                           name="contribution_id"
+                                           value="{{ contrib.id }}">
```
ThiefMaster: fits in the previous line

indico/modules/events/papers/templates/management/_paper_assignment_list.html
```diff
@@ -0,0 +1,157 @@
...
+                            {% endfor %}
+                            <td class="i-table revision-column">
+                                <span class="vertical-aligner">
+                                    {%- if contrib.paper_revisions -%}
+                                        {{ contrib.paper_revisions | length }}
+                                    {%- endif -%}
+                                </span>
+                            </td>
+                            <td class="i-table person-row-cell"
+                                data-searchable="{{ contrib.paper_judges|map(attribute='name')|join(', ')|lower }}">
```
ThiefMaster: `|join(', ', attribute='name')` - no need  for the separate `|map()`

http://jinja.pocoo.org/docs/dev/templates/#join

indico/modules/events/papers/templates/management/assignment_list_filter.html
```diff
@@ -0,0 +1,84 @@
...
+        </div>
+        {% if filter_choices %}
+            <div class="i-label filter">
+                <a class="icon-filter" data-toggle="dropdown"></a>
+                <ul class="dropdown" style="display: none">
+                    {% for value, caption in filter_choices.iteritems() %}
+                        <li>
+                            <input type="checkbox" name="field_{{ item_id }}" value="{{ value }}"
+                                   id="field_{{ item_id }}_option_{{ value }}"
+                                   {% if value|string in filters[item_id]|string %}
```
ThiefMaster: `{{ 'checked' if value|string in filters[item_id] }}`?

Also, `filters[item_id]|string` looks very wrong to me. Are you by any chance converting a list/set to a string and then check whether another string is in it?

indico/modules/events/papers/templates/management/assignment_list_filter.html
```diff
@@ -0,0 +1,84 @@
...
+                <li><a href="#" id="list-filter-select-all">{% trans %}All{% endtrans %}</a></li>
+                <li><a href="#" id="list-filter-select-none">{% trans %}None{% endtrans %}</a></li>
+            </ul>
+            <a class="i-button js-reset-btn icon-close warning" type="button">
+                {%- trans %}Clear filters{% endtrans -%}
+            </a>
+        </div>
+    </div>
+</div>
+<div class="success-message-box js-clear-filters-message" style="display: none">
```
ThiefMaster: `none;`

indico/modules/events/papers/lists.py
mvidalgarcia: `set(None)` -> `'NoneType' object is not iterable`

ThiefMaster: ah, didn't spot the set() around it

indico/modules/events/papers/templates/management/assignment_list_filter.html
mvidalgarcia: The problem is when the value is `None` (in cases like "Not yet submitted", "No track", etc.), then the check is `if "None" in [None, ...]` which is false and must be true.

indico/modules/events/papers/lists.py
```diff
@@ -0,0 +1,141 @@
...
+        Retrieve information needed for the header of the static columns.
+
+        :return: a list of {'id': ..., 'caption': ...} dicts
+        """
+        return [{'id': id_, 'caption': self.static_items[id_]['title']} for id_ in self.static_items if id_ in ids]
+
+    def _build_query(self):
+        return (Contribution.query.with_parent(self.event)
+                .order_by(Contribution.friendly_id)
+                .options(subqueryload('paper_last_revision')))
```
pferreir: It almost looks easy now. :trollface:

indico/modules/events/papers/lists.py
pferreir: (he means they should be in a separate block)

indico/modules/events/papers/templates/management/assignment_list_filter.html
mvidalgarcia: Yeah, that works! :+1:

indico/modules/events/contributions/models/contributions.py
```diff
@@ -332,6 +332,21 @@ def __table_args__(cls):
...
+        return db.relationship(
+            'PaperRevision',
+            uselist=False,
+            lazy=True,
+            viewonly=True,
+            primaryjoin=(db.and_((PaperRevision.contribution_id == cls.id),
+                         (PaperRevision.submitted_dt == (db.select([db.func.max(PaperRevision.submitted_dt)])
+                                                         .where(PaperRevision.contribution_id == cls.id)
+                                                         .correlate_except(PaperRevision)
+                                                         .as_scalar()))))
```
pferreir: Indentation seems wrong.

indico/modules/events/contributions/models/contributions.py
```diff
@@ -332,6 +332,21 @@ def __table_args__(cls):
...
+    @declared_attr
+    def paper_last_revision(cls):
+        # Incompatible with joinedload
+        return db.relationship(
+            'PaperRevision',
+            uselist=False,
+            lazy=True,
+            viewonly=True,
+            primaryjoin=(db.and_((PaperRevision.contribution_id == cls.id),
+                                 (PaperRevision.submitted_dt == (db.select([db.func.max(PaperRevision.submitted_dt)])
```
ThiefMaster: I would assign the subquery to a `subquery` variable and use `PaperRevision.submitted_dt == subquery.as_scalar()` here (or even move the `as_scalar()` to the `subquery = ...` too)

indico/modules/events/contributions/models/contributions.py
```diff
@@ -332,6 +332,21 @@ def __table_args__(cls):
...
 
+    @declared_attr
+    def paper_last_revision(cls):
+        # Incompatible with joinedload
+        return db.relationship(
+            'PaperRevision',
+            uselist=False,
+            lazy=True,
+            viewonly=True,
+            primaryjoin=(db.and_((PaperRevision.contribution_id == cls.id),
```
ThiefMaster: no need for the innermost `()` now; they were only needed because of the precedence of `&`

indico/modules/events/papers/controllers/management.py
```diff
@@ -22,6 +22,7 @@
 from indico.modules.events.papers.forms import PaperTeamsForm, make_competences_form
 from indico.modules.events.papers.operations import (set_reviewing_state, update_team_members, create_competences,
                                                      update_competences)
+from indico.modules.events.papers.lists import PaperAssignmentListGenerator
```
nop33: lists < operations

indico/modules/events/contributions/lists.py
```diff
@@ -122,7 +122,7 @@ def render_list(self, contrib=None):
         :param contrib: Used in RHs responsible for CRUD operations on a
                         contribution.
         :return: dict containing the list's entries, the fragment of
-                 displayed entries and whether the contrib passed is displayed
+                 displayed entries and whether the contribution passed is displayed
```
nop33: Docstrings should wrap after 80 characters and python code after 120. Make sure you setup your editor to indicate that 😉

indico/modules/events/papers/lists.py
```diff
@@ -0,0 +1,142 @@
...
+class PaperAssignmentListGenerator(ListGeneratorBase):
+    """Listing and filtering actions in a paper assignment list."""
+
+    endpoint = '.assignment'
+    list_link_type = 'paper_asssignment_management'
+
+    def __init__(self, event):
+        super(PaperAssignmentListGenerator, self).__init__(event)
+        self.default_list_config = {
+            'items': ('state', ),
```
nop33: extra space after `'state',`

indico/modules/events/contributions/lists.py
mvidalgarcia: :disappointed: http://stackoverflow.com/questions/24125957/multiple-right-margins-in-pycharm
I'll take it into account tho

indico/modules/events/papers/blueprint.py
```diff
@@ -39,6 +39,8 @@
 _bp.add_url_rule('/manage/papers/assignment-list/', 'assignment', management.RHPapersAssignmentList)
 _bp.add_url_rule('/manage/papers/assignment-list/customize', 'customize_assignment_list',
                  management.RHAssignmentListCustomize, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/papers/assignment-list/change-state', 'manage_papers_state',
```
ThiefMaster: Don't we call this "judging" in the abstracts module (RH name, URL, endpoint name, etc)? In that case I'd use similar naming here for consistency.

indico/modules/events/papers/controllers/management.py
```diff
@@ -141,3 +142,49 @@ def _process_GET(self):
...
+        ids = map(int, request.form.getlist('contribution_id'))
+        self.contributions = self._contrib_query.filter(Contribution.id.in_(ids)).all()
+
+
+class RHBulkPaperStateChange(RHManagePapersActionsBase):
+    """Perform bulk change state operations on selected papers"""
+
+    def _process(self):
+        form = BulkPaperJudgmentForm(event=self.event_new, judgment=request.form.get('judgment'),
+                                     contribution_id=[a.id for a in self.contributions])
```
ThiefMaster: Copied from abstracts I guess? ;)

`a` doesn't fit here much, I'd use `c` or the generic `x` (but usually I go for `c` for contributions)

indico/modules/events/papers/operations.py
```diff
@@ -57,3 +61,21 @@ def update_competences(user_competences, competences):
     event.log(EventLogRealm.management, EventLogKind.positive, 'Papers',
               "Updated competences for user {}".format(user_competences.user.full_name), session.user,
               data={'Competences': ', '.join(competences)})
+
+
+def paper_change_state(contribution, contrib_data, judgment, judge, send_notifications=False):
```
ThiefMaster: `judge_paper`?

indico/modules/events/papers/controllers/management.py
```diff
@@ -141,3 +142,49 @@ def _process_GET(self):
...
+            judgment_data, contrib_data = form.split_data
+            # TODO: For now, I'm considering that any submitted paper contribution can change
+            # its state (even rejected ones)
+            submitted_contributions = [c for c in self.contributions if c.paper_last_revision]
+            for contrib in submitted_contributions:
+                paper_change_state(contrib, contrib_data, judge=session.user, **judgment_data)
+            num_judged_contribs = len(submitted_contributions)
+            num_prejudged_contribs = len(self.contributions) - num_judged_contribs
+            if num_judged_contribs:
+                flash(ngettext("One paper contribution has been judged.",
```
ThiefMaster: just "paper" or "contribution's paper" - "paper contribution" sounds rather weird.

indico/modules/events/papers/forms.py
```diff
@@ -49,3 +54,38 @@ def __init__(self, *args, **kwargs):
...
+    judgment_comment = TextAreaField(_("Comment"), render_kw={'placeholder': _("Leave a comment for the submitter...")})
+    send_notifications = BooleanField(_("Send notifications to submitter"), default=True)
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event')
+        super(BulkPaperJudgmentForm, self).__init__(*args, **kwargs)
+        if kwargs['judgment']:
+            self._remove_unused_fields(kwargs['judgment'])
+
+    def _remove_unused_fields(self, judgment):
```
ThiefMaster: This can go away if we don't have any fields that are specific to certain judgment actions.

indico/modules/events/papers/operations.py
```diff
@@ -57,3 +61,21 @@ def update_competences(user_competences, competences):
...
+    log_data = {'New state': orig_string(judgment.title)}
```
ThiefMaster: I'd move that down so all the logging stuff is at the end

indico/modules/events/papers/operations.py
```diff
@@ -57,3 +61,21 @@ def update_competences(user_competences, competences):
...
+    if judgment == PaperAction.accept:
+        contribution.paper_last_revision.state = PaperRevisionState.accepted
+    elif judgment == PaperAction.reject:
+        contribution.paper_last_revision.state = PaperRevisionState.rejected
+    elif judgment == PaperAction.to_be_corrected:
+        contribution.paper_last_revision.state = PaperRevisionState.to_be_corrected
+    PaperReviewComment(paper_revision=contribution.paper_last_revision, user=judge,
```
ThiefMaster: So we do want to save the judgment comment  as a normal comment in the end and not in a separate field on the revision?

Anyway, while this works (due to being added to the SA session since you link it to he prevision) I think it's slighly ugly to create an instance without using it anywhere. I'd rather do it like this:

    if contrib_data['judgment_comment']:
        comment = PaperReviewComment(user=judge, text=contrib_data['judgment_comment'])
        contribution.paper_last_revision.comments.append(comment)

indico/modules/events/papers/operations.py
```diff
@@ -57,3 +61,21 @@ def update_competences(user_competences, competences):
...
+                       text=contrib_data['judgment_comment'])
+    log_data['Sent notifications'] = send_notifications
+    if send_notifications:
+        pass  # TODO: manage notifications
+    logger.info('Paper contribution %s changed state by %s', contribution.title, judge)
```
ThiefMaster: `%r` for the second one. I'd also include the new state.

ThiefMaster: like before, "paper contribution" sounds really strange.

indico/modules/events/papers/operations.py
OmeGak: Just clarifying, the rationale is to use verbs for functions/methods as much as possible.

indico/modules/events/papers/controllers/management.py
```diff
@@ -111,3 +113,78 @@ def _process(self):
...
+    """Perform bulk change state operations on selected papers"""
+
+    def _process(self):
+        form = BulkPaperJudgmentForm(event=self.event_new, judgment=request.form.get('judgment'),
+                                     contribution_id=[a.id for a in self.contributions])
+        if form.validate_on_submit():
+            judgment_data, contrib_data = form.split_data
+            # TODO: For now, I'm considering that any submitted paper contribution can change
+            # its state (even rejected ones)
+            submitted_contributions = [c for c in self.contributions if c.paper_last_revision]
```
OmeGak: This is actually how many papers have been submitted. Why not `num_papers`?

indico/modules/events/papers/controllers/management.py
```diff
@@ -111,3 +113,78 @@ def _process(self):
...
+            for contrib in submitted_contributions:
+                paper_change_state(contrib, contrib_data, judge=session.user, **judgment_data)
+            num_judged_contribs = len(submitted_contributions)
```
OmeGak: It's the papers that are judged, not the contributions. I would name these `num_judged_papers`.

indico/modules/events/papers/controllers/management.py
```diff
@@ -111,3 +113,78 @@ def _process(self):
...
+            num_prejudged_contribs = len(self.contributions) - num_judged_contribs
```
OmeGak: I don't think this has anything to do with the paper having been prejudged. Rather, it's simply contributions that haven't received a paper submission. What about renaming the variable `num_missing_paper`?

indico/modules/events/papers/operations.py
OmeGak: I also thought we settled for adding the judgment comment as a field in the revision. Not that I disagree with this approach, but just wondering if I missed something.

indico/modules/events/papers/operations.py
```diff
@@ -99,8 +98,7 @@ def judge_paper(contribution, contrib_data, judgment, judge, send_notifications=
     elif judgment == PaperAction.to_be_corrected:
         contribution.paper_last_revision.state = PaperRevisionState.to_be_corrected
     if contrib_data['judgment_comment']:
```
ThiefMaster: I don't think this check is needed - assigning an empty comment does no harm

indico/htdocs/js/indico/modules/papers.js
```diff
@@ -15,7 +15,26 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
-(function() {
+/* global setupListGenerator:false, applySearchFilters:false, setupTableSorter: false */
```
ThiefMaster: no space after the last `:` for consistency

indico/modules/events/papers/operations.py
```diff
@@ -81,3 +86,24 @@ def close_cfp(event):
     event.cfp.close()
     logger.info("Call for papers for %r closed by %r", event, session.user)
     event.log(EventLogRealm.management, EventLogKind.negative, 'Papers', 'Call for papers closed', session.user)
+
+
+@no_autoflush
+def judge_paper(contribution, contrib_data, judgment, judge, send_notifications=False):
+
```
ThiefMaster: there shouldn't be an empty line here

indico/modules/events/papers/operations.py
```diff
@@ -81,3 +86,24 @@ def close_cfp(event):
...
+    if judgment == PaperAction.accept:
+        contribution.paper_last_revision.state = PaperRevisionState.accepted
+    elif judgment == PaperAction.reject:
+        contribution.paper_last_revision.state = PaperRevisionState.rejected
+    elif judgment == PaperAction.to_be_corrected:
+        contribution.paper_last_revision.state = PaperRevisionState.to_be_corrected
+    if contrib_data['judgment_comment']:
```
ThiefMaster: you can always set the comment. that way judging it again (even though this shouldn't happen) would properly overwrite an existing comment

indico/modules/events/papers/operations.py
```diff
@@ -81,3 +86,24 @@ def close_cfp(event):
...
+        contribution.paper_last_revision.judgment_comment = contrib_data['judgment_comment']
+    db.session.flush()
+    log_data = {'New state': orig_string(judgment.title), 'Sent notifications': send_notifications}
+    if send_notifications:
+        pass  # TODO: manage notifications
+    logger.info('Paper %s changed state by %r to %s', contribution.title, judge,
+                orig_string(judgment.title))
+    contribution.event_new.log(EventLogRealm.management, EventLogKind.change, 'Papers',
+                               'Paper "{}" changed state'.format(contribution.title), judge, data=log_data)
```
ThiefMaster: ".. was judged .." and I'd include the friendly id in this log message and similar ones: `Paper #123 ("foo") was judged...`

Might be a good idea to add this property to `Contribution` (we also have it in `Abstract`) to make it easier/cleaner:

```python
    @property
    def verbose_title(self):
        return '#{} ({})'.format(self.friendly_id, self.title)
```

ThiefMaster: use `orig_string()` here, too. we don't want translated messages in logs

indico/modules/events/papers/operations.py
```diff
@@ -81,3 +86,24 @@ def close_cfp(event):
...
+        contribution.paper_last_revision.state = PaperRevisionState.rejected
+    elif judgment == PaperAction.to_be_corrected:
+        contribution.paper_last_revision.state = PaperRevisionState.to_be_corrected
+    if contrib_data['judgment_comment']:
+        contribution.paper_last_revision.judgment_comment = contrib_data['judgment_comment']
+    db.session.flush()
+    log_data = {'New state': orig_string(judgment.title), 'Sent notifications': send_notifications}
+    if send_notifications:
+        pass  # TODO: manage notifications
+    logger.info('Paper %s changed state by %r to %s', contribution.title, judge,
```
ThiefMaster: ".. was judged .."

ThiefMaster: `Paper %r ..` and use `contribution` instead of `contribution.title`.
This is for the logfile, having the repr is *much* more useful there than just the title

indico/modules/events/papers/lists.py
```diff
@@ -0,0 +1,142 @@
...
+            'items': ('state',),
+            'filters': {'items': {}}
+        }
+
+        state_not_submitted = {None: _('Not yet submitted')}
+        track_empty = {None: _('No track')}
+        session_empty = {None: _('No session')}
+        type_empty = {None: _('No type')}
+        state_choices = {state.value: state.title for state in PaperRevisionState}
+        track_choices = OrderedDict((unicode(t.id), t.title) for t in sorted(self.event.tracks,
```
OmeGak: Out of curiosity, why do we need to `unicode` the IDs?

ThiefMaster: most likely it goes through json encoding at some point and JSON only supports string keys

PR #2761, opened 2016-12-12
Paper Reviewing Migration

pferreir:
```sql
SET search_path = event_paper_reviewing, pg_catalog;
ALTER TABLE templates ADD COLUMN description text NOT NULL;
ALTER TABLE templates ADD COLUMN name text NOT NULL;
ALTER TABLE review_questions ADD COLUMN type smallint NOT NULL;
ALTER TABLE review_questions ADD CONSTRAINT ck_review_questions_valid_enum_type CHECK ((type = ANY (ARRAY[1, 2])));
```

Sill missing:
 - Paper files
 - Revisions
 - Reviews / Judgments

Still to discuss:
 - Question ranges

### DIFF
--- a/indico/modules/events/papers/models/templates.py
+++ b/indico/modules/events/papers/models/templates.py
@@ -42,6 +42,15 @@ class PaperTemplate(StoredFileMixin, db.Model):
         index=True,
         nullable=False
     )
+    name = db.Column(
+        db.String,
+        nullable=False
+    )
+    description = db.Column(
+        db.Text,
+        nullable=False,
+        default=''
+    )
 
     event_new = db.relationship(
         'Event',
--- a/indico/modules/events/papers/settings.py
+++ b/indico/modules/events/papers/settings.py
@@ -24,7 +24,13 @@ paper_reviewing_settings = EventSettingsProxy('paper_reviewing', {
     'end_dt': None,
     'content_reviewing_enabled': True,
     'layout_reviewing_enabled': False,
+    'judge_deadline': None,
+    'layout_reviewer_deadline': None,
+    'content_reviewer_deadline': None,
 }, converters={
     'start_dt': DatetimeConverter,
     'end_dt': DatetimeConverter,
+    'judge_deadline': DatetimeConverter,
+    'layout_reviewer_deadline': DatetimeConverter,
+    'content_reviewer_deadline': DatetimeConverter,
 })
--- a/indico/modules/events/papers/settings.py
+++ b/indico/modules/events/papers/settings.py
@@ -22,6 +22,7 @@ from indico.modules.events.settings import EventSettingsProxy
 paper_reviewing_settings = EventSettingsProxy('paper_reviewing', {
     'start_dt': None,
     'end_dt': None,
+    'enforce_deadlines': False,
     'content_reviewing_enabled': True,
     'layout_reviewing_enabled': False,
     'judge_deadline': None,
--- a/indico/modules/events/papers/settings.py
+++ b/indico/modules/events/papers/settings.py
@@ -16,8 +16,26 @@
 
 from __future__ import unicode_literals
 
-from indico.core.settings.converters import DatetimeConverter
+from indico.core.settings.converters import DatetimeConverter, SettingConverter
 from indico.modules.events.settings import EventSettingsProxy
+from indico.util.struct.enum import IndicoEnum
+
+
+class PaperReviewingRole(int, IndicoEnum):
+    judge = 1
+    layout_reviewer = 2
+    content_reviewer = 3
+
+
+class RoleConverter(SettingConverter):
+    @staticmethod
+    def from_python(value):
+        return [v.name for v in value]
+
+    @staticmethod
+    def to_python(value):
+        return {PaperReviewingRole.get(v) for v in value}
+
 
 paper_reviewing_settings = EventSettingsProxy('paper_reviewing', {
     'start_dt': None,
@@ -28,10 +46,20 @@ paper_reviewing_settings = EventSettingsProxy('paper_reviewing', {
     'judge_deadline': None,
     'layout_reviewer_deadline': None,
     'content_reviewer_deadline': None,
+
+    # Notifications
+    'notify_on_added_to_event': {},
+    'notify_on_assigned_contrib': {},
+    'notify_on_paper_submission': {PaperReviewingRole.layout_reviewer, PaperReviewingRole.content_reviewer},
+    'notify_judge_on_review': True,
+    'notify_author_on_judgment': True
 }, converters={
     'start_dt': DatetimeConverter,
     'end_dt': DatetimeConverter,
     'judge_deadline': DatetimeConverter,
     'layout_reviewer_deadline': DatetimeConverter,
     'content_reviewer_deadline': DatetimeConverter,
+    'notify_on_added_to_event': RoleConverter,
+    'notify_on_assigned_contrib': RoleConverter,
+    'notify_on_paper_submission': RoleConverter
 })
--- a/indico_zodbimport/modules/event_abstracts.py
+++ b/indico_zodbimport/modules/event_abstracts.py
@@ -69,7 +69,7 @@ def strict_sanitize_email(email, fallback=None):
 
 
 # Add "temporary" column to model
-Contribution.legacy_track_id = db.Column(db.Integer, nullable=True)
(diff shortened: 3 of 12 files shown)


indico_zodbimport/modules/event_papers.py
```diff
@@ -0,0 +1,224 @@
...
+        for n, q in enumerate(self.pr._reviewingQuestions, 1):
+            question = PaperReviewQuestion(text=q._text, type=PaperReviewType.content, position=n, event_new=self.event)
+            self.event.paper_review_questions.append(question)
+
+        for n, q in enumerate(self.pr._layoutQuestions, 1):
+            question = PaperReviewQuestion(text=q._text, type=PaperReviewType.layout, position=n, event_new=self.event)
+            self.event.paper_review_questions.append(question)
+
+    def _migrate_templates(self):
+        for tid, old_tpl in self.pr._templates.viewitems():
```
ThiefMaster: tid is currently unused

indico_zodbimport/modules/event_papers.py
```diff
@@ -0,0 +1,224 @@
...
+                self.importer.print_error(cformat('%{red!}File not found on disk; skipping it [{}]')
+                                          .format(convert_to_unicode(old_filename)),
+                                          event_id=self.event.id)
+                continue
+
+            filename = secure_filename(convert_to_unicode(old_filename), 'attachment')
+            content_type = mimetypes.guess_type(old_filename)[0] or 'application/octet-stream'
+            tpl = PaperTemplate(filename=filename, content_type=content_type, size=size,
+                                storage_backend=storage_backend, storage_file_id=storage_path,
+                                name=old_tpl._Template__name, description=old_tpl._Template__description)
```
ThiefMaster: `name` should probably be sanitized (`secure_filename`)

indico_zodbimport/modules/event_papers.py
```diff
@@ -0,0 +1,224 @@
...
+        for role in contribution.legacy_paper_reviewing_roles:
+            if role.role == PaperReviewingRoleType.referee:
+                contribution.paper_judges.add(role.user)
+            elif role.role == PaperReviewingRoleType.reviewer:
+                contribution.paper_content_reviewers.add(role.user)
+            elif role.role == PaperReviewingRoleType.editor:
+                contribution.paper_layout_reviewers.add(role.user)
+
+    def _migrate_papers(self):
+        contributions = {c.id: c for c in (Contribution.find(event_new=self.event)
```
ThiefMaster: I'd use `Contribution.query.with_parent(self.event)`

indico_zodbimport/modules/event_papers.py
```diff
@@ -0,0 +1,224 @@
...
+                with db.session.no_autoflush:
+                    mig.run()
+                    db.session.flush()
+
+    def _iter_events(self):
+        query = (Event.query
+                 .filter(~Event.is_deleted)
+                 .filter(Contribution.legacy_paper_files.any() | Contribution.legacy_paper_reviewing_roles.any() |
+                         (Event.type_ == EventType.conference))
+                 .join(Contribution).group_by(Event.id))
```
ThiefMaster: I'd wrap before `.group_by()`. Also, not sure  whether that join() can give you duplicates. I'd rather use `Event.contributions.any(Contribution.legacy_paper_files.any())` in the `filter()`

indico/modules/events/papers/models/templates.py
```diff
@@ -42,6 +42,15 @@ class PaperTemplate(StoredFileMixin, db.Model):
         index=True,
         nullable=False
     )
+    name = db.Column(
+        db.Text,
```
ThiefMaster: I'd use `db.String` for this one

migrations/versions/201612051744_2287c5d5680c_add_paper_tables.py
```diff
@@ -153,6 +154,8 @@ def upgrade():
         'templates',
         sa.Column('id', sa.Integer(), nullable=False),
         sa.Column('event_id', sa.Integer(), nullable=False, index=True),
+        sa.Column('name', sa.Text(), nullable=False),
```
ThiefMaster: `sa.String()` if you change it in the model

indico/modules/events/papers/settings.py
```diff
@@ -16,15 +16,50 @@
...
+
+
+class RoleConverter(SettingConverter):
+    @staticmethod
+    def from_python(value):
+        return [v.name for v in value]
+
+    @staticmethod
+    def to_python(value):
+        return {PaperReviewingRole.get(v) for v in value}
```
pferreir: Yeah. I thought of that, but since it's the first place where we need this...

indico_zodbimport/modules/event_papers.py
pferreir: Yeah, and it will probably stay that way.

indico_zodbimport/modules/event_papers.py
pferreir: I guess you mean `filename`, right?

ThiefMaster: yeah, so nevermind

indico_zodbimport/modules/event_papers.py
```diff
@@ -0,0 +1,224 @@
...
+    def _migrate_paper_roles(self, contribution):
+        for role in contribution.legacy_paper_reviewing_roles:
+            if role.role == PaperReviewingRoleType.referee:
+                contribution.paper_judges.add(role.user)
+            elif role.role == PaperReviewingRoleType.reviewer:
+                contribution.paper_content_reviewers.add(role.user)
+            elif role.role == PaperReviewingRoleType.editor:
+                contribution.paper_layout_reviewers.add(role.user)
+
+    def _migrate_papers(self):
```
OmeGak: But this is still missing migrating the actual papers, right?

indico_zodbimport/modules/event_papers.py
```diff
@@ -0,0 +1,224 @@
...
+class EventPapersImporter(LocalFileImporterMixin, Importer):
+    def __init__(self, **kwargs):
+        kwargs = self._set_config_options(**kwargs)
+        super(EventPapersImporter, self).__init__(**kwargs)
+
+    def has_data(self):
+        return PaperFile.query.has_rows() or PaperReview.query.has_rows() or PaperRevision.query.has_rows()
+
+    def migrate(self):
+        self.migrate_event_pr()
```
OmeGak: Just out of curiosity. Why not moving the contents of `migrate_event_pr` here?

indico_zodbimport/modules/event_papers.py
```diff
@@ -0,0 +1,224 @@
...
+                with db.session.no_autoflush:
+                    mig.run()
+                    db.session.flush()
+
+    def _iter_events(self):
+        query = (Event.query
+                 .filter(~Event.is_deleted)
+                 .filter(db.or_(Event.contributions.any(Contribution.legacy_paper_files.any() |
+                                                        Contribution.legacy_paper_reviewing_roles.any()),
+                                (Event.type_ == EventType.conference))))
```
OmeGak: What about events that were conferences but, for some reason (accidentally, or for testing purposes), have been changed to meetings at the time of the migration? Is it a problem iterating through all events regardless their type?

ThiefMaster: That's covered by the check whether there are any contribs with papers or roles

indico_zodbimport/modules/event_papers.py
ThiefMaster: consistency, so far we always have nothing but simple method calls in `migrate()`

indico_zodbimport/modules/event_abstracts.py
```diff
@@ -69,7 +69,7 @@ def strict_sanitize_email(email, fallback=None):
 
 
 # Add "temporary" column to model
-Contribution.legacy_track_id = db.Column(db.Integer, nullable=True)
+# Contribution.legacy_track_id = db.Column(db.Integer, nullable=True)
```
pferreir: Yes. That basically breaks the whole thing since the column doesn't exist anymore. I decided to leave it commented so that we read it when rewriting the migration scripts.

indico_zodbimport/modules/event_papers.py
pferreir: There may be a couple that slip through, but it should be fine. It's unlikely people will use that info once the event has become a meeting.

PR #2764, opened 2016-12-13
Render CfA announcement as markdown

pferreir:
This basically handles (automatic) conversion from HTML to Markdown as users edit the announcement. It is then rendered accordingly to the currently set format.

### DIFF
--- a/indico/modules/events/abstracts/controllers/management.py
+++ b/indico/modules/events/abstracts/controllers/management.py
@@ -16,6 +16,8 @@
 
 from __future__ import unicode_literals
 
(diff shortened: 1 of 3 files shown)


indico/core/settings/proxy.py
```diff
@@ -167,7 +167,7 @@ class SettingProperty(object):
     """Expose a SettingsProxy value as a property.
 
     Override `attr` in a subclass for settings proxies that are tied
-    to specific objects (such as `EventSettigsProxy`)
```
ThiefMaster: 🙈

PR #2758, opened 2016-12-09
Generalize review-related SCSS/HTML/JS

OmeGak:

PR #2765, opened 2016-12-14
Bundle all SASS in screen.scss

OmeGak:
Missing:
- [x] Adapt once #2758 is merged
- [x] Make sure this doesn't break meeting themes
- [x] Test thoroughly (visit as many indico pages as possible to see if there are CSS bugs)

### DIFF
--- a/indico/MaKaC/webinterface/pages/contributionReviewing.py
+++ b/indico/MaKaC/webinterface/pages/contributionReviewing.py
@@ -123,9 +123,6 @@ class WPContributionModifBase(WPConferenceModifBase):
                '\n'.join(['<script src="{0}" type="text/javascript"></script>'.format(url)
                           for url in self._asset_env['mathjax_js'].urls()])
 
-    def getCSSFiles(self):
-        return WPConferenceModifBase.getCSSFiles(self) + self._asset_env['contributions_sass'].urls()
-
 
 class WPContributionReviewing( WPContributionModifBase ):
 
--- a/indico/web/assets/bundles.py
+++ b/indico/web/assets/bundles.py
@@ -422,43 +422,9 @@ base_js = Bundle(palette, jquery, angular, jed, utils, presentation, calendar, i
 SASS_BASE_MODULES = ["sass/*.scss",
                      "sass/base/*.scss",
                      "sass/custom/*.scss",
-                     "sass/partials/*.scss"]
+                     "sass/partials/*.scss",
+                     "sass/modules/**/*.scss"]
 
-
-def sass_module_bundle(module_name, depends=[]):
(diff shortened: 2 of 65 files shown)


indico/MaKaC/webinterface/pages/contributionReviewing.py
```diff
@@ -123,12 +123,6 @@ def _getHeadContent(self):
                '\n'.join(['<script src="{0}" type="text/javascript"></script>'.format(url)
                           for url in self._asset_env['mathjax_js'].urls()])
 
-    def getCSSFiles(self):
-        return WPConferenceModifBase.getCSSFiles(self) + self._asset_env['contributions_sass'].urls()
-
-    def getJSFiles(self):
```
ThiefMaster: this one isn't needed?

nop33:
The `.big` class in `.i-button` seems to be broken:

image

pferreir:
> The .big class in .i-button seems to be broken:

Now that's BIG.

nop33:
Something to take into account: The rules in `jquery-ui-custom.css` sometimes have higher specificity than the `screen.scss`. An example is the color of the links in dialogs, where we have
``` css
.ui-widget-content a {
    color: #4C6A83;
}
```
but in the `_links.scss` we have
``` scss
a {
    color: $blue;
    ...
}
```

indico/MaKaC/webinterface/pages/contributionReviewing.py
OmeGak: Why would it be? The point is for `contributions_sass` to be included in `screen.scss`, so we don't need to serve extra bundles through `getCSSFiles`. Am I missing anything?

OmeGak:
The big buttons are now fixed.

OmeGak:
The specifity problems should now be fixed. I think it only affected links within dialogs thanks to annoying super-intrusive jQuery UI theme.

indico/web/assets/bundles.py
```diff
@@ -422,43 +422,10 @@ def rjs_bundle(name, *files):
...
-news_sass = sass_module_bundle('news')
-users_sass = sass_module_bundle('users')
-auth_sass = sass_module_bundle('auth')
-surveys_sass = sass_module_bundle('surveys')
-timetable_sass = sass_module_bundle('timetable')
-sessions_sass = sass_module_bundle('sessions')
-markdown_sass = sass_module_bundle('markdown')
+                     "sass/partials/*.scss",
+                     "sass/modules/*.scss",
+                     "sass/modules/*/*.scss"]
```
ThiefMaster: `sass/modules/**/*.scss` might cover both this and the previous one (`**` matches within the directory and any subdirectory). At least if webassets supports `**`

PR #2766, opened 2016-12-14
Replace "Registration/Register" with "Application/Apply" when reform is moderated

nop33:
Close `RQF0683122` when merged and deployed

### DIFF
--- a/indico/htdocs/js/indico/modules/registration/form/tpls/registrationform.tpl.html
+++ b/indico/htdocs/js/indico/modules/registration/form/tpls/registrationform.tpl.html
@@ -60,7 +60,8 @@
         <div ng-switch="updateMode">
             <input type="hidden" value="{{ csrfToken }}" name="csrf_token">
             <input type="submit" id="regformSubmit"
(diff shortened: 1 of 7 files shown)


indico/htdocs/js/indico/modules/registration/form/tpls/registrationform.tpl.html
```diff
@@ -60,7 +60,8 @@
         <div ng-switch="updateMode">
             <input type="hidden" value="{{ csrfToken }}" name="csrf_token">
             <input type="submit" id="regformSubmit"
-                 ng-switch-default class="i-button highlight" value="{{ 'Register' | i18n }}">
+                   ng-switch-default class="i-button highlight"
+                   value="{{ isModerated ? 'Apply' : 'Register' | i18n }}">
```
ThiefMaster: I doubt that works as it should as far as i18n is concerned. I'd go for this which is ugly but will probably work (dunno about the string extraction part though, if we even extract strings from `|i18n`)

    {{ isModerated ? 'Apply'|i18n : 'Register'|i18n }}

PR #2767, opened 2016-12-14
Fix contribution fields validator

nop33:


indico/modules/events/contributions/contrib_fields.py
```diff
@@ -50,10 +50,11 @@ class ContribField(BaseField):
     def __init__(self, obj, management=True):
         super(ContribField, self).__init__(obj)
         self.management = management
+        self.is_required = obj.is_required
 
     @property
     def required_validator(self):
-        return Optional if self.management else DataRequired
+        return Optional if self.management or not self.is_required else DataRequired
```
ThiefMaster: https://github.com/nop33/indico/blob/master/indico/web/fields/base.py#L139-L140

This property shouldn't be accessed at all if the field is not required

ThiefMaster: That seems very strange. Your change just adds another case where the field is *not* required.

nop33: I found out that the problem was the it always thought we are in the management area. Thanks for the insights!

PR #2772, opened 2016-12-15
Fix conference logo and title alignment

nop33:
When the logo became higher than the title of the conference, the subtitle row of the conference would wrap next to it:

image

Making the logo align with the title only, fixes the problem:

image


nop33:
Is it better when the title is aligned on the top, or in the center? I would say center.


image

pferreir:
👍  for center. But we should check some of the conference events we host at CERN, and see what would be the implications in their current layout.

nop33:
Well, the current layout is broken :stuck_out_tongue_closed_eyes:

nop33:
I can query the DB to find conferences with logo and see how it looks like.

PR #2773, opened 2016-12-16
Migration of PR revisions, reviews and files 🎄

pferreir:
Including answers to questions, etc.

Particular things to take into account:
 * Fixes in Revision model (added missing attributes, constraints);
 * Questionnaires answered by judges ("referees") are ignored. Another option would be adding a third type of review (e.g. 'legacy') or remove the constraint that enforces one review per user per revision, but I don't think it's worth it.
 * Answers to deleted questions are removed (old interface was actually showing them, which was super confusing);
 * Mechanism to remove duplicate files from subsequent reviews, based on CRC32;
 * Revisions that were judged before being submitted (!!!) are just marked as judged;
 * Revision submission dates are taken from the latest uploaded file when there is one;
 * Revision submitters are set to Indico Janitor, since we have no way of know which author uploaded the paper (legacy didn't keep track of that);

Don't forget:

### DIFF
--- a/indico/modules/events/papers/models/reviews.py
+++ b/indico/modules/events/papers/models/reviews.py
@@ -119,7 +119,7 @@ class PaperReview(RenderModeMixin, db.Model):
 
     @return_ascii
     def __repr__(self):
-        return format_repr(self, 'id', 'revision', 'user_id', proposed_action=None)
+        return format_repr(self, 'id', 'type', 'revision', 'user_id', proposed_action=None)
 
     def can_edit(self, user, check_state=False):
         if user is None:
--- a/indico/modules/events/papers/models/reviews.py
+++ b/indico/modules/events/papers/models/reviews.py
@@ -119,7 +119,7 @@ class PaperReview(RenderModeMixin, db.Model):
 
     @return_ascii
     def __repr__(self):
-        return format_repr(self, 'id', 'type', 'revision', 'user_id', proposed_action=None)
+        return format_repr(self, 'id', 'type', 'revision_id', 'user_id', proposed_action=None)
 
     def can_edit(self, user, check_state=False):
         if user is None:
--- a/indico/modules/events/papers/models/revisions.py
+++ b/indico/modules/events/papers/models/revisions.py
@@ -16,14 +16,14 @@
 
 from __future__ import unicode_literals
 
-from sqlalchemy.ext.hybrid import hybrid_property
 
 from indico.core.db import db
 from indico.core.db.sqlalchemy import UTCDateTime, PyIntEnum
+from indico.core.db.sqlalchemy.descriptions import RenderModeMixin, RenderMode
 from indico.util.date_time import now_utc
 from indico.util.i18n import _
 from indico.util.locators import locator_property
-from indico.util.string import return_ascii, format_repr, MarkdownText
+from indico.util.string import return_ascii, format_repr
 from indico.util.struct.enum import TitledIntEnum
 
 
@@ -35,11 +35,14 @@ class PaperRevisionState(TitledIntEnum):
     to_be_corrected = 4
 
 
-class PaperRevision(db.Model):
+class PaperRevision(RenderModeMixin, db.Model):
     __tablename__ = 'revisions'
     __table_args__ = (db.Index(None, 'contribution_id', unique=True,
                                postgresql_where=db.text('state = {}'.format(PaperRevisionState.accepted))),
                       {'schema': 'event_paper_reviewing'})
+
+    possible_render_modes = {RenderMode.markdown}
+    default_render_mode = RenderMode.markdown
 
     id = db.Column(
         db.Integer,
@@ -92,22 +95,12 @@ class PaperRevision(db.Model):
         )
     )
 
+    judgment_comment = RenderModeMixin.create_hybrid_property('_judgment_comment')
+
     # relationship backrefs:
     # - comments (PaperReviewComment.paper_revision)
     # - files (PaperFile.paper_revision)
     # - reviews (PaperReview.revision)
-
-    @hybrid_property
-    def judgment_comment(self):
-        return MarkdownText(self._judgment_comment)
-
-    @judgment_comment.setter
-    def judgment_comment(self, value):
-        self._judgment_comment = value
-
-    @judgment_comment.expression
-    def judgment_comment(cls):
-        return cls._judgment_comment
 
     @return_ascii
     def __repr__(self):
--- a/indico/modules/events/papers/models/revisions.py
+++ b/indico/modules/events/papers/models/revisions.py
@@ -39,6 +39,14 @@ class PaperRevision(RenderModeMixin, db.Model):
     __tablename__ = 'revisions'
     __table_args__ = (db.Index(None, 'contribution_id', unique=True,
                                postgresql_where=db.text('state = {}'.format(PaperRevisionState.accepted))),
+                      db.CheckConstraint('(state IN ({}, {}, {})) = (judge_id IS NOT NULL)'
+                                         .format(PaperRevisionState.accepted, PaperRevisionState.rejected,
+                                                 PaperRevisionState.to_be_corrected),
+                                         name='judge_if_judged'),
+                      db.CheckConstraint('(state IN ({}, {}, {})) = (judgment_dt IS NOT NULL)'
+                                         .format(PaperRevisionState.accepted, PaperRevisionState.rejected,
+                                                 PaperRevisionState.to_be_corrected),
+                                         name='judgment_dt_if_judged'),
                       {'schema': 'event_paper_reviewing'})
 
     possible_render_modes = {RenderMode.markdown}
@@ -47,6 +55,11 @@ class PaperRevision(RenderModeMixin, db.Model):
     id = db.Column(
         db.Integer,
         primary_key=True
+    )
+    state = db.Column(
+        PyIntEnum(PaperRevisionState),
+        nullable=False,
+        default=PaperRevisionState.submitted
     )
     contribution_id = db.Column(
         db.Integer,
@@ -65,10 +78,15 @@ class PaperRevision(RenderModeMixin, db.Model):
         nullable=False,
         default=now_utc
     )
-    state = db.Column(
-        PyIntEnum(PaperRevisionState),
-        nullable=False,
-        default=PaperRevisionState.submitted
+    judge_id = db.Column(
+        db.Integer,
+        db.ForeignKey('users.users.id'),
+        index=True,
+        nullable=True
+    )
+    judgment_dt = db.Column(
+        UTCDateTime,
+        nullable=True
     )
     _judgment_comment = db.Column(
         'judgment_comment',
@@ -89,8 +107,18 @@ class PaperRevision(RenderModeMixin, db.Model):
     submitter = db.relationship(
         'User',
         lazy=True,
+        foreign_keys=submitter_id,
         backref=db.backref(
             'paper_revisions',
+            lazy='dynamic'
+        )
+    )
+    judge = db.relationship(
+        'User',
+        lazy=True,
+        foreign_keys=judge_id,
+        backref=db.backref(
+            'judged_papers',
             lazy='dynamic'
         )
     )
--- a/indico/modules/events/papers/settings.py
+++ b/indico/modules/events/papers/settings.py
@@ -47,6 +47,8 @@ paper_reviewing_settings = EventSettingsProxy('paper_reviewing', {
     'layout_reviewer_deadline': None,
     'content_reviewer_deadline': None,
     'announcement': '',
+    'scale_lower': -3,
+    'scale_upper': 3,
 
     # Notifications
     'notify_on_added_to_event': {},
--- a/indico_zodbimport/modules/event_papers.py
+++ b/indico_zodbimport/modules/event_papers.py
@@ -16,26 +16,35 @@
 
 from __future__ import unicode_literals, division
 
+import itertools
 import mimetypes
+from collections import defaultdict
 from operator import attrgetter
 
+import click
 from pytz import utc
 from sqlalchemy.orm import joinedload
 
 from indico.core.db import db
+from indico.core.storage import StorageError
 from indico.modules.events import Event
 from indico.modules.events.contributions.models.contributions import Contribution
+from indico.modules.events.features.util import set_feature_enabled
 from indico.modules.events.models.events import EventType
 from indico.modules.events.papers.models.competences import PaperCompetence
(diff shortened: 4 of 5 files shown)


indico/modules/events/papers/models/revisions.py
```diff
@@ -60,10 +77,22 @@ class PaperRevision(db.Model):
         nullable=False,
         default=now_utc
     )
-    state = db.Column(
-        PyIntEnum(PaperRevisionState),
+    #: ID of the user who judged the abstract
```
ThiefMaster: i think that's obvious ;)

indico/modules/events/papers/models/revisions.py
```diff
@@ -60,10 +77,22 @@ class PaperRevision(db.Model):
...
+    judge_id = db.Column(
+        db.Integer,
+        db.ForeignKey('users.users.id'),
+        index=True,
+        nullable=True
+    )
+    judgment_dt = db.Column(
+        UTCDateTime,
+        nullable=True,
```
ThiefMaster: i think we don't have the trailing comma elsewhere

indico/modules/events/papers/models/revisions.py
```diff
@@ -78,11 +107,24 @@ class PaperRevision(db.Model):
...
             lazy='dynamic'
         )
     )
+    judge = db.relationship(
+        'User',
+        lazy=True,
+        foreign_keys=judge_id,
+        backref=db.backref(
+            'judged_papers',
+            primaryjoin='(PaperRevision.judge_id == User.id) & ~Contribution.is_deleted',
```
ThiefMaster: I don't think this is going to work. `Contribution` isn't even joined.

indico/modules/events/papers/settings.py
```diff
@@ -46,6 +46,8 @@ def to_python(value):
     'judge_deadline': None,
     'layout_reviewer_deadline': None,
     'content_reviewer_deadline': None,
+    'scale_lower': 0,
```
ThiefMaster: We don't want to keep a -3...+3 default?

indico_zodbimport/modules/event_papers.py
```diff
@@ -169,6 +216,83 @@ def _migrate_paper_roles(self, contribution):
...
+                     else PaperRevisionState.submitted)
+            judge = old_judgment._author.user if old_judgment._submitted else None
+            judgment_dt = _to_utc(old_judgment._submissionDate) if old_judgment._submitted else None
+            last_file = (LegacyPaperFile.query.with_parent(contribution)
+                         .filter(LegacyPaperFile.revision_id == old_revision._version)
+                         .order_by(LegacyPaperFile.created_dt.desc())
+                         .limit(1)).first()
+            # Legacy didn't keep track of the submission date (nor submitter for that matter)
+            # we're taking the most recent uploaded file and using that one
+            # if there are no files (possible yet unusual), the DB default (now) will be used
```
ThiefMaster: could use the event's `end_dt` in that case

indico_zodbimport/modules/event_papers.py
```diff
@@ -182,17 +306,78 @@ def _migrate_papers(self):
                 continue
             contribution = contributions[contrib_id]
             self._migrate_paper_roles(contribution)
+            self._migrate_revisions(contribution, rm)
+
+    def _migrate_paper_files(self):
+        n = 0
+        # we look for paper files in increasing order, so that we can detect file reuse
+        q = (LegacyPaperFile.find(Contribution.event_new == self.event).join(Contribution)
```
ThiefMaster: For consistency I'd format it like this:

```python
        q = (LegacyPaperFile.query
             .filter(Contribution.event_new == self.event)
             .join(Contribution)
             .order_by(Contribution.id, LegacyPaperFile.revision_id))
```

indico_zodbimport/modules/event_papers.py
```diff
@@ -182,17 +306,78 @@ def _migrate_papers(self):
...
+                    revision = self.legacy_contrib_revision_map[(contribution_id, revision_id)]
+                else:
+                    revision = None
+
+                for lpf in files:
+                    # check whether the same file has been uploaded to a subsequent revision
+                    try:
+                        with lpf.open() as f:
+                            checksum = crc32(f.read())
+                            checksum_map[checksum] = lpf
```
ThiefMaster: I think you can already outdent here, you don't seem to be using `f` again below

ThiefMaster: I think you can already outdent here, you don't seem to be using `f` again below

indico/modules/events/papers/settings.py
pferreir: Do we?

indico_zodbimport/modules/event_papers.py
pferreir: Yeah, that would be probably better indeed.

indico/modules/events/papers/settings.py
ThiefMaster: No idea, seems to fit for the common(?) usecase of "disagree/negative - neutral - agree/positive"

pferreir: That's a very good question.

pferreir: Let's discuss that next year 😛 
But yeah, for now I'll leave it at -3...+3 like legacy.

indico/modules/events/papers/models/revisions.py
pferreir: Yeah, indeed. Also, I can't find any way of making this work inside a backref.

indico_zodbimport/modules/event_papers.py
```diff
@@ -16,26 +16,34 @@
...
 from indico.modules.events.papers.models.review_questions import PaperReviewQuestion
-from indico.modules.events.papers.models.reviews import PaperReview, PaperReviewType
-from indico.modules.events.papers.models.revisions import PaperRevision
+from indico.modules.events.papers.models.review_ratings import PaperReviewRating
+from indico.modules.events.papers.models.reviews import PaperReview, PaperReviewType, PaperAction
+from indico.modules.events.papers.models.revisions import PaperRevision, PaperRevisionState
 from indico.modules.events.papers.models.templates import PaperTemplate
 from indico.modules.events.papers.settings import paper_reviewing_settings, PaperReviewingRole
 from indico.modules.events.paper_reviewing.models.roles import PaperReviewingRoleType
+from indico.modules.events.paper_reviewing.models.papers import LegacyPaperFile
```
nop33: Should be above the previous line

OmeGak:
> Revisions that were judged before being submitted (!!!) are just marked as judged

Is this just a timestamp problem?

indico_zodbimport/modules/event_papers.py
```diff
@@ -182,17 +306,80 @@ def _migrate_papers(self):
...
+                            self.file_checksums[checksum] = revision
+                    except (RuntimeError, StorageError):
+                        self.importer.print_error(cformat("%{red!}File not accessible; can't CRC it [{}]")
+                                                  .format(lpf.filename), event_id=self.event.id)
+
+                    if revision:
+                        revision.files.append(_paper_file_from_legacy(lpf))
+                    n += 1
+
+                # if a revision has no files (because they've all been ignored), then keep around a copy of each
```
OmeGak: Why do we need this?

pferreir: To avoid that we have revisions with zero files. Given the bad validation in past versions, it might happen that we have the same file in 3 different revisions (and nothing else). The code CRC-based checking mechanism would only preserve it in the first one.

indico/modules/events/papers/models/reviews.py
```diff
@@ -119,7 +119,7 @@ def locator(self):
 
     @return_ascii
     def __repr__(self):
-        return format_repr(self, 'id', 'revision', 'user_id', proposed_action=None)
+        return format_repr(self, 'id', 'type', 'revision', 'user_id', proposed_action=None)
```
ThiefMaster: Unrelated, but I would not include revision but revision_id here. Triggering lazyloads in the repr can easily cause errors that make debugging things harder.

indico_zodbimport/modules/event_papers.py
```diff
@@ -55,21 +83,45 @@ def _translate_notif_options(pr, options):
     return {PaperReviewingRole[role] for role, (attr, default) in options.viewitems() if getattr(pr, attr, default)}
 
 
+def _review_color(review, text):
+    return '%{{{}}}{}%{{reset}}'.format({
```
ThiefMaster: beautiful :D

indico_zodbimport/modules/event_papers.py
```diff
@@ -55,21 +83,45 @@ def _translate_notif_options(pr, options):
...
         self.event = event
+        self.janitor_user = janitor_user
         self.pr = conf._confPaperReview
+        self.legacy_question_map = {}
+        self.legacy_contrib_revision_map = {}
+        self.legacy_contrib_last_revision_map = dict()
 
     def run(self):
         self.importer.print_success(cformat('%{blue!}{}').format(self.event), event_id=self.event.id)
+        self._migrate_feature
```
ThiefMaster: :monkey: missing `()`

indico_zodbimport/modules/event_papers.py
```diff
@@ -169,6 +223,87 @@ def _migrate_paper_roles(self, contribution):
             elif role.role == PaperReviewingRoleType.editor:
                 contribution.paper_layout_reviewers.add(role.user)
 
+    def _migrate_review(self, contribution, old_judgment, review_type):
+        # Consider legacy custom states the same as "to be corrected"
+        proposed_action = JUDGMENT_STATE_PAPER_ACTION_MAP.get(int(old_judgment._judgement._id),
+                                                              PaperAction.to_be_corrected)
+        review = PaperReview(user=old_judgment._author.user, comment=old_judgment._comments, type=review_type,
```
ThiefMaster: `convert_to_unicode()` all the string values; this also takes care of stripping and a removing garbage (control chars).
Probably it's needed in a few other places too

indico_zodbimport/modules/event_papers.py
```diff
@@ -169,6 +223,87 @@ def _migrate_paper_roles(self, contribution):
...
+                             proposed_action=proposed_action, created_dt=_to_utc(old_judgment._submissionDate))
+        for old_answer in old_judgment._answers:
+            old_question = old_answer._question
+            question = self.legacy_question_map.get(old_question, None)
```
ThiefMaster: `None` is already the default and it looks like a good case to actually catch the `KeyError` from `[]` instead

indico_zodbimport/modules/event_papers.py
```diff
@@ -169,6 +223,87 @@ def _migrate_paper_roles(self, contribution):
...
+            if not question:
+                self.importer.print_warning(cformat('%{yellow!}Answer to deleted question {} has been ignored [{}, {}]')
+                                            .format(old_question._id, contribution, review_type),
+                                            event_id=self.event.id)
+                continue
+
+            rating = PaperReviewRating(question=question, value=old_answer._rbValue)
```
ThiefMaster: We don't have any garbage outside-scale data here? (we shouldn't since the scale was never configurable)

indico_zodbimport/modules/event_papers.py
```diff
@@ -169,6 +223,87 @@ def _migrate_paper_roles(self, contribution):
...
+            state = (JUDGMENT_STATE_REVISION_MAP.get(int(old_judgment._judgement._id),
+                                                     PaperRevisionState.to_be_corrected)
+                     if old_judgment._submitted
+                     else PaperRevisionState.submitted)
+            judge = old_judgment._author.user if old_judgment._submitted else None
+            judgment_dt = _to_utc(old_judgment._submissionDate) if old_judgment._submitted else None
+            last_file = (LegacyPaperFile.query.with_parent(contribution)
+                         .filter(LegacyPaperFile.revision_id == old_revision._version)
+                         .order_by(LegacyPaperFile.created_dt.desc())
+                         .limit(1)).first()
```
ThiefMaster: .limit(1)
    .first())

That's how we formatted it everythere else. Mostly a matter of preference though

indico_zodbimport/modules/event_papers.py
```diff
@@ -169,6 +223,87 @@ def _migrate_paper_roles(self, contribution):
...
+                    revision.reviews.append(review)
+                    review_colors += _review_color(review, 'C')
+            if old_layout_review._submitted:
+                review = self._migrate_review(contribution, old_layout_review, PaperReviewType.layout)
+                revision.reviews.append(review)
+                review_colors += _review_color(review, 'L')
+            contribution.paper_revisions.append(revision)
+
+            self.importer.print_info(cformat('\tRevision %{{blue!}}{}%{{reset}} %{{white,{}}}  %{{reset}} {}'.format(
+                n, STATE_COLOR_MAP[state], review_colors)))
```
ThiefMaster: I'd `cformat(...).format(n)` (even though it doesn't make a real difference here since it's just a number which won't ever interfere with the cformat logic)

indico_zodbimport/modules/event_papers.py
pferreir: Not easy to do it any other way 😉

indico_zodbimport/modules/event_papers.py
pferreir: oh, crap. it didn't crash, of course :)

indico_zodbimport/modules/event_papers.py
ThiefMaster: I usually go for `return '%%{%s}{}%%{reset}' % (...)` - but it's only slightly less ugly

indico_zodbimport/modules/event_papers.py
pferreir: Good question. I can double-check, but shouldn't be the case.

indico_zodbimport/modules/event_papers.py
pferreir: I can't do that, since I need to apply `.format()` before `cformat()` (custom color).

indico_zodbimport/modules/event_papers.py
pferreir: Added an assert, it seems to be fine. Also noticed I was just saving the stored value (0-6) instead of using `-3 ... 3` as it is displayed to the user.

indico_zodbimport/modules/event_papers.py
OmeGak: It's at least easier to parse with your eyes.

pferreir: I don't like using `%`

PR #2775, opened 2017-01-05
Include all primary authors in MARCXML export

nop33:
Fixes ticket `RQF0683434`

### DIFF
--- a/indico/MaKaC/common/output.py
+++ b/indico/MaKaC/common/output.py
@@ -15,6 +15,7 @@
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
 import os
+from collections import defaultdict
 from datetime import datetime
 from flask import request
 from hashlib import md5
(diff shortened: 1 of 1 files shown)


indico/MaKaC/common/output.py
```diff
@@ -494,40 +494,30 @@ def _generate_contrib_location_and_time(self, contrib, out):
...
-                    list_[user].append('Author')
+        users = {}
+        primary_authors = list(contrib.primary_authors)
+        secondary_authors = list(contrib.secondary_authors)
+        for primary_author in primary_authors:
+            users[primary_author] = ['Primary Author']
+
+        for user in secondary_authors:
+            if user in users:
+                users[user].append('Author')
```
ThiefMaster: why not make users a `defaultdict(list)`?

indico/MaKaC/common/output.py
```diff
@@ -494,40 +494,30 @@ def _generate_contrib_location_and_time(self, contrib, out):
...
-            out.writeTag('subfield', u'{} {}'.format(auth.last_name, auth.first_name), [['code', 'a']])
-            for val in list_[user]:
-                out.writeTag('subfield', val, [['code', 'e']])
-            out.writeTag('subfield', auth.affiliation, [['code', 'u']])
-            out.closeTag('datafield')
-            del list_[auth]
-
-        for user in list_.keys():
-            out.openTag('datafield', [['tag', '700'], ['ind1', ' '], ['ind2', ' ']])
+        for user in users.keys():
```
ThiefMaster: you get the same by iterating over `users` without `.keys()` - also, I'd use iteritems instead of accessing `users[user]` later

indico/MaKaC/common/output.py
```diff
@@ -494,41 +495,25 @@ def _generate_contrib_location_and_time(self, contrib, out):
...
-        if auth:
-            auth = auth[0]
-            list_[auth] = ['Primary Author']
-        for user in aList:
-            if user in list_:
-                if user != auth:
-                    list_[user].append('Author')
-            else:
-                list_[user] = ['Author']
+        users = defaultdict(list)
```
OmeGak: I would use `persons` for clarity. We are dealing with `User` not with `PersonLink` here.

nop33: Well, I won't bother changing it now, it's code that will be rewritten.

indico/MaKaC/common/output.py
```diff
@@ -494,41 +495,25 @@ def _generate_contrib_location_and_time(self, contrib, out):
...
+        primary_authors = list(contrib.primary_authors)
```
ThiefMaster: why is that needed? you never modify it

indico/MaKaC/common/output.py
```diff
@@ -494,41 +495,25 @@ def _generate_contrib_location_and_time(self, contrib, out):
...
+        secondary_authors = list(contrib.secondary_authors)
```
ThiefMaster: same here

PR #2776, opened 2017-01-06
Paper reviewing settings

nop33:
Don't forget:

### DIFF
--- a/indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
+++ b/indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
@@ -1,0 +1,72 @@
+/* This file is part of Indico.
+ * Copyright (C) 2002 - 2017 European Organization for Nuclear Research (CERN).
+ *
+ * Indico is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Indico is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+(function($) {
+    'use strict';
+
+    $.widget('indico.paperemailsettingswidget', {
+        options: {
+            containerElement: null,
+            multipleRecipientOptions: [
+                'notify_on_added_to_event', 'notify_on_assigned_contrib', 'notify_on_paper_submission'
+            ],
+            singleRecipientOptions: [
+                'notify_judge_on_review', 'notify_author_on_judgment'
+            ]
+        },
+
+        _initCheckboxes: function(data) {
+            var self = this;
+            var elementID = self.element.prop('id');
+            var multipleRecipientOptions = self.options.multipleRecipientOptions;
+            var singleRecipientOptions = self.options.singleRecipientOptions;
+
+            multipleRecipientOptions.forEach(function(condition) {
+                data[condition].forEach(function(role) {
+                    $('#{0}-{1}-{2}'.format(elementID, condition, role)).prop('checked', true);
+                });
+            });
+            singleRecipientOptions.forEach(function(condition) {
+                $('#{0}-{1}'.format(elementID, condition)).prop('checked', data[condition]);
+            });
+        },
+
+        _create: function() {
+            var self = this;
+            var element = self.element;
+            var $container = self.options.containerElement;
+            var hiddenData = element.val() ? JSON.parse(element.val()) : {};
+
+            self._initCheckboxes(hiddenData);
+
+            $container.find('.multiple-recipients input').on('change', function() {
+                var setting = hiddenData[this.name];
+                if (this.checked) {
+                    setting.push(this.value);
+                } else {
+                    setting.splice(setting.indexOf(this.value), 1);
+                }
+                element.val(JSON.stringify(hiddenData));
+            });
+
+            $container.find('.single-recipient input').on('change', function() {
+                hiddenData[this.name] = this.checked;
+                element.val(JSON.stringify(hiddenData));
+            });
+        }
+    });
+})(jQuery);
--- a/indico/modules/events/papers/controllers/management.py
+++ b/indico/modules/events/papers/controllers/management.py
@@ -19,9 +19,10 @@ from __future__ import unicode_literals
 from flask import request, render_template, flash, session
 
 from indico.modules.events.contributions import Contribution
+from indico.modules.events.papers import logger
 from indico.modules.events.papers.controllers.base import RHManagePapersBase
 from indico.modules.events.papers.forms import (BulkPaperJudgmentForm, make_competences_form, PapersScheduleForm,
-                                                PaperTeamsForm)
+                                                PaperTeamsForm, PaperReviewingSettingsForm)
 from indico.modules.events.papers.lists import PaperAssignmentListGenerator
 from indico.modules.events.papers.models.revisions import PaperRevisionState
 from indico.modules.events.papers.operations import (set_reviewing_state, update_team_members, create_competences,
@@ -226,3 +227,25 @@ class RHBulkPaperJudgment(RHManagePapersActionsBase):
                                num_not_submitted_papers).format(num=num_not_submitted_papers), 'warning')
             return jsonify_data(**self.list_generator.render_list())
         return jsonify_form(form=form, submit=_('Judge'), disabled_until_change=False)
+
+
+class RHManageReviewingSettings(RHManagePapersBase):
+    def _process(self):
+        defaults = FormDefaults(content_review_questions=self.event_new.cfp.content_review_questions,
+                                layout_review_questions=self.event_new.cfp.layout_review_questions)
+        form = PaperReviewingSettingsForm(event=self.event_new, obj=defaults)
+        if form.validate_on_submit():
+            data = form.data
+            content_review_questions = data.pop('content_review_questions', None)
+            layout_review_questions = data.pop('layout_review_questions', None)
+            if content_review_questions is None:
+                content_review_questions = self.event_new.cfp.content_review_questions
+            if layout_review_questions is None:
+                layout_review_questions = self.event_new.cfp.layout_review_questions
+            self.event_new.paper_review_questions = content_review_questions + layout_review_questions
+            paper_reviewing_settings.set_multi(self.event_new, data)
+            flash(_("The reviewing settings were saved successfully"), 'success')
+            logger.info("Paper reviewing settings of %r updated by %r", self.event_new, session.user)
+            return jsonify_data()
+        self.commit = False
+        return jsonify_form(form)
--- a/indico/modules/events/papers/controllers/management.py
+++ b/indico/modules/events/papers/controllers/management.py
@@ -232,7 +232,8 @@ class RHBulkPaperJudgment(RHManagePapersActionsBase):
 class RHManageReviewingSettings(RHManagePapersBase):
     def _process(self):
         defaults = FormDefaults(content_review_questions=self.event_new.cfp.content_review_questions,
-                                layout_review_questions=self.event_new.cfp.layout_review_questions)
+                                layout_review_questions=self.event_new.cfp.layout_review_questions,
+                                **paper_reviewing_settings.get_all(self.event_new))
         form = PaperReviewingSettingsForm(event=self.event_new, obj=defaults)
         if form.validate_on_submit():
             data = form.data
--- a/indico/modules/events/papers/controllers/management.py
+++ b/indico/modules/events/papers/controllers/management.py
@@ -27,7 +27,7 @@ from indico.modules.events.papers.lists import PaperAssignmentListGenerator
 from indico.modules.events.papers.models.revisions import PaperRevisionState
 from indico.modules.events.papers.operations import (set_reviewing_state, update_team_members, create_competences,
                                                      update_competences, judge_paper)
-from indico.modules.events.papers.settings import paper_reviewing_settings
+from indico.modules.events.papers.settings import paper_reviewing_settings, RoleConverter
 from indico.modules.events.papers.views import WPManagePapers
 from indico.modules.events.papers.operations import schedule_cfp, open_cfp, close_cfp
 from indico.modules.users.models.users import User
@@ -244,6 +244,10 @@ class RHManageReviewingSettings(RHManagePapersBase):
             if layout_review_questions is None:
                 layout_review_questions = self.event_new.cfp.layout_review_questions
             self.event_new.paper_review_questions = content_review_questions + layout_review_questions
+
+            email_settings = data.pop('email_settings')
+            data.update(email_settings)
+
             paper_reviewing_settings.set_multi(self.event_new, data)
             flash(_("The reviewing settings were saved successfully"), 'success')
             logger.info("Paper reviewing settings of %r updated by %r", self.event_new, session.user)
--- a/indico/modules/events/papers/fields.py
+++ b/indico/modules/events/papers/fields.py
@@ -1,0 +1,49 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2017 European Organization for Nuclear Research (CERN).
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 3 of the
+# License, or (at your option) any later version.
+#
+# Indico is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+import json
+
+from indico.modules.events.papers.settings import paper_reviewing_settings as settings, RoleConverter
+from indico.web.forms.fields import JSONField
+from indico.web.forms.widgets import JinjaWidget
+
+
+class PaperEmailSettingsField(JSONField):
+    CAN_POPULATE = True
+    widget = JinjaWidget('events/papers/forms/paper_email_settings_widget.html')
+
+    @property
+    def event(self):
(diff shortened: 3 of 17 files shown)


indico/modules/events/papers/controllers/management.py
```diff
@@ -111,3 +114,28 @@ def _process(self):
...
+    def _process(self):
+        defaults = FormDefaults(content_review_questions=self.event_new.cfp.content_review_questions,
+                                layout_review_questions=self.event_new.cfp.layout_review_questions,
+                                **paper_reviewing_settings.get_all(self.event_new))
+        form = PaperReviewingSettingsForm(event=self.event_new, obj=defaults)
+        if form.validate_on_submit():
+            data = form.data
+            content_review_questions = data.pop('content_review_questions', [])
+            layout_review_questions = data.pop('layout_review_questions', [])
+            self.event_new.paper_review_questions = []
```
ThiefMaster: That's not how this works ;)

By doing this, SQLAlchemy may delete the existing questions, and thus the associated answers (or try to delete them and fail if no cascading is set up). Either way, it's not what you want.

I see you already added a field similar to the one I added in the abstract dialog, but you need to use logic similar to what I did in `RHManageAbstractReviewing` too (do the assignment to deal with new questions and also - very important - don't commit in case the form wasn't submitted successfully to avoid committing editions/deletions if form validation failed somewhere else).

indico/modules/events/papers/fields.py
```diff
@@ -0,0 +1,47 @@
...
+                                                        **kwargs)
+
+    def process_formdata(self, valuelist):
+        super(PaperReviewQuestionsField, self).process_formdata(valuelist)
+        if valuelist:
+            existing = {x.id: x for x in self.object_data or ()}
+            data = []
+            for pos, entry in enumerate(self.data, 1):
+                question = (existing.pop(int(entry['id']))
+                            if entry.get('id') is not None else PaperReviewQuestionsField())
```
ThiefMaster: wrong class name. you want the model here, not a wtforms field

indico/modules/events/papers/fields.py
```diff
@@ -0,0 +1,47 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from indico.util.i18n import _
+from indico.web.forms.fields import MultipleItemsField
+
+
+class PaperReviewQuestionsField(MultipleItemsField):
```
ThiefMaster: This is almost identical to `AbstractReviewQuestionsField` - let's try to use a common class that handles both or a base class and then a subclass that adds the `no_score` field since the only differences seem to be the name of the question model and the (non-)existence of `no_score`.

One possible way of unifying it in a single class could be a constructor like this:

    ... = ReviewQuestionsField(_('whatever'), question_model=lambda: PaperReviewQuestion(type=PaperReviewType.xxx))

and for abstracts

    ... = ReviewQuestionsField(_('whatever'), question_model=AbstractReviewQuestion,
                               extra_fields=[{'id': 'no_score', 'caption': _("Exclude from score"),
                                              'type': 'checkbox'}])

indico/modules/events/papers/controllers/management.py
ThiefMaster: Right now, saving this form will delete all questions for a review type if it has been disabled.

indico/modules/events/papers/forms.py
```diff
@@ -49,3 +51,21 @@ def __init__(self, *args, **kwargs):
...
             del self.layout_reviewers
+
+
+class PaperReviewingSettingsForm(IndicoForm):
+    content_review_questions = PaperReviewQuestionsField(_("Content review questions"))
+    layout_review_questions = PaperReviewQuestionsField(_("Layout review questions"))
+    announcement = IndicoMarkdownField(_('Announcement'), editor=True)
+
+    @generated_data
+    def announcement_render_mode(self):
```
ThiefMaster: Did we have this announcement in legacy? If not, we don't need to store a render mode since it's always markdown. If we did, add the `handle_legacy_description` call like in `RHManageAbstractSubmission`. That way a description that has a HTML render mode (the ZODB migration script would have to set that) will be converted automatically when edited, combined with a warning asking the user to double-check that it still looks fine)

indico/modules/events/papers/views.py
```diff
@@ -16,22 +16,28 @@
 
 from __future__ import unicode_literals
 
+from indico.modules.events.abstracts.views import _MathjaxMixin
```
ThiefMaster: should be moved to a more suitable place than `abstracts`, and probably un-underscored

nop33: How about `events/util.py`?

indico/modules/events/papers/forms.py
nop33: Ah I see. No, we didn't have it in legacy. Thanks!

indico/modules/events/papers/views.py
nop33: I was about to move it, but then I saw that we import it in the tracks module as well.

ThiefMaster: `indico/util/mathjax.py`? it's not directly event-related

indico/modules/events/papers/forms.py
ThiefMaster: Then we can get rid of the render_mode and always use `MarkdownText(..)` in the `announcement` property in the `CallForPapers`

indico/modules/events/papers/views.py
nop33: hm ok

nop33: Would you also move the `util.py` from `events/util.py` inside the `events/util/` folder? And maybe rename it to `util/base.py`?

ThiefMaster: which events/util folder? also, not a fan of a base.py that's not for base classes

nop33: ah sorry, misread sth above

indico/modules/events/papers/fields.py
nop33: Seems like we are moving to what @OmeGak was suggesting in the beginning: a `review` module :boom:

ThiefMaster: just for rather generic stuff ;)

indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
```diff
@@ -0,0 +1,70 @@
...
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+(function($) {
+    'use strict';
+
+    $.widget('indico.paperemailsettingswidget', {
+        options: {
+            containerElement: null,
+            multipleRecipientOptions: ['notify_on_added_to_event', 'notify_on_assigned_contrib',
```
ThiefMaster: I don't think this wrapping style is common in JS. I'd go for what we do in other places if it doesn't fit on oneline:
```
foo: [
    ..., ..., ...,
    ...
]
```

indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
```diff
@@ -0,0 +1,70 @@
...
+                                       'notify_on_paper_submission'],
+            singleRecipientOptions: ['notify_judge_on_review', 'notify_author_on_judgment']
+        },
+
+        _setUp: function(data) {
```
ThiefMaster: not sure about that function name (besides disliking the `setUp` spelling of `setup` it sounds almost like a method that's part of the jquery ui widget system to me)

indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
```diff
@@ -0,0 +1,70 @@
...
+            var self = this;
+            var multipleRecipientOptions = self.options.multipleRecipientOptions;
+            var singleRecipientOptions = self.options.singleRecipientOptions;
+
+            for (var i = 0; i < multipleRecipientOptions.length; i++) {
```
ThiefMaster: I'd move the `var i;` out since you use it for multiple loops in this function. Easier to see that it's not missing in the next loop. Same for `var key;`

indico/util/mathjax.py
```diff
@@ -1,5 +1,5 @@
 # This file is part of Indico.
-# Copyright (C) 2002 - 2016 European Organization for Nuclear Research (CERN).
```
ThiefMaster: I'd amend the commit where you add the file instead of adding a new commit for it

indico/modules/events/papers/fields.py
```diff
@@ -0,0 +1,42 @@
...
+from indico.modules.events.papers.settings import paper_reviewing_settings as settings
+from indico.web.forms.fields import JSONField
+from indico.web.forms.widgets import JinjaWidget
+
+
+class PaperEmailSettingsField(JSONField):
+    CAN_POPULATE = True
+    widget = JinjaWidget('events/papers/forms/paper_email_settings_widget.html')
+
+    def __init__(self, *args, **kwargs):
```
ThiefMaster: not used/needed

indico/modules/events/papers/fields.py
```diff
@@ -0,0 +1,42 @@
...
+#
+# Indico is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+# from __future__ import unicode_literals
```
ThiefMaster: that shouldn't be commented out

indico/modules/events/papers/controllers/management.py
```diff
@@ -226,3 +228,34 @@ def _process(self):
...
+
+class RHManageReviewingSettings(RHManagePapersBase):
+    def _process(self):
+        defaults = FormDefaults(content_review_questions=self.event_new.cfp.content_review_questions,
+                                layout_review_questions=self.event_new.cfp.layout_review_questions,
+                                **paper_reviewing_settings.get_all(self.event_new))
+        form = PaperReviewingSettingsForm(event=self.event_new, obj=defaults)
+        if form.validate_on_submit():
+            data = form.data
+            content_review_questions = data.pop('content_review_questions', [])
```
ThiefMaster: The whole `*_review_questions` logic here is still not needed. See `RHManageAbstractReviewing`

indico/modules/events/papers/controllers/management.py
```diff
@@ -226,3 +228,34 @@ def _process(self):
...
+                                        event_new=self.event_new)
+                paper_review_questions.append(q)
+            self.event_new.paper_review_questions = paper_review_questions
+
+            for key, value in data.pop('email_settings').iteritems():
+                data[key] = RoleConverter.to_python(value) if isinstance(value, list) else value
+
+            paper_reviewing_settings.set_multi(self.event_new, data)
+            flash(_("The reviewing settings were saved successfully"), 'success')
+            return jsonify_data()
```
ThiefMaster: no logging?

indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
nop33: I stole it from the `track_role_widget.js`. I'll rename it to `_initilize` or sth.

ThiefMaster: don't use `_init` though, that one *is* used by the UI core (almost like _create but also runs when calling the constructor again on an element that already has the widget).

maybe `_initFields` or `_populateFields`?

indico/modules/events/papers/controllers/management.py
nop33: Well, we don't have logging in the abstract reviewing settings either.

indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
```diff
@@ -0,0 +1,75 @@
...
+            singleRecipientOptions: [
+                'notify_judge_on_review', 'notify_author_on_judgment'
+            ]
+        },
+
+        initCheckboxes: function(data) {
+            var self = this;
+            var multipleRecipientOptions = self.options.multipleRecipientOptions;
+            var singleRecipientOptions = self.options.singleRecipientOptions;
+            var i = 0;
```
ThiefMaster: no need to init

indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
```diff
@@ -0,0 +1,75 @@
...
+            var key = null;
```
ThiefMaster: same

indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
```diff
@@ -0,0 +1,75 @@
...
+
+            for (i = 0; i < multipleRecipientOptions.length; i++) {
```
ThiefMaster: What about this? More readable IMO

```
multipleRecipientOptions.forEach(function(key) {
    data[key].forEach(function(whatever) {
        $('#{0}-{1}'.format(key, whatever).prop('checked', true);
    });
});
```

indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
```diff
@@ -0,0 +1,75 @@
...
+            containerElement: null,
+            multipleRecipientOptions: [
+                'notify_on_added_to_event', 'notify_on_assigned_contrib', 'notify_on_paper_submission'
+            ],
+            singleRecipientOptions: [
+                'notify_judge_on_review', 'notify_author_on_judgment'
+            ]
+        },
+
+        initCheckboxes: function(data) {
```
ThiefMaster: can be underscored if not called from outside

indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
```diff
@@ -0,0 +1,75 @@
...
+            var self = this;
```
ThiefMaster: not really needed

indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
```diff
@@ -0,0 +1,75 @@
...
+                }
+            }
+            for (i = 0; i < singleRecipientOptions.length; i++) {
+                key = singleRecipientOptions[i];
+                $('#' + key).prop('checked', data[key]);
+            }
+        },
+
+        _create: function() {
+            var self = this;
```
ThiefMaster: not really needed, you only use it once to call `initCheckboxes` but never inside a callback

indico/modules/events/papers/controllers/management.py
```diff
@@ -226,3 +227,26 @@ def _process(self):
...
+                                **paper_reviewing_settings.get_all(self.event_new))
+        form = PaperReviewingSettingsForm(event=self.event_new, obj=defaults)
+        if form.validate_on_submit():
+            data = form.data
+            content_review_questions = data.pop('content_review_questions', [])
+            layout_review_questions = data.pop('layout_review_questions', [])
+            self.event_new.paper_review_questions = content_review_questions + layout_review_questions
+
+            for key, value in data.pop('email_settings').iteritems():
+                data[key] = RoleConverter.to_python(value) if isinstance(value, list) else value
```
ThiefMaster: Why is this needed?

indico/modules/events/papers/templates/forms/paper_email_settings_widget.html
```diff
@@ -0,0 +1,118 @@
...
+                        <th>{% trans %}Layout reviewer{% endtrans %}</th>
+                    {% endif %}
+                    <th>{% trans %}Author{% endtrans %}</th>
+                </tr>
+            </thead>
+            <tbody>
+                <tr class="multiple-recipients">
+                    <td>{% trans %}the person is added to/removed from the conference{% endtrans %}</td>
+                    <td>
+                        <input type="checkbox" name="notify_on_added_to_event" id="notify_on_added_to_event-judge"
```
OmeGak: I don't think we are using the `id` for anything. I would leave it out if that's the case. Also, inconsistent underscore/hyphen. There are some more instances of the same in this file.

indico/modules/events/papers/controllers/management.py
nop33: cause the value can be a list which needs to be converted to a set. Instead of reimplementing the logic of the `RoleConverter`, I just used it.

ThiefMaster: Wouldn't it make more sense to do that in the field class? You also pass a set to the form defaults so it'd be more consistent that the field also returns a set

indico/modules/events/papers/controllers/management.py
```diff
@@ -226,3 +227,26 @@ def _process(self):
...
+
+            paper_reviewing_settings.set_multi(self.event_new, data)
+            flash(_("The reviewing settings were saved successfully"), 'success')
+            logger.info("Paper reviewing settings of %s have been updated by %s", self.event_new, session.user)
```
ThiefMaster: `%r`, and for the sake of conciseness I'd skip the "have been" - it goes to a log file after all

indico/htdocs/js/indico/jquery/paper_email_settings_widget.js
nop33: In all our jQuery widgets we follow the same principle, I wouldn't change it here.

ThiefMaster: then `var $container = this.options.containerElement;` should probably be changed to use self ;)

indico/modules/events/papers/templates/forms/paper_email_settings_widget.html
ThiefMaster: I think `notify_on_added_to_event` is supposed to match the name of the field. Seems like a valid case to not follow the usual "dashed" convention. I believe it's used here: https://github.com/indico/indico/pull/2776/files#diff-fe7a46e1cd6f40ed1d66c84b01005e36R42

ThiefMaster: However, something that *is* missing here is the field ID. Even though we'll never have two instances of this wtforms field/widget on the same page, I think it's cleaner to have IDs that are guaranteed to be unique nonetheless

OmeGak: That's pretty obscure and I bet a hell to debug in the future.

nop33: @ThiefMaster you mean I should combine the `{{ field.id }}` with this ID to make sure it's unique?

ThiefMaster: yes

PR #2782, opened 2017-01-12
Call for Papers display page

DavidAndreev:


indico/modules/events/papers/__init__.py
```diff
@@ -90,3 +91,11 @@ class PaperLayoutReviewerRole(ManagementRole):
     name = 'paper_layout_reviewer'
     friendly_name = _('Layout reviewer')
     description = _('Grants layout reviewing rights for assigned papers.')
+
+
+@signals.event.sidemenu.connect
+def _extend_event_menu(sender, **kwargs):
+
```
ThiefMaster: too much whitespace

ThiefMaster: too much whitespace

indico/modules/events/papers/__init__.py
```diff
@@ -90,3 +91,11 @@ class PaperLayoutReviewerRole(ManagementRole):
...
+    yield MenuEntryData(title=_("Call for Papers"), name='call_for_papers',
+                        endpoint='papers.call_for_papers', position=2,
```
ThiefMaster: aren't we already using `position=2` for another menu entry? we could either add it to the end (need to check what the highest position is) or add it as `position=3` and increase all other positions >=3 by one. for now I'd go for `position=-1` (that adds it at the end) and check with @pferreir later if it's better to keep it at the end or insert it after CfA

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -51,6 +52,10 @@ def has_ended(self):
     def is_open(self):
         return self.has_started and not self.has_ended
 
+    def can_submit_papers(self, user):
```
ThiefMaster: Probably not needed; it's only used to implement the outside-deadlines-submission ACL in abstracts but we don't have this in papers (at least not yet and not in legacy)

indico/modules/events/papers/templates/display/_paper_list.html
```diff
@@ -0,0 +1,63 @@
...
+        <div class="display-contribution-list">
+            <ul>
+                {% for contribution in contributions %}
+                    {% set latest_revision = contribution.paper_last_revision %}
+                    <div class="contribution-row">
+                        <li>
+                            <div class="row-title">
+                                <a data-searchable="{{ contribution.title|lower }}">
+{#                                    TODO: Link to the paper timeline page #}
+                                    {#                                <a href="{{ url_for('abstracts.display_reviewable_track_abstracts', track) }}">#}
```
ThiefMaster: i know it's going to change but the indentation here is really weird

indico/modules/events/papers/templates/display/_paper_list.html
```diff
@@ -0,0 +1,63 @@
...
+                                                </div>
+                                            {% endif %}
+                                        </div>
+                                    {% endif %}
+                                    <div class="label i-tag {{ latest_revision.state.name }}">
+                                        {{ latest_revision.state.title }}
+                                    </div>
+                                </div>
+                                <div class="row-right">
+{#                                    TODO: List attached files #}
```
ThiefMaster: same here

indico/modules/events/papers/templates/display/call_for_papers.html
```diff
@@ -0,0 +1,86 @@
...
+                <span>{{ alternative_text }}</span>
+            {% endif %}
+        </div>
+    </li>
+{% endmacro %}
+
+{% macro render_timeline_section(cfp) %}
+    {% set tz = cfp.event.display_tzinfo %}
+    <section>
+        <ul class="cfa-timeline flexrow f-a-center f-j-center">
```
ThiefMaster: `cfa-timeline`?!

indico/modules/events/papers/util.py
```diff
@@ -0,0 +1,24 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2016 European Organization for Nuclear Research (CERN).
```
ThiefMaster: 2017

indico/modules/events/papers/controllers/display.py
```diff
@@ -0,0 +1,30 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2016 European Organization for Nuclear Research (CERN).
```
ThiefMaster: 2017

indico/modules/events/papers/util.py
```diff
@@ -0,0 +1,24 @@
...
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from indico.modules.events.contributions import Contribution
+from indico.modules.events.papers.models.revisions import PaperRevision
+
+
+def get_contributions_with_paper_submitted_by_user(event, user):
```
ThiefMaster: Please make sure this query is not awfully slow for very big events with lots of contributions (e.g. CHEP2016)!

indico/modules/events/papers/util.py
```diff
@@ -0,0 +1,24 @@
...
+    query = Contribution.query.with_parent(event)
+    query = query.filter(Contribution.paper_revisions.any(PaperRevision.submitter == user))
```
ThiefMaster: I'd use the same code style we use for other queries spanning multiple lines:

```python
return (Contribution.query.with_parent(event)
        .filter(Contribution.paper_revisions.any(PaperRevision.submitter == user))
        .all())
```

indico/modules/events/papers/views.py
```diff
@@ -32,7 +32,13 @@ class WPDisplayPapersBase(WPJinjaMixin, WPConferenceDefaultDisplayBase):
     template_prefix = 'events/papers/'
 
     def getJSFiles(self):
-        return WPConferenceDefaultDisplayBase.getJSFiles(self) + self._asset_env['modules_papers_js'].urls()
+        return (WPConferenceDefaultDisplayBase.getJSFiles(self) +
+                self._asset_env['modules_papers_js'].urls() +
+                self._asset_env['modules_event_management_js'].urls())
```
ThiefMaster: I'd include papers after event management - event management is more generic than papers

indico/modules/events/papers/templates/display/call_for_papers.html
DavidAndreev: can't I use the same css since its the very same thing (deadline timeline) as in CFA ?

ThiefMaster: what I'm saying is that this should probably be renamed and moved (currently in _cfa_page.scss) when it's used here. cc @OmeGak

indico/modules/events/papers/util.py
DavidAndreev: Yeah I did, I always code with CHEP as an example

indico/htdocs/sass/modules/_papers.scss
```diff
@@ -41,3 +43,81 @@
         }
     }
 }
+
+.display-contribution-list {
+    @extend %font-family-modern-body;
+    .contribution-row {
```
nop33: Thou shalt leave a blank line between selectors: https://github.com/indico/indico/wiki/SASS-Guidelines
Apply also below.

indico/htdocs/sass/modules/_papers.scss
```diff
@@ -41,3 +43,81 @@
...
+            border: 0;
+            margin-top: 7px;
+            display: inline-block;
+            padding: 5px 7px;
+
+            &.review-role {
+                margin-right: 10px;
+                background-color: white;
+                color: $light-black;
+                border: 1px solid $gray;
```
nop33: I think this can be done by `@include border-all($gray)` (or `@include border-all()` if `$pastel-gray` suits you).

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -36,6 +36,7 @@ def __repr__(self):
 
     start_dt = EventSettingProperty(paper_reviewing_settings, 'start_dt')
     end_dt = EventSettingProperty(paper_reviewing_settings, 'end_dt')
+    announcement = EventSettingProperty(paper_reviewing_settings, 'announcement')
```
nop33: I am also adding those in my PR where I worked on the announcement settings: https://github.com/indico/indico/pull/2776/commits/5e20878f9d5d9d64793afc88ce479368f52636c8#diff-f67bb6a77e79f4fc1c19acfaef552841R120

indico/modules/events/papers/settings.py
```diff
@@ -41,6 +41,7 @@ def to_python(value):
     'start_dt': None,
     'end_dt': None,
     'enforce_deadlines': False,
+    'announcement': '',
```
nop33: Same here

indico/modules/events/papers/templates/display/call_for_papers.html
DavidAndreev: So where should I move it ?

indico/modules/events/papers/models/call_for_papers.py
DavidAndreev: So should I delete it ?

nop33: Yes. My PR is merged, so make sure you rebase (after deleting it) ;)

indico/modules/events/papers/templates/display/call_for_papers.html
nop33: How about `indico/htdocs/sass/partials/_timelines.scss`? cc @OmeGak ?

indico/htdocs/sass/modules/_papers.scss
```diff
@@ -15,7 +15,9 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
-@import base;
+@import 'base';
```
OmeGak: You don't need any of these imports anymore. You needed them before because we were not bundling module-related `scss` together with `screen.scss`, but that's not the case anymore. We now have all variables and partials available out-of-the-box in module-related `scss`.

indico/htdocs/sass/modules/_papers.scss
```diff
@@ -41,3 +43,83 @@
         }
     }
 }
+
+.display-contribution-list {
+    @extend %font-family-modern-body;
+
+    .contribution-row {
+        padding-left: 10px;
+        border-left: 3px solid $pastel-gray;
```
OmeGak:
```scss
@include border-left($width: 3px);
```

indico/htdocs/sass/modules/_papers.scss
```diff
@@ -41,3 +43,83 @@
...
+        margin-bottom: 1em;
+
+        &:hover {
+            border-color: $blue;
+        }
+        .row-title {
+            font-size: 14pt;
+            @include border-bottom();
```
OmeGak: We specify `@include` before the rest of the rules.

indico/htdocs/sass/modules/_papers.scss
```diff
@@ -41,3 +43,83 @@
...
+                margin-right: 10px;
+                background-color: white;
+                color: $light-black;
+                border: 1px solid $gray;
+                i {
+                    color: $blue;
+                }
+            }
+
+            &.submitted {
```
OmeGak: Consider extending `%semantic-background`. You would then need a state-class mapping in the template.

```jinja
{% set mapping = {'submitted': 'highlight',
                  'accepted': 'success',
                  ...} %}
```

indico/modules/events/papers/templates/display/_common.html
```diff
@@ -0,0 +1,42 @@
...
+                </div>
+                <div>
+                    {%- trans %}You can submit a paper for reviewing.{% endtrans -%}
+                </div>
+                <div class="toolbar right">
+                    {% if session.user and not redirect_first %}
+                        <a class="i-button highlight js-show-paper-form"
+                           data-title="{% trans %}Submit new papers{% endtrans %}"
+{#                           TODO: Link to submit paper #}
+{#                           data-href="{{ url_for('papers.submit', event) }}"#}
```
OmeGak: You can do that already. Check [here](https://github.com/indico/indico/blob/devel/indico/modules/events/papers/templates/_contributions.html#L28) to see how.

indico/modules/events/papers/templates/display/_paper_list.html
```diff
@@ -0,0 +1,63 @@
...
+    <section>
+        <div class="display-contribution-list">
+            <ul>
+                {% for contribution in contributions %}
+                    {% set latest_revision = contribution.paper_last_revision %}
+                    <div class="contribution-row">
+                        <li>
+                            <div class="row-title">
+                                <a data-searchable="{{ contribution.title|lower }}">
+{#                                    TODO: Link to the paper timeline page #}
```
OmeGak:
```py
url_for('.paper_timeline', contribution)
```

indico/modules/events/papers/templates/display/_common.html
DavidAndreev: hmm.. but what contribution do I pass there? cause its in CFP page and there is no contribution specified

pferreir: @omegak, it's a different box.

indico/htdocs/sass/modules/_papers.scss
nop33: the state mapping won't be necessary anymore, after my PR gets merged. You'll be able to define the css classes in the State enum.

indico/modules/events/papers/templates/display/_paper_list.html
```diff
@@ -0,0 +1,61 @@
...
+                                    </div>
+                                </div>
+                                <div class="row-right">
+{#                                    TODO: List attached files #}
+            {#                        {% for file in latest_revision.files %}#}
+            {#                            <a href="... link to the gfile...">{{ file.title }}</a><br>#}
+            {#                        {% endfor %}#}
+                                    <a>file_name.pdf</a><br>
+                                    <a>file_name.pdf</a><br>
+                                    <a>file_name.pdf</a><br>
```
nop33: The submission of the paper is already implemented in `devel` if I am not mistaken. This means you can uncomment the code and remove this dummy data.

indico/modules/events/papers/templates/display/call_for_papers.html
```diff
@@ -0,0 +1,99 @@
+{% extends 'layout/conference_page_base.html' %}
+
+{% from 'events/papers/display/_common.html' import render_cfp_infoline %}
+{% from 'events/papers/display/_paper_list.html' import render_user_papers_list %}
+{% from 'events/management/_lists.html' import render_displayed_entries_fragment %}
```
nop33: Should moved above.

indico/modules/events/papers/templates/display/_paper_list.html
DavidAndreev: Well I implemented this displaying in Reviewing area which uses the same macro to render list, so once we merge that I will reabse and apply it here too ;)

indico/htdocs/sass/modules/_papers.scss
```diff
@@ -39,3 +39,92 @@
...
+                background-color: $green;
+            }
+
+            &.rejected {
+                background-color: $red;
+            }
+
+            &.to_be_corrected {
+                background-color: $yellow;
+            }
```
nop33: I suppose the comments I left in your other PR a few seconds ago, will be applied here as well :wink:

DavidAndreev: Yes ;)

indico/modules/events/papers/templates/display/_paper_list.html
```diff
@@ -1,5 +1,6 @@
 {% from 'events/management/_lists.html' import render_displayed_entries_fragment %}
 
+<<<<<<< 83d1b01137f5b4b358766d19e1fe6281d3cfe34e
```
ThiefMaster: :monkey:

indico/modules/events/papers/templates/display/call_for_papers.html
```diff
@@ -0,0 +1,99 @@
...
+
+{% from 'events/management/_lists.html' import render_displayed_entries_fragment %}
+{% from 'events/papers/display/_common.html' import render_cfp_infoline %}
+{% from 'events/papers/display/_paper_list.html' import render_user_papers_list %}
+
+{% block title %}
+    {%- trans %}Call for Papers{% endtrans -%}
+{% endblock %}
+
+{% macro render_timestamp(label, date, tz, completed=false, alternative_text=none) %}
```
ThiefMaster: Is this copied from the cfa template? in that case it should be moved into a generic place and imported in both this and the cfa template!

DavidAndreev: Good idea. Can I put it in indico/modules/events/templates/display ? Or is there already a place where we put things common for papers and abstracts ? And what should I call the argument representing CFP or CFA that can be passed there ?

ThiefMaster: indico/modules/events/templates/reviews/...

indico/modules/events/papers/controllers/display.py
```diff
@@ -26,8 +26,10 @@
 from indico.modules.events.papers.models.reviews import PaperReviewType, PaperReview, PaperTypeProxy
 from indico.modules.events.papers.operations import (create_paper_revision, create_review, create_comment,
                                                      update_comment, delete_comment, update_review)
-from indico.modules.events.papers.util import get_user_contributions_to_review, get_user_reviewed_contributions
-from indico.modules.events.papers.views import WPDisplayPapersBase, render_paper_page, WPDisplayReviewingArea
+from indico.modules.events.papers.util import get_user_contributions_to_review, get_user_reviewed_contributions, \
```
ThiefMaster: we don't wrap imports with `\`

indico/modules/events/papers/controllers/display.py
```diff
@@ -26,8 +26,10 @@
...
+    get_contributions_with_paper_submitted_by_user
+from indico.modules.events.papers.views import WPDisplayPapersBase, render_paper_page, WPDisplayReviewingArea, \
```
ThiefMaster: same (scumbag pycharm)

indico/modules/events/papers/templates/display/call_for_papers.html
```diff
@@ -0,0 +1,75 @@
+{% extends 'layout/conference_page_base.html' %}
```
ThiefMaster: Why not inherit from 'events/papers/_base.html'? that already sets the title block

indico/modules/events/papers/templates/display/call_for_papers.html
```diff
@@ -0,0 +1,75 @@
...
+                    </div>
+                </div>
+                <ul class="template-list">
+                    {% for tpl in event.paper_templates|sort(attribute='id')|sort(attribute='name') %}
+                        <li>
+                            <a href="{{ url_for('.download_template', tpl) }}" class="attachment {{ icon_from_mimetype(tpl.content_type, 'icon-file') }}">
+                                <span>{{ tpl.name }}</span>
+                            </a>
+                            {% if tpl.description %}
+                                <a class="info icon-info" title="{{ tpl.description }}"></a>
```
ThiefMaster: why an `a` tag? `<i>` seems much more appropriate here

indico/modules/events/papers/templates/display/_common.html
```diff
@@ -0,0 +1,42 @@
+{% macro render_cfp_infoline(event, cfp, show_icon=true, redirect_first=false) %}
+    {% set can_submit_papers = cfp.is_open or session.user %}
```
ThiefMaster: This check makes no sense - being logged in is not enough to submit a paper. Since we don't have a user-specific deadline override like we do in abstracts the only thing to check here is whether CFP is open

indico/modules/events/papers/templates/display/_common.html
```diff
@@ -0,0 +1,42 @@
...
+        {% if show_icon %}
+            <span class="icon icon-file-content"></span>
+        {% endif %}
+        <div class="text">
+            {%- if can_submit_papers %}
+                <div class="label">
+                    {% if cfp.is_open %}
+                        {%- trans %}The call for papers is open{% endtrans -%}
+                    {% else %}
+                        {%- trans %}You have been authorized to submit papers{% endtrans -%}
```
ThiefMaster: see above -- not needed

indico/modules/events/papers/templates/display/call_for_papers.html
```diff
@@ -0,0 +1,75 @@
...
+                                    <input type="text" id="search-input" placeholder="{% trans %}Enter #id or search string{% endtrans %}">
+                                </div>
+                            </div>
+                        </div>
+                    </div>
+                </div>
+                {{ render_user_papers_list(event, contributions) }}
+                <div id="filter-placeholder"></div>
+            </section>
+        {% endif %}
```
ThiefMaster: We need to show a message if the user has no contributions (in case he gets to this page somehow through a direct link)

indico/modules/events/papers/templates/display/call_for_papers.html
mvidalgarcia: > And what should I call the argument representing CFP or CFA that can be passed there ?  

I think we are calling them _Call for proposals_ as stated here: https://github.com/indico/indico/blob/master/indico/modules/events/models/reviews.py#L39
Not 100% sure though.

indico/modules/events/templates/reviews/_common.html
```diff
@@ -70,3 +70,28 @@
 {% macro render_review_state(review) %}
     <span class="bold underline semantic-text {{ review.proposed_action.css_class }}">
 {% endmacro %}
+
+{% macro render_timeline_section(cf) %}
```
mvidalgarcia: > And what should I call the argument representing CFP or CFA that can be passed there ?  

I think we are calling them _Call for proposals_ (so `cfp`) as stated here: https://github.com/indico/indico/blob/master/indico/modules/events/models/reviews.py#L39
Not 100% sure though.

nop33: Be sure.

indico/htdocs/sass/modules/_papers.scss
```diff
@@ -122,6 +122,16 @@
         color: $light-black;
         padding: 20px 0 10px;
     }
+
+    .template-list {
+        font-size: 13pt;
```
ThiefMaster: We use px/em everywhere else (except for legacy). I don't think we want to start using pt in a few places. `em` is probably the best choice here.

indico/modules/events/papers/templates/display/_common.html
```diff
@@ -0,0 +1,38 @@
...
+                </div>
+                <div>
+                    {%- trans %}You can submit a paper for reviewing.{% endtrans -%}
+                </div>
+                <div class="toolbar right">
+                    {% if session.user and not redirect_first %}
+                        <a class="i-button highlight js-show-paper-form"
+                           data-title="{% trans %}Submit new paper{% endtrans %}"
+{#                           TODO: Link to submit paper #}
+{#                           data-href="{{ url_for('papers.submit', event) }}"#}
```
ThiefMaster: I think you can already use `.submit_revision` here.

indico/modules/events/papers/templates/display/call_for_papers.html
```diff
@@ -0,0 +1,80 @@
...
+            <section>
+                <div class="header">
+                    <div class="header-row">
+                        <h3>{% trans %}Paper templates{% endtrans %}</h3>
+                    </div>
+                </div>
+                <ul class="template-list">
+                    {% for tpl in event.paper_templates|sort(attribute='id')|sort(attribute='name') %}
+                        <li>
+                            <a href="{{ url_for('.download_template', tpl) }}" class="attachment {{ icon_from_mimetype(tpl.content_type, 'icon-file') }}">
```
ThiefMaster: wrap before `class`

indico/modules/events/papers/templates/display/_common.html
DavidAndreev: for .submit_revision I need to pass some contribution but on this page there is no specific contribution in context

ThiefMaster: Then maybe this should actually show a dropdown where you can select the contribution in case you have more than one (and point to the submission page for your contribution directly if you have only one)?

DavidAndreev: I dont know, was that the plan ?

ThiefMaster: At least I don't see anything else that would make sense to link to from here...

nop33: Either what @ThiefMaster said, or just link to the "My Contributions" page. However, I don't like this solution very much since it will require at least 2 extra clicks just to get to the same page as you would get from the dropdown.

DavidAndreev: Ok, yeah I agree with dropdown, but would it be like Submit paper for contribution : ( and names of contributions in dropdown with a link to submit) ? Cause the names can be pretty long in most cases. Or how do you imagine it ?

ThiefMaster: `'%d: %s' | format(contrib.friendly_id, contrib.title) | truncate()`

http://jinja.pocoo.org/docs/dev/templates/#truncate

You just need to find a suitable max length.

DavidAndreev: Then I guess we have to get rid of that feature that will open a submit dialog for you if you clicked it when not logged in, that js-show-paper-form class right ? Cause user then still needs to choose the contribution.

ThiefMaster: Does it even make much sense to show the CFP page for users who are not logged in?

DavidAndreev: Good question, well we do that for abstracts..

ThiefMaster: But CFA is a call to action that also targets people who might not even have an Indico account yet - and we show a link on the main event page.

CFP on the other hand targets people who are already involved with the conference so they are more likely to be logged in or even get a notification email from the managers about the call for papers.

DavidAndreev: So I would change displaying of CFP menu item in init.py like this `visible=lambda event: event.has_feature('papers') and session.user)` or we need to change the protection checks as well ?

ThiefMaster: Yes, if we want to hide the menu item for people who aren't logged in then that's the correct change. Also raise Forbidden on that page if the user is not logged in

indico/htdocs/sass/partials/_timelines.scss
```diff
@@ -166,3 +166,61 @@
         margin-top: 0;
     }
 }
+
+ul.deadline-timeline {
```
OmeGak: This has nothing to do with the timeline component. We need to find a different file to place this and, probably, a better naming not to confuse it with the timeline component.

DavidAndreev: what about we place it to _infogrids.scss ? Maybe not that related but maybe not bad either.

indico/htdocs/sass/partials/_infogrids.scss
```diff
@@ -133,3 +133,61 @@
...
+
+    li {
+        flex-grow: 1;
+        flex-basis: 0;
+
+        &.completed {
+            .timestamp {
+                border-top-color: $indico-blue;
+
+                &:before {
```
mvidalgarcia: According to our sass-lint and [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements), pseudo-elements must start with double colons. So `&::before` would make sass-lint happier.

indico/htdocs/sass/partials/_infogrids.scss
```diff
@@ -133,3 +133,61 @@
...
+    }
+
+    .timestamp {
+        @include border-top($width: 3px);
+
+        padding: 0 40px;
+        position: relative;
+        font-weight: bold;
+
+        &:before {
```
mvidalgarcia: Same here

indico/modules/events/papers/__init__.py
```diff
@@ -106,9 +106,16 @@ def _reviewing_area_visible(event):
             return False
         return event.cfp.can_access_reviewing_area(session.user)
 
+    def _call_for_papers_visible(event):
+        from indico.modules.events.contributions.util import has_contributions_with_user_as_submitter
+        if not session.user or not event.has_feature('papers'):
+            return False
+        return (has_contributions_with_user_as_submitter(event, session.user) or event.cfp.is_reviewer(session.user) or
+                event.cfp.is_judge(session.user))
```
ThiefMaster: I'd add this method to the CFP class and use it instead of the last two criteria here:

```python
def is_staff(self, user):
    return self.is_manager(user) or self.is_judge(user) or self.is_reviewer(user)
```

indico/modules/events/papers/controllers/display.py
```diff
@@ -214,3 +217,30 @@ def _process(self):
...
+        return WPDisplayCallForPapers.render_template('display/call_for_papers.html', self._conf,
+                                                      event=self.event_new, contributions=self.contribs,
+                                                      papers=self.papers)
+
+
+class RHSelectContribution(RHCallForPapers):
+    """Select a contribution for which the user wants to submit a paper"""
+
+    def _process(self):
+        return WPDisplayCallForPapers.render_template('display/_select_contribution.html', contributions=self.contribs)
```
ThiefMaster: shouldn't be underscored

ThiefMaster: Also, if this is rendered inside a dialog use `jsonify_template` instead of going through a WP

indico/modules/events/papers/blueprint.py
```diff
@@ -26,7 +26,8 @@
 
 
 # Display pages
-_bp.add_url_rule('/papers/', 'call_for_papers', lambda **kw: 'TODO')
+_bp.add_url_rule('/papers/', 'call_for_papers', display.RHCallForPapers)
+_bp.add_url_rule('/papers/select_contribution/', 'select_contribution', display.RHSelectContribution)
```
ThiefMaster: We use dashes in URLs instead of underscores (also, don't put a trailing slash if there's nothing "inside")

PR #2784, opened 2017-01-12
Download paper files from management assignment page

mvidalgarcia:


indico/modules/events/papers/controllers/common.py
```diff
@@ -0,0 +1,41 @@
...
+
+from __future__ import unicode_literals
+
+import os
+
+from indico.modules.events.util import ZipGeneratorMixin
+from indico.util.fs import secure_filename
+
+
+class PapersDownloadAttachmentsMixin(ZipGeneratorMixin):
```
ThiefMaster: Will this be used in more than one place? If not let's move the logic to the RH itself instead of a separate mixin

ThiefMaster: Will this be used in more than one place? If not let's move the logic to the RH itself instead of a separate mixin

If we keep the mixin, I'd go for nicer name.. e.g. `DownloadPapersMixin`

indico/modules/events/papers/controllers/management.py
```diff
@@ -226,3 +227,7 @@ def _process(self):
                                num_not_submitted_papers).format(num=num_not_submitted_papers), 'warning')
             return jsonify_data(**self.list_generator.render_list())
         return jsonify_form(form=form, submit=_('Judge'), disabled_until_change=False)
+
+
+class RHPapersDownload(PapersDownloadAttachmentsMixin, RHManagePapersActionsBase):
```
ThiefMaster: `RHDownloadPapers`?

indico/modules/events/papers/controllers/common.py
DavidAndreev: It should be used in Judging area in display too, as far as i know.

indico/modules/events/papers/controllers/common.py
```diff
@@ -0,0 +1,41 @@
...
+        file_name = secure_filename('{}_{}'.format(unicode(item.id), item.filename), item.filename)
+        return os.path.join(*self._adjust_path_length([paper_title, file_name]))
+
+    def _iter_items(self, contributions):
+        contributions_with_paper = [c for c in self.contributions if c.paper_last_revision]
+        for contrib in contributions_with_paper:
+            for f in contrib.paper_last_revision.files:
+                yield f
+
+    def _process(self):
```
OmeGak: I would move `_process` to the actual RH. It will mostly be different in every case, won't it? If that's the case, I think it's cleaner forcing to specify the logic on each RH than having the option not to override.

Otherwise, if the `_process` doesn't need to be overridden, I would make this into a full-fledged `RHDownloadPapers` like so:

```py
class RHDownloadPapers(ZipGeneratorMixin, RHManagePapersActionsBase):
    # ...
```

indico/modules/events/papers/blueprint.py
```diff
@@ -41,6 +41,8 @@
                  management.RHAssignmentListCustomize, methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/papers/assignment-list/judge', 'manage_judge_papers',
                  management.RHBulkPaperJudgment, methods=('POST',))
+_bp.add_url_rule('/manage/papers/assignment-list/download', 'manage_download_papers', management.RHDownloadPapers,
```
OmeGak: Are we going to need different handling for downloading in management and display? Maybe we don't need an endpoint specific to management.

indico/modules/events/papers/controllers/common.py
mvidalgarcia: I see your point, but honestly I don't know if the downloads will be different. I will move the `_process` to the RH for now , I think that makes sense.

indico/modules/events/papers/blueprint.py
nop33: We could use a similar method to download a zip with the files of a particular paper revision.

ThiefMaster: FYI for actions that are the same in both areas we already have support to use the same RH in both places (same thing we do in abstracts basically)

mvidalgarcia: So should I declare the blueprint to be available in both areas and move the RH declaration to a different controller e.g `paper.py` (similar to abstracts) instead of leaving it in `management.py`?
Because I think this action will be the same in management and display, isn't?

indico/modules/events/papers/controllers/common.py
```diff
@@ -0,0 +1,37 @@
...
+
+from __future__ import unicode_literals
+
+import os
+
+from indico.modules.events.util import ZipGeneratorMixin
+from indico.util.fs import secure_filename
+
+
+class DownloadPapersMixin(ZipGeneratorMixin):
```
OmeGak: Do we need this as a mixin? I don't think we will need this functionality in any other RH for now, so I would move it into `RHDownloadPapers`.

nop33: > I don't think we will need this functionality in any other RH for now

What about downloading a zip with the files from a specific paper revision (paper timeline page)?

mvidalgarcia: I left it in `common.py` in case we need it in a different place as @nop33 says but no strong opinion. It would be easy to move though.

PR #2787, opened 2017-01-17
Add paper template management

ThiefMaster:


indico/modules/events/papers/operations.py
```diff
@@ -125,3 +126,39 @@ def judge_paper(contribution, contrib_data, judgment, judge, send_notifications=
...
+    template.save(file.file)
+
+
+def create_paper_template(event, data):
+    file = data.pop('template')
+    template = PaperTemplate(event_new=event)
+    template.populate_from_dict(data)
+    _store_paper_template_file(template, file)
+    db.session.flush()
+    logger.info('Paper template %r uploaded by %r', template, session.user)
```
nop33: no flash?

ThiefMaster: Considering that events rarely have that many templates flashing a message doesn't seem that useful

ThiefMaster:
Is showing the list of templates on the main CFP dashboard really a good idea? Managers upload them once and usually that's it

indico/modules/events/papers/operations.py
nop33: Why differentiate the user's feedback from Indico when there are many or not many of the kind they are adding/removing/updating?

ThiefMaster: We also don't flash after creating a contribution type for example

nop33: I don't necessarily think this is a good practice :) It doesn't cost anything anyhow. The user is already familiar with seeing green messages when saving something. To skip it in some places might create confusion.

ThiefMaster: Added flash messages

ThiefMaster:
@nop33 I changed to icon-file-text - looks better indeed!

PR #2786, opened 2017-01-16
Send email notifications when a paper is judged

mvidalgarcia:

PR #2788, opened 2017-01-17
WIP: Paper judging area

DavidAndreev:


indico/htdocs/sass/modules/_papers.scss
```diff
@@ -39,3 +39,21 @@
         }
     }
 }
+#assign-role-table {
```
ThiefMaster: empty line before

indico/modules/events/papers/__init__.py
```diff
@@ -90,3 +90,17 @@ class PaperLayoutReviewerRole(ManagementRole):
...
+    from indico.modules.events.layout.util import MenuEntryData
+
+    def _judging_area_visible(event):
+        if not session.user or not event.has_feature('papers'):
+            return False
+        return session.user in event.cfp.judges
+
+    yield MenuEntryData(title=_("Judging Area"), name='judging_area', endpoint='papers.display_judging_area',
+                        position=-1, visible=_judging_area_visible)
+                        # parent = 'call_for_papers')
```
ThiefMaster: I'd just omit it for now; committed commented-out code always has the risk of being forgotten later

indico/modules/events/papers/controllers/common.py
```diff
@@ -0,0 +1,72 @@
...
+# License, or (at your option) any later version.
+#
+# Indico is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+import os
```
ThiefMaster: empty line before, and unicode_literals future import

indico/modules/events/papers/controllers/common.py
```diff
@@ -0,0 +1,72 @@
...
+from indico.modules.events.papers.models.revisions import PaperRevisionState
+from indico.modules.events.papers.operations import judge_paper
+from indico.modules.events.util import ZipGeneratorMixin
+from indico.util.fs import secure_filename
+from indico.util.i18n import ngettext, _
+from indico.web.util import jsonify_data, jsonify_form
+
+
+class PaperJudgmentMixin:
+    """Perform bulk change state operations on selected papers"""
```
ThiefMaster: ...bulk judgment operations...

indico/modules/events/papers/controllers/common.py
```diff
@@ -0,0 +1,72 @@
...
+            return jsonify_data(**self.list_generator.render_list())
+        return jsonify_form(form=form, submit=_('Judge'), disabled_until_change=False)
+
+
+class PapersDownloadAttachmentsMixin(ZipGeneratorMixin):
+    """Generate a ZIP file with paper files for a given list of contributions"""
+
+    def _prepare_folder_structure(self, item):
+        paper_title = secure_filename('{}_{}'.format(item.contribution.title, unicode(item.contribution.id)), 'paper')
+        file_name = secure_filename('{}_{}'.format(unicode(item.id), item.filename), item.filename)
```
ThiefMaster: no need for secure_filename here (we already do it when saving), you fallback to a default containing the only user-provided part so it does almost nothing in the end

indico/modules/events/papers/controllers/display.py
```diff
@@ -53,3 +61,114 @@ def _process(self):
 class RHPaperTimeline(RHPapersBase):
     def _process(self):
         return NotImplementedError
+
+
+class RHDisplayJudgingArea(RHPapersBase):
+    def _checkProtection(self):
+        if not session.user:
+            raise Forbidden
+            RHPapersBase._checkProtection(self)
```
ThiefMaster: unreachable code. probably too much indentation

indico/modules/events/papers/blueprint.py
```diff
@@ -30,6 +30,11 @@
                  'submit_revision', display.RHSubmitPaper, methods=('GET', 'POST'))
 _bp.add_url_rule('/papers/<int:contrib_id>/', 'paper_timeline', display.RHPaperTimeline)
 
+# Juding area
+_bp.add_url_rule('/papers/judging/', 'display_judging_area', display.RHDisplayJudgingArea)
+_bp.add_url_rule('/papers/juding/customize', 'display_customize_judging_area_list',
```
mvidalgarcia: /papers/jud**g**ing/customize

indico/modules/events/papers/controllers/display.py
```diff
@@ -53,3 +61,114 @@ def _process(self):
 class RHPaperTimeline(RHPapersBase):
     def _process(self):
         return NotImplementedError
+
+
+class RHDisplayJudgingArea(RHPapersBase):
+    def _checkProtection(self):
```
ThiefMaster: Move that to a base class that checks whether the guy has judgment rights and inherit all the judgment-related RHs from that

indico/modules/events/papers/controllers/display.py
```diff
@@ -53,3 +61,114 @@ def _process(self):
...
+        self.action = request.args.get('action')
+
+    def _process(self):
+        if self.action == 'assign':
+            if self.role == PaperReviewingRole.content_reviewer.name:
+                person_list = self.event_new.cfp.content_reviewers
+            elif self.role == PaperReviewingRole.layout_reviewer.name:
+                person_list = self.event_new.cfp.layout_reviewers
+            else:
+                raise ArgumentParseError("Invalid paper reviewing role")
```
ThiefMaster: the legacy indico exceptions are best not used at all - BadRequest is fine for this stuff

indico/modules/events/papers/controllers/display.py
```diff
@@ -53,3 +61,114 @@ def _process(self):
...
+    def _checkParams(self, params):
+        RHPapersActionsBase._checkParams(self, params)
+        self.list_generator = PaperJudgingAreaListGeneratorDisplay(event=self.event_new, user=session.user)
+
+
+class RHDisplayPapersDownload(PapersDownloadAttachmentsMixin, RHPapersActionsBase):
+    pass
+
+
+class RHJudgingAreaAssignDialog(RHPapersBase):
```
ThiefMaster: I think splitting this into two RHs for assign/unassign would be cleaner. The common code can always be put in a base class of the two RHs

urls could be `/papers/judging/dialog/assign` | `/papers/judging/dialog/unassign`

indico/modules/events/papers/controllers/display.py
```diff
@@ -53,3 +61,114 @@ def _process(self):
...
+    """Dialog with person list to assign/unassign paper reviewing roles"""
+
+    def _checkParams(self, params):
+        RHPapersBase._checkParams(self, params)
+        ids = map(int, request.form.getlist('contribution_id'))
+        self.contributions = Contribution.query.with_parent(self.event_new).filter(Contribution.id.in_(ids)).all()
+        self.role = request.args.get('role')
```
ThiefMaster: `self.role = PaperReviewingRole[request.args['role']]` - then you have the enum item itself instaed of just its string name

indico/modules/events/papers/controllers/display.py
```diff
@@ -53,3 +61,114 @@ def _process(self):
...
+        person_ids = request.form.getlist('person_id')
+        contrib_ids = request.form.getlist('contribution_id')
+        self.persons = User.query.filter(User.id.in_(person_ids), User.email != '').all()
+        self.contributions = (Contribution.query.with_parent(self.event_new)
+                              .filter(Contribution.id.in_(contrib_ids)).all())
+        self.role = request.form.get('role')
+        self.action = request.form.get('action')
+
+    def _process(self):
+        manage_reviewing_roles(self.event_new, self.persons, self.contributions, self.role, self.action)
```
ThiefMaster: `manage_*` is not a good name for an operation

indico/modules/events/papers/templates/management/_paper_assignment_list.html
```diff
@@ -1,6 +1,6 @@
 {% from 'message_box.html' import message_box %}
 
-{% macro render_paper_assignment_list(event, total_entries, contribs, static_columns) %}
+{% macro render_paper_assignment_list(event, total_entries, contribs, static_columns, management=False) %}
```
ThiefMaster: jinja prefers lowercase `false` (same for true and none)

indico/modules/events/papers/controllers/management.py
```diff
@@ -21,17 +21,17 @@
...
                                                 PaperTeamsForm, PaperReviewingSettingsForm)
 from indico.modules.events.papers.lists import PaperAssignmentListGenerator
-from indico.modules.events.papers.models.revisions import PaperRevisionState
 from indico.modules.events.papers.operations import (set_reviewing_state, update_team_members, create_competences,
-                                                     update_competences, judge_paper)
-from indico.modules.events.papers.settings import paper_reviewing_settings, RoleConverter
-from indico.modules.events.papers.views import WPManagePapers
+                                                     update_competences)
+from indico.modules.events.papers.settings import paper_reviewing_settings
 from indico.modules.events.papers.operations import schedule_cfp, open_cfp, close_cfp
```
nop33: I know you didn't edit this line, but it can be merged with line 28.

indico/modules/events/papers/lists.py
```diff
@@ -140,3 +138,37 @@ def render_list(self):
...
+        super(PaperJudgingAreaListGeneratorDisplay, self).__init__(event)
+        self.user = user
+        self.default_list_config = {
+            'items': ('state',),
+            'filters': {'items': {}}
+        }
+
+    def _build_query(self):
+        return (super(PaperJudgingAreaListGeneratorDisplay, self)._build_query()
+                .filter(db.or_(Contribution.paper_judges.any(User.id == self.user.id))))
```
nop33: I think it would look nicer like this:
``` python
query = super(PaperJudgingAreaListGeneratorDisplay, self)._build_query()
return query.filter(db.or_(Contribution.paper_judges.any(User.id == self.user.id))))
```

indico/modules/events/papers/operations.py
```diff
@@ -125,3 +126,28 @@ def judge_paper(contribution, contrib_data, judgment, judge, send_notifications=
...
+            role_title = 'Layout reviewer'
+            role_group = contrib.paper_layout_reviewers
+        else:
+            role_title = 'Judge'
+            role_group = contrib.paper_judges
+        for person in persons:
+            if action == 'assign':
+                role_group.add(person)
+            else:
+                if person in role_group:
```
nop33: Why not `elif`?

indico/modules/events/papers/templates/display/judging_area.html
```diff
@@ -0,0 +1,158 @@
...
+                {% if event.cfp.content_reviewing_enabled %}
+                    <li>
+                        <a class="js-requires-selected-row disabled"
+                           data-title="{% trans %}Assign content reviewers to selected contributions{% endtrans %}"
+                           data-href="{{ url_for('.display_judging_assign_dialog', event, role='content_reviewer', action='assign') }}"
+                           data-params-selector="#assignment-list input[name=contribution_id]:checked"
+                           data-method="POST"
+                           data-reload-after
+                           data-ajax-dialog>
+                            {% trans %}Content reviewer{% endtrans %}
```
nop33: I'd add `-` in the `trans` as you did above. Same for all occurrences below.

indico/modules/events/papers/templates/display/judging_area.html
```diff
@@ -0,0 +1,158 @@
...
+                       data-update='{"html": "#assignment-list", "filter_statistics": "#filter-statistics"}'
+                       data-method="POST"
+                       data-ajax-dialog>
+                        {%- trans %}Rejected{% endtrans -%}
+                    </a>
+                </li>
+                <li>
+                    <a class="js-requires-selected-row js-dialog-action"
+                       data-href="{{ url_for('.display_judge_papers', event) }}"
+                       data-title="{% trans %}Mark to be corrected selected papers{% endtrans %}"
```
nop33: Maybe _"Mark selected papers as to be corrected"_ is nicer?

indico/modules/events/papers/templates/management/assign_role.html
```diff
@@ -0,0 +1,67 @@
...
+                        {% if action == 'assign' %}
+                            {% if role == 'content_reviewer' %}
+                                {%- trans %}There are no users with content reviewing rights yet.{% endtrans %}
+                            {% endif %}
+                            {% if role == 'layout_reviewer' %}
+                                {%- trans %}There are no users with layout reviewing rights yet.{% endtrans %}
+                            {% endif %}
+                        {% else %}
+                            {% if role == 'content_reviewer' %}
+                                {%- trans %}There are no content reviewers assigned to selected contributions.{% endtrans %}
```
nop33: _"...assigned to **the** selected contributions"_

indico/modules/events/papers/controllers/common.py
DavidAndreev: What do you mean by unicode_literals future import ?

ThiefMaster: `from __future__ import unicode_literals` like we do in all other new non-legacy files

indico/htdocs/sass/modules/_papers.scss
```diff
@@ -39,3 +39,23 @@
         }
     }
 }
+
+#assign-role-table {
```
OmeGak: Writing down this only to trigger a little discussion. I think we should try to avoid as much as possible using IDs for styling. For several reasons.

It's true that we want to scope changes and ID selectors are a good way to do that. When we have no other choice but to that I think we should try to use the broadest scope possible that fit. In this case it would be the ID of the `page_class` defined in the template.

However, I think we should go even further and try not to develop so many module/page specific components and component tweaks. Instead, if we need a slight variation we should try to add a new "mode" in the component. The reason is re-usability and maintainability. On the one hand if we ever make changes to the component we clearly see all the other modes that may be affected without having to search the full project. On the other hand, if we ever need to re-use the slight tweaks, they are already abstracted and ready to use simply adding a class. 

One side effect of this approach is that it's also easier to see in the HTML if the original component is modified. In this case, for instance, we would use `<div class="i-tag i-tag--special">blablabla</div>`.

Thoughts? /cc @pferreir @ThiefMaster @nop33 @DavidAndreev 
(We can also discuss offline)

indico/modules/events/papers/controllers/display.py
```diff
@@ -53,3 +60,123 @@ def _process(self):
...
+        self.action = request.form.get('action')
+        self.role = PaperReviewingRole[request.form.get('role')]
+
+    def _process(self):
+        update_reviewing_roles(self.event_new, self.persons, self.contributions, self.role, self.action)
+        if self.action == 'assign':
+            flash(_("Paper reviewing roles have been assigned."), 'success')
+        else:
+            flash(_("Paper reviewing roles have been unassigned."), 'success')
+        db.session.flush()
```
OmeGak: Why do we need a `flush()` here?

indico/modules/events/papers/operations.py
```diff
@@ -125,3 +126,24 @@ def judge_paper(contribution, contrib_data, judgment, judge, send_notifications=
...
+            role_group = contrib.paper_layout_reviewers
+        else:
+            role_group = contrib.paper_judges
+        for person in persons:
+            if action == 'assign':
+                role_group.add(person)
+            elif person in role_group:
+                role_group.remove(person)
+        event.log(EventLogRealm.management, EventLogKind.positive, 'Papers',
+                  "Modified paper reviewing roles for contribution {}.".format(contrib.title), session.user,
```
OmeGak: Not a sentence. I would go for:

```
"Paper reviewing roles modified for contribution {}."
```

indico/modules/events/papers/templates/display/judging_area.html
```diff
@@ -0,0 +1,158 @@
+{% extends 'layout/conference_page_base.html' %}
+{% from 'events/papers/management/_paper_assignment_list.html' import render_paper_assignment_list %}
+{% from 'events/management/_lists.html' import render_displayed_entries_fragment %}
+
+{% block title %}
+    {%- trans %}Judging Area{% endtrans -%}
+{% endblock %}
+{% block description %}
```
OmeGak: This could help an empty line before.

indico/modules/events/papers/controllers/display.py
ThiefMaster: we don't, there should be one in the operation anyway

indico/htdocs/sass/modules/_papers.scss
nop33: I think this is the best approach we can follow. Reusing components will save us some headaches and make our codebase cleaner.

pferreir: The problem is maintaining such components. If we're speaking of things that are clearly reusable (e.g. a table), yeah, that seems fair enough, however.

nop33: Almost everything is based on a more 'core' component though. The `#assign-role-table` for example is technically a `.i-table-widget`. So in this example, a new "mode" for the `.i-table-widget` could be created. Unfortunately, it's not always easy to come up with a new mode, especially when what you want to change in the core component is a minor thing.

nop33:
Getting an error when trying to bulk accept papers from the management area.
```
(psycopg2.IntegrityError) new row for relation "revisions" violates check constraint "ck_revisions_judge_if_judged" DETAIL: Failing row contains (1, 2389173, 61385, 2017-01-16 15:57:40.60894, 4, Can you please correct this and that and the other thing over th..., null, null).
```

ThiefMaster:
@nop33 does this come from an autoflush or is the judge never set? If it's autoflush-caused just add `@no_autoflush` to the operation

nop33:
Shouldn't this PR also add a menu item in the display page to give access to the judging area?

nop33:
The spacing between the buttons in the search field area doesn't seem quite right:
image

Make sure to include the word "WIP: " in front of your PR title when you push TODO stuff :wink:

nop33:
@ThiefMaster 
```
  File "/home/iliastrichopoulos/dev/indico/indico/modules/events/papers/operations.py", line 121, in judge_paper
    db.session.flush()
```
The judge is set.

ThiefMaster:
which state was set? this is the constraint that failed:

                      db.CheckConstraint('(state IN ({}, {}, {})) = (judge_id IS NOT NULL)'
                                         .format(PaperRevisionState.accepted, PaperRevisionState.rejected,
                                                 PaperRevisionState.to_be_corrected),
                                         name='judge_if_judged'),

nop33:
@ThiefMaster `PaperRevisionState.to_be_corrected`

mvidalgarcia:
@nop33 That error when bulk accepting will be solved with [this](https://github.com/indico/indico/pull/2786/commits/7ec2b63f670c2525bf9026f4204c030308e4972f) commit.

indico/modules/events/papers/templates/display/judging_area.html
```diff
@@ -0,0 +1,159 @@
...
+                           data-params-selector="#assignment-list input[name=contribution_id]:checked"
+                           data-method="POST"
+                           data-reload-after
+                           data-ajax-dialog>
+                            {%- trans %}Layout reviewer{% endtrans %}
+                        </a>
+                    </li>
+                {% endif %}
+            </ul>
+            <a class="i-button arrow js-dropdown js-enable-if-checked disabled" data-toggle="dropdown">
```
mvidalgarcia: You can add the hammer icon with the `icon-hammer` class

DavidAndreev:
@nop33 you need to be added as a judge in paper team to be able to see the menu item 
```python
def _judging_area_visible(event):
      if not session.user or not event.has_feature('papers'):
           return False
      return session.user in event.cfp.judges
```

indico/modules/events/papers/templates/display/judging_area.html
```diff
@@ -0,0 +1,159 @@
+{% extends 'layout/conference_page_base.html' %}
+{% from 'events/papers/management/_paper_assignment_list.html' import render_paper_assignment_list %}
+{% from 'events/management/_lists.html' import render_displayed_entries_fragment %}
```
nop33: Should be moved above.

indico/modules/events/papers/controllers/display.py
```diff
@@ -53,3 +60,122 @@ def _process(self):
...
+
+class RHJudgingAreaBase(RHPapersBase):
+    def _checkProtection(self):
+        if not session.user or session.user not in self.event_new.cfp.judges:
+            raise Forbidden
+        RHPapersBase._checkProtection(self)
+
+
+class RHDisplayJudgingArea(RHJudgingAreaBase):
+
```
nop33: remove empty line if there is no docstring.

indico/modules/events/papers/controllers/display.py
```diff
@@ -53,3 +60,122 @@ def _process(self):
...
+            person_list = self.event_new.cfp.content_reviewers
+        elif self.role == PaperReviewingRole.layout_reviewer:
+            person_list = self.event_new.cfp.layout_reviewers
+        else:
+            raise BadRequest
+        return self._render_template(person_list, 'assign')
+
+
+class RHJudgingAreaUnassign(RHRJudgingAreaAssigningBase):
+    """Dialog with person list to unassign paper reviewing roles"""
```
nop33: I don't think that the docstring should be coupled with the UI. Just writing *"Renders the person list to unassign paper reviewing roles"* would be enough IMO

ThiefMaster:
Superseded by #2811

PR #2789, opened 2017-01-17
Calculate room occupancy considering lunch break

pferreir:
We can actually have an arbitrary number of time periods. The goal is to have a more realistic estimate of the real occupancy of the room.

### DIFF
--- a/indico/MaKaC/webinterface/tpls/RoomBookingAdminLocation.tpl
+++ b/indico/MaKaC/webinterface/tpls/RoomBookingAdminLocation.tpl
@@ -386,7 +386,7 @@ indicoRequest(
               <span style="background-color: #C9FFC9; font-weight: bold;">
                 ${ '{0:.02f}'.format(kpi['occupancy'] * 100) }<small>%</small>
               </span>
-              ${inlineContextHelp('Average room occupancy in last 30 days during working hours (8H30-17H30, Monday-Friday including holidays). Only active, publically reservable rooms are taken into account.' )}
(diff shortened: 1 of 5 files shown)


indico/modules/rb/models/locations.py
```diff
@@ -32,8 +32,18 @@ class Location(db.Model):
     __tablename__ = 'locations'
     __table_args__ = {'schema': 'roombooking'}
 
-    working_time_start = time(8, 30)
-    working_time_end = time(17, 30)
+    # TODO: Turn this into a proper admin setting
```
OmeGak: I was about to suggest that. :smile:

indico/MaKaC/webinterface/tpls/RoomBookingAdminLocation.tpl
```diff
@@ -386,7 +386,7 @@ indicoRequest(
               <span style="background-color: #C9FFC9; font-weight: bold;">
                 ${ '{0:.02f}'.format(kpi['occupancy'] * 100) }<small>%</small>
               </span>
-              ${inlineContextHelp('Average room occupancy in last 30 days during working hours (8H30-17H30, Monday-Friday including holidays). Only active, publically reservable rooms are taken into account.' )}
+              ${inlineContextHelp('Average room occupancy in last 30 days during working hours (8:30 - 12:30 and 13:30 - 17:30, Monday-Friday including holidays). Only active, publically reservable rooms are taken into account.' )}
```
OmeGak: Is it worth maybe rendering the working hours from `working_time_periods` already?

pferreir: Thought of it, but then I will have to take care of i18n and the whole thing will end up being a mess. When we turn this into a true setting, then it makes sense to make it dynamic, IMHO.

PR #2791, opened 2017-01-18
Store relative path of cached BoA

pferreir:
Mostly discussed on IRC.

### DIFF
--- a/indico/modules/events/abstracts/util.py
+++ b/indico/modules/events/abstracts/util.py
@@ -16,6 +16,7 @@
 
 from __future__ import unicode_literals
 
+import errno
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


indico/modules/events/abstracts/util.py
```diff
@@ -278,17 +279,22 @@ def create_boa(event):
...
 
 
 def clear_boa_cache(event):
     """Delete the cached book of abstract"""
     path = boa_settings.get(event, 'cache_path')
     if path:
-        os.remove(path)
+        try:
+            os.remove(os.path.join(Config.getInstance().getXMLCacheDir(), path))
+        except OSError, e:
```
ThiefMaster: `as e` (i'll amend it)

PR #2793, opened 2017-01-19
Paper timeline

nop33:
This PR includes:
* Displaying the information and files of the 1st submitted revision
* Allows commenting
* Allows reviewing

Things to be included in the next PR:
* Allow judging
* Display info of every other revision
* Check that all permissions work fine
* ?

When merged, execute:

```sql
ALTER TABLE event_paper_reviewing.review_comments
        ADD COLUMN visibility smallint NOT NULL;

ALTER TABLE event_paper_reviewing.review_comments
        ADD CONSTRAINT ck_review_comments_valid_enum_visibility CHECK ((visibility = ANY (ARRAY[1, 2, 3, 4])));
```

### DIFF
--- a/indico/htdocs/sass/modules/_reviewing.scss
+++ b/indico/htdocs/sass/modules/_reviewing.scss
@@ -492,3 +492,9 @@
         }
     }
 }
+
+.review-item-content {
+    @extend .i-box;
+    @extend .i-box.no-shadow;
+    margin-top: 1rem;
+}
--- a/indico/modules/events/abstracts/models/abstracts.py
+++ b/indico/modules/events/abstracts/models/abstracts.py
@@ -32,7 +32,7 @@ from indico.modules.events.abstracts.models.reviews import AbstractAction, Abstr
 from indico.modules.events.abstracts.models.review_questions import AbstractReviewQuestion
 from indico.modules.events.abstracts.models.review_ratings import AbstractReviewRating
 from indico.modules.events.models.persons import AuthorsSpeakersMixin
-from indico.modules.events.models.reviews import ProposalMixin
+from indico.modules.events.models.reviews import ProposalMixin, ProposalRevisionMixin
 from indico.modules.events.contributions.models.contributions import _get_next_friendly_id, CustomFieldsMixin
 from indico.util.date_time import now_utc
 from indico.util.i18n import _
@@ -85,7 +85,8 @@ class EditTrackMode(int, IndicoEnum):
     reviewed_for = 2
 
 
-class Abstract(ProposalMixin, DescriptionMixin, CustomFieldsMixin, AuthorsSpeakersMixin, db.Model):
+class Abstract(ProposalMixin, ProposalRevisionMixin, DescriptionMixin, CustomFieldsMixin, AuthorsSpeakersMixin,
+               db.Model):
     """Represents an abstract that can be associated to a Contribution."""
 
     __tablename__ = 'abstracts'
@@ -124,6 +125,7 @@ class Abstract(ProposalMixin, DescriptionMixin, CustomFieldsMixin, AuthorsSpeake
     edit_comment_endpoint = 'abstracts.edit_abstract_comment'
     create_review_endpoint = 'abstracts.review_abstract'
     edit_review_endpoint = 'abstracts.edit_review'
+    revisions_enabled = False
 
     @declared_attr
     def __table_args__(cls):
--- a/indico/modules/events/abstracts/models/abstracts.py
+++ b/indico/modules/events/abstracts/models/abstracts.py
@@ -43,6 +43,7 @@ from indico.util.struct.enum import RichIntEnum, IndicoEnum
 
 class AbstractState(RichIntEnum):
     __titles__ = [None, _("Submitted"), _("Withdrawn"), _("Accepted"), _("Rejected"), _("Merged"), _("Duplicate")]
+    __css_classes__ = [None, '', 'outline dashed', 'success', 'error', 'visited', 'strong']
     submitted = 1
     withdrawn = 2
     accepted = 3
@@ -54,6 +55,8 @@ class AbstractState(RichIntEnum):
 class AbstractPublicState(RichIntEnum):
     __titles__ = {i: title for i, title in enumerate(AbstractState.__titles__[2:], 2)}
     __titles__.update({-1: _("Awaiting Review"), -2: _("Under Review")})
+    __css_classes__ = {i: css_class for i, css_class in enumerate(AbstractState.__css_classes__[2:], 2)}
+    __css_classes__.update({-1: '', -2: 'highlight'})
     # regular states (must match AbstractState!)
     withdrawn = 2
     accepted = 3
@@ -67,6 +70,7 @@ class AbstractPublicState(RichIntEnum):
 
 class AbstractReviewingState(RichIntEnum):
     __titles__ = [_("Not Started"), _("In progress"), _("Positive"), _("Conflicting"), _("Negative"), _("Mixed")]
+    __css_classes__ = ['', '', 'success', '', 'error', 'warning']
     not_started = 0
     in_progress = 1
     positive = 2
--- a/indico/modules/events/models/reviews.py
+++ b/indico/modules/events/models/reviews.py
@@ -167,21 +167,6 @@ class ProposalCommentMixin(object):
         raise NotImplementedError
 
 
-class ProposalCommentVisibility(RichIntEnum):
-    """Most to least restrictive visibility for abstract comments"""
-    __titles__ = [None,
-                  _("Visible only to judges"),
-                  _("Visible to conveners and judges"),
-                  _("Visible to reviewers, conveners, and judges"),
-                  _("Visible to contributors, reviewers, conveners, and judges"),
-                  _("Visible to all users")]
-    judges = 1
-    conveners = 2
-    reviewers = 3
-    contributors = 4
-    users = 5
-
-
 class ProposalReviewMixin(object):
     """Mixin for proposal reviews
 
@@ -206,10 +191,6 @@ class ProposalReviewMixin(object):
     @property
     def group(self):
         return self.group_proxy_cls(getattr(self, self.group_attr))
-
-    @property
-    def visibility(self):
-        return ProposalCommentVisibility.reviewers
 
     @property
     def score(self):
--- a/indico/modules/events/models/reviews.py
+++ b/indico/modules/events/models/reviews.py
@@ -19,9 +19,7 @@ from __future__ import unicode_literals
 from sqlalchemy.ext.hybrid import hybrid_property
 
 from indico.util.caching import memoize_request
-from indico.util.i18n import _
 from indico.util.locators import locator_property
-from indico.util.struct.enum import RichIntEnum
 from indico.web.flask.util import url_for
 
 
--- a/indico/modules/events/models/reviews.py
+++ b/indico/modules/events/models/reviews.py
@@ -20,43 +20,64 @@ from sqlalchemy.ext.hybrid import hybrid_property
 
 from indico.util.caching import memoize_request
 from indico.util.i18n import _
+from indico.util.locators import locator_property
 from indico.util.struct.enum import RichIntEnum
 from indico.web.flask.util import url_for
 
 
-class ProposalMixin(object):
-    proposal_type = None
-    call_for_proposals_attr = None
+class ProposalGroupProxy(object):
+    """Represents the object that the proposals can be grouped by.
 
-    # endpoints
-    delete_comment_endpoint = None
-    create_comment_endpoint = None
-    edit_comment_endpoint = None
-    create_review_endpoint = None
-    edit_review_endpoint = None
+    It provides all necessary methods for building the URLs, displaying the
+    grouping information, etc.
+    """
+
+    title_attr = 'title'
+    full_title_attr = 'full_title'
+
+    def __init__(self, group):
+        self.instance = group
+
+    def __eq__(self, other):
+        if not isinstance(other, ProposalGroupProxy):
+            return False
+        return self.instance == other.instance
+
+    def __hash__(self):
+        return hash(self.instance)
+
+    def __ne__(self, other):
+        return not (self == other)
 
     @property
-    def cfp(self):
-        return getattr(self.event_new, self.call_for_proposals_attr)
+    def title(self):
+        return getattr(self.instance, self.title_attr)
 
-    def can_comment(self, user):
-        raise NotImplementedError
+    @property
+    def full_title(self):
+        return (getattr(self.instance, self.full_title_attr)
+                if hasattr(self.instance, self.full_title_attr)
+                else self.title)
 
-    def can_review(self, user, check_state=False):
-        raise NotImplementedError
+    @locator_property
+    def locator(self):
+        return self.instance.locator
 
-    def get_delete_comment_url(self, comment):
-        return url_for(self.delete_comment_endpoint, comment)
 
-    def get_save_comment_url(self, comment=None):
-        return (url_for(self.edit_comment_endpoint, comment)
-                if comment
-                else url_for(self.create_comment_endpoint, self))
+class ProposalRevisionMixin(object):
+    """ Properties and methods of a proposal revision."""
+    #: The attribute  of the revision used to fetch the proposal object.
+    proposal_attr = None
+    #: Whether the reviewing process supports multiple revisions per proposal.
+    #: If set to false it is assumed that the reviewing process supports only
+    #: one revision per proposal.
+    revisions_enabled = True
 
-    def get_save_review_url(self, group=None, review=None):
-        return (url_for(self.edit_review_endpoint, review)
-                if review
-                else url_for(self.create_review_endpoint, self, group))
+    @property
+    def proposal(self):
+        # Property to fetch the proposal object. If multiple revisions are
+        # disabled, the revision is represented by the proposal object itself.
+        return getattr(self, self.proposal_attr) if self.revisions_enabled else self
 
     def get_timeline(self, user=None):
         raise NotImplementedError
@@ -84,6 +105,61 @@ class ProposalMixin(object):
                 'reviews': reviews}
 
 
+class ProposalMixin(object):
+    """Classes that represent a proposal object should extend this class (ex:
+    Abstract, Paper).
+    """
+    #: A unique identifier to handle rendering differences between proposal
+    #: types
+    proposal_type = None
+    #: Attribute to retrieve the object with access to the reviewing settings
+    call_for_proposals_attr = None
+    #: Whether there is support for multiple revisions per proposal or just one
+    revisions_enabled = True
+
+    # endpoints
+    delete_comment_endpoint = None
+    create_comment_endpoint = None
+    edit_comment_endpoint = None
+    create_review_endpoint = None
+    edit_review_endpoint = None
+
+    @property
+    def cfp(self):
+        return getattr(self.event_new, self.call_for_proposals_attr)
+
+    def get_revisions(self):
+        if self.revisions_enabled:
+            raise NotImplementedError
+        else:
+            return [self]
+
+    def get_last_revision(self):
+        if self.revisions_enabled:
+            raise NotImplementedError
+        else:
+            return self
+
+    def can_comment(self, user):
+        raise NotImplementedError
+
+    def can_review(self, user, check_state=False):
+        raise NotImplementedError
+
+    def get_delete_comment_url(self, comment):
+        return url_for(self.delete_comment_endpoint, comment)
+
+    def get_save_comment_url(self, comment=None):
+        return (url_for(self.edit_comment_endpoint, comment)
+                if comment
+                else url_for(self.create_comment_endpoint, self))
+
+    def get_save_review_url(self, group=None, review=None):
+        return (url_for(self.edit_review_endpoint, review)
+                if review
+                else url_for(self.create_review_endpoint, self, group))
+
+
 class ProposalCommentMixin(object):
     timeline_item_type = 'comment'
 
@@ -107,17 +183,29 @@ class ProposalCommentVisibility(RichIntEnum):
 
 
 class ProposalReviewMixin(object):
+    """Mixin for proposal reviews
+
+    Classes that represent a review of a proposal should extend this class
+    (ex: AbstractReview, PaperReview).
+    """
+    #: A unique identifier to handle rendering differences between timeline
+    #: items
     timeline_item_type = 'review'
-    proposal_relationship = None
-    group_relationship = None
+    #: The revision object that the review refers to
+    revision_attr = None
+    #: Object used to group reviews together
+    group_attr = None
+    #: Proxy class to provide the necessary properties and methods to the
+    #: review grouping object
+    group_proxy_cls = ProposalGroupProxy
 
     @hybrid_property
-    def proposal(self):
-        return getattr(self, self.proposal_relationship)
+    def revision(self):
+        return getattr(self, self.revision_attr)
 
     @property
     def group(self):
-        return getattr(self, self.group_relationship)
+        return self.group_proxy_cls(getattr(self, self.group_attr))
 
     @property
     def visibility(self):
--- a/indico/modules/events/papers/controllers/display.py
+++ b/indico/modules/events/papers/controllers/display.py
@@ -19,7 +19,7 @@ from __future__ import unicode_literals
 from flask import request, session
 
 from indico.modules.events.papers.controllers.base import RHPaperBase
-from indico.modules.events.papers.forms import PaperSubmissionForm
+from indico.modules.events.papers.forms import PaperSubmissionForm, PaperCommentForm
 from indico.modules.events.papers.models.files import PaperFile
 from indico.modules.events.papers.operations import create_paper_revision
 from indico.modules.events.papers.views import WPDisplayPapersBase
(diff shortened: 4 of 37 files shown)


nop33:
Current functional view that is going to be improved.
image

indico/htdocs/sass/modules/_reviewing.scss
```diff
@@ -492,3 +492,9 @@
         }
     }
 }
+
+.reviewing-item-content {
```
OmeGak: Not consistent with the other classes. I would rename this to `.review-item-content`.

indico/modules/events/models/reviews.py
```diff
@@ -19,44 +19,54 @@
...
+        return self.instance.locator
 
-    def get_delete_comment_url(self, comment):
-        return url_for(self.delete_comment_endpoint, comment)
 
-    def get_save_comment_url(self, comment=None):
-        return (url_for(self.edit_comment_endpoint, comment)
-                if comment
-                else url_for(self.create_comment_endpoint, self))
+class ProposalRevisionMixin(object):
```
OmeGak: I would add docstring to the class and maybe to the attributes as well (`#: ...`).

indico/modules/events/models/reviews.py
```diff
@@ -19,44 +19,54 @@
...
+    proposal_attr = None
+    revisions_enabled = True
 
-    def get_save_review_url(self, group=None, review=None):
-        return (url_for(self.edit_review_endpoint, review)
-                if review
-                else url_for(self.create_review_endpoint, self, group))
+    @property
+    def proposal(self):
```
OmeGak: I would add something along these lines here.

```python
# If revisions are disabled we assume ProposalRevisionMixin is also ProposalMixin.
```

indico/modules/events/models/reviews.py
```diff
@@ -84,44 +94,74 @@ def get_reviewer_render_data(self, user):
                 'reviews': reviews}
 
 
+class ProposalMixin(object):
```
OmeGak: Docstrings here maybe as well? Class and `proposal_type`, `call_for_proposals_attr`, and `revisions_enabled`.

indico/modules/events/models/reviews.py
```diff
@@ -84,44 +94,74 @@ def get_reviewer_render_data(self, user):
...
-    reviewers = 3
-    contributors = 4
-    users = 5
-
-
 class ProposalReviewMixin(object):
     timeline_item_type = 'review'
-    proposal_relationship = None
+    revision_attr = None
     group_relationship = None
```
OmeGak: I would rename `group_relationship` to `group_relationship` not to make unnecessary assumptions.

indico/modules/events/models/reviews.py
```diff
@@ -84,44 +94,74 @@ def get_reviewer_render_data(self, user):
...
-                  _("Visible to contributors, reviewers, conveners, and judges"),
-                  _("Visible to all users")]
-    judges = 1
-    conveners = 2
-    reviewers = 3
-    contributors = 4
-    users = 5
-
-
 class ProposalReviewMixin(object):
```
OmeGak: Docstrings.

indico/modules/events/papers/controllers/display.py
```diff
@@ -16,29 +16,39 @@
...
-from indico.modules.events.contributions.models.contributions import Contribution
-from indico.modules.events.papers.controllers.base import RHPapersBase
-from indico.modules.events.papers.forms import PaperSubmissionForm
-from indico.modules.events.papers.operations import create_paper_revision
-from indico.web.util import jsonify_form, jsonify_data
+from indico.modules.events.papers.controllers.base import RHPaperBase
+from indico.modules.events.papers.forms import PaperSubmissionForm, PaperCommentForm, build_review_form
+from indico.modules.events.papers.models.files import PaperFile
+from indico.modules.events.papers.models.reviews import PaperReviewType, PaperReview, PaperTypeProxy
+from indico.modules.events.papers.models.comments import PaperReviewComment
```
OmeGak: Import order.

indico/modules/events/papers/controllers/display.py
```diff
@@ -16,29 +16,39 @@
...
+from indico.web.util import jsonify_form, jsonify_data, jsonify
 
 
-class RHSubmitPaper(RHPapersBase):
-    normalize_url_spec = {
-        'locators': {
-            lambda self: self.contribution
-        }
-    }
+def render_paper_page(paper, view_class=None):
```
OmeGak: I would move this to `views.py`.

indico/modules/events/papers/forms.py
```diff
@@ -18,21 +18,64 @@
...
+    return form_class
+
+
+def build_review_form(paper_revision=None, type_=None, review=None):
+    if review:
+        paper_revision = review.revision
+        type_ = review.type
+    review_form_class = make_review_form(paper_revision.paper.event_new, type_=type_)
+    reviews = paper_revision.get_reviews(user=session.user, group=type_)
+    review_for_type = reviews[0] if reviews else None
```
OmeGak: Is this a type or a review? The naming is not very clear.

indico/modules/events/papers/forms.py
```diff
@@ -145,3 +188,42 @@ def __init__(self, *args, **kwargs):
...
+        if not paper.can_judge(user):
+            self.visibility.skip.add(PaperCommentVisibility.judges)
+        if not paper.can_review(user):
+            del self.visibility
+
+
+class PaperReviewForm(IndicoForm):
+    """Form for reviewing a paper"""
+
+    _order = ('proposed_action', 'comment')
```
OmeGak: Why do we need to specify an order here? Usually we do this when extending the from a base class, but that's not the case.

indico/modules/events/papers/models/papers.py
```diff
@@ -66,3 +68,45 @@ def state(self):
     @state.setter
     def state(self, state):
         self.last_revision.state = state
+
+    def is_in_final_state(self):
+        return self.state in {PaperRevisionState.accepted or PaperRevisionState.rejected}
+
+    def can_comment(self, user):
+        if not user:
```
OmeGak: This is starting to become a quite repeated pattern for protection checks. Maybe it would be interesting having a decorator `@protection_check` that does it for you. /cc @ThiefMaster

indico/modules/events/papers/models/papers.py
```diff
@@ -66,3 +68,45 @@ def state(self):
...
+            return False
+        return self.is_submitter(user) or self.can_judge(user) or self.can_review(user)
+
+    def is_submitter(self, user):
+        if not user:
+            return False
+        return user in {r.submitter for r in self.revisions} or user in self.contribution.submitters
```
OmeGak: Can you be a paper submitter if you are not a contribution submitter?

indico/modules/events/papers/models/papers.py
```diff
@@ -66,3 +68,45 @@ def state(self):
...
+        return user in self.contribution.paper_judges
+
+    def can_review(self, user, check_state=False):
+        if not user:
+            return False
+        elif check_state and self.is_in_final_state():
+            return False
+        elif self.can_manage(user):
+            return True
+        return user in self.contribution.paper_content_reviewers or user in self.contribution.paper_layout_reviewers
```
OmeGak: What about adding `Contribution.is_paper_reviewer` method?

```python
def is_paper_reviewer(self, user):
    return user in self.paper_content_reviewers or user in self.paper_layout_reviewers
```

indico/modules/events/papers/models/papers.py
```diff
@@ -66,3 +68,45 @@ def state(self):
...
+
+    def get_revisions(self):
```
OmeGak: I would add a comment specifying this is required because of `ProposalMixin`.

indico/modules/events/papers/models/papers.py
```diff
@@ -66,3 +68,45 @@ def state(self):
...
+        return self.revisions
+
+    def get_last_revision(self):
```
OmeGak: Same here.

indico/modules/events/papers/models/revisions.py
```diff
@@ -152,3 +158,26 @@ def paper(self):
     @paper.setter
     def paper(self, paper):
         self._contribution = paper.contribution
+
+    def get_timeline(self, user=None):
+        comments = [x for x in self.comments if x.can_view(user)] if user else self.comments
+        reviews = [x for x in self.reviews if x.can_view(user)] if user else self.reviews
+        return sorted(chain(comments, reviews), key=attrgetter('created_dt'))
+
+    def get_reviews(self, group=None, user=None):
```
OmeGak: I would rewrite this method as:

```python
def get_reviews(self, group=None, user=None):
    reviews = []
    if group and user:
        reviews = [x for x in self.reviews if x.group.instance == group and x.user == user]
    elif group:
        reviews = [x for x in self.reviews if x.group.instance == group]
    elif user:
        reviews = [x for x in self.reviews if x.user == user]
    return reviews
```

indico/modules/events/papers/operations.py
```diff
@@ -164,3 +168,86 @@ def delete_paper_template(template):
...
+    comment.populate_from_dict(comment_data)
+    comment.paper_revision = paper.last_revision
+    db.session.flush()
+    logger.info("Paper %s received a comment from %s", paper, session.user)
+    paper.event_new.log(EventLogRealm.management, EventLogKind.positive, 'Papers',
+                        'Paper {} received a comment'.format(paper.verbose_title),
+                        session.user)
+
+
+def update_comment(comment, comment_data):
```
OmeGak: I see you changed it from my commit. I think passing `*_data` is a bad pattern, especially taking into account operation functions should be easy to use from command line. My point is, you cannot know what `comment_data` is unless you go and check the form, while being explicit you can see right away what the function is expecting.

I would rewrite it like:

```python
def update_comment(comment, text, visibility, user):
    # ...
```

indico/modules/events/papers/operations.py
```diff
@@ -164,3 +168,86 @@ def delete_paper_template(template):
...
+    changes = comment.populate_from_dict(comment_data)
+    comment.modified_by = session.user
```
OmeGak: We need to pass the user explicitly since `session.user` resolves to `None` from the CLI. While that's fine for the logger, we don't want to end up with `comment.modified_dt` set to a timestamp while `comment.modified_by` pointing to a ghost user.

indico/modules/events/papers/operations.py
```diff
@@ -164,3 +168,86 @@ def delete_paper_template(template):
...
+        'proposed_action': 'Action',
+        'comment': 'Comment'
+    })
+    event.log(EventLogRealm.management, EventLogKind.change, 'Papers',
+              'Review for paper {} modified'.format(paper.verbose_title),
+              session.user, data={'Changes': make_diff_log(changes, log_fields)})
+
+
+@no_autoflush
+def create_comment(paper, comment_data, user):
```
OmeGak: Same as in `update_comment` I would rewrite this function to explicit in the arguments.

indico/modules/events/templates/reviews/_omnibox.html
```diff
@@ -72,7 +72,7 @@
                         {% if group in reviewer_data.reviewed_groups %}
                             {% set review = reviewer_data.reviews[group] %}
                             <a class="js-new-edit-review" data-review-id="{{ review.id }}">
-                                {{ group.full_title }}
+                                {{ group.full_title if group.full_title else group.title }}
```
OmeGak: Not needed anymore. `group.full_title` falls back to `group.title` at the proxy level.

indico/modules/events/templates/reviews/rating_widget.html
```diff
@@ -15,7 +15,7 @@
         </div>
         <div id="{{ field.id }}" class="rating-scale flexrow f-a-center">
             <span class="score">{{ '-' if field.object_data is none else field.data }}</span>
-            {% for bullet in range(cfa.rating_range[1], cfa.rating_range[0] - 1, -1) %}
+            {% for bullet in range(rating_range[1], rating_range[0] - 1, -1) %}
```
OmeGak: Where is `rating_range` coming from?

indico/modules/events/papers/models/papers.py
ThiefMaster: Seems somewhat obscure to me, especially to save just one line.

OmeGak: Two lines in every single method that is used to do protection check, which are starting to sprawl and it's easier to forget adding the check than the decorator. We could make it less obscure and call it `@require_user`.

indico/modules/events/models/reviews.py
nop33: image
love it :stuck_out_tongue_closed_eyes:

indico/modules/events/papers/controllers/display.py
nop33: Hm, isn't it more like a utility method that should be placed in `util.py` similarly to:
* `render_entry_info_balloon`
* `render_session_timetable`
* `render_placeholder_info`
* `render_engine_or_search_template`
* ...

Searching for `def render_` in our project I couldn't find any of those methods in a `views.py` file

OmeGak: Maybe they should. In my opinion, the rendering of stuff belongs precisely to the view.

indico/modules/events/papers/models/papers.py
nop33: Depends if the authors/co-authors of the contribution can submit a paper or if it's only the submitter of the contribution that can..

indico/modules/events/papers/models/papers.py
nop33: Fine!

indico/modules/events/papers/models/papers.py
nop33: Rly? It raises an `NotImplementedError`, why document it? Isn't it enough? Then we would have to do the same to all methods that raise such an error!

indico/modules/events/papers/operations.py
nop33: OK, fair enough!

indico/modules/events/templates/reviews/rating_widget.html
nop33: Same place where the `cfa` was coming from :) Since I had to reuse the widget in the papers module, I couldn't pass the `cfa` object anymore.

indico/modules/events/papers/models/papers.py
OmeGak: I would say only contribution submitters are also paper submitters.

indico/modules/events/templates/reviews/rating_widget.html
nop33: `make_review_form`

indico/modules/events/papers/models/papers.py
ThiefMaster: But the contribution submitters are configured by the event managers - while the author list usually comes from the abstract. I think for paper submission it makes much more sense to use the latter.

indico/modules/events/papers/forms.py
nop33: If we don't use it, the reviewing questions go to the bottom which is not what we want:
image

indico/modules/events/papers/models/papers.py
nop33: So what do you suggest @ThiefMaster? To do a check over the authors/co-authors/speakers of the paper and not check the submitters of the contribution?

ThiefMaster: Yes, I would allow all 'persons' associated with the contribution to upload papers. After all, the authors are the ones most likely to also write and upload the paper

indico/modules/events/papers/controllers/display.py
nop33: But we always call `render_template` in the RHs... I understand the MVC point of view (ha, I said "view" again, ta-dum-tsss). Anyway, not a strong opinion about this, what do the others think?

indico/modules/events/papers/models/papers.py
nop33: so basically the line changes to 
```python
return any(x.person.user == user for x in self.contribution.person_links)
```
right?

ThiefMaster: yep, but i'd add `if x.person.user` there (even though we bail out early anyway if user is None)

indico/modules/events/papers/controllers/display.py
ThiefMaster: > [14:58:07] <+ThiefMaster> if it's as big as e.g. render_abstract_page and also used in multiple controller .py files i'd go for views.py
[14:58:52] <+ThiefMaster> but if it's just used in one file i'd keep it in there, with an underscored function name

indico/modules/events/papers/views.py
```diff
@@ -38,7 +41,24 @@ class WPDisplayPapersBase(WPJinjaMixin, WPConferenceDefaultDisplayBase):
...
         return WPJinjaMixin._getPageContent(self, params).encode('utf-8')
+
+
+def render_paper_page(paper, view_class=None):
+    comment_form = PaperCommentForm(paper=paper, user=session.user, formdata=None)
+    review_form = None
+    reviewed_for_tracks = list(paper.last_revision.get_reviewed_for_groups(session.user))
+    if len(reviewed_for_tracks) == 1:
+        review_form = build_review_form(paper, reviewed_for_tracks[0])
+    params = {'paper': paper,
```
ThiefMaster: probably fits in one line

indico/modules/events/papers/models/revisions.py
```diff
@@ -152,3 +158,27 @@ def paper(self):
...
+        reviews = []
+        if user and group:
+            reviews = [x for x in self.reviews if x.group.instance == group and x.user == user]
```
nop33: because I think it looks better this way! I considered that too, but I prefer 1 return statement rather than 4, makes it cleaner imo.

indico/modules/events/papers/operations.py
```diff
@@ -180,3 +183,85 @@ def delete_paper_template(template):
...
+def create_review(paper, type_, user, review_data, questions_data):
+    review = PaperReview(revision=paper.last_revision, type=type_.instance, user=user)
+    review.populate_from_dict(review_data)
+    log_data = {}
+    for question in paper.event_new.cfp.get_questions_from_review_type(type_.instance):
+        value = int(questions_data['question_{}'.format(question.id)])
+        review.ratings.append(PaperReviewRating(question=question, value=value))
+        log_data[question.text] = value
+    db.session.flush()
+    logger.info("Paper %s received a review of type %s by %s", paper, type_, user)
```
nop33: Actually, I need `%r` for all of them since they are objects and not strings

indico/modules/events/papers/models/revisions.py
ThiefMaster: mkay, matter or preference i guess

indico/modules/events/papers/operations.py
ThiefMaster: `type_` looks like an enum member which has a  decent-ish string version. However, I'd use its `.name` here (with `%s`)

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -43,6 +43,7 @@
 
 class AbstractState(RichIntEnum):
     __titles__ = [None, _("Submitted"), _("Withdrawn"), _("Accepted"), _("Rejected"), _("Merged"), _("Duplicate")]
+    __css_classes__ = [None, '', 'outline dashed',  'success', 'error', 'visited', 'strong']
```
mvidalgarcia: Extra white space before `'success'`

indico/modules/events/papers/models/papers.py
pferreir: Discussed on IRC, will provide logfiles 😉

nop33: logfiles already provided :stuck_out_tongue_winking_eye:

indico/modules/events/models/reviews.py
```diff
@@ -84,44 +103,90 @@ def get_reviewer_render_data(self, user):
...
+    """Classes that represent a review of a proposal should extend this class
```
ThiefMaster: wrap after `"""` since it does not fit in one line. also, this doesn't look like a good "first" line of a docstring at all

I'd go for this instead:

```
"""Mixin for proposal reviews.

Classes that ...
"""
```

indico/modules/events/papers/controllers/display.py
```diff
@@ -19,37 +19,150 @@
...
+from indico.modules.events.papers.views import WPDisplayPapersBase, render_paper_page
+from indico.web.flask.templating import get_template_module
+from indico.web.util import jsonify_form, jsonify_data, jsonify
 
 
-class RHSubmitPaper(RHPapersBase):
+class RHSubmitPaper(RHPaperBase):
+    def _checkProtection(self):
+        RHPaperBase._checkProtection(self)
+        if not self.contribution.can_manage(session.user, role='submit'):
```
ThiefMaster: I don't think we want to check for submission rights here

indico/modules/events/papers/controllers/display.py
```diff
@@ -19,37 +19,150 @@
...
+        if form.validate_on_submit():
+            create_paper_revision(self.contribution, session.user, form.files.data)
+            return jsonify_data(flash=False)
+        return jsonify_form(form, form_header_kwargs={'action': request.relative_url})
+
+
+class RHPaperTimeline(RHPaperBase):
+    def _process(self):
+        comment_form = PaperCommentForm(paper=self.paper, user=session.user, formdata=None)
+        review_form = None
```
ThiefMaster: why not just use `None`  below?

indico/modules/events/papers/controllers/display.py
```diff
@@ -19,37 +19,150 @@
...
+
+
+class RHSubmitPaperComment(RHPaperBase):
+    def _check_paper_protection(self):
+        return self.paper.can_comment(session.user)
+
+    def _process(self):
+        form = PaperCommentForm(paper=self.paper, user=session.user)
+        if form.validate_on_submit():
+            create_comment(self.paper, form.data['text'], form.data['visibility'], session.user)
```
ThiefMaster: `form.text.data, form.visibility.data` - no need to build that dict twice...

indico/modules/events/papers/controllers/display.py
```diff
@@ -19,37 +19,150 @@
...
+class RHEditPaperComment(RHPaperCommentBase):
     def _process(self):
-        form = PaperSubmissionForm()
+        form = PaperCommentForm(obj=self.comment, paper=self.paper, user=session.user,
+                                prefix='edit-comment-{}-'.format(self.comment.id))
         if form.validate_on_submit():
-            create_paper_revision(self.contribution, session.user, form.files.data)
-            return jsonify_data(flash=False)
-        return jsonify_form(form, form_header_kwargs={'action': request.relative_url})
+            update_comment(self.comment, form.data['text'], form.data['visibility'])
```
ThiefMaster: `form.*.data`

indico/modules/events/papers/forms.py
```diff
@@ -18,21 +18,64 @@
...
 from indico.web.forms.fields import (PrincipalListField, IndicoDateTimeField, IndicoTagListField, HiddenEnumField,
-                                     HiddenFieldList, IndicoMarkdownField, FileField, EditableFileField)
+                                     HiddenFieldList, IndicoMarkdownField, FileField, EditableFileField,
+                                     IndicoEnumSelectField)
 from indico.web.forms.util import inject_validators
 from indico.web.forms.validators import LinkedDateTime
+from indico.web.forms.widgets import JinjaWidget
+
+
+def make_review_form(event, type_):
```
ThiefMaster: why don't we call it `review_type`? seems clearer than `type_`

indico/modules/events/papers/forms.py
```diff
@@ -18,21 +18,64 @@
...
+    """Extends the paper WTForm to add the extra fields.
+
+    Each extra field will use a field named ``custom_ID``.
+
+    :param event: The `Event` for which to create the paper review form.
+    :param type_: The `PaperReviewType` for which to create the paper review form.
+    :return: A `PaperForm` subclass.
```
ThiefMaster: `PaperReviewForm`

indico/modules/events/papers/controllers/display.py
```diff
@@ -17,39 +17,145 @@
...
+    }
+
+    def _check_paper_protection(self):
+        if self.paper.last_revision.get_reviews(user=session.user, group=self.type_.instance):
+            return False
+        return self.paper.can_review(session.user, check_state=True)
+
+    def _checkParams(self, params):
+        RHPaperBase._checkParams(self, params)
+        self.type_ = PaperTypeProxy(PaperReviewType(request.view_args['type_id']))
```
ThiefMaster: No need for the underscore since `self.type` wouldn't shadow `type()` - and I'd go for `review_type` here too (same for other places which I'm too lazy to look for now ;))

indico/modules/events/papers/models/reviews.py
```diff
@@ -39,12 +40,35 @@ class PaperReviewType(RichIntEnum):
     content = 2
 
 
-class PaperReview(RenderModeMixin, db.Model):
+class PaperTypeProxy(ProposalGroupProxy):
+    @locator_property
+    def locator(self):
+        return {'type_id': self.instance.value}
```
ThiefMaster: `type_id` is a good name IMO, it's not an ID. `review_type`?

PR #2796, opened 2017-01-19
Add paper proxy class

OmeGak:
We should try to merge this as soon as possible.

### DIFF
--- a/indico/core/settings/proxy.py
+++ b/indico/core/settings/proxy.py
@@ -203,3 +203,19 @@ class SettingProperty(object):
 
     def __delete__(self, obj):
         self.proxy.delete(*self._make_args(obj, self.name))
+
+
+class AttributeProxyProperty(object):
+    def __init__(self, attr):
+        self.attr = attr
+
+    def __get__(self, obj, objtype=None):
+        if obj is None:
+            return self
+        return getattr(getattr(obj, obj.proxied_attr), self.attr)
+
+    def __set__(self, obj, value):
+        setattr(getattr(obj, obj.proxied_attr), self.attr, value)
+
+    def __delete__(self, obj):
+        delattr(getattr(obj, obj.proxied_attr), self.attr)
--- a/indico/modules/events/contributions/models/contributions.py
+++ b/indico/modules/events/contributions/models/contributions.py
@@ -34,8 +34,10 @@ from indico.core.db.sqlalchemy.util.queries import increment_and_get
 from indico.core.db.sqlalchemy.util.session import no_autoflush
 from indico.modules.events.management.util import get_non_inheriting_objects
 from indico.modules.events.models.persons import PersonLinkDataMixin, AuthorsSpeakersMixin
+from indico.modules.events.papers.models.papers import Paper
 from indico.modules.events.papers.models.revisions import PaperRevisionState, PaperRevision
 from indico.modules.events.sessions.util import session_coordinator_priv_enabled
+from indico.util.caching import memoize_request
 from indico.util.locators import locator_property
 from indico.util.string import format_repr, return_ascii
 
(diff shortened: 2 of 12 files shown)


indico/modules/events/contributions/models/contributions.py
```diff
@@ -426,6 +428,11 @@ def locator(self):
     def verbose_title(self):
         return '#{} ({})'.format(self.friendly_id, self.title)
 
+    @property
+    @memoize_request
+    def paper(self):
+        return Paper(self) if self._paper_revisions else None
```
ThiefMaster: this looks like performance hell. it requires you to always load that relationship to do *anything* PR related. I'd either use `_paper_last_revision` (you never need to load all revisions on a page listing many papers) or even unconditionally `return Paper(self)`

indico/modules/events/papers/models/papers.py
```diff
@@ -0,0 +1,68 @@
...
+from __future__ import unicode_literals
+
+from indico.core.settings.proxy import ProxyProperty
+from indico.modules.events.models.reviews import ProposalMixin
+from indico.util.locators import locator_property
+from indico.util.string import return_ascii
+
+
+class Paper(ProposalMixin):
+    """Proxy class for facilitate access to all paper-related properties"""
```
ThiefMaster: "for access" or "to facilitate access" (but that sounds too enterprisey imo)

indico/core/settings/proxy.py
```diff
@@ -203,3 +203,19 @@ def __set__(self, obj, value):
 
     def __delete__(self, obj):
         self.proxy.delete(*self._make_args(obj, self.name))
+
+
+class ProxyProperty(object):
```
pferreir: Not sure about the naming:
  * SettingProperty - proxy property that maps to a setting;
  * ProxyProperty - proxy property that maps to an attribute of a proxied attribute;

What about `AttributeProperty`?

indico/modules/events/papers/models/files.py
```diff
@@ -49,11 +50,11 @@ class PaperFile(StoredFileMixin, db.Model):
         nullable=True
     )
 
-    contribution = db.relationship(
+    _contribution = db.relationship(
```
pferreir: Why can't `contribution` be a "first-class" attribute of `PaperFile`? We've also got `subcontrib.event_new == subcontrib.contribution.event_new`... I still don't see any problem with that.

indico/modules/events/papers/models/revisions.py
```diff
@@ -95,11 +96,11 @@ class PaperRevision(RenderModeMixin, db.Model):
         default=''
     )
 
-    contribution = db.relationship(
+    _contribution = db.relationship(
```
pferreir: Same. I can't see any problem with `revision.contribution`.

indico_zodbimport/modules/event_papers.py
```diff
@@ -304,7 +304,7 @@ def _migrate_revisions(self, contribution, rm):
                 review = self._migrate_review(contribution, old_layout_review, PaperReviewType.layout)
                 revision.reviews.append(review)
                 review_colors += _review_color(review, 'L')
-            contribution.paper_revisions.append(revision)
+            contribution._paper_revisions.append(revision)
```
pferreir: Maybe we should write `contribution.paper.revisions` instead?

nop33: We thought about it, but since it's the zodb import and we only run it once we kept it simple.

indico/core/settings/proxy.py
nop33: > ProxyProperty - proxy property that maps to an attribute of a proxied attribute;

Isn't it more like "proxy property that maps to an attribute of a proxied **class**?"

In any case, following the `SettingProperty` naming it's true that `AttributeProperty` keeps consistency. OTOH, it's very hard to imagine what a class named `AttributeProperty` does, at least for me. I'd suggest `AttributeProxyProperty` (which would also require the renaming of `SettingProperty` to `SettingProxyProperty`).

indico/modules/events/papers/models/files.py
nop33: No strong opinion here. On one hand, It might be considered cleaner to access the parent (`Contribution`) only from it's "virtual" direct children (`Paper`), on the other hand it doesn't harm since it's technically an ancestor of both (`Paper` and `PaperFile`).

PR #2798, opened 2017-01-20
Move general settings & social to new backend

ThiefMaster:
admin-settings


ThiefMaster:
@pferreir What's your opinion on adding another menu item "Admins" in "User Management" that shows the list of admins and lets you add/remove them (instead of showing them on the general settings page)?

nop33:
> @pferreir What's your opinion on adding another menu item "Admins" in "User Management" that shows the list of admins and lets you add/remove them (instead of showing them on the general settings page)?

I like the idea.

indico/MaKaC/services/implementation/error.py
```diff
@@ -19,8 +19,9 @@
 from werkzeug.urls import url_parse
 
 from MaKaC.services.implementation.base import ParameterManager, ServiceBase
-from MaKaC.webinterface.mail import GenericMailer, GenericNotification
+from MaKaC.common.mail import GenericMailer
```
nop33: Should go above `MaKaC.services` but I guess the file is going to be rewritten.

nop33:
Personally, I think the following view looks better and it's more consistent with our current interface. If the settings forms are not in a dialog, then they are editable right away when visiting the corresponding pages.

image

ThiefMaster:
Except that these are settings (like many other in the admin area) that are rarely, if ever, changed. So to me it seems cleaner when the forms only show up when you actually want to edit them...

PR #2802, opened 2017-01-24
Adapt conference pages to new template layout

mvidalgarcia:
Following https://github.com/indico/indico/wiki/Template-layout-(WIP)

Several changes to fulfill our sass-lint rules were made.

### DIFF
--- a/indico/htdocs/sass/modules/event_display/_conferences.scss
+++ b/indico/htdocs/sass/modules/event_display/_conferences.scss
@@ -26,18 +26,17 @@
 }
 
 .time-data, .location-data, .board-data {
+    line-height: 1.6em;
+    font-size: 1.2em;
+    color: $light-black;
+
     .header-data {
-        &:before {
+        &::before {
             color: $orange;
             vertical-align: middle;
             margin-right: 0.2em;
         }
     }
-
-    line-height: 1.6em;
-
-    font-size: 1.2em;
-    color: $light-black;
 }
 
 .conference-page {
@@ -46,27 +45,28 @@
     .item-text {
         clear: both;
         color: $black;
+        margin-bottom: 2em;
     }
 
     .page-content h3 {
+        margin: 1em 0 0.5em 0;
+        color: $orange;
+
         &:first-child {
             margin-top: 0;
         }
 
-        margin: 1em 0 0.5em 0;
-        color: $orange;
-
-        &[class^="icon-"]:before,
-        &[class*=" icon-"]:before {
+        &[class^='icon-']::before,
+        &[class*=' icon-']::before {
             margin-right: 5px;
         }
     }
 
-    #other-fields {
+    .other-fields {
         margin-left: 2em;
 
         th {
-            @include ellipsis;
+            @include ellipsis();
 
             color: $orange;
             text-align: right;
@@ -84,10 +84,6 @@
         margin-top: 0;
     }
 
-    .item-text {
-        margin-bottom: 2em;
-    }
-
     .item-subitems {
         margin-bottom: 1em;
         margin-top: 2em;
@@ -103,7 +99,7 @@
             list-style-type: none;
 
             li {
-                @include ellipsis;
+                @include ellipsis();
 
                 margin: 5px 0;
                 padding: 0 5px;
@@ -161,10 +157,16 @@
                 font-size: 1.5em;
                 font-weight: normal;
 
-                &[class^="icon-"]:before,
-                &[class*=" icon-"]:before {
+                &[class^='icon-']::before,
+                &[class*=' icon-']::before {
                     margin-right: 5px;
                 }
+            }
+
+            h4 {
+                color: $light-black;
+                font-size: 1.1em;
+                margin-top: 0.5em;
             }
         }
     }
@@ -184,18 +186,19 @@
 
         .description {
             position: relative;
-            &:before {
+            color: $dark-gray;
+            max-height: 4.2em;
+            overflow: hidden;
+
+            &::before {
                 content: '';
                 width: 100%;
                 height: 100%;
                 left: 0;
                 top: 0;
                 position: absolute;
-                background: linear-gradient(transparent 0.4em, white 4.2em)
+                background: linear-gradient(transparent 0.4em, white 4.2em);
             }
-            color: $dark-gray;
-            max-height: 4.2em;
-            overflow: hidden;
         }
 
         &:hover {
@@ -206,7 +209,7 @@
 
 .contrib-session, .contrib-track, .contrib-type, .contrib-state {
     @include border-radius();
-    @include ellipsis;
+    @include ellipsis();
 
     display: inline-block;
     margin-left: 5px;
@@ -251,11 +254,13 @@ ul.author-list {
         margin-bottom: 0.5em;
 
         .affiliation {
+            font-size: 1em;
+
             .text {
                 max-width: 30em;
             }
-            font-size: 1em;
         }
+
         .name {
             font-size: 1.5em;
         }
@@ -271,6 +276,7 @@ ul.author-list {
 
         > li {
             margin-bottom: 0.5em;
+
             .contrib-role {
                 color: $light-black;
             }
--- a/indico/htdocs/sass/partials/_boxes.scss
+++ b/indico/htdocs/sass/partials/_boxes.scss
@@ -41,7 +41,7 @@
     height: 0;
     overflow: hidden;
     background: transparent;
-    border: 0;
+    border: none;
     margin-left: -10px;
     margin-right: -10px;
 
@@ -64,8 +64,8 @@
     background-color: lighten($dark-gray, $color-variation);
     cursor: move;
 
-    &:before,
-    & > .handle:before {
+    &::before,
+    > .handle::before {
         color: $light-gray;
     }
 
@@ -73,8 +73,8 @@
     &:hover {
         background-color: white;
 
-        &:before,
-        & > .handle:before {
+        &::before,
+        > .handle::before {
             color: $dark-gray;
         }
     }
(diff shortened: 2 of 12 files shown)


indico/htdocs/sass/partials/_boxes.scss
```diff
@@ -64,17 +64,17 @@
     background-color: lighten($dark-gray, $color-variation);
     cursor: move;
 
-    &:before,
-    & > .handle:before {
+    &::before,
+    & > .handle::before {
```
pferreir: you don't need the `&` here. You can use `> .handle::before`. Applies in other places below.
I've added it to the [SASS Guidelines](https://github.com/indico/indico/wiki/SASS-Guidelines). /cc @nop33

indico/htdocs/sass/partials/_boxes.scss
```diff
@@ -283,15 +284,15 @@ $i-box-padding: 10px;
 }
 
 .i-box {
-    @include transition;
+    @include transition();
```
pferreir: Is this sass-lint? The SASS spec doesn't forbid the original form.

mvidalgarcia: Yes, it was sass-lint complaining.

indico/htdocs/sass/partials/_boxes.scss
pferreir: Actually, it would be nice to make sass-lint warn against this. But I can't find a rule for it.

indico/htdocs/sass/partials/_boxes.scss
pferreir: OK, no problem. Never mind.

mvidalgarcia: https://github.com/indico/indico/blob/devel/.sass-lint.yml#L24

indico/htdocs/sass/partials/_boxes.scss
nop33: Nice! I updated the wiki to always use `::` instead of `:` as suggested in [MDN], since `:` is a CSS2 syntax and `::` is CSS3 (https://developer.mozilla.org/en/docs/Web/CSS/Pseudo-elements)

indico/modules/events/templates/person_display.html
```diff
@@ -2,14 +2,10 @@
...
-            <td class="i-table caption">
-                {{ caption }}
-            </td>
-            <td class="i-table value">
-                {{ value }}
-            </td>
-        </tr>
+        <div>
+            <label>{{ caption }}</label>
+            <span class="text-normal">{% trans %}{{ value }}{% endtrans %}</span>
```
nop33: It should already be translated I believe.

indico/modules/events/templates/person_display.html
```diff
@@ -20,26 +16,36 @@
...
-                </tr>
-            {%- endfor %}
-        </table>
+        <section>
+            <div class="header">
+                <div class="header-row">
+                    <h3>
+                        {%- trans count=contribs|length -%}
+                            Author in the following contribution
+                            {%- pluralize -%}
```
nop33: Too much indentation

indico/modules/events/templates/person_display.html
ThiefMaster: And it wouldn't work properly anyway, since there is no way of extracting the actual value of `value` for translation. Text inside `trans` or `_()` must be fully static

indico/htdocs/sass/partials/_boxes.scss
```diff
@@ -407,10 +403,42 @@ $i-box-padding: 10px;
 
     > li {
         @include border-top();
-        display: block;
-        padding: 10px 10px;
+        padding: 10px;
         position: relative;
         overflow: hidden;
+
+        /* Sortable row with handle on the left */
```
ThiefMaster: Use a `//` comment - no need to include such comments in the final CSS (`//` are scss comments that are skipped when compiling it)

mvidalgarcia: So maybe we should enable the rule `no-css-comments` in the sass-lint file. It seems to fit exactly with your suggestion! Info: https://github.com/sasstools/sass-lint/blob/master/docs/rules/no-css-comments.md

indico/htdocs/sass/partials/_boxes.scss
OmeGak: Things like `::` can be enforced by `sass-lint` so we don't need to add them in the wiki. Let's keep it only for things that can be checked programmatically.

indico/htdocs/sass/partials/_boxes.scss
OmeGak: The only problem is the license header, that it uses `/* */`. Maybe we could update the script to maintain the headers to use `//` in `.scss` files and enable `no-css-comments` in `sass-lint`.

ThiefMaster: Yes, I already have the change for that on my git stash (and iirc also mentioned in the PR where I added the sass-lint config). But maybe let's wait either until we rebase devel on top of master the next time or until we merge devel into master to avoid having to do it in two places?

indico/modules/events/contributions/templates/display/contribution_display.html
```diff
@@ -126,104 +124,111 @@
...
                             Co-authors
                         {%- endtrans %}
                     </h3>
-                    {{ render_author_list(contribution.secondary_authors) }}
                 </div>
-            {% endif %}
-        </div>
-    </div>
+            </div>
+        {{ render_author_list(contribution.secondary_authors) }}
```
OmeGak: Missing some indentation here.

indico/htdocs/sass/partials/_boxes.scss
OmeGak: Let's just add a reminder somewhere. (Maybe in the *Projects* tab?)

indico/htdocs/sass/modules/event_display/_conferences.scss
```diff
@@ -161,11 +157,17 @@
...
 
-                &[class^="icon-"]:before,
-                &[class*=" icon-"]:before {
+                &[class^='icon-']::before,
+                &[class*=' icon-']::before {
                     margin-right: 5px;
                 }
             }
+
+            h4 {
```
nop33: We discussed that this would be a nice addition to our conference page layout template. Currently we don't account on an `h4` but I believe we should standardize how the `h4`s are displayed the same way we do with the `h3`s. That's why I'd like to suggest to move this code to `_conferences.scss` inside the `.header-row` selector.

indico/htdocs/sass/partials/_boxes.scss
mvidalgarcia: It's also possible to change the license header to use bang comments and will be ignored by the rule.
```
/*!
  * This is a good bang comment
**/
```
 I don't know if these bang comments have an specific use in SASS files though.

ThiefMaster: i wonder if sass is smart enough to strip all but one of these comments. otherwise we add tons of noise when keeping all license headers in the generated css file

ThiefMaster: i wonder if sass is smart enough to strip all but one of these comments. otherwise we add tons of noise when keeping all license headers in the generated css file

indico/htdocs/sass/modules/event_display/_conferences.scss
mvidalgarcia: Is it in the right place @nop33?

nop33: I am sorry, you're right, that's the right place. I mistook the diff!

indico/htdocs/sass/partials/_boxes.scss
```diff
@@ -401,16 +397,48 @@ $i-box-padding: 10px;
...
         @include border-top();
-        display: block;
-        padding: 10px 10px;
+        padding: 10px;
         position: relative;
         overflow: hidden;
+
+        // Sortable row with handle on the left
+        &.ui-sortable {
+            @extend .flexrow;
```
nop33: Sorry I didn't notice it before. When implementing the flexbox framework for Indico, I thought it would be a good idea to extend from the `.flexrow` and `.flexcol` classes. Later I realized that it creates a specificity hell so I stopped using it. It's better to apply `display: flex` instead.

indico/htdocs/sass/partials/_boxes.scss
```diff
@@ -401,16 +397,48 @@ $i-box-padding: 10px;
...
+            margin: 0;
+
+            > .ui-sortable-handle {
+                @extend %i-box-sortable-handle-background;
+                @extend .flexrow;
```
nop33: Same here

indico/modules/events/contributions/templates/display/user_contribution_list.html
```diff
@@ -5,41 +5,37 @@
...
+                                       data-title="{% trans title=contrib.title %}Edit contribution '{{ title }}'{% endtrans %}"
+                                       data-href="{{ url_for('.manage_update_contrib', contrib) }}"
+                                       data-ajax-dialog
+                                       data-reload-after></a>
+                                {%- endif %}
+                            </div>
+                        </li>
+                    {% else %}
+                        <li>
+                            {%- trans %}You don't have any contributions at the moment.{% endtrans -%}
```
nop33: This could nicely fit in one like (like you did in `contribution_display.html` for the `<h3>`), but that's just some personal preferences, feel free to use your own style! But then be consistent :stuck_out_tongue_winking_eye:

indico/modules/events/sessions/templates/display/session_display.html
```diff
@@ -8,11 +8,11 @@
 {% block page_class %}conference-page item-summary{% endblock %}
 
 {% block title %}
-    {% trans %}Session{% endtrans %}
+    {%- trans %}Session{% endtrans -%}
```
nop33: I :heart: you

indico/modules/events/sessions/templates/display/session_list.html
```diff
@@ -5,50 +5,44 @@
...
+                                       data-reload-after></a>
+                                {% endif %}
+                                <a href="{{ url_for('timetable.manage_session', sess) }}"
+                                   class="icon-time"
+                                   title="{% trans %}Manage timetable{% endtrans %}"></a>
+                            </div>
+                        </li>
+                    {% else %}
+                        <li>
+                            {%- trans %}You don't have any sessions at the moment.{% endtrans -%}
```
nop33: Same as I commented above. You can avoid using the `-` this way.

PR #2811, opened 2017-01-30
Unify paper judging and assignment areas

ThiefMaster:
Supersedes #2788


indico/modules/events/papers/templates/management/teams.html
```diff
@@ -0,0 +1,11 @@
+{% from 'forms/_form.html' import simple_form %}
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('warning') %}
+    {% trans %}
```
nop33: OK, I agree that everyone has their own style, but com'on! The content of the translatable string contains a lot of whitespace, that's even worse than forgetting the whitespace around the translatable string :stuck_out_tongue_closed_eyes:

ThiefMaster: oops ;)

PR #2801, opened 2017-01-24
Paper reviewing area

DavidAndreev:

PR #2814, opened 2017-01-31
Paper reviewing and judging deadlines

DavidAndreev:


indico/modules/events/papers/blueprint.py
```diff
@@ -34,6 +34,12 @@
 
 # Management
 _bp.add_url_rule('/manage/papers/', 'management', management.RHPapersDashboard)
+_bp.add_url_rule('/manage/papers/deadlines/judging', 'manage_judging_deadline', management.RHSetJudgingDeadline,
```
ThiefMaster: Why separate RHs? Could use `/deadlines/<any(judging,content,layout):type>` and then use `request.view_args['type']` in the RH

indico/modules/events/papers/controllers/management.py
```diff
@@ -253,3 +254,45 @@ def _process(self):
...
         return jsonify_form(form)
+
+
+class RHSetDeadlineBase(RHManagePapersBase):
+    def _process(self):
+        deadline = paper_reviewing_settings.get(self.event_new, self.role)
+        enforce = paper_reviewing_settings.get(self.event_new, 'enforce_{}'.format(self.role))
+        form = DeadlineForm(obj=FormDefaults(deadline=deadline, enforce=enforce), event=self.event_new)
+        if form.validate_on_submit():
+            if self.role == 'layout_reviewer_deadline':
```
ThiefMaster: This feels super dirty. How about reusing the PaperReviewingRole enum?

indico/modules/events/papers/blueprint.py
ThiefMaster: Or `<any(judge,content_reviewer,layout_reviewer):role>` if you reuse the enum (as I suggest in a comment below)

indico/modules/events/papers/controllers/management.py
```diff
@@ -253,3 +254,45 @@ def _process(self):
...
+                text = 'layout reviewing'
+            elif self.role == 'content_reviewer_deadline':
+                text = 'content reviewing'
+            elif self.role == 'judge_deadline':
+                text = 'judging'
+            modified = (paper_reviewing_settings.get(self.event_new, self.role) is not None or
+                        paper_reviewing_settings.get(self.event_new,
+                                                     'enforce_{}'.format(self.role)) is not form.data.get('enforce'))
```
ThiefMaster: `is not` is wrong here, you'd want `!=`.
But do we really need different flash messages depending on whether the deadline has been set/changed

indico/modules/events/papers/controllers/management.py
```diff
@@ -253,3 +254,45 @@ def _process(self):
...
+            set_deadline(self.event_new, form.data.get('deadline'), self.role, text, form.data.get('enforce'))
```
ThiefMaster: Why `get()`? These keys always exist so `[]` is more appropriate

indico/modules/events/papers/controllers/management.py
```diff
@@ -253,3 +254,45 @@ def _process(self):
...
+                flash(_("Paper {} deadline has been modified.".format(text)), 'success')
+            else:
+                flash(_("Paper {} deadline has been set.".format(text)), 'success')
+            return jsonify_data(html=_render_paper_dashboard(self.event_new))
+        return jsonify_form(form)
+
+
+class RHSetContentReviewingDeadline(RHSetDeadlineBase):
+    def _process(self):
+        self.role = 'content_reviewer_deadline'
```
ThiefMaster: This is something you'd set on the class level (then you wouldn't need to override `_process` at all)

indico/modules/events/papers/forms.py
```diff
@@ -145,3 +145,12 @@ def __init__(self, *args, **kwargs):
         if template is None:
             inject_validators(self, 'template', [DataRequired()])
         super(PaperTemplateForm, self).__init__(*args, **kwargs)
+
+
+class DeadlineForm(IndicoForm):
+    deadline = IndicoDateTimeField(_("Deadline"), [Optional()], default_time=time(23, 59))
+    enforce = BooleanField(_("Enforce deadline"), default=True)
```
ThiefMaster: switch widget please, and the `default=True` shouldn't be needed since you always provide the current settings when rendering the form

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -120,3 +120,7 @@ def user_competences(self):
     @property
     def announcement(self):
         return MarkdownText(paper_reviewing_settings.get(self.event, 'announcement'))
+
+    @property
+    def paper_reviewing_settings(self):
```
ThiefMaster: Not sure if this makes sense. It's not specific to any event. IF you want to do this, then return `paper_reviewing_settings.bind(self.event)` - that will let you access the settings without passing the event again to get(), set() etc.

indico/modules/events/papers/operations.py
```diff
@@ -180,3 +180,14 @@ def delete_paper_template(template):
     db.session.delete(template)
     db.session.flush()
     logger.info('Paper template %r deleted by %r', template, session.user)
+
+
+def set_deadline(event, deadline, role, text, enforce):
+    event.cfp.paper_reviewing_settings.set(event, role, deadline)
+    event.cfp.paper_reviewing_settings.set(event, 'enforce_{}'.format(role), enforce)
+    log_data = {'Enforced': enforce}
+    if deadline:
```
ThiefMaster: I'd always log it but use `None` if there's no deadline

indico/modules/events/papers/operations.py
```diff
@@ -180,3 +180,14 @@ def delete_paper_template(template):
     db.session.delete(template)
     db.session.flush()
     logger.info('Paper template %r deleted by %r', template, session.user)
+
+
+def set_deadline(event, deadline, role, text, enforce):
```
ThiefMaster: the `text` argument really doesn't belong here

indico/modules/events/papers/operations.py
```diff
@@ -180,3 +180,14 @@ def delete_paper_template(template):
...
+    event.cfp.paper_reviewing_settings.set(event, role, deadline)
+    event.cfp.paper_reviewing_settings.set(event, 'enforce_{}'.format(role), enforce)
+    log_data = {'Enforced': enforce}
+    if deadline:
+        log_data.update({'Deadline': deadline.isoformat()})
```
ThiefMaster: `log_data['Deadline'] = deadline.isoformat()` does the same and is much cleaner

indico/modules/events/papers/templates/management/overview.html
```diff
@@ -11,22 +12,35 @@
     {% endif %}
 
     <div class="action-box">
-        <div class="section semantic-text {{ 'highlight' if event.cfp.content_reviewing_enabled }}">
+        <div class="section semantic-text {{ 'highlight' if cfp.content_reviewing_enabled }}">
             <div class="icon icon-file-content"></div>
             <div class="text">
                 <div class="label">{% trans %}Content reviewing{% endtrans %}</div>
-                {% if event.cfp.content_reviewing_enabled %}
+                {% set content_deadline = cfp.paper_reviewing_settings.get(event,'content_reviewer_deadline') %}
```
ThiefMaster: space after `,`

ThiefMaster: Also, why not expose the deadline through a `SettingProperty` on the CFP object?

indico/modules/events/papers/models/call_for_papers.py
DavidAndreev: Well I think its better to put the whole settings there rather then each setting that we need separately, I also added it in reviewing area [#2801](https://github.com/indico/indico/pull/2801/commits/eceec90006c294bb92f94a7578c1675aca2aa9ce)  to display reviewing deadlines.  But yeah I can add that bind thing.

DavidAndreev: No, you' re right, I' ll do each setting separately ;D

indico/modules/events/papers/operations.py
```diff
@@ -238,3 +238,16 @@ def update_reviewing_roles(event, users, contributions, role, assign):
                   'Papers unassigned ({})'.format(orig_string(role.title)), session.user, data=log_data)
     db.session.flush()
     logger.info('Paper reviewing roles in event %r updated by %r', event, session.user)
+
+
+def set_deadline(event, deadline, role, enforce):
+    paper_reviewing_settings.set(event, '{}_deadline'.format(role), deadline)
```
ThiefMaster: you could use `set_multi()` here and pass a dict with the two settings

indico/modules/events/papers/operations.py
```diff
@@ -238,3 +238,16 @@ def update_reviewing_roles(event, users, contributions, role, assign):
...
+    paper_reviewing_settings.set(event, 'enforce_{}_deadline'.format(role), enforce)
+    log_data = {'Enforced': enforce}
+    if deadline:
+        log_data['Deadline'] = deadline.isoformat()
```
ThiefMaster: `log_data['Deadline'] = deadline.isoformat() if deadline else None` - no need to spread this over 4 lines

indico/modules/events/papers/operations.py
```diff
@@ -238,3 +238,16 @@ def update_reviewing_roles(event, users, contributions, role, assign):
...
+    else:
+        log_data['Deadline'] = 'None'
+    logger.info("Deadline for paper reviewing role '{}' in event %r set by %r".format(role), event, session.user)
```
ThiefMaster: `'Paper reviewing deadline (%s) set in %r by %r'` - this goes into a logfile, so being concise is perfectly fine. Also, don't use `.format()` for logging!

indico/modules/events/papers/operations.py
```diff
@@ -238,3 +238,16 @@ def update_reviewing_roles(event, users, contributions, role, assign):
...
+    event.log(EventLogRealm.management, EventLogKind.change, 'Papers',
+              "Deadline for paper reviewing role '{}' has been set".format(role), session.user, data=log_data)
```
ThiefMaster: `Paper reviewing deadline ({}) set`

indico/modules/events/papers/templates/management/overview.html
```diff
@@ -2,6 +2,7 @@
 
 {% block content %}
     {% set cfp = event.cfp %}
+    {% set tz = event.timezone %}
```
ThiefMaster: I'd use `event.tzinfo` (same behavior, it's just using the tzinfo object instead of the timezone string)

indico/modules/events/papers/templates/management/overview.html
```diff
@@ -11,22 +12,35 @@
...
 
     <div class="action-box">
-        <div class="section semantic-text {{ 'highlight' if event.cfp.content_reviewing_enabled }}">
+        <div class="section semantic-text {{ 'highlight' if cfp.content_reviewing_enabled }}">
             <div class="icon icon-file-content"></div>
             <div class="text">
                 <div class="label">{% trans %}Content reviewing{% endtrans %}</div>
-                {% if event.cfp.content_reviewing_enabled %}
+                {% set content_deadline = cfp.content_reviewer_deadline %}
+                {% if cfp.content_reviewing_enabled and content_deadline %}
```
ThiefMaster: By reordering it like this you could avoid checking whether it's enabled twice:
```
if not cfp.content_reviewing_enabled
elif content_deadline
else
```

indico/modules/events/papers/templates/management/overview.html
```diff
@@ -77,6 +104,30 @@
                 {% endif %}
             </div>
         </div>
+        <div class="section semantic-text">
```
ThiefMaster: no need for `semantic-text` here

indico/modules/events/papers/templates/management/overview.html
```diff
@@ -77,6 +104,30 @@
...
+            <div class="icon icon-hammer"></div>
+            <div class="text">
+                <div class="label">{% trans %}Paper judging{% endtrans %}</div>
+                {% set judge_deadline = cfp.judge_deadline %}
```
ThiefMaster: you could use `deadline` for all three places where you do this. I guess you do it to have something shorter, but in the trans line it's still quite long so it using just `deadline` would improve readability

indico/modules/events/papers/controllers/management.py
```diff
@@ -179,3 +180,21 @@ def _process(self):
...
+        deadline = paper_reviewing_settings.get(self.event_new, '{}_deadline'.format(role.name))
+        enforce = paper_reviewing_settings.get(self.event_new, 'enforce_{}_deadline'.format(role.name))
+        form = DeadlineForm(obj=FormDefaults(deadline=deadline, enforce=enforce), event=self.event_new)
+        if form.validate_on_submit():
+            messages = {
+                PaperReviewingRole.content_reviewer: _('Content reviewing deadline has been set.'),
+                PaperReviewingRole.layout_reviewer: _('Layout reviewing deadline has been set.'),
+                PaperReviewingRole.judge: _('Judging deadline has been set.')
+            }
+            set_deadline(self.event_new, form.data['deadline'], role.name, form.data['enforce'])
```
ThiefMaster: `form.deadline.data` and `form.enforce.data`

ThiefMaster: I'd move this line before the `messages = {...}` block

indico/modules/events/papers/templates/management/overview.html
```diff
@@ -44,22 +58,35 @@
...
-        <div class="section semantic-text {{ 'highlight' if event.cfp.layout_reviewing_enabled }}">
+        <div class="section semantic-text {{ 'highlight' if cfp.layout_reviewing_enabled }}">
             <div class="icon icon-layout"></div>
             <div class="text">
                 <div class="label">{% trans %}Layout reviewing{% endtrans %}</div>
-                {% if event.cfp.layout_reviewing_enabled %}
-                    {%- trans %}The layout reviewing process is enabled{% endtrans -%}
-                {% else %}
+                {% set deadline = cfp.layout_reviewer_deadline %}
+                {% if not cfp.layout_reviewing_enabled%}
```
ThiefMaster: missing space before `%}`

indico/modules/events/papers/templates/management/overview.html
```diff
@@ -77,6 +104,30 @@
...
+            <div class="icon icon-hammer"></div>
+            <div class="text">
+                <div class="label">{% trans %}Paper judging{% endtrans %}</div>
+                {% set deadline = cfp.judge_deadline %}
+                {% if deadline %}
+                    {% trans date=deadline|format_date(timezone=tz), time=deadline|format_time(timezone=tz) -%}
+                        The paper judging ends on <strong>{{ date }}</strong> at <strong>{{ time }}</strong>.
+                    {%- endtrans %}
+                {% else %}
+                    {%- trans %}The deadline for paper judging has not been set.{% endtrans -%}
```
ThiefMaster: "There is no deadline for paper judging."? Otherwise it sounds like it's something you are supposed to change

indico/modules/events/papers/templates/management/overview.html
```diff
@@ -77,6 +104,30 @@
...
             </div>
         </div>
+        <div class="section">
+            <div class="icon icon-hammer"></div>
+            <div class="text">
+                <div class="label">{% trans %}Paper judging{% endtrans %}</div>
+                {% set deadline = cfp.judge_deadline %}
+                {% if deadline %}
+                    {% trans date=deadline|format_date(timezone=tz), time=deadline|format_time(timezone=tz) -%}
+                        The paper judging ends on <strong>{{ date }}</strong> at <strong>{{ time }}</strong>.
```
ThiefMaster: "Paper judging ends ..." (without "the")?

indico/modules/events/papers/operations.py
```diff
@@ -322,3 +322,14 @@ def delete_comment(comment):
     paper = comment.paper_revision.paper
     paper.event_new.log(EventLogRealm.management, EventLogKind.negative, 'Papers',
                         'Comment on paper {} removed'.format(paper.verbose_title), session.user)
+
+
+def set_deadline(event, deadline, role, enforce):
+    paper_reviewing_settings.set_multi(event, {
+            '{}_deadline'.format(role): deadline,
```
ThiefMaster: indentation

indico/modules/events/papers/operations.py
```diff
@@ -322,3 +322,14 @@ def delete_comment(comment):
     paper = comment.paper_revision.paper
     paper.event_new.log(EventLogRealm.management, EventLogKind.negative, 'Papers',
                         'Comment on paper {} removed'.format(paper.verbose_title), session.user)
+
+
+def set_deadline(event, deadline, role, enforce):
```
ThiefMaster: I'd reorder the arguments like this: `event, role, deadline, enforce=True`

PR #2821, opened 2017-02-03
Fix event overview when events expand through days

nop33:
The issue was in the calculation of days between the start and end
datetime of the event. When the time of the event end datetime was
prior to the time of the start datetime, the last day was not
included in the calculation (ex: 1/1/16 14:00, 3/1/16 10:00)

### DIFF
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -16,7 +16,7 @@
 
 from __future__ import unicode_literals
 
-from datetime import datetime, timedelta, date
+from datetime import datetime, timedelta, date, time
 from functools import partial
 from io import BytesIO
 from itertools import chain, groupby, imap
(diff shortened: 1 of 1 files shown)


nop33:
Fixes `INC1271332`

indico/modules/categories/controllers/display.py
```diff
@@ -537,9 +537,13 @@ def _get_calendar_days(self):
     def _get_days(start_dt, end_dt):
         # Return all days in the open-ended interval
         current_dt = start_dt
+        event_tz = current_dt.tzinfo
```
ThiefMaster: We don't use `display_tzinfo` on that page?

nop33: How can I see if we use it or not?

nop33: we do, it's just easier to fetch it from there, I don't need to pass an extra arg

ThiefMaster: oh, `current_dt` - i had read it as `event.tzinfo`...

PR #2822, opened 2017-02-03
Allow customizable 'No accommodation' option in registration form

mvidalgarcia:
~I think a script similar to [this](https://gist.github.com/mvidalgarcia/b0fc94bd0a3b65a5dce0cb3b6a5cd92c) should be run to avoid errors in the existing registrations which have accommodation field.~
I think it's no necessary anymore.

image
image
image
image

### DIFF
--- a/indico/htdocs/js/indico/modules/registration/form/field.js
+++ b/indico/htdocs/js/indico/modules/registration/form/field.js
@@ -902,13 +902,13 @@ ndRegForm.directive('ndAccommodationField', function(url) {
                 // Adds `No accommodation` default choice the first time the Accommodation field is added
                 if (!field.choices) {
                     formData.choices.push({
-                        caption: '',
+                        caption: 'No accommodation',
                         isBillable: false,
                         isEnabled: true,
                         isNoAccommodation: true,
                         placesLimit: 0,
                         price: 0,
-                        placeholder: $T('No accommodation...')
+                        placeholder: $T('Title of the "None" option')
                     });
                 }
             };
--- a/indico/htdocs/js/indico/modules/registration/form/field.js
+++ b/indico/htdocs/js/indico/modules/registration/form/field.js
@@ -896,12 +896,13 @@ ndRegForm.directive('ndAccommodationField', function(url) {
                 // Adds `No accommodation` default choice the first time the Accommodation field is added
                 if (!field.choices) {
                     formData.choices.push({
-                        caption: "No accommodation",
+                        caption: '',
                         isBillable: false,
                         isEnabled: true,
                         isNoAccommodation: true,
                         placesLimit: 0,
-                        price: 0
+                        price: 0,
+                        placeholder: $T('No accommodation...')
                     });
                 }
             };
--- a/indico/htdocs/js/indico/modules/registration/form/field.js
+++ b/indico/htdocs/js/indico/modules/registration/form/field.js
@@ -821,8 +821,9 @@ ndRegForm.directive('ndAccommodationField', function(url) {
             };
 
             $scope.areAccommodationOptionsDefined = function(data) {
+                // `No accommodation` option does not count as an actual option
                 var choices = data.choices.filter(function(item) {
-                    return item.remove !== true;
+                    return item.remove !== true && !item.isNoAccommodation;
                 });
                 return choices.length !== 0;
             };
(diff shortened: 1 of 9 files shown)


indico/modules/events/registration/fields/choices.py
```diff
@@ -358,13 +358,14 @@ def _stay_dates_valid(form, field):
...
                 return
             data = snakify_keys(field.data)
-            try:
-                arrival_date = data['arrival_date']
-                departure_date = data['departure_date']
-            except KeyError:
-                raise ValidationError(_("Arrival/departure date is missing"))
-            if _to_date(arrival_date) > _to_date(departure_date):
-                raise ValidationError(_("Arrival date can't be set after the departure date."))
+            if 'choice' in data and data['choice']:
```
ThiefMaster: `if data.get('choice'):`

indico/modules/events/registration/fields/choices.py
```diff
@@ -413,11 +414,12 @@ def process_form_data(self, registration, value, old_data=None, billable_items_l
             return {}
         data = {}
         if value:
-            data = {
-                'choice': value['choice'],
-                'arrival_date': value['arrivalDate'],
-                'departure_date': value['departureDate']
-            }
+            if 'choice' in value and value['choice']:
```
ThiefMaster: same

indico/modules/events/registration/templates/display/_registration_summary_blocks.html
```diff
@@ -23,22 +23,26 @@
...
-        </div>
-        <div>
-            {% trans accommodation=friendly_data['choice'] -%}
-                <strong>Accommodation:</strong> {{ accommodation }}
-            {%- endtrans %}
-        </div>
+    {% if field.input_type == 'accommodation' %}
+        {% if friendly_data %}
+            <div>
+                {% trans arrival_date=friendly_data['arrival_date'] | format_date -%}
```
ThiefMaster: I know it was like this before, but it'd be a good opportunity to change this to `friendly_data.arrival_date` (same for the other ones below)

indico/htdocs/js/indico/modules/registration/form/tpls/fields/dialogs/base.tpl.html
```diff
@@ -38,7 +38,7 @@
                         ng-class="{hasError: validationStarted && optionsForm.title.$invalid}"
                         nd-init-focus
                         class="caption-input"/>
-                    <span ng-if="settings.isRequired">
+                    <span ng-if="field.inputType != 'accommodation'">
```
OmeGak: I don't think we want this. `accommodation` is not the only field we want to disallow making mandatory. With this change, for instance, `label` fields will be allowed to make mandatory, which doesn't make any sense.

Instead, what we need to do is adding `settings.isRequired = false;` in the `link` method of `ndAccommodationField`.
https://github.com/indico/indico/blob/v1.9.9/indico/htdocs/js/indico/modules/registration/form/field.js#L953

In the same way we do here:
https://github.com/indico/indico/blob/v1.9.9/indico/htdocs/js/indico/modules/registration/form/field.js#L444

OmeGak:
In general it is better to split linter-only fixes into a standalone PR. Otherwise, it makes it difficult to focus on functional changes while reviewing. This PR is small and didn't cause much trouble, but keep it mind.

mvidalgarcia:
@OmeGak: Not even if they are in a separated commit as here? Should I create a new PR each time I have to change just a couple of things related to linter?

ThiefMaster:
You could even ignore them if they are in semi-legacy code. Or ping someone to cherrypick the commit - no need for a PR!

indico/modules/events/registration/templates/display/_registration_summary_blocks.html
```diff
@@ -24,16 +24,18 @@
...
-            {% trans arrival_date=friendly_data['arrival_date'] | format_date -%}
-                <strong>Arrival:</strong> {{ arrival_date }}
-            {%- endtrans %}
-        </div>
-        <div>
-            {% trans departure_date=friendly_data['departure_date'] | format_date -%}
-                <strong>Departure:</strong> {{ departure_date }}
-            {%- endtrans %}
-        </div>
+         {% if not friendly_data['is_no_accommodation'] %}
```
pferreir: `friendly_data.is_no_accommodation`

indico/htdocs/js/indico/modules/registration/form/field.js
```diff
@@ -15,6 +15,8 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
+/*globals angular, ndRegForm */
```
ThiefMaster: /* global angular, ndRegForm */

PR #2835, opened 2017-02-14
Prevent reordering fields in registration form to break

mvidalgarcia:


indico/modules/events/registration/util.py
```diff
@@ -187,6 +187,8 @@ def create_personal_data_fields(regform):
             continue
         field = RegistrationFormPersonalDataField(registration_form=regform, personal_data_type=pd_type,
                                                   is_required=pd_type.is_required)
+        if not data.get('is_enabled', True):
+            field.position = data.get('position')
```
ThiefMaster: Assigning a `None` positions seems odd. Do we ever expect `position` to be missing? If not let's use `data['position']`

indico/modules/events/registration/controllers/management/fields.py
```diff
@@ -64,6 +64,8 @@ def _process(self):
         if (not enabled and self.field.type == RegistrationFormItemType.field_pd and
                 self.field.personal_data_type.is_required):
             raise BadRequest
+        positions = [item.position for item in self.regform.form_items if item.is_enabled == enabled]
+        self.field.position = positions[-1] + 1
```
ThiefMaster: What if there aren't any enabled fields?

indico/modules/events/registration/controllers/management/fields.py
```diff
@@ -64,6 +64,8 @@ def _process(self):
         if (not enabled and self.field.type == RegistrationFormItemType.field_pd and
                 self.field.personal_data_type.is_required):
             raise BadRequest
+        positions = [item.position for item in self.regform.form_items if item.is_enabled == enabled]
```
ThiefMaster: I think you could do this - you are looking for the highest position after all, aren't you?

    self.field.position = max(x.position for x in self.regform.form_items if item.is_enabled == enabled) + 1

However, see the comment below! max() on an empty collection would fail.

PR #2838, opened 2017-02-16
Keep event persons consistent on edit / merge

ThiefMaster:


indico/modules/events/models/persons.py
```diff
@@ -244,7 +262,7 @@ def merge_person_info(self, other):
             if existing_event_link is None:
                 event_link.person = self
             else:
-                db.session.delete(event_link)
+                other.event_links.remove(event_link)
```
ThiefMaster: Changed this because `db.session.delete()` doesn't update the relationship collection, resulting in a warning later when SA tries to delete things that are already deleted.

pferreir: Unrelated, but can an `EventPerson` ever have more than one associated `EventPersonLink`?

ThiefMaster: not for the same event

pferreir: Yeah. Anyway, what I meant is... what's the advantage of having `EventPersonLink` vis a vis a column in `EventPerson` telling us their role?

ThiefMaster: like this it's consistent with what we have in contributions, sessions, etc. also, you can override things on the link level. with just a column this wouldn't be possible

PR #2844, opened 2017-02-22
Paper reviewing email notifications

nop33:
Preview of the emails (text might have changed): http://imgur.com/a/gulmF

### DIFF
--- a/indico/modules/events/papers/controllers/display.py
+++ b/indico/modules/events/papers/controllers/display.py
@@ -198,7 +198,7 @@ class RHJudgePaper(RHPaperBase):
         return self.paper.can_judge(session.user, check_state=True)
 
     def _process(self):
-        form = PaperJudgmentForm()
+        form = PaperJudgmentForm(paper=self.paper)
         if form.validate_on_submit():
             judge_paper(self.paper, form.judgment.data, form.judgment_comment.data, judge=session.user)
             return jsonify_data(flash=False, html=render_paper_page(self.paper))
--- a/indico/modules/events/papers/controllers/display.py
+++ b/indico/modules/events/papers/controllers/display.py
@@ -200,8 +200,7 @@ class RHJudgePaper(RHPaperBase):
     def _process(self):
         form = PaperJudgmentForm()
         if form.validate_on_submit():
-            judge_paper(self.paper, form.judgment.data, form.judgment_comment.data, judge=session.user,
-                        send_notifications=form.send_notifications.data)
+            judge_paper(self.paper, form.judgment.data, form.judgment_comment.data, judge=session.user)
             return jsonify_data(flash=False, html=render_paper_page(self.paper))
 
 
--- a/indico/modules/events/papers/controllers/display.py
+++ b/indico/modules/events/papers/controllers/display.py
@@ -59,6 +59,12 @@ class RHPaperTimeline(RHPaperBase):
     def _process(self):
         return render_paper_page(self.paper, view_class=WPDisplayCallForPapers)
 
+    def _check_paper_protection(self):
+        return (self.contribution.is_user_associated(session.user, check_abstract=True) or
+                self.event_new.cfp.is_manager(session.user) or
+                self.paper.can_review(session.user) or
+                self.paper.can_judge(session.user))
+
 
 class RHDownloadPaperFile(RHPaperBase):
     """Download a paper file"""
--- a/indico/modules/events/papers/notifications.py
+++ b/indico/modules/events/papers/notifications.py
@@ -89,3 +89,12 @@ def notify_paper_assignment(user, role, contributions, event, assign):
                                    receiver=user, assign=assign, role=role)
     email = make_email(to_list=user.email, template=template, html=True)
     send_email(email, event=event, module='Papers', user=session.user)
+
+
+def notify_comment(person, paper, comment, submitter):
+    event = paper.event_new
+    receiver_name = person.first_name or 'user'
+    template = get_template_module('events/papers/emails/comment.html', event=event, receiver=receiver_name,
+                                   contribution=paper.contribution, comment=comment, submitter=submitter)
+    email = make_email(to_list=user.email, template=template, html=True)
+    send_email(email, event=event, module='Papers', user=session.user)
--- a/indico/modules/events/papers/notifications.py
+++ b/indico/modules/events/papers/notifications.py
@@ -96,5 +96,5 @@ def notify_comment(person, paper, comment, submitter):
     receiver_name = person.first_name or 'user'
     template = get_template_module('events/papers/emails/comment.html', event=event, receiver=receiver_name,
                                    contribution=paper.contribution, comment=comment, submitter=submitter)
-    email = make_email(to_list=user.email, template=template, html=True)
+    email = make_email(to_list=person.email, template=template, html=True)
     send_email(email, event=event, module='Papers', user=session.user)
--- a/indico/modules/events/papers/notifications.py
+++ b/indico/modules/events/papers/notifications.py
@@ -1,0 +1,91 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2017 European Organization for Nuclear Research (CERN).
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 3 of the
+# License, or (at your option) any later version.
+#
+# Indico is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from flask import session
+
+from indico.core.notifications import make_email, send_email
+from indico.modules.events.papers.settings import paper_reviewing_settings, PaperReviewingRole
+from indico.web.flask.templating import get_template_module
+
+
+def notify_paper_revision_submission(revision):
+    event = revision.paper.event_new
+    roles_to_notify = paper_reviewing_settings.get(event, 'notify_on_paper_submission')
+    if PaperReviewingRole.judge in roles_to_notify:
+        for judge in revision.paper.contribution.paper_judges:
+            template = get_template_module('events/papers/emails/revision_submission_to_judge.html', event=event,
+                                           revision=revision, receiver=judge)
+            email = make_email(to_list=judge.email, template=template, html=True)
+            send_email(email, event=event, module='Papers', user=session.user)
+    reviewers = set()
+    if PaperReviewingRole.layout_reviewer in roles_to_notify:
+        reviewers |= revision.paper.contribution.paper_layout_reviewers
+    if PaperReviewingRole.content_reviewer in roles_to_notify:
+        reviewers |= revision.paper.contribution.paper_content_reviewers
+    for reviewer in reviewers:
+        template = get_template_module('events/papers/emails/revision_submission_to_reviewer.html', event=event,
+                                       revision=revision, receiver=reviewer)
+        email = make_email(to_list=reviewer.email, template=template, html=True)
+        send_email(email, event=event, module='Papers', user=session.user)
+
+
+def notify_paper_review_submission(review):
+    event = review.revision.paper.event_new
+    if not paper_reviewing_settings.get(event, 'notify_judge_on_review'):
+        return
+    for judge in review.revision.paper.contribution.paper_judges:
+        template = get_template_module('events/papers/emails/review_submission_to_judge.html', event=event,
+                                       review=review, contribution=review.revision.paper.contribution, receiver=judge)
+        email = make_email(to_list=judge.email, template=template, html=True)
+        send_email(email, event=event, module='Papers', user=session.user)
+
+
+def notify_paper_judgment(paper, reset=False):
+    event = paper.event_new
+    receiver = paper.last_revision.submitter
+    if reset:
+        template = 'events/papers/emails/judgment_reset_to_author.html'
+    elif paper_reviewing_settings.get(event, 'notify_author_on_judgment'):
+        template = 'events/papers/emails/judgment_to_author.html'
+    template = get_template_module(template, event=event, paper=paper, contribution=paper.contribution,
+                                   receiver=receiver)
+    email = make_email(to_list=receiver.email, template=template, html=True)
+    send_email(email, event=event, module='Papers', user=session.user)
+
+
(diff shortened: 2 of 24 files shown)


indico/modules/events/papers/templates/emails/base.html
```diff
@@ -0,0 +1,40 @@
...
+{% block footer %}{% endblock %}
+
+{% block subject -%}
+    {% block subject_message %}Call for papers{% endblock %}
+    {%- block subject_name %}{% endblock %}
+{%- endblock %}
+
+{% block body %}
+    <div class="mail" style="width: 600px; font-family: Arial; border: 1px solid #dfdfdf; background-color: #fcfcfc; margin: auto auto 40px auto;">
+        <div class="header" style="background-color: #5D95EA; color: #eee; padding: 10px 20px;">
```
mvidalgarcia: Hexadecimal notation should be lowercase according to our set of sass-lint rules.

indico/modules/events/papers/templates/emails/base.html
```diff
@@ -0,0 +1,40 @@
...
+            <h3 style="margin: 0; margin-top: 10px; padding-top: 10px; border-top: 1px solid #eee;">
+                {% block header_subtitle %}{% endblock %}
+                <span style="float: right; font-style: italic;">
+                    {% block header_subtitle_ref %}{% endblock %}
+                </span>
+            </h3>
+        </div>
+        <div class="content" style="padding: 10px 20px; color: #555;">
+            {% block content %}
+                <div style="border-bottom: 1px solid #EBEBEB;">
```
mvidalgarcia: Same here.

indico/modules/events/papers/templates/emails/base.html
```diff
@@ -0,0 +1,40 @@
...
+                    {% block content_header %}{% endblock %}
+                </div>
+                {% block content_body %}{% endblock %}
+                <div>
+                    {% block content_footer %}{% endblock %}
+                </div>
+            {% endblock %}
+        </div>
+        <div style="text-align: center; padding: 10px; background: #EBEBEB; font-size: small; color: #999;">
```
mvidalgarcia: And here.

indico/modules/events/papers/templates/emails/revision_submission_to_judge.html
```diff
@@ -0,0 +1,12 @@
+{% extends 'events/papers/emails/revision_submission_base.html' %}
+
+{% block content %}
+    <p>Dear {{ receiver.first_name }},</p>
+    <p>An author has submitted a new revision for the paper <strong>"{{ revision.paper.contribution.title }}"</strong>.</p>
```
ThiefMaster: This is the same as in the reviewer email. Can we use a base class to avoid  repeating it? All that changes is the next paragraph.

indico/modules/events/papers/notifications.py
```diff
@@ -0,0 +1,99 @@
...
+    roles_to_notify = paper_reviewing_settings.get(event, 'notify_on_paper_submission')
+    if PaperReviewingRole.judge in roles_to_notify:
+        for judge in revision.paper.contribution.paper_judges:
+            template = get_template_module('events/papers/emails/revision_submission_to_judge.html', event=event,
+                                           revision=revision, receiver=judge)
+            email = make_email(to_list=judge.email, template=template, html=True)
+            send_email(email, event=event, module='Papers', user=session.user)
+    reviewers = set()
+    if PaperReviewingRole.layout_reviewer in roles_to_notify:
+        reviewers = revision.paper.cfp.layout_reviewers
```
ThiefMaster: I'd use `|=` for consistency. That way you also guarantee  that the next `|=` won't change anything (imagine if the `layout_reviewers` property returned something that's not generated on the fly - then the in-place change 2 lines below could end up modifying data in the DB)

indico/modules/events/papers/operations.py
```diff
@@ -302,6 +321,15 @@ def create_comment(paper, text, visibility, user):
     comment = PaperReviewComment(user=user, text=text, visibility=visibility)
     paper.last_revision.comments.append(comment)
     db.session.flush()
+    receivers = {x for x in paper.cfp.judges}
```
ThiefMaster: Why email judges not assigned to the paper? The might not even be able to see that paper.

indico/modules/events/papers/operations.py
```diff
@@ -302,6 +321,15 @@ def create_comment(paper, text, visibility, user):
...
+    if visibility == PaperCommentVisibility.contributors or visibility == PaperCommentVisibility.reviewers:
+        receivers |= paper.cfp.layout_reviewers if paper.cfp.layout_reviewing_enabled else {}
```
ThiefMaster: same here

indico/modules/events/papers/operations.py
```diff
@@ -302,6 +321,15 @@ def create_comment(paper, text, visibility, user):
...
+        receivers |= paper.cfp.content_reviewers if paper.cfp.content_reviewing_enabled else {}
+    if visibility == PaperCommentVisibility.contributors:
+        receivers |= {x.person.user for x in paper.contribution.person_links}
```
ThiefMaster: a person may not have an indico user associated so you should check for this - but I think it'd actually be better to use the person.email (which may also be empty, so take care of that!) here.

indico/modules/events/papers/templates/emails/added_to_reviewing_team.html
```diff
@@ -0,0 +1,25 @@
+{% extends 'events/papers/emails/base.html' %}
+
+{% block subject_message %}Added in paper reviewing team{% endblock %}
+
```
ThiefMaster: Doesn't it look less strange without empty lines between one-line blocks? Probably just a matter of personal taste though..

indico/modules/events/papers/templates/emails/added_to_reviewing_team.html
```diff
@@ -0,0 +1,25 @@
...
+{% block header_subtitle %}Added in paper reviewing team{% endblock %}
+
+{% block content %}
+    <p>Dear {{ receiver.first_name }},</p>
+    <p>You have been chosen as a <strong>{{ role.title }}</strong> for the conference <strong>{{ event.title }}</strong> in order to help with the paper reviewing process.</p>
+    {% if role.name == 'judge' %}
+        <p>Once there are some papers for you to judge, you will be able to do so in the judging area page:</p>
+        <p>
+            <a href="{{ url_for('papers.papers_list', event, _external=true) }}">
+                {{- url_for('papers.papers_list', event, _external=true) -}}
```
ThiefMaster: You need to add an explicit `management=false` here and in the other emails to make sure these emails always point to the page in the display area and not the equivalent in the management area.

ThiefMaster: FYI this applies to all the endpoints defined in the `for prefix, is_management in ...` loop, but I believe `papers_list` is the only one you are using.

indico/modules/events/papers/templates/emails/base.html
```diff
@@ -0,0 +1,40 @@
...
+
+{% block footer %}{% endblock %}
+
+{% block subject -%}
+    {% block subject_message %}Call for papers{% endblock %}
+    {%- block subject_name %}{% endblock %}
+{%- endblock %}
+
+{% block body %}
+    <div class="mail" style="width: 600px; font-family: Arial; border: 1px solid #dfdfdf; background-color: #fcfcfc; margin: auto auto 40px auto;">
```
ThiefMaster: What's the point of setting classes when we can't use non-inline CSS rules?
Also, do we really want to specify only `Arial` and not at least some fallback font available on non-Windows systems?

indico/modules/events/papers/templates/emails/judgment_reset_to_author.html
```diff
@@ -0,0 +1,18 @@
...
+{% block subject_message %}Paper assessment withdrawn{% endblock %}
+
+{% block header_subtitle %}Paper assessment withdrawn{% endblock %}
+
+{% block header_subtitle_ref %}ref #{{ contribution.friendly_id }}{% endblock %}
+
+{% block content %}
+    <p>Dear {{ receiver.first_name }},</p>
+    <p>Your paper <strong>{{ contribution.title }}</strong> has been put back to review, after the assessment has been withdrawn.</p>
+    <p>To see your paper follow the link to your paper page:</p>
```
ThiefMaster: "your paper's timeline"?

indico/modules/events/papers/templates/emails/judgment_to_author.html
```diff
@@ -0,0 +1,47 @@
...
+{% block subject_message %}Paper assessment{% endblock %}
+
+{% block header_subtitle %}Paper assessment{% endblock %}
+
+{% block header_subtitle_ref %}ref #{{ contribution.friendly_id }}{% endblock %}
+
+{% block content_header %}
+    <p>Dear {{ receiver.first_name }},</p>
+    {% if paper.state.name == 'accepted' %}
+        <p>Congratulations! A judge has <strong>accepted</strong> your paper <strong>{{ contribution.title }}</strong>.</p>
```
ThiefMaster: Isn't "Congratulations" a bit too enthusiastic? Also, I would get rid of "A judge .." in favor of "Your paper has been ...". It doesn't matter to the user that there are judges etc, all that's important is that it has been accepted/rejected/etc

indico/modules/events/papers/templates/emails/judgment_to_author.html
```diff
@@ -0,0 +1,47 @@
...
+        <p>A judge has <strong>rejected</strong> your paper <strong>{{ contribution.title }}</strong>.</p>
+        <p>No further actions are required.</p>
+    {% else %}
+        <p>A judge has requested some <strong>changes</strong> to your paper <strong>{{ contribution.title }}</strong>.</p>
+    {% endif %}
+
+{% endblock %}
+
+{% block content_body %}
+    {% if paper.last_revision._judgment_comment %}
```
ThiefMaster: Why not add a `judgment_comment` property to the `Paper` class?

indico/modules/events/papers/templates/emails/paper_assignment.html
```diff
@@ -0,0 +1,58 @@
...
+        New paper review assignment
+    {%- else -%}
+        Removed from paper review assignment
+    {%- endif %}
+{% endblock %}
+
+{% block content_header %}
+    <p>Dear {{ receiver.first_name }},</p>
+    {% if contribs | length == 1 %}
+        {% if assign %}
```
ThiefMaster: I'd `{% set contrib = contribs|first %}` and use that instead of `contribs[0]`

indico/modules/events/papers/templates/emails/base.html
OmeGak: :fire:

indico/modules/events/papers/templates/emails/base.html
OmeGak: :dark_sunglasses: 
:smiley:

indico/modules/events/papers/templates/emails/base.html
OmeGak: :sunglasses:

indico/modules/events/papers/templates/emails/judgment_reset_to_author.html
nop33: nah, I think paper page is fine..

ThiefMaster: sounds a bit weird tbh. "To see your paper, use the following link"?

indico/modules/events/papers/notifications.py
```diff
@@ -0,0 +1,100 @@
...
+        template = get_template_module('events/papers/emails/review_submission_to_judge.html', event=event,
+                                       review=review, contribution=review.revision.paper.contribution, receiver=judge)
+        email = make_email(to_list=judge.email, template=template, html=True)
+        send_email(email, event=event, module='Papers', user=session.user)
+
+
+def notify_paper_judgment(paper, reset=False):
+    event = paper.event_new
+    receiver = paper.last_revision.submitter
+    if paper_reviewing_settings.get(event, 'notify_author_on_judgment') and not reset:
```
ThiefMaster: This logic can be simplified:

```python
if reset:
    ...
elif paper_reviewing_settings.get(event, 'notify_author_on_judgment'):
    ...
```

indico/modules/events/papers/notifications.py
```diff
@@ -0,0 +1,100 @@
...
+def notify_paper_assignment(user, role, contributions, event, assign):
+    template = get_template_module('events/papers/emails/paper_assignment.html', event=event, contribs=contributions,
+                                   receiver=user, assign=assign, role=role)
+    email = make_email(to_list=user.email, template=template, html=True)
+    send_email(email, event=event, module='Papers', user=session.user)
+
+
+def notify_comment(person, paper, comment, submitter):
+    event = paper.event_new
+    receiver_name = person.first_name or 'user'
```
ThiefMaster: You already do that in the template so receiver should just be `person`

indico/modules/events/papers/operations.py
```diff
@@ -139,16 +158,17 @@ def create_paper_revision(paper, submitter, files):
                        _contribution=paper.contribution)
         pf.save(f.file)
     db.session.flush()
+    db.session.expire(revision._contribution, ['_paper_last_revision'])
```
ThiefMaster: Is this still needed? IIRC the problem you had there was caused by the memoization of the `paper` property.

nop33: It's needed. Otherwise:
```
File "/home/iliastrichopoulos/dev/indico/indico/modules/events/papers/notifications.py", line 27, in notify_paper_revision_submission
    event = revision.paper.event_new
AttributeError: 'NoneType' object has no attribute 'event_new'
```

indico/modules/events/papers/controllers/base.py
```diff
@@ -95,4 +95,7 @@ def _check_paper_protection(self):
         Override this in case you want to check for more specific
         privileges than the generic "can access".
         """
-        return self.contribution.can_access(session.user)
+        return (self.contribution.is_user_associated(session.user, check_abstract=True) or
```
ThiefMaster: wrong place.  check what inherits from this class... this code should go into the override in RHPaperTimeline

indico/modules/events/papers/controllers/display.py
```diff
@@ -59,6 +59,12 @@ class RHPaperTimeline(RHPaperBase):
     def _process(self):
         return render_paper_page(self.paper, view_class=WPDisplayCallForPapers)
 
+    def _check_paper_protection(self):
+        return (self.contribution.is_user_associated(session.user, check_abstract=True) or
+                self.paper.cfp.is_manager(session.user) or
```
ThiefMaster: I'd use `self.event_new.cfp` simply to make it clear that this is not related to that specific paper

PR #2846, opened 2017-02-22
Filter by unassigned paper role in assignment/judging area

mvidalgarcia:


indico/modules/events/papers/lists.py
```diff
@@ -97,6 +101,19 @@ def _filter_list_entries(self, query, filters):
...
+        if 'unassigned' in filters['items']:
+            role_map = {
+                PaperReviewingRole.judge.value: Contribution.paper_judges,
+                PaperReviewingRole.content_reviewer.value: Contribution.paper_content_reviewers,
+                PaperReviewingRole.layout_reviewer.value: Contribution.paper_layout_reviewers,
+            }
+            filtered_roles = filters['items']['unassigned']
+            unassigned_criteria = []
+            for filter_role in filtered_roles:
+                unassigned_criteria.append(role_map[int(filter_role)] == None)
```
ThiefMaster: `.is_(None)`

indico/htdocs/js/indico/modules/list_generator.js
```diff
@@ -15,6 +15,8 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
+/* global setupSearchBox: false, handleRowSelection: false, setupTableSorter: false */
```
ThiefMaster: everywhere else we don't put a `:` before the `false` in that line

indico/modules/events/papers/lists.py
ThiefMaster: Actually, this is a many-to-many relationship. I'd rather use this check:

`~role_map[int(filter_role)].any()`

indico/modules/events/papers/lists.py
```diff
@@ -176,6 +193,9 @@ def __init__(self, event, user):
             'items': ('state',),
             'filters': {'items': {}}
         }
+        judging_unassigned_choices = {role.value: role.title for role in PaperReviewingRole
+                                      if role is not PaperReviewingRole.judge}
+        self.static_items['unassigned']['filter_choices'] = OrderedDict(judging_unassigned_choices.items())
```
ThiefMaster: Adding something to an OrderedDict only makes sense when it's sorted before or comes from a sorted collection - but neither applies to a dict ;)

So you'll have to use `sorted()` there or you can simply use `judging_unassigned_choices` directly (but having a proper order would be better)

indico/modules/events/papers/templates/paper_list_filter.html
```diff
@@ -6,12 +6,14 @@
...
-            </div>
+            {% if item_id != 'unassigned' %}
+                <div class="actions">
+                    <span class="visibility icon-eye"></span>
+                </div>
+            {% endif %}
         </div>
         {% if filter_choices %}
-            <div class="i-label filter">
+            <div class="i-label filter" {% if item_id == 'unassigned' %} data-only-filter {% endif %}>
```
ThiefMaster: `{{ 'data-only-filter' if item_id == 'unassigned' }}`

mvidalgarcia: What's the difference? Just being curious.

ThiefMaster: none, it's just shorter

indico/modules/events/papers/lists.py
mvidalgarcia: I wanted to do it this way at the beginning but I was forgetting the `.any()`, thanks.

indico/modules/events/papers/models/call_for_papers.py
```diff
@@ -72,9 +72,19 @@ def close(self):
...
         elif reviewing_type == 'layout':
-            paper_reviewing_settings.set(self.event, 'layout_reviewing_enabled', enable)
+            self.layout_reviewing_enabled = enable
+        else:
+            raise ValueError('Invalid reviewing type: {}'.format(reviewing_type))
+
+    def get_reviewing_state(self, reviewing_type):
+        if reviewing_type == PaperReviewType.content:
+            return self.content_reviewing_enabled
+        elif reviewing_type == PaperReviewType.layour:
```
mvidalgarcia: layou**t**

ThiefMaster: oh shit, good thing it was late so i can safely force push ;)

PR #2851, opened 2017-02-23
Fix checks in render_paper_section macro

DavidAndreev:


indico/modules/events/papers/templates/_contributions.html
```diff
@@ -1,10 +1,11 @@
 {% macro render_paper_section(contribution, show_paper=True) %}
+    {% set can_access = (contribution.abstract and contribution.abstract.submitter == session.user) or contribution.is_user_associated(session.user) %}
```
nop33: How's the abstract related to the paper? I don't get it...

ThiefMaster: It's what we discussed during the meeting: If you submitted the abstract from which the contribution was created, you should be able to upload the paper for it even if you are not an author/speaker.

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -1,10 +1,11 @@
 {% macro render_paper_section(contribution, show_paper=True) %}
```
ThiefMaster: `true`

indico/modules/events/papers/templates/_contributions.html
nop33: I think you could also use `paper.can_submit`. Check it out in the `Paper` model.

ThiefMaster:
I think we should never show anything paper-related if the papers feature is disabled in the event.

indico/modules/events/papers/templates/_contributions.html
ThiefMaster: I thought so too, but we don't have a `Paper` instance for a contribution with no paper revisions...

nop33: I see. In that case I would rewrite this line with:
```jinja
{% set can_access = paper.can_submit(session.user) or (contribution.abstract and contribution.abstract.submitter == session.user)
```

nop33: Dammit github for not adding the comment before I post mine!

ThiefMaster: But `paper` is None :/

However, `Paper.can_submit()` could use the new `is_user_associated` method.

BTW, we could add a `check_abstract` kwarg to that method and then move the abstract submitter check in there.

nop33: In that case it would make sense to update the `paper.can_submit()` method to use the `is_user_associated`.

DavidAndreev:
@ThiefMaster so maybe like this in contribution_display.html : `{{ render_paper_section(contribution, show_paper=event.has_feature('papers')) }}` , I dont know for what was that show_paper intended at first but there s nothing being passed now as far as i know

indico/modules/events/papers/models/papers.py
```diff
@@ -82,7 +82,8 @@ def can_comment(self, user):
     def can_submit(self, user):
         if not user:
             return False
-        return any(x.person.user == user for x in self.contribution.person_links if x.person.user)
+        return (any(x.person.user == user for x in self.contribution.person_links if x.person.user) or
```
nop33: now you check the same thing twice :smile: 
```
any(pl.person.user == user for pl in self.person_links)

any(x.person.user == user for x in self.contribution.person_links if x.person.user)
```

indico/modules/events/contributions/models/contributions.py
```diff
@@ -454,6 +454,13 @@ def get_non_inheriting_objects(self):
         """Get a set of child objects that do not inherit protection."""
         return get_non_inheriting_objects(self)
 
+    def is_user_associated(self, user, **kwargs):
```
ThiefMaster: Why kwargs?! `check_abstract=False`

indico/modules/events/contributions/models/contributions.py
```diff
@@ -454,6 +454,13 @@ def get_non_inheriting_objects(self):
...
+        if user is None:
+            return False
+        if kwargs.get('check_abstract'):
```
ThiefMaster:
```
if check_abstract and self.abstract and self.abstract.submitter == user:
    return True
return any(...)
```

More readable than trying to squeeze this into a single statement

indico/modules/events/papers/models/papers.py
DavidAndreev: oh ..true :smile:  .. but there is also this check at the end ` if x.person.user` , I dont have that in the is_user_associated, I guess i should add it cause the user can be None right ?

indico/modules/events/contributions/models/contributions.py
```diff
@@ -454,6 +454,12 @@ def get_non_inheriting_objects(self):
         """Get a set of child objects that do not inherit protection."""
         return get_non_inheriting_objects(self)
 
+    def is_user_associated(self, user, check_abstract=False):
+        if user is None:
+            return False
+        if check_abstract and self.abstract and self.abstract.submitter == user:
+            return True
+        return any(pl.person.user == user for pl in self.person_links if pl.person.user)
```
ThiefMaster: FWIW, the `if pl.person.user` isn't necessary since `user` is guaranteed to be not None here. It doesn't harm though so we can keep it

indico/modules/events/contributions/templates/display/contribution_display.html
```diff
@@ -239,7 +239,7 @@
         </section>
     {%- endif %}
 
-    {{ render_paper_section(contribution, show_paper=show_paper) }}
+    {{ render_paper_section(contribution, show_paper=contribution.event_new.has_feature('papers')) }}
```
ThiefMaster: I would wrap this macro call in an if checking for the feature instead of having this argument.

indico/modules/events/papers/models/papers.py
```diff
@@ -82,7 +82,7 @@ def can_comment(self, user):
     def can_submit(self, user):
         if not user:
```
ThiefMaster: that's already checked in `is_user_associated`

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -1,10 +1,11 @@
-{% macro render_paper_section(contribution, show_paper=True) %}
+{% macro render_paper_section(contribution, show_paper=true) %}
+    {% set can_access = contribution.is_user_associated(session.user, check_abstract=true) %}
     {% set paper = contribution.paper %}
     {% set show_paper = paper.accepted_revision and show_paper %}
     {% set show_paper_status = paper.last_revision and contribution.can_manage(session.user, 'submit') %}
     {% set show_paper_submission = contribution.event_new.cfp.is_open and contribution.can_manage(session.user, 'submit') %}
 
-    {% if show_paper or show_paper_status or show_paper_submission %}
+    {% if (show_paper or show_paper_status or show_paper_submission) and can_access %}
```
ThiefMaster: This check is wrong, `can_access` should be called `can_submit` and only apply to the paper status/submission - showing the accepted paper revision doesn't need any special access.

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -1,10 +1,11 @@
-{% macro render_paper_section(contribution, show_paper=True) %}
+{% macro render_paper_section(contribution, show_paper=true) %}
+    {% set can_access = contribution.is_user_associated(session.user, check_abstract=true) %}
     {% set paper = contribution.paper %}
     {% set show_paper = paper.accepted_revision and show_paper %}
     {% set show_paper_status = paper.last_revision and contribution.can_manage(session.user, 'submit') %}
```
ThiefMaster: `= can_submit and ..` - and get rid of the `can_manage` check

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -1,10 +1,11 @@
-{% macro render_paper_section(contribution, show_paper=True) %}
+{% macro render_paper_section(contribution, show_paper=true) %}
+    {% set can_access = contribution.is_user_associated(session.user, check_abstract=true) %}
```
ThiefMaster: `can_submit`

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -1,10 +1,11 @@
...
     {% set paper = contribution.paper %}
     {% set show_paper = paper.accepted_revision and show_paper %}
     {% set show_paper_status = paper.last_revision and contribution.can_manage(session.user, 'submit') %}
     {% set show_paper_submission = contribution.event_new.cfp.is_open and contribution.can_manage(session.user, 'submit') %}
```
ThiefMaster: can_submit instead of the can_manage check

PR #2858, opened 2017-03-01
Allow e-mails to reviewers/judges from CFP page

mvidalgarcia:
image


indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -539,7 +539,9 @@ th.i-table {
     .roles-column {
         width: 5em;
 
-        .submitter-icon {
+        .submitter-icon,
+        .judge-icon,
+        .content-reviewer-icon {
```
ThiefMaster: Can't we just use `.icon` for all of them?

indico/modules/events/abstracts/templates/management/abstract_person_list.html
```diff
@@ -0,0 +1,22 @@
...
+        {% set submitter_title = _("This person submitted an abstract") %}
+    {%- else -%}
+        {% set submitter_title = _("This person hasn't submitted any abstracts") %}
+    {% endif -%}
+    <i class="icon-user-check submitter-icon {{ 'inactive' if not is_submitter }}"
+       title="{{ submitter_title }}"></i>
+{% endblock %}
+
+{% block no_persons_message %}
+    <td colspan="6">{% trans %}No persons in the selected abstracts yet{% endtrans %}</td>
```
ThiefMaster: I think this case is technically impossible since there's always a submitter, so you could just leave it empty (maybe with a comment mentioning that this is never empty).

indico/modules/events/contributions/controllers/management.py
```diff
@@ -289,6 +289,10 @@ def _membership_filter(self):
         contribution_ids = {contrib.id for contrib in self.contribs}
         return Contribution.id.in_(contribution_ids)
 
+    def _process(self):
+        self.template = 'events/contributions/management/contribution_person_list.html'
```
ThiefMaster: why set it on the instance and not just as a class attribute? Then you don't even need to override `_process`

indico/modules/events/contributions/templates/management/contribution_person_list.html
```diff
@@ -0,0 +1,5 @@
+{% extends 'events/management/_event_person.html' %}
+
+{% block no_persons_message %}
+    <td colspan="6">{% trans %}No persons in the selected contributions yet{% endtrans %}</td>
```
ThiefMaster: "There are no persons associated with the selected contributions." sounds better IMO

indico/modules/events/management/templates/_base_person.html
```diff
@@ -0,0 +1,50 @@
...
+                    {% block table_head %}{% endblock %}
+                </tr>
+            </thead>
+            <tbody class="scroll">
+                {% block table_body %}{% endblock %}
+            </tbody>
+        </table>
+    </div>
+    <div class="form-group-mail">
+        <a href="#" class="i-button icon-mail right highlight disabled js-requires-selected-row send-mails-btn"
```
ThiefMaster: `send-mails-btn` seems unused

indico/modules/events/management/templates/_base_person.html
```diff
@@ -0,0 +1,50 @@
...
+           data-href="{{ url_for('persons.email_event_persons', event) }}"
```
ThiefMaster: I think this need to go into a block (can leave the URL as the default) - as soon as you want to allow judge-level emails you need to use a custom RH with extra access checks so they can't email random users/persons.

indico/modules/events/management/templates/_base_person.html
```diff
@@ -0,0 +1,50 @@
...
+           data-params-selector="#person-list-table input[name=person_id]:checkbox:checked:visible, #person-list-table input[name=user_id]:checkbox:checked:visible"
+           data-ajax-dialog>
+            {%- trans %}Send emails{% endtrans -%}
+        </a>
+    </div>
+</div>
+<script>
+    setupEventPersonsList();
+</script>
+
```
ThiefMaster: that line shouldn't be there ;)

indico/modules/events/papers/controllers/management.py
```diff
@@ -123,6 +125,23 @@ def _process(self):
                                               form=form)
 
 
+class RHManageContact(RHManagePapersBase):
+    """Allow to send emails to members of reviewing team"""
```
ThiefMaster: empty line after a class-level docstring

indico/modules/events/papers/controllers/management.py
```diff
@@ -123,6 +125,23 @@ def _process(self):
...
+    def _process(self):
+        cfp = self.event_new.cfp
+        paper_persons_dict = defaultdict(lambda: {'judge': False, 'content_reviewer': False, 'layout_reviewer': False})
+        for judge in cfp.judges:
```
ThiefMaster: I'd use `user` for all these loop variables

indico/modules/events/papers/templates/management/paper_person_list.html
```diff
@@ -0,0 +1,65 @@
...
+                        {% set layout_reviewer_title = _('This person is not a layout reviewer') %}
+                    {% endif -%}
+                    <i class="icon-layout layout-reviewer-icon {{ 'inactive' if not is_layout_reviewer }}"
+                       title="{{ layout_reviewer_title }}"></i>
+                {% endif %}
+            </td>
+        </tr>
+    {%- else -%}
+        <tr class="i-table">
+            <td colspan="6">{% trans %}No reviewing team assigned to the selected papers yet{% endtrans %}</td>
```
ThiefMaster: this is not about selected papers

indico/modules/events/sessions/templates/management/session_person_list.html
```diff
@@ -0,0 +1,5 @@
+{% extends 'events/management/_event_person.html' %}
+
+{% block no_persons_message %}
+    <td colspan="6">{% trans %}No persons in the selected sessions yet{% endtrans %}</td>
```
ThiefMaster: same suggestion as for contributions

indico/modules/events/papers/controllers/management.py
mvidalgarcia: Do you know how to reduce the three loops to one? I'm quite sure that this could be done in less lines of code but don't know how.

ThiefMaster: iterate over `self.event_new.acl_entries` and then check the roles like the methods in the CFP class do

ThiefMaster: you also don't need a defaultdict now since you can assign the whole dict for a user at once

indico/htdocs/sass/modules/_event_management.scss
mvidalgarcia: Yes, better :+1:

indico/modules/events/papers/controllers/management.py
```diff
@@ -123,6 +125,23 @@ def _process(self):
                                               form=form)
 
 
+class RHManageContact(RHManagePapersBase):
+    """Allow to send emails to members of reviewing team"""
+
+    def _process(self):
+        paper_persons_dict = {}
+        for p in self.event.acl_entries:
+            paper_persons_dict[p]['judge'] = p.has_management_role('paper_judge', explicit=True)
```
ThiefMaster: Did you test this after changing it? It'll fail since it's not a defaultdict anymore. Also, you are adding people to the `paper_persons_dict` who are in the event ACL but not related to the papers module.

Check whether the user has the roles first and only then set `paper_persons_dict[p] = {...}` if he has at least one role.

indico/modules/events/management/templates/_base_person.html
```diff
@@ -0,0 +1,51 @@
+<div class="i-form no-max-width event-person-list js-event-person-list">
+    <div class="flashed-messages"></div>
+    <div>
+        <div class="toolbar space-after">
+            <a class="icon-checkbox-checked i-button arrow js-dropdown" href="#" data-toggle="dropdown"></a>
+            <ul class="dropdown">
+                <li>
+                    <a href="#" data-select-all="#person-list-table input.select-row:visible">
+                        {% trans %}All{% endtrans %}
```
nop33: It is more important to remove the whitespace that is enclosed inside an element (like in this case) than to remove it from cases like above
``` python
 +    {%- if is_submitter %}
 +        {% set submitter_title = _("This person submitted an abstract") %}
 +    {%- else -%}
 +        {% set submitter_title = _("This person hasn't submitted any abstracts") %}
 +    {% endif -%}
```

nop33: Sorry, by enclosed I mean that **text** is surrounded with extra whitespace. It doesn't matter if you have:
``` html
<div>
    {% if whatever %}
        {% set ... %}
    {% endif %}
    <span>...</span>
</div>
```
but it does matter if you have
``` html
<div>
    {% trans %}
        Dis is de text
    {% endtrans %}
</div>
```
Check the HTML generated in your browser and you'll see that in the first case not much changes (the browser removes the extra white space between HTML tags), whereas in the second it will keep the whitespace as part of the text.

indico/modules/events/management/templates/_event_person.html
```diff
@@ -0,0 +1,58 @@
...
+            </td>
+            <td class="i-table name-column">{{ event_person.full_name }}</td>
+            <td class="i-table email-column">{{ event_person.email }}</td>
+            <td class="i-table affiliation-column">{{ event_person.affiliation }}</td>
+            <td class="i-table author-type-column">
+                {% if person_roles.primary_author or person_roles.secondary_author %}
+                    {% set author_types = [_('Primary') if person_roles.primary_author,
+                                           _('Co-author') if person_roles.secondary_author] %}
+                    {{ author_types | reject('undefined') | join(', ') }}
+                {% else %}
```
nop33: Here for example it's needed :wink:

indico/modules/events/papers/controllers/management.py
mvidalgarcia: Yes, indeed I didn't test it. Don't know why I pushed it :worried:   

But if I do e.g. `paper_persons_dict[p] = {'judge': True}` and then `paper_persons_dict[p] = {'content_reviewer': True}` I will override the former. So perhaps it's better to keep using the `defaultdict` or am I missing something?

indico/modules/events/management/templates/_base_person.html
mvidalgarcia: Yes you're right. Since I was just rearranging templates I assumed it was right, but I should have double checked.

PR #2863, opened 2017-03-03
Change Booked for user section in Room booking

DavidAndreev:


indico/MaKaC/webinterface/tpls/RoomBookingNewBookingCollaborationWidget.tpl
```diff
@@ -18,7 +24,11 @@
     </div>
     <ul id="videoconferenceOptions" class="weak-hidden">
         % for subfield in [form.needs_vc_assistance] + eq_list:
-            <li class="js-vc-row">
+            % if subfield.short_name == 'needs_vc_assistance' and start_time < work_start or start_time > work_end:
+                <li class="js-vc-row unavailable" title="${ _('Assistance is not available because start time of booking is not within working hours.') }">
+            % else:
+                <li class="js-vc-row">
+            % endif
                 ${ subfield() }
```
ThiefMaster: `disabled=true` should work to disable the field

DavidAndreev: It doesn't work I get : NameError: Undefined , neither does : `disabled='true'` , although `readonly='true'` works, but thats not really what we want ..

indico/legacy/webinterface/tpls/RoomBookingNewBookingCollaborationWidget.tpl
```diff
@@ -18,23 +23,29 @@
...
-            <li class="js-vc-row">
-                ${ subfield() }
+            % if subfield.short_name == 'needs_vc_assistance' and start_dt.time() < work_start or start_dt.time() > work_end:
+                <li class="js-vc-row unavailable" title="${ _('Assistance is not available because start time of booking is not within working hours.') }">
+                    ${ subfield(disabled=True) }
+            % else:
+                <li class="js-vc-row">
+                    ${ subfield() }
+            % endif
                 ${ subfield.label(style='font-weight: normal;') }
```
ThiefMaster: For the sake of readability I'd duplicate this and the next line in both the if and else blocks.

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
```diff
@@ -1,44 +1,52 @@
...
+    ids = {
+        'booked_for_user': 'booked-for-user-wrapper',
+        'booking_reason': 'booking-reason'
     }
 %>
 
 <div id="bookedForInfo">
+    <div id="other-user">
+        <input type="radio" name="other_user" value="False">${ _("I'll be using the room") }<br>
+        <input type="radio" name="other_user" value="True">${ _("I'm making the booking for someone else") }<br>
```
ThiefMaster: I'm booking the room for someone else

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
```diff
@@ -1,44 +1,52 @@
...
+    }
+    ids = {
+        'booked_for_user': 'booked-for-user-wrapper',
+        'booking_reason': 'booking-reason'
     }
 %>
 
 <div id="bookedForInfo">
+    <div id="other-user">
+        <input type="radio" name="other_user" value="False">${ _("I'll be using the room") }<br>
```
ThiefMaster: I'm booking the room for myself (?)

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
```diff
@@ -1,44 +1,52 @@
...
+        <input type="radio" name="other_user" value="True">${ _("I'm making the booking for someone else") }<br>
+    </div>
     % for field in fields:
-        <div class="toolbar thin space-before space-after">
+        <div id="${ ids[field] }" class="toolbar thin space-before space-after">
```
ThiefMaster: Can't you construct an ID based on `field.id`?

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
```diff
@@ -1,44 +1,52 @@
...
-                    </span>
-                % endif
                 ${ form[field](**field_args.get(field, {})) }
                 % if helpers.get(field):
                     <i class="info-helper" title="${ helpers[field] }"></i>
                 % endif
             </div>
         </div>
     % endfor
+    <input id="current-user" type="hidden" value="">
```
ThiefMaster: what is this needed for?

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
```diff
@@ -1,44 +1,52 @@
...
-    $('#booked_for_user').on('change', function() {
-        var user = JSON.parse($(this).val())[0];
-        if (user) {
-            $('#contact_email').val(user.email);
-            $('#contact_phone').val(user.phone);
+    $('#other-user').on('change', function() {
+        if ($('input[name=other_user]:checked').length == 0) {
+            $('#current-user').val($('#booked_for_user').val());
+            $('#booked_for_user').val('');
+        }else{
```
ThiefMaster: space around `else`

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
```diff
@@ -1,44 +1,52 @@
...
 </div>
 
 <script>
-    $('#booked_for_user').on('change', function() {
-        var user = JSON.parse($(this).val())[0];
-        if (user) {
-            $('#contact_email').val(user.email);
-            $('#contact_phone').val(user.phone);
+    $('#other-user').on('change', function() {
+        if ($('input[name=other_user]:checked').length == 0) {
```
ThiefMaster: !...

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
```diff
@@ -1,44 +1,52 @@
...
+            $('#current-user').val($('#booked_for_user').val());
+            $('#booked_for_user').val('');
+        }else{
+            if ($('input[name=other_user]:checked').val() == 'True' ){
```
ThiefMaster: `:checked` only matches the checked item so you can just check for `!!$(...).length` instead of actually looking at the value

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
```diff
@@ -1,44 +1,52 @@
...
+                $('#booked-for-user-wrapper').slideDown();
+                $('#booked_for_user').val('');
+                $('#display-booked_for_user').val('');
+            }else {
```
ThiefMaster: space before `else`

indico/modules/rb/forms/reservations.py
```diff
@@ -111,8 +111,6 @@ class NewBookingPeriodForm(NewBookingFormBase):
 
 class NewBookingConfirmForm(NewBookingPeriodForm):
     booked_for_user = PrincipalField(_(u'User'), [DataRequired()], allow_external=True)
-    contact_email = StringField(_(u'Email'), [InputRequired(), IndicoEmail(multi=True)])
-    contact_phone = StringField(_(u'Telephone'))
```
ThiefMaster: We should decide whether to remove contact_email/phone everywhere! For contact_phone it makes sense for sure, but contact_email is actually used to control modification access on the booking...

    def is_booked_for(self, user):
        if user is None:
            return False
        return self.booked_for_user == user or bool(self.contact_emails & set(user.all_emails))

cc @pferreir

indico/htdocs/js/indico/RoomBooking/validation.js
```diff
@@ -75,10 +75,20 @@
         }
 
         // Booked for validator
-        $('#contact_email, #booking_reason').each(function() {
+        var validInput = function($element) {
```
DavidAndreev: So when do you use var and when only function ?

ThiefMaster: We pretty much never use `var foo = function() { ... }'` in new code. What we do use is stuff like this to  make a function global:

global.foo = function foo() { ... };

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
DavidAndreev: Did you mean `form[field].id` ? Or that i should make `fields` a dictionary with names and ids ? I need different id for the div and different for input element in booking-for-user

ThiefMaster: Ah yes if field is a string then `form[field].id`

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
DavidAndreev: Its to remember the logged in user when user switches between I'm booking the room and Booking for someone else

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
```diff
@@ -1,44 +1,55 @@
...
 
-    % if not reservation:
-        $('#booked_for_user').trigger('change');
-    % endif
+        if (!$('input[name=other_user]:checked').length) {
+            $('#current-user').val($('#booked_for_user').val());
+            $('#booked_for_user').val('');
+        }
+        $('#other-user').on('change', function() {
+            if ($('input[name=other_user]:checked').val() == 'True' ){
```
mvidalgarcia: You can move the space after `True` before the curly bracket.

indico/htdocs/sass/modules/_roombooking.scss
```diff
@@ -150,10 +150,37 @@
...
+        display: none;
+        min-height: 0;
+
+        &.hasError {
+            @include border-all($red);
+        }
+    }
+
+    #other-user {
+        display: table-cell;
```
nop33: Not very important, but did you check whether you can do the same with flexbox? :smile: If it takes too long and the layout is already implemented with tables, don't waste too much time, but I'd prefer seeing flexbox attributes in newly added code! :laughing:

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
```diff
@@ -1,44 +1,55 @@
...
 <%
-    fields = ['booked_for_user', 'contact_email', 'contact_phone', 'booking_reason']
+    fields = ['booked_for_user', 'booking_reason']
     field_args = {
         'booking_reason': {'rows': 3, 'placeholder': _(u'Reason...')}
     }
     helpers = {
-        'contact_email': 'You can specify multiple email addresses separated by commas.',
-        'booking_reason': 'Specify why you need this room.'
+        'booking_reason': 'Specify why you need this room.',
```
nop33: Shouldn't these be translated?

ThiefMaster: yes, probably

indico/htdocs/sass/modules/_roombooking.scss
DavidAndreev: Well I need to set the width of `other-user` to the width of its content, not sure how to do it with flexbox

indico/legacy/webinterface/tpls/RoomBookingNewBookingInfoWidget.tpl
pferreir: Shouldn't we fill `value` with the current user's ID then, instead of using JS to do it? We could just disable the field if the other option was chosen.

indico/modules/rb/forms/reservations.py
pferreir: Can't we just use the info from `booked_for_user`?

PR #2868, opened 2017-03-07
Implement event cloning using new stack

pferreir:
I know the UI is still a bit... broken, but the functionality should all be there.


indico/htdocs/js/indico/modules/events/cloning.js
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+
+        function updateCount() {
+            var $cloneButton = $('.clone-action-button');
+
+            $.ajax({
+                url: $formContainer.data('previewUrl'),
+                method: 'POST',
+                data: getFormParams($form),
```
ThiefMaster: Why not use `$form.ajaxSubmit(...)`? That takes are about serializing all the data

indico/htdocs/js/indico/modules/events/cloning.js
```diff
@@ -0,0 +1,129 @@
...
+                    return $('<div>').append(label, items);
+                }));
+            }
+        }
+
+        function updateCount() {
+            var $cloneButton = $('.clone-action-button');
+
+            $.ajax({
+                url: $formContainer.data('previewUrl'),
```
ThiefMaster: I thought we switched to dash notation :P

indico/htdocs/js/indico/modules/events/cloning.js
```diff
@@ -0,0 +1,129 @@
...
+                method: 'POST',
+                data: getFormParams($form),
+                success: function(data) {
+                    if (data.success) {
+                        var $countNumber = $eventCount.find('.count');
+                        $countNumber.text(data.count);
+                        $cloneErrors.hide();
+                        $eventCount.show();
+
+                        if (data.count) {
```
ThiefMaster:
```
$eventList.toggle(!!data.count);
$cloneButton.prop('disabled', !data.count);
```

indico/htdocs/js/indico/modules/events/cloning.js
```diff
@@ -0,0 +1,129 @@
...
+                    } else {
+                        $cloneErrors.show().find('.message-text').html(errorToHTML(data.error.message));
+                        $eventCount.hide();
+                    }
+                }
+            });
+        }
+
+        if (step === 4) {
+            $form.on('change', 'select,input', updateCount);
```
ThiefMaster: I'd put a space after the `,` like we do in CSS

indico/htdocs/js/indico/modules/events/cloning.js
```diff
@@ -0,0 +1,129 @@
...
+            $form.on('keyup', 'input', function(e) {
```
ThiefMaster: Could probably use the 'input' event instead of 'keyup'

indico/htdocs/js/indico/modules/events/cloning.js
```diff
@@ -0,0 +1,129 @@
...
+                e.preventDefault();
+                updateCount();
+            });
+        }
+
+        if ($eventCount.length) {
+            setTimeout(updateCount, 100);
```
ThiefMaster: why do we need the delay?

indico/web/assets/bundles.py
```diff
@@ -393,7 +393,8 @@ def rjs_bundle(name, *files):
                                        'Layout.js', 'Undo.js', 'Base.js', 'DragAndDrop.js', 'Draw.js', 'Actions.js')),
     'sessions': rjs_bundle('modules_sessions', 'js/indico/modules/sessions.js'),
     'users': rjs_bundle('modules_users', 'js/indico/modules/users.js'),
-    'designer': rjs_bundle('modules_designer', 'js/indico/modules/designer.js')
+    'designer': rjs_bundle('modules_designer', 'js/indico/modules/designer.js'),
+    'event_cloning': rjs_bundle('modules_event_Cloning', 'js/indico/modules/events/cloning.js')
```
ThiefMaster: typo (uppercase C)

indico/modules/events/cloning.py
```diff
@@ -71,34 +71,32 @@ class EventCloner(object):
 
     @classmethod
     def get_form_items(cls, old_event):
-        cloners = sorted([cloner_cls(old_event) for cloner_cls in get_event_cloners().itervalues()],
-                         key=lambda x: x.friendly_name)
-        return render_template('events/cloner_options.html', cloners=cloners)
+        return sorted([cloner_cls(old_event) for cloner_cls in get_event_cloners().itervalues()],
+                      key=lambda x: x.friendly_name)
```
ThiefMaster: attrgetter

indico/modules/events/management/blueprint.py
```diff
@@ -47,6 +47,8 @@
 _bp.add_url_rule('/protection/acl', 'acl', RHEventACL)
 _bp.add_url_rule('/protection/acl-message', 'acl_message', RHEventACLMessage)
 _bp.add_url_rule('/move', 'move', RHMoveEvent, methods=('POST',))
+_bp.add_url_rule('/clone', 'clone', RHCloneEvent, methods=('POST', 'GET'))
```
ThiefMaster: `('GET', 'POST')`

&lt;/nitpicking-mode&gt;

indico/modules/events/management/controllers.py
```diff
@@ -17,28 +17,35 @@
 from __future__ import unicode_literals
 
 import uuid
-from collections import defaultdict
+from collections import defaultdict, OrderedDict
 
-from flask import flash, redirect, session, request, signals
-from werkzeug.exceptions import Forbidden, NotFound
+from datetime import datetime
```
ThiefMaster: this one is stdlib

indico/modules/events/management/controllers.py
```diff
@@ -328,6 +400,89 @@ def _process(self):
...
+
+        if form.repeatability.data == 'interval':
+            clone_calculator = IntervalCloneCalculator(self.event_new)
+        elif form.repeatability.data == 'pattern':
+            clone_calculator = PatternCloneCalculator(self.event_new)
+        else:
+            raise BadRequest
+
+        try:
+            dates = clone_calculator.calculate(request.form)
```
ThiefMaster: can this one also raise a `ValueError`? if not the exception handling here is a bit overkill

indico/modules/events/management/controllers.py
```diff
@@ -328,6 +400,89 @@ def _process(self):
...
+            if len(dates) > 100:
+                raise ValueError(_("You can clone maximum of 100 times at once"))
+        except ValueError as e:
+            return jsonify(error={'message': e.message})
+        return jsonify_data(count=len(dates), dates=dates)
```
ThiefMaster: `flash=False` or, if this is not processed by ajaxdialog/ajaxform logic, simply `jsonify()`

indico/modules/events/management/templates/clone_event.html
```diff
@@ -0,0 +1,107 @@
+{% from 'forms/_form.html' import form_header, form_fieldset, form_footer, form_row, form_rows %}
+{% from 'message_box.html' import message_box %}
+
+{% macro invisible_field(field) -%}
+    {% set field_value = field._value() if field._value else field.data %}
```
ThiefMaster: I think you wanted `if field._value()` here

indico/modules/events/management/templates/clone_event.html
```diff
@@ -0,0 +1,107 @@
...
+    {% set data = [field_value] if field_value is string else ([field_value|tojson] if field_value is mapping else field_value) %}
```
ThiefMaster: that line is very readable ;) I'd wrap it and also move the `|forceescape` behind the `|tojson` (since that's the only case where it's needed)

indico/modules/events/management/templates/clone_event.html
```diff
@@ -0,0 +1,107 @@
...
+    {% for value in data %}
+        <input type="hidden" name="{{ field.name }}" value="{{ value | forceescape }}">
+    {% endfor %}
+{%- endmacro %}
+
+{% macro invisible_fields(form, skip) %}
+    {% for field in form %}
```
ThiefMaster: `{% for field in form if field.id not in skip %}`

Also, I'd use `field.short_name` even though it's usually the same as the ID (but wouldn't be if the form had a prefix set)

indico/modules/events/management/templates/clone_event.html
```diff
@@ -0,0 +1,107 @@
...
+        {% if field.id not in skip %}
+            {{ invisible_field(field) }}
+        {% endif %}
+    {% endfor %}
+{% endmacro %}
+
+{% block content %}
+    <div id="event-clone-form-container"
+         data-preview-url="{{ url_for('event_management.clone_preview', event) }}"
```
ThiefMaster: `event_management` can be omitted, I guess this template is only rendered from within the event_management blueprint

indico/util/date_time.py
```diff
@@ -27,14 +27,36 @@
...
+from dateutil.relativedelta import relativedelta as _relativedelta
 
 from indico.core.config import Config
 from indico.util.i18n import get_current_locale, _, ngettext, parse_locale
 from indico.util.string import inject_unicode_debug
 from indico.legacy.common.timezoneUtils import DisplayTZ
 
 
+class relativedelta(_relativedelta):
+    """Improved `relativedelta`"""
```
ThiefMaster: Out of curiosity, how is it improved?

indico/web/templates/forms/_form.html
```diff
@@ -9,13 +9,14 @@
 {%- endmacro %}
 
 {% macro form_header(form, method='post', action='', orientation='horizontal', id=none, classes='', multipart=false,
-                     i_form=true, no_ajax=false) -%}
+                     i_form=true, no_ajax=false, ajax_form=false) -%}
```
ThiefMaster: Not sure if this is good idea. You usually need lots of extra data attributes in an ajax form...

indico/htdocs/js/indico/modules/events/cloning.js
```diff
@@ -0,0 +1,129 @@
...
+        var $formContainer = $('#event-clone-form-container');
+        var $form = $('#event-clone-form');
+        var $eventCount = $('#event-count').hide();
+        var $eventList = $form.find('.event-list');
+        var $cloneErrors = $('#clone-errors').hide();
+        var step = $formContainer.data('step');
+
+        var clonerDependencies = $formContainer.data('clonerDependencies');
+
+
```
nop33: Are the 2 empty lines intended?

indico/modules/events/management/controllers.py
```diff
@@ -54,6 +61,71 @@
...
+        args = {}
+        if form.stop_criterion.data == 'day':
+            args['until'] = utc.localize(datetime.combine(form.until_dt.data, form.start_dt.data.time()))
+        else:
+            args['count'] = form.num_times.data
+        return args
+
+    def calculate(self, formdata):
+        form = self.form_class(self.event, formdata=formdata)
+        if not form.validate():
```
nop33: I would switch these conditions:
```python
if form.validate():
    return ...
else:
    return ...
```
or even
```python
if form.validate():
    return ...
return ...
```

indico/modules/events/management/controllers.py
```diff
@@ -328,6 +400,89 @@ def _process(self):
...
+            if form.validate_on_submit():
+                if form.repeatability.data == 'once':
+                    dates = [form.start_dt.data]
+                else:
+                    if form.repeatability.data == 'interval':
+                        clone_calculator = IntervalCloneCalculator(self.event_new)
+                    elif form.repeatability.data == 'pattern':
+                        clone_calculator = PatternCloneCalculator(self.event_new)
+                    else:
+                        raise BadRequest
```
nop33: The above if-else block could go into a function since it's repeated in `RHClonePreview`

indico/modules/events/management/templates/clone_event.html
```diff
@@ -0,0 +1,107 @@
...
+                        {{ form_row(form.recurrence) }}
+                        {{ form_rows(form, fields=visible_fields|reject('equalto', 'recurrence')) }}
+                    {% elif form.repeatability.data == 'pattern' %}
+                        {% set visible_fields = ('start_dt', 'day_number', 'week_day', 'num_months', 'stop_criterion',
+                                                 'until_dt', 'num_times') %}
+                        {{ form_rows(form, fields=visible_fields, skip_labels=true) }}
+                    {% endif %}
+                    {{ invisible_fields(form, skip=visible_fields) }}
+                    {% if form.repeatability.data != 'once' %}
+                        {% call message_box('info', fixed_width=True, id='event-count') -%}
```
nop33: lower case `true`

indico/modules/events/operations.py
```diff
@@ -118,6 +120,40 @@ def update_event(event, update_timetable=False, **data):
     _log_event_update(event, changes, visible_person_link_changes=visible_person_link_changes)
 
 
+def clone_event(event, start_dt, cloners, category=None):
+    """Clone an event on a given date/time.
+
+    Runs all required cloners.
+
+    :param start_dt: the start datetime of the new event;
```
nop33: I think we start the sentence with capital, like you do in the line below.

indico/web/forms/fields/datetime.py
```diff
@@ -110,6 +110,88 @@ def _value(self):
...
+            # Add whatever unit is necessary to represent the currenet value if we have one
+            choices.append((unit, '({})'.format(self.unit_names[unit])))
+        return choices
+
+    def process_formdata(self, valuelist):
+        if valuelist and len(valuelist) == 2:
+            value = int(valuelist[0])
+            unit = valuelist[1]
+            if unit not in self.magnitudes:
+                raise ValueError('Invalid unit')
```
nop33: Some `ValueError`s in your PR are translated and some others not.

ThiefMaster: I think errors you only get by tampering with the data aren't really worth translating - even though I'm not sure when you would end up with pre_validate failing without messing with the posted data

indico/util/date_time.py
pferreir: With `__abs__`. See the [ongoing discussion here](https://github.com/dateutil/dateutil/issues/350#issuecomment-284470151).

indico/htdocs/js/indico/modules/events/cloning.js
nop33: Not sure how to use `$form.ajaxSubmit` here, what are you trying to improve?

ThiefMaster: In the end you are submitting the form's data with AJAX. `$(someform).ajaxSubmit()` does exactly that, instead of having to use our custom `data: getFormParams($form)`

indico/modules/events/management/templates/clone_event.html
nop33: Why would the cloning event forms get a prefix?

indico/modules/events/management/templates/clone_event.html
nop33: You can use it in the display of a meeting.

indico/modules/events/management/templates/clone_event.html
ThiefMaster: There is no good reason to write code that can easily break if it can be done properly ;) `field.id` is the HTML ID which is just coincidentally identical to the name in most cases

indico/modules/events/management/templates/clone_event.html
ThiefMaster: But isn't it loaded in an AJAX dialog? So it'll still be inside the management blueprint

indico/htdocs/js/indico/modules/events/cloning.js
pferreir: Basically, when the form loads some of the fields are set from JavaScript. If we immediately send the request, we will send wrong information to the server. I know it's hacky, but I can't see a better solution right now. Suggestions are welcome!

nop33: I don't know, it works without it too. @pferreir ?

indico/modules/events/management/controllers.py
nop33: Yes it can. Why is it an overkill? There are 2 value errors that might happen, one is wrong data in the form and the other is out of range data (>100). How'd you do it instead?

ThiefMaster: read my comment again :P

"if not" - so since it can raise ValueError in there too it isn't overkill.

nop33: ok sorry, my bad

indico/web/templates/forms/_form.html
nop33: Well, we do have the caller in there so we could add all additional attributes. But then half of the ajax form related attributes will be put through the form_header parameters and others through the caller. I don't have a strong opinion.

ThiefMaster: Sounds like a very good reason for me to keep this one in the caller too instead of adding it to the normal args

nop33: @pferreir you agree with this? I can change it.

indico/web/forms/fields/datetime.py
pferreir: I fully agree with @ThiefMaster. I had overlooked the fact that the `caller` is passed in the attributed section.

indico/web/templates/forms/_form.html
pferreir: I fully agree with @ThiefMaster. I had overlooked the fact that the caller is passed in the attributed section.

PR #2869, opened 2017-03-07
Private surveys

DavidAndreev:

PR #2870, opened 2017-03-08
Option to send copy of an email to sender as well

### LINKED ISSUE #2225: Add "send copy to me" option to standard e-mail dialogs
Unless there is a good reason to hide it, it seems to be useful for most users.
Internal reference: `CHG34180`

DavidAndreev:
Closes #2225

### DIFF
--- a/indico/modules/events/persons/controllers.py
+++ b/indico/modules/events/persons/controllers.py
@@ -200,7 +200,9 @@ class RHEmailEventPersons(RHConferenceModifBase):
             email_subject = replace_placeholders('event-persons-email', form.subject.data, person=recipient,
                                                  event=self.event_new, register_link=self.no_account)
             tpl = get_template_module('emails/custom.html', subject=email_subject, body=email_body)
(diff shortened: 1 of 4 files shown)


ThiefMaster:
emailing registrants uses a different RH than emailing event persons / users

DavidAndreev:
So far I added this only to email dialog from Roles page, should this be also available when sending registration emails ? Cause there we already have preview email functionality so maybe its not necessary.  If required I can also put it to email dialogs when sending emails to reviewers after [ #2858](https://github.com/indico/indico/pull/2858) gets merged and when sending emails with survey links [#2869](https://github.com/indico/indico/pull/2869) , but again not sure if its needed.

ThiefMaster:
I think it's not so much about previews but about having a copy in your mailbox

DavidAndreev:
Ok so can I use the first registration from array as an example of an email sent ? Like we do when displaying the preview? Cause I need to pass some registration in order to replace the placeholders.

PR #2873, opened 2017-03-10
Allow to fill surveys progressively + Private surveys

mvidalgarcia:
image
image

### todo:

- [x] Make sure corner messages are the best option to inform the user
- [x] Make progressive surveys configurable (enable/disable)?
- [x] Delete `pending_answers` when a survey is submitted


indico/migrations/versions/201703021819_3b1d8a0a5267_add_is_anonymous_and_is_submitted_columns_to_submissionsurvey.py
```diff
@@ -0,0 +1,46 @@
...
+                               '(is_submitted AND is_anonymous) = (user_id IS NULL)', schema='event_surveys')
+
+
+def downgrade():
+    op.drop_constraint('ck_submissions_anonymous_or_user', 'submissions', schema='event_surveys')
+    op.drop_constraint('ck_submissions_dt_set_when_submitted', 'submissions', schema='event_surveys')
+    op.drop_constraint('ck_submissions_submitted_and_anonymous_no_user', 'submissions', schema='event_surveys')
+    op.drop_column('submissions', 'is_anonymous', schema='event_surveys')
+    op.drop_column('submissions', 'is_submitted', schema='event_surveys')
+    op.alter_column('submissions', 'submitted_dt', existing_type=postgresql.TIMESTAMP(), nullable=False,
```
ThiefMaster: `existing_type` is not needed with postgres

indico/migrations/versions/201703021819_3b1d8a0a5267_add_is_anonymous_and_is_submitted_columns_to_submissionsurvey.py
```diff
@@ -0,0 +1,46 @@
...
+                  sa.Column('is_anonymous', sa.Boolean(), nullable=False, server_default='false'),
+                  schema='event_surveys')
+    op.add_column('submissions',
+                  sa.Column('is_submitted', sa.Boolean(), nullable=False, server_default='false'),
+                  schema='event_surveys')
+    op.execute('UPDATE event_surveys.submissions SET is_anonymous = (user_id IS NULL)')
+    op.alter_column('submissions', 'is_anonymous', server_default=None, schema='event_surveys')
+    op.execute('UPDATE event_surveys.submissions SET is_submitted = (submitted_dt IS NOT NULL)')
+    op.alter_column('submissions', 'is_submitted', server_default=None, schema='event_surveys')
+    op.alter_column('submissions', 'submitted_dt', existing_type=postgresql.TIMESTAMP(), nullable=True,
```
ThiefMaster: `existing_type` is not needed with postgres

indico/modules/events/surveys/controllers/display.py
```diff
@@ -84,6 +89,7 @@ def _checkParams(self, params):
                        .options(joinedload('submissions'),
                                 joinedload('sections').joinedload('children'))
                        .one())
+        self.submission = SurveySubmission.find_first(user=session.user, survey=self.survey)
```
ThiefMaster:
```python
if session.user:
    self.submission = SurveySubmission.query.with_parent(self.survey).filter_by(user=session.user).first()
else:
    self.submission = None
```

Otherwise you might end up getting an anonymous submission when you aren't logged in. Also, we kind of stopped using the `find_*` shortcuts in favor of plain sqlalchemy functionality some time ago.

ThiefMaster: Actually, even better:

    session.user.survey_submissions.filter_by(survey=survey, is_submitted=False).first()

You don't want to get a "finalized" submission there after all.

indico/modules/events/surveys/controllers/display.py
```diff
@@ -18,20 +18,24 @@
...
 from indico.modules.auth.util import redirect_to_login
 from indico.modules.events.models.events import EventType
 from indico.modules.events.surveys.models.submissions import SurveyAnswer, SurveySubmission
 from indico.modules.events.surveys.models.surveys import Survey, SurveyState
-from indico.modules.events.surveys.util import make_survey_form, was_survey_submitted, save_submitted_survey_to_session
+from indico.modules.events.surveys.util import (make_survey_form, is_submission_in_progress,
+                                                save_submitted_survey_to_session, was_survey_submitted)
 from indico.modules.events.surveys.views import (WPDisplaySurveyConference, WPDisplaySurveyMeeting,
                                                  WPDisplaySurveyLecture)
+from indico.legacy.webinterface.rh.conferenceDisplay import RHConferenceBaseDisplay
```
nop33: Alphabetic order :wink:

indico/modules/events/surveys/templates/display/survey_list.html
```diff
@@ -37,6 +37,11 @@
                                title="{% trans %}Survey is closed{% endtrans %}">
                                 {% trans %}Fill out{% endtrans %}
                             </a>
+                        {% elif is_submission_in_progress(survey) %}
+                            <a href="{{ url_for('.display_survey_form', survey) }}"
+                               class="i-button text-color highlight outline">
+                                {% trans %}Continue{% endtrans %}
```
nop33: Pick a style! `{%-` or just `{%`

indico/modules/events/surveys/templates/display/survey_questionnaire.html
```diff
@@ -49,7 +57,23 @@
...
+        (function() {
+            'use strict';
+            // allow deselecting a horizontal radio button unless it's required
+            $('#survey-questionnaire .inline-vcentered input:radio:not([required])').resettableRadioButtons();
+            $('.js-save-answers').on('declarative:success', function() {
+                var $form = $(this).closest('form');
+                cornerMessage({
+                    actionLabel: $T.gettext('Submit'),
+                    progressMessage: $T.gettext('Submitting...'),
+                    message: $T.gettext('Your answers have been saved, do not forget to submit when you finish!'),
```
nop33: *do not forget to submit* **the survey** *when you finish!*

indico/modules/events/surveys/controllers/display.py
ThiefMaster: keeping indico.legacy separate is actually not too bad (like we did with makac).

however, considering that (AFAIK) none of the tools that sort imports let you enforce that, i guess going for alphabetical order whenever we add a legacy import is easier

indico/modules/events/surveys/templates/display/survey_questionnaire.html
mvidalgarcia: I've tried exactly that same sentence before but then the corner message split in two lines and was uglier. I can change it though.

nop33:
> Your answers have been saved, do not forget to submit when you finish!

In any case, this string is translatable, meaning that in other words 
the sentence can and will be longer, meaning that the corner message 
will grow bigger. The best is to fix the actual problem and make the 
corner message look better with 2 lines.

Try this:

```html
<div class="corner-message highlight save-corner-message FLEXROW 
F-A-CENTER">
     <SPAN>Do not forget to save your changes! Do not forget to save 
your changes!</SPAN>
     <a class="corner-message-text" href="#" style="FLEX: 0 0 
AUTO;">Save now</a>
</div>
```

I noticed that something is overriding the `display: flex` when you add 
the class `flexrow` to the `.corner-message` element.
The proper way to solve it would be to find out what it is and fix it 
(probably the order the CSS is loaded, I guess the flex partial needs to 
go higher up in the `_base.scss` file. The workaround would be to use 
`display: flex !important` to the element.

indico/modules/events/surveys/templates/display/survey_questionnaire.html
nop33: > Your answers have been saved, do not forget to submit when you finish!

In any case, this string is translatable, meaning that in other words the sentence can and will be longer, meaning that the corner message will grow bigger. The best is to fix the actual problem and make the corner message look better with 2 lines.

Try this:

```html
<div class="corner-message highlight save-corner-message FLEXROW  F-A-CENTER">
     <SPAN>Do not forget to save your changes! Do not forget to save your changes!</SPAN>
     <a class="corner-message-text" href="#" style="FLEX: 0 0 AUTO;">Save now</a>
</div>
```

I noticed that something is overriding the `display: flex` when you add the class `flexrow` to the `.corner-message` element. The proper way to solve it would be to find out what it is and fix it 
 (probably the order the CSS is loaded, I guess the flex partial needs togo higher up in the `_base.scss` file. The workaround would be to use  `display: flex !important` to the element.

mvidalgarcia:
Updated

indico/htdocs/sass/partials/_corner-messages.scss
```diff
@@ -62,17 +62,13 @@
...
     @include _progress(darken($light-gray, 0.3*$color-variation), $light-gray);
 
-    display: block;
-    float: right;
-    clear: both;
-
     padding: 1.2em;
     margin-bottom: 1em;
-    box-shadow: 0px 3px 2px rgba(150, 150, 150, 0.2);
+    box-shadow: 0 3px 2px rgba(150, 150, 150, 0.2);
```
nop33: We have the `box-shadow` mixin, check how it's used (ex: `/indico/htdocs/sass/partials/_forms.scss`)

indico/modules/events/surveys/templates/display/survey_questionnaire.html
```diff
@@ -49,7 +57,24 @@
...
+            'use strict';
+            // allow deselecting a horizontal radio button unless it's required
+            $('#survey-questionnaire .inline-vcentered input:radio:not([required])').resettableRadioButtons();
+            $('.js-save-answers').on('declarative:success', function() {
+                var $form = $(this).closest('form');
+                cornerMessage({
+                    actionLabel: $T.gettext('Submit'),
+                    progressMessage: $T.gettext('Submitting...'),
+                    message: $T.gettext('Your answers have been saved, do not forget to submit the survey when you ' +
+                    'finish!'),
```
pferreir: Not sure if our i18n processor is smart enough to figure this one out. I'd keep it in one line, even if that means putting it in a variable.

ThiefMaster: Quite sure it's not. Especially not with the `+`

I think exceeding the line length is not too bad for this - but if it fits by using a variable, even better!

mvidalgarcia: Didn't know that! :+1:

indico/modules/events/surveys/controllers/display.py
```diff
@@ -112,13 +121,30 @@ def _process(self):
...
+
+    @no_autoflush
+    def _save_answers(self, form, submission):
         survey = self.survey
-        submission = SurveySubmission(survey=survey)
-        if not survey.anonymous:
-            submission.user = session.user
+        if not submission:
+            submission = SurveySubmission(survey=survey, user=session.user)
+        submission.is_anonymous = survey.anonymous
```
ThiefMaster: I'd move this logic into `_save_answer`?

indico/modules/events/surveys/controllers/display.py
```diff
@@ -112,13 +121,30 @@ def _process(self):
...
         for question in survey.questions:
             answer = SurveyAnswer(question=question, data=getattr(form, 'question_{}'.format(question.id)).data)
             submission.answers.append(answer)
         db.session.flush()
         return submission
+
+
+class RHSaveSurveyAnswers(RHSubmitSurvey):
```
ThiefMaster: I'd move the common parts (everything until the end of checkParams) from `RHSubmitSurvey` to a new class `RHSubmitSurveyBase` and inherit both submission and saving from that class.

indico/modules/events/surveys/controllers/display.py
```diff
@@ -112,13 +121,30 @@ def _process(self):
...
+    def _process(self):
+        pending_answers = {k: v for k, v in request.form.iterlists() if k.startswith('question_')}
+        if not self.submission:
+            self.submission = SurveySubmission(survey=self.survey, user=session.user)
+        self.submission.pending_answers = pending_answers
+        self.submission.is_anonymous = self.survey.anonymous
```
ThiefMaster: I wonder if `is_anonymous` is even needed anymore. An incomplete submission is never shown anyway and we could just determine whether to keep the user or not based on the anonymity setting when doing the final submission.
It's quite unlikely that the setting changes while a survey is running anyway.

cc @pferreir

indico/modules/events/surveys/templates/display/survey_questionnaire.html
mvidalgarcia: Doesn't fit using a variable :hankey:

indico/modules/events/surveys/controllers/display.py
mvidalgarcia: I don't get your point. It's already in `_save_answers`. Do you mean moving it to `_process()` below `_save_answers` is called or create a new function `save_answer()`?

ThiefMaster: Nevermind, I somehow thought this was in some other method.

indico/modules/events/surveys/controllers/display.py
mvidalgarcia: Mmh I have to look deeper into it but yes, I think it could be removed since we can easily access to the anonimity from the submission with `subimission.survey.anonymous`

indico/migrations/versions/201703131140_3b1d8a0a5267_add_is_anonymous_and_is_submitted.py
```diff
@@ -0,0 +1,44 @@
+"""Add is_anonymous and is_submitted columns to SubmissionSurvey
```
ThiefMaster: SurveySubmission

indico/modules/events/surveys/placeholders.py
```diff
@@ -0,0 +1,51 @@
...
+class SurveyTitlePlaceholder(Placeholder):
+    name = 'survey_title'
+    description = _("The title of the survey")
+
+    @classmethod
+    def render(cls, event, survey, **kwargs):
+        return survey.title
+
+
+class SurveyLinkPlaceholder(Placeholder):
```
ThiefMaster: This one should be `required = True`. See [here](https://github.com/indico/indico/blob/master/indico/modules/events/registration/forms.py#L160-L163) for an example on how to validate this.

indico/modules/events/surveys/controllers/display.py
```diff
@@ -50,7 +53,7 @@ def view_class(self):
 class RHSurveyList(RHSurveyBaseDisplay):
     def _process(self):
         surveys = (Survey.query.with_parent(self.event_new)
-                   .filter(Survey.is_visible)
+                   .filter(Survey.is_visible, ~Survey.private)
```
ThiefMaster: Didn't we want to show a private survey if you have an imcomplete submission for it?

indico/modules/events/surveys/controllers/display.py
```diff
@@ -110,15 +122,39 @@ def _process(self):
...
+    def _make_form(self):
+        survey_form_class = make_survey_form(self.survey)
+        if self.submission and request.method == 'GET':
+            return survey_form_class(formdata=MultiDict(self.submission.pending_answers))
+        else:
+            return survey_form_class()
 
-    def _save_answers(self, form):
+    @no_autoflush
+    def _save_answers(self, form, submission):
```
ThiefMaster: You could actually use `self.submission` instead of passing it as an argument

indico/modules/events/surveys/controllers/display.py
```diff
@@ -110,15 +122,39 @@ def _process(self):
...
             submission.answers.append(answer)
         db.session.flush()
         return submission
+
+
+class RHSaveSurveyAnswers(RHSubmitSurveyBase):
+    def _checkProtection(self):
+        RHSubmitSurveyBase._checkProtection(self)
+        if not self.survey.partial_completion:
+            raise Forbidden(response=redirect_to_login(reason=_('You are trying to save a survey '
```
ThiefMaster: Why redirect to login? I'd just `raise Forbidden` - no need to specify messages when it's something only reachable by tampering with the data sent

indico/modules/events/surveys/forms.py
```diff
@@ -115,3 +121,22 @@ class ImportQuestionnaireForm(IndicoForm):
     json_file = FileField(_('File'), accepted_file_types="application/json,.json",
                           description=_("Choose a previously exported survey content to import. "
                                         "Existing sections will be preserved."))
+
+
+class EmailPeopleForm(IndicoForm):
```
ThiefMaster: I'd call this `InvitationForm` (we also called it like this in the registration module when inviting people)

indico/modules/events/surveys/templates/display/survey_list.html
```diff
@@ -30,24 +30,29 @@
...
-                                {% trans %}Fill out{% endtrans %}
+                                {%- trans %}Fill out{% endtrans -%}
                             </a>
                         {% elif not survey.is_active %}
                             <a class="i-button highlight disabled"
                                title="{% trans %}Survey is closed{% endtrans %}">
-                                {% trans %}Fill out{% endtrans %}
+                                {%- trans %}Fill out{% endtrans -%}
+                            </a>
+                        {% elif is_submission_in_progress(survey) %}
```
ThiefMaster: This results in `O(n)` queries. Not a big deal here since you rarely have tons of surveys and it's not on the main page of an event but it'd be nicer to get rid of this util and instead query the list of surveys within the event for which the user has a pending submission and put them in a set - then you can just check here whether the survey is in that set.

indico/modules/events/surveys/templates/display/survey_questionnaire.html
```diff
@@ -33,7 +33,17 @@
         {% endcall %}
     {% endfor %}
     {% call form_footer(form) %}
-        <input class="i-button big highlight" type="submit" value="{% trans %}Save answers{% endtrans %}">
+        {% if partial_completion %}
+            <a class="i-button big js-save-answers"
+               data-href="{{ url_for('.display_save_answers', survey) }}"
+               data-params-selector="#survey-questionnaire input[name^=question_]:checked,
+                                     #survey-questionnaire [name^=question_]:not(:radio):not(:checkbox)"
+               data-method="POST"
```
ThiefMaster: I'd move this one up (so method is right below href). I think we do it like that everywhere else

indico/modules/events/surveys/templates/management/email_dialog.html
```diff
@@ -0,0 +1,3 @@
+{% from 'forms/_form.html' import simple_form %}
+
+{{ simple_form(form, submit=_('Send'), disabled_until_change=false) }}
```
ThiefMaster: You don't need a template for this. `jsonify_form()` is your friend ;)

indico/modules/events/surveys/templates/management/survey.html
```diff
@@ -1,6 +1,38 @@
...
+    {% if survey.private and survey.state.name in ('active_and_clean', 'active_and_answered') %}
+        <div class="action-box highlight">
+            <div class="section">
+                <div class="icon icon-lock"></div>
+                <div class="text">
+                    <div class="label">
+                        {%- trans %}The survey is private{% endtrans -%}
+                    </div>
+                    {% trans -%}
+                        The survey is accessible only by people with this
```
ThiefMaster: fits nicely in one line with trans/endtrans

indico/modules/events/surveys/templates/management/survey.html
```diff
@@ -1,6 +1,38 @@
...
+                    {%- endtrans %}
+                    <a href="{{ url_for('.display_survey_form', survey, token=survey.uuid, _external=True) }}"
```
ThiefMaster: `true`

Also, check [this](https://github.com/indico/indico/blob/master/indico/modules/events/registration/models/registrations.py#L280) on how you could add another locator so you could do `survey.locator.uuid` here and then omit `token=survey.uuid`

indico/modules/events/surveys/templates/management/survey.html
```diff
@@ -1,6 +1,38 @@
...
+                       class="js-copy-to-clipboard"
+                       title="{% trans %}Copy link to clipboard{% endtrans %}"
+                       data-clipboard-text="{{ url_for('.display_survey_form', survey, token=survey.uuid, _external=True) }}"
+                       data-clipboard-action="copy">
+                        {%- trans %}link{% endtrans -%}
```
ThiefMaster: That's not a good way of doing i18n. Never build a sentence that spans multiple trans tags - otherwise you basically assume that "link" is the last word.

 Check [here](https://github.com/indico/indico/blob/devel/indico/modules/events/management/templates/_management_frame.html#L8-L11) for a way of doing it better

indico/modules/events/surveys/controllers/display.py
mvidalgarcia: I don't know, do we? The user can still access the incomplete submission through the emailed link, but maybe would be appropriate to show the unfinished submission in the survey list.

indico/modules/events/surveys/templates/display/survey_list.html
mvidalgarcia: Updated in https://github.com/indico/indico/pull/2873/commits/583143f2e1062cc0906e01458886738c759e0581

indico/modules/events/surveys/forms.py
DavidAndreev: So I guess I should update that 
form too, with the recipients field and copy to sender option right ? Cause I didn't know about that form that we send emails there too.

indico/modules/events/surveys/templates/management/survey.html
DavidAndreev: I don't really understand how it works, what should be in that dictionary that I would return form the locator ? in that example there is `return dict(self.registration_form.locator, token=self.uuid)` but what would be the first argument in my case ?

DavidAndreev:
Updated

indico/modules/events/surveys/controllers/display.py
```diff
@@ -84,6 +88,7 @@ def _checkParams(self, params):
                        .options(joinedload('submissions'),
                                 joinedload('sections').joinedload('children'))
                        .one())
+        self.submission = session.user.survey_submissions.filter_by(survey=self.survey, is_submitted=False).first()
```
ThiefMaster: :warning: This will fail if the user is not logged in (which is perfectly valid if the survey doesn't require a user)

mvidalgarcia:
```python
self.submission = (session.user.survey_submissions.filter_by(survey=self.survey, is_submitted=False).first()  
                   if session.user else None)
```
Would this be ok?

ThiefMaster: yes, that'd be my approach too

indico/modules/events/surveys/controllers/display.py
pferreir: I agree with @ThiefMaster on this. If you've started it already, then it's useful that you can continue it without having to use the link again.

indico/modules/events/surveys/__init__.py
```diff
@@ -62,8 +63,10 @@ def _get_active_surveys(event):
     if not event.has_feature('surveys'):
         return []
     from indico.modules.events.surveys.models.surveys import Survey
+    private_criterion = ~Survey.private | Survey.submissions.any(
```
ThiefMaster: Use `db.or_()` when you need to wrap it. It's quite ugly with the `|` operator. Or simply split it:


```
user_pending_criterion= ~SurveySubmission.is_submitted & (SurveySubmission.user == session.user)
private_criterion = ~Survey.private | Survey.submissions.any(user_pending_criterion)
```

actually, if there's no user you can skip the check:

```
private_criterion = ~Survey.private
if session.user:
    user_pending_criterion = ~SurveySubmission.is_submitted & (SurveySubmission.user == session.user)
    private_criterion |= Survey.submissions.any(user_pending_criterion)
```

not sure if `|=` works or you need `private_criterion = private_criterion | ...` (depends on how well sqlalchemy overloads this ;))

indico/modules/events/surveys/controllers/display.py
```diff
@@ -52,8 +52,10 @@ def view_class(self):
 
 class RHSurveyList(RHSurveyBaseDisplay):
     def _process(self):
+        private_criterion = ~Survey.private | Survey.submissions.any(
```
ThiefMaster: same here

ThiefMaster: if it's the same query/criteria consider adding a util function to get the query/criteria.

PR #2882, opened 2017-03-17
Different notification days before in room booking

DavidAndreev:


indico/modules/rb/models/reservation_occurrences_test.py
```diff
@@ -410,37 +410,83 @@ def test_overlaps_self(dummy_occurrence, skip_self):
...
     occurrence = create_occurrence(start_dt=datetime.now() + relativedelta(days=2))
-    dummy_room.notification_before_days = notification_window
+    occurrence.reservation.repeat_frequency = RepeatFrequency.WEEK
+    dummy_room.notification_before_repeating = notification_window
     db.session.flush()
     assert_is_in_notification_window(occurrence, expected, expected_with_exclude)
 
 
 @pytest.mark.parametrize(('notification_window', 'expected', 'expected_with_exclude'), (
+        (1, False, False,),
```
ThiefMaster: too much indentation (pycharm likes to mess it up in this specific case)

ThiefMaster:
This one needs a very thorough review of the logic - it's quite tricky.

indico/modules/rb/models/rooms.py
```diff
@@ -696,22 +699,32 @@ def is_owned_by(self, user):
         return user in GroupProxy.get_named_default_group(manager_group)
 
     @hybrid_method
-    def is_in_digest_window(self, exclude_first_day=False):
+    def is_in_digest_window(self, reservation, exclude_first_day=False):
```
ThiefMaster: I really dislike the fact that the regular and expression version of this hybrid method are different - also, the expression one will most likely retrieve the cross product between Room and Reservation when used in a query where Room and Reservation are not explicitly joined.

However, when I was with @DavidAndreev we couldn't find a cleaner solution for this quickly. Or at least none that wouldn't require rewriting the logic to handle these notifications (which, honestly, might not be too bad - even with unit tests the current notification logic is not that maintainable and will probably cause someone headaches in the future).

PR #2886, opened 2017-03-20
Show abstract state using badges in abstract list

### LINKED ISSUE #2880: Show review state in abstract list
Currently the state is just "Submitted" and there is no way of seeing whether the abstract is completely new, under review or has already been reviewed in all its tracks from the abstract list without going to each abstract's timeline.

DavidAndreev:
Closes #2880

### DIFF
--- a/indico/modules/events/abstracts/templates/management/_abstract_list.html
+++ b/indico/modules/events/abstracts/templates/management/_abstract_list.html
@@ -29,6 +29,27 @@
     </td>
 {%- endmacro %}
 
+{% macro _render_abstract_state(abstract) %}
+    <div class="i-tag outline semantic-text state-badge {{ abstract.public_state.css_class }}">
+        {% if abstract.public_state.name == 'under_review' and abstract.can_judge(session.user) %}
+            {% trans count=abstract.reviews|length -%}
+                {{- count }} review
+            {%- pluralize -%}
+                {{- count }} reviews
+            {%- endtrans %}
+            (
+                {%- trans count=abstract.reviewed_for_tracks|length -%}
+                    {{- count }} track
+                {%- pluralize -%}
+                    {{- count }} tracks
+                {%- endtrans -%}
+            )
+        {% else %}
+            {{- abstract.public_state.title -}}
+        {% endif %}
+    </div>
+{% endmacro %}
+
 {% macro render_abstract_list(abstracts, dynamic_columns, static_columns, total_abstracts, context_track=none, reviewed_abstracts=none) %}
     {% if abstracts %}
         <form method="POST">
@@ -96,8 +117,8 @@
                                 </td>
                                 {% for item in static_columns %}
                                     {% if item.id == 'state' %}
-                                        <td class="i-table" data-searchable="{{ abstract.state.title | lower }}">
-                                            {{- abstract.state.title -}}
+                                        <td class="i-table" data-searchable="{{ abstract.public_state.title | lower }}">
+                                            {{ _render_abstract_state(abstract) }}
                                             {% if context_track is not none and abstract.state.name == 'accepted' and abstract.accepted_track != context_track  %}
                                                 <span class="icon-warning"></span>
                                             {% endif %}
--- a/indico/htdocs/sass/modules/abstracts/_abstract_list.scss
+++ b/indico/htdocs/sass/modules/abstracts/_abstract_list.scss
@@ -24,3 +24,7 @@
         color: #777;
     }
 }
+
+.state-badge {
+    display: inline-block;
+}


mvidalgarcia:
I think a screenshot in this PR would help a lot! :wink:

DavidAndreev:
abstr list

indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -96,8 +96,17 @@
                                 </td>
                                 {% for item in static_columns %}
                                     {% if item.id == 'state' %}
-                                        <td class="i-table" data-searchable="{{ abstract.state.title | lower }}">
-                                            {{- abstract.state.title -}}
+                                        <td class="i-table" data-searchable="{{ abstract.public_state.title | lower }}">
+                                            {% if abstract.public_state.name == 'under_review' %}
+                                                <div class="i-tag state-badge under-review">
+                                                    {{- abstract.reviews|length|string + " " + ngettext("review on", "reviews on", abstract.reviews|length) + " "  +
```
ThiefMaster: I think this line is super ugly and needs to be improved quite a bit ;)

- Don't use `+` to build strings (you can just `}} {{` in a template to add that space)
- Don't use multiple gettext calls in the same sentence. That doesn't work well depending on the structure it has in another language
- There's {% trans %}{% pluralize %}{% endtrans %}`

DavidAndreev:
So I changed the badges as @nop33 suggested so they don't drag too much attention. Tell me if you liked it better before.
outline
reviwers

indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -96,8 +96,23 @@
                                 </td>
                                 {% for item in static_columns %}
                                     {% if item.id == 'state' %}
-                                        <td class="i-table" data-searchable="{{ abstract.state.title | lower }}">
-                                            {{- abstract.state.title -}}
+                                        <td class="i-table" data-searchable="{{ abstract.public_state.title | lower }}">
+                                            <div class="i-tag outline semantic-text {{ abstract.public_state.css_class }} state-badge">
```
pferreir: Maybe this could be moved inside a macro? Not that we're going to reuse it, but at least it doesn't look like a gigantic list of ifs and elses.

indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -96,8 +96,23 @@
...
+                                                {% if abstract.public_state.name == 'under_review' and abstract.can_judge(session.user) %}
+                                                    {% trans count=abstract.reviews|length -%}
+                                                        {{- count }} review on
```
ThiefMaster: This is not very good i8n since "x review on" and "x tracks" are translated separately. Unfortunately there is no way of doing it better with this structure. However, why not use parentheses instead of "on"?

"X review(s) (Y track(s))"

Like this you can translate/pluralize "X review(s)" and "Y track(s)" separately but it should be ok in any language

ThiefMaster:
Check whether we now spam queries for each abstract since we access `.reviews` in every row (go to an event with many abstracts and check in dblog if you have (much) more queries there than in an event with few abstracts)

DavidAndreev:
Yes I checked it, its ok.

indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -29,6 +29,25 @@
...
 
+{% macro _render_abstract_state(abstract) %}
+    <div class="i-tag outline semantic-text {{ abstract.public_state.css_class }} state-badge">
+        {% if abstract.public_state.name == 'under_review' and abstract.can_judge(session.user) %}
+            {% trans count=abstract.reviews|length -%}
+                {{- count }} review
+            {%- pluralize -%}
+                {{- count }} reviews
+            {%- endtrans %}
+            ({% trans count=abstract.reviewed_for_tracks|length -%}
```
ThiefMaster: I'd wrap after the `(` and then use `{%- trans` - like this it always looks like incorrect indentation before noticing the `(`

indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -29,6 +29,25 @@
     </td>
 {%- endmacro %}
 
+{% macro _render_abstract_state(abstract) %}
+    <div class="i-tag outline semantic-text {{ abstract.public_state.css_class }} state-badge">
```
ThiefMaster: I'd keep the variable class at the very end (ie after `state-badge`)

PR #1738, opened 2015-01-21
1586 fix booking days limit—1722 display nonbookable periods

jacquesd:
- check start date in query string of single page booking
- add new bar to the calendar to show days where booking is not possible.
- limit choice of days in 3-step booking 
- fixes #1586  
- display and check nonbookable periods in the view
- fixes #1722 
- sort days with (pre-)conflicts in the warning msg

> **Note:** This PR will close both the issue #1586 and the issue #1722.
> As a result, the PR #1609 for the issue #1586 has been closed preemptively.

PR #2887, opened 2017-03-21
Fix booking days limit / unavailability periods

pferreir:
Supersedes #1738

### DIFF
--- a/indico/modules/rb/views/calendar.py
+++ b/indico/modules/rb/views/calendar.py
@@ -349,9 +349,7 @@ class Bar(Serializer):
 
     @property
     def blocking_data(self):
-        if not self.blocking:
-            return None
-        elif self.blocking:
+        if self.blocking:
             return {
                 'id': self.blocking.id,
                 'creator': self.blocking.created_by_user.full_name,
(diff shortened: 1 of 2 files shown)


indico/modules/rb/controllers/user/reservations.py
```diff
@@ -375,6 +375,10 @@ def _make_form(self):
                 start_date = datetime.strptime(request.args['start_date'], '%Y-%m-%d').date()
             except ValueError:
                 pass
+            else:
+                if not self._room.check_advance_days(start_date, user=session.user, quiet=True):
```
ThiefMaster: elif

pferreir: With `try..except`? 😉

ThiefMaster: oops :monkey:

indico/modules/rb/views/calendar.py
```diff
@@ -325,14 +349,24 @@ def forReservation(self):
...
+        elif self.nb_period:
+            return {
+                'id': None,
+                'creator': None,
+                'reason': 'Unavailable from {0} until {1}.'.format(format_date(self.nb_period.start_dt),
+                                                                   format_date(self.nb_period.end_dt)),
+                'blocking_url': None,
+                'type': 'nonbookable'
+            }
+        return None
```
ThiefMaster: i'd move it to an `else:` block to have all the returns on the same level

PR #2890, opened 2017-03-22
Strip whitespace in {% trans %} Jinja tags

pferreir:


indico/util/translations.py
```diff
@@ -17,11 +17,26 @@
...
+
+
+# TODO: Remove this once there's proper support in upstream Jinja
+# https://github.com/pallets/jinja/pull/683
+def jinja2_babel_extract(fileobj, keywords, comment_tags, options):
+    """
+    Custom babel_extract for Jinja.
+
+    Hooks on to Jinja's babel_extract and handles
+    whitespace within `{% trans %}` tags.
```
ThiefMaster: double backticks

indico/util/translations.py
```diff
@@ -17,11 +17,26 @@
...
     def wrapper(*args, **kwargs):
         return encode_if_unicode(fn(*args, **kwargs))
     return wrapper
+
+
+# TODO: Remove this once there's proper support in upstream Jinja
+# https://github.com/pallets/jinja/pull/683
+def jinja2_babel_extract(fileobj, keywords, comment_tags, options):
+    """
+    Custom babel_extract for Jinja.
```
ThiefMaster: fits in the previous line ;)

setup.py
```diff
@@ -75,6 +75,9 @@
         'global_settings = indico_zodbimport.modules.global_settings:GlobalSettingsImporter',
         'badges_posters = indico_zodbimport.modules.badges_posters:BadgePosterImporter',
         'event_data = indico_zodbimport.modules.event_data:EventDataImporter',
+    ],
+    'babel.extractors': [
+        'jinja2_indico = indico.util.translations:jinja2_babel_extract'
```
ThiefMaster: You can omit that and add this to `babel.cfg` instead in the `[extractors]` section:

    jinja2_indico = indico.util.translations:jinja2_babel_extract

indico/util/translations.py
pferreir: I really don't like the asymmetry created by the lack of newline. But anyway, it's true that that's the style we mostly use.

setup.py
pferreir: You're right. Didn't recall that.

ThiefMaster:
unittests are broken

pferreir:
> unittests are broken

Apparently not by this... weird.

PR #2891, opened 2017-03-23
Add customization settings

ThiefMaster:
- [x] Jinja template overrides
- [x] CSS
- [x] JS

### DIFF
--- a/indico/indico.conf.sample
+++ b/indico/indico.conf.sample
@@ -452,10 +452,10 @@ MaxUploadFileSize = 0
 # Customization
 # ------------------------------------------------------------------------------
 
-# It is possible to override specific templates for advanced customizations.
-# When using this, be advised that depending on the modifications you perform
-# things may break after an Indico update.  Make sure to test all your
-# modifications whenever you update Indico!
+# It is possible to override specific templates and add CSS and JavaScript
+# for advanced customizations.  When using this, be advised that depending on
+# the modifications you perform things may break after an Indico update.  Make
+# sure to test all your modifications whenever you update Indico!
 #
 # Set the base directory for customizations.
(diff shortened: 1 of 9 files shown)


indico/web/flask/templating.py
```diff
@@ -194,6 +196,43 @@ def call_template_hook(*name, **kwargs):
...
+            try:
+                orig_path = rv[1]
+            except TemplateNotFound:
+                orig_path = None
+            self.logger.debug('Customizable: %s (original: %s, reference: ~%s)', path, orig_path, template)
+        return rv
+
+    def get_source(self, environment, template):
+        path = posixpath.join(*split_template_path(template))
+        if template[0] == '~':
```
ThiefMaster: I don't like this very much but being able to inherit from the original template is important - and from within a loader we have no way of knowing from where the template is being rendered.

indico/indico.conf.sample
```diff
@@ -446,3 +446,38 @@ MaxUploadFileSize = 0
...
+# However, instead of copying templates it is better to use Jinja inheritance
+# where possible. To make this easier the log entry will also contain a
+# "reference" path that can be used to reference the original template from the
+# customized one.
+#CustomizationDebug = False
+#
+# To include custom CSS and JavaScript, simply put *.scss and *.js files into
+# `<CustomizationDir>/scss` / `<CustomizationDir>/js`.  If there are multiple
+# files, they will be included in alphabetical order so prefixing them with
+# a number (e.g. `00-base.scss`, `10-events.scss`) is a good idea.
```
pferreir: Wondering if this shouldn't be put on the wiki. Or even in the docs.

ThiefMaster: Yeah, just wanted to have it *somewhere* for now

PR #2892, opened 2017-03-23
Get rid of i18nformat(...)

pferreir:
Good riddance.

### DIFF
--- a/indico/legacy/PDFinterface/conference.py
+++ b/indico/legacy/PDFinterface/conference.py
@@ -42,7 +42,7 @@ from indico.modules.events.tracks.settings import track_settings
 from indico.modules.events.util import create_event_logo_tmp_file
 from indico.util import json
 from indico.util.date_time import format_date, format_datetime, format_time, format_human_timedelta
-from indico.util.i18n import i18nformat, ngettext, _
+from indico.util.i18n import ngettext, _
 from indico.util.string import html_color_to_rgb, to_unicode, truncate
 from indico.legacy.PDFinterface.base import (escape, PDFLaTeXBase, PDFBase, PDFWithTOC, Paragraph, Spacer, PageBreak,
                                      modifiedFontSize)
@@ -557,7 +557,7 @@ class TimeTablePlain(PDFWithTOC):
                                 size=modifiedFontSize(9, self._fontsize), color=(0.5, 0.5, 0.5), align="left",
                                 lineSpacing=0.3, maximumWidth=maxi)
         c.drawCentredString(self._PAGE_WIDTH / 2.0, 0.5 * cm,
-                            i18nformat(""" _("Page") {} """).format(doc.page + self._firstPageNumber - 1))
+                            _('Page {}').format(doc.page + self._firstPageNumber - 1))
         c.drawRightString(self._PAGE_WIDTH - 1 * cm, self._PAGE_HEIGHT - 1 * cm, doc.getCurrentPart())
         c.restoreState()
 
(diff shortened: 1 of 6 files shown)


indico/legacy/PDFinterface/conference.py
```diff
@@ -624,11 +624,8 @@ def _processContribution(self, contrib, l):
         if self._ttPDFFormat.showContribAbstract():
             speaker_list = [self._get_speaker_name(spk) for spk in contrib.speakers]
             if speaker_list:
-                if len(speaker_list) == 1:
-                    speaker_title = i18nformat(""" _("Presenter"): """)
-                else:
-                    speaker_title = i18nformat(""" _("Presenters"): """)
-                speaker_content = speaker_title + ", ".join(speaker_list)
+                speaker_title = u' {}: '.format(ngettext('Presenter', 'Presenters', len(speaker_list)))
```
ThiefMaster: You probably need `u''` for the two strings too - otherwise you are interpolating bytestrings (which may be non-ascii when translated) into a unicode string.

ThiefMaster: I'd test those places with `DebugUnicode = 1` in indico.conf

indico/legacy/webinterface/pages/base.py
```diff
@@ -175,9 +175,9 @@ def _getHTMLHeader( self ):
 
         area=""
         if self.MANAGEMENT or isinstance(self._rh, RHModificationBaseProtected):
-            area=i18nformat(""" - _("Management area")""")
+            area = ' - {}'.format(_("Management area"))
```
ThiefMaster: `' - ' + _(..)` seems more readable here

indico/legacy/webinterface/wcomponents.py
```diff
@@ -380,7 +381,8 @@ def getVars( self ):
                     selected = "selected"
             else:
                 selectedSession = "all"
-            sessions = [ i18nformat(""" <select name="showSession" onChange="document.forms[0].submit();" style="font-size:8pt;"><option value="all" %s>- -  _("all sessions") - -</option> """)%selected]
+            sessions = ["""<select name="showSession" onChange="document.forms[0].submit();" style="font-size:8pt;">
```
ThiefMaster: `document.forms[0]`... because it's totally guaranteed that this is the first form. Can we use `this.form.submit()` instead?

indico/util/i18n.py
```diff
@@ -286,19 +286,6 @@ def parse_locale(locale):
...
-    """
-
-    # this is a bit of a dirty hack, but cannot risk emitting lazy proxies here
-    if not session.lang:
-        set_best_lang()
-
-    return RE_TR_FUNCTION.sub(lambda x: _(next(y for y in x.groups() if y is not None)), text)
-
-
 def extract(fileobj, keywords, commentTags, options):
```
ThiefMaster: this can also go away

ThiefMaster: same for the babel.cfg entry

indico/legacy/webinterface/pages/base.py
pferreir: Oh, come on.

indico/legacy/webinterface/wcomponents.py
pferreir: I didn't even look at that.

ThiefMaster: yeah, just noticed it. maybe not worth fixing now since this will eventually be rewritten anyway

PR #2893, opened 2017-03-24
Allow multiline for review/abstract questions

meluru:
Disable comments for accepted abstracts and papers

### DIFF
--- a/indico/htdocs/js/indico/widgets/multiple_items_widget.js
+++ b/indico/htdocs/js/indico/widgets/multiple_items_widget.js
@@ -241,6 +241,15 @@
                 });
                 initNumberErrorMessage(numberInput, col.min, col.max);
                 return {html: numberInput};
+            } else if (col.type === 'textarea') {
+                return {
+                    html: $('<textarea>', {
+                        'class': 'js-table-input multiline',
+                        'value': item ? item[col.id] : '',
+                        'placeholder': col.caption,
+                        'data-required': col.required
+                    })
+                };
             } else { // Assuming the type is 'text'
                 return {
                     html: $('<input>', {
@@ -312,7 +321,11 @@
                 $('<td>', {class: 'sort-handle'}).appendTo(row);
             }
             options.columns.forEach(function(col) {
-                $('<td>', makeColData(item, col)).appendTo(row);
+                var column = $('<td>', makeColData(item, col));
+                if (col.type === 'textarea') {
+                    column.addClass('multiline');
+                }
+                column.appendTo(row);
             });
             $('<td>', {
                 html: item ? deleteButton.clone().add(editButton.clone())
@@ -325,7 +338,11 @@
 
         function updateRow(row, editMode, moveTooltips) {
             row.children('td:not(.sort-handle):not(.js-action-col)').each(function(i) {
-                $(this).replaceWith($('<td>', makeColData(data[row.index()], options.columns[i], editMode)));
+                var column = $('<td>', makeColData(data[row.index()], options.columns[i], editMode));
+                if (options.columns[i].type === 'textarea') {
+                    column.addClass('multiline');
+                }
+                $(this).replaceWith(column);
             });
             if (editMode) {
                 row.children('.js-action-col').html(cancelButton.clone().add(saveButton.clone()));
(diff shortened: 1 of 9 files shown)


indico/htdocs/js/indico/widgets/multiple_items_widget.js
```diff
@@ -202,6 +202,8 @@
                             'text-align': 'center'
                         }
                     };
+                } else if (col.type === 'textarea') {
+                    return {html: item[col.id].replace(/(?:\r\n|\r|\n)/g, '<br>')};
```
ThiefMaster: I think replacing `\n` is enough (the `\r` - even if there is one - is just extra whitespace that gets collapsed, and all systems that used bare `\r` (legacy MacOS) are obsolete/unsupported)

Actually, you could even use `white-space: pre-line` on the container instead of converting anything. This would actually be better because right now any HTML in the content will be evaluated here - not such a good idea considering we don't allow HTML but just linebreaks ;)

indico/htdocs/sass/partials/_forms.scss
```diff
@@ -205,6 +205,7 @@ $i-form-button-line-height: 1.8;
     textarea {
         @extend %form-field-input;
         height: auto;
+        width: 100%;
```
ThiefMaster: Make sure this doesn't break the layout anywhere - it seems to affect almost every textarea in indico.

indico/modules/events/fields.py
```diff
@@ -267,7 +267,8 @@ def __init__(self, *args, **kwargs):
 class ReviewQuestionsField(MultipleItemsField):
     def __init__(self, *args, **kwargs):
         self.extra_fields = kwargs.pop('extra_fields', [])
-        self.fields = [{'id': 'text', 'caption': _("Question"), 'type': 'text', 'required': True}] + self.extra_fields
```
ThiefMaster: No strong opinion but IMO this looks better since it doesn't wrap inside the dict
```
        self.fields = ([{'id': 'text', 'caption': _("Question"), 'type': 'textarea', 'required': True}] + 
                       self.extra_fields)
```

indico/htdocs/js/indico/widgets/multiple_items_widget.js
pferreir: Yeah, I agree, `pre-line` would probably be the best.

indico/htdocs/sass/partials/_forms.scss
pferreir: Even if nothing breaks, I'd move this into a much more specific selector.

indico/modules/events/fields.py
pferreir: Agreed, but purely personal.

indico/modules/events/templates/reviews/rating_widget.html
```diff
@@ -8,12 +8,12 @@
             </span>
         </div>
         <div class="question-text f-self-stretch">
-            {{ field.label.text }}
+            {{ field.label.text | trim | replace('\n', '<br>'|safe) }}
```
pferreir: Maybe we can use `pre-line` here too?

meluru:
Updated

indico/modules/events/abstracts/templates/reviewing/judgment.html
```diff
@@ -109,11 +109,11 @@
                             <div id="question-list-{{ track.id }}" class="details-content weak-hidden">
                                 <div class="details-content-inner review-questions">
                                     {% for question, score in ratings[track.id].iteritems() %}
-                                        <div class="details-content-row flexrow f-a-center">
+                                        <div class="details-content-row flexrow">
                                             <div>
                                                 <span class="question-index">{{ loop.index }}</span>
                                             </div>
-                                            <div class="question-text f-self-stretch">{{ question.text }}</div>
+                                            <div class="question-text f-self-stretch">{{- question.text -}}</div>
```
ThiefMaster: Why? There's no whitespace to strip here...

indico/modules/events/templates/reviews/rating_widget.html
```diff
@@ -8,12 +8,12 @@
             </span>
         </div>
         <div class="question-text f-self-stretch">
-            {{ field.label.text }}
-            {%- if field.flags.required %}
+            {{- field.label.text -}}
+            {%- if field.flags.required -%}
```
ThiefMaster: Why truncate whitespace here? IMO it looks quite weird with no space between the question and the asterisk

meluru:
Updated

PR #2894, opened 2017-03-24
Remove CERN themes from the core

ThiefMaster:
Added as a CERN plugin in https://gitlab.cern.ch/indico/indico-plugins-cern/merge_requests/11


indico/modules/events/timetable/views/__init__.py
```diff
@@ -132,10 +132,14 @@ def inject_meeting_body(event, **kwargs):
     days = [(day, list(e)) for day, e in groupby(entries, lambda e: e.start_dt.astimezone(event_tz).date())]
     theme_id = view if view and view in theme_settings.themes else event.theme
     theme = theme_settings.themes[theme_id]
-    tt_tpl = theme.get('tt_template', theme['template'])
+    plugin = theme.get('plugin')
+    tpl_name = theme.get('tt_template', theme['template'])
+    tt_tpl = (plugin.name + tpl_name
+              if plugin and tpl_name[0] == ':'
```
pferreir: Maybe this is one of those cases where parenthesis would help to understand the precedence of the `+` operator over the `if`.

ThiefMaster: Thought about it but then assumed the wrapping already makes it clear

pferreir: The wrapping per se doesn't make it clear, IMHO, but anyway, not very important.

ThiefMaster: updated

PR #2897, opened 2017-03-28
Move core templates to Jinja

ThiefMaster:
- [x] HTMLHeader
- [x] Header
- [x] Footer
- [x] EventHeader
- [x] EventFooter
- [x] Event rendering - clean this up once and for all!
- [x] Theme handling (print url)
- [x] misc stuff
- [x] *.CERN.tpl


indico/modules/core/blueprint.py
```diff
@@ -25,4 +25,4 @@
 _bp.add_url_rule('/admin/settings/', 'settings', RHSettings, methods=('GET', 'POST'))
 
 # TODO: replace with an actual admin dashboard at some point
-_bp.add_url_rule('/admin/', 'dashboard', view_func=redirect_view('.settings'))
+_bp.add_url_rule('/admin/', 'admin_dashboard', view_func=redirect_view('.settings'))
```
ThiefMaster: `core.dashboard` doesn't sound like an admin thing

indico/web/templates/indico_base.html
```diff
@@ -0,0 +1,63 @@
...
+    is only called from a WP so just passing the contents is much easier
+    #}
+    {{ body|safe }}
+</body>
+</html>
+<!--
+Worker: {{ indico_config.WorkerName }}
+{%- if config.DEBUG or (session.user and session.user.is_admin) %}
+{%- if config.DEBUG %}
+Queries: {{ g.get('sql_query_count', 0) }}
```
ThiefMaster: Should be ok. We'd just need to register `@listens_for(Engine, 'after_cursor_execute')` all the time or maybe split query counting from the db logger. adding it to my TODO list for later

PR #2902, opened 2017-04-10
Improve style of event display top bar

nop33:
From

image

to

image

### DIFF
--- a/indico/modules/events/templates/header.html
+++ b/indico/modules/events/templates/header.html
@@ -1,31 +1,31 @@
 {%- macro _render_nav_bar() -%}
     {% set rel = event.get_relative_event_ids() %}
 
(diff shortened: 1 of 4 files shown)


ThiefMaster:
I don't like the thing blue line  on top very much

indico/modules/events/templates/header.html
```diff
@@ -1,31 +1,31 @@
 {%- macro _render_nav_bar() -%}
     {% set rel = event.get_relative_event_ids() %}
 
-    <a class="i-button icon-home" href="{{ url_for_index() }}"
+    <a class="i-button text-color subtle icon-home" href="{{ url_for_index() }}"
```
ThiefMaster: I'd rather import these two classes in the `.button-bar` `.i-button`  to keep everything in one place

PR #2905, opened 2017-04-11
Fix/papers

nop33:


indico/modules/events/papers/notifications.py
```diff
@@ -57,20 +57,21 @@ def notify_paper_review_submission(review):
 
 def notify_paper_judgment(paper, reset=False):
     event = paper.event_new
-    receiver = paper.last_revision.submitter
-    template = None
+    receivers = ([x for x in paper.contribution.submitters if x.email] if paper.last_revision.submitter.is_janitor
```
ThiefMaster: i think it's usually called `recipients`

ThiefMaster: why not use authors instead of submitters? material submission rights aren't taken into account for anything else in paper reviewing (iirc we discussed this a few times already ;)) so i don't see why they should be used here

indico/modules/events/papers/templates/reviewing/timeline.html
```diff
@@ -139,7 +139,8 @@ <h3 class="f-self-stretch">
             <div class="i-timeline-item-box header-indicator-left {{ header_class }}" id="revision-info-{{ revision.id }}">
                 <div class="i-box-header flexrow">
                     <div class="f-self-stretch">
-                        {% trans name=revision.submitter.name, number=revision.number -%}
+                        {% set name = revision.submitter.name if not revision.submitter.is_janitor else _('User') %}
```
ThiefMaster: 'A user'

"User submitted paper revision..." sounds quite strange

indico/modules/events/papers/notifications.py
nop33: @pferreir ?

pferreir: Authors makes more sense indeed.

PR #2906, opened 2017-04-13
Fix booked for user not shown

DavidAndreev:
Its still not finished, I get this error when radio field is not checked and trying to Check conflicts. I can continue working on this when I come back
err


DavidAndreev:
It should be fixed now, also when modifying the reservation.

PR #2912, opened 2017-04-24
Fix regform accommodation fields

mvidalgarcia:


indico/migrations/versions/201704241604_bc4e7682e7df_fix_regform_old_accommodation_fields.py
```diff
@@ -0,0 +1,58 @@
...
+def upgrade():
+    if context.is_offline_mode():
+        raise Exception('This upgrade is only possible in online mode')
+    conn = op.get_bind()
+
+    field_data_query = conn.execute("""
+        SELECT field_id, ffd.id, fi.is_required, ffd.versioned_data, fi.data
+        FROM event_registration.form_field_data ffd
+        JOIN event_registration.form_items fi ON fi.id = ffd.field_id
+        WHERE fi.input_type = 'accommodation' AND fi.is_deleted = FALSE
```
ThiefMaster: `AND NOT fi.is_deleted`

ThiefMaster: You need to `ORDER BY field_id` - otherwise the `groupby` below doesn't work properly.

> Generally, the iterable needs to already be sorted on the same key function.

https://docs.python.org/2/library/itertools.html#itertools.groupby

indico/migrations/versions/201704241604_bc4e7682e7df_fix_regform_old_accommodation_fields.py
```diff
@@ -0,0 +1,58 @@
...
+depends_on = None
+
+
+def upgrade():
+    if context.is_offline_mode():
+        raise Exception('This upgrade is only possible in online mode')
+    conn = op.get_bind()
+
+    field_data_query = conn.execute("""
+        SELECT field_id, ffd.id, fi.is_required, ffd.versioned_data, fi.data
```
ThiefMaster: I'd always qualify field names when more than one table is involed. i.e. `ffd.field_id` (or wherever it comes from) instead of just `field_id`

indico/migrations/versions/201704241604_bc4e7682e7df_fix_regform_old_accommodation_fields.py
```diff
@@ -0,0 +1,58 @@
...
+        FROM event_registration.form_field_data ffd
+        JOIN event_registration.form_items fi ON fi.id = ffd.field_id
+        WHERE fi.input_type = 'accommodation' AND fi.is_deleted = FALSE
+    """)
+
+    for field_id, data in itertools.groupby(field_data_query, attrgetter('field_id')):
+        choice_id = unicode(uuid.uuid4())
+        needs_caption = False
+        for (_, field_data_id, is_required, versioned_data, field_data) in data:
```
ThiefMaster: I'd just use `for row in data` and then use `row.field_data_id` etc. below.

indico/migrations/versions/201704241604_bc4e7682e7df_fix_regform_old_accommodation_fields.py
```diff
@@ -0,0 +1,58 @@
...
+            if not any(('is_no_accommodation' in choice) for choice in versioned_data['choices']):
+                needs_caption = True
+                no_acc_choice = {"is_no_accommodation": True,
+                                 "is_enabled": not is_required,
+                                 "price": 0,
+                                 "is_billable": False,
+                                 "places_limit": 0,
+                                 "placeholder": "Title of the \"None\" option",
```
ThiefMaster: If you use single quotes for all the keys/values like we usually do you don't even need to escape the inner double quotes ;)

indico/migrations/versions/201704241604_bc4e7682e7df_fix_regform_old_accommodation_fields.py
```diff
@@ -0,0 +1,58 @@
...
+                                 "id": choice_id}
+
+                versioned_data['choices'] += [no_acc_choice]
+                conn.execute("UPDATE event_registration.form_field_data SET versioned_data = %s::json WHERE id = %s",
```
ThiefMaster: Is the `::json` actually needed?

indico/migrations/versions/201704241604_bc4e7682e7df_fix_regform_old_accommodation_fields.py
```diff
@@ -0,0 +1,58 @@
...
+                             json.dumps(versioned_data), field_data_id)
+        if needs_caption:
+            field_data['captions'][choice_id] = 'No accommodation'
+            conn.execute("UPDATE event_registration.form_items SET data = %s::json WHERE id = %s",
+                         json.dumps(field_data), field_id)
```
ThiefMaster: Does this actually work? Don't you need to pass a tuple containing the values instead of separate arguments?

indico/migrations/versions/201704241604_bc4e7682e7df_fix_regform_old_accommodation_fields.py
mvidalgarcia: Not sure, I've seen it other revisions. I was trying it before using `json.dumps()` so maybe it's not needed anymore.

indico/migrations/versions/201704241604_bc4e7682e7df_fix_regform_old_accommodation_fields.py
mvidalgarcia: It seems to work because I tested it several times. I see that in other revisions we're using tuples though.

ThiefMaster: OK looks like [execute()][1] is quite flexible in what it accepts:

[1]: http://docs.sqlalchemy.org/en/latest/core/connections.html#sqlalchemy.engine.Connection.execute.params.*multiparams/**params

indico/migrations/versions/201704241604_bc4e7682e7df_fix_regform_old_accommodation_fields.py
```diff
@@ -0,0 +1,59 @@
...
+                needs_caption = True
+                no_acc_choice = {'is_no_accommodation': True,
+                                 'is_enabled': not row.is_required,
+                                 'price': 0,
+                                 'is_billable': False,
+                                 'places_limit': 0,
+                                 'placeholder': 'Title of the "None" option',
+                                 'id': choice_id}
+
+                row.versioned_data['choices'] += [no_acc_choice]
```
ThiefMaster: Why not simply `row.versioned_data['choices'].append(no_acc_choice)`?

ThiefMaster: Also, wouldn't it make more sense to `insert()` it at the beginning? I guess usually you want this to be the first option, not the last...

PR #2914, opened 2017-04-25
New notification system room booking

DavidAndreev:


indico/modules/rb/models/reservation_occurrences.py
```diff
@@ -289,22 +289,40 @@ def is_in_notification_window(self, exclude_first_day=False):
...
+        from indico.modules.rb import Reservation
         from indico.modules.rb import settings as rb_settings
+        from indico.modules.rb.models.reservations import RepeatFrequency
         from indico.modules.rb.models.rooms import Room
-        in_the_past = cast(self.start_dt, Date) < cast(func.now(), Date)
-        days_until_occurrence = cast(self.start_dt, Date) - cast(func.now(), Date)
-        notification_window = func.coalesce(Room.notification_before_days,
-                                            rb_settings.get('notification_before_days', 1))
+        in_the_past = cast(cls.start_dt, Date) < cast(func.now(), Date)
+        days_until_occurrence = cast(cls.start_dt, Date) - cast(func.now(), Date)
```
nop33: Why not storing the output of `cast(cls.start_dt, Date)` and `cast(func.now(), Date)` into variables? You use `cast` 4 times instead of 2 :smile:

indico/modules/rb/models/reservations.py
```diff
@@ -15,7 +15,7 @@
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
 from collections import defaultdict, OrderedDict
-from datetime import datetime, date
+from datetime import datetime
 from flask import session
```
nop33: I know you didn't change this line, but it should be grouped below.

indico/modules/rb/notifications/reservation_occurrences.py
```diff
@@ -72,60 +72,23 @@ def notify_rejection(occurrence):
...
-        raise ValueError("This reservation occurrence started in the past")
-
+def notify_upcoming_occurrences(occurrences):
     to_list = []
-    reservation_user = occurrence.reservation.booked_for_user
+    for occ in occurrences:
+        if occ.start_dt.date() < date.today():
+            raise ValueError("This reservation occurrence started in the past")
+
+    reservation_user = occurrences[0].reservation.booked_for_user
```
nop33: Any chance the `occurrences` is an empty list?

DavidAndreev: Not really, I skip to next occurence if we have an empty list and dont call the `notify_upcoming_occurrences`
```python
            if not same_user_occurrences:
                continue
            same_user_occurrences = sorted(same_user_occurrences, key=attrgetter('start_dt'))
            notify_upcoming_occurrences(same_user_occurrences)
```

indico/modules/rb/__init__.py
```diff
@@ -39,7 +39,8 @@
     'assistance_emails': [],
     'vc_support_emails': [],
     'notification_hour': 6,
-    'notification_before_days': 1,
+    'notification_before_repeating': 7,
+    'notification_before_single': 1,
```
pferreir: The default should actually be '2', IIRC

DavidAndreev:
Updated

indico/migrations/versions/201704251450_d34e73084377_remove_notification_for_responsible_.py
```diff
@@ -0,0 +1,25 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.drop_column('rooms', 'notification_for_responsible', schema='roombooking')
+
+
+def downgrade():
+    op.add_column('rooms', sa.Column('notification_for_responsible', sa.BOOLEAN(), autoincrement=False, nullable=True),
```
ThiefMaster: `sa.Boolean(), nullable=True, server_default='false', schema='roombooking'` - the autoincrement option is useless here but alembic always includes it when autogenerating such a downgrade. Also, you need to create it as nullable with a default and then use `op.alter_column()` to make it `nullable=False` right afterwards.

indico/modules/rb/templates/emails/reservations/reminders/upcoming_occurrence.txt
```diff
@@ -2,7 +2,13 @@
 {% block recipient %}{{ owner.first_name if owner else 'user' }}{% endblock %}
 
 {% block email_body -%}
-This is a reminder about a booking under your name:
+{% if occurrences | length == 1 -%}
+    This is a reminder about a booking under your name:
+{% else -%}
```
nop33: {%- else -%} :wink:

indico/modules/rb/tasks.py
```diff
@@ -111,9 +61,17 @@ def roombooking_occurrences():
...
     try:
         for occ in occurrences:
-            notify_upcoming_occurrence(occ)
-            occ.notification_sent = True
-            if occ.reservation.repeat_frequency == RepeatFrequency.DAY:
-                occ.reservation.occurrences.update({'notification_sent': True})
+            same_user_occurrences = []
+            for other in occurrences:
+                if (other.reservation.booked_for_user == occ.reservation.booked_for_user and
+                        other.first_start_date == occ.first_start_date and not other.notification_sent):
```
nop33: Shouldn't the indentation be like this?
```python
if (other.reservation.booked_for_user == occ.reservation.booked_for_user and
    other.first_start_date == occ.first_start_date and not other.notification_sent):
```

ThiefMaster: No, we use this option suggested in PEP8:

```
# Add some extra indentation on the conditional continuation line.
if (this_is_one_thing
        and that_is_another_thing):
    do_something()
```

indico/modules/rb/templates/emails/reservations/reminders/upcoming_occurrence.txt
DavidAndreev: But i need this empty space after the sentence, so it's easier to read.

indico/modules/rb/tasks.py
DavidAndreev: PyCharm says it shouldn't. So I don't know ..

DavidAndreev:
Updated

indico/modules/rb/models/reservation_occurrences.py
```diff
@@ -289,22 +289,42 @@ def is_in_notification_window(self, exclude_first_day=False):
...
+                                      else_=notification_window_repeating, value=Reservation.repeat_frequency)
         if exclude_first_day:
             return (days_until_occurrence < notification_window) & ~in_the_past
         else:
             return (days_until_occurrence <= notification_window) & ~in_the_past
+
+    @property
+    def first_start_date(self):
+        from indico.modules.rb.models.reservations import RepeatFrequency
+        if self.reservation.repeat_frequency == RepeatFrequency.DAY:
```
ThiefMaster: What about weekly etc?

ThiefMaster: Actually, why would the first occurrence of a daily booking not be on the first day of the booking?

DavidAndreev: Well, I'm grouping the reminders of bookings on the same day but sometimes it finds an occurrence which is not the first one in the reservation, therefore I need to know the first one of the daily booking reservation.

indico/modules/rb/tasks.py
```diff
@@ -111,9 +62,17 @@ def roombooking_occurrences():
 
     try:
         for occ in occurrences:
-            notify_upcoming_occurrence(occ)
-            occ.notification_sent = True
-            if occ.reservation.repeat_frequency == RepeatFrequency.DAY:
-                occ.reservation.occurrences.update({'notification_sent': True})
+            same_user_occurrences = []
+            for other in occurrences:
```
pferreir: I don't think this is the best/more efficient way of doing what you want to do. Have you thought of a group by query?

pferreir: And please pay attention to the number of queries that are generated by this task, it may need some joined/subquery loads.

DavidAndreev: No I haven't, since its celery task running every 8 hours I thought the performance is not that important. But I guess you are right ...

ThiefMaster: FWIW, I think the number of queries in this task was awful before. Dunno if it's worth to optimize this a lot - there's so much lazy=dynamic in roombooking code that you can't really joinedload/subqueryload much anyway.

pferreir: Well, my biggest complaint is not about query performance, it's about the way in which they're performed (cartesian product of all occurrences). We can probably use a GROUP BY query. And even if we do it client-side, we can use a dictionary to index the occurrences by user.

ThiefMaster:
Superseded by #2944

PR #2916, opened 2017-04-26
Fix override contrib type when accepting non-bulk

DavidAndreev:


indico/modules/events/abstracts/operations.py
```diff
@@ -211,7 +211,7 @@ def judge_abstract(abstract, abstract_data, judgment, judge, contrib_session=Non
     if judgment == AbstractAction.accept:
         abstract.state = AbstractState.accepted
         abstract.accepted_track = abstract_data.get('accepted_track')
-        if abstract_data.get('override_contrib_type'):
+        if abstract_data.get('override_contrib_type') or abstract_data.get('accepted_contrib_type'):
```
nop33: What is the `override_contrib_type` used for and how?

nop33: Nevermind, I think this fix is the right one.

nop33: Maybe it would be better if you swap these two checks, I think it's more frequent that we'll have an accepted_contrib_type than the override_contrib_type.

ThiefMaster: Not really worth it, they are both incredibly cheap.

DavidAndreev: Its a boolean that we pass when we select override type in bulk judgment 
overr
But when doing the single judgment we dont have this setting therefore it was always fallsy even when you selected a type to accept
ov

PR #2918, opened 2017-04-27
RegForm: allow modification end date to be set

pferreir:
There's currently no way to set it, even though it's included in the models (and has been migrated).


indico/modules/events/registration/forms.py
```diff
@@ -124,6 +124,10 @@ class RegistrationFormScheduleForm(IndicoForm):
                                    description=_("Moment when registrations will be open"))
     end_dt = IndicoDateTimeField(_("End"), [Optional(), LinkedDateTime('start_dt')], default_time=time(23, 59),
                                  description=_("Moment when registrations will be closed"))
+    modification_end_dt = IndicoDateTimeField(_("Modification deadline"), [Optional(), LinkedDateTime('end_dt')],
+                                              default_time=time(23, 59),
+                                              description=_("Deadline until which registration information can be "
+                                                            "modified"))
```
ThiefMaster: "Defaults to the end date" since this field is optional?

PR #2922, opened 2017-05-03
Hide/show "Modify registration" properly

mvidalgarcia:


indico/modules/events/registration/templates/display/event_header.html
```diff
@@ -3,7 +3,7 @@
         <div class="event-details-label align-top">{% trans %}Registration{% endtrans %}</div>
         <div class="event-details-content">
             {% for regform in regforms %}
-                {% set registration = session.user and regform.id in user_registrations %}
+                {% set registration = user_registrations[regform.id] if user_registrations[regform.id] else None %}
```
mvidalgarcia: I'm pretty sure this can be done in other way, is there anything similar to `.get()` in Jinja?

ThiefMaster: Yes, just use `.get()` - the code there is basically Python...

PR #2926, opened 2017-05-05
Do not silently ignore SMTP errors

pferreir:
Since we don't need anymore to retry transactions (as was the case with ZODB), we can just send any e-mails before the commit takes place.

### DIFF
--- a/indico/legacy/webinterface/rh/base.py
+++ b/indico/legacy/webinterface/rh/base.py
@@ -71,6 +71,7 @@ from indico.web.flask.util import ResponseUtil, url_for
 
 HTTP_VERBS = {'GET', 'PATCH', 'POST', 'PUT', 'DELETE'}
 
+logger = Logger.get('requestHandler')
 
 class RequestHandlerBase():
 
@@ -317,7 +318,7 @@ class RH(RequestHandlerBase):
                 except BuildError as e:
                     if current_app.debug:
                         raise
-                    Logger.get('requestHandler').warn('BuildError during normalization: %s', e)
+                    logger.warn('BuildError during normalization: %s', e)
                     raise NotFound
             else:
                 raise NotFound('The URL contains invalid data. Please go to the previous page and refresh it.')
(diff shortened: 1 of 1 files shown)


indico/legacy/webinterface/rh/base.py
```diff
@@ -551,13 +551,13 @@ def _do_process(self, profile):
...
 
     def _process_success(self):
-        Logger.get('requestHandler').info('Request successful')
-        # request is succesfull, now, doing tasks that must be done only once
+        logger = Logger.get('requestHandler')
+        logger.info('Request successful')
+        # request is succesfull, let's get rid of temporary files
         try:
-            GenericMailer.flushQueue(True)  # send emails
             self._deleteTempFiles()
```
ThiefMaster: This is actually obsolete and can go away - `self._tempFilesToDelete` is always an empty list in both RH and ServiceBase

indico/legacy/webinterface/rh/base.py
```diff
@@ -551,13 +551,13 @@ def _do_process(self, profile):
         return profile_name, res
 
     def _process_success(self):
-        Logger.get('requestHandler').info('Request successful')
-        # request is succesfull, now, doing tasks that must be done only once
+        logger = Logger.get('requestHandler')
+        logger.info('Request successful')
+        # request is succesfull, let's get rid of temporary files
```
ThiefMaster: successful

indico/legacy/webinterface/rh/base.py
```diff
@@ -587,6 +587,15 @@ def process(self, params):
                 self._check_auth(params)
                 profile_name, res = self._do_process(profile)
                 signals.after_process.send()
+
+                if self.commit:
+                    # ensure we fail early (before sending out e-mails)
+                    # in case there are DB constraint violations, etc...
+                    db.session.flush()
```
ThiefMaster: You could also call `db.enforce_constraints()` instead - this already flushes but also ensures that e.g. inconsistent-timetable errors are triggered

pferreir: Oh cool.

PR #2927, opened 2017-05-05
New Designer interface and back-sides

nop33:
* (Almost) Complete redesign of the UI for the category/event badge/poster designer.
* Ability to select/save a back side of a badge

> Specifications doc: https://gist.github.com/nop33/cf095188559b42a8fbc79883bf8250ba

### DIFF
--- a/indico/htdocs/js/indico/modules/designer.js
+++ b/indico/htdocs/js/indico/modules/designer.js
@@ -459,16 +459,14 @@
             template: {
                 width: templateDimensions.realWidth,
                 height: templateDimensions.realHeight,
+                background_position: $('input[name=bg-position]:checked').val(),
                 items: _.values(items).map(function(item) {
                     var itemCopy = $.extend(true, {}, item);
                     itemCopy.font_size = unzoomFont(item.font_size);
                     return item;
-                }),
-                background: {
-                    position: template.data.background.position,
-                    image_id: template.data.background.image_id.toString()
-                }
+                })
             },
+            clear_background: !template.background_url,
             title: $('.js-template-name').val(),
             backside_template_id: backsideTemplateID
         };
@@ -535,7 +533,7 @@
     function displayBackground(template, isBackside) {
         var templateSideClass = isBackside ? '.template-side.back' : '.template-side.front';
         var $backgroundElement = $(templateSideClass).find('.background-image');
-        var backgroundPos = template.data.background.position;
+        var backgroundPos = template.data.background_position;
 
         $backgroundElement.attr({
             src: template.background_url
@@ -562,7 +560,6 @@
     function removeBackground(template) {
         if (template.background_url) {
             template.background_url = null;
-            template.data.background.image_id = '';
             $('.template-side.front').trigger('indico:backgroundChanged');
         }
     }
@@ -593,7 +590,6 @@
         var $templateSide = $('.template-side.back');
         if (template.background_url) {
             template.background_url = null;
-            template.data.background.image_id = '';
             $templateSide.find('.background-image').remove();
             $templateSide.append($('<img>', {class: 'background-image'}));
         }
@@ -664,7 +660,6 @@
                         return;
                     }
                     template.background_url = data.image_url;
-                    template.data.background.image_id = data.image_id.toString();
                     displayBackground(template);
                     $('.template-side.front').trigger('indico:backgroundChanged');
                 }
@@ -676,7 +671,7 @@
                 var $backgroundElement = $('.template-side.front .background-image');
 
                 setBackgroundPos($backgroundElement, newPosition);
-                template.data.background.position = newPosition;
+                template.data.background_position = newPosition;
             });
 
             $('.js-remove-bg').click(function(e) {
@@ -769,8 +764,8 @@
                 }
                 $('.js-upload-bg').attr('disabled', !$backgroundFile.val());
                 $('.js-remove-bg').attr('disabled', !template.background_url);
-                $('#bg-position-stretch').prop('checked', template.data.background.position === 'stretch');
-                $('#bg-position-center').prop('checked', template.data.background.position === 'center');
+                $('#bg-position-stretch').prop('checked', template.data.background_position === 'stretch');
+                $('#bg-position-center').prop('checked', template.data.background_position === 'center');
             });
 
             $('.template-side.front').trigger('indico:backgroundChanged');
@@ -812,6 +807,8 @@
                     title: $this.data('title'),
                 });
             });
+
+            $('.backside-tools').addClass('hidden');
         });
 
         // We load the template if we are editing a template
--- a/indico/htdocs/js/indico/modules/designer.js
+++ b/indico/htdocs/js/indico/modules/designer.js
@@ -321,16 +321,37 @@
         });
     }
 
+    function updatePreset(templateWidth, templateHeight) {
+        $('.js-preset-tool option').each(function() {
+            var $this = $(this);
+
+            if ($this.val() === 'custom') {
+                return;
+            }
+
+            var width = $this.data('width');
+            var height = $this.data('height');
+
+            if (width === templateWidth && height === templateHeight) {
+                $this.prop('selected', true);
+                $('.js-template-dimension').prop('disabled', true);
+            }
+        });
+    }
+
     function changeTemplateSize(template, backsideTemplate) {
         var tpl = $('.template-container');
-        tpl.width($('.template-width').val() * pixelsPerCm);
-        tpl.height($('.template-height').val() * pixelsPerCm);
+        var templateWidth = parseFloat($('.template-width').val());
+        var templateHeight = parseFloat($('.template-height').val());
+
+        tpl.width(templateWidth * pixelsPerCm);
+        tpl.height(templateHeight * pixelsPerCm);
         previousTemplateDimensions.width = templateDimensions.width;
         previousTemplateDimensions.height = templateDimensions.height;
-        templateDimensions = new Dimensions($('.template-width').val() * DEFAULT_PIXEL_CM,
-                                            $('.template-height').val() * DEFAULT_PIXEL_CM);
+        templateDimensions = new Dimensions(templateWidth * DEFAULT_PIXEL_CM, templateHeight * DEFAULT_PIXEL_CM);
         updateRulers();
         displayAllBackgrounds(template, backsideTemplate);
+        updatePreset(templateWidth, templateHeight);
     }
 
     function moveSelectedItem(direction) {
@@ -607,8 +628,6 @@
         // Item class
         $(document).ready(function() {
             var removeBackgroundQtip = $('.js-remove-bg').qtip();
-            var templateWidth = $('.template-width').val();
-            var templateHeight = $('.template-height').val();
 
             $('#bg-form input[type="file"]').on('change', function() {
                 var $this = $(this);
@@ -756,22 +775,6 @@
 
             $('.template-side.front').trigger('indico:backgroundChanged');
 
-            $('.js-preset-tool option').each(function() {
-                var $this = $(this);
-
-                if ($this.val() === 'custom') {
-                    return;
-                }
-
-                var width = $this.data('width').toString();
-                var height = $this.data('height').toString();
-
-                if (width === templateWidth && height === templateHeight) {
-                    $this.prop('selected', true);
-                    $('.js-template-dimansion').attr('disabled', true);
-                }
-            });
-
             $('.template-side.back').on('indico:backsideUpdated', function(evt, data) {
                 backsideTemplateID = data.backside_template_id;
                 clearBacksideTemplate(backsideTemplate);
@@ -784,6 +787,16 @@
                 backsideTemplateID = null;
                 clearBacksideTemplate(backsideTemplate);
                 toggleBacksidePlaceholder(true);
+            });
+
+            $('.js-change-orientation').on('click', function() {
+                var $width = $('.template-width');
+                var $height = $('.template-height');
+                var widthValue = $width.val();
+
+                $width.val($height.val());
+                $height.val(widthValue);
+                changeTemplateSize(template, backsideTemplate);
             });
         });
 
--- a/indico/htdocs/js/indico/modules/designer.js
+++ b/indico/htdocs/js/indico/modules/designer.js
@@ -29,6 +29,7 @@
     var editing = false;
     var items = {};
     var backsideItems = {};
+    var backsideTemplateID = null;
     var itemTitles = {};
 
     var DEFAULT_PIXEL_CM = 50;
@@ -448,6 +449,7 @@
                 }
             },
             title: $('.js-template-name').val(),
+            backside_template_id: backsideTemplateID
         };
 
         $.ajax({
@@ -767,6 +769,7 @@
             });
 
             $('.template-side.back').on('indico:backsideUpdated', function(evt, data) {
+                backsideTemplateID = data.backside_template_id;
                 clearBacksideTemplate(backsideTemplate);
                 displayTemplate(data.template, true);
                 backsideTemplate = data.template;
--- a/indico/htdocs/js/indico/modules/designer.js
+++ b/indico/htdocs/js/indico/modules/designer.js
@@ -133,11 +133,11 @@
             left: item.x + 'px',
             top: item.y + 'px',
             zIndex: itemIdCounter + 10
-        }).data('id', item.id).appendTo('#template-div');
+        }).data('id', item.id).appendTo('.template-side.active');
 
         newDiv.draggable({
-            containment: '#template-div',
-            stack: '#template-div > div',
+            containment: '.template-side.active',
+            stack: '.template-side.active > div',
             opacity: 0.5,
             drag: function(e, ui) {
                 if (snapToGrid) {
@@ -174,7 +174,7 @@
         if (lastSelectedDiv) {
             delete items[lastSelectedDiv.data('id')];
             lastSelectedDiv.remove();
-            $('#selection-text').html('');
+            $('.selection-text').html('');
             lastSelectedDiv = null;
             $('#modify-panel').hide();
             $('#tab-format').hide();
@@ -186,7 +186,7 @@
         // Change the text that says which item is selected
         var id = selection.parent('.ui-draggable').data('id');
         var item = items[id];
-        $('#selection-text').html(item.type === 'fixed' ? $T.gettext('Fixed text') : itemTitles[item.type]);
+        $('.selection-text').html(item.type === 'fixed' ? $T.gettext('Fixed text') : itemTitles[item.type]);
 
         // Bring highlight to the element modification tab.
         if (controlTabs) {
@@ -304,7 +304,7 @@
 
 
     function changeTemplateSize(template) {
-        var tpl = $('#template-div');
+        var tpl = $('.template-container');
         tpl.width($('#badge_width').val() * pixelsPerCm);
         tpl.height($('#badge_height').val() * pixelsPerCm);
         previousTemplateDimensions.width = templateDimensions.width;
@@ -513,7 +513,7 @@
             $('#loadingIcon').hide();
             $('#remove-background').removeClass('hidden');
             setBackgroundPos($(this), template.data.background_position);
-        }).appendTo('#template-div');
+        }).appendTo('.template-side.active');
     }
 
     function removeBackground(template) {
@@ -547,15 +547,11 @@
         $(document).ready(function() {
             $('#bg-form input[type="file"]').on('change', function() {
                 var $this = $(this);
-                if ($this.val()) {
-                    $('#upload-background').removeClass('hidden');
-                } else {
-                    $('#upload-background').addClass('hidden');
-                }
+                $('#upload-background').toggleClass('disabled', !$this.val());
             });
 
-            if (template.backgroundId) {
-                $('#remove-background').removeClass('hidden');
+            if (template.background) {
+                $('#remove-background').removeClass('disabled');
             }
 
             // select and inline edit
@@ -645,6 +641,16 @@
             });
 
             controlTabs = $('#designer-control-tabs').tabs();
+
+            $('.js-toggle-side').on('click', function() {
+                var $this = $(this);
+                var flipSide = $this.hasClass('front') ? 'back' : 'front';
+                $('#template-div').toggleClass('flipped', flipSide === 'front');
+                $('.template-side.back').toggleClass('active', flipSide === 'front');
+                $('.template-side.front').toggleClass('active', flipSide === 'back');
+                $('.js-toggle-side').removeClass('highlight');
+                $this.toggleClass('highlight');
+            });
         });
 
         // We load the template if we are editing a template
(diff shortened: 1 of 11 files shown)


indico/modules/designer/templates/backside_list.html
```diff
@@ -0,0 +1,82 @@
...
+            </div>
+        {% endif %}
+    </section>
+    <section class="custom-template-list">
+        <h3>{% trans %}Custom templates{% endtrans %}</h3>
+        {% if custom_templates %}
+            <div class="i-box just-group-list">
+                <div class="i-box-content">
+                    <ul class="group-list no-content-before badge-backside-list">
+                        {% for tpl in custom_templates | sort(attribute='title') if tpl.data.width == template_width and tpl.data.height == template_height %}
```
ThiefMaster: Is hiding templates with the wrong dimensions really a good idea? IIRC we wanted to show a warning in that case and then update the dimensions of your current template.

indico/modules/designer/templates/backside_list.html
```diff
@@ -0,0 +1,82 @@
...
+{% block javascript %}
+    <script>
+        $('.backside-template').on('click', function() {
+            var $this = $(this);
+            $.ajax({
+                url: $this.data('href'),
+                type: 'POST',
+                dataType: 'json',
+                success: function(data) {
+                    if (handleAjaxError(data)) {
```
ThiefMaster: With new code you shouldn't ever end up in `success` in case of an error.
Instead, use `error: handleAjaxError` right above to catch real errors (ie non-2xx http status codes).

indico/modules/designer/templates/backside_list.html
```diff
@@ -0,0 +1,82 @@
...
+    </section>
+{% endblock %}
+
+{% block javascript %}
+    <script>
+        $('.backside-template').on('click', function() {
+            var $this = $(this);
+            $.ajax({
+                url: $this.data('href'),
+                type: 'POST',
```
ThiefMaster: why POST?

indico/modules/designer/controllers.py
```diff
@@ -268,14 +273,33 @@ def _process(self):
...
+
+
+class RHGetTemplateData(RHModifyDesignerTemplateBase):
+    def _process(self):
+        template_data = {
+            'title': self.template.title,
+            'data': self.template.data,
+            'background_url': self.template.background_image.download_url if self.template.background_image else None
+        }
+        return jsonify_data(template=template_data, backside_template_id=self.template.id)
```
ThiefMaster: `flash=False` or just use `jsonify()` since you don't really care about the `success: true` in the response

indico/modules/designer/blueprint.py
```diff
@@ -55,6 +58,8 @@ def view_func(**kwargs):
     _bp.add_url_rule(prefix + '/<int:template_id>/clone', 'clone_template',
                      _dispatch(RHCloneEventTemplate, RHCloneCategoryTemplate),
                      defaults={'object_type': object_type}, methods=('POST',))
+    _bp.add_url_rule(prefix + '/<int:template_id>/data', 'get_template_data',
+                     RHGetTemplateData, defaults={'object_type': object_type}, methods=('GET', 'POST'))
```
ThiefMaster: no need to POST here

indico/modules/designer/templates/backside_list.html
```diff
@@ -0,0 +1,82 @@
+{% extends 'layout/dialog_base.html' %}
+
+{% block description %}
+    {%- trans -%}
+        Listed below you will find the badge templates that can be used as the back side for your template. Note that only single-sided templates with the same dimentions can be used as the back side of a template.
```
ThiefMaster: linebreaks are free :p

indico/modules/designer/templates/template.html
```diff
@@ -1,268 +1,312 @@
-{% extends 'categories/management/base.html' if category else 'events/management/base.html' %}
+{% extends 'categories/management/base.html' if category else 'events/management/full_width_base.html' %}
```
ThiefMaster: We want full-width in events but not in categories?
Otherwise consider adding a similar `full_width_base.html` template for categories too

indico/modules/designer/templates/template.html
```diff
@@ -1,268 +1,312 @@
...
+                    <i class="icon-font-size" title="{% trans %}Font dimentions{% endtrans %}"></i>
+                    <div class="subtool">
+                        <select id='font-selector' class="js-font-tool" data-attr="font">
+                            <optgroup label="{% trans %}Normal Fonts{% endtrans %}">
+                                <option value="serif">Serif</option>
+                                <option value="courier">Courier</option>
+                                <option value="sans-serif">Sans Serif</option>
+                            </optgroup>
+                            <optgroup label="{% trans %}Special Character Fonts{% endtrans %}">
+                                <option>LinuxLibertine</option>
```
ThiefMaster: no values?

indico/modules/designer/templates/backside_list.html
nop33: I discussed this with @pferreir on Friday and that's what we concluded.. @pferreir what do you think?

nop33: We could gray them out instead, and use a tooltip.

indico/modules/designer/templates/template.html
nop33: I only moved the code. The JS takes care of the values.

indico/migrations/versions/201705221530_3ca338ed5192_remove_background_image_and_add.py
```diff
@@ -0,0 +1,37 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '3ca338ed5192'
+down_revision = '35d76c40ca48'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.drop_constraint(u'fk_designer_templates_background_image_id_designer_image_files', 'designer_templates',
```
ThiefMaster: no need for the `u` prefixes here

indico/migrations/versions/201705221530_3ca338ed5192_remove_background_image_and_add.py
```diff
@@ -0,0 +1,37 @@
...
+                       schema='indico', type_='foreignkey')
```
ThiefMaster: the type isn't needed

indico/migrations/versions/201705221530_3ca338ed5192_remove_background_image_and_add.py
```diff
@@ -0,0 +1,37 @@
...
+    op.drop_column('designer_templates', 'background_image_id', schema='indico')
+    op.add_column('designer_templates', sa.Column('backside_template_id', sa.Integer(), nullable=True), schema='indico')
+    op.create_foreign_key(op.f('fk_designer_templates_backside_template_id_designer_templates'), 'designer_templates',
```
ThiefMaster: `None` as the first arg so the name is autogenerated

indico/migrations/versions/201705221530_3ca338ed5192_remove_background_image_and_add.py
```diff
@@ -0,0 +1,37 @@
...
+                          'designer_templates', ['backside_template_id'], ['id'], source_schema='indico',
+                          referent_schema='indico')
+
+
+def downgrade():
+    op.drop_constraint(op.f('fk_designer_templates_backside_template_id_designer_templates'), 'designer_templates',
```
ThiefMaster: `op.f()` isn't needed here,

indico/migrations/versions/201705221530_3ca338ed5192_remove_background_image_and_add.py
```diff
@@ -0,0 +1,37 @@
...
+                       schema='indico', type_='foreignkey')
```
ThiefMaster: type can be omitted

indico/migrations/versions/201705221530_3ca338ed5192_remove_background_image_and_add.py
```diff
@@ -0,0 +1,37 @@
...
+    op.drop_column('designer_templates', 'backside_template_id', schema='indico')
+    op.add_column('designer_templates',
+                  sa.Column('background_image_id', sa.INTEGER(), autoincrement=False, nullable=True), schema='indico')
```
ThiefMaster: `sa.Integer`
`autoincrement=False` is the default and can be omitted

indico/modules/designer/controllers.py
```diff
@@ -268,14 +273,34 @@ def _process(self):
...
 
 
 class RHDeleteDesignerTemplate(RHModifyDesignerTemplateBase):
     def _process(self):
         db.session.delete(self.template)
         flash(_('The template has been removed'), 'success')
         return jsonify_data(html=_render_template_list(self.target, event=self.event_or_none))
+
+
+class RHListBackSideTemplates(RHModifyDesignerTemplateBase, RHListEventTemplates):
```
ThiefMaster: `RHListBacksideTemplates` - "Backside" is one word after all

indico/modules/designer/models/templates.py
```diff
@@ -105,7 +105,10 @@ def __init__(self, **kwargs):
         if data is None:
             data = {
                 'items': [],
-                'background_position': 'stretch'
+                'background': {
+                    'image_id': '',
```
ThiefMaster: do we REALLY want to go for an empty string here?

indico/modules/designer/models/templates.py
```diff
@@ -115,6 +118,11 @@ def __init__(self, **kwargs):
     def owner(self):
         return self.event_new if self.event_new else self.category
 
+    @property
+    def background_image(self):
+        image_id = self.data['background']['image_id']
+        return DesignerImageFile.get(int(image_id)) if image_id else None
```
ThiefMaster: Why are we storing the image id as a string? Actually, why are we storing the `image_id` in the JSON at all and not in a column as before?

indico/modules/designer/models/templates.py
```diff
@@ -64,9 +65,8 @@ class DesignerTemplate(db.Model):
         JSON,
         nullable=False
     )
-    background_image_id = db.Column(
```
ThiefMaster: Why remove this?

indico/modules/designer/models/templates.py
pferreir: Yeah, it used to be a column, and I thought it would stay as a column.

indico/modules/designer/models/templates.py
```diff
@@ -89,11 +89,11 @@ class DesignerTemplate(db.Model):
...
-    background_image = db.relationship(
-        'DesignerImageFile',
+    backside_template = db.relationship(
+        'DesignerTemplate',
         lazy=True,
-        foreign_keys=background_image_id,
-        post_update=True
+        foreign_keys=backside_template_id,
+        remote_side=id
     )
```
pferreir: Why do we need to remove `background_image` here? Didn't we agree things would stay the same, we'd only add a reference to a backside?

pferreir: OK, @ThiefMaster already mentions this above...

indico/migrations/versions/201705221530_3ca338ed5192_remove_background_image_and_add.py
```diff
@@ -0,0 +1,37 @@
...
+    op.create_foreign_key(u'fk_designer_templates_background_image_id_designer_image_files', 'designer_templates',
```
pferreir: I don't think you have to explicitly name the foreign key here either...

indico/modules/designer/models/templates.py
nop33: Well this started when I created the new JSON structure where you could configure the background for both sides. Instead of adding a second column for the backside background, I updated the JSON structure to support the same structure for both back and front side. Afterwards we changed out mind and added the the backside_template column.
We can still load the templates that use a specific background image (`DesignerImageFile.template`), so I don't see a harm doing it this way.
@pferreir can you confirm that this is what we discussed?

indico/migrations/versions/201705221530_3ca338ed5192_remove_background_image_and_add.py
ThiefMaster: yep, None works fine here

indico/modules/designer/models/templates.py
ThiefMaster: But even with two-sided templates storing it in a column would be much cleaner - like this we can have a proper sqlalchemy relationship and foreign keys while in JSON it's just a number that's meaningless to postgres (and thus has no enforced consistency)

pferreir: I wouldn't even go as far as discussing the pros of using a proper relationship. The previous solution was perfectly fine, and the changes we agreed don't change anything about it.

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -104,13 +104,20 @@
...
 
-        items[item.id] = item;
         return item;
     }
 
-    function createItemFromObject(obj) {
+    function createItemFromObject(obj, isBackside) {
         delete obj.id;
-        return _.extend(createItem(), obj);
+        var newItem = createItem();
```
ThiefMaster: just a matter of personal preference but IMO this is clearer instead of assigning a new item, just to replace it with another new item the line after

    var newItem = _.extend(createItem(), obj);

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -288,148 +307,169 @@
...
+
+    function updatePreset(templateWidth, templateHeight) {
+        $('.js-preset-tool option').each(function() {
+            var $this = $(this);
+
+            if ($this.val() === 'custom') {
+                return;
+            }
+
+            var width = $this.data('width').toString();
```
ThiefMaster: why `toString`? I'd just convert the numbers taken from the fields to ints in `changeTemplateSize`

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -288,148 +307,169 @@
...
+            var height = $this.data('height').toString();
+
+            if (width === templateWidth && height === templateHeight) {
+                $this.prop('selected', true);
+                $('.js-template-dimansion').attr('disabled', true);
```
ThiefMaster: s/attr/prop/
s/dimansion/dimension/

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -288,148 +307,169 @@
...
             }
         });
     }
 
+    function changeTemplateSize(template, backsideTemplate) {
+        var tpl = $('.template-container');
+        var templateWidth = $('.template-width').val();
```
ThiefMaster: var templateWidth = +$('.template-width').val();

(see my comment about toString above)

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -530,8 +571,46 @@
...
+    }
+
+    function clearBacksideTemplate(template) {
+        var $templateSide = $('.template-side.back');
+        if (template.background_url) {
+            template.background_url = null;
+            $templateSide.find('.background-image').remove();
+            $templateSide.append($('<img>', {class: 'background-image'}));
+        }
+        $templateSide.find('.designer-item').each(function() {
```
ThiefMaster: $templateSide.find('.designer-item').remove();

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -583,94 +660,178 @@
...
+                changeTemplateSize(template, backsideTemplate);
+            });
+
+            $('.js-backside-list-dialog').on('click', function() {
+                var $this = $(this);
+
+                ajaxDialog({
+                    trigger: $this,
+                    url: $this.data('href'),
+                    method: 'GET',
```
ThiefMaster: that's the default

indico/migrations/versions/201705221530_3ca338ed5192_remove_background_image_and_add.py
```diff
@@ -0,0 +1,29 @@
...
+# revision identifiers, used by Alembic.
+revision = '3ca338ed5192'
+down_revision = '35d76c40ca48'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('designer_templates', sa.Column('backside_template_id', sa.Integer(), nullable=True), schema='indico')
+    op.create_foreign_key('fk_designer_templates_backside_template_id_designer_templates', 'designer_templates',
```
ThiefMaster: `None` for the first argument

indico/modules/designer/blueprint.py
```diff
@@ -55,6 +58,8 @@ def view_func(**kwargs):
     _bp.add_url_rule(prefix + '/<int:template_id>/clone', 'clone_template',
                      _dispatch(RHCloneEventTemplate, RHCloneCategoryTemplate),
                      defaults={'object_type': object_type}, methods=('POST',))
+    _bp.add_url_rule(prefix + '/<int:template_id>/data', 'get_template_data',
+                     RHGetTemplateData, defaults={'object_type': object_type}, methods=('GET',))
```
ThiefMaster: `, methods=('GET',)` is not needed; GET-only is the default

indico/modules/designer/controllers.py
```diff
@@ -279,3 +281,24 @@ def _process(self):
...
+
+class RHListBacksideTemplates(RHModifyDesignerTemplateBase, RHListEventTemplates):
+    def _process(self):
+        inherited_templates = [tpl for tpl in get_inherited_templates(self.target)
+                               if not tpl.backside_template and tpl.type == TemplateType.badge]
+        custom_templates = [tpl for tpl in self.target.designer_templates
+                            if not tpl.backside_template and tpl != self.template]
+        return jsonify_template('designer/backside_list.html', target=self.target, custom_templates=custom_templates,
+                                inherited_templates=inherited_templates, current_template=self.template,
+                                width=int(request.values['width']), height=int(request.values['height']))
```
ThiefMaster: `request.args` since it's always query string data and never POST data

indico/modules/designer/models/templates.py
```diff
@@ -95,6 +100,12 @@ class DesignerTemplate(db.Model):
         foreign_keys=background_image_id,
         post_update=True
     )
+    backside_template = db.relationship(
+        'DesignerTemplate',
+        lazy=True,
+        foreign_keys=backside_template_id,
```
ThiefMaster: This may not be necessary since you already set `remote_side` and there is only one FK linking the table to itself

indico/modules/designer/models/templates.py
```diff
@@ -69,8 +70,12 @@ class DesignerTemplate(db.Model):
         db.ForeignKey('indico.designer_image_files.id'),
         index=False,
         nullable=True
+      )
+    backside_template_id = db.Column(
+        db.ForeignKey('indico.designer_templates.id'),
+        index=False,
```
ThiefMaster: I think it'd be a good idea to have an index on it so we can efficiently check whether a template is the backside of something.

indico/modules/designer/templates/backside_list.html
```diff
@@ -0,0 +1,64 @@
...
+    {% else %}
+        <div class="italic text-not-important">
+            {% trans %}No templates{% endtrans %}
+        </div>
+    {% endif %}
+{% endmacro %}
+
+{% block description %}
+    {%- trans -%}
+        Listed below you will find the badge templates that can be used as the back side for your template. Note that
```
ThiefMaster: "Listed below you fill find" sounds a bit strange to me. How about something like "The following badge templates can be used ..."?

indico/modules/designer/templates/backside_list.html
```diff
@@ -0,0 +1,64 @@
+{% extends 'layout/dialog_base.html' %}
+
+{% macro _render_template_list(templates) %}
+    {% if templates %}
+        <div class="i-box just-group-list">
+            <div class="i-box-content">
+                <ul class="group-list no-content-before badge-backside-list">
+                    {% for tpl in templates | sort(attribute='title') %}
+                        {% set different_size = tpl.data.width|int != width or tpl.data.height|int != height %}
```
ThiefMaster: having proper types in the JSON would be better than the `|int` here

indico/modules/designer/templates/template.html
```diff
@@ -1,268 +1,315 @@
...
+                                    <div class="back-side-placeholder">
+                                        <div class="placeholder-content">
+                                            <div>{% trans %}The back side is empty.{% endtrans %}</div>
+                                            <div>
+                                                {% set link_start %}
+                                                    <a href="#" class="js-backside-list-dialog"
+                                                       data-href="{{ url_for('.backside_template_list', template) }}"
+                                                       data-title="{% trans %}Select a template for the back side{% endtrans %}">
+                                                {% endset %}
+                                                {%- trans link_start=link_start|safe, link_end='</a>'|safe -%}
```
ThiefMaster: While technically the same I would use `{% set link_end %}</a>{% endset %}` above. Otherwise editors like to screw up closing tag detection and highlighting since they see the link start tag but not take the quoted end tag into account.

Editor quirks aside, this would also be more consistent since the start/end tag are set in the same way

indico/modules/designer/templates/template.html
```diff
@@ -271,9 +318,14 @@
             var template = {
                 title: {{ template.title | tojson }},
                 data: {{ template.data | tojson }},
-                background: {{ (template.background_image.download_url if template.background_image else none) | tojson }}
+                background_url: {{ (template.background_image.download_url if template.background_image else none) | tojson }}
+            };
+            var backsideTemplate = {
+                title: {{ (template.backside_template.title if template.backside_template else none) | tojson }},
+                data: {{ (template.backside_template.data if template.backside_template else none) | tojson }},
+                background_url: {{ (template.backside_template.background_image.download_url if template.backside_template and template.backside_template.background_image else none) | tojson }}
```
ThiefMaster: this line is awfully long.. might be nicer to build the whole `backside_template` dict in python and then just jsonify it here:

    var backsideTemplate = {{ backside_template | tojson }};

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -583,94 +660,178 @@
...
+
+            $('.js-preset-tool').on('change', function() {
+                var $selectedOption = $(this).find('option:selected');
+
+                if ($selectedOption.val() !== 'custom') {
+                    $('.template-width').val($selectedOption.data('width'));
+                    $('.template-height').val($selectedOption.data('height'));
+                    changeTemplateSize(template, backsideTemplate);
+                }
+                $('.js-template-dimansion').attr('disabled', $selectedOption.val() !== 'custom');
```
ThiefMaster: dimenison, prop

indico/htdocs/js/indico/modules/designer.js
nop33: I need `parseFloat`

nop33: ah nvm, it works for floats too.

nop33: However, I'd rather use `parseFloat`. I don't rly like the `+`, it's a bit of a weird way to convert a string to a number.

ThiefMaster: your choice ;)

indico/modules/designer/templates/template.html
nop33: Extra lines of code is my argument, but whatever, not a strong opinion

indico/modules/designer/templates/template.html
nop33: In that case I'll do the `var template` as well

PR #2933, opened 2017-05-09
Store system user flag ("janitor") in the DB and create it on db-prepare

ThiefMaster:


indico/cli/admin.py
```diff
@@ -70,7 +70,7 @@ def user_search(substring, include_deleted, include_pending, include_external, *
     assert set(criteria.viewkeys()) == {'first_name', 'last_name', 'email', 'affiliation'}
     criteria = {k: v for k, v in criteria.viewitems() if v is not None}
     res = search_users(exact=(not substring), include_deleted=include_deleted, include_pending=include_pending,
-                       external=include_external, **criteria)
+                       external=include_external, allow_system_user=True, **criteria)
```
pferreir: Should we actually allow this by default?

pferreir: LGTM otherwise.

ThiefMaster: I wouldn't since there's no good reason to add them e.g. to an event - also, users with no emails don't look super pretty in some places.

The only place where I could see some value is being able to add them as an admin (if you create an API token for them) but you can always do that on the CLI if necessary

pferreir: Yep. So, I would probably default it to `False` then.

ThiefMaster: Which [is already the case](Already the case: https://github.com/indico/indico/pull/2933/files#diff-6a85a90c99fa83a08eb0eed05308f410R174) - this line here is just from the user search on the CLI

pferreir: Ah right, sorry, hadn't realized that ;)

PR #2934, opened 2017-05-09
Support timezone info in XLSX export

nop33:
SNow ticket: `RQF0735420`

### DIFF
--- a/indico/modules/events/surveys/controllers/management/results.py
+++ b/indico/modules/events/surveys/controllers/management/results.py
@@ -51,8 +51,6 @@ class RHExportSubmissionsBase(RHManageSurveyBase):
     """Export submissions from the survey"""
 
     CSRF_ENABLED = False
-    EXT = None
-    FUNC = None
 
     def _process(self):
         if not self.survey.submissions:
@@ -61,22 +59,25 @@ class RHExportSubmissionsBase(RHManageSurveyBase):
 
         submission_ids = set(map(int, request.form.getlist('submission_ids')))
         headers, rows = generate_spreadsheet_from_survey(self.survey, submission_ids)
-        filename = 'submissions-{}.{}'.format(self.survey.id, self.EXT)
-        return self.FUNC(filename, headers, rows)
+        filename = 'submissions-{}'.format(self.survey.id)
+        return self._export(filename, headers, rows)
+
+    def _export(self, filename, headers, rows):
+        raise NotImplementedError
 
 
 class RHExportSubmissionsCSV(RHExportSubmissionsBase):
     """Export submissions as CSV"""
 
-    EXT = 'csv'
-    FUNC = staticmethod(send_csv)
+    def _export(self, filename, headers, rows):
+        return send_csv(filename + '.csv', headers, rows)
 
 
 class RHExportSubmissionsExcel(RHExportSubmissionsBase):
     """Export submissions as XLSX"""
 
-    EXT = 'xlsx'
-    FUNC = staticmethod(send_xlsx)
+    def _export(self, filename, headers, rows):
+        return send_xlsx(filename + '.xlsx', headers, rows, tz=self.event_new.tzinfo)
 
 
 class RHSurveySubmissionBase(RHManageSurveysBase):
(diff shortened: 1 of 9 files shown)


indico/util/spreadsheets.py
```diff
@@ -86,17 +88,19 @@ def generate_csv(headers, rows):
...
-def _prepare_excel_data(data):
+def _prepare_excel_data(data, tz=None):
     if isinstance(data, (list, tuple)):
         data = ', '.join(data)
     elif isinstance(data, set):
         data = ', '.join(sorted(data, key=unicode.lower))
     elif is_lazy_string(data) or isinstance(data, Markup):
         data = unicode(data)
+    elif isinstance(data, datetime.datetime) and tz:
+        data = format_datetime(data, timezone=tz)
```
ThiefMaster: `.decode('utf-8')` since these functions return UTF-8 but everything else here is unicode.

indico/util/spreadsheets.py
```diff
@@ -133,13 +137,13 @@ def send_csv(filename, headers, rows):
     return send_file(filename, buf, 'text/csv', inline=False)
 
 
-def send_xlsx(filename, headers, rows):
+def send_xlsx(filename, headers, rows, tz=None):
```
ThiefMaster: If you add a new argument to a function with a docstring documenting the params, please add that argument in there too ;)

indico/util/spreadsheets.py
```diff
@@ -86,17 +88,19 @@ def generate_csv(headers, rows):
...
 
-def _prepare_excel_data(data):
+def _prepare_excel_data(data, tz=None):
     if isinstance(data, (list, tuple)):
         data = ', '.join(data)
     elif isinstance(data, set):
         data = ', '.join(sorted(data, key=unicode.lower))
     elif is_lazy_string(data) or isinstance(data, Markup):
         data = unicode(data)
+    elif isinstance(data, datetime.datetime) and tz:
```
ThiefMaster: What happens if you have a datetime but no tz?

ThiefMaster: I'd use `from datetime import datetime` and then just use `datetime` in the check here.

indico/util/spreadsheets.py
```diff
@@ -86,17 +88,20 @@ def generate_csv(headers, rows):
...
-def _prepare_excel_data(data):
+def _prepare_excel_data(data, tz=None):
     if isinstance(data, (list, tuple)):
         data = ', '.join(data)
     elif isinstance(data, set):
         data = ', '.join(sorted(data, key=unicode.lower))
     elif is_lazy_string(data) or isinstance(data, Markup):
         data = unicode(data)
+    elif isinstance(data, datetime):
+        data = format_datetime(data, timezone=tz) if tz else format_datetime(data)
```
ThiefMaster: No need for the if..else, `timezone=None` is the default so you can just go for `data = format_datetime(data, timezone=tz).decode('utf-8')` without any extra checks

indico/modules/events/surveys/controllers/management/results.py
```diff
@@ -62,7 +62,7 @@ def _process(self):
         submission_ids = set(map(int, request.form.getlist('submission_ids')))
         headers, rows = generate_spreadsheet_from_survey(self.survey, submission_ids)
         filename = 'submissions-{}.{}'.format(self.survey.id, self.EXT)
-        return self.FUNC(filename, headers, rows)
+        return self.FUNC(filename, headers, rows, tz=self.event_new.tzinfo)
```
ThiefMaster: ⚠️ This is also called for csv! Maybe change it like this (and get rid of FUNC):

```python
def _export(self, filename, headers, rows):
    return send_csv(filename, headers, rows)

...

def _export(self, filename, headers, rows):
    return send_xlsx(filename, headers, rows, tz=self.event_new.tzinfo)
```

ThiefMaster: Actually, get rid of EXT too and just use `filename + '.csv'` and `filename + '.xlsx'` in the calls

nop33: but when we generate the data to pass to the send_csv, we are not passing datetime objects, so even if we pass the timezone, nothing changes.

ThiefMaster: `send_csv()` does not accept a `tz` argument...

PR #2935, opened 2017-05-10
New layout in Indico weeks wiew + settings in yaml

DavidAndreev:


indico/modules/events/layout/controllers/layout.py
```diff
@@ -65,12 +69,50 @@ def _process_lecture_meeting(self):
             return redirect(url_for('.index', self.event_new))
         return WPLayoutEdit.render_template('layout_meeting_lecture.html', self._conf, form=form, event=self.event_new)
 
+    def _make_weeks_view_form(self):
```
ThiefMaster: The ability to add custom settings is not specific to the "weeks view" so there should be no reference to it in any of th e names here.

indico/modules/events/layout/controllers/layout.py
```diff
@@ -65,12 +69,50 @@ def _process_lecture_meeting(self):
...
+        form_class = type(b'WeeksViewForm', (IndicoForm,), {})
+        fields = theme_settings.themes['indico-weeks-view']['arguments']
```
ThiefMaster: This should use the currently selected theme of the event, not hardcode a specific one (if we wanted it only for that theme we'd just define the form in Python)

indico/modules/events/layout/controllers/layout.py
```diff
@@ -65,12 +69,50 @@ def _process_lecture_meeting(self):
...
+        defaults = {}
+        for field_name in fields:
+            field_type = fields[field_name]['type']
+            field_class = getattr(indico_fields, field_type, None) or getattr(wtforms_fields, field_type, None)
+            if not field_class:
+                continue
```
ThiefMaster: fail with an exception instead of silently skipping invalid field classes

indico/modules/events/layout/controllers/layout.py
```diff
@@ -65,12 +69,50 @@ def _process_lecture_meeting(self):
...
+            field_object = type(str(field_name), (field_class,), {})
```
ThiefMaster: You don't need to subclass the field class. Just create an instance of `field_class` below...

indico/modules/events/layout/controllers/layout.py
```diff
@@ -65,12 +69,50 @@ def _process_lecture_meeting(self):
...
+            label = fields[field_name]['caption']
+            description = fields[field_name]['description']
+            defaults[field_name] = fields[field_name]['default'] if fields[field_name].get('default') else None
```
ThiefMaster: defaults[field_name] = fields[field_name].get('default')

indico/modules/events/layout/controllers/layout.py
```diff
@@ -65,12 +69,50 @@ def _process_lecture_meeting(self):
             return redirect(url_for('.index', self.event_new))
         return WPLayoutEdit.render_template('layout_meeting_lecture.html', self._conf, form=form, event=self.event_new)
 
+    def _make_weeks_view_form(self):
+        form_class = type(b'WeeksViewForm', (IndicoForm,), {})
+        fields = theme_settings.themes['indico-weeks-view']['arguments']
+        defaults = {}
+        for field_name in fields:
```
ThiefMaster: `for field_name, field_data in fields.iteritems()` - then you don't need to use `fields[field_name]` all the time

indico/modules/events/layout/controllers/layout.py
```diff
@@ -65,12 +69,50 @@ def _process_lecture_meeting(self):
...
+            field_type = fields[field_name]['type']
+            field_class = getattr(indico_fields, field_type, None) or getattr(wtforms_fields, field_type, None)
+            if not field_class:
+                continue
+            field_object = type(str(field_name), (field_class,), {})
+            label = fields[field_name]['caption']
+            description = fields[field_name]['description']
+            defaults[field_name] = fields[field_name]['default'] if fields[field_name].get('default') else None
+            validators = [DataRequired()] if fields[field_name].get('required') else []
+            if field_class == MultipleItemsField:
```
ThiefMaster: Why do we need custom logic here? Ideally this is all defined from the YAML data...

indico/modules/events/layout/templates/layout_conference.html
```diff
@@ -28,6 +28,11 @@
                 </a>
             {% endcall %}
         {% endcall %}
+        {% if theme_settings_form %}
```
ThiefMaster: Maybe this should be in a dialog (and only available once the theme has been saved for the event, since otherwise it's confusing if you change the theme but still see another theme's config form)?

indico/modules/events/timetable/templates/display/_weeks.html
```diff
@@ -44,16 +45,33 @@
                         {{ day | format_date(format='full', timezone=timezone) }}
                     </div>
                     {% set hidden_count = [] %}
+                    {% set short_contrib_count = [] %}
                     {% for start_time, slot_entries in day_entries.viewitems() %}
+                        {% set start_time_minutes = (start_time.strftime('%-H')|int*60 + start_time.strftime('%-M')|int) %}
+                        {% if day_entries.keys()[loop.index] %}
```
ThiefMaster: You use it that often so maybe assign `{% set days = day_entries.keys() %}` and then use that?

indico/modules/events/timetable/templates/display/_weeks.html
```diff
@@ -44,16 +45,33 @@
...
+                            {% set next_start_minutes = (day_entries.keys()[loop.index].strftime('%-H')|int*60 + day_entries.keys()[loop.index].strftime('%-M')|int)%}
+                        {% else %}
+                            {% set next_start_minutes = start_time_minutes %}
+                        {% endif %}
+                        {% set duration = next_start_minutes - start_time_minutes %}
+                        {% if prev_end[0] and not hidden_count[-1] %}
+                            {% set end_time = prev_end.pop().time() %}
+                            {% set new_dur  = start_time_minutes - (end_time.strftime('%-H')|int*60 + end_time.strftime('%-M')|int) %}
+                            {% if end_time < start_time and end_time.strftime('%-H')!= '0' %}
+                                <div class="row placeholder" data-slot="{{ start_time }}" style="{{ 'height: %dpx'|format(((new_dur/5) * px_per_5_minutes) -1) }}">
```
ThiefMaster: `style="height: {{ ((new_dur / 5) * px_per_5_minutes) - 1 }}px;'"`

indico/modules/events/timetable/templates/display/_weeks.html
```diff
@@ -44,16 +45,33 @@
...
+                                </div>
+                            {% endif %}
+                        {% endif %}
                         {% if not slot_entries %}
                             {% if hidden_count[-1] %}
                                 {% set __ = hidden_count.append(hidden_count[-1] - 1) %}
                             {% else %}
-                                <div class="row placeholder" data-slot="{{ start_time }}">
+                                <div class="row placeholder" data-slot="{{ start_time }}" style="{{ 'height: %dpx'|format(((duration/5) * px_per_5_minutes) -1) }}">
```
ThiefMaster: see above

indico/modules/events/timetable/templates/display/_weeks.html
```diff
@@ -44,16 +45,33 @@
...
                                 </div>
                             {% endif %}
                         {% else %}
                             {% for entry, count in slot_entries %}
+                            {% set __ = prev_end.append(entry.end_dt.astimezone(tz_object)) %}
```
ThiefMaster: wrong indentation

indico/modules/events/timetable/templates/display/_weeks.html
```diff
@@ -63,11 +81,32 @@
 
                                 {% set has_multi = false %}
                                 {% if loop.first %}
-                                    {% set extra_styles = ['height: %dpx'|format((18 + hidden_count[-1] * 19) if hidden_count[-1] else 18)] %}
+                                    {% set time_parts = (entry.duration.seconds/60) /5 %}
+                                    {% set height = (time_parts * px_per_5_minutes) %}
+                                    {% if short_contrib_count[-1] %}
+                                        {% set height = height - 7 * short_contrib_count| length %}
+                                        {%  set __ = short_contrib_count.pop() %}
```
ThiefMaster: too much space

indico/modules/events/timetable/views/weeks.py
```diff
@@ -105,5 +106,14 @@ def inject_week_timetable(event, days, tz_name):
...
         week_table.append(tmp)
 
-    return render_template('events/timetable/display/_weeks.html', event=event, week_table=week_table,
+    entry_colors = {}
+    if (layout_settings.get(event, 'weeks_view_settings') and
+            'colors' in layout_settings.get(event, 'weeks_view_settings')):
+        entry_colors = layout_settings.get(event, 'weeks_view_settings')['colors']
+    elif (theme_settings.themes['indico-weeks-view'].get('arguments') and
+          theme_settings.themes['indico-weeks-view'].get('arguments').get('colors') and
+          theme_settings.themes['indico-weeks-view'].get('arguments').get('colors').get('default')):
```
ThiefMaster: You could just do this and get rid of the previous two lines...

    theme_settings.themes['indico-weeks-view'].get('arguments', {}).get('colors', {}).get('default')

indico/modules/events/layout/__init__.py
```diff
@@ -45,7 +45,8 @@
     'timetable_theme': None,
     'use_custom_menu': False,
     'timetable_by_room': False,
-    'timetable_detailed': False
+    'timetable_detailed': False,
+    'weeks_view_settings': None
```
ThiefMaster: too specific, should be `theme_settings: {}` and cleared whenever the theme is changed (see my other comments below regarding how this should be much more generic)

indico/modules/events/layout/controllers/layout.py
DavidAndreev: And how can I get the currently selected theme ? In `views/_init_.py` we get the view from request: `view = request.args.get('view')`   and then use get theme like this : `theme = theme_settings.themes[get_theme(event, view)[0]]` but I dont know how to get the current theme in this context

ThiefMaster: Since we only keep settings for the currently *selected* theme (ie saved for the event, not for a theme specified with `?view=...`) you can just read it from the settings proxy: `layout_settings.get(event, 'timetable_theme')`

DavidAndreev: Are you sure ? That one always gives me None. Even when I try to print it in display when showing the timetable

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -195,4 +195,27 @@
             $(this).closest('tr').toggleClass('selected', this.checked);
         }).trigger('change');
     };
+
+    global.showThemeSettingsForm = function showThemeSettingsForm(formUrl, lastFieldId) {
```
ThiefMaster: We usually pass an `options` dict to these functions instead of having separate arguments (see some other global functions for examples)

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -195,4 +195,27 @@
...
+        $('#timetable_theme').on('change', function() {
+            $.ajax({
+                url: formUrl,
+                data: {
+                    theme: $(this).val()
+                },
+                complete: IndicoUI.Dialogs.Util.progress(),
+                error: handleAjaxError,
+                success: function(data) {
+                    var lastField = $(lastFieldId);
```
ThiefMaster: If the argument is called `lastFieldId` it shouldn't contain the `#` but it should rather be added here

indico/modules/events/timetable/views/weeks.py
```diff
@@ -105,5 +106,8 @@ def inject_week_timetable(event, days, tz_name):
             tmp.append((day, day_entries))
         week_table.append(tmp)
 
-    return render_template('events/timetable/display/_weeks.html', event=event, week_table=week_table,
+    entry_colors = {}
+    if layout_settings.get(event, 'timetable_theme') == 'indico-weeks-view':
```
ThiefMaster: I would rather expose all the theme settings to the template (and then get whatever item you need in the jinja template).
Also, the theme settings dict should be available when rendering a "normal" theme. Otherwise it's only usable for the weeks theme.

indico/modules/events/layout/templates/layout_meeting_lecture.html
```diff
@@ -1,10 +1,23 @@
...
-{% from 'forms/_form.html' import simple_form %}
+{% from 'forms/_form.html' import simple_form, form_rows %}
 
 {% block title %}
     {%- trans %}Layout{% endtrans -%}
 {% endblock %}
 
 {%- block content %}
-    {{ simple_form(form, save_reminder=true, back_button=false) }}
+     {% call simple_form(form, save_reminder=true, back_button=false) %}
```
ThiefMaster: too much indentation

indico/modules/events/layout/templates/layout_meeting_lecture.html
```diff
@@ -1,10 +1,23 @@
...
+         <div>
+            {{ form_rows(form, fields=('timetable_theme', 'timetable_by_room', 'timetable_detailed')) }}
```
ThiefMaster: the only field this form has is `timetable_theme`, and since it has only one field just keep the simple `form_rows(form)` call without specifying fields

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -195,4 +195,27 @@
             $(this).closest('tr').toggleClass('selected', this.checked);
         }).trigger('change');
     };
+
+    global.showThemeSettingsForm = function showThemeSettingsForm(options) {
+        $('#timetable_theme').on('change', function() {
```
ThiefMaster: take this ID from `options`.

also, check what we do in other places to specify the options that are available (`$.extend(..)` with some more or less useful defaults)

indico/modules/events/layout/templates/layout_meeting_lecture.html
```diff
@@ -1,10 +1,23 @@
...
+            {% if timetable_theme_settings_form %}
+                {{ form_rows(timetable_theme_settings_form) }}
+            {% endif %}
+         </div>
+    {% endcall %}
+    <script>
+        showThemeSettingsForm({
+            formUrl: {{ url_for('.timetable_theme_form', event) | tojson }},
+            lastFieldId: 'form-group-timetable_theme'
```
ThiefMaster: Just FYI, you could get the ID from the form: `'form-group-{{ form.timetable_theme.id }}'`

indico/modules/events/timetable/views/weeks.py
```diff
@@ -105,5 +106,6 @@ def inject_week_timetable(event, days, tz_name):
             tmp.append((day, day_entries))
         week_table.append(tmp)
 
-    return render_template('events/timetable/display/_weeks.html', event=event, week_table=week_table,
+    timetable_settings = layout_settings.get(event, 'timetable_theme_settings', {})
```
ThiefMaster: the `, {}` is not needed; that's already the default

ThiefMaster:
Right now the settings support is kind of useless for any non-weeks themes. I would make these settings available the same way as hardcoded YAML settings (e.g. `show_notes`) - actually, we could even change this setting to be a switch which would be off by default for "Indico style" and on by default for "Indico style - inline minutes".

indico/htdocs/js/indico/modules/events/management.js
DavidAndreev: That's a different ID. The one in options is an ID of a field after which we add the custom fields depending on which theme is selected.

ThiefMaster: Well, add a `themeFieldId` option ;)

DavidAndreev: but why if its always the same id ?

indico/modules/events/layout/templates/layout_meeting_lecture.html
```diff
@@ -1,10 +1,24 @@
...
+{% from 'forms/_form.html' import simple_form, form_rows %}
 
 {% block title %}
     {%- trans %}Layout{% endtrans -%}
 {% endblock %}
 
 {%- block content %}
-    {{ simple_form(form, save_reminder=true, back_button=false) }}
+    {% call simple_form(form, save_reminder=true, back_button=false) %}
+         <div>
```
ThiefMaster: why the extra div? we don't use it in the other template either

indico/modules/events/views.py
```diff
@@ -111,6 +111,7 @@ def _getBody(self, params):
                              category=self.event.category.title,
                              timezone=self.event.display_tzinfo,
                              theme_settings=self.theme.get('settings', {}),
+                             theme_user_settings=self.theme.get('user_settings', {}),
```
ThiefMaster: That's not what I meant - the settings definitions are quire useless in the template. Instead, get the actual settings dict here (from the layout_settings) - but only if the theme is the event's theme (ie not if it's just selected via `?view=...`)

indico/modules/events/layout/templates/layout_meeting_lecture.html
DavidAndreev: Cause in javascript I'm removing elements after the theme selector when changing the theme, so if its not in a div it will remove the Save button.

ThiefMaster: ah ok probably because it's not in a fieldset.. even though you could see whether there's any css selector you can use in the `nextAll()` call to stop it from removing the save button. maybe `:not(.form-group-footer)`

indico/modules/events/themes.yaml
```diff
@@ -47,3 +47,33 @@ definitions:
...
+        required: false
+        kwargs:
+          fields:
+            - id: entry
+              caption: Entry type
+            - id: color
+              caption: Colour
+          unique_field: entry
+        defaults:
+          - entry: foo
```
ThiefMaster: do we want to keep these defaults?

indico/modules/events/themes.yaml
```diff
@@ -47,3 +47,33 @@ definitions:
...
+    user_settings:
+      colors:
+        caption: Contribution Colours
+        description: 'Colours used for different types of timetable entries'
+        type: MultipleItemsField
+        required: false
+        kwargs:
+          fields:
+            - id: entry
+              caption: Entry type
```
ThiefMaster: Shouldn't we mention keywords here? Otherwise it's not clear what the entry type is!

indico/modules/events/themes.yaml
```diff
@@ -47,3 +47,33 @@ definitions:
...
+        description: 'First day of weeks view'
+        type: SelectField
+        required: true
+        kwargs:
+          choices:
+            - [event, Event start day]
+            - [monday, Monday]
+            - [sunday, Sunday]
+        defaults:
+            sunday
```
ThiefMaster: not monday?

indico/modules/events/timetable/templates/display/_weeks.html
```diff
@@ -1,40 +1,164 @@
...
+            {% if entry.object.contributions %}
+                {{ _render_contrib_list(entry.object) }}
+            {% endif %}
         {% endif %}
     </div>
 {% endmacro %}
 
+{% macro _render_time_slots(start_time, slot_entries, prev_end, hidden_count, short_contrib_count)%}
+    {% for entry, count in slot_entries %}
+        {% set __ = prev_end.append(entry.end_dt.astimezone(tz_object)) %}
```
pferreir: As I said the other day IRL, maybe we could move some of this stuff into macros, so that it's a little bit less confusing?

indico/htdocs/sass/themes/weeks.scss
```diff
@@ -15,6 +15,9 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
+@import "base/palette";
+@import "base/defaults";
+@import "base/borders";
```
nop33: You can also just import `base`.
```scss
@import 'base';
```
Also, we use single quotes in imports.

indico/htdocs/sass/themes/weeks.scss
```diff
@@ -136,7 +137,7 @@
 
 .week-timetable .row .time {
     position: absolute;
-    background: #ddd;
+    background: rgba(0,0,0,0.3);
```
nop33: spaces between values

indico/htdocs/sass/themes/weeks.scss
```diff
@@ -209,3 +201,55 @@
...
+    }
+
+    .description {
+        font-size: 0.8em;
+    }
+
+    .contrib-list {
+        width: 100%;
+        text-align: left;
+        ul {
```
nop33: empty line above

indico/htdocs/sass/themes/weeks.scss
```diff
@@ -209,3 +201,55 @@
...
+            @include border-bottom($dark-gray);
+            width: 100%;
+            display: block;
+            margin-top: 0.5em
+        }
+
+        .tooltip-contrib {
+            display: inline-flex;
+            justify-content: space-between;
+            .time {
```
nop33: empty line above

indico/modules/events/layout/controllers/layout.py
```diff
@@ -36,15 +38,50 @@
 from indico.util.fs import secure_filename
 from indico.util.i18n import _
 from indico.util.string import crc32, to_unicode
+from indico.web.flask.templating import get_template_module
 from indico.web.flask.util import send_file, url_for
-from indico.web.forms.base import FormDefaults
-from indico.web.util import jsonify_data
+from indico.web.forms.base import FormDefaults, IndicoForm
+from indico.web.forms import fields as indico_fields
```
nop33: Swap with line above

indico/modules/events/timetable/templates/display/_weeks.html
```diff
@@ -1,40 +1,170 @@
...
+        {% set speakers = item.person_links|selectattr("is_speaker")|list %}
+        {% if speakers %}
+           <span class="dash">-</span> {{ render_users(speakers, span_class="compact-speakers", separator=' ', title=false) }}
+        {% endif %}
+    </span>
+{% endmacro %}
+
+{% macro _render_contrib_list(session) %}
+    <div class="contrib-list">
+        <strong> Contributions</strong>
```
nop33: missing translation

nop33: and there is an extra space in front

indico/modules/events/timetable/templates/display/_weeks.html
```diff
@@ -1,40 +1,170 @@
...
+            {{ _render_contribution(entry.object) }}
+        {% else %}
+            <span class="main">
+                <span class="title">{{ entry.object.title }}</span>
+            </span>
+        {% endif %}
+        {{ _render_tooltip(entry, timezone) }}
+        {% if has_multi and entry.type.name == 'CONTRIBUTION' %}
+            <i class="icon-expand more-contribs"
+               title="There are {{ slot_entries|length - 1 }} more contributions at this time. Click this icon to show them."
```
nop33: missing translation

indico/modules/events/timetable/templates/display/_weeks.html
```diff
@@ -1,40 +1,170 @@
...
+        {% else %}
+            {% set same_time = True %}
+        {% endif %}
+        {% if entry.type.name == 'CONTRIBUTION' or session %}
+            {% if entry.type.name == 'CONTRIBUTION' %}
+                {% set entry_url = url_for('contributions.display_contribution', entry.object) %}
+            {% else %}
+                {% set entry_url = url_for('sessions.display_session', session) %}
+            {% endif %}
+            <a class="week-anchor {{ 'js-same-time' if same_time }}" href="{{ entry_url }}" style = "{{ 'display:none;' if same_time }}">
```
nop33: You can use the class `weak-hidden` instead of adding `display:none`

indico/modules/events/timetable/templates/display/_weeks.html
```diff
@@ -44,66 +174,33 @@
...
+                        {% set next_entry = day_entries.keys()[loop.index] %}
+                        {% if next_entry %}
+                            {% set next_start_minutes = (next_entry.strftime('%-H')|int*60 + next_entry.strftime('%-M')|int)%}
+                        {% else %}
+                            {% set next_start_minutes = start_time_minutes %}
+                        {% endif %}
+                        {% set duration = next_start_minutes - start_time_minutes %}
+                        {% if prev_end[0] and not hidden_count[-1] %}
+                            {% set end_time = prev_end.pop().time() %}
+                            {% set new_dur  = start_time_minutes - (end_time.strftime('%-H')|int*60 + end_time.strftime('%-M')|int) %}
```
nop33: extra space

indico/htdocs/sass/themes/weeks.scss
ThiefMaster: this is a theme css, do we really want to import everything from base there?

indico/modules/events/timetable/templates/display/_weeks.html
ThiefMaster: also, no space around `=` of attributes

ThiefMaster: didn't `.weak-hidden` not work well with .show()/.hide()/.toggle()/etc?

indico/htdocs/sass/themes/weeks.scss
DavidAndreev: Yeah thats what I wasn't sure about, since we don't need everything.

indico/modules/events/timetable/controllers/display.py
```diff
@@ -47,8 +48,9 @@ def _process(self):
         if self.theme is None:
             event_info = serialize_event_info(self.event_new)
             timetable_data = TimetableSerializer().serialize_timetable(self.event_new, strip_empty_days=True)
+            timetable_settings = layout_settings.get(self.event_new, 'timetable_theme_settings', {})
```
ThiefMaster: do we ever set them to None? if not the `, {}` is not needed since it's the default value anyway which is used if there's no value for the setting

ThiefMaster:
Merged into v1.9.11 (7aab7b7c5)

PR #2937, opened 2017-05-11
Remove delayed temp deletion & add task+cli for temp/cache/assets cleanup instead

ThiefMaster:


indico/cli/cleanup.py
```diff
@@ -0,0 +1,56 @@
...
+import click
+
+from indico.core.config import Config
+from indico.util.fs import cleanup_dir
+
+
+def _print_files(files):
+    if not files:
+        click.echo(click.style('  nothing to delete', fg='green'))
+        return
```
ThiefMaster: (I know it's not necessary but it feels cleaner to me and looks better than putting the loop inside `else`)

ThiefMaster:
Actually this could probably go into v1.9.10 so we can already change the cronjobs to call this script

PR #2942, opened 2017-05-15
Improve event locking

ThiefMaster:
Sorry for yet another MASSIVE PR ;)

- don't disallow viewing stuff
- make everything read-only instead
- reject non-GET requests

### DIFF
--- a/indico/htdocs/js/indico/jquery/actioninput.js
+++ b/indico/htdocs/js/indico/jquery/actioninput.js
@@ -32,7 +32,7 @@
 
             self._super();
 
-            self.actionIcon = $('<a class="i-link accept {0}"></a>'.format(self.options.actionIcon))
+            self.actionIcon = $('<a class="i-link accept hide-if-locked {0}"></a>'.format(self.options.actionIcon))
                 .css("line-height", input.css("height"))
                 .click(function() {
                     self._action();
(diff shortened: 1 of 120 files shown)


indico/htdocs/js/indico/jquery/actioninput.js
```diff
@@ -32,7 +32,7 @@
 
             self._super();
 
-            self.actionIcon = $('<a class="i-link accept {0}"></a>'.format(self.options.actionIcon))
+            self.actionIcon = $('<a class="i-link accept {0} hide-if-locked"></a>'.format(self.options.actionIcon))
```
nop33: I think it was you mentioning that the static classes go before the dynamic ones :wink:

indico/htdocs/sass/modules/_event_management.scss
```diff
@@ -727,3 +727,10 @@ form.horizontal .clone-event-dialog-footer .form-field {
         margin-left: auto;
     }
 }
+
+
+.event-locked {
+    .hide-if-locked {
+        display: none !important;
```
nop33: Isn't this the same style as the `hidden` class?

ThiefMaster: Yes, but the whole point of this is to conditionally hide things if a parent has `.event-locked`.

PR #2943, opened 2017-05-17
Refactor event management WP to use Jinja inheritance

ThiefMaster:
Code using the legacy `WPConferenceModifBase` base keeps using the old behavior of the WP taking care of the wrapping, but any new code that uses `WPEventManagement` is expected to inherit the template from `events/management/base.html`.


docs/source/api/event_management.rst
```diff
@@ -0,0 +1,11 @@
+Event Management
+================
+
+.. currentmodule::  indico.modules.events.management.controllers
+.. autoclass:: RHManageEventBase
```
nop33: Since when do we document controllers?!

indico/legacy/webinterface/pages/conferences.py
```diff
@@ -314,32 +315,24 @@ def _getTitle(self):
...
-        self._navigationTarget = self._conf = conference
+class WPConferenceModifBase(WPEventManagement):
+    """Base class for event management pages without Jinja inheritance.
 
-    def getJSFiles(self):
-        return (main.WPMainBase.getJSFiles(self) +
-                self._includeJSPackage('Management') +
-                self._asset_env['modules_event_cloning_js'].urls() +
-                self._asset_env['modules_event_management_js'].urls())
+    Do not use this for anything new.  Instead, use `WPEventManagement`
```
nop33: Let's make sure we mention this in the weekly meeting (and **emphasize**!) (delete the comment)

docs/source/api/event_management.rst
ThiefMaster: It's the base class for every single management RH. I think such classes (both RH and WP) deserve documentation.

PR #2944, opened 2017-05-17
Implement new RB notification logic

ThiefMaster:
Also simplify it A LOT by not caring about the extremely unlikely case where the celery task is not running for an extended amount of time. In such a case we now simply don't send the missed notifications - they are not important enough to justify overly complex code.

(sorry @OmeGak for removing your notification window magic :D)


OmeGak:
No hard feelings :P

PR #2946, opened 2017-05-18
Add .editorconfig and .isort.cfg

ThiefMaster:
I didn't add the isort config to .editorconfig since it feels very dirty to me that isort settings in there don't have any kind of prefix.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,7 @@
+[isort]
+line_length=120
+multi_line_output=0
+lines_after_imports=2
+add_imports=from __future__ import unicode_literals
+sections=FUTURE,STDLIB,THIRDPARTY,INDICO,FIRSTPARTY,LOCALFOLDER
+known_indico=indico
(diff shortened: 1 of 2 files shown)


.isort.cfg
```diff
@@ -0,0 +1,7 @@
+[isort]
+line_length=120
+multi_line_output=0
+lines_after_imports=2
+add_imports=from __future__ import unicode_literals
+sections=FUTURE,STDLIB,THIRDPARTY,INDICO,FIRSTPARTY,LOCALFOLDER
```
ThiefMaster: the separate INDICO section ensures that `indico_zodbimport` gets its own section. also like this the same file works fine for plugins where indico is kind-of thirdparty but should still have its own section

.isort.cfg
```diff
@@ -0,0 +1,7 @@
+[isort]
+line_length=120
+multi_line_output=0
+lines_after_imports=2
```
ThiefMaster: not 100% sure about this. when the first thing after an import is a function/class we already have two empty lines inbetween, but without this setting it's only one empty line for other things like variables.

PR #2948, opened 2017-05-24
WIP: Badge/Poster designer

nop33:


indico/migrations/versions/201705221530_3ca338ed5192_remove_background_image_and_add.py
```diff
@@ -22,9 +22,12 @@ def upgrade():
                           'designer_templates', 'designer_templates',
                           ['backside_template_id'], ['id'],
                           source_schema='indico', referent_schema='indico')
+    op.create_index('ix_designer_templates_backside_template_id', 'designer_templates', ['backside_template_id'],
```
ThiefMaster: `None` for the name works fine and `unique=False` is the default

indico/migrations/versions/201705241045_bbe2ac15f488_add_is_clonable_column_in_.py
```diff
@@ -0,0 +1,25 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'bbe2ac15f488'
+down_revision = '3ca338ed5192'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('designer_templates', sa.Column('is_clonable', sa.Boolean(), nullable=False, server_default='true'),
```
ThiefMaster: after creating it use `op,alter_column()` to set `server_default=None`

indico/modules/designer/forms.py
```diff
@@ -16,15 +16,25 @@
...
+from indico.web.forms.widgets import SwitchWidget
 
 
 class AddTemplateForm(IndicoForm):
     title = StringField(_('Title'), [DataRequired()])
     type = IndicoEnumSelectField(_('Template'), enum=TemplateType, default=TemplateType.poster)
+    is_clonable = BooleanField(_('Allow cloning'), widget=SwitchWidget(), default=True,
+                               description=_("Allow event managers to clone this template."))
+
+    def __init__(self,  *args, **kwargs):
```
ThiefMaster: too much whitespace

indico/modules/designer/controllers.py
```diff
@@ -177,9 +177,10 @@ def _checkProtection(self):
             raise Forbidden
 
     def _process(self):
-        form = AddTemplateForm()
+        form = AddTemplateForm(target=self.target)
         if form.validate_on_submit():
-            new_template = DesignerTemplate(title=form.title.data, type=form.type.data, **self.target_dict)
+            new_template = DesignerTemplate(title=form.title.data, type=form.type.data,
+                                            is_clonable=form.is_clonable.data, **self.target_dict)
```
ThiefMaster: accessing `form.is_clonable.data` is likely to fail when you deleted the field from the form.
I'd just do something like this after this statement:

    if 'is_clonable' in form and not form.is_clonable.data:
        new_template.is_clonable = False

indico/modules/designer/forms.py
```diff
@@ -16,15 +16,25 @@
...
+        target = kwargs.pop('target')
+        super(AddTemplateForm, self).__init__(*args, **kwargs)
+        if isinstance(target, Event):
+            del self.is_clonable
```
ThiefMaster: I'd do that in the RH after instantiating the form. You could even have an `always_clonable` attribute on `RHAddEventTemplate` / `RHAddCategoryTemplate` and use that flag to determine whether to remove the field from the form and whether to use the field's value when actually creating the template.

indico/migrations/versions/201705221530_3ca338ed5192_remove_background_image_and_add.py
nop33: Well, I checked in our codebase for `op.create_index(` and it seems like we don't have a pattern. But sure, I can use `None`!

indico/migrations/versions/201705241045_bbe2ac15f488_add_is_clonable_column_in_.py
nop33: I don't get that, why? All templates are clonable by default and it's specified in `server_default='true'` right above.

indico/modules/designer/forms.py
nop33: Why is that a better approach than the current one?

ThiefMaster: Like this the logic is in one place (RHs) - you need to check it once to determine whether to remove the field or not and then again when saving to determine whether to set the value on the model or not.

indico/migrations/versions/201705241045_bbe2ac15f488_add_is_clonable_column_in_.py
ThiefMaster: see DEVELOPMENT.md ;)

PR #2949, opened 2017-05-24
Badges designer

nop33:


indico/modules/designer/controllers.py
```diff
@@ -172,14 +172,20 @@ def _process(self):
...
         if not self.target.can_manage(session.user):
             raise Forbidden
 
     def _process(self):
         form = AddTemplateForm()
+        if self.always_clonable:
+            del form.is_clonable
         if form.validate_on_submit():
-            new_template = DesignerTemplate(title=form.title.data, type=form.type.data, **self.target_dict)
+            is_clonable = form.is_clonable.data if form.is_clonable else True
```
ThiefMaster: I'd use `if self.always_clonable` here but no strong opinon. probably your option works too ;)

nop33:
@ThiefMaster Does it make sense to approve a WIP PR?

indico/modules/designer/controllers.py
nop33: It works since `form.is_clonable` returns `None` in case the field doesn't exist in the form.

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -470,20 +471,33 @@
...
-            method: 'POST',
-            complete: IndicoUI.Dialogs.Util.progress(),
-            error: handleAjaxError,
-            success: function(data) {
-                handleFlashes(data, true);
-            }
-        });
+        if (showConfirmationDialog) {
+            confirmPrompt(
+                $T("Are you sure you want to change the dimensions of this template? " +
```
ThiefMaster: ` $T.gettext()` and you most likely cannot use `+` inside translated strings without breaking the extraction. However, having these two sentences in separate gettext calls doesn't seem too bad.

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -470,20 +471,33 @@
...
-            data: JSON.stringify(templateData),
-            contentType: 'application/json',
-            method: 'POST',
-            complete: IndicoUI.Dialogs.Util.progress(),
-            error: handleAjaxError,
-            success: function(data) {
-                handleFlashes(data, true);
-            }
-        });
+        if (showConfirmationDialog) {
```
ThiefMaster: I'd go for something like this:

```javascript
var confirmed = showConfirmationDialog ? confirmPrompt(...) : $.Deferred().resolve();
confirmed.then(saveData);
```

indico/modules/designer/operations.py
```diff
@@ -0,0 +1,57 @@
...
+    :param disassociate_backsides: Whether to remove the specific template from
+                                   being used as the backside of others
+    :param clear_background: Whether to remove the background image of the
+                             template
+    """
+    template.title = title
+    template.backside_template = DesignerTemplate.get(backside_template_id) if backside_template_id else None
+    template.data = dict({'background_position': 'stretch', 'items': []}, **data)
+
+    if disassociate_backsides:
```
ThiefMaster: I don't like the idea of making this an argument, even less letting the client specify its value. I'd rather check where whether the backside dimensions don't match and in that case disassociate it.

indico/modules/designer/templates/template.html
```diff
@@ -16,13 +16,12 @@
...
 {% block content %}
-    {% if related_tpls_per_type %}
+    {% if tpls_count %}
         <div class="affected-targets-warning weak-hidden">
             {% call message_box('warning', fixed_width=true) %}
                 <p>
-                    {%- trans strong_start='<strong>'|safe, strong_end='</strong>'|safe,
-                              link_start='<a href="#" class="js-affected-objects">'|safe, link_end='</a>'|safe,
-                              count=tpls_count -%}
+                    {%- trans count=tpls_count, strong_start='<strong>'|safe, strong_end='</strong>'|safe,
```
ThiefMaster: I'd put these two as HTML in the translated string. Moving very simple tags like `<strong>` to variables just makes the whole thing less readable with no real gain.

indico/modules/designer/templates/template.html
```diff
@@ -16,13 +16,12 @@
...
+                              link_start='<a href="#" class="js-affected-objects">'|safe, link_end='</a>'|safe -%}
```
ThiefMaster: IMO using set blocks would be cleaner, even if it makes the template a big longer. Especially in case we add more attributes in the future (kind of unlikely here...) this would quickly get messy.

indico/htdocs/js/indico/modules/designer.js
nop33: Does this really look prettier?

```js
        var confirmed = showConfirmationDialog ? confirmPrompt(
            $T.gettext("Are you sure you want to change the dimensions of this template?") +
            $T.gettext("The back side of the templates that use this one as the back side will be removed."),
            $T.gettext("Disassociate back sides")
        ) : $.Deferred().resolve();

        confirmed.then(saveData);
```

ThiefMaster: ok with the long call it's quite awful too. but i meant it more in general (creating the deferred and then using `then()` just once), doesn't necessarily need to use the ternary operator.

indico/modules/designer/templates/template.html
```diff
@@ -14,6 +16,37 @@
 {% endblock %}
 
 {% block content %}
+    {% if tpls_count %}
+        <div class="affected-targets-warning weak-hidden">
+            {% call message_box('warning', fixed_width=true) %}
+                <p>
+                    {%- trans count=tpls_count, strong_start='<strong>'|safe, strong_end='</strong>'|safe,
```
ThiefMaster: I'd put these two as HTML in the translated string. Moving very simple tags like `<strong>` to variables just makes the whole thing less readable with no real gain.

indico/modules/designer/templates/template.html
```diff
@@ -14,6 +16,37 @@
...
+                              link_start='<a href="#" class="js-affected-objects">'|safe, link_end='</a>'|safe -%}
```
ThiefMaster: IMO using set blocks would be cleaner, even if it makes the template a big longer. Especially in case we add more attributes in the future (kind of unlikely here...) this would quickly get messy.

indico/modules/designer/operations.py
ThiefMaster: I don't like the idea of making this an argument, even less letting the client specify its value. I'd rather check where whether the backside dimensions don't match and in that case disassociate it.

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -471,17 +474,27 @@
...
-                handleFlashes(data, true);
-            }
-        });
+        var confirmed = showConfirmationDialog ? confirmPrompt(
+            $T.gettext("Are you sure you want to change the dimensions of this template?") +
+            $T.gettext("The back side of the templates that use this one as the back side will be removed."),
+            $T.gettext("Disassociate back sides")
+        ) : $.Deferred().resolve();
+
+        confirmed.then(saveData);
```
ThiefMaster: You could for an anonymous function here since you only use it once now:

    confirmed.then(function() { ... })

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -707,7 +731,7 @@
                 save(template);
             });
 
-            $('.template-width, .template-height').on('keyup click', function() {
+            $('.template-width, .template-height').on('keyup change', function() {
```
ThiefMaster: `input` might be an even better event here

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -809,6 +834,33 @@
             });
 
             $('.backside-tools').addClass('hidden');
+
+            if ($('.affected-targets-warning').length) {
+                var $templateWidth = $('.template-width');
+                var $templateHeight = $('.template-height');
+                $templateWidth.on('change', function() {
```
ThiefMaster: `s/change/input` unless you want it to trigger only when the field loses focus

indico/modules/designer/templates/template.html
```diff
@@ -14,6 +16,38 @@
 {% endblock %}
 
 {% block content %}
+    {% if tpls_count %}
+        <div class="affected-targets-warning weak-hidden">
+            {% call message_box('warning', fixed_width=true) %}
+                <p>
+                    {% set link_start %}<a href="#" class="js-affected-objects">{% endset %}
+                    {% set link_end %}</a>{% endset %}
+                    {%- trans count=tpls_count, link_start=link_start|safe, link_end=link_end|safe -%}
```
ThiefMaster: [set blocks are safe by default](https://github.com/pallets/jinja/commit/6a754cb83a6686534dbc088f677f7a1e7ca3971e), so you can get rid of the whole `, link_start=link_start|safe, link_end=link_end|safe` part

ThiefMaster: You can also get rid of the `count=tpls_count` by using `{%- pluralize tpls_count -%}` below

indico/modules/designer/controllers.py
```diff
@@ -234,20 +242,31 @@ def _process_GET(self):
...
+                              .filter(DesignerTemplate.backside_template_id == self.template.id)
+                              .all())
+        related_tpls_per_event = defaultdict(list)
+        related_tpls_per_categ = defaultdict(list)
+        for bs_tpl in backside_templates:
+            if bs_tpl.event_new:
+                related_tpls_per_event[bs_tpl.event_new].append(bs_tpl)
+            elif bs_tpl.category:
+                related_tpls_per_categ[bs_tpl.category].append(bs_tpl)
+        related_tpls_per_type = {
```
ThiefMaster: You don't really use the type anywhere so why not just mix them together?

```python
for bs_tpl in backside_templates:
    related_tpls[bs_tpl.owner].append(bs_tpl)
```

If it's just for sorting to avoid mixing event/categories you could do that in the query or even split it in the jinja template using selectattr/rejectattr.

    .order_by(DesignerTemplate.event_id.isnot(None), DesignerTemplate.title)

indico/modules/designer/templates/template.html
nop33: The last thing you mentioned doesn't work, I guess I do need to specify the `count=tpls_count`
```
TemplateAssertionError: unknown variable 'tpls_count' for pluralization
```

ThiefMaster: oh yay, jinja being weird there i guess ;x

PR #2950, opened 2017-05-26
Fix typos after French translation

tgbaron:


indico/modules/events/registration/templates/display/registration_summary.html
```diff
@@ -12,7 +12,7 @@
                    {% trans %}Your registration has been completed{% endtrans %}
                </div>
                {% trans -%}
-                   Everything is done.
+                   The process is over.
```
ThiefMaster: IMO this message sounds a bit strange (but so does the old message).

Unless it looks odd maybe we can just omit the message here?

indico/modules/events/surveys/forms.py
```diff
@@ -43,7 +43,7 @@ class SurveyForm(IndicoForm):
     anonymous = BooleanField(_("Anonymous submissions"), widget=SwitchWidget(),
                              description=_("User information will not be attached to submissions"))
     require_user = BooleanField(_("Only logged-in users"), [HiddenUnless('anonymous')], widget=SwitchWidget(),
-                                description=_("Still require users to be logged in for submitting the survey"))
+                                description=_("Require users to be logged in for submitting the survey"))
```
ThiefMaster: IIRC the reason for the "still" was that this only applies to anonymous surveys - i.e. the submission is anonymous, but the user needs to be logged in anyway. But since we hide the option if the survey is not anonymous I guess it's clear without the "Still" (and it indeed sounds better without)

indico/modules/events/registration/templates/display/registration_summary.html
tgbaron: I agree, this additional line can be removed...

pferreir: Now that I see, we did get some complaints about this "Everything is done" message.
Maybe we can just omit it or replace it with something like "Thank you".

pferreir: OK, I'll amend the commit to remove it.

PR #2951, opened 2017-05-26
Fix not aligned abstract reviewing questions

DavidAndreev:


indico/htdocs/sass/modules/_reviewing.scss
```diff
@@ -331,7 +331,7 @@
 
             .details-content-row {
                 color: $dark-gray;
-                padding: 5px 10px 5px calc(20px + 1em);
+                padding: 5px 10px;
```
nop33: After this change the UI looks like this:
image

We want it to look like this: 
image

This can be achived with:
``` sass
padding: 5px 10px 5px calc(22px + 2em);
```
because of https://github.com/indico/indico/pull/2951/files#diff-65da6c2d419c1459c8181a9673c06dc3L292 (and the extra margin/padding)

DavidAndreev: Hmmm ok, I thought we want it all the way on the left, thanks !

PR #2963, opened 2017-06-12
Extend email sender options

nop33:


indico/modules/events/agreements/forms.py
```diff
@@ -43,10 +41,9 @@ class AgreementEmailForm(IndicoForm):
 
     def __init__(self, *args, **kwargs):
         self._definition = kwargs.pop('definition')
+        event = kwargs.pop('event')
         super(AgreementEmailForm, self).__init__(*args, **kwargs)
-        from_addresses = ['{} <{}>'.format(session.user.full_name, email)
-                          for email in sorted(session.user.all_emails, key=lambda x: x != session.user.email)]
-        self.from_address.choices = zip(from_addresses, from_addresses)
+        self.from_address.choices = event.get_allowed_sender_emails(extra=self.from_address.object_data).items()
```
ThiefMaster: The `extra=self.from_address.object_data` is only needed/useful in cases where the field's value is stored somewhere. The idea behind it is that even if that email would not be available anymore (e.g. because a manager was removed) saving the form should be possible without changing it. However, that's never the case in a form where no initial value is loaded from the db

PR #2965, opened 2017-06-13
Fix escaping in html email subjects

DavidAndreev:


indico/modules/vc/templates/emails/deleted.html
```diff
@@ -1,7 +1,9 @@
 {% extends 'emails/base.html' %}
 
 {% block subject -%}
-    [{{ plugin.friendly_name }}] Videoconference room deleted: {{ vc_room.name }}
+    {% autoescape false %}
```
ThiefMaster: strip whitespace inside. this is an email subject and it might look ugly if there's surrounding whitespace (not sure if we strip it in make_email or not)

same in the other file

PR #2966, opened 2017-06-13
Solve wrong date sorting in participant list

mvidalgarcia:
Related snow ticket `INC1383402`


indico/modules/events/registration/controllers/display.py
```diff
@@ -143,7 +143,8 @@ def _content(column_id):
                     # no data available for the field
                     return ''
 
-            columns = [{'text': _content(column_id)} for column_id in column_ids]
+            columns = [{'text': _content(column_id), 'sort_key': data_by_field.get(column_id).data}
```
ThiefMaster: `data_by_field.get(column_id)` may return `None`

PR #2977, opened 2017-06-22
Changes for the cern access plugin

meluru:


indico/core/signals/event/registration.py
```diff
@@ -20,6 +20,10 @@
 
 _signals = Namespace()
 
+registration_modified = _signals.signal('registration_modified', """
```
ThiefMaster: `registration-modified`

ThiefMaster: Right now `change` only includes personal data changes. What about making this signal specific to personal data changes too? Otherwise its a bit confusing if `change` is empty even though something did change

indico/modules/events/requests/models/requests.py
```diff
@@ -143,6 +143,8 @@ def definition(self, definition):
     @property
     def can_be_modified(self):
         """Determines if the request can be modified or if a new one must be sent"""
+        if (self.definition.modifiable is False):
```
ThiefMaster: `if not self.definition.modifiable` ... 🙈

indico/modules/events/requests/templates/event_request_details.html
```diff
@@ -76,7 +76,7 @@
     {% endif %}
 
     {% call form_footer(form) %}
-        <input class="i-button big highlight" type="submit" value="
+        <input id="send-request" class="i-button big highlight" type="submit" data-disabled-until-change value="
```
ThiefMaster: We usually put data attributes at the end, after all other attributes

meluru:
Updated

indico/modules/events/registration/util.py
```diff
@@ -236,6 +236,7 @@ def create_registration(regform, data, invitation=None, management=False, notify
 
 def modify_registration(registration, data, management=False, notify_user=True):
     old_price = registration.price
+    personal_data_change = {}
```
ThiefMaster: s/change/changes

indico/modules/events/requests/templates/event_request_details.html
```diff
@@ -76,15 +76,15 @@
...
-        <input class="i-button big highlight" type="submit" value="
+        <input id="send-request" class="i-button big highlight" type="submit" value="
             {%- if not req -%}
                 {%- trans -%}Send request{%- endtrans -%}
             {%- elif req.can_be_modified -%}
                 {%- trans -%}Modify request{%- endtrans -%}
             {%- else -%}
                 {%- trans -%}Send new request{%- endtrans -%}
             {%- endif -%}
-        ">
```
ThiefMaster: Maybe we can use the opportunity to move this huge block to  `{% set ... %}` block so the definition of this HTML tag is less awful (we could then just use `value="{{ button_title }}"`)?

meluru:
Updated

indico/modules/events/requests/base.py
```diff
@@ -62,6 +62,8 @@ class RequestDefinitionBase(object):
     manager_form = RequestManagerForm
     #: default values to use if there's no existing request
     form_defaults = {}
+    #: can request be modified
```
ThiefMaster: can *the* request be modified

indico/modules/events/requests/templates/event_request_details.html
```diff
@@ -5,10 +5,17 @@
 
 {% set has_req = req and not req.state.name == 'withdrawn' %}
 {% set show_management = is_manager and has_req %}
+{%- if not req -%}
```
ThiefMaster: missing i18n on the sets. you could either use `_()` in there or use a single set block:

```
{% set button_title -%}
    {%- if not req -%}
        {%- trans %}Send request{% endtrans -%}
    {%- elif ... -%}
        ...
    {%- endif -%}
{%- endset %}
```

indico/modules/events/requests/base.py
```diff
@@ -62,6 +62,8 @@ class RequestDefinitionBase(object):
...
+    modifiable = True
```
ThiefMaster: should we get rid of this since it's not needed anymore?

indico/core/signals/event/registration.py
```diff
@@ -18,8 +18,13 @@
 
 from blinker import Namespace
 
+
 _signals = Namespace()
 
+registration_personal_data_modified = _signals.signal('registration-personal-data-modified', """
+Called when the registration personal data is modified. The `sender` is the
+registration; the change is passed in the `change` kwarg.
```
nop33: Shouldn't you say here "The `sender` is the `Registration` object"? (as you do right below)

indico/modules/events/registration/util.py
```diff
@@ -18,33 +18,33 @@
...
 from indico.core.db import db
 from indico.modules.events import EventLogKind, EventLogRealm
 from indico.modules.events.models.events import Event
 from indico.modules.events.registration import logger
-from indico.modules.events.registration.fields.choices import (ChoiceBaseField, AccommodationField,
+from indico.modules.events.registration.fields.choices import (AccommodationField, ChoiceBaseField,
                                                                get_field_merged_options)
-from indico.modules.events.registration.models.form_fields import (RegistrationFormPersonalDataField,
-                                                                   RegistrationFormFieldData)
+from indico.modules.events.registration.models.form_fields import (RegistrationFormFieldData,
```
nop33: Isn't the `_` supposed to be sorted after the `s` character? (if so, swap line 35 and 37)

indico/modules/events/requests/templates/event_request_details.html
```diff
@@ -76,15 +83,8 @@
...
-            {%- if not req -%}
-                {%- trans -%}Send request{%- endtrans -%}
-            {%- elif req.can_be_modified -%}
-                {%- trans -%}Modify request{%- endtrans -%}
-            {%- else -%}
-                {%- trans -%}Send new request{%- endtrans -%}
-            {%- endif -%}
-        ">
+        <input id="send-request" class="i-button big highlight" type="submit"
+               value="{% trans %}{{ button_title }}{% endtrans %}" data-disabled-until-change>
```
nop33: As @ThiefMaster mentioned, you cannot do the translation like this. The content of variables is not translated inside the `{% trans %}` block.

ThiefMaster: it's translated if it's in the i18n dict but it won't be *extracted* (so it'd unlikely to be in the i18n dict)

indico/modules/events/registration/util.py
ThiefMaster: dunno, let isort decide

meluru:
Updated

indico/modules/events/requests/templates/event_request_details.html
```diff
@@ -5,10 +5,19 @@
 
 {% set has_req = req and not req.state.name == 'withdrawn' %}
 {% set show_management = is_manager and has_req %}
+{% set button_title -%}
+    {%- if not req -%}
+        {% trans %} Send request {% endtrans %}
```
ThiefMaster: Please don't keep surrounding whitespace inside trans blocks. This is rather ugly during translation since translators need to keep them there too. If it's relevant, you could always add it at the place where you use the translated string (ie keep the space outside the trans blocks) but here it doesn't seem to be needed at all.

ThiefMaster: Actually, we strip this whitespace when extracting it, so it makes no real difference but it's confusing to have this whitespace in the template

meluru:
Updated

indico/modules/events/requests/templates/event_request_details.html
```diff
@@ -5,10 +5,19 @@
...
+        {% trans %}Modify request{% endtrans %}
+    {%- else -%}
+        {% trans %}Send new request{% endtrans %}
+    {%- endif -%}
+{%- endset %}
 
 {% block title %}{% trans %}Services{% endtrans %}{% endblock %}
 {% block subtitle %}{{ definition.title }}{% endblock %}
-
+{% block description %}{% endblock %}
```
ThiefMaster: this shouldn't be necessary; it inherits from an already-empty description block

ThiefMaster:
Included in #3026

PR #2983, opened 2017-06-27
Change fs:/// to file:/// in migrate example

datty:
Example fails as is, indico-migrate doesn't recognise fs:///


ThiefMaster:
Thanks!

PR #2988, opened 2017-06-28
Typos in text string

DirkHoffmann:


ThiefMaster:
Thanks! However, we do not use an 80-col limit anywhere - in Python we use a hard 120 char limit (soft 72 for natural text blocks, ie comments and docstrings), in templates and JavaScript a soft 120 limit.

Considering that the timezone selector is legacy code which will eventually be replaced, I think it would make more sense to just focus the PR on fixing the i18n typo. Besides that, reformatting is generally better done in a separate commit - otherwise it's very hard to distinguish the actual change and the format changes when looking at the diff.

Also, small suggestion: rebase feature branches on top of this repo's master - merge commits in pull requests are usually noise (we can squash them away when merging a PR but it's cleaner to avoid them in the firs place).

DirkHoffmann:
Thank you for all those explanations. They all sound very reasonable to me. I just was not sure, which rules should be applied. And for many things I just do not know how to carry them out technically. Can you give me the two, three lines of git commands you want me to execute? I will be happy to perform them (and write them down for future purpose).

ThiefMaster:
`git rebase upstream/master` (assuming the remote pointing to our repo is called 'upstream')

However, for this specific commit it might be easier to just `git reset --hard upstream/master` (this throws away all commits in the branch and also local changes on tracked files and sets it to be the same as master) and then just apply the fix(es) on the translatable string(s). If you want to update the translation templates you'd then run `indico i18n extract_messages` and `indico i18n extract_messages_js`, but feel free to skip that - we do this from time to time anyway so we can just do it after merging your PR.

By the way, I noticed that your commit added trailing whitespace in some lines (the colored output of `git show` and `git diff` highlights this). You might want to configure your editor to always strip trailing whitespace when saving a file.

DirkHoffmann:
@ThiefMaster, is there anything you expect me to do on this?
If there are no simple commands, I can remove the commit from the branch, squash, make 2 separate edits and redo the PR. Is that what you recommend?

ThiefMaster:
Yeah, removing the commits (hence the suggestion to `git reset --hard upstream/master`) and doing the typo change again is the easiest option here. You can just force-push to the same branch after committing to update this PR, so no need to open a new one.

PR #2992, opened 2017-07-03
Always delete contribution on withdrawing abstract

### LINKED ISSUE #2991: Unacceptable abstracts caused by accept/reset/withdraw/restore
Withdrawing disassociates the contribution from the abstracts, but the contribution is only deleted when the abstract was in the 'accepted' state. So after restoring it there's still a contribution with the same friendly id, which is not tied to the abstract anymore, so accepting it again fails

DavidAndreev:
Closes #2991


ThiefMaster:
Check the template where the confirmation prompt is shown; it also checks the state there

PR #3000, opened 2017-07-07
Fix registration limit bug

meluru:


indico/modules/events/registration/controllers/display.py
```diff
@@ -258,8 +258,12 @@ def _checkParams(self, params):
     def _process(self):
         form = make_registration_form(self.regform)()
         if form.validate_on_submit():
-            registration = create_registration(self.regform, form.data, self.invitation)
-            return redirect(url_for('.display_regform', registration.locator.registrant))
+            if self.regform.limit_reached:
```
ThiefMaster: I would move this check to the very beginning (could even be in `_checkProtection`). Even displaying the form doesn't make much sense once the limit has been reached.

indico/modules/events/registration/controllers/display.py
```diff
@@ -258,8 +258,12 @@ def _checkParams(self, params):
...
+                flash(_("Your registration wasn't successful because registrations limit has been reached"), 'error')
```
ThiefMaster: "was not" (the "*n't" forms are quite colloquial)

But I'd go for something like "The maximum number of registrations has already been reached" - with the 'error' style it should still be clear that you did not successfully register.

indico/modules/events/registration/controllers/display.py
```diff
@@ -258,8 +258,12 @@ def _checkParams(self, params):
...
+                return redirect(url_for('.display_regform', self.regform))
```
ThiefMaster: Maybe better to redirect to `.display_regform_list`?

indico/modules/events/registration/controllers/display.py
```diff
@@ -248,6 +248,8 @@ def _checkProtection(self):
         if self.regform.require_login and not session.user and request.method != 'GET':
             raise Forbidden(response=redirect_to_login(reason=_('You are trying to register with a form '
                                                                 'that requires you to be logged in')))
+        if self.regform.limit_reached:
+            return redirect(url_for('.display_regform_list', self.regform))
```
ThiefMaster: the return value of `_checkProtection` is not used

meluru:
updated

PR #3002, opened 2017-07-12
Small corrections to apache.rst and upgrade.rst

shesselbach:
Debian apache.rst: Add two more dependencies: postgresql-contrib-9.6 and virtualenv

upgrade.rst: Changed `--archive-dir /opt/indico/archive` to `--archive-dir /opt/indico-legacy/archive` in Copy & Paste example for "Running indico-migrate"

### DIFF
--- a/docs/source/installation/production/debian/apache.rst
+++ b/docs/source/installation/production/debian/apache.rst
@@ -13,7 +13,7 @@ PostgreSQL is installed from its upstream repos to get a much more recent versio
     echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list
     wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
     apt update
-    apt install -y postgresql-9.6 postgresql-contrib-9.6 libpq-dev apache2 libapache2-mod-proxy-uwsgi libapache2-mod-xsendfile python-dev python-virtualenv virtualenv libxslt1-dev libxml2-dev libffi-dev libpcre3-dev libyaml-dev build-essential redis-server uwsgi uwsgi-plugin-python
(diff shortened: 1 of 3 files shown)


ThiefMaster:
On which exact OS version did you need to install these two packages? `python-virtualenv` should be fine to get virtualenv and IIRC the postgres extensions worked fine  when I was testing without having to install the contrib package too

ThiefMaster:
> root@indico-fromscratch:~# apt install postgresql-9.6 libpq-dev python-virtualenv
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following extra packages will be installed:
  libldap-2.4-2 libpq5 libsasl2-2 libsasl2-modules libsasl2-modules-db libsensors4 pgdg-keyring postgresql-client-9.6 postgresql-client-common postgresql-common **postgresql-contrib-9.6** python-chardet-whl python-colorama-whl python-distlib-whl python-html5lib-whl python-pip-whl python-requests-whl
  python-setuptools-whl python-six-whl python-urllib3-whl python3-pkg-resources python3-virtualenv ssl-cert sysstat **virtualenv**


Any chance you have some option enabled to no not install "extra" packages?

ThiefMaster:
OK I see, apparently you somehow have `--no-install-recommends` enabled. I wonder if adding `--install-recommends` is the better choice...

ThiefMaster:
It's only the two packages you added (also in the nginx-based setup), but based on https://unix.stackexchange.com/questions/77053/apt-installing-more-packages-than-specified-as-dependencies/77076#77076 I'd go for `--install-recommends`.

shesselbach:
I installed Indico on a newly setup Debian Jessie VM from our Linux experts. And indeed on these machines recommended packeges never get installed, i.e. `--no-install-recommends` is somehow the default. I agree that `--install-recommends` might be the better solution. I will update the PR.

ThiefMaster:
Thanks! Can you also update nginx.rst since it uses the same packages for postgres/virtualenv?

ThiefMaster:
thanks, merged!

PR #3006, opened 2017-07-13
Update upgrade.rst

shesselbach:
`/opt/indico-legacy` should be owned by user **indico** because
 * `indico-migrate` needs to create a lock file in db subdir
 * some files in archive may be only accessible for the owner

### DIFF
--- a/docs/source/installation/upgrade.rst
+++ b/docs/source/installation/upgrade.rst
@@ -132,10 +132,11 @@ When the wizard asks you about the **"Old archvive dir"**, make sure to set it t
 Running ``indico-migrate``
 ++++++++++++++++++++++++++
 
-First of all, make sure that you are using the **user** and **virtualenv** created using the step **"Install Indico"**:
+First of all, make sure that you are using the **user** and **virtualenv** created using the step **"Install Indico"** and that the legacy dir is owned by this **user**:
 
 .. code-block:: shell
 
+    chown -R indico /opt/indico-legacy
     su - indico
     source ~/.venv/bin/activate
 


docs/source/installation/upgrade.rst
```diff
@@ -132,10 +132,11 @@ When the wizard asks you about the **"Old archvive dir"**, make sure to set it t
 Running ``indico-migrate``
 ++++++++++++++++++++++++++
 
-First of all, make sure that you are using the **user** and **virtualenv** created using the step **"Install Indico"**:
+First of all, make sure that you are using the **user** and **virtualenv** created using the step **"Install Indico"** and that the legacy dir is owned by this **user**:
 
 .. code-block:: shell
 
+    chown -R indico /opt/indico-legacy
```
ThiefMaster:
```sh
chown indico /opt/indico-legacy
chown -R indico /opt/indico-legacy/archive
```

no need to chown everything else in there

shesselbach: And `chown -R indico /opt/indico-legacy/db` otherwise I get
```
IOError: [Errno 13] Permission denied: '/opt/indico-legacy/db/Data.fs.lock'
IOError: [Errno 13] Permission denied: '/opt/indico-legacy/db/Data.fs.tmp'
IOError: [Errno 13] Permission denied: '/opt/indico-legacy/db/Data.fs'
...
```

ThiefMaster: ah right, forgot that the database is also in there. yeah, then it's probably easier to chown everything in there.

shesselbach: Should I change it to
```
chown indico /opt/indico-legacy
chown -R indico /opt/indico-legacy/archive
chown -R indico /opt/indico-legacy/db
```
or should we stay with the simpler
```
chown -R indico /opt/indico-legacy
```

ThiefMaster: i think it's ok to keep the latter - if someone needs to go back to 1.2 for some reason they'll have to change ownership back either way

PR #3008, opened 2017-07-17
Fix unavailable periods calendar icon

DavidAndreev:


indico/htdocs/sass/modules/_roombooking.scss
```diff
@@ -349,3 +349,9 @@ table.booking-stats {
         }
     }
 }
+
+#modify-room {
+    .subFieldWidth {
+        width: 118px;
```
ThiefMaster: this feels like a dirty hack (especially adding that ID and that hardcoded size)... why not fix the code/css of the field. for example, you could use absolute positioning instead of a float to position the icon

DavidAndreev:
Now it looks like this, but maybe we can still reduce the width of the input or the field caption so the "add period" can fit in one line
unav

ThiefMaster:
I wouldn't bother too much. This page is still pretty much legacy and will eventually be improved a lot.

PR #3026, opened 2017-07-24
Ticket templates

DavidAndreev:


indico/migrations/versions/201707241125_75060428e775_add_ticket_template_to_registration_.py
```diff
@@ -0,0 +1,31 @@
...
+# revision identifiers, used by Alembic.
+revision = '75060428e775'
+down_revision = 'ebca3e91b139'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('forms', sa.Column('ticket_template_id', sa.Integer(), nullable=True), schema='event_registration')
+    op.create_index(op.f('ix_forms_ticket_template_id'), 'forms', ['ticket_template_id'], unique=False,
```
ThiefMaster: first argument can be `None`

indico/migrations/versions/201707241125_75060428e775_add_ticket_template_to_registration_.py
```diff
@@ -0,0 +1,31 @@
...
+                    schema='event_registration')
+    op.create_foreign_key(op.f('fk_forms_ticket_template_id_designer_templates'), 'forms', 'designer_templates',
```
ThiefMaster: first argument can be `None`

indico/migrations/versions/201707241125_75060428e775_add_ticket_template_to_registration_.py
```diff
@@ -0,0 +1,31 @@
...
+                          ['ticket_template_id'], ['id'], source_schema='event_registration', referent_schema='indico')
+
+
+def downgrade():
+    op.drop_constraint(op.f('fk_forms_ticket_template_id_designer_templates'), 'forms',
```
ThiefMaster: most likely you only need to drop the column here, and can omit the first two statements

indico/modules/events/registration/controllers/management/tickets.py
```diff
@@ -56,7 +57,9 @@ def _check_ticket_app_enabled(self):
         return True
 
     def _process(self):
-        form = TicketsForm(obj=self.regform)
+        all_templates = set(self.event_new.designer_templates) | get_inherited_templates(self.event_new)
+        badge_templates = [tpl for tpl in all_templates if tpl.type.name == 'badge']
```
ThiefMaster: `tpl.type == TemplateType.badge`

indico/modules/events/registration/forms.py
```diff
@@ -273,6 +273,14 @@ class TicketsForm(IndicoForm):
                                           description=_('Allow users to download their ticket from the registration '
                                                         'summary page.'))
 
+    ticket_template_id = SelectField(_('Ticket template'), [HiddenUnless('tickets_enabled', preserve_data=True),
+                                                            Optional()])
+
+    def __init__(self, *args, **kwargs):
+        super(TicketsForm, self).__init__(*args, **kwargs)
+        templates = kwargs.pop('templates')
```
ThiefMaster: do this before the `super()` call. also, I would move the logic to get the list of templates in here (and pass the event to the form instead of a list of templates)

indico/modules/events/registration/models/forms.py
```diff
@@ -228,6 +229,19 @@ class RegistrationForm(db.Model):
         default=True
     )
 
+    ticket_template_id = db.Column(
```
ThiefMaster: see the structure of the rest of the file: doc comment above the column, no empty line in between, and i would put the relationship right after the `event_new` one

indico/modules/events/registration/forms.py
ThiefMaster: Actually, I think you can use a QuerySelectField here. For the query:

```
self.ticket_template.query = (DesignerTemplate.query
                              .filter(DesignerTemplate.type == TemplateType.badge,
                                      db.or_(DesignerTemplate.category_id.in_(event.category_chain),
                                             DesignerTemplate.event_new == event)))
```

indico/migrations/versions/201707241125_75060428e775_add_ticket_template_to_registration_.py
```diff
@@ -0,0 +1,28 @@
...
+revision = '75060428e775'
+down_revision = 'ebca3e91b139'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('forms', sa.Column('ticket_template_id', sa.Integer(), nullable=True), schema='event_registration')
+    op.create_index(None, 'forms', ['ticket_template_id'], unique=False,
+                    schema='event_registration')
```
ThiefMaster: this fits find in the previous line

indico/modules/events/registration/forms.py
```diff
@@ -273,6 +276,17 @@ class TicketsForm(IndicoForm):
                                           description=_('Allow users to download their ticket from the registration '
                                                         'summary page.'))
 
+    ticket_template = QuerySelectField(_('Ticket template'), [HiddenUnless('tickets_enabled', preserve_data=True),
+                                                              Optional()], get_label='title', allow_blank=False)
```
ThiefMaster: why no allow_blank? Isn't "use the default" leaving the field blank?

indico/modules/events/registration/models/forms.py
```diff
@@ -239,6 +247,12 @@ class RegistrationForm(db.Model):
             lazy=True
         )
     )
+    #: The ticket template used in PDF
+    ticket_template = db.relationship(
+        'DesignerTemplate',
+        lazy=True,
+        foreign_keys=ticket_template_id
```
ThiefMaster: Usually we add a backref to almost all relationships. In this case I would call it e.g. `ticket_for_regforms`. Having such relationships - even if not used in the code - makes it easier to do stuff in indico shell.

Don't forget to run `python bin/maintenance/update_backrefs.py` after adding a backref to update the comment in the referenced model.

indico/modules/events/registration/models/forms.py
```diff
@@ -227,6 +228,13 @@ class RegistrationForm(db.Model):
         nullable=False,
         default=True
     )
+    #: The id of ticket template used in PDF
```
ThiefMaster: > The ID of the template used to generate tickets

indico/modules/events/registration/models/forms.py
```diff
@@ -239,6 +247,12 @@ class RegistrationForm(db.Model):
             lazy=True
         )
     )
+    #: The ticket template used in PDF
```
ThiefMaster: > The template used to generate tickets

indico/htdocs/sass/modules/_designer.scss
```diff
@@ -352,3 +352,24 @@
...
+
+    &.default-on-category {
+        opacity: 1;
+        color: $blue;
+
+        &.inherited-from-root {
+            opacity: 0.6;
+
+            &:hover {
+                color: $blue;
```
pferreir: Isn't the color already blue from `&.default-on-category`? Maybe you wanted to set opacity to 1 instead?

indico/migrations/versions/201707241125_75060428e775_add_ticket_template_to_registration_.py
```diff
@@ -0,0 +1,27 @@
...
+# revision identifiers, used by Alembic.
+revision = '75060428e775'
+down_revision = 'ebca3e91b139'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('forms', sa.Column('ticket_template_id', sa.Integer(), nullable=True), schema='event_registration')
+    op.create_index(None, 'forms', ['ticket_template_id'], unique=False, schema='event_registration')
```
pferreir: Isn't this equivalent to `index=True` in the previous line?

indico/migrations/versions/201707251214_fd47fc619616_add_default_ticket_template_to_category_.py
```diff
@@ -0,0 +1,28 @@
...
+revision = 'fd47fc619616'
+down_revision = '75060428e775'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('categories', sa.Column('default_ticket_template_id', sa.Integer(), nullable=True),
+                  schema='categories')
+    op.create_index(None, 'categories', ['default_ticket_template_id'], unique=False, schema='categories')
```
pferreir: Same here. I think this can be set on the previous line's `Column`?

indico/modules/categories/models/categories.py
```diff
@@ -175,6 +175,11 @@ def __table_args__(cls):
         nullable=False,
         default=False
     )
+    default_ticket_template_id = db.Column(
+        db.ForeignKey('indico.designer_templates.id'),
+        nullable=True,
```
pferreir: Maybe make it non-nullable and set the default to False?

indico/modules/designer/blueprint.py
```diff
@@ -53,6 +54,8 @@ def view_func(**kwargs):
                      defaults={'object_type': object_type}, methods=('GET', 'POST'))
     _bp.add_url_rule(prefix + '/<int:template_id>/', 'edit_template', RHEditDesignerTemplate,
                      defaults={'object_type': object_type}, methods=('GET', 'POST'))
+    _bp.add_url_rule(prefix + '/<int:template_id>/toggle_default/', 'toggle_category_default',
+                     RHToggleTemplateDefaultOnCategory, methods=('GET', 'POST'))
```
pferreir: Why do we need `GET` here?

indico/modules/designer/controllers.py
```diff
@@ -75,8 +77,10 @@
 
 
 def _render_template_list(target, event):
+    from indico.modules.designer.util import get_default_template_on_category
```
pferreir: Does this need to be imported from here?

indico/modules/designer/controllers.py
```diff
@@ -142,7 +146,10 @@ def target(self):
 
 class TemplateListMixin(TargetFromURLMixin):
     def _process(self):
-        return self._render_template('list.html', inherited_templates=get_inherited_templates(self.target))
+        from indico.modules.designer.util import get_default_template_on_category
```
pferreir: Same here.

indico/modules/designer/controllers.py
```diff
@@ -303,6 +310,12 @@ def _process(self):
 
 class RHDeleteDesignerTemplate(RHModifyDesignerTemplateBase):
     def _process(self):
+        affected_regforms = RegistrationForm.find_all(RegistrationForm.ticket_template_id == self.template.id)
```
pferreir: You can abbreviate this to:

```python
RegistrationForm.find_all(ticket_template_id=self.template.id)
```

indico/modules/designer/controllers.py
```diff
@@ -303,6 +310,12 @@ def _process(self):
...
+        for regform in affected_regforms:
+            regform.ticket_template = None
+        affected_categories = Category.find_all(Category.default_ticket_template_id == self.template.id)
```
pferreir: Same here.

indico/modules/designer/controllers.py
```diff
@@ -327,3 +340,20 @@ def _process(self):
...
+        RHManageCategoryBase._checkParams(self)
+        self.template = DesignerTemplate.get_one(request.view_args['template_id'])
+        category_id = request.view_args['category_id']
+        self.category = self._get_category(category_id)
+
+    def _process(self):
+        if self.category.default_ticket_template == self.template:
+            self.category.default_ticket_template = None
+        else:
+            self.category.default_ticket_template = self.template
```
pferreir: This could be reduced to a single-line conditional expression.

indico/modules/designer/models/templates.py
```diff
@@ -81,9 +83,15 @@ class DesignerTemplate(db.Model):
         nullable=False,
         default=True
     )
+    not_deletable = db.Column(
+        db.Boolean,
+        nullable=True,
```
pferreir: Are you sure this needs to be nullable?

indico/modules/designer/models/templates.py
```diff
@@ -130,6 +139,15 @@ def __init__(self, **kwargs):
     def owner(self):
         return self.event_new if self.event_new else self.category
 
+    @property
+    def is_not_deletable(self):
```
pferreir: I think this name is highly confusing, since `not_deletable` already exists and they hold different meanings.
I would call it `can_be_deleted` or similar. It's still not perfect, but at least it won't be confused with the column.

indico/modules/events/registration/controllers/management/tickets.py
```diff
@@ -90,7 +101,18 @@ def _checkProtection(self):
 
     def _process(self):
         filename = secure_filename('{}-Ticket.pdf'.format(self.event_new.title), 'ticket.pdf')
-        return send_file(filename, generate_ticket(self.registration), 'application/pdf')
+        template = self.regform.ticket_template
+        if template is None:
+            from indico.modules.designer.util import get_default_template_on_category
```
pferreir: Inline import once again. Is this really needed?

indico/modules/events/registration/forms.py
```diff
@@ -273,6 +276,21 @@ class TicketsForm(IndicoForm):
                                           description=_('Allow users to download their ticket from the registration '
                                                         'summary page.'))
 
+    ticket_template_id = SelectField(_('Ticket template'), [HiddenUnless('tickets_enabled', preserve_data=True),
+                                                            Optional()], coerce=lambda x: int(x))
+
+    def __init__(self, *args, **kwargs):
+        event = kwargs.pop('event')
+        from indico.modules.designer.util import get_default_template_on_category
```
pferreir: Inline import.

indico/modules/designer/util.py
```diff
@@ -38,3 +38,17 @@ def get_all_templates(obj):
 def get_inherited_templates(obj):
     """Get all templates inherited by a given event/category"""
     return get_all_templates(obj) - set(obj.designer_templates)
+
+
+def get_default_template_on_category(obj):
+    if isinstance(obj, Event):
```
pferreir: I believe you've added this here as a way to protect it from the case where `_render_template_list` calls it with an event. But I don't think it makes much sense to plan for a case like this in a function that is `_on_category`. I would instead do the check on the calling function(s).

indico/modules/designer/util.py
```diff
@@ -38,3 +38,17 @@ def get_all_templates(obj):
...
+        return None
+    if obj.default_ticket_template:
+        return obj.default_ticket_template
+    if obj.is_root:
+        default_root_tpl = DesignerTemplate.find_first(DesignerTemplate.not_deletable)
```
pferreir: I may be confused, but won't this return the first template it finds, regardless of whether it's attached to the root or not?

indico/modules/events/registration/controllers/management/tickets.py
```diff
@@ -90,7 +101,18 @@ def _checkProtection(self):
...
+            template = get_default_template_on_category(self.event_new.category)
+        signals.event.designer.print_badge_template.send(template, regform=self.regform)
+        if template.backside_template:
+            pdf_class = RegistrantsListToBadgesPDFFoldable
+        else:
+            pdf_class = RegistrantsListToBadgesPDF
```
pferreir: I'd use a conditional expression here.

indico/htdocs/sass/modules/_designer.scss
DavidAndreev: Well this is in case of default template inherited from root category which cannot be unclicked so I didn't want it to turn black on hover so it doesn't seem like you can click it

indico/modules/designer/blueprint.py
ThiefMaster: besides that, we use dashes instead of underscores in URLs

indico/modules/categories/models/categories.py
DavidAndreev: But why default to False if its an Integer field? And also in the code I check whether the `default_ticket_template` is None and I even set it to None when deleting the template, so should this `default_ticket_template_id` be non-nullable ?

pferreir: Whoops. Disregard my comment, my mistake.

indico/modules/designer/util.py
DavidAndreev: Well this is exactly in case we delete a template from the root category, so theres no template attached to the root category at this moment, therefore we look for the non-deletable one and assign it to the root category

pferreir: Isn't we weird we're doing such an assignment in a function called `get_default_template_on_category `? That's a side effect I wouldn't expect. Shouldn't we do that check outside the function, in the handler that calls it (right after deletion)?

DavidAndreev: Thats a good point, or can I do it directly upon deletion ? There I set it to None for all the affected categories so I would just check if its root category and do the same thing

indico/core/signals/event/designer.py
```diff
@@ -22,3 +22,8 @@
 Called when printing a badge template. Registration form is passed in the 'regform' kwarg.
 kwarg.
 """)
+
+generate_ticket_qr_code = _signals.signal('generate-ticket-qr-code', """
+Called when generating a ticket QR code. QR object and QR data are p
```
ThiefMaster: `p`? Also, mentioning kwargs without mentioning their names is a bit pointless ;)

indico/modules/events/registration/util.py
```diff
@@ -438,8 +439,12 @@ def generate_ticket_qr_code(registration):
         "event_id": unicode(registration.event_new.id),
         "server_url": Config.getInstance().getBaseURL()
     }
-    json_qr_data = json.dumps(qr_data)
-    qr.add_data(json_qr_data)
+    extended_qr = signals.event.designer.generate_ticket_qr_code.send(registration, qr_data=qr_data)
```
ThiefMaster: I would rather call the signal with just the registration as the sender and the `qr_data` kwarg (but maybe renamed to `ticket_data`), and let the signal modify the dict directly instead of returning a new one. Your current code won't work well when there's more than one plugin connecting to the signal.

Generally it's a good idea to use `values_from_signal(...)` when dealing with signals that return values, but if you go for changing the dict directly you won't need that.

indico/modules/events/registration/util.py
```diff
@@ -438,8 +439,12 @@ def generate_ticket_qr_code(registration):
...
+    if extended_qr:
+        qr = extended_qr[0][1]
```
ThiefMaster: does it really make sense to override the generation of the QR code here instead of just generating it with the updated data dict?

indico/core/signals/event/designer.py
```diff
@@ -22,3 +22,8 @@
 Called when printing a badge template. Registration form is passed in the 'regform' kwarg.
 kwarg.
 """)
+
+generate_ticket_qr_code = _signals.signal('generate-ticket-qr-code', """
+Called when generating a ticket QR code. QR data are passed passed in the 'qr_data' kwarg.
```
ThiefMaster: > Called when generating the QR code for a ticket. The data included in the QR code is passed
> in the `ticket_data` kwarg and may be modified.

indico/modules/events/registration/util.py
```diff
@@ -17,13 +17,16 @@
 from __future__ import unicode_literals
 
 from collections import OrderedDict
+from random import random
```
ThiefMaster: unused import?

indico/modules/events/registration/util.py
```diff
@@ -17,13 +17,16 @@
...
 
 from flask import current_app, json, session
+from pytz import timezone
```
ThiefMaster: unused

indico/core/signals/event/designer.py
```diff
@@ -22,3 +22,8 @@
 Called when printing a badge template. Registration form is passed in the 'regform' kwarg.
 kwarg.
 """)
+
+generate_ticket_qr_code = _signals.signal('generate-ticket-qr-code', """
```
ThiefMaster: isn't this signal more suitable in signals/event/registration.py? ticketing is more a registration feature than a designer feature.

indico/core/signals/event/designer.py
```diff
@@ -0,0 +1,23 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from blinker import Namespace
+
+_signals = Namespace()
+
+print_badge_template = _signals.signal('print-badge-template', """
+Called when printing a badge template. Registration form is passed in the 'regform' kwarg.
```
ThiefMaster: `*The* registration form`, and backticks to quote `regform`.

indico/modules/designer/models/templates.py
```diff
@@ -130,6 +139,15 @@ def __init__(self, **kwargs):
     def owner(self):
         return self.event_new if self.event_new else self.category
 
+    @property
+    def can_be_deleted(self):
+        from indico.modules.events.registration.models.forms import RegistrationForm
+        active_regforms = RegistrationForm.find_all(RegistrationForm.ticket_template == self,
+                                                    Event.ends_after(now_utc()), _join=Event)
```
ThiefMaster: Please do not use `find()` or `find_all()` with `_join` and similar shortcuts. We pretty much stopped using them many versions ago. I'd actually avoid using the `find` methods in favor of `query` at all - ie do this:

```python
active_regforms = (RegistrationForm.query
                   .filter(RegistrationForm.ticket_template == self,
                           RegistrationForm.event_new.has(Event.ends_after(now_utc())))
                   .all())
```

or this:

```python
active_regforms = (RegistrationForm.query
                   .join(Event)
                   .filter(RegistrationForm.ticket_template == self,
                           Event.ends_after(now_utc()))
                   .all())
```

I think the second one is slighly more performant, but considering that you already have a very small amount of rows due to the filter on ticket_template, I would go for the former since it's more concise.

---

In case you wonder why I'm against using find/find_all/first_first/find_one, even though it was me who added them initially: I thought it'd be nice and when using `indico shell` they are indeed very useful shortcuts. But when writing rael code it's better to stay close to what SQLAlchemy provides out of the box instead of obscure addons - and the various `_...` kwargs of find and friends are /very/ obscure.

indico/modules/designer/templates/_list.html
```diff
@@ -60,31 +58,55 @@
...
+                                title="{% trans %}This is the default ticket template from Home category{% endtrans %}"
+                            {% else %}
+                                title="{% trans %}Toggle ticket template for this category{% endtrans %}"
+                                data-href="{{ url_for('.toggle_category_default', target, template_id=tpl.id) }}"
+                                data-method="POST"
+                                data-update="#template-list"
+                            {% endif %}>
+                        </a>
+                    {% endif %}
+                        {% set clonable = tpl.is_clonable or target == tpl.category.get_root() %}
```
ThiefMaster: How about just checking `tpl.is_clonable or tpl.category.is_root`? No need to get the root category when you already have a category object.

indico/modules/designer/templates/_list.html
```diff
@@ -60,31 +58,55 @@
...
-                                data-href="{{ url_for('.delete_template', tpl) }}"
-                                data-method="DELETE"
-                                data-update="#template-list"
-                                data-confirm="{% trans title=tpl.title -%}
-                                                  Are you sure you want to delete template '{{ title }}'?
-                                              {%- endtrans %}">
+                        <button class="i-button icon-remove icon-only text-color borderless danger hide-if-locked
+                                {{ 'not-deletable' if not tpl.can_be_deleted }}"
+                                {% if not tpl.can_be_deleted %}
+                                    title="{% trans %}This template cannot be deleted because it's used by another registration form{% endtrans %}"
```
ThiefMaster: "it is" instead of "it's" (the shortcuts are very informal and not recommended in most written communication, chats etc. aside)

indico/modules/designer/templates/_list.html
```diff
@@ -60,31 +58,55 @@
...
+                                {% else %}
+                                    title="{% trans %}Delete tempalte{% endtrans %}"
```
ThiefMaster: typo -> template

indico/modules/designer/templates/_list.html
```diff
@@ -60,31 +58,55 @@
...
+                                    data-href="{{ url_for('.delete_template', tpl) }}"
+                                    data-method="DELETE"
+                                    data-update="#template-list"
+                                    data-confirm="{% trans title=tpl.title -%}
+                                                      Are you sure you want to delete template '{{ title }}'?
```
ThiefMaster: *the* template

indico/modules/designer/templates/_list.html
```diff
@@ -122,13 +144,13 @@
     </tr>
 {% endmacro -%}
 
-{% macro render_template_list(templates, target, inherited_templates=[], event=none) -%}
+{% macro render_template_list(templates, target, inherited_templates=[], event=none, default_template=None) -%}
```
ThiefMaster: lowercase `none`

indico/modules/designer/util.py
```diff
@@ -38,3 +40,11 @@ def get_all_templates(obj):
 def get_inherited_templates(obj):
     """Get all templates inherited by a given event/category"""
     return get_all_templates(obj) - set(obj.designer_templates)
+
+
+def get_default_template_on_category(category):
+    if category.default_ticket_template:
+        return category.default_ticket_template
+    parent_chain = reversed(category.parent_chain_query.options(joinedload('default_ticket_template')).all())
```
ThiefMaster: IMO it's more readable to do the `reversed()` below, i.e. `for category in reversed(parent_chain)`

indico/modules/events/registration/controllers/management/tickets.py
```diff
@@ -22,8 +22,13 @@
 from flask import flash, json, render_template
 from werkzeug.exceptions import Forbidden, NotFound
 
+from indico.core import signals
 from indico.core.config import Config
 from indico.core.db import db
+from indico.modules.designer import PageOrientation
+from indico.modules.designer import PageSize
```
ThiefMaster: merge with the previous line

indico/modules/events/registration/controllers/management/tickets.py
```diff
@@ -90,7 +102,14 @@ def _checkProtection(self):
 
     def _process(self):
         filename = secure_filename('{}-Ticket.pdf'.format(self.event_new.title), 'ticket.pdf')
-        return send_file(filename, generate_ticket(self.registration), 'application/pdf')
+        template = self.regform.ticket_template
+        if template is None:
+            template = get_default_template_on_category(self.event_new.category)
```
ThiefMaster: why not simply

    template = self.regform.ticket_template or get_default_template_on_category(self.event_new.category)

indico/modules/events/registration/controllers/management/tickets.py
```diff
@@ -90,7 +102,14 @@ def _checkProtection(self):
...
+        signals.event.designer.print_badge_template.send(template, regform=self.regform)
+        pdf_class = RegistrantsListToBadgesPDFFoldable if template.backside_template else RegistrantsListToBadgesPDF
+        config_params = DEFAULT_TICKET_PRINTING_SETTINGS
```
ThiefMaster: why assign it to a variable? you could just pass it directly in the next line

indico/modules/events/registration/forms.py
```diff
@@ -273,6 +273,20 @@ class TicketsForm(IndicoForm):
                                           description=_('Allow users to download their ticket from the registration '
                                                         'summary page.'))
 
+    ticket_template_id = SelectField(_('Ticket template'), [HiddenUnless('tickets_enabled', preserve_data=True),
+                                                            Optional()], coerce=lambda x: int(x))
```
ThiefMaster: `coerce=int`

indico/modules/events/registration/forms.py
```diff
@@ -273,6 +273,20 @@ class TicketsForm(IndicoForm):
...
+
+    def __init__(self, *args, **kwargs):
+        event = kwargs.pop('event')
+        default_tpl = get_default_template_on_category(event.category)
+        all_templates = set(event.designer_templates) | get_inherited_templates(event)
+        badge_templates = [(tpl.id, tpl.title) for tpl in all_templates
+                           if tpl.type == TemplateType.badge and tpl != default_tpl]
+        # Making the default category template the first option in the list of choices
+        badge_templates.insert(0, (default_tpl.id, default_tpl.title + ' ' + _('(Default category template)')))
```
ThiefMaster: `format()` instead of `+`

indico/modules/designer/templates/_list.html
DavidAndreev: But its to check whether the page we display this list is a root category page, thats the `target`object but it can also be an event,  or should I check `target.is_root` even though it can be an Event object ?

ThiefMaster: wrong logic! cloning should be possible if `target == tpl.owner` (or if the template is set to be always clonable)

indico/modules/designer/forms.py
```diff
@@ -30,4 +30,5 @@ class AddTemplateForm(IndicoForm):
     title = StringField(_('Title'), [DataRequired()])
     type = IndicoEnumSelectField(_('Template'), enum=TemplateType, default=TemplateType.poster)
     is_clonable = BooleanField(_('Allow cloning'), widget=SwitchWidget(), default=True,
-                               description=_("Allow event managers to clone this template."))
+                               description=_("Allow event managers to clone this template in children "
```
ThiefMaster: "child", i don't think the plural fits here.

indico/modules/designer/util.py
```diff
@@ -42,6 +45,21 @@ def get_inherited_templates(obj):
     return get_all_templates(obj) - set(obj.designer_templates)
 
 
+def get_not_deletable_templates(obj):
+    """Get all not deletable templates usable by an event/category"""
```
ThiefMaster: "Get all non-deletable ..."

indico/modules/designer/util.py
```diff
@@ -42,6 +45,21 @@ def get_inherited_templates(obj):
...
+    not_deletable_criteria = [
+        DesignerTemplate.is_system_template,
+        DesignerTemplate.backside_template_of != None,
+        DesignerTemplate.ticket_for_regforms.any(RegistrationForm.event_new.has(Event.ends_after(now_utc())))
+    ]
+    not_deletable = set(DesignerTemplate.query
+                        .filter(DesignerTemplate.event_new == obj if isinstance(obj, Event) else
+                                DesignerTemplate.category_id.in_(categ['id'] for categ in obj.chain))
+                        .filter(db.or_(*not_deletable_criteria)))
+    return not_deletable
```
ThiefMaster: why not `return set(...)` directly instead of assigning it first just to return it?

indico/modules/designer/util.py
```diff
@@ -38,3 +43,24 @@ def get_all_templates(obj):
...
+
+def get_not_deletable_templates(obj):
+    """Get all non-deletable templates usable by an event/category"""
+
+    not_deletable_criteria = [
+        DesignerTemplate.is_system_template,
+        DesignerTemplate.backside_template_of != None,
+        DesignerTemplate.ticket_for_regforms.any(RegistrationForm.event_new.has(Event.ends_after(now_utc())))
+    ]
+    return set(DesignerTemplate.query.filter(DesignerTemplate.event_new == obj if isinstance(obj, Event) else
```
ThiefMaster: Wrong indentation style.
```
return set(DesignerTemplate.query
           .filter(...)
           .whatever(...))
```

Also, I would really split this and move the statement with the isinstance check out of the query and assign it to a variable first, and then do e.g. `.filter(obj_criterion)`

pferreir:
When I first upgrade the DB and try to open the ticket configuration dialog...

```
  File "/Users/pedro/code/indico/indico/legacy/webinterface/rh/base.py", line 493, in process
    profile_name, res = self._do_process(profile)
  File "/Users/pedro/code/indico/indico/legacy/webinterface/rh/base.py", line 464, in _do_process
    res = self._process()
  File "/Users/pedro/code/indico/indico/modules/events/registration/controllers/management/tickets.py", line 75, in _process
    form = TicketsForm(obj=self.regform, event=self.event_new)
  File "/Users/pedro/code/indico/indico/web/forms/base.py", line 70, in __call__
    return super(IndicoFormMeta, cls).__call__(*args, **kwargs)
  File "/Users/pedro/.virtualenvs/indico/lib/python2.7/site-packages/wtforms/form.py", line 212, in __call__
    return type.__call__(cls, *args, **kwargs)
  File "/Users/pedro/code/indico/indico/modules/events/registration/forms.py", line 286, in __init__
    badge_templates.insert(0, (default_tpl.id, '{} {}'.format(default_tpl.title, _('(Default category template)'))))
AttributeError: 'NoneType' object has no attribute 'id'
```
We're missing a default template that gets added and set as default in the Home category.

ThiefMaster:
I guess there should be an alembic revision creating the default one which we also create in a new instance.

indico/modules/designer/controllers.py
```diff
@@ -327,3 +348,18 @@ def _process(self):
             'background_url': self.template.background_image.download_url if self.template.background_image else None
         }
         return jsonify(template=template_data, backside_template_id=self.template.id)
+
+
+class RHToggleTemplateDefaultOnCategory(RHManageCategoryBase):
+
```
ThiefMaster: there shouldn't be an empty line here.

ThiefMaster:
Actually, from what I can see this is also missing so far. The default ticket template should be created in `create_all_tables` where we also create the root category etc.

indico/modules/events/registration/notifications.py
```diff
@@ -33,11 +36,21 @@ def notify_invitation(invitation, email_subject, email_body, from_address):
 
 
 def _notify_registration(registration, template, to_managers=False):
+    attachments = None
+    regform = registration.registration_form
+    tickets_handled = values_from_signal(signals.event.is_ticketing_handled.send(regform), single_value=True)
+    if (not to_managers and
+            regform.tickets_enabled and
+            regform.ticket_on_email and
+            not (tickets_handled and all(tickets_handled)) and
```
ThiefMaster: @meluru I think `any(tickets_handled)`  would be more appropriate here. In the rare case that multiple plugins handle tickets (do we even support this) it should be enough if one of these plugins intercept ticketing..

indico/core/signals/event/registration.py
```diff
@@ -47,6 +47,14 @@
...
 
 registration_form_deleted = _signals.signal('registration-form-deleted', """
-Called when a registration form is removed.  The `sender` is the
+Called when a registration form is removed. The `sender` is the
 `RegistrationForm` object.
 """)
+
+is_ticketing_handled = _signals.signal('is-ticketing-handled', """
+Called when resolving whether Indico should send tickets with e-mails
+or it will be handled by other module. The `sender` is the
```
ThiefMaster: > or *if* it will be handled by *another* module. 

@meluru

indico/core/db/sqlalchemy/util/management.py
```diff
@@ -68,6 +100,8 @@ def create_all_tables(db, verbose=False, add_initial_data=True):
     """Create all tables and required initial objects"""
     from indico.modules.categories import Category
     from indico.modules.users import User
+    from indico.modules.designer import TemplateType
```
ThiefMaster: c < d < u

indico/core/db/sqlalchemy/util/management.py
```diff
@@ -78,4 +112,8 @@ def create_all_tables(db, verbose=False, add_initial_data=True):
         if verbose:
             print cformat('%{green}Creating root category')
         db.session.add(Category(id=0, title='Home', protection_mode=ProtectionMode.public))
+        if verbose:
+            print cformat('%{green}Creating default ticket template for root category ')
+        db.session.add(DesignerTemplate(category_id=0, title='Home category default ticket template',
```
ThiefMaster: 'Default Ticket' or something like that.

indico/modules/events/requests/controllers.py
```diff
@@ -182,6 +189,11 @@ def _checkProtection(self):
     def _process(self):
         if self.request.state not in {RequestState.pending, RequestState.accepted}:
             raise BadRequest
-        self.definition.withdraw(self.request)
-        flash(_('You have withdrawn your request ({0})').format(self.definition.title))
+        try:
+            self.definition.withdraw(self.request)
+        except RequestModuleError:
+            message = 'There was a problem withdrawing your request ({0}). Please contact support for further details.'
+            flash(_(message).format(self.definition.title), 'error')
```
ThiefMaster: @meluru this doesn't work. the `_()` must be right around the string  so it can be extracted for translation

if you did it due to the line length:

    msg = _('...')

should make enough space available ;)

indico/modules/events/requests/exceptions.py
```diff
@@ -0,0 +1,23 @@
...
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+
+class RequestModuleError(Exception):
+    def __init__(self, message=None):
```
ThiefMaster: why override this? Exception already takes an argument and you can get the string using `str(exc)`:

```
>>> exc = Exception('foo')
>>> str(exc)
'foo'
```

indico/modules/events/requests/controllers.py
```diff
@@ -120,13 +122,18 @@ def process_form(self):
...
             req = self.request
             new = False
-        self.definition.send(req, form.data)
-        if new:
-            flash_msg = _("Your request ({0}) has been sent. "
-                          "You will be notified by e-mail once it has been accepted or rejected.")
+        try:
+            self.definition.send(req, form.data)
+        except RequestModuleError:
+            db.session.delete(req)
```
ThiefMaster: @meluru is just deleting the whole request really a good idea?

indico/modules/events/registration/notifications.py
```diff
@@ -42,7 +42,7 @@ def _notify_registration(registration, template, to_managers=False):
     if (not to_managers and
             regform.tickets_enabled and
             regform.ticket_on_email and
-            not (tickets_handled and all(tickets_handled)) and
+            not (tickets_handled and any(tickets_handled)) and
```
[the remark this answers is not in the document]
ThiefMaster: oh heh, totally forgot to submit that comment last week or whenever i wrote it..

PR #3043, opened 2017-08-17
Cross-instance event import/export

ThiefMaster:


pferreir:
Would it maybe be interesting to have a verbose mode, which is especially useful for larger events?

indico/testing/fixtures/storage.py
```diff
@@ -0,0 +1,70 @@
...
+from indico.core import signals
+from indico.core.config import Config
+from indico.core.storage.backend import Storage
+from indico.modules.attachments.models.attachments import Attachment, AttachmentFile, AttachmentType
+from indico.modules.attachments.models.folders import AttachmentFolder
+
+
+@signals.get_storage_backends.connect
+def _get_storage_backends(sender, **kwargs):
+    yield MemoryStorage
```
ThiefMaster: can be `return` for a single item

indico/testing/fixtures/storage.py
```diff
@@ -0,0 +1,70 @@
...
+            'test': 'mem://test'
+        },
+        AttachmentStorage='test'
+    )
+
+
+@pytest.yield_fixture
+def dummy_attachment(dummy_user):
+    folder = AttachmentFolder(title='dummy_folder', description='a dummy folder')
+    file_ = AttachmentFile(user=dummy_user, filename='dummy_file.txt', content_type='text/plain')
```
ThiefMaster: I would just shadow `file` here. It's one of the most useless stdlib globals anyway (and was even removed from python 3)

indico/testing/fixtures/storage.py
```diff
@@ -0,0 +1,70 @@
...
+def storage():
+    Config.getInstance().update(
+        StorageBackends={
+            'test': 'mem://test'
+        },
+        AttachmentStorage='test'
+    )
+
+
+@pytest.yield_fixture
```
ThiefMaster: pytest recommends to use just `@pytest.fixture` since the yield logic has been included there: https://docs.pytest.org/en/latest/yieldfixture.html

indico/cli/event.py
```diff
@@ -0,0 +1,88 @@
...
+@click.option('--force', is_flag=True, help='Ignore Indico version mismatches (DANGER)')
+@click.option('-c', '--category', 'category_id', type=int, default=0, metavar='ID',
+              help='ID of the target category. Defaults to the root category.')
+def import_(source_file, force, category_id):
+    """Imports an event exported from another Indico instance."""
+    event = import_event(source_file, category_id, force=force)
+    if event is None:
+        click.secho('Import failed.', fg='red')
+        sys.exit(1)
+    if not click.confirm('Import finished. Commit the changes?', default=True):
```
ThiefMaster: yeah, and also a switch to set the default for user creation

ThiefMaster:
Thanks, @GitHub for not updating PRs when pushing to the base branch and now showing garbage commits/diffs here.

PR #3045, opened 2017-08-21
Fix Indico "Welcome" page UI bug

DavidAndreev:


DavidAndreev:
Fixed the height for firefox and made "additional" text bit darker : 
new

PR #3047, opened 2017-08-23
Add MD5 checksum to all stored files

pferreir:
Calculated upon saving.

### DIFF
--- a/indico/core/storage/backend.py
+++ b/indico/core/storage/backend.py
@@ -19,8 +19,8 @@ from __future__ import unicode_literals
 import os
 import sys
 from contextlib import contextmanager
+from hashlib import md5
 from io import BytesIO
-from shutil import copyfileobj
 from tempfile import NamedTemporaryFile
 
 from werkzeug.security import safe_join
@@ -106,6 +106,20 @@ class Storage(object):
         """Ensures that fileobj is a file-like object and not a string"""
         return BytesIO(fileobj) if not hasattr(fileobj, 'read') else fileobj
 
+    def _copy_file(self, source, target, chunk_size=1024*1024):
+        """Copy a file, in chunks, from ``source`` to ``target``.
+
+        The return value will be the MD5 checksum of the file (hex).
+        """
+        checksum = md5()
+        while True:
+            chunk = source.read(chunk_size)
+            if not chunk:
+                break
+            target.write(chunk)
+            checksum.update(chunk)
+        return checksum.hexdigest().decode('ascii')
+
     def open(self, file_id):  # pragma: no cover
         """Opens a file in the storage for reading.
 
@@ -129,7 +143,7 @@ class Storage(object):
         """
         with self.open(file_id) as fd:
             with NamedTemporaryFile(suffix='indico.tmp', dir=Config.getInstance().getTempDir()) as tmpfile:
-                copyfileobj(fd, tmpfile, 1024 * 1024)
+                self._copy_file(fd, tmpfile)
                 tmpfile.flush()
                 yield tmpfile.name
 
@@ -157,7 +171,8 @@ class Storage(object):
                          not be used depending on the backend).
         :param fileobj: A file-like object containing the file data as
                         bytes or a bytestring.
-        :return: unicode -- A unique identifier for the file.
+        :return: (unicode, unicode) -- A tuple containing a unique
+                        identifier for the file and an MD5 checksum.
         """
         raise NotImplementedError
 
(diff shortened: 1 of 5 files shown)


indico/core/storage/backend_test.py
```diff
@@ -102,15 +102,21 @@ def test_fs_dir(fs_storage):
...
 def test_fs_operations(fs_storage):
-    f1 = fs_storage.save('foo/bar/test.txt', 'unused/unused', 'unused', b'hello world')
-    f2 = fs_storage.save('foo/bar/test2.txt', 'unused/unused', 'unused', b'hello there')
-    f3 = fs_storage.save('test.txt', 'unused/unused', 'unused', b'hello test')
+    f1, h1 = fs_storage.save('foo/bar/test.txt', 'unused/unused', 'unused', b'hello world')
+    f2, h2 = fs_storage.save('foo/bar/test2.txt', 'unused/unused', 'unused', b'hello there')
+    f3, h3 = fs_storage.save('test.txt', 'unused/unused', 'unused', b'very very long file' * 1024 * 1024)
+
+    # check md5 checksums
+    assert h1 == b'5eb63bbbe01eeed093cb22bb8f5acdc3'
```
ThiefMaster: I would rather return the hashes as unicode. They're also unicode when stored in the DB.

indico/core/storage/models.py
```diff
@@ -136,6 +136,19 @@ def size(cls):
...
 
     @declared_attr
+    def md5(cls):
+        """
+        An MD5 hash of the file.
+
+        Automatically assigned when `save()` is called.
+        """
+        return db.Column(
+            db.String(32),
```
ThiefMaster: I'd just go for `db.String` since setting a length limit on a postgres varchar is pretty pointless

pferreir: I actually stopped to think about it and was wondering if it makes any sense given that it's always 32 char long.

ThiefMaster: https://stackoverflow.com/questions/4848964/postgresql-difference-between-text-and-varchar-character-varying is pretty interesting about postgres text types

indico/core/storage/backend.py
```diff
@@ -106,6 +106,20 @@ def _ensure_fileobj(self, fileobj):
         """Ensures that fileobj is a file-like object and not a string"""
         return BytesIO(fileobj) if not hasattr(fileobj, 'read') else fileobj
 
+    def _copy_file(self, source, target, chunk_size=1024 * 1024):
```
ThiefMaster: I think that's one of the few places where I would omit the space around `*` - having spaces in the default value of a kwarg always looks pretty odd to me, especially without `()` around it...

indico/core/storage/models.py
```diff
@@ -136,6 +136,19 @@ def size(cls):
...
+    def md5(cls):
+        """
+        An MD5 hash of the file.
+
+        Automatically assigned when `save()` is called.
+        """
+        return db.Column(
+            db.String,
+            nullable=False,
+            index=True
```
ThiefMaster: Do we ever want to search by hash? Otherwise I wouldn't bother adding an index to the column

indico/migrations/versions/201708231635_dc2c2fa6f5be_add_file_md5_checksum.py
```diff
@@ -0,0 +1,60 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'dc2c2fa6f5be'
+down_revision = 'ebca3e91b139'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('files', sa.Column('md5', sa.String, nullable=False, server_default='', index=True),
```
ThiefMaster: does this actually work? AFAIK `index=True` is a no-op if the table already exists (or at least i think it used to be the case)

indico/core/storage/backend.py
pferreir: I'm pretty agnostic in this, so I'll follow your advice.

indico/core/storage/models.py
pferreir: Well, not for now.

indico/migrations/versions/201708231635_dc2c2fa6f5be_add_file_md5_checksum.py
pferreir: If we're getting rid of the hash we don't need an index anyway.

pferreir: Anyway, it **does not** work, as you've guessed.

PR #3049, opened 2017-08-29
Update messages.po

rouet:
fix typo : missed space character

regards


ThiefMaster:
Thanks, I've updated it on Transifex and we'll include it the next time we sync the i18n strings.

Issue #2648, opened 2016-10-16
Tons of pycountry.db errors in other.log


danmichaelo:
I get tons of these in my `other.log`. Should I worry or not..? 

Fresh install of Indico v1.2.1rc10 installed using easy_install on Centos 7.2

```
2016-10-16 22:34:04,518  ERROR    000000000000  pycountry.db              Country False already taken in index 'deleted' and will be ignored.
2016-10-16 22:34:04,518  ERROR    000000000000  pycountry.db              Country False already taken in index 'deleted' and will be ignored.
2016-10-16 22:34:04,518  ERROR    000000000000  pycountry.db              Country False already taken in index 'deleted' and will be ignored.
2016-10-16 22:34:04,518  ERROR    000000000000  pycountry.db              Country u'BQ' already taken in index 'alpha2' and will be ignored.
2016-10-16 22:34:04,518  ERROR    000000000000  pycountry.db              Country True already taken in index 'deleted' and will be ignored.
2016-10-16 22:34:04,518  ERROR    000000000000  pycountry.db              Country True already taken in index 'deleted' and will be ignored.
2016-10-16 22:34:04,518  ERROR    000000000000  pycountry.db              Country u'BY' already taken in index 'alpha2' and will be ignored.
2016-10-16 22:34:04,519  ERROR    000000000000  pycountry.db              Country True already taken in index 'deleted' and will be ignored.
2016-10-16 22:34:04,519  ERROR    000000000000  pycountry.db              Country True already taken in index 'deleted' and will be ignored.
2016-10-16 22:34:04,519  ERROR    000000000000  pycountry.db              Country True already taken in index 'deleted' and will be ignored.
2016-10-16 22:34:04,519  ERROR    000000000000  pycountry.db              Country True already taken in index 'deleted' and will be ignored.
2016-10-16 22:34:04,519  ERROR    000000000000  pycountry.db              Country True already taken in index 'deleted' and will be ignored.
2016-10-16 22:34:04,519  ERROR    000000000000  pycountry.db              Country True already taken in index 'deleted' and will be ignored.
2016-10-16 22:34:04,519  ERROR    000000000000  pycountry.db              Country u'ET' already taken in index 'alpha2' and will be ignored.
2016-10-16 22:34:04,519  ERROR    000000000000  pycountry.db              Country True already taken in index 'deleted' and will be ignored.
[...]

PR #3052, opened 2017-08-30
Upgrade to a recent pycountry

bpedersen2:
Upgrade the pycountry dependency to the most recent stable
release. Adopt to changed attribute names.

Fixes: #2648

### DIFF
--- a/indico/util/countries.py
+++ b/indico/util/countries.py
@@ -25,7 +25,7 @@ from indico.util.caching import memoize
 
 @memoize
 def get_countries():
-    _countries = {country.alpha2: country.name for country in pycountry.countries}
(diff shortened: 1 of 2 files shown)


ThiefMaster:
We'd love to do that, unfortunately this is kind of a blocker: https://bitbucket.org/flyingcircus/pycountry/issues/13397/iso3166-3-is-missing-alpha_2-codes

They don't want to do this kind of change in pycountry, and their upstream is extremely inactive... (opened the issue in May, still no response).

ThiefMaster:
[Finally!](https://anonscm.debian.org/git/pkg-isocodes/iso-codes.git/commit/?id=d4e76a234a9d244fa695f22c7e9a977309c88449)

So we can upgrade soon (once there's a pycountry release including this change).

PR #3053, opened 2017-08-30
Improve handling of system OAuth apps

ThiefMaster:
- created automatically
- not deletable
- certain fields locked to the right values
- no need to set their id in the config file

if the apps already exist, their ids (either db id or client id) can be set via x-args, e.g.:

    indico db upgrade -x checkin_app_id=00000000-dead-c0de-beef-000000001111 -x flower_app_id=1

### DIFF
--- a/indico/core/celery/cli.py
+++ b/indico/core/celery/cli.py
@@ -21,9 +21,9 @@ import sys
 
 from celery.bin.celery import CeleryCommand, command_classes
 
-from indico.core.config import Config
 from indico.core.celery import celery
-from indico.modules.oauth.models.applications import OAuthApplication
+from indico.core.config import Config
+from indico.modules.oauth.models.applications import OAuthApplication, SystemAppType
 from indico.util.console import cformat
 from indico.web.flask.util import url_for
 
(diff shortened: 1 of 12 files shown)


indico/core/celery/cli.py
```diff
@@ -43,30 +43,20 @@ def celery_cmd(args):
...
-                        'INDICO_FLOWER_USER_URL': url_for('users.authenticated_user', _external=True)}
-        else:
-            print cformat('%{red!}WARNING: %{yellow!}Flower authentication is disabled.')
-            print cformat('%{yellow!}Having access to Flower allows one to shutdown Celery workers.')
-            print
-            auth_args = []
-            auth_env = {}
+        app = OAuthApplication.find_one(system_app_type=SystemAppType.flower)
+        if not app.redirect_uris:
+            print cformat('%{yellow!}Authentication fill fail unless you configure the redirect url for the {} OAuth '
```
pferreir: "Authentication **will** fail"

indico/modules/oauth/models/applications.py
```diff
@@ -34,6 +36,38 @@
...
 
 
+class SystemAppType(int, IndicoEnum):
+    none = 0
+    checkin = 1
+    flower = 2
+
+    @property
+    def enforced_data(self):
+        if self == SystemAppType.none:
```
pferreir: Ugh. No better way of doing this? 😵

indico/modules/oauth/models/applications.py
```diff
@@ -34,6 +36,38 @@
...
+                    'is_enabled': True}
+        elif self == SystemAppType.flower:
+            return {'default_scopes': {'read:user'},
+                    'is_enabled': True}
+
+    @property
+    def default_data(self):
+        data = self.enforced_data
+        data['is_trusted'] = True
+        if self == SystemAppType.checkin:
```
pferreir: Same here.

PR #3057, opened 2017-09-06
Add backend support for event groups/roles

ThiefMaster:


ThiefMaster:
@pferreir: Should we rename groups to roles and EventPrincipal roles to permissions? IMO it wouldn't be such a bad idea to call it "roles" in user-facing areas but keep calling it "event groups" internally (since that's pretty much what it is - a group that's local to the event) and keep "roles" in the code as it is right now (i.e. indicating what permissions an acl entry has)

indico/migrations/versions/20170905_1521_640584a3987e_add_event_groups_to_acls.py
```diff
@@ -0,0 +1,200 @@
...
+                          ['event_group_id'], ['id'],
+                          source_schema='events', referent_schema='events')
+    op.create_foreign_key(None, 'session_principals', 'groups',
+                          ['event_group_id'], ['id'],
+                          source_schema='events', referent_schema='events')
+    op.create_foreign_key(None, 'settings_principals', 'groups',
+                          ['event_group_id'], ['id'],
+                          source_schema='events', referent_schema='events')
+    op.execute('''
+        ALTER TABLE attachments.attachment_principals DROP CONSTRAINT ck_attachment_principals_valid_enum_type;
```
ThiefMaster: Using raw SQL here is much less ugly than writing alembic/sqlalchemy statements for it, especially since the CHECK constraint updates are not autogenerated (so like this the dbdiff output can be copied almost as-is, only thing I changed was removing the `SET search_path` stuff in favor of explicit schemas in table names)

PR #3068, opened 2017-09-15
Add event role management

ThiefMaster:


indico/modules/events/roles/templates/roles.html
```diff
@@ -0,0 +1,103 @@
...
+        <button class="i-button highlight icon-plus hide-if-locked"
+                data-href="{{ url_for('.add_role', event) }}"
+                data-title="{% trans %}Add a new role{% endtrans %}"
+                data-update="#event-roles"
+                data-ajax-dialog>
+            {% trans %}Add Role{% endtrans %}
+        </button>
+    </div>
+
+    {# TODO: move to JS file #}
```
pferreir: I guess this is the `roles.js` file that @meluru is creating in her branch?

ThiefMaster: Probably - initially we didn't add the separate CSS/JS file to avoid rebase hell in bundles.py with the changes from my config refactoring branch.

indico/modules/events/roles/controllers.py
```diff
@@ -0,0 +1,147 @@
...
+from indico.modules.events.roles.forms import RoleForm
+from indico.modules.events.roles.views import WPEventRoles
+from indico.modules.users import User
+from indico.util.user import principal_from_fossil
+from indico.web.flask.templating import get_template_module
+from indico.web.util import jsonify_data, jsonify_form
+
+
+def _get_roles(event):
+    return(EventRole.query.with_parent(event)
```
mvidalgarcia: Space missing?

ThiefMaster: yes

indico/htdocs/js/indico/modules/roles.js
```diff
@@ -0,0 +1,67 @@
...
+    $(document).ready(function() {
+        setupToggle();
+        setupButtons();
+    });
+
+    function setupToggle() {
+        var $roles = $('#event-roles');
+        $roles.on('click', '.toggle-members', function() {
+            var $row = $(this).closest('tr').next('tr').find('.slide');
+            $row[0].style.maxHeight = $row[0].scrollHeight + "px";
```
ThiefMaster: `$row.css('max-height', ...)` doesn't work?

indico/htdocs/js/indico/modules/roles.js
```diff
@@ -0,0 +1,67 @@
...
+            $row.toggleClass('open close');
+        });
+
+        $roles.on('indico:htmlUpdated', function() {
+            var $slide = $(this).find('.slide');
```
ThiefMaster: assigning this before calling `.each()` doesn't seem very useful since you don't use it again later

indico/htdocs/js/indico/modules/roles.js
```diff
@@ -0,0 +1,67 @@
...
+            $slide.each(function() {
+                $(this)[0].style.maxHeight =  $(this)[0].scrollHeight + "px";
```
ThiefMaster: `$(this)[0] === this` so no need to wrap in jquery just to unwrap it again

indico/htdocs/sass/modules/_roles.scss
```diff
@@ -0,0 +1,90 @@
...
+    padding: 10px;
+}
+
+.role-list {
+    padding: 10px;
+    box-sizing: border-box;
+    list-style-type: none;
+}
+
+.role-list li {
```
ThiefMaster: since this is SASS now, it can be improved (nesting instaedof repeating  selectors all the time)

indico/modules/events/roles/templates/_roles.html
```diff
@@ -77,8 +73,17 @@
                             </li>
                         {% endfor %}
                     </ul>
-                </div>
-            </td>
-        </tr>
-    {% endif %}
+                {% else %}
+                    <div class="empty">There are no members yet.</div>
```
ThiefMaster: i18n

indico/modules/events/roles/templates/_roles.html
```diff
@@ -77,8 +73,17 @@
...
+                {% endif %}
+                    <div class="toolbar f-j-end">
+                         <button class="i-button highlight text-color icon-plus js-add-members"
+                                data-href="{{ url_for('.add_members', role) }}"
```
ThiefMaster: indentation

indico/web/assets/bundles.py
```diff
@@ -437,7 +437,8 @@ def register_all_js(env):
         'sessions': rjs_bundle('modules_sessions', 'js/indico/modules/sessions.js'),
         'users': rjs_bundle('modules_users', 'js/indico/modules/users.js'),
         'designer': rjs_bundle('modules_designer', 'js/indico/modules/designer.js'),
-        'event_cloning': rjs_bundle('modules_event_cloning', 'js/indico/modules/events/cloning.js')
+        'event_cloning': rjs_bundle('modules_event_cloning', 'js/indico/modules/events/cloning.js'),
+        'roles': rjs_bundle('modules_roles', 'js/indico/modules/roles.js')
```
ThiefMaster: `event_roles` and it should be inside `modules/events/` and not just `modules/`

indico/htdocs/sass/modules/_roles.scss
```diff
@@ -15,55 +15,55 @@
...
+            }
+        }
+    }
+
+    .i-tag {
+        display: inline-flex;
+        justify-content: center;
+        align-items: baseline;
+
+        :not(:first-child) {
```
ThiefMaster: `&:not(..) {` - otherwise it's matching a child element with that selector and not the element itself

indico/web/assets/bundles.py
```diff
@@ -14,6 +14,8 @@
 # You should have received a copy of the GNU General Public License
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import unicode_literals
```
ThiefMaster: Usually not a good idea to add it to existing files - it probably works here but there are other (legacy) files where this  would break things (I actually removed this from the isort config since it's more annoying than useful for already-existing files)

indico/htdocs/sass/partials/_labels.scss
```diff
@@ -42,6 +42,34 @@ $i-label-color: $light-black;
     }
 }
 
+@mixin _i-label-outlines($color, $dark-color) {
```
mvidalgarcia: `$dark-color` is not being used in the mixin.

mvidalgarcia:
Maybe also a constraint to force the role code to be <= 3 characters?

ThiefMaster:
I was thinking about it but we don't enforce length limits anywhere else. FWIW, I only added the uppercase constraint to the database so I can use a simple unique index without having to go for a more complex one (`event_id, upper(code)`) ;)

PR #3069, opened 2017-09-15
Track-session linking

mvidalgarcia:
Temporary UI to show the default session linked to track:
image

### DIFF
--- a/indico/htdocs/js/indico/modules/abstracts.js
+++ b/indico/htdocs/js/indico/modules/abstracts.js
@@ -135,4 +135,15 @@
             applySearchFilters();
         }
     };
+
+    global.setupAbstractJudgment = function setupAbstractJudgment(options) {
+        options = $.extend({
+            trackSessionMap: {}
+        }, options);
+
+        $('#accepted_track').on('change', function() {
+            var sessionId = options.trackSessionMap[$(this).val()];
+            $('#session').val(sessionId || "__None");
+        }).trigger('change');
+    };
 })(window);
--- a/indico/htdocs/js/indico/modules/abstracts.js
+++ b/indico/htdocs/js/indico/modules/abstracts.js
@@ -141,9 +141,10 @@
             trackSessionMap: {}
         }, options);
 
-        $('#accepted_track').on('change', function() {
+        $('body').on('change', '#accepted_track', function() {
             var sessionId = options.trackSessionMap[$(this).val()];
             $('#session').val(sessionId || "__None");
-        }).trigger('change');
+        });
+        $('#accepted_track').trigger('change');
     };
 })(window);
--- a/indico/modules/events/abstracts/controllers/abstract_list.py
+++ b/indico/modules/events/abstracts/controllers/abstract_list.py
@@ -95,6 +95,10 @@ class RHAbstractList(DisplayAbstractListMixin, RHAbstractListBase):
     template = 'management/abstract_list.html'
     view_class = WPManageAbstracts
 
+    def _render_template(self, **kwargs):
+        kwargs['track_session_map'] = {track.id: track.default_session_id for track in self.event.tracks}
+        return super(RHAbstractList, self)._render_template(**kwargs)
+
 
 class RHAbstractListCustomize(CustomizeAbstractListMixin, RHAbstractListBase):
     view_class = WPManageAbstracts
--- a/indico/modules/events/tracks/forms.py
+++ b/indico/modules/events/tracks/forms.py
@@ -16,10 +16,12 @@
 
 from __future__ import unicode_literals
 
+from wtforms.ext.sqlalchemy.fields import QuerySelectField
 from wtforms.fields import StringField
 from wtforms.validators import DataRequired
 
 from indico.core.db.sqlalchemy.descriptions import RenderMode
+from indico.modules.events.sessions.models.sessions import Session
 from indico.util.i18n import _
 from indico.web.forms.base import IndicoForm, generated_data
 from indico.web.forms.fields import IndicoMarkdownField
@@ -28,7 +30,15 @@ from indico.web.forms.fields import IndicoMarkdownField
 class TrackForm(IndicoForm):
     title = StringField(_('Title'), [DataRequired()])
     code = StringField(_('Code'))
+    default_session = QuerySelectField(_('Default session'), default='', allow_blank=True, get_label='title',
+                                       description=_('Indico will preselect this session whenever an abstract is '
+                                                     'accepted for the track'))
     description = IndicoMarkdownField(_('Description'), editor=True)
+
+    def __init__(self, *args, **kwargs):
+        event = kwargs.pop('event')
+        super(TrackForm, self).__init__(*args, **kwargs)
+        self.default_session.query = Session.query.with_parent(event)
 
 
 class ProgramForm(IndicoForm):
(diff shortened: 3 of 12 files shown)


indico/htdocs/js/indico/modules/abstracts.js
```diff
@@ -135,4 +135,15 @@
             applySearchFilters();
         }
     };
+
+    global.setupAbstractJudgment = function setupAbstractJudgment(options) {
+        options = $.extend({
+            mapTrackSession: {}
```
ThiefMaster: `trackSessionMap`?

indico/modules/events/tracks/forms.py
```diff
@@ -28,8 +32,16 @@
 class TrackForm(IndicoForm):
     title = StringField(_('Title'), [DataRequired()])
     code = StringField(_('Code'))
+    default_session = QuerySelectField(_('Default session'), default='', allow_blank=True,
+                                       get_label=attrgetter('title'),
```
ThiefMaster: AFAIK`get_label='title'` does the job too

indico/modules/events/tracks/models/tracks.py
```diff
@@ -59,6 +59,11 @@ class Track(DescriptionMixin, db.Model):
         nullable=False,
         default=_get_next_position
     )
+    default_session_id = db.Column(
+        db.Integer,
+        db.ForeignKey('events.sessions.id'),
+        index=True
```
ThiefMaster: I'd use an explicit `nullable=True`

indico/modules/events/tracks/models/tracks.py
```diff
@@ -92,6 +97,10 @@ class Track(DescriptionMixin, db.Model):
             lazy=True
         )
     )
+    default_session = db.relationship(
+        'Session',
+        lazy=True
```
ThiefMaster: missing backref

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -14,31 +14,45 @@
...
-                        <span class="i-box-description">{{ track.code }}</span>
+                        <span class="i-label small">
+                            {{ track.code }}
+                        </span>
+                        {% set session = track.default_session %}
+                        {% if session %}
+                            <span class="i-label small default-session"
+                                  title='{% trans title=session.title %}Indico will preselect the session "{{ title }}"
+                                         whenever an abstract is accepted for this track{% endtrans %}'
+                                  style="background-color: #{{ session.background_color }};
```
ThiefMaster: `style="{{ session.css }}"`

indico/modules/events/tracks/models/tracks.py
mvidalgarcia: @pferreir mentioned it might not be needed, what do you think?

ThiefMaster: omitting the backref is usually not a good idea - AFAIK some cascades only work when there's a backref

mvidalgarcia:
```python
default_session = db.relationship(
    'Session',
    lazy=True,
    backref=db.backref(
        'assigned_tracks',
        lazy=True
    )
)
```
Something like this?

ThiefMaster: Yes, but I'd call it `default_for_tracks` and you could just use `backref='default_for_tracks'` when using all the defaults anyway (which is the case here)

indico/modules/events/tracks/forms.py
```diff
@@ -28,8 +30,15 @@
 class TrackForm(IndicoForm):
     title = StringField(_('Title'), [DataRequired()])
     code = StringField(_('Code'))
+    default_session = QuerySelectField(_('Default session'), default='', allow_blank=True, get_label='title',
+                                       description=_('Indico will preselect this session whenever an abstract is '
+                                                     'accepted for the track'))
     description = IndicoMarkdownField(_('Description'), editor=True)
 
+    def __init__(self, event, *args, **kwargs):
```
ThiefMaster: in all other forms requiring an event we pass it as a kwarg and pop it from `kwargs` before the super call so I'd do the same here

mvidalgarcia:
What do you mean exactly? The pre-select logic is in the last commit and I've modified it to trigger the event when loading the page for the first time.

pferreir:
Sorry, I started the comment but then forgot to delete it. Of course you've already done it.

ThiefMaster:
Shouldn't this also work with bulk abstract judgment?

indico/modules/events/abstracts/controllers/abstract_list.py
```diff
@@ -95,6 +95,13 @@ class RHAbstractList(DisplayAbstractListMixin, RHAbstractListBase):
     template = 'management/abstract_list.html'
     view_class = WPManageAbstracts
 
+    def _process(self):
+        if self.list_generator.static_link_used:
+            return redirect(self.list_generator.get_list_url())
+        kwargs = self.list_generator.get_list_kwargs()
+        kwargs['track_session_map'] = {track.id: track.default_session_id for track in self.event.tracks}
```
mvidalgarcia: Not sure if this is the best way of passing `track_session_map` to the template, I just didn't want to add it in `DisplayAbstractListMixin` since it's probably being used by other classes.

ThiefMaster: Yes, that's fine. We do the same e.g. in `PaperAssignmentListGenerator`

ThiefMaster: Actually, looking at it again, I'd rather override `get_list_kwargs` and do it in there (that's what we do in the class I mentioned above)

ThiefMaster: Why not override `get_list_kwargs` in `AbstractListGeneratorManagement`?

ThiefMaster: Or hm, it's not really related to the list generator. So maybe adding it there wouldn't be the best choice.

Why not override `_render_template` and add it to the kwargs in there, and then call super?

mvidalgarcia: Good point, updated.

mvidalgarcia: Whoops I missed your 2nd comment

mvidalgarcia: Updated

indico/modules/events/tracks/templates/_track_list.html
```diff
@@ -14,31 +14,44 @@
...
+                                           title="{% trans %}This track cannot be deleted because it contains accepted abstracts{% endtrans %}"></a>
+                                    {% endif %}
+                                </div>
                             </div>
                         </div>
-                        <span class="i-box-description">{{ track.code }}</span>
+                        <span class="i-label small">
+                            {{ track.code }}
+                        </span>
+                        {% set session = track.default_session %}
```
ThiefMaster: that's overwriting the flask `session` which is used e.g. for `session.user`. it only doesn't break stuff because we only access the session before this line and not after it.

actually, i don't think there's a need for this assignment at all - you even use `track.default_session` again in L52..

(i'll fix it)

PR #3074, opened 2017-09-26
Enforce isort rules in travis

ThiefMaster:


.isort.cfg
```diff
@@ -2,7 +2,6 @@
 line_length=120
 multi_line_output=0
 lines_after_imports=2
-add_imports=from __future__ import unicode_literals
```
ThiefMaster: there are still quite a few older files that don't use unicode literals and getting warnings (or now: travis failures) there is pretty annoying...

.travis.yml
```diff
@@ -2,17 +2,18 @@ sudo: false
 language: python
 python:
   - 2.7
+git:
+  submodules: false
```
ThiefMaster: only UI stuff that's not needed for tests

.travis.yml
```diff
@@ -2,17 +2,18 @@ sudo: false
...
 addons:
   postgresql: 9.6
-env:
-  global:
-    - INDICO_CONFIG=/dev/null
```
ThiefMaster: explicitly disabling the config is not needed anymore since the refactoring

PR #3076, opened 2017-09-27
List of participants

mvidalgarcia:
TODO:
- [x] Search box
- [x] Fix misaligned ticket icons in table rows
- [x] Show full role name in `qtip` when hovering
- [x] Make distinction between builtin and custom role labels
- [x] Disable reset filters button when it's in its initial state

### DIFF
--- a/indico/htdocs/js/indico/modules/events/management.js
+++ b/indico/htdocs/js/indico/modules/events/management.js
@@ -83,9 +83,12 @@
     };
 
     function refreshPersonFilters() {
+        $('#person-filters ul > li').removeClass('enabled');
         var personRows = $('.js-event-person-list tr[data-person-roles]');
         var filters = $('.js-event-person-list [data-filter]:checked').map(function() {
-            return $(this).data('filter');
+            var $this = $(this);
+            $this.closest('li').addClass('enabled');
+            return $this.data('filter');
         }).get();
 
         var visibleEntries = personRows.filter(function() {
@@ -184,6 +187,32 @@
                 delay: 100,
                 fixed: true
             }
+        });
+
+        var $personFilters = $('#person-filters');
+        // Sets background color of role filter items based on their colored squared color
+        $personFilters.find('li .colored-square').each(function() {
+            var $this = $(this);
+            $this.closest('li').css('background-color', $.Color($this.css('color')).alpha(0.1));
+        });
+
+        // Reset role filters
+        $personFilters.find('.js-reset-role-filter').on('click', function() {
+            $('.js-event-person-list [data-filter]').each(function() {
+                var $this = $(this);
+                $this.prop('checked', !$this.is('#filter-no-account'));
+                $this.parent().toggleClass('enabled', !$this.is('#filter-no-account'));
+            });
+            refreshPersonFilters();
+        });
+
+        // Allows to click in the whole list item area to enable/disable role filters
+        $personFilters.find('li label').on('click', function(evt) {
+            evt.preventDefault();
+        });
+        $personFilters.find('li').on('click', function() {
+            var $checkbox = $(this).find('[data-filter]');
+            $checkbox.prop('checked', !$checkbox.prop('checked')).trigger('change');
         });
     };
 
--- a/indico/htdocs/js/indico/modules/events/management.js
+++ b/indico/htdocs/js/indico/modules/events/management.js
@@ -144,7 +144,8 @@
             sortList: [[1, 0]]
         });
 
-        $('.js-event-person-list .roles-column > span').qbubble({
+        var $roleLabels = $('.js-event-person-list .roles-column > span');
+        $roleLabels.qbubble({
             show: {
                 event: 'mouseover'
             },
@@ -184,6 +185,12 @@
                     return html;
                 }
             }
+        });
+
+        // Sets background color of custom role labels based on their font color
+        $roleLabels.filter('.custom').each(function() {
+            var $this = $(this);
+            $this.css('background-color', $.Color($this.css('color')).alpha(0.1));
         });
 
         if (options.hasNoAccountFilter) {
--- a/indico/htdocs/js/indico/modules/events/management.js
+++ b/indico/htdocs/js/indico/modules/events/management.js
@@ -104,6 +104,12 @@
         $('.js-event-person-list').trigger('indico:syncEnableIfChecked');
     }
 
+    function toggleResetBtn() {
+        var isInitialState = $('#person-filters [data-filter]:checked').length ===
+                             $('#person-filters [data-filter]:not(#filter-no-account)').length;
+        $('.js-reset-role-filter').toggleClass('disabled', isInitialState);
+    }
+
     global.setupEventPersonsList = function setupEventPersonsList(options) {
         options = $.extend({
             hasNoAccountFilter: false
@@ -223,6 +229,7 @@
             });
             refreshPersonFilters();
             applySearchFilters();
+            toggleResetBtn();
         });
 
         // Allows to click in the whole list item area to enable/disable role filters
@@ -232,6 +239,7 @@
         $personFilters.find('li').on('click', function() {
             var $checkbox = $(this).find('[data-filter]');
             $checkbox.prop('checked', !$checkbox.prop('checked')).trigger('change');
+            toggleResetBtn();
         });
     };
 
--- a/indico/htdocs/js/indico/modules/events/management.js
+++ b/indico/htdocs/js/indico/modules/events/management.js
@@ -101,6 +101,7 @@
 
         personRows.addClass('hidden');
         visibleEntries.removeClass('hidden');
+        $('#filtering-state strong').text(visibleEntries.length);
         $('.js-event-person-list').trigger('indico:syncEnableIfChecked');
     }
 
--- a/indico/htdocs/js/indico/modules/events/management.js
+++ b/indico/htdocs/js/indico/modules/events/management.js
@@ -126,7 +126,7 @@
             sortList: [[1, 0]]
         });
 
-        $('.js-event-person-list .js-count-label:not(.no-role)').qbubble({
+        $('.js-event-person-list .js-count-label').qbubble({
             show: {
                 event: 'mouseover'
             },
--- a/indico/htdocs/js/indico/modules/events/management.js
+++ b/indico/htdocs/js/indico/modules/events/management.js
@@ -138,7 +138,7 @@
             sortList: [[1, 0]]
         });
 
-        $('.js-event-person-list .js-count-label').qbubble({
+        $('.js-event-person-list .roles-column > span').qbubble({
             show: {
                 event: 'mouseover'
             },
@@ -153,20 +153,27 @@
             },
             content: {
                 text: function() {
-                    var html = $('<ul class="qbubble-item-list">');
-                    var items = _.values($(this).data('items')).sort(function(a, b) {
-                        return strnatcmp(a.title.toLowerCase(), b.title.toLowerCase());
-                    });
+                    var $this = $(this);
+                    var html = $('<div>');
+                    var role = $('<strong>', {text: "{0}".format($(this).data('role-name'))});
+                    html.append(role);
+                    if ($this.is('.js-count-label')) {
+                        var list = $('<ul>', {class: 'qbubble-item-list'});
+                        var items = _.values($this.data('items')).sort(function(a, b) {
+                            return strnatcmp(a.title.toLowerCase(), b.title.toLowerCase());
+                        });
 
-                    $.each(items, function() {
-                        var item = $('<li>');
-                        if (this.url) {
-                            item.append($('<a>', {text: this.title, href: this.url}));
-                        } else {
-                            item.text(this.title);
-                        }
-                        html.append(item);
-                    });
+                        $.each(items, function() {
+                            var item = $('<li>');
+                            if (this.url) {
+                                item.append($('<a>', {text: this.title, href: this.url}));
+                            } else {
+                                item.text(this.title);
+                            }
+                            list.append(item);
+                        });
+                        html.append(list);
+                    }
 
                     return html;
                 }
--- a/indico/htdocs/js/indico/modules/events/management.js
+++ b/indico/htdocs/js/indico/modules/events/management.js
@@ -173,6 +173,18 @@
                 refreshPersonFilters();
             });
         }
+
+        $('.js-show-regforms').qtip({
+            content: {
+                text: function() {
+                    return $(this).data('title');
+                }
+            },
+            hide: {
+                delay: 100,
+                fixed: true
+            }
+        });
     };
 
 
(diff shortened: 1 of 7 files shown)


indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -83,8 +83,10 @@
     };
 
     function refreshPersonFilters() {
+        $('.js-event-person-list #person-filters ul>li').removeClass('enabled');
         var personRows = $('.js-event-person-list tr[data-person-roles]');
         var filters = $('.js-event-person-list [data-filter]:checked').map(function() {
+            $(this).parent('li').addClass('enabled');
```
ThiefMaster: - `var $this = $(this);`
- I'd use `.closest()` instead of `.parent()` to be more resilient to minor changes in the DOM

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -98,6 +100,7 @@
 
         personRows.addClass('hidden');
         visibleEntries.removeClass('hidden');
+        $('#filtering-state strong').text(visibleEntries.length);
```
ThiefMaster: I'd prefer using a .class instead of selecting `strong`

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -173,6 +176,44 @@
                 refreshPersonFilters();
             });
         }
+
+        $('.icon-ticket.js-show-regforms').qtip({
```
ThiefMaster: `.js-show-regforms` should be enough; selecting `.icon-*` in jquery is not a good idea (that's a purely display-related class that should not have any meaning beyond that)

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -173,6 +176,44 @@
...
+                }
+            },
+            hide: {
+                delay: 100,
+                fixed: true
+            }
+        });
+
+        var $personFilters = $('#person-filters');
+        /* Sets background color of role filter items based on their icon-stop color */
```
ThiefMaster: javascript has `// foo` comments ;)

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -173,6 +176,44 @@
...
+        $personFilters.find('li .icon-stop').each(function() {
```
ThiefMaster: see above - icons are not the best choice for selectors here

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -173,6 +176,44 @@
...
+            var $this = $(this);
+            $this.parent('li').css('background-color', $.Color($this.css('color')).alpha(0.1));
+        });
+
+        /* Reset role filters */
+        $personFilters.find('.js-reset-role-filter').on('click', function() {
+            $('.js-event-person-list [data-filter]').each(function() {
+                var $this = $(this);
+                $this.prop('checked', !$(this).is('#filter-no-account'))
+                     .parent().toggleClass('enabled', !$(this).is('#filter-no-account'));
```
ThiefMaster: I think it's more readable if you repeat the `$this` and make it two statements

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -173,6 +176,44 @@
...
+            });
+            refreshPersonFilters();
+        });
+
+        /* Allows to click in the whole list item area to enable/disable role filters */
+        $personFilters.find('li label').on('click', function(e) {
```
ThiefMaster: we usually use `evt` for the event argument

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -83,9 +83,12 @@
     };
 
     function refreshPersonFilters() {
+        $('.js-event-person-list #person-filters ul>li').removeClass('enabled');
```
ThiefMaster: the first part of the selector isn't really needed since a `#id` selector is already unique

indico/modules/events/persons/controllers.py
```diff
@@ -44,6 +47,16 @@
...
 
+BUILTIN_ROLES = {'chairperson': {'name': 'Chairperson', 'code': 'CHR', 'color': 'f7b076',
+                                 'css': 'color: #f7b076 !important; border-color: #f7b076 !important'},
+                 'author': {'name': 'Author', 'code': 'AUT', 'color': '7994f7',
+                            'css': 'color: #7994f7 !important; border-color: #7994f7 !important'},
+                 'convener': {'name': 'Convener', 'code': 'CON', 'color': 'dcb1f9',
+                              'css': 'color: #dcb1f9 !important; border-color: #dcb1f9 !important'},
+                 'speaker': {'name': 'Speaker', 'code': 'SPK', 'color': '7fd1c8',
+                             'css': 'color: #7fd1c8 !important; border-color: #7fd1c8 !important'}
+                 }
```
ThiefMaster: I'd either move this into the previous line or use this style:

```
BUILTIN_ROLES = {
    'chairperson': ...,
    'author': ...,
}
```

(whatever style you prefer or is more readable, but the indented `}` in a separate line is super ugly)

indico/modules/events/persons/controllers.py
```diff
@@ -63,62 +76,84 @@ def get_persons(self):
...
         chairpersons = {link.person for link in self.event.person_links}
-        persons = defaultdict(lambda: {'session_blocks': defaultdict(dict),
-                                       'contributions': defaultdict(dict),
-                                       'subcontributions': defaultdict(dict),
-                                       'abstracts': defaultdict(dict),
-                                       'roles': defaultdict(lambda: False)})
+        persons = defaultdict(lambda: {'roles': OrderedDict(), 'registrations': []})
 
-        event_persons_query = (self.event.persons
+        event_persons_query = (db.session.query(EventPerson, Registration)
```
ThiefMaster: Is using a single query for this the best idea? Seems cleaner to split it in two queries. Especially since we do want to show entries where there is only a registration but no `EventPerson`. BTW, we also want to show people who have neither but are in an `EventRole`.

indico/modules/events/persons/controllers.py
```diff
@@ -152,9 +187,11 @@ def _process(self):
             if not persons[principal.email].get('no_account'):
                 persons[principal.email]['roles']['no_account'] = True
                 num_no_account += 1
-
-        return WPManagePersons.render_template('management/person_list.html', self.event,
-                                               persons=person_list, num_no_account=num_no_account)
+        custom_roles = {'custom_{}'.format(r.id): {'name': r.name, 'code': r.code, 'color': r.color}
+                        for r in EventRole.query.with_parent(self.event).all()}
```
ThiefMaster: why not `self.event.roles` (or whatever I called the backref of the relationship)?

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -3,7 +3,24 @@
...
+                {%- trans %}The person has registered in:{% endtrans -%}
+                {% for reg in person_data.registrations %}
+                    {% set link_start %}
+                        <a href='{{ url_for('event_registration.registration_details', reg) }}'>
+                    {% endset %}
+                    {% set link_end %}</a>{% endset %}
+                    <div>{{ link_start }}{{ reg.registration_form.title }}{{ link_end }}</div>
+                {% endfor %}
+            {% else %}
+                {%- trans %}The person hasn't registered yet{% endtrans -%}
```
ThiefMaster: has not

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -3,7 +3,24 @@
...
+    <td class="i-table registration-column">
+        {% set tooltip %}
+            {% if person_data.registrations %}
+                {%- trans %}The person has registered in:{% endtrans -%}
+                {% for reg in person_data.registrations %}
+                    {% set link_start %}
+                        <a href='{{ url_for('event_registration.registration_details', reg) }}'>
+                    {% endset %}
+                    {% set link_end %}</a>{% endset %}
+                    <div>{{ link_start }}{{ reg.registration_form.title }}{{ link_end }}</div>
```
ThiefMaster: you don't need this link_start/link_end stuff outside translated strings

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -3,7 +3,24 @@
...
+                {% endfor %}
+            {% else %}
+                {%- trans %}The person hasn't registered yet{% endtrans -%}
+            {% endif %}
+        {% endset %}
+        <i class="icon-ticket js-show-regforms {{ 'exists' if person_data.registrations}}" data-title="{{ tooltip }}"></i>
```
ThiefMaster: missing space before `}}`

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -21,61 +38,32 @@
...
     </td>
-    <td class="i-table">
-        <span class="i-label {{ 'no-role' if not person_data.roles.chairperson }}">
-            {%- if person.event.type == 'lecture' -%}
-                {%- trans %}Speaker{% endtrans -%}
-            {%- else -%}
-                {%- trans %}Chairperson{% endtrans -%}
+    <td class="i-table roles-column">
+        {% for role, role_data in person_data.roles.iteritems() %}
+            {%- if role_data is mapping %}
```
ThiefMaster: when is it not a mapping?

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -21,61 +38,32 @@
...
+                {%- if 'elements' in role_data -%}
```
ThiefMaster: unless you really need to know whether it exists (even if empty) or is really not present, you could just use `role_data.elements`

indico/modules/events/persons/controllers.py
mvidalgarcia: Do we want to show all the registrants even if they don't belong to any role?
I've made this query together with @pferreir and in that moment we thought that having results as a tuple of  `(EventPerson, Registration)` was the best idea.

ThiefMaster: OK, if we don't want to show registrants without other roles then this makes sense

mvidalgarcia: Yes I agree, I missed those (`EventRole.members`)

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -141,20 +159,27 @@
...
-                    var html = $('<ul class="qbubble-item-list">');
-                    var items = _.values($(this).data('items')).sort(function(a, b) {
-                        return strnatcmp(a.title.toLowerCase(), b.title.toLowerCase());
-                    });
+                    var $this = $(this);
+                    var html = $('<div>');
+                    var role = $('<strong>', {text: "{0}".format($(this).data('role-name'))});
+                    html.append(role);
+                    if ($this.is('.js-count-label')) {
+                        var list = $('<ul class="qbubble-item-list">');
```
mic4ael: maybe it is a good opportunity to change it to `$('<ul>', {'class': '...'}`

indico/modules/events/persons/controllers.py
```diff
@@ -63,62 +76,84 @@ def get_persons(self):
...
+                        'url': url_for('abstracts.display_abstract', person_link.abstract, management=True)
+                    }
+                    for person_link in event_person.abstract_links if not person_link.abstract.is_deleted
+                }
+
+                if abstracts:
+                    data['roles']['author'] = BUILTIN_ROLES['author'].copy()
+                    data['roles']['author']['elements'] = abstracts
+
+            session_blocks = {
```
mic4ael: personally I find it a bit unreadable, what about a normal `for` statement instead of the list comprehensions

indico/modules/events/persons/controllers.py
```diff
@@ -63,62 +76,84 @@ def get_persons(self):
...
+                }
+                for person_link in event_person.session_block_links
+                if not person_link.session_block.session.is_deleted
+            }
+
+            if session_blocks:
+                data['roles']['convener'] = BUILTIN_ROLES['convener'].copy()
+                data['roles']['convener']['elements'] = session_blocks
+
+            contributions = {
```
mic4ael: the same comment as above

indico/modules/events/persons/controllers.py
ThiefMaster: agreed, this comprehension is pretty ugly. but i don't think using a normal for loop will be much better. why not add some helper functions that generate the data dicts? then the comprehension would be much more readable.

mvidalgarcia: Also, I pair-programmed this together with @pferreir, we thought about the possibility of using a helper function but, if you notice, there are some differences among each role. For instance, `abstracts` has `title` and `url` whereas `session_blocks` just has `title`, for the `subcontributions` we need to access the contribution title so the logic is different from the `contributions` itself and so on. I think the helper function would be very complex, unless you have a better idea on how to implement it.

ThiefMaster: i meant having one helper function for each type. writing one big (and ugly) function would be somewhat ugly indeed

mvidalgarcia:
Updated.

indico/modules/events/persons/controllers.py
```diff
@@ -148,7 +148,9 @@ def get_persons(self):
 
             if event_person.user:
                 for role in event_person.user.event_roles:
```
ThiefMaster: `.filter_by(event=self.event)`

mvidalgarcia: The relationship is not dynamic.
`AttributeError: 'InstrumentedSet' object has no attribute 'filter_by'`

ThiefMaster: OK, would be causing query spam anyway since it's in a loop..

But...o this always loads all event roles of the user - and since we don't subqueryload/joinedload it, it's an extra query for every single user anyway. Maybe we can build a dict of user->roles mappings only for people in the event outside the loop?

ThiefMaster:
Should we hide the registration icon/column if the event has no registration forms?

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -87,12 +75,18 @@
...
         {% endif %}
     </td>
-    <td class="i-table">
-        <a class="i-link icon-edit" title="{% trans %}Edit information{% endtrans %}"
-            data-href="{{ url_for('persons.edit_person', person) }}"
-            data-title="{% trans %}Edit information{% endtrans %}"
-            data-update="#person-{{ person.id }}"
-            data-highlight-update
-            data-ajax-dialog></a>
+    <td class="i-table edit-column">
```
ThiefMaster: I think this should also have the `thin-column` class

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -3,7 +3,24 @@
     <td class="i-table thin-column hide-if-locked">
         <input type="checkbox" value="{{ person.id }}" class="select-row" name="person_id" {{ 'disabled' if not person.email }}>
     </td>
-    <td class="i-table name-column">
+    <td class="i-table registration-column">
```
ThiefMaster: I think this should also have the `thin-column` class

PR #3079, opened 2017-09-28
Fix contribution type placeholder for abstract notifications

meluru:


indico/modules/events/abstracts/placeholders.py
```diff
@@ -105,9 +105,16 @@ class ContributionTypePlaceholder(Placeholder):
...
     def render(cls, abstract):
-        if abstract.state == AbstractState.accepted:
+        if abstract.state == AbstractState.withdrawn:
+            if abstract.accepted_contrib_type:
+                return abstract.accepted_contrib_type.name
+            elif abstract.submitted_contrib_type:
+                return abstract.submitted_contrib_type.name
+            return ''
+        elif abstract.state != AbstractState.accepted:
+            return abstract.submitted_contrib_type.name if abstract.submitted_contrib_type else ''
```
ThiefMaster: you can simplify the whole thing if you set `ctype = ...` and then at the very end do `return ctype.name if ctype else ''`

meluru:
updated

indico/modules/events/abstracts/placeholders.py
```diff
@@ -105,9 +105,15 @@ class ContributionTypePlaceholder(Placeholder):
 
     @classmethod
     def render(cls, abstract):
-        if abstract.state == AbstractState.accepted:
-            return abstract.accepted_contrib_type.name if abstract.accepted_contrib_type else ''
-        return ''
+        if abstract.state == AbstractState.withdrawn:
+            ctype = (abstract.accepted_contrib_type
```
ThiefMaster: ctype = abstract.accepted_contrib_type or abstract.submitted_contrib_type

meluru:
updated

indico/modules/events/abstracts/placeholders.py
```diff
@@ -105,9 +105,13 @@ class ContributionTypePlaceholder(Placeholder):
 
     @classmethod
     def render(cls, abstract):
-        if abstract.state == AbstractState.accepted:
-            return abstract.accepted_contrib_type.name if abstract.accepted_contrib_type else ''
-        return ''
+        if abstract.state == AbstractState.withdrawn:
+            ctype = abstract.accepted_contrib_type or abstract.submitted_contrib_type
+        elif abstract.state != AbstractState.accepted:
```
pferreir: This kind of logic is a bit confusing. I initially read that `!=` as `==` and it didn't make sense, of course. In cases where there's `if... else...` I normally favour using the `if` for the "positive" case.

PR #3081, opened 2017-09-29
Development docs

pferreir:
I've also added a `--dev` option to the setup wizard script.

Things to be discussed:
 * What should be the default BaseURL?
 * Should it be HTTP or HTTPS?
 * Should we assume the developer will be behind nginx?

I personally think we should keep things as simple as possible and then have some "optional" section where we explain how to move behind nginx.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,3 +7,7 @@ max-line-length = 120
 
 [pep257]
 ignore = D100
+
+[pycodestyle]
+exclude = .git,build,dist,ext_modules,node_modules,indico.egg-info,__pycache__,docs,htmlcov
+max-line-length = 120


ThiefMaster:
> What should be the default BaseURL?
> Should it be HTTP or HTTPS?
> Should we assume the developer will be behind nginx?


I would go for `http://127.0.0.1:8000` - that works everywhere, no matter if IPv6 is available or not, and doesn't require generating a certificate. As much as I think using nginx in development is a good idea, I'd keep it in a "bonus" section. The changes on running indico are minimal (sendfile method, no port in the base url + cli switch to enable the proxy middleware) so it shouldn't be too confusing to mention that in the nginx section. In that section I'd also enable SSL (ie terminate SSL on nginx and use plain HTTP between nginx and the dev server)

docs/source/installation/development.rst
```diff
@@ -3,5 +3,123 @@
...
 
-.. todo::
-    Write guide
+Install System Packages
+-----------------------
+
+We'll assume you've installed PostgreSQL (and its ``contrib package``, if on a Linux system) and managed to setup/start
+the server. PostgreSQL 9.6 at least is required. You can find instructions on how to install it on Fedora/CentOS
+`here <https://www.if-not-true-then-false.com/2012/install-postgresql-on-fedora-centos-red-hat-rhel/>`_.
+You can find the same for Debian/Ubuntu `here <https://www.tecmint.com/install-postgresql-on-debian-and-ubuntu/>`_.
```
ThiefMaster: I'd explain the steps here (almost copy&paste from the prod setup guides) instead of linking to external blogs that may or may not break in the future. Or at least link to something more official, like https://wiki.postgresql.org/wiki/Apt and https://wiki.postgresql.org/wiki/YUM_Installation

docs/source/installation/development.rst
```diff
@@ -3,5 +3,123 @@
 Installation guide (development)
 ===============================
 
-.. todo::
-    Write guide
+Install System Packages
```
ThiefMaster: "Install ..." but "Creating ..." etc. in the next sections -> "Installing" here?

docs/source/installation/development.rst
```diff
@@ -3,5 +3,123 @@
...
+you clone your own `GitHub fork of the project <https://help.github.com/articles/fork-a-repo/>`_. Otherwise, if you're
+just thinking of playing a bit with the source code, it's enough if you clone the upstream repo:
+
+.. code-block:: shell
+
+    git clone --recursive https://github.com/indico/indico.git src
+
+If you're going to be changing the standard Indico plugins and/or the documentation, you can also clone those::
+
+    git clone https://github.com/indico/indico-plugins.git plugins
```
ThiefMaster: I'd add a subfolder `plugins/something` for people who'll also work on their own plugins (or clone the CERN plugins in case of team members). Like this you can open "plugins" as another project in your IDE and have all plugins there, no matter from which repo.

docs/source/installation/development.rst
```diff
@@ -3,5 +3,123 @@
...
+Cloning Indico
+--------------
+
+First, let's clone Indico's code base. If you're going to contribute back to the project, it's probably best if
+you clone your own `GitHub fork of the project <https://help.github.com/articles/fork-a-repo/>`_. Otherwise, if you're
+just thinking of playing a bit with the source code, it's enough if you clone the upstream repo:
+
+.. code-block:: shell
+
+    git clone --recursive https://github.com/indico/indico.git src
```
ThiefMaster: What about `--origin upstream` so people can later add their own fork (if they want to contribute back to indico) as `origin`?

docs/source/installation/development.rst
```diff
@@ -3,5 +3,123 @@
...
+
+If you're going to be changing the standard Indico plugins and/or the documentation, you can also clone those::
+
+    git clone https://github.com/indico/indico-plugins.git plugins
+    git clone https://github.com/indico/indico-user-docs.git user-docs
+
+
+Setting up Maildump (advised)
+-----------------------------
+
```
ThiefMaster: I'd add a short explanation why maildump is a good idea

docs/source/installation/development.rst
```diff
@@ -3,5 +3,123 @@
...
+.. code-block:: shell
+
+    virtualenv maildump -p /usr/bin/python2.7
+    . ./maildump/bin/activate
```
ThiefMaster: not necessary if you use `./maildump/bin/pip install maildump` and `./maildump/bin/maildump -p /tmp/maildump.pid` below. avoids the risk of someone continuing with indico stuff  inside the maildump virtualenv

ThiefMaster: BTW, I'd use `source` instead of `.` in docs - makes  it less likely that someone forgets to copy/type it since the `.` looks a bit like a typo for someone who doesn't know it.

docs/source/installation/development.rst
```diff
@@ -3,5 +3,123 @@
...
+    pip install maildump
+    maildump -p /tmp/maildump.pid
+
+
+Creating the DB
+---------------
+
+.. code-block:: shell
+
+    sudo -u postgres createuser <your-user>
```
ThiefMaster: I would create the user with createdb privileges and create an `indico_template` db here where you create the extensions. Like this people will be able to use the dbdiff tool and they can create the indico database itself without sudo.

docs/source/installation/development.rst
```diff
@@ -3,5 +3,123 @@
...
+
+    indico db prepare
+
+
+Running Indico
+--------------
+
+.. code-block:: shell
+
+    indico run -h <your-hostname>
```
ThiefMaster: `-q` to not show requests for static files and maybe `--enable-evalex` because it's useful (since there's the PIN it's safe(-ish) even if someone exposes it to more than localhost)

indico/cli/setup.py
```diff
@@ -289,12 +296,15 @@ def _check_configured(self):
                 self.config_path))
             raise click.Abort
 
-    def _check_directories(self):
-        dirs = ['archive', 'assets', 'cache', 'etc', 'log', 'tmp', 'web']
+    def _check_directories(self, dev=False):
+        dirs = ['archive', 'assets', 'cache', 'log', 'tmp', 'web']
```
ThiefMaster: `web` isn't needed in dev mode

indico/cli/setup.py
```diff
@@ -402,12 +413,14 @@ def _get_noreply_email(email):
                                      help='When Indico sends email notifications they are usually sent from the '
                                           'noreply email.')
 
-    def _prompt_smtp_data(self):
+    def _prompt_smtp_data(self, dev=False):
```
ThiefMaster: Should we even prompt for this in dev mode instead of just setting 127.0.0.1:1025?

indico/cli/setup.py
```diff
@@ -550,9 +569,9 @@ def _setup(self):
 
         package_root = get_root_path('indico')
         _copy(os.path.normpath(os.path.join(package_root, 'logging.yaml.sample')),
-              os.path.join(self.root_path, 'etc', 'logging.yaml'))
-        _link(os.path.join(package_root, 'htdocs'), os.path.join(self.root_path, 'web', 'htdocs'))
-        _copy(os.path.join(package_root, 'web', 'indico.wsgi'), os.path.join(self.root_path, 'web', 'indico.wsgi'),
+              os.path.join(self.config_dir_path, 'logging.yaml'))
+        _link(os.path.join(package_root, 'htdocs'), os.path.join(self.data_root_path, 'web', 'htdocs'))
```
ThiefMaster: not needed in dev mode; since indico is installed in `--editable` mode the path to htdocs for a nginx config is pretty clear without the symlink

indico/cli/setup.py
```diff
@@ -550,9 +569,9 @@ def _setup(self):
...
+        _copy(os.path.join(package_root, 'web', 'indico.wsgi'), os.path.join(self.data_root_path, 'web', 'indico.wsgi'),
```
ThiefMaster: also not needed in dev mode

indico/cli/setup.py
pferreir: Should we assume people will always use maildump?

ThiefMaster: for development it might be a good idea instead of having people put their gmail credentials in the indico config (which i think is not too uncommon, as ugly as it is :/)

pferreir: What's wrong about having a default that looks good and still allowing for some flexibility? After all, people may choose to use other than maildump...

ThiefMaster:
The list of valid locales should be taken from the directories instead of looking for compiled locales since we have none compiled at this step.

pferreir:
> The list of valid locales should be taken from the directories instead of looking for compiled locales since we have none compiled at this step.

Done.

docs/source/installation/development.rst
```diff
@@ -3,5 +3,268 @@
...
+    git remote add upstream https://github.com/indico/indico.git src
+    cd ..
+
+Otherwise, cloning the upstream repository as the origin should be enough::
+
+    git clone --recursive https://github.com/indico/indico.git src
+
+If you're going to be changing the standard Indico plugins and/or the documentation, you can also clone those::
+
+    mkdir plugins
```
ThiefMaster: `cd ~/dev/indico` before doing that

pferreir: But... aren't we already there?

docs/source/installation/development.rst
```diff
@@ -3,5 +3,268 @@
...
+---------------------------------
+
+Some actions in Indico trigger automatic e-mails. Those will normally have to be routed through an SMTP server.
+This can become a problem if you're using production data and/or real e-mails, as users may end up being spammed
+unnecessarily. This is why we advise that you include a fake SMTP server in your development setup.
+`Maildump <https://github.com/ThiefMaster/maildump>`_ does exactly this and runs on Python. It should be quite simple
+to set up::
+
+    virtualenv maildump -p /usr/bin/python2.7
+    ./maildump/bin/pip install maildump
```
ThiefMaster: I think you need libevent-dev for maildump since it uses gevent

docs/source/installation/development.rst
ThiefMaster: ah right, didn't notice the `cd ..` above

docs/source/installation/development.rst
pferreir: I don't remember having installed anything when configuring my new box. Maybe it was already installed by something else?

docs/source/installation/development.rst
```diff
@@ -3,5 +3,268 @@
...
+
+You'll then be able to access the message log at `<http://localhost:1080>`_.
+
+
+Creating the DB
+---------------
+
+.. code-block:: shell
+
+    sudo -u postgres createuser <your-user> --createdb
```
ThiefMaster: `$USER` instead of `<your-user>`? (do all shells set that?)

docs/source/installation/development.rst
```diff
@@ -3,5 +3,268 @@
...
+    sudo -u postgres createdb indico_template -O <your-user>
+    sudo -u postgres psql indico_template -c "CREATE EXTENSION unaccent; CREATE EXTENSION pg_trgm;"
+    createdb -O <your-user> indico -T indico_template
```
ThiefMaster: the `-O <your-user>` isn't needed here since you are using your user by default

docs/source/installation/development.rst
```diff
@@ -3,5 +3,268 @@
...
+-----------
+
+Let's get into the Indico virtualenv::
+
+    source ./env/bin/activate
+
+    cd src
+    pip install -r requirements.dev.txt
+    pip install -e .
+    fab setup_deps:system_node=1
```
ThiefMaster: we never install nodejs above - so maybe safer to use the nodeenv node?

docs/source/installation/development.rst
```diff
@@ -3,5 +3,268 @@
...
+
+    indico db prepare
+
+
+Running Indico
+--------------
+
+.. code-block:: shell
+
+    indico run -h <your-hostname> -q --enable-evalex
```
ThiefMaster: `-h 127.0.0.1` since that's what we use in the default base url generated by the wizard?

setup.cfg
```diff
@@ -7,3 +7,7 @@ max-line-length = 120
 
 [pep257]
 ignore = D100
+
+[pycodestyle]
+exclude = .git,build,dist,ext_modules,node_modules,indico.egg-info
```
ThiefMaster: `__pycache__,docs,htmlcov` are missing there

docs/source/installation/development.rst
pferreir: argh. Yeah, maybe. Or we can add node as a dependency.

docs/source/installation/development.rst
pferreir: But the user is still free to choose whatever they want, so why confuse them?

ThiefMaster: true, the setting here overrides `BASE_URL` anyway...

docs/source/installation/development.rst
pferreir: OK, `libevent-devel` isn't even installed.

ThiefMaster: ok, maybe it was something else i had in mind - just tried it on the vm i used to test the prod setup and also managed to install maildump without anything else

PR #3083, opened 2017-10-03
Pass missing argument to the contribution list template

mic4ael:


indico/modules/events/contributions/controllers/common.py
```diff
@@ -31,4 +31,4 @@ def _process(self):
         return self._render_template(**self.list_generator.get_list_kwargs())
 
     def _render_template(self, selected_entry, **kwargs):
-        return self.view_class.render_template(self.template, self.event, **kwargs)
+        return self.view_class.render_template(self.template, self.event, **dict(kwargs, selected_entry=selected_entry))
```
ThiefMaster: `selected_entry=selected_entry, **kwargs` (but check that we never end up calling this with `kwargs` already containing it)

ThiefMaster:
Merged in 0a3e6b0b3a613400080945479aeb94b2289fe105

PR #3084, opened 2017-10-04
Make iCal UID more unique

mic4ael:


indico/modules/categories/serialize.py
```diff
@@ -82,7 +83,7 @@ def serialize_categories_ical(category_ids, user, event_filter=True, event_filte
                     if event.venue_name and event.room_name
                     else (event.venue_name or event.room_name))
         cal_event = ical.Event()
-        cal_event.add('uid', u'indico-event-{}@cern.ch'.format(event.id))
+        cal_event.add('uid', u'indico-event-{}@{}'.format(event.id, gethostname()))
```
ThiefMaster: i don't think using the hostname of the machine is a good idea. it'll break in multi-worker environment and also discloses the hostnames of the workers to users. why not use the host of the base url? `url_parse(config.BASE_URI).host`

indico/web/http_api/metadata/ical.py
```diff
@@ -55,7 +57,7 @@ def _deserialize_date(date_dict):
 
 def serialize_event(cal, fossil, now, id_prefix="indico-event"):
     event = ical.Event()
-    event.add('uid', '%s-%s@cern.ch' % (id_prefix, fossil['id']))
+    event.add('uid', '%s-%s@%s' % (id_prefix, fossil['id'], gethostname()))
```
ThiefMaster: lets change this to use `.format()`

indico/modules/categories/serialize.py
mic4ael: > also discloses the hostnames of the workers to users

good point

indico/modules/categories/serialize.py
```diff
@@ -18,6 +18,7 @@
 
 from io import BytesIO
 from itertools import ifilter
+from urlparse import urlparse
```
ThiefMaster: i think usually we use `from werkzeug.urls import url_parse`  - it's pretty much the same but probably more unicode-friendly

indico/web/http_api/metadata/ical.py
```diff
@@ -55,7 +58,7 @@ def _deserialize_date(date_dict):
 
 def serialize_event(cal, fossil, now, id_prefix="indico-event"):
     event = ical.Event()
-    event.add('uid', '%s-%s@cern.ch' % (id_prefix, fossil['id']))
+    event.add('uid', '%s-%s@%s'.format(id_prefix, fossil['id'], urlparse(config.BASE_URL).hostname))
```
ThiefMaster: format uses `{}` instead of `%s`

mic4ael: i know ;P my mistake I didn't change it

PR #3085, opened 2017-10-04
Refactor error handling

ThiefMaster:
...and ~~some~~ many more things I stumbled upon while doing so ;)


indico/htdocs/js/indico/Core/Dialogs/Util.js
```diff
@@ -15,23 +15,64 @@
...
+                    $content.find('.js-info-text').remove();
+                }
+            }
+        });
     }
-};
+
+    IndicoUI.Dialogs.Util = {
+        error: showErrorDialog,
+
```
mic4ael: Is this for readability or..?

ThiefMaster: I think I'll just expose and call `showErrorDialog` directly instead of keeping the legacy `IndicoUI.Dialogs.Util.error` call

indico/htdocs/js/presentation/Data/Remote.js
```diff
@@ -17,51 +12,51 @@ var ReadyState = new Enum("Uninitialized", "Loading", "Loaded", "Interactive", "
...
-        self.refresh = function() {
-                object.update({
-                        state: SourceState.Loading,
-                        error: null
-                });
-                return jsonRpc(url, method, params, process);
-        };
-        if (!dontStart) {
-            self.refresh();
+        if (callback) {
```
mic4ael: `$.isFunction(callback)`?

ThiefMaster: that's legacy code, and we can assume a `callback` is callable anyway ;)

indico/core/errors.py
```diff
@@ -106,14 +56,10 @@ def wrap_exc(cls, exc):
...
-    pass
-
-
 class UserValueError(NoReportError):
     """Error to indicate that the user entered invalid data.
 
     This behaves basically like NoReportError but it comes with
-    a 400 status code so AJAX error handling properly goes into
-    the error callback instead of the success callback.
+    a 400 status code isntead of the usual 500 code.
```
pferreir: Typo

indico/htdocs/js/presentation/Data/Remote.js
```diff
@@ -4,11 +4,6 @@
  */
```
pferreir: What is the advantage of changing these methods vs. just not using them and calling `$.ajax` directly instead? This is so legacy that we should just get rid of it if we have the chance to.

ThiefMaster: RB is still using JSONRPC and while the operations like rejecting blockings could easily be changed, the few areas using `indicoSource` might be uglier to adapt. OTOH, I think it's used only in two places (rb holiday/weekend warning and room blocking), and replacing the former with a normal ajax call would be pretty easy and the latter could probably just be moved to a WTForm since it's a "standard" form

pferreir: Yeah, we'll get to RB soon anyway. No strong opinion, just wondering.

PR #3086, opened 2017-10-04
Prevent submitting papers in some circumstances

mic4ael:


indico/modules/events/papers/controllers/display.py
```diff
@@ -43,6 +43,12 @@ class RHSubmitPaper(RHPaperBase):
     PAPER_REQUIRED = False
     ALLOW_LOCKED = True
 
+    def _check_paper_protection(self):
+        if not RHPaperBase._check_paper_protection(self):
+            return False
+        last_revision = self.contribution._paper_last_revision
+        return last_revision is None or last_revision.state == PaperRevisionState.to_be_corrected
```
ThiefMaster: Do we also want to require the last revision to have been submitted by the same user? Or should any author be able to submit a new one once it's in "to be corrected"?

indico/modules/events/papers/util.py
```diff
@@ -85,5 +85,9 @@ def has_contributions_with_user_paper_submission_rights(event, user):
     return _query_contributions_with_user_paper_submission_rights(event, user).has_rows()
 
 
-def get_contributions_with_user_paper_submission_rights(event, user):
-    return _query_contributions_with_user_paper_submission_rights(event, user).all()
+def get_contributions_with_submission_rights_and_proper_state(event, user):
```
ThiefMaster: Can't we find a nicer name for this? `get_user_submittable_contributions` isn't super nice either but doesn't sound as bad as both the previous name or the updated one.

indico/modules/events/papers/controllers/display.py
mic4ael: @pferreir what do you think?

mic4ael: But i guess we shouldn't limit it to the same user unless there is a good reason for that.

pferreir: I think we should not limit it to the same person. Being flexible in this case is probably the best.

pferreir: Think of co-authors working collaboratively on a paper.

indico/modules/events/papers/util.py
```diff
@@ -85,5 +85,9 @@ def has_contributions_with_user_paper_submission_rights(event, user):
     return _query_contributions_with_user_paper_submission_rights(event, user).has_rows()
 
 
-def get_contributions_with_user_paper_submission_rights(event, user):
-    return _query_contributions_with_user_paper_submission_rights(event, user).all()
+def get_user_submittable_contributions(event, user):
+    allowed_states = [PaperRevisionState.to_be_corrected]
+    return (_query_contributions_with_user_paper_submission_rights(event, user)
+            .filter(db.or_(Contribution._paper_last_revision == None,  # noqa
```
pferreir: why `noqa`?

ThiefMaster: `== None` (sqlalchemy does not implement `is_` and `isnot` there)

pferreir: Ah right.

ThiefMaster:
Merged in e2f4545ef (and fixed in 1c86d550e)

PR #3087, opened 2017-10-05
Hide table of contents when necessary

mic4ael:


ThiefMaster:
If a commit fixes an issue, can to add `closes #xxx` to the commit message?

indico/legacy/pdfinterface/base.py
```diff
@@ -621,7 +620,8 @@ def __init__(self, story=None, pagesize='A4', fontsize='normal', firstPageNumber
         self._fileDummy = FileDummy()
 
         self._doc = DocTemplateWithTOC(self._indexedFlowable, self._fileDummy, firstPageNumber=firstPageNumber,
-                                       pagesize=PDFSizes().PDFpagesizes[pagesize])
+                                       pagesize=PDFSizes().PDFpagesizes[pagesize],
+                                       include_table_of_contents=include_table_of_contents)
```
pferreir: `include_toc`, maybe?

indico/legacy/pdfinterface/base.py
```diff
@@ -503,13 +503,14 @@ def _doNothing(canvas, doc):
...
 
-    def __init__(self, indexedFlowable, filename, firstPageNumber = 1, **kw ):
+    def __init__(self, indexedFlowable, filename, firstPageNumber=1, **kw):
         """toc is the TableOfContents object
         indexedFlowale is a dictionnary with flowables as key and a dictionnary as value.
             the sub-dictionnary have two key:
                 text: the text which will br print in the table
                 level: the level of the entry( modifying the indentation and the police
         """
+        self.include_table_of_contents = kw.get('include_table_of_contents', False)
```
pferreir: Why not add this as a kwarg of the constructor?

indico/legacy/pdfinterface/base.py
mic4ael: ok, I don't mind changing it

indico/legacy/pdfinterface/base.py
mic4ael: don't mind changing it

mic4ael: changed

indico/legacy/pdfinterface/base.py
```diff
@@ -503,13 +503,14 @@ def _doNothing(canvas, doc):
 
 class DocTemplateWithTOC(SimpleDocTemplate):
 
-    def __init__(self, indexedFlowable, filename, firstPageNumber = 1, **kw ):
+    def __init__(self, indexedFlowable, filename, firstPageNumber=1, include_toc=False, **kw):
```
ThiefMaster: fyi, this class is used in other places  where the TOC is now disabled too - so this argument should be True by default (i'm changing it, no need to update the PR)

PR #3088, opened 2017-10-05
Properly sanitize output and run legacy sanitizer only for legacy code

### LINKED ISSUE #3004: Sanitize custom page HTML at render time
Currently we rely on the legacy HTML sanitizer and just assume the HTML that's in the database is safe.

ThiefMaster:
closes #3004

### DIFF
--- a/indico/modules/attachments/controllers/management/base.py
+++ b/indico/modules/attachments/controllers/management/base.py
@@ -100,7 +100,7 @@ class AddAttachmentFilesMixin:
                     attachment.acl = form.acl.data
                 content_type = mimetypes.guess_type(f.filename)[0] or f.mimetype or 'application/octet-stream'
                 attachment.file = AttachmentFile(user=session.user, filename=filename, content_type=content_type)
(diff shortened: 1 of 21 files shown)


indico/modules/attachments/controllers/management/base.py
```diff
@@ -152,7 +152,7 @@ def _process(self):
                 if file:
                     self.attachment.file = AttachmentFile(user=session.user, content_type=file.mimetype,
                                                           filename=secure_filename(file.filename, 'attachment'))
-                    self.attachment.file.save(file.file)
+                    self.attachment.file.save(file.stream)
```
mic4ael: what is the benefit of using `stream` over `file` property?

ThiefMaster: `file` does not exist, we were adding it in this legacy code that shouldn't even have such side-effects: https://github.com/indico/indico/pull/3088/files#diff-d96ee7e7041ebb8b679fc6200b567dcbL82

PR #3089, opened 2017-10-06
Add email hashes to person data

### LINKED ISSUE #3041: HTTP API: hash e-mails in timetable serializer
This is already done in the normal event serializer, there is no reason why we shouldn't do it here too.

mic4ael:
closes #3041

### DIFF
--- a/indico/modules/events/timetable/legacy.py
+++ b/indico/modules/events/timetable/legacy.py
@@ -31,8 +31,9 @@ from indico.web.flask.util import url_for
 
 
 class TimetableSerializer(object):
-    def __init__(self, management=False):
+    def __init__(self, management=False, user=None):
         self.management = management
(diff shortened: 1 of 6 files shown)


indico/modules/events/timetable/legacy.py
```diff
@@ -276,10 +278,11 @@ def _get_location_data(self, obj):
         return data
 
     def _get_person_data(self, person_link):
+        person = person_link.person
         return {'firstName': person_link.first_name,
                 'familyName': person_link.last_name,
                 'affiliation': person_link.affiliation,
-                'email': person_link.person.email,
```
ThiefMaster: should be added back when the user can manage the event (i'd set `self.can_manage_event` in `__init__` to avoid calling it over and over

PR #3091, opened 2017-10-09
Move conference stuff from Mako to Jinja

ThiefMaster:
Depends on changes in #3088; not really worth to review this one until that PR is merged.


indico/legacy/webinterface/pages/conferences.py
```diff
@@ -96,35 +91,62 @@ def render_event_footer(event, dark=False):
...
 
+    @unify_event_args
+    def __init__(self, rh, event_, **kwargs):
+        assert event_ == kwargs.setdefault('event', event_)
+        self.event = event_
+        kwargs['conf_layout_params'] = self._get_layout_params()
+        WPConferenceBase.__init__(self, rh, event_, **kwargs)
+
+    def _get_layout_params(self):
+        bg_color = layout_settings.get(self.event, 'header_background_color').replace('#', '').lower()
```
mic4ael: why are you placing `#` here? you are using it anyway a few lines below

indico/modules/events/templates/display/conference.html
```diff
@@ -0,0 +1,107 @@
...
+                    <div class="address text nohtml">{{ event.address }}</div>
+                {% endif %}
+            </div>
+        {% endif %}
+
+        {% if event.person_links %}
+            <div class="infoline chairs clear">
+                <span title="{% trans %}Chairpersons{% endtrans %}" class="icon icon-user-chairperson"
+                      aria-hidden="true"></span>
+                <ul class="chair_list text">
```
mic4ael: is `chair_list` a legacy class?

indico/legacy/webinterface/pages/conferences.py
ThiefMaster:
```
In [4]: Counter(x.value.startswith('#') for x in EventSetting.query.filter_by(name='header_background_color').filter(db.cast(EventSetting.value, db.Text) != '""
   ...: '))
Out[4]: Counter({False: 20, True: 2075})

In [5]: Counter(x.value.startswith('#') for x in EventSetting.query.filter_by(name='header_text_color').filter(db.cast(EventSetting.value, db.Text) != '""'))
Out[5]: Counter({False: 5, True: 1983})
```

indico/modules/events/templates/display/conference.html
ThiefMaster: yes, and some custom event stylesheets actually use it so i didn't want to change it

PR #3092, opened 2017-10-09
Remove e-mail about double payment

pferreir:
Let's just log it.

### DIFF
--- a/indico/modules/events/payment/models/transactions.py
+++ b/indico/modules/events/payment/models/transactions.py
@@ -233,26 +233,24 @@ class PaymentTransaction(db.Model):
         new_transaction = PaymentTransaction(amount=amount, currency=currency,
                                              provider=provider, data=data)
         registration.transaction = new_transaction
-        double_payment = False
         try:
             next_status = TransactionStatusTransition.next(previous_transaction, action, provider)
         except InvalidTransactionStatus as e:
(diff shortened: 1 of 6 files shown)


indico/modules/events/payment/models/transactions.py
```diff
@@ -253,6 +253,6 @@ def create_next(cls, registration, amount, currency, action, provider=None, data
         except DoublePaymentTransaction:
             next_status = TransactionStatus.successful
             double_payment = True
-            Logger.get('payment').warning("Received successful payment for an already paid registration")
+            Logger.get('payment').info("Received successful payment for an already paid registration")
         new_transaction.status = next_status
         return new_transaction, double_payment
```
ThiefMaster: We could just return the new transaction now

pferreir: I hesitated about removing it, since it may be useful in other places. But yeah, maybe it's best to keep it minimal.

pferreir:
Updated.

indico/modules/events/payment/util_test.py
```diff
@@ -22,23 +22,20 @@
 
 
 @pytest.mark.parametrize(('new', 'double', 'status'), (
```
ThiefMaster: needs to be updated too

pferreir: I actually fixed that in the meantime 😄

PR #3094, opened 2017-10-09
Permissions widget

mvidalgarcia:


indico/htdocs/js/indico/jquery/permissions_widget.js
```diff
@@ -0,0 +1,79 @@
...
+    var permissionClasses = {
+        access: 'accept',
+        timetable: 'highlight',
+        edit: 'danger',
+        submit: 'warning'
+    };
+
+    $.widget('indico.permissionswidget', {
+        options: {
+            a: null,
```
ThiefMaster: 🙈

indico/htdocs/js/indico/jquery/permissions_widget.js
```diff
@@ -0,0 +1,79 @@
...
+                return $labelBox.append($('<span class="flexrow f-a-center">').append($code).append($text));
+            } else {
+                var iconClass = principal.type === 'user' ? 'icon-user' : 'icon-users';
+                var text = principal.type === 'user' ? principal.name : principal.id;
+                return $labelBox.append($('<span class="label-icon text-normal">').addClass(iconClass).text(text));
+            }
+        },
+        _renderPermissions: function(permissions) {
+            var $permissions = $('<div class="permissions-box flexrow f-a-center f-self-stretch">');
+            var $permissionsUl = $('<ul>').appendTo($permissions);
```
ThiefMaster: I think `$permissionsList` is a better name - you'd also call something "fooLink" and not "fooA"

indico/modules/events/management/controllers/protection.py
```diff
@@ -94,10 +94,14 @@ def _get_defaults(self):
         event_session_settings = session_settings.get_all(self.event)
         coordinator_privs = {name: event_session_settings[val] for name, val in COORDINATOR_PRIV_SETTINGS.iteritems()
                              if event_session_settings.get(val)}
+        permissions = [[{'id': 91138, 'name': 'Marco Vidal', 'type': 'user'}, ['edit']],
```
ThiefMaster: :monkey:

indico/web/forms/fields/principals.py
```diff
@@ -119,3 +120,14 @@ def process_formdata(self, valuelist):
...
+
+
+class PermissionsField(JSONField):
+    widget = JinjaWidget('forms/permissions_widget.html', single_kwargs=True, acl=True)
+
+    def __init__(self, *args, **kwargs):
+        super(PermissionsField, self).__init__(*args, **kwargs)
+
+    def _value(self):
+        x = super(PermissionsField, self)._value() if self.data else '[]'
```
ThiefMaster: `return ...`

indico/htdocs/js/indico/jquery/permissions_widget.js
```diff
@@ -30,30 +30,47 @@
...
+                        'color': '#' + principal.color
+                    }
+                });
+
+                return $labelBox.append($('<span>', {class: 'flexrow f-a-center'}).append($code).append($text));
             } else {
                 var iconClass = principal.type === 'user' ? 'icon-user' : 'icon-users';
                 var text = principal.type === 'user' ? principal.name : principal.id;
-                return $labelBox.append($('<span class="label-icon text-normal">').addClass(iconClass).text(text));
+                return $labelBox.append($('<span>', {class: 'label-icon text-normal ' + iconClass, text: text}));
```
mvidalgarcia: Is it really better to add a class this way? Imagine you have to add two or more classes, how would you do?
IMO that space after `text-normal` is quite ugly. More opinions @ThiefMaster @pferreir @meluru?

ThiefMaster: Surrounding space in class attributes is fine. It's read by browsers, not humans ;) We do it all the time in Jinja templates since avoiding it would make it less readable to humans.

Anyway, generating markup from plain JS/jQuery is always ugly. No strong opinion here...

indico/modules/events/management/controllers/protection.py
```diff
@@ -94,11 +134,53 @@ def _get_defaults(self):
...
-                     'own_no_access_contact': self.event.own_no_access_contact}, **coordinator_privs)
+                     'own_no_access_contact': self.event.own_no_access_contact, 'permissions': permissions_data},
+                    **coordinator_privs)
 
     def _update_session_coordinator_privs(self, form):
         data = {field: getattr(form, field).data for field in form.priv_fields}
         update_session_coordinator_privs(self.event, data)
+
+    def _get_principal_permissions(self, principal):
+        """Retrieves a set containing the valid permissions of a principal."""
```
mic4ael: PEP257

indico/htdocs/js/indico/jquery/permissions_widget.js
```diff
@@ -0,0 +1,107 @@
...
+            var $permissions = $('<div>', {class: 'permissions-box flexrow f-a-center f-self-stretch'});
+            var $permissionsList = $('<ul>').appendTo($permissions);
+            permissions.forEach(function(item) {
+                $permissionsList.append($('<li>', {class: 'i-label bold ' + permissionClasses[item]}).append(item));
+            });
+
+            var $permissionsEditBtn = $('<button>', {
+                'type': 'button',
+                'class': 'i-button text-color borderless icon-only icon-edit',
+                'data-href': build_url(Indico.Urls.EventPermissions, {confId: this.options.event_id}),
```
mic4ael: You can put all those `data-` attributes inside an object, like `'data': {'href': '...', 'title': '...}`

indico/modules/events/management/controllers/protection.py
```diff
@@ -94,11 +134,53 @@ def _get_defaults(self):
...
+        permissions = set()
+        valid_labels = self._get_valid_permissions_labels()
+        if principal.full_access:
+            permissions.add('edit')
+        elif principal.read_access:
+            permissions.add('access')
+        return permissions | (set(principal.permissions) & valid_labels)
+
+    def _get_valid_permissions_labels(self):
+        """Retrieves a set containing all the valid permission labels
```
mic4ael: The same as above

indico/htdocs/js/indico/jquery/permissions_widget.js
mvidalgarcia: Apparently this is not possible.

ThiefMaster: That's because `data: {}` sets it as jquery data, but the declarative code actually needs the attributes, so `'data-...': ...` is correct here.

mvidalgarcia:
Not WIP anymore, have fun :tada:

indico/modules/events/management/controllers/protection.py
```diff
@@ -63,42 +70,86 @@ def _process(self):
                                 endpoint='event_management.acl')
 
 
-class RHEventProtection(RHManageEventBase):
+class RHEventProtectionBase(RHManageEventBase):
+    def __init__(self):
+        super(RHEventProtectionBase, self).__init__()
+        selectable_permissions = {k: v for k, v in get_available_permissions(Event).viewitems() if v.user_selectable}
+        self.whitelisted_permissions = set(selectable_permissions.keys()) | {'edit', 'access'}
```
ThiefMaster: calling `.keys()` is pretty much never necessary. Iterating a dict gives you the keys already, so just `set(selectable_permissions)` is enough.

indico/modules/events/management/controllers/protection.py
```diff
@@ -63,42 +70,86 @@ def _process(self):
...
+        self.permissions_tree = {
+            'edit': {
```
ThiefMaster: If this shares scope with actual permissions *anywhere*, please rename it to something like `_edit` or `_full_access` so it doesn't break if someone adds e.g. an 'edit' permission through a plugin.

ThiefMaster: Same for the 'access' permission by the way (`_read_access`)

indico/modules/events/management/controllers/protection.py
```diff
@@ -63,42 +70,86 @@ def _process(self):
...
+                'title': _('Event Management'),
+                'children': {
+                    v.name: {'title': v.friendly_name} for k, v in selectable_permissions.viewitems()
+                }
+            }
+        }
+        self.permissions_tree['edit']['children']['access'] = {'title': _('Access')}
+        self.permissions_tree['edit']['children'] = OrderedDict(
+            sorted(self.permissions_tree['edit']['children'].items(), key=lambda t: t[0])
```
ThiefMaster: `key=itemgetter(0)`, but I think you can completely omit the key function since sorting based on the `(key, value)` tuple works as well (with unique keys the values won't even be looked at)

indico/modules/events/management/controllers/protection.py
```diff
@@ -63,42 +70,86 @@ def _process(self):
                                 endpoint='event_management.acl')
 
 
-class RHEventProtection(RHManageEventBase):
+class RHEventProtectionBase(RHManageEventBase):
+    def __init__(self):
```
ThiefMaster: I would just just write a `get_permissions_info()` function that returns a `whitelisted_permissions, permissions_tree` tuple (btw, isn't `available_permissions` a better name than `whitelisted_permissions`?).

This would also make it easier in the future if we want to use the same widget e.g. for category permissions.

indico/modules/events/management/controllers/protection.py
```diff
@@ -63,42 +70,86 @@ def _process(self):
...
+class RHEventProtection(RHEventProtectionBase):
     """Show event protection"""
 
     NOT_SANITIZED_FIELDS = {'access_key'}
 
     def _process(self):
         form = EventProtectionForm(obj=FormDefaults(**self._get_defaults()), event=self.event)
         if form.validate_on_submit():
+            permission_principals = defaultdict(set)
+            for principal_data in form.permissions.data:
```
ThiefMaster: `for what, ever in form.permissions.data` - like this you have meaningful names instead of `principal_data[0]` and `principal_data[1]`

indico/modules/events/management/controllers/protection.py
```diff
@@ -63,42 +70,86 @@ def _process(self):
...
+                for permission in principal_data[1]:
+                    permission_principals[permission].add(principal_from_fossil(principal_data[0], allow_emails=True,
+                                                                                allow_networks=True))
+            for permission in self.whitelisted_permissions:
+                if permission == 'edit':
```
ThiefMaster: What about adding two module-level constants, `FULL_ACCESS_PERMISSION` and `READ_ACCESS_PERMISSION`?

indico/modules/events/management/controllers/protection.py
```diff
@@ -63,42 +70,86 @@ def _process(self):
...
+                    update_object_principals(self.event, permission_principals.get(permission, set()), full_access=True)
```
ThiefMaster: Don't call `update_object_principals` in a loop. This spams the event log with one entry per call! Instead, build the args (`full_access`, `read_access`, `permissions`) and then call it once. Since `update_object_principals` only supports updating one permission/access bit at a time, you might want to simply add another function that deals with the data you have here, so you'd just call something like `update_permissions(self.event, permission_principals, available_permissions)`

indico/modules/events/management/controllers/protection.py
```diff
@@ -63,42 +70,86 @@ def _process(self):
...
     def _update_session_coordinator_privs(self, form):
         data = {field: getattr(form, field).data for field in form.priv_fields}
         update_session_coordinator_privs(self.event, data)
+
+    def _get_principal_permissions(self, principal):
+        """Retrieve a set containing the valid permissions of a principal."""
+        permissions = set()
+        if principal.full_access:
+            permissions.add('edit')
+        elif principal.read_access:
```
ThiefMaster: Why `elif`? Both can be  set at the same time.

indico/modules/events/management/forms.py
```diff
@@ -228,28 +231,22 @@ class EventProtectionForm(IndicoForm):
...
                                                       'event'))
-    managers = PrincipalListField(_('Managers'), groups=True, allow_emails=True, allow_external=True,
-                                  description=_('List of users allowed to modify the event'))
-    submitters = PrincipalListField(_('Submitters'), groups=True, allow_emails=True, allow_external=True,
-                                    description=_('List of users with submission rights'))
     visibility = SelectField(_("Visibility"), [Optional()], coerce=lambda x: None if x == '' else int(x),
                              description=_("""From which point in the category tree this event will be visible from """
                                            """(number of categories upwards). Applies to "Today's events" and """
                                            """Calendar only. If the event is moved, this number will be preserved."""))
+    permissions = PermissionsField(_("Permissions"))
```
ThiefMaster: I would put this higher up so the field order here matches the order in which we show them in the form.

indico/modules/events/management/forms.py
```diff
@@ -270,6 +267,23 @@ def _init_visibility(self, event):
             self.visibility.warning = _("This event's visibility is currently limited by that of '{}'.").format(
                 real_horizon.title)
 
+    def validate_permissions(self, field):
+        entries = [{
+            'principal': principal_from_fossil(entry[0], allow_emails=True, allow_networks=True),
+            'permissions': entry[1]
+        } for entry in field.data]
```
ThiefMaster: `for foo, bar in field.data` to have more meaningful variable names in the lines above.

indico/modules/events/management/forms.py
```diff
@@ -228,28 +231,22 @@ class EventProtectionForm(IndicoForm):
...
     visibility = SelectField(_("Visibility"), [Optional()], coerce=lambda x: None if x == '' else int(x),
                              description=_("""From which point in the category tree this event will be visible from """
                                            """(number of categories upwards). Applies to "Today's events" and """
                                            """Calendar only. If the event is moved, this number will be preserved."""))
+    permissions = PermissionsField(_("Permissions"))
     priv_fields = set()
 
     def __init__(self, *args, **kwargs):
-        self.protected_object = event = kwargs.pop('event')
+        self.protected_object = event = self.event = kwargs.pop('event')
```
ThiefMaster: Just use `self.event` below.. no need to assign `event` as well. The performance difference is so minuscule that it's not really worth it.

indico/modules/events/management/forms.py
```diff
@@ -270,6 +267,23 @@ def _init_visibility(self, event):
             self.visibility.warning = _("This event's visibility is currently limited by that of '{}'.").format(
                 real_horizon.title)
 
+    def validate_permissions(self, field):
+        entries = [{
+            'principal': principal_from_fossil(entry[0], allow_emails=True, allow_networks=True),
+            'permissions': entry[1]
+        } for entry in field.data]
+        for e in entries:
```
ThiefMaster: `entry`? ;)

indico/modules/events/management/forms.py
```diff
@@ -270,6 +267,23 @@ def _init_visibility(self, event):
...
+            principal = e['principal']
+            permissions = e['permissions']
+            if isinstance(principal, IPNetworkGroup):
+                if permissions != ['access']:
+                    msg = _('IP Networks can only have "access" permission: {}:{}')
```
ThiefMaster: "IP networks cannot have management permissions"

And I think showing both name and ID here is not very pretty - if it's the ID of the network, it's meaningless for a user.

indico/modules/events/management/forms.py
```diff
@@ -270,6 +267,23 @@ def _init_visibility(self, event):
...
+    def validate_permissions(self, field):
+        entries = [{
+            'principal': principal_from_fossil(entry[0], allow_emails=True, allow_networks=True),
+            'permissions': entry[1]
+        } for entry in field.data]
+        for e in entries:
+            principal = e['principal']
+            permissions = e['permissions']
+            if isinstance(principal, IPNetworkGroup):
+                if permissions != ['access']:
```
ThiefMaster: While this works, I think using set logic would be nicer here: `if set(permissions) - {'access'}:`

indico/modules/events/management/forms.py
ThiefMaster: Also, if there are multiple placeholders in a translated string, *always* use the numbered or named version. In this particular case it may not matter much since `{}:{}` is meaningless enough so no translator would think about reordering it, but in other cases it may be necessary and for that the placeholders should be distinct and not use implicit positions. IMO the best choice is to always used named placeholders as soon as there's more than one in a translated string.

indico/modules/events/management/forms.py
```diff
@@ -270,6 +267,23 @@ def _init_visibility(self, event):
...
+                    msg = _('IP Networks can only have "access" permission: {}:{}')
+                    raise ValidationError(msg.format(principal.name, principal.id))
+            if 'edit' in permissions and len(permissions) != 1:
+                principal_id = principal.email if isinstance(principal, EmailPrincipal) else principal.id
+                msg = _('"Edit" permission must be unique: {}:{}')
```
ThiefMaster: see the comments above

indico/web/forms/fields/principals.py
```diff
@@ -21,15 +21,32 @@
...
 from indico.modules.networks.util import serialize_ip_network_group
 from indico.modules.users.util import serialize_user
 from indico.util.user import principal_from_fossil
+from indico.web.forms.fields import JSONField
 from indico.web.forms.widgets import JinjaWidget
 
 
+def serialize_principal(principal):
+    if principal.principal_type == PrincipalType.email:
+        return principal.fossilize()
```
ThiefMaster: Let's avoid calling legacy stuff here! Eventually we want to get rid of fossilize. You could just add a `serialize_email_principal` function and maybe even see if you can get rid of the fossilization stuff in `EmailPrincipal` (it seems to be used in just one other place).

ThiefMaster: Actually it looks like the "one other place" is the one you removed just below ;)

indico/web/forms/fields/principals.py
```diff
@@ -119,3 +124,22 @@ def process_formdata(self, valuelist):
...
+    @property
+    def event(self):
+        return self.get_form().event
+
+    @property
+    def roles(self):
+        return [serialize_role(role) for role in self.get_form().event.roles]
+
+    def _value(self):
+        return json.dumps(self.data, separators=(',', ':'), ensure_ascii=False).encode('utf-8') if self.data else '[]'
```
ThiefMaster: Why `separators=(',', ':'), ensure_ascii=False`?

indico/web/templates/forms/permissions_widget.html
```diff
@@ -0,0 +1,52 @@
+{% extends 'forms/base_widget.html' %}
+
+{% block html %}
+    <h3>Permissions</h3>
+    <div id="permissions-widget-{{ field.id }}" class="permissions-widget">
+        {% set event = field.event %}
+        <input type="hidden" value='{{ field._value() }}' name="permissions">
```
ThiefMaster: `{{ field.name }} `

indico/web/templates/forms/permissions_widget.html
```diff
@@ -0,0 +1,52 @@
...
+                </a>
+            </div>
+        </div>
+    </div>
+{% endblock %}
+
+{% block javascript %}
+    <script>
+        $('#permissions-widget-{{ field.id }}').permissionswidget({
+            event_id: {{ field.event.id }}
```
ThiefMaster: `eventId` (see one of the comments above)

ThiefMaster:
> Not WIP anymore, **have fun** :tada: 

Done. :trollface:

indico/web/forms/fields/principals.py
mvidalgarcia: I was having troubles with characters like 'ö' but I will change it.

ThiefMaster: Maybe a bytestring was passed somewhere?

indico/web/templates/forms/permissions_widget.html
```diff
@@ -0,0 +1,52 @@
+{% extends 'forms/base_widget.html' %}
+
+{% block html %}
+    <h3>Permissions</h3>
+    <div id="permissions-widget-{{ field.id }}" class="permissions-widget">
+        {% set event = field.event %}
+        <input type="hidden" value='{{ field._value()|tojson }}' name="permissions">
```
ThiefMaster: `name="{{ field.name }}'`

ThiefMaster: When using `|tojson` outside a script tag, you always need to use `|forceescape` as well

indico/web/templates/forms/permissions_widget.html
```diff
@@ -0,0 +1,52 @@
...
+        <ul class="permissions-widget-list"></ul>
+        <div class="toolbars">
+            <div class="toolbar">
+                <div class="group">
+                    <a href="#" class="i-button label text-color outline add-entry-decoration">
+                        {%- trans %}Add{% endtrans -%}
+                    </a>
+                    <a class="i-button js-add-user-group">{% trans %}User / Group{% endtrans %}</a>
+                    <a class="i-button arrow js-dropdown" data-toggle="dropdown">{% trans %}Role{% endtrans %}</a>
+                    <ul class="dropdown entry-role-dropdown" data-items='{{ field.roles|tojson }}'>
```
ThiefMaster: `|forceescape`

indico/web/templates/forms/permissions_widget.html
```diff
@@ -0,0 +1,52 @@
...
+                                <span class="icon-medal dropdown-icon"></span>
+                                <span>{% trans %}New role{% endtrans %}</span>
+                            </a>
+                        </li>
+                    </ul>
+                    {% if field.ip_networks %}
+                        <a class="i-button arrow js-dropdown" data-toggle="dropdown">
+                            {%- trans %}IP Network{% endtrans -%}
+                        </a>
+                        <ul class="dropdown entry-ip-network-dropdown" data-items='{{ field.ip_networks|tojson }}'></ul>
```
ThiefMaster: `|forceescape`

indico/modules/events/management/templates/event_permissions_dialog.html
```diff
@@ -0,0 +1,53 @@
+<form id="js-permissions-dialog-form">
+    <input type="hidden" id="selected_permissions" value='{{ permissions|tojson }}' name="selected_permissions">
```
ThiefMaster: `|forceescape`

indico/modules/events/management/templates/event_permissions_dialog.html
```diff
@@ -0,0 +1,53 @@
...
+    <input type="hidden" id="principal" value='{{ principal|tojson }}'>
```
ThiefMaster: `|forceescape`

indico/htdocs/js/indico/jquery/permissions_widget.js
```diff
@@ -0,0 +1,327 @@
...
+                });
+            });
+        },
+        _renderItem: function(item) {
+            var $item = $('<li>', {class: 'flexrow f-a-center'});
+            var principal = item[0];
+            var permissions = item[1];
+            $item.append(this._renderLabel(principal));
+            $item.append(this._renderPermissions(principal, permissions));
+            $item.toggleClass('disabled ' + principal.id, principal._type === 'DefaultEntry');
```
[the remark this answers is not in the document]
mvidalgarcia: In this case it's just for default entries, so the ids are `category-managers` and `anonymous`. I'm adding them as class because it later allows me to select the anonymous entry to change its permissions whenever the protection mode is changed.

indico/htdocs/js/indico/jquery/permissions_widget.js
```diff
@@ -0,0 +1,327 @@
...
+            } else {
+                $dropdownLink.removeClass('disabled');
+            }
+        },
+        _renderDropdownItem: function(principal) {
+            var self = this;
+            var $dropdownItem = $('<li>', {'class': 'entry-item', 'data-principal': JSON.stringify(principal)});
+            var $itemContent = $('<a>');
+            if (principal._type === 'EventRole') {
+                var $code = this._renderRoleCode(principal.code, principal.color).addClass('dropdown-icon');
```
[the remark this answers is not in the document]
mvidalgarcia: It fits.

indico/modules/events/management/controllers/protection.py
mvidalgarcia: I don't understand how I can do it in one single call. As you said, `update_object_principals` only supports updating one permission/access bit at a time, so how can I update all the permissions (since all of than might have changed when saving) in a unique call? Don't you need one call per permission (what I'm doing now)?

ThiefMaster: There is `event.update_principal()` where you pass the principal (user/group/etc.) and the list of roles to add/remove/set and the read/manage privs. You can call this directly, so it's one call per user no matter how many permissions.

ThiefMaster: Deleted entries need to be handled separately (like we probably do right now); either by removing all privileges via `update_principal` or by calling `delete_principal`

mvidalgarcia:
Updated.

indico/modules/events/management/controllers/protection.py
```diff
@@ -71,34 +101,85 @@ class RHEventProtection(RHManageEventBase):
...
+        available_permissions = get_permissions_info()[0]
+        return permissions | (set(principal.permissions) & set(available_permissions))
+
+    def _update_permissions(self, current, new):
+        """Handle the updates of permissions and creations/deletions of acl principals.
+
+        :param current: A dict mapping principals and a set with its current permissions
+        :param new: A dict mapping principals and a set with its new permissions
+        """
+        def _get_splitted_permissions(_permissions):
```
ThiefMaster: does this need to be nested inside the method? also, the `_permissions` argument name is a bit weird IMO

indico/modules/events/management/controllers/protection.py
```diff
@@ -71,34 +101,85 @@ class RHEventProtection(RHManageEventBase):
...
+            if full_access_permission:
+                _permissions.remove(FULL_ACCESS_PERMISSION)
+            read_access_permission = READ_ACCESS_PERMISSION in _permissions
+            if read_access_permission:
+                _permissions.remove(READ_ACCESS_PERMISSION)
+            return full_access_permission, read_access_permission, _permissions
+
+        for principal, permissions in current.viewitems():
+            if principal not in new:
+                self.event.update_principal(principal, False, False, del_permissions=current[principal])
```
ThiefMaster: Anything besides the first argument should be a kwarg.

Can't wait for Python 3 to enforce stuff like this! ;)

ThiefMaster: Also, you are removing someone's access here, we have `remove_principal()` for exactly this purpose. Just FYI, if you really needed to use `update_principal()` you could just pass `permissions=set()` to remove them all instead of explicitly specifying what to remove...

indico/modules/events/management/controllers/protection.py
```diff
@@ -71,34 +101,85 @@ class RHEventProtection(RHManageEventBase):
...
+            elif permissions != new[principal]:
+                full_access, read_access, new_permissions = _get_splitted_permissions(new[principal])
+                current_permissions = current[principal] - {FULL_ACCESS_PERMISSION, READ_ACCESS_PERMISSION}
+                add_permissions = new_permissions - current_permissions
+                del_permissions = current_permissions - new_permissions
+                self.event.update_principal(principal, read_access, full_access, add_permissions=add_permissions,
```
ThiefMaster: If you want to overwrite the whole permission set, just use `permissions=...` instead of `<add|del>_permissions=...`

indico/modules/events/management/controllers/protection.py
```diff
@@ -71,34 +101,85 @@ class RHEventProtection(RHManageEventBase):
...
+                                            del_permissions=del_permissions)
+        new_principals = set(new) - set(current)
+        for p in new_principals:
+            full_access, read_access, permissions = _get_splitted_permissions(new[p])
+            self.event.update_principal(p, read_access, full_access, add_permissions=permissions)
```
ThiefMaster: kwargs

indico/modules/events/management/controllers/protection.py
```diff
@@ -71,34 +101,85 @@ class RHEventProtection(RHManageEventBase):
...
+
+
+class RHEventPermissionsDialog(RHManageEventBase):
+    def _process(self):
+        principal = json.loads(request.args['principal'])
```
ThiefMaster: JSON in the query string is a bit ugly ;) I'd consider using a POST request... You just need to add `data-method=POST` and the dialog will be loaded with POST instead of GET.

indico/modules/events/management/forms.py
```diff
@@ -270,6 +267,22 @@ def _init_visibility(self, event):
...
+        } for principal, permissions in field.data]
+        for entry in entries:
+            principal = entry['principal']
+            permissions = entry['permissions']
+            if isinstance(principal, IPNetworkGroup):
+                if set(permissions) - {'_read_access'}:
+                    msg = _('IP networks cannot have management permissions: {}')
+                    raise ValidationError(msg.format(principal.name))
+            if '_full_access' in permissions and len(permissions) != 1:
+                msg = _('"Manage" permission must be unique: {}')
```
ThiefMaster: I remember discussing it but don't remember what we decided in the end... so we always remove explicit permissions once you make someone a full manager?

In any case, assuming this is already done on the client, I don't see a need to create a translatable error message here. You could even just discard these permissions instead of failing.

indico/modules/events/management/templates/event_protection.html
```diff
@@ -1,11 +1,12 @@
...
-{% from 'forms/_form.html' import form_header, form_footer, form_rows, form_fieldset %}
+{% from 'forms/_form.html' import form_header, form_footer, form_row, form_rows, form_fieldset %}
 
 {% block title %}{% trans %}Event protection{% endtrans %}{% endblock %}
 
 {% block content %}
     {{ form_header(form, id='event-protection-form') }}
-    {{ form_rows(form, skip=form.priv_fields,
+    {{ form_row(form.permissions, skip_label=true) }}
+    {{ form_rows(form, skip=('priv_fields', 'permissions'),
```
ThiefMaster: That doesn't do what you think it does!

`form.priv_fields` is a set containing the names of fields to skip. So you want `skip=form.priv_fields.union([permissions])` which is super ugly, but you don't have access to the `|` operator in Jinja.

indico/modules/events/roles/controllers.py
```diff
@@ -67,7 +68,7 @@ def _process(self):
             logger.info('Event role %r created by %r', role, session.user)
             self.event.log(EventLogRealm.management, EventLogKind.positive, 'Roles',
                            'Added role: "{}"'.format(role.name), session.user)
-            return jsonify_data(html=_render_roles(self.event))
+            return jsonify_data(html=_render_roles(self.event), role=serialize_principal(role))
```
ThiefMaster: You could use `serialize_role()` here directly.

indico/util/user.py
```diff
@@ -96,6 +97,12 @@ def principal_from_fossil(fossil, allow_pending=False, allow_groups=True, allow_
         if group.group is None and not allow_missing_groups:
             raise ValueError('Multipass group does not exist: {}:{}'.format(provider, id_))
         return group
+    elif type_ == 'EventRole':
+        role = EventRole.get(id_)
```
ThiefMaster: Do we ensure the role belongs to the same event somewhere? Maybe we should add an optional `event` arg to this function and only accept event roles if it's not None (like with the `allow_*` args) and the role belongs to the same event...

indico/modules/events/management/controllers/protection.py
mvidalgarcia: The problem is that this user can have other permissions not involved in the Protection page like for instance `paper_layout_reviewer` that we don't want to remove.

ThiefMaster: ok, valid point. maybe we could build a list containing such permissions as well? something like this:

    new_permissions = (existing_permissions & not_user_editable_permissions)

Like this everything that's not exposed in the permissions field is kept.

In the case below (updating instead of removing), you could do something like this:

    new_permissions = (existing_permissions & not_user_editable_permissions) | new_permissions

indico/modules/events/management/templates/event_protection.html
mvidalgarcia: Why isn't the way I'm doing valid? I guess it's just a matter of the stuff to skip into `priv_fields`?

ThiefMaster: Because there is no field named `priv_fields` - this is a set containing field names! So with your code, it skips `permissions` and that's it. since there is no field named `priv_fields`. But a conference event has extra fields for session coordinator privileges.

mvidalgarcia:
Updated.

ThiefMaster:
The description of a role isn't shown anywhere

indico/htdocs/js/indico/jquery/permissions_widget.js
```diff
@@ -0,0 +1,336 @@
...
+            var self = this;
+            return $('<button>', {
+                'type': 'button',
+                'class': 'i-button text-color borderless icon-only icon-remove',
+                'data-principal': JSON.stringify(principal)
+            }).on('click', function() {
+                var $this = $(this);
+                var title = $T.gettext("Delete entry '{0}'".format(principal.name || principal.id));
+                var message = $T.gettext("Are you sure you want to permanently delete this entry?");
+                confirmPrompt(message, title).then(function() {
```
ThiefMaster: A confirm prompt here is rather annoying, also the wording is somewhat unfortunate: "permanently delete" is something that usually implies data loss (like deleting an event), not something that can be restored by the user within seconds.

ThiefMaster: What we *do* need to restore here however is the confirmation prompt when you remove yourself from the ACL or remove your management rights. We had this before and I don't think it's completely obvious for a user that you, as the event creator, do not have any access or management rights just because you created the event.

indico/modules/events/management/forms.py
```diff
@@ -270,6 +268,20 @@ def _init_visibility(self, event):
             self.visibility.warning = _("This event's visibility is currently limited by that of '{}'.").format(
                 real_horizon.title)
 
+    def validate_permissions(self, field):
+        event = self.event
+        for idx, value in enumerate(field.data):
```
ThiefMaster: for principal_fossil, permissions in field.data:

(also see the comment a few lines below to remove the need for having the index)

---

just fyi, you could do this if you needed to keep the index:

    for idx, (principal_fossil, permissions) in enumerate(field.data):

indico/modules/events/management/forms.py
```diff
@@ -270,6 +268,20 @@ def _init_visibility(self, event):
...
+            principal_fossil = value[0]
+            permissions = value[1]
+            principal = principal_from_fossil(principal_fossil, allow_emails=True, allow_networks=True, event=event)
+            if isinstance(principal, IPNetworkGroup):
+                if set(permissions) - {READ_ACCESS_PERMISSION}:
+                    msg = _('IP networks cannot have management permissions: {}')
+                    raise ValidationError(msg.format(principal.name))
+            if FULL_ACCESS_PERMISSION in permissions and len(permissions) != 1:
+                # when full access permission is set, discard rest of permissions
+                field.data[idx][1] = [FULL_ACCESS_PERMISSION]
```
ThiefMaster: You can update the list in-place:

    permissions[:] = [FULL_ACCESS_PERMISSION]

indico/web/templates/forms/permissions_widget.html
```diff
@@ -0,0 +1,55 @@
...
+                        </a>
+                        <ul class="dropdown entry-ip-network-dropdown"
+                            data-items='{{ field.ip_networks|tojson|forceescape }}'></ul>
+                    {% endif %}
+                </div>
+            </div>
+            <div class="toolbar">
+                <a href="{{ url_for('event_roles.manage', event) }}"
+                   class="i-button icon-medal text-color borderless">
+                    {%- trans %}Manage roles{% endtrans -%}
```
ThiefMaster: IIRC. we wanted to remove this button. It's not very useful now that you can add roles from the dropdown, and clicking it results in you losing your changes in the form (which is not obvious beforehand).

indico/modules/events/management/forms.py
```diff
@@ -270,6 +268,20 @@ def _init_visibility(self, event):
...
                 real_horizon.title)
 
+    def validate_permissions(self, field):
+        event = self.event
+        for idx, value in enumerate(field.data):
+            principal_fossil = value[0]
+            permissions = value[1]
+            principal = principal_from_fossil(principal_fossil, allow_emails=True, allow_networks=True, event=event)
+            if isinstance(principal, IPNetworkGroup):
+                if set(permissions) - {READ_ACCESS_PERMISSION}:
```
ThiefMaster: if isinstance(principal, IPNetworkGroup) and set(permissions) - {READ_ACCESS_PERMISSION}:

mvidalgarcia:
Updated.

ThiefMaster:
Would it make sense to have the  "Access" entry on the same level as "Manage" instead of inside? Right now I think it could easily be misunderstoot as "Manage access" and not "Access the event" (unless you read the tooltip, but probably not everyone does)

indico/core/permissions.py
```diff
@@ -55,3 +66,37 @@ def check_permissions(type_):
...
+        READ_ACCESS_PERMISSION: {'title': _('Access'), 'css_class': 'accept',
+                                 'description': _('Access the public areas of the event')}
+    }
+    permissions_tree = {
+        FULL_ACCESS_PERMISSION: {
+            'title': special_permissions[FULL_ACCESS_PERMISSION]['title'],
+            'description': special_permissions[FULL_ACCESS_PERMISSION]['description'],
+            'children': {
+                v.name: {'title': v.friendly_name, 'description': v.description}
+                for k, v in selectable_permissions.viewitems()
```
ThiefMaster: `for name, perm in ...`?

indico/htdocs/js/indico/jquery/permissions_widget.js
```diff
@@ -0,0 +1,342 @@
...
+/* global ChooseUsersPopup:false, Palette:false */
+
+(function($) {
+    'use strict';
+
+    $.widget('indico.permissionswidget', {
+        options: {
+            eventId: null,
+            permissionsInfo: null,
+            userId: null
```
ThiefMaster: not needed, you can get it using`$('body').data('user-id')`

indico/modules/events/management/templates/event_permissions_dialog.html
```diff
@@ -0,0 +1,61 @@
...
+            {% for code, tree_elem in permissions_tree.viewitems() recursive %}
+                <li class="item">
+                    <input type="checkbox" id="cb-{{ code }}"
+                           class="circle-checkbox" {{ 'checked' if code in permissions or code == '_read_access' }}
+                           {{ 'disabled' if code == '_read_access' }}
+                           data-code="{{ code }}">
+                    <label for="cb-{{ code }}" title="{{ tree_elem.description }}">{{ tree_elem.title }}</label>
+                    {% if tree_elem.children %}
+                        <ul class="permissions">
+                            {{ loop(tree_elem.children|dictsort(by='key')) }}
```
ThiefMaster: `by='key'` is the default

indico/htdocs/js/indico/jquery/permissions_widget.js
```diff
@@ -0,0 +1,342 @@
...
+            }).on('click', function() {
+                var $this = $(this);
+                var title = $T.gettext("Delete entry '{0}'".format(principal.name || principal.id));
+                if (principal._type === 'Avatar' && principal.id === self.options.userId) {
+                    var message = $T.gettext("Are you sure you want to remove yourself from the list?");
+                    confirmPrompt(message, title).then(function() {
+                        self._updateItem($this.data('principal'), []);
+                    });
+                } else {
+                    self._updateItem($this.data('principal'), []);
```
ThiefMaster: You could avoid having this line twice using something like this (copied from the old place where we had this check):

                if (currentUserId === userId) {
                    confirmed = confirmPrompt($T.gettext('Are you sure you want to remove yourself from the list?'),
                                              $T.gettext('Confirm action'));
                } else {
                    confirmed = $.Deferred().resolve();
                }

                cnonfirmed.then(...)

indico/modules/events/management/templates/event_permissions_dialog.html
```diff
@@ -0,0 +1,61 @@
...
+    </div>
+</form>
+<script>
+    (function () {
+        'use strict';
+
+        function updateHiddenField() {
+            var data = [];
+            var $form = $('#js-permissions-dialog-form');
+            $form.find('[type=checkbox]:checked').each(function() {
```
ThiefMaster: *Way* too generic. Should be restricted to some container related to the permission dialog.

Also, use `input:checkbox` or `input[type=checkbox]` to select checkboxes.

mvidalgarcia:
> @ThiefMaster: Would it make sense to have the "Access" entry on the same level as "Manage" instead of inside? Right now I think it could easily be misunderstoot as "Manage access" and not "Access the event" (unless you read the tooltip, but probably not everyone does)

Either that or using a more meaningful friendly name like "Access the event" itself.

PS: I really hate GitHub no supporting inline replies :rage:

indico/modules/events/management/templates/event_permissions_dialog.html
mvidalgarcia: Isn't it already contained in the dialog?
```
var $form = $('#js-permissions-dialog-form');
```

ThiefMaster: Yes, indeed. But I think somewhere below there's one where you have `$('[type=checkbox]')`

mvidalgarcia:
Updated

ThiefMaster:
> When "manage" is enabled, I cannot toggle any other permission (which is correct), but clicking one of the circles still toggles its visual state

Better now, but still shows a hover effect.

indico/modules/events/__init__.py
```diff
@@ -191,6 +191,8 @@ class SubmitterPermission(ManagementPermission):
     name = 'submit'
     friendly_name = _('Submission')
     description = _('Grants access to materials and minutes.')
```
ThiefMaster: Let's clarify this, the current one is misleading since it sounds like you may need it to download these things.

How about this instead? "Grants management rights for event-wide materials and minutes"

If you change it, let's update the one for contributions (`SubmitterPermission` in `modules/events/contributions/__init__.py`) as well even though we currently do not show it anywhere:

"Grants management rights for materials and minutes." (since this would always be shown in the context of editing contribution permissions, it's probably not necessary to mention something like "associated with the contribution" there; that should be obvious)

PR #3096, opened 2017-10-13
Cern access plugin

meluru:


indico/htdocs/sass/partials/_icons.scss
```diff
@@ -94,445 +94,448 @@ i[class^="icon-"]:before {
 
 /* Icomoon's icon definitions */
 .icon-agreement:before {
-    content: "\e605";
+  content: "\e605";
```
ThiefMaster: select the whole generated block and let pycharm reformat it to keep the 4-space indentation

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -206,16 +208,21 @@ def _send_emails(self, form):
...
 
     def _process(self):
         tpl = get_template_module('events/registration/emails/custom_email_default.html')
         default_body = tpl.get_html_body()
         registration_ids = request.form.getlist('registration_id')
         form = EmailRegistrantsForm(body=default_body, regform=self.regform, registration_id=registration_ids,
-                                    recipients=[x.email for x in self.registrations])
+                                    recipients=[x.email for x in self.registrations], )
+        if self.regform.tickets_enabled:
+            setattr(EmailRegistrantsForm, 'attach_ticket', BooleanField(_('Attach ticket'), widget=SwitchWidget(),
```
ThiefMaster: ⚠️ This modifies the form class and will affect ALL EVENTS (for requests handled by the same process).

You need to subclass the form class on the fly:

    form_class = type(b'_EmailRegistrantsForm', (EmailRegistrantsForm,), {})

and then set `form_class.attach_ticket = ...` (no need to ever use setattr with a non-dynamic second argument) and use `form = form_class(...)`

---

However, You could just add the field to the form class like any other field and `del form.attach_ticket` if tickets are *not* enabled. That's much easier and thus probably the better solution here.

indico/modules/events/registration/lists.py
```diff
@@ -178,11 +178,13 @@ def get_list_kwargs(self):
         reg_list_config = self._get_config()
         registrations_query = self._build_query()
         total_entries = registrations_query.count()
+        regform = self.regform
         registrations = self._filter_list_entries(registrations_query, reg_list_config['filters']).all()
         dynamic_item_ids, static_item_ids = self._split_item_ids(reg_list_config['items'], 'dynamic')
         static_columns = self._get_static_columns(static_item_ids)
         regform_items = self._get_sorted_regform_items(dynamic_item_ids)
         return {
+            'regform': regform,
```
ThiefMaster: why not just use `self.regform` here? ;)

indico/modules/events/registration/templates/management/_reglist.html
```diff
@@ -19,8 +19,10 @@
                     <thead>
                         <tr class="i-table">
                             <th class="i-table thin-column"></th>
+                            {{ template_hook('registration-status-flag', regform=regform, header=True) }}
```
ThiefMaster: lowercase `true`

indico/modules/events/requests/controllers.py
```diff
@@ -122,18 +120,13 @@ def process_form(self):
         else:
             req = self.request
             new = False
-        try:
-            self.definition.send(req, form.data)
-        except RequestModuleError:
```
ThiefMaster: I would keep this. Could be useful for others implementing requests that need to call external services.

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -206,16 +208,20 @@ def _send_emails(self, form):
...
+                               attachments=attachments)
             send_email(email, self.event, 'Registration')
 
     def _process(self):
         tpl = get_template_module('events/registration/emails/custom_email_default.html')
         default_body = tpl.get_html_body()
         registration_ids = request.form.getlist('registration_id')
         form = EmailRegistrantsForm(body=default_body, regform=self.regform, registration_id=registration_ids,
-                                    recipients=[x.email for x in self.registrations])
+                                    recipients=[x.email for x in self.registrations], )
```
mic4ael: is that change necessary?

ThiefMaster: no, and it's rather misleading (one might first think you are using a tuple containing a list here). probably a leftover from some argument htat was there before

indico/htdocs/sass/partials/_icons.scss
```diff
@@ -96,441 +96,591 @@ i[class^="icon-"]:before {
 .icon-agreement:before {
     content: "\e605";
 }
+
```
ThiefMaster: i wouldn't add all these empty lines... it's a bit noisy considering it's just generated code anyway

indico/modules/events/registration/templates/management/_reglist.html
```diff
@@ -19,8 +19,10 @@
                     <thead>
                         <tr class="i-table">
                             <th class="i-table thin-column"></th>
+                            {{ template_hook('registration-status-flag', regform=regform, header=true) }}
```
ThiefMaster: i would add `registration=none` here

indico/modules/events/registration/templates/management/_reglist.html
```diff
@@ -38,11 +40,12 @@
                                            value="{{ registration.id }}"
                                            data-has-files="{{ registration.has_files | tojson }}">
                                 </td>
+                                {{ template_hook('registration-status-flag', regform=regform, registration=registration) }}
```
ThiefMaster: `header=false` - like this we always call the hook with the same list of arguments

indico/modules/events/requests/controllers.py
```diff
@@ -189,11 +189,12 @@ def _check_access(self):
     def _process(self):
         if self.request.state not in {RequestState.pending, RequestState.accepted}:
             raise BadRequest
-        try:
```
ThiefMaster: This doesn't look correct. the `else` below is now part of the `if` (doesn't make much sense, the if body raises an exception) and the "You have witdrawn" message is flashed even if there was a RequestModuleError.

The previous code looked fine; all that was missing was the `self.commit = False`..

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -206,8 +208,10 @@ def _send_emails(self, form):
             template = get_template_module('events/registration/emails/custom_email.html',
                                            email_subject=form.subject.data, email_body=email_body)
             bcc = [session.user.email] if form.copy_for_sender.data else []
+            attachments = get_ticket_attachment(registration) if form.attach_ticket.data else None
```
ThiefMaster: i think this might fail when ticketing is not enabled since you remove the field from the form. if it does fail, you need to check `'attach_ticket' in form and form.attach_ticket.data`

meluru:
Updated

indico/modules/events/registration/util.py
```diff
@@ -476,3 +476,11 @@ def get_event_regforms(event, user, with_registrations=False):
                                                       Registration.user == user,
                                                       Registration.is_active))
     return query.all()
+
+
+def get_ticket_attachment(registration):
```
ThiefMaster: Since this returns a list, wouldn't it be clearer to call this `get_ticket_attachments` (even though it returns only one attachment)?

Otherwise I'd return it without the list around it and wrap it into a list when calling it.

indico/modules/events/registration/util.py
```diff
@@ -476,3 +476,11 @@ def get_event_regforms(event, user, with_registrations=False):
...
+    from indico.modules.events.registration.controllers.management.tickets import generate_ticket
+    return [{
```
ThiefMaster: https://github.com/indico/indico/pull/3121/files#diff-47791ddd136c52d5259f676b675cac37R76

indico/modules/events/registration/util.py
```diff
@@ -476,3 +476,8 @@ def get_event_regforms(event, user, with_registrations=False):
                                                       Registration.user == user,
                                                       Registration.is_active))
     return query.all()
+
+
+def get_ticket_attachments(registration):
+    from indico.modules.events.registration.controllers.management.tickets import generate_ticket
```
ThiefMaster: Wouldn't this function be much better suited here in util?

meluru:
Updated

PR #3117, opened 2017-10-22
Add site name in <title>

### LINKED ISSUE #3018: Use configurable site title in <title>-
We have the setting and currently it's used only in opengraph metadata for events... It would probably make sense to use it in the `<title>` too.

nop33:
Closes #3018

Example when site name is the default ("Indico")
> Indico [Home] - Management area &middot; Indico

Example when site name is not "Indico":
> Indico [Home] - Management area &middot; Indico - My Awesome Organization

Current behaviour:

> Indico [Home] - Management area

### DIFF
--- a/indico/web/templates/indico_base.html
+++ b/indico/web/templates/indico_base.html
@@ -4,7 +4,9 @@
 <html prefix="og: http://ogp.me/ns#"
       data-static-site="{{ g.get('static_site', false)|tojson|forceescape }}">
 <head>
-    <title>{{ page_title }}</title>
+    <title>
+      {{- page_title }} &middot; {% if site_name != 'Indico' %}Indico - {% endif %}{{ site_name -}}
+    </title>
     <meta charset="UTF-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="csrf-token" id="csrf-token" content="{{ session.csrf_token }}">


ThiefMaster:
I don't think it makes sense to have "Indico" there twice. I would rather replace the default Indico (in one of the WP classes) with "Indico - SITE_NAME` if the site_name isn't "Indico".

nop33:
But I think it's more of a common practice to have this format:

{{Primary Keyword}} - {{Secondary Keyword}} | {{Brand Name}}
*8-foot Green Widgets - Widgets & Tools | Widget World*
or
*Home - {{site_name}} · Indico*

ThiefMaster:
ok fair enough. but that doesn't play well with the code that uses "Indico" as a default page title: https://github.com/indico/indico/blob/master/indico/web/views.py#L149

FWIW, maybe it'd be a good idea to move the "brand" title (eg "Indico - SITENAME or "Indico") into the template and use the title provided by the WP only for the part specific to your current location..

nop33:
So you mean something like 
WP.._title - BRAND,
where BRAND is either "Indico" or "Indico - SITENAME" ?

Imo, this would be the best format:
CURRENT LOCATION - SITE NAME · Indico

ThiefMaster:
Basically yes. not sure about the best format though.. I wonder if it's even important to show both Indico *and* the site name there. If I have many tabs open I don't care so much about the fact that something is an indico instance. If there's e.g. a tab with "ILC Agenda" in the title (be it "Home [ILC Agenda]" or "ILC Agenda [Home]") I know what it is. "Home [ILC Agenda] · Indico" doesn't look that bad either though, but it also doesn't add much valuable information.

These are the titles we have right now

```
Indico [CATEGORY TITLE]
Indico [CATEGORY TITLE] - ​Management area​
Indico - ​Management area
Indico - ​Administration area
Indico - Room Booking
```

some options could be

only the site title:

```
CATEGORY TITLE [Indico]
CATEGORY TITLE - ​Management area​ [Indico]
​Management area [Indico]
​Administration area [Indico]
Room Booking [Indico]
```

e.g.

```
CATEGORY TITLE [ILC Agenda]
CATEGORY TITLE - ​Management area​ [ILC Agenda]
​Management area [ILC Agenda]
​Administration area [ILC Agenda]
Room Booking [ILC Agenda]
```

or both:

```
CATEGORY TITLE [ILC Agenda] · Indico
CATEGORY TITLE - ​Management area​ [ILC Agenda] · Indico
​Management area [ILC Agenda] · Indico
​Administration area [ILC Agenda] · Indico
Room Booking [ILC Agenda] · Indico
```



@pferreir what do you think?

nop33:
I personally think it's good to put at the end of the title the product
name (Indico) in any case. It's the name of the software and I think it's
worth being mentioned. Also, it won't be visible in the tab title either if
it'll be at the end.
As for the brackets, I am not such a big fan, but I like the last
suggestion you made (using both site title and Indico at the end).

pferreir:
What about... `Management Area | Indico Workhop 2.0 · CERN · Indico`
Not sure about the separators.

pferreir:
> Adding the organization name there? Some current instances have very long names that would make the title quite ugly IMO

Yeah, maybe it's a bit too much.

ThiefMaster:
Superseded by #3130

PR #3118, opened 2017-10-23
Fix wording in session block adjustment dialog

### LINKED ISSUE #3080: Session block adjustment: incorrect message wording
When ajusting a session block start/end time based on the session block contents, the message incorrectly says that the *session* start/end time will be adjusted instead of the *session block*. Also the expected session block name displayed is the session name (consistent with the wrong wording!) where it should be `session:block`. And there is also a missing space after the session (block) name.

This is confusing but probably easy to fix!

mvidalgarcia:
Fixes #3080

### DIFF
--- a/indico/htdocs/js/indico/modules/timetable/timetable/Management.js
+++ b/indico/htdocs/js/indico/modules/timetable/timetable/Management.js
@@ -335,7 +335,11 @@ type("RescheduleDialog", ["ExclusivePopupWithButtons"], {
      * For an interval timetable, returns the title of the interval
      */
     __getIntervalTitle: function(){
-        return '"' + this.tt.contextInfo.title + '"';
+        if (this.tt.contextInfo.slotTitle) {
+            return '"{0}: {1}"'.format(this.tt.contextInfo.title, this.tt.contextInfo.slotTitle);
+        } else {
+            return '"{0}"'.format(this.tt.contextInfo.title);
+        }
     },
 
     /**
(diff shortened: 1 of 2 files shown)


indico/htdocs/js/indico/modules/timetable/timetable/Management.js
```diff
@@ -682,19 +682,19 @@ type("RescheduleDialog", ["ExclusivePopupWithButtons"], {
 type("FitInnerTimetableDialog", ["ConfirmPopup"], {
 
     /**
-     * Returns the title of the timetable/session/entry
+     * Returns the title of the session block
      */
-    __getSessionTitle: function(){
-        return '"' + this.tt.contextInfo.title + '"';
+    __generateSessionBlockTitle: function(slotData) {
+        return '"' + slotData.title + (slotData.slotTitle ? ": " + slotData.slotTitle : '') + '"';
```
ThiefMaster: I would use single quotes here: `": "` to use only one style of quoting within the same statement/line.

indico/htdocs/js/indico/modules/timetable/timetable/Management.js
```diff
@@ -682,19 +682,19 @@ type("RescheduleDialog", ["ExclusivePopupWithButtons"], {
 type("FitInnerTimetableDialog", ["ConfirmPopup"], {
 
     /**
-     * Returns the title of the timetable/session/entry
+     * Returns the title of the session block
      */
-    __getSessionTitle: function(){
-        return '"' + this.tt.contextInfo.title + '"';
+    __generateSessionBlockTitle: function(slotData) {
```
ThiefMaster: i'm curious, why pass it as an argument instead of accessing it via `this.tt.contextInfo` here?

mvidalgarcia: Not really a reason, I've just based it on a similar function but your way is actually cleaner.

PR #3121, opened 2017-10-24
Use celery to send emails (and retry on failure)

### LINKED ISSUE #2981: Fix sending emails to non-ascii email addresses
Easiest way to test is by adding a reminder with an email address containing some umlauts or other non-ascii chars.

ThiefMaster:
This also fixes #2981


indico/core/emails.py
```diff
@@ -0,0 +1,81 @@
...
+DELAYS = [30, 60, 120, 300, 600, 1800, 3600, 3600, 7200]
+
+
+@celery.task(name='send_email', bind=True, max_retries=None)
+def send_email_task(task, email, log_entry=None):
+    try:
+        do_send_email(email, log_entry)
+    except Exception as exc:
+        attempt = task.request.retries + 1
+        delay = (DELAYS + [0])[task.request.retries] if not config.DEBUG else 1
```
mvidalgarcia: Why `+ [0]`?

ThiefMaster: After the last failed attempt, `task.request.retries` is 9 so we still need to get a delay to pass to `task.retry()` to let it raise the `MaxRetriesExceededError`.

FWIW, we could just use custom code to handle giving up after 10 attempts, but I think it's cleaner to use the functionality already provided by celery.

pferreir:
A massive LGTM 👍

PR #3122, opened 2017-10-25
Fix alignment of parallel timetable blocks

### LINKED ISSUE #2999: Parallel contribution block is not shown in the correct place
All the sessions have a session code set.

https://indico.cern.ch/event/466934/timetable/#20170706.detailed

timetablebug

mic4ael:
fixes #2999

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,11 @@ Improvements
   (:issue:`3446`)
 - Add edit button to custom conference pages (:issue:`3284`)
 
+Bugfixes
+^^^^^^^^
+
+- Fix incorrect order of session blocks inside timetable (:issue:`2999`)
+
 Internal Changes
 ^^^^^^^^^^^^^^^^
 
--- a/indico/web/client/js/legacy/libs/timetable/Draw.js
+++ b/indico/web/client/js/legacy/libs/timetable/Draw.js
@@ -1010,7 +1010,6 @@ type("TimetableDrawer", ["IWidget", "DroppableTimetableMixin"],
              this._blockMap = {};
 
              each(blocks, function(blockData) {
-
                  var nCol = self.layoutChooser.get().getNumColumnsForGroup(groups[blockData.group]);
                  self.maxCol = ((self.maxCol == null) || (self.maxCol < nCol)) ? nCol : self.maxCol;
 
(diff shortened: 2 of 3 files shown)


indico/htdocs/js/indico/modules/timetable/timetable/Layout.js
```diff
@@ -402,7 +401,7 @@ type("CompactLayoutManager", ["IncrementalLayoutManager"],
                  algData.grid.push([startMin/60%24, algData.topPx]);
              }
 
-             if (!algData.active) {
+             if (algData.active == 0) {
```
ThiefMaster: why this change? if it's bool-ish we actually want `!`; and if you want to compare against the integer it should be `===`

mic4ael: it is not bool-ish :P the name is misleading, this variable is kind of a counter

ThiefMaster:
unfortunately this is now causing some misalignment in other places:

alignment

mic4ael:
fixed (I think..)

ThiefMaster:
Unfortunately not :/ (CHEP 2016 timetable)

image

mic4ael:
Looks a bit better now (at least the order is preserved)

screenshot from 2017-10-31 10-48-24

ThiefMaster:
IIRC this was still problematic and more likely to cause problems (especially with the high amount of edge cases in the timetable), so it should probably not be merged for now.

mic4ael:
As far as I remember nobody has reported to me any problems with the latest fix for this. I asked a few times if anyone could take a look. Either nobody has done that or there is no problem ;)

CHANGES.rst
```diff
@@ -236,6 +236,7 @@ Bugfixes
 - Disable caching when retrieving the list of unscheduled contributions
 - Process placeholders in the subject when emailing registrants
 - Fix Shibboleth login with non-ascii names (:issue:`3143`)
+- Fix incorrect order of session blocks inside timetable (:issue:`2999`)
```
ThiefMaster: move to 2.2 changes

CHANGES.rst
```diff
@@ -29,6 +29,7 @@ Internal Changes
 - Use webpack to build static assets
 - Add React+Redux for new frontend modules
 - Enable modern ES201x features
+- Fix incorrect order of session blocks inside timetable (:issue:`2999`)
```
ThiefMaster: it's a bugfix, not an internal change (add the "Bugfixes" section after "Improvements" like we did for other versions)

PR #3125, opened 2017-10-25
Removed length limit on option caption

### LINKED ISSUE #3119: Registration form: multiple choice text too short
I created a registration form with a multiple-choice field. It seems that the maximum length of a  choice item text is much shorter than it used to be in 1.2 (I have a similar event created a 1.2 and I cannot reenter the values used without having them truncated). Is there a good reason for this change? Is there a way to customize the max length? Or at least to change it without breaking the ability to update to future versions?
If it is too complicate to have this max length customizable, I'd suggest to increase it max lengh by adding allowing at least 20 more characters.

mic4ael:
fixes #3119

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,7 @@ Improvements
 
 - Hide category field in event creation dialog if there are no
   subcategories (:issue:`3112`)
+- Remove length limit from registration form field captions (:issue:`3119`)
 
 Bugfixes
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Can you add an entry to CHANGES.rst (2.0a2, Improvements section)?

```
- Remove length limit from registration form field captions (:issue:`3119`)
```

PR #3035, opened 2017-07-28
Changing some style issues for lectures

Problemkerze:
possible solution for issue #3033

PR #3128, opened 2017-10-25
Fix margins of event description

nop33:
This fix adds margin to the right of the description, equal to the space on the left:

image

In order to address the problem with the positioning of the "Description" label, how about reusing the labelling system of the agenda?

image

imo, the "Description" label is unnecessary (although the idea @pferreir mentioned in #3035 for a better styling of the description is interesting):

image

### DIFF
--- a/indico/htdocs/sass/themes/indico.scss
+++ b/indico/htdocs/sass/themes/indico.scss
@@ -577,6 +577,7 @@ div.event-details {
 
     .event-details-content {
         flex: 1;
+        padding-right: calc(120px + 1em);
     }
 }
 


ThiefMaster:
The labels look interesting. I wonder how it looks like for attachments and/or if there's also an actual timetable below

nop33:
I'll test it out and send some screenshots

ThiefMaster:
The margin on the right looks a bit weird when you have more than just a description that's rather wide...

nop33:
Gotta set up Indico on my new mac, and try again 😉

ThiefMaster:
Hahaha you really did wait for the next Hacktoberfest to resume it! :D

ThiefMaster:
I think the extra margin still looks a bit strange. I was hoping the pretty labels were still included in the PR, but they aren't :/

PR #3138, opened 2017-11-01
Handle non-public contribution types

mic4ael:


indico/migrations/versions/20171101_1149_566d5de4e0e5_add_is_public_column_to_contribution_types.py
```diff
@@ -0,0 +1,29 @@
...
+revision = '566d5de4e0e5'
+down_revision = '1d512a9ebb30'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        'contribution_types',
+        sa.Column('is_private', sa.Boolean(), nullable=False, server_default=sa.true()),
```
ThiefMaster: `server_default='true'` would be enough, but nothing wrong with letting SA generate that string for you

indico/modules/events/abstracts/forms.py
```diff
@@ -474,7 +474,10 @@ def __init__(self, *args, **kwargs):
         if abstracts_settings.get(self.event, 'contrib_type_required'):
             inject_validators(self, 'submitted_contrib_type', [DataRequired()])
         super(AbstractForm, self).__init__(*args, **kwargs)
-        self.submitted_contrib_type.query = self.event.contribution_types
+        if self.event.can_manage(session.user):
```
ThiefMaster: I would use the `management` argument that's available in `make_abstract_form` for this (pass it to the form constructor). No need to check the event ACL again.

indico/modules/events/contributions/forms.py
```diff
@@ -187,6 +188,9 @@ class ContributionTypeForm(IndicoForm):
     """Form to create or edit a ContributionType"""
 
     name = StringField(_("Name"), [DataRequired()])
+    is_private = BooleanField(_("Private contribution type"), widget=SwitchWidget(), default=False,
```
ThiefMaster: Maybe "Private" or "Internal" is a better title? "Private contribution type" sounds a lot like something that'll be wrapped on two lines..

indico/modules/events/contributions/forms.py
```diff
@@ -187,6 +188,9 @@ class ContributionTypeForm(IndicoForm):
...
+                              description=_("If unchecked, the contribution type can be assigned by managers/judges "
```
ThiefMaster: "If selected, this contribution type cannot be selected by users submitting an abstract."?

indico/modules/events/contributions/templates/management/_types_table.html
```diff
@@ -2,6 +2,7 @@
     <tr class="i-table">
         <td class="i-table name-column">{{ contrib_type.name }}</td>
         <td class="i-table">{{ contrib_type.description }}</td>
+        <td class="i-table">{{ 'Yes' if contrib_type.is_private else 'No' }}</td>
```
ThiefMaster: `_('Yes')` and `_('No')`

indico/modules/events/abstracts/forms.py
```diff
@@ -474,7 +474,10 @@ def __init__(self, *args, **kwargs):
...
+            self.submitted_contrib_type.query = self.event.contribution_types
+        else:
+            self.submitted_contrib_type.query = self.event.contribution_types.filter_by(is_private=False)
```
ThiefMaster: This is not covering that case where you are editing an existing abstract that has a private type. I think this might do the job:

```python
criteria = [~ContributionType.is_private]
if self.abstract and self.abstract.submitted_contrib_type:
    criteria.append(ContributionType.id == self.abstract.submitted_contrib_type.id)
self.submitted_contrib_type.query = self.event.contribution_types.filter(db.or_(*criteria))
```

mic4ael:
updated

indico/modules/events/contributions/forms.py
```diff
@@ -187,6 +188,9 @@ class ContributionTypeForm(IndicoForm):
     """Form to create or edit a ContributionType"""
 
     name = StringField(_("Name"), [DataRequired()])
+    is_private = BooleanField(_("Private"), widget=SwitchWidget(), default=False,
```
ThiefMaster: `default=False` isn't needed

indico/modules/events/contributions/forms.py
mvidalgarcia: "selected" twice sounds a bit weird. Maybe ".. cannot be chosen by users...?

mic4ael:
updated

indico/modules/events/abstracts/forms.py
```diff
@@ -474,7 +474,13 @@ def __init__(self, *args, **kwargs):
         if abstracts_settings.get(self.event, 'contrib_type_required'):
             inject_validators(self, 'submitted_contrib_type', [DataRequired()])
         super(AbstractForm, self).__init__(*args, **kwargs)
-        self.submitted_contrib_type.query = self.event.contribution_types
+        if kwargs['management']:
```
ThiefMaster: management = kwargs.pop('management')

before calling super (i'd put it after setting `self.abstract`); then use `management` here. no need to pass along  this kwarg to the superclass.

also, it sounds like something where a default value could make sense =>

    management = kwargs.pop('management', False)

PR #3144, opened 2017-11-08
Allow importing registrations/contributions from CSV files

pferreir:
For now it only supports personal data fields (registration) and single speakers (contributions).

### DIFF
--- a/indico/modules/events/contributions/controllers/management.py
+++ b/indico/modules/events/contributions/controllers/management.py
@@ -44,7 +44,7 @@ from indico.modules.events.contributions.operations import (create_contribution,
                                                             delete_contribution, delete_subcontribution,
                                                             update_contribution, update_subcontribution)
 from indico.modules.events.contributions.util import (contribution_type_row, generate_spreadsheet_from_contributions,
-                                                      make_contribution_form)
+                                                      import_contributions_from_csv, make_contribution_form)
 from indico.modules.events.contributions.views import WPManageContributions
 from indico.modules.events.logs import EventLogKind, EventLogRealm
 from indico.modules.events.management.controllers import RHManageEventBase
@@ -52,6 +52,7 @@ from indico.modules.events.management.controllers.base import RHContributionPers
 from indico.modules.events.management.util import flash_if_unregistered
 from indico.modules.events.models.references import ReferenceType
 from indico.modules.events.sessions import Session
+from indico.modules.events.timetable.forms import ImportContributionsForm
 from indico.modules.events.timetable.operations import update_timetable_entry
 from indico.modules.events.tracks.models.tracks import Track
 from indico.modules.events.util import (check_event_locked, get_field_values, track_time_changes,
@@ -470,6 +471,24 @@ class RHContributionsExportPDFBookSorted(RHManageContributionsExportActionsBase)
         pdf = ContributionBook(self.event, session.user, self.contribs, tz=self.event.timezone,
                                sort_by='board_number')
         return send_file('book-of-abstracts.pdf', pdf.generate(), 'application/pdf')
+
+
+class RHContributionsImportCSV(RHManageContributionsActionsBase):
+    """Import contributions from a CSV file"""
+
+    def _process(self):
+        form = ImportContributionsForm()
+
+        if form.validate_on_submit():
+            contributions, changes = import_contributions_from_csv(self.event, form.source_file.data)
+            flash(ngettext("{} contribution has been imported.",
+                           "{} contributions have been imported.",
+                           len(contributions)).format(len(contributions)), 'success')
+            if changes:
+                flash(_("Event dates/times adjusted due to imported data."), 'warning')
+            return jsonify_data(flash=False, redirect=url_for('.manage_contributions', self.event))
+        return jsonify_template('events/contributions/management/import_contributions.html', form=form,
+                                event=self.event)
 
 
 class RHManageContributionTypes(RHManageContributionsBase):
--- a/indico/modules/events/contributions/operations.py
+++ b/indico/modules/events/contributions/operations.py
@@ -62,6 +62,7 @@ def _ensure_consistency(contrib):
 
 
 def create_contribution(event, contrib_data, custom_fields_data=None, session_block=None, extend_parent=False):
+    user = session.user if session else None
     start_dt = contrib_data.pop('start_dt', None)
     contrib = Contribution(event=event)
     contrib.populate_from_dict(contrib_data)
@@ -71,9 +72,9 @@ def create_contribution(event, contrib_data, custom_fields_data=None, session_bl
         set_custom_fields(contrib, custom_fields_data)
     db.session.flush()
     signals.event.contribution_created.send(contrib)
-    logger.info('Contribution %s created by %s', contrib, session.user)
+    logger.info('Contribution %s created by %s', contrib, user)
     contrib.event.log(EventLogRealm.management, EventLogKind.positive, 'Contributions',
-                      'Contribution "{}" has been created'.format(contrib.title), session.user)
+                      'Contribution "{}" has been created'.format(contrib.title), user)
     return contrib
 
 
--- a/indico/modules/events/contributions/templates/management/contributions.html
+++ b/indico/modules/events/contributions/templates/management/contributions.html
@@ -127,7 +127,7 @@
                     <a class="i-button"
                        title="{% trans %}Import contributions from CSV file{% endtrans %}"
                        data-href="{{ url_for('.contributions_import', event) }}"
-                       data-title="{% trans %}Import Registrants from CSV File{% endtrans %}"
+                       data-title="{% trans %}Import contributions from CSV File{% endtrans %}"
                        data-ajax-dialog>
                         {%- trans %}Import{% endtrans -%}
                     </a>
--- a/indico/modules/events/contributions/templates/management/contributions.html
+++ b/indico/modules/events/contributions/templates/management/contributions.html
@@ -86,43 +86,52 @@
                    data-href="{{ url_for('.material_package', event) }}">
                     {%- trans %}Material package{% endtrans -%}
                 </a>
-                <a class="i-button js-enable-if-checked arrow disabled"
-                   title="{% trans %}Export data{% endtrans %}"
-                   data-toggle="dropdown">
-                    {%- trans %}Export{% endtrans -%}
-                </a>
-                <ul class="dropdown">
-                    <li>
-                        <a href="#" class="icon-file-pdf js-submit-form js-enable-if-checked disabled"
-                           data-href="{{ url_for('.contributions_pdf_export', event) }}">
-                            {%- trans %}PDF{% endtrans -%}
-                        </a>
-                    </li>
-                    <li>
-                        <a href="#" class="icon-file-pdf js-submit-form js-enable-if-checked disabled"
-                           data-href="{{ url_for('.contributions_pdf_export_book', event) }}">
-                            {%- trans %}PDF (book of abstracts){% endtrans -%}
-                        </a>
-                    </li>
-                    <li>
-                        <a href="#" class="icon-file-pdf js-submit-form js-enable-if-checked disabled"
-                           data-href="{{ url_for('.contributions_pdf_export_book_sorted', event) }}">
-                            {%- trans %}PDF (book of abstracts - sorted by board number){% endtrans -%}
-                        </a>
-                    </li>
-                    <li>
-                        <a href="#" class="icon-file-spreadsheet js-submit-form js-enable-if-checked disabled"
-                           data-href="{{ url_for('.contributions_csv_export', event) }}">
-                            {%- trans %}CSV{% endtrans -%}
-                        </a>
-                    </li>
-                    <li>
-                        <a href="#" class="icon-file-excel js-submit-form js-enable-if-checked disabled"
-                           data-href="{{ url_for('.contributions_excel_export', event) }}">
-                            {%- trans %}XLSX (Excel){% endtrans -%}
-                        </a>
-                    </li>
-                </ul>
+                <div class="group">
+                    <a class="i-button js-enable-if-checked arrow disabled"
+                       title="{% trans %}Export data{% endtrans %}"
+                       data-toggle="dropdown">
+                        {%- trans %}Export{% endtrans -%}
+                    </a>
+                    <ul class="dropdown">
+                        <li>
+                            <a href="#" class="icon-file-pdf js-submit-form js-enable-if-checked disabled"
+                               data-href="{{ url_for('.contributions_pdf_export', event) }}">
+                                {%- trans %}PDF{% endtrans -%}
+                            </a>
+                        </li>
+                        <li>
+                            <a href="#" class="icon-file-pdf js-submit-form js-enable-if-checked disabled"
+                               data-href="{{ url_for('.contributions_pdf_export_book', event) }}">
+                                {%- trans %}PDF (book of abstracts){% endtrans -%}
+                            </a>
+                        </li>
+                        <li>
+                            <a href="#" class="icon-file-pdf js-submit-form js-enable-if-checked disabled"
+                               data-href="{{ url_for('.contributions_pdf_export_book_sorted', event) }}">
+                                {%- trans %}PDF (book of abstracts - sorted by board number){% endtrans -%}
+                            </a>
+                        </li>
+                        <li>
+                            <a href="#" class="icon-file-spreadsheet js-submit-form js-enable-if-checked disabled"
+                               data-href="{{ url_for('.contributions_csv_export', event) }}">
+                                {%- trans %}CSV{% endtrans -%}
+                            </a>
+                        </li>
+                        <li>
+                            <a href="#" class="icon-file-excel js-submit-form js-enable-if-checked disabled"
+                               data-href="{{ url_for('.contributions_excel_export', event) }}">
+                                {%- trans %}XLSX (Excel){% endtrans -%}
+                            </a>
+                        </li>
+                    </ul>
+                    <a class="i-button"
+                       title="{% trans %}Import contributions from CSV file{% endtrans %}"
+                       data-href="{{ url_for('.contributions_import', event) }}"
+                       data-title="{% trans %}Import Registrants from CSV File{% endtrans %}"
+                       data-ajax-dialog>
+                        {%- trans %}Import{% endtrans -%}
+                    </a>
+                </div>
             </div>
             <div class="toolbar">
                 <div class="group" id="filter-statistics">
--- a/indico/modules/events/contributions/templates/management/import_contributions.html
+++ b/indico/modules/events/contributions/templates/management/import_contributions.html
@@ -34,8 +34,11 @@
         </li>
     </ul>
     {% trans -%}
-        Only the field "Title" is mandatory. Users will be matched with existing
-        Indico identities through their e-mail.
+        Only the field "Title" is mandatory.
+    {%- endtrans %}
+    <br>
+    {% trans -%}
+        Users will be matched with existing Indico identities through their e-mail.
     {%- endtrans %}
 {% endcall %}
 
--- a/indico/modules/events/contributions/templates/management/import_contributions.html
+++ b/indico/modules/events/contributions/templates/management/import_contributions.html
@@ -1,0 +1,43 @@
+{% from 'forms/_form.html' import simple_form  %}
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('highlight', large_icon=true) %}
+    {% trans -%}
+        You should upload a CSV (comma-separated values) file with exactly 7 columns in the
+        following order:
+    {%- endtrans %}
+    <ul>
+        <li>
+            {%- trans link='<a href="https://en.wikipedia.org/wiki/ISO_8601">'|safe, endlink='</a>'|safe,
+                      subitem='<ul><li><em>'|safe, endsubitem='</em></li></ul>'|safe -%}
+                Start date/time in {{ link }}ISO 8601{{ endlink }} format
+                {{ subitem }}e.g. 2017-12-03T12:00:00{{ endsubitem }}
+            {%- endtrans -%}
+        </li>
+        <li>
+            {%- trans %}Duration (minutes){% endtrans -%}
+        </li>
+        <li>
+            {%- trans %}Title{% endtrans -%}
+        </li>
+        <li>
+            {%- trans %}Speaker - First name{% endtrans -%}
+        </li>
+        <li>
+            {%- trans %}Speaker - Last name{% endtrans -%}
+        </li>
+        <li>
+            {%- trans %}Speaker - Affiliation{% endtrans -%}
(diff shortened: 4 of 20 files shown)


indico/modules/events/registration/forms.py
```diff
@@ -40,6 +40,7 @@
 from indico.web.forms.base import IndicoForm, generated_data
 from indico.web.forms.fields import (EmailListField, IndicoDateTimeField, IndicoEnumSelectField, JSONField,
                                      PrincipalListField)
+from indico.web.forms.fields.files import FileField
```
ThiefMaster: can be imported from `indico.web.forms.fields` in the statement right above

indico/modules/events/registration/forms.py
```diff
@@ -392,3 +393,14 @@ def __init__(self, event, **kwargs):
 
     def is_submitted(self):
         return super(BadgeSettingsForm, self).is_submitted() and 'submitted' in request.form
+
+
+class ImportRegistrationsForm(IndicoForm):
+    source_file = FileField(_("Source File"), accepted_file_types='.csv')
+    skip_moderation = BooleanField(_("Skip Moderation"), widget=SwitchWidget(), default=True,
+                                   description=_("If enabled, the registration will me immediately accepted"))
```
mvidalgarcia: > ... will **be** immediately ...

indico/modules/events/registration/forms.py
```diff
@@ -392,3 +393,14 @@ def __init__(self, event, **kwargs):
...
+    notify_users = BooleanField(_("E-mail users"), widget=SwitchWidget(),
+                                description=_("Whether the imported users should receive an e-mail notification"))
+    create_pending = BooleanField(_("Match missing users"), widget=SwitchWidget(), default=True,
```
ThiefMaster: Is there any good reason to create pending users here? If a user registers later they are associated with the registration anyway, and in the user search we never show pending users. So I'd rather offer to create an EventPerson so they can find him when searching to add e.g. a speaker - but never a pending indico User

indico/modules/events/registration/templates/management/import_registrations.html
```diff
@@ -0,0 +1,30 @@
...
+                <li>Position</li>
+                <li>Phone Number</li>
+                <li>E-mail</li>
+            </ul>
+            The fields "First Name", "Last Name" and "E-mail" are mandatory. Users will be matched with existing
+            Indico identities through their e-mail.
+        {%- endtrans %}
+    {% endcall %}
+
+    {{ form_header(form, multipart=True, action=url_for('.registrations_import', regform)) }}
```
ThiefMaster: `true`

indico/modules/events/registration/templates/management/import_registrations.html
```diff
@@ -0,0 +1,30 @@
...
+    {{ form_rows(form) }}
+
+    {% call form_footer(form) %}
+        <input type="submit" class="i-button big highlight" value="{% trans %}Submit{% endtrans %}" disabled>
```
ThiefMaster: why `disabled`? shouldn't this rather be `data-disabled-until-change`?

indico/modules/events/registration/templates/management/import_registrations.html
```diff
@@ -0,0 +1,30 @@
+{% extends 'layout/base.html' %}
```
ThiefMaster: It's a dialog so this and the block shouldn't be needed

indico/modules/events/registration/templates/management/import_registrations.html
```diff
@@ -0,0 +1,30 @@
...
+        {% trans num_columns=6 -%}
+            You should upload a CSV (comma-separated values) file with exactly {{num_columns}} columns in the
+            following order:
+            <ul>
+                <li>First Name</li>
+                <li>Last Name</li>
+                <li>Affiliation</li>
+                <li>Position</li>
+                <li>Phone Number</li>
+                <li>E-mail</li>
```
ThiefMaster: Wouldn't it make more sense to put First/Last/Email first since those are probably the most common fields?

indico/modules/events/registration/util.py
```diff
@@ -47,10 +48,15 @@
 from indico.util.date_time import format_date
 from indico.util.i18n import _
 from indico.util.spreadsheets import unique_col
+from indico.util.string import to_unicode
 from indico.web.forms.base import IndicoForm
 from indico.web.forms.widgets import SwitchWidget
 
 
+class CSVImportException(Exception):
```
ThiefMaster: inherit from `UserValueError` so it's not reportable?

indico/modules/events/registration/templates/management/import_registrations.html
ThiefMaster: the url is the same as the one the dialog is loaded from so omitting it should work. also, I think you can use `simple_form()` here

indico/modules/events/registration/forms.py
pferreir: And how do we phrase that option (EventPerson)?

indico/modules/events/registration/templates/management/import_registrations.html
pferreir: I put 'e-mail' at the end because some of the other fields may be empty (e.g. position), and the resulting file may end up having < 6 columns.

indico/modules/events/registration/templates/management/import_registrations.html
pferreir: Dropzone complains If I specify no URL.

indico/modules/events/registration/forms.py
pferreir: Then I'll just remove the option and set it to `False` by default. Anyway, the unification of Event Persons with registrations in version 2.1 should make this irrelevant.

indico/modules/events/registration/forms.py
```diff
@@ -392,3 +392,11 @@ def __init__(self, event, **kwargs):
 
     def is_submitted(self):
         return super(BadgeSettingsForm, self).is_submitted() and 'submitted' in request.form
+
+
+class ImportRegistrationsForm(IndicoForm):
+    source_file = FileField(_("Source File"), accepted_file_types='.csv')
+    skip_moderation = BooleanField(_("Skip Moderation"), widget=SwitchWidget(), default=True,
```
ThiefMaster: I'd remove this if moderation is disabled, like we do in `InvitationFormBase`

indico/modules/events/registration/util.py
```diff
@@ -492,3 +499,25 @@ def generate_ticket(registration):
...
+
+def import_registrations_from_csv(regform, f, skip_moderation=True, notify_users=False):
+    """Import event registrants from a CSV file into a form."""
+    reader = csv.reader(f)
+    registrations = []
+    for row in reader:
+        try:
+            first_name, last_name, affiliation, position, phone, email = [to_unicode(value).strip() for value in row]
+        except ValueError:
+            raise CSVImportException('Malformed CSV data - please check that the number of columns is correct')
```
ThiefMaster: `_(...)`?

indico/modules/events/registration/util.py
```diff
@@ -492,3 +499,25 @@ def generate_ticket(registration):
...
+
+        with db.session.no_autoflush:
+            registrations.append(create_registration(regform, {
+                'email': email,
+                'first_name': first_name.title(),
```
ThiefMaster: http://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/ ;)

Some of the things there are probably not super relevant, but I have the feeling titlecasing the name will screw e.g. many dutch names ("Foo van Bar" etc.) that have lowercase words. Maybe better to only titlecase if the original value is lowercase-only?

indico/modules/events/registration/util.py
```diff
@@ -492,3 +499,25 @@ def generate_ticket(registration):
 
 def get_ticket_attachments(registration):
     return [('Ticket.pdf', generate_ticket(registration).getvalue())]
+
+
+def import_registrations_from_csv(regform, f, skip_moderation=True, notify_users=False):
```
ThiefMaster: `s/f/fileobj/`?

ThiefMaster:
Probably good to rebase to v2.1-dev and add a changelog entry

indico/modules/events/registration/util.py
pferreir: The problem is actually UPPERCASE NAMES.

ThiefMaster: lowercase-only or uppercase-only then ;)

pferreir: Well, the question is whether we'd rather have people input all kinds of silly stuff (CAPITALS!) or miss a few ones here and there. I can't see any better solution, unfortunately.

pferreir:
I'll actually add some tests first.

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -86,43 +86,52 @@
...
-                    <li>
-                        <a href="#" class="icon-file-excel js-submit-form js-enable-if-checked disabled"
-                           data-href="{{ url_for('.contributions_excel_export', event) }}">
-                            {%- trans %}XLSX (Excel){% endtrans -%}
-                        </a>
-                    </li>
-                </ul>
+                <div class="group">
+                    <a class="i-button js-enable-if-checked arrow disabled"
+                    title="{% trans %}Export data{% endtrans %}"
```
ThiefMaster: wrong indentation

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -86,43 +86,52 @@
...
+                    data-toggle="dropdown">
+                        {%- trans %}Export{% endtrans -%}
+                    </a>
+                    <ul class="dropdown">
+                        <li>
+                            <a href="#" class="icon-file-pdf js-submit-form js-enable-if-checked disabled"
+                            data-href="{{ url_for('.contributions_pdf_export', event) }}">
```
ThiefMaster: wrong indentation (also a few more times below)

indico/modules/events/contributions/templates/management/import_contributions.html
```diff
@@ -0,0 +1,28 @@
+{% from 'forms/_form.html' import simple_form  %}
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('highlight', large_icon=true) %}
+    {% trans -%}
+        You should upload a CSV (comma-separated values) file with exactly 7 columns in the
+        following order:
+        <ul>
+            <li>Start date/time in <a href="https://en.wikipedia.org/wiki/ISO_8601">ISO 8601</a> format</li>
+            <ul>
```
ThiefMaster: an `<ul>` right inside an `<ul>` looks wrong

indico/modules/events/contributions/templates/management/import_contributions.html
```diff
@@ -0,0 +1,28 @@
+{% from 'forms/_form.html' import simple_form  %}
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('highlight', large_icon=true) %}
+    {% trans -%}
```
ThiefMaster: is it really a good idea to have this all in a single trans block? seems like it could be split easily (one per list item, one for the intro sentence, one for the sentence at the end) without having something that might be messy in other languages

indico/modules/events/contributions/util.py
```diff
@@ -16,22 +16,29 @@
 
 from __future__ import unicode_literals
 
+import csv
+import dateutil.parser
```
ThiefMaster: not stdlib

indico/modules/events/contributions/util.py
```diff
@@ -189,3 +196,50 @@ def get_contribution_ical_file(contrib):
...
+    for n_row, row in enumerate(reader):
+        try:
+            start_dt, duration, title, first_name, last_name, affiliation, email = row
+        except ValueError:
+            raise UserValueError(_('Row {}: malformed CSV data - please check that the number of columns is correct')
+                                 .format(n_row))
+        try:
+            parsed_start_dt = event.tzinfo.localize(dateutil.parser.parse(start_dt)) if start_dt else None
+        except ValueError:
+            raise UserValueError(_("Row {}: can't parse date: \"{}\"").format(n_row, start_dt))
```
ThiefMaster: sinlge quotes instead of escaping? ;)

indico/modules/events/contributions/util.py
```diff
@@ -189,3 +196,50 @@ def get_contribution_ical_file(contrib):
...
+            'firstName': first_name,
+            'familyName': last_name,
+            'affiliation': affiliation,
+            'email': email
+        })
+        link = ContributionPersonLink(person=person, is_speaker=True)
+        contribution.person_links.append(link)
+        contributions.append(contribution)
+
+    return contributions, changes
```
ThiefMaster: you return the `changes` from the last iteration but nothing from earlier.. maybe better to set a flag within the loop if there was a change and return that instead?

indico/modules/events/persons/util.py
```diff
@@ -0,0 +1,71 @@
...
+
+def get_event_person_for_user(event, user, create_untrusted_persons=False):
+    """Return the event person that links to a given User/Event (if any)."""
+    return EventPerson.for_user(user, event, is_untrusted=create_untrusted_persons)
+
+
+def get_event_person(event, data, create_untrusted_persons=False, allow_external=False, allow_emails=False,
+                     allow_networks=False):
+    """
+    Get an EventPerson from dictionary data.
```
ThiefMaster: should be in the `"""` line since it's a short single-line description

indico/modules/events/persons/util.py
```diff
@@ -0,0 +1,71 @@
...
+        # XXX: existing_data
+        principal = principal_from_fossil(data, allow_pending=allow_external, allow_emails=allow_emails,
+                                          allow_networks=allow_networks)
+        return get_event_person_for_user(event, principal, create_untrusted_persons=create_untrusted_persons)
+    elif person_type == 'EventPerson':
+        return event.persons.filter_by(id=data['id']).one()
+    elif person_type == 'PersonLink':
+        return event.persons.filter_by(id=data['personId']).one()
+    else:
+        raise ValueError(_("Unknown person type '{}'").format(person_type))
```
ThiefMaster: I don't think this should be translatable; users should never see this except when messing with the data sent to the server.

indico/modules/events/persons/util_test.py
```diff
@@ -0,0 +1,131 @@
...
+    db.session.flush()
+
+    person_2 = get_event_person(dummy_event, dict(data, _type='EventPerson', id=person_1.id))
+    assert person_2 == person_1
+
+    person_3 = get_event_person(dummy_event, dict(data, _type='PersonLink', personId=person_1.id))
+    assert person_3 == person_1
+
+    with pytest.raises(ValueError):
+        get_event_person(dummy_event, dict(data, _type='UnsupportedPersonType'))
```
ThiefMaster: missing trailing linebreak

indico/modules/events/registration/templates/management/import_registrations.html
```diff
@@ -0,0 +1,22 @@
+{% from 'forms/_form.html' import simple_form  %}
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('highlight', large_icon=true) %}
+    {% trans num_columns=6 -%}
+        You should upload a CSV (comma-separated values) file with exactly {{num_columns}} columns in the
```
ThiefMaster: space around `num_columns`

indico/modules/events/registration/templates/management/import_registrations.html
```diff
@@ -0,0 +1,22 @@
+{% from 'forms/_form.html' import simple_form  %}
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('highlight', large_icon=true) %}
+    {% trans num_columns=6 -%}
```
ThiefMaster: I think this big trans block could also be split into small ones

indico/modules/events/registration/util.py
```diff
@@ -493,3 +496,31 @@ def generate_ticket(registration):
...
+
+        with db.session.no_autoflush:
+            registrations.append(create_registration(regform, {
+                'email': email,
+                'first_name': first_name.title(),
+                'last_name': last_name.title(),
+                'affiliation': affiliation,
+                'phone': phone,
+                'position': position
+            }, notify_user=notify_users, management=skip_moderation))
```
ThiefMaster: Using the `management` kwarg also affects e.g. the log realm. Maybe better to add a new `skip_moderation` kwarg?

indico/modules/events/registration/util_test.py
```diff
@@ -0,0 +1,101 @@
...
+    assert registrations[2].full_name == 'Billy Bob Doe'
+    assert registrations[2].user == dummy_user
+    data = registrations[2].get_personal_data()
+    assert 'affiliation' not in data
+    assert data['email'] == '1337@example.com'
+    assert 'position' not in data
+    assert 'phone' not in data
+
+
+@pytest.mark.usefixtures('request_context')
```
ThiefMaster: I think we should rather use `session.user if session else None` (`__bool__` of LocalProxy catches the RuntimeError and returns False in that case) when logging the registration instead of using/requiring a request context. It doesn't make much sense that something as general as `create_registration` only works with an active request context...

indico/modules/events/timetable/forms.py
```diff
@@ -209,3 +209,7 @@ def data_for_format(self):
         for fieldname in fields:
             data.update(getattr(self, fieldname).data)
         return data
+
+
+class ImportContributionsForm(IndicoForm):
+    source_file = FileField(_("Source File"), accepted_file_types='.csv')
```
ThiefMaster: in some places we specify a mime type for `accepted_file_types`, in others a file extension... not sure if there's any advantage of using one or the other.

indico/modules/events/contributions/util.py
pferreir: There's an apostrophe in the sentence.

indico/modules/events/persons/util.py
pferreir: There's more stuff below. The first sentence should always be 1 line only IIRC.

ThiefMaster: > The summary line may be on the same line as the opening quotes or on the next line.

so yeah, it's not required - but consistent with our other docstrings ;)

pferreir: Well, I tend to like these two styles:

```
"""This function doesn't do anything."""
```

```
"""
This function doesn't do anything.

It's pretty useless indeed, and it returns a random value just to
confuse you.
"""
```

It annoys me a bit the lack of consistency of our current approach:

```
"""This function doesn't do anything.

It's pretty useless indeed, and it returns a random value just to
confuse you.
"""
```
But yeah, that's what we use mostly.

pferreir:
OK, I believe everything has been fixed. Awaiting final QA.

indico/modules/events/contributions/controllers/management.py
```diff
@@ -471,6 +472,25 @@ def _process(self):
...
+
+        if form.validate_on_submit():
+            contributions, changes = import_contributions_from_csv(self.event, form.source_file.data)
+            flash(ngettext("{} contribution has been imported.",
+                           "{} contributions have been imported.",
+                           len(contributions)).format(len(contributions)), 'success')
+            if changes:
+                flash(_("Event dates/times adjusted due to imported data."), 'warning')
+            return jsonify_data(flash=False, redirect=url_for('.manage_contributions', self.event),
+                                redirect_no_loading=True)
```
ThiefMaster: I don't think `redirect_no_loading` is useful here. I added it for actions that trigger a download prompt where you usually remain on the same page and thus don't want to block it with the "Loading" dialog

indico/modules/events/contributions/templates/management/import_contributions.html
```diff
@@ -0,0 +1,43 @@
...
+        You should upload a CSV (comma-separated values) file with exactly 7 columns in the
+        following order:
+    {%- endtrans %}
+    <ul>
+        <li>
+            {%- trans link='<a href="https://en.wikipedia.org/wiki/ISO_8601">'|safe, endlink='</a>'|safe,
+                        subitem='<ul><li><em>'|safe, endsubitem='</em></li></ul>'|safe -%}
+                Start date/time in {{link}}ISO 8601{{endlink}} format
```
ThiefMaster: missing space inside `{{ ... }}`

indico/modules/events/contributions/templates/management/import_contributions.html
```diff
@@ -0,0 +1,43 @@
...
+
+{% call message_box('highlight', large_icon=true) %}
+    {% trans -%}
+        You should upload a CSV (comma-separated values) file with exactly 7 columns in the
+        following order:
+    {%- endtrans %}
+    <ul>
+        <li>
+            {%- trans link='<a href="https://en.wikipedia.org/wiki/ISO_8601">'|safe, endlink='</a>'|safe,
+                        subitem='<ul><li><em>'|safe, endsubitem='</em></li></ul>'|safe -%}
```
ThiefMaster: the indentation seems weird

indico/modules/events/contributions/util.py
```diff
@@ -189,3 +196,60 @@ def get_contribution_ical_file(contrib):
...
     serializer = Serializer.create('ics')
     return BytesIO(serializer(data))
+
+
+def import_contributions_from_csv(event, f):
+    """Import timetable contributions from a CSV file into an event."""
+    reader = csv.reader(f)
+    contributions = []
+    all_changes = defaultdict(list)
+    for n_row, row in enumerate(reader, 1):
```
ThiefMaster: `row_num`?

pferreir: 😠

indico/modules/events/contributions/operations.py
```diff
@@ -62,6 +62,7 @@ def _ensure_consistency(contrib):
 
 
 def create_contribution(event, contrib_data, custom_fields_data=None, session_block=None, extend_parent=False):
+    user = session.user if session else None
```
mic4ael: is it only for making tests run?

ThiefMaster: It also lets you use this operation from the Indico shell without creating a request context.

indico/modules/events/contributions/templates/management/import_contributions.html
```diff
@@ -0,0 +1,43 @@
...
+        </li>
+        <li>
+            {%- trans %}Speaker - Affiliation{% endtrans -%}
+        </li>
+        <li>
+            {%- trans %}Speaker - E-mail address{% endtrans -%}
+        </li>
+    </ul>
+    {% trans -%}
+        Only the field "Title" is mandatory. Users will be matched with existing
```
mic4ael: what about E-mail address? isn't it mandatory as well?

ThiefMaster: No, you may have speakers without an email address.

pferreir: And no speakers at all 😉

PR #3150, opened 2017-11-14
Allow to change the default reloader type

mic4ael:


indico/cli/core.py
```diff
@@ -146,6 +144,7 @@ def cleanup(temp, cache, assets, verbose, dry_run, min_age):
 @click.option('--evalex-from', multiple=True,
               help='Restrict the debugger shell to the given ips (can be used multiple times)')
 @click.option('--proxy', is_flag=True, help='Use the ip and protocol provided by the proxy.')
+@click.option('--reloader-type', default='auto', help='Use specific reloader type for your devserver.')
```
ThiefMaster: I'd use `type=click.Choice(...)` so people actually know what they can use. And maybe rename the option to `--reloader`

indico/cli/core.py
```diff
@@ -146,6 +144,8 @@ def cleanup(temp, cache, assets, verbose, dry_run, min_age):
 @click.option('--evalex-from', multiple=True,
               help='Restrict the debugger shell to the given ips (can be used multiple times)')
 @click.option('--proxy', is_flag=True, help='Use the ip and protocol provided by the proxy.')
+@click.option('--reloader', type=click.Choice(['auto', 'stat', 'watchdog']), default='auto',
+              help='Use specific reloader type for your devserver.')
```
ThiefMaster: > Specify the auto-reloader to use.

your help text is missing a "the' but I think it can be made more concise like this since the whole command is for the dev server -> no need to mention it again

PR #3152, opened 2017-11-14
Properly support last-day-of-month event cloning

### LINKED ISSUE #1580: Event cloning with monthly interval fails
``` yaml
# Ticket imported from Trac
```

Cloning an event with monthly intervals fails when the start date is set to the last day of a month with 31 days. This is caused by the way we iterate months: creating a new date object using the day of start_date. This happens both for _Clone the event with a fixed interval_ and _Clone the agenda on given days_.

For example:
An event whose repetition starts a 31st October will fail when trying to compose the 31st of November.

The patch should be done using dateutil.rrule in both cases.

For the first type of repetitions we should keep the day of the month unless it doesn't exist for a particular month. In that case we will create the event on the last day of the month. i.e. January 31, February 28, March 31, ...

For the second type of repetitions we should make sure that we don't skip months that don't have 5th weekday and creating the event on the 4th weekday as a fallback. i.e. 5th Sunday of November, 4th Sunday of December, ... 5th Sunday of March.

meluru:
closes #1580

### DIFF
--- a/indico/htdocs/js/indico/modules/events/cloning.js
+++ b/indico/htdocs/js/indico/modules/events/cloning.js
@@ -61,13 +61,14 @@
                 success: function(data) {
                     if (data.success) {
                         var $countNumber = $eventCount.find('.count');
+                        var $lastDay = $eventCount.find('.last-day');
                         $countNumber.text(data.count);
                         $cloneErrors.hide();
                         $eventCount.show();
-
                         $eventList.toggle(!!data.count);
                         $cloneButton.prop('disabled', !data.count);
                         $eventCount.data('event-dates', data.dates);
+                        $lastDay.toggle(data.last_day_of_month);
                     } else {
                         $cloneErrors.show().find('.message-text').html(errorToHTML(data.error.message));
                         $eventCount.hide();
--- a/indico/modules/events/cloning_test.py
+++ b/indico/modules/events/cloning_test.py
@@ -1,0 +1,92 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2018 European Organization for Nuclear Research (CERN).
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 3 of the
+# License, or (at your option) any later version.
+#
+# Indico is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from datetime import datetime
+
+import pytest
+from pytz import timezone
+
+from indico.modules.events.management.controllers.cloning import (CloneCalculator, IntervalCloneCalculator,
+                                                                  PatternCloneCalculator)
+from indico.modules.events.management.forms import CloneRepeatIntervalForm, CloneRepeatPatternForm
+from indico.util.date_time import relativedelta
+
+
+@pytest.mark.parametrize(('start_dt', 'delta', 'stop_criterion', 'num_times', 'until_dt',
+                          'expected_dates', 'expected_flag'), (
+    (datetime(2017, 2, 28, 3, 0, 0), relativedelta(years=1), 'num_times', 2, None,
+     [datetime(2017, 2, 28, 3, 0, 0), datetime(2018, 2, 28, 3, 0, 0)], False),
+    (datetime(2017, 2, 28, 3, 0, 0), relativedelta(months=1), 'num_times', 3, None,
+     [datetime(2017, 2, 28, 3, 0, 0), datetime(2017, 3, 31, 3, 0, 0), datetime(2017, 4, 30, 3, 0, 0)], True),
+    (datetime(2017, 2, 27, 3, 0, 0), relativedelta(months=1), 'num_times', 3, None,
+     [datetime(2017, 2, 27, 3, 0, 0), datetime(2017, 3, 27, 3, 0, 0), datetime(2017, 4, 27, 3, 0, 0)], False),
+    (datetime(2017, 2, 28, 3, 0, 0), relativedelta(months=1), 'day', None, datetime(2017, 4, 30, 3, 0, 0),
+     [datetime(2017, 2, 28, 3, 0, 0), datetime(2017, 3, 31, 3, 0, 0)], True),
+    (datetime(2017, 2, 28, 3, 0, 0), relativedelta(weeks=2), 'num_times', 3, None,
+     [datetime(2017, 2, 28, 3, 0, 0), datetime(2017, 3, 14, 3, 0, 0), datetime(2017, 3, 28, 3, 0, 0)], False),
+    (datetime(2017, 2, 28, 3, 0, 0), relativedelta(days=1), 'day', None, datetime(2017, 3, 2, 3, 0, 0),
+     [datetime(2017, 2, 28, 3, 0, 0), datetime(2017, 3, 1, 3, 0, 0), datetime(2017, 3, 2, 3, 0, 0)], False)
+))
+def test_interval_clone(start_dt, delta, stop_criterion, num_times, until_dt,
+                        expected_dates, expected_flag, dummy_event):
+    dummy_event.timezone = 'Europe/Zurich'
+    clone_calculator = IntervalCloneCalculator(dummy_event)
+    form = CloneRepeatIntervalForm(dummy_event, csrf_enabled=False, recurrence=delta, stop_criterion=stop_criterion,
+                                   num_times=num_times)
+    expected_dates = [dummy_event.tzinfo.localize(dt) for dt in expected_dates]
+    form.start_dt.data = dummy_event.tzinfo.localize(start_dt)
+    form.until_dt.data = dummy_event.tzinfo.localize(until_dt) if until_dt else None
+    dates, last_day_of_the_month = clone_calculator._calculate(form)
+    assert last_day_of_the_month is expected_flag
+    assert dates == expected_dates
+
+
+@pytest.mark.parametrize(('start_dt', 'week_day', 'num_months',  'stop_criterion', 'num_times', 'until_dt',
+                          'expected_dates', 'expected_flag'), (
+    (datetime(2017, 12, 7, 3, 0, 0), (1, 0), 1, 'num_times', 2, None,
+     [datetime(2018, 1, 1, 3, 0, 0), datetime(2018, 2, 5, 3, 0, 0)], False),
+    (datetime(2018, 1, 1, 3, 0, 0), (1, 0), 1, 'num_times', 2, None,
+     [datetime(2018, 1, 1, 3, 0, 0), datetime(2018, 2, 5, 3, 0, 0)], False),
+    (datetime(2017, 11, 25, 3, 0, 0), (4, 6), 1, 'num_times', 3, None,
+     [datetime(2017, 11, 26, 3, 0, 0), datetime(2017, 12, 24, 3, 0, 0), datetime(2018, 1, 28, 3, 0, 0)], False),
+    (datetime(2017, 11, 25, 3, 0, 0), (-1, 6), 1, 'num_times', 3, None,
+     [datetime(2017, 11, 26, 3, 0, 0), datetime(2017, 12, 31, 3, 0, 0), datetime(2018, 1, 28, 3, 0, 0)], False),
+    (datetime(2017, 7, 19, 3, 0, 0), (3, 2), 2, 'day', None, datetime(2017, 11, 30, 3, 0, 0),
+     [datetime(2017, 7, 19, 3, 0, 0), datetime(2017, 9, 20, 3, 0, 0), datetime(2017, 11, 15, 3, 0, 0)], False),
+
+))
+def test_pattern_clone(start_dt, week_day, num_months, stop_criterion, num_times, until_dt, expected_dates,
+                       expected_flag, dummy_event):
+    dummy_event.timezone = 'Europe/Zurich'
+    clone_calculator = PatternCloneCalculator(dummy_event)
+    form = CloneRepeatPatternForm(dummy_event, csrf_enabled=False, week_day=week_day, num_months=num_months,
+                                  stop_criterion=stop_criterion, num_times=num_times, until_dt=until_dt)
+    expected_dates = [dummy_event.tzinfo.localize(dt) for dt in expected_dates]
+    form.start_dt.data = dummy_event.tzinfo.localize(start_dt)
+    dates, last_day_of_the_month = clone_calculator._calculate(form)
+    assert last_day_of_the_month is expected_flag
+    assert dates == expected_dates
+
+
+@pytest.mark.parametrize(('event_date', 'event_timezone'), (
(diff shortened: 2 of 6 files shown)


indico/htdocs/js/indico/modules/events/cloning.js
```diff
@@ -61,13 +61,15 @@
...
                         var $countNumber = $eventCount.find('.count');
+                        var $info = $eventCount.find('.info');
                         $countNumber.text(data.count);
                         $cloneErrors.hide();
                         $eventCount.show();
 
                         $eventList.toggle(!!data.count);
                         $cloneButton.prop('disabled', !data.count);
                         $eventCount.data('event-dates', data.dates);
+                        data.additional_info ? $info.text(data.additional_info) : $info.text('');
```
ThiefMaster: `?:` should not be used as a replacement for if..else but only when actually using the return value of the expressions.

`$info.text(data.additional_info ? data.additional_info : '');` would be OK, but this can be simplified to

    $info.text(data.additional_info || '');

indico/modules/events/management/controllers/cloning.py
```diff
@@ -99,23 +98,38 @@ class PatternCloneCalculator(CloneCalculator):
...
     def _calculate(self, form):
+        result = {}
         args = {'dtstart': self._naivify(form.start_dt.data)}
         args.update(self._calc_stop_criteria(form))
-        return self._tzify(rrule.rrule(rrule.MONTHLY,
-                                       interval=form.num_months.data,
-                                       byweekday=form.week_day.week_day_data,
-                                       bysetpos=form.week_day.day_number_data,
-                                       **args))
+        result['dates'] = self._tzify(rrule.rrule(rrule.MONTHLY,
```
ThiefMaster: You don't have anything else in `result`, so you can just `return {'dates': ...}`

indico/modules/events/management/controllers/cloning.py
```diff
@@ -99,23 +98,38 @@ class PatternCloneCalculator(CloneCalculator):
...
         freq, interval = relativedelta_to_rrule_interval(form.recurrence.data)
-        return self._tzify(rrule.rrule(freq, interval=interval, **args))
+        # check if last day of month
+        dtstart = self._naivify(form.start_dt.data)
+        next_day = dtstart + timedelta(days=1)
+        if freq == RRULE_FREQ_MAP['months'] and next_day.day == 1:
+            args = {'dtstart': next_day}
+            args.update(self._calc_stop_criteria(form))
+            dates = rrule.rrule(freq, interval=interval, **args)
+            info = (_("All cloned events will recur on the last day of the month."))
```
ThiefMaster: no need for the extra `()`.

indico/modules/events/management/controllers/cloning.py
```diff
@@ -99,23 +98,38 @@ class PatternCloneCalculator(CloneCalculator):
...
     def _calculate(self, form):
-        args = {'dtstart': self._naivify(form.start_dt.data)}
-        args.update(self._calc_stop_criteria(form))
+        result = {}
         freq, interval = relativedelta_to_rrule_interval(form.recurrence.data)
-        return self._tzify(rrule.rrule(freq, interval=interval, **args))
+        # check if last day of month
+        dtstart = self._naivify(form.start_dt.data)
+        next_day = dtstart + timedelta(days=1)
+        if freq == RRULE_FREQ_MAP['months'] and next_day.day == 1:
```
ThiefMaster: `freq == rrule.MONTHLY`

indico/modules/events/management/controllers/cloning.py
```diff
@@ -125,12 +139,14 @@ def _process(self):
         form = CloneRepeatabilityForm()
         clone_calculator = get_clone_calculator(form.repeatability.data, self.event)
         try:
-            dates = clone_calculator.calculate(request.form)
+            result = clone_calculator.calculate(request.form)
+            dates = result['dates']
+            info = result['info'] if 'info' in result else None
```
ThiefMaster: `info = result.get('info')`

indico/modules/events/management/templates/clone_event.html
```diff
@@ -91,8 +91,13 @@
...
                                 <span class="count"></span>
                             {% endset %}
+                            {% set info %}
+                                <span class="info"></span>
+                            {% endset %}
                             {% trans link='<a class="event-list">'|safe, endlink="</a>"|safe -%}
-                                {{ count }} new events will be created. {{ link }}When?{{ endlink}}
+                                {{ count }} new events will be created.
+                                <strong>{{ info }}</strong>
+                                {{ link }}When?{{ endlink}}
```
ThiefMaster: missing space after `endlink`

indico/modules/events/management/templates/clone_event.html
```diff
@@ -91,8 +91,13 @@
                             {% set count %}
                                 <span class="count"></span>
                             {% endset %}
+                            {% set info %}
+                                <span class="info"></span>
+                            {% endset %}
                             {% trans link='<a class="event-list">'|safe, endlink="</a>"|safe -%}
-                                {{ count }} new events will be created. {{ link }}When?{{ endlink}}
+                                {{ count }} new events will be created.
+                                <strong>{{ info }}</strong>
```
ThiefMaster: this feels dirty since we are combining two translated strings. i don't see much better options here

ThiefMaster: actually, if you returned just a flag about using last-day-of-month instead of a string, you could simply have two translated strings here instead of having to combine strings.

indico/htdocs/js/indico/modules/events/cloning.js
mic4ael: `$info.text(data.additional_info || '')`

indico/modules/events/management/controllers/cloning.py
```diff
@@ -125,12 +139,14 @@ def _process(self):
         form = CloneRepeatabilityForm()
         clone_calculator = get_clone_calculator(form.repeatability.data, self.event)
         try:
-            dates = clone_calculator.calculate(request.form)
+            result = clone_calculator.calculate(request.form)
```
ThiefMaster: I would return a tuple instead of a dict:

    dates, last_day_of_month = clone_calculator.calculate(request.form)

indico/modules/events/management/templates/clone_event.html
mvidalgarcia: Missing space after `endlink`

indico/modules/events/management/controllers/cloning.py
mic4ael: `result.get('info')`

indico/modules/events/management/templates/clone_event.html
meluru: Do you think it's better to use a flag? I considered that at first, but then I figured that a variable is more generic and could be used for some other messages as well (if ever we would need them :) )

ThiefMaster: The disadvantage of the free-text message is that you need to concatenate two translated messages which may or may not sound weird in other languages.

meluru: Yeah, that's a fair reason. Thanks

indico/modules/events/management/controllers/cloning.py
```diff
@@ -101,21 +100,32 @@ class PatternCloneCalculator(CloneCalculator):
...
         freq, interval = relativedelta_to_rrule_interval(form.recurrence.data)
-        return self._tzify(rrule.rrule(freq, interval=interval, **args))
+        # check if last day of month
+        dtstart = self._naivify(form.start_dt.data)
+        next_day = dtstart + timedelta(days=1)
+        if freq == rrule.MONTHLY and next_day.day == 1:
+            args = {'dtstart': next_day}
+            args.update(self._calc_stop_criteria(form))
+            dates = rrule.rrule(freq, interval=interval, **args)
+            last_day_of_month = True
```
ThiefMaster: why not just add `True` in the next line (and `False` further below)?

indico/modules/events/management/controllers/cloning.py
```diff
@@ -101,21 +100,32 @@ class PatternCloneCalculator(CloneCalculator):
     def _calculate(self, form):
         args = {'dtstart': self._naivify(form.start_dt.data)}
         args.update(self._calc_stop_criteria(form))
-        return self._tzify(rrule.rrule(rrule.MONTHLY,
-                                       interval=form.num_months.data,
-                                       byweekday=form.week_day.week_day_data,
-                                       bysetpos=form.week_day.day_number_data,
-                                       **args))
+        return (self._tzify(rrule.rrule(rrule.MONTHLY,
```
ThiefMaster: This will be more readable if you write it like this:

    dates = self._tzify(...)
    return dates, false

indico/modules/events/management/controllers/cloning.py
ThiefMaster: or, you could assign `dates = self._tzify(...)`, move the stuff after the return into an `else` block, and below all that `return dates, last_day_of_month`

indico/modules/events/management/templates/clone_event.html
```diff
@@ -91,8 +91,13 @@
                             {% set count %}
                                 <span class="count"></span>
                             {% endset %}
-                            {% trans link='<a class="event-list">'|safe, endlink="</a>"|safe -%}
-                                {{ count }} new events will be created. {{ link }}When?{{ endlink}}
+                            {% trans link='<a class="event-list">'|safe, endlink="</a>"|safe,
+                                last_day='<span class="last-day">'|safe, endlast_day="</span>"|safe -%}
```
ThiefMaster: I'd align this with `link` and consistently use single quotes (for the end tags). `endlast_day` also looks a bit weird. Maybe better to go for `lastday` and `endlastday` or `end_last_day` and `end_link`?

indico/modules/events/management/controllers/cloning.py
meluru: I can assign the dates, I didn't want to pass simply True/False to be more verbose about the parameter

ThiefMaster: I think this kind of verbosity would rather belong into a docstring, e,g, like this

    :return: a ``(dates, last_day_of_month)`` tuple

ThiefMaster:
This is safe to rebase to 2.1-dev now (since that branch is rebased to master since earlier today); don't forget to also change the target branch of this PR.

indico/modules/events/management/controllers/cloning.py
```diff
@@ -101,21 +104,34 @@ class PatternCloneCalculator(CloneCalculator):
...
+            args = {'dtstart': next_day}
+            args.update(self._calc_stop_criteria(form))
+            dates = rrule.rrule(freq, interval=interval, **args)
+            last_day_of_month = True
+            dates = self._tzify([date - timedelta(days=1) for date in dates])
+        else:
+            args = {'dtstart': dtstart}
+            args.update(self._calc_stop_criteria(form))
+            dates = self._tzify(rrule.rrule(freq, interval=interval, **args))
+        return dates, last_day_of_month
```
ThiefMaster: i'd still get rid of the flag and have to return statements (`return dates, True`  and `return dates, False`).

in any case, if you prefer to keep it like this i would move the `last_day_of_month = False` into the `else` block

indico/modules/events/management/templates/clone_event.html
```diff
@@ -91,8 +91,13 @@
                             {% set count %}
                                 <span class="count"></span>
                             {% endset %}
-                            {% trans link='<a class="event-list">'|safe, endlink="</a>"|safe -%}
-                                {{ count }} new events will be created. {{ link }}When?{{ endlink}}
+                            {% trans link='<a class="event-list">'|safe, endlink='</a>'|safe,
+                                     lastday='<span class="last-day">'|safe, endlastday='</span>'|safe -%}
+                                {{ count }} new events will be created.
+                                {{ lastday }}
+                                    <strong>All cloned events will recur on the last day of the month.</strong>
```
ThiefMaster: is "recur" correct here? since this is about the cloned events, i think "occur" or "take place" is more suitable

indico/modules/events/management/controllers/cloning.py
```diff
@@ -101,21 +104,34 @@ class PatternCloneCalculator(CloneCalculator):
...
     def _calculate(self, form):
-        args = {'dtstart': self._naivify(form.start_dt.data)}
-        args.update(self._calc_stop_criteria(form))
         freq, interval = relativedelta_to_rrule_interval(form.recurrence.data)
-        return self._tzify(rrule.rrule(freq, interval=interval, **args))
+        # check if last day of month
+        dtstart = self._naivify(form.start_dt.data)
+        next_day = dtstart + timedelta(days=1)
+        if freq == rrule.MONTHLY and next_day.day == 1:
+            args = {'dtstart': next_day}
```
mic4ael: I see you are unpacking it later on when calling `rrule` method. Those are keyword arguments, maybe it would be better to call it `kwargs` in that case since `args` is usually used for positional arguments.

indico/modules/events/management/controllers/cloning.py
```diff
@@ -101,21 +104,34 @@ class PatternCloneCalculator(CloneCalculator):
...
+            args.update(self._calc_stop_criteria(form))
```
mic4ael: Also, you could save one line by using something like this `kwargs  = dict(self._calc_stop_criteria(form), dtstart=next_day)`

indico/modules/events/cloning_test.py
```diff
@@ -0,0 +1,92 @@
...
+                                  stop_criterion=stop_criterion, num_times=num_times, until_dt=until_dt)
+    expected_dates = [dummy_event.tzinfo.localize(dt) for dt in expected_dates]
+    form.start_dt.data = dummy_event.tzinfo.localize(start_dt)
+    dates, last_day_of_the_month = clone_calculator._calculate(form)
+    assert last_day_of_the_month is expected_flag
+    assert dates == expected_dates
+
+
+@pytest.mark.parametrize(('event_date', 'event_timezone'), (
+    (datetime(2018, 03, 25, 2, 0, 0), 'Europe/Zurich'),
```
ThiefMaster: leading zeroes are not a good idea, better put an extra space if you want to align it

pferreir: Just to clarify @ThiefMaster's comment:
```python
>>> 012
10```
(octal)

Comment (ThiefMaster) [reply indico/modules/events/cloning_test.py]: And they are a syntax error in Python 3, due to how easy it was to mess things up by not realizing that the 0-prefix created an octal number.

Comment (meluru) [reply indico/modules/events/cloning_test.py]: Thanks, I actually wanted to avoid leading zeros, I think I just overlooked that one.

**Comment by meluru** [comment]
Updated

File: indico/modules/events/management/templates/clone_event.html [hunk 17]
```diff
@@ -91,8 +91,13 @@
                             {% set count %}
                                 <span class="count"></span>
                             {% endset %}
-                            {% trans link='<a class="event-list">'|safe, endlink="</a>"|safe -%}
-                                {{ count }} new events will be created. {{ link }}When?{{ endlink}}
+                            {% trans link='<a class="event-list">'|safe, endlink='</a>'|safe,
+                                     lastday='<span class="last-day">'|safe, endlastday='</span>'|safe -%}
+                                {{ count }} new events will be created.
+                                {{ lastday }}
```
ThiefMaster: I think you can strip whitespace on the right side here (`-}}`), and on the left side of `endlastday`.

meluru:
Updated

PR #3161, opened 2017-11-21
Add signal for checkin state changed

bpedersen2:
This signal can be used e.g. in a plugin to print badges or further
material on demand at a check-in desk.
See https://talk.getindico.io/t/print-on-registration-plugin/99 for
the discussion.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -48,6 +48,9 @@ Internal Changes
 - Rename *Roles* in ACL entries to *Permissions*.  This especially affects
   the ``can_manage`` method whose ``role`` argument has been renamed to
   ``permission`` (:issue:`3057`)
+- Add new ``registration_checkin_updated`` signal that can be used by
+  plugins to perform an action when the checkin state of a registration
+  changes (:issue:`3161`, thanks :user:`bpedersen2`)
 
 
 ----
--- a/indico/core/signals/event/registration.py
+++ b/indico/core/signals/event/registration.py
@@ -27,9 +27,14 @@ Called when the registration personal data is modified. The `sender` is the
 """)
 
 registration_state_updated = _signals.signal('registration-state-updated', """
-Called when the state of registration changes. The `sender` is the
+Called when the state of a registration changes. The `sender` is the
 `Registration` object; the previous state is passed in the `previous_state`
 kwarg.
+""")
+
+registration_checkin_updated = _signals.signal('registration-checkin-updated', """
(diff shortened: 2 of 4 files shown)


ThiefMaster:
You still need to actually send the signal ;)

indico/core/signals/event/registration.py
```diff
@@ -66,3 +66,8 @@
 If this signal returns ``True``, the user will not be able to download
 their ticket.
 """)
+
+registration_checkin_updated = _signals.signal('registration-checkin-updated', """
+Called when the checkin state of registration changes. The `sender` is the
```
ThiefMaster: > of *a* registration

(same typo in `registration_state_updated`, feel free to fix it there too).

I'd also move your new signal below that one since they are kind of similar so I think it makes more sense to have it up there.

bpedersen2: Done

indico/modules/events/registration/api.py
```diff
@@ -57,6 +57,7 @@ def _process_PATCH(self):
 
         if 'checked_in' in request.json:
             self._registration.checked_in = bool(request.json['checked_in'])
+            signals.event.registration_checkin_updated(self._registration)
```
ThiefMaster: Did you test this? I don't think signals implement `__call__` so you need to call `.send()` on them, e.g.

    signals.event.registration_checkin_updated.send(self._registration)

ThiefMaster:
Please add a CHANGES.rst entry (in the "Internal Changes" section):

```
- Add new ``registration_checkin_updated`` signal that can be used by
  plugins to perform an action when the checkin state of a registration
  changes (:issue:`3161`, thanks :user:`bpedersen2`)
```

bpedersen2:
Fixed the missing send and added changelog.

The signal now arrives in a simple test plugin

CHANGES.rst
```diff
@@ -34,6 +34,9 @@ Internal Changes
   the ``can_manage`` method whose ``role`` argument has been renamed to
   ``permission`` (:issue:`3057`)
 
+- Add new ``registration_checkin_updated`` signal that can be used by
```
ThiefMaster: there shouldn't be an empty line before this one (see the structore of the rest of the file)

bpedersen2: Should be fixed

ThiefMaster:
thanks!

Issue #3146, opened 2017-11-08
Ticket designer: Allow to pick any registration field


bpedersen2:
Currently, only a fixed and limited set of  registration data is available for tickets/badges.
It would be nice to  get additional data from the registration, e.g. booked social events or dietary preferences for badges/tickets.
One idea would be a parameterized item in the designer.

tomasr8:
Update after Angular removal. There are a couple more ideas how to go about this:

- An option to set a name on a registration field which would then uniquely identify the field within that registration form, and you could reference a field by name in other places such as badge generation but also csv import/export

- Being able to optionally link a regform to a ticket template. By default the template has access to global fields only, but when linked to a regform all fields become automatically available. Here we need to decide what to do when changing/removing the link. Should the fields be deleted/set as invalid/something else?

bpedersen2:
One of our main use cases is using a common template across more than one registration form( e.g ealrybird/student/regular), and often even across events,so a field-role kind  of naming would be really important.

PR #3164, opened 2017-11-24
Introduce a parameterized Badge placeholder

bpedersen2:
To allow arbitrary fields from the registraion on badges/tickets etc,
introduce a parameterized placeholder for the designer.

Currently in this simplest form the parameter should either be the
numeric field id or the field name, possibly followed by the subkey
separated with ':'.

This is related to  issue #3146.


indico/htdocs/js/indico/modules/designer.js
```diff
@@ -41,6 +41,16 @@
 
     var DEFAULT_PIXEL_CM = 50;
 
+    function _needs_param(field) {
```
ThiefMaster: in JS we use camelCase -> `_needsParam`

indico/htdocs/js/indico/modules/designer.js
```diff
@@ -41,6 +41,16 @@
...
+        if (field.type === "fixed") {
+            return true;
+        }
+        if (field.type === "genericfield"){
+            return true;
+        }
+        return false
```
ThiefMaster: the whole function can be simplified to `return field.type == 'fixed' || field.type == 'genericfield';`

indico/modules/designer/placeholders.py
```diff
@@ -297,3 +299,36 @@ class RegistrationTicketQRPlaceholder(Placeholder):
     @classmethod
     def render(cls, registration):
         return generate_ticket_qr_code(registration)
+
+class RegistrationFieldPlaceholder(ParametrizedPlaceholder):
+    group = 'registration'
+    name = 'genericfield'
```
ThiefMaster: I think `custom_field` would be a better name

bpedersen2:
What is open is how to best get a lis tof available fields, as this seems to depend heavily on where the templace is edited. Am I  correct in the assumption that that would probably need to go via signals.get_placeholders ?

indico/modules/designer/placeholders.py
```diff
@@ -297,3 +299,36 @@ class RegistrationTicketQRPlaceholder(Placeholder):
...
+    advanced = True
+
+    @classmethod
+    def render(cls, param, registration):
+        if ':' in param:
+            field_id, key = param.split(':', 1)
+        else:
+            field_id = param
+            key = None
+        import pdb; pdb.set_trace()
```
ThiefMaster: :monkey:

indico/modules/designer/placeholders.py
```diff
@@ -297,3 +299,36 @@ class RegistrationTicketQRPlaceholder(Placeholder):
...
+        try:
+            fid = int(field_id)
+        except (TypeError, ValueError):
+            for avail_field in registration.registration_form.active_fields:
+                if avail_field.title == field_id:
```
ThiefMaster: I don't think we should match by field title at all. We don't do so when emailing registrants either.

ThiefMaster:
Have a look at `FieldPlaceholder` and `render_placeholder_info` for getting a list of available placeholders.

In your case you can't use that function since you need to generate dropdown options from it, but the logic is pretty much the same, except that you would explicitly check whether it's a parametrized placeholder or maybe even add a new `show_in_designer` attribute to your placeholder (and then check `getattr(placeholder, 'show_in_designer', False)` to determine whether to show that placeholder as an option in the designer or not.

ThiefMaster:
Also, as usual please add a chancelog entry ("Improvements" section):

```rst
- Allow using any registration field in the badge designer (:issue:`3146`,
  thanks :user:`bpedersen2`)
```

indico/modules/designer/placeholders.py
```diff
@@ -297,3 +299,36 @@ class RegistrationTicketQRPlaceholder(Placeholder):
     @classmethod
     def render(cls, registration):
         return generate_ticket_qr_code(registration)
+
+class RegistrationFieldPlaceholder(ParametrizedPlaceholder):
```
ThiefMaster: PEP8: there should be two empty lines between top-level classes (and functions)

ThiefMaster: You might want to subclass `FieldPlaceholder`; like this you can skip almost all of your custom code in here.

Since you probably don't want to include all regform fields here (personal data is already handled with more explicit placeholder classes), you might want to move the `own_placeholder_types` set to the class level so you can override that in your subclass.

indico/modules/designer/placeholders.py
bpedersen2: Well, that was done on purpose here for two reason:
1) currently the user needs to enter this by hand, without good debuging knowledge the numeric value is hard to get.
2) When writing generic templates for more than a single registration form one can not match on the id here.

Getting available fields is planned as a follow-up

ThiefMaster: 1. Field titles are not necessarily unique, especially between sections - but 
2. Good point... that's actually a bigger problem I hadn't thought about. Can you join our IRC channel (#indico on freenode)? I think this (combined with (1)) needs more discussion and IRC is a better place for that

bpedersen2:
Still work in progress.

Missing:  Match semantic name entries with db and reuse them (currently, trying to use the same value on 2 fields fails)

indico/modules/events/registration/models/form_fields.py
```diff
@@ -59,13 +58,45 @@ def __repr__(self):
         return '<RegistrationFormFieldData({}, {})>'.format(self.id, self.field_id)
 
 
+class RegistrationFormFieldSemanticNames(db.Model):
```
ThiefMaster: Just saw that you updated this PR. I don't think adding a new table for this is a good idea - instead, I'd rather put a String column directly on `RegistrationFormItem`.

bpedersen2: Well, the idea is to add a autocompleter for know "semantic names" as these should be reused across conferences and registration forms.  But I think we should postpone this for after the regform angular cleanup anyway.

ThiefMaster: I don't like the idea of autocompleting this across events. And within an event having the extra table is overkill - getting it straight from the form_items table doesn't add noticable overhead

ThiefMaster: But yes, better to keep this for the future until after we got rid of the Angular mess :)

ThiefMaster:
Closing in favor of #6088

Issue #3165, opened 2017-11-24
Renaming pages should affect headings


Eising:
Hi,

As you well know, it is possible to rename things in the menu, for example renaming "Call for Abstracts" to something else, like "Call for Programme" or "Bob" or whatever you like.

This however does only affect the menu entry, and not the heading in the actual page.

I would like to suggest that you implement this for at least the heading also, and if not too difficult, further entries, such as buttons where it applies.

ThiefMaster:
> and if not too difficult, further entries, such as buttons where it applies.

I think that's going into the "overkill" area - for example, "submit abstract" etc can't really be changed since it's a different string, and usually people do use it for abstracts even if the menu is labelled differently.

Eising:
> I think that's going into the "overkill" area - for example, "submit abstract" etc can't really be changed since it's a different string, and usually people do use it for abstracts even if the menu is labelled differently.

That's fair. Fixing the heading alone would remove a bit of confusion I believe.

PR #3169, opened 2017-12-01
Keep page header and page title in sync

bpedersen2:
Use the same logic that is used for tracks/program display

Fixes: #3165


ThiefMaster:
superseded by #3236

PR #3170, opened 2017-12-01
Abstract fields

meluru:


indico/migrations/versions/20171130_1715_093533d27a96_add_visibility_options_to_contribution_.py
```diff
@@ -0,0 +1,33 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '093533d27a96'
+down_revision = '566d5de4e0e5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('contribution_fields', sa.Column('filled_by_submitter', sa.Boolean(), nullable=True), schema='events')
```
ThiefMaster: should this really be nullable?

indico/modules/events/contributions/contrib_fields.py
```diff
@@ -41,11 +42,16 @@ class ContribFieldConfigForm(IndicoForm):
     is_active = BooleanField(_('Active'), widget=SwitchWidget(),
                              description=_("Whether the field is available."),
                              default=True)
+    filled_by_submitter = BooleanField(_('Filled by submitter'), widget=SwitchWidget(),
```
ThiefMaster: That name sounds really misleading (as in "did the user fill it out"). What about something like `user_editable` instead?
Same for the title.

indico/modules/events/contributions/models/fields.py
```diff
@@ -74,6 +84,15 @@ class ContributionField(db.Model):
...
         default=True
     )
+    visibility = db.Column(
+        PyIntEnum(FieldVisibility),
+        nullable=False,
+        default=FieldVisibility.public
+    )
+    filled_by_submitter = db.Column(
+        db.Boolean,
+        default=True
```
ThiefMaster: `nullable=False` is missing here as well

indico/modules/events/abstracts/util.py
```diff
@@ -379,3 +382,15 @@ def get_events_with_abstract_persons(user, dt=None):
     for person in query:
         data[person.event_id].add('abstract_person')
     return data
+
+
+def filter_field_values(fields, user, managers, submitter):
+    if user in managers:
```
ThiefMaster: Just like above, I would check for management area instead of management rights

indico/modules/events/abstracts/util.py
```diff
@@ -379,3 +382,15 @@ def get_events_with_abstract_persons(user, dt=None):
...
+        return {field for field in fields if field.contribution_field.is_active}
+    if user == submitter:
```
ThiefMaster: I think we want the `abstract.user_owns()` check here and not an explicit check for the submitter.

indico/modules/events/abstracts/util.py
```diff
@@ -379,3 +382,15 @@ def get_events_with_abstract_persons(user, dt=None):
...
+        return {field for field in fields
```
ThiefMaster: I would pre-filter the fields based on `is_active` before the final set comprehensions. More readable that way.

indico/modules/events/abstracts/views.py
```diff
@@ -83,14 +84,18 @@ def render_abstract_page(abstract, view_class=None, management=False):
     judgment_form = AbstractJudgmentForm(abstract=abstract, formdata=None)
     review_track_list_form = AbstractReviewedForTracksForm(event=abstract.event, obj=abstract, formdata=None)
     track_session_map = {track.id: track.default_session_id for track in abstract.event.tracks}
+    managers = {x.principal for x in abstract.event.acl_entries if x.full_access}
```
ThiefMaster: see above

indico/modules/events/contributions/contrib_fields.py
```diff
@@ -37,15 +38,21 @@ class ContribFieldConfigForm(IndicoForm):
     title = StringField(_('Title'), [DataRequired()], description=_("The title of the field"))
     description = TextAreaField(_('Description'), description=_("The description of the field"))
     is_required = BooleanField(_('Required'), widget=SwitchWidget(),
-                               description=_("Whether the user has to fill out the field"))
+                               description=_("Whether the field has to be filled out"))
     is_active = BooleanField(_('Active'), widget=SwitchWidget(),
                              description=_("Whether the field is available."),
                              default=True)
+    user_editable = BooleanField(_('User editable'), widget=SwitchWidget(),
+                                 description=_("Whether the field is filled by user on abstract submission."),
```
ThiefMaster: "Whether the user can fill out the field during abstract submission"?

indico/modules/events/contributions/templates/forms/contribution_field.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'forms/form_common_fields_first.html' %}
+
+{% block javascript %}
+    <script>
+        var $userEditable = $('#user_editable');
```
ThiefMaster: wrap in a strict mode block; also because this currently creates globals.

ThiefMaster: Hardcoded IDs in widget JS are generally not a good idea. Also, have you checked the `HiddenUnless` validator? It automatically enables JS to hide the field if the condition is not met.

indico/modules/events/contributions/templates/forms/contribution_field.html
```diff
@@ -0,0 +1,20 @@
...
+        var $managers = $('#visibility').find('.managers').find('input');
+        if ($userEditable[0].checked){
+            $managers.prop('disabled', true);
+        } else {
+            $managers.prop('disabled', false);
+        }
+        $userEditable.on('change', function() {
+            if (this.checked) {
+                $managers.prop('disabled', true);
```
ThiefMaster: $managers.prop('disabled', this.checked);

indico/modules/events/contributions/templates/forms/contribution_field.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'forms/form_common_fields_first.html' %}
+
+{% block javascript %}
+    <script>
+        var $userEditable = $('#user_editable');
+        var $managers = $('#visibility').find('.managers').find('input');
+        if ($userEditable[0].checked){
+            $managers.prop('disabled', true);
```
ThiefMaster: $managers.prop('disabled', $userEditable[0].checked);

indico/modules/events/contributions/templates/forms/contribution_field.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'forms/form_common_fields_first.html' %}
+
+{% block javascript %}
+    <script>
+        var $userEditable = $('#user_editable');
+        var $managers = $('#visibility').find('.managers').find('input');
+        if ($userEditable[0].checked){
```
ThiefMaster: missing space before `{`

indico/modules/events/abstracts/util.py
```diff
@@ -186,13 +187,15 @@ def make_abstract_form(event, notification_option=False, management=False):
...
     form_class = type(b'_AbstractForm', tuple(mixins) + (AbstractForm,), {})
+    managers = {x.principal for x in event.acl_entries if x.full_access}
     for custom_field in event.contribution_fields:
         field_impl = custom_field.mgmt_field if management else custom_field.field
         if field_impl is None:
             # field definition is not available anymore
             continue
-        name = 'custom_{}'.format(custom_field.id)
-        setattr(form_class, name, field_impl.create_wtf_field())
+        if custom_field.user_editable or user in managers:
```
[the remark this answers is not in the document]
meluru: I also thought about the management mode at first, but a manger can edit contribution field from display page. So it's that the manager would be able to change some fields only in management mode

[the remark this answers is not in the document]
ThiefMaster: ok fair enough. anyway, don't do this weird `users in managers` check but use `event.can_manage(user)` ;)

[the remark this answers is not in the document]
pferreir: I agree with @meluru, since it's a pop-up anyway.

indico/modules/events/abstracts/util.py
meluru: And here also, wouldn't it be confusing for a manager to see different fields on display and different on manage?

ThiefMaster: yeah, let's stay with your logic and  just fix the way of checking for management rights ;)

indico/modules/events/contributions/templates/forms/contribution_field.html
meluru: But Hidden unless applies only when we want to hide a field, not an option of the field, right?

ThiefMaster: ah it's just an option.. you're right then. anyway, still not a good idea to hardcode the ids. you should be able to get them dynamically via the field (`field.id`, `field.get_form().otherfield.id`)

indico/modules/events/contributions/templates/forms/contribution_field.html
```diff
@@ -0,0 +1,15 @@
+{% extends 'forms/form_common_fields_first.html' %}
+
+{% block javascript %}
+    <script>
+        (function() {
+            'use strict';
+            var $userEditable = $('#user_editable');
```
ThiefMaster: `$('#{{ form.user_editable.id }}')`

meluru:
Updated

indico/modules/events/contributions/templates/forms/contribution_field_form.html
```diff
@@ -0,0 +1,19 @@
+{% extends 'forms/form_common_fields_first.html' %}
+
+{% block javascript %}
+    <script>
+        (function() {
+            'use strict';
+            var $userEditable = $('#{{ form.user_editable.id }}');
+            var $managers = $('#{{ form.visibility.id }}').find('.managers').find('input');
```
mvidalgarcia: Can't you do `.find('.managers input')`?

indico/modules/events/contributions/templates/forms/contribution_field_form.html
```diff
@@ -0,0 +1,19 @@
...
+            var $visibility = $("input[name=visibility]");
+            $visibility.on('change', function () {
```
mvidalgarcia: Extra space after `function`.

indico/migrations/versions/20171130_1715_093533d27a96_add_visibility_options_to_contribution_.py
```diff
@@ -0,0 +1,36 @@
...
+revision = '093533d27a96'
+down_revision = '566d5de4e0e5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('contribution_fields',
+                  sa.Column('user_editable', sa.Boolean(), nullable=False, server_default='true'),
+                  schema='events')
```
mic4ael: not aligned properly

indico/migrations/versions/20171130_1715_093533d27a96_add_visibility_options_to_contribution_.py
```diff
@@ -0,0 +1,36 @@
...
+    op.add_column('contribution_fields',
+                  sa.Column('visibility', PyIntEnum(ContributionFieldVisibility),
+                            nullable=False, server_default='1'),
+                  schema='events')
```
mic4ael: ditto

indico/modules/events/contributions/contrib_fields.py
```diff
@@ -37,15 +38,22 @@ class ContribFieldConfigForm(IndicoForm):
...
+                               description=_("Whether the field has to be filled out"))
     is_active = BooleanField(_('Active'), widget=SwitchWidget(),
                              description=_("Whether the field is available."),
                              default=True)
+    visibility = IndicoEnumRadioField(_('Visibility'), [DataRequired()], default=ContributionFieldVisibility.public,
+                                      enum=ContributionFieldVisibility,
+                                      description=_('Who will be able to see the field'))
+    user_editable = BooleanField(_('User editable'), widget=SwitchWidget(),
+                                 description=_("Whether the submitter/author can fill out the field during abstract"
+                                               " submission."),
```
mic4ael: usually we put a space at the end of a previous line rather than at the beginning of the next

indico/modules/events/contributions/contrib_fields.py
```diff
@@ -37,15 +38,22 @@ class ContribFieldConfigForm(IndicoForm):
...
     description = TextAreaField(_('Description'), description=_("The description of the field"))
     is_required = BooleanField(_('Required'), widget=SwitchWidget(),
-                               description=_("Whether the user has to fill out the field"))
+                               description=_("Whether the field has to be filled out"))
     is_active = BooleanField(_('Active'), widget=SwitchWidget(),
                              description=_("Whether the field is available."),
                              default=True)
+    visibility = IndicoEnumRadioField(_('Visibility'), [DataRequired()], default=ContributionFieldVisibility.public,
+                                      enum=ContributionFieldVisibility,
+                                      description=_('Who will be able to see the field'))
```
mic4ael: at least to me it doesn't sound right, maybe something like: "Determines who can fill in the field"

meluru: But it actually doesn't determine who will be able to fill the field, but who will be able to see the value of the field. How about 'Determines who can see the value of the field'?

PR #3172, opened 2017-12-04
Show validation message for radio_buttons_widget

mic4ael:
Validation message can't be shown because the `input[type=radio]` is not visible, thus not focusable.

### DIFF
--- a/indico/web/templates/forms/radio_buttons_widget.html
+++ b/indico/web/templates/forms/radio_buttons_widget.html
@@ -3,7 +3,7 @@
 {% set vertical_alignment = field.option_orientation == 'vertical' %}
 
 {% block html %}
-    <div class="{{ 'i-group' if vertical_alignment else 'group i-selection' }}">
+    <div id="{{ field.id }}" class="{{ 'i-group' if vertical_alignment else 'group i-selection' }}">
         {% for subfield in field %}
             <div class="{{ 'i-radio' if vertical_alignment else 'inline-vcentered'}} radio-item-{{ loop.index0 }}">
                 {{ subfield(**input_args) }}
@@ -11,4 +11,18 @@
             </div>
         {% endfor %}
     </div>
+{% endblock %}
+
+{% block javascript %}
+    <script>
+        (function() {
+            'use strict';
+
+            $('#{{ field.id }} input').on('invalid', function(evt) {
+                evt.preventDefault();
+
+                $('#{{ field.id }} div:last-child').stickyTooltip('warning', this.validationMessage);
+            });
+        })();
+    </script>
 {% endblock %}


indico/web/templates/forms/radio_buttons_widget.html
```diff
@@ -3,12 +3,33 @@
...
                 {{ subfield(**input_args) }}
                 {{ subfield.label(class_=('i-button' if not vertical_alignment)) }}
             </div>
         {% endfor %}
     </div>
+    <script>
+        (function() {
+            'use strict';
+
+            $('#{{ field.id }} input').bind('invalid', function(evt) {
```
ThiefMaster: `.on(...)`

indico/web/templates/forms/radio_buttons_widget.html
```diff
@@ -3,12 +3,33 @@
...
-    <div class="{{ 'i-group' if vertical_alignment else 'group i-selection' }}">
+    <div id="{{ field.id }}" class="{{ 'i-group' if vertical_alignment else 'group i-selection' }}">
         {% for subfield in field %}
             <div class="{{ 'i-radio' if vertical_alignment else 'inline-vcentered'}} radio-item-{{ loop.index0 }}">
                 {{ subfield(**input_args) }}
                 {{ subfield.label(class_=('i-button' if not vertical_alignment)) }}
             </div>
         {% endfor %}
     </div>
+    <script>
```
ThiefMaster: widget templates have a `javascript` jinja block

indico/web/templates/forms/radio_buttons_widget.html
```diff
@@ -3,12 +3,33 @@
...
+        (function() {
+            'use strict';
+
+            $('#{{ field.id }} input').bind('invalid', function(evt) {
+                $('#{{ field.id }} div:last-child').qtip({
+                    content: {
+                        text: $T.gettext(evt.target.validationMessage)
```
ThiefMaster: using i18n here is kind of pointless since the messages won't be in the dictionary

indico/web/templates/forms/radio_buttons_widget.html
```diff
@@ -3,12 +3,33 @@
...
+                    },
+                    show: {ready: true},
+                    position: {
+                        my: 'left center',
+                        at: 'right center',
+                        target: $('#{{ field.id }} div:last-child')
```
ThiefMaster: you use it above, so assign it to a variable instead of running the selector twice

indico/web/templates/forms/radio_buttons_widget.html
```diff
@@ -3,12 +3,33 @@
...
+                    }
+                });
+
+                return false;
```
ThiefMaster: we usually call `evt.preventDefault()` or `evt.stopPropagation()` since it's more explicit than returning false.

indico/web/templates/forms/radio_buttons_widget.html
mic4ael: yeah, but then we should also remember about moving JS code from existing templates to this block (e.g. switch_widget)

mic4ael:
updated

PR #3173, opened 2017-12-05
WIP: Webpack Integration

pferreir:
In order to run the code you'll have to install [this version of pywebpack](https://github.com/pferreir/pywebpack/tree/wip/indico).

Missing:
 * [x] Fixing static websites;
 * [x] Trailing slash sanitization;
 * [x] Plugin integration;
 * [x] Handling of custom (S)CSS;
 * [ ] Development docs;
 * [ ] Config docs (nginx), recommend no subdir;


docs/source/installation/development.rst
```diff
@@ -144,6 +144,14 @@ You can then initialize the DB::
 Running Indico
 --------------
 
+You will need two shells running in parallel. The first one will run the webpack watcher, which compiless
```
mic4ael: compiles

indico/web/client/styles/base/_layout.scss
```diff
@@ -135,3 +133,8 @@ body > * {
         width: 0;
     }
 }
+
+
+@mixin clearfix() {
+    overflow: hidden;
+}
```
ThiefMaster: missing trailing linebreak

indico/web/client/styles/screen.scss
```diff
@@ -17,10 +17,11 @@
...
-@import 'lib/compass/compass/css3/user-interface';
-
 @import 'partials';
 @import 'widgets';
 @import 'custom';
 @import 'modules';
+
+// Legacy CSS code
+@import './legacy/Default.css';
+@import './legacy/timetable.css';
```
ThiefMaster: missing trailing linebreak

indico/web/flask/app.py
```diff
@@ -239,6 +242,36 @@ def setup_assets():
...
+        pkg_path,
+        config=lambda: _get_webpack_config(app),
+        config_path=os.path.join(pkg_path, 'config.json')
+    )
+    app.config.update({
+        'WEBPACKEXT_PROJECT': project,
+        'WEBPACKEXT_MANIFEST_PATH': os.path.join('dist', 'manifest.json')
+    })
+
+    FlaskWebpackExt(app)
```
ThiefMaster: see my comment somewhere above about making this global and just calling `init_app` here

indico/web/flask/errors.py
```diff
@@ -38,22 +38,12 @@
 
 @errors_bp.app_errorhandler(NotFound)
```
ThiefMaster: I think this whole handler can now go away as it is covered by `handle_http_exception`

indico/web/templates/base.html
```diff
@@ -11,14 +11,20 @@
...
     <script>
-        window.CKEDITOR_BASEPATH = {{ url_for('assets.js', filename='lib/ckeditor/')|tojson() }};
+        window.CKEDITOR_BASEPATH = {{ url_for('assets.dist', filename='js/ckeditor/')|tojson() }};
     </script>
 
+    <!-- Webpack bundles -->
+        {{ webpack['common.js'] }}
+        {{ webpack['main.js'] }}
+        {{ webpack['main.css'] }}
+    <!-- end -->
```
ThiefMaster: is that comment (and the indentation of the last three lines) really useful? ;)

indico/web/views.py
```diff
@@ -19,14 +19,15 @@
 import posixpath
 from urlparse import urlparse
 
-from flask import g, render_template, request, session
+from flask import g, render_template, request, session, current_app
```
ThiefMaster: isort should complain about the order in this line

indico/web/views.py
```diff
@@ -142,6 +143,8 @@ def _getPageContent(self, params):
         if html is not None:
             return html
         template = params.pop('_jinja_template')
+        webpack = current_app.extensions['flask-webpackext']
+        params['bundles'] = map(lambda x: webpack.manifest[x], self.bundles)
```
ThiefMaster: I think the BDFL hates `map(lambda: ...)` in favor of list comprehensions ;)

`map(webpack.manifest.__getitem__, self.bundles)` would work but I think it's uglier than just using a list comprehension

indico/web/views.py
```diff
@@ -150,39 +153,26 @@ class WPBase(object):
...
-                for pkg_name in pkg_names
-                for url in self._asset_env[prefix + pkg_name.lower()].urls()]
+    @classproperty
+    @classmethod
+    def bundles(self):
+        _bundles = ('common.js', 'main.js', 'main.css', 'module_core.js', 'module_events.creation.js',
+                    'module_attachments.js')
+        if not g.get('static_site'):
+            _bundles += ('ckeditor.js',)
+        return _bundles
```
ThiefMaster: What about adding this for development?

```python
if config.DEBUG and len(_bundles) != len(set(_bundles)):
    raise Exception('Duplicate bundles found')
```

indico/web/views.py
```diff
@@ -219,14 +209,18 @@ def display(self, **params):
...
-        css_files = map(self._fix_path, self.getCSSFiles() + plugin_css + self.get_extra_css_files() + custom_css)
-        print_css_files = map(self._fix_path, self.getPrintCSSFiles())
-        js_files = map(self._fix_path, self.getJSFiles() + plugin_js + custom_js)
+        css_files = map(self._fix_path, plugin_css + self.get_extra_css_files() + custom_css)
+        js_files = map(self._fix_path, plugin_js + custom_js)
 
         body = to_unicode(self._display(params))
+        webpack = current_app.extensions['flask-webpackext']
+
+        bundles = map(lambda x: webpack.manifest[x], self.bundles)
```
ThiefMaster: list comprehension? ;)

postcss.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+    parser: 'postcss-scss',
+    plugins: {
+        'autoprefixer': {}
+    }
+}
```
ThiefMaster: missing trailing linebreak

webpack.config.js
```diff
@@ -0,0 +1,257 @@
...
+    h.update(fs.readFileSync(filePath));
+    return h.digest('hex').slice(0, 8);
+}
+
+function fileLoaderPublicPathGenerator(_prefix) {
+    return (filePath) => {
+        let pathParts = filePath.split(path.sep);
+        let prefix = _prefix;
+
+        // if no prefix specified, take the first parth segment
```
ThiefMaster: s/parth/path/

webpack.config.js
```diff
@@ -0,0 +1,257 @@
...
+                    }
+                }
+            ]
+        },
+        plugins: [
+            new ManifestPlugin({
+                fileName: 'manifest.json',
+                publicPath: config.build.distURL,
+                map: (file) => {
+                    let hash = null;
```
ThiefMaster: just `let hash;` since the default value is never used?

indico/modules/categories/js/index.js
```diff
@@ -0,0 +1,82 @@
...
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import './display';
+
+(function(global) {
+    global.setupCategoryStats = () => {
```
ThiefMaster: I would keep a named function expression here; especially having a name on it makes tracebacks more readable...

requirements.txt
```diff
@@ -72,3 +72,4 @@ flask-marshmallow==0.8.0
 ipython<6,>=5.4.1
 prompt-toolkit<2.0.0,>=1.0.14
 packaging>=16.8,<17
+flask-webpackext==0.1.0
```
ThiefMaster: What about also installing pywebpack from your fork here for now until your change is in a released version of it?

indico/modules/categories/js/index.js
```diff
@@ -0,0 +1,83 @@
...
+                            markerOptions: {
+                                color: '#005272',
+                                style: 'circle'
+                            }
+                        }
+                    ],
+                    width: 400
+                };
+                data = [
+                    _.pairs(data).map((datum) => {
```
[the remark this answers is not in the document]
pferreir: You realize this is just a copied file, right? 😉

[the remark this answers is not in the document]
ThiefMaster: you did change the function() to an arrow expression though :p

.babelrc
```diff
@@ -0,0 +1,9 @@
+{
+  "presets": ["es2015"],
```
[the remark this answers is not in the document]
pferreir: I'll actually use [env](https://babeljs.io/docs/plugins/preset-env/) instead.

indico/modules/categories/js/index.js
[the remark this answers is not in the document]
pferreir: You're right. Sorry 😉

indico/modules/events/themes.yaml
```diff
@@ -41,7 +41,7 @@ definitions:
     title: Compact style
     user_visible: true
 
-  indico-weeks-view:
+  indico_weeks_view:
```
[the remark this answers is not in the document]
pferreir: I don't like dashes in file names in general. But yeah, I'll add the conversion.

indico/web/client/styles/base/_layout.scss
pferreir: Sorry, my editor was misconfigured at the time 😞

indico/web/templates/base.html
pferreir: The comment was actually quite useful while moving to Webpack, I agree now it doesn't make much sense.

indico/web/views.py
pferreir: Yeah, that's the problem. I initially did it like that and then moved to the `lambda` 😉

indico/web/views.py
pferreir: I don't even need a list. A generator is enough 😉

webpack.config.js
```diff
@@ -0,0 +1,257 @@
...
+                    // some files are part of chunks, others (e.g. images) use file-loader
+                    if (file.chunk) {
+                        hash = file.chunk.renderedHash.slice(0, 8);
+                    } else {
+                        hash = hashFile(path.resolve(config.build.distPath, file.name));
+                    }
+                    file.path = file.path.replace(/dist/, 'dist/v/' + hash);
```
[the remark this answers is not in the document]
pferreir: Yeah, the regex slashes tricked me. 👍

indico/util/mathjax.py
```diff
@@ -16,11 +16,12 @@
...
+
 from indico.legacy.common.TemplateExec import render
 
 
 class MathjaxMixin(object):
     def _getHeadContent(self):
-        return (render('js/mathjax.config.js.tpl') +
-                b'\n'.join(b'<script src="{0}" type="text/javascript"></script>'.format(url)
-                           for url in self._asset_env['mathjax_js'].urls()))
+        webpack = current_app.extensions['flask-webpackext']
```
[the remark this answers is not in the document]
pferreir: Actually, that's not that simple. `flask-webpackext` registers itself in `extensions`, where it stores a (`_FlaskWebpackExtState`)[https://github.com/inveniosoftware/flask-webpackext/blob/master/flask_webpackext/ext.py#L37] object. There is no way to access it from anywhere else AFAIK.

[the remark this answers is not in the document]
ThiefMaster: Maybe a property `app_state` on the extension instance that gets `current_app.extensions['flask-webpackext']` would be good... or even properties proxying to the various app-specific things.

indico/web/client/js/jquery/utils/dropzone.js
```diff
@@ -72,7 +74,8 @@
                     }
                 }
 
-                $form.on('submit', function(e) {
+                // We'll first remove the default 'submit' handler
+                $form.off('submit').on('submit', function(e) {
```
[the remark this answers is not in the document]
pferreir: The `ready(...)` call that sets up the dropzone is registered first, so it should be called first. I agree it's not beautiful, but it's better than the undefined behaviour we had before...

indico/web/flask/util.py
```diff
@@ -321,6 +321,21 @@ def send_file(name, path_or_fd, mimetype, last_modified=None, no_cache=True, inl
     return rv
 
 
+def find_url_endpoint(url):
```
ThiefMaster: heh, i think we used to have a very similar function in indico/web/flask/util.py until some months ago when i removed it as it had become obsolete ;)

indico/core/plugins/__init__.py
```diff
@@ -180,6 +183,10 @@ def translation_domain(self):
         path = self.translation_path
         return Domain(path) if path else NullDomain()
 
+    @property
```
ThiefMaster: `@cached_property`?

webpack/base.js
```diff
@@ -124,7 +124,7 @@ export function webpackDefaults(env, config) {
                                     ctx: {
                                         urlnamespaces: {
                                             namespacePaths: (name) => {
-                                                return `${config.indico.build.staticURL}/plugins/${name}`;
+                                                return `${config.indico.build.staticURL}static/plugins/${name}`;
```
ThiefMaster: does staticURL already have a trailing slash? (pretty ugly IMO, but no idea if that's something webpack always does)

indico/core/plugins/__init__.py
pferreir: The manifest loader actually has its own cache, but yeah, we save ourselves the trouble of concatenating a few strings.

webpack/base.js
pferreir: Yeah, I have to work on trailing slash sanitization 😢

indico/cli/webpack.py
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+
+def _webpack_build_plugin(project, plugin_name, watch=False):
+    plugin = plugin_engine.get_plugin(plugin_name)
+    plugin_path = os.path.dirname(plugin.root_path)
+    webpack_config = os.path.join(plugin_path, 'webpack.config.js')
+
+    with open(os.path.join(plugin_path, 'config.json'), 'w') as f:
+        f.write(json.dumps(_generate_webpack_config(plugin)))
```
ThiefMaster: json.dump(_generate_webpack_config(plugin), f)

webpack.config.js
```diff
@@ -0,0 +1,152 @@
...
+
+// Add Meeting Themes
+entryPoints = Object.assign(entryPoints, base.getThemeEntryPoints(config, './themes/'));
+
+module.exports = env => {
+    const currentEnv = (env ? env.NODE_ENV : null) || 'development';
+
+    // Minification of copied files (e.g. CKEditor and MathJax)
+    const transform = currentEnv === 'development' ? null : (content) => {
+        if (!path.match(/\.js$/)) {
```
ThiefMaster: `path` is the "path" module so this fails during `indico webpack build`

indico/web/views.py
```diff
@@ -242,7 +242,8 @@ def display(self, **params):
         js_files = map(self._fix_path, custom_js)
 
         body = to_unicode(self._display(params))
-        bundles = itertools.chain((current_app.manifest[x] for x in self._resolve_bundles()),
+        bundles = itertools.chain((current_app.manifest[x] for x in self._resolve_bundles()
```
ThiefMaster: if we need it here, we probably need to change this in events/views.py as well

PR #3175, opened 2017-12-07
Reviewing questions

mic4ael:


indico/modules/events/papers/forms.py
```diff
@@ -236,6 +216,10 @@ def split_data(self):
         return {'questions_data': {k: v for k, v in data.iteritems() if k.startswith('question_')},
                 'review_data': {k: v for k, v in data.iteritems() if not k.startswith('question_')}}
 
+    @property
+    def has_questions(self):
+        return len(filter(lambda item: item[0].startswith('question_'), self.data.iteritems())) != 0
```
ThiefMaster: see my comment on the other incarnation of this method

indico/modules/events/reviewing_questions_fields.py
```diff
@@ -0,0 +1,103 @@
...
+from indico.web.forms.widgets import SwitchWidget
+
+
+class BaseReviewingQuestionConfigForm(IndicoForm):
+    text = StringField(_('Question'), [DataRequired()])
+    is_required = BooleanField(_('Required'), widget=SwitchWidget())
+
+    @property
+    def field_data(self):
+        return {'is_required': self.is_required.data}
```
ThiefMaster: shouldn't be in `field_data`

indico/modules/events/reviewing_questions_fields.py
```diff
@@ -0,0 +1,103 @@
...
+        return dict(base, multiline=self.multiline.data, max_words=self.max_words.data,
+                    max_length=self.max_length.data)
+
+
+class TextReviewingQuestion(TextField, BaseField):
+    common_settings = ('text',)
+    config_form_base = TextReviewingQuestionConfigForm
+
+
+def get_reviewing_question_types():
```
ThiefMaster: Why not add an argument for the type (abstracts/papers) instead of returning both?

indico/modules/events/templates/reviews/timeline.html
```diff
@@ -182,9 +182,16 @@
...
-                                                <span class="score">{{ rating.value }}</span>
+                                            {% if rating.question.question_type == 'rating' %}
+                                                <span class="rating-scale flexrow f-a-center">
+                                                    {{ _render_rating(proposal.cfp.rating_range, rating.value) }}
+                                                    <span class="score">{{ rating.value if rating.value is not none else '-' }}</span>
+                                                </span>
+                                            {% elif rating.question.question_type == 'bool' %}
+                                                {{ _('Yes') if rating.value else _('No') }}
+                                            {% else %}
+                                                {{ rating.value|safe }}
```
ThiefMaster: `|safe`? that sounds like a *very* bad idea since it lets you put HTML there...

indico/core/db/sqlalchemy/review_questions.py
```diff
@@ -111,3 +126,15 @@ def get_review_rating(self, review, allow_create=False):
...
+    def title(self):
+        return self.text
+
+    @property
+    def description(self):
+        return None
+
+    @property
+    def is_required(self):
+        return self.field_data['is_required']
```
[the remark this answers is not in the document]
mic4ael: ok, fair enough, I didn't notice that method

indico/modules/events/abstracts/forms.py
```diff
@@ -335,6 +324,10 @@ def split_data(self):
         return {'questions_data': {k: v for k, v in data.iteritems() if k.startswith('question_')},
                 'review_data': {k: v for k, v in data.iteritems() if not k.startswith('question_')}}
 
+    @property
+    def has_questions(self):
+        return len(filter(lambda item: item[0].startswith('question_'), self.data.iteritems())) != 0
```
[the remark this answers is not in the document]
mic4ael: I don't remember why you dislike lambdas so much :P could you please remind me?

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -605,10 +605,12 @@ def get_track_reviewing_state(self, track):
     def get_track_question_scores(self):
         query = (db.session.query(AbstractReview.track_id,
                                   AbstractReviewQuestion,
-                                  db.func.avg(AbstractReviewRating.value))
+                                  db.func.avg(AbstractReviewRating.value.cast(db.String).cast(db.Integer)))
```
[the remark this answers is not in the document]
mic4ael: Cool, good to know!

indico/modules/events/abstracts/forms.py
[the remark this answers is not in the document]
ThiefMaster: https://stackoverflow.com/questions/3013449/list-filtering-list-comprehension-vs-lambda-filter

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -605,10 +605,12 @@ def get_track_reviewing_state(self, track):
...
                  .join(AbstractReviewRating.review)
                  .join(AbstractReviewRating.question)
                  .filter(AbstractReview.abstract == self,
+                         AbstractReviewQuestion.question_type == 'rating',
+                         AbstractReviewRating.value.cast(db.String) != db.text("'null'"),
```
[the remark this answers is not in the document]
mic4ael: The value might be `null` whenever there is a non-required rating question with no value provided by the user.

indico/modules/events/abstracts/templates/management/abstract_reviewing_questions.html
```diff
@@ -0,0 +1,112 @@
...
+{% from '_sortable_list.html' import sortable_list %}
+
+<div class="manage-reviewing-questions">
+    <div class="toolbar space-after hide-if-locked">
+        <a class="i-button arrow icon-plus highlight" data-toggle="dropdown">
+            {%- trans %}Add new question{% endtrans -%}
+        </a>
+        <ul class="dropdown">
+            <li>
+                <a data-ajax-dialog class="js-action-button"
```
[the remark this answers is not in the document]
mic4ael: ouch, that was for sure by mistake

indico/core/db/sqlalchemy/review_questions.py
```diff
@@ -50,6 +50,13 @@ def event_id(cls):
             nullable=False
         )
 
+    @declared_attr
+    def question_type(cls):
+        return db.Column(
+            db.Text,
```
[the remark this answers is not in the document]
ThiefMaster: Not a great idea IMO, since that'd make it much harder if we ever wanted to use a signal to return the field types (like we do in surveys, contrib fields, etc) instead of supporting only a hardcoded list.

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -116,16 +118,10 @@ def _process(self):
                        .join(AbstractReview.abstract)
                        .filter(~Abstract.is_deleted, Abstract.event == self.event)
                        .has_rows())
-        defaults = FormDefaults(abstract_review_questions=self.event.abstract_review_questions,
-                                **abstracts_reviewing_settings.get_all(self.event))
+        defaults = FormDefaults(**abstracts_reviewing_settings.get_all(self.event))
         form = AbstractReviewingSettingsForm(event=self.event, obj=defaults, has_ratings=has_ratings)
```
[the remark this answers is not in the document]
mic4ael: adapted

indico/modules/events/papers/controllers/management.py
```diff
@@ -219,3 +214,88 @@ def _process(self):
...
+class RHReviewingQuestionsActionsBase(RHManagePapersBase):
+    def _process_args(self):
+        RHManagePapersBase._process_args(self)
+        try:
+            self.review_type = request.view_args['review_type']
+        except KeyError:
+            raise NotFound
+
+
+class RHCreateReviewingQuestion(RHReviewingQuestionsActionsBase):
```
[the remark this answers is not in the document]
mic4ael: extracted to functions which are in operations.py module

indico/modules/events/abstracts/models/reviews.py
```diff
@@ -184,7 +184,8 @@ def visibility(self):
 
     @property
     def score(self):
-        ratings = [r for r in self.ratings if not r.question.no_score and not r.question.is_deleted]
+        ratings = [r for r in self.ratings if not r.question.no_score and not r.question.is_deleted and
+                   r.value is not None]
```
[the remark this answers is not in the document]
mic4ael: done

mic4ael:
Updated

indico/core/db/sqlalchemy/review_questions.py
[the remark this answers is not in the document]
pferreir: Ah, you mean having it extended by plugins? Good point.

indico/core/db/sqlalchemy/review_questions.py
```diff
@@ -111,3 +135,20 @@ def get_review_rating(self, review, allow_create=False):
...
             rating = rating_class(question=self, review=review)
         return rating
+
+    @property
+    def title(self):
+        return self.text
+
+    @property
+    def description(self):
+        """Required by BaseField for creating an instance of WTForms field"""
```
ThiefMaster: I think this should be a comment, not a docstring

indico/modules/events/operations.py
```diff
@@ -278,3 +280,35 @@ def unlock_event(event):
     db.session.flush()
     logger.info('Event %r unlocked by %r', event, session.user)
     event.log(EventLogRealm.event, EventLogKind.change, 'Event', 'Event unlocked', session.user)
+
+
+def create_reviewing_question(event, question_model, wtf_field_cls, form, data=None):
+    new_question = question_model()
+    new_question.no_score = True if wtf_field_cls.name != 'rating' else getattr(form, 'no_score', True)
```
ThiefMaster: `wtf_field_cls.name != 'rating' or getattr(form, 'no_score', True)`

Also, if `form` is a wtform, then the else block is broken, since the attribute is a field, not its data.

indico/core/db/sqlalchemy/review_questions.py
```diff
@@ -111,3 +135,20 @@ def get_review_rating(self, review, allow_create=False):
...
+        return None
+
+    @property
+    def no_score(self):
+        return self.field_type != 'rating' or self._no_score
+
```
ThiefMaster: You might want to add a `.expression` for the property; like this you can use it in `filter()` criteria as well.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -605,12 +605,14 @@ def get_track_reviewing_state(self, track):
     def get_track_question_scores(self):
         query = (db.session.query(AbstractReview.track_id,
                                   AbstractReviewQuestion,
-                                  db.func.avg(AbstractReviewRating.value))
+                                  db.func.avg(AbstractReviewRating.value.op('#>>')('{}').cast(db.Integer)))
                  .join(AbstractReviewRating.review)
                  .join(AbstractReviewRating.question)
                  .filter(AbstractReview.abstract == self,
+                         AbstractReviewQuestion.field_type == 'rating',
+                         AbstractReviewRating.value.cast(db.String) != db.text("'null'"),
```
ThiefMaster: I'd still change the field type to JSONB and use `jsonb_typeof` for the check.

indico/migrations/versions/20171124_1138_2af245be72a6_review_questions_models.py
```diff
@@ -0,0 +1,65 @@
...
+    op.execute('ALTER TABLE event_abstracts.abstract_review_ratings ALTER COLUMN "value" TYPE INT '
+               'USING to_json(value)::TEXT::INT')
+    op.drop_column('abstract_review_questions', 'field_type', schema='event_abstracts')
+    op.drop_column('abstract_review_questions', 'is_required', schema='event_abstracts')
+    op.drop_column('abstract_review_questions', 'field_data', schema='event_abstracts')
+    op.execute("DELETE FROM event_paper_reviewing.review_ratings WHERE question_id "
+               "IN(SELECT id FROM event_paper_reviewing.review_questions WHERE field_type != 'rating')")
+    op.execute("DELETE FROM event_paper_reviewing.review_questions WHERE field_type != 'rating'")
+    op.execute('ALTER TABLE event_paper_reviewing.review_ratings ALTER COLUMN "value" TYPE INT '
+               'USING to_json(value)::TEXT::INT')
```
ThiefMaster: Why `to_json()`? value is already JSON here, isn't it?

indico/modules/events/abstracts/blueprint.py
```diff
@@ -71,6 +71,16 @@
                  methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/abstracts/review-settings', 'manage_reviewing_settings', management.RHManageAbstractReviewing,
                  methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/abstracts/question-settings', 'manage_reviewing_questions',
```
ThiefMaster: maybe just `/manage/abstracts/questions/`?

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -159,3 +159,66 @@ def _process(self):
...
+        self.question = AbstractReviewQuestion.get_one(request.view_args['question_id'])
+
+
+class RHEditAbstractReviewingQuestion(RHReviewingQuestionBase):
+    def _process(self):
+        defaults = FormDefaults(obj=self.question, **self.question.field_data)
+        form = self.question.field.create_config_form(obj=defaults)
+        if form.validate_on_submit():
+            edit_reviewing_question(self.question, form)
+            logger.info("Abstract reviewing question %s has been updated by %s", self.question, session.user)
```
ThiefMaster: server-side log messages can be more concise; i'd remove the "has been". Also, `%r` since we use the reprs here. Makes no difference since we don't use `__str__` on our models, but st ill cleaner.

You could also move it into the operation and just log it as `Reviewing question %r ...` so it'll be fine for all types of reviewing. The repr of the question makes it clear enough what it is.

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -605,12 +605,14 @@ def get_track_reviewing_state(self, track):
...
                          ~AbstractReviewQuestion.is_deleted,
-                         ~AbstractReviewQuestion.no_score)
+                         ~AbstractReviewQuestion._no_score)
```
ThiefMaster: if you use the `.expression` property suggested in another comment you can just use `~AbstractReviewQuestion.no_score` here

indico/modules/events/abstracts/operations.py
```diff
@@ -367,13 +367,15 @@ def update_abstract_review(review, review_data, questions_data):
...
         old_value = rating.value
-        rating.value = int(questions_data[field_name])
+        rating.value = questions_data[field_name]
         if old_value != rating.value:
-            changes[field_name] = (old_value, rating.value)
+            changes[field_name] = (question.field.get_friendly_value(old_value),
+                                   question.field.get_friendly_value(rating.value))
             log_fields[field_name] = {
                 'title': question.text,
                 'type': 'number'
```
ThiefMaster: this needs to be set according to the field type as it determines how the diff is displayed in the log message

indico/modules/events/papers/controllers/management.py
```diff
@@ -219,3 +214,86 @@ def _process(self):
...
+
+class RHManageReviewingQuestions(RHManagePapersBase):
+    def _process(self):
+        review_type = request.view_args['review_type']
+        if review_type == 'layout':
+            questions = self.event.cfp.layout_review_questions
+        else:
+            questions = self.event.cfp.content_review_questions
+
+        actions = {'create': 'papers.create_reviewing_question', 'edit': 'papers.edit_reviewing_question',
```
ThiefMaster: I think `endpoints` instead of `actions` would be more suitable.

indico/modules/events/papers/controllers/management.py
```diff
@@ -219,3 +214,86 @@ def _process(self):
...
+                   'delete': 'papers.delete_reviewing_question', 'sort': 'papers.sort_reviewing_questions'}
+        args = {'create': {'review_type': review_type}, 'sort': {'review_type': review_type}}
```
ThiefMaster: I don't think the nested dict is very useful here. I'd just go for `common_url_args = {'review_type': review_type}` and use that when building the URLs that are not for a specific question.

indico/modules/events/papers/controllers/management.py
```diff
@@ -219,3 +214,86 @@ def _process(self):
...
+        try:
+            self.field_cls = get_reviewing_field_types('papers')[request.args['field_type']]
+        except KeyError:
+            raise NotFound
+
+    def _process(self):
+        form = self.field_cls.create_config_form()
+        if form.validate_on_submit():
+            new_question = create_reviewing_question(self.event, PaperReviewQuestion, self.field_cls, form,
+                                                     {'type': PaperReviewType.get(self.review_type)})
```
ThiefMaster: `PaperReviewType[self.review_type]` since you don't want `None` in case of an invalid type...

indico/modules/events/papers/controllers/management.py
```diff
@@ -219,3 +214,86 @@ def _process(self):
...
+                         .filter_by(id=request.view_args['question_id'])
+                         .one())
+
+
+class RHEditReviewingQuestion(RHManageQuestionBase):
+    def _process(self):
+        defaults = FormDefaults(obj=self.question, **self.question.field_data)
+        form = self.question.field.create_config_form(obj=defaults)
+        if form.validate_on_submit():
+            edit_reviewing_question(self.question, form)
```
ThiefMaster: `update_reviewing_question` would be a better name. editing is usually what the user does, not the process of actually updating something

indico/modules/events/papers/operations.py
```diff
@@ -297,9 +297,10 @@ def update_review(review, review_data, questions_data):
...
         old_value = rating.value
-        rating.value = int(questions_data[field_name])
+        rating.value = questions_data[field_name]
         if old_value != rating.value:
-            changes[field_name] = (old_value, rating.value)
+            changes[field_name] = (question.field.get_friendly_value(old_value),
+                                   question.field.get_friendly_value(rating.value))
             log_fields[field_name] = {
                 'title': question.text,
                 'type': 'number'
```
ThiefMaster: needs to depend on the question type

indico/modules/events/reviewing_questions_fields.py
```diff
@@ -0,0 +1,102 @@
...
+    wtf_field_class = RatingReviewField
+
+    @property
+    def wtf_field_kwargs(self):
+        range_ = self.object.event.cfp.rating_range
+        choices = [(n, unicode(n)) for n in range(range_[0], range_[1] + 1)]
+        return {'coerce': int, 'choices': choices, 'rating_range': range_, 'question': self.object}
+
+
+class BoolReviewingQuestionConfigForm(BaseReviewingQuestionConfigForm):
```
ThiefMaster: Why not just use the base one? this subclass seems to be a bit pointless

indico/modules/events/reviewing_questions_fields.py
```diff
@@ -0,0 +1,102 @@
...
+from indico.modules.events.fields import RatingReviewField
+from indico.util.i18n import _
+from indico.web.fields.base import BaseField, IndicoForm
+from indico.web.fields.simple import BoolField, TextField
+from indico.web.forms.widgets import SwitchWidget
+
+
+class BaseReviewingQuestionConfigForm(IndicoForm):
+    text = StringField(_('Question'), [DataRequired()])
+    is_required = BooleanField(_('Required'), widget=SwitchWidget())
```
ThiefMaster: can we avoid replicating all the config forms here and just use the ones of the base fields? i think you are only replicating them to get rid of `description` but keeping a description wouldn't be so bad IMO! (on the review page we could show it as an info tooltip for example)

indico/modules/events/templates/reviewing_questions_management.html
```diff
@@ -0,0 +1,105 @@
...
+<div class="manage-reviewing-questions">
+    <div class="toolbar space-after hide-if-locked">
+        <a class="i-button arrow icon-plus highlight" data-toggle="dropdown">
+            {%- trans %}Add new question{% endtrans -%}
+        </a>
+        <ul class="dropdown">
+            {% for name, field_type in field_types.iteritems() -%}
+                <li>
+                    <a class="js-action-button"
+                       data-href="{{ url_for(actions.create, event, field_type=name, **(args.create|default({}))) }}"
```
ThiefMaster: if you get rid of the nested dict as suggested above, you can just use `**common_url_args` instead of the `default({})` stuff

indico/modules/events/templates/reviewing_questions_management.html
```diff
@@ -0,0 +1,105 @@
...
+                       data-title="{% trans %}Add custom reviewing question{% endtrans %}"
+                       data-ajax-dialog>
+                        {% trans name=field_type.friendly_name %}
```
ThiefMaster: what's the point of this trans block?

indico/migrations/versions/20171124_1138_2af245be72a6_review_questions_models.py
mic4ael: Not sure why.. my overlooking

indico/modules/events/abstracts/models/abstracts.py
mic4ael: why jsonb?

ThiefMaster: Because there is no `json_typeof`. `JSON` is stored as a string, while `JSONB` is stored in a binary format that allows postgres to actually look at its contents.

mic4ael: are you sure there is no `json_typeof`? https://www.postgresql.org/docs/9.6/static/functions-json.html

ThiefMaster: Indeed, looks like it's only the operators that are JSONB-only. Anyway, the performance is probably better with JSONB

mic4ael: Yeah, but do we need that performance in a field which holds only one simple value?

ThiefMaster: No, it probably doesn't make a noticeable difference here.

indico/modules/events/templates/reviewing_questions_management.html
mic4ael: Ah, indeed, the `friendly_name` is already translated here.

ThiefMaster: Even if it was, this wouldn't help, since the string would not be extracted for translation (it might just work by coincidence if it was extracted somewhere else)

indico/modules/events/reviewing_questions_fields.py
mic4ael: yeah, I just wanted to have separate classes for each field

indico/modules/events/reviewing_questions_fields.py
mic4ael: This `title = StringField(_('Title'), [DataRequired()], description=_("The title of the field"))` from `FieldConfigForm` makes a little bit of a problem. Since in the context of reviewing fields  `Title` for the field label is wrong.

mic4ael:
Updated

indico/modules/events/reviewing_questions_fields.py
ThiefMaster: "[Question] Title" is not so bad IMO, and you could modify the description after instantiating the form

indico/modules/events/abstracts/operations.py
```diff
@@ -367,13 +367,15 @@ def update_abstract_review(review, review_data, questions_data):
...
-        rating.value = int(questions_data[field_name])
+        rating.value = questions_data[field_name]
         if old_value != rating.value:
-            changes[field_name] = (old_value, rating.value)
+            changes[field_name] = (question.field.get_friendly_value(old_value),
+                                   question.field.get_friendly_value(rating.value))
             log_fields[field_name] = {
                 'title': question.text,
-                'type': 'number'
+                'type': question.field_type
```
ThiefMaster: I don't think we want `'type': 'rating'` here, since it's just a number in the end.

indico/modules/events/logs/util.py
```diff
@@ -89,7 +89,11 @@ def render_changes(a, b, type_):
     :param b: new value
     :param type_: the type determining how the values should be compared
     """
-    if type_ in ('number', 'enum', 'bool', 'datetime'):
+    if type_ in ('number', 'enum', 'bool', 'datetime', 'rating'):
+        if a is None or a == '':
```
ThiefMaster: `if a in (None, '')`

indico/modules/events/logs/util.py
```diff
@@ -89,7 +89,11 @@ def render_changes(a, b, type_):
     :param b: new value
     :param type_: the type determining how the values should be compared
     """
-    if type_ in ('number', 'enum', 'bool', 'datetime'):
+    if type_ in ('number', 'enum', 'bool', 'datetime', 'rating'):
```
ThiefMaster: see my comment above, i would not introduce a 'rating' type here

indico/modules/events/operations.py
```diff
@@ -278,3 +280,41 @@ def unlock_event(event):
     db.session.flush()
     logger.info('Event %r unlocked by %r', event, session.user)
     event.log(EventLogRealm.event, EventLogKind.change, 'Event', 'Event unlocked', session.user)
+
+
+def create_reviewing_question(event, question_model, wtf_field_cls, form, data=None):
+    new_question = question_model()
+    new_question.no_score = True
+    if 'no_score' in form.data:
+        new_question.no_score = form.data['no_score']
```
ThiefMaster: `if form.no_score:` and (in the next line) `form.no_score.data`. no need to build the data dict thrice

indico/modules/events/papers/operations.py
```diff
@@ -297,12 +297,13 @@ def update_review(review, review_data, questions_data):
...
-        rating.value = int(questions_data[field_name])
+        rating.value = questions_data[field_name]
         if old_value != rating.value:
-            changes[field_name] = (old_value, rating.value)
+            changes[field_name] = (question.field.get_friendly_value(old_value),
+                                   question.field.get_friendly_value(rating.value))
             log_fields[field_name] = {
                 'title': question.text,
-                'type': 'number'
+                'type': question.field_type
```
ThiefMaster: see my other comment

indico/modules/events/templates/reviewing_questions_management.html
```diff
@@ -0,0 +1,103 @@
...
+        });
+
+        var $questionsList = $('#reviewing-questions-list');
+        setupSortableList($questionsList);
+        $questionsList.on('sortupdate', function(evt, ui) {
+            var ids = $questionsList.find('li').map(function(index, item) {
+                return $(item).data('item-id');
+            });
+
+            ids = $.makeArray(ids);
```
ThiefMaster: If this is just converting the jquery object `ids` to a plain array, you can just call `.get()` after `.map()`

indico/modules/events/templates/reviews/rating_widget.html
```diff
@@ -1,27 +1,14 @@
...
-                       {{ 'checked' if field.data == bullet|string }}>
-                <label class="bullet" for="{{ radio_id }}" data-value="{{ bullet }}"></label>
-            {% endfor %}
-        </div>
+    <div id="{{ field.id }}" class="rating-scale flexrow f-a-start">
+        <span class="score">{{ '-' if field.object_data is none else field.data }}</span>
+        {% for bullet in range(rating_range[1], rating_range[0] - 1, -1) %}
+            {% set radio_id = '%s-%d'|format(field.id, bullet) %}
+            <input type="radio" id="{{ radio_id }}" name="{{ field.name }}" value="{{ bullet }}"
+                    {{ 'checked' if field.data == bullet }}>
```
ThiefMaster: 1 space too much indentation

indico/modules/events/operations.py
mic4ael: `form.no_score` will raise an `AttributeError` if there is no such a property

mic4ael: `if hasattr(form, 'no_score'):` will be a better option

ThiefMaster: You can use `'no_score' in form`

indico/migrations/versions/20171124_1138_2af245be72a6_review_questions_models.py
```diff
@@ -19,6 +19,7 @@
 def upgrade():
     for schema, a, b in (('event_abstracts', 'abstract_review_ratings', 'abstract_review_questions'),
```
ThiefMaster: `a` and `b`, really? ;)

how about some meaningful names. You can always define the list in a separate statement above to keep it readable.

indico/web/templates/forms/_form.html
```diff
@@ -41,7 +41,7 @@
     </div>
 {%- endmacro %}
 
-{% macro form_field(field, field_classes, widget_attrs={}, disabled=none) -%}
+{% macro form_field(field, field_classes, hide_description, widget_attrs={}, disabled=none) -%}
```
ThiefMaster: this should default to `false`

mic4ael:
Updated

PR #3176, opened 2017-12-07
Add header-only class when needed

mic4ael:
Sometime the ratings-details box may look broken, e.g. when

1) User clicks `show ratings`
2) User clicks icon `Edit review`
3) User clicks `Cancel`

untitled

### DIFF
--- a/indico/htdocs/js/indico/modules/reviews.js
+++ b/indico/htdocs/js/indico/modules/reviews.js
@@ -53,7 +53,10 @@
                 $item.addClass('header-indicator-left');
             }
             if ($item.data('no-comment') !== undefined) {
-                $item.addClass('header-only');
+                var $ratingsDetails = $item.find('.ratings-details');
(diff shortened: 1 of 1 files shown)


indico/htdocs/js/indico/modules/reviews.js
```diff
@@ -53,7 +53,10 @@
                 $item.addClass('header-indicator-left');
             }
             if ($item.data('no-comment') !== undefined) {
-                $item.addClass('header-only');
+                var $ratingsDetails = $item.find('.ratings-details');
+                if (!$ratingsDetails.length || $ratingsDetails.css('display') === 'none') {
```
ThiefMaster: `!$ratingsDetails.is(':visible')` doesn't work for the second check?

mic4ael: works, works ;)

PR #3177, opened 2017-12-07
[RFC] Make image upload better configurable

bpedersen2:
Add configs for image types and try to handle pdf as well.

This allows to upload 'hidden' files (e.g. pdf files) and link them
from pages or the menu, without having these files show up in the
materials listing on the event front page.

### DIFF
--- a/indico/modules/events/layout/controllers/images.py
+++ b/indico/modules/events/layout/controllers/images.py
@@ -18,11 +18,14 @@ from __future__ import unicode_literals
 
 import shutil
 from io import BytesIO
+from mimetypes import guess_type
 
 from flask import flash, render_template, request, session
 from PIL import Image
+from pyPdf import PdfFileReader
 
 from indico.core import signals
+from indico.core.config import config
 from indico.core.db import db
 from indico.modules.events.controllers.base import RHDisplayEventBase
 from indico.modules.events.layout import logger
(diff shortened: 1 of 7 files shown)


ThiefMaster:
RFC in the title == you want feedback? If yes, that shouldn't be in the commit message as well ;)

bpedersen2:
The alternative would be  to allow  'hidden' folders for material, that do only show up if you have edit access, but  still allow read access to the contained material.

indico/modules/events/layout/controllers/images.py
```diff
@@ -56,24 +59,36 @@ def _process(self):
         files = request.files.getlist('image')
         num = 0
         for f in files:
+
             filename = secure_filename(f.filename, 'image')
+            mtype = guess_type(filename)[0]
```
ThiefMaster: This uses the extension and is not very reliable. See e.g. `RHLayoutLogoUpload` on how to do a better check for image types.

indico/modules/events/layout/controllers/images.py
```diff
@@ -18,11 +18,14 @@
 
 import shutil
 from io import BytesIO
+from mimetypes import guess_type
 
 from flask import flash, render_template, request, session
 from PIL import Image
+from pyPdf import PdfFileReader
```
ThiefMaster: this adds a new dependency, with the last release being 7 years old. I don't think we want this in indico.

indico/modules/events/layout/controllers/images.py
```diff
@@ -56,24 +59,36 @@ def _process(self):
...
-                image_type = Image.open(data).format.lower()
-            except IOError:
-                # Invalid image data
-                continue
+            if mtype == 'application/pdf':
+                try:
+                    _pdf = PdfFileReader(data)
+                    image_type = 'pdf'
+                    content_type = mtype
+                except Exception:
```
ThiefMaster: way too broad

indico/modules/events/layout/controllers/images.py
```diff
@@ -56,24 +59,36 @@ def _process(self):
...
+                    continue
+            else:
+                try:
+                    image_type = Image.open(data).format.lower()
```
ThiefMaster: ah, didn't see this when posting the comment above regarding getting the image type properly

indico/modules/events/layout/templates/image_list.html
```diff
@@ -2,9 +2,13 @@
     <ul class="image-list">
     {% for image in images %}
         <li>
+            {%- if image.content_type.startswith('image/') -%}
             <a href="{{ url_for('event_images.image_display', image) }}" class="js-lightbox">
```
ThiefMaster: missing indentation

indico/modules/events/layout/templates/image_list.html
```diff
@@ -2,9 +2,13 @@
...
                 <img src="{{ url_for('event_images.image_display', image) }}" alt="{% trans %}Image{% endtrans %}">
             </a>
+            {%- else -%}
+            <a href="{{ url_for('event_images.image_display', image) }}"><span class="nonimage">{{ image.filename }}</span></a>
```
ThiefMaster: missing indentation

ThiefMaster:
I think being able to mark a material folder (probably only on the event (and maybe category) level) as hidden would be a better option. I probably wouldn't even do access checks there but completely hide them in display areas and only show them in management...

bpedersen2:
Yeah. I'll  try to find a solution for the hidden-folder approach.

Issue #3180, opened 2017-12-11
Support hidden folders in the frontend


bpedersen2:
Sometime not all material needs/should be listed on the event front page. 
Solution: Allow to  force-hide folders in the front-end.

See also discussion on pull request #3177

ThiefMaster:
Let's keep the discussion in one place (your PR)

PR #3181, opened 2017-12-11
Add support for hidden attachment folders

bpedersen2:
Not all uploaded files need to be shown in the materials listing
on the event front page, sometimes it should only be used for links
at some place in the event. Allow to always hide folders in the
frontend indepently from the access rights.

Fixes: #3180

### DIFF
--- a/indico/modules/attachments/forms.py
+++ b/indico/modules/attachments/forms.py
@@ -20,7 +20,7 @@ from wtforms.ext.sqlalchemy.fields import QuerySelectField
 from wtforms.fields import BooleanField, TextAreaField
 from wtforms.fields.html5 import URLField
 from wtforms.fields.simple import HiddenField, StringField
-from wtforms.validators import DataRequired, Optional
+from wtforms.validators import DataRequired, Optional, ValidationError
 
 from indico.core.db import db
 from indico.core.db.sqlalchemy.protection import ProtectionMode
@@ -101,10 +101,18 @@ class AttachmentFolderForm(IndicoForm):
     acl = AccessControlListField(_("Access control list"), [UsedIf(lambda form, field: form.protected.data)],
                                  groups=True, allow_external=True, default_text=_('Restrict access to this folder'),
                                  description=_("The list of users and groups allowed to access the folder"))
-    is_always_visible = BooleanField(_("Always Visible"), widget=SwitchWidget(),
+    is_always_visible = BooleanField(_("Always Visible"),
+                                     [HiddenUnless('is_hidden', value=False)],
+                                     widget=SwitchWidget(),
                                      description=_("By default, folders are always visible, even if a user cannot "
                                                    "access them. You can disable this behavior here, hiding the folder "
                                                    "for anyone who does not have permission to access it."))
+    is_hidden = BooleanField(_("Always Hidden in Event page"),
+                             [HiddenUnless('is_always_visible', value=False), ],
+                             widget=SwitchWidget(),
+                             description=_("Always hide the folder in the event page materials listing. "
+                                           "You can use this for folders to store non-image files used "
+                                           "in e.g. in download links. The access permissions still apply."))
 
     def __init__(self, *args, **kwargs):
         self.linked_object = kwargs.pop('linked_object')
@@ -119,6 +127,12 @@ class AttachmentFolderForm(IndicoForm):
         if self.title.data:
             suggestions.add(self.title.data)
         return sorted(suggestions)
+
+    def validate_is_always_visible(self, field):
+        if self.is_always_visible.data and self.is_hidden.data:
+            raise ValidationError('These two options cannot be used at the same time')
+
+    validate_is_hidden = validate_is_always_visible
 
     @generated_data
     def protection_mode(self):
--- a/indico/modules/attachments/forms.py
+++ b/indico/modules/attachments/forms.py
@@ -107,12 +107,12 @@ class AttachmentFolderForm(IndicoForm):
                                      description=_("By default, folders are always visible, even if a user cannot "
                                                    "access them. You can disable this behavior here, hiding the folder "
                                                    "for anyone who does not have permission to access it."))
-    is_hidden = BooleanField(_("Always Hidden in Event page"),
-                             [HiddenUnless('is_always_visible', value=False), ],
+    is_hidden = BooleanField(_("Always hidden"),
+                             [HiddenUnless('is_always_visible', value=False)],
                              widget=SwitchWidget(),
-                             description=_("Always hide the folder in the event page materials listing. "
-                                           "You can use this for folders to store non-image files used "
-                                           "in e.g. in download links. The access permissions still apply."))
+                             description=_("Always hide the folder and its contents from public display areas of "
+                                           "the event. You can use this for folders to store non-image files used "
+                                           "e.g. in download links. The access permissions still apply."))
 
     def __init__(self, *args, **kwargs):
         self.linked_object = kwargs.pop('linked_object')
--- a/indico/modules/attachments/models/folders.py
+++ b/indico/modules/attachments/models/folders.py
@@ -50,6 +50,7 @@ class AttachmentFolder(LinkMixin, ProtectionMixin, db.Model):
         return (db.CheckConstraint(default_inheriting, 'default_inheriting'),
                 db.CheckConstraint('is_default = (title IS NULL)', 'default_or_title'),
                 db.CheckConstraint('not (is_default and is_deleted)', 'default_not_deleted'),
+                db.CheckConstraint('not (is_hidden and is_always_visible)', 'is_hidden_not_is_always_visible'),
                 {'schema': 'attachments'})
 
     @declared_attr
@@ -89,6 +90,12 @@ class AttachmentFolder(LinkMixin, ProtectionMixin, db.Model):
         db.Boolean,
         nullable=False,
         default=True
+    )
+    #: If the folder is never shown in the frontend (even if you can access it)
+    is_hidden = db.Column(
+        db.Boolean,
+        nullable=False,
+        default=False
     )
 
     acl_entries = db.relationship(
@@ -159,6 +166,8 @@ class AttachmentFolder(LinkMixin, ProtectionMixin, db.Model):
         This does not mean the user can actually access its contents.
         It just determines if it is visible to him or not.
         """
+        if self.is_hidden:
+            return False
         if not self.object.can_access(user):
             return False
         return self.is_always_visible or super(AttachmentFolder, self).can_access(user)
@@ -213,6 +222,7 @@ class AttachmentFolder(LinkMixin, ProtectionMixin, db.Model):
             self.title,
             ', is_default=True' if self.is_default else '',
             ', is_always_visible=False' if not self.is_always_visible else '',
+            ', is_hidden=True' if self.is_hidden else '',
             ', is_deleted=True' if self.is_deleted else '',
             self.protection_repr,
             self.link_repr
--- a/indico/modules/attachments/templates/_attachments.html
+++ b/indico/modules/attachments/templates/_attachments.html
@@ -64,6 +64,14 @@
                    title="{% trans %}The access to this folder is restricted, regardless of the protection scheme of
                           the parent.{% endtrans %}"></i>
             {% endif %}
+            {% if folder.is_hidden and management %}
+                <i class="icon-eye-blocked"
+                   title="{% trans %}This folder is always hidden in the frontend.{% endtrans %}"></i>
+            {% endif %}
+            {% if folder.is_always_visible and management %}
+                <i class="icon-eye"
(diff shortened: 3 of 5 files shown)


indico/migrations/versions/20171211_0811_03dca79b401c_hidden_folders_support.py
```diff
@@ -0,0 +1,30 @@
...
+down_revision = '566d5de4e0e5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic  ###
+    op.add_column('folders', sa.Column('is_always_hidden_in_fe', sa.Boolean(), nullable=False, server_default='false'),
+                  schema='attachments')
+    op.alter_column('folders', 'is_always_hidden_in_fe', server_default=None, schema='attachments')
```
bpedersen2: Question: Is there something to be done to force the mutual exclusive setting of  hide_in_fe and always_visible at the database level?

indico/migrations/versions/20171211_0811_03dca79b401c_hidden_folders_support.py
```diff
@@ -0,0 +1,30 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '03dca79b401c'
+down_revision = '566d5de4e0e5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic  ###
```
mic4ael: You can remove  the comments from `upgrade` and `downgrade` functions.

bpedersen2: Done

mic4ael:
On a side note you don't have to create a separate commit for each fix/change which is a result of a comment. You could easily use either `git commit --amend` or interactive rebase along with squash operation.

indico/modules/attachments/models/folders.py
```diff
@@ -91,6 +91,14 @@ def __table_args__(cls):
         default=True
     )
 
+    #: If the folder is never shown in the frontend (even if you can access it)
+    is_always_hidden_in_fe = db.Column(
```
ThiefMaster: I kind of dislike this name. It's not very clear without reading the comment, and also too verbose. IMO `is_hidden` is a much better name for it. (also, we don't put an empty line between column definitions, and only one empty line between columns and relationships)

indico/modules/attachments/util.py
```diff
@@ -36,7 +36,7 @@ def get_attached_folders(linked_object, include_empty=True, include_hidden=True,
     folders = AttachmentFolder.get_for_linked_object(linked_object, preload_event=preload_event)
 
     if not include_hidden:
-        folders = [f for f in folders if f.can_view(session.user)]
+        folders = [f for f in folders if (f.can_view(session.user) and not f.is_always_hidden_in_fe)]
```
ThiefMaster: This check can be done in `can_view`

indico/modules/attachments/forms.py
```diff
@@ -101,10 +101,18 @@ class AttachmentFolderForm(IndicoForm):
...
                                  description=_("The list of users and groups allowed to access the folder"))
-    is_always_visible = BooleanField(_("Always Visible"), widget=SwitchWidget(),
+    is_always_visible = BooleanField(_("Always Visible"),
+                                     [HiddenUnless('is_always_hidden_in_fe', value=False), ],
+                                     widget=SwitchWidget(),
                                      description=_("By default, folders are always visible, even if a user cannot "
                                                    "access them. You can disable this behavior here, hiding the folder "
                                                    "for anyone who does not have permission to access it."))
+    is_always_hidden_in_fe = BooleanField(_("Always Hidden in Frontend"),
+                                          [HiddenUnless('is_always_visible', value=False), ],
```
ThiefMaster: no need for the `, ` at the end of the list - only single-element tuples need this.

indico/modules/attachments/util.py
bpedersen2: Done

indico/modules/attachments/models/folders.py
bpedersen2: Changed to is_hidden_in_fe .

indico/modules/attachments/forms.py
bpedersen2: Done

indico/modules/attachments/models/folders.py
ThiefMaster: I think the `_in_fe` is not very useful either; that's something worth mentioning in the docstring on top of it (like it already does)

bpedersen2: Ok, using is_hidden now

indico/modules/attachments/forms.py
```diff
@@ -101,10 +101,18 @@ class AttachmentFolderForm(IndicoForm):
...
                                  groups=True, allow_external=True, default_text=_('Restrict access to this folder'),
                                  description=_("The list of users and groups allowed to access the folder"))
-    is_always_visible = BooleanField(_("Always Visible"), widget=SwitchWidget(),
+    is_always_visible = BooleanField(_("Always Visible"),
+                                     [HiddenUnless('is_hidden', value=False)],
+                                     widget=SwitchWidget(),
                                      description=_("By default, folders are always visible, even if a user cannot "
                                                    "access them. You can disable this behavior here, hiding the folder "
                                                    "for anyone who does not have permission to access it."))
+    is_hidden = BooleanField(_("Always Hidden in Frontend"),
```
pferreir: Hi,
I'm not that sure about the wording here. "Frontend" is not a term that a end user will necessarily understand.

pferreir: I discussed it briefly with @ThiefMaster, "Hide in event page" or similar could be an option.

indico/modules/attachments/templates/_attachments.html
```diff
@@ -64,6 +64,14 @@
                    title="{% trans %}The access to this folder is restricted, regardless of the protection scheme of
                           the parent.{% endtrans %}"></i>
             {% endif %}
+            {% if folder.is_hidden and management%}
```
ThiefMaster: missing space before `%}`

indico/modules/attachments/templates/_attachments.html
```diff
@@ -64,6 +64,14 @@
...
+                <i class=" icon-eye-blocked"
+                title="{% trans %}This folder is always hidden in the frontend.{% endtrans %}"></i>
```
ThiefMaster: not enough indentation (attrs should be aligned)

indico/modules/attachments/templates/_attachments.html
```diff
@@ -64,6 +64,14 @@
                    title="{% trans %}The access to this folder is restricted, regardless of the protection scheme of
                           the parent.{% endtrans %}"></i>
             {% endif %}
+            {% if folder.is_hidden and management%}
+                <i class=" icon-eye-blocked"
```
ThiefMaster: there shouldn't be a space right after the opening quote

indico/modules/attachments/templates/_attachments.html
```diff
@@ -64,6 +64,14 @@
...
+                title="{% trans %}This folder is always hidden in the frontend.{% endtrans %}"></i>
+            {% endif %}
+            {% if folder.is_always_visible and management%}
+                <i class=" icon-eye"
+                title="{% trans %}This folder is always visible in the frontend.{% endtrans %}"></i>
```
ThiefMaster: Maybe this is a bit noisy since it's enabled by default? Explicitly showing the hidden flag makes sense but for this one I'm not sure if there's any benefit

indico/modules/attachments/models/folders.py
```diff
@@ -90,6 +90,13 @@ def __table_args__(cls):
         nullable=False,
         default=True
     )
+    #: If the folder is never shown in the frontend (even if you can access it)
+    is_hidden = db.Column(
+        db.Boolean,
+        nullable=False,
+        default=False
+    )
+
```
ThiefMaster: one linebreak too much (there should be one empty line between this and the the relationships)

indico/migrations/versions/20171211_0811_03dca79b401c_hidden_folders_support.py
```diff
@@ -0,0 +1,29 @@
...
+down_revision = '566d5de4e0e5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('folders', sa.Column('is_hidden', sa.Boolean(), nullable=False, server_default='false'),
+                  schema='attachments')
+    op.alter_column('folders', 'is_hidden', server_default=None, schema='attachments')
+    op.create_check_constraint('ck_folders_exclusive_is_hidden_is_always_visible', 'folders',
```
ThiefMaster: Not sure if a DB constraint is really important here - there is no harm if both are set for some reason.

Anyway, if you want to keep it, make sure to add it to the model as well!

indico/modules/attachments/templates/_attachments.html
bpedersen2: Well, it is only shown in managment views. And for symmetry ( all other flags one can set on a folder are indicated) I think it is helpful.

indico/migrations/versions/20171211_0811_03dca79b401c_hidden_folders_support.py
bpedersen2: Added constraint to model as well. I think this is needed, as both switches would get hidden if both flags are set in the db. Then the only way to recover would be to edit  values in the database by hand.

indico/modules/attachments/templates/_attachments.html
bpedersen2: Done

indico/modules/attachments/templates/_attachments.html
bpedersen2: Done

indico/modules/attachments/models/folders.py
bpedersen2: Done

indico/modules/attachments/forms.py
bpedersen2: Done

indico/migrations/versions/20171211_0811_03dca79b401c_hidden_folders_support.py
ThiefMaster: Good point indeed!

ThiefMaster:
```python
def validate_is_always_visible(self, field):
    if self.is_always_visible.data and self.is_hidden.data:
        raise ValidationError('These two options cannot be used at the same time')

validate_is_hidden = validate_is_always_visible
```

indico/modules/attachments/models/folders.py
```diff
@@ -50,7 +50,7 @@ def __auto_table_args():
         return (db.CheckConstraint(default_inheriting, 'default_inheriting'),
                 db.CheckConstraint('is_default = (title IS NULL)', 'default_or_title'),
                 db.CheckConstraint('not (is_default and is_deleted)', 'default_not_deleted'),
-                {'schema': 'attachments'})
```
ThiefMaster: what happened to this?!

indico/modules/attachments/models/folders.py
```diff
@@ -161,6 +167,8 @@ def can_view(self, user):
         """
         if not self.object.can_access(user):
             return False
+        if self.is_hidden:
```
ThiefMaster: do this before the access check; it's much cheaper!

indico/modules/attachments/models/folders.py
bpedersen2: Lost due to mis-reading braces..., readded

indico/modules/attachments/models/folders.py
bpedersen2: done

indico/migrations/versions/20171211_0811_03dca79b401c_hidden_folders_support.py
bpedersen2: Done

bpedersen2:
Bumped migrations

indico/migrations/versions/20180129_0842_813ea74ce8dc_hidden_folders_support.py
```diff
@@ -0,0 +1,38 @@
...
+
+def upgrade():
+    op.add_column('folders', sa.Column('is_hidden', sa.Boolean(), nullable=False, server_default='false'),
+                  schema='attachments')
+    op.alter_column('folders', 'is_hidden', server_default=None, schema='attachments')
+    op.create_check_constraint('ck_folders_is_hidden_not_is_always_visible', 'folders',
+                               'NOT (is_hidden AND is_always_visible)', schema='attachments')
+
+
+def validate_is_always_visible(self, field):
```
ThiefMaster: this doesn't belong in here

PR #3185, opened 2017-12-12
Add has_ended property to Event model

mic4ael:


ThiefMaster:
Assuming this is for the outlook plugin: Does it really make much sense to add a property for this? I'd just check this directly in the plugin...

PR #3187, opened 2017-12-14
Fix flower login if used behind a proxy via socket

bpedersen2:
The login redirect uri was always using the port, even if flower is served
via a proxy server, e.g. nginx.

Run flower as:

    indico celery flower --address=<yourhost>  --url-prefix=flower --unix-socket=/var/run/flower/flower.socket 

example nginx snippet:

```
server { ....
  location /flower {
    rewrite ^/flower/(.*)$ /$1 break;
    proxy_pass http://unix:/var/run/flower/flower.socket;
    proxy_set_header Host $host;
    proxy_redirect off;
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
    allow 127.0.0.1;
    deny all;
  }
}
```

### DIFF
--- a/indico/core/celery/flower.py
+++ b/indico/core/celery/flower.py
@@ -67,11 +67,19 @@ class FlowerAuthHandler(BaseHandler, OAuth2Mixin):
     def get_auth_http_client(self):
         return AsyncHTTPClient()
 
+    @property
+    def uri_base(self):
+        try:
+            return os.environ['INDICO_FLOWER_URL']
+        except KeyError:
+            return 'http{}://{}{}{}'.format('s' if 'ssl_options' in settings else '',
+                                            options.address or 'localhost',
+                                            ':{}'.format(options.port) if not options.unix_socket else '',
+                                            '/{}'.format(options.url_prefix) if options.url_prefix else '')
+
     @asynchronous
     def get(self):
-        redirect_uri = 'http{}://{}:{}/login'.format('s' if 'ssl_options' in settings else '',
-                                                     options.address or 'localhost',
-                                                     options.port)
+        redirect_uri = '{}/login'.format(self.uri_base.rstrip('/'))
         if self.get_argument('code', False):
             self.get_authenticated_user(
                 redirect_uri=redirect_uri,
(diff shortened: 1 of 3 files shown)


indico/core/celery/flower.py
```diff
@@ -69,9 +69,10 @@ def get_auth_http_client(self):
 
     @asynchronous
     def get(self):
-        redirect_uri = 'http{}://{}:{}/login'.format('s' if 'ssl_options' in settings else '',
-                                                     options.address or 'localhost',
-                                                     options.port)
+        redirect_uri = 'http{}://{}{}{}/login'.format('s' if 'ssl_options' in settings else '',
+                                                      options.address or 'localhost',
+                                                      ':' + options.port if not options.unix_socket else '',
```
ThiefMaster: The port should only be added if it's not the default (80/443 depending on the protocol)

bpedersen2: The patch removes adding the port (which is 5555 by default for flower) if flower is actually listening on a unix socket. What's not handled currently is an upstream server on non-default ports in this case.  As the options are handled by flower this is currently out of scope.

ThiefMaster: Oh right we were already always adding the port before. Nevermind then.

ThiefMaster:
While this is nice, there is still one problem: If you use SSL, you'd need to use flower's unix socket with SSL too (which doesn't make much sense) because otherwise the `redirect_uri` will use http.

Maybe there is some way to get the current URL that was used to access the page? This would always work, regardless of what type of socket or protocol is used...

`redirect_uri = self.request.full_url()` seems to be a good starting point, but the URL built like this lacks the URL prefix - tbh, using flower on anything but the url root (e.g. on a subdomain) feels a bit hacky. Searching for `url_prefix` in the list of flower  issues also yields lots of results of various issues with it.

bpedersen2:
Yes the http/https problem is not fully solved ( In my install it works, as  there is a redirect from http to https configured in nginx as well.  I think the probably the easiest solution would be adding further settings in indico to override the computed redirect uri ( then passed as enviroment vars in the flower execution)

ThiefMaster:
Doesn't even need a new setting - there is already `FLOWER_URL` which is currently only used to show a link in the admin area but could be used to build the `redirect_uri` as well

PR #3189, opened 2017-12-14
Session types

meluru:


indico/htdocs/sass/modules/_sessions.scss
```diff
@@ -20,3 +20,26 @@
...
+            white-space: normal;
+
+            &.name-column {
+                white-space: nowrap;
+            }
+        }
+
+        .action-column {
+            white-space: nowrap;
+            width: 1px;
```
mic4ael: 1px? why?

indico/migrations/versions/20171214_1059_9c4418d7a6aa_add_session_types_table.py
```diff
@@ -0,0 +1,34 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'session_types',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('event_id', sa.Integer(), nullable=False),
+        sa.Column('name', sa.String(), nullable=False),
+        sa.Column('is_poster', sa.Boolean(), nullable=False),
```
mic4ael: what about `server_default` since it is not nullable?

indico/modules/events/sessions/blueprint.py
```diff
@@ -52,6 +54,15 @@
 _bp.add_url_rule('/manage/sessions/<int:session_id>/acl', 'acl', RHSessionACL)
 _bp.add_url_rule('/manage/sessions/<int:session_id>/acl-message', 'acl_message', RHSessionACLMessage)
 
+# Session types
+_bp.add_url_rule('/manage/sessions/types/', 'manage_types', RHManageSessionTypes)
+_bp.add_url_rule('/manage/sessions/types/create', 'create_type', RHCreateSessionType,
```
mic4ael: what maxlinelength are you using in your IDE?

indico/modules/events/sessions/templates/management/_types_table.html
```diff
@@ -0,0 +1,18 @@
+{% macro types_table_row(session_type) %}
```
mic4ael: This is a duplicate of what we have for contribution types. Maybe it would be worth it to extract it to a more generic template that could be used in different places?

indico/modules/events/sessions/templates/management/types_dialog.html
```diff
@@ -0,0 +1,103 @@
+{% from 'message_box.html' import message_box %}
+{% from 'events/sessions/management/_types_table.html' import types_table_row %}
+
+<div class="manage-session-types">
+    <div class="toolbar space-after hide-if-locked">
+        <button id="js-new-session-type" class="i-button icon-plus highlight" data-ajax-dialog
```
mic4ael: I would `js-new-session-type` move to the `class` attribute since this is what we have been doing so far ;)

indico/modules/events/sessions/models/types.py
```diff
@@ -0,0 +1,74 @@
...
+        default=False
+    )
+
+    event = db.relationship(
+        'Event',
+        lazy=True,
+        backref=db.backref(
+            'session_types',
+            cascade='all, delete-orphan',
+            lazy='dynamic'
```
ThiefMaster: Why not just a normal lazy relationship?

indico/modules/events/sessions/forms.py
```diff
@@ -100,3 +102,24 @@ def session_fields(self):
...
+                             description=_("If selected, this session type will appear in timetable "
+                                           "as a poster session."))
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event')
+        self.session_type = kwargs.get('obj')
+        super(SessionTypeForm, self).__init__(*args, **kwargs)
+
+    def validate_name(self, field):
+        query = self.event.session_types.filter(db.func.lower(SessionType.name) == field.data.lower())
```
ThiefMaster: If you get rid of the lazy=dynamic of the relationship, you need to use this instead:

    query = SessionType.query.with_parent(self.event).filter(...)

indico/modules/events/sessions/templates/management/_types_table.html
ThiefMaster: Not sure if it's really worth the effort or even makes much sense since the poster column is only used here, and there are quite a few references to "session types" in other lines (e.g. confirmation messages), so the amount of reusable code isn't that big.

indico/modules/events/sessions/templates/management/types_dialog.html
ThiefMaster: :+1: on that; also move the `data-ajax-dialog` at the end (we put data attrs with no value last)

indico/modules/events/sessions/templates/management/types_dialog.html
```diff
@@ -0,0 +1,103 @@
...
+            var $row = $this.closest('tr');
+            $.ajax({
+                url: $this.data('href'),
+                method: $this.data('method'),
+                complete: IndicoUI.Dialogs.Util.progress(),
+                error: handleAjaxError,
+                success: function() {
+                    $row.remove();
+                    dialogModified();
+                    if ($('.manage-session-types table tbody tr').length == 0) {
```
ThiefMaster: `===` (and maybe this whole JS block should be in a js file?)

indico/modules/events/sessions/templates/management/types_dialog.html
```diff
@@ -0,0 +1,103 @@
...
+            if (data) {
+                var $lastRow = $('.manage-session-types table tr:last');
+                if ($lastRow.length) {
+                    $lastRow.after(data.html_row);
+                } else {
+                    $manageSessionTypes.trigger('ajaxDialog:reload');
+                }
+            }
+        });
+
```
ThiefMaster: too many blank lines

indico/modules/events/sessions/util.py
```diff
@@ -195,3 +197,9 @@ def get_session_timetable_pdf(sess, **kwargs):
     return TimeTablePlain(sess.event, session.user, showSessions=[sess.id], showDays=[],
                           sortingCrit=None, ttPDFFormat=pdf_format, pagesize='A4', fontsize='normal',
                           firstPageNumber=1, showSpeakerAffiliation=False, **kwargs)
+
+
+def session_type_row(session_type):
```
ThiefMaster: `render_session_type_row`

indico/htdocs/sass/modules/_sessions.scss
ThiefMaster: to use only as much space as actually needed for its contents

indico/migrations/versions/20171214_1059_9c4418d7a6aa_add_session_types_table.py
meluru: I thought we use `server_default` only when we add a new column to an existing table, so that the old records have some value.

indico/modules/events/sessions/blueprint.py
meluru: 120, but it's true that I didn't need to wrap this one :)

indico/modules/events/sessions/templates/management/_types_table.html
meluru: I agree with @ThiefMaster At first I thought about making some parts more generic, but I figured it's not really worth complicating.

indico/migrations/versions/20171214_1059_9c4418d7a6aa_add_session_types_table.py
ThiefMaster: yeah, no need to do that when creating a new table

indico/modules/events/contributions/templates/management/types_dialog.html
```diff
@@ -61,19 +63,19 @@
...
-        $('#js-new-contribution-type').on('ajaxDialog:closed', function(evt, data) {
+        $('.js-new-contribution-type').on('ajaxDialog:closed', function(evt, data) {
             evt.preventDefault();
             if (data) {
                 var $lastRow = $('.manage-contribution-types table tr:last');
                 if ($lastRow.length) {
                     $lastRow.after(data.html_row);
                 } else {
-                    $('.manage-contribution-types').trigger('ajaxDialog:reload');
+                    $manageContributionTypestrigger('ajaxDialog:reload');
```
mvidalgarcia: :warning: Missing dot! (`.trigger`)

indico/modules/events/sessions/controllers/management/sessions.py
```diff
@@ -50,7 +54,8 @@ def _get_session_list_args(event):
                          subqueryload('blocks').undefer('contribution_count'))
                 .order_by(db.func.lower(Session.title))
                 .all())
-    return {'sessions': sessions, 'default_colors': get_colors()}
+    types = [{'id': int(t.id), 'title': t.name} for t in SessionType.query.with_parent(event).all()]
```
mvidalgarcia: Why do you need to cast the id?

indico/modules/events/sessions/controllers/management/sessions.py
```diff
@@ -245,3 +264,73 @@ def _get_form_defaults(self):
 class RHSessionBlocks(RHManageSessionBase):
     def _process(self):
         return jsonify_template('events/sessions/management/session_blocks.html', sess=self.session)
+
+
+class RHManageSessionTypes(RHManageSessionsBase):
+    """Dialog to manage the SessionTypes of an event"""
```
mvidalgarcia: I'd go for "session types" instead of "SessionTypes" as you do in `RHManageSessionTypeBase` since this is an explanatory docstring.

indico/modules/events/sessions/templates/management/_types_table.html
```diff
@@ -0,0 +1,18 @@
...
+            <a class="js-edit-session-type icon-edit"
+               data-ajax-dialog
+               data-href="{{ url_for('.manage_type', session_type) }}"
+               data-title="{% trans %}Edit session type{% endtrans %}"></a>
+            <a class="js-delete-session-type icon-remove"
+               data-method="POST"
+               data-href="{{ url_for('.delete_type', session_type) }}"
+               data-confirm="{% trans %}Do you really want to delete this session type?{% endtrans %}"
+               data-title="{% trans %}Confirm deletion{% endtrans %}"></a>
+      </td>
```
mvidalgarcia: Indentation

indico/modules/events/sessions/templates/management/types_dialog.html
```diff
@@ -0,0 +1,104 @@
+{% from 'message_box.html' import message_box %}
+{% from 'events/sessions/management/_types_table.html' import types_table_row %}
+
+<div class="manage-session-types">
+    <div class="toolbar space-after hide-if-locked">
+        <button class="i-button js-new-session-type icon-plus highlight"
+                data-href="{{ url_for('.create_type', event) }}"
+                data-title="{% trans %}Add new session type{% endtrans %}"
+                data-ajax-dialog>
+            {% trans %}New session type{% endtrans %}
```
mvidalgarcia: `{%- trans %}`  

Same for other occurrences below.

indico/modules/events/sessions/controllers/management/sessions.py
ThiefMaster: Shouldn't be needed. And the `.all()` is not necessary; iterating a query does the job already.

indico/migrations/versions/20171214_1059_9c4418d7a6aa_add_session_types_table.py
mic4ael: ahh, sorry for that, didn't notice that it is indeed a new table ;)

indico/migrations/versions/20171214_1059_9c4418d7a6aa_add_session_types_table.py
```diff
@@ -0,0 +1,44 @@
...
+        sa.PrimaryKeyConstraint('id'),
+        schema='events'
+    )
+    op.create_index(op.f('ix_session_types_event_id'), 'session_types', ['event_id'], unique=False, schema='events')
+    op.create_foreign_key(op.f('fk_session_types_event_id_events'), 'session_types', 'events', ['event_id'], ['id'],
```
ThiefMaster: same here

indico/migrations/versions/20171214_1059_9c4418d7a6aa_add_session_types_table.py
```diff
@@ -0,0 +1,44 @@
...
+                          source_schema='events', referent_schema='events')
+    op.add_column('sessions', sa.Column('type_id', sa.Integer(), nullable=True), schema='events')
+    op.create_index(op.f('ix_sessions_type_id'), 'sessions', ['type_id'], unique=False, schema='events')
```
ThiefMaster: and here

indico/migrations/versions/20171214_1059_9c4418d7a6aa_add_session_types_table.py
```diff
@@ -0,0 +1,44 @@
...
+    op.create_foreign_key(op.f('fk_sessions_type_id_session_types'), 'sessions', 'session_types', ['type_id'], ['id'],
```
ThiefMaster: and here ;)

indico/migrations/versions/20171214_1059_9c4418d7a6aa_add_session_types_table.py
```diff
@@ -0,0 +1,44 @@
...
+                          source_schema='events', referent_schema='events')
+
+
+def downgrade():
+    op.drop_constraint(op.f('fk_sessions_type_id_session_types'), 'sessions', schema='events', type_='foreignkey')
```
ThiefMaster: no need for the `op.f()` wrapper here (but it doesn't do any harm either)

also, I think the FK and index will be deleted automatically when dropping the column. see if just dropping the columns isn't enough.

indico/migrations/versions/20171214_1059_9c4418d7a6aa_add_session_types_table.py
```diff
@@ -0,0 +1,44 @@
...
+    op.drop_index(op.f('ix_sessions_type_id'), table_name='sessions', schema='events')
+    op.drop_column('sessions', 'type_id', schema='events')
+    op.drop_constraint(op.f('fk_session_types_event_id_events'), 'session_types', schema='events', type_='foreignkey')
```
ThiefMaster: same here, dropping the table will delete all the indexes/constraints on its columns

indico/modules/events/sessions/controllers/management/sessions.py
```diff
@@ -146,16 +151,30 @@ def _process_DELETE(self):
     def _process_PATCH(self):
         data = request.json
         updates = {}
-        if set(data.viewkeys()) > {'colors'}:
+        if set(data.viewkeys()) > {'colors', 'type_id'}:
```
ThiefMaster: this isn't correct:

    >>> {'a', 'c'} > {'a', 'b'}
    False

What we actually want here is `set(data.viewkeys()) - {'colors', 'type_id'}` to fail if there is anything besides the allowed ones.

FWIW, with a single value in the setusing `>` worked, but with more than one on the right side it doesn't.

indico/modules/events/sessions/controllers/management/sessions.py
```diff
@@ -245,3 +264,73 @@ def _get_form_defaults(self):
...
         return jsonify_template('events/sessions/management/session_blocks.html', sess=self.session)
+
+
+class RHManageSessionTypes(RHManageSessionsBase):
+    """Dialog to manage the session types of an event"""
+
+    def _process(self):
+        types = SessionType.query.with_parent(self.event).all()
+        return jsonify_template('events/sessions/management/types_dialog.html', event=self.event,
+                                types=types)
```
ThiefMaster: `types=self.event.session_types`

indico/modules/events/sessions/controllers/management/sessions.py
```diff
@@ -245,3 +264,73 @@ def _get_form_defaults(self):
...
+            self.event.log(EventLogRealm.management, EventLogKind.change, 'Sessions',
+                           'Updated type: {}'.format(old_name), session.user)
+            return jsonify_data(html_row=render_session_type_row(self.session_type), flash=False)
+        return jsonify_form(form)
+
+
+class RHCreateSessionType(RHManageSessionsBase):
+    """Dialog to add a SessionType"""
+
+    def _get_response(self, new_type):
```
ThiefMaster: no need to split this into a separate method; we only did it in some other places where we wanted to override it in a subclass

indico/modules/events/sessions/controllers/management/sessions.py
```diff
@@ -245,3 +264,73 @@ def _get_form_defaults(self):
...
+        types = [{'id': t.id, 'title': t.name} for t in SessionType.query.with_parent(self.event)]
```
ThiefMaster: `self.event.session_types`

indico/modules/events/sessions/templates/display/session_display.html
```diff
@@ -131,7 +132,7 @@ <h3>{% trans %}Description{% endtrans %}</h3>
                     {%- trans %}Contribution list{% endtrans -%}
                 </a>
             {% endif %}
-            {% if not sess.is_poster %}
+            {% if not is_poster %}
```
ThiefMaster: keep this as it was before and add a property to Session that contains the logic above

indico/htdocs/js/indico/modules/contributions/types_dialog.js
```diff
@@ -0,0 +1,61 @@
...
+                    $lastRow.after(data.html_row);
+                } else {
+                    $manageContributionTypes.trigger('ajaxDialog:reload');
+                }
+            }
+        });
+
+        $manageContributionTypes.on('ajaxDialog:closed', '.js-edit-contribution-type', function(evt, data) {
+            evt.preventDefault();
+            var $row = $(this).closest('tr');
```
mic4ael: You are using it only if `data` is not empty. It should be better to move it inside the `if`.

`$(this).closest('tr').replaceWith(data.html_row);`

indico/htdocs/js/indico/modules/contributions/types_dialog.js
```diff
@@ -0,0 +1,61 @@
...
+            if (data) {
+                $row.replaceWith(data.html_row);
+                dialogModified();
+            }
+        });
+
+        $manageContributionTypes.on('indico:confirmed', '.js-delete-contribution-type', function(evt) {
+            evt.preventDefault();
+            var $this = $(this);
+            var $row = $this.closest('tr');
```
mic4ael: I think this should be also somewhere else, namely inside the `success` callback.

indico/htdocs/js/indico/modules/sessions/types_dialog.js
```diff
@@ -0,0 +1,61 @@
+(function(global) {
```
mic4ael: This file differs from `contributions/types_dialog.js` only in one class name. Shouldn't it be one file to make it more generic and not repeat the code?

indico/modules/events/sessions/models/sessions.py
```diff
@@ -205,6 +205,10 @@ def conveners(self):
                 .distinct(SessionBlockPersonLink.person_id)
                 .all())
 
+    @property
+    def is_poster(self):
+        return self.type and self.type.is_poster
```
ThiefMaster: that'll return None instead of False if there's no type.

indico/modules/events/sessions/controllers/management/sessions.py
```diff
@@ -146,16 +151,30 @@ def _process_DELETE(self):
     def _process_PATCH(self):
         data = request.json
         updates = {}
-        if set(data.viewkeys()) > {'colors'}:
+        if set(data.viewkeys()) - {'colors', 'type_id'}:
```
pferreir: Nitpicking here, but wouldn't `set(data)` give you the same?

CHANGES.rst
```diff
@@ -28,13 +28,17 @@ Improvements
   by users submitting an abstract (:issue:`3138`)
 - Add support for boolean (yes/no) and freetext questions in abstract
   reviewing (:issue:`3175`)
+- Add session types to sessions; allow marking session types as poster
+  so that the sessions with poster session type are displayed differently
+  in the timetable
```
ThiefMaster: poster sessions aren't a new feature, so I would change this a bit, maybe something like this:

```
- Add support for custom session types (:issue:`3189`)
- Move poster session flag from session settings to session type settings
```

CHANGES.rst
```diff
@@ -28,13 +28,17 @@ Improvements
...
 
 Internal Changes
 ^^^^^^^^^^^^^^^^
 
-- Rename *Roles* in ACL entries to *Permissions*.  This especially affects
+- Rename *Roles* in ACL entries to *Permissions*. This especially affects
   the ``can_manage`` method whose ``role`` argument has been renamed to
   ``permission`` (:issue:`3057`)
+- Change poster sessions to sessions with poster session type
```
ThiefMaster: I don't think this is really worth mentioning here; the idea of the "internal changes" section is to point out things likely to be relevant to developers

CHANGES.rst
```diff
@@ -28,13 +28,17 @@ Improvements
...
   reviewing (:issue:`3175`)
+- Add session types to sessions; allow marking session types as poster
+  so that the sessions with poster session type are displayed differently
+  in the timetable
 
 Internal Changes
 ^^^^^^^^^^^^^^^^
 
-- Rename *Roles* in ACL entries to *Permissions*.  This especially affects
+- Rename *Roles* in ACL entries to *Permissions*. This especially affects
```
ThiefMaster: The double spaces are intentional. In plaintext blobs it's not too uncommon to do this to keep some more spacing between sentences. You can find this in many docstrings or e.g. the GPL license text as well.

indico/htdocs/sass/modules/_types_dialog.scss
```diff
@@ -0,0 +1,39 @@
...
+
+    table.i-table {
+        table-layout: auto;
+        min-width: 360px;
+        margin-bottom: 1.5em;
+
+        th, td {
+            white-space: normal;
+
+            &.name-column {
```
ThiefMaster: I think we don't want this anymore - see #3215

indico/modules/events/sessions/templates/management/_types_table.html
```diff
@@ -0,0 +1,18 @@
+{% macro types_table_row(session_type) %}
+    <tr class="i-table">
+        <td class="i-table name-column">{{ session_type.name }}</td>
```
ThiefMaster: remove `name-column` (see my other comment in the css)

indico/modules/events/sessions/templates/management/types_dialog.html
```diff
@@ -0,0 +1,45 @@
...
+                data-ajax-dialog>
+            {%- trans %}New session type{% endtrans %}
+        </button>
+    </div>
+
+    {% if types %}
+        <table class="i-table">
+            <thead>
+                <tr class="i-table">
+                    <th class="i-table name-column">
```
ThiefMaster: remove `name-column` (see my other comment in the css)

indico/modules/events/sessions/controllers/management/sessions.py
```diff
@@ -146,16 +151,30 @@ def _process_DELETE(self):
...
             updates['colors'] = colors
+        if 'type_id' in data:
+            updates.update(self._get_session_type_updates(data['type_id']))
         update_session(self.session, updates)
         return jsonify()
 
+    def _get_session_type_updates(self, type_id):
+        updates = {}
+        if type_id is None:
+            updates['type_id'] = None
```
ThiefMaster: Try using `'type'` here; like this you use the relationship instead of the FK column

indico/modules/events/sessions/controllers/management/sessions.py
```diff
@@ -146,16 +151,30 @@ def _process_DELETE(self):
...
+        else:
+            type_ = SessionType.query.with_parent(self.event).filter_by(id=type_id).first()
+            if not type_:
+                raise BadRequest('Invalid type id')
+            if not self.session.type or type_id != self.session.type.id:
+                updates['type_id'] = type_id
```
ThiefMaster: same here; you already query the type so it's better to assign that to the relationship instead of the ID to the FK (which will be done internally by SA on flush)

indico/modules/events/sessions/controllers/management/sessions.py
```diff
@@ -245,3 +264,70 @@ def _get_form_defaults(self):
...
         return jsonify_template('events/sessions/management/session_blocks.html', sess=self.session)
+
+
+class RHManageSessionTypes(RHManageSessionsBase):
+    """Dialog to manage the session types of an event"""
+
+    def _process(self):
+        types = SessionType.query.with_parent(self.event).all()
+        return jsonify_template('events/sessions/management/types_dialog.html', event=self.event,
+                                types=types)
```
ThiefMaster: `types=self.event.session_types` and the query above can go away

indico/modules/events/sessions/templates/management/_session_list.html
```diff
@@ -43,13 +43,20 @@
...
-                                        {%- trans %}Poster{% endtrans -%}
-                                    {% else %}
-                                        {%- trans %}Standard{% endtrans -%}
-                                    {% endif -%}
-                                </span>
+                                <a class="i-button {{ 'session-type-picker' if not event.is_locked else 'disabled' }}"
+                                   data-items="{{ types | tojson | forceescape }}"
+                                   data-href="{{ url_for('.session_rest', sess) }}"
+                                   data-method="PATCH"
+                                   data-selected-item-id="{{ (sess.type.id if sess.type else None) | tojson }}">
```
ThiefMaster: lowercase `none` in jinja

ThiefMaster: theoretically you should use ` | forceescape` here as well, but since it's always a number or null it doesn't make a difference

meluru:
Updated

indico/migrations/versions/20171214_1059_9c4418d7a6aa_add_session_types_table.py
```diff
@@ -0,0 +1,68 @@
+"""add session types table
+
+Revision ID: 9c4418d7a6aa
+Revises: 566d5de4e0e5
```
ThiefMaster: 2af245be72a6

indico/migrations/versions/20171214_1059_9c4418d7a6aa_add_session_types_table.py
```diff
@@ -0,0 +1,68 @@
+"""add session types table
```
ThiefMaster: uppercase A ;)

PR #3195, opened 2018-01-08
Show warning when registering on Indico Hub

mic4ael:


indico/modules/bootstrap/controllers.py
```diff
@@ -44,14 +47,26 @@ class RHBootstrap(RH):
     def _process_GET(self):
         if User.query.filter_by(is_system=False).has_rows():
             return redirect(url_for_index())
+
+        try:
+            hostname = urlparse(request.url_root).hostname.strip('[]')
```
ThiefMaster: mostly a matter of preference, but i usually use `url_parse` from `werkzeug.urls` (we use this in many other places in indico as well)

indico/modules/bootstrap/controllers.py
```diff
@@ -44,14 +47,26 @@ class RHBootstrap(RH):
...
+            host_data = socket.getaddrinfo(hostname, 80, 0, 0, socket.IPPROTO_TCP)
+            for item in host_data:
```
ThiefMaster: local_instance = any(ipaddress.ip_address(unicode(item[4][0])).is_private for item in host_data)

indico/modules/bootstrap/controllers.py
```diff
@@ -44,14 +47,26 @@ class RHBootstrap(RH):
     def _process_GET(self):
         if User.query.filter_by(is_system=False).has_rows():
             return redirect(url_for_index())
+
+        try:
+            hostname = urlparse(request.url_root).hostname.strip('[]')
+            host_data = socket.getaddrinfo(hostname, 80, 0, 0, socket.IPPROTO_TCP)
```
ThiefMaster: I think you can pass `None` for the port since you don't really care about it

indico/modules/bootstrap/controllers.py
```diff
@@ -44,14 +47,26 @@ class RHBootstrap(RH):
...
         return render_template('bootstrap/bootstrap.html',
                                form=BootstrapForm(),
                                timezone=config.DEFAULT_TIMEZONE,
                                languages=get_all_locales(),
                                operating_system=get_os(),
                                postgres_version=get_postgres_version(),
                                indico_version=indico.__version__,
-                               python_version=python_version())
+                               python_version=python_version(),
+                               warn_before_hub_registration=local_instance or config.DEBUG)
```
ThiefMaster: `show_local_warning`?

and IMO it's cleaner to wrap the value in `()`

- `foo=x or y`
- `foo=(x or y)`

with the latter it's clearer that everything there belongs to foo

indico/modules/bootstrap/templates/bootstrap.html
```diff
@@ -100,41 +101,47 @@ <h1 class="step-title">{% trans %}Site configuration{% endtrans %}</h1>
...
+                            Please note that <strong>no private information</strong> will ever be sent to the
+                            Community Hub or the Indico Project and that you can disable it any time (from the
+                            "Community Hub" section of the administrative area).
+                        {%- endtrans -%}
+                    </div>
                 </div>
                 <div class="fields-container">
                     {{ form_rows(form, fields=('enable_tracking', 'contact_name', 'contact_email'), placeholder=True) }}
+                    {% call message_box('warning') %}
+                        {% trans %}
```
ThiefMaster: `{% trans -%}...{%- endtrans %}`

indico/modules/cephalopod/templates/cephalopod.html
```diff
@@ -42,22 +42,30 @@
             </div>
         </div>
     </div>
+    {% if show_local_warning %}
+        {% call message_box('warning') %}
+            {% trans %}
```
ThiefMaster: see my other comment

indico/util/network.py
```diff
@@ -0,0 +1,30 @@
...
+import ipaddress
+import socket
+
+from werkzeug.urls import url_parse
+
+
+def is_using_local_url(url):
+    """Check if the provided URL points to a private IP address."""
+    try:
+        hostname = url_parse(url).host.strip('[]')
```
ThiefMaster: I don't think this can raise any of the exceptions you are catching, so I'd move it outside the try block

indico/util/network.py
```diff
@@ -0,0 +1,30 @@
...
+        host_data = socket.getaddrinfo(hostname, None, 0, 0, socket.IPPROTO_TCP)
+        return any(ipaddress.ip_address(unicode(item[4][0])).is_private for item in host_data)
+    except socket.gaierror, ipaddress.AddressValueError:
```
ThiefMaster: Add some parentheses:

    except (socket.gaierror, ipaddress.AddressValueError):

This is clearer since `except Foo, exc` is valid and does something completely different (it's the deprecated legacy version of `except Foo as exc`)

indico/util/network_test.py
```diff
@@ -0,0 +1,31 @@
...
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+import pytest
+
+from indico.util.network import is_using_local_url
+
+
+@pytest.mark.parametrize(('url', 'expected'), (
+    ('http://127.0.0.1', True),
+    ('https://[::0]', True),
```
ThiefMaster: localhost is `[::1]` ;)

indico/util/network.py
```diff
@@ -0,0 +1,30 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+import ipaddress
+import socket
+
+from werkzeug.urls import url_parse
+
+
+def is_using_local_url(url):
```
ThiefMaster: `is_private_url` maybe?

indico/modules/bootstrap/templates/bootstrap.html
```diff
@@ -100,41 +101,47 @@ <h1 class="step-title">{% trans %}Site configuration{% endtrans %}</h1>
...
+                            "Community Hub" section of the administrative area).
+                        {%- endtrans -%}
+                    </div>
                 </div>
                 <div class="fields-container">
                     {{ form_rows(form, fields=('enable_tracking', 'contact_name', 'contact_email'), placeholder=True) }}
+                    {% call message_box('warning') %}
+                        {% trans -%}
+                            It seems like you're using a local URL for Indico.
+                            We recommend registering only production instances with the community hub.
```
mvidalgarcia: **C**ommunity **H**ub? We capitalize first letter as you can see some lines above.

Also, shouldn't sound better, "It's recommended". IMO using "We" doesn't sound very formal, but not really strong opinion.

indico/modules/bootstrap/controllers.py
```diff
@@ -51,7 +52,8 @@ def _process_GET(self):
                                operating_system=get_os(),
                                postgres_version=get_postgres_version(),
                                indico_version=indico.__version__,
-                               python_version=python_version())
+                               python_version=python_version(),
+                               show_local_warning=(is_private_url(request.url_root) or config.DEBUG))
```
mvidalgarcia: Are you using this in `bootstrap.html`? I think you forgot the check.

indico/modules/cephalopod/templates/cephalopod.html
```diff
@@ -42,22 +42,30 @@
             </div>
         </div>
     </div>
+    {% if show_local_warning %}
+        {% call message_box('warning') %}
+            {% trans -%}
+                It seems like you're using a local URL for Indico.
+                We recommend registering only production instances with the community hub.
```
mvidalgarcia: Same thing

indico/modules/bootstrap/templates/bootstrap.html
```diff
@@ -100,41 +101,47 @@ <h1 class="step-title">{% trans %}Site configuration{% endtrans %}</h1>
...
+                    <div>
+                        {%- trans -%}
+                            Please note that <strong>no private information</strong> will ever be sent to the
+                            Community Hub or the Indico Project and that you can disable it any time (from the
+                            "Community Hub" section of the administrative area).
+                        {%- endtrans -%}
+                    </div>
                 </div>
                 <div class="fields-container">
                     {{ form_rows(form, fields=('enable_tracking', 'contact_name', 'contact_email'), placeholder=True) }}
```
ThiefMaster: `=true` (since you are changing so much in here, let's fix that as well)

indico/modules/bootstrap/controllers.py
mic4ael: indeed! forgot about it!

mic4ael:
Updated

indico/modules/bootstrap/controllers.py
ThiefMaster: I'd reverse the order of the checks, reading `config.DEBUG` is cheaper than a call that involves network activity

indico/modules/bootstrap/templates/bootstrap.html
```diff
@@ -100,41 +101,49 @@ <h1 class="step-title">{% trans %}Site configuration{% endtrans %}</h1>
...
-                            {%- trans -%}
-                                Please note that <strong>no private information</strong> will ever be sent to the
-                                Community Hub or the Indico Project and that you can disable it any time (from the
-                                "Community Hub" section of the administrative area).
-                            {%- endtrans -%}
-                        </div>
+                    <div>
+                        {%- trans -%}
+                            Be part of the ever growing Indico community! Choose to receive news on the latest
+                            releases (2&ndash;3 times per year) as well as important security advisories.
```
ThiefMaster: why don't we just use the literal ndash instead of the entity? `–`

mic4ael: don't know, it was there already

PR #3200, opened 2018-01-10
Prevent repeated positions when disabling regform sections

mvidalgarcia:
Use the same position system that it's being used with fields. Disabled sections have positions >= 1000

### DIFF
--- a/indico/modules/events/registration/util.py
+++ b/indico/modules/events/registration/util.py
@@ -16,7 +16,9 @@
 
 from __future__ import unicode_literals
 
+import itertools
 from collections import OrderedDict
+from operator import attrgetter
 
 from flask import current_app, json, session
 from qrcode import QRCode, constants
@@ -493,3 +495,18 @@ def generate_ticket(registration):
 
 def get_ticket_attachments(registration):
     return [('Ticket.pdf', generate_ticket(registration).getvalue())]
+
+
+def update_regform_item_positions(regform):
+    """Update positions when deleting/disabling an item in order to prevent gaps"""
+    section_positions = itertools.count(1)
+    disabled_section_positions = itertools.count(1000)
+    for section in sorted(regform.sections, key=attrgetter('position')):
+        section_active = section.is_enabled and not section.is_deleted
+        section.position = next(section_positions if section_active else disabled_section_positions)
+        # ensure consistent field ordering
+        positions = itertools.count(1)
+        disabled_positions = itertools.count(1000)
+        for child in section.children:
+            child_active = child.is_enabled and not child.is_deleted
+            child.position = next(positions if child_active else disabled_positions)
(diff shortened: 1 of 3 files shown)


indico/modules/events/registration/controllers/management/util.py
```diff
@@ -0,0 +1,24 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+
+def remove_regform_item_gaps(form_items, from_position):
```
ThiefMaster: Why add a new util.py in controllers/management and not just add it to registration/util.py

indico/modules/events/registration/controllers/management/util.py
```diff
@@ -0,0 +1,24 @@
...
+    """Update positions when deleting/disabling an item in order to prevent gaps"""
+    for item in form_items:
+        if item.position > from_position and item.is_enabled:
```
ThiefMaster: Untested, but wouldn't something like this be safer? Unlike the current code it would ensure consistent positions on all items.

```python
positions = itertools.count(1)
disabled_positions = itertools.count(1)
for item in sorted(form_items, key=attrgetter('position')):
    pos = next(positions if item.is_enabled and not item.is_deleted else disabled_positions)
    item.position = pos
```

We could even go one step further and do this on the whole regform:

```
section_positions = itertools.count(1)
disabled_section_positions = itertools.count(1000)
for section in sorted(regform.sections, key=attrgetter('position')):
    section_active = section.is_enabled and not section.is_deleted
    section.position = next(section_positions if section_active else disabled_section_positions)
    # ensure consistent field ordering
    positions = itertools.count(1)
    disabled_positions = itertools.count(1000)
    for child in section.children:
        child_active = child.is_enabled and not child.is_deleted
        child.position = next(positions if child_active else disabled_positions)
```

indico/modules/events/registration/util.py
```diff
@@ -493,3 +493,10 @@ def generate_ticket(registration):
 
 def get_ticket_attachments(registration):
     return [('Ticket.pdf', generate_ticket(registration).getvalue())]
+
+
+def remove_regform_item_gaps(form_items, from_position):
+    """Update positions when deleting/disabling an item in order to prevent gaps"""
+    for item in form_items:
+        if item.position > from_position and item.is_enabled:
```
ThiefMaster: Untested, but wouldn't something like this be safer? Unlike the current code it would ensure consistent positions on all items.

```python
positions = itertools.count(1)
disabled_positions = itertools.count(1)
for item in sorted(form_items, key=attrgetter('position')):
    pos = next(positions if item.is_enabled and not item.is_deleted else disabled_positions)
    item.position = pos
```

We could even go one step further and do this on the whole regform:

```python
section_positions = itertools.count(1)
disabled_section_positions = itertools.count(1000)
for section in sorted(regform.sections, key=attrgetter('position')):
    section_active = section.is_enabled and not section.is_deleted
    section.position = next(section_positions if section_active else disabled_section_positions)
    # ensure consistent field ordering
    positions = itertools.count(1)
    disabled_positions = itertools.count(1000)
    for child in section.children:
        child_active = child.is_enabled and not child.is_deleted
        child.position = next(positions if child_active else disabled_positions)
```

mvidalgarcia: Looks quite safe, if we really want to be cautions sounds like a good idea.

PR #3207, opened 2018-01-17
Allow users to clone a contribution

mic4ael:


indico/modules/events/contributions/clone.py
```diff
@@ -98,14 +104,48 @@ def run(self, new_event, cloners, shared_data):
...
+                self._clone_contrib()
+            else:
+                self._clone_contribs(new_event)
+
         self._synchronize_friendly_id(new_event)
         db.session.flush()
         return {'contrib_map': self._contrib_map, 'subcontrib_map': self._subcontrib_map}
 
-    def _clone_contribs(self, new_event):
+    def _create_new_contribution(self, event, contribution, excluded_attrs=None):
```
ThiefMaster: I think keeping it named `old_contrib` is clearer, since then you have `<new|old>_contrib` instead of `new_contrib` and `contribution`.

indico/modules/events/contributions/controllers/management.py
```diff
@@ -679,3 +680,10 @@ def _process_POST(self):
         self.subcontrib.references.append(reference)
         db.session.flush()
         return self.jsonify_reference(reference)
+
+
+class RHCloneContribution(RHManageContributionBase):
+    def _process(self):
+        clone_contribution(self.event, self.contrib, False)
```
ThiefMaster: I would use a kwarg for the last one, or just omit it and set a default value of False.

indico/modules/events/contributions/controllers/management.py
```diff
@@ -679,3 +680,10 @@ def _process_POST(self):
...
+        db.session.commit()
+        return jsonify_data(html=self.list_generator.render_list()['html'])
```
ThiefMaster: What about using the same way like we do for all the other endpoints?

Not sure if there's anything besides `html` but if it's used to update the contribution list it'd probably break if we ever added something else since here that would then be missing.

indico/modules/events/contributions/operations.py
```diff
@@ -181,3 +182,22 @@ def create_contribution_from_abstract(abstract, contrib_session=None):
...
+        'sessions': {
+            'session_map': {},
+            'session_block_map': {}
+        },
+        'contribution_fields': {'contrib_field_map': {field: field for field in event.contribution_fields}},
+        'contribution_types': {'contrib_type_map': {type_: type_ for type_ in event.contribution_types}}
+    }
+
+    if preserve_session:
+        shared_data['sessions']['session_map'][contribution.session] = contribution.session
```
ThiefMaster: You could always populate this above; no harm in having it even when the session is not used.

indico/modules/events/contributions/operations.py
```diff
@@ -181,3 +182,22 @@ def create_contribution_from_abstract(abstract, contrib_session=None):
                                        'session': contrib_session,
                                        'person_link_data': {link: True for link in contrib_person_links}},
                                custom_fields_data=custom_fields_data)
+
+
+def clone_contribution(event, contribution, preserve_session):
+    shared_data = {
+        'event_persons': {'person_map': {person: person for person in event.persons}},
```
ThiefMaster: You could actually do this: `dict(zip(event.persons, event.persons))`

The literals you used are also quite clear, just mentioning that this would work too :p

indico/modules/events/contributions/operations.py
```diff
@@ -181,3 +182,22 @@ def create_contribution_from_abstract(abstract, contrib_session=None):
...
+        'contribution_fields': {'contrib_field_map': {field: field for field in event.contribution_fields}},
+        'contribution_types': {'contrib_type_map': {type_: type_ for type_ in event.contribution_types}}
+    }
+
+    if preserve_session:
+        shared_data['sessions']['session_map'][contribution.session] = contribution.session
+        shared_data['sessions']['session_block_map'][contribution.session_block] = contribution.session_block
+
+    cloner = ContributionCloner(event, contribution=contribution, preserve_session=preserve_session)
+    cloner.run(event, [], shared_data)
```
ThiefMaster: the 2nd argument is  usually a set. so maybe passing an empty `set()` is safer

indico/modules/events/timetable/controllers/manage.py
```diff
@@ -167,3 +170,17 @@ def _process_PATCH(self):
...
+
+class RHCloneContribution(RHManageTimetableBase):
+    def _process_args(self):
+        RHManageTimetableBase._process_args(self)
+        self.contribution = (Contribution.query.with_parent(self.event)
+                             .filter_by(id=request.args['contrib_id'])
+                             .first())
+
+    def _process(self):
+        clone_contribution(self.event, self.contribution, True)
```
ThiefMaster: kwarg style for the last arg

indico/modules/events/timetable/controllers/manage.py
```diff
@@ -167,3 +170,17 @@ def _process_PATCH(self):
...
+        db.session.commit()
```
ThiefMaster: why are you committing here? `db.session.flush()` is all you need if you just need to populate the id and default args

indico/modules/events/contributions/controllers/management.py
```diff
@@ -679,3 +680,10 @@ def _process_POST(self):
         self.subcontrib.references.append(reference)
         db.session.flush()
         return self.jsonify_reference(reference)
+
+
+class RHCloneContribution(RHManageContributionBase):
+    def _process(self):
+        clone_contribution(self.event, self.contrib, False)
+        db.session.commit()
```
ThiefMaster: why are you committing here? `db.session.flush()` is all you need if you just need to populate the id and default args

indico/modules/events/contributions/clone.py
```diff
@@ -98,14 +104,48 @@ def run(self, new_event, cloners, shared_data):
...
         attrs = (get_simple_column_attrs(Contribution) | {'own_room', 'own_venue'}) - {'abstract_id'}
+        if excluded_attrs is not None:
+            attrs -= excluded_attrs
+        new_contrib = Contribution()
+        new_contrib.populate_from_attrs(old_contrib, attrs)
+        new_contrib.subcontributions = list(self._clone_subcontribs(old_contrib.subcontributions))
+        new_contrib.acl_entries = clone_principals(ContributionPrincipal, old_contrib.acl_entries)
+        new_contrib.references = list(self._clone_references(ContributionReference, old_contrib.references))
+        new_contrib.person_links = list(self._clone_person_links(ContributionPersonLink,
+                                        old_contrib.person_links))
```
ThiefMaster: I think this now fits in one line

indico/modules/events/contributions/clone.py
```diff
@@ -98,14 +104,48 @@ def run(self, new_event, cloners, shared_data):
...
+                new_contrib.session_block = self._session_block_map[old_contrib.session_block]
+        event.contributions.append(new_contrib)
+        return new_contrib
+
+    def _clone_contrib(self):
+        new_contribution = self._create_new_contribution(self.old_event, self.contribution, {'friendly_id'})
+        if self.preserve_session:
+            entry = schedule_contribution(new_contribution, self.contribution.timetable_entry.start_dt,
+                                          session_block=new_contribution.session_block)
+            self.old_event.timetable_entries.append(entry)
```
ThiefMaster: I really dislike the idea of using `old_event` for *any* operation that creates something new. When cloning something within the event, the new event should be the same as `old_event` anyway, but by not using the old event you remove the risk of code actually messing with the old event during normal event cloning.

indico/modules/events/contributions/clone.py
```diff
@@ -98,14 +104,48 @@ def run(self, new_event, cloners, shared_data):
...
         self._subcontrib_map = {}
+
         with db.session.no_autoflush:
-            self._clone_contribs(new_event)
+            if new_event == self.old_event:
+                self._clone_contrib()
+            else:
+                self._clone_contribs(new_event)
+
         self._synchronize_friendly_id(new_event)
```
ThiefMaster: This shouldn't run when cloning a single contribution.

indico/modules/events/contributions/clone.py
mic4ael: Good point.

ThiefMaster:
Some unit tests for `clone_contribution` would be nice!

PR #3215, opened 2018-01-22
Properly show long contribution type names

mic4ael:


mvidalgarcia:
In these kind of PRs a screenshot (or a couple - after/before) would help a lot!

PR #3220, opened 2018-01-24
Support transparent background images in badges

mic4ael:


ThiefMaster:
changelog please ;)

    - Fix badges/posters with transparent background images (:issue:`3220`)

PR #3228, opened 2018-02-02
Pin default themes to the top for event managers

### LINKED ISSUE #3166: Request: "Default" themes pinned to the top for event administrators/managers
This concerns the instance at CERN.

When a user is not signed in or is not the administrator of an event, they are presented with a limited set of "default" themes to skin the event:

firefox - - screen shot 24 nov 2017 at 20 58 36

However, when you log in as an administrator or manager, you see a much longer list:

firefox - - screen shot 24 nov 2017 at 20 56 43

Is it possible to show this second list with the "default" themes from the first list pinned to the top and clearly separated (by a line of some sort) from the other options?

Thank you.

mvidalgarcia:
Resolves #3166

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,6 +44,7 @@ Improvements
 - Add contribution cloning within an event (:issue:`3207`)
 - Add option to include the event description in reminder emails
   (:issue:`3157`, thanks :user:`bpedersen2`)
+- Pin default themes to the top for event managers (:issue:`3166`)
 
 Internal Changes
 ^^^^^^^^^^^^^^^^
--- a/indico/modules/events/settings.py
+++ b/indico/modules/events/settings.py
@@ -226,6 +226,7 @@ class ThemeSettingsProxy(object):
             # to override a default style so let them do so...
             for name, definition in settings.get('definitions', {}).viewitems():
                 definition['plugin'] = plugin
+                definition.setdefault('user_visible', False)
                 core_settings['definitions'][name] = definition
         return core_settings
 
--- a/indico/modules/events/templates/header.html
+++ b/indico/modules/events/templates/header.html
@@ -38,15 +38,20 @@
        title="{% trans %}Change theme{% endtrans %}"></a>
     <ul class="dropdown">
         {% set endpoint = 'timetable.timetable' if event.type == 'conference' else 'events.display' %}
-        {% for id, data in themes.items()|sort(attribute='1.name')
-           if event.can_manage(session.user) or data.user_visible %}
-            <li>
-                {% if id == theme %}
-                    <a class="disabled"><strong>{{ data.name }}</strong></a>
-                {% else %}
-                    <a href="{{ url_for(endpoint, event, view=id) }}">{{ data.name }}</a>
-                {% endif %}
-            </li>
+        {% for group in themes.items()|sort(attribute='1.name')|groupby('1.user_visible')|reverse
+           if group.grouper or event.can_manage(session.user) %}
+            {% if not group.grouper %}
+                <li class="themes-separator"></li>
+            {% endif %}
+            {% for id, data in group.list %}
+                <li>
+                    {% if id == theme %}
+                        <a class="disabled"><strong>{{ data.name }}</strong></a>
+                    {% else %}
+                        <a href="{{ url_for(endpoint, event, view=id) }}">{{ data.name }}</a>
+                    {% endif %}
+                </li>
+            {% endfor %}
         {% endfor %}
     </ul>
 {%- endmacro -%}
--- a/indico/htdocs/sass/modules/event_display/_common.scss
+++ b/indico/htdocs/sass/modules/event_display/_common.scss
@@ -112,6 +112,12 @@
                 font-weight: bold;
             }
         }
+
+        .themes-separator {
+            @include border-bottom($black);
+            height: auto;
+            margin: 0 1em;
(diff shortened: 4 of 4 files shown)


indico/modules/events/templates/header.html
```diff
@@ -38,15 +38,20 @@
...
-            <li>
-                {% if id == theme %}
-                    <a class="disabled"><strong>{{ data.name }}</strong></a>
-                {% else %}
-                    <a href="{{ url_for(endpoint, event, view=id) }}">{{ data.name }}</a>
-                {% endif %}
-            </li>
+        {% for user_visible, themes in themes.items()|sort(attribute='1.name')|groupby('1.user_visible')|reverse
+           if event.can_manage(session.user) or user_visible %}
+            {% if user_visible == False %}
```
mic4ael: either `false` or just `not user_visible`?

indico/modules/events/templates/header.html
```diff
@@ -38,15 +38,20 @@
...
-            <li>
-                {% if id == theme %}
-                    <a class="disabled"><strong>{{ data.name }}</strong></a>
-                {% else %}
-                    <a href="{{ url_for(endpoint, event, view=id) }}">{{ data.name }}</a>
-                {% endif %}
-            </li>
+        {% for user_visible, themes in themes.items()|sort(attribute='1.name')|groupby('1.user_visible')|reverse
+           if event.can_manage(session.user) or user_visible %}
+            {% if user_visible == false %}
```
ThiefMaster: `if not user_visible`

indico/modules/events/templates/header.html
```diff
@@ -38,15 +38,20 @@
...
-           if event.can_manage(session.user) or data.user_visible %}
-            <li>
-                {% if id == theme %}
-                    <a class="disabled"><strong>{{ data.name }}</strong></a>
-                {% else %}
-                    <a href="{{ url_for(endpoint, event, view=id) }}">{{ data.name }}</a>
-                {% endif %}
-            </li>
+        {% for user_visible, themes in themes.items()|sort(attribute='1.name')|groupby('1.user_visible')|reverse
+           if event.can_manage(session.user) or user_visible %}
```
ThiefMaster: It was already like this before, but I would reverse the order of the checks. Checking the truthiness of a simple variable is usually much cheaper than any method call

indico/modules/events/templates/header.html
```diff
@@ -38,15 +38,20 @@
...
-        {% for id, data in themes.items()|sort(attribute='1.name')
-           if event.can_manage(session.user) or data.user_visible %}
-            <li>
-                {% if id == theme %}
-                    <a class="disabled"><strong>{{ data.name }}</strong></a>
-                {% else %}
-                    <a href="{{ url_for(endpoint, event, view=id) }}">{{ data.name }}</a>
-                {% endif %}
-            </li>
+        {% for user_visible, themes in themes.items()|sort(attribute='1.name')|groupby('1.user_visible')|reverse
```
ThiefMaster: I think shadowing `themes` with `themes` is rather ugly. I'd try to find a different name (which might be hard) or use the grouper/list style of using groupby as shown in the docs (didn't even know that existed, just saw it when looking if there's a `reversed` attribute for groupby!): http://jinja.pocoo.org/docs/2.10/templates/#groupby

indico/modules/events/templates/header.html
mvidalgarcia: There are themes with `user_visible = None`

mvidalgarcia:
Updated

indico/modules/events/settings.py
```diff
@@ -226,6 +226,7 @@ def settings(self):
             # to override a default style so let them do so...
             for name, definition in settings.get('definitions', {}).viewitems():
                 definition['plugin'] = plugin
+                definition['user_visible'] = definition.get('user_visible', False)
```
ThiefMaster: definition.setdefault('user_visible', False)

indico/modules/events/templates/header.html
```diff
@@ -38,15 +38,20 @@
...
-        {% for id, data in themes.items()|sort(attribute='1.name')
-           if event.can_manage(session.user) or data.user_visible %}
-            <li>
-                {% if id == theme %}
-                    <a class="disabled"><strong>{{ data.name }}</strong></a>
-                {% else %}
-                    <a href="{{ url_for(endpoint, event, view=id) }}">{{ data.name }}</a>
-                {% endif %}
-            </li>
+        {% for grouper, list in themes.items()|sort(attribute='1.name')|groupby('1.user_visible')|reverse
```
ThiefMaster: I was more thinking about `for group in ...` and then using `group.grouper` and `group.list` in the code below

ThiefMaster: FWIW, if you want to keep the argument unpacking style, maybe go for `user_visible` and `themes_list`... the latter is not very nice naming but at least I can't think of anything better..

CHANGES.rst
```diff
@@ -69,6 +69,7 @@ Improvements
   by Indico admins (:issue:`3210`)
 - Sort contribution types alphabetically
 - Add folding indicators when printing foldable badges (:issue:`3216`)
+- Pin default themes to the top for event managers (:issue:`3166`)
```
ThiefMaster: This is in the wrong section now (should go up to the 2.1 improvements)

PR #3246, opened 2018-02-21
Change the version of Google Maps API

mic4ael:
Without reverting to 3.31 the map in the RB module freezes IE11


ThiefMaster:
Let's reproduce the issue in a small example and open an issue with Google. If it's a problem on their side they should fix it instead of us locking the API version..

mic4ael:
Fixed in the latest release of the Google Maps API.

PR #3247, opened 2018-02-23
Permissions widget in contributions and sessions

mvidalgarcia:


indico/core/permissions.py
```diff
@@ -73,12 +71,29 @@ def get_permissions_info(_type):
     :param _type: The type of the permissions retrieved (e.g. Event, Category)
     :return: A tuple containing a dict with the available permissions and a dict with the permissions tree
     """
+    from indico.modules.events import Event
+    from indico.modules.events.contributions import Contribution
+    from indico.modules.events.sessions import Session
+
+    description_mapping = {
+        FULL_ACCESS_PERMISSION: {
+            Event: 'Unrestricted management access for the whole event',
```
ThiefMaster: missing i18n

indico/core/permissions.py
```diff
@@ -73,12 +71,29 @@ def get_permissions_info(_type):
...
+            Session: 'Access the public areas of the selected session',
+            Contribution: 'Access the public areas of the selected contribution'
+        }
+    }
+
     selectable_permissions = {k: v for k, v in get_available_permissions(_type).viewitems() if v.user_selectable}
     special_permissions = {
         FULL_ACCESS_PERMISSION: {'title': _('Manage'), 'css_class': 'danger',
-                                 'description': _('Unrestricted management access for the whole event')},
+                                 'description': _(description_mapping[FULL_ACCESS_PERMISSION][_type])},
```
ThiefMaster: this does not work, you **cannot** have dynamic things  inside `_()`

indico/htdocs/js/indico/jquery/permissions_widget.js
```diff
@@ -91,10 +91,20 @@
             var self = this;
             var $permissions = $('<div>', {class: 'permissions-box flexrow f-a-center f-self-stretch'});
             var $permissionsList = $('<ul>').appendTo($permissions);
+            // When full access is enabled, always show read access
+            if (_.contains(permissions, '_full_access') && !_.contains(permissions, '_read_access')) {
```
ThiefMaster: can we use `FULL_ACCESS_PERMISSION` / `READ_ACCESS_PERMISSION` constants here as well, just like we do in the python code?

indico/web/forms/fields/principals.py
```diff
@@ -133,17 +136,30 @@ class PermissionsField(JSONField):
     widget = JinjaWidget('forms/permissions_widget.html', single_kwargs=True, acl=True)
 
     def __init__(self, *args, **kwargs):
+        self.object_type = kwargs.pop('object_type')
         super(PermissionsField, self).__init__(*args, **kwargs)
         self.ip_networks = map(serialize_ip_network_group, IPNetworkGroup.query.filter_by(hidden=False))
-        self.permissions_info = get_permissions_info(Event)[0]
+
+    @staticmethod
+    def get_type_mapping():
```
ThiefMaster: this could simply be a class-level attribute

ThiefMaster:
Tests are failing due to a circular import. Might need to move some imports inside a function.

mvidalgarcia:
Updated.

indico/htdocs/sass/widgets/_permissions.scss
```diff
@@ -95,10 +95,21 @@
                     margin: 0.4em 0.5em;
                     padding: 0.6em 1em;
                     text-transform: uppercase;
+                    color: $light-gray;
 
                     &.disabled {
                         @include opacity(0.5);
                     }
+
+                    &.visited {
```
ThiefMaster: Is that really a good name? It doesn't really have anything to do with what it's used for...

mvidalgarcia: I wasn't sure about the name and I chose it because we used `.visited` in other places for that color. Do you think it'd be better to use the actual name of the permissions, i.e `.paper-manager`?

ThiefMaster: Yep, maybe we could automatically add a `permission-<objtype.lower>-<name>` class to these elements? So you'd have e.g. `.permission-event-paper_manager` for the paper manager permission.

PR #3248, opened 2018-02-26
Show warning box when reviewer roles do not apply (no tracks)

mvidalgarcia:


indico/modules/events/abstracts/templates/forms/track_role_widget.html
```diff
@@ -56,6 +56,18 @@
     <input type="hidden" id="{{ field.id }}" name="{{ field.name }}" value="{{ field._value() }}">
     <div id="role-widget-{{ field.id }}" class="abstract-role-widget">
         <ul class="abstract-role-list">
+            {% if not field.tracks %}
+                <div class="warning-message-box">
```
ThiefMaster: Why not use the `message_box` macro?

PR #3249, opened 2018-02-26
Add contact info to docs

mvidalgarcia:


docs/source/index.rst
```diff
@@ -96,3 +96,9 @@ Indices and tables
 .. |license| image:: https://img.shields.io/github/license/indico/indico.png
                    :alt: License
                    :target: https://github.com/indico/indico/blob/master/COPYING
+
+Contact
+++++++++++++++++++
```
ThiefMaster: that's way too many `+` chars... the underline should be just as long as the line before

CHANGES.rst
```diff
@@ -56,6 +56,7 @@ Improvements
 - When blocking a user account, the user will be forcefully logged out in
   addition to being prevented from logging in
 - Warn when editing a speaker/author would result in duplicate emails
+- Add contact information to documentation (:issue:`3111`)
```
ThiefMaster: IMO documentation changes aren't really useful in the changelog

docs/source/index.rst
```diff
@@ -96,3 +96,9 @@ Indices and tables
...
+
+* IRC Channel: Official `#indico <https://webchat.freenode.net/?channels=indico>`_ channel at Freenode (irc.freenode.net). Also accessible through `Matrix <https://riot.im/app/#/room/#indico:matrix.org>`_.
+* Discussion forum: `Indico Community <https://talk.getindico.io>`_ Discourse instance
```
ThiefMaster: is the fact that it's a Discourse instance really relevant?

mvidalgarcia: Just copied it from https://getindico.io/contact/, but I agree it's no really relevant.

ThiefMaster:
It would be nice if it showed up in the side menu (with the RTD theme) - otherwise people who are e.g. in the setup guide won't see it. Probably needs to be a separate page for that :/

ThiefMaster:
http://fabric-www.readthedocs.io/contact.html looks pretty nice for example

PR #3251, opened 2018-02-27
Update default columns for abstract list

### LINKED ISSUE #2813: Abstract list: Smarter default columns
If there are no tracks in an event it makes no sense to show the track-related columns (accepted track, submitted for tracks, reviewed for tracks) by default. They just waste space in this case.

meluru:
Show track related columns in abstract list only if there are tracks
available for the event

closes #2813


meluru:
With this change the track related columns will not show if there are no tracks, until we customize the the list.

PR #3262, opened 2018-03-12
Event page: Add keywords to metadata in header

bpedersen2:
Currently the  assigned keywords are only use in some calendar views,
but for search engines they should also be provided in the keywords
meta header.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -66,6 +66,7 @@ Improvements
 - Show a nicer error message when entering an excessively high base
   registration fee (:issue:`3260`)
 - Use proper British English for person titles (:issue:`3279`)
+- Add event keywords in meta tags (:issue:`3262`, thanks :user:`bpedersen2`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/events/templates/meta.html
+++ b/indico/modules/events/templates/meta.html
@@ -2,3 +2,6 @@
 <meta property="og:image" content="{{ event.logo_url if event.has_logo else
                                       (indico_config.IMAGES_BASE_URL + '/logo_indico.png') }}">
 <meta property="og:description" content="{{ event.description|striptags|truncate(500) }}">
+{% if event.keywords -%}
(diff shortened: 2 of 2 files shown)


indico/modules/events/templates/meta.html
```diff
@@ -2,3 +2,6 @@
 <meta property="og:image" content="{{ event.logo_url if event.has_logo else
                                       (indico_config.IMAGES_BASE_URL + '/logo_indico.png') }}">
 <meta property="og:description" content="{{ event.description|striptags|truncate(500) }}">
+{%- if event.keywords -%}
+<meta keywords="{{ event.keywords|join(',') }}">
```
ThiefMaster: I think the correct syntax is `<meta name="keywords" value="...">`

This line should be indented (we care about readable templates, even if this makes the final HTML output uglier).

Also, I think right now this ends up right behind the `og:description` line without a linebreak, so I would replace the `-%}` with `%}` or `{%-` with `{%}` (the `-` strips whitespace on the side where it is).

CHANGES.rst
```diff
@@ -65,6 +65,7 @@ Improvements
   reviewing
 - Show a nicer error message when entering an excessively high base
   registration fee (:issue:`3260`)
+- Add the keywords as meta tag to the page header(thanks :user:`bpedersen2`)
```
ThiefMaster: Missing space, and IMHO this can be a bit more concise:

```markdown
- Expose event keywords through an HTML meta tag (:issue:`3262`, thanks
  :user:`bpedersen2`)
```

ThiefMaster:
There's a double space in your commit message (`the  assigned`) - no big deal, but if you amend the commit with the suggestions above, this could be changed at the same time ;)

indico/modules/events/templates/meta.html
bpedersen2: Done

CHANGES.rst
bpedersen2: Done

PR #3263, opened 2018-03-13
Allow service requests to be auto-accepted

pferreir:
Required by https://github.com/indico/indico-plugins-cern/pull/11

### DIFF
--- a/indico/modules/events/requests/base.py
+++ b/indico/modules/events/requests/base.py
@@ -61,6 +61,8 @@ class RequestDefinitionBase(object):
     manager_form = RequestManagerForm
     #: default values to use if there's no existing request
     form_defaults = {}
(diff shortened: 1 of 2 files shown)


indico/modules/events/requests/base.py
```diff
@@ -61,6 +61,8 @@ class RequestDefinitionBase(object):
     manager_form = RequestManagerForm
     #: default values to use if there's no existing request
     form_defaults = {}
+    #: whether requests of this type should be automatically accepted
```
ThiefMaster: isn't this misleading since it's still up to the implementation to do the auto-accepting?

pferreir: OK, let's do this in the way we discussed IRL.

Issue #3272, opened 2018-03-20
Registrant list: Error if accomodation display is selected and a no-accomodation registration exists


bpedersen2:
Patch coming...
Traceback:
```
Traceback (most recent call last):
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/indico/web/flask/util.py", line 114, in wrapper
    return obj().process()
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/indico/web/rh.py", line 270, in process
    res = self._do_process()
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/indico/web/rh.py", line 249, in _do_process
    return self._process()
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/indico/modules/events/registration/controllers/management/reglists.py", line 90, in _process
    has_badges=has_badges, has_tickets=has_tickets, **reg_list_kwargs)
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/indico/web/views.py", line 123, in render_template
    return cls(g.rh, *wp_args, **context).display()
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/indico/web/views.py", line 231, in display
    body = to_unicode(self._display(params))
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/indico/web/views.py", line 261, in _display
    return self._applyDecoration(self._getBody(params))
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/indico/modules/events/management/views.py", line 74, in _getBody
    return self._getPageContent(params)
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/indico/web/views.py", line 150, in _getPageContent
    return self.render_template_func(template, **params)
  File "/usr/home/indicoadm/indicov2/.indicov2/lib/python2.7/site-packages/flask/templating.py", line 134, in render_template
[...]

PR #3273, opened 2018-03-20
Fix listing of accomodations in registrant listing #3272

bpedersen2:
As the templates assume nights to  always be set, add a default
value for the no_accomodation case as well.


bpedersen2:
Needs improvment, as it now erros out on plurals (this expects an integer)

PR #3277, opened 2018-03-27
Legacy LaTeX pdfs: Use colored links

bpedersen2:
Add the hyperref option colorlinks to generate colored links
in pdf files instead of the default red boxes.

User-visible change: The links will be colored both onscreen and
in the printed pdf.

### DIFF
--- a/indico/legacy/webinterface/tpls/latex/inc/document.tpl
+++ b/indico/legacy/webinterface/tpls/latex/inc/document.tpl
@@ -16,6 +16,7 @@
 </%block>
 
 \usepackage{hyperref}
+\usepackage[ocgcolorlinks]{ocgx2}[2017/03/30]
 \usepackage{amssymb}
 \usepackage{amsmath}
 \usepackage{fontspec}


ThiefMaster:
Just FYI, LaTeX PDFs themselves aren't legacy ;) It's just that the code running it is still very old and ugly and thus still in `indico/legacy`

indico/legacy/webinterface/tpls/latex/inc/document.tpl
```diff
@@ -15,7 +15,7 @@
 \usepackage[a4paper, top=1em, bottom=10em]{geometry}
 </%block>
 
-\usepackage{hyperref}
+\usepackage[ocgcolorlinks]{hyperref}
```
bpedersen2: What is the minium required texlive version currently? 
If a new enough version ( > 2017..) is allowed, then 
`\usepackage{hyperref}
\usepackage[ocgcolorlinks]{ocgx2}[2017/03/30]`
would be a n even better option. (see [
_
TeX - LaTeX Stack Exchange](https://tex.stackexchange.com/questions/4425/)

ThiefMaster: Since we did the stable release of 2.0 this year and texlive always installs the latest version (you can't even easily install an older one), I'd say we can safely assume people have at least 2017.

ThiefMaster:
Closing in favor of #3283

PR #3278, opened 2018-03-27
Prevent now happening element from changing height

mic4ael:


ThiefMaster:
changelog entry missing:

    - Fix display issue in "how happening" (:issue:`3278`)

PR #3280, opened 2018-03-29
Add event manager placeholder in perms widget for contrib/session

mvidalgarcia:


ThiefMaster:
Just thinking... shouldn't we show a "Category Managers" equivalent, i.e. "Event Managers" instead of the placeholder? After all, they always have access...

mvidalgarcia:
Updated with the discussed changes

PR #3283, opened 2018-04-03
LaTeX pdfs: Allow link format customisation

bpedersen2:
Make the hyperref link color options configurabable.

This is a more complete alternative to #3277

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,8 @@ Improvements
   :user:`bpedersen2`)
 - Improve message when the call for abstracts is scheduled to open but
   hasn't started yet
+- Make link color handling for LaTeX pdfs configurable (:issue:`3283`,
+  thanks :user:`bpedersen2`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/events/abstracts/settings.py
+++ b/indico/modules/events/abstracts/settings.py
@@ -42,7 +42,6 @@ class BOALinkFormat(RichEnum):
 
     frame = ('', '')
     colorlinks = ('[colorlinks]', '')
-    ocgx2colorlinks = ('', '\\usepackage[ocgcolorlinks]{ocgx2}[2017/03/30]')
     unstyled = ('[hidelinks]', '')
 
 
@@ -69,7 +68,6 @@ BOACorrespondingAuthorType.__titles__ = {
 BOALinkFormat.__titles__ = {
     BOALinkFormat.frame: _('Border around links (screen only)'),
     BOALinkFormat.colorlinks: _('Color links'),
-    BOALinkFormat.ocgx2colorlinks: _('Color links (screen only)'),
     BOALinkFormat.unstyled: _('Do not highlight links')
 }
 
--- a/indico/modules/events/abstracts/settings.py
+++ b/indico/modules/events/abstracts/settings.py
@@ -32,6 +32,20 @@ class BOACorrespondingAuthorType(RichEnum):
     speakers = 'speakers'
 
 
+class BOALinkFormat(RichEnum):
+    """LaTeX book of abstracts link format setting
+
+    value is a 2-tuple of strings:
+    first is the hyperref option to use
+    second sets additional tex commands
+    """
+
+    frame = ('', '')
+    colorlinks = ('[colorlinks]', '')
+    ocgx2colorlinks = ('', '\\usepackage[ocgcolorlinks]{ocgx2}[2017/03/30]')
+    unstyled = ('[hidelinks]', '')
+
+
 BOASortField.__titles__ = {
     BOASortField.id: _('ID'),
     BOASortField.abstract_title: _('Abstract title'),
@@ -51,6 +65,13 @@ BOACorrespondingAuthorType.__titles__ = {
     BOACorrespondingAuthorType.speakers: _('Speakers')
 }
 
+
+BOALinkFormat.__titles__ = {
+    BOALinkFormat.frame: _('Border around links (screen only)'),
+    BOALinkFormat.colorlinks: _('Color links'),
+    BOALinkFormat.ocgx2colorlinks: _('Color links (screen only)'),
+    BOALinkFormat.unstyled: _('Do not highlight links')
+}
 
 abstracts_settings = EventSettingsProxy('abstracts', {
     'description_settings': {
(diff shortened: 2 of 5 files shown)


indico/modules/events/abstracts/settings.py
```diff
@@ -53,6 +60,13 @@ class BOACorrespondingAuthorType(RichEnum):
 }
 
 
+BOALinkFormatting.__titles__ = {
+    BOALinkFormatting.frame: _('Frame'),
+    BOALinkFormatting.colorlinks: _('Color Links'),
+    BOALinkFormatting.ocgcolorlinks: _('Color Links for screen only'),
+    BOALinkFormatting.hidelinks: _('Hide Links'),
+    }
```
ThiefMaster: the `}` shouldn't be indented

CHANGES.rst
```diff
@@ -65,6 +65,8 @@ Improvements
   reviewing
 - Show a nicer error message when entering an excessively high base
   registration fee (:issue:`3260`)
+- Make link color handling for LaTeX pdfs configurable (thanks
```
ThiefMaster: can you add a link to the PR as well (like in other changelog entries)?

bpedersen2: Done

indico/modules/events/abstracts/settings.py
bpedersen2: Done

CHANGES.rst
```diff
@@ -79,6 +79,10 @@ Bugfixes
   (:issue:`3272`, thanks :user:`bpedersen2`)
 - Use the 'Reviewing' realm when logging actions from the abstract/paper
   reviewing modules
+- Make link color handling for LaTeX pdfs configurable (thanks
```
ThiefMaster: duplicate ;)

indico/modules/events/abstracts/settings.py
```diff
@@ -37,6 +37,24 @@ class BOACorrespondingAuthorType(RichEnum):
...
+    second sets additional tex commands
+
+    See `indico/legacy/webinterface/tpls/latex/inc/document.tpl`
+    for the usage.
+    '''
+
+    frame = ('', '')
+    colorlinks = ('[colorlinks]', '')
+    ocgcolorlinks = ('[ocgcolorlinks]', '')
+    ocgx2colorlinks = ('', '\\usepackage[ocgcolorlinks]{ocgx2}[2017/03/30]')
```
ThiefMaster: Matter of taste, but I would use a raw string (`r'\usepackage.....'`) instead of escaping the backslash.

indico/modules/events/abstracts/settings.py
```diff
@@ -37,6 +37,24 @@ class BOACorrespondingAuthorType(RichEnum):
...
 
 
+class BOALinkFormatting(RichEnum):
+    '''LaTeX book of abstracts link format setting
+
+    value is a 2-tuple of strings:
+    first is the hyperref option to use
+    second sets additional tex commands
+
+    See `indico/legacy/webinterface/tpls/latex/inc/document.tpl`
```
ThiefMaster: rst uses dual-backticks to wrap monospaced text, but I don't think the reference is needed. It's something that'll likely get outdated at some point ;)

indico/modules/events/abstracts/settings.py
```diff
@@ -53,6 +71,14 @@ class BOACorrespondingAuthorType(RichEnum):
 }
 
 
+BOALinkFormatting.__titles__ = {
+    BOALinkFormatting.frame: _('Frame'),
+    BOALinkFormatting.colorlinks: _('Color Links'),
+    BOALinkFormatting.ocgcolorlinks: _('Color Links for screen only'),
+    BOALinkFormatting.ocgx2colorlinks: _('Color Links for screen only(with ogcx2)'),
```
ThiefMaster: What's the difference to the option above? A typical user will not know what "ogcx2" is so it's a bit confusing. (also, missing space before the `(`).

indico/modules/events/abstracts/settings.py
```diff
@@ -53,6 +71,14 @@ class BOACorrespondingAuthorType(RichEnum):
...
+    BOALinkFormatting.hidelinks: _('Hide Links')
```
ThiefMaster: "Do not highlight links" or something like that would be better. I actually had to google to find out how the links are "hidden" (the answer is; "by using the normal text color/format")...

indico/modules/events/abstracts/settings.py
```diff
@@ -53,6 +68,14 @@ class BOACorrespondingAuthorType(RichEnum):
 }
 
 
+BOALinkFormatting.__titles__ = {
+    BOALinkFormatting.frame: _('Frame'),
+    BOALinkFormatting.colorlinks: _('Color Links'),
+    BOALinkFormatting.ocgcolorlinks: _('Color Links for screen only'),
+    BOALinkFormatting.ocgx2colorlinks: _('Color Links for screen only (alt., use ocgx2)'),
+    BOALinkFormatting.hidelinks: _('Do not highlight links')
```
ThiefMaster: what about renaming the enum option from `hidelinks` to `unstyled`?

indico/modules/events/abstracts/settings.py
```diff
@@ -53,6 +68,14 @@ class BOACorrespondingAuthorType(RichEnum):
 }
 
 
+BOALinkFormatting.__titles__ = {
+    BOALinkFormatting.frame: _('Frame'),
+    BOALinkFormatting.colorlinks: _('Color Links'),
```
ThiefMaster: lowercase `links` (also in then options below)

indico/modules/events/abstracts/settings.py
```diff
@@ -53,6 +68,14 @@ class BOACorrespondingAuthorType(RichEnum):
 }
 
 
+BOALinkFormatting.__titles__ = {
+    BOALinkFormatting.frame: _('Frame'),
```
ThiefMaster: What about something like "Border around links (screen only)"?

indico/modules/events/abstracts/settings.py
```diff
@@ -53,6 +68,14 @@ class BOACorrespondingAuthorType(RichEnum):
...
+    BOALinkFormatting.colorlinks: _('Color Links'),
+    BOALinkFormatting.ocgcolorlinks: _('Color Links for screen only'),
+    BOALinkFormatting.ocgx2colorlinks: _('Color Links for screen only (alt., use ocgx2)'),
```
ThiefMaster: I'm still unsure about this...

- if this is better than the option above and is supported with texlive 2017 and newer, let's just keep this one, without the `(...)` part in the title
- it's probably still confusing users - i guess even most LaTeX users have never heard of "ocgx2"

bpedersen2:
Adressed all comments.

ThiefMaster:
I just rebased this PR to the latest master and fixed some things, but I noticed that "Color links (screen only)" also affects printing. I tested it with texlive 2018 (installing 2019 atm in case this fixes it).

Can you confirm that a printed PDF does not colorize links for you when using this option?

bpedersen2:
Yes, seems like that depends on the pdf viewer used (I worked with  Adobe Reader, but seems to have no effect with evince under linux).

ThiefMaster:
I think I tried with sumatrapdf (and possibly foxit reader as well) and it was printed there...

ThiefMaster:
Maybe it'd be better to remove that particular option since we have absolutely no control about which PDF viewer people use?

Issue #3269, opened 2018-03-16
Meta information for schema.org/event is missing or incorrect


bpedersen2:
At least googlebot complains that some tags are missing and others use the wrong name:
![screenshot-search google com-2018 03 16-14-55-11](https://user-images.githubusercontent.com/5336494/37524406-a75be36c-2921-11e8-859f-f28bc0068226.png)

bpedersen2:
Need to  retest with 2.1 (this from the production system runnign 2.0.3)

pferreir:
We should definitely fix this.

bpedersen2:
Currently we add the meta info inline, split across the page. 

Shall  we consider switching to providing this information in a single place as json-ld? See ()[https://developers.google.com/search/docs/data-types/event] for an example. Then we could move all this meta-data stuff to a single template fragment.

PR #3287, opened 2018-04-04
Richcards

bpedersen2:
Instead of using inline microdata, render the metadata as json-ld in the header.

The resulting data validates in the [Google data testing tool](https://search.google.com/structured-data/testing-tool/) with just warnings about missing option entries (mainly: "offers")

Fixes: #3269

### DIFF
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -38,7 +38,7 @@ from indico.modules.categories.legacy import XMLCategorySerializer
 from indico.modules.categories.models.categories import Category
 from indico.modules.categories.serialize import (serialize_categories_ical, serialize_category, serialize_category_atom,
                                                  serialize_category_chain)
-from indico.modules.categories.util import get_category_stats, get_upcoming_events
+from indico.modules.categories.util import get_category_stats, get_upcoming_events, serialize_events_for_json_ld
 from indico.modules.categories.views import WPCategory, WPCategoryStatistics
 from indico.modules.events.models.events import Event
 from indico.modules.events.timetable.util import get_category_timetable
@@ -316,6 +316,7 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
         events_by_month = self.group_by_month(events)
 
         future_event_count = future_event_query.count()
+        future_events = future_event_query.all()
         past_event_count = past_event_query.count()
 
         show_future_events = bool(self.category.id in session.get('fetch_future_events_in', set()) or
@@ -338,10 +339,12 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
                   'past_event_count': past_event_count,
                   'show_past_events': show_past_events,
                   'past_threshold': past_threshold.strftime(threshold_format),
+                  'json_ld': serialize_events_for_json_ld(events + future_events),
                   'atom_feed_url': url_for('.export_atom', self.category),
                   'atom_feed_title': _('Events of "{}"').format(self.category.title)}
         params.update(get_base_ical_parameters(session.user, 'category',
                                                '/export/categ/{0}.ics'.format(self.category.id), {'from': '-31d'}))
+
 
         if not self.category.is_root:
             return WPCategory.render_template('display/category.html', self.category, **params)
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -243,8 +243,9 @@ class RHDisplayCategoryEventsBase(RHDisplayCategoryBase):
     _category_query_options = (joinedload('children').load_only('id', 'title', 'protection_mode'),
                                undefer('attachment_count'), undefer('has_events'))
     _event_query_options = (joinedload('person_links'), joinedload('series'), undefer_group('series'),
-                            load_only('id', 'category_id', 'created_dt', 'end_dt', 'protection_mode', 'start_dt',
-                                      'title', 'type_', 'series_pos', 'series_count'))
+                            load_only('id', 'category_id', 'created_dt', 'start_dt', 'end_dt', 'timezone',
+                                      'protection_mode', 'title', 'type_', 'series_pos', 'series_count',
+                                      'own_address', 'own_venue_id', 'own_venue_name'))
 
     def _process_args(self):
         RHDisplayCategoryBase._process_args(self)
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -313,12 +313,14 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
         future_event_query = event_query.filter(Event.start_dt >= future_threshold)
         current_event_query = event_query.filter(Event.start_dt >= past_threshold,
                                                  Event.start_dt < future_threshold)
-        events = current_event_query.filter(Event.start_dt < future_threshold).all()
+        json_ld_events = events = current_event_query.filter(Event.start_dt < future_threshold).all()
         events_by_month = self.group_by_month(events)
 
         future_event_count = future_event_query.count()
-        future_events = future_event_query.all()
         past_event_count = past_event_query.count()
+
+        if not session.user and future_event_count:
+            json_ld_events = json_ld_events + future_event_query.all()
 
         show_future_events = bool(self.category.id in session.get('fetch_future_events_in', set()) or
                                   (session.user and session.user.settings.get('show_future_events', False)))
@@ -340,7 +342,7 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
                   'past_event_count': past_event_count,
                   'show_past_events': show_past_events,
                   'past_threshold': past_threshold.strftime(threshold_format),
-                  'json_ld': map(serialize_event_for_json_ld, (events + future_events)),
+                  'json_ld': map(serialize_event_for_json_ld, json_ld_events),
                   'atom_feed_url': url_for('.export_atom', self.category),
                   'atom_feed_title': _('Events of "{}"').format(self.category.title)}
         params.update(get_base_ical_parameters(session.user, 'category',
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -345,7 +345,6 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
         params.update(get_base_ical_parameters(session.user, 'category',
                                                '/export/categ/{0}.ics'.format(self.category.id), {'from': '-31d'}))
 
-
         if not self.category.is_root:
             return WPCategory.render_template('display/category.html', self.category, **params)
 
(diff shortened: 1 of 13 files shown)


bpedersen2:
This pull request fixes the meta data problems for all event types and is also safe in case  a custom template overrides part of the content layout.

ThiefMaster:
If this is as widely supported as the previous attributes I really like it. Much less messy than having attributes spread all over for sure.

What I really dislike however is generating JSON with string operations. For example, in many places having a `"` would break the JSON. It's better to create a Python dict and then use the `|tojson` filter in Jinja to convert it to JSON. since `meta.html` is actually rendered from Python code, you can just create the dict there, pass it to the template, and then convert it to a JSON string in the template. Search for this line to find the place where it's rendered (events/views.py):

    meta = render_template('events/meta.html', event=self.event, site_name=site_name)

indico/modules/events/templates/display/conference/base.html
```diff
@@ -31,7 +31,7 @@ <h1>
                                        <img src="{{ event.logo_url }}" alt="{{ event.title }}" border="0" class="confLogo">
                                     </div>
                                 {% endif %}
-                                <span itemprop="title">{{ event.title }}</span>
```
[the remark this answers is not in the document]
bpedersen2: Done

bpedersen2:
Rewritten to prepare the dict first.

At least google states that the json-ld format is preferred over inline microdata.

indico/modules/events/views.py
```diff
@@ -120,9 +120,43 @@ def _getHeader(self):
...
+            "@type": "Event",
+            "url": self.event.external_url,
+            "name": self.event.title,
+            "startDate": self.event.start_dt.isoformat(),
+            "endDate": self.event.end_dt.isoformat(),
+            "location": {
+                "@type": "Place",
+                "name": self.event.venue_name or 'No location set',
+                "address": self.event.address or 'No address set'
+                },
```
ThiefMaster: you like this alignment, don't you :D

indico/modules/events/views.py
```diff
@@ -120,9 +120,43 @@ def _getHeader(self):
...
+            "description": strip_tags(self.event.description),
+            "performers": self._getJsonLDPerformers(self.event.person_links)
+        }
+        if self.event.has_logo:
+            lddata["image"] = url_for('event_images.logo_display', self.event, slug=self.event.logo_metadata['hash'],
```
ThiefMaster: Maybe you could add a `external_logo_url` property (right below the `logo_url`) property to `Event` and use that here?

indico/modules/events/views.py
```diff
@@ -120,9 +120,43 @@ def _getHeader(self):
     def getJSFiles(self):
         return WPDecorated.getJSFiles(self) + self._asset_env['modules_event_display_js'].urls()
 
+    def _getJsonLDData(self):
```
ThiefMaster: `_get_json_ld_data` (we don't use camelcase in new code)

indico/modules/events/views.py
```diff
@@ -120,9 +120,43 @@ def _getHeader(self):
...
+        lddata = {
```
ThiefMaster: `ld_data` or `data` (`lddata` looks a bit weird)

indico/modules/events/views.py
```diff
@@ -120,9 +120,43 @@ def _getHeader(self):
...
+            "@context": "http://schema.org",
```
ThiefMaster: usually we use single quotes everywhere

indico/modules/events/views.py
```diff
@@ -120,9 +120,43 @@ def _getHeader(self):
...
+            "@type": "Event",
+            "url": self.event.external_url,
+            "name": self.event.title,
+            "startDate": self.event.start_dt.isoformat(),
```
ThiefMaster: `start_dt `(UTC) or `start_dt_local` (event TZ)? Not sure what the spec requires

indico/modules/events/views.py
```diff
@@ -120,9 +120,43 @@ def _getHeader(self):
...
+            "endDate": self.event.end_dt.isoformat(),
+            "location": {
+                "@type": "Place",
+                "name": self.event.venue_name or 'No location set',
```
ThiefMaster: Is this required? Otherwise we could maybe just omit the location if there's no location information and/or use an empty string in case of just one of the two pieces of data being empty. IMO an empty string is more meaningful than a custom string where only a human knows that it means "not set/known"

indico/modules/events/views.py
```diff
@@ -120,9 +120,43 @@ def _getHeader(self):
...
+                },
+            "description": strip_tags(self.event.description),
+            "performers": self._getJsonLDPerformers(self.event.person_links)
+        }
+        if self.event.has_logo:
+            lddata["image"] = url_for('event_images.logo_display', self.event, slug=self.event.logo_metadata['hash'],
+                                      _external=True)
+        return lddata
+
+    def _getJsonLDPerformers(self, chairs):
```
ThiefMaster: no camelcase please ;)

indico/modules/events/views.py
```diff
@@ -120,9 +120,43 @@ def _getHeader(self):
...
+        performers = []
+        for chair in chairs:
+            performers.append({
+                "@type": "Person",
+                "name": chair.display_full_name,
+                "affiliation": {
+                    "@type": "Organization",
+                    "name": chair.affiliation
+                    }
```
ThiefMaster: `}` alignment - like above

indico/modules/events/views.py
ThiefMaster: I would change this method to act on a single person, and call it like this above:

    'performers': map(self._get_json_ld_performer, self.event.person_links)

indico/modules/events/views.py
bpedersen2: The spec wants isoformat, see http://schema.org/DateTime. But dt_local seems ok.

indico/modules/events/views.py
bpedersen2: At least the google tester flags this a errors, and an empty string is also not allowed.

indico/modules/events/views.py
ThiefMaster: sure, isoformat makes sense, the comment was just about whether to use UTC or local

indico/modules/events/views.py
bpedersen2: eclipse...

ThiefMaster: ouch ;) I didn't realize anyone still uses eclipse/pydev. Might want to look at PyCharm. But I guess eclipse/pydev also has a setting for the brace alignment

indico/modules/events/views.py
```diff
@@ -120,9 +120,40 @@ def _getHeader(self):
...
+            data['image'] = self.event.external_logo_url
+        return data
+
+    def _get_json_ldperformer(self, chair):
+        return {'@type': 'Person',
+                'name': chair.display_full_name,
+                'affiliation': {
+                    '@type': 'Organization',
+                    'name': chair.affiliation
+                }
```
ThiefMaster: return {'@type': 'Person',
            'name': chair.display_full_name,
            'affiliation': {'@type': 'Organization',
                            'name': chair.affiliation}}

or 

    return {
        '@type': 'Person',
        'name': chair.display_full_name,
        'affiliation': {
            '@type': 'Organization',
            'name': chair.affiliation
        }
    }

indico/modules/events/views.py
```diff
@@ -120,9 +120,40 @@ def _getHeader(self):
     def getJSFiles(self):
         return WPDecorated.getJSFiles(self) + self._asset_env['modules_event_display_js'].urls()
 
+    def _get_json_lddata(self):
```
ThiefMaster: `_get_json_ld_data`

indico/modules/events/views.py
```diff
@@ -120,9 +120,40 @@ def _getHeader(self):
...
+            },
+            'description': strip_tags(self.event.description),
+        }
+        if self.event.person_links:
+            data['performer'] = map(self._get_json_ldperformer, self.event.person_links)
+        if self.event.has_logo:
+            data['image'] = self.event.external_logo_url
+        return data
+
+    def _get_json_ldperformer(self, chair):
```
ThiefMaster: `_get_json_ld_performer`

indico/modules/events/views.py
bpedersen2: Done

indico/modules/events/views.py
bpedersen2: Done

bpedersen2:
On our production instance wiht these changes applied, google now renders the additional informations on the search results page  (at least for all pages re-crawled after application of this patch)

indico/modules/events/util.py
```diff
@@ -552,6 +553,39 @@ def serialize_event_for_ical(event, detail_level):
...
+        'endDate': event.end_dt_local.isoformat(),
+        'location': {
+            '@type': 'Place',
+            'name': event.venue_name or 'No location set',
+            'address': event.address or 'No address set'
+        },
+        'description': strip_tags(event.description),
+    }
+    if event.person_links:
+        data['performer'] = map(_get_json_ld_performer, event.person_links)
```
ThiefMaster: This breaks, I think you want to call `serialize_person_for_json_ld` here.

indico/modules/categories/controllers/display.py
```diff
@@ -338,11 +339,13 @@ def _process(self):
                   'past_event_count': past_event_count,
                   'show_past_events': show_past_events,
                   'past_threshold': past_threshold.strftime(threshold_format),
+                  'json_ld': serialize_events_for_json_ld(events + future_events),
```
ThiefMaster: this causes query spam (at least one extra query per event)

ThiefMaster:
I like the PR in general but I'm very unsure of adding the jsonld to the category page. Besides the query spam (which I'll take care of fixing if we decide to keep it): What's the benefit of including this information there? Do you have an example how Google etc. display this for the category page? Especially if there are many future events...

bpedersen2:
Data on category page: 
I added this as the data was already included in the page before this change. 

As for the rendering of  the data in google:
[https://www.google.com/search?q=cern+conference+category&q=cern+conference+category](https://www.google.com/search?q=cern+conference+category&oq=cern+conference+category)

indico/modules/events/util.py
```diff
@@ -552,6 +553,39 @@ def serialize_event_for_ical(event, detail_level):
...
+        data['performer'] = map(serialize_person_for_json_ld, event.person_links)
+    if event.has_logo:
+        data['image'] = event.external_logo_url
+    return data
+
+
+def serialize_person_for_json_ld(person):
+    return {
+        '@type': 'Person',
+        'name': person.display_full_name,
```
ThiefMaster: This should use `full_name` (`display_full_name` takes user preferences into account (whether the first or last name comes first etc) which aren't relevant anyway since search engines won't be logged in)

ThiefMaster:
> I added this as the data was already included in the page before this change.

Ah, hadn't noticed it since you didn't remove the old attributes.

I don't like the idea of having to query all the future events just to include them here... Also, we cannot include stuff like the speaker on the category listing page - you can have events there which are protected and not accessible by you (we do not want to do access checks for each displayed events, since at least when you are logged in these checks can be expensive when they result in group membership lookups).

So I'd keep only the same basic metadata we have right now: date, url, title
Like this you also avoid triggering extra queries to get e.g. the speakers.

bpedersen2:
Hmm, as the main users of these data are search engines, would it make sense to omit this if a user is logged in? Google and co. Are not logged typically.

Issue #3297, opened 2018-04-06
Conference timetable with indico style: add a link to event home page


jouvin:
Often, people prefer to use the Indico style timetable to the parallel one for conferences without parallel sessions as it is easier to navigate. This removes the conference menu on the left which is fine but has the drawback to remove any link to the event home page. Would it be possible to add a link back to the event page in the timetable in this case?

bpedersen2:
You can do this via template customisation:
locate indico/modules/events/templates/display/compact.html
copy it to custom/core/events/timetable/display/compact.html in your indico root

`@@ -15,7 +15,7 @@
                     </div>
                 </div>
             {% endif %}
-            {{ event.title }}
+            <a href="{{event.url}}">{{ event.title }}</a>
         </div>
         <br/>
`

This snippet would link the title to the 'normal' page. 

@ThiefMaster : Should I create a pull request with this change?

ThiefMaster:
I don't think this is only about the "compact" style but rather using meeting themes for conferences in general. While having a link back sounds like a good idea, I'm not sure how useful it would be in an actual meeting/lecture (the link would point to the current page after all).

jouvin:
@bpedersen2 thanks for snippet, it may help in the short term... but I agree with @ThiefMaster that it would be good if we could ensure that there is always a link back to the main event page from the conference timetables, whatever style is used. At least a useless link is harmless...

pferreir:
This is the good old problem of people "abusing" the meeting layout. Unfortunately, Pandora's box was opened long time ago and it's too late to just kill that possibility. Until we come up with a better conference timetable view, I agree that adding a link back won't hurt.

bpedersen2:
Agreed, this would need to be added to all meeting-like themes if the type is conference.

I think adding it to the header would work universally.

PR #3300, opened 2018-04-10
Add backlink to conference in title

bpedersen2:
For conference, if a theme is selected, add a backlink
to the main conference page.

Just using a home icon is not useful here, as the icon already
is used for the link to the indico home.

Fixes: #3297

### DIFF
--- a/indico/modules/events/templates/header.html
+++ b/indico/modules/events/templates/header.html
@@ -54,12 +54,6 @@
             {% endfor %}
         {% endfor %}
     </ul>
-    {% if event.type == 'conference' and theme %}
-        <span class="separator"></span>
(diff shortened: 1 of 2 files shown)


indico/modules/events/templates/header.html
```diff
@@ -54,6 +54,12 @@
             {% endfor %}
         {% endfor %}
     </ul>
+     {% if event.type == 'conference' and theme %}
+        <span class="separator"></span>
+         <a href="{{ event.url }}" class="i-button text-color subtle" title="">
```
ThiefMaster: ` title=""` isn't very useful ;)

bpedersen2: Remains from a version with icon :), fixed

PR #3319, opened 2018-04-19
Advanced filters (RB Room list)

mic4ael:


indico/modules/rb/models/locations.py
```diff
@@ -162,18 +162,22 @@ def get_equipment_by_name(self, name):
         return self.equipment_types.filter_by(name=name).first()
 
     def get_buildings(self):
-        building_rooms = defaultdict(list)
+        from indico.modules.rb_new.schemas import rooms_schema
```
ThiefMaster: I would add a new `get_buildings()` util function (or even method in the RH used for it) in the new RB module and keep this one as it is (and eventually get rid of it). It's quite specific to the room list and doesn't really need to be in the model.

indico/modules/rb/models/locations.py
```diff
@@ -162,18 +162,22 @@ def get_equipment_by_name(self, name):
...
                 continue
-            buildings.append({'number': building_name,
-                              'title': _(u'Building {}'.format(building_name)),
-                              'longitude': room_with_lat_lon.longitude,
-                              'latitude': room_with_lat_lon.latitude,
-                              'rooms': [r.to_serializable('__public_exhaustive__') for r in rooms]})
-        return buildings
+
+            buildings_tmp[building_name]['rooms'] = rooms_schema.dump(building_data['rooms']).data
+            buildings_tmp[building_name]['floors'] = sorted(list({room.floor for room in building_data['rooms']}))
```
ThiefMaster: no need to convert to a list, sorted works on any iterable

indico/modules/rb/models/locations.py
```diff
@@ -162,18 +162,22 @@ def get_equipment_by_name(self, name):
...
+            buildings_tmp[building_name].update({'number': building_name,
```
ThiefMaster: let's use either `[]` or `.update()` - but mixing it is a bit weird.

indico/modules/rb/models/locations.py
```diff
@@ -162,18 +162,22 @@ def get_equipment_by_name(self, name):
...
+                                                 'title': _(u'Building {}'.format(building_name)),
```
ThiefMaster: do we need this here? We could do this in client-side i18n

indico/modules/rb/models/locations.py
ThiefMaster: Also, I don't think it makes sense to have this limited to a specific location. As long as a room is active it should be used IMO.

Speaking of that: Are we already making sure to not include rooms that are disabled in the room list etc?

indico/modules/rb_new/client/js/components/RoomListFilters.jsx
```diff
@@ -0,0 +1,182 @@
...
+
+            if (filter === 'text') {
+                textValue = value;
+            } else {
+                if (building) {
+                    textValue = `building:${building}`;
+                }
+
+                if (floor) {
+                    textValue += ` floor:${floor}`;
```
ThiefMaster: let's build an array and `.join(' ')` it later?

ThiefMaster: also, what about a mapping for the criteria and iterating over that? something similar to this maybe:

```javascript
const stateToKeys = {building: 'building', floor: floor};
const searchParts = Object.entries(stateToKeys).map(([stateKey, searchKey]) => {
    // TODO: skip entries with no value (maybe a `filter()` before the `map()`
    return `${searchKey}:${this.state[stateKey]}`;
});
searchParts.push(text);  // the remaining searc text that's not a `foo:aar` option
seachParts.join(' ');
```

indico/modules/rb_new/client/js/components/RoomListFilters.jsx
```diff
@@ -0,0 +1,182 @@
...
+            selection: true,
+            style: {width: 200}
+        };
+
+        const {buildings: {list: buildingsList, isFetching}} = this.props;
+        const {building, floor} = this.state;
+        let floors = [{text: '', value: ''}];
+
+        if (building && buildingsList[building]) {
+            floors = floors.concat(buildingsList[building].floors.map((floorItem) => ({
```
ThiefMaster: I'd assign this to `buildingFloors` and later use `[...floors, ...buildingFloors]` to merge the two

indico/modules/rb_new/client/js/components/pages/RoomList.jsx
```diff
@@ -36,28 +36,32 @@ export default class RoomList extends React.Component {
...
+                                <Singular>
+                                    <Param name="count" value={list.length} /> result found
+                                </Singular>
+                                <Plural>
+                                    <Param name="count" value={list.length} /> results found
+                                </Plural>
+                            </PluralTranslate>
+                        </div>
+                        <Grid columns={5}>
+                            {list && list.map((room) => (
```
ThiefMaster: remove the `list &&` - calling `map()` on an empty array is fine

indico/modules/rb_new/client/js/util.js
```diff
@@ -21,3 +21,32 @@ import moment from 'moment';
...
 }
+
+export function parseRoomListFiltersText(text) {
+    const result = {text: null, building: null, floor: null};
+    if (!text) {
+        return result;
+    }
+
+    const reSplit = /((?:building:|floor:)(?:.*?)(?: |$))/;
+    const splitted = text.split(reSplit);
```
ThiefMaster: "splitted" is *very* uncommon or even wrong/outdated in English (it's irregular - split/split/split).
I'd call it `parts` or something like that.

indico/modules/rb_new/client/js/util.js
```diff
@@ -21,3 +21,32 @@ import moment from 'moment';
...
     return dt ? moment(dt, format) : null;
 }
+
+export function parseRoomListFiltersText(text) {
+    const result = {text: null, building: null, floor: null};
+    if (!text) {
+        return result;
+    }
+
+    const reSplit = /((?:building:|floor:)(?:.*?)(?: |$))/;
```
ThiefMaster: shouldn't we do something more generic here? ie split in space-separated tokens, and then see if it's matching `foo:bar` - if not it's just a normal part of the search string, otherwise you add it as `[key]: value` to the result dict

indico/modules/rb_new/controllers/backend.py
```diff
@@ -26,6 +26,7 @@
 from indico.modules.rb_new.schemas import aspect_schema, rooms_schema
 from indico.modules.rb_new.util import search_for_rooms
 from indico.util.string import natural_sort_key
+from indico.web.util import jsonify
```
ThiefMaster: that's not where `jsonify` comes from. it's `from flask import jsonify` (it only works because that module imports it from flask)

indico/modules/rb_new/util.py
```diff
@@ -30,6 +30,12 @@ def search_for_rooms(filters, only_available=False):
     if 'room_name' in filters:
         query = query.filter(Room.name.ilike('%{}%'.format(filters['room_name'])))
 
+    if 'building' in filters:
```
ThiefMaster: you could build a dict with all the simple filters and just do `query = query.filter_by(**whatever)`

indico/modules/rb/models/locations.py
mic4ael: We check against `is_active` property of a room

indico/modules/rb_new/client/js/components/RoomListFilters.jsx
```diff
@@ -0,0 +1,185 @@
...
+        this.setState({filtersVisible: visible});
+    }
+
+    applyFilters() {
+        const {onConfirm} = this.props;
+        this.setState({filtersVisible: false});
+        onConfirm();
+    }
+
+    handleFiltersChange(filter, value) {
```
ThiefMaster: `updateFilter`?

ThiefMaster: The function looks a bit convoluted. What about separating it into `updateTextFilter(text)` and `updateFilter(filter, value)`.
Parsing the string is only be needed in `updateTextFilter`, and updating the text  only in `updateFilter`.

indico/modules/rb_new/client/js/util.js
```diff
@@ -21,3 +21,26 @@ import moment from 'moment';
...
     return dt ? moment(dt, format) : null;
 }
+
+export function parseRoomListFiltersText(text) {
+    const result = {text: null, building: null, floor: null};
+    if (!text) {
+        return result;
+    }
+
+    const parts = text.split(" ");
```
ThiefMaster: single quotes :p

indico/modules/rb_new/util.py
```diff
@@ -49,3 +60,23 @@ def search_for_rooms(filters, only_available=False):
     else:
         rooms = query.all()
     return rooms
+
+
+def get_buildings(location=None):
+    buildings = defaultdict(dict)
+    rooms = location.rooms if location else Room.find_all()
```
ThiefMaster: `Room.query.filter_by(is_active=True).all()` (let's try to avoid adding more `Foo.find*()` calls, we kind of deprecated them quite some time ago_

indico/modules/rb/models/locations.py
```diff
@@ -162,18 +161,5 @@ def get_equipment_by_name(self, name):
...
-            room_with_lat_lon = next((r for r in rooms if r.longitude and r.latitude), None)
-            if not room_with_lat_lon:
-                continue
-            buildings.append({'number': building_name,
-                              'title': _(u'Building {}'.format(building_name)),
-                              'longitude': room_with_lat_lon.longitude,
-                              'latitude': room_with_lat_lon.latitude,
-                              'rooms': [r.to_serializable('__public_exhaustive__') for r in rooms]})
-        return buildings
+        from indico.modules.rb_new.util import get_buildings
```
ThiefMaster: let's keep this one and the tests as they were before - we can just remove them later when removing the old RB stuff

indico/modules/rb_new/util.py
```diff
@@ -49,3 +60,23 @@ def search_for_rooms(filters, only_available=False):
     else:
         rooms = query.all()
     return rooms
+
+
+def get_buildings(location=None):
```
ThiefMaster: location arg can go away when reverting the change to the legacy get_buildings in the model

indico/modules/rb_new/util.py
mic4ael: Good to know ;)

indico/modules/rb_new/client/js/components/RoomListFilters.jsx
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        const buildings = Object.entries(buildingsList).map(([key, val]) => ({
+            text: <Translate>Building <Param name="buildingNumber" value={val.number} /></Translate>,
+            value: key
+        }));
+        const content = (
+            <Form>
+                <Form.Field>
+                    <Select {...commonAttrs} placeholder="Select building"
```
mvidalgarcia: Translate.string('foo')

indico/modules/rb_new/client/js/components/RoomListFilters.jsx
```diff
@@ -0,0 +1,185 @@
...
+                            loading={isFetching}
+                            onChange={(event, data) => this.handleFiltersChange('building', data.value)}
+                            value={building}
+                            options={[{text: '', value: ''}, ...buildings]} />
+                </Form.Field>
+                <Form.Field>
+                    <Select {...commonAttrs} placeholder="Select floor" disabled={!building}
```
mvidalgarcia: Also translate

ThiefMaster:
applying filters always triggers another search request, even if nothing changed

indico/modules/rb_new/client/js/components/pages/RoomList.module.scss
```diff
@@ -15,39 +15,16 @@
...
-
-        .available-equipment {
-            text-align: right;
-        }
-
-        .room-card-description {
-            white-space: nowrap;
-            text-overflow: ellipsis;
-            overflow: hidden;
+        .room-image {
```
mvidalgarcia: Not sure, but I think this is unused.

indico/modules/rb_new/controllers/backend.py
```diff
@@ -34,7 +35,9 @@
     'start_dt': fields.DateTime(),
```
mvidalgarcia: I know it's not part of this PR, but maybe calling this dict `room_args` or `room_request_args`? `request_args` sounds rather generic.

ThiefMaster: Or just move it inside:

```python

class RHRoomBookingSearch(RHRoomBookingBase):
    @use_args({
        'capacity': fields.Int(),
        'room_name': fields.Str(),
        'start_dt': fields.DateTime(),
        'end_dt': fields.DateTime(),
        'repeat_frequency': EnumField(RepeatFrequency),
        'repeat_interval': fields.Int(missing=0),
        'building': fields.Str(),
        'floor': fields.Str()
    })
    def _process(self, args):
        rooms = sorted(search_for_rooms(args), key=lambda r: natural_sort_key(r.full_name))
        return rooms_schema.dumps(rooms).data
```

PR #3325, opened 2018-04-25
Allow using dl, dt, dd HTML tags in editors

mvidalgarcia:


indico/util/string.py
```diff
@@ -45,14 +45,14 @@
 # basic list of tags, used for markdown content
 BLEACH_ALLOWED_TAGS = bleach.ALLOWED_TAGS + [
     'sup', 'sub', 'small', 'br', 'p', 'table', 'thead', 'tbody', 'th', 'tr', 'td', 'img', 'hr', 'h1', 'h2', 'h3', 'h4',
-    'h5', 'h6', 'pre'
+    'h5', 'h6', 'pre', 'dl', 'dd', 'dt'
```
ThiefMaster: I don't think this change is necessary. For places where CKEditor fields are used, we use `BLEACH_ALLOWED_TAGS_HTML`

mvidalgarcia: `render_markdown`, which is being used in the contribution display page to print the description, uses `BLEACH_ALLOWED_TAGS`. This was the main reason of the change `(RQF1001171)`.

ThiefMaster: Ah I didn't realize it was there. I thought it was in the event description.

PR #3333, opened 2018-04-30
BOA latex pdf: Use single fancyhdr invocation

bpedersen2:
As the default template does not render contribution
specific content in the header/footer part, just invoke
the fancyhdr setup macros once before the list of abstracts.


ThiefMaster:
what's the advantage of this change?

bpedersen2:
The main adavantage is better readiblity (and probably performance) of the generated tex input. 
There is no technicall need to reset the header configuration on each contribution as it is not changing.

ThiefMaster:
This one would need to be adapted to the new templates as well :/

bpedersen2:
Updated to 2.2/master templates

PR #3335, opened 2018-04-30
BOA: Add some more formatting options

bpedersen2:
Add the following options:
 - Post-face text
 - Min. length per abstract


indico/modules/events/abstracts/forms.py
```diff
@@ -103,6 +103,8 @@ class BOASettingsForm(IndicoForm):
                                                  enum=BOACorrespondingAuthorType, sorted=True)
     show_abstract_ids = BooleanField(_('Show abstract IDs'), widget=SwitchWidget(),
                                      description=_("Show abstract IDs in the table of contents."))
+    min_lines_per_abstract = IntegerField(_("Minimum lines per abstract",
+                                          description=_("Minimum lines to reserver per abstracts.")))
```
mvidalgarcia: > reserve~s~ per abstract~s~

Issue #3337, opened 2018-05-04
Add event date/time in registration confirmation mail


pferreir:
This would probably not hurt and would save people the trouble of going back to the web browser and looking for the event.

PR #3339, opened 2018-05-08
Add event date/time to registration emails

bpedersen2:
Fixes: #3337

### DIFF
--- a/indico/modules/events/registration/templates/emails/base_registration.html
+++ b/indico/modules/events/registration/templates/emails/base_registration.html
@@ -1,4 +1,6 @@
 {% extends 'emails/base.html' %}
+{% from 'events/emails/_event_dates.html' import render_event_dates_html %}
+
 {% block header %}{% endblock %}
 {% block footer %}{% endblock %}
 
@@ -17,6 +19,9 @@
             <h1 style="margin: 0;">
                 {{ event.title }}
             </h1>
+            <div>
+                {{- render_event_dates_html(event) -}}
+            </div>
             <h3 style="margin: 0; margin-top: 10px; padding-top: 10px; border-top: 1px solid #eee;">
                 {{ registration.registration_form.title }}
                 <span style="float: right; font-style: italic;">
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -73,6 +73,8 @@ Improvements
   a meeting theme (:issue:`3297`, thanks :user:`bpedersen2`)
 - Allow definition lists in places where Markdown or HTML is accepted
   (:issue:`3325`)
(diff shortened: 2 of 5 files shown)


indico/modules/events/templates/emails/render_event_dates.html
```diff
@@ -0,0 +1,60 @@
+{% macro render_event_dates(event, occurences=None) %}
```
ThiefMaster: lowercase `none`

indico/modules/events/templates/emails/render_event_dates.html
```diff
@@ -0,0 +1,60 @@
...
+                {{ evt.start_dt|format_date(timezone=evt.timezone) }}, {# -#}
+                {{ evt.start_dt|format_time(timezone=evt.timezone) }} - {# -#}
+                {{ evt.end_dt|format_time(timezone=evt.timezone) }}: {# -#}
+                {{ evt.external_url -}}
+            {%- endfor -%}
+        {%- endif -%}
+    {%- endfilter -%}
+{% endmacro %}
+
+{% macro render_event_dates_html(event, occurences=None) %}
```
ThiefMaster: lowercase `none`

indico/modules/events/registration/templates/emails/base_registration.html
```diff
@@ -17,6 +18,9 @@
             <h1 style="margin: 0;">
                 {{ event.title }}
             </h1>
+            <div>
+            {{- render_event_dates_html(event) -}}
```
ThiefMaster: indent by 4 more spaces

indico/modules/events/registration/templates/emails/base_registration.html
```diff
@@ -1,4 +1,5 @@
 {% extends 'emails/base.html' %}
+{% from 'events/emails/render_event_dates.html' import render_event_dates_html %}
```
ThiefMaster: Usually we use an underscore-prefixed name for templates containing only macros, e.g. `_event_dates.html`

ThiefMaster:
I just tested it and it breaks the formatting of both the plaintext notification email for new events and the registration HTML emails (the header is rather compact, and adding big headings there doesn't look great).

Also I realized that having a macro that's meant to be used in a plaintext context in a .html file may cause problems: Jinja does autoescaping for HTML when the template is a .html file, but not if it's a .txt file. So calling this macro from the .txt file would probably result in stuff like `<` being escaped (and thus looking ugly in a text email).

I've added a simplified version to the registration confirmation email (like the original issue suggested anyway): 32ef3f28ee

Issue #3351, opened 2018-05-09
Scientific Programme PDF - Markdown not properly rendered


pferreir:
See for instance: https://indico.cern.ch/event/299118/program.pdf (page 2)

![image](https://user-images.githubusercontent.com/2699/39809645-b0622800-5382-11e8-8071-c8f334760135.png)

mic4ael:
The references PR partially fixes the issue. By partially I mean that reportlab can't handle HTML tags like `ul, ol` etc. which makes things a bit more complicated. A possible solution would be to implement a parser which would convert markdown/html to something that reportlab understands and resembles what one would get as a result using HTML.

culpinnis:
Hi together, 
we still have problems with the PDF rendering from markdown. Texts in bold are displayed correctly on the website, but in the PDF there is just `**Text**`, for example at https://events.ipb-halle.de/event/4/program.pdf
Indico version is 2.1.7. 
Best Wishes
Chris

PR #3352, opened 2018-05-09
Partially fix bug with the program PDF export

mic4ael:
#3351


ThiefMaster:
Could we at least keep linebreaks between the list items? I think losing all separation between the list elements is worse than not rendering any markdown since it results in something rather unreadable.

mic4ael:
With the latest update the result pdf looks like this

screenshot from 2018-05-15 08-39-53

indico/legacy/pdfinterface/conference.py
```diff
@@ -139,8 +139,13 @@ def getBody(self, story=None):
         styles = _get_sans_style_sheet()
         style = styles["Normal"]
         style.alignment = TA_JUSTIFY
-        p = Paragraph(escape(track_settings.get(self.event, 'program').encode('utf-8')), style)
-        story.append(p)
+
+        event_program = render_markdown(track_settings.get(self.event, 'program'))
+        for para in event_program.split('\n'):
+            if para.startswith('<p'):
```
ThiefMaster: what about regex matching `<p[ >]` to avoid e.g. `<pre>` from matching here as well?

mic4ael: yeah, good point

PR #3370, opened 2018-05-23
Add (friendly) registration id to designer fields

bpedersen2:
This sometimes to allow to easily identify users during (manual) check-in.


indico/modules/designer/placeholders.py
```diff
@@ -268,6 +268,12 @@ def render(cls, registration):
         return format_currency(registration.price, registration.currency, locale='en_GB')
 
 
+class RegistrationIdPlaceholder(RegistrationPlaceholder):
+    name = 'registrationid'
```
ThiefMaster: `registration_id` or `registration_friendly_id` (if you go for the latter, also update the class name)

indico/modules/designer/placeholders.py
```diff
@@ -268,6 +268,12 @@ def render(cls, registration):
         return format_currency(registration.price, registration.currency, locale='en_GB')
 
 
+class RegistrationIdPlaceholder(RegistrationPlaceholder):
```
ThiefMaster: `ID` instead of `Id`

bpedersen2: Done

indico/modules/designer/placeholders.py
bpedersen2: Done

PR #3395, opened 2018-06-07
Add working Room Booking dialog

pferreir:


indico/modules/rb_new/client/js/serializers/common.js
```diff
@@ -0,0 +1,59 @@
...
+ * Indico is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+export const filterDtHandler = (prefix) => {
```
ThiefMaster: s/Dt/DT/?

indico/modules/rb_new/controllers/backend.py
```diff
@@ -151,3 +155,43 @@ def _process_DELETE(self):
...
+    def _process(self, args):
+        room = Room.get_one(args.pop('room_id'))
+        user_id = args.pop('user_id', None)
+        booked_for = User.get_one(user_id) if user_id else session.user
+        is_prebooking = args.pop('is_prebooking')
+
+        # Check that the booking is not longer than allowed
+        booking_limit_days = room.booking_limit_days or rb_settings.get('booking_limit')
+        if not self._validate_room_booking_limit(args['start_dt'], args['end_dt'], booking_limit_days):
+            msg = (_(u'Bookings for the room "{}" may not be longer than {} days')
```
ThiefMaster: no need for the u prefix; we're using `unicode_literals`

indico/modules/users/controllers.py
```diff
@@ -517,3 +520,17 @@ def _process(self):
...
+
+class RHUserSearch(RHProtected):
+    """Search for users based on given criteria"""
+
+    @use_args({
+        'email': fields.Str(validate=lambda s: len(s) > 3 and '@' in s),
+        'name': fields.Str(validate=validate.Length(min=3))
+    })
+    def _process(self, args):
+        if not args:
```
ThiefMaster: i guess we can't do that on the webargs level?

indico/web/client/js/react/components/UserSearchField.jsx
```diff
@@ -0,0 +1,119 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Dropdown, Label} from 'semantic-ui-react';
+
+import searchUsersUrl from 'indico-url:users.user_search';
```
ThiefMaster: URL

indico/modules/rb_new/client/js/actions.js
```diff
@@ -362,3 +370,36 @@ export function fetchTimelineData() {
...
+export function createBooking(args) {
+    return async (dispatch) => {
+        let response;
+        const params = preProcessParameters(args, ajaxBookingRules);
+        dispatch({type: BOOKING_ONGOING});
+
+        try {
+            response = await indicoAxios.post(createBookingURL(), params);
+        } catch (error) {
+            handleAxiosError(error);
```
ThiefMaster: Have a look at [createSubmissionError](https://github.com/indico/indico/pull/3391/files#diff-67b3b116508a9dc437d4535a38add0c7R22); like that you can mark form fields as erroneous based on the error from webargs. Probably needs some adjustments to use with an ajax request from a redux action

indico/modules/rb_new/client/js/components/modals/BookRoomModal.jsx
```diff
@@ -0,0 +1,306 @@
...
+
+function validate({usage, user, reason}) {
+    const errors = {};
+    if (!usage) {
+        errors.usage = Translate.string('Please choose an option!');
+    }
+    if (usage === 'someone' && !user) {
+        errors.user = Translate.string('Please specify a user');
+    }
+    if (!reason) {
```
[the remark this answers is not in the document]
pferreir: We could have some kind of "validator factory" that generates a validation function based on a declarative definition. Pretty much in the way we're doing with AJAX/querystring serializers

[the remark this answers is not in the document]
ThiefMaster: Trimming should probably be done on a different level, not validators (so we actually get rid of the whitespace before even sending it). 

Should do the same thing in webargs nonetheless; opened an issue there to see if there's a clean way to do it: sloria/webargs#234

indico/modules/rb_new/client/js/actions.js
pferreir: Do I need that, though? I have already a client-side validator. If the user manages to mess it up, it's their fault that they'll see an error dialog.

ThiefMaster: In that case probably not - but I think it'd be nice to catch such errors nicely. Could have a generic util that takes care of it at some point. Not super important right now though

indico/modules/rb_new/client/js/components/modals/BookRoomModal.jsx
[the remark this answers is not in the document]
pferreir: Maybe with [normalize](https://redux-form.com/7.3.0/docs/api/field.md/#-code-normalize-value-previousvalue-allvalues-previousallvalues-gt-nextvalue-code-optional-)?

indico/modules/users/controllers.py
pferreir: https://github.com/sloria/webargs/issues/113

indico/modules/rb_new/client/js/components/modals/BookRoomModal.jsx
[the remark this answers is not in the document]
pferreir: > Maybe with normalize?

Yeah, that works 😃

indico/web/client/js/react/components/UserSearchField.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+    handleSearchChange = _.debounce(async (__, {searchQuery}) => {
+        if (searchQuery.trim().length < 3) {
+            return;
+        }
+
+        this.setState({
+            isFetching: true
+        });
+        const query = (searchQuery.indexOf('@') !== -1)
```
mic4ael: maybe it is better to use `includes` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

PR #3396, opened 2018-06-08
Time range picker

meluru:
improve date picker as well

### DIFF
--- a/indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
+++ b/indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
@@ -48,8 +48,8 @@ export default class TimeRangePicker extends React.Component {
     };
 
     static defaultProps = {
-        startTime: null,
-        endTime: null,
+        startTime: moment().startOf('hour').add(1, 'h'),
+        endTime: moment().startOf('hour').add(2, 'h')
     };
 
     constructor(props) {
@@ -74,27 +74,44 @@ export default class TimeRangePicker extends React.Component {
         });
         if (!found) {
             options.push({key: value, value, text: value});
+            return true;
         }
+        return false;
     }
 
     setStartTime(startTime, endTime) {
-        const start = toMoment(startTime, 'HH:mm');
+        const start = moment(startTime, ['HH:mm', 'HHmm', 'Hmm']);
         const end = toMoment(endTime, 'HH:mm');
-        this.setState({
-            startTime: start
-        });
-        const {onChange} = this.props;
-        onChange(start, end);
+        if (start.isValid()) {
+            const options = generateTimeOptions();
+            const added = this.addOptionIfNotExist(options, _serializeTime(moment(start)));
+            if (added) {
+                this.updateOptions(options, true);
+            }
+            this.setState({
+                startTime: start
+            });
+            const {onChange} = this.props;
+            onChange(start, end);
+        }
     }
+
 
     setEndTime(startTime, endTime) {
         const start = toMoment(startTime, 'HH:mm');
-        const end = toMoment(endTime, 'HH:mm');
-        this.setState({
-            endTime: end
-        });
-        const {onChange} = this.props;
-        onChange(start, end);
+        const end = moment(endTime, ['HH:mm', 'HHmm', 'Hmm']);
+        if (end.isValid()) {
+            const options = generateTimeOptions();
+            const added = this.addOptionIfNotExist(options, _serializeTime(moment(end)));
+            if (added) {
+                this.updateOptions(options, false);
+            }
+            this.setState({
+                endTime: end
+            });
+            const {onChange} = this.props;
+            onChange(start, end);
+        }
     }
 
     updateOptions(options, start) {
@@ -113,39 +130,27 @@ export default class TimeRangePicker extends React.Component {
         return [];
     };
 
-    itemAdded = (data, start) => {
-        const value = data['value'];
-        const options = generateTimeOptions();
-        const isValid = moment(value, 'HH:mm', true).isValid();
-        if (isValid) {
-            this.addOptionIfNotExist(options, value);
-        }
-        this.updateOptions(options, start);
-    };
-
     render() {
         const {startTime, endTime, startOptions, endOptions} = this.state;
         return (
             <div>
                 <Dropdown options={startOptions}
-                          onAddItem={(e, data) => this.itemAdded(data, true)}
                           search={(list, query) => this.searchTime(list, query)}
                           selection
                           allowAdditions
                           additionLabel=""
                           styleName="time-dropdown"
-                          defaultValue={_serializeTime(startTime)}
+                          value={_serializeTime(startTime)}
                           onChange={(_, {value}) => {
                               this.setStartTime(value, endTime);
                           }} />
                 <Dropdown options={endOptions}
-                          onAddItem={(e, data) => this.itemAdded(data, false)}
                           search={(list, query) => this.searchTime(list, query, false)}
                           selection
                           allowAdditions
                           additionLabel=""
                           styleName="time-dropdown"
-                          defaultValue={_serializeTime(endTime)}
+                          value={_serializeTime(endTime)}
                           onChange={(_, {value}) => {
                               this.setEndTime(startTime, value);
                           }} />
--- a/indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
+++ b/indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
@@ -1,0 +1,148 @@
+/* This file is part of Indico.
+ * Copyright (C) 2002 - 2018 European Organization for Nuclear Research (CERN).
+ *
+ * Indico is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Indico is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -0,0 +1,203 @@
...
+
+function _humanizeDuration(duration) {
+    const hours = duration.hours();
+    const minutes = duration.minutes();
+    if (hours === 1 && minutes === 0) {
+        return `${hours} hour`;
+    } else if (hours !== 0) {
+        const fraction = minutes / 60;
+
+        return `${hours + fraction} hours`;
```
mic4ael: You can limit it to only one line ``return `${hours + (minutes / 60)} hours`;``

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -0,0 +1,203 @@
...
+
+    generateEndTimeOptions = (start) => {
+        const options = [];
+        const end = moment().endOf('day');
+        const next = moment(start).add(30, 'm');
+        let serializedNext, duration, text;
+        while (next < end) {
+            duration = _humanizeDuration(moment.duration(next.diff(start)));
+            serializedNext = _serializeTime(moment(next));
+            text = <span>{serializedNext} <span styleName="duration">({duration })</span></span>;
```
mic4ael: extra space in `{duration }`

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -0,0 +1,203 @@
...
+    const hours = duration.hours();
+    const minutes = duration.minutes();
+    if (hours === 1 && minutes === 0) {
+        return `${hours} hour`;
+    } else if (hours !== 0) {
+        const fraction = minutes / 60;
+
+        return `${hours + fraction} hours`;
+    } else {
+        return `${minutes} minutes`;
```
mic4ael: i guess it should be a translated string (like the others above)

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -0,0 +1,203 @@
...
+            duration = _humanizeDuration(moment.duration(next.diff(start)));
+            serializedNext = _serializeTime(moment(next));
+            text = <span>{serializedNext} <span styleName="duration">({duration })</span></span>;
+            options.push({key: serializedNext, value: serializedNext, text});
+            next.add(30, 'm');
+        }
+        return options;
+    };
+
+    addOptionIfNotExist = (options, value) => {
```
mic4ael: maybe a better name would be `addOptionIfMissing`?

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -0,0 +1,203 @@
...
+                duration,
+                startOptions,
+                endOptions,
+            });
+            const {onChange} = this.props;
+            onChange(start, end);
+        }
+    };
+
+    searchTime = () => {
```
mic4ael: what is the point of this function when passed to `Dropdown`?

meluru: It overrides the default behaviour of search, so when you start typing you get no options

mic4ael: yeah, yeah, but isn't it enough to not include the `search` attribute in the first place? or at least do it as `search={true}`. I don't understand why you pass the `search` prop even if you don't need the searching functionality

meluru: So basically including search property let's you type in the field and it's required by addition properties. Keeping the default behavior of search would try to match current options with what you type, so if you started typing e.g. '1' you'd get all the time options that include 1 which doesn't make much sense in a time picker. That's why I need this method to make sure that whatever you type, you get no options as a result of search.

mic4ael: Ahh, that is what I didn;t understand - `search` being required by some other properties. However you can make it an inline arrow function `search={() => []}`

meluru: that's a good idea :)

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
meluru: Are you sure I need parentheses here?

mic4ael: actually you don't thanks to the operator precedence but I think it is usually clearer that way (at least to me)

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -0,0 +1,198 @@
...
+    } else if (hours !== 0) {
+        return Translate.string('{time} hours', {time: hours + (minutes / 60)});
+    } else {
+        return Translate.string('{time} minutes', {time: minutes});
+    }
+}
+
+export default class TimeRangePicker extends React.Component {
+    static propTypes = {
+        startTime: propTypes.object,
```
mvidalgarcia: The way to go pointed out by the [`prop-types`](https://www.npmjs.com/package/prop-types) lib is importing it capitalizing the first letter i.e `PropTypes`. I know we have it imported as `propTypes` in other places of our code base but we should choose one way and be consistent all over.

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -0,0 +1,198 @@
...
+import React from 'react';
+import propTypes from 'prop-types';
+import {Dropdown} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+import {toMoment} from '../util';
+
+import './TimeRangePicker.module.scss';
+
+
+const _serializeTime = time => (time ? time.format('HH:mm') : null);
```
mvidalgarcia: I've noticed we have the exact same function also in Landing.jsx and TimeForm.jsx, what about moving it to util and reusing it?

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
ThiefMaster: yeah, in such a case I prefer parentheses or not putting space around `/` if eslint allows that (PEP8 does, not sure about eslint)

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -0,0 +1,198 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import moment from 'moment';
+import React from 'react';
+import propTypes from 'prop-types';
```
ThiefMaster: Let's use the correct style in new files as @mvidalgarcia mentioned below. Existing files... either wait for the eslint-based fixer or update when changing something related.

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -0,0 +1,198 @@
...
+        });
+        if (!found) {
+            options.unshift({key: value, value, text: value});
+            return true;
+        }
+        return false;
+    };
+
+    updateStartTime = (startTime, endTime, duration) => {
+        const start = moment(startTime, ['HH:mm', 'HHmm', 'Hmm']);
```
ThiefMaster: should we except `HH` as well? so writing just `13` would turn into `13:00`

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -0,0 +1,198 @@
...
+        let end = toMoment(endTime, 'HH:mm');
+        if (start.isValid()) {
```
ThiefMaster:
```
if (!start.isValid()) {
    return;
}
```

indico/modules/rb_new/client/js/components/pages/Landing.jsx
```diff
@@ -19,21 +19,22 @@ import moment from 'moment';
 import PropTypes from 'prop-types';
 import React from 'react';
 import {Link} from 'react-router-dom';
-import Calendar from 'rc-calendar';
+import RcCalendar from 'rc-calendar';
```
ThiefMaster: `RCCalendar`?

indico/modules/rb_new/client/js/components/pages/Landing.jsx
```diff
@@ -204,11 +212,9 @@ export default class Landing extends React.Component {
...
                                 )}
                                 <Form.Group inline>
-                                    <TimePicker {...timePickerProps} value={startTime}
-                                                onChange={(value) => this.updateTimes(value, endTime)} />
-                                    -
-                                    <TimePicker {...timePickerProps} value={endTime}
-                                                onChange={(value) => this.updateTimes(startTime, value)} />
+                                    <TimeRangePicker startTime={startTime}
+                                                     endTime={endTime}
+                                                     onChange={(start, end) => this.updateTimes(start, end)} />
```
ThiefMaster: `onChange={this.updateTimes}`

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
meluru: The first formatting also accepts HH

PR #3405, opened 2018-06-11
add prenominal title Eur Ing

fgeorgatos:
ref. https://en.wikipedia.org/wiki/European_Engineer

### DIFF
--- a/indico/modules/users/models/users.py
+++ b/indico/modules/users/models/users.py
@@ -42,13 +42,14 @@ from indico.util.struct.enum import RichIntEnum
 
 
 class UserTitle(RichIntEnum):
-    __titles__ = ('', _('Mr'), _('Ms'), _('Mrs'), _('Dr'), _('Prof.'))
+    __titles__ = ('', _('Mr'), _('Ms'), _('Mrs'), _('Dr'), _('Prof.'), _('Eur Ing'))
     none = 0
     mr = 1
     ms = 2
     mrs = 3
     dr = 4
     prof = 5
+    euring = 6
 
 
 class NameFormat(RichIntEnum):


ThiefMaster:
Not sure if it's a good idea to add more titles than the basic ones... what about all the other ones some people might use?

M.Sc / B.Sc / Dr. rer. nat,, / Dipl.-Ing. / etc. - all titles that some people probably use, but are currently not present in Indico.

---

This aside, adding a new title is not just adding it to the enum, it'd also require creating a new alembic revision to update CHECK constraints on all database columns referencing a title and possibly adding it in some legacy JS which does not make use of the enum.

pferreir:
Can we somehow make this config-specific?

ThiefMaster:
not with the current models (where we're just storing the numeric enum value)... i18n would be tricky as well (except when letting people provide various language versions in the config)

fgeorgatos:
@ThiefMaster : 
you've already opened that can of worms, by assuming that `Prof.` is part of the `basic` titles... well!

one line of thinking could be to align with this (or any other passport standard for that matter): https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/118548/titles-included-in-passports.pdf

Regardless of if this one PR gets merged in, handled or not, I take the chance to ask you this question:
* Could there perhaps be a mechanism that allows to verify that someone is really entitled to that title (pun intended). I mean, some means to cross-check that people are what they say they are?
any ideas welcome...

ThiefMaster:
Since this PR is incomplete for the reasons mentioned above, maybe it's better to continue the discussion on supporting additional titles either in a normal issue or a [forum thread](https://talk.getindico.io/)?

> Could there perhaps be a mechanism that allows to verify that someone is really entitled to that title (pun intended). I mean, some means to cross-check that people are what they say they are?

I don't see any way how this could be done - and honestly, I don't think this is done anywhere, except maybe in cases where you get benefits by using a title (e.g. applying for a job requiring you to have a PhD). I'm pretty sure I could book an airplane ticket as "His Majesty" or "Prof. Dr." without having any of these titles ;)

PR #3407, opened 2018-06-11
Fix image dim issue

meluru:


indico/modules/rb_new/client/js/components/Dimmer.module.scss
```diff
@@ -0,0 +1,57 @@
...
+
+    &:hover .hover-content {
+        opacity: 1;
+    }
+
+    &:hover .content {
+        opacity: 0;
+    }
+}
+
```
mic4ael: Too many blank lines ;)

indico/modules/rb_new/client/js/components/Room.jsx
```diff
@@ -39,68 +40,30 @@ export default class Room extends React.Component {
...
-
-    shouldComponentUpdate(nextProps, nextState) {
+    shouldComponentUpdate(nextProps) {
         const {isFavorite: nextIsFavorite, room: nextRoom} = nextProps;
         const {room, isFavorite} = this.props;
-        const {dimmed} = this.state;
-        const {dimmed: nextDimmed} = nextState;
 
-        return nextIsFavorite !== isFavorite || !_.isEqual(room, nextRoom) || dimmed !== nextDimmed;
+        return nextIsFavorite !== isFavorite || !_.isEqual(room, nextRoom);
```
mic4ael: I am sure this will cause "wasted renders". Maybe it would be better to keep it somehow?

meluru: I removed the variable that we don't use anymore, did you think I removed the whole line?

PR #3419, opened 2018-06-14
Set proper day when managing session timetable

mic4ael:
Related request: `RQF1037887`


ThiefMaster:
probably needs a timezone conversion. try it with an itemized where the UTC time is on a different date in local time.

mic4ael:
Updated

PR #3429, opened 2018-06-26
Fix contribution pdf sorting

bpedersen2:
Only fall back to title if sort_by is empty after mapping.
Otherwise e.g. Contribution PDF export sorted by boardnumber
does not sort by board_number.

### DIFF
--- a/indico/legacy/pdfinterface/conference.py
+++ b/indico/legacy/pdfinterface/conference.py
@@ -365,6 +365,15 @@ class ContributionBook(PDFLaTeXBase):
                 if not c.speakers:
                     return True, None
                 return False, speakers[0].get_full_name(last_name_upper=False, abbrev_first_name=False).lower()
+        elif sort_by == BOASortField.board_number:
+            key_func = attrgetter('board_number')
+        elif sort_by == BOASortField.session_board_number:
+            key_func = lambda c: (c.session is None, c.session.title.lower() if c.session else '', c.board_number)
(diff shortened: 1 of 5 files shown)


indico/legacy/pdfinterface/conference.py
```diff
@@ -366,7 +366,7 @@ def key_func(c):
                     return True, None
                 return False, speakers[0].get_full_name(last_name_upper=False, abbrev_first_name=False).lower()
         else:
-            key_func = attrgetter(mapping.get(sort_by) or 'title')
+            key_func = attrgetter(mapping.get(sort_by) or sort_by or 'title')
```
ThiefMaster: `sort_by` is supposed to be an enum value (from `BOASortField`), and `attrgetter()` won't work with such a value.

`mapping.get(sort_by, sort_by.value) or 'title'` seems like the safer choice to also allow the special case of passing a string there (which is done only when called from `RHContributionsExportPDFBookSorted`). Or maybe we should just add `board_number` as a new option to the enum?

ThiefMaster: Actually, I wonder for which cases we added `mapping = {'number': 'id', 'name': 'title'}`... maybe legacy code back then that was passing strings instead of enum values.

bpedersen2: Reworked to add more enums. This also adds the possibility to sort by session or schedule and boardnumber  inside the sessions.

And from a fast git grep it really seems like the mapping part is actually unused now.

PR #3432, opened 2018-06-26
Create logo image for TeX in tex-gen subdir

bpedersen2:
This will make a later zipping of this folder easier.

### DIFF
--- a/indico/modules/events/util.py
+++ b/indico/modules/events/util.py
@@ -415,11 +415,11 @@ def get_base_ical_parameters(user, detail, path, params=None):
             'user_logged': user is not None, 'request_urls': request_urls}
 
 
-def create_event_logo_tmp_file(event):
+def create_event_logo_tmp_file(event, tmppath=None):
     """Creates a temporary file with the event's logo"""
     logo_meta = event.logo_metadata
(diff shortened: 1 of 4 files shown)


indico/modules/events/util.py
```diff
@@ -415,11 +415,11 @@ def get_base_ical_parameters(user, detail, path, params=None):
             'user_logged': user is not None, 'request_urls': request_urls}
 
 
-def create_event_logo_tmp_file(event):
+def create_event_logo_tmp_file(event, dir=config.TEMP_DIR):
```
ThiefMaster: call it `path` to avoid shadowing the builtin `dir`

ThiefMaster: the config is not accessible at import time (i'm surprised this doesn't break!). Set it to `None` here and do `dir or config.TEMP_DIR` inside the function

bpedersen2: Done

bpedersen2: Done

bpedersen2: Most probably there are no calls without setting the  dir path in the call left.

ThiefMaster: But default arguments are evaluated at import time, so it should already fail without even calling the function (and in the tests it did).

Issue #3039, opened 2017-08-09
Editable book of abstracts


meluru:
The idea is to let event managers export the book of abstracts in LaTeX format, so they can edit it in a preferable way. 

It was also suggested by the user that Indico could provide some basic settings to edit the book of abstract (add title page, programme, organising committee, be able to edit abstract page) and have an overview of how the exported book of abstracts will look.

When uploading the file we could use the new `File` model (same one we use for files in the editing module) and just store the ID of that file in the `boa_settings` SettingsProxy. Alternatively, we could add a new `custom_boa_file_id` FK column referencing a File to the `Event` model.

ThiefMaster:
> and have an overview of how the exported book of abstracts will look.

Except creating one with dummy data (like we do in the notification email preview) I don't see any good way to do so.

pferreir:
We should also have an "upload custom BoA" option,  which serves an uploaded PDF instead of the default generated one.

bpedersen2:
Downloading should be possible if #3434 get's accepted

bpedersen2:
As for the custom BoA: We do this by:
1. disable the standard BoA link
2. Upload the file as material (the hidden folder option is useful here)
3. Link this via a custom page pointing to the url

giordand:
will it be in Release 2.2?

ThiefMaster:
@giordand Unfortunately not, but we'll try to get at least the import/export part in 2.3.

joshuabarrow221:
Hello,
I have used the "Book of Abstracts" feature on a Fermi National Accelerator Laboratory (FNAL) server to collect contributions from our participants. I cannot find the seemingly newly added feature to output a LaTeX document directly instead of only a PDF. If I advocate that FNAL update to the newly merged build (I see this seems to have happened two weeks ago), will I be able to do so?

ThiefMaster:
@joshuabarrow221 

Hi, version 2.3 is only deployed on indico.cern.ch at the moment. While anyone could in principle build an indico page based on master (2.3-dev), we do not recommend this since we just deployed a few days ago and still find some (minor) bugs. So we recommend waiting with the upgrade a bit longer until we made a "real" 2.3 release.

joshuabarrow221:
Thanks; I'll keep an eye out for that.

PR #3434, opened 2018-06-27
Tex export

bpedersen2:
Note that this includes #3432  and #3429 currently

### DIFF
--- a/indico/modules/events/contributions/blueprint.py
+++ b/indico/modules/events/contributions/blueprint.py
@@ -37,16 +37,14 @@ _bp.add_url_rule('/manage/contributions/contributions.xlsx', 'contributions_exce
                  management.RHContributionsExportExcel, methods=('POST',))
 _bp.add_url_rule('/manage/contributions/contributions.pdf', 'contributions_pdf_export',
                  management.RHContributionsExportPDF, methods=('POST',))
-_bp.add_url_rule('/manage/contributions/book.pdf', 'contributions_pdf_export_book',
-                 management.RHContributionsExportPDFBook, methods=('POST',))
-_bp.add_url_rule('/manage/contributions/book-sorted.pdf', 'contributions_pdf_export_book_sorted',
-                 management.RHContributionsExportPDFBookSorted, methods=('POST',))
-_bp.add_url_rule('/manage/contributions/book-sorted-schedule.pdf', 'contributions_pdf_export_book_sorted_schedule',
-                 management.RHContributionsExportPDFBookSortedSchedule, methods=('POST',))
-_bp.add_url_rule('/manage/contributions/book-sorted-session.pdf', 'contributions_pdf_export_book_sorted_session',
-                 management.RHContributionsExportPDFBookSortedSession, methods=('POST',))
-_bp.add_url_rule('/manage/contributions/book-sorted-all.pdf', 'contributions_pdf_export_book_sorted_all',
-                 management.RHContributionsExportPDFBookSortedAll, methods=('POST',))
+
+# LaTeX-based exports
+_bp.add_url_rule('/manage/contributions/text-export/<uuid>', 'contributions_tex_export_book',
+                 management.RHContributionsExportTeXBook)
+_bp.add_url_rule('/manage/contributions/tex-export-dialog', 'contributions_tex_export_dialog',
+                 management.RHContributionExportTexConfig, methods=('POST',))
+
+# Import
 _bp.add_url_rule('/manage/contributions/import', 'contributions_import',
                  management.RHContributionsImportCSV, methods=('GET', 'POST'))
 
--- a/indico/modules/events/contributions/controllers/management.py
+++ b/indico/modules/events/contributions/controllers/management.py
@@ -503,17 +503,19 @@ def render_archive(event, contribs, sort_by, opts):
     pdf = opts(event, session.user, contribs, tz=event.timezone, sort_by=sort_by)
     res = pdf.generate(return_source=True)
 
-    files = os.listdir(pdf._dir)
     temp_file = NamedTemporaryFile(suffix='indico.tmp', dir=config.TEMP_DIR)
     with ZipFile(temp_file.name, 'w', allowZip64=True) as zip_handler:
-        for f in files:
-            af = os.path.abspath(os.path.join(pdf._dir, f))
-            print af
-            if os.path.islink(af):
-                continue
-            if os.path.isdir(af):
-                continue
-            zip_handler.write(af, os.path.basename(f).encode('utf-8'))
+        for dirpath, dirnames, files in os.walk(pdf.source_dir, followlinks=True):
+            for f in files:
+                if f.startswith('.'):
+                    continue
+                if f.endswith(('.py', '.pyc', '.pyo')):
+                    continue
+                pf = os.path.join(dirpath, f)
+                af = os.path.abspath(pf)
+                rp = os.path.relpath(pf, pdf.source_dir)
+                archivename = rp.encode('utf-8')
+                zip_handler.write(af, archivename)
 
     temp_file.delete = False
     zip_file_name = 'contributions-tex.zip'
--- a/indico/modules/events/contributions/controllers/management.py
+++ b/indico/modules/events/contributions/controllers/management.py
@@ -7,16 +7,22 @@
 
 from __future__ import unicode_literals
 
+import os
+import uuid
 from operator import attrgetter
+from tempfile import NamedTemporaryFile
+from zipfile import ZipFile
 
 from flask import flash, jsonify, redirect, request, session
 from sqlalchemy.orm import undefer
 from werkzeug.exceptions import BadRequest, Forbidden, NotFound
 
+from indico.core.config import config
 from indico.core.db import db
 from indico.core.db.sqlalchemy.protection import ProtectionMode, render_acl
 from indico.core.permissions import get_principal_permissions, update_permissions
-from indico.legacy.pdfinterface.latex import ContribsToPDF, ContributionBook
+from indico.legacy.common.cache import GenericCache
+from indico.legacy.pdfinterface.latex import AbstractBook, ContribsToPDF, ContributionBook, LaTeXRuntimeException
 from indico.modules.attachments.controllers.event_package import AttachmentPackageGeneratorMixin
 from indico.modules.events.abstracts.forms import AbstractContentSettingsForm
 from indico.modules.events.abstracts.settings import BOASortField, abstracts_settings
@@ -24,8 +30,9 @@ from indico.modules.events.contributions import contribution_settings, get_contr
 from indico.modules.events.contributions.clone import ContributionCloner
 from indico.modules.events.contributions.controllers.common import ContributionListMixin
 from indico.modules.events.contributions.forms import (ContributionDefaultDurationForm, ContributionDurationForm,
-                                                       ContributionProtectionForm, ContributionStartDateForm,
-                                                       ContributionTypeForm, SubContributionForm)
+                                                       ContributionExportTeXForm, ContributionProtectionForm,
+                                                       ContributionStartDateForm, ContributionTypeForm,
+                                                       SubContributionForm)
 from indico.modules.events.contributions.lists import ContributionListGenerator
 from indico.modules.events.contributions.models.contributions import Contribution
 from indico.modules.events.contributions.models.fields import ContributionField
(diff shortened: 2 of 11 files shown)


indico/modules/events/contributions/blueprint.py
```diff
@@ -50,6 +50,23 @@
                  management.RHContributionsExportPDFBook, methods=('POST',))
 _bp.add_url_rule('/manage/contributions/book-sorted.pdf', 'contributions_pdf_export_book_sorted',
                  management.RHContributionsExportPDFBookSorted, methods=('POST',))
+_bp.add_url_rule('/manage/contributions/book-sorted-schedule.pdf', 'contributions_pdf_export_book_sorted_schedule',
```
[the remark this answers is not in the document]
bpedersen2: Agreed.  I'll try to find out how the dialogs are created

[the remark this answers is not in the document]
bpedersen2: I am now able to get a dialog first, but struggle to break out of the ajax dialog for file reception,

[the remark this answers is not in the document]
ThiefMaster: check what we do when printing badges for selected registrations (generate uuid, store data in cache, redirect client-side to the endpoint generating the pdf based on the cached data)

bpedersen2:
One more question: Does the license of indico_fonts allow copying the needed files into the temporary folder (instead of specifying the full path that probably is only available serverside)?

ThiefMaster:
Copying stuff around on the server is always OK, but I think even including the fonts in a downloadable TeX archive should be fine. But a downloadable TeX archive should probably contain a README.txt file that has a link to the license information on this page: https://github.com/indico/indico-fonts

indico/modules/events/contributions/blueprint.py
```diff
@@ -46,10 +46,14 @@
                  management.RHContributionsExportExcel, methods=('POST',))
 _bp.add_url_rule('/manage/contributions/contributions.pdf', 'contributions_pdf_export',
                  management.RHContributionsExportPDF, methods=('POST',))
-_bp.add_url_rule('/manage/contributions/book.pdf', 'contributions_pdf_export_book',
-                 management.RHContributionsExportPDFBook, methods=('POST',))
-_bp.add_url_rule('/manage/contributions/book-sorted.pdf', 'contributions_pdf_export_book_sorted',
-                 management.RHContributionsExportPDFBookSorted, methods=('POST',))
+
+# Tex-based exports
+_bp.add_url_rule('/manage/contributions/<uuid>/book-tex.zip', 'contributions_tex_export_book',
```
ThiefMaster: I don't like mixing this UUID and contribution IDs on the same URL path level. I would go for a structure like this:

```
/manage/contributions/export/tex/  # for the form
/manage/contributions/export/tex/<uuid>  # for the download
```

indico/modules/events/contributions/blueprint.py
```diff
@@ -46,10 +46,14 @@
                  management.RHContributionsExportExcel, methods=('POST',))
 _bp.add_url_rule('/manage/contributions/contributions.pdf', 'contributions_pdf_export',
                  management.RHContributionsExportPDF, methods=('POST',))
-_bp.add_url_rule('/manage/contributions/book.pdf', 'contributions_pdf_export_book',
-                 management.RHContributionsExportPDFBook, methods=('POST',))
-_bp.add_url_rule('/manage/contributions/book-sorted.pdf', 'contributions_pdf_export_book_sorted',
-                 management.RHContributionsExportPDFBookSorted, methods=('POST',))
+
+# Tex-based exports
```
ThiefMaster: TeX ;)

indico/modules/events/contributions/forms.py
```diff
@@ -203,3 +205,22 @@ def validate_name(self, field):
...
+                                    sorted=True)
+    contrib_ids = HiddenFieldList()
+    submitted = HiddenField()
+
+    def __init__(self, *args, **kwargs):
+        self.contribs = kwargs.get('contribs')
+        super(ContributionExportTeXForm, self).__init__(*args, **kwargs)
+        self.contrib_ids.data = [c.id for c in self.contribs]
+        self.format.choices = {'PDF': 'PDF',
+                               'TeX': 'Zipped TeX directory'}.items()
```
ThiefMaster: `_('...')` for i18n, and I would just put `[(.., ..), (.., ..)]` here instead of creating a dict and then getting its items

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -93,6 +93,17 @@
                         {%- trans %}Export{% endtrans -%}
                     </a>
                     <ul class="dropdown">
+                        <li>
+                            <a href ="#" class="icon-file-pdf js-enable-if-checked disabled"
+                               data-href="{{ url_for('.contributions_tex_export_dialog', event) }}"
+                               data-method="POST"
+                               data-params-selector="#contribution-list input[name=contribution_id]:checked"
+                               data-ajax-dialog
+                             >
```
ThiefMaster: should go into the previous line

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -93,6 +93,17 @@
...
+                                {%- trans %}Export via LaTeX{% endtrans -%}
```
ThiefMaster: "LaTeX package" or something like that. "Export" is redundant since the button is already  labelled "Export"

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -101,14 +112,21 @@
...
+                            <a href="#" class="icon-file-pdf js-submit-form js-enable-if-checked disabled"
+                               data-href="{{ url_for('.contributions_pdf_export_book_sorted_session', event) }}">
+                                {%- trans %}PDF (book of abstracts - sorted by session and board number){% endtrans -%}
                             </a>
                         </li>
                         <li>
                             <a href="#" class="icon-file-pdf js-submit-form js-enable-if-checked disabled"
-                               data-href="{{ url_for('.contributions_pdf_export_book_sorted', event) }}">
-                                {%- trans %}PDF (book of abstracts - sorted by board number){% endtrans -%}
+                               data-href="{{ url_for('.contributions_pdf_export_book_sorted_all', event) }}">
```
ThiefMaster: I think this is obsolete after your latest chances

indico/modules/events/contributions/templates/management/exportdialog.html
```diff
@@ -0,0 +1,11 @@
+{% from 'forms/_form.html' import simple_form  %}
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('highlight', large_icon=true) %}
+    {% trans -%}
+    Please select the  desired format and sort order.
```
ThiefMaster: indentation, double space

ThiefMaster: Maybe adding this message is overkill. If you don't use it you can even get rid of this template altogether and just use `jsonify_form()` in the Python code.

If you do keep the template, rename it to `export_dialog.html` please

indico/modules/events/util.py
```diff
@@ -415,11 +415,11 @@ def get_base_ical_parameters(user, detail, path, params=None):
             'user_logged': user is not None, 'request_urls': request_urls}
 
 
-def create_event_logo_tmp_file(event):
+def create_event_logo_tmp_file(event, tmppath=None):
     """Creates a temporary file with the event's logo"""
     logo_meta = event.logo_metadata
     logo_extension = guess_extension(logo_meta['content_type']) or os.path.splitext(logo_meta['filename'])[1]
-    temp_file = NamedTemporaryFile(delete=False, dir=config.TEMP_DIR, suffix=logo_extension)
+    temp_file = NamedTemporaryFile(delete=False, dir=tmppath or config.TEMP_DIR, suffix=logo_extension)
```
ThiefMaster: `dir=(tmppath or config.TEMP_DIR)` for readability

indico/util/mdx_latex.py
```diff
@@ -408,7 +409,7 @@ def tolatex(self, ournode):
         elif ournode.tag == 'q':
             buffer += "`%s'" % subcontent.strip()
         elif ournode.tag == 'p':
-            if 'apply_br' in self.configs:
+            if self.configs.get('apply_br', False):
```
ThiefMaster: `, False` isn't needed, since the default `None` is already falsy

indico/modules/events/contributions/controllers/management.py
```diff
@@ -456,17 +466,60 @@ def _process(self):
...
-    def _process(self):
-        pdf = ContributionBook(self.event, session.user, self.contribs, tz=self.event.timezone,
-                               sort_by='board_number')
-        return send_file('book-of-abstracts.pdf', pdf.generate(), 'application/pdf')
+        return jsonify_template('events/contributions/management/exportdialog.html', form=form, event=self.event)
+
+
+class RHContributionsExportTeXBook(RHManageContributionsExportActionsBase):
+    ALLOW_LOCKED = True
+    normalize_url_spec = {
```
ThiefMaster: I don't think there is any URL normalization defined in the parent classes, so this is not necessary. Especially since a normalization spec without any locators is useless, since it won't do anything.

indico/modules/events/contributions/controllers/management.py
```diff
@@ -57,18 +64,21 @@
...
 from indico.util.string import handle_legacy_description
 from indico.web.flask.templating import get_template_module
-from indico.web.flask.util import send_file, url_for
+from indico.web.flask.util import redirect_or_jsonify, send_file, url_for
 from indico.web.forms.base import FormDefaults
 from indico.web.forms.fields.principals import serialize_principal
 from indico.web.util import jsonify_data, jsonify_form, jsonify_template
 
 
+export_list_cache = GenericCache('export-list-cache')
```
ThiefMaster: I'd remove the `-cache` from the GenericGache name

indico/modules/events/contributions/controllers/management.py
```diff
@@ -456,17 +466,60 @@ def _process(self):
...
-class RHContributionsExportPDFBookSorted(RHManageContributionsExportActionsBase):
-    def _process(self):
-        pdf = ContributionBook(self.event, session.user, self.contribs, tz=self.event.timezone,
-                               sort_by='board_number')
-        return send_file('book-of-abstracts.pdf', pdf.generate(), 'application/pdf')
+        return jsonify_template('events/contributions/management/exportdialog.html', form=form, event=self.event)
+
+
+class RHContributionsExportTeXBook(RHManageContributionsExportActionsBase):
+    ALLOW_LOCKED = True
```
ThiefMaster: not necessary since this is a GET endpoint and GET is always allowed, even in a locked event

ThiefMaster:
Regarding #3039, the BoA uses a slightly different template than what you get when downloading a PDF from the contribution list (see the `AbstractBook` class). Maybe it'd be good to add a link to download that particular BoA latex package (which then takes the BoA settings into account) somewhere in the abstracts module. Also, since building the PDF BoA can take some time for a big event, I wouldn't even try building it when downloading a TeX archive. After all, this is something meant for people who want to build their own PDF from the TeX sources after editing them...

indico/modules/events/contributions/templates/management/contributions.html
bpedersen2: Done

indico/modules/events/contributions/blueprint.py
bpedersen2: Dne

indico/modules/events/contributions/forms.py
bpedersen2: Changes with the signal introduction

bpedersen2:
currently I allow the signal only on the LaTeX based exports. I think it would also be possible to also include the other exports as well. 

Pros:  

- easier to add more exporters
- single interface again
- sort options possible on all export types(not strictly needed for e.g. table-like exports)

Cons:

- Endpoints for direct acces get more complex
- Even for simple (currently) option-less exports tow more clicks are needed
- harder to add the nice styling with icons

Comments?

bpedersen2:
I pushed a (yet untested) rebase in a new  branch: https://github.com/bpedersen2/indico/tree/texexport_v2 

(I am struggling with webpack barfing on lots of  missing/non-matching packages)

ThiefMaster:
What error do you get? Doing a `npm install` should install everything it needs...

ThiefMaster:
BTW, for zipping the directory you could use this util instead of doing it manually: https://docs.python.org/3/library/shutil.html#shutil.make_archive

bpedersen2:
> BTW, for zipping the directory you could use this util instead of doing it manually: https://docs.python.org/2/library/shutil.html#shutil.make_archive

I thought of that as  well, but the manual approach allows a finer-grained filtering of what gets packaged.

ThiefMaster:
Is there anything in the folder we don't want included?

bpedersen2:
> Is there anything in the folder we don't want included?

Currently e.g. the dummy __index__.py/pyc from the font package and any dot-prefixed files

bpedersen2:
Seems like this is superseded by #4035? If yes, then we can close this one.

ThiefMaster:
Yes indeed, that PR is actually based on yours.

PR #3439, opened 2018-06-27
react-router-redux => connected-react-router

### LINKED ISSUE #3438: Replace react-router-redux with connected-react-router
react-router-redux has been deprecated and we should move to [connected-react-router](https://github.com/supasate/connected-react-router).

pferreir:
Closes #3438


indico/web/client/js/utils/redux.js
```diff
@@ -29,7 +33,7 @@ function _combineReducers(...reducers) {
                 // it's a reducer
                 state = reducer(state, action);
             } else {
-                state = combineReducers(reducer)(state, action);
+                state = memoizedCombineReducers(reducer)(state, action);
```
ThiefMaster: Wouldn't it be cleaner to do the combining once when `_combineReducers()` is called instead of when the `(state, action) => ...` function gets called?

pferreir: Yeah, I thought of that but it complicated the code quite a bit. But I believe I have another idea.

indico/web/client/js/utils/redux.js
```diff
@@ -23,31 +24,24 @@ import {composeWithDevTools} from 'redux-devtools-extension';
...
-            }
-        }
-        return state;
-    };
-}
+    const reducerFunctions = reducers.map(reducer => (
+        (typeof reducer === 'function') ? reducer : combineReducers(reducer))
+    );
 
+    return (state, action) => reducerFunctions.reduce((st, fn) =>  fn(st, action), state);
```
ThiefMaster: double space after the second `=>`

pferreir: weird... ESLint doesn't seem to care about that.

PR #3445, opened 2018-06-29
Enforce survey submission limit

### LINKED ISSUE #3256: Enforce survey limit
Currently a limit for the number of submissions can be set, but we don't enforce it anywhere.

system7ltd:
Hi, this should fix #3256 .  Could you please review?
Cheers,
T

### DIFF
--- a/indico/modules/events/surveys/models/surveys.py
+++ b/indico/modules/events/surveys/models/surveys.py
@@ -242,7 +242,9 @@ class Survey(db.Model):
         if not self.has_ended:
             if not self.submissions:
                 return SurveyState.active_and_clean
-            return SurveyState.active_and_answered
+            # Do not accept any more submissions if submission limit reached.
+            if self.submission_limit is None or len(self.submissions) < self.submission_limit:
+                return SurveyState.active_and_answered
         return SurveyState.finished
 
     @property


indico/modules/events/surveys/models/surveys.py
```diff
@@ -242,7 +242,10 @@ def state(self):
         if not self.has_ended:
             if not self.submissions:
                 return SurveyState.active_and_clean
-            return SurveyState.active_and_answered
+            # Do not accept any more submissions if submission limit reached.
+            if self.submission_limit is None or\
```
mic4ael: Please, get rid of `\` and just put the second condition on the same line ;)

mic4ael:
Generally, I think it should do. But it would be also nice to limit that on the controller level `RHSubmitSurvey`

ThiefMaster:
Yeah, my first though was that I wouldn't touch the state for this as well but rather check it separately in the controller, and show a message that the max number of submissions has been reached.

OTOH, doing it in the state has the advantage that there's a single place for it - but in that case I would add a new state indicating that the limit has been reached and not use one of the existing states for it.

mic4ael:
Yep, totally agree that it would be much better to have a separate state for that.

ThiefMaster:
We use PEP8 but with a 120 char max line length (and around the usual 72 for text lines, i.e. comments/docstrings)

mic4ael:
I think there is no need for two commits like this. You could easily squash them ;)

ThiefMaster:
Regarding "just merge it as-is": I don't think it's very clear to a user (or a manager!) if the survey just shows up as closed without any indicator why that's the case.

system7ltd:
@ThiefMaster Thank you for your answer re. coding style.  As for the other concerns you raised, not sure if it matters to the user why they cannot make submissions anymore as long as it is clearly shown to them that the survey is closed for submissions.  I absolutely agree that it would be much nicer to have the fact clearly stated in the manager's view (not sure though if the managers would not prefer to have their survey limit enforced as is ASAP).  OTOH the solution I have provided fixes #3256 according to the bug specification you wrote (i.e. enforces the survey submission limit - "Currently a limit for the number of submissions can be set, but we don't enforce it anywhere") and the bug has been open for almost four months now - hence why I suggested that it might be a good idea to merge this and fix the bug, for the sake of agility and advancing forward with small iterations ;)... and then, while it is fixed and both the users and the managers are slightly happier ;), create a new enhancement proposal to address all the new issues and ideas not specified in #3256, have a lively discussion and decide on the best approach, the new state name, UI changes, messages, l10n/i18n etc..  Alternatively, if you really do not like the idea and have a reason to achieve all that in one step, I would suggest marking #3256 as 'wontfix', closing this issue, creating a new issue with an enhancement proposal described above (referencing #3256).  What do you think? 

@mic4ael I could, is that a style guideline for Indico?  I am happy to oblige, but I am under the impression that we will probably not be merging the fix anyway, so let's wait and see ;)

mic4ael:
@system7ltd I think it is just cleaner to have it in a single commit rather than two since it is a really small change. At least to me, this kind of commits only clutter the history a bit and apart from that it is much easier to look at all changes as a whole (as one commit; of course only when appropriate) rather than browsing through all of them just finding out that the rest are only code style changes ;)

system7ltd:
@mic4ael I think you are right.  I got accustomed to workflows where rebasing / squashing was generally discouraged.  Anyway, here it is, for the sake of completeness, even if we do not end up merging the fix, happy to oblige as I said ;)

ThiefMaster:
hen the limit is reached, the management UI is inconsistent:

image

- it shows the current time as "when it finished"
- "reopen now" doesn't do anything

Anyway, I'll adapt this and add a new state where no reopen button is shown.

ThiefMaster:
04d073214c4a575336320d230b3f4a3e0fc06922

PR #3450, opened 2018-07-05
Allow bookings to be made from 'room details'

### LINKED ISSUE #3441: Add button to book from room details
This should present the user with a modal to choose date/time/recurrence. We should check for conflicts and allow the possibility to go to the timeline view if the user wants to sort them out.

pferreir:
Closes #3441


indico/modules/rb_new/client/js/components/pages/RoomList.jsx
```diff
@@ -73,8 +72,26 @@ export default class RoomList extends React.Component {
         );
     };
 
+    closeBookingModal = () => {
+        const {pushState} = this.props;
+        pushState(`/rooms`, true);
```
ThiefMaster: `'`

indico/modules/rb_new/client/js/components/pages/RoomList.jsx
```diff
@@ -73,8 +72,26 @@ export default class RoomList extends React.Component {
...
+    };
+
+    roomPreloader(componentFunc) {
+        const {fetchRoomDetails} = this.props;
+        return ({match: {params: {roomId}}}) => (
+            <Preloader checkCached={({roomDetails: {rooms: cachedRooms}}) => {
```
ThiefMaster: like above, weird-looking (albeit correct) indentation after the callback

indico/modules/rb_new/controllers/backend.py
```diff
@@ -144,6 +144,10 @@ def _process(self, args):
             data['room'] = rooms_schema.dump(data['room'], many=False).data
             data.update({k: serialize_occurrences(data[k])
                          for k in ['candidates', 'pre_bookings', 'bookings', 'conflicts', 'pre_conflicts']})
+            data.update({
```
ThiefMaster: IMO bracket notation looks better for explicit/hardcoded keys

indico/modules/rb_new/controllers/backend.py
```diff
@@ -144,6 +144,10 @@ def _process(self, args):
...
+                'num_days_available': len(date_range) - len(data['conflicts']),
+                'all_days_available': not len(data['conflicts'])
```
ThiefMaster: `not data['conflicts']`

indico/web/client/js/react/util/Preloader.jsx
```diff
@@ -0,0 +1,49 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import {connect} from 'react-redux';
+
+
+class Preloader extends React.Component {
+    static propTypes = {
+        isCached: PropTypes.bool.isRequired,
+        action: PropTypes.func.isRequired,
+        children: PropTypes.func.isRequired,
```
ThiefMaster: what about using a [render= prop](https://reactjs.org/docs/render-props.html) instead of callable children?

indico/modules/rb_new/client/js/components/pages/RoomList.jsx
ThiefMaster: I was looking for a rule but couldn't find one - just the [other way round](https://eslint.org/docs/rules/no-template-curly-in-string)

indico/modules/rb_new/controllers/backend.py
pferreir: Too much JavaScript does this :rage2:

indico/web/client/js/react/util/Preloader.jsx
pferreir: Not sure what the "correct" way is.

ThiefMaster: at least in the form plugins they usually support both... so probably there is none. it's javascript after all :D

pferreir: I personally find it weird to have a massive render prop in a tag that then has no children. Not to mention indentation.

ThiefMaster: true, if it's massive it's probably ugly...

PR #3451, opened 2018-07-05
Make registration period display less confusing

### LINKED ISSUE #3359: Make registration period display less confusing
Currently some users are confusing the registration start/end dates with the event dates.

A possible solution would be showing both the start time and deadline (as it's right now) if registrations are still closed, but once they are open just show the deadline (labelled as such) since it's not relevant for a user when the registration period opened.

On the actual registration form it might be best to not show this information at all.

errikos:
- This resolves #3359.

### DIFF
--- a/indico/modules/events/registration/templates/display/_regform_info.html
+++ b/indico/modules/events/registration/templates/display/_regform_info.html
@@ -2,21 +2,37 @@
 
 {% macro render_regform_info(regform) %}
     <div class="infogrid condensed">
-        {% if regform.is_scheduled %}
+        {% if regform.is_scheduled and (not regform.has_started or regform.end_dt) %}
             <div class="infoline date">
                 <i class="icon icon-time"></i>
                 <div class="text">
                     <div>
-                        <span class="label">
-                            {% trans %}From{% endtrans %}
-                        </span>
-                        <span class="datetime">
-                            {{ regform.start_dt|format_date(timezone=regform.event.tzinfo) }}
-                        </span>
+                        {% if regform.is_open %}
+                            <span class="label">
+                                {% trans %}Registration is open{% endtrans %}
+                            </span>
+                        {% else %}
+                            <span class="label">
+                                {% if regform.has_ended %}
+                                    {% trans %}Opened{% endtrans %}
+                                {% else %}
+                                    {% trans %}Opens{% endtrans %}
+                                {% endif %}
+                            </span>
+                            <span class="datetime">
+                                {{ regform.start_dt|format_date(timezone=regform.event.tzinfo) }}
+                            </span>
+                        {% endif %}
                     </div>
                     <div>
                         <span class="label">
-                            {% trans %}To{% endtrans %}
+                            {% if regform.is_open %}
+                                {% trans %}Deadline{% endtrans %}
+                            {% elif regform.has_ended %}
+                                {% trans %}Closed{% endtrans %}
+                            {% else %}
+                                {% trans %}Closes{% endtrans %}
+                            {% endif %}
                         </span>
                         <span class="datetime">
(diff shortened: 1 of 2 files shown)


indico/modules/events/registration/templates/display/_regform_info.html
```diff
@@ -7,17 +7,29 @@
...
+                            <span class="datetime">
+                                {{ regform.start_dt|format_date(timezone=regform.event.tzinfo) }}
+                            </span>
+                        {% endif %}
                     </div>
                     <div>
-                        <span class="label">
-                            {% trans %}To{% endtrans %}
-                        </span>
+                        {% if regform.has_started %}
```
mvidalgarcia: What about having the if/else inside the `span` block so you avoid repetition?

ThiefMaster: Agreed, for this one it makes more sense. Also keep in mind that you can have a registration without an `end_dt`. In this case it'd probably be nicer to not show this block at all if the registration is running - "Registration is open" is kind of obvious.

errikos: Agreed and agreed. Although, for the second point, I don't like the empty space created by the missing end date span below the "Registration is open" label. Is there a nice way to push it in the middle of the vertical space?

ThiefMaster: That's why I would not show it at all if the registration is open and has no end date

errikos: I am not sure what you mean. Even if you don't show the span element, the empty space still exists. Check attachment.

capture du 2018-07-06 11-03-59

ThiefMaster: >  In this case it'd probably be nicer to not show this block at all if the registration is running - "Registration is open" is kind of obvious.

I meant hiding that whole block.

indico/modules/events/registration/templates/display/_regform_info.html
```diff
@@ -7,13 +7,17 @@
...
-                        {% if regform.has_started %}
+                        {% if regform.has_started and not regform.has_ended %}
                             <span class="label">
                                 {% trans %}Registration is open{% endtrans %}
                             </span>
                         {% else %}
                             <span class="label">
-                                {% trans %}Opens{% endtrans %}
+                                {% if regform.has_ended %}
+                                    {% trans %}Opened{% endtrans %}
```
ThiefMaster: i don't think it's really useful information for users to know when the registration period started

indico/modules/events/registration/templates/display/_regform_info.html
```diff
@@ -22,8 +26,10 @@
                     </div>
                     <div>
                         <span class="label">
-                            {% if regform.has_started %}
+                            {% if regform.has_started and not regform.has_ended %}
```
ThiefMaster: `is_open` covers this condition

indico/modules/events/registration/templates/display/_regform_info.html
errikos: It could be sometimes, for transparency reasons, e.g. if there are limited places for a conference and all get filled up quickly (just a quick thought). So you suggest to hide the schedule info altogether?

ThiefMaster: Yes, I would show:

- that the registration is open and what the deadline is [if there's a closing time]
- no schedule at all if there is no deadline and the registration is open
- when the registration opens(/closes) if the registration is not open
- when the registration opened/closed if the registration is closed

ThiefMaster:

    - Make registration period display less confusing (:issue:`3359`)

PR #3456, opened 2018-07-09
URI shortcut for upcoming category event

### LINKED ISSUE #3388: URI shortcut pointing to the upcoming event of a given category
Hi,

It'd be nice if it was possible to navigate to the next event of a given category by using an URI. For instance, something like:

```
https://indico.example.org/category/<catid>/upcoming
```

would take the user to the nearest event in time that's not due.

It's definitely more efficient than a human scanning all the dates listed in the category view and selecting the correct one based on the current date. Another advantage is that it can be bookmarked :)

Thanks.

errikos:
- This resolves #3388.

### DIFF
--- a/indico/modules/categories/blueprint.py
+++ b/indico/modules/categories/blueprint.py
@@ -22,11 +22,11 @@ from indico.modules.categories.compat import compat_category
 from indico.modules.categories.controllers.admin import RHManageUpcomingEvents
 from indico.modules.categories.controllers.display import (RHCategoryCalendarView, RHCategoryIcon, RHCategoryInfo,
                                                            RHCategoryLogo, RHCategoryOverview, RHCategorySearch,
-                                                           RHCategoryStatistics, RHDisplayCategory, RHEventList,
-                                                           RHExportCategoryAtom, RHExportCategoryICAL,
-                                                           RHReachableCategoriesInfo, RHShowFutureEventsInCategory,
-                                                           RHShowPastEventsInCategory, RHSubcatInfo,
-                                                           RHXMLExportCategoryInfo)
+                                                           RHCategoryStatistics, RHCategoryUpcomingEvent,
+                                                           RHDisplayCategory, RHEventList, RHExportCategoryAtom,
+                                                           RHExportCategoryICAL, RHReachableCategoriesInfo,
+                                                           RHShowFutureEventsInCategory, RHShowPastEventsInCategory,
+                                                           RHSubcatInfo, RHXMLExportCategoryInfo)
 from indico.modules.categories.controllers.management import (RHCreateCategory, RHDeleteCategory, RHDeleteEvents,
                                                               RHDeleteSubcategories, RHManageCategoryContent,
                                                               RHManageCategoryIcon, RHManageCategoryLogo,
@@ -83,6 +83,7 @@ _bp.add_url_rule('/show-past-events', 'show_past_events', RHShowPastEventsInCate
 _bp.add_url_rule('/statistics', 'statistics', RHCategoryStatistics)
 _bp.add_url_rule('/subcat-info', 'subcat_info', RHSubcatInfo)
 _bp.add_url_rule('/calendar', 'calendar', RHCategoryCalendarView)
+_bp.add_url_rule('/upcoming', 'upcoming_event', RHCategoryUpcomingEvent)
 
 # Event creation - redirect to anchor page opening the dialog
 _bp.add_url_rule('/create/event/<any(lecture,meeting,conference):event_type>', view_func=_redirect_event_creation)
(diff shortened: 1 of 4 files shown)


indico/modules/categories/blueprint.py
```diff
@@ -46,6 +46,9 @@ def _redirect_event_creation(category_id, event_type):
 _bp = IndicoBlueprint('categories', __name__, template_folder='templates', virtual_template_folder='categories',
                       url_prefix='/category/<int:category_id>')
 
+# Upcoming event of the category
```
ThiefMaster: I'd add it to the "Display" section below, without an extra header.

indico/modules/categories/controllers/display.py
```diff
@@ -670,3 +671,27 @@ def _get_event_data(self, event_query):
...
         return template.render_ongoing_events(ongoing_events, self.category.display_tzinfo)
+
+
+class RHCategoryUpcomingEvent(RHDisplayCategoryBase):
+    """Redirect to the upcoming event of a category."""
+
+    def _process(self):
+        event_id = self._get_upcoming_event()
+        if event_id:
+            resp = redirect(url_for('events.display', confId=event_id))
```
ThiefMaster: Return the event instead of its ID and you can use `event.url` here.

Also, we have some extra magic in our custom `url_for` wrapper that checks for a `locator` property on the object so you can also do something like `url_for('events.display', some_event)` because the locator returns `{'confId': self.id}` for an event.

indico/modules/categories/controllers/display.py
```diff
@@ -670,3 +671,27 @@ def _get_event_data(self, event_query):
...
+        else:
+            flash(_('There are no upcoming events for this category'))
+            resp = redirect(url_for('categories.display', category_id=self.category.id))
+        return resp
```
ThiefMaster: just `return redirect(...)` instead of assigning it

indico/modules/categories/controllers/display.py
```diff
@@ -670,3 +671,27 @@ def _get_event_data(self, event_query):
...
+
+    def _get_upcoming_event(self):
+        query = (Event.query
+                 .filter(Event.is_visible_in(self.category),
+                         Event.start_dt > now_utc())
+                 .options(load_only('id', 'start_dt'))
+                 .order_by(Event.start_dt)
+                 .order_by(Event.id))  # in case two events start exactly at the same time
```
ThiefMaster: that's obvious, no need to comment it ;) you can also do `.order_by(Event.start_dt, Event.id)`

indico/modules/categories/controllers/display.py
```diff
@@ -670,3 +671,27 @@ def _get_event_data(self, event_query):
...
+        else:
+            flash(_('There are no upcoming events for this category'))
+            resp = redirect(url_for('categories.display', category_id=self.category.id))
+        return resp
+
+    def _get_upcoming_event(self):
+        query = (Event.query
+                 .filter(Event.is_visible_in(self.category),
+                         Event.start_dt > now_utc())
+                 .options(load_only('id', 'start_dt'))
```
ThiefMaster: You don't need to load the `start_dt`, but the access check needs some other columns so currently your will spam queries for each access check. It's better to just not restrict the columns except for performance-critical places (which this isn't).

However, to avoid query spam in to the access checks you probably want `.options(subqueryload('acl_entries'))`.

indico/modules/categories/controllers/display.py
```diff
@@ -670,3 +671,27 @@ def _get_event_data(self, event_query):
...
+class RHCategoryUpcomingEvent(RHDisplayCategoryBase):
+    """Redirect to the upcoming event of a category."""
+
+    def _process(self):
+        event_id = self._get_upcoming_event()
+        if event_id:
+            resp = redirect(url_for('events.display', confId=event_id))
+        else:
+            flash(_('There are no upcoming events for this category'))
+            resp = redirect(url_for('categories.display', category_id=self.category.id))
```
ThiefMaster: `self.category.url`

indico/modules/categories/blueprint.py
```diff
@@ -83,6 +83,7 @@ def _redirect_event_creation(category_id, event_type):
 _bp.add_url_rule('/statistics', 'statistics', RHCategoryStatistics)
 _bp.add_url_rule('/subcat-info', 'subcat_info', RHSubcatInfo)
 _bp.add_url_rule('/calendar', 'calendar', RHCategoryCalendarView)
+_bp.add_url_rule('/upcoming', 'category_upcoming', RHCategoryUpcomingEvent, methods=('GET',))
```
ThiefMaster: GET is the default, no need to specify it

ThiefMaster:

    - Add shortcut to next upcoming event in a category (:issue:`3388`)

PR #3459, opened 2018-07-12
Upcoming events with subcategories

### LINKED ISSUE #3449: Support checking the category chain for upcoming events
This should be optional since the performance is better for a direct check.

Could be done using a new entry type "Category & Subcategories" in the upcoming events config which then checks using `is_visible_in` instead of just the `category_id`

errikos:
- Fix issue with Category CTEs not being usable to union queries.
- Add a new entry to the upcoming events administration section that allows to display the upcoming events of the hierarchy.
- This resolves #3449.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,7 @@ Improvements
   thanks :user:`nop33`)
 - Include event page title in the page's ``<title>`` (:issue:`3285`,
   thanks :user:`bpedersen2`)
+- Add option to include subcategories in upcoming events (:issue:`3449`)
 
 Bugfixes
 ^^^^^^^^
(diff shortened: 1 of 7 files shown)


ThiefMaster:

    - Add option to include subcategories in upcoming events (:issue:`3449`)

ThiefMaster:
I just noticed that this can easily take ~2 seconds to query against our production DB depending on the category. Maybe it'd be better to move it to an hourly celery task?

Or we can keep it as-is, assuming people are not gonna use this on huge categories.

ThiefMaster:
> Or we can keep it as-is, assuming people are not gonna use this on huge categories.

Easiest option for now. Most people have smaller DBs and someone using it on a huge category is unlikely.

PR #3463, opened 2018-07-17
Allow setting event-wide contribution duration

### LINKED ISSUE #3446: Allow setting event-wide contribution duration
This could be an event setting, maybe accessible from the 'cog' icon in the contribution list?

errikos:
- This resolves #3446.

### DIFF
--- a/indico/modules/events/contributions/blueprint.py
+++ b/indico/modules/events/contributions/blueprint.py
@@ -103,6 +103,10 @@ _bp.add_url_rule('/manage/contributions/types/<int:contrib_type_id>', 'manage_ty
 _bp.add_url_rule('/manage/contributions/types/<int:contrib_type_id>/delete', 'delete_type',
                  management.RHDeleteContributionType, methods=('POST',))
 
+# Default contribution duration
+_bp.add_url_rule('/manage/contributions/duration', 'manage_default_duration',
+                 management.RHManageDefaultContributionDuration, methods=('GET', 'POST',))
+
 # Custom contribution fields
 _bp.add_url_rule('/manage/contributions/fields/', 'manage_fields', management.RHManageContributionFields)
 _bp.add_url_rule('/manage/contributions/fields/create/<field_type>', 'create_field',
--- a/indico/modules/events/contributions/controllers/management.py
+++ b/indico/modules/events/contributions/controllers/management.py
@@ -29,12 +29,12 @@ from indico.legacy.pdfinterface.conference import ContribsToPDF, ContributionBoo
 from indico.modules.attachments.controllers.event_package import AttachmentPackageGeneratorMixin
 from indico.modules.events.abstracts.forms import AbstractContentSettingsForm
 from indico.modules.events.abstracts.settings import abstracts_settings
-from indico.modules.events.contributions import get_contrib_field_types
+from indico.modules.events.contributions import contribution_settings, get_contrib_field_types
 from indico.modules.events.contributions.clone import ContributionCloner
 from indico.modules.events.contributions.controllers.common import ContributionListMixin
-from indico.modules.events.contributions.forms import (ContributionDurationForm, ContributionProtectionForm,
-                                                       ContributionStartDateForm, ContributionTypeForm,
-                                                       SubContributionForm)
+from indico.modules.events.contributions.forms import (ContributionDefaultDurationForm, ContributionDurationForm,
+                                                       ContributionProtectionForm, ContributionStartDateForm,
+                                                       ContributionTypeForm, SubContributionForm)
 from indico.modules.events.contributions.lists import ContributionListGenerator
 from indico.modules.events.contributions.models.contributions import Contribution
 from indico.modules.events.contributions.models.fields import ContributionField
@@ -175,8 +175,10 @@ class RHCreateContribution(RHManageContributionsBase):
     def _process(self):
         inherited_location = self.event.location_data
         inherited_location['inheriting'] = True
+        default_duration = contribution_settings.get(self.event, 'default_duration')
         contrib_form_class = make_contribution_form(self.event)
-        form = contrib_form_class(obj=FormDefaults(location_data=inherited_location), event=self.event)
+        form = contrib_form_class(obj=FormDefaults(location_data=inherited_location, duration=default_duration),
+                                  event=self.event)
         if form.validate_on_submit():
             # Create empty contribution so it can be compared to the new one in flash_if_unregistered
             contrib = Contribution()
@@ -494,6 +496,18 @@ class RHManageContributionTypes(RHManageContributionsBase):
         contrib_types = self.event.contribution_types.order_by(db.func.lower(ContributionType.name)).all()
         return jsonify_template('events/contributions/management/types_dialog.html', event=self.event,
                                 contrib_types=contrib_types)
+
+
+class RHManageDefaultContributionDuration(RHManageContributionsBase):
+    """Dialog to manage the default contribution duration"""
+
+    def _process(self):
+        form = ContributionDefaultDurationForm(duration=contribution_settings.get(self.event, 'default_duration'))
+        if form.validate_on_submit():
+            contribution_settings.set(self.event, 'default_duration', form.duration.data)
+            flash(_("Default contribution duration was changed successfully"))
+            return jsonify_data()
+        return jsonify_form(form)
 
 
 class RHManageContributionTypeBase(RHManageContributionsBase):
--- a/indico/modules/events/contributions/forms.py
+++ b/indico/modules/events/contributions/forms.py
@@ -183,6 +183,11 @@ class ContributionDurationForm(IndicoForm):
                 raise ValidationError(error_msg)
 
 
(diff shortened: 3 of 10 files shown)


indico/modules/events/contributions/blueprint.py
```diff
@@ -103,6 +103,10 @@
 _bp.add_url_rule('/manage/contributions/types/<int:contrib_type_id>/delete', 'delete_type',
                  management.RHDeleteContributionType, methods=('POST',))
 
+# Default contribution duration
+_bp.add_url_rule('/manage/contributions/duration/', 'manage_default_duration',
```
ThiefMaster: `/manage/contributions/default-duration` (no trailing slash if the url segment has nothing inside)

indico/modules/events/contributions/controllers/management.py
```diff
@@ -175,8 +178,10 @@ class RHCreateContribution(RHManageContributionsBase):
     def _process(self):
         inherited_location = self.event.location_data
         inherited_location['inheriting'] = True
+        default_duration = TimedeltaConverter.to_python(event_core_settings.get(self.event, 'contrib_duration'))
```
ThiefMaster: Don't use the converter here; specify it in the EventSettingsProxy - it will automatically convert on get/set.

indico/modules/events/contributions/controllers/management.py
```diff
@@ -496,6 +501,18 @@ def _process(self):
                                 contrib_types=contrib_types)
 
 
+class RHManageDefaultContributionDuration(RHManageContributionsBase):
+    """Dialog to manage the event-wide contribution duration"""
+
+    def _process(self):
+        default_duration = TimedeltaConverter.to_python(event_core_settings.get(self.event, 'contrib_duration'))
```
ThiefMaster: see my comment above

indico/modules/events/contributions/forms.py
```diff
@@ -183,6 +183,11 @@ def validate_duration(self, field):
                 raise ValidationError(error_msg)
 
 
+class ContributionDefaultDurationForm(IndicoForm):
+    duration = TimeDeltaField(_('Duration'), [DataRequired(), MaxDuration(timedelta(days=1))],
+                              default=timedelta(minutes=20), units=('minutes', 'hours'))
```
ThiefMaster: having a default value here is not very useful, since you always provide on when instantiating the form anyway.

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -22,6 +22,15 @@
...
                     {% trans %}Contribution types{% endtrans %}</a>
             </li>
+            <li>
+                <a href="#"
+                   title="{% trans %}Set default contribution duration{% endtrans %}"
+                   data-title="{% trans %}Set event-wide contribution duration{% endtrans %}"
+                   data-href="{{ url_for('.manage_default_duration', event) }}"
+                   data-qtip-position="right"
+                   data-ajax-dialog>
+                    {% trans %}Contribution duration{% endtrans %}</a>
```
ThiefMaster: "Default contribution duration"?

indico/modules/events/models/events.py
```diff
@@ -419,6 +419,7 @@ def __table_args__(cls):
     end_dt_override = _EventSettingProperty(event_core_settings, 'end_dt_override')
     organizer_info = _EventSettingProperty(event_core_settings, 'organizer_info')
     additional_info = _EventSettingProperty(event_core_settings, 'additional_info')
+    contrib_duration = _EventSettingProperty(event_core_settings, 'contrib_duration')
```
ThiefMaster: I don't think we need this, in the few places where we access it we can get it from the proxy

indico/modules/events/settings.py
```diff
@@ -250,7 +251,8 @@ def get_themes_for(self, event_type):
     'start_dt_override': None,
     'end_dt_override': None,
     'organizer_info': '',
-    'additional_info': ''
+    'additional_info': '',
+    'contrib_duration': TimedeltaConverter.from_python(timedelta(minutes=20))
```
ThiefMaster: instead of using `event_core_settings` i would add a new `contribution_settings` EventSettingsProxy in `indico/modules/contributions/__init__.py`.

also, set the converter in the `converters` arg instead of calling it manually

indico/modules/events/settings.py
```diff
@@ -261,3 +262,9 @@ def get_themes_for(self, event_type):
     'emails': [],
     'phones': []
 })
+
+event_contribution_settings = EventSettingsProxy('contrib', {
```
ThiefMaster: this should be in `indico/modules/contributions/__init__.py` (just like we do for other modules such as `indico/modules/sessions/__init__.py`)

indico/modules/events/settings.py
```diff
@@ -261,3 +262,9 @@ def get_themes_for(self, event_type):
...
+    'contrib_duration': timedelta(minutes=20)
```
ThiefMaster: 'default_duration' is probably better now that it's in a contribution-specific settings proxy

indico/modules/events/settings.py
ThiefMaster: - `contribution_settings` for the proxy (when moving it)
- `'contributions'` as the first argument (the module name used when storing settings in the DB

indico/modules/events/contributions/forms.py
```diff
@@ -183,6 +183,14 @@ def validate_duration(self, field):
                 raise ValidationError(error_msg)
 
 
+class ContributionDefaultDurationForm(IndicoForm):
+    duration = TimeDeltaField(_('Duration'), [DataRequired(), MaxDuration(timedelta(days=1))],
+                              units=('minutes', 'hours'))
+
+    def __init__(self, *args, **kwargs):
```
ThiefMaster: this override isn't needed since it doesn't do anything besides calling the parent

indico/modules/events/contributions/controllers/management.py
```diff
@@ -496,6 +499,17 @@ def _process(self):
                                 contrib_types=contrib_types)
 
 
+class RHManageDefaultContributionDuration(RHManageContributionsBase):
+    """Dialog to manage the event-wide contribution duration"""
```
ThiefMaster: s/event-wide/default/ ;)

indico/modules/events/contributions/forms.py
errikos: oops I removed it and added it back while I was testing something

indico/modules/events/contributions/controllers/management.py
```diff
@@ -496,6 +497,18 @@ def _process(self):
...
 
 
+class RHManageDefaultContributionDuration(RHManageContributionsBase):
+    """Dialog to manage the default contribution duration"""
+
+    def _process(self):
+        form = ContributionDefaultDurationForm(duration=contribution_settings.get(self.event, 'default_duration'))
+        if form.validate_on_submit():
+            contribution_settings.set(self.event, 'default_duration', form.duration.data)
+            flash(_("Default contribution duration was changed successfully"))
```
errikos: added this flash, so that the user gets some sort of confirmation - maybe we need an error flash as well in case something goes wrong?

ThiefMaster: No, because if:

- form validation fails, the dialog stays open and the error is shown next to the field
- something goes completely wrong, that's handled automatically with the error report dialog

indico/modules/events/contributions/controllers/management.py
```diff
@@ -176,7 +176,8 @@ def _process(self):
         inherited_location = self.event.location_data
         inherited_location['inheriting'] = True
         contrib_form_class = make_contribution_form(self.event)
-        form = contrib_form_class(obj=FormDefaults(location_data=inherited_location), event=self.event)
+        form = contrib_form_class(obj=FormDefaults(location_data=inherited_location), event=self.event,
+                                  duration=contribution_settings.get(self.event, 'default_duration'))
```
ThiefMaster: convert the `_default_duration` in `ContributionEntryForm` to a property instead of setting it here; `self.event` should already be available there

indico/modules/events/contributions/controllers/management.py
```diff
@@ -496,6 +498,18 @@ def _process(self):
...
 
+class RHManageDefaultContributionDuration(RHManageContributionsBase):
+    """Dialog to manage the default contribution duration"""
+
+    def _process(self):
+        form = ContributionDefaultDurationForm(duration=contribution_settings.get(self.event, 'default_duration'))
+        if form.validate_on_submit():
+            contribution_settings.set(self.event, 'default_duration', form.duration.data)
+            flash(_("Default contribution duration was changed successfully"))
+            return jsonify_data(flash=True)
```
ThiefMaster: `flash=True` is already the default

ThiefMaster:
We also need to take this setting into account when accepting an abstract

Currently we have hardcoded 15 minutes in `create_contribution_from_abstract`:

    duration = contrib_session.default_contribution_duration if contrib_session else timedelta(minutes=15)

PR #3464, opened 2018-07-18
Update TeXLive installation instructions

### LINKED ISSUE #3460: documentation on latex installation has redundant information in texlive.profile
https://docs.getindico.io/en/latest/installation/latex/ defines `selected_scheme scheme-full`, which installs everything. It also includes `collection-* 1` lines for all currently available collections on x86_64-linux. The tlpkg/texlive.profile file created at the end of the installation has no `collection-*` lines. Removing the `collection-*` lines from the profile file given in the documentation does not change the created tlpkg/texlive.profile file (except the timestamp).

According to https://www.tug.org/texlive/doc/install-tl.html#PROFILES
>  If the variable selected_scheme is defined and no collection variables at all are defined, then the collections required by the specified scheme (which might change over time) are installed, without explicitly listing them. If any collections are specified in a profile, though, then all desired collections must be given explicitly.

Listing the collections would make sense, if the documentation would strive for a texlive installation which installs just the packages required for the application, but this does not seem to be the case here. For the 'just give me every texlive package, so that I do not need to track which package provides which functionality' `scheme-full` looks to be the easier and more maintainable approach.

### LINKED ISSUE #3461: documentation on latex installation sets options to unexpected values for server install
Correct me if I am wrong but some of the options set in texlive.profile feel wrong on an application server:

- `option_desktop_integration 1` would create a start menu entry.
- `option_doc 1` is anyone logging into the application server to read the documentation?
- `option_src 1` are the packages modified and rebuild?

Those options are documented in texlive/tlpkg/TeXlive/TLConfig.pem

errikos:
- Remove irrelevant/redundant entries from installation profile
- Add hint to periodically update TeXLive packages
- This resolves #3460 and resolves #3461

No need to have all collections listed, as `selected_scheme scheme-full` covers them anyway.

Also, regarding #3461, `option_desktop_integration 1` is only relevant for Win32 systems, so we can just remove that option. Documentation and source packages are also removed.

### DIFF
--- a/docs/source/installation/latex.rst
+++ b/docs/source/installation/latex.rst
@@ -88,4 +88,4 @@ a good idea to periodically update the TeXLive packages by running:
 
 .. code-block:: shell
 
-    /opt/texlive/bin/x86_64-linux/tlmgr update --all
+    /opt/texlive/bin/x86_64-linux/tlmgr update --self --all
--- a/docs/source/installation/latex.rst
+++ b/docs/source/installation/latex.rst
@@ -33,69 +33,27 @@ Create the setup config file to install all the packages you need:
     selected_scheme scheme-full
     TEXDIR /opt/texlive
     TEXMFCONFIG ~/.texlive/texmf-config
-    TEXMFLOCAL /opt/texlive/texmf-local
     TEXMFHOME ~/texmf
+    TEXMFLOCAL /opt/texlive/texmf-local
     TEXMFSYSCONFIG /opt/texlive/texmf-config
     TEXMFSYSVAR /opt/texlive/texmf-var
     TEXMFVAR ~/.texlive/texmf-var
     binary_x86_64-linux 1
-    collection-basic 1
-    collection-bibtexextra 1
-    collection-binextra 1
-    collection-context 1
-    collection-fontsextra 1
-    collection-fontsrecommended 1
-    collection-fontutils 1
-    collection-formatsextra 1
-    collection-games 1
-    collection-humanities 1
-    collection-langarabic 1
-    collection-langchinese 1
-    collection-langcjk 1
-    collection-langcyrillic 1
-    collection-langczechslovak 1
-    collection-langenglish 1
-    collection-langeuropean 1
-    collection-langfrench 1
-    collection-langgerman 1
-    collection-langgreek 1
-    collection-langitalian 1
-    collection-langjapanese 1
-    collection-langkorean 1
-    collection-langother 1
-    collection-langpolish 1
-    collection-langportuguese 1
-    collection-langspanish 1
-    collection-latex 1
-    collection-latexextra 1
-    collection-latexrecommended 1
-    collection-luatex 1
-    collection-mathscience 1
-    collection-metapost 1
-    collection-music 1
-    collection-pictures 1
-    collection-plaingeneric 1
-    collection-pstricks 1
-    collection-publishers 1
-    collection-texworks 1
-    collection-xetex 1
-    option_adjustrepo 0
-    option_autobackup 1
-    option_backupdir tlpkg/backups
-    option_desktop_integration 1
-    option_doc 1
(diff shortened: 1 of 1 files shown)


ThiefMaster:
nice! could you try it out to make sure it doesn't break anything? ie install a local texlive copy using that profile and see if pdf generation in indico still works fine (ideally also including a bunch of math/science stuff in the test contribution that would break without some of the packages we don't list explicitly anymore)

errikos:
Everything seems to work fine. I think `scheme-full` installs literally everything, so no packages will actually be missing with the new profile. Also the options removed are mainly Windows-related, plus the package documentation and sources.

ThiefMaster:
looks good! but i just noticed that at least on debian you need to install `libfontconfig1` or xelatex fails even on a simple `/opt/texlive/bin/x86_64-linux/xelatex --version`; not sure about other distros.

maybe we could add a section to the latex docs suggesting to run it with `--version` and see if there are any errors, and install any missing packages if necessary. we could even mention `libfontconfig1` for debian.

ThiefMaster:
Unrelated to your change, but let's fix all texlive setup issues in this PR:

There's an error about `xelatex.fmt` missing, which can be solved with `/opt/texlive/bin/x86_64-linux/fmtutil-sys --all` Not sure if that's the proper command or if it does more than needed; but someone mentioned it [in the forum](https://talk.getindico.io/t/latexruntimeexception-after-upgrade-to-2-0-2/263/5?u=thiefmaster) and it works.

errikos:
This should not be needed, because I have set the `tlpdbopt_create_formats` option in the profile, which does exactly that (i.e. generates all necessary formats upon installation completion).

Check [install-tl](https://ctan.org/tex-archive/systems/texlive/tlnet/install-tl) line 1147.

docs/source/installation/latex.rst
```diff
@@ -115,3 +73,10 @@ your new TeXLive installation:
 
 If you are in a production setup, reload uWSGI using
 ``touch /opt/indico/web/indico.wsgi`` to reload the config file.
+
+As security-related updates are released frequently, it is also
+a good idea to periodically update the TeXLive packages by running:
+
+.. code-block:: shell
+
+    /opt/texlive/bin/x86_64-linux/tlmgr update --all
```
omasanori: It may be better to add `--self` to this `tlmgr` command to update `tlmgr` itself.

omasanori:
@errikos @ThiefMaster Concerning `fmtutil` problem, I think it is better to install prerequisites before installation of TeX Live than to run `fmutil` by hand. From [installation manual](https://www.tug.org/texlive/doc/texlive-en/texlive-en.html):
> 3.1.1 Unix
> (...)
> Special considerations for Cygwin: Unlike other Unix-compatible systems, Cygwin does not by default include all of the prerequisite programs needed by the TEX Live installer. See Section 3.1.4.
>
> 3.1.4 Cygwin
> Before beginning the installation, use Cygwin’s setup.exe program to install the perl and wget packages if you have not already done so. The following additional packages are recommended:
>
> - fontconfig [needed by XeTEX and LuaTEX]
> - ghostscript [needed by various utilities]
> - libXaw7 [needed by xdvi]
> - ncurses [provides the clear command used by the installer]

The problem is, that non-Cygwin Unix setup may sometimes lacks these packages too. I'd suggest to add some sentences to our manual before the installation step:

Install prerequisites needed by various commands in TeX Live:
```shell
yum install fontconfig ghostscript libXaw ncurses # CentOS / CC7
apt install libfontconfig1 ghostscript libxaw7 ncurses-bin # Debian / Ubuntu
```

ThiefMaster:
i think we only need fontconfig since we don't use any GUI or frontend stuff

omasanori:
@ThiefMaster Yes, you are right. libXaw and ncurses (maybe?) are unnecessary to us.

PR #3470, opened 2018-07-24
Rooms sprite sheet

mvidalgarcia:
- [x] Actually use the sprites in the new rb UI
- [x] Remove all `small_photo` related stuff


indico/modules/rb_new/controllers/backend.py
```diff
@@ -76,7 +81,16 @@ def _process(self, args):
         query = search_for_rooms(args, only_available=filter_availability)
         query = query.limit(args['limit']).offset(args['offset'])
         rooms, total = with_total_rows(query)
-        return jsonify(total=total, rooms=rooms_schema.dump(rooms).data)
+        if _cache.get('rooms-sprite') is None or _cache.get('rooms-sprite-mapping') is None:
```
ThiefMaster: let's not read the whole image from cache here. better to assume it exists if we have a mapping

indico/modules/rb_new/controllers/backend.py
```diff
@@ -76,7 +81,16 @@ def _process(self, args):
...
+            build_rooms_spritesheet()
+        sprite_mapping = _cache.get('rooms-sprite-mapping')
```
ThiefMaster: i wouldn't read from the cache twice; just assign it first and then use that to check if it's there

indico/modules/rb_new/controllers/backend.py
```diff
@@ -76,7 +81,16 @@ def _process(self, args):
...
+        rooms = rooms_schema.dump(rooms).data
+        for room in rooms:
+            try:
+                room['sprite_position'] = sprite_mapping.index(room['id']) + 1
```
ThiefMaster: shouldn't this be a property on the Room?

indico/modules/rb_new/controllers/backend.py
```diff
@@ -236,3 +250,10 @@ class RHRoomSuggestions(RHRoomBookingBase):
     def _process(self, args):
         return jsonify([dict(suggestion, room=rooms_schema.dump(suggestion['room'], many=False).data)
                         for suggestion in get_suggestions(args, limit=NUM_SUGGESTIONS)])
+
+
+class RHRoomsSprite(RHRoomBookingBase):
+    def _process(self):
+        photo_data = _cache.get('rooms-sprite')
+        io = BytesIO(photo_data)
+        return send_file('rooms-sprite.jpg', io, 'image/jpeg')
```
ThiefMaster: this should have strong caching headers (and a cache buster token in the url that's changed whenever the cache is invalidated)

indico/modules/rb_new/util.py
```diff
@@ -541,3 +549,36 @@ def serialize_nonbookable_periods(data):
...
     return [bookable_hours_schema.dump(d).data for d in data]
+
+
+def build_rooms_spritesheet():
+    def _resize(image):
+        return image.resize(ROOM_PHOTO_DIMENSIONS, Image.ANTIALIAS)
+
+    image_width, image_height = ROOM_PHOTO_DIMENSIONS
+    query = Room.query.filter(Room.photo)
+    room_count = query.count()
```
ThiefMaster: just get `.all()` from the rooms, since you get them all later anyway. no need to do a separate query for the count

indico/modules/rb_new/util.py
```diff
@@ -541,3 +549,36 @@ def serialize_nonbookable_periods(data):
...
+    mapping = []
```
ThiefMaster: wouldn't a dict be better for the mapping?

indico/modules/rb_new/util.py
```diff
@@ -541,3 +549,36 @@ def serialize_nonbookable_periods(data):
...
+    sprite_height = image_height
+    sprite = Image.new(
+        mode='RGB',
+        size=(sprite_width, sprite_height),
+        color=(0, 0, 0))
+
+    # Placeholder image at position 0
+    assets_dir = os.path.join(current_app.root_path, 'web', 'static')
+    no_photo_path = 'rooms/large_photos/NoPhoto.jpg'
+    no_photo_image = Image.open(os.path.join(assets_dir, config.IMAGES_BASE_URL[1:], no_photo_path))
```
ThiefMaster: `IMAGES_BASE_URL` is not a good idea here; it may contain  e.g. `/indico/` if someone uses a subdir. just use `os.path.join(current_app.root_path, 'web/static/images/rooms/....')` directly

indico/modules/rb_new/util.py
```diff
@@ -541,3 +549,36 @@ def serialize_nonbookable_periods(data):
 
 def serialize_unbookable_hours(data):
     return [bookable_hours_schema.dump(d).data for d in data]
+
+
+def build_rooms_spritesheet():
+    def _resize(image):
```
ThiefMaster: isn't it a bit of overkill to have a function for a single statement used only twice? ;)

indico/modules/rb_new/controllers/backend.py
mvidalgarcia: How do you know this token (e.g. a hash of the photo) from the client-side when fetching the url?

indico/migrations/versions/20180803_1703_93985a8c11ed_remove_room_thumbnail.py
```diff
@@ -0,0 +1,26 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.drop_column('photos', 'thumbnail', schema='roombooking')
+
+
+def downgrade():
+    op.add_column('photos', sa.Column('thumbnail', postgresql.BYTEA(), autoincrement=False, nullable=True),
```
ThiefMaster: `sa.LargeBinary` for the type, and the `autoincrement`/`nullable` kwargs are not needed (they're both the defaults)

indico/modules/rb/blueprints/user.py
```diff
@@ -29,8 +29,7 @@
 
 
 # Photos
-_bp.add_url_rule('/room/<roomLocation>/<int:roomID>/photo-<any(small,large):size>.jpg', 'photo',
-                 photo_handlers.room_photo)
+_bp.add_url_rule('/room/<roomLocation>/<int:roomID>/photo-large.jpg', 'photo', photo_handlers.room_photo)
```
ThiefMaster: just `photo.jpg`?

indico/modules/rb/controllers/admin/rooms.py
```diff
@@ -112,12 +114,14 @@ def _save(self):
         form.populate_obj(room, skip=('bookable_hours', 'nonbookable_periods'), existing_only=True)
         room.update_name()
         # Photos
-        if form.small_photo.data and form.large_photo.data:
-            _cache.delete_multi('photo-{}-{}'.format(room.id, size) for size in {'small', 'large'})
-            room.photo = Photo(thumbnail=form.small_photo.data.read(), data=form.large_photo.data.read())
+        if form.large_photo.data:
+            _cache.delete('photo-{}-large'.format(room.id))
```
ThiefMaster: the `-large` here is redundant since we don't have a small one in the cache

indico/modules/rb_new/blueprint.py
```diff
@@ -43,3 +43,4 @@
 _bp.add_url_rule('/api/booking/create', 'create_booking', backend.RHCreateBooking,
                  methods=('POST',))
 _bp.add_url_rule('/api/suggestions', 'suggestions', backend.RHRoomSuggestions)
+_bp.add_url_rule('/rooms-sprite__v<version>.jpg', 'sprite', backend.RHRoomsSprite)
```
ThiefMaster: I'd avoid the `__vXXX` since we use that for special handling in nginx rules.
Instead I'd go for a simple `rooms-sprite-<version>.jpg`.

indico/web/assets/vars_js.py
```diff
@@ -205,7 +209,8 @@ def generate_global_file():
         'Data': {
             'WeekDays': ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'],
             'DefaultLocation': default_location,
-            'Locations': location_names
+            'Locations': location_names,
+            'RoomsSpriteToken': _cache.get('rooms-sprite-token'),
```
ThiefMaster: this is not a good place for it. that file has extremely strong caching (indico version + indico.conf hash) so an updated token is not likely to show up in here anytime soon.

indico/web/client/js/legacy/libs/room_booking/room_selector.js
```diff
@@ -379,8 +379,9 @@
                     .children(":first-child").addClass("roomname")
                     .next().addClass("roomlocation");
 
+                var sprite = "<div class='room-sprite mini' style='background: url("+self.options.spriteURL+") -"+rooms[index].sprite_position * 290+"px 0'></div>";
```
ThiefMaster: what about a backtick template string?

PR #3473, opened 2018-07-26
Fix missing import in registration API

femtobit:
While using a local Indico installation (version 2.1.2), I noticed that check-in did not work with the Indico Check-In Android app from Google Play Store. This was due to a `NameError` being thrown in line 61 here:
https://github.com/indico/indico/blob/9fa826ebb672021b59ecc2744e98ad1eaaabb4b8/indico/modules/events/registration/api.py#L59-L61

Importing the `signals` model from `indico.core` fixes this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@ Bugfixes
 - Fix error when exporting abstracts with review questions to JSON
 - Point the timetable to correct day in the session details
 - Fix massive performance issue on the material package page in big events
+- Fix error when using the checkin app to mark someone as checked in
+  (:issue:`3473`, thanks :user:`femtobit`)
 
 Version 2.1.2
 -------------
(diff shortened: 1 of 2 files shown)


mic4ael:
I am curious how it got into the master. However this is a fix that should be merged to the branch `2-1.maintenance`

ThiefMaster:
Indeed, if you could rebase this PR onto `2.1-maintenance` and change the upstream branch of your PR as well that would be great. Since this justifies a 2.1.3 release, I'll take the opportunity during the next few days to see which other small fixes/improvements we have in master that might be worth backporting now that there will be a 2.1.3 after all.

In any case, thanks for spotting and fixing this!

femtobit:
> In any case, thanks for spotting and fixing this!

You're welcome. Thanks for the quick response.

As requested, I rebased unto `2.1-maintenance` and changed the base branch of this PR accordingly.

ThiefMaster:
Thanks! We'll include this in a 2.1.3 bugfix release.

PR #3475, opened 2018-07-26
[WIP] Improve registration im/export

bpedersen2:
Add an export that exports all fields and handle more fields in the import.

Currently the re-import is limited to the same regform, as way do not
have a generic field mapping system (see #3164 where a similar matching would
be needed).

TODO: verify and fix the different input types(may need  id's instead of text
at some places).

While not yet fully ready, I would like to get some comments if this is a useful approach.

### DIFF
--- a/indico/modules/events/registration/util.py
+++ b/indico/modules/events/registration/util.py
@@ -7,7 +7,7 @@
 
 import csv
 import itertools
-from collections import OrderedDict
+from collections import Counter, OrderedDict
 from operator import attrgetter
 
 from flask import current_app, json, session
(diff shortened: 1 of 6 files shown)


bpedersen2:
One use-case I currently have in mind:
 - export a bunch of registrations
 - import on a different form 
- delete registrations in first form

to move registrations from one form to another. Not yet possible without manual interventions as the field are different.

indico/modules/events/registration/util.py
```diff
@@ -523,33 +565,30 @@ def import_registrations_from_csv(regform, fileobj, skip_moderation=True, notify
...
+            if not regdata.get('first_name') or not regdata.get('last_name'):
+                raise UserValueError(_('Row {}: missing first or last name').format(row_num))
+            if regdata['email'] in registered_emails:
+                raise UserValueError(_('Row {}: a registration with this email already exists').format(row_num))
+            if regdata['email'] in used_emails:
+                raise UserValueError(_('Row {}: email address is not unique').format(row_num))
+
+            used_emails.add(regdata['email'])
+            todo.append(regdata)
     return [create_registration(regform, data, notify_user=notify_users, skip_moderation=skip_moderation)
```
pferreir: Will the user get a meaningful error if they e.g. provide a field with the wrong name?

bpedersen2: Probably not  yet, error handling is not completed. But that's not different from the current situation if you have at least one mandatory extra field or removed  e.g. phone from your regform.

bpedersen2:
Update to resolve merge conflicts. I still need to check if e.g.choice fields work correctly this way or if I need to match the choice display string to the choice id first,

bpedersen2:
Q: Should I  add a fallback to the old format if none of the headers match a field name?

indico/modules/events/registration/util.py
```diff
@@ -524,40 +567,73 @@ def update_regform_item_positions(regform):
...
-        todo.append({
-            'email': email,
-            'first_name': first_name.title(),
-            'last_name': last_name.title(),
-            'affiliation': affiliation,
-            'phone': phone,
-            'position': position
-        })
+    keys = []
+    for row_num, row in enumerate(reader):
```
ThiefMaster: insert of the big if/else inside the loop better go for something like this:

```
header_row = next(reader)
# ... handle the header data
for row_num, row in enumerate(reader, 1):
    ... # logic to handle the registrant data
```

indico/modules/events/registration/util_test.py
```diff
@@ -39,11 +39,48 @@ def dummy_regform(db, dummy_event):
...
+        'phone:Personal Data#Phone Number,'
+        'email:Personal Data#Email Address,'
+    ], [
+        u'Personal Data#First Name,'
+        'Personal Data#Last Name,'
+        'Personal Data#Affiliation,'
+        'Personal Data#Position,'
+        'Personal Data#Phone Number,'
+        'Personal Data#Email Address,'
+    ], [u'first_name,'
```
ThiefMaster: no need for the unicode prefix

indico/modules/events/registration/util.py
bpedersen2: good idea. I'll try to prepare something (it may take a bit as I am currently busy with other things),

bpedersen2:
Still open: mapping of e.g. Choice fields

PR #3479, opened 2018-08-01
Room Blockings

mic4ael:
- [x] agree on UI (placement of buttons etc.)
- [x] allow to search not only for users but also for groups in the Blocking Modal
- [x] improve the blockings list


indico/modules/rb_new/client/js/containers/BlockingModal.js
```diff
@@ -0,0 +1,32 @@
+/* This file is part of Indico.
```
pferreir: Same here.

indico/modules/rb_new/client/js/containers/BlockingList.js
```diff
@@ -0,0 +1,34 @@
+/* This file is part of Indico.
```
pferreir: In a past meeting, we agreed we'd put these in the same file as the component. Since this is new code, I'd suggest that we do that here.

indico/modules/rb_new/util.py
```diff
@@ -541,3 +542,33 @@ def serialize_nonbookable_periods(data):
...
+    blocking.reason = reason
+    blocking.created_by_user = session.user
+    blocking.allowed = allowed_principals
+    blocking.blocked_rooms = [BlockedRoom(room_id=room.id) for room in rooms]
+    db.session.add(blocking)
+    db.session.flush()
+    return blocking
+
+
+def approve_blocking(blocking):
```
pferreir: I understand the fact that this method will "approve" a blocking if the user is allowed. However, it's not always the case. What about `register_blocking(...)` or similar? or even `approve_or_request_blocking(...)`...

indico/web/client/js/react/components/PrincipalSearchField.jsx
```diff
@@ -77,13 +93,23 @@ export default class UserSearchField extends React.Component {
...
     }
 
-    renderUser({first_name: firstName, last_name: lastName, email, id}) {
-        return {
-            text: `${firstName} ${lastName}`,
-            description: email,
-            value: id,
-            key: id
-        };
+    renderItem({first_name: firstName, last_name: lastName, email, id, is_group: isGroup, name: groupName}) {
```
pferreir: Matter of personal preference, but I'd do it like this:
```js
renderItem(...) => (isGroup ? {
  // ...
} : {
  // ...
})
```

indico/web/client/js/react/components/PrincipalSearchField.jsx
```diff
@@ -94,12 +120,18 @@ export default class UserSearchField extends React.Component {
...
-        const users = await searchUser(query);
-        const options = users.map(this.renderUser);
-        this.userCache = Object.assign(this.userCache, ...users.map(user => ({[user.id]: user})));
+
+        const query = (searchQuery.indexOf('@') !== -1) ? {email: searchQuery} : {name: searchQuery};
+        const {withGroups} = this.props;
+        let items = [];
+
+        items = items.concat(await searchUser(query));
+        if (withGroups) {
```
pferreir: You are running the two searches sequentially. What about using `Promise.all(...)` to save time?

indico/modules/groups/controllers.py
```diff
@@ -140,3 +141,10 @@ class RHGroupDeleteMember(RHLocalGroupBase):
     def _process(self):
         self.group.members.discard(User.get(request.view_args['user_id']))
         return jsonify(success=True)
+
+
+class RHGroupSearch(RHProtected):
+    def _process(self):
```
pferreir: Maybe use `webargs` here to do some basic sanitization?

indico/modules/rb_new/client/js/actions.js
```diff
@@ -88,6 +90,16 @@ export const OPEN_FILTER_DROPDOWN = 'OPEN_FILTER_DROPDOWN';
 export const CLOSE_FILTER_DROPDOWN = 'CLOSE_FILTER_DROPDOWN';
 
 export const SET_BOOKING_AVAILABILITY = 'SET_BOOKING_AVAILABILITY';
+
+// Blockings
+export const FETCH_BLOCKINGS_STARTED = 'FETCH_BLOCKINGS_STARTED';
+export const FETCH_BLOCKINGS_FAILED = 'FETCH_BLOCKINGS_FAILED';
+export const UPDATE_BLOCKINGS = 'UPDATE_BLOCKINGS';
+export const BLOCKING_ONGOING = 'BLOCKING_ONGOING';
```
pferreir: `CREATE_BLOCKING_...` seems more consistent to me.

indico/modules/rb_new/client/js/components/RoomSelector.jsx
```diff
@@ -0,0 +1,191 @@
...
+                </List.Content>
+            </List.Item>
+        );
+    };
+
+    render() {
+        const {onChange} = this.props;
+        const {locations, selectedLocation, selectedRoom, selectedRooms: rooms, isFetchingLocations} = this.state;
+        const locationOptions = locations
+            .sort((a, b) => a.name.localeCompare(b.name))
```
pferreir: Cool. I didn't know about `localeCompare`... 😎

indico/modules/rb_new/client/js/components/RoomSelector.jsx
```diff
@@ -0,0 +1,191 @@
...
+                    <Grid.Column width={5}>
+                        <Dropdown placeholder={Translate.string('Location')}
+                                  options={locationOptions}
+                                  value={selectedLocation && selectedLocation.id}
+                                  loading={isFetchingLocations}
+                                  onChange={(event, data) => {
+                                      const selected = locations.find((loc) => loc.id === data.value);
+                                      this.setState({selectedLocation: selected});
+                                  }}
+                                  onOpen={() => {
```
pferreir: We're only loading the list of locations once the dropdown is opened. That results in a delay after clicking it. Since it seems quite cheap as a request, what about loading it in `componentDidMount(...)`?

indico/modules/groups/controllers.py
mic4ael: oh, I forgot about it..

indico/modules/rb_new/client/js/components/RoomSelector.jsx
mic4ael: ahh, no.. I forgot I did that initially, however React would complain about setting the state in the `componentDidMount` method and that is why I moved it to the `onOpen` callback

indico/modules/groups/controllers.py
```diff
@@ -140,3 +143,12 @@ class RHGroupDeleteMember(RHLocalGroupBase):
     def _process(self):
         self.group.members.discard(User.get(request.view_args['user_id']))
         return jsonify(success=True)
+
+
+class RHGroupSearch(RHProtected):
+    @use_args({
```
ThiefMaster: `use_kwargs`

indico/modules/groups/controllers.py
```diff
@@ -140,3 +143,12 @@ class RHGroupDeleteMember(RHLocalGroupBase):
...
+        'name': fields.Str()
+    })
+    def _process(self, args):
+        groups = GroupProxy.search(args['name'])
+        return jsonify([{'is_group': group.is_group, 'name': group.name, 'id': group.name} for group in groups])
```
ThiefMaster: why do we need `is_group`? that's always true for a group so let's just set it to `True`

indico/modules/rb_new/client/js/actions.js
```diff
@@ -88,6 +90,16 @@ export const OPEN_FILTER_DROPDOWN = 'OPEN_FILTER_DROPDOWN';
 export const CLOSE_FILTER_DROPDOWN = 'CLOSE_FILTER_DROPDOWN';
 
 export const SET_BOOKING_AVAILABILITY = 'SET_BOOKING_AVAILABILITY';
+
+// Blockings
+export const FETCH_BLOCKINGS_STARTED = 'FETCH_BLOCKINGS_STARTED';
+export const FETCH_BLOCKINGS_FAILED = 'FETCH_BLOCKINGS_FAILED';
+export const UPDATE_BLOCKINGS = 'UPDATE_BLOCKINGS';
+export const CREATE_BLOCKING = 'CREATE_BLOCKING';
+export const BLOCKING_CONFIRMED = 'BLOCKING_CONFIRMED';
```
ThiefMaster: what about using `<whatever>_(REQUEST|SUCCESS|ERROR)` for the request-related actions? especially here for blockings, CONFIRMED is very misleading. i just forgot to rename it for the booking actions (where it's also misleading, "confirmed" sounds like accepting a prebooking) when adding the ajax util and added the three names mentioned above for `GET_BOOKING_AVAILABILITY_*`

indico/modules/rb_new/client/js/actions.js
```diff
@@ -462,3 +474,43 @@ export function fetchBookingAvailability(room, filters) {
         ({availability, date_range: dateRange}) => ({availability: availability[room.id], dateRange})
     );
 }
+
+export function fetchBlockingsStarted() {
```
ThiefMaster: let's not add action creators for "internal" actions that are only dispatched from within another action. if there's an argument like a namespace it kind of makes sense (but even then i wouldn't *export* it) - but for something as simple as this one, just `dispatch({type: ...})` directly

indico/modules/rb_new/client/js/actions.js
```diff
@@ -462,3 +474,43 @@ export function fetchBookingAvailability(room, filters) {
...
+    return {type: FETCH_BLOCKINGS_FAILED};
+}
+
+export function updateBlockings(blockings) {
+    return {type: UPDATE_BLOCKINGS, blockings};
+}
+
+export function fetchBlockings() {
+    return async (dispatch, getStore) => {
+        dispatch(fetchBlockingsStarted());
```
ThiefMaster: see if you can use `ajaxAction` here

indico/modules/rb_new/client/js/components/RoomSelector.jsx
```diff
@@ -0,0 +1,195 @@
...
+import PropTypes from 'prop-types';
+import {Button, Dropdown, Grid, Icon, Image, List, Message, Segment} from 'semantic-ui-react';
+import getLocationsURL from 'indico-url:rooms_new.locations';
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+import {Translate} from 'indico/react/i18n';
+
+import './RoomSelector.module.scss';
+
+
+const fetchLocations = async () => {
```
ThiefMaster: shouldn't the locations go into the redux state?

indico/modules/rb_new/client/js/components/RoomSelector.jsx
```diff
@@ -0,0 +1,195 @@
...
+            .sort((a, b) => a.name.localeCompare(b.name))
+            .map((location) => {
+                return {text: location.name, value: location.id};
+            });
+        let roomOptions = [];
+        if (selectedLocation) {
+            roomOptions = selectedLocation.rooms
+                .filter((room) => rooms.findIndex((item) => item.id === room.id) === -1)
+                .sort((a, b) => a.full_name.localeCompare(b.full_name))
+                .map((room) => {
```
ThiefMaster: .map((room) => ({text: room.full_name, value: room.id}))

indico/modules/rb_new/client/js/components/RoomSelector.jsx
```diff
@@ -0,0 +1,195 @@
...
+                                      relaxed
+                                      selection>
+                                    {rooms.map(this.renderRoomItem)}
+                                </List>
+                            </Segment>
+                        )}
+                        {rooms.length === 0 && (
+                            <Message info>
+                                <Translate>
+                                    There are no rooms selected for this blocking!
```
ThiefMaster: `.` instead of `!`?

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -0,0 +1,253 @@
...
+import PrincipalSearchField from 'indico/react/components/PrincipalSearchField';
+import DatePeriodField from 'indico/react/components/DatePeriodField';
+import RoomSelector from '../RoomSelector';
+import {createBlocking} from '../../actions';
+
+
+function validate({period, reason, rooms}) {
+    const errors = {};
+    if (period === undefined || !period.length) {
+        errors.period = Translate.string('Please choose a valid period!');
```
ThiefMaster: I think `.` would be sufficient here (and below)

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -0,0 +1,253 @@
...
+        rooms: PropTypes.array,
+        createBlocking: PropTypes.func.isRequired
+    };
+
+    static defaultProps = {
+        rooms: []
+    };
+
+    createBlocking = async (formData) => {
+        await this.props.createBlocking(formData); // eslint-disable-line react/destructuring-assignment
```
ThiefMaster: check for the error handling logic in `submitBooking`. assuming you have matching field names here and in the webargs schema on the server, returning the error here will make final-form automatically show the server-side error on the corresponding field.

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -0,0 +1,253 @@
...
+                                            You can also block rooms you do not own - however, those blockings have to
+                                            be approved by the owners of those rooms.
+                                        </Translate>
+                                    </Message.Content>
+                                </Message>
+                                <Message color="red" icon>
+                                    <Icon name="warning sign" />
+                                    <Message.Content>
+                                        <Translate>
+                                            When blocking rooms nobody but you, the rooms' managers and those
```
ThiefMaster: almost the same message as above

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -0,0 +1,253 @@
...
+                                <Field name="reason"
+                                       label={Translate.string('Reason')}
+                                       component={ReduxFormField}
+                                       as={Form.TextArea}
+                                       format={formatters.trim}
+                                       placeholder={Translate.string('Reason for blocking')}
+                                       disabled={submitting || submitSucceeded}
+                                       formatOnBlur
+                                       required />
+                                {submitSucceeded && (
```
ThiefMaster: I think you can use `<Message success>` here which will only show up when the form has `success` set.

indico/modules/rb_new/client/js/components/modals/BookRoomModal.jsx
```diff
@@ -145,6 +145,7 @@ export default class BookRoomModal extends React.Component {
             const {booking} = this.state;
             const bookingLink = (
                 // TODO: add link to view booking details
+                // eslint-disable-next-line no-alert
```
ThiefMaster: i intentionally didn't disable the rule since we want to remove the alert eventually

indico/modules/rb_new/client/js/components/pages/RoomList.jsx
```diff
@@ -70,26 +76,67 @@ export default class RoomList extends React.Component {
...
 
-        return (
-            <Room key={room.id} room={room} showFavoriteButton>
-                <Slot name="actions">
-                    <Popup trigger={showDetailsBtn} content={Translate.string('Room details')} position="top center" />
-                </Slot>
-            </Room>
-        );
+        if (blockingMode) {
+            const isRoomInTheList = room.id in blockings;
```
ThiefMaster: `isRoomBlocked`? `InTheList` sounds way too verbose in a variable name ;)

indico/modules/rb_new/controllers/backend.py
```diff
@@ -239,3 +241,46 @@ class RHRoomSuggestions(RHRoomBookingBase):
     def _process(self, args):
         return jsonify([dict(suggestion, room=rooms_schema.dump(suggestion['room'], many=False).data)
                         for suggestion in get_suggestions(args, limit=NUM_SUGGESTIONS)])
+
+
+class RHRoomBlockings(RHRoomBookingBase):
+    @use_args({
```
ThiefMaster: `use_kwargs`

indico/modules/rb_new/controllers/backend.py
```diff
@@ -239,3 +241,46 @@ class RHRoomSuggestions(RHRoomBookingBase):
...
+
+
+class RHLocations(RHRoomBookingBase):
+    def _process(self):
+        locations = Location.query.all()
+        return jsonify(locations_schema.dump(locations).data)
+
+
+class RHCreateRoomBlocking(RHRoomBookingBase):
+    @use_args({
```
ThiefMaster: `use_kargs`

PR #3480, opened 2018-08-03
Add basic calendar page

pferreir:


indico/web/client/js/utils/redux.js
```diff
@@ -108,12 +108,14 @@ export function submitFormAction(submitFunc, requestAction, successAction, error
...
-        dispatcher(dispatch, requestAction);
+        if (requestAction) {
+            dispatcher(dispatch, requestAction);
+        }
         let response;
         try {
             response = await requestFunc();
         } catch (error) {
-            if (_.get(error, 'response.status') === 418) {
+            if (errorAction && _.get(error, 'response.status') === 418) {
```
ThiefMaster: That looks wrong. If you get an error back you still go into the else block where you try to dispatch the action that is not set. I would only make the dispatch conditional but keep the return anyway.

indico/modules/rb_new/controllers/backend.py
```diff
@@ -153,19 +164,27 @@ def _process(self, args):
...
             })
-        return jsonify_data(flash=False, availability=availability, date_range=date_range)
+
+        return jsonify_data(flash=False,
+                            availability=_serialize_availability(availability),
+                            date_range=date_range)
+
+
+class RHCalendar(RHRoomBookingBase):
+    @use_args({
```
ThiefMaster: `use_kwargs`? you get the two values separately from the dict anyway just below

indico/modules/rb_new/util.py
```diff
@@ -303,6 +304,47 @@ def get_rooms_availability(rooms, start_dt, end_dt, repeat_frequency, repeat_int
     return date_range, availability
 
 
+def get_room_calendar(start_dt, end_dt):
+    start_dt = start_dt.replace(hour=0, minute=0)
+    end_dt = end_dt.replace(hour=23, minute=59)
+
+    query = (db.session.query(ReservationOccurrence)
```
ThiefMaster: `ReservationOccurrence.query`

indico/modules/rb_new/controllers/backend.py
pferreir: Didn't know that exists, thanks!

pferreir:
My latest push also orders the room in natural order.

mvidalgarcia:
According to Travis there's a blank line missing in `update_backrefs.py` after the imports :thinking:

pferreir:
@mvidalgarcia there was a problem sorting an import (uppercase comes first). Fixed now.

PR #3492, opened 2018-08-10
Unavailable periods in usage & booking modal

### LINKED ISSUE #3484: Display unavailable times in "room usage" and timeline in "book a room" modal

mvidalgarcia:
Closes #3484

### DIFF
--- a/indico/modules/rb_new/client/js/components/TimelineLegend.jsx
+++ b/indico/modules/rb_new/client/js/components/TimelineLegend.jsx
@@ -1,0 +1,65 @@
+/* This file is part of Indico.
+ * Copyright (C) 2002 - 2018 European Organization for Nuclear Research (CERN).
+ *
+ * Indico is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Indico is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import PropTypes from 'prop-types';
+import React from 'react';
+import {Label, Segment, List} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+
+import './TimelineLegend.module.scss';
+
+
+export default function TimelineLegend({labels, aside, compact}) {
+    return compact ? (
+        <List styleName="legend compact">
+            {labels.map(({label, color, style}) => (
+                <List.Item key={label}>
+                    <List.Content styleName="labels">
+                        <Label color={color} styleName={`compact ${style || ''}`} />
+                        <span styleName="text">{Translate.string(label)}</span>
+                    </List.Content>
+                </List.Item>
+            ))}
+        </List>
+    ) : (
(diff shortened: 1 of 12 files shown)


indico/modules/rb_new/client/js/components/TimelineLegend.jsx
```diff
@@ -0,0 +1,65 @@
...
+                ))}
+            </Label.Group>
+            {aside}
+        </Segment>
+    );
+}
+
+TimelineLegend.propTypes = {
+    labels: PropTypes.arrayOf(
+        PropTypes.shape({
```
ThiefMaster: same shape as above. define once as `labelShape` and use that here and above?

indico/modules/rb_new/util.py
```diff
@@ -348,6 +348,28 @@ def get_room_calendar(start_dt, end_dt):
...
+    nonbookable_periods = get_rooms_nonbookable_periods([room], start_dt, end_dt).get(room.id, [])
+
+    availability = []
+    for dt in dates:
+        iso_dt = dt.isoformat()
+        availability.append({
+            'bookings': serialize_occurrences(group_by_occurrence_date(bookings)).get(iso_dt),
+            'blockings': serialize_blockings(group_blockings(blockings, dates)).get(iso_dt),
+            'nonbookable_periods': serialize_nonbookable_periods(
+                group_nonbookable_periods(nonbookable_periods, dates)).get(iso_dt),
```
ThiefMaster: do we really want None in case the date has no entry and not e.g. an empty list?

indico/modules/rb_new/util.py
```diff
@@ -348,6 +348,28 @@ def get_room_calendar(start_dt, end_dt):
...
+    dates = [d.date() for d in iterdays(start_dt, end_dt)]
+
+    bookings = get_existing_room_occurrences(room, start_dt, end_dt, RepeatFrequency.DAY, 1, only_accepted=True)
+    blockings = get_rooms_blockings([room], start_dt.date(), end_dt.date()).get(room.id, [])
+    unbookable_hours = get_rooms_unbookable_hours([room]).get(room.id, [])
+    nonbookable_periods = get_rooms_nonbookable_periods([room], start_dt, end_dt).get(room.id, [])
+
+    availability = []
+    for dt in dates:
+        iso_dt = dt.isoformat()
```
ThiefMaster: `dt` is usually a datetime object. what about `day` instead?

indico/modules/rb_new/client/js/components/TimelineLegend.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+export default function TimelineLegend({labels, aside, compact}) {
+    return compact ? (
+        <List styleName="legend compact">
+            {labels.map(({label, color, style}) => (
+                <List.Item key={label}>
+                    <List.Content styleName="labels">
+                        <Label color={color} styleName={`compact ${style || ''}`} />
+                        <span styleName="text">{Translate.string(label)}</span>
+                    </List.Content>
```
mvidalgarcia: I guess this `Translate` is not needed then?
Thanks for taking care of the PRs :wink:

ThiefMaster: That would actually break during extraction since the value to translation functions cannot be dynamic. thanks for pointing it out, didn't notice it

ThiefMaster: Fixed. Actually found a bug in the i18n lib while doing so ;)

Tip: `indico i18n extract_messages_react` spots such problems since we fail with an error in case of invalid translation logic calls such as passing a dynamic value to `Translate.string()`. We don't keep the pot file updated for now, but running the command and then reverting the change to the pot file afterwards is still useful to check that everything is valid.

PR #3495, opened 2018-08-17
Edit/View a blocking

mic4ael:


indico/modules/rb_new/util.py
```diff
@@ -638,19 +640,35 @@ def get_room_blockings(start_dt=None, end_dt=None, created_by=None, in_rooms_own
     return query.all()
 
 
-def create_blocking(rooms, start_date, end_date, reason, allowed_principals):
+def _populate_blocking(blocking, room_ids, reason, allowed_principals, start_dt=None, end_dt=None):
+    blocking_data = {'reason': reason, 'created_by_user': session.user,
+                     'allowed': [GroupProxy(pr['id'], provider=pr['provider'])
+                                 if pr.get('is_group')
+                                 else User.query.filter(User.id == pr['id']).one()
```
ThiefMaster: `User.get_one(...)`

indico/modules/rb_new/util.py
```diff
@@ -638,19 +640,35 @@ def get_room_blockings(start_dt=None, end_dt=None, created_by=None, in_rooms_own
...
+                                 for pr in allowed_principals],
+                     'blocked_rooms': [BlockedRoom(room_id=room.id)
+                                       for room in Room.query.filter(Room.id.in_(room_ids))]}
+
+    if start_dt:
+        blocking_data['start_date'] = start_dt
+    if end_dt:
+        blocking_data['end_date'] = end_dt
+    blocking.populate_from_dict(blocking_data)
```
ThiefMaster: Why do we build a dict and populate from that instead of simply assigning things directly?

indico/web/client/js/react/components/PrincipalSearchField.jsx
```diff
@@ -54,40 +54,67 @@ const searchGroups = async (data) => {
...
-        const {defaultValue} = this.props;
-        const userId = defaultValue && defaultValue.id;
+
+        const {multiple, onChange} = this.props;
+        let {value} = this.props;
+        let dropdownValue;
+
+        if (multiple) {
+            value = value || [];
+            dropdownValue = value && value.length !== 0 ? value.map((val) => val.id) : [];
```
ThiefMaster: `value &&` isn't needed since you ensure it's an empty list in case of anything falsy one line above.

indico/modules/rb_new/blueprint.py
```diff
@@ -47,5 +47,6 @@
 _bp.add_url_rule('/api/blockings', 'blockings', backend.RHRoomBlockings)
 _bp.add_url_rule('/api/locations', 'locations', backend.RHLocations)
 _bp.add_url_rule('/api/blockings/create', 'create_blocking', backend.RHCreateRoomBlocking, methods=('POST',))
+_bp.add_url_rule('/api/blockings/update', 'update_blocking', backend.RHUpdateRoomBlocking, methods=('PATCH',))
```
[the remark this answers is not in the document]
mic4ael: So obvious. I think I was so occupied by making this all work I forgot to change the initial version of this `RH`..

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -44,124 +51,203 @@ function validate({period, reason, rooms}) {
...
+            reason: PropTypes.string,
+            created_by_id: PropTypes.number
+        })
     };
 
     static defaultProps = {
-        rooms: []
+        open: false,
+        mode: 'new',
+        blocking: {
```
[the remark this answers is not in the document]
mic4ael: Yeah, we do, for example for the `blocked_rooms`. I'd rather end up with an empty array rather than null or undefined (which would cause errors).

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -44,124 +51,203 @@ function validate({period, reason, rooms}) {
...
         onClose: PropTypes.func.isRequired,
-        rooms: PropTypes.array,
-        createBlocking: PropTypes.func.isRequired
+        createBlocking: PropTypes.func.isRequired,
+        updateBlocking: PropTypes.func.isRequired,
+        fetchBlockings: PropTypes.func.isRequired,
+        user: PropTypes.object.isRequired,
+        open: PropTypes.bool,
+        mode: PropTypes.oneOf(['read', 'edit', 'new']),
+        blocking: PropTypes.exact({
```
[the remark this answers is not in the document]
mic4ael: Don't remember now what it was exactly but I was trying to find a right `PropType` for something and I guess I didn't change it back to `shape`.

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -44,124 +51,203 @@ function validate({period, reason, rooms}) {
...
+            id: PropTypes.number,
+            blocked_rooms: PropTypes.array,
```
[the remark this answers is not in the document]
mic4ael: Nope, since this `Modal` can be used also to create a blocking which doesn't have `id` etc.

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -44,124 +51,203 @@ function validate({period, reason, rooms}) {
...
         const {form: {mutators}, submitting, submitSucceeded} = fprops;
+        const {mode} = this.state;
+        const formProps = mode === 'read' ? {} : {onSubmit: fprops.handleSubmit, success: submitSucceeded};
+        const canEdit = !!blocking.id && user.id === blocking.created_by_id && mode !== 'edit';
 
         // set `touched` flag so in case of a validation error we properly
         // show the error label
         mutators.setFieldTouched('rooms', true);
+        if (mode === 'edit') {
+            mutators.setFieldTouched('dates', true);
```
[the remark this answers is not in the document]
ThiefMaster: Basically what we discussed offline yesterday: The field should trigger being marked as touched when being focused and defocused by the user.

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -188,36 +274,53 @@ class BlockingModal extends React.Component {
...
                                        format={formatters.trim}
-                                       placeholder={Translate.string('Reason for blocking')}
-                                       disabled={submitting || submitSucceeded}
-                                       formatOnBlur
-                                       required />
+                                       render={(fieldProps) => {
+                                           const props = {};
+                                           if (mode === 'edit') {
+                                               props.defaultValue = blocking.reason;
+                                               props.value = undefined;
```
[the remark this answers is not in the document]
ThiefMaster: That seems strange...

```
> foo = {}
> foo.bar === undefined
true
```

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -44,124 +51,203 @@ function validate({period, reason, rooms}) {
...
+        const {createBlocking, updateBlocking, fetchBlockings, blocking: {id}} = this.props;
+        const {mode} = this.state;
+        let rv;
+
+        if (mode === 'new') {
+            rv = await createBlocking(formData);
+        } else if (mode === 'edit') {
+            rv = await updateBlocking({...formData, blockingId: id});
+            if (!rv.error) {
+                fetchBlockings();
```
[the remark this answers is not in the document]
ThiefMaster: Do we do some grouping on the server side? IMHO the API call to get the blockings should simply return a list of all the blockings - and in that case, adding the new blocking to that list after creating it will result in the same state as you would get after fetching all theblockings.

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -229,14 +332,26 @@ class BlockingModal extends React.Component {
     };
 
     render() {
-        const {open, onClose} = this.props;
+        const {
+            open, onClose,
+            blocking: {blocked_rooms: blockedRooms, allowed, start_date: startDate, end_date: endDate, reason}
+        } = this.props;
+        const {mode} = this.state;
+        const props = mode === 'read' ? {onSubmit() {}} : {validate, onSubmit: this.processBlocking};
```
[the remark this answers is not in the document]
mic4ael: `onSubmit` is apparently required and can't be undefined. Unfortunately wrapping it with the `FinalForm` is also necessary since when rendering the form I depend on some of the properties the rendering function receives.

indico/modules/rb_new/client/js/components/modals/BlockingModal.jsx
```diff
@@ -44,124 +51,203 @@ function validate({period, reason, rooms}) {
...
+    };
+
+    hasAllowedFieldChanged = (prevValue, nextValue) => {
+        if (!prevValue || prevValue.length !== nextValue.length) {
+            return true;
+        }
+
+        prevValue = _.sortBy(prevValue, 'id');
+        nextValue = _.sortBy(nextValue, 'id');
+        return !_.every(nextValue, (val, index) => val.id === prevValue[index].id);
```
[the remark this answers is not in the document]
mic4ael: Not exactly, it is because when viewing a specific blocking details I initialize the modal with principals serialized by marshmallow and at the same time data inside the `PrincipalSearchField` comes from a different endpoint, thus returning slightly different data. That makes comparison fail even if we compare data for the same principal.

indico/modules/groups/controllers.py
```diff
@@ -152,5 +152,6 @@ class RHGroupSearch(RHProtected):
     def _process(self, name):
         groups = GroupProxy.search(name)
         return jsonify([{'is_group': True, 'name': group.name, 'id': getattr(group, 'id', group.name),
-                         'provider': group.provider}
+                         'provider': group.provider,
+                         'identifier': 'Group:{}:{}'.format(group.provider or '', getattr(group, 'id', group.name))}
```
ThiefMaster: what about using the data in `group.as_principal` to build this identifier string?

indico/modules/rb_new/controllers/backend.py
```diff
@@ -298,26 +298,38 @@ def _process(self):
...
-        rooms = Room.query.filter(Room.id.in_(room_ids)).all()
-        allowed = []
-        for obj in allowed_principals:
-            if obj.get('is_group'):
-                allowed.append(GroupProxy(obj['id'], provider=obj['provider']))
-            else:
-                allowed.append(User.query.filter(User.id == obj['id']).one())
-        blocking = create_blocking(rooms, start_date, end_date, reason, allowed)
+    def _process(self, args):
+        blocking = create_blocking(args)
```
ThiefMaster: I know we do it in many other places, but I think passing/accepting an `args` dict in an operation function is not great - I'd rather have the operation take explicit arguments and then unpack `**args` here.

When just accepting a dict of "args" in the operation you need to go into the modal again to see what the operation actually accepts...

(nothing that needs to be changed now, just something that came to my mind a few times now when seeing operation functions)

indico/modules/rb_new/controllers/backend.py
```diff
@@ -298,26 +298,38 @@ def _process(self):
...
+        self.blocking = Blocking.get_one(request.view_args['blocking_id'])
+
+    @use_args({
+        'room_ids': fields.List(fields.Int(), required=True),
+        'reason': fields.Str(required=True),
+        'allowed_principals': fields.List(fields.Dict(), missing=[])
+    })
+    def _process(self, args):
+        update_blocking(self.blocking, args)
+        return jsonify_data(flash=False, blocking=blockings_schema.dump(self.blocking, many=False).data)
```
ThiefMaster: just use `jsonify()`; we don't include flashes anyway and the code doesn't check for a `.success` attribute (which is the only thing `jsonify_data(flash=False)` adds compared to plain `jsonify`)

indico/modules/rb_new/schemas.py
```diff
@@ -82,19 +84,43 @@ class Meta:
...
+    email = String(missing=None)
+    is_group = Boolean(missing=False)
+
+    class Meta:
+        model = BlockingPrincipal
+        fields = ('id', 'identifier', 'name', 'is_group', 'email', 'full_name', 'provider')
+
+    def get_identifier(self, blocking_principal):
+        if blocking_principal.is_group:
+            return 'Group:{}:{}'.format(blocking_principal.provider or '',
```
ThiefMaster: why not add a property for this on BlockingPrincipal?

or maybe a util function so you can use it both for the principal and the groupproxy itself?

indico/modules/rb_new/schemas.py
```diff
@@ -82,19 +84,43 @@ class Meta:
...
+                                        getattr(blocking_principal, 'id', blocking_principal.name))
+        else:
+            return 'User:{}'.format(blocking_principal.id)
+
+
 class BlockingSchema(mm.ModelSchema):
     blocked_rooms = Nested(BlockedRoomSchema, many=True)
+    allowed = Nested(BlockingPrincipalSchema, many=True)
+    can_edit = Method('get_can_edit')
```
ThiefMaster: What do you think about using `Function(lambda blocking: blocking.can_be_modified(session.user))` instead?

indico/modules/rb_new/util.py
```diff
@@ -638,19 +641,43 @@ def get_room_blockings(start_date=None, end_date=None, created_by=None, in_rooms
...
     blocking.created_by_user = session.user
-    blocking.allowed = allowed_principals
-    blocking.blocked_rooms = [BlockedRoom(room_id=room.id) for room in rooms]
+    blocking.reason = reason
+    blocking.allowed = [GroupProxy(_group_id_or_name(pr), provider=pr['provider'])
+                        if pr.get('is_group')
+                        else User.get_one(pr['id'])
+                        for pr in allowed_principals]
+
+    if start_date:
```
ThiefMaster: I would handle those in `create_blocking` (`blocking.start_date = blocking_data.pop('start_date')`)

indico/modules/users/schemas.py
```diff
@@ -16,14 +16,21 @@
...
 class UserSchema(mm.ModelSchema):
+    identifier = Method('get_identifier')
+
     class Meta:
         model = User
-        fields = ('id', 'first_name', 'last_name', 'email', 'affiliation', 'avatar_bg_color')
+        fields = ('id', 'identifier', 'first_name', 'last_name', 'email', 'affiliation', 'avatar_bg_color', 'full_name')
+
+    def get_identifier(self, user):
+        return 'User:{}'.format(user.id)
```
ThiefMaster: `Function(lambda...)`?

indico/web/client/js/react/components/PrincipalSearchField.jsx
```diff
@@ -139,39 +167,46 @@ export default class PrincipalSearchField extends React.Component {
 
     render() {
         const {multiple, onChange, disabled} = this.props;
-        const {isFetching, options, userId} = this.state;
+        const {isFetching, options, value} = this.state;
         let dropdownValues, selectedValues, dropdownOptions;
+        let placeholder = Translate.string('Write a name (of the group or user) or e-mail...');
```
ThiefMaster: But that's not always the case. Unless withGroups is used the message shouldn't mention groups.

indico/modules/rb_new/schemas.py
mic4ael: yeah, why not, no strong preferences

indico/modules/rb_new/client/js/components/filters/DateForm.jsx
```diff
@@ -90,7 +92,7 @@ export default class DateForm extends FilterFormComponent {
                                        await this.setDates(start, end);
                                        handleClose();
                                    }}
-                                   disabledDate={this.disabledDate}
+                                   disabledDate={disabledDate ? disabledDate : this.disabledDate}
```
ThiefMaster: `{disabledDate || this.disabledDate}`

indico/modules/rb_new/client/js/modules/blockings/serializers.js
```diff
@@ -27,6 +27,17 @@ export const ajax = {
...
         serializer: ({myBlockings}) => myBlockings,
     },
+    room_ids: {
+        onlyIf: ({rooms}) => rooms && rooms.length,
+        serializer: ({rooms}) => rooms.map((room) => room.id)
+    },
+    allowed_principals: {
+        onlyIf: ({allowed}) => !!allowed,
+        serializer: ({allowed}) => allowed.map((obj) => ({
+            id: obj.id, is_group: obj.isGroup, provider: obj.provider, name: obj.name
```
ThiefMaster: more readable with one item per line IMHO

indico/modules/rb_new/client/js/util.jsx
```diff
@@ -205,3 +205,20 @@ export function boolStateField(name) {
 export function isDateWithinRange(date, dateRange, _toMoment) {
     return date && dateRange.filter((dt) => _toMoment(dt).isSame(date, 'day')).length !== 0;
 }
+
+export function camelizeKeys(obj) {
+    if (_.isArray(obj)) {
+        return obj.map((item) => camelizeKeys(item));
```
ThiefMaster: `return obj.map(camelizeKeys);`

indico/modules/rb_new/client/js/util.jsx
```diff
@@ -205,3 +205,20 @@ export function boolStateField(name) {
...
+    }
+
+    const newObject = {};
+    for (const [key, value] of Object.entries(obj)) {
```
ThiefMaster: more-magic version would be like this:

```
const newObject = Object.entries.reduce((obj, [key, value]) => {
    return {...obj, [newKey]: camelizeKeys(value)}
}, {});
```

but regardless, i would support non-object/array values in `camelizeKeys` (and just return the input in that case); that way you don't need the extra check but can always call it

indico/modules/rb_new/util.py
```diff
@@ -638,19 +641,40 @@ def get_room_blockings(start_date=None, end_date=None, created_by=None, in_rooms
...
+    if provider is None or provider == 'indico':
+        return principal['id']
+    else:
+        return principal['name']
+
+
+@no_autoflush
+def _populate_blocking(blocking, room_ids, allowed_principals, reason):
+    blocking.blocked_rooms = [BlockedRoom(room_id=room.id) for room in Room.query.filter(Room.id.in_(room_ids))]
+    blocking.created_by_user = session.user
```
ThiefMaster: operations should not access the session whenever possible

indico/web/client/js/react/components/PrincipalSearchField.jsx
```diff
@@ -48,61 +49,88 @@ const searchGroups = async (data) => {
...
         };
-        this.userCache = {};
+
+        if (multiple) {
+            this.userCache = Object.assign({}, ...value.map((val) => ({[val.identifier]: val})));
+        } else {
+            this.userCache = {};
+        }
+
+        onChange(value);
```
ThiefMaster: @mic4ael do you remember why these calls to the onChange callback are/were needed?

mic4ael: yeah, I think in this case it is needed when editing a booking since it sets the initial value on the instance of the form which allows to properly perform later validations

PR #3502, opened 2018-08-24
Fix empty calendar bug

### LINKED ISSUE #3499: Empty calendar when using back button.
To reproduce: Go to the calendar, click some other page in the top menu, then use your browser's back button to go back to the calendar.

meluru:
Closes #3499

### DIFF
--- a/indico/modules/rb_new/client/js/components/pages/Calendar.jsx
+++ b/indico/modules/rb_new/client/js/components/pages/Calendar.jsx
@@ -54,7 +54,10 @@ class Calendar extends React.Component {
         // this updates the calendar when
         // 1) the query string changes due to the user changing the date
         // 2) the user clicks on "calendar" again (which also causes a query string change)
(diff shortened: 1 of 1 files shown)


indico/modules/rb_new/client/js/components/pages/Calendar.jsx
```diff
@@ -63,9 +63,8 @@ class Calendar extends React.Component {
         const {date, fetchCalendar, setDate} = this.props;
         if (!date) {
             setDate(moment());
-        } else {
-            fetchCalendar();
         }
+        fetchCalendar();
```
ThiefMaster: Are you sure this won't dispatch the fetch twice? See my comment [down here](https://github.com/indico/indico/blob/cdfb1c9c12bee72103e07e3d332f4aa77c91edf7/indico/modules/rb_new/client/js/components/pages/Calendar.jsx#L120-L121)

PR #3516, opened 2018-09-10
Add edit button to conference pages

### LINKED ISSUE #3284: Quicker way to edit a conference page
Currently, to edit a conference page (https://indico-server/event/XXXX/page/YYYY-page-name), one has to take several steps/clicks:  Manage -> Menu -> find page -> pen icon. 

It would be nice to have a more straightforward way to edit a page. E.g. if I am manager of the event, currently logged it, I'd be happy to see this "pen" icon on the page itself. 

(Same could apply for other page types, e.g. overview page, timetable etc.)

panagiotappl:
closes #3284

### DIFF
--- a/indico/modules/events/layout/templates/page.html
+++ b/indico/modules/events/layout/templates/page.html
@@ -5,6 +5,19 @@
     {{ page.menu_entry.title }}
 {% endblock %}
 
+{% block page_actions %}
+    {% if event.can_manage(session.user) %}
+        <div class="toolbar">
+            <a href="#"
+               class="i-button icon-edit"
+               data-href="{{ url_for('event_layout.menu_entry_edit', page.menu_entry) }}"
+               data-title="{% trans %}Edit page{% endtrans %}"
+               data-ajax-dialog
+               data-reload-after></a>
+        </div>
+    {% endif %}
+{% endblock %}
+
 {% block content %}
     {{ page.html|sanitize_html }}
 {% endblock %}
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,7 @@ Improvements
 - Make registration period display less confusing (:issue:`3359`)
 - Add setting for the default contribution duration of an event
   (:issue:`3446`)
+- Add edit button to custom conference pages (:issue:`3284`)
 
 Internal Changes
 ^^^^^^^^^^^^^^^^


indico/modules/events/layout/templates/page.html
```diff
@@ -3,6 +3,15 @@
 {% block title %}
     {# The title of a user-generated page isn't translated, localized title is not needed #}
     {{ page.menu_entry.title }}
+    {% if event.can_manage(session.user) %}
+    <a href="#"
```
mic4ael: You should fix the indentation of this whole block ;)

indico/modules/events/layout/templates/page.html
```diff
@@ -3,6 +3,15 @@
...
+        class="icon-edit"
+        data-href="{{ url_for('event_layout.menu_entry_edit', page.menu_entry) }}"
+        data-ajax-dialog
+        data-reload-after
+    ></a>
```
mic4ael: the closing `>` could be places at the end of `data-reload-after`

indico/modules/events/layout/templates/page.html
```diff
@@ -3,6 +3,15 @@
 {% block title %}
     {# The title of a user-generated page isn't translated, localized title is not needed #}
     {{ page.menu_entry.title }}
+    {% if event.can_manage(session.user) %}
+    <a href="#"
+        class="icon-edit"
+        data-href="{{ url_for('event_layout.menu_entry_edit', page.menu_entry) }}"
```
mic4ael: I think you can omit `event_layout` here

indico/modules/events/layout/templates/page.html
ThiefMaster: also of the attributes - check other templates, we align the attributes like this:

```
<a href="..."
   data-foo="bar">
```
```
<span class="..."
      whatever="foobar">
```

indico/modules/events/layout/templates/page.html
ThiefMaster: When called only from within the same blueprint you can use the relative version. However, this is not the case here since you can mark a custom page as the default event homepage and in that case it will be rendered from a different blueprint.

ThiefMaster:
Looks good now, can you also add a changelog entry (CHANGES.rst)? This would be for 2.2, in the "improvements section":

    - Add edit button to custom conference pages (:issue:`3284`)

PR #3529, opened 2018-09-13
Infinite scrolling in Booking Timeline

mic4ael:


indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -82,8 +82,9 @@ def _process(self, args):
                 'all_days_available': not data['conflicts']
             })
 
+        serialized = _serialize_availability(availability)
         return jsonify_data(flash=False,
-                            availability=_serialize_availability(availability),
+                            availability=list(serialized.viewitems()),
```
ThiefMaster: `serialized.items()` already gives you a list (and in python 3 either option would have to be changed, since viewitems goes away then)

PR #3530, opened 2018-09-13
Fix participant roles table

### LINKED ISSUE #3511: Table layout looks broken after editing person data in "participant roles"
image

This happens only if at least one participant is registered (so the registration indicator icon shows up)

panagiotappl:
Table layout does not break after editing a person and there is at least one
registration icon showing

closes #3511

### DIFF
--- a/indico/htdocs/js/indico/modules/events/management.js
+++ b/indico/htdocs/js/indico/modules/events/management.js
@@ -110,6 +110,20 @@
         $('.js-reset-role-filter').toggleClass('disabled', isInitialState);
     }
 
+    function initTooltip() {
+        $('.js-show-regforms').qtip({
+            content: {
+                text: function() {
+                    return $(this).data('title');
+                }
+            },
+            hide: {
+                delay: 100,
+                fixed: true
+            }
+        });
+    }
+
     global.setupEventPersonsList = function setupEventPersonsList(options) {
         options = $.extend({
             hasNoAccountFilter: false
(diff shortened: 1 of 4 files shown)


indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -110,6 +110,20 @@
         $('.js-reset-role-filter').toggleClass('disabled', isInitialState);
     }
 
+    function initTooltip() {
```
ThiefMaster: I would move this inside `setupEventPersonsList` since it's related to that (the whole management.js is a bit ugly/big, but not worth splitting/refactoring it now)

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -248,6 +252,12 @@
             $checkbox.prop('checked', !$checkbox.prop('checked')).trigger('change');
             toggleResetBtn();
         });
+
+        var $eventParticipantsList = $('#event-participants-list');
```
ThiefMaster: since you use it only once, no need to assign it first. btw, let's use `const` in new code, even when adding it in an old file

indico/htdocs/js/indico/modules/events/management.js
```diff
@@ -248,6 +252,12 @@
...
+
+        $eventParticipantsList.on('indico:htmlUpdated', function () {
```
ThiefMaster: This could be simplified like this (we don't put a space between `function` and `()` btw):

    $('#event-participants-list').on('indico:htmlUpdated', initTooltip);

PR #3536, opened 2018-09-17
Couple of fixes

mic4ael:


indico/modules/rb_new/client/js/components/TimelineContent.module.scss
```diff
@@ -98,5 +98,14 @@ $row-height: 50px;
...
     & + .timeline-loader {
         margin-top: 2em !important;
+        color: rgba(0, 0, 0, 0.87) !important;
+
+        &::before {
+            border-color: rgba(0, 0, 0, 0.1) !important;
+        }
+
+        &::after {
+            border-color: #767676 transparent transparent !important;
```
ThiefMaster: `$sui-grey`

indico/modules/rb_new/client/js/reducers/roomBooking/roomList.js
```diff
@@ -28,11 +28,11 @@ const initialRoomsState = {
...
 
 function mergeRooms(state, action) {
-    const {list: oldList} = state;
-    const {loadMore, matching, total, rooms: newList} = action;
+    const {list: oldList, total: stateTotal} = state;
+    const {loadMore, matching, rooms: newList, total} = action;
     return {
         matching,
-        total,
+        total: total === null ? stateTotal : total,
```
meluru: Is there a reason why you comapre to `null` instead of just `total ? total : stateTotal`

ThiefMaster: `total === 0` in case on on results

mic4ael: yes, `total ?  total : stateTotal` will return `stateTotal` for `total = 0` which I don't want. This whole change is based on the fact that we calculate `total` on the backend only on the first request. After that we don't return `total` thus `null`.

meluru: Okay :) thanks

PR #3541, opened 2018-09-18
Suggestions fix

mic4ael:


indico/modules/rb_new/client/js/modules/bookRoom/selectors.js
```diff
@@ -23,6 +23,6 @@ export const isFetchingFormTimeline = ({bookRoom}) => {
     return bookRoom.bookingForm.requests.timeline.state === RequestState.STARTED;
 };
 export const getUnavailableRoomInfo = ({bookRoom: {unavailableRooms}}) => unavailableRooms.availability;
-export const getTimelineDateRange = ({bookRoom: {timeline}}) => timeline.dateRange;
+export const getUnavailableTimelineDateRange = ({bookRoom: {unavailableRooms}}) => unavailableRooms.dateRange;
```
ThiefMaster: this change doesn't seem related to your fix; can we skip this (and the change in UnavailableRoomsModal) since it will already be changed in my PR (to take it from the meta data i return together with the search results - the "feels slightly dirty" one I mentioned in the meeting today)

The range for unavailable rooms is always the same as the range we searched for initially, right?

indico/modules/rb_new/operations/suggestions.py
```diff
@@ -40,6 +43,11 @@ def get_suggestions(filters, limit=None):
     query = search_for_rooms(room_filters)
     query = query.filter(~Room.filter_available(filters['start_dt'], filters['end_dt'],
                                                 (filters['repeat_frequency'], filters['repeat_interval'])))
+    query = (query
+             .join(Room.blocked_rooms)
+             .join(BlockedRoom.blocking)
+             .filter(~Room.blocked_rooms.any(db.and_(Blocking.start_date <= filters['end_dt'],
```
ThiefMaster: `db_dates_overlap`?

mic4ael:
Don't merge! FIXING TESTS!

indico/modules/rb/models/rooms.py
```diff
@@ -400,6 +400,11 @@ def sprite_position(self):
         sprite_mapping = _cache.get('rooms-sprite-mapping')
         return sprite_mapping.get(self.id, 0) if sprite_mapping else None  # placeholder at position 0
 
+    @property
+    def is_blocked(self):
```
ThiefMaster: This sends one query whenever accessed, so using this in a loop over many rooms is not a good idea.

indico/modules/rb_new/operations/blockings.py
```diff
@@ -110,3 +111,14 @@ def approve_or_request_blocking(blocking):
...
+
+
+def get_blocked_rooms(start_dt, end_dt, states=None):
+    query = (Room.query
+             .join(Room.blocked_rooms)
+             .join(BlockedRoom.blocking)
+             .filter(db_dates_overlap(Blocking, 'start_date', start_dt, 'end_date', end_dt)))
+
+    if states:
+        query = query.filter(BlockedRoom.state.in_(states or []))
```
ThiefMaster: `or []` never happens

mic4ael: ahh, yeah, forgot to remove it after refactoring

PR #3543, opened 2018-09-20
Add url-shortener template hooks

errikos:
- Parametrised hide event for copy URL tooltip (required from ursh plugin)
- URL shortener template hooks for events and contributions

Should also add hooks to attachments and sessions, but those proved more tricky and possibly require additional changes to the respective templates.

### DIFF
--- a/indico/modules/events/contributions/templates/display/contribution_display.html
+++ b/indico/modules/events/contributions/templates/display/contribution_display.html
@@ -115,6 +115,8 @@
                data-ajax-dialog
                data-reload-after></a>
         {%- endif %}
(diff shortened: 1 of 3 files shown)


ThiefMaster:
what happened to the ones for materials?

ThiefMaster:
Can you push the attachment/session one even if they aren't perfect right now? Better to have something in this PR than having to do that from scratch at some point.

errikos:
I think the attachment hooks are unusable. The reason is that once you have a long attachment name, the link icon disappears. I did not look into sessions at all. I will give it a try in the next few days (for both) and I'll let you know!

PR #3546, opened 2018-09-24
Make timeline filter bar/hours "sticky"

panagiotappl:
Move logic of TimelineLegend one component level above


indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
```diff
@@ -148,6 +174,18 @@ class BookRoom extends React.Component {
                     </div>
                     {this.renderViewSwitch()}
                 </div>
+                <SearchResultCount matching={results.length} total={totalResultCount}
+                                   isFetching={isSearching} />
```
pferreir: Style comment: you may want to put `total={...}` in a new line to be consistent and more merge-friendly.

indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
```diff
@@ -136,8 +146,24 @@ class BookRoom extends React.Component {
         );
     };
 
-    renderFilters(refName) {
-        const {[refName]: ref} = this.state;
+    renderFilters(refName, isTimelineVisible) {
```
pferreir: Isn't `isTimelineVisible` a prop? If so, is there a need to pass it as a parameter here?

indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
```diff
@@ -436,6 +469,7 @@ const mapStateToProps = (state) => {
         isInitializing: globalSelectors.isInitializing(state),
         queryString: stateToQueryString(state.bookRoom, qsFilterRules, qsBookRoomRules),
         showMap: globalSelectors.isMapVisible(state),
+        dateRange: state.bookRoom.timeline.data.dateRange
```
pferreir: I believe there is already a `getTimelineDateRange` selector for this in `bookRoom/selectors.js`.

indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
```diff
@@ -39,6 +39,7 @@ import BookFromListModal from '../../components/modals/BookFromListModal';
 
 
 import '../../components/Timeline.module.scss';
+import TimelineHeader from '../../components/TimelineHeader';
```
pferreir: We normally keep the CSS import at the bottom. It's not enforced by ESLint, but it makes it easier to figure out what is CSS vs what is JS.

indico/modules/rb_new/client/js/components/TimelineBase.jsx
```diff
@@ -175,9 +99,9 @@ export default class TimelineBase extends React.Component {
 
     render() {
         return (
-            <Container>
+            <div>
```
pferreir: Have you checked whether this `<div/>` is even needed? If not, we could probably do OK with renaming `renderContent()` => `render()`.

pferreir:
Very good PR, thanks! There are a few small comments, the most relevant of which being the one about the parameter in `renderFilters(...)`.

panagiotappl:
Thank you for the comments @pferreir. I pushed some changes as suggested!

ThiefMaster:
Looks good, but I think you missed one (moving the js import above the css imports)

PR #3548, opened 2018-09-26
React-Router related fixes

mic4ael:


indico/modules/rb_new/client/js/components/App.jsx
```diff
@@ -79,21 +79,20 @@ export default class App extends React.Component {
...
+                                    return null;
+                                }
+                                return filtersSet
+                                    ? (
+                                        <BookRoom location={location} />
+                                    ) : (
+                                        isExact ? <Landing /> : <Redirect to="/book" />
+                                    );
+                            }} />
+                            <Route path="/rooms" component={!isInitializing ? RoomList : null} />
```
ThiefMaster: What about something like this?
```
function ConditionalRoute({active, component, ...props}) {
    return <Route {...props} component={active ? component : null}>;
}
```

(also needs to handle the case of a render prop instead of component prop of course)

mic4ael: Yeah, I had sth very similar to this (but had to remove it due to some problems which eventually turned out to be unrelated).

indico/modules/rb_new/client/js/components/App.jsx
```diff
@@ -33,6 +33,19 @@ import Menu from './Menu';
...
 
 
+function ConditionalRoute({active, component, render, ...props}) {
+    const routeProps = {};
+    if (component) {
+        routeProps.component = active ? component : null;
+    } else if (render) {
+        routeProps.render = active ? render : null;
+    } else {
+        throw Error('either "component" or "render" should be provided as a prop');
```
ThiefMaster: Isn't it `new Error`?

mic4ael: can be as well

PR #3554, opened 2018-09-27
Show favorite users on top of the results

mic4ael:


indico/modules/users/util.py
```diff
@@ -208,6 +215,8 @@ def build_user_search_query(criteria, exact=False, include_deleted=False, includ
     for k, v in criteria.iteritems():
         query = query.filter(unaccent_match(getattr(User, k), v, exact))
 
+    if order_by_favorites:
+        query = query.order_by(nullslast(favorite_user_table.c.user_id))
```
ThiefMaster: nice idea with nullslast, but unless you apply a custom order somewhere later this now sorts favorite users by their id instead of by name (or whatever else we sort by atm)

PR #3557, opened 2018-09-28
Couple of fixes

mic4ael:


indico/modules/rb_new/client/js/modules/bookRoom/UnavailableRoomsModal.jsx
```diff
@@ -49,9 +49,18 @@ class UnavailableRoomsModal extends React.Component {
 
     render() {
         const {availability, dateRange, filters, isFetching, onClose} = this.props;
-        if (!dateRange) {
+        if (!dateRange || availability.length === 0) {
             return null;
         }
+
+        for (const av of availability) {
+            for (const [, candidates] of Object.entries(av[1].candidates)) {
```
ThiefMaster: use `Object.values()` if you don't need the keys...

mic4ael: yep, forgot to change it, since initially i was expecting there something else

PR #3558, opened 2018-09-28
Booking details modal

meluru:


indico/modules/rb_new/client/js/components/BookingTimeInformation.jsx
```diff
@@ -0,0 +1,73 @@
...
+                {(type === 'daily' || type === 'every') && (
+                    <Label basic pointing="left">
+                        {recurrenceRenderer(recurrence)}
+                    </Label>
+                )}
+            </Segment>
+            <Segment attached="bottom">
+                <Icon name="clock" />
+                <strong>{sTime.format('LT')}</strong>
+                {' '}&rarr;{' '}
```
ThiefMaster: just use the unicode char directly

indico/modules/rb_new/client/js/components/BookingTimeInformation.jsx
```diff
@@ -0,0 +1,73 @@
...
+import recurrenceRenderer from './RecurrenceRenderer';
+
+import './BookingTimeInformation.module.scss';
+
+
+export default function BookingTimeInformation({recurrence, dates, timeSlot}) {
+    const {startDate, endDate} = dates;
+    const {startTime, endTime} = timeSlot;
+    const {type} = recurrence;
+    const sDate = toMoment(startDate);
```
ThiefMaster: i don't like the idea of mixing sDate and startDate for moment/string versions...

indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
```diff
@@ -31,12 +30,15 @@ import roomDetailsModalFactory from '../../components/modals/RoomDetailsModal';
 import {pushStateMergeProps, roomPreloader} from '../../util';
 import * as calendarActions from './actions';
 import * as calendarSelectors from './selectors';
+import * as bookingActions from '../../common/bookings/actions';
```
ThiefMaster: import selectors from other modules through their entry point, ie `import {actions as bookingActions} from '../../common/bookings'`

indico/modules/rb_new/client/js/components/modals/BookingDetailsModal.jsx
```diff
@@ -0,0 +1,181 @@
...
+
+
+class BookingDetailsModal extends React.Component {
+    static propTypes = {
+        booking: PropTypes.object.isRequired,
+        onClose: PropTypes.func,
+    };
+
+    static defaultProps = {
+        onClose: () => {}
```
mic4ael: i think we usually assign it to `null` and then check whether it is not nullish when trying to call

indico/modules/rb_new/client/js/components/modals/BookingDetailsModal.jsx
```diff
@@ -0,0 +1,181 @@
...
+                candidates: occurrences[day].map((candidate) => ({...candidate, bookable: false})) || [],
+            },
+            label: moment(day).format('L'),
+            key: day,
+            room
+        });
+    }
+
+    renderTimeline = (occurrences, dateRange) => {
+        const hourSeries = _.range(6, 24, 2);
```
mic4ael: max as 24? I think in all cases we use 22 as maxHour

indico/modules/rb_new/client/js/common/bookings/reducers.js
```diff
@@ -0,0 +1,42 @@
...
+            bookingsActions.FETCH_BOOKING_DETAILS_REQUEST,
+            bookingsActions.FETCH_BOOKING_DETAILS_SUCCESS,
+            bookingsActions.FETCH_BOOKING_DETAILS_ERROR
+        )
+    }),
+    details: (state = {}, action) => {
+        switch (action.type) {
+            case bookingsActions.BOOKING_DETAILS_RECEIVED: {
+                const {id, ...data} = action.data;
+                return {...state, [id]: data};
```
mic4ael: maybe `camelizeKeys(data)`? thanks to that you will probably avoid renaming object keys down below `booked_for_user: bookedFor`

indico/modules/rb_new/client/js/components/modals/BookingDetailsModal.jsx
ThiefMaster: I think having a callable default is much better :)

indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
```diff
@@ -122,17 +125,25 @@ class Calendar extends React.Component {
...
     };
 
     closeModal = () => {
         const {pushState} = this.props;
         pushState('/calendar');
     };
 
+    openDetailsModal = (booking) => {
+        const {pushState} = this.props;
+        pushState(`/calendar/bookings/${booking.id}/details`);
```
mic4ael: I think you want to pass `true` as the second argument thus preserving the queryString (I think we will have filters there so opening Modal would probably clear the query parameters)

indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
```diff
@@ -167,12 +179,24 @@ class Calendar extends React.Component {
...
                 </Grid.Row>
-                <Route exact path="/calendar/:roomId/details" render={roomPreloader((roomId) => (
+                <Route exact path="/calendar/rooms/:roomId/details" render={roomPreloader((roomId) => (
                     <RoomDetailsModal roomId={roomId} onClose={this.closeModal} />
                 ), fetchRoomDetails)} />
                 <Route exact path="/calendar/:roomId/book" render={roomPreloader((roomId) => (
                     <BookFromListModal roomId={roomId} onClose={this.closeModal} />
                 ), fetchRoomDetails)} />
+                <Route exact path="/calendar/bookings/:bookingId/details" render={({match: {params: {bookingId}}}) => {
+                    return (
```
mic4ael: you don't need an explicit `return` here. You can replace enclosing curly braces with brackets and it will return the component automatically

indico/modules/rb_new/client/js/components/BookingTimeInformation.jsx
meluru: I actually just extracted this to standalone component and didn't look much at the logic inside, but yeah, it's a good opportunity to change it :)

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -143,3 +145,16 @@ class RHRoomSuggestions(RHRoomBookingBase):
...
+
+class RHBookingDetails(RHRoomBookingBase):
+    def _process_args(self):
+        self.booking = Reservation.get_one(request.view_args['booking_id'])
+
+    def _process(self):
+        attributes = reservation_details_schema.dump(self.booking).data
+        date_range, occurrences = get_booking_occurrences(self.booking)
+        date_range = [dt.isoformat() for dt in date_range]
+        serialized = {dt.isoformat(): reservation_details_occurrences_schema.dump(data).data
```
mic4ael: i am a little bit fussy here but you could easily bind to `occurrences` as well ;) `serialized` is not very explanatory

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -143,3 +145,16 @@ class RHRoomSuggestions(RHRoomBookingBase):
...
     def _process(self, args):
         return jsonify(get_suggestions(args, limit=NUM_SUGGESTIONS))
+
+
+class RHBookingDetails(RHRoomBookingBase):
+    def _process_args(self):
+        self.booking = Reservation.get_one(request.view_args['booking_id'])
+
+    def _process(self):
+        attributes = reservation_details_schema.dump(self.booking).data
```
mic4ael: why `attributes`? Isn't it more like `booking_details`?

indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
```diff
@@ -122,17 +125,25 @@ class Calendar extends React.Component {
 
     onClickLabel = (id) => {
         const {pushState} = this.props;
-        pushState(`/calendar/${id}/details`);
+        pushState(`/calendar/rooms/${id}/details`);
```
mic4ael: the same as below

indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
meluru: I didn't add it here for now intentionally. Because with no query parameters I end up with 'undefined' in the querystring

indico/modules/rb_new/controllers/backend/bookings.py
meluru: well, it is, but the whole thing is booking details, if you look in the redux state, then you have this data under bookings/details/ so I thought it may be strange to use details again

indico/modules/rb_new/schemas.py
```diff
@@ -77,6 +77,25 @@ class Meta:
...
+
+class ReservationDetailsSchema(mm.ModelSchema):
+    room = Nested(RoomSchema)
+    booked_for_user = Nested(UserSchema, only=('full_name', 'phone', 'email'))
+    created_by_user = Nested(UserSchema, only=('full_name'))
+
+    class Meta:
+        model = Reservation
+        fields = ('start_dt', 'end_dt', 'repetition', 'booking_reason', 'created_dt', 'status_string',
+                  'uses_vc', 'needs_vc_assistance', 'booked_for_user', 'room', 'created_by_user', )
```
ThiefMaster: let's omit things like `status_string` and `uses_*` as they'll likely go away. also, no need to serialize the room again. having the ID is enough (you can get the room data from the redux state)

indico/modules/rb_new/client/js/common/bookings/reducers.js
meluru: camelizeKeys messes up the keys in occurrences (when we use strings with dates as keys) so I won't use it

mic4ael: then I think it might be a good thing to actually fix this method since I don't think we want to camelize something like `2018-05-01`

indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
mic4ael: weren't filters already added there?

indico/modules/rb_new/client/js/common/bookings/reducers.js
ThiefMaster: Yep, camelize shouldn't break a string that's doesn't even contain any letters... (or rather it should probably only modify strings consisting solely of `[a-zA-A_]`)

indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
ThiefMaster: pushState calls to open modals are going  away anyway, so I wouldn't worry too much about any code related to opening modals

indico/modules/rb_new/client/js/common/bookings/reducers.js
meluru: Good idea, I'll do so

ThiefMaster:
.../bookRoom/filters/RecurrenceRenderer.jsx → .../components/RecurrenceRenderer.jsx

intentional? also, those *Renderer functions aren't components. Let's either keep them as functions returning jsx (which is fine), and move them to a better place than ../components/.., or rename them to have an `UpperCamelCase` name like all components should have. But to me they look more like utils than components...

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -134,3 +136,20 @@ class RHRoomSuggestions(RHRoomBookingBase):
...
+        date_range, occurrences = get_booking_occurrences(self.booking)
+        date_range = [dt.isoformat() for dt in date_range]
+        occurrences = {dt.isoformat(): reservation_details_occurrences_schema.dump(data).data
+                       for dt, data in occurrences.iteritems()}
+        edit_logs = sorted(reservation_edit_log_schema.dump(self.booking.edit_logs).data,
+                           key=itemgetter('timestamp'),
+                           reverse=True)
+        return jsonify(attributes=attributes, id=self.booking.id, occurrences=occurrences, date_range=date_range,
```
ThiefMaster: what about `booking=attributes` - in the end it's the basic booking data we have here.

Or, crazy idea, do we even need this separate? ie serialize the booking details, then add the other information to that dict.
On the Python side we have all on the same "layer" as well (booking.id, booking.occurrences, etc.)

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -134,3 +136,20 @@ class RHRoomSuggestions(RHRoomBookingBase):
...
+    def _process_args(self):
+        self.booking = Reservation.get_one(request.view_args['booking_id'])
+
+    def _process(self):
+        attributes = reservation_details_schema.dump(self.booking).data
+        date_range, occurrences = get_booking_occurrences(self.booking)
+        date_range = [dt.isoformat() for dt in date_range]
+        occurrences = {dt.isoformat(): reservation_details_occurrences_schema.dump(data).data
+                       for dt, data in occurrences.iteritems()}
+        edit_logs = sorted(reservation_edit_log_schema.dump(self.booking.edit_logs).data,
```
ThiefMaster: you can do the sorting on the schema as well, see the changes in schemas.py in #3582 for an example

indico/web/client/js/utils/camelize.js
```diff
@@ -28,6 +28,10 @@ export default function camelizeKeys(obj) {
     }
 
     return Object.entries(obj).reduce((accum, [key, value]) => {
-        return {...accum, [_.camelCase(key)]: camelizeKeys(value)};
+        if (key.match('^[A-Za-z_]+$')) {
```
ThiefMaster: `key.match(/^[A-Za-z_]+$/)`?

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -0,0 +1,263 @@
...
+                    </Translate>
+                </List.Item>
+            );
+        }));
+        return (
+            <div styleName="booking-logs">
+                <Header><Translate>Booking history</Translate></Header>
+                <List divided styleName="log-list">
+                    {items}
+                    <List.Item>
```
ThiefMaster: what about injecting the fake log item on the server side? you could append it `post_dump` in the `ReservationDetailsSchema` (assuming you move log handling there). if you prefer to do it on the client:

    editLogs = [...editLogs, {your fake log entry here}];

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -0,0 +1,263 @@
...
+                            <Param name="created-by" value={createdBy} />
+                        </Translate>
+                    </List.Item>
+                </List>
+            </div>
+        );
+    };
+
+    renderBookedFor = (bookedForUser) => {
+        if (bookedForUser) {
```
ThiefMaster:
```
if (!bookedForUser) {
    return null;
}
```

or even call `renderBookedFor` conditionally

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -0,0 +1,263 @@
...
+            );
+        }
+    };
+
+    renderReason = (reason) => (
+        <Message info icon styleName="message-icon">
+            <Icon name="info" />
+            <Message.Content>
+                <Message.Header>
+                    <Translate>Booking reason: </Translate>
```
ThiefMaster: do we really want the `:` there? also, let's not put trailing whitespace in translate - we strip it anyway.

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -0,0 +1,263 @@
...
+                            {this.renderBookedFor(bookedForUser)}
+                            {this.renderReason(bookingReason)}
+                            {this.renderBookingHistory(editLogs, createdOn, createdByUser)}
+                            {showActionButtons && this.renderActionButtons(canAccept, canCancel, canReject, isAccepted)}
+                        </Grid.Column>
+                    </Grid>
+                </Modal.Content>
+                <Modal open={occurrencesVisible}
+                       onClose={() => this.setState({occurrencesVisible: false})}
+                       size="large" closeIcon>
```
ThiefMaster: i'd put `closeIcon` in a new line, since all other attributes are in their own line as well

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -0,0 +1,263 @@
...
+                </Modal.Header>
+                <Modal.Content>
+                    <Grid columns={2}>
+                        <Grid.Column>
+                            <RoomBasicDetails room={room} />
+                            <RoomKeyLocation room={room} />
+                            <TimeInformation recurrence={recurrence}
+                                             dates={dates}
+                                             timeSlot={times}
+                                             onClickOccurrences={() => this.setState({occurrencesVisible: true})}
```
ThiefMaster: what about adding a `toggleOccurrences` method (or showOccurrences and hideOccurrences), and just doing `onClickOccurrences={this.toggleOccurrences}` or `onClickOccurrences={this.showOccurrences}` here?

It doesn't make a huge difference in this case, but it's always better if defining new functions in render() can be avoided (those functions are redefined on each render after all)

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -0,0 +1,263 @@
...
+                <Modal open={occurrencesVisible}
+                       onClose={() => this.setState({occurrencesVisible: false})}
+                       size="large" closeIcon>
+                    <Modal.Header className="legend-header">
+                        <Translate>Occurrences</Translate>
+                        <Popup trigger={<Icon name="info circle" className="legend-info-icon" />}
+                               content={<TimelineLegend labels={legendLabels} compact />} />
+                    </Modal.Header>
+                    <Modal.Content scrolling>
+                        {occurrences && this.renderTimeline(occurrences, dateRange)}
```
ThiefMaster: i think `occurrences` is an object or array, so it's never falsy

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.module.scss
```diff
@@ -0,0 +1,53 @@
...
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+@import 'rb_new:styles/palette';
+
+
+.log-list {
+    max-height: 140px;
+    overflow-y: auto;
+    overflow-x: auto;
```
ThiefMaster: `overflow: auto;` covers both x and y

indico/modules/rb_new/client/js/common/bookings/BookingDetailsPreloader.jsx
```diff
@@ -0,0 +1,51 @@
...
+import {Preloader} from 'indico/react/util';
+import * as bookingsActions from './actions';
+import * as bookingsSelectors from './selectors';
+
+
+const BookingDetailsPreloader = ({bookingId, component: Component, fetchDetails, onClose}) => (
+    <Preloader checkCached={state => bookingsSelectors.hasDetails(state, {bookingId})}
+               action={() => fetchDetails(bookingId)}
+               dimmer={<Dimmer active page><Loader /></Dimmer>}>
+        {() => {
```
ThiefMaster: {() => <Component bookingId={bookingId} onClose={onClose} />}

indico/modules/rb_new/client/js/components/TimeInformation.jsx
```diff
@@ -0,0 +1,117 @@
...
+            <Segment attached="top" color="teal">
+                <div styleName="occurences-details">
+                    <div>
+                        <Icon name="calendar outline" />
+                        {(mEndDate && !mStartDate.isSame(mEndDate, 'day'))
+                            ? (
+                                <Translate>
+                                    <Param name="startDate"
+                                           wrapper={<strong />}
+                                           value={mStartDate.format('L')} /> to <Param name="endDate"
```
ThiefMaster: this can be made much more readable by writing the spaces as `{' '}` so you don't have to wrap in this weird way

indico/modules/rb_new/client/js/components/TimeInformation.jsx
```diff
@@ -0,0 +1,117 @@
...
+                                <strong>{mStartDate.format('L')}</strong>
+                            )
+                        }
+                        {(type === 'daily' || type === 'every') && (
+                            <Label basic pointing="left">
+                                {renderRecurrence(recurrence)}
+                            </Label>
+                        )}
+                    </div>
+                    {(occurrencesNumber) && (
```
ThiefMaster: no need for the `()` around occurrencesNumber

indico/modules/rb_new/client/js/components/TimeInformation.jsx
```diff
@@ -0,0 +1,117 @@
...
+                        <Button color="blue" size="small" hovercontent="See timeline" styleName="hover-button"
+                                onClick={() => onClickOccurrences()}>
```
ThiefMaster: `onClick={onClickOccurrences}`

indico/modules/rb_new/client/js/util.jsx
```diff
@@ -133,3 +137,51 @@ export function boolStateField(name) {
 export function isDateWithinRange(date, dateRange, _toMoment) {
     return date && dateRange.filter((dt) => _toMoment(dt).isSame(date, 'day')).length !== 0;
 }
+
+
+export function PopupParam({content, children}) {
```
ThiefMaster: nothing for now but I have the feeling we soon need to split this file. We now have various utils and a component in here...

indico/modules/rb_new/operations/bookings.py
```diff
@@ -200,3 +200,10 @@ def get_room_details_availability(room, start_dt, end_dt):
             'day': iso_day,
         })
     return sorted(availability, key=itemgetter('day'))
+
+
+def get_booking_occurrences(booking):
+    candidates = booking.occurrences
```
ThiefMaster: those aren't candidates. you could just use booking.occurrences directly in the next line

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -0,0 +1,220 @@
...
+        const {booking} = this.props;
+        const {
+            room, bookedForUser, startDt, endDt, repetition, createdByUser, createdDt, bookingReason
+        } = booking.attributes;
+        const {occurrences, dateRange, editLogs} = booking;
+        const {occurrencesVisible} = this.state;
+        const {fullName: bookedForName, email: bookedForEmail, phone: bookedForPhone} = bookedForUser;
+        const dates = {startDate: startDt, endDate: endDt};
+        const createdOn = serializeDate(toMoment(createdDt));
+        const times = {startTime: moment(startDt).utc().format('HH:mm'), endTime: moment(endDt).utc().format('HH:mm')};
```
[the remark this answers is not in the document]
ThiefMaster: OK, they are correct, but I think semantically they are not. We get e.g. `2018-01-10T11:00:00+00:00` from the server, but those timestamps are naive (without timezone information). Ideally they shouldn't be serialized with a timezone. With a timezone-less string such as `2018-01-10T11:00:00` you get the correct output without the `.utc()` call.

Unfortunately marshmallow's DateTime type doesn't support this so let's stick with your workaround for now. It's not pretty but we can always add a custom marshmallow field type for this later... (and I'm tempted to do so once this is merged ;))

indico/modules/rb_new/schemas.py
```diff
@@ -81,6 +84,30 @@ class Meta:
...
+    reservation = Nested(ReservationSchema)
+
+    class Meta:
+        model = ReservationOccurrence
+        fields = ('start_dt', 'end_dt', 'is_valid')
+
+
+class ReservationDetailsSchema(mm.ModelSchema):
+    booked_for_user = Nested(UserSchema, only=('full_name', 'phone', 'email'))
+    created_by_user = Nested(UserSchema, only=('full_name'))
```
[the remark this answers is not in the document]
ThiefMaster: Looks like the string is correct, but in that case remove the `()` around it.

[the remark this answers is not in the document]
meluru: I have already changed it to a tuple, so I'll keep it that way.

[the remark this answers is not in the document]
ThiefMaster: ok, either way is fine -  just wanted to update the comment since i noticed that it was working fine because your code was actually expecting a flat string instead of an object (which is what nested apparently gives you when specifying a string)

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -0,0 +1,263 @@
...
+    renderBookedFor = (bookedForUser) => {
+        if (bookedForUser) {
+            const {fullName: bookedForName, email: bookedForEmail, phone: bookedForPhone} = bookedForUser;
+            return (
+                <>
+                    <Header><Icon name="user" /><Translate>Booked for</Translate></Header>
+                    <div>{bookedForName}</div>
+                    {bookedForPhone && <div><Icon name="phone" />{bookedForPhone}</div>}
+                    {bookedForEmail && <div><Icon name="mail" />{bookedForEmail}</div>}
+                </ >
```
ThiefMaster: should be `</>` without the space

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
meluru: I am no sure about it, I would need to fake some id there

ThiefMaster: what about `"created"`? Doesn't need to be a number ;)

meluru: yeah, okay, that makes sense :)

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -0,0 +1,251 @@
...
+    renderTimeline = (occurrences, dateRange) => {
+        const hourSeries = _.range(6, 22, 2);
+        const rows = dateRange.map((day) => this._getRowSerializer(day)(occurrences));
+        return <DailyTimelineContent rows={rows} hourSeries={hourSeries} />;
+    };
+
+    renderBookingHistory = (editLogs, createdOn, createdByUser) => {
+        if (createdByUser) {
+            const {fullName: createdBy} = createdByUser;
+            editLogs = [...editLogs, {id: 'created',
```
ThiefMaster:
```
            editLogs = [...editLogs, {
                id: 'created',
                timestamp: createdOn,
                info: ['Booking created'],
                userName: createdBy
            }];
```

PR #3561, opened 2018-10-01
Implement filtering by division

pferreir:
Also making a few more components `Overridable`.


indico/web/client/js/react/util/Overridable.jsx
```diff
@@ -4,25 +4,26 @@ import {connect} from 'react-redux';
 
 
 function Overridable({children, overrides, id, ...restProps}) {
-    const childProps = children.props;
+    const childProps = children ? children.props : {};
```
mvidalgarcia: What about adding some brief JSDocs to the `Overridable` component?

PR #3564, opened 2018-10-03
Centralize modal handling

ThiefMaster:


ThiefMaster:
Just thinking, does it make much sense to have the UnavailableRoomsModal handled centrally **or be part of the query string at all**? This is not something where it makes much sense for someone to link to directly so maybe using component-local state to show/hide that would make more sense?

PR #3570, opened 2018-10-08
Validate email address in contribution csv import

### LINKED ISSUE #3568: Validate email addresses in contribution csv import
Currently we don't, and if someone mixes up e.g. affiliation and email that results in very messy data.

`import_contributions_from_csv` already has some checks, but it's missing one using `validate_email`. Note: The email is optional, so we should only validate it if it's not empty.

Kush22:
An exception is raised only before ascertaining that the data is ok.

fixes #3568

### DIFF
--- a/indico/modules/events/contributions/util.py
+++ b/indico/modules/events/contributions/util.py
@@ -16,7 +16,6 @@
 
 from __future__ import unicode_literals
 
-import re
 import csv
 from collections import defaultdict
 from datetime import timedelta
@@ -40,7 +39,7 @@ from indico.modules.events.persons.util import get_event_person
 from indico.modules.events.util import serialize_person_link, track_time_changes
 from indico.util.date_time import format_human_timedelta
 from indico.util.i18n import _
-from indico.util.string import to_unicode
+from indico.util.string import to_unicode, validate_email
 from indico.web.flask.templating import get_template_module
 from indico.web.flask.util import url_for
 from indico.web.http_api.metadata.serializer import Serializer
@@ -199,11 +198,6 @@ def get_contribution_ical_file(contrib):
     serializer = Serializer.create('ics')
     return BytesIO(serializer(data))
 
-def is_valid_email(email):
-    if len(email) > 7:
-        if re.match("^.+@([?)[a-zA-Z0-9-.]+.([a-zA-Z]{2,3}|[0-9]{1,3})(]?)$", email) != None:
-            return True
-    return False
 
 def import_contributions_from_csv(event, f):
     """Import timetable contributions from a CSV file into an event."""
@@ -232,10 +226,8 @@ def import_contributions_from_csv(event, f):
         if not title:
             raise UserValueError(_("Row {}: contribution title is required").format(num_row))
 
-        
-        if not email or not is_valid_email(email):
+        if not validate_email(email):
             raise UserValueError(_("Row {row}: doesn't have valid email address: {email}").format(row=num_row, email=email))
-
 
         contrib_data.append({
             'start_dt': parsed_start_dt,
@@ -266,7 +258,7 @@ def import_contributions_from_csv(event, f):
             all_changes[key].append(val)
 
         email = speaker_data['email']
-        if not email or not is_valid_email(email):
+        if not email:
             continue
 
         # set the information of the speaker
--- a/indico/modules/events/contributions/util.py
+++ b/indico/modules/events/contributions/util.py
@@ -16,6 +16,7 @@
 
 from __future__ import unicode_literals
 
+import re
 import csv
 from collections import defaultdict
 from datetime import timedelta
--- a/indico/modules/events/contributions/util.py
+++ b/indico/modules/events/contributions/util.py
@@ -198,6 +198,11 @@ def get_contribution_ical_file(contrib):
     serializer = Serializer.create('ics')
     return BytesIO(serializer(data))
 
+def is_valid_email(email):
+    if len(email) > 7:
+        if re.match("^.+@([?)[a-zA-Z0-9-.]+.([a-zA-Z]{2,3}|[0-9]{1,3})(]?)$", email) != None:
+            return True
+    return False
 
 def import_contributions_from_csv(event, f):
     """Import timetable contributions from a CSV file into an event."""
@@ -225,6 +230,11 @@ def import_contributions_from_csv(event, f):
 
         if not title:
             raise UserValueError(_("Row {}: contribution title is required").format(num_row))
+
+        
+        if not email or not is_valid_email(email):
+            raise UserValueError(_("Row {row}: doesn't have valid email address: {email}").format(row=num_row, email=email))
+
 
         contrib_data.append({
             'start_dt': parsed_start_dt,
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Hi, first of all, thanks for the contribution!

Please do not add hashtags to commit messages. Hashtags in commit messages are not a thing and will hopefully never be (*please* tell me Hbcktoberfest doesn't actually encourage people to do this...).

There are a few issues with the code itself which I'll point out in a review.

PS: Someone else already commented on the issue stating they were interested in working on it. Generally it's better to not pick the same thing someone else is probably already working on.

indico/modules/events/contributions/util.py
```diff
@@ -16,6 +16,7 @@
 
 from __future__ import unicode_literals
 
+import re
```
ThiefMaster: As you can see in the failing tests on Travis, we are using isort to ensure consistent sorting of imports. Please fix the order.

indico/modules/events/contributions/util.py
```diff
@@ -198,6 +199,11 @@ def get_contribution_ical_file(contrib):
     serializer = Serializer.create('ics')
     return BytesIO(serializer(data))
 
+def is_valid_email(email):
+    if len(email) > 7:
```
ThiefMaster: Why `>7`? `a@b.cc` could be a perfectly valid email address and it's only 6 characters.

indico/modules/events/contributions/util.py
```diff
@@ -198,6 +199,11 @@ def get_contribution_ical_file(contrib):
...
+        if re.match("^.+@([?)[a-zA-Z0-9-.]+.([a-zA-Z]{2,3}|[0-9]{1,3})(]?)$", email) != None:
```
ThiefMaster: PEP8 states that you should use `is not` for this comparison:

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.

Also, did you even test this regex? 

```
>>> re.match("^.+@([?)[a-zA-Z0-9-.]+.([a-zA-Z]{2,3}|[0-9]{1,3})(]?)$", 'foo@bar.google')
  File "<stdin>", line 1, in <module>
  File "/home/adrian/.python-env/lib64/python2.7/re.py", line 141, in match
    return _compile(pattern, flags).match(string)
  File "/home/adrian/.python-env/lib64/python2.7/re.py", line 251, in _compile
    raise error, v # invalid expression
sre_constants.error: unbalanced parenthesis
```

Regardless of this, the regex doesn't make much sense and also doesn't take anything but ASCII into account.

indico/modules/events/contributions/util.py
```diff
@@ -198,6 +199,11 @@ def get_contribution_ical_file(contrib):
     serializer = Serializer.create('ics')
     return BytesIO(serializer(data))
 
+def is_valid_email(email):
```
ThiefMaster: We already have a function `validate_email` as stated in the original issue:

https://github.com/indico/indico/blob/master/indico/util/string.py#L288

indico/modules/events/contributions/util.py
```diff
@@ -226,6 +232,11 @@ def import_contributions_from_csv(event, f):
         if not title:
             raise UserValueError(_("Row {}: contribution title is required").format(num_row))
 
+
```
ThiefMaster: Why so much space around everything? The only place a Python file should have two consecutive empty lines is between top-level functions/classes.

indico/modules/events/contributions/util.py
```diff
@@ -226,6 +232,11 @@ def import_contributions_from_csv(event, f):
         if not title:
             raise UserValueError(_("Row {}: contribution title is required").format(num_row))
 
+        
+        if not email or not is_valid_email(email):
```
ThiefMaster: As stated in the issue, having no email is perfectly acceptable.

indico/modules/events/contributions/util.py
```diff
@@ -226,6 +232,11 @@ def import_contributions_from_csv(event, f):
...
+            raise UserValueError(_("Row {row}: doesn't have valid email address: {email}").format(row=num_row, email=email))
```
ThiefMaster: > doesn't have *a* valid email address

indico/modules/events/contributions/util.py
```diff
@@ -226,6 +232,11 @@ def import_contributions_from_csv(event, f):
...
+
```
ThiefMaster: same as above

indico/modules/events/contributions/util.py
```diff
@@ -255,7 +266,7 @@ def import_contributions_from_csv(event, f):
             all_changes[key].append(val)
 
         email = speaker_data['email']
-        if not email:
+        if not email or not is_valid_email(email):
```
ThiefMaster: Why check this again? This code is not reachable if there is an invalid email.

ThiefMaster:
Note: This PR is currently labelled as [invalid], i.e. it does not count for Hacktoberfest. If you improve it to meet the quality standards of our code base, I will gladly remove that label again.

Kush22:
Hi,
Thank you so much for your comments & and sorry for making erroneous updates. I should have made cross-checks before making any change.
Please now look at the changes made again. Thank you so much!

ThiefMaster:
much better! can you please fix the typo i commented? (missing "a")

Kush22:
> much better! can you please fix the typo I commented? (missing "a")
Thank you! 
I couldn't find the typo.

ThiefMaster:
'doesn't have **a** valid email address' - the "a" is missing.

anyway, let's go for something more concise, such as this:

    raise UserValueError(_("Row {row}: invalid email address: {email}").format(row=num_row, email=email))

indico/modules/events/contributions/util.py
```diff
@@ -226,6 +226,9 @@ def import_contributions_from_csv(event, f):
         if not title:
             raise UserValueError(_("Row {}: contribution title is required").format(num_row))
 
+        if not validate_email(email):
```
ThiefMaster: Still not the correct logic. The email address is optional, but if it's set it must be valid. So you need `if email and not validate_email(email)` here (this is actually why the tests are currently failing).

ThiefMaster:
Please also add a another test case to [`test_import_contributions_errors`](https://github.com/indico/indico/blob/master/indico/modules/events/contributions/util_test.py#L110) that covers the case of an invalid email (just put `foobar` or something similar that is obviously invalid as the email address there)

ThiefMaster:
A changelog entry would be good as well. You can paste this to CHANGES.rst, at the end of the "Bugfixes" of v2.2 (please keep the line wrap and indentation of the second line as-is):

```rst
- Add missing email validation to contribution CSV import (:issue:`3568`,
  thanks :user:`Kush22`)
```

indico/modules/events/contributions/util_test.py
```diff
@@ -125,3 +125,6 @@ def test_import_contributions_errors(db, dummy_event):
 
     e = _check_importer_exception(dummy_event, b'2010-02-23T00:00:00,15min,Test,,,,')
     assert 'parse duration' in e.message
+
+    e = _check_importer_exception(dummy_event, b'2010-02-23T00:00:00,15min,Test,,,,foobar')
```
ThiefMaster: this still contains an invalid duration so the test might fail depending on the order of the checks. better take an otherwise valid string (except for the bad email of course)

PR #3576, opened 2018-10-10
Improve UX of forms using PrincipalField

ThiefMaster:
This PR also massively improves the performance in the booking and blocking dialog by avoiding (very) expensive re-renders of fields like the date or user picker if not necessary.

### DIFF
--- a/indico/modules/users/util.py
+++ b/indico/modules/users/util.py
@@ -178,7 +178,8 @@ def serialize_user(user):
 
 def _build_name_search(name_list):
     text = remove_accents('%{}%'.format('%'.join(escape_like(name) for name in name_list)))
-    return db.func.indico.indico_unaccent(db.func.concat(User.first_name, ' ', User.last_name)).ilike(text)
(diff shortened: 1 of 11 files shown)


indico/modules/users/util.py
```diff
@@ -206,8 +207,7 @@ def build_user_search_query(criteria, exact=False, include_deleted=False, includ
             raise ValueError("'name' is not compatible with 'exact'")
         if 'first_name' in criteria or 'last_name' in criteria:
             raise ValueError("'name' is not compatible with (first|last)_name")
-        name = name.split()
-        query = query.filter(_build_name_search(name))
+        query = query.filter(_build_name_search(name.replace(',', '').split()))
```
mvidalgarcia: What if someone types 'Monnich,Adrian' (no space)? It would end up being 'MonnichAdrian'. I wonder if `ilike` is able to find 'Monnich__Adrian' (two spaces), so we could safely do `name.replace(',', ' ')`.

ThiefMaster: I thought about it as well, but in the end I thought "garbage in, garbage out". If someone enters a name in a way nobody uses.. their fault ;)

PR #3590, opened 2018-10-16
Contribution excel export:  person => speaker full name 

mvidalgarcia:


ThiefMaster:
Changelog entry please (just copy the 2.1.4 section and set the date as Unreleased).

Even if we probably never release a 2.1.5, let's still keep the changelog up to date

PR #3599, opened 2018-10-17
Align footer-links in the center of the page

nop33:
I noticed that the footer links are clearly not in the center of the page and it seems like it was intended to be. This PR fixes this issue. To demonstrate:

## Before
screen shot 2018-10-17 at 23 21 36

## After
screen shot 2018-10-17 at 23 22 34

Also, is there a reason why the right padding of the footer is `5px` but the left is `10px` in `indico/web/client/styles/partials/_footer.scss`?

### DIFF
--- a/indico/web/templates/footer.html
+++ b/indico/web/templates/footer.html
@@ -4,7 +4,7 @@
             <a href="https://getindico.io">Indico</a>
         {%- endset -%}
         <div class="flexrow f-j-space-between">
-            <div class="flexrow f-a-center">
+            <div class="flexrow f-a-center f-self-stretch">
                 {% block footer_logo %}
                     {% set filename = 'indico_small_white.png' if dark else 'indico_small.png' %}
                     <img src="{{ url_for('assets.image', filename=filename) }}" class="footer-logo" alt="Indico">
@@ -32,7 +32,7 @@
                     </li>
                 {% endfor %}
             </ul>
-            <div class="footer-extra">
+            <div class="footer-extra f-self-stretch">
                 {% block footer_extra %}
                 {% endblock %}
             </div>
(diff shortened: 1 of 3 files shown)


indico/web/templates/footer.html
```diff
@@ -4,7 +4,7 @@
             <a href="https://getindico.io">Indico</a>
         {%- endset -%}
         <div class="flexrow f-j-space-between">
-            <div class="flexrow f-a-center">
+            <div class="flexrow f-a-center f-self-stretch">
```
ThiefMaster: I'm curious, why do we need to stretch the outer elements for this? Shouldn't space-between on the container already spread them evenly?

nop33: If all items have `flex-grow` set to `1`, the remaining space in the container will be distributed equally to all children. In this case we have 3 children and we want the middle one to be at the exact center. So distributing the remaining space to the first and third child, we make sure that the 2nd one will always be in the middle.

nop33: Even if the third element has no content ;)

nop33:
yay, merged!

ThiefMaster:
Hey @nop33, I just noticed that this PR broke the footer logo text on the login page:

image

I know it's not october anymore, but would you be willing to fix that?

nop33:
Sure, I can have a look!

ThiefMaster:
`rm -rf node_modules` is the easiest option; npm doesn't broken symlinks in there

nop33:
same result

ThiefMaster:
try deleting package-lock.json as well

nop33:
I did it right before I saw your comment. Worked, thx!

PR #3600, opened 2018-10-17
Fix right spacing of session bar in event page

nop33:
The navigation bar in the event page has different left and right spacing. This fix unifies the spacings by updating the right spacing to be the same as the left. Demonstration:

## Before
image

## After
image


nop33:
image

wut?

Am I not seeing something? It's the same XD

ThiefMaster:
leading space

ThiefMaster:
It's a bit weird that now the session bar alignment is even more different between inside and outside events than before. Before this change it was only vertically, now it's different horizontally as well.

nop33:
Then maybe it should be fixed globally.

pferreir:
Someone really wants a t-shirt :trollface:

nop33:
Yep! And I'll get it :trollface:

PR #3603, opened 2018-10-18
Move SearchBar next to the filters

mic4ael:


ThiefMaster:
Something in this change causes the "Actions" button to look really bad when the screen is small enough to cause wrapping:

image

mic4ael:
Thanks! Fixed now.

PR #3607, opened 2018-10-18
Improve event list

nop33:
This MR fixes the alignment of labeled dates compared to the not labeled ones as well as the text arrangement of long titled events. See demonstration:

## Before
screen shot 2018-10-18 at 20 42 00

## After
screen shot 2018-10-18 at 20 45 52


ThiefMaster:
Looks great :)

PS: MR? Someone did too much GitLab!

PR #3612, opened 2018-10-22
Improve the BlockingList component

mic4ael:


indico/modules/rb_new/schemas.py
```diff
@@ -158,10 +158,12 @@ class BlockingSchema(mm.ModelSchema):
     blocked_rooms = Nested(BlockedRoomSchema, many=True)
     allowed = Nested(BlockingPrincipalSchema, many=True)
     can_edit = Function(lambda blocking: blocking.can_be_modified(session.user))
+    created_by = Function(lambda blocking: blocking.created_by_user.full_name)
```
ThiefMaster: I think this would be cleaner in terms of semantics (untested!):

    created_by = Nested(UserSchema, attribute='created_by_user', only='full_name')

indico/modules/rb_new/schemas.py
```diff
@@ -158,10 +158,12 @@ class BlockingSchema(mm.ModelSchema):
...
 
     class Meta:
         model = Blocking
-        fields = ('id', 'start_date', 'end_date', 'reason', 'blocked_rooms', 'allowed', 'created_by_id', 'can_edit')
+        fields = ('id', 'start_date', 'end_date', 'reason', 'blocked_rooms', 'allowed', 'created_by_id', 'can_edit',
```
ThiefMaster: `created_by_id` is unused in the client code; no need to expose it.

indico/modules/rb_new/schemas.py
```diff
@@ -158,10 +158,11 @@ class BlockingSchema(mm.ModelSchema):
     blocked_rooms = Nested(BlockedRoomSchema, many=True)
     allowed = Nested(BlockingPrincipalSchema, many=True)
     can_edit = Function(lambda blocking: blocking.can_be_modified(session.user))
+    created_by = Nested(UserSchema, attribute='created_by_user', only=('full_name',))
```
ThiefMaster: if you use a string instead of a tuple you get rid of the nesting so `created_by` is a string. imho that'd be cleaner here since then you can just use `blocking.createdBy` in the frontend

mic4ael: Good to know, quite useful.

PR #3629, opened 2018-10-23
Room/TimelineItem Placeholders

### LINKED ISSUE #3328: Loading placeholders for list of rooms/timeline
I don't think Semantic UI has any built-in element for it, but should be easy to make one.

mic4ael:
closes #3328


ThiefMaster:
Some issues I found:

- Wrong margin for blockings:
image
- Do we really want 20 placeholders for blockings? I'd go for much less there...
- Timeline view of book-a-room now loads all pages before showing anything now instead of using the lazy scroller for it.

mic4ael:
Thanks! Didn't notice the problem with margin. As to the number of placeholders I don't have a strong opinion about it. Can be less but I feel like it looks a bit better with a higher number of placeholders.

PR #3633, opened 2018-10-23
Add Modal.Actions bar to the booking details modal

### LINKED ISSUE #3614: Booking details: fix styling
In case of missing key location information some parts of Booking History are not visible

meluru:
Closes  #3614

### DIFF
--- a/indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
+++ b/indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
@@ -160,8 +160,8 @@ class BookingDetailsModal extends React.Component {
         </Message>
     );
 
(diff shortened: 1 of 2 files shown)


indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -200,7 +200,8 @@ class BookingDetailsModal extends React.Component {
         const legendLabels = [
             {label: Translate.string('Occurrence'), color: 'orange'},
         ];
-        const showActionButtons = (!isCancelled && !isRejected);
+        const showAccept = canAccept && !isAccepted;
```
ThiefMaster: I think that should be covered by `canAccept`. but we can revisit the logic here when cleaning up the permission checks in the backend and/or actually adding functionality to those buttons

PR #3634, opened 2018-10-23
Add setting for custom room name format

### LINKED ISSUE #2095: Room Booking: Adapt notation of rooms to match CERN's
Related ticket in SNOW `CHG34013`.

We are using the old notation for rooms at CERN (`{building}-{floor}-{room}`), but [maps.cern.ch](maps.cern.ch) has stopped recognizing it in favor for the new one (`{building}/{floor}-{room}`). Links keep working just fine, but it's not ideal when people copy&paste information coming from Indico into CERN maps.

Even better than simply hard-coding a CERN policy, we could probably allow overriding the default notation in the configuration file.

pferreir:
Also updated default to `X/Y-Z` instead of `X-Y-Z`.

Closes #2095

### DIFF
--- a/indico/migrations/versions/20181023_1051_55893713f6b7_add_room_name_format.py
+++ b/indico/migrations/versions/20181023_1051_55893713f6b7_add_room_name_format.py
@@ -1,0 +1,27 @@
+"""Add room_name_format
+
+Revision ID: 55893713f6b7
+Revises: 93985a8c11ed
+Create Date: 2018-10-23 10:51:55.999263
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '55893713f6b7'
+down_revision = '93985a8c11ed'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('locations',
+                  sa.Column('room_name_format', sa.String(), nullable=False, server_default='%1$s/%2$s-%3$s'),
+                  schema='roombooking')
+    op.alter_column('locations', 'room_name_format', server_default=None, schema='roombooking')
+
+
+def downgrade():
+    op.drop_column('locations', 'room_name_format', schema='roombooking')
--- a/indico/modules/rb/models/locations.py
+++ b/indico/modules/rb/models/locations.py
@@ -14,10 +14,12 @@
 # You should have received a copy of the GNU General Public License
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
+import re
 from collections import defaultdict
 from datetime import time
 
 from sqlalchemy import func
+from sqlalchemy.ext.hybrid import hybrid_property
 
 from indico.core.db import db
 from indico.modules.rb.models.aspects import Aspect
@@ -26,6 +28,9 @@ from indico.util.decorators import classproperty
 from indico.util.i18n import _
 from indico.util.locators import locator_property
 from indico.util.string import return_ascii
+
+
+ROOM_NAME_TPL_RE = re.compile(r'%(\d)\$s')
 
 
 class Location(db.Model):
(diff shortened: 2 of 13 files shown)


indico/migrations/versions/20181023_1051_55893713f6b7_add_room_name_format.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = '55893713f6b7'
+down_revision = '93985a8c11ed'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('locations',
+                  sa.Column('room_name_format', sa.String(), nullable=False, server_default=u'%1$s/%2$s-%3$s'),
```
ThiefMaster: - no need for the unicode prefix
- remove the `server_default` (`alter_column` to set it to None) after adding the column to be in sync with the model

indico/modules/rb/models/locations.py
```diff
@@ -49,17 +54,20 @@ def working_time_end(cls):
         db.Integer,
         primary_key=True
     )
+
```
ThiefMaster: i think we usually don't have empty lines there

indico/modules/rb/models/locations_test.py
```diff
@@ -42,6 +43,19 @@ def test_default_location(create_location):
...
+def test_room_name_format(create_location, create_room, db, dummy_user):
+    location = create_location(u'Foo', is_default=True)
+    location.room_name_format = '{building}|{floor}|{number}'
+    assert location._room_name_format == '%1$s|%2$s|%3$s'
+
+    Room(building=1, floor=2, number=3, verbose_name='First amphitheater', location=location, owner=dummy_user)
+    Room(building=1, floor=3, number=4, verbose_name='Second amphitheater', location=location, owner=dummy_user)
+    Room(building=1, floor=2, number=4, verbose_name='Room 3', location=location, owner=dummy_user)
+    db.session.flush()
+    print Room.query.all()
```
ThiefMaster: :see_no_evil:

indico/modules/rb/models/rooms.py
```diff
@@ -314,23 +315,30 @@ def map_url(self):
...
         return self.photo_id is not None
 
+    @hybrid_property
+    def name(self):
+        return self.generate_name()
+
+    @name.expression
+    def name(cls):
+        q = (db.session.query(db.m.Location.room_name_format).
+             filter(db.m.Location.id == cls.location_id).
```
ThiefMaster: `.` at the beginning of each line, not at the end of the previous one

indico/modules/rb/models/rooms.py
```diff
@@ -95,9 +95,10 @@ class Room(versioned_cache(_cache, 'id'), db.Model, Serializer):
         db.Integer,
         db.ForeignKey('roombooking.photos.id')
     )
-    name = db.Column(
+    #: Verbose name for the room (long)
+    verbose_name = db.Column(
         db.String,
-        nullable=False
+        nullable=True
```
ThiefMaster: null vs empty string? I think in most places we use empty strings and `nullable=False`

indico/migrations/versions/20181023_1051_55893713f6b7_add_room_name_format.py
pferreir: 👍

indico/modules/rb/models/locations_test.py
pferreir: Whoops 😳

indico/modules/rb/models/rooms.py
pferreir: I personally think that when doing a DB query that involves checking whether there is a `verbose_name` or not, it's much cleaner to do a `.isnot(None)`. I know we don't normally do that for strings, but in this case we have to do checks on whether the field is set.

indico/modules/rb/models/locations.py
pferreir: Which is a bit weird, cause we don't have spaces between columns but then add them between relationships. Go figure.

indico/modules/rb/models/rooms.py
ThiefMaster: In that case I would add a check constraint to prevent empty strings there. Otherwise I can guarantee that at some point some code will set it to an empty string when editing/updating ;)

indico/modules/rb/models/locations.py
ThiefMaster: I think in newer models we are consistent:

- no whitespace between columns
- no whitespace between relationships
- whitespace between [...columns...]/[...relationships...]

pferreir: Can we document that somewhere? Some code style wiki page or something for stuff that is not done by flake8/ESLint. I think you're the right guy to do that 😉

indico/modules/rb/models/rooms.py
pferreir: Done!

indico/modules/rb_new/operations/rooms.py
```diff
@@ -50,7 +50,7 @@ def _filter_coordinates(query, filters):
 
 def _make_room_text_filter(text):
     text = '%{}%'.format(escape_like(text))
-    columns = ('name', 'site', 'division', 'building', 'floor', 'number', 'comments', 'full_name')
+    columns = ('full_name', 'site', 'division', 'building', 'floor', 'number', 'comments', 'full_name')
```
ThiefMaster: ah we actually had `full_name` at the end of the list, so one of them can go away

PR #3639, opened 2018-10-25
Booking status

mic4ael:


indico/modules/rb_new/client/js/common/bookings/reducers.js
```diff
@@ -27,13 +27,22 @@ export default combineReducers({
...
     }),
     details: (state = {}, action) => {
         switch (action.type) {
             case bookingsActions.BOOKING_DETAILS_RECEIVED: {
-                const {id, ...data} = camelizeKeys(action.data);
-                return {...state, [id]: data};
+                const data = camelizeKeys(action.data);
+                return {...state, [data.id]: data};
+            }
+            case bookingsActions.BOOKING_STATE_UPDATED: {
```
ThiefMaster: since you return all booking data you can just send dispatch `BOOKING_DETAILS_RECEIVED` instead of adding a new action

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -153,3 +153,17 @@ def _process(self):
         booking_details['occurrences'] = occurrences
         booking_details['date_range'] = date_range
         return jsonify(booking_details)
+
+
+class RHBookingStateActions(RHRoomBookingBase):
+    def _process_args(self):
+        self.booking = Reservation.get_one(request.view_args['booking_id'])
+
```
ThiefMaster: missing access checks

indico/modules/rb_new/client/js/common/bookings/selectors.js
```diff
@@ -28,3 +30,6 @@ export const getDetailsWithRoom = createSelector(
     }
 );
 export const hasDetails = (state, {bookingId}) => getDetails(state, {bookingId}) !== undefined;
+export const isBookingChangeInProgress = ({bookings: {requests}}) => (
```
ThiefMaster: usually we only destructurize the top-level state slice and use the usual dot notation for the rest in selectors

indico/modules/rb_new/client/js/common/bookings/reducers.js
mic4ael: I don't return everything. If I want to do it in `BOOKING_DETAILS_RECEIVED` I will have to make sure it doesn't override entirely the object that we received from the `BookingDetails` endpoint. I could also just return state-related data and leave the additional action.

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -160,30 +170,184 @@ class BookingDetailsModal extends React.Component {
...
+                   content={message}
+                   position="bottom center" />
+        );
+    };
+
+    renderDeleteButton = () => {
+        const content = (
+            <>
+                <Message icon="exclamation triangle"
+                         header={Translate.string('Are you sure you want to delete this reservation?')} />
```
ThiefMaster: What about using [Confirm](https://react.semantic-ui.com/addons/confirm/) for this? For something as dangerous as deletion (even if it's admin-only) I think I'd prefer something a bit more intrusive.

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -160,30 +170,184 @@ class BookingDetailsModal extends React.Component {
...
+    };
+
+    renderBookingStatus = ({isPending, isAccepted, isCancelled, isRejected, rejectionReason}) => {
+        let color, status, icon, message;
+
+        if (isPending) {
+            icon = <Icon name="wait" />;
+            color = 'yellow';
+            status = Translate.string('Pending Confirmation');
+            message = Translate.string('The booking is subject to acceptance by the room owner');
```
meluru: what about changing all the `The booking...` to `This booking...` ?

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -197,18 +361,22 @@ class BookingDetailsModal extends React.Component {
...
         return (
             <Modal open onClose={this.handleCloseModal} size="large" closeIcon>
                 <Modal.Header styleName="booking-modal-header">
-                    <Translate>Booking Details</Translate>
+                    <span styleName="header-text">
+                        <Translate>Booking Details</Translate>
+                    </span>
+                    <span styleName="booking-status">
+                        {/* eslint-disable-next-line react/destructuring-assignment */}
+                        {this.renderBookingStatus(this.props.booking)}
```
ThiefMaster: Why pass a prop to a method of the same component? You could just do this inside `renderBookingStatus`:

    const {booking: {isPending, isAccepted, isCancelled, isRejected, rejectionReason}} = this.props;

indico/modules/rb_new/client/js/common/bookings/actions.js
```diff
@@ -27,6 +29,15 @@ export const FETCH_BOOKING_DETAILS_REQUEST = 'bookings/FETCH_BOOKING_DETAILS_REQ
 export const FETCH_BOOKING_DETAILS_SUCCESS = 'bookings/FETCH_BOOKING_DETAILS_SUCCESS';
 export const FETCH_BOOKING_DETAILS_ERROR = 'bookings/FETCH_BOOKING_DETAILS_ERROR';
 
+export const BOOKING_STATE_CHANGE_REQUEST = 'bookings/BOOKING_STATE_CHANGE_REQUEST';
+export const BOOKING_STATE_CHANGE_SUCCESS = 'bookings/BOOKING_STATE_CHANGE_SUCCESS';
+export const BOOKING_STATE_CHANGE_ERROR = 'bookings/BOOKING_STATE_CHANGE_ERROR';
+export const BOOKING_STATE_UPDATED = 'bookings/BOOKING_STATE_UPDATED';
+
+export const BOOKING_DELETE_REQUEST = 'bookings/BOOKING_DELETE_REQUEST';
```
ThiefMaster: `DELETE_BOOKING_...`

indico/modules/rb_new/client/js/common/bookings/actions.js
```diff
@@ -40,3 +51,25 @@ export function fetchBookingDetails(id) {
 }
 
 export const openBookingDetails = (bookingId) => modalActions.openModal('booking-details', bookingId);
+
+export function changeBookingState(id, action, params = {}) {
+    return async (dispatch) => {
```
ThiefMaster: `return ajaxAction(...);` - no need for the extra wrapper since you do not need to access the state

indico/modules/rb_new/client/js/modules/calendar/reducers.js
```diff
@@ -24,6 +24,7 @@ import {requestReducer} from 'indico/utils/redux';
 import {actions as bookRoomActions} from '../../modules/bookRoom';
 import * as actions from '../../actions';
 import * as calendarActions from './actions';
+import * as bookingActions from '../../common/bookings/actions';
```
ThiefMaster: import {actions as bookingActions} from '../../common/bookings';

indico/modules/rb_new/client/js/modules/calendar/reducers.js
```diff
@@ -76,17 +126,44 @@ export default combineReducers({
...
+                const {bookingId, roomId} = camelizeKeys(action.data);
+                const {rows} = state;
+                return {...state, rows: filterDeletedBooking(rows, bookingId, roomId)};
+            }
+            case bookingActions.BOOKING_STATE_UPDATED: {
+                const {booking: {id, roomId}, bookingState} = camelizeKeys(action.data);
+                const {rows} = state;
+                let newRows;
+
+                if (['rejected', 'cancelled'].indexOf(bookingState) !== -1) {
```
ThiefMaster: do you really consider that more readable than a simple `||`?

    if (bookingState === 'rejected' || bookingState === 'cancelled')

BTW, ES6 has `[...].includes(somevalue)`

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -153,3 +158,50 @@ def _process(self):
...
+        RHRoomBookingBase._check_access(self)
+        if self.action == 'approve':
+            if not self.booking.can_be_accepted(session.user):
+                raise Forbidden
+        elif self.action == 'reject':
+            if not self.booking.can_be_rejected(session.user):
+                raise Forbidden
+        elif self.action == 'cancel':
+            if not self.booking.can_be_cancelled(session.user):
+                raise Forbidden
```
ThiefMaster:
```
else:
    raise BadRequest  # or Forbidden
```

just to avoid bad surprises if we ever make a type

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -153,3 +158,50 @@ def _process(self):
         booking_details['occurrences'] = occurrences
         booking_details['date_range'] = date_range
         return jsonify(booking_details)
+
+
+class RHBookingStateActions(RHRoomBookingBase):
+    def _check_access(self):
```
ThiefMaster: move this after `_process_args`

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -153,3 +158,50 @@ def _process(self):
...
+    def _process_args(self):
+        self.booking = Reservation.get_one(request.view_args['booking_id'])
+        self.action = request.view_args['action']
+
+    def _process(self):
+        if self.action == 'approve':
+            self.booking.accept(session.user)
+            state = 'approved'
+        elif self.action == 'reject':
+            reason = request.json['reason']
```
ThiefMaster: let's use webargs for this (could be `self.reject()`, with `reject` being decorated with the webargs decorator)

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -153,3 +158,50 @@ def _process(self):
...
+    def _check_access(self):
+        RHRoomBookingBase._check_access(self)
+        if not self.booking.can_be_deleted(session.user):
+            raise Forbidden
+
+    def _process_args(self):
+        self.booking = Reservation.get_one(request.view_args['booking_id'])
+
+    def _process(self):
+        booking_id, room_id = self.booking.id, self.booking.room.id
```
ThiefMaster: why?

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -160,30 +170,184 @@ class BookingDetailsModal extends React.Component {
...
+        const cancelButton = (
+            <Button type="button"
+                    icon="remove circle"
+                    size="small"
+                    loading={actionInProgress === 'cancel' && bookingStateChangeInProgress}
+                    disabled={actionInProgress !== 'cancel' && bookingStateChangeInProgress}
+                    content={<Translate>Cancel booking</Translate>} />
+        );
+
+        const validate = ({reason}) => {
```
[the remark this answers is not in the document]
mic4ael: mhm, didn't know about it! Good to know.

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -160,30 +170,184 @@ class BookingDetailsModal extends React.Component {
...
+                    <Popup trigger={cancelButton}
+                           position="right center"
+                           on="click">
+                        <FinalForm onSubmit={(data) => this.changeState('cancel', data)}
+                                   validate={validate}
+                                   render={renderForm(
+                                       Translate.string('Provide the cancellation reason'),
+                                       Translate.string('Cancel the booking')
+                                   )}
+                                   initialValues={{reason: ''}} />
```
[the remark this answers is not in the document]
mic4ael: It is, because we use `formatters.trim` there which will not operate on `null`

indico/modules/rb_new/client/js/modules/calendar/reducers.js
mic4ael: Yeah, I know about `includes` I just remember I had some problems with it at some point.. however I don't remember what it was exactly.

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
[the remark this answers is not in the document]
ThiefMaster: hmm i thought i had fixed it to not do anything on falsy values..

indico/modules/rb_new/controllers/backend/bookings.py
mic4ael: I need this information in order to properly update `Calendar` redux state

ThiefMaster: I meant the unpacking..

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
mic4ael: Oh, I think my brain just froze at the point of writing this.

indico/modules/rb_new/controllers/backend/bookings.py
mic4ael: What is wrong with it? I am just saving one line here which looks much better (at least to me).

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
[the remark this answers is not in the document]
mic4ael: yeah, indeed, you don't perform any action on `undefined`

indico/modules/rb_new/client/js/modules/calendar/reducers.js
mic4ael: Then why do we have sth similar above?

ThiefMaster: `calendarActions` is from the same module

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -153,3 +158,57 @@ def _process(self):
...
+
+
+class RHBookingStateActions(RHRoomBookingBase):
+    def _process_args(self):
+        self.booking = Reservation.get_one(request.view_args['booking_id'])
+        self.action = request.view_args['action']
+
+    def _check_access(self):
+        RHRoomBookingBase._check_access(self)
+        if self.action == 'approve':
```
ThiefMaster: This could be simplified like this:

```
funcs = {'approve': self.booking.can_be_accepted,
         'reject': self.booking.can_be_rejected,
         'cancel': self.booking.can_be_cancelled}
if not funcs[self.action](session.user):
    raise Forbidden
```

(or simply merge the inner ifs with the outer ones)

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -160,33 +171,185 @@ class BookingDetailsModal extends React.Component {
...
+            <Modal.Actions>
+                {canCancel && (
+                    <>
+                        <Button type="button"
+                                icon="remove circle"
+                                size="small"
+                                onClick={() => this.showConfirm('cancel')}
+                                loading={actionInProgress === 'cancel' && bookingStateChangeInProgress}
+                                disabled={bookingStateChangeInProgress}
+                                content={<Translate>Cancel booking</Translate>} />
```
mvidalgarcia: Is this going to cancel all the occurrences (if there are)? For me as user it's not clear.

ThiefMaster: Yes (we still need a way to cancel individual ones). The confirmation prompt should clarify this.

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -160,33 +171,185 @@ class BookingDetailsModal extends React.Component {
...
+                        <Confirm header={Translate.string('Confirm cancellation')}
+                                 content={Translate.string('Are you sure you want to cancel this reservation?')}
```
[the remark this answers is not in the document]
ThiefMaster: I'd go for "booking" everywhere in the UI (even in the backend we usually use "booking" now and not the old "reservation" except in the model's name).

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -35,15 +39,22 @@ import './BookingDetailsModal.module.scss';
...
 
     static defaultProps = {
         onClose: () => {}
     };
 
     state = {
-        occurrencesVisible: false
+        occurrencesVisible: false,
+        actionInProgress: null,
+        openConfirm: null,
```
ThiefMaster: `activeConfirmation`? the current name sounds more like a function

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -160,33 +171,186 @@ class BookingDetailsModal extends React.Component {
...
+                        <Button type="button"
+                                icon="cancel"
+                                size="small"
+                                onClick={() => this.showConfirm('cancel')}
+                                loading={actionInProgress === 'cancel' && bookingStateChangeInProgress}
+                                disabled={bookingStateChangeInProgress}
+                                content={<Translate>Cancel booking</Translate>} />
+                        <Confirm header={Translate.string('Confirm cancellation')}
+                                 content={Translate.string('Are you sure you want to cancel this booking?' +
+                                                           ' This will cancel all occurrences of this booking.')}
```
ThiefMaster: what about keeping the trailing space in the previous line like we do in python?

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
```diff
@@ -160,33 +171,186 @@ class BookingDetailsModal extends React.Component {
...
+            <Modal.Actions>
+                {canCancel && (
+                    <>
+                        <Button type="button"
+                                icon="cancel"
+                                size="small"
+                                onClick={() => this.showConfirm('cancel')}
+                                loading={actionInProgress === 'cancel' && bookingStateChangeInProgress}
+                                disabled={bookingStateChangeInProgress}
+                                content={<Translate>Cancel booking</Translate>} />
```
ThiefMaster: Does `Translate.string()` not work here?  most other props below use that one

indico/modules/rb_new/client/js/common/bookings/actions.js
```diff
@@ -27,16 +29,41 @@ export const FETCH_BOOKING_DETAILS_REQUEST = 'bookings/FETCH_BOOKING_DETAILS_REQ
...
+    );
 }
 
 export const openBookingDetails = (bookingId) => modalActions.openModal('booking-details', bookingId);
+
+export function changeBookingState(id, action, params = {}) {
+    return ajaxAction(
+        () => indicoAxios.post(bookingStateActionsURL({booking_id: id, action}), params),
+        BOOKING_STATE_CHANGE_REQUEST,
+        [BOOKING_STATE_CHANGE_SUCCESS, BOOKING_STATE_UPDATED],
```
ThiefMaster: usually we trigger the "updated" action before the "success" one

mic4ael: why? what is the reasoning behind that? I would say it is more correct to send the action stating that  the request was successful before any other action.

indico/modules/rb_new/client/js/common/bookings/BookingDetailsModal.jsx
mic4ael: It should. Don't know why I did it like that here.

indico/modules/rb_new/client/js/common/bookings/actions.js
ThiefMaster: When stepping through with the redux devtools for example, the "loading" state only goes away once the new data is available.
Also, in case we had something rendered based on the fact that a request was successful, this would be guaranteed to happen only once the new data is available (since we usually do not use the `_SUCCESS` event to update the store).

FWIW, if we had some kind of batched actions this proably wouldn't matter. But last time I checked, they had some drawbacks when it came to middlewares (since depending on the batching implemention they'd see only the batch action but not the actual underlying actions)

PR #3642, opened 2018-10-26
Add event links

### LINKED ISSUE #3625: Add link back to event
In bookings that were created from an event, we should have a link back to the latter.

pferreir:
Example:

image

Closes #3625

### DIFF
--- a/indico/modules/rb_new/client/js/common/bookings/BookingEventLink.jsx
+++ b/indico/modules/rb_new/client/js/common/bookings/BookingEventLink.jsx
@@ -1,0 +1,93 @@
+/* This file is part of Indico.
+ * Copyright (C) 2002 - 2018 European Organization for Nuclear Research (CERN).
+ *
+ * Indico is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Indico is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Message, Placeholder, Segment, Icon} from 'semantic-ui-react';
+import getBookingEventDataURL from 'indico-url:rooms_new.booking_event_data';
+import {Translate} from 'indico/react/i18n';
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+import camelizeKeys from 'indico/utils/camelize';
+
+import './BookingEventLink.module.scss';
+
+
+export default class BookingEventLink extends React.PureComponent {
+    static propTypes = {
+        bookingId: PropTypes.number.isRequired,
+    };
+
+    state = {
+        loaded: false,
+    };
+
+    componentDidMount() {
+        const {bookingId} = this.props;
+        this.fetchEventData(bookingId);
+    }
+
+    async fetchEventData(bookingId) {
+        try {
+            const {data} = await indicoAxios.get(getBookingEventDataURL({booking_id: bookingId}));
+            this.setState({
+                loaded: true,
+                data: camelizeKeys(data)
+            });
+        } catch (error) {
+            handleAxiosError(error);
+        }
+    }
+
+    renderEventInfo(data) {
+        const {url, title, canAccess} = data;
+        return (
+            <Message icon color="teal">
+                <Icon name="linkify" />
+                <Message.Content>
+                    {canAccess ? (
+                        <>
+                            <Translate>This booking is linked to an event:</Translate>
+                            <div styleName="booking-event-link">
+                                <a href={url}>{title}</a>
+                            </div>
+                        </>
+                    ) : (
+                        <Translate>This booking is linked to an event.</Translate>
+                    )}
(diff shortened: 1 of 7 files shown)


indico/modules/rb_new/client/js/common/bookings/BookingEventLink.jsx
```diff
@@ -0,0 +1,89 @@
...
+import {Message, Placeholder, Segment, Icon} from 'semantic-ui-react';
+import getBookingEventData from 'indico-url:rooms_new.booking_event_data';
+import {Translate} from 'indico/react/i18n';
+import {indicoAxios} from 'indico/utils/axios';
+import camelizeKeys from 'indico/utils/camelize';
+
+import './BookingEventLink.module.scss';
+
+
+export default class BookingEventLink extends React.Component {
```
ThiefMaster: I think this can be a `PureComponent`

indico/modules/rb_new/client/js/common/bookings/BookingEventLink.jsx
```diff
@@ -0,0 +1,89 @@
...
+        loaded: false
+    };
+
+    componentDidMount() {
+        const {bookingId} = this.props;
+        this.fetchEventData(bookingId);
+    }
+
+    async fetchEventData(bookingId) {
+        const {data} = await indicoAxios.get(getBookingEventData({booking_id: bookingId}));
```
ThiefMaster: try..catch?

indico/modules/rb_new/client/js/common/bookings/BookingEventLink.jsx
```diff
@@ -0,0 +1,89 @@
...
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Message, Placeholder, Segment, Icon} from 'semantic-ui-react';
+import getBookingEventData from 'indico-url:rooms_new.booking_event_data';
```
ThiefMaster: `getBookingEventDataURL`?

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -153,3 +154,12 @@ def _process(self):
         booking_details['occurrences'] = occurrences
         booking_details['date_range'] = date_range
         return jsonify(booking_details)
+
+
+class RHBookingEventData(RHBookingDetails):
```
ThiefMaster: what about adding a `RHBookingBase` that contains the `_process_args` that's currently in `RHBookingDetails` and then use that class as a base for both `RHBookingDetails` and `RHBookingEventData`?

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -153,3 +154,12 @@ def _process(self):
...
+    def _process(self):
+        if self.booking.event is None:
+            raise NotFound
```
ThiefMaster: hmm... 404 vs response indicating that there is no booking. i guess with 404 the error handling in the JS is a bit trickier

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -153,3 +154,12 @@ def _process(self):
...
+        if not self.booking.event.can_access(session.user):
+            return jsonify({'can_access': False})
```
ThiefMaster: `jsonify(can_access=False)`

indico/modules/rb_new/schemas.py
```diff
@@ -77,6 +77,13 @@ class Meta:
         fields = ('id', 'booking_reason', 'booked_for_name')
 
 
+class ReservationEventDataSchema(mm.ModelSchema):
```
ThiefMaster: don't use `ModelSchema` if you don't base it on an actual modal class (specified in the `class Meta`); instead use a normal `mm.Schema`.. Or you could use a ModelSchema with `model = Event` and simply restrict the fields. That way you don't have to define the types of the fields since they'll be taken from the Event model.

indico/modules/rb_new/controllers/backend/bookings.py
pferreir: Well, we will never handle it in JS, since we only send the request in the first place if we know there is an event.

indico/modules/rb_new/controllers/backend/bookings.py
pferreir: Thought of it, but thought it was an overkill.

indico/modules/rb_new/controllers/backend/bookings.py
ThiefMaster: ah even better, hadn't noticed that the request is conditional. then i agree that 404 is a good option

indico/modules/rb_new/client/js/common/bookings/BookingEventLink.jsx
pferreir: What's the advantage of that?

ThiefMaster: the component only updates when state/props changed instead of (by default) whenever the parent updates

pferreir: Is that still the case? [The docs](https://reactjs.org/docs/react-api.html#reactpurecomponent) are not very clear about that.

indico/modules/rb_new/schemas.py
pferreir: Yeah, copy-paste.

indico/modules/rb_new/client/js/common/bookings/BookingEventLink.jsx
pferreir: Anyway, changed.

PR #3641, opened 2018-10-26
Replace has_projector etc with something more flexible.

ThiefMaster:
### TODO

- Audiovisual plugin: update to check for feature (add plugin setting to select the feature) instead of looking for equipment with a special name


ThiefMaster:
### Feature filter buttons

image


Could also be a separate bar, but that might look weird when there are no features and that bar would contain only the favorites filter.

indico/modules/rb/models/room_features.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+    name = db.Column(
+        db.String,
+        nullable=False,
+        index=True,
+        unique=True
+    )
+    title = db.Column(
+        db.String,
+        nullable=False,
```
mic4ael: you put trailing comma here and below, but e.g. you don't have it above

indico/modules/rb_new/client/js/common/roomSearch/queryString.js
```diff
@@ -85,6 +85,21 @@ export const rules = {
...
     },
+    feat: {
+        validator: () => true,
+        stateField: {
+            serialize: ({filters: {features}}) => features,
+            parse: (value, state) => {
+                if (!Array.isArray(value)) {
+                    value = [value];
+                }
+                if (!state.filters) {
```
mic4ael: When can it be `true`? Don't we set it to an empty object?

indico/modules/rb_new/client/js/common/rooms/selectors.js
```diff
@@ -15,15 +15,68 @@
...
+    features => features.filter(f => f.showFilterButton).map(f => _.pick(f, ['icon', 'name', 'title']))
+);
+
+export const getAllRooms = createSelector(
+    ({rooms}) => rooms.rooms,
+    getEquipmentTypes,
+    (rawRooms, equipmentTypes) => {
+        equipmentTypes = equipmentTypes.reduce((obj, eq) => ({...obj, [eq.id]: eq}), {});
+        return _.fromPairs(rawRooms.map(room => {
+            const {available_equipment: equipment, ...roomData} = room;
```
mic4ael: Shouldn't we somewhere use `camelizeKeys` to change `available_equipment` to `availableEquipment`?

indico/modules/rb_new/client/js/common/roomSearch/queryString.js
ThiefMaster: just copied from equipments tbh.

pferreir:
> ### Feature filter buttons
> 
> image
> 
> Could also be a separate bar, but that might look weird when there are no features and that bar would contain only the favorites filter.

Do we really want these separate buttons?

ThiefMaster:
It's configurable for each feature whether to show a search button or not. IMO it's useful though since it lets people filter by those features quickly, instead of having to go into the equipment dropdown. Probably I would only enable it for videoconference and maybe projector since webcast/recording is probably not used that much compared to the rest

indico/modules/rb_new/client/js/common/roomSearch/queryString.js
ThiefMaster: It's based on an [empty object](https://github.com/indico/redux-router-querystring/blob/master/src/validation.js#L19), which then gets merged into the redux state later.

indico/modules/rb_new/client/js/common/rooms/selectors.js
ThiefMaster: Something for a separate commit since it'd just add lots of clutter to this PR.

ThiefMaster:
image

indico/modules/rb_new/client/js/common/rooms/selectors.js
```diff
@@ -15,15 +15,84 @@
...
 
 export const hasLoadedEquipmentTypes = ({rooms}) => rooms.requests.equipmentTypes.state === RequestState.SUCCESS;
-export const getEquipmentTypes = ({rooms}) => rooms.equipmentTypes;
+const getEquipmentTypes = ({rooms}) => rooms.equipmentTypes;
+const getFeaturesMapping = createSelector(
+    getEquipmentTypes,
+    equipmentTypes => {
+        const features = {};
+        equipmentTypes
+            .forEach(({name: eqName, features: eqFeatures}) => {
```
mic4ael: Honestly, it is not more readable to me. Why did you put it in the next line? It seems like it could fit perfectly fine in the previous one - that would save one indentation level also.

ThiefMaster:
For those with a CERN roombooking DB copy, run these queries to get the correct features and equipment associations:

```sql
INSERT INTO roombooking.features VALUES (1, 'projector', 'Projector', 'film');
INSERT INTO roombooking.features VALUES (3, 'vc', 'Video conference', 'video camera');
INSERT INTO roombooking.features VALUES (2, 'webcast-recording', 'Webcast/Recording', 'podcast');
INSERT INTO roombooking.equipment_features VALUES (5, 1);
INSERT INTO roombooking.equipment_features VALUES (27, 1);
INSERT INTO roombooking.equipment_features VALUES (28, 1);
INSERT INTO roombooking.equipment_features VALUES (9, 2);
INSERT INTO roombooking.equipment_features VALUES (20, 3);
INSERT INTO roombooking.equipment_features VALUES (8, 3);
SELECT pg_catalog.setval('roombooking.features_id_seq', 3, true);
```

PR #3647, opened 2018-11-01
Clear calendar timeline when loading new data

ThiefMaster:
That way we always see the loading indicator which was previously pushed out of the visible are if there were more than a few rooms.

Also fixed the query string mapping for the calendar view mode.

---

image

In case of weekly/monthly we don't have data to show in the header, so we keep it empty. IMHO it's not worth adding the extra complexity of calculating the days for the currently-loading month/week in JS just for this.

image

### DIFF
--- a/indico/modules/rb_new/client/js/util.jsx
+++ b/indico/modules/rb_new/client/js/util.jsx
@@ -103,9 +103,32 @@ export const pushStateMergeProps = (stateProps, dispatchProps, ownProps) => {
     };
 };
 
+/**
+ * Defines a query string state field for a boolean that is only added
+ * to the query string if it's true.
+ * @param {string} name - Name of the state field.
+ */
 export function boolStateField(name) {
     return {
         serialize: state => _.get(state, name) || null,
+        parse: (value, state) => {
+            _.set(state, name, value);
+        }
+    };
+}
+
+/**
+ * Defines a query string state field that is only added to the query
+ * string if the value doesn't match the default.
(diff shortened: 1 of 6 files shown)


indico/modules/rb_new/client/js/util.jsx
```diff
@@ -103,6 +103,11 @@ export const pushStateMergeProps = (stateProps, dispatchProps, ownProps) => {
     };
 };
 
+/**
+ * Defines a query string state field for a boolean that will only added
```
mic4ael: be added/that is only added ?

indico/modules/rb_new/client/js/util.jsx
```diff
@@ -112,6 +117,24 @@ export function boolStateField(name) {
...
+/**
+ * Defines a query string state field that is only added to the query
+ * string if the value doesn't match the default.
+ * @param {string} name - Name of the state field.
+ * @param defaultValue - The default value that will not go to the query string.
+ */
+export function defaultStateField(name, defaultValue) {
+    return {
+        serialize: state => {
+            const value = _.get(state, name, null);
```
mic4ael: why do you have to use `_.get` here? isn't it enough to use `return state[name] === defaultValue ? null : value`? Does it prevent from cases where `defaultValue` is `undefined`?

ThiefMaster: No, the state field can use dot notation

PR #3652, opened 2018-11-06
Room statistics

### LINKED ISSUE #3617: Add Room statistics
Currently missing, as you can see:

image

Important statistics:
 * % occupancy of room (during working hours as defined by `Location.working_time_periods`) in the last 7, 30 and 365 days (or similar);
 * Number of bookings this week/month/year;
 * Any other ideas?

mvidalgarcia:
closes #3617

### DIFF
--- a/indico/modules/rb/statistics.py
+++ b/indico/modules/rb/statistics.py
@@ -30,8 +30,8 @@ def calculate_rooms_booked_time(rooms, start_date=None, end_date=None):
     # Reservations on working days
     reservations = Reservation.find(Reservation.room_id.in_(r.id for r in rooms),
                                     db.extract('dow', ReservationOccurrence.start_dt).between(1, 5),
-                                    ReservationOccurrence.start_dt >= start_date,
-                                    ReservationOccurrence.end_dt <= end_date,
+                                    db.cast(ReservationOccurrence.start_dt, db.Date) >= start_date,
+                                    db.cast(ReservationOccurrence.end_dt, db.Date) <= end_date,
                                     ReservationOccurrence.is_valid,
                                     _join=ReservationOccurrence)
 
--- a/indico/modules/rb_new/blueprint.py
+++ b/indico/modules/rb_new/blueprint.py
@@ -46,8 +46,8 @@ _bp.add_url_rule('/api/rooms/search', 'search_rooms', rooms.RHSearchRooms)
 _bp.add_url_rule('/api/rooms/<int:room_id>/availability', 'room_availability', rooms.RHRoomAvailability)
 _bp.add_url_rule('/api/rooms/<int:room_id>/attributes', 'room_attributes', rooms.RHRoomAttributes)
 _bp.add_url_rule('/api/rooms/<int:room_id>/timeline', 'timeline', bookings.RHTimeline)
-_bp.add_url_rule('/api/suggestions', 'suggestions', bookings.RHRoomSuggestions)
 _bp.add_url_rule('/api/rooms/<int:room_id>/stats', 'room_stats', rooms.RHRoomStats)
+_bp.add_url_rule('/api/suggestions', 'suggestions', bookings.RHRoomSuggestions)
 
 # User
 _bp.add_url_rule('/api/user/', 'user_info', misc.RHUserInfo)
--- a/indico/modules/rb_new/blueprint.py
+++ b/indico/modules/rb_new/blueprint.py
@@ -47,6 +47,7 @@ _bp.add_url_rule('/api/rooms/<int:room_id>/availability', 'room_availability', r
 _bp.add_url_rule('/api/rooms/<int:room_id>/attributes', 'room_attributes', rooms.RHRoomAttributes)
 _bp.add_url_rule('/api/rooms/<int:room_id>/timeline', 'timeline', bookings.RHTimeline)
 _bp.add_url_rule('/api/suggestions', 'suggestions', bookings.RHRoomSuggestions)
+_bp.add_url_rule('/api/rooms/<int:room_id>/stats', 'room_stats', rooms.RHRoomStats)
 
 # User
 _bp.add_url_rule('/api/user/', 'user_info', misc.RHUserInfo)
--- a/indico/modules/rb_new/client/js/common/rooms/RoomStats.jsx
+++ b/indico/modules/rb_new/client/js/common/rooms/RoomStats.jsx
@@ -78,12 +78,15 @@ export default class RoomStats extends React.PureComponent {
 
     render() {
         const {loaded, data} = this.state;
-        const mapping = {
-            times_booked: Translate.string('Times booked'),
-            occupancy: Translate.string('Occupancy'),
+        const sections = {
+            times_booked: {title: Translate.string('Times booked')},
+            occupancy: {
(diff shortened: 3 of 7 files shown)


indico/modules/rb_new/client/js/common/rooms/RoomStats.jsx
```diff
@@ -0,0 +1,100 @@
...
+                        </Placeholder>
+                    </Grid.Column>
+                ))}
+            </Grid>
+        );
+    }
+
+    render() {
+        const {loaded, data} = this.state;
+        return loaded ? (
```
ThiefMaster:
```jsx
if (!loaded) {
    return this.renderPlaceholder();
}
```

indico/modules/rb_new/client/js/common/rooms/RoomStats.jsx
```diff
@@ -0,0 +1,100 @@
...
+            <div styleName="room-stats">
+                {Object.entries(data).map(([key, {title, values, note}]) => (
+                    <div key={key} styleName="stats-box">
+                        <div styleName="title">{Translate.string(title)}{!!note && '*'}</div>
+                        {values.map(({days, value}) => (
+                            <div styleName="value-box" key={value}>
+                                <div styleName="days">
+                                    <Translate>Last <Param name="days" value={days} /> days</Translate>
```
ThiefMaster: This is a case for `PluralTranslate`

indico/modules/rb_new/operations/rooms.py
```diff
@@ -90,6 +94,39 @@ def get_managed_room_ids(user):
         return {r.id for r in Room.get_owned_by(user)}
 
 
+# @memoize_redis(3600)
+def get_room_statistics(room):
+    data = {
+        'count': {
+            'title': 'Times booked',
```
ThiefMaster: i18n.. best to just use an identifier here (e.g. `times_booked`) and then keep the human-friendly display on the client side

mvidalgarcia: You mean having a mapping const in the client like this?
```js
const mapping = {
    times_booked: 'Times booked',
    occupancy: 'Occupancy',
}
```

ThiefMaster: yes (but with i18n ofc)

indico/modules/rb/statistics.py
```diff
@@ -30,8 +30,8 @@ def calculate_rooms_booked_time(rooms, start_date=None, end_date=None):
     # Reservations on working days
     reservations = Reservation.find(Reservation.room_id.in_(r.id for r in rooms),
                                     db.extract('dow', ReservationOccurrence.start_dt).between(1, 5),
-                                    ReservationOccurrence.start_dt >= start_date,
-                                    ReservationOccurrence.end_dt <= end_date,
+                                    ReservationOccurrence.start_dt >= datetime.combine(start_date, time.min),
+                                    ReservationOccurrence.end_dt <= datetime.combine(end_date, time.max),
```
pferreir: I'm not sure that this is the best solution. You're accepting a `datetime` but then just ignoring its time part. It seems a bit misleading to me.
Maybe we should use a `date` instead of `datetime` for `start_date` and `end_date`? This way it will be clear that time isn't taken into account.

indico/modules/rb_new/operations/rooms.py
```diff
@@ -90,6 +95,40 @@ def get_managed_room_ids(user):
...
+    data = {
+        'count': {
+            'id': 'times_booked',
+            'values': [],
+            'note': None
+        },
+        'percentage': {
+            'id': 'occupancy',
+            'values': [],
+            'note': _('excluding weekends')
```
pferreir: If the result is memoized on Redis, the translated string will be as well. Either you add it later in the RH or directly on the client.

ThiefMaster: huh weird, i could have sworn i already commented somewhere that this should be handled on the client..

mvidalgarcia:
Updated.

indico/modules/rb_new/blueprint.py
```diff
@@ -47,6 +47,7 @@
 _bp.add_url_rule('/api/rooms/<int:room_id>/attributes', 'room_attributes', rooms.RHRoomAttributes)
 _bp.add_url_rule('/api/rooms/<int:room_id>/timeline', 'timeline', bookings.RHTimeline)
 _bp.add_url_rule('/api/suggestions', 'suggestions', bookings.RHRoomSuggestions)
+_bp.add_url_rule('/api/rooms/<int:room_id>/stats', 'room_stats', rooms.RHRoomStats)
```
ThiefMaster: what about moving this one line up so it's there together with the other room-specific endpoints?

indico/modules/rb_new/client/js/common/rooms/RoomStats.jsx
```diff
@@ -0,0 +1,117 @@
...
+        );
+    }
+
+    render() {
+        const {loaded, data} = this.state;
+        const mapping = {
+            times_booked: Translate.string('Times booked'),
+            occupancy: Translate.string('Occupancy'),
+        };
+        const hasStats = Object.entries(data)
```
ThiefMaster: `Object.values()` since you never use the keys...

indico/modules/rb_new/client/js/common/rooms/RoomStats.jsx
```diff
@@ -0,0 +1,117 @@
...
+                        </Placeholder>
+                    </Grid.Column>
+                ))}
+            </Grid>
+        );
+    }
+
+    render() {
+        const {loaded, data} = this.state;
+        const mapping = {
```
ThiefMaster: `titles` or `sections` (the latter if you move the note texts in here as well)?

indico/modules/rb_new/operations/rooms.py
```diff
@@ -90,6 +94,40 @@ def get_managed_room_ids(user):
...
+    data = {
+        'count': {
+            'id': 'times_booked',
+            'values': [],
+            'note': None
+        },
+        'percentage': {
+            'id': 'occupancy',
+            'values': [],
+            'note': 'excluding weekends'
```
ThiefMaster: this should be on the client side, not returned by the server

indico/modules/rb_new/client/js/common/rooms/RoomStats.jsx
```diff
@@ -0,0 +1,118 @@
...
+                        </Placeholder>
+                    </Grid.Column>
+                ))}
+            </Grid>
+        );
+    }
+
+    render() {
+        const {loaded, data} = this.state;
+        const sections = {
```
ThiefMaster: wouldn't something like this be better?

```
        const sections = {
            times_booked: {title: Translate.string('Times booked')},
            occupancy: {title: Translate.string('Occupancy'), note: Translate.string('excluding weekends')},
        };
```

mvidalgarcia:
This PR got forgotten somehow, thanks @mic4ael for the reminder! Just rebased it, do you have any further feedback on it or it can be merged?

mic4ael:
As far as I am concerned I haven't noticed anything to comment on ;)

PR #3655, opened 2018-11-08
Highlight map markers when rooms are hovered

### LINKED ISSUE #3621: Highlight on map hovered rooms
This could also work in the opposite direction, even if it depends on whether the room is currently visible, to be effective.
Map markers should also open a tooltip when clicked, which allows for the actions that are present normally on the room.

pferreir:
highlight

Closes #3621


indico/modules/rb_new/client/js/common/map/selectors.js
```diff
@@ -44,3 +44,5 @@ export const makeGetMapData = (namespace) => ({[namespace]: stateSlice}) => ({
 });
 
 export const makeIsMapSearchEnabled = (namespace) => ({[namespace]: stateSlice}) => stateSlice.map.search;
+
+export const getHoveredRoom = ({map: {ui}}) => ui.hoveredRoom;
```
ThiefMaster: I would go for this:

```js
export const getHoveredRoom = ({map}) => map.ui.hoveredRoom;
```

Like this it's consistent with all the other selectors and it's easy to find accesses to a given state slice by searching for "foo.bar" in the code

PR #3659, opened 2018-11-14
Some Booking creation improvements

pferreir:
Two main enhancements:
 * Warning whenever user about to make a pre-booking
 * Link to newly created booking now works;

image

### DIFF
--- a/indico/modules/rb_new/client/js/common/map/RoomBookingMap.jsx
+++ b/indico/modules/rb_new/client/js/common/map/RoomBookingMap.jsx
@@ -29,6 +29,65 @@ import 'leaflet/dist/leaflet.css';
 import './RoomBookingMap.module.scss';
 
 
+function groupIconCreateFunction(cluster) {
+    const highlight = cluster.getAllChildMarkers().some(m => m.options.highlight);
+    return Leaflet.divIcon({
+        html: `<span>${cluster.getChildCount()}</span>`,
+        className: `marker-cluster marker-cluster-small rb-map-cluster ${highlight ? 'highlight' : ''}`,
+        iconSize: Leaflet.point(40, 40, true),
+    });
+}
+
+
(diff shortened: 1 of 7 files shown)


indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -133,10 +133,10 @@ def _process(self, args):
 
         serialized_occurrences = serialize_occurrences(group_by_occurrence_date(resv.occurrences.all()))
         if is_prebooking:
-            data = {'pre_bookings': serialized_occurrences}
+            data = {'pre_booking_occurrences': serialized_occurrences}
         else:
-            data = {'bookings': serialized_occurrences}
-        return jsonify(room_id=room.id, **data)
+            data = {'booking_occurrences': serialized_occurrences}
```
ThiefMaster: the reducer doesn't know about this rename so new bookings don't show up on the calendar now

indico/modules/rb_new/client/js/common/map/RoomBookingMap.jsx
```diff
@@ -29,6 +29,63 @@ import 'leaflet/dist/leaflet.css';
...
+    hoveredRoomId: PropTypes.number
+};
+
+_MapMarkers.defaultProps = {
+    rooms: [],
+    clusterProps: {},
+    hoveredRoomId: null
+};
+
+const MapMarkers = connect(state => ({
```
ThiefMaster: same formatting as for other connects?

```js
const MapMarkers = connect(
    state => ({
        hoveredRoomId: mapSelectors.getHoveredRoom(state),
    })
)(_MapMarkers);
```

indico/modules/rb_new/controllers/backend/bookings.py
pferreir: This is actually not intentional, now that I see it. I forgot to undo these changes.

PR #3666, opened 2018-11-21
Refresh data after leaving RB administration

ThiefMaster:
Feels a bit hacky to me but otherwise we'd need to add extra logic to every single admin action that changes data (and it'd also result in reloading data more often than needed).

### DIFF
--- a/indico/modules/rb_new/client/js/setup.jsx
+++ b/indico/modules/rb_new/client/js/setup.jsx
@@ -10,6 +10,7 @@ import createRBStore, {history} from './store';
 import {init} from './actions';
 import {selectors as configSelectors} from './common/config';
 import {selectors as userSelectors} from './common/user';
+import {actions as roomsActions} from './common/rooms';
 
 import 'semantic-ui-css/semantic.css';
 import '../styles/main.scss';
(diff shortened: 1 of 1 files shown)


indico/modules/rb_new/client/js/setup.jsx
```diff
@@ -20,6 +21,17 @@ export default function setup(overrides = {}, postReducers = []) {
         const appContainer = document.getElementById('rb-app-container');
         const store = createRBStore(overrides, postReducers);
 
+        let oldPath = history.location.pathname;
+        history.listen(({pathname: newPath}) => {
```
pferreir: Couldn't this be rewritten as a reducer (on `LOCATION_CHANGE`)?

ThiefMaster: reducers are not supposed to dispatch new actions

pferreir: Ah right, I didn't see you have to fetch new stuff. If we saved it on the redux state when saving it remotely, then we could just use a reducer. But yeah, probably not worth the trouble.

PR #3668, opened 2018-11-22
Add room feature management & make components reusable

ThiefMaster:


indico/modules/rb_new/client/js/modules/admin/EditableList.jsx
```diff
@@ -0,0 +1,173 @@
...
+            this.setState({adding: false});
+        }
+        return rv;
+    };
+
+    render() {
+        const {
+            title, addModalTitle, renderAddForm, renderEditForm, renderDeleteMessage, renderItem, initialEditValues,
+            isFetching, items,
+            initialAddValues, onDelete, onUpdate,
```
mic4ael: Can't it appear on the previous line with `isFetching, items`?

indico/modules/rb_new/client/js/modules/admin/EditableListItem.jsx
```diff
@@ -0,0 +1,134 @@
...
+        onUpdate: PropTypes.func.isRequired,
+    };
+
+    static defaultProps = {
+        initialEditValues: item => item,
+    };
+
+    state = {
+        // delete
+        confirming: false,
```
mic4ael: Wouldn't it be better to have sth like `state = {delete: {state: ...}, edit: {state: ...}}`?

indico/modules/rb_new/client/js/modules/admin/EquipmentTypeList.jsx
```diff
@@ -0,0 +1,172 @@
...
+        }));
+
+        return (
+            <>
+                <Field name="name" component={ReduxFormField} as="input"
+                       format={formatters.trim} formatOnBlur
+                       label={Translate.string('Name')}
+                       disabled={fprops.submitting}
+                       autoFocus />
+                {!!featureOptions.length && (
```
mic4ael: maybe just use `>0` since you are using it somewhere down below `numEquipmentTypes > 0 && (
`

indico/modules/rb_new/controllers/backend/admin.py
```diff
@@ -47,12 +47,70 @@ def _process(self):
...
+    def _jsonify_one(self, equipment_type):
+        return jsonify(room_feature_schema.dump(equipment_type).data)
+
+    def _jsonify_many(self):
+        return jsonify(self._dump_features())
+
+    def _process_GET(self):
+        if self.feature:
+            return self._jsonify_one(self.feature)
+        else:
```
mic4ael: What about removing `else` and saving one line?

indico/modules/rb_new/client/js/modules/admin/EditableList.jsx
ThiefMaster: Initially I had only the actions on that line, but yes, can be merged now

indico/modules/rb_new/controllers/backend/admin.py
ThiefMaster: It's save 3 lines but I don't think it's more readable in this case

ThiefMaster: I usually do that if the if block is just error handling or some other "exit early" case, but not if it's really two cases that are both "normal".

indico/modules/rb_new/client/js/modules/admin/EditableListItem.jsx
ThiefMaster: react doesn't do deep merges for state changes so i'd always have to specify the whole object

PR #3669, opened 2018-11-23
Fix uwsgi package name

DrLove73:
On CentOS 7.5 I see only "uwsgi-plugin-python2" package, not "uwsgi-plugin-python"

### DIFF
--- a/docs/source/installation/production/centos/apache.rst
+++ b/docs/source/installation/production/centos/apache.rst
@@ -29,7 +29,7 @@ to the ``[base]`` and ``[updates]`` sections, as described in the
     yum install -y https://download.postgresql.org/pub/repos/yum/9.6/redhat/rhel-7-x86_64/pgdg-centos96-9.6-3.noarch.rpm
     yum install -y postgresql96 postgresql96-server postgresql96-libs postgresql96-devel postgresql96-contrib
     yum install -y httpd mod_proxy_uwsgi mod_ssl mod_xsendfile
-    yum install -y gcc redis uwsgi uwsgi-plugin-python
+    yum install -y gcc redis uwsgi uwsgi-plugin-python2
     yum install -y python-devel python-virtualenv libjpeg-turbo-devel libxslt-devel libxml2-devel libffi-devel pcre-devel libyaml-devel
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Thanks! Can you confirm that this was the only change needed for everything to work?

DrLove73:
I wish it is finished :) . Big "ATTENTION: DO NOT use this in production as it is. There are important things missing, such as HTTPS support." is a turn off, I need indico to be visible from outside.
If I can not progress in next few days, I will be installing in KVM VM using this howto.

DrLove73:
I am just installing indico in CentOS 7 VM. I am installing with Apache, but the problem is same for both howto's, and is located in chapter 7., while installing indico it self: "pip install indico"

> Successfully built Markdown MarkupSafe translitcodec zope.interface blinker itsdangerous termcolor terminaltables pytest-localserver alembic csscompressor pyScss Mako pycountry WTForms-Components pyatom Flask-BabelEx simplejson PyYAML html2text WTForms pojson webassets colorclass SQLAlchemy jsmin pyPdf python-editor pathlib validators intervals simplegeneric speaklater billiard functools32 infinity scandir pycparser
requests 2.18.4 has requirement idna<2.7,>=2.5, but you'll have idna 2.7 which is incompatible.

In Chapter 8. there is following warning:

>/opt/indico/.venv/lib/python2.7/site-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.""")

It should be solved with "pip install psycopg2-binary", so appropriate change should be edited in both howto's.

After installation, it seams to work correctly.

ThiefMaster:
You can ignore the warning. Instead of using `psycopg2-binary` it's better to install a non-binary psycopg2:

    pip uninstall -y psycopg2 && pip install --no-binary psycopg2 'psycopg2==2.7.4'

FWIW, this will fix itself on its own once psycopg2 2.8 is out and we switch to that.

PR #3670, opened 2018-11-23
Update apache.rst

DrLove73:
On CentOS 7.5 I see only "uwsgi-plugin-python2" package, not "uwsgi-plugin-python"


ThiefMaster:
I've amended your commit in #3669 to include this one as well.

PR #3674, opened 2018-11-26
Add template hook to vc page

mvidalgarcia:
Related to https://github.com/indico/indico-plugins-cern/pull/32


indico/modules/vc/templates/manage_event.html
```diff
@@ -7,6 +7,7 @@
     <div class="module-vc vc-management" id="vc-info-container">
         {% if plugins %}
             {% if event_vc_rooms %}
+                {{ template_hook('vc-assistance-message', event=event) }}
```
ThiefMaster: that name is too specific. something like `before-vc-list`?

PR #3675, opened 2018-11-26
Fix footer in login page

nop33:
Fixes issue reported in https://github.com/indico/indico/pull/3599#issuecomment-441657522

image


ThiefMaster:
Thanks!

PR #3679, opened 2018-11-29
Event roombooking link

### LINKED ISSUE #898: Room Booking: Closer link between room booking and events
``` yaml
# Ticket imported from Trac
```

We should add 'warnings'/'help' around, for instance:
1) when creating an event, if the room is chosen, we should say that they should book it too (from inside the event).
2) the same thing as 1, but from general settings.
3) if booking a room with webcast/recording facilities, we should say that they must create an event and request webcast/recording, etc..

and more to discuss.

meluru:
closes #898


indico/modules/rb_new/controllers/backend/rooms.py
```diff
@@ -110,3 +111,14 @@ def _process_PUT(self):
     def _process_DELETE(self):
         session.user.favorite_rooms.discard(self.room)
         return '', 204
+
+
+class RHCheckRoomAvailable(RHRoomBase):
+    @use_args({
```
ThiefMaster: use_kwargs since you use the args separately anyway

indico/modules/rb_new/controllers/backend/rooms.py
```diff
@@ -110,3 +111,14 @@ def _process_PUT(self):
...
+        'start_dt': fields.String(),
+        'end_dt': fields.String(),
+    })
+    def _process(self, args):
+        start_dt = dateutil.parser.parse(args['start_dt'])
```
ThiefMaster: can't we let webargs do the parsing?

indico/modules/rb_new/operations/bookings.py
```diff
@@ -206,3 +211,54 @@ def get_booking_occurrences(booking):
...
+                     if overlaps((start_dt.time(), end_dt.time()), (hours.start_time, hours.end_time))]
+    nonbookable_periods = get_rooms_nonbookable_periods([room], start_dt, end_dt)
+    blockings = get_rooms_blockings([room], start_dt, end_dt).get(room.id, [])
+    blocked_for_user = [blocking for blocking in blockings
+                        if not blocking.blocking.can_be_overridden(session.user, room)]
+    user_bookings = [booking for booking in bookings if booking.reservation.booked_for_id == session.user.id]
+    user_prebookings = [prebooking for prebooking in prebookings
+                        if prebooking.reservation.booked_for_id == session.user.id]
+
+    conflict_booking = True if bookings else False
```
ThiefMaster: `bool(bookings)`

indico/modules/events/templates/forms/event_creation_form.html
```diff
@@ -99,3 +99,215 @@
...
+            var startTime = $('#event-creation-start_dt-time').val();
+            var endDate = $('#event-creation-end_dt-date').val();
+            var endTime = $('#event-creation-end_dt-time').val();
+            var occurrences = $('#event-creation-occurrences').val();
+
+            roomData = JSON.parse($("#event-creation-location_data").val());
+            timezone = $('#event-creation-timezone').val();
+            category = JSON.parse($("#event-creation-category").val());
+
+            if (occurrences && occurrences.length === 1) {
```
mic4ael: Are you sure that is gonna work? `occurrences` here will contain a string not an array. So this basically will never match at this moment since you are checking the length of the stringified value of the field.

mic4ael:
Also I guess you saw it already but there is one test failing.

indico/modules/events/templates/forms/event_creation_form.html
```diff
@@ -99,3 +99,215 @@
...
+                setLectureTimes(JSON.parse(occurrences[0]));
+            } else {
+                startDt = moment(startDate + " " + startTime, 'DD/MM/YYYY HH:mm');
+                endDt = moment(endDate + " " + endTime, 'DD/MM/YYYY HH:mm');
+            }
+        }
+
+        initValues();
+
+        function isCategoryExcluded(categoryId) {
```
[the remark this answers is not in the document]
meluru: maybe `array.some` ?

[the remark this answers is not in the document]
ThiefMaster: once this goes through babel i'm sure we have a polyfill enabled for it. But yes, `.findIndex() !== -1`  should be replaced with `.some()`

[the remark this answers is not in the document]
ThiefMaster: or if you just need the IDs in that array just put only the IDs in the array so you can just use `.includes()`

indico/modules/events/client/js/creation.js
```diff
@@ -58,5 +76,209 @@
...
+                $createBooking.val(false);
+                return;
+            }
+
+            const roomId = roomData['room_id'];
+            previousRoomId = roomId;
+
+            const requestParams = {
+                room_id: roomId,
+                start_dt: startDt.format('YYYY-MM-DDTHH:mm:ss'),
```
ThiefMaster: i don't think we need to include seconds here, and instead of hardcoding the format you can use `moment.HTML5_FMT.DATETIME_LOCAL`

indico/modules/events/client/js/creation.js
```diff
@@ -58,5 +76,209 @@
...
+            };
+
+            $.ajax({
+                url: build_url(Indico.Urls.RoomBooking.room.check_available, requestParams),
+                method: 'GET',
+                dataType: 'json',
+                contentType: 'application/json',
+                error: handleAjaxError,
+                success(data) {
+                    if (data['user_booking']) {
```
ThiefMaster: `data.user_booking` (same in the other cases below)

ThiefMaster: `data = camelizeKeys(data);`?

indico/modules/events/client/js/creation.js
```diff
@@ -15,37 +15,55 @@
...
+        const $createBooking = $('#event-creation-create_booking');
+        const $availableMessage = $('#room-available');
+        const $availablePrebookMessage = $('#room-available-prebook');
+        const $conflictBookingMessage = $('#room-conflict-booking');
+        const $conflictPrebookingMessage = $('#room-conflict-prebooking');
+        const $conflictPrebookingPrebookMessage = $('#room-conflict-prebooking-prebook');
+        const $userBookingMessage = $('#room-user-booking');
+        const $userPrebookingMessage = $('#room-user-prebooking');
+        const $unbookableMessage = $('#room-unbookable');
+        const $cantBookMessage = $('#room-cant-book');
```
ThiefMaster: cannot

indico/modules/rb/forms/settings.py
```diff
@@ -32,6 +32,7 @@
 class SettingsForm(IndicoForm):
     admin_principals = PrincipalListField(_('Administrators'), groups=True)
     authorized_principals = PrincipalListField(_('Authorized users/groups'), groups=True)
+    excluded_categories = MultipleItemsField('Excluded categories', fields=[{'id': 'id', 'caption': 'Category ID'}])
```
ThiefMaster: needs a description, "Disable quick-book on event creation in these categories" or something similar

indico/modules/events/client/js/creation.js
```diff
@@ -58,5 +76,209 @@
...
+                timezone = ($(this).val());
+                updateAvailability();
+            });
+
+            $('#event-creation-category').on('change', function() {
+                category = JSON.parse($(this).val());
+                updateAvailability();
+            });
+
+            $('#create-booking').on('change', function() {
```
ThiefMaster: you repeat this code 3 times below, why not add a class for all these items and bind on that?

indico/modules/events/operations.py
```diff
@@ -97,6 +99,17 @@ def create_event(category, event_type, data, add_creator_as_manager=True, featur
     logger.info('Event %r created in %r by %r ', event, category, session.user)
     event.log(EventLogRealm.event, EventLogKind.positive, 'Event', 'Event created', session.user)
     db.session.flush()
+    if create_booking:
+        room_id = data['location_data'].pop('room_id', None)
+        if room_id:
+            booking = create_booking_for_event(room_id, event)
+            if booking:
+                logger.info('Booking %r created for event %r', booking, event)
+                log_data = {'Date': booking.start_dt.strftime('%d/%m/%Y'),
```
ThiefMaster: the log data should include the name of the room as well

indico/modules/rb_new/blueprint.py
```diff
@@ -48,6 +48,7 @@
 _bp.add_url_rule('/api/rooms/<int:room_id>/timeline', 'timeline', bookings.RHTimeline)
 _bp.add_url_rule('/api/rooms/<int:room_id>/stats', 'room_stats', rooms.RHRoomStats)
 _bp.add_url_rule('/api/suggestions', 'suggestions', bookings.RHRoomSuggestions)
+_bp.add_url_rule('/api/rooms/<int:room_id>/check_available', 'check_room_available', rooms.RHCheckRoomAvailable)
```
ThiefMaster: `check-available` or maybe `availability/simple` since it's similar to the 'availability' endpoint just without all the detailed data?

ThiefMaster: move it up one line to group it together with the other `/api/rooms/...` urls

indico/modules/rb_new/controllers/backend/rooms.py
```diff
@@ -115,3 +115,12 @@ def _process_PUT(self):
     def _process_DELETE(self):
         session.user.favorite_rooms.discard(self.room)
         return '', 204
+
+
+class RHCheckRoomAvailable(RHRoomBase):
+    @use_kwargs({
+        'start_dt': fields.DateTime(),
```
ThiefMaster: `NaiveDateTime()`?

indico/modules/rb_new/operations/bookings.py
```diff
@@ -209,3 +214,47 @@ def get_booking_occurrences(booking):
...
+        default_timezone = timezone(config.DEFAULT_TIMEZONE)
+        start_dt = event.start_dt.astimezone(default_timezone).replace(tzinfo=None)
+        end_dt = event.end_dt.astimezone(default_timezone).replace(tzinfo=None)
+        booking_reason = "Event '%s'" % (event.title)
+        data = dict(start_dt=start_dt, end_dt=end_dt, booked_for_user=event.creator, booking_reason=booking_reason,
+                    repeat_frequency=RepeatFrequency.NEVER, event_id=event.id)
+        resv = Reservation.create_from_data(room, data, session.user)
+        return resv
+    except NoReportError as e:
+        if 'Reservation has no valid occurrences' == e:
```
ThiefMaster: image

We don't do Yoda conditions :)

Also, that string is translated when the error is raised, so it won't work properly in other locales. Do we get any other NoReport errors there? If not let's just not check the reason at all.

indico/modules/rb_new/operations/bookings.py
```diff
@@ -209,3 +214,47 @@ def get_booking_occurrences(booking):
...
+            flash(_("Couldn't create the booking. Probably somebody else booked the room in the meantime."), 'error')
+        else:
+            raise e
```
ThiefMaster: always use a bare `raise` to re-raise an exception to avoid messing up the traceback

indico/web/templates/forms/location_widget.html
```diff
@@ -31,9 +31,8 @@
       the current venue. This is to avoid clearing the venue field by accident
       (see actions after selecting a room option).
 #}
-
```
ThiefMaster: i would keep this separation between comment and code

indico/modules/rb_new/operations/bookings.py
meluru: hahaha, Yoda conditions : D
I agree it's not perfect this way. There's another NoReport error if user is not authorized to book the room

ThiefMaster: But we check that before and don't even allow selection to book, right?

indico/modules/events/client/js/creation.js
meluru: it just seemed like a better idea, than adding a class parameter to switch macro

ThiefMaster: you can also use `#foo, #bar, #...` in a single `$(...)` call

meluru: thats a good idea

indico/modules/rb_new/operations/bookings.py
meluru: Yes, we do in the frontend, we also don't allow selection to book if room is occupied. But I figured that there's some chance somebody will book the room in the meantime and it would be good to show some error. But yeah, I guess it won't happen that somebody looses the rights to book the room in the meantime

ThiefMaster: Yaeh, enough of an edge case that a slightly incorrect error message is fine. People are used to almost-useless error messages from mobile apps etc anyway :/

indico/modules/rb/forms/settings.py
```diff
@@ -32,6 +32,8 @@
 class SettingsForm(IndicoForm):
     admin_principals = PrincipalListField(_('Administrators'), groups=True)
     authorized_principals = PrincipalListField(_('Authorized users/groups'), groups=True)
+    excluded_categories = MultipleItemsField(_('Excluded categories'), fields=[{'id': 'id', 'caption': 'Category ID'}],
+                                             description='Disable quick-book on event creation in these categories')
```
ThiefMaster: `_(...)`

indico/modules/rb_new/operations/bookings.py
```diff
@@ -209,3 +214,44 @@ def get_booking_occurrences(booking):
...
+    nonbookable_periods = get_rooms_nonbookable_periods([room], start_dt, end_dt)
+    blockings = get_rooms_blockings([room], start_dt, end_dt).get(room.id, [])
+    blocked_for_user = [blocking for blocking in blockings
+                        if not blocking.blocking.can_be_overridden(session.user, room)]
+    user_bookings = [booking for booking in bookings if booking.reservation.booked_for_id == session.user.id]
+    user_prebookings = [prebooking for prebooking in prebookings
+                        if prebooking.reservation.booked_for_id == session.user.id]
+
+    return {
+        'can_book': room.can_be_booked(session.user),
```
ThiefMaster: should we use `ignore_admin=False` here since there's no way to downgrade to prebooking?

indico/modules/rb_new/operations/bookings.py
```diff
@@ -209,3 +214,44 @@ def get_booking_occurrences(booking):
...
+        'can_prebook': room.can_be_prebooked(session.user),
```
ThiefMaster: same here for rooms that are not supposed to be bookable at all

indico/modules/rb_new/operations/bookings.py
```diff
@@ -209,3 +214,44 @@ def get_booking_occurrences(booking):
...
+    prebookings = [occ for occ in occurrences if not occ.reservation.is_accepted]
+    bookings = [occ for occ in occurrences if occ.reservation.is_accepted]
+    unbookable_hours = get_rooms_unbookable_hours([room])
+    hours_overlap = [hours for hours in unbookable_hours
+                     if overlaps((start_dt.time(), end_dt.time()), (hours.start_time, hours.end_time))]
+    nonbookable_periods = get_rooms_nonbookable_periods([room], start_dt, end_dt)
+    blockings = get_rooms_blockings([room], start_dt, end_dt).get(room.id, [])
+    blocked_for_user = [blocking for blocking in blockings
+                        if not blocking.blocking.can_be_overridden(session.user, room)]
+    user_bookings = [booking for booking in bookings if booking.reservation.booked_for_id == session.user.id]
```
ThiefMaster: what about using `any(...)` instead of `[...]` here instead of converting to `bool()` below? probably there are a few more cases where we could do that

indico/modules/rb_new/operations/bookings.py
```diff
@@ -209,3 +214,44 @@ def get_booking_occurrences(booking):
...
+    }
+
+
+def create_booking_for_event(room_id, event):
+    try:
+        room = Room.get_one(room_id)
+        default_timezone = timezone(config.DEFAULT_TIMEZONE)
+        start_dt = event.start_dt.astimezone(default_timezone).replace(tzinfo=None)
+        end_dt = event.end_dt.astimezone(default_timezone).replace(tzinfo=None)
+        booking_reason = "Event '%s'" % (event.title)
```
mvidalgarcia: i18n?

indico/modules/rb_new/operations/bookings.py
```diff
@@ -209,3 +214,44 @@ def get_booking_occurrences(booking):
...
+        data = dict(start_dt=start_dt, end_dt=end_dt, booked_for_user=event.creator, booking_reason=booking_reason,
+                    repeat_frequency=RepeatFrequency.NEVER, event_id=event.id)
+        resv = Reservation.create_from_data(room, data, session.user)
+        return resv
+    except NoReportError:
+        flash(_("Couldn't create the booking. Probably somebody else booked the room in the meantime."), 'error')
```
mvidalgarcia: "It wasn't possible to create.."? for me it sounds a bit too informal omitting the subject here.
Or maybe "Booking couldn't be created."? What do you think?

indico/modules/rb_new/operations/bookings.py
ThiefMaster: No, we don't translate log messages

mvidalgarcia: Isn't it the reason of the booking?

ThiefMaster: ah indeed. still, it's visible to everyone so IMO the same applies

ThiefMaster: or should we just use the event title without "Event '...'" around it?

mvidalgarcia: yeah, maybe it's the easiest solution

mvidalgarcia: Just a use case, if I have my Indico instance in Polish I'd find strange ending up with a partially English booking reason.

ThiefMaster: As a non-polish room owner I'd find it strange to have a polish justification why someone's using my room :)

We don't take the user's language into account anywhere else where content will become visible to others than the current user (e.g. emails)

meluru: I've used `Event '...'` to indicate it's booked for the event itself. I figured we will have `Contribution '...'` and `Session block '...'` for the same event as well.

indico/modules/rb_new/operations/bookings.py
meluru: yeah, sounds better :)

PR #3681, opened 2018-12-04
"Disable" days in the timeline view that are not part of the booking period

### LINKED ISSUE #3667: ElasticTimeline: add data for days in-between
Example:

image

We should have the booking data for all displayed days. It could somehow be shown in some kind of paler color or with some transparency.

mic4ael:
closes #3667

### DIFF
--- a/indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
+++ b/indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
@@ -145,38 +145,39 @@ export default class DailyTimelineContent extends React.Component {
         const labelWidth = longLabel ? 200 : 150;
 
         return (
-            <div styleName="timeline-content"
-                 className={!selectable && 'timeline-non-selectable'}
-                 style={{width}}>
-                <div style={{
-                    left: labelWidth,
-                    width: `calc(100% - ${labelWidth}px)`,
-                }}
-                     styleName="timeline-lines">
-                    {this.renderDividers(hourSpan, hourStep)}
+            <>
+                <div styleName="timeline-content"
+                     className={!selectable ? 'timeline-non-selectable' : ''}
+                     style={{width}}>
+                    <div style={{left: labelWidth, width: `calc(100% - ${labelWidth}px)`}}
+                         styleName="timeline-lines">
+                        {this.renderDividers(hourSpan, hourStep)}
+                    </div>
+                    <div>
+                        <List width={width}
+                              height={height}
+                              rowCount={rows.length}
+                              overscanRowCount={15}
+                              rowHeight={50}
+                              rowRenderer={({index, style, key}) => (
+                                  this.renderTimelineRow(rows[index], key, style)
+                              )}
+                              {...extraProps}
+                              tabIndex={null} />
+                    </div>
                 </div>
-                <div>
-                    <List width={width}
-                          height={height}
-                          rowCount={rows.length}
-                          overscanRowCount={15}
-                          rowHeight={50}
-                          rowRenderer={({index, style, key}) => (
-                              this.renderTimelineRow(rows[index], key, style)
-                          )}
-                          {...extraProps}
-                          tabIndex={null} />
-                    {isLoading && (
-                        _.range(0, 10).map((i) => (
-                            <Placeholder key={i} styleName="timeline-item-placeholder" fluid>
+                {isLoading && (
+                    <div style={{marginTop: 10}}>
+                        {_.range(0, 10).map((i) => (
+                            <Placeholder key={i} styleName="timeline-item-placeholder" style={{width}} fluid>
                                 <Placeholder.Paragraph>
                                     <Placeholder.Line />
                                 </Placeholder.Paragraph>
                             </Placeholder>
-                        ))
-                    )}
-                </div>
-            </div>
+                        ))}
+                    </div>
+                )}
+            </>
         );
     }
 
--- a/indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
+++ b/indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
@@ -122,7 +122,7 @@ export default class DailyTimelineContent extends React.Component {
         const labelWidth = longLabel ? 200 : 150;
 
         return (
-            <div styleName="timeline-header" className={!selectable && 'timeline-non-selectable'}>
+            <div styleName="timeline-header" className={!selectable ? 'timeline-non-selectable' : ''}>
                 <div style={{width: labelWidth}} />
                 <div styleName="timeline-header-labels">
                     {_.range(0, hourSpan, hourStep).map((i, n) => (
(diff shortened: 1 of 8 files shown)


indico/modules/rb_new/client/js/common/timeline/MonthlyTimelineContent.jsx
```diff
@@ -75,10 +75,28 @@ export default class MonthlyTimelineContent extends WeeklyTimelineContent {
...
+            return this.dates.findIndex((el) => el === day) + 1;
+        });
+
+        const additionalStyles = (dayNumber) => {
+            if (emptyDays.includes(dayNumber)) {
+                return {
+                    backgroundColor: '#FBE1E1',
+                    opacity: 0.35,
+                };
+            } else {
```
ThiefMaster: I'd negate the if and return this empty object early

indico/modules/rb_new/client/js/common/timeline/MonthlyTimelineContent.jsx
```diff
@@ -75,10 +75,28 @@ export default class MonthlyTimelineContent extends WeeklyTimelineContent {
         const nDays = this.dates.length;
         const daySize = (100 / nDays);
 
+        const {dateRange} = this.props;
+        const emptyDays = dateRange.length === 0 ? [] : this.dates.filter((day) => {
```
ThiefMaster: I think the ternary here makes it much less readable. I'd rather assign an empty list and then, if there are dates, assign the generated value

or is this.dates an empty list in case there's no dateRange? in that case no extra code would be needed

indico/modules/rb_new/client/js/common/timeline/MonthlyTimelineContent.jsx
```diff
@@ -75,10 +75,28 @@ export default class MonthlyTimelineContent extends WeeklyTimelineContent {
...
+            return this.dates.findIndex((el) => el === day) + 1;
+        });
+
+        const additionalStyles = (dayNumber) => {
+            if (!emptyDays.includes(dayNumber)) {
+                return {};
+            }
+
+            return {
+                backgroundColor: '#FBE1E1',
```
mvidalgarcia: Could it be possible to keep this somewhere else in a variable? If someday we decide to change the palette it will be hard to change all these values. I wonder if it'd be possible to export sass variables to js... Ok, I found [this](https://www.bluematador.com/blog/how-to-share-variables-between-js-and-sass), it might be interesting to apply to the project, what do you think? cc: @pferreir @ThiefMaster @meluru @panagiotappl

mic4ael:
Yeah, I saw the placeholders overlapping with the timetable and completely forgot about it.

mic4ael:
Besides, I think we also agreed at some point that loading this additional data is not necessary at all thus it suffices to "disable" the days that are not part of the booking period. It would be quite heavy if we were to load all the occurrences happening during a specific month.

mvidalgarcia:
But we do that in the calendar anyway, don't we?

mic4ael:
I guess, there are less filters which make queries not that heavy.

mic4ael:
So that is how it looks like right now

screenshot from 2018-12-05 15-20-15

indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
```diff
@@ -145,38 +145,39 @@ export default class DailyTimelineContent extends React.Component {
...
-                 style={{width}}>
-                <div style={{
-                    left: labelWidth,
-                    width: `calc(100% - ${labelWidth}px)`,
-                }}
-                     styleName="timeline-lines">
-                    {this.renderDividers(hourSpan, hourStep)}
+            <>
+                <div styleName="timeline-content"
+                     className={!selectable && 'timeline-non-selectable'}
```
pferreir: Won't this end up in `className=false`? IIRC React doesn't like those.

indico/modules/rb_new/client/js/common/timeline/WeeklyTimelineContent.jsx
```diff
@@ -26,6 +28,19 @@ import style from './WeeklyTimelineContent.module.scss';
...
 
 export default class WeeklyTimelineContent extends DailyTimelineContent {
+    static propTypes = {
+        ...DailyTimelineContent.propTypes,
+        dateRange: PropTypes.array.isRequired,
+    };
+
+    get dates() {
+        const {rows} = this.props;
+        if (rows.length) {
```
pferreir: Just code golfing, but what about this?
```js
const [row] = rows;
return row ? row.availability.map(([dt]) => dt) : [];
```

indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
mic4ael: Hm, I can change it. I didn't look at it since I didn't modify the code. It has just been moved.

indico/modules/rb_new/client/js/common/timeline/TimelineItem.jsx
```diff
@@ -44,12 +44,14 @@ const types = {
...
-function getKeyForOccurrence({reservation, startDt, endDt}) {
+function getKeyForOccurrence(name, {reservation, startTime, endTime, startDt, endDt}) {
     let key = '';
     if (reservation) {
         key += `${reservation.id}-`;
     }
-    return `${name}-${key}${startDt}-${endDt}`;
+    // startDt ? startDt : startTime: this handles unbookableHours which
+    // don't have start/end dates only start/end times
+    return `${name}-${key}${startDt ? startDt : startTime}-${endDt ? endDt : endTime}`;
```
ThiefMaster: you can simpllify this to `${startDt || startTime}` and `${endDt || endTime}`

indico/modules/rb_new/client/js/common/timeline/WeeklyTimelineContent.jsx
```diff
@@ -71,27 +85,34 @@ export default class WeeklyTimelineContent extends DailyTimelineContent {
             return null;
         }
 
+        const {dateRange} = this.props;
+        const emptyDays = this.dates.filter((day) => {
+            return dateRange.length !== 0 && !dateRange.includes(day);
+        }).map((day) => {
```
ThiefMaster:
```js
.map(day => this.dates.findIndex((el) => el === day))
```

indico/modules/rb_new/client/js/common/timeline/WeeklyTimelineContent.jsx
```diff
@@ -71,27 +85,34 @@ export default class WeeklyTimelineContent extends DailyTimelineContent {
...
+        });
+
         return (
             _.times(7, n => (
                 <div styleName="style.timeline-day-divider"
+                     className={toClasses({hidden: emptyDays.includes(n), visible: !emptyDays.includes(n)})}
                      style={{left: `${n * daySize}%`, width: `${daySize}%`}}
                      key={`day-divider-${n}`}>
-                    {super.renderDividers(hourSpan, hourStep)}
+                    {!emptyDays.includes(n) && super.renderDividers(hourSpan, hourStep)}
```
ThiefMaster: What about `emptyDays.includes(n) || ...`?

PR #3689, opened 2018-12-06
Filter calendar by user bookings

### LINKED ISSUE #3683: "My Bookings"
This should normally link to the Calendar, with a filter set for the user's bookings (booked by/for him/her).

mic4ael:
closes #3683

### DIFF
--- a/indico/modules/rb_new/client/js/common/roomSearch/reducers.js
+++ b/indico/modules/rb_new/client/js/common/roomSearch/reducers.js
@@ -65,6 +65,10 @@ export function initialRoomFilterStateFactory(namespace) {
                 endTime: null
             },
         });
+    } else if (namespace === 'calendar') {
+        Object.assign(state, {
+            myBookings: false,
+        });
     }
 
     return state;
--- a/indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
+++ b/indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
@@ -27,6 +27,7 @@ import searchBarFactory from '../../components/SearchBar';
 import * as calendarActions from './actions';
 import * as calendarSelectors from './selectors';
 import {actions as bookingsActions} from '../../common/bookings';
+import {actions as filtersActions} from '../../common/filters';
 import {selectors as roomsSelectors, actions as roomsActions} from '../../common/rooms';
 import {EditableTimelineItem, ElasticTimeline, TimelineHeader, TimelineItem} from '../../common/timeline';
 import {actions as bookRoomActions} from '../bookRoom';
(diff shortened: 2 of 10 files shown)


indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
```diff
@@ -115,15 +117,27 @@ class Calendar extends React.Component {
...
+                                        onClick={this.toggleShowUnused} />}>
+                    {showUnused
+                        ? <Translate>Hide unused rooms</Translate>
+                        : <Translate>Show unused rooms</Translate>}
+                </Popup>
+                <Popup trigger={<Button size="large"
+                                        primary={myBookings}
+                                        icon="user circle"
+                                        onClick={() => setFilterParameter('myBookings', !myBookings || null)} />}>
+                    <Translate>
```
mvidalgarcia: Can nicely fit in one line

indico/modules/rb_new/client/js/modules/calendar/queryString.js
```diff
@@ -37,7 +37,12 @@ const rules = {
     mode: {
         validator: v.isIn(['days', 'weeks', 'months']),
         stateField: defaultStateField('datePicker.mode', 'days'),
-    }
+    },
+    myBookings: {
```
mvidalgarcia: I've noticed that we use camel case for some query params (`myBlockings`, `myRooms`) and snake case for some others (map lat & lng). I don't know which case is more common but I think we should be consistent.

indico/modules/rb_new/client/js/modules/calendar/Calendar.jsx
mic4ael: Yeah, but this is really subjective ;P I like it much more like this.

indico/modules/rb_new/client/js/modules/calendar/queryString.js
mic4ael: Yeah, true. Maybe it is a good opportunity to change it in blockings as well

indico/modules/rb_new/client/js/common/roomSearch/reducers.js
```diff
@@ -67,7 +67,7 @@ export function initialRoomFilterStateFactory(namespace) {
         });
     }
 
-    return state;
+    return Object.assign(state, customFilters);
```
ThiefMaster: we already conditionally add default state above based on the namespace. what about doing the same for calendar as well instead of passing the customFilters?

indico/modules/rb_new/client/js/modules/calendar/actions.js
```diff
@@ -71,7 +71,7 @@ export function fetchCalendar(fetchRooms = true) {
             }
         }
 
-        const params = preProcessParameters(datePicker, ajaxRules);
+        const params = preProcessParameters({...datePicker, myBookings: filters.myBookings}, ajaxRules);
```
ThiefMaster: `myBookings` does not affect the room list so this should not be part of `filters` here (as we use it above to check if the room list must be reloaded as well). You could go for this and then use `roomFilters` instead of filters in the code above:

```js
const {
    calendar: {
        filters: {myBookings, ...roomFilters},
        data: {roomIds},
        datePicker,
    }
} = getState();
```

PR #3692, opened 2018-12-10
Separate booking and pre-booking actions

### LINKED ISSUE #3684: Add "pre-booking" button to room renderer
Managers should somehow see both (dropdown menu?)

mvidalgarcia:
closes #3684

### DIFF
--- a/indico/modules/rb_new/client/js/common/rooms/RoomDetailsModal.jsx
+++ b/indico/modules/rb_new/client/js/common/rooms/RoomDetailsModal.jsx
@@ -28,6 +28,7 @@ import RoomKeyLocation from '../../components/RoomKeyLocation';
 import RoomStats from './RoomStats';
 import {DailyTimelineContent, TimelineLegend} from '../timeline';
 import * as roomsSelectors from './selectors';
+import {selectors as userSelectors} from '../user';
 import {actions as bookRoomActions} from '../../modules/bookRoom';
 
 import './RoomDetailsModal.module.scss';
@@ -40,16 +41,18 @@ class RoomDetailsModal extends React.Component {
         attributes: PropTypes.array.isRequired,
         onClose: PropTypes.func.isRequired,
         promptDatesOnBook: PropTypes.bool,
+        title: IndicoPropTypes.i18n,
+        isAdmin: PropTypes.bool,
         actions: PropTypes.exact({
             openBookRoom: PropTypes.func.isRequired,
             openBookingForm: PropTypes.func.isRequired,
         }).isRequired,
-        title: IndicoPropTypes.i18n
     };
 
     static defaultProps = {
         promptDatesOnBook: false,
-        title: <Translate>Room Details</Translate>
+        title: <Translate>Room Details</Translate>,
+        isAdmin: false,
     };
 
     handleCloseModal = () => {
@@ -59,8 +62,8 @@ class RoomDetailsModal extends React.Component {
 
     render() {
         const {
-            room, availability, attributes, promptDatesOnBook,
-            actions: {openBookRoom, openBookingForm}, title
+            room, availability, attributes, promptDatesOnBook, title, isAdmin,
+            actions: {openBookRoom, openBookingForm}
         } = this.props;
         return (
             <Modal open onClose={this.handleCloseModal} size="large" closeIcon>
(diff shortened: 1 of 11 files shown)


indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
```diff
@@ -275,8 +261,16 @@ class BookRoom extends React.Component {
                                     <RoomRenderer rooms={this.visibleRooms}>
                                         {room => (
                                             <Slot name="actions">
-                                                <Popup trigger={bookingModalBtn(room)}
-                                                       content={labels.bookButton}
+                                                {isAdmin && !room.isAutoConfirm && (
```
mvidalgarcia: How can I make the selector work from this page if I don't have `roomId` (obviously) available in the props of the component (`mapStateToProps`)?

ThiefMaster: Damn that's a good question.. my first thought was to have RoomRenderer use the selector but it also doesn't have a single room... I think the problem here is that there's tons of indirection going on with RoomRenderer and the Slots.

The easiest option I see is changing the `getRoom` selector to take care of adding this information...

```js
const _getRoom = (state, {roomId}) => getAllRooms(state)[roomId];
export const getRoom = createSelector(
    _getRoom,
    isUserAdmin,
    (room, isAdmin) => {
        return {
            ...room,
            canUserBook: isUserAdmin || room.isAutoConfirm,
            canUserPrebook: !room.isAutoConfirm,
        };
    }
);
```

and later once we fetch the user's privileges from the server we'd then change the selector to something like this:

```js
const _getRoom = (state, {roomId}) => getAllRooms(state)[roomId];
export const getRoom = createSelector(
    _getRoom,
    getUserBookingPrivs,
    (room, userRoomPrivs) => {
        return {
            ...room,
            canUserBook: userRoomPrivs.canBook,
            canUserPrebook: userRoomPrivs.canPrebook,
        };
    }
);
```

mvidalgarcia: I get the idea, but I still don't get how `getRoom` can be used in this file as we don't have a single room in the props and all the checks are done here.

ThiefMaster: It doesn't - `room` will simply contain it, since that data originally comes from `getRoom`

mvidalgarcia: As far as I can see, the `results` in BookRoom come from `getSearchResults` selector that uses `getAllRooms` but not `getRoom`.

ThiefMaster: Indeed. Same solution though - just implement it in getAllRooms then ;)

indico/modules/rb_new/client/js/common/rooms/RoomDetailsModal.jsx
```diff
@@ -142,10 +142,18 @@ function RoomDetails({bookRoom, room, availability, attributes}) {
...
-                        </Button>
+                        {room.canUserBook && (
+                            <Button color="green" onClick={() => bookRoom(room.id, {isPrebooking: false})}>
+                                <Icon name="check circle" />
+                                <Translate>Start booking</Translate>
+                            </Button>
+                        )}
+                        {room.canUserPreBook && (
+                            <Button color="orange" onClick={() => bookRoom(room.id, {isPrebooking: true})}>
+                                <Icon name="check circle" />
```
pferreir: I believe we said we'd look for another icon?

pferreir: Anyway, it can be done after...

mvidalgarcia: That's true, I had forgotten. Any suggestions? We use the clock in some places to point out that a booking is actually a pre-booking but this makes sense as the pre-booking has been already created. I don't know if it'd be suitable in this case.

meluru: In my opinion a clock is suitable also for creation. Since we use it in other places already, the users may recognize this icon as anything connected to prebooking

mvidalgarcia:
Updated.

PR #3695, opened 2018-12-11
Landing page - "My bookings" + Carousel

### LINKED ISSUE #3687: Add "my bookings" preview in landing page
The initial mockups included this:
image

... and then the statistics in case there was nothing else to show. We could think of a carousel, since I believe the statistics are useful information too (or just "squeeze" the statistics?).

pferreir:
kapture 2018-12-11 at 17 46 57

It should be nice and responsive now (to a certain extent).
Closes #3687

### DIFF
--- a/indico/modules/rb_new/client/js/modules/landing/Landing.jsx
+++ b/indico/modules/rb_new/client/js/modules/landing/Landing.jsx
@@ -20,12 +20,16 @@ import {connect} from 'react-redux';
 import PropTypes from 'prop-types';
 import {Card, Checkbox, Form, Grid} from 'semantic-ui-react';
 import {Translate} from 'indico/react/i18n';
+import {Carousel} from 'indico/react/components';
 import {Overridable} from 'indico/react/util';
 
 import {actions as filtersActions} from '../../common/filters';
 import BookingBootstrapForm from '../../components/BookingBootstrapForm';
 import LandingStatistics from './LandingStatistics';
+import UpcomingBookings from './UpcomingBookings';
 import {selectors as userSelectors} from '../../common/user';
+import * as landingActions from './actions';
+import * as landingSelectors from './selectors';
 
 import './Landing.module.scss';
 
@@ -34,8 +38,18 @@ class Landing extends React.Component {
     static propTypes = {
         actions: PropTypes.exact({
             setFilters: PropTypes.func.isRequired,
+            fetchUpcomingBookings: PropTypes.func.isRequired
         }).isRequired,
-        userHasFavorites: PropTypes.bool.isRequired
+        userHasFavorites: PropTypes.bool.isRequired,
+        showUpcomingBookings: PropTypes.bool,
+        fetchedUpcomingBookings: PropTypes.bool.isRequired,
+        hasUpcomingBookings: PropTypes.bool.isRequired,
+        upcomingBookings: PropTypes.arrayOf(PropTypes.object)
+    };
+
+    static defaultProps = {
+        upcomingBookings: null,
+        showUpcomingBookings: true
     };
 
     state = {
@@ -44,6 +58,13 @@ class Landing extends React.Component {
             onlyFavorites: false
         }
     };
+
+    componentDidMount() {
+        const {actions: {fetchUpcomingBookings}, showUpcomingBookings} = this.props;
+        if (showUpcomingBookings) {
+            fetchUpcomingBookings();
+        }
+    }
 
     doSearch = (formState) => {
         const {extraState, text} = this.state;
@@ -70,8 +91,19 @@ class Landing extends React.Component {
         this.setState({extraState: {...extraState, ...attrs}});
     };
 
+    renderCarousel() {
+        const {upcomingBookings} = this.props;
+        const panes = [
+            {key: 'upcoming', content: <UpcomingBookings bookings={upcomingBookings} />, delay: 20},
+            {key: 'stats', content: <LandingStatistics />}
+        ];
+        return (
+            <Carousel panes={panes} />
+        );
+    }
+
     render() {
-        const {userHasFavorites} = this.props;
+        const {userHasFavorites, showUpcomingBookings, hasUpcomingBookings, fetchedUpcomingBookings} = this.props;
         const {extraState} = this.state;
         return (
             <div className="landing-wrapper">
@@ -103,11 +135,17 @@ class Landing extends React.Component {
                             </Card.Content>
                         </Card>
                     </Grid.Row>
-                    <Grid.Row styleName="landing-page-statistics">
-                        <div styleName="statistics">
-                            <LandingStatistics />
-                        </div>
-                    </Grid.Row>
+                    {(!showUpcomingBookings || fetchedUpcomingBookings) && (
+                        <Grid.Row styleName="landing-page-lower-row">
+                            <div styleName="lower-row">
+                                {hasUpcomingBookings ? (
+                                    this.renderCarousel()
+                                ) : (
+                                    <LandingStatistics />
+                                )}
+                            </div>
+                        </Grid.Row>
+                    )}
                 </Grid>
             </div>
         );
@@ -117,13 +155,19 @@ class Landing extends React.Component {
 
 export default connect(
     state => ({
-        userHasFavorites: userSelectors.hasFavoriteRooms(state)
+        userHasFavorites: userSelectors.hasFavoriteRooms(state),
+        fetchedUpcomingBookings: landingSelectors.hasFetchedUpcomingBookings(state),
+        hasUpcomingBookings: landingSelectors.hasUpcomingBookings(state),
+        upcomingBookings: landingSelectors.getUpcomingBookings(state)
     }),
     dispatch => ({
         actions: {
             setFilters(data) {
                 dispatch(filtersActions.setFilters('bookRoom', data));
+            },
+            fetchUpcomingBookings() {
+                dispatch(landingActions.fetchUpcomingBookings());
             }
         }
     })
-)(Landing);
+)(Overridable.component('Landing', Landing));
--- a/indico/modules/rb_new/client/js/modules/landing/UpcomingBookings.jsx
+++ b/indico/modules/rb_new/client/js/modules/landing/UpcomingBookings.jsx
@@ -1,0 +1,105 @@
+/* This file is part of Indico.
+ * Copyright (C) 2002 - 2018 European Organization for Nuclear Research (CERN).
+ *
+ * Indico is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Indico is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import PropTypes from 'prop-types';
+import React from 'react';
+import {bindActionCreators} from 'redux';
+import {Header, List, Popup, Label} from 'semantic-ui-react';
+import {connect} from 'react-redux';
+import {toMoment} from 'indico/utils/date';
+import SpriteImage from '../../components/SpriteImage';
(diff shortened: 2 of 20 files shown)


indico/modules/rb_new/client/js/modules/landing/Landing.jsx
```diff
@@ -20,12 +20,16 @@ import {connect} from 'react-redux';
 import PropTypes from 'prop-types';
 import {Card, Checkbox, Form, Grid} from 'semantic-ui-react';
 import {Translate} from 'indico/react/i18n';
+import Carousel from 'indico/react/components/Carousel';
```
ThiefMaster: export via indico/react/components/index.js?

even though i'm starting to feel like there wouldn't be any real disadvantage in not having that index.js and then importing it like you did

indico/modules/rb_new/client/js/modules/landing/Landing.jsx
```diff
@@ -34,8 +38,16 @@ class Landing extends React.Component {
...
         }).isRequired,
-        userHasFavorites: PropTypes.bool.isRequired
+        userHasFavorites: PropTypes.bool.isRequired,
+        fetchedUpcomingBookings: PropTypes.bool.isRequired,
+        hasUpcomingBookings: PropTypes.bool.isRequired,
+        upcomingBookings: PropTypes.arrayOf(PropTypes.object)
+    };
+
+    static defaultProps = {
+        upcomingBookings: null
```
ThiefMaster: does this ever happen? i guess the default in the redux state is an empty list

indico/modules/rb_new/client/js/modules/landing/UpcomingBookings.jsx
```diff
@@ -0,0 +1,105 @@
...
+import SpriteImage from '../../components/SpriteImage';
+import {actions as bookingActions} from '../../common/bookings';
+
+import './UpcomingBookings.module.scss';
+
+
+/**
+ * A widget that shows the user a list of their upcoming bookings
+ */
+class UpcomingBookings extends React.Component {
```
ThiefMaster: this component can probably be pure

indico/modules/rb_new/client/js/modules/landing/UpcomingBookings.jsx
```diff
@@ -0,0 +1,105 @@
...
+    static propTypes = {
+        /** The list of booking objects */
+        bookings: PropTypes.arrayOf(PropTypes.object).isRequired,
+        actions: PropTypes.shape({
```
ThiefMaster: `.exact()` instead of `.shape()` so it shows a proptypes error if we ever add another action without adding it here

indico/modules/rb_new/client/js/modules/landing/UpcomingBookings.jsx
```diff
@@ -0,0 +1,105 @@
...
+        return (
+            <Label corner="right"
+                   icon="time"
+                   color="yellow"
+                   size="small" />
+        );
+    }
+
+    renderCard({reservation: {room, id: resvId, bookingReason, isAccepted}, startDt}) {
+        const mStartDt = toMoment(startDt);
```
ThiefMaster: DT instead of Dt? I think we use DT everywhere else

indico/modules/rb_new/client/js/modules/landing/reducers.js
```diff
@@ -38,4 +38,19 @@ export default combineReducers({
             }
         }
     }),
+    upcomingBookings: combineReducers({
+        request: requestReducer(
+            landingActions.FETCH_BOOKINGS_REQUEST,
+            landingActions.FETCH_BOOKINGS_SUCCESS,
+            landingActions.FETCH_BOOKINGS_ERROR
+        ),
+        data: (state = null, action) => {
```
ThiefMaster: default to an empty list/object? (see my other comment on the component proptype)

indico/modules/rb_new/client/js/modules/landing/Landing.jsx
```diff
@@ -71,8 +88,12 @@ class Landing extends React.Component {
     };
 
     render() {
-        const {userHasFavorites} = this.props;
+        const {userHasFavorites, hasUpcomingBookings, upcomingBookings, fetchedUpcomingBookings} = this.props;
         const {extraState} = this.state;
+        const panes = () => [
```
ThiefMaster: why a function?

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -218,3 +221,15 @@ def _process(self):
         if not self.booking.event.can_access(session.user):
             return jsonify(can_access=False)
         return jsonify(reservation_event_data_schema.dump(self.booking.event).data)
+
+
+class RHMyUpcomingBookings(RHRoomBookingBase):
+    def _process(self):
+        q = ReservationOccurrence.query.filter(
+            ReservationOccurrence.start_dt > now_utc(),
```
ThiefMaster: aren't those datetimes naive? ==> now instead of now_utc?

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -218,3 +221,15 @@ def _process(self):
...
+            db.or_(
+                Reservation.booked_for_user == session.user,
+                Reservation.created_by_user == session.user,
+            )
+        ).join(Reservation).order_by(ReservationOccurrence.start_dt.asc()).limit(5)
```
ThiefMaster: what about using the formatting we use for all other queries?

```python
q = (ReservationOccurrence.query
     .filter(ReservationOccurrence.start_dt > now_utc(),
             db.or_(Reservation.booked_for_user == session.user,
                    Reservation.created_by_user == session.user))
     .join(Reservation)
     .order_by(ReservationOccurrence.start_dt.asc())
     .limit(5)

```

indico/modules/rb_new/schemas.py
```diff
@@ -43,6 +38,11 @@
 from indico.util.marshmallow import NaiveDateTime
 from indico.util.string import natural_sort_key
 
+from flask import session
+from marshmallow import Schema, ValidationError, fields, post_dump, validate, validates_schema
```
ThiefMaster: this doesn't belong down here (something broken in your isort?)

indico/modules/rb_new/schemas.py
```diff
@@ -266,8 +284,9 @@ class Meta:
 rooms_schema = RoomSchema(many=True)
 room_attribute_values_schema = RoomAttributeValuesSchema(many=True)
 map_areas_schema = MapAreaSchema(many=True)
-reservation_occurrences_schema = ReservationOccurrenceSchema(many=True)
-reservation_schema = ReservationSchema()
+reservation_occurrences_min_schema = ReservationOccurrenceMinSchema(many=True)
+reservation_occurrences_full_schema = ReservationOccurrenceFullSchema(many=True)
```
ThiefMaster: just thinking, should we start to get rid of `many=True` when instantiating schemas and instead pass it when serializing? right now we have an unholy mix of both and then sometimes setting/unsetting it at serialization time

indico/modules/rb_new/util.py
```diff
@@ -20,16 +20,17 @@
...
 
 from indico.legacy.common.cache import GenericCache
 from indico.modules.rb.models.rooms import Room
 from indico.modules.rb_new.schemas import (bookable_hours_schema, nonbookable_periods_schema,
-                                           reservation_occurrences_schema, simple_blockings_schema)
+                                           reservation_occurrences_min_schema, simple_blockings_schema)
 from indico.util.string import crc32
 from indico.util.struct.iterables import group_list
 
+from flask import current_app
```
ThiefMaster: isort striking again

indico/web/client/js/react/components/Carousel.jsx
```diff
@@ -0,0 +1,193 @@
...
+    };
+
+    componentDidMount() {
+        const {delay: globalDelay, panes: [{delay: paneDelay}]} = this.props;
+        this.timeout = setTimeout(this.switchPane, (paneDelay || globalDelay) * 1000);
+    }
+
+    componentWillUnmount() {
+        // cancel timeout for next transition
+        window.clearTimeout(this.timeout);
```
ThiefMaster: remove `window.` here or add it to `setTimeout` as well (i'd remove it)

indico/web/client/js/react/components/Carousel.jsx
```diff
@@ -0,0 +1,193 @@
...
+            activePane,
+        });
+
+        // Schedule next switch
+        this.timeout = setTimeout(this.switchPane, (paneDelay || globalDelay) * 1000);
+    };
+
+    forcePaneSwitch = (newPane) => {
+        // cancel automatic switching (forever!)
+        window.clearTimeout(this.timeout);
```
ThiefMaster: on unmount we'll call it again. set to null and conditionally call it there?

indico/modules/rb_new/client/js/modules/landing/Landing.jsx
pferreir: I think that importing from `components` is cleaner, especially if you have to import several classes from the same module.

indico/modules/rb_new/client/js/modules/landing/Landing.jsx
pferreir: Indeed.

pferreir: `data` is `null` by default. `state = null` as default parameter in the reducer.

indico/modules/rb_new/client/js/modules/landing/UpcomingBookings.jsx
pferreir: Indeed!

indico/modules/rb_new/client/js/modules/landing/reducers.js
pferreir: That's already the default in the `stats` key. And TBH, it's much easier to check for `!state` than for `!state.length`. So, I'd tend to prefer leaving it like this.

indico/modules/rb_new/client/js/modules/landing/Landing.jsx
pferreir: It's a trick to avoid rendering it unconditionally. Since I'm then rendering it within a ternary expression, it's easier than extracting the whole thing into its own function.

indico/modules/rb_new/client/js/modules/landing/UpcomingBookings.jsx
pferreir: The thing is that the `startDt` in the parameters already comes like this from `camelizeKeys`. I was just keeping the capitalization for consistency. But yeah, I can just change the derived variable to `mStartDT`.

indico/modules/rb_new/client/js/modules/landing/Landing.jsx
ThiefMaster: Thought so, but I don't think just running the JSX snippets renders - so the ternary later should already avoid rendering it

indico/web/client/js/react/components/Carousel.jsx
pferreir: Yeah, that's an option. Even if calling it with `null` as a parameter doesn't break anything.

indico/modules/rb_new/schemas.py
pferreir: Yep. I noticed that too.

indico/modules/rb_new/client/js/modules/landing/Landing.jsx
pferreir: OK, I moved it all to a separate function, which makes things clearer.

indico/modules/rb_new/controllers/backend/bookings.py
pferreir: `)` missing 😉 
Yeah, you're right!

indico/modules/rb_new/controllers/backend/bookings.py
pferreir: Good point. I guess it should rather be `pytz.timezone(config.DEFAULT_TIMEZONE).localize(now_utc())` in that case.

ThiefMaster: We have a `utc_to_server()` util that does it

pferreir: Ah cool.

pferreir:
OK, suggestions applied. I've also fixed the bug with the page in the background showing up as placeholders (+ error 422).

PR #3701, opened 2019-01-07
Change state of accepted Prebookings when modified

### LINKED ISSUE #3387: Room Booking: Subject modifications of confirmed pre-bookings to acceptance
For a room with pre-booking enabled, when a booking is accepted, the user that booked it can change it to a different date and the booking gets automatically updated.
Since pre-booking is enabled, modifications of these bookings should be subject to acceptance again.

Related incident `INC1693961`

panagiotappl:
closes #3387

### DIFF
--- a/indico/modules/rb/models/reservations.py
+++ b/indico/modules/rb/models/reservations.py
@@ -35,8 +35,8 @@ from indico.modules.rb.models.reservation_edit_logs import ReservationEditLog
 from indico.modules.rb.models.reservation_occurrences import ReservationOccurrence, ReservationOccurrenceState
 from indico.modules.rb.models.room_nonbookable_periods import NonBookablePeriod
 from indico.modules.rb.models.util import unimplemented
-from indico.modules.rb.notifications.reservations import (notify_cancellation, notify_confirmation, notify_creation,
-                                                          notify_modification, notify_rejection)
+from indico.modules.rb.notifications.reservations import (notify_cancellation, notify_change_state, notify_confirmation,
+                                                          notify_creation, notify_modification, notify_rejection)
 from indico.modules.rb.util import rb_is_admin
 from indico.util.date_time import format_date, format_time, now_utc
 from indico.util.i18n import N_, _
@@ -423,6 +423,10 @@ class Reservation(Serializer, db.Model):
             if not occurrence.is_valid:
                 continue
             occurrence.reject(user, u'Rejected due to collision with a confirmed reservation')
+
+    def change_state(self, user):
+        notify_change_state(self)
+        self.add_edit_log(ReservationEditLog(user_name=user.full_name, info=['Reservation requires approval']))
 
     def cancel(self, user, reason=None, silent=False):
         self.state = ReservationState.cancelled
--- a/indico/modules/rb/models/reservations.py
+++ b/indico/modules/rb/models/reservations.py
@@ -35,8 +35,8 @@ from indico.modules.rb.models.reservation_edit_logs import ReservationEditLog
 from indico.modules.rb.models.reservation_occurrences import ReservationOccurrence, ReservationOccurrenceState
 from indico.modules.rb.models.room_nonbookable_periods import NonBookablePeriod
 from indico.modules.rb.models.util import unimplemented
-from indico.modules.rb.notifications.reservations import (notify_cancellation, notify_change_state, notify_confirmation,
-                                                          notify_creation, notify_modification, notify_rejection)
+from indico.modules.rb.notifications.reservations import (notify_cancellation, notify_confirmation, notify_creation,
+                                                          notify_modification, notify_rejection, notify_reset_approval)
 from indico.modules.rb.util import rb_is_admin
 from indico.util.date_time import format_date, format_time, now_utc
 from indico.util.i18n import N_, _
@@ -424,8 +424,8 @@ class Reservation(Serializer, db.Model):
                 continue
             occurrence.reject(user, u'Rejected due to collision with a confirmed reservation')
 
-    def change_state(self, user):
-        notify_change_state(self)
+    def reset_approval(self, user):
+        notify_reset_approval(self)
         self.add_edit_log(ReservationEditLog(user_name=user.full_name, info=['Reservation requires approval']))
 
     def cancel(self, user, reason=None, silent=False):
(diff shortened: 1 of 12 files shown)


indico/modules/rb/notifications/reservations.py
```diff
@@ -64,6 +64,27 @@ def compose_email_to_assistance(self, **mail_params):
                     return make_email(to_list=to_list, subject=subject, body=body)
 
 
+@email_sender
+def notify_change_state(reservation):
+    if reservation.is_accepted:
+        raise ValueError('Reservation has not changed state')
```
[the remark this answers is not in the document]
panagiotappl: Never :sweat_smile:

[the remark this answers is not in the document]
ThiefMaster: ok, no need for this check then. btw, my comment about the name from above applies to this function as well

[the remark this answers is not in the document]
panagiotappl: Yep, I figured. Thanks for the comments!

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -240,6 +241,10 @@ def _process(self, args):
         }
 
         self.booking.modify(data, session.user)
+
+        room = self.booking.room
+        if room.reservations_need_confirmation and self.booking.is_accepted:
```
ThiefMaster: this should also check for `not room.can_be_booked(session.user, ignore_admin=True)` - if the user can create regular bookings for the room anyway, there's no need to reset the booking back to pending (since it probably wasn't pending before either)

panagiotappl: Okay, I didn't think of this case. Thanks for pointing out!

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -240,6 +241,11 @@ def _process(self, args):
         }
 
         self.booking.modify(data, session.user)
+
+        room = self.booking.room
+        if room.reservations_need_confirmation and self.booking.is_accepted and\
```
ThiefMaster: we use the `(...)` notation to wrap statements

indico/modules/rb/templates/emails/reservations/change_state_email_to_manager.txt
```diff
@@ -0,0 +1,5 @@
+{% extends 'rb/emails/reservations/base_email_to_manager.txt' %}
+
+{% block body -%}
+A booking on a room you manage has been MODIFIED and the changes require approval.
```
meluru: I'd say: `A booking in a room...`

indico/modules/rb/templates/emails/reservations/change_state_email_to_vc_support.txt
```diff
@@ -0,0 +1,6 @@
+{% extends 'rb/emails/reservations/base_email_to_vc_support.txt' %}
+
+{% block body -%}
+A booking on {{ reservation.room.full_name }} was modified and the changes require approval.
```
meluru: Same here: `booking in room...`

mvidalgarcia: In the other two email templates the word "modified" is in capital letters, I don't know what's the correct way but I think we should be consistent.

ThiefMaster: I'd really like those AWFUL CAPS to GO AWAY ;)

indico/modules/rb/models/reservations.py
```diff
@@ -411,6 +411,12 @@ def get_with_data(*args, **kwargs):
 
         return result.values()
 
+    def is_equal_with(self, new_booking):
```
ThiefMaster: The name is too generic (it sounds like a poor man's `__eq__`), so something like `has_same_dates` would be more suitable.

But maybe it's a bit too simple to justify a method in the model unless we use this in multiple places. And even then, do we have it in a dict in both places? Not sure how the code for splitting looks like right now.
I'd probably simply check it in the controller and eventually move it into an operation function.

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -260,8 +260,13 @@ def _process(self, args):
             'repeat_frequency': args['repeat_frequency'],
             'repeat_interval': args['repeat_interval'],
         }
-
+        has_date_changed = not self.booking.is_equal_with(data)
         self.booking.modify(data, session.user)
+
+        room = self.booking.room
+        if (has_date_changed and room.reservations_need_confirmation and self.booking.is_accepted and
```
ThiefMaster: instead of checking if the room generally uses prebookings i'd check `not can_book and can_prebook`

indico/modules/rb/models/reservations.py
```diff
@@ -432,6 +438,11 @@ def accept(self, user):
                 continue
             occurrence.reject(user, u'Rejected due to collision with a confirmed reservation')
 
+    def reset_approval(self, user):
+        self.is_accepted = False
+        notify_reset_approval(self)
+        self.add_edit_log(ReservationEditLog(user_name=user.full_name, info=['Requiring new approval due to change by']))
```
ThiefMaster: the "by" shouldn't be in there

indico/modules/rb/notifications/reservations.py
```diff
@@ -64,6 +64,25 @@ def compose_email_to_assistance(self, **mail_params):
...
+    return filter(None, [
+        notification.compose_email_to_user(
+            subject='Booking approval changed state on',
+            template_name='change_state_email_to_user'
+        ),
+        notification.compose_email_to_manager(
+            subject='Booking approval changed state on',
+            template_name='change_state_email_to_manager'
+        ),
+        notification.compose_email_to_vc_support(
```
ThiefMaster: i think we can omit this one; VC support is going away from the core anyway

indico/modules/rb/templates/emails/reservations/change_state_email_to_vc_support.txt
```diff
@@ -0,0 +1,6 @@
+{% extends 'rb/emails/reservations/base_email_to_vc_support.txt' %}
+
+{% block body -%}
+A booking in {{ reservation.room.full_name }} was modified and the changes require approval.
```
ThiefMaster: this file isn't used anymore

PR #3702, opened 2019-01-07
Room ACLs

### LINKED ISSUE #1712: Room Booking: Allow certain groups to book a room without confirmation while everyone else needs confirmation
Suggested in RQF0411107.

This could be easily done by having another custom attribute containing a group name which can override "bookings require confirmation". It's most likely a one-line change (unittests not counted) in `Room._can_be_booked`.

### LINKED ISSUE #2165: Room Booking: make "manager group" a 1st class attribute
This seems like something that many people will use, and having it as a "custom attribute" really decreases it importance. We should make it more visible and make sure people understand those peple have the same privileges as room owners.

### LINKED ISSUE #3591: Use a proper ACL for room permissions
Currently we show some rooms as "is not bookable" or "is not publicly available".
Does this distinction in the UI make much sense? Shouldn't we simply show an indicator whether the user can book a given room or not? It doesn't really make a difference if there's a group restriction or if the room is simply configured to be bookable only by its managers.

### LINKED ISSUE #3696: Properly check booking privileges on the server
This depends on #3591.

Currently we simply check for admin privs and if a room uses prebookings to determine what kind of booking someone can create.

Instead of that we should actually return this information from the server to have a single source of truth, especially since we need to perform the check again when actually creating the booking.

Client-side code should already be using `room.canUserBook` and `room.canUserPrebook` for this, so all that's needed on the client is fetching the data with an API call and updating the `getAllRooms` selector to get the information from the correct place instead of the current "simple" check mentioned above.

ThiefMaster:
New room ACLs can't be edited anywhere so far, but that's outside the scope of this PR

closes #3591 
closes #3696
closes #2165
closes #1712


indico/migrations/versions/20181213_1110_cbe630695800_add_room_principals_table.py
```diff
@@ -0,0 +1,216 @@
...
+# revision identifiers, used by Alembic.
+revision = 'cbe630695800'
+down_revision = '252c0015c9a0'
+branch_labels = None
+depends_on = None
+
+
+def _create_acl_entry(conn, room_id, user_id=None, group_id=None, mp_group_provider=None, mp_group_name=None,
+                      full_access=False, permissions=()):
+    permissions = list(permissions)
```
mic4ael: Why not pass the list directly as an argument? It is being used only in one case with one element in a set.

ThiefMaster: i don't like mutable default arguments even if it'd be safe in this case

indico/migrations/versions/20181213_1110_cbe630695800_add_room_principals_table.py
```diff
@@ -0,0 +1,204 @@
...
+    booking_group_attr_id = conn.execute('SELECT id FROM roombooking.room_attributes WHERE name = %s',
+                                         ('allowed-booking-group',)).scalar()
+    manager_group_attr_id = conn.execute('SELECT id FROM roombooking.room_attributes WHERE name = %s',
+                                         ('manager-group',)).scalar()
+    if booking_group_attr_id is None:
+        booking_group_attr_id = _create_attribute(conn, 'allowed-booking-group', 'Allowed Booking Group')
+    if manager_group_attr_id is None:
+        manager_group_attr_id = _create_attribute(conn, 'manager-group', 'Manager Group')
+
+    query = 'SELECT id, owner_id, reservations_need_confirmation, protection_mode from roombooking.rooms'
```
mic4ael: I know it is something trivial but to keep it consistent maybe change `from` to `FROM`?

indico/modules/rb_new/client/js/common/user/reducers.js
```diff
@@ -60,6 +66,7 @@ export default combineReducers({
                     avatarBgColor: user.avatar_bg_color,
                     language: user.language,
                     isAdmin: user.is_admin,
+                    isRBAdmin: user.is_rb_admin,
```
mic4ael: Can't we just do sth like this `return camelizeKeys(action.data.user)`?

ThiefMaster: I think that'd give us `isRbAdmin` here

indico/modules/rb_new/client/js/modules/bookRoom/SearchResultCount.jsx
```diff
@@ -106,6 +108,26 @@ class SearchResultCount extends React.Component {
...
     }
 
+    renderUnbookable(count) {
+        const label = <Label color="red" horizontal size="small">{count}</Label>;
+        const trigger = (
+            <Menu.Item as="span">
+                <Icon name="lock" />
+                <Translate>
+                    Unauthorized
+                    <Param name="count" value={label} />
```
mic4ael: Was it intended to put it in a separate line?

ThiefMaster: There'd be an actual space if it was in the same line (linebreaks between text and elements are ignored in JSX). Not sure if it'd cause any display issues here though. Anyway, nothing wrong with having the param in a separate line IMO.

PR #3704, opened 2019-01-10
rb_new: use upstream react-leaflet

pferreir:
See [this](https://github.com/PaulLeCam/react-leaflet/pull/537) for context. I cannot find any other way of making it work with the upstream version besides creating a wrapper component. See the comments for context.


indico/modules/rb_new/client/js/common/map/MapMarkers.jsx
```diff
@@ -0,0 +1,119 @@
...
+    shouldComponentUpdate({rooms: newRooms, hoveredRoomId: newHovered}) {
+        const {rooms: oldRooms, hoveredRoomId: oldHovered} = this.props;
+        return !_.isEqual(oldRooms, newRooms) || newHovered !== oldHovered;
+    }
+
+    render() {
+        const {rooms, clusterProps, hoveredRoomId} = this.props;
+
+        return (
+            !!rooms.length && (
```
ThiefMaster: what about an `if (!rooms.length) { return null; }` to get out early?

indico/modules/rb_new/client/js/common/map/MapMarkers.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+const icon = Leaflet.divIcon({className: 'rb-map-marker', iconSize: [20, 20]});
+const hoveredIcon = Leaflet.divIcon({className: 'rb-map-marker highlight', iconSize: [20, 20]});
+
+/**
+ * This wrapper is used to avoid re-rendering of <Marker> on every parent
+ * cycle. By adding MarkerWrapper, we gain control of that through
+ * shouldComponentUpdate(...)
+ */
+class MarkerWrapper extends React.Component {
```
mvidalgarcia: Don't you need to declare `highlight` as `propTypes` here?

pferreir: Indeed. I wonder why ESLint didn't complain...

pferreir: Oh, it did (Travis), it's my local ESLint that is silly.

PR #3717, opened 2019-01-17
Booking link to event/contribution/session block

### LINKED ISSUE #3707: Link room booking to event/contribution/session block
This issue is part from #3693 
- Update `Reservation` model to add `contribution_id` and `session_block_id`
- URL query parameters that allow linking the booking to an event, contribution or session block
- Display appropriate information in the book a room modal indicating the that the booking is going to be linked
- Only allow linking bookings to event managers
- Display appropriate info in the booking modal ("This booking is linked to foo event")

mvidalgarcia:
Closes #3707 

TODO:
- [ ] When linking a booking to an object, update location+room of that object


indico/modules/rb/models/reservations.py
```diff
@@ -105,6 +106,12 @@ class Reservation(Serializer, db.Model):
         'is_rejected', ('location_name', 'location'), ('contact_email', 'booked_for_user_email')
     ]
 
+    allowed_link_types = {LinkType.contribution, LinkType.session_block}
```
ThiefMaster: `LinkType.event` is missing here

indico/modules/rb/models/reservations.py
```diff
@@ -105,6 +106,12 @@ class Reservation(Serializer, db.Model):
...
+    unique_links = True
+    optional_links = True
+    events_backref_name = 'reservations'
```
ThiefMaster: `room_reservations`?

indico/modules/rb/models/reservations.py
```diff
@@ -105,6 +106,12 @@ class Reservation(Serializer, db.Model):
...
+    link_backref_name = 'reservation'
```
ThiefMaster: `room_reservation`?

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -249,13 +252,27 @@ def _process(self):
...
     def _process(self):
-        if self.booking.event is None:
+        if self.linked_object is None:
             raise NotFound
-        if not self.booking.event.can_access(session.user):
+        if isinstance(self.linked_object, Event):
+            event = self.linked_object
+        else:
+            event = self.linked_object.event
+        if not event.can_access(session.user):
```
ThiefMaster: Why not check access on the linked object? An event you can access might have protected contributions!

indico/modules/rb_new/blueprint.py
```diff
@@ -39,6 +39,8 @@
 _bp.add_url_rule('/api/map-areas', 'map_areas', misc.RHMapAreas)
 _bp.add_url_rule('/api/equipment', 'equipment_types', misc.RHEquipmentTypes)
 _bp.add_url_rule('/api/locations', 'locations', locations.RHLocations)
+_bp.add_url_rule('/api/data/<any(event,contrib,session_block):type>/<int:id>', 'linked_object_data',
```
ThiefMaster: `/api/link-info/...`? "data" is way too generic..

also, i'd `s/contrib/contribution/`

indico/modules/rb/models/reservations.py
```diff
@@ -91,7 +92,7 @@ def convert_legacy_repeatability(cls, repeat):
             raise KeyError('Undefined old repeat: {}'.format(repeat))
 
 
-class Reservation(Serializer, db.Model):
+class Reservation(Serializer, LinkMixin, db.Model):
```
ThiefMaster: What about keeping all the link information in a separate model, something like `ReservationLink`, that has an 1:1 relationship to `Reservation.?
That way we keep stuff we don't need very often out of the Reservation model and there aren't any drawbacks except that linking is slightly more complicated (need to create the ReservationLink if it doesn't exist yet).

indico/modules/rb_new/client/js/common/bookings/BookingObjectLink.jsx
```diff
@@ -0,0 +1,117 @@
...
+            sessionBlock: Translate.string('a session block')
+        };
+        const verb = potential ? <Translate> will be </Translate> : <Translate> is </Translate>;
+        return !!mapping[type] && canAccess && (
+            <Message icon color="teal">
+                <Icon name="linkify" />
+                <Message.Content>
+                    <>
+                        <Translate>
+                            This booking <Param name="verb" value={verb} /> linked to <Param name="type" wrapper={<strong />}
```
ThiefMaster: this can be wrapped in a more readable way, even without requiring `{' '}` to keep proper spacing:
```jsx
                            This booking <Param name="verb" value={verb} /> linked
                            to <Param name="type" wrapper={<strong />} value={mapping[type]} />:
```

indico/modules/rb_new/client/js/common/bookings/BookingObjectLink.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+    renderObjectInfo(data) {
+        const {type, potential} = this.props;
+        const {url, title, canAccess, eventTitle, eventUrl} = data;
+        const mapping = {
+            event: Translate.string('an event'),
+            contrib: Translate.string('a contribution'),
+            sessionBlock: Translate.string('a session block')
+        };
+        const verb = potential ? <Translate> will be </Translate> : <Translate> is </Translate>;
```
ThiefMaster: no no no, assembling a sentence from multiple translated snippets doesn't work. it's not the same structure in all languages!

so you will need to duplicate the whole sentence a few times here. not super pretty but that way it can be translated properly

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -249,13 +252,27 @@ def _process(self):
...
 
-class RHBookingEventData(RHBookingBase):
+class RHLinkedObjectData(RHRoomBookingBase):
+    """Fetch data from event, contribution or session block"""
+    def _process_args(self):
+        type = request.view_args['type']
+        id = request.view_args['id']
+        self.linked_object = {
+            'event': lambda x: Event.get_one(x),
+            'contrib': lambda x: Contribution.get_one(x),
```
ThiefMaster: ensure it's not deleted / in a deleted event

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -249,13 +252,27 @@ def _process(self):
...
+            'session_block': lambda x: SessionBlock.get_one(x)
```
ThiefMaster: ensure it's not in a deleted event

indico/modules/rb_new/schemas.py
```diff
@@ -130,7 +132,10 @@ class ReservationDetailsSchema(mm.ModelSchema):
     can_delete = Function(lambda booking: booking.can_be_deleted(session.user))
     can_modify = Function(lambda booking: booking.can_be_modified(session.user))
     can_reject = Function(lambda booking: booking.can_be_rejected(session.user))
-    is_linked_to_event = Function(lambda booking: booking.event_id is not None)
+    is_linked_to_object = Function(lambda booking:
+                                   booking.event_id is not None or
```
ThiefMaster: just check the `link_type`

mvidalgarcia:
Updated.

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -171,6 +175,19 @@ def _validate_room_booking_limit(self, start_dt, end_dt, booking_limit_days):
...
+        if id_ and type_:
+            obj = None
+            if type_ is LinkType.event:
+                obj = Event.get(id_)
+            elif type_ is LinkType.contribution:
+                obj = Contribution.get(id_)
+            elif type_ is LinkType.session_block:
+                obj = SessionBlock.get(id_)
+
+            if obj is not None:
```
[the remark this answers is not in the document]
ThiefMaster: put this in a util function since you need it again in `RHLinkedObjectData`

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -258,13 +276,33 @@ def _process(self):
...
+    def _process_args(self):
+        type_ = request.view_args['type']
+        id_ = request.view_args['id']
+        obj = {
+            'event': lambda x: Event.get_one(x),
+            'contribution': lambda x: Contribution.get_one(x),
+            'session_block': lambda x: SessionBlock.get_one(x)
+        }[type_](id_)
+
+        if type_ is 'event':
```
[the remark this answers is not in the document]
ThiefMaster: also, it's not needed since `Event` has a property named `event` which simply returns `self`

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -258,13 +276,33 @@ def _process(self):
...
+            raise ValidationError(_('The associated event is deleted'))
+        else:
+            self.linked_object = obj
+
     def _process(self):
-        if self.booking.event is None:
+        if self.linked_object is None:
             raise NotFound
-        if not self.booking.event.can_access(session.user):
+        if not self.linked_object.can_access(session.user):
```
ThiefMaster: maybe `can_manage` makes more sense since the link to link is only available from the management area of the event?

indico/modules/rb_new/schemas.py
```diff
@@ -77,11 +78,13 @@ class Meta:
         fields = ('id', 'booking_reason', 'booked_for_name', 'room_id', 'is_accepted')
 
 
-class ReservationEventDataSchema(Schema):
+class ReservationLinkedObjectDataSchema(Schema):
     id = Number()
     title = String()
     url = String()
-    can_access = Function(lambda event: event.can_access(session.user))
+    event_title = Function(lambda obj: obj.event.title)
```
ThiefMaster: we shouldn't be returning this information at all if the user cannot access the object

indico/modules/rb_new/schemas.py
```diff
@@ -133,7 +136,9 @@ class ReservationDetailsSchema(mm.ModelSchema):
     can_edit = Function(lambda booking: booking.can_edit(session.user))
     can_reject = Function(lambda booking: booking.can_reject(session.user))
     state = EnumField(ReservationState)
-    is_linked_to_event = Function(lambda booking: booking.event_id is not None)
+    is_linked_to_object = Function(lambda booking: booking.link_id is not None)
+    link_type = Function(lambda booking: booking.link.link_type.name if booking.link_id else None)
```
ThiefMaster: Would be much nicer if we could use an `EnumField` here... But I guess marshmallow can't handle this yet due to the nesting? (marshmallow-code/marshmallow#611)

ThiefMaster: you can also use `if booking.link` here; sqlalchemy is smart enough to not trigger a load if the FK is None

ThiefMaster: actually, what about making link a `Nested` here? then you can use a subschema with an EnumField

indico/modules/rb_new/schemas.py
ThiefMaster: actually, we abort early in the RH if the user has no access so we can always return True for `can_access` here since we did the proper access check in the RH

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -258,13 +276,33 @@ def _process(self):
...
+            'session_block': lambda x: SessionBlock.get_one(x)
+        }[type_](id_)
+
+        if type_ is 'event':
+            event = obj
+        else:
+            event = obj.event
+
+        if event.is_deleted:
+            raise ValidationError(_('The associated event is deleted'))
```
ThiefMaster: `ValidationError` seems wrong here, we aren't in any validation function. what about just handling this like an object with no access?

mvidalgarcia:
Updated.

indico/modules/rb_new/client/js/common/bookings/BookingObjectLink.jsx
```diff
@@ -80,7 +80,7 @@ export default class BookingObjectLink extends React.PureComponent {
             contribution: Translate.string('This booking is linked to a contribution:'),
             sessionBlock: Translate.string('This booking is linked to a session block:'),
         };
-        return canAccess && (
+        return canManage && (
```
ThiefMaster: for this one access is fine (that's when viewing a booking)

mvidalgarcia: Not always, it could be right before linking (book a room). I guess I need to check `pending` also.

ThiefMaster: I think that's an edge case which is OK-ish since you won't get there except by messing with the link. The whole point of these access checks is to:

- not allow someone to link an event they don't manage
- not disclose information about events they cannot see otherwise

being able to see information they could access anyway isn't a problem

mvidalgarcia: Fair enough, updated.

indico/modules/rb/models/reservations.py
```diff
@@ -251,6 +279,15 @@ def __table_args__(cls):
         )
     )
 
+    link = db.relationship(
+        'ReservationLink    ',
```
mic4ael: Trailing spaces?

indico/modules/rb_new/client/js/common/bookings/BookingObjectLink.jsx
```diff
@@ -0,0 +1,118 @@
...
+        const linkedMessages = {
+            event: Translate.string('This booking is linked to an event:'),
+            contribution: Translate.string('This booking is linked to a contribution:'),
+            sessionBlock: Translate.string('This booking is linked to a session block:'),
+        };
+        return canAccess && (
+            <Message icon color="teal">
+                <Icon name="linkify" />
+                <Message.Content>
+                    <>
```
mic4ael: I don't think you need `React.Fragment` here.

indico/migrations/versions/20190118_1514_416f9c877300_create_reservationlink_model.py
```diff
@@ -0,0 +1,88 @@
...
+            WHERE event_id IS NOT NULL
+        ), link_ids_data AS (
+            INSERT INTO roombooking.reservation_links (id, event_id, linked_event_id, link_type)
+            SELECT new_link_id, event_id, event_id, 2 -- LinkType.event
+            FROM reserv_data
+            RETURNING id AS link_id
+        )
+        UPDATE roombooking.reservations
+        SET link_id = link_ids_data.link_id
+        FROM link_ids_data,reserv_data
```
ThiefMaster: missing space after `,`

indico/migrations/versions/20190118_1514_416f9c877300_create_reservationlink_model.py
```diff
@@ -0,0 +1,88 @@
...
+        WHERE id = reserv_data.reserv_id and link_ids_data.link_id = reserv_data.new_link_id;
+    ''')
+
+    op.drop_column('reservations', 'event_id', schema='roombooking')
+
+
+def downgrade():
+    op.add_column('reservations', sa.Column('event_id', sa.Integer(), autoincrement=False, nullable=True),
```
ThiefMaster: `autoincrement=False` can be omitted, alembic is just overzealous when autogenerating this

indico/modules/rb_new/client/js/common/bookings/BookingObjectLink.jsx
```diff
@@ -15,20 +15,31 @@
...
     static propTypes = {
-        bookingId: PropTypes.number.isRequired,
+        type: PropTypes.oneOf(['event', 'contribution', 'sessionBlock']).isRequired,
+        id: PropTypes.number.isRequired,
+        /** Whether it is a pending link or the booking is already linked */
+        pending: PropTypes.bool,
+    };
+
+    static defaultProps = {
+        pending: false
```
ThiefMaster: trailing comma? ;)

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -174,6 +176,13 @@ def _validate_room_booking_limit(self, start_dt, end_dt, booking_limit_days):
         selected_period_days = (day_end_dt - day_start_dt).days
         return selected_period_days <= booking_limit_days
 
+    def _link_booking(self, booking, type_, id_):
+        if id_ and type_:
```
ThiefMaster: 0 could be a valid id :) very unlikely to happen but migrated databases to have an event and contribution with id 0.

so better to explicitly check for None here - and you could return early if type/id are missing and thus outdent the rest of this method

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -261,13 +271,20 @@ def _process(self):
         return jsonify(booking_id=booking_id, room_id=room_id)
 
 
-class RHBookingEventData(RHBookingBase):
+class RHLinkedObjectData(RHRoomBookingBase):
+    """Fetch data from event, contribution or session block"""
```
ThiefMaster: empty line after a class docstring

indico/modules/rb_new/util.py
```diff
@@ -81,3 +87,22 @@ def serialize_nonbookable_periods(data):
 
 def serialize_unbookable_hours(data):
     return [bookable_hours_schema.dump(d).data for d in data]
+
+
+def get_linked_object(type_, id_):
+    obj = None
+    if type_ == LinkType.event:
+        obj = Event.get_one(id_, is_deleted=False)
```
ThiefMaster: `return ...` inside the ifs; no need to assign first and return at the end

ThiefMaster:
For session blocks with no title the message in the modal looks weird. Better use the `full_title` of the SessionBlock instead. I'm afraid that's going to make `ReservationLinkedObjectDataSchema` much more ugly than it's right now :/ Maybe you can use a Method field and check the object's type to determine whether to use `title` or `full_title`..

ThiefMaster:
image

Shouldn't the title be italic/bold? To me it feels a bit odd that it's the same style as the text right above.

indico/modules/rb_new/client/js/common/bookings/BookingObjectLink.jsx
```diff
@@ -15,20 +15,31 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
+import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import {Message, Placeholder, Segment, Icon} from 'semantic-ui-react';
-import getBookingEventDataURL from 'indico-url:rooms_new.booking_event_data';
+import getLinkedObjectDataURL from 'indico-url:rooms_new.linked_object_data';
```
ThiefMaster: move that import to the top (separate block of imports) like we do e.g. in the actions

mvidalgarcia: Then we should also change this on `SpriteImage.jsx`, `RoomSelector.jsx` and `RoomStats.jsx`

mvidalgarcia:
Updated :arrows_clockwise:

PR #3724, opened 2019-01-18
Prevent from showing invalid suggestions

mic4ael:


indico/modules/rb_new/operations/suggestions.py
```diff
@@ -52,8 +53,15 @@ def get_single_booking_suggestions(rooms, start_dt, end_dt, limit=None):
     data = []
     new_start_dt = start_dt - timedelta(minutes=BOOKING_TIME_DIFF)
     new_end_dt = end_dt + timedelta(minutes=BOOKING_TIME_DIFF)
+    blocked_rooms = get_rooms_blockings(rooms, start_dt.date(), end_dt.date())
+    unbookable_hours = get_rooms_unbookable_hours(rooms)
+    nonbookable_periods = get_rooms_nonbookable_periods(rooms, start_dt, end_dt)
+    conflicts = set(get_rooms_conflicts(rooms, start_dt, end_dt, RepeatFrequency.NEVER, 0, blocked_rooms,
+                                        nonbookable_periods, unbookable_hours)[0])
```
meluru: Correct me if I am wrong, but in case, when we are trying to book a room from 10:45 to 12.00 and the room is only bookable between 11:00 and 18:00, then it won't show up at all? it could be a suggestion (15 minutes later)

ThiefMaster: Possibly, there are still a few cases where suggestions don't work that great :/

mic4ael: Yeah, for now it is a bit limited. I think we want to improve that part anyway later on.

PR #3730, opened 2019-01-22
Fix sticky filter bars

### LINKED ISSUE #3722: Fix broken sticky filter bar
(apparently broken in all pages)

mic4ael:
fixes #3722

### DIFF
--- a/indico/modules/rb_new/client/js/components/App.module.scss
+++ b/indico/modules/rb_new/client/js/components/App.module.scss
@@ -96,7 +96,15 @@
     }
 }
 
-.rb-pushable, .rb-pusher:global(.pusher) {
+.rb-pushable {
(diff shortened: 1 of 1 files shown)


indico/modules/rb_new/client/js/components/App.module.scss
```diff
@@ -96,7 +96,13 @@
     }
 }
 
-.rb-pushable, .rb-pusher:global(.pusher) {
+.rb-pushable {
     // make sure that side bar takes the whole remaining height of the page
     min-height: calc(100vh - #{$menu-height} - 50px);
+    transform: initial !important;
```
pferreir: A comment explaining the need for this would be nice. We won't be able to figure it out in a few months, otherwise.

PR #3733, opened 2019-01-22
Plan B improvements

### LINKED ISSUE #3500: Improve wording in "plan B" results
It should be clearer what they mean and that they are available. We should also make it clearer what they are in timeline view (e.g. a different header?)

pferreir:
image

This PR contains:
 * Better Plan B option display;
 * Tooltips when hovering Plan B options;
 * Refactored Plan B suggestion rendering;
 * CSS now from [Indico Semantic UI Theme](https://github.com/pferreir/indico-sui-theme/);
 * Indico palette slightly adapted to fit the theme's colour scheme (have a look to see if you find any inconsistencies);
Closes #3500

### DIFF
--- a/indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
+++ b/indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
@@ -22,15 +22,14 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import {bindActionCreators} from 'redux';
 import _ from 'lodash';
-import {Button, Card, Grid, Header, Icon, Label, Message, Popup, Sticky, Divider} from 'semantic-ui-react';
+import {Button, Card, Grid, Header, Icon, Popup, Sticky, Divider} from 'semantic-ui-react';
 import LazyScroll from 'redux-lazy-scroll';
 
-import {PluralTranslate, Translate, Singular, Param, Plural} from 'indico/react/i18n';
+import {Translate} from 'indico/react/i18n';
 import {Overridable, Slot, toClasses, IndicoPropTypes} from 'indico/react/util';
 import {serializeTime, serializeDate} from 'indico/utils/date';
 
 import searchBarFactory from '../../components/SearchBar';
-import Room from '../../components/Room';
 import CardPlaceholder from '../../components/CardPlaceholder';
 import BookingFilterBar from './BookingFilterBar';
 import {roomFilterBarFactory} from '../../modules/roomList';
@@ -44,6 +43,7 @@ import {actions as filtersActions} from '../../common/filters';
 import {actions as roomsActions, RoomRenderer} from '../../common/rooms';
 import * as bookRoomSelectors from './selectors';
 import {mapControllerFactory, selectors as mapSelectors} from '../../common/map';
+import BookingSuggestion from './BookingSuggestion';
 
 import './BookRoom.module.scss';
 
@@ -315,72 +315,17 @@ class BookRoom extends React.Component {
                 <Header styleName="header">
                     <Translate>Here are some alternatives we've found for you!</Translate>
                 </Header>
-                <Card.Group styleName="suggestions" stackable>
-                    {suggestions.map((suggestion) => this.renderSuggestion(suggestion))}
+                <Card.Group stackable>
+                    {suggestions.map(({room, suggestions: roomSuggestions}) => (
+                        <BookingSuggestion key={room.id}
+                                           room={room}
+                                           suggestions={roomSuggestions}
+                                           onClick={(overrides = null) => {
+                                               this.openBookingForm(room, overrides, !room.canUserBook);
+                                           }} />
+                    ))}
                 </Card.Group>
             </div>
-        );
-    };
-
-    renderSuggestion = ({room, suggestions}) => {
-        return (
-            <Room key={room.id} room={room}>
-                <Slot>
-                    <Grid centered styleName="suggestion" columns={3}>
-                        <Grid.Column verticalAlign="middle" width={14}>
-                            {this.renderSuggestionText(room, suggestions)}
-                        </Grid.Column>
-                    </Grid>
-                </Slot>
-            </Room>
-        );
-    };
-
-    renderSuggestionText = (room, {time, duration, skip}) => {
-        return (
-            <>
-                {time && (
-                    <Message styleName="suggestion-text" size="mini"
-                             onClick={() => this.openBookingForm(room, {time}, !room.canUserBook)}
-                             warning compact>
-                        <Message.Header>
-                            <Icon name="clock" />
-                            <PluralTranslate count={time}>
-                                <Singular>
-                                    One minute <Param name="modifier" value={time < 0 ? 'earlier' : 'later'} />
-                                </Singular>
-                                <Plural>
-                                    <Param name="count" value={Math.abs(time)} /> minutes{' '}
-                                    <Param name="modifier" value={time < 0 ? 'earlier' : 'later'} />
-                                </Plural>
-                            </PluralTranslate>
-                        </Message.Header>
-                    </Message>
-                )}
-                {duration && time && (
-                    <div>
-                        <Label color="brown" circular>{Translate.string('or')}</Label>
-                    </div>
-                )}
-                {duration && (
-                    <Message styleName="suggestion-text" size="mini"
-                             onClick={() => this.openBookingForm(room, {duration}, !room.canUserBook)}
-                             warning compact>
-                        <Message.Header>
-                            <Icon name="hourglass full" /> {PluralTranslate.string('One minute shorter', '{duration} minutes shorter', duration, {duration})}
-                        </Message.Header>
-                    </Message>
-                )}
-                {skip && (
-                    <Message styleName="suggestion-text" size="mini"
-                             onClick={() => this.openBookingForm(room, null, !room.canUserBook)}
-                             warning compact>
-                        <Message.Header>
-                            <Icon name="calendar times" /> {PluralTranslate.string('Skip one day', 'Skip {skip} days', skip, {skip})}
-                        </Message.Header>
-                    </Message>
-                )}
-            </>
         );
     };
 
--- a/indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
+++ b/indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
@@ -22,7 +22,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import {bindActionCreators} from 'redux';
 import _ from 'lodash';
-import {Button, Card, Grid, Header, Icon, Label, Message, Popup, Sticky} from 'semantic-ui-react';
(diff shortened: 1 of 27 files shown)


indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
```diff
@@ -309,78 +309,26 @@ class BookRoom extends React.Component {
 
         return (
             <>
-                <Header as="h2">
-                    <Translate>Rooms that you might be interested in</Translate>
+                <Divider horizontal>
+                    <Icon name="magic" style={{fontSize: '2em'}} circular />
+                </Divider>
+                <Header style={{margin: '0 0 2em 0'}}>
+                    <Translate>Here are some other possibilities we've calculated for you!</Translate>
```
ThiefMaster: "calculated" sounds a bit strange IMO - I've never seen that term used for any kind of recommendation on websites

indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
```diff
@@ -309,78 +309,26 @@ class BookRoom extends React.Component {
 
         return (
             <>
-                <Header as="h2">
-                    <Translate>Rooms that you might be interested in</Translate>
+                <Divider horizontal>
+                    <Icon name="magic" style={{fontSize: '2em'}} circular />
```
ThiefMaster: not covered by any of the predefined sizes?

indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
```diff
@@ -309,78 +309,26 @@ class BookRoom extends React.Component {
...
+                </Divider>
+                <Header style={{margin: '0 0 2em 0'}}>
```
ThiefMaster: move to scss file?

indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
```diff
@@ -0,0 +1,130 @@
+/* This file is part of Indico.
+ * Copyright (C) 2002 - 2018 European Organization for Nuclear Research (CERN).
```
ThiefMaster: 2019 (we should run the update script at some point, but maybe we can wait with that until we do the license change as well?)

indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
```diff
@@ -0,0 +1,130 @@
...
+
+import Room from '../../components/Room';
+
+import './BookingSuggestion.module.scss';
+
+
+function SuggestionOption({onClick, overrides, icon, text, tooltip}) {
+    const message = (
+        <Message styleName="suggestion-text" size="mini"
+                 onClick={onClick.bind(undefined, overrides)}
```
ThiefMaster: just curious: why do we need to explicitly clear `this`?

indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Renders a "suggested room" containing one or more alternative options
+ * e.g. "10 minutes later", "Skip one day"...
+ */
+export default class BookingSuggestion extends React.PureComponent {
+    static propTypes = {
+        /** Room information */
+        room: PropTypes.object.isRequired,
+        /** Suggestions as {time, duration, skip} */
+        suggestions: PropTypes.object.isRequired,
```
ThiefMaster: i'd document the keys via `PropTypes.shape()`

indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
```diff
@@ -0,0 +1,130 @@
...
+
+    renderSuggestionText(room, {time, duration, skip}) {
+        const {onClick} = this.props;
+        return (
+            <>
+                {time && (
+                    <SuggestionOption onClick={onClick}
+                                      overrides={{time}}
+                                      icon="clock"
+                                      text={(
```
ThiefMaster: what about using children for this instead of a normal prop?

indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
```diff
@@ -0,0 +1,130 @@
...
+                                          <PluralTranslate count={time}>
+                                              <Singular>
+                                                  One minute <Param name="modifier" value={time < 0 ? 'earlier' : 'later'} />
+                                              </Singular>
+                                              <Plural>
+                                                  <Param name="count" value={Math.abs(time)} /> minutes{' '}
+                                                  <Param name="modifier" value={time < 0 ? 'earlier' : 'later'} />
```
ThiefMaster: that's not how i18n works :P 

probably need to repeat the whole translate block for the two versions :/

indico/modules/rb_new/client/js/setup.jsx
```diff
@@ -11,7 +11,7 @@ import {selectors as configSelectors} from './common/config';
 import {selectors as userSelectors, actions as userActions} from './common/user';
 import {actions as roomsActions} from './common/rooms';
 
-import 'semantic-ui-css/semantic.css';
+import 'indico-sui-theme/dist/semantic.css';
```
ThiefMaster: any easy way to skip the `dist/` when building the theme package?

package.json
```diff
@@ -24,6 +24,7 @@
     "fullcalendar": "^3.10.0",
     "geodesy": "^1.1.3",
     "history": "^4.7.2",
+    "indico-sui-theme": "github:pferreir/indico-sui-theme#master",
```
ThiefMaster: npm? ;)

indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
mic4ael: Out of curiosity, what is the problem here?

pferreir: It wasn't me writing the original, but you are indeed right.
@mic4ael there are two problems:
 * `earlier` and `later` are not translated;
 * Building sentences from pieces is not great for translation, due to languages that work differently from English (e.g. declensions, etc...). We actually mention this [here](https://github.com/indico/indico/wiki/Internationalisation#translating-variable-sentences).

indico/modules/rb_new/client/js/setup.jsx
pferreir: We could put the dist files in the package root. But is it that bad that you get to see the `dist/`?

ThiefMaster: just seems a bit ugly to have this in the import since it's the only thing from the package that ever gets imported :see_no_evil:

indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
mic4ael: Mhm, right! Thanks!

package.json
pferreir: Well, I was waiting for it to be a little bit more stable.

indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
pferreir: Actually, setting the size doesn't change much because somehow the icon limits itself to the height of the divider. But I will move the override to the SCSS module.

indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
pferreir: Thought of it, but then there's the "tooltip" field as well. We could also use `Slot`s, but passing as an arg doesn't seem to be an uncommon pattern in React.

ThiefMaster: Just thought because it's quite big. But since the code to generate this message will probably end up in a separate function to fix the i18n I guess that point becomes moot anyway

indico/modules/rb_new/client/styles/palette.scss
```diff
@@ -22,12 +22,10 @@ $highlight-color: #7dcfb6;
 $dark-highlight-color: #45ba97;
 $primary-color: #2185d0;
 
-$sui-red: #db2828;
-$sui-orange: #f2711c;
-$sui-green: #21ba45;
-$sui-grey: #767676;
-
-$light-red: #ffbdbd;
+$sui-red: $red;
```
mvidalgarcia: Just out of curiosity, these new `sui` colors are being changed because are the default ones in the new indico theme?

pferreir: Yes. I tried to find a palette that doesn't look too different from the one we use in the rest of Indico and that fits the Semantic UI... ehm... semantics.
The colours are [here](https://github.com/pferreir/indico-sui-theme/blob/master/theme/indico/globals/site.variables#L23).

indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
```diff
@@ -0,0 +1,146 @@
...
+                    <Param name="count" value={Math.abs(time)} /> minutes earlier
+                </Plural>
+            </PluralTranslate>
+        ) : (
+            <PluralTranslate count={time}>
+                <Singular>
+                  One minute later
+                </Singular>
+                <Plural>
+                    <Param name="count" value={Math.abs(time)} /> minutes later
```
ThiefMaster: no need for abs here

indico/modules/rb_new/client/styles/palette.scss
```diff
@@ -22,12 +22,10 @@ $highlight-color: #7dcfb6;
...
+$sui-orange: $orange;
+$sui-green: $green;
+$sui-grey: $gray;
```
ThiefMaster: :see_no_evil: yay for different spellings of gr[ea]y :disappointed: 

`s/sui-grey/sui-gray/` after merging this?

indico/web/client/styles/base/_palette.scss
```diff
@@ -48,8 +48,8 @@ $yellow: #e99e18;
 $postit-yellow: #ffffa5;
 $light-yellow: #ffe6ba;
 
-$orange: #b14300;
-$bright-orange: #e25300;
+$orange: #f80;
```
ThiefMaster: won't these changes in here affect other parts of indico quite a bit?

ThiefMaster: yeah, "main categories" on the indico front page looks different now. looks a bit strange IMO

indico/modules/rb_new/client/styles/palette.scss
pferreir: Actually, `$sui-*` will go away, now that the palettes are unified.

pferreir: ... done.

PR #3739, opened 2019-01-29
Consider max-advance-days in room filtering

### LINKED ISSUE #3734: Check max_advance_days
Rooms that cannot be booked due to this should show up in "unauthorized".
The check should be done client side, to avoid the cost of a check for permissions server-side.

panagiotappl:
Filtering happens in client side
closes #3734


indico/modules/rb_new/client/js/common/user/selectors.js
```diff
@@ -45,7 +47,18 @@ export const getAllUserRoomPermissions = createSelector(
 
 export const getUnbookableRoomIds = createSelector(
     getAllUserRoomPermissions,
-    permissions => Object.entries(permissions).filter(([, perms]) => !perms.book && !perms.prebook).map(([id]) => +id)
+    ({rooms}) => rooms.rooms,
+    ({bookRoom}) => bookRoom.filters.dates.startDate,
+    (permissions, rooms, startDate) => {
```
ThiefMaster: `getUnbookableRoomIds` is not the right place for this. It's meant to return the room IDs the user can *never* book regardless of date/time.

ThiefMaster: `getUnbookableRoomIds` is used in many places, including the calendar, and there no bookRoom date filter is present so this might break.

panagiotappl: Does this mean it probably affects the code somewhere else when it shouldn't be?

panagiotappl: Okay, you replied even before I asked :P

ThiefMaster: You could add a new selector that combines data from `getUnbookableRoomIds` and your extra logic and then use that one when getting the data in the various selectors inside the bookRoom module.

indico/modules/rb_new/client/js/common/user/selectors.js
```diff
@@ -15,6 +15,7 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
+
```
ThiefMaster: let's remove this ;)

indico/modules/rb_new/client/js/modules/bookRoom/selectors.js
```diff
@@ -15,6 +15,8 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
+import moment from 'moment';
+
```
ThiefMaster: no need for that empty line either

indico/modules/rb_new/client/js/modules/bookRoom/selectors.js
```diff
@@ -98,13 +100,28 @@ export const isSearchAndPermissionCheckFinished = createSelector(
...
 
+/** Get rooms ids that are unbookable by the user and also take into
+ *  account the maximum advance days property of the room
+ */
+export const getUnbookableRoomIdsWithMaxAdvanceDays = createSelector(
+    userSelectors.getUnbookableRoomIds,
+    roomsSelectors.getAllRooms,
+    getFilters,
+    (unbookableRoomIds, rooms, {dates: {startDate}}) => {
+        const roomIds = Object.entries(rooms).filter(([, room]) => (
```
ThiefMaster: you have to filter out rooms with `!maxAdvanceDays` since that means there's no restriction

indico/modules/rb_new/client/js/modules/bookRoom/selectors.js
```diff
@@ -98,13 +99,28 @@ export const isSearchAndPermissionCheckFinished = createSelector(
...
+            room.maxAdvanceDays && moment().add(room.maxAdvanceDays, 'days') < moment(startDate))
```
ThiefMaster: in case of `room.canUserOverride` there's no need to check anything here

ThiefMaster: You could split the filters to improve readability:

```js
        const roomIds = Object.entries(rooms)
            .filter(([, room]) => !room.canUserOverride && room.maxAdvanceDays)
            .filter(([, room]) => moment(startDate) > moment().add(room.maxAdvanceDays, 'days'))
            .map(([id]) => +id);
```

PR #3744, opened 2019-02-01
Book from event

### LINKED ISSUE #3693: "Book from event" option
After some discussion, the best solution seems to be something that doesn't look that different from what we currently have right now.
So, the user would see three options:
 1. Book room for event;
 2. Book room for session block;
 3. Book room for contribution;

Selecting 2 or 3 would result in a "select box" with a search functionality being shown.
The user would click a "Book" button which would open a new browser tab with RB pre-configured through the magic parameter implemented in #3707.

mvidalgarcia:
closes #3693


indico/modules/rb_new/controllers/management/contributions.py
```diff
@@ -0,0 +1,40 @@
...
+
+from indico.modules.events.contributions import Contribution
+from indico.modules.events.management.controllers import RHManageEventBase
+
+
+class RHListOtherContributions(RHManageEventBase):
+    """AJAX endpoint that lists all contributions in the event (dict representation)."""
+
+    def _process(self):
+
```
ThiefMaster: save the planet, don't waste valuable bytes on empty lines! :D

indico/modules/rb_new/controllers/management/contributions.py
```diff
@@ -0,0 +1,40 @@
...
+        query = (Contribution.query
+                 .with_parent(self.event)
+                 .filter(Contribution.is_scheduled)
+                 .filter(Contribution.room_reservation_link == None)
+                 .order_by(Contribution.friendly_id))
+
+        result = [{'id': contrib.id, 'friendly_id': contrib.friendly_id, 'title': contrib.title,
+                   'full_title': '#{}: {}'.format(contrib.friendly_id, contrib.title)}
+                  for contrib in query
+                  if contrib.can_access(session.user)]
```
ThiefMaster: not needed, this RH is restricted to event managers who can per definition access and manage everything in there

indico/modules/rb_new/controllers/management/contributions.py
```diff
@@ -0,0 +1,40 @@
...
+from __future__ import unicode_literals
+
+from flask import jsonify, session
+
+from indico.modules.events.contributions import Contribution
+from indico.modules.events.management.controllers import RHManageEventBase
+
+
+class RHListOtherContributions(RHManageEventBase):
+    """AJAX endpoint that lists all contributions in the event (dict representation)."""
```
ThiefMaster: not sure what the "(dict representation)" means here, you return a list ;) (and i think the docstring is more than clear enough without that)

indico/modules/rb_new/controllers/management/events.py
```diff
@@ -0,0 +1,38 @@
...
+from indico.modules.rb_new.views.base import WPEventBookingList
+
+
+class RHEventBookingList(RHRoomBookingEventBase):
+    def _process(self):
+        form = BookingListForm(event=self.event)
+        reservations = self.event.reservations
+        has_unlinked_contribs = (Contribution.query.with_parent(self.event)
+                                 .filter(Contribution.is_scheduled)
+                                 .filter(Contribution.room_reservation_link == None).has_rows())
```
ThiefMaster: wrap before `.has_rows()` and assuming that PEP8/editors complain about the `== None` see if `.is_(None)` works (not sure if SA implements this for relationships); otherwise add a `# noqa` comment at the end of the line

indico/modules/rb_new/controllers/management/events.py
```diff
@@ -0,0 +1,38 @@
...
+        has_unlinked_session_blocks = (SessionBlock.query
+                                       .filter(SessionBlock.session.has(event=self.event))
+                                       .filter(SessionBlock.room_reservation_link == None)).has_rows()
```
ThiefMaster: no need to call `filter()` again, you can pass multiple arguments to it which have the same effect.

and the same things as in the comment above

indico/modules/rb_new/controllers/management/sessions.py
```diff
@@ -0,0 +1,42 @@
...
+from indico.modules.events.management.controllers import RHManageEventBase
+from indico.modules.events.sessions import Session
+from indico.modules.events.sessions.models.blocks import SessionBlock
+
+
+class RHListOtherSessionBlocks(RHManageEventBase):
+    """AJAX endpoint that lists all session blocks in the event (dict representation)."""
+
+    def _process(self):
+
```
ThiefMaster: 🙈

indico/modules/rb_new/controllers/management/sessions.py
```diff
@@ -0,0 +1,42 @@
...
+from flask import jsonify, session
+
+from indico.modules.events.management.controllers import RHManageEventBase
+from indico.modules.events.sessions import Session
+from indico.modules.events.sessions.models.blocks import SessionBlock
+
+
+class RHListOtherSessionBlocks(RHManageEventBase):
+    """AJAX endpoint that lists all session blocks in the event (dict representation)."""
```
ThiefMaster: same as above

indico/modules/rb_new/controllers/management/sessions.py
```diff
@@ -0,0 +1,42 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from flask import jsonify, session
+
+from indico.modules.events.management.controllers import RHManageEventBase
+from indico.modules.events.sessions import Session
```
ThiefMaster: better import it from the model file, it's pure coincidence that `__init__.py` imports it

indico/modules/rb_new/controllers/management/sessions.py
```diff
@@ -0,0 +1,42 @@
...
+
+        query = (SessionBlock.query
+                 .filter(SessionBlock.session.has(event=self.event))
+                 .filter(SessionBlock.room_reservation_link == None)
+                 .join(Session)
+                 .order_by(Session.friendly_id))
+
+        result = [{'id': session_block.id, 'friendly_id': session_block.session.friendly_id,
+                   'title': session_block.full_title,
+                   'full_title': '#{}: {}'.format(session_block.session.friendly_id, session_block.full_title)}
```
ThiefMaster: Even with this, won't you get duplicate entries? If yes let's include the date/time of the block in the title as well

indico/modules/rb_new/controllers/management/sessions.py
```diff
@@ -0,0 +1,42 @@
...
+                  for session_block in query
+                  if session_block.can_access(session.user)]
```
ThiefMaster: not needed

indico/modules/rb_new/fields/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# Indico is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
```
ThiefMaster: leave empty. also, `fields.py` should be more than enough. that way  you're also much more likely to not duplicate lots of code between the two fields ;)

indico/modules/rb_new/fields/contributions.py
```diff
@@ -0,0 +1,61 @@
...
+    def _serialize_contrib(cls, contrib):
+        return {'id': contrib.id, 'friendly_id': contrib.friendly_id, 'title': contrib.title,
+                'full_title': '#{}: {}'.format(contrib.friendly_id, contrib.title)}
+
+    def _value(self):
+        return self._serialize_contrib(self.data) if self.data else None
+
+    def pre_validate(self, form):
+        super(ContributionField, self).pre_validate(form)
+        if self.data is not None and self.data.id in self.excluded_abstract_ids:
```
ThiefMaster: abstract ids?

indico/modules/rb_new/fields/sessions.py
```diff
@@ -0,0 +1,61 @@
...
+        return {'id': contrib.id, 'friendly_id': contrib.friendly_id, 'title': contrib.title,
+                'full_title': '#{}: {}'.format(contrib.friendly_id, contrib.title)}
+
+    def _value(self):
+        return self._serialize_contrib(self.data) if self.data else None
+
+    def pre_validate(self, form):
+        super(SessionBlockField, self).pre_validate(form)
+        if self.data is not None:
+            raise ValueError(_('This contribution cannot be selected.'))
```
ThiefMaster: contribution?

indico/modules/rb_new/forms/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# Indico is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
```
ThiefMaster: leave empty and again a case where we probably don't need a package

ThiefMaster: btw, maybe add `indico/modules/rb_new/event/...` and put controllers.py, forms.py, fields.py, etc. in there? that way all the event stuff is grouped together. basically a submodule of rb_new ;)

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -0,0 +1,72 @@
...
+    </div>
+    {{ reservations }}
+    <script>
+        (function() {
+            'use strict';
+
+            $('#contribution, #session_block').on('change', (e) => {
+                const $target = $(e.currentTarget);
+                $target.closest('.searchable-field').find('.book-btn').toggleClass('disabled', !$target.val());
+                const linkType = {contribution: 'contribution',
```
ThiefMaster: put all in one line or wrap after `{`

indico/modules/rb_new/controllers/management/contributions.py
mvidalgarcia: Copy paste from https://github.com/indico/indico/blob/master/indico/modules/events/abstracts/controllers/reviewing.py#L46 :confused:

ThiefMaster: OK weird. Anyway, then the "other" makes more sense too since for the abstract one it only shows abstracts that are not the one the user currently views

indico/modules/rb_new/controllers/management/events.py
mvidalgarcia: > NotImplementedError: <function is_ at 0x7fe0d7e83ed8>

ThiefMaster: damn thought so... pretty sure i got this error at least once and then had to go for the uglier `# noqa` comment to disable linters..

indico/modules/rb_new/event/controllers.py
```diff
@@ -41,6 +41,8 @@ def _process(self):
         has_unlinked_session_blocks = (SessionBlock.query
                                        .filter(SessionBlock.session.has(event=self.event),
                                                SessionBlock.room_reservation_link == None)).has_rows()  # noqa
+        book_room_url = '{}book?'.format(url_for('rooms_new.roombooking'))
```
ThiefMaster: `url_for('rooms_new.roombooking', path='book', linkType='event', linkId=self.event.id)`

indico/modules/events/management/client/js/index.js
```diff
@@ -17,6 +17,10 @@
 
 /* global setupSearchBox:false, strnatcmp:false */
 
+import 'selectize';
```
ThiefMaster: If we import this in management now, can we remove it from vc? The VC pages using it are all in management so I think we are just duplicating the CSS there.

indico/modules/rb_new/blueprint.py
```diff
@@ -96,3 +97,10 @@
 _bp.add_url_rule('/api/admin/attributes', 'admin_attributes', admin.RHAttributes, methods=('GET', 'POST'))
 _bp.add_url_rule('/api/admin/attributes/<int:attribute_id>', 'admin_attributes', admin.RHAttributes,
                  methods=('GET', 'DELETE', 'PATCH'))
+
+# Event linking
+_bp.add_url_rule('!/event/<confId>/manage/rooms/', 'booking_list', event.RHEventBookingList)
+_bp.add_url_rule('!/event/<confId>/manage/contributions/other-list', 'other_contributions',
```
ThiefMaster: not sure what happened to my previous comment, but:

- this is not "other" here unlike in the abstracts version it was copied from
- since the RHs have room-specific logic, I think they belong inside `/manage/rooms/...`, e.g. `!/event/<confId>/manage/rooms/linking/contributions`

indico/modules/rb_new/event/controllers.py
```diff
@@ -0,0 +1,86 @@
...
+from indico.modules.rb_new.views.base import WPEventBookingList
+from indico.util.date_time import format_datetime
+from indico.web.flask.util import url_for
+
+
+class RHEventBookingList(RHRoomBookingEventBase):
+    def _process(self):
+        form = BookingListForm(event=self.event)
+        reservations = self.event.reservations
+        has_unlinked_contribs = (Contribution.query.with_parent(self.event)
```
ThiefMaster:
```python
        has_unlinked_contribs = (Contribution.query.with_parent(self.event)
                                 .filter(Contribution.is_scheduled,
                                         Contribution.room_reservation_link == None)  # noqa
                                 .has_rows())
```

indico/modules/rb_new/event/controllers.py
```diff
@@ -0,0 +1,86 @@
...
+                                 .filter(Contribution.is_scheduled,
+                                         Contribution.room_reservation_link == None)).has_rows()  # noqa
+        has_unlinked_session_blocks = (SessionBlock.query
+                                       .filter(SessionBlock.session.has(event=self.event),
+                                               SessionBlock.room_reservation_link == None)).has_rows()  # noqa
```
ThiefMaster: same as above

indico/modules/rb_new/event/controllers.py
```diff
@@ -0,0 +1,86 @@
...
+        book_room_url = url_for('rooms_new.roombooking', path='book')
+        event_book_room_url = url_for('rooms_new.roombooking', path='book', link_type='event', link_id=self.event.id)
+
+        def booking_details_url(reservation):
+            return url_for('rooms_new.roombooking', path='calendar', modal='booking-details:{}'.format(reservation.id))
```
ThiefMaster: shouldn't this be a jinja macro in the template?

indico/modules/rb_new/event/controllers.py
```diff
@@ -0,0 +1,86 @@
...
+    def _process(self):
+        form = BookingListForm(event=self.event)
+        reservations = self.event.reservations
+        has_unlinked_contribs = (Contribution.query.with_parent(self.event)
+                                 .filter(Contribution.is_scheduled,
+                                         Contribution.room_reservation_link == None)).has_rows()  # noqa
+        has_unlinked_session_blocks = (SessionBlock.query
+                                       .filter(SessionBlock.session.has(event=self.event),
+                                               SessionBlock.room_reservation_link == None)).has_rows()  # noqa
+        book_room_url = url_for('rooms_new.roombooking', path='book')
```
ThiefMaster: this can be done in the template

indico/modules/rb_new/event/fields.py
```diff
@@ -0,0 +1,79 @@
...
+
+from wtforms.ext.sqlalchemy.fields import QuerySelectField
+
+from indico.util.i18n import _
+from indico.web.flask.util import url_for
+from indico.web.forms.widgets import SelectizeWidget
+
+
+class LinkedObjectField(QuerySelectField):
+
```
ThiefMaster: no empty line here

indico/modules/rb_new/event/fields.py
```diff
@@ -0,0 +1,79 @@
...
+class SessionBlockField(LinkedObjectField):
+    """A selectize-based field to select a session block that has no reservation yet."""
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('render_kw', {}).setdefault('placeholder', _('Enter session block title or #id'))
+        super(SessionBlockField, self).__init__(*args, **kwargs)
+
+    @classmethod
+    def _serialize_linked_object(cls, session_block):
+        return {'id': session_block.id, 'friendly_id': session_block.friendly_id, 'title': session_block.title,
```
ThiefMaster: session blocks don't have a friendly ID and their title is optional -> use its `.full_title` instead.
since you can easily have multiple session blocks with the exact same `full_title` (multiple unnamed blocks in the same session), I think the date/time of it needs to be included as well

indico/modules/rb_new/event/forms.py
```diff
@@ -0,0 +1,35 @@
...
+
+from wtforms.validators import DataRequired
+
+from indico.modules.rb_new.event.fields import ContributionField, SessionBlockField
+from indico.util.i18n import _
+from indico.web.forms.base import IndicoForm
+
+
+class BookingListForm(IndicoForm):
+    contribution = ContributionField(_("Contribution"), [DataRequired()], description=_("Enter the contribution name."),
```
ThiefMaster: iirc we always put the description kwarg last and on a new line (easier to read and less wrapping in case of longer descriptions)

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -0,0 +1,122 @@
+{% extends 'events/management/base.html' %}
+{% from 'forms/_form.html' import form_header, form_row, form_footer %}
```
ThiefMaster: remove unused imports

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -0,0 +1,122 @@
...
+        <div class="action-box event-bookings">
+            <div class="section">
+                <div class="text">
+                    <div class="label">
+                        {%- trans %}Event{% endtrans -%}
+                    </div>
+                    {{ event.title }}
+                </div>
+                <div class="toolbar">
+                    <a href="{{ event_book_room_url }}" target="_blank" class="i-button highlight book-btn">
```
ThiefMaster: `js-book-btn` since it's not used for styling

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -0,0 +1,122 @@
...
+            $('#contribution, #session_block').on('change', (e) => {
+                const $target = $(e.currentTarget);
+                const $bookBtn = $target.closest('.searchable-field').find('.book-btn');
+                const linkType = {
+                    contribution: 'contribution',
+                    session_block: 'sessionBlock',
+                }[$target.attr('id')];
+                const params = {
+                    link_type: linkType,
+                    link_id: +$target.val(),
```
ThiefMaster: since it all becomes a string there's no need to convert to a number here

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -0,0 +1,122 @@
...
+                };
+                $bookBtn.toggleClass('disabled', !$target.val())
```
ThiefMaster: i'd wrap the `.toggleClass` call into a new line as well. looks weird to have multiple chained calls that don't all start in the same column

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -0,0 +1,122 @@
...
+                    .attr('href', $bookBtn.data('href') + (params.link_id ? `?${$.param(params)}`: ''))
```
ThiefMaster: Could we move all this JS into a .js file? That's cleaner anyway and then you could import `indico-url:rooms_new.roombooking` there and build the url like this:

    rbURL({path: 'book', ...params})

ThiefMaster: BTW, when would `params.link_id` be unset?

indico/modules/rb_new/templates/booking_list.html
ThiefMaster: also, i think for event booking this class isn't needed at all

indico/modules/rb/blueprints/events.py
```diff
@@ -24,32 +24,33 @@
 
 
 # Booking and event assignment list
-event_mgmt.add_url_rule('/rooms/', 'rooms_booking_list', event.RHRoomBookingEventBookingList)
```
ThiefMaster: what happened to this one? shouldn't it also be renamed to `rooms-old`?

indico/modules/rb_new/blueprint.py
```diff
@@ -97,3 +98,10 @@
 _bp.add_url_rule('/api/admin/attributes', 'admin_attributes', admin.RHAttributes, methods=('GET', 'POST'))
 _bp.add_url_rule('/api/admin/attributes/<int:attribute_id>', 'admin_attributes', admin.RHAttributes,
                  methods=('GET', 'DELETE', 'PATCH'))
+
+# Event linking
+_bp.add_url_rule('!/event/<confId>/manage/rooms/', 'booking_list', event.RHEventBookingList)
```
ThiefMaster: what about `event_...` for the endpoint names here?

indico/modules/rb_new/event/controllers.py
```diff
@@ -0,0 +1,95 @@
...
+                                         Contribution.room_reservation_links == None)  # noqa
+                                 .has_rows())
+        has_unlinked_session_blocks = (SessionBlock.query
+                                       .filter(SessionBlock.session.has(event=self.event),
+                                               SessionBlock.room_reservation_links == None)  # noqa
+                                       .has_rows())
+
+        is_single_day = self.event.start_dt.date() == self.event.end_dt.date()
+        event_rb_params = {'link_type': 'event', 'link_id': self.event.id,
+                           'recurrence': 'single' if is_single_day else 'daily', 'number': 1, 'interval': 'week',
```
ThiefMaster: maybe one-item-per-line is more readable here?

indico/modules/rb_new/event/controllers.py
```diff
@@ -0,0 +1,95 @@
...
+                           'sd': self.event.start_dt_local.date().isoformat(),
+                           'ed': None if is_single_day else self.event.end_dt_local.date().isoformat(),
+                           'st': format_time(self.event.start_dt_local.time()),
+                           'et': format_time(self.event.end_dt_local.time()),
+                           'text': self.event.room.name if self.event.room else None}
```
ThiefMaster: do we properly filter the None at some point or do we end up searching for `text=None` if the event has no room?

indico/modules/rb_new/event/controllers.py
```diff
@@ -0,0 +1,95 @@
...
+
+        return WPEventBookingList.render_template('booking_list.html', self.event, form=form, reservations=reservations,
```
ThiefMaster: I'd wrap after `self.event,`

indico/modules/rb_new/event/controllers.py
```diff
@@ -0,0 +1,95 @@
...
+        query = (Contribution.query
+                 .with_parent(self.event)
+                 .filter(Contribution.is_scheduled, Contribution.room_reservation_links == None)  # noqa
+                 .order_by(Contribution.friendly_id))
+
+        result = [{'value': json.dumps({'id': contrib.id,
+                                        'start_dt': contrib.start_dt.isoformat(),
+                                        'end_dt': contrib.end_dt.isoformat()}),
+                   'friendly_id': contrib.friendly_id, 'title': contrib.title,
+                   'full_title': '#{}: {}'.format(contrib.friendly_id, contrib.title)}
```
ThiefMaster: `contrib.verbose_title` already does this

indico/modules/rb_new/event/controllers.py
```diff
@@ -0,0 +1,95 @@
...
+                 .filter(SessionBlock.session.has(event=self.event), SessionBlock.room_reservation_links == None)  # noqa
+                 .join(Session)
+                 .order_by(Session.friendly_id))
+
+        result = [{'value': json.dumps({'id': session_block.id,
+                                        'start_dt': session_block.start_dt.isoformat(),
+                                        'end_dt': session_block.end_dt.isoformat()}),
+                   'friendly_id': session_block.session.friendly_id, 'title': session_block.full_title,
+                   'full_title': '#{}: {} ({})'.format(session_block.session.friendly_id, session_block.full_title,
+                                                       format_datetime(session_block.timetable_entry.start_dt))}
```
ThiefMaster: careful, `format_datetime` returns a byestring that needs to be wrapped in `to_unicode()`

indico/modules/rb_new/event/client/js/index.js
```diff
@@ -0,0 +1,49 @@
...
+            session_block: 'sessionBlock',
+        }[$target.attr('id')];
+        if ($target.val()) {
+            const values = JSON.parse($target.val());
+            params = {
+                link_type: linkType,
+                link_id: values.id,
+                recurrence: 'single',
+                number: 1,
+                interval: 'week',
```
ThiefMaster: do we really need to specify number/interval here?

indico/modules/rb_new/event/fields.py
```diff
@@ -0,0 +1,62 @@
...
+    """A selectize-based field to select a contribution that has no reservation yet."""
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('render_kw', {}).setdefault('placeholder', _('Enter contribution title or #id'))
+        super(ContributionField, self).__init__(*args, **kwargs)
+
+
+class SessionBlockField(LinkedObjectField):
+    """A selectize-based field to select a session block that has no reservation yet."""
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('render_kw', {}).setdefault('placeholder', _('Enter session block title or #id'))
```
ThiefMaster: `or #id` doesn't apply here

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -0,0 +1,109 @@
...
+                                <td>
+                                    {{ reservation.start_dt | format_date()}}
+                                    {% if reservation.is_repeating %}
+                                        ({% trans %}recurring{% endtrans %})
+                                    {% endif %}
+                                </td>
+                                <td>{{ reservation.start_dt | format_time() }} - {{ reservation.end_dt | format_time() }}</td>
+                                {% set link = reservation.link %}
+                                <td class="link-data">
+                                <span title="{{ link.object.full_title if link.link_type.name == 'session_block' else link.object.title }}">
```
ThiefMaster: missing indentation

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -0,0 +1,109 @@
...
+                                    {{ link.link_type.name.replace('_', ' ') | capitalize }}
```
ThiefMaster: Please use a dict with translated names for this...

ThiefMaster: Or better, replace the `LinkType` enum with this:

```python
class LinkType(RichIntEnum):
    __title__ = (None, _('Category'), _('Event'), ...)

    category = 1
    event = 2
    contribution = 3
    subcontribution = 4
    session = 5
    session_block = 6
```

Then you can just use `link.link_type.title`

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -0,0 +1,109 @@
...
+{% endblock %}
+
+{% macro booking_details_link(reservation) %}
+    <a href="{{ url_for('rooms_new.roombooking', path='calendar', modal='booking-details:{}'.format(reservation.id)) }}"
+       target="_blank"
+       class="icon-arrow-right-sparse action-icon"
+       title="{% trans %}See details{% endtrans %}"></a>
+{% endmacro %}
+
+
```
ThiefMaster: too many empty lines are the end

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -0,0 +1,109 @@
+{% extends 'events/management/base.html' %}
+{% from 'forms/_form.html' import form_row %}
+
+{% block title %}
+    {% trans %}Event bookings{% endtrans %}
```
ThiefMaster: "Rooms" or "Room bookings"?

indico/modules/rb_new/event/client/js/index.js
mvidalgarcia: I doesn't work without them, the search works fine but when you book it fails.

indico/modules/rb_new/event/controllers.py
mvidalgarcia: `url_for` is filtering them

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -70,8 +70,18 @@ <h3>Room bookings</h3>
...
                         {% for reservation in reservations %}
-                            <tr>
-                                <td>{{ reservation.room.full_name }}</td>
+                            <tr class="{{ 'disabled' if reservation.is_rejected }}">
+                                <td>
+                                    {{ reservation.room.full_name }}
+                                    {% if reservation.is_rejected %}
+                                        <span class="icon-disable booking-status"
+                                              title="{% trans %}Rejected:{% endtrans %}
+                                                     {{ reservation.rejection_reason }}"></span>
```
ThiefMaster: am i the only one who finds wrapping inside an attribute quite ugly?

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -17,7 +17,7 @@
                 </div>
                 <div class="toolbar">
                     <a href="{{ url_for('rooms_new.roombooking', path='book', **event_rb_params) }}"
-                       target="_blank" class="i-button highlight">
+                       target="_blank" class="i-button highlight {{ 'disabled' if is_past_event }}">
```
ThiefMaster: Ideally that's done for contributions/blocks as well, but maybe checking for the whole event is enough... In that case I'd also disable the session/contribution buttons though...

mvidalgarcia: Just the buttons or hide the fields altogether?

ThiefMaster: Disabling with a tooltip is better I guess. But now that you mention hiding.. we could just filter out contributions/blocks that started in the past!

mvidalgarcia:
Updated :arrows_clockwise:

indico/modules/rb_new/templates/booking_list.html
```diff
@@ -0,0 +1,118 @@
...
+                    {% set title = _('Event has already finished') %}
+                    <a href="{{ url_for('rooms_new.roombooking', path='book', **event_rb_params) }}"
+                       target="_blank" class="i-button highlight {{ 'disabled' if is_past_event }}"
+                       title="{{ title if is_past_event }}">
+                        {%- trans %}Book{% endtrans -%}
+                    </a>
+                </div>
+            </div>
+            {% if has_contribs %}
+                <div class="section searchable-field" data-values={{ contribs_data | tojson }}>
```
ThiefMaster: :warning: `data-values="{{ contribs_data | tojson | forceescape }}"`

indico/modules/rb_new/event/fields.py
```diff
@@ -0,0 +1,62 @@
...
+from __future__ import unicode_literals
+
+from wtforms.ext.sqlalchemy.fields import QuerySelectField
+
+from indico.util.i18n import _
+from indico.web.flask.util import url_for
+from indico.web.forms.widgets import SelectizeWidget
+
+
+class LinkedObjectField(QuerySelectField):
```
ThiefMaster: QSF doesn't make much sense here - since we don't use the form for submission even a simple `Field` should do the job for rendering

PR #3745, opened 2019-02-01
Consider booking_limit_days in search_for_rooms

### LINKED ISSUE #3567: Expose booking_limit_days validation on client-side
Whenever the limit is being surpassed...
Ideas:
1. Hiding the room at all from search results
2. Showing a warning icon directly in the results with a explanatory tooltip.
3. Showing a warning/error in the booking dialog with a explanation and disabling the booking button.

I'm mostly inclined to the 2nd + 3rd.

panagiotappl:
closes #3567

Solution based on this: 
https://github.com/indico/indico/issues/3567#issuecomment-459345824

### DIFF
--- a/indico/modules/rb_new/client/js/common/roomSearch/actions.js
+++ b/indico/modules/rb_new/client/js/common/roomSearch/actions.js
@@ -32,13 +32,14 @@ export function roomSearchActionsFactory(namespace) {
 
     function searchRooms() {
         return async (dispatch, getStore) => {
-            const {[namespace]: {filters}} = getStore();
+            const store = getStore();
+            const {filters} = store[namespace];
             if (!validateFilters(filters, namespace, dispatch)) {
                 return;
             }
-            const isAdmin = userSelectors.isUserAdminOverrideEnabled(getStore());
+            const isAdminOverrideEnabled = userSelectors.isUserAdminOverrideEnabled(store);
             const params = preProcessParameters(filters, ajaxFilterRules);
-            params.is_admin = isAdmin;
+            params.admin_override_enabled = isAdminOverrideEnabled;
             return await ajaxAction(
                 () => indicoAxios.get(searchRoomsURL(), {params}),
                 SEARCH_ROOMS_REQUEST,
--- a/indico/modules/rb_new/client/js/common/roomSearch/actions.js
+++ b/indico/modules/rb_new/client/js/common/roomSearch/actions.js
@@ -21,6 +21,7 @@ import {ajaxAction} from 'indico/utils/redux';
 import {preProcessParameters} from '../../util';
 import {ajax as ajaxFilterRules} from './serializers';
 import {validateFilters} from '../filters';
+import {selectors as userSelectors} from '../../common/user';
 
 
 export function roomSearchActionsFactory(namespace) {
@@ -35,8 +36,9 @@ export function roomSearchActionsFactory(namespace) {
             if (!validateFilters(filters, namespace, dispatch)) {
                 return;
             }
+            const isAdmin = userSelectors.isUserAdminOverrideEnabled(getStore());
             const params = preProcessParameters(filters, ajaxFilterRules);
-
+            params.is_admin = isAdmin;
             return await ajaxAction(
                 () => indicoAxios.get(searchRoomsURL(), {params}),
                 SEARCH_ROOMS_REQUEST,
--- a/indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
+++ b/indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
@@ -67,7 +67,8 @@ export default class BookingSuggestion extends React.PureComponent {
         suggestions: PropTypes.shape({
             time: PropTypes.number,
             duration: PropTypes.number,
-            skip: PropTypes.number
+            skip: PropTypes.number,
+            shorten: PropTypes.number,
         }).isRequired,
         /** Will be called when a suggestion is chosen (clicked on) */
         onClick: PropTypes.func.isRequired
@@ -95,7 +96,7 @@ export default class BookingSuggestion extends React.PureComponent {
         ));
     }
 
-    renderSuggestionText(room, {time, duration, skip}) {
+    renderSuggestionText(room, {time, duration, skip, shorten}) {
         const {onClick} = this.props;
         return (
             <>
@@ -124,6 +125,13 @@ export default class BookingSuggestion extends React.PureComponent {
                                       text={PluralTranslate.string('Skip one day', 'Skip {skip} days', skip, {skip})}
                                       tooltip={PluralTranslate.string("We'll skip one occurrence to avoid any conflicts",
                                                                       "We'll skip {skip} occurrences to avoid any conflicts", skip, {skip})} />
+                )}
+                {shorten && (
+                    <SuggestionOption onClick={onClick}
+                                      overrides={{shorten}}
+                                      icon="calendar times"
+                                      text={PluralTranslate.string('Shorten by one day', 'Shorten by {shorten} days', shorten, {shorten})}
+                                      tooltip={Translate.string('Bookings in this room cannot last this long and will be shortened accordingly')} />
                 )}
             </>
         );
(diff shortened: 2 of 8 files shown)


indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
```diff
@@ -125,6 +126,14 @@ export default class BookingSuggestion extends React.PureComponent {
                                       tooltip={PluralTranslate.string("We'll skip one occurrence to avoid any conflicts",
                                                                       "We'll skip {skip} occurrences to avoid any conflicts", skip, {skip})} />
                 )}
+                {shorten && (
+                    <SuggestionOption onClick={onClick}
+                                      overrides={{shorten}}
```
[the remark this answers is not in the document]
ThiefMaster: i think that's intentional somce the override is e.g. `{shorten: 30}` to shorten by 30 days

indico/modules/rb_new/client/js/common/roomSearch/actions.js
```diff
@@ -35,8 +36,9 @@ export function roomSearchActionsFactory(namespace) {
             if (!validateFilters(filters, namespace, dispatch)) {
                 return;
             }
+            const isAdminOverrideEnabled = userSelectors.isUserAdminOverrideEnabled(getStore());
```
ThiefMaster: I'd call `getStore()` only once, ie something like this:
```
const store = getStore();
const {filters} = store[namespace];
const isAdminOverrideEnabled = userSelectors.isUserAdminOverrideEnabled(store);
```

indico/modules/rb_new/operations/rooms.py
```diff
@@ -137,10 +137,11 @@ def get_room_statistics(room):
     return data
 
 
-def search_for_rooms(filters, is_admin_override, availability=None):
+def search_for_rooms(filters, admin_override_enabled, availability=None):
     """Search for a room, using the provided filters.
 
     :param filters: The filters, provided as a dictionary
+    :param admin_override_enabled: A boolean specifying whether admins have override privilages
```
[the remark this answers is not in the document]
ThiefMaster: I'd call the param `allow_admin` like we use everywhere else in the backend (e.g. in can_*) and give it a default value

indico/modules/rb_new/controllers/backend/rooms.py
```diff
@@ -62,22 +62,26 @@ def _process(self):
...
+        admin_override_enabled = args.pop('admin_override_enabled')
         if not filter_availability:
             availability = None
             if only_unavailable:
                 raise UnprocessableEntity('Required data to filter by availability is not present')
         else:
             availability = not only_unavailable
 
-        search_query = search_for_rooms(args, availability=availability)
+        search_query = search_for_rooms(args, admin_override_enabled, availability=availability)
```
ThiefMaster: i'd use kwarg syntax for it

indico/modules/rb_new/operations/suggestions.py
```diff
@@ -100,10 +102,17 @@ def get_number_of_skipped_days_for_rooms(rooms, start_dt, end_dt, repeat_frequen
         if limit and len(data) == limit:
             break
 
+        suggestions = {}
+        if room.booking_limit_days is not None and room.booking_limit_days < booking_length:
```
ThiefMaster: you need to take the global `booking_limit` into account if the room doesn't have one set:

booking_limit = room.booking_limit_days or rb_settings.get('booking_limit')

indico/modules/rb_new/operations/rooms.py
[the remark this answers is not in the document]
panagiotappl: Should the default value be False?

indico/modules/rb_new/controllers/backend/rooms.py
```diff
@@ -62,22 +62,26 @@ def _process(self):
 
 class RHSearchRooms(RHRoomBookingBase):
     @use_args(dict(search_room_args, **{
-        'unavailable': fields.Bool(missing=False)
+        'unavailable': fields.Bool(missing=False),
+        'admin_override_enabled': fields.Bool()
```
mic4ael: is it always in the request or should we be on the safe side and add `missing` to the `fields.Bool`?

indico/modules/rb_new/operations/rooms.py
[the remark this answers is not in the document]
ThiefMaster: That's a good question... generally we have `allow_admin=True` as the default for all access checks, since we usually don't have opt-in admin privs. But for this one I think `False` makes more sense (and it's not a typical access check anyway).

indico/modules/rb_new/controllers/backend/rooms.py
ThiefMaster: missing defaults to `Missing`, which is falsy

indico/modules/rb_new/operations/rooms.py
```diff
@@ -191,21 +192,21 @@ def search_for_rooms(filters, availability=None):
...
-                           db.and_(Room.filter_bookable_hours(start_dt.time(), end_dt.time()),
-                                   Room.filter_nonbookable_periods(start_dt, end_dt),
-                                   db.or_(booking_limit_days.is_(None),
-                                          selected_period_days <= booking_limit_days)))
-        availability_filters.append(criterion)
+        criterion = db.and_(Room.filter_bookable_hours(start_dt.time(), end_dt.time()),
+                            Room.filter_nonbookable_periods(start_dt, end_dt),
+                            db.or_(booking_limit_days.is_(None),
+                            selected_period_days <= booking_limit_days))
+        cant_book = [room.id for room in query.filter(db.and_(*availability_filters)).filter(~criterion).all() if
```
mic4ael: That looks a bit unreadable, what about sth like this?

```python

cant_book = [room.id
             for room in query.filter(db.and_(*availability_filters)).filter(~criterion).all()
             if not room.can_override(session.user, allow_admin=False)]
```

ThiefMaster: :+1: wrapping after `if` doesn't look amazing

Also, the querying part can be simplified to `query.filter(db.and_(*availability_filters), ~criterion)`

indico/modules/rb_new/controllers/backend/rooms.py
panagiotappl: Again here, if there was a missing option that would be False, correct?

ThiefMaster: it'd be a falsy object. if you specify `missing=False` it will be actually `False`

indico/modules/rb_new/client/js/modules/bookRoom/BookingSuggestion.jsx
```diff
@@ -125,6 +126,14 @@ export default class BookingSuggestion extends React.PureComponent {
...
+                                      icon="calendar times"
+                                      text={PluralTranslate.string('Make booking non-recurring', 'Shorten by {shorten} days', shorten, {shorten})}
```
ThiefMaster: Are you sure this is correct? If you shorten BY 1 occurrence it doesn't make it non-recurring. Also, you're shortening by days - unless it's a daily booking those counts are different.

panagiotappl: You are right. This is not what should be applied here at all. :eyes: It should be something like "Shorten by one day"

indico/modules/rb_new/operations/rooms.py
```diff
@@ -191,21 +192,22 @@ def search_for_rooms(filters, availability=None):
...
-                           db.and_(Room.filter_bookable_hours(start_dt.time(), end_dt.time()),
-                                   Room.filter_nonbookable_periods(start_dt, end_dt),
-                                   db.or_(booking_limit_days.is_(None),
-                                          selected_period_days <= booking_limit_days)))
-        availability_filters.append(criterion)
+        criterion = db.and_(Room.filter_bookable_hours(start_dt.time(), end_dt.time()),
+                            Room.filter_nonbookable_periods(start_dt, end_dt),
+                            db.or_(booking_limit_days.is_(None),
+                            selected_period_days <= booking_limit_days))
+        cant_book = [room.id
```
ThiefMaster: `unbookable_ids`?

indico/modules/rb_new/operations/rooms.py
```diff
@@ -191,21 +192,22 @@ def search_for_rooms(filters, availability=None):
...
+                     for room in query.filter(db.and_(*availability_filters), ~criterion)
+                     if not room.can_override(session.user, allow_admin=False)]
+        availability_filters.append(db.not_(Room.id.in_(cant_book)))
```
ThiefMaster: I't use one of these two versions

```python
availability_filters.append(~Room.id.in_(cant_book))
availability_filters.append(Room.id.notin_(cant_book))
```

PR #3747, opened 2019-02-05
Add more persistent way of linking stuff to bookings

ThiefMaster:
Opening RB with `linkType=...&linkId=...` sets it in the state and keeps it there until the user leaves linking mode or creates a booking.

### DIFF
--- a/indico/modules/rb_new/client/js/common/linking/LinkBar.jsx
+++ b/indico/modules/rb_new/client/js/common/linking/LinkBar.jsx
@@ -23,6 +23,7 @@ import {Icon, Popup} from 'semantic-ui-react';
 import {Translate} from 'indico/react/i18n';
 import * as linkingActions from './actions';
 import * as linkingSelectors from './selectors';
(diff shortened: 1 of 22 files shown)


indico/modules/rb_new/client/js/common/linking/LinkBar.jsx
```diff
@@ -0,0 +1,87 @@
...
+ * an event, contribution or session block.
+ */
+const LinkBar = ({visible, clear, data}) => {
+    if (!visible) {
+        return null;
+    }
+
+    const {type, title, eventURL, eventTitle} = data;
+    return (
+        <header styleName="link-bar">
```
mic4ael: Intentionally `header` instead of SUI's `<Header />`?

ThiefMaster: I used the same element already used for the rb-menu-bar. I think SUI Header comes with extra styling.

PR #3760, opened 2019-02-11
Add enlarged photo view option to room details

### LINKED ISSUE #3754: Add some way to view the full-size room photo

panagiotappl:
closes #3754

### DIFF
--- a/indico/modules/rb_new/blueprint.py
+++ b/indico/modules/rb_new/blueprint.py
@@ -32,6 +32,7 @@ _bp.add_url_rule('/', 'roombooking', RHRoomBooking)
 _bp.add_url_rule('/<path:path>', 'roombooking', RHRoomBooking)
 _bp.add_url_rule('/rooms-sprite-<version>.jpg', 'sprite', misc.RHRoomsSprite)
 _bp.add_url_rule('/rooms-sprite.jpg', 'sprite', misc.RHRoomsSprite)
+_bp.add_url_rule('/rooms/<int:room_id>.jpg', 'room_photo', rooms.RHRoomPhoto)
 
 # General backend
 _bp.add_url_rule('/api/<path:path>', '404', lambda path: (jsonify(), 404))
--- a/indico/modules/rb_new/client/js/components/RoomBasicDetails.jsx
+++ b/indico/modules/rb_new/client/js/components/RoomBasicDetails.jsx
@@ -15,9 +15,11 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
+import roomImageURL from 'indico-url:rooms_new.room_photo';
+
 import React from 'react';
 import PropTypes from 'prop-types';
-import {Grid, Header, Icon, Popup} from 'semantic-ui-react';
+import {Dimmer, Grid, Header, Icon, Image, Loader, Modal, Popup} from 'semantic-ui-react';
 import {Param, Plural, PluralTranslate, Singular, Translate} from 'indico/react/i18n';
 import {formatLatLon} from '../common/map/util';
 
@@ -67,8 +69,58 @@ AnnotatedIcon.propTypes = {
 
 export default class RoomBasicDetails extends React.PureComponent {
     static propTypes = {
-        room: PropTypes.object.isRequired
+        room: PropTypes.object.isRequired,
     };
+
+    state = {
+        imageModalOpen: false,
+        imageLoaded: false,
+    };
+
+    renderImage() {
+        const {room} = this.props;
+        const {imageModalOpen, imageLoaded} = this.state;
+        const isClickable = room.hasPhoto;
+        const spriteImage = (
+            <SpriteImage styleName="clickable"
+                         pos={room.spritePosition}
+                         height="100%"
+                         width="100%"
+                         fillVertical
+                         onClick={isClickable ? (() => this.setState({imageModalOpen: true})) : undefined} />
+        );
+        if (isClickable) {
+            return (
+                <>
+                    <Modal closeIcon open={imageModalOpen}
+                           onClose={() => this.setState({imageModalOpen: false, imageLoaded: false})}>
+                        <Modal.Content image styleName="photo-dimmer">
+                            <Dimmer.Dimmable styleName="photo-dimmable-width">
+                                <Dimmer active={!imageLoaded} inverted>
+                                    <Loader />
+                                </Dimmer>
+                                <Image styleName="photo-width"
+                                       wrapped
+                                       onLoad={() => this.setState({imageLoaded: true})}
+                                       src={roomImageURL({room_id: room.id})} />
+                            </Dimmer.Dimmable>
+                        </Modal.Content>
+                    </Modal>
+                    <div styleName="image-container clickable">
+                        {spriteImage}
+                        <RoomFeaturesBox room={room} />
+                    </div>
+                </>
+            );
+        } else {
+            return (
+                <div styleName="image-container">
+                    {spriteImage}
+                    <RoomFeaturesBox room={room} />
+                </div>
+            );
+        }
+    }
 
     render() {
         const {room} = this.props;
@@ -76,17 +128,10 @@ export default class RoomBasicDetails extends React.PureComponent {
             ownerName: owner, latitude, longitude, division, locationName: location, surfaceArea: surface, capacity,
             telephone, fullName: name
         } = room;
-
         return (
             <Grid columns={2}>
                 <Grid.Column textAlign="center" styleName="photo-column">
-                    <div styleName="image-container">
-                        <SpriteImage pos={room.spritePosition}
-                                     height="100%"
-                                     width="100%"
-                                     fillVertical />
-                        <RoomFeaturesBox room={room} />
-                    </div>
+                    {this.renderImage()}
                 </Grid.Column>
                 <Grid.Column styleName="data-column">
                     <Header>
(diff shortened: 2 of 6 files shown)


indico/modules/rb/models/rooms.py
```diff
@@ -299,6 +299,10 @@ def is_auto_confirm(self):
     def is_auto_confirm(self):
         return ~self.reservations_need_confirmation
 
+    @property
```
mvidalgarcia: This property is already defined, look a bit below :wink:

indico/modules/rb_new/blueprint.py
```diff
@@ -31,6 +31,7 @@
 _bp.add_url_rule('/<path:path>', 'roombooking', RHRoomBooking)
 _bp.add_url_rule('/rooms-sprite-<version>.jpg', 'sprite', misc.RHRoomsSprite)
 _bp.add_url_rule('/rooms-sprite.jpg', 'sprite', misc.RHRoomsSprite)
+_bp.add_url_rule('/rooms-image/<int:room_id>/room.jpg', 'rooms_image', rooms.RHRoomsImage)
```
mvidalgarcia: What about `/rooms/<int:room_id>/photo.jpg` and then the RH in singular `RHRoomPhoto`? I don't have a strong opinion on Image vs Photo.
Also, this might fit better in # Rooms

indico/modules/rb_new/client/js/components/RoomBasicDetails.jsx
```diff
@@ -80,11 +93,32 @@ export default class RoomBasicDetails extends React.PureComponent {
...
+                                <Dimmer active={!imageLoaded} inverted>
+                                    <Loader />
+                                </Dimmer>
+                                <Image wrapped
+                                       onLoad={() => this.setState({imageLoaded: true})}
+                                       src={fetchRoomImageURL({room_id: room.id})} />
+                            </Dimmer.Dimmable>
+                        </Modal.Content>
+                    </Modal>
+                    <div styleName={isClickable ? 'image-container clickable-image' : 'image-container'}>
```
mvidalgarcia:
```jsx
<div styleName=`image-container {isClickable ? 'clickable-image' : ''}`>
```

indico/modules/rb_new/client/js/components/RoomBasicDetails.jsx
```diff
@@ -80,11 +93,32 @@ export default class RoomBasicDetails extends React.PureComponent {
...
+                        <SpriteImage styleName="clickable-image"
+                                     pos={room.spritePosition}
                                      height="100%"
                                      width="100%"
-                                     fillVertical />
+                                     fillVertical
+                                     onClick={() => {
```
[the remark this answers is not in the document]
ThiefMaster: This could also be split into a `renderImage` method. That way `render` doesn't become so huge and if you define the `<SpriteImage>` at the beginning, you can return it wrapped in the div straight away if the image isn't clickable. No need to always have the modal stuff around.

indico/modules/rb_new/client/js/components/RoomBasicDetails.jsx
ThiefMaster: Is that valid JSX?

ThiefMaster: Also, does this work with the styleName magic which gets resolved at compile time?

indico/modules/rb_new/client/js/components/RoomBasicDetails.module.scss
```diff
@@ -17,6 +17,15 @@
 
 @import 'rb_new:styles/palette';
 
+.image-dimmer {
+    display: flex;
+    justify-content: center;
+}
+
+.clickable-image {
```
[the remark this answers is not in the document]
panagiotappl: I wanted to move it there but there are cases where I don't want the image to be clickable (when the room doesn't have an image). How does it work that way?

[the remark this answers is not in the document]
ThiefMaster: You can still conditionally set the class... I meant this:

```scss
.photo-column {
    overflow: hidden;

    .image-container {
        border: 1px solid $pastel-gray;
        position: relative;
        height: 100%;
        width: 100%;

        &.clickable {
            cursor: pointer !important;
        }
    }
}
```

indico/modules/rb_new/client/js/components/RoomBasicDetails.jsx
mvidalgarcia: Forgot the `$` and brackets.
I'd swear we have this structure in other places

mvidalgarcia: For instance - https://github.com/indico/indico/blob/master/indico/modules/rb_new/client/js/common/timeline/TimelineLegend.jsx#L33

ThiefMaster: Not sure if css module name mangling is applied to the conditional class name...

mic4ael: why not use `className={toClasses{...`?

ThiefMaster: className/toClasses don't support css modules

indico/modules/rb_new/client/js/components/RoomBasicDetails.module.scss
```diff
@@ -17,6 +17,15 @@
...
+    cursor: pointer !important;
```
ThiefMaster: let's try the `zoom-in` cursor instead

indico/modules/rb_new/blueprint.py
pferreir: "Photo" seems a bit better. "Image" is more vague.

indico/modules/rb_new/blueprint.py
```diff
@@ -55,6 +56,7 @@
 _bp.add_url_rule('/api/rooms/<int:room_id>/availability/simple', 'check_room_available', rooms.RHCheckRoomAvailable)
 _bp.add_url_rule('/api/suggestions', 'suggestions', bookings.RHRoomSuggestions)
 
+
```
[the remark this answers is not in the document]
pferreir: I guess it was accidental.

indico/modules/rb_new/client/js/components/RoomBasicDetails.jsx
pferreir: I believe that what @mvidalgarcia suggested would work. But it's a matter of trying it out 😉

panagiotappl: I followed another route in the end that didn't require me to try that. (Having a different render method and not returning the Modal in cases of unclickable like it's  been mentioned in other comments)

indico/modules/rb_new/blueprint.py
panagiotappl: I have no strong opinion on this. I could go with "photo".

indico/modules/rb_new/client/js/components/RoomBasicDetails.jsx
```diff
@@ -67,26 +69,75 @@ AnnotatedIcon.propTypes = {
...
+        const {imageModalOpen, imageLoaded} = this.state;
+        const isClickable = room.hasPhoto;
+        const spriteImage = (
+            <SpriteImage styleName="clickable"
+                         pos={room.spritePosition}
+                         height="100%"
+                         width="100%"
+                         fillVertical
+                         onClick={() => {
+                             if (isClickable) {
```
mvidalgarcia: What about returning earlier and saving the `else` part?
```jsx
if (!isClickable) {
    return null;
}
this.setState({imageModalOpen: true});
```

indico/modules/rb_new/controllers/backend/rooms.py
```diff
@@ -17,11 +17,12 @@
 from __future__ import unicode_literals
 
 from datetime import date, datetime, time, timedelta
+from io import BytesIO
 
 from flask import jsonify, request, session
 from sqlalchemy.orm import subqueryload
 from webargs import fields
-from webargs.flaskparser import use_args, use_kwargs
+from webargs.flaskparser import abort, use_args, use_kwargs
```
mvidalgarcia: You're not using it anymore!

indico/modules/rb_new/client/js/components/SpriteImage.jsx
```diff
@@ -42,14 +42,16 @@ class SpriteImage extends React.Component {
         /** Additional styles to be applied on the component */
         styles: PropTypes.object,
         /** Whether the image should try to fit vertically instead of horizontally */
-        fillVertical: PropTypes.bool
+        fillVertical: PropTypes.bool,
+        onClick: PropTypes.func
```
mvidalgarcia: JSDoc? I know it's obvious but just to be consistent with the rest of props.

indico/modules/rb_new/controllers/backend/rooms.py
ThiefMaster: unrelated, but `abort` is from flask and if used it should be imported from there and not from some library that also imports it (probably also from flask)

indico/modules/rb_new/client/js/components/RoomBasicDetails.jsx
panagiotappl: I forgot to follow the earlier suggestion that was this: 
onClick={isClickable ? (() => this.setState({openImageModal: true})) : undefined}
Thanks for pointing it out and reminding me about it anyway!

mvidalgarcia: True, even better

PR #3778, opened 2019-02-15
Burotel override adjustments

mvidalgarcia:
Related to https://github.com/indico/indico-plugins-cern/pull/27

### DIFF
--- a/indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
+++ b/indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
@@ -28,6 +28,7 @@ import PrincipalSearchField from 'indico/react/components/PrincipalSearchField';
 import {ReduxFormField, ReduxRadioField, formatters, validators as v} from 'indico/react/forms';
 import {SingleDatePicker, DatePeriodField} from 'indico/react/components';
 import {serializeDate, serializeTime, toMoment} from 'indico/utils/date';
+import {Overridable} from 'indico/react/util';
 import {Translate} from 'indico/react/i18n';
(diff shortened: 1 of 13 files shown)


indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -43,10 +44,12 @@ class BookingEditForm extends React.Component {
         booking: PropTypes.object.isRequired,
         formProps: PropTypes.object.isRequired,
         onBookingPeriodChange: PropTypes.func,
+        onlyDaily: PropTypes.bool,
     };
 
     static defaultProps = {
         onBookingPeriodChange: () => {},
+        onlyDaily: false,
```
ThiefMaster: what about using a list of disabled recurrence types here?

indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -43,10 +44,14 @@ class BookingEditForm extends React.Component {
         booking: PropTypes.object.isRequired,
         formProps: PropTypes.object.isRequired,
         onBookingPeriodChange: PropTypes.func,
+        hideOptions: PropTypes.objectOf(PropTypes.bool),
+        optionalFields: PropTypes.objectOf(PropTypes.bool),
```
ThiefMaster: I'd use a simple `reasonRequired` bool (default true) here. I think `reason` is the only field where making it optional makes any sense or even works (at least if we ever make `booked_for_id` NOT NULL as it should be)

PR #3784, opened 2019-02-19
Disable pointer cursor when not clickable

mvidalgarcia:


indico/modules/rb_new/client/js/common/timeline/TimelineItem.module.scss
```diff
@@ -145,3 +150,31 @@
...
+:global(.book-room-timeline-view) {
+    .timeline-occurrence {
+        &:global(.booking),
+        &:global(.pre-booking) {
+            cursor: default;
+        }
+    }
+}
+
+:global(.unavailable-rooms-modal),
```
ThiefMaster: is there a better way than using global class names? e.g. checking which types of click handlers are actually defined and then using that info to determine which bars are clickable

indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
```diff
@@ -51,6 +51,7 @@ export default class DailyTimelineContent extends React.Component {
         renderHeader: PropTypes.func,
         showActions: PropTypes.bool,
         booking: PropTypes.object,
+        classes: PropTypes.string,
```
ThiefMaster: `className`? or does it behave differently from what `className` usually does?

indico/modules/rb_new/client/js/common/timeline/TimelineItem.module.scss
mvidalgarcia: Don't know what I was thinking of :man_facepalming: Thanks!

PR #3791, opened 2019-02-20
Startup assistance

mic4ael:


indico/modules/rb_new/schemas.py
```diff
@@ -70,7 +70,7 @@ class Meta:
         fields = ('id', 'name', 'capacity', 'building', 'floor', 'number', 'is_public', 'location_name', 'full_name',
                   'comments', 'division', 'is_reservable', 'reservations_need_confirmation', 'sprite_position',
                   'surface_area', 'latitude', 'longitude', 'telephone', 'key_location', 'max_advance_days',
-                  'owner_name', 'available_equipment', 'has_photo')
+                  'owner_name', 'available_equipment', 'has_photo', 'notification_for_assistance')
```
ThiefMaster: this doesn't belong in the core

indico/web/client/js/react/forms/fields.jsx
```diff
@@ -117,7 +117,7 @@ export function ReduxCheckboxField({input: {value, ...input}, ...props}) {
     return (
         <ReduxFormField input={input}
                         {...props}
-                        checked={value}
+                        checked={!!value}
```
ThiefMaster: :eyes:  (have to check if there's a better way to get `false` here even if there's no way to put that value in the form's initialValues)

indico/modules/rb/models/reservation_occurrences.py
```diff
@@ -232,7 +232,9 @@ def find_with_filters(cls, filters, user=None):
         if filters.get('needs_vc_assistance'):
             q = q.filter(Reservation.needs_vc_assistance)
         if filters.get('needs_assistance'):
-            q = q.filter(Reservation.needs_assistance)
+            # TODO: properly filter by those reservations that have an associated request
+            # q = q.filter(Reservation.needs_assistance)
+            pass
```
ThiefMaster: don't bother changing this function, it'll go away with old-rb eventually

indico/modules/rb_new/schemas.py
mic4ael: So I think I should have a separate table in the plugin with a single column storing the ids of the rooms that allow to request assistance, right?

ThiefMaster: I'd just use a plugin setting storing a list of room ids

ThiefMaster: See `RoomConverter` in the cronjobs plugin

mic4ael: Event better, thanks!

indico/modules/rb_new/client/js/modules/bookRoom/BookRoomModal.jsx
```diff
@@ -553,6 +554,10 @@ class BookRoomModal extends React.Component {
                                            placeholder={Translate.string('Reason for booking')}
                                            disabled={bookingBlocked(fprops)}
                                            required />
+                                    <Overridable id="BookRoomModal.extraFields"
+                                                 disabled={bookingBlocked(fprops)}
+                                                 form={fprops.form}
```
mic4ael: This is a workaround until we start sharing the same `react` bundle between Indico and the plugins.

indico/modules/rb_new/client/js/setup.jsx
```diff
@@ -38,6 +38,10 @@ export default function setup(overrides = {}, postReducers = []) {
         const appContainer = document.getElementById('rb-app-container');
         const store = createRBStore(overrides, postReducers);
 
+        window.registerOverride = (externalOverrides) => {
```
pferreir: Documenting this function could be useful.

indico/core/signals/rb.py
```diff
@@ -25,3 +25,22 @@
 rooms_fetched = _signals.signal('rooms_fetched', """
```
ThiefMaster: can you remove this signal and the place where we're calling it since it's not needed anymore?

indico/migrations/versions/20190301_1027_0cf535c6311a_remove_needs_assistance_column.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def upgrade():
+    op.drop_column('reservations', 'needs_assistance', schema='roombooking')
+    op.drop_column('rooms', 'notification_for_assistance', schema='roombooking')
+
+
+def downgrade():
+    op.add_column('reservations',
+                  sa.Column('needs_assistance', sa.Boolean(), nullable=False, server_default='false'),
```
ThiefMaster: remove the `server_default` afterwards

indico/modules/events/management/templates/settings.html
```diff
@@ -4,7 +4,30 @@
 {% block title %}{% trans %}Settings{% endtrans %}{% endblock %}
 
 {% block content %}
+    {{ template_hook('event-actions', event=event) }}
+    {%- if event.room and event.room_reservation_links|length == 0 -%}
```
ThiefMaster: `not event.room_reservation_links`

indico/modules/events/management/templates/settings.html
```diff
@@ -4,7 +4,30 @@
...
+        <div class="action-box warning">
+            <div class="section">
+                <div class="icon icon-bookmark"></div>
+                <div class="text">
+                    <div class="label">{% trans %}Book the event location{% endtrans %}</div>
+                    {% trans %}
+                        There is no booking linked to this event. Click the button on the right side to
```
ThiefMaster: isn't "Click the button on the right side to book the location of this event." obvious when there's a "book the room" button next to it?

indico/modules/rb/models/reservation_occurrences.py
```diff
@@ -180,69 +180,6 @@ def find_overlapping_with(cls, room, occurrences, skip_reservation_id=None):
...
-            else:
-                q = q.filter(~Reservation.is_cancelled & ~ReservationOccurrence.is_cancelled)
-
-        if filters.get('is_archived'):
-            q = q.filter(Reservation.is_archived)
-
-        if filters.get('uses_vc'):
-            q = q.filter(Reservation.uses_vc)
-        if filters.get('needs_vc_assistance'):
-            q = q.filter(Reservation.needs_vc_assistance)
```
ThiefMaster: i would have just removed this part since some of the old rb code still calls this method. anyway, more things in old-rb are broken so i guess losing this function now isn't really a problem...

indico/modules/rb_new/client/js/reducers.js
```diff
@@ -31,7 +31,17 @@ import {reducer as adminReducer} from './modules/admin';
 import {reducer as linkingReducer} from './common/linking';
 
 
+function overridesReducer(state = {}, action) {
+    switch (action.type) {
+        case 'EXTEND_OVERRIDES':
```
ThiefMaster: what about using the constant?

indico/core/signals/rb.py
ThiefMaster: nevermind, removed in master by #3841

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -387,4 +391,5 @@ def _process(self):
             self.reject()
         elif self.action == 'cancel':
             self.occurrence.cancel(session.user)
+        signals.rb.booking_occurrence_state_changed.send(self.occurrence)
```
ThiefMaster: Shouldn't this be done in the cancel/reject methods of `ReservationOccurrence`? Otherwise it won't be triggered when something else cancels it (e.g. a colliding booking that has priority)

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -265,6 +267,7 @@ def _process(self):
             self.reject()
         elif self.action == 'cancel':
             self.booking.cancel(session.user)
+        signals.rb.booking_state_changed.send(self.booking)
```
ThiefMaster: I think this should be called from the corresponding methods in `Reservation`.

indico/modules/rb_new/controllers/backend/bookings.py
mic4ael: Hm, I wasn't sure about this. Initially it was inside the corresponding methods and I moved it here at some point because of some reason (don't remember now what the reason was :P)

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -214,6 +215,7 @@ def _process(self):
             db.session.rollback()
             raise ExpectedError(unicode(e))
 
+        signals.rb.booking_created.send(resv)
```
ThiefMaster: call this in `Reservation.create_from_data`? (right before `notify_creation`, also flush before triggering the signal just in case)

ThiefMaster:
I'm gonna merge this without the commit that removes the DB columns (5724b1c) for now so it's easier to test the plugin without losing data when doing an upgrade here. I stored the commit as a patch locally so we don't lose the changes.

ThiefMaster:
Merged in 0259163d54

PR #3783, opened 2019-02-21
Markdown rendering of room comments

ThiefMaster:
Script to fix existing comments & key locations:

```python
import re
from html2text import HTML2Text

def _fix(s):
    s = re.sub(r'<font color=[^> ]+>(.+?)</font>', r'<strong>\1</strong>', s)
    return HTML2Text(bodywidth=0).handle(s).strip()

for r in Room.query:
    for attr in ('comments', 'key_location'):
        if not getattr(r, attr):
            continue
        old = getattr(r, attr).strip()
        new = _fix(old)
        if old != new:
            print 'fixing {} in {}'.format(attr, r.name), '\n', old, '\n', new, '\n\n'
            setattr(r, attr, new)
```

### DIFF
--- a/indico/web/client/js/react/util/Markdown.jsx
+++ b/indico/web/client/js/react/util/Markdown.jsx
@@ -1,0 +1,52 @@
+/* This file is part of Indico.
+ * Copyright (C) 2002 - 2019 European Organization for Nuclear Research (CERN).
+ *
+ * Indico is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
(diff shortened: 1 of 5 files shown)


indico/web/client/js/react/util/Markdown.jsx
```diff
@@ -0,0 +1,51 @@
...
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactMarkdown from 'react-markdown';
+
+// eslint-disable-next-line react/prop-types
+const ExternalLink = ({href, children}) => (
+    <a href={href} target="_blank" rel="noopener noreferrer">{children}</a>
```
pferreir: Cool 👍

PR #3793, opened 2019-02-22
Add A6 size

pferreir:
Pretty straightforward, but had to force sorting of the enum.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,16 @@ Changelog
 =========
 
 
+Version 2.1.8
+-------------
+
+*Unreleased*
+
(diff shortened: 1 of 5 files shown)


ThiefMaster:
CHANGES.rst entry please:

```
Version 2.1.8
-------------

*Unreleased*

Improvements
^^^^^^^^^^^^

- Add A6 to page size options (:issue:`3793`)
```

Even if we probably/hopefully never release that version, that way we have a list of stuff to copy into the 2.2 release notes eventually.

pferreir:
@ThiefMaster sure, forgot this is not master.

PR #3794, opened 2019-02-22
Ask users if they're OK with deleting events w/ bookings

pferreir:
image

image

### DIFF
--- a/indico/modules/categories/controllers/management.py
+++ b/indico/modules/categories/controllers/management.py
@@ -35,6 +35,7 @@ from indico.modules.categories.util import get_image_data
 from indico.modules.categories.views import WPCategoryManagement
 from indico.modules.events import Event
 from indico.modules.events.util import update_object_principals
+from indico.modules.rb.models.reservations import Reservation, ReservationLink
 from indico.util.fs import secure_filename
 from indico.util.i18n import _, ngettext
 from indico.util.string import crc32
@@ -351,7 +352,15 @@ class RHDeleteEvents(RHManageCategorySelectedEventsBase):
     def _process(self):
         is_submitted = 'confirmed' in request.form
         if not is_submitted:
-            return jsonify_template('events/management/delete_events.html', events=self.events)
+            # find out which active bookings are linked to the events in question
+            num_bookings = (Reservation.query
+                            .join(ReservationLink)
+                            .join(Event, Event.id == ReservationLink.linked_event_id)
+                            .filter(Event.id.in_(e.id for e in self.events),
+                                    Reservation.is_pending | Reservation.is_accepted)
+                            .count())
+            return jsonify_template('events/management/delete_events.html',
+                                    events=self.events, num_bookings=num_bookings)
         for ev in self.events[:]:
             ev.delete('Bulk-deleted by category manager', session.user)
         flash(ngettext('You have deleted one event', 'You have deleted {} events', len(self.events))
--- a/indico/modules/events/management/templates/delete_event.html
+++ b/indico/modules/events/management/templates/delete_event.html
@@ -1,10 +1,19 @@
 {% from 'confirmation_dialog.html' import confirmation_dialog %}
 
+{% set num_bookings = event.all_room_reservation_links.count() %}
 {% set confirmation_message %}
-    {% trans -%}
-        Please note that if you delete the event you will lose all the information contained in it.
-        This operation is irreversible!
-    {%- endtrans %}
+    {% if num_bookings %}
+        {% trans strong='<strong>'|safe, endstrong='</strong>'|safe -%}
+            Please note that if you delete the event you will lose all the information contained in it,
+            {{ strong }}including {{ num_bookings }} Room Booking(s){{ endstrong }} linked to it.
+            This operation is irreversible!
+        {%- endtrans %}
+    {% else %}
+        {% trans -%}
+            Please note that if you delete the event you will lose all the information contained in it.
+            This operation is irreversible!
+        {%- endtrans %}
+    {% endif %}
 {% endset %}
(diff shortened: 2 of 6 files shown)


indico/modules/categories/controllers/management.py
```diff
@@ -351,7 +353,15 @@ class RHDeleteEvents(RHManageCategorySelectedEventsBase):
     def _process(self):
         is_submitted = 'confirmed' in request.form
         if not is_submitted:
-            return jsonify_template('events/management/delete_events.html', events=self.events)
+            # find out which active bookings are linked to the events in question
+            num_bookings = (Reservation.query
+                            .join(ReservationLink)
+                            .join(Event, Event.id == ReservationLink.linked_event_id)
+                            .filter(Event.id.in_(map(attrgetter('id'), self.events)),
```
ThiefMaster: `.in_(e.id for id in self.events)` is actually shorter than map+attrgetter

indico/modules/events/management/templates/delete_event.html
```diff
@@ -1,10 +1,18 @@
 {% from 'confirmation_dialog.html' import confirmation_dialog %}
 
 {% set confirmation_message %}
-    {% trans -%}
-        Please note that if you delete the event you will lose all the information contained in it.
-        This operation is irreversible!
-    {%- endtrans %}
+    {% if event.reservations %}
```
ThiefMaster: `{% set num_bookings = event.all_room_reservation_links.count() %}` and use this here and below

indico/modules/rb_new/__init__.py
```diff
@@ -40,6 +43,27 @@ def _sidemenu_items(sender, event, **kwargs):
                            icon='location')
 
 
+@signals.users.merged.connect
+def _merge_users(target, source, **kwargs):
+    BlockingPrincipal.merge_users(target, source, 'blocking')
+    Blocking.find(created_by_id=source.id).update({Blocking.created_by_id: target.id})
```
ThiefMaster: Let's use `XXX.query.filter_by(...)` instead of `XXX.find()` - less stuff to update when we finally get rid of it

indico/modules/rb_new/__init__.py
```diff
@@ -40,6 +43,27 @@ def _sidemenu_items(sender, event, **kwargs):
...
+    Reservation.find(created_by_id=source.id).update({Reservation.created_by_id: target.id})
+    Reservation.find(booked_for_id=source.id).update({Reservation.booked_for_id: target.id})
+    Room.find(owner_id=source.id).update({Room.owner_id: target.id})
+    rb_settings.acls.merge_users(target, source)
```
ThiefMaster: we also need this one (forgot that when adding the acls):

    RoomPrincipal.merge_users(target, source, 'room')

indico/modules/rb_new/__init__.py
```diff
@@ -40,6 +43,27 @@ def _sidemenu_items(sender, event, **kwargs):
...
+
+
+@signals.event.deleted.connect
+def _event_deleted(event, user, **kwargs):
+    reservation_links = (ReservationLink.query.with_parent(event)
```
ThiefMaster: `with_parent(event, 'event')` to make sure it uses the `event` relationship and not `linked_event` (which would be None for contribs/blocks)

ThiefMaster: Actually, since this is a single event you can get all links using `event.all_room_reservation_links`

indico/modules/categories/controllers/management.py
pferreir: I guess you mean `... for e ...` 😉 but yeah, thanks :)

indico/modules/rb_new/__init__.py
pferreir: This was just copy-paste from the other file.

indico/modules/categories/controllers/management.py
```diff
@@ -18,6 +18,7 @@
 
 import os
 from io import BytesIO
+from operator import attrgetter
```
ThiefMaster: unused

indico/modules/events/management/templates/delete_event.html
```diff
@@ -1,10 +1,19 @@
 {% from 'confirmation_dialog.html' import confirmation_dialog %}
 
+{% set num_bookings = event.all_room_reservation_links.count() %}
 {% set confirmation_message %}
-    {% trans -%}
-        Please note that if you delete the event you will lose all the information contained in it.
-        This operation is irreversible!
-    {%- endtrans %}
+    {% if num_bookings %}
+        {% trans num_bookings=num_bookings, strong='<strong>'|safe, endstrong='</strong>'|safe -%}
```
ThiefMaster: i think the `num_bookings=num_bookings, ` can be omitted

indico/modules/rb_new/__init__.py
```diff
@@ -40,6 +44,27 @@ def _sidemenu_items(sender, event, **kwargs):
                            icon='location')
 
 
+@signals.users.merged.connect
+def _merge_users(target, source, **kwargs):
+    BlockingPrincipal.merge_users(target, source, 'blocking')
```
ThiefMaster: move down to the other principal merge (or the other one up here)?

indico/modules/categories/controllers/management.py
pferreir: 🤦‍♂️

indico/modules/rb_new/__init__.py
pferreir: Well, my logic was actually to group things together: first blockings, then reservations and finally rooms.

PR #3796, opened 2019-02-28
RB Admin: General settings

ThiefMaster:
image

`admin_principals` and `authorized_principals` are still missing ad those are more complex. I'll do that in a separate PR - they need a modified `PrincipalSearchField` component that can work with simple data and doesn't need a dict with details for each users.

### DIFF
--- a/indico/modules/rb_new/client/js/modules/admin/CategoryList.jsx
+++ b/indico/modules/rb_new/client/js/modules/admin/CategoryList.jsx
@@ -1,0 +1,74 @@
+/* This file is part of Indico.
+ * Copyright (C) 2002 - 2019 European Organization for Nuclear Research (CERN).
+ *
+ * Indico is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Indico is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import _ from 'lodash';
+import React, {useState} from 'react';
+import PropTypes from 'prop-types';
+import {Dropdown} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+
+
+const isValid = value => (+value >= 0);
+
+
+// TODO: make this nicer (searching by title/id, showing titles instead of IDs)
+
+/**
+ * A field that lets the user enter category IDs.
+ */
+const CategoryList = (props) => {
+    const {value, disabled, onChange, onFocus, onBlur} = props;
+    const [options, setOptions] = useState(value.map(x => ({text: x, value: x})));
+
+    const handleAddItem = (e, {value: newValue}) => {
+        if (isValid(newValue)) {
+            setOptions([...options, {text: newValue, value: +newValue}]);
+        }
+    };
+
+    const handleChange = (e, {value: newValue}) => {
+        newValue = _.uniq(newValue.filter(isValid).map(x => +x));
+        setOptions(newValue.map(x => ({text: x, value: x})));
+        onChange(newValue);
+        onFocus();
+        onBlur();
+    };
+
+    return (
+        <Dropdown options={options}
+                  value={value}
+                  disabled={disabled}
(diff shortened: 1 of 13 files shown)


indico/modules/rb_new/client/js/modules/admin/CategoryList.jsx
```diff
@@ -0,0 +1,74 @@
...
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Indico; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import _ from 'lodash';
+import React, {useState} from 'react';
```
mic4ael: Oh, is that the first usage of the hooks in the Indico codebase? :D

indico/modules/rb_new/client/js/modules/admin/CategoryList.jsx
```diff
@@ -0,0 +1,74 @@
...
+CategoryList.propTypes = {
+    value: PropTypes.arrayOf(PropTypes.number).isRequired,
+    disabled: PropTypes.bool.isRequired,
+    onChange: PropTypes.func.isRequired,
+    onFocus: PropTypes.func.isRequired,
+    onBlur: PropTypes.func.isRequired,
+};
+
+
+export default CategoryList;
```
mic4ael: Why don't you just define `CategoryList` as a normal function and export it when declaring it `export function CategoryList..`?

indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
```diff
@@ -0,0 +1,220 @@
...
+                    {fprops => (
+                        <Form onSubmit={fprops.handleSubmit}
+                              success={fprops.submitSucceeded && !fprops.dirtySinceLastSubmit}>
+                            <Field name="tileserver_url" component={ReduxFormField} as="input"
+                                   format={formatters.trim} formatOnBlur
+                                   label={Translate.string('Tileserver URL')}
+                                   disabled={fprops.submitting}
+                                   validate={val => {
+                                       if (!val) {
+                                           return undefined;
```
mic4ael: Could just return but I guess being explicit here is nice.

indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
```diff
@@ -0,0 +1,220 @@
...
+                                       }
+                                   }}>
+                                <p className="field-description">
+                                    <Translate>
+                                        If you want to use the map, specify the URL to a tileserver covering the
+                                        area in which your rooms are.
+                                    </Translate>
+                                </p>
+                            </Field>
+                            <Field name="booking_limit" component={ReduxFormField} as="input"
```
mic4ael: I think we use camelCase notation for form fields in other forms. Why did you choose to use snake_case here?

indico/core/settings/converters.py
```diff
@@ -81,3 +82,52 @@ def from_python(self, value):
...
+        self.collection_class = collection_class
+        pks = inspect(model).primary_key
+        assert len(pks) == 1
+        self.column = pks[0]
+
+    def from_python(self, value):
+        if not value:
+            return []
+        assert all(isinstance(x, self.model) for x in value)
+        return [inspect(x).identity_key[1][0] for x in value]
```
mic4ael: What if we have a composite key? I guess this will always take the first value of a PK. Anyway, not sure if this field should support composite keys anyway.

ThiefMaster: fails the assertion in the constructor

indico/modules/rb_new/client/js/modules/admin/CategoryList.jsx
ThiefMaster: yep ;)

indico/modules/rb_new/client/js/modules/admin/CategoryList.jsx
ThiefMaster: that'd work as well, just a matter of preference i guess

indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
ThiefMaster: yes, i don't like mixing bare returns and returns with a value in the same function.

indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
ThiefMaster: i had that initially, but then realized we camelize and snakify for no real reason. I totally understand doing it when we actually need to work with the keys a lot in JS, but otherwise i think having consistent names in frontend and backend is great e.g. to be able to search for them

PR #3798, opened 2019-03-01
Reminders before bookings coming to an end

### LINKED ISSUE #3772: Send e-mail message when booking is about to finish
In the core RB system, we send out messages to remind people that they have an ongoing booking. In Burotel, we should instead warn them that they have a booking that is about to end. The number of days in advance could be a configuration setting.
We should disable the normal reminder booking notifications  in Burotel.

mic4ael:
closes #3772


indico/migrations/versions/20190227_1523_081d4c97060a_add_notification_sent_column_to_.py
```diff
@@ -0,0 +1,35 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '081d4c97060a'
+down_revision = '7aabedfb5e3a'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('reservations', sa.Column('notification_sent', sa.Boolean(), nullable=False, server_default='false'),
```
ThiefMaster: `end_notification_sent`, it's not very clear what kind of notification it is otherwise.

indico/migrations/versions/20190227_1523_081d4c97060a_add_notification_sent_column_to_.py
```diff
@@ -0,0 +1,35 @@
...
+                  schema='roombooking')
+    op.add_column('rooms', sa.Column('notification_before_end_daily', sa.Integer()), schema='roombooking')
```
ThiefMaster: what about calling it `end_notification`? it's shorter but still conveys all necessary information

indico/migrations/versions/20190227_1523_081d4c97060a_add_notification_sent_column_to_.py
```diff
@@ -0,0 +1,35 @@
...
+    op.add_column('rooms', sa.Column('notification_before_end_weekly', sa.Integer()), schema='roombooking')
+    op.add_column('rooms', sa.Column('notification_before_end_monthly', sa.Integer()), schema='roombooking')
+    op.add_column('rooms',
+                  sa.Column('notifications_before_end_enabled', sa.Boolean(), nullable=False, server_default='true'),
```
ThiefMaster: remove the `server_default` values again below, you just need them so you can create the non-nullable columns

indico/modules/rb/__init__.py
```diff
@@ -42,8 +42,13 @@
     'notification_before_days_weekly': 5,
     'notification_before_days_monthly': 7,
     'notifications_enabled': True,
+    'notification_before_end_daily': 1,
```
ThiefMaster: not sure about the defaults, maybe 7 for everything makes more sense?

indico/modules/rb/models/rooms.py
```diff
@@ -139,6 +139,15 @@ class Room(versioned_cache(_cache, 'id'), ProtectionManagersMixin, db.Model, Ser
     notification_before_days_monthly = db.Column(
         db.Integer
     )
+    notification_before_end_daily = db.Column(
+        db.Integer
```
ThiefMaster: at least in newer code we explicitly specify `nullable=True` even though it's the default if omitted

indico/modules/rb/templates/emails/reservations/reminders/finishing_bookings.html
```diff
@@ -0,0 +1,32 @@
+{% extends 'rb/emails/base_email.html' %}
+
+{% block header_recipient %}{{ user.first_name }}{% endblock %}
+{% block subject %}Some of your bookings will come to an end soon{% endblock %}
+
+
+{% block body -%}
+    {%- if reservations | length == 1 -%}
```
ThiefMaster: clearer without space around the `|`

indico/modules/rb_new/client/js/common/rooms/RoomEditModal.module.scss
```diff
@@ -27,3 +27,8 @@
     margin-bottom: 10px;
 }
 
+.room-form {
+    :global(.fields) > :global(.field) {
+        flex: 1 1 auto !important;
```
ThiefMaster: I was going to suggest to not use the shorthand property, but at least https://css-tricks.com/snippets/css/a-guide-to-flexbox/ recommends it.. :/

Anyway, the second/third params are optional so unless you need them to override something let's omit them.

indico/migrations/versions/20190227_1523_081d4c97060a_add_notification_sent_column_to_.py
mic4ael: Agreed. Honestly, I am not a big fan of naming conventions in this PR.

indico/modules/rb/__init__.py
mic4ael: Yeah, those are the initial values that I came up with. And I think it is something we should discuss.

indico/modules/rb_new/client/js/common/rooms/RoomEditModal.module.scss
mic4ael: This is a fix to how (as far as I know) Chrome renders flex elements with `flex-basis: auto`. On the other hand I think it will be enough to just override `flex-grow` here because that is the only thing I care about.

indico/modules/rb/templates/emails/reservations/reminders/finishing_bookings.html
```diff
@@ -0,0 +1,32 @@
...
+        There is one booking under your name that will come to an end soon:<br>
+    {%- else -%}
+        There are bookings under your name that will soon come to an end:<br>
```
mvidalgarcia: Any reason why the messages are different depending on the number of bookings?
"...will soon..." vs "...will come...soon"

indico/modules/rb_new/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -254,35 +260,82 @@ const columns = [
...
+                type: 'number',
+                min: 1,
+                max: 30,
+            },
+            required: false
+        }]
+    }, {
+        type: 'formgroup',
+        key: 'notificationsOfFinishingBookings',
+        label: Translate.string('How many days before the end of a booking should we send a reminder?'),
```
mvidalgarcia: For consistency reasons, I'd phrase this sentence in the same way as you did for the `notificationBeforeDays`
What about something like "How many days before the end of a booking reminders should be sent?" ?

indico/modules/rb/templates/emails/reservations/reminders/finishing_bookings.html
mic4ael: Probably forgot to fix it since I was modified those messages a few times

indico/modules/rb_new/client/js/common/rooms/RoomEditModal.jsx
mic4ael: I don't mind changing it. I was experimenting with different length of those messages to make it look nicer in the modal.

indico/migrations/versions/20190227_1523_081d4c97060a_add_notification_sent_column_to_.py
mic4ael: Ahh, true. I completely forgot that I need to remove that afterwards.

indico/modules/rb/forms/settings.py
```diff
@@ -42,6 +42,17 @@ class SettingsForm(IndicoForm):
     notification_before_days_monthly = IntegerField(_('Send booking reminders X days before (monthly)'),
                                                     [InputRequired(), NumberRange(min=1, max=30)])
     notifications_enabled = BooleanField(_('Reminders enabled'))
+    notification_before_end_daily = IntegerField(_('Send reminders about ending bookings X days before their end date '
+                                                   '(daily)'),
```
pferreir: What about "(daily bookings)"? Because otherwise people may get the impression that those reminders are sent daily.

indico/modules/rb/forms/settings.py
```diff
@@ -42,6 +42,17 @@ class SettingsForm(IndicoForm):
...
+                                                 [InputRequired(), NumberRange(min=1, max=30)])
+    notification_before_end_weekly = IntegerField(_('Send reminders about ending bookings X days before their end date '
+                                                    '(weekly)'),
+                                                  [InputRequired(), NumberRange(min=1, max=30)])
+    notification_before_end_monthly = IntegerField(_('Send reminders about ending bookings X days before their end '
+                                                     'date (monthly)'),
+                                                   [InputRequired(), NumberRange(min=1, max=30)])
+    notifications_before_end_enabled = BooleanField(_('Reminders of ending bookings enabled'))
```
pferreir: I think this text doesn't read very well. It's already confusing what this field does. What about "Send reminders when bookings are about to end" or similar?

indico/migrations/versions/20190227_1523_081d4c97060a_add_notification_sent_column_to_.py
pferreir: I guess this wasn't fixed yet? Just saying because the conversation is marked as "resolved".

indico/modules/rb/forms/settings.py
pferreir: (same for options below, of course...)

indico/modules/rb/models/reservations.py
```diff
@@ -243,6 +243,11 @@ def __table_args__(cls):
         nullable=True,
         index=True
     )
+    notification_sent = db.Column(
```
pferreir: I guess this will change to `end_notification_sent`?

mic4ael:
Updated

indico/modules/rb/tasks.py
```diff
@@ -118,3 +128,44 @@ def roombooking_occurrences(debug=False):
             _notify_occurrences(user, user_occurrences)
     if not debug:
         db.session.commit()
+
+
+@celery.periodic_task(name='notifications_about_finishing_bookings', run_every=crontab(minute='0', hour='8'))
+def notifications_about_finishing_bookings():
```
pferreir: Maybe `booking_end_notifications`?

indico/modules/rb/forms/settings.py
```diff
@@ -42,6 +42,17 @@ class SettingsForm(IndicoForm):
...
+    end_notification = IntegerField(_('Send reminders about ending bookings X days before their end date '
+                                      '(daily)'),
+                                    [InputRequired(), NumberRange(min=1, max=30)])
+    end_notification_weekly = IntegerField(_('Send reminders about ending bookings X days before their end date '
+                                             '(weekly)'),
+                                           [InputRequired(), NumberRange(min=1, max=30)])
+    end_notification_monthly = IntegerField(_('Send reminders about ending bookings X days before their end '
+                                              'date (monthly)'),
+                                            [InputRequired(), NumberRange(min=1, max=30)])
+    end_notifications_enabled = BooleanField(_('Reminders of ending bookings enabled'))
```
pferreir: What about "Send a reminder whenever a booking is about to end"?

indico/modules/rb/forms/settings.py
```diff
@@ -42,6 +42,17 @@ class SettingsForm(IndicoForm):
     notification_before_days_monthly = IntegerField(_('Send booking reminders X days before (monthly)'),
                                                     [InputRequired(), NumberRange(min=1, max=30)])
     notifications_enabled = BooleanField(_('Reminders enabled'))
+    end_notification = IntegerField(_('Send reminders about ending bookings X days before their end date '
+                                      '(daily)'),
```
pferreir: I think this is very confusing. Maybe we should say "(daily bookings)" instead of just "(daily)'? Otherwise people may think that the reminders will be sent daily.

indico/modules/rb/models/reservations.py
mic4ael: yup, changed already

indico/modules/rb/forms/settings.py
mic4ael: Yeah, I will change it soon when I will move those settings to the new RB.

indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
```diff
@@ -170,7 +170,51 @@ class SettingsPage extends React.PureComponent {
...
-                                               label={Translate.string('Monthly')}
+                                               label={Translate.string('Monthly bookings')}
+                                               validate={v.range(1, 30)}
+                                               disabled={fprops.submitting} />
+                                    </Form.Group>
+                                </Message>
+                            </FieldCondition>
+                            <Field name="end_notifications_enabled" component={ReduxCheckboxField}
+                                   componentLabel={Translate.string('Send reminders when bookings are about to end')}
+                                   disabled={fprops.submitting} />
```
ThiefMaster: let's a add a description here for consistency with the other checkbox

mic4ael: I didn't add it there since, at least to me, the label is pretty explanatory.

ThiefMaster: then let's remove it from the other checkbox

mic4ael: fair enough ;)

indico/modules/rb/tasks.py
```diff
@@ -118,3 +128,44 @@ def roombooking_occurrences(debug=False):
             _notify_occurrences(user, user_occurrences)
     if not debug:
         db.session.commit()
+
+
+@celery.periodic_task(name='notifications_about_finishing_bookings', run_every=crontab(minute='0', hour='8'))
```
ThiefMaster: what about `roombooking_end_notifications`?

indico/modules/rb/templates/emails/reservations/reminders/finishing_bookings.html
```diff
@@ -0,0 +1,34 @@
+{% extends 'rb/emails/base_email.html' %}
+
+{% block header_recipient %}{{ user.first_name }}{% endblock %}
+{% block subject %}Some of your bookings will come to an end soon{% endblock %}
+
+
+{% block body -%}
+    {% block message %}
+        {%- if reservations|length == 1 -%}
+            There is one booking under your name that will soon come to an end:<br>
```
ThiefMaster: There is *a* booking...

indico/modules/rb/templates/emails/reservations/reminders/finishing_bookings.html
```diff
@@ -0,0 +1,34 @@
+{% extends 'rb/emails/base_email.html' %}
+
+{% block header_recipient %}{{ user.first_name }}{% endblock %}
+{% block subject %}Some of your bookings will come to an end soon{% endblock %}
```
ThiefMaster: pluralize the subject as well?

indico/modules/rb_new/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -254,35 +260,82 @@ const columns = [
...
+                type: 'number',
+                min: 1,
+                max: 30,
+            },
+            required: false
+        }]
+    }, {
+        type: 'formgroup',
+        key: 'notificationsOfFinishingBookings',
+        label: Translate.string('How many days before the end of a booking reminders should be sent?'),
```
ThiefMaster: s/reminders should/should reminders/ - also in all the other places where you have a `?` at the end. the way it's currently written it'd be fine as an statement but not as a question.

PR #3800, opened 2019-03-04
Split timeline labels in two rows

### LINKED ISSUE #3770: Solve issue with desk names being cut off
In the "Book a Room" view in particular, this is quite annoying. But it's also a bit of a problem in the Calendar.

image

In Burotel, the desk numbers are quite important and they're normally at the end of the text string.

panagiotappl:
closes #3770

### DIFF
--- a/indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
+++ b/indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
@@ -20,10 +20,9 @@ import moment from 'moment';
 import PropTypes from 'prop-types';
 import React from 'react';
 import {AutoSizer, List, WindowScroller} from 'react-virtualized';
-import {Icon, Placeholder, Popup} from 'semantic-ui-react';
+import {Icon, Placeholder} from 'semantic-ui-react';
 import LazyScroll from 'redux-lazy-scroll';
 import {toMoment} from 'indico/utils/date';
-import {Translate} from 'indico/react/i18n';
 import {TooltipIfTruncated} from 'indico/react/components';
 
 import OccurrenceActionsDropdown from '../bookings/OccurrenceActionsDropdown';
@@ -278,27 +277,14 @@ export default class DailyTimelineContent extends React.Component {
 }
 
 export function TimelineRowLabel({label, name, verboseName, availability, longLabel, onClickLabel}) {
-    let color, tooltip;
-    switch (availability) {
-        case 'conflict':
-            color = 'red';
-            tooltip = Translate.string('Conflicts for the selected period');
-            break;
-        case 'alternatives':
-            color = 'orange';
-            tooltip = Translate.string('Room suggested based on the selected criteria');
-            break;
-        default:
-            color = 'green';
-            tooltip = Translate.string('Room available');
-    }
+    const colorMap = {conflict: 'red', alternatives: 'orange'};
+    const color = colorMap[availability] || 'green';
 
     const icon = availability ? <Icon name="circle" size="tiny" color={color} styleName="dot" /> : null;
     const labelContent = verboseName ? (
         <span styleName="split-label">
             <div>
-                {icon}
-                {name}
+                {icon}{name}
             </div>
             <TooltipIfTruncated>
                 <div styleName="sub-label">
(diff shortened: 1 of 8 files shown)


indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
```diff
@@ -290,10 +292,20 @@ export function TimelineRowLabel({label, availability, longLabel, onClickLabel})
             color = 'green';
             tooltip = Translate.string('Room available');
     }
+
+    const labelContent = verboseName ? (
+        <span styleName="split-label">
+            <div>{name}</div>
+            <div styleName="sub-label">
+                <small>{verboseName}</small>
```
mvidalgarcia: Maybe it's worth trying [`<TooltipIfTruncated />`](https://github.com/indico/indico/blob/master/indico/web/client/js/react/components/TooltipIfTruncated.jsx) here in case the `verboseName` is too long that it gets _ellipsis'ed_.

panagiotappl: Does the `<TooltipIfTruncated>` that wraps around the label <div> also apply here?

mvidalgarcia: Ah, I didn't notice it. But it seems that the `<TooltipIfTruncated />` is only working for the `name` and not for the `verboseName`.

I also noticed this.
image
The `name` is not being _ellipsized_ as it used to be that might be ugly if there are too long room names (I'm not sure if there are)

meluru:
Are you aware of this problem on `book a room` page?  
image

indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
panagiotappl: Good observation but I think the name is always going to be in the format of: Building/Floor-RoomNo right?

mvidalgarcia: I think so, so maybe it's worth only applying the tooltip for the `verboseName`. Let's ping the others to see what they think. @pferreir @ThiefMaster @mic4ael @meluru

meluru: I think it's a good idea.

indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
```diff
@@ -275,45 +276,57 @@ export default class DailyTimelineContent extends React.Component {
     }
 }
 
-export function TimelineRowLabel({label, availability, longLabel, onClickLabel}) {
-    let color, tooltip;
+export function TimelineRowLabel({label, name, verboseName, availability, longLabel, onClickLabel}) {
+    let color;
     switch (availability) {
```
ThiefMaster: what about replacing this switch with a dict lookup?

```js
const colorMap = {conflict: 'red', alternatives: 'orange'};
const color = colorMap[availability] || 'green';
```

indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
```diff
@@ -275,45 +276,57 @@ export default class DailyTimelineContent extends React.Component {
...
+                {icon}
+                {name}
+            </div>
+            <TooltipIfTruncated>
+                <div styleName="sub-label">
+                    <small>{verboseName}</small>
+                </div>
+            </TooltipIfTruncated>
+        </span>
+    ) : (<span>{icon}{label}</span>);
```
ThiefMaster: `<span>{icon}{label}</span>` should go in its own line

indico/modules/rb_new/client/js/common/timeline/DailyTimelineContent.jsx
```diff
@@ -275,45 +276,57 @@ export default class DailyTimelineContent extends React.Component {
...
             color = 'green';
-            tooltip = Translate.string('Room available');
     }
+
+    const icon = availability ? <Icon name="circle" size="tiny" color={color} styleName="dot" /> : null;
+    const labelContent = verboseName ? (
+        <span styleName="split-label">
+            <div>
+                {icon}
+                {name}
```
ThiefMaster: I'd remove the linebreak between the two or add one below to be consistent

PR #3811, opened 2019-03-08
Turn single radio button into text

### LINKED ISSUE #3810: Burotel: display only "Daily booking" text instead of radio button
Since in burotel this is the only option, it's pretty useless to show a single "radio button" for it.

Happens in:
  * Landing;
  * When creating booking from List of Spaces;
  * when editing a booking.

panagiotappl:
closes #3810

### DIFF
--- a/indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
+++ b/indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
@@ -208,35 +208,39 @@ class BookingEditForm extends React.Component {
         const today = moment();
         const bookingFinished = today.isAfter(startDt, 'day') && (endDt && today.isAfter(endDt, 'day'));
 
+        // all but one option are hidden
+        const showRecurrenceOptions = ['single', 'daily', 'recurring'].filter(x => hideOptions[x]).length !== 2;
         return (
             <Form id="booking-edit-form" styleName="booking-edit-form" onSubmit={handleSubmit}>
                 <Segment>
-                    <Form.Group inline>
-                        {!hideOptions.single && (
-                            <Field name="recurrence.type"
-                                   component={ReduxRadioField}
-                                   componentLabel={Translate.string('Single booking')}
-                                   radioValue="single"
-                                   disabled={submitSucceeded || bookingFinished}
-                                   onClick={() => this.recurrenceTypeChanged('single')} />
-                        )}
-                        {!hideOptions.daily && (
-                            <Field name="recurrence.type"
-                                   component={ReduxRadioField}
-                                   componentLabel={Translate.string('Daily booking')}
-                                   radioValue="daily"
-                                   disabled={submitSucceeded || bookingFinished}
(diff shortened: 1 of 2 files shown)


indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -207,36 +207,41 @@ class BookingEditForm extends React.Component {
         const bookedByCurrentUser = sessionUser.id === bookedForUser.id;
         const today = moment();
         const bookingFinished = today.isAfter(startDt, 'day') && (endDt && today.isAfter(endDt, 'day'));
-
+        const onlyDaily = hideOptions.single && hideOptions.recurring && !hideOptions.daily;
```
ThiefMaster: shouldn't this check for all possible combinations where only one option is visible?

panagiotappl: In the code for the Landing page the logic is a bit different and we only check for `onylDaily`, so I wasn't sure what kind of approach I should follow.

ThiefMaster: I guess we should change it there as well then :)

panagiotappl: Discard the prop `onlyDaily` and replace it with `hideOptions.single` etc?

ThiefMaster:
i wouldn't mention burotel in the commit message since all this commit does is hiding the radio buttons if there's only a single option selected - the fact that the burotel plugin uses this doesn't matter for this commit

indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
ThiefMaster: yep

ThiefMaster:
image


I don't think this looks amazing... what about hiding this text altogether when there's only one option?

indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -208,35 +208,40 @@ class BookingEditForm extends React.Component {
         const today = moment();
         const bookingFinished = today.isAfter(startDt, 'day') && (endDt && today.isAfter(endDt, 'day'));
 
+        const onlySingle = !hideOptions.single && hideOptions.recurring && hideOptions.daily;
+        const onlyDaily = hideOptions.single && hideOptions.recurring && !hideOptions.daily;
+        const onlyRecurring = hideOptions.single && !hideOptions.recurring && hideOptions.daily;
```
ThiefMaster: You could simplify these checks like this:

```js
// all but one option are hidden
const showRecurrenceOptions = ['single', 'daily', 'recurring'].filter(x => hideOptions[x]).length !== 2;
```

and then just use `showRecurrenceOptions` below

PR #3816, opened 2019-03-13
Fetch rooms signal

mvidalgarcia:


indico/modules/rb_new/schemas.py
```diff
@@ -72,6 +73,12 @@ class Meta:
                   'surface_area', 'latitude', 'longitude', 'telephone', 'key_location', 'max_advance_days',
                   'owner_name', 'available_equipment', 'has_photo', 'verbose_name')
 
+    @post_dump(pass_many=True)
+    def inject_fields(self, data, many):
+        if many:
```
ThiefMaster: Currently this doesn't work after editing a room when just a single room is fetched. You can use this instead so your signal is always called properly:

```
        rooms = data if many else [data]
        signals.rb.rooms_fetched.send(rooms)
```

ThiefMaster:
Merging this for now, but I think eventually we could see if we can get this callback somehow into the base schema class and use a generic signal (using the actual schema as the sender)

PR #3833, opened 2019-03-14
Timeline fixes

### LINKED ISSUE #3777: Fix several timeline issues
1. Nonbookable periods/hours show up in the disabled days in weekly/monthly view

image

2. New booking appears below the current booking in booking edit when:
- we shorten the time of the booking
- we change booking from weekly to biweekly (in case of booking that didn't start yet) (screenshot below)
image

3. Element's border is cut out on the right side of the timeline

image

meluru:
Closes #3777


indico/modules/rb_new/client/js/common/timeline/ElasticTimeline.jsx
```diff
@@ -89,7 +89,7 @@ export default class ElasticTimeline extends React.Component {
                 preConflicts: preConflicts[dt] || [],
                 blockings: blockings[dt] || [],
                 nonbookablePeriods: nonbookablePeriods[dt] || [],
-                unbookableHours: unbookableHours || [],
+                unbookableHours: dateRange.length !== 0 && !_.includes(dateRange, dt) ? [] : unbookableHours || [],
```
ThiefMaster: please add some parentheses, `a && b ? c : d || e` isn't really clear IMO

indico/modules/rb_new/client/js/common/timeline/TimelineLegend.module.scss
```diff
@@ -70,6 +70,7 @@
             &.pre-booking {
                 @include stripes($pre-booking-stripe-colors...);
                 background-size: 35px 35px;
+                color: #c4930d;
```
ThiefMaster: shouldn't this come from the palette?

indico/modules/rb_new/client/js/common/timeline/MonthlyTimelineContent.jsx
```diff
@@ -73,38 +73,56 @@ export default class MonthlyTimelineContent extends WeeklyTimelineContent {
...
         const nDays = this.dates.length;
         const daySize = (100 / nDays);
-
         const {dateRange} = this.props;
+
         const emptyDays = this.dates
             .filter(day => dateRange.length !== 0 && !dateRange.includes(day))
             .map(day => this.dates.findIndex((el) => el === day));
 
+        const weekendDays = this.dates
```
mvidalgarcia: I noticed you have this same operation in `renderHeader()`, perhaps could this be a separate function? It even could be a util function (`date.js`).

Apart from that, I think the weekend days aren't being properly calculated. For instance for the March you're highlighting Sunday and Monday.
image

indico/modules/rb_new/client/js/common/bookings/selectors.js
```diff
@@ -27,8 +27,8 @@ const getRawDetails = (state, {bookingId}) => state.bookings.details[bookingId];
 
 
 function applyPermissionsToOccurrences(occurrences, override) {
-    return _.fromPairs(Object.entries(occurrences).map(([key, data]) => {
-        data = _.fromPairs(Object.entries(data).map(([day, dayData]) => {
+    return _.update(occurrences, 'bookings', (data) => {
```
ThiefMaster: this mutates occurrences :/

indico/modules/rb_new/client/js/common/timeline/MonthlyTimelineContent.jsx
```diff
@@ -35,6 +35,12 @@ export default class MonthlyTimelineContent extends WeeklyTimelineContent {
         return [];
     }
 
+    getWeekendDays = () => (
```
mvidalgarcia: What about `get weekendDays() {...}`? Then you can use directly `this.weekendDays`.

ThiefMaster: I'd use bracket style with return for this method. Using `=> ( ... )` over multiple lines without something that's obviously a return value (especially JSX) in the body isn't immediately clear without a closer look at the brackets. While looking at this PR I first thought there's a missing `return` statement

indico/modules/rb_new/client/js/common/timeline/TimelineContent.module.scss
```diff
@@ -156,3 +153,19 @@ $split-row-height: $row-height/2;
     overflow-x: hidden;
     overflow-y: auto;
 }
+
+
+.daily-timeline {
+    margin-right: 2em;
```
mic4ael: Why do you need it here? This margin makes the scrollbar e.g. in the `UnavailableRoomsModal` look a bit out of place.

indico/modules/rb_new/client/js/common/timeline/TimelineContent.module.scss
```diff
@@ -156,3 +153,19 @@ $split-row-height: $row-height/2;
...
+    display: flex;
```
mic4ael: Why do you need those flex-related properties here?

indico/modules/rb_new/client/js/common/timeline/MonthlyTimelineContent.jsx
```diff
@@ -73,38 +81,49 @@ export default class MonthlyTimelineContent extends WeeklyTimelineContent {
     renderDividers() {
         const nDays = this.dates.length;
         const daySize = (100 / nDays);
-
         const {dateRange} = this.props;
+        const weekendDays = this.getWeekendDays();
```
mic4ael: This causes an error. IT should be `this.weekendDays`.

indico/modules/rb_new/client/js/common/timeline/MonthlyTimelineContent.jsx
```diff
@@ -73,38 +81,49 @@ export default class MonthlyTimelineContent extends WeeklyTimelineContent {
...
+                         key={`day-divider-${n}`} />
+                );
+            })
         );
     }
 
     renderHeader() {
         const {longLabel, selectable} = this.props;
         const labelWidth = longLabel ? 200 : 150;
+        const weekendDays = this.getWeekendDays();
```
mic4ael: Error, there is no such method. `this.weekendDays`.

indico/modules/rb_new/client/js/common/timeline/MonthlyTimelineContent.jsx
mvidalgarcia: Indeed [(getter docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)

meluru: true, just forgot to change it here.

indico/modules/rb_new/client/js/common/timeline/TimelineContent.module.scss
meluru: Good point, it's not useful anymore

indico/modules/rb_new/client/js/common/timeline/ElasticTimeline.jsx
```diff
@@ -89,7 +89,7 @@ export default class ElasticTimeline extends React.Component {
                 preConflicts: preConflicts[dt] || [],
                 blockings: blockings[dt] || [],
                 nonbookablePeriods: nonbookablePeriods[dt] || [],
-                unbookableHours: unbookableHours || [],
+                unbookableHours: (dateRange.length !== 0 && !_.includes(dateRange, dt)) ? [] : (unbookableHours || []),
```
mic4ael: You can use `Array.prototype.includes` instead of using `_.includes`

indico/modules/rb_new/client/js/common/timeline/MonthlyTimelineContent.jsx
```diff
@@ -73,38 +81,47 @@ export default class MonthlyTimelineContent extends WeeklyTimelineContent {
...
                     <div styleName="style.timeline-header-labels">
-                        {_.map(this.dates, (dt, n) => (
-                            <div styleName="style.timeline-header-label"
-                                 key={`timeline-header-${n}`}>
-                                <span styleName="style.timeline-label-text">
-                                    {toMoment(dt, 'YYYY-MM-DD').format('D')}
-                                </span>
-                            </div>
-                        ))}
+                        {_.map(this.dates, (dt, n) => {
```
mic4ael: I am curious why we use `_.map` when we could use `Array.prototype.map` instead.

indico/modules/rb_new/client/js/common/timeline/TimelineItem.jsx
```diff
@@ -148,7 +148,7 @@ class TimelineItem extends React.Component {
         const {startHour, endHour, onClickCandidate, onClickReservation, room, dayBased} = this.props;
         if (type === 'blocking') {
             segmentStartDt = moment(startHour, 'HH:mm');
-            segmentEndDt = moment(endHour, 'HH:mm');
+            segmentEndDt = (endHour === 24 ? moment('23:59', 'HH:mm') : moment(endHour, 'HH:mm'));
```
mic4ael: This is a pretty weird fix and anyway it is not needed as far as I can tell. The issue described in the task

> Element's border is cut out on the right side of the timeline

is solved by two places that you have already fixed in lines 129 and 239.

meluru: The issue with element border is solved in line 239.
This is a different problem I discovered on the way: blockings don't render properly if timeline `maxHour` is set to 24, because `segmentEndDt.hours()` will resolve to 0, so `segmentWidth` won't be correct

meluru: Line 129 fixes rendering occurrences that start at timeline maxHour, because we should treat them as out of bounds ones and we didn't

mic4ael: Ahh, I see now.

indico/modules/rb_new/client/js/common/timeline/ElasticTimeline.jsx
ThiefMaster: `dateRange.includes(dt)` to be exact ;)

mic4ael: Yeah, yeah. That is what I meant :P

indico/modules/rb_new/client/js/common/timeline/TimelineContent.module.scss
```diff
@@ -156,10 +156,6 @@ $split-row-height: $row-height/2;
 
 
 .daily-timeline {
-    margin-right: 2em;
-    display: flex;
-    flex-direction: column;
-    flex: 1;
```
mvidalgarcia: Removing `flex` and `flex-direction` is breaking the edit booking timeline.
image

mic4ael: Good catch.

PR #3834, opened 2019-03-15
Travis CI: Start testing on Python 3 in allow_failures mode

cclauss:
https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail

Also, [Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

"_If you currently specify __sudo: false__ in your __.travis.yml__, we recommend removing that configuration_"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
-sudo: false
 language: python
 python:
   - 2.7
+  - 3.6
+matrix:
+  allow_failures:
+    - python: 3.6
 git:
   submodules: false
   depth: 500
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 python:
   - 2.7
-  - 3.6
+  # - 3.6  # TODO: Enable Python 3
 matrix:
   allow_failures:
     - python: 3.6


ThiefMaster:
Thanks, but we'll add Python 3 once we actually start supporting it. But feel free to update the PR to remove `sudo: false` if Travis recommends that.

cclauss:
Did you read the URL in the commit message?  
> [_Allow failures_] lets you add in experimental and preparatory builds to test against versions or configurations that you are not ready to officially support.

Travis CI designed this feature for the current state of this repo.

ThiefMaster:
Yes, but we know that many things in Indico are broken in Python 3 right now (but luckily most are easy to solve once we switch - which will be a hard switch with a new major release btw, no 2+3-compatible version) so there is no point in wasting Travis' resources to run tests in a known-broken environment.

cclauss:
Not best practice the make a cold switch from 2 to 3 without being able to run both with a common codebase...  I have watched that fail _many_ times.  The A/B comparisons are too valuable to be lost.

ThiefMaster:
For a library that has good test coverage for everything I'd fully agree. But I think it only adds more complexity to not fully switch over to Python 3 at some point but rather use `six` etc instead.

Anyway, a major v3 release that uses Python 3 is a good opportunity to bump our other requirements as well (e.g. Postgres 10/11 instead of 9.6) even if there is no technical need for this right now, just to be future-proof and not locked out from features that might be nice to use.

PR #3835, opened 2019-03-15
Modernize Python 2 code to get ready for Python 3

cclauss:
* Use __print()__ function in both Python 2 and Python 3
    * Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.
* Use __exec()__ function in both Python 2 and Python 3
    * __exec()__ is a function in Python 3.
* Old style exceptions --> new style for Python 3
    * Old style exceptions are syntax errors in Python 3 but new style exceptions work as expected in both Python 2 and Python 3.

Related to #3834

### DIFF
--- a/bin/maintenance/update_backrefs.py
+++ b/bin/maintenance/update_backrefs.py
@@ -15,6 +15,7 @@
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import unicode_literals
+from __future__ import print_function
 
 import sys
 from collections import defaultdict
@@ -90,10 +91,10 @@ def main(ci):
         if in_backrefs and not backrefs_written:
             _write_backrefs(rels, new_source)
         if not backrefs_written:
-            print cformat('%{yellow}Class {} has no comment for backref information').format(cls.__name__)
+            print(cformat('%{yellow}Class {} has no comment for backref information').format(cls.__name__))
             has_missing = True
         if source != new_source:
-            print cformat('%{green!}Updating backref info for {} in {}').format(cls.__name__, path)
+            print(cformat('%{green!}Updating backref info for {} in {}').format(cls.__name__, path))
             has_updates = True
             with open(path, 'w') as f:
                 f.writelines(line + '\n' for line in new_source)
--- a/bin/maintenance/update_backrefs.py
+++ b/bin/maintenance/update_backrefs.py
@@ -14,19 +14,20 @@
 # You should have received a copy of the GNU General Public License
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
-from __future__ import unicode_literals
-from __future__ import print_function
+from __future__ import print_function, unicode_literals
 
 import sys
 from collections import defaultdict
 from operator import itemgetter
 
 import click
-from sqlalchemy import inspect
 
 from indico.core.db import db
 from indico.core.db.sqlalchemy.util.models import import_all_models
 from indico.util.console import cformat
+
+from sqlalchemy import inspect
+
 
 click.disable_unicode_literals_warning = True
 
--- a/bin/maintenance/update_backrefs.py
+++ b/bin/maintenance/update_backrefs.py
@@ -21,12 +21,11 @@ from collections import defaultdict
 from operator import itemgetter
 
 import click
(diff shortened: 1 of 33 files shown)


bin/maintenance/update_backrefs.py
```diff
@@ -15,6 +15,7 @@
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import unicode_literals
+from __future__ import print_function
```
ThiefMaster: We merge imports from the same source -> 

    from __future__ import print_function, unicode_literals

bin/utils/apiProxy.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
ThiefMaster: should come after the header (separate block of imports before the stdlib ones)

bin/utils/createDatabaseGraphs.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
ThiefMaster: same

bin/utils/db_log.py
```diff
@@ -15,6 +15,7 @@
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import unicode_literals
+from __future__ import print_function
```
ThiefMaster: see above

indico/cli/database.py
```diff
@@ -15,6 +15,7 @@
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import unicode_literals
+from __future__ import print_function
```
ThiefMaster: see above

indico/cli/devserver.py
```diff
@@ -15,6 +15,7 @@
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import unicode_literals
+from __future__ import print_function
```
ThiefMaster: merge

indico/core/celery/cli.py
```diff
@@ -15,6 +15,7 @@
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import unicode_literals
+from __future__ import print_function
```
ThiefMaster: merge

indico/legacy/webinterface/common/tools.py
```diff
@@ -233,7 +233,7 @@ def escape_html(text, escape_quotes=False):
                           " => &quot;
                           ' => &#34;
     """
-    if type(text) != str:
+    if not isinstance(text, str):
```
ThiefMaster: This is legacy code, not sure if it can get str subclasses that should not be converted. Safer to not touch this particular one. It's part of the very legacy HTML sanitizer that will soon go away anyway.

ThiefMaster:
Thanks for helping with the cleanup of legacy Python 2 things that can be easily changed! But please update your PR to fix the issues mentioned above.

cclauss:
I must be doing something wrong with the isort command.

ThiefMaster:
Did you `pip install -r requirements.txt`? If not, that's why it moves stuff in the wrong section, since it doesn't know it's a thirdparty dependency.

cclauss:
Perfect.  Thanks.

bin/utils/db_log.py
```diff
@@ -30,12 +30,13 @@
 from threading import Lock
 
 import click
-import sqlparse
 from pygments import highlight
 from pygments.formatters.terminal256 import Terminal256Formatter
 from pygments.lexers.agile import PythonLexer, PythonTracebackLexer
 from pygments.lexers.sql import SqlLexer
 
+import sqlparse
```
ThiefMaster: sorry, forgot to mention that you need to install the dev requirements (`requirements.dev.txt`) as well

ThiefMaster:
Just a heads-up @cclauss, changing `exec` to `exec()` breaks on Python 2.7.5 (e.g. on CentOS 7) if the function containing the exec contains a dict/list comprehension. I've reverted this part of your changes for now, but it might be useful for contributions to other repos if you double-check for such cases.

PR #3840, opened 2019-03-18
Add extension point to postprocess event metadata

pferreir:
I am not sure whether this is the best approach. The initial requirements were that we should be able to add the URL of the corresponding Vidyo room at the bottom of an event's description in:
 * ICS feeds
 * Exchange events

This patch tries to address those two scenarios with a single signal `event.metadata_postprocess` that is called from different places with a different string as `sender` parameter. We would then implement the same signal-based extension logic in the `outlook` (Exchange) plugin.

### DIFF
--- a/indico/modules/events/api.py
+++ b/indico/modules/events/api.py
@@ -26,6 +26,7 @@ from sqlalchemy import Date, cast
 from sqlalchemy.orm import joinedload, subqueryload, undefer
 from werkzeug.exceptions import ServiceUnavailable
 
+from indico.core import signals
 from indico.core.config import config
 from indico.core.db import db
 from indico.core.db.sqlalchemy.principals import PrincipalType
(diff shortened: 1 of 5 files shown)


indico/modules/events/api.py
```diff
@@ -614,6 +616,10 @@ def _build_event_api_data(self, event):
             data['occurrences'] = fossilize(self._calculate_occurrences(event, self._fromDT, self._toDT,
                                             pytz.timezone(config.DEFAULT_TIMEZONE)),
                                             {Period: IPeriodFossil}, tz=self._tz, naiveTZ=config.DEFAULT_TIMEZONE)
+        # check whether the plugins want to add/override any data
+        for update in values_from_signal(
+                signals.event.metadata_postprocess.send('http-api', event=event, data=data), as_list=True):
+            data.update(update)
```
ThiefMaster: what about simply letting the signal mutate data instead of using a return value? nothing stops someone from doing this right now anyway..

pferreir: True.

pferreir: At the same time, this kind of makes things a little bit more explicit, doesn't it.

ThiefMaster:
Let's merge this (unless you want to change to mutable first)?

pferreir:
Personally, I wouldn't touch it, but if you think it makes a big difference, we can change it.

ThiefMaster:
nah mostly personal preference i guess

PR #3841, opened 2019-03-18
Remove redundant signal

mvidalgarcia:


ThiefMaster:
#3791 adds more rb signals, so it's probably better to just remove the signal and keep the file itself.

https://github.com/indico/indico/pull/3791/files#r266459026

mvidalgarcia:
Done

PR #3848, opened 2019-03-20
Images folder name and nginx group name are modified

Ahmet-Kaplan:
web/static folder name is changed to web/htdocs. So, nginx.conf file should be updated.

Also, nginx web user is named www-data.

### DIFF
--- a/docs/source/installation/production/debian/nginx.rst
+++ b/docs/source/installation/production/debian/nginx.rst
@@ -76,7 +76,7 @@ most cases.
     cat > /etc/uwsgi/apps-available/indico.ini <<'EOF'
     [uwsgi]
     uid = indico
-    gid = nginx
+    gid = www-data
     umask = 027
 
     processes = 4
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Thanks, but your change to the folder name is not correct:

- Version 2.0 (obsolete) and 2.1 (current stable) use htdocs: https://docs.getindico.io/en/stable/installation/production/debian/nginx/#configure-uwsgi-nginx
- Version 2.2 (under development) and newer use static: https://docs.getindico.io/en/latest/installation/production/debian/nginx/#configure-uwsgi-nginx

The regexps for the aliases are also different - you need to use the correct docs for the version you are using..

ThiefMaster:
About the users: Which Debian version are you using?

Ahmet-Kaplan:
I am using Debian 9 bionic.

ThiefMaster:
I'm not sure what "Debian Bionic" is. AFAIK Bionic is an Ubuntu release. Could you post the output of `lsb_release -a`?

Ahmet-Kaplan:
Really sorry about that. It's Ubuntu bionic. 
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.2 LTS
Release:	18.04
Codename:	bionic

ThiefMaster:
Then they must have changed this between xenial and bionic... I did all testing with 16.04 when writing the docs... great, updating the docs for that won't be fun since debian most likely still has a different username.

ThiefMaster:
I just tested the install guide on the latest Ubuntu 18.04.2 and it works perfectly with with the `nginx` group, so I don't think there is any need to switch to `www-data`.

PR #3858, opened 2019-03-28
Add warning when editing confirmed pre-booking

### LINKED ISSUE #3831: Show warning when editing will revert state to pending
This happens when editing a booking for a room that requires confirmation. I will automatically revert to a pre-booking, which the user should be aware of before clicking "Save".

panagiotappl:
closes #3831 
example

### DIFF
--- a/indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
+++ b/indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
@@ -20,8 +20,8 @@ import moment from 'moment';
 import React from 'react';
 import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
-import {Form, Input, Segment, Select, TextArea} from 'semantic-ui-react';
-import {Field} from 'react-final-form';
+import {Form, Input, Message, Segment, Select, TextArea} from 'semantic-ui-react';
+import {Field, FormSpy} from 'react-final-form';
 import {START_DATE} from 'react-dates/constants';
 
 import PrincipalSearchField from 'indico/react/components/PrincipalSearchField';
@@ -195,7 +195,7 @@ class BookingEditForm extends React.Component {
     render() {
         const {
             user: sessionUser,
-            booking: {bookedForUser, startDt, endDt},
+            booking: {bookedForUser, startDt, endDt, room, isAccepted},
             onBookingPeriodChange,
             formProps,
             hideOptions,
@@ -278,6 +278,24 @@ class BookingEditForm extends React.Component {
                                disabled={submitSucceeded}
                                isEqual={_.isEqual}
                                render={this.renderTimeForm} />
+                    )}
+                    {!room.canUserBook && room.canUserPrebook && isAccepted && (
+                        <FormSpy subscription={{dirtyFields: true}}>
+                            {({dirtyFields}) => (
+                                Object.keys(dirtyFields).some(key => (
+                                    key === 'dates' || key === 'timeSlot' || key.startsWith('recurrence.')
+                                )) && (
+                                    <Message warning visible>
+                                        <Message.Header>
+                                            <Translate>This booking will require approval!</Translate>
+                                        </Message.Header>
+                                        <Translate>
+                                            Changing date or time will revert it back to a pre-booking.
+                                        </Translate>
+                                    </Message>
+                                )
+                            )}
+                        </FormSpy>
                     )}
                 </Segment>
                 <Segment color="blue" inverted>


indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -268,17 +275,31 @@ class BookingEditForm extends React.Component {
...
-                               onChange={(newTimeSlot) => onBookingPeriodChange(dates, newTimeSlot, recurrence)}
+                               onChange={(newTimeSlot) => {
+                                   this.setState({formChanged: true});
+                                   onBookingPeriodChange(dates, newTimeSlot, recurrence);
+                               }}
                                disabled={submitSucceeded}
                                isEqual={_.isEqual}
                                render={this.renderTimeForm} />
                     )}
+                    {formChanged && showWarning && (
```
ThiefMaster: You can use a FormSpy instead of tracking the state manually (which doesn't reset e.g. when you switch back to the old value). Something like this:

```jsx
{!room.canUserBook && room.canUserPrebook && (
    <FormSpy subscription={{dirtyFields: true}}>
        {({dirtyFields}) => (
            !!_.intersection(dirtyFields, ['recurrence.type', ...]).length && (
                <Message ...>...</Message>
            )
        )}
    </FormSpy>
)}
```

ThiefMaster: I just checked again and dirtyFields is a dict mapping field names to booleans and not just an array as I initially thought. But it never contains false values, so you could just throw in an `Object.keys()` there... Or better, add a util function that checks the relevant keys. Like this you could also use `startsWith('recurrence.')` instead of listing every `recurrence.*` name.

panagiotappl: I noticed that and added `Object.keys()`. I also noticed that the intersection should be between the `timeSlot` and `dates` and not with anything related to recurrences but when you switch from `single` to anything else the dates are also affected, thus the warning appears.

ThiefMaster: Are you sure the dates will become dirty if you change the recurrence in a way where the start and end date are still the same?

panagiotappl: I think it applies only when you go from single -> to something else since there is an `endDate` that is added there

indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -279,6 +279,20 @@ class BookingEditForm extends React.Component {
                                isEqual={_.isEqual}
                                render={this.renderTimeForm} />
                     )}
+                    {!room.canUserBook && room.canUserPrebook && isAccepted && (
+                        <FormSpy subscription={{dirtyFields: true}}>
+                            {({dirtyFields}) => (
+                                !!_.intersection(Object.keys(dirtyFields), ['timeSlot', 'dates']).length && (
+                                    <Message warning visible>
+                                        <Message.Header>
+                                                 This booking will require approval!
```
ThiefMaster: too much indentation, missing i18n

indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -279,6 +279,20 @@ class BookingEditForm extends React.Component {
...
+                                        </Message.Header>
+                                        <p>Changing date or time will revert it back to a pre-booking.</p>
```
ThiefMaster: missing i18n

indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -52,6 +52,7 @@ class BookingEditForm extends React.Component {
         hideOptions: {},
     };
 
+
```
mvidalgarcia: Is this intended? If so, why?

panagiotappl: I don't see a reason why that newline would be intended. :P Obviously a miss, thanks for pointing out!

indico/modules/rb_new/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -207,7 +207,6 @@ class BookingEditForm extends React.Component {
         const bookedByCurrentUser = sessionUser.id === bookedForUser.id;
         const today = moment();
         const bookingFinished = today.isAfter(startDt, 'day') && (endDt && today.isAfter(endDt, 'day'));
-
```
ThiefMaster: is this one intentional? ;)

panagiotappl: I never intend to change already existing spacing in parts of the code. I should be more careful when adding to git. Thanks!

ThiefMaster: `git add -p` ;)

panagiotappl: I know but sometimes when there's a bulk of code along with it you focus on the important parts and miss those little line adds/deletes. That's what I meant by more careful.

meluru: @panagiotappl I usually browse through my changes once more just before creating the PR. It's easy to spot unintended changes then.

panagiotappl: That's a good tip @meluru

PR #3861, opened 2019-03-29
Fix dropzone issue in abstract submission

### LINKED ISSUE #3572: Abstract submission broken if attachments are enabled and no file is provided
Seems like something is going wrong with the dropzone integration in that case. Submitting the abstract triggers the onbeforeunload prompt and when confirming it the form is submitted without AJAX so the user sees the JSON response.

This was working fine in 2.1, so it must be something we changed in the dropzone JS in 2.2

---

Any change there needs to be tested in various cases:

- Abstracts: without attachments enabled
- Abstracts: with attachments enabled, with & without a file selected
- Other dropzone uploads (attachments, event logo, ...)

mvidalgarcia:
closes #3572

This is a workaround that fixes the issue. I've tried hundreds of things but couldn't come up with something better :sweat: 

TODO:
- [x] Test other dropzone uploads (attachments, event logo, ...)

### DIFF
--- a/indico/web/client/js/jquery/utils/dropzone.js
+++ b/indico/web/client/js/jquery/utils/dropzone.js
@@ -99,9 +99,8 @@ import 'dropzone/dist/dropzone.css';
                             $dz.find('.dz-progress').show();
                             self.processQueue();
                         }
-                    } else {
-                        handlerRemoved = false;
                     }
+                    handlerRemoved = false;
                 });
 
                 self.on('addedfile', function(file) {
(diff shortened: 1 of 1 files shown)


indico/web/client/js/jquery/utils/dropzone.js
```diff
@@ -74,8 +77,15 @@ import 'dropzone/dist/dropzone.css';
                     }
                 }
 
-                // We'll first remove the default 'submit' handler
-                $form.off('submit').on('submit', function(e) {
+                $form.on('ajaxForm:validateBeforeSubmit', (e) => {
+                    let isAjaxDialog = !!$(e.target).parent('.exclusivePopup').length;
```
mic4ael: `const`?

ThiefMaster:
This is breaking the attachment upload in a somewhat subtle way:

- Upload a single file and it works fine.
- Then upload another file, and the progress dialog remains open.

PR #3862, opened 2019-04-01
Make submitter button color more distinct

### LINKED ISSUE #3731: "Submitter" switch is almost invisible
Many users have a hard time figuring out that it's there, thanks to low contrast and sub-standard screens.

image

panagiotappl:
closes #3731


mvidalgarcia:
This change is making the following style almost invisible
```css
&.selected:hover {
    color: $blue !important;
    border-color: $blue;
}
```

as `$indico-blue` and `$blue` are quite similar, should we change this rule so the user gets feedback when hovering?

PR #3863, opened 2019-04-01
Add location management

### LINKED ISSUE #3830: Admin Area: Manage locations - delete/edit/create
This is still unimplemented.

image

ThiefMaster:
closes #3830

### DIFF
--- a/indico/modules/rb_new/client/js/modules/admin/LocationPage.jsx
+++ b/indico/modules/rb_new/client/js/modules/admin/LocationPage.jsx
@@ -15,20 +15,47 @@
  * along with Indico; if not, see <http://www.gnu.org/licenses/>.
  */
 
+import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import {bindActionCreators} from 'redux';
 import {connect} from 'react-redux';
+import createDecorator from 'final-form-calculate';
 import {Field} from 'react-final-form';
 import {List} from 'semantic-ui-react';
 import {Param, Plural, PluralTranslate, Singular, Translate} from 'indico/react/i18n';
-import {formatters, validators as v, ReduxFormField} from 'indico/react/forms';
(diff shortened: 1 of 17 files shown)


indico/modules/rb_new/client/js/modules/admin/LocationPage.jsx
```diff
@@ -0,0 +1,178 @@
...
+
+        return (
+            <EditableList title={Translate.string('Locations')}
+                          addModalTitle={Translate.string('Add location')}
+                          isFetching={isFetching}
+                          items={locations}
+                          initialEditValues={snakifyKeys}
+                          initialAddValues={{
+                              room_name_format: '{building}/{floor}-{number}',
+                              map_url_template: 'https://www.google.com/maps/place/{lat},{lng}',
```
mvidalgarcia: Do we want to promote google maps? :trollface:
This could also be a placeholder

ThiefMaster: Placeholder wouldn't be great for the user since he'd have to type the URL. Having a link to google maps is the most convenient option for someone who doesn't want to bother setting up their own map etc. But we could also use OSM if they have a public site similar to google maps.

ThiefMaster: - OSM: https://www.openstreetmap.org/?mlat=46.23227762&mlon=6.05491571&zoom=18
- Google: https://www.google.com/maps/place/46.23227762,.05491571

mic4ael: +1 for OSM

ThiefMaster: One benefit of linking to Google would be that people who want to use it for GPS navigation are already in the right place.

ThiefMaster: Problem solved, now there's no default and you can choose between google, osm and a custom url

PR #3865, opened 2019-04-03
Optional gutter to timeline row label + overrides

mvidalgarcia:


indico/modules/rb_new/client/js/common/rooms/RoomDetailsModal.jsx
```diff
@@ -145,10 +145,15 @@ function RoomDetails({bookRoom, room, availability, attributes}) {
                 <Grid.Column>
                     <div>
                         <RoomBasicDetails room={room} />
-                        <RoomAvailabilityBox room={room} />
-                        <RoomCommentsBox room={room} />
-                        <RoomKeyLocation room={room} />
-                        <RoomCustomAttributesBox attributes={attributes} />
+                        <Overridable id="RoomDetails.infoBoxes"
+                                     room={room}>
```
ThiefMaster: that fits nicely in the previous line ;)

indico/modules/rb_new/client/js/common/rooms/RoomDetailsModal.jsx
```diff
@@ -145,10 +145,14 @@ function RoomDetails({bookRoom, room, availability, attributes}) {
                 <Grid.Column>
                     <div>
                         <RoomBasicDetails room={room} />
-                        <RoomAvailabilityBox room={room} />
-                        <RoomCommentsBox room={room} />
-                        <RoomKeyLocation room={room} />
-                        <RoomCustomAttributesBox attributes={attributes} />
+                        <Overridable id="RoomDetails.infoBoxes" room={room}>
+                            <>
```
mic4ael: `React.Fragment` should not be necessary here.

mvidalgarcia: Apparently it's necessary, otherwise it breaks and I get:
> Uncaught Invariant Violation: React.Children.only expected to receive a single React element child.

mic4ael: Ahh, right. It is handled by `Overridable`.

```javascript

const child = children ? React.Children.only(children) : null;

```

PR #3868, opened 2019-04-04
Prevent booking in the past

### LINKED ISSUE #3826: Prevent booking in the past
(Unless admin override is enabled)

`now - 1h` or something like that (could also be the current day) should probably never be considered "past". But then we should probably adapt checks like the one in `can_cancel` to also allow cancelling a booking e.g. within 30m of its creation (even if it's in the past) since right now you could end up creating a past booking which you cannot cancel even right after creating it.

mic4ael:
closes #3826

### DIFF
--- a/indico/modules/rb/models/reservation_occurrences.py
+++ b/indico/modules/rb/models/reservation_occurrences.py
@@ -30,6 +30,7 @@ from indico.core.db.sqlalchemy.util.queries import db_dates_overlap
 from indico.core.errors import IndicoError
 from indico.modules.rb.models.reservation_edit_logs import ReservationEditLog
 from indico.modules.rb.models.util import proxy_to_reservation_if_last_valid_occurrence
+from indico.modules.rb.util import rb_is_admin
 from indico.util import date_time
 from indico.util.date_time import format_date, iterdays
 from indico.util.serializer import Serializer
@@ -250,9 +251,17 @@ class ReservationOccurrence(db.Model, Serializer):
         return self.reservation.can_reject(user, allow_admin=allow_admin)
 
     def can_cancel(self, user, allow_admin=True):
+        if user is None:
+            return False
         if not self.is_valid or self.end_dt < datetime.now():
             return False
-        return self.reservation.can_cancel(user, allow_admin=allow_admin)
+        booking = self.reservation
+        booked_or_owned_by_user = booking.is_owned_by(user) or booking.is_booked_for(user)
+        if booking.is_rejected or booking.is_cancelled or booking.is_archived:
+            return False
+        if booked_or_owned_by_user and self.start_dt >= datetime.now():
+            return True
+        return allow_admin and rb_is_admin(user)
 
     @proxy_to_reservation_if_last_valid_occurrence
     @unify_user_args
--- a/indico/modules/rb/models/reservations.py
+++ b/indico/modules/rb/models/reservations.py
@@ -534,7 +534,7 @@ class Reservation(Serializer, db.Model):
         if self.is_rejected or self.is_cancelled or self.is_archived:
             return False
         is_booked_or_owned_by_user = self.is_owned_by(user) or self.is_booked_for(user)
-        is_valid_start_dt = is_booking_start_within_grace_period(self.start_dt)
+        is_valid_start_dt = is_booking_start_within_grace_period(self.start_dt, user)
         return (is_booked_or_owned_by_user and is_valid_start_dt) or (allow_admin and rb_is_admin(user))
 
     def can_edit(self, user, allow_admin=True):
--- a/indico/modules/rb/models/reservations.py
+++ b/indico/modules/rb/models/reservations.py
@@ -528,11 +528,14 @@ class Reservation(Serializer, db.Model):
         return self.room.can_moderate(user, allow_admin=allow_admin)
 
     def can_cancel(self, user, allow_admin=True):
+        from indico.modules.rb_new.util import is_booking_start_within_grace_period
         if user is None:
             return False
         if self.is_rejected or self.is_cancelled or self.is_archived:
             return False
-        return self.is_owned_by(user) or self.is_booked_for(user) or (allow_admin and rb_is_admin(user))
+        is_booked_or_owned_by_user = self.is_owned_by(user) or self.is_booked_for(user)
+        is_valid_start_dt = is_booking_start_within_grace_period(self.start_dt)
+        return (is_booked_or_owned_by_user and is_valid_start_dt) or (allow_admin and rb_is_admin(user))
 
     def can_edit(self, user, allow_admin=True):
         if user is None:
--- a/indico/modules/rb_new/client/js/components/BookingBootstrapForm.jsx
+++ b/indico/modules/rb_new/client/js/components/BookingBootstrapForm.jsx
@@ -25,7 +25,7 @@ import {SingleDatePicker, DateRangePicker} from 'indico/react/components';
 import {PluralTranslate, Translate} from 'indico/react/i18n';
 import {Overridable} from 'indico/react/util';
 import {
-    serializeDate, serializeTime, isBookingStartDtValid, createDt, isBookingStartDateValid
+    serializeDate, serializeTime, isBookingStartDTValid, createDT, isBookingStartDateValid
 } from 'indico/utils/date';
 import TimeRangePicker from './TimeRangePicker';
 import {selectors as userSelectors} from '../common/user';
@@ -176,7 +176,7 @@ class BookingBootstrapForm extends React.Component {
         } = this.state;
 
         const {buttonCaption, buttonDisabled, children, dayBased, hideOptions, isAdminOverrideEnabled} = this.props;
-        const isStartDtValid = isBookingStartDtValid(createDt(startDate, startTime), isAdminOverrideEnabled);
+        const isStartDtValid = isBookingStartDTValid(createDT(startDate, startTime), isAdminOverrideEnabled);
         const recurrenceOptions = [
             {text: PluralTranslate.string('Week', 'Weeks', number), value: 'week'},
             {text: PluralTranslate.string('Month', 'Months', number), value: 'month'}
--- a/indico/modules/rb_new/client/js/components/BookingBootstrapForm.jsx
+++ b/indico/modules/rb_new/client/js/components/BookingBootstrapForm.jsx
@@ -176,10 +176,7 @@ class BookingBootstrapForm extends React.Component {
         } = this.state;
 
         const {buttonCaption, buttonDisabled, children, dayBased, hideOptions, isAdminOverrideEnabled} = this.props;
-        const isStartDtValid = isBookingStartDtValid(
-            createDt(moment(startDate).format('YYYY-MM-DD'), startTime.format('HH:mm')),
-            isAdminOverrideEnabled
-        );
+        const isStartDtValid = isBookingStartDtValid(createDt(startDate, startTime), isAdminOverrideEnabled);
         const recurrenceOptions = [
             {text: PluralTranslate.string('Week', 'Weeks', number), value: 'week'},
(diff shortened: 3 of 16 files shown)


indico/modules/rb_new/client/js/modules/bookRoom/BookFromListModal.jsx
```diff
@@ -132,10 +143,36 @@ class BookFromListModal extends React.Component {
                         <Grid.Column width={8}>
                             <BookingBootstrapForm buttonCaption={isPrebooking ? labels.preBookBtn : labels.bookBtn}
                                                   buttonDisabled={buttonDisabled}
-                                                  onChange={refreshCollisions}
+                                                  onChange={(filters) => {
```
mvidalgarcia: What about wrapping this logic in a function?

indico/web/client/js/utils/date.js
```diff
@@ -55,3 +55,31 @@ export function dayRange(start, end, step = 1) {
...
+    return moment([...momentDate.toArray().splice(0, 3), ...momentTime.toArray().splice(3)]);
+}
+
+function isBookingStartValid(dt, isAdminOverrideEnabled = false, granularity = 'minute') {
+    if (!dt || !dt.isValid()) {
+        return false;
+    } else if (isAdminOverrideEnabled) {
+        return true;
+    }
+    const gracePeriod = granularity === 'day' ? moment() : moment().subtract(1, 'hour');
```
mvidalgarcia: :thinking: I wonder whether this grace period will be enough. I'm more inclined to be more flexible. Anyway we can wait if we get any complaint and modify it afterwards.

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -43,60 +43,39 @@ function _humanizeDuration(duration) {
...
+                serializedNext = serializeTime(momentNext);
+                options.push({key: serializedNext, value: serializedNext, text: serializedNext});
+            }
             next.add(30, 'm');
         }
         return options;
     };
 
-    generateEndTimeOptions = (start) => {
+    static generateEndTimeOptions = (start) => {
```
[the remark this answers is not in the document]
mic4ael: It is being used inside a static method so I can't have as a normal instance method. I could have it as a standalone function though.

indico/web/client/js/utils/date.js
mic4ael: I think that is a good compromise. Anyway, we will be able to easily change it later on if there is a need.

mic4ael:
> I figure that the `can_cancel` part stated in the issue description is not done?

Yeah, I started looking at other PRs and I will add the missing thing soon.

indico/modules/rb/models/reservations.py
```diff
@@ -528,11 +528,14 @@ def can_reject(self, user, allow_admin=True):
...
 
     def can_cancel(self, user, allow_admin=True):
+        from indico.modules.rb_new.util import is_booking_start_within_grace_period
         if user is None:
             return False
         if self.is_rejected or self.is_cancelled or self.is_archived:
             return False
-        return self.is_owned_by(user) or self.is_booked_for(user) or (allow_admin and rb_is_admin(user))
+        is_booked_or_owned_by_user = self.is_owned_by(user) or self.is_booked_for(user)
+        is_valid_start_dt = is_booking_start_within_grace_period(self.start_dt)
```
ThiefMaster: Do we fail gracefully if this returned true (when loading the booking data), but the grace period ended before the user actually cancelled?

indico/modules/rb_new/client/js/components/BookingBootstrapForm.jsx
```diff
@@ -170,7 +175,8 @@ class BookingBootstrapForm extends React.Component {
             dates: {startDate, endDate}
         } = this.state;
 
-        const {buttonCaption, buttonDisabled, children, dayBased, hideOptions} = this.props;
+        const {buttonCaption, buttonDisabled, children, dayBased, hideOptions, isAdminOverrideEnabled} = this.props;
+        const isStartDtValid = isBookingStartDtValid(createDt(startDate, startTime), isAdminOverrideEnabled);
```
ThiefMaster: s/Dt/DT/?

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -41,29 +41,65 @@ function _humanizeDuration(duration) {
     }
 }
 
+function generateStartTimeOptions(allowPastTimes) {
+    const options = [];
+    const end = moment().endOf('day');
+    const next = moment(START_HOUR, 'HH:mm');
+    let serializedNext;
```
ThiefMaster: why not just make it const and declare it inside the if when assigning it?

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -78,42 +114,18 @@ export default class TimeRangePicker extends React.Component {
         };
     }
 
+    static getDerivedStateFromProps({allowPastTimes, startTime}, prevState) {
```
ThiefMaster: ugh can't we avoid using derived state here? :/

unless the options are expensive to generate i guess we could just generate them when needed during rendering?

indico/modules/rb_new/client/js/modules/bookRoom/actions.js
```diff
@@ -77,8 +77,9 @@ export const FETCH_RELATED_EVENTS_ERROR = 'bookRoom/FETCH_RELATED_EVENTS_ERROR';
 export const RELATED_EVENTS_RECEIVED = 'bookRoom/RELATED_EVENTS_RECEIVED';
 export const RESET_RELATED_EVENTS = 'bookRoom/RESET_RELATED_EVENTS';
 
-export function createBooking(args) {
+export function createBooking(args, isAdminOverrideEnabled) {
     const params = preProcessParameters(args, ajaxRules);
+    params.admin_override_enabled = isAdminOverrideEnabled;
```
ThiefMaster: only add this param if it's enabled

indico/modules/rb_new/util.py
```diff
@@ -104,3 +105,9 @@ def get_linked_object(type_, id_):
                         SessionBlock.session.has(db.and_(~Session.is_deleted,
                                                          Session.event.has(is_deleted=False))))
                 .first())
+
+
+def is_booking_start_within_grace_period(start_dt, allow_admin=False):
+    if allow_admin:
```
ThiefMaster: :warning: you aren't checking if the user actually IS an admin

mic4ael: Ahh, right!

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
mic4ael: Parts of this function were copy-pasted. But I will take a look.

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
mic4ael: It was because of the fact that options are stored in the state. But now, when I looked at it closely I might be able to get rid of that.

indico/modules/rb/models/reservations.py
mic4ael: It will throw a `Forbidden` error right now.

indico/modules/rb/models/reservation_occurrences.py
```diff
@@ -250,9 +251,17 @@ def can_reject(self, user, allow_admin=True):
         return self.reservation.can_reject(user, allow_admin=allow_admin)
 
     def can_cancel(self, user, allow_admin=True):
+        if user is None:
+            return False
         if not self.is_valid or self.end_dt < datetime.now():
             return False
-        return self.reservation.can_cancel(user, allow_admin=allow_admin)
+        resv = self.reservation
```
ThiefMaster: i think we usually use `booking` in the new code

indico/modules/rb/models/reservation_occurrences.py
```diff
@@ -250,9 +251,17 @@ def can_reject(self, user, allow_admin=True):
...
+        booked_or_owned_by_user = resv.is_owned_by(user) or resv.is_booked_for(user)
+        if booked_or_owned_by_user and self.start_dt >= datetime.now():
+            return True
+        if resv.is_rejected or resv.is_cancelled or resv.is_archived:
+            return False
+        return booked_or_owned_by_user or allow_admin and rb_is_admin(user)
```
ThiefMaster: parentheses!

indico/modules/rb_new/client/js/components/TimeRangePicker.jsx
```diff
@@ -178,27 +181,37 @@ export default class TimeRangePicker extends React.Component {
...
                 }
+            } else {
+                duration = moment.duration(end.diff(start));
             }
         } else {
-            duration = moment.duration(end.diff(start));
+            this.setState({
+                endSearchQuery: serializeTime(previousEndTime)
+            });
+            return;
```
ThiefMaster: I would negate the `if` condition. the current flow seems pretty strange to me:

```
if (x) {
    foo;
} else {
    bar;
    return;
}
more foo;
```

indico/modules/rb_new/client/js/modules/bookRoom/BookFromListModal.jsx
```diff
@@ -105,6 +108,10 @@ class BookFromListModal extends React.Component {
         }
     };
 
+    state = {
+        pastDtChosen: false,
```
ThiefMaster: `DT`

indico/modules/rb_new/controllers/backend/bookings.py
```diff
@@ -207,6 +208,9 @@ def _link_booking(self, booking, type_, id_, link_back):
 
     def _process(self):
         args = self.args
+        if not is_booking_start_within_grace_period(args['start_dt'], session.user, args['admin_override_enabled']):
+            raise ExpectedError(_('You cannot create a booking with start date in the past'))
```
ThiefMaster: "which starts in the past" or "with **a** start date in the past"

indico/modules/rb_new/util.py
```diff
@@ -104,3 +106,9 @@ def get_linked_object(type_, id_):
                         SessionBlock.session.has(db.and_(~Session.is_deleted,
                                                          Session.event.has(is_deleted=False))))
                 .first())
+
+
+def is_booking_start_within_grace_period(start_dt, user, allow_admin=False):
+    if allow_admin and rb_is_admin(user):
+        return True
+    return start_dt >= datetime.now() - timedelta(hours=1)
```
ThiefMaster: this will break if the DEFAULT_TIMEZONE does not match the server's timezone. use utc and convert to the server's tz

indico/modules/rb/models/reservation_occurrences.py
mic4ael: Yeah, but the `and` operator has higher precedence anyway ;P

ThiefMaster: Yes, it's well-defined but knowing the and/or precedence is something I consider "stuff you should not NEED to know" ;)

mic4ael:
Updated.

indico/modules/rb/models/reservation_occurrences.py
```diff
@@ -250,9 +251,17 @@ def can_reject(self, user, allow_admin=True):
         return self.reservation.can_reject(user, allow_admin=allow_admin)
 
     def can_cancel(self, user, allow_admin=True):
```
ThiefMaster: Shouldn't this also take the grace period into account? Also, below you return True in some cases without checking the false cases first (you can never cancel an occurrence if the whole booking is already cancelled). FWIW, cancelling the whole booking cancels all occurrences, so this isn't really an issue, but I think we should still change the order of those checks.

PR #3882, opened 2019-04-12
Add PrincipalField & use it for room owner change

ThiefMaster:
This is based on the same code already used for PrincipalListField.

image

### DIFF
--- a/indico/web/client/js/react/components/principals/PrincipalField.jsx
+++ b/indico/web/client/js/react/components/principals/PrincipalField.jsx
@@ -32,7 +32,7 @@ import './items.module.scss';
  * A field that lets the user select a user.
  */
 const PrincipalField = (props) => {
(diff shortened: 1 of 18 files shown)


indico/web/client/js/react/components/principals/PrincipalField.jsx
```diff
@@ -0,0 +1,137 @@
...
+                    onAddItems={handleAddItem}
+                    onClose={markTouched}
+                    favorites={favoriteUsers}
+                    disabled={disabled}
+                    single />
+    );
+
+    return (
+        <div className="ui input">
+            <div className="fake-input">
```
ThiefMaster: FYI this class is defined in indico-sui-theme since it needs access to the SUI variables. It applies the same styling any other `<input>` would get.

PR #3885, opened 2019-04-15
Fix possible invalid time

mvidalgarcia:


indico/web/client/js/utils/date.js
```diff
@@ -83,3 +83,8 @@ export function isBookingStartDateValid(date, isAdminOverrideEnabled) {
 export function isBookingStartDTValid(dt, isAdminOverrideEnabled) {
     return isBookingStartValid(dt, isAdminOverrideEnabled);
 }
+
+export function initialEndTime(end) {
+    const endOfDay = moment().endOf('day');
+    return end > endOfDay ? endOfDay : end;
```
meluru: is it okay to use `>` when comparing two Moment objects? I always use `isBefore` etc...

mvidalgarcia: Good question, it seems to work. I based my logic on [this](https://github.com/indico/indico/blob/master/indico/modules/rb_new/client/js/components/TimeRangePicker.jsx#L145). But maybe it's cleaner `isBefore`..

ThiefMaster: If `<` works then I think that's better since it's more straightforward.
JS uses `.valueOf()` for comparison, so if that returns something suitable for comparison, then the usual operators should work.

mvidalgarcia: I remember fixing something related to this.., yeah [this](https://github.com/indico/indico/commit/504424c28ff793ce4f8c720844b8dd1802f1cd44#diff-5ac3fca07d60192bf271d60e991b5f45). But I guess that for full timestamp comparison it's ok to use operators.

mvidalgarcia:
So far it seems to work fine with Burotel. I need to test further once @mic4ael pushes his fixes.

mic4ael:
> So far it seems to work fine with Burotel. I need to test further once @mic4ael pushes his fixes.

https://github.com/indico/indico/pull/3890

PR #3890, opened 2019-04-15
Make booking grace period configurable

mic4ael:
fixes indico/indico-plugins-cern#49

### DIFF
--- a/indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
+++ b/indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
@@ -248,10 +248,8 @@ const SettingsPage = props => {
                                component={ReduxFormField}
                                as="input"
                                type="number"
-                               min="0"
                                max="24"
-                               required
-                               parse={value => (value ? +value : null)}
+                               parse={value => (value.length !== 0 ? +value : null)}
                                label={Translate.string('Grace period')}>
                             <p className="field-description">
                                 <Translate>
--- a/indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
+++ b/indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
@@ -248,13 +248,16 @@ const SettingsPage = props => {
                                component={ReduxFormField}
                                as="input"
                                type="number"
+                               min="0"
                                max="24"
-                               parse={value => (value.length !== 0 ? +value : null)}
+                               parse={value => (value !== '' ? +value : null)}
                                label={Translate.string('Grace period')}>
                             <p className="field-description">
                                 <Translate>
                                     Usually booking a space in the past is not allowed. This setting will allow to
                                     book a room with a start date within a specified number of hours in the past.
+                                    Leaving the field empty will allow any start date that is not in the past
+                                    without restricting the start time as well.
                                 </Translate>
                             </p>
                         </Field>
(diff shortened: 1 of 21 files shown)


indico/modules/rb_new/controllers/backend/admin.py
```diff
@@ -70,6 +71,7 @@ class SettingsSchema(mm.Schema):
     end_notification_weekly = fields.Int(validate=[validate.Range(min=1, max=30)])
     end_notification_monthly = fields.Int(validate=[validate.Range(min=1, max=30)])
     excluded_categories = ModelList(Category)
+    grace_period = fields.Int()
```
ThiefMaster: `validate.Range(min=0)` (and if we still reject past dates: `max=24`)

indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
```diff
@@ -244,6 +244,20 @@ const SettingsPage = props => {
                                 </Translate>
                             </p>
                         </Field>
+                        <Field name="grace_period"
+                               component={ReduxFormField}
+                               as="input"
+                               type="number"
+                               max="24"
+                               parse={value => (value ? +value : null)}
```
ThiefMaster: This will also convert 0 to null. Also, the description below should mention the special meaning of leaving it empty.

mic4ael: Ah, right. It is not needed now since we will allow empty values now.

indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
```diff
@@ -244,6 +244,20 @@ const SettingsPage = props => {
                                 </Translate>
                             </p>
                         </Field>
+                        <Field name="grace_period"
+                               component={ReduxFormField}
+                               as="input"
+                               type="number"
+                               max="24"
```
ThiefMaster: `min` value is missing

indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
```diff
@@ -244,6 +244,20 @@ const SettingsPage = props => {
...
+                               parse={value => (value.length !== 0 ? +value : null)}
```
ThiefMaster: `value !== ''`?

indico/modules/rb_new/client/js/modules/admin/SettingsPage.jsx
```diff
@@ -244,6 +244,20 @@ const SettingsPage = props => {
...
+                               label={Translate.string('Grace period')}>
+                            <p className="field-description">
+                                <Translate>
+                                    Usually booking a space in the past is not allowed. This setting will allow to
+                                    book a room with a start date within a specified number of hours in the past.
```
ThiefMaster: doesn't explain what "no value" means

indico/modules/rb_new/controllers/backend/admin.py
```diff
@@ -69,6 +70,7 @@ class SettingsSchema(mm.Schema):
     end_notification_weekly = fields.Int(validate=[validate.Range(min=1, max=30)])
     end_notification_monthly = fields.Int(validate=[validate.Range(min=1, max=30)])
     excluded_categories = ModelList(Category)
+    grace_period = fields.Int(missing=None, validate=[validate.Range(max=24)])
```
ThiefMaster: - `missing=None` is wrong here - you now always set the grace period to null when saving something else and not touching this field at all!
- missing `min` in the range
- probably needs `allow_none=True`

PR #3894, opened 2019-04-16
Support external users in "booked for"

ThiefMaster:


indico/modules/users/controllers.py
```diff
@@ -522,18 +525,83 @@ def _process(self):
...
+
+search_result_schema = UserSearchResultSchema()
+
+
 class RHUserSearch(RHProtected):
     """Search for users based on given criteria"""
 
-    @use_args({
+    def _serialize_pending_user(self, entry):
+        first_name = entry.data.get('first_name') or ''
```
mic4ael: why not `entry.data.get('first_name', '')`?

indico/util/user.py
```diff
@@ -129,6 +129,23 @@ def principal_from_identifier(identifier, allow_groups=False):
...
+            raise ValueError('External users are not allowed')
+        cache = GenericCache('external-user')
+        external_user_data = cache.get(data)
+        if not external_user_data:
+            raise ValueError('Invalid data')
+        user = User.query.filter(User.all_emails == external_user_data['email'], ~User.is_deleted).first()
+        if user:
+            return user
+        # create a pending user. this user isn't sent to the DB unless it gets added
+        # to the sqlalchemy session somehow (e.g. by using it to an ACL).
```
mic4ael: Did you mean "by adding it to an ACL"?

indico/web/client/js/react/components/principals/PrincipalListField.jsx
```diff
@@ -100,6 +102,7 @@ const PrincipalListField = (props) => {
                                        name={data.name}
                                        detail={data.detail}
                                        isGroup={data.group}
+                                       isPendingUser={!data.group && data.userId === null}
```
mic4ael: Isn't it enough to check if `data.userId === null`?

indico/modules/users/controllers.py
ThiefMaster: That won't cover the value being `None` in the dict

indico/web/client/js/react/components/principals/PrincipalListField.jsx
ThiefMaster: Yes, but that's just because `null !== undefined` and I really don't want to depend on that :)

indico/util/user.py
ThiefMaster: yep, "using it in an ACL" / "adding it to an ACL"

PR #3896, opened 2019-04-17
Allow cancelling future booking occurrences

### LINKED ISSUE #3886: Allow cancelling a booking with future occurrences
Currently we cannot cancel a booking once it started. This means if you have 1 past occurrence and 99 future occurrences you have no way to cancel that booking.

The best option is probably to just cancel future occurrences when cancelling a booking, and changing `can_cancel` to return true unless the end time is in the past instead of checking the start time.

mvidalgarcia:
closes #3886

TODO:
- [x] Refresh occurrences in booking details'  'See on timeline'

### DIFF
--- a/indico/modules/rb/models/reservation_occurrences.py
+++ b/indico/modules/rb/models/reservation_occurrences.py
@@ -14,7 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Indico; if not, see <http://www.gnu.org/licenses/>.
 
-from datetime import datetime
+from datetime import datetime, timedelta
 from math import ceil
 
 from dateutil import rrule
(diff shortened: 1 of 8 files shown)


meluru:
Shouldn't the grace period be the same as for booking creation? Now it seems a bit confusing that I can create a booking an hour ago but I can't cancel it.

Also, maybe you could update the tests that are failing now?

mvidalgarcia:
@meluru We've already discussed this. Problem of having 1h grace period for cancellations is that you can accidentally cancel short past recent occurrences e.g. you have a recurring booking with occurrences (meetings) of 20m. If right after finishing a meeting you decide to cancel the booking (because you want to cancel future occurrences) you'd end up cancelling the meeting you just had which is a valid one. Said that, I'm open to different approaches.

Concerning the tests, I'm working on it, thanks for mentioning!

ThiefMaster:
What about adding an `is_within_cancel_grace_period` hybrid property to ReservationOccurrence? That way you don't have to specify it in so many places.

PR #3899, opened 2019-04-18
Add day of week in reservation reminders

### LINKED ISSUE #2982: Room Booking: Reservation reminder - add day of week
Can you please add day of week next to the date, for room reservation reminder e-mails?

panagiotappl:
closes #2982

### DIFF
--- a/indico/modules/rb/templates/emails/reservations/reminders/finishing_bookings.html
+++ b/indico/modules/rb/templates/emails/reservations/reminders/finishing_bookings.html
@@ -29,7 +29,7 @@
 
 
 {% macro _render_dates(reservation) -%}
-    <strong>{{ reservation.start_dt|format_datetime(locale='en_GB') }}</strong> - <strong>{{ reservation.end_dt|format_datetime(locale='en_GB') }}</strong>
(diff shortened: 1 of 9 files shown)


indico/modules/rb/templates/emails/reservations/reminders/finishing_bookings.html
```diff
@@ -29,7 +29,7 @@
 
 
 {% macro _render_dates(reservation) -%}
-    <strong>{{ reservation.start_dt|format_datetime(locale='en_GB') }}</strong> - <strong>{{ reservation.end_dt|format_datetime(locale='en_GB') }}</strong>
+    <strong>{{ reservation.start_dt|format_datetime('EEEE dd/MM/yyyy', locale='en_GB') }}</strong> - <strong>{{ reservation.end_dt|format_datetime('EEEE dd/MM/yyyy', locale='en_GB') }}</strong>
```
ThiefMaster: i'm pretty sure we lose information here since the format you have only includes the data but previously we had the time as well

panagiotappl: You're right. Also, do you think we should include the full name of the day or just the first three letters (eg. Tue)?

PR #3900, opened 2019-04-18
Dynamic timeline legend 

### LINKED ISSUE #3824: Create dynamic timeline legend
Currently the timeline legend is predefined and sometimes shows items that are not displayed on the timeline.

E.g.

image

meluru:
Closes #3824

### DIFF
--- a/indico/modules/rb_new/client/js/common/bookings/BookingDetails.jsx
+++ b/indico/modules/rb_new/client/js/common/bookings/BookingDetails.jsx
@@ -295,8 +295,9 @@ class BookingDetails extends React.Component {
     };
 
     getLegendLabels = (availability) => {
+        const inactiveTypes = ['blockings', 'overridableBlockings', 'nonbookablePeriods', 'unbookableHours'];
         const occurrenceTypes = getOccurrenceTypes(availability);
-        return transformToLegendLabels(occurrenceTypes);
+        return transformToLegendLabels(occurrenceTypes, inactiveTypes);
     };
 
     renderActionButtons = (canCancel, canReject, showAccept) => {
--- a/indico/modules/rb_new/client/js/common/bookings/BookingDetails.jsx
+++ b/indico/modules/rb_new/client/js/common/bookings/BookingDetails.jsx
@@ -302,7 +302,7 @@ class BookingDetails extends React.Component {
                 return {label: Translate.string('Cancelled'), style: 'cancellation'};
             case 'rejections':
                 return {label: Translate.string('Rejected'), style: 'rejection'};
-            case 'otherBookings':
+            case 'other':
                 return {label: Translate.string('Other bookings'), style: 'other'};
             default:
                 return undefined;
--- a/indico/modules/rb_new/client/js/common/bookings/BookingDetails.jsx
+++ b/indico/modules/rb_new/client/js/common/bookings/BookingDetails.jsx
@@ -297,13 +297,13 @@ class BookingDetails extends React.Component {
     transformToLabel = (type) => {
         switch (type) {
             case 'bookings':
-                return {label: Translate.string('Current'), style: 'booking'};
+                return {label: Translate.string('Current'), style: 'booking', order: 1};
             case 'cancellations':
-                return {label: Translate.string('Cancelled'), style: 'cancellation'};
+                return {label: Translate.string('Cancelled'), style: 'cancellation', order: 2};
             case 'rejections':
-                return {label: Translate.string('Rejected'), style: 'rejection'};
+                return {label: Translate.string('Rejected'), style: 'rejection', order: 3};
             case 'other':
-                return {label: Translate.string('Other bookings'), style: 'other'};
+                return {label: Translate.string('Other bookings'), style: 'other', order: 4};
             default:
                 return undefined;
         }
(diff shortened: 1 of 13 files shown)


indico/modules/rb_new/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -294,6 +294,27 @@ class BookingDetails extends React.Component {
...
+            default: return undefined;
+        }
+    };
+
+    getLegendLabels = (availability) => {
+        const legendLabels = [];
+        Object.entries(availability).forEach(([type, occurrences]) => {
+            if (occurrences && Object.keys(occurrences).length > 0) {
+                const label = this.transformToLabel(type);
+                label && !legendLabels.some(lab => _.isEqual(lab, label)) && legendLabels.push(label);
```
ThiefMaster: use a normal `if ()` statement here. this kind of "optimization" is something valid if you are a script that's minifying JS, but not something that should be in original source code ;)

indico/modules/rb_new/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -294,6 +294,27 @@ class BookingDetails extends React.Component {
         });
     };
 
+    transformToLabel = (type) => {
+        switch (type) {
+            case 'bookings': return {label: Translate.string('Current'), style: 'booking'};
```
ThiefMaster: what about a linebreak after `:`?

indico/modules/rb_new/client/js/common/rooms/RoomDetailsModal.jsx
```diff
@@ -117,15 +118,32 @@ export default connect(
...
-        {label: Translate.string('Blocked'), style: 'blocking'},
-        {label: Translate.string('Not bookable'), style: 'unbookable'}
-    ];
+const _getLegendLabels = (availability) => {
+    const legendLabels = [];
+    availability.forEach((day) => {
+        Object.entries(day).forEach(([type, occurrences]) => {
+            if (occurrences && Object.keys(occurrences).length > 0) {
+                const label = _transformToLabel(type);
+                label && !legendLabels.some(lab => _.isEqual(lab, label)) && legendLabels.push(label);
```
ThiefMaster: same as above

indico/modules/rb_new/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -294,6 +294,34 @@ class BookingDetails extends React.Component {
         });
     };
 
+    transformToLabel = (type) => {
+        switch (type) {
```
pferreir: A pity that JS doesn't have native pattern-matching 😢

indico/modules/rb_new/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -294,6 +294,34 @@ class BookingDetails extends React.Component {
...
+            case 'other':
+                return {label: Translate.string('Other bookings'), style: 'other', order: 4};
+            default:
+                return undefined;
+        }
+    };
+
+    getLegendLabels = (availability) => {
+        const legendLabels = [];
+        Object.entries(availability).forEach(([type, occurrences]) => {
```
pferreir: An alternative that would be probably faster... (?)
```js
const legendLabels = Object.entries(availability).reduce((accum, [type, occurrences]) => {
    //...
   return accum;
}, {});
```
... and just set `accum[label] = true` if there's a match. The fact that we'd use an object would make the membership check O(1) instead of O(N).
It's not a big deal since the set of labels will always be bound to be small. So, maybe not worth the effort...

indico/modules/rb_new/client/js/common/bookings/BookingEditCalendar.jsx
```diff
@@ -44,19 +45,52 @@ class BookingEditCalendar extends React.Component {
...
+                return {label: Translate.string('Not bookable'), style: 'unbookable', order: 11};
+            default:
+                return undefined;
+        }
+    };
+
+    getLegendLabels = (calendars) => {
+        const legendLabels = [];
+        calendars.forEach((calendar) => {
+            calendar.forEach(({availability}) => {
```
pferreir: This seems very similar to [some other bit](https://github.com/indico/indico/pull/3900/files#diff-ba6397d7aaad7dd18fb13f6c1a947fe3R314) as well as many others after. Maybe it can be moved out to its own function?

indico/modules/rb_new/client/js/modules/bookRoom/BookRoom.jsx
```diff
@@ -157,6 +158,47 @@ class BookRoom extends React.Component {
         openBookingForm(roomId, {...bookingData, isPrebooking});
     }
 
+    transformToLabel(type) {
+        switch (type) {
```
pferreir: By the way, since the translation is always the same anyway, why not split these exhaustive switches in two declarative structures?
 * An object that maps `style -> translation`
 * A list of style IDs. e.g. `['available', 'booking', 'pre-booking', ...]` (no need for `order`, since the list is already ordered anyway)

Each widget would need to have its own version of the second, but they could all share a common dictionary of translations.

meluru:
@pferreir I'll give it a try, I've considered it before, but legends seemed to be too different and I saw some problems on the way, but maybe if I simplify some of the current ones it will work out

meluru:
Two things that come with those changes:
1. I've unified the label translations - now every label that used to be called `Current booking`, `Booked` etc is called simply `Booking`
2. In every component we define which labels we will use and in which order

PR #3905, opened 2019-04-24
Travis CI: Start testing on Python 3 in allow_failures mode

cclauss:
Another attempt at #3834:


ThiefMaster:
Thanks, but we'll add this when we start supporting Python 3. Right now it's a waste of resources since we know things will fail.

PR #3910, opened 2019-04-26
Theme improvements

pferreir:
This is meant to pave the way for https://github.com/indico/indico-plugins-cern/pull/50.

Some of the decisions I took:
 * Passing `theme_settings` around instead of overloading the function calls with many arguments;
 * Creating `theme_context` so that the macros generating the timetable can have a shared state (once again, the alternative would have been even more argument passing);
 * Adding `hide_duration`, `hide_session_block_time` and `number_contributions` to `theme_settings`, which will be used by https://github.com/indico/indico-plugins-cern/pull/50;

I am not very sure about the approach when it comes to creating and updating the `theme_context`. The alternatives are not great, but a possibility would be a global function that just does a dict update. Ideas welcome!

### DIFF
--- a/indico/modules/attachments/templates/_display.html
+++ b/indico/modules/attachments/templates/_display.html
@@ -37,7 +37,7 @@
     {% set folders = item.attached_items.folders if item else folders %}
 
     <div class="attachments-display-container toolbar">
-        <div class="folder folder-root">
+        <div class="folder folder-root {{ 'folder-root-empty' if not files }}">
             {% for attachment in files %}
                 {{ render_attachment(attachment, classes='i-button') }}
                 {{ template_hook('event-display-after-attachment', attachment=attachment, top_level=true, has_label=false) }}
--- a/indico/modules/events/timetable/templates/display/indico/_contribution.html
+++ b/indico/modules/events/timetable/templates/display/indico/_contribution.html
@@ -4,7 +4,7 @@
 {% from 'events/timetable/display/indico/_common.html' import render_speakers, render_references, render_attachments,
                                                               render_notes, render_description, render_time %}
 
-{% macro render_contribution(contrib, event, parent=none, nested=false, hide_end_time=false, timezone=none,
+{% macro render_contribution(contrib, event, theme_settings, parent=none, nested=false, hide_end_time=false, timezone=none,
                              show_notes=false, show_location=false) -%}
     {% set anchor = slugify(contrib.friendly_id, contrib.title, maxlen=30) %}
     <li class="timetable-item timetable-contrib" id="{{ anchor }}">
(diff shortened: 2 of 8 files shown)


ThiefMaster:
You can use a namespace object instead of the dict: http://jinja.pocoo.org/docs/2.10/templates/#namespace

That lets you use `{% set foo.bar = ... %}` instead of the hack with calling `update()`

pferreir:
Applied @ThiefMaster's suggestion (thanks!)

indico/modules/attachments/templates/_display.html
```diff
@@ -37,7 +37,7 @@
     {% set folders = item.attached_items.folders if item else folders %}
 
     <div class="attachments-display-container toolbar">
-        <div class="folder folder-root">
+        <div class="folder folder-root {{ 'folder-root-empty' if not files|length }}">
```
ThiefMaster: shouldn't `... if not files` be enough here?

indico/modules/events/timetable/templates/display/indico/_subcontribution.html
```diff
@@ -3,16 +3,20 @@
...
-{% macro render_subcontribution(subcontrib, event, show_notes=true) %}
+{% macro render_subcontribution(subcontrib, event, theme_settings, theme_context, show_notes=true) %}
     {% set contrib = subcontrib.contribution %}
     {% set anchor = slugify('sc', contrib.friendly_id, subcontrib.friendly_id, subcontrib.title, maxlen=30) %}
     <li class="timetable-subcontrib timetable-item" id="{{ anchor }}">
         <div class="timetable-item-body flexcol">
             <div class="timetable-item-header flexrow">
                 <span class="timetable-title" data-anchor="{{ anchor }}">
+                    {% if theme_settings.number_contributions %}
+                        {{- theme_context.num_contribution}}.
```
ThiefMaster: missing space before `}}`

indico/modules/events/timetable/templates/display/indico/meeting.html
```diff
@@ -3,6 +3,9 @@
 {% from 'events/timetable/display/indico/_break.html' import render_break %}
 
 {% if entries %}
+    {% set theme_context = namespace({
```
ThiefMaster: `namespace(num_contribution=0)` works as well

indico/modules/events/timetable/templates/display/indico/_contribution.html
```diff
@@ -4,20 +4,26 @@
 {% from 'events/timetable/display/indico/_common.html' import render_speakers, render_references, render_attachments,
                                                               render_notes, render_description, render_time %}
 
-{% macro render_contribution(contrib, event, parent=none, nested=false, hide_end_time=false, timezone=none,
-                             show_notes=false, show_location=false) -%}
+{% macro render_contribution(contrib, event, theme_settings, theme_context, parent=none, nested=false, hide_end_time=false,
```
ThiefMaster: I think the administrative theme calls this macro so it probably needs to be updated.

pferreir: IIRC, the admin theme has its own macros.

ThiefMaster: indeed, nvm

PR #3912, opened 2019-04-29
Support readOnly prop in some fields

### LINKED ISSUE #3908: Approving/Rejecting blockings is broken
This is because we are abusing disabled fields for the 'view' mode, but we now set disabled on the `<Form.Field>` when a field is disabled, so all mouse events are disabled (and an opacity is applied).

Two possible solutions:

1) add a `disableFieldWrapper={false}` prop so no `disabled` prop is set on the `<Form.Field>` even if the field is disabled
2) Do not disable the fields at all but set a `readOnly` prop (and implement it correctly in the custom fields used in the blocking modal)

I think (2) is the cleaner solution since disabled fields are usually just disabled, and don't suddenly hide the UI related to changing things...

mic4ael:
fixes #3908

### DIFF
--- a/indico/modules/rb_new/client/js/modules/blockings/BlockingModal.jsx
+++ b/indico/modules/rb_new/client/js/modules/blockings/BlockingModal.jsx
@@ -233,7 +233,7 @@ class BlockingModal extends React.Component {
         return (
             <Button type="submit"
                     form="blocking-form"
-                    disabled={hasValidationErrors || pristine || submitSucceeded}
+                    disabled={submitting || hasValidationErrors || pristine || submitSucceeded}
                     loading={submitting}
                     primary>
                 {mode === 'edit' ? (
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Let's only disable elements and not hide them in disabled mode; for "view" mode there's readOnly now, and it's a bit weird if the UI changes (beyond disabling everything) while submitting.

indico/modules/rb_new/client/js/modules/blockings/BlockingModal.jsx
```diff
@@ -381,14 +381,13 @@ class BlockingModal extends React.Component {
                                                favoriteUsersController={favoriteUsersController}
                                                isEqual={(a, b) => _.isEqual(a.sort(), b.sort())}
                                                label={Translate.string('Authorized users/groups')}
-                                               readOnly={mode === 'view'}
-                                               disabled={mode === 'view' || submitSucceeded} />
+                                               readOnly={mode === 'view' || submitSucceeded} />
```
ThiefMaster: I guess submitSucceeded should keep setting disabled disabled?

PR #3917, opened 2019-05-06
Include IE11 as outdated browser

### LINKED ISSUE #3914: Include IE11 in outdatedBrowser warnings
Also make sure those warnings show up in the new room booking module! I don't think we currently include it there at all.

We might need to find a new library for this as the one we are using seems to be broken (`lowerThan: 'Edge'` triggers in Firefox)

mvidalgarcia:
closes #3914 

`'object-fit'` works as `'Edge'` (https://github.com/burocratik/outdated-browser/issues/198#issuecomment-431383916)

### DIFF
--- a/indico/modules/rb_new/views/base.py
+++ b/indico/modules/rb_new/views/base.py
@@ -25,7 +25,7 @@ class WPRoomBookingBase(WPNewBase):
     template_prefix = 'rb/'
     title = _('Room Booking')
     bundles = ('common.js', 'common.css', 'react.js', 'react.css', 'semantic-ui.js', 'semantic-ui.css',
-               'module_rb_new.js', 'module_rb_new.css')
+               'module_rb_new.js', 'module_rb_new.css', 'outdatedbrowser.js', 'outdatedbrowser.css')
 
 
 class WPEventBookingList(WPEventManagement):
--- a/indico/modules/rb_new/views/base.py
+++ b/indico/modules/rb_new/views/base.py
@@ -25,7 +25,7 @@ class WPRoomBookingBase(WPNewBase):
     template_prefix = 'rb/'
     title = _('Room Booking')
     bundles = ('common.js', 'common.css', 'react.js', 'react.css', 'semantic-ui.js', 'semantic-ui.css',
-               'module_rb_new.js', 'module_rb_new.css', 'outdatedbrowser.js', 'outdatedbrowser.css')
+               'module_rb_new.js', 'module_rb_new.css')
 
 
 class WPEventBookingList(WPEventManagement):
--- a/indico/web/templates/_outdated_browser.html
+++ b/indico/web/templates/_outdated_browser.html
@@ -6,12 +6,12 @@
 </div>
 
 <script>
-    $(document).ready(function() {
+    window.addEventListener('load', function() {
         outdatedBrowser({
             bgColor: '#0b63a5',
             color: '#ffffff',
-            lowerThan: 'IE11',
+            lowerThan: 'object-fit', // Edge
             languagePath: ''
-        });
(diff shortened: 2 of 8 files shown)


indico/web/templates/_outdated_browser.html
```diff
@@ -6,12 +6,27 @@ <h6>Your browser is out of date!</h6>
 </div>
 
 <script>
-    $(document).ready(function() {
+    function addLoadEvent(func) {
+        const oldonload = window.onload;
+        if (typeof window.onload != 'function') {
```
ThiefMaster: `!==`

ThiefMaster: Can't we just use `addEventListener`? Or do some old browsers not support this? (we don't care about IE6)

indico/web/templates/_outdated_browser.html
```diff
@@ -6,12 +6,27 @@ <h6>Your browser is out of date!</h6>
 </div>
 
 <script>
-    $(document).ready(function() {
+    function addLoadEvent(func) {
+        const oldonload = window.onload;
```
ThiefMaster: I doubt old browsers support `const`.

indico/web/templates/_outdated_browser.html
```diff
@@ -6,12 +6,27 @@ <h6>Your browser is out of date!</h6>
 </div>
 
 <script>
-    $(document).ready(function() {
+    function addLoadEvent(func) {
```
ThiefMaster: Let's not create a global function :)

webpack.config.babel.js
```diff
@@ -35,7 +35,8 @@ let entryPoints = {
     markdown: ['./js/jquery/markdown.js'],
     mathjax: ['./js/jquery/compat/mathjax.js'],
     statistics: ['./js/jquery/statistics.js'],
-    fonts: ['./styles/partials/_fonts.scss']
+    fonts: ['./styles/partials/_fonts.scss'],
+    outdatedbrowser: ['./js/outdatedbrowser/index.js']
```
ThiefMaster: since we just have a single file in `js/outdatebrowser/`, what about just making it `./js/outdatedbrowser.js`?

indico/modules/rb_new/views/base.py
```diff
@@ -25,7 +25,7 @@ class WPRoomBookingBase(WPNewBase):
     template_prefix = 'rb/'
     title = _('Room Booking')
     bundles = ('common.js', 'common.css', 'react.js', 'react.css', 'semantic-ui.js', 'semantic-ui.css',
-               'module_rb_new.js', 'module_rb_new.css')
+               'module_rb_new.js', 'module_rb_new.css', 'outdatedbrowser.js', 'outdatedbrowser.css')
```
ThiefMaster: what about adding those two in `WPNewBase`? That way it's guaranteed to be included on all 'new' pages in case we add more in the future!

indico/web/templates/_outdated_browser.html
```diff
@@ -6,12 +6,12 @@ <h6>Your browser is out of date!</h6>
...
+    window.addEventListener('load', function() {
         outdatedBrowser({
             bgColor: '#0b63a5',
             color: '#ffffff',
-            lowerThan: 'IE11',
+            lowerThan: 'object-fit', // Edge
             languagePath: ''
-        });
-    })
+        })
```
ThiefMaster: missing semicolon

indico/modules/rb_new/views/base.py
ThiefMaster: moved it there since burotel inherits directly from `WPNewBase`

PR #3920, opened 2019-05-09
removed ur'' strings (resolves #3838)

### LINKED ISSUE #3838: ur'strings' are syntax errors in Python 3
In Python 3, you can have __u'strings'__ or __r'strings'__ but __ur'string' is a ___syntax error___.

$ __python3 -c "ur'string'"__  # --> syntax error

[flake8](http://flake8.pycqa.org) testing of https://github.com/indico/indico on Python 3.7.1

$ __flake8 . --count --select=E9,F63,F72,F822 --show-source --statistics__
```
./indico/legacy/pdfinterface/conference.py:148:56: E999 SyntaxError: invalid syntax
            if i > 0 and re.match(ur'<(p|ul|ol)\b[^>]*>', part):
                                                       ^
./indico/web/http_api/metadata/xml.py:47:90: E999 SyntaxError: invalid syntax
    def _convert(self, value, _control_char_re=re.compile(ur'[\x00-\x08\x0b\x0c\x0e-\x1f]')):
                                                                                         ^
./indico/modules/rb/models/rooms_test.py:470:16: E999 SyntaxError: invalid syntax
    (ur'foo\bar', ur'foo\bar', ''),
               ^
./indico/util/string.py:164:31: E999 SyntaxError: invalid syntax
    value = re.sub(ur'[^\w\s-]', u'', value).strip()
                              ^
./indico/util/spreadsheets.py:55:65: E999 SyntaxError: invalid syntax
def _prepare_csv_data(data, _linebreak_re=re.compile(ur'(\r?\n)+')):
                                                                ^
./indico/util/console.py:114:77: E999 SyntaxError: invalid syntax
    string = re.sub(ur'%\{(?P<fg>[a-z]+)(?P<fg_bold>!?)(?:,(?P<bg>[a-z]+))?}', _cformat_sub, string)
                                                                            ^
./indico/util/mdx_latex.py:627:52: E999 SyntaxError: invalid syntax
        new_blocks = [re.sub(ur'<a[^>]*>([^<]+)</a>', lambda m: convert_link_to_latex(m.group(0)).strip(), block)
                                                   ^
```
__E901,E999,F821,F822,F823__ are the "_showstopper_" [flake8](http://flake8.pycqa.org) issues that can halt the runtime with a SyntaxError, NameError, etc. These 5 are different from most other flake8 issues which are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

rooterkyberian:
The only difference between `re.compile(ur'')` and `re.compile(r'')` that I know of is that they are not equal in `re.compile(ur'') != re.compile(r'')`. Otherwise the behavior will be the same - on Python 2 they will happily parse both `str` and `unicode` objects alike.

### DIFF
--- a/indico/legacy/pdfinterface/conference.py
+++ b/indico/legacy/pdfinterface/conference.py
@@ -111,7 +111,7 @@ class ProgrammeToPDF(PDFBase):
         event_program = sanitize_for_platypus(render_markdown(track_settings.get(self.event, 'program')))
         parts = []
         for i, part in enumerate(re.split(r'\n+', event_program)):
-            if i > 0 and re.match(ur'<(p|ul|ol)\b[^>]*>', part):
+            if i > 0 and re.match(r'<(p|ul|ol)\b[^>]*>', part):
                 # extra spacing before a block-level element
                 parts.append(u'<br/>')
             parts.append(part)
--- a/indico/util/console.py
+++ b/indico/util/console.py
@@ -102,7 +102,7 @@ def cformat(string):
     """
     reset = colored(u'')
     string = string.replace(u'%{reset}', reset)
-    string = re.sub(ur'%\{(?P<fg>[a-z]+)(?P<fg_bold>!?)(?:,(?P<bg>[a-z]+))?}', _cformat_sub, string)
+    string = re.sub(r'%\{(?P<fg>[a-z]+)(?P<fg_bold>!?)(?:,(?P<bg>[a-z]+))?}', _cformat_sub, string)
     if not string.endswith(reset):
         string += reset
     return Color(string)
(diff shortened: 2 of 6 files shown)


ThiefMaster:
Thanks! I always thought bytestring pattern might result in implicit encoding/decoding when parsing unicode with them.

Could you use imperative form in the commit msg though? i.e. "Remove ..." (and ideally move the issue reference in one of the extra lines since it's not very useful in the summary line)

PR #3922, opened 2019-05-10
Convert exec statements to exec() function calls

cclauss:
__exec__ statements are syntax errors in Python 3 but __exec()__ function works as expected in both Python 2 and Python 3.  https://portingguide.readthedocs.io/en/latest/builtins.html#the-exec-function

After this PR lands, the only remaining Python 3 syntax errors that I can detect are [__reraise__](https://portingguide.readthedocs.io/en/latest/exceptions.html#the-new-raise-syntax) issues at:
* ./indico/core/storage/backend.py:236 and
* ./indico/core/db/sqlalchemy/core.py:57

### DIFF
--- a/indico/core/config.py
+++ b/indico/core/config.py
@@ -141,7 +141,7 @@ def _parse_config(path):
     locals_ = {}
     with codecs.open(path, encoding='utf-8') as config_file:
         # XXX: unicode_literals is inherited from this file
-        exec compile(config_file.read(), path, 'exec') in globals_, locals_
+        exec(compile(config_file.read(), path, 'exec'), globals_, locals_)
     return {unicode(k if k.isupper() else _convert_key(k)): v
             for k, v in locals_.iteritems()
             if k[0] != '_'}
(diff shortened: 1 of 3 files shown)


indico/core/config.py
```diff
@@ -141,7 +141,7 @@ def _parse_config(path):
     locals_ = {}
     with codecs.open(path, encoding='utf-8') as config_file:
         # XXX: unicode_literals is inherited from this file
-        exec compile(config_file.read(), path, 'exec') in globals_, locals_
+        exec(compile(config_file.read(), path, 'exec'), globals_, locals_)
```
ThiefMaster: Unfortunately this isn't an option right now, see 03f0334aab

ThiefMaster:
Actually, I just noticed that you had this already in one of your earlier PRs, and I reverted this change in 6ad534cd33 because of CentOS 7's Python 2.7 version failing when using this syntax in a function that also contains a dict comprehension.

cclauss:
Sorry about that.  CentOS does have very long lifespans.
https://en.wikipedia.org/wiki/CentOS#End-of-support_schedule

ThiefMaster:
FWIW, we'll expect people to use CentOS 8 (or get newer Python packages using any other way) long before 7 goes EOL ;)

PR #3928, opened 2019-05-15
Allow setting name format per event

mvidalgarcia:


indico/modules/users/models/users.py
```diff
@@ -44,16 +44,17 @@ class UserTitle(RichIntEnum):
...
                   _('John DOE'), _('DOE, John'), _('DOE, J.'), _('J. DOE'))
-    first_last = 0
-    last_first = 1
-    last_f = 2
-    f_last = 3
-    first_last_upper = 4
-    last_first_upper = 5
-    last_f_upper = 6
-    f_last_upper = 7
+    inherit = 0
```
ThiefMaster: unless you add an alembic revision to update all existing name form at user settings you can't just change those values since they are stored in the database...

Maybe just use `-1` for inherit? Or like i initially suggested, don't change the enum at all but instead use `None` on the event-level name format to indicate inheritance

indico/modules/users/models/users.py
```diff
@@ -131,7 +132,10 @@ def display_full_name(self):
 
 
 def format_display_full_name(user, obj):
-    name_format = user.settings.get('name_format') if user else NameFormat.first_last
+    from indico.modules.events.layout import layout_settings
+    name_format = layout_settings.get(g.rh.event, 'name_format') if 'rh' in g else NameFormat.inherit
```
ThiefMaster: `and hasattr(g.rh, 'event')`

indico/modules/users/models/users.py
```diff
@@ -131,7 +132,10 @@ def display_full_name(self):
...
+    if name_format is NameFormat.inherit:
```
ThiefMaster: don't use `is` here

indico/modules/users/models/users.py
pferreir: Are we not storing this anywhere in the DB?

indico/modules/events/layout/forms.py
```diff
@@ -85,6 +88,8 @@ def validate_use_custom_css(self, field):
 
 
 class LectureMeetingLayoutForm(IndicoForm):
+    name_format = IndicoEnumSelectField(_('Name format'), enum=NameFormat, default=NameFormat.inherit,
+                                        description=_('Default format in which names are displayed'))
```
ThiefMaster: This is not a "default" since it forces the selected format regardless of the user's preference.

indico/modules/events/layout/__init__.py
```diff
@@ -37,7 +40,9 @@
     'timetable_theme_settings': {},
     'use_custom_menu': False,
     'timetable_by_room': False,
-    'timetable_detailed': False,
```
ThiefMaster: why remove that comma? ;x

pferreir: Why not? 😉

ThiefMaster: unrelated to the other changes and no real benefit in removing it :p

mvidalgarcia: Whoops, it wasn't intentional

pferreir: This is why we need Black.

indico/modules/users/models/users.py
pferreir: This is useful if you're not sure whether you should use `is` or `==`:
https://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is

indico/modules/events/layout/__init__.py
ThiefMaster: I still don't like it very much because of their weird decision on quote style, but without quote normalization and 120-char line length I think it wouldn't really change a lot for us.

mvidalgarcia:
Updated.

PR #3934, opened 2019-05-17
Use prettier for JS formatting

ThiefMaster:
TODO: 

- [x] Actually reformat all the files
- [x] Ensure that CI checks are still fine (should be since prettier is integrated in eslint)

```sh
# Reformat all our JS
npx prettier --write indico/**/*.{js,jsx}
# or if we don't bother about old code:
npx prettier --write indico/modules/events/logs/**/*.{jsx,js}
npx prettier --write indico/modules/rb/client/js/**/*.{jsx,js}
npx prettier --write indico/web/client/js/react/**/*.{jsx,js}
npx prettier --write indico/web/client/js/utils/**/*.js
```

I will apply the prettier changes once I'm back from vacation (by then I think most ongoing work that involves bigger JS changes should be done and likely ready to merge); if it need to be done earlier please use `git commit --author 'Indico Team <indico-team@cern.ch>` as the commit will be MASSIVE!

---

Opinions welcome:

- [jsxBracketSameLine](https://prettier.io/docs/en/options.html#jsx-brackets) 
- [printWidth](https://prettier.io/docs/en/options.html#print-width) - 120 vs 100 vs 80 (I didn't find anything looking awful with 120 so I'm inclined to stick with that)


mvidalgarcia:
My opinion: 
- `jsxBracketSameLine: true`
- `printWidth` - No strong opinion, either 100 or 120 look reasonable.

pferreir:
>     * `jsxBracketSameLine: true`

I would probably prefer `false` because of the visual confusion that `true` causes with nested content (see e.g. [here](https://prettier.io/docs/en/options.html#jsx-brackets))

>     * `printWidth` - No strong opinion, either 100 or 120 look reasonable.

Same here.

ThiefMaster:
Good point; and it makes adding a new prop at the end of the list easier because you don't have to put the cursor before the `>`

PR #3935, opened 2019-05-22
Show correct number of registrations

mic4ael:


indico/modules/events/registration/models/forms.py
```diff
@@ -364,6 +364,12 @@ def sender_address(self):
         contact_email = self.event.contact_emails[0] if self.event.contact_emails else None
         return self.notification_sender_address or contact_email
 
+    @property
```
ThiefMaster: I don't like the idea of a property triggering a query on every access.

I'd either make this a `column_property` (can be deferred to not run it all the time) or a normal method (`get_registration_count()`)

ThiefMaster: I just saw that the other property is already done as a column property, so just do the same for the new one: https://github.com/indico/indico/blob/211375ede936dd5f2ad9dc7b2e4d15ad1b94a050/indico/modules/events/registration/models/forms.py#L418-L423

mic4ael:
Updated ;)

CHANGES.rst
```diff
@@ -47,6 +47,7 @@ Bugfixes
 - Fix paper reviewers getting notifications even if their type of reviewing
   has been disabled (:issue:`3852`)
 - Correctly handle merging users in the paper reviewing module (:issue:`3895`)
+- Show correct number of registrations (:issue:`#3935`)
```
ThiefMaster: without the `#`

ThiefMaster: "....in management area" ?

indico/modules/events/registration/models/forms.py
```diff
@@ -415,3 +415,7 @@ def _mappers_configured():
              .where((Registration.registration_form_id == RegistrationForm.id) & Registration.is_active)
              .correlate_except(Registration))
     RegistrationForm.active_registration_count = column_property(query, deferred=True)
+    query = (select([db.func.count(Registration.id)])
```
ThiefMaster: empty line before?

CHANGES.rst
mic4ael: It is not only management area. There is one place in the display area where we show the number of registrations.

indico/modules/events/registration/templates/display/regform_list.html
```diff
@@ -44,7 +44,7 @@
                     {% if any_published_registrations %}
                         {% if regform.publish_registration_count %}
                             <td class="icon-user">
-                                {{- regform.active_registration_count -}}
+                                {{- regform.existing_registrations_count -}}
```
ThiefMaster: I'm not sure if this one should be changed. Why would anyone but a manager care about a rejected registration? Such a person is not participating in the event and also not counting against the limit for example!

mic4ael: Yeah, I wasn't really sure about this one. Updated.

PR #3936, opened 2019-05-22
Add Help, Contact & Terms on Sidebar

### LINKED ISSUE #3627: Show footer on RB pages
Especially footer links (contact, privacy, etc),

panagiotappl:
Disable scrolling when Sidebar is open

closes #3627

### DIFF
--- a/indico/modules/core/templates/impersonation.html
+++ b/indico/modules/core/templates/impersonation.html
@@ -22,14 +22,6 @@
                 title="{% trans user=message %}Switch back to {{ original_name }}{% endtrans %}"></i>
         </div>
     </div>
-    <script>
-        (() => {
-            const numBars = Array.from(document.querySelectorAll('.announcement-bar'))
-                .filter((e) => e.style.display === 'flex')
-                .length;
-            document.body.style.setProperty('--extra-bars', numBars + 1);
-        })();
-    </script>
 {% endmacro %}
 
 
--- a/indico/modules/core/templates/impersonation.html
+++ b/indico/modules/core/templates/impersonation.html
@@ -1,4 +1,4 @@
-{% macro  render_impersonation_legacy(original_name) %}
+{% macro render_impersonation_legacy(original_name) %}
     <div class="impersonation-header clearfix">
         <span class="icon-eye text">
             <small>{% trans %}Impersonating{% endtrans %}</small>
@@ -22,6 +22,14 @@
                 title="{% trans user=message %}Switch back to {{ original_name }}{% endtrans %}"></i>
         </div>
     </div>
+    <script>
+        (() => {
+            const numBars = Array.from(document.querySelectorAll('.announcement-bar'))
+                .filter((e) => e.style.display === 'flex')
+                .length;
+            document.body.style.setProperty('--extra-bars', numBars + 1);
+        })();
+    </script>
 {% endmacro %}
 
 
--- a/indico/modules/rb/client/js/common/config/reducers.js
+++ b/indico/modules/rb/client/js/common/config/reducers.js
@@ -20,6 +20,8 @@ export const initialState = {
     helpURL: '',
     hasTOS: false,
     tosHTML: null,
+    hasPrivacyPolicy: false,
+    privacyPolicyHTML: null,
     contactEmail: null,
 };
 
@@ -29,9 +31,10 @@ export default combineReducers({
         switch (action.type) {
             case configActions.CONFIG_RECEIVED: {
                 const {
-                    roomsSpriteToken, gracePeriod, helpURL, contactEmail,
+                    roomsSpriteToken, gracePeriod, helpURL, contactEmail, hasPrivacyPolicy,
                     hasTos: hasTOS,
                     tosHtml: tosHTML,
+                    privacyPolicyHtml: privacyPolicyHTML,
                     tileserverURL: tileServerURL,
                 } = camelizeKeys(action.data);
                 const {languages} = action.data;
@@ -43,6 +46,8 @@ export default combineReducers({
                     helpURL,
                     hasTOS,
                     tosHTML,
+                    hasPrivacyPolicy,
+                    privacyPolicyHTML,
                     contactEmail,
                 };
             }
--- a/indico/modules/rb/client/js/common/config/reducers.js
+++ b/indico/modules/rb/client/js/common/config/reducers.js
@@ -17,6 +17,10 @@ export const initialState = {
     languages: {},
     tileServerURL: '',
     gracePeriod: 1,
+    helpURL: '',
+    tosURL: null,
+    tosHTML: null,
+    contactEmail: null,
 };
 
 export default combineReducers({
@@ -24,13 +28,21 @@ export default combineReducers({
     data: (state = initialState, action) => {
         switch (action.type) {
             case configActions.CONFIG_RECEIVED: {
-                const {roomsSpriteToken, tileserverURL: tileServerURL, gracePeriod} = camelizeKeys(action.data);
+                const {
+                    roomsSpriteToken, gracePeriod, helpURL, contactEmail, tosURL,
+                    tosHtml: tosHTML,
+                    tileserverURL: tileServerURL,
+                } = camelizeKeys(action.data);
                 const {languages} = action.data;
                 return {
                     roomsSpriteToken,
                     languages,
                     tileServerURL,
-                    gracePeriod
+                    gracePeriod,
+                    helpURL,
+                    tosURL,
+                    tosHTML,
+                    contactEmail,
                 };
             }
             case configActions.SET_ROOMS_SPRITE_TOKEN:
--- a/indico/modules/rb/client/js/common/config/reducers.js
+++ b/indico/modules/rb/client/js/common/config/reducers.js
@@ -18,7 +18,7 @@ export const initialState = {
     tileServerURL: '',
     gracePeriod: 1,
     helpURL: '',
-    tosURL: null,
+    hasTOS: false,
     tosHTML: null,
     contactEmail: null,
 };
@@ -29,7 +29,8 @@ export default combineReducers({
         switch (action.type) {
             case configActions.CONFIG_RECEIVED: {
                 const {
-                    roomsSpriteToken, gracePeriod, helpURL, contactEmail, tosURL,
+                    roomsSpriteToken, gracePeriod, helpURL, contactEmail,
+                    hasTos: hasTOS,
                     tosHtml: tosHTML,
                     tileserverURL: tileServerURL,
                 } = camelizeKeys(action.data);
@@ -40,7 +41,7 @@ export default combineReducers({
                     tileServerURL,
                     gracePeriod,
                     helpURL,
-                    tosURL,
+                    hasTOS,
                     tosHTML,
                     contactEmail,
                 };
--- a/indico/modules/rb/client/js/components/App.jsx
+++ b/indico/modules/rb/client/js/components/App.jsx
@@ -86,6 +86,7 @@ class App extends React.Component {
     }
 
     onSidebarHide = () => {
+        document.body.classList.remove('scrolling-disabled');
         this.setState({userActionsVisible: false});
     };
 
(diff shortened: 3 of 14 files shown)


indico/modules/rb/blueprint.py
```diff
@@ -32,6 +32,9 @@
 _bp.add_url_rule('/api/stats', 'stats', misc.RHStats)
 _bp.add_url_rule('/api/map-areas', 'map_areas', misc.RHMapAreas)
 _bp.add_url_rule('/api/equipment', 'equipment_types', misc.RHEquipmentTypes)
+_bp.add_url_rule('/api/tos', 'tos', misc.RHTos)
```
ThiefMaster: I'd go for `RHTOS` or `RHTermsOfService`; I think mixed-cases acronyms look weird 🙈

indico/modules/rb/blueprint.py
```diff
@@ -32,6 +32,9 @@
...
+_bp.add_url_rule('/api/help', 'help_url', misc.RHHelp)
```
ThiefMaster: Do we really need API endpoints for this? I think this can all go into `RHConfig` - I'd go as far as including the ToS URL/text in there, so none of the new endpoints here would be needed.

indico/modules/rb/client/js/components/App.jsx
```diff
@@ -86,6 +86,7 @@ class App extends React.Component {
     }
 
     onSidebarHide = () => {
+        document.body.style.overflowY = 'auto';
```
ThiefMaster: 🙈 but I guess it's needed as a workaround.

Could we use a class for it and add/remove it (using `document.body.classList`)? Like this it's clear that this and the line where the style attribute is set to hidden are related (since searching for the class name will show all relevant places in the code)

indico/modules/rb/client/js/components/SidebarMenu.jsx
```diff
@@ -42,97 +47,177 @@ SidebarTrigger.propTypes = {
...
-        {
-            key: 'adminOverride',
-            icon: isAdminOverrideEnabled ? 'unlock' : 'lock',
-            text: Translate.string('Admin Override'),
-            iconColor: isAdminOverrideEnabled ? 'orange' : undefined,
-            active: isAdminOverrideEnabled,
-            onClick: toggleAdminOverride,
-            onlyIf: isAdmin,
-            tooltip: Translate.string('Admin Override gives you unrestricted access to all rooms and bookings.')
+class SidebarMenu extends React.Component {
```
ThiefMaster: I don't think we need to convert this to a class component :)

When moving the URLs to the config store (see my previous comment) those are provided by redux, and the simple state used here right now would be a perfect use case for hooks otherwise!

indico/modules/rb/client/js/components/SidebarMenu.jsx
```diff
@@ -42,97 +47,177 @@ SidebarTrigger.propTypes = {
...
+                            <div>
+                                <a href={`mailto:${contactEmail}`}><Translate>{contactEmail}</Translate></a>
+                            </div>
+                        </Modal.Content>
+                    </Modal>
+                )}
+                {tos && (
+                    <Modal open={termsVisible}
+                           closeIcon
+                           onClose={() => this.setState({termsVisible: false})}>
```
ThiefMaster: Do we really need state for this? Wouldn't it be possible to make the menu item the trigger for an uncontrolled Modal?

indico/modules/rb/templates/room_booking.html
```diff
@@ -2,5 +2,4 @@
 
 {% block body %}
     <div id="rb-app-container"></div>
-    {% include '_outdated_browser.html' %}
```
ThiefMaster: was this really intended?

indico/modules/rb/client/js/components/SidebarMenu.jsx
```diff
@@ -42,97 +47,177 @@ SidebarTrigger.propTypes = {
...
-                }
-                return <Popup trigger={item} content={tooltip} key={key} position="left center" />;
-            })}
-        </Sidebar>
-    );
+                        return <Popup trigger={item} content={tooltip} key={key} position="left center" />;
+                    })}
+                    <div styleName="bottom-align">
+                        <Menu.Item as="a" onClick={this.redirectToHelpURL}>
+                            Help
```
mvidalgarcia: i18n

indico/modules/rb/client/js/components/SidebarMenu.jsx
```diff
@@ -42,97 +47,177 @@ SidebarTrigger.propTypes = {
...
+                        </Menu.Item>
+                        <Menu.Item onClick={this.showContact}>
+                            Contact
```
mvidalgarcia: same

indico/modules/rb/client/js/components/SidebarMenu.jsx
```diff
@@ -5,13 +5,17 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
-import React from 'react';
+import fetchConfig from 'indico-url:rb.config';
```
ThiefMaster: `configURL`, but the config is in the redux state so you do not need to fetch it in here

indico/modules/rb/client/js/components/SidebarMenu.jsx
```diff
@@ -103,35 +107,106 @@ function SidebarMenu({
         }
     ].filter(({onlyIf}) => onlyIf === undefined || onlyIf);
 
+    const [contactVisible, setContactVisible] = useState(false);
+    const [termsVisible, setTermsVisible] = useState(false);
+    const [config, setConfig] = useState(null);
+
+    useEffect(() => {
```
ThiefMaster: this would be a good case for `useIndicoAxios`, but see my comment above about this data already being available

indico/modules/rb/controllers/backend/misc.py
```diff
@@ -27,15 +29,27 @@
...
 from indico.util.i18n import get_all_locales
+from indico.util.string import sanitize_html
 from indico.web.flask.util import send_file, url_for
 
 
 class RHConfig(RHRoomBookingBase):
     def _process(self):
+        url = legal_settings.get('tos_url')
+        if url:
+            tos_type = 'url'
```
ThiefMaster: matter of personal opinion, but i'd probably just send `tos_url` and `tos_html`.

anyway, with a `tos_type` the case of not having tos at all should be covered (`'none'` or `null`) as well

indico/modules/rb/controllers/backend/misc.py
```diff
@@ -27,15 +29,27 @@
...
+            tos = url
+        else:
+            tos_type = 'html'
+            tos = sanitize_html(legal_settings.get('tos'))
         return jsonify(rooms_sprite_token=unicode(_cache.get('rooms-sprite-token', '')),
                        languages=get_all_locales(),
                        tileserver_url=rb_settings.get('tileserver_url'),
-                       grace_period=rb_settings.get('grace_period'))
+                       grace_period=rb_settings.get('grace_period'),
+                       help_URL=config.HELP_URL,
```
ThiefMaster: `help_url=...`

indico/modules/rb/client/js/components/SidebarMenu.jsx
```diff
@@ -103,35 +107,106 @@ function SidebarMenu({
...
+                    <Modal.Header>
+                        <Translate>Contact</Translate>
+                    </Modal.Header>
+                    <Modal.Content>
+                        <div>
+                            <Translate>If you need support, you can contact the following email address:</Translate>
+                        </div>
+                        <div>
+                            <a href={`mailto:${config.contactEmail}`}>
+                                <Translate>{config.contactEmail}</Translate>
```
mic4ael: Use `<Param />` here ;)

ThiefMaster: more like "don't use i18n here" ;)

indico/modules/rb/client/js/common/config/reducers.js
```diff
@@ -17,20 +17,29 @@ export const initialState = {
...
 };
 
 export default combineReducers({
     request: requestReducer(configActions.FETCH_REQUEST, configActions.FETCH_SUCCESS, configActions.FETCH_ERROR),
     data: (state = initialState, action) => {
         switch (action.type) {
             case configActions.CONFIG_RECEIVED: {
-                const {roomsSpriteToken, tileserverURL: tileServerURL, gracePeriod} = camelizeKeys(action.data);
+                const {roomsSpriteToken, tileserverURL: tileServerURL, gracePeriod, helpURL, contactEmail, tosURL,
+                       tosHtml} = camelizeKeys(action.data);
```
ThiefMaster: `tosHtml: tosHTML` so you don't have to use the awful mixed-cased acronym spelling ;)

indico/modules/rb/client/js/components/SidebarMenu.jsx
```diff
@@ -103,35 +108,96 @@ function SidebarMenu({
...
+                    );
+                    if (!tooltip) {
+                        return item;
+                    }
+                    return <Popup trigger={item} content={tooltip} key={key} position="left center" />;
+                })}
+                <div styleName="bottom-align">
+                    {helpURL && (
+                        <Menu.Item as="a" onClick={() => {
+                            window.location.href = helpURL;
```
ThiefMaster: can't we set a normal href? Using JS to change location is pretty ugly since you can't middle-click for example...

indico/modules/rb/client/js/components/SidebarMenu.jsx
```diff
@@ -103,35 +108,96 @@ function SidebarMenu({
...
+                    )}
+                    {contactEmail && (
+                        <Menu.Item onClick={() => setContactVisible(true)}>
+                            <Translate>Contact</Translate>
+                        </Menu.Item>
+                    )}
+                    {(tosURL || tosHtml) && (
+                        <Menu.Item onClick={() => {
+                            if (tosURL) {
+                                window.location.href = tosURL;
```
ThiefMaster: same as above

ThiefMaster: FWIW, if a text is used you can still set the href to the existing tos page (in the indico core) and preventDefault the click event to avoid going there when opening the modal

indico/modules/rb/controllers/backend/misc.py
```diff
@@ -27,15 +29,26 @@
 from indico.modules.rb.util import build_rooms_spritesheet
 from indico.util.caching import memoize_redis
 from indico.util.i18n import get_all_locales
+from indico.util.string import sanitize_html
 from indico.web.flask.util import send_file, url_for
 
 
 class RHConfig(RHRoomBookingBase):
     def _process(self):
+        tos_url, tos_html = None, None
```
ThiefMaster: tos_url = tos_html = None

indico/modules/rb/controllers/backend/misc.py
```diff
@@ -27,15 +29,25 @@
...
+from indico.util.string import sanitize_html
 from indico.web.flask.util import send_file, url_for
 
 
 class RHConfig(RHRoomBookingBase):
     def _process(self):
+        tos_url = None
+        url = legal_settings.get('tos_url')
+        if url:
+            tos_url = url
```
ThiefMaster: this whole thing can be replaced with `tos_url = legal_settings.get('tos_url')` - and if the setting can be an empty string while you need `None`, just add `or None`

indico/modules/core/templates/impersonation.html
```diff
@@ -22,6 +22,14 @@
                 title="{% trans user=message %}Switch back to {{ original_name }}{% endtrans %}"></i>
         </div>
     </div>
+    <script>
+        (() => {
+            let numBars = Array.from(document.querySelectorAll('.announcement-bar'))
```
mvidalgarcia: Can't you use `const`?

indico/modules/core/templates/impersonation.html
```diff
@@ -22,6 +22,14 @@
...
+                .filter(function(e) {return e.style.display === 'flex';})
```
mvidalgarcia: Isn't it possible to use arrow functions here?

ThiefMaster: It is. We don't support browsers anymore that do not support arrow functions.

PR #3938, opened 2019-05-24
Allow request workflows to set their own Reply-To

pferreir:
This PR should set more user-friendly defaults when it comes to e-mail notifications sent by request (e.g. Webcast/Recording, in CERN plugins) workflows.

Messages sent to the event manager:
 * `From:` *no-reply address*
 * `Reply-To`:  Defined by Request Definition, *support e-mail* otherwise

This is because the user will normally want to reply back to a service-specific address, which each plugin can define.

Messages sent to the request managers (e.g. Webcast manager):
 * `From:` *no-reply address*
 * `Reply-To`: *support e-mail*

This is the same as before, only that we send the message from a *no-reply* address, which is more a more standard way of doing things.

### DIFF
--- a/indico/modules/events/requests/base.py
+++ b/indico/modules/events/requests/base.py
@@ -39,7 +39,7 @@ class RequestManagerForm(IndicoForm):
 
 
 class RequestDefinitionBase(object):
-    """Defines a service request which can be sent by event managers."""
+    """A service request which can be sent by event managers."""
 
     #: the plugin containing this request definition - assigned automatically
     plugin = None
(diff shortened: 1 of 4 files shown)


indico/modules/events/requests/base.py
```diff
@@ -117,6 +118,11 @@ def get_manager_notification_emails(cls):
         """
         return set()
 
+    @classmethod
+    def get_notification_reply_email(cls):
+        """Return the `Reply-To:` e-mail address for notifications."""
```
ThiefMaster: double backticks

PR #3939, opened 2019-05-24
Manage map areas

### LINKED ISSUE #3931: Add MapArea management
Add/delete/edit map areas. Should ideally use an actual map to select an area and then save it.

mic4ael:
closes #3931


indico/modules/rb/blueprint.py
```diff
@@ -30,7 +30,7 @@
 _bp.add_url_rule('/api/<path:path>', '404', lambda path: (jsonify(), 404))
 _bp.add_url_rule('/api/config', 'config', misc.RHConfig)
 _bp.add_url_rule('/api/stats', 'stats', misc.RHStats)
-_bp.add_url_rule('/api/map-areas', 'map_areas', misc.RHMapAreas)
+_bp.add_url_rule('/api/map-areas', 'map_areas', misc.RHMapAreas, methods=('GET', 'POST', 'PATCH', 'DELETE'))
```
ThiefMaster: ~~DELETE/PATCH should have the ID in the URL (same endpoint name is fine though)~~

all the management stuff should be in `/api/admin/...` below with the other admin endpoints.

indico/modules/rb/client/js/common/map/reducers.js
```diff
@@ -25,7 +25,13 @@ export default combineReducers({
     areas: (state = [], action) => {
         switch (action.type) {
             case mapActions.AREAS_RECEIVED:
+            case mapActions.MAP_AREA_UPDATED:
+            case mapActions.MAP_AREA_CREATED:
```
ThiefMaster: I'd keep admin stuff separate and just reload the list of map areas when leaving the admin area

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,372 @@
...
+                            <FinalCheckbox name="default"
+                                           label={Translate.string('Default map area')}
+                                           toggle />
+                        </Form.Group>
+                        <Form.Group>
+                            <FinalInput name="bounds.southWest.lat"
+                                        type="number"
+                                        label={Translate.string('SW Latitude')}
+                                        disabled={submitting || submitSucceeded}
+                                        validate={v.number()}
```
ThiefMaster: This shouldn't be needed; `input type=number` never returns non-numeric values so this validator will never fail.

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,372 @@
...
+                        <Form.Group>
+                            <FinalCheckbox name="default"
+                                           label={Translate.string('Default map area')}
+                                           toggle />
+                        </Form.Group>
+                        <Form.Group>
+                            <FinalInput name="bounds.southWest.lat"
+                                        type="number"
+                                        label={Translate.string('SW Latitude')}
+                                        disabled={submitting || submitSucceeded}
```
ThiefMaster: submitting is handled internally; no need to do it here

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,372 @@
...
+                                        label={Translate.string('NE Longitude')}
+                                        type="number"
+                                        disabled={submitting || submitSucceeded}
+                                        validate={v.number()}
+                                        required />
+                        </Form.Group>
+                    </Form>
+                </Modal.Content>
+                <Modal.Actions>
+                    <Button type="submit" form="map-area-form"
```
ThiefMaster: I added a FinalSubmit component when refactoring the form stuff; check if you can use that here.

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,372 @@
...
+
+    renderModalForm = () => {
+        const {areaProps} = this.state;
+        if (!areaProps) {
+            return null;
+        }
+
+        const {id, bounds, name, isDefault} = areaProps;
+        return (
+            <FinalForm onSubmit={this.onSubmit}
```
ThiefMaster: specify the subscription

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,372 @@
...
+                        },
+                        id: area.id,
+                        isDefault: area.is_default,
+                        name: area.name
+                    }
+                });
+            };
+
+            return (
+                <Rectangle key={`${area.name}-${center.x}-${center.y}`}
```
ThiefMaster: why not the ID?

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,372 @@
...
+        });
+    };
+
+    onSubmit = (data) => {
+        const {actions: {createMapArea, updateMapAreas}} = this.props;
+        const {areaProps: {id}} = this.state;
+        let result;
+
+        if (!id) {
+            result = createMapArea(data);
```
ThiefMaster: why no async/await?

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
ThiefMaster: Also, do we really want to disable on success? If we keep the modal open after saving (on edit), there's no need to prevent further edits. If we close it, it doesn't matter anyway.

indico/modules/rb/client/js/modules/admin/MapAreasPage.module.scss
```diff
@@ -0,0 +1,26 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+@import 'rb:styles/palette';
```
ThiefMaster: are you using anything from the palette here?

indico/modules/rb/controllers/backend/misc.py
```diff
@@ -79,9 +83,33 @@ def _get_stats(date):
...
+        'bounds': fields.Dict(required=True),
+        'name': fields.String(required=True),
+        'default': fields.Bool(missing=False)
+    })
+    def _process_POST(self, args):
+        create_area(args)
         return jsonify(map_areas_schema.dump(MapArea.query))
 
+    @use_kwargs({
+        'areas': fields.List(fields.Dict(), required=True)
```
ThiefMaster: `fields.Dict` doesn't validate anything of the actual data in there...

indico/modules/rb/controllers/backend/misc.py
```diff
@@ -79,9 +83,33 @@ def _get_stats(date):
...
+    })
+    def _process_PATCH(self, areas):
+        for area in areas:
+            update_area(area.pop('id'), area)
+        return jsonify(map_areas_schema.dump(MapArea.query))
```
ThiefMaster: `return map_areas_schema.jsonify(...)`

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,372 @@
...
+        this.map = null;
+        this.featureGroup = null;
+    }
+
+    state = {
+        areaProps: null,
+        suppressClick: false
+    };
+
+    componentDidMount() {
```
ThiefMaster: let's also fetch the latest map areas from the server in here

indico/modules/rb/blueprint.py
ThiefMaster: OK, after reading the rest of the PR I see the reason for not having an ID there (updating multiple areas with a single request).

Not sure if the RESTful verbs are really appropriate for something like this. I can't say for sure which approach I'd use here but some that come to my mind:

- single POST endpoint where we simply send the new list - and then determine on the server what do delete/add/update based on whether there's an ID in the area's dict (update) or whether an areas is missing in the payload but exists on the server (delete)
- one request per area (maybe even immediately after changing something)

indico/modules/rb/client/js/common/map/reducers.js
mic4ael: But I need this to rerender the map with new areas

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
mic4ael: I close it so yeah, probably it is not that necessary here.

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
mic4ael: Ohh, nice. Didn't notice that one.

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
mic4ael: Oh yeah, I was actually to do that! Thanks for reminding me about it ;)

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
mic4ael: Actually, not sure why I didn't use id here

indico/modules/rb/blueprint.py
mic4ael: Hm, I think in terms of simplicity my approach is not that bad. Those operations are nicely separated which makes it easier to understand them.

indico/modules/rb/client/js/common/map/reducers.js
ThiefMaster: hadn't noticed that you are using the same state for admin and the normal rb interface until after writing that comment

mic4ael: I mean, is there any advantage of putting the same data in the admin-related part of redux store?

ThiefMaster: not if the data is the same which seems to be the case here

mic4ael: Yep, exactly.

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
mic4ael: Yeah, but in case someone tampers with the `type` attribute, it will still be possible to send it to backend, causing a DB error. So, additional layer of validation doesn't hurt here.

ThiefMaster: You know what they say about client-side validation... it's just for user convenience. If someone tampers with the attribute, they can just send random garbage as well.

The schema on the server should be more specific than just `Dict` though, to avoid this :)

mic4ael: Yeah, of course it doesn't stop the user from tampering with the data even more. But isn't it better at least to have clear error messages? :P (I mean I don't see a big problem with having this extra check there)

BTW: good point regarding the schema

ThiefMaster: I don't see the point in adding a check that will never be triggered except when the user messes around with the DOM.

Also, if we want such a validator it'd be more appropriate to add inside FinalInput so it applies to all number fields. But like I said, I don't think there's a real benefit there (I wanted to add it there initially to show a nicer error if someone enters non-numeric data, but since that doesn't work...)

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
mic4ael: Ah, now I remember. In case when I use `id` as the value for the `key` prop, rectangles (areas) are not rerendered when e.g. editing the size/coordinates (which for example affects position of labels).

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
mic4ael: Hm, now when I am using `<Final...` fields I started wondering whether I even have to specify subscription on the form.

mic4ael: I guess an empty object would be fine here?

ThiefMaster: If you don't use any fprops (except the submit callback) then it can indeed be empty

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
mic4ael: Hm, but what is the benefit of refetching it here?

mic4ael:
Updated

indico/modules/rb/controllers/backend/admin.py
```diff
@@ -503,3 +504,45 @@ def _process_POST(self, args):
         db.session.flush()
         RHRoomsPermissions._jsonify_user_permissions.clear_cached(session.user)
         return jsonify(id=room.id)
+
+
+_base_schema = {
+    'default': fields.Bool(missing=False),
```
ThiefMaster: are you sure this is correct when used in the `update_schema`? unless you always send a value for `default` (in which case it should be `required`), omitting it will be just like sending `False`.

indico/modules/rb/controllers/backend/admin.py
```diff
@@ -503,3 +504,45 @@ def _process_POST(self, args):
         db.session.flush()
         RHRoomsPermissions._jsonify_user_permissions.clear_cached(session.user)
         return jsonify(id=room.id)
+
+
+_base_schema = {
```
ThiefMaster: what about using actual `mm.Schema` subclasses here and then proper inheritance below instead of `dict(.., **..)`?

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,356 @@
...
+            return null;
+        }
+
+        const {id, bounds, name, isDefault} = areaProps;
+        return (
+            <FinalForm onSubmit={this.onSubmit}
+                       render={this.renderForm}
+                       subscription={{}}
+                       initialValues={{
+                           name: !id ? null : name,
```
ThiefMaster: `id === null` (or `undefined`) - while there shouldn't be MapAreas with id 0, I'd rather do an explicit check for null/undefined here.

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,356 @@
...
+                           bounds,
+                           default: isDefault
+                       }} />
+        );
+    };
+
+    renderForm = (fprops) => {
+        const {handleSubmit} = fprops;
+        const {areaProps: {id}} = this.state;
+        const isNewArea = !id;
```
ThiefMaster: same as above

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,356 @@
...
+                       initialValues={{
+                           name: !id ? null : name,
+                           bounds,
+                           default: isDefault
+                       }} />
+        );
+    };
+
+    renderForm = (fprops) => {
+        const {handleSubmit} = fprops;
```
ThiefMaster: I'd destructure this one in the line before - it's small enough now to not look awful

indico/modules/rb/controllers/backend/admin.py
mic4ael: Is it really worth it to create a class here? :P

indico/modules/rb/controllers/backend/admin.py
mic4ael: Actually I think I might remove it!

mic4ael:
Updated

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,355 @@
...
+    };
+
+    drawAreas = () => {
+        if (this.map === null) {
+            return null;
+        }
+
+        const {areas} = this.props;
+        return areas.map((area) => {
+            const southWest = [area.top_left_latitude, area.top_left_longitude];
```
pferreir: GIven the amount of times you write `area`, I would consider de-structuring in the function args already.
Actually, this could be a good candidate for a `_drawArea` method. It's a bit of a massive method right now.

indico/modules/rb/operations/admin.py
```diff
@@ -62,3 +63,45 @@ def update_room(room, args):
...
+
+def create_area(area_data):
+    bounds = area_data['bounds']
+    top, bottom = bounds['north_east'], bounds['south_west']
+    is_default = area_data.get('default', False)
+
+    if is_default:
+        MapArea.query.update({MapArea.is_default: False}, synchronize_session='fetch')
+
+    new_area = MapArea()
```
pferreir: It's a choice of style, but wouldn't it be more readable to pass those as `kwargs` of the constructor?

indico/modules/rb/blueprint.py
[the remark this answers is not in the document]
mic4ael: Good point! I forgot to remove those after moving some code to admin area

mic4ael:
Updated

ThiefMaster:
"Clear all" without confirmation is super dangerous. If we can't easily add a prompt there let's just disable that one (not sure why you don't have to click "save" for it...)

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,358 @@
...
+                                        required />
+                        </Form.Group>
+                        <Form.Group>
+                            <FinalCheckbox name="default"
+                                           label={Translate.string('Default map area')}
+                                           toggle />
+                        </Form.Group>
+                        <Form.Group>
+                            <FinalInput name="bounds.southWest.lat"
+                                        type="number"
```
ThiefMaster: `step="0.00000000000001"` - otherwise there are html5 validation errors as soon as you change the value manually (default step is 1)

indico/modules/rb/controllers/backend/admin.py
```diff
@@ -532,9 +532,9 @@ def _process_POST(self, args):
 
 
 class RHMapAreas(RHRoomBookingAdminBase):
-    @use_args(_create_args)
-    def _process_POST(self, args):
-        create_area(args)
+    @use_kwargs(_create_args)
```
ThiefMaster: You can just do `create_area(**args)` below without changing anything here - no need to convert between dicts and function args before calling that function ;)

mic4ael: Oh yeah, I think I forgot to change it after I experimented with `use_kwargs`. Thanks for fixing it!

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,366 @@
...
+import 'leaflet/dist/leaflet.css';
+import 'leaflet-draw/dist/leaflet.draw.css';
+import './MapAreasPage.module.scss';
+
+
+Leaflet.EditToolbar.Delete.include({
+    removeAllLayers: false
+});
+
+const DEFAULT_CENTER = [46.233832398, 6.053166454];
```
mvidalgarcia: Is it OK to keep this hardcoded?

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
```diff
@@ -0,0 +1,366 @@
...
+                const {options: {id}} = layer;
+                if (id === undefined) {
+                    this.featureGroup.leafletElement.removeLayer(layer);
+                }
+            });
+        }
+    };
+
+    editArea = (evt) => {
+        if (!Object.keys(evt.layers._layers).length) {
```
mvidalgarcia: Any particular reason why this is before the `const` declarations?

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
ThiefMaster: Probably not amazing. Could we query the tile server for the center of the area it has available somehow? @pferreir

mic4ael: Yep, not the best thing to do. But couldn't find a way to get that information from the tile server

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
mic4ael: Why would you put it here? I don't need those definitions at this point.

mic4ael:
> Looks very good!  just one thing that I noticed: When editing an area and either moving it or resizing it, the popup with the name doesn't follow the bottom right corner. I've noticed that leaflet-draw have `onEditMove` event. Perhaps can be called to move the popup as well?

Hm, I didn't think it was worth fixing. Since I would have to redraw the whole map every time an area was moved.

indico/modules/rb/client/js/modules/admin/MapAreasPage.jsx
mvidalgarcia: I know, just code style. IIRC some linters have this rule to have definitions on top.

PR #3944, opened 2019-05-31
[WIP] Simple webhook-based notification system

pferreir:
This implements a very simple webhook-based notification system that should be compatible with CERN's Push Notification service and any other web service using a REST API. Authentication is done using Bearer tokens and data is sent in JSON.

A few things to discuss:
 * Content types - I am using the usual `RenderModeMixin` and allowing Markdown, HTML and Plain Text. For now we're just sending out whatever type the notification claims to be in. It could be interesting to discuss Markdown-based notifications in Indico, since it can be converted to HTML easily or gracefully degraded to plaintext if needed;
 * The dispatching mechanism - right now notifications are appended in a DB table and then dispatched by a task that runs every ~1 min. This is not real time, of course, but is more than enough for this use case (notifications about starting/ending bookings);
    - I chose to store the notification in the DB instead of just sending it out to Celery for failure recovery and tracking reasons;
    - An interesting alternative would be to immediately send out the notification but also save it in the DB (it could even be saved asynchronously, post-request). Then we could have a task that just goes over notifications that have been pending for longer than X;
 * It would be interesting to have a "notification area" for users, where they can check which notifications they've received. Since we have that information in the DB, why not?
    - In that case it could be interesting to provide the PN service with a callback URL that it could optionally call to let us know that the person has read the notification;

Things to change:
 * Using `Bearer` token instead of proprietary field (we have to tell the PN guys to use that instead).

### DIFF
--- a/indico/core/notifications.py
+++ b/indico/core/notifications.py
@@ -79,12 +79,13 @@ def _log_email(email, event, module, user, meta=None):
                      user, type_='email', data=log_data, meta=meta)
 
 
-def init_email_queue():
+def init_notification_queues():
     """Enable email queueing for the current context."""
     g.setdefault('email_queue', [])
+    g.setdefault('webhook_queue', set())
 
 
-def flush_email_queue():
+def _flush_email_queue():
     """Send all the emails in the queue.
 
     Note: This function does a database commit to update states
@@ -94,7 +95,7 @@ def flush_email_queue():
     been pending.
     """
     from indico.core.emails import store_failed_email, update_email_log_state
-    queue = g.get('email_queue', [])
+    queue = g.get('email_queue')
     if not queue:
         return
     logger.debug('Sending %d queued emails', len(queue))
@@ -116,6 +117,31 @@ def flush_email_queue():
             time.sleep(0.25)
     del queue[:]
     db.session.commit()
+
+
+def _flush_webhook_queue():
+    """Send enqueued webhook notifications."""
+    from indico.modules.notifications.tasks import send_webhook_notification
+    queue = g.get('webhook_queue')
+    if not queue:
+        return
+    logger.debug('Sending %d queued webhook notifications', len(queue))
+
+    for notification in queue:
+        try:
+            send_webhook_notification.delay(notification)
+        except Exception:
+            logger.exception('Flushing queued webhook notification %d failed', notification.id)
+            # Wait for a short moment in case it's a very temporary issue
+            time.sleep(0.25)
+    queue.clear()
+    db.session.commit()
+
+
+def flush_notification_queues():
+    """Send enqueued notifications (e-mail and webhook)."""
+    _flush_email_queue()
+    _flush_webhook_queue()
 
 
 def make_email(to_list=None, cc_list=None, bcc_list=None, from_address=None, reply_address=None, attachments=None,
--- a/indico/migrations/versions/20190531_1220_87d248818199_add_table_for_notifications.py
+++ b/indico/migrations/versions/20190531_1220_87d248818199_add_table_for_notifications.py
@@ -1,0 +1,39 @@
+"""Add table for notifications
+
+Revision ID: 87d248818199
+Revises: 06a4ec717b84
+Create Date: 2019-05-31 12:20:19.089312
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+from indico.core.db.sqlalchemy import PyIntEnum, UTCDateTime
+from indico.core.db.sqlalchemy.descriptions import RenderMode
+
+
+# revision identifiers, used by Alembic.
+revision = '87d248818199'
+down_revision = '06a4ec717b84'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'notifications',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('user_id', sa.Integer(), nullable=False, index=True),
+        sa.Column('subject', sa.Text(), nullable=False),
+        sa.Column('render_mode', PyIntEnum(RenderMode), nullable=False),
+        sa.Column('body', sa.Text(), nullable=False),
+        sa.Column('sent_dt', UTCDateTime, index=True),
+        sa.Column('created_dt', UTCDateTime, index=True, nullable=False),
+        sa.ForeignKeyConstraint(['user_id'], ['users.users.id']),
+        sa.PrimaryKeyConstraint('id'),
+        schema='indico'
+    )
+
+
+def downgrade():
+    op.drop_table('notifications', schema='indico')
--- a/indico/modules/notifications/forms.py
+++ b/indico/modules/notifications/forms.py
@@ -1,0 +1,34 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2019 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from __future__ import unicode_literals
+
+from wtforms.fields import StringField
+from wtforms.validators import Optional, ValidationError
+
+from indico.util.i18n import _
+from indico.web.forms.base import IndicoForm
+from indico.web.forms.fields import IndicoPasswordField
+from indico.web.forms.validators import IndicoRegexp
+
+
+class NotificationSettingsForm(IndicoForm):
+    webhook_url = StringField(
+        _('Webhook URL'),
+        [IndicoRegexp(r'^https://.+'), Optional()],
+        description=_("Webhook to contact for notifications.")
+    )
+    secret_token = IndicoPasswordField(
+        _('Secret Token'),
+        toggle=True,
+        description=_("The bearer token used for authentication with the Webhook.")
+    )
+
+    def validate_secret_token(self, field):
+        # We shouldn't allow a secret token without a URL
+        if field.data and not self.webhook_url.data:
+            raise ValidationError(_('You should specify a Webhook URL'))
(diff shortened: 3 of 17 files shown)


indico/modules/notifications/tasks.py
```diff
@@ -0,0 +1,48 @@
...
+    token = notification_settings.get('secret_token')
+
+    if not url:
+        raise ValueError('Webhook URL not set!')
+
+    sent_items = set()
+
+    try:
+        for notification in Notification.query.filter(~Notification.is_sent):
+            if notification.user.email == 'pedro.ferreira@cern.ch':
```
pferreir: This is intentional. It's here to avoid accidental spamming of the PN service.

indico/migrations/versions/20190531_1220_87d248818199_add_table_for_notifications.py
```diff
@@ -0,0 +1,39 @@
...
+        sa.Column('user_id', sa.Integer(), nullable=False),
+        sa.Column('subject', sa.Text(), nullable=False),
+        sa.Column('is_sent', sa.Boolean(), nullable=False),
+        sa.Column('render_mode', PyIntEnum(RenderMode), nullable=False),
+        sa.Column('body', sa.Text(), nullable=False),
+        sa.ForeignKeyConstraint(['user_id'], ['users.users.id']),
+        sa.PrimaryKeyConstraint('id'),
+        schema='indico'
+    )
+    op.create_index(None, 'notifications', ['user_id'], unique=False, schema='indico')
```
ThiefMaster: `index=True` on the `sa.Column()` above

indico/modules/notifications/forms.py
```diff
@@ -0,0 +1,32 @@
...
+from indico.web.forms.validators import IndicoRegexp
+
+
+class NotificationSettingsForm(IndicoForm):
+    webhook_url = StringField(
+        _('Webhook URL'),
+        [IndicoRegexp(r'^https://.+'), Optional()],
+        description=_("Webhook to contact for notifications.")
+    )
+    secret_token = StringField(
```
ThiefMaster: `IndicoPasswordField`? (that way it shows up as a password field but lets the user reveal its content)

indico/modules/notifications/forms.py
```diff
@@ -0,0 +1,32 @@
...
+        _('Secret Token'),
+        description=_("The bearer token used for authentication with the Webhook.")
+    )
+
+    def validate_secret_token(self, field):
+        # We shouldn't allow a secret token without a URL
+        if field.data and not self.webhook_url.data:
+            raise ValidationError('You should specify a Webhook URL')
```
ThiefMaster: i18n, and I'd skip this validation - it's annoying if you can't just clear the non-secret data to disable something. Or maybe add a checkbox and HiddenUnless on the two other fields (which can then both be required)

indico/modules/notifications/models/notifications.py
```diff
@@ -0,0 +1,70 @@
...
+from sqlalchemy.ext.declarative import declared_attr
+
+from indico.core.db.sqlalchemy import db
+from indico.core.db.sqlalchemy.descriptions import RenderMode, RenderModeMixin
+
+
+class Notification(db.Model, RenderModeMixin):
+    """Notifications sent by Indico to its users"""
+    __tablename__ = 'notifications'
+    __table_args__ = (db.Index(None, 'user_id'),
```
ThiefMaster: `index=True` on the column

ThiefMaster: ...which you already have, so this one is not needed at all

indico/modules/notifications/models/notifications.py
```diff
@@ -0,0 +1,70 @@
...
+    #: ID of the user
+    user_id = db.Column(
+        db.Integer,
+        db.ForeignKey('users.users.id'),
+        index=True,
+        nullable=False
+    )
+    #: subject of the message
+    subject = db.Column(
+        db.Text,
```
ThiefMaster: `db.String` - usually we use Text for longer/multiline content

indico/modules/notifications/templates/admin.html
```diff
@@ -0,0 +1,20 @@
...
+
+{% block content %}
+    {{ simple_form(form) }}
+    {% if form.webhook_url.data %}
+        <h2>Test</h2>
+        <button class="i-button icon-bullhorn"
+                data-href="{{ url_for('notifications.send_test') }}"
+                data-method="POST"
+                data-ajax
+                data-update="#dummy">
```
ThiefMaster: do we actually need that? :/

indico/modules/notifications/util.py
```diff
@@ -0,0 +1,70 @@
...
+
+from indico.core.db import db
+from indico.core.db.sqlalchemy.descriptions import RenderMode
+from indico.core.notifications import email_sender
+from indico.modules.notifications import settings as notification_settings
+from indico.modules.notifications.models.notifications import Notification
+from indico.util.caching import memoize
+
+
+@memoize
```
ThiefMaster: Not a good idea, that way changing the setting won't ever be taking into account unless the server is restarted. `memoize_request` would work, but settings are already cached during the request so no need to memoize anything here.

indico/modules/notifications/views.py
```diff
@@ -0,0 +1,14 @@
...
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from __future__ import unicode_literals
+
+from indico.modules.admin.views import WPAdmin
+
+
+class WPNotificationSettings(WPAdmin):
+    template_prefix = 'notifications/'
```
ThiefMaster: set the active sidemenu item either here or when calling `render_template()`

ThiefMaster:
Regarding dispatching/storing: I think the minutely task is fine for this particular case, but for something where you can get a notification as a result of your own action I think this is somewhat annoying - imagine performing some action. and then getting an email/notification about it only a minute later. Much more disrupting than getting (and then probably deleting) it immediatey.

In terms of reliability there's no need to store them - even if Celery is down, sending a task won't fail. It will still be enqueued and picked up as soon as celery is restarted. And when redis is down, then more things are broken anyway (especially sessions and celery, so very low chance of something triggering a notification).

That aside, I think persistent storage of notifications is not a bad idea (maybe delete old ones after a while though, no need to keep that stuff for a long time..), since it'd allow showing them inside indico (like on fb, github, etc.) at some point as well.

So I'd put the notification in the DB, but also immediately send it to celery. We just need to make sure that it's committed before sending it to Celery or there can be cases where celery tries to load the notification from the DB before it's been committed.

indico/modules/notifications/templates/admin.html
pferreir: Yes, I'm afraid so.

indico/modules/notifications/util.py
pferreir: Indeed. Good point.

pferreir:
> In terms of reliability there's no need to store them - even if Celery is down, sending a task won't fail. It will still be enqueued and picked up as soon as celery is restarted.

That was my initial approach, but then I thought what a mess it will be if a few thousand tasks fail and we end up with Celery retrying all of them in an interval of minutes/seconds. Or if there's a bug that causes a task to fail and we just want to get rid of it and have to manually delete it from the Celery queue...

ThiefMaster:
> Or if there's a bug that causes a task to fail and we just want to get rid of it and have to manually delete it from the Celery queue...

If a task fails unexpectedly, it won't be retried. It'll only retry it if you catch the exception and raise a different one telling celery to retry (or maybe you can also configure a task to auto-retry, but I don't think we do that anywhere)

pferreir:
Yeah, I know. But do you really want to ignore a task that fails, right away? Imagine there's a bug and a few thousands of them just fail. That's why keeping a log isn't a bad idea IMHO.

indico/migrations/versions/20190531_1220_87d248818199_add_table_for_notifications.py
pferreir: I seem to forget about this every 6 months.

indico/modules/notifications/util.py
```diff
@@ -0,0 +1,68 @@
...
+
+def enqueue_notification(notification):
+    for receiver in notification['to']:
+        notification = Notification(
+            user=receiver,
+            subject=notification['subject'],
+            body=notification['body'],
+            render_mode=RenderMode[notification.get('content_type', 'plain_text')]
+        )
+        db.session.add(notification)
```
mic4ael: Don't we want to flush here as well?

pferreir: Why? Does it make any difference, since we commit at the end anyway?

ThiefMaster: no, flush is only relevant if you need e.g. the id of a newly inserted object or want to ensure everything has been sent to the DB (e.g. because of other queries depending on it)

pferreir:
I've updated the PR to use a different strategy:
 * Sending the notification right away and graciously failing if something is wrong (logging it, of course);
 * Running, every 5 minutes, a task that re-sends all notifications older than 5 minutes that haven't yet been marked as sent;

indico/core/notifications.py
```diff
@@ -119,6 +120,31 @@ def flush_email_queue():
     db.session.commit()
 
 
+def _flush_webhook_queue():
+    """Send enqueued webhook notifications."""
+    from indico.modules.notifications.tasks import send_webhook_notification
+    queue = g.get('webhook_queue', set())
```
ThiefMaster: no need for the default value since you'd return right below anyway if it's none

ThiefMaster: can be changed for the email queue as well...

PR #3952, opened 2019-06-04
make http api example doc python 3 compatible

pseyfert:
The HTTP API access example [here](https://docs.getindico.io/en/stable/http_api/access/#request-signing-for-python) is for python 2 only. Updated it according to [these pointers](https://python-future.org/compatible_idioms.html) to be python 2 and python 3 compatible.

 - `print` with parentheses
 - import of `urlencode` (alternatives kept in comments in case `future` isn't installed as convenience for the reader)
 - convert `dict_items` to `list`
 - encode `str` before hashing
 - while on it, ran `yapf` to deal with longer lines.

### DIFF
--- a/docs/source/http_api/access.rst
+++ b/docs/source/http_api/access.rst
@@ -99,12 +99,16 @@ A simple example in Python::
 
     import hashlib
     import hmac
-    import urllib
     import time
+
+    try:
+        from urllib.parse import urlencode
+    except ImportError:
+        from urllib import urlencode
 
 
     def build_indico_request(path, params, api_key=None, secret_key=None, only_public=False, persistent=False):
-        items = params.items() if hasattr(params, 'items') else list(params)
+        items = list(params.items()) if hasattr(params, 'items') else list(params)
         if api_key:
             items.append(('apikey', api_key))
         if only_public:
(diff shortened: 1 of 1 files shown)


docs/source/http_api/access.rst
```diff
@@ -99,12 +99,21 @@ A simple example in Python::
...
-        items = params.items() if hasattr(params, 'items') else list(params)
+    from future.moves.urllib.parse import urlencode
+    ## python 3
+    # from urllib.parse import urlencode
+    ## python 2
+    # from urllib import urlencode
+
+
+    def build_indico_request(path,
+                             params,
```
ThiefMaster: I don't think wrapping it increases readability - and if someone pastes it in their code changes are good they'll reformat it according to their own codebase's style anyway.

docs/source/http_api/access.rst
```diff
@@ -99,12 +99,21 @@ A simple example in Python::
...
     import hmac
-    import urllib
     import time
-
-
-    def build_indico_request(path, params, api_key=None, secret_key=None, only_public=False, persistent=False):
-        items = params.items() if hasattr(params, 'items') else list(params)
+    from future.moves.urllib.parse import urlencode
+    ## python 3
+    # from urllib.parse import urlencode
```
ThiefMaster: Why not try importing this and then catch `ImportError` to use the python2 fallback?

ThiefMaster:
Thanks a lot for the PR! I left you some comments on how it could be improved so people can use it regardless of which Python version they're on.

docs/source/http_api/access.rst
pseyfert: whichever you prefer, maitainer edits are enabled.

docs/source/http_api/access.rst
pseyfert: again, no preference from my side what you want to recommend your users.

PR #3963, opened 2019-06-05
Room Booking: make DateRangePicker more responsive

pferreir:
This should solve the issues with people using screens with lower resolutions. We fall back to a full-screen date picker in those cases.


ThiefMaster:
Looks good!

I guess we want the fullscreen thingy for the single date picker, too?

pferreir:
> Looks good!
> 
> I guess we want the fullscreen thingy for the single date picker, too?

Yeah, indeed. Good idea.

PR #3968, opened 2019-06-11
Show territory next to the language name

mic4ael:


indico/web/client/js/react/components/UserMenu.jsx
```diff
@@ -110,8 +114,11 @@ export default function UserMenu({userData, languages}) {
           <Translate>My Preferences</Translate>
         </Dropdown.Item>
         <Dropdown.Header>
-          <Icon name="globe" />
-          {languageSelector}
+          {languageSelector(
+            <>
+              <Icon name="globe" /> {`${languages[language][0]} (${languages[language][1]})`}
```
ThiefMaster: `{languages[language][0]} ({languages[language][1]})`

and maybe assign `const langInfo = languages[language];` and then use that

PR #3976, opened 2019-06-19
Bookings export

mvidalgarcia:


indico/modules/rb/client/js/common/bookings/BookingExportModal.jsx
```diff
@@ -0,0 +1,130 @@
...
+import {Button, Form, Modal} from 'semantic-ui-react';
+import PropTypes from 'prop-types';
+
+import {Translate} from 'indico/react/i18n';
+import {FinalDatePeriod} from 'indico/react/components';
+import {FinalRadio} from 'indico/react/forms';
+import {actions as bookingActions} from '../bookings';
+import FinalRoomSelector from '../../components/RoomSelector';
+
+function validate({format, dates, rooms}) {
```
ThiefMaster: Am I the only one who doesn't like the separate validation function (in favor of having validation functions specified on the Field itself)?

indico/modules/rb/client/js/common/bookings/BookingExportModal.jsx
```diff
@@ -0,0 +1,130 @@
...
+};
+
+BookingExportModal.defaultProps = {
+  rooms: [],
+};
+
+export default connect(
+  null,
+  dispatch => ({
+    actions: bindActionCreators(
```
ThiefMaster: For simple components that don't have tons of actions (let's say no more than 2 or 3?) I probably wouldn't nest them inside `actions`. Because that way you can get this much simpler version:

```js
export default connect(
  null,
  {exportBookings: bookingActions.exportBookings}
)(BookingExportModal);
```

ThiefMaster: However, I don't see a reason to use Redux here at all. There is no benefit in not sending the request directly from within the component.

If you move, the request into the component you can use the `handleSubmitError` util to convert errors to the format final-form expects:

```js
    try {
      response = await indicoAxios.XXX(....);
    } catch (error) {
      return handleSubmitError(error);
    }
```

indico/modules/rb/client/js/common/bookings/BookingExportModal.jsx
```diff
@@ -0,0 +1,130 @@
...
+            <Form.Group>
+              <FinalRadio name="format" value="csv" label={Translate.string('CSV')} />
+              <FinalRadio name="format" value="xlsx" label={Translate.string('XLSX')} />
+            </Form.Group>
+            <FinalDatePeriod name="dates" label={Translate.string('Period')} required allowNull />
+            <FinalRoomSelector name="rooms" label={Translate.string('Rooms to export')} required />
+          </Form>
+        </Modal.Content>
+        <Modal.Actions>
+          <Button
```
ThiefMaster: Use `FinalSubmitButton` here; it handles all the disabled/loading logic for you.

indico/modules/rb/client/js/common/bookings/BookingExportModal.jsx
```diff
@@ -0,0 +1,130 @@
...
+  if (!dates || !Object.values(dates).every(x => x)) {
+    errors.dates = Translate.string('Please choose a valid period.');
+  }
+  if (!rooms || !rooms.length) {
+    errors.rooms = Translate.string('Please choose at least one room to export its bookings.');
+  }
+  return errors;
+}
+
+const BookingExportModal = props => {
```
ThiefMaster: I also didn't do it a few times but I think we should try to use `function` statements for top-level components, especially those that are exported. That way it's guaranteed that the component name is properly set in all cases.

indico/modules/rb/client/js/common/bookings/BookingExportModal.jsx
```diff
@@ -0,0 +1,130 @@
...
+
+  const renderModalContent = fprops => {
+    const {submitting, hasValidationErrors, pristine, submitSucceeded} = fprops;
+    return (
+      <>
+        <Modal.Header>
+          <Translate>Export bookings</Translate>
+        </Modal.Header>
+        <Modal.Content>
+          <Form id="export-bookings-form" onSubmit={fprops.handleSubmit} success={submitSucceeded}>
```
ThiefMaster: You don't need the `success` here unless you have a `<Message success ...` inside it.

indico/modules/rb/client/js/common/bookings/BookingExportModal.jsx
```diff
@@ -0,0 +1,130 @@
...
+
+  return (
+    <Modal open onClose={onClose} size="tiny" closeIcon>
+      <FinalForm
+        validate={validate}
+        onSubmit={handleSubmit}
+        render={renderModalContent}
+        initialValues={{rooms, dates: null}}
+        initialValuesEqual={_.isEqual}
+        subscription={{
```
ThiefMaster: if you apply the other suggestions this can be empty

indico/modules/rb/controllers/backend/bookings.py
```diff
@@ -353,3 +362,38 @@ def _process(self):
...
+                               ReservationOccurrence.is_valid,
+                               db_dates_overlap(ReservationOccurrence,
+                                                'start_dt', datetime.combine(start_date, time()),
+                                                'end_dt', datetime.combine(end_date, time.max)))).all()
+
+        token = unicode(uuid.uuid4())
+        headers, rows = generate_spreadsheet_from_occurrences(occurrences)
+        _cache.set(token, {'headers': headers, 'rows': rows}, time=1800)
+        download_url = url_for('rb.export_bookings_{}'.format(format), token=token)
+        return jsonify_data(url=download_url)
```
ThiefMaster: `jsonify_data` is not needed since you don't flash or need `success: True`. Use flask's `jsonify()` instaed

indico/modules/rb/controllers/backend/bookings.py
```diff
@@ -353,3 +362,38 @@ def _process(self):
...
+                       .filter(Reservation.room_id.in_(room_ids),
+                               ReservationOccurrence.is_valid,
+                               db_dates_overlap(ReservationOccurrence,
+                                                'start_dt', datetime.combine(start_date, time()),
+                                                'end_dt', datetime.combine(end_date, time.max)))).all()
+
+        token = unicode(uuid.uuid4())
+        headers, rows = generate_spreadsheet_from_occurrences(occurrences)
+        _cache.set(token, {'headers': headers, 'rows': rows}, time=1800)
+        download_url = url_for('rb.export_bookings_{}'.format(format), token=token)
```
ThiefMaster: Please don't use dynamic endpoint names; that's awful when searching for the place where an endpoint is used.
Maybe use `bookings.<any(csv,xlsx):format>` in the route and handle both in a single RH? Then you can pass it as a kwarg to `url_for` here.

indico/modules/rb/client/js/common/bookings/BookingExportModal.jsx
```diff
@@ -0,0 +1,113 @@
...
+  );
+}
+
+BookingExportModal.propTypes = {
+  rooms: PropTypes.array,
+  onClose: PropTypes.func.isRequired,
+};
+
+BookingExportModal.defaultProps = {
+  rooms: [],
```
ThiefMaster: this is never undefined/null so no need to have it optional

mvidalgarcia: It is if you access the url directly `rooms/bookings/export`

ThiefMaster: isn't it an empty array in that case?

indico/modules/rb/client/js/common/bookings/BookingExportModal.jsx
```diff
@@ -70,8 +69,26 @@ export default function BookingExportModal({rooms, onClose}) {
...
+              required
+              allowNull
+            />
+            <FinalRoomSelector
+              name="rooms"
+              label={Translate.string('Rooms to export')}
+              validate={val => {
+                if (!val || !val.length) {
+                  return Translate.string(
+                    'Please choose at least one room to export its bookings.'
```
ThiefMaster: "to export its bookings" sounds a bit weird; i think "Please choose at least one room" is clear enough

indico/modules/rb/client/js/common/bookings/BookingExportModal.jsx
```diff
@@ -0,0 +1,120 @@
...
+  handleSubmitError,
+  FinalSubmitButton,
+} from 'indico/react/forms';
+
+import FinalRoomSelector from '../../components/RoomSelector';
+
+function validate({format}) {
+  const errors = {};
+  if (!format) {
+    errors.type = Translate.string('Please choose a file format to export the bookings.');
```
ThiefMaster: `errors.format`?

PR #3989, opened 2019-07-04
Room booking responsiveness

mvidalgarcia:
TODO:
- [x] Fix building number input getting too small on type
- [x] Modals
- [ ] ~Timelines~ (different PR)

### DIFF
--- a/indico/modules/rb/client/js/components/SidebarFooter.jsx
+++ b/indico/modules/rb/client/js/components/SidebarFooter.jsx
@@ -9,12 +9,68 @@ import contactURL from 'indico-url:core.contact';
 import tosURL from 'indico-url:legal.display_tos';
 import privacyPolicyURL from 'indico-url:legal.display_privacy';
 
+import _ from 'lodash';
 import React, {useState} from 'react';
 import PropTypes from 'prop-types';
-import {Menu, Modal} from 'semantic-ui-react';
+import {Icon, Menu, Modal, Popup} from 'semantic-ui-react';
 import {connect} from 'react-redux';
 import {Translate} from 'indico/react/i18n';
+import {Responsive} from 'indico/react/util';
 import {selectors as configSelectors} from '../common/config';
+
+const buildMenuItems = (helpURL, hasTOS, hasPrivacyPolicy) => {
+  const menuItems = [
+    [
+      'help',
+      {
+        href: helpURL,
+        target: '_blank',
+        rel: 'noopener noreferrer',
+        caption: Translate.string('Help'),
+        icon: 'question',
+      },
+    ],
+    [
+      'contact',
+      {
+        href: contactURL(),
+        modal: true,
+        caption: Translate.string('Contact'),
+        icon: 'at',
+      },
+    ],
+    [
+      'tos',
+      {
+        href: tosURL(),
+        modal: true,
+        target: '_blank',
+        rel: 'noopener noreferrer',
+        caption: Translate.string('Terms and Conditions'),
+        icon: 'file text',
+      },
+    ],
+    [
+      'privacy',
+      {
+        href: privacyPolicyURL(),
+        modal: true,
+        target: '_blank',
+        rel: 'noopener noreferrer',
+        caption: Translate.string('Privacy Policy'),
+        icon: 'eye',
+      },
+    ],
+  ];
+
+  if (!hasTOS) {
+    _.remove(menuItems, ([id]) => id === 'tos');
+  }
+  if (!hasPrivacyPolicy) {
+    _.remove(menuItems, ([id]) => id === 'privacy');
+  }
+  return menuItems;
+};
 
 function SidebarFooter({
   hasTOS,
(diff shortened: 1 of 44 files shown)


indico/modules/rb/client/js/util.jsx
```diff
@@ -161,19 +161,26 @@ export function renderRecurrence({type, number, interval}) {
   }
   if (type === 'single') {
     return (
-      <Responsive.Tablet andLarger orElse="S">
+      <Responsive.Tablet andLarger orElse={<Translate context="single">S</Translate>}>
```
ThiefMaster: i think the context should be something more verbose like "single booking shortcut".

fyi, `Translate.string()` also has a context arg, so you don't need to use the jsx version here

indico/modules/rb/client/js/modules/landing/Landing.module.scss
```diff
@@ -80,12 +81,17 @@
         transition: max-height 0.25s ease-out;
         margin: 10vh auto auto auto;
 
-        @media (min-width: 600px) {
+        @media (max-width: $tablet-width) {
+            width: 100%;
+            max-height: 100%;
+            margin-top: 1rem;
+        }
```
mic4ael: Maybe a line break after that closing brace?

indico/modules/rb/client/styles/responsive.scss
```diff
@@ -0,0 +1,11 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+
+$tablet-width: 768px;
```
mic4ael: Out of curiosity, where did you get those values from?

indico/web/client/js/react/util/Responsive.jsx
```diff
@@ -0,0 +1,59 @@
...
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import PropTypes from 'prop-types';
+import React from 'react';
+import Responsive from 'react-responsive';
+
+const DIMENSIONS = {
+  tablet: 768,
```
mic4ael: How about using https://github.com/css-modules/icss#export (at least checking how easy it is to import those variables from `responsive.scss`)

mvidalgarcia: Good idea, I'll give it a try :+1:

indico/modules/rb/client/styles/responsive.scss
mvidalgarcia: I'm not sure, from [semantic-ui](https://semantic-ui.com/elements/container.html) I think, they're kind of _standard_ if you do a quick search (responsive breakpoints), for instance Bootstrap uses the same. Probably @pferreir knows better as he added those.

indico/modules/rb/client/js/modules/bookRoom/filters/RecurrenceForm.jsx
```diff
@@ -35,7 +35,7 @@ export default class RecurrenceForm extends FilterFormComponent {
       this.stateChanger('type')(e.target.value);
     };
     this.onTypeChange = this.stateChanger('type');
-    this.onNumberChange = this.stateChanger('number', num => Math.abs(parseInt(num, 10) || 1));
+    this.onNumberChange = this.stateChanger('number', num => Math.abs(parseInt(num, 10)));
```
ThiefMaster: what's the reason for this change?

mvidalgarcia: From mobile you can't delete nor modify the `1` that is there by default. It's true that this makes the the form more error prone as you can try to submit it without a value but I couldn't think of any other solution that is mobile-friendly, do you have any other suggestion?

indico/web/client/js/react/util/Responsive.jsx
pferreir: Not sure if it's worth using that just for this use case.

indico/web/client/js/react/util/Responsive.jsx
```diff
@@ -0,0 +1,62 @@
...
+    andLarger: false,
+    andSmaller: false,
+    onlyIf: true,
+    orElse: null,
+  };
+
+  return _SizeSpec;
+};
+
+export default Object.assign(Responsive, {
```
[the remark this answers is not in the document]
mvidalgarcia: I've tried using the spread operator to clone the object like this:
```es6
export default {
  ...Responsive,
  WideScreen: factory(DIMENSIONS.wideScreen, null),
  Desktop: factory(DIMENSIONS.computer, DIMENSIONS.wideScreen),
  Tablet: factory(DIMENSIONS.tablet, DIMENSIONS.computer),
  Phone: factory(null, DIMENSIONS.tablet),
};
```
but then I get `Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: object.`, any idea?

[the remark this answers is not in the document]
ThiefMaster: `Responsive` is probably a react component so you can't just clone it like that (that's just copying object attrs). I guess in this case it's better to stick with the original option of modifying it - it's not pretty but less messy than creating a new class and inheriting from that (especially since that would break when they switch from class to function at some point)

indico/modules/rb/client/js/components/SidebarFooter.jsx
```diff
@@ -11,11 +11,64 @@ import privacyPolicyURL from 'indico-url:legal.display_privacy';
...
+      {
+        href: helpURL,
+        target: '_blank',
+        rel: 'noopener noreferrer',
+        caption: Translate.string('Help'),
+        icon: 'question',
+      },
+    ],
+    [
+      'contact',
```
ThiefMaster: This entry is conditional as well (only if `contactEmail` is set)

ThiefMaster:
Some things to not forget for the next PR:

- hide tooltips that aren't really needed (e.g. on the room tile buttons)
- hide edit room button
- timelines (obviously ;))

mvidalgarcia:
> Some things to not forget for the next PR:
> 
> * hide tooltips that aren't really needed (e.g. on the room tile buttons)
> * hide edit room button
> * timelines (obviously ;))

Thanks! I'll open a 2nd PR soon ;)

PR #3990, opened 2019-07-05
Always include start_dt changes in the signal

mic4ael:


indico/modules/events/operations.py
```diff
@@ -108,12 +109,15 @@ def update_event(event, update_timetable=False, **data):
...
     old_person_links = event.person_links[:]
+    changes = {}
     if (update_timetable or event.type == EventType.lecture) and 'start_dt' in data:
         # Lectures have no exposed timetable so if we have any timetable entries
         # (e.g. because the event had a different type before) we always update them
         # silently.
-        event.move_start_dt(data.pop('start_dt'))
-    changes = event.populate_from_dict(data)
+        start_dt = data.pop('start_dt')
+        changes['start_dt'] = (copy(event.start_dt), start_dt)
```
ThiefMaster: I'm pretty sure that `datetime` objects are immutable, so no need to copy

PR #4000, opened 2019-07-12
Fix convener affiliation

pferreir:
This was triggered by a CERN support ticket. We should be displaying the `SessionBlockPersonLink`'s attributes, not the `EventPerson`'s.


ThiefMaster:
huh i thought i had fixed it not too long ago... maybe it was somewhere else. we have everything from 2.1-maintenance merged, right?

pferreir:
> huh i thought i had fixed it not too long ago... maybe it was somewhere else. we have everything from 2.1-maintenance merged, right?

Yes, it's all merged.

PR #4001, opened 2019-07-12
Fix jquery slide animation

panagiotappl:
Fix slide right animation in timetable
Fix slide down animation in social button


indico/modules/events/timetable/templates/move_entry.html
```diff
@@ -85,7 +85,7 @@
         function toggleMoveButton(item) {
             item.toggleClass('selected');
             if (item.data('item-id') != $initiallySelected.data('item-id')) {
-                item.find('.move-btn').toggle('slide', {direction: 'right'}, 200);
+                item.find('.move-btn').animate({ width: "toggle"});
```
ThiefMaster: that space after `{` shouldn't be there

PR #4004, opened 2019-07-15
Handle page metadata a little bit better

pferreir:
Right now we're relying on `_get_head_content(...)`, which makes it a little harder to manage.


pferreir:
@ThiefMaster thanks for noticing that. Whoops!
That should be OK now. While fixing it, I realized that we could probably get rid of some duplicated code: https://github.com/indico/indico/pull/4004/commits/f49a5e32079c7e9f647395159c25e98488ef83bc
Does anyone see any issues with that? It seems to work fine.

PR #4009, opened 2019-07-18
Hide warning when booking already exists

### LINKED ISSUE #3987: Hide booking warning in case there is a booking for one of the managers during that period.

panagiotappl:
closes #3987

### DIFF
--- a/indico/modules/events/management/controllers/settings.py
+++ b/indico/modules/events/management/controllers/settings.py
@@ -11,6 +11,9 @@ from flask import redirect, session
 from werkzeug.exceptions import Forbidden
 
 from indico.core import signals
+from indico.core.config import config
+from indico.core.db import db
+from indico.core.db.sqlalchemy.util.queries import db_dates_overlap
 from indico.modules.events.management.controllers.base import RHManageEventBase
 from indico.modules.events.management.forms import (EventClassificationForm, EventContactInfoForm, EventDataForm,
                                                     EventDatesForm, EventLocationForm, EventPersonsForm)
@@ -18,6 +21,9 @@ from indico.modules.events.management.util import flash_if_unregistered
 from indico.modules.events.management.views import WPEventSettings, render_event_management_header_right
 from indico.modules.events.operations import update_event
 from indico.modules.events.util import track_time_changes
+from indico.modules.rb.models.reservation_occurrences import ReservationOccurrence
+from indico.modules.rb.models.reservations import Reservation
+from indico.modules.rb.models.rooms import Room
 from indico.util.signals import values_from_signal
 from indico.web.flask.templating import get_template_module
 from indico.web.forms.base import FormDefaults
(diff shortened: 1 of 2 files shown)


indico/modules/events/management/controllers/settings.py
```diff
@@ -41,7 +46,21 @@ def _check_access(self):
         RHManageEventBase._check_access(self)  # mainly to trigger the legacy "event locked" check
 
     def _process(self):
-        return WPEventSettings.render_template('settings.html', self.event, 'settings')
+        # Check if any of the managers of the event already have a booking that overlaps with the event datetime
+        manager_ids = [p.user.id for p in self.event.acl_entries if p.user]
```
ThiefMaster: that's everyone in the ACL, not just managers. the correct way to get only the managers is:

```python
manager_ids = {p.user.id
               for p in self.event.acl_entries
               if p.type == PrincipalType.user and p.has_management_permission()}
```

indico/modules/events/management/controllers/settings.py
```diff
@@ -41,7 +46,21 @@ def _check_access(self):
...
+        has_overlap = (ReservationOccurrence.query
+                       .filter(ReservationOccurrence.is_valid,
+                               db.or_(Reservation.booked_for_id.in_(manager_ids),
+                                      Reservation.created_by_id.in_(manager_ids)),
+                               db_dates_overlap(ReservationOccurrence,
+                                                'start_dt', self.event.start_dt_local,
+                                                'end_dt', self.event.end_dt_local),
+                               Reservation.room_id == self.event.room.id,
```
ThiefMaster: I think this breaks the settings page if the event has no room assigned...

indico/modules/events/management/templates/settings.html
```diff
@@ -5,7 +5,8 @@
 
 {% block content %}
     {{ template_hook('event-actions', event=event) }}
-    {%- if indico_config.ENABLE_ROOMBOOKING and not event.has_ended and event.room and not event.room_reservation_links -%}
+    {%- if indico_config.ENABLE_ROOMBOOKING and not event.has_ended and event.room and not event.room_reservation_links
+    and not has_overlap -%}
```
ThiefMaster: that's a weird indentation level for wrapping this. how about moving this logic into the RH class and just check e.g. `if show_no_booking_warning` here?

indico/modules/events/management/controllers/settings.py
```diff
@@ -11,13 +11,18 @@
...
+from indico.core.db import db
+from indico.core.db.sqlalchemy.util.queries import db_dates_overlap
 from indico.modules.events.management.controllers.base import RHManageEventBase
 from indico.modules.events.management.forms import (EventClassificationForm, EventContactInfoForm, EventDataForm,
                                                     EventDatesForm, EventLocationForm, EventPersonsForm)
 from indico.modules.events.management.util import flash_if_unregistered
 from indico.modules.events.management.views import WPEventSettings, render_event_management_header_right
 from indico.modules.events.operations import update_event
 from indico.modules.events.util import track_time_changes
+from indico.modules.rb import Room
```
mic4ael: It works just by coincidence. Use `from indico.modules.rb.models.rooms import Room`

PR #4013, opened 2019-07-23
RB responsiveness 2.0

mvidalgarcia:
##### TODO
- [x] Investigate map issue

### DIFF
--- a/indico/web/client/js/react/components/ResponsivePopup.jsx
+++ b/indico/web/client/js/react/components/ResponsivePopup.jsx
@@ -9,9 +9,7 @@ import React from 'react';
 import {Popup} from 'semantic-ui-react';
 import {useResponsive} from 'indico/react/util';
 
-const ResponsivePopup = props => {
+export default function ResponsivePopup(props) {
   const {isPhone, isTablet} = useResponsive();
   return <Popup {...props} disabled={isPhone || isTablet} />;
-};
-
-export default ResponsivePopup;
+}
--- a/indico/web/client/js/react/components/ResponsivePopup.jsx
+++ b/indico/web/client/js/react/components/ResponsivePopup.jsx
@@ -1,0 +1,17 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React from 'react';
+import {Popup} from 'semantic-ui-react';
+import {useResponsive} from 'indico/react/util';
+
+const ResponsivePopup = props => {
+  const {isPhone} = useResponsive();
+  return <Popup {...props} disabled={isPhone} />;
+};
+
+export default ResponsivePopup;
--- a/indico/web/client/js/react/components/ResponsivePopup.jsx
+++ b/indico/web/client/js/react/components/ResponsivePopup.jsx
@@ -10,8 +10,8 @@ import {Popup} from 'semantic-ui-react';
 import {useResponsive} from 'indico/react/util';
 
 const ResponsivePopup = props => {
-  const {isPhone} = useResponsive();
-  return <Popup {...props} disabled={isPhone} />;
+  const {isPhone, isTablet} = useResponsive();
+  return <Popup {...props} disabled={isPhone || isTablet} />;
 };
 
 export default ResponsivePopup;
--- a/indico/web/client/js/react/components/StickyWithScrollBack.jsx
+++ b/indico/web/client/js/react/components/StickyWithScrollBack.jsx
@@ -10,16 +10,22 @@ import PropTypes from 'prop-types';
 import {Sticky} from 'semantic-ui-react';
 
 import {ScrollButton} from 'indico/react/components';
+import {useResponsive} from 'indico/react/util';
+
+import {useScrollUp} from '../hooks';
 
 import './StickyWithScrollBack.module.scss';
 
-export default function StickyWithScrollBack({children, context}) {
+export default function StickyWithScrollBack({children, context, responsive}) {
   const [scrollButtonVisible, setScrollButtonVisible] = useState(false);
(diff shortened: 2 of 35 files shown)


package.json
```diff
@@ -64,7 +64,7 @@
     "react-modal": "^3.9.1",
     "react-popper": "^1.3.3",
     "react-redux": "^7.1.0",
-    "react-responsive": "^7.0.0",
+    "react-responsive": "8.0.0-beta.1",
```
ThiefMaster: why not `^8.0.0-beta.1`?

indico/web/client/js/react/util/Responsive.jsx
```diff
@@ -60,3 +60,18 @@ export default Object.assign(Responsive, {
   Tablet: factory(DIMENSIONS.tablet, DIMENSIONS.computer),
   Phone: factory(null, DIMENSIONS.tablet),
 });
+
+export function useResponsive() {
+  return {
+    isWideScreen: useMediaQuery({query: `(min-width: ${DIMENSIONS.computer}px)`}),
+    isDesktop: useMediaQuery({
+      query: `(min-width: ${DIMENSIONS.computer}px) and (max-width: ${DIMENSIONS.wideScreen -
+        1}px)`,
```
ThiefMaster: this wrapping here is super awful :/

indico/web/client/js/react/hooks.js
```diff
@@ -119,3 +119,31 @@ export function useIndicoAxios({camelize, ...args}) {
...
+      prevRef.current = window.scrollY;
+      window.addEventListener('scroll', _.throttle(handleScroll, 500));
+
+      return function cleanup() {
+        window.removeEventListener('scroll', _.throttle(handleScroll, 500));
+      };
+    }
+  }, [isScrollingUp, enabled]);
+
+  return {isScrollingUp};
```
ThiefMaster: just return a plain value instead of an object

indico/web/client/js/react/hooks.js
```diff
@@ -119,3 +119,31 @@ export function useIndicoAxios({camelize, ...args}) {
...
+        setIsScrollingUp(prevRef.current > window.scrollY);
+        prevRef.current = window.scrollY;
+      }
+    };
+
+    if (enabled) {
+      prevRef.current = window.scrollY;
+      window.addEventListener('scroll', _.throttle(handleScroll, 500));
+
+      return function cleanup() {
```
ThiefMaster: arrow function

indico/web/client/js/react/components/ResponsivePopup.jsx
```diff
@@ -0,0 +1,17 @@
...
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React from 'react';
+import {Popup} from 'semantic-ui-react';
+import {useResponsive} from 'indico/react/util';
+
+const ResponsivePopup = props => {
```
ThiefMaster: `export default function ResponsivePopup(props) { ...`

indico/web/client/js/react/components/StickyWithScrollBack.jsx
```diff
@@ -10,16 +10,22 @@ import PropTypes from 'prop-types';
...
+import {useResponsive} from 'indico/react/util';
+
+import {useScrollUp} from '../hooks';
 
 import './StickyWithScrollBack.module.scss';
 
-export default function StickyWithScrollBack({children, context}) {
+export default function StickyWithScrollBack({children, context, responsive}) {
   const [scrollButtonVisible, setScrollButtonVisible] = useState(false);
+  const {isScrollingUp} = useScrollUp(responsive);
```
ThiefMaster: no need for object unpacking here if you return a plain value from the hook

indico/web/client/js/react/util/Responsive.jsx
ThiefMaster: defining

    const belowWideScreen = DIMENSIONS.wideScreen - 1;

and then using this here would avoid the wrapping

indico/web/client/js/react/components/ResponsivePopup.jsx
mvidalgarcia: Is this because of the naming problem when importing it?

ThiefMaster: i think in this particular case the name would be derived correctly, but generally yes (and CRA etc. all use `function` for their top-level functions representing components)

ThiefMaster:
The booking details tooltip on the calendar bars  should probably be disabled on all touch devices (is there a media query for it)? 

---

Can we hide more of the UI in portrait mode here? It looks a bit strange to have e.g. the legend there.

image

ThiefMaster:
More UI to hide in portrait mode:

image

ThiefMaster:
> Error: Invalid LatLng object: (NaN, NaN)

I guess that's because the map is hidden? I don't think this happens if the map is generally disabled, so worth investigating why this happens here.

indico/web/client/js/react/util/Responsive.jsx
```diff
@@ -54,9 +59,58 @@ const factory = (minDimension, maxDimension) => {
...
+  };
+
+  return _OrientationSpec;
+};
+
 export default Object.assign(Responsive, {
   WideScreen: factory(DIMENSIONS.wideScreen, null),
   Desktop: factory(DIMENSIONS.computer, DIMENSIONS.wideScreen),
   Tablet: factory(DIMENSIONS.tablet, DIMENSIONS.computer),
   Phone: factory(null, DIMENSIONS.tablet),
```
ThiefMaster: maybe we can rename this to `dimensionFactory`?

mvidalgarcia:
> The only things I noticed so far is a problem I encountered on my Phone (Galaxy S8)
> There is no enough right margin in a few places which makes some elements partially invisible.
> _[...]_

Could you please check now and let me know? It's up on my server

PR #4016, opened 2019-07-23
Add Author List button in Paper Management

### LINKED ISSUE #3978: Add author list button to CfP list
Event organizers often need to send emails to authors that haven't sent their papers yet.
The idea is implementing an _Author list_ button in the Call for Papers list as we have in abstract and contribution modules. From there sending emails must be possible.

`RQF1335983`

panagiotappl:
closes #3978

### DIFF
--- a/indico/modules/events/papers/templates/_paper_list.html
+++ b/indico/modules/events/papers/templates/_paper_list.html
@@ -351,6 +351,14 @@
                     </a>
                 </li>
             </ul>
+            <button class="i-button icon-users js-enable-if-checked disabled"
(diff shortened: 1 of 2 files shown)


indico/modules/events/papers/templates/_paper_list.html
```diff
@@ -351,6 +351,14 @@
                     </a>
                 </li>
             </ul>
+            <button class="i-button icon-users js-enable-if-checked disabled"
+                        data-title="{% trans %}Author list{% endtrans %}"
```
mic4ael: I think you should align the following lines with the `class` attribute.

PR #4019, opened 2019-07-29
Sort Surveys by title on display page

### LINKED ISSUE #3802: Surveys: Sort by title on display page
Currently we use whatever order the DB gives us.

Easiest way would be adding the sorting to `query_active_surveys` and removing it from code that applies it after calling that function.

panagiotappl:
closes #3802

### DIFF
--- a/indico/modules/events/surveys/__init__.py
+++ b/indico/modules/events/surveys/__init__.py
@@ -54,9 +54,7 @@ def _get_active_surveys(event):
     if not event.has_feature('surveys'):
         return []
     from indico.modules.events.surveys.models.surveys import Survey
-    return (query_active_surveys(event)
-            .order_by(db.func.lower(Survey.title))
-            .all())
+    return query_active_surveys(event).all()
 
 
 @template_hook('event-header')
--- a/indico/modules/events/surveys/util.py
+++ b/indico/modules/events/surveys/util.py
@@ -139,4 +139,6 @@ def query_active_surveys(event):
     if session.user:
         user_pending_criterion = ~SurveySubmission.is_submitted & (SurveySubmission.user == session.user)
         private_criterion |= Survey.submissions.any(user_pending_criterion)
-    return Survey.query.with_parent(event).filter(Survey.is_active, private_criterion)
+    return (Survey.query.with_parent(event)
+            .filter(Survey.is_active, private_criterion)
+            .order_by(db.func.lower(Survey.title)))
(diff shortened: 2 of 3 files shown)


indico/modules/events/surveys/__init__.py
```diff
@@ -55,7 +55,6 @@ def _get_active_surveys(event):
         return []
     from indico.modules.events.surveys.models.surveys import Survey
     return (query_active_surveys(event)
-            .order_by(db.func.lower(Survey.title))
```
ThiefMaster: this now fits nicely in one line: `return query_active_surveys(event).all()`

indico/modules/events/surveys/util.py
```diff
@@ -139,4 +139,8 @@ def query_active_surveys(event):
     if session.user:
         user_pending_criterion = ~SurveySubmission.is_submitted & (SurveySubmission.user == session.user)
         private_criterion |= Survey.submissions.any(user_pending_criterion)
-    return Survey.query.with_parent(event).filter(Survey.is_active, private_criterion)
+    return (
```
mvidalgarcia: Shouldn't the wrapping be like this in python?
```python
return (Survey.query.with_parent(event)
        .filter(Survey.is_active, private_criterion)
        .order_by(db.func.lower(Survey.title)))
```

mic4ael: I guess it was `black` formatting it this way..

panagiotappl: I wasn't really sure about the wrapping but this was suggested by black. I decided to follow black since I wasn't sure.

mvidalgarcia: Oh, didn't know we were using `black` in Indico

panagiotappl: We are not.. I just wanted a suggestion on a formatting and I used that tool for it. If you want I can change it to the one you suggested.

mic4ael: Actually @mvidalgarcia is right. We are not using `black` for Indico for now.

panagiotappl: Changed it to @mvidalgarcia 's suggestion. As I explained, I wasn't trying to add `black` in indico and I know we are not using it. I just took this suggestion as I was unsure about indentation at that part.

ThiefMaster: black in indico is not gonna happen for now (but something to consider when we go for python 3) - so best to not have it enabled in your editor either.

Anyway, I think asking tools like black and prettier for advice when you are unsure on how to indent something is generally not a bad idea. However for chained calls in queries we already have a good style we use everywhere in Indico ;)

PR #4020, opened 2019-07-30
Always explicitly install react-jsx-i18n

mic4ael:


setup.py
```diff
@@ -47,6 +47,7 @@ def _compile_languages_react(self):
             if not os.path.exists(po_file):
                 continue
             with open(os.devnull, 'w') as devnull:
+                subprocess.check_call(['npm', 'install', 'react-jsx-i18n'], stdout=devnull, stderr=devnull)
```
ThiefMaster: that shouldn't be needed since npx downloads all requirements to a temporary location to run something that is not installed. And I really do not like the idea of a setup script installing new packages to the local environment while it's just supposed to be building stuff.

mic4ael: Yep, that is what I mentioned on mattermost - this PR is not needed if we fix the issue in the `react-jsx-i18n`.

mic4ael: Basically I can close the PR anyway.

PR #4024, opened 2019-08-07
Track groups

mic4ael:
So, `update/create` operations are just stubs for now since I am not sure at the moment how we will let users assign tracks to track groups.

At the moment I see two options:
- Using one form to `create/update` a `TrackGroup` which would proabably include a new field allowing to select which tracks belong to the new/updated track group;
- Using the `TrackGroupForm` only for stuff like e.g. `title` and then have a separate place in which users would assign tracks to track groups.

### DIFF
--- a/indico/modules/events/tracks/blueprint.py
+++ b/indico/modules/events/tracks/blueprint.py
@@ -7,8 +7,10 @@
 
 from __future__ import unicode_literals
 
-from indico.modules.events.tracks.controllers import (RHCreateTrack, RHDeleteTrack, RHDisplayTracks, RHEditProgram,
-                                                      RHEditTrack, RHManageTracks, RHSortTracks, RHTracksPDF)
+from indico.modules.events.tracks.controllers import (RHCreateTrack, RHCreateTrackGroup, RHDeleteTrack,
+                                                      RHDeleteTrackGroup, RHDisplayTracks, RHEditProgram, RHEditTrack,
+                                                      RHEditTrackGroup, RHManageTracks, RHSortTracks, RHTrackGroups,
+                                                      RHTracksPDF)
 from indico.web.flask.util import make_compat_redirect_func
 from indico.web.flask.wrappers import IndicoBlueprint
 
@@ -22,6 +24,13 @@ _bp.add_url_rule('/manage/tracks/create', 'create_track', RHCreateTrack, methods
 _bp.add_url_rule('/manage/tracks/sort', 'sort_tracks', RHSortTracks, methods=('POST',))
 _bp.add_url_rule('/manage/tracks/<int:track_id>', 'edit_track', RHEditTrack, methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/tracks/<int:track_id>', 'delete_track', RHDeleteTrack, methods=('DELETE',))
+
+_bp.add_url_rule('/manage/track-groups', 'track_groups', RHTrackGroups)
+_bp.add_url_rule('/manage/track-groups/create', 'create_track_group', RHCreateTrackGroup, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/track-groups/<int:track_group_id>', 'edit_track_group', RHEditTrackGroup,
+                 methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/track-groups/<int:track_group_id>', 'delete_track_group', RHDeleteTrackGroup,
+                 methods=('DELETE',))
 
 _bp.add_url_rule('/program', 'program', RHDisplayTracks)
 _bp.add_url_rule('/program.pdf', 'program_pdf', RHTracksPDF)
--- a/indico/modules/events/tracks/controllers.py
+++ b/indico/modules/events/tracks/controllers.py
@@ -17,9 +17,11 @@ from indico.core.db.sqlalchemy.descriptions import RENDER_MODE_WRAPPER_MAP
 from indico.legacy.pdfinterface.conference import ProgrammeToPDF
 from indico.modules.events.controllers.base import RHDisplayEventBase
 from indico.modules.events.management.controllers import RHManageEventBase
-from indico.modules.events.tracks.forms import ProgramForm, TrackForm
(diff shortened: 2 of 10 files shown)


indico/modules/events/tracks/models/tracks.py
```diff
@@ -94,6 +100,14 @@ class Track(DescriptionMixin, db.Model):
         lazy=True,
         backref='default_for_tracks'
     )
+    group = db.relationship(
+        'TrackGroup',
+        lazy=True,
+        backref=db.backref(
+            'tracks',
+            lazy='dynamic'
```
ThiefMaster: do we really need/want `lazy='dynamic'` here?

indico/modules/events/tracks/models/groups.py
```diff
@@ -0,0 +1,42 @@
...
+
+from __future__ import unicode_literals
+
+from indico.core.db.sqlalchemy import db
+from indico.core.db.sqlalchemy.descriptions import DescriptionMixin, RenderMode
+from indico.util.locators import locator_property
+from indico.util.string import format_repr, return_ascii, text_to_repr
+
+
+class TrackGroup(DescriptionMixin, db.Model):
```
ThiefMaster: this need a FK+relationship to the Event

indico/modules/events/tracks/models/groups.py
```diff
@@ -0,0 +1,42 @@
...
+        db.String,
+        nullable=False
+    )
+
+    # relationship backrefs:
+    # - tracks (Track.group)
+
+    @locator_property
+    def locator(self):
+        return dict(self.track.locator, group_id=self.id)
```
ThiefMaster: this isn't correct; there's no `self.track` here.

indico/modules/events/tracks/controllers.py
```diff
@@ -53,6 +55,18 @@ def _process_args(self):
...
+class RHManageTrackGroupBase(RHManageEventBase):
+    normalize_url_spec = {
+        'locators': {
+            lambda self: self.track_group
+        }
+    }
+
+    def _process_args(self):
+        RHManageEventBase._process_args(self)
+        self.track_group = TrackGroup.get_one(request.view_args['group_id'])
```
ThiefMaster: I would call this `track_group_id` (also in the models etc) to avoid confusion with user groups

indico/migrations/versions/20190731_1515_eefba82b42c5_add_track_groups_table.py
```diff
@@ -0,0 +1,35 @@
...
+def upgrade():
+    op.create_table(
+        'track_groups',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('title', sa.String(), nullable=False),
+        sa.Column('description', sa.Text(), nullable=False, server_default=''),
+        sa.PrimaryKeyConstraint('id'),
+        schema='events'
+    )
+    op.add_column('tracks', sa.Column('group_id', sa.Integer(), nullable=True), schema='events')
```
ThiefMaster: you also need to create the index since you have `index=True` in the model

indico/modules/events/tracks/models/tracks.py
mic4ael: I think I used it here since I wanted to filter the `tracks` but I don't think we are gonna need it so probably I don't need `dynamic` here.

ThiefMaster:
Since a track can only be in one group I'd make it an option in the `TrackForm`. That way we don't need a new field/widget and if people first define the track groups and then create tracks, it's also more straightforward.

mic4ael:
> Since a track can only be in one group I'd make it an option in the `TrackForm`. That way we don't need a new field/widget and if people first define the track groups and then create tracks, it's also more straightforward.

But this would make it a bit cumbersome to assign multiple tracks to one track group. You would have update all of them one by one.

indico/modules/events/tracks/operations.py
```diff
@@ -43,3 +44,25 @@ def update_program(event, data):
     track_settings.set_multi(event, data)
     logger.info('Program of %r updated by %r', event, session.user)
     event.log(EventLogRealm.management, EventLogKind.change, 'Tracks', 'The program has been updated', session.user)
+
+
+def create_track_group(event, data):
+    track_group = TrackGroup()
```
panagiotappl: Shouldn't the `track_group` be associated with the `event` here?

mic4ael: Yup, forgot to add it after I created a relationship between `TrackGroup` and `Event`.

mic4ael: Updated

panagiotappl: Okay, I'm adding it in my PR. Just letting you know.

mic4ael: I have just updated my PR ;) you can update

indico/modules/events/tracks/blueprint.py
```diff
@@ -23,6 +25,11 @@
 _bp.add_url_rule('/manage/tracks/<int:track_id>', 'edit_track', RHEditTrack, methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/tracks/<int:track_id>', 'delete_track', RHDeleteTrack, methods=('DELETE',))
 
+_bp.add_url_rule('/manage/track-groups', 'track_groups', RHTrackGroups)
+_bp.add_url_rule('/manage/track-groups/create', 'create_track_group', RHCreateTrackGroup, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/track-groups/<int:group_id>', 'edit_track_group', RHEditTrackGroup, methods=('GET', 'POST'))
```
panagiotappl: Are we going with `group_id` or `track_group_id`? I think I've noticed it written differently in some parts.

mic4ael: Should be `track_group_id`, let me fix that ;)

mic4ael: Updated and thanks for noticing that!

PR #4025, opened 2019-08-07
Remove ExternalIDs label when useless

### LINKED ISSUE #3857: Hide "external IDs" in event settings if none are defined
We currently show the field for this setting even if no external ID types are defined. This makes no sense since it's useless in that case.

aristofanischionis:
Make a query to see if there are any ExternalIDs
closes #3857

### DIFF
--- a/indico/modules/events/management/controllers/settings.py
+++ b/indico/modules/events/management/controllers/settings.py
@@ -19,6 +19,7 @@ from indico.modules.events.management.forms import (EventClassificationForm, Eve
                                                     EventDatesForm, EventLocationForm, EventPersonsForm)
 from indico.modules.events.management.util import flash_if_unregistered
 from indico.modules.events.management.views import WPEventSettings, render_event_management_header_right
+from indico.modules.events.models.references import ReferenceType
 from indico.modules.events.operations import update_event
 from indico.modules.events.util import track_time_changes
 from indico.modules.rb.models.reservation_occurrences import ReservationOccurrence
@@ -65,8 +66,10 @@ class RHEventSettings(RHManageEventBase):
                            .join(Room)
                            .has_rows())
             show_booking_warning = not has_overlap
+        has_reference_types = ReferenceType.query.has_rows()
         return WPEventSettings.render_template('settings.html', self.event, 'settings',
-                                               show_booking_warning=show_booking_warning)
+                                               show_booking_warning=show_booking_warning,
+                                               has_reference_types=has_reference_types)
 
 
 class RHEditEventDataBase(RHManageEventBase):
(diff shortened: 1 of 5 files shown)


indico/modules/events/management/forms.py
```diff
@@ -211,7 +211,8 @@ class EventClassificationForm(IndicoForm):
     def __init__(self, *args, **kwargs):
         event = kwargs.pop('event')
         super(EventClassificationForm, self).__init__(*args, **kwargs)
-        if event.type_ != EventType.meeting:
+
```
ThiefMaster: I don't think we need this blank line but no problem with having it either ;)

indico/modules/events/management/controllers/settings.py
```diff
@@ -65,6 +66,11 @@ def _process(self):
                            .join(Room)
                            .has_rows())
             show_booking_warning = not has_overlap
+        if not ReferenceType.query.has_rows():
```
ThiefMaster: has_reference_type = ReferenceType.query.has_rows()

indico/modules/events/management/controllers/settings.py
```diff
@@ -65,6 +66,11 @@ def _process(self):
...
+            has_reference_type = False
+        else:
+            has_reference_type = True
+        self.event.has_reference_type = has_reference_type
```
ThiefMaster: never set random stuff on the event object! pass it as a kwarg to the template instead. (also, it should be plural `has_reference_types`)

indico/modules/events/management/controllers/settings.py
```diff
@@ -79,7 +85,12 @@ def render_form(self, form):
     def render_settings_box(self):
         tpl = get_template_module('events/management/_settings.html')
         assert self.section_name
-        return tpl.render_event_settings(self.event, section=self.section_name, with_container=False)
+        if not ReferenceType.query.has_rows():
```
ThiefMaster: same as above

indico/modules/events/management/controllers/settings.py
```diff
@@ -79,7 +85,12 @@ def render_form(self, form):
...
+            has_reference_type = False
+        else:
+            has_reference_type = True
+        return tpl.render_event_settings(self.event,
+                                         has_reference_type, section=self.section_name, with_container=False)
```
ThiefMaster: the wrapping is a bit weird here

indico/modules/events/management/templates/settings.html
```diff
@@ -1,6 +1,5 @@
 {% extends 'events/management/base.html' %}
 {% from 'events/management/_settings.html' import render_event_settings %}
-
```
ThiefMaster: let's keep that one

indico/modules/events/management/templates/settings.html
```diff
@@ -27,6 +26,6 @@
         </div>
     {%- endif -%}
     <div class="action-box event-settings">
-        {{ render_event_settings(event) }}
+        {{ render_event_settings(event, event.has_reference_type) }}
```
ThiefMaster: without `event.` (see my earlier comment)

PR #4035, opened 2019-08-20
BoA TeX export

mvidalgarcia:
TODO:
- [x] Use BytesIO download the zip file directly instead of caching


indico/modules/events/contributions/util.py
```diff
@@ -268,3 +303,40 @@ def import_contributions_from_csv(event, f):
...
+    temp_file = NamedTemporaryFile(suffix='indico.tmp', dir=config.TEMP_DIR)
+    with ZipFile(temp_file.name, 'w', allowZip64=True) as zip_handler:
+        for dirpath, dirnames, files in os.walk(tex.source_dir, followlinks=True):
+            for f in files:
+                if f.startswith('.'):
+                    continue
+                if f.endswith(('.py', '.pyc', '.pyo')):
+                    continue
+                pf = os.path.join(dirpath, f)
+                af = os.path.abspath(pf)
```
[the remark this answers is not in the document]
mvidalgarcia: :man_shrugging:

indico/legacy/pdfinterface/latex.py
```diff
@@ -176,6 +176,16 @@ def run(self, template_name, **kwargs):
         with codecs.open(source_filename, 'wb', encoding='utf-8') as f:
             f.write(source)
 
+        distribution = pkg_resources.get_distribution('indico-fonts')
+        font_dir = os.path.join(distribution.location, 'indico_fonts')
+        try:
+            os.symlink(font_dir, os.path.join(self.source_dir, 'fonts'))
+        except OSError:
```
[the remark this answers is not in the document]
mvidalgarcia: Maybe it shouldn't fail at all.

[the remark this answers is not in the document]
ThiefMaster: So I would rather symlink only - and then, if it ever fails for some reason, we can then fix it (after finding out why it failed)

indico/modules/events/abstracts/templates/management/overview.html
```diff
@@ -94,9 +94,17 @@
                 <div class="label">
                     {%- trans %}Book of abstracts{% endtrans -%}
                 </div>
-                {%- trans url=url_for('.export_boa', event) -%}
-                    Customize and <a href={{ url }}>download</a> the book of abstracts
+                {%- trans -%}
+                    Customize and <a class="arrow js-dropdown" data-toggle="dropdown">download</a> the book of abstracts
```
ThiefMaster: I think it's time to use `{{ link_start }}` and `{{ link_end }}` in here - it's quite ugly to have that much HTML in the translatable string. If you search for `{% set link_start %}` you'll find some examples.

indico/modules/events/abstracts/blueprint.py
```diff
@@ -44,6 +44,7 @@
 # Book of Abstracts
 _bp.add_url_rule('/manage/abstracts/boa', 'manage_boa', boa.RHManageBOA, methods=('GET', 'POST'))
 _bp.add_url_rule('/book-of-abstracts.pdf', 'export_boa', boa.RHExportBOA)
+_bp.add_url_rule('/book-of-abstracts.zip', 'export_boa_tex', boa.RHExportBOATeX)
```
ThiefMaster: I think this should be manager-only. There is no need a participant of the event needs to get a zip archive with the latex source code of the BoA. And by making it manager-only there is also no need to do any caching, so we don't need to complicate those functions to handle separate caches.

That way you could even use a BytesIO to write the zip file to instead of creating it on disk.

ThiefMaster:
The spacing at the bottom looks strange now:

image

mvidalgarcia:
> The spacing at the bottom looks strange now:
> 
> image

I think it slightly changed on `master` as well, maybe related with SUI? A comparative:

`master`:
image

Production:
image

indico/modules/events/contributions/controllers/management.py
```diff
@@ -454,38 +461,45 @@ def _process(self):
...
-class RHContributionsExportPDFBookSortedAll(RHManageContributionsExportActionsBase):
     def _process(self):
-        pdf = ContributionBook(self.event, session.user, self.contribs, tz=self.event.timezone,
-                               sort_by=BOASortField.session_schedule_board)
-        return send_file('book-of-abstracts.pdf', pdf.generate(), 'application/pdf')
+        config_params = export_list_cache.get(request.view_args['uuid'])
+        output_format = config_params['format']
+        sort_by = config_params['sort_by']
+        contribs = (Contribution.query.with_parent(self.event)
+                    .filter(Contribution.id.in_(config_params['contribution_ids'])).all())
```
mic4ael: Maybe put `.all()` in the next line?

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -151,6 +120,21 @@
...
                             </a>
                         </li>
+                        <li>
+                            <a href ="#" class="icon-file-text js-submit-form js-enable-if-checked disabled"
+                               data-href="{{ url_for('.contributions_tex_export', event) }}">
+                                {%- trans %}LaTeX{% endtrans -%}
+                            </a>
+                        </li>
+                        <li>
+                            <a href ="#" class="icon-file-text js-enable-if-checked disabled"
```
mic4ael:
```suggestion
                            <a href="#" class="icon-file-text js-enable-if-checked disabled"
```

indico/modules/events/contributions/util.py
```diff
@@ -268,3 +303,38 @@ def import_contributions_from_csv(event, f):
...
+
+def render_archive(event, contribs, sort_by, opts):
+    pdf = opts(event, session.user, contribs, tz=event.timezone, sort_by=sort_by)
+    pdf.generate(return_source=True)
+    return send_file('contributions-tex.zip', zip_tex_file(pdf), 'application/zip', inline=False)
+
+
+def zip_tex_file(tex):
+    temp_file = NamedTemporaryFile(suffix='indico.tmp', dir=config.TEMP_DIR)
+    with ZipFile(temp_file.name, 'w', allowZip64=True) as zip_handler:
```
mic4ael: `allowZip64` is `True` by default

ThiefMaster: Not in Python 2...

```
$ grep '__init__.*allowZip64' /usr/lib/python2.7/zipfile.py
    def __init__(self, file, mode="r", compression=ZIP_STORED, allowZip64=False):
```

PR #4045, opened 2019-08-23
Remove pyatom from dependencies

mic4ael:


pferreir:
We should find a better solution to this before we move to Werkzeug 1.0, since this contrib module will be removed by then.

PR #4054, opened 2019-09-06
Add marshmallow post load

pferreir:
This PR adds a signal that will be needed for https://github.com/indico/indico-plugins-cern/pull/65.
It also makes it so that errors in the room editing modal are properly displayed. There should now be a button that allows the user to reset the form in case there are issues (the previous behaviour was to reset it by default, which also erased the error messages).

image

### DIFF
--- a/indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
+++ b/indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
@@ -662,17 +662,18 @@ class RoomEditModal extends React.Component {
       await this.saveEquipment(roomId, availableEquipment);
       await this.saveAttributes(roomId, attributes);
       await this.saveAvailability(roomId, changedValues, nonbookablePeriods, bookableHours);
+
+      // reload room so the form gets new initialValues
+      await this.fetchRoomData(roomId);
+      this.setState({
+        wasEverSaved: true,
+        submitState,
+      });
     } catch (e) {
       submitError = handleSubmitError(e);
       submitState = 'error';
+      this.setState({submitState});
     }
-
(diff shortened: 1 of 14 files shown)


indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -931,7 +932,21 @@ class RoomEditModal extends React.Component {
...
+                  <Button
+                    type="button"
+                    icon="undo"
+                    size="mini"
+                    color="red"
+                    content={Translate.string('Reset form')}
+                    onClick={async () => {
+                      // reset the form
+                      this.setState({submitState: ''});
+                      await this.fetchRoomData(roomId);
```
ThiefMaster: I don't think `await`ing has any benefit here, so just calling it (and not making the onClick handler async) sohuld be enough.

ThiefMaster: Actually, you can simply call `fprops.form.reset();` to reset the form without the need for extra requests.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -931,7 +932,21 @@ class RoomEditModal extends React.Component {
...
+                  <p>
+                    <Translate>Room could not be updated.</Translate>
+                  </p>
+                  <Button
+                    type="button"
+                    icon="undo"
+                    size="mini"
+                    color="red"
+                    content={Translate.string('Reset form')}
+                    onClick={async () => {
```
mic4ael: You don't need `async` here.

PR #4057, opened 2019-09-09
Add endpoint that exports dashboard as ICS

pferreir:
image

image

I also made the dashboard a little bit less claustrophobic and more readable (and responsive!):
image

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -25,6 +25,7 @@ settings:
       map:
         - ['indico/modules/core', './indico/modules/core/client/js']
         - ['indico/modules/rb', './indico/modules/rb/client/js']
+        - ['indico/modules/users', './indico/modules/users/client/js']
         - ['indico', './indico/web/client/js']
       extensions: [.js, .jsx, .json]
   react:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ script:
   - isort -rc --diff --check-only indico/
   - python bin/maintenance/update_backrefs.py --ci
   - npx eslint --ext .js --ext .jsx indico/modules/rb/ indico/modules/events/logs/ indico/web/client/js/react/
+    indico/modules/users/
   - pytest
   - python bin/maintenance/build-assets.py indico --dev
 notifications:
--- a/indico/modules/api/templates/user_profile.html
+++ b/indico/modules/api/templates/user_profile.html
@@ -91,6 +91,33 @@
             </div>
         </div>
 
+        <div class="i-box">
+            <div class="i-box-header">
+                <div class="i-box-title">
+                    {%- trans %}Token-based links{% endtrans -%}
+                </div>
+                <div class="i-box-buttons toolbar">
+                    <div class="group">
+                        <button type="button"
+                                class="i-button warning icon-remove"
+                                data-href="{{ url_for('core.reset_signature_tokens') }}"
+                                data-method="POST"
+                                data-confirm="{% trans %}Are you sure you'd like to invalidate all your token-based links?{% endtrans %}">
+                            {%- trans %}Invalidate all links{% endtrans -%}
+                        </button>
+                    </div>
+                </div>
+            </div>
+            <div class="i-box-content">
+                {% trans %}
+                    Token-based unique links are secret URLs which Indico uses to allow you to easily fetch data
+                    from within other applications (e.g. calendar schedulers through iCalendar). They are generated
+                    through a digital signature based on a secret key that is unique per Indico user. If one of your
+                    links is compromised, you should click this button. All other links will be invalidated.
+                {% endtrans %}
+            </div>
+        </div>
+
         {% if session.user.is_admin and (key or old_keys) %}
             <div class="i-box">
                 <div class="i-box-header">
--- a/indico/modules/core/controllers.py
+++ b/indico/modules/core/controllers.py
@@ -31,6 +31,7 @@ from indico.modules.cephalopod import cephalopod_settings
 from indico.modules.core.forms import ReportErrorForm, SettingsForm
 from indico.modules.core.settings import core_settings, social_settings
 from indico.modules.core.views import WPContact, WPSettings
+from indico.modules.users.controllers import RHUserBase
 from indico.util.i18n import _, get_all_locales
 from indico.util.marshmallow import PrincipalList
 from indico.util.user import principal_from_identifier
@@ -284,3 +285,10 @@ class RHSignURL(RHProtected):
         return jsonify(
             url=signed_url_for(session.user, endpoint, url_params=url_params, _external=True, **query_params)
         )
+
+
+class RHResetSignatureTokens(RHUserBase):
+    def _process(self):
+        self.user.reset_signing_secret()
+        flash(_('All your token-based links have been invalidated'), 'success')
+        return redirect(url_for('api.user_profile'))
--- a/indico/modules/core/controllers.py
+++ b/indico/modules/core/controllers.py
@@ -16,7 +16,7 @@ from packaging.version import Version
 from pkg_resources import DistributionNotFound, get_distribution
 from pytz import common_timezones_set
 from webargs import fields
-from werkzeug.exceptions import NotFound, ServiceUnavailable
+from werkzeug.exceptions import BadRequest, NotFound, ServiceUnavailable
 from werkzeug.urls import url_join, url_parse
 
 import indico
@@ -40,7 +40,7 @@ from indico.web.flask.templating import get_template_module
 from indico.web.flask.util import url_for
 from indico.web.forms.base import FormDefaults
 from indico.web.rh import RH, RHProtected
-from indico.web.util import jsonify_data, jsonify_form
+from indico.web.util import jsonify_data, jsonify_form, signed_url_for
 
 
 class RHContact(RH):
@@ -269,3 +269,18 @@ class RHPrincipals(RHProtected):
     def _process(self, values):
         return jsonify({identifier: self._serialize_principal(identifier, principal)
                         for identifier, principal in values.viewitems()})
+
+
+class RHSignURL(RHProtected):
+    def _process(self):
+        endpoint = request.json.get('endpoint')
+        if not endpoint:
+            raise BadRequest
+        # filter out non-standard args
+        url_params = request.json.get('url_params', {})
+        url_params = {k: v for k, v in url_params.viewitems() if not k.startswith('_')}
+        query_params = request.json.get('query_params', {})
+        query_params = {k: v for k, v in query_params.viewitems() if not k.startswith('_')}
+        return jsonify(
+            url=signed_url_for(session.user, endpoint, url_params=url_params, _external=True, **query_params)
+        )
--- a/indico/modules/core/controllers.py
+++ b/indico/modules/core/controllers.py
@@ -282,9 +282,9 @@ class RHSignURL(RHProtected):
         url_params = {k: v for k, v in url_params.viewitems() if not k.startswith('_')}
         query_params = request.json.get('query_params', {})
         query_params = {k: v for k, v in query_params.viewitems() if not k.startswith('_')}
-        return jsonify(
-            url=signed_url_for(session.user, endpoint, url_params=url_params, _external=True, **query_params)
-        )
+        url = signed_url_for(session.user, endpoint, url_params=url_params, _external=True, **query_params)
+        Logger.get('url_signing').info("%s signed URL for endpoint '%s' (%s)", session.user, endpoint, url)
+        return jsonify(url=url)
 
 
 class RHResetSignatureTokens(RHUserBase):
--- a/indico/modules/users/blueprint.py
+++ b/indico/modules/users/blueprint.py
@@ -10,9 +10,9 @@ from __future__ import unicode_literals
 from flask import request
 
 from indico.modules.users.api import RHUserFavoritesAPI, fetch_authenticated_user
-from indico.modules.users.controllers import (RHAcceptRegistrationRequest, RHAdmins, RHPersonalData,
-                                              RHRegistrationRequestList, RHRejectRegistrationRequest, RHUserDashboard,
-                                              RHUserEmails, RHUserEmailsDelete, RHUserEmailsSetPrimary,
+from indico.modules.users.controllers import (RHAcceptRegistrationRequest, RHAdmins, RHExportDashboardICS,
+                                              RHPersonalData, RHRegistrationRequestList, RHRejectRegistrationRequest,
+                                              RHUserDashboard, RHUserEmails, RHUserEmailsDelete, RHUserEmailsSetPrimary,
                                               RHUserEmailsVerify, RHUserFavorites, RHUserFavoritesCategoryAPI,
                                               RHUserFavoritesUserRemove, RHUserFavoritesUsersAdd, RHUserPreferences,
                                               RHUsersAdmin, RHUsersAdminCreate, RHUsersAdminMerge,
@@ -39,7 +39,6 @@ _bp.add_url_rule('!/admin/users/registration-requests/<int:request_id>/accept', 
 _bp.add_url_rule('!/admin/users/registration-requests/<int:request_id>/reject', 'reject_registration_request',
                  RHRejectRegistrationRequest, methods=('POST',))
 
-
 # User profile
 with _bp.add_prefixed_rules('/<int:user_id>'):
     _bp.add_url_rule('/dashboard/', 'user_dashboard', RHUserDashboard)
@@ -57,6 +56,8 @@ with _bp.add_prefixed_rules('/<int:user_id>'):
     _bp.add_url_rule('/emails/verify/<token>', 'user_emails_verify', RHUserEmailsVerify)
     _bp.add_url_rule('/emails/<email>', 'user_emails_delete', RHUserEmailsDelete, methods=('DELETE',))
     _bp.add_url_rule('/emails/make-primary', 'user_emails_set_primary', RHUserEmailsSetPrimary, methods=('POST',))
+
+_bp.add_url_rule('/<int:user_id>/dashboard.ics', 'export_dashboard_ics', RHExportDashboardICS)
 
 # User search
 _bp.add_url_rule('/search/info', 'user_search_info', RHUserSearchInfo)
--- a/indico/modules/users/client/js/ICSCalendarLink.jsx
+++ b/indico/modules/users/client/js/ICSCalendarLink.jsx
@@ -1,0 +1,142 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import signURL from 'indico-url:core.sign_url';
+
+import React, {useState} from 'react';
+import PropTypes from 'prop-types';
+import {Button, Dropdown, Icon, Input, Label, Grid, Popup, Header} from 'semantic-ui-react';
+
+import {Translate} from 'indico/react/i18n';
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+import {snakifyKeys} from 'indico/utils/case';
+
+export default function ICSCalendarLink({endpoint, urlParams, options, ...restProps}) {
+  const [copied, setCopied] = useState(false);
+  const [option, setOption] = useState(null);
+  const [open, setOpen] = useState(false);
+
+  const copyButton = (
+    <Button
+      icon="copy"
+      onClick={async () => {
+        await navigator.clipboard.writeText(option.url);
+        setCopied(true);
+      }}
+    />
+  );
+
+  const fetchURL = async queryParams => {
+    try {
+      const {
+        data: {url: signedURL},
+      } = await indicoAxios.post(signURL(), snakifyKeys({endpoint, urlParams, queryParams}));
+      return signedURL;
+    } catch (error) {
+      handleAxiosError(error);
+    }
+  };
+
+  const trigger = (
+    <Dropdown
+      {...restProps}
+      icon={null}
+      trigger={
+        <Button icon size="small">
+          <Icon name="calendar alternate outline" />
+          <Icon name="caret down" />
+        </Button>
+      }
+      pointing="top right"
+    >
+      <Dropdown.Menu>
+        <Dropdown.Header>
+          <Translate>Synchronise with your calendar</Translate>
+        </Dropdown.Header>
+        <Dropdown.Divider />
+        {options.map(({key, text, queryParams}) => (
+          <Dropdown.Item
+            key={key}
+            text={text}
+            onClick={async () => {
+              setOption({
+                text,
+                url: await fetchURL(queryParams),
+              });
+              setOpen(true);
+            }}
+          />
+        ))}
+      </Dropdown.Menu>
+    </Dropdown>
+  );
+
+  return (
+    <Popup
+      trigger={trigger}
+      position="left center"
+      open={open}
+      popperDependencies={[copied]}
+      on="click"
+      onOpen={() => {
+        setCopied(false);
+      }}
+      onClose={() => {
+        setOpen(false);
+      }}
+      wide
+    >
+      <Header
+        content={Translate.string('Sync with your calendar')}
+        subheader={option && <Label color="blue">{option.text}</Label>}
+      />
+      <Popup.Content>
+        <p>
+          <Translate>
+            You may copy-paste the following URL into your scheduling application. Contents will be
+            automatically synchronised.
+          </Translate>
+        </p>
+        <Input
+          placeholder={Translate.string('Loading…')}
+          readOnly
+          loading={!option}
+          size="mini"
+          fluid
+          value={option ? option.url : ''}
+          action={option && navigator.clipboard && copyButton}
+        />
+        {copied && (
+          <Grid centered>
+            <Grid.Row>
+              <Label pointing="above" color="teal">
+                <Translate>Copied!</Translate>
+              </Label>
+            </Grid.Row>
+          </Grid>
+        )}
+      </Popup.Content>
+    </Popup>
+  );
+}
+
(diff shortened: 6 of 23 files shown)


ThiefMaster:
I would strongly prefer the unique token, but in a different way; Store it on the User, and use it for all links of this kind (dashboard, category, etc); possibly through itsdangerous so the same token cannot be used on different links.

The advantage compared to storing separate tokens is that getting those links is idempotent and getting them multiple times won't create new tokens (or have to check the DB whether a token for a specific link already exists).

The only UI needed for that would be a "reset private links" button, telling the user that they have to update all links to protected dashboard, caetgory, etc. feeds they are using.

Speaking of category feeds: this would be a good opportunity to rip out the connection to the HTTP API from category/event ical links and use the token there - that's a bunch of legacy JS that could go away (currently we have stuff like creating an API key and enabling persistent api signatures in there).

indico/modules/users/controllers.py
```diff
@@ -102,37 +123,56 @@ class RHUserDashboard(RHUserBase):
...
+                                      user=self.user,
                                       categories=categories,
                                       categories_events=categories_events,
                                       suggested_categories=get_suggested_categories(self.user),
                                       linked_events=linked_events)
 
 
+class RHExportDashboardICS(RHUserBase):
+    @use_kwargs({
+        'from_': HumanizedDate(data_key='from', missing=now_utc(False) - relativedelta(weeks=1)),
```
ThiefMaster: that `missing` value is evaluated at import time..

indico/modules/users/controllers.py
```diff
@@ -102,37 +123,56 @@ class RHUserDashboard(RHUserBase):
...
+        'include': fields.List(fields.Str(), missing={'linked', 'categories'}),
+        'limit': fields.Integer(missing=100, validate=lambda v: v > 0 and v <= 500)
```
ThiefMaster: `0 < v <= 500`

indico/modules/users/controllers.py
```diff
@@ -102,37 +123,56 @@ class RHUserDashboard(RHUserBase):
...
+            categories_events = get_events_in_categories(category_ids, self.user, limit=limit)
+
+        linked_events = get_linked_events(
+            self.user,
+            from_,
+            limit=limit,
+            load_also=('description', 'own_room_id', 'own_venue_id', 'own_room_name', 'own_venue_name')
+        )
+
+        all_events = []
```
ThiefMaster: shouldn't this be a set to avoid duplicates?

indico/util/marshmallow.py
```diff
@@ -163,3 +174,23 @@ def _deserialize(self, value, attr, data):
...
+
+    def _deserialize(self, value, attr, obj):
+        m = HUMANIZED_DATE_RE.match(value)
+        if not m:
+            raise ValidationError("Can't parse humanized date!")
+
+        if m.groupdict().get('iso_date'):
+            return parser.isoparse(value)
+        else:
+            unit = m.groupdict()['unit']
```
ThiefMaster: Isn't there `.group('unit')` for that?

indico/modules/users/controllers.py
pferreir: Very good catch, thanks!

indico/modules/users/controllers.py
pferreir: Always forget the Python can do chained inequalities.

indico/util/marshmallow.py
pferreir: I remember telling myself "I should definitely check the re API before I send the PR, since I'm pretty sure there was a nicer way of doing this". That clearly didn't happen 😆

indico/modules/users/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -0,0 +1,135 @@
+import React, {useState} from 'react';
+import PropTypes from 'prop-types';
+import {Button, Dropdown, Icon, Input, Label, Grid, Popup, Header} from 'semantic-ui-react';
+
+import signURL from 'indico-url:core.sign_url';
```
ThiefMaster: I think in other files we have those imports on top. also, missing file header. `python bin/maintenance/update_header.py indico` should tell you about all files with missing headers

indico/modules/users/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -0,0 +1,135 @@
...
+        <Button icon size="small">
+          <Icon name="calendar alternate outline" />
+          <Icon name="caret down" />
+        </Button>
+      }
+      pointing="top right"
+    >
+      <Dropdown.Menu>
+        <Dropdown.Header>
+          <Translate>Synchronize with your calendar</Translate>
```
ThiefMaster: `Synchronise` :gb: ?

(at some point I'd like to move all source strings to en_US to be consistent with non-user-facing code and just use transifex for the en_GB-isms)

indico/modules/users/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -0,0 +1,135 @@
...
+import signURL from 'indico-url:core.sign_url';
+
+import {Translate} from 'indico/react/i18n';
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+import {snakifyKeys} from 'indico/utils/case';
+
+export default function ICSCalendarLink({endpoint, urlParams, options, ...restProps}) {
+  const [copied, setCopied] = useState(false);
+  const [option, setOption] = useState(null);
+  const [open, setOpen] = useState(false);
```
ThiefMaster: do we need state and a controlled component for this? `Popup` can handle this on its own when not using the `open` prop

indico/modules/users/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -0,0 +1,135 @@
...
+  );
+}
+
+ICSCalendarLink.propTypes = {
+  endpoint: PropTypes.string.isRequired,
+  urlParams: PropTypes.objectOf(PropTypes.string),
+  options: PropTypes.arrayOf(
+    PropTypes.shape({
+      key: PropTypes.string,
+      text: PropTypes.string,
```
ThiefMaster: `isRequired`? (same for key)

indico/modules/users/controllers.py
```diff
@@ -33,32 +33,53 @@
...
+                     Event.category_chain_overlaps(category_ids),
+                     Event.start_dt.astimezone(session.tzinfo) >= today)
+             .options(joinedload('category').load_only('id', 'title'),
+                      joinedload('series'),
+                      subqueryload('acl_entries'),
+                      load_only('id', 'category_id', 'start_dt', 'end_dt', 'title', 'access_key',
+                                'protection_mode', 'series_id', 'series_pos', 'series_count'))
+             .order_by(Event.start_dt, Event.id))
+    if limit:
+        query = query.limit(limit)
```
ThiefMaster: I don't think this is correct. `get_n_matching` takes care of it, so you probably want to pass the limit to that instead of the hardcoded `10` (and probably use `10` as the default value for `limit`).

indico/modules/users/blueprint.py
```diff
@@ -58,6 +57,8 @@
     _bp.add_url_rule('/emails/<email>', 'user_emails_delete', RHUserEmailsDelete, methods=('DELETE',))
     _bp.add_url_rule('/emails/make-primary', 'user_emails_set_primary', RHUserEmailsSetPrimary, methods=('POST',))
 
+_bp.add_url_rule('/<int:user_id>/dashboard.ics', 'export_dashboard_ics', RHExportDashboardICS)
```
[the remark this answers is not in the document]
ThiefMaster: If we use an unique-enough name like `user_token` we could even have an `RHTokenProtected` base RH that takes care of assigning the user and checking.

On a side note, this is where it sucks a bit that we went for `session.user` as the way to get the current user. Had we used a `current_user` proxy (and template global) we could just populate that when a token has been used and thus have access to the current user in the exact same way regardless of how they are authenticated (token/cookie/apikey/whatever)

indico/modules/users/models/users.py
```diff
@@ -500,6 +507,12 @@ def synced_values(self):
             return {}
         return {field: (identity.data.get(field) or '') for field in multipass.synced_fields}
 
+    @property
+    def signing_secret(self):
+        if not self._signing_secret:
+            self._signing_secret = str(uuid4())
```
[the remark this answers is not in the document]
ThiefMaster: Looks like there's even a way to generate UUIDs without an extension. Pretty ugly, but it'd only be in the upgrade script...

https://stackoverflow.com/a/21327318/298479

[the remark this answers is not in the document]
pferreir: Yeah, I thought of it but thought it would be too heavy/complex. But I agree, much better solution.

indico/modules/users/models/users.py
```diff
@@ -243,6 +244,12 @@ class User(PersonMixin, db.Model):
         nullable=False,
         default=False
     )
+    #: a unique secret used to generate signed URLs
+    _signing_secret = db.Column(
+        'signing_secret',
+        db.String,
```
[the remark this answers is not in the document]
ThiefMaster: Maybe it should be unique as well

.eslintrc.yml
```diff
@@ -25,6 +25,7 @@ settings:
       map:
         - ['indico/modules/core', './indico/modules/core/client/js']
         - ['indico/modules/rb', './indico/modules/rb/client/js']
+        - ['indico/modules/users', './indico/modules/rb/users/js']
```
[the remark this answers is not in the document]
pferreir: Oh, a typo. Tragic.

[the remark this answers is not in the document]
ThiefMaster: https://www.vice.com/en_us/article/4x3n9b/sometimes-a-typo-means-you-need-to-blow-up-your-spacecraft ;)

indico/modules/users/models/users.py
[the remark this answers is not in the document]
pferreir: Why? If we're to follow what we do with the [API secret](https://github.com/indico/indico/blob/master/indico/modules/api/models/keys.py#L39), then it shouldn't. In any case, it's such an unlikely thing...

indico/web/util.py
```diff
@@ -179,3 +182,32 @@ def get_request_info(hide_passwords=True):
 def url_for_index(_external=False, _anchor=None):
     from indico.web.flask.util import url_for
     return url_for('categories.display', _external=_external, _anchor=_anchor)
+
+
+def signed_url_for(user, blueprint, url_params=None, *args, **kwargs):
```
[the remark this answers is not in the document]
pferreir: Done.

indico/modules/users/models/users.py
```diff
@@ -243,6 +245,12 @@ class User(PersonMixin, db.Model):
         nullable=False,
         default=False
     )
+    #: a unique secret used to generate signed URLs
+    signing_secret = db.Column(
+        UUID,
+        nullable=True,
```
ThiefMaster: `nullable=False`

indico/modules/api/templates/user_profile.html
```diff
@@ -91,6 +91,33 @@
...
+                                data-confirm="{% trans %}Are you sure you'd like to invalidate all your token-based links?{% endtrans %}">
+                            {%- trans %}Invalidate all Links{% endtrans -%}
+                        </button>
+                    </div>
+                </div>
+            </div>
+            <div class="i-box-content">
+                {% trans %}
+                    Token-based unique links are secret URLs which Indico uses to allow you to easily fetch data
+                    from within other applications (e.g. calendar schedulers through ICS). They are generated through a
```
ThiefMaster: s/ICS/ICal?

indico/migrations/versions/20190925_1548_229bb73a558a_add_signature_token_column_to_user.py
```diff
@@ -0,0 +1,29 @@
...
+down_revision = 'f6fba869a27c'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # give every user a 'signing_secret' column with a random UUID
+    # thanks to the author of https://stackoverflow.com/a/21327318
+    op.execute('''
+        ALTER TABLE users.users ADD COLUMN signing_secret uuid unique not null default
```
ThiefMaster: I think you can keep using `op.add_column` and just put the default string inside `server_default` (probably as `sa.text()` or something similar that marks it as a literal sql string)

ThiefMaster: and of course remove the `server_default` afterwards

indico/web/util.py
```diff
@@ -179,3 +181,34 @@ def get_request_info(hide_passwords=True):
...
+    from indico.web.flask.util import url_for
+    _external = kwargs.pop('_external', False)
+    base_url = url_for(blueprint, *args, **(url_params or {}))
+    qs = url_encode(sorted(kwargs.items()))
+    url = '{}?{}'.format(base_url, qs) if qs else base_url
+    signer = Signer(user.signing_secret, salt='url-signing')
+    print 'signing', url
```
ThiefMaster: :eyes: :printer:

indico/modules/users/client/js/dashboard.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import ICSCalendarLink from 'indico/modules/users/react/components/ICSCalendarLink';
+
+import {Translate} from 'indico/react/i18n';
+
+document.addEventListener('DOMContentLoaded', () => {
+  const userId = document.querySelector('body').dataset.userId;
```
mic4ael: Maybe it is better to use `Indico.User.id` instead of searching the `DOM`

indico/modules/users/client/js/dashboard.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
```
mic4ael: Missing headers

indico/modules/users/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -0,0 +1,135 @@
...
+  );
+
+  return (
+    <Popup
+      trigger={trigger}
+      position="left center"
+      open={open}
+      popperDependencies={[copied]}
+      on="click"
+      onOpen={async () => {
```
mic4ael: Does it have to be `async`?

indico/web/util.py
```diff
@@ -179,3 +181,34 @@ def get_request_info(hide_passwords=True):
...
+    signature = params.pop('token')
+    url = url_unparse((
+        '',
+        '',
+        parsed.path,
+        url_encode(sorted(params.items()), sort=True),
+        parsed.fragment
+    ))
+    signer = Signer(user.signing_secret, salt='url-signing')
+    print url, signature, signer.verify_signature(url, signature)
```
mic4ael: Forgot to remove? :P

indico/modules/users/client/js/dashboard.jsx
ThiefMaster: hm when all you need is the ID, I think using the dataset is fine (but: `document.body.dataset.userId`); feels cleaner to me than using that global

indico/modules/users/blueprint.py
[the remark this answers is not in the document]
pferreir: What would be the advantage of going that way?

indico/modules/users/client/js/dashboard.jsx
```diff
@@ -0,0 +1,25 @@
...
+import ReactDOM from 'react-dom';
+import ICSCalendarLink from 'indico/modules/users/react/components/ICSCalendarLink';
```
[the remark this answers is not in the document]
pferreir: True.

indico/modules/users/blueprint.py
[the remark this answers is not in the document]
ThiefMaster: - not having to show the userid in the URL (it kind of invites people to try to change it, even though it of course won't do anything nasty)
- URL looks simpler - but since some of them have a query string anyway I guess it doesn't matter that much
- not sure if there aren't some calendaring tools that don't like query strings

anyway, I guess it boils down to personal preference

---

but regardless of the URL style, I would move the logic that checks the token into a custom RH subclass

indico/modules/users/client/js/dashboard.jsx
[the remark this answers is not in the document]
pferreir: Jolly good idea.

indico/modules/users/client/js/react/components/ICSCalendarLink.jsx
pferreir: We should run that as part of the commit hook.

ThiefMaster:
```
$ time python bin/maintenance/update_header.py indico
Updating headers to the year 2019 for all git-tracked files...
python bin/maintenance/update_header.py indico  1.93s user 0.04s system 99% cpu 1.977 total
```

Not sure whether I like that idea. CI is probably the better place for it.

indico/modules/users/client/js/react/components/ICSCalendarLink.jsx
pferreir: ESLint doesn't complain about missing `isRequired` within `.shape({})`. I never add those because of that...

indico/web/util.py
pferreir: 🤦‍♂

pferreir: It was actually added on a later fixup, because before opening a PR I grepped for `print` and `console` :D

indico/modules/users/client/js/react/components/ICSCalendarLink.jsx
pferreir: No. Must have been before 😉 . Thanks!

indico/web/util.py
pferreir: Debugging code I added because of the unit tests. We should really add a pre-commit hook to look for print and console ☹️

indico/modules/users/client/js/react/components/ICSCalendarLink.jsx
ThiefMaster: I guess it only affects the actual proptypes checking at runtime, whether it'll complain about it being undefined or not.

indico/web/util.py
```diff
@@ -179,3 +181,32 @@ def get_request_info(hide_passwords=True):
...
+    full_base_url = url_for(blueprint, *args, _external=_external, **(url_params or {}))
+    return '{}?{}'.format(full_base_url, qs) if qs else base_url
+
+
+def is_signed_url_valid(user, url):
+    parsed = url_parse(url)
+    params = url_decode(parsed.query)
+    if 'token' not in params:
+        return False
+    signature = params.pop('token')
```
ThiefMaster:
```python
try:
    signature = params.pop('token')
except KeyError:
    return False
```

-> no need for the check above anymore

(and especially now that the `[0]` is gone I think this is the cleanest, most readable and most pythonic option)

.travis.yml
```diff
@@ -15,7 +15,8 @@ script:
   - FORCE_COLOR=1 npx react-jsx-i18n extract --ext jsx indico/web/client/ indico/modules/ > /dev/null
   - isort -rc --diff --check-only indico/
   - python bin/maintenance/update_backrefs.py --ci
-  - npx eslint --ext .js --ext .jsx indico/modules/rb/ indico/modules/events/logs/ indico/web/client/js/react/
+  - npx eslint --ext .js --ext .jsx indico/modules/rb/ indico/modules/events/logs/ indico/web/client/js/react/ \
```
ThiefMaster: no need for the backslash (and it's breaking the build since it remains there as a literal backslash)

indico/modules/users/client/js/react/components/ICSCalendarLink.jsx
pferreir: We do, because we are opening the pop-up on the parent element instead of each option. Opening one per element would not be feasible since any click outside would close the dropdown menu (basically: a mess).

indico/modules/api/templates/user_profile.html
pferreir: I choose "[iCalendar](https://en.wikipedia.org/wiki/ICalendar)."

indico/modules/api/templates/user_profile.html
```diff
@@ -91,6 +91,33 @@
...
+            <div class="i-box-header">
+                <div class="i-box-title">
+                    {%- trans %}Token-based links{% endtrans -%}
+                </div>
+                <div class="i-box-buttons toolbar">
+                    <div class="group">
+                        <button type="button"
+                                class="i-button warning icon-remove"
+                                data-href="{{ url_for('core.reset_signature_tokens') }}"
+                                data-method="post"
```
mic4ael: `POST`?

indico/modules/users/client/js/ICSCalendarLink.jsx
```diff
@@ -0,0 +1,142 @@
...
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React, {useState} from 'react';
+import PropTypes from 'prop-types';
+import {Button, Dropdown, Icon, Input, Label, Grid, Popup, Header} from 'semantic-ui-react';
+
+import signURL from 'indico-url:core.sign_url';
```
mic4ael: Shouldn't that be at the top of the file?

pferreir: Yes, thanks!

indico/modules/users/client/js/dashboard.jsx
```diff
@@ -0,0 +1,33 @@
...
+import ICSCalendarLink from './ICSCalendarLink';
+
+document.addEventListener('DOMContentLoaded', () => {
+  const userId = document.querySelector('body').dataset.userId;
+  ReactDOM.render(
+    <ICSCalendarLink
+      endpoint="users.export_dashboard_ics"
+      urlParams={{user_id: userId}}
+      options={[
+        {key: 'events', text: Translate.string('Events at hand'), queryParams: {linked: 'true'}},
```
mic4ael: Do we have to use `'true'`? Wouldn't that work as well if we passed `true`?

pferreir: Query params can only be strings, so I find it less confusing if we pass them as a string, as opposed to relying on whatever implicit mechanism to do so.

mic4ael: Yeah, I was just curious whether those query params get converted to string automatically an exception is thrown in case you pass something else than a string.

indico/modules/api/templates/user_profile.html
```diff
@@ -91,6 +91,33 @@
...
+                    {%- trans %}Token-based links{% endtrans -%}
+                </div>
+                <div class="i-box-buttons toolbar">
+                    <div class="group">
+                        <button type="button"
+                                class="i-button warning icon-remove"
+                                data-href="{{ url_for('core.reset_signature_tokens') }}"
+                                data-method="POST"
+                                data-confirm="{% trans %}Are you sure you'd like to invalidate all your token-based links?{% endtrans %}">
+                            {%- trans %}Invalidate all Links{% endtrans -%}
```
ThiefMaster: s/Links/links/?

indico/modules/core/controllers.py
```diff
@@ -269,3 +269,18 @@ def _serialize_principal(self, identifier, principal):
     def _process(self, values):
         return jsonify({identifier: self._serialize_principal(identifier, principal)
                         for identifier, principal in values.viewitems()})
+
+
+class RHSignURL(RHProtected):
```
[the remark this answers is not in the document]
pferreir: Good idea.

indico/modules/users/blueprint.py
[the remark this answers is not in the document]
pferreir: I think that adding the token to the URL and removing the user makes the API less coherent. And if we ever implement the `current_user` proxy you talk about, then it shouldn't matter which auth method they're using, that should be the URL.

indico/modules/core/controllers.py
```diff
@@ -269,3 +270,25 @@ def _serialize_principal(self, identifier, principal):
...
+        endpoint = request.json.get('endpoint')
+        if not endpoint:
+            raise BadRequest
+        # filter out non-standard args
+        url_params = request.json.get('url_params', {})
+        url_params = {k: v for k, v in url_params.viewitems() if not k.startswith('_')}
+        query_params = request.json.get('query_params', {})
+        query_params = {k: v for k, v in query_params.viewitems() if not k.startswith('_')}
+        url = signed_url_for(session.user, endpoint, url_params=url_params, _external=True, **query_params)
+        Logger.get('url_signing').info("Created signed URL for endpoint '%s' (%s)", endpoint, url)
```
ThiefMaster: maybe including the signature in the logged url isn't such a great idea.. logging the user who created it is probably more useful

pferreir: I thought of that too, but is it much different from what we get in webserver logs?

ThiefMaster: good point ;)

indico/modules/users/controllers.py
```diff
@@ -102,35 +121,62 @@ class RHUserDashboard(RHUserBase):
...
-                                      linked_events=linked_events)
+        return WPUserDashboard.render_template('dashboard.html', 'dashboard',
+                                               user=self.user,
+                                               categories=categories,
+                                               categories_events=categories_events,
+                                               suggested_categories=get_suggested_categories(self.user),
+                                               linked_events=linked_events)
+
+
+class RHExportDashboardICS(RHProtected):
```
ThiefMaster: should this inherit from `RHProtected`? you aren't using anything from it since you are fully replacing its `_check_access` method, so i think inheriting from `RH` would make more sense

ThiefMaster:
image

Truncating the title at 50% width is a bit strange now

PR #4090, opened 2019-10-02
Cloning ACL involving event roles links to role from source event.

ThiefMaster:
When cloning an event that has event roles in the ACL, the new event will still reference the event roles from the old event. We need to clone any event roles that are used in an ACL instead (or just clone all of them when using the `EventProtectionCloner` since they are kind of related to "ACLs and protection settings" anyway)

https://talk.getindico.io/t/participant-role-clone-problem/1452

---

Added an explicit "Event roles" cloner option, and depending on whether it's selected or not, role entries in the ACLs will either be skipped or correctly cloned (with a reference to the cloned event role).

Also added a new CLI command `indico maint fix-event-role-acls` which fixes any ACL entries that reference a role from another event.

### DIFF
--- a/indico/modules/events/roles/clone.py
+++ b/indico/modules/events/roles/clone.py
@@ -1,0 +1,45 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2019 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from __future__ import unicode_literals
+
+from indico.core.db import db
+from indico.core.db.sqlalchemy.util.models import get_simple_column_attrs
+from indico.core.db.sqlalchemy.util.session import no_autoflush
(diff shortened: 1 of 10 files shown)


indico/modules/events/roles/clone.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @no_autoflush
+    def _clone_event_roles(self, new_event):
+        attrs = get_simple_column_attrs(EventRole) | {'members'}
+        for old_event_role in self.old_event.roles:
+            event_role = EventRole()
+            event_role.populate_from_attrs(old_event_role, attrs)
+            new_event.roles.append(event_role)
+            self._event_role_map[old_event_role] = event_role
+        print self._event_role_map
```
mic4ael: Did you leave it on purpose?

ThiefMaster: of course! :p

PR #4095, opened 2019-10-02
Add draft mode for contributions/timetable

### LINKED ISSUE #4056: Add "published" flag to List of Contributions
This is a switch that would be visible on the "Contributions" page in management mode. By default it would be in "draft" state. Once all abstracts have been processed and the resulting contributions are ready for display, one would just need to click it. We'd then be in "published" mode, which would activate:
 * the *Contribution List* in display view;
 * the *Book of Abstracts* (also in display view);

The process should be reversible, of course.

panagiotappl:
closes #4056 

TODO:

- [x] Write proper modal text

### DIFF
--- a/indico/modules/events/abstracts/__init__.py
+++ b/indico/modules/events/abstracts/__init__.py
@@ -123,14 +123,13 @@ def _extend_event_menu(sender, **kwargs):
     from indico.modules.events.contributions import contribution_settings
 
     def _boa_visible(event):
-        return config.LATEX_ENABLED and event.has_feature('abstracts') and contribution_settings.get(event, 'published')
-
+        return (config.LATEX_ENABLED and event.has_feature('abstracts')
+                and contribution_settings.get(event, 'published'))
 
     def _reviewing_area_visible(event):
         if not session.user or not event.has_feature('abstracts'):
             return False
         return has_user_tracks(event, session.user)
-
 
     yield MenuEntryData(title=_("Book of Abstracts"), name='abstracts_book', endpoint='abstracts.export_boa',
                         position=9, visible=_boa_visible, static_site=True)
--- a/indico/modules/events/abstracts/__init__.py
+++ b/indico/modules/events/abstracts/__init__.py
@@ -120,14 +120,17 @@ def _get_notification_placeholders(sender, **kwargs):
 def _extend_event_menu(sender, **kwargs):
     from indico.modules.events.abstracts.util import has_user_tracks
     from indico.modules.events.layout.util import MenuEntryData
+    from indico.modules.events.contributions import contribution_settings
 
     def _boa_visible(event):
-        return config.LATEX_ENABLED and event.has_feature('abstracts')
+        return config.LATEX_ENABLED and event.has_feature('abstracts') and contribution_settings.get(event, 'published')
+
 
     def _reviewing_area_visible(event):
         if not session.user or not event.has_feature('abstracts'):
             return False
         return has_user_tracks(event, session.user)
+
 
     yield MenuEntryData(title=_("Book of Abstracts"), name='abstracts_book', endpoint='abstracts.export_boa',
                         position=9, visible=_boa_visible, static_site=True)
--- a/indico/modules/events/abstracts/controllers/boa.py
+++ b/indico/modules/events/abstracts/controllers/boa.py
@@ -8,7 +8,7 @@
 from __future__ import unicode_literals
 
 from flask import flash
-from werkzeug.exceptions import Forbidden, NotFound
+from werkzeug.exceptions import NotFound
 
 from indico.core.config import config
 from indico.modules.events.abstracts.controllers.base import RHAbstractsBase, RHManageAbstractsBase
@@ -42,7 +42,7 @@ class RHExportBOA(RHAbstractsBase):
         RHAbstractsBase._check_access(self)
         published = contribution_settings.get(self.event, 'published')
         if not published:
-            raise Forbidden
+            raise NotFound(_("The contributions of this event haven't been published yet"))
 
     def _process(self):
         if not config.LATEX_ENABLED:
--- a/indico/modules/events/abstracts/controllers/boa.py
+++ b/indico/modules/events/abstracts/controllers/boa.py
@@ -8,13 +8,14 @@
 from __future__ import unicode_literals
 
 from flask import flash
-from werkzeug.exceptions import NotFound
+from werkzeug.exceptions import Forbidden, NotFound
 
 from indico.core.config import config
 from indico.modules.events.abstracts.controllers.base import RHAbstractsBase, RHManageAbstractsBase
 from indico.modules.events.abstracts.forms import BOASettingsForm
 from indico.modules.events.abstracts.settings import boa_settings
 from indico.modules.events.abstracts.util import clear_boa_cache, create_boa, create_boa_tex
+from indico.modules.events.contributions import contribution_settings
 from indico.util.i18n import _
 from indico.web.flask.util import send_file
 from indico.web.forms.base import FormDefaults
@@ -36,6 +37,12 @@ class RHManageBOA(RHManageAbstractsBase):
 
 class RHExportBOA(RHAbstractsBase):
     """Export the book of abstracts"""
+
+    def _check_access(self):
+        RHAbstractsBase._check_access(self)
+        published = contribution_settings.get(self.event, 'published')
+        if not published:
+            raise Forbidden
 
     def _process(self):
         if not config.LATEX_ENABLED:
--- a/indico/modules/events/abstracts/controllers/boa.py
+++ b/indico/modules/events/abstracts/controllers/boa.py
@@ -42,7 +42,7 @@ class RHExportBOA(RHAbstractsBase):
         RHAbstractsBase._check_access(self)
         published = contribution_settings.get(self.event, 'published')
         if not published:
-            raise NotFound(_("The contributions of this event haven't been published yet"))
+            raise NotFound(_("The contributions of this event have not been published yet"))
 
     def _process(self):
         if not config.LATEX_ENABLED:
--- a/indico/modules/events/contributions/__init__.py
+++ b/indico/modules/events/contributions/__init__.py
@@ -18,6 +18,7 @@ from indico.core.settings.converters import TimedeltaConverter
 from indico.modules.events.contributions.contrib_fields import get_contrib_field_types
 from indico.modules.events.contributions.models.contributions import Contribution
 from indico.modules.events.contributions.models.fields import ContributionField
+from indico.modules.events.models.events import EventType
 from indico.modules.events.settings import EventSettingsProxy
 from indico.util.i18n import _, ngettext
 from indico.web.flask.util import url_for
@@ -119,6 +120,12 @@ def _extend_event_menu(sender, **kwargs):
                         is_enabled=False, static_site=True)
     yield MenuEntryData(title=_("Speaker List"), name='speaker_index', endpoint='contributions.speaker_list',
                         position=6, is_enabled=False, static_site=True)
+
+
+@signals.event.created.connect
+def _event_created(event, **kwargs):
+    if event.type_ == EventType.conference:
+        contribution_settings.set(event, 'published', False)
 
 
 contribution_settings = EventSettingsProxy('contributions', {
--- a/indico/modules/events/contributions/__init__.py
+++ b/indico/modules/events/contributions/__init__.py
@@ -120,7 +120,8 @@ def _extend_event_menu(sender, **kwargs):
 
 
 contribution_settings = EventSettingsProxy('contributions', {
-    'default_duration': timedelta(minutes=20)
+    'default_duration': timedelta(minutes=20),
+    'published': True
 }, converters={
     'default_duration': TimedeltaConverter
 })
--- a/indico/modules/events/contributions/__init__.py
+++ b/indico/modules/events/contributions/__init__.py
@@ -104,10 +104,12 @@ def _extend_event_menu(sender, **kwargs):
     from indico.modules.events.layout.util import MenuEntryData
 
     def _visible_my_contributions(event):
-        return session.user and has_contributions_with_user_as_submitter(event, session.user)
+        published = contribution_settings.get(event, 'published')
+        return session.user and has_contributions_with_user_as_submitter(event, session.user) and published
 
     def _visible_list_of_contributions(event):
-        return Contribution.query.filter(Contribution.event == event).has_rows()
+        published = contribution_settings.get(event, 'published')
+        return Contribution.query.filter(Contribution.event == event).has_rows() and published
 
     yield MenuEntryData(title=_("My Contributions"), name='my_contributions', visible=_visible_my_contributions,
                         endpoint='contributions.my_contributions', position=2, parent='my_conference')
--- a/indico/modules/events/contributions/__init__.py
+++ b/indico/modules/events/contributions/__init__.py
@@ -105,11 +105,11 @@ def _extend_event_menu(sender, **kwargs):
 
     def _visible_my_contributions(event):
         published = contribution_settings.get(event, 'published')
-        return session.user and has_contributions_with_user_as_submitter(event, session.user) and published
+        return session.user and published and has_contributions_with_user_as_submitter(event, session.user)
 
     def _visible_list_of_contributions(event):
         published = contribution_settings.get(event, 'published')
-        return Contribution.query.filter(Contribution.event == event).has_rows() and published
+        return published and Contribution.query.filter(Contribution.event == event).has_rows()
 
     yield MenuEntryData(title=_("My Contributions"), name='my_contributions', visible=_visible_my_contributions,
                         endpoint='contributions.my_contributions', position=2, parent='my_conference')
--- a/indico/modules/events/contributions/client/js/index.js
+++ b/indico/modules/events/contributions/client/js/index.js
@@ -19,7 +19,7 @@ import PublicationSwitch from 'indico/react/components/PublicationSwitch';
 
   function setupPublicationButton() {
     const eventId = $('#pub-switch').data('event-id');
-    const element = React.createElement(PublicationSwitch, {confId: eventId})
+    const element = React.createElement(PublicationSwitch, {eventId})
     ReactDOM.render(element, document.getElementById('pub-switch'));
   }
 
--- a/indico/modules/events/contributions/client/js/index.js
+++ b/indico/modules/events/contributions/client/js/index.js
@@ -12,20 +12,19 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 
 import 'indico/modules/events/util/types_dialog';
-import PublicationSwitch from 'indico/react/components/PublicationSwitch';
+import {PublicationSwitch} from 'indico/react/components';
 
 
 (function(global) {
-
   function setupPublicationButton() {
-    const eventId = $('#pub-switch').data('event-id');
-    const element = React.createElement(PublicationSwitch, {eventId})
-    ReactDOM.render(element, document.getElementById('pub-switch'));
+    const element = document.querySelector('#pub-switch');
+    const component = React.createElement(PublicationSwitch, {eventId: element.dataset.eventId});
+    ReactDOM.render(component, element);
   }
 
   global.setupContributionConfig = function setupContributionConfig() {
     setupPublicationButton();
-  }
+  };
 
   function setupTableSorter(selector) {
     $(selector).tablesorter({
--- a/indico/modules/events/contributions/client/js/index.js
+++ b/indico/modules/events/contributions/client/js/index.js
@@ -8,9 +8,25 @@
 /* global showUndoWarning:false, setupListGenerator:false, setupSearchBox:false */
 /* global reloadManagementAttachmentInfoColumn:false */
 
+import React from 'react';
+import ReactDOM from 'react-dom';
+
 import 'indico/modules/events/util/types_dialog';
+import PublicationSwitch from 'indico/react/components/PublicationSwitch';
+
 
 (function(global) {
+
+  function setupPublicationButton() {
+    const eventId = $('#pub-switch').data('event-id');
+    const element = React.createElement(PublicationSwitch, {confId: eventId})
+    ReactDOM.render(element, document.getElementById('pub-switch'));
+  }
+
+  global.setupContributionConfig = function setupContributionConfig() {
+    setupPublicationButton();
+  }
+
   function setupTableSorter(selector) {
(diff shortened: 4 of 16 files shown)


indico/modules/events/contributions/controllers/management.py
```diff
@@ -541,6 +541,22 @@ def _process(self):
...
+class RHManageContributionPublicationREST(RHManageContributionsBase):
+    """Manage contribution publication setting"""
+
+    def _process_GET(self):
+        published = contribution_settings.get(self.event, 'published')
+        return jsonify(published=published)
+
+    def _process_PUT(self):
+        contribution_settings.set(self.event, 'published', True)
+        return jsonify_data(flash=False)
```
ThiefMaster: `return '', 204` (ie an empty no-content response)

indico/modules/events/contributions/controllers/management.py
```diff
@@ -541,6 +541,22 @@ def _process(self):
...
+
+    def _process_DELETE(self):
+        contribution_settings.set(self.event, 'published', False)
+        return jsonify_data(flash=False)
```
ThiefMaster: same here

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -43,6 +43,7 @@
             </li>
         </ul>
     {%- endif -%}
+    <div id="pub-switch" data-event-id={{event.id}} ></div>
```
ThiefMaster:
```suggestion
    <div id="pub-switch" data-event-id="{{ event.id }}"></div>
```

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,79 @@
...
+export default function PublicationSwitch({confId}) {
+  const [published, setPublished] = useState(true);
+  const [modalOpen, setModalOpen] = useState(false);
+
+  useEffect(() => {
+    const fetchPublicationSetting = async () => {
+      let response;
+      try {
+        response = await indicoAxios.get(publicationURL({confId}));
+        setPublished(response.data.published);
```
ThiefMaster: move this after the try..catch (and return at the end of the catch)

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,79 @@
...
+      } catch (error) {
+        handleAxiosError(error);
+      }
+    };
+    fetchPublicationSetting();
+  }, [confId]);
+
+  const publish = shouldPublish => {
```
ThiefMaster: maybe `state` or `published` would be a better name for this argument?

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,79 @@
...
+    const updatePublicationSetting = async () => {
+      try {
+        if (shouldPublish) {
+          await indicoAxios.put(publicationURL({confId}));
+        } else {
+          await indicoAxios.delete(publicationURL({confId}));
+        }
+        setPublished(shouldPublish);
```
ThiefMaster: outside the try..catch block

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,79 @@
...
+  };
+
+  return (
+    <Modal
+      open={modalOpen}
+      onClose={() => setModalOpen(false)}
+      size="tiny"
+      trigger={
+        <Checkbox
+          label={published ? 'Published' : 'Draft'}
```
ThiefMaster: i18n

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,79 @@
...
+          toggle
+          onClick={() => setModalOpen(true)}
+          checked={published}
+        />
+      }
+      closeIcon
+    >
+      <Header content={!published ? 'Publish Contributions' : 'Unpublish Contributions'} />
```
ThiefMaster: i18n

and the negation is weird

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,79 @@
...
+        >
+          Yes
+        </Button>
+      </Modal.Actions>
+    </Modal>
+  );
+}
+
+PublicationSwitch.propTypes = {
+  confId: PropTypes.number.isRequired,
```
ThiefMaster: `eventId`; it's ugly enough that we still use `confId` in the url routes, let's not use it anywhere else :)

indico/modules/events/contributions/__init__.py
```diff
@@ -104,10 +104,12 @@ def _extend_event_menu(sender, **kwargs):
     from indico.modules.events.layout.util import MenuEntryData
 
     def _visible_my_contributions(event):
-        return session.user and has_contributions_with_user_as_submitter(event, session.user)
+        published = contribution_settings.get(event, 'published')
+        return session.user and has_contributions_with_user_as_submitter(event, session.user) and published
```
mic4ael: Maybe you could put `and published` before `and has_contributions`. This way we avoid querying the DB in case `published` is `False`.

indico/modules/events/contributions/__init__.py
```diff
@@ -104,10 +104,12 @@ def _extend_event_menu(sender, **kwargs):
...
 
     def _visible_list_of_contributions(event):
-        return Contribution.query.filter(Contribution.event == event).has_rows()
+        published = contribution_settings.get(event, 'published')
+        return Contribution.query.filter(Contribution.event == event).has_rows() and published
```
mic4ael: The same as above

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,79 @@
...
+import PropTypes from 'prop-types';
+import {Header, Modal, Button, Checkbox} from 'semantic-ui-react';
+import indicoAxios, {handleAxiosError} from 'indico/utils/axios';
+
+export default function PublicationSwitch({confId}) {
+  const [published, setPublished] = useState(true);
+  const [modalOpen, setModalOpen] = useState(false);
+
+  useEffect(() => {
+    const fetchPublicationSetting = async () => {
```
mic4ael: Is there any reason why this function has been defined inside the `useEffect` callback? Even though this effect will not be called more than once (since `confId` will not change) I think it would be nice to move the function outside of the callback and then you could also remove the `confId` from the `useEffect` dependencies.

ThiefMaster: You could also move it outside the component and pass the `eventId` as an argument.

But I think the `eventId` *should* remain in the effect's dependency list, because in the (not-gonna-happen) case of the prop changing you DO need the effect to re-run and fetch the state of the new event.

ThiefMaster: I think the `eventId` should stay in the dependency list. In the case of the prop changing to another ID (even if that's not gonna happen), you want to reload the data for the new ID.

indico/web/client/js/react/components/PublicationSwitch.jsx
panagiotappl: Maybe `changeState` as what it really does is change the state from published to unpublished and the other way around?

indico/modules/events/contributions/controllers/display.py
```diff
@@ -52,8 +53,11 @@ class RHContributionDisplayBase(RHDisplayEventBase):
 
     def _check_access(self):
         RHDisplayEventBase._check_access(self)
-        if not self.contrib.can_access(session.user):
+        published = contribution_settings.get(self.event, 'published')
+        if not self.contrib.can_access(session.user) or not published:
             raise Forbidden
+        if not published:
+            raise NotFound("The contributions of this event haven't been published yet.")
```
ThiefMaster: i18n

indico/modules/events/contributions/controllers/display.py
```diff
@@ -63,6 +67,10 @@ def _process_args(self):
 class RHDisplayProtectionBase(RHDisplayEventBase):
     def _check_access(self):
         RHDisplayEventBase._check_access(self)
+        published = contribution_settings.get(self.event, 'published')
+        if not published:
+            raise NotFound("The contributions of this event haven't been published yet.")
```
ThiefMaster: i18n

indico/modules/events/contributions/controllers/display.py
```diff
@@ -72,11 +80,6 @@ class RHMyContributions(RHDisplayProtectionBase):
 
     MENU_ENTRY_NAME = 'my_contributions'
 
-    def _check_access(self):
-        RHDisplayProtectionBase._check_access(self)
-        if not session.user:
-            raise Forbidden
```
ThiefMaster: Why are you removing this?

indico/modules/events/contributions/controllers/management.py
```diff
@@ -541,6 +541,22 @@ def _process(self):
...
 
+class RHManageContributionPublicationREST(RHManageContributionsBase):
+    """Manage contribution publication setting"""
+
+    def _process_GET(self):
+        published = contribution_settings.get(self.event, 'published')
+        return jsonify(published=published)
+
+    def _process_PUT(self):
+        contribution_settings.set(self.event, 'published', True)
```
ThiefMaster: This should be logged in the event log (`self.event.log(...)`)

indico/modules/events/contributions/controllers/management.py
```diff
@@ -541,6 +541,22 @@ def _process(self):
...
+        return '', 204
+
+    def _process_DELETE(self):
+        contribution_settings.set(self.event, 'published', False)
```
ThiefMaster: same

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -43,6 +43,7 @@
             </li>
         </ul>
     {%- endif -%}
+    <div id="pub-switch" data-event-id="{{ event.id }}" ></div>
```
ThiefMaster:
```suggestion
    <div id="pub-switch" data-event-id="{{ event.id }}"></div>
```

indico/modules/events/controllers/creation.py
```diff
@@ -115,6 +116,8 @@ def _process(self):
                 notify_event_creation(event, occurrences=events)
             else:
                 event = self._create_event(form.data)
+                if self.event_type == EventType.conference:
+                    contribution_settings.set(event, 'published', False)
```
ThiefMaster: I'd rather do this from a signal handler in `contributions/__init__.py`:

```python
@signals.event.created.connect
def _event_created(event, **kwargs):
    if event.type_ == EventType.conference:
        contribution_settings.set(event, 'published', False)
```

indico/modules/events/timetable/controllers/display.py
```diff
@@ -26,12 +27,21 @@
 from indico.web.util import jsonify_data, jsonify_template
 
 
-class RHTimetable(RHDisplayEventBase):
+class RHTimetableProtectionBase(RHDisplayEventBase):
+
```
ThiefMaster: :space_invader:

indico/modules/events/timetable/controllers/display.py
```diff
@@ -65,7 +75,8 @@ def _process(self):
         return jsonify(html=html)
 
 
-class RHTimetableExportPDF(RHDisplayEventBase):
+class RHTimetableExportPDF(RHTimetableProtectionBase):
+
```
ThiefMaster: :space_invader:

indico/modules/events/timetable/controllers/display.py
```diff
@@ -92,7 +103,8 @@ def _process(self):
                                 back_url=url_for('.timetable', self.event))
 
 
-class RHTimetableExportDefaultPDF(RHDisplayEventBase):
+class RHTimetableExportDefaultPDF(RHTimetableProtectionBase):
+
```
ThiefMaster: :space_invader:

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,137 @@
...
+  const changeState = shouldPublish => {
+    const updatePublicationSetting = async () => {
+      try {
+        if (shouldPublish) {
+          await indicoAxios.put(publicationURL({confId: eventId}));
+        } else {
+          await indicoAxios.delete(publicationURL({confId: eventId}));
+        }
+      } catch (error) {
+        handleAxiosError(error);
```
ThiefMaster: `return;` after handling the error

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,137 @@
...
+      }
+      setPublished(response.data.published);
+    };
+    fetchPublicationSetting();
+  }, [eventId]);
+
+  const changeState = shouldPublish => {
+    const updatePublicationSetting = async () => {
```
ThiefMaster: No need for this nesting. Just make `changeState` async. We only need this kind of wrapping in useEffect, because the effect's function should not be returning a promise (which all async functions do)

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,137 @@
...
+          published
+            ? Translate.string('Set contribution list in Draft mode')
+            : Translate.string('Publish Contributions')
+        }
+      />
+      <Modal.Content>
+        {published ? (
+          <>
+            <p>
+              <Translate>Are you sure you want to set contribution list in draft mode?</Translate>
```
ThiefMaster: > Are you sure you want to change the contribution list back to draft mode?

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,137 @@
...
+          </>
+        ) : (
+          <>
+            <p>
+              <Translate>Are you sure you want to publish the contribution list?</Translate>
+            </p>
+            <p>
+              <Translate>By doing so the following menu items will be accessible: </Translate>
+            </p>
+            {menuList}
```
ThiefMaster: why not move this after the ternary instead of having it in both places? then you don't even need the variable but can put the JSX there directly

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,137 @@
...
+          </>
+        )}
+      </Modal.Content>
+      <Modal.Actions>
+        <Button onClick={() => setModalOpen(false)}>
+          <Translate>No</Translate>
+        </Button>
+        <Button
+          color="blue"
```
ThiefMaster: `primary` instead of `color="blue"`?

indico/modules/events/contributions/client/js/index.js
```diff
@@ -8,9 +8,24 @@
...
 (function(global) {
+  function setupPublicationButton() {
+    const element = document.querySelector('#pub-switch');
+    const component = React.createElement(PublicationSwitch, {eventId: element.dataset.eventId});
+    ReactDOM.render(component, element);
+  }
+
+  global.setupContributionConfig = function setupContributionConfig() {
+    setupPublicationButton();
+  }
```
mic4ael: missing semicolon

indico/modules/events/contributions/controllers/display.py
```diff
@@ -52,8 +54,11 @@ class RHContributionDisplayBase(RHDisplayEventBase):
 
     def _check_access(self):
         RHDisplayEventBase._check_access(self)
-        if not self.contrib.can_access(session.user):
+        published = contribution_settings.get(self.event, 'published')
+        if not self.contrib.can_access(session.user) or not published:
```
mic4ael: I would put `not published` before `not self.contrib.can_access`

indico/modules/events/controllers/creation.py
```diff
@@ -18,6 +18,7 @@
 from indico.core.config import config
 from indico.core.db.sqlalchemy.util.models import get_simple_column_attrs
 from indico.modules.categories import Category
+from indico.modules.events.contributions import contribution_settings
```
mic4ael: Something you forgot to remove?

indico/modules/events/abstracts/__init__.py
```diff
@@ -120,15 +120,18 @@ def _get_notification_placeholders(sender, **kwargs):
 def _extend_event_menu(sender, **kwargs):
     from indico.modules.events.abstracts.util import has_user_tracks
     from indico.modules.events.layout.util import MenuEntryData
+    from indico.modules.events.contributions import contribution_settings
 
     def _boa_visible(event):
-        return config.LATEX_ENABLED and event.has_feature('abstracts')
+        return config.LATEX_ENABLED and event.has_feature('abstracts') and contribution_settings.get(event, 'published')
```
ThiefMaster: I think this line is just one character too long...

indico/modules/events/abstracts/__init__.py
```diff
@@ -120,15 +120,18 @@ def _get_notification_placeholders(sender, **kwargs):
...
+
 
     def _reviewing_area_visible(event):
         if not session.user or not event.has_feature('abstracts'):
             return False
         return has_user_tracks(event, session.user)
 
+
```
ThiefMaster: I think functions that are not on the top level usually don't have two empty lines in between

indico/modules/events/abstracts/controllers/boa.py
```diff
@@ -37,6 +38,12 @@ def _process(self):
 class RHExportBOA(RHAbstractsBase):
     """Export the book of abstracts"""
 
+    def _check_access(self):
+        RHAbstractsBase._check_access(self)
+        published = contribution_settings.get(self.event, 'published')
+        if not published:
+            raise NotFound(_("The contributions of this event haven't been published yet"))
```
ThiefMaster: "have not"? "haven't" is pretty colloquial

indico/modules/events/contributions/controllers/display.py
```diff
@@ -53,8 +55,11 @@ class RHContributionDisplayBase(RHDisplayEventBase):
 
     def _check_access(self):
         RHDisplayEventBase._check_access(self)
-        if not self.contrib.can_access(session.user):
+        published = contribution_settings.get(self.event, 'published')
+        if not published or not self.contrib.can_access(session.user):
             raise Forbidden
+        if not published:
```
ThiefMaster: this code never runs since you raise Forbidden above...

indico/modules/events/contributions/controllers/management.py
```diff
@@ -544,6 +544,26 @@ def _process(self):
...
+class RHManageContributionPublicationREST(RHManageContributionsBase):
+    """Manage contribution publication setting"""
+
+    def _process_GET(self):
+        published = contribution_settings.get(self.event, 'published')
+        return jsonify(published=published)
+
+    def _process_PUT(self):
+        contribution_settings.set(self.event, 'published', True)
+        self.event.log(EventLogRealm.management, EventLogKind.change, 'Contributions',
```
ThiefMaster: I think we could use `EventLogKind.positive` here (and `negative` for unpublishing below)

indico/modules/events/timetable/controllers/display.py
```diff
@@ -26,12 +27,20 @@
 from indico.web.util import jsonify_data, jsonify_template
 
 
-class RHTimetable(RHDisplayEventBase):
+class RHTimetableProtectionBase(RHDisplayEventBase):
+    def _check_access(self):
+        RHDisplayEventBase._check_access(self)
+        published = contribution_settings.get(self.event, 'published')
+        if not published:
+            raise Forbidden
```
ThiefMaster: shouldn't this use a proper message as well? also, you're using NotFound in some places and Forbidden in others..

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,141 @@
...
+          onClick={() => setModalOpen(true)}
+          checked={published}
+        />
+      }
+      closeIcon
+    >
+      <Header
+        content={
+          published
+            ? Translate.string('Set contribution list in Draft mode')
```
ThiefMaster: "Draft" or "draft"?

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,141 @@
...
+            : Translate.string('Publish Contributions')
```
ThiefMaster: lowercase "contributions"

indico/web/client/js/react/components/PublicationSwitch.jsx
```diff
@@ -0,0 +1,141 @@
...
+              <Translate>By doing so the following menu items won't be accessible:</Translate>
+            </p>
+          </>
+        ) : (
+          <>
+            <p>
+              <Translate>Are you sure you want to publish the contribution list?</Translate>
+            </p>
+            <p>
+              <Translate>By doing so the following menu items will be accessible: </Translate>
```
ThiefMaster: remove the trailing space

indico/web/client/styles/modules/_contributions.scss
```diff
@@ -21,6 +21,12 @@
     }
 }
 
+#pub-switch {
```
ThiefMaster: shouldn't this be a class?

indico/modules/events/contributions/controllers/display.py
panagiotappl: I should just remove the `not published` check from there. Probably a mistake.

indico/web/client/styles/modules/_contributions.scss
panagiotappl: I use it for the querySelector ` const element = document.querySelector('#pub-switch');` Should I make a separate class for the styles?

ThiefMaster: ah you query a single element; then the ID makes sense

indico/modules/events/contributions/controllers/display.py
```diff
@@ -53,8 +55,11 @@ class RHContributionDisplayBase(RHDisplayEventBase):
 
     def _check_access(self):
         RHDisplayEventBase._check_access(self)
+        published = contribution_settings.get(self.event, 'published')
         if not self.contrib.can_access(session.user):
             raise Forbidden
+        if not published:
+            raise NotFound(_("The contributions of this event haven't been published yet."))
```
ThiefMaster: let's be consistent with "haven't" vs "have not" (right now we have both in this PR)

PR #4097, opened 2019-10-03
Add include_groups option in get_manager_list(...)

pferreir:
This is needed for https://github.com/indico/indico-plugins-cern/pull/67

### DIFF
--- a/indico/core/db/sqlalchemy/protection.py
+++ b/indico/core/db/sqlalchemy/protection.py
@@ -462,10 +462,14 @@ class ProtectionManagersMixin(ProtectionMixin):
                                        old_data=old_data, quiet=quiet)
         return entry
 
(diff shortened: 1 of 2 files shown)


indico/core/db/sqlalchemy/protection.py
```diff
@@ -462,10 +462,14 @@ def update_principal(self, principal, read_access=None, full_access=None, permis
                                        old_data=old_data, quiet=quiet)
         return entry
 
-    def get_manager_list(self, recursive=False):
-        managers = {x.principal for x in self.acl_entries if x.has_management_permission()}
+    def get_manager_list(self, recursive=False, include_groups=True):
```
ThiefMaster: For the sake of consistency I would update the signature of `SubContribution.get_manager_list` as well.

PR #4100, opened 2019-10-07
Abstract permissions

### LINKED ISSUE #4041: Allow event roles to be assigned to abstract permissions
image

Right now it's only possible to add users. No groups and no roles.

meluru:
closes #4041


indico/migrations/versions/20191002_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,231 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '2496c4adc7e9'
+down_revision = 'eefba82b42c5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
ThiefMaster: those comments should be removed

indico/migrations/versions/20191002_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,231 @@
...
+            [u'users.users.id'],
+            name=op.f('fk_track_principals_user_id_users'),
+        ),
+        sa.PrimaryKeyConstraint('id', name=op.f('pk_track_principals')),
+        schema='events',
+    )
+    op.create_index(
+        op.f('ix_track_principals_email'),
+        'track_principals',
+        ['email'],
```
ThiefMaster: all those simple single-column indexes can be written as `index=True` on the corresponding column definition above

indico/migrations/versions/20191002_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,231 @@
...
+        unique=True,
+        schema='events',
+        postgresql_where=sa.text(u'type = 1'),
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index(
```
ThiefMaster: none of those commands are needed since dropping a table will drop everything associated with it

indico/modules/events/abstracts/__init__.py
```diff
@@ -96,8 +97,27 @@ def is_allowed_for_event(cls, event):
 
 
 @signals.acl.get_management_permissions.connect_via(Event)
+@signals.acl.get_management_permissions.connect_via(Track)
 def _get_management_permissions(sender, **kwargs):
-    return AbstractReviewerPermission
+    if isinstance(sender, Track):
```
ThiefMaster: please use two functions for this instead of checking the sender's type

indico/modules/events/abstracts/controllers/management.py
```diff
@@ -134,12 +134,16 @@ def _process(self):
 
         if form.validate_on_submit():
             role_data = form.roles.role_data
-            self.event.global_conveners = set(role_data['global_conveners'])
-            self.event.global_abstract_reviewers = set(role_data['global_reviewers'])
+            for principal in role_data['global_conveners']:
+                self.event.update_principal(principal, permissions={'convene_all'})
```
ThiefMaster: I'd go for `convene_all_abstracts`.. that's clearer (not sure why we didn't do that previously)

ThiefMaster: Also, if you rename the permissions you need SQL in the migration step to update the permissions array (hopefully postgres provides functions to replace data in an array) in each acl entry

indico/modules/events/abstracts/util.py
```diff
@@ -197,15 +199,25 @@ def get_roles_for_event(event):
...
         roles[str(track.id)].setdefault('reviewer', [])
         roles[str(track.id)].setdefault('convener', [])
-        for reviewer in track.abstract_reviewers:
-            roles[str(track.id)]['reviewer'].append(reviewer.id)
-        for convener in track.conveners:
-            roles[str(track.id)]['convener'].append(convener.id)
-    roles['*']['reviewer'] = [reviewer.id for reviewer in event.global_abstract_reviewers]
-    roles['*']['convener'] = [reviewer.id for reviewer in event.global_conveners]
+        for principal in track.acl_entries:
+            if not principal.user_id:
```
ThiefMaster: check `principal.type` instead. user 0 may actually be a real user for anyone who migrated from 1.2

indico/modules/events/abstracts/util.py
```diff
@@ -222,17 +234,30 @@ def get_user_abstracts(event, user):
...
+    for role in user.event_roles:
+        criteria.append(db.and_(TrackPrincipal.type == PrincipalType.event_role,
+                                TrackPrincipal.event_role_id == role.id,
+                                TrackPrincipal.has_management_permission(permission='ANY')))
     query = Track.query.with_parent(event)
-    if user not in event.global_abstract_reviewers and user not in event.global_conveners:
-        query = query.filter(db.or_(Track.conveners.any(User.id == user.id),
-                                    Track.abstract_reviewers.any(User.id == user.id)))
+    if (not event.can_manage(user, permission='review_all', explicit_permission=True) and not
+       event.can_manage(user, permission='convene_all', explicit_permission=True)):
```
ThiefMaster: continued if conditions have 4 more spaces of indentation in python

indico/modules/events/abstracts/util.py
```diff
@@ -222,17 +234,30 @@ def get_user_abstracts(event, user):
...
+    convener_tracks = set(track for track in event.tracks
+                          if track.can_manage(user, permission='convene', explicit_permission=True))
     return abstract.reviewed_for_tracks & convener_tracks
 
 
 def _query_user_tracks(event, user):
+    criteria = [db.and_(TrackPrincipal.type == PrincipalType.user,
+                        TrackPrincipal.user_id == user.id,
+                        TrackPrincipal.has_management_permission(permission='ANY'))]
+    for group in user.local_groups:
```
ThiefMaster: I think we need to support multipass groups as well here (which isn't great performance-wise), OR we explicitly disallow anything but users (and event roles) for the abstract-related ACLs

indico/modules/events/tracks/__init__.py
```diff
@@ -44,20 +44,37 @@ def _program_visible(event):
...
 @signals.event_management.get_cloners.connect
 def _get_cloners(sender, **kwargs):
     yield TrackCloner
 
 
+@signals.acl.get_management_permissions.connect_via(Track)
 @signals.acl.get_management_permissions.connect_via(Event)
 def _get_management_permissions(sender, **kwargs):
-    return TrackConvenerPermission
+    if isinstance(sender, Track):
```
ThiefMaster: two functions again

indico/core/permissions.py
```diff
@@ -144,7 +147,8 @@ def get_unified_permissions(principal):
     """Convert principal's permissions into a list including read and full access."""
     if principal.full_access:
         return {FULL_ACCESS_PERMISSION}
-    return (set(principal.permissions) + {READ_ACCESS_PERMISSION}) if principal.read_access else principal.permissions
+    return ((set(principal.permissions) + {READ_ACCESS_PERMISSION}) if principal.read_access
```
mic4ael: why not sth like this

```python

perms = set(principal.permissions)
if principal.read_access:
    perms.add(READ_ACCESS_PERMISSION)
return perms
```

it would be a bit more readable IMO

indico/migrations/versions/20191023_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,304 @@
...
+    ''', (track_id, user_id, PrincipalType.user.value, permissions))
+
+
+def _upgrade_permissions():
+    conn = op.get_bind()
+
+    # Create track acl entries
+
+    track_reviewers_conveners = '''
+        SELECT user_id, track_id FROM events.track_abstract_reviewers as reviewers WHERE EXISTS
```
mic4ael: TBH I find this SQL query a bit unreadable. How about sth like this

```sql

SELECT user_id, track_id
FROM   events.track_abstract_reviewers AS reviewers
WHERE EXISTS (
  SELECT 1
  FROM   events.track_conveners AS conveners
  WHERE  ( reviewers.user_id = conveners.user_id
           AND reviewers.track_id = conveners.track_id )
) 
```

indico/migrations/versions/20191023_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,304 @@
...
+        (SELECT 1 FROM events.track_conveners as conveners WHERE
+        (reviewers.user_id = conveners.user_id AND
+        reviewers.track_id = conveners.track_id))
+    '''
+
+    for user_id, track_id in conn.execute(track_reviewers_conveners):
+        _create_track_acl_entry(conn, user_id, track_id, ['convene', 'review'])
+
+    track_reviewers = '''
+        SELECT user_id, track_id FROM events.track_abstract_reviewers as reviewers WHERE
```
mic4ael: Similar to my comment above

indico/migrations/versions/20191023_1820_2496c4adc7e9_add_track_principal.py
mic4ael: I don't want to nit-pick but if we uppercase SQL keywords then we should also use `AS` instead of `as`

indico/migrations/versions/20191023_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,304 @@
...
+        ),
+        sa.CheckConstraint(
+            u'NOT read_access', name=op.f('ck_track_principals_no_read_access')
+        ),
+        sa.CheckConstraint(
+            u'email IS NULL OR email = lower(email)',
+            name=op.f('ck_track_principals_lowercase_email'),
+        ),
+        sa.CheckConstraint(
+            u'read_access OR full_access OR array_length(permissions, 1) IS NOT NULL',
```
mic4ael: will `array_length` ever return `NULL`? I think it should always return an `int`

indico/migrations/versions/20191023_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,304 @@
...
+            name=op.f('ck_track_principals_valid_email'),
+        ),
+        sa.CheckConstraint(
+            u'type != 6 OR (email IS NULL AND local_group_id IS NULL AND mp_group_name IS NULL AND '
+            'mp_group_provider IS NULL AND user_id IS NULL AND event_role_id IS NOT NULL)',
+            name=op.f('ck_track_principals_valid_event_role'),
+        ),
+        sa.ForeignKeyConstraint(
+            ['event_role_id'],
+            [u'events.roles.id'],
```
mic4ael: why `u` here? does it have to be unicode? I don't think we use `unicode` strings in other places

indico/migrations/versions/20191023_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,304 @@
...
+            name=op.f('fk_track_principals_event_role_id_roles'),
+        ),
+        sa.ForeignKeyConstraint(
+            ['local_group_id'],
+            [u'users.groups.id'],
```
mic4ael: the same as above

indico/modules/events/abstracts/templates/forms/track_role_widget.html
```diff
@@ -14,41 +12,7 @@
...
-                            <li class="event-user" data-user-id="{{ user_id }}">
-                                {{- user_data[user_id].name -}}
-                                <a class="icon-cross remove-user hide-if-locked" href="#"></a>
-                            </li>
-                        {% endfor %}
-                    </ul>
-                </div>
-            {% endfor %}
-        </div>
+        <div id="track-roles-{{'global' if global else track_id}}" data-track-id="{{ track_id }}"></div>
```
mic4ael:
```suggestion
        <div id="track-roles-{{ 'global' if global else track_id }}" data-track-id="{{ track_id }}"></div>
```

indico/modules/events/schemas.py
```diff
@@ -0,0 +1,19 @@
...
+# LICENSE file for more details.
+
+from __future__ import unicode_literals
+
+from indico.core.marshmallow import mm
+from indico.modules.events.models.principals import EventPrincipal
+from indico.util.marshmallow import PrincipalPermissionList
+
+
+class EventPermissionsSchema(mm.ModelSchema):
```
mic4ael: does it have to inherit from `mm.ModelSchema` if it doesn't specify the actualy `Meta` class with `model` attribute? Can't you use `mm.Schema`?

indico/modules/events/tracks/models/principals.py
```diff
@@ -0,0 +1,57 @@
...
+        return auto_table_args(cls, schema='events')
+
+    @strict_classproperty
+    @staticmethod
+    def __auto_table_args():
+        constraints = [
+            db.CheckConstraint('NOT read_access', 'no_read_access'),
+            db.CheckConstraint('NOT full_access', 'no_full_access'),
+        ]
+        return tuple(constraints)
```
mic4ael:
```python

return (
    db.CheckConstraint...,
    db.CheckConstraint
)

File: indico/modules/events/tracks/schemas.py [hunk 19]
```diff
@@ -17,5 +19,10 @@ class Meta:
         fields = ('id', 'title', 'code', 'description')
 
 
+class TrackPermissionsSchema(mm.ModelSchema):
```
mic4ael: do you have to inherit from `mm.ModelSchema` here as well? (similar to my comment above)

indico/migrations/versions/20191023_1820_2496c4adc7e9_add_track_principal.py
mic4ael: Ahh, I see now.. docs are a bit misleading :P (it returns `NULL` for an empty array and docs specify that it always returns an `int`)

ThiefMaster:
Many of the commits could probably be squashed together. I think the best approach to commits is to have them as logical units, so for example "Allow disabling protection mode in ProtectionMixin" is a change that does not affect existing code/functionality and not part of the "abstract permissions" feature itself, so it makes perfect sense as a separate commit.

But many/most other commits are really just steps from implementing the feature, so having those as separate commits isn't *that* helpful...

(opinions on this are welcome @indico/development-team)

mic4ael:
> Many of the commits could probably be squashed together. I think the best approach to commits is to have them as logical units, so for example "Allow disabling protection mode in ProtectionMixin" is a change that does not affect existing code/functionality and not part of the "abstract permissions" feature itself, so it makes perfect sense as a separate commit.
> 
> But many/most other commits are really just steps from implementing the feature, so having those as separate commits isn't _that_ helpful...
> 
> (opinions on this are welcome @indico/development-team)

I fully agree. Commits should represent a given functionality so that if I want to checkout a random commit on a given branch I don't want to end up in the middle of the development process (leaving the whole thing in an unusable state).

indico/modules/events/abstracts/util.py
```diff
@@ -238,27 +219,27 @@ def get_user_abstracts(event, user):
...
-        query = query.filter(db.or_(Track.conveners.any(User.id == user.id),
-                                    Track.abstract_reviewers.any(User.id == user.id)))
-    return query
-
-
 def get_user_tracks(event, user):
     """Get the list of tracks where the user is a reviewer/convener"""
-    return _query_user_tracks(event, user).order_by(Track.title).all()
+    tracks = Track.query.with_parent(event).order_by(Track.title).all()
+    if (not event.can_manage(user, permission='review_all_abstracts', explicit_permission=True) and not
```
ThiefMaster: I think the last `not` should go in the next line

indico/modules/events/abstracts/util.py
```diff
@@ -345,24 +326,31 @@ def get_events_with_abstract_reviewer_convener(user, dt=None):
...
-        query = (Track.query.with_parent(user, rel)
-                 .join(Track.event)
-                 .filter(Event.ends_after(dt), ~Event.is_deleted)
-                 .options(load_only('event_id')))
-        for track in query:
-            data[track.event_id].add(role)
+
+    event_query = (user.in_event_acls
+                   .join(Event)
+                   .options(noload('user'), noload('local_group'), load_only('event_id', 'permissions'))
```
ThiefMaster: `noload('*')`?

indico/modules/events/abstracts/util.py
```diff
@@ -345,24 +326,31 @@ def get_events_with_abstract_reviewer_convener(user, dt=None):
...
+        roles = data[principal.event_id]
+        if 'review_all_abstracts' in principal.permissions:
+            roles.add('abstract_reviewer')
+        if 'convene_all_abstracts' in principal.permissions:
+            roles.add('track_convener')
+
+    query = (user.in_track_acls
+             .options(load_only('track_id', 'permissions'))
+             .options(noload('*'))
+             .options(contains_eager(TrackPrincipal.track).load_only('event_id'))
```
ThiefMaster: You aren't using the track for any filtering, so no need for the explicit join + contains_eager - you can just `joinedload()` it instead

indico/modules/events/tracks/models/principals.py
```diff
@@ -0,0 +1,57 @@
...
+        nullable=False,
+        index=True
+    )
+
+    # relationship backrefs:
+    # - track (Track.acl_entries)
+
+    @return_ascii
+    def __repr__(self):
+        return format_repr(self, 'id', 'track_id', 'principal', read_access=False, full_access=False, permissions=[])
```
ThiefMaster: omit `read_access=False, full_access=False,` since those are *always* false anyway

indico/web/client/js/jquery/widgets/track_role_widget.js
```diff
@@ -5,86 +5,41 @@
...
     },
 
-    _setUp: function($trackConfig) {
-      var self = this;
-      $trackConfig.find('.js-role-container').each(function() {
-        self._setUpRoleContainer($(this));
-      });
+    _renderACLField: function(trackId, value) {
+      const permissionInfo = this.options.permissionsInfo;
+      const onChange = (newValue) => this._updateValue(newValue, trackId);
```
ThiefMaster: shouldn't prettier remove the `()` around `newValue`?

indico/web/client/js/jquery/widgets/track_role_widget.js
```diff
@@ -5,86 +5,41 @@
...
     },
 
     _create: function() {
-      var self = this;
-      var $container = self.options.containerElement;
-      self.roleData = JSON.parse(self.element.val());
-      $container.find('.js-track-config').each(function() {
-        self._setUp($(this));
+      const roleData = JSON.parse(this.element.val());
+      Object.entries(roleData).map(([trackId, value]) => {
```
ThiefMaster: Don't use `.map()` if you don't need the return value; use `.forEach()` instaed.

indico/util/marshmallow.py
```diff
@@ -160,11 +160,15 @@ def _deserialize(self, value, attr, data):
...
 
-    def __init__(self, principal_class, **kwargs):
+    def __init__(self, principal_class, explicit_permissions=False, **kwargs):
         self.principal_class = principal_class
+        self.explicit_permissions = explicit_permissions
         super(PrincipalPermissionList, self).__init__(**kwargs)
 
     def _serialize(self, value, attr, obj):
+        if self.explicit_permissions:
+            return [(entry.principal.identifier, sorted(get_principal_permissions(entry, type(obj))))
```
ThiefMaster: no need to call `type(obj)` for  every iteration.. assign `obj_cls = type(obj)` before and just use that ;)

indico/web/client/js/react/components/principals/TrackACLField.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+export default function TrackACLField({value, permissionInfo, onChange}) {
+  const [currentValue, setCurrentValue] = useState(value);
+  const permissionManager = new PermissionManager(permissionInfo.tree, permissionInfo.default);
+  return (
+    <FavoritesProvider>
+      {favoriteUsersController => (
+        <ACLField
+          value={currentValue}
+          disabled={false}
```
ThiefMaster: this prop is actually required? if not, just omit it

indico/web/client/js/react/components/principals/TrackACLField.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+import {ACLField} from 'indico/react/components';
+import {FavoritesProvider} from 'indico/react/hooks';
+import {PermissionManager} from 'indico/react/components/principals/util';
+
+export default function TrackACLField({value, permissionInfo, onChange}) {
+  const [currentValue, setCurrentValue] = useState(value);
+  const permissionManager = new PermissionManager(permissionInfo.tree, permissionInfo.default);
+  return (
+    <FavoritesProvider>
```
ThiefMaster: You don't need this, it's just a wrapper for class components which can't use hooks. Use `const favoriteUsersController = useFavoriteUsers();` directly.

indico/modules/events/schemas.py
meluru: Good point, I've used `Meta` before and then I forgot to change that

meluru:
> Many of the commits could probably be squashed together. I think the best approach to commits is to have them as logical units, so for example "Allow disabling protection mode in ProtectionMixin" is a change that does not affect existing code/functionality and not part of the "abstract permissions" feature itself, so it makes perfect sense as a separate commit.
> 
> But many/most other commits are really just steps from implementing the feature, so having those as separate commits isn't _that_ helpful...
> 
> (opinions on this are welcome @indico/development-team)

Good point, this one has definitely too many commits.

indico/web/client/js/react/components/principals/TrackACLField.jsx
meluru: It is required

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -522,9 +522,10 @@ def can_convene(self, user):
             return False
         elif not self.event.can_manage(user, permission='track_convener', explicit_permission=True):
             return False
-        elif self.event in user.global_convener_for_events:
+        elif self.event.can_manage(user, permission='convene_all_abstracts', explicit_permission=True):
             return True
-        elif user.convener_for_tracks & self.reviewed_for_tracks:
+        elif any(track.can_manage(user, permission='convene', explicit_permission=True)
```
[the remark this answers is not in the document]
meluru: Those shortcuts already exist, I actually wanted to avoid the repeated checks on user and event permissions so I didn't use the shortcut here

indico/web/client/js/react/components/principals/Search.jsx
```diff
@@ -461,19 +461,19 @@ export const UserSearch = ({withExternalUsers, initialFormValues, ...props}) =>
...
 
-UserSearch.defaultProps = {
+DefaultUserSearch.defaultProps = {
   ...InnerUserSearch.defaultProps,
   withExternalUsers: false,
   initialFormValues: {},
 };
 
-export const OverridableUserSearch = Overridable.component('OverridableUserSearch', UserSearch);
+export const UserSearch = Overridable.component('UserSearch', DefaultUserSearch);
```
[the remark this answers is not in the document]
meluru: This diff is wrong, I didn't change the name of overridable version of the component

indico/web/client/js/jquery/widgets/track_role_widget.js
meluru: I think prettier is ignoring this file

ThiefMaster: Indeed. You can run `npx eslint --no-ignore --fix indico/web/client/js/jquery/widgets/track_role_widget.js` to explicitly fix it.

indico/migrations/versions/20191023_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+def _downgrade_permissions():
+    conn = op.get_bind()
+    query_track_permissions = 'SELECT user_id, track_id, permissions FROM events.track_principals'
+    for user_id, track_id, permissions in conn.execute(query_track_permissions):
+        for permission in permissions:
+            if permission == 'review':
+                _create_abstract_reviewers_entry(conn, user_id, track_id=track_id)
+            if permission == 'convene':
```
mic4ael: `elif`? This way we will save an additional comparison

indico/migrations/versions/20191023_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,304 @@
...
+                _create_conveners_entry(conn, user_id, track_id=track_id)
+    query_event_permissions = '''
+        SELECT user_id, event_id, permissions FROM events.principals
+        WHERE 'review_all_abstracts' = ANY(permissions) OR 'convene_all_abstracts' = ANY(permissions)
```
mic4ael: I think a nicer way would be to use `ARRAY` operators

```sql
WHERE permissions && ARRAY['review_all_abstracts', 'convene_all_abstracts']::character varying[];
```

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -630,9 +632,11 @@ def get_track_question_scores(self):
 
     def get_reviewed_for_groups(self, user, include_reviewed=False):
         already_reviewed = {each.track for each in self.get_reviews(user=user)} if include_reviewed else set()
-        if self.event in user.global_abstract_reviewer_for_events:
+        if self.event.can_manage(user, permission='review_all_abstracts',  explicit_permission=True):
```
mic4ael:
```suggestion
        if self.event.can_manage(user, permission='review_all_abstracts', explicit_permission=True):
```

indico/modules/events/tracks/__init__.py
```diff
@@ -56,8 +54,30 @@ def _get_cloners(sender, **kwargs):
...
+
+@signals.acl.get_management_permissions.connect_via(Track)
+def _get_track_management_permissions(sender, **kwargs):
+    yield ConvenePermission
+
+
+class ConvenePermission(ManagementPermission):
+    name = 'convene'
+    friendly_name = _('Convene')
+    description = _('Grants track convener rights in a track')
```
mic4ael: Missing `.` or just redundant. (Since you have it in `GlobalConvenePermission.description`

indico/migrations/versions/20191023_1820_2496c4adc7e9_add_track_principal.py
meluru: this is autogenerated, I should have taken a closer look at that :) thanks for pointing out

indico/migrations/versions/20191023_1820_2496c4adc7e9_add_track_principal.py
meluru: but we can have both for one entry

meluru: I'm gonna change those checks :)

indico/util/marshmallow.py
```diff
@@ -160,11 +160,14 @@ def _deserialize(self, value, attr, data):
...
 
-    def __init__(self, principal_class, **kwargs):
+    def __init__(self, principal_class, explicit_permissions=False, **kwargs):
         self.principal_class = principal_class
+        self.explicit_permissions = explicit_permissions
         super(PrincipalPermissionList, self).__init__(**kwargs)
 
     def _serialize(self, value, attr, obj):
+        if self.explicit_permissions:
+            return [(entry.principal.identifier, set(entry.permissions)) for entry in value]
```
meluru: I've set the global track permissions as not user selectable (so that they don't appear in event permissions list), so I had to make a direct call on permissions here to be able to retrieve them

indico/core/db/sqlalchemy/protection.py
```diff
@@ -173,6 +186,8 @@ def can_access(self, user, allow_admin=True):
         # If there's a valid access key we can skip all other ACL checks
         elif self.allow_access_key and self.check_access_key():
             rv = True
+        elif self.disable_protection_mode:
```
ThiefMaster: Move this to the beginning of the function

indico/migrations/versions/20191111_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,277 @@
...
+        (%s,       %s,      %s,   false,       false,       %s)
+    ''', (track_id, user_id, PrincipalType.user.value, permissions))
+
+
+def _upgrade_permissions():
+    conn = op.get_bind()
+
+    # Create track acl entries
+
+    track_reviewers_conveners = '''
```
ThiefMaster: I'd just call this `stmt` since it only contains a sql statement

indico/migrations/versions/20191111_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,277 @@
...
+        SELECT user_id, track_id
+        FROM events.track_abstract_reviewers r
+        WHERE (
+            track_id IS NOT NULL AND NOT EXISTS (
+                SELECT 1 FROM events.track_conveners c
+                WHERE r.user_id = c.user_id AND r.track_id = c.track_id
+            )
+        )
+    '''
+
```
ThiefMaster: 👾

indico/migrations/versions/20191111_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,277 @@
...
+    for user_id, track_id in conn.execute(track_conveners):
+        _create_track_acl_entry(conn, user_id, track_id, ['convene'])
+
+    # Update event acl entries
+
+    event_reviewers = '''
+        SELECT reviewers.user_id, reviewers.event_id, principals.permissions
+        FROM events.track_abstract_reviewers as reviewers
```
ThiefMaster: no need for `as` + short aliases

indico/migrations/versions/20191111_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,277 @@
...
+        JOIN events.principals as principals
+        ON reviewers.user_id = principals.user_id AND reviewers.event_id = principals.event_id
```
ThiefMaster: ON usually goes in the same line as the JOIN

indico/migrations/versions/20191111_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,277 @@
...
+        WHERE reviewers.event_id IS NOT NULL
```
ThiefMaster: I think `track_id IS NULL` is clearer

indico/migrations/versions/20191111_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,277 @@
...
+    '''
+
+    for user_id, event_id, permissions in conn.execute(event_reviewers):
+        bisect.insort(permissions, 'review_all_abstracts')
+        _update_event_acl_entry(conn, user_id, event_id, permissions)
+
+    event_conveners = '''
+        SELECT conveners.user_id, conveners.event_id, principals.permissions
+        FROM events.track_conveners as conveners
```
ThiefMaster: aliases etc again

indico/migrations/versions/20191111_1820_2496c4adc7e9_add_track_principal.py
```diff
@@ -0,0 +1,277 @@
...
+        JOIN events.principals as principals
+        ON conveners.user_id = principals.user_id AND conveners.event_id = principals.event_id
+        WHERE conveners.event_id IS NOT NULL
```
ThiefMaster: I think `track_id IS NULL` is clearer

indico/modules/events/abstracts/util.py
```diff
@@ -238,27 +219,27 @@ def get_user_abstracts(event, user):
...
 def get_user_tracks(event, user):
     """Get the list of tracks where the user is a reviewer/convener"""
-    return _query_user_tracks(event, user).order_by(Track.title).all()
+    tracks = Track.query.with_parent(event).order_by(Track.title).all()
+    if (not event.can_manage(user, permission='review_all_abstracts', explicit_permission=True) and
+            not event.can_manage(user, permission='convene_all_abstracts', explicit_permission=True)):
+        return [track for track in event.tracks if
+                (track.can_manage(user, permission='review', explicit_permission=True) or
+                 track.can_manage(user, permission='convene', explicit_permission=True))]
+    return tracks
```
ThiefMaster: maybe invert the first if check and return early in that case?

indico/modules/events/tracks/models/principals.py
```diff
@@ -0,0 +1,56 @@
...
+        return (
+            db.CheckConstraint('NOT read_access', 'no_read_access'),
+            db.CheckConstraint('NOT full_access', 'no_full_access')
+        )
+
+    id = db.Column(
+        db.Integer,
+        primary_key=True
+    )
+
```
ThiefMaster: 👾

indico/modules/events/tracks/models/tracks.py
```diff
@@ -140,29 +126,27 @@ def locator(self):
...
 
+    def can_access(self, user):
+        # not relevant to tracks
+        raise NotImplementedError
+
     def can_delete(self, user):
         return self.event.can_manage(user) and not self.abstracts_accepted
 
-    def can_review_abstracts(self, user):
+    def can_review_abstracts(self, user, allow_admin=True):
```
ThiefMaster: `allow_admin` isn't used anywhere and `explicit_permissions` makes it ineffective anyway

indico/util/marshmallow.py
```diff
@@ -160,11 +160,14 @@ def _deserialize(self, value, attr, data):
 class PrincipalPermissionList(Field):
     """Marshmallow field for a list of principals and their permissions."""
 
-    def __init__(self, principal_class, **kwargs):
+    def __init__(self, principal_class, explicit_permissions=False, **kwargs):
```
ThiefMaster: `all_permissions`? and i'd explain it in the docstring

indico/util/marshmallow.py
```diff
@@ -160,11 +160,14 @@ def _deserialize(self, value, attr, data):
...
         self.principal_class = principal_class
+        self.explicit_permissions = explicit_permissions
         super(PrincipalPermissionList, self).__init__(**kwargs)
 
     def _serialize(self, value, attr, obj):
+        if self.explicit_permissions:
+            return [(entry.principal.identifier, set(entry.permissions)) for entry in value]
         return [(entry.principal.identifier, sorted(get_unified_permissions(entry))) for entry in value]
```
ThiefMaster: you could pass an arg to `get_unified_permissions` on whether to include all permissions even if full access is granted

indico/modules/events/abstracts/util.py
```diff
@@ -238,27 +219,27 @@ def get_user_abstracts(event, user):
 
 def get_visible_reviewed_for_tracks(abstract, user):
     event = abstract.event
-    if abstract.can_judge(user, check_state=True) or user in event.global_conveners:
+    if (abstract.can_judge(user, check_state=True) or
+            event.can_manage(user, permission='convene_all_abstracts', explicit_permission=True)):
         return abstract.reviewed_for_tracks
-    convener_tracks = {track for track in event.tracks if track.can_convene(user)}
+    convener_tracks = set(track for track in event.tracks
```
ThiefMaster: `{...}` instead of `set(...)`

PR #4107, opened 2019-10-14
Call for Papers in React

mic4ael:
WIP:
- [ ] edit a comment
- [ ] show/add/edit ratings
- [ ] render fields
- [ ] add/edit a review
- [ ] remove old CSS classes and replace them with SUI-React
- [x] submit a revision


indico/modules/events/papers/schemas.py
```diff
@@ -0,0 +1,121 @@
...
+from indico.modules.events.papers.models.files import PaperFile
+from indico.modules.events.papers.models.reviews import PaperReview
+from indico.modules.events.papers.models.revisions import PaperRevision, PaperRevisionState
+from indico.modules.events.schemas import EventSchema
+from indico.modules.users.schemas import UserSchema
+from indico.util.mimetypes import icon_from_mimetype
+
+
+class PaperFileSchema(mm.ModelSchema):
+    icon = Method('get_icon')
```
ThiefMaster: Doesn't really matter, but I'd probably just use an inline function here:

    icon = Function(lambda pf: icon_from_mimetype(pf.content_type))

indico/modules/events/papers/schemas.py
```diff
@@ -0,0 +1,121 @@
...
+                serialized_item.update(paper_review_schema.dump(timeline_item))
+            serialized.append(serialized_item)
+        return serialized
+
+
+class PaperRevisionSchema(mm.ModelSchema):
+    submitter = Nested(UserSchema)
+    judge = Nested(UserSchema)
+    spotlight_file = Nested(PaperFileSchema)
+    files = Nested(PaperFileSchema, many=True)
```
ThiefMaster: Read through marshmallow-code/marshmallow#779 and see whether we should use this or `List(Nested(PaperFileSchema))`

indico/modules/events/papers/schemas.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class PaperReviewSchema(mm.ModelSchema):
+    score = Float(as_string=True)
+    created_dt = DateTime()
+    user = Nested(UserSchema)
+    modified_dt = DateTime()
+    visibility = Nested(PaperCommentVisibilitySchema)
+    proposed_action = Nested(PaperAction)
+    group = Function(lambda review: {'title': str(review.group.title)})
```
ThiefMaster: `unicode`

indico/modules/events/schemas.py
```diff
@@ -0,0 +1,20 @@
...
+from indico.modules.events.models.events import Event
+
+
+class EventSchema(mm.ModelSchema):
+    class Meta:
+        model = Event
+        fields = ('id', 'title', 'is_locked')
+
+
+event_schema = EventSchema()
```
ThiefMaster: I think we can stop doing that and just instantiate Schema classes whenever we need them...

indico/modules/events/schemas.py
```diff
@@ -0,0 +1,20 @@
...
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from __future__ import unicode_literals
+
+from indico.core.marshmallow import mm
+from indico.modules.events.models.events import Event
+
+
+class EventSchema(mm.ModelSchema):
```
ThiefMaster: I don't like this name for a schema that simple. I'd rather have a `PaperEventSchema` inside the papers module that defines this (and if we end up needing this particular schema in multiple places we can always move it back with a more generic name like `LockedEventSchema`)

indico/web/client/js/react/components/MathJax.jsx
```diff
@@ -0,0 +1,23 @@
...
+// LICENSE file for more details.
+
+import React, {useEffect, useRef} from 'react';
+import PropTypes from 'prop-types';
+
+export default function MathJax({children}) {
+  const wrapperRef = useRef(null);
+
+  useEffect(() => {
+    jQuery(wrapperRef.current).mathJax();
```
ThiefMaster: `$` instead of `jQuery`?

ThiefMaster: But let's just export the non-jquery version of this and use that one here. Because all the jQuery one does is this:

```js
  $.fn.mathJax = function() {
    $(this).each(function() {
      pd.mathJax(this);
    });
    return this;
  };
```

indico/modules/events/papers/blueprint.py
```diff
@@ -9,13 +9,27 @@
...
 
 
 _bp = IndicoBlueprint('papers', __name__, url_prefix='/event/<confId>', template_folder='templates',
                       virtual_template_folder='events/papers')
 
+# CfP React
+_bp.add_url_rule('/papers/<int:contrib_id>/new', 'new_paper_timeline', display.RHNewPaperTimeline)
+
+# API
+_api_bp = IndicoBlueprint('papers_api', __name__, url_prefix='/event/<confId>/api', template_folder='templates',
```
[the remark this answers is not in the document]
mic4ael: So, the goal here was to have the same endpoint names without having to come up with some weird names e.g. there is already an endpoint `reset_paper_state`. At some point we will probably use just one endpoint.

indico/modules/events/papers/client/js/actions.js
```diff
@@ -0,0 +1,87 @@
...
+
+export const DELETE_COMMENT_REQUEST = 'papers/DELETE_COMMENT_REQUEST';
+export const DELETE_COMMENT_SUCCESS = 'papers/DELETE_COMMENT_SUCCESS';
+export const DELETE_COMMENT_ERROR = 'papers/DELETE_COMMENT_ERROR';
+
+export const JUDGE_PAPER_REQUEST = 'papers/JUDGE_PAPER_REQUEST';
+export const JUDGE_PAPER_SUCCESS = 'papers/JUDGE_PAPER_SUCCESS';
+export const JUDGE_PAPER_ERROR = 'papers/JUDGE_PAPER_ERROR';
+
+export function fetchPaperDetails(confId, contributionId) {
```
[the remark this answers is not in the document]
mic4ael: Right :P

indico/modules/events/papers/client/js/components/PaperContent.jsx
```diff
@@ -0,0 +1,73 @@
...
+import PaperMetadata from './PaperMetadata';
+
+export default function PaperContent() {
+  const {
+    lastRevision: {spotlightFile, files},
+    event: {id: eventId},
+    contribution: {id: contributionId},
+  } = useSelector(getPaperDetails);
+  const onlySpotlightFile = files.length === 1;
+  const content = [];
```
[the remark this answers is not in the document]
mic4ael: Yeah, I think that it can be easily done. Don't know why I did it like that

indico/modules/events/papers/client/js/index.js
```diff
@@ -79,3 +88,27 @@ import 'indico/modules/events/reviews';
...
+      user: {
+        fullName: Indico.User.full_name,
+        avatarBgColor: Indico.User.avatar_bg_color,
+      },
+    },
+  };
+  const store = createReduxStore('paper-timeline', {paper: paperReducers}, initialData);
+
+  ReactDOM.render(
+    React.createElement(Provider, {store}, React.createElement(Paper, {eventId, contributionId})),
```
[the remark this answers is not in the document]
mic4ael: Hm, would it incur some extra overhead when? We still have plenty of JS files in which we will not use `JSX` at all. IMO we shouldn't enforce that

indico/modules/events/papers/controllers/api.py
```diff
@@ -0,0 +1,100 @@
...
+from indico.modules.events.papers.models.reviews import PaperAction
+from indico.modules.events.papers.models.revisions import PaperRevisionState
+from indico.modules.events.papers.operations import delete_comment, judge_paper, reset_paper_state
+from indico.modules.events.papers.schemas import paper_schema
+from indico.web.args import use_kwargs
+
+
+class RHPaperDetails(RHPaperBase):
+    def _process(self):
+        return paper_schema.dumps(self.paper)
```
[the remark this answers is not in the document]
mic4ael: Oops, I forgot to actually change that :P

indico/modules/events/papers/controllers/api.py
```diff
@@ -0,0 +1,100 @@
...
+        if self.paper.state == PaperRevisionState.submitted:
+            return False
+        # managers and judges can always reset
+        if self.paper.event.can_manage(session.user) or self.paper.can_judge(session.user):
+            return True
+
+    def _process(self):
+        if self.paper.state != PaperRevisionState.submitted:
+            reset_paper_state(self.paper)
+        return paper_schema.dumps(self.paper)
```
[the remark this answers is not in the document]
mic4ael: Yeah, unfortunately, to have the consistent state in Redux I had to serialize the whole paper. Maybe another `GET` request wouldn't be a bad idea.. sth similar to `client.useBackend` in newdle?

indico/modules/events/papers/client/js/index.js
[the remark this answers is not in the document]
ThiefMaster: i don't think enabling jsx transforms would have a noticable overhead (at build time)..

anyway, i think especially short-term my second comment would be the better option

indico/modules/events/papers/controllers/api.py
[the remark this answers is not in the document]
ThiefMaster: I'd just dispatch the action to fetch the paper from the action after the request to do something has been successful. You don't even need to do anything redux-y for sending the update request itself.

indico/modules/events/papers/client/js/components/PaperFiles.jsx
```diff
@@ -0,0 +1,56 @@
...
+            href={paperFileDownloadURL({
+              file_id: file.id,
+              filename: file.filename,
+              confId: eventId,
+              contrib_id: contributionId,
+            })}
+            className={`attachment paper-file i-button text-color borderless ${file.icon}`}
+            title={file.filename}
+          >
+            <span className="title truncate-text">{file.filename}</span>
```
panagiotappl: Is translation missing here?

indico/web/client/js/react/components/MathJax.jsx
```diff
@@ -0,0 +1,23 @@
...
+// LICENSE file for more details.
+
+import React, {useEffect, useRef} from 'react';
+import PropTypes from 'prop-types';
+
+export default function MathJax({children}) {
+  const wrapperRef = useRef(null);
+
+  useEffect(() => {
+    window.mathJax(wrapperRef.current);
```
ThiefMaster: can't current be `null` initially?

indico/modules/events/papers/schemas.py
```diff
@@ -0,0 +1,126 @@
...
+        model = PaperReview
+        fields = ('id', 'score', 'created_dt', 'user', 'comment', 'modified_dt', 'visibility', 'proposed_action',
+                  'group', 'ratings')
+
+
+class PaperReviewCommentSchema(mm.ModelSchema):
+    user = Nested(UserSchema)
+    text = String()
+    visibility = Nested(PaperCommentVisibilitySchema)
+    created_dt = DateTime()
```
ThiefMaster: do you actually need this? it's a ModelSchema so most of the simple fields are usually autogenerated

indico/web/client/js/react/components/MathJax.jsx
mic4ael: Yep, but I haven't noticed any problems with that. However I think I will just make sure I don't pass `null` to `mathJax`

indico/web/client/js/react/components/FileSubmission.jsx
```diff
@@ -0,0 +1,113 @@
...
+      setFiles(acceptedFiles);
+      if (onChange) {
+        onChange(acceptedFiles);
+      }
+    },
+    [setFiles, onChange]
+  );
+
+  const removeFile = file => {
+    const newFiles = files.filter(f => !Object.is(f, file));
```
ThiefMaster: `===`

indico/web/client/js/react/components/FileSubmission.jsx
```diff
@@ -0,0 +1,113 @@
...
+                          raised
+                        >
+                          <Icon name="trash" onClick={() => removeFile(file)} color="red" />
+                          <Card.Content>
+                            <TooltipIfTruncated>
+                              <Card.Header textAlign="center">{file.name}</Card.Header>
+                            </TooltipIfTruncated>
+                            <Card.Meta textAlign="center">{`${(file.size / 1024).toFixed(
+                              2
+                            )} KB`}</Card.Meta>
```
ThiefMaster: It's `kB`.

Eventually we should add a util that formats a file size nicely (using the most suitable unit)

indico/modules/events/papers/client/js/setup.jsx
```diff
@@ -0,0 +1,39 @@
...
+      user: {
+        fullName: Indico.User.full_name,
+        avatarBgColor: Indico.User.avatar_bg_color,
+      },
+    },
+  };
+  const store = createReduxStore('paper-timeline', {paper: paperReducers}, initialData);
+
+  ReactDOM.render(
+    React.createElement(Provider, {store}, React.createElement(Paper, {eventId, contributionId})),
```
ThiefMaster: it's a `.jsx` file, you can use JSX in there! ;)

indico/modules/events/papers/client/js/reducers.js
```diff
@@ -0,0 +1,66 @@
...
+        return state;
+    }
+  },
+  permissions: (state = null, action) => {
+    switch (action.type) {
+      case paperActions.FETCH_PAPER_PERMISSIONS_SUCCESS:
+        return camelizeKeys(action.data);
+      case paperActions.DELETE_COMMENT_SUCCESS: {
+        const {commentId} = action.data;
+        const {comments: oldComments} = state;
```
ThiefMaster: not sure if destructuring really increases readability here ;)

indico/modules/events/papers/client/js/index.js
```diff
@@ -59,23 +61,27 @@ import 'indico/modules/events/reviews';
...
       };
 
-      var applySearchFilters = setupSearchBox(filterConfig);
+      const applySearchFilters = setupSearchBox(filterConfig);
       applySearchFilters();
     }
   };
 })(window);
+
+window.addEventListener('load', () => {
```
ThiefMaster: `document.addEventListener('DOMContentLoaded', ...);`?

indico/modules/events/papers/client/js/components/SubmitRevision.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+    try {
+      await indicoAxios.post(
+        submitRevisionURL({confId: eventId, contrib_id: contributionId}),
+        formData,
+        {headers}
+      );
+    } catch (e) {
+      handleAxiosError(e);
+      setSubmitting(false);
```
ThiefMaster: this one can go in a finally block since you run it in any case

indico/modules/events/papers/schemas.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class PaperRevisionSchema(mm.ModelSchema):
+    submitter = Nested(UserSchema)
+    judge = Nested(UserSchema)
+    spotlight_file = Nested(PaperFileSchema)
+    files = List(Nested(PaperFileSchema))
+    state = EnumField(PaperRevisionState)
+    timeline = PaperRevisionTimelineField()
+    judgment_comment = Function(lambda revision: escape(revision.judgment_comment))
```
ThiefMaster: IIRC we said that we wanted to have `judgment_comment_html` etc. for the rendered versions. This will be more important for anything that's editable, since you need the source version there as well.

indico/modules/events/papers/client/js/components/RevisionJudgment.jsx
```diff
@@ -0,0 +1,125 @@
...
+                header={
+                  <Header icon="warning sign" content={Translate.string('Confirm the operation')} />
+                }
+                open={confirmOpen}
+                content={Translate.string(
+                  'Do you really want to reset the judgment? This operation is irreversible.'
+                )}
+                onCancel={() => setConfirmOpen(false)}
+                onConfirm={() => {
+                  dispatch(resetPaperJudgment(event.id, contribution.id)).then(({error}) => {
```
ThiefMaster: why not async/await?

indico/modules/events/papers/client/js/components/RevisionJudgment.jsx
```diff
@@ -0,0 +1,125 @@
...
+                    // only hide the modal in case there was an error. This will prevent calling `setState` on
+                    // an already unmounted component.
+                    if (error) {
+                      setConfirmOpen(false);
```
ThiefMaster: actually does it even make sense to close the confirm in case of an error? I'd expect it to be open so I can try again if it failed..

indico/modules/events/papers/client/js/components/RevisionJudgment.jsx
```diff
@@ -0,0 +1,125 @@
...
+                <a
+                  href="#"
+                  className="i-link icon-remove"
+                  title={Translate.string('Reset judgment')}
+                  onClick={() => setConfirmOpen(true)}
+                />
+              </div>
+              <Confirm
+                header={
+                  <Header icon="warning sign" content={Translate.string('Confirm the operation')} />
```
ThiefMaster: Wouldn't that text make more sense as a child? According to the docs this is perfectly valid:

    <Header icon="...">content here</Header>

indico/modules/events/papers/client/js/components/RevisionInfo.jsx
```diff
@@ -0,0 +1,90 @@
...
+                <Translate>Show old revision</Translate>
+              </a>
+            )}
+          </div>
+          <div className={`i-box-content submission-info ${!isLastRevision ? 'weak-hidden' : ''}`}>
+            <ul className="file-list">
+              {_.sortBy(files, 'filename').map(file => (
+                <li className="truncate-text" key={file.id}>
+                  <a
+                    href={paperFileDownloadURL({
```
ThiefMaster: Maybe including the url in the schema would be better here instead of building it client-side?

indico/modules/events/papers/client/js/components/RevisionInfo.jsx
```diff
@@ -0,0 +1,90 @@
...
+              <time dateTime={serializeDate(submittedDt, moment.HTML5_FMT.DATETIME_LOCAL)}>
+                {serializeDate(submittedDt, 'LL')}
+              </time>
+            </div>
+            {!isLastRevision && (
+              <a
+                href="#"
+                className="revision-info-link i-link"
+                data-toggle={`#revision-timeline-${id}, #revision-info-${id} .submission-info`}
+                data-toggle-class={`{"target": "#revision-info-${id}", "class": "header-only"}`}
```
ThiefMaster: shouldn't we use react state for this stuff?

indico/modules/events/papers/client/js/components/RevisionComment.jsx
```diff
@@ -0,0 +1,129 @@
...
+                      negative
+                    />
+                  }
+                  closeIcon={!isDeletingCommentInProgress}
+                />
+              </div>
+            )}
+          </div>
+          <div className="i-box-content js-form-container">
+            <div className="markdown-text">{comment.text}</div>
```
ThiefMaster: still needs markdown rendering i guess?

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
```diff
@@ -0,0 +1,12 @@
...
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React from 'react';
+
+export default function PaperReviewForm() {
+  return <div />;
```
ThiefMaster: Let's render a TODO message instead?

indico/modules/events/papers/blueprint.py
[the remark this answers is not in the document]
ThiefMaster: What about `api_*` names for the endpoints instead? Separate blueprint are be problematic if we build URLs using blueprint-relative endpoint names during some requests (e.g. for email notifications).

indico/modules/events/papers/client/js/setup.jsx
mic4ael: Oh, forgot to actually change it after moving it to a separate file.

indico/modules/events/papers/client/js/components/RevisionInfo.jsx
mic4ael: Yup, we should ofc. But I didn't think it was that important for now.

indico/modules/events/papers/client/js/components/RevisionReview.jsx
```diff
@@ -0,0 +1,134 @@
...
+                            className="highlight bold semantic-text"
+                            title={Translate.string('Score: {score} / {maxScore}', {
+                              score: review.score,
+                              maxScore,
+                            })}
+                          />
+                        }
+                      />
+                    </Translate>{' '}
+                    {'('}
```
ThiefMaster: why not a single `{' ('}`?

indico/modules/events/papers/client/js/components/RevisionReview.jsx
```diff
@@ -0,0 +1,134 @@
...
+                    <a className="js-ratings-toggle toggle-link">
+                      <span className="js-show-ratings">
+                        <Translate>show ratings</Translate>
+                      </span>
+                      <span className="weak-hidden js-hide-ratings">
+                        <Translate>hide ratings</Translate>
+                      </span>
+                    </a>
+                    {')'}
```
ThiefMaster: actually, can't we just put a literal `)` here or will that result in an extra space?

indico/modules/events/papers/client/style/papers.scss
```diff
@@ -0,0 +1,10 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+.ui.page.dimmer {
+    z-index: 100 !important; // Make it less visible than Indico error-dialog
```
ThiefMaster: why do we need this? shouldn't the error dialog (which also uses a dimmer) be on top since it's rendered later?

indico/web/client/js/react/components/FileSubmission.jsx
```diff
@@ -0,0 +1,138 @@
...
+import PropTypes from 'prop-types';
+import Dropzone from 'react-dropzone';
+import {Button, Card, Divider, Grid, Header, Icon, Segment} from 'semantic-ui-react';
+
+import {TooltipIfTruncated} from 'indico/react/components';
+import {Param, Translate} from 'indico/react/i18n';
+
+import './FileSubmission.module.scss';
+
+function humanReadableBytes(bytes) {
```
ThiefMaster: What about moving this to util?

indico/web/client/js/react/components/FileSubmission.jsx
```diff
@@ -0,0 +1,138 @@
...
+  const kiloBytes = 1000;
+  const megaBytes = 1000 * kiloBytes;
+
+  if (bytes < kiloBytes) {
+    return (
+      <Translate>
+        <Param name="numberOfBytes" value={bytes} /> bytes
```
ThiefMaster: what about just `size` for those param names?

indico/web/client/js/react/components/FileSubmission.jsx
mic4ael: Hm, I wasn't sure about this. This is really specific to this component and I don't think we will use it anywhere else.

indico/modules/events/papers/client/style/papers.scss
mic4ael: Nope if you already have a SUIR dialog which has z-index of 101

indico/modules/events/papers/client/js/components/RevisionJudgment.jsx
```diff
@@ -0,0 +1,117 @@
...
+                />
+              </div>
+              <Confirm
+                header={Translate.string('Confirm the operation')}
+                open={confirmOpen}
+                content={Translate.string(
+                  'Do you really want to reset the judgment? This operation is irreversible.'
+                )}
+                onCancel={() => setConfirmOpen(false)}
+                onConfirm={async () => {
```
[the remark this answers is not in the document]
mic4ael: Don't really know how I ended up with `async` here :P

indico/modules/events/papers/schemas.py
```diff
@@ -0,0 +1,134 @@
...
+class PaperRevisionTimelineField(Field):
+    def _serialize(self, value, attr, obj, **kwargs):
+        if not value:
+            return []
+
+        serialized = []
+        for timeline_item in value:
+            serialized_item = {'timeline_item_type': timeline_item.timeline_item_type}
+            if timeline_item.timeline_item_type == 'comment':
+                serialized_item.update(PaperReviewCommentSchema(context=dict(self.context)).dump(timeline_item))
```
ThiefMaster: - instantiate the schema once before the loop
- no need to clone the context

indico/modules/events/papers/schemas.py
```diff
@@ -0,0 +1,134 @@
...
+            elif timeline_item.timeline_item_type == 'review':
+                serialized_item.update(PaperReviewSchema(context=dict(self.context)).dump(timeline_item))
```
ThiefMaster: same here

indico/modules/events/papers/client/js/components/RevisionJudgment.jsx
```diff
@@ -0,0 +1,116 @@
...
+import UserAvatar from './UserAvatar';
+
+export default function RevisionJudgment({revision}) {
+  const {state, judge, isLastRevision, judgmentCommentHtml, judgmentDt} = revision;
+  const {event, contribution, canJudge} = useSelector(getPaperDetails);
+  const [confirmOpen, setConfirmOpen] = useState(false);
+  const isResetInProgress = useSelector(isPaperStateResetInProgress);
+  const dispatch = useDispatch();
+
+  if (state === 'submitted') {
```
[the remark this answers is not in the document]
mic4ael: Yeah, I actually wanted to do that at some point but didn't know where to put it.

indico/modules/events/papers/client/js/components/PaperTimeline.jsx
```diff
@@ -0,0 +1,134 @@
...
+  const [toggled, setToggled] = useState([]);
+  const toggleRevisionInfo = id => {
+    if (toggled.includes(id)) {
+      setToggled(toggled.filter(el => el !== id));
+    } else {
+      setToggled([...toggled, id]);
+    }
+  };
+
+  return revisions.map((revision, index) => {
```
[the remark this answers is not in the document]
mic4ael: That is how it was before but I changed this to avoid passing callbacks to make the revision info toggleable.

indico/modules/events/papers/client/js/util.js
```diff
@@ -0,0 +1,13 @@
...
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+export const PaperState = {
+  accepted: 'accepted',
+  rejected: 'rejected',
+  submitted: 'submitted',
+  toBeCorrected: 'to_be_corrected',
```
ThiefMaster: Not sure about `camelCase` vs `snake_case` here... being able to search for the state name and find both backend and frontend code would be pretty convenient...

indico/modules/events/papers/schemas.py
```diff
@@ -0,0 +1,136 @@
...
+
+class PaperReviewSchema(mm.ModelSchema):
+    score = Float(as_string=True)
+    user = Nested(UserSchema)
+    visibility = Nested(PaperCommentVisibilitySchema)
+    proposed_action = Nested(PaperAction)
+    group = Function(lambda review: {'title': unicode(review.group.title)})
+    ratings = Function(lambda review: [{'value': rating.value} for rating in review.ratings])
+    comment_html = Function(lambda review: escape(review.comment))
+    can_edit = Function(lambda review, ctx: review.can_edit(ctx['user']))
```
ThiefMaster: `ctx.get('user')` so it works with no context as well

indico/modules/events/papers/client/js/components/TimelineItem.jsx
```diff
@@ -0,0 +1,117 @@
...
+  return (
+    <>
+      <div className="i-timeline">
+        <div className="i-timeline-item">
+          <UserAvatar user={submitter} />
+          <div
+            className={`i-timeline-item-box header-indicator-left ${
+              !isLastRevision && !visible ? 'header-only' : ''
+            }`}
+            id={`revision-info-${revision.id}`}
```
ThiefMaster: do we still need that id?

mic4ael: Yeah, link inside `PaperMetadata` allows to quickly jump to a specific revision.

indico/modules/events/papers/client/js/components/RevisionReview.jsx
```diff
@@ -0,0 +1,135 @@
...
+                          />
+                        }
+                      />
+                    </Translate>
+                    {' ('}
+                    <a className="js-ratings-toggle toggle-link">
+                      <span className="js-show-ratings">
+                        <Translate>show ratings</Translate>
+                      </span>
+                      <span className="weak-hidden js-hide-ratings">
```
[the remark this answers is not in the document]
mic4ael: Yep, ratings are still missing, so I could remove it altogether ;)

indico/modules/events/papers/client/js/util.js
```diff
@@ -0,0 +1,13 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+export const PaperState = {
```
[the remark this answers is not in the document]
mic4ael: Yep, I wasn't sure about the naming as well ;) so I am open to suggestions

indico/modules/events/papers/client/js/components/PaperDecisionForm.jsx
```diff
@@ -0,0 +1,94 @@
...
+  const canJudge = useSelector(canJudgePaper);
+
+  const submitPaperJudgment = useCallback(
+    formData => {
+      dispatch(judgePaper(eventId, contributionId, formData));
+    },
+    [dispatch, eventId, contributionId]
+  );
+
+  if (isLocked || !canJudge) {
```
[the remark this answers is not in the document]
mic4ael: Nope, I think that I should move it ;)

mic4ael:
> Indeed a pretty big PR. But it's very well structured and was a pleasure to read!

Thanks a lot for the review! ;)

indico/modules/events/papers/client/js/components/PaperDecisionForm.jsx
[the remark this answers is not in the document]
mic4ael: Ah, actually I now remember why I put it inside `PaperDecisionForm`. This boils down to the fact that `useSelector(isEventLocked)` or `useSelector(canJudgePaper)` will fail as long as the paper details have not been loaded. So in order to make it work I would have to either fix the selector (so that it first checks whether the required data have been loaded) or put it inside `PaperDecisionForm`.

indico/modules/events/papers/client/js/components/RevisionJudgment.jsx
```diff
@@ -0,0 +1,118 @@
...
+          {canJudge && isLastRevision && (
+            <>
+              <div className="hide-if-locked">
+                <a
+                  className="i-link icon-remove"
+                  title={Translate.string('Reset judgment')}
+                  onClick={() => setConfirmOpen(true)}
+                />
+              </div>
+              <Confirm
```
ThiefMaster: I think `size="tiny"` is big enough for it. Same for all the other confirmation prompts.

indico/modules/events/papers/client/js/actions.js
```diff
@@ -47,6 +52,20 @@ export function resetPaperJudgment(eventId, contributionId) {
   );
 }
 
+export function addComment(eventId, contributionId, commentData) {
+  const params = {
+    confId: eventId,
+    contrib_id: contributionId,
+  };
+
+  return ajaxAction(
```
ThiefMaster: `submitFormAction`

indico/modules/events/papers/client/js/actions.js
```diff
@@ -21,6 +22,10 @@ export const RESET_PAPER_JUDGMENT_REQUEST = 'papers/RESET_PAPER_JUDGMENT_REQUEST
 export const RESET_PAPER_JUDGMENT_SUCCESS = 'papers/RESET_PAPER_JUDGMENT_SUCCESS';
 export const RESET_PAPER_JUDGMENT_ERROR = 'papers/RESET_PAPER_JUDGMENT_ERROR';
 
+export const ADD_COMMENT_REQUEST = 'papers/ADD_COMMENT_REQUEST';
```
ThiefMaster: tracking the state won't be needed when using `fprops.submitting` instead

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
```diff
@@ -5,8 +5,111 @@
...
+    {
+      value: 'contributors',
+      text: Translate.string('Visible to contributors, reviewers and judges'),
+    },
+  ];
+
+  const onCommentClickHandler = commentFormVisible
+    ? null
+    : () => {
+        setCommentFormVisible(true);
```
ThiefMaster: I'd check for `commentFormVisible` inside.. looks less ugly IMHO

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
```diff
@@ -5,8 +5,111 @@
...
+      };
+
+  const createComment = useCallback(
+    formData => {
+      dispatch(addComment(eventId, contributionId, formData));
```
ThiefMaster:
```js
const rv = await dispatch(...);
if (rv.error) {
  return rv.error;
}
```

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
```diff
@@ -5,8 +5,111 @@
...
+
+  return (
+    <div className="i-timeline-item" styleName="review-timeline-input">
+      <UserAvatar user={user} />
+      <div className="i-timeline-item-box footer-only header-indicator-left">
+        <div className="i-box-footer" style={{overflow: 'visible'}}>
+          {canComment && (
+            <div className="flexrow">
+              <div className="f-self-stretch">
+                <FinalForm
```
ThiefMaster: `subscription={{submitting: true}}`

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
```diff
@@ -5,8 +5,111 @@
...
+                  onSubmit={createComment}
+                  initialValues={{comment: '', visibility: 'judges'}}
+                >
+                  {fprops => (
+                    <Form
+                      id="comment-form"
+                      onSubmit={fprops.handleSubmit}
+                      loading={isCommentBeingAdded}
```
ThiefMaster: `fprops.submitting`

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
```diff
@@ -5,8 +5,111 @@
...
+                            options={visibilityOptions}
+                            selection
+                            required
+                          />
+                          <Form.Group inline>
+                            <FinalSubmitButton
+                              label={Translate.string('Comment')}
+                              form="comment-form"
+                            />
+                            <Button
```
ThiefMaster: this one should also be disabled while submitting

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
```diff
@@ -5,8 +5,111 @@
...
+                          <FinalDropdown
+                            name="visibility"
+                            options={visibilityOptions}
+                            selection
+                            required
+                          />
+                          <Form.Group inline>
+                            <FinalSubmitButton
+                              label={Translate.string('Comment')}
+                              form="comment-form"
```
ThiefMaster: we're inside the form so this is most likely not needed (neither is the `id` on the form in this case)

indico/modules/events/papers/client/js/components/RevisionJudgment.jsx
```diff
@@ -31,85 +31,87 @@ export default function RevisionJudgment({revision}) {
...
+                  changes.
+                </Translate>
+              )}{' '}
+              <time dateTime={serializeDate(judgmentDt, moment.HTML5_FMT.DATETIME_LOCAL)}>
+                {serializeDate(judgmentDt, 'LL')}
+              </time>
+            </div>
+            {canJudge && isLastRevision && (
+              <>
+                <div className="hide-if-locked">
```
ThiefMaster: if `canJudge` takes the event being locked into account we don't need this (it's just a quick way to hide action buttons etc. in locked events without the need for explicit checks or JS)

indico/modules/events/papers/client/js/reducers.js
```diff
@@ -27,6 +27,11 @@ export default combineReducers({
       paperActions.RESET_PAPER_JUDGMENT_SUCCESS,
       paperActions.RESET_PAPER_JUDGMENT_ERROR
     ),
+    addComment: requestReducer(
```
ThiefMaster: can probably go away when using `fprops.submitting`

indico/modules/events/papers/client/js/selectors.js
```diff
@@ -15,6 +15,8 @@ export const isJudgingInProgress = state =>
   state.paper.requests.judgment.state === RequestState.STARTED;
 export const isPaperStateResetInProgress = state =>
   state.paper.requests.resetState.state === RequestState.STARTED;
+export const isAddingComment = state =>
```
ThiefMaster: can probably go away when using `fprops.submitting`

indico/modules/events/papers/client/js/components/PaperContent.jsx
```diff
@@ -0,0 +1,66 @@
...
+            files={files.filter(file => file.id !== spotlightFile.id)}
+          />
+        )}
+      </>
+    );
+  } else {
+    content = (
+      <div key="paper-files" className="paper-files-list flexrow f-a-baseline">
+        <div className="paper-files-list-label f-self-no-shrink">
+          <Translate>Paper files</Translate>
```
meluru: Do we need that? I think it's clear without the title
image

indico/web/client/js/react/components/FileSubmission.jsx
```diff
@@ -0,0 +1,138 @@
...
+  } else {
+    return (
+      <Translate>
+        <Param name="size" value={(bytes / megaBytes).toFixed(2)} /> MB
+      </Translate>
+    );
+  }
+}
+
+export default function FileSubmission({onChange}) {
```
ThiefMaster: the dropzone and buttons inside should be disabled based on a `disabled` prop

indico/modules/events/papers/client/js/components/SubmitRevision.jsx
```diff
@@ -0,0 +1,86 @@
...
+        <div className="i-box-header flexrow">
+          <Translate>Upload corrected revision</Translate>
+        </div>
+        <div className="i-box-content">
+          <FileSubmission onChange={newFiles => setFiles(newFiles)} />
+          <Button
+            onClick={submitFiles}
+            content={Translate.string('Submit new revision')}
+            style={{marginTop: 10}}
+            disabled={files.length === 0}
```
ThiefMaster: also disable it while submitting

indico/modules/events/papers/client/js/components/PaperContent.jsx
mic4ael: For now I just did it 1:1 to what we had before. I think that along the way we will improve it anyway ;)

mic4ael:
> Looks good overall, the code is very clean, good job :)
> 
> Do you plan to improve the animation when showing/hiding the old revision ?

I didn't want to spend much more time on this than necessary since it is something related to the `SUIR` `Transition`

indico/web/client/styles/modules/_papers.scss
```diff
@@ -243,10 +243,14 @@
...
+        margin: 0 3px;
 
         &:hover {
             color: $light-black;
         }
     }
 }
+
+.paper-timeline {
+    min-height: 500px;
```
meluru: Why do we need this?

mic4ael: To make the `Loader` look nice when fetching the paper-related data ;)

mic4ael: By nicer I mean centered

indico/modules/events/papers/client/js/components/PaperDecisionForm.jsx
```diff
@@ -0,0 +1,94 @@
...
+            {fprops => (
+              <Form id="judgment-form" onSubmit={fprops.handleSubmit}>
+                <FinalDropdown name="action" options={actionOptions} selection required />
+                <FinalTextArea
+                  name="comment"
+                  placeholder={Translate.string('Leave a comment for the submitter...')}
+                />
+                <FinalSubmitButton
+                  label={Translate.string('Judge')}
+                  form="judgment-form"
```
ThiefMaster: probably not needed

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
```diff
@@ -0,0 +1,112 @@
...
+      <UserAvatar user={user} />
+      <div className="i-timeline-item-box footer-only header-indicator-left">
+        <div className="i-box-footer" style={{overflow: 'visible'}}>
+          {canComment && (
+            <div className="flexrow">
+              <div className="f-self-stretch">
+                <FinalForm
+                  onSubmit={createComment}
+                  initialValues={{comment: '', visibility: 'judges'}}
+                  subscription={{values: true, submitting: true}}
```
ThiefMaster: no need to subscribe to `values`

indico/modules/events/papers/client/js/components/RevisionComment.jsx
```diff
@@ -0,0 +1,113 @@
...
+                  className="i-link icon-cross js-delete-comment"
+                  title={Translate.string('Remove comment')}
+                />
+                <Confirm
+                  size="tiny"
+                  header={Translate.string('Remove comment')}
+                  open={confirmOpen}
+                  content={Translate.string('Are you sure you want to remove this comment?')}
+                  onCancel={() => setConfirmOpen(false)}
+                  onConfirm={() => {
```
ThiefMaster: Does the Confirm auto-close? If not, what happens if the request is slow and you click twice?

indico/web/client/styles/modules/_papers.scss
meluru: I see now, cool :)

meluru:
> > Looks good overall, the code is very clean, good job :)
> > Do you plan to improve the animation when showing/hiding the old revision ?
> 
> I didn't want to spend much more time on this than necessary since it is something related to the `SUIR` `Transition`

Makes sense, tet's make a task for it once this is merged then

indico/modules/events/papers/client/js/components/RevisionComment.jsx
mic4ael: Discussed in person

ThiefMaster:
Adding/removing a comment results in a reload of the paper which temporarily replaces it with a spinner. This results in the scroll position being reset (since the page becomes much shorter.

Let's only show the spinner instead of the paper during the first load (ie when there's no paper in the store).

mic4ael:
> Adding/removing a comment results in a reload of the paper which temporarily replaces it with a spinner. This results in the scroll position being reset (since the page becomes much shorter.
> 
> Let's only show the spinner instead of the paper during the first load (ie when there's no paper in the store).

Honestly, when I was developing it I wasn't sure whether users would like the `Loader` on each action. I think that showing it on the initial page load is the right thing to do.

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
```diff
@@ -0,0 +1,119 @@
...
+      setCommentFormVisible(true);
+    }
+  };
+
+  const createComment = useCallback(
+    async (formData, form) => {
+      const rv = await dispatch(addComment(eventId, contributionId, formData));
+
+      setCommentFormVisible(false);
+      setTimeout(() => form.reset(), 0);
```
ThiefMaster: move it after the error check

indico/modules/events/papers/client/js/components/RevisionComment.jsx
```diff
@@ -0,0 +1,120 @@
...
+      <UserAvatar user={comment.user} />
+      <div className="flexrow i-timeline-item-content">
+        <div className="i-timeline-item-box header-indicator-left">
+          <div className="i-box-header flexrow">
+            <div className="f-self-stretch">
+              <Translate>
+                <Param name="userName" value={comment.user.fullName} wrapper={<strong />} /> left a
+                comment
+              </Translate>{' '}
+              {comment.visibility !== 'contributors' && (
```
ThiefMaster: `CommentVisibility` enum?

indico/modules/events/papers/controllers/api.py
```diff
@@ -0,0 +1,108 @@
...
+            reset_paper_state(self.paper)
+        return '', 204
+
+
+class RHSubmitPaperComment(RHPaperBase):
+    def _check_paper_protection(self):
+        return self.paper.can_comment(session.user)
+
+    @use_kwargs({
+        'comment': fields.String(required=True),
```
ThiefMaster: `, validate=validate.Length(min=1)` - otherwise empty strings are accepted

PR #4114, opened 2019-10-27
Add traditional Chinese (zh_TW) translations.

fweng322:
Add translations for Traditional Chinese (zh_TW, zh_Hant_TW).


ThiefMaster:
Thanks, but we do not use Pull Requests for translations since we use Transifex as a central place for all translations (and then pull it from there before a release).

I'll ping @mic4ael to contact you with details on how to start using Transifex since he's managing i18n for the project. Since you already have .po files it's most likely easily possible to import your existing translation there.

PR #4116, opened 2019-10-28
Track groups

### LINKED ISSUE #4080: Show track group name in addition to track name
For example, when submitting an abstract we only show the track names, but we should show e.g. `groupname: trackname` instead.

panagiotappl:
This PR is still in **Draft mode**

- Fixes 3 track sorting bugs
    -  when dragging large track inside group
    -  fix sorting endpoint being called twice
    - don't allow dragging group inside another group
- Adds `SelectMultipleWidget` and `_SingleChoiceQuerySelectMultipleFieldGrouped` to support track group display (as `<optgroup />`) in track select field
- Adds group support in MultipleSelectFields
- Shows group name in various places in Abstract View page

closes #4080


indico/modules/events/abstracts/forms.py
```diff
@@ -534,8 +562,26 @@ def process_formdata(self, valuelist):
         super(_SingleChoiceQuerySelectMultipleField, self).process_formdata(valuelist)
 
 
+class _SingleChoiceQuerySelectMultipleFieldGrouped(_SingleChoiceQuerySelectMultipleField):
+    widget = SelectMultipleWidget()
+
+    def __init__(self, *args, **kwargs):
+        if 'get_group' in kwargs:
+            self.get_group = kwargs.pop('get_group')
```
ThiefMaster: self.get_group = kwargs.pop('get_group', lambda x: x)

indico/modules/events/abstracts/forms.py
```diff
@@ -534,8 +562,26 @@ def process_formdata(self, valuelist):
...
+        else:
+            self.get_group = lambda x: x
+        super(_SingleChoiceQuerySelectMultipleFieldGrouped, self).__init__(*args, **kwargs)
+
+    def iter_choices(self):
+        yield ('__None', self.blank_text, self.data is None, None)
+        for pk, obj in self._get_object_list():
+            yield (pk, self.get_label(obj), obj in self.data, self.get_group(obj))
```
ThiefMaster: no need for the outermost `()` around everything

indico/modules/events/abstracts/forms.py
```diff
@@ -514,6 +515,33 @@ def __init__(self, *args, **kwargs):
...
+    """
+
+    def __call__(self, field, **kwargs):
+        kwargs.setdefault('id', field.id)
+        if self.multiple:
+            kwargs['multiple'] = True
+        html = ['<select %s>' % html_params(name=field.name, **kwargs)]
+        group_iter = groupby(field.iter_choices(), lambda (_val, _label, _selected, group): group)
+        for group, tracks in group_iter:
+            print group, tracks
```
ThiefMaster: :eyes:

indico/modules/events/abstracts/forms.py
```diff
@@ -514,6 +515,33 @@ def __init__(self, *args, **kwargs):
...
+            if group:
+                html.append('<optgroup %s>' % html_params(label=group.title))
```
ThiefMaster: use `get_label` or add a new `get_group_label` kwarg to the field

panagiotappl: Why is this neceessary here?

ThiefMaster: You already have `get_group` to make this field pretty generic, so I think it'd make sense to make this part generic as well instead of hardcoding a title attribute

panagiotappl: I see what you mean

indico/modules/events/abstracts/forms.py
```diff
@@ -514,6 +516,32 @@ def __init__(self, *args, **kwargs):
...
+        for (group, group_label), tracks in group_iter:
+            if group:
+                html.append('<optgroup %s>' % html_params(label=group_label))
+            for val, label, selected, _group, _group_label in tracks:
+                option_params = {'value': val}
+                if selected:
+                    option_params['selected'] = 'true'
+                html.append('<option %s>%s</option>' % (html_params(**option_params), label))
+            if group:
+                html.append("</optgroup>")
```
mic4ael: mixing `"` with `'`

indico/modules/events/abstracts/forms.py
```diff
@@ -549,8 +593,48 @@ def __init__(self, *args, **kwargs):
...
+        html = []
+        id = 0
+        for (group, group_label), tracks in group_iter:
+            if group:
+                html.append('<label><strong>%s</strong></label>' % group_label)
+            else:
+                html.append('<label><i>No group</i></label>')
+            for val, label, selected, _group, _group_label in tracks:
+                input_id = '{}-{}'.format(field.id, str(id))
+                html.append('<div class="i-checkbox ">')
```
mic4ael:
```suggestion
                html.append('<div class="i-checkbox">')
```

PR #4118, opened 2019-10-29
Ignore HTTP API errors from `Authorization: Basic` header

uxmaster:
Ignore errors from Basic HTTP Authorization headers, see
 * https://talk.getindico.io/t/http-api-error-with-basic-auth-header/1450
 * https://talk.getindico.io/t/repeated-unexpected-exception-errors/987.


ThiefMaster:
Couldn't we just skip the oauth check altogether if there's a basic auth header?

uxmaster:
I think we could, but I'm not so familiar with the Indico code, and Python is not my native language. Handling the exception seemed easier.

uxmaster:
@ThiefMaster, I've got more familiar with the code and prepared another pull request based on your comment: https://github.com/indico/indico/pull/4123

Please merge the one you like more or tell me what I should change. We really need these errors fixed, and can only update the Indico installation at KEK if the patch is official.

PR #4123, opened 2019-10-31
Check OAuth only if `Authorization` header is OK

uxmaster:
So only if there is an `Authorization` header, and its value does not start with `Basic`.

This fixes the following errors:
 * https://talk.getindico.io/t/http-api-error-with-basic-auth-header/1450
 * https://talk.getindico.io/t/repeated-unexpected-exception-errors/987.

### DIFF
--- a/indico/web/http_api/handlers.py
+++ b/indico/web/http_api/handlers.py
@@ -129,8 +129,8 @@ def handler(prefix, path):
     onlyPublic = get_query_parameter(queryParams, ['op', 'onlypublic'], 'no') == 'yes'
     onlyAuthed = get_query_parameter(queryParams, ['oa', 'onlyauthed'], 'no') == 'yes'
     scope = 'read:legacy_api' if request.method == 'GET' else 'write:legacy_api'
-    authHeader = request.headers.get('Authorization')
-    if authHeader and authHeader[:5].lower() != 'basic':
+
+    if not request.headers.get('Authorization', '').lower().startswith('basic '):
         try:
             oauth_valid, oauth_request = oauth.verify_request([scope])
             if not oauth_valid and oauth_request and oauth_request.error_message != 'Bearer token not found.':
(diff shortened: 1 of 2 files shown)


indico/web/http_api/handlers.py
```diff
@@ -128,16 +128,20 @@ def handler(prefix, path):
...
-        if not oauth_valid and oauth_request and oauth_request.error_message != 'Bearer token not found.':
-            raise BadRequest('OAuth error: {}'.format(oauth_request.error_message))
-        elif g.get('received_oauth_token') and oauth_request.error_message == 'Bearer token not found.':
-            raise BadRequest('OAuth error: Invalid token')
-    except ValueError:
-        # XXX: Dirty hack to workaround a bug in flask-oauthlib that causes it
-        #      not to properly urlencode request query strings
-        #      Related issue (https://github.com/lepture/flask-oauthlib/issues/213)
+    authHeader = request.headers.get('Authorization')
+    if authHeader and authHeader[:5].lower() != 'basic':
```
ThiefMaster: We don't use camelCase anymore. Those places where we do are legacy leftovers.

And I think it'd be nicer to combine those two lines into one:

```python
    if request.headers.get('Authorization', '').lower().startswith('basic ')
```

uxmaster: Oh sorry about the camelCase. I mimicked the code just before the patch.

I can combine the two lines but it would then do the OAuth when there's no `Authorization` header present.

Also,
```
authHeader[:5].lower() != 'basic'
```
is faster than
```
request.headers.get('Authorization', '').lower().startswith('basic ')
```
see https://stackoverflow.com/questions/13578916/case-insensitive-string-startswith-in-python

HTTP Headers may be up to 16K on some servers.

What do you think?

uxmaster:
Thanks a lot for merging. Could you merge it also into the `indico:2.1-maintenance` branch?

I'm asking because apparently, we have an old 2.1 branch at KEK and this is also the case at BNL (the original issue reporters).

ThiefMaster:
I'm afraid not - this is not a critical fix that justifies backporting it to an otherwise unsupported version (and making a release).

- If you don't use HTTP basic auth, you can just disable the error emails and not bother about the errors getting logged
- If you do use HTTP basic auth, either configure your webserver to exempt the affected API routes from it or apply the patch directly to your code.

And of course, I strongly recommend you to update to 2.2 as soon as possible. We will include the fix in the next 2.2 patch release soon.

uxmaster:
When do you plan to do the next 2.2 patch release?

ThiefMaster:
I actually wanted to do it earlier this week but we were busy with some other things....

Tomorrow or monday should work though :)

PR #4126, opened 2019-11-05
Allow restricting admin privs to secure logins (e.g. 2FA)

ThiefMaster:
This depends on indico/flask-multipass#28, so the minimum multipass version should be bumped to 0.3 before merging this! (leaving it as draft until that's merged and released)

Also, there are of course alternatives on how to implement this, e.g. by having something less magic, but it would require changing MANY places, and often we only pass around the user


ThiefMaster:
Considering that optional/step-up 2FA is no longer really common and always-on-2FA is what everyone uses nowadays, this is no longer relevant.

PR #4132, opened 2019-11-07
Add generic file upload system

### LINKED ISSUE #4130: Create backend to receive paper files
An API endpoint which receives a single file and returns a response containing a UUID.
The file should be present in the body of the request (no weird form encodings etc...)

ThiefMaster:
closes #4130

---

To add the upload endpoint e.g. for the editing workflow, add a RH similar to this one that uses a suitable context:

```python
class RHEditingUploadFile(UploadFileMixin, RHWhateverContributionBase):
    def get_file_context(self):
        return 'event', self.event.id, 'papers', self.contribution.id
```

### DIFF
--- a/indico/modules/files/tasks.py
+++ b/indico/modules/files/tasks.py
@@ -13,6 +13,7 @@ from celery.schedules import crontab
 from sqlalchemy.orm.attributes import flag_modified
 
 from indico.core.celery import celery
+from indico.core.config import config
 from indico.core.db import db
 from indico.core.storage import StorageReadOnlyError
 from indico.modules.files import logger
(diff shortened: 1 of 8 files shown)


indico/modules/files/tasks.py
```diff
@@ -0,0 +1,46 @@
...
+from indico.modules.files.models.files import File
+from indico.util.date_time import now_utc
+
+
+@celery.periodic_task(name='delete_unclaimed_files', run_every=crontab(minute='0', hour='6'))
+def delete_unclaimed_files():
+    unclaimed_files = (File.query
+                       .filter(~File.claimed,
+                               File.created_dt <= (now_utc() - timedelta(days=1)),
+                               ~File.meta.op('?')('deletion_failed'))
```
panagiotappl: Maybe you could add a comment explaining what this part of the query does.

Issue #4120, opened 2019-10-30
Paper Editing: create management interface to set up file types


pferreir:
We'll need a management interface to set up which file "types" are expected per `Reviewable` type. These types could be restricted by MIME types/extension. A "file type" could include one or several files.

For now the main task would be to create a React-based interface which lists the existing types and contains (mock) action buttons to add/remove them.

PR #4133, opened 2019-11-08
Add EditingFileType model

### LINKED ISSUE #4121: Paper Editing: create DB models for file management
We need some sort of entity which represents a set of files and which will be used in #4119 and #4120.

ThiefMaster:
An empty `extensions` list indicates any extension is valid.

closes #4121

---

Later we will add a frontend for it (#4120); probably a simple wtforms-based form does the job well enough.

For now, create file types using `indico shell`:

```python
e = E(EVENT_ID)
e.editing_file_types = [EditingFileType(name='PDF', extensions=['pdf'], required=True, publishable=True), EditingFileType(name='Source', extensions=['tex', 'docx', 'doc', 'txt'], required=True, allow_multiple_files=True)]
db.session.commit()
```

### DIFF
--- a/indico/modules/events/editing/models/file_types.py
+++ b/indico/modules/events/editing/models/file_types.py
@@ -1,0 +1,72 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2019 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from __future__ import unicode_literals
+
+from sqlalchemy.dialects.postgresql import ARRAY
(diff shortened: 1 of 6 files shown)


indico/modules/events/editing/models/file_types.py
```diff
@@ -0,0 +1,72 @@
...
+            'editing_file_types',
+            cascade='all, delete-orphan',
+            lazy=True
+        )
+    )
+
+    @return_ascii
+    def __repr__(self):
+        return format_repr(self, 'id', 'extensions', allow_multiple_files=False, required=False, publishable=False,
+                           _text=self.name)
```
meluru: How about adding event id here as well?

PR #4135, opened 2019-11-11
Fix legacy time table pdf page breaks

bpedersen2:
If 'Each session on a separate page' was selected,
only the first of many parallel sessions appeared in the
pdf.

Do not try to remove the previous page breaks.

The takewhile would only consider entries up to the first page break,
effectivly hiding all other parallel sessions instead of removing
only the last page break.


ThiefMaster:
Oh, good catch! This bug only happens when the *last* top-level timetable entry of the day is a contribution/break and the option you mentioned is enabled.

PR #4137, opened 2019-11-11
Paper editing: FileManager component

### LINKED ISSUE #4119: Paper Editing: create widget to manage files
* Event manager should be able to define "file types", restricted by content type/extension (see task #4120);
 * Authors will be shown inputs for the different file types;
 * When uploading a new revision, the author/editor should be able to select which files to keep from the previous one and which ones to discard (internally, we'd save a full a snapshot);
 * A "file type" can require a single file or several files.

image

panagiotappl:
closes #4119

### DIFF
--- a/indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
+++ b/indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
@@ -1,0 +1,104 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React, {useContext, useRef} from 'react';
+import PropTypes from 'prop-types';
+import Dropzone from 'react-dropzone';
+import {Icon} from 'semantic-ui-react';
+import {FileManagerContext, filePropTypes} from './util';
+import styles from './FileManager.module.scss';
+
+export default function FileList({files, fileTypeId, multiple}) {
+  const dispatch = useContext(FileManagerContext);
+  const ref = useRef(null);
+
+  function onModifyDrop(fileId) {
+    return acceptedFiles => {
+      dispatch({
+        type: 'MODIFY',
+        fileTypeId,
+        fileId,
+        file: {name: acceptedFiles[0].name, id: 'file2', url: 'file2url', claimed: false},
+      });
+    };
+  }
+
+  return (
+    <ul styleName="file-list">
+      {files.map(({name, id, state}) => (
+        <li key={id} styleName="file-row">
+          <span className={`${styles['file-state']} ${state || ''}`}>{name}</span>
+          <span>
+            {!state && multiple && (
+              <>
+                <Icon
+                  styleName="exchange-icon"
+                  name="exchange"
+                  onClick={() => {
+                    ref.current.open();
+                  }}
+                />
+                <Dropzone ref={ref} onDrop={onModifyDrop(id)}>
+                  {({getRootProps, getInputProps}) => (
+                    <span {...getRootProps()}>
+                      <input {...getInputProps()} />
+                    </span>
+                  )}
+                </Dropzone>
+              </>
+            )}
+            {state !== 'deleted' && state !== 'modified' && (
+              <Icon
+                styleName="delete-icon"
+                onClick={() =>
+                  dispatch({
+                    type: 'DELETE',
+                    fileTypeId,
+                    fileId: id,
+                  })
+                }
+                name="trash"
+              />
+            )}
+            {state === 'deleted' && (
+              <Icon
+                styleName="undo-icon"
+                onClick={() =>
+                  dispatch({
+                    type: 'UNDELETE',
+                    fileTypeId,
+                    fileId: id,
+                  })
+                }
+                name="undo"
+              />
+            )}
+            {state === 'modified' && (
+              <Icon
+                styleName="undo-icon"
+                onClick={() =>
+                  dispatch({
+                    type: 'REVERT',
+                    fileTypeId,
+                    fileId: id,
+                  })
+                }
+                name="undo"
+              />
+            )}
+          </span>
+        </li>
+      ))}
+    </ul>
+  );
+}
+
+FileList.propTypes = {
+  files: filePropTypes.isRequired,
+  fileTypeId: PropTypes.string.isRequired,
+  multiple: PropTypes.bool.isRequired,
+};
--- a/indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
+++ b/indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
@@ -36,7 +36,12 @@ FileAction.propTypes = {
   className: PropTypes.string.isRequired,
 };
 
-function FileEntry({eventId, fileTypeId, multiple, file: {uuid, filename, state, claimed}}) {
+function FileEntry({
+  eventId,
+  fileTypeId,
+  allowMultipleFiles,
+  file: {uuid, filename, state, claimed},
+}) {
   const ref = useRef(null);
   const dispatch = useContext(FileManagerContext);
   const [activeButton, setActiveButton] = useState(null);
@@ -47,7 +52,7 @@ function FileEntry({eventId, fileTypeId, multiple, file: {uuid, filename, state,
         {filename}
       </span>
       <span>
-        {!state && multiple && (
+        {!state && allowMultipleFiles && (
           <>
             <FileAction
               icon="exchange"
@@ -117,17 +122,22 @@ function FileEntry({eventId, fileTypeId, multiple, file: {uuid, filename, state,
 
 FileEntry.propTypes = {
   eventId: PropTypes.string.isRequired,
-  fileTypeId: PropTypes.string.isRequired,
-  multiple: PropTypes.bool.isRequired,
+  fileTypeId: PropTypes.number.isRequired,
+  allowMultipleFiles: PropTypes.bool.isRequired,
   file: PropTypes.shape(filePropTypes).isRequired,
 };
 
-export default function FileList({files, fileTypeId, multiple, eventId}) {
+export default function FileList({files, fileTypeId, allowMultipleFiles, eventId}) {
   return (
     <ul styleName="file-list">
       {files.map(file => (
         <li key={file.uuid} styleName="file-row">
-          <FileEntry fileTypeId={fileTypeId} eventId={eventId} multiple={multiple} file={file} />
+          <FileEntry
+            fileTypeId={fileTypeId}
+            eventId={eventId}
+            allowMultipleFiles={allowMultipleFiles}
+            file={file}
+          />
         </li>
       ))}
     </ul>
@@ -136,7 +146,7 @@ export default function FileList({files, fileTypeId, multiple, eventId}) {
 
 FileList.propTypes = {
   files: PropTypes.arrayOf(PropTypes.shape(filePropTypes)).isRequired,
-  fileTypeId: PropTypes.string.isRequired,
-  multiple: PropTypes.bool.isRequired,
+  fileTypeId: PropTypes.number.isRequired,
+  allowMultipleFiles: PropTypes.bool.isRequired,
   eventId: PropTypes.string.isRequired,
 };
--- a/indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
+++ b/indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
@@ -5,79 +5,129 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
-import React, {useContext, useRef} from 'react';
+import React, {useContext, useRef, useState} from 'react';
 import PropTypes from 'prop-types';
 import Dropzone from 'react-dropzone';
 import {Icon} from 'semantic-ui-react';
-import {FileManagerContext, filePropTypes, uploadFiles} from './util';
+import {FileManagerContext, filePropTypes, uploadFiles, deleteFile} from './util';
 import * as actions from './actions';
 
 import './FileManager.module.scss';
 
-export default function FileList({files, fileTypeId, multiple, eventId}) {
-  const dispatch = useContext(FileManagerContext);
+function FileAction({onClick, active, icon, className}) {
+  return (
+    <Icon
+      className={className}
+      name={active ? 'spinner' : icon}
+      loading={active}
+      onClick={() => {
+        if (!active) {
+          onClick();
+        }
+      }}
+    />
+  );
+}
+
+FileAction.propTypes = {
+  onClick: PropTypes.func.isRequired,
+  active: PropTypes.bool.isRequired,
+  icon: PropTypes.string.isRequired,
+  className: PropTypes.string.isRequired,
+};
+
+function FileEntry({eventId, fileTypeId, multiple, file: {uuid, filename, state, claimed}}) {
   const ref = useRef(null);
+  const dispatch = useContext(FileManagerContext);
+  const [activeButton, setActiveButton] = useState(null);
 
   return (
+    <>
+      <span styleName="file-state" className={state || ''}>
+        {filename}
+      </span>
+      <span>
+        {!state && multiple && (
+          <>
+            <FileAction
+              icon="exchange"
+              active={activeButton === 'replace'}
+              styleName="exchange-icon"
+              onClick={() => {
+                ref.current.open();
+              }}
+            />
+            <Dropzone
+              ref={ref}
+              onDrop={async acceptedFiles => {
+                setActiveButton('replace');
+                await uploadFiles(
+                  actions.markModified,
+                  fileTypeId,
+                  acceptedFiles,
+                  eventId,
+                  dispatch,
+                  uuid
+                );
+              }}
+            >
+              {({getRootProps, getInputProps}) => (
+                <span {...getRootProps()}>
+                  <input {...getInputProps()} />
+                </span>
+              )}
+            </Dropzone>
+          </>
+        )}
+        {state !== 'deleted' && state !== 'modified' ? (
+          <FileAction
+            styleName="delete-icon"
+            icon="trash"
+            active={activeButton === 'delete'}
+            onClick={async () => {
+              if (!claimed) {
+                // if the file is not part of the revision yet,
+                // we can safely delete it from the server
+                setActiveButton('delete');
+                await deleteFile(uuid);
+                setActiveButton(null);
+              }
+              dispatch(actions.markDeleted(fileTypeId, uuid));
+            }}
+          />
+        ) : (
+          <FileAction
+            icon="undo"
+            active={activeButton === 'undo'}
+            styleName="undo-icon"
+            onClick={async () => {
+              if (!claimed) {
+                setActiveButton('undo');
+                await deleteFile(uuid);
+                setActiveButton(null);
+              }
+              dispatch((state === 'deleted' ? actions.undelete : actions.revert)(fileTypeId, uuid));
+            }}
+          />
+        )}
+      </span>
+    </>
+  );
+}
+
+FileEntry.propTypes = {
+  eventId: PropTypes.string.isRequired,
+  fileTypeId: PropTypes.string.isRequired,
+  multiple: PropTypes.bool.isRequired,
+  file: PropTypes.shape(filePropTypes).isRequired,
+};
+
+export default function FileList({files, fileTypeId, multiple, eventId}) {
+  return (
     <ul styleName="file-list">
-      {files.map(({name, id, state}) => (
-        <li key={id} styleName="file-row">
-          <span styleName="file-state" className={state || ''}>
-            {name}
-          </span>
-          <span>
-            {!state && multiple && (
-              <>
-                <Icon
-                  styleName="exchange-icon"
-                  name="exchange"
-                  onClick={() => {
-                    ref.current.open();
-                  }}
-                />
-                <Dropzone
-                  ref={ref}
-                  onDrop={acceptedFiles =>
-                    uploadFiles(
-                      actions.markModified,
-                      fileTypeId,
-                      acceptedFiles,
-                      eventId,
-                      dispatch,
-                      id
-                    )
-                  }
-                >
-                  {({getRootProps, getInputProps}) => (
-                    <span {...getRootProps()}>
-                      <input {...getInputProps()} />
-                    </span>
-                  )}
-                </Dropzone>
-              </>
-            )}
-            {state !== 'deleted' && state !== 'modified' && (
-              <Icon
-                styleName="delete-icon"
-                onClick={() => dispatch(actions.markDeleted(fileTypeId, id))}
-                name="trash"
-              />
-            )}
-            {state === 'deleted' && (
-              <Icon
-                styleName="undo-icon"
-                onClick={() => dispatch(actions.undelete(fileTypeId, id))}
-                name="undo"
-              />
-            )}
-            {state === 'modified' && (
-              <Icon
-                styleName="undo-icon"
-                onClick={() => dispatch(actions.revert(fileTypeId, id))}
-                name="undo"
-              />
-            )}
-          </span>
+      {files.map(file => (
+        <li key={file.uuid} styleName="file-row">
+          <FileEntry fileTypeId={fileTypeId} eventId={eventId} multiple={multiple} file={file} />
         </li>
       ))}
     </ul>
@@ -85,7 +135,7 @@ export default function FileList({files, fileTypeId, multiple, eventId}) {
 }
 
 FileList.propTypes = {
-  files: filePropTypes.isRequired,
+  files: PropTypes.arrayOf(PropTypes.shape(filePropTypes)).isRequired,
   fileTypeId: PropTypes.string.isRequired,
   multiple: PropTypes.bool.isRequired,
   eventId: PropTypes.string.isRequired,
--- a/indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
+++ b/indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
@@ -54,39 +54,21 @@ export default function FileList({files, fileTypeId, multiple}) {
             {state !== 'deleted' && state !== 'modified' && (
               <Icon
                 styleName="delete-icon"
-                onClick={() =>
-                  dispatch({
-                    type: 'DELETE',
-                    fileTypeId,
-                    fileId: id,
-                  })
-                }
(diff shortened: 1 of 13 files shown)


indico/modules/events/editing/client/js/components/FileManager/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+    },
+    [id, multiple, files, dispatch]
+  );
+
+  const {getRootProps, getInputProps, isDragActive} = useDropzone({onDrop});
+
+  return (
+    <div {...getRootProps()}>
+      <input {...getInputProps()} />
+      <div className={`${styles['dropzone']} ${isDragActive ? 'active' : ''}`}>
```
ThiefMaster: `styles.dropzone`

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,60 @@
...
+    case 'MODIFY': {
+      const originalFile = _.find(fileType.files, {id: action.fileId});
+      fileType.files = fileType.files.map(file => {
+        if (file.id === action.fileId) {
+          return {state: 'modified', originalFile, ...action.file};
+        } else {
+          return file;
+        }
+      });
+      console.warn(state);
```
ThiefMaster: 👀

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,60 @@
...
+      action.files.forEach(file => {
+        file.state = 'added';
+      });
+      if (fileType.multiple) {
+        fileType.files = fileType.files.concat(action.files);
+      } else {
+        fileType.files = action.files;
+      }
+      return [...state];
+    case 'DELETE': {
```
ThiefMaster: I'd add an empty line at the end of a case

indico/modules/events/editing/templates/editing.html
```diff
@@ -0,0 +1,9 @@
+{% extends 'events/management/base.html' %}
+
+{% block title %}
+    Paper editing
+{% endblock %}
+
+{% block content %}
+    <div id="file-manager" data-event-id={{ event.id }}></div>
```
pferreir: Missing quotes.

indico/modules/events/editing/client/js/components/FileManager/util.js
```diff
@@ -0,0 +1,21 @@
...
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const FileManagerContext = React.createContext(null);
+
+export const filePropTypes = PropTypes.arrayOf(
```
panagiotappl: @pferreir Wouldn't it be better if we removed the `arrayOf` from this prop declaration and have `filePropTypes` represent just one file object?

pferreir: Yeah, that might make sense.

indico/modules/events/editing/client/js/components/FileManager/FileManager.module.scss
```diff
@@ -0,0 +1,100 @@
+@import 'base/palette';
```
mic4ael: missing LICENSE header

indico/modules/events/editing/client/js/components/FileManager/Uploads.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Icon, Progress, Segment} from 'semantic-ui-react';
+
+import './FileManager.module.scss';
+
+export default function Uploads({uploads}) {
+  const uploadList = _.sortBy(_.toPairs(uploads), x => x[0]);
```
mic4ael:
```suggestion
  const uploadList = _.sortBy(Object.entries(uploads), x => x[0]);
```

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,110 @@
...
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import _ from 'lodash';
+import {combineReducers} from 'redux';
+import * as actions from './actions';
+
+export default combineReducers({
+  fileTypes: (state = [], action) => {
+    const fileType = _.find(state, {id: action.fileTypeId});
```
mic4ael: We don't have to always use lodash if ES supports sth like that out of the box :P

```javascript
state.find(type => type.id === action.fileTypeId);
```

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,110 @@
...
+    switch (action.type) {
+      case actions.MARK_UPLOADED:
+        action.file.state = 'added';
```
mic4ael: I am wondering whether we should directly update the `file` object like that. It could probably have some side effects.

```javascript
const file = {...action.file, state: 'added'}l
...
    fileType.files = [...fileType.files, file];
} else {
    fileType.files = [file];

File: indico/modules/events/editing/client/js/components/FileManager/reducer.js [hunk 10]
```diff
@@ -0,0 +1,110 @@
...
+        if (fileType.multiple) {
+          fileType.files = fileType.files.concat(action.file);
+        } else {
+          fileType.files = [action.file];
+        }
+        return [...state];
+
+      case actions.MARK_DELETED: {
+        const file = _.find(fileType.files, {uuid: action.fileId});
```
mic4ael:
```suggestion
        const file = fileType.files.find(file => file.uuid === action.fileId);
```

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,110 @@
...
+
+import _ from 'lodash';
+import {combineReducers} from 'redux';
+import * as actions from './actions';
+
+export default combineReducers({
+  fileTypes: (state = [], action) => {
+    const fileType = _.find(state, {id: action.fileTypeId});
+    switch (action.type) {
+      case actions.MARK_UPLOADED:
```
mic4ael:
```suggestion
      case actions.MARK_UPLOADED: {
      ...
      }
```

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,110 @@
...
+        if (file.claimed) {
+          file.state = 'deleted';
+        } else {
+          _.pull(fileType.files, file);
+        }
+        return [...state];
+      }
+
+      case actions.MARK_MODIFIED: {
+        const currentFile = _.find(fileType.files, {uuid: action.fileId});
```
mic4ael: `Array.prototype.find` as in previous examples

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,110 @@
...
+            };
+          } else {
+            return file;
+          }
+        });
+        return [...state];
+      }
+
+      case actions.UNDELETE: {
+        const file = _.find(fileType.files, {uuid: action.fileId});
```
mic4ael: ditto

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,110 @@
...
+          progress: 0,
+          file,
+        }));
+        return {...state, [action.fileTypeId]: {...state[action.fileTypeId], ...newUploads}};
+      }
+
+      case actions.MARK_MODIFIED:
+      case actions.MARK_UPLOADED: {
+        const fileType = state[action.fileTypeId];
+        delete fileType[action.tmpFileId];
```
mic4ael: Hm, you are directly modifying the state here which is usually not a good idea.

indico/modules/events/editing/client/js/components/FileManager/util.js
```diff
@@ -0,0 +1,82 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import _ from 'lodash';
+import uploadFileURL from 'indico-url:event_editing.file_upload';
```
mic4ael: thise imports should be at the top of the file before all other imports

indico/modules/events/editing/client/js/components/FileManager/reducer.js
pferreir: Should we use `Array.prototype.find` if we're using lodash anyway and `_.find` is nicer (at leas in my opinion)?

indico/modules/events/editing/client/js/components/FileManager/reducer.js
pferreir: Just for the sake of style? Because the others are there for a reason.

indico/modules/events/editing/client/js/components/FileManager/reducer.js
pferreir: See my comment below. I don't agree that it's just 'out of the box'. `lodash` is more convenient IMHO.

indico/modules/events/editing/client/js/components/FileManager/reducer.js
ThiefMaster: How is it nicer? :p

indico/modules/events/editing/client/js/components/FileManager/reducer.js
mic4ael: It is a matter of preference. I e.g. prefer to use something that is provided directly by the engine.

indico/modules/events/editing/client/js/components/FileManager/reducer.js
ThiefMaster: I think curly braces only make sense for cases if you use `let` or `const` and thus need the scope

mic4ael: Not only, this will create another scope for all variables defined in that `case` (even though it might not be necessary in this case)

indico/modules/events/editing/client/js/components/FileManager/reducer.js
pferreir: It doesn't make much of a difference here. The old object is not used anywhere else. But this could be nicer:

```js
        return {
          ...state,
          [action.fileTypeId]: _.omit(fileType, action.tmpFileId),
        };
```

indico/modules/events/editing/client/js/components/FileManager/reducer.js
ThiefMaster: Oh, I just realized it searches by UUID. I agree with @pferreir that it's nicer than `[].find()`...

indico/modules/events/editing/client/js/components/FileManager/reducer.js
mic4ael: Yeah, maybe it is not being used right now, but who knows what the future holds :P

mic4ael: But yeah, using `omit` here is a much better way to do this

indico/modules/events/editing/client/js/components/FileManager/reducer.js
panagiotappl: I think it's better to leave them as they are, since there are no `let` or `const` there, for now.

indico/modules/events/editing/client/js/components/FileManager/reducer.js
pferreir: It's not very likely, since the action object is not reused AFAIK. But it's probably nicer to omit that line and do:
```js
        if (fileType.allowMultipleFiles) {
          fileType.files = fileType.files.concat({...action.file, state: 'added'});
        } else {
```
What do you guys think?

pferreir: @mic4ael your version is also fine, of course...

indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
```diff
@@ -0,0 +1,152 @@
...
+          <>
+            <FileAction
+              icon="exchange"
+              active={activeButton === 'replace'}
+              styleName="exchange-icon"
+              onClick={() => {
+                ref.current.open();
+              }}
+            />
+            <Dropzone
```
ThiefMaster: What about using the hook version?

indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
```diff
@@ -0,0 +1,152 @@
...
+          <FileAction
+            styleName="delete-icon"
+            icon="trash"
+            active={activeButton === 'delete'}
+            onClick={async () => {
+              if (!claimed) {
+                // if the file is not part of the revision yet,
+                // we can safely delete it from the server
+                setActiveButton('delete');
+                await deleteFile(uuid);
```
ThiefMaster: Deleting the file is more a "nice to do" thing, so we could immediately delete it without waiting for the request. Anyway, nothing wrong with doing it this way.

indico/modules/events/editing/client/js/components/FileManager/reducer.js
ThiefMaster: Throwing another version in the ring ;)

```js
fileTypes.files = [...fileType.files, {...action.file, state: 'added'}];
```

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,109 @@
...
+    const fileType = _.find(state, {id: action.fileTypeId});
+    switch (action.type) {
+      case actions.MARK_UPLOADED:
+        action.file.state = 'added';
+        if (fileType.allowMultipleFiles) {
+          fileType.files = fileType.files.concat(action.file);
+        } else {
+          fileType.files = [action.file];
+        }
+        return [...state];
```
ThiefMaster: do we need to clone the state here?

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,109 @@
...
+
+      case actions.MARK_DELETED: {
+        const file = _.find(fileType.files, {uuid: action.fileId});
+        if (file.claimed) {
+          file.state = 'deleted';
```
ThiefMaster: like above i'd prefer avoiding state mutations

indico/modules/events/editing/controllers/backend.py
```diff
@@ -36,3 +39,14 @@ def _process_args(self, editable_id):
 
     def _process(self):
         return EditableSchema().jsonify(self.editable)
+
+
+class RHEditing(RHManageEventBase):
```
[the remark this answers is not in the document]
panagiotappl: Does it really matter for this case? This shouldn't be merged in master anyway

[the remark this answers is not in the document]
ThiefMaster: no, if we remove it before merging the PR then it doesn't matter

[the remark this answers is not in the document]
panagiotappl: I think it's best to remove it after all the fixes are applied

indico/modules/events/editing/client/js/components/FileManager/util.js
```diff
@@ -0,0 +1,92 @@
...
+  claimed: PropTypes.bool,
+  state: PropTypes.oneOf(['added', 'modified', 'deleted']),
+};
+
+async function uploadFile(acceptedFile, urlParams, onUploadProgress) {
+  const formData = new FormData();
+  formData.append('file', acceptedFile);
+
+  try {
+    const {data} = await indicoAxios.post(uploadFileURL(urlParams), formData, {
```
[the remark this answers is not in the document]
pferreir: I see it the other way round: there's no need to "create" a variable to store the result if `data` is all we need.

indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
panagiotappl: In this case we needed to use the reference to know which dropzone to open whenever the user clicks the exchange button and it was too complicated with the hook version.

pferreir: We cannot, because we need a `ref` to the `Dropzone`, so that we can trigger the opening of the fiile selection dialog.

indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
pferreir: Yeah, I thought of that, but anyway, doesn't hurt to wait...

indico/modules/events/editing/client/js/components/FileManager/reducer.js
pferreir: It's just to signal that the state has changed, as a shallow comparison will not work.

indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
pferreir: (what @panagiotappl said :wink:)

indico/modules/events/editing/client/js/components/FileManager/reducer.js
ThiefMaster: that implies mutating state ;)

indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
ThiefMaster: The callable `open()` like in https://github.com/react-dropzone/react-dropzone/tree/master/examples/file-dialog doesn't work?

indico/modules/events/editing/client/js/components/FileManager/reducer.js
pferreir: Not sure I understand.

indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
pferreir: Oh, the docs don't mention that. Cool.

indico/modules/events/editing/client/js/components/FileManager/reducer.js
ThiefMaster: Is mutating state more acceptable in `useReducer` than with redux?

Because if you don't mutate the old state then by definition the new state you return is either the same (if nothing changed) or a new object (if something changed).

indico/modules/events/editing/client/js/components/FileManager/util.js
```diff
@@ -0,0 +1,92 @@
...
+  return Promise.all(
+    _.zip(acceptedFiles, tmpFileIds).map(async ([acceptedFile, tmpFileId]) => {
+      const uploadedFile = await uploadFile(acceptedFile, {confId: eventId}, e =>
+        dispatch(actions.progress(fileTypeId, tmpFileId, Math.floor((e.loaded / e.total) * 100)))
+      );
+      dispatch(
+        action(fileTypeId, fileId, tmpFileId, {
+          filename: uploadedFile.filename,
+          uuid: uploadedFile.uuid,
+          downloadURL: 'url2file',
```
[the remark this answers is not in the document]
panagiotappl: Thanks for noticing, this was leftover from "dummy data"

indico/modules/events/editing/client/js/components/FileManager/reducer.js
pferreir: Sure. But then instead of "cloning" the state we should build a new state object. Anyway, I agree that we should not mutate the state. We just did it like that initially because it becomes annoying once you have nested states.

indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
pferreir: Done.

indico/modules/events/editing/client/js/components/FileManager/util.js
```diff
@@ -0,0 +1,91 @@
...
+  } catch (e) {
+    handleAxiosError(e, false, true);
+  }
+}
+
+export function mapFileTypes(fileTypes, files) {
+  return fileTypes.map(fileType => {
+    return {
+      ...fileType,
+      files: files.map(f => ({...f, claimed: true})).filter(file => file.fileType === fileType.id),
```
ThiefMaster: Wouldn't it make more sense to `filter()` before the `map()`?

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,118 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import _ from 'lodash';
+import {combineReducers} from 'redux';
```
ThiefMaster: Interesting, I didn't know you could use this for useReducer-based reducers as well! But I guess it makes sense since this is just a functional util with no ties to the redux store etc.

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @param {Array} files - the array of files to look into
+ * @param {String} uuid - the uuid of the file to look for
+ * @param {Function} func - function used to "reduce" the file in question
+ */
+function replaceFile(files, uuid, func) {
+  return files
+    .map(file => {
+      return file.uuid === uuid ? func(file) : file;
+    })
```
ThiefMaster: `.map(file => (file.uuid === uuid ? func(file) : file))`?

indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
```diff
@@ -0,0 +1,144 @@
...
+        {filename}
+      </span>
+      <span>
+        {!state && allowMultipleFiles && (
+          <>
+            <FileAction
+              icon="exchange"
+              active={activeButton === 'replace'}
+              styleName="exchange-icon"
+              onClick={() => {
```
ThiefMaster: `onClick={open}`? `FileAction` never passes arguments to it anyway.

indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
```diff
@@ -0,0 +1,144 @@
...
+  allowMultipleFiles,
+  file: {uuid, filename, state, claimed},
+}) {
+  const dispatch = useContext(FileManagerContext);
+  const [activeButton, setActiveButton] = useState(null);
+
+  const onDrop = useCallback(
+    async acceptedFiles => {
+      setActiveButton('replace');
+      await uploadFiles(actions.markModified, fileTypeId, acceptedFiles, eventId, dispatch, uuid);
```
ThiefMaster: We don't need a `setActiveButton(null)` afterwards?

pferreir: No, actually React will complain about that, since the component will be unmounted by then.

ThiefMaster: might be worth adding a comment mentioning that the component will be unmounted afterwards and thus no update is needed

indico/modules/events/editing/client/js/components/FileManager/util.js
pferreir: Indeed, thanks!

indico/modules/events/editing/client/js/components/FileManager/reducer.js
pferreir: Yeah, it's just a normal function.

indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
pferreir: Good idea.

indico/modules/events/editing/client/js/components/FileManager/util.js
```diff
@@ -0,0 +1,88 @@
...
+  formData.append('file', file);
+
+  try {
+    const {data} = await indicoAxios.post(url, formData, {
+      headers: {'content-type': 'multipart/form-data'},
+      onUploadProgress,
+    });
+    return data;
+  } catch (e) {
+    handleAxiosError(e, false, true);
```
ThiefMaster: what happens to the calling code if it fails? does it go kaboom or just hide the failed upload?

indico/modules/events/editing/controllers/backend.py
```diff
@@ -250,3 +250,4 @@ def _process_PATCH(self, text, internal):
     def _process_DELETE(self):
         delete_revision_comment(self.comment)
         return '', 204
+        return EditableSchema().jsonify(self.editable)
```
ThiefMaster: i don't think this should be there

indico/modules/events/editing/client/js/components/FileManager/util.js
pferreir: I believe it just kabooms 💣

pferreir: We would have to add an additional action to "finish" the upload once there is an error. I think it can be done in a future task as it's not urgent right now.

ThiefMaster: yeah, agreed

indico/modules/events/editing/controllers/frontend.py
```diff
@@ -6,3 +6,16 @@
...
 
 from __future__ import unicode_literals
+
+from indico.modules.events.editing.controllers.base import RHContributionEditableBase
+from indico.modules.events.editing.views import WPEditingBase
+
+
+class RHEditableTimeline(RHContributionEditableBase):
+    def _process(self):
+        return WPEditingBase.render_template(
```
mic4ael: Was black used here? :D

indico/modules/events/editing/views.py
```diff
@@ -6,3 +6,10 @@
 # LICENSE file for more details.
 
 from __future__ import unicode_literals
+
+from indico.modules.events.management.views import WPEventManagement
+
+
+class WPEditingBase(WPEventManagement):
```
ThiefMaster: `WPEditingBase` is a bit strange since this is not a base class

indico/modules/events/editing/controllers/frontend.py
pferreir: Nope, that was actually me 🤣

PR #4138, opened 2019-11-12
Add models related to editing revisions

### LINKED ISSUE #4111: Paper Editing: create DB model for Reviewable
Right now we have foreseen two types of `Reviewable`:
 * Conference Papers
 * Slides
but we should keep in mind that others may need to be added in the future.

`Reviewable`s should have their separate reviewing timelines. Slide and paper reviewing are different things which are reviewed in a different way.

### LINKED ISSUE #4112: Paper Editing: Implement DB models for Reviews/Revisions
Our current structure in `indico.modules.events.papers.models` looks like this:

image

We could probably base ourselves on this, while taking into account that a `Reviewable` will be the target of the review, not a `Contribution`. We don't need to worry about ratings and questions for now.

ThiefMaster:
closes #4111 
closes #4112


indico/modules/events/editing/models/editable.py
```diff
@@ -0,0 +1,71 @@
...
+            lazy='dynamic'
+        )
+    )
+    published_revision = db.relationship(
+        'EditingRevision',
+        foreign_keys=published_revision_id,
+        lazy=True,
+    )
+
+    # relationship backrefs:
```
pferreir: Have you guys not included the relationship for `contribution` on purpose?

ThiefMaster: nope, just forgot adding it

PR #4144, opened 2019-11-12
Paper editing: basic timeline

### LINKED ISSUE #4139: Paper Reviewing: basic timeline
Create a basic timeline based on the work done in #4107.

image

pferreir:
**TODO:**
 * [x] List of revisions (no files, no forms)

Closes #4139


indico/modules/events/client/js/reviewing/components/TimelineItem.jsx
```diff
@@ -0,0 +1,85 @@
...
+                    </div>
+                    <div dangerouslySetInnerHTML={{__html: revision.commentHtml}} />
+                  </>
+                )}
+              </div>
+            )}
+          </div>
+        </div>
+      </div>
+      {(visible || isLastRevision) && revision.items.length !== 0 && (
```
meluru: Are you sure this `isLastRevision` is needed here? You already set `visible` default state based on `isLastRevision` and changing this state is not allowed for last revision

mic4ael: Yeah, I already changed that earlier today ;)

indico/modules/events/client/js/reviewing/components/JudgmentModal.module.scss
```diff
@@ -0,0 +1,12 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+:global(.ui.modal) {
+    :global(.disabled.field) {
+        background: none;
```
ThiefMaster: why do we need this?

indico/modules/events/editing/client/js/components/Timeline/reducer.js
```diff
@@ -0,0 +1,19 @@
...
+// LICENSE file for more details.
+
+import {SET_LOADING, SET_DETAILS} from './actions';
+
+export default function reducer(state, action) {
+  switch (action.type) {
+    case SET_LOADING:
+      return {...state, isLoading: action.loading};
+    case SET_DETAILS:
+      return {...state, details: action.details};
```
ThiefMaster: shouldn't this imply changing `isLoading` to `false`?

mic4ael: Yeah, but in case of an error we would still have to dispatch a separate action to disable loading, thus we just send an action to set `loading` to `false` in the `finally` block.
```javascript

      try {
        const {data} = await indicoAxios.get(
          editableDetailsURL({confId: eventId, contrib_id: contributionId, type})
        );
        dispatch(actions.setDetails(camelizeKeys(data)));
      } catch (error) {
        handleAxiosError(error, false, true);
      } finally {
        dispatch(actions.setLoading(false));
      }
```

indico/modules/events/client/js/reviewing/components/JudgmentModal.module.scss
mic4ael: Because `disabled` fields have some color that really stands out when used in conjunction with out `i-timeline` styles

indico/modules/events/editing/client/js/components/Timeline/reducer.js
ThiefMaster: sure, just wondered if it'd be better to only dispatch the explicit action in case of an error. anyway, i don't think there's a problem with the separate dispatches

indico/modules/events/editing/views.py
```diff
@@ -12,4 +12,5 @@
 
 class WPEditing(WPEventManagement):
     template_prefix = 'events/editing/'
-    bundles = ('module_events.editing.js', 'module_events.editing.css')
+    # remove mathjax later on from here (it is just here to make the timeline work)
```
ThiefMaster: `# TODO: ...` (otherwise we'll forget it for sure - i usually search only for "TODO")

PR #4152, opened 2019-11-13
Add Editing backend

### LINKED ISSUE #4131: Create backend to receive new revisions.
The endpoint should receive:
 * the UUIDs of files to be added/replaced/deleted;
 * the target state ("Assigned" or "Requires confirmation". We should check its validity, of course);

This action can be triggered:
 * by the author/submitter
 * by an editor (it will then trigger to "requires confirmation")

ThiefMaster:
Also changed the API endpoint to get the timeline to use the contrib id and type instead of the editable id.

closes #4131

### DIFF
--- a/indico/modules/events/editing/blueprint.py
+++ b/indico/modules/events/editing/blueprint.py
@@ -29,3 +29,5 @@ _bp.add_url_rule('/api/contributions/<int:contrib_id>/editing/<any(paper):type>/
                  'api_replace_revision', backend.RHReplaceRevision, methods=('POST',))
 _bp.add_url_rule('/api/contributions/<int:contrib_id>/editing/<any(paper):type>/<int:revision_id>/new',
                  'api_create_submitter_revision', backend.RHCreateSubmitterRevision, methods=('POST',))
+_bp.add_url_rule('/api/contributions/<int:contrib_id>/editing/<any(paper):type>/<int:revision_id>/review',
+                 'api_undo_review', backend.RHUndoReview, methods=('DELETE',))
--- a/indico/modules/events/editing/blueprint.py
+++ b/indico/modules/events/editing/blueprint.py
@@ -16,4 +16,7 @@ _bp = IndicoBlueprint('event_editing', __name__, url_prefix='/event/<confId>', t
 
 _bp.add_url_rule('/editing/api/file-types', 'api_file_types', backend.RHEditingFileTypes)
 _bp.add_url_rule('/editing/api/tags', 'api_tags', backend.RHEditingTags)
-_bp.add_url_rule('/editing/api/editable/<int:editable_id>', 'api_editable', backend.RHEditable)
+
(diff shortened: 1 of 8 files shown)


indico/modules/events/editing/blueprint.py
```diff
@@ -16,4 +16,7 @@
 
 _bp.add_url_rule('/editing/api/file-types', 'api_file_types', backend.RHEditingFileTypes)
 _bp.add_url_rule('/editing/api/tags', 'api_tags', backend.RHEditingTags)
-_bp.add_url_rule('/editing/api/editable/<int:editable_id>', 'api_editable', backend.RHEditable)
+
+_bp.add_url_rule('/api/contributions/<int:contrib_id>/editing/<type>', 'api_editable', backend.RHEditable)
+_bp.add_url_rule('/api/contributions/<int:contrib_id>/editing/<type>', 'api_create_editable',
+                 backend.RHCreateEditable, methods=('PUT',))
```
mic4ael: Why `PUT` here? Wouldn't `POST` be more suitable?

ThiefMaster: I think both are valid here; with POST I'd rather add a `/create` at the end since especially once there are the other urls, just POSTing to that url won't be super clear.

indico/modules/events/editing/controllers/backend.py
```diff
@@ -7,32 +7,193 @@
...
+    def _check_access(self):
+        RHContributionEditableBase._check_access(self)
+        if not self._user_is_authorized_submitter():
+            # XXX: should event managers be able to submit on behalf of the user?
+            raise Forbidden
+        # TODO: check if submitting papers for editing is allowed in the event
+
+    def _process(self):
+        if self.editable:
+            raise UserValueError('Editable already exists')
```
mic4ael: Missing i18n?

ThiefMaster: If an error is not going to show up to users during normal usage there's not much benefit in translating it. But in theory this one could happen if you have multiple tabs open and already submitted your editable in another one, so can't hurt to add i18n to it..

indico/modules/events/editing/operations.py
```diff
@@ -0,0 +1,200 @@
...
+        EditingReviewAction.request_update: FinalRevisionState.needs_submitter_changes,
+    }[action]
+
+    db.session.flush()
+    if action == EditingReviewAction.accept:
+        publish_editable_revision(revision)
+    elif action == EditingReviewAction.update:
+        new_revision = EditingRevision(submitter=editor,
+                                       initial_state=InitialRevisionState.needs_submitter_confirmation,
+                                       files=_make_editable_files(revision.editable, files))
```
mic4ael: `_make_editable_files` will fail with the default argument for `files`

ThiefMaster: that will never happen because `files` is always specified for the `update` action. anyway, updated the function to handle it

mic4ael: Yeah, but then I guess it shouldn't be a keyword argument?

PR #4160, opened 2019-11-19
Add FileDisplay component

### LINKED ISSUE #4143: Paper Editing: create FileDisplay component
Create a component that only displays submitted files.

panagiotappl:
closes #4143 
image

### DIFF
--- a/indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
+++ b/indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
@@ -25,3 +25,8 @@
         font-weight: bold;
     }
 }
+
+.no-files {
+    font-style: italic;
+    color: $gray;
+}
--- a/indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
+++ b/indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
@@ -7,6 +7,11 @@
 
 @import 'base/palette';
 
+.file-display-wrapper {
+    display: flex;
+    flex-direction: column;
+}
+
 .file-display {
     display: flex;
     flex-wrap: wrap;
--- a/indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
+++ b/indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
@@ -1,0 +1,27 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+@import 'base/palette';
+
+.file-display {
+    display: flex;
+
+    .file-type-display {
+        padding: 1em;
+        margin: 1em;
+        width: 33%;
+    }
+}
+
+.file-list-display {
+    list-style-type: none;
+    padding: 0;
+
+    .file-name {
+        font-weight: bold;
+    }
+}
--- a/indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
+++ b/indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
@@ -9,11 +9,11 @@
 
 .file-display {
     display: flex;
+    flex-wrap: wrap;
 
     .file-type-display {
         padding: 1em;
-        margin: 1em;
-        width: 33%;
+        flex: 0 0 32%;
     }
 }
 
--- a/indico/modules/events/editing/client/js/components/FileDisplay/index.jsx
+++ b/indico/modules/events/editing/client/js/components/FileDisplay/index.jsx
@@ -1,0 +1,57 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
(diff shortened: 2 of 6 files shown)


indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
```diff
@@ -0,0 +1,20 @@
+@import 'base/palette';
+
+.file-display {
+    display: flex;
+
+    .file-type-display {
+        padding: 1em;
+        margin: 1em;
+        width: 33%;
```
mic4ael: I think you can achieve the same with `flex: 1` which will be probably better in case there are more than 3 columns

indico/modules/events/editing/client/js/components/FileDisplay/index.jsx
```diff
@@ -0,0 +1,55 @@
+import React, {useMemo} from 'react';
```
mic4ael: missing headers

indico/modules/events/editing/client/js/components/FileDisplay/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+  files: PropTypes.arrayOf(PropTypes.shape(filePropTypes)).isRequired,
+};
+
+function FileTypeDisplay({fileType}) {
+  return (
+    <div styleName="file-type-display">
+      <h3>{fileType.name}</h3>
+      <FileListDisplay
+        files={fileType.files}
+        fileTypeId={fileType.id}
```
mic4ael: These two props are not being used at all

indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
panagiotappl: I think the initial idea behind this (we use the same in FileManager) is to move to another row whenever we have more than 3 items and not try to squeeze in everything in the same row.

ThiefMaster: that sounds a perfect use case for flexbox features

indico/modules/events/editing/client/js/components/FileDisplay/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+    </div>
+  );
+}
+
+FileTypeDisplay.propTypes = {
+  fileType: PropTypes.shape(fileTypePropTypes).isRequired,
+};
+
+export default function FileDisplay({fileTypes, files}) {
+  const _fileTypes = useMemo(() => mapFileTypes(fileTypes, files), [fileTypes, files]);
```
mic4ael: I am wondering whether the benefit of `useMemo` is enough to justify its usage here. This component probably won't rerender so often, right? It doesn't hurt to use it but I am just thinking whether it is so critical here

indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
mic4ael: Yep, that is why I suggested `flex` attribute ;)

meluru:
Can you provide a screenshot? :)

indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
panagiotappl: But you would still need to use a percentage to specify how many items you want per line even with flexbox (ex: `flex: 0 0 33%;`), right?

mic4ael: Yeah, and probably `flex-wrap` as well

ThiefMaster:
I rebased it to the latest master, but I think it looks a bit strange if there are no files for a given type:

image

Maybe we can somehow indicatean empty file type ?

indico/modules/events/editing/client/js/components/FileDisplay/index.jsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <ul styleName="file-list-display">
+      {files.map(({filename, uuid}) => (
+        <li key={uuid}>
+          <span styleName="file-name">
+            <a>{filename}</a>
+          </span>
+        </li>
+      ))}
+      {!files.length && <span styleName="no-files">No files uploaded</span>}
```
mic4ael: Missing i18n

indico/modules/events/editing/client/js/components/FileDisplay/FileDisplay.module.scss
```diff
@@ -0,0 +1,25 @@
+@import 'base/palette';
```
mic4ael: missing header

indico/modules/events/editing/client/js/components/FileDisplay/index.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+FileTypeDisplay.propTypes = {
+  fileType: PropTypes.shape(fileTypePropTypes).isRequired,
+};
+
+export default function FileDisplay({fileTypes, files}) {
+  const _fileTypes = mapFileTypes(fileTypes, files);
+  return (
+    <div styleName="file-display">
+      {_fileTypes.map(fileType => (
```
mic4ael:
```suggestion
      {mapFileTypes(fileTypes, files).map(fileType => (
```

PR #4164, opened 2019-11-19
Accommodation: Allow single arrival/departure date

bpedersen2:
Allow to restrict the date range to a single day for
arrival/departue  in the setup form.


ThiefMaster:
Does it make sense to select accommodation at all in this case? In case of payment, the price would be zero (0 nights), so usually people are supposed to select "no accommodation" in such a case.

bpedersen2:
> Does it make sense to select accommodation at all in this case? In case payment is used, the price would be zero (0 nights), so usually people are supposed to select "no accommodation" in such a case.

There you  misunderstand the change, it is about allowing only a single arrival and/or departure date to  be selected.  Currently you always need to allow at least 2 days for arrival and departure.

ThiefMaster:
Make sense. Thanks!

PR #4165, opened 2019-11-19
Display only public contrib fields in pdf exports

meluru:
Fix contrib fields display in pdf exports

### DIFF
--- a/indico/legacy/pdfinterface/latex.py
+++ b/indico/legacy/pdfinterface/latex.py
@@ -457,7 +457,6 @@ class ContributionBook(PDFLaTeXBase):
             'event': event,
             'tz': timezone(tz or event.timezone),
             'url': event.url,
-            'fields': [f for f in event.contribution_fields if f.is_active],
             'sorted_by': sort_by,
             'user': user,
             'boa_text': boa_settings.get(event, 'extra_text'),
--- a/indico/legacy/pdfinterface/latex_templates/inc/contribution.tex
+++ b/indico/legacy/pdfinterface/latex_templates/inc/contribution.tex
@@ -48,11 +48,12 @@
     \end{addmargin}
 
 
-    \JINJA{for field in contrib.event.contribution_fields if field.is_active}
+    \JINJA{for field_value in contrib.field_values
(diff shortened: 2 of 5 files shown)


indico/legacy/pdfinterface/latex.py
```diff
@@ -457,7 +457,6 @@ def __init__(self, event, user, contribs=None, tz=None, sort_by=""):
             'event': event,
             'tz': timezone(tz or event.timezone),
             'url': event.url,
-            'fields': [f for f in event.contribution_fields if f.is_active],
```
ThiefMaster: I guess this was unused?

indico/legacy/pdfinterface/latex_templates/inc/contribution.tex
```diff
@@ -180,16 +180,18 @@
         \setlength{\leftskip}{0pt}
     \JINJA{endif}
 
-    \JINJA{for field_value in contrib.field_values}
-        \textbf{\VAR{ field_value.contribution_field.title}}:
+    \JINJA{for field in contrib.event.contribution_fields if field.is_active and field.is_public}
```
ThiefMaster: I think this change results in fields that had no value (e.g. because the field was created after the contribution) showing up

indico/legacy/pdfinterface/latex_templates/inc/contribution.tex
```diff
@@ -180,16 +180,18 @@
...
+        \textbf{\VAR{field.title}}:
 
         %% Markdown content
         \rmfamily {
             \allsectionsfont{\rmfamily}
             \sectionfont{\normalsize\rmfamily}
             \subsectionfont{\small\rmfamily}
             \small
-            \VAR{field_value.friendly_data|markdown}
+            \VAR{contrib.get_field_value(field.id) | markdown}
```
ThiefMaster: `get_field_value` iterates over `field_values` internally, so this is much less efficient than the old code

indico/legacy/pdfinterface/latex.py
meluru: I will double check, but seems like unusued

Issue #2715, opened 2016-11-16
Registration Form: allow users to cancel their registration


pferreir:
There's already backend code for this (`Registration.withdrawn`), we have to make it available from the interface. It should be configurable.

SebastianLopienski:
It would be also good to allow event managers to withdraw an application/registration (e.g. one the person withdraws by email to event organizers)

brabemi:
Hi, in my [fork](https://github.com/brabemi/indico/tree/registration_withdraw) i added withdraw button with confirmation popup into registration detail view. Behavior is the same as for modify button.

I will think about how to add withdraw action for event managers

![02](https://user-images.githubusercontent.com/6898184/57967699-5917d200-7962-11e9-9cb0-77cc30dedda6.png)

bpedersen2:
@brabemi : Would you mind to create a pull request?

PR #4166, opened 2019-11-19
Withdraw popup in registration summary

brabemi:
Add withdraw button with confirmation popup into registration detail view. Behavior is the same as for the modify button.

closes: #2715 
CC: @ThiefMaster @bpedersen2 

indico_02

### DIFF
--- a/indico/modules/events/registration/controllers/display.py
+++ b/indico/modules/events/registration/controllers/display.py
@@ -308,6 +308,24 @@ class RHRegistrationDisplayEdit(RegistrationEditMixin, RHRegistrationFormRegistr
         return url_for('.display_regform', self.registration.locator.registrant)
 
 
+class RHRegistrationWithdraw(RHRegistrationFormRegistrationBase):
+    """Withdraw a registration"""
+
+    def _process(self):
+        if not self.registration.can_be_modified:
+            if not self.registration.registration_form.is_modification_open:
+                flash(_("The modification/withdrawal period is over"), "warning")
+            elif self.registration.registration_form.modification_mode.name == 'allowed_until_payment':
+                flash(_("Withdrawal is not allowed after payment"), "warning")
+            else:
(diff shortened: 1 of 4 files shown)


indico/modules/events/registration/controllers/display.py
```diff
@@ -308,6 +308,24 @@ def success_url(self):
         return url_for('.display_regform', self.registration.locator.registrant)
 
 
+class RHRegistrationWithdraw(RHRegistrationFormRegistrationBase):
+    """Withdraw a registration"""
+
+    def _process(self):
+        if not self.registration.can_be_modified:
+            if not self.registration.registration_form.is_modification_open:
```
ThiefMaster: Should we check for the modification period here at all? I think withdrawal should always be possible, especially if no payment is involved. After paying it does make sense that you need to contact the organizer though (regardless of the modification setting) - and they should then have a button in the management view to mark the registration as withdrawn.

indico/modules/events/registration/controllers/display.py
```diff
@@ -308,6 +308,24 @@ def success_url(self):
...
+                flash(_("The modification/withdrawal period is over"), "warning")
+            elif self.registration.registration_form.modification_mode.name == 'allowed_until_payment':
+                flash(_("Withdrawal is not allowed after payment"), "warning")
+            else:
+                flash(_("The registration cannot be modified/withdrawn"), "warning")
+            return redirect(url_for('.display_regform', self.registration.locator.registrant))
+
+        self.registration.state = RegistrationState.withdrawn
```
ThiefMaster: This should use [`update_state()`](https://github.com/indico/indico/blob/9955b2ab920ac8af753937e79a038b78754ba4d7/indico/modules/events/registration/models/registrations.py#L458) (with a new kwarg for withdrawal) so the `registration_state_updated` signal gets triggered.

PR #4168, opened 2019-11-20
Jest/Enzyme unit tests for React components

pferreir:
This PR adds unit tests for React components based on [`jest`](https://jestjs.io/docs/en/getting-started.html) and [`enzyme`](https://airbnb.io/enzyme/).

 * The `axios` and `react` modules are mocked, which makes it possible to simulate backend requests and tap into the inner workings of components (e.g. intercepting `useReducer(...)`'s actions);
 * There are currently only two simple tests;
 * CSS files imported from JavaScript are just ignored;

I have moved parts of our webpack config to Babel's, which required moving from `.babelrc` to `babel.config.js`, as it was initially needed. Even if it's not strictly needed anymore, I believe it's a better place to have it anyway.


__mocks__/axios.js
```diff
@@ -0,0 +1,9 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import mockAxios from 'jest-mock-axios';
+export default mockAxios;
```
ThiefMaster: `export default from 'jest-mock-axios';`

__mocks__/react.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const PatchedReact = {
+  ...React,
+  useReducer,
+  resetMocks,
+  mockDispatches,
+};
+
+export {PatchedReact as default};
```
ThiefMaster: `export default PatchedReact;`?

__mocks__/react.js
```diff
@@ -0,0 +1,42 @@
...
+ * a way to tap into some of the internal methods (e.g. dispatch)
+ */
+
+import React from 'react';
+
+const mockDispatches = [];
+
+export * from 'react';
+
+export function useReducer(reducer, state) {
```
ThiefMaster: `initialState`? (and then s/st/state below)

indico/modules/events/editing/client/js/components/FileManager/util.js
```diff
@@ -62,7 +62,7 @@ export function uploadFiles(action, fileTypeId, acceptedFiles, uploadURL, dispat
         action(fileTypeId, fileId, tmpFileId, {
           filename: uploadedFile.filename,
           uuid: uploadedFile.uuid,
-          downloadURL: null,
+          downloadURL: uploadedFile.downloadURL,
```
ThiefMaster: Is this intended? We only have a download URL for files that are already attached to a revision (we could probably create a blob url from the locally-available file data until then).

package.json
```diff
@@ -154,5 +162,23 @@
   "bugs": {
     "url": "https://github.com/indico/indico/issues"
   },
-  "homepage": "https://github.com/indico/indico#readme"
+  "homepage": "https://github.com/indico/indico#readme",
+  "jest": {
+    "moduleNameMapper": {
+      "^.+\\.(s?css)$": "identity-obj-proxy",
+      "^indico/modules/([\\w]+)/(.*)(.jsx?|)$": "<rootDir>/indico/modules/$1/client/js/$2$3",
```
ThiefMaster: Does the `(.jsx?|)` part make sense? No matter what you have there, it will either go in `$3` or `$2`

ThiefMaster:
eslint is currently broken when there's no `webpack-build-config.js` - which is usually the case unless you are actively building or in `--watch` mode

indico/modules/events/editing/client/js/components/FileManager/util.js
pferreir: I added it so that we can at least mock the response from the server and it will work. If there isn't a `downloadURL` we'll get `undefined`, which I guess is OK here.

pferreir: You're right. We will only get `downloadURL` for claimed files, which will never come from this handler.

__mocks__/axios.js
pferreir: I don't think you can do that. At least ESLint complains about it.

ThiefMaster: `export {default} from '...';` works for sure. the syntax from my initial comment may still be in a proposal stage and not supported

https://stackoverflow.com/questions/39999282/re-export-default-in-es-6-modules

pferreir: It's an experimental feature.

pferreir:
> eslint is currently broken when there's no `webpack-build-config.js` - which is actually always the case unless you are actively building (or in `--watch` mode)

Thanks! It should be fine now.

PR #4169, opened 2019-11-21
File manager download zip

### LINKED ISSUE #4163: Add "download" button in FileManager/display which downloads zip file.
image

mic4ael:
closes #4163

### DIFF
--- a/indico/modules/events/editing/blueprint.py
+++ b/indico/modules/events/editing/blueprint.py
@@ -17,6 +17,9 @@ _bp = IndicoBlueprint('event_editing', __name__, url_prefix='/event/<confId>', t
 _bp.add_url_rule('/editing/api/file-types', 'api_file_types', backend.RHEditingFileTypes)
 _bp.add_url_rule('/editing/api/tags', 'api_tags', backend.RHEditingTags)
 
+
+_bp.add_url_rule('/contributions/<int:contrib_id>/editing/<any(paper):type>/<int:revision_id>/files.zip',
+                 'revision_files_export', backend.RHExportRevisionFiles)
 _bp.add_url_rule('/api/contributions/<int:contrib_id>/editing/<any(paper):type>/upload', 'api_upload',
                  backend.RHEditingUploadFile, methods=('POST',))
 _bp.add_url_rule('/api/contributions/<int:contrib_id>/editing/<any(paper):type>',
--- a/indico/modules/events/editing/client/js/components/FileManager/FileManager.module.scss
+++ b/indico/modules/events/editing/client/js/components/FileManager/FileManager.module.scss
@@ -24,15 +24,25 @@
     }
 }
 
+.file-manager-wrapper {
+    display: flex;
+    flex-direction: column;
+}
+
 .file-manager {
     display: flex;
+    flex-wrap: wrap;
 
     .file-type {
+        &:not(:nth-child(3)) {
+            margin-right: 1em;
+        }
+
+        margin-bottom: 1em;
         border: 1px solid $gray;
         border-radius: 3px;
         padding: 1em;
-        margin: 1em;
-        width: 33%;
+        flex: 0 0 32%;
     }
 }
 
@@ -78,6 +88,7 @@
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
+    max-width: 80%;
 
     &:global(.added) {
         color: $green;
--- a/indico/modules/events/editing/client/js/components/FileManager/index.jsx
+++ b/indico/modules/events/editing/client/js/components/FileManager/index.jsx
@@ -9,8 +9,11 @@ import _ from 'lodash';
 import React, {useCallback, useReducer, useContext, useRef, useMemo} from 'react';
 import PropTypes from 'prop-types';
 import {useDropzone} from 'react-dropzone';
-import {Icon} from 'semantic-ui-react';
-import {FileManagerContext, filePropTypes, uploadFiles, deleteFile, mapFileTypes} from './util';
+import {Button, Icon} from 'semantic-ui-react';
+
+import {Translate} from 'indico/react/i18n';
+
+import {deleteFile, FileManagerContext, filePropTypes, mapFileTypes, uploadFiles} from './util';
 import FileList from './FileList';
 import Uploads from './Uploads';
 import reducer from './reducer';
@@ -101,7 +104,7 @@ FileType.defaultProps = {
   uploads: {},
 };
 
-export default function FileManager({uploadURL, fileTypes, files}) {
+export default function FileManager({downloadURL, uploadURL, fileTypes, files}) {
   const _fileTypes = useMemo(() => mapFileTypes(fileTypes, files), [fileTypes, files]);
   const [state, dispatch] = useReducer(reducer, {
     fileTypes: _fileTypes,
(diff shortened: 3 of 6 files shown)


indico/modules/events/editing/client/js/components/FileManager/FileManager.module.scss
```diff
@@ -24,15 +24,25 @@
...
+            margin-right: 1em;
+        }
+
+        margin-bottom: 1em;
         border: 1px solid $gray;
         border-radius: 3px;
         padding: 1em;
-        margin: 1em;
-        width: 33%;
+        flex: 0 0 32%;
```
panagiotappl: Why do we use 32% instead of 33% here?

mic4ael: Because of the margin. With 33% there will be usually two items per row.

indico/modules/events/editing/blueprint.py
```diff
@@ -17,6 +17,9 @@
 _bp.add_url_rule('/editing/api/file-types', 'api_file_types', backend.RHEditingFileTypes)
 _bp.add_url_rule('/editing/api/tags', 'api_tags', backend.RHEditingTags)
 
+
+_bp.add_url_rule('/contributions/<int:contrib_id>/editing/<any(paper):type>/<int:revision_id>/revision-files.zip',
```
ThiefMaster: `.../files.zip`?

indico/modules/events/editing/client/js/components/FileManager/index.jsx
```diff
@@ -9,8 +9,11 @@ import _ from 'lodash';
 import React, {useCallback, useReducer, useContext, useRef, useMemo} from 'react';
 import PropTypes from 'prop-types';
 import {useDropzone} from 'react-dropzone';
-import {Icon} from 'semantic-ui-react';
-import {FileManagerContext, filePropTypes, uploadFiles, deleteFile, mapFileTypes} from './util';
+import {Button, Icon} from 'semantic-ui-react';
+
+import {Param, Translate} from 'indico/react/i18n';
+
```
ThiefMaster: 👾

indico/modules/events/editing/client/js/components/FileManager/index.jsx
```diff
@@ -101,30 +104,42 @@ FileType.defaultProps = {
...
+              uploads={state.uploads[fileType.id]}
+            />
+          ))}
+        </FileManagerContext.Provider>
+      </div>
+      {downloadURL && files.length !== 0 && (
+        <div>
+          <Button as="a" href={downloadURL} floated="right" icon primary>
+            <Translate>
+              <Param name="icon" value={<Icon name="download" />} /> Download ZIP
```
ThiefMaster: Just move the icon outside the Translate..

indico/modules/events/editing/controllers/backend.py
```diff
@@ -250,3 +259,27 @@ def _process_PATCH(self, text, internal):
...
+        temp_file = NamedTemporaryFile(suffix='indico.tmp', dir=config.TEMP_DIR)
+        with ZipFile(temp_file.name, 'w', allowZip64=True) as zip_handler:
+            for revision_file in self.revision.files:
+                file = revision_file.file
+                file_type = revision_file.file_type
+
+                with file.storage.get_local_path(file.storage_file_id) as filepath:
+                    filename = secure_filename(file.filename, 'file-{}'.format(file.id))
+                    folder_name = secure_filename(file_type.name, 'file-type-{}'.format(file_type.id))
+                    zip_handler.write(filepath.encode('utf-8'), os.path.join(folder_name, filename))
```
ThiefMaster: i don't think we need/want `secure_filename` for the folder / filename used in the second arg. Just make sure that putting `../` and similar stuff there doesn't break anything (`secure_filename` strips much more, like any whitespace, non-ascii, etc).

indico/modules/events/editing/controllers/backend.py
```diff
@@ -250,3 +259,27 @@ def _process_PATCH(self, text, internal):
     def _process_DELETE(self):
         delete_revision_comment(self.comment)
         return '', 204
+
+
+class RHRevisionFilesExportZIP(RHContributionEditableRevisionBase):
```
ThiefMaster: `RHExportRevisionFiles`?

indico/modules/events/editing/controllers/backend.py
```diff
@@ -250,3 +259,27 @@ def _process_PATCH(self, text, internal):
...
+    def _check_revision_access(self):
+        return self._user_is_authorized_submitter() or self._user_is_authorized_editor()
+
+    def _process(self):
+        if not self.revision.files:
+            raise NotFound
```
ThiefMaster: When would you have a revision without files? Anyway, not sure if we even need this check, you'd just get an empty zip file in this case.

indico/modules/events/editing/controllers/backend.py
```diff
@@ -250,3 +259,27 @@ def _process_PATCH(self, text, internal):
...
+
+        temp_file = NamedTemporaryFile(suffix='indico.tmp', dir=config.TEMP_DIR)
```
ThiefMaster: What about using a BytesIO instead of an actual file?

indico/modules/events/editing/client/js/components/FileManager/index.jsx
mic4ael: So, I did it like that because prettier was complaining when I added `{' '}` after `<Icon />` componentrr

indico/modules/events/editing/blueprint.py
mic4ael: Hm, no strong opinion. `revision-files.zip` doesn't look bad to me.

indico/modules/events/editing/client/js/components/FileManager/index.jsx
ThiefMaster: how did it complain (and without being able to fix it)?

indico/modules/events/editing/controllers/backend.py
mic4ael: Hm, yeah, good suggestion.

indico/modules/events/editing/client/js/components/FileManager/index.jsx
mic4ael: It complained about `{' '}`. The way to fix it was to remove it which made the icon inside the button look weird. That is a bit weird since we have a few other places where we do sth similar. Maybe I will take a closer look at that.

indico/modules/events/editing/client/js/components/FileManager/index.jsx
mic4ael: That was intentional - to separate our general modules from the local ones (local to the component path)

indico/modules/events/editing/controllers/backend.py
mic4ael: Hm, but `secure_filename` also makes sure that characters like `../` are stripped. I still would use it to assure security and not care about some characters being stripped.

ThiefMaster: if you pass bytes to werkzeug's secure_filename it will only strip nasty stuff

mic4ael: What do you mean exactly?

```python
In [70]: secure_filename(u'../łąe/as/asd/'.encode('utf-8'), 'test')                                                                                                                                                                                                                                                           
Out[70]: 'lae_as_asd'
```

That looks fine for me

ThiefMaster: ah i didn't notice that the werkzeug function strips more stuff later with a regex..  so nevermind on that.

ThiefMaster: anyway, since this is in a zipfile (opened on many different systems) maybe it's easier to stick with our `secure_filename`..

indico/modules/events/editing/controllers/backend.py
```diff
@@ -250,3 +259,23 @@ def _process_PATCH(self, text, internal):
...
+        buf = BytesIO()
+        with ZipFile(buf, 'w', allowZip64=True) as zip_handler:
+            for revision_file in self.revision.files:
+                file = revision_file.file
+                filename = secure_filename(file.filename, 'file-{}'.format(file.id))
+                file_type = revision_file.file_type
+                folder_name = secure_filename(file_type.name, 'file-type-{}'.format(file_type.id))
+
+                with file.storage.get_local_path(file.storage_file_id) as filepath:
+                    zip_handler.write(filepath.encode('utf-8'), os.path.join(folder_name, filename))
```
ThiefMaster: is the `.encode('utf-8')` on the local path actually needed? and in case of local storage, `sys.getfilesystemencoding()` would be more appropriate

mic4ael: Nope, it is not.

PR #4171, opened 2019-11-21
Add template_hook 'before-regform'

giusedb:
There is already a `after-regform` template_hook. This change adds a `before-regform` with the same arguments of the `after-regform`.

### DIFF
--- a/indico/modules/events/registration/templates/display/regform_display.html
+++ b/indico/modules/events/registration/templates/display/regform_display.html
@@ -98,6 +98,7 @@
             {% endcall %}
         {% endif %}
 
+        {{ template_hook('before-regform', event=event, regform=regform, management=management, registration=none) }}
         <div ng-app="nd" ng-controller="AppCtrl">
             <div nd-reg-form
                  conf-id="{{ event.id }}"
--- a/indico/modules/events/registration/templates/display/registration_modify.html
+++ b/indico/modules/events/registration/templates/display/registration_modify.html
@@ -16,6 +16,8 @@
 
 {% block content %}
     {% if registration.can_be_modified or registration.invitation %}
+        {{ template_hook('before-regform', event=event, regform=registration.registration_form,
+                         registration=registration, management=false) }}
         <div ng-app="nd" ng-controller="AppCtrl">
             <div nd-reg-form
                  conf-id="{{ event.id }}"
@@ -30,7 +32,7 @@
             </div>
         </div>
         {{ template_hook('after-regform', event=event, regform=registration.registration_form,
-                         registration=registration, management=False) }}
+                         registration=registration, management=false) }}
     {% else %}
         {% call message_box('error') %}
             {% if not registration.registration_form.is_modification_open -%}
--- a/indico/modules/events/registration/templates/management/registration_modify.html
+++ b/indico/modules/events/registration/templates/management/registration_modify.html
@@ -7,6 +7,7 @@
 {% endblock %}
 
 {% block content %}
+    {{ template_hook('before-regform', event=event, regform=regform, registration=registration, management=true) }}
     <div ng-app="nd" ng-controller="AppCtrl">
         <div nd-reg-form
              conf-id="{{ event.id }}"
@@ -21,4 +22,5 @@
              management="true"
              notify-default="{{ session.get('registration_notify_user_default', true) | tojson | forceescape }}"></div>
     </div>
+    {{ template_hook('after-regform', event=event, regform=regform, registration=registration, management=true) }}
 {% endblock %}
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,7 @@ Internal Changes
 ^^^^^^^^^^^^^^^^
 
 - Make React and SemanticUI usable everywhere (:issue:`3955`)
+- Add ``before-regform`` template hook (:issue:`4171`, thanks :user:`giusedb`)
 
 
 ----


indico/modules/events/registration/templates/display/regform_display.html
```diff
@@ -97,7 +97,7 @@
                 {%- endtrans %}
             {% endcall %}
         {% endif %}
-
```
ThiefMaster: I would keep this blank line

indico/modules/events/registration/templates/display/registration_modify.html
```diff
@@ -16,6 +16,8 @@
 
 {% block content %}
     {% if registration.can_be_modified or registration.invitation %}
+        {{ template_hook('before-regform', event=event, regform=registration.registration_form,
+                         registration=registration, management=False) }}
```
ThiefMaster: lowercase `false` (can you also fix it in the existing hook below)?

(jinja recommends lowercase true/false/none even though the python-style capitalization works as well)

indico/modules/events/registration/templates/management/regform_modify.html
```diff
@@ -7,6 +7,7 @@
 {% endblock %}
 
 {% block content %}
+    {{ template_hook('before-regform', event=event, regform=regform, management=management, registration=none) }}
```
ThiefMaster: are you sure this is needed here? it's for editing the regform itself, not for using it for a registration

indico/modules/events/registration/templates/management/registration_modify.html
```diff
@@ -7,6 +7,7 @@
 {% endblock %}
 
 {% block content %}
+    {{ template_hook('before-regform', event=event, regform=regform, management=management, registration=none) }}
```
ThiefMaster: I don't think this one is correct. `registration=none` but we are modifying an existing registration here so we should pass it..

ThiefMaster: Also, I'd add an `after-regform` to be consistent

indico/modules/events/registration/templates/display/regform_display.html
giusedb: Ok, I'm going to restore it.

indico/modules/events/registration/templates/display/registration_modify.html
giusedb: Good, I just copy and paste from the line below for `after-regform`
. I'm going to fix also the line below.

indico/modules/events/registration/templates/management/regform_modify.html
giusedb: I'm sure. It's useful if the conference manager wants to customize the block above from the UI.
In my case, this block looks like a section of the registration form, and the manager can consistently customize title and subtitle as per the other section in the registration form.
The manager should perceive it as a "_regular_" section

indico/modules/events/registration/templates/management/registration_modify.html
giusedb: You're right it's not correct indeed.
Thank you to noticed it.
I'm going to fix it

indico/modules/events/registration/templates/management/regform_modify.html
ThiefMaster: I wonder if it should be a different signal. If not, this signal **needs** a new kwarg (`regform_management`? sounds awful but i can't think of a better one right now) that is usually set to false and true here.

giusedb: I get your concerns, but consistently with the show up of _angular_ registration-form, It sounds to me more like `above-regform` and `below-regform` for two pieces of bread that you graphically need to create your sandwich, with the registration form in between.

ThiefMaster: yes sure, i just meant something like `above-regform-management` because i think it's unlikely that one wants to inject the same code on the two pages

giusedb: I consider these as an _extension_ of the _angular_ registration form.

mic4ael:
I would squash all the commits into one.

indico/modules/events/registration/templates/management/regform_modify.html
giusedb: So, what's the purpose of the `management` parameter?
you can always implement a function like:
```
if management:
 do_this()
else:
 do_that()
return render_template(same_template)
```
In my case, the only difference is just a different parameter passed to my javascript function.
Everything else is exactly the same.

ThiefMaster: But the current `management` parameter indicates if the registration is being edited  in the management area or not. 

Now that you are adding a hook to `regform_modify.html`, this will run when managing the **registration form itself**. And a plugin that wants to show something during registration, most likely does not want to show this when the registration form is being edited.

ThiefMaster: And the reason why I suggested a separate signal:

Right now the cern_access plugin uses this code to inject its fields when someone registers for the event:

```python
    def _get_personal_data_form(self, event, regform, management, registration=None, **kwargs):
if management or registration is not None:
    return
```

If you start calling the hooks when managing the regform, this code will also run. Of course it's no problem for us to update the plugin - but IMO those are two totally different things (registrations vs registration forms), so separate hooks would avoid this problem altogether.

giusedb: Ok, but at this point, both of them should be separated. How can we proceed? you separate yours while I separate mine?
Do you separate both?

ThiefMaster: not sure what you mean... the existing hooks only run when you create/edit registrations. so i'd add the before-regform hooks in those files where we already have `after-regform` ones.

and if you do need hooks in regform management as well, then (maybe as a separate commit) add new before/after signals. `<before|after>-regform-setup` or something like that

giusedb: OK, let's proceed this way. I need those hooks for the regform management. Later on, I'll create the `<before|after>-regform-setup`.

PR #4179, opened 2019-11-21
Add event feature for the editing module

ThiefMaster:


indico/modules/events/editing/__init__.py
```diff
@@ -7,7 +7,27 @@
...
+from indico.util.i18n import _
 
 
 logger = Logger.get('events.editing')
+
+
+class EditingFeature(EventFeature):
+    name = 'editing'
+    friendly_name = _('Paper Editing')
+    description = _('Gives event managers the opportunity to let contributors submit papers to be reviewed and '
```
meluru: `to be edited` ?

ThiefMaster: I thought the same, but is "edited" in that context clear?

ThiefMaster: I thought the same, but is "edited" in that context clear?

pferreir: I'd choose "edited".

PR #4181, opened 2019-11-22
Use flake8 instead of pycodestyle in the hook

mic4ael:


pre-commit.githook
```diff
@@ -4,12 +4,10 @@
 # Interesting post on max line length:
 # http://stackoverflow.com/questions/88942/why-should-python-pep-8-specify-a-maximum-line-length-of-79-characters
 
-PEP8_CMD='pycodestyle'
-PEP8_OPTIONS='--max-line-length=120'
```
mic4ael: We don't need that here since it is already present in the `setup.cfg`

```

[flake8]
exclude = .git,build,dist,ext_modules,node_modules,indico.egg-info
max-line-length = 120
```

PR #4182, opened 2019-11-22
Improve comment visibility dropdown

### LINKED ISSUE #4172: Improve "comments" input.
We currently have a dropdown to specify the visibility of a comment:

image

The mockups suggest another strategy:

image

We should check which option is best and come up with something user-friendly.

mic4ael:
closes #4172 

Peek 2019-11-22 09-23

### DIFF
--- a/indico/modules/events/client/js/reviewing/components/ReviewForm.jsx
+++ b/indico/modules/events/client/js/reviewing/components/ReviewForm.jsx
@@ -9,24 +9,13 @@ import React, {useState} from 'react';
 import {Form as FinalForm} from 'react-final-form';
 import {Button, Dropdown, Form} from 'semantic-ui-react';
 
-import {FinalDropdown, FinalSubmitButton, FinalTextArea} from 'indico/react/forms';
+import {FinalCheckbox, FinalInput, FinalSubmitButton, FinalTextArea} from 'indico/react/forms';
 import {Translate} from 'indico/react/i18n';
 
 import JudgmentModal from './JudgmentModal';
 import UserAvatar from './UserAvatar';
 
 import './ReviewForm.module.scss';
-
-const visibilityOptions = [
-  {
-    value: 'editors',
-    text: Translate.string('Visible only to editors'),
-  },
-  {
-    value: 'authors',
-    text: Translate.string('Visible to editors and authors'),
-  },
-];
 
 const judgmentOptions = [
   {
@@ -61,6 +50,8 @@ export default function ReviewForm() {
     }
   };
 
+  const InputComponent = commentFormVisible ? FinalTextArea : FinalInput;
+  const inputProps = commentFormVisible ? {autoFocus: true} : {};
   return (
     <>
       <div className="i-timeline-item review-timeline-input">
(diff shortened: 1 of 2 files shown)


ThiefMaster:
I think this needs more discussion on the UI, because:

- showing the visibility selector only in the unexpanded view isn't very user-friendly
- you currently don't see which visibility is selected in any view unless you open the dropdown again (maybe have different icons?)

---

Besides the dropdown, maybe a checkbox/toggle could be another option? After all, the only choices are making it internal or not.

meluru:
At least in JACoW case internal comment wasn't used so often. Maybe a toggle with a default as public will work well?

mic4ael:
> At least in JACoW case internal comment wasn't used so often. Maybe a toggle with a default as public will work well?

Yeah, if we only have two options, then let's go for a switch-like component. But then we should also agree where to put it..

mic4ael:
Honestly, I would just leave it as it is right now (without the PR) and maybe just change the component we use.

meluru:
> Yeah, if we only have two options, then let's go for a switch-like component. But then we should also agree where to put it..

I would try placing it where the selector is currently and see if it works :)

mic4ael:
To me at least, it doesn't make sense to put this in the "unexpanded" view, since in order to comment you would have to:

1) Select visibility
2) Put text in the textarea
3) Press "Comment" button

The first step shouldn't be required. That is why I would just put it in the expanded view.

meluru:
> To me at least, it doesn't make sense to put this in the "unexpanded" view, since in order to comment you would have to:
> 
>     1. Select visibility
> 
>     2. Put text in the textarea
> 
>     3. Press "Comment" button
> 
> 
> The first step shouldn't be required. That is why I would just put it in the expanded view.

Yeah, that actually sounds much better. I just realized the visibility disappears when you expand the comment.

indico/modules/events/client/js/reviewing/components/ReviewForm.jsx
```diff
@@ -71,41 +62,40 @@ export default function ReviewForm() {
...
-                            options={visibilityOptions}
-                            width={8}
-                            selection
-                            required
+                        <FinalCheckbox
+                          label={Translate.string(
+                            'Restrict visibility of this comment to other editors'
+                          )}
+                          name="protected"
+                          toggle
```
ThiefMaster: hm, wouldn't an actual checkbox be nicer?

indico/modules/events/client/js/reviewing/components/ReviewForm.jsx
```diff
@@ -71,41 +62,40 @@ export default function ReviewForm() {
               <div className="f-self-stretch">
                 <FinalForm
                   onSubmit={() => {}}
-                  initialValues={{comment: '', visibility: 'editors'}}
-                  subscription={{submitting: true}}
+                  initialValues={{comment: '', protected: false}}
+                  subscription={{submitting: true, values: true}}
```
ThiefMaster: you don't need a values subscription here

indico/modules/events/client/js/reviewing/components/ReviewForm.jsx
meluru: I've suggested a toggle :) What others think? @panagiotappl @pferreir

indico/modules/events/client/js/reviewing/components/ReviewForm.jsx
mic4ael: Ah, yeah. Forgot to remove it. At some point I needed that when I wanted to change icon's color based on the selected value :P

indico/modules/events/client/js/reviewing/components/ReviewForm.jsx
mic4ael: To be honest, I prefer a checkbox a bit more since it is a bit more obvious what it means when I check it off.

mic4ael: But let's see what others think.

indico/modules/events/client/js/reviewing/components/ReviewForm.jsx
ThiefMaster: Tip, you can wrap code that needs the value of a particular field like this:

```jsx
<Field name="thefieldyoucareabout" subscription={{value: true}}>
  {({input: {value}}) => (
    ...
  )}
</Field>
```

That way you only re-render the part that needs it and only when the value of the field you care about changes - avoiding re-renders whenever someone types in a text field.

I used that for `_map_url_template_choice` in RB for example.

mic4ael: Yeah, that is pretty useful if you actually want to render something "conditionally". In my case, I wanted to pass a different value as a prop to a component so I think it would be still better to just get it directly from `fprops.values`

PR #4183, opened 2019-11-22
Add submitter confirmation component

meluru:
Closes #4183 

image

### DIFF
--- a/indico/modules/events/editing/client/js/actions.js
+++ b/indico/modules/events/editing/client/js/actions.js
@@ -26,6 +26,12 @@ export function reviewEditable(revision, formData) {
   );
 }
 
+export function confirmEditableChanges(revision, formData) {
+  return submitFormAction(() => indicoAxios.post(revision.confirmURL, formData), null, () =>
+    loadTimeline()
+  );
+}
+
 export function createRevisionComment(url, formData) {
   return submitFormAction(() => indicoAxios.post(url, formData), null, () => loadTimeline());
 }
--- a/indico/modules/events/editing/client/js/components/Timeline/ChangesConfirmation.jsx
+++ b/indico/modules/events/editing/client/js/components/Timeline/ChangesConfirmation.jsx
@@ -1,0 +1,71 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React from 'react';
+import {Form as FinalForm} from 'react-final-form';
+import {Divider, Form, Message} from 'semantic-ui-react';
+import {FinalSubmitButton, FinalTextArea} from 'indico/react/forms';
+import {Translate} from 'indico/react/i18n';
+
+import './ChangesConfirmation.module.scss';
+
+export default function ChangesConfirmation() {
+  const confirmChanges = formData => {
+    console.log(formData);
+  };
+
+  return (
+    <div>
+      <Divider />
+      <FinalForm onSubmit={confirmChanges}>
+        {fprops => (
+          <Form onSubmit={fprops.handleSubmit}>
+            <Message
+              info
+              size="tiny"
+              icon="exclamation circle"
+              header={Translate.string('Confirmation required')}
+              content={
+                <p>
+                  <Translate>
+                    Editor uploaded changes to your paper. Please download the updated files and
+                    verify their changes. If you reject them, you can upload a new version
+                    afterwards.
+                  </Translate>
+                </p>
+              }
+            />
+            <FinalTextArea
+              name="comment"
+              placeholder={Translate.string('You can leave a comment if you wish')}
+            />
+            <Form.Group inline styleName="submit-buttons">
+              <FinalSubmitButton
+                label={Translate.string('Accept')}
+                color="green"
+                icon="check"
+                onClick={() => {
+                  fprops.form.change('action', 'accept');
+                }}
+                disabledUntilChange={false}
+              />
+              <FinalSubmitButton
+                label={Translate.string('Reject')}
+                color="red"
+                icon="times"
+                onClick={() => {
+                  fprops.form.change('action', 'reject');
+                }}
+                disabledUntilChange={false}
+              />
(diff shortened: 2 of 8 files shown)


indico/modules/events/client/js/reviewing/components/ChangesConfirmation.jsx
```diff
@@ -0,0 +1,68 @@
...
+                  reject the changes and provide a new version by yourself.'
+              )}
+            />
+            <FinalTextArea
+              name="comment"
+              placeholder={Translate.string('If you wish you can leave a comment...')}
+            />
+            <div styleName="submit-buttons">
+              <Form.Button
+                form="confirmation-form"
```
mic4ael: You don't need this prop on these `Button`s since they are already placed within a form.

indico/modules/events/client/js/reviewing/components/ChangesConfirmation.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+import './ChangesConfirmation.module.scss';
+
+export default function ChangesConfirmation() {
+  const confirmChanges = formData => {
+    console.log(formData);
+  };
+
+  return (
+    <div styleName="confirmation">
```
mic4ael: I think that you don't need that at all since `Divider` already adds some nice margin.

Additional margin


Only `Divider` (without `confirmation` class)

indico/modules/events/client/js/reviewing/components/ChangesConfirmation.jsx
```diff
@@ -0,0 +1,66 @@
...
+      <FinalForm onSubmit={confirmChanges}>
+        {fprops => (
+          <Form id="confirmation-form" onSubmit={fprops.handleSubmit}>
+            <Message
+              info
+              size="tiny"
+              icon="exclamation circle"
+              header={Translate.string('Confirmation required')}
+              content={Translate.string(
+                'Editor uploaded changes to your paper. Please download and confirm. You can \
```
ThiefMaster: I guess "Editor" will be replaced with their name?

ThiefMaster: Can we avoid having to wrapping the string here with a `\`? It's super ugly.. :/

Maybe `<Translate>..,</Translate>` works here as well?

indico/modules/events/client/js/reviewing/components/ChangesConfirmation.jsx
```diff
@@ -0,0 +1,66 @@
...
+                  reject the changes and provide a new version by yourself.'
+              )}
+            />
+            <FinalTextArea
+              name="comment"
+              placeholder={Translate.string('If you wish you can leave a comment...')}
```
ThiefMaster: "You can leave a comment if you wish"?

indico/modules/events/client/js/reviewing/components/ChangesConfirmation.jsx
```diff
@@ -0,0 +1,66 @@
...
+            />
+            <Form.Group inline styleName="submit-buttons">
+              <Form.Button
```
ThiefMaster: Use the logic from `FinalSubmitButton` to disable those buttons while submitting etc.

Best option is probably to pass along extra props in the FinalSubmitButton component so you can use it here.

indico/modules/events/client/js/reviewing/components/ChangesConfirmation.jsx
```diff
@@ -0,0 +1,66 @@
...
+  const confirmChanges = formData => {
+    console.log(formData);
+  };
+
+  return (
+    <div>
+      <Divider />
+      <FinalForm onSubmit={confirmChanges}>
+        {fprops => (
+          <Form id="confirmation-form" onSubmit={fprops.handleSubmit}>
```
mic4ael: You don't need that `id` attribute anymore ;)

indico/modules/events/client/js/reviewing/components/ChangesConfirmation.jsx
meluru: yeah, i think it would be nice to replace it with their name later

indico/modules/events/client/js/reviewing/components/ChangesConfirmation.jsx
```diff
@@ -0,0 +1,70 @@
...
+            <Message
+              info
+              size="tiny"
+              icon="exclamation circle"
+              header={Translate.string('Confirmation required')}
+              content={
+                <p>
+                  <Translate>
+                    Editor uploaded changes to your paper. Please download and confirm. You can
+                    reject the changes and provide a new version by yourself.
```
mic4ael: Just `You can reject the changes and provide a new version of the paper`?

indico/modules/events/client/js/reviewing/components/TimelineItem.jsx
```diff
@@ -67,6 +68,10 @@ export default function TimelineItem({revision, isLastRevision, state}) {
                     <div dangerouslySetInnerHTML={{__html: revision.commentHtml}} />
                   </>
                 )}
+                {/* TODO: Check whether the current user is submitter */}
+                {isLastRevision && state.name === 'needs_submitter_confirmation' && (
```
ThiefMaster: should use the enum (InitialRevisionState or FinalRevisionState) depending on which state we are using here

indico/modules/events/client/js/reviewing/components/ChangesConfirmation.jsx
```diff
@@ -0,0 +1,70 @@
...
+import {Form as FinalForm} from 'react-final-form';
+import {Divider, Form, Message} from 'semantic-ui-react';
+import {FinalSubmitButton, FinalTextArea} from 'indico/react/forms';
+import {Translate} from 'indico/react/i18n';
+
+import './ChangesConfirmation.module.scss';
+
+export default function ChangesConfirmation() {
+  const confirmChanges = formData => {
+    console.log(formData);
```
ThiefMaster: we already have the APIs for this so let's actually implement answering. can probably use a redux action similar to what we use when submitting a review

PR #4186, opened 2019-11-25
Add interface to upload editables

### LINKED ISSUE #4140: Paper Editing: way to upload editables
Right now there is no interface to upload an editable. This should be based on the file manager (#4119) of course.

panagiotappl:
This PR assumes that we upload papers only.

In this PR:
- [x] Add editable submission area in Contribution page
- [x] Create modal that renders an empty FileManager and creates editable when saving
- [x] When editable exists show link to timeline (draft view)
- [x] Redirect to timeline when success

closes #4140

### DIFF
--- a/indico/modules/events/contributions/client/js/index.jsx
+++ b/indico/modules/events/contributions/client/js/index.jsx
@@ -8,13 +8,51 @@
 /* global showUndoWarning:false, setupListGenerator:false, setupSearchBox:false */
 /* global reloadManagementAttachmentInfoColumn:false */
 
+import fileTypesURL from 'indico-url:event_editing.api_file_types';
+
 import React from 'react';
 import ReactDOM from 'react-dom';
 
 import 'indico/modules/events/util/types_dialog';
+import EditableSubmissionButton from 'indico/modules/events/editing/components/EditableSubmissionButton';
 import {PublicationSwitch} from 'indico/react/components';
+import {camelizeKeys} from 'indico/utils/case';
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
 
 (function(global) {
+  global.setupEditableSubmissionButton = async function setupEditableSubmissionButton() {
+    const editableSubmissionButton = document.querySelector('#editable-submission-button');
+    if (!editableSubmissionButton) {
+      return;
+    }
+    const {
+      eventId,
+      contributionId,
+      apiEditableSubmitUrl,
+      type,
+      uploadUrl,
+    } = editableSubmissionButton.dataset;
+
+    let response;
+    try {
+      response = await indicoAxios.get(fileTypesURL({confId: eventId}));
+    } catch (e) {
+      handleAxiosError(e);
+    }
+    const fileTypes = camelizeKeys(response.data);
+    ReactDOM.render(
+      <EditableSubmissionButton
+        uploadURL={uploadUrl}
+        fileTypes={fileTypes}
+        type={type}
+        submitRevisionURL={apiEditableSubmitUrl}
+        eventId={eventId}
+        contributionId={contributionId}
+      />,
+      editableSubmissionButton
+    );
+  };
+
   function setupPublicationButton() {
     const element = document.querySelector('#pub-switch');
     const component = React.createElement(PublicationSwitch, {eventId: element.dataset.eventId});
--- a/indico/modules/events/contributions/templates/display/contribution_display.html
+++ b/indico/modules/events/contributions/templates/display/contribution_display.html
@@ -2,7 +2,7 @@
 
 {% from 'events/display/common/_conferences.html' import render_attachments %}
 {% from 'events/display/indico/_common.html' import render_location, render_users, render_speakers %}
-{% from 'events/papers/_contributions.html' import render_paper_section %}
+{% from 'events/papers/_contributions.html' import render_paper_section, render_editables_section %}
 {% from 'message_box.html' import message_box %}
 
 {% block page_class %}conference-page item-summary{% endblock %}
@@ -258,6 +258,12 @@
     {%- endif %}
     {% if contribution.event.has_feature('papers') %}
         {{ render_paper_section(contribution) }}
+    {% endif %}
+    {% if contribution.event.has_feature('editing') %}
+        {{ render_editables_section(contribution) }}
+        <script>
+            setupEditableSubmissionButton();
+        </script>
     {% endif %}
     <script>
         setupAttachmentTreeView();
--- a/indico/modules/events/editing/client/js/components/EditableSubmissionButton.jsx
+++ b/indico/modules/events/editing/client/js/components/EditableSubmissionButton.jsx
@@ -1,0 +1,87 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import editableURL from 'indico-url:event_editing.editable';
+import React, {useCallback, useState} from 'react';
+import PropTypes from 'prop-types';
+import {Button, Modal} from 'semantic-ui-react';
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+import {Translate} from 'indico/react/i18n';
+import FileManager from './FileManager';
+import {fileTypePropTypes} from './FileManager/util';
+
+export default function EditableSubmissionButton({
+  eventId,
+  contributionId,
+  uploadURL,
+  fileTypes,
+  type,
+  submitRevisionURL,
+}) {
+  const [open, setOpen] = useState(false);
+  const [submissionFiles, setSubmissionFiles] = useState({});
+  const [submitting, setSubmitting] = useState(false);
+  const submitRevision = async () => {
+    setSubmitting(true);
+    try {
+      await indicoAxios.put(submitRevisionURL, {
+        files: submissionFiles,
+      });
+    } catch (e) {
+      handleAxiosError(e, false, true);
+      setSubmitting(false);
+      return;
+    }
+    setSubmitting(false);
+    location.href = editableURL({confId: eventId, contrib_id: contributionId, type});
+  };
+
+  const handleChange = useCallback(files => setSubmissionFiles(files), []);
+  return (
+    <>
+      <Modal
+        open={open}
+        onClose={() => setOpen(false)}
+        closeIcon
+        closeOnDimmerClick={false}
+        closeOnEscape={false}
+      >
+        <Modal.Header>
+          {type === 'paper' && <Translate>Submit your paper</Translate>}
+          {type === 'slides' && <Translate>Submit your slides</Translate>}
+        </Modal.Header>
+        <Modal.Content>
+          <FileManager onChange={handleChange} fileTypes={fileTypes} uploadURL={uploadURL} />
+        </Modal.Content>
+        <Modal.Actions>
+          <Button onClick={() => setOpen(false)}>
+            <Translate>Cancel</Translate>
+          </Button>
+          <Button
+            primary
+            disabled={submitting || !Object.keys(submissionFiles).length}
+            onClick={submitRevision}
+          >
+            <Translate>Save</Translate>
+          </Button>
+        </Modal.Actions>
+      </Modal>
+      <button type="submit" className="i-button highlight" onClick={() => setOpen(true)}>
+        <Translate>Submit Files</Translate>
+      </button>
+    </>
+  );
+}
+
+EditableSubmissionButton.propTypes = {
+  uploadURL: PropTypes.string.isRequired,
+  type: PropTypes.string.isRequired,
+  submitRevisionURL: PropTypes.string.isRequired,
+  fileTypes: PropTypes.arrayOf(PropTypes.shape(fileTypePropTypes)).isRequired,
+  eventId: PropTypes.string.isRequired,
+  contributionId: PropTypes.string.isRequired,
+};
--- a/indico/modules/events/papers/templates/_contributions.html
+++ b/indico/modules/events/papers/templates/_contributions.html
@@ -11,6 +11,11 @@
         <section>
             <div class="header">
                 <div class="header-row">
+                    <h2>{% trans %}Peer reviewing{% endtrans %}</h2>
+                </div>
+            </div>
+            <div class="header">
+                <div class="header-row">
                     <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
                     {% if show_paper and paper.state.name == 'accepted' %}
                         {{ render_paper_metadata(paper, contribution_page=true) }}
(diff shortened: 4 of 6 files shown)


indico/modules/events/contributions/client/js/index.jsx
```diff
@@ -8,13 +8,40 @@
...
 (function(global) {
+  global.setupEditableSubmissionButton = async function setupEditableSubmissionButton() {
+    const editableSubmissionButton = document.querySelector('#editable-submission-button');
+    const {eventId, apiEditableSubmitUrl, type, uploadUrl} = editableSubmissionButton.dataset;
+
+    let response;
+    try {
+      response = await indicoAxios.get(fileTypesURL({confId: eventId}));
+    } catch (e) {
+      handleAxiosError(e);
```
ThiefMaster: return in this case to avoid breaking the code below which would get an undefined response

indico/modules/events/editing/client/js/components/FileManager/index.jsx
```diff
@@ -104,13 +104,17 @@ FileType.defaultProps = {
...
 
-export default function FileManager({downloadURL, uploadURL, fileTypes, files}) {
+export default function FileManager({onChange, downloadURL, uploadURL, fileTypes, files}) {
   const _fileTypes = useMemo(() => mapFileTypes(fileTypes, files), [fileTypes, files]);
   const [state, dispatch] = useReducer(reducer, {
     fileTypes: _fileTypes,
     uploads: {},
   });
 
+  useEffect(() => {
```
ThiefMaster: using an effect for this seems to be a bit strange. I'd rather call onChange explicitly after an action affecting the final list of file types happened (e.g. the user clicking on delete, an upload finishing, etc.)

indico/web/client/js/react/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,62 @@
...
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React, {useCallback, useState} from 'react';
+import PropTypes from 'prop-types';
+import {Button, Modal} from 'semantic-ui-react';
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+
+import FileManager from '../../../../../modules/events/editing/client/js/components/FileManager';
```
ThiefMaster: `indico/modules/...`

indico/web/client/js/react/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,62 @@
...
+          Submit Files
+        </button>
+      }
+      closeIcon
+    >
+      <Modal.Header>Submit a {type}</Modal.Header>
+      <Modal.Content>
+        <FileManager
+          onChange={useCallback(fts => {
+            const fs = {};
```
ThiefMaster: fts, fs, ft. Very confusing.

indico/web/client/js/react/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,62 @@
...
+            fts.forEach(ft => (fs[ft.id] = ft.files.map(f => f.uuid)));
```
ThiefMaster: use an arrow function with curly braces if you need statements in there and not just expressions

indico/web/client/js/react/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,62 @@
...
+        <button type="submit" className="i-button highlight">
+          Submit Files
+        </button>
+      }
+      closeIcon
+    >
+      <Modal.Header>Submit a {type}</Modal.Header>
+      <Modal.Content>
+        <FileManager
+          onChange={useCallback(fts => {
```
ThiefMaster: having useCallback inlined here doesn't look great. `const handleChange = useCallback(...);` and then just pass `onChange={handleChange}` here

indico/modules/events/editing/client/js/components/FileManager/selectors.js
```diff
@@ -0,0 +1,5 @@
+export const getFiles = state => {
+  const files = {};
+  state.fileTypes.forEach(fileType => (files[fileType.id] = fileType.files.map(file => file.uuid)));
```
ThiefMaster: `=` assignment does not belong in an arrow function without curly braces. Doing an assignment as an expression is pretty ugly in this context..

```js
export const getFiles = state => {
  const files = {};
  state.fileTypes.forEach(fileType => {
    files[fileType.id] = fileType.files.map(file => file.uuid);
  });
  return files;
};
```

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -31,6 +36,56 @@ <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
     {% endif %}
 {% endmacro %}
 
+{% macro render_editables_section(contribution) %}
+    {% set editable = contribution.editables[0] if contribution.editables if contribution.editables else None %}
+    {% set show_editable = true if editable else false %}
```
ThiefMaster: Just use `if editable` below and get rid of this

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -31,6 +36,56 @@ <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
     {% endif %}
 {% endmacro %}
 
+{% macro render_editables_section(contribution) %}
+    {% set editable = contribution.editables[0] if contribution.editables if contribution.editables else None %}
```
ThiefMaster: lowercase `none` in jinja

ThiefMaster: you have `if contribution.editables` there twice

ThiefMaster: eventually this needs to support multiple editables instead of doing the `editables[0]` 'hack' - maybe add a `{# TODO: ... #}` comment for this

ThiefMaster: `{% set editable = contribution.editables|selectattr('type', 'equalto', 'paper')|first %}` probably works as well

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -31,6 +36,56 @@ <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
...
+                </div>
+            </div>
+            <div class="header">
+                <div class="header-row">
+                    <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
+                </div>
+            </div>
+            {% if show_editable %}
+                <div>
+                    <a href="{{ url_for('event_editing.editable', contribution.event, contrib_id=contribution.id, type='paper') }}">
```
ThiefMaster: you can just pass the `editable` to `url_for` instead of the kwargs since there are locators now that provide all the necessary data

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -31,6 +36,56 @@ <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
...
+                                <div>
+                                    {% trans %}You can submit a paper{% endtrans %}
+                                </div>
+                            </div>
+                        <div class="toolbar">
+                            <div id="editable-submission-button"
+                                data-event-id="{{ contribution.event_id }}"
+                                data-contribution-id="{{ contribution.id }}"
+                                data-type="paper"
+                                data-upload-url="{{ url_for('event_editing.api_upload', contribution.event, contrib_id=contribution.id, type='paper') }}"
```
ThiefMaster: `contribution.event, contrib_id=contribution.id` can be replaced with `contribution` (its locator provides both values)

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -31,6 +36,56 @@ <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
...
+                                data-api-editable-submit-url="{{ url_for('event_editing.api_create_editable', contribution.event, contrib_id=contribution.id, type='paper') }}">
```
ThiefMaster: same here

indico/web/client/js/react/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,71 @@
...
+  eventId,
+  contributionId,
+  uploadURL,
+  fileTypes,
+  type,
+  submitRevisionURL,
+}) {
+  const [submissionFiles, setSubmissionFiles] = useState({});
+  const [submitting, setSubmitting] = useState(false);
+  async function submitRevision() {
```
ThiefMaster: usually we use arrow functions for nested functions

indico/web/client/js/react/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,71 @@
...
+      await indicoAxios.put(submitRevisionURL, {
+        files: submissionFiles,
+      });
+    } catch (e) {
+      handleAxiosError(e, false, true);
+      setSubmitting(false);
+      return;
+    }
+    setSubmitting(false);
+    window.location.href = `/event/${eventId}/contributions/${contributionId}/editing/${type}`;
```
ThiefMaster: import `indico-url:event_editing.editable` and use that to build the url. hardcoding doesn't only break in case we ever change the url blueprint, but also when running indico in a subdir (even though we don't really support that anymore)

indico/web/client/js/react/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  const handleChange = useCallback(files => setSubmissionFiles(files), []);
+  return (
+    <Modal
+      trigger={
+        <button type="submit" className="i-button highlight">
+          Submit Files
```
ThiefMaster: i18n

indico/web/client/js/react/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,71 @@
...
+        </button>
+      }
+      closeIcon
+    >
+      <Modal.Header>Submit a {type}</Modal.Header>
```
ThiefMaster: i18n and you have to explicitly write two strings for paper and slides since interpolating the `{type}` in the translated string won't work for anything but english.

"Submit your paper" and "Submit your slides" are probably good texts for it

indico/web/client/js/react/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,71 @@
...
+      <Modal.Content>
+        <FileManager onChange={handleChange} fileTypes={fileTypes} uploadURL={uploadURL} />
+      </Modal.Content>
+      <Modal.Actions>
+        <Button>Close</Button>
+        <Button primary disabled={submitting} onClick={submitRevision}>
+          Save
```
ThiefMaster: i18n

indico/web/client/js/react/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,71 @@
...
+        </button>
+      }
+      closeIcon
+    >
+      <Modal.Header>Submit a {type}</Modal.Header>
+      <Modal.Content>
+        <FileManager onChange={handleChange} fileTypes={fileTypes} uploadURL={uploadURL} />
+      </Modal.Content>
+      <Modal.Actions>
+        <Button>Close</Button>
```
ThiefMaster: i18n

indico/web/client/js/react/components/index.js
```diff
@@ -32,3 +32,4 @@ export {default as ResponsivePopup} from './ResponsivePopup';
 export {default as PublicationSwitch} from './PublicationSwitch';
 export {default as MathJax} from './MathJax';
 export {default as FileSubmission} from './FileSubmission';
+export {default as EditableSubmissionButton} from './EditableSubmissionButton';
```
ThiefMaster: I think this component would fit better in the editing module than here, since this place is meant for common components not specific to some module.

indico/modules/events/editing/client/js/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,77 @@
...
+      await indicoAxios.put(submitRevisionURL, {
+        files: submissionFiles,
+      });
+    } catch (e) {
+      handleAxiosError(e, false, true);
+      setSubmitting(false);
+      return;
+    }
+    setSubmitting(false);
+    window.location.href = editableURL({confId: eventId, contrib_id: contributionId, type});
```
ThiefMaster: I just noticed that we use `location.href` instead of `window.location.href` in most other places.

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -31,6 +36,56 @@ <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
...
+                                <div class="label">
+                                    {% trans %}Paper submission is open{% endtrans %}
+                                </div>
+                                <div>
+                                    {% trans %}You can submit a paper{% endtrans %}
+                                </div>
+                            </div>
+                        <div class="toolbar">
+                            <div id="editable-submission-button"
+                                data-event-id="{{ contribution.event_id }}"
```
ThiefMaster: incorrect indentation (in html we currently align the attributes)

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -31,6 +36,56 @@ <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
     {% endif %}
 {% endmacro %}
 
+{% macro render_editables_section(contribution) %}
+    {# TODO: Support mulitple editable rendering #}
+    {% set editable = contribution.editables[0] if contribution.editables[0] else none %}
```
ThiefMaster: did what i suggested in my earlier comment not work?

indico/modules/events/contributions/client/js/index.jsx
```diff
@@ -8,13 +8,48 @@
...
 
 import 'indico/modules/events/util/types_dialog';
+import EditableSubmissionButton from 'indico/modules/events/editing/components/EditableSubmissionButton';
 import {PublicationSwitch} from 'indico/react/components';
+import {camelizeKeys} from 'indico/utils/case';
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
 
 (function(global) {
+  global.setupEditableSubmissionButton = async function setupEditableSubmissionButton() {
+    const editableSubmissionButton = document.querySelector('#editable-submission-button');
```
pferreir: While testing I noticed that this element may not exist (e.g. no submission rights) and that will mean you'll get `undefined`, which will blow up the whole thing.

indico/modules/events/editing/client/js/components/EditableSubmissionButton.jsx
```diff
@@ -0,0 +1,77 @@
...
+  };
+
+  const handleChange = useCallback(files => setSubmissionFiles(files), []);
+  return (
+    <Modal
+      trigger={
+        <button type="submit" className="i-button highlight">
```
mic4ael: Did you use `button` instead of `Button` on purpose?

panagiotappl: I used indico css so I didn't think of any use for `Button`

indico/modules/events/papers/templates/_contributions.html
panagiotappl: I thought that since were both temporary approaches and we will eventually iterate over all editables to not change it for now.

indico/modules/events/contributions/client/js/index.jsx
panagiotappl: Is it the best approach to use `if` inside the setup function or not call the setup function at all in that case?

ThiefMaster: I don't think it makes much sense to call `setupEditableSubmissionButton` if you don't have the button

indico/modules/events/contributions/templates/display/contribution_display.html
```diff
@@ -259,7 +259,13 @@ <h3>{% trans %}Subcontributions{% endtrans %}</h3>
     {% if contribution.event.has_feature('papers') %}
         {{ render_paper_section(contribution) }}
     {% endif %}
+    {% if contribution.event.has_feature('editing') %}
+        {{ render_editables_section(contribution) }}
+    {% endif %}
     <script>
+        if (document.querySelector('#editable-submission-button')) {
```
ThiefMaster: That's a bit ugly. I thought about checking (in the jinja/pyhjon code) whether a new revision can be submitted but it'd result in duplicating logic, so I'd move the check inside the function then (just return early if you don't have the button).

Also, no need to call this if the editing feature isn't enabled. You can just use a jinja if to not render the function call at all in this case.

PR #4188, opened 2019-11-27
Allow downloading files from File{Display,Manager}

### LINKED ISSUE #4162: Allow files to be downloaded from FileManager and FileDisplay components.
Right now they don't link anywhere.

mic4ael:
closes #4162

### DIFF
--- a/indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
+++ b/indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
@@ -40,7 +40,7 @@ function FileEntry({
   uploadURL,
   fileTypeId,
   allowMultipleFiles,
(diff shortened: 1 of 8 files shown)


indico/modules/events/editing/client/js/components/FileManager/FileList.jsx
```diff
@@ -40,7 +40,7 @@ function FileEntry({
   uploadURL,
   fileTypeId,
   allowMultipleFiles,
-  file: {uuid, filename, state, claimed},
+  file: {uuid, filename, state, claimed, downloadURL: singleFileDownloadURL},
```
ThiefMaster: Why not just `downloadURL`? This component is for a specific file after all.

PR #4189, opened 2019-11-27
Support event roles in ACL field

### LINKED ISSUE #4185: Create a component for event roles assignment
Currently track permissions dialog allows only assigning users and groups. The goal is to implement a react component that will take care of assigning event roles there. This is only front end task, the back end already exists.

ThiefMaster:
closes #4185

### DIFF
--- a/indico/modules/core/controllers.py
+++ b/indico/modules/core/controllers.py
@@ -11,7 +11,6 @@ import re
 
 import requests
 from flask import flash, jsonify, redirect, request, session
-from marshmallow import ValidationError
 from packaging.version import Version
 from pkg_resources import DistributionNotFound, get_distribution
 from pytz import common_timezones_set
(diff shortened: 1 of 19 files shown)


meluru:
The code looks good so far. 

I'd only add a bit more spacing between the role name and permissions, because right now it looks squeezed compared to user.

image

ThiefMaster:
Thought about it, the problem is that it starts looking a bit funny if you don't have any roles

meluru:
> Thought about it, the problem is that it starts looking a bit funny if you don't have any roles

How would it look like if the spacing was only for roles? (so that all entries have the same height)

ThiefMaster:
Oh you mean vertical spacing. I thought you mean horizontal spacing between the icon and the title.

ThiefMaster:
an empty `List.Description` doesn't render at all (or is collapsed). so i'd have to replace that with some custom placeholder in case of roles or local groups (which are shown in the same way). i'd say that's not really in the scope of this PR ;)

PR #4191, opened 2019-11-28
Track groups fixes

meluru:
Fixes:

- track groups cloning
- tracks sorting and grouping

### DIFF
--- a/indico/modules/events/tracks/client/js/index.js
+++ b/indico/modules/events/tracks/client/js/index.js
@@ -8,6 +8,34 @@
 (function(global) {
   'use strict';
 
+  function updateSorting() {
+    const container = $('#track-list-container');
+    const sortedList = container
+      .find('li.track-row')
+      .map(function() {
+        const $this = $(this);
+        if ($this.hasClass('track-group-box')) {
+          return {id: $this.data('id'), type: 'group'};
+        } else {
+          let parent = null;
+          const parentDiv = $this.closest('.track-group-box');
+          if (parentDiv.length) {
+            parent = parentDiv.data('id');
+          }
+          return {id: $this.data('id'), type: 'track', parent};
+        }
+      })
+      .get();
+    $.ajax({
+      url: container.data('url'),
+      method: 'POST',
+      contentType: 'application/json',
+      data: JSON.stringify({sort_order: sortedList}),
+      complete: IndicoUI.Dialogs.Util.progress(),
+      error: handleAjaxError,
+    });
+  }
+
   global.setupTrackManagement = function setupTrackManagement() {
     var heightLimit = 50;
     $('#track-list-container')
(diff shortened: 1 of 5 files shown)


indico/modules/events/tracks/templates/management.html
```diff
@@ -57,7 +57,7 @@ <h3 class="f-self-stretch">{% trans %}List of tracks and track groups{% endtrans
         </a>
     </div>
 
-    <div id="track-list-container">
+    <div id="track-list-container" data-url="{{ url_for('.sort_tracks', event) }}">
```
ThiefMaster: can you also remove the two attributes that aren't needed anymore?

indico/modules/events/tracks/client/js/index.js
```diff
@@ -8,6 +8,34 @@
 (function(global) {
   'use strict';
 
+  function updateSorting(item) {
```
ThiefMaster: `item` is always `$('#track-list-container')` so I'd just use that directly 2 lines below and remove the arg

indico/modules/events/tracks/client/js/index.js
```diff
@@ -32,36 +60,12 @@
...
-                complete: IndicoUI.Dialogs.Util.progress(),
-                error: handleAjaxError,
-              });
+            // call update only once and only for the receiver
+            if (this === ui.item.parent()[0]) {
+              updateSorting($('#track-list-container'));
             }
           },
-          receive: function(event, ui) {
+          receive: function(_event, ui) {
```
ThiefMaster: why the `_event`?

indico/modules/events/models/events.py
```diff
@@ -681,6 +681,9 @@ def get_sorted_tracks(self):
         """Return tracks and track groups in the correct order"""
         track_groups = self.track_groups
         tracks = [track for track in self.tracks if not track.track_group]
+        for tg in track_groups:
+            if tg.tracks:
+                tg.tracks = sorted(tg.tracks, key=attrgetter('position'))
```
ThiefMaster: that's not a good idea, you modify the relationship's collection so it might actually emit UPDATEs on flush/commit. the proper option here is to add `order_by=position` below this line: https://github.com/indico/indico/blob/9838bc48f01c4f21c8902c0614f0971bd924fb4a/indico/modules/events/tracks/models/tracks.py#L99

afterwards the whole change in here can go away.

indico/modules/events/tracks/client/js/index.js
panagiotappl: Hm, I think that was done accidentally by the ide because event is an unused variable and I decided to keep it.

ThiefMaster: i'd rather use use `_` without anything after it in that case

indico/modules/events/tracks/client/js/index.js
```diff
@@ -8,6 +8,33 @@
 (function(global) {
   'use strict';
 
+  function updateSorting() {
+    const sortedList = $('#track-list-container')
```
ThiefMaster: `const container = $('#track-list-container');` and use that one here and below to get the URL

PR #4198, opened 2019-12-03
Improve process revisions

mic4ael:


indico/modules/events/editing/client/js/actions.js
```diff
@@ -5,41 +5,29 @@
...
 
 export const SET_LOADING = 'SET_LOADING';
 export const SET_DETAILS = 'SET_DETAILS';
 
-export function loadTimeline(eventId, contributionId, type) {
-  const url = editableDetailsURL({confId: eventId, contrib_id: contributionId, type});
-  return ajaxAction(() => indicoAxios.get(url), SET_LOADING, SET_DETAILS);
+export function loadTimeline() {
+  return async (dispatch, getStore) => {
+    const url = getStore().staticData.editableDetailsURL;
```
ThiefMaster: You can still use the selector here: `getStaticDate(getStore())`

indico/modules/events/editing/controllers/base.py
```diff
@@ -56,4 +56,4 @@ def _user_is_authorized_editor(self):
         if session.user.is_admin:
             # XXX: not sure if we want to keep this, but for now it's useful to have!
             return True
-        return self.editable.editor == session.user
+        return self.event.can_manage(session.user, permission='paper_editing')
```
ThiefMaster: You need to be the assigned editor to submit a review etc. so this isn't correct.

indico/modules/events/editing/client/js/actions.js
mic4ael: Ahh, I forgot to change it after I tested it (this was meant to make the whole thing testable)

PR #4197, opened 2019-12-04
Fix FileManager issues

ThiefMaster:
- Replacing a file and then uploading a new file in the same (multi-file) type deletes the file uploaded as a replacement.
- Replacing a freshly uploaded file in a single-file type results in both files disappearing, and the old file does not get deleted
- Replacing an existing file twice in a single-file type is broken
- Extensions / multiple aren't checked in the dropzones

We should ideally add some unit tests to cover those cases!

### DIFF
--- a/indico/modules/events/editing/client/js/__tests__/FileManager.spec.jsx
+++ b/indico/modules/events/editing/client/js/__tests__/FileManager.spec.jsx
@@ -28,19 +28,19 @@ const fileTypes = [
   {
     id: 1,
     name: 'Source file',
-    extensions: ['.txt', '.md', '.tex'],
+    extensions: ['txt', 'md', 'tex'],
     allowMultipleFiles: false,
   },
   {
     id: 2,
     name: 'PDF file',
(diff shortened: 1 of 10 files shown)


indico/modules/events/editing/client/js/__tests__/FileManager.spec.jsx
```diff
@@ -119,6 +140,108 @@ describe('File manager', () => {
...
+
+    // upload another file which should replace the previous one
+    const uuid2 = await simulateFileUpload(dropzone, 'test2.pdf', 'pdf');
+    expect(onChange).toHaveBeenCalledWith({'2': [uuid2]});
+    expect(mockAxios.delete).toHaveBeenCalledWith('/files/newtest1.pdf');
+    await act(async () => {
+      mockAxios.mockResponse(); // respond to the DELETE request
+    });
+    onChange.mockClear();
+    mockAxios.delete.mockClear();
```
pferreir: The two tests look pretty much the same up to here, but not sure how we can improve that.

PR #4211, opened 2019-12-05
Support editing comments in the UI

mic4ael:
editing_comments

### DIFF
--- a/indico/modules/events/editing/client/js/components/Timeline/CommentForm.jsx
+++ b/indico/modules/events/editing/client/js/components/Timeline/CommentForm.jsx
@@ -1,0 +1,104 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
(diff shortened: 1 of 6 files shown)


indico/modules/events/editing/client/js/components/Timeline/CommentForm.jsx
```diff
@@ -0,0 +1,104 @@
...
+  const InputComponent = commentFormVisible ? FinalTextArea : FinalInput;
+  const inputProps = commentFormVisible ? {autoFocus: true} : {};
+  const onCommentClickHandler = () => {
+    if (!commentFormVisible) {
+      setCommentFormVisible(true);
+      onToggleExpand(true);
+    }
+  };
+  const handleSubmit = async (formData, form) => {
+    const rv = await onSubmit(formData, form);
```
ThiefMaster: I don't think you are using `form` inside onSubmit

PR #4212, opened 2019-12-05
Use migra instead of apgdiff for schema diffs

pferreir:
This will make it much easier to get diffs of the DB schema. We could even consider adding this as a command under `indico db`. There are no external dependencies outside of PyPI.

Until https://github.com/djrobstep/schemainspect/pull/33 is merged, the script will always return a trigger function in addition to the actual differences (due to differences in line breaks).

### DIFF
--- a/bin/utils/db_diff.py
+++ b/bin/utils/db_diff.py
@@ -143,8 +143,13 @@ def main(dbname, verbose, reverse):
         base_conn = base_eng.connect()
         target_conn = target_eng.connect()
 
+        version = base_conn.execute("SELECT current_setting('server_version_num')::int").scalar()
+        if version < 100000:
+            click.echo(click.style('!! This utility requires at least Postgres 10', fg='red', bold=True), err=True)
+            sys.exit(1)
+
         if verbose:
-            click.echo(click.style('** Calculating differences', fg='magenta', bold=True), err=True)
+            click.echo(click.style('** Calculating differences', fg='cyan'), err=True)
         # use migra to figure out the SQL diff
         m = Migration(base_conn, target_conn) if reverse else Migration(target_conn, base_conn)
         m.set_safety(False)
--- a/bin/utils/db_diff.py
+++ b/bin/utils/db_diff.py
@@ -11,10 +11,11 @@ import os
 import pipes
 import subprocess
 import sys
-import tempfile
 
 import click
 from click._compat import should_strip_ansi
+from migra import Migration
+from sqlalchemy import create_engine
 
 
 click.disable_unicode_literals_warning = True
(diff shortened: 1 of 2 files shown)


bin/utils/db_diff.py
```diff
@@ -137,30 +126,42 @@ def main(dbname, verbose, apgdiff):
...
+        diff = m.sql
     finally:
+        # clean up connections and engines, so that no open connection remains
+        # (otherwise the DROPDB operation won't work)
+        if base_conn:
+            base_conn.close()
+            base_eng.dispose()
+        if target_conn:
+            target_conn.close()
+            base_eng.dispose()
```
ThiefMaster: shouldn't this be `target_eng`?

pferreir: Indeed, thanks!

bin/utils/db_diff.py
```diff
@@ -137,30 +126,42 @@ def main(dbname, verbose, apgdiff):
...
-    specify the path to `apgdiff` apgdiff or its .jar file.
     """
     temp_dbname = 'indico_dbdiff'
-    apgdiff_cmd = _get_apgdiff_cmd(apgdiff)
-    if not apgdiff_cmd:
-        raise click.exceptions.UsageError('Could not find apgdiff in PATH; specify the path to a script or the .jar '
-                                          'file manually.')
+    base_conn = None
+    target_conn = None
+    diff = ""
```
mic4ael: We are mixing `"` with `'` :P (sorry for nit-picking)

pferreir: Thanks! It's not nitpicking, it's a violation of our quotes policy 😉

ThiefMaster:
The latest version isn't compatible with Postgres 9.6 :(

djrobstep/migra#59

pferreir:
> The latest version of migra isn't compatible with Postgres 9.6 :( [djrobstep/migra#59](https://github.com/djrobstep/migra/issues/59)

Is that a big deal in what is mostly a development tool?

ThiefMaster:
Personally I liked having 9.6 locally to avoid bad surprises (like only realizing in production that something doesn't work on 9.6) - but I did upgrade on friday not too long after posting this comment since the "big" 2.3 changes are done and 3.0 will most likely require 11 anyway...

pferreir:
Yeah, I think it's a matter of time until this becomes an issue. I would say we can leave the PR open until https://github.com/djrobstep/schemainspect/pull/33 is solved and merge afterwards.

bin/utils/db_diff.py
```diff
@@ -137,30 +126,42 @@ def main(dbname, verbose, apgdiff):
...
-    specify the path to `apgdiff` apgdiff or its .jar file.
     """
     temp_dbname = 'indico_dbdiff'
-    apgdiff_cmd = _get_apgdiff_cmd(apgdiff)
-    if not apgdiff_cmd:
-        raise click.exceptions.UsageError('Could not find apgdiff in PATH; specify the path to a script or the .jar '
-                                          'file manually.')
+    base_conn = None
+    target_conn = None
+    diff = ''
```
ThiefMaster: this one isn't needed

PR #4213, opened 2019-12-05
Add tag input

pferreir:
In order to quickly migrate your DB:
```sql
alter table "event_editing"."tags" drop column "name";
alter table "event_editing"."tags" add column "code" character varying not null;
alter table "event_editing"."tags" add column "title" character varying not null;
CREATE UNIQUE INDEX ix_uq_tags_event_id_code_lower ON event_editing.tags USING btree (event_id, lower((code)::text));
```

If you want to add a few testing tags:

```python
TAGS = [
    ('ERR1', 'Missing bold title', 'red'),
    ('ERR2', 'Margin too small', 'red'),
    ('PRB1', 'Author list incomplete', 'purple'),
    ('MISC3', 'Some other error', 'teal')
]
for code, title, color in TAGS:
    EditingTag(code=code, title=title, color=color, event=your_test_event)
db.session.commit()
```

### DIFF
--- a/indico/modules/events/editing/client/js/components/Timeline/judgment/AcceptRejectForm.jsx
+++ b/indico/modules/events/editing/client/js/components/Timeline/judgment/AcceptRejectForm.jsx
@@ -15,17 +15,19 @@ import {FinalSubmitButton, FinalTextArea} from 'indico/react/forms';
 import {Translate} from 'indico/react/i18n';
 
 import {reviewEditable} from '../../../actions';
-import {getLastRevision} from '../../../selectors';
+import {getLastRevision, getStaticData} from '../../../selectors';
+import FinalTagInput from './TagInput';
 
 import './JudgmentBox.module.scss';
 
 export default function AcceptRejectForm({action, setLoading}) {
   const lastRevision = useSelector(getLastRevision);
+  const {tags: tagOptions} = useSelector(getStaticData);
   const dispatch = useDispatch();
 
   return (
     <FinalForm
-      initialValues={{comment: ''}}
+      initialValues={{comment: '', tags: []}}
       onSubmit={async formData => {
         setLoading(true);
         const rv = await dispatch(reviewEditable(lastRevision, {...formData, action}));
@@ -48,7 +50,7 @@ export default function AcceptRejectForm({action, setLoading}) {
               /* otherwise changing required doesn't work properly if the field has been touched */
               key={action}
             />
-            <div>TODO: Tags field</div>
+            <FinalTagInput name="tags" options={tagOptions} />
           </Form>
           <div styleName="judgment-submit-button">
             <FinalSubmitButton
--- a/indico/modules/events/editing/client/js/components/Timeline/judgment/AcceptRejectForm.jsx
+++ b/indico/modules/events/editing/client/js/components/Timeline/judgment/AcceptRejectForm.jsx
@@ -27,7 +27,7 @@ export default function AcceptRejectForm({action, setLoading}) {
 
   return (
     <FinalForm
-      initialValues={{comment: '', tags: []}}
+      initialValues={{comment: '', tags: lastRevision.tags}}
       onSubmit={async formData => {
         setLoading(true);
         const rv = await dispatch(reviewEditable(lastRevision, {...formData, action}));
--- a/indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
+++ b/indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
@@ -1,0 +1,79 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Dropdown} from 'semantic-ui-react';
+import {FinalField} from 'indico/react/forms/fields';
+import {Translate} from 'indico/react/i18n';
+
+import './TagInput.module.scss';
+
+const optionPropTypes = PropTypes.shape({
+  id: PropTypes.number.isRequired,
+  code: PropTypes.string.isRequired,
+  title: PropTypes.string.isRequired,
+  verboseTitle: PropTypes.string.isRequired,
+  color: PropTypes.string.isRequired,
+});
+
+/**
+ * A dropdown which allows tags to be searched and added
+ */
+function TagInput({onChange, value, placeholder, options}) {
+  return (
+    <Dropdown
+      placeholder={placeholder}
+      styleName="tag-input"
+      upward={false}
+      fluid
+      multiple
+      search
+      selection
+      value={value}
+      options={options.map(({id, verboseTitle, ...rest}) => ({
+        value: id,
+        text: verboseTitle,
+        key: id,
+        ...rest,
+      }))}
+      onChange={(_, {value: v}) => onChange(v)}
+      closeOnChange
+      renderLabel={({color, text}) => ({
+        color,
+        content: text,
+      })}
+    />
+  );
+}
+
+TagInput.propTypes = {
+  /** Called whenever the value changes (receives array of Strings) */
+  onChange: PropTypes.func.isRequired,
+  /** The default (initial) value */
+  value: PropTypes.arrayOf(PropTypes.number).isRequired,
+  /** The text placeholder which should should up in the input box */
+  placeholder: PropTypes.string,
+  /** The list of options to show as valid tags */
+  options: PropTypes.arrayOf(optionPropTypes).isRequired,
+};
+
+TagInput.defaultProps = {
+  placeholder: Translate.string('Tag code or title...'),
+};
+
+/** final-form-compatible <TagInput /> */
+export default function FinalTagInput({name, ...rest}) {
+  return <FinalField name={name} component={TagInput} {...rest} />;
+}
+
+FinalTagInput.propTypes = {
+  /** The field's name */
+  name: PropTypes.string.isRequired,
+  /** The list of options to show as valid tags */
+  options: PropTypes.arrayOf(optionPropTypes).isRequired,
+};
--- a/indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
+++ b/indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
@@ -35,11 +35,11 @@ function TagInput({onChange, value, placeholder, options}) {
       search
       selection
       value={value}
-      options={options.map(({id, verboseTitle, ...rest}) => ({
+      options={options.map(({id, verboseTitle, color}) => ({
         value: id,
         text: verboseTitle,
         key: id,
-        ...rest,
+        color,
       }))}
       onChange={(_, {value: v}) => onChange(v)}
       closeOnChange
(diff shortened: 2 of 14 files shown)


ThiefMaster:
That SQL won't work if there are already tags in the DB. you might want to rename name to title there instead ;)

indico/modules/events/editing/client/js/components/Timeline/judgment/AcceptRejectForm.jsx
```diff
@@ -48,7 +51,7 @@ export default function AcceptRejectForm({action, setLoading}) {
               /* otherwise changing required doesn't work properly if the field has been touched */
               key={action}
             />
-            <div>TODO: Tags field</div>
+            <FinalTagInput name="tags" options={_.sortBy(tagOptions, 'id')} />
```
ThiefMaster: I'd do the sorting inside the field

indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
```diff
@@ -0,0 +1,78 @@
...
+      defaultValue={defaultValue}
+      options={options.map(({id, code, title, ...rest}) => ({
+        value: id,
+        text: `${code}: ${title}`,
+        key: id,
+        ...rest,
+      }))}
+      onChange={(_, {value: v}) => onChange(v)}
+      closeOnChange
+      renderLabel={({color, text}) => {
```
ThiefMaster: `({color, text}) => ({color, content: text})`

indico/modules/events/editing/models/tags.py
```diff
@@ -31,10 +31,15 @@ def __table_args__(cls):
...
         nullable=False
     )
-    name = db.Column(
+    title = db.Column(
         db.String,
         nullable=False
     )
+    code = db.Column(
+        db.String,
+        index=True,
```
ThiefMaster: is that index really useful?

pferreir:
> That SQL won't work if there are already tags in the DB. you might want to rename name to title there instead ;)

Sure, but how many people will have tags in their DBs?

indico/modules/events/editing/client/js/index.jsx
```diff
@@ -42,6 +48,7 @@ document.addEventListener('DOMContentLoaded', async () => {
       contrib_id: contributionId,
       type: editableType,
     }),
+    tags,
```
ThiefMaster: more readable if you move it above the url

ThiefMaster:
I do ;)

    alter table "event_editing"."tags" rename column "name" to "title";

indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
```diff
@@ -0,0 +1,78 @@
...
+      styleName="tag-input"
+      upward={false}
+      fluid
+      multiple
+      search
+      selection
+      defaultValue={defaultValue}
+      options={options.map(({id, code, title, ...rest}) => ({
+        value: id,
+        text: `${code}: ${title}`,
```
ThiefMaster: what about adding a `verbose_title` to the schema for this? otherwise we'll probably have inconsistent styles (`: ` vs ` - `) at some point

pferreir: Good idea.

indico/modules/events/editing/models/tags.py
pferreir: IIRC I made the diff before I added that (?)

pferreir: My thought was that sorting by code might come in helpful in the future, when we implement the µS.

ThiefMaster: I guess those queries would be event-scoped so even if we queried a single tag (instead of just fetching them all which at least to me sounds more likely since a "get a single tag" api sounds like overkill considering that tags are small and there won't be hundreds in an event) there would be little benefit in having this extra index...

pferreir: OK, makes sense. I'll remove that.

indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
```diff
@@ -0,0 +1,77 @@
...
+      title: PropTypes.string.isRequired,
+      verboseTitle: PropTypes.string.isRequired,
+      color: PropTypes.string.isRequired,
+    })
+  ).isRequired,
+};
+
+TagInput.defaultProps = {
+  value: [],
+  placeholder: Translate.string('Tag ID or title...'),
```
ThiefMaster: s/ID/code/?

indico/modules/events/editing/client/js/components/Timeline/judgment/AcceptRejectForm.jsx
pferreir: Why?

indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
```diff
@@ -0,0 +1,77 @@
...
+};
+
+/** final-form-compatible <TagInput /> */
+export default function FinalTagInput({name, ...rest}) {
+  return <FinalField name={name} component={TagInput} {...rest} />;
+}
+
+FinalTagInput.propTypes = {
+  name: PropTypes.string.isRequired,
```
ThiefMaster: options should be here as well since that's always required

indico/modules/events/editing/schemas.py
```diff
@@ -39,7 +39,9 @@ class Meta:
 class EditingTagSchema(mm.ModelSchema):
     class Meta:
         model = EditingTag
-        fields = ('id', 'name', 'color', 'system')
+        fields = ('id', 'code', 'title', 'color', 'system', 'verbose_title')
+
+    verbose_title = fields.Function(lambda tag: '{}: {}'.format(tag.code, tag.title))
```
ThiefMaster: hm, property on the model vs this? if we need it in server-side code (notification emails?) at some point the property would be better

indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A dropdown which allows tags to be searched and added
+ */
+function TagInput({onChange, value: defaultValue, placeholder, options}) {
+  return (
+    <Dropdown
+      placeholder={placeholder}
+      styleName="tag-input"
+      upward={false}
```
ThiefMaster: isn't that the default anyway?

indico/modules/events/editing/client/js/components/Timeline/judgment/AcceptRejectForm.jsx
ThiefMaster: we probably always want to show it sorted (by code btw, not by id!), so having to do it whenever you use the field is a bit repetitive

ThiefMaster: another option would be just returning it sorted from the API (in the schema's post_dump function)... natural-order sorting would be easier there as well

indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
pferreir: Hm... true. ESLint doesn't complain about it because we're just using `...rest`.

indico/modules/events/editing/schemas.py
pferreir: Good point.

indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
pferreir: If I understood correctly, if it's not set the component tries to figure out whether to open upward or downward. Upward looks pretty ugly IMHO.

indico/modules/events/editing/client/js/components/Timeline/judgment/AcceptRejectForm.jsx
pferreir: Yeah, I think I prefer that last option.

indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
ThiefMaster: ah good point, i didn't know it was a tristate option

ThiefMaster: ah, good point, didn't know it was a tristate option

indico/modules/events/editing/schemas.py
```diff
@@ -39,12 +41,12 @@ class Meta:
...
         model = EditingTag
-        fields = ('id', 'name', 'color', 'system')
+        fields = ('id', 'code', 'title', 'color', 'system', 'verbose_title')
 
     @post_dump(pass_many=True)
-    def convert_to_dict(self, data, many, **kwargs):
+    def sort_list(self, data, many, **kwargs):
         if many:
-            data = {x['id']: x for x in data}
+            data = sorted(data, key=itemgetter('verbose_title'))
```
ThiefMaster: `natural_sort_key`

indico/modules/events/editing/client/js/components/Timeline/judgment/TagInput.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+TagInput.propTypes = {
+  /** Called whenever the value changes (receives array of Strings) */
+  onChange: PropTypes.func.isRequired,
+  /** The default (initial) value */
+  value: PropTypes.arrayOf(PropTypes.number),
+  /** The text placeholder which should should up in the input box */
+  placeholder: PropTypes.string,
+  /** The list of options to show as valid tags */
+  options: PropTypes.arrayOf(optionPropTypes).isRequired,
```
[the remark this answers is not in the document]
pferreir: I prefer "options" because "tags" can be confused with the actual tags that have been set. "Options" is the set of usable tags.

indico/modules/events/editing/schemas.py
pferreir: Indeed. Thanks!

ThiefMaster:
i pushed some improvements to the final-form integration (and rebased)

indico/modules/events/editing/models/tags.py
```diff
@@ -31,7 +31,11 @@ def __table_args__(cls):
         index=True,
         nullable=False
     )
-    name = db.Column(
+    title = db.Column(
+        db.String,
+        nullable=False
+    )
+    code = db.Column(
```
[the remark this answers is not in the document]
pferreir: Yes, I think it makes sense, otherwise using the full name may become hard to manage, right? Opinions? @mic4ael @meluru @panagiotappl

indico/web/client/js/jquery/extensions/global.js
```diff
@@ -48,7 +48,8 @@ $(document).ready(function() {
       var qtipHTML =
         qtipHTMLContainer && qtipHTMLContainer.length ? $(this).next(qtipHTMLContainer) : null;
 
-      if ((!qtipHTML && !title) || this.disabled || $target.data('no-auto-tooltip')) {
```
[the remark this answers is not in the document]
pferreir: Good point! Thanks!

PR #4215, opened 2019-12-10
Tag management

### LINKED ISSUE #4210: Paper Editing: Create tag setup interface
A simple interface to create/modify/delete tags.

mic4ael:
closes #4210


indico/modules/events/editing/models/tags.py
```diff
@@ -13,6 +13,11 @@
 from indico.util.string import format_repr, return_ascii
 
 
+tag_colors = [
```
ThiefMaster: What about putting this as `get_sui_colors()` in indico.web.forms.colors?

indico/modules/events/editing/controllers/backend.py
```diff
@@ -44,12 +48,44 @@ def _process(self):
...
+        'title': fields.String(),
+        'color': fields.String(validate=validate.OneOf(tag_colors)),
+    })
+    def _process_PATCH(self, args):
+        update_tag(self.tag, args)
+        return EditingTagSchema(many=True).jsonify(self.event.editing_tags)
+
+    def _process_DELETE(self):
+        delete_tag(self.tag)
+        return '', 201
```
ThiefMaster: 204

indico/modules/events/editing/models/tags.py
mic4ael: good idea!

indico/modules/events/editing/client/js/index.jsx
```diff
@@ -5,56 +5,5 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
```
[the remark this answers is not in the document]
mic4ael: Ah, forgot about it after moving the code to separate files

indico/modules/events/editing/client/js/components/TagManager.jsx
```diff
@@ -0,0 +1,175 @@
...
+      return {...state, isLoading: action.isLoading};
+    case 'TAGS_RECEIVED':
+      return {isLoading: false, tags: camelizeKeys(action.data)};
+    default:
+      return state;
+  }
+}
+
+export default function TagManager({eventId}) {
+  const [[tagModalVisible, initialValues], setTagModalVisible] = useState([false, undefined]);
```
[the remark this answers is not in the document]
mic4ael: Hm, honestly I don't see why it would be "ugly" :P

indico/modules/events/editing/client/js/components/TagManager.module.scss
```diff
@@ -0,0 +1,31 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+.tags-container {
+    max-width: 600px !important;
```
ThiefMaster: max-width vs width? Might look weird if there's only a very short tag and then the whole thing shrinks a lot

indico/modules/events/editing/client/js/components/TagManager.module.scss
```diff
@@ -0,0 +1,31 @@
...
+        .tag-actions {
+            width: 45px;
+            display: flex;
+            align-items: center;
+            justify-content: space-between;
+
+            i:global(.icon) {
+                margin-right: 0;
+
+                &:hover {
```
ThiefMaster: what's the benefit of setting this on hover? you don't *have* a cursor unless you hover it :D

indico/modules/events/editing/client/js/components/TagModal.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Form as FinalForm} from 'react-final-form';
+import {Button, Form, Label, Modal} from 'semantic-ui-react';
+
+import {FinalDropdown, FinalInput, FinalSubmitButton} from 'indico/react/forms';
+import {Translate} from 'indico/react/i18n';
+
+const colors = {
```
ThiefMaster: What about putting this in a `SUIPalette` or `Palette.SUI` object in `indico/web/client/js/utils/palette.js`?

indico/modules/events/editing/controllers/backend.py
```diff
@@ -14,26 +14,31 @@
 from flask import request, session
 from marshmallow import fields
 from marshmallow_enum import EnumField
+from webargs import validate
```
ThiefMaster: Import this from marshmallow; webargs just re-exports it

indico/modules/events/editing/controllers/backend.py
```diff
@@ -44,12 +49,44 @@ def _process(self):
 
 
 class RHEditingTags(RHEditingBase):
-    """Return all editing tags defined in the event."""
+    """Return all editing tags defined in the event or create a new tag."""
```
ThiefMaster: not the case anymore

indico/modules/events/editing/controllers/backend.py
```diff
@@ -44,12 +49,44 @@ def _process(self):
...
 
     def _process(self):
         return EditingTagSchema(many=True).jsonify(self.event.editing_tags)
 
 
+class RHCreateTag(RHEditingManagementBase):
```
ThiefMaster: You need to check for unique codes here and when editing. Check e.g. `RHEquipmentTypes` (the admin one) for an example how to properly return the error in the same style webargs would report a failure so it shows up on the correct field in the client.

indico/modules/events/editing/templates/editing.html
```diff
@@ -5,5 +5,22 @@
...
-    <div class="page-content"></div>
+    <div class="page-content">
+        <div class="action-box">
+            <div class="section">
+                <span class="icon icon-tag" style="align-self: center;"></span>
+                <div class="text">
+                    <div class="label">
+                        {% trans %}Editing tags{% endtrans %}
+                    </div>
+                    {% trans %}Configure editable tags{% endtrans %}
```
ThiefMaster: I don't think we need to mention "editable" and "editing" here. We are in the management areas of the Editing module, so it's obvious.

indico/modules/events/editing/templates/tags.html
```diff
@@ -0,0 +1,9 @@
+{% extends 'events/management/base.html' %}
+
+{% block title %}
+    Tag management
```
ThiefMaster: i18n

indico/modules/events/editing/controllers/frontend.py
```diff
@@ -31,3 +31,8 @@ def _process(self):
             editable=self.editable,
             contribution=self.contrib
         )
+
+
+class RHEditingTags(RHEditingBase):
+    def _process(self):
+        return WPEditing.render_template('tags.html', self.event)
```
ThiefMaster: What about `management/tags.html` (and `management/editing.html` in `RHEditingDashboard`)?

indico/modules/events/editing/client/js/components/TagManager.jsx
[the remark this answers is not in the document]
ThiefMaster: I still think using a dispatch with these actions would make this much nicer and more readable:

- `{type: 'ADD_TAG'}`
- `{type: 'EDIT_TAG', tag}`
- `{type: 'HIDE'}`

indico/modules/events/editing/client/js/components/TagManager.jsx
```diff
@@ -0,0 +1,164 @@
...
+  if (isLoadingTags) {
+    return <Loader inline="centered" active />;
+  }
+
+  return (
+    <div styleName="tags-container">
+      {tags && (
+        <>
+          {tags.map(tag => (
+            <Segment key={tag.id} styleName="tag-segment">
```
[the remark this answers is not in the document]
mic4ael: Yeah, that is why I just left it as it is.

indico/modules/events/editing/client/js/components/TagManager.jsx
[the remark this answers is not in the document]
mic4ael: Really.. then what did you mean by > This feels a bit ugly to me.. and then > maybe better to simply use two useStates

indico/modules/events/editing/client/js/components/TagManager.jsx
```diff
@@ -0,0 +1,164 @@
...
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+import RequestConfirm from './RequestConfirm';
+import TagModal from './TagModal';
+
+import './TagManager.module.scss';
+
+export default function TagManager({eventId}) {
+  const [[tagModalVisible, initialValues], setTagModalVisible] = useState([false, undefined]);
+  const [deletedTag, setDeletedTag] = useState(null);
+  const [isLoadingTags, setIsLoadingTags] = useState(false);
```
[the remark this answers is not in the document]
mic4ael: I think that `setIsLoadingTags` is a bit clearer and I don't see any advantage of `setLoadingTags` :P

indico/modules/events/editing/client/js/components/TagManager.jsx
```diff
@@ -0,0 +1,164 @@
...
+  const [tags, setTags] = useState(null);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
mic4ael: Nice, that might be actually useful :D I was thinking about having something like that and it turns out that there is already a nice hook

indico/modules/events/editing/controllers/backend.py
```diff
@@ -44,12 +49,44 @@ def _process(self):
...
+        self.tag = EditingTag.query.with_parent(self.event).filter_by(id=request.view_args['tag_id']).first_or_404()
+
+    @use_args({
+        'code': fields.String(),
+        'title': fields.String(),
+        'color': fields.String(validate=validate.OneOf(get_sui_colors())),
+    })
+    def _process_PATCH(self, args):
+        update_tag(self.tag, args)
+        return EditingTagSchema(many=True).jsonify(self.event.editing_tags)
```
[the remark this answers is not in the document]
mic4ael: Yeah, I agree. I just did it so that the frontend doesn't have to perform operations on the internal state. However I think that in the end it doesn't hurt and I will just return the updated object from this RH

indico/modules/events/editing/client/js/components/TagManager.jsx
```diff
@@ -0,0 +1,164 @@
...
+          }
+          onSubmit={async formData => {
+            let rv;
+            if (initialValues) {
+              rv = await editTag(initialValues.id, formData);
+            } else {
+              rv = await createTag(formData);
+            }
+
+            return rv;
```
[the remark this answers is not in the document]
mic4ael: I did test it exactly like that and it worked with just `return rv`

indico/modules/events/editing/client/js/components/TagManager.module.scss
mic4ael: What do you mean? I used `max-width` so that it is a bit more responsive with smaller screens

indico/modules/events/editing/client/js/components/TagModal.jsx
mic4ael: Nice, didn't notice this file

indico/modules/events/editing/client/js/components/TagManager.jsx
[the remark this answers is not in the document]
mic4ael: So, it won't work since `handleSubmitError` returns `{[FORM_ERROR: ...}`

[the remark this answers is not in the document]
mic4ael: it is only `submitFormAction` which returns `{error: {[FORM_ERROR]...`

indico/modules/events/editing/client/js/components/TagManager.jsx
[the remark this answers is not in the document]
ThiefMaster: already explained offline, but for the sake of not leaving it unanswered here: I haven't had a closer look at the rest of the code the first comment, especially not at the places where we're actually using `setTagModalVisible`

PR #4216, opened 2019-12-10
FileType manager interface

### LINKED ISSUE #4120: Paper Editing: create management interface to set up file types

panagiotappl:
image

TODO

- [x] Fetch file types inside the file type manager component and not inside `index.jsx`

closes #4120

### DIFF
--- a/indico/modules/events/editing/client/js/fileTypes.jsx
+++ b/indico/modules/events/editing/client/js/fileTypes.jsx
@@ -1,0 +1,20 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import FileTypeManager from './components/FileTypeManager';
+
+document.addEventListener('DOMContentLoaded', async () => {
+  const fileTypeManager = document.querySelector('#editing-filetypes');
+  if (!fileTypeManager) {
+    return null;
+  }
+  const eventId = parseInt(fileTypeManager.dataset.eventId, 10);
+  ReactDOM.render(<FileTypeManager eventId={eventId} />, fileTypeManager);
+});
--- a/indico/modules/events/editing/controllers/frontend.py
+++ b/indico/modules/events/editing/controllers/frontend.py
@@ -9,7 +9,8 @@ from __future__ import unicode_literals
 
 from werkzeug.exceptions import NotFound
 
-from indico.modules.events.editing.controllers.base import RHContributionEditableBase, RHEditingBase
+from indico.modules.events.editing.controllers.base import (RHContributionEditableBase, RHEditingBase,
+                                                            RHEditingManagementBase)
 from indico.modules.events.editing.views import WPEditing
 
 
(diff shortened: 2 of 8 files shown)


indico/modules/events/editing/controllers/frontend.py
```diff
@@ -36,3 +36,8 @@ def _process(self):
 class RHEditingTags(RHEditingBase):
     def _process(self):
         return WPEditing.render_template('management/tags.html', self.event)
+
+
+class RHEditingFileTypes(RHEditingBase):
```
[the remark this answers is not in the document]
ThiefMaster: And maybe rename this class (and the one for tags above as well) to `RHManageEditing***` to distinguish more between the management RH and the (public) api to list these things?

indico/modules/events/editing/client/js/components/FileTypeInterface.module.scss
```diff
@@ -0,0 +1,40 @@
...
+    .filetype h3 {
+        margin-bottom: 0;
+    }
+
+    .file-extension {
+        font-style: italic;
+        color: $gray;
+    }
+
+    .file-extension::after {
```
ThiefMaster: Can't you use `:not(:last-child)` here and then avoid the next rule altogether?

indico/modules/events/editing/controllers/frontend.py
```diff
@@ -33,6 +34,11 @@ def _process(self):
         )
 
 
-class RHEditingTags(RHEditingBase):
+class RHManageEditingTags(RHEditingBase):
```
ThiefMaster: Inherit from `RHEditingManagementBase`

indico/modules/events/editing/client/js/timeline.jsx
```diff
@@ -18,6 +18,7 @@ import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
 
 import storeFactory from './store';
 import Timeline from './components/Timeline';
+import FileTypeInterface from './components/FileTypeInterface';
```
ThiefMaster: not needed

indico/modules/events/editing/client/js/fileTypes.jsx
```diff
@@ -0,0 +1,21 @@
...
+// Copyright (C) 2002 - 2019 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import FileTypeInterface from './components/FileTypeInterface';
```
ThiefMaster: What about `FileTypeManager` for consistency (since there's also `TagManager`)?

indico/modules/events/editing/client/js/fileTypes.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+document.addEventListener('DOMContentLoaded', async () => {
+  const fileTypeInterface = document.querySelector('#editing-filetypes');
+  const eventId = parseInt(fileTypeInterface.dataset.eventId, 10);
```
ThiefMaster: move this after the if check since it'll fail if `fileTypeInterface` doesn't exist

PR #4229, opened 2020-01-08
Add backend to edit/delete file types

### LINKED ISSUE #4206: File Type Manager: add action to delete a type
Should just be an action icon which deletes the type when clicked.

- [x] Backend to delete a type (like in tags)
- [x] Frontend to delete a type (with confirmation prompt, like in tags)
- [x] Disallow deleting types that have files

ParthS007:
for #4206

### DIFF
--- a/indico/modules/events/editing/blueprint.py
+++ b/indico/modules/events/editing/blueprint.py
@@ -26,6 +26,8 @@ _bp.add_url_rule('/contributions/<int:contrib_id>/editing/<any(paper):type>/<int
 
 # Event-level APIs
 _bp.add_url_rule('/editing/api/file-types', 'api_file_types', backend.RHEditingFileTypes)
+_bp.add_url_rule('/editing/api/file-types/<int:file_type_id>', 'api_edit_file_type', backend.RHEditFileType,
+                 methods=('PATCH', 'DELETE'))
 _bp.add_url_rule('/editing/api/tags', 'api_tags', backend.RHEditingTags)
 _bp.add_url_rule('/editing/api/tags', 'api_create_tag', backend.RHCreateTag, methods=('POST',))
 _bp.add_url_rule('/editing/api/tag/<int:tag_id>', 'api_edit_tag', backend.RHEditTag, methods=('PATCH', 'DELETE'))
--- a/indico/modules/events/editing/controllers/backend.py
+++ b/indico/modules/events/editing/controllers/backend.py
@@ -21,17 +21,18 @@ from indico.modules.events.editing.controllers.base import (RHContributionEditab
                                                             RHEditingManagementBase)
 from indico.modules.events.editing.fields import EditingFilesField, EditingTagsField
 from indico.modules.events.editing.models.comments import EditingRevisionComment
+from indico.modules.events.editing.models.file_types import EditingFileType
 from indico.modules.events.editing.models.revision_files import EditingRevisionFile
 from indico.modules.events.editing.models.revisions import EditingRevision
 from indico.modules.events.editing.models.tags import EditingTag
 from indico.modules.events.editing.operations import (confirm_editable_changes, create_new_editable, create_new_tag,
                                                       create_revision_comment, create_submitter_revision,
-                                                      delete_revision_comment, delete_tag, replace_revision,
-                                                      review_editable_revision, undo_review, update_revision_comment,
-                                                      update_tag)
-from indico.modules.events.editing.schemas import (EditableSchema, EditableTagArgs, EditingConfirmationAction,
-                                                   EditingFileTypeSchema, EditingReviewAction, EditingTagSchema,
-                                                   ReviewEditableArgs)
+                                                      delete_file_type, delete_revision_comment, delete_tag,
+                                                      replace_revision, review_editable_revision, undo_review,
+                                                      update_file_type, update_revision_comment, update_tag)
+from indico.modules.events.editing.schemas import (EditableFileTypeArgs, EditableSchema, EditableTagArgs,
+                                                   EditingConfirmationAction, EditingFileTypeSchema,
+                                                   EditingReviewAction, EditingTagSchema, ReviewEditableArgs)
 from indico.modules.files.controllers import UploadFileMixin
 from indico.util.fs import secure_filename
 from indico.util.i18n import _
@@ -80,6 +81,26 @@ class RHEditTag(RHEditingManagementBase):
 
     def _process_DELETE(self):
         delete_tag(self.tag)
+        return '', 204
+
+
+class RHEditFileType(RHEditingManagementBase):
+    """Delete/update a file type."""
+
+    def _process_args(self):
+        RHEditingManagementBase._process_args(self)
+        self.file_type = (EditingFileType.query
+                          .with_parent(self.event)
+                          .filter_by(id=request.view_args['file_type_id'])
+                          .first_or_404())
+
+    @use_rh_args(EditableFileTypeArgs, partial=True)
+    def _process_PATCH(self, data):
+        update_file_type(self.file_type, **data)
+        return EditingFileTypeSchema().jsonify(self.file_type)
+
+    def _process_DELETE(self):
+        delete_file_type(self.file_type)
         return '', 204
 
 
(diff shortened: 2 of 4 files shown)


indico/modules/events/editing/blueprint.py
```diff
@@ -26,6 +26,7 @@
 
 # Event-level APIs
 _bp.add_url_rule('/editing/api/file-types', 'api_file_types', backend.RHEditingFileTypes)
+_bp.add_url_rule('/editing/api/file-types/<int:file_type_id>', 'api_edit_file_type', backend.RHEditFileType, methods=('PATCH', 'DELETE'))
```
ThiefMaster: line is too long

indico/modules/events/editing/controllers/backend.py
```diff
@@ -24,12 +24,13 @@
...
 from indico.modules.events.editing.models.tags import EditingTag
+from indico.modules.events.editing.models.file_types import EditingFileType
 from indico.modules.events.editing.operations import (confirm_editable_changes, create_new_editable, create_new_tag,
                                                       create_revision_comment, create_submitter_revision,
-                                                      delete_revision_comment, delete_tag, replace_revision,
+                                                      delete_revision_comment, delete_tag, delete_file_type, replace_revision,
                                                       review_editable_revision, undo_review, update_revision_comment,
-                                                      update_tag)
-from indico.modules.events.editing.schemas import (EditableSchema, EditableTagArgs, EditingConfirmationAction,
+                                                      update_tag, update_file_type)
```
ThiefMaster: `isort indico/modules/events/editing/controllers/backend.py `

indico/modules/events/editing/controllers/backend.py
```diff
@@ -82,6 +83,25 @@ def _process_DELETE(self):
         delete_tag(self.tag)
         return '', 204
```
ThiefMaster: 2 empty lines between top-level classes/functions

indico/modules/events/editing/controllers/backend.py
```diff
@@ -82,6 +83,25 @@ def _process_DELETE(self):
...
 
+class RHEditFileType(RHEditingManagementBase):
+    """Delete/update a file type."""
+
+    def _process_args(self):
+        RHEditingManagementBase._process_args(self)
+        self.file_type = (EditingFileType.query
+                    .with_parent(self.event)
```
ThiefMaster: align the `.` below the `E` (also in the next two lines)

indico/modules/events/editing/operations.py
```diff
@@ -229,3 +229,18 @@ def update_tag(tag, code=None, title=None, color=None):
 def delete_tag(tag):
     logger.info('Tag %r deleted by %r', tag, session.user)
     db.session.delete(tag)
+
+
+def update_file_type(file_type, name=None, extensions=None):
```
ThiefMaster: needs to handle `allow_multiple_files`, `required` and `publishable` as well

indico/modules/events/editing/schemas.py
```diff
@@ -162,3 +162,20 @@ def _check_for_unique_tag_code(self, code):
             query = query.filter(EditingTag.id != tag.id)
         if query.has_rows():
             raise ValidationError(_('Tag code must be unique'))
+
+class EditableFileTypeArgs(mm.Schema):
+    class Meta:
+        rh_context = ('event', 'file_type')
+
+    name = fields.String(required=True, validate=not_empty)
+    extension = fields.String(required=True, validate=not_empty)
```
ThiefMaster: that's a list of extensions -> `fields.List(...)`

and it may be empty, so no need for any validate

indico/modules/events/editing/blueprint.py
```diff
@@ -26,6 +26,8 @@
 
 # Event-level APIs
 _bp.add_url_rule('/editing/api/file-types', 'api_file_types', backend.RHEditingFileTypes)
+_bp.add_url_rule('/editing/api/file-types/<int:file_type_id>',
+                 'api_edit_file_type', backend.RHEditFileType, methods=('PATCH', 'DELETE'))
```
ThiefMaster: the endpoint fits nicely in the previous line - we use up to 120 chars per line, so no need to wrap that early

indico/modules/events/editing/schemas.py
```diff
@@ -162,3 +162,24 @@ def _check_for_unique_tag_code(self, code):
...
+
+    @validates('name')
+    def _check_for_unique_file_type_name(self, name):
+        event = self.context['event']
+        file_type = self.context['file_type']
+        query = EditingFileType.query.with_parent(event).filter(func.lower(EditingFileType.name) == name.lower())
+        if file_type:
+            query = query.filter(EditingFileType.id != file_type.id)
+        if query.has_rows():
+            raise ValidationError(_('File type name must be unique'))
```
ThiefMaster: Just "Name must be unique" sounds better IMO

indico/modules/events/editing/schemas.py
```diff
@@ -162,3 +162,24 @@ def _check_for_unique_tag_code(self, code):
...
             raise ValidationError(_('Tag code must be unique'))
+
+
+class EditableFileTypeArgs(mm.Schema):
+    class Meta:
+        rh_context = ('event', 'file_type')
+
+    name = fields.String(required=True, validate=not_empty)
+    extension = fields.List(required=True)
+    allow_multiple_files = fields.Boolean(required=True, validate=not_empty)
```
ThiefMaster: you do not need those validates; most likely having those breaks the case of sending `false` because `not_empty` checks for `if not value`...

indico/modules/events/editing/schemas.py
```diff
@@ -162,3 +162,24 @@ def _check_for_unique_tag_code(self, code):
...
         if query.has_rows():
             raise ValidationError(_('Tag code must be unique'))
+
+
+class EditableFileTypeArgs(mm.Schema):
+    class Meta:
+        rh_context = ('event', 'file_type')
+
+    name = fields.String(required=True, validate=not_empty)
+    extension = fields.List(required=True)
```
ThiefMaster: specify the type of the list members: `fields.List(fields.String(), required=True)`

indico/modules/events/editing/operations.py
```diff
@@ -229,3 +229,24 @@ def update_tag(tag, code=None, title=None, color=None):
 def delete_tag(tag):
     logger.info('Tag %r deleted by %r', tag, session.user)
     db.session.delete(tag)
+
+
+def update_file_type(file_type, name=None, extensions=None, allow_multiple_files=None, required=None, publishable=None):
```
ThiefMaster: Maybe using kwargs would be better here...

```python
def update_file_type(file_type, **data):
    file_type.update_from_dict(data, keys=('name', 'extensions', ...))
```

(the `keys=...` list is not required, but I'd like to make it somewhat clear when looking at the code which kwargs are accepted)

ParthS007: I have to create `update_from_dict `in the `EditingFileType` model then.

ThiefMaster: No, it's part of `IndicoModel` which all models inherit from since `db.Model` uses it.

ThiefMaster: Oh, I meant `populate_from_dict`, not `update_from_dict`...

ParthS007:
I have checked both `delete` and `patch` requests with http. Its working

indico/modules/events/editing/schemas.py
```diff
@@ -162,3 +162,24 @@ def _check_for_unique_tag_code(self, code):
...
+
+class EditableFileTypeArgs(mm.Schema):
+    class Meta:
+        rh_context = ('event', 'file_type')
+
+    name = fields.String(required=True, validate=not_empty)
+    extension = fields.List(fields.String(), required=True)
+    allow_multiple_files = fields.Boolean(required=True)
+    required = fields.Boolean(required=True)
+    publishable = fields.Boolean(required=True)
```
ThiefMaster: `extension` instead of `extensions`, and the list items may not be empty.

Also, with the changes below this schema will make the API more flexible once it's also used for creating new file types (since only the name will be required then).

```python
    extensions = fields.List(fields.String(validate=not_empty))
    allow_multiple_files = fields.Boolean()
    required = fields.Boolean()
    publishable = fields.Boolean()
```

ParthS007: okay, got it :+1:

PR #4232, opened 2020-01-09
Handle file upload errors in FileManager

### LINKED ISSUE #4203: FileManager: handle upload errors
Make it resilient to upload errors. The whole thing shouldn't blow up when an upload fails.

mic4ael:
closes #4203

### DIFF
--- a/indico/modules/events/editing/client/js/components/FileManager/reducer.js
+++ b/indico/modules/events/editing/client/js/components/FileManager/reducer.js
@@ -108,6 +108,18 @@ export default combineReducers({
         };
       }
 
+      case actions.UPLOAD_ERROR: {
+        const fileType = state[action.fileTypeId];
+        const tmpFile = fileType[action.tmpFileId];
+        return {
+          ...state,
+          [action.fileTypeId]: {
+            ...fileType,
+            [action.tmpFileId]: {...tmpFile, failed: true},
+          },
+        };
+      }
+
       default:
         return state;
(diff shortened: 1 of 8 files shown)


indico/modules/events/editing/client/js/components/FileManager/util.js
```diff
@@ -66,14 +66,20 @@ export function uploadFiles(action, fileTypeId, acceptedFiles, uploadURL, dispat
...
-      dispatch(
-        action(fileTypeId, fileId, tmpFileId, {
-          filename: uploadedFile.filename,
-          uuid: uploadedFile.uuid,
-          claimed: false,
-          fileType: fileTypeId,
-        })
-      );
+
+      if (uploadedFile === undefined) {
```
ThiefMaster: I'd rather make the function return `null` and check for that, instead of using an implicit `undefined` return value.

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -119,6 +131,7 @@ export default combineReducers({
       case actions.MARK_DELETED:
       case actions.UNDELETE:
       case actions.REVERT:
+      case actions.UPLOAD_ERROR:
```
ThiefMaster: probably not needed since we use `isDirty` to trigger the `useEffect` that calls `onChange`, and an error does not change the value

Issue #4207, opened 2019-12-04
File Type Manager: add form to edit a type


pferreir:
It should be based on the form created in #4204 and allow the attributes to be changed.

Issue #4204, opened 2019-12-04
File Type Manager: add form to add a new type


pferreir:
Fields which we will need:
 * Name of the type;
 * List of extensions;
 * Whether it's a "multi-file" type (`allow_multiple_files`);
 * Whether it's required;

PR #4238, opened 2020-01-13
File types: create, edit, delete 

meluru:
Closes: #4204
Closes: #4206
Closes: #4207

image

image

### DIFF
--- a/indico/modules/events/editing/client/js/components/ExtensionList.jsx
+++ b/indico/modules/events/editing/client/js/components/ExtensionList.jsx
@@ -5,6 +5,7 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
+import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import {Dropdown} from 'semantic-ui-react';
@@ -12,7 +13,7 @@ import {Translate} from 'indico/react/i18n';
 
 export default function ExtensionList({value, disabled, onChange, onFocus, onBlur}) {
   const handleChange = (e, {value: newValue}) => {
-    onChange(newValue);
+    onChange(_.uniq(newValue.map(x => x.trim().replace(/^[*.]+/, ''))));
     onFocus();
     onBlur();
   };
--- a/indico/modules/events/editing/client/js/components/ExtensionList.jsx
+++ b/indico/modules/events/editing/client/js/components/ExtensionList.jsx
@@ -1,0 +1,45 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2020 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Dropdown} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+
+export default function ExtensionList({value, disabled, onChange, onFocus, onBlur}) {
+  const handleChange = (e, {value: newValue}) => {
+    onChange(newValue);
+    onFocus();
+    onBlur();
+  };
+
+  return (
+    <Dropdown
+      options={value.map(x => ({text: x, value: x}))}
+      value={value}
+      disabled={disabled}
+      search
+      selection
+      multiple
+      allowAdditions
+      fluid
+      icon={null}
+      closeOnChange
+      noResultsMessage={Translate.string('Enter an extension')}
+      additionLabel={Translate.string('Add extension') + ' '} // eslint-disable-line prefer-template
+      onChange={handleChange}
+    />
+  );
+}
+
+ExtensionList.propTypes = {
+  value: PropTypes.arrayOf(PropTypes.string).isRequired,
+  disabled: PropTypes.bool.isRequired,
+  onChange: PropTypes.func.isRequired,
+  onFocus: PropTypes.func.isRequired,
+  onBlur: PropTypes.func.isRequired,
+};
--- a/indico/modules/events/editing/client/js/components/FileTypeManager.jsx
+++ b/indico/modules/events/editing/client/js/components/FileTypeManager.jsx
@@ -6,21 +6,82 @@
 // LICENSE file for more details.
 
 import fileTypesURL from 'indico-url:event_editing.api_file_types';
+import createFileTypeURL from 'indico-url:event_editing.api_add_file_type';
+import editFileTypeURL from 'indico-url:event_editing.api_edit_file_type';
 
-import React from 'react';
+import React, {useReducer} from 'react';
 import PropTypes from 'prop-types';
 import {Button, Icon, Loader, Message, Segment} from 'semantic-ui-react';
-import {Translate} from 'indico/react/i18n';
+import {Param, Translate} from 'indico/react/i18n';
+import {getChangedValues, handleSubmitError} from 'indico/react/forms';
 import {useIndicoAxios} from 'indico/react/hooks';
+import {handleAxiosError, indicoAxios} from 'indico/utils/axios';
+import {snakifyKeys} from 'indico/utils/case';
+import FileTypeModal from './FileTypeModal';
+import RequestConfirm from './RequestConfirm';
 
 import './FileTypeManager.module.scss';
 
+const initialState = {
+  fileType: null,
+  operation: null,
+};
+
+function fileTypesReducer(state, action) {
+  switch (action.type) {
+    case 'ADD_FILE_TYPE':
+      return {operation: 'add', fileType: null};
+    case 'EDIT_FILE_TYPE':
+      return {operation: 'edit', fileType: action.fileType};
+    case 'DELETE_FILE_TYPE':
+      return {operation: 'delete', fileType: action.fileType};
+    case 'CLEAR':
+      return {...initialState};
+    default:
+      return state;
+  }
+}
+
 export default function FileTypeManager({eventId}) {
-  const {data, loading: isLoadingFileTypes, lastData} = useIndicoAxios({
+  const [state, dispatch] = useReducer(fileTypesReducer, initialState);
+
+  const {data, loading: isLoadingFileTypes, reFetch, lastData} = useIndicoAxios({
     url: fileTypesURL({confId: eventId}),
     camelize: true,
     trigger: eventId,
   });
+
+  const createFileType = async formData => {
+    try {
+      await indicoAxios.post(createFileTypeURL({confId: eventId}), snakifyKeys(formData));
+      reFetch();
+    } catch (e) {
+      return handleSubmitError(e);
+    }
+  };
+
+  const editFileType = async (fileTypeId, fileTypeData) => {
+    const url = editFileTypeURL({confId: eventId, file_type_id: fileTypeId});
+
+    try {
+      await indicoAxios.patch(url, snakifyKeys(fileTypeData));
+      reFetch();
+    } catch (e) {
+      return handleSubmitError(e);
+    }
+  };
+
+  const deleteFileType = async fileTypeId => {
+    const url = editFileTypeURL({confId: eventId, file_type_id: fileTypeId});
+
+    try {
+      await indicoAxios.delete(url);
+      reFetch();
+    } catch (e) {
+      handleAxiosError(e);
+      return true;
+    }
+  };
 
   const fileTypes = data || lastData;
   if (isLoadingFileTypes && !lastData) {
@@ -29,6 +90,7 @@ export default function FileTypeManager({eventId}) {
     return null;
   }
 
+  const {operation, fileType: currentFileType} = state;
   return (
     <div styleName="file-types-container">
       {fileTypes.length === 0 && (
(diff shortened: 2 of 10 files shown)


indico/modules/events/editing/operations.py
```diff
@@ -231,6 +232,14 @@ def delete_tag(tag):
     db.session.delete(tag)
 
 
+def create_new_file_type(event, name, extensions=[], allow_multiple_files=False, required=False, publishable=False):
```
ThiefMaster: never put mutable default values

indico/modules/events/editing/operations.py
```diff
@@ -231,6 +232,14 @@ def delete_tag(tag):
...
+    file_type = EditingFileType(name=name, extensions=extensions, allow_multiple_files=allow_multiple_files,
```
ThiefMaster: What about `= EditingFileType()` and then using `populate_from_dict` like below when editing?

indico/modules/events/editing/client/js/components/ExtensionList.jsx
```diff
@@ -0,0 +1,41 @@
...
+      disabled={disabled}
+      search
+      selection
+      multiple
+      allowAdditions
+      fluid
+      icon={null}
+      closeOnChange
+      noResultsMessage={Translate.string('Enter an extension')}
+      additionLabel={Translate.string('Add extension ')} // eslint-disable-line prefer-template
```
ThiefMaster: `Translate.string('Add extension') + ' '`

indico/modules/events/editing/client/js/components/FileTypeModal.jsx
```diff
@@ -0,0 +1,104 @@
...
+        >
+          <Modal.Header content={header} />
+          <Modal.Content>
+            <Form id="file-type-form" onSubmit={fprops.handleSubmit}>
+              <FinalInput name="name" label={Translate.string('Name')} required />
+              <FinalField
+                name="extensions"
+                component={ExtensionList}
+                label={Translate.string('Extensions')}
+                isEqual={_.isEqual}
```
ThiefMaster: You probably want `isEqual={unsortedArraysEqual}` here since the file type list isn't sorted so changing the order (e.g. by deleting in the middle and re-adding) doesn't need to mark the form as dirty.

indico/modules/events/editing/schemas.py
```diff
@@ -38,7 +38,15 @@ class RevisionState(mm.Schema):
...
         model = EditingFileType
-        fields = ('id', 'name', 'extensions', 'allow_multiple_files', 'required', 'publishable')
+        fields = ('id', 'name', 'extensions', 'allow_multiple_files', 'required', 'publishable', 'is_used')
+
+    is_used = fields.Function(lambda file_type: EditingRevisionFile.query.with_parent(file_type).has_rows())
+
+    @post_dump(pass_many=True)
+    def sort_list(self, data, many, **kwargs):
+        if many:
+            data = sorted(data, key=lambda e: natural_sort_key(e['name']))
```
ThiefMaster: What does `e` stand for?

I'd go for `x` as a generic name, or `ft` for filetype

indico/modules/events/editing/operations.py
```diff
@@ -231,6 +232,14 @@ def delete_tag(tag):
     db.session.delete(tag)
 
 
+def create_new_file_type(event, **data):
+    file_type = EditingFileType(event=event)
+    file_type.populate_from_dict(data)
```
ThiefMaster: I'd specify the list of `keys` like in `update_file_type` to make it clear what args this function accepts when looking at its code.

indico/modules/events/editing/client/js/components/ExtensionList.jsx
```diff
@@ -0,0 +1,41 @@
+import React, {useState} from 'react';
+import PropTypes from 'prop-types';
+import {Dropdown} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+
+export default function ExtensionList({value, disabled, onChange, onFocus, onBlur}) {
+  const [options, setOptions] = useState(value.map(x => ({text: x, value: x})));
```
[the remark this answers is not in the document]
meluru: We need the local state to handle the additions to the options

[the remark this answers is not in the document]
ThiefMaster: I think you are right, the two other fields use it as well. I thought I did it without back then... Anyway, I think I'll have another look again tomorrow, because I'm curious *why* the state was necessary (should have added a comment when writing the other fields...)

indico/modules/events/editing/schemas.py
meluru: for me `e` as an entry, but I see no problem with using `ft`

indico/modules/events/editing/client/js/components/ExtensionList.jsx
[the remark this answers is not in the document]
ThiefMaster: I just tried this and it worked fone:

```diff
diff --git a/indico/modules/events/editing/client/js/components/ExtensionList.jsx b/indico/modules/events/editing/client/js/components/ExtensionList.jsx
index 1b7fa73034..7005997148 100644
--- a/indico/modules/events/editing/client/js/components/ExtensionList.jsx
+++ b/indico/modules/events/editing/client/js/components/ExtensionList.jsx
@@ -4,10 +4,7 @@ import {Dropdown} from 'semantic-ui-react';
 import {Translate} from 'indico/react/i18n';

 export default function ExtensionList({value, disabled, onChange, onFocus, onBlur}) {
-  const [options, setOptions] = useState(value.map(x => ({text: x, value: x})));
-
   const handleChange = (e, {value: newValue}) => {
-    setOptions(newValue.map(x => ({text: x, value: x})));
     onChange(newValue);
     onFocus();
     onBlur();
@@ -15,7 +12,7 @@ export default function ExtensionList({value, disabled, onChange, onFocus, onBlu

   return (
     <Dropdown
-      options={options}
+      options={value.map(x => ({text: x, value: x}))}
       value={value}
       disabled={disabled}
       search
```

panagiotappl:
This PR looks very good!

I just have one, not so crucial, suggestion. 

I found it a bit confusing when a delete icon is disabled the cursor still shows as a pointer. Made me think that the icon is clickable and just shows in a lighter red color (especially where no file type is deletable and you have no dark red delete icons showing up for comparison).

indico/web/client/js/react/forms/index.js
```diff
@@ -14,6 +14,7 @@ export {
   FinalRadio,
   FinalSubmitButton,
   FinalTextArea,
+  unsortedArraysEqual,
```
ThiefMaster: I think there's at least one place importing it directly from `indico/react/forms/fields`; maybe you could update that place to import it from here instead?

meluru:
@panagiotappl 
That's a good point, it shouldn't be a pointer in that case. I would also go one step further and add a tooltip saying 'File type has files attached to it, it's not possible to delete it' (or something along those lines). @ParthS007

indico/modules/events/editing/client/js/components/ExtensionList.jsx
```diff
@@ -0,0 +1,39 @@
+import _ from 'lodash';
```
mic4ael: missing LICENSE header

ThiefMaster: run `python bin/maintenance/update_header.py indico` should add it

ParthS007:
@meluru @panagiotappl I have worked on suggestions and made the channges. Please have a look and let me know if any other changes are required.

indico/modules/events/editing/client/js/components/ExtensionList.jsx
```diff
@@ -0,0 +1,39 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Dropdown} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+
+export default function ExtensionList({value, disabled, onChange, onFocus, onBlur}) {
+  const handleChange = (e, {value: newValue}) => {
+    onChange(_.uniq(newValue.map(x => x.replace(/\W/g, ''))));
```
ThiefMaster: Already mentioned offline, but for the same of completeness: This breaks e.g. `tar.gz` extensions; the only things to change here is stripping surrounding whitespace (if the dropdown component doesn't already do it) and leading dots.

indico/modules/events/editing/client/js/components/FileTypeManager.jsx
```diff
@@ -40,24 +102,106 @@ export default function FileTypeManager({eventId}) {
...
+              <Icon
+                color="blue"
+                name="pencil"
+                onClick={() => dispatch({type: 'EDIT_FILE_TYPE', fileType})}
+              />
+              <Popup
+                on="hover"
+                disabled={!fileType.isUsed}
+                position="right center"
+                content={Translate.string('This type has files attached')}
```
meluru: I would change it to something like: 'Not possible to delete: type has files attached'

ThiefMaster: not sure if that's needed - with a grayed-out delete button it's quite obvious. and it looks a bit ugly if it gets wrapped on a second line.

meluru: For us it is obvious, but I am always in for being more explicit, because some users get confused easily. Having said that, we can leave it up to @ParthS007 to decide

mic4ael:
There is a typo in the b058ede96da1895c72150d46088cc07e0b4570ed message

indico/modules/events/editing/client/js/components/ExtensionList.jsx
mic4ael: `update_header` just prints a warning that the header is missing (doesn't actually add it to the file). By the way, this comment has been resolved but no header is present in the file :P

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -48,6 +53,18 @@ def _process_DELETE(self):
...
 
 
+class RHCreateFileType(RHEditingManagementBase):
+    """Create a new file type."""
+
+    @use_rh_args(EditableFileTypeArgs)
+    def _process(self, data):
+        for extension in data.get('extensions', []):
+            if re.match(r'^\.', extension):
+                raise ExpectedError(_('Extensions cannot have leading dots'))
```
mic4ael: This should be probably replaced with:

```python
from webargs.flaskparser import abort
...
abort(422, messages={'extensions': [_('Extensions cannot have leading dots')]})
```

this way, users will see a nice error message attached to the field. At the moment nothing indicates what caused the error.

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -60,9 +77,14 @@ def _process_args(self):
 
     @use_rh_args(EditableFileTypeArgs, partial=True)
     def _process_PATCH(self, data):
+        for extension in data.get('extensions', []):
+            if re.match(r'^\.', extension):
+                raise ExpectedError(_('Extensions cannot have leading dots'))
```
mic4ael: The same as above

indico/modules/events/editing/controllers/backend/management.py
ThiefMaster: or better, add a validation method for the field in EditableFileTypeArgs

indico/modules/events/editing/client/js/components/ExtensionList.jsx
```diff
@@ -0,0 +1,39 @@
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Dropdown} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+
+export default function ExtensionList({value, disabled, onChange, onFocus, onBlur}) {
+  const handleChange = (e, {value: newValue}) => {
+    onChange(_.uniq(newValue.map(x => x.trim().replace(/^\./, ''))));
```
mic4ael: What if you have `....pdf` (which will break the backend validation)? Shouldn't it be `/^\.+/`?

meluru: That's a good point, I fixed that. I've actually added `/^\**\.+/` on front end in case someone would write `*.txt`

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -60,9 +74,14 @@ def _process_args(self):
 
     @use_rh_args(EditableFileTypeArgs, partial=True)
     def _process_PATCH(self, data):
+        for extension in data.get('extensions', []):
+            if re.match(r'^\.', extension):
```
mic4ael: Forgot to remove it?

indico/modules/events/editing/client/js/components/ExtensionList.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Dropdown} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+
+export default function ExtensionList({value, disabled, onChange, onFocus, onBlur}) {
+  const handleChange = (e, {value: newValue}) => {
+    onChange(_.uniq(newValue.map(x => x.trim().replace(/^\**\.+/, ''))));
```
mic4ael: Maybe just `/^[*.]+/` will be easier to read.

indico/modules/events/editing/schemas.py
```diff
@@ -183,3 +193,9 @@ def _check_for_unique_file_type_name(self, name):
             query = query.filter(EditingFileType.id != file_type.id)
         if query.has_rows():
             raise ValidationError(_('Name must be unique'))
+
+    @validates('extensions')
+    def _check_for_correct_extensions_format(self, extensions):
+        for extension in extensions:
```
ThiefMaster:
```python
if any(re.match(r'^[*.]+', ext) for ext in extensions):
    raise ValidationError(_('Extensions cannot have leading dots'))
```

BTW, if this check never happens unless someone messes with the client-side validation, we could skip i18n here.

meluru: Does it hurt to have it? I would leave it as it is

ThiefMaster: not sure if it was about i18n or replacing the loop with any() or both, but like i said, just a suggestion ;)

PR #4242, opened 2020-01-15
Don't attach tickets to emails for invalid registrations

mic4ael:


indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -204,9 +204,8 @@ def _send_emails(self, form):
             template = get_template_module('events/registration/emails/custom_email.html',
                                            email_subject=email_subject, email_body=email_body)
             bcc = [session.user.email] if form.copy_for_sender.data else []
-            attachments = (get_ticket_attachments(registration)
-                           if 'attach_ticket' in form and form.attach_ticket.data
-                           else None)
+            attach_ticket = not registration.is_ticket_blocked and 'attach_ticket' in form and form.attach_ticket.data
```
ThiefMaster: I'd move `not registration.is_ticket_blocked` to the end since it's likely more expensive than just accessing the form data.

indico/modules/events/registration/templates/management/email.html
```diff
@@ -1,4 +1,15 @@
 {% from 'forms/_form.html' import form_header, form_rows, form_footer  %}
+{% from 'message_box.html' import message_box %}
+
+{% if registrations_without_ticket_count > 0 %}
```
ThiefMaster: I'd also add some JS to only show this when the user selects to attach tickets.

ThiefMaster: no need for the `> 0`

indico/modules/events/registration/templates/management/email.html
```diff
@@ -8,3 +19,19 @@
            data-href="{{ url_for('.email_registrants_preview', regform) }}">
     <a class="i-button big" data-button-back>{% trans %}Cancel{% endtrans %}</a>
 {% endcall %}
+
+<script>
+    (function() {
+        'use strict';
+
+        var attachTicketField = $('#{{ form.attach_ticket.id }}');
```
ThiefMaster: let's use ES6 here :)

indico/modules/events/registration/templates/management/email.html
mic4ael: I know I just prefer to be explicit

indico/modules/events/registration/templates/management/email.html
```diff
@@ -1,4 +1,17 @@
 {% from 'forms/_form.html' import form_header, form_rows, form_footer  %}
+{% from 'message_box.html' import message_box %}
+
+{% if registrations_without_ticket_count %}
+    <div class="attach-ticket-warning" style="display: none;">
```
ThiefMaster: I think we need to move this farther down - unless you are on at least a 1440p screen, you'll most likely have to scroll to see the switch, but then the warning will not be visible since it'll be in the area from which you scrolled away...

CHANGES.rst
```diff
@@ -14,6 +14,8 @@ Bugfixes
   required even though they should be optional
 - Avoid browsers prefilling stored passwords in togglable password fields
   such as the event access key
+- Make sure that tickets are not attached to emails sent to registrants for whom
+  tickets are blocked
```
ThiefMaster: the reference to the PR is missing here

PR #4243, opened 2020-01-15
Add map_url for geographic coordinates in room details

### LINKED ISSUE #4146: Room booking: add link to map in room parameters
When looking at room details, you can see its GPS coordinates (if defined) but it is not a clickable link redirecting to a map service, like it is in an Indico event. It makes difficult to locate a room which is not linked to an event... Would it be possible to add it?

ParthS007:
closes #4146

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,7 @@ Improvements
   refreshing identity information
 - Show more verbose error when email validation fails during event
   registration (:issue:`4177`)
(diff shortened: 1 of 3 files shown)


ThiefMaster:
please add this to the changelog (for 2.3, in "improvements"):

```rst
- Add link to external map in room details view (:issue:`4146`)
```

ThiefMaster:
the commit message is a bit long (>50 chars), but could be easily shortened to "Add map link in room details view"

PR #4245, opened 2020-01-15
Show allowed file extensions in FileManager

### LINKED ISSUE #4218: File Manager: Display file extensions
Displaying allowed file extensions in FileManager is currently missing:

image

panagiotappl:
closes #4218 

image

### DIFF
--- a/indico/modules/events/editing/client/js/components/FileManager/FileManager.module.scss
+++ b/indico/modules/events/editing/client/js/components/FileManager/FileManager.module.scss
@@ -35,16 +35,38 @@
 
     .file-type {
         overflow: hidden;
-
-        &:not(:nth-child(3n)) {
-            margin-right: 0.9em;
-        }
-
         margin-bottom: 1em;
         border: 1px solid $gray;
         border-radius: 3px;
         padding: 1em;
         flex: 0 0 32%;
+
+        &:not(:nth-child(3n)) {
+            margin-right: 0.9em;
+        }
+
+        .file-extensions {
+            font-style: italic;
+            color: $gray;
+            list-style-type: none;
+            margin: 0.2em 0;
+            padding: 0;
+            white-space: nowrap;
+            overflow: hidden;
+            text-overflow: ellipsis;
+
+            li {
+                display: inline;
+
+                &:not(:last-child)::after {
+                    content: ', ';
+                }
+            }
+        }
+    }
+
+    .file-type h3 {
+        margin-bottom: 0;
     }
 }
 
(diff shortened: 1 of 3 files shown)


indico/modules/events/editing/client/js/components/FileManager/index.jsx
```diff
@@ -91,6 +93,13 @@ function FileType({uploadURL, fileType, uploads}) {
   return (
     <div styleName="file-type">
       <h3>{fileType.name}</h3>
+      <TooltipIfTruncated customText={fileType.extensions.map(ext => ext).join(', ')}>
```
ThiefMaster: `.map(ext => ext)` doesn't do anything

indico/web/client/js/react/components/TooltipIfTruncated.jsx
```diff
@@ -11,15 +11,21 @@ import PropTypes from 'prop-types';
 export default class TooltipIfTruncated extends React.Component {
   static propTypes = {
     children: PropTypes.any.isRequired,
+    customText: PropTypes.string,
```
ThiefMaster: `customText` vs `text` vs `tooltip`.. any opinions @indico/development-team

indico/modules/events/editing/client/js/components/FileManager/index.jsx
panagiotappl: Oops, it was leftover from previous implementations. Thanks for pointing out

indico/modules/events/editing/client/js/components/FileManager/FileManager.module.scss
```diff
@@ -35,16 +35,38 @@
...
+            padding: 0;
+            white-space: nowrap;
+            overflow: hidden;
+            text-overflow: ellipsis;
+
+            li {
+                display: inline;
+            }
+
+            li:not(:last-child)::after {
```
mic4ael: Maybe move it inside the `li` selector?

```scss

li {
    &:not(:last-child)::after {
```

indico/web/client/js/react/components/TooltipIfTruncated.jsx
mic4ael: Hm.. `tooltip` seems like a good name for that (I would be more in favor of `customText` if `element.innerText` had precedence over the prop value, but apparently it is not the case)

panagiotappl: I think I'll just go with `tooltip`. What do you think?

PR #4248, opened 2020-01-17
Editing file type filename template

### LINKED ISSUE #4222: File Type Manager: add a field to define filename template
Add a possibility to define allowed filenames for file types:
e.g. pdf file type should be named '*_paper'

mic4ael:
closes #4222


indico/modules/events/editing/operations.py
```diff
@@ -241,7 +241,8 @@ def create_new_file_type(event, **data):
 
 
 def update_file_type(file_type, **data):
-    file_type.populate_from_dict(data, keys=('name', 'extensions', 'allow_multiple_files', 'required', 'publishable'))
+    attrs_to_update = ('name', 'extensions', 'allow_multiple_files', 'required', 'publishable', 'filename_template')
```
ThiefMaster: shouldn't this be updated in `create_new_file_type` as well? (maybe move the list to the top level and use it in both places)

mic4ael: Indeed, I only tested update process :P

indico/migrations/versions/20200113_1340_6a185fdcd4ee_add_filename_template_column.py
```diff
@@ -0,0 +1,24 @@
+"""Add filename_template column
```
ThiefMaster: Before we merge, I would move this into the existing alembic revision and just paste some SQL in the PR's description for devs to update their DB.

OTOH, we'll probably reset the alembic steps with 3.0 anyway, so it doesn't matter that much.

indico/modules/events/editing/fields.py
```diff
@@ -49,6 +50,13 @@ def validate_files(self, value):
                     raise ValidationError('File type "{}" does not allow these extensions: {}'
                                           .format(file_type.name, ', '.join(invalid_extensions)))
 
+            # ensure all filenames conform to the template
+            if file_type.filename_template:
+                filenames = {os.path.splitext(f.filename)[0] for f in files}
+                if not all(fnmatch.fnmatch(filename, file_type.filename_template) for filename in filenames):
+                    raise ValidationError("Some files don't conform to the filename template '{}'"
```
ThiefMaster: unless we check the template locally (which I think would be nice), this needs to be translatable

indico/modules/events/editing/models/file_types.py
```diff
@@ -55,6 +55,11 @@ def __table_args__(cls):
         nullable=False,
         default=False
     )
+    filename_template = db.Column(
+        db.String,
+        nullable=True,
+        default=''
```
ThiefMaster: That's contradicting with `nullable=True`: We should decide whether we want to use an empty string or None for "no template". If we go for None, there shouldn't be a default (implicitly defaults to None/NULL); otherwise it should be `nullable=False`

indico/modules/events/editing/client/js/components/FileTypeModal.jsx
```diff
@@ -56,6 +56,13 @@ export default function FileTypeModal({onClose, onSubmit, initialValues, header}
             )}
             <Form id="file-type-form" onSubmit={fprops.handleSubmit}>
               <FinalInput name="name" label={Translate.string('Name')} required />
+              <FinalInput
```
ThiefMaster: if we go for `None` instead of `''`, we need `nullIfEmpty` here to convert empty strings to null

indico/modules/events/editing/schemas.py
```diff
@@ -179,6 +180,7 @@ class Meta:
         rh_context = ('event', 'file_type')
 
     name = fields.String(required=True, validate=not_empty)
+    filename_template = fields.String()
```
ThiefMaster: If we go for `None`, this should have `validate=not_empty, allow_none=True`

indico/modules/events/editing/models/file_types.py
ThiefMaster: In the past (WTForms) we usually used empty strings with `nullable=False` because this was more straightforward - zero extra steps necessary, and no risk of getting a mix of the two values.

With final-form and webargs/marshmallow the overhead for empty-to-null is minimal, but of course there's still the risk of getting empty strings in the DB if we don't validate properly (unless we add a CHECK, and personally I'd prefer to avoid creating CHECKs for every nullable string field).

cc @indico/development-team

mic4ael: Yeah, you are right. This shouldn't have any default value.

> With final-form and webargs/marshmallow the overhead for empty-to-null is minimal

Yep, that sounds like a good reason to stop providing defaults on the SQLAlchemy level.

indico/modules/events/editing/fields.py
mic4ael: What do you mean by "locally" in this context?

ThiefMaster: client-side (no need to waste people's time translating errors they never see unless they tinker with the data sent to the API)

mic4ael: For now it is only backend validation (I will probably look into client-side validation once you are done with your PR)

indico/modules/events/editing/models/file_types.py
```diff
@@ -72,4 +76,4 @@ def __table_args__(cls):
     @return_ascii
     def __repr__(self):
         return format_repr(self, 'id', 'event_id', 'extensions', allow_multiple_files=False, required=False,
-                           publishable=False, _text=self.name)
+                           publishable=False, filename_template='', _text=self.name)
```
ThiefMaster: `filename_template=None`

webpack.config.babel.js
```diff
@@ -91,6 +91,9 @@ export default env => {
     // in base.js are more general
     'resolve.alias': 'prepend',
   })(webpackDefaults(env, config), {
+    node: {
+      fs: 'empty',
```
ThiefMaster: what does this do (and why is it needed)? might want to leave a commend here since it's not straightforward

pferreir: Sounds like some naughty lib tries to use node's `fs` module.

pferreir: (BTW, this just tells webpack to mock it, IIRC)

mic4ael: So, the problem is with the addition of `micromatch`. I think some of its dependencies uses `fs` and this says to mock the module entirely (as @pferreir pointed out correctly)

PR #4251, opened 2020-01-20
Fix the copy link icon

ParthS007:


indico/web/client/js/react/components/ClipboardButton.jsx
```diff
@@ -29,11 +29,14 @@ export default function ClipboardButton({text, successText}) {
       on="click"
       position="bottom center"
       trigger={
-        <Icon
+        <Button
+          basic
+          type="button"
+          icon="linkify"
+          onClick={copied ? null : handleOpen}
```
ThiefMaster: `copied ? null : ` might not be necessary; I think disabled buttons don't call onclick (but try it)

ParthS007: I have checked it, you are correct. :+1:

PR #4253, opened 2020-01-22
Approval not required for same slots of a day

### LINKED ISSUE #4214: Room Booking: don't reset confirmation state if booking is just shorter
If I book a room from `14:00 - 15:00` and then change the booking to `14:00 - 14:30`, I shouldn't need further confirmation from the room manager as I am staying within the same time slot.

ParthS007:
closes #4214

### DIFF
--- a/indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
+++ b/indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
@@ -47,6 +47,18 @@ class BookingEditForm extends React.Component {
     hideOptions: {},
   };
 
+  showApprovalWarning = (dirtyFields, initialValues, values) => {
+    const initialStartTime = moment(initialValues.timeSlot.startTime, 'HH:mm');
+    const initialEndTime = moment(initialValues.timeSlot.endTime, 'HH:mm');
+    const newStartTime = moment(values.timeSlot.startTime, 'HH:mm');
+    const newEndTime = moment(values.timeSlot.endTime, 'HH:mm');
+    return (
+      Object.keys(dirtyFields).some(key => key === 'dates' || key.startsWith('recurrence.')) ||
+      newStartTime.isBefore(initialStartTime) ||
+      newEndTime.isAfter(initialEndTime)
+    );
+  };
+
   recurrenceTypeChanged = newType => {
     const {
       booking: {startDt},
@@ -226,11 +238,9 @@ class BookingEditForm extends React.Component {
             />
           )}
           {!room.canUserBook && room.canUserPrebook && isAccepted && (
-            <FormSpy subscription={{dirtyFields: true}}>
-              {({dirtyFields}) =>
-                Object.keys(dirtyFields).some(
-                  key => key === 'dates' || key === 'timeSlot' || key.startsWith('recurrence.')
-                ) && (
+            <FormSpy subscription={{dirtyFields: true, initialValues: true, values: true}}>
+              {({dirtyFields, initialValues, values}) =>
+                this.showApprovalWarning(dirtyFields, initialValues, values) && (
                   <Message warning visible>
                     <Message.Header>
                       <Translate>This booking will require approval!</Translate>
--- a/indico/modules/rb/operations/bookings.py
+++ b/indico/modules/rb/operations/bookings.py
@@ -363,6 +363,15 @@ def has_same_dates(old_booking, new_booking):
             old_booking.repeat_frequency == new_booking['repeat_frequency'])
 
 
+def has_same_slots(old_booking, new_booking):
+    if (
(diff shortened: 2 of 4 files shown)


indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -226,10 +237,13 @@ class BookingEditForm extends React.Component {
...
-            <FormSpy subscription={{dirtyFields: true}}>
-              {({dirtyFields}) =>
+            <FormSpy subscription={{dirtyFields: true, initialValues: true, values: true}}>
+              {({dirtyFields, initialValues, values}) =>
                 Object.keys(dirtyFields).some(
-                  key => key === 'dates' || key === 'timeSlot' || key.startsWith('recurrence.')
+                  key =>
+                    key === 'dates' ||
+                    key.startsWith('recurrence.') ||
+                    this.showWarning(initialValues, values)
```
ThiefMaster: This doesn't really belong here; you are inside the `some()` call, but the check is not specific to any of the dirtyFields key. Move it outside the `some()` call instead.

FWIW, you could change it to `showWarning(dirtyFields, initialValues, values)` and then move the `Object.keys(dirtyFields).some()` check into the function as well.

indico/modules/rb/operations/bookings.py
```diff
@@ -363,6 +363,16 @@ def has_same_dates(old_booking, new_booking):
...
 
 
+def has_same_slots(old_booking, new_booking):
+    if not (old_booking.repeat_interval == new_booking['repeat_interval'] and
+            old_booking.repeat_frequency == new_booking['repeat_frequency']):
+        return False
+    if (old_booking.start_dt <= new_booking['start_dt']) and (old_booking.end_dt >= new_booking['end_dt']):
+        return True
+    else:
+        return False
```
ThiefMaster:
```python
return old_booking.start_dt <= new_booking['start_dt'] and old_booking.end_dt >= new_booking['end_dt']
```

indico/modules/rb/operations/bookings.py
```diff
@@ -363,6 +363,16 @@ def has_same_dates(old_booking, new_booking):
             old_booking.repeat_frequency == new_booking['repeat_frequency'])
 
 
+def has_same_slots(old_booking, new_booking):
+    if not (old_booking.repeat_interval == new_booking['repeat_interval'] and
+            old_booking.repeat_frequency == new_booking['repeat_frequency']):
+        return False
```
ThiefMaster: multi-line if conditions are usually wrapped like this in Python. Also, using or instead of he negation aroudn everything is more readable:

```python
    if (old_booking.repeat_interval != new_booking['repeat_interval'] or
            old_booking.repeat_frequency != new_booking['repeat_frequency']):
        return False
```

or in black-style:

```python
    if (
        old_booking.repeat_interval != new_booking["repeat_interval"]
        or old_booking.repeat_frequency != new_booking["repeat_frequency"]
    ):
        return False
```

indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -47,6 +47,17 @@ class BookingEditForm extends React.Component {
     hideOptions: {},
   };
 
+  showWarning = (initialValues, values) => {
+    return !(
+      moment(values['timeSlot']['startTime'], 'HH:mm').isSameOrAfter(
```
ThiefMaster: `values.timeSlot.startTime`

ThiefMaster: assign the moment() values; that'll make the final condition more readable

indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -47,6 +47,17 @@ class BookingEditForm extends React.Component {
...
+        moment(initialValues['timeSlot']['startTime'], 'HH:mm')
+      ) &&
```
ThiefMaster: `||`, remove the negation and use the negated versions of isSameOrAFter / isSameOrBefore

indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -47,6 +47,17 @@ class BookingEditForm extends React.Component {
     hideOptions: {},
   };
 
+  showWarning = (initialValues, values) => {
```
ThiefMaster: Let's find a better name for this. `shouldShowApprovalWarning` or something like that

indico/modules/rb/operations/bookings.py
ParthS007: Going with black style

PR #4257, opened 2020-01-24
Add Popup for showing filename template info

### LINKED ISSUE #4254: Display Filename Template in File Manager
Display the `FileType`'s `filename_template` in the `FileManager` (so that users know that the filenames of uploaded files need to conform to some specific pattern)

ParthS007:
closes #4254 

image

### DIFF
--- a/indico/modules/events/editing/client/js/components/FileManager/index.jsx
+++ b/indico/modules/events/editing/client/js/components/FileManager/index.jsx
@@ -114,7 +114,21 @@ function FileType({uploadURL, fileType, uploads}) {
   const ref = useRef(null);
   return (
     <div styleName="file-type">
-      <h3>{fileType.name}</h3>
+      <h3>
+        {fileType.name}
+        {fileType.filenameTemplate !== null && (
+          <Popup
+            position="bottom center"
+            content={
+              <Translate>
+                Filenames must have the format{' '}
+                <Param name="template" value={fileType.filenameTemplate} wrapper={<code />} />
+              </Translate>
+            }
+            trigger={<Icon corner="top right" name="info" />}
+          />
+        )}
+      </h3>
       <TooltipIfTruncated tooltip={fileType.extensions.join(', ')}>
         <ul styleName="file-extensions">
           {fileType.extensions.length !== 0


indico/modules/events/editing/client/js/components/FileManager/index.jsx
```diff
@@ -114,7 +114,20 @@ function FileType({uploadURL, fileType, uploads}) {
...
-      <h3>{fileType.name}</h3>
+      <h3>
+        {fileType.name}
+        <Popup
+          position="bottom center"
+          content={
+            fileType.filenameTemplate !== null ? (
+              <Translate>Filenames must have the format: {fileType.filenameTemplate}</Translate>
+            ) : (
+              Translate.string('No file name restrictions')
```
ThiefMaster: I'd simply not render the popup if there are no restrictions

indico/modules/events/editing/client/js/components/FileManager/index.jsx
```diff
@@ -114,7 +114,20 @@ function FileType({uploadURL, fileType, uploads}) {
...
   return (
     <div styleName="file-type">
-      <h3>{fileType.name}</h3>
+      <h3>
+        {fileType.name}
+        <Popup
+          position="bottom center"
+          content={
+            fileType.filenameTemplate !== null ? (
+              <Translate>Filenames must have the format: {fileType.filenameTemplate}</Translate>
```
ThiefMaster: You cannot interpolate values like this in translated strings; if you search for `<Param` in the codebase you'll find examples how to use parameters there. You can also use this to wrap the template in `<code>` (using the `wrapper` prop of `Param`)

indico/modules/events/editing/client/js/components/FileManager/index.jsx
```diff
@@ -114,7 +114,20 @@ function FileType({uploadURL, fileType, uploads}) {
...
+            ) : (
+              Translate.string('No file name restrictions')
+            )
+          }
+          trigger={<Icon corner="top right" name="info" style={{cursor: 'pointer'}} />}
```
ThiefMaster: I don't think this cursor is a good idea here; it's not something clickable!

PR #4259, opened 2020-01-27
Fix Layout Issue in Filetype Manager

### LINKED ISSUE #4258: Fix layout issues in the `FileTypeManager`
Screenshot from 2020-01-24 12-50-13

ParthS007:
closes #4258 

**Updated Screenshot** 

Screenshot from 2020-01-28 15-31-51

### DIFF
--- a/indico/modules/events/editing/client/js/components/FileTypeManager.jsx
+++ b/indico/modules/events/editing/client/js/components/FileTypeManager.jsx
@@ -11,7 +11,8 @@ import editFileTypeURL from 'indico-url:event_editing.api_edit_file_type';
 
 import React, {useReducer} from 'react';
 import PropTypes from 'prop-types';
-import {Button, Icon, Loader, Message, Segment, Popup} from 'semantic-ui-react';
+import {Button, Icon, Loader, Message, Segment, Popup, Label} from 'semantic-ui-react';
+import {TooltipIfTruncated} from 'indico/react/components';
 import {Param, Translate} from 'indico/react/i18n';
 import {getChangedValues, handleSubmitError} from 'indico/react/forms';
 import {useIndicoAxios} from 'indico/react/hooks';
@@ -40,6 +41,18 @@ function fileTypesReducer(state, action) {
       return state;
   }
 }
+
+function StatusIcon({active, icon, text}) {
+  return (
+    <Icon size="small" color={active ? 'blue' : 'grey'} name={icon} title={active ? text : ''} />
+  );
+}
+
+StatusIcon.propTypes = {
+  active: PropTypes.bool.isRequired,
+  icon: PropTypes.string.isRequired,
+  text: PropTypes.string.isRequired,
+};
 
 export default function FileTypeManager({eventId}) {
   const [state, dispatch] = useReducer(fileTypesReducer, initialState);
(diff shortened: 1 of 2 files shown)


indico/modules/events/editing/client/js/components/FileTypeManager.jsx
```diff
@@ -91,18 +92,20 @@ export default function FileTypeManager({eventId}) {
 
   const {operation, fileType: currentFileType} = state;
   return (
-    <div styleName="file-types-container">
+    <Segment className="ui basic" styleName="file-types-container">
```
mic4ael: `<Segment basic...`

mic4ael:
I still don't think it is the best solution for the problem. This is what I managed to accomplish with only `TooltipIfTruncated` and a couple of CSS changes

image

indico/modules/events/editing/client/js/components/FileTypeManager.jsx
```diff
@@ -41,6 +42,18 @@ function fileTypesReducer(state, action) {
   }
 }
 
+function statusIcon(fileParam, iconName, text) {
```
ThiefMaster: It's a component so it should be named `StatusIcon`, and used as `<StatusIcon ... />` below instead of `{statusIcon(...)}`

indico/modules/events/editing/client/js/components/FileTypeManager.jsx
```diff
@@ -41,6 +42,18 @@ function fileTypesReducer(state, action) {
...
+  return (
+    <Icon
+      styleName={fileParam ? '' : 'disabled'}
+      size="small"
+      color={fileParam ? 'blue' : 'grey'}
+      name={iconName}
+      title={fileParam ? Translate.string(text) : ''}
```
ThiefMaster: Don't pass dynamic values to `Translate.string()`; this doesn't work since there's no way those values can be extracted for translation. Pass `Translate.string('...')` when using the StatusIcon component instead.

indico/modules/events/editing/client/js/components/FileTypeManager.jsx
```diff
@@ -41,6 +42,17 @@ function fileTypesReducer(state, action) {
   }
 }
 
+function StatusIcon({fileParam, iconName, text}) {
```
ThiefMaster: `active` would be a clearer name instead of `fileParam`

indico/modules/events/editing/client/js/components/FileTypeManager.jsx
```diff
@@ -208,3 +212,14 @@ export default function FileTypeManager({eventId}) {
 FileTypeManager.propTypes = {
   eventId: PropTypes.number.isRequired,
 };
+
+StatusIcon.propTypes = {
```
ThiefMaster: move this up to the component

indico/modules/events/editing/client/js/components/FileTypeManager.jsx
```diff
@@ -208,3 +212,14 @@ export default function FileTypeManager({eventId}) {
...
+  fileParam: PropTypes.bool.isRequired,
+  iconName: PropTypes.string,
+  text: PropTypes.string,
+};
+
+StatusIcon.defaultProps = {
```
ThiefMaster: you can make all props required and get rid of this

indico/modules/events/editing/client/js/components/FileTypeManager.jsx
```diff
@@ -99,65 +111,57 @@ export default function FileTypeManager({eventId}) {
...
+            }
+          />
+          <Icon
+            style={{cursor: 'pointer'}}
+            color="blue"
+            name="pencil"
+            corner="top right"
+            onClick={() => dispatch({type: 'EDIT_FILE_TYPE', fileType})}
+          />
+          <div styleName="filetype-head">
```
ThiefMaster: `filetype-header`

indico/modules/events/editing/client/js/components/FileTypeManager.jsx
```diff
@@ -41,6 +42,17 @@ function fileTypesReducer(state, action) {
   }
 }
 
+function StatusIcon({fileParam, iconName, text}) {
+  return (
+    <Icon
+      size="small"
+      color={fileParam ? 'blue' : 'grey'}
+      name={iconName}
+      title={fileParam ? text : ''}
```
mic4ael: You can just use `text` (defaultProp for `text` is null which is valid here)

indico/modules/events/editing/client/js/components/FileTypeManager.jsx
mic4ael: `icon` instead of `iconName`

indico/modules/events/editing/client/js/components/FileTypeManager.jsx
ThiefMaster: There shouldn't be a default (as the prop is always specified); and `title={false}` would be very odd (and likely a proptypes violation)

mic4ael: I meant that `title` has default `null` which works fine in this case

mic4ael: ahh, I see now.. nvm my comment

PR #4262, opened 2020-01-28
Don't delete or unset the last publishable type

### LINKED ISSUE #4205: File Type Manager: handle publishable types
There should be a minimum of a publishable type. This means that we shouldn't allow the last publishable type to be deleted or unset.

There should be some way to (un-)set a publishable type. Something similar to a checkbox for each list entry could be an option. Otherwise, a field with a list of publishable types could also work (as long as it cannot be set to empty).

panagiotappl:
closes #4205

### DIFF
--- a/indico/modules/events/editing/controllers/backend/management.py
+++ b/indico/modules/events/editing/controllers/backend/management.py
@@ -78,5 +78,12 @@ class RHEditFileType(RHEditingManagementBase):
     def _process_DELETE(self):
         if EditingRevisionFile.query.with_parent(self.file_type).has_rows():
             raise ExpectedError(_('Cannot delete file type which already has files'))
+        if self.file_type.publishable:
+            is_last = not (EditingFileType.query
+                           .with_parent(self.event)
+                           .filter(EditingFileType.publishable, EditingFileType.id != self.file_type.id)
+                           .has_rows())
+            if is_last:
+                raise ExpectedError(_('Cannot delete the only publishable file type'))
         delete_file_type(self.file_type)
         return '', 204
--- a/indico/modules/events/editing/schemas.py
+++ b/indico/modules/events/editing/schemas.py
@@ -206,3 +206,15 @@ class EditableFileTypeArgs(mm.Schema):
         for extension in extensions:
             if re.match(r'^[*.]+', extension):
                 raise ValidationError(_('Extensions cannot have leading dots'))
+
+    @validates('publishable')
+    def _check_if_can_unset_or_delete(self, publishable):
+        event = self.context['event']
+        file_type = self.context['file_type']
+        if file_type and not publishable:
+            is_last = not (EditingFileType.query
+                           .with_parent(event)
+                           .filter(EditingFileType.publishable, EditingFileType.id != file_type.id)
+                           .has_rows())
+            if is_last:
+                raise ValidationError(_('Cannot unset the only publishable file type'))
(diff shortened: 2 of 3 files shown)


indico/modules/events/editing/operations.py
```diff
@@ -252,3 +253,12 @@ def update_file_type(file_type, **data):
 def delete_file_type(file_type):
     logger.info('File type %r deleted by %r', file_type, session.user)
     db.session.delete(file_type)
+
+
+def check_if_can_unset_or_delete(event, file_type):
+    types = (EditingFileType.query
+                            .with_parent(event)
+                            .filter(EditingFileType.publishable, EditingFileType.id != file_type.id)
+                            .all())
```
ThiefMaster: You can use `.has_rows()` if you only care whether there are rows or not

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -72,11 +73,15 @@ def _process_args(self):
 
     @use_rh_args(EditableFileTypeArgs, partial=True)
     def _process_PATCH(self, data):
+        if 'publishable' in data and not data['publishable']:
+            check_if_can_unset_or_delete(self.event, self.file_type)
```
ThiefMaster: Can you try whether you can put this check in the schema instead of having it here?

indico/modules/events/editing/schemas.py
```diff
@@ -206,3 +206,15 @@ def _check_for_correct_extensions_format(self, extensions):
...
+
+    @validates('publishable')
+    def _check_if_can_unset_or_delete(self, publishable):
+        event = self.context['event']
+        file_type = self.context['file_type']
+        if not publishable:
+            types = (EditingFileType.query
+                                    .with_parent(event)
+                                    .filter(EditingFileType.publishable, EditingFileType.id != file_type.id)
+                                    .all())
```
ThiefMaster: You can use `.has_rows()` here as well ;)

indico/modules/events/editing/controllers/backend/management.py
panagiotappl: I put it on the schema as discussed in MM but that doesn't cover the case of deletion, right? Should I leave the function in `operations.py` and call it both in the schema and `management.py`?

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -78,5 +78,12 @@ def _process_PATCH(self, data):
     def _process_DELETE(self):
         if EditingRevisionFile.query.with_parent(self.file_type).has_rows():
             raise ExpectedError(_('Cannot delete file type which already has files'))
+        if self.file_type.publishable:
+            types = (EditingFileType.query
```
ThiefMaster: I'd name this `is_last` instead of `types`, since the latter implies more a list of types instead of a bool

indico/modules/events/editing/schemas.py
```diff
@@ -206,3 +206,15 @@ def _check_for_correct_extensions_format(self, extensions):
         for extension in extensions:
             if re.match(r'^[*.]+', extension):
                 raise ValidationError(_('Extensions cannot have leading dots'))
+
+    @validates('publishable')
+    def _check_if_can_unset_or_delete(self, publishable):
+        event = self.context['event']
+        file_type = self.context['file_type']
+        if file_type and not publishable:
+            types = (EditingFileType.query
```
ThiefMaster: same thing here

PR #4265, opened 2020-01-30
Add new  LOCAL_GROUP setting that can be used to fully disable local groups

### LINKED ISSUE #4260: Add LOCAL_GROUPS setting to disable local groups everywhere
This is a prerequisite for #3040 (see the comments there).

- [x] Add a new `LOCAL_GROUPS` (default: `True`) setting for indico.conf
- [x] Add it to the docs for the indico.conf settings
- [x] If local groups are disabled with this setting, then:
  - [x] Hide local group features in the admin area
  - [x] Do not return local groups when searching for groups
  - [x] Always return false when checking membership against local group (in case local groups were created and added to ACLs before turning off the setting)
  - [x] Make sure saving such an ACL fails gracefully. We do something similar for remote (multipass) groups which do not exist anymore. Keep in mind that we have different fields for ACLs; the React one used e.g. in the Room Booking settings, and the classic one used e.g. when editing the protection of an event. I think  `principal_from_fossil` and `principal_from_identifier` are the places to look at for this.

ParthS007:
closes #4260

### DIFF
--- a/docs/source/config/settings.rst
+++ b/docs/source/config/settings.rst
@@ -34,6 +34,16 @@ Authentication
 
     Default: ``True``
 
+.. data:: LOCAL_GROUPS
+
+    This setting controls whether local Indico groups are available.
+    If no centralized authentication infrastructure that supports groups
+    (e.g. LDAP) is used, local groups are the only way to define groups in 
+    Indico, but if you do have central groups it may be useful to disable 
+    local ones to have all groups in one central place.
+
+    Default: ``True``
+
 .. data:: LOCAL_REGISTRATION
 
     This setting controls whether people accessing Indico can create a
--- a/indico/modules/groups/controllers.py
+++ b/indico/modules/groups/controllers.py
@@ -12,9 +12,10 @@ from operator import attrgetter
 from flask import flash, jsonify, redirect, request
 from sqlalchemy.orm import joinedload
 from webargs import fields
-from werkzeug.exceptions import NotFound
+from werkzeug.exceptions import Forbidden, NotFound
 
 from indico.core.auth import multipass
+from indico.core.config import config
 from indico.core.db import db
 from indico.modules.admin import RHAdminBase
 from indico.modules.groups import GroupProxy
@@ -39,20 +40,27 @@ class RHGroups(RHAdminBase):
         groups = [g.proxy for g in query]
         providers = [p for p in multipass.identity_providers.itervalues() if p.supports_groups]
         form = SearchForm(obj=FormDefaults(exact=True))
-        if not providers:
+        groups_enabled = True
+        if not providers and not config.LOCAL_GROUPS:
+            groups_enabled = False
+        elif not providers:
             del form.provider
         else:
-            form.provider.choices = ([('', _('All')), ('indico', _('Local Groups'))] +
-                                     [(p.name, p.title) for p in sorted(providers, key=attrgetter('title'))])
+            choices = [('', _('All'))]
+            if config.LOCAL_GROUPS:
+                choices.append(('indico', _('Local Groups')))
+            choices += [(p.name, p.title) for p in sorted(providers, key=attrgetter('title'))]
+            form.provider.choices = choices
         search_results = None
-        if form.validate_on_submit():
+        if groups_enabled and form.validate_on_submit():
             search_providers = None if not providers or not form.provider.data else {form.provider.data}
             search_results = GroupProxy.search(form.name.data, exact=form.exact.data, providers=search_providers)
             search_results.sort(key=attrgetter('provider', 'name'))
         provider_titles = {p.name: p.title for p in multipass.identity_providers.itervalues()}
         provider_titles[None] = _('Local')
         return WPGroupsAdmin.render_template('groups.html', groups=groups, providers=providers, form=form,
-                                             search_results=search_results, provider_titles=provider_titles)
+                                             search_results=search_results, provider_titles=provider_titles,
+                                             groups_enabled=groups_enabled)
 
 
 class RHGroupBase(RHAdminBase):
@@ -63,6 +71,8 @@ class RHGroupBase(RHAdminBase):
             group = None
         if group is None or group.group is None:
             raise NotFound
+        if not config.LOCAL_GROUPS and group.is_local:
+            raise Forbidden('Local groups are disabled.')
         self.group = group
 
 
@@ -88,6 +98,8 @@ class RHGroupEdit(RHAdminBase):
     """Admin group modification/creation"""
 
     def _process_args(self):
+        if not config.LOCAL_GROUPS:
+            raise Forbidden('Local groups are disabled.')
         if 'group_id' in request.view_args:
             self.new_group = False
             self.group = LocalGroup.get(request.view_args['group_id'])
@@ -115,6 +127,8 @@ class RHGroupEdit(RHAdminBase):
 
 class RHLocalGroupBase(RHAdminBase):
     def _process_args(self):
+        if not config.LOCAL_GROUPS:
+            raise Forbidden('Local groups are disabled.')
         self.group = LocalGroup.get(request.view_args['group_id'])
         if self.group is None:
             raise NotFound
--- a/indico/modules/groups/core.py
+++ b/indico/modules/groups/core.py
@@ -13,6 +13,7 @@ from flask_multipass import MultipassException
 from werkzeug.utils import cached_property
 
 from indico.core.auth import multipass
+from indico.core.config import config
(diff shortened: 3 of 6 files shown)


docs/source/config/settings.rst
```diff
@@ -34,6 +34,12 @@ Authentication
 
     Default: ``True``
 
+.. data:: LOCAL_GROUPS
+
+    This setting controls whether local Indico groups are available.
```
ThiefMaster: I'd add this below your text:

```
If no centralized authentication infrastructure that supports groups (e.g. LDAP)
is used, local groups are the only way to define groups in Indico, but if you do
have central groups it may be useful to disable local ones to have all groups
in one central place.
```

indico/modules/groups/controllers.py
```diff
@@ -31,7 +32,15 @@
 from indico.web.rh import RHProtected
 
 
-class RHGroups(RHAdminBase):
+class RHLocalGroupConfig:
```
ThiefMaster: All classes in Python 2 should be new-style, i.e. inherit from `object` (or another base class that already inherits from `object`).

Since this particular one is a mixin, it should be named e.g. `LocalGroupConfigMixin` - but I don't think a mixin is the best choice here, since you can just have this class (renamed to `RHLocalGroupsBase`) inherit from `RHAdminBase`, add the check here, and then inherit all the other classes that currently use `RHAdminBase` directly from `RHLocalGroupsBase` instead.

indico/modules/groups/controllers.py
```diff
@@ -54,8 +63,11 @@ def _process(self):
         return WPGroupsAdmin.render_template('groups.html', groups=groups, providers=providers, form=form,
                                              search_results=search_results, provider_titles=provider_titles)
 
+    def _check_access(self):
+        RHLocalGroupConfig._check_access(self)
```
ThiefMaster: ⚠️ Always call the methods from both base classes, otherwise you lose that functionality - so in this case, these endpoints would suddenly be completely unrestricted!

(also in the RHs below)

indico/modules/groups/core.py
```diff
@@ -138,8 +139,9 @@ def search(cls, name, exact=False, providers=None):
         else:
             criterion = db.func.lower(LocalGroup.name).contains(name.lower())
         result = set()
-        if providers is None or 'indico' in providers:
-            result |= {g.proxy for g in LocalGroup.find(criterion)}
+        if config.LOCAL_GROUPS:
+            if providers is None or 'indico' in providers:
```
ThiefMaster: these two lines can be merged into one

indico/modules/groups/core.py
```diff
@@ -162,7 +164,9 @@ def name(self):
     @cached_property
     def group(self):
         """The underlying :class:`.LocalGroup`"""
-        return LocalGroup.get(self.id)
+        if config.LOCAL_GROUPS:
```
ThiefMaster: I think doing it the other way round would be clearer, i.e. if local groups are disabled, return none early.

BTW, I'd add a comment here mentioning why we are doing this:

```
# pretend local groups do not exist if they are disabled
# this way they'll be rejected in acl fields
```

indico/modules/groups/core.py
```diff
@@ -174,7 +178,9 @@ def as_legacy_group(self):
         return LocalGroupWrapper(self.id)
 
     def has_member(self, user):
-        return user and self.group in user.local_groups
+        if config.LOCAL_GROUPS:
```
ThiefMaster: like above, I'd return false if local groups are disabled and then continue with the normal logic afterwards

indico/modules/groups/controllers.py
```diff
@@ -31,7 +32,16 @@
...
+class RHLocalGroupsBase(RHAdminBase):
+    """Class for checking Local group configuration"""
+
+    def _check_access(self):
+        RHAdminBase._check_access(self)
+        if not config.LOCAL_GROUPS:
+            raise Forbidden('Local groups are disabled.')
+
+
+class RHGroups(RHLocalGroupsBase):
```
ThiefMaster: I just realized that this page is not only for local groups but all types of groups. So disabling everything through inheritance won't work... Sorry for that!

So the check needs to be in `RHGroupBase`, by checking if `group.is_local` there. and `RHLocalGroupBase` (unconditionally since that one is only used for local ones).

indico/modules/groups/__init__.py
```diff
@@ -21,7 +22,7 @@
 
 @signals.menu.items.connect_via('admin-sidemenu')
 def _extend_admin_menu(sender, **kwargs):
-    if session.user.is_admin:
+    if session.user.is_admin and config.LOCAL_GROUPS:
```
ThiefMaster: For the reason mentioned in my other comment, this change needs to be reverted as well since there can still be remote groups. :/

indico/modules/groups/controllers.py
```diff
@@ -39,7 +40,11 @@ def _process(self):
         groups = [g.proxy for g in query]
         providers = [p for p in multipass.identity_providers.itervalues() if p.supports_groups]
         form = SearchForm(obj=FormDefaults(exact=True))
-        if not providers:
+        local_group = True
+        if not providers and not config.LOCAL_GROUPS:
+            local_group = False
+            del form.provider
```
ThiefMaster: You're not showing the form in this case so this isn't needed

indico/modules/groups/controllers.py
```diff
@@ -39,7 +40,11 @@ def _process(self):
         groups = [g.proxy for g in query]
         providers = [p for p in multipass.identity_providers.itervalues() if p.supports_groups]
         form = SearchForm(obj=FormDefaults(exact=True))
-        if not providers:
+        local_group = True
```
ThiefMaster: `groups_enabled`

indico/modules/groups/templates/groups.html
```diff
@@ -31,24 +33,11 @@
...
-                            {%- trans %}Create new group{% endtrans -%}
-                        </a>
-                    </div>
-                </div>
-            </div>
-            <div class="i-box-table-widget">
-                {{ group_table(groups, show_members=true) }}
-            </div>
-        </div>
+    {% if not providers and not local_group %}
```
ThiefMaster: Checking for the new `groups_enabled` flag here should be enough

indico/modules/groups/templates/groups.html
```diff
@@ -31,24 +33,11 @@
...
+        {% call message_box('info', fixed_width=true) %}
+            {% trans %}Local groups are disabled.{% endtrans %}
```
ThiefMaster: > Group functionality is not available as local groups are disabled and there are no providers for remote groups configured.

indico/modules/groups/controllers.py
```diff
@@ -39,7 +40,11 @@ def _process(self):
...
+        if not providers and not config.LOCAL_GROUPS:
+            local_group = False
+            del form.provider
+        elif not providers:
             del form.provider
         else:
             form.provider.choices = ([('', _('All')), ('indico', _('Local Groups'))] +
```
ThiefMaster:
```python
choices = [('', _('All'))]
if config.LOCAL_GROUPS:
    choices.append(('indico', _('Local Groups')))
choices += [(p.name, p.title) for p in sorted(providers, key=attrgetter('title'))]
form.provider.choices = choices
```

indico/modules/groups/controllers.py
```diff
@@ -39,11 +40,18 @@ def _process(self):
         groups = [g.proxy for g in query]
         providers = [p for p in multipass.identity_providers.itervalues() if p.supports_groups]
         form = SearchForm(obj=FormDefaults(exact=True))
-        if not providers:
+        groups_enabled = True
+        local_groups = config.LOCAL_GROUPS
```
ThiefMaster: Just use `config.LOCAL_GROUPS` directly below

indico/modules/groups/controllers.py
```diff
@@ -52,7 +60,8 @@ def _process(self):
         provider_titles = {p.name: p.title for p in multipass.identity_providers.itervalues()}
         provider_titles[None] = _('Local')
         return WPGroupsAdmin.render_template('groups.html', groups=groups, providers=providers, form=form,
-                                             search_results=search_results, provider_titles=provider_titles)
+                                             search_results=search_results, provider_titles=provider_titles,
+                                             groups_enabled=groups_enabled, local_groups=local_groups)
```
ThiefMaster: `local_groups` is not needed: in templates we have a global `indico_config` so you can access `indico_config.LOCAL_GROUPS` there

PR #4266, opened 2020-01-30
Allow submitters to access contributions

### LINKED ISSUE #4224: Paper Editing Timeline: Allow the submitter to access the timeline even when contributions are not published

panagiotappl:
closes #4224

### DIFF
--- a/indico/modules/events/contributions/__init__.py
+++ b/indico/modules/events/contributions/__init__.py
@@ -105,8 +105,9 @@ def _extend_event_menu(sender, **kwargs):
     from indico.modules.events.layout.util import MenuEntryData
 
     def _visible_my_contributions(event):
-        published = contribution_settings.get(event, 'published')
-        return session.user and published and has_contributions_with_user_as_submitter(event, session.user)
+        if not session.user:
+            return False
+        return has_contributions_with_user_as_submitter(event, session.user)
 
     def _visible_list_of_contributions(event):
         published = contribution_settings.get(event, 'published')
--- a/indico/modules/events/contributions/templates/display/contribution_display.html
+++ b/indico/modules/events/contributions/templates/display/contribution_display.html
@@ -141,7 +141,7 @@
 {%- endblock %}
 
 {% block content -%}
-    {%- if not published -%}
+    {%- if not published and contribution.event.can_manage(session.user) and not contribution.is_user_associated(session.user) -%}
         {% call message_box('warning') -%}
             {% trans %}
                 The contributions of this event have not been published yet. You can only access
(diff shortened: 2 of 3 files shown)


indico/modules/events/contributions/templates/display/contribution_display.html
```diff
@@ -145,7 +145,7 @@
         {% call message_box('warning') -%}
             {% trans %}
                 The contributions of this event have not been published yet. You can only access
-                this page because you are an event manager.
+                this page because you are an event manager or you are associated with this contribution.
```
ThiefMaster: I don't think this message should be shown in this case at all - it's just going to confuse users...

panagiotappl: So, only show it to managers and the submitters see it as a regurlar page?

ThiefMaster: I think that's the most logical solution: For a manager it's important to know because they may otherwise think the link can be shared with others or forget to publish in the end. But for a user it's pretty much expected that they can see their own contribution even if the full list of contributions hasn't been released yet.

indico/modules/events/contributions/templates/display/contribution_display.html
```diff
@@ -141,7 +141,7 @@
 {%- endblock %}
 
 {% block content -%}
-    {%- if not published -%}
+    {%- if not published and contribution.event.can_manage(session.user) -%}
```
ThiefMaster: That'll show it for a manager even if they are associated with it.. I'd add the association check here as well, i.e. `not published and manager and not associated`

indico/modules/events/contributions/__init__.py
```diff
@@ -106,7 +106,9 @@ def _extend_event_menu(sender, **kwargs):
 
     def _visible_my_contributions(event):
         published = contribution_settings.get(event, 'published')
-        return session.user and published and has_contributions_with_user_as_submitter(event, session.user)
+        if not session.user:
+            return False
+        return has_contributions_with_user_as_submitter(event, session.user) or published
```
ThiefMaster: check for `published` first - no need to send the extra query in that case

ThiefMaster: I don't think `has_contributions_with_user_as_submitter` is correct - AFAIK we allow any author to submit editables (even though I think there's an issue to change this?)

It should be possible to create a util that has similar logic to check for the whole event, just like `is_user_associated` does for a single contribution

panagiotappl: This about showing "My Contributions" menu entry to users who are contribution authors though, right?

indico/modules/events/contributions/__init__.py
```diff
@@ -106,7 +106,9 @@ def _extend_event_menu(sender, **kwargs):
 
     def _visible_my_contributions(event):
         published = contribution_settings.get(event, 'published')
-        return session.user and published and has_contributions_with_user_as_submitter(event, session.user)
+        if not session.user or not published:
```
ThiefMaster: i think the published check in here can completely go away now

PR #4270, opened 2020-02-03
Ready for review conditions

### LINKED ISSUE #4221: Paper editing: Define ready for review file types
Create interface to define which file types are needed to allow editing of given paper. Allow using logical operators (OR, AND) e.g. 'pdf' OR 'source'

mic4ael:
closes #4221


indico/modules/events/editing/__init__.py
```diff
@@ -14,12 +14,16 @@
...
 from indico.modules.events.features.base import EventFeature
 from indico.modules.events.models.events import Event, EventType
+from indico.modules.events.settings import EventSettingsProxy
 from indico.util.i18n import _
 from indico.web.flask.util import url_for
 from indico.web.menu import SideMenuItem
 
 
 logger = Logger.get('events.editing')
+editing_settings = EventSettingsProxy('editing', {
```
ThiefMaster: I'd move this to a `settings.py` (we did the same in many other modules except the early ones)

indico/modules/events/editing/blueprint.py
```diff
@@ -19,13 +19,19 @@
...
+                 frontend.RHManageEditingReviewConditions)
 _bp.add_url_rule('/contributions/<int:contrib_id>/editing/<any(paper):type>', 'editable', frontend.RHEditableTimeline)
 _bp.add_url_rule('/contributions/<int:contrib_id>/editing/<any(paper):type>/<int:revision_id>/files.zip',
                  'revision_files_export', timeline.RHExportRevisionFiles)
 _bp.add_url_rule('/contributions/<int:contrib_id>/editing/<any(paper):type>/<int:revision_id>/<int:file_id>/<filename>',
                  'download_file', timeline.RHDownloadRevisionFile)
 
 # Event-level APIs
+_bp.add_url_rule('/editing/api/review-conditions', 'api_review_conditions', management.RHEditingReviewConditions,
+                 methods=('GET', 'POST',))
```
ThiefMaster: no trailing comma needed

indico/modules/events/editing/client/js/components/ConditionInfo.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+import ReviewConditionForm from './ReviewConditionForm';
+
+import './ConditionInfo.module.scss';
+
+export default function ConditionInfo({condition, index, onUpdate}) {
+  const [isEditing, setIsEditing] = useState(false);
+  const [isDeleting, setIsDeleting] = useState(false);
+  const {eventId} = useSelector(state => state.staticData);
+  const url = editReviewConditionURL({confId: eventId, index});
```
ThiefMaster: Using the index seems pretty dangerous here. That means if someone else deleted a rule after you opened the page, you'll delete/edit a different rule!

I'd probably just send all data even if that means possibly overwriting someone else's changes, but you could assign a UUID to each rule and use that instead of the index.

indico/modules/events/editing/client/js/components/ConditionInfo.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const deleteCondition = async () => {
+    try {
+      await indicoAxios.delete(url);
+      if (onUpdate) {
+        onUpdate();
+      }
```
ThiefMaster: I don't think this should be in the try block - unless it can throw something as well.

indico/modules/events/editing/client/js/components/ConditionInfo.jsx
```diff
@@ -0,0 +1,103 @@
...
+        <Icon color="blue" name="pencil" onClick={() => setIsEditing(true)} />
+        <Icon color="red" name="trash" onClick={() => setIsDeleting(true)} />
+      </div>
+      <RequestConfirm
+        header={Translate.string('Delete review condition')}
+        confirmText={Translate.string('Yes')}
+        cancelText={Translate.string('No')}
+        onClose={() => setIsDeleting(false)}
+        content={
+          <div className="content">
```
ThiefMaster: Where's that class coming from? SUI?

indico/modules/events/editing/client/js/components/ReviewConditionsManager.jsx
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  const eventConditions = eventConditionsSetting || lastData;
+  if (loading && !lastData) {
+    return <Loader inline="centered" active />;
+  } else if (!eventConditions) {
+    return null;
+  }
+
+  const reviewConditions = eventConditions.map(condition =>
```
ThiefMaster: I think this function could use a comment... it's not very obvious what it's doing (replacing the ID with the actual file type?)

indico/modules/events/editing/client/js/components/ReviewConditionsManager.jsx
```diff
@@ -0,0 +1,104 @@
...
+      </Message>
+      {eventConditions.length === 0 && (
+        <Message info>
+          <Translate>No reviewing conditions have been defined</Translate>
+        </Message>
+      )}
+      <div>
+        {reviewConditions.map((reviewCondition, index) => (
+          // eslint-disable-next-line react/no-array-index-key
+          <React.Fragment key={`condition-${index}`}>
```
ThiefMaster: if you go for the uuid, you can use that in the key

indico/modules/events/editing/client/js/components/ReviewConditionsManager.jsx
```diff
@@ -0,0 +1,104 @@
...
+            <div styleName="condition-row">
+              <ConditionInfo condition={reviewCondition} index={index} onUpdate={() => reFetch()} />
+            </div>
+            {index !== reviewConditions.length - 1 && (
+              <Divider horizontal>
+                <Translate>Or</Translate>
```
ThiefMaster: `OR` / `or`?

indico/modules/events/editing/client/js/reviewConditions.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  const store = createReduxStore(
+    'editing-review-conditions',
+    {},
+    {
+      staticData: {
+        eventId,
+        fileTypes,
+      },
```
ThiefMaster: Is it worth using redux just for this instead of passing a prop?

indico/web/client/js/react/forms/validators.js
```diff
@@ -55,7 +57,11 @@ function url(value) {
 }
 
 function required(value) {
-  return value || value === 0 ? undefined : Translate.string('This field is required.');
+  const errorMsg = Translate.string('This field is required.');
+  if (_.isArray(value)) {
```
ThiefMaster: `Array.isArray(value)` - doesn't need lodash anymore ;)

indico/modules/events/editing/client/js/components/ConditionInfo.jsx
mic4ael: Yes. It is here to keep the correct styles

indico/modules/events/editing/client/js/components/ConditionInfo.jsx
mic4ael: Hm, yeah I was not very happy with the fact that I am relying on the index. Maybe having an object with `uuid` keys would be much better in this case.

indico/modules/events/editing/client/js/reviewConditions.js
mic4ael: Yeah, I would have to pass it down to different components so I thought that using redux would be cleaner here

indico/core/settings/converters.py
```diff
@@ -144,3 +146,13 @@ def to_python(self, value):
         if not value:
             return []
         return self.collection_class(self.model.query.filter(self.column.in_(value)))
+
+
+class OrderedDictConverter(SettingConverter):
+    @staticmethod
+    def from_python(value):
+        return json.dumps(value)
```
ThiefMaster: I really don't like the idea of storing nested JSON... I'd rather convert it to a list of lists...

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -87,3 +96,43 @@ def _process_DELETE(self):
...
+        editing_settings.set(self.event, 'review_conditions', new_conditions)
+        return '', 204
+
+
+class RHEditingEditReviewCondition(RHEditingManagementBase):
+    def _process_args(self):
+        RHEditingManagementBase._process_args(self)
+        uuid = request.view_args['uuid']
+        try:
+            UUID(uuid)
```
ThiefMaster: tip, you can use `<uuid:uuid>` in the url rule, then it'll only match on  valid uuids

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -87,3 +96,43 @@ def _process_DELETE(self):
...
+            raise NotFound
+
+        conditions = editing_settings.get(self.event, 'review_conditions')
+        if uuid not in conditions:
+            raise NotFound
+        self.uuid = uuid
+
+    def _process_DELETE(self):
+        new_conditions = OrderedDict(editing_settings.get(self.event, 'review_conditions'))
+        del new_conditions[self.uuid]
```
ThiefMaster: mutating the existing dict is fine here

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -87,3 +96,43 @@ def _process_DELETE(self):
...
+        editing_settings.set(self.event, 'review_conditions', new_conditions)
+        return '', 204
+
+    @use_rh_kwargs(EditingReviewConditionArgs)
```
ThiefMaster: I'd always use `partial=True` for PATCH endpoints

indico/core/settings/converters.py
mic4ael: So, does it really hurt? By having an actual `dict` here the `update/delete` operations are much easier than iterating over the entire list of lists

indico/modules/events/editing/controllers/backend/management.py
mic4ael: Ah, true.. I forgot about it.. (we could change other places that don't use it)

indico/modules/events/editing/controllers/backend/management.py
mic4ael: Ahh, true. I was thinking that we use redis as the settings cache (I see that the cache is only per request thus there is no problem with mutating it here).

indico/modules/events/editing/controllers/backend/management.py
mic4ael: why? This RH always expects `file_types` to be present in the request

indico/core/settings/converters.py
ThiefMaster: i mean for storage - nothing wrong with the dict itself

mic4ael: Ah.. I see

indico/modules/events/editing/controllers/backend/management.py
ThiefMaster: Indeed, doesn't make much sense then.

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -87,3 +95,38 @@ def _process_DELETE(self):
...
+
+
+class RHEditingEditReviewCondition(RHEditingManagementBase):
+    def _process_args(self):
+        RHEditingManagementBase._process_args(self)
+        uuid = request.view_args['uuid']
+        conditions = editing_settings.get(self.event, 'review_conditions')
+        if unicode(uuid) not in conditions:
+            raise NotFound
+        self.uuid = unicode(uuid)
```
ThiefMaster: I've move this before the check and then just use `self.uuid not in conditions` there

indico/modules/events/editing/schemas.py
```diff
@@ -218,3 +224,21 @@ def _check_if_can_unset_or_delete(self, publishable):
...
+
+    @validates('file_types')
+    def _validate_file_types(self, file_types):
+        event = self.context['event']
+        event_file_types = {ft.id for ft in event.editing_file_types}
+        if set(file_types) - event_file_types:
+            raise ValidationError(_('Invalid file type used'))
+
+        event_conditions = editing_settings.get(event, 'review_conditions')
+        if file_types in event_conditions.values():
```
ThiefMaster: Doesn't this depend on the order in which you added the conditions? Or are we sorting them somewhere?

PR #4273, opened 2020-02-04
Add category groups/roles

### LINKED ISSUE #3346: Abstracts: UX improvements in role management
It's currently quite cumbersome to set up a list of roles of a large conference. In addition to adding users one by one, the manager will have to add the user once for each track/role. This can get quite time-expensive.

There are currently two ideas:
 * Allowing for (local) groups to be added (#3040 will help);
 * Allowing for some kind of drag and drop mechanism, or similar, to move/copy people across lists;

### LINKED ISSUE #4271: Add Category based roles/groups

ParthS007:
Closes #4271 
Closes #3346

### DIFF
--- a/indico/migrations/versions/20200204_1201_4d263fa78830_add_category_roles_table.py
+++ b/indico/migrations/versions/20200204_1201_4d263fa78830_add_category_roles_table.py
@@ -1,0 +1,46 @@
+"""Add category roles table
+
+Revision ID: 4d263fa78830
+Revises: 6a185fdcd4ee
+Create Date: 2020-02-04 12:01:02.554724
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '4d263fa78830'
+down_revision = '6a185fdcd4ee'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'roles',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('category_id', sa.Integer(), nullable=False, index=True),
+        sa.Column('name', sa.String(), nullable=False),
+        sa.Column('code', sa.String(), nullable=False),
+        sa.Column('color', sa.String(), nullable=False),
+        sa.Index(None, 'category_id', 'code', unique=True),
+        sa.CheckConstraint('code = upper(code)', name='uppercase_code'),
+        sa.ForeignKeyConstraint(['category_id'], ['categories.categories.id']),
+        sa.PrimaryKeyConstraint('id'),
+        schema='categories'
+    )
+    op.create_table(
+        'role_members',
+        sa.Column('role_id', sa.Integer(), nullable=False, index=True),
+        sa.Column('user_id', sa.Integer(), nullable=False, index=True),
+        sa.ForeignKeyConstraint(['role_id'], ['categories.roles.id']),
+        sa.ForeignKeyConstraint(['user_id'], ['users.users.id']),
+        sa.PrimaryKeyConstraint('role_id', 'user_id'),
+        schema='categories'
+    )
+
+
+def downgrade():
+    op.drop_table('role_members', schema='categories')
+    op.drop_table('roles', schema='categories')
--- a/indico/modules/categories/blueprint.py
+++ b/indico/modules/categories/blueprint.py
@@ -18,11 +18,14 @@ from indico.modules.categories.controllers.display import (RHCategoryCalendarVie
                                                            RHExportCategoryICAL, RHReachableCategoriesInfo,
                                                            RHShowFutureEventsInCategory, RHShowPastEventsInCategory,
                                                            RHSubcatInfo, RHXMLExportCategoryInfo)
-from indico.modules.categories.controllers.management import (RHCreateCategory, RHDeleteCategory, RHDeleteEvents,
-                                                              RHDeleteSubcategories, RHManageCategoryContent,
-                                                              RHManageCategoryIcon, RHManageCategoryLogo,
-                                                              RHManageCategoryProtection, RHManageCategorySettings,
-                                                              RHMoveCategory, RHMoveEvents, RHMoveSubcategories,
+from indico.modules.categories.controllers.management import (RHAddCategoryRole, RHAddCategoryRoleMembers,
+                                                              RHCategoryRoles, RHCreateCategory, RHDeleteCategory,
+                                                              RHDeleteCategoryRole, RHDeleteEvents,
+                                                              RHDeleteSubcategories, RHEditCategoryRole,
+                                                              RHManageCategoryContent, RHManageCategoryIcon,
+                                                              RHManageCategoryLogo, RHManageCategoryProtection,
+                                                              RHManageCategorySettings, RHMoveCategory, RHMoveEvents,
+                                                              RHMoveSubcategories, RHRemoveCategoryRoleMember,
                                                               RHSortSubcategories, RHSplitCategory)
 from indico.modules.users import User
 from indico.web.flask.util import make_compat_redirect_func, redirect_view, url_for
@@ -45,6 +48,15 @@ _bp.add_url_rule('/manage/logo', 'manage_logo', RHManageCategoryLogo, methods=('
 _bp.add_url_rule('/manage/move', 'move', RHMoveCategory, methods=('POST',))
 _bp.add_url_rule('/manage/protection', 'manage_protection', RHManageCategoryProtection, methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/settings', 'manage_settings', RHManageCategorySettings, methods=('POST', 'GET'))
+
+# Role management
+_bp.add_url_rule('/manage/roles', 'manage_roles', RHCategoryRoles, methods=('POST', 'GET'))
+_bp.add_url_rule('/manage/roles/create', 'add_role', RHAddCategoryRole, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/roles/<int:role_id>/edit', 'edit_role', RHEditCategoryRole, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/roles/<int:role_id>', 'delete_role', RHDeleteCategoryRole, methods=('DELETE',))
+_bp.add_url_rule('/manage/roles/<int:role_id>/members', 'add_members', RHAddCategoryRoleMembers, methods=('POST',))
+_bp.add_url_rule('/manage/roles/<int:role_id>/members/<int:user_id>', 'remove_member', RHRemoveCategoryRoleMember,
+                 methods=('DELETE',))
 
 # Event management
 _bp.add_url_rule('/manage/events/delete', 'delete_events', RHDeleteEvents, methods=('GET', 'POST'))
--- a/indico/modules/categories/client/js/management.js
+++ b/indico/modules/categories/client/js/management.js
@@ -458,8 +458,8 @@
 
   function setupRolesButtons() {
     $('#event-roles').on('click', '.js-add-members', function(evt) {
-      var $this = $(this);
       evt.stopPropagation();
+      var $this = $(this);
       $('<div>')
         .principalfield({
           multiChoice: true,
--- a/indico/modules/categories/client/js/management.js
+++ b/indico/modules/categories/client/js/management.js
@@ -435,4 +435,55 @@
       },
     });
   }
+
+  function setupRolesToggle() {
+    var $roles = $('#event-roles');
+    $roles.on('click', '.toggle-members', function() {
+      var $row = $(this)
+        .closest('tr')
+        .next('tr')
+        .find('.slide');
+      $row.css('max-height', $row[0].scrollHeight + 'px');
+      $row.toggleClass('open close');
+    });
+
+    $roles.on('indico:htmlUpdated', function() {
+      $(this)
+        .find('.slide')
+        .each(function() {
+          $(this).css('max-height', this.scrollHeight + 'px');
+        });
+    });
+  }
+
+  function setupRolesButtons() {
+    $('#event-roles').on('click', '.js-add-members', function(evt) {
+      var $this = $(this);
+      evt.stopPropagation();
+      $('<div>')
+        .principalfield({
+          multiChoice: true,
+          onAdd: function(users) {
+            $.ajax({
+              url: $this.data('href'),
+              method: $this.data('method'),
+              data: JSON.stringify({users: users}),
+              dataType: 'json',
+              contentType: 'application/json',
+              error: handleAjaxError,
+              complete: IndicoUI.Dialogs.Util.progress(),
+              success: function(data) {
+                updateHtml($this.data('update'), data);
+              },
+            });
+          },
+        })
+        .principalfield('choose');
+    });
+  }
+
+  global.setupCategoryRolesList = function setupCategoryRolesList() {
+    setupRolesToggle();
+    setupRolesButtons();
+  };
 })(window);
--- a/indico/modules/categories/controllers/management.py
+++ b/indico/modules/categories/controllers/management.py
@@ -8,6 +8,7 @@
 from __future__ import unicode_literals
 
 import os
+import random
 from io import BytesIO
 
 from flask import flash, redirect, request, session
@@ -19,17 +20,22 @@ from indico.core.db import db
 from indico.modules.categories import logger
 from indico.modules.categories.controllers.base import RHManageCategoryBase
 from indico.modules.categories.forms import (CategoryIconForm, CategoryLogoForm, CategoryProtectionForm,
-                                             CategorySettingsForm, CreateCategoryForm, SplitCategoryForm)
+                                             CategorySettingsForm, CreateCategoryForm, RoleForm, SplitCategoryForm)
 from indico.modules.categories.models.categories import Category
+from indico.modules.categories.models.roles import CategoryRole
 from indico.modules.categories.operations import create_category, delete_category, move_category, update_category
 from indico.modules.categories.util import get_image_data
 from indico.modules.categories.views import WPCategoryManagement
 from indico.modules.events import Event
+from indico.modules.events.roles.util import get_role_colors, serialize_role
 from indico.modules.events.util import update_object_principals
 from indico.modules.rb.models.reservations import Reservation, ReservationLink
+from indico.modules.users import User
 from indico.util.fs import secure_filename
 from indico.util.i18n import _, ngettext
 from indico.util.string import crc32
+from indico.util.user import principal_from_fossil
+from indico.web.flask.templating import get_template_module
 from indico.web.flask.util import url_for
 from indico.web.forms.base import FormDefaults
 from indico.web.util import jsonify_data, jsonify_form, jsonify_template, url_for_index
@@ -37,6 +43,22 @@ from indico.web.util import jsonify_data, jsonify_form, jsonify_template, url_fo
 
 CATEGORY_ICON_DIMENSIONS = (16, 16)
 MAX_CATEGORY_LOGO_DIMENSIONS = (200, 200)
+
+
+def _get_roles(category):
+    return (CategoryRole.query.with_parent(category)
+            .options(joinedload('members'))
+            .all())
+
+
+def _render_roles(category):
+    tpl = get_template_module('events/roles/_roles.html')
+    return tpl.render_roles(_get_roles(category), email_button=False)
+
+
+def _render_role(role, collapsed=True):
+    tpl = get_template_module('events/roles/_roles.html')
+    return tpl.render_role(role, collapsed=collapsed, email_button=False)
 
 
 class RHManageCategoryContent(RHManageCategoryBase):
@@ -402,3 +424,94 @@ class RHMoveEvents(RHManageCategorySelectedEventsBase):
                        'You have moved {count} events to the category "{cat}"', len(self.events))
               .format(count=len(self.events), cat=self.target_category.title), 'success')
         return jsonify_data(flash=False)
+
+
+class RHCategoryRoles(RHManageCategoryBase):
+    """Category role management"""
+
+    def _process(self):
+        return WPCategoryManagement.render_template('management/roles.html', self.category, 'roles',
+                                                    roles=_get_roles(self.category))
+
+
+class RHAddCategoryRole(RHManageCategoryBase):
+    """Add a new category role"""
+
+    def _process(self):
+        form = RoleForm(category=self.category, color=self._get_color())
+        if form.validate_on_submit():
+            role = CategoryRole(category=self.category)
+            form.populate_obj(role)
+            db.session.flush()
+            logger.info('Category role %r created by %r', role, session.user)
+            return jsonify_data(html=_render_roles(self.category), role=serialize_role(role))
+        return jsonify_form(form)
+
+    def _get_color(self):
+        used_colors = {role.color for role in self.category.roles}
+        unused_colors = set(get_role_colors()) - used_colors
+        return random.choice(tuple(unused_colors) or get_role_colors())
+
+
+class RHManageCategoryRole(RHManageCategoryBase):
+    """Base class to manage a specific category role"""
+
+    normalize_url_spec = {
+        'locators': {
+            lambda self: self.role
+        }
+    }
+
+    def _process_args(self):
+        RHManageCategoryBase._process_args(self)
+        self.role = CategoryRole.get_one(request.view_args['role_id'])
+
+
+class RHEditCategoryRole(RHManageCategoryRole):
+    """Edit an category role"""
+
+    def _process(self):
+        form = RoleForm(obj=self.role, category=self.category)
+        if form.validate_on_submit():
+            form.populate_obj(self.role)
+            db.session.flush()
+            logger.info('Category role %r updated by %r', self.role, session.user)
+            return jsonify_data(html=_render_role(self.role))
+        return jsonify_form(form)
+
+
+class RHDeleteCategoryRole(RHManageCategoryRole):
+    """Delete an category role"""
+
+    def _process(self):
+        db.session.delete(self.role)
+        logger.info('Category role %r deleted by %r', self.role, session.user)
+        return jsonify_data(html=_render_roles(self.category))
+
+
+class RHRemoveCategoryRoleMember(RHManageCategoryRole):
+    """Remove a user from an category role"""
+
+    normalize_url_spec = dict(RHManageCategoryRole.normalize_url_spec, preserved_args={'user_id'})
+
+    def _process_args(self):
+        RHManageCategoryRole._process_args(self)
+        self.user = User.get_one(request.view_args['user_id'])
+
+    def _process(self):
+        if self.user in self.role.members:
+            self.role.members.remove(self.user)
+            logger.info('User %r removed from role %r by %r', self.user, self.role, session.user)
+        return jsonify_data(html=_render_role(self.role, collapsed=False))
+
+
+class RHAddCategoryRoleMembers(RHManageCategoryRole):
+    """Add users to an category role"""
+
+    def _process(self):
+        for data in request.json['users']:
+            user = principal_from_fossil(data, allow_pending=True, allow_groups=False)
+            if user not in self.role.members:
+                self.role.members.add(user)
+                logger.info('User %r added to role %r by %r', user, self.role, session.user)
+        return jsonify_data(html=_render_role(self.role, collapsed=False))
--- a/indico/modules/categories/controllers/management.py
+++ b/indico/modules/categories/controllers/management.py
@@ -17,18 +17,18 @@ from sqlalchemy.orm import joinedload, load_only, undefer_group
 from werkzeug.exceptions import BadRequest, Forbidden
 
 from indico.core.db import db
+from indico.core.permissions import get_principal_permissions, update_permissions
 from indico.modules.categories import logger
 from indico.modules.categories.controllers.base import RHManageCategoryBase
 from indico.modules.categories.forms import (CategoryIconForm, CategoryLogoForm, CategoryProtectionForm,
-                                             CategorySettingsForm, CreateCategoryForm, RoleForm, SplitCategoryForm)
+                                             CategoryRoleForm, CategorySettingsForm, CreateCategoryForm,
+                                             SplitCategoryForm)
 from indico.modules.categories.models.categories import Category
 from indico.modules.categories.models.roles import CategoryRole
 from indico.modules.categories.operations import create_category, delete_category, move_category, update_category
-from indico.modules.categories.util import get_image_data
+from indico.modules.categories.util import get_image_data, serialize_category_role
 from indico.modules.categories.views import WPCategoryManagement
 from indico.modules.events import Event
-from indico.modules.events.roles.util import get_role_colors, serialize_role
-from indico.modules.events.util import update_object_principals
 from indico.modules.rb.models.reservations import Reservation, ReservationLink
 from indico.modules.users import User
 from indico.util.fs import secure_filename
@@ -38,6 +38,8 @@ from indico.util.user import principal_from_fossil
 from indico.web.flask.templating import get_template_module
 from indico.web.flask.util import url_for
 from indico.web.forms.base import FormDefaults
+from indico.web.forms.colors import get_role_colors
+from indico.web.forms.fields.principals import serialize_principal
 from indico.web.util import jsonify_data, jsonify_form, jsonify_template, url_for_index
 
 
@@ -46,9 +48,7 @@ MAX_CATEGORY_LOGO_DIMENSIONS = (200, 200)
 
 
 def _get_roles(category):
-    return (CategoryRole.query.with_parent(category)
-            .options(joinedload('members'))
-            .all())
+    return (CategoryRole.query.with_parent(category).options(joinedload('members')).all())
 
 
 def _render_roles(category):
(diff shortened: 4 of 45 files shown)


indico/migrations/versions/20200204_1201_4d263fa78830_add_category_roles_table.py
```diff
@@ -0,0 +1,46 @@
...
+        sa.Column('category_id', sa.Integer(), nullable=False, index=True),
+        sa.Column('name', sa.String(), nullable=False),
+        sa.Column('code', sa.String(), nullable=False),
+        sa.Column('color', sa.String(), nullable=False),
+        sa.CheckConstraint('code = upper(code)', name='uppercase_code'),
+        sa.ForeignKeyConstraint(['category_id'], ['categories.categories.id']),
+        sa.PrimaryKeyConstraint('id'),
+        schema='categories'
+    )
+    op.create_index(None, 'roles', ['category_id', 'code'], unique=True, schema='categories')
```
ThiefMaster: you can add this to the create_table call above instead:

    sa.Index(None, 'category_id', 'code', unique=True),

indico/modules/events/models/roles.py
```diff
@@ -21,6 +21,7 @@ class EventRole(db.Model):
 
     is_group = False
     is_event_role = True
+    is_category_role = False
```
ThiefMaster: let's add this to the other objects that have `is_event_role = False` as well.

right now neither of them are being read, but that way we are consistent (and either keep them or remove both of them later)

indico/modules/categories/models/roles.py
```diff
@@ -0,0 +1,123 @@
...
+    color = db.Column(
+        db.String,
+        nullable=False
+    )
+
+    category = db.relationship(
+        'Category',
+        lazy=True,
+        backref=db.backref(
+            'category_roles',
```
ThiefMaster: just `'roles'` is enough (it's on Category anyway)

indico/migrations/versions/20200204_1201_4d263fa78830_add_category_roles_table.py
ParthS007: Tested with `downgrade` and `upgrade`

indico/modules/categories/blueprint.py
```diff
@@ -45,6 +48,15 @@ def _redirect_event_creation(category_id, event_type):
 _bp.add_url_rule('/manage/move', 'move', RHMoveCategory, methods=('POST',))
 _bp.add_url_rule('/manage/protection', 'manage_protection', RHManageCategoryProtection, methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/settings', 'manage_settings', RHManageCategorySettings, methods=('POST', 'GET'))
+_bp.add_url_rule('/manage/roles', 'manage_roles', RHCategoryRoles, methods=('POST', 'GET'))
```
ThiefMaster: I'd move this one down to "Role management" (and lowercase the m in that comment to be consistent with the others)

indico/modules/categories/client/js/management.js
```diff
@@ -435,4 +435,55 @@
...
+            });
+          },
+        })
+        .principalfield('choose');
+    });
+  }
+
+  global.setupCategoryRolesList = function setupCategoryRolesList() {
+    setupToggle();
+    setupButtons();
```
ThiefMaster: what about `setupRolesToggle` and `setupRolesButtons`?

indico/modules/categories/templates/management/roles.html
```diff
@@ -0,0 +1,23 @@
+{% extends 'categories/management/base.html' %}
+{% from 'categories/management/_roles.html' import render_roles %}
```
ThiefMaster: Didn't we discuss yesterday that we want to reuse the one from events here (and add a new macro param to hide stuff only needed in events, like the email buttons)?

indico/modules/categories/views.py
```diff
@@ -88,3 +88,13 @@ def _get_breadcrumbs(self):
 
 class WPCategoryStatistics(WPCategory):
     bundles = ('statistics.js', 'statistics.css')
+
+
+class WPCategoryRoles(WPCategory):
```
ThiefMaster: You can use `WPCategoryManagement` directly; no need for a custom WP class here

indico/modules/categories/templates/management/roles.html
```diff
@@ -0,0 +1,23 @@
+{% extends 'categories/management/base.html' %}
+{% from 'events/roles/_roles.html' import render_roles %}
+
+{% block title %}{% trans %}Roles Setup{% endtrans %}{% endblock %}
+
+{% block content %}
+    <table class="i-table-widget multi-tbody" id="event-roles">
+        {{ render_roles(roles, collapsed=true, email_button=False) }}
```
ThiefMaster: `false` (jinja uses lowercase true/false/none, even though the python-style versions work as well)

indico/modules/events/roles/templates/_roles.html
```diff
@@ -13,7 +13,7 @@
 {%- endmacro %}
 
 
-{% macro render_role(role, collapsed=true) -%}
+{% macro render_role(role, collapsed=collapsed, email_button=email_button) -%}
```
ThiefMaster: `collapsed=true, email_button=true`

indico/modules/events/roles/templates/_roles.html
```diff
@@ -1,7 +1,7 @@
-{% macro render_roles(roles) -%}
+{% macro render_roles(roles, collapsed, email_button) -%}
     {% for role in roles|sort(attribute='code') %}
         <tbody id="event-role-{{ role.id }}">
-            {{ render_role(role) }}
+            {{ render_role(role, collapsed, email_button) }}
```
ThiefMaster: `render_role(role, email_button=email_button)`

indico/modules/events/roles/templates/roles.html
```diff
@@ -5,7 +5,7 @@
 
 {% block content %}
     <table class="i-table-widget multi-tbody" id="event-roles">
-        {{ render_roles(roles) }}
+        {{ render_roles(roles, collapsed=true, email_button=true) }}
```
ThiefMaster: no need to change anything here when having the correct default values for those two params

indico/modules/events/roles/templates/_roles.html
```diff
@@ -1,7 +1,7 @@
-{% macro render_roles(roles) -%}
+{% macro render_roles(roles, collapsed, email_button) -%}
```
ThiefMaster: no need for `collapsed` here, and `email_button` should default to `true`

indico/modules/categories/templates/management/roles.html
ThiefMaster: and `collapsed=true` isn't needed here

indico/web/forms/fields/principals.py
```diff
@@ -144,9 +149,18 @@ def event(self):
...
     @property
-    def roles(self):
+    def event_roles(self):
         return [serialize_role(role) for role in sorted(self.event.roles, key=attrgetter('code'))]
 
+    @property
+    def category_roles(self):
+        category_roles = list(
+            itertools.chain.from_iterable(
+                c.roles for c in self.event.category.chain_query.options(joinedload("roles"))
```
ThiefMaster: single quotes

indico/web/templates/forms/permissions_widget.html
```diff
@@ -24,6 +24,8 @@ <h3>Permissions</h3>
                         </a>
                     </li>
                 </ul>
+                <a class="i-button arrow js-dropdown" data-toggle="dropdown">{% trans %}Category role{% endtrans %}</a>
+                <ul class="i-dropdown entry-category-role-dropdown" data-items='{{ field.category_roles|tojson|forceescape }}'></ul>
```
ThiefMaster: this should only be rendered if there are actually some roles (showing an empty dropdown is weird)

indico/modules/categories/util.py
```diff
@@ -195,3 +195,22 @@ def get_image_data(image_type, category):
...
+
+def serialize_category_role(role, legacy=True):
+    """Serialize role to JSON-like object"""
+    if legacy:
+        return {
+            'id': role.id,
+            'name': role.name,
+            'code': role.code,
+            'color': role.color,
+            'identifier': 'Role:{}'.format(role.id),
```
ThiefMaster: `CategoryRole:...` (also `EventRole:...` for event roles)

indico/util/user.py
```diff
@@ -88,6 +90,12 @@ def principal_from_fossil(fossil, allow_pending=False, allow_groups=True, allow_
         if group.group is None and not allow_missing_groups:
             raise ValueError('Multipass group does not exist: {}:{}'.format(provider, id_))
         return group
+    elif allow_category_roles and event.category.roles:
```
ThiefMaster: Instead of the `allow_*` flag  better add a `category` kwarg (and default to `event.category` when an event but no category is present).

indico/util/user.py
```diff
@@ -88,6 +90,12 @@ def principal_from_fossil(fossil, allow_pending=False, allow_groups=True, allow_
...
+        role = CategoryRole.get(id_)
```
ThiefMaster: You need to handle roles from parent categories here as well (similar to what you do when displaying the list of roles to select from)

indico/util/user.py
```diff
@@ -88,6 +90,12 @@ def principal_from_fossil(fossil, allow_pending=False, allow_groups=True, allow_
...
+        role_name = fossil.get('name')
+        if role is None:
+            raise ValueError('Category role does not belong to provided event: {}:{}'.format(role_name, id_))
```
ThiefMaster: `This category role is not available here: ...`

indico/web/forms/fields/principals.py
```diff
@@ -63,6 +67,7 @@ def __init__(self, *args, **kwargs):
         self.allow_emails = kwargs.pop('allow_emails', False)
         self.groups = kwargs.pop('groups', False)
         self.allow_networks = kwargs.pop('allow_networks', False)
+        self.allow_category_roles = kwargs.pop('allow_category_roles', True)
```
ThiefMaster: If you don't need it in the end (due to using PermissionsField in category protection), remove it from here again. 

Otherwise document the new kwarg in the docstring above.

indico/modules/categories/forms.py
```diff
@@ -76,12 +79,8 @@ class CategoryLogoForm(IndicoForm):
 
 class CategoryProtectionForm(IndicoForm):
     _event_creation_fields = ('event_creation_restricted', 'event_creators', 'event_creation_notification_emails')
```
ThiefMaster: `event_creators` isn't needed in here anymore

indico/util/user.py
```diff
@@ -88,6 +97,17 @@ def principal_from_fossil(fossil, allow_pending=False, allow_groups=True, allow_
         if group.group is None and not allow_missing_groups:
             raise ValueError('Multipass group does not exist: {}:{}'.format(provider, id_))
         return group
+    elif category:
+        category_roles = list(
+            itertools.chain.from_iterable(
+                c.roles for c in category.chain_query.options(joinedload("roles"))
```
ThiefMaster: we usually use single quotes except for english strings that contain `'` in the string itself (or any other cases where using single quotes would require more escaping)

indico/util/user.py
```diff
@@ -170,5 +191,16 @@ def principal_from_identifier(identifier, allow_groups=False, allow_external_use
...
+    elif type_ == 'CategoryRole':
+        if not allow_category_roles:
+            raise ValueError('Category roles are not allowed')
+        try:
+            category_role_id = int(data)
+        except ValueError:
+            raise ValueError('Invalid data')
+        category_role = CategoryRole.get(category_role_id)
+        if category_role is None:
+            raise ValueError('Invalid category role: {}'.format(category_role_id))
```
ThiefMaster: We also need to check whether the role is available here (based on the category of the event from `event_id`).

indico/web/client/js/jquery/widgets/jinja/permissions_widget.js
```diff
@@ -271,18 +271,25 @@ import Palette from 'indico/utils/palette';
       ];
       this.$permissionsWidgetList.append(this._renderItem(anonymous));
 
-      const managersTitle =
-        this.options.objectType === 'event'
-          ? $T.gettext('Category Managers')
-          : $T.gettext('Event Managers');
-      const categoryManagers = [
+      var managersTitle;
```
ThiefMaster: `let managersTitle;` (`var` doesn't belong in any new JS code nowadays)

indico/web/client/js/jquery/widgets/jinja/permissions_widget.js
```diff
@@ -271,18 +271,25 @@ import Palette from 'indico/utils/palette';
...
+      if (this.options.objectType === 'category') {
+        managersTitle = $T.gettext('Category Managers');
+      } else {
+        managersTitle = $T.gettext('Parent Category Managers');
+      }
```
ThiefMaster: That logic here isn't correct:

- event -> category managers
- category -> parent category managers
- anything else -> event managers

indico/web/client/js/jquery/widgets/jinja/permissions_widget.js
```diff
@@ -271,18 +271,25 @@ import Palette from 'indico/utils/palette';
...
         {_type: 'DefaultEntry', name: managersTitle},
         [FULL_ACCESS_PERMISSIONS],
       ];
-      this.$permissionsWidgetList.prepend(this._renderItem(categoryManagers));
+      this.$permissionsWidgetList.prepend(this._renderItem(managers));
       this.$permissionsWidgetList.find('.anonymous').toggle(!this.isEventProtected);
 
-      this._renderDropdown(this.$roleDropdown);
+      if (this.$roleDropdown.length) {
+        this._renderDropdown(this.$roleDropdown);
```
ThiefMaster: let's rename the event role parts to `$eventRoleDropdown`

indico/web/client/js/react/components/principals/util.js
```diff
@@ -126,7 +129,7 @@ export class PrincipalType {
       [PrincipalType.multipassGroup]: Translate.string(
         'This group does not exist anymore. Please choose a different one.'
       ),
-      // event roles are hard-deleted
+      // event roles and category roles are hard-deleted
```
ThiefMaster: absolutely nothing wrong with writing it like this, but personally i'd use the more concise *"event/category roles ..."* here (and in the comment below)

indico/web/client/js/react/components/principals/util.js
```diff
@@ -144,6 +147,7 @@ export class PrincipalType {
       [PrincipalType.localGroup]: 0,
       [PrincipalType.multipassGroup]: 0,
       [PrincipalType.eventRole]: 1,
+      [PrincipalType.categoryRole]: 2,
```
ThiefMaster: I wouldn't group it with users but event roles, so `1`. Another option would be keeping this as `2` and changing the user sort order to `3`.

indico/web/forms/fields/principals.py
```diff
@@ -7,13 +7,16 @@
...
 
+from sqlalchemy.orm import joinedload
 from wtforms import HiddenField
 
 from indico.core.db.sqlalchemy.principals import PrincipalType, serialize_email_principal
 from indico.core.permissions import get_permissions_info
+from indico.modules.categories.models.categories import Category
 from indico.modules.events import Event
 from indico.modules.events.contributions.models.contributions import Contribution
 from indico.modules.events.roles.util import serialize_role
```
ThiefMaster: I'd move this one inside the function as well... just a matter of time before it causes import issues as well 🙈 

And maybe rename it to `serialize_event_role`...

indico/web/client/js/react/components/principals/util.js
ParthS007: Going with the second option

indico/modules/categories/blueprint.py
```diff
@@ -46,6 +49,15 @@ def _redirect_event_creation(category_id, event_type):
 _bp.add_url_rule('/manage/protection', 'manage_protection', RHManageCategoryProtection, methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/settings', 'manage_settings', RHManageCategorySettings, methods=('POST', 'GET'))
 
+# Role management
+_bp.add_url_rule('/manage/roles', 'manage_roles', RHCategoryRoles, methods=('POST', 'GET'))
+_bp.add_url_rule('/manage/roles/create', 'add_role', RHAddCategoryRole, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/roles/<int:role_id>/edit', 'edit_role', RHEditCategoryRole, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/roles/<int:role_id>', 'delete_role', RHDeleteCategoryRole, methods=('DELETE',))
```
mic4ael: Why not having all the `HTTP` methods handled by the same `RH`?

indico/modules/categories/client/js/management.js
```diff
@@ -435,4 +435,55 @@
...
+        .each(function() {
+          $(this).css('max-height', this.scrollHeight + 'px');
+        });
+    });
+  }
+
+  function setupRolesButtons() {
+    $('#event-roles').on('click', '.js-add-members', function(evt) {
+      var $this = $(this);
+      evt.stopPropagation();
```
mic4ael: It is usually the first thing to appear in the callback (however there is nothing wrong with putting it after `var $this`)

indico/modules/categories/controllers/management.py
```diff
@@ -16,29 +17,51 @@
...
+from indico.web.forms.fields.principals import serialize_principal
 from indico.web.util import jsonify_data, jsonify_form, jsonify_template, url_for_index
 
 
 CATEGORY_ICON_DIMENSIONS = (16, 16)
 MAX_CATEGORY_LOGO_DIMENSIONS = (200, 200)
 
 
+def _get_roles(category):
+    return (CategoryRole.query.with_parent(category)
```
mic4ael: The whole function can fit in a single line

indico/modules/categories/controllers/management.py
```diff
@@ -185,27 +208,25 @@ def _set_image(self, data, metadata):
 
 class RHManageCategoryProtection(RHManageCategoryBase):
     def _process(self):
-        form = CategoryProtectionForm(obj=self._get_defaults(), category=self.category)
+        category = self.category
```
mic4ael: Why?

indico/modules/categories/controllers/management.py
```diff
@@ -402,3 +423,94 @@ def _process(self):
...
+        }
+    }
+
+    def _process_args(self):
+        RHManageCategoryBase._process_args(self)
+        self.role = CategoryRole.get_one(request.view_args['role_id'])
+
+
+class RHEditCategoryRole(RHManageCategoryRole):
+    """Edit an category role"""
```
mic4ael: a category

indico/modules/categories/controllers/management.py
```diff
@@ -402,3 +423,94 @@ def _process(self):
...
+        if form.validate_on_submit():
+            form.populate_obj(self.role)
+            db.session.flush()
+            logger.info('Category role %r updated by %r', self.role, session.user)
+            return jsonify_data(html=_render_role(self.role))
+        return jsonify_form(form)
+
+
+class RHDeleteCategoryRole(RHManageCategoryRole):
+    """Delete an category role"""
```
mic4ael: a category

indico/modules/categories/controllers/management.py
```diff
@@ -402,3 +423,94 @@ def _process(self):
...
+
+    def _process(self):
+        db.session.delete(self.role)
+        logger.info('Category role %r deleted by %r', self.role, session.user)
+        return jsonify_data(html=_render_roles(self.category))
+
+
+class RHRemoveCategoryRoleMember(RHManageCategoryRole):
+    """Remove a user from an category role"""
```
mic4ael: a category

indico/modules/categories/controllers/management.py
```diff
@@ -402,3 +423,94 @@ def _process(self):
...
+
+    def _process(self):
+        if self.user in self.role.members:
+            self.role.members.remove(self.user)
+            logger.info('User %r removed from role %r by %r', self.user, self.role, session.user)
+        return jsonify_data(html=_render_role(self.role, collapsed=False))
+
+
+class RHAddCategoryRoleMembers(RHManageCategoryRole):
+    """Add users to an category role"""
```
mic4ael: a category

indico/modules/categories/forms.py
```diff
@@ -92,13 +91,11 @@ class CategoryProtectionForm(IndicoForm):
...
                                                            'to a list of specific persons'))
-    event_creators = PrincipalListField(_('Event creators'), groups=True, allow_external=True,
-                                        description=_('Users allowed to create events in this category'))
 
     def __init__(self, *args, **kwargs):
-        self.protected_object = category = kwargs.pop('category')
+        self.protected_object = self.category = kwargs.pop('category')
         super(CategoryProtectionForm, self).__init__(*args, **kwargs)
-        self._init_visibility(category)
+        self._init_visibility(self.category)
```
mic4ael: Why do you have to create `self.category`? Is it used anywhere?

indico/modules/categories/forms.py
```diff
@@ -171,3 +168,25 @@ def validate_entries(self, field):
...
+        self.role = kwargs.get('obj')
+        self.category = kwargs.pop('category')
+        super(RoleForm, self).__init__(*args, **kwargs)
+
+    def validate_code(self, field):
+        query = CategoryRole.query.with_parent(self.category).filter_by(code=field.data)
+        if self.role is not None:
+            query = query.filter(CategoryRole.id != self.role.id)
+        if query.has_rows():
+            raise ValueError(_('A role with this abbreviation already exists.'))
```
mic4ael: `A role with this code`?

indico/modules/categories/blueprint.py
ThiefMaster: i asked him to just copy from the event roles RHs ;) if we move this to react at some point i'd use that moment to switch to the usual REST-style rh structure

indico/modules/categories/forms.py
ThiefMaster: It's accessed by one of the fields via `self.get_form().catgory`

indico/modules/categories/forms.py
ThiefMaster: sounds better, i'd update this for event roles as well

ParthS007: We are using `abbreviation`  in event role also and in both places, in UI it is named as `abbreviation`  not `code` .

mic4ael: We use `code` in many other places for similar purposes.

mic4ael: An apart from that we store it as `code` :P

mic4ael: And we could just change the label for the `EventRole` field

indico/modules/categories/forms.py
mic4ael: Ah ok. I didn't notice that.

indico/modules/categories/forms.py
ParthS007: okay, I will make the change in both places. :+1:

indico/modules/categories/client/js/management.js
ParthS007: I didn't get this one.

indico/web/client/js/react/components/principals/util.js
```diff
@@ -126,7 +129,7 @@ export class PrincipalType {
       [PrincipalType.multipassGroup]: Translate.string(
         'This group does not exist anymore. Please choose a different one.'
       ),
-      // event roles are hard-deleted
+      // event roles/category roles are hard-deleted
```
ThiefMaster: "event/category roles" (or put a space around the `/` if you want to repeat the "roles") 🙈

indico/modules/categories/controllers/management.py
ThiefMaster: Good point indeed, not sure why we did it like this for event protection.... maybe the longer form with `self` required some wrapping below that made it look worse? But now it should be fine so maybe update it here and in `RHEventProtection`

indico/util/user.py
```diff
@@ -170,5 +192,24 @@ def principal_from_identifier(identifier, allow_groups=False, allow_external_use
...
+            raise ValueError('Invalid event id: {}'.format(event_id))
+        if not allow_category_roles:
+            raise ValueError('Category roles are not allowed')
+        try:
+            category_role_id = int(data)
+        except ValueError:
+            raise ValueError('Invalid data')
+        available_category_roles = list(
+            itertools.chain.from_iterable(
+                c.roles for c in Event.get(event_id).category.chain_query.options(joinedload('roles'))
```
ThiefMaster: you can use `event` here; no need to `Event.get()` again

indico/web/forms/fields/principals.py
ThiefMaster: I don't think this is resolved - either remove the change to this field (if it's not needed), otherwise add it to the docstring...

indico/modules/categories/forms.py
```diff
@@ -171,3 +168,25 @@ def validate_entries(self, field):
                 raise ValidationError(_('Invalid category: {}').format(entry['id']))
             if entry['type'] == 'event' and not Event.get(entry['id'], is_deleted=False):
                 raise ValidationError(_('Invalid event: {}').format(entry['id']))
+
+
+class RoleForm(IndicoForm):
+    name = StringField(_('Name'), [DataRequired()],
+                       description=_('The full name of the role'))
+    code = StringField(_('Abbreviation'), [DataRequired(), Length(max=3)], filters=[lambda x: x.upper() if x else ''],
```
ThiefMaster: Abbreviation -> Code?

indico/web/forms/fields/principals.py
```diff
@@ -76,7 +81,8 @@ def __init__(self, *args, **kwargs):
     def _convert_principal(self, principal):
         return principal_from_fossil(principal, allow_pending=self.allow_external,
                                      allow_emails=self.allow_emails, allow_networks=self.allow_networks,
-                                     existing_data=self.object_data, event=self._event)
+                                     existing_data=self.object_data, event=self._event,
+                                     allow_category_roles=self.allow_category_roles)
```
ThiefMaster: this change can go away as well (it breaks the field in its current state)

ParthS007: oh I missed it

pferreir:
Also closes #3346

indico/util/user.py
```diff
@@ -170,5 +192,24 @@ def principal_from_identifier(identifier, allow_groups=False, allow_external_use
         if event_role is None or event_role.event_id != event_id:
             raise ValueError('Invalid event role: {}'.format(event_role_id))
         return event_role
+    elif type_ == 'CategoryRole':
+        event = Event.get(event_id)
+        if event is None:
+            raise ValueError('Invalid event id: {}'.format(event_id))
+        if not allow_category_roles:
+            raise ValueError('Category roles are not allowed')
```
ThiefMaster: I'd move this check up a bit (ie right before `event = Event.get(event_id)`) since getting the event is pointless if category roles are disabled. No harm in having it down here though, since the only case where you'd fail that condition is when sending nonsense data...

indico/web/client/js/jquery/widgets/jinja/permissions_widget.js
```diff
@@ -64,7 +64,7 @@ import Palette from 'indico/utils/palette';
       return [$code, $text];
     },
     _renderOtherLabel(principal, type) {
-      let iconClass, extraText;
+      let iconClass, extraText, $code;
```
ThiefMaster: I don't think any of the changes in `_renderOtherLabel` are needed - you never call this for a `CategoryRole` type since `_renderLabel` calls `_renderRoleLabel` for `EventRole` and `CategoryRole`

indico/web/forms/fields/principals.py
```diff
@@ -30,16 +32,19 @@
...
         return serialize_ip_network_group(principal)
     elif principal.principal_type == PrincipalType.user:
         return serialize_user(principal)
     elif principal.principal_type == PrincipalType.event_role:
-        return serialize_role(principal)
+        return serialize_event_role(principal)
     elif principal.is_group:
         return serialize_group(principal)
+    elif principal.principal_type == PrincipalType.category_role:
+        return serialize_category_role(principal)
```
ThiefMaster: I'd move this up two lines to have category roles right after event roles (it's a bit strange to have groups in between).

indico/web/client/js/jquery/widgets/jinja/permissions_widget.js
ParthS007: Added two functions namely : `_renderEventRoleLabel` and `_renderCategoryRoleLabel`

indico/modules/events/abstracts/fields.py
```diff
@@ -209,7 +211,16 @@ def permissions_info(self):
 
     @property
     def event_roles(self):
-        return [serialize_role(role, legacy=False) for role in sorted(self.event.roles, key=attrgetter('code'))]
+        return [serialize_event_role(role, legacy=False) for role in sorted(self.event.roles, key=attrgetter('code'))]
+
+    @property
+    def category_roles(self):
+        category_roles = list(
```
ThiefMaster: I found a much nicer way to get all available category roles for a category:

    CategoryRole.query.join(cat.chain_query.subquery()).all()

This would probably fit well in a `get_for_category(cat)` staticmethod of the `CategoryRole` model.

indico/web/forms/fields/principals.py
```diff
@@ -144,8 +151,24 @@ def event(self):
...
+        return self.event.category
+
+    @property
+    def event_roles(self):
+        return [serialize_event_role(role) for role in sorted(self.event.roles, key=attrgetter('code'))]
+
+    @property
+    def category_roles(self):
+        from indico.modules.categories.util import serialize_category_role
+        category_roles = list(
```
ThiefMaster: see my other comment about the nicer way of getting the role (you could even move the sorting in the query)

indico/util/user.py
```diff
@@ -88,6 +97,17 @@ def principal_from_fossil(fossil, allow_pending=False, allow_groups=True, allow_
         if group.group is None and not allow_missing_groups:
             raise ValueError('Multipass group does not exist: {}:{}'.format(provider, id_))
         return group
+    elif category:
+        category_roles = list(
+            itertools.chain.from_iterable(
+                c.roles for c in category.chain_query.options(joinedload('roles'))
+            )
+        )
+        role = CategoryRole.get(id_)
```
ThiefMaster: This can also be simplified like this:

    role = CategoryRole.query.filter_by(id=id_).join(category.chain_query.subquery()).first()

This could also go into another `CategoryRole` staticmethod (`get_for_category(cat, id)`).

indico/util/user.py
```diff
@@ -170,5 +192,25 @@ def principal_from_identifier(identifier, allow_groups=False, allow_external_use
...
+
+        event = Event.get(event_id)
+        if event is None:
+            raise ValueError('Invalid event id: {}'.format(event_id))
+        try:
+            category_role_id = int(data)
+        except ValueError:
+            raise ValueError('Invalid data')
+        available_category_roles = list(
+            itertools.chain.from_iterable(
```
ThiefMaster: see my above comment ;)

FWIW, for the case here the staticmethod would need another argument (false by default) that skips the subquery filter in case of `soft_fail`. Or you can just use `CategoryRole.get()` in case of `soft_fail`, and otherwise the staticmethod that checks the ancestry.

indico/modules/categories/models/roles.py
```diff
@@ -0,0 +1,135 @@
...
+        return 'color: #{0} !important; border-color: #{0} !important'.format(self.color)
+
+    @property
+    def style(self):
+        return {'color': '#' + self.color, 'borderColor': '#' + self.color}
+
+    @staticmethod
+    def get_category_roles(cat):
+        category_roles = CategoryRole.query.join(cat.chain_query.subquery()).all()
+        return sorted(category_roles, key=attrgetter('code'))
```
ThiefMaster: You can just add `.order_by(CategoryRole.code)` to the query:

    return CategoryRole.query....order_by(...).all()

indico/modules/categories/models/roles.py
```diff
@@ -0,0 +1,135 @@
...
+    @property
+    def css(self):
+        return 'color: #{0} !important; border-color: #{0} !important'.format(self.color)
+
+    @property
+    def style(self):
+        return {'color': '#' + self.color, 'borderColor': '#' + self.color}
+
+    @staticmethod
+    def get_category_roles(cat):
```
ThiefMaster: docstring please :)

```python
"""Get the category roles available for the specified category."""
```

indico/modules/categories/models/roles.py
```diff
@@ -0,0 +1,135 @@
...
+        category_roles = CategoryRole.query.join(cat.chain_query.subquery()).all()
+        return sorted(category_roles, key=attrgetter('code'))
+
+    @staticmethod
+    def get_category_role_by_id(cat, id):
```
ThiefMaster: docstring please :)

```python
"""Get a category role in the context of the specified category.

If the role is not defined in the category or one of its parents,
it is considered non-existing.
"""
```

PR #4276, opened 2020-02-04
Hide judgment button when submitter_needs_changes

### LINKED ISSUE #4274: Hide judge button from showing when the submitter has rejected the proposed changes

panagiotappl:
closes #4274

### DIFF
--- a/indico/modules/events/editing/client/js/selectors.js
+++ b/indico/modules/events/editing/client/js/selectors.js
@@ -199,3 +199,8 @@ export const getLastRevertableRevisionId = createSelector(
     return lastFinalRev.id;
   }
 );
+
+export const canJudgeLastRevision = createSelector(
+  getLastRevision,
+  lastRevision => lastRevision.finalState.name === FinalRevisionState.none
+);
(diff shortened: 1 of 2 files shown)


indico/modules/events/editing/client/js/selectors.js
```diff
@@ -199,3 +199,10 @@ export const getLastRevertableRevisionId = createSelector(
     return lastFinalRev.id;
   }
 );
+
+export const canJudgeLastRevision = createSelector(
+  getLastRevision,
+  lastRevision => {
```
ThiefMaster: I think this would fit in one line using the `=> foo` style instead of `=> { return foo; }`

indico/modules/events/editing/client/js/selectors.js
```diff
@@ -199,3 +199,10 @@ export const getLastRevertableRevisionId = createSelector(
...
+    return lastRevision.finalState.name !== FinalRevisionState.needs_submitter_changes;
```
ThiefMaster: Checking for `=== FinalRevisionState.none` seems like the safer choice to me - none of the other states allow judging anyway.

PR #4279, opened 2020-02-06
Allow padding of contribution ids in programme codes

### LINKED ISSUE #4220: Progamme codes: Add placeholder to define number of id digits
Allow defining how many digits should id placeholder have and add zeros to small numbers. For example if digit number is defined as 3 the code for contribution id 2 should be 002

Leats:
Closes #4220

### DIFF
--- a/indico/modules/events/management/program_codes.py
+++ b/indico/modules/events/management/program_codes.py
@@ -12,7 +12,7 @@ from collections import OrderedDict
 from indico.modules.events.management.settings import program_codes_settings
 from indico.util.date_time import format_datetime
 from indico.util.i18n import _
-from indico.util.placeholders import Placeholder, get_empty_placeholders, replace_placeholders
+from indico.util.placeholders import ParametrizedPlaceholder, Placeholder, get_empty_placeholders, replace_placeholders
 
 
 def generate_program_codes(event, object_type, objects):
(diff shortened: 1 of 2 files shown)


indico/modules/events/management/program_codes.py
```diff
@@ -49,13 +49,21 @@ def _make_date_placeholder(class_name, name, render_kwarg, date_format, descript
...
     @classmethod
-    def render(cls, contribution):
-        return unicode(contribution.friendly_id)
+    def render(cls, param, contribution):
+        if param is None:
+            return unicode(contribution.friendly_id)
+        try:
+            padding = max(1, min(int(param), 10))
+            return unicode(contribution.friendly_id).zfill(padding)
+        except TypeError:
```
ThiefMaster: i think it's ValueError here; and I'd just default to no padding in that case until we can validate while saving

mic4ael: Yup, `ValueError`

mic4ael:
@Leats you can easily squash these commits into one :)

indico/modules/events/management/program_codes.py
```diff
@@ -49,13 +49,21 @@ def _make_date_placeholder(class_name, name, render_kwarg, date_format, descript
...
 
     @classmethod
-    def render(cls, contribution):
-        return unicode(contribution.friendly_id)
+    def render(cls, param, contribution):
+        if param is None:
+            return unicode(contribution.friendly_id)
+        try:
+            padding = max(1, min(int(param), 10))
+            return unicode(contribution.friendly_id).zfill(padding)
```
ThiefMaster: I'd move this one into the `else` part of the try..except since errors in here would be exceptional and thus shouldn't be caught

ThiefMaster:
Very nice first PR! 🎉
And thanks for fixing that bug ;x

indico/modules/events/management/program_codes.py
```diff
@@ -49,13 +49,22 @@ def _make_date_placeholder(class_name, name, render_kwarg, date_format, descript
     return type(str(class_name), (DatePlaceholder,), dct)
 
 
-class ContributionIDPlaceholder(Placeholder):
+class ContributionIDPlaceholder(ParametrizedPlaceholder):
     name = 'id'
-    description = _('The ID of the contribution')
+    description = _('The ID of the contribution, optionally padded with zeros to a specified length '
```
ThiefMaster: > The ID of the contribution (optionally padded with zeros)

might be enough, since the param is documented in the param list. you could set `param_friendly_name = 'length'` as well to have something nicer than "param" in the help text there - but feel free to keep an example in there as well!

PR #4280, opened 2020-02-06
Log bulk event moves via category management area

### LINKED ISSUE #4241: Log bulk event moves
Right now we log only [moves of single events](https://github.com/indico/indico/blob/master/indico/modules/events/management/controllers/actions.py#L96-L97), but not when multiple events are moved from the category management area.

Leats:
Closes #4241.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,7 @@ Improvements
   category settings form (:issue:`4231`)
 - Added new :data:`LOCAL_GROUPS` setting that can be used to fully disable
   local groups (:issue:`4260`)
+- Log bulk event category changes in the event log (:issue:`4241`)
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Thanks, looks good! But for this one I'd add a changelog entry (CHANGES.rst) in the 2.3 improvements section:

```rst
- Log bulk event category changes in the event log (:issue:`4241`)
```

ThiefMaster:
I just noticed that `RHSplitCategory` could also use this kind of logging (since it moves events as well in the end).

And since that means every `Event.move()` call will need the logging, I'd simply move the logic out of the RHs into that method (not sure why we didn't put it there in the first place 😳).

PR #4283, opened 2020-02-07
Show warning when merging users and one of them is blocked

### LINKED ISSUE #3845: Show warning when merging users and one of them is blocked
There's no UI to block users yet (bonus: you could add an `indico user <block|unblock> <id>` CLI command), but `indico shell` can be used for it

```python
User.get(12345).is_blocked = True
db.session.commit()
```

Leats:
Closes #3845

### DIFF
--- a/indico/cli/user.py
+++ b/indico/cli/user.py
@@ -30,6 +30,15 @@ def cli():
 
 
 def _print_user_info(user):
+    flags = []
+    if user.is_admin:
+        flags.append('%{yellow}admin%{reset}')
+    if user.is_blocked:
+        flags.append('%{red}blocked%{reset}')
+    if user.is_deleted:
+        flags.append('%{red!}deleted%{reset}')
+    if user.is_pending:
+        flags.append('%{cyan}pending%{reset}')
     print()
     print('User info:')
     print("  ID: {}".format(user.id))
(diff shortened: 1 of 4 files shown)


indico/cli/user.py
```diff
@@ -166,3 +166,39 @@ def revoke_admin(user_id):
...
+
+@cli.command()
+@click.argument('user_id', type=int)
+def block(user_id):
+    """Blocks a given user"""
+    user = User.get(user_id)
+    if user is None:
+        print(cformat("%{red}This user does not exist"))
+        return
+    _print_user_info(user)
```
ThiefMaster: I'd print an extra line in that function if the user is blocked (and nothing if they aren't, since blocked users are somewhat uncommon)

ThiefMaster:
If you want to dig a bit into how custom [`click`](https://click.palletsprojects.com/en/7.x/) types work, you could see if you can add some type for the `user_id` arg that automatically gets the user by ID and aborts if the user doesn't exist (instead of having that code in every of the four(?) user functions)

Leats:
As @ThiefMaster suggested an extra line will be added when the user has flags:
> I'd print an extra line in that function if the user is blocked (and nothing if they aren't, since blocked users are somewhat uncommon)

PR #4286, opened 2020-02-10
Improve access checks in Editable-related RHs

### LINKED ISSUE #4284: Fix access checks on timeline view
Currently accessing it without being logged in results in many ugly errors since the page itself isn't protected but the APIs are.

The page itself should always require logging in, so you are redirected to the login, and also check if you are authorized to access that editable at all.

mic4ael:
closes #4284

### DIFF
--- a/indico/modules/events/editing/controllers/base.py
+++ b/indico/modules/events/editing/controllers/base.py
@@ -8,31 +8,31 @@
 from __future__ import unicode_literals
 
 from flask import request, session
-from werkzeug.exceptions import Forbidden
 
 from indico.modules.events.contributions.controllers.display import RHContributionDisplayBase
-from indico.modules.events.controllers.base import RHEventBase
+from indico.modules.events.controllers.base import RHDisplayEventBase
 from indico.modules.events.editing.models.editable import Editable, EditableType
 from indico.modules.events.management.controllers.base import RHManageEventBase
+from indico.web.rh import RequireUserMixin
 
 
-class RHEditingBase(RHEventBase):
+class RHEditingBase(RequireUserMixin, RHDisplayEventBase):
     """Base class for editing RHs that don't reference an editable."""
 
     EVENT_FEATURE = 'editing'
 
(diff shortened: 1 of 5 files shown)


indico/modules/events/editing/controllers/base.py
```diff
@@ -11,28 +11,30 @@
...
 
 
 class RHEditingManagementBase(RHManageEventBase):
+    """Base class for editing RHs that don't reference an editable."""
+
     EVENT_FEATURE = 'editing'
 
 
-class RHContributionEditableBase(RHContributionDisplayBase):
+class RHContributionEditableBase(RHContributionDisplayBase, RHProtected):
```
ThiefMaster: What about adding a `RequireUserMixin` and using that here and in `RHEditingBase`? That way we don't need to inherit from two RH base classes, and still avoid having the `session.user` check in many different places..

indico/modules/events/editing/controllers/base.py
```diff
@@ -8,31 +8,31 @@
...
 
 
 class RHEditingManagementBase(RHManageEventBase):
+    """Base class for editing RHs that don't reference an editable."""
+
     EVENT_FEATURE = 'editing'
 
 
-class RHContributionEditableBase(RHContributionDisplayBase):
+class RHContributionEditableBase(RHContributionDisplayBase, RequireUserMixin):
```
ThiefMaster: i'd put the mixin on the left of the base class

indico/modules/events/editing/controllers/frontend.py
```diff
@@ -25,6 +25,14 @@ def _process_args(self):
         if not self.editable:
             raise NotFound
 
+    def _check_access(self):
+        RHContributionEditableBase._check_access(self)
+
+        if self.event.can_manage(session.user) or self.event.can_manage(session.user, permission='paper_editing'):
```
ThiefMaster: the second check already includes full management permissions; no need to do two checks here

PR #4288, opened 2020-02-10
OpenGraph: show event title

pferreir:
Indico page previews currently only show "Indico" as the title.

e.g.

image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,7 @@ Bugfixes
   tickets are blocked (:issue:`4242`)
 - Fix event access key prompt not showing when accessing an attachment link
   (:issue:`4255`)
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -18,6 +18,7 @@ Bugfixes
   tickets are blocked (:issue:`4242`)
 - Fix event access key prompt not showing when accessing an attachment link
   (:issue:`4255`)
+- Include event title in OpenGraph metadata (:issue:`4288`)
```
ThiefMaster: bugfix or improvement?

pferreir: Thought of it too. Looks like something we overlooked to me, hence a bug.

PR #4296, opened 2020-02-18
Extract logic of registration form display to util

### LINKED ISSUE #4122: Registration Form: hide it if not scheduled and not open
Right now we always show all registration forms on the display page of an event. Those which are closed are shown as disabled.
I suggest we just hide any forms which are not available. The exception would be scheduled forms whose end date has already passed (as it's useful for the user to notice that registration has already closed).

Leats:
Logic which was used both for conferences and meeting/lectures now is placed in a util function.
The template for conferences was fixed to work with these changes.

Probably closes #4122. 
The requested behavior is implemented in the conference display. Since meetings/lectures do not show opening and closing dates of scheduling it was not changed here to avoid confusion.

### DIFF
--- a/indico/modules/events/registration/util.py
+++ b/indico/modules/events/registration/util.py
@@ -488,6 +488,27 @@ def get_event_regforms(event, user, with_registrations=False):
     return query.all()
 
 
+def get_event_regforms_registrations(event, user, include_scheduled=True):
+    """Get regforms and the associated registrations for an event+user.
+
+    :param event: the `Event` to get registration forms for
+    :param user: A `User`
+    :param include_scheduled: Whether to include scheduled
+                              but not open registration forms
+    :return: A tuple, which includes:
+            - All registration forms which are scheduled, open or registered.
+            - A dict mapping all registration forms to the user's registration if they have one.
+    """
+    all_regforms = get_event_regforms(event, user, with_registrations=True)
+    if include_scheduled:
+        displayed_regforms = [regform for regform, registration in all_regforms
+                              if regform.is_scheduled or registration]
+    else:
+        displayed_regforms = [regform for regform, registration in all_regforms
+                              if regform.is_open or registration]
+    return displayed_regforms, dict(all_regforms)
+
+
 def generate_ticket(registration):
     from indico.modules.designer.util import get_default_template_on_category
     from indico.modules.events.registration.controllers.management.tickets import DEFAULT_TICKET_PRINTING_SETTINGS
--- a/indico/modules/events/registration/util_test.py
+++ b/indico/modules/events/registration/util_test.py
@@ -7,12 +7,14 @@
 
 from __future__ import unicode_literals
 
+from datetime import datetime
 from io import BytesIO
 
 import pytest
 
 from indico.core.errors import UserValueError
-from indico.modules.events.registration.util import create_registration, import_registrations_from_csv
+from indico.modules.events.registration.util import (create_registration, get_event_regforms_registrations,
+                                                     import_registrations_from_csv)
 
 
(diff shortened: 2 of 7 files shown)


indico/modules/events/registration/util.py
```diff
@@ -488,6 +488,27 @@ def get_event_regforms(event, user, with_registrations=False):
     return query.all()
 
 
+def get_event_regforms_registrations(event, user, include_scheduled=True):
+    """Get a tuple, which includes:
```
ThiefMaster: First docstring line should be a concise single-sentence summary - if there's no good way to write one, wrap right after the `"""` and start in the next line (without extra indentation).

This could be a good first line here:


```python
"""Get regforms and the associated registrations for an event+user.

...
"""
```

And then explain details later (similar to how it's done in git commits with the empty line between).

indico/modules/events/registration/util.py
```diff
@@ -488,6 +488,27 @@ def get_event_regforms(event, user, with_registrations=False):
...
+       All registration forms which are scheduled, open or registered.
+       All user registrations corresponding to them.
+
+    :param event: the `Event` to get registration forms for
+    :param user: A `User`
+    :param include_scheduled: Whether to include scheduled
+                              but not open registration
+                              forms
```
ThiefMaster: let's move this into the previous line? 🙈

indico/modules/events/registration/util.py
```diff
@@ -488,6 +488,27 @@ def get_event_regforms(event, user, with_registrations=False):
     return query.all()
 
 
+def get_event_regforms_registrations(event, user, include_scheduled=True):
+    """Get a tuple, which includes:
+       All registration forms which are scheduled, open or registered.
+       All user registrations corresponding to them.
```
ThiefMaster: > A dict mapping all registration forms to the user's registration if they have one

ThiefMaster:
A unit test for `get_event_regforms_registrations` could be nice - with some `@pytest.mark.parametrize` you could have a list of start/end date and `include_scheduled` combinations (including some `None`) and a boolean indicating whether the regform should be returned or not. That way the test case itself can be extremely simply and still cover everything.

And even better, it'll serve as additional documentation on what the behavior should be.

Tip: We have a `freeze_time` fixture that lets you set the current time during the test to something fixed

indico/modules/events/registration/util_test.py
```diff
@@ -109,3 +111,53 @@ def test_import_error(dummy_regform):
...
+                                            include_scheduled, expected_regform_flag):
+    freeze_time(datetime(2019, 12, 13, 8, 0, 0))
+    if start_dt:
+        dummy_regform.start_dt = dummy_event.tzinfo.localize(start_dt)
+    if end_dt:
+        dummy_regform.end_dt = dummy_event.tzinfo.localize(end_dt)
+
+    regforms, registrations = get_event_regforms_registrations(dummy_event, dummy_user, include_scheduled)
+
+    assert ('Registration Form' in str(regforms)) == expected_regform_flag
```
ThiefMaster: `str(regforms)` gives you a repr of the list.. while it works, it feels a bit dirty.. Better check for the object directly:

    assert (dummy_regform in regforms) == expected_regform_flag

indico/modules/events/registration/util_test.py
```diff
@@ -109,3 +111,53 @@ def test_import_error(dummy_regform):
...
+    assert 'None' in str(registrations.values())
```
ThiefMaster: same here, `assert None in registrations.values()`

or if you know there's exactly one entry, just check `registrations.values() == [None]`

PR #4301, opened 2020-02-19
Add import of role members

### LINKED ISSUE #3346: Abstracts: UX improvements in role management

ParthS007:
related to #3346

### DIFF
--- a/indico/modules/categories/blueprint.py
+++ b/indico/modules/categories/blueprint.py
@@ -19,9 +19,9 @@ from indico.modules.categories.controllers.display import (RHCategoryCalendarVie
                                                            RHShowFutureEventsInCategory, RHShowPastEventsInCategory,
                                                            RHSubcatInfo, RHXMLExportCategoryInfo)
 from indico.modules.categories.controllers.management import (RHAddCategoryRole, RHAddCategoryRoleMembers,
-                                                              RHCategoryRoles, RHCreateCategory, RHDeleteCategory,
-                                                              RHDeleteCategoryRole, RHDeleteEvents,
-                                                              RHDeleteSubcategories, RHEditCategoryRole,
+                                                              RHCategoryRoleMembersImportCSV, RHCategoryRoles,
+                                                              RHCreateCategory, RHDeleteCategory, RHDeleteCategoryRole,
+                                                              RHDeleteEvents, RHDeleteSubcategories, RHEditCategoryRole,
                                                               RHManageCategoryContent, RHManageCategoryIcon,
                                                               RHManageCategoryLogo, RHManageCategoryProtection,
                                                               RHManageCategorySettings, RHMoveCategory, RHMoveEvents,
@@ -57,6 +57,8 @@ _bp.add_url_rule('/manage/roles/<int:role_id>', 'delete_role', RHDeleteCategoryR
 _bp.add_url_rule('/manage/roles/<int:role_id>/members', 'add_members', RHAddCategoryRoleMembers, methods=('POST',))
 _bp.add_url_rule('/manage/roles/<int:role_id>/members/<int:user_id>', 'remove_member', RHRemoveCategoryRoleMember,
                  methods=('DELETE',))
+_bp.add_url_rule('/manage/roles/<int:role_id>/members/import', 'add_members_import_csv',
+                 RHCategoryRoleMembersImportCSV, methods=('GET', 'POST'))
 
 # Event management
 _bp.add_url_rule('/manage/events/delete', 'delete_events', RHDeleteEvents, methods=('GET', 'POST'))
--- a/indico/modules/categories/blueprint.py
+++ b/indico/modules/categories/blueprint.py
@@ -54,8 +54,8 @@ _bp.add_url_rule('/manage/roles', 'manage_roles', RHCategoryRoles, methods=('POS
 _bp.add_url_rule('/manage/roles/create', 'add_role', RHAddCategoryRole, methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/roles/<int:role_id>/edit', 'edit_role', RHEditCategoryRole, methods=('GET', 'POST'))
 _bp.add_url_rule('/manage/roles/<int:role_id>', 'delete_role', RHDeleteCategoryRole, methods=('DELETE',))
-_bp.add_url_rule('/manage/roles/<int:role_id>/members', 'add_members', RHAddCategoryRoleMembers, methods=('POST',))
-_bp.add_url_rule('/manage/roles/<int:role_id>/members/<int:user_id>', 'remove_member', RHRemoveCategoryRoleMember,
+_bp.add_url_rule('/manage/roles/<int:role_id>/members', 'add_role_members', RHAddCategoryRoleMembers, methods=('POST',))
+_bp.add_url_rule('/manage/roles/<int:role_id>/members/<int:user_id>', 'remove_role_member', RHRemoveCategoryRoleMember,
                  methods=('DELETE',))
 _bp.add_url_rule('/manage/roles/<int:role_id>/members/import', 'add_members_import_csv',
                  RHCategoryRoleMembersImportCSV, methods=('GET', 'POST'))
--- a/indico/modules/categories/controllers/management.py
+++ b/indico/modules/categories/controllers/management.py
@@ -33,6 +33,7 @@ from indico.modules.rb.models.reservations import Reservation, ReservationLink
 from indico.modules.users import User
 from indico.util.fs import secure_filename
 from indico.util.i18n import _, ngettext
+from indico.util.roles import ImportRoleMembersMixin
 from indico.util.string import crc32
 from indico.util.user import principal_from_fossil
 from indico.web.flask.templating import get_template_module
@@ -512,3 +513,9 @@ class RHAddCategoryRoleMembers(RHManageCategoryRole):
                 self.role.members.add(user)
                 logger.info('User %r added to role %r by %r', user, self.role, session.user)
         return jsonify_data(html=_render_role(self.role, collapsed=False))
+
+
+class RHCategoryRoleMembersImportCSV(ImportRoleMembersMixin, RHManageCategoryRole):
+    """Add users to a category role from CSV"""
+
+    logger = logger
--- a/indico/modules/events/roles/blueprint.py
+++ b/indico/modules/events/roles/blueprint.py
@@ -8,7 +8,8 @@
 from __future__ import unicode_literals
 
 from indico.modules.events.roles.controllers import (RHAddEventRole, RHAddEventRoleMembers, RHDeleteEventRole,
-                                                     RHEditEventRole, RHEventRoles, RHRemoveEventRoleMember)
+                                                     RHEditEventRole, RHEventRoleMembersImportCSV, RHEventRoles,
+                                                     RHRemoveEventRoleMember)
 from indico.web.flask.wrappers import IndicoBlueprint
 
 
@@ -23,3 +24,6 @@ _bp.add_url_rule('/<int:role_id>', 'delete_role', RHDeleteEventRole, methods=('D
 
 _bp.add_url_rule('/<int:role_id>/members', 'add_members', RHAddEventRoleMembers, methods=('POST',))
 _bp.add_url_rule('/<int:role_id>/members/<int:user_id>', 'remove_member', RHRemoveEventRoleMember, methods=('DELETE',))
+
+_bp.add_url_rule('/<int:role_id>/members/import', 'add_members_import_csv', RHEventRoleMembersImportCSV,
+                 methods=('GET', 'POST'))
--- a/indico/modules/events/roles/blueprint.py
+++ b/indico/modules/events/roles/blueprint.py
@@ -22,8 +22,9 @@ _bp.add_url_rule('/create', 'add_role', RHAddEventRole, methods=('GET', 'POST'))
 _bp.add_url_rule('/<int:role_id>/edit', 'edit_role', RHEditEventRole, methods=('GET', 'POST'))
 _bp.add_url_rule('/<int:role_id>', 'delete_role', RHDeleteEventRole, methods=('DELETE',))
 
-_bp.add_url_rule('/<int:role_id>/members', 'add_members', RHAddEventRoleMembers, methods=('POST',))
-_bp.add_url_rule('/<int:role_id>/members/<int:user_id>', 'remove_member', RHRemoveEventRoleMember, methods=('DELETE',))
+_bp.add_url_rule('/<int:role_id>/members', 'add_role_members', RHAddEventRoleMembers, methods=('POST',))
+_bp.add_url_rule('/<int:role_id>/members/<int:user_id>', 'remove_role_member', RHRemoveEventRoleMember,
+                 methods=('DELETE',))
 
 _bp.add_url_rule('/<int:role_id>/members/import', 'add_members_import_csv', RHEventRoleMembersImportCSV,
                  methods=('GET', 'POST'))
--- a/indico/modules/events/roles/controllers.py
+++ b/indico/modules/events/roles/controllers.py
@@ -21,6 +21,7 @@ from indico.modules.events.roles.forms import EventRoleForm
 from indico.modules.events.roles.util import serialize_event_role
 from indico.modules.events.roles.views import WPEventRoles
 from indico.modules.users import User
+from indico.util.roles import ImportRoleMembersMixin
 from indico.util.user import principal_from_fossil
 from indico.web.flask.templating import get_template_module
 from indico.web.forms.colors import get_role_colors
@@ -145,3 +146,9 @@ class RHAddEventRoleMembers(RHManageEventRole):
                                data={'Name': user.full_name,
                                      'Email': user.email})
         return jsonify_data(html=_render_role(self.role, collapsed=False))
+
+
+class RHEventRoleMembersImportCSV(ImportRoleMembersMixin, RHManageEventRole):
+    """Add users to an event role from CSV"""
+
+    logger = logger
--- a/indico/modules/events/roles/forms.py
+++ b/indico/modules/events/roles/forms.py
@@ -7,14 +7,15 @@
 
 from __future__ import unicode_literals
 
-from wtforms.fields.core import StringField
-from wtforms.validators import DataRequired, Length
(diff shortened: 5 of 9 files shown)


indico/modules/events/roles/util.py
```diff
@@ -25,3 +32,31 @@ def serialize_event_role(role, legacy=True):
...
+
+    for num_row, row in enumerate(reader, 1):
+        try:
+            email = [to_unicode(value).strip() for value in row]
+        except ValueError:
+            raise UserValueError(_('Row {}: malformed CSV data').format(num_row))
+        try:
+            email = email[0].lower()
+        except IndexError:
+            continue
```
ThiefMaster:
```python
if not email:
    continue
```

indico/modules/events/roles/util.py
```diff
@@ -25,3 +32,31 @@ def serialize_event_role(role, legacy=True):
...
+
+
+def import_members_from_csv(f):
+    """Import members from a CSV file into an event role."""
+    reader = csv.reader(f.read().splitlines())
+    members_data = set()
+
+    for num_row, row in enumerate(reader, 1):
+        try:
+            email = [to_unicode(value).strip() for value in row]
```
ThiefMaster: you can already `.lower()` it here

ThiefMaster: You expect only one row, so simply use `email = to_unicode(row[0]).strip().lower()` directly (and possibly abort if `len(row) != 1`)

indico/modules/events/roles/util.py
```diff
@@ -25,3 +32,31 @@ def serialize_event_role(role, legacy=True):
             'name': role.name,
             'identifier': 'EventRole:{}'.format(role.id),
         }
+
+
+def import_members_from_csv(f):
+    """Import members from a CSV file into an event role."""
+    reader = csv.reader(f.read().splitlines())
+    members_data = set()
```
ThiefMaster: `emails` might be a better name?

indico/modules/events/roles/util.py
```diff
@@ -25,3 +32,31 @@ def serialize_event_role(role, legacy=True):
...
+            continue
+
+        if email and not validate_email(email):
+            raise UserValueError(_("Row {row}: invalid email address: {email}").format(row=num_row, email=email))
+        if email in members_data:
+            raise UserValueError(_('Row {}: email address is not unique').format(num_row))
+        members_data.add(email)
+
+    registered_users = set(User.query.filter(~User.is_deleted, User.all_emails.in_(list(members_data))))
+    registered_users_email = set(user.email for user in list(registered_users))
```
ThiefMaster: you can iterate a set directly, without converting it to a list first, and python 2.7+ has set comprehensions so no need to use `set(...)` - you can use `{...}` instead

indico/modules/events/roles/util.py
```diff
@@ -25,3 +32,31 @@ def serialize_event_role(role, legacy=True):
...
+        except IndexError:
+            continue
+
+        if email and not validate_email(email):
+            raise UserValueError(_("Row {row}: invalid email address: {email}").format(row=num_row, email=email))
+        if email in members_data:
+            raise UserValueError(_('Row {}: email address is not unique').format(num_row))
+        members_data.add(email)
+
+    registered_users = set(User.query.filter(~User.is_deleted, User.all_emails.in_(list(members_data))))
```
ThiefMaster: is the list() around `members_data` needed? I'd assume SQLAlchemy is fine with you passing a set there

indico/modules/categories/controllers/management.py
```diff
@@ -512,3 +514,40 @@ def _process(self):
...
+                not_registered_email,
+            ) = import_members_from_csv(form.source_file.data)
+            new_members = registered_users - self.role.members
+            not_registered = csv_email - registered_users_email
+            if bool(new_members):
+                if form.remove_existing.data:
+                    self.role.members = registered_users
+                    flash(ngettext("{} members has been imported.",
+                                   "{} members have been imported.",
+                                   registered_users.__len__()).format(registered_users.__len__()), 'success')
```
ThiefMaster: use `len(x)` instead of `x.__len__()`

indico/modules/categories/controllers/management.py
```diff
@@ -512,3 +514,40 @@ def _process(self):
...
+                    flash(ngettext("{} new member has been imported.",
+                                   "{} new members have been imported.",
+                                   new_members.__len__()).format(new_members.__len__()), 'success')
+            else:
+                flash(ngettext("{} member is already present.",
+                               "{} members are already present.",
+                               registered_users.__len__()).format(registered_users.__len__()), 'success')
+
+            if bool(not_registered):
+                flash(_("{} not registered email.".format(', '.join(not_registered))), 'warning')
```
ThiefMaster: you need to do the `.format()` outside the `_()` call - the string in there must be fully static.

also, the text is strange.. what about something like "There are no users with these email addresses: ..." (with proper pluralization of course)

indico/modules/categories/controllers/management.py
```diff
@@ -512,3 +514,40 @@ def _process(self):
...
+                    self.role.members |= new_members
+                    flash(ngettext("{} new member has been imported.",
+                                   "{} new members have been imported.",
+                                   new_members.__len__()).format(new_members.__len__()), 'success')
+            else:
+                flash(ngettext("{} member is already present.",
+                               "{} members are already present.",
+                               registered_users.__len__()).format(registered_users.__len__()), 'success')
+
+            if bool(not_registered):
```
ThiefMaster: no need for the `bool()`

indico/modules/categories/controllers/management.py
```diff
@@ -512,3 +514,40 @@ def _process(self):
                 self.role.members.add(user)
                 logger.info('User %r added to role %r by %r', user, self.role, session.user)
         return jsonify_data(html=_render_role(self.role, collapsed=False))
+
+
+class RHAddCategoryRoleMembersImportCSV(RHManageCategoryRole):
```
ThiefMaster: lets create a mixin from this so this can be simplified to

    class RHAddCategoryRoleMembersImportCSV(ImportRoleMembersMixin, RHManageCategoryRole):
        pass

indico/modules/events/roles/util.py
ParthS007: okay, I will check it.

indico/modules/categories/controllers/management.py
```diff
@@ -512,3 +513,8 @@ def _process(self):
                 self.role.members.add(user)
                 logger.info('User %r added to role %r by %r', user, self.role, session.user)
         return jsonify_data(html=_render_role(self.role, collapsed=False))
+
+
+class RHAddCategoryRoleMembersImportCSV(ImportRoleMembersMixin, RHManageCategoryRole):
+    """Add users to a category role from CSV"""
+    pass
```
ThiefMaster: the `pass` isn't needed since the docstring already make the class' body non-empty

indico/modules/events/roles/templates/_roles.html
```diff
@@ -83,6 +83,11 @@
                     <div class="empty">{% trans %}There are no members yet.{% endtrans %}</div>
                 {% endif %}
                     <div class="toolbar f-j-end">
+                         <button class="i-button highlight text-color icon-upload"
+                                data-href="{{ url_for('.add_members_import_csv', role) }}"
+                                data-update="#event-role-{{ role.id }}"
+                                data-ajax-dialog>{% trans %}Import users from CSV{% endtrans %}
```
ThiefMaster: wrap after the `>`

indico/util/roles.py
```diff
@@ -0,0 +1,84 @@
...
+from indico.modules.events.roles.forms import ImportMembersCSVForm
+from indico.modules.users import User
+from indico.util.i18n import _, ngettext
+from indico.util.string import to_unicode, validate_email
+from indico.web.flask.templating import get_template_module
+from indico.web.util import jsonify_data, jsonify_template
+
+
+class ImportRoleMembersMixin():
+
```
ThiefMaster: don't create old-style classes - always inherit from `object` if there's no other base class. also, no empty line after the class header (it's only needed after docstrings)

indico/util/roles.py
```diff
@@ -0,0 +1,84 @@
...
+    def import_members_from_csv(self, f):
+        """Import members from a CSV file into an event role."""
+        reader = csv.reader(f.read().splitlines())
+        emails = set()
+
+        for num_row, row in enumerate(reader, 1):
+            if len(row) != 1:
+                continue
```
ThiefMaster: I'd raise an error here to tell the user that the row doesn't contain the right amount of columns

indico/modules/events/roles/blueprint.py
```diff
@@ -23,3 +24,6 @@
 
 _bp.add_url_rule('/<int:role_id>/members', 'add_members', RHAddEventRoleMembers, methods=('POST',))
 _bp.add_url_rule('/<int:role_id>/members/<int:user_id>', 'remove_member', RHRemoveEventRoleMember, methods=('DELETE',))
+
+_bp.add_url_rule('/<int:role_id>/members/import', 'add_members_import_csv',
```
mic4ael: You can still put `RHAddEventRoleMembersImportCSV` in the previous line

indico/util/roles.py
```diff
@@ -0,0 +1,84 @@
...
+            try:
+                email = to_unicode(row[0]).strip().lower()
+            except ValueError:
```
ThiefMaster: this won't ever happen

indico/modules/events/roles/forms.py
```diff
@@ -37,3 +38,8 @@ def validate_code(self, field):
             query = query.filter(EventRole.id != self.role.id)
         if query.has_rows():
             raise ValueError(_('A role with this code already exists.'))
+
+
+class ImportMembersCSVForm(IndicoForm):
+    source_file = FileField(_("Source File"), [InputRequired(), DataRequired()], accepted_file_types='.csv,.txt')
```
mic4ael: mixing `"` with `'`. Let's use `"` only if necessary

indico/util/roles.py
```diff
@@ -0,0 +1,84 @@
...
+                raise UserValueError(_('Row {}: malformed CSV data').format(num_row))
+
```
ThiefMaster:
```python
if not email:
    continue
```

indico/modules/events/roles/templates/_roles.html
mic4ael: `from CSV file`?

indico/util/roles.py
```diff
@@ -0,0 +1,84 @@
...
+from indico.core.errors import UserValueError
+from indico.modules.events.roles.forms import ImportMembersCSVForm
+from indico.modules.users import User
+from indico.util.i18n import _, ngettext
+from indico.util.string import to_unicode, validate_email
+from indico.web.flask.templating import get_template_module
+from indico.web.util import jsonify_data, jsonify_template
+
+
+class ImportRoleMembersMixin():
```
mic4ael: inherit from `object`

indico/util/roles.py
```diff
@@ -0,0 +1,84 @@
...
+        for num_row, row in enumerate(reader, 1):
+            if len(row) != 1:
+                continue
+            try:
+                email = to_unicode(row[0]).strip().lower()
+            except ValueError:
+                raise UserValueError(_('Row {}: malformed CSV data').format(num_row))
+
+            if email and not validate_email(email):
+                raise UserValueError(_("Row {row}: invalid email address: {email}").format(row=num_row, email=email))
```
mic4ael: replace `"` with `'`? Since two lines below you are using `'` (let's just be consistent)

indico/modules/events/roles/templates/_roles.html
ParthS007: `from CSV` would be good, I think

mic4ael: an example of what we have in a few other places
```html
<a class="i-button"
    title="{% trans %}Import contributions from CSV file{% endtrans %}"
    data-href="{{ url_for('.contributions_import', event) }}"
    data-title="{% trans %}Import contributions from CSV File{% endtrans %}"
    data-ajax-dialog>
    {%- trans %}Import{% endtrans -%}
</a>
```

indico/modules/events/roles/templates/_roles.html
```diff
@@ -82,13 +82,20 @@
...
-                                 data-method="POST"
-                                 data-update="#event-role-{{ role.id }}">{% trans %}Add user{% endtrans %}
-                         </button>
-                    </div>
+                <div class="toolbar f-j-end">
+                    <button class="i-button highlight text-color icon-upload"
+                            data-href="{{ url_for('.add_members_import_csv', role) }}"
+                            data-update="#event-role-{{ role.id }}"
+                            data-ajax-dialog>
+                        {% trans %}Import users from CSV File{% endtrans %}
```
ThiefMaster: lowercase "file"

indico/util/roles.py
```diff
@@ -0,0 +1,69 @@
...
+            email = to_unicode(row[0]).strip().lower()
+
+            if email and not validate_email(email):
+                raise UserValueError(_('Row {row}: invalid email address: {email}').format(row=num_row, email=email))
+            if email in emails:
+                raise UserValueError(_('Row {}: email address is not unique').format(num_row))
+            emails.add(email)
+
+        users = set(User.query.filter(~User.is_deleted, User.all_emails.in_(emails)))
+        registered_users_email = {user.email for user in users}
```
ThiefMaster: `users_emails`

indico/util/roles.py
```diff
@@ -0,0 +1,69 @@
...
+        new_members = users - self.role.members
+        return new_members, users, unknown_emails
+
+    def _process(self):
+        form = ImportMembersCSVForm()
+
+        if form.validate_on_submit():
+            new_members, users, unknown_emails = self.import_members_from_csv(form.source_file.data)
+            if form.remove_existing.data:
+                self.role.members = users
```
ThiefMaster: also log removed users

indico/modules/events/roles/controllers.py
```diff
@@ -83,6 +84,7 @@ class RHManageEventRole(RHManageEventBase):
     def _process_args(self):
         RHManageEventBase._process_args(self)
         self.role = EventRole.get_one(request.view_args['role_id'])
+        self.logger = logger
```
ThiefMaster: You don't need this here, since you already have access to `self.logger` in the classes that need it.

indico/modules/events/roles/controllers.py
```diff
@@ -145,3 +147,9 @@ def _process(self):
                                data={'Name': user.full_name,
                                      'Email': user.email})
         return jsonify_data(html=_render_role(self.role, collapsed=False))
+
+
+class RHAddEventRoleMembersImportCSV(ImportRoleMembersMixin, RHManageEventRole):
+    """Add users to a category role from CSV"""
```
ThiefMaster: s/category/event

indico/modules/events/roles/forms.py
```diff
@@ -37,3 +38,8 @@ def validate_code(self, field):
             query = query.filter(EventRole.id != self.role.id)
         if query.has_rows():
             raise ValueError(_('A role with this code already exists.'))
+
+
+class ImportMembersCSVForm(IndicoForm):
+    source_file = FileField(_('Source File'), [InputRequired(), DataRequired()], accepted_file_types='.csv,.txt')
+    remove_existing = BooleanField('Remove existing members', widget=SwitchWidget())
```
ThiefMaster: missing i18n

indico/modules/events/roles/templates/import_members.html
```diff
@@ -0,0 +1,15 @@
+{% from 'forms/_form.html' import simple_form  %}
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('highlight', large_icon=true) %}
+    {% trans -%}
+        You should upload a CSV or TXT file with exactly 1 column of user email.
```
ThiefMaster: I think in this case "with one email address per line" is clearer

indico/modules/categories/blueprint.py
```diff
@@ -57,6 +57,8 @@ def _redirect_event_creation(category_id, event_type):
 _bp.add_url_rule('/manage/roles/<int:role_id>/members', 'add_members', RHAddCategoryRoleMembers, methods=('POST',))
 _bp.add_url_rule('/manage/roles/<int:role_id>/members/<int:user_id>', 'remove_member', RHRemoveCategoryRoleMember,
                  methods=('DELETE',))
+_bp.add_url_rule('/manage/roles/<int:role_id>/members/import', 'add_members_import_csv',
+                 RHAddCategoryRoleMembersImportCSV, methods=('GET', 'POST'))
```
mic4ael: I think this `RH` name doesn't need the `Add` part. `RHCategoryRolesMembersImportCSV` should be enough

indico/modules/events/roles/blueprint.py
```diff
@@ -23,3 +24,6 @@
 
 _bp.add_url_rule('/<int:role_id>/members', 'add_members', RHAddEventRoleMembers, methods=('POST',))
 _bp.add_url_rule('/<int:role_id>/members/<int:user_id>', 'remove_member', RHRemoveEventRoleMember, methods=('DELETE',))
+
+_bp.add_url_rule('/<int:role_id>/members/import', 'add_members_import_csv', RHAddEventRoleMembersImportCSV,
```
mic4ael: The same as before. Do you really need that `Add` part in tthe name of the `RH`?

indico/modules/events/roles/controllers.py
ParthS007: Ohh, I forgot to remove it.

indico/modules/categories/blueprint.py
ParthS007: Yeah, that's also fine.

indico/modules/events/roles/blueprint.py
ParthS007: changed it.

PR #4307, opened 2020-02-21
Add optional message on pre-booking acceptance

### LINKED ISSUE #4086: Room Booking: add optional message on pre-booking acceptance
A user suggested that we allow a message to be provided upon acceptance of a booking, pretty much like when it is rejected. In this case, the message wouldn't be mandatory and would also be included in the acceptance e-mail.
@ThiefMaster suggested a button similar to what GitHub uses for merging options:

image

Underneath the "arrow" menu we could have an "Accept with message" option.

Leats:
Closes #4086.

The optional message is saved in the logs and is sent via email but does not have its own column in the database.

### DIFF
--- a/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
+++ b/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
@@ -70,6 +70,15 @@ class BookingDetails extends React.Component {
   state = {
     occurrencesVisible: false,
     activeConfirmation: null,
+    acceptanceFormVisible: false,
+  };
+
+  showAcceptanceForm = () => {
+    this.setState({acceptanceFormVisible: true});
+  };
+
+  hideAcceptanceForm = () => {
+    this.setState({acceptanceFormVisible: false});
   };
 
   showOccurrences = () => {
@@ -384,7 +393,7 @@ class BookingDetails extends React.Component {
 
   renderActionButtons = (canCancel, canReject, showAccept) => {
     const {bookingStateChangeInProgress} = this.props;
-    const {actionInProgress, activeConfirmation} = this.state;
+    const {actionInProgress, activeConfirmation, acceptanceFormVisible} = this.state;
     const rejectButton = (
       <Button
         type="button"
@@ -397,7 +406,16 @@ class BookingDetails extends React.Component {
       />
     );
 
-    const renderForm = ({
+    const acceptWithMessage = (
+      <Button
+        type="button"
+        icon="edit outline"
+        styleName="accept-with-message"
+        disabled={bookingStateChangeInProgress || acceptanceFormVisible}
+      />
+    );
+
+    const renderRejectionForm = ({
       handleSubmit,
       hasValidationErrors,
       submitSucceeded,
@@ -421,6 +439,36 @@ class BookingDetails extends React.Component {
             primary
           >
             <Translate>Reject</Translate>
+          </Button>
+        </Form>
+      );
+    };
+
+    const renderAcceptanceForm = ({
+      handleSubmit,
+      hasValidationErrors,
+      submitSucceeded,
+      submitting,
+    }) => {
+      return (
+        <Form styleName="rejection-form" onSubmit={handleSubmit}>
+          <FinalTextArea
+            name="reason"
+            placeholder={Translate.string(
+              'You can provide an optional message when accepting this booking'
+            )}
+            disabled={submitSucceeded}
+            rows={2}
+            autoFocus
+          />
+          <Button
+            type="submit"
+            disabled={submitting || hasValidationErrors || submitSucceeded}
+            loading={submitting}
+            floated="right"
+            primary
+          >
+            <Translate>Accept</Translate>
           </Button>
         </Form>
       );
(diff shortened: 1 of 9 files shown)


indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -70,6 +70,15 @@ class BookingDetails extends React.Component {
   state = {
     occurrencesVisible: false,
     activeConfirmation: null,
+    activeAcceptanceForm: false,
+  };
+
+  activateAcceptanceForm = () => {
```
ThiefMaster: show/hide sounds better than activate/deactivate IMO

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -70,6 +70,15 @@ class BookingDetails extends React.Component {
   state = {
     occurrencesVisible: false,
     activeConfirmation: null,
+    activeAcceptanceForm: false,
```
ThiefMaster: `acceptanceFormVisible`? (this one is a boolean while `activeConfirmation` above which you probably used as a naming hint is a string indicating *which* confirm prompt is visible)

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -458,20 +506,35 @@ class BookingDetails extends React.Component {
...
+            <Button
+              type="button"
+              icon="check circle"
+              onClick={() => this.changeState('approve')}
+              loading={actionInProgress === 'approve' && bookingStateChangeInProgress}
+              disabled={bookingStateChangeInProgress || activeAcceptanceForm}
+              content={Translate.string('Accept booking')}
+            />
+            <Popup
+              trigger={acceptWithMessage}
```
ThiefMaster: You could put the `<Button ... />` JSX directly in here, but nothing wrong with leaving it separate, especially since that's what we already do for `rejectionButton`.

I think the main reason why we have some of those buttons separate is actually that we weren't using prettier for code formatting back then, and it was really messy to have this kind of nesting at that point.

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -458,20 +506,35 @@ class BookingDetails extends React.Component {
...
+              position="bottom right"
+              on="click"
+              onOpen={this.activateAcceptanceForm}
+              onClose={this.deactivateAcceptanceForm}
+            >
+              <FinalForm
```
ThiefMaster: I think both this form and the rejection one could use a subscription prop to avoid unnecessary re-renders:

    subscription={{hasValidationErrors: true, submitSucceeded: true, submitting: true, pristine: true}}

(you basically list all props you want to receive in the form's render function, and anything else which you don't use anyway won't be passed and thus also won't trigger a re-render)

indico/modules/rb/client/js/common/bookings/BookingDetails.module.scss
```diff
@@ -54,3 +54,7 @@
 .rejection-form {
     width: 350px;
 }
+
+.accept-with-message {
+    border-left: 1px solid darken($color: $green, $amount: 5%) !important;
```
ThiefMaster: in other places we use the shorter way of calling scss functions: `darken($green, 5%)`

indico/modules/rb/templates/emails/reservations/confirmation_email_to_user.txt
```diff
@@ -4,6 +4,11 @@
 {% block state -%}
 Your booking has been ACCEPTED.
 This is the final confirmation.
+
+{% if reservation.acceptance_reason %}
```
ThiefMaster: I think you need `{%- ...}` here to swallow the whitespace in case there's no reason, and add the empty line inside the if's body instead. Probably also want `{%- endif -%}` below to avoid having an extra linebreak a the end.

Best to test with and without a commend and make sure there's no weird change in empty lines.

indico/modules/rb/models/reservations_test.py
```diff
@@ -194,6 +194,23 @@ def test_reject(smtp, create_reservation, dummy_user, silent):
...
+                                     end_dt=date.today() + relativedelta(days=10, hour=17),
+                                     repeat_frequency=RepeatFrequency.DAY)
+    assert not reservation.is_rejected
+    assert not any(occ.is_rejected for occ in reservation.occurrences)
+    reservation.accept(user=dummy_user, reason=reason)
+    assert reservation.is_accepted
+    assert len(smtp.outbox) == 2
+
+    if reason:
+        assert ['My reason' in mail.as_string() for mail in smtp.outbox]
```
ThiefMaster: You could also add an assertion on the log message here

indico/modules/rb/templates/emails/reservations/confirmation_email_to_user.txt
```diff
@@ -4,6 +4,11 @@
 {% block state -%}
 Your booking has been ACCEPTED.
 This is the final confirmation.
+{% if reason %}
+
+The following message was provided by the room manager:
+{{reason }}
```
ThiefMaster: missing space after `{{`

indico/modules/rb/templates/emails/reservations/confirmation_email_to_user.txt
```diff
@@ -4,6 +4,11 @@
 {% block state -%}
 Your booking has been ACCEPTED.
 This is the final confirmation.
+{% if reason %}
```
ThiefMaster: Matter of style, but if you use `{%- if reason %}` you can add a linebreak before this line to keep things a bit more separate in the template. Jinja removes any whitespace on the side where the `-` is.

PR #4316, opened 2020-02-25
Enhance cancellation of recurring bookings

### LINKED ISSUE #4092: Room booking: deleting an occurence of recurring reservation is not obvious
If you created a recurring reservation and you go to the booking calendar view, select an occurence that you want to delete and click the "Cancel booking" button, you get the message:

> Are you sure you want to cancel this booking? This will cancel future occurrences of this booking.

which let think it will delete the occurence you selected and all the following when it fact it delete the whole booking (all occurences). This is misleading and probably the date of "this booking" should be displayed in the message or the message should be more clear that it will delete all the occurences in the booking.

If in fact, you want to delete one specific occurence, you cannot do it from this button but you have to first click "See on timeline" and then select the occurence you want to delete. It is far from obvious if you never did it. IMO, it should either be better exposed or their should be some text in the windows explaining what to do for the different kind of deletions.

Leats:
Single and recurring bookings now have different confirmation messages, making it clearer what exactly gets deleted.
Closes #4092. 

The new layout will look like this:
Screenshot from 2020-02-25 17-07-42

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -48,6 +48,8 @@ Improvements
 - Allow importing event role members from a CSV file (:issue:`4301`)
 - Allow optional comment when accepting a pre-booking (:issue:`4086`)
 - Log event restores in event log (:issue:`4309`)
+- Warn about cancelling/rejecting whole recurring bookings instead of just
+  specific occurrences (:issue:`4092`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
+++ b/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
@@ -30,7 +30,7 @@ import {
 
 import {toMoment, serializeDate} from 'indico/utils/date';
 import {Param, Plural, PluralTranslate, Singular, Translate} from 'indico/react/i18n';
-import {FinalTextArea} from 'indico/react/forms';
+import {FinalCheckbox, FinalTextArea} from 'indico/react/forms';
 import {Responsive} from 'indico/react/util';
 import {ClipboardButton} from 'indico/react/components';
 import {DailyTimelineContent, TimelineLegend} from '../timeline';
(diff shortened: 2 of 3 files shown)


ThiefMaster:
Looks and works very well so far! What about adding some text indicating that you're rejecting the whole reservation in the popup shown when you are about to reject an already-confirmed reservation as well?

Changelog entry could be useful as well.

CHANGES.rst
```diff
@@ -48,6 +48,7 @@ Improvements
 - Allow importing event role members from a CSV file (:issue:`4301`)
 - Allow optional comment when accepting a pre-booking (:issue:`4086`)
 - Log event restores in event log (:issue:`4309`)
+- Enhance cancellation of recurring bookings (:issue:`4092`)
```
ThiefMaster: I'd go a bit more into detail here. What about something like this?

```rst
- Warn about cancelling/rejecting whole recurring bookings instead of just
  specific occurrences (:issue:`4092`)
```

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -431,9 +438,23 @@ class BookingDetails extends React.Component {
...
+                  acceptRejection: checked,
+                });
+              }}
+            />
+          )}
           <Button
             type="submit"
-            disabled={submitting || pristine || hasValidationErrors || submitSucceeded}
+            disabled={
+              submitting || pristine || hasValidationErrors || submitSucceeded || !acceptRejection
```
ThiefMaster: Since we're inside a FinalForm there's probably a nicer way to actually mark the form as invalid until the checkbox has been checked, but I have to try this out myself first...

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -487,21 +508,58 @@ class BookingDetails extends React.Component {
...
+                )}
+                <Button
+                  onClick={() => {
+                    this.hideConfirm();
+                  }}
+                  content={PluralTranslate.string(
+                    'Cancel booking',
+                    'Cancel all occurrences',
+                    occurrenceCount
+                  )}
```
ThiefMaster: I think using separate Translates (with the `?:` conditional) would be slightly more correct here, since the translation does not depend on the number, just on the fact whether it's a single occurrence or not.

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
ThiefMaster: BTW, I think this will currently break rejecting a pending booking, since this condition still runs, but there's no checkbox to select.

PR #4322, opened 2020-02-26
Editing view

### LINKED ISSUE #4278: Find something between management/display view for editing timeline and similar pages
This could for example be based on the management layout, but without showing stuff like the event creator and a better title than "Switch to display view" to get back to the normal event view.

panagiotappl:
closes #4278


indico/modules/events/editing/client/js/components/EditingView/EditingFooter.jsx
```diff
@@ -0,0 +1,46 @@
...
+import contactURL from 'indico-url:core.contact';
+import tosURL from 'indico-url:legal.display_tos';
+
+import React from 'react';
+import {Image} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+
+import './EditingFooter.module.scss';
+
+export default function Editingfooter() {
```
ThiefMaster: should be `EditingFooter` but do we really want to put the "base" part of this in the editing module?

indico/modules/events/editing/client/js/components/EditingView/EditingFooter.jsx
```diff
@@ -0,0 +1,46 @@
...
+      <div styleName="logo">
+        <Image src={logoURL({filename: 'indico_small.png'})} styleName="image" />
+        <span styleName="text">
+          <Translate>Powered by </Translate>
+          <a href="http://getindico.io">Indico</a>
+        </span>
+      </div>
+      <div styleName="information">
+        <span>
+          <a href="https://learn.getindico.io">
```
ThiefMaster: I think we have a configurable URL for this (since e.g. in case of CERN indico we want to point somewhere else)

indico/modules/events/editing/client/js/components/EditingView/MenuBar.jsx
```diff
@@ -0,0 +1,72 @@
...
+import {Header, Icon} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+
+import './MenuBar.module.scss';
+
+export default function MenuBar({eventId, eventTitle, menuItems}) {
+  return (
+    <div styleName="menu-bar">
+      <Header as="h2" styleName="header">
+        Paper Editing
```
ThiefMaster: i18n

indico/modules/events/editing/client/js/components/EditingView/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+  const [menuItems, setMenuItems] = useState([]);
+  const {eventId} = staticData;
+
+  useEffect(() => {
+    const fetchMenuEntries = async () => {
+      let response;
+      try {
+        response = await indicoAxios.get(menuEntriesURL({confId: eventId}));
+      } catch (e) {
+        handleAxiosError(e);
```
ThiefMaster: return after this

indico/modules/events/editing/views.py
```diff
@@ -14,3 +14,6 @@ class WPEditing(WPEventManagement):
     template_prefix = 'events/editing/'
     # TODO: remove mathjax later on from here (it is just here to make the timeline work)
     bundles = ('module_events.editing.js', 'module_events.editing.css', 'mathjax.js')
+
+    def _get_footer(self):
+        return u""
```
ThiefMaster: `''` since the file uses unicode literals

indico/modules/events/editing/client/js/components/EditingView/EditingFooter.jsx
panagiotappl: We have a URL for this but is the a dedicated endpoint? Apart from the one we have created in RB which returns other information as well.

ThiefMaster: We should move it out from the RB one to a more generic place. Feel free to use the RB endpoint for the time being.

panagiotappl: We should actually separate the RB one into two, as it also contains the room sprite and other rb related stuff.

indico/modules/events/editing/client/js/components/EditingView/Footer.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+import React, {useState, useEffect} from 'react';
+import {Image, Modal} from 'semantic-ui-react';
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+import {Translate} from 'indico/react/i18n';
+import {camelizeKeys} from 'indico/utils/case';
+
+import './Footer.module.scss';
+
+export default function Editingfooter() {
```
ThiefMaster: `Footer`

indico/modules/events/editing/client/js/editing.jsx
```diff
@@ -0,0 +1,67 @@
...
+    return;
+  }
+
+  const eventId = parseInt(editingElement.dataset.eventId, 10);
+  const eventTitle = editingElement.dataset.eventTitle;
+
+  let fileTypes, tags;
+
+  try {
+    const [fileTypeResponse, tagResponse] = await Promise.all([
```
ThiefMaster: It looks like this was copied from `timeline.jsx`. If yes, can it be removed there?

panagiotappl: `timeline.jsx` was supposed to be deleted after this file was added

indico/modules/events/editing/client/js/components/EditingView/EditingView.module.scss
```diff
@@ -0,0 +1,28 @@
...
+
+@import 'base/palette';
+
+.editing-view {
+    display: flex;
+
+    .contents {
+        display: flex;
+        flex-direction: column;
+        width: 100%;
```
mic4ael: I think with flex you can just use `flex: 1`?

indico/modules/events/editing/client/js/components/EditingView/Footer.jsx
```diff
@@ -0,0 +1,86 @@
...
+import {Translate} from 'indico/react/i18n';
+import {camelizeKeys} from 'indico/utils/case';
+
+import './Footer.module.scss';
+
+export default function Footer() {
+  const [config, setConfig] = useState(null);
+  const [tosModal, setTosModal] = useState(false);
+
+  useEffect(() => {
```
mic4ael: Maybe just use `useIndicoAxios`?

indico/modules/core/controllers.py
```diff
@@ -283,3 +285,22 @@ def _process(self):
         self.user.reset_signing_secret()
         flash(_('All your token-based links have been invalidated'), 'success')
         return redirect(url_for('api.user_profile'))
+
+
+class RHConfig(RH):
+    def _process(self):
+        tos_url = legal_settings.get('tos_url')
+        tos_html = sanitize_html(legal_settings.get('tos')) or None
```
mic4ael: Why `or None`?

indico/modules/events/editing/client/js/components/EditingView/Footer.module.scss
```diff
@@ -0,0 +1,43 @@
...
+
+        span {
+            color: $gray;
+        }
+
+        span:not(:last-child)::after {
+            content: ' | ';
+        }
+    }
+
```
mic4ael: redundant empty lne

indico/modules/events/editing/client/js/components/EditingView/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+import Footer from './Footer';
+
+import './EditingView.module.scss';
+
+export default function EditingView({eventTitle}) {
+  const staticData = useSelector(getStaticData);
+  const [menuItems, setMenuItems] = useState([]);
+  const {eventId} = staticData;
+
+  useEffect(() => {
```
mic4ael: you could probably use `useIndicoAxios` here

indico/modules/core/controllers.py
ThiefMaster: it may be an empty string, but we want explicit none on the client side

mic4ael: Yeah, but I don't think we explicitly check for `None`. Empty string will anyway evaluate to false:

```javascript
{hasTos && tosHtml && (
    <Modal
    open={tosModal}
    closeIcon
    onClose={() => setTosModal(false)}
    trigger={
        <span>
        <a onClick={() => setTosModal(true)}>
            <Translate>Terms and Conditions</Translate>
        </a>
        </span>
    }
    >
    <Modal.Header>
        <Translate>Terms and Conditions</Translate>
    </Modal.Header>
    <Modal.Content>
        <div dangerouslySetInnerHTML={{__html: tosHtml}} />
    </Modal.Content>
    </Modal>
)}
```

mic4ael: Ah, I guess it is better to have `null` here to avoid rendering empty strings in the DOM.

indico/modules/events/editing/client/js/components/EditingView/EditingView.module.scss
panagiotappl: For what purpose?

mic4ael: To make it grow/shrink appropriately thus this is more `flex`y than using `width`

indico/modules/events/editing/client/js/components/EditingView/Footer.jsx
```diff
@@ -0,0 +1,78 @@
...
+  const config = data;
+  if (!config) {
+    return null;
+  }
+
+  const {helpURL, hasTos, contactEmail, tosHtml} = config;
+  return (
+    <div styleName="editing-footer">
+      <div styleName="logo">
+        <Image src={logoURL({filename: 'indico_small.png'})} styleName="image" />
```
ThiefMaster: Like this there's currently no way to override the footer logo like we do everywhere else (e.g. to put the CERN logo in the footer), but I don't see any good solution here for now, so I think we need to live with this for now, and find some solution for it later that does not involve react overrides / plugins (we can't expect people to write a plugin for this, and we don't want to do this for our own customizations since it's moer maintenance).

indico/modules/events/editing/client/js/components/EditingView/MenuBar.jsx
```diff
@@ -0,0 +1,71 @@
...
+              <a key={item.name} href={item.url}>
+                <li styleName="inner-list-item">{item.title}</li>
+              </a>
+            ))}
+          </ul>
+        </li>
+      </ul>
+      <ul styleName="list">
+        <li>
+          <Translate>OTHER VIEWS</Translate>
```
ThiefMaster: I'd write this normally and [use CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform) to capitalize it.

indico/modules/events/editing/schemas.py
```diff
@@ -244,3 +244,17 @@ def _validate_file_types(self, file_types):
         event_conditions = editing_settings.get(event, 'review_conditions')
         if any(condition_types == set(event_condition) for event_condition in event_conditions.values()):
             raise ValidationError(_('Conditions have to be unique'))
+
+
+class SideMenuItemSchema(mm.Schema):
+    name = fields.String(required=True)
+    title = fields.String(required=True)
+    url = fields.String(required=True)
+    weight = fields.Int()
```
ThiefMaster: Only return stuff that's needed on the client. Assuming ordering is done server-side, we don't need that here.

indico/modules/events/editing/schemas.py
```diff
@@ -244,3 +244,17 @@ def _validate_file_types(self, file_types):
...
+    active = fields.Boolean()
+    disabled = fields.Boolean()
+    section = fields.String()
+    icon = fields.String()
+
+
+side_menu_items_schema = SideMenuItemSchema(many=True)
```
ThiefMaster: I'd try to avoid those instances; they are not very useful and instantiating the schema when you need it works perfectly fine - and it avoids having e.g. singular and plural instances around (with/without `many`)

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -130,3 +130,11 @@ def _process_PATCH(self, file_types):
         new_conditions[self.uuid] = file_types
         editing_settings.set(self.event, 'review_conditions', new_conditions)
         return '', 204
+
+
+class RHMenuEntries(RHEditingManagementBase):
+    def _process(self):
+        from indico.core import signals
+        from indico.util.signals import named_objects_from_signal
```
ThiefMaster: those imports should be safe at the top of the module

indico/modules/events/editing/schemas.py
```diff
@@ -244,3 +244,17 @@ def _validate_file_types(self, file_types):
...
             raise ValidationError(_('Conditions have to be unique'))
+
+
+class SideMenuItemSchema(mm.Schema):
+    name = fields.String(required=True)
+    title = fields.String(required=True)
+    url = fields.String(required=True)
+    weight = fields.Int()
+    active = fields.Boolean()
+    disabled = fields.Boolean()
```
ThiefMaster: Do we even return disabled items when jsonifying them?
If not, let's not include it here since it'll always be false anyway.
If yes, should we be doing it?

indico/modules/events/editing/client/js/components/EditingView/MenuBar.jsx
```diff
@@ -0,0 +1,71 @@
...
+const menuEntryPropTypes = {
+  name: PropTypes.string.isRequired,
+  title: PropTypes.string.isRequired,
+  url: PropTypes.string.isRequired,
+  weight: PropTypes.number.isRequired,
+  active: PropTypes.bool.isRequired,
+  disabled: PropTypes.bool.isRequired,
+  section: PropTypes.string.isRequired,
+  icon: PropTypes.string,
+};
```
ThiefMaster: Let's only include proptypes here which we actually use. `shape` doesn't complain if there's more in the object (only `exact` would)

indico/modules/events/editing/client/js/components/EditingView/MenuBar.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+const menuEntryPropTypes = {
+  name: PropTypes.string.isRequired,
+  title: PropTypes.string.isRequired,
+  url: PropTypes.string.isRequired,
+  weight: PropTypes.number.isRequired,
+  active: PropTypes.bool.isRequired,
+  disabled: PropTypes.bool.isRequired,
+  section: PropTypes.string.isRequired,
+  icon: PropTypes.string,
```
ThiefMaster: shouldn't we take the icon into account if one has been defined on the menu item?

indico/modules/events/editing/client/js/components/EditingView/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+import Timeline from '../Timeline';
+import MenuBar from './MenuBar';
+import Footer from './Footer';
+
+import './EditingView.module.scss';
+
+export default function EditingView({eventTitle}) {
+  const staticData = useSelector(getStaticData);
+  const {eventId} = staticData;
```
ThiefMaster:
```suggestion
  const {eventId} = useSelector(getStaticData);
```

indico/modules/events/abstracts/__init__.py
```diff
@@ -46,6 +46,7 @@ def _clear_boa_cache(sender, obj=None, **kwargs):
     clear_boa_cache(event)
 
 
+@signals.menu.items.connect_via('event-editing-sidemenu')
```
ThiefMaster: The CfA link in the hybrid view shouldn't point to the management page but rather the reviewing area, i.e. in a new function that uses the same logic used in `_extend_event_menu`.

Similar for the CfP link.

indico/modules/events/editing/client/js/components/EditingView/Footer.jsx
```diff
@@ -0,0 +1,78 @@
...
+export default function Footer() {
+  const [tosModal, setTosModal] = useState(false);
+
+  const {data} = useIndicoAxios({
+    url: configURL(),
+    camelize: true,
+    trigger: 'once',
+  });
+
+  const config = data;
```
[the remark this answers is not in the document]
panagiotappl: It was supposed to be fixed but maybe got lost in rebasing

indico/modules/events/editing/client/js/components/EditingView/MenuBar.jsx
panagiotappl: What are these icon strings exactly? Do they align with semantic-ui icon names?

ThiefMaster: whatever you specify when defining the menu item in the signal.. so for a react-based menu you could use sui names; the existing ones use the legacy icons (from `_icons.scss`)

indico/modules/events/abstracts/__init__.py
panagiotappl: MenuEntryData contains an endpoint not a url (like the management menu classes). Should I create a custom field that serializes the endpoint to a url in the schema or is using the endpoint with a url value acceptable here?

ThiefMaster: But that's just for the conference menu, keep using `SideMenuItem` of course - my comment was just about the visibility logic. Sorry if that wasn't clear.

indico/modules/events/editing/client/js/components/EditingView/Footer.jsx
```diff
@@ -0,0 +1,78 @@
...
+      <div styleName="logo">
+        <Image src={logoURL({filename: 'indico_small.png'})} styleName="image" />
+        <span styleName="text">
+          <Translate>Powered by </Translate>
+          <a href="http://getindico.io">Indico</a>
+        </span>
+      </div>
+      <div styleName="information">
+        {helpURL && (
+          <span>
```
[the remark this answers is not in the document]
panagiotappl: They are needed before of this 
```
span {
    color: $gray;
}

span:not(:last-child)::after {
    content: ' | ';
}
```
but maybe I should make it more obvious

indico/modules/events/editing/client/js/components/EditingView/Footer.jsx
```diff
@@ -0,0 +1,78 @@
...
+    camelize: true,
+    trigger: 'once',
+  });
+
+  const config = data;
+  if (!config) {
+    return null;
+  }
+
+  const {helpURL, hasTos, contactEmail, tosHtml} = config;
```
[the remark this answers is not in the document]
panagiotappl: Do we show privacy policy anywhere in the footer? Or do you mean that I should just include it in the schema.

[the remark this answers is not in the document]
ThiefMaster: yes, both in normal pages and the rb menu i think (if a link or text is configured)

indico/modules/events/papers/__init__.py
```diff
@@ -36,6 +36,17 @@ def _extend_event_management_menu(sender, event, **kwargs):
                         section='organization')
 
 
+@signals.menu.items.connect_via('event-editing-sidemenu')
+def _extend_editing_menu(sender, event, **kwargs):
+    from indico.modules.events.papers.util import has_contributions_with_user_paper_submission_rights
+    if not session.user or not event.has_feature('papers'):
+        return False
```
ThiefMaster: `return None`

PR #4331, opened 2020-03-02
Add quick cancel link to room booking reminder emails

### LINKED ISSUE #4324: Add "quick cancel" link to room booking reminder emails
This link would show the booking details page, and immediately open a confirmation to cancel this particular occurrence.

Things to keep in mind:

- you may click the link after the occurrence finished, so you cannot cancel it anymore. I'd just ignore it in that case and not show anything besides the normal booking details dialog
- for a non-recurring booking including this link is not very useful since you can just cancel the booking itself

For the link in the email I'd use `.../booking/<id>/cancel/<date>`; and then just redirect to the normal booking details modal with another arg (see `payload` in `ModalController`) to specify the day that should be cancelled.

related: https://github.com/indico/indico/issues/4092#issuecomment-565469493

Leats:
Closes #4324.

### DIFF
--- a/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
+++ b/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
@@ -58,13 +58,17 @@ class BookingDetails extends React.Component {
     bookingStateChangeInProgress: PropTypes.bool.isRequired,
     actions: PropTypes.exact({
       deleteBooking: PropTypes.func.isRequired,
+      changeBookingOccurrenceState: PropTypes.func,
       changeBookingState: PropTypes.func.isRequired,
       openBookingDetails: PropTypes.func.isRequired,
+      fetchBookingDetails: PropTypes.func.isRequired,
     }).isRequired,
+    cancelDate: PropTypes.string,
   };
 
   static defaultProps = {
     onClose: () => {},
+    cancelDate: null,
   };
 
   state = {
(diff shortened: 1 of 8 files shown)


indico/modules/rb/templates/emails/reservations/reminders/upcoming_occurrence.html
```diff
@@ -18,6 +18,10 @@
             </li>
         {%- endfor -%}
     </ul>
+    {%- if occurrences | length > 1 -%}
+        If you wish to cancel all occurrences, please <a href="{{ reservation.external_details_url }}">cancel the
```
ThiefMaster: I'd wrap after "please" to avoid wrapping in the middle of a tag's content (especially of a tag that's not in its own line). [same in the block below]

Since it's HTML whitespace doesn't really matter here anyway. In any case, the next line has 4 spaces indentation too much 👾

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -114,6 +118,76 @@ class BookingDetails extends React.Component {
...
+    } else if (!bookings[cancelDate][0].canCancel) {
+      return (
+        <Modal open size="tiny" onClose={() => openBookingDetails(id)}>
+          <Modal.Header>
+            <Translate>Cancellation not possible</Translate>
+          </Modal.Header>
+          <Modal.Content>
+            <Translate>
+              The occurrence you chose (
+              <Param name="serializedDate" value={serializedDate} />) cannot be cancelled.
```
ThiefMaster: `name="date"` (it's the placeholder shown to translators, so no need to go into technical details not relevant for translation)

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -114,6 +118,76 @@ class BookingDetails extends React.Component {
...
+            </Translate>
+          </Modal.Content>
+          <Modal.Actions>
+            <Button content={Translate.string('Close')} onClick={() => openBookingDetails(id)} />
+            <Button
+              content={Translate.string('Cancel this occurrence')}
+              onClick={() => {
+                this.changeOccurrenceState('cancel');
+                openBookingDetails(id);
+              }}
```
ThiefMaster: It would be safer to make this function async and await the `this.changeOccurrenceState` call to make sure the details don't open until after you fully finished cancelling

PR #4334, opened 2020-03-02
Implement category stats in react

### LINKED ISSUE #4269: Implement React charts to categories statistics
Implement React charts to categories statistics.

These are the libraries which can be used:

1. React Charts - https://github.com/tannerlinsley/react-charts
2. Victory - https://github.com/FormidableLabs/victory

After some discussion, we decided to go with **React Charts**

ParthS007:
closes #4269 

Updated SS
Screenshot from 2020-04-03 14-24-39

### DIFF
--- a/indico/modules/categories/client/js/components/CategoryStatistics.jsx
+++ b/indico/modules/categories/client/js/components/CategoryStatistics.jsx
@@ -1,0 +1,195 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2020 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import categoryStatisticsURL from 'indico-url:categories.statistics_json';
+
+import moment from 'moment';
+import React, {useContext, useMemo} from 'react';
+import PropTypes from 'prop-types';
+import {Chart} from 'react-charts';
+import {Container, Header, Segment, Statistic, Message, Loader} from 'semantic-ui-react';
+import {Param, Translate} from 'indico/react/i18n';
+import {useIndicoAxios} from 'indico/react/hooks';
+import {LocaleContext} from '../context.js';
+import './CategoryStatistics.module.scss';
+
+export default function CategoryStatistics({categoryId}) {
+  const {data, loading: isLoading} = useIndicoAxios({
+    url: categoryStatisticsURL({category_id: categoryId}),
+    camelize: true,
+    trigger: categoryId,
+  });
+
+  if (isLoading) {
+    return <Loader size="massive" active />;
+  } else if (!data) {
+    return null;
+  }
+
+  return <CategoryStatisticsDisplay data={data} />;
+}
+
+CategoryStatistics.propTypes = {
+  categoryId: PropTypes.number.isRequired,
+};
+
+function getGraphData(data) {
+  const serialize = statsData =>
+    Object.entries(statsData)
+      .filter(([year]) => year <= data.maxYear + 3 && year >= data.minYear)
+      .map(([year, number]) => [moment(year, 'YYYY'), number]);
+
+  const series = {type: 'line'};
+  const tooltip = {align: 'left', anchor: 'left'};
+  const axes = [
+    {primary: true, type: 'time', position: 'bottom'},
+    {type: 'linear', position: 'left'},
+  ];
+  const eventGraph = [
+    {
+      label: Translate.string('Events'),
+      data: serialize(data.events),
+    },
+  ];
+  const contributionsGraph = [
+    {
+      label: Translate.string('Contributions'),
+      data: serialize(data.contributions),
+    },
+  ];
+
+  return {eventGraph, contributionsGraph, series, axes, tooltip};
+}
+
+function CategoryStatisticsDisplay({data}) {
+  const lang = useContext(LocaleContext);
+  const userLanguage = lang.replace('_', '-');
+  const updatedTime = moment.utc(data.updated).format('LLL');
+  const fmt = new Intl.NumberFormat(userLanguage);
+  const totalEvents = fmt.format(data.totalEvents);
+  const totalContributions = fmt.format(data.totalContributions);
+  const files = fmt.format(data.files);
+  const users = fmt.format(data.users);
+  const {eventGraph, contributionsGraph, series, axes, tooltip} = useMemo(
+    () => getGraphData(data),
+    [data]
+  );
+
+  return (
+    <Container>
+      <Segment.Group horizontal>
+        <Segment>
+          <Header as="h2">
+            <Header.Content>
+              <Translate>Number of events</Translate>
+              <Header.Subheader>
+                <Translate>The year is the one of the start date of the event.</Translate>
+              </Header.Subheader>
+            </Header.Content>
+          </Header>
+          <Segment basic>
+            {eventGraph[0].data.length === 0 ? (
+              <Message info>
+                <Translate>No data available</Translate>
+              </Message>
+            ) : (
+              <div styleName="chart-container">
+                <Chart
+                  styleName="chart"
+                  data={eventGraph}
+                  series={series}
+                  axes={axes}
+                  tooltip={tooltip}
+                />
+              </div>
+            )}
+          </Segment>
+        </Segment>
+        <Segment>
+          <Header as="h2">
+            <Header.Content>
+              <Translate>Number of contributions</Translate>
+              <Header.Subheader>
+                <Translate>The year is the one of the start date of the contribution.</Translate>
+              </Header.Subheader>
+            </Header.Content>
+          </Header>
+          <Segment basic>
+            {contributionsGraph[0].data.length === 0 ? (
+              <Message info>
+                <Translate>No data available</Translate>
+              </Message>
+            ) : (
+              <div styleName="chart-container">
+                <Chart
+                  styleName="chart"
+                  data={contributionsGraph}
+                  series={series}
+                  axes={axes}
+                  tooltip={tooltip}
+                />
+              </div>
+            )}
+          </Segment>
+        </Segment>
+      </Segment.Group>
+      <Segment.Group horizontal>
+        <Segment inverted color="blue" textAlign="center">
+          <Statistic inverted>
+            <Statistic.Value>{totalEvents}</Statistic.Value>
+            <Statistic.Label>
+              <Translate>Total number of events</Translate>
+            </Statistic.Label>
+          </Statistic>
+        </Segment>
+        <Segment inverted color="blue" textAlign="center">
+          <Statistic inverted>
+            <Statistic.Value>{totalContributions}</Statistic.Value>
+            <Statistic.Label>
+              <Translate>Total number of contributions</Translate>
+            </Statistic.Label>
+          </Statistic>
+        </Segment>
+        <Segment inverted color="blue" textAlign="center">
+          <Statistic inverted>
+            <Statistic.Value>{files}</Statistic.Value>
+            <Statistic.Label>
+              <Translate>Number of attachments</Translate>
+            </Statistic.Label>
+          </Statistic>
+        </Segment>
+        {data.users !== undefined && (
+          <Segment inverted color="blue" textAlign="center">
+            <Statistic inverted>
+              <Statistic.Value>{users}</Statistic.Value>
+              <Statistic.Label>
+                <Translate>Number of users</Translate>
+              </Statistic.Label>
+            </Statistic>
+          </Segment>
+        )}
+      </Segment.Group>
+      <Translate>
+        Last updated on: <Param name="time" value={updatedTime} />
+      </Translate>
+    </Container>
+  );
+}
+
+CategoryStatisticsDisplay.propTypes = {
+  data: PropTypes.shape({
+    contributions: PropTypes.object.isRequired,
+    events: PropTypes.object.isRequired,
+    files: PropTypes.number.isRequired,
+    updated: PropTypes.string.isRequired,
+    users: PropTypes.number,
+    totalEvents: PropTypes.number.isRequired,
+    totalContributions: PropTypes.number.isRequired,
+    minYear: PropTypes.number.isRequired,
+    maxYear: PropTypes.number.isRequired,
+  }).isRequired,
+};
--- a/indico/modules/categories/client/js/components/CategoryStatistics.jsx
+++ b/indico/modules/categories/client/js/components/CategoryStatistics.jsx
@@ -68,7 +68,10 @@ function getGraphData(data) {
 function CategoryStatisticsDisplay({data}) {
   const lang = useContext(LocaleContext);
   const userLanguage = lang.replace('_', '-');
-  const updatedTime = moment.utc(data.updated).format('LLL');
+  const updatedTime = moment
+    .utc(data.updated)
+    .local()
+    .format('LLL');
   const fmt = new Intl.NumberFormat(userLanguage);
   const totalEvents = fmt.format(data.totalEvents);
   const totalContributions = fmt.format(data.totalContributions);
--- a/indico/modules/categories/client/js/index.js
+++ b/indico/modules/categories/client/js/index.js
@@ -5,64 +5,30 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
+import React from 'react';
+import ReactDOM from 'react-dom';
+
 import './display';
+import {setMomentLocale} from 'indico/utils/date';
+import CategoryStatistics from './components/CategoryStatistics';
+import {LocaleContext} from './context.js';
 
 (function(global) {
   global.setupCategoryStats = function setupCategoryStats() {
-    $(document).ready(() => {
-      $('.plot-container .plot').each((__, element) => {
-        const $this = $(element);
-        const xMin = $this.data('min-x');
-        const xMax = $this.data('max-x');
-        const yMin = $this.data('min-y');
-        const yMax = $this.data('max-y');
-
-        var data = $this.data('values') || {};
-
-        const currentYear = new Date().getFullYear();
-        const current = [[currentYear, data[currentYear]]];
-
-        const options = {
-          axes: {
-            xaxis: {
-              label: $this.data('label-xaxis'),
-              max: xMax,
-              min: xMin,
-              tickOptions: {showGridline: false},
-            },
-            yaxis: {
-              label: $this.data('label-yaxis'),
-              max: yMax,
-              min: yMin,
-            },
-          },
-          height: 400,
-          highlighter: {
-            location: 'n',
-            tooltipAxes: 'yx',
-            tooltipSeparator: $this.data('qbubble'),
-          },
-          series: [
-            {
-              fillAlpha: 0.9,
-              fill: true,
-              fillAndStroke: true,
-            },
-            {
-              markerOptions: {
-                color: '#005272',
-                style: 'circle',
-              },
-            },
-          ],
-          width: 400,
-        };
-        data = [_.pairs(data).map(datum => [parseInt(datum[0], 10), datum[1]]), current];
-        const plot = global.$.jqplot(element.id, data, global.processJqPlotOptions(options));
-        $(window).resize(() => {
-          plot.replot({resetAxes: true});
-        });
-      });
+    document.addEventListener('DOMContentLoaded', async () => {
+      const rootElement = document.querySelector('#category-stats-root');
+      if (!rootElement) {
+        return;
+      }
+      const categoryId = parseInt(rootElement.dataset.categoryId, 10);
+      const lang = rootElement.dataset.lang;
+      setMomentLocale(lang);
+      ReactDOM.render(
+        <LocaleContext.Provider value={lang}>
+          <CategoryStatistics categoryId={categoryId} />
+        </LocaleContext.Provider>,
+        rootElement
+      );
     });
   };
 })(window);
--- a/indico/modules/categories/client/js/index.js
+++ b/indico/modules/categories/client/js/index.js
@@ -22,7 +22,7 @@ import {LocaleContext} from './context.js';
       }
       const categoryId = parseInt(rootElement.dataset.categoryId, 10);
       const lang = rootElement.dataset.lang;
-      setMomentLocale(lang);
+      await setMomentLocale(lang);
       ReactDOM.render(
         <LocaleContext.Provider value={lang}>
           <CategoryStatistics categoryId={categoryId} />
--- a/indico/modules/categories/templates/category_statistics.html
+++ b/indico/modules/categories/templates/category_statistics.html
@@ -1,4 +1,3 @@
-{% from '_statistics.html' import scatter_plot, stats_badges, stats_box %}
 {% from 'message_box.html' import message_box %}
 
 <div class="category-container">
--- a/indico/modules/categories/templates/category_statistics.html
+++ b/indico/modules/categories/templates/category_statistics.html
@@ -15,35 +15,11 @@
             </div>
         </div>
     </div>
-    {% if has_stats %}
-        <div class="category-content-wrapper category-statistics center">
-            <div class="i-box-group horz vert">
-                {% for plot in plots %}
-                    {% set title, subtitle, data = plot %}
-                    {% call stats_box(title=title, subtitle=subtitle, label=data.total,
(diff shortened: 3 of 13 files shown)


indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,147 @@
...
+import {Translate} from 'indico/react/i18n';
+
+export default function CategoryStatistics({data}) {
+  const updatedTime = moment.utc(data.updated).format('D MMM YYYY HH:mm');
+  const serialize = statsData => {
+    const serializeData = [];
+    for (const [year, number] of Object.entries(statsData)) {
+      serializeData.push([moment(year, 'YYYY'), number]);
+    }
+    return serializeData;
```
ThiefMaster: I think you can use `Object.entries(statsData.map(...)` for this

indico/modules/categories/client/js/index.js
```diff
@@ -5,64 +5,31 @@
...
-        const xMax = $this.data('max-x');
-        const yMin = $this.data('min-y');
-        const yMax = $this.data('max-y');
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+
+import './display';
+import CategoryStatistics from './components/CategoryStatistics';
 
-        var data = $this.data('values') || {};
+document.addEventListener('DOMContentLoaded', async () => {
```
ThiefMaster: I'd keep this wrapped in a `setupCategoryStats` function to avoid running it on every category page where you don't use stats at all

indico/web/client/styles/modules/_category.scss
```diff
@@ -24,6 +24,20 @@
...
+        width: 50%;
+    }
+
+    .chart {
+        width: 96%;
+        height: 350px;
+        padding: 2% 2% 0 2%;
+    }
+
+    td {
```
ThiefMaster: Could we put all this stuff in a separate `category-stats-container`? I think `category-container` is used for all the category pages.

indico/modules/categories/templates/category_statistics.html
```diff
@@ -15,35 +15,8 @@
...
-        </div>
-    {% else %}
-        <div class="category-content-wrapper category-statistics center" style="text-align: left;">
-            {% call message_box('info', fixed_width=true) %}
-                {% trans %}There are no statistics available for this category.{% endtrans %}
-            {% endcall %}
-        </div>
-    {% endif %}
+    <br><br>
+    <div id="category-content-wrapper" data-category-id="{{ category.id }}"></div>
```
ThiefMaster: What about something like `category-stats-root` for the ID here?

indico/modules/categories/client/js/index.js
```diff
@@ -5,64 +5,33 @@
...
+      const categoryId = parseInt(rootElement.dataset.categoryId, 10);
+      let response;
+      try {
+        response = await indicoAxios.get(categoryStatisticsURL({category_id: categoryId}));
+      } catch (e) {
+        handleAxiosError(e);
+        return;
+      }
+      const data = response.data;
+      ReactDOM.render(<CategoryStatistics categoryId={categoryId} data={data} />, rootElement);
```
ThiefMaster: you aren't using the categoryId prop. what about moving the data loading into the statistics component though instead of doing it here?

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,147 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Chart} from 'react-charts';
+import {Container, Header, Segment, Statistic, Label, Message} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+
+export default function CategoryStatistics({data}) {
+  const updatedTime = moment.utc(data.updated).format('D MMM YYYY HH:mm');
+  const serialize = statsData => {
+    const serializeData = Object.entries(statsData).map(([year, number]) => [
```
ThiefMaster: you can `return ...` here instead of assigning it first.

indico/modules/categories/templates/category_statistics.html
```diff
@@ -15,35 +15,12 @@
...
-            </span>
-        </div>
-    {% else %}
-        <div class="category-content-wrapper category-statistics center" style="text-align: left;">
-            {% call message_box('info', fixed_width=true) %}
-                {% trans %}There are no statistics available for this category.{% endtrans %}
-            {% endcall %}
-        </div>
-    {% endif %}
+    <br><br>
```
ThiefMaster: what about setting some padding-top on the `.category-stats-container` instead?

indico/modules/categories/util.py
```diff
@@ -71,6 +71,17 @@ def get_contribs_by_year(category_id=None):
...
 
+def get_min_year(category_id=None):
+    """Get the min year.
+
+    :param category_id: The category ID to get statistics for..
+    :return: A `Datetime Object`.
+    """
+    category_filter = Event.category_chain_overlaps(category_id) if category_id else True
+    min_year = db.session.query(db.func.min(Event.created_dt)).filter(~Event.is_deleted, category_filter).scalar()
+    return min_year
```
ThiefMaster: move the `return` to the previous statement

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,220 @@
...
+      },
+    ],
+    [data.contributions, serialize]
+  );
+  const series = React.useMemo(
+    () => ({
+      type: 'bar',
+    }),
+    []
+  );
```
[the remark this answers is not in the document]
ParthS007: I think, it's better to keep same style in all the functions as it is the only which will fit in single line.

indico/web/client/styles/modules/categories/_category_statistics.scss
```diff
@@ -0,0 +1,28 @@
...
+    width: 480px;
+    height: 300px;
+}
+
+.chart > svg {
+    width: 500px !important;
+    height: 300px !important;
+}
+
+td {
```
ThiefMaster: this is way too generic, and breaks all kinds of things (e.g. many other tables in indico such as the programme code assignment dialog)

ParthS007: working on it. :smiley:

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,211 @@
...
+
+import categoryStatisticsURL from 'indico-url:categories.statistics_json';
+
+import moment from 'moment';
+import React, {useCallback, useMemo} from 'react';
+import PropTypes from 'prop-types';
+import {Chart} from 'react-charts';
+import {Container, Header, Segment, Statistic, Message, Loader} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+
```
mic4ael: Why an empty line here?

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,211 @@
...
+}
+
+CategoryStatistics.propTypes = {
+  categoryId: PropTypes.number.isRequired,
+  lang: PropTypes.string.isRequired,
+};
+
+function CategoryStatisticsDisplay({data, lang}) {
+  const userLanguage = lang.replace('_', '-');
+  const updatedTime = moment.utc(data.updated).format('D MMM YYYY HH:mm');
```
mic4ael: This format will not take into account the current locale.. use `L` or something similar ;) and also take a look at `serializeDate`

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,211 @@
...
+  const contributionsGraph = useMemo(
+    () => [
+      {
+        label: Translate.string('Contributions'),
+        data: serialize(data.contributions),
+      },
+    ],
+    [data.contributions, serialize]
+  );
+  const series = useMemo(
```
mic4ael: Does it have to be inside this component? Can't you move it outside of the function? The same applies to `axes` and `tooltip`

indico/modules/categories/client/js/index.js
```diff
@@ -5,64 +5,22 @@
...
-          plot.replot({resetAxes: true});
-        });
-      });
+    document.addEventListener('DOMContentLoaded', async () => {
+      const rootElement = document.querySelector('#category-stats-root');
+      if (!rootElement) {
+        return;
+      }
+      const categoryId = parseInt(rootElement.dataset.categoryId, 10);
+      const lang = rootElement.dataset.lang;
```
mic4ael: this you could also just get from `Indico.User.language`. This way you will not have to put it in the `data-attributes`

indico/modules/categories/util.py
```diff
@@ -71,6 +71,16 @@ def get_contribs_by_year(category_id=None):
     return OrderedDict(query)
 
 
+def get_min_year(category_id=None):
+    """Get the min year.
+
+    :param category_id: The category ID to get statistics for..
```
mic4ael:
```suggestion
    :param category_id: The category ID to get statistics for.
```

indico/modules/categories/util.py
```diff
@@ -71,6 +71,16 @@ def get_contribs_by_year(category_id=None):
...
+    :return: A `Datetime Object`.
```
mic4ael: A `datetime` object or something similar

indico/web/client/styles/modules/categories/_category_statistics.scss
```diff
@@ -0,0 +1,28 @@
...
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+.ui.segments > .segment {
+    width: 50%;
+}
+
+.chart-container {
+    width: 480px;
```
mic4ael: Hm, it is not very flexible if we take into account resizing of the browsers etc.

indico/modules/categories/client/js/index.js
ParthS007: I have first used `Indico.User.language` but It gives an error if the user is not logged in, so that's why I used data-attribute which is available in all the templates.

indico/modules/categories/client/js/components/CategoryStatistics.jsx
ParthS007: I will give it a try to keep them outside and then calling it where it is used. What will be the difference if we keep it inside or outside?

indico/modules/categories/client/js/components/CategoryStatistics.jsx
ParthS007: Will remove it :sweat_smile:

indico/modules/categories/util.py
ParthS007: why something similar, it will always be a `Dateime object`

indico/modules/categories/client/js/components/CategoryStatistics.jsx
ThiefMaster: If something is completely static then you can indeed define it outside; that way it doesn't change during renders so no `useMemo` is needed. Since most of the values do depend on data though and thus need to be inside the component, I'm not sure how useful it is.

One option would be to create a single function outside that returns everything, and then only put something like this in the component:

    const {eventGraph, contribGrapy, ...} = useMemo(() => getGraphData(data), [...])

indico/modules/categories/util.py
ThiefMaster: it's `datetime`, not `Datetime` ;)

ThiefMaster: Also, the `Object ` part shouldn't be inside the backticks

ThiefMaster: btw, there's a `.` too much in the line before.

indico/modules/categories/client/js/components/CategoryStatistics.jsx
mic4ael: Adrian was faster than me xD yeah, my main point was this is purely static and there is no need to even call `useMemo` on every render

indico/modules/categories/client/js/index.js
mic4ael: Ahh, ok. Good point!

indico/web/client/styles/modules/categories/_category_statistics.scss
ParthS007: I understand that. If I don't specify the width and height here specifically, the SVG inside the chart will have a `0` height and not be visible. It is the requirement to pass the width and height to div containing the chart.

indico/modules/categories/util.py
mic4ael: So the return value is of type `datetime` not `Datetime` and IMO what is between '`' refers to the actual type. `Datetime Object` is not a type.

indico/web/client/styles/modules/categories/_category_statistics.scss
mic4ael: Can't you use some relative units?

indico/modules/categories/client/js/components/CategoryStatistics.jsx
ParthS007: Can we show time as well by `serializeDate`?

ParthS007: By using `LL`.

Screenshot from 2020-03-09 12-20-21

By using `L`:
Screenshot from 2020-03-09 12-21-00

we need to also show time as well.

indico/modules/categories/util.py
ParthS007: okay, got it. changed :+1:

indico/modules/categories/client/js/components/CategoryStatistics.jsx
mic4ael: Just check the momentjs docs for locale-aware formats :)

mic4ael: and yeah, you can show times with `serializeDate` as well.. it will use whatever format you pass to it ;)

indico/modules/categories/client/js/components/CategoryStatistics.jsx
ParthS007: This is the reason for `useMemo` - https://github.com/tannerlinsley/react-charts#memoize-your-props

mic4ael: Yeah, but you will achieve pretty much the same if you move the definitions of these props outside of the function. They won't change thus having the same effect as if you were using `useMemo`

indico/modules/categories/client/js/components/CategoryStatistics.jsx
ParthS007: okay, I can set the locale using [`moment.locale(userLanguage)`](https://momentjs.com/docs/#/i18n/) and then instead of formatting the date as `'D MMM YYYY HH:mm'`, I can pass just the format as `L`, `LL`, `LLL`. Now the time and date both are visible. Resolving this. :)

indico/web/client/styles/modules/categories/_category_statistics.scss
ParthS007: Done !

indico/web/client/styles/modules/categories/_category_statistics.scss
ParthS007: Opened an Issue on Github: https://github.com/tannerlinsley/react-charts/issues/58

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,193 @@
...
+
+  return <CategoryStatisticsDisplay data={data} lang={lang} />;
+}
+
+CategoryStatistics.propTypes = {
+  categoryId: PropTypes.number.isRequired,
+  lang: PropTypes.string.isRequired,
+};
+
+function getGraphData({data}) {
```
ThiefMaster: just use `data` here and when calling it; no need to wrap/unwrap

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,193 @@
...
+      data: serialize(data.contributions),
+    },
+  ];
+
+  return {eventGraph, contributionsGraph, series, axes, tooltip};
+}
+
+function CategoryStatisticsDisplay({data, lang}) {
+  const userLanguage = lang.replace('_', '-');
+  moment.locale(userLanguage);
```
ThiefMaster: You probably want to use `setMomentLocale` here.

    import {setMomentLocale} from 'indico/utils/date';

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,193 @@
...
+  data: PropTypes.shape({
+    contributions: PropTypes.object,
+    events: PropTypes.object,
+    files: PropTypes.number,
+    updated: PropTypes.string,
+    users: PropTypes.number,
+    totalEvents: PropTypes.number,
+    totalContributions: PropTypes.number,
+    minYear: PropTypes.number,
+    maxYear: PropTypes.number,
```
ThiefMaster: I think all of them could be `isRequired`

ParthS007: Apart from `users`.

indico/web/client/styles/modules/categories/_category_statistics.scss
mic4ael: Why don't you just use `.ReactChart.chart .tooltip-wrap td`?

mic4ael: And I don't think there is another way to customize the color (https://github.com/tannerlinsley/react-charts/blob/a454216a5cdbf30b401200892a52b515a14d516f/src/components/Tooltip.js#L387)

ParthS007: ` :global(.tooltip-wrap) td` worked :tada:

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,196 @@
...
+      label: Translate.string('Contributions'),
+      data: serialize(data.contributions),
+    },
+  ];
+
+  return {eventGraph, contributionsGraph, series, axes, tooltip};
+}
+
+function CategoryStatisticsDisplay({data, lang}) {
+  const userLanguage = lang.replace('_', '-');
```
ThiefMaster: not needed (check what setMomentLocale does)

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,196 @@
...
+  setMomentLocale(lang);
```
mic4ael: I would move this `setMomentLocale` to `categories/index.js` since this is configuration which should happen before mounting the component. This also actually fixes `moment.utc(data.updated).format('LLL');` since right now `updatedTime` isn't properly formatted using the current locale.

ParthS007: Sounds good to me, I will update this :+1:

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,196 @@
...
+            <Statistic inverted>
+              <Statistic.Value>{users}</Statistic.Value>
+              <Statistic.Label>
+                <Translate>Number of users</Translate>
+              </Statistic.Label>
+            </Statistic>
+          </Segment>
+        )}
+      </Segment.Group>
+      <Translate>Last updated on:</Translate> {updatedTime}
```
mic4ael:
```suggestion
      <Translate>Last updated on: <Param name="time" value={updatedTime}></Translate>
```

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,196 @@
...
+        </Segment>
+        <Segment inverted color="blue" textAlign="center">
+          <Statistic inverted>
+            <Statistic.Value>{files}</Statistic.Value>
+            <Statistic.Label>
+              <Translate>Number of attachments</Translate>
+            </Statistic.Label>
+          </Statistic>
+        </Segment>
+        {data.users && (
```
mic4ael: Shouldn't this be also displayed if `data.users == 0` or we only want to show it once `data.users > 0`?

ThiefMaster: you never have 0 users

mic4ael: Yeah, I was actually thinking about checking against `null` that is why I added this comment.

ParthS007: okay, I will add the null check.

ParthS007: After wotking on it, When we go in some subcategory, their number of users is not available and `data.users` is `undefined` and `users` is `NaN`, It satisfies `data.users !== null` and then shows `NAN` like the image below, but when we have just `data.users`, it will not pass the `NAN` value and the segment will not be shown. As @ThiefMaster mentioned, we will never have `0` users, `data.users` will never be `null`, It may be `undefined` in case of sub category.  Do we still `null` check?
Screenshot from 2020-04-01 12-24-49

ThiefMaster: then check for `undefined` instead of `null`... :)

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,196 @@
...
+import React, {useMemo} from 'react';
+import PropTypes from 'prop-types';
+import {Chart} from 'react-charts';
+import {Container, Header, Segment, Statistic, Message, Loader} from 'semantic-ui-react';
+import {Param, Translate} from 'indico/react/i18n';
+import {useIndicoAxios} from 'indico/react/hooks';
+
+import './CategoryStatistics.module.scss';
+
+export default function CategoryStatistics({categoryId, lang}) {
```
mic4ael, line 20: maybe use `React.Context` to have the `lang` value accessibile from different components instead of passing it down the tree?

ParthS007: Something like this?
 After this [line](https://github.com/indico/indico/pull/4334/files#diff-399df67992d6e77330d0e55fc8a1e072R24)
```
const localeLang = React.createContext(lang);
```
and then using like this here:
```
static lang = localeLang

// To use its value
this.context.value 
```

or I can use the `Context.Provider ` in the parent component.

What's your say on this?
@mic4ael

mic4ael: I thought more about wrapping the parent component with a context provider.

ThiefMaster: You create a context on the top level of some module: `export const LocaleContext = React.reateContext();`

Then in the components where you want to consume it: 
`const lang = useContext(LocaleContext);`

and to provide it you wrap it the topmost component in this:

`<LocaleContext.Provider value={lang}>...</LocaleContext.Provider>`

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -0,0 +1,195 @@
...
+import categoryStatisticsURL from 'indico-url:categories.statistics_json';
+
+import moment from 'moment';
+import React, {useContext, useMemo} from 'react';
+import PropTypes from 'prop-types';
+import {Chart} from 'react-charts';
+import {Container, Header, Segment, Statistic, Message, Loader} from 'semantic-ui-react';
+import {Param, Translate} from 'indico/react/i18n';
+import {useIndicoAxios} from 'indico/react/hooks';
+import {LocaleContext} from '../index.js';
```
ThiefMaster, line 17: don't import from `index.js` - either import from `../` (but I think that's quite ugly) or put it in a new `context.js` and then import it from `../context`

```js
import React from 'react';

const LocaleContext = React.createContext(null);
export default LocaleContext;
```

PR #4337, opened 2020-03-05
Add info about registration status to contributions

### LINKED ISSUE #4318: Display contributions for which none of the presenters is registered
The idea is to provide a way for an event manager to see for which contributions none of the presenters have registered for the event and provide an easy way to contact them. 

This could be done by adding a 'presenter registered' column in the List of Contributions and a filter. 'Author list' button might be enough to be able to mail them.

Leats:
- Presenters' user icons will now display their registration status in the contribution list
- Filtering contributions according to registration status of the speakers is possible
- Add column to the author list to display registration status
- Add filtering option to the author list for registration status

Closes #4318.

### DIFF
--- a/indico/modules/events/contributions/lists.py
+++ b/indico/modules/events/contributions/lists.py
@@ -16,6 +16,10 @@ from sqlalchemy.orm import joinedload, subqueryload
 
 from indico.core.db import db
 from indico.modules.events.contributions.models.contributions import Contribution
+from indico.modules.events.contributions.models.persons import ContributionPersonLink
+from indico.modules.events.models.persons import EventPerson
+from indico.modules.events.registration.models.registrations import Registration
+from indico.modules.events.registration.util import get_registered_event_persons
 from indico.modules.events.util import ListGeneratorBase
 from indico.util.i18n import _
 from indico.web.flask.templating import get_template_module
@@ -49,7 +53,9 @@ class ContributionListGenerator(ListGeneratorBase):
             ('type', {'title': _('Type'),
                       'filter_choices': OrderedDict(type_empty.items() + type_choices.items())}),
             ('status', {'title': _('Status'), 'filter_choices': {'scheduled': _('Scheduled'),
-                                                                 'unscheduled': _('Not scheduled')}})
+                                                                 'unscheduled': _('Not scheduled')}}),
+            ('speakers', {'title': _('Speakers'), 'filter_choices': {'registered': _('Registered'),
+                                                                     'not_registered': _('Not registered')}})
         ])
 
         self.list_config = self._get_config()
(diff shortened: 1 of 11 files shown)


indico/modules/events/contributions/lists.py
```diff
@@ -80,6 +87,25 @@ def _filter_list_entries(self, query, filters):
...
                 criteria.append(db.or_(*status_criteria))
 
+        if 'speakers' in filters['items']:
+            contrib_query = (db.session.query(Contribution.id)
+                             .join(ContributionPersonLink)
+                             .join(EventPerson)
+                             .join(User)
+                             .filter(Contribution.event_id == self.event.id)
+                             .filter(User.registrations.any(Registration.is_active &
+                                                            (Registration.event_id == Contribution.event_id))))
```
ThiefMaster: I'd use sqlalchemy's `and_(..)` instead of the `&` operator as soon as it needs to be wrapped on two lines. Wrapping binary operators looks a bit weird ;)

```python
                             .filter(User.registrations.any(and_(Registration.is_active,
                                                                 Registration.event_id == Contribution.event_id))))
```

(`&` is simply a shortcut for it)

indico/modules/events/contributions/templates/management/_contribution_list.html
```diff
@@ -125,7 +125,18 @@
...
+                                        {% set speaker_is_registered = speaker.person.user.id in registered_users %}
+                                        <div class="person-row">
+                                            {% set tooltip %}
+                                                {% if speaker_is_registered %}
+                                                    {%- trans %}This speaker registered for the event{% endtrans -%}
+                                                {% else %}
+                                                    {%- trans %}This speaker did not register yet{% endtrans -%}
+                                                {% endif %}
+                                            {% endset %}
+                                            <i class="icon-user js-show-registrations {{ 'exists' if speaker_is_registered }}" title="{{ tooltip }}" data-qtip-position="bottom"></i>
```
ThiefMaster: might be nicer to wrap the attributes on separate lines

indico/modules/events/registration/util.py
```diff
@@ -577,3 +581,14 @@ def import_registrations_from_csv(regform, fileobj, skip_moderation=True, notify
         })
     return [create_registration(regform, data, notify_user=notify_users, skip_moderation=skip_moderation)
             for data in todo]
+
+
+def get_registered_contribution_participants(event_id):
```
ThiefMaster: I'd pass the `event` itself as an argument even if you only needs the ID - that way it'd be easier to change things in the future if needed.

indico/modules/events/registration/util.py
```diff
@@ -577,3 +581,14 @@ def import_registrations_from_csv(regform, fileobj, skip_moderation=True, notify
...
+    """Get user ids of registered users who participate in contributions for specific event."""
```
ThiefMaster: While thinking about a different working for this docstring I just realized that the contribution joining isn't really needed here... so this whole util could probably be replaced with something like this:

```python
query = event.registrations.with_entities(Registration.user_id).filter(Registration.user_id.isnot(None)
return {id_ for id_, in query)
```

Might still be nice to keep it as a util here and name it something like `get_registered_user_ids`.

indico/modules/events/management/templates/_event_person_list.html
```diff
@@ -21,12 +24,23 @@
 {% block table_body %}
     {% for event_person, person_roles in event_persons.iteritems() -%}
         {% set is_speaker = person_roles.speaker %}
+        {% set is_registered = person_roles.no_registration == false %}
```
ThiefMaster: `not person_roles.no_registration`

indico/modules/events/contributions/lists.py
```diff
@@ -80,6 +88,25 @@ def _filter_list_entries(self, query, filters):
...
+                             .filter(Contribution.event_id == self.event.id)
+                             .filter(User.registrations.any(and_(Registration.is_active,
+                                                                 Registration.event_id == Contribution.event_id))))
+            registered_contribs = {id_ for id_, in contrib_query}
+
+            filtered_speakers = filters['items']['speakers']
+            speakers_criteria = []
+            if 'registered' in filtered_speakers:
+                speakers_criteria.append(Contribution.id.in_(registered_contribs))
+            if 'notregistered' in filtered_speakers:
```
ThiefMaster: i'd use `not_registered` here

PR #4338, opened 2020-03-06
RB: add additional overriding hooks

pferreir:
This will be needed to add a "Statistics" pane in Burotel, with an icon inside the `SidebarMenu`.

### DIFF
--- a/indico/modules/rb/client/js/components/SidebarMenu.jsx
+++ b/indico/modules/rb/client/js/components/SidebarMenu.jsx
@@ -5,6 +5,7 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
+import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import {Icon, Sidebar, Menu} from 'semantic-ui-react';
@@ -52,6 +53,7 @@ function SidebarMenu({
   visible,
   onClickOption,
   hideOptions,
+  extraOptions,
 }) {
   const options = [
     {
@@ -97,7 +99,13 @@ function SidebarMenu({
       onClick: toggleAdminOverride,
       onlyIf: isAdmin,
     },
-  ].filter(({onlyIf}) => onlyIf === undefined || onlyIf);
+  ]
+    .concat(extraOptions)
+    .filter(
+      ({onlyIf}) =>
+        onlyIf === undefined ||
+        (_.isFunction(onlyIf) ? onlyIf({isAdmin, isAdminOverrideEnabled, hasOwnedRooms}) : onlyIf)
+    );
 
   return (
     <Sidebar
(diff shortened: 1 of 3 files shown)


indico/modules/rb/client/js/components/SidebarMenu.jsx
```diff
@@ -165,6 +180,7 @@ export default connect(
     hasOwnedRooms: userSelectors.hasOwnedRooms(state),
   }),
   dispatch => ({
+    dispatchAction: dispatch,
```
pferreir: If you have a better idea on how to do this (allow plugins to dispatch actions), you're welcome to share it!

indico/web/client/js/react/util/Overridable.jsx
```diff
@@ -18,10 +18,10 @@ import {withRouter} from 'react-router-dom';
  * @param {Object|Function} extraProps - additional properties passed to the final component
  */
 export function parametrize(Component, extraProps) {
-  const ParametrizedComponent = ({...props}) => {
```
ThiefMaster: This looks like it was meant to explicitly copy the props (or maybe it was just a leftover from when we had a non-...rest prop as well?)

pferreir: I didn't find any reason to actually copy the properties over there. That's why I removed it.

indico/modules/rb/client/js/components/SidebarMenu.jsx
ThiefMaster: `const dispatch = useDispatch()` inside the plugin's component? not sure if importing stuff from react-redux in plugins works though... and i guess this type of redirection of calling hooks would be pretty nasty.

so probably your option is already the way to go there (i'd just keep the `dispatch` name and add a comment mentioning that we want to provide it to plugins so nobody removes it as being unused)

pferreir: The problem with the `dispatch` name is that along the way it got "stolen" by one of the HOCs. But I think we don't need to actually steal it (probably a remnant from some other thing).

pferreir: Added the comment too.

indico/modules/rb/client/js/components/SidebarMenu.jsx
```diff
@@ -165,6 +180,9 @@ export default connect(
     hasOwnedRooms: userSelectors.hasOwnedRooms(state),
   }),
   dispatch => ({
+    // ATTENTION: this is **intentionally** passed as a prop, despite not being used,
```
ThiefMaster: s/ATTENTION/XXX (most editors highlight FIXME, TODO and XXX in comments) ;)

pferreir: Mine doesn't 😉

pferreir: Isn't `XXX` meant for thing which is not quite ready?

ThiefMaster: looks like there might be an extension for it: microsoft/vscode#9899

ThiefMaster: I always interpreted it as "heads-up"

SO seems to agree: https://stackoverflow.com/questions/1452934/what-is-the-meaning-of-xxx-in-code-comments

pferreir: Other sources disagree, and to me it looks more like a FIXME or a TODO, so I'll leave "ATTENTION", which seems to be pretty self-explanatory. You can always configure your IDE to highlight those 😉

PR #4339, opened 2020-03-06
Move the pre-commit hook to husky

### LINKED ISSUE #3110: Mention pre-commit githook in dev setup guide
Currently `pre-commit.githook` needs to be symlinked to `.git/hooks/pre-commit`, but we could look into something more fancy such as https://pre-commit.com/ if this can handle everything we currently do.

pedroo21:
Closes #3110 

I simply added a new dependency `husky` and moved the `pre-commit.hook` file to be handled by it. This should remove the need to add a symlink setup guide and bring effortless hook integration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ git:
   depth: 500
 addons:
   postgresql: 9.6
+env:
+  - NODE_VERSION="12.16.1"
+before_install:
+  - nvm install $NODE_VERSION
 install:
   - pip install -U pip setuptools
   - pip install -r requirements.dev.txt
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Thanks! Could you test it using a fresh clone of your branch (e.g. somewhere in /tmp) if you haven't already done this? ie just do the usual `pip install -e .` (also the dev deps) and `npm install` and then try committing something that should trigger a warning.

ThiefMaster:
> Husky requires Node 10, skipping Git hooks installation.

on travis.

While husky setup will auto-skip on travis anyway (i think it checks for `CI=1` or a similar env var that's set there), I think it'd be a good opportunity to use a less outdated node version (maybe node 12) on travis.

PR #4341, opened 2020-03-07
Make isinstance calls Python 3 compatible

claudep:


ThiefMaster:
Thanks, but our current plan is to move fully to Python 3 in Indico 3.0 (and drop 2.7 at the same time).

Anyway, I guess having six there doesn't harm and makes thing easier (are there any automated tools that convert code using six to the native Python3-only versions?). If we use six, we should explicitly list it in requirements.txt though and not just rely on the fact that some other package installed it.

claudep:
OK, so you've decided to not make a version which is both Python 2 and 3 compatible?
Then are you planning to create a dedicated Python 3 branch at some point?

The advantage of using six, even in your scenario, is that you can introduce Python 3 compatibility progressively in master without having to split your code and maintain two branches in parallel during the transition period. Of course, it's a bit more work because you have to introduce six, then later remove it.

There are automated tools, but IMHO they should be considered as helpers, not as the ultimate solution.

ThiefMaster:
We'll move 2.3 to a separate branch once we start working on 3.0, so master will become the Python3-only branch then... or maybe we'll temporarily use a dev branch for the time where the code isn't in a fully working state, not sure yet.

claudep:
In that case, I guess this PR is of poor value. Sorry for the noise.
If you need help at some point for the Python 3 move, feel free to ping me.

ThiefMaster:
no problem - and contributions are always welcome, so i'll keep that in mind! :)

PR #4343, opened 2020-03-09
Fix misalignment of icons in registration scheduling

### LINKED ISSUE #4342: Schedule Registrations: Misaligned clear button
The clear button and the timezone icon are on top of each other when scheduling registrations.

Leats:
Closes #4342.
Will look like this:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -61,6 +61,7 @@ Bugfixes
 - Use a more consistent order when cloning the timetable (:issue:`4227`)
 - Do not show unrelated rooms with similar names when booking room from an
   event (:issue:`4089`)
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Looks much better now! :)

Just in case, can you check some other places where we use the same widget? Not sure if we have any where it's used in a different way...

Not sure if anything changes if it's required / not required (could also use devtools to toggle the flag), but I think either Chrome or Firefox automatically shows (or at least used to show) its *own* clear icon inside the time picker field if it did not have the `required` attribute...

Leats:
Event creation and screen dates do not have a problem with the new alignment.

Required/not required does not lead to any apparent change, everything still seems to be working.  The browser's clear icon gets shown, though, but that is the same in master branch.

PR #4346, opened 2020-03-09
Fix alignment of materials for events

### LINKED ISSUE #4344: Icon and material folders are misaligned in meetings
Regular top-level attachments look fine, but as soon as there's a folder the alignment is very weird.

image

Something important to test as well (after fixing it) is how it looks like when adding more files/folders so they get wrapped in multiple lines.

Leats:
Closes #4344.
Fixes the misalignment when only folders are present. 
Also fixes the misalignment for the second line of items.

### DIFF
--- a/indico/modules/attachments/templates/_display.html
+++ b/indico/modules/attachments/templates/_display.html
@@ -37,12 +37,12 @@
     {% set folders = item.attached_items.folders if item else folders %}
 
     <div class="attachments-display-container toolbar">
-        <div class="folder folder-root {{ 'folder-root-empty' if not files }}">
-            {% for attachment in files %}
+        {% for attachment in files %}
+            <div class="folder">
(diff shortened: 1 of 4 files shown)


Leats:
Changes the view from this:


To this:


Another picture with only folders:


If anyone has a different solution I'd be happy to hear it!

indico/modules/attachments/templates/_display.html
```diff
@@ -37,12 +37,14 @@
     {% set folders = item.attached_items.folders if item else folders %}
 
     <div class="attachments-display-container toolbar">
-        <div class="folder folder-root {{ 'folder-root-empty' if not files }}">
+        {% if files %}
```
mic4ael: I don't think it is needed. `for` loop will not render anything if `files` is an empty array.

PR #4349, opened 2020-03-11
Add warning for accepting concurrent pre-bookings

### LINKED ISSUE #4129: Room booking: show concurrent pre-booking warning when accepting a pre-booking
Currently, there is no warning when accepting a pre-booking that indicates that one or more concurrent pre-bookings will be rejected. The user can only know this information by checking the Bookings timeline.

This is even less distinct when the user has to deal with weekly/bi-weekly bookings, where they would have to check all the occurrences in the timeline.

There could be a warning in the pre-booking dialogue with links to the concurrent pre-bookings and/or a checkbox that confirms the pre-booking rejection, as suggested by a user.

Leats:
Add a modal to inform the user about the automatic rejection of concurrent bookings when accepting another one.
Closes #4129.

### DIFF
--- a/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
+++ b/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
@@ -75,6 +75,11 @@ class BookingDetails extends React.Component {
     occurrencesVisible: false,
     activeConfirmation: null,
     acceptanceFormVisible: false,
+    preBookingConflicts: {
+      warningVisible: false,
+      concurrentBookings: [],
+      acceptanceData: null,
+    },
   };
 
   showAcceptanceForm = () => {
@@ -91,6 +96,16 @@ class BookingDetails extends React.Component {
 
   hideOccurrences = () => {
     this.setState({occurrencesVisible: false});
+  };
+
+  hidePreBookingWarning = () => {
+    this.setState({
+      preBookingConflicts: {
+        warningVisible: false,
+        concurrentBookings: [],
+        acceptanceData: null,
+      },
+    });
   };
 
   renderBookedFor = bookedForUser => {
@@ -115,6 +130,69 @@ class BookingDetails extends React.Component {
           </div>
         )}
       </>
+    );
+  };
+
+  renderPreBookingWarning = () => {
+    const {bookingStateChangeInProgress} = this.props;
+    const {
+      actionInProgress,
+      preBookingConflicts: {warningVisible, concurrentBookings, acceptanceData},
+    } = this.state;
+    const items = concurrentBookings.map(booking => {
+      const reservation = booking.reservation;
+      const startDate = serializeDate(toMoment(booking.start_dt), 'L LT');
+      const endDate = serializeDate(toMoment(booking.end_dt), 'LT');
+      return (
+        <List.Item key={reservation.id}>
+          <a
+            href={bookingLinkURL({booking_id: reservation.id})}
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            {startDate} - {endDate}
+          </a>{' '}
+          <Translate>
+            by <Param name="user" value={reservation.booked_for_name} />
+          </Translate>{' '}
+          ({reservation.booking_reason})
+        </List.Item>
+      );
+    });
+    return (
+      <Modal open={warningVisible} size="small" onClose={this.hidePreBookingWarning}>
+        <Modal.Header>
+          <Translate>Warning</Translate>
+        </Modal.Header>
+        <Modal.Content>
+          <Translate>
+            There is more than one booking request during this time. Accepting this booking will
+            automatically reject the following:
+          </Translate>
+          <div styleName="concurrent-reservations">
+            <List divided styleName="concurrent-reservations-list">
+              {items}
+            </List>
+          </div>
+        </Modal.Content>
+        <Modal.Actions>
+          <Button
+            content={Translate.string('Close without accepting')}
+            onClick={this.hidePreBookingWarning}
+          />
+          <Button
+            icon="check circle"
+            onClick={() => {
+              this.changeState('approve', {...acceptanceData, force: true});
+              this.hidePreBookingWarning();
+            }}
+            loading={actionInProgress === 'approve' && bookingStateChangeInProgress}
+            disabled={bookingStateChangeInProgress}
+            content={Translate.string('Accept anyway')}
+            positive
+          />
+        </Modal.Actions>
+      </Modal>
     );
   };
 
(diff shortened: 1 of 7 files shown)


indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -118,6 +129,76 @@ class BookingDetails extends React.Component {
...
+            rel="noopener noreferrer"
+          >
+            {startDate} - {endDate}
+          </a>
+          <Translate>
+            {' '}
+            by <Param name="user" value={reservation.booked_for_name} /> {'('}
+            <Param name="info" value={reservation.booking_reason} />
+            {')'}
+          </Translate>
```
ThiefMaster: Don't put leading/trailing whitespace inside Translate tags - it's not pleasant when translating and I think we even strip it somewhere.

```suggestion
          </a>{' '}
          <Translate>
            by <Param name="user" value={reservation.booked_for_name} /> (
            <Param name="info" value={reservation.booking_reason} />)
          </Translate>
```

In principle you could even move the ` (reason)` part outside the translate block, since i don't think any language needs to change the order of name and reason:

```jsx
          </a>{' '}
          <Translate>
            by <Param name="user" value={reservation.booked_for_name} />
          </Translate>{' '}
          ({reservation.booking_reason})
```

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -118,6 +129,76 @@ class BookingDetails extends React.Component {
...
+      <Modal open={preBookingWarningVisible} size="small" onClose={this.hidePreBookingWarning}>
+        <Modal.Header>
+          <Translate>Warning</Translate>
+        </Modal.Header>
+        <Modal.Content>
+          <Translate>
+            There is more than one booking request during this time. Accepting this booking will
+            automatically reject the following:
+          </Translate>
+          {!!concurrentBookings.length && (
```
ThiefMaster: Is this check needd? I wouldn't expect this warning to show up at all unless there *are* conflicting prebookings ;)

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -75,6 +75,9 @@ class BookingDetails extends React.Component {
     occurrencesVisible: false,
     activeConfirmation: null,
     acceptanceFormVisible: false,
+    preBookingWarningVisible: false,
+    concurrentBookings: [],
+    acceptanceData: null,
```
ThiefMaster: I'd merge those into a single state field, since you usually change all of them at the same time anyway.

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -93,6 +96,14 @@ class BookingDetails extends React.Component {
     this.setState({occurrencesVisible: false});
   };
 
+  showPreBookingWarning = () => {
```
ThiefMaster: It's only used in one place so when merging the state fields you can probably just create a function that sets everything or move the logic directly to the place where you currently call this.

indico/modules/rb/controllers/backend/bookings.py
```diff
@@ -225,9 +225,15 @@ def reject(self, reason):
...
+        'reason': fields.String(required=False),
+        'force': fields.Bool(required=False)
     })
-    def accept(self, reason=None):
+    def accept(self, reason=None, force=False):
+        if not force:
+            collisions = get_prebooking_collisions(self.booking)
+            collision_data = reservation_occurrences_schema.dump(collisions)
+            if collision_data:
+                raise ExpectedError('Pre-booking collision', data=collision_data)
```
ThiefMaster: Since this error is never shown to users I wonder if it wouldn't be better to use a more machine-targetted message, like `'prebooking_collision'` (and then also check for it in the JS code instead of just checking for *an* error)

indico/modules/rb/models/reservations.py
```diff
@@ -417,8 +417,7 @@ def accept(self, user, reason=None):
         self.add_edit_log(ReservationEditLog(user_name=user.full_name, info=[log_msg]))
         notify_confirmation(self, reason)
         signals.rb.booking_state_changed.send(self)
-        valid_occurrences = self.occurrences.filter(ReservationOccurrence.is_valid).all()
-        pre_occurrences = ReservationOccurrence.find_overlapping_with(self.room, valid_occurrences, self.id).all()
+        pre_occurrences = get_prebooking_collisions(self)
         for occurrence in pre_occurrences:
             if not occurrence.is_valid:
```
ThiefMaster: This check should probably be inside `get_prebooking_collisions` - otherwise an already-cancelled prebooking occurrence would also show up as a collision when you call the util from the RH.

indico/modules/rb/controllers/backend/bookings.py
```diff
@@ -225,9 +225,15 @@ def reject(self, reason):
...
-        'reason': fields.String(required=False)
+        'reason': fields.String(required=False),
+        'force': fields.Bool(required=False)
     })
-    def accept(self, reason=None):
+    def accept(self, reason=None, force=False):
+        if not force:
+            collisions = get_prebooking_collisions(self.booking)
+            collision_data = reservation_occurrences_schema.dump(collisions)
+            if collision_data:
```
ThiefMaster: I'd check `if collisions` and only dump inside the if when you need the jsonifiable data

indico/modules/rb/util.py
```diff
@@ -321,3 +321,9 @@ def get_booking_params_for_event(event):
             'params': params,
             'time_info': time_info
         }
+
+
+def get_prebooking_collisions(reservation):
```
ThiefMaster: unit test would probably be nice for it (that way we can also guarantee that we don't end up getting already-cancelled occurrences ;))

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -448,8 +526,21 @@ class BookingDetails extends React.Component {
       booking: {id},
       actions: {changeBookingState},
     } = this.props;
+    const {acceptanceFormVisible} = this.state;
     this.setState({actionInProgress: action});
-    return changeBookingState(id, action, data).then(() => {
+    return changeBookingState(id, action, data).then(({error}) => {
+      if (error && error.message === 'prebooking_collision' && action === 'approve') {
+        if (acceptanceFormVisible) {
```
ThiefMaster: I think it'd be safe to hide it unconditionally, since the setState will simply do nothing in that case. Nothing wrong with keeping the check though.

PR #4352, opened 2020-03-12
Move the timetable logic to ellipsize entries to CSS

### LINKED ISSUE #4326: Fix timetable entry title being fully ellipsized in management if the block is small
I first thought e76b9cb610 might have caused this, but even with this commit reverted it still happens.

image

The blocks in the screenshot above are 5min each.

The up/down arrows on hover and the resizable handle also look a bit weird, but that was already the case before...

plourenco:
Closes #4326 

The timetable truncation logic has been rewritten to CSS with the following structure:

Captura de ecrã 2020-03-12, às 17 00 00
- Each timetable block contains a header and footer (.timetableBlockWrapper).
  - The header contains, respectively, the title, presenter and convener (if any).
  - The footer contains, respectively, the location and the time.
  - Header/footer wrappers are disposed of as columns and styled with `flex-wrap: wrap`. Which means that the footer (location and time) will be hidden if they do not fully fit inside the timetable block.

- Inside each wrapper, the entries will also `wrap` to display evenly when the width of the block is too small.

With `flex-wrap: wrap`      | Without
:-------------------------:|:-------------------------:
image | image

- If the header is too big to fit inside the timetable block, line wrappings will be removed to attempt to recover some spacing (handled by JS <sup>[1](#myfootnote1)</sup>) (ellipsis will also be applied):

With `white-space: nowrap`  | Without
:----------------------------:|:----------------------------:
image | image

- To conclude, if the content still doesn't fit because the timetable block is too small, we'll reduce the content to a single line (handled by JS <sup>[1](#myfootnote1)</sup>):

With `flex-wrap: nowrap`       | Without
:----------------------------:|:----------------------------:
image | image

<a name="myfootnote1">1</a>: The last two approaches need to be handled by JS because, at the time of this issue, I couldn't find any bulletproof CSS approach to prevent multi-line text from being cut with `overflow: hidden`. For future reference, this [article](http://hackingui.com/front-end/a-pure-css-solution-for-multiline-text-truncation/) includes the alternatives I considered.

### DIFF
--- a/indico/web/client/js/legacy/libs/timetable/Draw.js
+++ b/indico/web/client/js/legacy/libs/timetable/Draw.js
@@ -9,30 +9,6 @@ type(
   'TimetableBlockBase',
   [],
   {
-    truncateTitle: function(numChars, title) {
-      if (numChars >= title.length) {
-        return title;
-      }
-
-      if (numChars < 0) {
-        // Never truncate more than the length of  the string
-        if (-numChars >= title.length) {
-          return title;
-        }
-
-        // Since we add 3 periods in the end truncating 3 or less chars
-        // doen't result in shortening the string. Force it to be <= -4
-        if (numChars >= -3) {
-          numChars = -4;
-        }
-
-        numChars = title.length + numChars;
-      }
-
-      title = title.substring(0, numChars) + '...';
-      return title;
-    },
-
     openPopup: function(event) {
       var self = this;
       self.popupActive = true;
@@ -180,11 +156,12 @@ type(
         this._getTitle()
       );
 
-      this.titleWrapper = Html.div({}, this.titleDiv);
+      this.headerBlock = Html.div({className: 'timetableBlockWrapper'}, this.titleDiv);
+      this.footerBlock = Html.div({className: 'timetableBlockWrapper', style: {marginTop: 'auto'}});
 
       this.div = Html.div(
         {className: 'entry-content', style: {width: '100%', height: '100%'}},
-        this.titleWrapper
+        this.headerBlock
       );
 
       if (this.compactMode) {
@@ -235,7 +212,7 @@ type(
           if (self.eventData.presenters.length > 1) {
             this.presentersDiv.append(' ' + $T('et al.'));
           }
-          this.titleWrapper.insert(this.presentersDiv);
+          this.headerBlock.insert(this.presentersDiv);
         }
 
         // Add material menu
@@ -244,7 +221,7 @@ type(
           self.eventData.attachments &&
           self.eventData.attachments.files
         ) {
-          this.titleWrapper.insert(this.createMaterialButton(this.eventData.attachments));
+          this.headerBlock.insert(this.createMaterialButton(this.eventData.attachments));
         }
 
         //this.titleWrapper.insert(this.createManageButton());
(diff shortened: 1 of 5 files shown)


indico/web/client/js/legacy/libs/timetable/Draw.js
```diff
@@ -180,11 +156,12 @@ type(
         this._getTitle()
       );
 
-      this.titleWrapper = Html.div({}, this.titleDiv);
+      this.headerBlock = Html.div({className: 'timetableBlockWrapper'}, this.titleDiv);
+      this.footerBlock = Html.div({className: 'timetableBlockWrapper', style: { marginTop: 'auto' }});
```
mic4ael: even though this is legacy code and will probably be removed at some point I would still recommend to be consistent and maybe remove the spaces inside in the object literal

pedroo21: Agree!

indico/web/client/js/legacy/libs/timetable/Draw.js
```diff
@@ -255,9 +232,12 @@ type(
           self._formatConveners(this.eventData.conveners)
         );
 
-        this.div.append(this.convenerDiv);
-        this.div.append(this.timeDiv);
-        this.div.append(this.locationDiv);
+        if (this.eventData.conveners) {
+          this.headerBlock.append(this.convenerDiv);
+        }
+        this.footerBlock.append(this.locationDiv);
```
mic4ael: `jQuery.append` allows you to pass an array of nodes to append `this.footerBlock.append([this.locationDiv, this.timeDiv])`

indico/web/client/js/legacy/libs/timetable/Draw.js
```diff
@@ -273,7 +253,7 @@ type(
 
       this.leftPos = leftPos;
       this.width = width;
-      this.height = this.blockData.end - this.blockData.start - 3;
+      this.height = this.blockData.end - this.blockData.start - 2;
```
mic4ael: what is this mysterious `2`? :D

indico/web/client/js/legacy/libs/timetable/Draw.js
plourenco: I would like to avoid using jQuery here as much as possible since this is just plain JS.
Considering it was just two divs, I kept it as two calls, but I can always do:
`[this.locationDiv, this.timeDiv].forEach(x => this.footerBlock.append(x))`

ThiefMaster: I don't think that change would improve readability, and it's extremely legacy code anyway...

indico/web/client/js/legacy/libs/timetable/Draw.js
plourenco: You're right, it's not self-explanatory, had the same question before. This `2` is an extra padding to prevent block entries from being too close together.

I'm adding a comment referring such.

image
image

indico/web/client/js/legacy/libs/timetable/Draw.js
mic4ael: Ah, nvm.. I thought that `this.footerBlock` is a `jQuery` object which then allows to pass an array to the `append` method.

mic4ael: `append` is just a method that calls `Node.appendChild` internally.

indico/web/client/js/legacy/libs/timetable/Draw.js
mic4ael: Ok, thanks!

PR #4353, opened 2020-03-17
Add event labels

### LINKED ISSUE #3199: mark POSTPONED or CANCELED Indico events with a nice flag and notifications where appropriate
When an Indico event is  POSTPONED or CANCELLED, it would be nice to provide a relevant flag, rather than the event manager putting vulgar upper-case letters in the title. Example: 
https://indico.cern.ch/event/673053/

Also to generate notifications where appropriate, e.g. the audiovisual team, if recording were planned and the registered participants, when registration is involved.

I suggested it to Pedro who said this is a good enhancement. Thanks for your attention.

ThiefMaster:
The implementation using the SettingsProxy for storage is somewhat dirty, but I don't want to introduce new alembic revisions in patch releases (2.3 will move it to a proper DB model), especially now ones that would need to add a column to the `Event` models and thus may be rather expensive on large databases.

closes #3199

---

image

image

image

image

image

image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,12 @@ Version 2.2.7
 
 *Unreleased*
 
+Improvements
+^^^^^^^^^^^^
+
+- Add support for event labels to indicate e.g. postponed or cancelled
+  events (:issue:`3199`)
+
 Bugfixes
 ^^^^^^^^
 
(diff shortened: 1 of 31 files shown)


pferreir:
Screenshots missing :wink:

ThiefMaster:
all on mattermost so i was lazy :P

ThiefMaster:
actually, management isn't on mm...

ThiefMaster:
now there are screenshots ;)

PR #4355, opened 2020-03-18
Separate revision 'comment' into 'initial_comment' and 'final_comment'

### LINKED ISSUE #4350: Do not leave dangling comments when undoing the submitter's accept/reject of proposed changes
This will allow us to keep the comment tied to the revision, and to make it go away once we "undo" a decision.

Leats:
This allows the comment to be tied to the revision instead and delete it once the decision is undone.
Closes #4350.


ThiefMaster:
Unfortunately this turned out to be a waste of time, since it wasn't needed at all :/ See https://github.com/indico/indico/issues/4350#issuecomment-600698363 for details; sorry for not noticing that earlier...

(don't forget to `indico db downgrade` before leaving the branch)

PR #4356, opened 2020-03-18
Allow accepting after making changes

### LINKED ISSUE #4321: Paper Editing: Allow uploading files when assigning accepted state
The idea is to allow editors to upload files even when they are assigning 'accepted' state. Sometimes they provide very small changes to the papers and don't need author's confirmation

ParthS007:
closes #4321

### DIFF
--- a/indico/modules/events/editing/client/js/components/Timeline/judgment/UpdateFilesForm.jsx
+++ b/indico/modules/events/editing/client/js/components/Timeline/judgment/UpdateFilesForm.jsx
@@ -8,11 +8,11 @@
 import uploadURL from 'indico-url:event_editing.api_upload';
 
 import _ from 'lodash';
-import React from 'react';
+import React, {useState} from 'react';
 import {useDispatch, useSelector} from 'react-redux';
 import PropTypes from 'prop-types';
 import {Form as FinalForm} from 'react-final-form';
-import {Form} from 'semantic-ui-react';
+import {Form, Dropdown, Button} from 'semantic-ui-react';
 
 import {FinalSubmitButton, FinalTextArea} from 'indico/react/forms';
 import {Translate} from 'indico/react/i18n';
@@ -26,20 +26,36 @@ import FinalTagInput from './TagInput';
 
 import './JudgmentBox.module.scss';
 
+const confirmOptions = [
+  {
+    value: 'confirm',
+    text: Translate.string('Confirm'),
+  },
+  {
+    value: 'confirm_and_approve',
+    text: Translate.string('Confirm & Approve'),
+  },
+];
+
 export default function UpdateFilesForm({setLoading}) {
+  const [confirmationType, setConfirmationType] = useState('confirm');
   const lastRevision = useSelector(selectors.getLastRevision);
   const staticData = useSelector(selectors.getStaticData);
   const {eventId, contributionId, editableType} = staticData;
   const fileTypes = useSelector(selectors.getFileTypes);
   const dispatch = useDispatch();
   const files = getFilesFromRevision(fileTypes, lastRevision);
+  const option = confirmOptions.find(x => x.value === confirmationType);
 
   const submitReview = async formData => {
     setLoading(true);
     const rv = await dispatch(
       reviewEditable(lastRevision, {
         ...formData,
-        action: EditingReviewAction.update,
+        action:
+          confirmationType === 'confirm_and_approve'
+            ? EditingReviewAction.update_accept
+            : EditingReviewAction.update,
       })
     );
     if (rv.error) {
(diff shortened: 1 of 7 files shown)


indico/modules/events/editing/client/js/components/Timeline/judgment/UpdateFilesForm.jsx
```diff
@@ -78,7 +92,21 @@ export default function UpdateFilesForm({setLoading}) {
...
+                className="button icon"
+                form="judgment-form"
+                text={null}
+                button
+                floating
+                disabled={hasValidationErrors || pristine || submitting}
+                options={confirmOptions}
+                onChange={(e, data) => setConfirmationType(data.value)}
+                // eslint-disable-next-line react/jsx-no-useless-fragment
+                trigger={<Fragment />}
```
ParthS007: `trigger` - Custom element to trigger the menu to become visible. Takes place of 'text'.
I tried to supply `{null}` to avoid the use of `Fragment` but it still shows in the dropdown.

indico/modules/events/editing/client/js/components/Timeline/judgment/JudgmentBox.module.scss
```diff
@@ -32,4 +32,14 @@
...
     display: flex;
     justify-content: flex-end;
+
+    :global(.ui.buttons .button:last-child) {
+        border-top-right-radius: 0;
+        border-bottom-right-radius: 0;
+    }
+
+    :global(.ui.button.floating.dropdown.icon) {
+        border-top-right-radius: 0.28571429rem;
```
ThiefMaster: Why do we need to copy(?) this from somewhere else?

indico/modules/events/editing/client/js/components/Timeline/judgment/UpdateFilesForm.jsx
```diff
@@ -8,11 +8,11 @@
 import uploadURL from 'indico-url:event_editing.api_upload';
 
 import _ from 'lodash';
-import React from 'react';
+import React, {useState, Fragment} from 'react';
```
ThiefMaster: for fragment it's more common to write `<React.Fragment>` instead of importing `Fragment` itself

indico/modules/events/editing/client/js/components/Timeline/judgment/UpdateFilesForm.jsx
```diff
@@ -26,20 +26,34 @@ import FinalTagInput from './TagInput';
...
+
 export default function UpdateFilesForm({setLoading}) {
+  const [confirmationType, setConfirmationType] = useState('confirm');
   const {fileTypes} = useSelector(selectors.getStaticData);
   const lastRevision = useSelector(selectors.getLastRevision);
   const staticData = useSelector(selectors.getStaticData);
   const {eventId, contributionId, editableType} = staticData;
   const dispatch = useDispatch();
   const files = getFilesFromRevision(fileTypes, lastRevision);
+  const option = confirmOptions.find(x => x.value === confirmationType);
```
mic4ael: maybe do sth like

```javascript
const confirmOptions = {
  confirm: {
    value: 'confirm',
    text: Translate.string('Confirm'),
  },
  confirm_and_approve: {
    value: 'confirm_and_approve',
    text: Translate.string('Confirm & Approve'),
  },
};
```

this way you will not have to loop over the array just to get the object. Then you would also have to change `options={confirmOptions}` to `options={Object.values(confirmOptions)}`.

This is just a suggestion. Feel free to leave it like that

ThiefMaster: I don't think this is needed at all. You can just check `confirmationType` directly below instead of `option.value`

mic4ael: Yeah, true.

indico/modules/events/editing/client/js/components/Timeline/judgment/UpdateFilesForm.jsx
```diff
@@ -26,20 +26,34 @@ import FinalTagInput from './TagInput';
...
 
   const submitReview = async formData => {
     setLoading(true);
     const rv = await dispatch(
       reviewEditable(lastRevision, {
         ...formData,
         action: EditingReviewAction.update,
+        accepted_with_changes: confirmationType === 'confirm_and_approve',
```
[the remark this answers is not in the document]
ParthS007: I would go for `accepted_with_changes`.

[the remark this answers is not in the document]
ThiefMaster: to me `accepted_with_changes` sounds more like the name of a state.

PS: Another option would be to add a new `EditingReviewAction` (`update_accept` or something like that) and use that instead of the separate flag. Not sure in how many places we'd have to add logic to handle it though...

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -115,7 +115,7 @@ def _check_revision_access(self):
     @use_kwargs(ReviewEditableArgs)
     def _process(self, action, comment):
         argmap = {'tags': EditingTagsField(self.event, missing=set())}
-        if action == EditingReviewAction.update:
+        if action in [EditingReviewAction.update, EditingReviewAction.update_accept]:
```
ThiefMaster: `in (...)` instead of `in [...]`

indico/modules/events/editing/operations.py
```diff
@@ -97,15 +97,23 @@ def review_editable_revision(revision, editor, action, comment, tags, files=None
...
         EditingReviewAction.update: FinalRevisionState.needs_submitter_confirmation,
+        EditingReviewAction.update_accept: FinalRevisionState.needs_submitter_confirmation,
         EditingReviewAction.request_update: FinalRevisionState.needs_submitter_changes,
     }[action]
 
     db.session.flush()
     if action == EditingReviewAction.accept:
         publish_editable_revision(revision)
-    elif action == EditingReviewAction.update:
+    elif action in [EditingReviewAction.update, EditingReviewAction.update_accept]:
```
ThiefMaster: same here

PR #4357, opened 2020-03-18
Add event logs for registration and layout changes

### LINKED ISSUE #4360: Missing event logs for registration-related actions and layout changes
Some management actions in events fail to produce report logs. In particular, I'm missing:
- [ ] [re]opening/closing a registration form
- [ ] approving/rejecting a registration
- [ ] checking-in a registrant
- [ ] resetting check-in for a registrant
- [ ] updating the event layout

Let's simply log these actions.

OmeGak:
Closes #4360.

### DIFF
--- a/indico/modules/events/layout/controllers/layout.py
+++ b/indico/modules/events/layout/controllers/layout.py
@@ -17,12 +17,14 @@ from wtforms import fields as wtforms_fields
 from wtforms.validators import DataRequired
 
 from indico.core.db import db
+from indico.modules.events import EventLogKind, EventLogRealm
 from indico.modules.events.controllers.base import RHDisplayEventBase
 from indico.modules.events.layout import layout_settings, logger, theme_settings
 from indico.modules.events.layout.forms import (ConferenceLayoutForm, CSSForm, CSSSelectionForm,
                                                 LectureMeetingLayoutForm, LogoForm)
 from indico.modules.events.layout.util import get_css_file_data, get_css_url, get_logo_data
 from indico.modules.events.layout.views import WPLayoutEdit
+from indico.modules.events.logs.util import make_diff_log
 from indico.modules.events.management.controllers import RHManageEventBase
 from indico.modules.events.models.events import EventType
 from indico.modules.events.views import WPConferenceDisplay
@@ -74,11 +76,28 @@ class RHLayoutTimetableThemeForm(RHLayoutBase):
 
 
 class RHLayoutEdit(RHLayoutBase):
-    def _process(self):
+    def _process_request(self):
         if self.event.type_ == EventType.conference:
             return self._process_conference()
         else:
             return self._process_lecture_meeting()
+
+    def _process_GET(self):
+        return self._process_request()
+
+    def _process_POST(self):
+        old_values = layout_settings.get_all(self.event)
+        ret = self._process_request()
+        new_values = layout_settings.get_all(self.event)
+        # Skip `timetable_theme_settings` as they are dynamically generated from themes.yaml
+        changes = {k: (old_values[k], v) for k, v in new_values.iteritems()
+                   if old_values[k] != v and k != 'timetable_theme_settings'}
+        if changes:
+            form_cls = ConferenceLayoutForm if self.event.type_ == EventType.conference else LectureMeetingLayoutForm
+            form = form_cls(event=self.event)
+            self.event.log(EventLogRealm.participants, EventLogKind.change, 'Layout', summary='Layout was updated',
+                           user=session.user, data={'Changes': make_diff_log(changes, form.log_fields_metadata)})
+        return ret
 
     def _get_form_defaults(self):
         defaults = FormDefaults(**layout_settings.get_all(self.event))
--- a/indico/modules/events/layout/forms.py
+++ b/indico/modules/events/layout/forms.py
@@ -16,7 +16,7 @@ from indico.core.config import config
 from indico.modules.events.layout import theme_settings
 from indico.modules.events.layout.util import get_css_file_data, get_logo_data, get_plugin_conference_themes
 from indico.modules.users import NameFormat
-from indico.util.i18n import _
+from indico.util.i18n import _, orig_string
 from indico.web.forms.base import IndicoForm
 from indico.web.forms.fields import EditableFileField, FileField, IndicoEnumSelectField
 from indico.web.forms.validators import HiddenUnless, UsedIf
@@ -39,7 +39,31 @@ def _get_conference_theme_choices():
     return THEMES + sorted(plugin_themes, key=lambda x: x[1].lower())
 
 
-class ConferenceLayoutForm(IndicoForm):
+class LoggedLayoutForm(IndicoForm):
+    @classmethod
+    def build_field_metadata(cls, field):
+        if field.short_name == 'name_format':
+            return {'title': orig_string(field.label.text),
+                    'default': orig_string(field.none)}
+        elif field.short_name == 'theme':
+            choices = {k if k else None: orig_string(v) for k, v in field.choices}
+            return {'title': orig_string(field.label.text),
+                    'type': 'string',
+                    'convert': lambda changes: [choices[x] for x in changes]}
+        elif field.short_name == 'timetable_theme':
+            choices = {k if k else None: v for k, v in field.choices}
+            return {'title': orig_string(field.label.text),
+                    'type': 'string',
+                    'convert': lambda changes: [choices[x] for x in changes]}
+        else:
+            return orig_string(field.label.text)
+
+    @property
+    def log_fields_metadata(self):
+        return {k: self.build_field_metadata(v) for k, v in self._fields.iteritems()}
+
+
+class ConferenceLayoutForm(LoggedLayoutForm):
     is_searchable = BooleanField(_("Enable search"), widget=SwitchWidget(),
                                  description=_("Enable search within the event"))
     show_nav_bar = BooleanField(_("Show navigation bar"), widget=SwitchWidget(),
@@ -87,7 +111,7 @@ class ConferenceLayoutForm(IndicoForm):
             raise ValidationError(_('Cannot enable custom stylesheet unless there is one.'))
 
 
-class LectureMeetingLayoutForm(IndicoForm):
+class LectureMeetingLayoutForm(LoggedLayoutForm):
     name_format = IndicoEnumSelectField(_('Name format'), enum=NameFormat, none=_('Inherit from user preferences'),
                                         description=_('Format in which names are displayed'))
     timetable_theme = SelectField(_('Timetable theme'), [DataRequired()])
--- a/indico/modules/events/registration/controllers/management/regforms.py
+++ b/indico/modules/events/registration/controllers/management/regforms.py
@@ -214,12 +214,20 @@ class RHRegistrationFormOpen(RHManageRegFormBase):
     """Opens registration for a registration form"""
 
     def _process(self):
+        old_dts = (self.regform.start_dt, self.regform.end_dt)
         if self.regform.has_ended:
             self.regform.end_dt = None
         else:
             self.regform.start_dt = now_utc()
         logger.info("Registrations for %s opened by %s", self.regform, session.user)
         flash(_("Registrations for {} are now open").format(self.regform.title), 'success')
+        new_dts = (self.regform.start_dt, self.regform.end_dt)
+        if new_dts != old_dts:
+            if not old_dts[1]:
+                log_text = 'Registration form "{}" was opened'.format(self.regform.title)
+            else:
+                log_text = 'Registration form "{}" was reopened'.format(self.regform.title)
+            self.event.log(EventLogRealm.event, EventLogKind.change, 'Registration', log_text, session.user)
         return redirect(url_for('.manage_regform', self.regform))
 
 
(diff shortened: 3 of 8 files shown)


indico/modules/events/registration/logging.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def connect_log_signals():
+    signals.event.registration_checkin_updated.connect(log_registration_check_in)
+    signals.event.registration_state_updated.connect(log_registration_updated)
+
+
+def log_registration_check_in(sender, *args, **kwargs):
+    """Log a registration check-in action to the event log."""
+    registration = sender
```
ThiefMaster: just call the arg `registration`; sender is a positional arg and not a kwargs. there's also never going to be any other posarg, so `*args` is not needed

indico/modules/events/registration/logging.py
```diff
@@ -0,0 +1,56 @@
...
+        RegistrationState.rejected: 'rejected',
+    }
+    if registration.state == RegistrationState.complete:
+        log_text = 'Registration for "{}" has been approved'
+    elif registration.state == RegistrationState.rejected:
+        log_text = 'Registration for "{}" has been rejected'
+    elif registration.state == RegistrationState.pending:
+        old_state = status_descriptions.get(previous_state)
+        if not old_state:
+            old_state = str(old_state).split('.')[-1]
```
ThiefMaster: that looks quite strange to me...

indico/modules/events/layout/controllers/layout.py
```diff
@@ -74,12 +75,24 @@ def _process(self):
...
 
+    def _process_POST(self):
+        old_vals = self.event.get_settings('layout')
+        self._process_GET()
+        changes = {name: [old_vals.get(name), value, 'string']
+                   for name, value in self.event.get_settings('layout').items()
+                   if old_vals.get(name) != value}
+        changes['_diff'] = True
+        if changes:
+            log_text = "Layout was updated"
```
mic4ael: why did you assign it here to `log_text`?

indico/modules/events/registration/logging.py
OmeGak: fixed

indico/modules/events/layout/controllers/layout.py
OmeGak: This has been ported almost as-is from a plugin. I didn't change it because I think it helps understanding which text will be ultimately logged, rather than passing the string literal as a positional argument to `self.event.log(...)`.

mic4ael: Ok, then at least use consistent type of quotes and replace `"` with `'` :P

indico/modules/events/registration/logging.py
giusedb: it's a bit tricky but, the `__nonzero__` of a `TransactionStatus.pending` is evaluated to `False` while the `__str__` of it is: "TransactionStatus.pending", and `str.split('.')[-1] is "pending"

indico/modules/events/registration/logging.py
giusedb: BTW, for consistency i would like to keep that all the signal handler with the same signature, but, since this will be a "core"code, it's up to you.

ThiefMaster: I think for the `sender` it makes sense to use the name that suits the object, and not the meaningless `sender` name. `*args` would just help shadowing problems in case there are ever new posargs (which shouldn't happen, unlike new kwargs!)

indico/modules/events/layout/controllers/layout.py
OmeGak: As far as I remember, we were using double-quote strings for natural language strings so that apostrophes, common in English, don't need to be escaped. I used single-quote strings in the previous natural language strings since they contain `"` inside and I preferred instead of escaping them.

But it's been a while, so let me know what I should stick to.

ThiefMaster: mostly it's just "single quotes unless the string contains single quotes" nowadays ;) it's not *that* likely that a string suddenly gets changed to include single quotes anyway when it didn't before

indico/modules/events/registration/logging.py
OmeGak: Updated in favor of making use of `RegistrationState` titles.

```python
previous_state_title = orig_string(previous_state.title).lower()
log_text = 'Registration for "{}" has been changed from %s to pending' % previous_state_title
```

indico/modules/events/layout/controllers/layout.py
OmeGak: `'fixed'` xP

indico/modules/events/registration/logging.py
OmeGak: Kept `**kwargs` 👌

indico/modules/events/layout/controllers/layout.py
```diff
@@ -74,12 +75,24 @@ def _process(self):
...
 class RHLayoutEdit(RHLayoutBase):
-    def _process(self):
+    def _process_GET(self):
         if self.event.type_ == EventType.conference:
             return self._process_conference()
         else:
             return self._process_lecture_meeting()
 
+    def _process_POST(self):
+        old_vals = self.event.get_settings('layout')
```
ThiefMaster: That function is not part of the indico codebase nor this PR...

indico/modules/events/layout/controllers/layout.py
```diff
@@ -74,12 +75,24 @@ def _process(self):
...
+        self._process_GET()
+        changes = {name: [old_vals.get(name), value, 'string']
```
ThiefMaster: you can most likely use `old_vals[name]` here

OmeGak: It's necessary to use `.get()`. On a freshly created event, layout settings will be empty before making any change:

```python
>>> self.event.settings.filter(EventSetting.module == 'layout').all()
[]
```

indico/modules/events/layout/controllers/layout.py
OmeGak: Sorry, didn't notice that was a function coming from the plugin.

indico/modules/events/layout/controllers/layout.py
ThiefMaster: That's why the EventSettingsProxy (e.g. `layout_settings` here) should be used. It contains all the defaults.

OmeGak: Fixed 👌

indico/modules/events/layout/controllers/layout.py
OmeGak: Fixed. Now this uses `layout_settings.get_all(self.event)`.

indico/modules/events/layout/controllers/layout.py
```diff
@@ -74,12 +76,26 @@ def _process(self):
...
-    def _process(self):
+    def _process_GET(self):
         if self.event.type_ == EventType.conference:
             return self._process_conference()
         else:
             return self._process_lecture_meeting()
 
+    def _process_POST(self):
+        old_values = layout_settings.get_all(self.event)
+        ret = self._process_GET()
```
mic4ael: I think the contents of `process_GET` could be extracted to a separate method which could be called here instead of `process_GET`

indico/modules/events/registration/controllers/management/regforms.py
```diff
@@ -214,24 +214,37 @@ class RHRegistrationFormOpen(RHManageRegFormBase):
...
+        old_dts = (self.regform.start_dt, self.regform.end_dt)
         if self.regform.has_ended:
             self.regform.end_dt = None
         else:
             self.regform.start_dt = now_utc()
         logger.info("Registrations for %s opened by %s", self.regform, session.user)
         flash(_("Registrations for {} are now open").format(self.regform.title), 'success')
+        new_dts = (self.regform.start_dt, self.regform.end_dt)
+        if new_dts != old_dts:
+            if not old_dts[1]:
```
mic4ael: This way I will not have track `old_dts` somewhere else in the code to see what the first element refers to
```suggestion
            if not self.regform.end_dt:
```

indico/modules/events/registration/logging.py
```diff
@@ -0,0 +1,49 @@
...
+    from indico.modules.events.registration import logger
+    if not previous_state:
+        return
+    if registration.state == RegistrationState.complete:
+        log_text = 'Registration for "{}" has been approved'
+    elif registration.state == RegistrationState.rejected:
+        log_text = 'Registration for "{}" has been rejected'
+    elif registration.state == RegistrationState.pending:
+        previous_state_title = orig_string(previous_state.title).lower()
+        log_text = 'Registration for "{}" has been changed from %s to pending' % previous_state_title
```
mic4ael: you are mixing usage of `%` with `format` method. We usually use `format`

indico/modules/events/layout/controllers/layout.py
OmeGak: Unless there is magic happening under the hood when calling `self.process_GET`, I think this adds unnecessary indirection. Is this what you would prefer?

```python
def __process(self):
    if self.event.type_ == EventType.conference:
        return self._process_conference()
    else:
        return self._process_lecture_meeting()

def _process_GET(self):
    self.__process()

def _process_POST(self):
    old_values = layout_settings.get_all(self.event)
    ret = self.__process()
    new_values = layout_settings.get_all(self.event)
```

Up to you. Let me know what you prefer.

mic4ael: Yeah, there might no be anything happening under the hood but I still think this method should be used only internally and not overused by the classes implementing it.

OmeGak: Done 👌

indico/modules/events/registration/logging.py
```diff
@@ -0,0 +1,49 @@
...
+    signals.event.registration_state_updated.connect(log_registration_updated)
+
+
+def log_registration_check_in(registration, **kwargs):
+    """Log a registration check-in action to the event log."""
+    if registration.checked_in:
+        log_text = '"{}" has been checked in'.format(registration.full_name)
+    else:
+        log_text = '"{}" check-in has been reset'
+    registration.event.log(EventLogRealm.participants, EventLogKind.change, 'Event',
```
ThiefMaster: Shouldn't the module be 'Registration' instead of 'Event'?

indico/modules/events/layout/controllers/layout.py
```diff
@@ -74,12 +75,28 @@ def _process(self):
...
+        return self._process_request()
+
+    def _process_POST(self):
+        old_values = layout_settings.get_all(self.event)
+        ret = self._process_request()
+        new_values = layout_settings.get_all(self.event)
+        changes = {name: [old_values[name], value, 'string']
+                   for name, value in new_values.items()
+                   if old_values[name] != value}
+        changes['_diff'] = True
```
ThiefMaster: This breaks when changing the "Name format" since it tried to store an enum member in the DB. Have a look at `_log_event_update` and more important `make_diff_log` - you may want to do something similar here.

That way you'd also get nicer entries in the log message than just the raw names of the settings.

OmeGak: It took a bit to figure out but it's now done.

indico/modules/events/layout/forms.py
```diff
@@ -39,7 +39,33 @@ def _get_conference_theme_choices():
     return THEMES + sorted(plugin_themes, key=lambda x: x[1].lower())
 
 
-class ConferenceLayoutForm(IndicoForm):
+class LoggedLayoutForm(IndicoForm):
+    @classmethod
+    def build_field_metadata(cls, field):
+        if field.short_name == 'name_format':
+            return {'title': orig_string(field.label.text),
+                    'type': 'string',
```
ThiefMaster: For this particular one the enum type should be fine, shouldn't it?

OmeGak: Yep, but I had to fix the `Enum` interpretation in `make_diff_log`.

indico/modules/events/registration/logging.py
```diff
@@ -0,0 +1,49 @@
...
+        return
+    if registration.state == RegistrationState.complete:
+        log_text = 'Registration for "{}" has been approved'
+    elif registration.state == RegistrationState.rejected:
+        log_text = 'Registration for "{}" has been rejected'
+    elif registration.state == RegistrationState.pending:
+        previous_state_title = orig_string(previous_state.title).lower()
+        log_text = 'Registration for "{}" has been changed from %s to pending' % previous_state_title
+    else:
+        logger.error('Unknown registration state "%s" for registration "%s"', (registration.state, registration))
```
ThiefMaster: "not enough format params" because loggers take varargs instead of a single args for the format params

indico/modules/events/registration/logging.py
```diff
@@ -0,0 +1,49 @@
...
+    registration.event.log(EventLogRealm.participants, EventLogKind.change, 'Registration',
+                           log_text.format(registration.full_name), session.user)
+
+
+def log_registration_updated(registration, previous_state, **kwargs):
+    """Log the registration status change to the event log."""
+    from indico.modules.events.registration import logger
+    if not previous_state:
+        return
+    if registration.state == RegistrationState.complete:
```
ThiefMaster: you need to check the previous state as well. some cases that are not correctly handled atm:

- `unpaid -> complete`
- `complete -> unpaid`
- for the approved message you probably want to check `pending -> *`

indico/modules/events/registration/controllers/management/regforms.py
```diff
@@ -214,24 +214,37 @@ class RHRegistrationFormOpen(RHManageRegFormBase):
...
 
     def _process(self):
+        old_dts = (self.regform.start_dt, self.regform.end_dt)
         self.regform.end_dt = now_utc()
         if not self.regform.has_started:
             self.regform.start_dt = self.regform.end_dt
         flash(_("Registrations for {} are now closed").format(self.regform.title), 'success')
         logger.info("Registrations for %s closed by %s", self.regform, session.user)
+        new_dts = (self.regform.start_dt, self.regform.end_dt)
+        if new_dts != old_dts:
```
ThiefMaster: In which case would they be the same? I think in this RH you can unconditionally log closing the regform

OmeGak: Right! Unconditionally logging now.

PR #4358, opened 2020-03-18
Editing settings 

### LINKED ISSUE #4336: Editable setup: enabling/disabling module and editables
Create the initial enabling/disabling page of the entire module and each editable based on [this](https://codimd.web.cern.ch/_D0Aqc87RiyYk7J6cjdLtw)

panagiotappl:
This PR implements the initial view of the Editing settings based on [these notes](https://codimd.web.cern.ch/_D0Aqc87RiyYk7J6cjdLtw)
image
image
image

closes #4336

### DIFF
--- a/indico/modules/events/editing/client/js/components/EditableTypeList.jsx
+++ b/indico/modules/events/editing/client/js/components/EditableTypeList.jsx
@@ -1,0 +1,137 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2020 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import enabledEditableTypesURL from 'indico-url:event_editing.api_enabled_editable_types';
+
+import React, {useState} from 'react';
+import PropTypes from 'prop-types';
+import {Button, Checkbox, Form, Loader, Popup} from 'semantic-ui-react';
+import {Form as FinalForm} from 'react-final-form';
+import {Translate} from 'indico/react/i18n';
+import {useIndicoAxios} from 'indico/react/hooks';
+import {FinalCheckbox, FinalSubmitButton, handleSubmitError} from 'indico/react/forms';
+import {indicoAxios} from 'indico/utils/axios';
+import {editableTypeOrder, EditableTypeTitles} from '../models';
+
+import './EditableTypeList.module.scss';
+
+export default function EditableTypeList({eventId}) {
+  const [editMode, setEditMode] = useState(false);
+  const {
+    data: enabledEditableTypes,
+    loading: isLoadingEnabledEditableTypes,
+    reFetch,
+  } = useIndicoAxios({
+    url: enabledEditableTypesURL({confId: eventId}),
+    camelize: true,
+    trigger: eventId,
+  });
+
+  if (isLoadingEnabledEditableTypes) {
+    return <Loader inline="centered" active />;
+  } else if (!enabledEditableTypes) {
+    return null;
+  }
+
+  const handleSubmit = async formData => {
+    const url = enabledEditableTypesURL({confId: eventId});
+    const enabledTypes = Object.keys(formData).filter(name => formData[name]);
+    try {
+      await indicoAxios.post(url, {editable_types: enabledTypes});
+    } catch (e) {
+      return handleSubmitError(e);
+    }
+    reFetch();
+    setEditMode(false);
+  };
+
+  const renderEditMode = () => {
+    return (
+      <FinalForm
+        onSubmit={handleSubmit}
+        subscription={{submitting: true}}
+        initialValues={Object.assign(
+          ...editableTypeOrder.map(id => ({[id]: enabledEditableTypes.includes(id)}))
+        )}
+      >
+        {fprops => (
+          <>
+            <Form id="editable-type-form" onSubmit={fprops.handleSubmit}>
+              <div className="action-box">
+                {editableTypeOrder.map(id => (
+                  <div key={id} className="section">
+                    <span className="icon icon-file" />
+                    <div className="text">
+                      <div className="label">{EditableTypeTitles[id]}</div>
+                    </div>
+                    <div className="toolbar" styleName="switch-toolbar">
+                      <FinalCheckbox name={id} toggle label="" styleName="type-switch" />
+                    </div>
+                  </div>
+                ))}
+              </div>
+            </Form>
+            <div className="toolbar f-j-end">
+              <Button onClick={() => setEditMode(false)} disabled={fprops.submitting}>
+                <Translate>Cancel</Translate>
+              </Button>
+              <FinalSubmitButton form="editable-type-form" label={Translate.string('Save')} />
+            </div>
+          </>
+        )}
+      </FinalForm>
+    );
+  };
+
+  const renderListMode = () => {
+    return enabledEditableTypes.length === 0 ? (
+      <span styleName="disabled">
+        <Translate>
+          No editable types enabled yet. Click the toggle to enable any editable types.
+        </Translate>
+      </span>
+    ) : (
+      <div className="action-box">
+        {enabledEditableTypes.map(id => (
+          <div key={id} className="section">
+            <span className="icon icon-file" />
+            <div className="text">
+              <div className="label">{EditableTypeTitles[id]}</div>
+            </div>
+            <div className="toolbar">
+              <a className="i-button icon-list">
+                <Translate>List</Translate>
+              </a>
+              <a className="i-button icon-settings">
+                <Translate>Manage</Translate>
+              </a>
+            </div>
+          </div>
+        ))}
+      </div>
+    );
+  };
+
+  return (
+    <div styleName="editable-list">
+      <div styleName="header">
+        <h3>
+          <Translate>List of editable types</Translate>
+        </h3>
+        <Popup
+          content={Translate.string('Toggle editable types')}
+          trigger={<Checkbox checked={editMode} toggle onClick={() => setEditMode(!editMode)} />}
+        />
+      </div>
+      {editMode ? renderEditMode() : renderListMode()}
+    </div>
+  );
+}
+
+EditableTypeList.propTypes = {
+  eventId: PropTypes.number.isRequired,
+};
--- a/indico/modules/events/editing/client/js/models.js
+++ b/indico/modules/events/editing/client/js/models.js
@@ -5,6 +5,8 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
+import {Translate} from 'indico/react/i18n';
+
 export const FinalRevisionState = {
   none: 'none',
   replaced: 'replaced',
@@ -26,3 +28,11 @@ export const EditingReviewAction = {
   update_accept: 'update_accept',
   requestUpdate: 'request_update',
 };
+
+export const EditableTypeTitles = {
+  paper: Translate.string('Paper'),
+  slides: Translate.string('Slides'),
+  poster: Translate.string('Poster'),
+};
+
+export const editableTypeOrder = ['paper', 'slides', 'poster'];
--- a/indico/modules/events/editing/controllers/backend/management.py
+++ b/indico/modules/events/editing/controllers/backend/management.py
@@ -20,12 +20,13 @@ from indico.modules.events.editing.models.revision_files import EditingRevisionF
 from indico.modules.events.editing.models.tags import EditingTag
 from indico.modules.events.editing.operations import (create_new_file_type, create_new_tag, delete_file_type,
                                                       delete_tag, update_file_type, update_tag)
-from indico.modules.events.editing.schemas import (EditableFileTypeArgs, EditableTagArgs, EditingFileTypeSchema,
-                                                   EditingMenuItemSchema, EditingReviewConditionArgs, EditingTagSchema)
+from indico.modules.events.editing.schemas import (EditableFileTypeArgs, EditableTagArgs, EditableTypeArgs,
+                                                   EditingFileTypeSchema, EditingMenuItemSchema,
+                                                   EditingReviewConditionArgs, EditingTagSchema)
 from indico.modules.events.editing.settings import editing_settings
 from indico.util.i18n import _
 from indico.util.signals import named_objects_from_signal
-from indico.web.args import use_rh_args, use_rh_kwargs
+from indico.web.args import use_kwargs, use_rh_args, use_rh_kwargs
 
 
 class RHCreateTag(RHEditingManagementBase):
@@ -138,3 +139,14 @@ class RHMenuEntries(RHEditingManagementBase):
     def _process(self):
         menu_entries = named_objects_from_signal(signals.menu.items.send('event-editing-sidemenu', event=self.event))
         return EditingMenuItemSchema(many=True).jsonify(menu_entries.values())
+
+
+class RHEnabledEditableTypes(RHEditingManagementBase):
+    def _process_GET(self):
+        return jsonify(editing_settings.get(self.event, 'editable_types'))
+
+    @use_kwargs(EditableTypeArgs)
+    def _process_POST(self, editable_types):
+        editable_types_names = [t.name for t in editable_types]
+        editing_settings.set(self.event, 'editable_types', editable_types_names)
+        return '', 204
--- a/indico/modules/events/editing/controllers/frontend.py
+++ b/indico/modules/events/editing/controllers/frontend.py
@@ -18,10 +18,8 @@ class RHEditingDashboard(RHEditingManagementBase):
     EVENT_FEATURE = None
 
     def _process(self):
-        if not self.event.has_feature('editing'):
-            return WPEditing.render_template('management/disabled.html', self.event)
-        else:
-            return WPEditing.render_template('management/editing.html', self.event)
+        template = 'editing.html' if self.event.has_feature('editing') else 'disabled.html'
+        return WPEditing.render_template('management/{}'.format(template), self.event)
 
 
 class RHEditableTimeline(RHContributionEditableBase):
--- a/indico/modules/events/editing/controllers/frontend.py
+++ b/indico/modules/events/editing/controllers/frontend.py
@@ -15,8 +15,13 @@ from indico.modules.events.editing.views import WPEditing
 
 
 class RHEditingDashboard(RHEditingManagementBase):
+    EVENT_FEATURE = None
+
     def _process(self):
-        return WPEditing.render_template('management/editing.html', self.event)
+        if not self.event.has_feature('editing'):
+            return WPEditing.render_template('management/disabled.html', self.event)
+        else:
+            return WPEditing.render_template('management/editing.html', self.event)
 
 
 class RHEditableTimeline(RHContributionEditableBase):
--- a/indico/modules/events/editing/models/editable.py
+++ b/indico/modules/events/editing/models/editable.py
@@ -17,6 +17,7 @@ from indico.util.struct.enum import IndicoEnum
 class EditableType(int, IndicoEnum):
     paper = 1
     slides = 2
+    poster = 3
 
 
 class Editable(db.Model):
--- a/indico/modules/events/editing/schemas.py
+++ b/indico/modules/events/editing/schemas.py
@@ -17,7 +17,7 @@ from sqlalchemy import func
 from indico.core.marshmallow import mm
 from indico.modules.events.contributions.schemas import ContributionSchema
 from indico.modules.events.editing.models.comments import EditingRevisionComment
-from indico.modules.events.editing.models.editable import Editable
+from indico.modules.events.editing.models.editable import Editable, EditableType
 from indico.modules.events.editing.models.file_types import EditingFileType
 from indico.modules.events.editing.models.revision_files import EditingRevisionFile
 from indico.modules.events.editing.models.revisions import EditingRevision, InitialRevisionState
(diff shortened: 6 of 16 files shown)


indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -130,3 +130,17 @@ def _process_PATCH(self, file_types):
...
+class RHEditableTypes(RHEditingManagementBase):
+    def _process_GET(self):
+        return jsonify(editing_settings.get(self.event, 'editable_types').items())
+
+    @use_rh_kwargs(EditableTypeArgs)
+    def _process_POST(self, editable_types):
+        new_editable_types = editing_settings.get(self.event, 'editable_types')
+        enabled = [t.name for t in editable_types]
+        for name, editable in new_editable_types.items():
+            editable['enabled'] = name in enabled
```
ThiefMaster: wouldn't it make more sense to store a dict mapping type names to bool?

indico/modules/events/editing/settings.py
```diff
@@ -14,7 +14,13 @@
 
 
 editing_settings = EventSettingsProxy('editing', {
-    'review_conditions': OrderedDict()
+    'review_conditions': OrderedDict(),
+    'editable_types': OrderedDict([
```
ThiefMaster: Do we want to allow changing the title? If not, I'd just map type name to enabled bool (using a normal dict)

indico/modules/events/editing/schemas.py
```diff
@@ -244,3 +244,10 @@ def _validate_file_types(self, file_types):
         event_conditions = editing_settings.get(event, 'review_conditions')
         if any(condition_types == set(event_condition) for event_condition in event_conditions.values()):
             raise ValidationError(_('Conditions have to be unique'))
+
+
+class EditableTypeArgs(mm.Schema):
+    class Meta:
+        rh_context = ()
```
ThiefMaster: you can completely omit this

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,126 @@
...
+      reFetch();
+    } catch (e) {
+      return handleSubmitError(e);
+    }
+    setEditMode(false);
+  };
+
+  return (
+    <div styleName="editable-list">
+      <h3>List of editable types</h3>
```
ThiefMaster: i18n

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,126 @@
...
+      <FinalForm
+        onSubmit={handleSubmit}
+        subscription={{submitting: true}}
+        initialValues={Object.assign(...editableTypes.map(([id, {enabled}]) => ({[id]: enabled})))}
+      >
+        {fprops => (
+          <>
+            <Form id="editable-type-form" onSubmit={fprops.handleSubmit}>
+              <div className="action-box">
+                {!editMode &&
```
ThiefMaster: I'd keep the two modes more separate, i.e. render the FinalForm and the form content itself only while in edit mode, and the normal display version only while outside edit mode. The duplication would be minimal (some contaienr elements), but I think it would greatly improve readability.

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,126 @@
...
+                  editableTypes.map(([id, {title, enabled}]) => {
+                    return enabled ? (
+                      <div key={id} className="section">
+                        <span className="icon icon-file" />
+                        <div className="text">
+                          <div className="label">
+                            <Param name="title" value={title} />
```
ThiefMaster: there's no i18n here so no need for `Param`

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,126 @@
...
+                          </div>
+                        </div>
+                        <div className="toolbar">
+                          <a className="i-button icon-list">
```
ThiefMaster: Shouldn't we use SUI buttons here since it'll look inconsistent with the buttons in edit view otherwise?

indico/modules/events/editing/models/editable.py
```diff
@@ -17,6 +17,7 @@
 class EditableType(int, IndicoEnum):
     paper = 1
     slides = 2
+    poster = 3
```
ThiefMaster: This needs to go along with an alembic revision that updates the check constraint on the columns using this enum.

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,126 @@
...
+  } else if (!editableTypes) {
+    return null;
+  }
+
+  const handleSubmit = async formData => {
+    const url = editableTypesURL({confId: eventId});
+    const enabledTypes = Object.keys(formData).filter(name => formData[name]);
+    try {
+      await indicoAxios.post(url, {editable_types: enabledTypes});
+      reFetch();
```
ThiefMaster: this should be after the try..catch

indico/modules/events/editing/settings.py
panagiotappl: We probably don't want to allow changing the title. Where should I store it, if not there?

ThiefMaster: Just a mapping in the JS i'd say...

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
panagiotappl: Since we are already using action-boxes in this page I think it's not too bad to use the old buttons since they integrate well with all the old css used by the boxes. Unless we are planning to re-do the entire settings page in semantic-ui style, which I don't think is the case

ThiefMaster: ok, makes sense. i don't think we want to spend on redoing that now...

indico/migrations/versions/20200319_1605_56a26a721717_.py
```diff
@@ -0,0 +1,29 @@
...
+revision = '56a26a721717'
+down_revision = '18a1088f1ea8'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute('''
+        ALTER TABLE "event_editing"."editables" DROP CONSTRAINT "ck_editables_valid_enum_type";
+        ALTER TABLE "event_editing"."editables" ADD CONSTRAINT "ck_editables_valid_enum_type" CHECK ((type = ANY (ARRAY[1, 2, 3])));
```
mic4ael: Maybe wrap this like to make it more readable (it is a bit too long.

```sql
         ALTER TABLE event_editing.editables
         ADD CONSTRAINT "ck_editables_valid_enum_type"
         CHECK ((type = ANY (ARRAY[1, 2, 3])));
```

indico/modules/events/editing/controllers/frontend.py
```diff
@@ -15,8 +15,13 @@
...
 
 class RHEditingDashboard(RHEditingManagementBase):
+    EVENT_FEATURE = None
+
     def _process(self):
-        return WPEditing.render_template('management/editing.html', self.event)
+        if not self.event.has_feature('editing'):
+            return WPEditing.render_template('management/disabled.html', self.event)
+        else:
+            return WPEditing.render_template('management/editing.html', self.event)
```
mic4ael: Maybe sth like this

```suggestion
        template = 'disabled.html' if not self.event.has_feature('editing') else 'editing.html'
        return WPEditing.render_template('management/{}',format(template), self.event)
```

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,140 @@
...
+    );
+  };
+
+  return (
+    <div styleName="editable-list">
+      <h3>
+        <Translate>List of editable types</Translate>
+      </h3>
+      {editMode && renderEditMode()}
+      {!editMode && renderListMode()}
```
mic4ael:
```suggestion
      {editMode ? renderEditMode() : renderListMode()}
```

indico/migrations/versions/20200319_1605_56a26a721717_.py
ThiefMaster: I really wouldn't bother with wrapping those raw SQL statements. IMHO it doesn't even help readability (not that you really *want* to read them).

mic4ael: I do read them :P and for example with GH view it is hard to see the whole statement

ThiefMaster: fair enough, i usually just use dbdiff to check that the revision works, and is in sync with the models (which are much more readable)..

anyway, i still think 1 statement per line for *bulk* raw sql (especially in cases where you have maybe 20+ lines of generated sql) is easier.

fwiw, in this particular case (one constraint) using the standard alembic operations to create/drop constraints wouldn't be so bad (i didn't look at how much sql we have here when writing my last comment about it being a waste of time)

panagiotappl: I personally didn't find those particular two lines unreadable. I think it's kind of a matter of preference in this case as I don't prefer so much wrapping as posted in the comment.

mic4ael: To me, as a reviewer, it was a bit unreadable but in the end I am not against leaving it as it is.

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,152 @@
...
+    trigger: eventId,
+  });
+
+  if (isLoadingEditableTypes) {
+    return <Loader inline="centered" active />;
+  } else if (!editableTypes) {
+    return null;
+  }
+
+  const handleSubmit = async formData => {
```
plourenco: Perhaps wrap this inside `useCallback()`. Even though there are no dependencies, every time a re-render happens this will be referentially different.

ThiefMaster: I don't think this is a problem for onSubmit, since that function is only relevant during submission so the function changing should not cause anything to refresh unnecessarily. So using useCallback with a dependency list just adds more complexity to each component using it...

But if you noticed strange behavior or saw something about it in the docs or code of the library, I'd be interested in hearing about it for sure!

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,152 @@
...
+                        label=""
+                        styleName="type-switch"
+                      />
+                    </div>
+                  </div>
+                ))}
+              </div>
+            </Form>
+            <div className="toolbar f-j-end">
+              {editMode && (
```
plourenco: If the whole `renderEditMode` depends on `editMode` in line 144.
```
{editMode && renderEditMode()}
```
Is there any other reason why we need to check it again here?

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
plourenco: I agree, the impact shouldn't be relevant, just checked it.
However, we should keep an eye for other cases, because the lack of `useCallback` will cause it to be referentially different and affect the props change within FinalForm whenever a submit is done. Final form uses deep equality. Imo, it's avoidable.

```
// FinalForm.js
function useWhenValueChanges(value, callback, isEqual) {
  if (isEqual === void 0) {
    isEqual = function isEqual(a, b) {
      return a === b;
    };
  }
...
```

ThiefMaster: let's see what the developer of the library @erikras says ;)

https://twitter.com/ThiefMaster/status/1241046930303713282

plourenco: Great idea! 😄 That can be really useful for future cases

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,144 @@
...
+      </FinalForm>
+    );
+  };
+
+  const renderListMode = () => {
+    const hasEnabled = editableTypes.some(([_, enabled]) => enabled);
+    return (
+      <>
+        {!hasEnabled && (
+          <span styleName="no-enabled">
```
mic4ael:
```suggestion
          <span styleName="disabled">
```
?

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,144 @@
...
+            <Translate>
+              No editable types enabled yet. Click the toggle to enable any editable types.
+            </Translate>
+          </span>
+        )}
+        {hasEnabled && (
+          <div className="action-box">
+            {editableTypes.map(([id, enabled]) => {
+              return enabled ? (
```
mic4ael: Maybe just return early?

```javascript
if (!enabled) {
  return null;
}
return (
  <div>
    ...
```

indico/modules/events/editing/settings.py
```diff
@@ -14,7 +14,9 @@
 
 
 editing_settings = EventSettingsProxy('editing', {
-    'review_conditions': OrderedDict()
+    'review_conditions': OrderedDict(),
+    'editable_types': OrderedDict({'paper': True, 'slides': False, 'poster': False}),
```
ThiefMaster: I don't think we need an OrderedDict here - especially since you lose the ordering by initializing it from a regular dict!

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -138,3 +139,17 @@ class RHMenuEntries(RHEditingManagementBase):
     def _process(self):
         menu_entries = named_objects_from_signal(signals.menu.items.send('event-editing-sidemenu', event=self.event))
         return EditingMenuItemSchema(many=True).jsonify(menu_entries.values())
+
+
+class RHEditableTypes(RHEditingManagementBase):
+    def _process_GET(self):
+        return jsonify(editing_settings.get(self.event, 'editable_types').items())
+
+    @use_kwargs(EditableTypeArgs)
```
ThiefMaster: No strong opinion, but since this schema is pretty simple you could also inline it:

```python
@use_kwargs({
    'editable_types': fields.List(EnumField(EditableType), required=True)
})
```

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -138,3 +139,17 @@ class RHMenuEntries(RHEditingManagementBase):
...
+    def _process_POST(self, editable_types):
+        new_editable_types = editing_settings.get(self.event, 'editable_types')
+        enabled = [t.name for t in editable_types]
+        for key in new_editable_types:
+            new_editable_types[key] = key in enabled
+        editing_settings.set(self.event, 'editable_types', new_editable_types)
```
ThiefMaster: Can't this whole block be replaced with this?

```python
editing_settings.set(self.event, 'editable_types', editable_types)
```

or if you want to keep a dict instead of storing a list of enabled types in the settings:

```python
types_config = {t.name: t.name in editable_types for t in EditableType}
editing_settings.set(self.event, 'editable_types', types_config)
```

indico/modules/events/editing/client/js/models.js
```diff
@@ -25,3 +25,9 @@ export const EditingReviewAction = {
   update: 'update',
   requestUpdate: 'request_update',
 };
+
+export const EditingType = {
+  paper: 'Paper',
+  slides: 'Slides',
+  poster: 'Poster',
+};
```
ThiefMaster: My initial suggestion about replicating the enum was for cases where you need to check the internal name (ie the lowercased version). If you want a mapping for the user-facing name fine, but then it should be named e.g. `EditableTypeTitles` and use `Translate.string()`

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,148 @@
...
+  };
+
+  return (
+    <div styleName="editable-list">
+      <div styleName="header">
+        <h3>
+          <Translate>List of editable types</Translate>
+        </h3>
+        <Popup
+          content="Toggle editable types"
```
ThiefMaster: i18n

indico/modules/events/editing/settings.py
panagiotappl: If the OrderedDict is initialized properly here why shouldn't we keep it?

ThiefMaster: what's the benefit in maintaining an ordered structure for something's that just a collection of "enabled/disabled" flags?

panagiotappl: So that they appear in the same order in all events, and they don't arbitrarily change when saving the editable types.

ThiefMaster: I'd prefer to have the order hardcoded on the client side... using the the list of enabled/disabled types for this feels very hacky.

panagiotappl: Isn't hardcoding it on the client side hacky as well? What do the others think: @indico/development-team

ThiefMaster: The display order sounds like something very much client-related to me...

mic4ael: Yeah, first of all, you should initialize the `OrderedDict` with a list of items to preserve the order. However, I think that the way wee store that data shouldn't matter that much and we can easily sort it on the client side.

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,148 @@
...
+    reFetch();
+    setEditMode(false);
+  };
+
+  const renderEditMode = () => {
+    return (
+      <FinalForm
+        onSubmit={handleSubmit}
+        subscription={{submitting: true}}
+        initialValues={Object.assign(...editableTypes.map(([id, enabled]) => ({[id]: enabled})))}
```
plourenco:
```suggestion
        initialValues={editableTypes.reduce((acc, [id, enabled]) => {
          acc[id] = enabled;
          return acc;
        }, {})}
```

Imo, this is more readable, but since spread has been widely used across other components and the latter is smaller, I'll leave this up to your criteria.

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -138,3 +139,14 @@ class RHMenuEntries(RHEditingManagementBase):
     def _process(self):
         menu_entries = named_objects_from_signal(signals.menu.items.send('event-editing-sidemenu', event=self.event))
         return EditingMenuItemSchema(many=True).jsonify(menu_entries.values())
+
+
+class RHEditableTypes(RHEditingManagementBase):
```
ThiefMaster: `RHEnabledEditableTypes` maybe?

indico/modules/events/editing/client/js/models.js
```diff
@@ -25,3 +27,11 @@ export const EditingReviewAction = {
...
   requestUpdate: 'request_update',
 };
+
+export const EditableTypeTitles = {
+  paper: Translate.string('Paper'),
+  slides: Translate.string('Slides'),
+  poster: Translate.string('Poster'),
+};
+
+export const EditableTypeOrder = ['paper', 'slides', 'poster'];
```
ThiefMaster: `editableTypeOrder`? (the other names are using UpperCase to match the enums, but this one doesn't match an enum)

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,145 @@
...
+              <FinalSubmitButton form="editable-type-form" label={Translate.string('Save')} />
+            </div>
+          </>
+        )}
+      </FinalForm>
+    );
+  };
+
+  const renderListMode = () => {
+    const hasEnabled = editableTypes.some(([_, enabled]) => enabled);
```
ThiefMaster: `editableTypes.length > 0`?

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,145 @@
...
+    return (
+      <>
+        {!hasEnabled && (
```
ThiefMaster: `{!hasEnabled ? ... : ...}`?

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,144 @@
...
+                {editableTypeOrder.map(id => (
+                  <div key={id} className="section">
+                    <span className="icon icon-file" />
+                    <div className="text">
+                      <div className="label">{EditableTypeTitles[id]}</div>
+                    </div>
+                    <div className="toolbar" styleName="switch-toolbar">
+                      <FinalCheckbox
+                        name={id}
+                        as={Checkbox}
```
ThiefMaster: I don't think you need this; it's set by default in `CheckboxAdapter`

indico/modules/events/editing/client/js/components/EditableTypeList.jsx
```diff
@@ -0,0 +1,144 @@
...
+            </div>
+          </>
+        )}
+      </FinalForm>
+    );
+  };
+
+  const renderListMode = () => {
+    const hasEnabled = enabledEditableTypes.length > 0;
+    return !hasEnabled ? (
```
ThiefMaster: return enabledEditableTypes.length === 0 ? ...

PR #4359, opened 2020-03-19
Add category navigator in category display pages

### LINKED ISSUE #4282: Better category navigation from category display pages
Navigating to nested categories from the home pages requires several page loads. This feature is proposed to provide an easier mechanism to navigate to categories 

Reusing the category navigation widget, we can provide a quick way to find and jump to a category. I'm proposing to add a `Navigate` button on the category display page that opens a category navigation widget dialog. The configured action for selecting a category in the widget is redirecting the browser to the display page of that category.

This approach would also help navigating to very nested categories https://github.com/indico/indico/issues/3727. A drawaback of a "deep" tree view is that for instances with many categories this list will be very long. Power users will be comfortable with Ctrl+F, but others will likely end up browsing by scrolling down. This would probably require using some sort of real time filtering, but the category navigator widget already comes with it.

**Additional context**
I've prototyped this feature in https://github.com/UNOG-Indico/indico-core/commit/1f4d740163dca2a29719122df3d05d57a44c4f70.

`Navigate` button:
Screen Shot 2020-02-06 at 17 23 36

Configured category navigation widget:
Screen Shot 2020-02-06 at 17 24 05

OmeGak:
Closes #4282.


indico/modules/categories/templates/display/base.html
```diff
@@ -86,6 +91,19 @@ <h1 class="category-title {{ category_title_classes }}">
     </div>
 </div>
 <script>
+    $('#navigate-button').on('click', function(evt){
```
mic4ael:
```suggestion
    $('#navigate-button').on('click', function(evt) {
```

indico/modules/categories/templates/display/base.html
```diff
@@ -86,6 +91,19 @@ <h1 class="category-title {{ category_title_classes }}">
...
+        evt.preventDefault();
+        $('<div>').categorynavigator({
+            openInDialog: true,
+            category: "{{ category.id }}",
```
mic4ael: I guess it shouldn't matter match, but the initial value for `category` in `categorynavigator` widget is a number. Here you are passing a string which is a bit inconsistent

indico/modules/categories/templates/display/base.html
OmeGak: Applied.

indico/modules/categories/templates/display/base.html
OmeGak: True. Thanks. Just changed it.

indico/modules/categories/templates/display/base.html
```diff
@@ -86,6 +91,19 @@ <h1 class="category-title {{ category_title_classes }}">
...
+    $('#navigate-button').on('click', function(evt) {
+        evt.preventDefault();
+        $('<div>').categorynavigator({
+            openInDialog: true,
+            category: {{ category.id }},
+            actionButtonText: $T.gettext('Navigate to'),
+            dialogTitle: $T.gettext('Navigate to a category'),
+            onAction: function(category) {
+                var url = build_url(Indico.Urls.Categories.display, {category_id: category.id});
+                window.location.href = url;
```
ThiefMaster: `window.location.href = build_url(...)` isn't much longer ;)

(if you want to keep it, use `const` instead of `var`)

OmeGak: Changed.

mic4ael:
Is it normal that so many lines have been deleted from selection.json?

ThiefMaster:
icomoon is now generating a minified/compact version of the json

mic4ael:
> icomoon is now generating a minified/compact version of the json

Mhm, ok. Good to know

indico/modules/categories/templates/display/base.html
```diff
@@ -22,6 +22,11 @@ <h1 class="category-title {{ category_title_classes }}">
             <div class="group">
                 {{ create_event_button(category, classes="highlight", text=_("Create event"), with_tooltip=false) }}
             </div>
+            <div class="group">
+                <div id="navigate-button" class="i-button icon-compass2">
+                    {% trans %}Navigate categories{% endtrans %}
```
ThiefMaster: Just wondering.. do we need the text at all?

image

or at least shorter text?

image


or maybe a button in the bar next to the parent one?

image
image

(i prefer the last option tbh)

```diff
diff --git a/indico/modules/categories/templates/display/base.html b/indico/modules/categories/templates/display/base.html
index 5d92c13a8b..120f61148b 100644
--- a/indico/modules/categories/templates/display/base.html
+++ b/indico/modules/categories/templates/display/base.html
@@ -22,17 +22,14 @@
             <div class="group">
                 {{ create_event_button(category, classes="highlight", text=_("Create event"), with_tooltip=false) }}
             </div>
-            <div class="group">
-                <div id="navigate-button" class="i-button icon-compass2">
-                    {% trans %}Navigate categories{% endtrans %}
-                </div>
-            </div>
             <div class="group">
                 {% if not category.is_root %}
                     <a class="i-button icon-arrow-up" href="{{ category.parent.url }}">
                         {% trans %}Parent category{% endtrans %}
                     </a>
                 {% endif %}
+                <button id="navigate-button" class="i-button icon-only icon-compass2"
+                        title="{% trans %}Navigate categories{% endtrans %}"></button>
                 <a href="#"
                    class="i-button icon-calendar arrow js-export-ical"
                    title="{% trans %}Export to scheduling tool{% endtrans %}"
```

OmeGak: This is a matter of ux affordances, discoverability and how much attention we want to draw towards this feature. 

While the icon is nonintrusive, it doesn't communicate that it's now possible to navigate categories. The button with the *Navigate* text does a better job at drawing attention, but it doesn't communicate what exactly can be navigated. 

The text could be avoided if the button was located somewhere that is intimately related with category navigation (like right above list of subcategories or next to breadcrumbs), but since it's not, I prefer to keep the action explicit.

At the same time, I don't have a particularly strong feeling about this. I would, at the very least, leave the button with the *Navigate* text on it. Let me know what you think.

ThiefMaster: Shorter text plus more verbose tooltip? (i'd still move it in the existing group though)

Leats: While I feel the icon without text looks the best, I feel like that will get easily overlooked. Keeping the text short (`Navigate`) sounds like the best compromise. And I think a more verbose tooltip would be good!

panagiotappl: I agree with @Leats even though I don't see a huge difference in having just the icon.

ThiefMaster: let's go for that option then.. seems like a good middle ground between a large button and just the icon

OmeGak: - Button text changed to `Navigate`.
- Tooltip `Navigate categories` added.

Regarding the grouping, I think it makes more sense to keep the button in a separate group.
|a|b|
|:-|:-|
|Screen Shot 2020-03-25 at 17 14 43|Screen Shot 2020-03-25 at 17 15 04| 

In `a`, each group has a clearly different semantic: 1) create stuff, 2) navigate *all* categories, 3) do stuff in *this* category. In `b`, the semantics get mixed up.

CHANGES.rst
```diff
@@ -58,6 +58,8 @@ Improvements
 - Add event logging to opening/closing registration forms, approval/rejection of
   registrations, and updates to event layout (:issue:`4360`,
   thanks :user:`giusedb` & :user:`omegak`)
+- Add category navigation dialog from category display page (:issue:`4282`,
+  thanks, :user:`omegak`)
```
ThiefMaster: no comm after the "thanks"

indico/modules/categories/templates/display/base.html
```diff
@@ -86,6 +92,18 @@ <h1 class="category-title {{ category_title_classes }}">
     </div>
 </div>
 <script>
+    $('#navigate-button').on('click', function(evt) {
+        evt.preventDefault();
```
ThiefMaster: not needed if you convert it to a `<button>`

PR #4363, opened 2020-03-19
Expose user blocking in the admin UI

### LINKED ISSUE #3243: Expose user blocking in the admin UI
There's `User.is_blocked` which is properly implemented in the backend but currently only settable via indico shell.

Ideally, the UI should:

- Let admins block/unblock users
- Show that a user account is blocked when viewing their profile and maybe also in the user search results
- Filter out blocked users when searching in non-admin areas (?)

---

Since we don't have any amazing place to put admin-level stuff for specific users I would either:

- Put it on the "Accounts" tab of the user settings. That's the least weird place right now.
- Add a new "Admin" tab, but with just block/unblock it might look a bit naked

The actual blocking/unblocking would be done in a new RH (similar to what `RHAPIBlockKey` does for blocking/unblocking API keys), but the action should be logged as well (not sure why we don't do that for API keys)

For the information in the profile I would show a message like we currently do for the system user (base.html in the users module). The search results can simply use an icon.

plourenco:
Closes #3243

This PR will expose the endpoint and UI ability to block/unblock users, and:
- [x] Show that a user account is blocked ~~when viewing their profile~~ and maybe also in the user search results
- [x] *Filter out blocked users when searching in non-admin areas* - RFC

image
---
image
---
image

### Side-effects

Since this commit used some semantic-ui raw elements, I took the opportunity to upgrade most of the previous custom CSS within the user profile section. These include, but not limited to, toggles, buttons and header panels (alignment refactoring only).

image

### DIFF
--- a/indico/modules/api/templates/user_profile.html
+++ b/indico/modules/api/templates/user_profile.html
@@ -12,15 +12,17 @@
                 {% if key and can_modify %}
                     <div class="i-box-buttons toolbar">
                         <div class="group">
-                            <button type="button" class="i-button warning icon-plus" data-href="{{ url_for('.key_create') }}" data-method="post"
+                            <button type="button" class="ui orange button" data-href="{{ url_for('.key_create') }}" data-method="post"
                                     data-params='{"force": "1"}'
                                     data-title="{% trans %}Create new API key{% endtrans %}"
                                     data-confirm="{% trans %}Do you really want to create a new API key? Your old key will stop working immediately!{% endtrans %}">
+                                <i aria-hidden="true" class="icon plus"></i>
                                 {%- trans %}Create new API key{% endtrans -%}
                             </button>
-                            <button type="button" class="i-button danger icon-remove" data-href="{{ url_for('.key_delete') }}" data-method="post"
+                            <button type="button" class="ui negative button" data-href="{{ url_for('.key_delete') }}" data-method="post"
                                     data-title="{% trans %}Delete API key{% endtrans %}"
                                     data-confirm="{% trans %}Do you really want to delete your API key?{% endtrans %}">
+                                <i aria-hidden="true" class="icon trash alternate"></i>
                                 {%- trans %}Delete API key{% endtrans -%}
                             </button>
                         </div>
@@ -36,7 +38,7 @@
                         {%- endtrans -%}
                     </p>
 
-                    <button type="button" class="i-button highlight" data-href="{{ url_for('.key_create') }}" data-method="post">
+                    <button type="button" class="ui primary button" data-href="{{ url_for('.key_create') }}" data-method="post">
                         {%- trans %}Create API key{% endtrans -%}
                     </button>
                 {% else %}
@@ -99,10 +101,11 @@
                 <div class="i-box-buttons toolbar">
                     <div class="group">
                         <button type="button"
-                                class="i-button warning icon-remove"
+                                class="ui warning button"
                                 data-href="{{ url_for('core.reset_signature_tokens') }}"
                                 data-method="POST"
                                 data-confirm="{% trans %}Are you sure you'd like to invalidate all your token-based links?{% endtrans %}">
+                            <i aria-hidden="true" class="icon trash alternate"></i>
                             {%- trans %}Invalidate all links{% endtrans -%}
                         </button>
                     </div>
@@ -128,13 +131,14 @@
                         <div class="i-box-buttons toolbar">
                             <div class="group">
                                 {% if key.is_blocked %}
-                                    <button type="button" class="i-button accept" data-href="{{ url_for('.key_block') }}"
+                                    <button type="button" class="ui positive button" data-href="{{ url_for('.key_block') }}"
                                             data-method="post">
                                         {%- trans %}Unblock API key{% endtrans -%}
                                     </button>
                                 {% else %}
-                                    <button type="button" class="i-button warning icon-disable" data-href="{{ url_for('.key_block') }}"
+                                    <button type="button" class="ui negative button" data-href="{{ url_for('.key_block') }}"
                                             data-method="post">
+                                        <i aria-hidden="true" class="icon ban"></i>
                                         {%- trans %}Block API key{% endtrans -%}
                                     </button>
                                 {% endif %}
--- a/indico/modules/api/templates/user_profile.html
+++ b/indico/modules/api/templates/user_profile.html
@@ -12,14 +12,21 @@
                 {% if key and can_modify %}
                     <div class="i-box-buttons toolbar">
                         <div class="group">
-                            <button type="button" class="ui orange button" data-href="{{ url_for('.key_create') }}" data-method="post"
+                            <button type="button"
+                                    class="ui orange button"
+                                    data-href="{{ url_for('.key_create') }}"
+                                    data-method="post"
                                     data-params='{"force": "1"}'
                                     data-title="{% trans %}Create new API key{% endtrans %}"
-                                    data-confirm="{% trans %}Do you really want to create a new API key? Your old key will stop working immediately!{% endtrans %}">
+                                    data-confirm="{% trans %}Do you really want to create a new API key?
+                                    Your old key will stop working immediately!{% endtrans %}">
                                 <i aria-hidden="true" class="icon plus"></i>
                                 {%- trans %}Create new API key{% endtrans -%}
                             </button>
-                            <button type="button" class="ui negative button" data-href="{{ url_for('.key_delete') }}" data-method="post"
+                            <button type="button"
+                                    class="ui negative button"
+                                    data-href="{{ url_for('.key_delete') }}"
+                                    data-method="post"
                                     data-title="{% trans %}Delete API key{% endtrans %}"
                                     data-confirm="{% trans %}Do you really want to delete your API key?{% endtrans %}">
                                 <i aria-hidden="true" class="icon trash alternate"></i>
@@ -38,7 +45,10 @@
                         {%- endtrans -%}
                     </p>
 
-                    <button type="button" class="ui primary button" data-href="{{ url_for('.key_create') }}" data-method="post">
+                    <button type="button"
+                            class="ui primary button"
+                            data-href="{{ url_for('.key_create') }}"
+                            data-method="post">
                         {%- trans %}Create API key{% endtrans -%}
                     </button>
                 {% else %}
@@ -47,7 +57,10 @@
                         <dd>
                             {{ key.token }}
                             {% if key.is_blocked %}
-                                <i class="icon-warning warningText" title="{% trans %}Your API key has been blocked. Please contact an administrator for further details.{% endtrans %}"></i>
+                                <i class="icon-warning warningText"
+                                   title="{% trans %}Your API key has been blocked.
+                                   Please contact an administrator for further details.{% endtrans %}">
+                                </i>
                             {% endif %}
                         </dd>
                         {% if use_signatures %}
@@ -71,9 +84,13 @@
                             <dd>
                                 {{ key.last_used_uri }}
                                 {% if key.last_used_auth %}
-                                    <i class="icon-shield" title="{% trans %}Last request could access private data.{% endtrans %}"></i>
+                                    <i class="icon-shield"
+                                       title="{% trans %}Last request could access private data.{% endtrans %}">
+                                    </i>
                                 {% else %}
-                                    <i class="icon-earth" title="{% trans %}Last request could only access public data.{% endtrans %}"></i>
+                                    <i class="icon-earth"
+                                       title="{% trans %}Last request could only access public data.{% endtrans %}">
+                                    </i>
                                 {% endif %}
                             </dd>
                         {% endif %}
@@ -104,7 +121,8 @@
                                 class="ui warning button"
                                 data-href="{{ url_for('core.reset_signature_tokens') }}"
                                 data-method="POST"
-                                data-confirm="{% trans %}Are you sure you'd like to invalidate all your token-based links?{% endtrans %}">
+                                data-confirm="{% trans %}Are you sure you'd like to invalidate all your
+                                token-based links?{% endtrans %}">
                             <i aria-hidden="true" class="icon trash alternate"></i>
                             {%- trans %}Invalidate all links{% endtrans -%}
                         </button>
@@ -131,12 +149,16 @@
                         <div class="i-box-buttons toolbar">
                             <div class="group">
                                 {% if key.is_blocked %}
-                                    <button type="button" class="ui positive button" data-href="{{ url_for('.key_block') }}"
+                                    <button type="button"
+                                            class="ui positive button"
+                                            data-href="{{ url_for('.key_block') }}"
                                             data-method="post">
                                         {%- trans %}Unblock API key{% endtrans -%}
                                     </button>
                                 {% else %}
-                                    <button type="button" class="ui negative button" data-href="{{ url_for('.key_block') }}"
+                                    <button type="button"
+                                            class="ui negative button"
+                                            data-href="{{ url_for('.key_block') }}"
                                             data-method="post">
                                         <i aria-hidden="true" class="icon ban"></i>
                                         {%- trans %}Block API key{% endtrans -%}
--- a/indico/modules/api/templates/user_profile.html
+++ b/indico/modules/api/templates/user_profile.html
@@ -13,7 +13,7 @@
                     <div class="i-box-buttons toolbar">
                         <div class="group">
                             <button type="button"
-                                    class="ui orange button"
+                                    class="ui orange small button"
                                     data-href="{{ url_for('.key_create') }}"
                                     data-method="post"
                                     data-params='{"force": "1"}'
@@ -24,7 +24,7 @@
                                 {%- trans %}Create new API key{% endtrans -%}
                             </button>
                             <button type="button"
-                                    class="ui negative button"
+                                    class="ui negative small button"
                                     data-href="{{ url_for('.key_delete') }}"
                                     data-method="post"
                                     data-title="{% trans %}Delete API key{% endtrans %}"
@@ -118,7 +118,7 @@
                 <div class="i-box-buttons toolbar">
                     <div class="group">
                         <button type="button"
-                                class="ui warning button"
+                                class="ui negative small button"
                                 data-href="{{ url_for('core.reset_signature_tokens') }}"
                                 data-method="POST"
                                 data-confirm="{% trans %}Are you sure you'd like to invalidate all your
@@ -150,14 +150,14 @@
                             <div class="group">
                                 {% if key.is_blocked %}
                                     <button type="button"
-                                            class="ui positive button"
+                                            class="ui positive small button"
                                             data-href="{{ url_for('.key_block') }}"
                                             data-method="post">
                                         {%- trans %}Unblock API key{% endtrans -%}
                                     </button>
                                 {% else %}
                                     <button type="button"
-                                            class="ui negative button"
+                                            class="ui negative small button"
                                             data-href="{{ url_for('.key_block') }}"
                                             data-method="post">
                                         <i aria-hidden="true" class="icon ban"></i>
--- a/indico/modules/events/features/controllers.py
+++ b/indico/modules/events/features/controllers.py
@@ -37,8 +37,7 @@ class RHFeatures(RHFeaturesBase):
         for name, feature in sorted(get_feature_definitions().iteritems(), key=lambda x: x[1].friendly_name):
             if name in disallowed:
                 continue
-            field = BooleanField(feature.friendly_name, widget=SwitchWidget(),
-                                 description=feature.description)
+            field = BooleanField(feature.friendly_name, widget=SwitchWidget(), description=feature.description)
             setattr(form_class, name, field)
         defaults = {name: True for name in get_enabled_features(self.event)}
         return form_class(csrf_enabled=False, obj=FormDefaults(defaults))
--- a/indico/modules/events/features/controllers.py
+++ b/indico/modules/events/features/controllers.py
@@ -17,7 +17,7 @@ from indico.modules.events.features.util import (format_feature_names, get_disal
 from indico.modules.events.features.views import WPFeatures
 from indico.modules.events.logs import EventLogKind, EventLogRealm
 from indico.modules.events.management.controllers import RHManageEventBase
-from indico.util.i18n import _, ngettext
+from indico.util.i18n import ngettext
 from indico.web.forms.base import FormDefaults, IndicoForm
 from indico.web.forms.widgets import SwitchWidget
 from indico.web.menu import render_sidemenu
@@ -37,7 +37,7 @@ class RHFeatures(RHFeaturesBase):
         for name, feature in sorted(get_feature_definitions().iteritems(), key=lambda x: x[1].friendly_name):
             if name in disallowed:
                 continue
-            field = BooleanField(feature.friendly_name, widget=SwitchWidget(on_label=_('On'), off_label=_('Off')),
+            field = BooleanField(feature.friendly_name, widget=SwitchWidget(),
                                  description=feature.description)
             setattr(form_class, name, field)
         defaults = {name: True for name in get_enabled_features(self.event)}
--- a/indico/modules/users/blueprint.py
+++ b/indico/modules/users/blueprint.py
@@ -12,12 +12,13 @@ from flask import request
 from indico.modules.users.api import RHUserFavoritesAPI, fetch_authenticated_user
 from indico.modules.users.controllers import (RHAcceptRegistrationRequest, RHAdmins, RHExportDashboardICS,
                                               RHPersonalData, RHRegistrationRequestList, RHRejectRegistrationRequest,
-                                              RHUserDashboard, RHUserEmails, RHUserEmailsDelete, RHUserEmailsSetPrimary,
-                                              RHUserEmailsVerify, RHUserFavorites, RHUserFavoritesCategoryAPI,
-                                              RHUserFavoritesUserRemove, RHUserFavoritesUsersAdd, RHUserPreferences,
-                                              RHUsersAdmin, RHUsersAdminCreate, RHUsersAdminMerge,
-                                              RHUsersAdminMergeCheck, RHUsersAdminSettings, RHUserSearch,
-                                              RHUserSearchInfo, RHUserSuggestionsRemove)
+                                              RHUserBlock, RHUserDashboard, RHUserEmails, RHUserEmailsDelete,
+                                              RHUserEmailsSetPrimary, RHUserEmailsVerify, RHUserFavorites,
+                                              RHUserFavoritesCategoryAPI, RHUserFavoritesUserRemove,
+                                              RHUserFavoritesUsersAdd, RHUserPreferences, RHUsersAdmin,
+                                              RHUsersAdminCreate, RHUsersAdminMerge, RHUsersAdminMergeCheck,
+                                              RHUsersAdminSettings, RHUserSearch, RHUserSearchInfo,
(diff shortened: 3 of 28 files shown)


indico/modules/users/controllers.py
```diff
@@ -639,3 +639,16 @@ class RHUserSearchInfo(RHProtected):
     def _process(self):
         external_users_available = any(auth.supports_search for auth in multipass.identity_providers.itervalues())
         return jsonify(external_users_available=external_users_available)
+
+
+class RHUserBlock(RHUserBase):
+    def _block(self, blocked):
+        self.user.is_blocked = blocked
+        flash(_('The user has been {}.'.format('blocked' if blocked else 'unblocked')), 'success')
```
ThiefMaster: This is not how i18n works ;)

- you need to have a static string inside `_(...)` for extraction to work
- concatenating strings makes translation hard to impossible (for example, blocked/unblocked are not translated right now, and even if you wrapped those in `_()`, it would only work for languages where the word is at the end of the sentence)

So better to use two separate (fully translatable) sentences.

indico/modules/users/controllers.py
```diff
@@ -639,3 +639,16 @@ class RHUserSearchInfo(RHProtected):
...
+        return jsonify(success=True)
+
+    def _process_PUT(self):
+        self._block(True)
```
ThiefMaster: I don't think using a separate method is useful considering my i18n comment above. Having 3 lines here is much simpler and more readable:

- set the blocked flag
- flash a message
- (currently missing) use the `logger` to log who's blocking who
- return json (you can use `jsonify_data(flash=False)` to already have the success flag included btw!)

indico/modules/users/templates/personal_data.html
```diff
@@ -9,6 +9,27 @@
                 <div class="i-box-title">
                     {%- trans %}Details{% endtrans -%}
                 </div>
+                <div class="i-box-buttons">
+                    {% if not user.is_blocked %}
+                        <button class="ui negative small button"
+                                data-href="{{ url_for('.user_block') }}"
+                                data-method="PUT"
+                                data-confirm="This user will no longer have access to Indico.<br>
```
ThiefMaster: missing i18n

indico/modules/users/templates/personal_data.html
```diff
@@ -9,6 +9,27 @@
...
+                                Are you sure you want to revoke his access?"
```
ThiefMaster: *their* instead of *his* is a more neutral form

indico/modules/users/templates/personal_data.html
```diff
@@ -9,6 +9,27 @@
...
+                                data-reload-after>
+                            {% trans %}Block Access{% endtrans %}
```
ThiefMaster: "Block user"?

plourenco: I had "Block user" but changed it because *blocking a user* could be confused with you blocking a person rather than blocking the person's access to the website.

indico/modules/events/features/controllers.py
```diff
@@ -37,7 +37,7 @@ def _make_form(self):
         for name, feature in sorted(get_feature_definitions().iteritems(), key=lambda x: x[1].friendly_name):
             if name in disallowed:
                 continue
-            field = BooleanField(feature.friendly_name, widget=SwitchWidget(on_label=_('On'), off_label=_('Off')),
+            field = BooleanField(feature.friendly_name, widget=SwitchWidget(),
                                  description=feature.description)
```
mic4ael: This could be a single line.

indico/modules/users/templates/users_admin.html
```diff
@@ -79,6 +79,15 @@ <h3>
                                             {{ user_entry.full_name }}
                                         {%- endif -%}
                                     </span>
+                                    {% if user_entry.user.is_admin %}
+                                        <span class="ui blue tiny label">Admin</span>
```
mic4ael: Missing i18n in these strings

indico/modules/users/blueprint.py
```diff
@@ -56,6 +57,7 @@
     _bp.add_url_rule('/emails/verify/<token>', 'user_emails_verify', RHUserEmailsVerify)
     _bp.add_url_rule('/emails/<email>', 'user_emails_delete', RHUserEmailsDelete, methods=('DELETE',))
     _bp.add_url_rule('/emails/make-primary', 'user_emails_set_primary', RHUserEmailsSetPrimary, methods=('POST',))
+    _bp.add_url_rule('/block/', 'user_block', RHUserBlock, methods=('PUT', 'DELETE'))
```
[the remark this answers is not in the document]
plourenco: I can omit the trailing slash but it will become the only entry of this block without one.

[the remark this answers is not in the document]
plourenco: Not sure about the `/blocked` though. In theory, we will PUT to `block` and DELETE a `block`. The same with the previous `/emails/make-primary`, we will POST to "make primary".

I think it makes more sense to have the route reflecting an action we will trigger in the present, and not in the past.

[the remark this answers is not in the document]
ThiefMaster: there's nothing "inside" `/block/`, that's why i'd skip the slash (`/emails/` has `/emails/<email>` inside for example)

indico/modules/users/controllers.py
plourenco: I agree, this is over-engineering.

indico/modules/users/blueprint.py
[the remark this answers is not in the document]
ThiefMaster: the idea behind PUTting and DELETEing `/user/blocked` is that you are setting/clearing the "blocked" flag on the user, while you POST the make-primary verb (even though this isn't fully restful - usually one would PATCH the email with `{"is_primary": true}` to be fully restful)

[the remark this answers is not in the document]
plourenco: > there's nothing "inside" `/block/`, that's why i'd skip the slash (`/emails/` has `/emails/<email>` inside for example)

Should we keep the following without it?
```
_bp.add_url_rule('/profile/', 'user_profile', RHPersonalData, methods=('GET', 'POST'))
_bp.add_url_rule('/preferences/', 'user_preferences', RHUserPreferences, methods=('GET', 'POST'))
_bp.add_url_rule('/favorites/', 'user_favorites', RHUserFavorites)
```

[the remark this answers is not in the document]
ThiefMaster: those are user-facing so i wouldn't change them since there's no redirect from slash to no-slash, only vice versa (but we should probably have done them without the slash in the first place).

anyway, the block action will only be called in the background and never show up for an end user

indico/modules/users/controllers.py
plourenco: That makes sense, I forgot about languages where the word could change place.

indico/modules/users/blueprint.py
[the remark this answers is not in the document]
plourenco: I think the crux here is that we're trying to approach the ideal PATCH solution. I agree with both solutions even though `block` initially seemed more straight-forward when going against the pattern.

I've changed it to `blocked`.

plourenco:
Updated my last commit to include labels in the user dashboard.
image

indico/modules/users/templates/dashboard.html
```diff
@@ -42,6 +43,9 @@
                     </div>
                     <div class="your-details-wrapper">
                         <h3>{{ user.full_name }}</h3>
+                        <div class="your-labels">
+                            {{ labels(user) }}
+                        </div>
                         {% if config.DEBUG %}
                             <div>
                                 <i class="icon-info" title="{% trans %}User ID{% endtrans %}"></i><span>{{ user.id }}</span>
```
ThiefMaster: What about moving this one to something like a label as well? I added it a long time ago as a quick way to get my own ID, but always found the way it's shown quite ugly.. ;x

indico/modules/users/controllers.py
```diff
@@ -639,3 +639,17 @@ class RHUserSearchInfo(RHProtected):
     def _process(self):
         external_users_available = any(auth.supports_search for auth in multipass.identity_providers.itervalues())
         return jsonify(external_users_available=external_users_available)
+
+
+class RHUserBlock(RHUserBase):
+    def _process_PUT(self):
+        self.user.is_blocked = True
+        logger.info("User %s blocked %s's access", session.user, self.user)
```
ThiefMaster: Can be more concise: `'User %r blocked %r'` (it goes to the logfile, so no need to write verbose sentences)

indico/modules/users/templates/users_admin.html
```diff
@@ -79,10 +80,15 @@ <h3>
                                             {{ user_entry.full_name }}
                                         {%- endif -%}
                                     </span>
+                                    {% if user_entry.user %}
+                                        {{ labels(user_entry.user) }}
+                                    {% endif %}
                                     <span class="right">
                                         {% if user_entry.user and user_entry.user.is_system %}
                                             <em title="{% trans %}This is the system user.{% endtrans %}">
```
ThiefMaster: I don't think we need to keep the `<em>` here; the title could be set directly on the span below

indico/modules/api/templates/user_profile.html
```diff
@@ -12,15 +12,17 @@
                 {% if key and can_modify %}
                     <div class="i-box-buttons toolbar">
                         <div class="group">
-                            <button type="button" class="i-button warning icon-plus" data-href="{{ url_for('.key_create') }}" data-method="post"
+                            <button type="button" class="ui orange button" data-href="{{ url_for('.key_create') }}" data-method="post"
```
mic4ael: Maybe you could use this as an opportunity to wrap this line and put some of the properties in the next line?

indico/modules/api/templates/user_profile.html
```diff
@@ -12,15 +12,17 @@
...
                                     data-params='{"force": "1"}'
                                     data-title="{% trans %}Create new API key{% endtrans %}"
                                     data-confirm="{% trans %}Do you really want to create a new API key? Your old key will stop working immediately!{% endtrans %}">
+                                <i aria-hidden="true" class="icon plus"></i>
                                 {%- trans %}Create new API key{% endtrans -%}
                             </button>
-                            <button type="button" class="i-button danger icon-remove" data-href="{{ url_for('.key_delete') }}" data-method="post"
+                            <button type="button" class="ui negative button" data-href="{{ url_for('.key_delete') }}" data-method="post"
```
mic4ael: Maybe the same as above?

mic4ael:
Just noticed a small thing which might not be related to your additions but I think it is something worth looking into.

The elements in each line seem to be misaligned.

Screenshot 2020-03-24 at 10 22 17

indico/modules/api/templates/user_profile.html
plourenco: Wouldn't this imply changing it in all the files to keep it consistent? 
I think in that case we should enforce it by linting.

indico/modules/users/templates/dashboard.html
plourenco: I'm not sure if having the user id as a label will fit very well, considering labels are intended for user status/roles and ids may look disproportional. But I'll try that.

indico/modules/api/templates/user_profile.html
ThiefMaster: I don't think there's a good linter tool that works with Jinja. Anyway, we usually wrap if the line becomes longer than 120 chars otherwise. No need to be super consistent with wrapping things everywhere though - just wrap it nicely in new code ;)

indico/modules/users/templates/dashboard.html
ThiefMaster: yeah, not sure - but worth a try IMO (it's only shown in debug mode anyway)

plourenco:
> Just noticed a small thing which might not be related to your additions but I think it is something worth looking into.
> 
> The elements in each line seem to be misaligned.
> 
> Screenshot 2020-03-24 at 10 22 17

It is related, yes. The new semantic-ui buttons are a bit bigger, adding some extra white-space. 
Nice catch.

indico/modules/users/controllers.py
plourenco: Agree

ThiefMaster:
Maybe we can use a smaller size for the button there?

indico/modules/api/templates/user_profile.html
mic4ael: Yeah, we don't have to update every file, just the files that introduce some changes.

ThiefMaster: (but please don't start major reformattings of unchanged code parts, it makes reviewing PRs harder)

indico/modules/users/templates/personal_data.html
ThiefMaster: but isn't that the same?

plourenco: We agreed to keep "Block User" since the opposite, a feature to just block someone's activity, won't be considered.

indico/web/client/styles/base/_typography.scss
```diff
@@ -62,35 +66,36 @@ body {
     background: #fff;
     padding: 0px;
     margin: 0px;
-}
-
-body > :not(.rb-app-container) {
```
plourenco: It makes sense to read this and understand that the "Liberation Sans" font styles are applied except for the room booking container. However, I couldn't find any evidence of this CSS selector (.rb-app-container) ever being used. That's why it has been removed.

ThiefMaster: i don't think we include anything from the "base" css in room booking anyway

panagiotappl: I don't think this is supposed to be removed. There are cases where no other CSS rules from `_typography.scss` (like `.conference-page` or `.description` etc) are applied and then you need this rule to ensure that "Liberation Sans" and not "Muli" is applied.

ThiefMaster: But the behavior doesn't change as far as I can see, since `:not(.rb-app-container)` was always true.

The only problem I could see is that `body > xxx` is more specific than `body`, so in some cases it may not apply this anymore now.

panagiotappl: But since we have included SUI everywhere in indico, it's trying to enforce a different font in the body and we need this rule in order to avoid that.

panagiotappl: Ah, I understand now. My bad. Thought he meant that the entire rule was not needed.

indico/modules/users/templates/dashboard.html
```diff
@@ -42,11 +43,15 @@
...
-                        {% if config.DEBUG %}
-                            <div>
-                                <i class="icon-info" title="{% trans %}User ID{% endtrans %}"></i><span>{{ user.id }}</span>
-                            </div>
-                        {% endif %}
+                        <div class="your-labels">
+                            {% if config.DEBUG %}
+                                <div class="ui label">
+                                  ID
+                                  <div class="detail">{{ user.id }}</div>
```
ThiefMaster: We use 4-space alignment in HTM

indico/modules/users/controllers.py
```diff
@@ -639,3 +639,17 @@ class RHUserSearchInfo(RHProtected):
...
         external_users_available = any(auth.supports_search for auth in multipass.identity_providers.itervalues())
         return jsonify(external_users_available=external_users_available)
+
+
+class RHUserBlock(RHUserBase):
+    def _process_PUT(self):
+        self.user.is_blocked = True
+        logger.info("User %s blocked %s", session.user, self.user)
+        flash(_('{name} has been blocked.').format(name=self.user.name), 'success')
+        return jsonify(flash=False)
```
ThiefMaster: Use either `jsonify(success=True)` or `jsonify_data(flash=False)` (check the code of the latter)

indico/modules/users/controllers.py
```diff
@@ -639,3 +639,17 @@ class RHUserSearchInfo(RHProtected):
...
+
+    def _process_DELETE(self):
+        self.user.is_blocked = False
+        logger.info("User %s unblocked %s", session.user, self.user)
+        flash(_('{name} has been unblocked.').format(name=self.user.name), 'success')
+        return jsonify(flash=False)
```
ThiefMaster: same here

indico/modules/users/templates/users_admin.html
```diff
@@ -79,11 +87,14 @@ <h3>
                                             {{ user_entry.full_name }}
                                         {%- endif -%}
                                     </span>
+                                    {% if user_entry.user %}
+                                        {{ labels(user_entry.user) }}
+                                    {% endif %}
                                     <span class="right">
```
ThiefMaster: This is no longer right-aligned. It looks especially strange when there's a label for the user as well.

indico/web/client/styles/partials/_boxes.scss
```diff
@@ -398,7 +402,9 @@ $i-box-padding: 10px;
     > li {
         @include border-top();
         padding: 10px;
-        position: relative;
+        display: flex;
```
ThiefMaster: I think it's this change that breaks the alignment of the search result list and email management page

indico/modules/api/templates/user_profile.html
```diff
@@ -99,10 +118,12 @@
                 <div class="i-box-buttons toolbar">
                     <div class="group">
                         <button type="button"
-                                class="i-button warning icon-remove"
+                                class="ui warning button"
```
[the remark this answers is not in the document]
plourenco: Yes, I agree. I've restored the size to small. 
However, the header will still expand to cope with the button size and the padding. Imo, I would keep it like this (with `small`). I expect the headers to expand if there's a button inside.

indico/web/client/styles/base/_typography.scss
```diff
@@ -62,35 +66,36 @@ body {
...
     @extend %font-family-title-light;
 }
 
-body > :not(.rb-app-container) h2 {
+body > h2 {
     font-size: 1.6em;
 }
 
-body > :not(.rb-app-container) h3 {
+body > h3 {
```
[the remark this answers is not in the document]
plourenco: Agree! Restored to standalone rules, as before.

indico/modules/users/templates/dashboard.html
plourenco: Thanks.

indico/web/client/styles/partials/_boxes.scss
plourenco: I've refactored these alignments. The previous rules should be intact now.

indico/modules/users/templates/_favorites.html
```diff
@@ -1,9 +1,16 @@
...
+        <li data-user-id="{{ fav.id }}" class="flexrow f-a-center js-user-row">
             <span>{{ fav.full_name }}</span>
-            <a class="toggle icon-close right js-delete-user" title="{% trans %}Remove from favourites{% endtrans %}" href="#"
-               data-href="{{ url_for('.user_favorites_user_remove', fav_user_id=fav.id) }}"></a>
+            <span class="right">
+                <span class="ui icon tiny button right js-delete-user"
+                     data-tooltip="{% trans %}Remove from favourites{% endtrans %}"
+                     data-inverted=""
+                     data-position="left center"
+                     data-href="{{ url_for('.user_favorites_user_remove', fav_user_id=fav.id) }}">
```
ThiefMaster: needs one more space for indentation

indico/modules/users/controllers.py
```diff
@@ -639,3 +639,17 @@ class RHUserSearchInfo(RHProtected):
     def _process(self):
         external_users_available = any(auth.supports_search for auth in multipass.identity_providers.itervalues())
         return jsonify(external_users_available=external_users_available)
+
+
+class RHUserBlock(RHUserBase):
+    def _process_PUT(self):
+        self.user.is_blocked = True
+        logger.info("User %s blocked %s", session.user, self.user)
```
mic4ael: Sorry for being nitpicky but we usually use single quotes.

indico/modules/users/controllers.py
```diff
@@ -639,3 +639,17 @@ class RHUserSearchInfo(RHProtected):
...
+        flash(_('{name} has been blocked.').format(name=self.user.name), 'success')
+        return jsonify(success=True)
+
+    def _process_DELETE(self):
+        self.user.is_blocked = False
+        logger.info("User %s unblocked %s", session.user, self.user)
```
mic4ael: ditto

indico/modules/users/templates/_favorites.html
```diff
@@ -1,9 +1,16 @@
...
     {% for fav in user.favorite_users|sort(attribute='full_name') %}
-        <li data-user-id="{{ fav.id }}" class="js-user-row">
+        <li data-user-id="{{ fav.id }}" class="flexrow f-a-center js-user-row">
             <span>{{ fav.full_name }}</span>
-            <a class="toggle icon-close right js-delete-user" title="{% trans %}Remove from favourites{% endtrans %}" href="#"
-               data-href="{{ url_for('.user_favorites_user_remove', fav_user_id=fav.id) }}"></a>
+            <span class="right">
+                <span class="ui icon tiny button right js-delete-user"
+                     data-tooltip="{% trans %}Remove from favourites{% endtrans %}"
+                     data-inverted=""
```
mic4ael: Is this attribute handled somehwere? (ahh I guess it is handled by SUIR)

indico/modules/users/templates/emails.html
```diff
@@ -18,22 +18,31 @@
...
                             <span>{{ email }}</span>
-                            <a class="toggle icon-remove right js-delete-email" href="#" data-href="{{ url_for('.user_emails_delete', email=email) }}" style="margin-left: 10px;"></a>
-                            <button type="button" class="i-button right" data-href="{{ url_for('.user_emails_set_primary') }}" data-method='POST' data-params="{{ {'email': email} | tojson | forceescape }}">
-                                {%- trans %}Set as primary{% endtrans -%}
-                            </button>
+                            <span class="right">
+                                <button type="button"
+                                        class="ui primary small button"
+                                        data-href="{{ url_for('.user_emails_set_primary') }}"
+                                        data-method='POST'
```
mic4ael: `"` instead of `'`

indico/modules/users/templates/_favorites.html
ThiefMaster: I asked him the same on mattermost since I also didn't find anything: It's using the pure-css tooltips from SUI.

mic4ael: Ah, ok. I thought so..

plourenco: I ended up removing this because it was starting to look very odd, having different tooltips. When I came to myself I was already changing the tooltips in five files. So I rolled it back.

indico/modules/users/controllers.py
plourenco: No problem! I thought the linter was picking this up.

ThiefMaster: Does flake8 have a plugin which can check for this (single quotes, unless it would require escaping inside)? But even if it does, I think it'd require quite a few changes in existing files for English strings where we defaulted to double quotes for a while even if there were no apostrophes inside.

indico/modules/users/templates/_category.html
```diff
@@ -17,25 +17,27 @@
...
-            {# The caller must provide the closing span tag (</span>). #}
-            {# TODO This should be changed once Jinja is updated to 2.8 with support for set block. #}
-        </div>
+    <li class="category-box flexrow f-a-center">
+        <span class="category-box-info">
+            <a href="{{ category.url }}">{{ category.title }}</a>
+            {{ _category_path(category, truncated_path) }}
+            {{ caller() }}
+        {# The caller must provide the closing span tag (</span>). #}
+        {# TODO This should be changed once Jinja is updated to 2.8 with support for set block. #}
```
plourenco: Perhaps we could take the opportunity to refactor this?

indico/modules/users/controllers.py
```diff
@@ -402,7 +402,7 @@ def _process(self):
             external = form_data.pop('external')
             form_data = {k: v for (k, v) in form_data.iteritems() if v and v.strip()}
             matches = search_users(exact=exact, include_deleted=include_deleted, include_pending=include_pending,
-                                   external=external, allow_system_user=True, **form_data)
+                                   include_blocked=True, external=external, allow_system_user=True, **form_data)
```
plourenco: In the admin search, I'm just including the blocked users by default because we already have way too many options and a label will be shown anyway. Nonetheless, I will leave this thread for change requests if anyone disagrees.

indico/modules/users/templates/_category.html
ThiefMaster: Sure! I also just noticed that `suggested_category` renders a div inside a span.. isn't that actually invalid?

plourenco: > Sure! I also just noticed that `suggested_category` renders a div inside a span.. isn't that actually invalid?

Yes, that's not really inline. I've updated both!

indico/modules/users/controllers.py
plourenco: Something like this?
https://pypi.org/project/flake8-quotes/

indico/modules/users/templates/users_admin.html
```diff
@@ -48,9 +52,13 @@
                     {% endif %}
                 </div>
             </div>
-            <div class="i-box-content">
+            <span class="i-box-content">
                 <div class="info-message-box">
-                    <div class="message-text">{% trans num_of_users=num_of_users %}There are {{ num_of_users }} users, {{ num_deleted_users }} of which are deleted{% endtrans %}</div>
+                    <div class="message-text">
+                        {%- trans num_of_users=num_of_users -%}
```
ThiefMaster, line 58: `num_of_users=num_of_users` isn't needed; plain variables already work out of the box. you only need to do stuff here in cases where you are accessing e.g. some attribute (since you cannot do that inside a trans block)

indico/modules/users/controllers.py
ThiefMaster: Oh nice, this looks great! I just gave it a try and with some configuration it works pretty well to avoid stuff like unnecessary escapes.

unfortunately the "wrong type of quotes" check isn't really suitable for us:

```
$ flake8 indico/ | grep Q000 | wc -l
1366
```

a bit better for multiline strings, but that's more because we don't have that many which aren't docstrings..

```
$ flake8 indico/ | grep Q001 | wc -l
138
```

and even better when ignoring it for `indico/core/signals/` since the multiline strings there are basically docstrings (and thus double quotes are more appropriate).

```
$ flake8 indico/ | grep Q001 | wc -l
34
```

But I don't think it really matters for multiline strings, so I'm inclined to disable Q000 and Q001 globally and just let it complain about unnecessary escapes and docstrings not using double-quotes.

Too bad there's no way to run this check just for newly added code...

indico/modules/users/templates/users_admin.html
plourenco: Thanks, this was already here before the PR, but I didn't even notice it.

indico/modules/users/templates/_category.html
```diff
@@ -16,55 +16,59 @@
...
-{% macro _category_item(category, truncated_path, extra_info='') %}
-    <li class="category-box">
-        <div>
-            <span class="category-box-info">
-                <a href="{{ category.url }}">{{ category.title }}</a>
-                {{ _category_path(category, truncated_path) }}
-                {{ caller() }}
-            {# The caller must provide the closing span tag (</span>). #}
-            {# TODO This should be changed once Jinja is updated to 2.8 with support for set block. #}
+{% macro _category_item(category, truncated_path, extra_info) %}
```
ThiefMaster, line 19: `extra_info=''` so we don't rely on `undefined` evaluating as an empty string in cases where we don't provide this argument (i still want to enable `StrictUndefined` at some point, at least when in debug mode)

PR #4365, opened 2020-03-20
Port missing features to the new Call for Papers

mic4ael:


indico/modules/events/papers/client/js/components/PaperDecisionForm.jsx
```diff
@@ -58,6 +57,15 @@ export default function PaperDecisionForm() {
...
   }
 
+  if (state.name === PaperState.submitted && !actionOptions[PaperState.to_be_corrected]) {
+    actionOptions[PaperState.to_be_corrected] = {
+      value: 'to_be_corrected',
+      text: Translate.string('To be corrected'),
+    };
+  } else if (state.name === PaperState.to_be_corrected) {
+    delete actionOptions[PaperState.to_be_corrected];
+  }
```
ThiefMaster: Modifying an object that's not defined inside the component feels pretty dirty.. maybe better to just define it inside?

indico/modules/events/papers/client/js/components/GroupReviewForm.jsx
```diff
@@ -0,0 +1,209 @@
...
+      {fprops => (
+        <Form onSubmit={fprops.handleSubmit} styleName="group-review-form">
+          {!review && (
+            <>
+              <div className="form-preface">
+                <Translate>
+                  Reviewing in{' '}
+                  <Param
+                    name="group"
+                    value={<span>{group.title}</span>}
```
ThiefMaster: is that span needed?

indico/modules/events/papers/client/js/components/CommentForm.jsx
```diff
@@ -0,0 +1,148 @@
...
+    : {
+        onFocus: () => {
+          setCommentFormVisible(true);
+          onToggleExpand(true);
+        },
+      };
+
+  const handleSubmit = async (formData, form) => {
+    const error = await onSubmit(formData, form);
+    if (!error) {
```
ThiefMaster:
```js
if (error) {
  return error;
}
...
```

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
```diff
@@ -5,114 +5,136 @@
...
+            {reviewGroups[0] in reviewerData.reviews ? (
+              <Translate>Change review</Translate>
+            ) : (
+              <Translate>Review</Translate>
+            )}
+          </Button>
+        ) : (
+          <Dropdown
+            text={
+              _.isEqual(reviewGroups, reviewedGroups)
```
ThiefMaster: do they always have the correct order or could this check sometimes fail due to a different order in (one of) the arrays? if the order isn't guaranteed, our `unsortedArraysEqual` util would help

indico/modules/events/papers/operations.py
```diff
@@ -329,8 +329,13 @@ def create_comment(paper, text, visibility, user):
                     session.user)
 
 
-def update_comment(comment, text, visibility):
-    changes = comment.populate_from_dict({'text': text, 'visibility': visibility})
+def update_comment(comment, text=None, visibility=None):
+    changed = {}
+    if text:
+        changed['text'] = text
+    if visibility:
```
ThiefMaster: I think `is not None` would be better for those checks (this one here would fail if we had a visibility type with a value of 0 for example)

indico/modules/events/papers/controllers/api.py
```diff
@@ -67,12 +70,23 @@ def _check_access(self):
...
+                        .filter(PaperReviewComment.id == request.view_args['comment_id'],
+                                ~PaperReviewComment.is_deleted)
+                        .first_or_404())
 
     def _process_DELETE(self):
         delete_comment(self.comment)
         return '', 204
 
+    @use_kwargs({
+        'comment': fields.String(validate=validate.Length(min=1)),
```
ThiefMaster: Use our `not_empty` validator instead. `Length(min=1)` doesn't have an amazing error message

indico/web/client/js/react/components/ReviewRating.jsx
```diff
@@ -0,0 +1,52 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Rating} from 'semantic-ui-react';
+
+import './ReviewRating.module.scss';
+
+export default function ReviewRating({min, max, value, disabled, onChange}) {
+  const ratingRange = _.range(min, max + 1);
+  const suiRatingRange = _.range(1, ratingRange.length + 1);
+  const maxRating = suiRatingRange[suiRatingRange.length - 1];
```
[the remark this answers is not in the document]
mic4ael: Nope, it is not. `maxRating` is the prop passed to  SUIR `Rating` field which operates on values between 0 and `max`. However, users can declare the rating range to be less than 0 so in case of a negative `min`, `max + 1` won't work correctly here.

indico/web/client/js/react/forms/fields.jsx
```diff
@@ -408,6 +421,21 @@ FinalDropdown.defaultProps = {
   multiple: false,
 };
 
+export function FinalRating({name, label, ...rest}) {
```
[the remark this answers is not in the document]
mic4ael: Hm.. I didn't notice we have `FinalFileManager` in the file where the component declaration is. I don't mind moving it even though I think it is cleaner to put it in the `forms/fields.jsx`

indico/web/client/js/react/components/ReviewRating.jsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <span styleName="rating-field">
+      <Rating
+        maxRating={maxRating}
+        rating={value === null ? null : rating}
+        disabled={disabled}
+        onRate={(evt, {rating: newRating, ...rest}) => {
+          // map the value from SUIR Rating to the corresponding value between min and max
+          const val = ratingRange[suiRatingRange.findIndex(item => item === newRating)];
+          onChange(evt, {rating: val, ...rest});
```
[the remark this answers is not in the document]
mic4ael: Nope, the only thing I care about is the new value of the field.

indico/modules/events/papers/schemas.py
```diff
@@ -80,28 +101,64 @@ class PaperRevisionSchema(mm.ModelSchema):
...
+            return None
+
+        data = dict(revision.get_reviewer_render_data(self.context.get('user')))
+        review_type_schema = PaperReviewTypeSchema()
+        for name in {'groups', 'missing_groups', 'reviewed_groups'}:
+            data[name] = [review_type_schema.dump(item.instance) for item in data[name]]
+
+        reviews = {}
+        for key, value in data['reviews'].iteritems():
+            reviews[str(key.instance.title)] = paper_review_schema.dump(value)
```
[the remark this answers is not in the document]
mic4ael: Yup, it is a `LazyString`. I just realized I can replace it with `indico.util.i18n.orig_string(key.instance.title)`

indico/modules/events/papers/operations.py
```diff
@@ -329,8 +329,13 @@ def create_comment(paper, text, visibility, user):
...
+        changed['visibility'] = visibility
+    changes = comment.populate_from_dict(changed)
```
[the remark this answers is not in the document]
mic4ael: Hm, yeah, I completely agree. Good catch

indico/modules/events/papers/client/js/components/GroupReviewForm.jsx
mic4ael: Hm, probably not. Before it used to have a tooltip attached but I removed it and probably forgot to remove the `span` altogether.

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
mic4ael: I am pretty sure the order is guaranteed but I think it won't hurt to use `unsortedArraysEqual`

indico/web/client/js/react/forms/fields.jsx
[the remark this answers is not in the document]
ThiefMaster: hm to me it feels like we're overloading that file a bit if we put all the special/custom components there as well. it's already quite large without them!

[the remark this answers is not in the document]
mic4ael: Yeah, I completely agree and I also thought that this file is getting a bit bloated.

indico/modules/events/papers/client/js/components/CommentForm.jsx
```diff
@@ -0,0 +1,148 @@
...
+    if (onToggleExpand) {
+      onToggleExpand(false);
+    }
+
+    setTimeout(() => form.reset(), 0);
+  };
+
+  const initialValues = comment ? {comment: comment.text} : {};
+  if (!canJudge) {
+    delete visibilityOptions.judges;
```
ThiefMaster: this is still modifying something defined outside this function..

indico/modules/events/papers/client/js/components/CommentForm.jsx
```diff
@@ -0,0 +1,148 @@
...
+}
+
+CommentForm.propTypes = {
+  onSubmit: PropTypes.func.isRequired,
+  onToggleExpand: PropTypes.func,
+  comment: PropTypes.shape({
+    text: PropTypes.string,
+    visibility: PropTypes.shape({
+      name: PropTypes.string.isRequired,
+    }),
```
ThiefMaster: Just curious, why is this nested into an object instead of just the visibility.

indico/modules/events/papers/client/js/components/CommentForm.jsx
```diff
@@ -0,0 +1,148 @@
...
+  );
+}
+
+CommentForm.propTypes = {
+  onSubmit: PropTypes.func.isRequired,
+  onToggleExpand: PropTypes.func,
+  comment: PropTypes.shape({
+    text: PropTypes.string,
+    visibility: PropTypes.shape({
+      name: PropTypes.string.isRequired,
```
ThiefMaster: `PropTypes.oneOf([...]).isRequired`?

indico/modules/events/papers/client/js/components/PaperDecisionForm.jsx
```diff
@@ -58,6 +46,26 @@ export default function PaperDecisionForm() {
...
+    },
+  };
+
+  if (state.name === PaperState.submitted && !actionOptions[PaperState.to_be_corrected]) {
+    actionOptions[PaperState.to_be_corrected] = {
+      value: 'to_be_corrected',
+      text: Translate.string('To be corrected'),
+    };
+  } else if (state.name === PaperState.to_be_corrected) {
+    delete actionOptions[PaperState.to_be_corrected];
```
ThiefMaster: not needed anymore

indico/modules/events/papers/client/js/components/PaperReviewForm.jsx
```diff
@@ -7,112 +7,134 @@
...
+          </Dropdown>
+        )}
+      </div>
+    );
+  };
+
+  const renderForm = () => {
+    if (reviewGroup) {
+      const group = reviewerData.groups.find(item => item.name === reviewGroup);
+      const review = reviewGroup in reviewerData.reviews ? reviewerData.reviews[reviewGroup] : null;
```
ThiefMaster: `reviewerData.reviews[reviewGroup] || null`?

indico/modules/events/papers/controllers/api.py
```diff
@@ -106,3 +120,82 @@ def _check_paper_protection(self):
...
+        elif question.field_type == 'text':
+            validators = []
+            if question.field_data['max_length']:
+                validators.append(validate.Length(max=question.field_data['max_length']))
+            if question.field_data['max_words']:
+                validators.append(max_words(question.field_data['max_words']))
+
+            attrs['validate'] = validators
+            field_cls = fields.String
+        else:
```
ThiefMaster: wouldn't it be safer to use an elif here as well and raise an exception in else or use else for the `text` type?

indico/modules/events/papers/schemas.py
```diff
@@ -80,28 +102,64 @@ class PaperRevisionSchema(mm.ModelSchema):
...
+                  'is_last_revision', 'number', 'state', 'timeline', 'judgment_comment', 'judgment_comment_html',
+                  'reviewer_data')
+
+    def _get_reviewer_data(self, revision):
+        if not revision.is_last_revision:
+            return None
+
+        data = dict(revision.get_reviewer_render_data(self.context.get('user')))
+        review_type_schema = PaperReviewTypeSchema()
+        for name in {'groups', 'missing_groups', 'reviewed_groups'}:
```
ThiefMaster: `()` instead of `{}`? (no benefit of using a set and AFAIK there are more efficient opcodes for creating a small tuple - not that it really matters here...)

indico/modules/events/papers/schemas.py
```diff
@@ -80,28 +102,64 @@ class PaperRevisionSchema(mm.ModelSchema):
...
+            data[name] = [review_type_schema.dump(item.instance) for item in data[name]]
+
+        reviews = {}
+        for key, value in data['reviews'].iteritems():
```
ThiefMaster: I'd use `viewitems()` in new code; that can be replaced with `items()` in the (python 3) future straight away

indico/web/client/js/react/components/ReviewRating.jsx
```diff
@@ -0,0 +1,71 @@
...
+  disabled: false,
+  onChange: null,
+};
+
+function RatingAdapter(props) {
+  return <FormFieldAdapter {...props} as={ReviewRating} undefinedValue={null} />;
+}
+
+export function FinalRating({name, label, ...rest}) {
+  return <FinalField {...rest} name={name} label={label} adapter={RatingAdapter} format={v => v} />;
```
ThiefMaster: I don't think you need a custom adapter here. `component={ReviewRating} undefinedValue={[]}` should be enough.

indico/modules/events/papers/client/js/components/CommentForm.jsx
mic4ael: Ah.. ok. I think I changed it somewhere else :P

indico/modules/events/papers/schemas.py
mic4ael: Yeah I can change it so it is more consistent since we use tuples pretty much everywhere.

indico/modules/events/papers/controllers/api.py
mic4ael: Yeah, however it is pretty much improbable to happen :P

indico/modules/events/papers/client/js/components/CommentForm.jsx
mic4ael: I am using other visibility properties like `title` somewhere else. Here I only need the `name`

ThiefMaster: OK, didn't realize there's more than the name sent back from the server.

indico/web/client/js/react/components/ReviewRating.jsx
mic4ael: why an empty array?

ThiefMaster: copied from the wrong component. i meant `null`

indico/modules/events/papers/client/js/components/PaperDecisionForm.jsx
```diff
@@ -58,6 +46,24 @@ export default function PaperDecisionForm() {
...
+      value: 'accept',
+      text: Translate.string('Accept'),
+    },
+    [PaperState.rejected]: {
+      value: 'reject',
+      text: Translate.string('Reject'),
+    },
+  };
+
+  if (state.name === PaperState.submitted && !actionOptions[PaperState.to_be_corrected]) {
```
ThiefMaster: ` && !actionOptions[PaperState.to_be_corrected]` is always the case here

indico/modules/events/papers/client/js/components/RevisionReview.jsx
```diff
@@ -44,81 +82,134 @@ export default function RevisionReview({review}) {
...
-                </Translate>
-              )}{' '}
-              {review.ratings.length > 0 &&
-                (review.score !== null && (
+          {!isEditing && (
+            <div className="i-box-header flexrow">
+              <div className="review-badges">
+                {review.proposedAction.name === 'accept' && (
+                  <Translate>
+                    Proposed to <Param name="actionName" value="accept" wrapper={wrapper} />
```
ThiefMaster: The "accept" isn't translatable here... I think you want this:

```jsx
<Translate>
  Proposed to <Param name="wrapper" wrapper={wrapper}>accept</Param>
</Translate>
```

ThiefMaster:
This should be rounded (it was the case before)

image

mic4ael:
> This should be rounded (it was the case before)
> 
> image

So, I actually round it in any other case. You managed to find one usecase in which it somehow doesn't round :P

PR #4366, opened 2020-03-23
Enhance making changes to files in paper editing

### LINKED ISSUE #4362: Paper editing: Making changes not conform with filenameTemplate will still create entry in timeline
**How to recreate:**

1. Add a filename template in Paper Editing
2. In the timeline of paper editing, try to "Make changes"
3. Try to upload a file not conforming to the filename template (a warning should be displayed)
4. Add a comment and confirm
5. The comment and an entry `Editor has made some changes to the paper` should appear with the old, conform file on it

Instead of this the confirm button should not be clickable at all or there should be some kind of error displayed when trying to confirm the wrong file.

Leats:
Closes #4362.
It is no longer possible to create a new revision without changing at least one file.
It is no longer possible to click 'Confirm' when there are errors regarding the file name according to the filename template.

### DIFF
--- a/indico/modules/events/editing/client/js/components/FileManager/actions.js
+++ b/indico/modules/events/editing/client/js/components/FileManager/actions.js
@@ -16,6 +16,7 @@ export const CLEAR_DIRTY = 'CLEAR_DIRTY';
 export const UPLOAD_ERROR = 'UPLOAD_ERROR';
 export const RESET = 'RESET';
 export const INVALID_TEMPLATE = 'INVALID_TEMPLATE';
+export const REMOVE_INVALID_FILENAME = 'REMOVE_INVALID_FILENAME';
 
 export const startUploads = (fileTypeId, files, tmpFileIds) => ({
   type: START_UPLOADS,
@@ -82,6 +83,12 @@ export const reset = fileTypes => ({
 
 export const invalidTemplate = (fileTypeId, filename) => ({
   type: INVALID_TEMPLATE,
+  fileTypeId,
+  filename,
+});
+
+export const removeInvalidFilename = (fileTypeId, filename) => ({
+  type: REMOVE_INVALID_FILENAME,
   fileTypeId,
   filename,
 });
--- a/indico/modules/events/editing/client/js/components/FileManager/index.jsx
+++ b/indico/modules/events/editing/client/js/components/FileManager/index.jsx
@@ -12,7 +12,7 @@ import PropTypes from 'prop-types';
 import {fromEvent} from 'file-selector';
 import {useDropzone} from 'react-dropzone';
 import {Field} from 'react-final-form';
-import {Icon, Popup, Segment} from 'semantic-ui-react';
+import {Icon, Message, Popup} from 'semantic-ui-react';
 import {TooltipIfTruncated} from 'indico/react/components';
 import {Param, Translate} from 'indico/react/i18n';
 import {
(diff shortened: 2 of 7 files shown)


indico/modules/events/editing/client/js/components/FileManager/actions.js
```diff
@@ -16,6 +16,7 @@ export const CLEAR_DIRTY = 'CLEAR_DIRTY';
 export const UPLOAD_ERROR = 'UPLOAD_ERROR';
 export const RESET = 'RESET';
 export const INVALID_TEMPLATE = 'INVALID_TEMPLATE';
+export const INVALID_FILENAME = 'INVALID_FILENAME';
```
ThiefMaster: `REMOVE_INVALID_FILENAME` or something like that

indico/modules/events/editing/client/js/components/FileManager/index.jsx
```diff
@@ -151,10 +151,14 @@ function FileType({uploadURL, fileType, uploads}) {
                 </Translate>
               }
               trigger={
-                <Segment>
+                <Message
+                  onDismiss={() => {
+                    dispatch(actions.removeInvalidFilename(fileType.id, invalidFile));
+                  }}
```
mic4ael:
```suggestion
                  onDismiss={() => dispatch(actions.removeInvalidFilename(fileType.id, invalidFile))}
```

indico/modules/events/editing/client/js/components/FileManager/index.jsx
```diff
@@ -151,10 +151,14 @@ function FileType({uploadURL, fileType, uploads}) {
                 </Translate>
               }
               trigger={
-                <Segment>
+                <Message
+                  onDismiss={() => {
```
ThiefMaster: `{() => actions....(...)}` doesn't fit?

indico/modules/events/editing/client/js/components/FileManager/reducer.js
```diff
@@ -78,6 +78,18 @@ export default combineReducers({
     if (action.type === actions.RESET) {
       return action.fileTypes;
     }
+    if (action.type === actions.INVALID_FILENAME) {
+      return state.map(fileType => {
+        if (fileType.id === action.fileTypeId) {
+          return {
+            ...fileType,
+            invalidFiles: fileType.invalidFiles.filter(ift => ift !== action.filename),
```
ThiefMaster: s/ift/name/ (earlier i wasn't sure what it was so i just abbreviated whatever long name there was)

indico/modules/events/editing/client/js/components/FileManager/index.jsx
mic4ael: Hm, not sure about this one but you can also omit `{}` which will be the return value for `onDismis`. I don't think `Message` relies on the return value so it would be purely personal preference to change it.

indico/modules/events/editing/client/js/components/FileManager/index.jsx
Leats: That was how I did it at first but prettier is complaining that way

mic4ael: weird.. what did it complain about?

indico/modules/events/editing/client/js/components/FileManager/index.jsx
Leats: With prettier not wanting it in one line (how I had it originally) I think I personally prefer how it is now :)

PR #4368, opened 2020-03-24
Use a proper model for event labels

ThiefMaster:
Get rid of all the dirty workarounds used when implementing the feature in 2.2 without the need for a new alembic revision.

### DIFF
--- a/indico/migrations/versions/20200324_1004_a3295d628e3b_migrate_event_labels_from_settings.py
+++ b/indico/migrations/versions/20200324_1004_a3295d628e3b_migrate_event_labels_from_settings.py
@@ -1,0 +1,77 @@
+"""Migrate event labels from settings
+
+Revision ID: a3295d628e3b
+Revises: 266d78b1c5db
+Create Date: 2020-03-24 10:04:10.926079
+"""
+
+import json
+from collections import defaultdict
+from uuid import uuid4
+
+from alembic import context, op
+
+
+# revision identifiers, used by Alembic.
+revision = 'a3295d628e3b'
+down_revision = '266d78b1c5db'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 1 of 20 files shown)


indico/migrations/versions/20200324_1004_a3295d628e3b_migrate_event_labels_from_settings.py
```diff
@@ -0,0 +1,77 @@
...
+
+def upgrade():
+    if context.is_offline_mode():
+        raise Exception('This upgrade is only possible in online mode')
+    conn = op.get_bind()
+    # label definitions
+    mapping = {}
+    res = conn.execute("SELECT name, value FROM indico.settings WHERE module = 'event_labels'")
+    for uuid, data in res:
+        assert uuid == data['id']
```
mic4ael: Why do you check it here?

ThiefMaster: probably not really needed (since it should never be different) but can't hurt either

mic4ael: Yeah, I was actually thinking when these values can be different.

plourenco:
Most likely related to the previous label issue but I noticed existing label actions are off-margin, maybe we could fix it with this PR?
image

I would suggest:
```
.i-table-widget {
...
  a.action-icon {
        margin-right: 5px;

        &:first-child {
            margin-left: 0;
        }
    }
}
```
Or, for a vertical display:
```
.i-table-widget {
...
  a.action-icon {
        margin-right: 5px;
    }
}
```

ThiefMaster:
Oh thanks, this only happens on Chrome (also on the "external id types" page)... In Firefox it looks fine.

PR #4371, opened 2020-03-25
Add editing module to event cloning

### LINKED ISSUE #4319: Add editing module to event cloning
Useful especially for JACoW, as they are considering having a template event with their basic editing setup and clone it. The idea is to be able to clone editing settings (file types, tags etc.)

Leats:
Closes #4319.

### DIFF
--- a/indico/core/signals/event/core.py
+++ b/indico/core/signals/event/core.py
@@ -43,7 +43,8 @@ the old category is in the `old_parent` kwarg.
 """)
 
 created = _signals.signal('created', """
-Called when a new event is created. The `sender` is the new Event.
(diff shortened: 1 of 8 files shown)


CHANGES.rst
```diff
@@ -58,6 +58,7 @@ Improvements
 - Add event logging to opening/closing registration forms, approval/rejection of
   registrations, and updates to event layout (:issue:`4360`,
   thanks :user:`giusedb` & :user:`omegak`)
+- Add editing module to event cloning (:issue:`4319`)
```
ThiefMaster: We don't need a changelog entry here since the whole editing module is new in 2.3

Issue #4375, opened 2020-03-27
Handle different editable types during submission


ThiefMaster:
- use e.g. a dropdown to select which editable type to submit from the contribution page (if more than one is enabled)
- only allow submitting editables for enabled types

PR #4378, opened 2020-03-30
Associate file types with editable types

### LINKED ISSUE #4373: Associate file types with editable types
- new column + alembic revision
- update filetype-related urls to include the name
- create a default one for each editable type when enabling the event feature

Leats:
Closes #4373

### DIFF
--- a/indico/modules/events/editing/blueprint.py
+++ b/indico/modules/events/editing/blueprint.py
@@ -16,53 +16,53 @@ _bp = IndicoBlueprint('event_editing', __name__, url_prefix='/event/<confId>', t
                       virtual_template_folder='events/editing')
 
 # Frontend
+contrib_prefix = '/contributions/<int:contrib_id>/editing/<any(paper,slides,poster):type>'
 _bp.add_url_rule('/manage/editing/', 'dashboard', frontend.RHEditingDashboard)
 _bp.add_url_rule('/manage/editing/tags', 'manage_tags', frontend.RHManageEditingTags)
-_bp.add_url_rule('/manage/editing/types', 'manage_file_types', frontend.RHManageEditingFileTypes)
-_bp.add_url_rule('/manage/editing/review-conditions', 'manage_review_conditions',
+_bp.add_url_rule('/manage/editing/<any(paper,slides,poster):type>/types', 'manage_file_types',
+                 frontend.RHManageEditingFileTypes)
+_bp.add_url_rule('/manage/editing/review-conditions/<any(paper,slides,poster):type>', 'manage_review_conditions',
                  frontend.RHManageEditingReviewConditions)
-_bp.add_url_rule('/contributions/<int:contrib_id>/editing/<any(paper):type>', 'editable', frontend.RHEditableTimeline)
-_bp.add_url_rule('/contributions/<int:contrib_id>/editing/<any(paper):type>/<int:revision_id>/files.zip',
-                 'revision_files_export', timeline.RHExportRevisionFiles)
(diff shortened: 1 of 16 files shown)


indico/modules/events/editing/blueprint.py
```diff
@@ -16,53 +16,53 @@
                       virtual_template_folder='events/editing')
 
 # Frontend
+contrib_prefix = '/contributions/<int:contrib_id>/editing/<any(paper,slides,poster):type>'
```
panagiotappl: Is it better to leave a space between the commas here? ex: <any(paper, slides, poster):type>
If not, I'll change it in my PR also.

ThiefMaster: IMHO it's more readable without the space in this particular case (just like not having spaces around the `=` of kwargs)

Leats: It seems like other blueprints (for example `indico/modules/rb`) don't have spaces in there as well, so I left them out
I don't know if it's better or not, though

indico/modules/events/editing/controllers/frontend.py
```diff
@@ -51,10 +51,19 @@ def _process(self):
 
 
 class RHManageEditingFileTypes(RHEditingManagementBase):
+    def _process_args(self):
+        RHEditingManagementBase._process_args(self)
+        self.editable_type = request.view_args['type']
```
ThiefMaster: I think it would be more consistent to use `EditableType[request.view_args['type']]` here and below, and then use the `.name` of the enum member when passing it to the template (or inside the template).

indico/modules/events/editing/models/file_types.py
```diff
@@ -19,8 +21,16 @@ class EditingFileType(db.Model):
...
-                {'schema': 'event_editing'})
+        return (
+            db.Index(
+                'ix_uq_file_types_event_id_type_name_lower',
+                cls.event_id,
+                cls.type,
+                db.func.lower(cls.name),
+                unique=True,
+            ),
+            {"schema": "event_editing"},
```
ThiefMaster:
```suggestion
            {'schema': 'event_editing'},
```

indico/modules/events/editing/controllers/frontend.py
panagiotappl: Should we allow accessing/managing an editable type that is not enabled?

ThiefMaster: I'm leaning towards yes, since there's no harm in it - I'd just check it for user/editor-facing endpoints (#4375).

Just like we also allow accessing `/manage/contributions` for a meeting, even though this page is usually not used/shown outside conferences.

panagiotappl: I agree but we should also display it in the editable list even if it is disabled, as the only way to edit the disabled ones would be via typing the url for them.

ThiefMaster: That's a very good point! Maybe showing it but keeping it grayed-out to make it clear it's disabled?

panagiotappl: Yeah, could go in a separate commit/task though

ThiefMaster: Yep, I was about to suggest that you could include it in your current task but I think it makes even more sense to have a separate task for it.

ThiefMaster: Actually, thinking about it again, do we need this? If you enable a type but don't open submission/editing for it nor assign anyone, the type would not show up for users/editors anyway, and thus there's no real need to configure it while it's disabled.

panagiotappl: I thought that one might want to configure them all first before enabling them.

PR #4379, opened 2020-03-30
Add checking of reviewing conditions

### LINKED ISSUE #4302: Make sure ready for review conditions are properly checked before an `Editable` can be reviewed

ParthS007:
closes #4302

### DIFF
--- a/indico/modules/events/client/js/reviewing/components/TimelineHeader.jsx
+++ b/indico/modules/events/client/js/reviewing/components/TimelineHeader.jsx
@@ -10,10 +10,18 @@ import contributionDisplayURL from 'indico-url:contributions.display_contributio
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import {Message, Icon} from 'semantic-ui-react';
 import {Param, Translate} from 'indico/react/i18n';
 import {MathJax} from 'indico/react/components';
 
-export default function TimelineHeader({children, contribution, state, submitter, eventId}) {
+export default function TimelineHeader({
+  children,
+  contribution,
+  state,
+  submitter,
+  eventId,
+  reviewConditionsValid,
+}) {
   return (
     <>
       <div className="submission-title flexrow">
@@ -27,7 +35,13 @@ export default function TimelineHeader({children, contribution, state, submitter
       <div className="paper-public">
         <div className="review-summary flexrow f-a-baseline">
           <div className="review-summary-badge">
-            <div className={`i-tag ${state.cssClass}`}>{state.title}</div>
+            {reviewConditionsValid ? (
+              <div className={`i-tag ${state.cssClass}`}>{state.title}</div>
+            ) : (
+              <div className="i-tag">
+                <Translate>Not Ready</Translate>
+              </div>
+            )}
           </div>
           <div className="review-summary-content f-self-stretch">
             <div>
@@ -47,6 +61,12 @@ export default function TimelineHeader({children, contribution, state, submitter
             </div>
           </div>
         </div>
+        {!reviewConditionsValid && (
+          <Message warning>
+            <Icon name="warning sign" />
+            <Translate>This editable is not fulfilling reviewing conditions.</Translate>
+          </Message>
+        )}
         <div className="review-item-content">{children}</div>
       </div>
     </>
@@ -69,4 +89,9 @@ TimelineHeader.propTypes = {
   }).isRequired,
   eventId: PropTypes.number.isRequired,
   children: PropTypes.node.isRequired,
+  reviewConditionsValid: PropTypes.bool,
+};
+
+TimelineHeader.defaultProps = {
+  reviewConditionsValid: true,
 };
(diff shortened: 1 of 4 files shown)


indico/modules/events/editing/models/editable.py
```diff
@@ -91,3 +92,11 @@ def event(self):
...
         return (self.event.can_manage(user, permission='paper_editing')
                 or self.contribution.is_user_associated(user, check_abstract=True))
+
+    def review_conditions_ok(self):
+        review_conditions = editing_settings.get(self.event, 'review_conditions').values()
+        file_types = [file.file_type_id for file in self.revisions[-1].files]
+        for review_condition in review_conditions:
+            if all(file_type in file_types for file_type in review_condition):
+                return True
+        return False
```
ThiefMaster: `return all(...)` with two loops inside the `all()`. This also ensures that not having any conditions will be considered OK, since right now you;d return False in this case.

indico/modules/events/editing/models/editable.py
```diff
@@ -91,3 +92,11 @@ def event(self):
     def can_comment(self, user):
         return (self.event.can_manage(user, permission='paper_editing')
                 or self.contribution.is_user_associated(user, check_abstract=True))
+
+    def review_conditions_ok(self):
+        review_conditions = editing_settings.get(self.event, 'review_conditions').values()
+        file_types = [file.file_type_id for file in self.revisions[-1].files]
+        for review_condition in review_conditions:
+            if all(file_type in file_types for file_type in review_condition):
```
ThiefMaster: using `ft` as the loop var is more readable here IMO since `file_type` and `file_types` looks very similar.

indico/modules/events/editing/models/editable.py
```diff
@@ -91,3 +92,11 @@ def event(self):
     def can_comment(self, user):
         return (self.event.can_manage(user, permission='paper_editing')
                 or self.contribution.is_user_associated(user, check_abstract=True))
+
+    def review_conditions_ok(self):
+        review_conditions = editing_settings.get(self.event, 'review_conditions').values()
+        file_types = [file.file_type_id for file in self.revisions[-1].files]
```
ThiefMaster: `{...}` since sets are more efficient for this

indico/modules/events/editing/schemas.py
```diff
@@ -138,6 +139,7 @@ class Meta:
     can_create_internal_comments = fields.Function(
         lambda editable, ctx: editable.contribution.event.can_manage(ctx.get('user'), permission='paper_editing')
     )
+    review_conditions_ok = fields.Function(lambda editable, ctx: editable.review_conditions_ok())
```
ThiefMaster: Why not make the method on the Editable a property? That way you can use a simple `fields.Bool()` here

indico/modules/events/editing/client/js/components/Timeline/index.jsx
```diff
@@ -45,6 +45,7 @@ export default function Timeline() {
         state={lastState}
         submitter={timelineBlocks[0].submitter}
         eventId={eventId}
+        reviewConditionsOk={details.reviewConditionsOk}
```
plourenco: Maybe we could name this `validReviewConditions`? Seems more appropriate.

ThiefMaster: `reviewConditionsValid` is better IMO. But thanks for suggesting a better name; I also wasn't a huge fan of the `Ok` but couldn't think of something better on the fly

indico/modules/events/client/js/reviewing/components/TimelineHeader.jsx
```diff
@@ -69,4 +89,5 @@ TimelineHeader.propTypes = {
   }).isRequired,
   eventId: PropTypes.number.isRequired,
   children: PropTypes.node.isRequired,
+  reviewConditionsOk: PropTypes.bool.isRequired,
```
plourenco: This prop is marked as required but not defined in `events/papers/clients/js/components/Paper.jsx`.

ParthS007: Thanks @plourenco for pointing out, I will keep a default value for this Prop.

indico/modules/events/editing/models/editable.py
```diff
@@ -91,3 +92,11 @@ def event(self):
     def can_comment(self, user):
         return (self.event.can_manage(user, permission='paper_editing')
                 or self.contribution.is_user_associated(user, check_abstract=True))
+
+    @property
+    def review_conditions_valid(self):
+        review_conditions = editing_settings.get(self.event, 'review_conditions').values()
+        file_types = {file.file_type_id for file in self.revisions[-1].files}
+        if not file_types:
```
ThiefMaster: wrong check - you need to check `if not review_conditions` here

indico/modules/events/editing/models/editable.py
```diff
@@ -91,3 +92,11 @@ def event(self):
...
+            return True
+        return any(all(ft in file_types for ft in cond) for cond in review_conditions)
```
ThiefMaster: looking at it again i think we can simplify this even more using set logic:

    return any(file_types >= set(cond) for cond in review_conditions)

nothing wrong with keeping the `all()` though - whatever you prefer

ParthS007: Will go for the first one, Looks much simpler and readable.

PR #4383, opened 2020-03-31
Add Workflow section in management sidebar

### LINKED ISSUE #4382: Simplify "organization" event management menu
Call for Abstracts, Call for Papers and Editing could go into a new "Workflows" menu group.

ParthS007:
closes #4382

### DIFF
--- a/indico/modules/events/abstracts/__init__.py
+++ b/indico/modules/events/abstracts/__init__.py
@@ -51,7 +51,7 @@ def _extend_event_management_menu(sender, event, **kwargs):
     if not event.can_manage(session.user) or not AbstractsFeature.is_allowed_for_event(event):
         return
     return SideMenuItem('abstracts', _('Call for Abstracts'), url_for('abstracts.management', event),
-                        section='organization')
+                        section='workflows')
 
 
 @signals.event.get_feature_definitions.connect
@@ -141,8 +141,7 @@ def _get_notification_placeholders(sender, **kwargs):
 @signals.menu.items.connect_via('event-editing-sidemenu')
 def _extend_editing_menu(sender, event, **kwargs):
     if event.has_feature('abstracts'):
-        yield SideMenuItem('abstracts', _('Call for Abstracts'), url_for('abstracts.call_for_abstracts', event),
-                           section='organization')
+        yield SideMenuItem('abstracts', _('Call for Abstracts'), url_for('abstracts.call_for_abstracts', event))
 
 
 @signals.event.sidemenu.connect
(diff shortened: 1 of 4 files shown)


indico/modules/events/abstracts/__init__.py
```diff
@@ -142,7 +142,7 @@ def _get_notification_placeholders(sender, **kwargs):
 def _extend_editing_menu(sender, event, **kwargs):
     if event.has_feature('abstracts'):
         yield SideMenuItem('abstracts', _('Call for Abstracts'), url_for('abstracts.call_for_abstracts', event),
-                           section='organization')
+                           section='workflows')
```
ThiefMaster: This one is a separate menu. Do we even need the section here? cc @panagiotappl

indico/modules/events/papers/__init__.py
```diff
@@ -44,7 +44,7 @@ def _extend_editing_menu(sender, event, **kwargs):
     if (has_contributions_with_user_paper_submission_rights(event, session.user) or
             event.cfp.is_staff(session.user)):
         return SideMenuItem('papers', _('Call for Papers'), url_for('papers.call_for_papers', event),
-                            section='organization')
+                            section='workflows')
```
ThiefMaster: Same here, most likely the section can be removed from all menu items that are returned for `event-editing-sidemenu`

indico/modules/events/management/__init__.py
```diff
@@ -17,7 +17,8 @@
 
 @signals.menu.sections.connect_via('event-management-sidemenu')
 def _sidemenu_sections(sender, **kwargs):
-    yield SideMenuSection('organization', _("Organization"), 50, icon='list', active=True)
+    yield SideMenuSection('organization', _("Organization"), 60, icon='list', active=True)
```
ThiefMaster: This change affects the order of other items as well (protection has 50 as well so with making organization 60 it now shows up before that item).

Keep 50 for this one and use e.g. 45 for workflows (or 40 and renumber the others below ccordingly)

indico/modules/events/abstracts/__init__.py
ParthS007: I don't think we need the section here, I will remove it fully.

panagiotappl: Yeah, we don't at least for now that it's so small.

indico/modules/events/management/__init__.py
ParthS007: I have set `60` for `organization`, `55` for `workflows` and `70` for `protection`. If I don't change the `protection` here, I will have to change the `room booking` one.

PR #4386, opened 2020-04-02
Use react for editing management dashboard

ThiefMaster:
And while at it:

- use react/-router for editing management
- move related components into subdirectories
- get rid of redux in condition manager (context is more than enough for static data)
- use new Section component to render sections on both the dashboard and the editable type page


indico/modules/events/editing/client/js/components/EditingManagement.jsx
```diff
@@ -0,0 +1,39 @@
...
+import {BrowserRouter as Router, Route, Switch} from 'react-router-dom';
+import {ManagementPageBackButton, ManagementPageSubTitle} from 'indico/react/components';
+import {Translate} from 'indico/react/i18n';
+
+import EditingManagementDashboard from './EditingManagementDashboard';
+import TagManager from './TagManager';
+
+export default function EditingManagement({eventId}) {
+  return (
+    <Router>
```
panagiotappl, line 22: Wouldn't it be nice if we also included each editable type page in the routes and not reload for those either?

indico/modules/events/editing/client/js/components/Section.jsx
```diff
@@ -0,0 +1,29 @@
...
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function Section({icon, label, description, children}) {
+  return (
+    <div className="section">
+      <span className={`icon icon-${icon}`} style={{alignSelf: 'center'}} />
```
panagiotappl, line 14: I don't think this style is needed anymore after the latest commit to center all icons

indico/modules/events/editing/client/js/components/EditingManagement.jsx
ThiefMaster: yeah i'm looking into that as well; should be easily doable

indico/modules/events/editing/client/js/components/Section.jsx
ThiefMaster: indeed, removed it

indico/modules/events/editing/client/js/components/EditingManagement.jsx
ThiefMaster: "should be easily doable".. famous last words!

anyway, it's done now, and the PR suddenly became much bigger 🙈

indico/modules/events/editing/client/js/components/ReviewConditions/index.jsx
```diff
@@ -0,0 +1,40 @@
...
+    camelize: true,
+    trigger: [eventId, type],
+  });
+
+  return (
+    <>
+      <EditingTypeSubPageNav title={Translate.string('Review conditions')} />
+      {fileTypes !== null && (
+        <ReviewConditionsContext.Provider value={{eventId, editableType: type, fileTypes}}>
+          <ReviewConditionsManager eventId={eventId} fileTypes={fileTypes} editableType={type} />
```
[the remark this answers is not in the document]
ThiefMaster: I thought the exact same thing when encountering it, but didn't want to spend more time on refactoring this part on friday evening ;)

indico/modules/events/editing/client/js/components/EditingTagManagement.jsx
```diff
@@ -172,3 +178,14 @@ export default function TagManager({eventId}) {
 TagManager.propTypes = {
   eventId: PropTypes.number.isRequired,
 };
+
+export default function EditingTagManagement() {
```
plourenco, line 182: Why not just editing TagManager and adding
```
const eventId = useNumericParam('confId');
...
<ManagementPageBackButton url={dashboardURL({confId: eventId})} />
<ManagementPageSubTitle title={Translate.string('Tags')} />
...
```

indico/modules/events/editing/client/js/components/EditableTypeDashboard.jsx
```diff
@@ -0,0 +1,102 @@
...
+          description={Translate.string('Configure file types')}
+        >
+          <Link className="i-button icon-settings" to={manageFileTypesURL({confId: eventId, type})}>
+            <Translate>Configure</Translate>
+          </Link>
+        </Section>
+        <Section
+          icon="equalizer"
+          label={Translate.string('Ready for review conditions')}
+          description={Translate.string('Configure conditions for reviewing<')}
```
plourenco, line 65:
```suggestion
          description={Translate.string('Configure conditions for reviewing')}
```

indico/modules/events/editing/client/js/components/EditingTagManagement.jsx
ThiefMaster: The problem of doing this in `TagManager` is that we return `null` or a `Loader` in some cases - but we always need to render those two components to have the correct title/subtitle on the page. So having a separate component for it is easier.

Also, it separates the actual tag manager from setting up the surrounding page.

PR #4387, opened 2020-04-06
Check whether reviewing is actually possible

### LINKED ISSUE #4367: Reviewing deadlines are not taken into account
**Describe the bug**
Once an event manager enforces a reviewing deadline it is still possible to review the assigned papers.

**To Reproduce**
1. Define a reviewing deadline
2. Make sure the deadline is enforced

**Expected behavior**
It should be impossible to review a paper.

mic4ael:
closes #4367

### DIFF
--- a/indico/modules/events/papers/models/revisions.py
+++ b/indico/modules/events/papers/models/revisions.py
@@ -185,10 +185,15 @@ class PaperRevision(ProposalRevisionMixin, RenderModeMixin, db.Model):
 
     def get_reviewed_for_groups(self, user, include_reviewed=False):
         from indico.modules.events.papers.models.reviews import PaperTypeProxy
-        reviewed_for = {x.type for x in self.reviews if x.user == user} if include_reviewed else set()
-        if self.paper.cfp.content_reviewing_enabled and user in self.paper.cfp.content_reviewers:
+        from indico.modules.events.papers.util import is_type_reviewing_possible
+
+        cfp = self.paper.cfp
+        reviewed_for = set()
+        if include_reviewed:
+            reviewed_for = {x.type for x in self.reviews if x.user == user and is_type_reviewing_possible(cfp, x.type)}
+        if is_type_reviewing_possible(cfp, PaperReviewType.content) and user in self.paper.cfp.content_reviewers:
             reviewed_for.add(PaperReviewType.content)
-        if self.paper.cfp.layout_reviewing_enabled and user in self.paper.cfp.layout_reviewers:
+        if is_type_reviewing_possible(cfp, PaperReviewType.layout) and user in self.paper.cfp.layout_reviewers:
             reviewed_for.add(PaperReviewType.layout)
         return set(map(PaperTypeProxy, reviewed_for))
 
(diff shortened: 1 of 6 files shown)


indico/modules/events/papers/util.py
```diff
@@ -83,3 +85,14 @@ def get_user_submittable_contributions(event, user):
...
+
+def is_type_reviewing_possible(cfp, review_type):
+    if review_type == PaperReviewType.content:
+        return (cfp.content_reviewing_enabled
+                and (not cfp.content_reviewer_deadline_enforced or cfp.content_reviewer_deadline > now_utc()))
+    elif review_type == PaperReviewType.layout:
+        return (cfp.layout_reviewing_enabled
+                and (not cfp.layout_reviewer_deadline_enforced or cfp.layout_reviewer_deadline > now_utc()))
+    else:
+        return False
```
ThiefMaster, line 98: is that's never supposed to happen, better raise an exception.. `ValueError('invalid review type')` maybe

indico/modules/events/papers/models/revisions.py
```diff
@@ -185,10 +185,15 @@ def get_reviews(self, group=None, user=None):
...
     def get_reviewed_for_groups(self, user, include_reviewed=False):
         from indico.modules.events.papers.models.reviews import PaperTypeProxy
-        reviewed_for = {x.type for x in self.reviews if x.user == user} if include_reviewed else set()
-        if self.paper.cfp.content_reviewing_enabled and user in self.paper.cfp.content_reviewers:
+        from indico.modules.events.papers.util import is_type_reviewing_possible
+
+        cfp = self.paper.cfp
+        reviewed_for = set()
+        if include_reviewed:
+            reviewed_for = {x.type for x in self.reviews if x.user == user and is_type_reviewing_possible(cfp, x.type)}
```
ThiefMaster, line 193: Not sure how exactly this function i used, but if it shows the reviewing types in which the user has already reviewed checking if they are still within the deadline seems incorrect here.

mic4ael: This is the list of available groups (content, layout). So user will be available to review in any of the groups that are in this set. I don't want to include any of these even if there is an existing review.

ThiefMaster: the name and iteration over `self.reviews` made me assume it's also used to show the groups in which you *already* reviewed

mic4ael: As far as I can tell there is another place where we return the list of groups that had been reviewed.

```python
    def get_reviewer_render_data(self, user):
        groups = self.get_reviewed_for_groups(user, include_reviewed=True)
        reviews = {x.group: x for x in self.get_reviews(user=user)}
        reviewed_groups = {x.group for x in reviews.itervalues()}
        missing_groups = groups - reviewed_groups
        return {'groups': groups,
                'missing_groups': missing_groups,
                'reviewed_groups': reviewed_groups,
                'reviews': reviews}
```

ThiefMaster: OK, I guess you tested that it works correctly anyway ;)

mic4ael: Hah, of course as always :D I hope I didn't miss anything.

PR #4388, opened 2020-04-06
Move review conditions to SQLAlchemy model

### LINKED ISSUE #4385: Move ready-to-review conditions to an SQLAlchemy model
Especially after #4372 using the SettingsProxy for this will be somewhat ugly (one key for each editable type). It'd be nicer to have a table for this, similar to how we store e.g. FileTypes.

The actual rules for each condition set could still be stored in a single column (`jsonb` or `int[]` array), but adding a many-to-many mapping table would probably be cleaner since that way we ensure that there are no broken references (even though we disallow deleting file types used in rules). It also makes saving the rules easier since you can just assign the file types to the relationship.

Leats:
Closes #4385

### DIFF
--- a/indico/modules/events/editing/clone.py
+++ b/indico/modules/events/editing/clone.py
@@ -7,17 +7,13 @@
 
 from __future__ import unicode_literals
 
-from collections import OrderedDict
-from uuid import uuid4
-
 from indico.core.db import db
 from indico.core.db.sqlalchemy.util.models import get_simple_column_attrs
 from indico.core.db.sqlalchemy.util.session import no_autoflush
 from indico.modules.events.cloning import EventCloner
-from indico.modules.events.editing.models.editable import EditableType
 from indico.modules.events.editing.models.file_types import EditingFileType
+from indico.modules.events.editing.models.review_conditions import EditingReviewCondition
 from indico.modules.events.editing.models.tags import EditingTag
-from indico.modules.events.editing.settings import editing_settings
 from indico.modules.events.models.events import EventType
 from indico.util.i18n import _
 
@@ -54,16 +50,11 @@ class EditingSettingsCloner(EventCloner):
             filetype.populate_from_attrs(old_filetype, attrs)
             new_event.editing_file_types.append(filetype)
             db.session.flush()
-            self._filetype_map[old_filetype.id] = filetype.id
+            self._filetype_map[old_filetype] = filetype
 
     def _clone_review_conditions(self, new_event):
-        for type_ in EditableType:
-            review_conditions = editing_settings.get(self.old_event, type_.name + '_review_conditions')
-            new_conditions = OrderedDict(
-                self._build_review_conditions(new_event, cond)
-                for cond in review_conditions.viewvalues()
-            )
-            editing_settings.set(new_event, type_.name + '_review_conditions', new_conditions)
-
-    def _build_review_conditions(self, new_event, value):
-        return unicode(uuid4()), [self._filetype_map[filetype_id] for filetype_id in value]
+        old_conditions = EditingReviewCondition.query.with_parent(self.old_event).all()
+        for condition in old_conditions:
+            new_filetypes = {self._filetype_map[ft] for ft in condition.file_types}
+            new_condition = EditingReviewCondition(type=condition.type, file_types=new_filetypes)
+            new_event.editing_review_conditions.append(new_condition)
--- a/indico/modules/events/editing/controllers/backend/management.py
+++ b/indico/modules/events/editing/controllers/backend/management.py
@@ -7,19 +7,18 @@
 
 from __future__ import unicode_literals
 
-from uuid import uuid4
-
 from flask import jsonify, request
-from werkzeug.exceptions import NotFound
 
 from indico.core.errors import UserValueError
 from indico.modules.events.editing.controllers.base import RHEditingManagementBase
 from indico.modules.events.editing.models.editable import EditableType
 from indico.modules.events.editing.models.file_types import EditingFileType
+from indico.modules.events.editing.models.review_conditions import EditingReviewCondition
 from indico.modules.events.editing.models.revision_files import EditingRevisionFile
 from indico.modules.events.editing.models.tags import EditingTag
-from indico.modules.events.editing.operations import (create_new_file_type, create_new_tag, delete_file_type,
-                                                      delete_tag, update_file_type, update_tag)
+from indico.modules.events.editing.operations import (create_new_file_type, create_new_review_condition, create_new_tag,
+                                                      delete_file_type, delete_review_condition, delete_tag,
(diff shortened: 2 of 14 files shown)


indico/modules/events/editing/models/file_types.py
```diff
@@ -83,6 +83,17 @@ def __table_args__(cls):
             lazy=True
         )
     )
+    review_conditions = db.relationship(
```
ThiefMaster, line 86: I'd move the relationship into the `EditingReviewCondition` model - having it on the other side doesn't really make a technical difference, but IMO it's more closely related to the condition rather than the file type ("mapping conditions to file types" and not "mapping file types to conditions")

indico/modules/events/editing/models/file_types.py
```diff
@@ -83,6 +83,17 @@ def __table_args__(cls):
...
+        'EditingReviewCondition',
+        secondary='event_editing.review_condition_file_types',
+        collection_class=set,
+        lazy=True,
+        backref=db.backref(
+            'file_types',
+            collection_class=set,
+            lazy=True
```
ThiefMaster, line 94: The number of queries on the file type management page increases with each condition added. Not a big problem since we don't have *that*many conditions, but better make this one `lazy=False` (if you end up moving the relationship to the other side, this will then be in the "outer" part, not inside the backref)

indico/modules/events/editing/clone.py
```diff
@@ -54,16 +50,13 @@ def _clone_filetypes(self, new_event):
...
-            )
-            editing_settings.set(new_event, type_.name + '_review_conditions', new_conditions)
-
-    def _build_review_conditions(self, new_event, value):
-        return unicode(uuid4()), [self._filetype_map[filetype_id] for filetype_id in value]
+        old_conditions = (EditingReviewCondition.query.with_parent(self.old_event).all())
+        for condition in old_conditions:
+            new_filetypes = set()
+            for file_type in condition.file_types:
+                new_filetypes.add(self._filetype_map[file_type])
```
ThiefMaster, line 60:
```python
new_filetypes = {self._filetype_map[ft] for ft in condition.file_types}
```

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -108,37 +107,36 @@ def _process_args(self):
         self.editable_type = EditableType[request.view_args['type']]
 
     def _process_GET(self):
-        return jsonify(editing_settings.get(self.event, self.editable_type.name + '_review_conditions').items())
+        query = (
+            EditingReviewCondition.query.with_parent(self.event)
+            .filter_by(type=self.editable_type)
+            .order_by(EditingReviewCondition.id)
+            .all()
```
ThiefMaster, line 114: either remove the `.all()` or don't name the variable you are assigning it to `query`

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -108,37 +107,36 @@ def _process_args(self):
...
+        )
+        return jsonify([[cond.id, sorted([ft.id for ft in cond.file_types])] for cond in query])
```
ThiefMaster, line 116: `sorted(...)` instead of `sorted([...])` (internally sorted() already creates a list when passing a generator)

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -108,37 +107,36 @@ def _process_args(self):
...
     def _process_args(self):
         RHEditingManagementBase._process_args(self)
         self.editable_type = EditableType[request.view_args['type']]
-        uuid = request.view_args['uuid']
-        conditions = editing_settings.get(self.event, self.editable_type.name + '_review_conditions')
-        self.uuid = unicode(uuid)
-        if self.uuid not in conditions:
-            raise NotFound
+        id_ = request.view_args['condition_id']
+        self.condition = (EditingReviewCondition.query.with_parent(self.event).filter_by(id=id_).first_or_404())
```
ThiefMaster, line 130: no need for the `()` around the whole thing since it's all on one line.

alternatively, go for this instead:

```python
        self.condition = (EditingReviewCondition.query
                          .with_parent(self.event)
                          .filter_by(id=request.view_args['condition_id'])
                          .first_or_404())
```

indico/modules/events/editing/models/review_conditions.py
```diff
@@ -0,0 +1,82 @@
...
+            lazy=True
+        )
+    )
+    file_types = db.relationship(
+        'EditingFileType',
+        secondary='event_editing.review_condition_file_types',
+        collection_class=set,
+        lazy=False,
+        backref=db.backref(
+            'editing_review_conditions',
```
ThiefMaster, line 50: `review_conditions` is enough here (since it's on a model specific to editing, unlike the backref on event where your name is much better)

indico/modules/events/editing/operations.py
```diff
@@ -264,3 +265,21 @@ def update_file_type(file_type, **data):
 def delete_file_type(file_type):
     logger.info('File type %r deleted by %r', file_type, session.user)
     db.session.delete(file_type)
+
+
+def create_new_review_condition(event, editable_type, file_types):
+    review_condition = EditingReviewCondition(event=event, type=editable_type, file_types=file_types)
+    db.session.flush()
+    logger.info('Review condition %r created by %r', repr(review_condition), session.user)
```
ThiefMaster, line 273: `%r` already uses the repr, so if you call `repr()` again it will look *quite* ugly

indico/modules/events/editing/operations.py
```diff
@@ -264,3 +265,21 @@ def update_file_type(file_type, **data):
...
+
+
+def update_review_condition(condition, file_types):
+    condition.file_types = file_types
+    db.session.flush()
+    logger.info('Review condition %r updated by %r', repr(condition), session.user)
+
+
+def delete_review_condition(condition):
+    ('Review condition %r deleted by %r', repr(condition), session.user)
```
ThiefMaster, line 284: missing call 🙈

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -88,8 +87,8 @@ def _process_DELETE(self):
         if EditingRevisionFile.query.with_parent(self.file_type).has_rows():
             raise UserValueError(_('Cannot delete file type which already has files'))
 
-        review_conditions = editing_settings.get(self.event, self.editable_type.name + '_review_conditions')
-        if any(self.file_type.id in cond for cond in review_conditions.values()):
+        review_conditions = EditingReviewCondition.query.with_parent(self.event).filter_by(type=self.editable_type)
+        if any(self.file_type in review_conditions.file_types):
```
ThiefMaster, line 91: you can simplify this a lot now by checking `if self.file_type.review_conditions`

PR #4389, opened 2020-04-06
Fix alignment on management pages

### LINKED ISSUE #4299: Fix alignment on management pages
This is especially bad and noticeable in category management with the actions next to the title:

image

...but also less strongly in event management:

image

---

This is how it looked like in 2.2:

image

image

---

The tricky part is to make sure it doesn't start pushing down the menu in event management, otherwise adding a `margin-top` would be enough...

ParthS007:
closes #4299 
Screenshot from 2020-04-06 14-18-57

### DIFF
--- a/indico/web/client/styles/base/_layout.scss
+++ b/indico/web/client/styles/base/_layout.scss
@@ -97,7 +97,7 @@ body > * {
 
         .action-menu {
             flex-shrink: 0;
-            margin-top: 5px;
+            margin-top: -0.5%;
         }
     }
 
--- a/indico/web/client/styles/base/_pages.scss
+++ b/indico/web/client/styles/base/_pages.scss
@@ -102,9 +102,11 @@
 
     color: $banner-color;
     max-width: 800px;
+    margin-top: 10px;
 
     &.full-width {
         max-width: none;
+        margin-bottom: -10%;
 
         .title {
             flex-grow: 1;


ThiefMaster:
Unfortunately the change results in the menu being pushed further down in case of a subtitle (like the event creator information in case of event management) - ie exactly what is mentioned in the original issue:

> The tricky part is to make sure it doesn't start pushing down the menu in event management, otherwise adding a margin-top would be enough...

mic4ael:
Hey, I don't know what you had done before but what about just removing the `margin` on the `.group` element?
```diff
diff --git i/indico/web/client/styles/base/_layout.scss w/indico/web/client/styles/base/_layout.scss
index 2db273434f..ee2f817f30 100644
--- i/indico/web/client/styles/base/_layout.scss
+++ w/indico/web/client/styles/base/_layout.scss
@@ -107,7 +107,6 @@ body > * {
         margin-right: 30px;

         .group:first-child {
-            margin-top: 5px;
             margin-bottom: 23px;

             a {
```

ParthS007:
Removing the `margin-top` from here will move the menu column more upwards, in this issue, I am trying to keep `switch display button` and title on the same baseline.

mic4ael:
> Removing the `margin-top` from here will move the menu column more upwards, in this issue, I am trying to keep `switch display button` and title on the same baseline.

Yeah, ofc. But is that really a problem? They will still be on the same baseline.

ThiefMaster:
IMHO it looks bad if there are different spacings above the menu in event/category management

mic4ael:
Ahh.. my bad.. I see now what you mean. I will have a quick look into that

mic4ael:
So, the misalignment is mainly caused by SemanticUI and their CSS rules changing `line-height`. Removing it from the affected nodes fixes the misalignment. But yeah, that will not fix the issue with the menus

ThiefMaster:
Layout-wise it looks fine now. Unless there's a much better solution or any problems caused by this one (I didn't find any) I'm inclined to merge it the way it is now...

mic4ael:
> Layout-wise it looks fine now. Unless there's a much better solution or any problems caused by this one (I didn't find any) I'm inclined to merge it the way it is now...

Looks fine to me as well. I just wanted to mention what actually caused the issue.

PR #4394, opened 2020-04-07
Add team manager

### LINKED ISSUE #4377: Editing setup: teams
- Split `EditableEditorPermission` into three separate permissions for the editable types
- Add a page for assigning the editors for the event (can probably use one of our ACL widgets that lets you assign permissions? maybe a filtered version of the one we use on the event protection page that only shows editing-related permissions?); but better ideas are welcome of course!

panagiotappl:
closes #4377 
image

### DIFF
--- a/indico/modules/events/editing/blueprint.py
+++ b/indico/modules/events/editing/blueprint.py
@@ -53,6 +53,8 @@ _bp.add_url_rule('/editing/api/enabled-editable-types', 'api_enabled_editable_ty
                  management.RHEnabledEditableTypes, methods=('GET', 'POST'))
 _bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/editable-assignment/self-assign-enabled',
                  'api_self_assign_enabled', management.RHEditableSetSelfAssign, methods=('GET', 'PUT', 'DELETE'))
+_bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/principals', 'api_editable_type_principals',
+                 management.RHEditableTypePrincipals, methods=('GET', 'POST'))
 
 # Contribution/revision-level APIs
 contrib_api_prefix = '/api' + contrib_prefix
--- a/indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
+++ b/indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
@@ -10,7 +10,7 @@ import manageFileTypesURL from 'indico-url:event_editing.manage_file_types';
 import manageReviewConditionsURL from 'indico-url:event_editing.manage_review_conditions';
 import selfAssignURL from 'indico-url:event_editing.api_self_assign_enabled';
 
-import React from 'react';
+import React, {useState} from 'react';
 import {useParams, Link} from 'react-router-dom';
 import {Checkbox, Loader} from 'semantic-ui-react';
 
@@ -20,12 +20,14 @@ import {useNumericParam} from 'indico/react/util/routing';
 import {useTogglableValue} from 'indico/react/hooks';
 import {EditableTypeTitles} from '../../models';
 import Section from '../Section';
+import TeamManagerModal from './TeamManagerModal';
 
 import './EditableTypeDashboard.module.scss';
 
 export default function EditableTypeDashboard() {
   const eventId = useNumericParam('confId');
   const {type} = useParams();
+  const [editorManagerVisible, setEditorManagerVisible] = useState(false);
 
   const [
     selfAssignEnabled,
@@ -95,9 +97,15 @@ export default function EditableTypeDashboard() {
               <a className="i-button icon-mail">
                 <Translate>Contact</Translate>
               </a>
-              <a className="i-button icon-users">
+              <a className="i-button icon-users" onClick={() => setEditorManagerVisible(true)}>
                 <Translate>Manage team</Translate>
               </a>
+              {editorManagerVisible && (
+                <TeamManagerModal
+                  editableType={type}
+                  onClose={() => setEditorManagerVisible(false)}
+                />
+              )}
             </Section>
             <Section
               icon="copy1"
--- a/indico/modules/events/editing/client/js/management/editable_type/TeamManagerModal.jsx
+++ b/indico/modules/events/editing/client/js/management/editable_type/TeamManagerModal.jsx
@@ -1,0 +1,95 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2020 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import principalsURL from 'indico-url:event_editing.api_editable_type_principals';
+
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Form as FinalForm} from 'react-final-form';
+import {Button, Form, Loader, Message, Modal} from 'semantic-ui-react';
+import {Translate} from 'indico/react/i18n';
+import {FinalPrincipalList} from 'indico/react/components';
+import {
+  getChangedValues,
+  handleSubmitError,
+  FinalSubmitButton,
+  FinalUnloadPrompt,
+} from 'indico/react/forms';
+import {indicoAxios} from 'indico/utils/axios';
+import {useFavoriteUsers, useIndicoAxios} from 'indico/react/hooks';
+import {useNumericParam} from 'indico/react/util/routing';
+
+export default function TeamManager({editableType, onClose}) {
+  const eventId = useNumericParam('confId');
+  const favoriteUsersController = useFavoriteUsers();
+
+  const {data: principals, loading: isLoadingPrincipals} = useIndicoAxios({
+    url: principalsURL({confId: eventId, type: editableType}),
+    trigger: [eventId, editableType],
+  });
+
+  const handleSubmit = async (data, form) => {
+    const changedValues = getChangedValues(data, form);
+    try {
+      await indicoAxios.post(principalsURL({confId: eventId, type: editableType}), changedValues);
+    } catch (error) {
+      return handleSubmitError(error);
+    }
+    onClose();
+  };
+
+  if (isLoadingPrincipals) {
+    return <Loader active />;
+  } else if (!principals) {
+    return null;
+  }
+
+  return (
+    <div style={{padding: '30'}}>
+      <FinalForm
+        onSubmit={handleSubmit}
+        initialValues={{principals}}
+        initialValuesEqual={_.isEqual}
+        subscription={{}}
+      >
+        {fprops => (
+          <Modal onClose={onClose} size="small" closeOnDimmerClick={false} open>
+            <Modal.Header content={Translate.string('Set editors')} />
+            <Modal.Content>
+              <Form id="team-manager-form" onSubmit={fprops.handleSubmit}>
+                <FinalUnloadPrompt />
+                <Message>
+                  <FinalPrincipalList
+                    name="principals"
+                    withGroups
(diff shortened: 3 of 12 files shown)


indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -145,3 +146,25 @@ def _process_POST(self, editable_types):
...
+        self.editable_type = EditableType[request.view_args['type']]
+
+    def _process_GET(self):
+        principals = {p.principal for p in self.event.acl_entries
+                      if p.has_management_permission(self.editable_type.name + '_editing', explicit=True)}
+        return EditableTypePrincipalsSchema().jsonify({'principals': principals})
+
+    @use_rh_kwargs(EditableTypePrincipalsSchema)
+    def _process_POST(self, principals):
+        permission_name = self.editable_type.name + '_editing'
```
ThiefMaster, line 163: I'd prefer a dict mapping editable types to permission names. Otherwise searching for the permission name won't find this.

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -145,3 +146,25 @@ def _process_POST(self, editable_types):
...
+
+class RHEditableTypePrincipals(RHEditingManagementBase):
+    def _process_args(self):
+        RHEditingManagementBase._process_args(self)
+        self.editable_type = EditableType[request.view_args['type']]
+
+    def _process_GET(self):
+        principals = {p.principal for p in self.event.acl_entries
+                      if p.has_management_permission(self.editable_type.name + '_editing', explicit=True)}
+        return EditableTypePrincipalsSchema().jsonify({'principals': principals})
```
ThiefMaster, line 159: I'd just return a list here:

    return jsonify([p.principal.identifier for p in principals])

indico/modules/events/editing/client/js/management/editable_type/TeamManagerModal.jsx
```diff
@@ -0,0 +1,93 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2020 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+import principalsURL from 'indico-url:event_editing.api_editable_type_principals';
```
ThiefMaster, line 7: missing empty line above this

indico/modules/events/editing/client/js/management/editable_type/TeamManagerModal.jsx
```diff
@@ -0,0 +1,93 @@
...
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+import {useFavoriteUsers, useIndicoAxios} from 'indico/react/hooks';
+import {useNumericParam} from 'indico/react/util/routing';
+
+export default function TeamManager({editableType, onClose}) {
+  const eventId = useNumericParam('confId');
+
+  const {data: principals, loading: isLoadingPrincipals} = useIndicoAxios({
+    url: principalsURL({confId: eventId, type: editableType}),
+    trigger: eventId,
```
ThiefMaster, line 26: `trigger: [eventId, editableType]`

While we don't ever expect this to happen, the editable type changing is something that would require a re-fetch, so like in the dependency list of other hooks it needs to be included here.

indico/modules/events/editing/client/js/management/editable_type/TeamManagerModal.jsx
```diff
@@ -0,0 +1,93 @@
...
+          </Modal>
+        )}
+      </FinalForm>
+    </div>
+  );
+}
+
+TeamManager.propTypes = {
+  editableType: PropTypes.string.isRequired,
+  onClose: PropTypes.func,
```
plourenco, line 88: I wouldn't set this as null in the default props, since we end up calling the function without checking it. IMO, I would suggest setting it as required. As we do it in other similar modals.

```suggestion
  onClose: PropTypes.func.isRequired,
```

indico/modules/events/editing/controllers/backend/management.py
panagiotappl: Wouldn't this break in case of `CategoryRole` for example?

ThiefMaster: No, any principal class has an `identifier` property that returns a string like `User:123` or `CategoryRole:123`.

This is actually what PrincipalList does when serializing:

```python
    def _serialize(self, value, attr, obj):
        return [x.identifier for x in value]
```

panagiotappl: Ah, I see. I was already accessing `p.principal`

indico/modules/events/editing/blueprint.py
```diff
@@ -53,6 +53,8 @@
                  management.RHEnabledEditableTypes, methods=('GET', 'POST'))
 _bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/editable-assignment/self-assign-enabled',
                  'api_self_assign_enabled', management.RHEditableSetSelfAssign, methods=('GET', 'PUT', 'DELETE'))
+_bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/principals', 'api_editable_type_principals',
+                 management.RHEditableTypePrincipals, methods=('GET', 'POST'))
```
ThiefMaster, line 57: s/principals/editors/?

indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
```diff
@@ -20,12 +20,14 @@ import {useNumericParam} from 'indico/react/util/routing';
...
 import {EditableTypeTitles} from '../../models';
 import Section from '../Section';
+import TeamManagerModal from './TeamManagerModal';
 
 import './EditableTypeDashboard.module.scss';
 
 export default function EditableTypeDashboard() {
   const eventId = useNumericParam('confId');
   const {type} = useParams();
+  const [modalVisible, setModalVisible] = useState(false);
```
ThiefMaster, line 30: `[editorManagerVisible, setEditorManagerVisible]`?

indico/modules/events/editing/client/js/management/editable_type/TeamManagerModal.jsx
```diff
@@ -0,0 +1,90 @@
...
+    url: principalsURL({confId: eventId, type: editableType}),
+    trigger: [eventId, editableType],
+  });
+
+  const handleSubmit = async (data, form) => {
+    const changedValues = getChangedValues(data, form);
+    try {
+      await indicoAxios.post(principalsURL({confId: eventId, type: editableType}), changedValues);
+    } catch (error) {
+      handleAxiosError(error);
```
ThiefMaster, line 35: return handleSubmitError(error);

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -157,4 +158,22 @@ def _process_PUT(self):
 
     def _process_DELETE(self):
         editable_type_settings[self.editable_type].set(self.event, 'self_assign_allowed', False)
+
+
+class RHEditableTypePrincipals(RHEditableTypeManagementBase):
+    def _process_GET(self):
+        principals = {p.principal for p in self.event.acl_entries
```
ThiefMaster, line 165: `= [p.principal.identifier for ...]` and then just `return jsonify(principals)` below (or even merge everything in a single statement if it doesn't look bad)

indico/modules/events/management/controllers/protection.py
```diff
@@ -114,13 +114,11 @@ def _process(self):
...
     ALLOW_LOCKED = True
 
-    def _process_args(self):
-        RHManageEventBase._process_args(self)
-        args = parser.parse({
-            'values': PrincipalDict(allow_groups=True, allow_external_users=True, allow_event_roles=True,
-                                    allow_category_roles=True, event_id=self.event.id, missing={})
-        })
-        self.values = args['values']
+    @use_rh_kwargs({'values': PrincipalDict(allow_groups=True, allow_external_users=True, allow_event_roles=True,
```
ThiefMaster, line 117: I'd wrap after `{` since we usually do this in other places where we have `use_*()`

indico/modules/events/editing/blueprint.py
panagiotappl: Is it really wrong to use `principals` here?

ThiefMaster: just a suggestion ;)

I tend to use "principals" when there's no more meaningful word available (let's say for a general ACL where you can assign different permissions), but go for the explicit term ("editors", etc) if one is available

panagiotappl: I don't want to change the naming of this everywhere across the PR, I think `principals` is fine.

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -157,4 +158,21 @@ def _process_PUT(self):
 
     def _process_DELETE(self):
         editable_type_settings[self.editable_type].set(self.event, 'self_assign_allowed', False)
+
+
+class RHEditableTypePrincipals(RHEditableTypeManagementBase):
+    def _process_GET(self):
+        return jsonify([p.principal.identifier for p in self.event.acl_entries
+                       if p.has_management_permission(self.editable_type.name + '_editing', explicit=True)])
```
ThiefMaster, line 166: an you use the mapping here as well?

ThiefMaster:
When you search for `'paper_editing'` you'll find some places where we have this permission name hardcoded; I think most/all of them need to be updated to take the editable type into account. Since you need the permission name for an editable type in more places, I'd add a new property `editor_permission` to the `EditableType` enum (similar to how the `title` property on the RichEnum base class is implemented) so you can just do `self.editable_type.editor_permission` etc

indico/modules/events/editing/models/editable.py
```diff
@@ -17,10 +17,15 @@
 
 class EditableType(RichIntEnum):
     __titles__ = [None, _("Paper"), _("Slides"), _("Poster")]
```
mic4ael, line 19: Why `"` instead of `'`? We usually use single quotes and probably it would be best to be consistent

indico/modules/events/editing/models/editable.py
```diff
@@ -17,10 +17,15 @@
...
+    __editor_permissions__ = [None, "paper_editing", "slides_editing", "poster_editing"]
```
mic4ael, line 20: The same as above

indico/modules/events/editing/client/js/management/editable_type/TeamManagerModal.jsx
```diff
@@ -0,0 +1,95 @@
...
+  const favoriteUsersController = useFavoriteUsers();
+
+  if (isLoadingPrincipals) {
+    return <Loader active />;
+  } else if (!principals) {
+    return null;
+  }
+
+  return (
+    <div style={{padding: '30px'}}>
```
mic4ael, line 53:
```suggestion
    <div style={{padding: 30}}>
```

the edfault unit is `px` anyway

indico/modules/events/editing/client/js/management/editable_type/TeamManagerModal.jsx
```diff
@@ -0,0 +1,95 @@
...
+  const handleSubmit = async (data, form) => {
+    const changedValues = getChangedValues(data, form);
+    try {
+      await indicoAxios.post(principalsURL({confId: eventId, type: editableType}), changedValues);
+    } catch (error) {
+      return handleSubmitError(error);
+    }
+    onClose();
+  };
+  const favoriteUsersController = useFavoriteUsers();
```
mic4ael, line 44: Hm, maybe move this hook to the top so that we don't mix hooks with normal functions? (it is just a personal preference)

PR #4396, opened 2020-04-07
Fix category statistics display issues

mic4ael:


indico/modules/categories/controllers/display.py
```diff
@@ -104,7 +104,7 @@ def _process(self):
             'total_events': sum(stats['events_by_year'].values()),
             'total_contributions': sum(stats['contribs_by_year'].values()),
             'min_year': stats['min_year'].year,
-            'max_year': date.today().year,
+            'max_year': stats['max_year'].year,
```
ThiefMaster, line 107: The point of using the current year here was to avoid getting useless data when you have people adding events that end a few decades from now. That's why the JS logic uses up to `maxYear + 3` (if there's data for that long) when displaying the chart.

mic4ael: I see.. I was actually wondering why it has been implemented like that. In this case, shouldn't we limit the categories/contributions only to those happening before `date.today().year`?

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -38,12 +38,15 @@ CategoryStatistics.propTypes = {
...
 function getGraphData(data) {
-  const serialize = statsData =>
-    Object.entries(statsData)
+  const serialize = statsData => [
+    [moment(data.minYear - 1, 'YYYY'), null],
+    ...Object.entries(statsData)
       .filter(([year]) => year <= data.maxYear + 3 && year >= data.minYear)
-      .map(([year, number]) => [moment(year, 'YYYY'), number]);
+      .map(([year, number]) => [moment(year, 'YYYY'), number]),
+    [moment(data.maxYear + 4, 'YYYY'), null],
```
ThiefMaster, line 46: This won't work, because we only take up to 3 years after the current year depending on how much data there is

indico/modules/categories/util.py
```diff
@@ -78,7 +78,7 @@ def get_min_year(category_id=None):
     :return: A `datetime` Object.
     """
     category_filter = Event.category_chain_overlaps(category_id) if category_id else True
-    return db.session.query(db.func.min(Event.created_dt)).filter(~Event.is_deleted, category_filter).scalar()
+    return db.session.query(db.func.min(Event.start_dt)).filter(~Event.is_deleted, category_filter).scalar()
```
ParthS007, line 81: what's the reason for changing this?

mic4ael: We probably don't want the max `created_dt`. We group events by their `start_dt` and then we also want to get the max `start_dt` not `created_dt`. Unless there is a reason for this?

indico/modules/categories/client/js/components/CategoryStatistics.jsx
mic4ael: What do you mean? I just went here for a simple solution instead of actually checking what the max date in the returned list is.

indico/modules/categories/util.py
ParthS007: I think I got it. LGTM!

ParthS007: I was using `created_dt` as it is only truly useful for the lower bound

ThiefMaster, line 81: keep `created_dt`; `start_dt` may be WAY in the past

indico/modules/categories/client/js/components/CategoryStatistics.jsx
```diff
@@ -38,12 +38,23 @@ CategoryStatistics.propTypes = {
...
+    const maxDate = parseInt(
+      Object.keys(statsData)
+        .reverse()
+        .find(date => date <= data.maxYear + 3),
+      10
+    );
+    return [
+      [moment(data.minYear - 1, 'YYYY'), null],
+      ...Object.entries(statsData)
+        .filter(([year]) => year <= data.maxYear + 3 && year >= data.minYear)
```
ThiefMaster, line 51: `year <= maxDate`? Same functionality, but only one place where we use the year+3 logic

PR #4398, opened 2020-04-07
Add useTogglableValue react hook & simplify PublicationSwitch

ThiefMaster:
This hook is useful for other cases as well where we have a switch (or something switch-like) that should load its initial state from the server and then saves the new state immediately whenever it gets toggled.

Also updated the `PublicationSwitch` for the event's contribution list to use the new hook and moved it to a more appropriate place (since it's not a generic component).


indico/web/client/js/react/hooks.js
```diff
@@ -122,6 +122,58 @@ export function useIndicoAxios({camelize, ...args}) {
...
+  if (data === null) {
+    return [false, () => null, true, false];
+  }
+
+  // if we changed the state locally, that one has priority, otherwise use the fetched value
+  const checked = newState === null ? data : newState;
+
+  const save = async enabled => {
+    try {
+      // we are optimisting and flip the switch without waiting for the action to finish
```
mic4ael, line 157:
```suggestion
      // we are optimistic and flip the switch without waiting for the action to finish
```
?

PR #4380, opened 2020-03-30
Editable setup

panagiotappl:
closes #4369 

image

PR #4400, opened 2020-04-09
Implement self-assign toggle in editing setup

### LINKED ISSUE #4390: Editing setup: Self-assign toggle
Currently it doesn't do anything. We don't have anything for assignment yet, but it can already toggle a setting stored in the settings proxy.

Note: We can either use prefixed settings in the existing settings proxy for each editable type, or we could add more settings proxies for each editable types. something like this:

```python
editable_type_settings = {
    EditableType.paper: EventSettingsProxy('editing_paper', ...),
    ...
}
```

(but please don't triplicate the defaults(/converters) ;))

Leats:
Also fixes definition of default fonts, which were defined in two different files.
Closes #4390

### DIFF
--- a/indico/modules/events/editing/blueprint.py
+++ b/indico/modules/events/editing/blueprint.py
@@ -51,6 +51,8 @@ _bp.add_url_rule('/editing/api/tag/<int:tag_id>', 'api_edit_tag', management.RHE
 _bp.add_url_rule('/editing/api/menu-entries', 'api_menu_entries', common.RHMenuEntries)
 _bp.add_url_rule('/editing/api/enabled-editable-types', 'api_enabled_editable_types',
                  management.RHEnabledEditableTypes, methods=('GET', 'POST'))
+_bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/editable-assignment/self-assign-enabled',
+                 'api_self_assign_enabled', management.RHEditableSetSelfAssign, methods=('GET', 'PUT', 'DELETE'))
 
 # Contribution/revision-level APIs
 contrib_api_prefix = '/api' + contrib_prefix
--- a/indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
+++ b/indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
@@ -8,14 +8,16 @@
 import dashboardURL from 'indico-url:event_editing.dashboard';
 import manageFileTypesURL from 'indico-url:event_editing.manage_file_types';
 import manageReviewConditionsURL from 'indico-url:event_editing.manage_review_conditions';
+import selfAssignURL from 'indico-url:event_editing.api_self_assign_enabled';
 
 import React from 'react';
 import {useParams, Link} from 'react-router-dom';
-import {Checkbox} from 'semantic-ui-react';
+import {Checkbox, Loader} from 'semantic-ui-react';
 
 import {Translate} from 'indico/react/i18n';
 import {ManagementPageSubTitle, ManagementPageBackButton} from 'indico/react/components';
 import {useNumericParam} from 'indico/react/util/routing';
+import {useTogglableValue} from 'indico/react/hooks';
 import {EditableTypeTitles} from '../../models';
 import Section from '../Section';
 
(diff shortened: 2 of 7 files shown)


indico/modules/events/editing/blueprint.py
```diff
@@ -51,6 +51,8 @@
 _bp.add_url_rule('/editing/api/menu-entries', 'api_menu_entries', common.RHMenuEntries)
 _bp.add_url_rule('/editing/api/enabled-editable-types', 'api_enabled_editable_types',
                  management.RHEnabledEditableTypes, methods=('GET', 'POST'))
+_bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/editable-assignment/self-assign', 'api_self_assign',
```
ThiefMaster, line 54: `.../self-assign-enabled` and `api_self_assign_enabled`?

indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
```diff
@@ -25,78 +27,95 @@ export default function EditableTypeDashboard() {
...
+            <Section
+              icon="copy1"
+              label={Translate.string('Editable assignment')}
+              description={Translate.string('Assign editors')}
+            >
+              <Checkbox
+                styleName="toolbar-checkbox"
+                toggle
+                checked={selfAssignEnabled}
+                onClick={!selfAssignSaving && toggleSelfAssign}
```
plourenco, line 109: image

When `selfAssignSaving` is false and you click the toggle. The component will re-render with `selfAssignSaving` as true and `onClick` will be `!true && toggleSelfAssign = false`

ThiefMaster: Ah it was me suggesting that shortcut 🙈

But you are right, this is a problem here - the code I had in mind when making this suggestion either have something that's `null` and not a boolean before the `&&` part or do something like this:

    onClick={() => !foo && bar()}

indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.module.scss
```diff
@@ -5,11 +5,17 @@
...
     align-self: center;
-}
+    margin-top: 0.4rem;
 
-.toolbar-label {
-    align-self: center;
-    padding: 0 1rem 0 0.5rem;
+    &:global(.ui.toggle.checkbox) input ~ label {
+        @extend %font-family-body;
+        color: $gray !important;
```
panagiotappl, line 17: I personally don't think that the color should be lighter here. I think the color provided by `action-box` css is more visible. Especially since we make the font even smaller in this case.

indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
```diff
@@ -25,78 +27,95 @@ export default function EditableTypeDashboard() {
...
-          <Checkbox styleName="toolbar-checkbox" toggle />
-          <span styleName="toolbar-label">
-            <Translate>Allow editors to self-assign editables</Translate>
-          </span>
-          <a className="i-button icon-settings">
-            <Translate>Assign</Translate>
-          </a>
-        </Section>
-      </div>
+      {!selfAssignLoading && (
```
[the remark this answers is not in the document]
panagiotappl: Unless we don't want the users to perform any other actions when this toggle is loading.. Either way I would add a `<Loader />` from SUIR while `selfAssignLoading`

indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
plourenco: Yes, the shortcut will work just fine until the re-render. I would suggest either that `() => {}`, which IMO is probably the simplest.
Or converting that value to `null/undefined`, even though it looks bad, can't think of another alternative rn.

`!true && toggleSelfAssign || null`
or
`!selfAssignSaving ? toggleSelfAssign : null

indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.module.scss
ThiefMaster: Isn't it the same color? Without it, it's just black.

indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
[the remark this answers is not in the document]
Leats: I thought it looked weird when that this one part just popped up a few miliseconds later. With everything popping up just that tiny bit later it isn't really noticeable.

indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.module.scss
Leats: Bildschirmfoto 2020-04-09 um 17 48 36
This is how it looks like for me when I remove the `color: $gray !important`. Alternatively I could try a darker gray than `$gray`?

panagiotappl: It wasn't black before. There was css being applied through `action-box` which was a darker version of grey... You can check the screenshot in the closed PR: https://github.com/indico/indico/pull/4380

Leats: I moved the label into the Checkbox to make the label clickable as well, and that somehow made it black. But it seems like action-box did use $dark-gray, so I'll use that as well then! That should make it more visible.

PR #4401, opened 2020-04-09
Add location_info blocks in event display templates

OmeGak:
This allows plugins to override only location information in conference display pages rather than the whole content block. This is preferable in order to prevent overridden blocks to diverge from vanilla templates.

### DIFF
--- a/indico/modules/events/templates/display/conference.html
+++ b/indico/modules/events/templates/display/conference.html
@@ -28,29 +28,31 @@
                 </div>
             </div>
 
-            {% if event.has_location_info or event.address %}
-                <div class="infoline location">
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -95,6 +95,12 @@ Bugfixes
 
 - None so far :)
 
+Internal Changes
+^^^^^^^^^^^^^^^^
+
+- Add ``location_info`` block in conference display template (thanks
+  :user:`omegak`)
+
```
ThiefMaster, line 103: Wrong version; should be listed in 2.3 (also add a reference to the PR now that you know its ID)

ThiefMaster: not sure if this really needs a changelog entry though...

OmeGak: Removed.

PR #4402, opened 2020-04-09
Add map URL to events

OmeGak:
This allows adding a map URL to an event regardless of room booking being enabled or not. In case the event takes place in a room with with map URL associated, this one overrides it.

# UI changes

Screen Shot 2020-04-09 at 18 06 25
Screen Shot 2020-04-09 at 18 06 14
Screen Shot 2020-04-09 at 18 05 52

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -63,6 +63,8 @@ Improvements
 - Add UI for admins to block/unblock users (:issue:`3243`)
 - Show labels indicating whether a user is an admin, blocked or soft-deleted
   (:issue:`4363`)
+- Add map URL to events, allowing also to override room map URL (:issue:`4402`,
+  thanks :user:`omegak`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/migrations/versions/20200409_1711_7f56252c73ab_add_map_url_to_events_table.py
+++ b/indico/migrations/versions/20200409_1711_7f56252c73ab_add_map_url_to_events_table.py
@@ -1,0 +1,25 @@
+"""Add map_url to events table
+
+Revision ID: 7f56252c73ab
+Revises: 933665578547
+Create Date: 2020-04-09 17:11:13.685047
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '7f56252c73ab'
+down_revision = '933665578547'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('events', sa.Column('map_url', sa.String(), nullable=False, server_default=''), schema='events')
+    op.alter_column('events', 'map_url', server_default=None, schema='events')
+
+
+def downgrade():
+    op.drop_column('events', 'map_url', schema='events')
--- a/indico/modules/events/management/forms.py
+++ b/indico/modules/events/management/forms.py
@@ -17,8 +17,8 @@ from pytz import timezone
 from werkzeug.datastructures import ImmutableMultiDict
 from wtforms import BooleanField, FloatField, SelectField, StringField, TextAreaField
 from wtforms.ext.sqlalchemy.fields import QuerySelectField
-from wtforms.fields.html5 import IntegerField
-from wtforms.validators import DataRequired, InputRequired, NumberRange, Optional, ValidationError
+from wtforms.fields.html5 import IntegerField, URLField
+from wtforms.validators import URL, DataRequired, InputRequired, NumberRange, Optional, ValidationError
 
 from indico.core.config import config
 from indico.core.db import db
@@ -168,6 +168,13 @@ class EventDatesForm(IndicoForm):
 
 class EventLocationForm(IndicoForm):
     location_data = IndicoLocationField(_('Location'), allow_location_inheritance=False)
+    own_map_url = URLField(_('Map URL'), [Optional(), URL()])
+
+    def __init__(self, *args, **kwargs):
+        event = kwargs['event']
+        super(EventLocationForm, self).__init__(*args, **kwargs)
+        if event.room:
+            self.own_map_url.render_kw = {'placeholder': event.room.map_url}
 
 
 class EventPersonsForm(IndicoForm):
--- a/indico/modules/events/models/events.py
+++ b/indico/modules/events/models/events.py
@@ -244,6 +244,14 @@ class Event(SearchableTitleMixin, DescriptionMixin, LocationMixin, ProtectionMan
         index=True,
         nullable=True
     )
+    #: The url to a map for the event
+    own_map_url = db.Column(
+        'map_url',
+        db.String,
+        nullable=False,
+        default=''
+    )
+
     #: The last user-friendly registration ID
     _last_friendly_registration_id = db.deferred(db.Column(
         'last_friendly_registration_id',
@@ -554,6 +562,14 @@ class Event(SearchableTitleMixin, DescriptionMixin, LocationMixin, ProtectionMan
     def short_external_url(self):
         id_ = self.url_shortcut or self.id
         return url_for('events.shorturl', confId=id_, _external=True)
+
+    @property
+    def map_url(self):
+        if self.own_map_url:
+            return self.own_map_url
+        elif not self.room:
+            return ''
+        return self.room.map_url or ''
 
     @property
     def tzinfo(self):
--- a/indico/modules/events/operations.py
+++ b/indico/modules/events/operations.py
@@ -129,7 +129,8 @@ def update_event(event, update_timetable=False, **data):
     assert set(data.viewkeys()) <= {'title', 'description', 'url_shortcut', 'location_data', 'keywords',
                                     'person_link_data', 'start_dt', 'end_dt', 'timezone', 'keywords', 'references',
                                     'organizer_info', 'additional_info', 'contact_title', 'contact_emails',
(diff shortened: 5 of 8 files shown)


indico/migrations/versions/20200409_1711_7f56252c73ab_add_map_url_to_events_table.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '7f56252c73ab'
+down_revision = '933665578547'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('events', sa.Column('map_url', sa.String(), nullable=True), schema='events')
```
ThiefMaster, line 20: hm.. nullable vs default `''`.. i think with the current wtforms code we get `''` when saving, so we'll have a mix of both - so better go for non-nullable

CHANGES.rst
```diff
@@ -13,6 +13,8 @@ Major Features
 - Add category roles, which are similar to local groups but within the
   scope of a category and its subcategories. They can be used for assigning
   permissions in any of these categories and events within such categories.
+- Add map URL to events, allowing also to override room map URL (:issue:`4402`,
+  thanks :user:`omegak`)
```
ThiefMaster, line 17: Not really a "major feature" IMO. I'd list this under "improvements"

indico/modules/events/templates/display/conference.html
```diff
@@ -49,6 +45,10 @@
                             {% if event.address %}
                                 <div class="address nohtml">{{ event.address }}</div>
                             {% endif %}
+
+                            {% if event.map_url or (event.room and event.room.map_url) %}
```
ThiefMaster, line 49: What about adding a property to `Event` that returns either the custom map url or the room's map url?

indico/modules/events/templates/display/conference.html
```diff
@@ -49,6 +45,10 @@
...
+                                <a href="{{ event.map_url or event.room.map_url }}" target="_blank">Go to map</a>
```
ThiefMaster, line 50: I'm not a friend of `target="_blank"` - but *if* it's used, it should go along with `rel="noopener noreferrer"`

OmeGak: From a UX perspective, `target="_blank"` is preferred. The user is navigating Indico and is expecting to remain in it when clicking links. I understand the need for `noopener`, but why `noreferrer`?

ThiefMaster: Better privacy by not sending a referer to the target URL.

https://developers.google.com/web/tools/lighthouse/audits/noopener

indico/migrations/versions/20200409_1711_7f56252c73ab_add_map_url_to_events_table.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '7f56252c73ab'
+down_revision = '933665578547'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('events', sa.Column('_map_url', sa.String(), nullable=True, server_default=''), schema='events')
```
ThiefMaster, line 20: please don't change the name here; you can pass `'map_url'` as the first posarg to `db.Column()` to set the column name separately from the

indico/modules/events/models/events.py
```diff
@@ -244,6 +244,13 @@ def __table_args__(cls):
         index=True,
         nullable=True
     )
+    #: The url to a map for the event
+    _map_url = db.Column(
```
ThiefMaster, line 248: what about `own_map_url` for the column attr (`map_url` in the DB), and `map_url` for the property?

indico/modules/events/models/events.py
```diff
@@ -555,6 +562,17 @@ def short_external_url(self):
         id_ = self.url_shortcut or self.id
         return url_for('events.shorturl', confId=id_, _external=True)
 
+    @property
+    def map_url(self):
+        map_url = self._map_url
```
ThiefMaster, line 567:
```
if self._map_url:
    return self._map_url:
elif not self.room:
    return None  # or empty string? not sure what's better
return self.room.map_url
```

OmeGak: Since `_map_url` defaults to empty string, I think it would be more consistent to do something like this:

```python
def map_url(self):
    if self._map_url:
        return self._map_url
    elif not self.room:
        return ''
    return self.room.map_url or ''
```

indico/modules/events/models/events.py
```diff
@@ -555,6 +563,18 @@ def short_external_url(self):
...
+    def map_url(self):
+        if self.own_map_url:
+            return self.own_map_url
+        elif not self.room:
+            return ''
+        return self.room.map_url or ''
+
+    @map_url.setter
+    def map_url(self, value):
+        self.own_map_url = value
```
ThiefMaster, line 576: Do we need this setter in the end?

indico/modules/events/operations.py
```diff
@@ -129,7 +129,8 @@ def update_event(event, update_timetable=False, **data):
     assert set(data.viewkeys()) <= {'title', 'description', 'url_shortcut', 'location_data', 'keywords',
                                     'person_link_data', 'start_dt', 'end_dt', 'timezone', 'keywords', 'references',
                                     'organizer_info', 'additional_info', 'contact_title', 'contact_emails',
-                                    'contact_phones', 'start_dt_override', 'end_dt_override', 'label', 'label_message'}
+                                    'contact_phones', 'start_dt_override', 'end_dt_override', 'label', 'label_message',
+                                    'map_url'}
```
ThiefMaster, line 133: Shouldn't we use the `own_map_url` here and in the form?

indico/modules/events/models/events.py
OmeGak: Nope, if I also change it the form.

ThiefMaster: yeah, i think `map_url` should only be used in the display areas, and anything else would use the `own_map_url`

as a bonus, you could set the room-provided map url as a placeholder in the custom map url field to make it clear which url is used when you don't enter anything there. it would only be applied after saving in case you edit the room but i don't think this would be an issue

OmeGak: Done!

indico/modules/events/management/forms.py
```diff
@@ -168,6 +168,13 @@ def validate_end_dt(self, field):
 
 class EventLocationForm(IndicoForm):
     location_data = IndicoLocationField(_('Location'), allow_location_inheritance=False)
+    own_map_url = URLField(_('Map URL'))
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event')
```
ThiefMaster, line 174: no need to store it on `self` since it's not used anywhere else

indico/modules/events/operations.py
```diff
@@ -212,7 +213,8 @@ def _log_event_update(event, changes, visible_person_link_changes=False):
         'contact_emails': 'Contact emails',
         'contact_phones': 'Contact phone numbers',
         'label': {'title': 'Label', 'type': 'string', 'attr': 'title'},
-        'label_message': 'Label message'
+        'label_message': 'Label message',
+        'own_map_url': 'Map URL'
```
ThiefMaster, line 217: `{'title': 'Map URL', 'type': 'string'}`

Like this you get "old -> new" in the log instead of a diff-like view. I think for a URL this makes more sense.

indico/modules/events/management/forms.py
OmeGak: Also, I see no need to pop `event`. `IndicoForm` is not expecting it, but it was receiving it already anyways before this PR.

ThiefMaster: I was just about to ask why you aren't popping it anymore ;) Anyway, passing it doesn't do any harm (usually we pop it unless we inherit from a custom base class that still needs it)

indico/modules/events/management/forms.py
```diff
@@ -168,6 +168,13 @@ def validate_end_dt(self, field):
 
 class EventLocationForm(IndicoForm):
     location_data = IndicoLocationField(_('Location'), allow_location_inheritance=False)
+    own_map_url = URLField(_('Map URL'))
```
mic4ael, line 171: don't you want to use `wtforms.validators.URL` validator to make sure it is a valid URL?

ThiefMaster:
Good point, the `own_map_url` should be included in the `data` dict in `clone_event`!

indico/modules/events/management/forms.py
ThiefMaster: I think everyone (including me) usually assumes that `URLField` would do this by default.. so good point ;x

OmeGak: Yep, surprised about this. Adding the URL validator ;)

indico/modules/events/management/forms.py
```diff
@@ -168,6 +168,13 @@ def validate_end_dt(self, field):
 
 class EventLocationForm(IndicoForm):
     location_data = IndicoLocationField(_('Location'), allow_location_inheritance=False)
+    own_map_url = URLField(_('Map URL'), [URL()])
```
ThiefMaster, line 171: You need `[Optional(), URL()]` or an empty field will fail validation.

OmeGak: Done! (So easy to forget all the quirks...)

PR #4404, opened 2020-04-14
Submit Editable to different types

### LINKED ISSUE #4375: Handle different editable types during submission

mic4ael:
closes #4375

### DIFF
--- a/indico/modules/events/contributions/client/js/index.jsx
+++ b/indico/modules/events/contributions/client/js/index.jsx
@@ -29,28 +29,27 @@ import PublicationSwitch from './PublicationSwitch';
     if (!editableSubmissionButton) {
       return;
     }
-    const {
-      eventId,
-      contributionId,
-      contributionCode,
-      apiEditableSubmitUrl,
-      type,
-      uploadUrl,
-    } = editableSubmissionButton.dataset;
 
-    let response;
+    const availableTypes = JSON.parse(editableSubmissionButton.dataset.availableTypes);
+    const {eventId, contributionId, contributionCode} = editableSubmissionButton.dataset;
+
+    let responses;
     try {
-      response = await indicoAxios.get(fileTypesURL({confId: eventId, type}));
+      responses = await Promise.all(
+        availableTypes.map(type => indicoAxios.get(fileTypesURL({confId: eventId, type})))
+      );
     } catch (e) {
       handleAxiosError(e);
+      return;
     }
-    const fileTypes = camelizeKeys(response.data);
+
+    const fileTypes = _.fromPairs(
+      responses.map((response, index) => [availableTypes[index], camelizeKeys(response.data)])
+    );
+
     ReactDOM.render(
       <EditableSubmissionButton
-        uploadURL={uploadUrl}
         fileTypes={fileTypes}
-        type={type}
-        submitRevisionURL={apiEditableSubmitUrl}
         eventId={eventId}
         contributionId={contributionId}
         contributionCode={contributionCode}
--- a/indico/modules/events/contributions/models/contributions.py
+++ b/indico/modules/events/contributions/models/contributions.py
@@ -494,6 +494,14 @@ class Contribution(DescriptionMixin, ProtectionManagersMixin, LocationMixin, Att
     def paper(self):
         return Paper(self) if self._paper_last_revision else None
 
+    @property
+    def allowed_types_for_editable(self):
+        if not self.event.has_feature('editing'):
+            return []
+
+        submitted_for = {editable.type.name for editable in self.editables}
+        return [editable_type for editable_type in self.event.editable_types if editable_type not in submitted_for]
+
     def is_paper_reviewer(self, user):
         return user in self.paper_content_reviewers or user in self.paper_layout_reviewers
 
--- a/indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
+++ b/indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
@@ -6,40 +6,43 @@
 // LICENSE file for more details.
 
 import editableURL from 'indico-url:event_editing.editable';
+import submitRevisionURL from 'indico-url:event_editing.api_create_editable';
+import apiUploadURL from 'indico-url:event_editing.api_upload';
 
 import _ from 'lodash';
 import React, {useState} from 'react';
 import PropTypes from 'prop-types';
-import {Button, Form, Modal} from 'semantic-ui-react';
+import {Button, Dropdown, Form, Modal} from 'semantic-ui-react';
 import {Form as FinalForm} from 'react-final-form';
+
 import {indicoAxios} from 'indico/utils/axios';
 import {FinalSubmitButton, handleSubmitError} from 'indico/react/forms';
-import {Translate} from 'indico/react/i18n';
+import {Param, Translate} from 'indico/react/i18n';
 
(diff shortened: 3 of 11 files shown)


indico/modules/events/contributions/client/js/index.jsx
```diff
@@ -29,28 +29,29 @@ import PublicationSwitch from './PublicationSwitch';
...
-      response = await indicoAxios.get(fileTypesURL({confId: eventId, type}));
+      responses = await Promise.all(
+        availableTypes.map(type => indicoAxios.get(fileTypesURL({confId: eventId, type})))
+      );
     } catch (e) {
       handleAxiosError(e);
     }
-    const fileTypes = camelizeKeys(response.data);
+
+    const fileTypes = Object.assign(
```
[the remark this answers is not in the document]
mic4ael: It is really a personal preference I think. `Object.fromEntries`, `_.fromPairs` or `Object.assign`. I think all of them are equally good.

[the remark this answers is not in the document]
ThiefMaster: I think `Object.fromEntries` is still missing in Edge, otherwise I'd have suggested that. ;)

IMHO using fromEntries/fromPairs is be superior to `Object.assign(...)` since the former does exactly what you are doing here: creating a dict from a list of tuples

In any case, I think the `{},` isn't needed, is it?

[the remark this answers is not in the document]
mic4ael: Yeah, but you can still polyfill it if you want :P And yeah `{}` isn't needed. It will use the first object as the "base" for the returned object (it is just that I want to be explicit about it).

indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
```diff
@@ -51,47 +54,69 @@ export default function EditableSubmissionButton({
...
-      </button>
+      {editableTypes.length === 1 ? (
+        <Button onClick={() => setCurrentType(editableTypes[0])} primary>
+          <Translate>Submit files</Translate>
+        </Button>
+      ) : (
+        <Dropdown
+          className="primary"
+          text={Translate.string('Submit files')}
+          options={editableTypes.map(editableType => ({
```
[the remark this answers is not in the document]
mic4ael: Nope, since it defaults to `option.value` if `key` is not provided https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Dropdown/Dropdown.js#L1298

indico/modules/events/contributions/client/js/index.jsx
```diff
@@ -29,28 +29,26 @@ import PublicationSwitch from './PublicationSwitch';
...
+    const {eventId, contributionId, contributionCode} = editableSubmissionButton.dataset;
+
+    let responses;
     try {
-      response = await indicoAxios.get(fileTypesURL({confId: eventId, type}));
+      responses = await Promise.all(
+        availableTypes.map(type => indicoAxios.get(fileTypesURL({confId: eventId, type})))
+      );
     } catch (e) {
       handleAxiosError(e);
```
ThiefMaster, line 42: return after this to avoid ugly errors in case of a failure

indico/modules/events/contributions/models/contributions.py
```diff
@@ -494,6 +494,14 @@ def verbose_title(self):
     def paper(self):
         return Paper(self) if self._paper_last_revision else None
 
+    @property
+    def allowed_types_for_editable(self):
+        if not self.event.has_feature('editing'):
+            return []
+
+        submitted_for = [editable.type.name for editable in self.editables]
```
ThiefMaster, line 502: could use a `{...}` set here

indico/modules/events/editing/client/js/models.js
```diff
@@ -29,6 +29,12 @@ export const EditingReviewAction = {
   requestUpdate: 'request_update',
 };
 
+export const EditableType = {
```
ThiefMaster, line 32: We use `Object.keys(EditableTypeTitles)` in two places; can you update those to use your new enum instead?

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -92,14 +92,15 @@ def _check_access(self):
         if not self._user_is_authorized_submitter():
             # XXX: should event managers be able to submit on behalf of the user?
             raise Forbidden
-        # TODO: check if submitting papers for editing is allowed in the event
```
ThiefMaster, line 95: I think it's be worth keeping the TODO comment since we'll still need to check whether submission is currently open for the given editable type.

indico/modules/events/contributions/models/contributions.py
mic4ael: I was using `set` initially but I didn't like the fact that I had to explicitly convert it to a `list` to make it serializable.

ThiefMaster: But you never return this...

mic4ael: Ahh.. ok, I thought you meant I should use sets in the entire function :P That is what I did before (sorry, i got confused)

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -53,14 +52,17 @@ <h2>{% trans %}Editing{% endtrans %}</h2>
                     <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
```
ThiefMaster, line 52: This heading, and "Paper submission is open" / "You can submit a paper" further below, don't really apply in case of other editable types. This is especially misleading if you don't have Paper editing enabled or already submitted one and can only submit in some other editable type - its name is not mentioned anywhere until you click the button in this case.

ThiefMaster: Example where I submitted a paper, and still have to submit slides:

image

mic4ael: Maybe I could change `Paper submission is open` to `Submission is open` and then change `You can submit a paper` to sth like `You can submit file(s) to the following types: poster,slides`

ThiefMaster: "Submission is open" sounds good (and just get rid of the "Paper" heading I guess?)

Maybe "You can submit files for editing." for the smaller text below? With the dropdown we don't need another place listing the types, and without the dropdown I would use a specific button title that mentions for which type you will submit, i.e. "Submit <paper|slides|poster>"

indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
```diff
@@ -51,47 +54,74 @@ export default function EditableSubmissionButton({
...
+        </Button>
+      ) : (
+        <Dropdown
+          className="primary"
+          text={Translate.string('Submit files')}
+          onChange={(__, {value}) => setCurrentType(value)}
+          options={editableTypes.map(editableType => ({
+            text: EditableTypeTitles[editableType],
+            value: editableType,
+            onClick: () => setCurrentType(editableType),
```
ThiefMaster, line 112: do we still need this one? I think the `onChange` handler is enough

mic4ael: Nope.

PR #4405, opened 2020-04-14
Editing service integration - basic setup

ThiefMaster:


indico/modules/events/editing/client/js/management/Section.jsx
```diff
@@ -23,7 +23,7 @@ export default function Section({icon, label, description, children}) {
```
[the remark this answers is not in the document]
ThiefMaster: yeah, either now or when we end up using it somewhere else

ThiefMaster:
For testing purposes you can run [this dummy service](https://gist.github.com/ThiefMaster/e3579f02874e5576988bd9bfbbca09c8). The easiest way to create an error after connecting the service is to delete the event from the service's DB via psql.

indico/modules/events/editing/client/js/management/ManageService.jsx
```diff
@@ -0,0 +1,228 @@
...
+                      setServiceURLInfo(null);
+
+                      const validationError = v.required(value) || v.url(value);
+                      if (validationError) {
+                        return validationError;
+                      }
+
+                      let resp;
+                      try {
+                        // this does send a request on each keypress, but in this particular case
```
plourenco, line 169: I agree, and this is probably the simplest approach.

But an idea to consider: What about moving this to a function and adding a small delay to check the URL?
The validation would be pending until you stop writing for, say, 500 ms. Each key press within this range would cancel the previous promise (with a spinner loading).

indico/modules/events/editing/client/js/management/ManageService.jsx
```diff
@@ -0,0 +1,228 @@
...
+                        // fail validation nicely.
+                        const submitError = handleSubmitError(err);
+                        return submitError.url || submitError[FORM_ERROR];
+                      }
+
+                      if (resp.data.error) {
+                        return resp.data.error;
+                      }
+
+                      setServiceURLInfo(resp.data.info);
```
plourenco, line 188: If I close the modal, the state of this variable will be retained. It should probably be cleared `onClose`.

indico/modules/events/editing/client/js/management/ManageService.jsx
```diff
@@ -0,0 +1,228 @@
...
+  const checkPending = loading || (data === null && error === null);
+  let label = Translate.string('Custom editing workflow');
+  let description;
+  let showConnect = false;
+  let showDisconnect = false;
+  let forceDisconnect = false;
+
+  if (checkPending) {
+    description = Translate.string('Checking custom editing workflow status');
+  } else if (error || data.error) {
```
plourenco, line 48: Do we need to double-check for the error? Isn't it already guaranteed to be true if `data.error` exists?

image

ThiefMaster: `error` means the request to our own backend failed, `data.error` means that checking the service itself failed. You never have both at the same time.

In any case, we should never end up here with neither an error nor data (since we're in checkPending in that case)

indico/modules/events/editing/client/js/management/ManageService.jsx
ThiefMaster: That's indeed an option. To be honest, I just didn't want to implement an async debounce function ;)

indico/modules/events/editing/client/js/management/ManageService.jsx
plourenco: I didn't realize earlier that `data.error` was referring to the service. That makes sense.

indico/modules/events/editing/client/js/management/ManageService.jsx
ThiefMaster: I don't think `onClose` triggers when using the controlled version of the Modal. Anyway, fixed it and also added the missing onClose when prevented you from closing the modal by clicking e.g. on the dimmer outside.

indico/modules/events/editing/client/js/management/ManageService.jsx
ThiefMaster: Added some debouncing. Not super pretty but I don't think async debounce can be made much prettier...

plourenco: Looks good! I would suggest a Spinner to show the loading state (setLoading(true) when starting and setLoading(false) right after), but apart from that, 👍.

ThiefMaster: not sure if that would look good - in general i like the idea, but the question is always where to put those things... more often than not when trying to put spinners somewhere they looked weird

plourenco: My idea was this: https://react.semantic-ui.com/elements/input/#states-loading

ThiefMaster: oh nice, i forgot that this existed. looks good indeed, and can most likely be done in a generic way directly in the FormFieldAdapter!

ThiefMaster: image

indico/web/client/js/react/forms/fields.jsx
```diff
@@ -70,6 +71,10 @@ export function FormFieldAdapter({
     delete input.checked;
   }
 
+  if (loaderWhileValidating) {
```
plourenco, line 74: Probably not worth re-opening just for this. But isn't this (in the props) equivalent?
```
loading={loaderWhileValidating && validating}
```

ThiefMaster: No, some fields (the react-dates date picker) complain about any unexpected prop in their proptypes validation. I had `loading={loaderWhileValidating ? validating : undefined}` initially hoping that `undefined` props would simply not be passed along, but this wasn't the case...

PR #4408, opened 2020-04-15
Refactor room edit modal

### LINKED ISSUE #3958: Room Booking: refactoring Room edit modal
A tabbed layout would be an interesting option.

plourenco:
Closes #3958

Update the room edit modal to a tabbed layout. Also refactor the form
into multiple sub-components from a single object.

Under heavy development and refactoring with a couple of TODOs:
- [x] Prevent the errors from showing when switching tabs (possibly with a form reset that keeps dirty)
- [x] Handle submit errors by showing the affected tab
- [x] Handle validation errors ~by showing the affected tab, or prevent switching if any error is triggered~
 - This may include moving the validation into final-form.

image


indico/modules/rb/client/js/common/rooms/RoomEditNotifications.jsx
```diff
@@ -0,0 +1,77 @@
...
+        <Header as="h5">
+          <Translate>How many days in advance booking reminders should be sent</Translate>
+        </Header>
+        <Form.Group widths="equal">
+          <FinalInput
+            fluid
+            name="notificationBeforeDays"
+            label={Translate.string('Single/Daily')}
+            type="number"
+            max={30}
```
plourenco, line 24: Missing `min={1}`

indico/modules/rb/client/js/common/rooms/DailyAvailability.jsx
```diff
@@ -26,38 +26,37 @@ export default class DailyAvailability extends React.Component {
...
+          : v
+      )
     );
   };
 
-  handleRemoveTimes = key => {
+  handleRemoveTimes = index => {
     const {onChange, value} = this.props;
-    onChange([...value.filter(bH => bH.key !== key)]);
+    onChange(value.filter((_, i) => i !== index));
```
ThiefMaster, line 42: I'd use `__` here since we often import lodash as `_`

indico/modules/rb/client/js/common/rooms/RoomEditDetails.jsx
```diff
@@ -0,0 +1,48 @@
+import {Form, Header} from 'semantic-ui-react';
```
ThiefMaster, line 1: missing file header

ThiefMaster: I'd move all the editing stuff into an `edit` subdirectory

indico/modules/rb/client/js/common/rooms/RoomEditDetails.jsx
```diff
@@ -0,0 +1,48 @@
...
+import {FinalInput, FinalTextArea} from 'indico/react/forms';
+
+function RoomEditDetails() {
+  return (
+    <>
+      <Header>
+        <Translate>Contact</Translate>
+      </Header>
+      <FavoritesProvider>
+        {favoriteUsersController => (
```
ThiefMaster, line 15: Not needed; you can use the `useFavoriteUsers` hook directly here; the only reason `FavoritesProvider` exists is to use it in class-based components.

ThiefMaster: Also, unless this component stays mounted even when switching tabs I would move this hook one level up into the parent component and pass the controller as a prop - otherwise the favorite users ids are resolved every time this component gets mounted.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,200 @@
...
+import RoomEditDetails from './RoomEditDetails';
+import RoomEditPermissions from './RoomEditPermissions';
+import RoomEditOptions from './RoomEditOptions';
+
+function RoomEditModal({roomId, locationId, onClose, afterCreation}) {
+  const [globalAttributes, setGlobalAttributes] = useState([]);
+  const [newRoomId, setNewRoomId] = useState(null);
+  const [wasEverUpdated, setWasEverUpdated] = useState(null);
+
+  const [roomDetails, setRoomDetails] = useState({
```
ThiefMaster, line 41: Do we need this state? Ideally it should be enough to pass all the data into the FinalForm's `initialValues`.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,200 @@
...
-  };
-
-  handleSubmit = async (data, form) => {
-    let {roomId} = this.props;
+    return camelizeKeys(response.data);
+  }, []);
+
+  const fetchRoomData = useCallback(async () => {
+    const resp = await Promise.all([
+      fetch(roomURL, {room_id: roomId}),
```
ThiefMaster, line 76: I'd call the url builder here instead of passing the params separately, i.e. `roomURL({room_id: roomId})`

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,200 @@
...
-      });
+      if (!isNewRoom()) {
+        setWasEverUpdated(true);
+        fetchRoomData();
+      }
     } catch (e) {
-      submitError = handleSubmitError(e);
-      submitState = 'error';
-      this.setState({submitState});
+      return camelizeKeys(handleSubmitError(e));
```
ThiefMaster, line 123: I'd rather keep the `snake_case` names on the various form fields to avoid having to snakify/camelize here.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,200 @@
...
+    if (!isNewRoom()) {
+      fetchRoomData();
     }
-  };
+  }, [isNewRoom, fetchRoomData]);
+
+  const tabPanes = useMemo(
+    () => [
+      {
+        menuItem: 'Basic Details',
```
ThiefMaster, line 144: missing i18n

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,200 @@
...
+            {({submitSucceeded}) => {
+              return (
+                <Message styleName="submit-message" positive hidden={!submitSucceeded || dirty}>
                   <Translate>Room has been successfully updated.</Translate>
                 </Message>
-                <Message styleName="submit-message" negative hidden={submitState !== 'error'}>
+              );
+            }}
+          </FormSpy>
+          <FormSpy subscription={{submitFailed: true}}>
```
ThiefMaster, line 201: I'd use a single FormSpy that subscribes to both `submitSucceeded` and `submitFailed`.

In principle you can probably avoid using `FormSpy` altogether and just add it to the global subscription in the FinalForm and then get these two values from the fprops.

indico/modules/rb/client/js/common/rooms/RoomEditModal.module.scss
```diff
@@ -13,11 +13,15 @@
...
+    margin: 0 0 1em !important;
 }
 
 .room-form {
     :global(.fields) > :global(.field) {
         flex-grow: 1 !important;
     }
 }
+
+.DateRangePicker_picker { // sass-lint:disable-line class-name-format
```
ThiefMaster, line 25: don't you need `:global()` here?

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,200 @@
...
-    const roomId = this.newRoom ? this.state.newRoomId : this.props.roomId;
-    this.setState({closing: true});
-    if (roomId !== null && wasEverSaved) {
-      await Promise.all([
-        fetchEquipmentTypes(),
-        fetchRoom(roomId),
-        fetchRoomPermissions(roomId),
-        fetchRoomDetails(roomId, true),
-      ]);
-    }
```
ThiefMaster, line 650: It looks like you didn't add this back in the new code. It's required to update the data in redux when editing a room from outside the admin area.

indico/modules/rb/client/js/common/rooms/DailyAvailability.jsx
plourenco: Agree, easier to distinguish.

indico/modules/rb/client/js/common/rooms/RoomEditDetails.jsx
plourenco: > missing file header

Thanks.

indico/modules/rb/client/js/common/rooms/RoomEditDetails.jsx
plourenco: > Also, unless this component stays mounted even when switching tabs I would move this hook one level up into the parent component and pass the controller as a prop - otherwise the favorite users ids are resolved every time this component gets mounted.

Unfortunately, the component needs to stay mounted when switching tabs to maintain the form state consistent. Nonetheless, I think I may end up moving it up just to have more control and keep this unidirectional flow.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Do you mean the state inside `roomDetails`? I copied that from the previous default room state set when a new room is created, instead of edited. I believe those are the default values. Not sure if any of them are needed, but I'll give it a try.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: And convert all the input names into `snake_case`?

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Thanks

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: I've had both approaches, ended up with this because using two `FormSpy`'s will result in an extra empty `<>` to hold both `Message` components (unless we merge them and use conditionals).
Adding it as a global subscription would be perfect, but wouldn't that make the whole form re-render when we only need it in a small part?

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
ThiefMaster: Yeah, in this particular case I'd do it.

I think camelize for responses coming back from the server still makes sense in many cases, but for form fields I'd rather use the same names - it's nice if you can search for the name of a field and find both the server-side and client-side code for it!

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Sounds good!

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Yes, that is helpful. I've updated it. Even though having to convert from snake case to camel case in between code and input names can become annoying e.g.
from:
`const {availableEquipment, availableSomething} = camel(input)`
to:
`const {available_equiment: availableEquipment, available_something: availableSomething} = input`

However, I think for this case it is still manageable.

indico/modules/rb/client/js/common/rooms/RoomEditModal.module.scss
plourenco: Works fine with both approaches. I'm keeping `:global()`.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: I was wondering what that was for. Eventually, I noticed there are different room fetch calls for admin and non-admin data. Thanks!

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
ThiefMaster: Indeed this is a bit ugly... Well you can always use e.g. `input.available_equipment` depending on how you are using things. We don't have to destructure everywhere - it's a convenient thing to do in many places, but I think it's best to decide case-by-case whether it's the right / more readable thing to do or not.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
ThiefMaster: Exactly - the admin data is freshly loaded all the time (since you REALLY don't want to accidentally edit a room using stale data), while the data for users is loaded once on page load, and then never reloaded (except when you are an admin and enter+leave the admin area, or edit a room from the normal area)

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
```diff
@@ -0,0 +1,162 @@
...
+import EquipmentList from './EquipmentList';
+import DailyAvailability from './DailyAvailability';
+import NonBookablePeriods from './NonBookablePeriods';
+
+function RoomEditOptions({active, showEquipment, globalAttributes}) {
+  const attributeTitles = useMemo(
+    () => (globalAttributes ? _.fromPairs(globalAttributes.map(x => [x.name, x.title])) : []),
+    [globalAttributes]
+  );
+  const withKeyAttribute = useCallback(
```
plourenco, line 25: Remove this method since the methods below already generate the keys internally.

indico/web/client/js/react/forms/validators.js
```diff
@@ -68,7 +68,7 @@ function optional(arg = null) {
     return chain(optional(), arg);
   }
   return value => {
-    if (value === null) {
+    if ([null, undefined].includes(value)) {
```
ThiefMaster, line 71: Not sure if `includes()` really makes things more readable here.. I'd go for good old `... || ...` here.

plourenco: Yea, sounds good!

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: I updated this part to only contain the essential. Should be fine now.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,309 @@
...
-    onClose(wasEverSaved || afterCreation);
-  };
-
-  handleSubmit = async (data, form) => {
-    let {roomId} = this.props;
+      ].map(x => ({
+        ...x,
+        menuItem: (
+          <Menu.Item key={x.key}>
+            {x.menuItem} {tabsWithError.includes(x.key) ? <Label color="red">Error</Label> : null}
```
ParthS007, line 186: `<Translate>Error</Translate>`

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,309 @@
...
-  parsers as p,
-} from 'indico/react/forms';
-import {FavoritesProvider} from 'indico/react/hooks';
+import {useDispatch, useSelector} from 'react-redux';
+import {snakifyKeys} from 'indico/utils/case';
 import {Translate} from 'indico/react/i18n';
-import {FinalEmailList, FinalPrincipal, ACLField} from 'indico/react/components';
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+
+import './RoomEditModal.module.scss';
```
panagiotappl, line 39: We usually have `scss` imports below all other imports. 

In general, you could reorganize the imports a bit (eg react import before SUIR). You can check the other components we have!

Same in other files of this PR.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,309 @@
...
+        dispatch(roomsActions.fetchEquipmentTypes()),
+        dispatch(userActions.fetchRoomPermissions(roomId)),
+        dispatch(roomsActions.fetchDetails(roomId, true)),
+      ]);
+    }
+    onClose(wasEverUpdated || afterCreation);
+  }, [dispatch, onClose, roomId, newRoomId, wasEverUpdated, afterCreation]);
+
+  useEffect(() => {
+    (async () => setGlobalAttributes(await fetch(fetchAttributesURL())))();
```
panagiotappl, line 264: Maybe you could improve the naming here a bit. For example, remove the word `fetch` at the beginning of the URL imports to something like this: `await fetch(attributesURL())` or maybe change the name of the `fetch` const to something else. I think the naming is a bit confusing here.

indico/modules/rb/client/js/common/rooms/edit/RoomEditLocation.jsx
```diff
@@ -0,0 +1,114 @@
...
+          validate={v.optional(v.min(0))}
+          hideErrorPopup={!active}
+        />
+        <FinalInput
+          fluid
+          type="text"
+          name="latitude"
+          label={Translate.string('Latitude')}
+          parse={f => p.number(f, false)}
+          validate={v.optional(v.number())}
```
ThiefMaster, line 71: Not sure if I'm just missing it, but while latitude/longitude are both optional, once one of them is set, the other becomes required as well.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,309 @@
...
-  }
-  if (latitude !== null && typeof latitude !== 'number') {
-    errors.latitude = Translate.string('Coordinates should be valid numbers.');
-  }
-  if (longitude !== null && typeof longitude !== 'number') {
-    errors.longitude = Translate.string('Coordinates should be valid numbers.');
-  }
-  return errors;
-}
+  const isNewRoom = useCallback(() => roomId === undefined, [roomId]);
```
ThiefMaster, line 81: Do we need a function/callback for this? Wouldn't just `const isNewRoom = roomId === undefined;` work?

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,309 @@
...
-          label: Translate.string('Single/Daily'),
-          inputArgs: {
-            type: 'number',
-            min: 1,
-            max: 30,
-            fluid: true,
-          },
-          required: false,
-        },
+  const fetch = useCallback(async url => {
```
ThiefMaster, line 83: I'd still prefer a different name since `fetch` is also the name of a builtin.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,309 @@
...
-            type: 'number',
-            min: 1,
-            max: 30,
-            fluid: true,
-          },
-          required: false,
+          key: 'location',
+          menuItem: <Translate>Location</Translate>,
+          pane: <RoomEditLocation key="location" />,
+          fields: [
```
ThiefMaster, line 121: Any chance we could do some trickery with jest tests that ensures that the list here matches the final-form fields rendered by the pane?

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
ThiefMaster, line 186: What about using an icon instead of a text label?

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -8,951 +8,309 @@
...
-        fetchRoomPermissions(roomId),
-        fetchRoomDetails(roomId, true),
-      ]);
-    }
-    onClose(wasEverSaved || afterCreation);
-  };
-
-  handleSubmit = async (data, form) => {
-    let {roomId} = this.props;
+      ].map(x => ({
```
ThiefMaster, line 182: maybe `pane` or something similar would be better than `x` here since it's used over a few lines, which makes using a slightly more meaningful name is more appropriate IMO

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
ThiefMaster, line 264: This only runs once since `fetch` is defined with an empty dependency list and thus never changes, right?

If yes you could probably use `useIndicoAxios`. And by no longer having a dependency on your `fetch` util here it means that you may not need to wrap that util in `useCallback` anymore.

indico/modules/rb/client/js/common/rooms/edit/RoomEditLocation.jsx
ThiefMaster: You might be able to wrap both of these components in a `<Field name="theotherfieldsname" subscription={{value: true}}>` to have access to the other field's value, and then set `validate` based on this.

indico/modules/rb/client/js/common/rooms/edit/RoomEditLocation.jsx
```diff
@@ -0,0 +1,114 @@
...
+  );
+}
+
+RoomEditLocation.propTypes = {
+  active: PropTypes.bool,
+};
+
+RoomEditLocation.defaultProps = {
+  active: true,
+};
```
ThiefMaster, line 112: Does having a default here make sense? Especially one that's not `false`?

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
```diff
@@ -0,0 +1,162 @@
...
+                      name: 'remove',
+                      color: 'red',
+                      link: true,
+                      onClick: () => fields.remove(index),
+                    }}
+                  />
+                ))}
+                {fields.length === 0 && (
+                  <div>
+                    <Translate>No custom attributes found</Translate>
```
ThiefMaster, line 112: s/found/defined/?

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
```diff
@@ -0,0 +1,162 @@
...
+        <FinalCheckbox
+          name="notifications_enabled"
+          label={Translate.string('Reminders enabled')}
+          hideErrorPopup={!active}
+        />
+        <FinalCheckbox
+          name="end_notifications_enabled"
+          label={Translate.string('Reminders of finishing bookings enabled')}
+          hideErrorPopup={!active}
+        />
```
ThiefMaster, line 139: Should those be here or on the notifications tab? I'd be inclined to put the actual toggleable checkboxes on the notifications tab, and show some kind of placeholder (disabled checkbox?) here indicating the current state and maybe a link that switches to the notifications tab where one can change them.

Right now it's a bit confusing to not see some fields in the notifications tab based on a checkbox in a different tab - you have no way of knowing that toggling them will show/enable fields elsewhere.

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
```diff
@@ -0,0 +1,162 @@
...
+              return null;
+            }
+            const fieldsByName = fields.map(x => x.name);
+            const options = globalAttributes
+              .map(x => ({
+                key: x.name,
+                text: x.title,
+                value: x.name,
+              }))
+              .filter(attr => !fieldsByName.includes(attr.key));
```
ThiefMaster, line 76: This doesn't work correctly; I can still add an attribute which is already defined.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Thanks!

plourenco:
> why `export default` is not written with the function itself as there is only one function in most of the files?

I think this was a personal preference. But I'm changing the style since the latter is more straightforward.

plourenco:
> @plourenco I have tested the edit modal and observed these 2 things:
> 
> 
> Screenshot from 2020-04-22 10-53-38
> 
> When we upload a picture and click on save (button just below picture), It sends a `POST` and `GET` to show the image. After that, when we delete the picture just after that, it sends a `DELETE` and then `GET` to get the image with the same query string when we saved the image which gives a 404 on network tab. Is it supposed to work that way?
> 
> 
> Screenshot from 2020-04-22 10-34-11
> Now, we have validation error on 3 fields continuously, I can't select the Building field. Maybe the position of the tooltip can be changed?

1. Completely not expected. It was the simple order of setState updates in the async call that caused the files to be set to null before the preview flag was false.

2. That was actually a big problem, probably affecting other fields. Nice catch! 
It seems like we are using `open` together with `trigger`, however, the first one is preventing the latter.
Ideally, we don't need the Popups to be persistent, we only need to show them if there is an error and you actually want to find more information and **click** or **hover** on the input. (This is what the `trigger` controls).

Having the `open` set to `showErrorPopup` means it will either always stay closed if there is no error (expected) or always stay open if there is an error (not expected, the trigger should control this).

So what I did was not setting the `open` prop if there is an error.
`open={showErrorPopup && undefined}`

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Agree, I did some re-organization based on the warnings I was getting. However, I would probably expect this to be linted, it is pretty hard to track.

ThiefMaster: Is there any decent linter for this that works with eslint? The one we have isn't very powerful and just checks some basic things like relative imports after absolute ones...

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: I agree it is a bit confusing. Thanks.
I'm removing the `fetch` as a prefix from those URL imports.

indico/modules/rb/client/js/common/rooms/RoomEditModal.module.scss
```diff
@@ -13,11 +13,15 @@
...
 
 .room-form {
     :global(.fields) > :global(.field) {
         flex-grow: 1 !important;
     }
 }
+
+:global(.DateRangePicker_picker) { // sass-lint:disable-line class-name-format
+    z-index: 10;
+}
```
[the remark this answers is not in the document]
plourenco: Yes, should be enough. Agree. 
Not sure if there is any implication in applying this everywhere though.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: We don't, actually, I'm not even sure how that ended up being a callback.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Any suggestions?

indico/modules/rb/client/js/common/rooms/RoomEditModal.module.scss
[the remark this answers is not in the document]
ThiefMaster: Should be safe. Just make sure it's before the 1001 z-index for the fullscreen one in that file. 

FWIW I suggested @Leats to do the exact same in her task since she also had a case of a switch appearing above a date picker, so if this solution works for you as well I'm actually thinking about just pushing it straight to master since both of you need it anyway.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
ThiefMaster: loadData, fetchData, ...

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Yes, sure, I can do a test for that.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Agree, I'm way too used to `x`.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Exactly, we can actually remove the `fetch` dependency here.

indico/modules/rb/client/js/common/rooms/edit/RoomEditLocation.jsx
plourenco: When I was building this, I was thinking about the generic case someone would re-use this component and expect by default it would be active without having to specify the `active` prop.

I would always expect a TabPane component to be active by default if I'm not controlling that prop.

indico/modules/rb/client/js/common/rooms/RoomEditModal.module.scss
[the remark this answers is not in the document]
plourenco: I cannot confirm it works for me since react-dates' `_datepicker.css` is now overriding the new `dates.scss` style.

image

We can solve it by adding a !important though.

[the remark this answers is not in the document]
ThiefMaster: ah I guess we're including their css later... `!important` seems reasonable then.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Sure, looks good!
image

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Actually, `useIndicoAxios` doesn't seem to provide the error handling we have in fetch, does it?
Unless we have a utils method for it, I would like to avoid having more than one line here.

ThiefMaster: It triggers the error dialog automatically in case of an error, and returns `{data, error, loading}` (and some other) things

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
plourenco: Yes, it is confusing from a user point of view. Imo, I think this should be in the notifications tab without any links in the other tab. I don't think the assumption of considering it a (generic) option was correct anyway.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: After finding some problems most likely related to dependencies when testing this component, we've decided to move this to a separate task.

indico/modules/rb/client/js/common/rooms/edit/RoomEditLocation.jsx
plourenco: Unfortunately, even though this is the simplest solution, it will not work properly. It seems that the validation happens before the field can change it again.
One simple test case is:
- Set latitude to 123
- Set longitude to 123
- Remove latitude
- Remove longitude
- Latitude will remain required

I believe the alternative would be moving this to an after-submit validation. I'm open for suggestions.

ThiefMaster: validating it on the server side is better than nothing for sure (and i hope we already do that?!).

or you could use a form-level validation function just for this case. i think this would actually be logic: validate field-specific stuff on the field-level, and stuff spanning multiple fields on the form level

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: I did a quick research and couldn't find any solution for this particular case. However, I believe we can come back to this later in another task.

indico/modules/rb/client/js/common/rooms/edit/RoomEditLocation.jsx
plourenco: > validating it on the server side is better than nothing for sure (and i hope we already do that?!).
> 
> and you could use a form-level validation function just for this case. i think this would actually be logic: validate field-specific stuff on the field-level, and stuff spanning multiple fields on the form level

Agree, I'll make this the solution.

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
```diff
@@ -0,0 +1,132 @@
...
+import PropTypes from 'prop-types';
+import {Translate} from 'indico/react/i18n';
+import {FinalCheckbox, FinalField, FinalInput} from 'indico/react/forms';
+import EquipmentList from './EquipmentList';
+import DailyAvailability from './DailyAvailability';
+import NonBookablePeriods from './NonBookablePeriods';
+
+export default function RoomEditOptions({active, showEquipment, globalAttributes}) {
+  const attributeTitles = useMemo(
+    () => (globalAttributes ? _.fromPairs(globalAttributes.map(x => [x.name, x.title])) : []),
```
mic4ael, line 21: can `globalAttributes` end up being `null` or `undefined`? If not, the default value for it is `[]` thus you don't need `?` operator at all.

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
```diff
@@ -0,0 +1,132 @@
...
+      <Form.Group>
+        {showEquipment && (
+          <FinalField
+            name="available_equipment"
+            component={EquipmentList}
+            isEqual={_.isEqual}
+            componentLabel={Translate.string('Add new equipment')}
+          />
+        )}
+        {globalAttributes && (
```
mic4ael, line 46: Don't you want to check for `globalAttributes.length > 0`? `[]` evaluates to `true`

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
plourenco: Yes, the default value is `[]` but I always expect the error case where `globalAttributes` ends up being null.

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
plourenco: The purpose of `globalAttributes &&` was to hide them if they did not load. But I'm guessing we probably don't need to show them if we did not define any, globally. So, maybe `globalAttributes && globalAttributes.length > 0`.

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
mic4ael: Hm.. then, in case of an invalid value for the component's prop, should we really account for that? This seems to me like it should never happen and we should just let any expection caused by the invalid value to propagate.

indico/modules/rb/client/js/common/rooms/RoomDetailsModal.jsx
```diff
@@ -81,34 +81,31 @@ class RoomDetailsModal extends React.Component {
...
-          <RoomEditModal roomId={room.id} onClose={this.handleCloseRoomEditModal} />
-        )}
-      </>
+    return !roomEditVisible ? (
+      <Modal onClose={this.handleCloseModal} size="large" closeIcon open>
+        <Modal.Header styleName="room-details-header">
+          {title}
+          <Responsive.Tablet andLarger>
+            {room.canUserEdit && (
+              <span>
```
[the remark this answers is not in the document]
plourenco: I don't think we do. I'll remove it.

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
plourenco: The problem is, if by any reason the `globalAttributes` request fails we have a null value that is not being correctly handled here (will crash). It is not a default prop, so I think we should always keep this as defensive.

mic4ael: Ahh.. I see now. Then yeah, it is perfectly fine to account for that ;)

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -7,952 +7,330 @@
...
+      if (!isNewRoom) {
+        setLoading(true);
+        await fetchRoomData();
+        setLoading(false);
       }
+    })();
+  }, [isNewRoom, fetchRoomData]);
+
+  const formValidation = values => {
+    if (!values.latitude !== !values.longitude) {
```
ThiefMaster, line 289: I think the negation here is bit weird. Since you just use it to cast to boolean, maybe using the more common `!!` here?

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -7,952 +7,330 @@
```
ThiefMaster, line 7: Shouldn't this file also go inside the `edit/` subdir?

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
```diff
@@ -962,123 +340,83 @@ class RoomEditModal extends React.Component {
...
         <Modal.Actions>
-          <Button onClick={this.handleCloseModal}>
-            {submitState === 'success' ? (
-              <Translate>Close</Translate>
-            ) : (
-              <Translate>Cancel</Translate>
+          <FormSpy subscription={{submitSucceeded: true}}>
+            {({submitSucceeded}) => (
+              <Button onClick={onClose}>
+                {submitSucceeded ? <Translate>Close</Translate> : <Translate>Cancel</Translate>}
```
ThiefMaster, line 366: Not sure if that does what you intend: The old `submitState` was reset as soon as something changed in the form (so it went back from Close to Cancel if you started editing after saving) - I don't think this happens with `submitSucceeded`.

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
```diff
@@ -0,0 +1,132 @@
...
+            componentLabel={Translate.string('Add new equipment')}
+          />
+        )}
+        {globalAttributes && globalAttributes.length > 0 && (
+          <FieldArray name="attributes" isEqual={_.isEqual}>
+            {({fields}) => {
+              if (!fields.value) {
+                return null;
+              }
+              const fieldsByName = fields.value.map(x => x.name);
```
ThiefMaster, line 52: Maybe use `new Set(...)` since you only check for containment below (you'll have to replace `includes` with `has` then)?

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: I'm not sure, what do you think?
Since this is a `core` file, connected to the `RoomEditDetails` for me it made more sense to highlight it and move to the `rooms`.

ThiefMaster: hm you could always re-export it from edit/index.js so you can `import RoomEditModal from './edit';` which I think is clearer (and that way anything for editing is in one place)

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: That was my attempt of implementing an XOR without using the bitwise operator `^`. Is your suggestion `!!a !== !!b`?

ThiefMaster: yeah.. or since you probably always have null or a value you could also go for this: `(a === null) != (b === null)`

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: Yes, the only difference I saw was not resetting after a successful submission. 
However, do you think it is worth changing this to the internal state (another re-render) just because of a button change after the first submit?
It could also seem reasonable for the button to say `Close` if you already saved previously.

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
plourenco: Sure, I can do that.

indico/modules/rb/client/js/common/rooms/RoomEditModal.jsx
plourenco: I'm not sure if that is a better solution since the only thing I am importing and exporting is `RoomEditModal` itself, it is an unnecessary barrel. Nonetheless, I moved it into the `edit`.

mic4ael:
Really great job! I am just thinking about one thing. Right now, when you change between tabs the height of the modal changes according to the tab's content. Wouldn't it be nicer if we could have (if possible) fixed height and prevent those changes?

mic4ael:
Another thing that I just noticed. Not sure if we haven't noticed it so far, but when one uploads a new room photo, it is possible to upload for example a pdf. Backend of course will reject it, but as far as I know `Dropzone` also allows to define a list of allowed mimetypes.

plourenco:
> Really great job! I am just thinking about one thing. Right now, when you change between tabs the height of the modal changes according to the tab's content. Wouldn't it be nicer if we could have (if possible) fixed height and prevent those changes?

I agree, it is annoying. Even though the modal is already top-aligned, I'm not sure it is possible to prevent it from stretching vertically besides adding a fixed height. We were previously using a scrollable modal, but I don't like that approach.

> but as far as I know Dropzone also allows to define a list of allowed mimetypes.

This is a great suggestion, we can prevent this.

indico/modules/rb/client/js/common/rooms/edit/RoomEditOptions.jsx
```diff
@@ -0,0 +1,150 @@
...
+            componentLabel={Translate.string('Add new equipment')}
+          />
+        )}
+        {globalAttributes && globalAttributes.length > 0 && (
+          <FieldArray name="attributes" isEqual={_.isEqual}>
+            {({fields}) => {
+              if (!fields.value) {
+                return null;
+              }
+              const fieldsByName = new Set(fields.value.map(x => x.name));
```
mic4ael, line 52: Why `Set` here? Don't you get unique values here in `fields`?

plourenco: Suggested [here](https://github.com/indico/indico/pull/4408/files/088a9cfcce9e3cb2f544c55be8132e71b6560bdc#diff-cd3d7e574e8639b08a5920392ad4b9ba) as a minor improvement for the `includes` check below.

indico/modules/rb/client/js/common/rooms/edit/RoomEditModal.jsx
```diff
@@ -0,0 +1,435 @@
...
+        setLoading(false);
+      }
+    })();
+  }, [isNewRoom, fetchRoomData]);
+
+  const formValidation = values => {
+    if (!!values.latitude !== !!values.longitude) {
+      // Validation for dependent fields is made at the form level, since field
+      // level won't handle it properly.
+      const error = 'Both latitude and longitude need to be set (or omitted).';
```
ThiefMaster, line 288: missing i18n

indico/modules/rb/client/js/common/rooms/edit/RoomEditModal.jsx
```diff
@@ -0,0 +1,435 @@
...
+          pristine: true,
+          dirty: true,
+        }}
+        mutators={{...arrayMutators}}
+        validate={formValidation}
+      />
+    );
+  };
+
+  return renderForm();
```
ThiefMaster, line 428: What's the reason for having this inside a function? I think removing the `const renderForm = () => { ... }` around it would still keep the exact same behavior

indico/modules/rb/client/js/common/rooms/edit/RoomEditModal.jsx
plourenco: Thanks!

indico/modules/rb/client/js/common/rooms/edit/RoomEditModal.jsx
plourenco: I don't see any reason either. Initially, it was because the function was going to be placed somewhere else. I'll move it outside.

indico/modules/rb/client/js/common/rooms/edit/RoomEditModal.jsx
```diff
@@ -0,0 +1,438 @@
...
+      validate={formValidation}
+    />
+  );
+}
+
+RoomEditModal.propTypes = {
+  roomId: PropTypes.number,
+  locationId: PropTypes.number,
+  onClose: PropTypes.func.isRequired,
+  afterCreation: PropTypes.func,
```
ThiefMaster, line 431: Isn't this a `bool`?

Issue #4412, opened 2020-04-16
Make rendering of Category UI views more plugin customization friendly


marengeti:
Implementing Category-related UI features for the Indico.UN plugin requires a numerous changes and adjustments to the current core's UI rendering for different views. 

A shortlist of currently needed adjustments are:
- blockify the Category toolbar template code
- parametrize the big utility get_category_timetable() with an additional callable to allow for arbitrary custom filtering as needed per case
- factor the big call to get_category_timetable() in RHCategoryOveriew  out of the _process() method, for easier overriding of only this key method, without needing to redefine the full _process()
- make the params dict that is passed to the rendering template to be an instance attribute, so it can be added to conveniently from any method in the RH for the view, and so not requiring to redefine whole methods when all is needed is to add a parameter to pass to the template.

Will propose an implementation to each of the above, that will not change core behavior in any way.

PR #4413, opened 2020-04-16
#4412 Make Category UI views more plugin friendly

marengeti:
Several non-behaviour-modifying changes to RH and template code for Category view rendering, as listed in #4412.

Overall description of changes in #4412. 

I was trying to be consistent with your code base, but please feel free to adjust any variable/parameter/template-block name chosen here as you feel is best. 

Another small aspect -- if you feel a new method is not warranted for this, the new _get_data() method on RHCategoryOverview could in practice be made part of _process_args (in base RH._process_args, the docstring actually suggests that is where data fetches should probably go). So, for example, could just make the call within _process_args(), setting the result on an instance parameter, such as self.info. However, I feel given that this is a rather hefty db data fetch & process, I feel everything is better off to having it factored out as a completely separate customizable/optimizable/cacheable method.

### DIFF
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -316,6 +316,7 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
                   'atom_feed_title': _('Events of "{}"').format(self.category.title)}
         params.update(get_base_ical_parameters(session.user, 'category',
                                                '/export/categ/{0}.ics'.format(self.category.id), {'from': '-31d'}))
+
         if not self.category.is_root:
             return WPCategory.render_template('display/category.html', self.category, **params)
 
@@ -424,7 +425,7 @@ class RHXMLExportCategoryInfo(RH):
 class RHCategoryOverview(RHDisplayCategoryBase):
     """Display the events for a particular day, week or month"""
 
-    def _get_data(self):
+    def _get_timetable(self):
         return get_category_timetable([self.category.id], self.start_dt, self.end_dt,
                                       detail_level=self.detail, tz=self.category.display_tzinfo,
                                       from_categ=self.category, grouped=False)
@@ -457,7 +458,7 @@ class RHCategoryOverview(RHDisplayCategoryBase):
             self.end_dt = self.start_dt + relativedelta(months=1)
 
     def _process(self):
-        info = self._get_data()
+        info = self._get_timetable()
         events = info['events']
 
         # Only categories with icons are listed in the sidebar
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -299,23 +299,21 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
         managers = sorted(self.category.get_manager_list(), key=attrgetter('principal_type.name', 'name'))
 
         threshold_format = '%Y-%m'
-        params = {
-            'event_count': len(events),
-            'events_by_month': events_by_month,
-            'format_event_date': self.format_event_date,
-            'future_event_count': future_event_count,
-            'show_future_events': show_future_events,
-            'future_threshold': future_threshold.strftime(threshold_format),
-            'happening_now': self.happening_now,
-            'is_recent': self.is_recent,
-            'managers': managers,
-            'past_event_count': past_event_count,
-            'show_past_events': show_past_events,
-            'past_threshold': past_threshold.strftime(threshold_format),
-            'json_ld': map(serialize_event_for_json_ld, json_ld_events),
-            'atom_feed_url': url_for('.export_atom', self.category),
-            'atom_feed_title': _('Events of "{}"').format(self.category.title)
-        }
+        params = {'event_count': len(events),
+                  'events_by_month': events_by_month,
+                  'format_event_date': self.format_event_date,
+                  'future_event_count': future_event_count,
+                  'show_future_events': show_future_events,
+                  'future_threshold': future_threshold.strftime(threshold_format),
+                  'happening_now': self.happening_now,
+                  'is_recent': self.is_recent,
+                  'managers': managers,
+                  'past_event_count': past_event_count,
+                  'show_past_events': show_past_events,
+                  'past_threshold': past_threshold.strftime(threshold_format),
+                  'json_ld': map(serialize_event_for_json_ld, json_ld_events),
+                  'atom_feed_url': url_for('.export_atom', self.category),
+                  'atom_feed_title': _('Events of "{}"').format(self.category.title)}
         params.update(get_base_ical_parameters(session.user, 'category',
                                                '/export/categ/{0}.ics'.format(self.category.id), {'from': '-31d'}))
         if not self.category.is_root:
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -299,8 +299,7 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
         managers = sorted(self.category.get_manager_list(), key=attrgetter('principal_type.name', 'name'))
 
         threshold_format = '%Y-%m'
-        params = self.params
-        params.update({
+        params = {
             'event_count': len(events),
             'events_by_month': events_by_month,
             'format_event_date': self.format_event_date,
@@ -316,7 +315,7 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
             'json_ld': map(serialize_event_for_json_ld, json_ld_events),
             'atom_feed_url': url_for('.export_atom', self.category),
             'atom_feed_title': _('Events of "{}"').format(self.category.title)
-        })
+        }
         params.update(get_base_ical_parameters(session.user, 'category',
                                                '/export/categ/{0}.ics'.format(self.category.id), {'from': '-31d'}))
         if not self.category.is_root:
@@ -478,8 +477,7 @@ class RHCategoryOverview(RHDisplayCategoryBase):
                     -mktime(event.first_occurence_start_dt.timetuple()) if ongoing else event.start_dt.time())
         events = sorted(events, key=_event_sort_key)
 
-        params = self.params
-        params.update({
+        params = {
             'detail': self.detail,
             'period': self.period,
             'subcategories': subcategories,
@@ -492,7 +490,7 @@ class RHCategoryOverview(RHDisplayCategoryBase):
             'previous_year_url': self._other_day_url(self.start_dt - relativedelta(years=1)),
             'next_year_url': self._other_day_url(self.start_dt + relativedelta(years=1)),
             'mathjax': True
-        })
+        }
 
         if self.detail != 'event':
             cte = self.category.get_protection_parent_cte()
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -299,7 +299,8 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
         managers = sorted(self.category.get_manager_list(), key=attrgetter('principal_type.name', 'name'))
 
         threshold_format = '%Y-%m'
-        params = {'event_count': len(events),
+        params = self.params
+        params.update({'event_count': len(events),
                   'events_by_month': events_by_month,
                   'format_event_date': self.format_event_date,
                   'future_event_count': future_event_count,
@@ -313,10 +314,9 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
                   'past_threshold': past_threshold.strftime(threshold_format),
                   'json_ld': map(serialize_event_for_json_ld, json_ld_events),
                   'atom_feed_url': url_for('.export_atom', self.category),
-                  'atom_feed_title': _('Events of "{}"').format(self.category.title)}
+                  'atom_feed_title': _('Events of "{}"').format(self.category.title)})
         params.update(get_base_ical_parameters(session.user, 'category',
                                                '/export/categ/{0}.ics'.format(self.category.id), {'from': '-31d'}))
-
         if not self.category.is_root:
             return WPCategory.render_template('display/category.html', self.category, **params)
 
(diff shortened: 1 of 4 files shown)


ThiefMaster:
- While those are a lot of blocks, I don't see any reason against adding them (but I'd add indentation for their contents later, even though that will make the diff uglier of course)
- I'm fine with `_get_data`
- `includible`: the name sounds a bit strange (i think `includable` would be more correct), but in principle it's OK, even though it's the kind of code that may get removed in the future if someone sees it and realizes it's not being used. so maybe there should be a comment mentioning that it might be used by plugins calling the function..
- I really don't want this `self.params` thing in `RH`. What about moving more code from `_process` into a method? For example, you could move pretty much all of the `_process` method of `RHDisplayCategory` into a `_get_display_params` method that returns the params in the end. Like this you can then patch the method, call the original one and modify its return value. Moving `event_query = ...` into a `_get_event_query` method might also be helpful for you if you want to modify the events queried for display (e.g. to hide certain ones).

marengeti:
> * While those are a lot of blocks, I don't see any reason against adding them (but I'd add indentation for their contents later, even though that will make the diff uglier of course)

I did not indent in the template because I see `block` really as part and parcel of the root element inside it -- and, it is not written out, so the output would have really bizarre indenting (unless reformatted by the consumer). So, as a rule, I prefer to favour logically correct *output* indenting. But, that is totally a personal preference... please let me know exactly how you would like to have it, or just adjust directly.

> * I'm fine with `_get_data`

Very good ;-) Also, with the name?

> * `includible`: the name sounds a bit strange (i think `includable` would be more correct), but in principle it's OK, even though it's the kind of code that may get removed in the future if someone sees it and realizes it's not being used. so maybe there should be a comment mentioning that it might be used by plugins calling the function..

The two are synonyms, took what seems to me the most commonly used. Am ok with either, or any other!

> * I really don't want this `self.params` thing in `RH`. What about moving more code from `_process` into a method? For example, you could move pretty much all of the `_process` method of `RHDisplayCategory` into a `_get_display_params` method that returns the params in the end. Like this you can then patch the method, call the original one and modify its return value. Moving `event_query = ...` into a `_get_event_query` method might also be helpful for you if you want to modify the events queried for display (e.g. to hide certain ones).

OK, RH is base to many others of course, not just in Category module, which may be one of your objections? I think, though, that we will need it also in all the other modules!

Having a separate method to build this dict (presumably, in one place) seems much less convenient, as it would mean that if there is some need elsewhere in the RH to add a template param, it must somehow ensure that that name/value is passed on to this method, probably by setting as some multiple attribute on the RH instance...

To illustrate this, look at for example, RHDisplayCategory._process() -- the params dict there sets 16 or so entries on params, and most of them or not attrs on self. To get these values out to a separate method means a lot more code changes, and a lot more intermediate storage of data. 

In comparison, in RHCategoryOverview._process(), the params there is set up from almost exclusively attrs on self. With a self.params paradigm, this could presumably be simplified, and most likely remove to need of some of these attrs of self in the first place (their values being recorded directly on self.params where they are needed).

I would suggest maybe a more specific name, to reduce ambiguity given the wider context e.g. self.template_params or self.render_params? And, adopt it consistently as a simple flexible pattern for all RH classes?

The idea for a separate _get_query() is completely fine, as a further refinement of _get_data()? So, as protocol, it is always called only from within _get_data()? Thus, offering flexibility to only redefine it, if that would be enough, to to redefine _get_data() entirely? Am totally ok with that, just that I have not yet needed that specifically as yet.

ThiefMaster:
> so the output would have really bizarre indenting (unless reformatted by the consumer)

We generally don't care much about how the indentation of the HTML looks like. It's not meant to be read by humans - you have the browser's developer tools for that. Having clearly structured source files is more important IMO.

> Very good ;-) Also, with the name?

It didn't sound too strange to me and I couldn't think of something better when reading it, so yes.

> To illustrate this, look at for example, `RHDisplayCategory._process()` -- the params dict there sets 16 or so entries on params, and most of them or not attrs on self. To get these values out to a separate method means a lot more code changes, and a lot more intermediate storage of data.

I meant moving pretty much everything from there into the new method. something like this:

```python
    def _get_display_params(self):
        # Current events, which are always shown by default are events of this month and of the previous month.
        # If there are no events in this range, it will include the last and next month containing events.
        past_threshold = self.now - relativedelta(months=1, day=1, hour=0, minute=0)
        # ...
        params = {...}
        params.update(get_base_ical_parameters(session.user, 'category',
                                               '/export/categ/{0}.ics'.format(self.category.id), {'from': '-31d'}))
        return params

    def _process(self):
        params = self._get_display_params()
        if not self.category.is_root:
            return WPCategory.render_template('display/category.html', self.category, **params)
        news = get_recent_news()
        upcoming_events = get_upcoming_events()
        return WPCategory.render_template('display/root_category.html', self.category, news=news,
                                          upcoming_events=upcoming_events, **params)
```

As an alternative, what do you think about calling a signal from `WPJinjaMixin.render_template()` that can be used to modify the context passed to the params?

marengeti:
> Having clearly structured source files is more important IMO.

Fine with me. I will then indent all the blocks.
 
> I meant moving pretty much everything from there into the new method. something like this:

Such an external method will satisfy this need just fine. Still it does have a couple disadvantages to the simple pattern of setting a dedicated dict (or whatever it may evolve to be later) on each RH instance, and putting entries on that dict _as you go_, while doing whatever processing, in the context of that processing. So, for example, we add a url query param e.g. `openonly`, that we want to process and pass a resulting value to the template. Well, processing is supposed to be in _process_args(), and all that would be needed is:  

```
class CustomRHCategoryOverview(RHCategoryOverview):

    def _process_args(self):
        self.params['openonly'] = True if request.args.get('openonly', 'false') == 'true' else False
        super(_RHCategoryOverview, self)._process_args()
```

If we have a separate method, and follow 'protocol', we would need to do something like, so adding need to touch/override more methods unnecessarily:
```
class CustomRHCategoryOverview(RHCategoryOverview):

    def _process_args(self):
        self.openonly = True if request.args.get('openonly', 'false') == 'true' else False
        super(_RHCategoryOverview, self)._process_args()

    def _get_display_params(self):
        params = super(CustomRHCategoryOverview, self)._get_display_params()
        params['openonly'] = self.openonly
        return params
```

An additional consideration, is that what if the method _get_display_params() is called multiple times? Will it be the *same* dict instance that is returned, or a different one each time? This usually is not a concern, but it may lead to issues. Having self.params cuts the chase on all such concerns, as can be referred to as many times as needed, from any method, always same instance.

> As an alternative, what do you think about calling a signal from `WPJinjaMixin.render_template()` that can be used to modify the context passed to the params?

That is less direct, so adds opacity (and coding inconvenience). To what gain? Are we aware of a need where a listening "3rd party" needs to *adjust* the context to any "render template" event? As I see now, the added complication is not justifiable?

ThiefMaster:
The signal would be explicit in giving third parties a way to access/modify the data passed to templates, so we wouldn't need to add code that's likely to be refactored away (like your "display params" approach) by someone not realizing there's thirdparty code relying on it (and honestly, I don't like the idea of having this kind of "implicit" extensibility especially if it results in code in the core that, in my opinion, is less clean).

And from what I can see this signal would work well for your use case...

marengeti:
> The signal would be explicit in giving third parties a way to access/modify the data passed to templates, so we wouldn't need to add code that's likely to be refactored away (like your "display params" approach) by someone not realizing there's thirdparty code relying on it 

My approach is to simply add a self.render_params dict to all RH. 

You suggested adding a dedicated method _get_display_params() to RH. As I tried to say, yes, this will work, but implies more code changes, a little less direct, and so on (my comparison of the two approaches for an example case above). 

I fail to see how the dict attr on self adds to, as you say, 'a "kind of "implicit" extensibility especially if it results in code in the core that, in my opinion, is less clean).' ?

For the signal-based approach, i think this brings even more overhead than the method approach... and I do not see any gain for it? Even in core code? Is it possible to show how the above simple example (adding an 'openonly' url query param) to a view will be code in such a signal-based approach?

ThiefMaster:
> You suggested adding a dedicated method _get_display_params() to RH. As I tried to say, yes, this will work, but implies more code changes, a little less direct, and so on (my comparison of the two approaches for an example case above).

No, I meant adding one to the category-related RHs where you need it for easier overriding. Not to all RHs or the base RH class.

ThiefMaster:
In the majority of our code we don't have `params = {...}` at all, we just pass some keyword args to render_template. So in all those places the new dict you propose would not really help.

And I can't really see us starting to do something like this all the time:

```
params = self.render_params
params.update(...)
return WPWhatever.render_template(..., **params)
```

instead of the much simpler

```
return WPWhatever.render_template(...,, foo=foo, bar=bar, ...)
```

marengeti:
> In the majority of our code we don't have `params = {...}` at all, we just pass some keyword args to render_template. So in all those places the new dict you propose would not really help.
> 
> And I can't really see us starting to do something like this all the time:

Well, in these two views there were many params being added. But an inheriting RH only needs to add one... so it needs acces to *that* params dict. So, to allow such access, just moving it as a an attr on self. You do not have to set self.render_params to any local var, i did that to minimize change to the current core cat-views code, e.g.: 

```
self.render_params.update(...)
return WPWhatever.render_template(..., **self.render_params)
```

> instead of the much simpler
> ```
> return WPWhatever.render_template(...,, foo=foo, bar=bar, ...)
> ```

You have 16 or so params in those 2 Category views? 

Anyhow, if the self attr or the method are not on base RH, but only on RHCategoryBase or so, am perfectly fine with that. 

Still would like to see what the signal-based code for the `openonly` extra param example would look like?

ThiefMaster:
Actually for this particular param you need absolutely no python code change ad all. You can just access `request.args.openonly` directly in the template...

marengeti:
> Actually for this particular param you need absolutely no python code change ad all. You can just access `request.args.openonly` directly in the template...

That's true. I think that would be definitely less clean, but it will work. I can go along with that.

ThiefMaster:
But the flask `request` is always available in jinja templates... so you can just access it without having to pass it explicitly.

marengeti:
> But the flask `request` is always available in jinja templates... so you can just access it without having to pass it explicitly.

Ok, that does work. But again, that is not what template code should be doing. In any case, am fine to workaround this this in this way. Will revert the self.params changes back.

marengeti:
OK, done the mentioned changes:
- extended comment to mention that includible callable may be in use by client/plugin code
- indented cat display/base.html blocks
- dropped the RH.params dict attribute pattern

indico/modules/categories/controllers/display.py
```diff
@@ -299,24 +299,25 @@ def _process(self):
...
-                  'happening_now': self.happening_now,
-                  'is_recent': self.is_recent,
-                  'managers': managers,
-                  'past_event_count': past_event_count,
-                  'show_past_events': show_past_events,
-                  'past_threshold': past_threshold.strftime(threshold_format),
-                  'json_ld': map(serialize_event_for_json_ld, json_ld_events),
-                  'atom_feed_url': url_for('.export_atom', self.category),
-                  'atom_feed_title': _('Events of "{}"').format(self.category.title)}
+        params = {
```
ThiefMaster, line 302: This is just a format change now so I'd fully omit it

marengeti: Fine, omitted.

indico/modules/categories/controllers/display.py
```diff
@@ -453,8 +457,7 @@ def _process_args(self):
             self.end_dt = self.start_dt + relativedelta(months=1)
 
     def _process(self):
-        info = get_category_timetable([self.category.id], self.start_dt, self.end_dt, detail_level=self.detail,
-                                      tz=self.category.display_tzinfo, from_categ=self.category, grouped=False)
+        info = self._get_data()
```
ThiefMaster, line 460: Now that I'm looking at this again, `_get_timetable` would probably be a better name ;)

indico/modules/categories/controllers/display.py
```diff
@@ -316,7 +316,6 @@ def _process(self):
                   'atom_feed_title': _('Events of "{}"').format(self.category.title)}
         params.update(get_base_ical_parameters(session.user, 'category',
                                                '/export/categ/{0}.ics'.format(self.category.id), {'from': '-31d'}))
-
```
ThiefMaster, line 319: could you undo this one as well?

indico/modules/categories/templates/display/base.html
```diff
@@ -18,64 +18,87 @@ <h1 class="category-title {{ category_title_classes }}">
...
+                    {% endblock %}
+
+                    {% block cat_manage %}
+                        {% if category.can_manage(session.user) %}
+                            <a href="{{ url_for('categories.manage_content', category) }}"
+                               class="i-button icon-edit"
+                               title="{% trans %}Manage category{% endtrans %}"></a>
+                        {% endif %}
+                    {% endblock %}
+
```
ThiefMaster, line 86: I don't think this empty line was here before

indico/modules/categories/templates/display/base.html
```diff
@@ -18,64 +18,87 @@ <h1 class="category-title {{ category_title_classes }}">
...
+                            <button type="button"
+                                    class="i-button fav-button icon-only icon-bookmark {{ 'enabled' if category in session.user.favorite_categories }}"
+                                    data-href="{{ url_for('users.user_favorites_category_api', category) }}"
+                                    data-favorites-href="{{ url_for('users.user_favorites') }}"></button>
+                        </div>
+                    {% endif %}
+            {% endblock %}
+
         </div>
+        {% endblock %}
```
ThiefMaster, line 101: There's something wrong with the indentation here

indico/modules/categories/controllers/display.py
marengeti: OK

indico/modules/categories/controllers/display.py
marengeti: OK, i'll change it to this.

indico/modules/categories/templates/display/base.html
marengeti: There are also other new empty lines above (between blocks, i think). I'll remove them.

ThiefMaster: Having an empty line between two blocks seems reasonable. Having one right before the closing tag not so much

marengeti: Removed them.

indico/modules/categories/templates/display/base.html
marengeti: Corrected.

marengeti:
closes #4412

PR #4414, opened 2020-04-20
Make "Go to timeline" nicer

### LINKED ISSUE #4407: Make "Go to timeline" in contribution view nicer
Currently we only have a somewhat ugly/unformatted "Go to timeline (paper)" link there there. Maybe we could show a box similar to what we have for submissions, and also indicate the current state?

panagiotappl:
closes #4407 
image

### DIFF
--- a/indico/modules/events/contributions/models/contributions.py
+++ b/indico/modules/events/contributions/models/contributions.py
@@ -502,6 +502,16 @@ class Contribution(DescriptionMixin, ProtectionManagersMixin, LocationMixin, Att
         submitted_for = {editable.type.name for editable in self.editables}
         return [editable_type for editable_type in self.event.editable_types if editable_type not in submitted_for]
 
+    @property
+    def enabled_editables(self):
+        """Return all submitted editables with enabled types."""
+        from indico.modules.events.editing.settings import editing_settings
+        if not self.event.has_feature('editing'):
+            return []
+
+        enabled_editable_types = editing_settings.get(self.event, 'editable_types')
+        return [editable for editable in self.editables if editable.type.name in enabled_editable_types]
+
     def is_paper_reviewer(self, user):
         return user in self.paper_content_reviewers or user in self.paper_layout_reviewers
 
--- a/indico/modules/events/papers/templates/_contributions.html
+++ b/indico/modules/events/papers/templates/_contributions.html
@@ -39,6 +39,8 @@
 {% macro render_editables_section(contribution) %}
     {% set can_create_editable = contribution.allowed_types_for_editable|length > 0 %}
     {% set can_submit = contribution.is_user_associated(session.user, check_abstract=true) %}
+    {% set enabled_editables = contribution.enabled_editables %}
+    {% set has_enabled_types = enabled_editables|length > 0 %}
 
     {% if can_submit %}
         <section>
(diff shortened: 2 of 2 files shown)


indico/modules/events/papers/templates/_contributions.html
```diff
@@ -47,15 +47,22 @@ <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
...
-            {% for editable in contribution.editables -%}
-                <div>
-                    <a href="{{ url_for('event_editing.editable', editable) }}">
-                        {%- trans type=editable.type.name %}
-                            Go to timeline ({{ type }})
-                        {% endtrans -%}
-                    </a>
-                </div>
-            {%- endfor %}
+            <div class="action-box">
```
ThiefMaster, line 50: How does it look like if there are no editables for the contribution?

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -47,15 +47,22 @@ <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
...
+                {% for editable in contribution.editables -%}
+                    {% set latest_revision = editable.revisions|last %}
+                    {% set latest_state = latest_revision.final_state or latest_revision.initial_state %}
+                    <div class="section f-j-space-between">
+                        <div class="i-tag {{ latest_state.css_class }}">
+                            {%- trans type=editable.type.title %} {{ type }} {% endtrans -%}: {% trans state=latest_state.title %} {{ state }} {% endtrans -%}
```
ThiefMaster, line 56: None of these trans blocks here are needed

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -47,15 +47,22 @@ <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
...
+                        </div>
+                        <div>
+                            <a href="{{ url_for('event_editing.editable', editable) }}">
+                                {%- trans %} Go to timeline {% endtrans -%}
```
ThiefMaster, line 60: don't put surrounding space inside trans blocks; it gets stripped anyway

indico/modules/events/papers/templates/_contributions.html
panagiotappl: Why is it needed [here](https://github.com/indico/indico/blob/master/indico/modules/events/papers/templates/_contributions.html#L54)?

ThiefMaster: There you have text that can be translated: `Go to timeline: {type}`

But in this case you only have a variable (which is already a translatable string itself)

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -47,15 +47,24 @@ <h3 class="icon-file">{% trans %}Paper{% endtrans %}</h3>
...
                 </div>
             </div>
-            {% for editable in contribution.editables -%}
-                <div>
-                    <a href="{{ url_for('event_editing.editable', editable) }}">
-                        {%- trans type=editable.type.name %}
-                            Go to timeline ({{ type }})
-                        {% endtrans -%}
-                    </a>
+            {% if contribution.editables %}
```
ThiefMaster, line 50: Shouldn't we also take `contribution.allowed_types_for_editable` into account here, e.g. using a new property on Contribution that returns the editable that both exist but are also for enabled types?

I guess it's an edge case but it's strange to show editable types there even if they've been disabled later..

indico/modules/events/contributions/models/contributions.py
```diff
@@ -502,6 +502,15 @@ def allowed_types_for_editable(self):
         submitted_for = {editable.type.name for editable in self.editables}
         return [editable_type for editable_type in self.event.editable_types if editable_type not in submitted_for]
 
+    @property
+    def enabled_editable_types(self):
```
ThiefMaster, line 506: Not sure about the naming... What about something like `enabled_editables` giving you directly the list of editables which are enabled?

Also a docstring to explain in more detail what this does

```
Return the active editables submitted for the contribution.

These are all the editables which have been submitted, but only if
their type is still enabled for the event.
```

(not 100% satisfied with that text either, but can't think of something better right now)

panagiotappl: Why not use the word "type" in there? This is supposed to return the names of all the editable types that have been submitted and are also enabled at the moment.

ThiefMaster: Yeah, if you keep returning type names that's fine - I'd have returned the editables themselves instead (and wrote the docstring based on that suggestion)

panagiotappl: What's the advantage of returning the entire editable?

ThiefMaster: Being able to use it directly in the template instead of this:

```
                    {% for editable in contribution.editables -%}
                        {% if editable.type.name in enabled_editable_types %}
```

PS: Jinja supports `{% for foo in bar if whatever %}` in case you decide to not change it.

panagiotappl: Maybe something like this for the docstring?
```
Return all the submitted editabled with enabled types.

These are all the editables which have been submitted and their type
is still enabled for the event.
```

ThiefMaster: "Return all submitted editables for enabled types."

i think we don't even need the rest in this case since it's already clear :)

PR #4419, opened 2020-04-21
Events visibility

openprojects:
Show/hide Events in Category based on can_display

### DIFF
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -348,8 +348,10 @@ class RHEventList(RHDisplayCategoryEventsBase):
         after = self._parse_year_month(request.args.get('after'))
         if before is None and after is None:
             raise BadRequest('"before" or "after" parameter must be specified')
+        hidden_event_ids = {e.id for e in self.category.get_hidden_events(user=session.user)}
         event_query = (Event.query.with_parent(self.category)
                        .options(*self._event_query_options)
+                       .filter(Event.id.notin_(hidden_event_ids))
                        .order_by(Event.start_dt.desc(), Event.id.desc()))
         if before:
             event_query = event_query.filter(Event.start_dt < before)
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -263,12 +263,16 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
         # If there are no events in this range, it will include the last and next month containing events.
         past_threshold = self.now - relativedelta(months=1, day=1, hour=0, minute=0)
         future_threshold = self.now + relativedelta(months=1, day=1, hour=0, minute=0)
+        hidden_event_ids = tuple([event.id for event in Event.get_hidden_events(self.category.id)
+                                  if not event.can_display()])
         next_event_start_dt = (db.session.query(Event.start_dt)
-                               .filter(Event.start_dt >= self.now, Event.category_id == self.category.id)
+                               .filter(Event.start_dt >= self.now, Event.category_id == self.category.id,
+                                       ~Event.id.in_(hidden_event_ids))
                                .order_by(Event.start_dt.asc(), Event.id.asc())
                                .first() or (None,))[0]
         previous_event_start_dt = (db.session.query(Event.start_dt)
-                                   .filter(Event.start_dt < self.now, Event.category_id == self.category.id)
+                                   .filter(Event.start_dt < self.now, Event.category_id == self.category.id,
+                                           ~Event.id.in_(hidden_event_ids))
                                    .order_by(Event.start_dt.desc(), Event.id.desc())
                                    .first() or (None,))[0]
         if next_event_start_dt is not None and next_event_start_dt > future_threshold:
@@ -277,6 +281,7 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
             past_threshold = previous_event_start_dt.replace(day=1, hour=0, minute=0)
         event_query = (Event.query.with_parent(self.category)
                        .options(*self._event_query_options)
+                       .filter(~Event.id.in_(hidden_event_ids))
                        .order_by(Event.start_dt.desc(), Event.id.desc()))
         past_event_query = event_query.filter(Event.start_dt < past_threshold)
         future_event_query = event_query.filter(Event.start_dt >= future_threshold)
@@ -287,6 +292,7 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
 
         future_event_count = future_event_query.count()
         past_event_count = past_event_query.count()
+        hiding_events = len(hidden_event_ids) > 0
 
         if not session.user and future_event_count:
             json_ld_events = json_ld_events + future_event_query.all()
@@ -311,6 +317,7 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
                   'past_event_count': past_event_count,
                   'show_past_events': show_past_events,
                   'past_threshold': past_threshold.strftime(threshold_format),
+                  'hiding_events': hiding_events,
                   'json_ld': map(serialize_event_for_json_ld, json_ld_events),
                   'atom_feed_url': url_for('.export_atom', self.category),
                   'atom_feed_title': _('Events of "{}"').format(self.category.title)}
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -263,16 +263,15 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
         # If there are no events in this range, it will include the last and next month containing events.
         past_threshold = self.now - relativedelta(months=1, day=1, hour=0, minute=0)
         future_threshold = self.now + relativedelta(months=1, day=1, hour=0, minute=0)
-        hidden_event_ids = tuple([event.id for event in Event.get_hidden_events(self.category.id)
-                                  if not event.can_display()])
+        hidden_event_ids = {e.id for e in self.category.get_hidden_events(user=session.user)}
         next_event_start_dt = (db.session.query(Event.start_dt)
                                .filter(Event.start_dt >= self.now, Event.category_id == self.category.id,
-                                       ~Event.id.in_(hidden_event_ids))
+                                       Event.id.notin_(hidden_event_ids))
                                .order_by(Event.start_dt.asc(), Event.id.asc())
                                .first() or (None,))[0]
         previous_event_start_dt = (db.session.query(Event.start_dt)
                                    .filter(Event.start_dt < self.now, Event.category_id == self.category.id,
-                                           ~Event.id.in_(hidden_event_ids))
+                                           Event.id.notin_(hidden_event_ids))
                                    .order_by(Event.start_dt.desc(), Event.id.desc())
                                    .first() or (None,))[0]
         if next_event_start_dt is not None and next_event_start_dt > future_threshold:
@@ -281,7 +280,7 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
             past_threshold = previous_event_start_dt.replace(day=1, hour=0, minute=0)
         event_query = (Event.query.with_parent(self.category)
                        .options(*self._event_query_options)
-                       .filter(~Event.id.in_(hidden_event_ids))
+                       .filter(Event.id.notin_(hidden_event_ids))
                        .order_by(Event.start_dt.desc(), Event.id.desc()))
         past_event_query = event_query.filter(Event.start_dt < past_threshold)
         future_event_query = event_query.filter(Event.start_dt >= future_threshold)
@@ -292,7 +291,7 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
 
         future_event_count = future_event_query.count()
         past_event_count = past_event_query.count()
-        hiding_events = len(hidden_event_ids) > 0
+        has_hidden_events = bool(hidden_event_ids)
 
         if not session.user and future_event_count:
             json_ld_events = json_ld_events + future_event_query.all()
@@ -317,7 +316,7 @@ class RHDisplayCategory(RHDisplayCategoryEventsBase):
                   'past_event_count': past_event_count,
                   'show_past_events': show_past_events,
                   'past_threshold': past_threshold.strftime(threshold_format),
-                  'hiding_events': hiding_events,
+                  'has_hidden_events': has_hidden_events,
                   'json_ld': map(serialize_event_for_json_ld, json_ld_events),
                   'atom_feed_url': url_for('.export_atom', self.category),
                   'atom_feed_title': _('Events of "{}"').format(self.category.title)}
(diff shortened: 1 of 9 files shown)


indico/modules/events/models/events.py
```diff
@@ -669,6 +669,20 @@ def can_lock(self, user):
         """Check whether the user can lock/unlock the event"""
         return user and (user.is_admin or user == self.creator or self.category.can_manage(user))
 
+    def can_display(self):
+        """Check whether the user can display the event in the category.
+        To be replaced with new permission mechanism: can('display', self, session.user)."""
+        return self.visibility is None or self.can_manage(session.user)
```
ThiefMaster, line 675: `self.visibility != 0` since any other visibility includes the current category as well

indico/modules/events/models/events.py
```diff
@@ -669,6 +669,20 @@ def can_lock(self, user):
         """Check whether the user can lock/unlock the event"""
         return user and (user.is_admin or user == self.creator or self.category.can_manage(user))
 
+    def can_display(self):
+        """Check whether the user can display the event in the category.
+        To be replaced with new permission mechanism: can('display', self, session.user)."""
```
ThiefMaster, line 674: i don't think this belongs here; but without this line the docstring is fine

indico/modules/events/models/events.py
ThiefMaster, line 675: Don't use `session.user` in here; pass it as an argument instead (like we do for the other `can_*` methods)

indico/modules/categories/controllers/display.py
```diff
@@ -287,6 +292,7 @@ def _process(self):
 
         future_event_count = future_event_query.count()
         past_event_count = past_event_query.count()
+        hiding_events = len(hidden_event_ids) > 0
```
ThiefMaster, line 295: `has_hidden_events = bool(hidden_event_ids)`

indico/modules/categories/controllers/display.py
```diff
@@ -277,6 +281,7 @@ def _process(self):
             past_threshold = previous_event_start_dt.replace(day=1, hour=0, minute=0)
         event_query = (Event.query.with_parent(self.category)
                        .options(*self._event_query_options)
+                       .filter(~Event.id.in_(hidden_event_ids))
```
ThiefMaster, line 284: `Event.id.notin_(hidden_event_ids)`

indico/modules/categories/controllers/display.py
```diff
@@ -263,12 +263,16 @@ def _process(self):
         # If there are no events in this range, it will include the last and next month containing events.
         past_threshold = self.now - relativedelta(months=1, day=1, hour=0, minute=0)
         future_threshold = self.now + relativedelta(months=1, day=1, hour=0, minute=0)
+        hidden_event_ids = tuple([event.id for event in Event.get_hidden_events(self.category.id)
```
ThiefMaster, line 266: Better use a set instead of a tuple here; and move this method/logic into the Category model as you act on a category anyway.

    hidden_event_ids = {e.id for e in self.category.get_hidden_events()}

indico/modules/events/models/events.py
```diff
@@ -669,6 +669,20 @@ def can_lock(self, user):
...
+    @staticmethod
+    def get_hidden_events(category_id):
+        """Get all hidden events within the given category."""
+        category = Category.get(int(category_id), is_deleted=False)
+        return (Event.query.with_parent(category)
+                .filter(Event.category_id == category_id, Event.visibility == 0)
+                .order_by(Event.start_dt.desc(), Event.id.desc())
+                .all())
+
     def get_relative_event_ids(self):
```
ThiefMaster, line 686: I think you should exclude hidden events in this function as well (since you cannot easily check can_display there). But for this type of navigation (the next/prev/first/last buttons on top of an event display page) excluding those events is probably fine anyway.

indico/modules/events/models/events.py
```diff
@@ -669,6 +669,20 @@ def can_lock(self, user):
...
+        return self.visibility is None or self.can_manage(session.user)
+
+    @staticmethod
+    def get_hidden_events(category_id):
+        """Get all hidden events within the given category."""
+        category = Category.get(int(category_id), is_deleted=False)
+        return (Event.query.with_parent(category)
+                .filter(Event.category_id == category_id, Event.visibility == 0)
+                .order_by(Event.start_dt.desc(), Event.id.desc())
+                .all())
```
ThiefMaster, line 684: Why so complicated?

    Event.query.with_parent(category).filter_by(visibility=0)

indico/modules/categories/templates/management/_events_list.html
```diff
@@ -6,6 +6,11 @@
         <td class="i-table column-icon">
             <input type="checkbox" name="event_id" value="{{ event.id }}">
         </td>
+        <td class="i-table column-icon">
+            {% if (event.visibility == 0) %}
```
ThiefMaster, line 10: no `()` here

indico/modules/categories/templates/display/event_list.html
```diff
@@ -98,6 +110,12 @@ <h4 class="{% if month.is_current %}current-month{% endif %}">
                                 {{ event.get_verbose_title(show_speakers=is_lecture, show_series_pos=is_lecture) | striptags }}
                             </a>
                             <span class="protected">
+                                {% if (event.visibility == 0) %}
```
ThiefMaster, line 113: no `()` here

ThiefMaster:
What do you think about not checking for event management permissions but only category management permissions when determining whether to show hidden events anyway or not? This would allow making this much simpler as you could always hide events with `visibility==0` unless you are a category manager (ie you can do this fully on the query level without and extra queries etc)!

openprojects:
… but then any Conference Manager (and other custom role we will create, like “Security manager”) will not see the Event he should be able to manage…

___________________________________
Giorgio Pieretti (Mr.)
Information Systems Assistant
Division of Conference Management (DCM)
United Nations Office in Geneva (UNOG)
Palais des Nations, Room E-164
1211 Geneva 10
Tel. +41 22 917 2691
Mobile +41 79 444 39 43
e-mail: giorgio.pieretti@un.org<mailto:giorgio.pieretti@un.org>
website: www.unog.ch<http://www.unog.ch/>

ThiefMaster:
Yeah, depends on whether it's common to have managers that aren't category managers... I guess in your case this happens more often so having the event-level check probably indeed makes sense...

openprojects:
Exactly.
I proceed fixing the PR.
Thank you!

___________________________________
Giorgio Pieretti (Mr.)
Information Systems Assistant
Division of Conference Management (DCM)
United Nations Office in Geneva (UNOG)
Palais des Nations, Room E-164
1211 Geneva 10
Tel. +41 22 917 2691
Mobile +41 79 444 39 43
e-mail: giorgio.pieretti@un.org<mailto:giorgio.pieretti@un.org>
website: www.unog.ch<http://www.unog.ch/>

indico/modules/events/models/events.py
openprojects: Ok

indico/modules/events/models/events.py
openprojects: Agree, but I saw the "can_lock" that's why I decided to put it there. 
BTW, you want me to remove the "To be replaced with new permission ..." line and keep the rest as is?

indico/modules/events/models/events.py
openprojects: Ok

indico/modules/events/models/events.py
ThiefMaster: yes, removing this line is actually all this comment was about :)

indico/modules/categories/controllers/display.py
openprojects: Nicer :)

indico/modules/categories/controllers/display.py
openprojects: Ok

indico/modules/categories/controllers/display.py
openprojects: OK, so you want me to also move the can_display(user) check inside the get_hidden_events(), correct?

ThiefMaster: I think it makes sense, unless doing the check outside allows you to do some optimization... But right now I can't think of anything in particular, except that you could skip the call to this altogether in case the user is a category manager (this avoids the event ACL check for non-admin users since if someone is a category manager they are allowed to see everything anyway)

openprojects: OK, I'm fine with it.

indico/modules/events/models/events.py
openprojects: Not clear... so can I keep it as is?

ThiefMaster: .where((Event.category_id == self.category_id) & ~Event.is_deleted)

You need to add ` & (Event.visibility != 0)` here (and test it, I'm not sure if it works correctly due to how NULL behaves, you may need something more explicit such as `(Event.visibility.is_(None) | Event.visibility != 0))`

indico/modules/events/models/events.py
openprojects: Because I'm not yet used to this kind of query way... but I'll get to it :)
Changed.

indico/modules/events/models/events.py
openprojects: Wait... here I need to retrieve **only** the **invisible** ones... so visibility=0.
As you proposed:
hidden_events = Event.query.with_parent(self).filter_by(visibility=0)

indico/modules/categories/templates/management/_events_list.html
openprojects: Ok

indico/modules/categories/templates/display/event_list.html
openprojects: Ok

indico/modules/events/models/events.py
ThiefMaster: Check which function this comment is on ;)

openprojects: Ahhhh ok, I got it now! :)

indico/modules/categories/models/categories.py
```diff
@@ -486,6 +486,12 @@ def visible_categories_query(self):
         cte_query = Category.get_visible_categories_cte(self.id)
         return Category.query.join(cte_query, Category.id == cte_query.c.id)
 
+    def get_hidden_events(self, user=None):
+        """Get all hidden events within the given category and user."""
+        from indico.modules.events import Event
+        hidden_events = Event.query.with_parent(self).filter_by(visibility=0)
+        return filter(lambda e: not e.can_display(user), hidden_events)
```
ThiefMaster, line 493: Use a list comprehension here

indico/modules/categories/models/categories.py
```diff
@@ -486,6 +486,12 @@ def visible_categories_query(self):
         cte_query = Category.get_visible_categories_cte(self.id)
         return Category.query.join(cte_query, Category.id == cte_query.c.id)
 
+    def get_hidden_events(self, user=None):
+        """Get all hidden events within the given category and user."""
+        from indico.modules.events import Event
+        hidden_events = Event.query.with_parent(self).filter_by(visibility=0)
```
ThiefMaster, line 492: I'd name this `query` or call `.all()` on it (having a query object in something not named query is a bit strange)

indico/modules/events/models/events.py
```diff
@@ -683,16 +687,20 @@ def get_relative_event_ids(self):
...
         rv = (db.session.query(subquery.c.first, subquery.c.last, subquery.c.prev, subquery.c.next)
               .filter(subquery.c.id == self.id)
-              .one()
-              ._asdict())
-        if rv['first'] == self.id:
-            rv['first'] = None
-        if rv['last'] == self.id:
-            rv['last'] = None
+              .first())
+        if not rv:
```
ThiefMaster, line 696: I think you can avoid this case altogether by not applying the visibility filter in case, i.e. `(Event.visibility.is_(None) | (Event.visibility != 0) | (Event.id == self.id))`

openprojects: Looks much cleaner: I'm going to test it, tnx.

indico/modules/categories/models/categories.py
openprojects: Ok.

indico/modules/categories/models/categories.py
openprojects: Yep, added .all()

ThiefMaster:
You are not correctly checking for hidden events when loading past/future events (`RHEventList`).

ThiefMaster:
Please update the field description of the event visibility field to mention that "Invisible" also hides the event from the category list

indico/modules/categories/templates/management/_events_list.html
```diff
@@ -6,6 +6,11 @@
         <td class="i-table column-icon">
             <input type="checkbox" name="event_id" value="{{ event.id }}">
         </td>
+        <td class="i-table column-icon">
+            {% if event.visibility == 0 %}
+                <span class="icon-eye icon-eye-blocked"></span>
```
ThiefMaster, line 11: `title="{% trans %}Hidden{% endtrans %}"`

openprojects: Ok

openprojects:
> You are not correctly checking for hidden events when loading past/future events (`RHEventList`).
OK... from the category view I don't hit that class... where can I reproduce it?

ThiefMaster:
Just move the data long enough in the past that it's part of the "past events" that are loaded with a separate request

openprojects:
> Please update the field description of the event visibility field to mention that "Invisible" also hides the event from the category list

Sure. Should I also update the msgid in the .po files? and the relative translations too (BR,FR,ES)?

ThiefMaster:
no, we do that later when coming closer to a release

indico/modules/categories/models/categories.py
```diff
@@ -486,6 +486,12 @@ def visible_categories_query(self):
         cte_query = Category.get_visible_categories_cte(self.id)
         return Category.query.join(cte_query, Category.id == cte_query.c.id)
 
+    def get_hidden_events(self, user=None):
+        """Get all hidden events within the given category and user."""
+        from indico.modules.events import Event
+        hidden_events = Event.query.with_parent(self).filter_by(visibility=0).all()
+        return [event for event in hidden_events if not event.can_display(user)]
```
plourenco, line 493: Isn't this the same or more explicit to write?
```suggestion
        return [event for event in hidden_events if not event.can_manage(user)]
```

Since `can_display` is `self.visibility != 0 or self.can_manage(user)` and we have `filter_by(visibility=0)`.

openprojects: Yes, but this is just because we do not have a specific permission at the moment. 
Consider a "Custom role", or "Security role" or "Focal point role" (that you do not have but we are working on them). All these roles will not have "event manage" permission, but a specific "can display" (and in the future: can('display', self, session.user).

plourenco: Sounds good then.

openprojects: Well... there was... ask Adrian :)

ThiefMaster: Yeah I suggested removing it since the idea about adding `can(...)` is something for 3.1, i.e. a long time in the future ;)

openprojects:
Is there anything else I have to change/fix in the PR?

ThiefMaster:
no, looks good. i'll give it a last quick test tomorrow, add a changelog entry and then merge it

ThiefMaster:
I pushed some small fixes/improvements; the most important one was including `visibility` in the list of queried columns when querying the events - without this there was one extra query for each event, making this whole check very slow for large categories.

PR #4424, opened 2020-04-23
Add UI for List of editables

### LINKED ISSUE #4391: List of editables: UI
Implement components for the list of editables for a given type:

Ideally this should support infinite scrolling and windowed rendering for performance reasons

Use dummy data for now.

ParthS007:
closes #4391 

Screenshot from 2020-05-08 10-06-36
Screenshot from 2020-05-08 10-06-47
Screenshot from 2020-05-08 10-26-36

### DIFF
--- a/indico/modules/events/editing/client/js/editing/timeline/StateIndicator.jsx
+++ b/indico/modules/events/editing/client/js/editing/timeline/StateIndicator.jsx
@@ -8,6 +8,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {Label, Popup} from 'semantic-ui-react';
+import Palette from 'indico/utils/palette';
+import {EditableStatus} from '../../models';
+
+import './StateIndicator.module.scss';
 
 const colors = {
   replaced: 'grey',
@@ -18,33 +22,45 @@ const colors = {
   needs_submitter_changes: 'orange',
 };
 
-export default function StateIndicator({text, circular, tooltip, state}) {
-  const trigger = (
-    <Label size="tiny" color={colors[state]} circular={circular}>
-      {text}
-    </Label>
-  );
+const labelColors = {
+  replaced: Palette.grey,
+  needs_submitter_confirmation: Palette.yellow,
+  rejected: Palette.red,
+  accepted: Palette.green,
+  assigned: Palette.purple,
+  needs_submitter_changes: Palette.orange,
+};
+
+export default function StateIndicator({label, circular, tooltip, state}) {
+  const trigger = <Label size="tiny" color={colors[state]} circular={circular} />;
 
   return (
-    <Popup
-      position="bottom center"
-      trigger={trigger}
-      content={tooltip}
-      on="hover"
-      disabled={!tooltip}
-    />
+    <>
+      <Popup
+        position="bottom center"
+        trigger={trigger}
+        content={tooltip}
+        on="hover"
+        disabled={!tooltip}
+      />
+      {label && (
+        <div styleName="label-text" style={{color: labelColors[state]}}>
+          {EditableStatus[state]}
+        </div>
+      )}
+    </>
   );
 }
 
 StateIndicator.propTypes = {
-  text: PropTypes.string,
+  label: PropTypes.bool,
   circular: PropTypes.bool,
   tooltip: PropTypes.string,
   state: PropTypes.oneOf(Object.keys(colors)).isRequired,
 };
 
 StateIndicator.defaultProps = {
-  text: null,
+  label: false,
   circular: false,
   tooltip: null,
 };
--- a/indico/modules/events/editing/client/js/editing/timeline/StateIndicator.module.scss
+++ b/indico/modules/events/editing/client/js/editing/timeline/StateIndicator.module.scss
@@ -1,0 +1,13 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2020 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+.label-text {
+    margin-left: 5px;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
+}
--- a/indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
+++ b/indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
@@ -1,0 +1,177 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2020 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import dashboardURL from 'indico-url:event_editing.dashboard';
+// import manageEditableTypeListURL from 'indico-url:event_editing.manage_editable_type_list'; Uncomment when we have backend ready
+
+import React, {useState} from 'react';
+import {useParams} from 'react-router-dom';
+import {Button, Icon, Checkbox, Message, Search} from 'semantic-ui-react';
+import {Column, Table, SortDirection, WindowScroller} from 'react-virtualized';
+import _ from 'lodash';
+import {
+  TooltipIfTruncated,
+  ManagementPageSubTitle,
+  ManagementPageBackButton,
+} from 'indico/react/components';
+import {useNumericParam} from 'indico/react/util/routing';
+import {Translate} from 'indico/react/i18n';
+import {EditableType} from '../../models';
+import StateIndicator from '../../editing/timeline/StateIndicator';
+
+import './EditableList.module.scss';
+
+export default function EditableList() {
+  const list = []; // Replace it with list from server
+  const eventId = useNumericParam('confId');
+  const {type} = useParams();
+  const [sortBy, setSortBy] = useState('id');
+  const [sortDirection, setSortDirection] = useState('ASC');
+  const [sortedList, setSortedList] = useState(list); // list from server
+  const [checked, setChecked] = useState([]);
+  const hasCheckedRows = checked.length === 0;
+  const title = {
+    [EditableType.paper]: Translate.string('List of papers'),
+    [EditableType.slides]: Translate.string('List of slides'),
+    [EditableType.poster]: Translate.string('List of posters'),
+  }[type];
+  const columnHeaders = [
+    ['id', Translate.string('ID'), 60],
+    ['code', Translate.string('Code'), 80],
+    ['title', Translate.string('Title'), 600],
+    ['status', Translate.string('Status'), 150],
+    ['assigned', Translate.string('Assigned'), 400],
+  ];
+
+  // eslint-disable-next-line no-shadow
+  const _sortList = ({sortBy, sortDirection}) => {
+    const newList = _.sortBy(list, [sortBy]);
+    if (sortDirection === SortDirection.DESC) {
+      newList.reverse();
+    }
+    return newList;
+  };
+
+  // eslint-disable-next-line no-shadow
+  const _sort = ({sortBy, sortDirection}) => {
+    setSortBy(sortBy);
+    setSortDirection(sortDirection);
+    setSortedList(_sortList({sortBy, sortDirection}));
+  };
+
+  // eslint-disable-next-line react/prop-types
+  const renderCell = ({dataKey, rowIndex}) => {
+    return (
+      <TooltipIfTruncated>
+        <div index={rowIndex} styleName="rowcolumn-tooltip" role="gridcell">
+          {dataKey === 'assigned' && <Icon name="user outline" />}
+          {dataKey === 'status' ? (
+            <StateIndicator index={rowIndex} state={sortedList[rowIndex][dataKey]} circular label />
+          ) : (
+            sortedList[rowIndex][dataKey]
+          )}
+        </div>
+      </TooltipIfTruncated>
+    );
+  };
+
+  const toggleSelectAll = dataChecked => {
+    if (dataChecked) {
+      setChecked(sortedList.map(row => row.id));
+    } else {
+      setChecked([]);
+    }
+  };
+
+  const toggleSelectRow = dataIndex => {
+    const newRow = sortedList[dataIndex].id;
+    if (checked.includes(newRow)) {
+      setChecked(old => old.filter(row => row !== newRow));
+    } else {
+      setChecked(old => [...old, newRow]);
+    }
+  };
+
+  return (
+    <>
+      <ManagementPageSubTitle title={title} />
+      <ManagementPageBackButton url={dashboardURL({confId: eventId})} />
+      <div styleName="editable-topbar">
+        <div>
+          <Button disabled={hasCheckedRows} content={Translate.string('Assign')} />
+          <Button disabled={hasCheckedRows} content={Translate.string('Unassign')} />
+          <Button disabled={hasCheckedRows} content={Translate.string('Set status')} />
+          <Button
+            disabled={hasCheckedRows}
+            color="blue"
+            content={Translate.string('Assign to myself')}
+          />
+          <Button disabled={hasCheckedRows} content={Translate.string('Download all files')} />
+        </div>
+        <Search />
+      </div>
+      {sortedList.length ? (
+        <div styleName="editable-list">
+          <WindowScroller>
+            {({height, isScrolling, onChildScroll, scrollTop}) => (
+              <Table
+                autoHeight
+                width={1000}
+                styleName="table"
+                height={height}
+                isScrolling={isScrolling}
+                onScroll={onChildScroll}
+                headerHeight={30}
+                rowHeight={40}
+                sort={_sort}
+                sortBy={sortBy}
+                sortDirection={sortDirection}
+                rowCount={sortedList.length}
+                scrollTop={scrollTop}
+                rowGetter={({index}) => sortedList[index]}
+              >
+                <Column
+                  disableSort
+                  dataKey="checkbox"
+                  width={30}
+                  headerRenderer={() => (
+                    <Checkbox
+                      indeterminate={checked.length > 0 && checked.length < sortedList.length}
+                      checked={checked.length === sortedList.length}
+                      onChange={(e, data) => toggleSelectAll(data.checked)}
+                    />
+                  )}
+                  cellRenderer={({rowIndex}) => (
+                    <Checkbox
+                      checked={checked.includes(sortedList[rowIndex].id)}
+                      onChange={(e, data) => toggleSelectRow(data.index)}
+                      index={rowIndex}
+                    />
+                  )}
+                />
+                {columnHeaders.map(([key, label, width, extraProps]) => (
+                  <Column
+                    key={key}
+                    dataKey={key}
+                    label={label}
+                    width={width}
+                    {...extraProps}
+                    cellRenderer={renderCell}
+                  />
+                ))}
+              </Table>
+            )}
+          </WindowScroller>
+        </div>
+      ) : (
+        <Message info>
+          <Translate>There are no editables yet.</Translate>
+        </Message>
+      )}
+    </>
+  );
+}
--- a/indico/modules/events/editing/client/js/management/editable_type/EditableList.module.scss
+++ b/indico/modules/events/editing/client/js/management/editable_type/EditableList.module.scss
@@ -1,0 +1,85 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2020 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+@import 'base/palette';
+
+$border: 1px solid $pastel-gray;
+
+.editable-topbar {
+    display: flex;
+    width: 1000px;
+    justify-content: space-between;
+}
+
+.editable-list {
+    .table {
+        font-family: 'Muli', 'Helvetica Neue', Arial, Helvetica, sans-serif;
+        margin-top: 20px;
+        margin-bottom: 50px;
+    }
+
+    .rowcolumn-tooltip {
+        text-overflow: ellipsis;
+        white-space: nowrap;
+        overflow: hidden;
+        display: flex;
+    }
+
+    :global(.ReactVirtualized__Table__Grid) {
+        outline: none;
+    }
+
+    :global(.ReactVirtualized__Table__headerRow) {
+        font-weight: 700;
+        text-transform: uppercase;
+        display: flex;
+        border-top: $border;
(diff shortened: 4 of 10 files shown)


indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,138 @@
...
+            {sortedList[rowIndex][dataKey]}
+          </div>
+        </TooltipIfTruncated>
+      </>
+    );
+  }
+
+  return (
+    <>
+      <ManagementPageSubTitle title={`List of ${EditableTypeTitles[type]}s`} />
```
ThiefMaster, line 67: missing i18n (needs 3 separate strings since interpolation is not good in translatable strings)

indico/modules/events/editing/client/js/management/editable_type/EditableList.module.scss
```diff
@@ -0,0 +1,79 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2020 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+$border: 1px solid rgba(34, 36, 38, 0.15);
```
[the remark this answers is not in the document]
ParthS007: `$pastel-gray` looks nice. Added!

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,138 @@
...
+
+export default function EditableList() {
+  const eventId = useNumericParam('confId');
+  const {type} = useParams();
+  const [sortBy, setSortBy] = useState('id');
+  const [sortDirection, setSortDirection] = useState('ASC');
+  const [sortedList, setSortedList] = useState(list);
+
+  // eslint-disable-next-line no-shadow
+  function _sortList({sortBy, sortDirection}) {
```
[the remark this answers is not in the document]
ParthS007: Should I keep it or remove it?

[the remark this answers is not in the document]
ThiefMaster: i think for now we can keep it

indico/modules/events/editing/client/js/management/editable_type/EditableList.module.scss
```diff
@@ -0,0 +1,83 @@
...
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+@import 'base/palette';
+
+$border: 1px solid $pastel-gray;
+
+.editable-list {
+
```
mic4ael, line 13: redundant empty line

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,145 @@
...
+  const [sortedList, setSortedList] = useState(list);
+  const title = {
+    [EditableType.paper]: Translate.string('List of papers'),
+    [EditableType.slides]: Translate.string('List of slides'),
+    [EditableType.poster]: Translate.string('List of posters'),
+  }[type];
+
+  // eslint-disable-next-line no-shadow
+  function _sortList({sortBy, sortDirection}) {
+    const newList = _.sortBy(list, [sortBy]);
```
plourenco, line 44:
```suggestion
    const newList = _.orderBy(list, [sortBy], [sortDirection]);
```

Note: Seems like the lodash's `sortDirection` needs to either be `asc` or `desc` lowercase. Is that plausible changing? Not sure if we should convert it here or in the state.

ThiefMaster: Considering that this code is temporary (sorting will be server-side) I don't think we really need to care much ;)

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,145 @@
...
+              />
+              <Column
+                dataKey="id"
+                label="ID"
+                width={50}
+                cellRenderer={({dataKey, rowIndex}) => _cellRenderer(dataKey, rowIndex)}
+              />
+              <Column
+                dataKey="code"
+                label="Code"
```
mic4ael, line 115: Missing `i18n` (in other `Column`s as well

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,145 @@
...
+      <Button content={Translate.string('Download all files')} />
+      <Button color="blue" floated="right" content={Translate.string('Assign to myself')} />
+      <div styleName="editable-list">
+        <WindowScroller>
+          {({height}) => (
+            <Table
+              autoHeight
+              width={1000}
+              styleName="table"
+              height={height}
```
mic4ael, line 91: Is it needed if you also provide `autoHeight` prop?

ParthS007: Yes, `height` is a required `prop`.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,177 @@
...
+              <Column
+                disableSort
+                dataKey="checkbox"
+                width={30}
+                headerRenderer={() => (
+                  <Checkbox
+                    indeterminate={checked.length > 0 && checked.length < sortedList.length}
+                    checked={checked.length === sortedList.length}
+                    onChange={(e, data) => {
+                      data.checked
```
ParthS007, line 121: Should I create a new function for this or keep it here only?

plourenco: I would say move it to a new function. The render part becomes clearer and the onChange should no longer be referentially different (a new anonymous function) in every re-render.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,179 @@
...
+        <TooltipIfTruncated>
+          <div index={rowIndex} className="rowColumn_tooltip" role="gridcell">
+            {sortedList[rowIndex][dataKey]}
+          </div>
+        </TooltipIfTruncated>
+      </>
+    );
+  }
+
+  function _headerCheckboxState(data) {
```
plourenco, line 73: What do you think about naming these as `onChange{Header,Row}Checkbox`? 
My point is because having the `onChange` makes us clearly understand that this is a handler for a controlled checkbox change event (change -> set -> re-render).

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,179 @@
...
+            >
+              <Column
+                disableSort
+                dataKey="checkbox"
+                width={30}
+                headerRenderer={() => (
+                  <Checkbox
+                    indeterminate={checked.length > 0 && checked.length < sortedList.length}
+                    checked={checked.length === sortedList.length}
+                    onChange={(e, data) => {
```
plourenco, line 126: I think we can avoid the brackets here:
```suggestion
                    onChange={(e, data) => _headerCheckboxState(data)}
```

Or even better, and you just have to ignore the event in the callback function (that way we don't need an anonymous wrapper).
```suggestion
                    onChange={_headerCheckboxState(data)}
```

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ParthS007: Sounds good!
Will go with `onChange{Header,Row}Checkbox`. Let me know if anyone else has any other suggestions.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,179 @@
...
+            {sortedList[rowIndex][dataKey]}
+          </div>
+        </TooltipIfTruncated>
+      </>
+    );
+  }
+
+  function _headerCheckboxState(data) {
+    data.checked
+      ? setChecked([...Array(sortedList.length).keys()].map(row => sortedList[row].id))
```
panagiotappl, line 75: I think you could replace this with `? setChecked(sortedList.map(row => row.id))`

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,179 @@
...
+      : setChecked([]);
+  }
+
+  function _rowCheckboxState(data) {
```
panagiotappl, line 79: You could use a const instead of a function here and in all the other functions you've used. 
For example: `const _rowCheckboxState = data => { ... }`

indico/modules/events/editing/client/js/management/editable_type/EditableList.module.scss
```diff
@@ -0,0 +1,82 @@
...
+        margin-right: 10px;
+        align-items: center;
+        display: flex;
+    }
+
+    :global(#status.ReactVirtualized__Table__headerColumn) {
+        justify-content: center;
+    }
+
+    :global(.rowColumn_tooltip) {
```
panagiotappl, line 55: If it's a function that is not part of React Virtualized then it could be named the same way we use for other css classes. For example: `row-column-tooltip` or something like this.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,179 @@
...
+                  <Checkbox
+                    checked={checked.includes(sortedList[rowIndex].id)}
+                    onChange={(e, data) => {
+                      _rowCheckboxState(data);
+                    }}
+                    index={rowIndex}
+                  />
+                )}
+              />
+              <Column
```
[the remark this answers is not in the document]
ParthS007: `{style={justifyContent: 'space-evenly'}}`, I am not able to pass the `style` prop by this way.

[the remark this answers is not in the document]
panagiotappl: @ParthS007 There was a mistake in my comment, I fixed it now!

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ParthS007: I need to pass the `data` to the function.

plourenco: It doesn't work if you use `onChange={_headerCheckboxState}` and then receive it as: ?
```
const _headerCheckboxState = (__, data) => {
    console.log(data); // ...
}
```

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,174 @@
...
+          <Button
+            disabled={checked.length === 0}
+            color="blue"
+            content={Translate.string('Assign to myself')}
+          />
+          <Button content={Translate.string('Download all files')} />
+        </div>
+        <Search />
+      </div>
+      {sortedList.length > 0 ? (
```
panagiotappl, line 114: You can replace this with `{sortedList.length ? ( ... ) : ( ... ) }`

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,174 @@
...
+    setSortBy(sortBy);
+    setSortDirection(sortDirection);
+    setSortedList(_sortList({sortBy, sortDirection}));
+  };
+
+  const _cellRenderer = (dataKey, rowIndex) => {
+    return (
+      <>
+        {dataKey === 'status' && (
+          <StateIndicator index={rowIndex} state={sortedList[rowIndex][dataKey]} circular />
```
panagiotappl, line 73: I'm not sure whether the STATUS column renders properly at the moment. 
image

You need to provide one of [replaced, needs_submitter_confirmation, rejected, accepted ,assigned, needs_submitter_changes] for the `StateIndicator` component to display the correct color but if you do that then the text right after the indicator will appear in lowercase as well  (since you are providing the same variable there). 

If we want to render the staus text starting with a capital letter one solution would be to keep the titles of each state the same way we do with `EditableTypeTitles`. Another idea would be to add this as extra functionality in the `StateIndicator`. Maybe as some optional prop called `text`. What do the others think? @indico/development-team

indico/modules/events/editing/client/js/management/editable_type/EditableList.module.scss
```diff
@@ -0,0 +1,88 @@
...
+
+    :global(.ReactVirtualized__Table__headerColumn),
+    :global(.ReactVirtualized__Table__rowColumn) {
+        margin-right: 10px;
+        align-items: center;
+        display: flex;
+    }
+
+    :global(#status.ReactVirtualized__Table__headerColumn) {
+        justify-content: center;
```
panagiotappl, line 64: By adding this css you have the STATUS column spaced evenly in the center (even the header) but the rest of the columns are spaced on the left side. I would suggest to remove this all together and just add some spacing between the `StateIndicator` and the state text.

Before: 
image
After:
image

Also, that could be avoided if the status text was part of the `StateIndicator`

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ParthS007: @ThiefMaster @mic4ael @plourenco What do you guys think about it?

plourenco: I agree. I would go for having that extra inside `StateIndicator` since it is straightforward.
Not sure if there's a use case where a prop called `text` would show a different description from the state itself, that could become confusing.

indico/modules/events/editing/client/js/editing/timeline/StateIndicator.jsx
```diff
@@ -18,33 +18,41 @@ const colors = {
...
-    />
+    <>
+      <Popup
+        position="bottom center"
+        trigger={trigger}
+        content={tooltip}
+        on="hover"
+        disabled={!tooltip}
+      />
+      {label && <div style={{marginLeft: '5px', color: colors[state]}}>{labelText[state]}</div>}
```
panagiotappl, line 42: For longer text it looks like this 
image

One option would be to use `white-space: nowarp; text-overflow: ellipsis; overflow: hidden;` and it would look like this:
image

Maybe adding a tooltip would help in this case also. 

Other option would be to make the column wider maybe?

ParthS007: I will use this: `white-space: nowarp; text-overflow: ellipsis; overflow: hidden;`

Nice catch :sweat_smile:

panagiotappl: But then you need the tooltip for sure.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,171 @@
...
+              sortedList[rowIndex][dataKey]
+            )}
+          </div>
+        </TooltipIfTruncated>
+      </>
+    );
+  };
+
+  const _onChangeHeaderCheckbox = data => {
+    data.checked ? setChecked(sortedList.map(row => row.id)) : setChecked([]);
```
ThiefMaster, line 83: Please do not use the ternary operator as a replacement for normal if/else. Basically, don't use the ternary operator unless you do something with the return value of the expression.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,171 @@
...
+  };
+
+  const _onChangeRowCheckbox = data => {
+    const newRow = sortedList[data.index].id;
+    checked.includes(newRow)
+      ? setChecked(old => old.filter(row => row !== newRow))
+      : setChecked(old => [...old, newRow]);
```
ThiefMaster, line 90: same here

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,171 @@
...
+  };
+
+  return (
+    <>
+      <ManagementPageSubTitle title={title} />
+      <ManagementPageBackButton url={dashboardURL({confId: eventId})} />
+      <div styleName="editable-topbar">
+        <div>
+          <Button disabled={checked.length === 0} content={Translate.string('Assign')} />
```
ThiefMaster, line 99: what about a `const hasCheckedRows = ...` and using that in all the places where you check the length atm?

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ParthS007: oh okay, I will change it :+1:

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,183 @@
...
+    setSortBy(sortBy);
+    setSortDirection(sortDirection);
+    setSortedList(_sortList({sortBy, sortDirection}));
+  };
+
+  // eslint-disable-next-line react/prop-types
+  const _cellRenderer = ({dataKey, rowIndex}) => {
+    return (
+      <>
+        {dataKey === 'assigned' && <Icon name="user outline" />}
```
panagiotappl, line 69: I would suggest to move this inside `TooltipIfTruncated` and also inside `<div index={rowIndex} styleName="rowcolumn-tooltip" role="gridcell">` . That way the user icon is part of the div that contains both the icon and the user name information.

indico/modules/events/editing/client/js/management/editable_type/EditableList.module.scss
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.editable-list {
+    .table {
+        font-family: 'Muli', 'Helvetica Neue', Arial, Helvetica, sans-serif;
+        margin-top: 20px;
+        margin-bottom: 50px;
+    }
+
+    .rowcolumn-tooltip {
```
panagiotappl, line 25: You can add `display: flex` here for both items (in one case the dot and the status and in another case the user icon and the user name) are aligned in a row.

indico/modules/events/editing/client/js/editing/timeline/StateIndicator.module.scss
```diff
@@ -0,0 +1,12 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2020 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+.label-text {
+    margin-left: 5px;
```
panagiotappl, line 9: `position: relative;` and `bottom: 2px` are not needed here, as they don't help much when aligning the contents of `StateIndicator` but you need to also add here the three rules you have in the other class (`text-overflow: ellipsis;  white-space: nowrap; overflow: hidden;`). That way the ellipsis dots also appear in the same color as the text.

I'm not sure whether this is the correct way have proper text-wrapping, by having these three rules both in this and the parents class. Maybe someone else knows better.

ParthS007: I will do this right away. Thanks for checking this out.

indico/modules/events/editing/client/js/editing/timeline/StateIndicator.jsx
plourenco: I'm wondering if it would be possible to make one of the columns grow, perhaps using something such as `flexGrow={1}`. In the initial example, the title has an excessive width of `600px`, when we could reduce and shrink it (`flexShrink`) so the content fits without wrapping. Would this work?

ParthS007: When we will have actual data, the title will be long enough. I don't think then we need this.

indico/modules/events/editing/client/js/editing/timeline/StateIndicator.jsx
```diff
@@ -18,33 +21,54 @@ const colors = {
...
-      {text}
-    </Label>
-  );
+const labelText = {
+  replaced: 'Replaced',
+  needs_submitter_confirmation: 'Needs submitter confirmation',
+  rejected: 'Rejected',
+  accepted: 'Accepted',
+  assigned: 'Assigned',
+  needs_submitter_changes: 'Needs submitter changes',
```
ThiefMaster, line 30: Missing i18n; and I'd put this in models.js where we also have `EditableTypeTitles` etc

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,177 @@
...
+
+  // eslint-disable-next-line no-shadow
+  const _sort = ({sortBy, sortDirection}) => {
+    setSortBy(sortBy);
+    setSortDirection(sortDirection);
+    setSortedList(_sortList({sortBy, sortDirection}));
+  };
+
+  // eslint-disable-next-line react/prop-types
+  const _cellRenderer = ({dataKey, rowIndex}) => {
```
ThiefMaster, line 67: I think imperative names would be better here, i.e. `renderCell`

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,177 @@
...
+            <StateIndicator index={rowIndex} state={sortedList[rowIndex][dataKey]} circular label />
+          ) : (
+            sortedList[rowIndex][dataKey]
+          )}
+        </div>
+      </TooltipIfTruncated>
+    );
+  };
+
+  const _onChangeHeaderCheckbox = data => {
```
ThiefMaster, line 82: `toggleSelectAll`, and I'd pass only the `checked` bool to this function

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -0,0 +1,177 @@
...
+    if (data.checked) {
+      setChecked(sortedList.map(row => row.id));
+    } else {
+      setChecked([]);
+    }
+  };
+
+  const _onChangeRowCheckbox = data => {
```
ThiefMaster, line 90: `toggleSelectRow`, and I'd pass `index, checked` as separate arguments

PR #4425, opened 2020-04-23
Make PyIntEnum freeze enums in Alembic revisions

OmeGak:


indico/core/db/sqlalchemy/custom/int_enum.py
```diff
@@ -66,13 +66,22 @@ def coerce_set_value(self, value):
...
-        autogen_context.imports.add('from {} import {}'.format(self.enum.__module__, self.enum.__name__))
+        name = '_{}'.format(self.enum.__name__)
+        members = '\n'.join(['\t{} = {}'.format(k, v.value) for k, v in self.enum.__members__.iteritems()])
+        enum = ("""
+# ### code auto-generated by PyIntEnum - please adjust
+from enum import Enum
+class {name}(int, Enum):
+{members}
+# ### end of PyIntEnum code - please adjust
+        """.format(name=name, members=members))
```
ThiefMaster, line 77: You can use `textwrap.dedent()` to make this less ugly

indico/core/db/sqlalchemy/custom/int_enum.py
```diff
@@ -66,13 +66,22 @@ def coerce_set_value(self, value):
 
     def alembic_render_type(self, autogen_context):
         autogen_context.imports.add('from indico.core.db.sqlalchemy import PyIntEnum')
-        autogen_context.imports.add('from {} import {}'.format(self.enum.__module__, self.enum.__name__))
+        name = '_{}'.format(self.enum.__name__)
+        members = '\n'.join(['\t{} = {}'.format(k, v.value) for k, v in self.enum.__members__.iteritems()])
```
ThiefMaster, line 70: no tabs pls

ThiefMaster, line 70: i think you can just iterate `self.enum` directly

mic4ael:
Out of curiosity, what problem are you trying to solve here? In which case importing the Enum directly wouldn't suffice?

ThiefMaster:
Imagine changes like this:

- rev n: create enum column
- rev n+x: add new member to enum

Now you update from a version that was at rev n-1 to one that has at least n+x. When running rev n it will create the check constraint for the enum with all fields the enum has, even the ones that were added later.

Technically this is not a problem since it'll make the change in n+x just a no-op, but it means this revision is no longer idempotent, because if you upgrade and downgrade it, you will not have the same database structure you had before it (since the downgrade will remove the extra values from the check constraint)

mic4ael:
> Imagine changes like this:
> 
>     * rev n: create enum column
> 
>     * rev n+x: add new member to enum
> 
> 
> Now you update from a version that was at rev n-1 to one that has at least n+x. When running rev n it will create the check constraint for the enum with all fields the enum has, even the ones that were added later.
> 
> Technically this is not a problem since it'll make the change in n+x just a no-op, but it means this revision is no longer idempotent, because if you upgrade and downgrade it, you will not have the same database structure you had before it (since the downgrade will remove the extra values from the check constraint)

Thanks! Didn't think of such case.

indico/core/db/sqlalchemy/custom/int_enum.py
OmeGak: 🤮True! How could I!?

OmeGak:
> Technically this is not a problem since it'll make the change in n+x just a no-op, but it means this revision is no longer idempotent, because if you upgrade and downgrade it, you will not have the same database structure you had before it (since the downgrade will remove the extra values from the check constraint)

To complete the explanation, this is relevant in the context of the [idempotency checks that we implemented for the UN plugin](https://github.com/UNOG-Indico/indico-un/blob/master/bin/db_test.py#L217).

mic4ael:
> > Technically this is not a problem since it'll make the change in n+x just a no-op, but it means this revision is no longer idempotent, because if you upgrade and downgrade it, you will not have the same database structure you had before it (since the downgrade will remove the extra values from the check constraint)
> 
> To complete the explanation, this is relevant in the context of the [idempotency checks that we implemented for the UN plugin](https://github.com/UNOG-Indico/indico-un/blob/master/bin/db_test.py#L217).

Unfortunately I have no access to this repository :(

indico/core/db/sqlalchemy/custom/int_enum.py
```diff
@@ -65,14 +67,24 @@ def coerce_set_value(self, value):
         return self.enum(value)
 
     def alembic_render_type(self, autogen_context):
+        name = '_{}'.format(self.enum.__name__)
+        members = '\n'.join(['    {} = {}'.format(x.name, x.value) for x in self.enum])
+        enum_tpl = textwrap.dedent("""
+            # ### code auto-generated by PyIntEnum - please adjust
+            from enum import Enum
```
ThiefMaster, line 74: Can you move this into a separate `imports` entry? That way it doesn't get duplicated in case more than one such column is used in the same revision.

ThiefMaster: Damn, this breaks the file because now the imports is after the code... Anyway let's see if there is a nicer way (as mentioned on IRC)

indico/core/db/sqlalchemy/custom/int_enum.py
```diff
@@ -64,15 +66,22 @@ def coerce_set_value(self, value):
             return None
         return self.enum(value)
 
-    def alembic_render_type(self, autogen_context):
+    def alembic_render_type(self, autogen_context, toplevel_code):
+        name = '_{}'.format(self.enum.__name__)
+        members = '\n'.join(['    {} = {}'.format(x.name, x.value) for x in self.enum])
```
ThiefMaster, line 71: the `[]` isn't needed here; passing a generator to join() works fine

and I think `{!r}` would be slightly cleaner for the value, even though for an int it doesn't really matter whether you use the repr explicitly or not...

PR #4426, opened 2020-04-24
Allow microservices to configure editing workflow

### LINKED ISSUE #4422: Microservice - add setup APIs in Indico
Specifically:
 * Label setting API
 * Editable creation API

pferreir:
This closes #4422 

I have also updated @ThiefMaster's little service to reflect the changes (and do some setting up):         https://gist.github.com/pferreir/a1ca1c68b95ebc4dceb990ba441e3729

The service should be able to set up:
 * Tags
 * Active editables;
 * File types within each editable

Once the service is disconnected from the event, it should also do some tidying up, removing tags and file types which it created and are not used.

### DIFF
--- a/indico/modules/events/editing/controllers/base.py
+++ b/indico/modules/events/editing/controllers/base.py
@@ -7,30 +7,62 @@
 
 from __future__ import unicode_literals
 
-from flask import request
+from flask import request, session
+from werkzeug.exceptions import Unauthorized
 
 from indico.modules.events.contributions.controllers.display import RHContributionDisplayBase
 from indico.modules.events.controllers.base import RHDisplayEventBase
 from indico.modules.events.editing.models.editable import Editable, EditableType
+from indico.modules.events.editing.settings import editing_settings
 from indico.modules.events.management.controllers.base import RHManageEventBase
+from indico.modules.events.models.events import Event
 from indico.web.rh import RequireUserMixin
 
 
-class RHEditingBase(RequireUserMixin, RHDisplayEventBase):
+class TokenAccessMixin(object):
+    SERVICE_ALLOWED = False
+
+    def _token_can_access(self):
+        # we need to "fish" the event here because at this point _check_params
+        # hasn't run yet
+        event = Event.get_one(int(request.view_args['confId']), is_deleted=False)
+        if not self.SERVICE_ALLOWED or not request.bearer_token:
+            return False
+
+        event_token = editing_settings.get(event, 'service_token')
+        if request.bearer_token != event_token:
+            raise Unauthorized('Invalid bearer token')
+
+        return True
+
+    def _check_csrf(self):
+        # check CSRF if there is no bearer token or there's a session cookie
+        if session.user or not request.bearer_token:
+            super(TokenAccessMixin, self)._check_csrf()
+
+
+class RHEditingBase(TokenAccessMixin, RequireUserMixin, RHDisplayEventBase):
     """Base class for editing RHs that don't reference an editable."""
 
     EVENT_FEATURE = 'editing'
 
     def _check_access(self):
-        RHDisplayEventBase._check_access(self)
-        RequireUserMixin._check_access(self)
+        self.is_service_call = TokenAccessMixin._token_can_access(self)
+        if not self.is_service_call:
+            RHDisplayEventBase._check_access(self)
+            RequireUserMixin._check_access(self)
 
 
-class RHEditingManagementBase(RHManageEventBase):
+class RHEditingManagementBase(TokenAccessMixin, RHManageEventBase):
     """Base class for editing RHs that don't reference an editable."""
 
     EVENT_FEATURE = 'editing'
     PERMISSION = 'editing_manager'
+
+    def _check_access(self):
+        self.is_service_call = TokenAccessMixin._token_can_access(self)
+        if not self.is_service_call:
+            super(RHEditingManagementBase, self)._check_access()
 
 
 class RHEditableTypeManagementBase(RHEditingManagementBase):
--- a/indico/modules/events/editing/controllers/base.py
+++ b/indico/modules/events/editing/controllers/base.py
@@ -25,7 +25,7 @@ class TokenAccessMixin(object):
     def _token_can_access(self):
         # we need to "fish" the event here because at this point _check_params
         # hasn't run yet
-        event = Event.get_one(int(request.view_args['confId']), is_deleted=False)
+        event = Event.get_or_404(int(request.view_args['confId']), is_deleted=False)
         if not self.SERVICE_ALLOWED or not request.bearer_token:
             return False
 
--- a/indico/modules/events/editing/schemas.py
+++ b/indico/modules/events/editing/schemas.py
@@ -174,11 +174,12 @@ class EditingConfirmationAction(IndicoEnum):
 
 class EditableTagArgs(mm.Schema):
     class Meta:
-        rh_context = ('event', 'tag')
+        rh_context = ('event', 'tag', 'is_service_call')
 
     code = fields.String(required=True, validate=not_empty)
     title = fields.String(required=True, validate=not_empty)
     color = fields.String(required=True, validate=validate.OneOf(get_sui_colors()))
+    system = fields.Bool(missing=False)
 
     @validates('code')
     def _check_for_unique_tag_code(self, code):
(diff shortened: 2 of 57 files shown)


indico/modules/events/editing/controllers/base.py
```diff
@@ -7,30 +7,65 @@
...
+
+    def _token_can_access(self):
+        auth_header = request.headers.get('Authorization')
+        if not self.SERVICE_ALLOWED or not auth_header:
+            return False
+
+        m = AUTH_BEARER_RE.match(auth_header)
+        token = editing_settings.get(self.event, 'service_token')
+        if not m or m.group(1) != token:
+            raise Unauthorized("Invalid bearer token")
```
mic4ael, line 37:
```suggestion
            raise Unauthorized('Invalid bearer token')
```

indico/modules/events/editing/controllers/base.py
```diff
@@ -7,30 +7,65 @@
...
-class RHEditingManagementBase(RHManageEventBase):
+class RHEditingManagementBase(TokenAccessMixin, RHManageEventBase):
     """Base class for editing RHs that don't reference an editable."""
 
     EVENT_FEATURE = 'editing'
 
+    def _check_access(self):
+        self.is_service_call = TokenAccessMixin._token_can_access(self)
+        if not self.is_service_call:
+            super(RHEditingManagementBase, self)._check_access
```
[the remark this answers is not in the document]
pferreir: Whoops ;)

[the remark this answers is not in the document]
pferreir: I'm starting to hate Python.

indico/modules/events/editing/schemas.py
```diff
@@ -182,6 +192,11 @@ def _check_for_unique_tag_code(self, code):
         if query.has_rows():
             raise ValidationError(_('Tag code must be unique'))
 
+    @validates('system')
+    def _check_only_services_set_system_tags(self, value):
+        if value and not self.context['is_service_call']:
+            raise ValidationError(_('Only custom editing workflows can set system tags'))
```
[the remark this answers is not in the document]
pferreir: Yeah, I was going to do that, but within the same file there is another error which is actually translated. Maybe we should just not translate that one.

indico/modules/events/editing/controllers/base.py
```diff
@@ -7,30 +7,65 @@
...
 from indico.modules.events.contributions.controllers.display import RHContributionDisplayBase
 from indico.modules.events.controllers.base import RHDisplayEventBase
 from indico.modules.events.editing.models.editable import Editable, EditableType
+from indico.modules.events.editing.settings import editing_settings
 from indico.modules.events.management.controllers.base import RHManageEventBase
 from indico.web.rh import RequireUserMixin
 
 
-class RHEditingBase(RequireUserMixin, RHDisplayEventBase):
+AUTH_BEARER_RE = re.compile(r"Bearer ([\w-]+)")
```
mic4ael, line 23:
```suggestion
AUTH_BEARER_RE = re.compile(r'Bearer ([\w-]+)')
```

indico/modules/events/editing/controllers/base.py
```diff
@@ -7,30 +7,65 @@
...
+
+class TokenAccessMixin(object):
+    SERVICE_ALLOWED = False
+
+    def _token_can_access(self):
+        auth_header = request.headers.get('Authorization')
+        if not self.SERVICE_ALLOWED or not auth_header:
+            return False
+
+        m = AUTH_BEARER_RE.match(auth_header)
```
ThiefMaster, line 34: What about adding a `bearer_token` property to `IndicoRequest`, which returns either the token or None? Like this we avoid having to put this in any place where we want to use Bearer tokens (which I hope may be some API more endpoints in the future as well..).

In that case I'd allow anything in the token though, not just `[\w-]` - simply because IMO it should be up to the code checking the token if it wants to do any further checks beyond "does it match".

indico/modules/events/editing/schemas.py
```diff
@@ -42,10 +42,14 @@ class EditingFileTypeSchema(mm.ModelSchema):
     class Meta:
         model = EditingFileType
         fields = ('id', 'name', 'extensions', 'allow_multiple_files', 'required', 'publishable', 'is_used',
-                  'filename_template', 'is_used_in_condition')
+                  'filename_template', 'is_used_in_condition', 'url')
 
     is_used = fields.Function(lambda ft: EditingRevisionFile.query.with_parent(ft).has_rows())
     is_used_in_condition = fields.Function(lambda ft: EditingReviewCondition.query.with_parent(ft).has_rows())
+    url = fields.Function(lambda ftype: url_for('.api_edit_file_type',
```
ThiefMaster, line 49: I'd keep using `ft` here or change it in the two lines above as well for consistency

indico/modules/events/editing/service.py
```diff
@@ -78,10 +80,17 @@ def service_handle_enabled(event):
         'title': event.title,
         'url': event.external_url,
         'token': editing_settings.get(event, 'service_token'),
+        'config_endpoints': {
+            'tags': url_for('.api_create_tag', event, _external=True),
```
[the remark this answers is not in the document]
pferreir: But they're all using the same URL, only different verbs...

indico/modules/events/editing/controllers/base.py
```diff
@@ -7,30 +7,65 @@
...
+        token = editing_settings.get(self.event, 'service_token')
+        if not m or m.group(1) != token:
+            raise Unauthorized("Invalid bearer token")
+
+        return True
+
+    def _check_csrf(self):
+        # only check CSRF if a service call is not allowed
+        if not self.SERVICE_ALLOWED:
```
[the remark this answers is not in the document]
pferreir: Indeed. Good catch, thanks!

indico/modules/events/editing/controllers/base.py
pferreir: Can't wait to start using Black.

ThiefMaster: unfortunately black cannot enforce single quotes :/

pferreir: No way!

pferreir:
Thanks @mic4ael and @ThiefMaster for the comments! I believed I've fixed everything, let me know if you find anything else!

indico/modules/events/editing/service.py
```diff
@@ -78,10 +81,39 @@ def service_handle_enabled(event):
...
+                    } for tag in event.editing_tags
+                }
+            },
+            'editable_types': url_for('.api_enabled_editable_types', event, _external=True),
+            'file_types': {
+                t.name: {
+                    'create': url_for('.api_add_file_type', event, type=t.name, _external=True),
+                    'list': url_for('.api_file_types', event, type=t.name, _external=True),
+                    'file_type_actions': {
+                        ft.name: {
```
ThiefMaster, line 101: Not sure if using the name as a key is a good idea here, since this cannot be used as a stable identifier (it's the user-facing name which can be changed at any time).

Actually, I'm wondering whether sending those `<tag|file_type>_actions` here is a good idea to begin with. The service would not only have to store this data (should we add an endpoint the service can use to retrieve this kind of config data again?) but also lack any information about tags/filetype created after enabling the service.

Maybe it'd be better to just send the URLs that never change (ie create and list), and include the *update* and *delete* URLs when retrieving the actual tags/filetypes via the *list* url? Looks like you already added the `url` to the schemas anyway

indico/modules/events/editing/controllers/base.py
```diff
@@ -7,31 +7,63 @@
...
+from indico.modules.events.models.events import Event
 from indico.web.rh import RequireUserMixin
 
 
-class RHEditingBase(RequireUserMixin, RHDisplayEventBase):
+class TokenAccessMixin(object):
+    SERVICE_ALLOWED = False
+
+    def _token_can_access(self):
+        # we need to "fish"  the event here because at this point _check_params
```
ThiefMaster, line 26: extra space after `"fish"` 👾

indico/modules/events/editing/controllers/base.py
```diff
@@ -7,31 +7,63 @@
...
+        # hasn't run yet
+        event = Event.get(int(request.view_args['confId']))
```
ThiefMaster, line 28: `, is_deleted=False`, and I think right now we may also get a rather ugly error when accessing an URL using this mixin with an invalid event id that doesn't exist, `.get_one(...)` should avoid this.

indico/modules/events/editing/service.py
pferreir: I think that makes sense. I believe that was actually my original approach.

PR #4430, opened 2020-04-29
Show accepted editable files

### LINKED ISSUE #4415: Display files from published editables on contribution page
This happens when the Editable has a `published_revision`. All files from this revision that belong to a publishable file type should then be shown nicely on the contribution page.

You can probably have a look at how we show accepted papers, but don't take this as a requirement - there may be a nicer way.

It's probably a good idea to take into account whether a given publishable file type only has (or allows?) one file or multiple - in case of a single file we could show it more prominently compared to a list of files.

panagiotappl:
Section title is temporary and should be changed to something more appropriate than "Proceedings". Any ideas are welcome. 
image

closes #4415

### DIFF
--- a/indico/modules/events/contributions/models/contributions.py
+++ b/indico/modules/events/contributions/models/contributions.py
@@ -24,6 +24,7 @@ from indico.core.db.sqlalchemy.protection import ProtectionManagersMixin
 from indico.core.db.sqlalchemy.util.models import auto_table_args
 from indico.core.db.sqlalchemy.util.queries import increment_and_get
 from indico.core.db.sqlalchemy.util.session import no_autoflush
+from indico.modules.events.editing.models.editable import EditableType
 from indico.modules.events.management.util import get_non_inheriting_objects
 from indico.modules.events.models.persons import AuthorsSpeakersMixin, PersonLinkDataMixin
 from indico.modules.events.papers.models.papers import Paper
@@ -510,7 +511,9 @@ class Contribution(DescriptionMixin, ProtectionManagersMixin, LocationMixin, Att
             return []
 
         enabled_editable_types = editing_settings.get(self.event, 'editable_types')
-        return [editable for editable in self.editables if editable.type.name in enabled_editable_types]
+        enabled_editables = [editable for editable in self.editables if editable.type.name in enabled_editable_types]
+        order = list(EditableType)
+        return sorted(enabled_editables, key=lambda editable: order.index(editable.type))
 
     @property
     def has_published_editables(self):
--- a/indico/modules/events/contributions/models/contributions.py
+++ b/indico/modules/events/contributions/models/contributions.py
@@ -512,6 +512,10 @@ class Contribution(DescriptionMixin, ProtectionManagersMixin, LocationMixin, Att
         enabled_editable_types = editing_settings.get(self.event, 'editable_types')
         return [editable for editable in self.editables if editable.type.name in enabled_editable_types]
 
+    @property
+    def has_published_editables(self):
+        return any(e.published_revision_id is not None for e in self.enabled_editables)
+
     def is_paper_reviewer(self, user):
         return user in self.paper_content_reviewers or user in self.paper_layout_reviewers
 
--- a/indico/modules/events/editing/models/revisions.py
+++ b/indico/modules/events/editing/models/revisions.py
@@ -165,3 +165,10 @@ class EditingRevision(RenderModeMixin, db.Model):
     @locator_property
     def locator(self):
         return dict(self.editable.locator, revision_id=self.id)
+
+    def get_spotlight_file(self):
+        files = [file for file in self.files if file.file_type.publishable]
+        return files[0] if len(files) == 1 else None
+
+    def get_files_based_on_file_type(self, file_type):
+        return [file for file in self.files if file.file_type == file_type]
--- a/indico/modules/events/editing/models/revisions.py
+++ b/indico/modules/events/editing/models/revisions.py
@@ -8,6 +8,7 @@
 from __future__ import unicode_literals
 
 import re
+from collections import defaultdict
 
 from indico.core.db import db
 from indico.core.db.sqlalchemy import PyIntEnum, UTCDateTime
@@ -170,5 +171,10 @@ class EditingRevision(RenderModeMixin, db.Model):
         files = [file for file in self.files if file.file_type.publishable]
         return files[0] if len(files) == 1 else None
 
-    def get_files_based_on_file_type(self, file_type):
-        return [file for file in self.files if file.file_type == file_type]
+    def get_published_files(self):
+        """Get the published files, grouped by file type."""
+        files = defaultdict(list)
+        for file in self.files:
+            if file.file_type.publishable:
+                files[file.file_type].append(file)
+        return dict(files)
--- a/indico/modules/events/papers/templates/_contributions.html
+++ b/indico/modules/events/papers/templates/_contributions.html
@@ -46,7 +46,7 @@
         <section>
             <div class="header">
                 <div class="header-row">
-                <h2>{% trans %}Editing{% endtrans %}</h2>
+                    <h2>{% trans %}Editing{% endtrans %}</h2>
                 </div>
             </div>
             {% if has_enabled_types %}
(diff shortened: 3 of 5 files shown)


indico/modules/events/editing/models/revisions.py
```diff
@@ -165,3 +165,10 @@ def __repr__(self):
     @locator_property
     def locator(self):
         return dict(self.editable.locator, revision_id=self.id)
+
+    def get_spotlight_file(self):
+        files = [file for file in self.files if file.file_type.publishable]
+        return files[0] if len(files) == 1 else None
```
plourenco, line 171:
```suggestion
        return files[0] if len(files) else None
```

You can leave this as is, up to your criteria ;) but here's a suggestion I saw the other day for a simple and pretty cool idiom to return the first element:
```suggestion
        return next(iter(files), None)
```

indico/modules/events/contributions/models/contributions.py
```diff
@@ -510,7 +511,19 @@ def enabled_editables(self):
             return []
 
         enabled_editable_types = editing_settings.get(self.event, 'editable_types')
-        return [editable for editable in self.editables if editable.type.name in enabled_editable_types]
+        enabled_editables = [editable for editable in self.editables if editable.type.name in enabled_editable_types]
+        order = list(EditableType)
+        return sorted(enabled_editables, key=lambda editable: order.index(editable.type))
```
pferreir, line 516: This is a quite interesting way of doing the sorting, but I would probably find something like this more readable:

```python
available_editable_types = {editable.type.name for editable in self.editables}
return [t for t in EditableType if t.name in (available_editable_types & enabled_editable_types)]
```
(untested, though)

indico/modules/events/editing/models/revisions.py
pferreir: @plourenco yeah, that's quite nice looking, but it has two drawbacks IMHO:
 * it creates a whole iterator for what is a single operation;
 * it makes the whole thing less readable

ThiefMaster: `if len(files)` doesn't look like a good idea to me in any case - you'd use `if files` there. But I think this check here is especially about checking that there is exactly one file, not that there are any number of files.

indico/modules/events/contributions/models/contributions.py
ThiefMaster: but this returns the actual editables, no the types

indico/modules/events/contributions/models/contributions.py
```diff
@@ -510,7 +511,19 @@ def enabled_editables(self):
...
+
+    @property
+    def has_accepted_editables(self):
+        for editable in self.enabled_editables:
+            if editable.revisions:
+                latest_revision = editable.revisions[-1]
+                latest_state = latest_revision.final_state or latest_revision.initial_state
+                if latest_state.name == 'accepted':
+                    return True
+        return False
```
ThiefMaster, line 526: return any(e.published_revision_id is not None for e in self.enabled_editables)

The state doesn't matter (even though it should always be accepted) - what matters is that the revision is marked as published. With the microservice it is actually possible that a revision is accepted but not published yet ("awaiting QA")

indico/modules/events/contributions/models/contributions.py
```diff
@@ -510,7 +511,19 @@ def enabled_editables(self):
             return []
 
         enabled_editable_types = editing_settings.get(self.event, 'editable_types')
-        return [editable for editable in self.editables if editable.type.name in enabled_editable_types]
+        enabled_editables = [editable for editable in self.editables if editable.type.name in enabled_editable_types]
+        order = list(EditableType)
+        return sorted(enabled_editables, key=lambda editable: order.index(editable.type))
+
+    @property
+    def has_accepted_editables(self):
```
ThiefMaster, line 519: Should probably be renamed to `has_published_editables`

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -166,3 +166,71 @@ <h2>{% trans %}Editing{% endtrans %}</h2>
...
+
+{% macro render_proceedings_section(contribution) %}
+    {% set enabled_editables = contribution.enabled_editables %}
+    <section>
+        <div class="header">
+            <div class="header-row">
+                <h2>{% trans %}Proceedings{% endtrans %}</h2>
+            </div>
+        </div>
+        {% for editable in enabled_editables -%}
```
ThiefMaster, line 203: {% for editable in enabled_editables if editable.published_revision -%}

and use `editable.published_revision` below instead of `latest_revision`; the state check can go away as well

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -166,3 +166,71 @@ <h2>{% trans %}Editing{% endtrans %}</h2>
...
+                        </div>
+                    </div>
+                    {% set spotlight_file = latest_revision.get_spotlight_file() %}
+                    {% if spotlight_file %}
+                        {% set icon = icon_from_mimetype(spotlight_file.file.content_type, 'icon-file-filled') %}
+                        <div class="paper-content flexcol">
+                            <div class="flexrow f-a-center spotlight-file i-button text-color borderless paper-file"
+                                title="{{ spotlight_file.filename }}">
+                                <a href="{{ spotlight_file.download_url }}"
+                                class="attachment spotlight-file-name flexrow f-a-center truncate-text {{ icon }}">
```
mic4ael, line 220: misaligned `class` attribute

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -166,3 +166,71 @@ <h2>{% trans %}Editing{% endtrans %}</h2>
...
+                        <div class="header-row">
+                            <h3 class="icon-file">{{ editable.type.title }}</h3>
+                        </div>
+                    </div>
+                    {% set spotlight_file = latest_revision.get_spotlight_file() %}
+                    {% if spotlight_file %}
+                        {% set icon = icon_from_mimetype(spotlight_file.file.content_type, 'icon-file-filled') %}
+                        <div class="paper-content flexcol">
+                            <div class="flexrow f-a-center spotlight-file i-button text-color borderless paper-file"
+                                title="{{ spotlight_file.filename }}">
```
mic4ael, line 218: misaligned `title` attribute

indico/modules/events/editing/models/revisions.py
panagiotappl: Yes, we want to check if there's exactly one file, which is this case is our spotlight file.

indico/modules/events/contributions/models/contributions.py
pferreir: > but this returns the actual editables, not the types

Indeed. Never mind 🤦

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -166,3 +166,68 @@ <h2>{% trans %}Editing{% endtrans %}</h2>
...
+                               class="attachment spotlight-file-name flexrow f-a-center truncate-text {{ icon }}">
+                                <span class="title truncate-text">{{ spotlight_file.file.filename }}</span>
+                            </a>
+                        </div>
+                    </div>
+                {% else %}
+                    {% set file_types = contribution.event.editing_file_types %}
+                    {% for file_type in file_types if file_type.publishable and file_type.type == editable.type %}
+                        {% set published_revision_files = published_revision.get_files_based_on_file_type(file_type) %}
+                        {{ _render_file_type_section(oublished_revision_files, file_type) }}
```
ThiefMaster, line 227: typo...

```suggestion
                        {{ _render_file_type_section(published_revision_files, file_type) }}
```

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -166,3 +166,68 @@ <h2>{% trans %}Editing{% endtrans %}</h2>
         </div>
     {% endif %}
 {% endmacro %}
+
+{% macro _render_file_type_section(latest_revision_files, file_type) %}
```
ThiefMaster, line 170: `latest_revision_files` -> `revision_files` or `published_files`?

indico/modules/events/papers/templates/_contributions.html
```diff
@@ -166,3 +166,68 @@ <h2>{% trans %}Editing{% endtrans %}</h2>
...
+                    </tr>
+                {% endfor %}
+            </table>
+        </div>
+    </div>
+{% endmacro %}
+
+
+{% macro render_proceedings_section(contribution) %}
+    {% set enabled_editables = contribution.enabled_editables %}
```
ThiefMaster, line 196: you only use this once; i don't think there's much benefit in this assignment.

BTW: using methods instead of properties for things that trigger queries on each call might be better - that way it's clearer that it's not something you should just access multiple times instead of assigning it in case you do need to access it multiple times

panagiotappl: This particular one gets the information from the event settings. Does that count as triggering a query?

ThiefMaster: no. i somehow thought we were explicitly querying something in there...

indico/modules/events/editing/models/revisions.py
```diff
@@ -165,3 +165,10 @@ def __repr__(self):
...
+
+    def get_files_based_on_file_type(self, file_type):
```
ThiefMaster, line 173: What about `get_files_for_file_type` instead?

Issue #936, opened 2011-10-24
Avatar: add pic


jbenito3:
``` yaml
# Ticket imported from Trac
```

This could help to identify the participants of a conference. 
Original request:

```
Juste au passage, je voulais te faire une suggestion pour la V3 ;-) :
Je fais partie du comité de programme d'une conférence importante en France, les JRES (www.jres.org). Le comité d'organisation a développé quasiment en temps réel l'appli de support pour le workflow des abstracts/articles/présentations (et tout le reste, inscriptions, programme etc). Pour l'instant, ils ne prévoient pas d'en faire un produit de gestion de conférence générique, mais il y a quelques idées que je trouve sympathiques. À l'occasion, je vous ferai une démo (partielle, je ne suis pas admin du système). Une des idées sympathiques : tous les gens qui ont un compte (donc en fait tout le monde) peut uploader une photo dans son profil, et chaque fois qu'on survole le nom de quelqu'un, on peut faire apparaitre sa photo. C'est extrêmement précieux pour les futurs chairpersons (j'en suis un !) pour retrouver le moment venu ses speakers par exemple. Je suis sûr que Marco aurait apprécié pour ICHEP. Je m'aperçois même à l'instant que c'est vrai aussi pour le programme (mais j'ai l'impression qu'il faut être loggué pour voir les photos). Du coup je trouve que c'est aussi un vrai plus pour les participants.

Amicalement,

Christian.
```

PR #4431, opened 2020-04-29
Add user profile picture

OmeGak:
Partially addresses #936.

This PR adds:
- A file upload field in the user personal details form.
- Display of the profile picture in the user profile dashboard.

## Functional testing instructions

Upload profile pic:
1. Go to user profile settings (`user/<user_id>/profile/`)
2. Upload a file from your computer
3. Verify that the picture is uploaded successfully
4. Verify that the profile picture is displayed in the user profile dashboard (`user/<user_id>/dashboard/`)

Delete profile pic:
1. Go to user profile settings (`user/<user_id>/profile/`)
2. Delete the profile picture
3. Verify that the profile picture is not displayed in the user profile dashboard (`user/<user_id>/dashboard/`)

## Screenshots

image

image

image

### DIFF
--- a/indico/migrations/versions/20200428_1807_532f0ea25bb1_add_user_profile_picture.py
+++ b/indico/migrations/versions/20200428_1807_532f0ea25bb1_add_user_profile_picture.py
@@ -1,0 +1,31 @@
+"""Add user profile picture
+
+Revision ID: 532f0ea25bb1
+Revises: 7f56252c73ab
+Create Date: 2020-04-28 18:07:51.879932
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+
+# revision identifiers, used by Alembic.
+revision = '532f0ea25bb1'
+down_revision = '7f56252c73ab'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('users', sa.Column('picture', sa.LargeBinary(), nullable=True), schema='users')
+    op.add_column('users', sa.Column('picture_metadata', postgresql.JSONB(), nullable=False, server_default='null'),
+                  schema='users')
+    op.alter_column('users', 'picture_metadata', server_default=None, schema='users')
+    op.create_check_constraint('valid_picture', 'users', "(picture IS NULL) = (picture_metadata::text = 'null')",
+                               schema='users')
+
+
+def downgrade():
+    op.drop_column('users', 'picture_metadata', schema='users')
+    op.drop_column('users', 'picture', schema='users')
--- a/indico/modules/users/controllers.py
+++ b/indico/modules/users/controllers.py
@@ -7,6 +7,7 @@
 
 from __future__ import unicode_literals
 
+import os
 from collections import namedtuple
 from io import BytesIO
 from operator import attrgetter, itemgetter
@@ -15,6 +16,7 @@ from dateutil.relativedelta import relativedelta
 from flask import flash, jsonify, redirect, request, session
 from markupsafe import Markup, escape
 from marshmallow import fields
+from PIL import Image
 from sqlalchemy.orm import joinedload, load_only, subqueryload, undefer
 from sqlalchemy.orm.exc import StaleDataError
 from webargs import validate
@@ -24,6 +26,7 @@ from indico.core import signals
 from indico.core.auth import multipass
 from indico.core.db import db
 from indico.core.db.sqlalchemy.util.queries import get_n_matching
+from indico.core.errors import UserValueError
 from indico.core.marshmallow import mm
 from indico.core.notifications import make_email, send_email
 from indico.legacy.common.cache import GenericCache
@@ -36,18 +39,21 @@ from indico.modules.events import Event
 from indico.modules.events.util import serialize_event_for_ical
 from indico.modules.users import User, logger, user_management_settings
 from indico.modules.users.forms import (AdminAccountRegistrationForm, AdminsForm, AdminUserSettingsForm, MergeForm,
-                                        SearchForm, UserDetailsForm, UserEmailsForm, UserPreferencesForm)
+                                        SearchForm, UserDetailsForm, UserEmailsForm, UserPictureForm,
+                                        UserPreferencesForm)
 from indico.modules.users.models.emails import UserEmail
 from indico.modules.users.operations import create_user
-from indico.modules.users.util import (get_linked_events, get_related_categories, get_suggested_categories, merge_users,
-                                       search_users, serialize_user)
+from indico.modules.users.util import (get_linked_events, get_picture_data, get_related_categories,
+                                       get_suggested_categories, merge_users, search_users, serialize_user)
 from indico.modules.users.views import WPUser, WPUserDashboard, WPUsersAdmin
 from indico.util.date_time import now_utc
 from indico.util.event import truncate_path
+from indico.util.fs import secure_filename
 from indico.util.i18n import _
+from indico.util.images import square
 from indico.util.marshmallow import HumanizedDate, validate_with_message
 from indico.util.signals import values_from_signal
-from indico.util.string import make_unique_token
+from indico.util.string import crc32, make_unique_token
 from indico.web.args import use_kwargs
 from indico.web.flask.templating import get_template_module
 from indico.web.flask.util import send_file, url_for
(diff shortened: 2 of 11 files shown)


indico/modules/users/controllers.py
```diff
@@ -177,13 +182,63 @@ class RHPersonalData(RHUserBase):
...
+        return WPUser.render_template('personal_data.html', 'personal_data', user=self.user, form=form,
+                                      pic_form=pic_form)
+
+
+class RHProfilePictureDisplay(RHUserBase):
+    def _process(self):
+        if not self.user.has_picture:
+            raise NotFound
+        metadata = self.user.picture_metadata
+        return send_file(metadata['filename'], BytesIO(self.user.picture), mimetype=metadata['content_type'],
```
ThiefMaster, line 204: I don't think we use the original filename here. Imagine a user using a somewhat inappropriate name...

FWIW I think it isn't a good idea to send the original filename for the other images where we do this either (even though it was probably me writing some of those RHs 🙈) - storing them is one thing (and that's nice so the user sees their original filename), but for sending I'd use something generic, like "user.png"

indico/modules/users/controllers.py
```diff
@@ -177,13 +182,63 @@ class RHPersonalData(RHUserBase):
...
+                         conditional=True)
+
+
+class RHProfilePicture(RHUserBase):
+    def _process(self):
+        f = request.files['picture']
+        try:
+            pic = Image.open(f)
+        except IOError:
+            raise UserValueError(_('You cannot upload this file as a room picture.'))
```
ThiefMaster, line 214: room picture?

indico/modules/users/controllers.py
```diff
@@ -177,13 +182,63 @@ class RHPersonalData(RHUserBase):
...
+        if pic.format.lower() not in {'jpeg', 'png'}:
```
ThiefMaster, line 215: what about allowing gif as well?

indico/modules/users/models/users.py
```diff
@@ -252,6 +254,17 @@ class User(PersonMixin, db.Model):
         nullable=False,
         default=lambda: unicode(uuid4())
     )
+    #: the user profile picture
+    picture = db.deferred(db.Column(
+        db.LargeBinary,
+        nullable=True
+    ))
+    # user profile picture metadata
```
ThiefMaster, line 262: missing `:` (to make the comment docstring-ish)

indico/modules/users/templates/personal_data.html
```diff
@@ -48,4 +63,29 @@
             </div>
         </div>
     </div>
+
+    <script>
+        $('.pic-form').on('indico:fieldsSaved', function(e, response) {
+            var hidden = !response.content;
```
ThiefMaster, line 69: Let's use `const` (or `let` if necessary) and arrow functions for callbacks (at least when `this` is not needed, so not in this callback) in new code

indico/modules/users/controllers.py
```diff
@@ -177,13 +182,63 @@ class RHPersonalData(RHUserBase):
...
+        content = image_bytes.read()
+        self.user.picture = content
+        self.user.picture_metadata = {
+            'hash': crc32(content),
+            'size': len(content),
+            'filename': os.path.splitext(secure_filename(f.filename, 'logo'))[0] + '.png',
+            'content_type': 'image/png'
+        }
+        flash(_('Profile picture uploaded'), 'success')
+        logger.info("Profile picture of user %s uploaded by %s", self.user, session.user)
```
mic4ael, line 231:
```suggestion
        logger.info('Profile picture of user %s uploaded by %s', self.user, session.user)
```

indico/modules/users/controllers.py
```diff
@@ -177,13 +182,63 @@ class RHPersonalData(RHUserBase):
...
+        return jsonify_data(content=get_picture_data(self.user))
+
+
+class RHProfilePictureDelete(RHUserBase):
+    def _process(self):
+        self.user.picture = None
+        self.user.picture_metadata = None
+        flash(_('Profile picture deleted'), 'success')
+        logger.info("Profile picture of user %s deleted by %s", self.user, session.user)
```
mic4ael, line 240:
```suggestion
        logger.info('Profile picture of user %s deleted by %s', self.user, session.user)
```

indico/web/client/styles/modules/_dashboard.scss
```diff
@@ -193,10 +193,19 @@
...
+
+            &.placeholder {
+                font-size: 70px;
+            }
+
+            &:not(.placeholder) {
+                height: 140px;
+            }
         }
+
```
mic4ael, line 207: redundant empty lines?

indico/modules/users/controllers.py
OmeGak: Copypasta leftovers.

indico/modules/users/controllers.py
OmeGak: Would you really want people to use animated gifs as profiles pictures?

OmeGak: aAYrp7R

ThiefMaster: no but that part goes away when converting to PNG anyway

indico/migrations/versions/20200428_1807_532f0ea25bb1_add_user_profile_picture.py
```diff
@@ -0,0 +1,33 @@
...
+    op.add_column('users', sa.Column('picture_metadata', postgresql.JSONB(astext_type=sa.Text()), nullable=False,
+                                     server_default='null'),
+                  schema='users')
+    op.alter_column('users', 'picture_metadata', server_default=None, schema='users')
+    op.create_check_constraint('valid_picture', 'users', "(picture IS NULL) = (picture_metadata::text = 'null')",
+                               schema='users')
+
+
+def downgrade():
+    op.drop_constraint('ck_users_valid_picture', 'users', schema='users')
```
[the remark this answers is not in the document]
OmeGak: That's correct.

indico/modules/users/controllers.py
OmeGak: Agreed!

ThiefMaster:
When testing with my [octocat avatar](https://user-images.githubusercontent.com/179599/81169165-ac7e9800-8f98-11ea-8eb8-06c2e987547b.png) it somehow got [completely destroyed](https://user-images.githubusercontent.com/179599/81169217-c6b87600-8f98-11ea-97b5-527748648c73.png) during the conversion. Not sure why this happened though... 

It also looks a bit strange right now if the picture is smaller than the circle where it's displayed.

ThiefMaster:
Regarding the upload UI, would it make sense to have a separate box for it on the page instead of including it in the "Details" box?

OmeGak:
> When testing with my [octocat avatar](https://user-images.githubusercontent.com/179599/81169165-ac7e9800-8f98-11ea-8eb8-06c2e987547b.png) it somehow got [completely destroyed](https://user-images.githubusercontent.com/179599/81169217-c6b87600-8f98-11ea-97b5-527748648c73.png) during the conversion. Not sure why this happened though...

I'll look into this. Probably something that I need to pass to `PIL` to preserve the transparency.

> Regarding the upload UI, would it make sense to have a separate box for it on the page instead of including it in the "Details" box?

I've tried a few variations, but I'm not crazy about any of them.

||label|no label|
|-|-|-|
|top|Screen Shot 2020-05-06 at 13 31 44|Screen Shot 2020-05-06 at 13 31 33|
|bottom|Screen Shot 2020-05-06 at 13 35 19|Screen Shot 2020-05-06 at 13 35 12|

ThiefMaster:
True. Doesn't look super amazing. Any other ideas here @indico/development-team?

panagiotappl:
I personally like both the top label and bottom label ideas. IMO, they look good. I would probably expect to see that on top rather than the bottom part of the page.

OmeGak:
> I personally like both the top label and bottom label ideas. IMO, they look good. I would probably expect to see that on top rather than the bottom part of the page.

Two concerns:
- What would you do with the `Block User` button? I would not leave it on the bottom box, but putting it on the `Profile Picture` box looks semantically wrong.
- The label looks redundant, as the box has the same title. We could change the label or box title to something else, but at that point we'll be contorting the UI just for the sake of having two boxes.

Personally, I would just leave it as it is, with a single box.

panagiotappl:
Maybe it's better to leave it in the same box after all. I don't think it looks bad if it's in the same box anyway. I don't see why the picture cannot be in basic details.

panagiotappl:
I'm sorry. I accidentally clicked "Close and comment" . :(

OmeGak:
@ThiefMaster, upload of transparent images is now working fine. I was removing the alpha channel. This fixes it:

```diff
- if pic.mode != 'RGB':
+ if pic.mode not in {'RGB', 'RGBA'}:
     pic = pic.convert('RGB')
```

plourenco:
I like the top options, I'm not sure why but it seems straightforward to have a picture as one of the first details to be edited.

Apart from that, perhaps I could suggest having the image squared (side-by-side) with the details inputs? The same way we're now using in #4408.

indico/modules/users/controllers.py
```diff
@@ -177,13 +183,63 @@ class RHPersonalData(RHUserBase):
...
+        try:
+            pic = Image.open(f)
+        except IOError:
+            raise UserValueError(_('You cannot upload this file as profile picture.'))
+        if pic.format.lower() not in {'jpeg', 'png', 'gif'}:
+            raise UserValueError(_('The file has an invalid format ({format}).').format(format=pic.format))
+        if pic.mode not in {'RGB', 'RGBA'}:
+            pic = pic.convert('RGB')
+        image_bytes = BytesIO()
+        pic = square(pic)
```
ThiefMaster, line 220: `make_square`?

ThiefMaster, line 220: I think we also need to resize the image to a max size (256x256px or something like that).

I just managed to upload an 8mb photo of a very pretty cat, but as much as I like hi-res cat pictures, a user picture/avatar is not the place for them ;)

OmeGak: `square` is a also a verb.

PR #4433, opened 2020-05-01
Fix item sorting in badge designer

openprojects:
Fix sorting items in "Insert element" combo: additional custom items (not "fields") are not correctly sorted.

### DIFF
--- a/indico/modules/designer/templates/template.html
+++ b/indico/modules/designer/templates/template.html
@@ -301,7 +301,7 @@
                         {% for group_id, group in placeholders.viewitems() | sort(attribute='1.title') %}
                             {% if group_id not in config.disallow_groups %}
                                 <optgroup label="{{ group.title }}">
-                                    {% for option_id, option in group.options|dictsort(by='value') %}
+                                    {% for option_id, option in group.options|dictsort(by='key') %}
                                         <option value="{{ option_id }}">{{ option }}</option>
                                     {% endfor %}
                                 </optgroup>
--- a/indico/modules/designer/templates/template.html
+++ b/indico/modules/designer/templates/template.html
@@ -301,7 +301,7 @@
                         {% for group_id, group in placeholders.viewitems() | sort(attribute='1.title') %}
                             {% if group_id not in config.disallow_groups %}
                                 <optgroup label="{{ group.title }}">
-                                    {% for option_id, option in group.options|dictsort(by='key') %}
+                                    {% for option_id, option in group.options|dictsort(by='value') %}
                                         <option value="{{ option_id }}">{{ option }}</option>
                                     {% endfor %}
                                 </optgroup>
--- a/indico/modules/designer/util.py
+++ b/indico/modules/designer/util.py
@@ -28,7 +28,7 @@ def get_placeholder_options():
 def get_nested_placeholder_options():
     groups = {group_id: {'title': group_title, 'options': {}} for group_id, group_title in GROUP_TITLES.viewitems()}
     for name, placeholder in get_placeholder_options().viewitems():
-        groups[placeholder.group]['options'][name] = placeholder.description
+        groups[placeholder.group]['options'][name] = unicode(placeholder.description)
     return groups
 
 


ThiefMaster:
I'm not sure I understand the problem you are fixing. The sorting here is for the user-facing string, so sorting by the internal name doesn't looks like the right thing to do, since it's pure coincidence that the internal name is usually a snake_cased version of the user-facing title. And it certainly is not likely to be the case in non-english locales.

openprojects:
I agree, it should sort by -option value- but it seems it is not doing it or I'm doing something wrong:
This is what I have in my placeholders.py (in UN plugin):

class EventLogoPlaceholder(DesignerPlaceholder):
    group = 'event'
    name = 'event_logo'
    description = _("Event Logo")

And the result from the BadgeDesigner UI is:

Screen Shot 2020-05-05 at 11 08 39

openprojects:
...probably it is something about the dictsort: the "keys" are mixed _LazyString (coming from core Indico) and unicode (coming from the plugin)

openprojects:
OK, I got it.
In indico.modules.designer.util.py , in def get_nested_placeholder_options , changing 
from:
`groups[placeholder.group]['options'][name] = placeholder.description`
to:
`groups[placeholder.group]['options'][name] = unicode(placeholder.description)`

What do you think about it?

ThiefMaster:
looks good

openprojects:
OK then, I'll close this PR and create a new one for that change. THANK YOU!

openprojects:
Perfect THANKS! Committing...

pferreir:
> OK, I got it.
> In indico.modules.designer.util.py , in def get_nested_placeholder_options , changing
> from:
> `groups[placeholder.group]['options'][name] = placeholder.description`
> to:
> `groups[placeholder.group]['options'][name] = unicode(placeholder.description)`
> 
> What do you think about it?

Shouldn't `name` be a `unicode` string in any case? At least that's what we have in `placeholder.py`. I suspect that your file is not using `from __future__ import unicode_literals`?

openprojects:
> Shouldn't `name` be a `unicode` string in any case? At least that's what we have in `placeholder.py`. I suspect that your source file is not using `from __future__ import unicode_literals`?

Agree it should be unicode and the ones coming from my plugin are unicode, but the ones coming from your placeholder.py are not (are _LazyString). 
This is from modules.designer.util.py:
Screen Shot 2020-05-05 at 18 25 53

ThiefMaster:
LGTM! But I'm curious, how is it *not* a LazyString for your placeholder? Assuming you also have the placeholder class defined on the module level, I'd expect yours to have a LazyString there as well...

openprojects:
> LGTM! But I'm curious, how is it _not_ a LazyString for your placeholder? Assuming you also have the placeholder class defined on the module level, I'd expect yours to have a LazyString there as well...

I totally agree, it is "bizarre" and I do not have an answer for that. 
The 2 placeholders.py have the same kind of declaration, but the final result is different.

pferreir:
As I told @ThiefMaster on Mattermost, this might be caused by a dynamic import when a request is already going on. If there's a request context, then you won't get a `LazyString`. Could that be the case @openprojects?

openprojects:
> As I told @ThiefMaster on Mattermost, this might be caused by a dynamic import when a request is already going on. If there's a request context, then you won't get a `LazyString`. Could that be the case @openprojects?

Well, both (in the core and in the plugin) PlaceHolder class objects properties (group, name, description) are loaded run time. Then yes, the instance is created while displaying the designer page, within a request context...  so I'm not sure...

ThiefMaster:
I think what he meant is whether the `import` of the module defining the class happens normally or at runtime (ie from within a function). If it's the latter, then the `_(...)` call runs during a request and thus does not return a LazyString.

openprojects:
> I think what he meant is whether the `import` of the module defining the class happens normally or at runtime (ie from within a function). If it's the latter, then the `_(...)` call runs during a request and thus does not return a LazyString.

Ah ok. That's too complex for me :) So you are then, probably that is the reason.

PR #4440, opened 2020-05-05
Move occurrences widget to react

### LINKED ISSUE #4399: Replace <input type="time"> fields with text-based fields
The way browsers handle time fields is not consistent across vendors and creates lots of problems especially to people on Edge, who have to deal with an awful built-in time picker widget.
Additionally, these widgets rely on the system locale, which may differ from the Indico locale.

We should replace these with a widget which is easy to use and based on whatever locale is selected in Indico.

Leats:
Also replaces time field in regform.
Closes #4399.

Bildschirmfoto 2020-05-07 um 09 33 40
Bildschirmfoto 2020-05-07 um 09 33 57
Bildschirmfoto 2020-05-07 um 09 34 08

### DIFF
--- a/indico/web/client/js/react/components/WTFOccurrencesField.jsx
+++ b/indico/web/client/js/react/components/WTFOccurrencesField.jsx
@@ -1,0 +1,177 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2020 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import _ from 'lodash';
+import React, {useState, useEffect, useMemo, useRef} from 'react';
+import moment from 'moment';
+import TimePicker from 'rc-time-picker';
+import PropTypes from 'prop-types';
+import {Translate} from 'indico/react/i18n';
+import {SingleDatePicker} from 'indico/react/components';
+
+function triggerChange(field) {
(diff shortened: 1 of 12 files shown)


panagiotappl:
It would be useful to provide a screenshot of a test case for this.

indico/web/client/js/react/components/WTFOccurrencesField.jsx
```diff
@@ -0,0 +1,177 @@
...
+  };
+
+  const updateDuration = event => {
+    changeOccurrence(occurrence.date, parseInt(event.target.value, 10), occurrence.time);
+  };
+
+  return (
+    <>
+      <SingleDatePicker
+        id=""
```
panagiotappl, line 115: `react-dates` documentation mentions that this prop is required but they use a `defaultProp` (with the value `"date"`) nonetheless (I also noticed that in another usage of this component we don't pass an `id` prop). So, I think you can omit this.

ThiefMaster: Nope. We actually fought with this one on the previous PR because if you don't set it to an empty string, it also adds `name="date"`, and thus breaks the logic to disable the submit button when there are no changes because then you have an additional named field in the form which gets included in determining its values...

So in cases where you don't care about the id and name of the underlying input field, you can indeed omit it. But in the wtforms context where the actual html form contents (as seen by jquery etc) matter, this is the only workaroud :/

PR #4445, opened 2020-05-06
Badgedesigner updates

openprojects:
As discussed in https://talk.getindico.io/t/badge-designer-adding-a-custom-image-field/1802/2
Adding "Event Logo" item in BadgeDesigner and "Height" property in the toolbar (only for image fields)

### DIFF
--- a/indico/modules/designer/client/js/index.js
+++ b/indico/modules/designer/client/js/index.js
@@ -469,12 +469,16 @@
       width: function() {
         selectedItem.width = Math.round($('.js-element-width').val() * pixelsPerCm);
         if (selectedItem.type === 'ticket_qr_code') {
-          $('#element-height').val($('.js-element-width').val());
+          $('.js-element-height').val($('.js-element-width').val());
           selectedItem.height = selectedItem.width;
         }
       },
       height: function() {
         selectedItem.height = Math.round($('.js-element-height').val() * pixelsPerCm);
+        if (selectedItem.type === 'ticket_qr_code') {
+          $('.js-element-width').val($('.js-element-height').val());
+          selectedItem.width = selectedItem.height;
+        }
       },
     }[attribute]());
 
--- a/indico/modules/designer/client/js/index.js
+++ b/indico/modules/designer/client/js/index.js
@@ -34,7 +34,7 @@
   var DEFAULT_PIXEL_CM = 50;
 
   // Should add "is_image" property in the json and the PlaceHolder
-  var imageTypes = ["ticket_qr_code","event_logo"];
+  const imageTypes = ['ticket_qr_code', 'event_logo'];
 
   function isImage(type) {
     return imageTypes.indexOf(type) > -1;
--- a/indico/modules/designer/client/js/index.js
+++ b/indico/modules/designer/client/js/index.js
@@ -6,7 +6,7 @@
 // LICENSE file for more details.
 
 /* global AlertPopup: true */
-/* eslint-disable max-len */
+/* eslint-disable max-len, import/unambiguous */
 
 (function(global) {
   'use strict';
--- a/indico/modules/designer/client/js/index.js
+++ b/indico/modules/designer/client/js/index.js
@@ -33,10 +33,10 @@
 
   var DEFAULT_PIXEL_CM = 50;
 
-  var imageTypes = ['ticket_qr_code', 'event_logo'];
+  let imageTypes = ['ticket_qr_code', 'event_logo'];
 
   function isImage(type) {
-    return imageTypes.indexOf(type) > -1;
+    return imageTypes.includes(type);
   };
 
   function zoom(val) {
@@ -757,7 +757,6 @@
 
     backsideTemplateID = backsideTemplate.id;
 
-    // Get image placeholder list
     imageTypes = image_types;
 
     // Item class
--- a/indico/modules/designer/client/js/index.js
+++ b/indico/modules/designer/client/js/index.js
@@ -33,6 +33,13 @@
 
   var DEFAULT_PIXEL_CM = 50;
 
+  // Should add "is_image" property in the json and the PlaceHolder
+  var imageTypes = ["ticket_qr_code","event_logo"];
+
+  function isImage(type) {
+    return imageTypes.indexOf(type) > -1;
+  };
+
   function zoom(val) {
     return val * zoomFactor;
   }
@@ -76,8 +83,8 @@
       text_align: 'center',
       color: 'black',
       font_size: '15pt',
-      width: type === 'ticket_qr_code' ? 150 : 400,
-      height: type === 'ticket_qr_code' ? 150 : null,
+      width: isImage(type) ? 150 : 400,
+      height: isImage(type) ? 150 : null,
       text: $T('Fixed text'),
 
       // The following attributes have no meaning to the server
@@ -237,6 +244,7 @@
     $('#style-selector').val(itemStyles.length ? itemStyles.join('_') : 'normal');
     $('#color-selector').val(item.color);
     $('.js-element-width').val(item.width / pixelsPerCm);
+    $('.js-element-height').val(item.height / pixelsPerCm);
 
     var $fixedTextField = $('#fixed-text-field');
     var $fontTools = $('.font-tools');
@@ -245,7 +253,7 @@
       $fontTools.fadeIn();
       $fixedTextField.closest('.tool').fadeIn();
       $fixedTextField.val(item.text);
-    } else if (item.type === 'ticket_qr_code') {
+    } else if (isImage(item.type)) {
       $fontTools.fadeOut();
       $fixedTextField.closest('.tool').fadeOut();
     } else {
@@ -462,8 +470,12 @@
       width: function() {
         selectedItem.width = Math.round($('.js-element-width').val() * pixelsPerCm);
         if (selectedItem.type === 'ticket_qr_code') {
+          $('#element-height').val($('.js-element-width').val());
           selectedItem.height = selectedItem.width;
         }
+      },
+      height: function() {
+        selectedItem.height = Math.round($('.js-element-height').val() * pixelsPerCm);
       },
     }[attribute]());
 
@@ -847,6 +859,10 @@
 
       $('.js-element-width').on('keyup click', function() {
         setSelectedItemAttribute('width', config);
+      });
+
+      $('.js-element-height').on('keyup click', function() {
+        setSelectedItemAttribute('height', config);
       });
 
       $('#fixed-text-field').on('keyup', function() {
--- a/indico/modules/designer/client/js/index.js
+++ b/indico/modules/designer/client/js/index.js
@@ -33,7 +33,7 @@
 
   var DEFAULT_PIXEL_CM = 50;
 
-  let imageTypes = ['ticket_qr_code', 'event_logo'];
+  let imageTypes = null;
 
   function isImage(type) {
     return imageTypes.includes(type);
--- a/indico/modules/designer/client/js/index.js
+++ b/indico/modules/designer/client/js/index.js
@@ -33,8 +33,7 @@
 
   var DEFAULT_PIXEL_CM = 50;
 
-  // Should add "is_image" property in the json and the PlaceHolder
-  const imageTypes = ['ticket_qr_code', 'event_logo'];
+  var imageTypes = ['ticket_qr_code', 'event_logo'];
 
   function isImage(type) {
     return imageTypes.indexOf(type) > -1;
@@ -741,7 +740,7 @@
     showConfirmationDialog = false;
   }
 
-  global.setupDesigner = function setupDesigner(template, backsideTemplate, config, placeholders) {
+  global.setupDesigner = function setupDesigner(template, backsideTemplate, config, placeholders, image_types) {
     editing = !!template;
     itemTitles = _.partial(_.extend, {}).apply(
       null,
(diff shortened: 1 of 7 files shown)


indico/modules/designer/placeholders.py
```diff
@@ -71,6 +75,22 @@ def render(cls, registration):
...
+    group = 'event'
+    name = 'event_logo'
+    description = _("Event Logo")
+
+    @classmethod
+    def render(cls, registration):
+        try:
+            buf = io.BytesIO(registration.logo)
+            img = Image.open(buf)
+            return img
```
Leats, line 88:
```suggestion
            buf = io.BytesIO(registration.logo)
            return Image.open(buf)
```
or even
```suggestion
            return Image.open(io.BytesIO(registration.logo))
```
(though in my opinion it feels less cluttered with the first option)

indico/modules/designer/placeholders.py
```diff
@@ -71,6 +75,22 @@ def render(cls, registration):
...
+        except Exception as e:
+            logger.error('Cannot get Event image logo: %s', e)
```
Leats, line 90:
```suggestion
            logger.error('Cannot get event image logo: %s', e)
```

indico/modules/designer/client/js/index.js
```diff
@@ -33,6 +33,13 @@
 
   var DEFAULT_PIXEL_CM = 50;
 
+  // Should add "is_image" property in the json and the PlaceHolder
+  var imageTypes = ["ticket_qr_code","event_logo"];
+
+  function isImage(type) {
+    return imageTypes.indexOf(type) > -1;
```
ThiefMaster, line 40: const isImage = type => imageTypes.includes(type);

(if you prefer keeping the `function` statement, still use `includes` instead of `indexOf`)

indico/modules/designer/client/js/index.js
```diff
@@ -33,6 +33,13 @@
 
   var DEFAULT_PIXEL_CM = 50;
 
+  // Should add "is_image" property in the json and the PlaceHolder
+  var imageTypes = ["ticket_qr_code","event_logo"];
```
ThiefMaster, line 37: `const`, single quotes and space after `,`

indico/modules/designer/placeholders.py
ThiefMaster, line 90: I don't think this try..except is really useful.

- catching any `Exception` is way too broad
- what kind of failures do you expect here? and even if you catch it, returning `None` will most likely break later

So I would just remove this altogether and if something breaks, then you keep a meaningful exception and traceback, which is better than handling it in a way where you either get a failure later or have a broken result (e.g. a badge with a missing image)

indico/modules/designer/client/js/index.js
ThiefMaster, line 37: What about actually passing the list of image placeholders to this function as well instead of hardcoding it? I had a look at the surrounding code and it would be pretty straightforward (build the list of image placeholder names in python, and pass it just like we pass `placeholders`).

indico/modules/designer/client/js/index.js
openprojects: I agree.
I was also thinking... in my plugin I have to add a 3rd Image object, that is really "UN custom": Band. What can I do to add it in the plugin without customising the entire index.js ? can I keep the imageTypes in a separate JS file? Or maybe add a suffix: like event_logo_image , ticket_qr_code_image ? Not elegant but...

indico/modules/designer/placeholders.py
openprojects: Yes, the Exception is not nice... I'll change it for an IOException. I want to catch the missing file.
On the other hand... imagine being a Security Officer trying to print a badge with a long queue in front: better get an Indico error, or a badge even without the correct logo? :)
I'll fix the Exception and return something...

indico/modules/designer/client/js/index.js
ThiefMaster: with https://github.com/indico/indico/pull/4445/files#r420941411 this won't be a problem

indico/modules/designer/placeholders.py
```diff
@@ -71,6 +75,21 @@ def render(cls, registration):
         return name
 
 
+class EventLogoPlaceholder(DesignerPlaceholder):
+    group = 'event'
+    name = 'event_logo'
+    description = _("Event Logo")
+
+    @classmethod
+    def render(cls, registration):
```
ThiefMaster, line 84: this is an `event`, not a `registration`

indico/modules/designer/placeholders.py
ThiefMaster: But... you cannot have a missing file here. the event logo is stored in the database.

What can happen of course is that the event has no logo at all, because none has been uploaded, or someone deleted it. But this case should *not* be handled here - instead, check `event.has_logo` and if there either use a placeholder image (like in case of broken images on the web) or return None **but make sure it's handled cleanly in the code later (which may be hard, because it checks for isinstance Image to know whether it's an image or text)

openprojects: Ah ok. 
I had in mind the logo was on FS.
OK, I'm going to change it

openprojects: I checked it out, returning an empty string will print the badge without logo and no errors.

indico/modules/designer/client/js/index.js
openprojects: Ahhhh!! Now what you wrote above IS CLEAR :)
YES!! Totally agree... ! This will be flexible and allow additional images declared in the plugin.
PERFECT. I'll take a look at it.

indico/modules/designer/placeholders.py
openprojects: Yeah, sorry.

indico/modules/designer/util.py
```diff
@@ -32,6 +32,14 @@ def get_nested_placeholder_options():
     return groups
 
 
+def get_images_placeholder_types():
+    image_types = []
+    for name, placeholder in get_placeholder_options().viewitems():
+        if getattr(placeholder, 'is_image', False):
+            image_types.append(name)
+    return image_types
```
ThiefMaster, line 40: This can be simplified a lot:

```suggestion
    return [name for name, placeholder in get_placeholder_options().viewitems() if placeholder.is_image]
```

to avoid the getattr check you need to set `is_image = False` in the `DesignerPlaceholder` base class, but I think makes sense anyway:

```python

    #: Whether this placeholder is rendering an image
    is_image = False
```

indico/modules/designer/client/js/index.js
```diff
@@ -746,6 +757,9 @@
 
     backsideTemplateID = backsideTemplate.id;
 
+    // Get image placeholder list
+    imageTypes = image_types;
```
ThiefMaster, line 761: Oh yuck, I was just about to comment on why not just put `imageTypes` in the function's argument list, then I realized the huge mess we have there with global-like variables...

I don't think this line needs a comment btw.

indico/modules/designer/client/js/index.js
```diff
@@ -33,6 +33,12 @@
 
   var DEFAULT_PIXEL_CM = 50;
 
+  var imageTypes = ['ticket_qr_code', 'event_logo'];
```
ThiefMaster, line 36: I don't think we still need this hardcoded list. Also, use `let` so you can reassign later; `var` should not be used in any new code.

indico/modules/designer/client/js/index.js
```diff
@@ -33,6 +33,12 @@
...
+
+  function isImage(type) {
+    return imageTypes.indexOf(type) > -1;
```
ThiefMaster, line 39: `imageTypes.includes(type)`

indico/modules/designer/client/js/index.js
```diff
@@ -462,9 +469,13 @@
       width: function() {
         selectedItem.width = Math.round($('.js-element-width').val() * pixelsPerCm);
         if (selectedItem.type === 'ticket_qr_code') {
```
ThiefMaster, line 471: shouldn't this use `isImage` as well?

indico/modules/designer/util.py
openprojects: Yes, I agree.

indico/modules/designer/client/js/index.js
openprojects: OK, i'll put `let`. I would like to keep it just as a default...

indico/modules/designer/client/js/index.js
openprojects: No, this is specific to ticket_qr_code because it is squared: so if you modify width, I set the height accordingly

ThiefMaster: OK fair enough

indico/modules/designer/client/js/index.js
ThiefMaster: I disagree with keeping a default that is never used anyway... if the code never uses the array I'd just default to `null` so if something does end up using it too early, it breaks instead of silently using different data..

indico/modules/designer/placeholders.py
```diff
@@ -71,6 +76,21 @@ def render(cls, registration):
...
 
+class EventLogoPlaceholder(DesignerPlaceholder):
+    group = 'event'
+    name = 'event_logo'
+    description = _("Event Logo")
+    is_image = True
+
+    @classmethod
+    def render(cls, event):
+        if event.has_logo:
```
ThiefMaster, line 87:
```python
if not event.has_logo:
    return ''
buf = BytesIO(event.logo)
return Image.open(buf)
```

(and use `from io import BytesIO` for the import)

indico/modules/designer/util.py
```diff
@@ -32,6 +32,10 @@ def get_nested_placeholder_options():
     return groups
 
 
+def get_images_placeholder_types():
```
ThiefMaster, line 35: I think `get_image_placeholder_types` (singular "image") would be more correct

indico/modules/events/registration/badges.py
```diff
@@ -87,7 +87,12 @@ def _draw_badge(self, canvas, registration, template, tpl_data, pos_x, pos_y):
 
         placeholders = get_placeholders('designer-fields')
 
-        for item in tpl_data.items:
+        # Print images first
+        images = [key for key in placeholders if hasattr(placeholders[key], 'is_image')]
```
ThiefMaster, line 91: this won't work since all placeholders have it

indico/modules/events/registration/badges.py
```diff
@@ -87,7 +87,12 @@ def _draw_badge(self, canvas, registration, template, tpl_data, pos_x, pos_y):
...
+        items = []
+        map(lambda x: items.insert(0, x) if x['type'] in images else items.append(x), tpl_data.items)
```
ThiefMaster, line 93: Please don't use `map()` as a replacement for a for loop where you don't care about return values. This is not how `map()` is intended to be used and will cease to work in Python 3 where it returns an interator. What about something like this instead?

```python
image_placeholders = {name for name, placeholder in placeholders.viewitems() if placeholder.is_image}
image_items = [x for x in items if x['type'] in image_placeholders]
text_items = [x for x in items if x['type'] not in image_placeholders]

for item in chain(image_items, text_items):
    ...
```

ThiefMaster: Actually, even easier:

```python
image_placeholders = {name for name, placeholder in placeholders.viewitems() if placeholder.is_image}
items = tpl_data.items.sort(key=lambda item: item['type'] not in image_placeholders)
for item in items:
    ...
```

This will sort anything that is in `image_placeholders` first.

indico/modules/events/registration/badges.py
openprojects: Correct. I'm going to fix it:

`images = [key for key in placeholders if placeholders[key].is_image]`

indico/modules/events/registration/badges.py
openprojects: I really like this last version!
I didn't like the way I did it but I was not able to find a nicer way... this definitely nailed it :)

openprojects: Just one thing... the `for` should be using` tpl_data.items` :
`for item in tpl_data.items:`
because `items` is `None`

openprojects: I can even change this:
`items = tpl_data.items.sort(key=lambda item: item['type'] not in image_placeholders)`
to this:
`tpl_data.items.sort(key=lambda item: item['type'] not in image_placeholders)`
What do you think?

ThiefMaster: sorry, i meant this of course. not sure why i went for the inplace sort (which is wrong here since you don't want to modify the original list!):

```python
items = sorted(tpl_data.items, key=lambda item: item['type'] not in image_placeholders)
```

indico/modules/events/registration/badges.py
ThiefMaster: please see my line in the comment thread just below ;) it's a bit cleaner

indico/modules/designer/templates/template.html
```diff
@@ -370,7 +377,7 @@
             'use strict';
 
             setupDesigner({{ template_data | tojson }}, {{ backside_template_data | tojson}}, {{ config | tojson }},
```
ThiefMaster, line 379:
```suggestion
            setupDesigner({{ template_data | tojson }}, {{ backside_template_data | tojson }}, {{ config | tojson }},
```

(not related to your change, just noticed it)

openprojects:
Perfect, thank you!

PR #4452, opened 2020-05-12
Apply proper editing permission checks

### LINKED ISSUE #4417: Editing permission checks
Right now we only have very basic checks, i.e. whether you are an editor in general, but you can edit anything even without being assigned as the current editor on the editable.

Unless you are a manager (or admin) you should only be able to edit an editable where you are assigned as the editor.

It's probably a good idea to also add a new 'editing' permission ("Editing Manager") for general Editing management, which would be like event manager access but only within the editing module. No need for new UI here; especially for now it can be assigned through the normal event protection page. But this could be made a separate task as well...

ThiefMaster:
closes #4417

### DIFF
--- a/indico/modules/events/contributions/models/contributions.py
+++ b/indico/modules/events/contributions/models/contributions.py
@@ -515,9 +515,6 @@ class Contribution(DescriptionMixin, ProtectionManagersMixin, LocationMixin, Att
         order = list(EditableType)
         return sorted(enabled_editables, key=lambda editable: order.index(editable.type))
 
-    def can_submit_editables(self, user):
-        return self.is_user_associated(user, check_abstract=True)
-
     @property
     def has_published_editables(self):
         return any(e.published_revision_id is not None for e in self.enabled_editables)
@@ -550,6 +547,16 @@ class Contribution(DescriptionMixin, ProtectionManagersMixin, LocationMixin, Att
         if check_abstract and self.abstract and self.abstract.submitter == user:
             return True
         return any(pl.person.user == user for pl in self.person_links if pl.person.user)
+
+    def can_submit_proceedings(self, user):
+        """Whether the user can submit editables/papers."""
+        if user is None:
+            return False
+        # The submitter of the original abstract is always authorized
+        if self.abstract and self.abstract.submitter == user:
+            return True
+        # Otherwise only users with submission rights are authorized
+        return self.can_manage(user, 'submit', allow_admin=False, check_parent=False)
 
 
 Contribution.register_protection_events()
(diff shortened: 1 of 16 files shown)


indico/modules/events/contributions/models/contributions.py
```diff
@@ -515,6 +515,9 @@ def enabled_editables(self):
         order = list(EditableType)
         return sorted(enabled_editables, key=lambda editable: order.index(editable.type))
 
+    def can_submit_editables(self, user):
+        return self.is_user_associated(user, check_abstract=True)
```
mic4ael, line 519: Why `check_abstract=True`? Do we care about abstracts here?

indico/modules/events/editing/client/js/editing/timeline/TimelineItem.jsx
```diff
@@ -84,8 +85,9 @@ export default function TimelineItem({block}) {
                   files={block.files}
                   downloadURL={block.downloadFilesURL}
                 />
-                {/* TODO: Check whether the current user is submitter */}
-                {isLastBlock && needsSubmitterConfirmation && <ChangesConfirmation />}
+                {isLastBlock && needsSubmitterConfirmation && canPerformSubmitterActions && (
```
mic4ael, line 88: maybe `canPerformSubmitterAction` should be the first condition? I think it would be logically more correct (even though it doesn't change much)

ThiefMaster: since it's not doing any function calls i don't care much about the order ;) but you are right, reversing the order of these conditions is more logical

indico/modules/events/contributions/models/contributions.py
ThiefMaster: That way the user who submitted the abstract always has access, even if they are not listed as an author. We do the same e.g. for papers, and I think it makes sense: After all, if you submitted the abstract you may be the person who's "handling" that contribution, even if you are not an author/speaker yourself (but e.g. a secretary for the team who wrote it).

indico/modules/events/editing/models/editable.py
```diff
@@ -95,8 +95,76 @@ def locator(self):
...
+        or approving/rejecting the editable.
+        """
+        # If the user can't even see the timeline, we never allow any modifications
+        if not self.can_see_timeline(user):
+            return False
+        # Editing/event managers can perform actions without being the assigned editor
+        # XXX: Do we want this? Or should they have to assign themselves first if they
+        #      want to do actions that would usually be done by the assigned editor?
+        if self.event.can_manage(user, permission='editing_manager'):
+            return True
```
ThiefMaster, line 152: I'll send a separate PR to require them to be assigned as well.

PR #4454, opened 2020-05-12
Use react-overridable for Overridable

ThiefMaster:
Depends on indico/react-overridable#1

# TODO

- [x] Install package from npm
- [x] Get rid of `Overridable.jsx` (check what to do with `RouteAwareOverridable`)
- [x] Update burotel plugin as well - indico/indico-plugins-cern#76


indico/modules/rb/client/js/setup.jsx
```diff
@@ -33,14 +34,7 @@ export default function setup(overrides = {}, postReducers = []) {
 
   document.addEventListener('DOMContentLoaded', () => {
     const appContainer = document.getElementById('rb-app-container');
-    const store = createRBStore(overrides, postReducers);
-
-    // Use this function whenever you would like to override
-    // a component e.g. from within a plugin without
-    // using the `setup` function
-    window.registerOverride = externalOverrides => {
```
ThiefMaster, line 41: I removed this for now since we:

- don't actually use it
- will probably need react state to store the overrides once we want to add something like this back

pferreir: Maybe we were using this in the shower site? 😆

ThiefMaster: nope, that's using `setup()`

PR #4456, opened 2020-05-15
Remove circular dependencies from the room booking module 🎉

plourenco:
This issue was first tracked in #4408, causing tests to fail when resolving imports with cycles. It became pretty hard to track where these imports where cycling. ESLint rule `import/no-cycle` returned:

```
> npx eslint --rule 'import/no-cycle:1' indico/modules/rb
✖ 39 problems (38 error, 1 warning)
```

Initially thought to be caused by `index.js` files (where most imports overlap), I ended up tracking down `import {history} from 'store.js'` as a cause of about 95% of them. As a result, this was moved into a new file.

Some minor changes were also applied to remove a circular dependency in `modules/roomSearch` caused by importing with `index.js`. Those were updated to individual imports.

Also enables eslint rule `import/no-cycle` with a warning. 
Even though in the future we should always avoid it.


package.json
```diff
@@ -169,6 +169,7 @@
   },
   "homepage": "https://github.com/indico/indico#readme",
   "jest": {
+    "transformIgnorePatterns": ["/node_modules/(?!geodesy)"],
```
pferreir, line 172: Maybe I missed some discussion, but why are we not ignoring the `geodesy` package?

ThiefMaster: correct me if i missed anything @plourenco, but iirc this package uses *only* ES modules so it's the only one that needs to be transformed atm when running jest even though it's something from `node_modules`

plourenco: Yes, because in `indico/modules/rb/client/js/common/map/util.js` line 14:

```
import LatLon from 'geodesy/latlon-nvector-spherical';
```

We are importing that module directly which was published untranspiled, it is a pure ES6 class. Since all files inside node_modules are not transformed by default, Jest will not understand the code in these modules, resulting in syntax errors.

The same happens with `semantic-ui-react` in #4461, where I'm ignoring it.

pferreir: OK, just wanted to clarify that, thanks!

indico/modules/rb/client/js/store.js
```diff
@@ -5,26 +5,25 @@
...
 import {
   routeConfig as calendarRouteConfig,
   queryStringReducer as qsCalendarReducer,
-} from './modules/calendar';
+} from './modules/calendar/queryString';
 import {
   routeConfig as blockingsRouteConfig,
   queryStringReducer as qsBlockingsReducer,
-} from './modules/blockings';
+} from './modules/blockings/queryString';
```
pferreir, line 26: I guess you did this to avoid importing the `index.js` directly. Have you checked whether in these cases you still need to re-export `routeConfig` and `queryStringReducer` in it? I mean [this line](https://github.com/indico/indico/blob/master/indico/modules/rb/client/js/modules/blockings/index.js#L16). If we don't need it, we're better off removing as it may be misleading for future code.

indico/modules/rb/client/js/modules/bookRoom/actions.js
```diff
@@ -18,7 +18,8 @@ import {validateFilters} from '../../common/filters';
 import {preProcessParameters} from '../../util';
 import {actions as modalActions} from '../../modals';
 
-import {roomSearchActionsFactory, ajaxRules as ajaxFilterRules} from '../../common/roomSearch';
+import {roomSearchActionsFactory} from '../../common/roomSearch/actions';
+import {ajax as ajaxFilterRules} from '../../common/roomSearch/serializers';
```
pferreir, line 22: Same here.

indico/modules/rb/client/js/modules/bookRoom/selectors.js
```diff
@@ -8,9 +8,9 @@
 import moment from 'moment';
 import {createSelector} from 'reselect';
 import {RequestState} from 'indico/utils/redux';
-import {roomSearchSelectorFactory} from '../../common/roomSearch';
-import {selectors as roomsSelectors} from '../../common/rooms';
-import {selectors as userSelectors} from '../../common/user';
+import {roomSearchSelectorFactory} from '../../common/roomSearch/selectors';
+import * as roomsSelectors from '../../common/rooms/selectors';
+import * as userSelectors from '../../common/user/selectors';
```
pferreir, line 13: Same here.

indico/modules/rb/client/js/modules/bookRoom/actions.js
plourenco: The thing is not all the imports at `roomSearch/index` resolve to a circular dependency. The only ones that are very deeply linked are `roomSearch/index -> importing -> actions/reducers` with `bookRoom/index -> importing -> actions/reducers`. 
Which means that we can still use the index on other occasions.

I agree that this becomes unconventional because some are direct imports, others go through the index so doing that change is one option.

However, in the case below of `rooms/selectors` and `user/selectors`, multiple imports depend on that, breaking them out of index would result in a lot of file changes.

Nonetheless, since this seems like an isolated case, what if we kept the imports and just ignored the rule here?

pferreir: I don't know. I agree that if the "indirect" (shorter) import works OK in other places, we can keep it. OTOH, does it really hurt changing all occurrences to the direct one? Let's maybe agree on a standard? @ThiefMaster what do you think?

ThiefMaster: I guess being consistent is better - and it's a one-time change so even if it affects more files, it's probably the Right Thing ™ to do...

And since it's only within roombooking it should be easily testable too and thus be a rather safe change.

plourenco: Just as a quick comparison, before going into erradicating all indexes, check my latest commit.

ThiefMaster: hm am i misunderstanding something? this commit looks more like the reverse of what i thought you'd do. or is it just to demonstrate the reverse of what you'd apply to the other files?

pferreir: I don't think the point would be to eradicate the indices, I would only do it for these particular exports...

indico/modules/rb/client/js/common/roomSearch/reducers.js
```diff
@@ -12,7 +12,7 @@ import {requestReducer} from 'indico/utils/redux';
 import {filterReducerFactory} from '../filters';
 import {mapSearchReducerFactory} from '../map';
 import {roomSearchActionsFactory} from './actions';
-import {actions as bookRoomActions} from '../../modules/bookRoom';
+import * as bookRoomActions from '../../modules/bookRoom/actions';
```
plourenco, line 15: Even though the previous commit was to avoid direct imports clashing with importing from the index. There isn't really anything I can do about this particular case.

Removing the actions export in bookRoom would result in multiple changes that aren't really effective since `roomSearch` is the only import that creates a cycle. I think this is reasonable.

PR #4458, opened 2020-05-18
Add editing notifications

### LINKED ISSUE #4437: Editing: Notifications
Send email notifications on various events happening to an editable

- No message customizations or settings to enable/disable emails
- Plaintext emails are probably enough
- Send notifications on all **relevant events** to the people for **whom they matter**
  - new comment by submitter -> notify editor
  - new comment by editor -> notify submitter
  - judged by editor -> notify submitter
  - new revision uploaded by submitter -> notify editor
  - revision changes made by editor -> notify submitter who made the previous revision
  - revision changes approved/rejected by submitter -> notify editor (the currently assigned one and the one who was assigned for that particular revision in case they are different)
  - Anything else? TBD

Corner cases mentioned in MM chat:
- user is not an editor nor a submitter for the editable (!can_perform_submitter_actions, !can_perform_editor_actions): send emails to both the submitter of the revision and the editor of the editable
- same if he's both?
- user is not an editor for the editable: send the email to the submitter of the revision
- user is not a submitter for the editable: send the email to the editor of the editable

panagiotappl:
closes #4437 

Set in draft mode because some corner cases are missing.

### DIFF
--- a/indico/modules/events/editing/notifications.py
+++ b/indico/modules/events/editing/notifications.py
@@ -1,0 +1,82 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2020 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from indico.core.notifications import make_email, send_email
+from indico.modules.events.editing.schemas import EditingConfirmationAction
+from indico.web.flask.templating import get_template_module
+
+
+def notify_submitter_comment(comment):
+    """
+    Notify the editor about comments made by a submitter.
+    """
+    revision = comment.revision
+    editor = revision.editable.editor
+    submitter = comment.user
+    if not editor:
+        return
+    tpl = get_template_module('events/editing/emails/submitter_comment_notification.txt',
+                              submitter_name=submitter.first_name,
+                              timeline_url=revision.editable.external_timeline_url,
+                              recipient_name=editor.first_name)
+    send_email(make_email(editor.email, template=tpl))
+
+
+def notify_editor_comment(comment):
+    """
+    Notify the submitter about comments made by an editor.
+    """
+    revision = comment.revision
+    editor = comment.user
+    submitter = revision.submitter
+    tpl = get_template_module('events/editing/emails/editor_comment_notification.txt',
+                              editor_name=editor.first_name,
+                              timeline_url=revision.editable.external_timeline_url,
+                              recipient_name=submitter.first_name)
+    send_email(make_email(submitter.email, template=tpl))
+
+
+def notify_editor_judgment(revision, editor):
+    """Notify the submitter about a judgment made by an editor."""
+    submitter = revision.submitter
+    tpl = get_template_module('events/editing/emails/editor_judgment_notification.txt',
+                              editor_name=editor.first_name,
+                              timeline_url=revision.editable.external_timeline_url,
+                              recipient_name=submitter.first_name)
+    send_email(make_email(submitter.email, template=tpl))
+
+
+def notify_submitter_upload(revision):
+    """Notify the editor about the submitter uploading a new revision."""
+    submitter = revision.submitter
+    editor = revision.editable.editor
+    if not editor:
+        return
+    tpl = get_template_module('events/editing/emails/submitter_upload_notification.txt',
+                              submitter_name=submitter.first_name,
+                              timeline_url=revision.editable.external_timeline_url,
+                              recipient_name=editor.first_name)
+    send_email(make_email(editor.email, template=tpl))
+
+
+def notify_submitter_confirmation(revision, submitter, action):
+    """Notify the editor(s) about submitter accepting/rejecting revision changes."""
+    editable = revision.editable
+    current_editor = editable.editor
+    prev_revision_editor = editable.revisions[-2].editor
+    recipients = {current_editor, prev_revision_editor}
+    recipients.discard(None)
+    if action == EditingConfirmationAction.accept:
+        template_path = 'events/editing/emails/submitter_confirmation_notification.txt'
+    else:
+        template_path = 'events/editing/emails/submitter_rejection_notification.txt'
+    for recipient in recipients:
+        tpl = get_template_module(template_path,
+                                  submitter_name=submitter.first_name,
+                                  timeline_url=revision.editable.external_timeline_url,
+                                  recipient_name=recipient.first_name)
+        send_email(make_email(recipient.email, template=tpl))
--- a/indico/modules/events/editing/notifications.py
+++ b/indico/modules/events/editing/notifications.py
@@ -10,34 +10,40 @@ from indico.modules.events.editing.schemas import EditingConfirmationAction
 from indico.web.flask.templating import get_template_module
 
 
-def notify_submitter_comment(comment):
-    """
-    Notify the editor about comments made by a submitter.
-    """
+def _get_commenting_users(revision, check_internal_access=False):
+    return {
+        c.user
+        for c in revision.comments
+        if not check_internal_access or revision.editable.can_use_internal_comments(c.user)
+    }
+
+
+def notify_comment(comment):
+    """Notify about a new comments on a revision."""
     revision = comment.revision
     editor = revision.editable.editor
-    submitter = comment.user
-    if not editor:
-        return
-    tpl = get_template_module('events/editing/emails/submitter_comment_notification.txt',
-                              submitter_name=submitter.first_name,
-                              timeline_url=revision.editable.external_timeline_url,
-                              recipient_name=editor.first_name)
-    send_email(make_email(editor.email, template=tpl))
+    author = comment.user
+    if comment.internal:
+        # internal comments notify the editor and anyone who commented + can see internal comments
+        recipients = _get_commenting_users(revision, check_internal_access=True) | {editor}
+    elif author == editor:
+        # editor comments notify the submitter and anyone else who commented
+        recipients = _get_commenting_users(revision) | {revision.submitter}
+    elif revision.editable.can_perform_submitter_actions(author):
+        # submitter comments notify the editor and anyone else who commented
+        recipients = _get_commenting_users(revision) | {editor}
+    else:
+        # comments from someone else (managers) notify everyone
+        recipients = _get_commenting_users(revision) | {editor, revision.submitter}
 
-
-def notify_editor_comment(comment):
-    """
-    Notify the submitter about comments made by an editor.
-    """
-    revision = comment.revision
-    editor = comment.user
-    submitter = revision.submitter
-    tpl = get_template_module('events/editing/emails/editor_comment_notification.txt',
-                              editor_name=editor.first_name,
-                              timeline_url=revision.editable.external_timeline_url,
-                              recipient_name=submitter.first_name)
-    send_email(make_email(submitter.email, template=tpl))
+    recipients.discard(None)  # in case there's no editor assigned
+    recipients.discard(author)  # never bother people about their own comments
+    for recipient in recipients:
+        tpl = get_template_module('events/editing/emails/comment_notification.txt',
+                                  author_name=author.first_name,
+                                  timeline_url=revision.editable.external_timeline_url,
+                                  recipient_name=recipient.first_name)
+        send_email(make_email(recipient.email, template=tpl))
 
 
 def notify_editor_judgment(revision, editor):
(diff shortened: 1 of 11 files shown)


indico/modules/events/editing/notifications.py
```diff
@@ -0,0 +1,96 @@
...
+
+from indico.core.notifications import email_sender, make_email
+from indico.web.flask.templating import get_template_module
+from indico.web.flask.util import url_for
+
+
+def _get_timeline_url(revision):
+    editable = revision.editable
+    contribution = editable.contribution
+    return url_for('event_editing.editable', contribution.event, contribution, type=editable.type.name, _external=True)
```
mic4ael, line 16: You can pass only `contribution` since `Contribution`'s locator_property already returns event-related attributes.

```suggestion
    return url_for('event_editing.editable', contribution, type=editable.type.name, _external=True)
```

indico/modules/events/editing/notifications.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@email_sender
+def notify_submitter_comment(revision):
+    """
+    Notifies the editor about comments made by a submitter.
+    """
+    editor = revision.editable.editor
+    subject = 'New submitter comment'
```
mic4ael, line 25: I am wondering whether the subjects should be passed like that to the templates. As far as I can tell each template has a different subject thus you could probably put those strings directly in the `txt` files.

mic4ael: Of course, still using the `block subject` since it is being used in the `get_subject` macro

ThiefMaster: Yes, that's exactly what the subject block is meant for :)

indico/modules/events/editing/notifications.py
ThiefMaster: even easier, `url_for('event_editing.editable', editable, _external=True)` since `Editable` has its own locator property.

mic4ael: Yeah, even better ;)

indico/modules/events/editing/templates/emails/base_email.txt
```diff
@@ -0,0 +1,8 @@
+{% extends 'emails/base.txt' %}
+
+{% block header_recipient -%} {{ recipient_name }} {%- endblock %}
```
ThiefMaster, line 3: this space here and in the other templates is a bit weird.. since it gets stripped anyway I'd just use linebreaks (and then indentation as well) if you want to keep some space for readability:

```
{% block header_recipient -%}
    {{ recipient_name }}
{%- endblock %}
```

or

```
{% block header_recipient -%}{{ recipient_name }}{%- endblock %}
```

indico/modules/events/editing/templates/emails/submitter_upload_notification.txt
```diff
@@ -0,0 +1,7 @@
+{% extends 'events/editing/emails/base_email.txt' %}
+
+{% block subject -%} New submitter upload {%- endblock %}
+
+{% block body_text -%}
+{ submitter_name } has uploaded a new revision to an editable you are editor of.
```
ThiefMaster, line 6: missing a pair of `{...}`?

indico/modules/events/editing/operations.py
```diff
@@ -205,6 +211,13 @@ def create_revision_comment(revision, user, text, internal=False):
     comment = EditingRevisionComment(user=user, text=text, internal=internal)
     revision.comments.append(comment)
     db.session.flush()
+    editable = revision.editable
+    if editable.can_perform_submitter_actions(user):
+        notify_submitter_comment(revision)
+    if editable.can_perform_editor_actions(user):
+        notify_editor_comment(revision)
+    if not editable.can_perform_submitter_actions(user) and not editable.can_perform_editor_actions(user):
```
ThiefMaster, line 219: `elif not editable.can_perform_submitter_actions(user):` is enough here, since the other criterion is implied when using elif

indico/modules/events/editing/notifications.py
```diff
@@ -0,0 +1,88 @@
...
+from indico.web.flask.templating import get_template_module
+from indico.web.flask.util import url_for
+
+
+def _get_timeline_url(revision):
+    editable = revision.editable
+    return url_for('event_editing.editable', editable, _external=True)
+
+
+@email_sender
```
ThiefMaster, line 20: I'm not a big fan of the decorator... I would just use `send_email(make_email(...))` in the function itself.

indico/modules/events/editing/notifications.py
```diff
@@ -0,0 +1,88 @@
...
+    current_editor = editable.editor
+    prev_revision_editor = editable.revisions[-2].editor
+    recipients = {current_editor, prev_revision_editor}
+    template_path = 'events/editing/emails/submitter_confirmation_notification.txt'
+    return [make_email(recipient.email,
+                       template=get_template_module(template_path,
+                                                    submitter_name=submitter.first_name,
+                                                    timeline_url=_get_timeline_url(revision),
+                                                    recipient_name=recipient.first_name)
+                       ) for recipient in recipients]
```
ThiefMaster, line 88: This one actually becomes more readable when either not using the decorator or using it with `yield`:

```python
    for recipient in recipients:
        tpl = get_template_module('events/editing/emails/submitter_confirmation_notification.txt',
                                  submitter_name=submitter.first_name,
                                  timeline_url=_get_timeline_url(revision),
                                  recipient_name=recipient.first_name)
        yield make_email(recipient.email, template=tpl)  
```

indico/modules/events/editing/operations.py
```diff
@@ -119,6 +122,7 @@ def review_editable_revision(revision, editor, action, comment, tags, files=None
                                        tags=revision.tags)
         revision.editable.revisions.append(new_revision)
     db.session.flush()
+    notify_editor_judgement(revision)
```
ThiefMaster, line 125: We use `judgment` everywhere else

indico/modules/events/editing/operations.py
```diff
@@ -136,6 +140,7 @@ def confirm_editable_changes(revision, submitter, action, comment):
     if action == EditingConfirmationAction.accept:
         publish_editable_revision(revision)
     db.session.flush()
+    notify_submitter_confirmation(revision)
```
ThiefMaster, line 143: If the submitter rejects the changes the email you are sending here is rather misleading. I'd send a different one mentioning that the submitter rejected the editor's changed in case of `EditingConfirmationAction.reject`.

indico/modules/events/editing/notifications.py
```diff
@@ -0,0 +1,88 @@
...
+
+from indico.core.notifications import make_email, send_email
+from indico.modules.events.editing.schemas import EditingConfirmationAction
+from indico.web.flask.templating import get_template_module
+from indico.web.flask.util import url_for
+
+
+def _get_timeline_url(revision):
+    editable = revision.editable
+    return url_for('event_editing.editable', editable, _external=True)
```
ThiefMaster, line 18: What about adding this as an `external_timeline_url` property to the `Editable` model?

indico/modules/events/editing/notifications.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def notify_submitter_comment(revision):
+    """
+    Notifies the editor about comments made by a submitter.
```
ThiefMaster, line 23: "Notify ..." (PEP257)

Feel free to change the wording in the other docstrings that are wrong, too. We didn't follow this rule initially and only started at some point between 1.9.x and 2.0, so older docstrings are still "wrong"

indico/modules/events/editing/templates/emails/submitter_confirmation_notification.txt
```diff
@@ -0,0 +1,7 @@
+{% extends 'events/editing/emails/base_email.txt' %}
+
+{% block subject -%}New submitter confirmation{%- endblock %}
+
+{% block body_text -%}
+Your revision changes have been {{ action }} by {{ submitter_name }}.
```
ThiefMaster, line 6: What about two separate templates for the actions? I think the subject could be different based on the action as well (e.g. "Submitter confirmed changes" and "Submitter rejected changes"). It's a rather important difference, so I think having it in the subject is worth it.

indico/modules/events/editing/notifications.py
```diff
@@ -0,0 +1,84 @@
...
+from indico.modules.events.editing.schemas import EditingConfirmationAction
+from indico.web.flask.templating import get_template_module
+
+
+def notify_submitter_comment(revision):
+    """
+    Notify the editor about comments made by a submitter.
+    """
+    submitter = revision.submitter
+    editor = revision.editable.editor
```
ThiefMaster, line 20: If `editor` is None this fails below; let's just return if there is no editor.

indico/modules/events/editing/notifications.py
```diff
@@ -0,0 +1,84 @@
...
+    tpl = get_template_module('events/editing/emails/submitter_comment_notification.txt',
+                              submitter_name=submitter.first_name, timeline_url=revision.editable.external_timeline_url,
+                              recipient_name=editor.first_name)
+    send_email(make_email(editor.email, template=tpl))
+
+
+def notify_editor_comment(revision, admin=False):
```
ThiefMaster, line 27: maybe `manager` instead of `admin` is more suitable here? while an admin is manager-like in this case, it's more likely to be a manager than an indico admin leaving a comment

indico/modules/events/editing/notifications.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    Notify the submitter about comments made by an editor.
+    """
+    submitter = revision.submitter
+    editor = session.user if admin else revision.editable.editor
```
ThiefMaster, line 32: Better pass the `user` as an argument - functions in `operations` and anything called from there should usually not rely on the session...

The code here also *assumes* that the comment was created by `revision.editable.editor`, even though you *know* who the creator is in the operation.

indico/modules/events/editing/notifications.py
```diff
@@ -0,0 +1,84 @@
...
+                              recipient_name=submitter.first_name)
+    send_email(make_email(submitter.email, template=tpl))
+
+
+def notify_editor_judgment(revision):
+    """
+    Notify the submitter about a judgement made by an editor.
+    """
+    submitter = revision.submitter
+    editor = revision.editable.editor
```
ThiefMaster, line 45: Same here, when a user performs an operation we know who the user is - I would not infer the editor whom who's assigned. While it should usually match, it's better to stay on the safe side for cases where we know the correct user anyway.

indico/modules/events/editing/notifications.py
panagiotappl: So, there's no reason to use the `if` here and just always use the user?

indico/modules/events/editing/templates/emails/editor_comment_notification.txt
```diff
@@ -0,0 +1,7 @@
+{% extends 'events/editing/emails/base_email.txt' %}
+
+{% block subject -%} New editor comment {%- endblock %}
```
mic4ael, line 3:
```suggestion
{% block subject -%}New editor comment{%- endblock %}
```

indico/modules/events/editing/operations.py
```diff
@@ -205,6 +211,10 @@ def create_revision_comment(revision, user, text, internal=False):
     comment = EditingRevisionComment(user=user, text=text, internal=internal)
     revision.comments.append(comment)
     db.session.flush()
+    editable = revision.editable
+    if editable.can_perform_submitter_actions(user):
+        notify_submitter_comment(revision)
```
ThiefMaster, line 216: Pass the `user` here as well instead of relying on `revision.submitter` in there (there can be multiple people with submitter permissions and the email should mention the user who actually posted the comment).

Alternatively, pass the `comment` to the function since that one already has both the `user` and `revision`...

indico/modules/events/editing/notifications.py
```diff
@@ -0,0 +1,82 @@
...
+                              timeline_url=revision.editable.external_timeline_url,
+                              recipient_name=editor.first_name)
+    send_email(make_email(editor.email, template=tpl))
+
+
+def notify_submitter_confirmation(revision, action):
+    """
+    Notify the editor(s) about submitter accepting/rejecting revision changes.
+    """
+    submitter = revision.submitter
```
ThiefMaster, line 68: This may not actually be the user performing the action. Pass the `user` (coming from `session.user`) through to know who is actually doing the action.

PR #4459, opened 2020-05-19
Implement backend actions for multiple editables

### LINKED ISSUE #4418: Editable list: Backend actions
These are action that can be executed to selected editables, so they are all POST requests that take a list of editable ids.

- download all files - download a zip archive containing all the files, with a folder structure e.g. like this: "editable_type_name/code_or_id/file_type/file_name" (i think codes are not required to be unique so we may need to always include the ID as well even if there is a code, or we could do it only if there are duplicate codes)
- assign to me - set current user as editor
- assign to <user> (any editor for the given editable type) - separate endpoint than "to me" since normal editors can only assign to themselves
- unassign - remove the editor
- set status - ??? (doesn't seem to make much sense, since you cannot just set a state without taking the current state into account?)

mic4ael:
closes #4418

### DIFF
--- a/indico/modules/events/editing/blueprint.py
+++ b/indico/modules/events/editing/blueprint.py
@@ -8,7 +8,7 @@
 from __future__ import unicode_literals
 
 from indico.modules.events.editing.controllers import frontend
-from indico.modules.events.editing.controllers.backend import common, management, service, timeline
+from indico.modules.events.editing.controllers.backend import common, editable_list, management, service, timeline
 from indico.web.flask.wrappers import IndicoBlueprint
 
 
@@ -66,12 +66,22 @@ _bp.add_url_rule('/editing/api/service/connect', 'api_service_connect', service.
 _bp.add_url_rule('/editing/api/service/disconnect', 'api_service_disconnect', service.RHDisconnectService,
                  methods=('POST',))
 _bp.add_url_rule('/editing/api/service/status', 'api_service_status', service.RHServiceStatus)
+_bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/editables/prepare-archive', 'prepare_editables_archive',
+                 editable_list.RHPrepareEditablesArchive, methods=('POST',))
+_bp.add_url_rule('/editing/<any(paper,slides,poster):type>/editables/archive/<uuid:uuid>.zip', 'download_archive',
+                 editable_list.RHDownloadArchive)
+_bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/editables/assign', 'assign_editor',
+                 editable_list.RHAssignEditor, methods=('POST',))
+_bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/editables/assign/me', 'assign_myself',
+                 editable_list.RHAssignMyselfAsEditor, methods=('POST',))
+_bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/editables/unassign', 'unassign_editor',
+                 editable_list.RHUnassignEditor, methods=('POST',))
 
 # Contribution/revision-level APIs
 contrib_api_prefix = '/api' + contrib_prefix
-_bp.add_url_rule(contrib_api_prefix + '/upload', 'api_upload', timeline.RHEditingUploadFile, methods=('POST',))
 _bp.add_url_rule(contrib_api_prefix, 'api_editable', timeline.RHEditable)
 _bp.add_url_rule(contrib_api_prefix, 'api_create_editable', timeline.RHCreateEditable, methods=('PUT',))
+_bp.add_url_rule(contrib_api_prefix + '/upload', 'api_upload', timeline.RHEditingUploadFile, methods=('POST',))
 _bp.add_url_rule(contrib_api_prefix + '/<int:revision_id>/review', 'api_review_editable',
                  timeline.RHReviewEditable, methods=('POST',))
 _bp.add_url_rule(contrib_api_prefix + '/<int:revision_id>/confirm', 'api_confirm_changes',
--- a/indico/modules/events/editing/controllers/backend/editable_list.py
+++ b/indico/modules/events/editing/controllers/backend/editable_list.py
@@ -1,0 +1,79 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2020 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from __future__ import unicode_literals
+
+import uuid
+
+from flask import jsonify, request, session
+from werkzeug.exceptions import Forbidden
+
+from indico.legacy.common.cache import GenericCache
+from indico.modules.events.editing.controllers.base import RHEditablesBase, RHEditableTypeManagementBase
+from indico.modules.events.editing.models.editable import Editable
+from indico.modules.events.editing.operations import assign_editor, generate_editables_zip, unassign_editor
+from indico.util.i18n import _
+from indico.util.marshmallow import Principal
+from indico.web.args import use_kwargs
+from indico.web.flask.util import url_for
+
+
+archive_cache = GenericCache('editables-archive')
+
+
+class RHPrepareEditablesArchive(RHEditablesBase):
+    def _process(self):
+        key = unicode(uuid.uuid4())
+        data = [editable.id for editable in self.editables]
+        archive_cache.set(key, data, time=1800)
+        download_url = url_for('.download_archive', self.event, type=self.editable_type.name, uuid=key)
+        return jsonify(download_url=download_url)
+
+
+class RHDownloadArchive(RHEditableTypeManagementBase):
+    def _process(self):
+        editable_ids = archive_cache.get(unicode(request.view_args['uuid']), [])
+        editables = Editable.query.filter(Editable.id.in_(editable_ids)).all()
+        return generate_editables_zip(editables)
+
+
+class RHAssignEditor(RHEditablesBase):
+    @use_kwargs({
+        'editor': Principal(required=True)
+    })
+    def _process_args(self, editor):
+        RHEditablesBase._process_args(self)
+        if (not self.event.can_manage(editor, self.editable_type.editor_permission)
+                and not self.event.can_manage(editor, 'editing_manager')):
+            raise Forbidden(_('This user is not an editor of the {} type').format(self.editable_type.name))
+
+        self.editor = editor
+
+    def _process(self):
+        for editable in self.editables:
+            assign_editor(editable, self.editor)
+        return '', 204
+
+
+class RHAssignMyselfAsEditor(RHEditablesBase):
+    def _check_access(self):
+        RHEditablesBase._check_access(self)
+        if (not self.event.can_manage(session.user, self.editable_type.editor_permission)
+                and not self.event.can_manage(session.user, 'editing_manager')):
+            raise Forbidden(_('You are not an editor of the {} type').format(self.editable_type.name))
+
+    def _process(self):
+        for editable in self.editables:
+            assign_editor(editable, session.user)
+        return '', 204
+
+
+class RHUnassignEditor(RHEditablesBase):
+    def _process(self):
+        for editable in self.editables:
+            unassign_editor(editable)
+        return '', 204
--- a/indico/modules/events/editing/controllers/base.py
+++ b/indico/modules/events/editing/controllers/base.py
@@ -12,10 +12,13 @@ from werkzeug.exceptions import Unauthorized
 
 from indico.modules.events.contributions.controllers.display import RHContributionDisplayBase
 from indico.modules.events.controllers.base import RHDisplayEventBase
+from indico.modules.events.editing.fields import EditableList
 from indico.modules.events.editing.models.editable import Editable, EditableType
 from indico.modules.events.editing.settings import editing_settings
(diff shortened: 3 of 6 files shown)


indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -176,3 +179,33 @@ def _process_POST(self, principals):
...
+
+class RHDownloadArchive(RHEditablesBase):
+    def _process(self):
+        return generate_editables_zip(self.editables)
+
+
+class RHAssignEditor(RHEditablesBase):
+    def _process(self):
+        argmap = {
+            'editor': Principal(required=True)
```
ThiefMaster, line 192: You can use the `use_kwargs` decorator for this

indico/modules/events/editing/controllers/base.py
```diff
@@ -64,3 +67,13 @@ def _process_args(self):
...
                          .first())
+
+
+class RHEditablesBase(RHEditingManagementBase):
+    """Base class for operations on multiple editables."""
+
+    def _process_args(self):
+        RHEditingManagementBase._process_args(self)
+        self.editables = parser.parse({
+            'editables': EditableList(required=True, validate=not_empty)
```
ThiefMaster, line 78: should work with the decorator as well

indico/modules/events/editing/controllers/backend/management.py
mic4ael: Ah.. true.. I should have done that. Initially I was passing some redundant objects to `EditableList` and that is why I didn't use available decorators. But after some time I realized I don't need to pass anything like `self.event` or `self.contribution`.

ThiefMaster: Actually, you do! You need to ensure that you cannot pass editable ids which do not belong to the given event!

So, the decorator may not work after all unless you can do it with the context coming from the RH..

indico/modules/events/editing/controllers/base.py
ThiefMaster: Ah no, you need to ensure that you cannot pass editable ids which do not belong to the given event!

So, the decorator may not work after all unless you can do it with the context coming from the RH...

indico/modules/events/editing/blueprint.py
```diff
@@ -62,12 +62,17 @@
 _bp.add_url_rule('/editing/api/service/disconnect', 'api_service_disconnect', service.RHDisconnectService,
                  methods=('POST',))
 _bp.add_url_rule('/editing/api/service/status', 'api_service_status', service.RHServiceStatus)
+_bp.add_url_rule('/editing/api/editables/files.zip', 'download_archive', management.RHDownloadArchive)
```
ThiefMaster, line 65: I would not use a `.../api/...` URL for this one since it's something user-facing

indico/modules/events/editing/blueprint.py
```diff
@@ -62,12 +62,17 @@
...
+_bp.add_url_rule('/editing/api/editables/assign', 'assign_editor', management.RHAssignEditor, methods=('POST',))
+_bp.add_url_rule('/editing/api/editables/assign/me', 'assign_myself', management.RHAssignMyselfAsEditor,
+                 methods=('POST',))
+_bp.add_url_rule('/editing/api/editables/unassign', 'unassign_editor', management.RHUnassignEditor, methods=('POST',))
```
ThiefMaster, line 69: I think those should be within `/editing/api/<any(paper,slides,poster):type>/...`

indico/modules/events/editing/blueprint.py
mic4ael: I wasn't actually very sure about this. I didn't like the idea of having it as part of the `/api` but in the end it is the same action as any other exposed by the API

indico/modules/events/editing/blueprint.py
mic4ael: Ahh, right, we will have separate lists for each type. For some reason I thought we will have a list of all Editables.

indico/modules/events/editing/blueprint.py
ThiefMaster: is it? Other APIs are called in the background, but for this one we need to redirect the user to it.

Actually, maybe the best option would be to do it like when printing badges etc, i.e. you make an API call that stores the (properly validated) data in redis (via GenericCache) and then returns a UUID to use in a URL to redirect to which starts the download.

Otherwise you'd need to create a HTML form to be able to submit everything which isn't amazing...

indico/modules/events/editing/operations.py
```diff
@@ -283,3 +289,47 @@ def update_review_condition(condition, file_types):
...
+
+    buf.seek(0)
+    return send_file('files.zip', buf, 'application/zip', inline=False)
+
+
+def _compose_filepath(editable, revision_file):
+    file_obj = revision_file.file
+    contrib = editable.contribution
+    editable_type = editable.type.name
+    code = 'Editable-{}'.format(editable.id)
```
ThiefMaster, line 324: I don't think the internal ID makes sense here - I'd use the friendly id of the contribution instead.

indico/modules/events/editing/blueprint.py
```diff
@@ -62,12 +62,20 @@
 _bp.add_url_rule('/editing/api/service/disconnect', 'api_service_disconnect', service.RHDisconnectService,
                  methods=('POST',))
 _bp.add_url_rule('/editing/api/service/status', 'api_service_status', service.RHServiceStatus)
+_bp.add_url_rule('/editing/<any(paper,slides,poster):type>/editables/files.zip', 'download_archive',
+                 editable_list.RHDownloadArchive)
```
ThiefMaster, line 66: I don't think a GET request will work well for this unless you select only few editables. It will also result in a really ugly URL. What about my suggestion of storing the IDs in redis with a token (with a POST request), and then returning only the URL so the client-side code can simply redirect to the download URL? I.e. similar to what we do with the `badge_cache` when printing badges.

indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class RHDownloadArchive(RHEditablesBase):
+    def _process(self):
+        return generate_editables_zip(self.editables)
+
+
+class RHAssignEditor(RHEditablesBase):
+    @use_kwargs({
+        'editor': Principal(required=True)
```
ThiefMaster, line 25: Check that the selected user is actually an editor for the given editable type.

indico/modules/events/editing/operations.py
```diff
@@ -283,3 +289,47 @@ def update_review_condition(condition, file_types):
 def delete_review_condition(condition):
     logger.info('Review condition %r deleted by %r', condition, session.user)
     db.session.delete(condition)
+
+
+def assign_editor(editable, editor):
+    editable.editor = editor
+    db.session.flush()
+    logger.info('Editable %r assigned to %r editor', editable, editor)
```
ThiefMaster, line 297: `... to editor %r`

indico/modules/events/editing/operations.py
```diff
@@ -283,3 +289,47 @@ def update_review_condition(condition, file_types):
...
+    logger.info('Editor %r has been unassigned from %r', editor, editable)
+
+
+def generate_editables_zip(editables):
+    buf = BytesIO()
+    with ZipFile(buf, 'w', allowZip64=True) as zip_file:
+        for editable in editables:
+            for revision_file in editable.revisions[-1].files:
+                file_obj = revision_file.file
+                with file_obj.storage.get_local_path(file_obj.storage_file_id) as src_file:
```
ThiefMaster, line 313: `with file_obj.get_local_path() as src_file` (let's fix this in `RHExportRevisionFiles` and `ZipGeneratorMixin` too)

indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,60 @@
...
+        for editable in self.editables:
+            assign_editor(editable, self.editor)
+        return '', 204
+
+
+class RHAssignMyselfAsEditor(RHEditablesBase):
+    def _check_access(self):
+        RHEditablesBase._check_access(self)
+        if not self.event.can_manage(session.user, self.editable_type.editor_permission, explicit_permission=True):
+            raise Forbidden(_('You are not an editor of the {} type').format(self.editable_type.name))
```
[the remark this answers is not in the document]
mic4ael: you mean without `explicit_permission`?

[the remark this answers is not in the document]
ThiefMaster: `self.event.can_manage(session.user, permission='editing_manager') or self.event.can_manage(session.user, self.editable_type.editor_permission, explicit_permission=True)`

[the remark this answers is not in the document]
mic4ael: > but it also needs the check for more general management permissions mentioned above.

I had this part in mind

indico/modules/events/editing/blueprint.py
mic4ael: Why not something like `RHContributionsMaterialPackage`? I would just have to use `POST` in this case and let the frontend handle it properly.

ThiefMaster: But there we use classic forms that just do a normal submit, while editing is all in react... So when doing it like in `RHContributionsMaterialPackage` the client code would need to get the binary data from the AJAX request and use the Blob API to start a download... vs just redirecting to the url starting the download. And from what I can see `URL.createObjectURL()` still doesn't let you specify a file name shown in the download prompt :/

indico/modules/events/editing/blueprint.py
```diff
@@ -62,12 +62,22 @@
 _bp.add_url_rule('/editing/api/service/disconnect', 'api_service_disconnect', service.RHDisconnectService,
                  methods=('POST',))
 _bp.add_url_rule('/editing/api/service/status', 'api_service_status', service.RHServiceStatus)
+_bp.add_url_rule('/editing/<any(paper,slides,poster):type>/editables/prepare-archive', 'prepare_editables_archive',
+                 editable_list.RHPrepareEditablesArchive, methods=('POST',))
+_bp.add_url_rule('/editing/<any(paper,slides,poster):type>/editables/archive/<uuid:uuid>', 'download_archive',
```
ThiefMaster, line 67: `.zip` at the end of the url? ;)

indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,78 @@
...
+        return generate_editables_zip(editables)
+
+
+class RHAssignEditor(RHEditablesBase):
+    @use_kwargs({
+        'editor': Principal(required=True)
+    })
+    def _process_args(self, editor):
+        RHEditablesBase._process_args(self)
+        if not self.event.can_manage(self.editor, self.editable_type.editor_permission):
```
ThiefMaster, line 51: This does not take `editing_manager` into account (only full event managers).

indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,78 @@
...
+    def _process(self):
+        for editable in self.editables:
+            assign_editor(editable, self.editor)
+        return '', 204
+
+
+class RHAssignMyselfAsEditor(RHEditablesBase):
+    def _check_access(self):
+        RHEditablesBase._check_access(self)
+        if not self.event.can_manage(session.user, self.editable_type.editor_permission):
```
ThiefMaster, line 65: same here

indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,80 @@
...
+archive_cache = GenericCache('editables-archive')
+
+
+class RHPrepareEditablesArchive(RHEditablesBase):
+    def _process(self):
+        key = unicode(uuid.uuid4())
+        data = [editable.id for editable in self.editables]
+        archive_cache.set(key, data, time=1800)
+        download_url = url_for('.download_archive', self.event, type=self.editable_type.name, uuid=key)
+        return jsonify_data(flash=False, download_url=download_url)
```
ThiefMaster, line 35: `jsonify(download_url=download_url)` (`jsonify_data` is only needed for the old jquery-based stuff that expects a `success: true` field in the json)

indico/modules/events/editing/operations.py
```diff
@@ -283,3 +289,47 @@ def update_review_condition(condition, file_types):
 def delete_review_condition(condition):
     logger.info('Review condition %r deleted by %r', condition, session.user)
     db.session.delete(condition)
+
+
+def assign_editor(editable, editor):
+    editable.editor = editor
+    db.session.flush()
+    logger.info('Editable %r assigned to editor %r', editable, editor)
```
ThiefMaster, line 297: Shouldn't this also be logged to the event log? Might be useful for event managers to see who performed an assignment and when...

indico/modules/events/editing/operations.py
```diff
@@ -283,3 +290,49 @@ def update_review_condition(condition, file_types):
...
+    editable.event.log(EventLogRealm.management, EventLogKind.change, 'Editing',
+                       'Editable assigned to an editor', session.user, data={'Editor': editor.full_name})
+    db.session.flush()
+
+
+def unassign_editor(editable):
+    editor = editable.editor
+    editable.editor = None
+    db.session.flush()
+    logger.info('Editor %r has been unassigned from %r', editor, editable)
```
ThiefMaster, line 307: I'd log this to the event log as well...

PR #4460, opened 2020-05-19
Add backend for list of editables

### LINKED ISSUE #4416: List of editables: List backend
The most important functionality here is to have an endpoint returning a list of editables with basic information. Feel free to return dummy data for things that are more expensive to get (like the current state / bullet color that require diving into the revisions of the editable).

Also, for performance reasons we need simple pagination, i.e. being able to specify:

- the max number of editables to load
- what to sort by (e.g. ID, title - not sure if anything else makes sense); ID should always be used as a last criterion to make the sorting stable in case of duplicate titles
- before/after id - needed when loading a later/earlier page; if one (both = invalid) is present, you'd only load items with an ID less/greater than what's specified. we may not need "before" at all unless we ditch virtualized rendering in favor of explizit prev/next navigation, or we don't always keep already-loaded data.

ParthS007:
closes #4416 
Screenshot from 2020-06-09 16-19-30

### DIFF
--- a/indico/modules/events/contributions/models/contributions.py
+++ b/indico/modules/events/contributions/models/contributions.py
@@ -564,6 +564,10 @@ class Contribution(DescriptionMixin, ProtectionManagersMixin, LocationMixin, Att
         # Otherwise only users with submission rights are authorized
         return self.can_manage(user, 'submit', allow_admin=False, check_parent=False)
 
+    def get_editable(self, editable_type):
+        """Get the editable of the given type."""
+        return next((e for e in self.editables if e.type == editable_type), None)
+
 
 Contribution.register_protection_events()
 
--- a/indico/modules/events/editing/blueprint.py
+++ b/indico/modules/events/editing/blueprint.py
@@ -45,6 +45,8 @@ _bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/file-types', 'api
                  common.RHEditingFileTypes)
 _bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/file-types', 'api_add_file_type',
                  management.RHCreateFileType, methods=('POST',))
+_bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/list', 'api_editable_list',
+                 editable_list.RHEditableList)
 _bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/file-types/<int:file_type_id>', 'api_edit_file_type',
                  management.RHEditFileType, methods=('PATCH', 'DELETE'))
 _bp.add_url_rule('/editing/api/tags', 'api_tags', common.RHEditingTags)
--- a/indico/modules/events/editing/client/js/editing/timeline/StateIndicator.jsx
+++ b/indico/modules/events/editing/client/js/editing/timeline/StateIndicator.jsx
@@ -20,6 +20,8 @@ const colors = {
   accepted: 'green',
   assigned: 'purple',
   needs_submitter_changes: 'orange',
+  not_submitted: 'blue',
+  ready_for_review: 'olive',
 };
 
 const labelColors = {
@@ -29,6 +31,8 @@ const labelColors = {
   accepted: Palette.green,
   assigned: Palette.purple,
   needs_submitter_changes: Palette.orange,
+  not_submitted: Palette.blue,
+  ready_for_review: Palette.olive,
 };
 
 export default function StateIndicator({label, circular, tooltip, state}) {
--- a/indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
+++ b/indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
@@ -6,11 +6,12 @@
 // LICENSE file for more details.
 
 import dashboardURL from 'indico-url:event_editing.dashboard';
-// import manageEditableTypeListURL from 'indico-url:event_editing.manage_editable_type_list'; Uncomment when we have backend ready
+import editableListURL from 'indico-url:event_editing.api_editable_list';
 
 import React, {useState} from 'react';
+import PropTypes from 'prop-types';
 import {useParams} from 'react-router-dom';
-import {Button, Icon, Checkbox, Message, Search} from 'semantic-ui-react';
+import {Button, Icon, Loader, Checkbox, Message, Search} from 'semantic-ui-react';
 import {Column, Table, SortDirection, WindowScroller} from 'react-virtualized';
 import _ from 'lodash';
 import {
@@ -20,36 +21,75 @@ import {
 } from 'indico/react/components';
 import {useNumericParam} from 'indico/react/util/routing';
 import {Translate} from 'indico/react/i18n';
+import {useIndicoAxios} from 'indico/react/hooks';
+import {userPropTypes} from '../../editing/timeline/util';
 import {EditableType} from '../../models';
 import StateIndicator from '../../editing/timeline/StateIndicator';
 
 import './EditableList.module.scss';
 
 export default function EditableList() {
-  const list = []; // Replace it with list from server
   const eventId = useNumericParam('confId');
   const {type} = useParams();
-  const [sortBy, setSortBy] = useState('id');
+  const {data, loading: isLoadingEditableList, lastData} = useIndicoAxios({
+    url: editableListURL({confId: eventId, type}),
+    camelize: true,
+    trigger: eventId,
+  });
+  const editableList = data || lastData;
+  if (isLoadingEditableList && !lastData) {
+    return <Loader inline="centered" active />;
+  } else if (!editableList) {
+    return null;
+  }
+  const codePresent = Object.values(editableList).some(c => c.code);
+  return (
+    <EditableListDisplay
+      editableList={editableList}
+      codePresent={codePresent}
+      editableType={type}
+      eventId={eventId}
+    />
+  );
+}
+
+function EditableListDisplay({editableList, codePresent, editableType, eventId}) {
+  const [sortBy, setSortBy] = useState('friendly_id');
   const [sortDirection, setSortDirection] = useState('ASC');
-  const [sortedList, setSortedList] = useState(list); // list from server
+  const [sortedList, setSortedList] = useState(editableList);
   const [checked, setChecked] = useState([]);
   const hasCheckedRows = checked.length === 0;
   const title = {
     [EditableType.paper]: Translate.string('List of papers'),
     [EditableType.slides]: Translate.string('List of slides'),
     [EditableType.poster]: Translate.string('List of posters'),
-  }[type];
+  }[editableType];
   const columnHeaders = [
-    ['id', Translate.string('ID'), 60],
-    ['code', Translate.string('Code'), 80],
+    ['friendlyId', Translate.string('ID'), 60],
+    ...(codePresent ? [['code', Translate.string('Code'), 80]] : []),
     ['title', Translate.string('Title'), 600],
-    ['status', Translate.string('Status'), 150],
-    ['assigned', Translate.string('Assigned'), 400],
+    ['status', Translate.string('Status'), 200],
+    ['editor', Translate.string('Editor'), 400],
   ];
+
+  const sortEditor = contribution =>
+    contribution.editable &&
+    contribution.editable.editor &&
+    contribution.editable.editor.fullName.toLowerCase();
+  const sortStatus = contribution =>
+    contribution.editable && contribution.editable.state.toLowerCase();
+  const sortTitle = contribution => contribution.title.toLowerCase();
+
+  const sortFuncs = {
+    title: sortTitle,
+    status: sortStatus,
+    editor: sortEditor,
+  };
 
   // eslint-disable-next-line no-shadow
   const _sortList = ({sortBy, sortDirection}) => {
-    const newList = _.sortBy(list, [sortBy]);
+    const fn = sortFuncs[sortBy] || (x => x);
+    const newList = _.sortBy(editableList, fn);
     if (sortDirection === SortDirection.DESC) {
       newList.reverse();
     }
@@ -63,17 +103,51 @@ export default function EditableList() {
     setSortedList(_sortList({sortBy, sortDirection}));
   };
 
+  const renderCode = code => code || Translate.string('n/a');
+  const renderEditor = editable => {
+    return editable && editable.editor ? (
+      <div>
+        <Icon name="user outline" />
+        {editable.editor.fullName}
+      </div>
+    ) : (
+      <div />
+    );
+  };
+  // eslint-disable-next-line no-shadow
+  const renderTitle = (title, index) => {
+    return sortedList[index].editable ? (
+      <a href={sortedList[index].editable.timelineURL}>{title}</a>
+    ) : (
+      <div>{title}</div>
+    );
+  };
+  const renderStatus = editable => {
+    return <StateIndicator state={editable ? editable.state : 'not_submitted'} circular label />;
+  };
+  const renderFuncs = {
+    title: renderTitle,
+    code: renderCode,
+    status: renderStatus,
+    editor: renderEditor,
+  };
+
   // eslint-disable-next-line react/prop-types
   const renderCell = ({dataKey, rowIndex}) => {
+    let data = sortedList[rowIndex][dataKey];
+    if (dataKey === 'status' || dataKey === 'editor') {
+      data = sortedList[rowIndex]['editable'];
+    }
+    const fn = renderFuncs[dataKey] || (x => x);
     return (
       <TooltipIfTruncated>
-        <div index={rowIndex} styleName="rowcolumn-tooltip" role="gridcell">
-          {dataKey === 'assigned' && <Icon name="user outline" />}
-          {dataKey === 'status' ? (
-            <StateIndicator index={rowIndex} state={sortedList[rowIndex][dataKey]} circular label />
-          ) : (
-            sortedList[rowIndex][dataKey]
-          )}
+        <div
+          index={rowIndex}
+          styleName="rowcolumn-tooltip"
+          role="gridcell"
+          style={dataKey === 'title' ? {display: 'block'} : {}}
+        >
+          {fn(data, rowIndex)}
         </div>
       </TooltipIfTruncated>
     );
@@ -112,7 +186,7 @@ export default function EditableList() {
           />
           <Button disabled={hasCheckedRows} content={Translate.string('Download all files')} />
         </div>
-        <Search />
+        <Search disabled={!sortedList.length} />
       </div>
       {sortedList.length ? (
         <div styleName="editable-list">
@@ -175,3 +249,24 @@ export default function EditableList() {
     </>
   );
 }
+
+EditableListDisplay.propTypes = {
+  editableList: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.number.isRequired,
+      friendlyId: PropTypes.number.isRequired,
+      title: PropTypes.string.isRequired,
+      code: PropTypes.string.isRequired,
+      editable: PropTypes.shape({
+        id: PropTypes.number,
+        editor: PropTypes.shape(userPropTypes),
+        state: PropTypes.string,
+        type: PropTypes.oneOf(Object.values(EditableType)),
+        timelineURL: PropTypes.string,
+      }),
+    })
+  ).isRequired,
+  codePresent: PropTypes.bool.isRequired,
+  editableType: PropTypes.oneOf(Object.values(EditableType)).isRequired,
+  eventId: PropTypes.number.isRequired,
+};
--- a/indico/modules/events/editing/controllers/backend/editable_list.py
+++ b/indico/modules/events/editing/controllers/backend/editable_list.py
@@ -10,12 +10,15 @@ from __future__ import unicode_literals
 import uuid
 
 from flask import jsonify, request, session
+from sqlalchemy.orm import joinedload
 from werkzeug.exceptions import Forbidden
 
 from indico.legacy.common.cache import GenericCache
+from indico.modules.events.contributions.models.contributions import Contribution
 from indico.modules.events.editing.controllers.base import RHEditablesBase, RHEditableTypeManagementBase
 from indico.modules.events.editing.models.editable import Editable
 from indico.modules.events.editing.operations import assign_editor, generate_editables_zip, unassign_editor
+from indico.modules.events.editing.schemas import EditingEditableListSchema
 from indico.util.i18n import _
 from indico.util.marshmallow import Principal
 from indico.web.args import use_kwargs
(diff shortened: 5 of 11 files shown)


indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,57 @@
...
+from indico.modules.events.contributions.models.contributions import Contribution
+from indico.modules.events.editing.controllers.base import RHEditingBase
+from indico.modules.events.editing.models.editable import EditableType
+
+
+class RHEditableList(RHEditingBase):
+    """Return editable list of the event for the editable type."""
+    def _process_args(self):
+        RHEditingBase._process_args(self)
+        self.event = Event.get(int(request.view_args['confId']))
```
mic4ael, line 22: Not needed since one of the base classes already does that

indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,57 @@
...
+        self.editable_type = EditableType[request.view_args['type']]
```
mic4ael, line 23: I guess you can inherit from `RHEditableTypeManagementBase` to have `self.editable_type` as well

indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,57 @@
...
+        self.contributions = Contribution.query.with_parent(self.event)
+        self.is_code_present = False
+        self.serialized_contributions = []
+        for contribution in self.contributions:
+            serialized_editable = None
+            if contribution.editables:
```
mic4ael, line 29: no need for that `if` since if `contribution.editables` is empty it will not do anything anyway ;) (and we will save some indentation)

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -153,13 +189,21 @@ export default function EditableList() {
                     />
                   )}
                 />
-                {columnHeaders.map(([key, label, width, extraProps]) => (
+                {code && (
+                  <Column
+                    key="code"
+                    dataKey="code"
+                    label="Code"
```
mic4ael, line 196: missing i18n

indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,57 @@
...
+                for editable in contribution.editables:
+                    if editable.type.name == self.editable_type.name:
+                        serialized_editable = self.serialize_editable(editable)
+            self.serialized_contributions.append(self.serialize_contribution(contribution, serialized_editable))
```
mic4ael, line 33: I am not sure if that is the best place to serialize the contributions. Usually we "deserialize" the request arguments to a Pythonic form in `_process_args` which then can be serialized inside RH (what you are doing with `self.contributions`). Maybe to keep the `process_args` small and have only one responsibility you should move the serialization to a separate method which then would be called from `_process`.

ThiefMaster: indeed, this is way too much logic for `_process_args` - this all belongs in `_process`. also, i think we could make use of a marshmallow schema

ParthS007: Do we need schema here?

ThiefMaster: I think using one would certainly be cleaner.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -153,13 +189,21 @@ export default function EditableList() {
                     />
                   )}
                 />
-                {columnHeaders.map(([key, label, width, extraProps]) => (
+                {code && (
```
panagiotappl, line 192: Maybe you could add this as a `columnHeader` conditionally somehow?
I couldn't find any pretty solution though...
Here's an example:
``` jsx
const columnHeaders = [
    ['id', Translate.string('ID'), 60],
    ...code ? [['code', Translate.string('Code'), 80]] : [],
    ['title', Translate.string('Title'), 600],
    ['status', Translate.string('Status'), 200],
    ['editor', Translate.string('Editor'), 400],
  ];
```

indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,57 @@
...
+from flask import jsonify, request
+
+from indico.modules.events import Event
+from indico.modules.events.contributions.models.contributions import Contribution
+from indico.modules.events.editing.controllers.base import RHEditingBase
+from indico.modules.events.editing.models.editable import EditableType
+
+
+class RHEditableList(RHEditingBase):
+    """Return editable list of the event for the editable type."""
```
panagiotappl, line 19: Maybe you could rephrase this to "Return the list of editables of the event for a given type"

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ParthS007: It was looking a bit ugly but I can go ahead with this if everyone is fine with this. :+1:

indico/modules/events/editing/client/js/editing/timeline/StateIndicator.jsx
```diff
@@ -20,6 +20,8 @@ const colors = {
   accepted: 'green',
   assigned: 'purple',
   needs_submitter_changes: 'orange',
+  not_submitted: 'blue',
+  ready_for_review: 'olive',
```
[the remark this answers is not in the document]
ParthS007: What about brown or grey?

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -66,16 +89,29 @@ export default function EditableList() {
...
-        <div index={rowIndex} styleName="rowcolumn-tooltip" role="gridcell">
-          {dataKey === 'assigned' && <Icon name="user outline" />}
-          {dataKey === 'status' ? (
-            <StateIndicator index={rowIndex} state={sortedList[rowIndex][dataKey]} circular label />
-          ) : (
-            sortedList[rowIndex][dataKey]
-          )}
-        </div>
-      </TooltipIfTruncated>
+      sortedList[rowIndex][dataKey] !== null && (
```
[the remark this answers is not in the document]
ParthS007: I checked with zero code, it's working fine. Going ahead with this. :+1:

indico/modules/events/editing/schemas.py
```diff
@@ -139,6 +140,29 @@ class Meta:
     review_conditions_valid = fields.Boolean()
 
 
+class EditableBasicSchema(mm.ModelSchema):
+    class Meta:
+        model = Editable
+        fields = ('id', 'type', 'state', 'editor')
+
+    state = fields.String()
+    editor = fields.Nested(UserSchema, only=('id', 'avatar_bg_color', 'full_name'))
```
ParthS007, line 149: Currently, Assigning of editor is not implemented so it will give error.

indico/modules/events/editing/schemas.py
```diff
@@ -139,6 +140,29 @@ class Meta:
...
+class EditingEditableListSchema(mm.ModelSchema):
+    class Meta:
+        model = Contribution
+        rh_context = ('editable_type')
+        fields = ('id', 'friendly_id', 'title', 'code', 'editable')
+
+    editable = fields.Method('_get_editable')
+
+    def _get_editable(self, contribution):
+        editable_type = self.context['editable_type']
```
ParthS007, line 161: Why `editable_type` is not available in `self.context`?

ThiefMaster: Look carefully at this line again :)

https://github.com/indico/indico/pull/4460/files#diff-14a5c7d2bef671ab20ad3825f8c7d38aR155

ParthS007: No, that is a different error. Fixed that!

ThiefMaster: The `rh_context` not being a tuple would explain why you don't have the `editable_type` in the context though...

indico/modules/events/editing/schemas.py
```diff
@@ -139,6 +140,29 @@ class Meta:
...
+        editable = {editable for editable in contribution.editables if editable.type.name == editable_type.name}
```
ThiefMaster, line 162: why are you comparing by name here?

ParthS007: Comparing whether it is a `paper`, `slides` or `poster`.

indico/modules/events/editing/schemas.py
ThiefMaster: Why does it give an error? If there is no editor assigned it should serialize as `None`

indico/modules/events/editing/schemas.py
ParthS007: If I did `rh_context = ('editable_type',)`, still gives error.

indico/modules/events/editing/schemas.py
ThiefMaster: What I mean is, why not just compare `editable.type == editable_type`

indico/modules/events/editing/schemas.py
ThiefMaster: time to add some print statements directly in `use_rh_args` to debug what's going wrong :)

indico/modules/events/editing/schemas.py
ParthS007: Resolved!

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -20,36 +21,67 @@ import {
...
+  });
+
+  let editableList = data || lastData;
+  if (isLoadingEditableList && !lastData) {
+    return <Loader inline="centered" active />;
+  } else if (!editableList) {
+    return null;
+  }
+
+  editableList = [...data];
```
mic4ael, line 46: What is the point of this?

indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,28 @@
...
+    """Return the list of editables of the event for a given type"""
+    def _process_args(self):
+        RHEditableTypeManagementBase._process_args(self)
+        self.contributions = (Contribution.query
+                              .with_parent(self.event)
+                              .options(joinedload('editables'))
+                              .all())
+
+    def _process(self):
+        return (EditingEditableListSchema(many=True, context={"editable_type": self.editable_type})
```
mic4ael, line 27:
```suggestion
        return (EditingEditableListSchema(many=True, context={'editable_type': self.editable_type})
```

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -20,36 +21,67 @@ import {
...
+  editableList.forEach(contribution => {
+    if (_.isEmpty(contribution.editable)) {
+      contribution.status = 'not_submitted';
+      contribution.editor = null;
+    } else {
+      contribution.status = contribution.editable.state;
+      contribution.editor = contribution.editable.editor.name;
+    }
+    if (!contribution.code) {
+      contribution.code = 'N/A';
```
mic4ael, line 56: Wouldn't it be better to actually move this to `renderCell` instead of modifying the object directly?

```javascript
{dataKey === 'code' && !code && Translate.string('N/A')}
```

indico/modules/events/editing/schemas.py
```diff
@@ -139,6 +140,28 @@ class Meta:
...
+    class Meta:
+        model = Contribution
+        fields = ('id', 'friendly_id', 'title', 'code', 'editable')
+
+    editable = fields.Method('_get_editable')
+
+    def _get_editable(self, contribution):
+        editable_type = self.context['editable_type']
+        editable = {editable for editable in contribution.editables if editable.type == editable_type}
+        return EditableBasicSchema().dump(editable)
```
mic4ael, line 162: This will silently error, since you are passing a `set` to the `dump` method which expects a single object.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ParthS007: From the backend, data is coming in multiple json objects and I just destructure them to store in a list.

ThiefMaster: But this line of code just shallow-copies a list...

mic4ael: Yeah, you are just copying the array.  It is like a noop in this case

ParthS007: oh okay, I got it. Thanks for the catch :+1:

indico/modules/events/editing/schemas.py
ParthS007: What should be the correct way to do it since editable will always be a single object?

ThiefMaster: add a `get_editable(self, editable_type)` method to `Contribution` which contains e.g. this:

```python
return next((e for e in self.editables if e.type == editable_type), None)
```

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ParthS007: In `renderCell`, I am checking `sortedList[rowIndex][dataKey]` already, so I thought of adding info in the object itself.

mic4ael: But it is a bit dirty. In case you e.g. access `contribution.code` later in the code, you would get `N/A` instead of null|undefined. IMO this shouldn't modify the data at all.

ThiefMaster: Yes, please don't mutate data (unless it's something known to be regenerated each render).

ParthS007: Considering this, should I remove the addition of `status` and `editor` attribute in object and add it in `renderCell` instead?

ThiefMaster: yes, this is also mutating the object

indico/modules/events/editing/client/js/editing/timeline/StateIndicator.jsx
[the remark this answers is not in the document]
panagiotappl: I don't know. I can't think of anything better, so if no one has a different suggestion you can leave it as it is.

indico/modules/events/editing/blueprint.py
```diff
@@ -45,6 +45,8 @@
                  common.RHEditingFileTypes)
 _bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/file-types', 'api_add_file_type',
                  management.RHCreateFileType, methods=('POST',))
+_bp.add_url_rule('/editing/api/<any(paper,slides,poster):type>/list', 'api_editable_list',
+                 editable_list.RHEditableList, methods=('GET',))
```
ThiefMaster, line 49: GET is the default, no need to specify `methods` at all

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -20,36 +21,65 @@ import {
...
+  });
+
+  const editableList = data || lastData;
+  if (isLoadingEditableList && !lastData) {
+    return <Loader inline="centered" active />;
+  } else if (!editableList) {
+    return null;
+  }
+  editableList.forEach(contribution => {
+    if (_.isEmpty(contribution.editable)) {
```
ThiefMaster, line 46: `contribution.editable === null` should be enough here

indico/modules/events/editing/schemas.py
```diff
@@ -139,6 +140,28 @@ class Meta:
...
+
+class EditingEditableListSchema(mm.ModelSchema):
+    class Meta:
+        model = Contribution
+        fields = ('id', 'friendly_id', 'title', 'code', 'editable')
+
+    editable = fields.Method('_get_editable')
+
+    def _get_editable(self, contribution):
+        editable_type = self.context["editable_type"]
```
ThiefMaster, line 160: single quotes

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ParthS007: I have removed this, can you check again?

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -63,17 +80,28 @@ export default function EditableList() {
...
+          role="gridcell"
+          style={dataKey === 'title' ? {display: 'block'} : {}}
+        >
+          {dataKey === 'friendlyId' && row[dataKey]}
+          {dataKey === 'code' && code(row[dataKey])}
+          {dataKey === 'title' && row[dataKey]}
+          {dataKey === 'status' && (
+            <StateIndicator index={rowIndex} state={status(row['editable'])} circular label />
           )}
+          {dataKey === 'editor' && <Icon name="user outline" /> && row['editable'][dataKey]}
```
ThiefMaster, line 104: What about a mapping object here?

```js
const renderFuncs = {
  code: renderCode,
  status: renderStatus,
  editor: renderEditor,
};

const data = sortedList[rowIndex][dataKey];
const fn = renderFuncs[dataKey] || (x => x);
```

And then here you can just do `{fn(data)}`

indico/modules/events/editing/controllers/backend/editable_list.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class RHEditableList(RHEditableTypeManagementBase):
+    """Return the list of editables of the event for a given type"""
+    def _process_args(self):
+        RHEditableTypeManagementBase._process_args(self)
+        self.contributions = (Contribution.query
+                              .with_parent(self.event)
+                              .options(joinedload('editables'))
+                              .all())
```
ThiefMaster, line 24: We should probably apply a default order here to have something consistent even if the user does not apply any order manually e.g. `.order_by(Contribution.friendly_id)`

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -63,17 +80,28 @@ export default function EditableList() {
     setSortedList(_sortList({sortBy, sortDirection}));
   };
 
+  const code = data => (!data ? Translate.string('N/A') : data);
+  const status = data => (_.isEmpty(data) ? 'not_submitted' : data.state);
```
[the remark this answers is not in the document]
ParthS007: Yes, It's an empty object.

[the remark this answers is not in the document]
ThiefMaster: That's strange. I'd expect `Nested` to give you `null` if the relationship is is `None`...

indico/modules/events/contributions/models/contributions.py
```diff
@@ -564,6 +564,10 @@ def can_submit_proceedings(self, user):
         # Otherwise only users with submission rights are authorized
         return self.can_manage(user, 'submit', allow_admin=False, check_parent=False)
 
+    def get_editable(self, editable_type):
+        """Get the editable of given editable type."""
```
ThiefMaster, line 568: > Get the editable of the given type.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -175,3 +206,9 @@ export default function EditableList() {
     </>
   );
 }
+
+EditableListDisplay.propTypes = {
+  editableList: PropTypes.array.isRequired,
+  editableType: PropTypes.string.isRequired,
+  eventId: PropTypes.any.isRequired,
```
ThiefMaster, line 213: why `any`? this should always be a `number`..

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -175,3 +206,9 @@ export default function EditableList() {
     </>
   );
 }
+
+EditableListDisplay.propTypes = {
+  editableList: PropTypes.array.isRequired,
+  editableType: PropTypes.string.isRequired,
```
ThiefMaster, line 212: `editableType: PropTypes.oneOf(Object.values(EditableType)).isRequired,`

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -175,3 +206,9 @@ export default function EditableList() {
     </>
   );
 }
+
+EditableListDisplay.propTypes = {
+  editableList: PropTypes.array.isRequired,
```
ThiefMaster, line 211: `PropTypes.arrayOf(PropTypes.shape(...)).isRequired` (in the shape specify the props you are actually using).

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ParthS007: Ohh, I forgot to change this.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -63,17 +81,42 @@ export default function EditableList() {
     setSortedList(_sortList({sortBy, sortDirection}));
   };
 
+  const renderCode = code => code || Translate.string('n/a');
+  const renderEditor = editable => {
+    !_.isEmpty(editable) ?? (<Icon name="user outline" /> && editable.editor.fullName);
```
ThiefMaster, line 86: This line looks really weird. Why do we need null-coalescing here (does our babel config transpile this properly?), and the `&&` seems strange since the left side is never falsy...

ParthS007: null-coalescing worked here but the the editor name was not visible because of `&&`. Changes that.

ThiefMaster: but why would `_.isEmpty(editable)` ever return null?

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -20,36 +21,80 @@ import {
...
+  } else if (!editableList) {
+    return null;
+  }
+
+  let codePresent = false;
+  for (const contribution of editableList) {
+    if (contribution.code !== '') {
+      codePresent = true;
+    }
+  }
```
ThiefMaster, line 51:
```js
const codePresent = Object.values(editableList).some(c => c.code);
```

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -63,17 +103,52 @@ export default function EditableList() {
...
       <TooltipIfTruncated>
-        <div index={rowIndex} styleName="rowcolumn-tooltip" role="gridcell">
-          {dataKey === 'assigned' && <Icon name="user outline" />}
-          {dataKey === 'status' ? (
-            <StateIndicator index={rowIndex} state={sortedList[rowIndex][dataKey]} circular label />
-          ) : (
-            sortedList[rowIndex][dataKey]
-          )}
+        <div
+          id={dataKey}
```
ThiefMaster, line 145: is this ID needed for anything? it's adding tons of duplicate IDs..

ParthS007: I don't think so. Removing it

PR #4466, opened 2020-05-24
Preserve non-ASCII chars in file names

### LINKED ISSUE #4465: Downloading attachments have wrong filenames (Unicode)
**Describe the bug**
Files uploaded with the material editor do not preserve their filenames when downloaded, if they include (non-ASCII) unicode characters.

**To Reproduce**
1. Go to 'Material Editor'
2. Upload file with filename (for example) 'TEST日本語123.txt'
3. Go to the event page, right-click and download the file above.
4. The downloaded file has a filename 'TEST123.txt', which is different from the original.

**Expected behavior**
In the example above, the downloaded filename should be 'TEST日本語123.txt' instead of 'TEST123.txt'.

**Screenshots**
Here is a screenshot of how the file looks in the agenda page (it is correct here):

unicode-filename

This shows that, at least in this intermediate step, the filename seems to be properly stored.
However, the actual URL to the file (and therefore how the web browser saves the file) does not have the correct filename.

Here is a screenshot of the web browser pointing to the file:

unicode-filename2

**Additional context**
* This behavior was seen at http://indico.cern.ch/
* Uploading seems to be working fine, since the internal representation shows the correct filename.
* It seems the generation of the download link alters the filename.
* It seems the non-ASCII characters get omitted.

ThiefMaster:
On disk filenames are still cleaned up and converted to ASCII, but the filename used for display purposes and the content-disposition header sent to clients now uses the original name.

closes #4465

### DIFF
--- a/indico/modules/events/papers/operations.py
+++ b/indico/modules/events/papers/operations.py
@@ -34,7 +34,7 @@ from indico.modules.events.papers.settings import PaperReviewingRole, paper_revi
 from indico.modules.events.util import update_object_principals
 from indico.modules.users import User
 from indico.util.date_time import now_utc
-from indico.util.fs import secure_filename
+from indico.util.fs import secure_client_filename
 from indico.util.i18n import orig_string
 
 
@@ -143,7 +143,7 @@ def close_cfp(event):
 def create_paper_revision(paper, submitter, files):
     revision = PaperRevision(paper=paper, submitter=submitter)
     for f in files:
-        filename = secure_filename(f.filename, 'paper')
+        filename = secure_client_filename(f.filename)
         content_type = mimetypes.guess_type(f.filename)[0] or f.mimetype or 'application/octet-stream'
         pf = PaperFile(filename=filename, content_type=content_type, paper_revision=revision,
                        _contribution=paper.contribution)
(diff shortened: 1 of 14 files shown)


indico/modules/events/papers/operations.py
```diff
@@ -189,13 +187,12 @@ def reset_paper_state(paper):
 
 def _store_paper_template_file(template, file):
     content_type = mimetypes.guess_type(file.filename)[0] or file.mimetype or 'application/octet-stream'
-    filename = secure_filename(file.filename, 'template')
```
plourenco, line 192: Just to confirm, why was this check removed? Is it safe to trust the file?

ThiefMaster: No, it's a name coming from the user. But we still apply it for the filename that is used for storage on the file system (or whatever storage backend is used). It's just no longer applied to the file name stored in the database, so when you download a file stored in Indico later it will give you the same filename used during upload.

plourenco: All right, sounds good.

ThiefMaster: FWIW it's not unsafe per se, but I did notice that it lets you upload garbage files containing `../` or `/` in general which would then break the download links since the filenames are not supposed to contain either of those. So I'll add a new util that sanitizes a filename for display purposes and strips this stuff while keeping non-ASCII and spaces.

ThiefMaster: Updated to strip this kind of garbage anyway while preserving useful stuff.

PR #4472, opened 2020-05-26
Show actual editor name in timeline

### LINKED ISSUE #4467: Show actual name in timeline
Currently we only show "Editor has made some changes to the paper". It would be nicer to show something like

> Editor *Adrian Mönnich* has made some changes to the paper

or

> Adrian Mönnich (editor) has made some changes to the paper

---

BTW, we have something similar when a submitter accepts/rejected proposed changes ("Submitter rejected proposed changes") but we do not store who performed this action, so we have no way to show it there. :/

panagiotappl:
closes #4467

### DIFF
--- a/indico/modules/events/editing/client/js/editing/timeline/selectors.js
+++ b/indico/modules/events/editing/client/js/editing/timeline/selectors.js
@@ -38,18 +38,21 @@ export function processRevisions(revisions) {
   const newRevisions = [];
   let numberOfRevisions = 1;
   let currentRevision = revisions[0];
+  let previousRevision = null;
   let items = [...currentRevision.comments];
 
   for (const [index, revision] of revisions.entries()) {
     const {initialState, finalState} = revision;
-
     if (!currentRevision) {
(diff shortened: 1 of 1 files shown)


indico/modules/events/editing/client/js/editing/timeline/selectors.js
```diff
@@ -38,18 +38,20 @@ export function processRevisions(revisions) {
...
     if (!currentRevision) {
       currentRevision = {...revision};
       items = [...currentRevision.comments];
     }
 
     if (initialState.name === InitialRevisionState.needs_submitter_confirmation) {
-      currentRevision.header = Translate.string('Editor has made some changes to the paper');
+      currentRevision.header = `${previousRevision.editor.fullName} ${Translate.string(
+        '(editor) has made some changes to the paper'
+      )}`;
```
ThiefMaster, line 54: Use a placeholder in the translatable string - the name may be somewhere else in other languages.

PR #4473, opened 2020-05-27
Add functionality to opening/closing of submissions/editing

### LINKED ISSUE #4469: Editing: Open/Close submission
We already have buttons for this in the management area on the editable type page, but they have no functionality yet. When submission is closed the corresponding element on the contribution display page should be hidden as well.

### LINKED ISSUE #4470: Editing: Open/Close editing
We already have buttons for this in the management area on the editable type page, but they have no functionality yet. When editing is closed, editors who aren't managers as well should not be able to perform any editing-related actions.

Leats:
Closes #4469 and closes #4470.

### DIFF
--- a/indico/modules/events/editing/settings.py
+++ b/indico/modules/events/editing/settings.py
@@ -18,7 +18,11 @@ editing_settings = EventSettingsProxy('editing', {
     'editable_types': ['paper'],
 })
 
-_defaults = {'self_assign_allowed': False, 'submission_enabled': False}
+_defaults = {
+    'self_assign_allowed': False,
+    'submission_enabled': False,
+    'editing_enabled': False,
+}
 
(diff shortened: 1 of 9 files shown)


indico/modules/events/editing/settings.py
```diff
@@ -18,7 +18,7 @@
     'editable_types': ['paper'],
 })
 
-_defaults = {'self_assign_allowed': False}
+_defaults = {'self_assign_allowed': False, 'submission_enabled': False, 'editing_enabled': False}
```
ThiefMaster, line 21: I think this would look a bit nicer:

```python
_defaults = {
    'self_assign_allowed': False,
    'submission_enabled': False,
    'editing_enabled': False,
}
```

PR #4475, opened 2020-05-28
Implement modifiedSinceLastSubmit in room edit modal form

### LINKED ISSUE #4435: New room edit modal: restore back old success msg behavior
Once #4408 is merged, the "submit successful" message doesn't permanently disappear (until next submission) after submitting, but rather goes away correctly when the form goes `dirtySinceLastSubmit` but shows up again when the form loses that state.

By implementing final-form/final-form#215 this can be fixed nicely without having to keep state on our side.

plourenco:
Closes #4435


package-lock.json
```diff
@@ -4527,6 +4527,11 @@
       "resolved": "https://registry.npmjs.org/@reach/observe-rect/-/observe-rect-1.0.5.tgz",
       "integrity": "sha512-EQLHhquzwAh2TFtMJU0PXJRmhpOrAhzApgoI4Lwn72J7ygrMk3Ys3a3i8KDnV0TetbwAxn4XSHi4pE8sPWUYLg=="
     },
+    "@scarf/scarf": {
+      "version": "1.0.5",
+      "resolved": "https://registry.npmjs.org/@scarf/scarf/-/scarf-1.0.5.tgz",
+      "integrity": "sha512-9WKaGVpQH905Aqkk+BczFEeLQxS07rl04afFRPUG9IcSlOwmo5EVVuuNu0d4M9LMYucObvK0LoAe+5HfMW2QhQ=="
+    },
```
ThiefMaster, line 4534: I was curious what this new dependency final-form added was, and it turns out to be some [analytics package](https://final-form.org/docs/react-final-form/getting-started#installation).

Maybe we should opt-out from this in our package.json? While I can understand the developer wanting to get some analytics about where his package is installed, and it would only affect developers (since our end users don't install any JS from npm), I usually don't like analytics that much (unless I see the results ;)), so I'd probably [opt-out](https://github.com/scarf-sh/scarf-js#as-a-user-of-a-package-using-scarf-how-can-i-opt-out-of-analytics) from them.

PR #4489, opened 2020-06-10
Create a closer link between peer reviewing and editing

### LINKED ISSUE #4303: Closer link between Paper Peer Reviewing and Paper Edtiting
There should be a way to (automatically?) pass a paper from layout editing to peer reviewing ("paper reviewing") (or was it vice versa? to be confirmed when we have access to the flowchart for this again...)

plourenco:
Closes #4303 
Under development

TODOs:
- [x] Re-uploading existing files results in multiple copies (if multiple files are enabled)
- [x] Confirm if the copies of the files are deleted accordingly
- [x] Add a link to the editable from the paper reviewing
- [ ] Add a warning if the editable was rejected in peer-reviewing? - probably for another task
- [x] Check the slow loadings
- [x] Disable the editing step if editing is not enabled
- [x] Restrict the steps view to authors only
- [x] Decide between uploading files from any last revision or only when the peer reviewing finished
- [x] Consider whether it is worth merging the `existingFiles` inside `_fileTypes` for FileManager children

PS: We considered re-using the `files` property in the `FileManager` like we do when submitting revisions. However, PaperFiles use `ids`, while editables use `uuids`, it's very complicated to merge them within the `FileManager`. To avoid a complete refactoring, I had to make this a new feature.

### DIFF
--- a/indico/modules/events/contributions/client/js/index.jsx
+++ b/indico/modules/events/contributions/client/js/index.jsx
@@ -33,23 +33,27 @@ import PublicationSwitch from './PublicationSwitch';
     const availableTypes = JSON.parse(editableSubmissionButton.dataset.availableTypes);
     const {eventId, contributionId, contributionCode} = editableSubmissionButton.dataset;
 
-    const tasks = await Promise.allSettled([
-      Promise.all(
-        availableTypes.map(type => indicoAxios.get(fileTypesURL({confId: eventId, type})))
-      ),
-      indicoAxios.get(paperInfoURL({confId: eventId, contrib_id: contributionId})),
-    ]);
-    if (tasks[0].status === 'rejected') {
-      handleAxiosError(tasks[0].reason);
+    let responses, lastRevFiles;
+    try {
+      responses = await Promise.all([
+        Promise.all(
+          availableTypes.map(type => indicoAxios.get(fileTypesURL({confId: eventId, type})))
+        ),
+        indicoAxios
+          .get(paperInfoURL({confId: eventId, contrib_id: contributionId}))
+          // eslint-disable-next-line no-unused-vars
+          .catch(e => undefined),
+      ]);
+    } catch (e) {
+      handleAxiosError(e);
       return;
     }
     const fileTypes = _.fromPairs(
-      tasks[0].value.map((response, index) => [availableTypes[index], camelizeKeys(response.data)])
+      responses[0].map((response, index) => [availableTypes[index], camelizeKeys(response.data)])
     );
 
-    let lastRevFiles;
-    if (tasks[1].status === 'fulfilled') {
-      const {isInFinalState, lastRevision} = camelizeKeys(tasks[1].value.data);
+    if (responses[1]) {
+      const {isInFinalState, lastRevision} = camelizeKeys(responses[1].data);
       if (isInFinalState && lastRevision) {
         lastRevFiles = lastRevision.files;
       }
--- a/indico/modules/events/contributions/client/js/index.jsx
+++ b/indico/modules/events/contributions/client/js/index.jsx
@@ -49,19 +49,19 @@ import PublicationSwitch from './PublicationSwitch';
 
     let lastRevFiles;
     if (tasks[1].status === 'fulfilled') {
-      const paperInfo = camelizeKeys(tasks[1].value.data);
-      if (paperInfo.lastRevision) {
-        lastRevFiles = paperInfo.lastRevision.files;
+      const {isInFinalState, lastRevision} = camelizeKeys(tasks[1].value.data);
+      if (isInFinalState && lastRevision) {
+        lastRevFiles = lastRevision.files;
       }
     }
 
     ReactDOM.render(
       <EditableSubmissionButton
         fileTypes={fileTypes}
-        eventId={eventId}
-        contributionId={contributionId}
+        eventId={Number(eventId)}
+        contributionId={Number(contributionId)}
         contributionCode={contributionCode}
-        existingFiles={lastRevFiles}
+        uploadableFiles={lastRevFiles}
       />,
       editableSubmissionButton
     );
--- a/indico/modules/events/contributions/client/js/index.jsx
+++ b/indico/modules/events/contributions/client/js/index.jsx
@@ -10,6 +10,7 @@
 /* global enableIfChecked:true */
 
 import fileTypesURL from 'indico-url:event_editing.api_file_types';
+import paperInfoURL from 'indico-url:papers.api_paper_details';
 
 import _ from 'lodash';
 import moment from 'moment';
@@ -29,23 +30,30 @@ import PublicationSwitch from './PublicationSwitch';
     if (!editableSubmissionButton) {
       return;
     }
-
     const availableTypes = JSON.parse(editableSubmissionButton.dataset.availableTypes);
     const {eventId, contributionId, contributionCode} = editableSubmissionButton.dataset;
 
-    let responses;
-    try {
-      responses = await Promise.all(
+    const tasks = await Promise.allSettled([
+      Promise.all(
         availableTypes.map(type => indicoAxios.get(fileTypesURL({confId: eventId, type})))
-      );
-    } catch (e) {
-      handleAxiosError(e);
+      ),
+      indicoAxios.get(paperInfoURL({confId: eventId, contrib_id: contributionId})),
+    ]);
+    if (tasks[0].status === 'rejected') {
+      handleAxiosError(tasks[0].reason);
       return;
     }
-
     const fileTypes = _.fromPairs(
-      responses.map((response, index) => [availableTypes[index], camelizeKeys(response.data)])
+      tasks[0].value.map((response, index) => [availableTypes[index], camelizeKeys(response.data)])
     );
+
+    let lastRevFiles;
+    if (tasks[1].status === 'fulfilled') {
+      const paperInfo = camelizeKeys(tasks[1].value.data);
+      if (paperInfo.lastRevision) {
+        lastRevFiles = paperInfo.lastRevision.files;
+      }
+    }
 
     ReactDOM.render(
       <EditableSubmissionButton
@@ -53,6 +61,7 @@ import PublicationSwitch from './PublicationSwitch';
         eventId={eventId}
         contributionId={contributionId}
         contributionCode={contributionCode}
+        existingFiles={lastRevFiles}
       />,
       editableSubmissionButton
     );
--- a/indico/modules/events/contributions/client/js/index.jsx
+++ b/indico/modules/events/contributions/client/js/index.jsx
@@ -33,14 +33,14 @@ import PublicationSwitch from './PublicationSwitch';
     const availableTypes = JSON.parse(editableSubmissionButton.dataset.availableTypes);
     const {eventId, contributionId, contributionCode} = editableSubmissionButton.dataset;
 
-    let responses, lastRevFiles;
+    let fileTypeResponses, paperInfoResponse, lastRevFiles;
     try {
-      responses = await Promise.all([
+      [fileTypeResponses, paperInfoResponse] = await Promise.all([
         Promise.all(
           availableTypes.map(type => indicoAxios.get(fileTypesURL({confId: eventId, type})))
         ),
         indicoAxios.get(paperInfoURL({confId: eventId, contrib_id: contributionId}), {
-          validateStatus: status => status >= 200 && status <= 404,
+          validateStatus: status => (status >= 200 && status < 300) || status === 404,
         }),
       ]);
     } catch (e) {
@@ -48,11 +48,14 @@ import PublicationSwitch from './PublicationSwitch';
       return;
     }
     const fileTypes = _.fromPairs(
-      responses[0].map((response, index) => [availableTypes[index], camelizeKeys(response.data)])
+      fileTypeResponses.map((response, index) => [
+        availableTypes[index],
+        camelizeKeys(response.data),
+      ])
     );
 
-    if (responses[1]) {
-      const {isInFinalState, lastRevision} = camelizeKeys(responses[1].data);
+    if (paperInfoResponse) {
+      const {isInFinalState, lastRevision} = camelizeKeys(paperInfoResponse.data);
       if (isInFinalState && lastRevision) {
         lastRevFiles = lastRevision.files;
       }
--- a/indico/modules/events/contributions/client/js/index.jsx
+++ b/indico/modules/events/contributions/client/js/index.jsx
@@ -39,10 +39,9 @@ import PublicationSwitch from './PublicationSwitch';
         Promise.all(
           availableTypes.map(type => indicoAxios.get(fileTypesURL({confId: eventId, type})))
         ),
-        indicoAxios
-          .get(paperInfoURL({confId: eventId, contrib_id: contributionId}))
-          // eslint-disable-next-line no-unused-vars
-          .catch(e => undefined),
+        indicoAxios.get(paperInfoURL({confId: eventId, contrib_id: contributionId}), {
+          validateStatus: status => status >= 200 && status <= 404,
+        }),
       ]);
     } catch (e) {
       handleAxiosError(e);
@@ -62,8 +61,8 @@ import PublicationSwitch from './PublicationSwitch';
     ReactDOM.render(
       <EditableSubmissionButton
         fileTypes={fileTypes}
-        eventId={Number(eventId)}
-        contributionId={Number(contributionId)}
+        eventId={+eventId}
+        contributionId={+contributionId}
         contributionCode={contributionCode}
         uploadableFiles={lastRevFiles}
       />,
--- a/indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
+++ b/indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
@@ -8,7 +8,7 @@
 import editableURL from 'indico-url:event_editing.editable';
 import submitRevisionURL from 'indico-url:event_editing.api_create_editable';
 import apiUploadURL from 'indico-url:event_editing.api_upload';
-import apiUploadExistingURL from 'indico-url:event_editing.api_upload_last_revision';
+import apiUploadExistingURL from 'indico-url:event_editing.api_add_paper_file';
 
 import _ from 'lodash';
 import React, {useState} from 'react';
@@ -31,7 +31,7 @@ export default function EditableSubmissionButton({
   contributionId,
   contributionCode,
   fileTypes,
-  existingFiles,
+  uploadableFiles,
 }) {
   const [currentType, setCurrentType] = useState(null);
   const submitRevision = async formData => {
@@ -89,7 +89,7 @@ export default function EditableSubmissionButton({
                       contrib_id: contributionId,
                       type: currentType,
                     })}
-                    existingFiles={existingFiles}
+                    uploadableFiles={uploadableFiles}
                     mustChange
                   />
                 </Form>
@@ -131,12 +131,12 @@ export default function EditableSubmissionButton({
 
 EditableSubmissionButton.propTypes = {
   fileTypes: PropTypes.objectOf(PropTypes.arrayOf(PropTypes.shape(fileTypePropTypes))).isRequired,
-  eventId: PropTypes.string.isRequired,
-  contributionId: PropTypes.string.isRequired,
+  eventId: PropTypes.number.isRequired,
+  contributionId: PropTypes.number.isRequired,
   contributionCode: PropTypes.string.isRequired,
-  existingFiles: PropTypes.arrayOf(PropTypes.shape(uploadablePropTypes)),
+  uploadableFiles: PropTypes.arrayOf(PropTypes.shape(uploadablePropTypes)),
 };
 
 EditableSubmissionButton.defaultProps = {
-  existingFiles: [],
+  uploadableFiles: [],
 };
--- a/indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
+++ b/indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
@@ -18,13 +18,13 @@ import {Form as FinalForm} from 'react-final-form';
 
 import {indicoAxios} from 'indico/utils/axios';
 import {FinalSubmitButton, handleSubmitError} from 'indico/react/forms';
-import {Param, Translate} from 'indico/react/i18n';
+import {Translate} from 'indico/react/i18n';
 
 import {fileTypePropTypes, uploadablePropTypes} from './timeline/FileManager/util';
 import {FinalFileManager} from './timeline/FileManager';
 
 import {getFileTypes} from './timeline/selectors';
-import {EditableTypeTitles, EditableType} from '../models';
+import {EditableTypeTitles} from '../models';
 
 export default function EditableSubmissionButton({
   eventId,
@@ -32,6 +32,7 @@ export default function EditableSubmissionButton({
   contributionCode,
   fileTypes,
   uploadableFiles,
+  text,
 }) {
   const [currentType, setCurrentType] = useState(null);
   const submitRevision = async formData => {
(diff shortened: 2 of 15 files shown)


indico/modules/events/papers/client/js/components/Paper.jsx
```diff
@@ -59,3 +85,54 @@ Paper.propTypes = {
...
+  const {event, contribution} = useSelector(getPaperDetails);
+
+  return (
+    <Step.Group ordered fluid>
+      <Step completed={isAccepted}>
+        <Step.Content>
+          <Step.Title>Peer Reviewing</Step.Title>
+          <Step.Description style={{marginBottom: '0'}}>
+            {isAccepted ? (
+              <Translate>The paper was accepted</Translate>
```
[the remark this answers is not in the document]
plourenco: I used `was` because `has been` looks more like something that happened just now, an immediate warning. While the first refers to a past action, can be an hour, day, week.

indico/modules/events/papers/client/js/components/Paper.jsx
```diff
@@ -59,3 +85,54 @@ Paper.propTypes = {
...
+
+function PaperSteps({isAccepted, hasEditable, fileTypes, existingFiles}) {
+  const {event, contribution} = useSelector(getPaperDetails);
+
+  return (
+    <Step.Group ordered fluid>
+      <Step completed={isAccepted}>
+        <Step.Content>
+          <Step.Title>Peer Reviewing</Step.Title>
+          <Step.Description style={{marginBottom: '0'}}>
```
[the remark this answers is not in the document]
plourenco: I've seen it both ways. But zero works and definitely looks better.

indico/web/client/js/react/hooks.js
```diff
@@ -103,10 +103,11 @@ FavoritesProvider.propTypes = {
   children: PropTypes.func.isRequired,
 };
 
-export function useIndicoAxios({camelize, ...args}) {
+export function useIndicoAxios({camelize, unhandledErrors = [], ...args}) {
   const lastData = useRef(null);
   const {response, error, loading, reFetch} = useAxios({
-    customHandler: err => err && handleAxiosError(err),
+    customHandler: err =>
+      err && !unhandledErrors.includes(err.response.status) && handleAxiosError(err),
```
ThiefMaster, line 110: can `err.response` be `null` in case the network request failed?

indico/modules/events/papers/schemas.py
```diff
@@ -192,6 +194,10 @@ class PaperSchema(mm.Schema):
     can_judge = Function(lambda paper, ctx: paper.can_judge(ctx.get('user')))
     can_comment = Function(lambda paper, ctx: paper.can_comment(ctx.get('user'), check_state=True))
     can_review = Function(lambda paper, ctx: paper.can_review(ctx.get('user')))
+    can_submit_proceedings = Function(lambda paper, ctx: paper.contribution.can_submit_proceedings(ctx.get('user')))
+    editing_enabled = Function(
+        lambda paper, : editable_type_settings[EditableType.paper].get(paper.event, 'editing_enabled')
```
ThiefMaster, line 199: i didn't even know you can omit an argument like this... looks super ugly IMO, so i'd go for `paper, ctx` even if `ctx` isn't used.

indico/web/client/js/react/hooks.js
plourenco: I asked the same thing, I didn't want to expand the condition too much. But I guess it is safer to check it.

indico/modules/events/contributions/client/js/index.jsx
```diff
@@ -29,30 +30,38 @@ import PublicationSwitch from './PublicationSwitch';
...
       return;
     }
-
     const availableTypes = JSON.parse(editableSubmissionButton.dataset.availableTypes);
     const {eventId, contributionId, contributionCode} = editableSubmissionButton.dataset;
 
-    let responses;
-    try {
-      responses = await Promise.all(
+    const tasks = await Promise.allSettled([
```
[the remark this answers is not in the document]
plourenco: I did this mainly because `Promise.all` rejects with the reason of the first promise that rejects. However, the second API call to `paperInfoURL` is expected to reject if there is no paper.
But since we require a polyfill, I can go around and not throw.

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -85,6 +99,8 @@ def _editable_query_options(self):
 
     def _check_access(self):
         RHContributionEditableBase._check_access(self)
+        if not self.editable:
+            raise NotFound(_('There is no editable for this contribution yet.'))
```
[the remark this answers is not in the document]
plourenco: Good idea!

indico/modules/events/contributions/client/js/index.jsx
```diff
@@ -29,30 +30,42 @@ import PublicationSwitch from './PublicationSwitch';
...
-        availableTypes.map(type => indicoAxios.get(fileTypesURL({confId: eventId, type})))
-      );
+      responses = await Promise.all([
+        Promise.all(
+          availableTypes.map(type => indicoAxios.get(fileTypesURL({confId: eventId, type})))
+        ),
+        indicoAxios
+          .get(paperInfoURL({confId: eventId, contrib_id: contributionId}))
+          // eslint-disable-next-line no-unused-vars
+          .catch(e => undefined),
```
ThiefMaster, line 45: `() => undefined` to avoid the unused var

But I'd rather override `validateStatus` in the options for this particular request to avoid rejecting the promise on the expected error status code altogether...

indico/modules/events/contributions/client/js/index.jsx
```diff
@@ -29,30 +30,42 @@ import PublicationSwitch from './PublicationSwitch';
...
+        lastRevFiles = lastRevision.files;
+      }
+    }
+
     ReactDOM.render(
       <EditableSubmissionButton
         fileTypes={fileTypes}
-        eventId={eventId}
-        contributionId={contributionId}
+        eventId={Number(eventId)}
```
ThiefMaster, line 65: `parseInt(eventId, 10)`

indico/modules/events/editing/client/js/editing/timeline/FileManager/FileManager.module.scss
```diff
@@ -34,7 +34,7 @@
   flex-wrap: wrap;
 
   .file-type {
-    overflow: hidden;
+    // overflow: hidden;
```
ThiefMaster, line 37: let's remove that if it's not needed?

indico/modules/events/editing/client/js/editing/timeline/FileManager/util.js
```diff
@@ -48,33 +54,45 @@ async function uploadFile(file, url, onUploadProgress) {
...
   fileTypeId,
   acceptedFiles,
-  uploadURL,
+  uploadCallback,
   dispatch,
-  fileId = null,
-  onError = null
+  replaceFileId = null,
+  onError = null,
+  fileId = null
```
ThiefMaster, line 87: Maybe it's time to pass a singe argument (an object) to this function? Having that many positional arguments is getting really messy!

indico/modules/events/papers/client/js/components/Paper.jsx
```diff
@@ -59,3 +105,90 @@ Paper.propTypes = {
...
+        </Step.Content>
+      </Step>
+      <Step active={isInFinalState} completed={hasEditable} disabled={isRejected}>
+        <Icon name="pencil alternate" />
+        <Step.Content>
+          <Step.Title>
+            {hasEditable || !isInFinalState || isRejected ? (
+              <Translate>Submit for Editing</Translate>
+            ) : (
+              <EditableSubmissionButton
```
ThiefMaster, line 156: "Editing" should be mentioned here somewhere because right now we don't once the button is displayed:

image

indico/modules/events/editing/client/js/editing/timeline/FileManager/util.js
```diff
@@ -48,33 +54,45 @@ async function uploadFile(file, url, onUploadProgress) {
   }
 }
 
+export async function uploadAnExistingFile(url, file) {
```
ThiefMaster, line 57: Just `uploadExistingFile`? Having articles in function names is rather uncommon and sounds a bit weird

indico/modules/events/editing/client/js/editing/timeline/FileManager/util.js
```diff
@@ -48,33 +54,45 @@ async function uploadFile(file, url, onUploadProgress) {
...
+ * @param {String?} replaceFileId - the ID of the file to modify, if any
  * @param {Function} onError - the function to be called on file upload error
+ * @param {Number} fileId - the id of the file, if any, used to cross-reference uploadables
  */
 export function uploadFiles(
   action,
   fileTypeId,
   acceptedFiles,
-  uploadURL,
+  uploadCallback,
```
ThiefMaster, line 83: I don't think `callback` is really fitting here... `uploadFunc`?

indico/modules/events/editing/client/js/editing/timeline/FileManager/util.js
```diff
@@ -108,11 +127,14 @@ export async function deleteFile(uuid) {
...
 }
 
-export function mapFileTypes(fileTypes, files) {
+export function mapFileTypes(fileTypes, files, uploadableFiles = []) {
   return fileTypes.map(fileType => ({
     ...fileType,
     files: files.filter(file => file.fileType === fileType.id).map(f => ({...f, claimed: true})),
     invalidFiles: [],
+    uploadableFiles: uploadableFiles.filter(file =>
+      fileType.extensions.includes(file.filename.split('.').pop())
```
ThiefMaster, line 136: I think you are not handling the case where a file type has no extension restrictions

indico/modules/events/contributions/client/js/index.jsx
plourenco: Is there any difference, isn't `Number` cleaner?

indico/modules/events/editing/client/js/editing/timeline/FileManager/FileManager.module.scss
plourenco: Yes, I agree!

indico/modules/events/editing/client/js/editing/timeline/FileManager/util.js
plourenco: I was starting to think it is getting messy too. I'm thinking about moving the delete - when replacing - to a util function and I should be able to refactor this too.

indico/modules/events/papers/client/js/components/Paper.jsx
plourenco: I tried to, the problem is, adding a title near the button completely breaks the positioning. Even if I try to make the button smaller, either the whole container will expand or they will seem out of margin. Perhaps changing the text in button?

indico/modules/events/contributions/client/js/index.jsx
ThiefMaster: I think in this case it [doesn't really matter](https://stackoverflow.com/questions/4564158/what-is-the-difference-between-parseintstring-and-numberstring-in-javascript). You could even use the `eventId={+eventId}` shortcut ;)

When dealing with user input, `+` and `Number` have the disadvantage of converting `0x*` and (I think) `0*` (octal) as well which is confusing if you type e.g. `0123` and get `83`.

indico/modules/events/papers/client/js/components/Paper.jsx
ThiefMaster: That's what I had in mind ;)

indico/modules/events/editing/client/js/editing/timeline/FileManager/util.js
plourenco: I’ll add an exception. Thanks!

plourenco: PS: I added a condition inside the `filter` instead of an outside check because I just thought it would be a good idea to preserve the immutability `filter` provides.

indico/modules/events/editing/client/js/editing/timeline/FileManager/util.js
plourenco: I've seen `Func` being used multiple times before so, seems like a candidate

indico/modules/events/editing/client/js/editing/timeline/FileManager/util.js
plourenco: Sounds good

indico/modules/events/contributions/client/js/index.jsx
plourenco: Totally forgot about the `+`. It seems like a good idea considering ids should be safe here!

indico/modules/events/contributions/client/js/index.jsx
plourenco: Didn't know about `validateStatus`. I tried searching for something that would catch them.
For the sake of simplicity, I made it resolve errors between 200 and 404, since I don't think resolving errors between 400 and 404 is a big problem.

ThiefMaster: I don't think that's a good idea. Especially 403 is something where we usually want error handling, since a Forbidden error is rather unexpected there...

plourenco: I didn't expect that to happen in the front-end but I've made it strict.

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -39,10 +39,26 @@
...
 class RHEditingUploadFile(UploadFileMixin, RHContributionEditableBase):
+    EDITABLE_REQUIRED = False
+
     def get_file_context(self):
         return 'event', self.event.id, 'editing', self.contrib.id, self.editable_type.name
 
 
+class RHEditingUploadPaperLastRevision(RHEditingUploadFile):
+    @use_kwargs({
+        'id': fields.Int()
```
plourenco, line 50: Is it worth changing this to a list of numbers (ids) instead?

ThiefMaster: when would the be useful?

indico/modules/events/contributions/client/js/index.jsx
```diff
@@ -29,30 +30,41 @@ import PublicationSwitch from './PublicationSwitch';
...
+        lastRevFiles = lastRevision.files;
+      }
+    }
+
     ReactDOM.render(
       <EditableSubmissionButton
         fileTypes={fileTypes}
-        eventId={eventId}
-        contributionId={contributionId}
+        eventId={+eventId}
```
ParthS007, line 64: what's this `+` for? 

Sorry if it has already been discussed, just point me to that.

indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
```diff
@@ -81,6 +85,12 @@ export default function EditableSubmissionButton({
                       contrib_id: contributionId,
                       type: currentType,
                     })}
+                    uploadExistingURL={apiUploadExistingURL({
+                      confId: eventId,
+                      contrib_id: contributionId,
```
ParthS007, line 90:
```suggestion
                      contribId: contributionId,
```

indico/modules/events/papers/client/js/components/Paper.jsx
```diff
@@ -8,37 +8,72 @@
...
+  const {data: fileTypes, loading: isFileTypesLoading} = useIndicoAxios({
+    url: fileTypesURL({confId: eventId, type: EditableType.paper}),
+    trigger: eventId,
+    unhandledErrors: [404], // if editing module is disabled
+    camelize: true,
+  });
+  const {data: editable, loading: isEditableLoading} = useIndicoAxios({
+    url: editableDetailsURL({
+      confId: eventId,
+      contrib_id: contributionId,
```
ParthS007, line 46:
```suggestion
      contribId: contributionId,
```

indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
ThiefMaster: Those are parameters to build an URL. They must be in snake_case because that's how they are defined in the blueprint..

indico/modules/events/editing/controllers/backend/timeline.py
plourenco: Since `id` is the only parameter used. Instead of `[{ "id": 17 }, ...]`. I would have `[17,...]`.

indico/modules/events/contributions/client/js/index.jsx
plourenco: Returns the numerical representation, similar to `Number()`
https://github.com/indico/indico/pull/4489#discussion_r442167578

indico/modules/events/editing/controllers/backend/timeline.py
ThiefMaster: What I mean is you never have more than one `id` argument... so it's `{id: 17}` vs `17`.. and I don't think webargs can process the latter

indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
plourenco: Yes, unfortunately, there are way too many references to change. Perhaps in a separate task.

indico/modules/events/papers/client/js/components/Paper.jsx
plourenco: Similar to the above.

indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
ThiefMaster: We do not want to change this because that would introduce camelCase on the Python side ;)

indico/modules/events/editing/controllers/backend/timeline.py
plourenco: Never mind, I mixed this with `RHSubmitNewRevision`. Forgot only one file was accepted here.

indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
plourenco: Yes, I was thinking about `confId` to `conf_id`.

ThiefMaster: That would be nice in principle - except that the standard for names in JS is camelCase...

The `snakifyKeys` util would work, but for less widespread names using it would have the disadvantage that searching for the param name using one style would not find occurrences of it in the other style. With the explicit mapping you at least find the place with the mapping basically telling you that there's something else to search for..

CHANGES.rst
```diff
@@ -76,6 +76,8 @@ Improvements
 - Preserve non-ascii characters in file names (:issue:`4465`)
 - Allow resetting moderation state from registration management view
   (:issue:`4498`, thanks :user:`omegak`)
+- Create a new editing interface inside paper reviewing and add the option to
+  re-use accepted papers as editable files (:issue:`4489`).
```
ThiefMaster, line 80: Just wondering if a changelog entry makes sense here since this is related to a new "major feature" (even though Editing is not listed there yet)

indico/modules/events/papers/schemas.py
```diff
@@ -192,6 +194,13 @@ class PaperSchema(mm.Schema):
     can_judge = Function(lambda paper, ctx: paper.can_judge(ctx.get('user')))
     can_comment = Function(lambda paper, ctx: paper.can_comment(ctx.get('user'), check_state=True))
     can_review = Function(lambda paper, ctx: paper.can_review(ctx.get('user')))
+    can_submit_proceedings = Function(lambda paper, ctx: paper.contribution.can_submit_proceedings(ctx.get('user')))
+    editing_open = Function(
+        lambda paper, ctx: editable_type_settings[EditableType.paper].get(paper.event, 'submission_enabled')
+    )
+    editing_enabled = Function(
+        lambda paper, ctx: 'paper' in editing_settings.get(paper.event, 'editable_types')
+    )
```
ThiefMaster, line 203: What if you enabled paper editing but then disabled the Editing feature altogether? I think you also need a check for `paper.event.has_feature('editing')`

indico/modules/events/papers/client/js/components/Paper.jsx
```diff
@@ -59,3 +108,77 @@ Paper.propTypes = {
...
+                uploadableFiles={uploadableFiles}
+                text={Translate.string('Submit for editing')}
+              />
+            )}
+          </Step.Title>
+          {hasEditable && (
+            <Step.Description style={{marginBottom: 0}}>
+              <Translate>
+                The paper was{' '}
+                <Param name="submittedURL" value="submitted" wrapper={<a href={editableURL} />} />{' '}
```
ThiefMaster, line 161: `name="url"` (no need to be overly verbose there).

Also, the hardcoded value kind of breaks i18n here. You need to use Param like this instead:

    <Param name="url" wrapper={<a href={editableURL} />}>submitted</Param>

indico/modules/events/editing/client/js/editing/EditableSubmissionButton.jsx
```diff
@@ -97,9 +107,11 @@ export default function EditableSubmissionButton({
       </FinalForm>
       {editableTypes.length === 1 ? (
         <Button onClick={() => setCurrentType(editableTypes[0])} primary>
-          <Translate>
-            Submit <Param name="editableType" value={editableTypes[0]} />
-          </Translate>
+          {text || (
+            <Translate>
+              Submit <Param name="editableType" value={editableTypes[0]} />
```
ThiefMaster, line 112: I know you just reindented it, but this is not properly translatable. it should use a lookup dict with static messages for the different types or simply 3 separate conditions like we do for the `<Modal.Header>` in the same file.

indico/modules/events/contributions/client/js/index.jsx
```diff
@@ -29,30 +30,41 @@ import PublicationSwitch from './PublicationSwitch';
...
     const availableTypes = JSON.parse(editableSubmissionButton.dataset.availableTypes);
     const {eventId, contributionId, contributionCode} = editableSubmissionButton.dataset;
 
-    let responses;
+    let responses, lastRevFiles;
     try {
-      responses = await Promise.all(
-        availableTypes.map(type => indicoAxios.get(fileTypesURL({confId: eventId, type})))
-      );
+      responses = await Promise.all([
```
ThiefMaster, line 38: What about `[fileTypeResponses, paperInfoResponse] = ...`?

indico/modules/events/papers/schemas.py
plourenco: I can add that, but I don't think it will affect this PR because the call to `file-types` returns 404 if the editing feature is disabled, no file types. I'm using that as the third condition to disable it.

CHANGES.rst
plourenco: Thought about the same but wasn't sure if you/others would agree. I can move it there.

ThiefMaster: I think we'd have to write something else up there, since that changelog entry would need to mention the editing module itself, not just one small thing from it

indico/modules/events/contributions/client/js/index.jsx
plourenco: Looks good

indico/modules/events/papers/client/js/components/Paper.jsx
plourenco: Thought I had moved the value within, maybe stashed it. Naming's good 👍

indico/modules/events/papers/schemas.py
ThiefMaster: feels a bit implicit even if it works... and `editing_enabled: true` would be incorrect if editing isn't enabled ;)

PR #4490, opened 2020-06-10
Add Contact dialog in Editing

### LINKED ISSUE #4444: Editing: Contact editors
This can be done similar to what we have in the paper reviewing module. Let's see if we can easily reuse something from there.

panagiotappl:
closes #4444

### DIFF
--- a/indico/modules/events/editing/controllers/backend/management.py
+++ b/indico/modules/events/editing/controllers/backend/management.py
@@ -27,6 +27,7 @@ from indico.modules.events.editing.util import get_editors
 from indico.modules.events.logs import EventLogKind, EventLogRealm
 from indico.util.i18n import _, orig_string
 from indico.web.args import use_kwargs, use_rh_args, use_rh_kwargs
+from indico.web.util import jsonify_template
 
 
 class RHCreateTag(RHEditingManagementBase):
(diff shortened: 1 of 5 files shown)


panagiotappl:
Also, regarding the separate task. Maybe we could avoid adding a category role/event role in the editing team if there are no users in it yet?

ThiefMaster:
I think that would be rather confusing - you may want to set things up first before deciding on who the users in the role shall be.

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -223,3 +225,12 @@ def _process_DELETE(self):
                        'Closed {} editing'.format(orig_string(self.editable_type.title)), session.user)
         editable_type_settings[self.editable_type].set(self.event, 'editing_enabled', False)
         return '', 204
+
+
+class RHContactEditingTeam(RHEditableTypeManagementBase):
+    """Send emails to editing team"""
```
Leats, line 231:
```suggestion
    """Send emails to editing team."""
```
To keep it consistent with the rest of the file

indico/core/db/sqlalchemy/principals.py
```diff
@@ -413,13 +413,29 @@ def get_emails(self):
...
         """
         if self.type == PrincipalType.user and not self.user.is_system:
             return {self.user.email}
         elif self.type in (PrincipalType.local_group, PrincipalType.multipass_group):
             return {x.email for x in self.principal.get_members() if not x.is_system}
+        elif self.type in (PrincipalType.event_role, PrincipalType.category_role):
+            return {x.email for x in self.principal.members if not x.is_system}
+        return set()
+
+    def get_users(self):
```
plourenco, line 427: What do you think about making this generic (getting a principal's attribute)? Such as:
```
get_attribute(attribute, include_system=False): ...

get_attribute('email', include_system=True) // this would be emails
```

Could this be useful in the future?

PR #4485, opened 2020-06-11
Editable list: Action button functionality

ThiefMaster:
Connect the action buttons above the editable list to the actions (which have already been implemented in the backend).

We also need to find a solution for updating the displayed list after performing an action. I don't think reloading the whole list after every action would be OK here since it may be large and thus take a few moments.

depends on #4460

---

Keeping the list in state is a bit ugly, but I didn't want to spend time on any kind of big refactoring in this component during the task. And since we never change the `eventId` prop we won't have a case where the initial list is reloaded but the state still has a stale version.

### DIFF
--- a/indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
+++ b/indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
@@ -5,14 +5,18 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
-import dashboardURL from 'indico-url:event_editing.dashboard';
+import editableTypeURL from 'indico-url:event_editing.manage_editable_type';
+import editorsURL from 'indico-url:event_editing.api_editable_type_editors';
 import editableListURL from 'indico-url:event_editing.api_editable_list';
 import editablesArchiveURL from 'indico-url:event_editing.api_prepare_editables_archive';
+import assignEditorURL from 'indico-url:event_editing.api_assign_editor';
+import assignSelfEditorURL from 'indico-url:event_editing.api_assign_myself';
+import unassignEditorURL from 'indico-url:event_editing.api_unassign_editor';
 
-import React, {useState} from 'react';
+import React, {useState, useMemo} from 'react';
 import PropTypes from 'prop-types';
 import {useParams} from 'react-router-dom';
-import {Button, Icon, Loader, Checkbox, Message, Search} from 'semantic-ui-react';
+import {Button, Icon, Loader, Checkbox, Message, Search, Dropdown} from 'semantic-ui-react';
 import {Column, Table, SortDirection, WindowScroller} from 'react-virtualized';
 import _ from 'lodash';
 import {
@@ -24,6 +28,7 @@ import {useNumericParam} from 'indico/react/util/routing';
 import {Translate} from 'indico/react/i18n';
 import {useIndicoAxios} from 'indico/react/hooks';
 import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+import {camelizeKeys} from 'indico/utils/case';
 import {userPropTypes} from '../../editing/timeline/util';
 import {EditableType} from '../../models';
 import StateIndicator from '../../editing/timeline/StateIndicator';
@@ -33,15 +38,19 @@ import './EditableList.module.scss';
 export default function EditableList() {
   const eventId = useNumericParam('confId');
   const {type} = useParams();
-  const {data, loading: isLoadingEditableList, lastData} = useIndicoAxios({
+  const {data: editableList, loading: isLoadingEditableList} = useIndicoAxios({
     url: editableListURL({confId: eventId, type}),
     camelize: true,
     trigger: eventId,
   });
-  const editableList = data || lastData;
-  if (isLoadingEditableList && !lastData) {
+  const {data: editors, loading: isLoadingEditors} = useIndicoAxios({
+    url: editorsURL({confId: eventId, type}),
+    camelize: true,
+    trigger: eventId,
+  });
+  if (isLoadingEditableList || isLoadingEditors) {
     return <Loader inline="centered" active />;
-  } else if (!editableList) {
+  } else if (!editableList || !editors) {
     return null;
   }
   const codePresent = Object.values(editableList).some(c => c.code);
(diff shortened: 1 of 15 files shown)


indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -158,36 +206,114 @@ function EditableListDisplay({editableList, codePresent, editableType, eventId})
...
+    } finally {
+      setActiveRequest(null);
+    }
+
+    return camelizeKeys(response.data);
+  };
+
+  const downloadAllFiles = async () => {
+    setActiveRequest('download');
+    const rv = await checkedEditablesRequest(editablesArchiveURL);
```
ParthS007, line 243: We can have a more meaningful name?

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -49,21 +60,46 @@ export default function EditableList() {
...
-  const hasCheckedRows = checked.length === 0;
+  const editables = editableList.filter(x => x.editable);
+  const hasCheckedEditables = checked.length > 0;
+  const checkedSet = new Set(checked);
+  const checkedEditables = editables.filter(x => checkedSet.has(x.editable.id));
+  const [activeRequest, setActiveRequest] = useState(null);
+
+  const editorOptions = useMemo(
+    () =>
+      _.sortBy(editors, 'fullName').map(e => ({
```
ParthS007, line 88: Do we need to case-insensitive sorting here as we have done in case of titles?

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ThiefMaster: `rv` is a common name for return/response value ;)

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ThiefMaster: +-0 on that, usually the first letter of a name is an uppercase latin character or a character where it doesn't really matter whether or not to do it case-insensitively

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -49,21 +60,46 @@ export default function EditableList() {
...
       editableType={type}
       eventId={eventId}
+      editors={editors}
     />
   );
 }
 
-function EditableListDisplay({editableList, codePresent, editableType, eventId}) {
+function EditableListDisplay({
+  editableList: origEditableList,
```
plourenco, line 69: Perhaps we could change this to `initialEditableList` to distinguish from a controlled prop, since we are immediately deriving state from it?

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -49,21 +60,46 @@ export default function EditableList() {
...
+  codePresent,
+  editableType,
+  eventId,
+  editors,
+}) {
   const [sortBy, setSortBy] = useState('friendly_id');
   const [sortDirection, setSortDirection] = useState('ASC');
+  const [editableList, setEditableList] = useState(origEditableList);
   const [sortedList, setSortedList] = useState(editableList);
```
plourenco, line 78: Why do we need both `editableList` and `sortedList`? Wouldn't just using the sorted for the same work?

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -158,36 +206,114 @@ function EditableListDisplay({editableList, codePresent, editableType, eventId})
...
+
+  const downloadAllFiles = async () => {
+    setActiveRequest('download');
+    const rv = await checkedEditablesRequest(editablesArchiveURL);
+    if (rv) {
+      location.href = rv.downloadURL;
+    }
+  };
+
+  const assignEditor = async editor => {
```
plourenco, line 249: What do you think about merging the last three functions? Since we're repeating most of the codes perhaps we could move the `setActiveRequest` and `checkedEditablesRequest` arguments to a method.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -158,36 +206,114 @@ function EditableListDisplay({editableList, codePresent, editableType, eventId})
...
-          <Button disabled={hasCheckedRows} content={Translate.string('Set status')} />
+          <Button.Group>
+            <Dropdown
+              disabled={!hasCheckedEditables || !editors.length || !!activeRequest}
+              options={editorOptions}
+              icon={null}
+              value={null}
+              selectOnBlur={false}
+              selectOnNavigation={false}
+              onChange={(evt, {value}) => {
```
plourenco, line 287:
```suggestion
              onChange={(evt, {value}) => assignEditor(value)}
```

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ThiefMaster: Yes, most likely that would be possible by using `useMemo` for the sorted list instead of putting it in state. But I did not want to refactor this component in this PR. Also because the tricky part with useMemo is invalidating it after patching the data with updated entries.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ThiefMaster: Yeah, even though I'd say it's right on the edge between where deduplicating makes sense or adds more complexity ;)

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ThiefMaster: I don't think the return value of the callback matters so using the braces-less version is indeed a good idea.

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
ThiefMaster: good idea, initial* is also the name used to set the default value of other uncontrolled components

indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -158,36 +206,114 @@ function EditableListDisplay({editableList, codePresent, editableType, eventId})
...
       <ManagementPageSubTitle title={title} />
-      <ManagementPageBackButton url={dashboardURL({confId: eventId})} />
+      <ManagementPageBackButton url={editableTypeURL({confId: eventId, type: editableType})} />
       <div styleName="editable-topbar">
         <div>
-          <Button disabled={hasCheckedRows} content={Translate.string('Assign')} />
-          <Button disabled={hasCheckedRows} content={Translate.string('Unassign')} />
-          <Button disabled={hasCheckedRows} content={Translate.string('Set status')} />
+          <Button.Group>
+            <Dropdown
```
plourenco, line 280: The assignments look really nice! But I noticed that for a high number of editors the dropdown will become very large and sometimes also opening outside of the margins.
I'm wondering if we could give it a fixed size and make it scrollable? 
A "search" function would also look dope but I'm featurizing now 😄

image

ThiefMaster: adding `scrolling` in the dropdown already avoided this ugliness :)

PR #4494, opened 2020-06-16
Disallow groups from receiving editor permissions

### LINKED ISSUE #4491: Disallow adding groups as editors
Especially multipass groups can contain huge amounts of users, and we neither want to disclose those user lists to event managers, nor end up with a huge dropdown from which to select editors.

- Selecting groups when selecting editors shoul be disabled
- We could add a check constraint that ensures these two principal types never have the editor permissions

panagiotappl:
closes #4491

### DIFF
--- a/indico/modules/events/models/principals.py
+++ b/indico/modules/events/models/principals.py
@@ -10,7 +10,7 @@ from __future__ import unicode_literals
 from sqlalchemy.ext.declarative import declared_attr
 
 from indico.core.db import db
-from indico.core.db.sqlalchemy.principals import PrincipalPermissionsMixin
+from indico.core.db.sqlalchemy.principals import PrincipalPermissionsMixin, PrincipalType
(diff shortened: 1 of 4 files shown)


indico/modules/events/models/principals.py
```diff
@@ -27,7 +27,12 @@ class EventPrincipal(PrincipalPermissionsMixin, db.Model):
 
     @declared_attr
     def __table_args__(cls):
-        return auto_table_args(cls, schema='events')
+        group_perm_constraint = db.CheckConstraint(
+            '(type != {} AND type != {}) OR (NOT CAST(permissions AS text[]) && ARRAY[\'paper_editing\','
```
ThiefMaster, line 31: `type NOT IN ({}, {})` and I think `CAST(permission AS text[])` can be simplified to `permissions::text[]`.

PR #4497, opened 2020-06-17
Editable list: implement search functionality

### LINKED ISSUE #4486: Editable list: Implement search functionality
Should search in all relevant columns, but `#xxx` input should filter *only* by ID.

Let's keep it client-side since we fetch all at data once anyway.

depends on #4460

Leats:
Closes #4486 

Bildschirmfoto 2020-06-18 um 10 24 05
Bildschirmfoto 2020-06-18 um 10 24 23

### DIFF
--- a/indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
+++ b/indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
@@ -16,7 +16,7 @@ import unassignEditorURL from 'indico-url:event_editing.api_unassign_editor';
 import React, {useState, useMemo} from 'react';
 import PropTypes from 'prop-types';
 import {useParams} from 'react-router-dom';
-import {Button, Icon, Loader, Checkbox, Message, Search, Dropdown} from 'semantic-ui-react';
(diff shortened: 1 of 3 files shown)


indico/modules/events/editing/client/js/management/editable_type/EditableList.jsx
```diff
@@ -325,31 +388,44 @@ function EditableListDisplay({initialEditableList, codePresent, editableType, ev
...
+                      disabled={!!activeRequest}
                     />
                   )}
+                  cellRenderer={({rowIndex}) =>
+                    sortedList[rowIndex].editable && (
+                      <Checkbox
+                        disabled={!!activeRequest || !filteredSet.has(sortedList[rowIndex].id)}
+                        checked={
+                          sortedList[rowIndex].editable
+                            ? checked.includes(sortedList[rowIndex].editable.id)
```
ThiefMaster, line 421: We can use `checkedSet.has(sortedList[rowIndex].editable.id)` here as well... it was already using `includes` before but it makes sense to use the Set here..

Same for the other place where we currently use `checked.includes`.

PR #4498, opened 2020-06-17
Allow resetting registration moderation status

OmeGak:
This PR introduces the possibility for managers to reset the approval state of a given registration from the registration details view. The button is displayed as part of available actions for "complete" registrations only if registration form moderation is enabled.

## Screenshot

image

### DIFF
--- a/indico/modules/events/registration/controllers/management/reglists.py
+++ b/indico/modules/events/registration/controllers/management/reglists.py
@@ -565,6 +565,20 @@ class RHRegistrationReject(RHManageRegistrationBase):
         return jsonify_data(html=_render_registration_details(self.registration))
 
 
+class RHRegistrationReset(RHManageRegistrationBase):
+    """Reset a registration back to a non-approved status"""
+
+    def _process(self):
+        if self.registration.state in (RegistrationState.complete, RegistrationState.unpaid):
+            self.registration.update_state(approved=False)
+        elif self.registration.state == RegistrationState.rejected:
+            self.registration.update_state(rejected=False)
+        else:
+            raise BadRequest(_('The registration cannot be reset in its current state.'))
+        logger.info('Registration %r was reset by %r', self.registration, session.user)
+        return jsonify_data(html=_render_registration_details(self.registration))
+
+
 class RHRegistrationCheckIn(RHManageRegistrationBase):
     """Set checked in state of a registration"""
 
--- a/indico/modules/events/registration/models/registrations.py
+++ b/indico/modules/events/registration/models/registrations.py
@@ -470,7 +470,7 @@ class Registration(db.Model):
         moderation_required = (regform.moderation_enabled and not _skip_moderation and
                                (not invitation or not invitation.skip_moderation))
         with db.session.no_autoflush:
-            payment_required = regform.event.has_feature('payment') and self.price
+            payment_required = regform.event.has_feature('payment') and bool(self.price)
         if self.state == RegistrationState.pending:
             if approved and payment_required:
                 self.state = RegistrationState.unpaid
@@ -488,6 +488,11 @@ class Registration(db.Model):
                 self.state = RegistrationState.pending
             elif paid is False and payment_required:
                 self.state = RegistrationState.unpaid
+        elif self.state == RegistrationState.rejected:
+            if rejected is False and moderation_required:
+                self.state = RegistrationState.pending
+            elif rejected is False and not moderation_required:
+                self.state = RegistrationState.complete
         if self.state != initial_state:
             signals.event.registration_state_updated.send(self, previous_state=initial_state)
 
--- a/indico/modules/events/registration/models/registrations.py
+++ b/indico/modules/events/registration/models/registrations.py
@@ -491,7 +491,9 @@ class Registration(db.Model):
         elif self.state == RegistrationState.rejected:
             if rejected is False and moderation_required:
                 self.state = RegistrationState.pending
-            elif rejected is False and not moderation_required:
+            elif rejected is False and payment_required:
+                self.state = RegistrationState.unpaid
+            elif rejected is False:
                 self.state = RegistrationState.complete
         if self.state != initial_state:
             signals.event.registration_state_updated.send(self, previous_state=initial_state)
--- a/indico/modules/events/registration/templates/management/_registration_details.html
+++ b/indico/modules/events/registration/templates/management/_registration_details.html
@@ -189,12 +189,12 @@
 
 {% macro _render_reset_registration_button(registration, action_text) %}
     <a class="i-button warning"
-        data-update="#registration-details"
-        data-method="POST"
-        data-href="{{ url_for('.reset_registration', registration) }}"
-        data-confirm="{% trans %}Are you sure you want to reset this registration status? This action will not be notified to the registrant.{% endtrans %}"
-        data-title="{{ action_text }}">
-        {{ action_text }}
+       data-update="#registration-details"
+       data-method="POST"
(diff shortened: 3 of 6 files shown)


indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -565,6 +565,16 @@ def _process(self):
         return jsonify_data(html=_render_registration_details(self.registration))
 
 
+class RHRegistrationReset(RHManageRegistrationBase):
+    """Reset a registration back to a non-approved status"""
+
+    def _process(self):
+        self.registration.update_state(approved=False)
+        db.session.flush()
+        logger.info('Registration %s was reset by %s', self.registration, session.user)
```
ThiefMaster, line 574: `%r` for both format placeholders (even though `%s` does the same here)

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -565,6 +565,16 @@ def _process(self):
         return jsonify_data(html=_render_registration_details(self.registration))
 
 
+class RHRegistrationReset(RHManageRegistrationBase):
+    """Reset a registration back to a non-approved status"""
+
+    def _process(self):
```
ThiefMaster, line 571: Shouldn't we also check if moderation is enabled? Or are we already ignoring an `approved=False` state update in that case?

indico/modules/events/registration/controllers/management/reglists.py
ThiefMaster, line 574: I think logging this action to the event log (`event.log(...)`) would be useful (let me guess, it's missing for all the other actions as well? ;))

indico/modules/events/registration/controllers/management/reglists.py
OmeGak: I don't think it's necessary to check for moderation on the RH, unless you want to raise an explicit error. `update_state` will not do anything if moderation is not enabled:

```python
elif self.state == RegistrationState.complete:
    if approved is False and payment_required is False and moderation_required:
        self.state = RegistrationState.pending
```

ThiefMaster: Fine for me then since we hide the button anyway if moderation is not enabled.

indico/modules/events/registration/controllers/management/reglists.py
OmeGak: `update_state()` triggers the `registration_state_updated` signal, which is connected to [`log_registration_updated`](https://github.com/indico/indico/blob/master/indico/modules/events/registration/logging.py#L33) and effectively registers all registration state changes on the event log.

ThiefMaster: forgot about that ;) never mind th en

indico/modules/events/registration/templates/management/_registration_details.html
```diff
@@ -150,11 +150,23 @@
...
+    {% if can_reset or can_get_ticket %}
         <div class="toolbar">
-            <a href="{{ url_for('.ticket_download', registration.locator.registrant) }}" class="i-button accept icon-ticket">
-                {% trans %}Get ticket{% endtrans %}
-            </a>
+            {% if can_reset %}
+                <a class="i-button warning"
+                    data-update="#registration-details"
+                    data-method="POST"
+                    data-href="{{ url_for('.reset_registration', registration) }}">
```
ThiefMaster, line 161: one space too much indentation

ThiefMaster:
Resetting a rejected or withdrawn (even if there's no UI for the latter) registration is not possible yet

OmeGak:
> Resetting a rejected or withdrawn (even if there's no UI for the latter) registration is not possible yet

Rejected state is now properly handled.

indico/modules/events/registration/models/registrations.py
```diff
@@ -488,6 +488,11 @@ def update_state(self, approved=None, paid=None, rejected=None, _skip_moderation
                 self.state = RegistrationState.pending
             elif paid is False and payment_required:
                 self.state = RegistrationState.unpaid
+        elif self.state == RegistrationState.rejected:
+            if rejected is False and payment_required is False and moderation_required:
+                self.state = RegistrationState.pending
+            if rejected is False and payment_required:
+                self.state = RegistrationState.unpaid
```
ThiefMaster, line 495: Payment happens only after approval, so I don't think rejected=>unpaid can happen. Resetting rejected should always go back to pending. The only edge case where you would not go back to pending is if moderation has been disabled in the meantime - in that case you'd go to complete.

indico/modules/events/registration/models/registrations.py
```diff
@@ -488,6 +488,11 @@ def update_state(self, approved=None, paid=None, rejected=None, _skip_moderation
                 self.state = RegistrationState.pending
             elif paid is False and payment_required:
                 self.state = RegistrationState.unpaid
+        elif self.state == RegistrationState.rejected:
+            if rejected is False and payment_required is False and moderation_required:
+                self.state = RegistrationState.pending
+            if rejected is False and payment_required:
```
ThiefMaster, line 494: `elif`

indico/modules/events/registration/templates/management/_registration_details.html
```diff
@@ -91,7 +91,9 @@
             <a class="i-button accept"
                data-update="#registration-details"
                data-method="POST"
-               data-href="{{ url_for('.approve_registration', registration) }}">
+               data-href="{{ url_for('.approve_registration', registration) }}"
+               data-confirm="{% trans %}Are you sure you want to approve this registration? This will trigger a notification email.{% endtrans %}"
+               data-title="{% trans %}Approve registration{% endtrans %}">
```
ThiefMaster, line 96: Is a confirmation prompt for what's probably the most common action really needed, especially now that it can be undone in case someone does misclick?

OmeGak: I think it's generally a good idea to let managers know that an action will trigger an email. But now, precisely because approvals can be undone, there is the risk for managers to inadvertently triggering multiple emails. There could be better ways to hint that an email will be sent, but I think this is better than nothing. At the same time, I don't have a very strong opinion on this. Let me know if I should remove it.

pferreir: I don't have a very strong opinion either. But then it would be logical that we add this warning also when we bulk-approve registrations.

OmeGak: Reworded the already-existing confirmation dialog for the bulk-moderation actions. 👌

indico/modules/events/registration/models/registrations.py
```diff
@@ -488,6 +488,11 @@ def update_state(self, approved=None, paid=None, rejected=None, _skip_moderation
                 self.state = RegistrationState.pending
             elif paid is False and payment_required:
                 self.state = RegistrationState.unpaid
+        elif self.state == RegistrationState.rejected:
+            if rejected is False and payment_required is False and moderation_required:
```
ThiefMaster, line 492: the `payment_required is False` check is not needed/correct here.

PS: can you add `bool()` around `self.price` above? while debugging i realized that `payment_required` is the price and not `True` ;x

ThiefMaster:
When the registration is in the `unpaid` state the reset approval button is not visible.
After marking it as paid, it shows up, but clicking it doesn't do anything...

OmeGak:
Fixed:
- The button is now hidden when a registration is complete and paid.
- The button is now displayed when a registration is unpaid.

PR #4501, opened 2020-06-22
Update new Editing view

### LINKED ISSUE #4484: Editing: Editor dashboard in the hybrid view
This is basically a list of editables, similar to the one from #4460.

panagiotappl:
closes #4484 

Changes:
- Looks
- Only Timeline is wrapped by redux store (instead of entire EditingView)
- Added router to handle links from timeline to editable list
- Information about editable type, contribution and event are taken from router params and not WP

Issues:
- How to get `contributionCode` now that `contributionId` is taken from url parameters? Use new endpoint? (currently contribution code is hardcoded)

Timeline view:

Editable list view:

### DIFF
--- a/indico/modules/events/editing/client/js/editing/index.jsx
+++ b/indico/modules/events/editing/client/js/editing/index.jsx
@@ -5,18 +5,17 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
-import fileTypesURL from 'indico-url:event_editing.api_file_types';
-import editableDetailsURL from 'indico-url:event_editing.api_editable';
-import tagsURL from 'indico-url:event_editing.api_tags';
+import timelineURL from 'indico-url:event_editing.editable';
+import editableTypeListURL from 'indico-url:event_editing.editable_type_list';
 
 import React from 'react';
 import ReactDOM from 'react-dom';
-
-import {camelizeKeys} from 'indico/utils/case';
-import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+import {BrowserRouter as Router, Route, Switch} from 'react-router-dom';
+import {routerPathFromFlask} from 'indico/react/util/routing';
 
 import EditingView from './page_layout';
 import ReduxTimeline from './ReduxTimeline';
+import {EditableList} from '../management/editable_type';
 
 document.addEventListener('DOMContentLoaded', async () => {
   const editingElement = document.querySelector('#editing-view');
@@ -30,44 +29,30 @@ document.addEventListener('DOMContentLoaded', async () => {
     document.querySelector('div.main-breadcrumb').getBoundingClientRect().height;
   document.body.style.setProperty('--header-height', headerHeight);
 
-  const eventId = parseInt(editingElement.dataset.eventId, 10);
   const eventTitle = editingElement.dataset.eventTitle;
-  const contributionId = parseInt(editingElement.dataset.contributionId, 10);
-  const contributionCode = editingElement.dataset.contributionCode;
-  const editableType = editingElement.dataset.editableType;
-
-  let fileTypes, tags;
-
-  try {
-    const [fileTypeResponse, tagResponse] = await Promise.all([
-      indicoAxios.get(fileTypesURL({confId: eventId, type: editableType})),
-      indicoAxios.get(tagsURL({confId: eventId})),
-    ]);
-    fileTypes = camelizeKeys(fileTypeResponse.data);
-    tags = camelizeKeys(tagResponse.data);
-  } catch (e) {
-    handleAxiosError(e);
-    return;
-  }
-
-  const storeData = {
-    eventId,
-    contributionId,
-    contributionCode,
-    editableType,
-    fileTypes,
-    tags,
-    editableDetailsURL: editableDetailsURL({
-      confId: eventId,
-      contrib_id: contributionId,
-      type: editableType,
-    }),
-  };
 
   ReactDOM.render(
-    <EditingView eventId={eventId} eventTitle={eventTitle} editableType={editableType}>
-      <ReduxTimeline storeData={storeData} />
-    </EditingView>,
+    <Router>
+      <Route
+        path={[
+          routerPathFromFlask(timelineURL, ['confId', 'contrib_id', 'type']),
+          routerPathFromFlask(editableTypeListURL, ['confId', 'type']),
+        ]}
+      >
+        <EditingView eventTitle={eventTitle}>
+          <Switch>
+            <Route
+              exact
+              path={routerPathFromFlask(timelineURL, ['confId', 'contrib_id', 'type'])}
+              component={ReduxTimeline}
+            />
+            <Route exact path={routerPathFromFlask(editableTypeListURL, ['confId', 'type'])}>
+              <EditableList management={false} />
+            </Route>
+          </Switch>
+        </EditingView>
+      </Route>
+    </Router>,
     editingElement
   );
 });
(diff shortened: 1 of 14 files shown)


indico/modules/events/editing/client/js/editing/index.jsx
```diff
@@ -33,52 +29,31 @@ document.addEventListener('DOMContentLoaded', async () => {
...
-        <Timeline />
-      </EditingView>
-    </Provider>,
+    <Router>
+      <Switch>
+        <Route
+          exact
+          path={routerPathFromFlask(timelineURL, ['confId', 'contrib_id', 'type'])}
+          component={() => (
+            <EditingView eventTitle={eventTitle}>
```
ThiefMaster, line 41: I'd keep this outside the Switch/Router to avoid repeating it for every Route.

indico/modules/events/editing/client/js/editing/index.jsx
```diff
@@ -33,52 +29,31 @@ document.addEventListener('DOMContentLoaded', async () => {
...
-      <EditingView eventId={eventId} eventTitle={eventTitle}>
-        <Timeline />
-      </EditingView>
-    </Provider>,
+    <Router>
+      <Switch>
+        <Route
+          exact
+          path={routerPathFromFlask(timelineURL, ['confId', 'contrib_id', 'type'])}
+          component={() => (
```
ThiefMaster, line 40: Even though this works, `component` is meant to be used with just that, e.g. `component={ReduxTimeline}`. But since you don't rely on any render props, you can just pass the JSX you return here as children to the `Route`.

indico/modules/events/editing/client/js/editing/index.jsx
ThiefMaster: and by doing that, you can use `component={ReduxTimeline}` above ;)

panagiotappl: I need `MenuBar` to be inside router since it acts as a navigation bar, though.

ThiefMaster: Try this structure:

```
<Router>
  <EditingView>
    <Switch>
      <Route>
      <Route>
```

panagiotappl: I tried this but I'll try it again.

indico/modules/events/editing/client/js/editing/page_layout/MenuBar.jsx
```diff
@@ -6,58 +6,78 @@
...
+      </Menu.Item>
+    );
+  };
+
   return (
     <div styleName="menu-bar">
       <Header as="h2" styleName="header">
-        <Translate>Paper Editing</Translate>
+        <Translate>
+          <Param name="editableType" value={EditableTypeTitles[type]} /> Editing
```
ThiefMaster, line 43: This won't work well for any language where the translated word for Editing doesn't come *before* the type.. better to use a dict mapping types to complete translatable strings instead.

indico/modules/events/editing/client/js/editing/page_layout/MenuBar.jsx
```diff
@@ -6,58 +6,78 @@
...
+        </Translate>
       </Header>
-      <ul styleName="list">
-        <li>
-          <span styleName="capitalized">
+      <Menu vertical>{renderEditableListMenu()}</Menu>
```
ThiefMaster, line 46: considering how simple `renderEditableListMenu` is I'd move its content here

panagiotappl: I'd prefer it in a separate function to avoid writing `if` code inside jsx twice.

ThiefMaster: maybe worth making it a separate component (still in this same file) then?

indico/modules/events/editing/client/js/editing/page_layout/MenuBar.jsx
```diff
@@ -6,58 +6,85 @@
...
-              </li>
-            </a>
-          </ul>
-        </li>
-      </ul>
+        </Menu.Item>
+        <Menu.Item name="display">
+          <a href={displayURL({confId: eventId})}>
+            <Icon name="tv" />
+            <Translate>Display</Translate>
```
ThiefMaster, line 78: Can you add a space between those two elements (same below)? Right now the text and icon don't have any space between which looks a bit weird. Replacing the linebreak with a space should work (not sure if prettier prefers that or `{' '}` here)

indico/modules/events/editing/controllers/frontend.py
```diff
@@ -37,6 +39,16 @@ def _process(self):
...
             self.event,
-            editable=self.editable,
-            contribution=self.contrib
+        )
+
+
+class RHEditableTypeList(RHEditingBase):
+    def _process_args(self):
+        RHEditingBase._process_args(self)
+        self.editable_type = EditableType[request.view_args['type']]
```
ThiefMaster, line 48: Is this used anywhere?

indico/modules/events/editing/client/js/editing/page_layout/MenuBar.jsx
```diff
@@ -6,58 +6,83 @@
...
-            </a>
-          </ul>
-        </li>
-      </ul>
+        </Menu.Item>
+        <Menu.Item name="display">
+          <a href={displayURL({confId: eventId})}>
+            <Icon name="tv" /> <Translate>Display</Translate>
+          </a>
+        </Menu.Item>
```
ThiefMaster, line 79: I just noticed that you have to click the text itself, instead of anywhere on the menu item. This can be avoided by making the menu item itself the link. but it needs some extra CSS to keep the usual link color:

```jsx
        <Menu.Item name="display" as="a" href={displayURL({confId: eventId})}>
          <span>
            <Icon name="tv" /> <Translate>Display</Translate>
          </span>
        </Menu.Item>
```

(similar for all the other menu items)

indico/modules/events/editing/client/js/editing/page_layout/MenuBar.jsx
```diff
@@ -6,58 +6,83 @@
...
+          <Translate>Editable list</Translate>
+        </Link>
+      </Menu.Item>
+    </Menu>
+  );
+}
+
+EditableListMenu.propTypes = {
+  eventId: PropTypes.number.isRequired,
+  type: PropTypes.string.isRequired,
```
ThiefMaster, line 44: `editableType`?

and i'd use `PropTypes.oneOf(...).isRequired` since we have a well-defined list of editable types

indico/modules/events/editing/controllers/frontend.py
```diff
@@ -37,6 +38,12 @@ def _process(self):
         return WPEditingView.render_template(
             'editing.html',
             self.event,
-            editable=self.editable,
-            contribution=self.contrib
+        )
+
+
+class RHEditableTypeList(RHEditingBase):
```
ThiefMaster, line 44: We need a permission check in this RH. `RHEditablesList` (the underlying API) probably needs to be updated as well since it currently requires management access. But editors who aren't managers can still use the new view, so the API must let them access it as well.

indico/modules/events/editing/client/js/editing/timeline/selectors.js
```diff
@@ -173,14 +173,15 @@ export const getStaticData = state => state.staticData;
 
 export const getFileTypes = createSelector(
   getStaticData,
-  staticData => {
+  getDetails,
```
plourenco, line 176: Isn't this a dangerous change?

There's an exception when I click `EditableSubmissionButton` because it also uses `getFileTypes` to pass it as prop to the `FileManager` but I don't think the store contains any timeline details, since we're passing the contribution code as `staticData`. Perhaps we could find a balance here?

panagiotappl: That's a nice observation I didn't notice, thanks. I'll find a way to fix this.

PR #4503, opened 2020-06-23
Allow filtering event log by related entries

OmeGak:
Event log filtering by related entries relies on the new `internal_data` column added to the `event.logs` table. For now, the only logged events that use this new column are those related to registrations.

## Screenshots
||
|-|
|image|

||
|-|
|image|

### DIFF
--- a/indico/modules/events/logs/client/js/actions.js
+++ b/indico/modules/events/logs/client/js/actions.js
@@ -5,6 +5,8 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
+import qs from 'qs';
+
 import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
 
 export const SET_KEYWORD = 'SET_KEYWORD';
@@ -36,6 +38,32 @@ export function setDetailedView(entryIndex) {
 
 export function setMetadataQuery(metadataQuery) {
   return {type: SET_METADATA_QUERY, metadataQuery};
+}
+
+export function clearMetadataQuery() {
+  return dispatch => {
+    dispatch(setMetadataQuery({}));
+    dispatch(setPage(1));
+    dispatch(setDetailedView(null));
+    dispatch(fetchLogEntries());
+    history.replaceState({}, null, location.pathname);
+  };
+}
+
+export function showRelatedEntries() {
+  return (dispatch, getStore) => {
+    const {
+      logs: {entries, currentViewIndex},
+    } = getStore();
+
+    const entry = entries[currentViewIndex];
+    const queryString = qs.stringify({meta: entry.meta}, {allowDots: true});
+    dispatch(setMetadataQuery(entry.meta));
+    dispatch(setPage(1));
+    dispatch(setDetailedView(null));
+    dispatch(fetchLogEntries());
+    history.replaceState({}, null, `${location.pathname}?${queryString}`);
+  };
 }
 
 export function viewPrevEntry() {
--- a/indico/modules/events/logs/client/js/actions.js
+++ b/indico/modules/events/logs/client/js/actions.js
@@ -131,7 +131,7 @@ export function fetchLogEntries() {
     const params = {
       page: currentPage,
       filters: [],
-      data: JSON.stringify(metadataQuery),
+      meta: metadataQuery,
     };
     if (keyword) {
       params.q = keyword;
@@ -145,7 +145,10 @@ export function fetchLogEntries() {
 
     let response;
     try {
-      response = await indicoAxios.get(fetchLogsUrl, {params});
+      response = await indicoAxios.get(fetchLogsUrl, {
+        params,
+        paramsSerializer: p => qs.stringify(p, {arrayFormat: 'repeat', allowDots: true}),
+      });
     } catch (error) {
       handleAxiosError(error);
       dispatch(fetchFailed());
--- a/indico/modules/events/logs/client/js/actions.js
+++ b/indico/modules/events/logs/client/js/actions.js
@@ -16,6 +16,7 @@ export const FETCH_FAILED = 'FETCH_FAILED';
 export const SET_DETAILED_VIEW = 'SET_DETAILED_VIEW';
 export const VIEW_PREV_ENTRY = 'VIEW_PREV_ENTRY';
 export const VIEW_NEXT_ENTRY = 'VIEW_NEXT_ENTRY';
+export const SET_METADATA_QUERY = 'SET_METADATA_QUERY';
 
 export function setKeyword(keyword) {
   return {type: SET_KEYWORD, keyword};
@@ -31,6 +32,10 @@ export function setPage(currentPage) {
 
 export function setDetailedView(entryIndex) {
   return {type: SET_DETAILED_VIEW, currentViewIndex: entryIndex};
+}
+
+export function setMetadataQuery(metadataQuery) {
+  return {type: SET_METADATA_QUERY, metadataQuery};
 }
 
 export function viewPrevEntry() {
@@ -90,15 +95,15 @@ export function fetchFailed() {
 export function fetchLogEntries() {
   return async (dispatch, getStore) => {
     dispatch(fetchStarted());
-
     const {
-      logs: {filters, keyword, currentPage},
+      logs: {filters, keyword, currentPage, metadataQuery},
       staticData: {fetchLogsUrl},
     } = getStore();
 
     const params = {
       page: currentPage,
       filters: [],
+      data: JSON.stringify(metadataQuery),
     };
     if (keyword) {
       params.q = keyword;
--- a/indico/modules/events/logs/client/js/components/EventLog.jsx
+++ b/indico/modules/events/logs/client/js/components/EventLog.jsx
@@ -6,13 +6,46 @@
 // LICENSE file for more details.
 
 import React from 'react';
+import {useDispatch, useSelector} from 'react-redux';
+
+import {Param, Translate} from 'indico/react/i18n';
 
 import LogEntryList from '../containers/LogEntryList';
 import Toolbar from './Toolbar';
+import {fetchLogEntries, setMetadataQuery, setPage} from '../actions';
+
+function MetadataQueryMessage() {
+  const dispatch = useDispatch();
+  const resetURLFiltering = () => {
+    dispatch(setMetadataQuery({}));
+    dispatch(setPage(1));
+    dispatch(fetchLogEntries());
+    history.replaceState({}, null, location.pathname);
+  };
+
+  return (
+    <div className="highlight-message-box">
+      <span className="icon" />
+      <div className="message-text">
+        <Translate>
+          Log entries are currently filtered by URL.{' '}
+          <Param name="link" wrapper={<a onClick={resetURLFiltering} />}>
+            Click here
+          </Param>{' '}
+          to disable the filter.
+        </Translate>
+      </div>
+    </div>
+  );
+}
+
+MetadataQueryMessage.propTypes = {};
 
 export default function EventLog() {
+  const metadataQuery = useSelector(state => state.logs.metadataQuery);
   return (
     <>
+      {Object.keys(metadataQuery).length !== 0 && <MetadataQueryMessage />}
       <Toolbar />
       <LogEntryList />
     </>
--- a/indico/modules/events/logs/client/js/components/EventLog.jsx
+++ b/indico/modules/events/logs/client/js/components/EventLog.jsx
@@ -12,15 +12,12 @@ import {Param, Translate} from 'indico/react/i18n';
 
 import LogEntryList from '../containers/LogEntryList';
 import Toolbar from './Toolbar';
-import {fetchLogEntries, setMetadataQuery, setPage} from '../actions';
+import {clearMetadataQuery} from '../actions';
 
 function MetadataQueryMessage() {
   const dispatch = useDispatch();
   const resetURLFiltering = () => {
-    dispatch(setMetadataQuery({}));
-    dispatch(setPage(1));
-    dispatch(fetchLogEntries());
-    history.replaceState({}, null, location.pathname);
+    dispatch(clearMetadataQuery());
   };
 
   return (
--- a/indico/modules/events/logs/client/js/components/LogEntryModal.jsx
+++ b/indico/modules/events/logs/client/js/components/LogEntryModal.jsx
@@ -6,7 +6,6 @@
 // LICENSE file for more details.
 
 import moment from 'moment';
-import qs from 'qs';
 import React from 'react';
 import PropTypes from 'prop-types';
 
@@ -66,11 +65,8 @@ export default class LogEntryModal extends React.Component {
   }
 
   relatedEntries() {
-    const {currentViewIndex, entries, relatedEntries} = this.props;
-    const entry = entries[currentViewIndex];
-    const queryString = qs.stringify({meta: entry.meta}, {allowDots: true});
-    relatedEntries(entry.meta);
-    history.replaceState({}, null, `${location.pathname}?${queryString}`);
+    const {relatedEntries} = this.props;
+    relatedEntries();
   }
 
   _isFirstEntry() {
--- a/indico/modules/events/logs/client/js/components/LogEntryModal.jsx
+++ b/indico/modules/events/logs/client/js/components/LogEntryModal.jsx
@@ -6,6 +6,7 @@
 // LICENSE file for more details.
 
 import moment from 'moment';
+import qs from 'qs';
 import React from 'react';
 import PropTypes from 'prop-types';
 
@@ -20,6 +21,7 @@ export default class LogEntryModal extends React.Component {
     setDetailedView: PropTypes.func.isRequired,
     prevEntry: PropTypes.func.isRequired,
     nextEntry: PropTypes.func.isRequired,
+    relatedEntries: PropTypes.func.isRequired,
     currentPage: PropTypes.number.isRequired,
     totalPageCount: PropTypes.number.isRequired,
   };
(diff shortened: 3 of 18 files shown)


indico/modules/events/logs/models/entries.py
```diff
@@ -96,6 +97,11 @@ class EventLogEntry(db.Model):
         JSON,
         nullable=False
     )
+    #: Non-displayable data
+    internal_data = db.Column(
+        JSON,
```
ThiefMaster, line 102: Let's use JSONB here so we can use an index. JSON is stored as opaque text and won't be indexed. The only reason `data` is still JSON is that converting it would have taken ages in our own database.

indico/modules/events/models/events.py
```diff
@@ -843,7 +844,7 @@ def log(self, realm, kind, module, summary, user=None, type_='simple', data=None
         if self.__logging_disabled:
             return
         entry = EventLogEntry(user=user, realm=realm, kind=kind, module=module, type=type_, summary=summary,
-                              data=data or {})
+                              data=data or {}, internal_data=internal_data or {})
```
ThiefMaster, line 847:
```suggestion
                              data=(data or {}), internal_data=(internal_data or {}))
```

No technical difference but kwarg values with spaces look very weird without parentheses.

indico/migrations/versions/20200623_1718_05f227f4b938_add_internal_data_to_event_logs.py
```diff
@@ -0,0 +1,30 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '05f227f4b938'
+down_revision = 'c0fc1e46888b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('logs', sa.Column('internal_data', postgresql.JSON(astext_type=sa.Text()), nullable=False,
```
ThiefMaster, line 21: I think the `astext_type` is not relevant here since it doesn't affect the DDL

indico/migrations/versions/20200623_1718_05f227f4b938_add_internal_data_to_event_logs.py
```diff
@@ -0,0 +1,30 @@
...
+                                    server_default='null'), schema='events')
+    op.alter_column('logs', 'internal_data', server_default=None, schema='events')
+    op.create_index(op.f('ix_logs_internal_data'), 'logs', [sa.text('internal_data')], unique=False, schema='events',
```
ThiefMaster, line 24: `None` for the index name since it's automatically generated

indico/migrations/versions/20200623_1718_05f227f4b938_add_internal_data_to_event_logs.py
```diff
@@ -0,0 +1,30 @@
...
+                    postgresql_using=u'gin')
```
ThiefMaster, line 25: no need for that `u` prefix here

indico/migrations/versions/20200623_1718_05f227f4b938_add_internal_data_to_event_logs.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def downgrade():
+    op.drop_index(op.f('ix_logs_internal_data'), table_name='logs', schema='events')
```
ThiefMaster, line 29: I think we don't need this since dropping the column will already drop the index as well

indico/migrations/versions/20200623_1718_05f227f4b938_add_internal_data_to_event_logs.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision = '05f227f4b938'
+down_revision = 'c0fc1e46888b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('logs', sa.Column('internal_data', postgresql.JSON(astext_type=sa.Text()), nullable=False,
+                                    server_default='null'), schema='events')
```
ThiefMaster, line 22: `null` or `'{}'`? Since we default to `{}` when creating a new log entry I think using `{}` for existing rows would make more sense..

indico/modules/events/logs/controllers.py
```diff
@@ -37,9 +39,10 @@ class RHEventLogsJSON(RHManageEventBase):
     def _process(self):
         page = int(request.args.get('page', 1))
         filters = request.args.getlist('filters')
+        data = json.loads(request.args.get('data'))
```
ThiefMaster, line 42: This fails if the param is missing, since `json.loads(None)` won't work

indico/modules/events/logs/client/js/actions.js
```diff
@@ -33,6 +34,10 @@ export function setDetailedView(entryIndex) {
   return {type: SET_DETAILED_VIEW, currentViewIndex: entryIndex};
 }
 
+export function setDataQuery(query) {
+  return {type: SET_DATA_QUERY, dataQuery: query};
```
ThiefMaster, line 38:
```suggestion
export function setDataQuery(dataQuery) {
  return {type: SET_DATA_QUERY, dataQuery};
```

indico/modules/events/logs/client/js/actions.js
```diff
@@ -90,15 +95,15 @@ export function fetchFailed() {
...
     const {
-      logs: {filters, keyword, currentPage},
+      logs: {filters, keyword, currentPage, dataQuery},
       staticData: {fetchLogsUrl},
     } = getStore();
 
     const params = {
       page: currentPage,
       filters: [],
+      data: JSON.stringify(dataQuery)
```
ThiefMaster, line 106: I'm very surprised that CI did not complain about a missing trailing `,` here...

indico/modules/events/logs/client/js/components/EventLog.jsx
```diff
@@ -6,13 +6,45 @@
...
 
 import LogEntryList from '../containers/LogEntryList';
 import Toolbar from './Toolbar';
+import {fetchLogEntries, setDataQuery, setPage} from '../actions';
+
+function DataQueryMessage() {
+  const dispatch = useDispatch();
+  const dataQuery = useSelector(state => state.logs.dataQuery);
+
+  function resetUrlFiltering() {
```
ThiefMaster, line 21: `URL` please 🙈 

and in most places we use arrow functions when nesting functions inside other functions

indico/modules/events/logs/client/js/components/EventLog.jsx
```diff
@@ -6,13 +6,45 @@
...
+  if (Object.keys(dataQuery).length === 0) {
+    return null;
+  }
+  return (
+    <>
+      <div className="highlight-message-box">
+        <span className="icon"></span>
+        <div className="message-text">
+          <Translate>
+            Some entries may currently be filtered by URL. <Param name="link" wrapper={<a onClick={resetUrlFiltering} />}>Click here</Param> to disable URL filtering.
```
ThiefMaster, line 36: I don't like the term "url filtering" - it's not really clear what that's supposed to mean! Can we find a better word for it?

> Log entries are currently filtered. Click hear to [clear the filter].

What about something like this? Even though it might be a bit misleading because there's this filter and the search box... which you wouldn't want to clear by clicking here.

OmeGak: > Log entries **are** currently filtered. Click here to [clear the filter].

We don't know for sure if entries are filtered in the current implementation. I would still keep __"may be"__.

> Even though it might be a bit misleading because there's this filter and the search box... which you wouldn't want to clear by clicking here.

Exactly. I would rather make a clear distinction between the two kinds of filtering. How about something like this?

> Log entries may be currently filtered by URL. Click here to clear the filter.

"Filtered by URL" is more semantically accurate than "URL filtering" and, although tech-savvy, when looking at the URL (`?filter.xyz=...`) one can imagine what is going on.

ThiefMaster: I think they are always being filtered, even if the filter happens to match everything. But that is actually impossible because "event created" will never show up with such a filter active.

indico/migrations/versions/20200623_1718_05f227f4b938_add_internal_data_to_event_logs.py
```diff
@@ -0,0 +1,28 @@
...
+down_revision = 'b6dd0a4ed40d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('logs', sa.Column('internal_data', postgresql.JSONB(), nullable=False, server_default='{}'),
+                  schema='events')
+    op.alter_column('logs', 'internal_data', server_default=None, schema='events')
+    op.create_index(None, 'logs', [sa.text('internal_data')], unique=False, schema='events', postgresql_using='gin')
```
ThiefMaster, line 24: the `sa.text(...)` around `'internal_data'` is not necessary

indico/modules/events/logs/client/js/actions.js
```diff
@@ -33,6 +34,10 @@ export function setDetailedView(entryIndex) {
   return {type: SET_DETAILED_VIEW, currentViewIndex: entryIndex};
 }
 
+export function setDataQuery(dataQuery) {
+  return {type: SET_DATA_QUERY, dataQuery};
```
pferreir, line 38: "Data query" seems like a very broad name to me. I know you probably want to avoid "filter" since that's already used for the regular filters in the event log. Maybe "metadataQuery" would make it a bit more intuitive?

indico/modules/events/logs/templates/entry_simple.html
```diff
@@ -34,4 +34,14 @@
             </td>
         </tr>
     {% endfor %}
+    {% if data_query %}
+        <tr class="i-table content">
+            <td class="i-table caption log-caption"></td>
+            <td class="i-table value">
+                <a href="{{ url_for('.index', entry.event, **data_query) }}">
```
ThiefMaster, line 41: I think this should be done in the react app, not in the template. That way you:

- avoid the full page reload
- integrate nicely with the rest of the UI

Maybe a "Related log entries" button between previous/next in the modal? Unless you have a better idea...

indico/modules/events/logs/controllers.py
```diff
@@ -29,17 +31,20 @@ class RHEventLogs(RHManageEventBase):
     """Shows the modification/action log for the event"""
 
     def _process(self):
+        data_query = {k[len('filter.'):]: int(v) if v.isdigit() else v
```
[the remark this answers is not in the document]
pferreir: And maybe using `filter.` here will generate confusion, since we're calling "filters" something else?

ThiefMaster:
I noticed that we are not attaching the registration id when logging emails related to a registration. I thing doing it there too would be a good idea.

You'll have to add a way to pass-through internal data (maybe call it `log_data` there?) in `send_email` and `_log_email`; the rest should be straightforward.

indico/migrations/versions/20200623_1718_05f227f4b938_add_internal_data_to_event_logs.py
```diff
@@ -0,0 +1,28 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '05f227f4b938'
+down_revision = 'b6dd0a4ed40d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('logs', sa.Column('internal_data', postgresql.JSONB(), nullable=False, server_default='{}'),
```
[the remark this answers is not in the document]
ThiefMaster: Just `metadata` is reserved by SQLAlchemy, but `meta` would work. I actually like this, especially because we could then use `meta.FIELD=...` in the URL!

indico/modules/events/logs/controllers.py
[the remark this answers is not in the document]
ThiefMaster: I had suggested on IRC to rename our filters since we call those realms everywhere else. But with your `meta` suggestion I think `meta` would fit nicely here as well.

indico/modules/events/logs/client/js/components/EventLog.jsx
```diff
@@ -6,13 +6,43 @@
...
+    dispatch(fetchLogEntries());
+  };
+
+  return (
+    <div className="highlight-message-box">
+      <span className="icon" />
+      <div className="message-text">
+        <Translate>
+          Log entries are currently filtered by URL.{' '}
+          <Param name="link" wrapper={<a onClick={resetURLFiltering} />}>
```
plourenco, line 31: `resetURLFiltering` works just fine. However, due to the fact this is combining react with server-side filtering, we will remain with the same URL query as before. In other words, we are only resetting this component's state. 

While this is not relevant, could be worth considering doing it all in the react app like https://github.com/indico/indico/pull/4503#discussion_r447708547 suggested or any workaround.

indico/modules/events/logs/client/js/components/EventLog.jsx
```diff
@@ -6,13 +6,43 @@
...
+      </div>
+    </div>
+  );
+}
 
 export default function EventLog() {
+  const metadataQuery = useSelector(state => state.logs.metadataQuery);
   return (
     <>
+      {Object.keys(metadataQuery).length !== 0 && <MetadataQueryMessage />}
```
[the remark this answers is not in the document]
ThiefMaster: that would render a `0`...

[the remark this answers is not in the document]
plourenco: Sorry, I forgot the `!!` before. Thanks.
Nevertheless, now that I'm looking at it, I think this suggestion is opinionated and irrelevant so I'm closing this thread.

OmeGak:
> I noticed that we are not attaching the registration id when logging emails related to a registration. I thing doing it there too would be a good idea.

Done!

indico/modules/events/logs/client/js/components/EventLog.jsx
OmeGak: `resetURLFiltering` now removes the arguments from the URL.

indico/modules/events/logs/client/js/components/LogEntryModal.jsx
```diff
@@ -62,6 +64,18 @@ export default class LogEntryModal extends React.Component {
     nextEntry();
   }
 
+  relatedEntries() {
+    const {currentViewIndex, entries, relatedEntries} = this.props;
+    const entry = entries[currentViewIndex];
+    const urlArgs = [];
+    Object.entries(entry.meta).forEach(function(item) {
```
ThiefMaster, line 71: arrow function please :)

ThiefMaster: Actually, I think you can use `qs.stringify(...)` to build the query string. Something like this

```jsx
const query = qs.stringify({meta: entry.meta}, {allowDots: true});
```

indico/modules/events/logs/client/js/components/LogEntryModal.jsx
```diff
@@ -62,6 +64,18 @@ export default class LogEntryModal extends React.Component {
     nextEntry();
   }
 
+  relatedEntries() {
```
ThiefMaster, line 67: `showRelatedEntries`?

indico/modules/events/logs/client/js/containers/LogEntryModal.js
```diff
@@ -26,6 +33,11 @@ const mapDispatchToProps = dispatch => ({
   nextEntry: () => {
     dispatch(viewNextEntry());
   },
+  relatedEntries: metadataQuery => {
+    dispatch(setMetadataQuery(metadataQuery));
+    dispatch(setPage(1));
+    dispatch(fetchLogEntries());
+  },
```
ThiefMaster, line 40: `dispatch(setDetailedView(null))` to close the modal as well?

indico/modules/events/logs/client/js/containers/LogEntryModal.js
```diff
@@ -26,6 +33,11 @@ const mapDispatchToProps = dispatch => ({
   nextEntry: () => {
     dispatch(viewNextEntry());
   },
+  relatedEntries: metadataQuery => {
+    dispatch(setMetadataQuery(metadataQuery));
+    dispatch(setPage(1));
+    dispatch(fetchLogEntries());
```
ThiefMaster, line 39: if you close the modal, `await` this dispatch to only close the modal after loading finished

indico/modules/events/logs/controllers.py
```diff
@@ -29,17 +31,20 @@ class RHEventLogs(RHManageEventBase):
     """Shows the modification/action log for the event"""
 
     def _process(self):
+        metadata_query = {k.lstrip('meta.'): int(v) if v.isdigit() else v
```
ThiefMaster, line 34: `lstrip` doesn't work this way. it removes all of the chars, so `meta.event_id` would turn into `vent_id`.. Only python 3.9 has a function to strip a prefix, so for now I'd keep the slice you had before..

indico/modules/events/logs/client/js/components/LogEntryModal.jsx
```diff
@@ -116,6 +136,10 @@ export default class LogEntryModal extends React.Component {
             >
               <Translate>Previous</Translate>
             </IButton>
+
+            <IButton onClick={this.relatedEntries} disabled={!this._hasRelatedEntries()}>
```
ThiefMaster, line 140: Should we also hide/disable it if the current metadata query already matches this entry's metadata?

OmeGak: I think that would be more confusing unless we have a good way to show in the UI the reason why the button is hidden/disabled. At the same time, I don't think it harms much to show the button even if it will trigger the same results. I would keep this as a nice to have.

indico/modules/events/logs/client/js/components/LogEntryModal.jsx
OmeGak: Yes, but then I would also rename the `nextEntry` and `prevEntry` to `showNextEntry` and `showPrevEntry`.

PR #4504, opened 2020-06-25
Add template hook to event management title

meluru:
- Adds `event-management-after-title` template hook
- Fixes `banner` class margins

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -105,6 +105,8 @@ Internal Changes
   :user:`omegak`)
 - Add ``extra-registration-actions`` template hook (:issue:`4500`, thanks
   :user:`omegak`)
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Looks fine. What exactly are you fixing in the banner margins? And example how it's broken without the fix? I see a tiny (1px or something in that area) shift between master and your branch, but nothing significant.

meluru:
When we add more content to the banner with the hook, 10% is more than 17px and things overlap.

ThiefMaster:
OK, makes sense. I was afraid the fix would break the alignment fixed in #4389 but that's not the case so everything is ok!

However, I don't think this fix needs a changelog entry since it's not broken in the core, and the hook also did not exist in a previous version.

meluru:
Great! I've removed the changelog entry.

PR #4510, opened 2020-07-01
Connect revision endpoints to the service mixin

### LINKED ISSUE #4502: Microservice task: Submit new editable
When submitting a new editable, the microservice can decide which state it should get in case it wants to do some extra processing on the submitted files and then replace the revision.

To allow the microservice to fetch files for the new revision, I'd suggest that the microservice returns the state for newly submitted revisions during setup time so Indico can just store it. That way during submission you just use that state, commit, and only then notify the microservice about the newly submitted editable.

plourenco:
Closes #4502 

- [ ] Have the micro-service store an initial state before it sets the editable as ready for review (do we really need this?)
- [x] Possibly deleting the editable if the micro-service is not available/fails

Known bugs:
- [x] Creating a revision right after the service replacement causes the front-end to crash. For some reason, it tries to get the editor from the previous revision (which is System) while showing the new dialog.

Service for testing:
- [Openreferee](https://github.com/indico/openreferee) (recommended)
- [Dummy service](https://gist.github.com/plourenco/930bfc440f3f88b591840ab4f51513bd).

### DIFF
--- a/indico/modules/events/editing/controllers/backend/timeline.py
+++ b/indico/modules/events/editing/controllers/backend/timeline.py
@@ -19,6 +19,7 @@ from werkzeug.exceptions import BadRequest, Forbidden, NotFound
 
 from indico.core.db import db
 from indico.core.errors import UserValueError
+from indico.modules.events import logger
 from indico.modules.events.editing.controllers.base import RHContributionEditableBase, TokenAccessMixin
 from indico.modules.events.editing.fields import EditingFilesField, EditingTagsField
 from indico.modules.events.editing.models.comments import EditingRevisionComment
@@ -88,8 +89,7 @@ class RHContributionEditableRevisionBase(RHContributionEditableBase):
         raise NotImplementedError
 
     def _check_access(self):
-        self.is_service_call = TokenAccessMixin._token_can_access(self)
-        if not self.is_service_call:
+        if not TokenAccessMixin._token_can_access(self):
             RHContributionEditableBase._check_access(self)
             if not self._check_revision_access():
                 raise UserValueError(_('You cannot perform this action on this revision'))
@@ -143,9 +143,10 @@ class RHCreateEditable(RHContributionEditableBase):
         if service_url:
             try:
                 service_handle_editable(editable)
-            except ServiceRequestFailed:
+            except ServiceRequestFailed as exc:
                 delete_editable(editable)
                 db.session.commit()
+                logger.exception('Failed to contact service. Got: %r', exc.message)
                 raise BadRequest(_('Failed while submitting for editing. The associated service is not available.'))
         return '', 201
 
--- a/indico/modules/events/editing/controllers/backend/timeline.py
+++ b/indico/modules/events/editing/controllers/backend/timeline.py
@@ -12,17 +12,18 @@ from io import BytesIO
 from zipfile import ZipFile
 
 from flask import request, session
-from marshmallow import fields
+from marshmallow import fields, validate
 from marshmallow_enum import EnumField
 from sqlalchemy.orm import joinedload
 from werkzeug.exceptions import Forbidden, NotFound
 
+from indico.core.db import db
 from indico.core.errors import UserValueError
-from indico.modules.events.editing.controllers.base import RHContributionEditableBase
+from indico.modules.events.editing.controllers.base import RHContributionEditableBase, TokenAccessMixin
 from indico.modules.events.editing.fields import EditingFilesField, EditingTagsField
 from indico.modules.events.editing.models.comments import EditingRevisionComment
 from indico.modules.events.editing.models.revision_files import EditingRevisionFile
-from indico.modules.events.editing.models.revisions import EditingRevision
+from indico.modules.events.editing.models.revisions import EditingRevision, InitialRevisionState
 from indico.modules.events.editing.operations import (assign_editor, confirm_editable_changes, create_new_editable,
                                                       create_revision_comment, create_submitter_revision,
                                                       delete_revision_comment, replace_revision,
@@ -30,7 +31,10 @@ from indico.modules.events.editing.operations import (assign_editor, confirm_edi
                                                       update_revision_comment)
 from indico.modules.events.editing.schemas import (EditableSchema, EditingConfirmationAction, EditingReviewAction,
                                                    ReviewEditableArgs)
+from indico.modules.events.editing.service import service_handle_editable
+from indico.modules.events.editing.settings import editing_settings
 from indico.modules.files.controllers import UploadFileMixin
+from indico.modules.users import User
 from indico.util.fs import secure_filename
 from indico.util.i18n import _
 from indico.util.marshmallow import not_empty
@@ -39,6 +43,7 @@ from indico.web.flask.util import send_file
 
 
 class RHEditingUploadFile(UploadFileMixin, RHContributionEditableBase):
+    SERVICE_ALLOWED = True
     EDITABLE_REQUIRED = False
 
     def get_file_context(self):
@@ -83,9 +88,11 @@ class RHContributionEditableRevisionBase(RHContributionEditableBase):
         raise NotImplementedError
 
     def _check_access(self):
-        RHContributionEditableBase._check_access(self)
-        if not self._check_revision_access():
-            raise UserValueError(_('You cannot perform this action on this revision'))
+        self.is_service_call = TokenAccessMixin._token_can_access(self)
+        if not self.is_service_call:
+            RHContributionEditableBase._check_access(self)
+            if not self._check_revision_access():
+                raise UserValueError(_('You cannot perform this action on this revision'))
 
 
 class RHEditable(RHContributionEditableBase):
@@ -128,8 +135,13 @@ class RHCreateEditable(RHContributionEditableBase):
         args = parser.parse({
             'files': EditingFilesField(self.event, self.contrib, self.editable_type, required=True)
         })
+        service_url = editing_settings.get(self.event, 'service_url')
+        initial_state = InitialRevisionState.new if service_url else InitialRevisionState.ready_for_review
 
-        create_new_editable(self.contrib, self.editable_type, session.user, args['files'])
+        editable = create_new_editable(self.contrib, self.editable_type, session.user, args['files'], initial_state)
+        db.session.commit()
+        if service_url:
+            service_handle_editable(editable)
         return '', 201
 
 
@@ -168,20 +180,24 @@ class RHConfirmEditableChanges(RHContributionEditableRevisionBase):
 class RHReplaceRevision(RHContributionEditableRevisionBase):
     """Replace the latest revision of an Editable."""
 
+    SERVICE_ALLOWED = True
+
     def _check_revision_access(self):
         # XXX: we still need UI for this or expose it only to the microservice
         return self.editable.editor is None and self.editable.can_perform_submitter_actions(session.user)
 
     @use_kwargs({
-        'comment': fields.String(missing='')
+        'comment': fields.String(missing=''),
+        'state': fields.String(validate=validate.OneOf(InitialRevisionState.__members__))
     })
-    def _process(self, comment):
+    def _process(self, comment, state):
         args = parser.parse({
             'files': EditingFilesField(self.event, self.contrib, self.editable_type, allow_claimed_files=True,
                                        required=True)
         })
 
-        replace_revision(self.revision, session.user, comment, args['files'])
+        user = User.get_system_user() if self.is_service_call else session.user
+        replace_revision(self.revision, user, comment, args['files'], InitialRevisionState[state])
         return '', 204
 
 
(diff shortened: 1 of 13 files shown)


indico/modules/events/editing/operations.py
```diff
@@ -157,15 +157,15 @@ def confirm_editable_changes(revision, submitter, action, comment):
...
+def replace_revision(revision, user, comment, files, initial_state=None):
     _ensure_latest_revision(revision)
     _ensure_state(revision,
                   initial=(InitialRevisionState.new, InitialRevisionState.ready_for_review),
                   final=FinalRevisionState.none)
     revision.comment = comment
     revision.final_state = FinalRevisionState.replaced
     new_revision = EditingRevision(submitter=user,
-                                   initial_state=revision.initial_state,
+                                   initial_state=initial_state or revision.initial_state,
```
ThiefMaster, line 168:
```suggestion
                                   initial_state=(initial_state or revision.initial_state),
```

indico/modules/events/editing/controllers/base.py
```diff
@@ -92,8 +92,10 @@ class RHContributionEditableBase(RequireUserMixin, RHContributionDisplayBase):
     _editable_query_options = ()
 
     def _check_access(self):
-        RequireUserMixin._check_access(self)
-        RHContributionDisplayBase._check_access(self)
+        self.is_service_call = TokenAccessMixin._token_can_access(self)
```
ThiefMaster, line 95: What do you think about moving this to `TokenAccessMixin`?

`is_service_call = False` on the class level (so it's always set even if `TokenAccessMixin._token_can_access` is not executed, and set it to true in `_token_can_access`

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -168,20 +180,24 @@ def _process(self, action, comment):
...
+    SERVICE_ALLOWED = True
+
     def _check_revision_access(self):
         # XXX: we still need UI for this or expose it only to the microservice
         return self.editable.editor is None and self.editable.can_perform_submitter_actions(session.user)
 
     @use_kwargs({
-        'comment': fields.String(missing='')
+        'comment': fields.String(missing=''),
+        'state': fields.String(validate=validate.OneOf(InitialRevisionState.__members__))
```
ThiefMaster, line 191:
```suggestion
        'state': EnumField(InitialRevisionState)
```

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -168,20 +180,24 @@ def _process(self, action, comment):
...
-    def _process(self, comment):
+    def _process(self, comment, state):
         args = parser.parse({
             'files': EditingFilesField(self.event, self.contrib, self.editable_type, allow_claimed_files=True,
                                        required=True)
         })
 
-        replace_revision(self.revision, session.user, comment, args['files'])
+        user = User.get_system_user() if self.is_service_call else session.user
+        replace_revision(self.revision, user, comment, args['files'], InitialRevisionState[state])
```
ThiefMaster, line 200:
```suggestion
        replace_revision(self.revision, user, comment, args['files'], state)
```

indico/modules/events/editing/service.py
```diff
@@ -123,3 +124,29 @@ def service_get_status(event):
...
+
+def service_handle_editable(editable):
+    revision = editable.revisions[-1]
+    data = {
+        'files': EditingRevisionFileSchema().dump(revision.files, many=True),
+        'endpoints': {
+            'revisions': {
+                'replace': url_for('.api_replace_revision', editable.event,
+                                   contrib_id=editable.contribution_id, type=editable.type.name,
+                                   revision_id=revision.id, _external=True)
```
ThiefMaster, line 137:
```suggestion
                'replace': url_for('.api_replace_revision', revision, _external=True)
```

indico/modules/events/editing/service.py
```diff
@@ -123,3 +124,29 @@ def service_get_status(event):
...
+            },
+            'file_upload': url_for('.api_upload', editable.event,
+                                   contrib_id=editable.contribution_id, type=editable.type.name,
+                                   _external=True)
```
ThiefMaster, line 141:
```suggestion
            'file_upload': url_for('.api_upload', editable.contribution, type=editable.type.name, _external=True)
```

indico/modules/events/editing/service.py
```diff
@@ -123,3 +124,29 @@ def service_get_status(event):
...
+        }
+    }
+    try:
+        resp = requests.post(_build_url(editable.event, '/event/{}/contributions/{}/editing/{}'
+                                        .format(_get_event_identifier(editable.event),
+                                                editable.contribution_id, editable.type.name)),
+                             headers=_get_headers(editable.event), json=data)
```
ThiefMaster, line 148: I think this would make it more readable:

```suggestion
        path = '/event/{}/contributions/{}/editing/{}'.format(
            _get_event_identifier(editable.event),
            editable.contribution_id,
            editable.type.name
        )
        resp = requests.post(_build_url(editable.event, path), headers=_get_headers(editable.event), json=data)
```

indico/modules/events/editing/controllers/backend/timeline.py
plourenco: I had the impression there was a better way of doing this 😄

indico/modules/events/editing/service.py
plourenco: Looks good!

indico/modules/events/editing/controllers/base.py
plourenco: And then just calling `TokenAccessMixin._token_can_access(self)` and expect the `is_service_call` to be assigned?

ThiefMaster: In here: `if not TokenAccessMixin._token_can_access(self):`

and just use the attribute when you need to check it elsewhere

plourenco: I like the idea of simplifying this service call assignment. But for some reason, it feels dangerous to "hide" this assignment (or implicitly set it). Nevertheless, since it will get re-executed every time on `_check_access` I think it is fine.

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -128,8 +135,19 @@ def _process(self):
...
+
+        editable = create_new_editable(self.contrib, self.editable_type, session.user, args['files'], initial_state)
+        db.session.commit()
+        if service_url:
+            try:
+                service_handle_editable(editable)
+            except ServiceRequestFailed as exc:
+                delete_editable(editable)
+                db.session.commit()
+                logger.exception('Failed to contact service. Got: %r', exc.message)
```
ThiefMaster, line 149: No need to include exception details again; `logger.exception()` already adds the full exception and traceback to the logging record.

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -128,8 +135,19 @@ def _process(self):
...
+                raise BadRequest(_('Failed while submitting for editing. The associated service is not available.'))
```
ThiefMaster, line 150: Sounds a bit strange. What about something more generic like "Submission failed, please try again later."? The end user does not really know about the microservice etc.

ThiefMaster, line 150: `BadRequest` does not fit here since it's not the client's fault. `ServiceUnavailable` (503) would be more appropriate.

indico/modules/events/editing/controllers/base.py
```diff
@@ -25,6 +25,10 @@
 class TokenAccessMixin(object):
     SERVICE_ALLOWED = False
 
+    def __init__(self):
+        super(TokenAccessMixin, self).__init__()
+        self.is_service_call = False
```
ThiefMaster, line 30: No need for this, you can just add `is_service_call = False` above below `SERVICE_ALLOWED = False`. It's not mutable so nothing wrong with setting the default on the class.

indico/modules/events/editing/controllers/base.py
```diff
@@ -36,7 +40,8 @@ def _token_can_access(self):
         if request.bearer_token != event_token:
             raise Unauthorized('Invalid bearer token')
 
-        return True
+        self.is_service_call = True
+        return self.is_service_call
```
ThiefMaster, line 44: `return True`? it's always the same value anyway!

indico/modules/events/editing/operations.py
```diff
@@ -78,17 +78,25 @@ def _make_editable_files(editable, files):
...
     db.session.flush()
     logger.info('Editable [%s] created by %s for %s', type_.name, submitter, contrib)
     return editable
 
 
+def delete_editable(editable):
+    db.session.delete(editable)
+    for revision in editable.revisions:
+        for ef in revision.files:
+            ef.file.unclaim()
```
ThiefMaster, line 96: This is a bit dangerous: Even though it's never the case here (for now), a file could be referenced by other revisions etc. 

So I'm not sure if it's worth adding this just to avoid orphaned files. If we really *do* want to avoid orphaned files, I think we need to use something more sophisticated, like a reference counter (+1 on claim, -1 on unclaim) or check if there are any incoming relationship references left...

indico/modules/events/editing/service.py
```diff
@@ -123,3 +124,27 @@ def service_get_status(event):
     except requests.RequestException as exc:
         return {'status': None, 'error': unicode(ServiceRequestFailed(exc))}
     return {'status': resp.json(), 'error': None}
+
+
+def service_handle_editable(editable):
```
ThiefMaster, line 129: `service_handle_editable_submission` or `service_handle_new_editable`?

indico/modules/events/editing/service.py
```diff
@@ -123,3 +124,27 @@ def service_get_status(event):
...
+    revision = editable.revisions[-1]
+    data = {
+        'files': EditingRevisionFileSchema().dump(revision.files, many=True),
+        'endpoints': {
+            'revisions': {
+                'replace': url_for('.api_replace_revision', revision, _external=True)
+            },
+            'file_upload': url_for('.api_upload', editable.contribution, type=editable.type.name, _external=True)
```
ThiefMaster, line 137: You can just pass the `editable` to get everything since is locator includes the contribution's locator and the type.

indico/modules/events/editing/controllers/backend/timeline.py
plourenco: It is a good idea but might induce into the mistake of thinking the submission error was related to the editable itself, instead of the service that processes it.

plourenco: Sounds good. I made it `BadRequest` initially because I was considering errors coming from corrupted client input.

ThiefMaster: Not really important for the user, and I think "try again later" indicates that it's not something the user did wrong. In any case, the service should be stable ;)

indico/modules/events/editing/operations.py
plourenco: The thing is, if I do not un-claim the files, the next submissions after an error will fail (e.g. when re-trying).

ThiefMaster: OK good point. Maybe we can restrict this function to only work on an editable with a single revision? We don't expect to ever use it for other editables anyway. And since we do not allow previously-claimed files when creating a new editable, this would ensure you never have a file claimed elsewhere.

Could even go one step further and do the unclaiming in the RH itself and not in this operation..

ThiefMaster:
> Have the micro-service store an initial state before it sets the editable as ready for review (do we really need this?)

Depends on how hard we want to depend on the microservice. If we are fine with expecting the microservice to always take care of updating the state (we probably still need an endpoint for this, since not every editable will have its initial revision replaced) then no, we don't need to store an initial state.

indico/modules/events/editing/controllers/base.py
plourenco: Of course, static here doesn't really apply. Thanks!

indico/modules/events/editing/service.py
plourenco: I like the second one.

indico/modules/events/editing/operations.py
plourenco: Not sure how I feel about having this in the RH itself since I expect to control this kind of operations in the file model. I'm leaving it as-is for now, perhaps later I could add a single revision check.

indico/modules/files/models/files.py
```diff
@@ -18,6 +17,8 @@
 from indico.util.fs import secure_filename
 from indico.util.string import format_repr, return_ascii, strict_unicode
 
+import posixpath
```
plourenco, line 20: CI is failing here. Not sure yet why this is happening. But please ignore it, I'll fix it before merging.

ThiefMaster: > Not sure yet why this is happening. 

`posixpath` is stdlib` so that import should remain at its previous position

PR #4511, opened 2020-07-01
Several core improvements for plugins

pferreir:
A few improvements in the core related to several plugins:

 * VC module - better handling of exceptions in some error situations;
 * Timetable - flash messages from timetable updates;
 * Adding a `plusdelta` filter to Jinja templates (useful for e-mails);
 * Fixing the wording of Room Booking e-mails, which where saying e.g. "booking cancelled on XXX" when XXX is actually the date of the booking, not the operation.
 * Adding a `decodeprincipal` filter to Jinja;
 * Triggering custom even when an ajaxDialog loads;
 * Removing the forced sanitization of VCRoom names;
 * Some minor CSS fixes;

### DIFF
--- a/indico/modules/rb/models/reservation_occurrences_test.py
+++ b/indico/modules/rb/models/reservation_occurrences_test.py
@@ -258,7 +258,7 @@ def test_cancel(smtp, create_reservation, dummy_user, silent, reason):
     else:
         assert occurrence.reservation.edit_logs.count() == 1
         assert occurrence.reservation.edit_logs[0].user_name == dummy_user.full_name
-        extract_emails(smtp, count=2, regex=True, subject=r'Booking cancelled on .+ \(SINGLE OCCURRENCE\)')
+        extract_emails(smtp, count=2, regex=True, subject=r'Booking cancelled .+ \(SINGLE OCCURRENCE\)')
         if reason:
             assert len(occurrence.reservation.edit_logs[0].info) == 2
         else:
@@ -280,7 +280,7 @@ def test_cancel_single_occurrence(smtp, create_occurrence, dummy_user, silent, f
         assert not occ.reservation.edit_logs.count()
     else:
         assert occ.reservation.edit_logs.count()
-        mails = extract_emails(smtp, count=2, regex=True, subject=r'Booking cancelled on')
+        mails = extract_emails(smtp, count=2, regex=True, subject=r'Booking cancelled')
         assert not any('SINGLE OCCURRENCE' in mail['subject'] for mail in mails)
     assert not smtp.outbox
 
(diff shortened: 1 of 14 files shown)


indico/modules/rb/models/reservation_occurrences_test.py
```diff
@@ -258,7 +258,7 @@ def test_cancel(smtp, create_reservation, dummy_user, silent, reason):
     else:
         assert occurrence.reservation.edit_logs.count() == 1
         assert occurrence.reservation.edit_logs[0].user_name == dummy_user.full_name
-        extract_emails(smtp, count=2, regex=True, subject=r'Booking cancelled on .+ \(SINGLE OCCURRENCE\)')
+        extract_emails(smtp, count=2, regex=True, subject=r'Booking cancelled .+ \(SINGLE OCCURRENCE\)')
```
ThiefMaster, line 261: What do you think about using a subject like `Booking cancelled: XXX` instead of `Booking cancelled XXX` (the latter sounds like the booking cancelled XXX, and I hope neither our meeting rooms nor their bookings ever become sentient :D)

pferreir: I added some [parenthesis](https://github.com/indico/indico/pull/4511/files#diff-8d9f968452beacfa32590ba89f5283b0R51).

ThiefMaster: you're right, i was just looking at the `subject=...` when creating the notifications

indico/web/templates/forms/_form.html
```diff
@@ -8,14 +8,22 @@
     </ul>
 {%- endmacro %}
 
+{% macro _serialize_data_attrs(attrs) -%}
```
ThiefMaster, line 11: AFAIK the `html_params` Jinja filter (coming from WTForms) does this for you

indico/web/client/js/jquery/utils/declarative.js
```diff
@@ -163,6 +163,9 @@ import {$T} from '../../utils/i18n';
             dialogClasses: $this.data('dialog-classes'),
             hidePageHeader: $this.data('hide-page-header') !== undefined,
             confirmCloseUnsaved: $this.data('confirm-close-unsaved') !== undefined,
+            onOpen: ({dialogElement}) => {
+              dialogElement.trigger('indico:dialog-open');
```
ThiefMaster, line 167: I think we use camelCase for other custom events

indico/web/flask/templating.py
```diff
@@ -80,6 +81,17 @@ def groupby(environment, value, attribute, reverse=False):
             for key, values in itertools.groupby(sorted(value, key=expr, reverse=reverse), expr)]
 
 
+def plusdelta(value, **kwargs):
+    """Add a relativedelta to a datetime object."""
+    return value + relativedelta(**kwargs)
+
+
+def decodeprincipal(value, **kwargs):
+    """Decodes a principal into an actual object"""
```
ThiefMaster, line 90:
```suggestion
    """Decode a principal into an actual object."""
```

PR #4516, opened 2020-07-02
Editing: Self-assignment ("Paper Bingo")

### LINKED ISSUE #4487: Editing: Self-assignment ("Paper Bingo")
Provide filters for file types that have multiple extensions configured.

Example: The editable type has three file types: PDF, Source (.docx, .tex) and Image (.png, .jpg). Then the filter options would let you select docx/tex for Source and png/jpg for Image. If the file type is not required, we oculd also show "Any" (even for types with no extension requirements) to include anything that has files present.

---

This will probably be linked from the editor-visible editable list (does not exist yet) in the hybrid area. When working on this task before this is done, just add a button in the Editable Type dashboard (e.g. next to the "List" button)

Leats:
Closes #4487 
Bildschirmfoto 2020-07-03 um 09 48 59


indico/modules/events/editing/client/js/management/editable_type/NextEditable.jsx
```diff
@@ -0,0 +1,273 @@
...
+
+  return (
+    <Modal onClose={onClose} closeOnDimmerClick={false} open>
+      <Modal.Header>{titleByType[editableType]}</Modal.Header>
+      <Modal.Content>
+        <div styleName="filetype-list">
+          {fileTypes.map(fileType => {
+            return (
+              <div key={fileType.id} styleName="filetype-list-item">
+                <div>
```
panagiotappl, line 134: I think this `div` can be omitted here since it's plain text inside and doesn't need to add any extra css. I can also be replaced with a `span`.

indico/modules/events/editing/client/js/management/editable_type/NextEditable.jsx
```diff
@@ -0,0 +1,273 @@
...
+                        {fileType.extensions.map(extension => (
+                          <Button
+                            disabled={loading}
+                            content={extension}
+                            key={extension}
+                            onClick={() => updateExtensionFilter(fileType.id, extension)}
+                            color={
+                              Array.isArray(filters[fileType.id]) &&
+                              filters[fileType.id].includes(extension)
+                                ? 'blue'
```
panagiotappl, line 150: You could import Palette from `web/client/js/utils/palette.js` here and use one of those colors in case we want to stick to our palette. Same goes for other color usages here (like orange etc)

ThiefMaster: we cannot use our palette there since those are SUI colors

PR #4525, opened 2020-07-09
Add link in Editable List for "Paper Bingo"

### LINKED ISSUE #4519: "Paper Bingo": Add link in Editing View and adjust permissions
Currently it's manager-only and only accessible by managers. Now that #4501 is merged, we should add a button there (maybe in the editable list?) and change the inheritance of the RHs so it can also be accessed by editors.

Not sure if the permissions actually need changing or are already fine... please test while logged in as a user who is not an Indico admin or event manager but really just an editor.

Leats:
Closes #4519
Adds link and changes permissions accordingly.

### DIFF
--- a/indico/modules/events/editing/client/js/management/editable_type/NextEditable.jsx
+++ b/indico/modules/events/editing/client/js/management/editable_type/NextEditable.jsx
@@ -11,6 +11,7 @@ import fileTypesURL from 'indico-url:event_editing.api_file_types';
 
 import _ from 'lodash';
 import React, {useState, useEffect} from 'react';
+import {useHistory} from 'react-router-dom';
 import PropTypes from 'prop-types';
 import {Button, Loader, Modal, Table, Checkbox, Dimmer} from 'semantic-ui-react';
 import {camelizeKeys} from 'indico/utils/case';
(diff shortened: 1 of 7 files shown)


indico/modules/events/editing/client/js/management/editable_type/NextEditable.jsx
```diff
@@ -49,13 +51,15 @@ NextEditable.propTypes = {
   eventId: PropTypes.number.isRequired,
   editableType: PropTypes.oneOf(Object.values(EditableType)).isRequired,
   onClose: PropTypes.func.isRequired,
+  management: PropTypes.bool.isRequired,
 };
 
-function NextEditableDisplay({eventId, editableType, onClose, fileTypes}) {
+function NextEditableDisplay({eventId, editableType, onClose, fileTypes, management}) {
```
plourenco, line 57: Perhaps later, to avoid stalling this PR, we could merge `NextEditable` with `NextEditableDisplay` since the latter isn't used individually. I don't think there's any reason to keep them separate.

ThiefMaster: The `NextEditableDisplay` component has a `useEffect` which loads the actual editables - having a separate component preload all the things that never change has the advantage that you can just return null or the loader there, and don't have to run the hooks of the main component all the time.

Ideally we would just load all the static data needed on the various pages on a higher level and provide it via context though... But I don't think it's that important here since there's not much overhead in loading it here.

PR #4526, opened 2020-07-09
Prevent managers from editing/deleting system tags

ThiefMaster:
image

### DIFF
--- a/indico/modules/events/editing/client/js/management/tags/TagManager.jsx
+++ b/indico/modules/events/editing/client/js/management/tags/TagManager.jsx
@@ -110,7 +110,7 @@ export default function TagManager({eventId}) {
                     disabled={tag.system}
                     circular
                     inverted
-                  />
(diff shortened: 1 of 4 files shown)


indico/modules/events/editing/client/js/management/tags/TagManager.jsx
```diff
@@ -96,22 +96,37 @@ export default function TagManager({eventId}) {
...
-              color="red"
-              size="small"
-              onClick={() => dispatch({type: 'DELETE_TAG', tag})}
-              circular
-              inverted
-            />
+            <Popup
+              on="hover"
+              position="right center"
+              // disabled={!tag.system}
```
Leats, line 102: uncomment :)

ThiefMaster: oops :D

PR #4530, opened 2020-07-10
Introduce tighter controls on abstract modification

### LINKED ISSUE #3431: Abstracts: Introduce tighter controls on modification
Right now, all authors/speakers [can modify an abstract](https://github.com/indico/indico/blob/master/indico/modules/events/abstracts/models/abstracts.py#L652). While this is a sensible default for most events, in larger ones, stricter controls may be required. A good example is the JACoW workflow, according to which:
 * **Speakers** should have submission rights (uploading files);
 * Only **primary authors** should be able to edit the title/description and authors of the abstract;

This should be ideally configurable.

Leats:
- [x] Speakers should have submission rights (uploading files);
- [x] Only primary authors should be able to edit the title/description and authors of the abstract (ideally configurable)

closes #3431

### DIFF
--- a/indico/modules/events/abstracts/forms.py
+++ b/indico/modules/events/abstracts/forms.py
@@ -23,8 +23,8 @@ from indico.modules.events.abstracts.fields import (AbstractField, AbstractPerso
                                                     TrackRoleField)
 from indico.modules.events.abstracts.models.abstracts import EditTrackMode
 from indico.modules.events.abstracts.models.reviews import AbstractAction, AbstractCommentVisibility
-from indico.modules.events.abstracts.settings import (BOACorrespondingAuthorType, BOALinkFormat, BOASortField,
-                                                      abstracts_settings)
+from indico.modules.events.abstracts.settings import (AllowEditingType, BOACorrespondingAuthorType, BOALinkFormat,
+                                                      BOASortField, abstracts_settings)
 from indico.modules.events.contributions.models.persons import AuthorType
 from indico.modules.events.contributions.models.types import ContributionType
 from indico.modules.events.sessions.models.sessions import Session
@@ -121,6 +121,7 @@ class AbstractSubmissionSettingsForm(IndicoForm):
                                   description=_("Allow the selection of the abstract speakers"))
     speakers_required = BooleanField(_('Require a speaker'), [HiddenUnless('allow_speakers')], widget=SwitchWidget(),
                                      description=_("Make the selection of at least one author as speaker mandatory"))
+    allow_editing = IndicoEnumSelectField(_('Allow editing'), enum=AllowEditingType, sorted=True)
     authorized_submitters = PrincipalListField(_("Authorized submitters"),
                                                description=_("These users may always submit abstracts, even outside "
                                                              "the regular submission period."))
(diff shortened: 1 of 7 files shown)


indico/modules/events/abstracts/forms.py
```diff
@@ -121,6 +121,10 @@ class AbstractSubmissionSettingsForm(IndicoForm):
                                   description=_("Allow the selection of the abstract speakers"))
     speakers_required = BooleanField(_('Require a speaker'), [HiddenUnless('allow_speakers')], widget=SwitchWidget(),
                                      description=_("Make the selection of at least one author as speaker mandatory"))
+    allow_editing = IndicoEnumSelectField(_('Allow editing'), enum=AllowEditingType, sorted=True)
+    only_speaker_submitter = BooleanField(_('Submission rights'), widget=SwitchWidget(),
```
pferreir, line 125: The naming of the option seems a bit vague, maybe? It would be nice to see how this renders, but from what I see it may be perceived as confusing. What about using an enum like in the field just before?

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -564,13 +570,31 @@ def can_judge(self, user, check_state=False):
...
+            return True
+        elif editing_allowed == AllowEditingType.submitter_all:
+            return True
+        elif editing_allowed == AllowEditingType.submitter_primary and is_primary:
             return True
-        elif (self.public_state in (AbstractPublicState.awaiting, AbstractPublicState.invited) and
-                (is_manager or self.event.cfa.can_edit_abstracts(user))):
+        elif editing_allowed == AllowEditingType.submitter_authors and (is_primary or is_secondary):
             return True
         else:
```
pferreir, line 599: This `else` is actually kind of useless...

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -564,13 +570,31 @@ def can_judge(self, user, check_state=False):
...
             return False
-        elif is_manager and self.public_state in (AbstractPublicState.under_review, AbstractPublicState.withdrawn):
+        elif not self.user_owns(user) or not self.event.cfa.can_edit_abstracts(user):
+            return False
+
+        editing_allowed = self.event.cfa.allow_editing
+        author_type = next((x.author_type for x in self.person_links if x.person.user == user), None)
+        is_primary = author_type == AuthorType.primary
+        is_secondary = author_type == AuthorType.secondary
+        if user == self.submitter:
```
plourenco, line 591: Shouldn't this be ...?
```suggestion
        if self.user_owns(user):
```

ThiefMaster: that would allow any linked person again. `user_owns` would fit in case of `editing_allowed == AllowEditingType.submitter_all` but since it's checked above it's not needed there

plourenco: That makes sense.

Leats:
Updated; Please let me know if you can think of a better name for `submission_rights`! It is very general but I can't think of anything short that's more descriptive. It might be enough with the enum description, though.

pferreir:
> Updated; Please let me know if you can think of a better name for `submission_rights`! It is very general but I can't think of anything short that's more descriptive. It might be enough with the enum description, though.

`submission_allowed_for`? I'm not sure it's that much better, though.

ThiefMaster:
`contribution_submitters` maybe?

Issue #4505, opened 2020-06-26
Allow selecting event/category roles and registrants in attachment/folder ACLs


ThiefMaster:
These types are already whitelisted in the models, but the `AccessControlListField` (which is basically a `PrincipalListField`) used there does not contain an option to select those types.

Unfortunately this field is still rendered using *very* legacy (pre-jquery) JS code. I see two options here:

- Add support for these three principal types in the legacy field. Not fun, and the code will be ugly
- Use the react-based `PrincipalListField`. You'd need to wrap it in a `WTFPrincipalListField` to integrate it in a classic form (where the initial value is loaded form a hidden field, and making any changes update the same hidden field). @Leats added two fields a while ago for the date/time pickers that make use of a similar integration, so you could have a look how it was done there.

Note: If you go for the second option (which would be the cleaner one), you will have to update the Python-WTForms `PrincipalListField` as well to use identifiers instead of fossils for the submitted data.

A good first step here before doing anything related to the wtforms integration would be adding support to select registration forms there (like it's done for event/category roles, ie with a hook similar to `useFetchEventCategoryRoles` to get the list of regforms to select etc). This could even be a separate Pull Request.

PR #4532, opened 2020-07-13
Add support for registration forms in the PrincipalList

plourenco:
Related to #4505 

Adds support to select registration forms in the PrincipalListField (like it's done for event/category roles).


indico/modules/core/controllers.py
```diff
@@ -244,6 +244,11 @@ def _serialize_principal(self, identifier, principal):
                     'name': principal.name,
                     'meta': {'style': principal.style, 'code': principal.code},
                     'detail': principal.category.title}
+        elif principal.principal_type == PrincipalType.registration_form:
+            return {'identifier': identifier,
+                    'type': 'registration_form',
+                    'invalid': False,
```
ThiefMaster, line 250: I think we can use `principal.is_deleted` here

indico/modules/events/management/controllers/protection.py
```diff
@@ -147,3 +148,10 @@ class RHCategoryRolesJSON(RHManageEventBase):
     def _process(self):
         category_roles = CategoryRole.get_category_roles(self.event.category)
         return jsonify([serialize_category_role(cr, legacy=False) for cr in category_roles])
+
+
+class RHRegistrationFormsJSON(RHManageEventBase):  # TODO: Give this the API prefix instead of JSON?
```
ThiefMaster, line 153: I wonder if this should stay here or go in the registrations module...

indico/modules/events/registration/schemas.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class RegistrationFormSchema(mm.ModelSchema):
+    class Meta:
+        model = RegistrationForm
+        fields = ('id', 'name', 'identifier')
+
+    id = Integer(attribute='id')
+    name = String(attribute='title')
+    identifier = 'RegistrationForm:{}'.format(id)
```
ThiefMaster, line 14: I don't think this works...

indico/modules/events/management/controllers/protection.py
```diff
@@ -147,3 +148,10 @@ class RHCategoryRolesJSON(RHManageEventBase):
...
+    PERMISSION = 'ANY'
```
ThiefMaster, line 154: Materials can be managed by people who aren't managers as long as they have submission rights on something, so I think we need to give anyone who can access the event access to this endpoint :/

We could also check if the user has submission access on the event or any session/contribution but maybe that's overkill.

indico/modules/events/registration/schemas.py
```diff
@@ -0,0 +1,14 @@
...
+from indico.modules.events.registration.models.forms import RegistrationForm
+
+
+class RegistrationFormSchema(mm.ModelSchema):
+    class Meta:
+        model = RegistrationForm
+        fields = ('id', 'name', 'identifier')
+
+    id = Integer(attribute='id')
+    name = String(attribute='title')
```
ThiefMaster, line 13: no need to specify `attribute` if it matches the actual class attribute

indico/modules/events/registration/schemas.py
```diff
@@ -0,0 +1,14 @@
+from marshmallow.fields import Integer, String
```
ThiefMaster, line 1: missing file header

ThiefMaster: tip, you can use `python bin/maintenance/update_header.py indico` to check (we should probably add this to CI)

indico/modules/events/registration/schemas.py
ThiefMaster: But you don't need any of the last three lines, because they are all attributes of RegistrationForm and should be handled implicitly. In case `identifier` (not a column) isn't, then `identifier = String()` would be enough.

indico/web/client/js/react/components/principals/hooks.js
```diff
@@ -100,3 +101,18 @@ export const useFetchEventCategoryRoles = eventId => {
 
   return data || [];
 };
+
+/**
+ * This hook fetches
```
ThiefMaster, line 106: something's missing here ;)

indico/modules/events/management/controllers/protection.py
plourenco: I had it like that before, until I noticed both categories and event roles were here.
I also don't like to see it here, I'll get them back. Not sure if we should do something about the others tho.

ThiefMaster: hm, category roles don't really have a better place... maybe keeping it here for now is fine since it's somewhat generic and used everywhere in events, not just in registration forms. that way all of these endpoints are next to each other

indico/modules/events/registration/schemas.py
plourenco: Agreed, we could add it in CI. Just noticed a test from the room booking passed through.

indico/modules/events/management/controllers/protection.py
plourenco: I see what you mean, I also don't think it is critical to leave it open to the event. But I can always change it eventually.

indico/modules/events/registration/schemas.py
plourenco: It seemed to work, but I did not notice there was already a property in the model. I was trying to avoid adding a new one solely for the principal list purpose. Thanks!

indico/modules/events/registration/api.py
```diff
@@ -67,3 +69,8 @@ def _process_args(self):
 
     def _process_GET(self):
         return jsonify(registrants=build_registrations_api_data(self.event))
+
+
+class RHAPIRegistrationForms(RHProtectedEventBase):
+    def _process(self):
+        return jsonify(RegistrationFormSchema().dump(self.event.registration_forms, many=True))
```
ThiefMaster, line 76:
```suggestion
        return RegistrationFormSchema(many=True).jsonify(self.event.registration_forms)
```

indico/web/client/js/react/components/principals/util.js
```diff
@@ -115,6 +117,7 @@ export class PrincipalType {
       [PrincipalType.multipassGroup]: Translate.string('Unknown group'),
       [PrincipalType.eventRole]: Translate.string('Unknown event role'),
       [PrincipalType.categoryRole]: Translate.string('Unknown category role'),
+      [PrincipalType.registrationForm]: Translate.string('Unknown registration form'),
     }[type];
   }
```
ThiefMaster, line 123: you need an entry in `getDeletedText` as well

indico/modules/events/registration/api.py
plourenco: Isn't it the same thing?

ThiefMaster: yes, it's just a shortcut to do it:

https://github.com/marshmallow-code/flask-marshmallow/blob/92d2c20ab8da5e669c7327f39c1f7a94f021ca3d/src/flask_marshmallow/schema.py#L16

indico/modules/events/controllers/base.py
```diff
@@ -25,21 +25,29 @@ def _process_args(self):
             raise NotFound(_('This event has been deleted.'))
 
 
-class RHDisplayEventBase(RHEventBase):
+class RHProtectedEventBase(RHEventBase):
+    """
+    Requires event access, regardless of the user being logged in
+    """
```
ThiefMaster, line 31:
```suggestion
    """Base class for events that check if he user can access it.
    
    This includes unauthenticated users who have access to the event
    for any other reason (e.g. due to a whitelisted IP address).
    """
```

indico/modules/events/registration/schemas.py
```diff
@@ -0,0 +1,20 @@
...
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from marshmallow.fields import String
+
+from indico.core.marshmallow import mm
+
+
+class RegistrationFormSchema(mm.ModelSchema):
```
ThiefMaster, line 13: I'd rename this to `RegistrationFormPrincipalSchema` since it's very basic and if we ever add a more complete schema we don't want to suddenly start returning more details from the existing RH.

indico/modules/events/registration/schemas.py
```diff
@@ -0,0 +1,20 @@
...
+    class Meta:
+        from indico.modules.events.registration.models.forms import RegistrationForm
```
ThiefMaster, line 15: I don't think having the import here has any benefit. This still runs at import time...

indico/web/client/js/react/components/principals/PrincipalListField.jsx
```diff
@@ -156,6 +163,20 @@ const PrincipalListField = props => {
               onChange={(e, data) => handleAddItems([{identifier: data.value}])}
             />
           )}
+          {registrationForms.length !== 0 && (
+            <Dropdown
```
ThiefMaster, line 167: Now that we have 3 of these dropdowns, what about wrapping it in a component?

```jsx
<AddPrincipalDropdown
  text={Translate.string('Registration Form')}
  options={registrationFormOptions}
  onAdd={value => handleAddItems([{identifier: value}])}
/>
```

indico/web/rh.py
```diff
@@ -327,6 +327,9 @@ def wrapper(*args, **kwargs):
 
 
 class RHProtected(RH):
+    """
+    Requires the session to have a user
+    """
```
ThiefMaster, line 332:
```suggestion
    """Base RH that requires the user to be authenticated."""
```

https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

indico/modules/events/registration/schemas.py
plourenco: This is the fix for [this](https://github.com/indico/indico/pull/4532/checks?check_run_id=870558678). I previously thought it wasn’t working but I mistakenly imported it incorrectly.

ThiefMaster: If this particular line fixes it I'm very curious how and why... Because the code here runs immediately:

```
>>> class Foo(object):
...     class Meta:
...             print 'hi'
...             x = 1
...
hi
```

And at the point where it runs, nothing is defined compared to what would happen if you did the import right before the class.

indico/modules/events/registration/schemas.py
plourenco: Good idea

indico/web/client/js/react/components/principals/PrincipalListField.jsx
plourenco: Sounds good, didn't even notice this. Much better! I tried to build a map but the text doesn't help. Didn't move the `ifs` because it feels weird to let the child control rendering.

indico/modules/events/controllers/base.py
```diff
@@ -25,21 +25,31 @@ def _process_args(self):
             raise NotFound(_('This event has been deleted.'))
 
 
-class RHDisplayEventBase(RHEventBase):
+class RHProtectedEventBase(RHEventBase):
+    """Base class for events that check if he user can access it.
```
ThiefMaster, line 29:
```suggestion
    """Base class for events that check if the user can access it.
```

indico/modules/events/controllers/base.py
```diff
@@ -25,21 +25,31 @@ def _process_args(self):
...
+
+    This includes unauthenticated users who have access to the event
+    for any other reason (e.g. due to a whitelisted IP address).
+    """
+    def _check_access(self):
```
ThiefMaster, line 34: add an empty line before this one (see [PEP257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings))

indico/web/client/js/react/components/principals/PrincipalListField.jsx
```diff
@@ -182,10 +178,33 @@ PrincipalListField.defaultProps = {
   withExternalUsers: false,
   withEventRoles: false,
   withCategoryRoles: false,
+  withRegistrants: false,
   eventId: null,
   readOnly: false,
 };
 
+export function AddPrincipalDropdown({text, options, onChange}) {
```
ThiefMaster, line 186: No need to export this. It's only used in here.

indico/web/client/js/react/components/principals/PrincipalListField.jsx
```diff
@@ -182,10 +178,33 @@ PrincipalListField.defaultProps = {
...
+    />
+  );
+}
+AddPrincipalDropdown.propTypes = {
+  text: PropTypes.string.isRequired,
+  options: PropTypes.array.isRequired,
+  onChange: PropTypes.func.isRequired,
+};
+
 export default React.memo(PrincipalListField);
```
ThiefMaster, line 208: Having this after `AddPrincipalDropdown` is a bit strange since it belongs to `PrincipalListField`..

indico/web/rh.py
```diff
@@ -327,6 +327,7 @@ def wrapper(*args, **kwargs):
 
 
 class RHProtected(RH):
+    """Base RH that requires the user to be authenticated."""
```
ThiefMaster, line 330:
```suggestion
    """Base RH that requires the user to be authenticated."""
    
```

PR #4531, opened 2020-07-13
Add link from event menu to editing hybrid view

ThiefMaster:
Currently the only way to get there is by going to an editable timeline and then using the link in the menu there...

We probably need a top-level menu entry with submenus for the editable types (otherwise we'd have to add some kind of dashboard).

Editing
- Papers
- Posters
- Slides

Alternatively we could also add a single link to Editing which goes to the editing view. But then we need a way to change the editable type there, and also add some page the user ends up when going there (since going to any specific editable type would be a bit weird, especially if more than one is enabled).

---

Note when testing: If event menu customization is enabled for an event, chances are good that you need to temporarily return `True` from `_menu_needs_recheck` and load the event page once since the event menu update logic only runs once for each indico version and event.

### DIFF
--- a/indico/modules/events/editing/__init__.py
+++ b/indico/modules/events/editing/__init__.py
@@ -81,6 +81,41 @@ def _get_cloners(sender, **kwargs):
     yield EditingSettingsCloner
 
 
+@signals.event.sidemenu.connect
+def _extend_event_menu(sender, **kwargs):
+    from indico.modules.events.editing.models.editable import EditableType
+    from indico.modules.events.layout.util import MenuEntryData
+
+    def _editing_visible(event, specific_type=None):
+        if not session.user or not event.has_feature('editing'):
+            return False
+
+        enabled_types = {
+            EditableType[et]
+            for et in event.editable_types
+            if (specific_type is None or specific_type.name == et)
+        }
+        if not enabled_types:
+            return False
+        elif event.can_manage(session.user, 'editing_manager'):
+            return True
+        return any(event.can_manage(session.user, et.editor_permission) for et in enabled_types)
+
+    def _make_visible(editable_type):
+        return lambda event: _editing_visible(event, editable_type)
+
+    yield MenuEntryData(title=_('Editing'), name='editing', position=8, visible=_editing_visible)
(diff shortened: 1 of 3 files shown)


indico/modules/events/layout/util.py
```diff
@@ -198,7 +201,7 @@ def _check_menu(event):
                       joinedload('children').load_only('id', 'parent_id', 'name', 'position')))
 
     existing = {entry.name: entry for entry in query}
-    pos_gen = count(start=(max(x.position for x in existing.itervalues() if not x.parent)))
+    pos_gen = count(start=(max(x.position for x in existing.itervalues() if not x.parent) + 1))
```
pferreir, line 204: Nice catch.

indico/modules/events/editing/__init__.py
```diff
@@ -81,6 +81,41 @@ def _get_cloners(sender, **kwargs):
...
+        enabled_types = {
+            EditableType[et]
+            for et in event.editable_types
+            if (specific_type is None or specific_type.name == et)
+        }
+        if not enabled_types:
+            return False
+        elif event.can_manage(session.user, 'editing_manager'):
+            return True
+        return any(event.can_manage(session.user, et.editor_permission) for et in enabled_types)
```
pferreir, line 102: Does this need to be a clouse? Wouldn't it be better to keep it at the top level if it doesn't use any of the external scope vars (importing `EditableType` inside it shouldn't hurt)?

ThiefMaster: No, doesn't *need* to be a closure, but I find it a bit ugly to define this globally in `__init__.py` even though it's completely local to the menu stuff. If we ever need it elsewhere I'd move it to a util though.

indico/modules/events/layout/util.py
ThiefMaster: didn't really cause problems, just looked a bit ugly when having the same position displayed twice in the menu editing view

indico/modules/events/editing/__init__.py
pferreir: Ah, it's because it's in `__init__.py`. You could always use `__all__`, but anyway, no big deal.

PR #4533, opened 2020-07-15
Import from event feature

### LINKED ISSUE #4518: Event cloning: Allow cloning into an existing event
The idea is to extend Event Cloner functionality to cloning into an existing event.

The implementation could allow cloning only modules that are empty in the target event (in order to avoid conflicts or replacing data).

meluru:
Closes #4518

image


indico/modules/events/operations.py
```diff
@@ -353,3 +373,17 @@ def sort_reviewing_questions(questions, new_positions):
         field.position = index
     db.session.flush()
     logger.info("Reviewing questions of %r reordered by %r", questions[0].event, session.user)
+
+
+def get_event_from_url(url):
+    data = urlparse(url)
+    if not all([data.scheme, data.netloc, data.path]):
+        raise ValueError(_('Invalid event URL'))
+    event_path = re.search('^/event/([0-9]+)/*', data.path)
```
ThiefMaster, line 382: I was thinking whether to suggest letting Flask parse the path instead of a regex but IIRC doing that is somewhat messy and we also don't really care *which* endpoint it matches as long as there's an event ID...

So keeping the regex is probably the better option here.

indico/modules/events/operations.py
```diff
@@ -187,6 +190,23 @@ def clone_event(event, start_dt, cloners, category=None):
...
+    Runs all required cloners.
+
+    :param source_event: The `Event` to clone data from;
+    :param target_event: The `Event` to clone data into;
+    :param cloners: A set containing the names of all enabled cloners.
+    """
+
+    # Run the modular cloning system
+    EventCloner.run_cloners(source_event, target_event, cloners, event_exists=True)
+    signals.event.cloned.send(source_event, new_event=target_event)
```
ThiefMaster, line 205: I think we need a new `existing_event` kwarg here since probably none of the logic we currently have when handling this signal should run in that case. Or just don't trigger the signal and maybe add a new `signals.event.imported` one?

indico/modules/events/cloning.py
```diff
@@ -163,6 +175,14 @@ def is_available(self):
         """
         return True
 
+    def has_conflicts(self, target_event):
+        """Checks for conflicts between source event and target event
+
+        Use this when cloning into an existing event to disable options
+        where target_event data would conflict with cloned data.
+        """
+        return False
```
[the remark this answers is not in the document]
meluru: Good point, I'd go for assuming conflicts by default. I'd prefer cloners to opt-out by `new_event_only = True`. It seems more like a special case when they don't allow importing.

indico/modules/events/clone.py
```diff
@@ -45,7 +51,10 @@ class EventPersonCloner(EventCloner):
     # We do not override `is_available` as we have cloners depending
     # on this internal cloner even if it won't clone anything.
 
-    def run(self, new_event, cloners, shared_data):
+    def has_conflicts(self, target_event):
+        return bool(target_event.persons)
```
plourenco, line 55: The same as my previous comment seems to be happening here.

indico/modules/events/contributions/clone.py
```diff
@@ -35,7 +35,10 @@ class ContributionTypeCloner(EventCloner):
     # We do not override `is_available` as we have cloners depending
     # on this internal cloner even if it won't clone anything.
 
-    def run(self, new_event, cloners, shared_data):
+    def has_conflicts(self, target_event):
+        return bool(target_event.contribution_types)
```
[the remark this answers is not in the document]
ThiefMaster: Good catch! `contribution_types ` is a `lazy='dynamic'` 😞 relationship so `.has_rows()` is indeed needed.

meluru:
@plourenco Thank you for the review, it was a good catch indeed. 
All updated now.

CHANGES.rst
```diff
@@ -92,6 +92,7 @@ Improvements
 - Add CfA setting to control whether authors can edit abstracts (:issue:`3431`)
 - Add CfA setting to control whether only speakers or also authors should
   get submission rights once the abstract gets accepted (:issue:`3431`)
+- Event modules can now be improted from other event (:issue:`4518`, thanks :user:`meluru`)
```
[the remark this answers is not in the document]
plourenco: Just missing clearing the line above 🙂

indico/modules/events/editing/clone.py
```diff
@@ -26,8 +26,12 @@ class EditingSettingsCloner(EventCloner):
     def is_visible(self):
         return self.old_event.type_ == EventType.conference
 
+    def has_conflicts(self, target_event):
+        return (bool(target_event.editing_file_types) or bool(target_event.editing_review_conditions)
```
plourenco, line 30: I just noticed that it doesn't seem to be possible to clone the event editing settings. I created a brand new event, but I can't import them because `bool(target_event.editing_file_types)` will always be Truthy, since it needs to have at least one publishable file-type, right? Perhaps we could disable this all-together, overwrite or ignore these cases?

CHANGES.rst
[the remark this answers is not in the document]
ThiefMaster: "from another event" or "from other events"

indico/modules/events/editing/clone.py
ThiefMaster: Should we just disable the importer for editing? Checking if there are only these three file types, unmodified, feels a bit much for this...

meluru: `bool(target_event.editing_file_types)` is False when editing is disabled in the target event, so import is possible only then. But, like @ThiefMaster said, checking the default file types seems like too much and overall importing editing settings is not so important so I can disable it.

meluru:
@plourenco 
I didn't manage to reproduce this problem here and nothing changed ui-wise in the last push.
Does it always happen for you? Maybe it's some corner case or it's browser-specific?

indico/modules/events/management/forms.py
```diff
@@ -407,3 +407,33 @@ class ProgramCodesForm(IndicoForm):
     session_block_template = StringField(_('Session blocks'))
     contribution_template = StringField(_('Contributions'))
     subcontribution_template = StringField(_('Subcontributions'))
+
+
+class ImportSourceEventForm(IndicoForm):
+    source_event_url = StringField(_('Event url'))
```
ThiefMaster, line 413:
```suggestion
    source_event_url =  URLField(_('Event URL'), [DataRequired(), URL()])
```

indico/modules/events/management/forms.py
```diff
@@ -407,3 +407,33 @@ class ProgramCodesForm(IndicoForm):
...
+            }
+            kwargs['formdata'] = ImmutableMultiDict(form_params)
+
+        super(ImportContentsForm, self).__init__(**kwargs)
+        self.selected_items.choices = [(option.name, option.friendly_name) for option in visible_options]
+        # We disable all cloners that are not available, handle only cloning to new events,
+        # have conflicts with target_event or any of their dependencies has conflicts with target_event
+        self.selected_items.disabled_choices = [option.name for option in visible_options if not option.is_available
+                                                or conflicts[option.name]
+                                                or option.__class__.new_event_only
```
ThiefMaster, line 438:
```suggestion
                                                or type(option).new_event_only
```

ThiefMaster: or better just use `option.new_event_only`; there's nothing wrong with accessing a class attribute through the instance. and that way a class could in principle override it as an actual `@property`

indico/core/signals/event/core.py
```diff
@@ -32,6 +32,11 @@
 the old event, the new event is passed in the `new_event` kwarg.
 """)
 
+imported = _signals.signal('imported', """
+Called when data is imported to an event. The *sender* is the `Event`
+data was imported into, source event is passed in the `source_event` kwarg.
```
ThiefMaster, line 37:
```suggestion
data was imported into, the source event is passed in the `source_event` kwarg.
```

indico/modules/attachments/clone.py
```diff
@@ -45,6 +48,9 @@ def run(self, new_event, cloners, shared_data):
             self._clone_attachments(new_event)
         db.session.flush()
 
+    def _has_content(self, event):
+        return bool(event.all_attachment_folders.filter(AttachmentFolder.attachments.any()).count())
```
ThiefMaster, line 52:
```suggestion
        return event.all_attachment_folders.filter(AttachmentFolder.attachments.any()).has_rows()
```

indico/modules/categories/schemas.py
```diff
@@ -0,0 +1,18 @@
...
+# LICENSE file for more details.
+
+from __future__ import unicode_literals
+
+from indico.core.marshmallow import mm
+from indico.modules.categories.models.categories import Category
+
+
+class CategorySchema(mm.ModelSchema):
+
```
ThiefMaster, line 15: no empty line here 👾

indico/modules/events/schemas.py
```diff
@@ -16,4 +20,14 @@ class EventPermissionsSchema(mm.Schema):
     acl_entries = PrincipalPermissionList(EventPrincipal, all_permissions=True)
 
 
+class EventDetailsSchema(mm.ModelSchema):
+
```
ThiefMaster, line 24: 👾

indico/modules/events/schemas.py
```diff
@@ -16,4 +20,14 @@ class EventPermissionsSchema(mm.Schema):
...
+    class Meta:
+        model = Event
+        fields = ('id', 'category', 'title', 'start_dt', 'end_dt')
+
+    category = fields.Nested(CategorySchema)
+
+
 event_permissions_schema = EventPermissionsSchema()
+event_details_schema = EventDetailsSchema()
```
ThiefMaster, line 33: let's not add new global schema instances (we pretty much stopped doing it for most newer schemas). better to just instantiate it when we need it

indico/modules/events/timetable/clone.py
```diff
@@ -35,13 +35,19 @@ def is_default(self):
...
+
+    def run(self, new_event, cloners, shared_data, event_exists=False):
         self._session_block_map = shared_data['sessions']['session_block_map']
         self._contrib_map = shared_data['contributions']['contrib_map']
         with db.session.no_autoflush:
             self._clone_timetable(new_event)
         db.session.flush()
 
+    def _has_content(self, event):
+        return bool(event.timetable_entries.count())
```
ThiefMaster, line 49:
```suggestion
        return event.timetable_entries.has_rows()
```

indico/modules/events/sessions/clone.py
```diff
@@ -51,30 +58,43 @@ def _clone_sessions(self, new_event):
                           subqueryload('acl_entries')))
         for old_sess in query:
             sess = Session()
-            sess.populate_from_attrs(old_sess, attrs)
-            sess.blocks = list(self._clone_session_blocks(old_sess.blocks))
+            if event_exists:
+                sess.populate_from_attrs(old_sess, attrs, skip='friendly_id')
```
ThiefMaster, line 62: This doesn't work, `skip` is supposed to be a collection... Anyway, you do not need to skip since you overwrite it in the next line anyway. I would add something like this before the loop though since it's cleaner:

```python
if event_exists:
    attrs.discard('friendly_id')
```

indico/modules/events/sessions/clone.py
```diff
@@ -51,30 +58,43 @@ def _clone_sessions(self, new_event):
...
+                sess.friendly_id = self._get_next_friendly_id(new_event.id)
+            else:
+                sess.populate_from_attrs(old_sess, attrs)
```
ThiefMaster, line 65: This whole snippet can then be replaced with this:

```python
sess.populate_from_attrs(old_sess, attrs)
if event_exists:
    sess.friendly_id = self._get_next_friendly_id(new_event.id)
```

indico/modules/events/sessions/clone.py
```diff
@@ -51,30 +58,43 @@ def _clone_sessions(self, new_event):
                           subqueryload('acl_entries')))
         for old_sess in query:
             sess = Session()
-            sess.populate_from_attrs(old_sess, attrs)
-            sess.blocks = list(self._clone_session_blocks(old_sess.blocks))
+            if event_exists:
+                sess.populate_from_attrs(old_sess, attrs, skip='friendly_id')
+                sess.friendly_id = self._get_next_friendly_id(new_event.id)
```
ThiefMaster, line 63: ⚠️ I already added some more comments around these lines, with the stuff in here most of them are probably obsolete.

Why do we need this? If you (actually) skip populating, the friendly IDs should be auto-assigned when the sessions get inserted.

Also, if there are any events we assume there's a conflict and do not allow importing. So in principle none of the special friendly-id handling should be needed. If you get an error due to non-unique friendly IDs it means we have a bug in our model and do not exclude deleted rows from the uniqueness check.

```python
    __auto_table_args = (db.Index(None, 'friendly_id', 'event_id', unique=True),
                         {'schema': 'events'})
```

should probably be

```python
    __auto_table_args = (db.Index(None, 'friendly_id', 'event_id', unique=True,
                                  postgresql_where=db.text('NOT is_deleted')),
                         {'schema': 'events'})
```

indico/modules/events/sessions/clone.py
```diff
@@ -51,30 +58,43 @@ def _clone_sessions(self, new_event):
...
 
-    def _clone_person_links(self, person_links):
+    def _clone_person_links(self, person_links, event_exists=False):
         attrs = get_simple_column_attrs(SessionBlockPersonLink)
         for old_link in person_links:
             link = SessionBlockPersonLink()
             link.populate_from_attrs(old_link, attrs)
-            link.person = self._person_map[old_link.person]
+            current_person = self._person_map[old_link.person]
+            if event_exists and current_person.user_id in self._new_event_persons.keys():
```
ThiefMaster, line 88:
```suggestion
            if event_exists and current_person.user_id in self._new_event_persons:
```

no need for `.keys()` to check if a key exists

indico/modules/events/clone.py
```diff
@@ -45,7 +51,10 @@ class EventPersonCloner(EventCloner):
     # We do not override `is_available` as we have cloners depending
     # on this internal cloner even if it won't clone anything.
 
-    def run(self, new_event, cloners, shared_data):
+    def has_conflicts(self, target_event):
+        return bool(target_event.persons.has_rows())
```
ThiefMaster, line 55:
```suggestion
        return target_event.persons.has_rows()
```

indico/modules/events/clone.py
```diff
@@ -98,13 +113,21 @@ class EventProtectionCloner(EventCloner):
...
     uses = {'event_roles', 'registration_forms'}
 
-    def run(self, new_event, cloners, shared_data):
+    def has_conflicts(self, target_event):
+        if target_event.access_key != '':
+            return True
+        entries = list(target_event.acl_entries)
+        if len(entries) != 1 or entries[0].user != target_event.creator:
+            return True
+        return False
```
ThiefMaster, line 122:
```suggestion
        return len(entries) != 1 or entries[0].user != target_event.creator
```

indico/modules/events/cloning.py
```diff
@@ -163,6 +175,14 @@ def is_available(self):
         """
         return True
 
+    def has_conflicts(self, target_event):
+        """Checks for conflicts between source event and target event
```
ThiefMaster, line 179:
```suggestion
        """Check for conflicts between source event and target event.
```

indico/modules/events/contributions/clone.py
```diff
@@ -35,7 +35,10 @@ class ContributionTypeCloner(EventCloner):
     # We do not override `is_available` as we have cloners depending
     # on this internal cloner even if it won't clone anything.
 
-    def run(self, new_event, cloners, shared_data):
+    def has_conflicts(self, target_event):
+        return bool(target_event.contribution_types.has_rows())
```
ThiefMaster, line 39:
```suggestion
        return target_event.contribution_types.has_rows()
```

indico/modules/events/management/blueprint.py
```diff
@@ -45,6 +45,8 @@
 # Cloning
 _bp.add_url_rule('/clone', 'clone', cloning.RHCloneEvent, methods=('GET', 'POST'))
 _bp.add_url_rule('/clone/preview', 'clone_preview', cloning.RHClonePreview, methods=('GET', 'POST'))
+_bp.add_url_rule('/import', 'import', cloning.RHImportFromEvent, methods=('GET', 'POST'))
+_bp.add_url_rule('/import/event-details', 'import_event_details', cloning.RHImportEventDetails, methods=('GET', 'POST'))
```
ThiefMaster, line 49: I don't think this one uses anything but POST

```suggestion
_bp.add_url_rule('/import/event-details', 'import_event_details', cloning.RHImportEventDetails, methods=('POST',))
```

indico/modules/events/management/controllers/cloning.py
```diff
@@ -196,3 +198,63 @@ def _process(self):
                         for c in EventCloner.get_cloners(self.event)}
         return jsonify_template('events/management/clone_event.html', event=self.event, step=step, form=form,
                                 cloner_dependencies=dependencies, **tpl_args)
+
+
+class RHImportFromEvent(RHManageEventBase):
+    """Imports from another event."""
+
+    ALLOW_LOCKED = True
```
ThiefMaster, line 206: This shouldn't be here; locked  events do not allow anything that modifies it.

indico/modules/events/sessions/clone.py
meluru: Yes, there was a friendly_id uniqueness issue. I didn't think of that solution, but it should be much better this way.

meluru:
All updated now! I removed all `friendly_id` related changes and fixed the index on sessions.

indico/migrations/versions/20200804_1308_497c61b68050_fix_session_friendly_id_index.py
```diff
@@ -0,0 +1,29 @@
...
+        DROP INDEX events.ix_uq_sessions_friendly_id_event_id;
+        CREATE UNIQUE INDEX ix_uq_sessions_friendly_id_event_id ON events.sessions USING btree (friendly_id, event_id) WHERE NOT is_deleted;
+    ''')
+
+
+def downgrade():
+    op.execute('''
+        DROP INDEX events.ix_uq_sessions_friendly_id_event_id;
+        CREATE UNIQUE INDEX ix_uq_sessions_friendly_id_event_id ON events.sessions USING btree (friendly_id, event_id);
+    ''')
```
ThiefMaster, line 29: I think this can all be done with the high-level alembic API to drop/create indexes.

(but that said, I'd also be tempted to just put SQL there since it's less work :P)

ThiefMaster: FYI, I just cherry-picked this commit into master and amended it with the suggested change (71539618f4). that way we have less PRs with alembic revisions and it's something i'd like to have included when deploying next week, while i want to test this PR more thoroughly first :)

meluru: Great, thanks a lot. I've rebased the PR now :)

plourenco:
It would be an interesting idea if we added an indication on why some fields are greyed out, as in: "You can't import because you already made changes there" or "Only unmodified/empty modules of the event can be imported". Could even be something generic, it would be more user friendly.

meluru:
@plourenco that's a good point. I'll add a message. :+1: I think I should also remove from the list the modules that don't support importing at all, because having them greyed out might be confusing.

meluru:
Updated. Let me know if the message is alright.

image

indico/modules/events/management/controllers/cloning.py
```diff
@@ -196,3 +198,62 @@ def _process(self):
...
+            step = step - 1
+
+        elif step > 2:
+            # last step - perform actual cloning
+            source_event = get_event_from_url(request.form.get('source_event_url'))
+            form = ImportContentsForm(source_event, self.event)
+
+            if form.validate_on_submit():
+                updated_event = clone_into_event(source_event, self.event, set(form.selected_items.data))
+                flash(_('Data properly imported!'), 'success')
```
ThiefMaster, line 231: properly sounds weird in this context. "successful" fits better IMO

```suggestion
                flash(_('Import successful!'), 'success')
```

indico/modules/events/management/controllers/cloning.py
```diff
@@ -196,3 +198,62 @@ def _process(self):
...
+
+    def _process(self):
+        from indico.modules.events.schemas import EventDetailsSchema
+        schema = EventDetailsSchema()
+        form = ImportSourceEventForm()
+        try:
+            event = get_event_from_url(form.source_event_url.data)
+            if event == self.event:
+                raise ValueError(_('Cannot import from the same event'))
+            if event.type != self.event.type:
```
ThiefMaster, line 253:
```suggestion
            if event.type_ != self.event.type_:
```

`type_` is the enum, `type` is its name (for legacy reasons). No need to use the latter in new code ;)

indico/modules/events/management/templates/_action_menu.html
```diff
@@ -100,6 +100,12 @@
         <div class="toolbar">
             {% if event.can_manage(session.user) %}
                 <div class="group">
+                    <a class="i-button icon-upload" title="{% trans %}Import from event{% endtrans %}"
+                        data-href="{{ url_for('event_management.import', event) }}"
+                        data-title="{% trans %}Import from Event{% endtrans %}"
+                        data-ajax-dialog>
+                     {% trans %}Import{% endtrans %}
```
ThiefMaster, line 107:
```suggestion
                       data-href="{{ url_for('event_management.import', event) }}"
                       data-title="{% trans %}Import from Event{% endtrans %}"
                       data-ajax-dialog>
                        {% trans %}Import{% endtrans %}
```

indico/modules/events/management/templates/import_event.html
```diff
@@ -0,0 +1,113 @@
...
+                        {% endset %}
+                        {% set event_category %}
+                            <div class="import-event-category"></div>
+                        {% endset %}
+                        <div>
+                            {% trans %}You will import data from the following event:{% endtrans %}
+                        </div>
+                        {{ event_title }}
+                        {{ event_dates }}
+                        {{ event_category }}
```
ThiefMaster, line 62: I don't think we need those set blocks:

```suggestion
                        <div>
                            {% trans %}You will import data from the following event:{% endtrans %}
                        </div>
                        <div class="import-event-title"></div>
                        <div class="import-event-dates"></div>
                        <div class="import-event-category"></div>
```

indico/modules/events/notes/clone.py
```diff
@@ -37,6 +40,9 @@ def run(self, new_event, cloners, shared_data):
             self._clone_notes(new_event)
         db.session.flush()
 
+    def _has_content(self, event):
+        return bool(event.all_notes.filter_by(is_deleted=False).count())
```
ThiefMaster, line 44:
```suggestion
        return event.all_notes.filter_by(is_deleted=False).has_rows()
```

indico/modules/events/operations.py
```diff
@@ -187,6 +187,23 @@ def clone_event(event, start_dt, cloners, category=None):
     return new_event
 
 
+def clone_into_event(source_event, target_event, cloners):
+    """Clone data into an exisiting event.
```
ThiefMaster, line 191:
```suggestion
    """Clone data into an existing event.
```

indico/modules/events/cloning.py
```diff
@@ -163,6 +175,14 @@ def is_available(self):
         """
         return True
 
+    def has_conflicts(self, target_event):
+        """Check for conflicts between source event and target event
+
+        Use this when cloning into an existing event to disable options
+        where target_event data would conflict with cloned data.
```
ThiefMaster, line 182:
```suggestion
        where ``target_event`` data would conflict with cloned data.
```

indico/modules/events/cloning.py
```diff
@@ -163,6 +175,14 @@ def is_available(self):
         """
         return True
 
+    def has_conflicts(self, target_event):
+        """Check for conflicts between source event and target event
```
ThiefMaster, line 179:
```suggestion
        """Check for conflicts between source event and target event.
```

indico/modules/events/cloning.py
```diff
@@ -58,6 +58,9 @@ class EventCloner(object):
     #: would result in people having to explicitly enable the other
     #: cloner even if it makes no sense to not run it.
     always_available_dep = False
+    #: Whether cloner allows cloning only into new events and excludes
+    #: cloning into exisitng events
```
ThiefMaster, line 62:
```suggestion
    #: Whether this cloner only allows cloning into new events and
    #: is not available when importing into an existing event.
```

indico/modules/events/clone.py
```diff
@@ -25,13 +25,19 @@ class EventLocationCloner(EventCloner):
...
+    def has_conflicts(self, target_event):
+        return self._has_content(target_event)
+
+    def run(self, new_event, cloners, shared_data, event_exists=False):
         with db.session.no_autoflush:
             self._clone_location(new_event)
         db.session.flush()
 
+    def _has_content(self, event):
+        return bool(event.has_location_info)
```
ThiefMaster, line 39:
```suggestion
        return event.has_location_info
```

it's already a bool ;)

ThiefMaster:
The last commit message has a typo in "existing" (exisiting)

ThiefMaster:
I pushed some small fixes/improvements I noticed while testing.

When trying to import the timetable from a larger event from the past (indico workshop 2.0) I'm currently getting an error about the timetable being inconsistent; I'll have a look and see if it's a quick fix, otherwise I'll ask you to look into it :p

ThiefMaster:
The timetable error happens because my target event was shorter. Should we make it a conflict in the timetable importer when the source event is longer than the target event?

ThiefMaster:
I added a commit doing just that. Importing the indico workshop event works fine now, I'll test with some other events but I think it's pretty much ready to be merged!

ThiefMaster:
After discussing with the team we thought the button next to clone is a bit too prominent so I moved it inside the cog menu:

image

Not completely sure about the icon... but so far I couldn't find any better one. The import/export icons in icomoon don't look amazing either. If you have any good idea please let me know!

ThiefMaster:
@meluru from our side everything is fine now and I think this PR is ready to be merged. Maybe you can have a quick look at the changes I made before I merge it?

PR #4534, opened 2020-07-15
Build changes from getattr instead of passed value

OmeGak:
This makes `populate_from_dict` correctly build the changes for properties where `getattr` returns something different than what `setattr` sets.

### DIFF
--- a/indico/core/db/sqlalchemy/util/models.py
+++ b/indico/core/db/sqlalchemy/util/models.py
@@ -212,10 +212,11 @@ class IndicoModel(Model):
             if not hasattr(cls, key):
                 raise ValueError("{} has no attribute '{}'".format(cls.__name__, key))
             old_value = getattr(self, key, None)
-            if old_value != value:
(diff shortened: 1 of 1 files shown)


indico/core/db/sqlalchemy/util/models.py
```diff
@@ -212,10 +212,11 @@ def populate_from_dict(self, data, keys=None, skip=None):
             if not hasattr(cls, key):
                 raise ValueError("{} has no attribute '{}'".format(cls.__name__, key))
             old_value = getattr(self, key, None)
-            if old_value != value:
-                # XXX: we copy because of https://bitbucket.org/zzzeek/sqlalchemy/issues/3913/
-                changed[key] = (copy(old_value), value)
-                setattr(self, key, value)
+            setattr(self, key, value)
+            new_value = getattr(self, key, None)
```
ThiefMaster, line 216: i don't think you need the `, None` here - after the setattr the attribute should always exist

PR #4535, opened 2020-07-15
Improve IndicoDateField facilities

OmeGak:
Added:
- `DateRange` validator
- `LinkedDate` validator

Improved:
- `IndicoDateField` now renders with the React date picker

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -127,6 +127,8 @@ Internal Changes
   :user:`omegak`)
 - Add ``before-registration-actions`` template hook (:issue:`4524`, thanks
   :user:`omegak`)
+- Add ``LinkedDate`` and ``DateRange`` form field validators (:issue:`4535`,
+  thanks :user:`omegak`)
 
 
 ----
--- a/indico/web/forms/fields/datetime.py
+++ b/indico/web/forms/fields/datetime.py
@@ -219,7 +219,7 @@ class IndicoDateTimeField(DateTimeField):
     during form submission.
     """
 
-    widget = JinjaWidget('forms/datetime_widget.html', single_line=True)
+    widget = JinjaWidget('forms/datetime_widget.html', single_line=True, single_kwargs=True)
 
     def __init__(self, *args, **kwargs):
         self._timezone = kwargs.pop('timezone', None)
--- a/indico/web/forms/fields/datetime.py
+++ b/indico/web/forms/fields/datetime.py
@@ -23,7 +23,7 @@ from indico.core.config import config
 from indico.util.date_time import localize_as_utc, relativedelta
 from indico.util.i18n import _, get_current_locale
 from indico.web.forms.fields import JSONField
-from indico.web.forms.validators import DateTimeRange, LinkedDateTime
+from indico.web.forms.validators import DateRange, DateTimeRange, LinkedDate, LinkedDateTime
 from indico.web.forms.widgets import JinjaWidget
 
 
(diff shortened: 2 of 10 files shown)


CHANGES.rst
```diff
@@ -127,6 +127,8 @@ Internal Changes
   :user:`omegak`)
 - Add ``before-registration-actions`` template hook (:issue:`4524`, thanks
   :user:`omegak`)
+- Improve ``IndicoDateField`` rendering and validators (:issue:`4535`, thanks
+  :user:`omegak`)
```
ThiefMaster, line 131: hm not sure if it's useful... it's not really relevant for end users nor plugin developers

OmeGak: It would still be relevant to specify that `DateRange ` and `LinkedDate` validators have been added.

indico/web/templates/forms/date_widget.html
```diff
@@ -1,12 +1,32 @@
 {% extends 'forms/base_widget.html' %}
 
 {% block html %}
-    <input type="text" id="{{ field.id }}" value="{{ field._value() }}" name="{{ field.name }}"
-           placeholder="DD/MM/YYYY" {{ input_args | html_params }}>
+    <div class="datetime-widget i-form-field-fixed-width">
+        {# the hidden field has autofocus on purpose to prevent our JS from focusing the date picker #}
+        <input type="hidden" name="{{ field.name }}" id="{{ field.id }}-datestorage" autofocus
+               {% if field.data and not field.date_missing %}value="{{ field.data|format_date('code') }}"{% endif %}>
```
ThiefMaster, line 7: We need to keep `{{ input_args | html_params }}` on the field, otherwise validators like HiddenUnless don't work anymore (can be tested in the clone with pattern dialog or when creating a reminder) since this adds the data attributes enabling the HiddenUnless logic.

The same bug probably exists in the datetime variant, but I guess we never use this with HiddenUnless and thus didn't notice it there.

indico/web/forms/fields/datetime.py
```diff
@@ -179,9 +179,36 @@ class IndicoDateField(DateField):
     widget = JinjaWidget('forms/date_widget.html', single_line=True, single_kwargs=True)
 
     def __init__(self, *args, **kwargs):
+        self.allow_clear = kwargs.pop('allow_clear', True)
```
ThiefMaster, line 182: Let's default to `False` here since we did not have a clear button before, and only make it clearable in `AttachmentPackageForm` and `VCRoomListFilterForm` - in all the other places the field is required.


Another option would be defaulting to `not self.flags.required` (if the flags from validators are available at this point - might have to do it after the super call)

indico/web/templates/forms/date_widget.html
```diff
@@ -1,12 +1,32 @@
 {% extends 'forms/base_widget.html' %}
 
 {% block html %}
-    <input type="text" id="{{ field.id }}" value="{{ field._value() }}" name="{{ field.name }}"
-           placeholder="DD/MM/YYYY" {{ input_args | html_params }}>
+    <div class="datetime-widget i-form-field-fixed-width">
```
ThiefMaster, line 4: image

There's some weird spacing before the clear button. Seems to be coming from `.datetime-widget .clear-pickers`. This fixes it:

```patch
diff --git a/indico/web/client/styles/partials/_inputs.scss b/indico/web/client/styles/partials/_inputs.scss
index 364df0f6e8..66f06f4efc 100644
--- a/indico/web/client/styles/partials/_inputs.scss
+++ b/indico/web/client/styles/partials/_inputs.scss
@@ -276,7 +276,8 @@ $dropdown-transition: $dropdown-transition-step ease-out;
   }
 }

-.datetime-widget {
+.datetime-widget,
+.date-widget {
   .rc-time-picker-input {
     width: 100% !important;
     font-size: 1em;
@@ -348,9 +349,18 @@ $dropdown-transition: $dropdown-transition-step ease-out;
   }
 }

-.i-form .form-group .form-field .datetime-widget input[type='text'] {
-  font-size: 1em;
-  width: 150px;
+.date-widget .clear-pickers {
+  left: 0;
+}
+
+.i-form .form-group .form-field {
+  .datetime-widget,
+  .date-widget {
+    input[type='text'] {
+      font-size: 1em;
+      width: 150px;
+    }
+  }
 }

 .has-warning {
diff --git a/indico/web/templates/forms/date_widget.html b/indico/web/templates/forms/date_widget.html
index 1107fcb8b7..403f88ce87 100644
--- a/indico/web/templates/forms/date_widget.html
+++ b/indico/web/templates/forms/date_widget.html
@@ -1,10 +1,11 @@
 {% extends 'forms/base_widget.html' %}

 {% block html %}
-    <div class="datetime-widget i-form-field-fixed-width">
+    <div class="date-widget i-form-field-fixed-width">
         {# the hidden field has autofocus on purpose to prevent our JS from focusing the date picker #}
         <input type="hidden" name="{{ field.name }}" id="{{ field.id }}-datestorage" autofocus
-               {% if field.data and not field.date_missing %}value="{{ field.data|format_date('code') }}"{% endif %}>
+               {% if field.data and not field.date_missing %}value="{{ field.data|format_date('code') }}"{% endif %}
+               {{ input_args | html_params }}>
         <span id="{{ field.id }}"></span>
     </div>
 {% endblock %}
```

PR #4536, opened 2020-07-15
Add Flask-Caching

ThiefMaster:
Not used by indico itself yet, but useful if e.g. a flask-multipass extension (flask-multipass-cern ;)) wants to use caching.

It could also be the successor of the legacy GenericCache in 3.0.


pferreir:
Sounds like a great idea.

PR #4537, opened 2020-07-16
Allow registration access in restricted events if they have registrants in the ACL

### LINKED ISSUE #4528: Allow registration access in restricted events if they have registrants in the ACL
This is a follow-up on #2460 and #4477.

When accessing an event where the usual access check fails but the ACL contains registration forms, then redirect to that form (if it's exactly one) or the list of available registration forms (if more than one), and show a message there indicating that the event is restricted to registered participants.

Some important points:

- Events may have a custom logo/css; those endpoints need to skip the usual event can_access check if the ACL contains any regform (I would suggest adding a new method/property on `Event` that checks for this)
- We need to make sure the event header does not show too much information (no filter button in the header for example, since that would reveal the list of sessions/contributions)
- No conference menu should be shown or if we do show it, only the registration item should be visible (this would most likely require adding `event.can_access` checks to all the *other* menu items' visibility functions). Always using the meeting layout could be an option as well, but maybe not what event organizers expect

Leats:
Closes #4528.

### DIFF
--- a/indico/modules/events/layout/controllers/layout.py
+++ b/indico/modules/events/layout/controllers/layout.py
@@ -18,7 +18,7 @@ from wtforms.validators import DataRequired
 
 from indico.core.db import db
 from indico.modules.events import EventLogKind, EventLogRealm
-from indico.modules.events.controllers.base import RHDisplayEventBase
+from indico.modules.events.controllers.base import RegistrationRequired, RHDisplayEventBase
 from indico.modules.events.layout import layout_settings, logger, theme_settings
 from indico.modules.events.layout.forms import (ConferenceLayoutForm, CSSForm, CSSSelectionForm,
                                                 LectureMeetingLayoutForm, LogoForm)
@@ -203,7 +203,6 @@ class RHLayoutCSSUpload(RHLayoutBase):
 
 
 class RHLayoutCSSDelete(RHLayoutBase):
-
     def _process(self):
         self.event.stylesheet = None
         self.event.stylesheet_metadata = None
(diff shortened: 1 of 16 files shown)


indico/modules/events/layout/controllers/layout.py
```diff
@@ -234,6 +233,13 @@ def _process(self):
 
 
 class RHLogoDisplay(RHDisplayEventBase):
+    def _check_access(self):
+        try:
+            RHDisplayEventBase._check_access(self)
+        except RegistrationRequired:
+            # in case registrants need to register they should be able to see the logo
```
pferreir, line 240: :+1:

indico/modules/events/registration/templates/display/regform_display.html
```diff
@@ -26,6 +26,13 @@
 
 
 {% block content %}
+    {% if not event.can_access(session.user) %}
```
pferreir, line 29: I guess there is no other situation in which `.can_access` returns `False`, since otherwise this template wouldn't even be rendered? I was wondering whether we should check for `.has_regform_in_acl` anyway, or just add a comment explaining that...

indico/modules/events/layout/controllers/layout.py
pferreir, line 240: 👍

indico/modules/events/registration/templates/display/regform_list.html
```diff
@@ -4,12 +4,21 @@
...
+{% from 'message_box.html' import message_box %}
+
 {% block subtitle %}
     {% trans %}Available forms{% endtrans %}
 {% endblock %}
 
 {% block content %}
-{% set any_published_registrations = regforms|selectattr('publish_registration_count')|any %}
+    {% set any_published_registrations = regforms|selectattr('publish_registration_count')|any %}
+    {% if not event.can_access(session.user) %}
```
pferreir, line 15: Same as my point above.

indico/modules/events/registration/templates/display/regform_display.html
ThiefMaster: comment sounds reasonable - and you are right, there's no other way to get there if you do not have access

PR #4539, opened 2020-07-16
Add a missing headers check to CI

plourenco:
Didn't want to modify this script too much so I just added a simple return that tracks the exit code. 
It only checks if headers are missing, still performs the year update. I just thought that was irrelevant enough to extend this. Suggestions are welcome!

Also adds missing file headers.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,10 +14,7 @@ indent_size=2
 [{*.css,*.scss}]
 indent_size=2
 
-[{*.yml,*.yaml}]
-indent_size=2
-
-[{package.json,.babelrc}]
+[{*.yml,*.yaml,*.json,.babelrc}]
 indent_size=2
 
 [Makefile]
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,7 +14,7 @@ indent_size=2
 [{*.css,*.scss}]
 indent_size=2
 
-[{*.yml,*.yaml,*.json,.babelrc}]
+[{*.yml,*.yaml,*.json}]
 indent_size=2
 
 [Makefile]
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ script:
   - FORCE_COLOR=1 npx react-jsx-i18n extract --ext jsx indico/web/client/ indico/modules/ > /dev/null
   - isort -rc --diff --check-only indico/
   - python bin/maintenance/update_backrefs.py --ci
-  - python bin/maintenance/update_header.py indico
+  - python bin/maintenance/update_header.py indico --ci
   - flake8
   - npx eslint --ext .js --ext .jsx indico/modules/rb/ indico/modules/events/logs/ indico/web/client/js/react/
     indico/modules/users/
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ script:
   - FORCE_COLOR=1 npx react-jsx-i18n extract --ext jsx indico/web/client/ indico/modules/ > /dev/null
   - isort -rc --diff --check-only indico/
   - python bin/maintenance/update_backrefs.py --ci
+  - python bin/maintenance/update_header.py indico
   - flake8
   - npx eslint --ext .js --ext .jsx indico/modules/rb/ indico/modules/events/logs/ indico/web/client/js/react/
     indico/modules/users/
--- a/bin/maintenance/update_header.py
+++ b/bin/maintenance/update_header.py
@@ -126,6 +126,7 @@ def _update_header(project, file_path, year, substring, regex, data):
         print(cformat('%{green!}Updating header of %{blue!}{}').format(os.path.relpath(file_path)))
         with open(file_path, 'wb') as file_write:
             file_write.write(content)
+        return True
     elif not found:
         print('Missing header in {}'.format(os.path.relpath(file_path)))
         return True
--- a/bin/maintenance/update_header.py
+++ b/bin/maintenance/update_header.py
@@ -127,7 +127,7 @@ def _update_header(project, file_path, year, substring, regex, data):
         with open(file_path, 'wb') as file_write:
             file_write.write(content)
     elif not found:
-        print(cformat('%{yellow}Missing header in %{yellow!}{}').format(os.path.relpath(file_path)))
+        print('Missing header in {}'.format(os.path.relpath(file_path)))
         return True
 
 
--- a/bin/maintenance/update_header.py
+++ b/bin/maintenance/update_header.py
@@ -95,7 +95,7 @@ to the current year.
 You need to specify which project it is (one of {projects}) to the correct
 headers are used.
 
-You can specify a year (1000-2999) to update to as well as a file or directory.
+You can specify a year (1000+) to update to as well as a file or directory.
 This will update all the supported files in the scope including those not tracked
 by git. If the directory does not contain any supported files (or if the file
 specified is not supported) nothing will be updated.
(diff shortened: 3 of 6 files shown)


bin/maintenance/update_header.py
```diff
@@ -125,6 +125,7 @@ def _update_header(project, file_path, year, substring, regex, data):
             file_write.write(content)
     elif not found:
         print(cformat('%{yellow}Missing header in %{yellow!}{}').format(os.path.relpath(file_path)))
+        return True
```
ThiefMaster, line 128: I'd prefer an explicit `return False` in the other cases. True vs None is a bit ugly ;)

ThiefMaster: The "Updating header" case should also cause a CI failure IMO

bin/maintenance/update_header.py
```diff
@@ -179,31 +181,35 @@ def main():
...
         for root, _, filenames in os.walk(path):
             for filename in filenames:
                 if not blacklisted(path, root):
-                    update_header(project, os.path.join(root, filename), year)
+                    error = update_header(project, os.path.join(root, filename), year) or error
     elif file_ is not None:
         print(cformat("Updating headers to the year %{yellow!}{year}%{reset} for the file "
                       "%{yellow!}{file}%{reset}...").format(year=year, file=file_))
-        update_header(project, file_, year)
+        error = update_header(project, file_, year) or error
```
ThiefMaster, line 196:
```suggestion
        if update_header(project, file_, year):
            error = True
```

bin/maintenance/update_header.py
```diff
@@ -179,31 +181,35 @@ def main():
...
     else:
         print(cformat("Updating headers to the year %{yellow!}{year}%{reset} for all "
                       "git-tracked files...").format(year=year))
         try:
             for path in subprocess.check_output(['git', 'ls-files']).splitlines():
                 path = os.path.abspath(path)
                 if not blacklisted(os.getcwd(), os.path.dirname(path)):
-                    update_header(project, path, year)
+                    error = update_header(project, path, year) or error
```
ThiefMaster, line 204: same here. IMO it looks better and is a bit more explicit

bin/maintenance/update_header.py
```diff
@@ -179,31 +181,35 @@ def main():
...
         except subprocess.CalledProcessError:
             print(cformat('%{red!}[ERROR] you must be within a git repository to run this script.'), file=sys.stderr)
             print(USAGE)
             sys.exit(1)
 
+    if error:
+        sys.exit(1)
```
ThiefMaster, line 211: I'd prefer to put this behind a `--ci` flag like we did in `update_backrefs.py`.
Having a non-zero exit code in a script that does its job as intended and without any errors just doesn't seem right.

bin/maintenance/update_header.py
plourenco: I was trying to avoid nesting it further. But that works too!

bin/maintenance/update_header.py
plourenco: Alright, I can do that. I previously made it True to avoid changing all returns (quite a few).

bin/maintenance/update_header.py
plourenco: Sounds good 👍

bin/maintenance/update_header.py
ThiefMaster: I see only one in this function (plus the implicit one at the end where nothing is returned).. ;)

plourenco: I meant `update_header`. We should return True (success) if e.g. `if ext not in SUPPORTED_FILES` right?

ThiefMaster: I'd return False ("nothing missing / to be done") in those cases, i.e. returning `True` only when there's something to be done...

plourenco: I see what you mean now.

.editorconfig
```diff
@@ -14,10 +14,7 @@ indent_size=2
 [{*.css,*.scss}]
 indent_size=2
 
-[{*.yml,*.yaml}]
-indent_size=2
-
-[{package.json,.babelrc}]
+[{*.yml,*.yaml,*.json,.babelrc}]
```
ThiefMaster, line 17: We don't have a `.babelrc` anymore since ac6760d6dc

.travis.yml
```diff
@@ -23,6 +23,7 @@ script:
   - FORCE_COLOR=1 npx react-jsx-i18n extract --ext jsx indico/web/client/ indico/modules/ > /dev/null
   - isort -rc --diff --check-only indico/
   - python bin/maintenance/update_backrefs.py --ci
+  - python bin/maintenance/update_header.py indico
```
ThiefMaster, line 26: `--ci` even though it doesn't really matter since we don't use travis anymore on master

bin/maintenance/update_header.py
```diff
@@ -173,36 +161,46 @@ def blacklisted(root, path, _cache={}):
     return _cache[orig_path]
 
 
-def main():
-    if '-h' in sys.argv[1:] or '--help' in sys.argv[1:] or not sys.argv[1:]:
-        print(USAGE)
+@click.command(help=USAGE)
+@click.option('--ci', is_flag=True, help='Indicate that the script is running during CI and should use a non-zero '
+                                         'exit code unless all headers were already up to date.')
+@click.argument('project', nargs=1)
```
ThiefMaster, line 167: Isn't `nargs=1` the default?

bin/maintenance/update_header.py
```diff
@@ -124,41 +127,26 @@ def _update_header(project, file_path, year, substring, regex, data):
...
-        return
-    _update_header(project, file_path, year, SUBSTRING, SUPPORTED_FILES[ext]['regex'], SUPPORTED_FILES[ext]['format'])
+        return False
+    return _update_header(project, file_path, year, SUBSTRING, SUPPORTED_FILES[ext]['regex'],
+                          SUPPORTED_FILES[ext]['format'])
 
 
-def _process_args(args):
+def _process_year(ctx, param, value):
     year_regex = re.compile('^[12][0-9]{3}$')  # Year range 1000 - 2999
```
ThiefMaster, line 145: i don't think we need to keep this. it was just to be "smart" to have optional positional arguments that aren't at the end

bin/maintenance/update_header.py
```diff
@@ -173,36 +161,46 @@ def blacklisted(root, path, _cache={}):
...
 
-def main():
-    if '-h' in sys.argv[1:] or '--help' in sys.argv[1:] or not sys.argv[1:]:
-        print(USAGE)
+@click.command(help=USAGE)
+@click.option('--ci', is_flag=True, help='Indicate that the script is running during CI and should use a non-zero '
+                                         'exit code unless all headers were already up to date.')
+@click.argument('project', nargs=1)
+@click.argument('year', nargs=1, required=False, callback=_process_year)
+@click.argument('path', nargs=1, required=False, type=click.Path(exists=True))
```
ThiefMaster, line 169: What about making these two args options instead of arguments? Having them as positional arguments isn't particularly intuitive anyway.

bin/maintenance/update_header.py
plourenco: But isn't it intended to also check if you provided a year in the right format instead of, say, `foo`?

ThiefMaster: I'd say requiring a numeric value is fine for that... it's not like this tool is exposed to untrusted users. OR a simple `2020 <= year <= 2999` (as if indico (or humanity) will still be a thing by then :D)

ThiefMaster: there's actually a click validator for this: https://click.palletsprojects.com/en/7.x/api/#click.IntRange

bin/maintenance/update_header.py
plourenco: Is it, got carried by the docs.

bin/maintenance/update_header.py
plourenco: I was wondering the same previously.

PR #4551, opened 2020-07-21
Update AccessControlListField principal list to React

### LINKED ISSUE #4505: Allow selecting event/category roles and registrants in attachment/folder ACLs

plourenco:
Closes #4505 

- [x] Requires testing
- [x] Checking which legacy classes can already re-use the new version.
- [x] Update WTFPrincipalListField spreaded props
- [x] Update PrincipalListField props (such as required, onFocus, blur)

### DIFF
--- a/indico/modules/events/abstracts/forms.py
+++ b/indico/modules/events/abstracts/forms.py
@@ -36,7 +36,7 @@ from indico.web.forms.base import FormDefaults, IndicoForm, generated_data
 from indico.web.forms.fields import (EditableFileField, EmailListField, HiddenEnumField, HiddenFieldList,
                                      IndicoDateTimeField, IndicoEnumSelectField, IndicoMarkdownField,
                                      IndicoQuerySelectMultipleCheckboxField, IndicoQuerySelectMultipleField,
-                                     IndicoRadioField, PrincipalField, PrincipalListField)
+                                     IndicoRadioField, PrincipalField, _LegacyPrincipalListField)
 from indico.web.forms.util import inject_validators
 from indico.web.forms.validators import HiddenUnless, LinkedDateTime, SoftLength, UsedIf, WordCount
 from indico.web.forms.widgets import JinjaWidget, SwitchWidget
@@ -127,9 +127,9 @@ class AbstractSubmissionSettingsForm(IndicoForm):
                                                     enum=SubmissionRightsType, sorted=True,
                                                     description=_("Specify who will get contribution submission rights "
                                                                   "once an abstract has been accepted"))
-    authorized_submitters = PrincipalListField(_("Authorized submitters"),
-                                               description=_("These users may always submit abstracts, even outside "
-                                                             "the regular submission period"))
+    authorized_submitters = _LegacyPrincipalListField(_("Authorized submitters"),
+                                                      description=_("These users may always submit abstracts, "
+                                                                    "even outside the regular submission period."))
     submission_instructions = IndicoMarkdownField(_('Instructions'), editor=True,
                                                   description=_("These instructions will be displayed right before the "
                                                                 "submission form"))
--- a/indico/modules/events/abstracts/forms.py
+++ b/indico/modules/events/abstracts/forms.py
@@ -36,7 +36,8 @@ from indico.web.forms.base import FormDefaults, IndicoForm, generated_data
 from indico.web.forms.fields import (EditableFileField, EmailListField, HiddenEnumField, HiddenFieldList,
                                      IndicoDateTimeField, IndicoEnumSelectField, IndicoMarkdownField,
                                      IndicoQuerySelectMultipleCheckboxField, IndicoQuerySelectMultipleField,
-                                     IndicoRadioField, PrincipalField, _LegacyPrincipalListField)
+                                     IndicoRadioField, PrincipalField)
+from indico.web.forms.fields.principals import PrincipalListField
 from indico.web.forms.util import inject_validators
 from indico.web.forms.validators import HiddenUnless, LinkedDateTime, SoftLength, UsedIf, WordCount
 from indico.web.forms.widgets import JinjaWidget, SwitchWidget
@@ -127,9 +128,9 @@ class AbstractSubmissionSettingsForm(IndicoForm):
                                                     enum=SubmissionRightsType, sorted=True,
                                                     description=_("Specify who will get contribution submission rights "
                                                                   "once an abstract has been accepted"))
-    authorized_submitters = _LegacyPrincipalListField(_("Authorized submitters"),
-                                                      description=_("These users may always submit abstracts, "
-                                                                    "even outside the regular submission period."))
+    authorized_submitters = PrincipalListField(_("Authorized submitters"),
+                                               description=_("These users may always submit abstracts, "
+                                                             "even outside the regular submission period."))
     submission_instructions = IndicoMarkdownField(_('Instructions'), editor=True,
                                                   description=_("These instructions will be displayed right before the "
                                                                 "submission form"))
--- a/indico/modules/events/abstracts/forms.py
+++ b/indico/modules/events/abstracts/forms.py
@@ -617,7 +617,8 @@ class InvitedAbstractMixin(object):
                                              choices=(('existing', _('Existing user')),
                                                       ('new', _('New user'))))
     submitter = PrincipalField(_('Submitter'),
-                               [HiddenUnless('users_with_no_account', 'existing'), DataRequired()], allow_external=True,
+                               [HiddenUnless('users_with_no_account', 'existing'), DataRequired()],
+                               allow_external_users=True,
                                description=_('The person invited to submit the abstract'))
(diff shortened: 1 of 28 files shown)


indico/modules/events/abstracts/forms.py
```diff
@@ -127,9 +127,9 @@ class AbstractSubmissionSettingsForm(IndicoForm):
                                                     enum=SubmissionRightsType, sorted=True,
                                                     description=_("Specify who will get contribution submission rights "
                                                                   "once an abstract has been accepted"))
-    authorized_submitters = PrincipalListField(_("Authorized submitters"),
-                                               description=_("These users may always submit abstracts, even outside "
-                                                             "the regular submission period"))
+    authorized_submitters = _LegacyPrincipalListField(_("Authorized submitters"),
```
ThiefMaster, line 130: This one can most likely use the new field

indico/modules/events/papers/forms.py
```diff
@@ -35,16 +36,16 @@ def make_competences_form(event):
...
-                                  description=_('List of users allowed to manage the call for papers'))
-    judges = PrincipalListField(_('Judges'),
-                                description=_('List of users allowed to judge papers'))
-    content_reviewers = PrincipalListField(_('Content reviewers'),
-                                           description=_('List of users allowed to review the content of the assigned '
-                                                         'papers'))
-    layout_reviewers = PrincipalListField(_('Layout reviewers'),
-                                          description=_('List of users allowed to review the layout of the assigned '
-                                                        'papers'))
+    managers = _LegacyPrincipalListField(_('Paper managers'), groups=True, event=lambda form: form.event,
```
ThiefMaster, line 39: Same here; all cases where we use the `PrincipalListField` directly should be fine with the new field.

indico/web/client/js/jquery/widgets/jinja/principal_list_widget.js
```diff
@@ -1,28 +1,49 @@
...
+        withRegistrationForms: false,
+        withNetworks: false,
         networks: [],
       },
-      options
+      ...options,
+    };
+    return options.legacy
+      ? _setupLegacyPrincipalListWidget(options)
+      : _setupPrincipalListWidget(options);
```
ThiefMaster, line 26: Is this return value used for anything? If not, let's just use a normal if/else.

indico/web/client/js/jquery/widgets/jinja/principal_list_widget.js
```diff
@@ -1,28 +1,49 @@
...
+  };
+
+  function _setupPrincipalListWidget(options) {
+    const field = document.getElementById(options.fieldId);
+    const principals = JSON.parse(field.value);
```
ThiefMaster, line 31: Should this really be here and not inside the component like it's done for the date-related fields?

indico/web/client/js/react/components/WTFPrincipalListField.jsx
```diff
@@ -0,0 +1,57 @@
...
+  const protectedField = useMemo(
+    () => protectedFieldId && document.getElementById(protectedFieldId),
+    [protectedFieldId]
+  );
+  const inputField = useMemo(() => document.getElementById(fieldId), [fieldId]);
+  const [value, setValue] = useState(defaultValue);
+  const [disabled, setDisabled] = useState(!!protectedField && !protectedField.checked);
+
+  useEffect(() => {
+    if (!protectedField) return;
```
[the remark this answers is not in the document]
plourenco: Seems like it isn't a rule. We could probably add [this](https://eslint.org/docs/rules/curly).

indico/web/forms/fields/principals.py
```diff
@@ -53,6 +53,52 @@ def serialize_principal(principal):
...
+                                 should be available.  Selecting such a user
+                                 will automatically create a pending user once
+                                 the form is submitted, even if other fields
+                                 in the form fail to validate!
+    """
+
+    widget = JinjaWidget('forms/principal_list_widget.html', single_kwargs=True)
+
+    def __init__(self, *args, **kwargs):
+        self.data = []
```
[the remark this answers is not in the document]
plourenco: It doesn't seem to be initialized by the Field class, I added it previously to silence a warning. 
Still, this doesn't make much sense because `_get_data` below checks if this is null. 
But now I'm wondering, perhaps we could add it?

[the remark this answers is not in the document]
ThiefMaster: isn't the warning just about it not being set in the constructur? ;)

indico/web/client/js/jquery/widgets/jinja/principal_list_widget.js
plourenco: Good question. Since this one doesn't really change and is only the initial value I thought we could avoid manipulating it through react. Having the manipulation in JS and the prop in react seems cleaner.

indico/web/forms/fields/principals.py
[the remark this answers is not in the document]
plourenco: Yes, it is just a warning. But perhaps we could avoid the null check in the `get_data` method below if we directly initialised this as zero. Unless something else depends on this initial value?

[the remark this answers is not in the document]
ThiefMaster: honestly, i would not try to refactor wtforms components too much as long as they work... hopefully we can eventually replace more and more with react+webargs/marshmallow

indico/modules/events/papers/forms.py
plourenco: Thanks.

indico/web/client/js/jquery/widgets/jinja/principal_list_widget.js
plourenco: Yes, it isn't.

indico/web/templates/forms/principal_list_widget.html
```diff
@@ -26,11 +26,15 @@
...
             fieldId: {{ field.id | tojson }},
             eventId: {{ (field.event.id if field.event else none) | tojson }},
-            openImmediately: {{ field.open_immediately | tojson }},
-            groups: {{ field.groups | tojson }},
-            allowExternal: {{ field.allow_external | tojson }},
-            allowNetworks: {{ field.allow_networks | tojson }},
-            networks: {{ field.ip_networks | sort(attribute='name') | tojson }}
+            openImmediately: {{ field.open_immediately | default(False) | tojson }},
+            withGroups: {{  (field.groups or field.allow_groups) | tojson }},
+            withExternalUsers: {{ (field.allow_external or field.allow_external_users) | tojson }},
```
plourenco, line 31: We should probably settle on a naming convention here. I was keeping `withX` on the templating and react side and `allowX` Python-wise. But it's starting to get confusing, can we perhaps keep one of them across all places?

indico/web/forms/fields/principals.py
```diff
@@ -53,74 +51,56 @@ def serialize_principal(principal):
...
-    :param allow_external: If "search users with no indico account"
-                           should be available.  Selecting such a user
-                           will automatically create a pending user once
-                           the form is submitted, even if other fields
-                           in the form fail to validate!
+                                        to an event should be selectable.
+    :param allow_event_roles: If event roles should be selectable.
+    :param allow_category_roles: If category roles should be selectable.
+    :param allow_external_users: If "search users with no indico account"
+                                 should be available.  Selecting such a user
```
Leats, line 60: One space too much before `Selecting`

plourenco: Thanks!

indico/web/templates/forms/principal_widget.html
```diff
@@ -16,8 +16,7 @@
             fieldId: {{ field.id | tojson }},
             eventId: {{ (field.event.id if field.event else none) | tojson }},
             required: {{ field.flags.required | tojson }},
-            allowExternal: {{ field.allow_external | tojson }}
+            withExternalUsers: {{ field.allow_external_users | default(False) | tojson }}
```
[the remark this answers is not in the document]
ThiefMaster: Actually, I think you can just remove the `| default(False)` since this attribute always exists now.

PR #4571, opened 2020-07-29
Limit visibility of pages to registered participants

### LINKED ISSUE #4507: Limit visibility of pages to registered participants

We have information (like the links to Zoom or Slack) which we would like to make available only to the registered participants of a conference. So far I was unable to figure out how to do this.

In the *menu* settings one can change the visibility in general. Here I would expect to be able to apply ACLs.

My current workaround is a hidden page I'm going to share via e-mail.

panagiotappl:
closes #4507 
image

### DIFF
--- a/indico/modules/events/layout/controllers/menu.py
+++ b/indico/modules/events/layout/controllers/menu.py
@@ -242,6 +242,12 @@ class RHPageDisplay(RHDisplayEventBase):
         }
     }
 
+    def _check_access(self):
+        RHDisplayEventBase._check_access(self)
+        if (self.page.menu_entry.registered_only and not self.event.is_user_registered(session.user)
+                and not self.event.can_manage(session.user)):
+            raise Forbidden
+
     def _process_args(self):
         RHDisplayEventBase._process_args(self)
         self.page = EventPage.get_or_404(request.view_args['page_id'])
--- a/indico/modules/events/layout/forms.py
+++ b/indico/modules/events/layout/forms.py
@@ -161,6 +161,8 @@ class MenuUserEntryFormBase(IndicoForm):
     title = StringField(_("Title"), [DataRequired()])
     is_enabled = BooleanField(_("Show"), widget=SwitchWidget())
     new_tab = BooleanField(_("Open in a new tab"), widget=SwitchWidget())
+    registered_only = BooleanField(_("Restricted"), widget=SwitchWidget(),
+                                   description=_("Visible to registered users only."))
 
 
 class MenuLinkForm(MenuUserEntryFormBase):
--- a/indico/modules/events/layout/models/menu.py
+++ b/indico/modules/events/layout/models/menu.py
@@ -7,7 +7,7 @@
 
 from __future__ import unicode_literals
 
-from flask import g
+from flask import g, session
 from sqlalchemy.orm import joinedload
 from werkzeug.utils import cached_property
 
@@ -125,12 +125,17 @@ class MenuEntryMixin(object):
     def is_visible(self):
         if not self.is_enabled:
             return False
+        if (self.registered_only and not self.event_ref.is_user_registered(session.user)
+                and not self.event_ref.can_manage(session.user)):
+            return False
         if not self.name:
             return True
         if self.is_orphaned:
             return False
 
         data = self.default_data
+        if data.hide_if_restricted and not self.event_ref.can_access(session.user):
+            return False
         if not data.static_site and g.get('static_site'):
             return False
         return data.visible(self.event_ref)
@@ -177,6 +182,7 @@ class TransientMenuEntry(MenuEntryMixin):
         self.link_url = None
         self.page_id = None
         self.is_root = True
+        self.registered_only = False
         for child in self.children:
             child.is_root = False
 
@@ -263,6 +269,12 @@ class MenuEntry(MenuEntryMixin, db.Model):
     )
     #: Whether the menu entry should be opened in a new tab or window
     new_tab = db.Column(
+        db.Boolean,
+        nullable=False,
+        default=False
+    )
+    #: Whether the menu entry should be viewable only by registered users
+    registered_only = db.Column(
         db.Boolean,
         nullable=False,
         default=False
--- a/indico/modules/events/layout/models/menu.py
+++ b/indico/modules/events/layout/models/menu.py
@@ -129,7 +129,9 @@ class MenuEntryMixin(object):
                 and not self.event_ref.can_manage(session.user)):
             return False
         if not self.name:
-            return True
+            # we don't have `hide_if_restricted` for custom menu items, so we
+            # always hide them if the user cannot access the event
+            return self.event_ref.can_access(session.user)
         if self.is_orphaned:
             return False
 
--- a/indico/migrations/versions/20200729_1842_d89abffb0f63_add_menu_entries_registered_only.py
+++ b/indico/migrations/versions/20200729_1842_d89abffb0f63_add_menu_entries_registered_only.py
@@ -1,0 +1,26 @@
+"""Add menu_entries.registered_only column.
+
+Revision ID: d89abffb0f63
+Revises: 4f0b4dd412b5
+Create Date: 2020-07-28 18:42:54.968880
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = 'd89abffb0f63'
+down_revision = '4f0b4dd412b5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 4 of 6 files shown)


indico/migrations/versions/20200728_1842_d89abffb0f63_add_registered_only.py
```diff
@@ -0,0 +1,30 @@
+"""Add registered_only column for showing menu entries only to registered users.
```
ThiefMaster, line 1: I'd go for something more concise here, e.g. "Add menu_entries.registered_only column"
and `20200728_1842_d89abffb0f63_add_menu_entries_registered_only.py` for the filename

indico/migrations/versions/20200728_1842_d89abffb0f63_add_registered_only.py
```diff
@@ -0,0 +1,30 @@
...
+    op.add_column('menu_entries', sa.Column('registered_only', sa.Boolean(), nullable=True), schema='events')
+    op.execute('UPDATE events.menu_entries SET registered_only = false')
+    op.alter_column('menu_entries', 'registered_only', nullable=False, schema='events')
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('menu_entries', 'registered_only', schema='events')
+    # ### end Alembic commands ###
```
ThiefMaster, line 30:
```suggestion
def upgrade():
    op.add_column('menu_entries', sa.Column('registered_only', sa.Boolean(), nullable=False, server_default='false'),
                  schema='events')
    op.alter_column('menu_entries', 'registered_only', server_default=None, schema='events')


def downgrade():
    op.drop_column('menu_entries', 'registered_only', schema='events')
```

indico/modules/events/layout/controllers/menu.py
```diff
@@ -247,4 +248,8 @@ def _process_args(self):
         self.page = EventPage.get_or_404(request.view_args['page_id'])
 
     def _process(self):
+        registered_only = self.page.menu_entry.registered_only
+        registered_users = [x.user for x in get_registered_event_persons(self.event)]
+        if registered_only and session.user not in registered_users and not self.event.can_manage(session.user):
+            raise Forbidden
```
ThiefMaster, line 254: This is an access check so it should be in `_check_access`.

`get_registered_event_persons` is not the best choice here - you do not really care about EventPersons here but just if the user is registered. Querying all registered users is also expensive, you only need to know about the current user here.

I'd suggest adding this method to `User` to a cheap check:

```python
def is_registered_in_event(self, event):
    """Check whether the user is registered in the event.
    
    This takes both unpaid and complete registrations into account.
    """
    from indico.modules.events.registration.models.registrations import Registration, RegistrationState
    return (Registration.query.with_parent(self)
            .filter(Registration.event == event,
                    Registration.state.in_([RegistrationState.unpaid, RegistrationState.complete]),
                    ~Registration.is_deleted,
                    ~RegistrationForm.is_deleted)
            .has_rows())
```

indico/modules/events/layout/forms.py
```diff
@@ -161,6 +161,8 @@ class MenuUserEntryFormBase(IndicoForm):
     title = StringField(_("Title"), [DataRequired()])
     is_enabled = BooleanField(_("Show"), widget=SwitchWidget())
     new_tab = BooleanField(_("Open in a new tab"), widget=SwitchWidget())
+    registered_only = BooleanField(_("Restricted"), description=_("Visible to registered users only."),
```
ThiefMaster, line 164: we usually put the `description` kwarg last in wtforms

indico/modules/events/views.py
```diff
@@ -227,6 +228,8 @@ def __init__(self, rh, event_, **kwargs):
         self.event = event_
         kwargs['conf_layout_params'] = self._get_layout_params()
         kwargs.setdefault('page_title', self.sidemenu_title)
+        registered_users = [x.user for x in get_registered_event_persons(self.event)]
+        kwargs.setdefault('registered_users', registered_users)
```
ThiefMaster, line 232: Shouldn't be here, this is called **a lot** and will result in poor performance for large events.

indico/modules/events/templates/display/conference/base.html
```diff
@@ -58,7 +58,9 @@ <h1>
             <div class="conf_leftMenu">
                 <ul id="outer">
                     {%- for entry in conf_layout_params.menu if entry.is_visible %}
-                        {{ menu_entry_display(entry, active_entry_id=conf_layout_params.active_menu_item) }}
+                        {% if not entry.registered_only or session.user in registered_users %}
+                            {{ menu_entry_display(entry, active_entry_id=conf_layout_params.active_menu_item) }}
+                        {% endif %}
```
ThiefMaster, line 63: Better check this in the `MenuEntryData.visible` property.

indico/modules/events/layout/forms.py
ThiefMaster: I would only show this option for menu items corresponding to pages. Because for most other menu items, their visibility do not imply access restrictions, so being able to hide e.g. "Timetable" for unregistered users would be misleading, since you can still access that by URL.

panagiotappl: Isn't it only showing for the custom items that way, just like the "open in a new tab" option?

ThiefMaster: You're right, it's only used for pages and links. And for external links, it's actually nice to be able to restrict it as well (if someone wants to put e.g. a zoom url there).

indico/modules/events/layout/controllers/menu.py
```diff
@@ -242,6 +242,12 @@ class RHPageDisplay(RHDisplayEventBase):
         }
     }
 
+    def _check_access(self):
+        if (self.page.menu_entry.registered_only and not session.user.is_registered_in_event(self.event)
+                and not self.event.can_manage(session.user)):
+            raise Forbidden
+        RHDisplayEventBase._check_access(self)
```
ThiefMaster, line 249: I would keep this check first, and only apply the additional checks afterwards.

indico/modules/events/layout/controllers/menu.py
```diff
@@ -242,6 +242,12 @@ class RHPageDisplay(RHDisplayEventBase):
         }
     }
 
+    def _check_access(self):
+        if (self.page.menu_entry.registered_only and not session.user.is_registered_in_event(self.event)
```
ThiefMaster, line 246: I think this will fail if the user is not logged in

indico/modules/events/layout/models/menu.py
```diff
@@ -125,6 +125,8 @@ def is_orphaned(self):
     def is_visible(self):
         if not self.is_enabled:
             return False
+        if self.registered_only and not session.user.is_registered_in_event(self.event):
```
ThiefMaster, line 128: needs to check for the `not session.user` case since it may be None.

Alternatively (maybe that keeps the code saner here), you could replace `User.is_registered_in_event` with `Event.is_user_registered`:

```python
def is_user_registered(self, user):
    """Check whether the user is registered in the event.
    
    This takes both unpaid and complete registrations into account.
    """
    from indico.modules.events.registration.models.registrations import Registration, RegistrationState
    if user is None:
        return False
    return (Registration.query.with_parent(self)
            .filter(Registration.user == user,
                    Registration.state.in_([RegistrationState.unpaid, RegistrationState.complete]),
                    ~Registration.is_deleted,
                    ~RegistrationForm.is_deleted)
            .has_rows())
```

indico/modules/events/layout/models/menu.py
```diff
@@ -125,6 +125,8 @@ def is_orphaned(self):
     def is_visible(self):
         if not self.is_enabled:
             return False
+        if self.registered_only and not self.event.is_user_registered(session.user):
+            return False
```
ThiefMaster, line 129: We now have access checks both in here and in `MenuEntryData.visible`. Can you either move your check in there or the one from there in here:

```
        if self.hide_if_restricted and not event.can_access(session.user):
            return False
```

If you keep the checks here IIRC you are supposed to use `self.event_ref` instead of `self.event`.

indico/migrations/versions/20200728_1842_d89abffb0f63_add_menu_entries_registered_only.py
```diff
@@ -0,0 +1,26 @@
+"""Add menu_entries.registered_only column.
+
+Revision ID: d89abffb0f63
+Revises: 05f227f4b938
```
ThiefMaster, line 4: Could you update the alembic revision so it revises `4f0b4dd412b5`? This is a small fix I added earlier today. You'll also need to rename the file to update the date to today so it shows up at the end of the file list in the versions folder.

PR #4572, opened 2020-07-29
Connect the revision endpoints to the service

### LINKED ISSUE #4563: Inform editing service about reviews
Call during the `review_editable_revision` operation, so the microservice can decide whether to publish or not (in case of accepting). We could also let the microservice modify the tags being assigned to the revision. It may be useful to let the microservice reject the review altogether (in which case it should fail gracefully for the editor).

plourenco:
Closes #4563

Please use the following server version: indico/openreferee#5

### DIFF
--- a/indico/modules/events/editing/controllers/backend/timeline.py
+++ b/indico/modules/events/editing/controllers/backend/timeline.py
@@ -163,19 +163,19 @@ class RHReviewEditable(RHContributionEditableRevisionBase):
         args = parser.parse(argmap)
         service_url = editing_settings.get(self.event, 'service_url')
 
-        revision = review_editable_revision(self.revision, session.user, action, comment, args['tags'],
-                                            args.get('files'))
+        new_revision = review_editable_revision(self.revision, session.user, action, comment, args['tags'],
+                                                args.get('files'))
 
         publish = True
         if service_url:
             try:
-                resp = service_handle_review_editable(self.editable, action, self.revision, revision)
+                resp = service_handle_review_editable(self.editable, action, self.revision, new_revision)
                 publish = resp.get('publish', True)
             except ServiceRequestFailed:
-                raise ServiceUnavailable(_('Revision failed, please try again later.'))
+                raise ServiceUnavailable(_('Failed processing review, please try again later.'))
 
         if publish and action in (EditingReviewAction.accept, EditingReviewAction.update_accept):
-            publish_editable_revision(revision or self.revision)
+            publish_editable_revision(new_revision or self.revision)
         return '', 204
 
 
@@ -199,7 +199,7 @@ class RHConfirmEditableChanges(RHContributionEditableRevisionBase):
                 resp = service_handle_review_editable(self.editable, action, self.revision)
                 publish = resp.get('publish', True)
             except ServiceRequestFailed:
-                raise ServiceUnavailable(_('Revision failed, please try again later.'))
+                raise ServiceUnavailable(_('Failed processing review, please try again later.'))
 
         if publish and action == EditingConfirmationAction.accept:
             publish_editable_revision(self.revision)
--- a/indico/modules/events/editing/controllers/backend/timeline.py
+++ b/indico/modules/events/editing/controllers/backend/timeline.py
@@ -17,7 +17,6 @@ from marshmallow_enum import EnumField
 from sqlalchemy.orm import joinedload
 from werkzeug.exceptions import Forbidden, NotFound, ServiceUnavailable
 
-from indico.core.db import db
 from indico.core.errors import UserValueError
 from indico.modules.events.editing.controllers.base import RHContributionEditableBase, TokenAccessMixin
 from indico.modules.events.editing.fields import EditingFilesField, EditingTagsField
@@ -26,12 +25,13 @@ from indico.modules.events.editing.models.revision_files import EditingRevisionF
 from indico.modules.events.editing.models.revisions import EditingRevision, InitialRevisionState
 from indico.modules.events.editing.operations import (assign_editor, confirm_editable_changes, create_new_editable,
                                                       create_revision_comment, create_submitter_revision,
-                                                      delete_editable, delete_revision_comment, replace_revision,
-                                                      review_editable_revision, unassign_editor, undo_review,
-                                                      update_revision_comment)
+                                                      delete_revision_comment, publish_editable_revision,
+                                                      replace_revision, review_editable_revision, unassign_editor,
+                                                      undo_review, update_revision_comment)
 from indico.modules.events.editing.schemas import (EditableSchema, EditingConfirmationAction, EditingReviewAction,
                                                    ReviewEditableArgs)
-from indico.modules.events.editing.service import ServiceRequestFailed, service_handle_new_editable
+from indico.modules.events.editing.service import (ServiceRequestFailed, service_handle_new_editable,
+                                                   service_handle_review_editable)
 from indico.modules.events.editing.settings import editing_settings
 from indico.modules.files.controllers import UploadFileMixin
 from indico.modules.users import User
@@ -81,6 +81,7 @@ class RHContributionEditableRevisionBase(RHContributionEditableBase):
                          .with_parent(self.editable, 'revisions')
                          .filter_by(id=request.view_args['revision_id'])
                          .first_or_404())
+
         if self.revision is None:
             raise NotFound
 
@@ -138,14 +139,12 @@ class RHCreateEditable(RHContributionEditableBase):
         initial_state = InitialRevisionState.new if service_url else InitialRevisionState.ready_for_review
 
         editable = create_new_editable(self.contrib, self.editable_type, session.user, args['files'], initial_state)
-        db.session.commit()
         if service_url:
             try:
                 service_handle_new_editable(editable)
             except ServiceRequestFailed:
-                delete_editable(editable)
-                db.session.commit()
                 raise ServiceUnavailable(_('Submission failed, please try again later.'))
+
         return '', 201
 
 
@@ -162,7 +161,21 @@ class RHReviewEditable(RHContributionEditableRevisionBase):
             argmap['files'] = EditingFilesField(self.event, self.contrib, self.editable_type, allow_claimed_files=True,
                                                 required=True)
         args = parser.parse(argmap)
-        review_editable_revision(self.revision, session.user, action, comment, args['tags'], args.get('files'))
+        service_url = editing_settings.get(self.event, 'service_url')
+
+        revision = review_editable_revision(self.revision, session.user, action, comment, args['tags'],
+                                            args.get('files'))
+
+        publish = True
+        if service_url:
+            try:
+                resp = service_handle_review_editable(self.editable, action, self.revision, revision)
+                publish = resp.get('publish', True)
+            except ServiceRequestFailed:
+                raise ServiceUnavailable(_('Revision failed, please try again later.'))
+
+        if publish and action in (EditingReviewAction.accept, EditingReviewAction.update_accept):
+            publish_editable_revision(revision or self.revision)
         return '', 204
 
 
@@ -178,6 +191,18 @@ class RHConfirmEditableChanges(RHContributionEditableRevisionBase):
     })
     def _process(self, action, comment):
         confirm_editable_changes(self.revision, session.user, action, comment)
+
+        service_url = editing_settings.get(self.event, 'service_url')
+        publish = True
+        if service_url:
+            try:
+                resp = service_handle_review_editable(self.editable, action, self.revision)
+                publish = resp.get('publish', True)
+            except ServiceRequestFailed:
+                raise ServiceUnavailable(_('Revision failed, please try again later.'))
+
+        if publish and action == EditingConfirmationAction.accept:
+            publish_editable_revision(self.revision)
         return '', 204
 
 
@@ -236,6 +261,8 @@ class RHUndoReview(RHContributionEditableRevisionBase):
 class RHCreateRevisionComment(RHContributionEditableRevisionBase):
     """Create new revision comment"""
 
+    SERVICE_ALLOWED = True
+
     def _check_revision_access(self):
         return self.editable.can_comment(session.user)
 
@@ -244,9 +271,13 @@ class RHCreateRevisionComment(RHContributionEditableRevisionBase):
         'internal': fields.Bool(missing=False)
     })
     def _process(self, text, internal):
-        if internal and not self.editable.can_use_internal_comments(session.user):
+        user = session.user
+        if self.is_service_call:
+            user = User.get_system_user()
+        elif internal and not self.editable.can_use_internal_comments(session.user):
             internal = False
-        create_revision_comment(self.revision, session.user, text, internal)
+
+        create_revision_comment(self.revision, user, text, internal)
         return '', 201
 
 
--- a/indico/modules/events/editing/controllers/backend/timeline.py
+++ b/indico/modules/events/editing/controllers/backend/timeline.py
@@ -243,7 +243,15 @@ class RHCreateSubmitterRevision(RHContributionEditableRevisionBase):
                                        required=True)
         })
 
-        create_submitter_revision(self.revision, session.user, args['files'])
+        service_url = editing_settings.get(self.event, 'service_url')
+        new_revision = create_submitter_revision(self.revision, session.user, args['files'])
+
+        if service_url:
+            try:
+                service_handle_review_editable(self.editable, EditingReviewAction.update,
+                                               self.revision, new_revision)
+            except ServiceRequestFailed:
+                raise ServiceUnavailable(_('Failed processing review, please try again later.'))
         return '', 204
 
 
--- a/indico/modules/events/editing/schemas.py
+++ b/indico/modules/events/editing/schemas.py
@@ -382,8 +382,13 @@ class EditableTypePrincipalsSchema(mm.Schema):
     principals = PrincipalList(many=True, allow_event_roles=True, allow_category_roles=True)
 
 
+class ReviewCommentSchema(mm.Schema):
+    text = fields.String(required=True)
+    internal = fields.Boolean(default=False)
+
+
 class ServiceReviewEditableSchema(mm.Schema):
     publish = fields.Boolean(default=True)
     comment = fields.String()
-    comments = fields.List(fields.String())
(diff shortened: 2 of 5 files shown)


indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -162,7 +165,25 @@ def _process(self, action, comment):
             argmap['files'] = EditingFilesField(self.event, self.contrib, self.editable_type, allow_claimed_files=True,
                                                 required=True)
         args = parser.parse(argmap)
-        review_editable_revision(self.revision, session.user, action, comment, args['tags'], args.get('files'))
+        service_url = editing_settings.get(self.event, 'service_url')
+
+        revision = review_editable_revision(self.revision, session.user, action, comment, args['tags'],
+                                            args.get('files'))
+        publish = True
+        db.session.commit()
```
ThiefMaster, line 173: I'm starting to dislike the fact that we need to commit because there may be new files uploaded during the review. This is a mess, but i guess we cannot avoid this.

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -162,7 +165,25 @@ def _process(self, action, comment):
...
+        if service_url:
+            try:
+                resp = service_handle_review_editable(self.editable, revision or self.revision)
+                publish = resp.get('publish', True)
+            except ServiceRequestFailed:
+                self.revision.final_state = FinalRevisionState.none
+                if revision:
+                    delete_revision(revision)
+                db.session.commit()
+                raise ServiceUnavailable(_('Revision failed, please try again later.'))
```
ThiefMaster, line 183: I guess since we raise an exception here, the email queued for sending in `notify_editor_judgment` won't be sent?

indico/modules/events/editing/service.py
```diff
@@ -148,3 +150,25 @@ def service_handle_new_editable(editable):
     except requests.RequestException as exc:
         logger.exception('Failed calling listener for editable')
         raise ServiceRequestFailed(exc)
+
+
+def service_handle_review_editable(editable, revision):
+
+    class PublishSchema(mm.Schema):
+        publish = fields.Boolean()
+        success = fields.Boolean()  # TODO: We can probably remove this, or find a simpler alternative
```
ThiefMaster, line 159: I don't think we really need a schema for this, do we? Just using the dict coming from the json response is probably more than enough...

indico/modules/events/editing/service.py
```diff
@@ -148,3 +150,25 @@ def service_handle_new_editable(editable):
     except requests.RequestException as exc:
         logger.exception('Failed calling listener for editable')
         raise ServiceRequestFailed(exc)
+
+
+def service_handle_review_editable(editable, revision):
```
ThiefMaster, line 155: Shouldn't we also send the type of review action that was performed to the service?

indico/modules/events/editing/controllers/backend/timeline.py
plourenco: I agree. I was wondering that maybe we could discuss an alternative to this. Not sure how well it would work, but the idea would be to make this a pre-listener rather than a post-listener (before changes are committed). In sum, we would send the object to the service and only commit it after the service confirmation.

This comes with a couple of pitfalls like trying to modify an object that doesn't exist yet, but ideally, we could also receive everything as a response from the service (instead of doing new API calls from the service). Can't say if this is a better alternative because it is also less flexible, but it ensures nothing is committed in-between.

indico/modules/events/editing/controllers/backend/timeline.py
plourenco: Great question. I had a note here to do the `notify_editor_judgment` within the RH because I thought notifications were probably being sent. But even better now that I found out about the queue, looks like raising an exception here doesn't even let it get to the `flush_emails` part!

indico/modules/events/editing/service.py
plourenco: I agree, I was debating about the same thing.

indico/modules/events/editing/service.py
plourenco: Maybe, but wouldn't that be also available in the `final_state`? It is set right before going into the service.

ThiefMaster: I think having to check the final state to guess what action was performed isn't particularly developer-friendly ;)

indico/modules/events/editing/controllers/backend/timeline.py
ThiefMaster: The main reason why we commit first is so the microservice can retrieve files, right?

Could we maybe send URLs with a token that allows downloading the revision's files based on their UUID?
Basically, we'd add an endpoint like this (next to `RHFileInfo`):

```python
# /files/<uuid:uuid>/download?token=<token>
class RHFileDownload(RHFileBase):
    def _check_access(self):
        # this raises if the token is invalid/expired
        uuid = secure_serializer.loads(request.view_args['token'], salt='file-download', max_age=86400)
        if uuid != self.file.uuid:
            raise Forbidden
    
    def _process(self):
        return self.file.send()
```

and then build an url pointing to that endpoint and send this as the `external_url` when serializing the revision's files for the microservice.

For basic operations the microservice should perform immediately, we'd use returned data from the request to the microservice, and for anything else the microservice would be expected to perform then asynchronously to happen after the commit (we could add an endpoint to check if a given revision already exists, or the microservice could simply retry if necessary).

plourenco: Exactly, this would be a much cleaner alternative. There are other use-cases where we might need to do something immediately, such as adding a comment, maybe even tags? But we can always make that the response of the service and perform them immediately on the first commit. Allowing the service to modify any of the attributes of the object we're about to commit.

ThiefMaster: I would say anything that must be done immediately or affect the action being performed itself (e.g. modify the set of tags during a review), should be done by returning data from the microservice. And anything beyond that should be done later, asynchronously.

ThiefMaster: I think the only place where it certainly makes sense to keep the intermediate commit is when *enabling* the microservice initially.

indico/modules/events/editing/service.py
```diff
@@ -148,3 +154,48 @@ def service_handle_new_editable(editable):
...
+        publish = fields.Boolean(default=True)
+        comment = fields.String()
+        comments = fields.List(fields.String())
+        tags = fields.List(fields.String())
+
+    # TODO: consider not sending the endpoints
+    new_revision = revision or parent_revision
+    data = {
+        'revision': EditingRevisionUnclaimedSchema().dump(new_revision),
+        'endpoints': {
```
plourenco, line 171: Passing the endpoints as part of the request starts to get excessive when multiple service endpoints might need them all together. I'm starting to considering these should be extraneous to the type of request but, instead, be built by the micro-service. IMO, we should document these endpoints and let the service build the ones it intends to use.

plourenco: We are no longer considering this.

indico/modules/files/schemas.py
```diff
@@ -14,7 +14,7 @@
 class FileSchema(mm.ModelSchema):
     class Meta:
         model = File
-        fields = ('claimed', 'content_type', 'created_dt', 'filename', 'size', 'uuid')
+        fields = ('claimed', 'content_type', 'created_dt', 'filename', 'size', 'uuid', 'external_download_url')
```
ThiefMaster, line 17: Including the signed URL here is dangerous!

Using `RHFileInfo` you can get the metadata of any file (using this schema) based on the UUID. And if this includes a signed download link it means as soon as you have access to any endpoint where you can upload a File, you could use Indico as a 24-hour one-click-hoster.

I would subclass `FileSchema` as `FileDownloadSchema` and add the signed download url only in there.

plourenco: Great point! TBH I can't seem to find any further needs for `external_download_url` considering it is only being used in `RHFileInfo` and `UploadFileMixin`. The attribute we need in the service is being exposed directly in `EditingRevisionUnclaimedFileSchema`, so I'm just removing it.

ThiefMaster: I would keep the property, but maybe rename it to `signed_download_url`? Just not in a schema if you don't need it there anyway

indico/modules/events/editing/service.py
```diff
@@ -148,3 +134,64 @@ def service_handle_new_editable(editable):
...
+            editable.contribution_id,
+            new_revision.id
+        )
+        resp = requests.post(_build_url(editable.event, path), headers=_get_headers(editable.event),
+                             json=data)
+        resp.raise_for_status()
+        try:
+            resp = ServiceReviewEditableSchema().dump(resp.json())
+        except ValueError:
+            return dict()
```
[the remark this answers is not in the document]
plourenco: Always returning a JSON-parsable response is easier to handle than having empties or nulls in between. I added this after an use-case with our demo service where an empty body wasn't expected.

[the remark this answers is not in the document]
ThiefMaster: `{}` and `dict()` are literally the same ;)

but my question was when we would get a ValueError there in the first place

indico/modules/events/editing/service.py
```diff
@@ -148,3 +134,64 @@ def service_handle_new_editable(editable):
...
+
+        if 'comment' in resp:
+            parent_revision.comment = resp.get('comment')
```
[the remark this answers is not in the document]
plourenco: But wouldn't that override the previous comment with an empty string if the service does not provide one?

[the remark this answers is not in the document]
ThiefMaster: indeed, you're right

indico/modules/events/editing/service.py
```diff
@@ -148,3 +134,64 @@ def service_handle_new_editable(editable):
...
+            _get_event_identifier(editable.event),
+            editable.type.name,
+            editable.contribution_id,
+            new_revision.id
+        )
+        resp = requests.post(_build_url(editable.event, path), headers=_get_headers(editable.event),
+                             json=data)
+        resp.raise_for_status()
+        try:
+            resp = ServiceReviewEditableSchema().dump(resp.json())
```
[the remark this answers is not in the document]
plourenco: Yes, nice catch, clumsily it was working too.

indico/modules/files/models/files.py
```diff
@@ -73,6 +75,11 @@ def save(self, context, data):
         super(File, self).save(data)
         del self.__context
 
+    @property
+    def external_download_url(self):
```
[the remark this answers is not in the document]
plourenco: Less confusing, I'm updating the service as well.

indico/modules/events/editing/service.py
```diff
@@ -148,3 +134,64 @@ def service_handle_new_editable(editable):
     except requests.RequestException as exc:
         logger.exception('Failed calling listener for editable')
         raise ServiceRequestFailed(exc)
+
+
+def service_handle_review_editable(editable, action, parent_revision, revision=None):
+    new_revision = revision or parent_revision
+    data = {
+        'action': action.value,
```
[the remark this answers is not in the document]
plourenco: Sounds good

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -162,7 +161,21 @@ def _process(self, action, comment):
             argmap['files'] = EditingFilesField(self.event, self.contrib, self.editable_type, allow_claimed_files=True,
                                                 required=True)
         args = parser.parse(argmap)
-        review_editable_revision(self.revision, session.user, action, comment, args['tags'], args.get('files'))
+        service_url = editing_settings.get(self.event, 'service_url')
+
+        revision = review_editable_revision(self.revision, session.user, action, comment, args['tags'],
```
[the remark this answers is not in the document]
plourenco: I agree

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -162,7 +161,21 @@ def _process(self, action, comment):
...
+                                            args.get('files'))
+
+        publish = True
+        if service_url:
+            try:
+                resp = service_handle_review_editable(self.editable, action, self.revision, revision)
+                publish = resp.get('publish', True)
+            except ServiceRequestFailed:
+                raise ServiceUnavailable(_('Revision failed, please try again later.'))
```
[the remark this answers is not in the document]
plourenco: Sounds clearer

indico/modules/events/editing/service.py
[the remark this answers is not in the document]
plourenco: Yes, I've changed it too `{}`, I meant the return type in general.

The `ValueError` is what I expect if the service response doesn't conform with the schema. But after checking, I guess that is actually `ValidationError`, right?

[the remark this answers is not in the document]
ThiefMaster: not sure... but if we get crap back from the service I'd prefer to fail with an error instead of catching it and being silent about it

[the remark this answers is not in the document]
plourenco: All right, I can raise it. But should it prevent the revision from happening?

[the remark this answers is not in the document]
ThiefMaster: I would say so - after all, if you didn't get something decent back from the service call, it's likely that something went wrong there

ThiefMaster:
Since we needed the signed download link in another PR it's cherry-picked into master now. You might have a much easier time rebasing if you squash your commits first ;)

indico/modules/events/editing/service.py
```diff
@@ -148,3 +135,61 @@ def service_handle_new_editable(editable):
...
+        path = '/event/{}/editable/{}/{}/{}'.format(
+            _get_event_identifier(editable.event),
+            editable.type.name,
+            editable.contribution_id,
+            new_revision.id
+        )
+        resp = requests.post(_build_url(editable.event, path), headers=_get_headers(editable.event),
+                             json=data)
+        resp.raise_for_status()
+        resp = ServiceReviewEditableSchema().load(resp.json())
```
ThiefMaster, line 157: This schema also contains a `comments` field, but this is not being handled anywhere. Also see my related TODO comment in `process_revision` in indico/openreferee#5

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -162,7 +161,21 @@ def _process(self, action, comment):
...
+        if service_url:
+            try:
+                resp = service_handle_review_editable(self.editable, action, self.revision, new_revision)
+                publish = resp.get('publish', True)
+            except ServiceRequestFailed:
+                raise ServiceUnavailable(_('Failed processing review, please try again later.'))
+
+        if publish and action in (EditingReviewAction.accept, EditingReviewAction.update_accept):
+            publish_editable_revision(new_revision or self.revision)
         return '', 204
```
ThiefMaster, line 180: When the submitter creates a new revision after "request changes", we currently do not notify the service. `RHCreateSubmitterRevision` needs some service-reloated logic as well.

indico/modules/events/editing/schemas.py
```diff
@@ -363,3 +380,10 @@ class Meta:
         rh_context = ('event',)
 
     principals = PrincipalList(many=True, allow_event_roles=True, allow_category_roles=True)
+
+
+class ServiceReviewEditableSchema(mm.Schema):
+    publish = fields.Boolean(default=True)
+    comment = fields.String()
+    comments = fields.List(fields.String())
```
ThiefMaster, line 388: See my other comment:

- this is currently not taken into account
- we should probably accept the same data structure here like the one we'd POST to the comments endpoint

ThiefMaster (changes requested):

indico/modules/events/editing/service.py
plourenco: I had some changes stashed for this. But ended up concluding we don't actually need it, after all, using the endpoint works!

ThiefMaster: not if the revision does not exist yet...

indico/modules/events/editing/controllers/backend/timeline.py
plourenco: Thought about the same, but had the impression we didn't actually need it because it wasn't considered as a "revision". Nonetheless, sounds good to me.

indico/modules/events/editing/service.py
```diff
@@ -148,3 +137,65 @@ def service_handle_new_editable(editable):
...
+
+        if 'comment' in resp:
+            parent_revision.comment = resp.get('comment')
+        if 'tags' in resp:
+            parent_revision.tags = {tag for tag in editable.event.editing_tags
+                                    if tag.id in map(int, resp.get('tags'))}
+        if 'comments' in resp:
+            for comment in resp.get('comments'):
+                create_revision_comment(new_revision, User.get_system_user(), comment.get('text'),
+                                        internal=comment.get('internal'))
```
ThiefMaster, line 169: only use `.get()` if you expect something to not exist. but if fields in the schema are required or have a default (`missing=...`) value those keys always exist

```suggestion
        for comment in resp.get('comments', []):
            create_revision_comment(new_revision, User.get_system_user(), comment['text'],
                                    internal=comment['internal'])
```

indico/modules/events/editing/schemas.py
```diff
@@ -363,3 +374,15 @@ class Meta:
         rh_context = ('event',)
 
     principals = PrincipalList(many=True, allow_event_roles=True, allow_category_roles=True)
+
+
+class ReviewCommentSchema(mm.Schema):
+    text = fields.String(required=True)
+    internal = fields.Boolean(default=False)
```
ThiefMaster, line 381: I think it's `missing` and not `default`

indico/modules/events/editing/schemas.py
```diff
@@ -363,3 +374,15 @@ class Meta:
...
+
+
+class ServiceReviewEditableSchema(mm.Schema):
+    publish = fields.Boolean(default=True)
```
ThiefMaster, line 385: ditto

indico/modules/events/editing/service.py
plourenco: For some reason, with dictionaries, I find it more idiomatic than pure parentheses, must be from Java.
But good catch, didn't expect comments to be undefined at this point.

ThiefMaster (approved):

PR #4577, opened 2020-07-31
Add ability to upload custom PDF for BOA

### LINKED ISSUE #3039: Editable book of abstracts

Leats:
Closes #3039.


indico/modules/events/abstracts/controllers/management.py
```diff
@@ -51,8 +52,12 @@ def _process(self):
             return WPManageAbstracts.render_template('management/disabled.html', self.event)
         else:
             abstracts_count = Abstract.query.with_parent(self.event).count()
+            custom_boa = None
+            if self.event.has_custom_boa:
+                custom_boa = BasicFileSchema().dump(self.event.custom_boa)
```
plourenco, line 57: Would this work the same using the previous schema but with [see below]?
```
FileSchema(only=('filename', 'size', 'uuid')).dump(self.event.custom_boa)
```

indico/modules/events/abstracts/client/js/boa.jsx
```diff
@@ -0,0 +1,160 @@
...
+  };
+
+  const handleSubmit = async ({file}) => {
+    try {
+      await indicoAxios.post(customBOAURL({confId: eventId}), {file});
+    } catch (e) {
+      return handleSubmitError(e);
+    }
+    location.reload();
+    // never finish submitting to avoid fields being re-enabled
```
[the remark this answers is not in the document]
ThiefMaster: This would close the modal - while that would work, we thought it's nicer to reload without showing a "non submitting" state before that temporarily.

indico/modules/events/abstracts/controllers/management.py
ThiefMaster: Yep, that would work. But since this is something we likely need in more places in the future (whenever we use this field), a custom schema for it felt cleaner.

PR #4585, opened 2020-08-05
Allow registrants to withdraw their registration

OmeGak:
Closes: #2715
Supersedes and thus closes: #4166

Replying to @ThiefMaster's [comment on the superseded PR](https://github.com/indico/indico/pull/4166#discussion_r348124148):
> Should we check for the modification period here at all? I think withdrawal should always be possible, especially if no payment is involved. After paying it does make sense that you need to contact the organizer though (regardless of the modification setting) - and they should then have a button in the management view to mark the registration as withdrawn.

The logic I have applied to allow withdrawals:
- The registration has not been paid.
- The event the registration is associated to has not finished yet.

Regarding allowing managers to undo withdrawals, this is now possible via the "Reset registration" button in the registration management view.

## Screenshots

||
|-|
|image|
|image|

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -106,6 +106,8 @@ Improvements
   :pr:`4577`)
 - Display each news item on a separate page instead of together with all the
   other news items (:pr:`4587`)
+- Allow registrants to withdraw their application (:issue:`2715`, :pr:`4166`,
+  thanks :user:`brabemi` & :user:`omegak`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/events/registration/models/registrations.py
+++ b/indico/modules/events/registration/models/registrations.py
@@ -525,6 +525,23 @@ class Registration(db.Model):
         if self.state != initial_state:
             signals.event.registration_state_updated.send(self, previous_state=initial_state)
 
+    def has_conflict(self):
+        """Check if there are other valid registrations for the same user.
+
+        This is intended for cases where this registration is currenly invalid
+        (rejected or withdrawn) to determine whether it would be acceptable to
+        restore it.
+        """
+        conflict_criteria = [Registration.email == self.email]
+        if self.user_id is not None:
+            conflict_criteria.append(Registration.user_id == self.user_id)
+        return (Registration.query
+                .with_parent(self.registration_form)
+                .filter(Registration.id != self.id,
+                        db.or_(*conflict_criteria),
+                        Registration.state.notin_([RegistrationState.rejected, RegistrationState.withdrawn]))
+                .has_rows())
+
 
 class RegistrationData(StoredFileMixin, db.Model):
     """Data entry within a registration for a field in a registration form"""
--- a/indico/modules/events/registration/models/registrations.py
+++ b/indico/modules/events/registration/models/registrations.py
@@ -290,6 +290,20 @@ class Registration(db.Model):
         return regform.is_modification_open and regform.is_modification_allowed(self)
 
     @property
+    def can_be_withdrawn(self):
+        from indico.modules.events.registration.models.forms import ModificationMode
+        if self.is_paid:
+            return False
+        elif self.event.end_dt < now_utc():
+            return False
+        elif self.registration_form.modification_mode == ModificationMode.not_allowed:
+            return False
+        elif self.registration_form.modification_end_dt and self.registration_form.modification_end_dt < now_utc():
+            return False
+        else:
+            return True
+
+    @property
     def data_by_field(self):
         return {x.field_data.field_id: x for x in self.data}
 
@@ -456,13 +470,13 @@ class Registration(db.Model):
         if self.state != initial_state:
             signals.event.registration_state_updated.send(self, previous_state=initial_state)
 
-    def update_state(self, approved=None, paid=None, rejected=None, _skip_moderation=False):
+    def update_state(self, approved=None, paid=None, rejected=None, withdrawn=None, _skip_moderation=False):
         """Update the state of the registration for a given action
 
         The accepted kwargs are the possible actions. ``True`` means that the
         action occured and ``False`` that it was reverted.
         """
-        if sum(action is not None for action in (approved, paid, rejected)) > 1:
+        if sum(action is not None for action in (approved, paid, rejected, withdrawn)) > 1:
             raise Exception("More than one action specified")
         initial_state = self.state
         regform = self.registration_form
@@ -478,22 +492,35 @@ class Registration(db.Model):
                 self.state = RegistrationState.complete
             elif rejected:
                 self.state = RegistrationState.rejected
+            elif withdrawn:
+                self.state = RegistrationState.withdrawn
         elif self.state == RegistrationState.unpaid:
             if paid:
                 self.state = RegistrationState.complete
             elif approved is False:
                 self.state = RegistrationState.pending
+            elif withdrawn:
+                self.state = RegistrationState.withdrawn
         elif self.state == RegistrationState.complete:
             if approved is False and payment_required is False and moderation_required:
                 self.state = RegistrationState.pending
             elif paid is False and payment_required:
                 self.state = RegistrationState.unpaid
+            elif withdrawn:
+                self.state = RegistrationState.withdrawn
         elif self.state == RegistrationState.rejected:
             if rejected is False and moderation_required:
                 self.state = RegistrationState.pending
             elif rejected is False and payment_required:
                 self.state = RegistrationState.unpaid
             elif rejected is False:
+                self.state = RegistrationState.complete
+        elif self.state == RegistrationState.withdrawn:
+            if withdrawn is False and moderation_required:
+                self.state = RegistrationState.pending
+            elif withdrawn is False and payment_required:
+                self.state = RegistrationState.unpaid
+            elif withdrawn is False:
                 self.state = RegistrationState.complete
         if self.state != initial_state:
             signals.event.registration_state_updated.send(self, previous_state=initial_state)
--- a/indico/modules/events/registration/templates/display/registration_summary.html
+++ b/indico/modules/events/registration/templates/display/registration_summary.html
@@ -46,6 +46,20 @@
                         {% endif %}>
                         {% trans %}Modify{% endtrans %}
                     </a>
+                    {% if registration.can_be_withdrawn %}
+                        <a class="i-button icon-exit"
+                           data-href="{{ url_for('.withdraw_registration', registration.locator.registrant) }}"
+                           data-method="POST"
+                           data-title="{% trans %}Withdraw{% endtrans %}"
+                           data-confirm="{% trans %}Are you sure that you want to withdraw your registration? This action cannot be undone.{% endtrans %}">
+                            {% trans %}Withdraw{% endtrans %}
+                        </a>
+                    {% else %}
+                        <a class="i-button icon-exit disabled"
+                           title="{% trans %}The registration can no longer be withdrawn. Please contact an organizer.{% endtrans %}">
+                            {% trans %}Withdraw{% endtrans %}
+                        </a>
+                    {% endif %}
                 </div>
             {% endif %}
             {% if registration.state.name == 'complete' and registration.registration_form.tickets_enabled and
--- a/indico/modules/events/registration/blueprint.py
+++ b/indico/modules/events/registration/blueprint.py
@@ -161,6 +161,8 @@ _bp.add_url_rule('/registrations/<int:reg_form_id>/', 'display_regform', display
                  methods=('GET', 'POST'))
 _bp.add_url_rule('/registrations/<int:reg_form_id>/edit', 'edit_registration_display',
                  display.RHRegistrationDisplayEdit, methods=('GET', 'POST'))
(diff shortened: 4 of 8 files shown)


CHANGES.rst
```diff
@@ -104,6 +104,8 @@ Improvements
 - Show the Indico version in the footer again (:issue:`4558`)
 - Event managers can upload a custom Book of Abstract PDF (:issue:`3039`,
   :pr:`4577`)
+- Allow registrants to withdraw their application (:issue:`2715`, thanks
+  :user:`brabemi` & :user:`omegak`)
```
ThiefMaster, line 108: I'd also reference the PR since sphinx-issues supports that as well nowadays:

```suggestion
- Allow registrants to withdraw their application (:issue:`2715`, :pr:`4585`,
  thanks :user:`brabemi` & :user:`omegak`)
```

indico/modules/events/registration/models/registrations.py
```diff
@@ -289,6 +289,15 @@ def can_be_modified(self):
...
         return regform.is_modification_open and regform.is_modification_allowed(self)
 
+    @property
+    def can_be_withdrawn(self):
+        if self.is_paid:
+            return False
+        elif self.event.end_dt < now_utc():
+            return False
+        else:
+            return True
```
ThiefMaster, line 299:
```suggestion
        return not self.is_paid and self.event.end_dt > now_utc()
```

I think here we don't lose any readability by using a more compact version

indico/modules/events/registration/templates/display/registration_summary.html
```diff
@@ -4,60 +4,76 @@
 {% set class_mapping = {'complete': 'accept', 'pending': 'warning', 'unpaid': 'warning'} %}
 
 <div class="action-box {{ class_mapping[registration.state.name] }}">
-   <div class="section">
```
ThiefMaster, line 7: I'd fix the indentation separately... it doesn't help the readability of the PR :/

indico/modules/events/registration/templates/display/registration_summary.html
```diff
@@ -4,60 +4,76 @@
...
+                        {% endif %}>
+                        {% trans %}Modify{% endtrans %}
+                    </a>
+                    {% if registration.can_be_withdrawn %}
+                        <a class="i-button icon-exit {% if not registration.can_be_modified %}disabled{% endif %}"
+                           data-href="{{ url_for('.withdraw_registration', registration.locator.registrant) }}"
+                           data-method="POST"
+                           data-title="{% trans %}Withdraw{% endtrans %}"
+                           data-confirm=
+                               "{% trans %}Are you sure that yo want to withdraw your registration? This action cannot be undone.{% endtrans %}">
```
ThiefMaster, line 55: I don't think wrapping here is particularly useful...

indico/modules/events/registration/templates/display/registration_summary.html
```diff
@@ -4,60 +4,76 @@
...
+                            {% trans %}Withdraw{% endtrans %}
+                        </a>
+                    {% else %}
+                        <a class="i-button icon-exit disabled"
+                           title="{% trans %}The registration can no longer be withdrawn. Please contact an
+                                    organizer.{% endtrans %}">
```
ThiefMaster, line 61: same here

ThiefMaster:
How does it currently work for events where the registration feature is used solely for participants, i.e. the registration was never open/scheduled and no modification is allowed.

I think in such a case it doesn't make much sense to allow withdrawing, especially since you'd only be able to get there by using the url to the registrations directly (in case of a meeting)...

OmeGak:
> How does it currently work for events where the registration feature is used solely for participants, i.e. the registration was never open/scheduled and no modification is allowed.

Same as with conferences. If the registration is not paid and it's before the end of the event, the withdraw button will be visible. Modifications being allowed and regform being open/scheduled are not taken into account to allow or disallow withdrawals.

> I think in such a case it doesn't make much sense to allow withdrawing, especially since you'd only be able to get there by using the url to the registrations directly (in case of a meeting)...

I would not disallow this. Someone may want to update their RSVP to a meeting after having accepted an invitation.

ThiefMaster:
I was more thinking about cases like our weekly meeting where we added everyone as participants but do not use the user-facing registration module at all. People withdrawing there would also mess up clones of the event where they'd be withdrawn from the beginning...

OmeGak:
While testing, I found that inviting existing Indico users to a registration is currently broken. 

```
  File "/Users/omegak/Code/orgs/indico/indico-un/.venv/lib/python2.7/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/omegak/Code/orgs/indico/indico-un/.venv/lib/python2.7/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/omegak/Code/orgs/indico/indico-un/indico/indico/web/flask/util.py", line 83, in wrapper
    return obj().process()
  File "/Users/omegak/Code/orgs/indico/indico-un/indico/indico/web/rh.py", line 275, in process
    res = self._do_process()
  File "/Users/omegak/Code/orgs/indico/indico-un/indico/indico/web/rh.py", line 245, in _do_process
    rv = self._process()
  File "/Users/omegak/Code/orgs/indico/indico-un/indico/indico/modules/events/registration/controllers/management/invitations.py", line 70, in _process
    if form.validate_on_submit():
  File "/Users/omegak/Code/orgs/indico/indico-un/.venv/lib/python2.7/site-packages/flask_wtf/form.py", line 100, in validate_on_submit
    return self.is_submitted() and self.validate()
  File "/Users/omegak/Code/orgs/indico/indico-un/indico/indico/web/forms/base.py", line 141, in validate
    valid = super(IndicoForm, self).validate()
  File "/Users/omegak/Code/orgs/indico/indico-un/.venv/lib/python2.7/site-packages/wtforms/form.py", line 318, in validate
    return super(Form, self).validate(extra)
  File "/Users/omegak/Code/orgs/indico/indico-un/.venv/lib/python2.7/site-packages/wtforms/form.py", line 150, in validate
    if not field.validate(self, extra):
  File "/Users/omegak/Code/orgs/indico/indico-un/.venv/lib/python2.7/site-packages/wtforms/fields/core.py", line 226, in validate
    stop_validation = self._run_validation_chain(form, chain)
  File "/Users/omegak/Code/orgs/indico/indico-un/.venv/lib/python2.7/site-packages/wtforms/fields/core.py", line 246, in _run_validation_chain
    validator(form, self)
  File "/Users/omegak/Code/orgs/indico/indico-un/indico/indico/modules/events/registration/forms.py", line 191, in validate_users_field
    emails = {x['email'].lower() for x in field.data}
  File "/Users/omegak/Code/orgs/indico/indico-un/indico/indico/modules/events/registration/forms.py", line 191, in <setcomp>
    emails = {x['email'].lower() for x in field.data}
TypeError: 'User' object has no attribute '__getitem__'
```

This may be due to recent changes in `PrincipalListField`. I've added [a commit in this PR that solves it](https://github.com/indico/indico/pull/4585/commits/5c67206617b3ec891dfec08d003752f3280e846b), but I don't know if it's not taking into account some of the possible return values of `principal_from_identifier` in `PrincipalListField._convert_principal`. Could you take a look and confirm? Also, let me know if you want me to add this bug fix in `CHANGES.rst`.

OmeGak:
> I was more thinking about cases like our weekly meeting where we added everyone as participants but do not use the user-facing registration module at all. People withdrawing there would also mess up clones of the event where they'd be withdrawn from the beginning...

This sounds very specific to you workflow and this makes me think that perhaps it should be configurable. I would apply the same logic as for hiding/showing the "Modify" button and let managers configure until when withdrawals will be allowed with "modification deadline". 

Hard-coding end-of-event as last possible moment to withdraw is currently not documented and conference managers will have no way to control corner cases like the one you are describing.

ThiefMaster:
I cherrypicked the invitation fix into master.

> Also, let me know if you want me to add this bug fix in CHANGES.rst.

No, this bug was introduced in 2.3, so it's not part of any released version.

> but I don't know if it's not taking into account some of the possible return values

The way the field is configured, it only returns `User` objects. so everything fine

ThiefMaster:
> This sounds very specific to you workflow

It's actually the vast majority for meetings (event type 2). People add participants (like they did with the legacy "participants" feature in 1.x) but do not use any actual registration features.

```
indico_prod=> select e.type, count(*) from event_registration.forms rf join events.events e on (e.id = rf.event_id) where not e.is_deleted and not rf.is_deleted and not (rf.start_dt is null and rf.end_dt is null and rf.modification_mode = 3) and exists (select 1 from event_registration.registrations r where r.id = rf.id) group by e.type;
 type | count
------+-------
    1 |   284
    2 |  5443
    3 |  7442
(3 rows)

indico_prod=> select e.type, count(*) from event_registration.forms rf join events.events e on (e.id = rf.event_id) where not e.is_deleted and not rf.is_deleted and (rf.start_dt is null and rf.end_dt is null and rf.modification_mode = 3) and exists (select 1 from event_registration.registrations r where r.id = rf.id
) group by e.type;
 type | count
------+-------
    1 |   116
    2 | 38385
    3 |    53
(3 rows)
```

We could add a new column to control the withdrawal mode and set a reasonable default depending on the event type (not allowed for meetings, allowed for the other event types).

OmeGak:
> We could add a new column to control the withdrawal mode and set a reasonable default depending on the event type (not allowed for meetings, allowed for the other event types).

Isn't this already the case for allowing modifications and modification deadline? Modifications are enabled by default in conferences and disabled by default in lectures and meetings. Then, it's unlikely that conference organizers expect registrants to withdraw when there is a modification deadline date, so we can also use `modification_end_dt` for that.

In the end I've gone for the following logic:
```python
@property
def can_be_withdrawn(self):
    from indico.modules.events.registration.models.forms import ModificationMode
    if self.is_paid:
        return False
    elif self.event.end_dt < now_utc():
        return False
    elif self.registration_form.modification_mode == ModificationMode.not_allowed:
        return False
    elif (self.registration_form.modification_end_dt < now_utc()
            if self.registration_form.modification_end_dt else False):
        return False
    else:
        return True
```

CHANGES.rst
```diff
@@ -104,6 +104,8 @@ Improvements
 - Show the Indico version in the footer again (:issue:`4558`)
 - Event managers can upload a custom Book of Abstract PDF (:issue:`3039`,
   :pr:`4577`)
+- Allow registrants to withdraw their application (:issue:`2715`, :pr:`#4166`,
```
ThiefMaster, line 107:
```suggestion
- Allow registrants to withdraw their application (:issue:`2715`, :pr:`4166`,
```

indico/modules/events/registration/models/registrations.py
```diff
@@ -289,6 +289,21 @@ def can_be_modified(self):
...
+    def can_be_withdrawn(self):
+        from indico.modules.events.registration.models.forms import ModificationMode
+        if self.is_paid:
+            return False
+        elif self.event.end_dt < now_utc():
+            return False
+        elif self.registration_form.modification_mode == ModificationMode.not_allowed:
+            return False
+        elif (self.registration_form.modification_end_dt < now_utc()
+              if self.registration_form.modification_end_dt else False):
```
ThiefMaster, line 302:
```suggestion
        elif (self.registration_form.modification_end_dt and
                self.registration_form.modification_end_dt < now_utc()):
```

indico/modules/events/registration/templates/display/registration_summary.html
```diff
@@ -34,7 +34,7 @@
         </div>
         <div class="toolbar">
             {% if registration.registration_form.modification_mode.name != 'not_allowed' %}
-                <div class="group">
+            <div class="group">
```
ThiefMaster, line 37: That change doesn't look intentional.

indico/modules/events/registration/templates/display/registration_summary.html
```diff
@@ -46,7 +46,21 @@
                         {% endif %}>
                         {% trans %}Modify{% endtrans %}
                     </a>
-                </div>
+                {% if registration.can_be_withdrawn %}
+                    <a class="i-button icon-exit"
+                        data-href="{{ url_for('.withdraw_registration', registration.locator.registrant) }}"
+                        data-method="POST"
+                        data-title="{% trans %}Withdraw{% endtrans %}"
+                        data-confirm="{% trans %}Are you sure that yo want to withdraw your registration? This action cannot be undone.{% endtrans %}">
```
ThiefMaster, line 54: 1 space too much indentation

indico/modules/events/registration/templates/display/registration_summary.html
```diff
@@ -46,7 +46,21 @@
...
+                        {% trans %}Withdraw{% endtrans %}
+                    </a>
+                {% else %}
+                    <a class="i-button icon-exit disabled"
+                        title="{% trans %}The registration can no longer be withdrawn. Please contact an organizer.{% endtrans %}">
```
ThiefMaster, line 59: same here

indico/modules/events/registration/templates/display/registration_summary.html
ThiefMaster, line 54: yo, wanna withdraw your registration? :D 🙈

Issue #4589, opened 2020-08-11
Refactor the room booking main page


plourenco:
The room booking could use a style refactor, to become easier to maintain and cope with new changes.
There are some problems affecting the current implementation, such as:
- The top bar content isn't centred, this might seem normal but can misplace new elements we might add.
- The bottom numbers hide when the vertical space is too small, causing the scroll to show an empty space below.

![image](https://user-images.githubusercontent.com/12183954/89902918-f18d1880-dbde-11ea-80d5-f483e4123213.png)

ThiefMaster:
I think hiding the stats if there's not enough space makes sense - it just should not result in blank space showing up.

ThiefMaster:
Just so we don't forget about it: The hacky "fix" from #4590 should be removed again once we fix it in a cleaner way.

PR #4590, opened 2020-08-11
Quick-fix room booking un-scrollable content

plourenco:
Quick-fixes a problem causing the room booking to disable overflow even when the content is too small. 
This is being caused by the side-bar. We should come back to this in #4589.

### DIFF
--- a/indico/modules/rb/client/js/components/BookingBootstrapForm.jsx
+++ b/indico/modules/rb/client/js/components/BookingBootstrapForm.jsx
@@ -291,6 +291,7 @@ class BookingBootstrapForm extends React.Component {
               isOutsideRange={dt => {
                 return !isBookingStartDateValid(dt, isAdminOverrideEnabled, bookingGracePeriod);
               }}
+              withFullScreenPortal={window.innerHeight < 730}
             />
           </Form.Group>
         )}
@@ -302,6 +303,7 @@ class BookingBootstrapForm extends React.Component {
               disabledDate={dt => {
                 return !isBookingStartDateValid(dt, isAdminOverrideEnabled, bookingGracePeriod);
               }}
+              withFullScreenPortal={window.innerHeight < 730} // Temporary fix to ensure vertical responsiveness
             />
           </Form.Group>
         )}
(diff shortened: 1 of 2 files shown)


panagiotappl:
Doesn't that introduce a scrollbar on the right even when there's no scrolling to be done?
image

plourenco:
> Doesn't that introduce a scrollbar on the right even when there's no scrolling to be done?
> image

Yes, it was thanks. I did not notice that because my browser was merging them.
Should be good now, the child isn't scrollable anymore.

Also made the date field full-screen if the screen height is small, as suggested by @ThiefMaster.

indico/modules/rb/client/js/components/BookingBootstrapForm.jsx
```diff
@@ -291,9 +291,11 @@ class BookingBootstrapForm extends React.Component {
               isOutsideRange={dt => {
                 return !isBookingStartDateValid(dt, isAdminOverrideEnabled, bookingGracePeriod);
               }}
+              withFullScreenPortal={window.innerHeight < 730}
             />
           </Form.Group>
         )}
+        yolo
```
ThiefMaster, line 298: YØLØ

indico/modules/rb/client/js/components/BookingBootstrapForm.jsx
```diff
@@ -291,9 +291,11 @@ class BookingBootstrapForm extends React.Component {
               isOutsideRange={dt => {
                 return !isBookingStartDateValid(dt, isAdminOverrideEnabled, bookingGracePeriod);
               }}
+              withFullScreenPortal={window.innerHeight < 730}
```
ThiefMaster, line 294: We have a `responsiveReactDates` util where you can use the actual react-responsive utils to do this..

plourenco: But wouldn't that affect all date components that depend on this? 
I could, however, re-use a tablet dimension, even though we're using height here instead.

indico/modules/rb/client/js/components/BookingBootstrapForm.jsx
plourenco: omg

indico/modules/rb/client/js/components/BookingBootstrapForm.jsx
ThiefMaster: True.. depends on whether we want it for all responsive date components. If not, maybe add a prop to override the dimension if that can be done easily?

Issue #4594, opened 2020-08-17
Prevent rejected participant to register again


OmeGak:
**Is your feature request related to a problem? Please describe.**
After a registration is rejected, the associated user or email address should not be able to register again in the same registration form.

**Describe the solution you'd like**
The event registration form view should show the registration status as existing and the button should be "Check details", as with approved or pending registrations. The registration details view will then display "Your registration has been rejected". In case the rejected participant wants to retry, they can contact the conference organizer. Then the conference organizer can reset the rejected registration back to pending.

pferreir:
I don't have a strong opinion on this since it seems to be a pretty uncommon situation. I think staying in a "rejected" status is fine as long as we clearly show that.

OmeGak:
Is it so uncommon? To me it seems like the current UX encourages this situation. After a registration has been rejected/withdrawn, the event's main page welcomes the user with:

![image](https://user-images.githubusercontent.com/716307/90424270-1e996980-e0be-11ea-9bb2-4d1a4d2d257b.png)

, which will send them through the registration workflow and result in a new registration. I think this is problematic because I cannot imagine any circumstance where an event organizer is expecting rejected/withdrawn people to register again.

ThiefMaster:
In case of a withdrawn registration, I'd say it makes much more sense to let the user restore their registration if registering again would still be possible anyway (see #4588).

BTW maybe @pferreir meant is that it's uncommon that someone who was rejected wants to register again?

pferreir:
> BTW maybe @pferreir meant is that it's uncommon that someone who was rejected wants to register again?

Yes, indeed.

PR #4595, opened 2020-08-17
Prevent rejected/withdrawn participants to register again

OmeGak:
Addresses #4594.


indico/modules/events/registration/templates/display/registration_summary.html
```diff
@@ -11,6 +12,20 @@
...
+                </div>
+                {% trans -%}
+                    Contact an event manager if you think this is an error.
+                {%- endtrans %}
+            {% elif registration.state.name == 'withdrawn' %}
+                <div class="label">
+                    {% trans %}You withdrew your registration{% endtrans %}
+                </div>
+                {% trans -%}
+                    Contact an event manager if this was an error.
```
ThiefMaster, line 27: > Contact an event manager if you would like to restore your registration.

or "if you changed your mind" or something like that?

indico/modules/events/registration/templates/display/registration_summary.html
```diff
@@ -33,7 +48,7 @@
             {% endif %}
         </div>
         <div class="toolbar">
-            {% if registration.registration_form.modification_mode.name != 'not_allowed' %}
+            {% if registration.registration_form.modification_mode.name != 'not_allowed' and registration.is_active %}
```
ThiefMaster, line 51: Do we also enforce this on the server side? Or would an edit go through if you sent the right request (easiest way to test: start editing while the registration is active, withdraw in another tab, then try saving the edit).

CHANGES.rst
```diff
@@ -116,6 +116,9 @@ Improvements
 - Event modules can now be imported from another event (:issue:`4518`, thanks :user:`meluru`)
 - Event modules can now be imported from another event (:issue:`4518`, :pr:`4533`,
   thanks :user:`meluru`)
+- Allow registrants to check details for non-active registrations and prevent
+  them from registering twice with the same registration form (:issue:`4594`,
+  thanks :user:`omegak`)
```
ThiefMaster, line 121:
```suggestion
  :pr:`4595`, thanks :user:`omegak`)
```

indico/modules/events/registration/templates/display/registration_summary.html
OmeGak: Implicitly, it was like this already. This view could be accessed only for active registrations, but `RHRegistrationDisplayEdit` did not check if the registration is active, so someone could have potentially edited their non-active registration. I just added a `_check_access` that checks if `registration.is_active` here and in `RHRegistrationWithdraw`.

PR #4597, opened 2020-08-19
Typos

DirkHoffmann:
It's a trivial typo. But as we are using the english text as index for all translations, it's better not to delay its correction too much.
I would have liked to squash it, but failed. Maybe you are luckier than me.


ThiefMaster:
thx!

ThiefMaster:
i've also extracted+pushed the updated pots to transifex

PR #4599, opened 2020-08-20
Add event keywords in API data

### LINKED ISSUE #4598: Exposing the event's keywords in the API
This is a feature request to include the keywords of an event to the API. 

At the moment the keywords are seldom used, but when using the API that would be a solid use case, for example to refine the classification of events within a single category.

See the discussion on discourse: https://talk.getindico.io/t/event-keywords-in-the-export-api/1889/2

chernals:
closes #4598


ThiefMaster (approved):
Thanks! I amended your commit to give you proper credit in the changelog and updated the example output in the docs. It's pretty outdated anyway, but that way people at least know that the keywords are now included there!

PR #4603, opened 2020-08-21
Set requirement for flask_multipass to version 0.2

kadet1090:
Version 0.3 breaks compatibility with oauth providers.


ThiefMaster:
2.3 is not released yet; and when we do release that version we'll fully switch to the new multipass version as a requirement and thus provide documentation on how to update the config accordingly or see if we can somehow allow reuse of the old config with the new backend (basically automatically create an authlib-based config if the old oauth-based one is detected)

PR #4608, opened 2020-08-27
Allow extending polymorphic models in plugins

OmeGak:
Subclassing polymorphic models is not currently allowed from plugins. Example:

```python
class RegistrationFormSessionsSection(RegistrationFormSection):
    __mapper_args__ = {
        'polymorphic_identity': RegistrationFormItemType.section_sessions
    }
```

Blows up like this:

```
Exception: Plugin 'un' added a model which is not in a plugin schema ('RegistrationFormSessionsSection' in 'event_registration')
```

This PR solves it.

### DIFF
--- a/indico/core/plugins/__init__.py
+++ b/indico/core/plugins/__init__.py
@@ -116,6 +116,10 @@ class IndicoPlugin(Plugin):
         # which could contain more than one plugin this is not easily possible.
         for name, model in added_models:
             schema = model.__table__.schema
+            # Allow models with non-plugin schema if they specify `polymorphic_identity` without a dedicated table
+            if ('polymorphic_identity' in getattr(model, '__mapper_args__', ())
+                    and '__tablename__' not in model.__dict__):
+                continue
             if not schema.startswith('plugin_'):
                 raise Exception("Plugin '{}' added a model which is not in a plugin schema ('{}' in '{}')"
                                 .format(self.name, name, schema))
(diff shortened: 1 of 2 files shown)


indico/core/plugins/__init__.py
```diff
@@ -116,6 +116,9 @@ def _import_models(self):
         # which could contain more than one plugin this is not easily possible.
         for name, model in added_models:
             schema = model.__table__.schema
+            # Allow models with non-plugin schema if they specify `polymorphic_identity` without a dedicated table
+            if not hasattr(model, 'tablename') and 'polymorphic_identity' in getattr(model, '__mapper_args__', ()):
```
ThiefMaster, line 120: is it `tablename` or `__tablename__`? can the model have a `Non` tablename? If yes maybe better to check getattr

OmeGak: > can the model have a Non tablename?

True, this is not possible. I'm trying to account here for the possibility to use `polymorphic_identity` for [joined table inheritance](https://docs.sqlalchemy.org/en/13/orm/inheritance.html#joined-table-inheritance). In which case, you probably want the plugin to define the new table in its own plugin schema.

ThiefMaster: `'__tablename__' in model.__dict__` should do the job

ThiefMaster (approved):
I think our check in `_import_models` doesn't even work all the time, since usually the model already got imported when loading the plugin.. anyway, for the cases where it does work your change looks reasonable :)

PR #4618, opened 2020-09-02
Add more prominent draft mode indicator

ThiefMaster:
We had quite a few tickets from people not realizing that draft mode is enabled (or exists) and thus wondering why their timetable was not visible. This warning should make it much clearer.

image


Leats (approved):
Looks good! The warning seems very helpful!

PR #4620, opened 2020-09-07
Allow overriding language territory names

ThiefMaster:
Apparently "Chinese (China)" is not acceptable in some communities for political reasons, and they want "Chinese (Simplified)" instead.

### DIFF
--- a/docs/source/config/settings.rst
+++ b/docs/source/config/settings.rst
@@ -38,8 +38,8 @@ Authentication
 
     This setting controls whether local Indico groups are available.
     If no centralized authentication infrastructure that supports groups
-    (e.g. LDAP) is used, local groups are the only way to define groups in 
-    Indico, but if you do have central groups it may be useful to disable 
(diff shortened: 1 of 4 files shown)


docs/source/config/settings.rst
```diff
@@ -308,6 +308,30 @@ Customization
...
 
+.. data:: CUSTOM_LANGUAGES
+
+    A dict with language/territory name overrides.  This can be useful if the
+    official territory name that goes along with a language does not match what
+    your Indico instance's target audience expects for a country, e.g. due to
+    political situations.
+
+    For example, to use "Chinese (Simplified)" instead of "Chinese (China)",
+    you would use this
```
plourenco, line 319:
```suggestion
    For example, to replace "Chinese (Simplified)" with "Chinese (China)",
    you would use:
```

plourenco (approved):
Works great here! 👍
Just a minor semantic suggestion below.

PR #4623, opened 2020-09-08
Catch requests.TooManyRedirects

bcc:
Simple fix to catch requests.TooManyRedirects exception and put a ‘broken image’ placeholder in the pdf, much like already happens when a downloaded image isn’t a valid image.

The current handling results in an exception email that doesn't identify the failing URL, nor does the feedback to the user help them identify the problematic content. We see this every few weeks on a moderately popular Indico installation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -142,6 +142,7 @@ Bugfixes
 - Fix empty entries in corresponding authors (:pr:`4604`)
 - Actually prevent users from editing registrations if modification is
   disabled
+- Handle latex images with broken redirects (:pr:`4623`, thanks :user:`bcc`)
 
 Internal Changes
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Thanks, looks reasonable. I guess we never expected people to link to images that redirect forever ;)

ThiefMaster (approved):
If you want to add a changelog entry ("Handle latex images with broken redirects" or something like that) with credits for yourself (same way as for other entries in CHANGES.rst from external contributors) feel free to add that! Otherwise I'll merge it later/tomorrow once CI finished.

CHANGES.rst
```diff
@@ -142,6 +142,7 @@ Bugfixes
 - Fix empty entries in corresponding authors (:pr:`4604`)
 - Actually prevent users from editing registrations if modification is
   disabled
+- Handle latex images with broken redirects (:pr:`4623`, thanks :user:`bcc`)
```
ThiefMaster, line 145: `LaTeX` ;)

bcc: doh! ;)

PR #4628, opened 2020-09-15
Add endpoint for event preparation

adl1995:
This PR related to this issue: https://github.com/indico/newdle/issues/128

It will be used by [Newdle](https://github.com/indico/newdle) to automate event (meeting) creation on Indico.


indico/web/client/js/jquery/extensions/global.js
```diff
@@ -247,19 +247,31 @@ $(document).ready(function() {
   showFormErrors();
 
   // Show form creation dialog if hash is present in the URL
-  var match = location.hash.match(/^#create-event:(lecture|meeting|conference)(?::(\d+))?$/);
+  var match = location.hash.match(/^#create-event:(lecture|meeting|conference)(?::(\d*))(?::(\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b))?$/);
```
ThiefMaster, line 250: Please, no uuid regex :D I'd go for something like this...

    const match = location.hash.match(/^#create-event:(lecture|meeting|conference)(?::(\d*)(?::([a-f0-9-]{36}))?)?$/);

Like this it will still match all these cases, but without this awfully long regex:

- `#create-event:meeting`
- `#create-event:meeting:123`
- `#create-event:meeting::c730a998-c8eb-4d85-b398-c8821354213b`
- `#create-event:meeting:123:c730a998-c8eb-4d85-b398-c8821354213b`

`match[2]` is always the category id (empty string or undefined if missing)
`match[3]` is the uuid (or undefined if missing)

indico/web/client/js/jquery/extensions/global.js
```diff
@@ -247,19 +247,31 @@ $(document).ready(function() {
...
       lecture: $T.gettext('Create new lecture'),
       meeting: $T.gettext('Create new meeting'),
       conference: $T.gettext('Create new conference'),
     }[eventType];
-    var url =
-      match[2] !== undefined
-        ? build_url(Indico.Urls.EventCreation, {event_type: eventType, category_id: categoryId})
-        : build_url(Indico.Urls.EventCreation, {event_type: eventType});
+    var url;
+    if (match[2] !== "") {
```
ThiefMaster, line 264: `if (match[2]) {` is enough

indico/web/client/js/jquery/extensions/global.js
```diff
@@ -247,19 +247,31 @@ $(document).ready(function() {
...
     var title = {
       lecture: $T.gettext('Create new lecture'),
       meeting: $T.gettext('Create new meeting'),
       conference: $T.gettext('Create new conference'),
     }[eventType];
-    var url =
-      match[2] !== undefined
-        ? build_url(Indico.Urls.EventCreation, {event_type: eventType, category_id: categoryId})
-        : build_url(Indico.Urls.EventCreation, {event_type: eventType});
+    var url;
```
ThiefMaster, line 263: always use `let` or `const` in new code

indico/web/client/js/jquery/extensions/global.js
```diff
@@ -247,19 +247,31 @@ $(document).ready(function() {
...
+    if (match[2] !== "") {
+      url = build_url(Indico.Urls.EventCreation, {event_type: eventType, category_id: categoryId});
+    } else if (match[3] !== "") {
+      url = build_url(Indico.Urls.EventCreation, {event_type: eventType, event_uuid: eventUUID});
+    } else {
+      url = build_url(Indico.Urls.EventCreation, {event_type: eventType});
+    }
```
ThiefMaster, line 270: While not relevant for now I'd also support the case where we have both a category id and a uuid.

indico/modules/events/controllers/creation.py
```diff
@@ -123,3 +137,44 @@ def _process(self):
                                 event_type=self.event_type.name, single_category=self.single_category,
                                 check_room_availability=check_room_availability,
                                 rb_excluded_categories=rb_excluded_categories)
+
+class RHCreateEventUUID(RH):
```
ThiefMaster, line 141: I don't think we need a new class here. Let's just pass the UUID as a query string argument to the one we already have and populate the form with the details from cache in that case instead of the usual defaults.

ThiefMaster (changes requested):

indico/web/client/js/jquery/extensions/global.js
adl1995: Sure, updated :)

indico/web/client/js/jquery/extensions/global.js
adl1995: Added a condition for when both category ID and a UUID are present. Does this need to be implemented now?

indico/web/client/js/jquery/extensions/global.js
adl1995: Thanks, updated.

indico/modules/events/controllers/creation.py
adl1995: I can't recall why this was added (it's not used). Removed

adl1995:
@ThiefMaster - Can you review this again (sorry for the long delay)?

indico/web/client/js/jquery/extensions/global.js
ThiefMaster: I would just build the URL accordingly:

    url = build_url(Indico.Urls.EventCreation, {event_type: eventType, category_id: categoryId, event_uuid: eventUUID});

That way it will Just Work ™ if we ever have something passing data like this.

indico/modules/events/controllers/preparation.py
```diff
@@ -0,0 +1,42 @@
...
+    def _process(self, title, start_dt, duration):
+        event_key = unicode(uuid4())
+        self._cache.set(
+            event_key,
+            {
+                "title": title,
+                "start_dt": start_dt,
+                "duration": duration,
+            },
+            90
```
ThiefMaster, line 40: `3600` maybe? I mean otherwise reloading the event creation page (for whatever reason) after 90s will result in an error (or the form not being pre-populated with data)

indico/modules/events/controllers/creation.py
```diff
@@ -57,16 +59,25 @@ def _get_form_defaults(self):
...
-            start_dt = tzinfo.localize(datetime.combine(now.date() + relativedelta(days=1), time(9)))
-        end_dt = start_dt + relativedelta(hours=2)
+        title = None
+        try:
+            cached_data = self._cache.get(request.args['event_uuid'])
+            title = cached_data['title']
+            start_dt = datetime.strptime(cached_data['start_dt'], "%Y-%m-%d %H:%M:%S")
+            start_dt = tzinfo.localize(start_dt)
+            end_dt = start_dt + relativedelta(minutes=int(cached_data['duration']))
+        except Exception:
```
ThiefMaster, line 69: Please don't ever do that except in VERY exceptional cases where you really want to ignore all errors (or when you re-`raise` at the end)

indico/modules/events/controllers/creation.py
```diff
@@ -36,6 +37,7 @@
 
 class RHCreateEvent(RHProtected):
     """Create a new event."""
+    _cache = GenericCache('event-preparation')
```
ThiefMaster, line 40: classes always have an empty line after their docstring

indico/modules/events/controllers/preparation.py
```diff
@@ -0,0 +1,42 @@
...
+from indico.legacy.common.cache import GenericCache
+from indico.web.args import use_kwargs
+from indico.web.rh import RH
+
+
+class RHPrepareEvent(RH):
+    """Prepare a new event, store it using GenericCache util, and create a UUID."""
+
+    CSRF_ENABLED = False
+    _cache = GenericCache('event-preparation')
```
ThiefMaster, line 24: I'd move this inside `_process`, i.e.:

```python
cache = GenericCache(...)
cache.set(..)
```

indico/modules/events/controllers/creation.py
ThiefMaster:
```python
prepared_data = self._get_prepared_data()
if prepared_data:
    tz, start_dt, end_dt, title = prepared_data
else:
    # try to find good dates/times
    ...
```

and a method like this:

```python
def _get_prepared_data(self):
    try:
        data = self._cache.get(request.args['event_uuid'])
    except KeyError:
        return None
    if not data:
        return None
    tz = data['timezone']
    start_dt = datetime.strptime(data['start_dt'], '%Y-%m-%d %H:%M')
    start_dt = timezone(tz).localize(start_dt)
    end_dt = start_dt + relativedelta(minutes=cached_data['duration'])
    return tz, start_dt, end_dt, title
```

Note: the timezone needs to be passed from newdle since we know the timezone there, but all we have here would be the default timezone which doesn't necessarily match the one from the newdle!

You could also move the parsing of the time string into the preparation endpoint and already store a datetime object in the cache entry.

indico/modules/events/controllers/preparation.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @use_kwargs({
+        'title': fields.String(missing=""),
+        'start_dt': fields.String(missing=""),
```
ThiefMaster, line 28: Maybe these should be `required=True` instead of having a `missing` value?

ThiefMaster, line 28: We can use the custom `NaiveDateTime` field type (from `indico.util.marshmallow`) here to get an actual datetime object instead of just a string.

indico/modules/events/controllers/preparation.py
```diff
@@ -0,0 +1,42 @@
...
+        self._cache.set(
+            event_key,
+            {
+                "title": title,
+                "start_dt": start_dt,
+                "duration": duration,
+            },
+            90
+        )
+        return jsonify(url="{}#create-event:meeting::{}".format(request.url_root, event_key))
```
ThiefMaster, line 42:
```python
return jsonify(url=url_for_index(_anchor='create-event:...'))
```

indico/modules/events/controllers/preparation.py
adl1995: `3600` sounds reasonable. Updated.

indico/modules/events/controllers/creation.py
adl1995: Thanks for explaining ;)

I've added a `_get_prepared_data` method and am now passing timezone info from Newdle and storing the parsed datetime in cache.

indico/modules/events/controllers/preparation.py
adl1995: Makes sense, updated.

adl1995: Updated.

indico/modules/events/controllers/preparation.py
adl1995: Thanks, updated.

indico/modules/events/controllers/preparation.py
adl1995: Updated, thanks.

indico/modules/events/blueprint.py
```diff
@@ -42,6 +43,10 @@
 _bp.add_url_rule('/event/create/<any(lecture,meeting,conference):event_type>', 'create', RHCreateEvent,
                  methods=('GET', 'POST'))
 
+# Preparation
+_bp.add_url_rule('/event/prepare', 'prepare', RHPrepareEvent,
+                 methods=('GET', 'POST'))
```
ThiefMaster, line 48: This fits nicely in one line and we do not need GET

```suggestion
_bp.add_url_rule('/event/prepare', 'prepare', RHPrepareEvent, methods=('POST',))
```

indico/modules/events/controllers/preparation.py
```diff
@@ -0,0 +1,46 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2020 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
```
ThiefMaster, line 7: one blank line too much 👾

indico/modules/events/controllers/preparation.py
```diff
@@ -0,0 +1,46 @@
...
+from indico.legacy.common.cache import GenericCache
+from indico.util.marshmallow import NaiveDateTime
+from indico.web.args import use_kwargs
+from indico.web.rh import RH
+from indico.web.util import url_for_index
+
+
+class RHPrepareEvent(RH):
+    """Prepare a new event, store it using GenericCache util, and create a UUID."""
```
ThiefMaster, line 23: That's a bit too technical IMO (GenericCache usage is very much an implementation detail). What about something like this instead?

```suggestion
    """Prepare the creation of an event with some data."""
```

indico/modules/events/controllers/preparation.py
```diff
@@ -0,0 +1,46 @@
...
+
+    CSRF_ENABLED = False
+
+    @use_kwargs({
+        'title': fields.String(required=True),
+        'start_dt': NaiveDateTime(required=True),
+        'tz': fields.String(),
+        'duration': fields.Integer(),
```
ThiefMaster, line 31: I think these fields should be required as well

indico/web/client/js/jquery/extensions/global.js
```diff
@@ -247,19 +247,26 @@ $(document).ready(function() {
...
-      match[2] !== undefined
-        ? build_url(Indico.Urls.EventCreation, {event_type: eventType, category_id: categoryId})
-        : build_url(Indico.Urls.EventCreation, {event_type: eventType});
+    let url;
+    if (match[2]) {
+      url = build_url(Indico.Urls.EventCreation, {event_type: eventType, category_id: categoryId});
+    } else if (match[3]) {
+      url = build_url(Indico.Urls.EventCreation, {event_type: eventType, event_uuid: eventUUID});
+    } else if (match[2] && match[3]) {
+      // TODO: When both categoryId and eventUUID are present.
```
ThiefMaster, line 266: I think when we discussed this the last time we said we'd pass both params to `build_url`. It most likely works out of the box, even if nothing generates such URLs yet. You can easily test it yourself by opening `#create-event:meeting:1:<uuid>` in which case the dialog should have category 1 preselected.

indico/web/client/js/jquery/extensions/global.js
```diff
@@ -247,19 +247,26 @@ $(document).ready(function() {
...
-  var match = location.hash.match(/^#create-event:(lecture|meeting|conference)(?::(\d+))?$/);
+  let match = location.hash.match(/^#create-event:(lecture|meeting|conference)(?::(\d*)(?::([a-f0-9-]{36}))?)?$/);
   if (match) {
-    var eventType = match[1];
-    var categoryId = match[2];
-    var title = {
+    let eventType = match[1];
+    let categoryId = match[2];
+    let eventUUID = match[3];
+    let title = {
```
ThiefMaster, line 255: I don't think we ever reassign those, so `const` it is.

For the match you could also do something like this:

```js
const [, eventType, categoryId, eventUUID] = match;
```

indico/web/client/js/jquery/extensions/global.js
```diff
@@ -247,19 +247,26 @@ $(document).ready(function() {
...
       lecture: $T.gettext('Create new lecture'),
       meeting: $T.gettext('Create new meeting'),
       conference: $T.gettext('Create new conference'),
     }[eventType];
-    var url =
-      match[2] !== undefined
-        ? build_url(Indico.Urls.EventCreation, {event_type: eventType, category_id: categoryId})
-        : build_url(Indico.Urls.EventCreation, {event_type: eventType});
+    let url;
+    if (match[2]) {
```
ThiefMaster, line 261: Since you assign all those things to variables you can use them here as well instead of checking `match[x]` directly.

indico/modules/events/controllers/preparation.py
```diff
@@ -0,0 +1,46 @@
...
+
+class RHPrepareEvent(RH):
+    """Prepare a new event, store it using GenericCache util, and create a UUID."""
+
+    CSRF_ENABLED = False
+
+    @use_kwargs({
+        'title': fields.String(required=True),
+        'start_dt': NaiveDateTime(required=True),
+        'tz': fields.String(),
```
ThiefMaster, line 30: `fields.String(required=True, validate=OneOf(common_timezones_set))`

(you can import it from pytz)

indico/modules/events/controllers/preparation.py
ThiefMaster, line 31: this one could have some basic validation (>0)

indico/modules/events/controllers/creation.py
```diff
@@ -49,22 +50,40 @@ def _default_category(self):
...
-        now = now_utc(exact=False)
-        start_dt = now + relativedelta(hours=1, minute=0)
-        if start_dt.astimezone(tzinfo).time() > time(18):
-            start_dt = tzinfo.localize(datetime.combine(now.date() + relativedelta(days=1), time(9)))
-        end_dt = start_dt + relativedelta(hours=2)
+        prepared_data = self._get_prepared_data()
+        if prepared_data:
+            start_dt, end_dt, title = prepared_data
+        else:
+            # try to find good dates/times
```
ThiefMaster, line 76: title is never set when you get in this block, so let's add `title = ''` here

indico/web/client/js/jquery/extensions/global.js
```diff
@@ -247,19 +247,26 @@ $(document).ready(function() {
...
+      url = build_url(Indico.Urls.EventCreation, {event_type: eventType, category_id: categoryId});
+    } else if (match[3]) {
+      url = build_url(Indico.Urls.EventCreation, {event_type: eventType, event_uuid: eventUUID});
+    } else if (match[2] && match[3]) {
```
ThiefMaster, line 265: this one needs to come first, you never end up in this block right now because the last two if/elif conditions already matched when this one would match as well.

ThiefMaster: bah, github doesn't let me comment on the whole block with a suggestion, so i just paste it here. This is a version with all my comments fixed (and reformatted with prettier)

```js
  const match = location.hash.match(
    /^#create-event:(lecture|meeting|conference)(?::(\d*)(?::([a-f0-9-]{36}))?)?$/
  );
  if (match) {
    const [, eventType, categoryId, eventUUID] = match;
    const title = {
      lecture: $T.gettext('Create new lecture'),
      meeting: $T.gettext('Create new meeting'),
      conference: $T.gettext('Create new conference'),
    }[eventType];
    let url;
    if (categoryId && eventUUID) {
      url = build_url(Indico.Urls.EventCreation, {
        event_type: eventType,
        category_id: categoryId,
        event_uuid: eventUUID,
      });
    } else if (categoryId) {
      url = build_url(Indico.Urls.EventCreation, {event_type: eventType, category_id: categoryId});
    } else if (eventUUID) {
      url = build_url(Indico.Urls.EventCreation, {event_type: eventType, event_uuid: eventUUID});
    } else {
      url = build_url(Indico.Urls.EventCreation, {event_type: eventType});
    }
    ajaxDialog({url, title});
  }
```

indico/modules/events/controllers/preparation.py
```diff
@@ -0,0 +1,46 @@
...
+        cache.set(
+            event_key,
+            {
+                'title': title,
+                'start_dt': start_dt,
+                'duration': duration,
+            },
+            3600
+        )
+        return jsonify(url=url_for_index(_external=True, _anchor=f'create-event:meeting::{event_key}'))
```
ThiefMaster, line 46: While we do not need it for the newdle integration, let's add an optional `type` param (lecture/meeting/conference) that defaults to `meeting` if it's `missing`.

indico/modules/events/controllers/preparation.py
ThiefMaster: instead of making it required you can also use `missing=config.DEFAULT_TIMEZONE` to use the server's default timezone if none has been specified... not sure if it's really useful though, maybe it's cleaner to force people to be explicit..

indico/modules/events/blueprint.py
adl1995: Right, thanks.

indico/modules/events/controllers/preparation.py
adl1995: Simplified.

indico/web/client/js/jquery/extensions/global.js
adl1995: Didn't realize it would work as is. Updated.

indico/web/client/js/jquery/extensions/global.js
adl1995: Much cleaner, thanks.

indico/modules/events/controllers/preparation.py
adl1995: Also added `missing=config.DEFAULT_TIMEZONE`. I think it adds some flexibility.

indico/modules/events/controllers/preparation.py
adl1995: Added `'duration': fields.Integer(required=True, validate=lambda v: v > 0),`.

indico/modules/events/controllers/creation.py
adl1995: Thanks.

indico/web/client/js/jquery/extensions/global.js
adl1995: Thanks, updated.

indico/modules/events/controllers/preparation.py
adl1995: Added.

ThiefMaster (approved):

PR #4629, opened 2020-09-15
Add commands for building translations

adl1995:


docs/source/installation/development.rst
```diff
@@ -139,6 +139,13 @@ Then, follow the instructions given by the wizard::
 
     indico setup wizard --dev
 
+To build translations for locales other than "en_US", use:
```
ThiefMaster, line 142:
```suggestion
To build the locales, use:
```

default is enGB and even for that things don't work properly if they aren't built

PR #4630, opened 2020-09-16
Move material package generation to Celery

Leats:


indico/modules/attachments/controllers/display/event.py
```diff
@@ -53,3 +55,16 @@ def wp(self):
             return WPPackageEventAttachmentsDisplayConference
         else:
             return WPPackageEventAttachmentsDisplay
+
+
+class RHPackageEventAttachmentsStatus(RHDisplayEventBase):
```
ThiefMaster, line 60: I just realized this might be better suited in `event_package.py`

indico/modules/attachments/controllers/display/event.py
```diff
@@ -53,3 +55,16 @@ def wp(self):
...
+    def _process(self):
+        res = AsyncResult(request.view_args['task_id'])
+        try:
+            download_url = res.get(5, propagate=False)
+        except TimeoutError:
+            return jsonify(pending=True)
+        if res.successful():
+            return {'pending': False, 'download_url': download_url}
+        else:
+            raise RuntimeError('Material package generation failed')
```
ThiefMaster, line 70: i18n

indico/modules/attachments/tasks.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@celery.task(ignore_result=False)
+def generate_materials_package(attachment_ids, event):
+    from indico.modules.attachments.controllers.event_package import AttachmentPackageGeneratorMixin
+    attachments = Attachment.query.filter(Attachment.id.in_(attachment_ids)).all()
+    attachment_package_mixin = AttachmentPackageGeneratorMixin()
+    attachment_package_mixin.event = event
+    generated_zip = attachment_package_mixin._generate_zip_file(attachments, return_file=True)
+    f = File(filename='material-package.zip', content_type='application/zip')
```
ThiefMaster, line 21:
```suggestion
    f = File(filename='material-package.zip', content_type='application/zip', meta={'event_id': event.id})
```

It's not really needed but maybe still nice to have some metadata there, especially since we never claim this File...

indico/modules/attachments/controllers/display/event.py
ThiefMaster, line 70: We could discard the result once we are done since we'll never need it again anyway:

```suggestion
        try:
            if res.successful():
                return {'pending': False, 'download_url': download_url}
            else:
                raise RuntimeError('Material package generation failed')
        finally:
            res.forget()
```

ThiefMaster: `IndicoError` instead of `RuntimeError` looks nicer for the user (no `RuntimeError:` prefix in the message)

ThiefMaster, line 70: `IndicoError` instead of `RuntimeError` looks nicer for the user (no `RuntimeError:` prefix in the message)

ThiefMaster (approved):

PR #4632, opened 2020-09-16
Typo

DirkHoffmann:
No comment.

Is there a better way to push trivial corrections than to open a PR?


ThiefMaster:
pushing the commit to a branch and pinging us in our IRC channel would be an alternative.. then we can just cherrypick the commit without a PR

PR #4634, opened 2020-09-17
Add compilation documentation

plourenco:
Includes a guide to compile Indico. 
Particularly useful for translators that which to expose a translation we do not officially support, in their production version.


docs/source/compilation/index.rst
```diff
@@ -0,0 +1,23 @@
+.. _compilation:
+
+Compilation
```
ThiefMaster, line 3: I think "Building" would be a better title. Also, the number of `=` in the line below should be equal to the length of this one ;)

docs/source/compilation/index.rst
```diff
@@ -0,0 +1,23 @@
...
+=============
+
+Before starting Indico compilation, this guide assumes you've previously
+:ref:`cloned the development version <cloning>`.
+The first step is to generate a local distribution archive. Navigate to the Indico source folder
+(by default, it is ``~/dev/indico/src``) and run the following commands::
+
+    ./bin/maintenance/build-wheel.py --target-dir dist indico --add-version-suffix
```
ThiefMaster, line 11: `--target-dir dist` can be omitted, it's the default anyway and people pretty much never need to override it.

ThiefMaster, line 11: We should tell people that they must not be running the `build-assets.py` script in `--watch` mode while building a release wheel

docs/source/compilation/index.rst
```diff
@@ -0,0 +1,23 @@
...
+
+
+This will generate a wheel distribution in the ``dist`` folder::
+
+    dist/
+      indico-2.3.dev0-py2-none-any.whl
```
ThiefMaster, line 17: maybe use a filename that contains the suffix here? ;)

docs/source/compilation/index.rst
```diff
@@ -0,0 +1,23 @@
...
+
+
+To deploy this distribution, you should follow the `production installation guide <compilation/index.rst>`_,
```
ThiefMaster, line 20: wrong link target (best to use `:ref:` for it)

docs/source/compilation/index.rst
```diff
@@ -0,0 +1,23 @@
...
+but replacing the indexed package in Indico installation with your local wheel from the previous step::
+
+    pip install ~/dev/indico/src/dist/indico-2.3.dev0-py2-none-any.whl
```
ThiefMaster, line 23: People should not be building indico on their production machine, that's just asking for them to make a mess or mix things between dev and prod. So I would be clear about *coping* the wheel to their prod machine (e.g. wit `scp`) and installing it there.

docs/source/compilation/index.rst
```diff
@@ -0,0 +1,23 @@
...
+
+
+This will generate a wheel distribution in the ``dist`` folder::
+
+    dist/
+      indico-2.3.dev0-py2-none-any.whl
+
+
+To deploy this distribution, you should follow the `production installation guide <compilation/index.rst>`_,
+but replacing the indexed package in Indico installation with your local wheel from the previous step::
```
ThiefMaster, line 21: but instead of doing `pip install indico`, use `pip install /tmp/indico-....whl` instead

or something like that

docs/source/index.rst
```diff
@@ -39,6 +39,15 @@ Configuration
     config/index.rst
 
 
+Compilation
++++++++++++++
```
ThiefMaster, line 43: see my other comment

docs/source/compilation/index.rst
```diff
@@ -0,0 +1,23 @@
+.. _compilation:
```
ThiefMaster, line 1: If you don't use a ref to link to this page you don't need this

docs/source/compilation/index.rst
```diff
@@ -0,0 +1,23 @@
...
+
+Compilation
+=============
+
+Before starting Indico compilation, this guide assumes you've previously
+:ref:`cloned the development version <cloning>`.
```
ThiefMaster, line 7: ...and followed the dev setup guide

because if they *only* clone it they won't be able to build things

ThiefMaster:
when adding a new translation people should also enable the moment locale in `indico/web/client/js/jquery/index.js`:

```js
// moment.js locales
import 'moment/locale/zh-cn';
import 'moment/locale/es';
import 'moment/locale/fr';
import 'moment/locale/en-gb';
```

new locales should be added to the top (never at the end, since the default english one needs to be the last oen in the list)

docs/source/compilation/index.rst
plourenco: Well, I didn't because this one looks simpler. But sure I can include it.

docs/source/compilation/index.rst
plourenco: Good idea

docs/source/compilation/index.rst
plourenco: Yes, I guess I can include the whole setup. I was trying to avoid pointing someone towards setting up mail-dump, the database and running.

docs/source/compilation/index.rst
plourenco: Sounds good, it was a bit hard to find a place to include it but should be good now.

PR #4638, opened 2020-09-21
Add Session basic schema

OmeGak:

### DIFF
--- a/indico/core/db/sqlalchemy/locations.py
+++ b/indico/core/db/sqlalchemy/locations.py
@@ -199,20 +199,30 @@ class LocationMixin(object):
     def venue_name(self, venue_name):
         self.own_venue_name = venue_name
 
-    def get_room_name(self, full=True):
+    def get_room_name(self, full=True, verbose=False):
         """The name of the room where this item is located.
 
-        :param full: by default, if the room has a "friendly name" too
-                     (e.g. 'Main Amphitheatre'), a composite name will be
-                     returned. If ``full`` is set to ``False``, only the
-                     "friendly name" will be returned in that case.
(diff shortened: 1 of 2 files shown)


indico/modules/events/sessions/schemas.py
```diff
@@ -0,0 +1,33 @@
...
+
+class SessionSchema(mm.ModelSchema):
+    blocks = fields.Nested(SessionBlockSchema, many=True)
+
+    class Meta:
+        model = Session
+        fields = ('id', 'title', 'friendly_id', 'blocks')
+
+
+session_schema_basic = SessionSchema()
```
ThiefMaster, line 33: I'd call the class itself `BasicSessionSchema` (iirc we discussed this on matrix?), and completely get rid of this line - instantiating the schema is cheap, so you can just do this whenever you need to use it.. `SessionSchema().dump()` etc (and that way you can even specify e.g. `many=True` in the constructor when you need it)

ThiefMaster (approved):

indico/core/db/sqlalchemy/locations.py
```diff
@@ -199,20 +199,30 @@ def venue_name(self):
...
+        assert sum([full, verbose]) <= 1
         if self.inherit_location and self.location_parent is None:
             return ''
         room = self.room
         if room is not None:
-            return room.full_name if full else room.name
-        return self.own_room_name if not self.inherit_location else self.location_parent.room_name
+            if full:
+                return room.full_name
+            elif verbose:
```
ThiefMaster, line 220:
```suggestion
            elif verbose and room.verbose_name:
```

otherwise this returns None for rooms that have no verbose name, which is probably not what you want

PR #4642, opened 2020-09-28
Add signals to intercept regform fields processing

OmeGak:
This allows plugins to add arbitrary WTForms fields in a registration form and later process them on registration creation/modification.

### DIFF
--- a/indico/modules/events/registration/util.py
+++ b/indico/modules/events/registration/util.py
@@ -161,6 +161,8 @@ def make_registration_form(regform, management=False, registration=None):
 
         field_impl = form_item.field_impl
         setattr(RegistrationFormWTF, form_item.html_field_name, field_impl.create_wtf_field())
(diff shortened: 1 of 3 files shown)


indico/modules/events/registration/util.py
```diff
@@ -161,6 +161,7 @@ def validate_email(self, field):
 
         field_impl = form_item.field_impl
         setattr(RegistrationFormWTF, form_item.html_field_name, field_impl.create_wtf_field())
+    signals.event.registration_form_wtform_created.send(regform, wtform=RegistrationFormWTF)
```
ThiefMaster, line 164: I'd pass `management` and `registration` to the signal as well. Just in case it's useful/needed at some point.

ThiefMaster: What about `form_cls` or `wtform_cls` instead of `wtform` (which could also be an *instance* of a form)?

ThiefMaster (approved):

PR #4647, opened 2020-10-01
Allow signal handlers to modify list of registrations to be printed

OmeGak:


indico/modules/events/registration/util.py
```diff
@@ -527,7 +527,7 @@ def generate_ticket(registration):
     signals.event.designer.print_badge_template.send(template, regform=registration.registration_form,
                                                      registrations=[registration])
```
ThiefMaster, line 528: modifying the list here won't do anything. maybe assign the list to a variable before calling the signal and use that below?

ThiefMaster (approved):

PR #4649, opened 2020-10-02
Make docstrings conform to PEP 257

Leats:
Docstrings should now conform to [PEP 257](https://www.python.org/dev/peps/pep-0257/).

This includes:
- descriptions as command (`Return` vs `Returns`)
- docstring phrases ending in a period
- correct indentation and line breaks for multi-line comments
- blank lines after docstrings for classes
- docstrings that were too long were split into several lines instead

Especially in newer code these things were already mostly present. The older code was changed accordingly.


ThiefMaster (approved):
Thanks a lot! I only skimmed the diff but I didn't spot any problems, so I'll go ahead and merge this :)

PR #4657, opened 2020-10-09
Refactor processRevisions selector logic

plourenco:
The `processRevisions` selector is becoming too complex mainly because it is describing all the possible combination of states in revisions, such as:

```
[InitialRevisionState.needs_submitter_confirmation, FinalRevisionState.needs_submitter_changes]
```
meaning **Submitter rejected proposed changes**.

The purpose of this is to refactor it, implementing a less verbose state machine.

---

Should also fix a bug causing custom revision actions not to show up.

This is a bug introduced in #4641 causing custom actions to disappear in a block with more than one revision. Caused because we may group multiple revisions in blocks, causing the client to use the actions for the first revision in the block, instead of the latest (the most recent one).

## How to reproduce

- Make sure the micro-service is connected with an action to replace the revision as soon as it is submitted
- Make sure the micro-service exposes custom actions
- Upload a paper for editing
- Accept the paper

The actions will no longer show up.

### DIFF
--- a/indico/modules/events/editing/client/js/editing/timeline/RevisionLog.jsx
+++ b/indico/modules/events/editing/client/js/editing/timeline/RevisionLog.jsx
@@ -10,15 +10,18 @@ import React from 'react';
 
 import CommentItem from './CommentItem';
 import CustomItem from './CustomItem';
+import {blockItemPropTypes} from './util';
 
-export default function RevisionLog({items, children, separator}) {
+export default function RevisionLog({items, state, children, separator}) {
   return (
     <div className="i-timeline">
       <div className="i-timeline with-line">
         <div className="i-timeline-connect-up" />
         {items.map(item => {
-          const Component = item.custom ? CustomItem : CommentItem;
-          return <Component key={item.id} {...item} />;
+          if (item.custom) {
+            return <CustomItem key={item.id} item={item} state={state} />;
+          }
+          return <CommentItem key={item.id} {...item} />;
         })}
         {children}
       </div>
@@ -35,7 +38,8 @@ export default function RevisionLog({items, children, separator}) {
 }
 
 RevisionLog.propTypes = {
-  items: PropTypes.array.isRequired,
+  items: PropTypes.arrayOf(PropTypes.shape(blockItemPropTypes)).isRequired,
+  state: PropTypes.string,
   children: PropTypes.node,
   separator: PropTypes.bool,
 };
@@ -43,4 +47,5 @@ RevisionLog.propTypes = {
 RevisionLog.defaultProps = {
   children: null,
   separator: false,
+  state: null,
 };
--- a/indico/modules/events/editing/client/js/editing/timeline/RevisionLog.jsx
+++ b/indico/modules/events/editing/client/js/editing/timeline/RevisionLog.jsx
@@ -8,6 +8,8 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
+import {FinalRevisionState} from '../../models';
+
 import CommentItem from './CommentItem';
 import CustomItem from './CustomItem';
 import {blockItemPropTypes} from './util';
@@ -39,7 +41,7 @@ export default function RevisionLog({items, state, children, separator}) {
 
 RevisionLog.propTypes = {
   items: PropTypes.arrayOf(PropTypes.shape(blockItemPropTypes)).isRequired,
-  state: PropTypes.string,
+  state: PropTypes.oneOf(Object.values(FinalRevisionState)),
   children: PropTypes.node,
   separator: PropTypes.bool,
 };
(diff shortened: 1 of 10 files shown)


ThiefMaster:
I moved the PR back to draft because of the changes to the processRevisions logic we discussed offline.

indico/modules/events/editing/client/js/editing/timeline/util.js
```diff
@@ -7,8 +7,44 @@
...
+  any: {
+    [FinalRevisionState.replaced]: Translate.string('Revision has been replaced'),
+    [FinalRevisionState.accepted]: Translate.string('Revision has been accepted'),
+    [FinalRevisionState.rejected]: Translate.string('Revision has been rejected'),
+    [FinalRevisionState.needs_submitter_changes]: Translate.string(
+      'Submitter has been asked to make some changes'
+    ),
+    [FinalRevisionState.needs_submitter_confirmation]: revision =>
+      Translate.string('{editorName} (editor) has made some changes to the paper', {
+        editorName: revision.editor ? revision.editor.fullName : '',
```
ThiefMaster, line 37: Would this work here?

```suggestion
        editorName: revision.editor?.fullName || '',
```

indico/modules/events/editing/client/js/editing/timeline/selectors.js
```diff
@@ -144,7 +57,9 @@ export const isInitialEditableDetailsLoading = state =>
 export const isTimelineOutdated = createSelector(
   getDetails,
   getNewDetails,
-  (details, newDetails) => newDetails !== null && !_.isEqual(details, newDetails)
+  (details, newDetails) => {
+    return newDetails !== null && !_.isEqual(details, newDetails);
```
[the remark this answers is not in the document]
plourenco: Exactly

indico/modules/events/editing/client/js/editing/timeline/util.js
plourenco: Yes, sounds good 👍

indico/modules/events/editing/client/js/editing/timeline/RevisionLog.jsx
```diff
@@ -35,12 +38,14 @@ export default function RevisionLog({items, children, separator}) {
 }
 
 RevisionLog.propTypes = {
-  items: PropTypes.array.isRequired,
+  items: PropTypes.arrayOf(PropTypes.shape(blockItemPropTypes)).isRequired,
+  state: PropTypes.string,
```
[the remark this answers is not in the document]
plourenco: Not sure I understood, could you elaborate a bit more?

indico/modules/events/editing/operations.py
```diff
@@ -171,6 +171,7 @@ def replace_revision(revision, user, comment, files, tags, initial_state=None):
     revision.comment = comment
     revision.tags = tags
     revision.final_state = FinalRevisionState.replaced
+    revision.editor = user
```
plourenco, line 174: Just a small highlight on this change. Even though it helps this PR identifying the "author" of a revision action, I think it makes sense for replaced revisions to be identified by the *editor* that decided to replace them.

indico/modules/events/editing/client/js/editing/timeline/RevisionLog.jsx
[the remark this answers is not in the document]
ThiefMaster: if you know which states you can get there, you can use `PropTypes.oneOf(['somestate', 'otherstate'])`

indico/modules/events/editing/operations.py
ThiefMaster: the user is pretty much always the indico system account, but no harm in setting it. at least i think so.. ;)

indico/modules/events/editing/client/js/editing/timeline/RevisionLog.jsx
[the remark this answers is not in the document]
plourenco: Never mind, I confused it with the diff above in `items`, since it can only be `blockItemPropTypes`.

indico/modules/events/editing/client/js/editing/timeline/selectors.js
```diff
@@ -8,135 +8,45 @@
...
-    (initialState.name === InitialRevisionState.ready_for_review &&
-      ![
-        FinalRevisionState.accepted,
-        FinalRevisionState.rejected,
-        FinalRevisionState.replaced,
-      ].includes(finalState.name))
-  );
+// This method defines each revision as a block
+// with a label referring to its previous state transition
+export function processRevisions(revisions) {
```
ThiefMaster, line 17: could you move this whole function into `util.js` since it's not a selector?

plourenco: I agree, always hated to see this here.

ThiefMaster (approved):

PR #4658, opened 2020-10-13
Send more verbose user data to editing service

ThiefMaster:


indico/modules/events/editing/schemas.py
```diff
@@ -383,6 +385,30 @@ class ReviewCommentSchema(mm.Schema):
...
+        return 'category' if isinstance(role, CategoryRole) else 'event'
+
+
+class ServiceUserSchema(mm.ModelSchema):
+    class Meta:
+        model = User
+        fields = ('id', 'full_name', 'email', 'roles', 'event_manager')
+
+    roles = fields.Method('_get_roles')
+    event_manager = fields.Function(lambda u, ctx: ctx['event'].can_manage(u, 'editing_manager'))
```
ThiefMaster, line 403: I'm including people who are "just" editing managers here, since the service is only editing-related so there should no need to differentiate between full event managers and editing managers.

plourenco (approved):

PR #4662, opened 2020-10-20
Fix an issue causing authorisation to fail in the legacy API

plourenco:
The categories endpoint in the legacy API fails while checking if the token is invalid.

The implementation should be replaced in Indico 3.0, but these changes ensure that it works as expected until.

### DIFF
--- a/indico/web/http_api/handlers.py
+++ b/indico/web/http_api/handlers.py
@@ -37,6 +37,7 @@ from indico.web.http_api.responses import HTTPAPIError, HTTPAPIResult
 from indico.web.http_api.util import get_query_parameter
 
 
+
 # Remove the extension at the end or before the querystring
 RE_REMOVE_EXTENSION = re.compile(r'\.(\w+)(?:$|(?=\?))')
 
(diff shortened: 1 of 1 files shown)


indico/web/http_api/handlers.py
```diff
@@ -130,19 +131,11 @@ def handler(prefix, path):
...
-            oauth_valid, oauth_request = oauth.verify_request([scope])
-            if not oauth_valid and oauth_request and oauth_request.error_message != 'Bearer token not found.':
+        oauth_valid, oauth_request = oauth.verify_request([scope])
+        if not oauth_valid and oauth_request:
+            if oauth_request.error_message != 'Bearer token not found.':
                 raise BadRequest('OAuth error: {}'.format(oauth_request.error_message))
-            elif g.get('received_oauth_token') and oauth_request.error_message == 'Bearer token not found.':
-                raise BadRequest('OAuth error: Invalid token')
-        except ValueError:
-            # XXX: Dirty hack to workaround a bug in flask-oauthlib that causes it
```
ThiefMaster, line 140: Are you sure removing this is safe?

indico/web/http_api/handlers.py
```diff
@@ -37,6 +37,7 @@
 from indico.web.http_api.util import get_query_parameter
 
 
+
```
ThiefMaster, line 40: 👾

plourenco:
A bug was found with potential data leakage between OAuth-authenticated and unauthenticated HTTP API requests for the same resource, causing this to be replaced by the fix above.

PR #4692, opened 2020-11-03
Hide title field by default & add Mx title

### LINKED ISSUE #4688: Update list of titles in Registration Form and disable it by default
Actions:
 * There is a gender-neutral non-academic title ("Mx") which seems to be the standard in several official documents and procedures in the UK these days. We have had requests from event organizers to include it in the default list and we believe this would be helpful to others;
 * Titles seem to be losing relevance in daily life. Although useful in some contexts, our experience with the overwhelming majority of Indico events is that only a few people end up filling them in and we almost always end up with a handful of those in a list of mostly "title-less" participants. The idea would be to have the "title" field disabled by default.

This field should remain fully configurable. New titles may be added and the field can even be made mandatory if needed.

ThiefMaster:
closes #4688

### DIFF
--- a/indico/modules/users/models/users.py
+++ b/indico/modules/users/models/users.py
@@ -38,13 +38,14 @@ from indico.web.flask.util import url_for
 
 
 class UserTitle(RichIntEnum):
-    __titles__ = ('', _('Mr'), _('Ms'), _('Mrs'), _('Dr'), _('Prof.'))
+    __titles__ = ('', _('Mr'), _('Ms'), _('Mrs'), _('Dr'), _('Prof.'), _('Mx'))
     none = 0
     mr = 1
     ms = 2
     mrs = 3
     dr = 4
     prof = 5
+    mx = 6
 
 
 class NameFormat(RichIntEnum):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,11 @@ Version 2.3.2
 
 *Unreleased*
 
+Improvements
+^^^^^^^^^^^^
+
+- Disable title field by default in new registration forms (:issue:`4688`, :pr:`4692`)
+
 Bugfixes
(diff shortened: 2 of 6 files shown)


plourenco (approved):
👍

indico/modules/users/models/users.py
```diff
@@ -38,13 +38,14 @@
...
 class UserTitle(RichIntEnum):
-    __titles__ = ('', _('Mr'), _('Ms'), _('Mrs'), _('Dr'), _('Prof.'))
+    __titles__ = ('', _('Mr'), _('Ms'), _('Mrs'), _('Dr'), _('Prof.'), _('Mx'))
     none = 0
     mr = 1
     ms = 2
     mrs = 3
     dr = 4
     prof = 5
+    mx = 6
```
Leats, line 48: Is it better to have it at the end of the list or would it be more inclusive to add it after Mrs? 
Semantically it would make more sense to add it after Mrs to keep the non-academic titles and academic titles separated.
I am not sure if it makes much difference, though.

ThiefMaster: Not possible without updating all existing values in the DB :/
But the `sorted=True` for the form field will use alphabetical order for the dropdown options, so the order is blank-dr-m*-prof which isn't perfect but better than having the non-academic titles separated ;)

ThiefMaster: actually, I think in python 3 this will be possible since the `__dict__` of a class is ordered there

```python
>>> class MyEnum(int, Enum):
...     a = 2
...     b = 1
...     c = 3
...

>>> list(MyEnum)
[<MyEnum.a: 2>, <MyEnum.b: 1>, <MyEnum.c: 3>]

```

Leats (approved):

PR #4694, opened 2020-11-06
Fix a bug causing system user checks for group membership

plourenco:
Fixes a bug causing the system user to be checked for group membership.

Introduced by #4657 

Can be reproduced by (within editing) accepting a revision where the previous revision was made by a system user. For example, when connected to the openreferee service.

### DIFF
--- a/indico/core/db/sqlalchemy/protection.py
+++ b/indico/core/db/sqlalchemy/protection.py
@@ -365,10 +365,13 @@ class ProtectionManagersMixin(ProtectionMixin):
         if permission is not None and permission != 'ANY' and permission not in get_available_permissions(type(self)):
             raise ValueError("permission '{}' is not valid for '{}' objects".format(permission, type(self).__name__))
 
-        if user is None or user.is_system:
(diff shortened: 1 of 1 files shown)


indico/core/db/sqlalchemy/protection.py
```diff
@@ -365,7 +365,7 @@ def can_manage(self, user, permission=None, allow_admin=True, check_parent=True,
         if permission is not None and permission != 'ANY' and permission not in get_available_permissions(type(self)):
             raise ValueError("permission '{}' is not valid for '{}' objects".format(permission, type(self).__name__))
 
-        if user is None:
+        if user is None or user.is_system:
```
ThiefMaster, line 368: Maybe a separate `if` with a comment explaining that the system user has no email and thus some access checks (against groups) may fail, and the system user never needs special access, would be better?

plourenco: Sounds better, yes

ThiefMaster (approved):

PR #4695, opened 2020-11-06
Python 3... LET'S DO THIS! 🐍

### LINKED ISSUE #4673: Migrate to Python 3.9
We can look into `modernize` and `pyupgrade` to automate some of the things.

ThiefMaster:
Everything seems to work fine on 3.9! 🎉

If anyone wants to look at the changes I strongly recommend going through individual commits since some are large ones from `modernize` and `pyupgrade`.

closes #4673


bpedersen2:
bin/maintenance/dump-url-mpa misses the 

#! /usr/bin/env python line  

( this causes problems if the default python  is not the  correct one...)

ThiefMaster:
@bpedersen2 that file is not meant to be executed directly but called via `python bin/maintenance/dump_url_map.py`. however, i agree that it'd be nice if we were more consistent in adding shebang lines instead having them in some commandline tools but not in others.

bpedersen2:
> @bpedersen2 that file is not meant to be executed directly but called via `python bin/maintenance/dump_url_map.py` (it wouldn't work regardless of shebang since it's not marked as executable).
> 
> however, i agree that it'd be nice if we were more consistent in adding shebang lines instead having them in some commandline tools but not in others.

Yes, and that is where this breaks ( manually calling it in the venv works, but  broke from the other scripts ( they could also use sys.executable instead of  a hardcoded 'python' )

ThiefMaster:
but since we only support virtualenvs, i'd say this point is moot ;) do you have any example of a script calling this in a broken way? during development we assume people always have the virtualenv active

bpedersen2:
> but since we only support virtualenvs, i'd say this point is moot ;) do you have any example of a script calling this in a broken way? during development we assume people always have the virtualenv active

Yes, currently I can not reproduce the problem (also I am sure I had the venv activated then).

PR #4697, opened 2020-11-11
Some changes needed for Zoom plugin

pferreir:
This adds:
 * Possibility to disable WTFPrincipalField;
 * Template hook to add additional interface items in list of VC rooms;
 * New `is_new` flag in `update_data_vc_room`.

### DIFF
--- a/indico/modules/vc/plugins.py
+++ b/indico/modules/vc/plugins.py
@@ -185,7 +185,7 @@ class VCPluginMixin(object):
         if event_vc_room.data is None:
             event_vc_room.data = {}
 
-    def update_data_vc_room(self, vc_room, data):
+    def update_data_vc_room(self, vc_room, data, is_new=False):
         if 'name' in data:
             vc_room.name = data.pop('name')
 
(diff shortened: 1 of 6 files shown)


indico/web/templates/forms/principal_widget.html
```diff
@@ -13,7 +13,8 @@
         setupPrincipalWidget({
             fieldId: {{ field.id | tojson }},
             required: {{ input_args.required | default(false) | tojson }},
-            withExternalUsers: {{ field.allow_external_users | tojson }}
+            withExternalUsers: {{ field.allow_external_users | tojson }},
+            disabled: {{ field.render_kw.disabled if field.render_kw else None | tojson }}
```
ThiefMaster, line 17:
```suggestion
            disabled: {{ (field.render_kw.disabled if field.render_kw else none) | tojson }}
```

indico/modules/vc/plugins.py
```diff
@@ -185,7 +185,7 @@ def update_data_association(self, event, vc_room, event_vc_room, data):
         if event_vc_room.data is None:
             event_vc_room.data = {}
 
-    def update_data_vc_room(self, vc_room, data):
+    def update_data_vc_room(self, vc_room, data, is_new=False):
```
ThiefMaster, line 188: :warning: This requires a change in the `vc_vidyo` plugin as well, unless we call it with `is_new=True` only if the method accepts this argument.

pferreir: Indeed!

pferreir: Solved, thanks!

PR #4705, opened 2020-11-13
Room booking: fix queries in conflict checking

pferreir:
Small fix in conflict checks. Cancelled/rejected bookings were being taken into account.


pferreir:
This solves the wrong problem. Closing. See #4719

PR #4706, opened 2020-11-13
Check for admin override in blockings

ThiefMaster:


indico/modules/rb/models/blockings.py
```diff
@@ -109,7 +109,7 @@ def can_override(self, user, room=None, explicit_only=False, allow_admin=True):
         if not explicit_only:
             if allow_admin and rb_is_admin(user):
                 return True
-            if room and room.can_manage(user):
+            if room and room.can_manage(user, allow_admin=allow_admin):
```
plourenco, line 112: Would it make sense to still use the `allow_admin` here?
From what I read here, if `allow_admin` is True and the user is actually a room booking admin, it will early return true. Thus, causing this to have no effect.

Is there any occasion where `allow_admin` is still useful even if the user isn't one?

indico/modules/rb/models/reservations.py
```diff
@@ -349,7 +349,7 @@ def create_from_data(cls, room, data, user, prebook=None, ignore_admin=False):
         reservation.booked_for_name = reservation.booked_for_user.full_name
         reservation.state = ReservationState.pending if prebook else ReservationState.accepted
         reservation.created_by_user = user
-        reservation.create_occurrences(True)
+        reservation.create_occurrences(True, allow_admin=(not ignore_admin))
```
plourenco, line 352: Completely unrelated, but I just noticed that `ignore_admin` is missing in the docstring, perhaps we could add it with these changes?

indico/modules/rb/models/reservations.py
```diff
@@ -492,15 +492,15 @@ def can_delete(self, user, allow_admin=True):
...
+    def create_occurrences(self, skip_conflicts, user=None, allow_admin=True):
         ReservationOccurrence.create_series_for_reservation(self)
         db.session.flush()
 
         if user is None:
             user = self.created_by_user
 
         # Check for conflicts with nonbookable periods
-        if not rb_is_admin(user) and not self.room.can_manage(user, permission='override'):
+        if (not allow_admin or not rb_is_admin(user)) and not self.room.can_manage(user, permission='override'):
```
plourenco, line 503: What do you think about moving this part to simplify the condition?
(Can't think of a good var name though 😄)
```suggestion
        admin = allow_admin and rb_is_admin(user)
        if not admin and not self.room.can_manage(user, permission='override'):
```

indico/modules/rb/models/blockings.py
ThiefMaster: `allow_admin` indicates whether admin permissions should be checked at all. so if it's false, we do not return early above, so we don't want the `can_manage` check to return early either

indico/modules/rb/models/reservations.py
ThiefMaster: no strong opinion on it, so why not ;)

PR #4708, opened 2020-11-17
Add revision numbers to each timeline block

plourenco:
Regarding revision blocks, we previously only appended the revision reference when a custom header wasn't set.

This change converts a revision title such as "Revision as been replaced" to "Revision has been replaced #\2".

Will depend on #4709.

### DIFF
--- a/indico/modules/events/editing/client/js/editing/timeline/TimelineItem.jsx
+++ b/indico/modules/events/editing/client/js/editing/timeline/TimelineItem.jsx
@@ -57,7 +57,7 @@ export default function TimelineItem({block, index}) {
           >
             <div className="i-box-header flexrow">
               <div className="f-self-stretch" styleName="header">
-                <span styleName="item-index">#{index + 1}</span>
+                <span styleName="item-index">#{index + 1}</span>{' '}
                 <span>
                   {block.header ? (
                     <strong>{block.header}</strong>
@@ -67,7 +67,7 @@ export default function TimelineItem({block, index}) {
                       submitted a revision
                     </Translate>
                   )}
-                </span>
+                </span>{' '}
                 <time dateTime={serializeDate(createdDt, moment.HTML5_FMT.DATETIME_LOCAL)}>
                   {serializeDate(createdDt, 'LL')}
                 </time>
--- a/indico/modules/events/editing/client/js/editing/timeline/TimelineItem.jsx
+++ b/indico/modules/events/editing/client/js/editing/timeline/TimelineItem.jsx
@@ -56,17 +56,18 @@ export default function TimelineItem({block, index}) {
             id={`block-info-${block.id}`}
           >
             <div className="i-box-header flexrow">
-              <div className="f-self-stretch">
-                {block.header ? (
-                  <span>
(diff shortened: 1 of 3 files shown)


indico/modules/events/editing/client/js/editing/timeline/TimelineItem.jsx
```diff
@@ -58,7 +58,9 @@ export default function TimelineItem({block, index}) {
             <div className="i-box-header flexrow">
               <div className="f-self-stretch">
                 {block.header ? (
-                  <strong>{block.header}</strong>
+                  <span>
+                    <strong>{block.header}</strong> {`#${index + 1}`}
```
ThiefMaster, line 62:
```suggestion
                    <strong>{block.header}</strong> #{index + 1}
```

ThiefMaster:
image

I don't really like the way it looks right now. Maybe we could show it in a nicer way at the beginning of the line, similar to how we show the date at the end of it?

plourenco:
> I don't really like the way it looks right now. Maybe we could show it in a nicer way at the beginning of the line, similar to how we show the date at the end of it?

I agree, it looks a bit weird, the problem is moving it at the beginning of the line doesn't play very well with the first line.
Maybe something like this?
imagem

ThiefMaster:
Not bad :) Alternatively, maybe something like `#1 * Pedto submitted initial revision * November 12, 2020`?

other opinions @indico/development-team?

plourenco:
> Not bad (maybe put a separator dot before the ID like we have before the date ?) :)
> 
> Alternatively, maybe something like `#1 * Pedto submitted initial revision * November 12, 2020`?
> 
> other opinions @indico/development-team?

I like both alternatives! The only inconvenient with the second one is that "Name submitted revision #1" is the header we use for every revision that doesn't have one. So, unless we add an exception for the first one, ideally it would be something like `#1 * Name submitted a revision * November 12, 2020`.

indico/modules/events/editing/client/js/editing/timeline/TimelineItem.module.scss
```diff
@@ -5,6 +5,18 @@
...
   margin-left: 5px;
 }
+
+.item-index {
+  color: $gray;
+}
+
+.header > :not(:first-child)::before {
+  content: '\00B7' !important;
+  padding: 0 0.4em;
```
plourenco, line 20: I updated this css to provide some padding instead of a special character plus the `{' '}`.

I could have updated it globally, we use it on `_timelines.scss` however I'm not exactly sure of the impact, we would have to update some code and remove the extra spaces.

ThiefMaster: What I don't like with this approach is that you get something like this when copying content from the page (at least in Firefox):

    #1Adrian Mönnich submitted a revisionNovember 12, 2019

Otherwise it'd be a great option!

plourenco: To be completely honest, I would probably still prefer not having the spaces, but I can see it makes sense for some use-cases.

ThiefMaster (approved):

PR #4709, opened 2020-11-17
Update deprecated set-env and add-path commands

plourenco:
Update `set-env` and `add-path` commands that are being deprecated according to [changelog](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,8 +89,8 @@ jobs:
 
       - name: Activate virtualenv for later steps
         run: |
-          echo "::set-env name=VIRTUAL_ENV::$(pwd)/.venv"
-          echo "::add-path::$(pwd)/.venv/bin"
+          echo "VIRTUAL_ENV=$(pwd)/.venv" >> $GITHUB_ENV
+          echo "$(pwd)/.venv/bin" >> $GITHUB_PATH
 
       - name: Install Indico
         run: pip install -e .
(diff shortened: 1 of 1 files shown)


ThiefMaster:
huh where did someone fix this is recently? was it in newdle? i could've sworn this was already done
..

plourenco:
> huh where did someone fix this is recently? was it in newdle? i could've sworn this was already done
> ..

Yes, that's right, it was introduced in indico/newdle@697bb0c. Didn't notice this was changed until my latest run.

.github/workflows/ci.yml
```diff
@@ -85,8 +85,8 @@ jobs:
 
       - name: Activate virtualenv for later steps
         run: |
-          echo "::set-env name=VIRTUAL_ENV::$(pwd)/.venv"
-          echo "::add-path::$(pwd)/.venv/bin"
+          echo "name=VIRTUAL_ENV::$(pwd)/.venv" >> $GITHUB_ENV
```
ThiefMaster, line 88:
```suggestion
          echo "VIRTUAL_ENV=$(pwd)/.venv" >> $GITHUB_ENV
```

same below

ThiefMaster:
Could you change the PR to be on top of 2.3-maintenance? That branch also uses GH actions so I'd rather merge the PR in there and then that branch into master.

.github/workflows/ci.yml
plourenco: My bad, my latest force push was supposed to fix this.

ThiefMaster:
you'll also have to rebase the branch itself from master onto 2.3-maintenance ;)

PR #4712, opened 2020-11-18
Contribution list: Filter speakers correctly

bpedersen2:
Do not attempt to match on all ContributionPersonLinks, only on
'speakers'.

Otherwise a contribution with a second, non-speaker author that has
registered will not show up if searching for non-registered speakers.


ThiefMaster (approved):
Nice catch, thanks for fixing this :)

PR #4713, opened 2020-11-18
Add codes in the contribution list

plourenco:
Includes the codes in the contribution lists and a `data-searchable` tag.
I've decided not to opt for a label (like we have in the editing view) since this one looks simpler. However, depending on the feedback we can change it either way.
imagem

### DIFF
--- a/indico/web/client/styles/modules/event_display/_conferences.scss
+++ b/indico/web/client/styles/modules/event_display/_conferences.scss
@@ -195,6 +195,11 @@
       font-size: 1.2em;
     }
 
+    .contrib-code {
+      color: $dark-gray;
+      font-size: 0.8em;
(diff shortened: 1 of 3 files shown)


indico/web/client/styles/modules/event_display/_conferences.scss
```diff
@@ -195,6 +195,11 @@
       font-size: 1.2em;
     }
 
+    .contrib-code {
+      color: #999;
```
ThiefMaster, line 199:
```suggestion
      color: $dark-gray;
```

ThiefMaster:
I think this is stuff that can still go in 2.3-maintenance

ThiefMaster (approved):

PR #4714, opened 2020-11-19
Update marshmallow & webargs

### LINKED ISSUE #4682: Update webargs & marshmallow
6.x has [breaking changes](https://webargs.readthedocs.io/en/latest/changelog.html#b1-2020-01-06) that require updating every single call of `use_args` and `use_kwargs`, but since we consistently use only query strings are or POST data for pretty much all the places where we use it, the change should be straightforward and require only little manual testing.

There's also a 7 (beta) release of webargs. Maybe we can go directly for that one since it's still some time until the indico 3 release...

ThiefMaster:
closes #4682


ThiefMaster:
Resolving principals (and other fields that use a custom list-like field not inheriting from the List field) currently only works with a JSON payload (and not with regular form data) but that's what we use anyway, so not a blocker, could even set `location='json'` there...

https://github.com/marshmallow-code/webargs/issues/563 for details.

indico/modules/events/abstracts/schemas.py
```diff
@@ -60,19 +60,19 @@ class Meta:
...
 
 
-class AbstractSchema(mm.ModelSchema):
+class AbstractSchema(mm.SQLAlchemyAutoSchema):
     submitter = Nested(UserSchema)
     judge = Nested(UserSchema)
     modified_by = Nested(UserSchema)
-    duplicate_of = Nested('self', only=_basic_abstract_fields)
-    merged_into = Nested('self', only=_basic_abstract_fields)
+    duplicate_of = Nested(lambda: AbstractSchema, only=_basic_abstract_fields)
```
plourenco, line 74: I was just reading that `fields.Nested` can now take a callable (such as, `Nested(lambda: AbstractSchema(only=)`. But isn't this still the same thing?
```suggestion
    duplicate_of = Nested(AbstractSchema, only=_basic_abstract_fields)
```

ThiefMaster: No, a class cannot reference itself during definition time. That's why they added support for passing a callable instead of the schema itself.

plourenco: Ah, I see, this is a particular case!

PR #4715, opened 2020-11-20
Fix existing files not conforming with name patterns

plourenco:


indico/modules/events/editing/client/js/editing/timeline/FileManager/util.js
```diff
@@ -108,11 +109,18 @@ export function mapFileTypes(fileTypes, files, uploadableFiles = []) {
...
     invalidFiles: [],
-    uploadableFiles: uploadableFiles.filter(
+    uploadableFiles: filesForFileType(uploadableFiles, fileType),
+  }));
+}
+
+function filesForFileType(files, fileType) {
+  const templateRe = fileType.filenameTemplate && globToRegExp(fileType.filenameTemplate);
+  return files
+    .filter(file => !fileType.filenameTemplate || templateRe.test(file.filename.split('.')[0]))
```
ThiefMaster, line 119: Does this work as expected if the filename itself contains any dots?

plourenco: That's a good point. I had the impression we weren't going to consider them as the filename templates also made the exact same check. But now it's fixed on both sides. 👍

ThiefMaster (approved):

PR #4718, opened 2020-11-20
Expose links the user can access

### LINKED ISSUE #2599: Link material - make URL visible to the user
As a user, how can I know what is the URL behind a link material? When I hover my mouse over the link, the popup just repeats the link material's _name_ (not very useful), and the browser's status bar shows the URL of Indico's redirection, but not the final URL. 

image

For usability, but also security reasons, I believe it would be better if the popup contains the final URL instead of the link's name. (BTW, wasn't that the case some time ago?)

plourenco:
Closes #2599

Uses the simplistic approach where links are visible if the user has permissions on them. It causes an access check on our side for every button, but realistically I'm assuming no impact while doing that.

We also considered creating an Indico redirector (could even contain a signature to avoid malicious access) but concluded it could become a burden for a simple use-case.

View with permissions (the redirection is immediate):
imagem

View without permission (the redirection goes through the server):
imagem

### DIFF
--- a/indico/modules/attachments/templates/_attachments.html
+++ b/indico/modules/attachments/templates/_attachments.html
@@ -8,8 +8,7 @@
 
 {% macro render_attachment(attachment, has_label=false, classes='') %}
     {% set previewable = (attachment.file.is_previewable and not g.static_site) or false %}
-    {% set restricted = attachment.is_self_protected and not attachment.can_access(session.user) %}
-    {% set visible_link = attachment.type.name == 'link' and not restricted %}
+    {% set visible_link = attachment.type.name == 'link' and (not attachment.is_self_protected or attachment.can_access(session.user)) %}
     <a class="attachment {{ get_attachment_icon(attachment) }} {% if attachment.is_self_protected %}is-protected{% endif %} {% if has_label %}has-label{% endif %} {%- if previewable %} js-preview-dialog{% endif %} {{ classes }}"
        data-previewable="{{ previewable|tojson }}"
        data-attachment-id="{{ attachment.id }}"
@@ -17,9 +16,7 @@
        title="{{ attachment.description or attachment.title }}
               {%- if attachment.is_self_protected %} ({% trans %}Protected{% endtrans %}){% endif %}">
         {%- if not has_label -%}
-            <span class="title">
-                {{ attachment.link_url if visible_link else attachment.title }}
-            </span>
+            <span class="title">{{ attachment.title }}</span>
         {%- endif -%}
     </a>
 {% endmacro %}
--- a/indico/modules/attachments/templates/_attachments.html
+++ b/indico/modules/attachments/templates/_attachments.html
@@ -8,14 +8,18 @@
 
 {% macro render_attachment(attachment, has_label=false, classes='') %}
     {% set previewable = (attachment.file.is_previewable and not g.static_site) or false %}
+    {% set restricted = attachment.is_self_protected and not attachment.can_access(session.user) %}
+    {% set visible_link = attachment.type.name == 'link' and not restricted %}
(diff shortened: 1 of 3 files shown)


indico/modules/attachments/templates/_attachments.html
```diff
@@ -8,14 +8,18 @@
 
 {% macro render_attachment(attachment, has_label=false, classes='') %}
     {% set previewable = (attachment.file.is_previewable and not g.static_site) or false %}
+    {% set restricted = attachment.is_self_protected and not attachment.can_access(session.user) %}
```
ThiefMaster, line 11: let's not perform the access check unless we need it (ie it's actually a link attachment)

indico/modules/attachments/templates/_attachments.html
```diff
@@ -8,14 +8,18 @@
...
        data-previewable="{{ previewable|tojson }}"
        data-attachment-id="{{ attachment.id }}"
-       href="{{ attachment.download_url }}" target="_blank"
+       href="{{ attachment.link_url if visible_link else attachment.download_url }}" target="_blank"
        title="{{ attachment.description or attachment.title }}
               {%- if attachment.is_self_protected %} ({% trans %}Protected{% endtrans %}){% endif %}">
         {%- if not has_label -%}
-            <span class="title">{{ attachment.title }}</span>
+            <span class="title">
+                {{ attachment.link_url if visible_link else attachment.title }}
```
ThiefMaster, line 21: i don't think that's a good idea - the link may be awfully long, and we want to keep the title. people can copy the URL or check the status bar if they want to know where the link goes IMO

indico/modules/attachments/templates/_attachments.html
plourenco: I moved these up because it was getting a bit confusing, but that's probably better.

indico/modules/attachments/templates/_attachments.html
plourenco: Alright, I added after testing manually and noticing we already have some pretty nice truncation (could be more user friendly):
imagem
But I'm guessing this steals the purpose of link titles.

indico/modules/attachments/templates/_attachments.html
ThiefMaster: that makes sense, I'd just merge this into the `set visible_link` statement so we short-circuit in case it's not a link

indico/modules/attachments/templates/_attachments.html
ThiefMaster: exactly ;) and it would probably look worse in meetings where the attachments are shown more like buttons

indico/modules/attachments/templates/_attachments.html
```diff
@@ -8,10 +8,11 @@
 
 {% macro render_attachment(attachment, has_label=false, classes='') %}
     {% set previewable = (attachment.file.is_previewable and not g.static_site) or false %}
+    {% set visible_link = attachment.type.name == 'link' and (not attachment.is_self_protected or attachment.can_access(session.user)) %}
```
ThiefMaster, line 11:
```suggestion
    {% set visible_link = attachment.type.name == 'link' and (not attachment.is_protected or attachment.can_access(session.user)) %}
```

otherwise you reveal links in protected folders

indico/modules/attachments/templates/_attachments.html
```diff
@@ -8,10 +8,11 @@
...
     <a class="attachment {{ get_attachment_icon(attachment) }} {% if attachment.is_self_protected %}is-protected{% endif %} {% if has_label %}has-label{% endif %} {%- if previewable %} js-preview-dialog{% endif %} {{ classes }}"
        data-previewable="{{ previewable|tojson }}"
        data-attachment-id="{{ attachment.id }}"
-       href="{{ attachment.download_url }}" target="_blank"
+       href="{{ attachment.link_url if visible_link else attachment.download_url }}" target="_blank"
```
ThiefMaster, line 15: Maybe this would be a good time to start adding `rel="noopener noreferrer"` to those links

PR #4721, opened 2020-11-23
Fix too-restricted access check for attachments

ThiefMaster:
When an attachment was inside a public session/contribution of a protected event, we always checked for event access even though having something less protected inside an event is possible and may be used in case of "public sessions" of otherwise protected events.

Also added a bunch of unit tests covering this.


plourenco (approved):
Nice one!
Even though I'm a bit concerned about how well we can scale this (keep using `except` to redirect to the correct exception). Now I see why less protected childs are a mess.

I guess it wouldn't make much sense to mix the logics/singularity and have the `RHDisplayEventBase._check_access(self)` check within `DownloadAttachmentMixin._check_access(self)`...?

PR #4725, opened 2020-11-26
Allow categories to have both events and subcats

### LINKED ISSUE #4679: Allow mixing events + subcategories in a category
This is fully supported in the backend and the frontend does not look awful. We just prohibit it when creating new events/categories.

Removing those restrictions and ensuring (especially in the category picker) that a case of a category containing both is properly handled (we may even be able to remove some code there related to restricting things based on what a category contains!) should be enough.

Once we did this, we can finally cache certain things such as category event counts, instead of querying freshly from the DB every single time (ideally in a separate task).

plourenco:
Closes #4679

### DIFF
--- a/indico/modules/categories/client/js/management.js
+++ b/indico/modules/categories/client/js/management.js
@@ -340,7 +340,6 @@
         )
         .format(ids.length),
       actionOn: {
-        categoriesWithEvents: {disabled: true},
         categoriesDescendingFrom: {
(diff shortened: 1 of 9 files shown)


ThiefMaster:
I just cherrypicked the eslint commit into master; could you rebase?

ThiefMaster (approved):

PR #4731, opened 2020-11-30
Exclude some user agents from sso redirection

### LINKED ISSUE #4720: SSO Authentication is incorrectly redirected within MS Office
Opening a restricted event from within a link in MS Office (with SSO) will cause a failed login.

imagem

## The cause

MS Office seems to resolve redirects first before opening the URL in the browser, so it opens the SSO site in the browser, and thus there are no Indico cookies before coming back from SSO.

https://docs.microsoft.com/en-us/office/troubleshoot/office-suite-issues/click-hyperlink-to-sso-website

## Solutions

A possible workaround could be checking for whatever user agent Office sends and in that case disable the automatic `single_auth_provider` redirect in `RHLogin`.

plourenco:
Closes #4720

### DIFF
--- a/indico/modules/auth/controllers.py
+++ b/indico/modules/auth/controllers.py
@@ -65,9 +65,14 @@ class RHLogin(RH):
             multipass.set_next_url()
             return multipass.redirect_success()
 
+        # Some clients attempt to incorrectly resolve redirections internally.
+        # See https://github.com/indico/indico/issues/4720 for details
+        user_agent = request.headers.get('User-Agent', '')
+        sso_redirect = not any(s in user_agent for s in ('ms-office', 'Microsoft Office'))
+
         # If we have only one provider, and this provider is external, we go there immediately
         # However, after a failed login we need to show the page to avoid a redirect loop
-        if not session.pop('_multipass_auth_failed', False) and 'provider' not in request.view_args:
+        if not session.pop('_multipass_auth_failed', False) and 'provider' not in request.view_args and sso_redirect:
             single_auth_provider = multipass.single_auth_provider
             if single_auth_provider and single_auth_provider.is_external:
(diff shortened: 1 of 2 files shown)


indico/modules/auth/controllers.py
```diff
@@ -65,9 +66,13 @@ def _process(self):
             multipass.set_next_url()
             return multipass.redirect_success()
 
+        sso_redirect = True
+        if 'User-Agent' in request.headers:
+            sso_redirect = all(a not in request.headers['User-Agent'] for a in self.REDIRECT_EXEMPT)
```
ThiefMaster, line 71: `not any(...)`

indico/modules/auth/controllers.py
```diff
@@ -56,6 +56,7 @@ class RHLogin(RH):
     # CSRF token as Indico forms use. But since the form has its own
     # CSRF check anyway disabling the global check is perfectly fine.
     CSRF_ENABLED = False
+    REDIRECT_EXEMPT = ["ms-office", "Microsoft Office"]
```
ThiefMaster, line 59: I'd put an empty line before this (so it's separated from the documented csrf block ) or even just keep the list below inside the method since it's never meant to be overridden anyway

indico/modules/auth/controllers.py
plourenco: I started with `not any` too 😄, but it boils down to the same right?

plourenco: Never mind, I've just re-read what I wrote shortly after my previous comment.

ThiefMaster (approved):
I can't reproduce it on my local instance (MS Office only sends HTTP HEAD which 400s anyway), but the logic works fine so I don't see any reason to not merge this.

PR #4735, opened 2020-12-03
Add Last Login Date block to dashboard.html

vasantvohra:
Need to add an empty block to the user dashboard template for displaying the last login date from the UN template override.


indico/modules/users/templates/dashboard.html
```diff
@@ -49,6 +49,7 @@ <h3>{{ user.full_name }}</h3>
                                 </div>
                             {% endif %}
                             {{ labels(user) }}
+                            {% block last_login_dt %} {% endblock %}
```
ThiefMaster, line 52: The name is too specific. Please find a generic name. `after_user_labels` maybe. Also, no need to put a space in the default content.

ThiefMaster (changes requested):
Please don't keep your JIRA IDs in commit messages going upstream and use present tense :)

`Add xxx block in dashboard template` for example

ThiefMaster:
Showing the last login date may be useful in the core. Are you using the highest `safe_last_login_dt` of all the user's identities for this? If yes, maybe contributing the actual feature of showing the last login timestamp upstream would be better than just adding the block for you to override?

vasantvohra:
> Showing the last login date may be useful in the core. Are you using the highest `safe_last_login_dt` of all the user's identities for this? If yes, maybe contributing the actual feature of showing the last login timestamp upstream would be better than just adding the block for you to override?

@ThiefMaster Yes, I'm using the `user.local_identity.safe_last_login_dt` with format_datetime function in the template, I guess the user.local_identity gives the most recently used local identity. Correct me, if I'm wrong. Also, In my UN Jira ticket, we are showing it only when `{% if session.user.is_admin %}` does the core wants this feature for all the users?
Apologize for the above silly mistakes just had an intro session from @OmeGak :)

ThiefMaster:
Since you can only view your own dashboard unless you are an admin, I think showing it all the time would be fine. I don't have objections in showing it only to admins there, since we eventually want to add a nice sessions overview for users as well where we could show it to the user themself.

vasantvohra:
> since we eventually want to add a nice sessions overview for users as well where we could show it to the user themself.

Sessions overview will be good let me know the features you are thinking to add there. Also, I think we can show users last login location from the IP we are storing in identities. For the time being, I've committed the entire block, please have a review and let me know if any changes required.

indico/modules/users/templates/dashboard.html
```diff
@@ -49,6 +49,14 @@ <h3>{{ user.full_name }}</h3>
                                 </div>
                             {% endif %}
                             {{ labels(user) }}
+                            {% block session_details %}
+                            <br>
+                            <div class="ui small label">
+                                Last Login Date
```
ThiefMaster, line 55:
```suggestion
                                {% trans %}Last login date{% endtrans %}
```

indico/modules/users/templates/dashboard.html
```diff
@@ -49,6 +49,14 @@ <h3>{{ user.full_name }}</h3>
                                 </div>
                             {% endif %}
                             {{ labels(user) }}
+                            {% block session_details %}
```
ThiefMaster, line 52: Unless you still need to override it, let's get rid of the block around it.

indico/modules/users/templates/dashboard.html
```diff
@@ -49,6 +49,14 @@ <h3>{{ user.full_name }}</h3>
...
+                            <br>
+                            <div class="ui small label">
+                                Last Login Date
+                                {% set last_login_dt = user.local_identity.safe_last_login_dt %}
+                                <div class="detail">{{ last_login_dt | format_datetime  }}</div>
```
ThiefMaster, line 57: This will fail if the user has no `local_identity`.

indico/modules/users/templates/dashboard.html
```diff
@@ -49,6 +49,17 @@ <h3>{{ user.full_name }}</h3>
...
                             {% endif %}
                             {{ labels(user) }}
+                            <br>
+                            <div class="ui small label">
+                                {% trans %}Last login date{% endtrans %}
+                                <div class="detail">
+                                {% if user.last_login_dt %}
+                                    {{ user.last_login_dt | format_datetime  }}
+                                {% else %}
+                                    Never
```
ThiefMaster, line 59: Wrap this in trans as well for i18n

indico/modules/users/templates/dashboard.html
```diff
@@ -49,6 +49,17 @@ <h3>{{ user.full_name }}</h3>
                                 </div>
                             {% endif %}
                             {{ labels(user) }}
+                            <br>
+                            <div class="ui small label">
+                                {% trans %}Last login date{% endtrans %}
+                                <div class="detail">
+                                {% if user.last_login_dt %}
+                                    {{ user.last_login_dt | format_datetime  }}
```
ThiefMaster, line 57:
```suggestion
                                    {{ user.last_login_dt | format_datetime }}
```

indico/modules/users/templates/dashboard.html
```diff
@@ -49,6 +49,17 @@ <h3>{{ user.full_name }}</h3>
                                 </div>
                             {% endif %}
                             {{ labels(user) }}
+                            <br>
+                            <div class="ui small label">
+                                {% trans %}Last login date{% endtrans %}
+                                <div class="detail">
```
ThiefMaster, line 55: the content of this div should be indented

vasantvohra: Did you mean for `<div class="detail">`?

ThiefMaster: yes, anything that's in a nested element (regardless if html or jinja) gets a new level of indentation

vasantvohra: I've intended the if conditions in div element, is it correct?

ThiefMaster: yep

PR #4067, opened 2019-09-16
Search Interface 2.0

aristofanischionis:
# Search Component
## General Description :
* Filter results in 4 types: 
   * Categories
   * Contributions
   * Events
   * Files
* Paginate results
* URL Manipulation
* Icons and nice way of presenting the results
* API to fetch results

## Issues:
* Back button doesn't update the search term of the Search Box.

## Screenshots:
Screenshot from 2019-09-17 11-11-43
Screenshot from 2019-09-17 11-11-29
Screenshot from 2019-09-17 11-11-13
Screenshot from 2019-09-17 11-10-54

PR #4736, opened 2020-12-03
New search interface

ThiefMaster:
For now this draft PR just contains a squashed commit of all the changes in #4067 since that PR got massive over time.

Things to do (most likely this list is non-exhaustive):

- [ ] Define a new search API to integrate with plugins (especially citadel - the livesync part is in indico/indico-plugins#86)
- [ ] Address the comments in the original PR.
- [ ] Possibly polish the UI (for later, since functionality is more important)

(please add new things up there as they come up)

Let's use this branch as a starting point for additional development regarding the new search in the core, so we have one PR/branch where everything gets consolidated.

Some ideas for the plugin API:

- Let's not create a new plugin class mixin unless we realize there's a strong need for this.
- No need to strongly enforce a single search engine - that way we can most likely just use a signal to trigger get a list of search implementations (like we do with other things like placeholders being provided by a plugin). And if we do want to keep enforcing a single active engine, we can simply fail when the signal call returns more than one search implementation.
- For the API we can look at the `search()` util in Michal's PR. This still needs tweaking of course, but should be a good start. We surely need to pass the user (and possibly a list of all their (multipass-based) groups) as well.
- For the search counterpart to `livesync_citadel` we could either put it in the same plugin (maybe easier, and allows reusing utils if needed) or create a separate `search_citadel` plugin. I'm more inclined to keep both parts (livesync + search) in a single plugin, and rename it to `search_citadel` in the end. It's very uncommon to use only one part, and in case of e.g. dev/staging instances we'd simply disable the livesync agent and only use the search part.
- It may be a good idea to create a `search_dummy` plugin that simply implements the new search API so we have something basic to test with, without depending on Citadel being fully functional. That also allows us to move out the demo data that's currently hardcoded in `indico/modules/search/controllers.py` out of the core. Ideally we can do this 
- The search UI supports applying extra filters to the current search results (by speaker affiliation / name) which are based on the data in those results, and not subject to the usual pagination. I think at the beginning we can ignore this to keep things easier, but eventually this need to be handled in the API between core and plugin as well

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -28,6 +28,7 @@ settings:
         - ['indico/modules/core', './indico/modules/core/client/js']
         - ['indico/modules/rb', './indico/modules/rb/client/js']
         - ['indico/modules/users', './indico/modules/users/client/js']
+        - ['indico/modules/search', './indico/modules/search/client/js']
         - ['indico/modules/events/reviewing', './indico/modules/events/client/js/reviewing']
         - ['indico/modules/events/editing', './indico/modules/events/editing/client/js']
         - ['indico/modules/events', './indico/modules/events/client/js']
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,7 @@ jobs:
              indico/modules/events/papers/client/js/
              indico/web/client/js/react/
              indico/modules/users/
+             indico/modules/search/
 
       - name: Run stylelint for all files
         if: github.event_name == 'push' && (success() || failure())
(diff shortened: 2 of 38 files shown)


plourenco:
Maybe we could merge this into a separate new branch, so all PRs and diffs would be based and tracked from Indico repo itself?

ThiefMaster:
the idea was to just push directly to my branch and not have individual PRs against that branch ;)

ThiefMaster:
This is in the upstream `search` branch now; opening a new PR for that one.

PR #4737, opened 2020-12-04
change signals.category.updated name to 'updated'

giusedb:
As per the discussion on https://talk.getindico.io/t/misleading-category-created-signal/2048
I propose to change the name of the `NamedSignal` to `updated` accordingly with the variable name.

### DIFF
--- a/indico/core/signals/category.py
+++ b/indico/core/signals/category.py
@@ -21,7 +21,7 @@ created = _signals.signal('created', """
 Called when a new category is created. The `sender` is the new category.
 """)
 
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Thanks, but this is already fixed in 2.3-maintenance (3cd07d2e1e). I just haven't merged that branch it into master yet since you guys are based on 2.3-maintenance anyway (unless you ended up switching to Python 3 during the last few days) ;)

giusedb:
Sorry, I didn't see it

PR #4738, opened 2020-12-04
Upgrade to webpack 5

### LINKED ISSUE #4675: Webpack 5
It has some very interesting features like module federation. But the painful part will be upgrading our existing webpack build configs to Webpack 5 first.

plourenco:
Closes #4675

### DIFF
--- a/indico/web/client/js/react/components/index.js
+++ b/indico/web/client/js/react/components/index.js
@@ -5,6 +5,14 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
+export {
+  FinalPrincipal,
+  FinalPrincipalList,
+  PrincipalField,
+  PrincipalListField,
+  ACLField,
+  TrackACLField,
+} from './principals';
 export {default as IButton} from './IButton';
 export {default as MessageBox} from './MessageBox';
 export {default as Modal} from './Modal';
@@ -19,14 +27,6 @@ export {default as Carousel} from './Carousel';
 export {default as ScrollButton} from './ScrollButton';
 export {default as StickyWithScrollBack} from './StickyWithScrollBack';
 export {default as EmailListField, FinalEmailList} from './EmailListField';
-export {
-  FinalPrincipal,
-  FinalPrincipalList,
-  PrincipalField,
-  PrincipalListField,
-  ACLField,
-  TrackACLField,
-} from './principals';
 export {default as PopoverDropdownMenu} from './PopoverDropdownMenu';
 export {default as ResponsivePopup} from './ResponsivePopup';
 export {default as MathJax} from './MathJax';
--- a/webpack.config.babel.js
+++ b/webpack.config.babel.js
@@ -6,11 +6,12 @@
 // LICENSE file for more details.
 
 import path from 'path';
-import glob from 'glob';
-import {customizeArray, mergeWithCustomize} from 'webpack-merge';
+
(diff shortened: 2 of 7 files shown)


indico/web/client/js/react/components/index.js
```diff
@@ -5,6 +5,14 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
+export {
```
ThiefMaster, line 8: Why was this move needed? (Requiring a particular export order seems rather fragile!)

webpack.config.babel.js
```diff
@@ -150,8 +151,5 @@ export default env => {
     resolve: {
       alias: extraResolveAliases,
     },
-    output: {
-      jsonpFunction: 'coreJsonp',
-    },
```
ThiefMaster, line 155: I think removing this may break plugins; those are separate webpack builds and thus need separate callbacks.

plourenco: What was the cause for adding it?

indico/web/client/js/react/components/index.js
plourenco: I forgot to comment this in one of the commits. The cause is this exact webpack-contrib/mini-css-extract-plugin#382
A clearer explanation is also [here](https://github.com/facebook/create-react-app/issues/5372#issuecomment-685932009).

It's possibly safe to ignore these warnings, however, this was the only change required to silence then alll.

webpack.config.babel.js
ThiefMaster: @pferreir may remember the details, otherwise the git blame and then the commit/PR adding it may also help..

pferreir: https://github.com/indico/indico/commit/902c217299d89c927ac9df961bff4c1946dfc623

pferreir: > This keeps the plugin environment from colliding with the core in
production builds.

webpack/base.js
```diff
@@ -284,14 +282,6 @@ export function webpackDefaults(env, config, bundles, isPlugin = false) {
       new MiniCssExtractPlugin({
         filename: 'css/[name].[contenthash:8].css',
       }),
-      new ProgressBarPlugin({
```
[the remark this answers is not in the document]
plourenco: It's still supported, yes. I'm suggesting (in the latest commit) replacing it with webpack's `--progress` and `--profile` flags which are more informative. But I can always revert this.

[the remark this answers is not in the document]
ThiefMaster: i just tested it and the new progress display is nice. :) ("sending code to the moon and back" was fun though :p)

ThiefMaster:
Room booking (e.g. the list of rooms) is broken due to "process is not defined". Likely this issue: https://stackoverflow.com/a/65018686/298479

webpack.config.babel.js
ThiefMaster: https://webpack.js.org/blog/2020-10-10-webpack-5-release/#automatic-unique-naming

So I guess this is no longer needed :)

plourenco:
> Room booking (e.g. the list of rooms) is broken due to "process is not defined". Likely this issue: https://stackoverflow.com/a/65018686/298479

Thanks, added one for path but forgot this one.

ThiefMaster (approved):

PR #4739, opened 2020-12-08
Add a prefixed field input widget

plourenco:
Includes a prefix in a normal TextInput field as in [2]. We previously used labels to indicate this behaviour [1].

[1] imagem

[2] imagem

Later, updated to

imagem

Perhaps I could name it `PrefixedTextWidget`? 
Or even make it more generic and allow any field to be prefixed, instead of only an input?
I'll leave this as RFC.


ThiefMaster:
Looks really nice now! Just one small thing: When focused, the border looks a bit strange:

image

indico/web/forms/widgets.py
```diff
@@ -288,3 +288,13 @@ def __init__(self, show_field=True):
 
     def __call__(self, field, **kwargs):
         return super().__call__(field, input_args=kwargs)
+
+
+class TextCompleteWidget(JinjaWidget):
```
ThiefMaster, line 293: What about `LabelledTextWidget`? `TextCompleteWidget` sounds like something with autocomplete to me...

ThiefMaster: > Perhaps I could name it `PrefixedTextWidget`?

sounds fine as well

> Or even make it more generic and allow any field to be prefixed, instead of only an input?

i would not spend time on that. wtforms is semi-legacy, and there's a very good chance any fields that are not basic text input fields would not just work with something generic

indico/web/forms/widgets.py
```diff
@@ -288,3 +288,13 @@ def __init__(self, show_field=True):
...
+
+
+class PrefixedTextWidget(JinjaWidget):
+    """Render a TextInput with a prefix"""
+
+    def __init__(self, prefix=None, show_field=True):
+        super().__init__('forms/prefixed_text_widget.html', single_line=True, show_field=show_field, prefix=prefix)
+
+    def __call__(self, field, **kwargs):
+        return super().__call__(field, input_args=kwargs)
```
ThiefMaster, line 300: I don't think this is needed. Have a look at the `single_kwargs` option of JinjaWidget though.

indico/web/templates/forms/prefixed_text_widget.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'forms/base_widget.html' %}
+
+{% block html %}
+    <div class="i-form-field-fixed-width labeled input">
+        <span class="label">{{ prefix }}</span>
+        <input id="{{ field.id }}" name="{{ field.name }}" value="{{ field._value() }}" type="text">
```
ThiefMaster, line 6: You need `{{ input_args|html_params }}` to forward any extra args (such as `disabled` or `required`) to the field.

indico/web/forms/widgets.py
plourenco: Yes, I think I got led by the previous field.

ThiefMaster (approved):

PR #4743, opened 2020-12-11
Get rid of legacy user/group search APIs

### LINKED ISSUE #4674: Get rid of legacy user/group search APIs
Maybe redo the dialogs still them in SUI+React? We can probably reuse most of what we have for the reactified PrincipalField / PrincipalListField!

Most likely we can remove the whole jsonrpc legacy code after having removed these two APIs.

Leats:
Closes #4674


indico/modules/events/persons/controllers.py
```diff
@@ -347,3 +359,54 @@ def _process(self):
...
+        'email': fields.Str(validate=lambda s: len(s) > 3),
+        'affiliation': fields.Str(validate=validate.Length(min=1)),
+        'exact': fields.Bool(missing=False),
+    }, validate=validate_with_message(
+        lambda args: args.keys() & {'first_name', 'last_name', 'email', 'affiliation'},
+        'No criteria provided'
+    ), location='query')
+    def _process(self, exact, **criteria):
+        matches = self._search_event_persons(exact=exact, **criteria)
+        results = sorted((EventPersonSchema().dump(entry) for entry in matches), key=itemgetter('full_name'))
```
ThiefMaster, line 409: Since you already sort the results in the query, you can use `EventPersonSchema().dump(matches[:10], many=True)` here.

Also do the slicing here; no need to serialize more results. Or even better, do something like this in `_search_event_persons`:

    return query.limit(10).all(), query.count()

That way you get the total count but only *retrieve* 10 rows. Not a big deal here since there aren't a ton of EventPersons even in bigger events, but no need to fetch objects from the DB if you never use them.

indico/modules/events/persons/controllers.py
```diff
@@ -347,3 +359,54 @@ def _process(self):
             tpl = get_template_module('events/persons/management/_person_list_row.html')
             return jsonify_data(html=tpl.render_person_row(person_data, bool(self.event.registration_forms)))
         return jsonify_form(form)
+
+
+class RHEventPersonSearch(RHPersonsBase):
+    def _search_event_persons(self, exact=False, **criteria):
+        criteria = {key: value.strip() for key, value in criteria.items() if value.strip()}
```
ThiefMaster, line 366: I think you can use some Python 3.9 magic here ;)

```suggestion
        criteria = {key: v for key, value in criteria.items() if (v := value.strip())}
```

Nothing wrong with your solution though, stripping is cheap!

indico/modules/events/persons/controllers.py
```diff
@@ -347,3 +359,54 @@ def _process(self):
...
+        if not criteria:
+            return set()
+
+        query = EventPerson.query.distinct(EventPerson.id).filter(EventPerson.event_id == self.event.id)
+
+        query = (
+            query.filter(or_(
+                EventPerson.abstract_links.any(AbstractPersonLink.abstract.has(~Abstract.is_deleted)),
+                EventPerson.contribution_links.any(ContributionPersonLink.contribution.has(~Contribution.is_deleted)),
+                EventPerson.event_links.any(EventPersonLink.event.has(~Event.is_deleted)),
```
ThiefMaster, line 377: `.any()` is enough here, you never search for a deleted event anyway (and if we did have an option to search persons for a deleted event, then filtering them out here wouldn't make much sense)

indico/modules/events/persons/controllers.py
```diff
@@ -347,3 +359,54 @@ def _process(self):
...
+                EventPerson.subcontribution_links.any(
+                    SubContributionPersonLink.subcontribution.has(~SubContribution.is_deleted)
```
ThiefMaster, line 379: needs another criterion in case only the parent contribution is explicitly d eleted: `~SubContribution.contribution.has(Contribution.is_deleted)`

indico/modules/events/persons/schemas.py
```diff
@@ -0,0 +1,21 @@
...
+from indico.modules.events.models.persons import EventPerson
+
+
+class EventPersonSchema(mm.SQLAlchemyAutoSchema):
+    class Meta:
+        model = EventPerson
+        fields = ('id', 'identifier', 'email', 'affiliation', 'full_name',
+                  'first_name', 'last_name', 'user_identifier')
+
+    full_name = fields.Function(lambda ft: f'{ft.first_name} {ft.last_name}'.strip() or 'Unknown')
```
ThiefMaster, line 20: what's `ft`? `ep` seems more logical

indico/modules/events/persons/schemas.py
```diff
@@ -0,0 +1,21 @@
...
+    user_identifier = fields.Function(lambda ft: f'User:{ft.user_id}' if ft.user else None)
```
ThiefMaster, line 21: I think you can simplify this to

    user_identifier = fields.String(attribute='user.identifier')

to get the identifier from thw userlying User. Needs testing to see if it correctly handles the case where no user is present. Also, in case this triggers an extra query for each eventperson with a user (to query the user even though you only need their id) then better stick with what you have there right now.

indico/modules/events/persons/controllers.py
```diff
@@ -347,3 +359,54 @@ def _process(self):
...
+            query = query.filter(unaccent_match(getattr(EventPerson, k), v, exact))
+
+        # wrap as subquery so we can apply order regardless of distinct-by-id
+        query = query.from_self()
+        query = query.order_by(
+            db.func.lower(db.func.indico.indico_unaccent(EventPerson.first_name)),
+            db.func.lower(db.func.indico.indico_unaccent(EventPerson.last_name)),
+            EventPerson.id,
+        )
+        return query
```
ThiefMaster, line 395:
```suggestion
        return query.all()
```

indico/modules/users/client/js/Favorites.jsx
```diff
@@ -0,0 +1,171 @@
...
+};
+
+function FavoriteCatManager({userId}) {
+  const [favoriteCats, setFavoriteCats] = useState(null);
+  const [favoriteUsers, [addFavoriteUser, deleteFavoriteUser], loading] = useFavoriteUsers();
+
+  const getFavoriteCats = useCallback(async () => {
+    let res;
+    try {
+      res = await indicoAxios.get(categoryFavoriteURL(userId !== null ? {userId} : {}));
```
ThiefMaster, line 38:
```suggestion
      res = await indicoAxios.get(categoryFavoriteURL(userId !== null ? {user_id: userId} : {}));
```

didn't notice that earlier when suggesting this change 🙈

indico/modules/users/templates/favorites.html
```diff
@@ -1,97 +1,14 @@
...
-                            </span>
-                        {% endif %}
-                    </div>
-                </div>
-            </div>
+            <div id="user-favorites"></div>
         </div>
     </div>
     <script>
+        setupFavoriteSelection({{ (user.id if user else none) | tojson }});
```
ThiefMaster, line 10:
```suggestion
        setupFavoriteSelection({{ (user.id if user != session.user else none) | tojson }});
```

indico/web/client/js/react/components/principals/PersonLinkSearch.jsx
```diff
@@ -0,0 +1,64 @@
...
+  single: PropTypes.bool,
+  triggerFactory: PropTypes.func,
+};
+
+PersonLinkSearch.defaultProps = {
+  disabled: false,
+  withExternalUsers: false,
+  withFavorites: true,
+  withEventPersons: false,
+  eventId: 0,
```
ThiefMaster, line 61: `0` is an actual event when you have a DB migrated from legacy, better default to `null`

indico/web/client/js/react/components/principals/Search.jsx
```diff
@@ -144,17 +145,33 @@ const searchFactory = config => {
...
+                position="top left"
+              />
+            )}
           </Icon.Group>
         </div>
         <div styleName="content">
-          <List.Content>{name}</List.Content>
+          <List.Content>
+            {name}
+            {/* {existsInEvent && (
```
ThiefMaster, line 168: 👀

indico/web/client/js/react/components/principals/Search.jsx
```diff
@@ -360,6 +384,8 @@ const searchFactory = config => {
     alwaysConfirm: false,
     onOpen: () => {},
     onClose: () => {},
+    withEventPersons: false,
+    eventId: 0,
```
ThiefMaster, line 388: `null`

indico/web/client/js/react/hooks.js
```diff
@@ -21,15 +21,26 @@ export const useFavoriteUsers = () => {
   // once on the same page we keep the favorites in sync and don't send multiple
   // requests to load the initial list
   const [favorites, setFavorites] = useState({});
+  const [loading, setLoading] = useState(false);
+  const userId = null;
 
   const apiCall = async (method, id = null, source = null) => {
+    setLoading(true);
```
ThiefMaster, line 28: Oh, I guess that's what caused the behavior we were talking about before! Maybe you could only set loading during the initial load of the favorite list, but not when performing a modification?

indico/web/templates/forms/permissions_widget.html
```diff
@@ -10,7 +10,7 @@ <h3>Permissions</h3>
                 <a href="#" class="i-button label text-color outline add-entry-decoration">
                     {%- trans %}Add{% endtrans -%}
                 </a>
-                <a class="i-button js-add-user-group">{% trans %}User / Group{% endtrans %}</a>
+                <span id='js-add-user-group'></span>
```
ThiefMaster, line 13:
```suggestion
                <span id="js-add-user-group"></span>
```

we usually use single quotes in html attributes

indico/modules/users/controllers.py
```diff
@@ -306,40 +306,39 @@ def _process(self):
...
-        self.user.favorite_users |= {_f for _f in users if _f}
-        tpl = get_template_module('users/_favorites.html')
-        return jsonify(success=True, users=[serialize_user(user) for user in users],
-                       html=tpl.favorite_users_list(self.user))
+class RHUserFavoritesAPI(RHProtected):
+    def _process_args(self):
+        self.user = User.get_or_404(request.view_args['fav_user_id']) if 'fav_user_id' in request.view_args else None
 
+    def _process_GET(self):
+        return jsonify(sorted(u.id for u in session.user.favorite_users))
```
ThiefMaster, line 317: `self.user`

indico/modules/users/controllers.py
```diff
@@ -306,40 +306,39 @@ def _process(self):
...
-class RHUserFavoritesUsersAdd(RHUserBase):
-    def _process(self):
-        users = [User.get(int(id_)) for id_ in request.form.getlist('user_id')]
-        self.user.favorite_users |= {_f for _f in users if _f}
-        tpl = get_template_module('users/_favorites.html')
-        return jsonify(success=True, users=[serialize_user(user) for user in users],
-                       html=tpl.favorite_users_list(self.user))
+class RHUserFavoritesAPI(RHProtected):
+    def _process_args(self):
+        self.user = User.get_or_404(request.view_args['fav_user_id']) if 'fav_user_id' in request.view_args else None
```
ThiefMaster, line 314: Careful, this exposes everyone's favorites to any other user. Probably better to just inherit from `RHUserBase` which already assigns `self.user` properly and performs an access check when specifying a user that is not you.

indico/modules/users/client/js/Favorites.jsx
```diff
@@ -0,0 +1,171 @@
...
+            )}
+          </div>
+        </div>
+      </div>
+    </>
+  );
+}
+
+FavoriteCatManager.propTypes = {
+  userId: PropTypes.number.isRequired,
```
ThiefMaster, line 170: not required & default to null (due to the change i suggested in the template in order to only pass a user if if it's not the current user like we do on the profile picture page)

indico/modules/events/persons/controllers.py
```diff
@@ -347,3 +358,51 @@ def _process(self):
...
+        query = EventPerson.query.distinct(EventPerson.id).filter(EventPerson.event_id == self.event.id)
+
+        query = query.filter(
+            or_(
+                EventPerson.abstract_links.any(AbstractPersonLink.abstract.has(~Abstract.is_deleted)),
+                EventPerson.contribution_links.any(ContributionPersonLink.contribution.has(~Contribution.is_deleted)),
+                EventPerson.event_links.any(),
+                EventPerson.subcontribution_links.any(
+                    SubContributionPersonLink.subcontribution.has(~SubContribution.is_deleted)),
+                EventPerson.subcontribution_links.any(~SubContribution.contribution.has(Contribution.is_deleted)),
```
ThiefMaster, line 379: This needs to be one condition - otherwise having a link to a deleted subcontribution of a non-deleted contribution would work.

I think this works (but untested):

```suggestion
                EventPerson.subcontribution_links.any(
                    SubContributionPersonLink.subcontribution.has(
                        ~SubContribution.is_deleted,
                        SubContribution.contribution.has(~Contribution.is_deleted)
                    )
                ),
```

indico/modules/users/client/js/Favorites.jsx
```diff
@@ -0,0 +1,173 @@
...
+import {PersonLinkSearch} from 'indico/react/components/principals/PersonLinkSearch';
+import {useFavoriteUsers} from 'indico/react/hooks';
+import {Translate} from 'indico/react/i18n';
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+
+import './Favorites.module.scss';
+
+window.setupFavoriteSelection = function setupFavoriteSelection(userId) {
+  ReactDOM.render(
+    <FavoriteCatManager userId={userId} />,
```
ThiefMaster, line 26: `FavoriteManager` since it contains both users and cats; and maybe split the user/cat-specific parts in separate components

indico/web/client/js/jquery/widgets/jinja/permissions_widget.js
```diff
@@ -479,9 +461,67 @@ import Palette from 'indico/utils/palette';
...
+        </a>
+      );
+
+      const existing = JSON.parse(this.element.find('input[name="permissions"]').val()).map(
+        e => e[0].identifier
+      );
+
+      ReactDOM.render(
+        <>
+          <PersonLinkSearch
```
ThiefMaster, line 482: In the end, `PersonLinkSearch` is only wrapping `UserSearch` and including the `useFavoriteUsers` hook. I think we could just get rid of `PersonLinkSearch` altogether and get the favorite users in the two places where `<PersonLinkSearch` is currently used and use `UserSearch` directly.

indico/web/client/js/react/components/principals/Search.jsx
```diff
@@ -431,14 +444,60 @@ const InnerUserSearch = searchFactory({
...
+        lastName,
+        existsInEvent: false,
+        email,
+        affiliation,
+      })
+    );
     delete resultData.users;
+    if (withEventPersons) {
+      let epResponse;
+      values.confId = eventId;
```
ThiefMaster, line 465: Instead of modifying this I'd rather do something like this below:

```
        epResponse = await indicoAxios.get(eventPersonSearchURL({...values, confId: eventId}));
```

Assuming `favorites_first` isn't event used in this endpoint here, we could do the same for it above, so `values` really only contains the search criteria (in which case I'd rename it to something like `criteria`).

.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         id: cache-pip
         with:
           path: .venv
-          key: ${{ runner.os }}-pip-${{ hashFiles('requirements*.txt') }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements*.txt') }}-x
```
ThiefMaster, line 23: TODO remove this again later. for some reason the github actions cache with the usual key is corrupted and there's no way to clear that cache...

indico/modules/users/client/js/Favorites.jsx
```diff
@@ -0,0 +1,198 @@
...
+window.setupFavoriteSelection = function setupFavoriteSelection(userId) {
+  ReactDOM.render(<FavoriteManager userId={userId} />, document.getElementById('user-favorites'));
+};
+
+function FavoriteManager({userId}) {
+  const [favoriteUsers, [addFavoriteUser, deleteFavoriteUser], loading] = useFavoriteUsers(userId);
+
+  if (loading) {
+    return <Loader active />;
+  }
```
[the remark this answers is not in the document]
Leats: I had it here because otherwise the category fav box jumps around after the user fav box is loaded. but I guess I can just place the loader further inside the divs to stop it from jumping around!

ThiefMaster (approved):

PR #4746, opened 2020-12-16
Show vc join buttons in timetable

bpedersen2:
Try and show the relevant join buttons:
 In over view mode:
   - session block vc room
   - or event vc room if no session block room is found
 In detail mode:
   - contribution vc room
   - or as above

### DIFF
--- a/indico/modules/events/timetable/legacy.py
+++ b/indico/modules/events/timetable/legacy.py
@@ -140,6 +140,8 @@ class TimetableSerializer(object):
                      'pdf': url_for('sessions.export_session_timetable', block.session),
                      'url': url_for('sessions.display_session', block.session),
                      'friendlyId': block.session.friendly_id})
+        vcdata = self._get_vc_joinbuttons(entry)
+        data['vcrooms'] = ''.join(vcdata)
         return data
 
     def serialize_contribution_entry(self, entry):
@@ -174,6 +176,8 @@ class TimetableSerializer(object):
                      'friendlyId': contribution.friendly_id,
                      'references': map(SerializerBase.serialize_reference, contribution.references),
                      'board_number': contribution.board_number})
+        vcdata = self._get_vc_joinbuttons(entry)
+        data['vcrooms'] = ''.join(vcdata)
         return data
 
     def serialize_break_entry(self, entry, management=False):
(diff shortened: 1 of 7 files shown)


ThiefMaster:
I like the idea, but it needs some polishing for sure... currently it doesn't look great:

image

indico/modules/events/timetable/legacy.py
```diff
@@ -195,6 +199,24 @@ def serialize_break_entry(self, entry, management=False):
...
+            if plugin:
+                vcdata.append(plugin.render_event_buttons(vc_room, vc_room_assoc))
+        if not vcdata and entry.parent:
+            vcdata = self._get_vc_joinbuttons(entry.parent)
+        if not vcdata and event:
+            for vc_room_assoc in event.vc_room_associations:
+                vc_room = vc_room_assoc.vc_room
+                plugin = vc_room.plugin
+                if plugin:
+                    vcdata.append(plugin.render_event_buttons(vc_room, vc_room_assoc))
```
ThiefMaster, line 217: This code looks almost like the code above. It can be refactored to avoid duplicating it...

indico/web/client/js/legacy/libs/timetable/Draw.js
```diff
@@ -202,9 +202,12 @@ type(
         if (this.eventData.location) {
           this.locationDiv.append(addComma ? ', ' : '');
           this.locationDiv.append(this.eventData.location);
-        }
-
-        // If it's a contribution add the speakers information
+        } 
+      	this.vcdiv = Html.div('timetableBlockVC');
+	if (this.eventData.vcrooms) {
```
ThiefMaster, line 207: Something went wrong with the indentation here. Accidental use of tabs instead of spaces?

ThiefMaster (changes requested):

bpedersen2:
> I like the idea, but it needs some polishing for sure... currently it doesn't look great:

> Maybe it could go in the info balloon instead?

There I  did put it as ewll (with a local override template, I'll add this  as well), but the idea was to make it really easy to discover.
But I' ll take a look if I can position it better

indico/web/client/js/legacy/libs/timetable/Draw.js
bpedersen2: Should be fixed now

indico/modules/events/timetable/legacy.py
bpedersen2: refactored

kewisch:
Maybe it would make sense to designate a special area for "buttons", and then give the template hook a more general name? I could imagine this would be useful for creating more timetable entry actions that other plugins might use as well.

ThiefMaster:
Something we'll certainly consider  for the new timetable in v3.4 :)

ThiefMaster:
old PR, closing in favor of hopefully coming up with something nice in the v3.4 timetable

PR #4747, opened 2020-12-16
Start using the user's profile picture in more places

### LINKED ISSUE #4625: Start using the user's profile picture in more places
We now have a proper `picture_url` for every user which will serve a default icon (colored bg + initial) or a custom picture/avatar. Currently this is only used in the user dashboard, but we can now start using this in other places as well.

Look for `avatarBgColor` in the code to find those places!

Leats:
Closes #4625.

Add a default picture for the Indico Bot in editing:
Bildschirmfoto 2020-12-16 um 12 00 33

Start using avatars in the meeting's participant list as well:
Bildschirmfoto 2020-12-16 um 15 42 31
This also changes that avatars are also visible if nobody is logged in.


indico/modules/events/client/js/reviewing/components/UserAvatar.jsx
```diff
@@ -5,24 +5,28 @@
...
   return (
-    <div
-      className="i-timeline-item-label avatar-placeholder"
-      style={{backgroundColor: user.avatarBgColor}}
-    >
-      {user.fullName[0]}
+    <div>
+      <Image
+        avatar
+        src={user.pictureURL ? user.pictureURL : imageURL({filename: 'robot.svg'})}
```
ThiefMaster, line 19:
```suggestion
        src={user.pictureURL || imageURL({filename: 'robot.svg'})}
```

indico/modules/events/registration/models/registrations.py
```diff
@@ -321,6 +321,11 @@ def display_full_name(self):
         """Return the full name using the user's preferred name format."""
         return format_display_full_name(session.user, self)
 
+    @property
+    def avatar_url(self):
+        """Return the url of the user's avatar."""
+        return self.user.picture_url
```
ThiefMaster, line 327: `self.user` can be None

indico/modules/users/controllers.py
```diff
@@ -223,10 +223,11 @@ def _process(self, source):
             return send_file('avatar.png', BytesIO(gravatar), mimetype='image/png')
 
 
-class RHProfilePictureDisplay(RHUserBase):
+class RHProfilePictureDisplay(RH):
     """Display the user's profile picture."""
 
-    allow_system_user = True
+    def _process_args(self):
+        self.user = User.get(request.view_args['user_id'], is_deleted=False)
```
ThiefMaster, line 230:
```suggestion
        self.user = User.get_or_404(request.view_args['user_id'], is_deleted=False)
```

indico/modules/users/schemas.py
```diff
@@ -16,8 +16,8 @@ class UserSchema(mm.SQLAlchemyAutoSchema):
 
     class Meta:
         model = User
-        fields = ('id', 'identifier', 'first_name', 'last_name', 'email', 'affiliation', 'avatar_bg_color', 'full_name',
-                  'phone')
+        fields = ('id', 'identifier', 'first_name', 'last_name', 'email', 'affiliation', 'full_name',
+                  'phone', 'picture_url')
```
ThiefMaster, line 20: how ugly would it be to change it to `avatar_url` everywhere?

Leats: depends, do we want to change stuff like `picture_metadata` to as well?

ThiefMaster: I wouldn't especially since those are DB columns... but the `picture_url` is just a property IIRC

indico/modules/events/registration/controllers/display.py
```diff
@@ -392,3 +395,20 @@ def _check_access(self):
...
+
+
+class RHRegistrationAvatar(RH):
+    """Display a standard avatar for a registration based on the full name."""
+
+    def _process_args(self):
+        self.registration = (Registration
+                             .find(Registration.id == request.view_args['registration_id'],
+                                   ~Registration.is_deleted,
+                                   ~RegistrationForm.is_deleted).one())
```
ThiefMaster, line 407: This causes an extremely expensive query since without JOINs it will do a cross-join between the registration form and registration table, so suddenly the DB needs to look at ten thousands of rows (all registration forms that are not deleted multiplied by the number of registrations matching your filter (luckily that's always 1 here)).

Also, `find()` is deprecated ;)

ThiefMaster:
```python
class RHRegistrationAvatar(RHDisplayEventBase):
    """Display a standard avatar for a registration based on the full name."""

    normalize_url_spec = {
        'locators': {
            lambda self: self.registration
        }
    }

    def _process_args(self):
        RHDisplayEventBase._process_args(self)
        self.registration = (Registration.query
                             .filter(Registration.id == request.view_args['registration_id'],
                                     ~Registration.is_deleted,
                                     ~RegistrationForm.is_deleted)
                             .one())
```

Here's a similar version that avoids this problem and also enables the url normalization in order to ensure it's not returning a registration from the wrong event.

indico/modules/events/registration/controllers/display.py
```diff
@@ -392,3 +395,20 @@ def _check_access(self):
...
+
+    def _process(self):
+        first_name = self.registration.full_name[0].upper() if self.registration.full_name else ''
+        avatar = render_template('users/avatar.svg', bg_color=get_color_for_username(self.registration.full_name),
+                                 text=first_name)
+        return send_file('avatar.svg', BytesIO(avatar.encode()), mimetype='image/svg+xml',
+                         no_cache=False, inline=True, safe=False, cache_timeout=(86400*7))
```
ThiefMaster, line 414: What about adding a `send_default_avatar(name)` util to `.../users/util.py` containing this logic? Then you could just `return send_avatar(self.registration.first_name)` in here and do something similar in the other places where we return this default avatar.

indico/modules/events/registration/templates/display/event_header.html
```diff
@@ -83,7 +83,12 @@
             <div class="participant-list-wrapper">
                 <ul class="participant-list">
                     {%- for participant in event.published_registrations|sort(attribute='display_full_name') -%}
-                        <li class="event-user {% if participant.checked_in and participant.registration_form.publish_checkin_enabled %}checked-in{% endif %}">
+                        <li class="ui image label meeting-participant {% if participant.checked_in and participant.registration_form.publish_checkin_enabled %} checked-in {% endif %}">
+                            {%- if participant.user -%}
+                                <img src="{{ participant.avatar_url }}" alt="">
+                            {%- else -%}
+                                <img src="{{ url_for('event_registration.registration_avatar', participant)}}" alt="">
+                            {%- endif -%}
```
ThiefMaster, line 91: This logic would better fit in the `avatar_url` property of `Registration`.

indico/modules/events/registration/models/registrations.py
```diff
@@ -321,6 +321,11 @@ def display_full_name(self):
         """Return the full name using the user's preferred name format."""
         return format_display_full_name(session.user, self)
 
+    @property
+    def avatar_url(self):
+        """Return the url of the user's avatar."""
+        return self.user.avatar_url if self.user else None
```
ThiefMaster, line 327: Let's always return the registrant-specific URL here

indico/modules/events/registration/controllers/display.py
ThiefMaster: And together with that, we could also move the `RHProfilePictureDisplay` logic into a `send_avatar(user)`, which then uses  either send_default_avatar or the usual code to serve the custom one.

That allows you to simplify the code in here to something like this:

```python
if self.registration.user:
    return send_avatar(self.registration.user) 
return send_default_avatar(self.registration.first_name)
```

indico/modules/users/controllers.py
```diff
@@ -223,22 +222,14 @@ def _process(self, source):
...
-            first_name = self.user.first_name[0].upper() if self.user.first_name else ''
-            avatar = render_template('users/avatar.svg', bg_color=self.user.avatar_bg_color, text=first_name)
-            return send_file('avatar.svg', BytesIO(avatar.encode()), mimetype='image/svg+xml',
-                             no_cache=False, inline=True, safe=False, cache_timeout=(86400*7))
-
-        metadata = self.user.picture_metadata
-        return send_file('avatar.png', BytesIO(self.user.picture), mimetype=metadata['content_type'],
-                         inline=True, conditional=True, last_modified=parse_date(metadata['lastmod']),
-                         cache_timeout=(86400*7))
+        send_avatar(self.user)
```
ThiefMaster, line 232:
```suggestion
        return send_avatar(self.user)
```

PR #4748, opened 2020-12-28
Add uk_UA translation

andriish:
Dear Indico developers team,

please have a look and this PR that includes a new translation.
The translation files were combined from the individual files obtained from the Transifex. 

Best regards,

Andii and Indico Ukrainian translation team


ThiefMaster:
thanks, but we do not use PRs for i18n updates (and translations are currently based on the stable version which is in the 2.3-maintenance branch, not master) - instead, we pull translations directly from transifex shortly before making a release.

anyway, since the translation is complete we can put it on one of our test servers next week (unless you already tested it on your own indico instance?)

andriish:
Hi @ThiefMaster ,

thanks for the quick reply.

0) > the stable version which is in the 2.3-maintenance 

Thanks! I'm planning to have a look at the 2.3 a bit before the upgrade, so that is good to know.

1) > since the translation is complete

The translation has been ready for 2.2 1-1.5 years ago ( and on many occasions before) and I've used the 2.2 version in production since that time.  This translation is for 2.3. The difference is small, ~10%, but testing it on the test servers would be great. Thank you!

2)Another question I have is about the documentation. The docs on the deployment of translations in Indico cover mostly the case of deployment of the translations during the installation of Indico. Would that make sense to cover the case of adding/updating a translation to an already running instance, starting from the *po files on Transifex? I could provide some description/docs. What is your opinion?

Best regards,

Andrii

ThiefMaster:
Do you mean [this documentation](https://docs.getindico.io/en/stable/building/#including-a-new-translation)? ;)

ThiefMaster:
I reopened the PR and included some fixes needed to actually use the translation in indico

andriish:
Hi @ThiefMaster ,

> Do you mean this documentation? ;)

Yes.  If you find it is a good idea I could prepare a separate PR with documentation. What do you think?

> I reopened the PR and included some fixes needed to actually use the translation in indico

Cool! Thanks a lot! Anything I can do?


Best regards,

Andrii

ThiefMaster:
> Yes. If you find it is a good idea I could prepare a separate PR with documentation. What do you think?

What I meant is that we already have documentation on how to build your own package with a custom translation included..

> Cool! Thanks a lot! Anything I can do?

I just deployed the translation on a test server; my colleague will send you the details to access it shortly.

andriish:
Hi @ThiefMaster ,

>  build your own package with a custom translation included..

Exactly.  What I mean is to provide docs for deployment a custom/additional translation into the running instance, i.e. translation-only.

Best regards,

Andrii

ThiefMaster:
That's not really supported - as you can see in the PR here we currently need some minor JS changes (including a rebuild of the assets) when adding a new locale. Otherwise e.g. Room Booking is broken when using the new locale.

andriish:
> Room Booking is broken when using the new locale.

I remember some extra actions were needed to make the Room Booking work with the new translations installed on the running instance, but if

> That's not really supported

of course, there should be no docs for unsupported options.

Best regards,

Andrii

PR #4752, opened 2021-01-08
Remove extra margin in roles column

### LINKED ISSUE #4751: Layout bug in Participant Roles list
image

The little pencil icon is cut off, on the right hand side.

plourenco:
Closes #4751


plourenco:
Not relevant anymore as the original issue was related to the right margin of the clickable pencil item, and we couldn't reproduce it again.

Nonetheless, this seems to be a problem with the column organization as we could instead use a flex layout and align the content of the last column to the leftmost side.

PR #4753, opened 2021-01-11
Improve VC linking dropdowns

ThiefMaster:


plourenco (approved):
Not sure if `contrib_choices` wouldn't be more readable as a plain loop. Nonetheless, looks good to go!

PR #4760, opened 2021-01-14
Refresh user data when cloning an event

### LINKED ISSUE #4750: Event cloning: add "update user data" checkbox
Scenario:
 * User is added to an event;
 * User updates their profile data (on their profile page);
 * Someone else clones the aforementioned event;
 * Event still contains the old user data.

This is a pretty common source of misunderstandings. I suggest we add an "update user data" checkbox which will tell Indico to fetch the users' latest profile data upon cloning. I would also suggest to have that enabled by default.

ThiefMaster:
closes #4750

image

### DIFF
--- a/indico/modules/events/models/persons.py
+++ b/indico/modules/events/models/persons.py
@@ -12,6 +12,7 @@ from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import mapper
 
+from indico.core import signals
 from indico.core.db.sqlalchemy import PyIntEnum, UTCDateTime, db
 from indico.core.db.sqlalchemy.principals import EmailPrincipal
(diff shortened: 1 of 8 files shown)


indico/modules/events/models/persons.py
```diff
@@ -242,6 +242,17 @@ def link_user_by_email(cls, user):
                 db.session.delete(event_person)
         db.session.flush()
 
+    def sync_user(self):
```
plourenco, line 245: Not sure about the further uses of `sync_user` but would it make sense to trigger the `person-updated` signal?

plourenco (approved):
👍

indico/modules/events/models/persons.py
ThiefMaster: When cloning no, but when doing it in an existing event it probably makes sense. So I'll probably add a kwarg for it and default to true..

PR #4764, opened 2021-01-20
Let cloners know about multiple clones

pferreir:
This is needed currently for Zoom, but will basically allow any cloner to know whether only one event is being cloned is supposed to be the only occurrence (`None`) or one of many (0-indexed).
We're now also allowing VC plugins to not clone a room (return `None`).

### DIFF
--- a/indico/modules/events/cloning.py
+++ b/indico/modules/events/cloning.py
@@ -69,8 +69,8 @@ class EventCloner(object):
                       key=attrgetter('friendly_name'))
 
     @classmethod
-    def run_cloners(cls, old_event, new_event, cloners, event_exists=False):
-        all_cloners = OrderedDict((name, cloner_cls(old_event))
+    def run_cloners(cls, old_event, new_event, cloners, n_occurrence=0, event_exists=False):
+        all_cloners = OrderedDict((name, cloner_cls(old_event, n_occurrence))
                                   for name, cloner_cls in get_event_cloners().iteritems())
         if any(cloner.is_internal for name, cloner in all_cloners.iteritems() if name in cloners):
             raise Exception('An internal cloner was selected')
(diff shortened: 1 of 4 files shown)


indico/modules/events/operations.py
```diff
@@ -153,11 +153,12 @@ def update_event(event, update_timetable=False, **data):
     _log_event_update(event, changes, visible_person_link_changes=visible_person_link_changes)
 
 
-def clone_event(event, start_dt, cloners, category=None):
+def clone_event(event, n_occurrence, start_dt, cloners, category=None):
     """Clone an event on a given date/time.
 
     Runs all required cloners.
 
+    :param n_occurrence: The number of the occurrence, if this is a "recurring" clone, otherwise `None`
```
ThiefMaster, line 161: Let's use `0` in case of non-recurring? Otherwise the `> 0` check will fail in Python 3...

pferreir: Oh, I didn't realize that, good point. Thanks.

indico/modules/events/cloning.py
```diff
@@ -65,12 +65,12 @@ class EventCloner(object):
     @classmethod
     def get_cloners(cls, old_event):
         """Return the list of cloners (sorted for display)."""
-        return sorted((cloner_cls(old_event) for cloner_cls in get_event_cloners().itervalues()),
+        return sorted((cloner_cls(old_event, 0) for cloner_cls in get_event_cloners().itervalues()),
```
ThiefMaster, line 68: hm, default value `0` for that arg instead of passing it explicitly here?

indico/modules/events/operations.py
pferreir: Ok, done.

ThiefMaster (approved):
works fine!

PR #4765, opened 2021-01-21
Move package metadata from setup.py to setup.cfg

### LINKED ISSUE #4680: Modernize setuptools packaging
- Move stuff from setup.py to setup.cfg
- Use `[dev]` extra to install dev dependencies
- Check for dev dependencies currently not marked as such (e.g. all the pytest stuff)

~~We could also look into using `poetry` and `pyproject.toml` instead.~~ (something for a different issue.. maybe!)

ThiefMaster:
closes #4680

related changes (without PRs) for the plugins:

- https://github.com/indico/indico-plugins/compare/b3abd4f...865abd6
- https://github.com/indico/indico-plugins-cern/commit/a9e7887


ThiefMaster:
Keeping requirements in separate files instead of adding them to setup.cfg because:

- GitHub's dependency graph doesn't take setup.cfg into account
- Not sure if dependabot uses it
- CI setup with caching the pyhon env is slightly easier with requirements*.txt files

PR #4769, opened 2021-02-01
Add a new form to capture rejection reason for registration(s)

vasantvohra:
Test Instructions
- Create an event with enabled moderation in the registration form.
- Register for the event
- Go to the registration display page, reject the participant, a form in the popup window will appear to enter a reason for rejection and a switch to enable or disable attaching the reason with the notification.
- The reason will be stored in the database as `registration.rejection_reason`, added as data into the event log,
- Also available to use in email templates as `rejection_reason` tag.
- Similarly, while rejecting multiple registrations from the registration list the same popup window would appear.

> Squash and Merge pl.

- Let me know if there are any changes required.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+Version 2.3.4
+-------------
+
+Improvements
+^^^^^^^^^^^^
+
+- Store a reason for rejection of registration(s) if stated by the moderator
+  through a form in a dialog box, can be sent to users along with notification,
+  Add a new 'rejection_reason' placeholder for an email template.
+  (:pr:`4740`, thanks :user:`vasantvohra`)
+
 
 Version 2.3.4
 -------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,10 +7,11 @@ Version 2.3.4
 Improvements
 ^^^^^^^^^^^^
 
-- Store a reason for rejection of registration(s) if stated by the moderator
-  through a form in a dialog box, can be sent to users along with notification,
-  Add a new 'rejection_reason' placeholder for an email template.
-  (:pr:`4740`, thanks :user:`vasantvohra`)
+- When the moderator states the reason for the rejection of registrations in a form available in the popup window.
+  The reason is saved in the database as `registration.rejection_reason` and added to the event log.
+  A switch allows the reason to be attached to an email notification.
+  Also, add a new 'rejection_reason' placeholder/tag to use in email templates.
+  (:pr:`4769`, thanks :user:`vasantvohra`)
 
 
 Version 2.3.4
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,22 +10,12 @@ Version 2.3.4
 Improvements
 ^^^^^^^^^^^^
 
-- Allow managers to specify a reason when rejecting registrants and add a new placeholder
-  for the rejection reason when emailing registrants (:pr:`4769`, thanks :user:`vasantvohra`)
-
-
-Version 2.3.4
--------------
-
-*Unreleased*
-
-Improvements
-^^^^^^^^^^^^
-
 - Fail more gracefully is a user has an invalid locale set and fall back to the default
   locale or English in case the default locale is invalid as well
 - Log an error if the configured default locale does not exist
 - Add ID-1 page size for badge printing (:pr:`4774`, thanks :user:`omegak`)
+- Allow managers to specify a reason when rejecting registrants and add a new placeholder
+  for the rejection reason when emailing registrants (:pr:`4769`, thanks :user:`vasantvohra`)
 
 Bugfixes
 ^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,10 +10,8 @@ Version 2.3.4
 Improvements
 ^^^^^^^^^^^^
 
-- Add a new form to capture rejection reason for registration, store it in DB, and event log.
-  If enabled send the reason with notification email to registrant and event managers.
-  Add a new 'rejection_reason' placeholder to use in email templates.
-  (:pr:`4769`, thanks :user:`vasantvohra`)
+- Allow managers to specify a reason when rejecting registrants and add a new placeholder
+  for the rejection reason when emailing registrants (:pr:`4769`, thanks :user:`vasantvohra`)
 
 
 Version 2.3.4
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,16 +1,18 @@
 Changelog
 =========
 
+
 Version 2.3.4
 -------------
+
+*Unreleased*
 
 Improvements
 ^^^^^^^^^^^^
 
-- When the moderator states the reason for the rejection of registrations in a form available in the popup window.
-  The reason is saved in the database as `registration.rejection_reason` and added to the event log.
-  A switch allows the reason to be attached to an email notification.
-  Also, add a new 'rejection_reason' placeholder/tag to use in email templates.
+- Add a new form to capture rejection reason for registration, store it in DB, and event log.
+  If enabled send the reason with notification email to registrant and event managers.
+  Add a new 'rejection_reason' placeholder to use in email templates.
   (:pr:`4769`, thanks :user:`vasantvohra`)
 
 
--- a/indico/modules/events/registration/controllers/management/reglists.py
+++ b/indico/modules/events/registration/controllers/management/reglists.py
@@ -569,11 +569,12 @@ class RHRegistrationReject(RHManageRegistrationBase):
 
     def _process(self):
         form = RejectRegistrantsForm()
+        message = _("Rejecting this registration will trigger a notification email.")
         if form.validate_on_submit():
             self.registration.rejection_reason = form.rejection_reason.data
             _modify_registration_status(self.registration, approve=False)
             return jsonify_data(html=_render_registration_details(self.registration))
-        return jsonify_form(form, disabled_until_change=False, submit='Reject')
+        return jsonify_form(form, disabled_until_change=False, submit='Reject', message=message)
 
 
 class RHRegistrationReset(RHManageRegistrationBase):
@@ -651,19 +652,19 @@ class RHRegistrationsModifyStatus(RHRegistrationsActionBase):
         if approve:
             for registration in self.registrations:
                 _modify_registration_status(registration, approve)
-            flash(_("The status of the selected registrations was successfully '{}'.").format('approved'), 'success')
+            flash(_("The selected registrations was successfully '{}'.").format('approved'), 'success')
             return jsonify_data(**self.list_generator.render_list())
         else:
             form = RejectRegistrantsForm(obj=FormDefaults(flag=request.args.get('flag'),
                                                           registration_id=request.args.getlist('registration_id')))
+            message = _("Rejecting the selected registrations will trigger a notification email for each registrant.")
             if form.validate_on_submit():
                 for registration in self.registrations:
                     registration.rejection_reason = form.rejection_reason.data
                     _modify_registration_status(registration, approve)
-                flash(_("The status of the selected registrations was successfully '{}'.").format('rejected'),
-                      'success')
+                flash(_("The selected registrations was successfully '{}'.").format('rejected'), 'success')
                 return jsonify_data(**self.list_generator.render_list())
-            return jsonify_form(form, disabled_until_change=False, submit='Reject')
+            return jsonify_form(form, disabled_until_change=False, submit='Reject', message=message)
 
 
 class RHRegistrationsExportAttachments(RHRegistrationsExportBase, ZipGeneratorMixin):
--- a/indico/modules/events/registration/controllers/management/reglists.py
+++ b/indico/modules/events/registration/controllers/management/reglists.py
@@ -587,7 +587,7 @@ class RHRegistrationReset(RHManageRegistrationBase):
         if self.registration.state in (RegistrationState.complete, RegistrationState.unpaid):
             self.registration.update_state(approved=False)
         elif self.registration.state == RegistrationState.rejected:
-            self.registration.rejection_reason = None
+            self.registration.rejection_reason = ''
             self.registration.update_state(rejected=False)
         elif self.registration.state == RegistrationState.withdrawn:
             self.registration.update_state(withdrawn=False)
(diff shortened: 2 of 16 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,18 @@
 Changelog
 =========
 
+Version 2.3.4
```
ThiefMaster, line 4: Please rebase, we already have a stub for this version in the latest 2.3-maintenance branch.

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -640,11 +648,23 @@ class RHRegistrationsModifyStatus(RHRegistrationsActionBase):
     """Accept/Reject selected registrations."""
 
     def _process(self):
-        approve = request.form['flag'] == '1'
-        for registration in self.registrations:
-            _modify_registration_status(registration, approve)
-        flash(_("The status of the selected registrations was updated successfully."), 'success')
-        return jsonify_data(**self.list_generator.render_list())
+        approve = request.form.get('flag') == '1'
```
ThiefMaster, line 651: I think this is probably a good time to split accept/reject into two separate RHs. Previously combining the two actions made sense because there was almost no difference in the code, but that's no longer the case now...

indico/modules/events/registration/forms.py
```diff
@@ -407,3 +407,11 @@ def __init__(self, *args, **kwargs):
         super(ImportRegistrationsForm, self).__init__(*args, **kwargs)
         if not self.regform.moderation_enabled:
             del self.skip_moderation
+
+
+class RejectRegistrantsForm(IndicoForm):
+    rejection_reason = TextAreaField(_('Reason'), description=_("Provide a reason for rejection of registration(s)."))
```
ThiefMaster, line 413: What about something like "You can provide a reason for the rejection here"? This avoids the ugly "(s)" pluralization.

indico/modules/events/registration/forms.py
```diff
@@ -407,3 +407,11 @@ def __init__(self, *args, **kwargs):
...
+    attach_reason = BooleanField(_('Attach reason'), [HiddenUnless('rejection_reason')], widget=SwitchWidget(),
+                                 description=_("Whether to attach reason with the reject notification"))
+    flag = HiddenField()
```
ThiefMaster, line 416: This can go away when splitting the RHs

indico/modules/events/registration/logging.py
```diff
@@ -35,13 +35,15 @@ def log_registration_updated(registration, previous_state, **kwargs):
...
     previous_state_title = orig_string(previous_state.title)
+    data = {'Previous state': previous_state_title}
     if (previous_state == RegistrationState.pending
             and registration.state in (RegistrationState.complete, RegistrationState.unpaid)):
         log_text = 'Registration for "{}" has been approved'
         kind = EventLogKind.positive
     elif previous_state == RegistrationState.pending and registration.state == RegistrationState.rejected:
         log_text = 'Registration for "{}" has been rejected'
         kind = EventLogKind.negative
+        data["Reason"] = registration.rejection_reason if registration.rejection_reason else None
```
ThiefMaster, line 46: `registration.rejection_reason or None` but I'd simply omit it in that case.

indico/modules/events/registration/models/registrations.py
```diff
@@ -186,7 +186,11 @@ class Registration(db.Model):
         UTCDateTime,
         nullable=True
     )
-
+    #: If given a reason for rejection
+    rejection_reason = db.Column(
+        db.String,
+        nullable=True
```
ThiefMaster, line 192: We usually use `nullable=False, default=''` to avoid a mix of `None` and `''` in the database.

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -566,8 +568,13 @@ class RHRegistrationReject(RHManageRegistrationBase):
     """Reject a registration."""
 
     def _process(self):
-        _modify_registration_status(self.registration, approve=False)
-        return jsonify_data(html=_render_registration_details(self.registration))
+        form = RejectRegistrantsForm()
+        message = _("Rejecting this registration will trigger a notification email.")
+        if form.validate_on_submit():
+            self.registration.rejection_reason = form.rejection_reason.data
+            _modify_registration_status(self.registration, approve=False)
```
ThiefMaster, line 575: What about moving the reason update into `_modify_registration_status` (only used in case of `approve=False` of course)?

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -636,17 +644,31 @@ def _process(self):
...
+class RHRegistrationsApprove(RHRegistrationsActionBase):
+    """Accept selected registrations from registration list."""
 
     def _process(self):
-        approve = request.form['flag'] == '1'
         for registration in self.registrations:
-            _modify_registration_status(registration, approve)
-        flash(_("The status of the selected registrations was updated successfully."), 'success')
+            _modify_registration_status(registration, approve=True)
+        flash(_("The selected registrations was successfully '{}'.").format('approved'), 'success')
```
ThiefMaster, line 653: Feel free to make the string more clear ("The selected registrations were successfully approved" or something like that), but do not break i18n like this particular change does.

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -636,17 +644,31 @@ def _process(self):
...
         return jsonify_data(**self.list_generator.render_list())
 
 
+class RHRegistrationsReject(RHRegistrationsActionBase):
+    """Reject selected registrations from registration list."""
+
+    def _process(self):
+        form = RejectRegistrantsForm(obj=FormDefaults(registration_id=request.args.getlist('registration_id')))
+        message = _("Rejecting the selected registrations will trigger a notification email for each registrant.")
```
ThiefMaster, line 662:
```suggestion
        message = _("Rejecting these registrations will trigger a notification email for each registrant.")
```

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -636,17 +644,31 @@ def _process(self):
...
+            _modify_registration_status(registration, approve=True)
+        flash(_("The selected registrations was successfully '{}'.").format('approved'), 'success')
         return jsonify_data(**self.list_generator.render_list())
 
 
+class RHRegistrationsReject(RHRegistrationsActionBase):
+    """Reject selected registrations from registration list."""
+
+    def _process(self):
+        form = RejectRegistrantsForm(obj=FormDefaults(registration_id=request.args.getlist('registration_id')))
```
ThiefMaster, line 661: I think this works as well:

```suggestion
        form = RejectRegistrantsForm(registration_id=request.args.getlist('registration_id'))
```

At least that's what we do e.g. in `BulkAbstractJudgmentForm` which works very similarly (form opened with a list of selected items).

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -636,17 +644,31 @@ def _process(self):
...
+        message = _("Rejecting the selected registrations will trigger a notification email for each registrant.")
+        if form.validate_on_submit():
+            for registration in self.registrations:
+                registration.rejection_reason = form.rejection_reason.data
+                _modify_registration_status(registration, approve=False)
+            flash(_("The selected registrations was successfully '{}'.").format('rejected'), 'success')
```
ThiefMaster, line 667: same here, don't use format on translated strings when the interpolated data is an english word (which needs to be translated btw) as well:

```suggestion
            flash(_("The selected registrations were successfully rejected."), 'success')
```

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -636,17 +644,31 @@ def _process(self):
...
+            return jsonify_data(**self.list_generator.render_list())
+        return jsonify_form(form, disabled_until_change=False, submit='Reject', message=message)
```
ThiefMaster, line 669:
```suggestion
        return jsonify_form(form, disabled_until_change=False, submit=_('Reject'), message=message)
```

indico/modules/events/registration/controllers/management/reglists.py
ThiefMaster: Also, this should be `request.form`, not `request.args`. Use `data-method="POST"` in the HTML to load the form with POST as well.

To know whether the form was actually submitted or not, add this code:

```python
    submitted = HiddenField()

    ...

    def is_submitted(self):
```

You can have a look at `BulkAbstractJudgmentForm` to see a place where this is already done as well :)

indico/modules/events/registration/templates/management/regform_reglist.html
```diff
@@ -150,19 +150,19 @@
...
                                    data-method="POST"
                                    data-confirm="{% trans %}Do you really want to approve the selected registrations? This will trigger a notification email for each registrant.{% endtrans %}">
                                     {%- trans %}Approve registrations{% endtrans -%}
                                 </a>
                             </li>
                             <li>
-                                <a href="#" class="icon-close js-requires-selected-row disabled js-modify-status"
-                                   data-href="{{ url_for('.registrations_modify_status', regform) }}"
-                                   data-flag="0"
-                                   data-method="POST"
```
ThiefMaster, line 164: we want to keep POST here - see my comment above about `request.form`

(selecting many registrations without POST would result in a VERY long query string otherwise)

ThiefMaster (changes requested):

indico/modules/events/registration/notifications.py
```diff
@@ -63,6 +64,8 @@ def notify_registration_modification(registration, notify_user=True):
 
 
 def notify_registration_state_update(registration):
-    _notify_registration(registration, 'registration_state_update_to_registrant.html')
+    attach_rejection_reason = request.form.get('attach_rejection_reason') == 'y'
```
ThiefMaster, line 67: This should be passed as an argument as weld - the data should be coming from `form.attach_rejection_reason.data`, and not from manually reading the form data again..

indico/modules/events/registration/templates/emails/registration_creation_to_registrant.html
```diff
@@ -49,6 +49,13 @@
 {% endmacro %}
 
 
+{% macro render_rejection_reason() %}
+    {% if registration.state.name == 'rejected' and registration.rejection_reason %}
+    	<br>Rejection reason: {{ registration.rejection_reason }}.<br>
```
ThiefMaster, line 54: please keep the indentation consistent (4 spaces) :)

indico/modules/events/registration/templates/emails/registration_creation_to_managers.html
```diff
@@ -31,6 +31,13 @@
 {%- endmacro %}
 
 
+{% macro render_rejection_reason() %}
+    {% if registration.state.name == 'rejected' and registration.rejection_reason %}
+    	<br>Rejection reason: {{ registration.rejection_reason }}.<br>
```
ThiefMaster, line 36: please keep the indentation consistent (4 spaces)

indico/modules/events/registration/templates/emails/registration_state_update_to_registrant.html
```diff
@@ -8,6 +8,7 @@
 {% block registration_header_text -%}
     Your registration {{ render_registration_info() }}
     is now <strong>{{ registration.state.title|lower }}</strong>.
+    {{ render_rejection_reason() if attach_rejection_reason  }}
```
ThiefMaster, line 11:
```suggestion
    {{ render_rejection_reason() if attach_rejection_reason }}
```

ThiefMaster (changes requested):

indico/modules/events/registration/templates/emails/registration_creation_to_managers.html
```diff
@@ -31,6 +31,13 @@
 {%- endmacro %}
 
 
+{% macro render_rejection_reason() %}
+    {% if registration.state.name == 'rejected' and registration.rejection_reason %}
+        <br>Rejection reason: {{ registration.rejection_reason }}.<br>
```
ThiefMaster, line 36:
```suggestion
        <br>Rejection reason: {{ registration.rejection_reason }}<br>
```

indico/modules/events/registration/templates/emails/registration_creation_to_registrant.html
```diff
@@ -49,6 +49,13 @@
 {% endmacro %}
 
 
+{% macro render_rejection_reason() %}
+    {% if registration.state.name == 'rejected' and registration.rejection_reason %}
+        <br>Rejection reason: {{ registration.rejection_reason }}.<br>
```
ThiefMaster, line 54:
```suggestion
        <br>Rejection reason: {{ registration.rejection_reason }}<br>
```

indico/modules/events/registration/forms.py
```diff
@@ -407,3 +407,14 @@ def __init__(self, *args, **kwargs):
         super(ImportRegistrationsForm, self).__init__(*args, **kwargs)
         if not self.regform.moderation_enabled:
             del self.skip_moderation
+
+
+class RejectRegistrantsForm(IndicoForm):
+    rejection_reason = TextAreaField(_('Reason'), description=_("You can provide a reason for the rejection here."))
+    attach_rejection_reason = BooleanField(_('Attach reason'), [HiddenUnless('rejection_reason')],
```
ThiefMaster, line 414: I would remove the HiddenUnless. It only triggers when the field loses focus, so its UX is not very good right now.

Just make sure that we never show an empty reason in a notification email when this is enabled but no reason is specified ;)

ThiefMaster:
Another small improvement would be displaying the rejection reason - feel free to just apply this patch:

```patch
diff --git a/indico/modules/events/registration/templates/management/_registration_details.html b/indico/modules/events/registration/templates/management/_registration_details.html
index 5dd26d1b75..deb5dd669d 100644
--- a/indico/modules/events/registration/templates/management/_registration_details.html
+++ b/indico/modules/events/registration/templates/management/_registration_details.html
@@ -183,6 +183,9 @@
                 This registration is {{ state }}
             {%- endtrans %}
         </div>
+        {% if registration.state.name == 'rejected' and registration.rejection_reason %}
+            {% trans reason=registration.rejection_reason %}Reason: {{ reason }}{% endtrans %}
+        {% endif %}
     </div>
     <div class="toolbar hide-if-locked">
         {% set action = _('Reset rejection') if registration.state.name == 'rejected' else _('Reset withdrawal') %}
```

indico/modules/events/registration/forms.py
ThiefMaster: I think my second point is already covered since `render_rejection_reason` doesn't render anything if there is no reason.

ThiefMaster (approved):
PR looks and works fine now, leaving it open for now until we decided whether this is for 2.3 or rather 3.0 due to the DB change.

vasantvohra:
Thanks

> PR looks and works fine now, leaving it open for now until we decided whether this is for 2.3 or rather 3.0 due to the DB change.

Thanks a ton @ThiefMaster

PR #4771, opened 2021-02-03
Replace GenericCache and get rid of non-redis cache backends

### LINKED ISSUE #4681: Replace GenericCache
There are extensions such as flask-caching which do pretty much the same thing (just without the extra level of namespacing the argument passed to `GenericCache(...)` provides, but this could be easily worked around since flask-caching has a `key_prefix`). #4536 already added this particular extension.

We could also just use redis directly since everyone uses it as a cache backend in Indico anyway...

ThiefMaster:
- Removed the legacy GenericCache and replaced it with a lightweight wrapper around Flask-Caching
- The cache prefix now also contains the major/minor indico version number, so any cached data goes away after significant upgrades.

closes #4681


javfg (approved):
Don't see anything out of place. It's been useful for as an first contact with Indico codebase.

PR #4772, opened 2021-02-04
Be more forgiving with invalid locales

ThiefMaster:
See this forum thread for some more details: https://talk.getindico.io/t/localization-en-us-issue-in-2-3-3/2104

Basically having an invalid locale stored in the session/user currently results in a 404 error (instead of silently falling back to a minimal valid file) when trying to retrieve the locale information for JS. However, not having the locale data completely breaks our i18n JS, which in turns breaks pretty much all JS we have.

With this change we now simply redirect to the default locale (if it's valid, otherwise to the known-valid `en_GB` locale) instead of failing with a 404.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,13 @@ Version 2.3.4
 
 *Unreleased*
 
+Improvements
+^^^^^^^^^^^^
+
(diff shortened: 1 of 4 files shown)


plourenco (approved):
👍

Issue #4776, opened 2021-02-09
Use signed URL instead of legacy API for persistent ical links


ThiefMaster:
We have this widget in various places within categories and events to get persistent URLs to an iCAL file:

![image](https://user-images.githubusercontent.com/179599/107376972-c54d4880-6aea-11eb-94f1-7a6f9226e6a8.png) ![image](https://user-images.githubusercontent.com/179599/107377290-165d3c80-6aeb-11eb-8c0e-5150a4da92f9.png)

It is extremely legacy (see the code in `Export.js`) and using the (legacy) HTTPAPI for this is not ideal - let's use the signed URL utility we already use in the dashboard instead:

![image](https://user-images.githubusercontent.com/179599/107377430-40166380-6aeb-11eb-8afb-4d5a39467912.png)

One thing to keep in mind is that we need to include the userid in the signature or the URL, since downloading the ical file needs to be tied to the user who created the token of course, since generally there won't be a valid session cookie when accessing the URL with the token.

We can also reconsider whether the persistent link actually makes sense in some of the places. For categories it's useful for sure, for events it *may* also be useful, but for sessions/contributions it's probably not really relevant.

PR #4777, opened 2021-02-09
Remove semi-broken persistent ical links for contributions

ThiefMaster:
- They pointed to the event ical (without any contribution details)
- When enabling (persistent) API keys from that widget, the URLs generated pointed to a legacy contribution-only ical api which 503s since v1.9.x
- Persistent calendar links for a single contribution are not particularly useful

When we improve the ical links in all the other places (see #4776), we can consider doing the same for contributions for the sake if being consistent, i.e. having the option to generate a persistent link in any place where we also have a regular ical link.

Nothing changed in Export.js btw; I just thought it may need some changes so I ran eslint on it before, and didn't feel like reverting those in the end, even though I seriously hope this file goes away for good with #4776!

---

image

I did not bother removing the dropdown in favor of a direct link, since there's a very good chance that the issue mentioned above will land in 3.0 anyway, and I didn't feel like improving UI that won't make it into the final release of this version.

### DIFF
--- a/indico/web/templates/_ical_export.html
+++ b/indico/web/templates/_ical_export.html
@@ -8,6 +8,7 @@
 
     {% block extra_download %}{% endblock %}
 
+    {% block persistent_ui %}
     <div id="iCalSeparator{{ item.id }}" class="icalSeparator" style="display: none;"></div>
     <div id="exportICalDialogs" style="display: none;">
         <div id="agreementApiKey{{ item.id }}">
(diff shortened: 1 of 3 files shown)


indico/web/templates/_ical_export.html
```diff
@@ -8,6 +8,7 @@
 
     {% block extra_download %}{% endblock %}
 
+    {% block persistent_ui %}
```
ThiefMaster, line 11: I did not indent the contents of this rather big block on purpose since this template or at least most of its content should become obsolete soon anyway.

javfg (approved):
:+1:

PR #4778, opened 2021-02-11
Replace OrderedDict instances with dicts (#4703)

### LINKED ISSUE #4703: Replace OrderedDict with regular dicts
Since Python 3.7 dicts are guaranteed to preserve insertion order - ie same behavior as OrderedDict. Let's get rid of the latter.

Check 3b13a91447 for an example how it's done in `ListGenerator` subclasses; in most other places it should be even easier. Just make sure to test the updated code to ensure things work fine.

Note: In case someone who's not an Indico core developer wants to contribute here: If you pick straightforward cases and don't want to setup a local indico dev environment, feel free to send a PR even without having actually tested it; just let us know so we can test more thoroughly in that case.

errikos:
Fixes #4703. Changes:
* Replace all `OrderedDict` instances with normal `dict`s.
* Remove the `OrderedDictConverter` class.

I have also removed the `OrderedDict` constructor calls where unnecessary, but I may have missed some spots (hopefully not).

`pytest` looks OK, but I haven't tested every single change on the locally deployed site.

I can also send a PR for `plugins-cern` if we need the same changes there too.

### DIFF
--- a/indico/modules/events/contributions/lists.py
+++ b/indico/modules/events/contributions/lists.py
@@ -36,18 +36,16 @@ class ContributionListGenerator(ListGeneratorBase):
         session_empty = {None: _('No session')}
         track_empty = {None: _('No track')}
         type_empty = {None: _('No type')}
-        session_choices = {str(s.id): s.title for s in sorted(self.event.sessions,
-                                                              key=attrgetter('title'))}
+        session_choices = {str(s.id): s.title for s in sorted(self.event.sessions, key=attrgetter('title'))}
         track_choices = {str(t.id): t.title for t in sorted(self.event.tracks, key=attrgetter('title'))}
-        type_choices = {str(t.id): t.name for t in sorted(self.event.contribution_types,
-                                                          key=attrgetter('name'))}
+        type_choices = {str(t.id): t.name for t in sorted(self.event.contribution_types, key=attrgetter('name'))}
         self.static_items = {
             'session': {'title': _('Session'),
-                        'filter_choices': dict(session_empty | session_choices)},
+                        'filter_choices': session_empty | session_choices},
             'track': {'title': _('Track'),
-                      'filter_choices': dict(track_empty | track_choices)},
+                      'filter_choices': track_empty | track_choices},
             'type': {'title': _('Type'),
-                     'filter_choices': dict(type_empty | type_choices)},
+                     'filter_choices': type_empty | type_choices},
             'status': {'title': _('Status'), 'filter_choices': {'scheduled': _('Scheduled'),
                                                                 'unscheduled': _('Not scheduled')}},
             'speakers': {'title': _('Speakers'), 'filter_choices': {'registered': _('Registered'),
--- a/indico/modules/events/contributions/lists.py
+++ b/indico/modules/events/contributions/lists.py
@@ -5,7 +5,6 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
-from collections import OrderedDict
 from datetime import timedelta
 from operator import attrgetter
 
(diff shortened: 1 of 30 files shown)


ThiefMaster:
Thanks a lot, I'll test the changes.

> I can also send a PR for plugins-cern if we need the same changes there too.

That would be great! With all dicts being ordered in Python now, there's no need to keep ODs around.

indico/modules/events/contributions/lists.py
```diff
@@ -37,23 +36,23 @@ def __init__(self, event):
...
-            ('status', {'title': _('Status'), 'filter_choices': {'scheduled': _('Scheduled'),
-                                                                 'unscheduled': _('Not scheduled')}}),
-            ('speakers', {'title': _('Speakers'), 'filter_choices': {'registered': _('Registered'),
-                                                                     'not_registered': _('Not registered')}})
-        ])
+        session_choices = {str(s.id): s.title for s in sorted(self.event.sessions,
+                                                              key=attrgetter('title'))}
+        track_choices = {str(t.id): t.title for t in sorted(self.event.tracks, key=attrgetter('title'))}
+        type_choices = {str(t.id): t.name for t in sorted(self.event.contribution_types,
+                                                          key=attrgetter('name'))}
```
ThiefMaster, line 43: I think each of those statements may fit in one line without the need for wrapping before `key=` nor exceeding our usual 120 char max line length

indico/modules/events/contributions/lists.py
```diff
@@ -37,23 +36,23 @@ def __init__(self, event):
...
+        self.static_items = {
+            'session': {'title': _('Session'),
+                        'filter_choices': dict(session_empty | session_choices)},
```
ThiefMaster, line 46: You can omit the `dict()` around it - `dict | dict` already returns a dict.

indico/modules/events/papers/lists.py
```diff
@@ -35,26 +34,26 @@ def __init__(self, event):
...
-                                                                           key=attrgetter('title')))
-        type_choices = OrderedDict((str(t.id), t.name) for t in sorted(self.event.contribution_types,
-                                                                       key=attrgetter('name')))
+        state_choices = {state.value: state.title for state in PaperRevisionState}
+        unassigned_choices = {role.value: role.title for role in PaperReviewingRole}
+        track_choices = {str(t.id): t.title for t in sorted(self.event.tracks, key=attrgetter('title'))}
+        session_choices = {str(s.id): s.title for s in sorted(self.event.sessions,
+                                                              key=attrgetter('title'))}
+        type_choices = {str(t.id): t.name for t in sorted(self.event.contribution_types,
+                                                          key=attrgetter('name'))}
```
ThiefMaster, line 43: These statements may fit in a single line as well

indico/modules/events/papers/lists.py
```diff
@@ -35,26 +34,26 @@ def __init__(self, event):
...
 
-        self.static_items = OrderedDict([
-            ('state', {'title': _('State'), 'filter_choices': OrderedDict(state_not_submitted | state_choices)}),
-            ('track', {'title': _('Track'), 'filter_choices': OrderedDict(track_empty | track_choices)}),
-            ('session', {'title': _('Session'), 'filter_choices': OrderedDict(session_empty | session_choices)}),
-            ('type', {'title': _('Type'), 'filter_choices': OrderedDict(type_empty | type_choices)}),
-            ('unassigned', {'title': _('Unassigned'), 'filter_choices': unassigned_choices}),
-        ])
+        self.static_items = {
+            'state': {'title': _('State'), 'filter_choices': dict(state_not_submitted | state_choices)},
```
ThiefMaster, line 51: no need for the `dict()` around it

indico/modules/events/papers/lists.py
```diff
@@ -186,7 +185,7 @@ def __init__(self, event, user):
         }
         judging_unassigned_choices = {role.value: role.title for role in PaperReviewingRole
                                       if role is not PaperReviewingRole.judge}
-        self.static_items['unassigned']['filter_choices'] = OrderedDict(sorted(judging_unassigned_choices.items()))
+        self.static_items['unassigned']['filter_choices'] = judging_unassigned_choices
```
ThiefMaster, line 188: This can be simplified into a single statement (and also sort by displayed name, not the internal enum member name):

```python
        self.static_items['unassigned']['filter_choices'] = {
            role.value: role.title
            for role in sorted(PaperReviewingRole, key=attrgetter('title'))
            if role is not PaperReviewingRole.judge
        }
```

indico/modules/events/persons/controllers.py
```diff
@@ -162,12 +162,12 @@ def get_persons(self):
             event_user_roles_data = {}
             for role in event_user_roles[event_person.user]:
                 event_user_roles_data[f'custom_{role.id}'] = {'name': role.name, 'code': role.code, 'css': role.css}
-            event_user_roles_data = OrderedDict(sorted(list(event_user_roles_data.items()), key=lambda t: t[1]['code']))
-            data['roles'] = OrderedDict(data['roles'] | event_user_roles_data)
+            event_user_roles_data = dict(sorted(list(event_user_roles_data.items()), key=lambda t: t[1]['code']))
```
ThiefMaster, line 165: the `list()` isn't needed

indico/modules/events/persons/controllers.py
```diff
@@ -162,12 +162,12 @@ def get_persons(self):
...
+            data['roles'] = dict(data['roles'] | event_user_roles_data)
```
ThiefMaster, line 166: no need for `dict()` around it

indico/modules/users/util.py
```diff
@@ -77,7 +76,7 @@ def get_related_categories(user, detailed=True):
             'managed': categ in managed,
             'path': truncate_path(categ.chain_titles[:-1], chars=50)
         }
-    return OrderedDict(sorted(list(res.items()), key=itemgetter(0)))
+    return dict(sorted(list(res.items()), key=itemgetter(0)))
```
ThiefMaster, line 79:
```suggestion
    return dict(sorted(res.items(), key=itemgetter(0)))
```

indico/modules/events/registration/fields/simple.py
```diff
@@ -212,7 +211,7 @@ def unprocess_field_data(cls, versioned_data, unversioned_data):
 
     @property
     def filter_choices(self):
-        return OrderedDict(self.wtf_field_kwargs['choices'])
+        return self.wtf_field_kwargs['choices']
```
ThiefMaster, line 214: This fixes the error I mentioned on IRC:

```suggestion
        return dict(self.wtf_field_kwargs['choices'])
```

ThiefMaster (approved):
thanks for the nice PR!

PR #4796, opened 2021-02-18
Avoid querying active registrations if not needed

OmeGak:


indico/modules/events/registration/fields/choices.py
```diff
@@ -128,6 +128,8 @@ def process_field_data(cls, data, old_data=None, old_versioned_data=None):
 
     def get_places_used(self):
         places_used = Counter()
+        if not self.form_item.data.get('places_limit'):
```
ThiefMaster, line 131: That won't work here, since the place limit for choices is stored in the choice data within the field's `versioned_data`

ThiefMaster (approved):

PR #4798, opened 2021-02-18
Refactor OAuth provider code and switch from Flask-OAuthlib to Authlib

### LINKED ISSUE #4685: Replace flask-oauthlib with authlib
Probably rewrite the oauth provider and improve it overall?

TODO: Describe how it should be improved, but let's see whether we can do anything from #1934 while we're at it

ThiefMaster:
- [x] Authlib instead of (flask-)oauthlib
- [x] Proper test coverage for all supported oauth flows
- [x] [Authorization code with PKCE](https://www.oauth.com/oauth2-servers/pkce/authorization-code-exchange/) flow instead of implicit flow (**breaking change**)
- [x] Provider metadata endpoint (`/.well-known/oauth-authorization-server`)
- [x] Token introspection endpoint
- [x] Token revocation endpoint
- [x] Allow multiple tokens for a user/app combo (~~maybe group by scope to~~ keep the total number of tokens still limited? grouping is no longer possible due to hashing, keeping up to 50 tokens for now)
- [x] Make PKCE flow opt-in (most apps are server-side and can just use a client secret)
- [x] Store oauth tokens in hashed form, maybe sha256 (no salting is needed for random tokens, and bcrypt etc are slow and thus not a good choice for API tokens)
- [x] Changelog entry

Regarding the breaking change, this needs an update in the checkin app which I think is the only client using the implicit flow right now. An alternative, especially since hello.js which we use in that app does not support PKCE yet (see MrSwitch/hello.js#633), could be enabling the implicit flow but allowing it *only* for the checkin app so we can remove it once everyone is on 3.0 and there is no need to support older Indico versions in the checkin app. (**Update:** The checkin system app is allowed to use the implicit grant and token-in-querystring for now)

closes #4685


javfg (approved):
Looks good to me, but keep in mind my still limited review value at the moment.

PR #4801, opened 2021-02-19
Persistent ical links with signed URL, updated calendar widgets

### LINKED ISSUE #4776: Use signed URL instead of legacy API for persistent ical links

javfg:
Closes #4776.

* There might be different options to download icals that I was not aware of, which could be added to dropdowns in the new widgets.
* The change in Event uses a new module.
* I might have left some unused code around, not sure if I found all the templates/hooks for the old widget.


indico/modules/events/sessions/client/js/session_display.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import {IButton, ICSCalendarLink} from 'indico/react/components';
+import {Translate} from 'indico/react/i18n';
+
+document.addEventListener('DOMContentLoaded', () => {
+  const userId = document.querySelector('body').dataset.userId;
+  const {eventId, sessionId} = document.querySelector('#session-calendar-link').dataset;
```
ThiefMaster, line 16: I would assign the element to a variable instead of querying it here and again below. Also, I think this code should be a bit more resilient against the element not existing (just return early if querying it doesn't return something)

indico/modules/categories/templates/display/base.html
```diff
@@ -42,13 +42,7 @@ <h1 class="category-title {{ category_title_classes }}">
...
                     {% endblock %}
                     {% block cat_export %}
-                        <a href="#"
-                           class="i-button icon-calendar arrow js-export-ical"
-                           title="{% trans %}Export to scheduling tool{% endtrans %}"
-                           data-id="{{ category.id }}"></a>
-                        {% with item=category, ics_url=url_for('categories.export_ical', category) -%}
-                            {% include 'categories/category_export_ical.html' %}
-                        {%- endwith %}
+                        <div id="category-calendar-link" class="calendar-ics-link" data-category-id="{{category.id}}"></div>
```
ThiefMaster, line 45: Is the `calendar-ics-link` class used anywhere?

ThiefMaster, line 45: Usually we put some spaces inside Jinja constructs, ie `{{ category.id }}` (unfortunately there's no linter for that :/)

indico/web/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -41,44 +50,67 @@ export default function ICSCalendarLink({endpoint, urlParams, options, ...restPr
...
+  let trigger;
+
+  if (options.length > 1) {
+    trigger = (
+      <Dropdown
+        styleName="height-full"
+        icon={null}
+        trigger={
+          renderButton ? (
+            renderButton()
```
ThiefMaster, line 70: This doesn't trigger any warnings/errors since we end up doing `{...props}` with props being `undefined`?

indico/modules/categories/templates/display/base.html
javfg: You are right, it is also unused in the original dashboard template I fetched this from.

indico/web/client/js/react/components/ICSCalendarLink.jsx
javfg: It should not, React always returns an empty object even if props are undefined: https://codepen.io/javfg/pen/jOVGGNv

ThiefMaster: I was more worried about ths jsx unpacking, ie this:

```jsx
const foo = undefined;
const test = <Bar {...foo} />;
```

javfg: Indeed. While I was at it, I removed the spread of props and just passed onClick, as that is considered cleaner.

indico/web/client/js/react/components/IButton.jsx
```diff
@@ -31,16 +32,31 @@ export default class IButton extends React.PureComponent {
...
+    const {
+      classes,
+      disabled,
+      highlight,
+      href,
+      title,
+      children,
+      onClick,
+      icon,
+      dropDown,
```
ThiefMaster, line 48: Isn't `dropdown` more common?

CHANGES.rst
```diff
@@ -30,6 +30,9 @@ Improvements
   event (:issue:`4750`, :pr:`4760`)
 - Add options for attaching iCal files to complete registration and event reminder
   emails (:issue:`1158`, :pr:`4780`)
+- Using new signed URL instead of legacy API for persistent ical links, replaced widgets
+  in category, event, session and contribution views with the modern one used in
+  dashboard. (:issue:`4776`, :pr:`4801`)
```
ThiefMaster, line 35:
```suggestion
- Use the new token-based URLs instead of API keys for persistent ical links and replace
  the calendar link widgets in category, event, session and contribution views with the
  more modern ones used in dashboard (:issue:`4776`, :pr:`4801`)
```

ThiefMaster (commented):
Very nice PR. Just two minor suggestions!

indico/modules/core/controllers.py
```diff
@@ -277,15 +277,32 @@ def _process_args(self, values):
         self.values = values
 
 
-class RHSignURL(RHProtected):
+class RHSignURL(RH):
+    """Create a persistent signed URL for a user.
+
+    This build a url and adds a signature that authenticates the user without
+    requiring them to have session cookies. It it means for cases where the
```
javfg, line 284:
```suggestion
    requiring them to have session cookies. It is meant for cases where the
```

javfg (commented):
Just a typo in the docstring.

ThiefMaster (approved):

PR #4802, opened 2021-02-25
Add signal for primary email change

openprojects:
As discussed in talk.getindico.io

### DIFF
--- a/indico/core/signals/users.py
+++ b/indico/core/signals/users.py
@@ -49,3 +49,8 @@ preferences for the user preference page. The *sender* is the user for whom the
 preferences page is being shown which might not be the currently logged-in
 user!
 """)
+
+primary_email_changed = _signals.signal('primary-email-changed', """
+Called when the primary address is changed. The *sender* is
+the user object and the `new` and `old` values are passed in the kwarg.
+""")
--- a/indico/core/signals/users.py
+++ b/indico/core/signals/users.py
@@ -52,5 +52,5 @@ user!
 
 primary_email_changed = _signals.signal('primary-email-changed', """
 Called when the primary address is changed. The *sender* is
-the user object and the `new` and `old` values are passed in the kwarg.
+the user object and the `new` and `old` values are passed as kwargs.
 """)
--- a/indico/modules/users/models/users.py
+++ b/indico/modules/users/models/users.py
@@ -622,12 +622,12 @@ class User(PersonMixin, db.Model):
         secondary = next((x for x in self._secondary_emails if x.email == email), None)
         if secondary is None:
             raise ValueError('email is not a secondary email address')
+        old = self.email
         self._primary_email.is_primary = False
         db.session.flush()
         secondary.is_primary = True
         db.session.flush()
-        kwargs = dict(old=self._primary_email.email, new=secondary.email)
-        signals.users.primary_email_changed.send(self, **kwargs)
+        signals.users.primary_email_changed.send(self, old=old, new=self.email)
 
     def reset_signing_secret(self):
         self.signing_secret = unicode(uuid4())
--- a/indico/modules/users/models/users.py
+++ b/indico/modules/users/models/users.py
@@ -12,7 +12,14 @@ from enum import Enum
 from operator import attrgetter
 from uuid import uuid4
 
+from flask import flash, g, has_request_context, session
 from flask_multipass import IdentityRetrievalFailed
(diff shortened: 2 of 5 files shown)


indico/modules/users/models/users.py
```diff
@@ -624,6 +626,8 @@ def make_email_primary(self, email):
         db.session.flush()
         secondary.is_primary = True
         db.session.flush()
+        kwargs = dict(old=self._primary_email.email, new=secondary.email)
+        signals.users.primary_email_changed.send(self, **kwargs)
```
ThiefMaster, line 630:
```suggestion
        signals.users.primary_email_changed.send(self, old=self._primary_email.email, new=secondary.email)
```

indico/modules/users/models/users.py
```diff
@@ -12,15 +12,9 @@
...
 from uuid import uuid4
 
-from flask import flash, g, has_request_context, session
 from flask_multipass import IdentityRetrievalFailed
-from sqlalchemy.dialects.postgresql import JSONB, UUID
-from sqlalchemy.event import listens_for
-from sqlalchemy.ext.associationproxy import association_proxy
-from sqlalchemy.ext.hybrid import hybrid_property
-from sqlalchemy.orm import object_session
-from werkzeug.utils import cached_property
```
ThiefMaster, line 22: Please undo this

indico/modules/users/models/users.py
ThiefMaster: Also getting the old value here seems to be a bit fragile. Why not set `old = self.email` before changing it and use `email` for new?

indico/core/signals/users.py
```diff
@@ -49,3 +49,8 @@
 preferences page is being shown which might not be the currently logged-in
 user!
 """)
+
+primary_email_changed = _signals.signal('primary-email-changed', """
+Called when the primary address is changed. The *sender* is
+the user object and the `new` and `old` values are passed in the kwarg.
```
ThiefMaster, line 55:
```suggestion
the user object and the `new` and `old` values are passed as kwargs.
```

indico/modules/users/models/users.py
openprojects: you mean... to keep the imports, right?

ThiefMaster: keeping them as they are, since the previous sorting was correct (not sure why your isort tries to move them like this)

openprojects: I changed the imports to this:
```
import itertools
from enum import Enum
from operator import attrgetter
from uuid import uuid4

from flask import flash, g, has_request_context, session
from flask_multipass import IdentityRetrievalFailed
from sqlalchemy.dialects.postgresql import JSONB, UUID
from sqlalchemy.event import listens_for
from sqlalchemy.ext.associationproxy import association_proxy
from sqlalchemy.ext.hybrid import hybrid_property
from sqlalchemy.orm import object_session
from werkzeug.utils import cached_property

from indico.core import signals
from indico.core.auth import multipass
from indico.core.db import db
from indico.core.db.sqlalchemy import PyIntEnum
from indico.core.db.sqlalchemy.custom.unaccent import define_unaccented_lowercase_index
from indico.core.db.sqlalchemy.principals import PrincipalType
from indico.core.db.sqlalchemy.util.models import get_default_values
from indico.modules.users.models.affiliations import UserAffiliation
from indico.modules.users.models.emails import UserEmail
from indico.modules.users.models.favorites import favorite_category_table, favorite_user_table
from indico.util.locators import locator_property
from indico.util.string import format_full_name, format_repr, return_ascii
from indico.util.struct.enum import RichIntEnum
```

... but now I cannot commit: the pre-commit checks are blocking me.
I already tried:
git diff --staged --name-only --diff-filter d indico/indico/modules/users/models/users.py | xargs isort

...but nothing changed.
Suggestions?

ThiefMaster: Something must be wrong with your local setup if isort tries to mess up your imports. Maybe the virtualenv is not active? Anyway, `git commit -n` to skip checks is your friend (at least as a workaround).

openprojects: Thank you, it worked

ThiefMaster (approved):
I fixed the logic (`self.email` does *not* contain the new email until the user object gets expired) and added a unit test and changelog entry for this.

PR #4803, opened 2021-02-26
Unify handling of the current user of a request

ThiefMaster:
My approach is to first use the slightly ugly solution where `session.user` returns the current user, and then later (in a separate PR) replace `session.user` with something else.

### DIFF
--- a/indico/modules/events/registration/util.py
+++ b/indico/modules/events/registration/util.py
@@ -413,7 +413,7 @@ def _build_base_registration_info(registration):
         'registrant_id': str(registration.id),
         'checked_in': registration.checked_in,
         'checkin_secret': registration.ticket_uuid,
-        'full_name': '{} {}'.format(personal_data.get('title', ''), registration.full_name),
+        'full_name': '{} {}'.format(personal_data.get('title', ''), registration.full_name).strip(),
         'personal_data': personal_data
     }
 
--- a/indico/web/util.py
+++ b/indico/web/util.py
@@ -296,7 +296,7 @@ def get_oauth_user(scopes):
     try:
         oauth_token = require_oauth.acquire_token(scopes)
     except OAuth2Error as exc:
-        raise BadRequest(f'OAuth error: {exc}')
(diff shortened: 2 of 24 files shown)


indico/web/util.py
```diff
@@ -181,31 +185,84 @@ def url_for_index(_external=False, _anchor=None):
...
+
+    url = url_unparse((
+        '',
+        '',
+        parsed.path,
+        url_encode(sorted(params.items()), sort=True),
+        parsed.fragment
+    ))
+    signer = Signer(user.signing_secret, salt='url-signing')
+    return signer.verify_signature(url.encode(), signature)
```
plourenco, line 211: Don't we need to encode the signature as before? (also applies to other occasions below)
```suggestion
    return signer.verify_signature(url.encode(), signature.encode())
```

plourenco (approved):
Nice, I like the way this is heading 👍

For future reference, we talked about the possibility of keeping the simple assignment `session.user =` instead of the setter, but that behaviour is only meant to be restored later when `session.user` is replaced and setting and reading to the exact same thing.

indico/modules/events/registration/util.py
```diff
@@ -413,7 +413,7 @@ def _build_base_registration_info(registration):
         'registrant_id': str(registration.id),
         'checked_in': registration.checked_in,
         'checkin_secret': registration.ticket_uuid,
-        'full_name': '{} {}'.format(personal_data.get('title', ''), registration.full_name),
+        'full_name': '{} {}'.format(personal_data.get('title', ''), registration.full_name).strip(),
```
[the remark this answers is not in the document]
ThiefMaster: I spotted it while testing and was too lazy to push it directly to master ;)

indico/web/util.py
ThiefMaster: No, it uses `wants_bytes()` internally in most places. The next itsdangerous version will do that for the data arg as well, but it's still a prerelease...

PR #4804, opened 2021-02-26
Room booking: handle errors in room details modal

pferreir:
An error condition coming from the backend was being ignored unless it was a pre-booking collision.

### DIFF
--- a/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
+++ b/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
@@ -77,6 +77,7 @@ class BookingDetails extends React.Component {
     occurrencesVisible: false,
     activeConfirmation: null,
     acceptanceFormVisible: false,
+    error: null,
     preBookingConflicts: {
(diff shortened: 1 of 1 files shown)


indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -529,19 +543,25 @@ class BookingDetails extends React.Component {
...
+            preBookingConflicts: {
+              concurrentBookings: error.data,
+              acceptanceData: data,
+              warningVisible: true,
+            },
+          });
+        } else {
+          this.setState({
+            error,
+          });
```
ThiefMaster, line 563: Do we need to clear `preBookingConflicts` here?

pferreir: As far as I can see, `preBookingConflicts` is reset once the corresponding warning modal is closed. We could reset it just for the sake of being explicit, but since the two state slices never interact, I wouldn't bother...

PR #4805, opened 2021-02-27
UI: add horizontal scroll to body

ParthS007:


indico/web/client/styles/themes/indico.scss
```diff
@@ -53,6 +53,7 @@ $selected-color: $indico-blue !default;
 
 body {
   background: #424242;
+  overflow-x: scroll;
```
ThiefMaster, line 56: I'm pretty sure this results in a permanent horizontal scrollbar. What about `auto` to show it only if one is needed?

ParthS007: I just checked  again, It does not end up creating permanent horizontal scrollbar but I think `auto` would be enough since both `scroll` and `auto` produce same results on my local instance.

ParthS007:
@ThiefMaster Do I need to add anything else?

ThiefMaster:
I haven't tested it yet - once I (or someone else from the team) did we'll let you know ;)

ThiefMaster:
So the change works fine, but I just checked why we do not get horizontal scrolling by default, and it looks like it's semantic-ui that sets `overflow-x: hidden;` globally on `body`.

Maybe we should rather reset it back to `inherit` (to use the default) in `indico/web/client/styles/base/_layout.scss` globally?

ParthS007:
I checked with `inherit` and it works as expected. 👍

ThiefMaster:
Check my comment about the file as well ;)

ThiefMaster (approved):

PR #4810, opened 2021-03-05
Add bulk export for editing data to JSON

### LINKED ISSUE #4767: Bulk JSON dump for Editing / Paper Peer Reviewing modules
In the Abstracts module we have an option to dump everything as JSON (with all the review details included).

Having similar functionality in the Editing (and maybe Peer Reviewing) module(s) as well would be useful.

- [x] In the Editing module we can add the UI for it in the same way as "Download all files"
- [ ] If we want to add this in Peer Reviewing as well (apparently NOT needed for the conference, but best to confirm again), next to the "Download papers" button on the Paper Assignment page would be a good place for it.

(Context: @tgbaron mentioned that vCHEP 2021 (happening in May) needs this (shortly after the conference))

ThiefMaster:
closes #4767

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,7 @@ Improvements
 - Use the new token-based URLs instead of API keys for persistent ical links and replace
   the calendar link widgets in category, event, session and contribution views with the
   more modern ones used in dashboard (:issue:`4776`, :pr:`4801`)
+- Add an option to export editables to JSON (:issue:`4767`, :pr:`4810`)
 
 Internal Changes
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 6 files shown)


plourenco (approved):
Everything seems fine!

Not sure for how long the 

```
f for f in EditableSchema.Meta.fields if not f.startswith('can_')
```

is going to hold for, if somehow the `EditableSchema` is changed and something may go in here. But I'm guessing you also assumed it's fine for now.

ThiefMaster:
I think if we add new attributes to that schema there's a good chance they are worth including in the export as well.

I'm just excluding those `can_*` attributes because anything about the current user's permissions is completely useless when exporting data to use outside Indico.

PR #4812, opened 2021-03-09
Fix editable access for editors in case of unpublished contributions

### LINKED ISSUE #4811: Editing Team isn't allowed to see unpublished contributions
When a user part of the editing team tries to access a contribution he's not part of, the following error will come up:
`The contributions of this event have not been published yet.`

Checking that specific controller RHContributionEditableBase, we call RHContributionDisplayBase which as:
```
if (not published and not self.event.can_manage(session.user)
                and not self.contrib.is_user_associated(session.user)):
    raise NotFound(_("The contributions of this event have not been published yet."))
```

`can_manage` doesn't include reviewing managers or paper editors. On the other hand, the editing team is also unable to see a contribution under `RHContributionDisplayBase`.

ThiefMaster:
fixes #4811

I did not enable access to the full contribution itself, since that shouldn't be needed when someone is just an Editor. In any case, this is pretty much an edge case, since Editing usually happens *after* an event, when everything is published.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,7 @@ Bugfixes
   no longer installed
 - Do not auto-redirect to SSO when an MS office user agent is detected (:issue:`4720`,
   :pr:`4731`)
+- Allow Editing team to view editables of unpublished contributions (:issue:`4811`, :pr:`4812`)
 
 Internal Changes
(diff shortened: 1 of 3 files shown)


plourenco (approved):
Looks good, thanks!

PR #4822, opened 2021-03-15
Add option to show only unregistered users in participant roles list

### LINKED ISSUE #4763: Participant roles: filter view by (un)registered status
Right now there is no way to get a list, for instance, of all authors who have not registered yet.

An idea would be to add an "Unregistered users" option in here:

image

ThiefMaster:
closes #4763

### DIFF
--- a/indico/modules/events/persons/templates/management/person_list.html
+++ b/indico/modules/events/persons/templates/management/person_list.html
@@ -176,6 +176,15 @@
                             <i class="icon-stop colored-square no-account"></i>
                             <label for="filter-no-account">{% trans %}Users with no account{% endtrans %}</label>
                         </li>
+                        <li class="separator">
+                            <span class="titled-rule">{% trans %}or{% endtrans %}</span>
+                        </li>
+                        <li for="filter-no-registration">
+                            <input type="checkbox" id="filter-no-registration" data-filter="no_registration">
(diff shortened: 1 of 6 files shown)


indico/modules/events/persons/templates/management/person_list.html
```diff
@@ -176,6 +176,15 @@
...
                             <label for="filter-no-account">{% trans %}Users with no account{% endtrans %}</label>
                         </li>
+                        <li class="separator">
+                            <span class="titled-rule">{% trans %}or{% endtrans %}</span>
+                        </li>
+                        <li for="filter-no-registration">
+                            <input type="checkbox" id="filter-no-registration" data-filter="no_registration">
+                            <i class="icon-checkmark"></i>
+                            <i class="icon-stop colored-square no-registration"></i>
+                            <label for="filter-no-registration">{% trans %}Users who didn't register{% endtrans %}</label>
```
plourenco, line 186: Perhaps `Unregistered users` would better fit the dialog?

plourenco (approved):
👍

indico/modules/events/persons/templates/management/person_list.html
ThiefMaster: I tried to stay consistent with "Users with no account", but "Unregistered users" sounds good as well

PR #4823, opened 2021-03-19
Registration CSV import: Check for unique users as well

ThiefMaster:
We check for [unique emails](https://github.com/indico/indico/blob/aeee94f34f07d745802c5ae1c6f422f1eed4adbd/indico/modules/events/registration/util.py#L576-L577) but if two emails belong to the same user the import still fails with an ugly database error.

### DIFF
--- a/indico/modules/events/registration/util.py
+++ b/indico/modules/events/registration/util.py
@@ -553,9 +553,14 @@ def import_registrations_from_csv(regform, fileobj, skip_moderation=True, notify
     """Import event registrants from a CSV file into a form."""
     with csv_text_io_wrapper(fileobj) as ftxt:
         reader = csv.reader(ftxt.read().splitlines())
-    query = db.session.query(Registration.email).with_parent(regform).filter(Registration.is_active)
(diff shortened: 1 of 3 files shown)


indico/modules/events/registration/util.py
```diff
@@ -573,10 +578,20 @@ def import_registrations_from_csv(regform, fileobj, skip_moderation=True, notify
             raise UserValueError(_('Row {}: missing first or last name').format(row_num))
         if email in registered_emails:
             raise UserValueError(_('Row {}: a registration with this email already exists').format(row_num))
+
+        user = get_user_by_email(email)
+        if user and user.id in registered_user_ids:
+            raise UserValueError(_('Row {}: a registration for this user already exists').format(row_num))
         if email in used_emails:
             raise UserValueError(_('Row {}: email address is not unique').format(row_num))
+        if conflict_row_num := email_row_map.get(email):
```
pferreir, line 587: 😲 shiny new assignment operator!

pferreir (approved):

PR #4842, opened 2021-03-25
Update the search controller into a single endpoint

plourenco:
Eventually, this controller would also receive multiple search targets (normally all of them) and return a mixed result set such as: `Event, Event, Contribution, Event, Attachment`.

### DIFF
--- a/indico/modules/search/client/js/components/SearchApp.jsx
+++ b/indico/modules/search/client/js/components/SearchApp.jsx
@@ -8,9 +8,9 @@
 import searchURL from 'indico-url:search.api_search';
 
 import PropTypes from 'prop-types';
-import React, {useEffect, useState, useMemo, useCallback} from 'react';
+import React, {useCallback, useEffect, useMemo, useState} from 'react';
 import {useHistory} from 'react-router-dom';
-import {Loader, Menu, Grid} from 'semantic-ui-react';
+import {Grid, Loader, Menu} from 'semantic-ui-react';
 
 import {useIndicoAxios} from 'indico/react/hooks';
 import {Translate} from 'indico/react/i18n';
@@ -112,9 +112,7 @@ function useQueryParams() {
   );
 
   useEffect(() => {
-    history.listen(location => {
-      _setQuery(location.search);
-    });
+    return history.listen(location => _setQuery(location.search));
   }, [history, query, _setQuery]);
 
   return [queryObject, setQuery];
@@ -123,10 +121,6 @@ function useQueryParams() {
 export default function SearchApp() {
   const [query, setQuery] = useQueryParams();
   const [activeMenuItem, setActiveMenuItem] = useState(undefined);
-  const handleClick = (e, {index}) => {
-    setActiveMenuItem(index);
-  };
-
   const {q, ...filters} = query;
   const [categoryResults, setCategoryPage] = useSearch(searchURL({type: 'category'}), query);
   const [eventResults, setEventPage] = useSearch(searchURL({type: 'event'}), query);
@@ -148,10 +142,7 @@ export default function SearchApp() {
   const [_, results, setPage, Component] = searchMap[menuItem];
   const isAnyLoading = searchMap.some(x => x[1].loading);
 
-  const handleQuery = (value, type) => {
-    const _type = type || 'q';
-    setQuery(_type, value, _type === 'q');
-  };
+  const handleQuery = (value, type = 'q') => setQuery(type, value, type === 'q');
 
   return (
     <Grid padded>
@@ -172,7 +163,7 @@ export default function SearchApp() {
                   title={Translate.string(_label)}
                   total={_results.total}
                   loading={_results.loading}
-                  onClick={handleClick}
+                  onClick={(e, {index}) => setActiveMenuItem(index)}
                 />
               ))}
             </Menu>
--- a/indico/modules/search/client/js/components/SearchApp.jsx
+++ b/indico/modules/search/client/js/components/SearchApp.jsx
@@ -5,15 +5,12 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
-import searchAttachmentURL from 'indico-url:search.api_search_attachment';
-import searchCategoryURL from 'indico-url:search.api_search_category';
-import searchContributionURL from 'indico-url:search.api_search_contribution';
-import searchEventURL from 'indico-url:search.api_search_event';
+import searchURL from 'indico-url:search.api_search';
 
 import PropTypes from 'prop-types';
-import React, {useEffect, useReducer, useState, useMemo} from 'react';
+import React, {useEffect, useState, useMemo, useCallback} from 'react';
+import {useHistory} from 'react-router-dom';
 import {Loader, Menu, Grid} from 'semantic-ui-react';
-import {useQueryParam, StringParam} from 'use-query-params';
 
 import {useIndicoAxios} from 'indico/react/hooks';
 import {Translate} from 'indico/react/i18n';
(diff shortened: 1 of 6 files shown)


indico/modules/search/client/js/components/SideBar.jsx
```diff
@@ -6,130 +6,84 @@
...
-          </Table.Row>
-        </Table.Header>
+function AggregationList({name, title, items, query, onChange}) {
+  const _items = items
+    .filter(b => b.key)
+    .map(({key, fromAsString: from, toAsString: to, docCount: count}) => ({
+      key: from || to ? `[${from || '*'} TO ${to || '*'}]` : key,
+      label: key.charAt(0).toUpperCase() + key.slice(1),
+      count,
+    }));
```
plourenco, line 44: A note here: While most aggregations that come from elastic search are simple composed of the attributes: `key` and `docCount`, the recent case of dates introduced the concept of ranges. In range aggregations, there are two other attributes in the following form:
```
{
    "key": "1 year ago"
    "docCount": 30,
    "from": <timestamp>
    "fromAsString": "2020-03-25"
    "to": <timestamp>
    "toAsString": "*"
}
```

This mapping is simply checking whether or not our key is in fact a range or a keyword to go as query params in the `onChange` below.

indico/modules/search/controllers.py
```diff
@@ -5,60 +5,77 @@
...
 from indico.modules.search.base import IndicoSearchProvider, SearchTarget, get_search_provider
-from indico.modules.search.schemas import CategoryResultSchema, EventResultSchema
+from indico.modules.search.schemas import DetailedCategorySchema, EventSchema
 from indico.modules.search.views import WPSearch
+from indico.util.caching import memoize
 from indico.web.args import use_kwargs
 from indico.web.rh import RH
 
 
+@memoize
```
ThiefMaster, line 28: `@memoize_request` - we don't want to keep this forever in the process!

indico/modules/search/controllers.py
```diff
@@ -5,60 +5,77 @@
...
-            if session.user.can_get_all_multipass_groups:
-                access += [GroupProxy(x.name, x.provider.name).identifier
-                           for x in session.user.iter_all_multipass_groups()]
-            pages, total, results = search_provider().search(q, access, self.TARGET)
-            return {'pages': pages, 'total': total, 'results': results}
+        if not search_provider or target is SearchTarget.category:
+            search_provider = InternalSearch
+        access = []
+        if session.user:
+            access = get_groups(session.user)
```
ThiefMaster, line 56:
```suggestion
        access = get_groups(session.user) if session.user else []
```

indico/modules/search/controllers.py
```diff
@@ -5,60 +5,77 @@
...
         search_provider = get_search_provider()
-        if search_provider and self.USE_PROVIDER:
-            # TODO: memoize
-            access = [session.user.identifier] + [x.identifier for x in session.user.local_groups]
-            if session.user.can_get_all_multipass_groups:
-                access += [GroupProxy(x.name, x.provider.name).identifier
-                           for x in session.user.iter_all_multipass_groups()]
-            pages, total, results = search_provider().search(q, access, self.TARGET)
-            return {'pages': pages, 'total': total, 'results': results}
+        if not search_provider or target is SearchTarget.category:
```
ThiefMaster, line 52:
```suggestion
        if not search_provider or target == SearchTarget.category:
```

indico/modules/search/controllers.py
```diff
@@ -5,60 +5,77 @@
...
     @use_kwargs({
         'page': fields.Int(missing=1),
         'q': fields.String(required=True),
-    }, location='query')
-    def _process(self, page, q):
+    }, location='query', unknown=INCLUDE)
+    def _process(self, page, q, **params):
+        target = SearchTarget.get(request.view_args['type'], SearchTarget.event)
+        if not target:
+            raise BadRequest('Invalid search target')
```
ThiefMaster, line 50: Are you sure this will ever happen? You are passing a default value to `.get()`

indico/modules/search/client/js/components/SideBar.jsx
```diff
@@ -6,130 +6,84 @@
...
+}
 
-      <Table celled padded>
-        <Table.Header>
-          <Table.Row>
-            <Table.HeaderCell>Affiliations</Table.HeaderCell>
-          </Table.Row>
-        </Table.Header>
+function AggregationList({name, title, items, query, onChange}) {
+  const _items = items
```
ThiefMaster, line 38: IMHO `_items` vs `items` gets a bit confusing... what about just reassigning `items = items.map(...)` instead?

indico/modules/search/client/js/components/SideBar.jsx
```diff
@@ -6,130 +6,84 @@
...
-      setAffiliationFilter(affiliationValue === value ? '' : value, 'pushIn');
-      setAffiliationValue(affiliationValue === value ? '' : value);
-    }
-  };
-
-  const handleChange = (e, {value}) => {
-    addSearchFilter(value.name, value.type);
+export default function SideBar({query, aggregations, onChange}) {
+  const handleChange = (type, value) => {
+    onChange && onChange(type in query && query[type] === value ? undefined : value, type);
```
ThiefMaster, line 16: if (onChange) { ... }

ThiefMaster: `type in query && ` isn't needed, `query[type]` will just be undefined otherwise

indico/modules/search/client/js/components/SearchApp.jsx
```diff
@@ -106,51 +88,87 @@ SearchTypeMenuItem.defaultProps = {
...
+        pathname: window.location.pathname,
+        search: `?${_query}`,
+      });
+      _setQuery(_query);
+    },
+    [history, query]
+  );
+
+  useEffect(() => {
+    history.listen(location => {
```
ThiefMaster, line 115: this listener should probably be removed in the useEffect's cleanup function

indico/modules/search/client/js/components/SideBar.jsx
```diff
@@ -6,130 +6,84 @@
...
-        <Table.Header>
-          <Table.Row>
-            <Table.HeaderCell>Affiliations</Table.HeaderCell>
-          </Table.Row>
-        </Table.Header>
+function AggregationList({name, title, items, query, onChange}) {
+  const _items = items
+    .filter(b => b.key)
+    .map(({key, fromAsString: from, toAsString: to, docCount: count}) => ({
+      key: from || to ? `[${from || '*'} TO ${to || '*'}]` : key,
```
ThiefMaster, line 41: do we really need something this fancy for the react key? something as simple as `${from}|${to}|${key}` might do the job just as well

indico/modules/search/client/js/components/SideBar.jsx
```diff
@@ -6,130 +6,84 @@
...
             <Table.Cell>
               <Checkbox
-                label="Giota's (333)"
-                name="Giota"
-                value={{name: 'Giota', type: 'affiliation'}}
-                checked={affiliationValue === 'Giota'}
-                onChange={handleChange}
+                styleName="checkbox"
+                label={`${label} (${count})`}
+                checked={name in query && query[name] === key}
```
ThiefMaster, line 60:
```suggestion
                checked={query[name] === key}
```

indico/modules/search/client/js/components/SideBar.jsx
plourenco: Indeed, and in the begging, I wasted some time with alternatives such as `${from}-${to}` or the above.
However, while this can be changed, [ES range filters](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_ranges) are exactly made up as `[DATE TO DATE]`. So I thought I could just use the native form.

indico/modules/search/client/js/components/SearchApp.jsx
plourenco: Fair point!

indico/modules/search/client/js/components/SideBar.jsx
plourenco: Right! I don't think initially I was expecting the query to be an empty object.

indico/modules/search/client/js/components/SideBar.jsx
plourenco: I agree, it seems confusing.

indico/modules/search/controllers.py
plourenco: No, it doesn't make sense, I was undecided initially, but I believe my latest idea was to default it to Events.

indico/modules/search/client/js/components/SideBar.jsx
plourenco: As for the `onChange`, I assume you're not a fan of the one-liner? As I could get rid of the brackets instead.

ThiefMaster: Exactly, I don't mind it when using it in an expression or jsx attribute, but not in a normal statement

indico/modules/search/controllers.py
ThiefMaster: Isn't it coming from `any()` in the url rule? If yes, I'd do `SearchTarget[request.view_args['type']]` since it makes sense to fail loudly if we get crap data there.

But actually, the even nicer option would be this:

```python
@use_kwargs({'type': EnumField(SearchTarget)}, location='view_args')
```

indico/modules/search/client/js/components/SearchApp.jsx
```diff
@@ -106,55 +88,85 @@ SearchTypeMenuItem.defaultProps = {
...
-  // On every search, the tab will be either the first loading or with results
-  const menuItem = searchMap.findIndex(x => x[1].loading || x[1].total > 0);
-  // TODO: we could make this initially uncontrolled, but if the user controls it, it no longer changes
-  useEffect(() => {
-    setActiveMenuItem(Math.max(0, menuItem));
-  }, [menuItem]);
-
-  const handleQuery = value => setQuery(value, 'pushIn');
+  const handleQuery = (value, type) => {
+    const _type = type || 'q';
```
ThiefMaster, line 146: Not sure if it works, but in case it gets called without an argument for type (and not just a falsy one), you could just use a default value:

```suggestion
  const handleQuery = (value, type = 'q') => {
```

indico/modules/search/controllers.py
```diff
@@ -15,50 +18,60 @@
...
+from indico.util.caching import memoize_request
 from indico.web.args import use_kwargs
 from indico.web.rh import RH
 
 
+@memoize_request
+def get_groups(user):
+    access = [user.identifier] + [x.identifier for x in user.local_groups]
+    if user.can_get_all_multipass_groups:
+        access += [GroupProxy(x.name, x.provider.name).identifier
```
ThiefMaster, line 32:
```suggestion
        access += [GroupProxy(x.name, x.provider.name, x).identifier
```

While `.identifier` should never trigger a lookup of the remote group data, passing the group explicitly via the third constructor argument ensures that it's always available.

plourenco: Sounds good

indico/modules/search/client/js/components/SearchApp.jsx
plourenco: Sure, neither the value should be strictly undefined or false, so this is fine too.

PR #4850, opened 2021-04-01
Add download option to calendar export popups

### LINKED ISSUE #4827: Restore download option for event ical files
With the changes in #4801 we lost the option to simply download the ical file without bothering with a signed download URL etc:

image

Now we *only* have the copy signed URL option

image

Let's also keep a download link there. We can use the same signed URL for it (even though the signature is not strictly necessary since we have a valid session)

For the UI we can either see whether we can have two action buttons on the input line, but that may be tricky and/or look ugly; or we can just show something like "[Download] the current calendar" link above or below.

javfg:
This PR remodels the popup shown when clicking on the different 'export calendar' buttons.

After https://github.com/indico/indico/pull/4801, the option to download the ICS file was lost, as we started using the dashboard popup everywhere. With the new layout, it is back; and the popup is uniform across the whole application:

image

Also, the different buttons that open the popup will now shade/sink properly when clicked.

Closes #4827.

### DIFF
--- a/indico/modules/events/client/js/header.jsx
+++ b/indico/modules/events/client/js/header.jsx
@@ -32,11 +32,13 @@ document.addEventListener('DOMContentLoaded', () => {
     <ICSCalendarLink
       endpoint="events.export_event_ical"
       params={{event_id: eventId}}
-      renderButton={onClick => (
+      renderButton={(onClick, classes) => (
         <IButton
           icon="calendar"
-          classes={{'height-full': true, 'text-color': true, 'subtle': true}}
+          dropdown
+          classes={{'height-full': true, 'text-color': true, 'subtle': true, ...classes}}
           onClick={onClick}
+          title="Export"
         />
       )}
       dropdownPosition="top left"
--- a/indico/modules/events/client/js/header.jsx
+++ b/indico/modules/events/client/js/header.jsx
@@ -38,7 +38,7 @@ document.addEventListener('DOMContentLoaded', () => {
           dropdown
           classes={{'height-full': true, 'text-color': true, 'subtle': true, ...classes}}
           onClick={onClick}
-          title="Export"
+          title={Translate.string('Export')}
         />
       )}
       dropdownPosition="top left"
--- a/indico/web/client/js/react/components/ICSCalendarLink.jsx
+++ b/indico/web/client/js/react/components/ICSCalendarLink.jsx
@@ -84,7 +84,7 @@ export default function ICSCalendarLink({
       >
         <Dropdown.Menu>
           <Dropdown.Header>
-            <Translate>Synchronise with your calendar</Translate>
+            <Translate>Export</Translate>
           </Dropdown.Header>
           <Dropdown.Divider />
           {options.map(({key, text, extraParams}) => (
@@ -125,11 +125,12 @@ export default function ICSCalendarLink({
       wide
     >
       <Header
-        styleName="centered"
-        content={Translate.string('Sync with your calendar')}
+        styleName="export-header"
+        content={Translate.string('Export')}
         subheader={option && <Label color="blue">{option.text}</Label>}
       />
       <Popup.Content>
+        <strong styleName="export-option">Synchronise with your calendar</strong>
         <p>
           <Translate>
             You may copy-paste the following URL into your scheduling application. Contents will be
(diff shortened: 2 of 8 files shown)


indico/web/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -154,6 +161,20 @@ export default function ICSCalendarLink({
...
+        <div styleName="download-option">
+          <span>
+            <Translate>
+              Download an iCalendar file that you can use in calendaring applications.
+            </Translate>
+          </span>
+          <Button
+            styleName="download-button"
+            as="a"
+            href={option ? option.url : ''}
```
ThiefMaster, line 174: I guess this is missing while still retrieving the URL? If yes, you may want to add `disabled={!option}`

indico/web/client/js/react/components/ICSCalendarLink.module.scss
```diff
@@ -9,7 +9,39 @@
...
+
+.export-option {
+  display: block;
+  text-align: center;
+  margin: 1rem 0 0.5rem 0;
+}
+
+.export-option::before,
+.export-option::after {
+  box-shadow: 1px 1px 1px #f9f9f9;
```
ThiefMaster, line 26: This is exposed as `$light-gray` from _palette.scss

indico/web/client/js/react/components/ICSCalendarLink.module.scss
```diff
@@ -9,7 +9,39 @@
...
+  background-color: #dfdfdf;
```
ThiefMaster, line 27: `$pastel-gray`

indico/web/client/js/react/components/ICSCalendarLink.module.scss
```diff
@@ -9,7 +9,39 @@
...
+  vertical-align: middle;
+  width: 10%;
+  margin: 0 0.5rem;
+}
+
+.download-button:global(.ui.icon.button) {
+  display: flex;
+  align-items: center;
+  margin: 0;
+  height: 2.125rem;
```
ThiefMaster, line 41: Where's this number coming from?

indico/web/client/js/react/components/ICSCalendarLink.jsx
javfg: I thought about disabling the button, but [line 153](https://github.com/javfg/indico/blob/ical-downloadbutton/indico/web/client/js/react/components/ICSCalendarLink.jsx#L153) hides it while loads, so discarded the idea in the end.

ThiefMaster: Sure, but that's only hiding the copy button, not the download button...

javfg: Aaah, I see the problem. There was another thing going on: [`setOpen`](https://github.com/javfg/indico/blob/ical-downloadbutton/indico/web/client/js/react/components/ICSCalendarLink.jsx#L63) was being run after the `await`, so I had no way to check what was happening in the form before the URL loaded. It just did not show until `option.url` was there.

javfg: I'll fix that.

indico/web/client/js/react/components/ICSCalendarLink.module.scss
javfg: Comes from the height of SUI input component. The idea is for both buttons (this and the copy one on top of him) to have the same size.

indico/modules/events/client/js/header.jsx
```diff
@@ -32,11 +32,13 @@ document.addEventListener('DOMContentLoaded', () => {
...
       params={{event_id: eventId}}
-      renderButton={onClick => (
+      renderButton={(onClick, classes) => (
         <IButton
           icon="calendar"
-          classes={{'height-full': true, 'text-color': true, 'subtle': true}}
+          dropdown
+          classes={{'height-full': true, 'text-color': true, 'subtle': true, ...classes}}
           onClick={onClick}
+          title="Export"
```
ThiefMaster, line 41:
```suggestion
          title={Translate.string('Export')}
```

ThiefMaster (approved):

Issue #4719, opened 2020-11-23
Corner case cancelling full booking - misleading description of behaviour


pferreir:
**Describe the bug**
When cancelling a full booking which has started already, Indico indicates that it's going to remove all the occurrences, but then removes all occurrences minus one.

**To Reproduce**
1. Create a booking which spans over 8 days. Make it so that it starts exactly today;
2. Cancel the booking;
3. Indico will ask you if you want to delete the 8 occurrences - do so;

**Expected behavior**
All occurrences should go away (or Indico should let me know that one of the occurrences will stay nonetheless).

**Screenshots**

![image](https://user-images.githubusercontent.com/2699/99969736-2c384c80-2d9b-11eb-93fc-f5e8d7aa6694.png)
![image](https://user-images.githubusercontent.com/2699/99969761-322e2d80-2d9b-11eb-9d6f-b4e7d0c18c73.png)

**Additional context**
This is was tested on Burotel. If it doesn't happen in vanilla Indico, then we should move it to the relevant repo.

plourenco:
I tried to reproduce this in vanilla with 8 occurrences but all of them were cancelled.

PR #4861, opened 2021-04-09
Various fixes for Room Booking

javfg:
This PR closes https://github.com/indico/indico/issues/4719, by showing the correct number  of booking occurrences that will be cancelled with a cancel action.

Also:
* Fixes an incorrect parameter passed to Popper's offset modifier which was causing the popup for the time on the click-and-drag ui for the day calendar not to follow the mouse pointer.
* Fixes the fetching of bookings so it doesn't get the same bookings infinitely.

### DIFF
--- a/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
+++ b/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
@@ -571,7 +571,13 @@ class BookingDetails extends React.Component {
     return transformToLegendLabels(occurrenceTypes, inactiveTypes);
   };
 
-  renderActionButtons = (canCancel, canReject, showAccept, occurrenceCount, isAccepted) => {
+  renderActionButtons = (
+    canCancel,
+    canReject,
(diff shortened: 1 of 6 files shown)


indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -825,6 +831,9 @@ class BookingDetails extends React.Component {
     const showActionButtons = !isCancelled && !isRejected && (canCancel || canReject || showAccept);
     const activeBookings = _.omitBy(occurrences.bookings, value => _.isEmpty(value));
     const occurrenceCount = Object.keys(activeBookings).length;
+    const cancellableOccurrenceCount = Object.keys(activeBookings).filter(date =>
```
ThiefMaster, line 834: `Object.entries()` is the better choice here since you need the values as well. Or `Object.values()` since you don't use the keys except to get the values..

ThiefMaster (approved):

PR #4865, opened 2021-04-12
Add an option to keep Editing team anonymous

### LINKED ISSUE #4829: Editing: Option to anonymize editor/commenter names for the author/submitter
This came up when discussing with the vCHEP organizers. It seems to be pretty common in some reviewing workflows that the authors/submitters of papers never see the names of the editors; especially in case they may be a bit more junior than the submitter (to prevent any pressure from submitters to editors).

A setting to anonymize editor information for those people would help there.

This would be needed around mid april, so depending on whether 3.0 is ready by then it may need to go in 2.3!

ThiefMaster:
image

closes #4829

### DIFF
--- a/indico/modules/events/client/js/reviewing/components/UserAvatar.jsx
+++ b/indico/modules/events/client/js/reviewing/components/UserAvatar.jsx
@@ -14,13 +14,14 @@ function UserAvatar({user}) {
       className="i-timeline-item-label avatar-placeholder"
       style={{backgroundColor: user.avatarBgColor}}
     >
-      {user.fullName[0]}
+      {user.anonymous ? '' : user.fullName[0]}
     </div>
   );
 }
 
 UserAvatar.propTypes = {
   user: PropTypes.shape({
+    anonymous: PropTypes.bool,
     fullName: PropTypes.string.isRequired,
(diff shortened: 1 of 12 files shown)


indico/modules/events/client/js/reviewing/components/UserAvatar.jsx
```diff
@@ -14,13 +14,14 @@ function UserAvatar({user}) {
       className="i-timeline-item-label avatar-placeholder"
       style={{backgroundColor: user.avatarBgColor}}
     >
-      {user.fullName[0]}
+      {user.anonymous ? '' : user.fullName[0]}
```
plourenco, line 17: We could check if the `id` is `-1` and probably skip having an anonymous attribute 
(as it seems to be just for the avatar)

plourenco (approved):

indico/modules/events/client/js/reviewing/components/UserAvatar.jsx
ThiefMaster: I had this initially; but I preferred being more explicit in the end. The only reason I'm sending the `-1` id (instead of `null`) is to make PropTypes happy without finding every single place where we're using it or passing it along ;)

Issue #2462, opened 2016-06-10
Better time/duration widget for contributions


pferreir:
The fact that Indico is used mostly by people who are presumably good at maths is not enough to justify that our current duration widget can work either in minutes or hours, but never in both.

![screen shot 2016-06-10 at 15 38 12](https://cloud.githubusercontent.com/assets/2699/15966010/5d75be04-2f21-11e6-8141-6713be03ab9b.png)

It would be better to just allow both hours and minutes, so that people don't need to mentally figure out how many minutes 1h15m corresponds to.

ThiefMaster:
In newdle we adapted the rc-time-picker for it which works pretty well. We could probably do the same here.

PR #4873, opened 2021-04-15
New duration field for durations under 24h

javfg:
This PR implements the same widget Newdle uses for time durations (https://github.com/indico/newdle/pull/213). The new `IndicoDurationField` replaces `TimedeltaField` in places where values lower than a day are used.

Closes https://github.com/indico/indico/issues/2462.

Also, the focus stays in the input element when clicking in the field, so no double click is needed anymore.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,6 +52,8 @@ Improvements
   conferences now redirect to the meeting view page (:pr:`4847`)
 - Use a more compact setup QR code for the mobile *Indico check-in* app; the latest version of
   the app is now required. (:pr:`4844`)
+- Duration fields in different forms throughout the application are now easier to use.
+  (:issue:`2462`, :pr:`4873`)
 
 Bugfixes
 ^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,8 +52,8 @@ Improvements
   conferences now redirect to the meeting view page (:pr:`4847`)
 - Use a more compact setup QR code for the mobile *Indico check-in* app; the latest version of
   the app is now required. (:pr:`4844`)
-- Duration fields in different forms throughout the application are now easier to use.
-  (:issue:`2462`, :pr:`4873`)
+- Contribution duration fields now use a widget similar to the time picker that makes selecting
+  durations easier. (:issue:`2462`, :pr:`4873`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/web/client/js/react/components/WTFDurationField.jsx
+++ b/indico/web/client/js/react/components/WTFDurationField.jsx
@@ -1,0 +1,65 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2021 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import moment from 'moment';
+import PropTypes from 'prop-types';
+import TimePicker from 'rc-time-picker';
+import React, {useEffect, useMemo, useRef, useState, useCallback} from 'react';
+
+const timeToSeconds = time => (time ? time.diff(moment().startOf('day'), 'seconds') : null);
+const secondsToTime = seconds =>
+  moment()
+    .startOf('day')
+    .seconds(seconds);
+
+export default function WTFDurationField({timeId, required, disabled}) {
+  const timeField = useMemo(() => document.getElementById(timeId), [timeId]);
+  const [time, setTime] = useState(secondsToTime(timeField.value));
+  const timePickerRef = useRef(null);
+
+  const handleTimePickerChange = useCallback(
+    value => {
+      setTime(value || null);
+      timeField.value = timeToSeconds(value);
+      timeField.dispatchEvent(new Event('change', {bubbles: true}));
+    },
+    [timeField]
+  );
+
+  useEffect(() => {
+    timePickerRef.current.picker.required = required;
+  }, [required]);
+
+  return (
+    <TimePicker
+      showSecond={false}
+      value={time}
+      focusOnOpen
+      format="H:mm"
+      onChange={handleTimePickerChange}
+      allowEmpty={false}
+      placeholder="h:mm"
+      disabled={disabled}
+      ref={timePickerRef}
+      // keep the picker in the DOM tree of the surrounding element to avoid
+      // e.g. qbubbles from closing when a picker is used inside one and the
+      // user clicks something in the picker
+      getPopupContainer={node => node}
+    />
+  );
+}
+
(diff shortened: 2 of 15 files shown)


indico/web/forms/fields/datetime.py
```diff
@@ -177,6 +177,20 @@ class IndicoTimeField(TimeField):
...
+    widget = JinjaWidget('forms/duration_widget.html', single_line=True, single_kwargs=True)
+
+    def _value(self):
+        if self.data is None:
+            return 0
+        else:
+            return int(self.data.total_seconds())
+
+    def process_formdata(self, valuelist):
+        if valuelist and len(valuelist) == 1:
```
ThiefMaster, line 190: I don't think we need the length check here; in the other field we just have it because we need more than one value

indico/web/client/styles/modules/_contributions.scss
```diff
@@ -141,6 +141,10 @@
 
 .qbubble-contrib-duration {
   max-width: 420px !important;
+
+  & div form {
```
ThiefMaster, line 145: I don't think the `&` is needed here

indico/web/client/js/react/components/WTFDurationField.jsx
```diff
@@ -0,0 +1,65 @@
...
+import moment from 'moment';
+import PropTypes from 'prop-types';
+import TimePicker from 'rc-time-picker';
+import React, {useEffect, useMemo, useRef, useState, useCallback} from 'react';
+
+const timeToSeconds = time => (time ? time.diff(moment().startOf('day'), 'seconds') : null);
+const secondsToTime = seconds =>
+  moment()
+    .startOf('day')
+    .seconds(seconds);
```
ThiefMaster, line 17: Just wondering.. will this work correctly on days with a DST change?

javfg: Looks like it does, as adding seconds to a DST day works just as with a normal day:

```javascript
moment('2021-03-28 08:00').startOf('day').seconds('43200').toDate()
Sun Mar 28 2021 12:00:00 GMT+0200 (Central European Summer Time)
moment('2021-03-27 08:00').startOf('day').seconds('43200').toDate()
Sat Mar 27 2021 12:00:00 GMT+0100 (Central European Standard Time)
```
When using `diff`, it accounts for the missing hour:

```javascript
a = moment('2021-03-28 00:00')
Moment {_isAMomentObject: true, _i: "2021-03-28 00:00", _f: "YYYY-MM-DD HH:mm", _isUTC: false, _pf: {…}, …}
b = moment('2021-03-28 23:59:59')
Moment {_isAMomentObject: true, _i: "2021-03-28 23:59:59", _f: "YYYY-MM-DD HH:mm:ss", _isUTC: false, _pf: {…}, …}
b.diff(a, 'hour')
22
```
Because `a` and `b` are different timezones:
```javascript
a.toDate()
Sun Mar 28 2021 00:00:00 GMT+0100 (Central European Standard Time)
b.toDate()
Sun Mar 28 2021 23:59:59 GMT+0200 (Central European Summer Time)
```

CHANGES.rst
```diff
@@ -52,6 +52,8 @@ Improvements
   conferences now redirect to the meeting view page (:pr:`4847`)
 - Use a more compact setup QR code for the mobile *Indico check-in* app; the latest version of
   the app is now required. (:pr:`4844`)
+- Duration fields in different forms throughout the application are now easier to use.
+  (:issue:`2462`, :pr:`4873`)
```
ThiefMaster, line 56: What do you think about this?

```suggestion
- Contribution duration fields now use a widget similar to the time picker that makes selecting
  durations easier. (:issue:`2462`, :pr:`4873`)
```

indico/web/client/styles/modules/_contributions.scss
```diff
@@ -141,6 +141,10 @@
 
 .qbubble-contrib-duration {
   max-width: 420px !important;
+
+  div form {
+    height: 200px;
+  }
```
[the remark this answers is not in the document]
ThiefMaster: We need the same logic for this tooltip in the timetable:

image

This seems to do the job:

```patch
diff --git a/indico/web/client/styles/partials/_timetable-balloons.scss b/indico/web/client/styles/partials/_timetable-balloons.scss
index 4c9d1bcd6f..dce11dbf06 100644
--- a/indico/web/client/styles/partials/_timetable-balloons.scss
+++ b/indico/web/client/styles/partials/_timetable-balloons.scss
@@ -157,6 +157,10 @@
   max-width: 410px;
   min-width: 410px;

+  .qtip-content {
+    overflow: visible;
+  }
+
   .i-form {
     input {
       width: 7em !important;
```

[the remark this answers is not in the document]
javfg: Much better indeed.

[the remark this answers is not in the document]
javfg: > We need the same logic for this tooltip in the timetable:
> 
> image
> 
> This seems to do the job:
> 
> ```diff
> diff --git a/indico/web/client/styles/partials/_timetable-balloons.scss b/indico/web/client/styles/partials/_timetable-balloons.scss
> index 4c9d1bcd6f..dce11dbf06 100644
> --- a/indico/web/client/styles/partials/_timetable-balloons.scss
> +++ b/indico/web/client/styles/partials/_timetable-balloons.scss
> @@ -157,6 +157,10 @@
>    max-width: 410px;
>    min-width: 410px;
> 
> +  .qtip-content {
> +    overflow: visible;
> +  }
> +
>    .i-form {
>      input {
>        width: 7em !important;
> ```

I missed that popup!

ThiefMaster:
Would it make sense to disallow a zero duration? It fails the server-side validation anyway, but it'd be nicer to not allow submitting it in the first place.

Excluding the `0` minute if the hour is `0` as well is probably the easiest way to do that. TimePicker has a `disabledMinutes`; something like `h => (h === 0 ? [0] : [])` probably does the job (didn't test it myself)

javfg:
> Would it make sense to disallow a zero duration? It fails the server-side validation anyway, but it'd be nicer to not allow submitting it in the first place.
> 
> Excluding the `0` minute if the hour is `0` as well is probably the easiest way to do that. TimePicker has a `disabledMinutes` prop; something like `h => (h === 0 ? [0] : [])` probably does the job (didn't test it myself)

We could do the same with hour zero when minutes are zero, but I've tried and it's bad on usability: if the user is at 1:00 and wants to set 0:20, he would have to set minutes first, potentially scrolling down, which feels annoying to do before setting the hour.

ThiefMaster:
Could we automatically switch to 0:01 in that case?

ThiefMaster (approved):

PR #4874, opened 2021-04-15
Add before_notification_send signal

OmeGak:

### DIFF
--- a/indico/core/signals/core.py
+++ b/indico/core/signals/core.py
@@ -86,3 +86,10 @@ db_schema_created = _signals.signal('db-schema-created', """
 Executed when a new database schema is created.  The *sender* is the
 name of the schema.
 """)
+
+before_notification_send = _signals.signal('before-notification-send', """
+Executed before a notification is sent.  The *sender* is a string representing
+the type of notification.  The notification email that will be sent is passed in
+the ``email`` kwarg.  The additional kwargs passed to this signal depend on the
+context.
+""")
--- a/indico/modules/events/registration/notifications.py
+++ b/indico/modules/events/registration/notifications.py
@@ -27,7 +27,7 @@ def notify_invitation(invitation, email_subject, email_body, from_address):
     send_email(email, invitation.registration_form.event, 'Registration', user)
 
 
-def _notify_registration(registration, template, to_managers=False, attach_rejection_reason=False):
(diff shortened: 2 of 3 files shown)


indico/core/signals/core.py
```diff
@@ -86,3 +86,9 @@
 Executed when a new database schema is created.  The *sender* is the
 name of the schema.
 """)
+
+before_notification_sent = _signals.signal('before-notification', """
+Executed before a notification is sent.  The *sender* is a string representing
+the type of notification.  The additional kwargs passed to this signal depend on
```
ThiefMaster, line 92: `mail` could be documented here as well since that should always be passed

indico/modules/events/registration/notifications.py
```diff
@@ -39,12 +39,15 @@ def _notify_registration(registration, template, to_managers=False, attach_rejec
...
         attachments = get_ticket_attachments(registration)
 
+    template_name = template
     template = get_template_module('events/registration/emails/{}'.format(template),
                                    registration=registration, attach_rejection_reason=attach_rejection_reason)
     to_list = registration.email if not to_managers else registration.registration_form.manager_notification_recipients
     from_address = registration.registration_form.sender_address if not to_managers else None
     mail = make_email(to_list=to_list, template=template, html=True, from_address=from_address, attachments=attachments)
     user = session.user if session else None
+    mail = signals.core.before_notification_send.send('notify-registration', mail=mail, registration=registration,
```
ThiefMaster, line 49: Did you test this? The return value of `.send()` is not what you may expect - remember, you can have more than one handler for a signal... Probably better to not use it at all and document that `mail` should be mutated if necessary.

indico/core/signals/core.py
```diff
@@ -86,3 +86,10 @@
 Executed when a new database schema is created.  The *sender* is the
 name of the schema.
 """)
+
+before_notification_send = _signals.signal('before-notification', """
```
ThiefMaster, line 90:
```suggestion
before_notification_send = _signals.signal('before-notification-send', """
```

indico/modules/events/registration/notifications.py
OmeGak: > Probably better to not use it at all and document that mail should be mutated if necessary.

👍

ThiefMaster (approved):

PR #4875, opened 2021-04-16
Fix issues with NPM 7

ThiefMaster:


javfg (approved):
This fixes the NPM problem with SHA512 mismatch. Also, it is using the correct protocol to fetch from GitHub now.

PR #4878, opened 2021-04-19
Update search result listing and attributes

plourenco:


indico/modules/search/client/js/components/results/Contribution.jsx
```diff
@@ -13,44 +15,55 @@ import {toMoment, serializeDate} from 'indico/utils/date';
...
         <List.Item>
-          <Icon name="calendar alternate outline" />
-          {serializeDate(toMoment(startDt), 'DD MMMM YYYY HH:mm')}
+          {persons.length !== 0 && (
+            <ul>
+              {persons.length > 1 ? <Icon name="users" /> : <Icon name="user" />}
+              {persons.map(person => (
+                <li key={person}>
+                  {person.name}
+                  {person.affiliation ? ` (${person.affiliation.substring(0, 12)})` : ''}
```
ThiefMaster, line 34: Shouldn't we ellipsize this instead of just truncating?

indico/modules/search/client/js/components/results/Event.jsx
```diff
@@ -5,12 +5,14 @@
...
 
+import eventURL from 'indico-url:events.display';
+
 import moment from 'moment';
 import PropTypes from 'prop-types';
 import React from 'react';
 import {List, Icon} from 'semantic-ui-react';
-import '../ResultList.module.scss';
 
+import '../ResultList.module.scss';
```
ThiefMaster, line 15: move after the non-css imports

indico/modules/search/client/js/components/results/Event.jsx
```diff
@@ -76,15 +79,14 @@ const personShape = PropTypes.shape({
 
 Event.propTypes = {
   title: PropTypes.string.isRequired,
-  url: PropTypes.string.isRequired,
-  typeFormat: PropTypes.oneOf(['lecture', 'meeting', 'conference']).isRequired,
+  eventId: PropTypes.number.isRequired,
+  typeFormat: PropTypes.oneOf(['lecture', 'meeting', 'conference']),
```
ThiefMaster, line 83: when is this missing?

and just wondering, why `typeFormat` and not `type`?

indico/modules/search/client/js/components/results/EventNote.jsx
```diff
@@ -10,34 +10,32 @@ import React from 'react';
...
-  );
-};
+const EventNote = ({title, url, content, highlight}) => (
+  <div styleName="item">
+    <List.Header styleName="header">
+      <a href={url}>{title}</a>
+    </List.Header>
+    <List.Description styleName="description">
+      {highlight?.content
+        ?.slice(0, 3) // eslint-disable-next-line react/no-array-index-key
```
ThiefMaster, line 20: wtf prettier really wraps before the `?`?!

indico/modules/search/schemas.py
```diff
@@ -58,7 +58,7 @@ class Meta:
 
     event_id = fields.Int(attribute='id')
     type = fields.Constant(SearchTarget.event.name)
-    type_format = fields.String(attribute='type.name')
+    type_format = fields.String(attribute='type')
```
ThiefMaster, line 61: ⚠️ I think this breaks the export!

plourenco: Indeed, I trusted the property value, the latest one should fix it.

indico/modules/search/client/js/components/results/EventNote.jsx
plourenco: Seems really weird right? Perhaps I should use another alternative here?

indico/modules/search/client/js/components/results/Event.jsx
plourenco: Not supposed to be missing - however, the component should work completely fine without it, for that reason I decided to default it to minimal behaviour.

indico/modules/search/client/js/components/results/Contribution.jsx
plourenco: Well, we could but I noticed there were some really long affiliations such as University of Somewhere Really Far Away, and having all of those occupying so much space from other person's names is excessive.

indico/modules/search/client/js/components/results/Event.jsx
plourenco: Not supposed to be missing, however since the component works just fine without it, I decided to default it to a minimal behaviour.

indico/modules/search/client/js/components/results/Contribution.jsx
ThiefMaster: It looks a bit strange:

image

but even without trailing space it's a bit strange if it's cut off in the middle

image



Maybe we could just show the names there and put the affiliation in a tooltip? Alternatively, wrapping the affiliation in a span with a max width + text-overflow ellipsis could be an option as well.

But I'm not sure if showing the affiliation there immediately is that important: I just checked the old CERN Search and we don't show the affiliation there at all. So maybe for now we can just omit it and if someone asks for it we revisit this?

indico/modules/search/client/js/components/results/Event.jsx
plourenco: Forgot the second line. `type` is supposed to be a generic type that identifies the record, such as, Event, Contribution, Attachment. `typeFormat` is a record-specific format such as Event being a meeting, conference, etc.

ThiefMaster: ah... unless citadel uses `type_format` by default, maybe `event_type` (`eventType` here) would be clearer?

plourenco: Yes, citadel is already using it. The problem is `type_format` is already being indexed by all records, and it applies to both a contribution, event and attachment independently of their type, so changing the prop wouldn't be simple. This would have to be a change at schema-citadel level.

ThiefMaster: let's leave it then; i was just curious about the naming :)

indico/modules/search/client/js/components/results/Contribution.jsx
```diff
@@ -13,44 +15,57 @@ import {toMoment, serializeDate} from 'indico/utils/date';
...
+      </List.Header>
+      <List.Description styleName="description">
         <List.Item>
-          <Icon name="calendar alternate outline" />
-          {serializeDate(toMoment(startDt), 'DD MMMM YYYY HH:mm')}
+          {persons.length !== 0 && (
+            <ul>
+              {persons.length > 1 ? <Icon name="users" /> : <Icon name="user" />}
+              {persons.map(person => (
+                <li key={person}>
```
ThiefMaster, line 32: the keys need to be strings, so e.g. `key={person.name + person.affiliation}`

indico/modules/search/client/js/components/results/Event.jsx
```diff
@@ -30,28 +32,26 @@ const renderDates = (startDt, endDt) =>
...
       <List.Description styleName="description">
-        {/* if it's a lecture print the list of speakers */}
         {type === 'lecture' && chairPersons.length !== 0 && (
           <List.Item>
-            {chairPersons.map(i => (
-              <div key={i.name}>
-                {i.title
-                  ? `${i.title} ${i.name} (${i.affiliation})`
-                  : `${i.name} (${i.affiliation})`}
+            {chairPersons.map(person => (
```
ThiefMaster, line 44: We need the same deduplication logic like in contributions for name+affiliation here.

indico/modules/search/client/js/components/results/PersonList.jsx
```diff
@@ -0,0 +1,35 @@
...
+import {Translate} from 'indico/react/i18n';
+
+import '../ResultList.module.scss';
+
+export default function PersonList({persons}) {
+  persons = [...new Map(persons.map(p => [p.name + p.affiliation, p])).values()];
+
+  return (
+    <ul>
+      {persons.length > 1 ? <Icon name="users" /> : <Icon name="user" />}
```
ThiefMaster, line 21: Shouldn't this be outside the `<ul>`? It's not an `<li>` after all...

indico/modules/search/client/js/components/results/PersonList.jsx
```diff
@@ -0,0 +1,35 @@
...
+      {persons.slice(0, 2).map((person, idx, arr) => (
+        <li key={person.name + person.affiliation}>
+          {person.name}
+          <span styleName="muted">{person.affiliation ? ` (${person.affiliation})` : ''}</span>
+          {persons.length > 2 && idx + 1 === arr.length && ` ${Translate.string('et al.')}`}
```
ThiefMaster, line 26: What about adding this as a separate list element?

plourenco: It works too, I was just looking to keep the CSS property that tried to limit the amount of `li` elements (in case it becomes useful for another search result element). Anyway, it shouldn't be a problem except for the minor difference of adding a comma before `et al.`.

ThiefMaster: yeah it was more about semantics ;) if we already use a list it's a bit strange to mix different things in a single item

PR #4879, opened 2021-04-19
Increase search responsiveness

plourenco:
Depends on #4878


indico/modules/search/client/js/components/SearchPagination.jsx
```diff
@@ -8,10 +8,13 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import {Pagination} from 'semantic-ui-react';
+
 import './SearchPagination.module.scss';
+import {useResponsive} from 'indico/react/util';
```
ThiefMaster, line 13: let's keep css imports at the very end (eslint cannot enforce this because it doesn't know that css imports have no significant side effects :/)

plourenco: Yes, that one slipped through

ThiefMaster:
I'm getting some weird scrollbars:

image

PR #4882, opened 2021-04-20
Add log_metadata while sending custom emails

vasantvohra:
Although custom sent emails are logged at the event's log, but couldn't be filtered by registration Id. Therefore, adding metadata to the ` func: send_email`


ThiefMaster:
Your git `user.name` seems to be misconfigured: `Author: “Vasant <vasantvohra@gmail.com>` (see that quote at the beginning of the name?).

Anyway I'll amend your commit with the data from one of your prior commits...

ThiefMaster (approved):

PR #4883, opened 2021-04-20
Update production setup docs for 3.0

### LINKED ISSUE #4700: Update prod setup docs for Python 3
Let's also combine it with updating our docs for more modern Linux distributions (CentOS 8, Debian 10/11, Ubuntu 20 LTS); possibly drop CentOS 7 unless it's easy to keep supported.

Since Python 3.9 is required and not all distributions include it out of the box, we need to check what's the best way to install this on a production system.

Using [pyenv](https://github.com/pyenv/pyenv/#installation) would be one option, but this means they need to install uwsgi with `pip` as well in order to be linked against the correct Python libraries, so we need to provide a systemd startup script for it. The big advantage of this is that we have full control over the Python version, and can in fact make updating Python part of our upgrade guide, and thus we can go along with the yearly Python release cycle instead of being stuck with 3.9 for the next 5 years or so.

The alternative would be looking for official package sources (ie from the distribution or Python maintainers) for recent Python versions, like SCL/AppStreams on CentOS8, but this is likely more messy since it's *extremely* distribution-specific.

Let's also update the nginx/apache TLS settings to the latest mozilla intermediate preset. TLSv1 and TLSv1.1 from our current settings have no place in 2021 anymore.

ThiefMaster:
Only focusing on the basic setup for now; ie getting indico to run and using modern TLS settings.

- [x] debian10/nginx
- [x] debian10/apache
- [x] debian11/nginx
- [x] debian11/apache
- [x] ubuntu20/nginx
- [x] ubuntu20/apache
- [x] centos8/nginx
- [ ] centos8/apache ❌  
- [x] centos7/nginx
- [x] centos7/apache

Debian 11 isn't released yet but it'll be good to know if it actually works there as well since someone will certainly use it once it's out.

closes #4700


ThiefMaster:
Debian 11 works fine with minor tweaks:

- no nginx.org package yet, but the original debian package is recent enough
- that package uses `www-data` gid instead of `nginx`, so some adaptions are needed

In any case, nginx will most likely publish debian packages once bullseye is released, so this shouldn't matter.

ThiefMaster:
- CentOS Stream 8 doesn't have official Postgres 13 packages yet
- CentOS 8 has no packages for mod_xsendfile and mod_proxy_uwsgi yet (not event in EPEL, mod_xsendfile is in epel-testing though)

So for now these distributions are not supported. Once the necessary packages become available, they should work fine though.

Issue #4876, opened 2021-04-16
New search: Search only within a category


ThiefMaster:

javfg:
There are two parts to this issue:

1. Creating a `/category/<int:category_id>/search`, which shows a search page just like `/search`, and limits the output to the given category.
2. Adding a popup to the search field in the category view page (`/category/<int:category_id>`), which tells the user his search will be limited to that category, and allows him to search globally. This can be similar to GitHub's search in a repo:
![image](https://user-images.githubusercontent.com/6058151/115897016-4bc3c600-a45c-11eb-84fc-dbdf0ece85f2.png)

PR #4885, opened 2021-04-23
Add search within categories with limited scope

javfg:
Closes https://github.com/indico/indico/issues/4876.


indico/modules/search/controllers.py
```diff
@@ -63,11 +63,17 @@ class InternalSearch(IndicoSearchProvider):
...
+                total, results = InternalSearch.search_categories(page, query)
         elif object_type == SearchTarget.event:
-            total, results = InternalSearch.search_events(page, query)
+            if params.get('category'):
+                total, results = InternalSearch.search_category_events(page, query, params['category'])
+            else:
+                total, results = InternalSearch.search_events(page, query)
         else:
-            return 0, [], []
+            return 0, 0, [], []
```
javfg, line 76: Even if this commit ends up being discarded, keep in mind this line was returning an incorrect tuple @plourenco.

indico/modules/search/controllers.py
```diff
@@ -82,6 +88,19 @@ def search_categories(page, q):
         # this would be nicer but then we can't easily paginate...
         return results.total, DetailedCategorySchema(many=True).dump(results.items)
 
+    @staticmethod
+    def search_subcategories(page, q, category_id):
+        # TODO: This must filter subcategories.
+        results = (Category.query
```
ThiefMaster, line 94: You can use `category.deep_children_query` to get a query that searches only within children of that category

indico/modules/search/controllers.py
plourenco: Yes, `pages` was not supposed to be here, but it had to be added just recently. I'm updating this on the base branch and we'll rebase it.

indico/modules/search/client/js/components/SearchBox.jsx
```diff
@@ -7,30 +7,69 @@
...
+    </span>
+    <Label content={label} size="small" />
+  </div>
+);
+
+export default function SearchBox({onSearch, category}) {
   const [keyword, setKeyword] = useState('');
-  const handleChange = event => {
+  const options =
+    category.id !== 0
```
ThiefMaster, line 28: I'd prefer if we checked for the category's `is_root` attribute instead of relying on the ID. While I think even the most legacy databases had ID 0 for the root category, this is not guaranteed

indico/modules/search/client/js/components/SearchBox.jsx
```diff
@@ -7,30 +7,69 @@
...
 SearchBox.propTypes = {
   onSearch: PropTypes.func.isRequired,
+  category: PropTypes.object,
+};
+
+SearchBox.defaultProps = {
+  category: {
+    title: undefined,
+    id: undefined,
+  },
```
ThiefMaster, line 74: If we always pass it we can just make it required instead of providing defaults

indico/modules/search/client/js/components/SearchBox.jsx
```diff
@@ -7,30 +7,69 @@
...
+        fluid
+        selectFirstResult
       />
     </Form>
   );
 }
 
 SearchBox.propTypes = {
   onSearch: PropTypes.func.isRequired,
+  category: PropTypes.object,
```
ThiefMaster, line 67: `PropTypes.shape()` is usually preferred since it specifies what kind of object you expect

indico/modules/search/views.py
```diff
@@ -19,3 +20,17 @@ def _get_breadcrumbs(self):
...
         return self._get_page_content(params)
+
+
+class WPCategorySearch(WPCategory):
+    """WP for category-scoped search"""
+    template_prefix = 'search/'
+    bundles = ('module_search.js', 'module_search.css')
+
+    def __init__(self, rh, category, **kwargs):
+        WPCategory.__init__(self, rh, category, **kwargs)
```
ThiefMaster, line 31: not needed (just a reminder to remove this if it stays empty)

indico/modules/search/client/js/components/SearchBox.jsx
javfg: Ah yes, I was not sure about checking for id 0 or is_root, as I've seen those checks around. I'll change this.

indico/modules/search/client/js/components/SearchBox.jsx
```diff
@@ -7,30 +7,71 @@
...
   };
+
+  // form submit happens when no option is selected in search (eg. in home category)
   const handleSubmit = () => {
-    onSearch(keyword);
+    if (!keyword) {
+      return;
+    }
+
+    console.log('handleSubmit', keyword, category.isRoot);
```
ThiefMaster, line 44: 👀

indico/modules/search/client/js/components/SearchBox.jsx
```diff
@@ -7,30 +7,71 @@
...
+  </div>
+);
+
+export default function SearchBox({onSearch, category}) {
   const [keyword, setKeyword] = useState('');
-  const handleChange = event => {
+  const options = !category.isRoot
+    ? [
+        {value: 'category', title: category.title, label: 'In this category ↵'},
+        {value: 'global', title: 'Global', label: 'All Indico ↵'},
```
ThiefMaster, line 30: Needs i18n (I'd keep the icon out of the translatable string of course)

javfg: As discussed, we found it better to keep the `↵` inside the string for RLT languages.

indico/modules/search/client/js/components/SearchBox.jsx
```diff
@@ -7,30 +7,74 @@
...
   const [keyword, setKeyword] = useState('');
-  const handleChange = event => {
+  const options = !category.isRoot
+    ? [
+        {
+          value: 'category',
+          title: category.title,
+          label: Translate.string('In this category ↵'),
+        },
+        {value: 'global', title: 'Global', lalbe: Translate.string('All Indico ↵')},
```
plourenco, line 34:
```suggestion
        {value: 'global', title: 'Global', label: Translate.string('All Indico ↵')},
```

javfg: :grimacing:

plourenco (approved):

indico/modules/search/controllers.py
```diff
@@ -72,12 +78,18 @@ def _process(self):
 class InternalSearch(IndicoSearchProvider):
     def search(self, query, access, page=1, object_types=(), **params):
         if SearchTarget.category in object_types:
-            total, results = InternalSearch.search_categories(page, query)
+            if params.get('category'):
+                total, results = InternalSearch.search_subcategories(page, query, params['category'])
+            else:
+                total, results = InternalSearch.search_categories(page, query)
```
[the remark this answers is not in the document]
javfg: Indeed, this changed in https://github.com/indico/indico/pull/4885/commits/48c94e5a675ff798e8ef5583dc960a81c839fbcc.

indico/modules/search/client/js/components/SearchApp.jsx
```diff
@@ -182,10 +240,15 @@ export default function SearchApp() {
...
       </Grid.Column>
     </Grid>
   );
 }
+
+SearchApp.propTypes = {
+  category: PropTypes.shape({
+    id: PropTypes.number,
+    title: PropTypes.string,
+  }).isRequired,
```
ThiefMaster, line 253: Wouldn't it make more sense to pass a `null` category when not filtering by category (and make the attributes inside required in case one is passed?)

indico/modules/search/client/js/index.jsx
```diff
@@ -13,9 +13,12 @@ import SearchApp from './components/SearchApp';
 
 document.addEventListener('DOMContentLoaded', () => {
   const root = document.querySelector('#search-root');
+  const id = parseInt(root.dataset.id, 10) || undefined;
+  const title = root.dataset.title;
```
ThiefMaster, line 17: `const category = root.dataset.category ? JSON.parse(root.dataset.category) : null;`

and in the template put something like `data-category="{{ {'id': category.id, 'title': category.title} | tojson | forceescape }}"`

indico/modules/search/client/js/index.jsx
```diff
@@ -13,9 +13,12 @@ import SearchApp from './components/SearchApp';
...
+
   ReactDOM.render(
     <Router>
-      <SearchApp />
+      <SearchApp category={{id, title}} />
```
ThiefMaster, line 21: `category={category}`

indico/modules/search/client/js/components/SearchApp.jsx
javfg: Good point, I started with only the ID, and passing a null ID was being translated into an empty string in the backend db query.

Now that it's an object it makes more sense to make it `null`.

indico/modules/search/controllers.py
```diff
@@ -84,25 +90,33 @@ def search(self, query, access, page=1, object_types=(), **params):
...
         # XXX should we only show categories the user can access?
         # this would be nicer but then we can't easily paginate...
         res = DetailedCategorySchema(many=True).dump(results.items)
         return results.total, CategoryResultSchema(many=True).load(res)
 
     @staticmethod
-    def search_events(page, q):
+    def search_events(page, q, category_id):
+        filters = [Event.title_matches(q), Event.effective_protection_mode == ProtectionMode.public,
+                   ~Event.is_deleted]
```
ThiefMaster, line 111:
```suggestion
        filters = [
            Event.title_matches(q),
            Event.effective_protection_mode == ProtectionMode.public,
            ~Event.is_deleted
        ]
```

indico/modules/search/templates/category_search.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'layout/base.html' %}
+
+{% block page_class %}search-page{% endblock %}
+
+{% block title %}
+    {% trans %}Searching within{% endtrans %} "{{ category.title }}"
```
ThiefMaster, line 6:
```suggestion
    {% trans cat=category.title %}Searching within "{{ cat }}"{% endtrans %}
```

not sure if all languages put it at the end

indico/modules/search/templates/category_search.html
```diff
@@ -0,0 +1,11 @@
...
+{% endblock %}
+
+{% block content %}
+    <div id="search-root" data-id="{{ category.id }}" data-title="{{ category.title }}"></div>
```
ThiefMaster, line 10: see my comment above about using a single attribute with an object

indico/modules/search/client/js/index.jsx
javfg: I've made this same change to `SearchBox` and the `Category` base template.

ThiefMaster (approved):

PR #4886, opened 2021-04-28
Update the search API to support multiple types 

plourenco:
This PR finally lifts the restriction of a single type per search. 
This brings up some considerations about the current API such as receiving a `category` together with an `event` object type. Since the former should be queried directly from the InternalSearch (categories are not initially designed to be indexed by any elastic search service), we don't really have any simple means of merging results from two different sources (InternalSearch and the currently used search provider).

My approach keeps these are two different sources apart by ignoring any `object_types` and return only categories if that target is provided as a type. 
This should prevent us from running into the previous situation while preserving the possibility of querying categories directly from the `InternalSearch`.

In other news, the `InternalSearch` was kept as originally, only supporting a single `SearchTarget`.

In addition, merges contributions with sub-contributions.

### DIFF
--- a/indico/modules/search/blueprint.py
+++ b/indico/modules/search/blueprint.py
@@ -15,5 +15,4 @@ _bp = IndicoBlueprint('search', __name__, template_folder='templates', virtual_t
 _bp.add_url_rule('/search/', 'search', RHSearchDisplay)
 
 # Search APIs
-_bp.add_url_rule('/search/api/<any(event,category,contribution,attachment,event_note):type>',
-                 'api_search', RHAPISearch)
+_bp.add_url_rule('/search/api', 'api_search', RHAPISearch)
--- a/indico/modules/search/blueprint.py
+++ b/indico/modules/search/blueprint.py
@@ -15,4 +15,4 @@ _bp = IndicoBlueprint('search', __name__, template_folder='templates', virtual_t
 _bp.add_url_rule('/search/', 'search', RHSearchDisplay)
 
 # Search APIs
-_bp.add_url_rule('/search/api', 'api_search', RHAPISearch)
+_bp.add_url_rule('/search/api/search', 'api_search', RHAPISearch)
--- a/indico/modules/search/controllers.py
+++ b/indico/modules/search/controllers.py
@@ -56,7 +56,7 @@ class RHAPISearch(RH):
         if not search_provider or SearchTarget.category in type:
             search_provider = InternalSearch
         access = get_groups(session.user) if session.user else []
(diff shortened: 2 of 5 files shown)


indico/modules/search/blueprint.py
```diff
@@ -15,5 +15,4 @@
 _bp.add_url_rule('/search/', 'search', RHSearchDisplay)
 
 # Search APIs
-_bp.add_url_rule('/search/api/<any(event,category,contribution,attachment,event_note):type>',
-                 'api_search', RHAPISearch)
+_bp.add_url_rule('/search/api', 'api_search', RHAPISearch)
```
ThiefMaster, line 18: I'm not a big fan of this URL

```suggestion
_bp.add_url_rule('/search/api/search', 'api_search', RHAPISearch)
```

indico/modules/search/controllers.py
```diff
@@ -40,17 +40,23 @@ def _process(self):
...
         'page': fields.Int(missing=1),
         'q': fields.String(required=True),
+        'type': fields.List(EnumField(SearchTarget), missing=None)
     }, location='query', unknown=INCLUDE)
-    @use_kwargs({'type': EnumField(SearchTarget)}, location='view_args')
-    def _process(self, type, page, q, **params):
+    def _process(self, page, q, type, internal, **params):
         search_provider = get_search_provider()
-        if not search_provider or type == SearchTarget.category:
+        if not search_provider or SearchTarget.category in type:
```
ThiefMaster, line 56: Maybe this is clearer since you won't get useful results when you pass multiple types including `category`?

```suggestion
        if not search_provider or type == [SearchTarget.category]:
```

plourenco: In this case, it may pass them through the actual provider which should have no entries either, but who knows.
Anyway, I find it equally reasonable.

indico/modules/search/controllers.py
```diff
@@ -40,17 +40,24 @@ def _process(self):
...
+    """
+
     @use_kwargs({
         'page': fields.Int(missing=1),
         'q': fields.String(required=True),
+        'type': fields.List(EnumField(SearchTarget), missing=None)
     }, location='query', unknown=INCLUDE)
-    @use_kwargs({'type': EnumField(SearchTarget)}, location='view_args')
-    def _process(self, type, page, q, **params):
+    def _process(self, page, q, type, internal, **params):
```
ThiefMaster, line 55: I don't think this should be there:

```suggestion
    def _process(self, page, q, type, **params):
```

PR #4887, opened 2021-04-29
Move placeholders to the search provider

plourenco:


indico/modules/search/schemas.py
```diff
@@ -19,6 +19,11 @@
 from indico.web.flask.util import url_for
 
 
+class PlaceholderSchema(mm.Schema):
+    key = fields.String()
+    label = fields.String()
```
ThiefMaster, line 24: shouldn't they be `required=True`?

plourenco: Yes, I didn't even remember they weren't by default

indico/modules/search/controllers.py
```diff
@@ -66,6 +66,13 @@ def _process(self, page, q, type, **params):
         }
 
 
+class RHAPISearchPlaceholders(RH):
+    def _process(self):
+        search_provider = get_search_provider()
+        placeholders = search_provider().get_placeholders()
+        return jsonify(placeholders if not PlaceholderSchema(many=True).validate(placeholders) else [])
```
ThiefMaster, line 73: I don't think we really need this validation - it's data coming from the plugin, that's most likely fully static - so unlike data returned from an actual search engine with extra metadata we don't want to show, I'd expect plugins to always return clean data here.

PR #4892, opened 2021-05-05
Create search sort options

plourenco:
imagem

### DIFF
--- a/indico/modules/search/client/js/components/SearchApp.jsx
+++ b/indico/modules/search/client/js/components/SearchApp.jsx
@@ -6,11 +6,12 @@
 // LICENSE file for more details.
 
 import searchURL from 'indico-url:search.api_search';
+import searchOptionsURL from 'indico-url:search.api_search_options';
 import searchPageURL from 'indico-url:search.search';
 
 import PropTypes from 'prop-types';
 import React, {useEffect, useMemo, useState} from 'react';
-import {Grid, Loader, Menu, Message} from 'semantic-ui-react';
+import {Dropdown, Grid, Loader, Menu, Message} from 'semantic-ui-react';
 
 import {useIndicoAxios, useQueryParams} from 'indico/react/hooks';
 import {Param, Translate} from 'indico/react/i18n';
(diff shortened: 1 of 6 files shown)


indico/modules/search/client/js/components/SearchApp.jsx
```diff
@@ -123,10 +124,40 @@ NoResults.defaultProps = {
   query: undefined,
 };
 
+function SearchOptions({sort, sortOptions, onSortChange}) {
+  const selected = sortOptions.find(x => x.key === sort) || sortOptions[0];
+
+  return !sortOptions.length ? null : (
```
ThiefMaster, line 130: IMO this would be nicer:

```js
if (!sortOptions.length) {
  return null;
}

return ...;
```

indico/modules/search/client/js/components/SearchBar.jsx
```diff
@@ -81,4 +77,9 @@ export default function SearchBar({onSearch, searchTerm}) {
 SearchBar.propTypes = {
   onSearch: PropTypes.func.isRequired,
   searchTerm: PropTypes.string.isRequired,
+  placeholders: PropTypes.array,
```
ThiefMaster, line 80: `PropTypes.arrayOf(PropTypes.shape(...))` would be cleaner

indico/modules/search/client/js/components/SearchApp.jsx
plourenco: It's funny because I changed that shortly before pushing 😄

indico/modules/search/client/js/components/SearchApp.jsx
```diff
@@ -175,10 +176,44 @@ ResultHeader.defaultProps = {
...
+        onChange={(e, data) => onSortChange(data.value)}
+        options={sortOptions.map(x => ({value: x.key, text: x.label}))}
+      />
+    </div>
+  );
+}
+
+SearchOptions.propTypes = {
+  sort: PropTypes.string,
+  sortOptions: PropTypes.array,
```
ThiefMaster, line 199: `.arrayOf(PropTypes.shape(...))`

indico/modules/search/client/js/components/SearchApp.jsx
```diff
@@ -175,10 +176,49 @@ ResultHeader.defaultProps = {
...
+  }
+
+  const selected = sortOptions.find(x => x.key === sort) || sortOptions[0];
+
+  return (
+    <div styleName="search-options">
+      <Dropdown
+        text={Translate.string('Sort by: {value}', {value: selected.label})}
+        onChange={(e, data) => onSortChange(data.value)}
+        options={sortOptions.map(x => ({value: x.key, text: x.label}))}
```
ThiefMaster, line 191:
```suggestion
        options={sortOptions.map(x => ({value: x.key, text: x.label}))}
        selectOnNavigation={false}
        selectOnBlur={false}
        value={selected.key}
```

This ensures the active option is shown in bold, and using the cursor keys without actively choosing an option won't immediately submit a query.

indico/modules/search/client/js/components/SearchApp.jsx
```diff
@@ -175,10 +176,49 @@ ResultHeader.defaultProps = {
...
+  if (!sortOptions.length) {
+    return null;
+  }
+
+  const selected = sortOptions.find(x => x.key === sort) || sortOptions[0];
+
+  return (
+    <div styleName="search-options">
+      <Dropdown
+        text={Translate.string('Sort by: {value}', {value: selected.label})}
```
ThiefMaster, line 189: Probably a matter of preference, but just putting `text={value}` here doesn't look bad either:

image

plourenco: I like you idea but it seems less user-friendly and more hidden from the regular user.

PR #4903, opened 2021-05-12
Add event-scoped search

### LINKED ISSUE #4902: Search: Event-scoped search
The old search plugin displayed a search box (similar to what we have in categories) within an event (if enabled - the setting for that is in the core).

It would be nice to restore this with the new search. We probably need a more lightweight version of the search results; some things that are not needed are:

- aggregations
- category breadcrumbs
- category/event results (categories are useless, and the event is always the current one)
- event note results (the UI is only shown in conferences, but in conferences we don't even show the UI to add/edit notes)

ThiefMaster:
closes #4902

image

image

### DIFF
--- a/indico/modules/search/client/js/components/SearchApp.jsx
+++ b/indico/modules/search/client/js/components/SearchApp.jsx
@@ -37,12 +37,12 @@ const camelizeValues = obj =>
     {}
   );
 
-function useSearch(url, query, type, categoryId) {
+function useSearch(url, query, type, scope = {}) {
   const [page, setPage] = useState(1);
 
   const {data, error, loading, lastData} = useIndicoAxios({
     url,
-    options: {params: {...query, type, page, category_id: categoryId}},
+    options: {params: {...query, type, page, ...scope}},
     forceDispatchEffect: () => query?.q,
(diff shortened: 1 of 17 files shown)


indico/modules/search/client/js/components/SearchApp.jsx
```diff
@@ -222,28 +223,49 @@ export default function SearchApp({category}) {
     trigger: 'once',
     camelize: true,
   });
+  let scope = {};
+  if (eventId !== null) {
+    scope = {event_id: eventId};
+  } else if (category !== null) {
+    scope = {category_id: category.id};
```
plourenco, line 230: Good idea to merge these into a scope argument in the hook above. But do we think we're going to need this scope check? As it doesn't seem wrong to send both at the same time. Or, even if one or both are undefined, I believe `useIndicoAxios` or the axios hook seems to correctly remove any undefined query params.

ThiefMaster: Yeah, might work, but if for whatever reason both an event and a category gets passed (not that doing that would make any sense) you might end up sending both..

PR #4905, opened 2021-05-17
Support prev/next pagination and use it in InternalSearch

ThiefMaster:
This allows us to perform `can_access()` checks in Indico without getting terrible performance.

image

Event on an instance as big as our production one searching by event title takes less than 1.5 seconds - and instances that big aren't even expected to use the internal search.

### DIFF
--- a/indico/modules/search/controllers.py
+++ b/indico/modules/search/controllers.py
@@ -62,7 +62,7 @@ class RHAPISearch(RH):
     """
 
     @use_kwargs({
-        'page': fields.Int(missing=1),
+        'page': fields.Int(missing=None),
         'q': fields.String(required=True),
         'type': fields.List(EnumField(SearchTarget), missing=None)
     }, location='query', unknown=INCLUDE)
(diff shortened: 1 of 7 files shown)


indico/modules/search/controllers.py
```diff
@@ -84,50 +69,79 @@ def _process(self):
 
 
 class InternalSearch(IndicoSearchProvider):
-    def search(self, query, access, page=1, object_types=(), **params):
+    def search(self, query, user, page=None, object_types=(), **params):
```
plourenco, line 72:
```suggestion
    def search(self, query, user=None, page=None, object_types=(), **params):
```

indico/modules/search/controllers.py
```diff
@@ -84,50 +69,79 @@ def _process(self):
...
+
+        def _can_access(obj):
+            return obj.effective_protection_mode == ProtectionMode.public or obj.can_access(user, allow_admin=False)
+
+        res = get_n_matching(query, self.RESULTS_PER_PAGE + 1, _can_access, prefetch_factor=20)
+
+        if len(res) > self.RESULTS_PER_PAGE:
+            # we queried 1 more so we can see if there are more results available
+            assert len(res) == self.RESULTS_PER_PAGE + 1
+            del res[-1]
```
plourenco, line 108: Feel free to ignore this if you prefer it this way. But we could in principle just trim it as it seems harmless.

```suggestion
            del res[self.RESULTS_PER_PAGE:]
```

plourenco (approved):

indico/modules/search/controllers.py
ThiefMaster: yeah, i thought the same, but i was too lazy to test the alternative ;) (i always want to be sure to avoid off-by-one errors when using slices)

Issue #4678, opened 2020-10-28
Use stricter Undefined in Jinja


ThiefMaster:
This should ideally not break much, but by doing it early in 3.0 development we'll most likely find out if anything does break because we are accessing undefined variables in templates.

TBD: Whether to use `StrictUndefined` or a custom `Undefined` subclass which is only a little bit stricter than the default. The disadvantage of `StrictUndefined` is that it is *really* strict and fails on stuff like `{% if somedict.foo %}` if it does not have a `foo` key, which is a bit annoying.

We could probably use a custom subclass like this:

```python
class SemiStrictUndefined(Undefined):
    __slots__ = ()
    # like StrictUndefined but without failing on truthiness check and comparisons
    __iter__ = __str__ = __len__ = Undefined._fail_with_undefined_error
    __hash__ = Undefined._fail_with_undefined_error
```

It could be nice to customize it even more so doing anything with an undefined top-level variable fails (even bool/eq checks), but not for items/attrs.

PR #4906, opened 2021-05-18
Fix undefined id in abstract fields sortable list

plourenco:
You can reproduce this error by clicking "Abstract fields" in "Cal for Abstracts" management, let alone add any fields (as `disabled_until_change` was undefined as well).

I believe this was triggered when #4678 was introduced as `id` was previously being incorrectly sent as undefined as part of the sortable list.

### DIFF
--- a/indico/web/templates/forms/form_common_fields_first.html
+++ b/indico/web/templates/forms/form_common_fields_first.html
@@ -1,12 +1,12 @@
 {% from 'forms/_form.html' import form_header, form_footer, form_rows %}
 
+{% set common_fields = form._common_fields|default([]) %}
 {{ form_header(form) }}
-{{ form_rows(form, fields=form._common_fields) }}
-{{ form_rows(form, skip=form._common_fields) }}
+{{ form_rows(form, fields=common_fields) }}
+{{ form_rows(form, skip=common_fields) }}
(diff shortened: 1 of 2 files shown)


indico/web/templates/forms/form_common_fields_first.html
```diff
@@ -5,7 +5,7 @@
 {{ form_rows(form, skip=form._common_fields) }}
 {% call form_footer(form) %}
     <input class="i-button big highlight" type="submit" value="{% trans %}Save{% endtrans %}"
-           {% if disabled_until_change %}data-disabled-until-change{% endif %}>
+           {% if disabled_until_change|default(true) %}data-disabled-until-change{% endif %}>
```
ThiefMaster, line 8: I don't think we ever pass `disabled_until_change` to this template... do we really want to start defaulting to true when the previous implicit default was falsy? We could just remove this altogether unless all the places where it's used are fine with disabled-until-change behavior

plourenco: I initially thought it would make more sense to have it `true`, but now that you mention it there might be indeed some occasions where we want to immediately save. I can see some usages of `disabled_until_change` being both `true` and `false`. So I can only think of defaulting it to false.

PR #4909, opened 2021-05-19
Remove search descriptions without content

plourenco:
Highlight's div class summary applies a margin-bottom, when we render it empty, it causes a double margin.

imagem

imagem

Up for discussion, but I went for keeping this responsibility on the parent, i.e. decide whether or not highlighted descriptions should render.

### DIFF
--- a/indico/modules/search/client/js/components/results/Contribution.jsx
+++ b/indico/modules/search/client/js/components/results/Contribution.jsx
@@ -37,7 +37,7 @@ export default function Contribution({
         <a href={url}>{title}</a>
       </List.Header>
       <List.Description styleName="description">
-        <Highlight text={description} highlight={highlight.description} />
+        {description && <Highlight text={description} highlight={highlight.description} />}
         {persons.length !== 0 && (
           <List.Item>
             <PersonList persons={persons} />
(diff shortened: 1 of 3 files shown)


indico/modules/search/client/js/components/results/Contribution.jsx
```diff
@@ -37,7 +37,7 @@ export default function Contribution({
         <a href={url}>{title}</a>
       </List.Header>
       <List.Description styleName="description">
-        <Highlight text={description} highlight={highlight.description} />
+        {description && <Highlight text={description} highlight={highlight.description} />}
```
ThiefMaster, line 40: I guess we always get a description as well as a highlight?

ThiefMaster: What do you think about returning `null` in `Highlight` if there's neither text nor a highlight?

plourenco: I was undecided... it's the balance between responsibility and verboseness. I think it made more sense to have the parent decide whether or not the child should render. On the other hand, the highlight shouldn't be empty. So if you agree, it's fine for me.

ThiefMaster: I don't have a strong opinion either, but a slight preference on having the check in the component since that's what's rendering the `<div styleName="summary">` causing the margin anyway.

PR #4910, opened 2021-05-21
Add support for survey cloning

### LINKED ISSUE #2045: Clone surveys while cloning an event
We used to have this with the old evaluation system. Probably only few people ever used it but it would be nice to add it back at some point.

plourenco:
Closes #2045

### DIFF
--- a/indico/modules/events/clone.py
+++ b/indico/modules/events/clone.py
@@ -13,8 +13,6 @@ from indico.modules.events.models.events import EventType
 from indico.modules.events.models.persons import EventPerson, EventPersonLink
 from indico.modules.events.models.principals import EventPrincipal
 from indico.modules.events.sessions import session_settings
-from indico.modules.events.surveys.models.items import SurveyItem
-from indico.modules.events.surveys.models.surveys import Survey
 from indico.modules.events.util import track_location_changes
 from indico.util.i18n import _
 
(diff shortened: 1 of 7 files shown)


indico/modules/events/clone.py
```diff
@@ -151,3 +152,46 @@ def _clone_acl(self, new_event, event_exists):
...
+        survey_attrs = get_simple_column_attrs(Survey) - {'uuid', 'start_dt', 'end_dt', '_last_friendly_submission_id'}
+        for old_survey in self.old_event.surveys:
+            survey = Survey()
+            survey.populate_from_attrs(old_survey, survey_attrs)
+            item_map = {}
+            for old_item in old_survey.items:
+                item = self._clone_item(old_item)
+                if old_item.parent:
+                    assert old_item.parent != old_item
+                    if not item_map.get(old_item.parent):
```
ThiefMaster, line 188: `if old_item.parent not in item_map:`

indico/modules/events/clone.py
```diff
@@ -151,3 +152,46 @@ def _clone_acl(self, new_event, event_exists):
...
+                        item_map[old_item.parent] = self._clone_item(old_item.parent)
+                    item.parent = item_map[old_item.parent]
+                item_map[old_item] = item
+            survey.items.extend(item_map.values())
+            new_event.surveys.append(survey)
+
+    def _clone_item(self, old_item):
+        item = old_item.__class__()
```
ThiefMaster: I guess we are using polymorphism here so it's not always the same class? I'd probably be more explicit and do `item_cls = type(old_item)`, and then `item = item_cls()` but no strong opinion.

plourenco: Yes, that's right. Sure, why not!

indico/modules/events/surveys/clone_test.py
```diff
@@ -0,0 +1,30 @@
...
+from indico.modules.events.surveys.clone import EventSurveyCloner
+from indico.modules.events.surveys.models.items import SurveyItem, SurveyQuestion, SurveySection
+from indico.modules.events.surveys.models.surveys import Survey
+
+
+def test_survey_clone(db, create_event, dummy_event):
+    survey = Survey(title='test')
+    section = SurveySection(title='test', display_as_section=True)
+    survey.items.append(section)
+    SurveyQuestion(title='question', field_type='text', is_required=True, parent=section)
```
ThiefMaster, line 18: Maybe we could add a `SurveyText` item as well?

plourenco: I was actually thinking about parametrizing, but it's not that easy with parents. The latest test should ensure more correctness.

indico/modules/events/surveys/clone.py
```diff
@@ -0,0 +1,56 @@
...
+    def has_conflicts(self, target_event):
+        return self._has_content(target_event)
+
+    def run(self, new_event, cloners, shared_data, event_exists=False):
+        with db.session.no_autoflush:
+            self._clone_surveys(new_event)
+        db.session.flush()
+
+    def _has_content(self, event):
+        return len([survey for survey in event.surveys if not survey.is_deleted])
```
ThiefMaster, line 33:
```suggestion
        return any(survey for survey in event.surveys if not survey.is_deleted)
```

indico/modules/events/surveys/clone.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def _clone_surveys(self, new_event):
+        survey_attrs = get_simple_column_attrs(Survey) - {'uuid', 'start_dt', 'end_dt', '_last_friendly_submission_id'}
+        for old_survey in self.old_event.surveys:
```
ThiefMaster, line 37: `Event.surveys` doesn't exclude deleted surveys, so we need to skip those explicitly

```suggestion
        for old_survey in self.old_event.surveys:
            if old_survey.is_deleted:
                continue
```

ThiefMaster (approved):

PR #4915, opened 2021-05-25
Fixes to iCalendar exports

javfg:
* Show agenda link back in description, as Exchange does not parse `URL` field
* Remove duplicate `DESCRIPTION` field in some entities
* Do not export unscheduled contributions in sessions

### DIFF
--- a/indico/modules/events/ical.py
+++ b/indico/modules/events/ical.py
@@ -61,7 +61,7 @@ def generate_basic_component(entity, uid=None, url=None):
             # this happens if desc_text only contains a html comment
             pass
     if url:
-        description.append(f'Agenda: {url}')
+        description.append(url)
     if description:
         component.add('description', '\n'.join(description))
 
--- a/indico/modules/events/ical.py
+++ b/indico/modules/events/ical.py
@@ -36,10 +36,10 @@ def generate_basic_component(entity, uid=None, url=None):
     if uid:
         component.add('uid', uid)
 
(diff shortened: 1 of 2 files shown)


indico/modules/events/ical.py
```diff
@@ -87,7 +89,8 @@ def generate_event_component(event, user=None):
         as_list=True
     ):
         data.update(update)
-    component.add('description', data['description'])
+    if data['description']:
+        component['description'] = data['description']
```
ThiefMaster, line 93: does this work even if there is no description yet? (can this happen?)

indico/modules/events/sessions/ical.py
```diff
@@ -42,8 +43,11 @@ def session_to_ical(session, detailed=False):
         calendar.add_component(component)
     else:
         from indico.modules.events.contributions.ical import generate_contribution_component
+
+        contributions = (Contribution.query.with_parent(session)
+                         .filter(Contribution.is_scheduled).all())
```
ThiefMaster, line 48:
```suggestion
        contributions = (Contribution.query.with_parent(session)
                         .filter(Contribution.is_scheduled)
                         .all())
```

indico/modules/events/ical.py
javfg: It works, but it should not happen except for entities without any URL, as the description will include that as a minimum now.

ThiefMaster (approved):

PR #4918, opened 2021-05-26
Fix character count inconsistencies while dealing with line breaks

plourenco:
May-26-2021 16-13-57

I first went for a client-sided approach (fix it before sending), but it's not that easy with multiple change listeners. Then I noticed on top of that we have a SoftLength validator which seemed like a good candidate to accept line-breaks as a single character.

### DIFF
--- a/indico/web/forms/validators.py
+++ b/indico/web/forms/validators.py
@@ -380,8 +380,12 @@ class SoftLength(Length):
     """
 
     def __call__(self, form, field):
+        orig_data = field.data
         field.data = re.sub(r'(\r\n|\r)', '\n', field.data)
-        super().__call__(form, field)
+        try:
+            super().__call__(form, field)
+        finally:
+            field.data = orig_data
 
 
 class SecurePassword:
--- a/indico/web/forms/validators.py
+++ b/indico/web/forms/validators.py
@@ -379,6 +379,10 @@ class SoftLength(Length):
     where surrounding whitespace is stripped before validation.
     """
 
+    def __call__(self, form, field):
+        field.data = re.sub(r'(\r\n|\r)', '\n', field.data)
+        super().__call__(form, field)
+
 
 class SecurePassword:
     """Validate that a string is a secure password."""
--- a/indico/web/forms/validators.py
+++ b/indico/web/forms/validators.py
@@ -381,7 +381,8 @@ class SoftLength(Length):
 
     def __call__(self, form, field):
         orig_data = field.data
-        field.data = re.sub(r'(\r\n|\r)', '\n', field.data)
+        if field.data is not None:
+            field.data = re.sub(r'(\r\n|\r)', '\n', field.data)
         try:
             super().__call__(form, field)
         finally:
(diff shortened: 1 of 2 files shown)


indico/web/forms/validators.py
```diff
@@ -379,6 +379,10 @@ class SoftLength(Length):
     where surrounding whitespace is stripped before validation.
     """
 
+    def __call__(self, form, field):
+        field.data = re.sub(r'(\r\n|\r)', '\n', field.data)
```
ThiefMaster, line 383: I'm not sure I like the idea of modifying a field's data from within a validator... AFAIK that's what filters are meant for (defined on the field level - searching for `filters=[` will give you a few examples)

ThiefMaster:
IMO this change deserves a changelog entry (in bugfixes):

```rst
- Correctly count line breaks in length-limited abstracts (:pr:`4918`)
```

indico/web/forms/validators.py
plourenco: Me neither, my initial idea wasn't to modify it within a validator but to make it ignore any double line-breaks. Filters definitely seem better suited here.

plourenco: The downside of filtering (instead of ignoring) is that we need to make sure this filter is applied anywhere we used a `TextArea` (or any of its superclasses) with a length validator, which isn't that easy to control. Even adding the filter by default in `IndicoMarkdownField` doesn't prevent this from happening with `TextArea` in the future. Should I just worry about `IndicoMarkdownField` for now?

ThiefMaster: The alternative would be something like this:

```python
old_data = field.data
field.data = re.sub(...)
try:
    super()...
finally:
    field.data = old_data
```

plourenco: Seems fair, and simpler than cloning the field.

indico/web/forms/validators.py
```diff
@@ -379,6 +379,14 @@ class SoftLength(Length):
     where surrounding whitespace is stripped before validation.
     """
 
+    def __call__(self, form, field):
+        _data = field.data
```
ThiefMaster, line 383: I don't really like `_foo` names inside local scopes ;)

What's wrong with `orig_data` which is clearer about its purpose IMO?

plourenco: 👍

indico/web/forms/validators.py
```diff
@@ -379,6 +379,14 @@ class SoftLength(Length):
     where surrounding whitespace is stripped before validation.
     """
 
+    def __call__(self, form, field):
+        orig_data = field.data
+        field.data = re.sub(r'(\r\n|\r)', '\n', field.data)
```
ThiefMaster, line 384: I think `field.data` might be `None` in some cases - at least the original Length validator takes that into account, so maybe safer to do so here as well..

```suggestion
        if field.data is not None:
            field.data = re.sub(r'(\r\n|\r)', '\n', field.data)
```

ThiefMaster (approved):

PR #4921, opened 2021-05-27
Convert translatable strings to en_US

### LINKED ISSUE #4639: Homogenise msgid strings to AE
As a matter of fact, Indico codebase and base textstrings have been mostly written in in American English (as opposed to British Englsh, [read more](https://www.boredpanda.com/british-american-english-differences-language/). As a consequence, the basic (untranslated) version of Indico is **en_US in general, but not always**. (An [en_GB section on Transifex](https://www.transifex.com/indico/indico/translate/#en_GB) was opened recently to provide a BE version of Indico.) 
The issue is to make all translatable text in Indico correct en_US, as it is presently a mixture of BE and AE writing. This is particularly important, because the (untranslated) base textstrings of Indico are used as identifiers (**msgid**) for the translation via PO/MO files. Hence they have an impact on _all_ translated languages, when they are updated. An appropriate automatic method to avoid too much overhead for each translator group should be applied.

From earlier exchanges about the topic, for reference:
> Organize (https://www.merriam-webster.com/dictionary/organise)
> Synchronize, authorize, customize and so on, and their derivatives.
> 
> Then there are "center" (centre), "gray" (grey) and "color" (colour) for example.
> 
> I am not sure about "convener" (convenor), which seems to be a matter of taste rather than geography.
> 
> Programme / program.

The proposal is to finish the cleanup by Indico 3.0.

plourenco:
Closes #4639 

A starting point to convert some of our translatable strings in en_GB to the standardized en_US.

### DIFF
--- a/indico/web/client/js/utils/palette.js
+++ b/indico/web/client/js/utils/palette.js
@@ -36,7 +36,7 @@ export const SUIPalette = {
   purple: Translate.string('Purple'),
   pink: Translate.string('Pink'),
   brown: Translate.string('Brown'),
-  grey: Translate.string('Grey'),
+  gray: Translate.string('Gray'),
   black: Translate.string('Black'),
 };
 export default Palette;
--- a/indico/web/client/js/utils/palette.js
+++ b/indico/web/client/js/utils/palette.js
@@ -36,7 +36,7 @@ export const SUIPalette = {
   purple: Translate.string('Purple'),
   pink: Translate.string('Pink'),
   brown: Translate.string('Brown'),
-  gray: Translate.string('Gray'),
+  grey: Translate.string('Gray'),
   black: Translate.string('Black'),
 };
 export default Palette;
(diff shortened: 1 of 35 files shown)


indico/web/client/js/utils/palette.js
```diff
@@ -36,7 +36,7 @@ export const SUIPalette = {
   purple: Translate.string('Purple'),
   pink: Translate.string('Pink'),
   brown: Translate.string('Brown'),
-  grey: Translate.string('Grey'),
+  gray: Translate.string('Gray'),
```
ThiefMaster, line 39:
```suggestion
  grey: Translate.string('Gray'),
```

The key can't be changed; it's one of the SUI colors.

ThiefMaster:
I dumped the msgids in a file and pasted them into word to see what it complains about:

- programme -> program
- favourite -> favorite
- synchronisation -> synchronization

and some unrelated typos which we could also fix:

- "You should upload a CSV or TXT file with ~~with~~ one email address per line."
- occuring -> occurring

PR #4924, opened 2021-05-31
Enhance the PrincipalListField

### LINKED ISSUE #4917: User search dialog: Make better use of screen space
On larger screens the user search dialog - especially if there are many search results - uses lots of vertical spaces and rather little horizontal space. It would be nice to make it a bit more responsive and make better use of screens pace, e.g. by putting the results on the side (probably needs some playing around to see what looks & feels best).

image

plourenco:
Closes #4917 
These changes convert the PrincipalListField into a column-based layout.
Also includes a small staged users refactoring into a new component and shows favourite users by default.

### DIFF
--- a/indico/web/client/js/react/components/principals/Search.jsx
+++ b/indico/web/client/js/react/components/principals/Search.jsx
@@ -13,7 +13,7 @@ import userSearchInfoURL from 'indico-url:users.user_search_info';
 import {FORM_ERROR} from 'final-form';
 import _ from 'lodash';
 import PropTypes from 'prop-types';
-import React, {useContext, useState} from 'react';
+import React, {useContext, useRef, useState} from 'react';
 import {Form as FinalForm} from 'react-final-form';
 import Overridable from 'react-overridable';
 import {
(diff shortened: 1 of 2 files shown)


indico/web/client/js/react/components/principals/Search.jsx
```diff
@@ -221,30 +221,86 @@ const searchFactory = config => {
...
     eventId,
   }) => {
-    const [result, setResult] = useState(null);
+    const [loading, setLoading] = useState(false);
+    const [result, _setResult] = useState(null);
+    const lastResult = useRef(null);
+    const setResult = value => {
+      lastResult.current = result;
+      _setResult(value);
+    };
```
ThiefMaster, line 230: just wondering.. do we need a ref for this? or can we use useReducer for result+lastResult where sending a new result sets the last result to the previous one?

plourenco: Yes, I don't see why not, `useRef` just seems simpler as it won't have any side-effects (we just want to silently keep track of the results), but since both are being assigned at the exact same time it's pretty much the same (without a reducing logic):
```
  lastResult.current = result;
  _setResult(value)
```
I can do both.

ThiefMaster: yeah nothing wrong with keeping it as it is right now :)

indico/web/client/js/react/components/principals/Search.jsx
```diff
@@ -221,30 +221,86 @@ const searchFactory = config => {
...
+    const handleSearch = async data => {
+      setLoading(true);
+      try {
+        await runSearch(data, setResult, withEventPersons, eventId);
+      } finally {
+        setLoading(false);
+      }
+    };
+    const favoriteResults = Object.values(favorites);
```
ThiefMaster, line 239: this breaks on pages that do not use favorites (e.g. the favorite user management page).

This avoids it:

```suggestion
    const favoriteResults = favorites ? Object.values(favorites) : [];
```

It would be nice to show something like "n favorite users" instead of "n users found" when we show the favorites initially... (and if there are no favorites, something else)

plourenco: Thanks, hadn't realized this could be null.
I thought about something like "Suggested users" but not sure it's worth it since `noResultsText` is actually a prop.

ThiefMaster: You could add `initialResultsText` (for the favorites - in case of groups we don't have anything there so we'd just leave it empty) and `notSearchedText` that's shown when there are no initial results nor a search.

ThiefMaster:
In the old version the smaller-screen version made the input fields full width. Any way to keep this behavior? It was kind of nice, especially on phone-sized screens.

image

image

indico/web/client/js/react/components/principals/Search.jsx
plourenco: I tried using a different text for the initial results, but having a different string doesn't look great as there's still a lot of white space. 
As an alternative, I tried to initially hide the results column if there are neither results nor favourites - but only if the search is "pristine". How does it look to you?

The only part left to address is `initialResultsText` **when the favourites are being shown**, where I can have something as "n favorite users".

ThiefMaster: user search is currently broken (undefined results in SearchResults), but besides the fact that it currently shows "no results found" while searching (ldap group search without "exact match" takes a while) I kind of like it..

ThiefMaster: works pretty nicely now, just one small thing: while the right side is disabled because of a running search, the widgets there are still active - ie i can select users and get tooltips when hovering the "user exists in event" icon

indico/web/client/js/react/components/principals/Search.jsx
```diff
@@ -263,6 +264,11 @@ const searchFactory = config => {
               onAdd={onAdd}
               onRemove={onRemove}
               isAdded={isAdded}
+              resultsText={total =>
+                pristine.current
+                  ? Translate.string('{total} favorite users', {total})
```
ThiefMaster, line 269: this needs pluralization

indico/web/client/js/react/components/principals/Search.jsx
plourenco: I disabled the pointer-events. Not exactly sure if it's worth changing the faded "no results found" while the next search state is still loading since I'm not sure what to set there instead, unless I got rid of it.

ThiefMaster: I like it the way it is now :)

You could add `user-select: none;` though; it doesn't really matter but being able to select content by clicking and dragging the cursor on a selected area feels a bit strange.

PR #4925, opened 2021-06-01
Refactor ICSCalendarLink

plourenco:
We previously got some feedback on the dropdown being a multi-step selection that could be simplified into fewer clicks.

These changes update the popup into a simplified multi-state (tabbed liked) component.

imagem

### DIFF
--- a/indico/modules/events/client/js/header.jsx
+++ b/indico/modules/events/client/js/header.jsx
@@ -19,7 +19,14 @@ document.addEventListener('DOMContentLoaded', () => {
   }
 
   const {eventId, eventContribCount} = calendarContainer.dataset;
-  const options = [{key: 'event', text: Translate.string('Event'), extraParams: {}}];
+  const options = [
+    {
+      key: 'event',
+      text: Translate.string('Basic'),
+      description: Translate.string('Just the event.'),
+      extraParams: {},
+    },
+  ];
   if (parseInt(eventContribCount, 10) > 0) {
     options.push({
       key: 'contributions',
--- a/indico/modules/events/client/js/header.jsx
+++ b/indico/modules/events/client/js/header.jsx
@@ -32,12 +32,11 @@ document.addEventListener('DOMContentLoaded', () => {
     <ICSCalendarLink
       endpoint="events.export_event_ical"
       params={{event_id: eventId}}
-      renderButton={(onClick, classes) => (
+      renderButton={classes => (
         <IButton
           icon="calendar"
           dropdown
           classes={{'height-full': true, 'text-color': true, 'subtle': true, ...classes}}
-          onClick={onClick}
           title={Translate.string('Export')}
         />
       )}
--- a/indico/modules/events/client/js/header.jsx
+++ b/indico/modules/events/client/js/header.jsx
@@ -18,7 +18,7 @@ document.addEventListener('DOMContentLoaded', () => {
     return;
   }
 
-  const {eventId, eventContribCount} = calendarContainer.dataset;
+  const {eventId, eventContribCount, eventSessionBlockCount} = calendarContainer.dataset;
   const options = [
     {
       key: 'event',
@@ -27,18 +27,20 @@ document.addEventListener('DOMContentLoaded', () => {
       extraParams: {},
     },
   ];
+  if (parseInt(eventSessionBlockCount, 10) > 0) {
+    options.push({
+      key: 'sessions',
+      description: Translate.string('A detailed timetable containing individual session blocks.'),
+      text: Translate.string('Compact'),
+      extraParams: {scope: 'session'},
+    });
+  }
   if (parseInt(eventContribCount, 10) > 0) {
     options.push({
       key: 'contributions',
       text: Translate.string('Detailed'),
       description: Translate.string('A detailed timetable containing individual contributions.'),
       extraParams: {scope: 'contribution'},
-    });
-    options.push({
-      key: 'sessions',
-      description: Translate.string('A detailed timetable containing individual session blocks.'),
-      text: Translate.string('Compact'),
-      extraParams: {scope: 'session'},
     });
   }
 
--- a/indico/modules/events/client/js/header.jsx
+++ b/indico/modules/events/client/js/header.jsx
@@ -23,8 +23,15 @@ document.addEventListener('DOMContentLoaded', () => {
   if (parseInt(eventContribCount, 10) > 0) {
     options.push({
       key: 'contributions',
-      text: Translate.string('Detailed timetable'),
-      extraParams: {detail: 'contributions'},
+      text: Translate.string('Detailed'),
+      description: Translate.string('A detailed timetable scoped to contributions.'),
+      extraParams: {scope: 'contribution'},
+    });
+    options.push({
+      key: 'sessions',
+      description: Translate.string('A detailed timetable scoped to sessions.'),
+      text: Translate.string('Compact'),
+      extraParams: {scope: 'session'},
     });
   }
 
--- a/indico/modules/events/client/js/header.jsx
+++ b/indico/modules/events/client/js/header.jsx
@@ -30,7 +30,9 @@ document.addEventListener('DOMContentLoaded', () => {
   if (parseInt(eventSessionBlockCount, 10) > 0) {
     options.push({
       key: 'sessions',
-      description: Translate.string('A detailed timetable containing individual session blocks.'),
+      description: Translate.string(
+        'A detailed timetable containing individual session blocks and top-level contributions.'
+      ),
       text: Translate.string('Compact'),
       extraParams: {scope: 'session'},
     });
@@ -39,7 +41,9 @@ document.addEventListener('DOMContentLoaded', () => {
     options.push({
       key: 'contributions',
       text: Translate.string('Detailed'),
-      description: Translate.string('A detailed timetable containing individual contributions.'),
+      description: Translate.string(
+        'A detailed timetable containing all individual contributions.'
+      ),
       extraParams: {scope: 'contribution'},
     });
   }
--- a/indico/modules/events/client/js/header.jsx
+++ b/indico/modules/events/client/js/header.jsx
@@ -24,12 +24,12 @@ document.addEventListener('DOMContentLoaded', () => {
     options.push({
       key: 'contributions',
       text: Translate.string('Detailed'),
-      description: Translate.string('A detailed timetable scoped to contributions.'),
+      description: Translate.string('A detailed timetable containing individual contributions.'),
       extraParams: {scope: 'contribution'},
     });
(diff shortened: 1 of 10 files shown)


indico/web/client/js/react/components/ICSCalendarLink.module.scss
```diff
@@ -13,10 +13,15 @@
 
 .export-header:global(.ui.header) {
   display: flex;
-  justify-content: center;
+  flex-direction: column;
+  align-items: center;
   margin-bottom: 0;
 }
 
+.export-header:global(.ui.header:first-child) {
```
ThiefMaster, line 21:
```suggestion
.export-header:global(.ui.header):first-child {
```

IMO it's more logical to only keep the stuff that actually needs global scoping in there. (untested btw, but I don't see how it would make a difference in the resulting CSS)

indico/web/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -82,90 +66,58 @@ export default function ICSCalendarLink({
...
+            <>
+              <Button
+                key={key}
+                content={text}
+                onClick={() => handleSetOption(key, extraParams)}
+                primary={key === popupState.key}
+                label={idx === 0 ? Translate.string('Export') : undefined}
+                labelPosition={idx === 0 ? 'left' : undefined}
+              />
+              {idx < options.length - 1 && <Button.Or />}
```
ThiefMaster, line 116: Maybe that `<Button.Or />` should have a `text={Translate.string('or')}` so we can translate it?

indico/web/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -82,90 +66,58 @@ export default function ICSCalendarLink({
...
     >
-      <Header
-        styleName="export-header"
-        content={Translate.string('Export')}
-        subheader={<Label color="blue">{popupState.text}</Label>}
-      />
+      <Header styleName="export-header">
+        <Button.Group size="small">
+          {options.map(({key, text, extraParams}, idx) => (
+            <>
```
ThiefMaster, line 107: React warns about a missing key here -> `<React.Fragment key={key}>`

indico/web/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -82,90 +66,58 @@ export default function ICSCalendarLink({
...
+              <Button
+                key={key}
+                content={text}
+                onClick={() => handleSetOption(key, extraParams)}
+                primary={key === popupState.key}
+                label={idx === 0 ? Translate.string('Export') : undefined}
```
ThiefMaster, line 113: can we render the label separately somehow? it's a bit weird that it acts as a label for the first actual button (and is clickable / has the cursor associated with clickable elements)

if it's not possible to render the label separately, maybe some css to disable pointer-events on it would help...

indico/web/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -9,46 +9,30 @@ import signURL from 'indico-url:core.sign_url';
...
-  params,
-  renderButton,
-  dropdownPosition,
-  popupPosition,
-  options,
-  ...restProps
-}) {
-  const [popupState, dispatch] = useReducer(popupReducer, initialState);
+export default function ICSCalendarLink({endpoint, params, renderButton, popupPosition, options}) {
+  const [popupState, dispatch] = useReducer(popupReducer, {open: false});
```
ThiefMaster, line 34: What's the reason for this change? Isn't it cleaner to keep any state fields we set in the reducer in the initial state as well?

indico/web/client/js/react/components/ICSCalendarLink.jsx
ThiefMaster: ok looks like disabling pointer events doesn't work because of the DOM structure used for those labels..

indico/web/client/js/react/components/ICSCalendarLink.jsx
plourenco: Yes, saw that then forgot about it

indico/web/client/js/react/components/ICSCalendarLink.jsx
plourenco: Yes, I noticed that too, seems like `cursor: initial` in the wrapper `ui left labeled button` works, the button preserves the click.

indico/web/client/js/react/components/ICSCalendarLink.jsx
plourenco: Sure, I can restore this if you feel like it. The reason is that it seems like boilerplate to me, all the fields were undefined (unused) except for `open` which is the only one we actually expect to use in the first state.

indico/web/client/js/react/components/ICSCalendarLink.jsx
plourenco: Good point

indico/web/client/js/react/components/ICSCalendarLink.jsx
ThiefMaster: at least i always saw "full" initial state definitions.. (and the advantage is that e.g. `something === null` would work, but with undefined keys that's not the case. also, there's always the full state visible in the react dev tools)

indico/modules/events/client/js/header.jsx
```diff
@@ -23,21 +23,27 @@ document.addEventListener('DOMContentLoaded', () => {
   if (parseInt(eventContribCount, 10) > 0) {
     options.push({
       key: 'contributions',
-      text: Translate.string('Detailed timetable'),
-      extraParams: {detail: 'contributions'},
+      text: Translate.string('Detailed'),
+      description: Translate.string('A detailed timetable scoped to contributions.'),
```
ThiefMaster, line 27:
```suggestion
      description: Translate.string('A detailed timetable containing individual contributions.'),
```

indico/modules/events/client/js/header.jsx
```diff
@@ -23,21 +23,27 @@ document.addEventListener('DOMContentLoaded', () => {
...
+      extraParams: {scope: 'contribution'},
+    });
+    options.push({
+      key: 'sessions',
+      description: Translate.string('A detailed timetable scoped to sessions.'),
```
ThiefMaster, line 32:
```suggestion
      description: Translate.string('A detailed timetable containing individual session blocks.'),
```

indico/modules/events/controllers/display.py
```diff
@@ -8,22 +8,26 @@
...
 from indico.modules.events.views import WPConferenceDisplay, WPSimpleEventDisplay
+from indico.web.args import use_kwargs
 from indico.web.flask.util import send_file, url_for
 from indico.web.rh import allow_signed_url
 
 
 @allow_signed_url
 class RHExportEventICAL(RHDisplayEventBase):
-    def _process(self):
-        detailed = request.args.get('detail') == 'contributions'
```
ThiefMaster, line 25: I think we should keep this as a fallback for people who already have existing signed links...

indico/modules/events/controllers/display.py
```diff
@@ -8,22 +8,26 @@
...
-        event_ical = event_to_ical(self.event, session.user, detailed)
+    @use_kwargs({
+        'scope': EnumField(CalendarScope)
```
ThiefMaster, line 27: not required + no default value = the kwarg doesn't get passed if it's missing, which resutls in an error. `missing=None` and then using the old `'detail'` logic is probably a good idea...

indico/modules/events/controllers/display.py
```diff
@@ -8,22 +8,26 @@
...
+    })
```
ThiefMaster, line 28:
```suggestion
    }, location='query')
```

otherwise it only checks form/json data

indico/modules/events/ical.py
```diff
@@ -95,38 +101,46 @@ def generate_event_component(event, user=None):
...
 
 
-def event_to_ical(event, user=None, detailed=False):
+def event_to_ical(event, user=None, scope=None):
     """Serialize an event into an ical.
 
     :param event: The event to serialize
     :param user: The user who needs to be able to access the events
-    :param detailed: If True, iCal will include the event's contributions
+    :param scope: If True, iCal will include a detailed timetable up to this event scope
```
ThiefMaster, line 109:
```suggestion
    :param scope: If specified, use a more detailed timetable using the given scope
```

indico/modules/events/ical.py
```diff
@@ -95,38 +101,46 @@ def generate_event_component(event, user=None):
...
 
 
-def events_to_ical(events, user=None, detailed=False):
+def events_to_ical(events, user=None, scope=None):
     """Serialize multiple events into an ical.
 
     :param events: A list of events to serialize
     :param user: The user who needs to be able to access the events
-    :param detailed: If True, iCal will include the event's contributions
+    :param scope: If True, iCal will include a detailed timetable up to this event scope
```
ThiefMaster, line 119:
```suggestion
    :param scope: If specified, use a more detailed timetable using the given scope
```

indico/modules/events/ical.py
```diff
@@ -95,38 +101,46 @@ def generate_event_component(event, user=None):
...
     """
     calendar = icalendar.Calendar()
     calendar.add('version', '2.0')
     calendar.add('prodid', '-//CERN//INDICO//EN')
 
     for event in events:
-        if not detailed:
+        if not scope or not CalendarScope.get('scope'):
```
ThiefMaster, line 126: Not sure what the second check is meant for (i guess the quotes weren't intended), but I don't think it's necessary - we always expect none or an enum member.

Maybe the whole nested if/else construct here could be unnested to something like this?

```python
if scope == CalendarScope.X:
    ....
elif scope == CalendarScope.Y:
    ....
else:
    ...
```

The duplicated code would be minimal (just the loop over components).

plourenco: Yes, this was stupid, the reason was that shortly before pushing I was using strings as scope, then I decided to update it to an Enum.

indico/modules/events/ical.py
```diff
@@ -95,39 +101,52 @@ def generate_event_component(event, user=None):
...
+        elif scope == CalendarScope.session:
+            components = [
+                generate_session_component(session)
+                for session in event.sessions
+                if session.start_dt
+            ]
+            components += [
+                generate_contribution_component(contrib)
+                for contrib in event.contributions
+                if contrib.start_dt and not contrib.session_id
```
ThiefMaster, line 144:
```suggestion
                if contrib.start_dt and not contrib.session_id is not None
```

just because technically there could be a session with id 0

ThiefMaster (approved):

Issue #4912, opened 2021-05-21
Search: Add a setting for admins to get all results


ThiefMaster:
Not sure on the UI - suggestions welcome!

For the implementation, this would be a new `allow_admin` arg to the `search()` method; in InternalSearch it would be passed on to the `can_access()` check; in the citadel plugin it would add `IndicoAdmin` to the access list sent to citadel

PR #4926, opened 2021-06-02
Skip access checks option for admin search

javfg:
Enables administrators to include protected entries in their search.

Provisionally includes two changes in the UI:

* A checkbox in the search page:
image

* A new option with a checkbox in the search box popup (not sure yet if we are keeping this one):
image

Closes https://github.com/indico/indico/issues/4912.

### DIFF
--- a/indico/modules/categories/templates/display/base.html
+++ b/indico/modules/categories/templates/display/base.html
@@ -25,7 +25,7 @@
                 <div class="group">
                     <div id="search-box" class="search-box"
                          data-category="{{ {'id': category.id, 'isRoot': category.is_root, 'title': category.title} | tojson | forceescape }}"
-                         data-is-admin="{{ session.user.is_admin | tojson }}"
+                         {% if session.user.is_admin %}data-is-admin{% endif %}
                     ></div>
                 </div>
                 <div class="group">
--- a/indico/modules/categories/templates/display/base.html
+++ b/indico/modules/categories/templates/display/base.html
@@ -25,7 +25,7 @@
                 <div class="group">
                     <div id="search-box" class="search-box"
                          data-category="{{ {'id': category.id, 'isRoot': category.is_root, 'title': category.title} | tojson | forceescape }}"
-                         {% if session.user.is_admin %}data-is-admin{% endif %}
+                         {% if session.user and session.user.is_admin %}data-is-admin{% endif %}
                     ></div>
                 </div>
                 <div class="group">
--- a/indico/modules/categories/templates/display/base.html
+++ b/indico/modules/categories/templates/display/base.html
@@ -25,6 +25,7 @@
                 <div class="group">
                     <div id="search-box" class="search-box"
                          data-category="{{ {'id': category.id, 'isRoot': category.is_root, 'title': category.title} | tojson | forceescape }}"
+                         data-is-admin="{{ session.user.is_admin | tojson }}"
                     ></div>
                 </div>
                 <div class="group">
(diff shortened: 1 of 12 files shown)


indico/modules/search/client/js/components/SearchBox.jsx
```diff
@@ -23,33 +23,66 @@ const renderResult = ({label}, keyword) => (
...
-          {value: 'global', title: 'Global', label: Translate.string('All Indico ↵')},
         ]
       : [];
+  if (isAdmin) {
+    options.push({
+      value: 'global',
+      title: 'allowAdmin',
+      className: 'cursor-default',
+      onClick: () => {},
+      // eslint-disable-next-line
```
ThiefMaster, line 52: only disable the problematic rule here, not everything

indico/modules/search/client/js/components/SearchBox.jsx
```diff
@@ -60,11 +93,12 @@ export default function SearchBox({onSearch, category}) {
         results={options}
         value={keyword}
         showNoResults={false}
+        open={!!keyword.length}
```
ThiefMaster, line 96:
```suggestion
        open={!!keyword}
```

this should be enough assuming that `keyword` is a string

indico/modules/search/internal.py
```diff
@@ -17,11 +17,11 @@
 
 
 class InternalSearch(IndicoSearchProvider):
-    def search(self, query, user=None, page=None, object_types=(), **params):
+    def search(self, query, user=None, page=None, object_types=(), allow_admin=False, **params):
```
ThiefMaster, line 20: I'd make `allow_admin` keyword-only (also in the base class and other subclasses)

indico/modules/search/templates/search.html
```diff
@@ -7,5 +7,5 @@
 {% endblock %}
 
 {% block content %}
-    <div id="search-root"></div>
+    <div id="search-root" {% if session.user.is_admin %}data-is-admin{% endif %}></div>
```
ThiefMaster, line 10: `session.user and session.user.is_admin`

indico/modules/categories/templates/display/base.html
```diff
@@ -25,6 +25,7 @@ <h1 class="category-title {{ category_title_classes }}">
                 <div class="group">
                     <div id="search-box" class="search-box"
                          data-category="{{ {'id': category.id, 'isRoot': category.is_root, 'title': category.title} | tojson | forceescape }}"
+                         {% if session.user.is_admin %}data-is-admin{% endif %}
```
ThiefMaster, line 28: `session.user and session.user.is_admin`

indico/modules/search/client/js/components/SearchApp.jsx
```diff
@@ -214,11 +214,12 @@ SearchOptions.defaultProps = {
   sortOptions: undefined,
 };
 
-export default function SearchApp({category, eventId}) {
+export default function SearchApp({category, eventId, isAdmin}) {
   const eventSearch = eventId !== null;
   const [query, setQuery] = useQueryParams();
   const [activeMenuItem, setActiveMenuItem] = useState(undefined);
-  const {q, sort, ...filters} = query;
+  const {q, sort, admin_override_enabled: adminOverrideEnabledStr = false, ...filters} = query;
```
ThiefMaster, line 221: shouldn't it be `'false'`? `JSON.parse(false) == false` but that seems to be only because it converts the arg to a string..

indico/modules/search/client/js/components/SearchBox.jsx
```diff
@@ -23,33 +23,66 @@ const renderResult = ({label}, keyword) => (
...
   const handleSubmit = () => {
     if (keyword.trim()) {
-      onSearch(keyword.trim(), category ? category.isRoot : true);
+      onSearch(keyword.trim(), category ? category.isRoot : true, adminOverrideEnabled);
     }
   };
 
   const handleResultSelect = (e, data) => {
-    onSearch(keyword.trim(), data.result.value === 'global');
+    if (data.result.title === 'enableAdminOverride') {
```
ThiefMaster, line 81: can't we use the value for this? title is usually meant for humans, and while using the custom renderer never displays it, it feels wrong to use it like this..

ThiefMaster (approved):

PR #4927, opened 2021-06-02
Fix ACL dropdown overflow in modals

plourenco:
Fixes dropdown cropping in modals with hidden overflow. I've tried many different approaches resorting to CSS only and while the menu is `absolute` positioned, its parent - the buttons - are relative, making it very hard to ignore the modal overflow. In the original implementation, semantic doesn't expect any surrounding overflow.

For that reason, I'm removing the forced `upward` positioning, including scroll, which should cause it to play with the direction nicely.

imagem

### DIFF
--- a/indico/web/client/js/react/components/principals/ACLField.jsx
+++ b/indico/web/client/js/react/components/principals/ACLField.jsx
@@ -167,7 +167,8 @@ const ACLField = props => {
             <Dropdown
               text={Translate.string('Event Role')}
               button
-              upward
+              defaultUpward
+              scrolling
               floating
               disabled={eventRoleOptions.length === 0}
               options={eventRoleOptions}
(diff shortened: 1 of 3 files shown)


indico/web/client/js/react/components/principals/ACLField.jsx
```diff
@@ -100,6 +101,20 @@ const ACLField = props => {
       text: r.name,
     }));
 
+  const onDropdownOpen = event => {
+    // Fixes semantic dropdown cropping within constrained elements
+    // `defer` ensures the scroll is queued after all state updates
+    const target = event.currentTarget;
+    if (scrollOnOpen) {
+      _.defer(() =>
+        target?.querySelector('.menu.visible')?.scrollIntoView({
```
ThiefMaster, line 110: I'd use `if (scrollOnOpen && target)` above and remove the `?` after `target` here

ThiefMaster (approved):

ThiefMaster (approved):

PR #4934, opened 2021-06-04
Vertically align affiliation with a person's name

tomasr8:
Vertically aligns the affiliation next a person's name. 

Before:
Screenshot from 2021-06-04 15-38-50
After:
Screenshot from 2021-06-04 15-38-23

The fix is based on this SO thread: https://stackoverflow.com/q/20566710

Basically, `inline-block` together with `overflow: hidden` moves the element up which is fixed by adding `vertical-align: bottom`.


ThiefMaster (approved):
Finally! This small display issue annoyed me since a long time... 🎉

PR #4937, opened 2021-06-07
Show placeholders in room-specific notification settings

### LINKED ISSUE #4840: RB: Show placeholder in room-specific notification settings
image

We have global values that can be used as placeholders (like the "3" in the first field which I added manually before taking the screenshot, and yes I'm totally too lazy to take another screenshot, which is why I'm writing this comment now :P).

Note: We probably need a new endpoint to get those defaults, since room owners can edit rooms as well and unless they are indico admins they cannot get them via `/api/admin/settings` (exposing it to them would show too much, since e.g. the ACLs of who can access RB are none of their business). Alternatively we can return a limited set of data to non-admins, but a separate endpoint returning just the notification defaults (e.g. `/api/admin/settings/notifications`) is probably cleaner and less confusing.

tomasr8:
Closes #4840 

Added notification placeholders in  the `Notifications` tab
of the `Edit Room Details` modal. The notification settings
were previously only available from the `api/admin/settings`
endpoint which is admin-protected.

This commit adds a new unprotected endpoint `api/config/notifications`
which exposes only the notification settings. This endpoint is
used to show the corresponding placeholder values in the input
elements.

Screenshot from 2021-06-07 17-41-44

### DIFF
--- a/indico/modules/rb/client/js/common/rooms/edit/RoomEditNotifications.jsx
+++ b/indico/modules/rb/client/js/common/rooms/edit/RoomEditNotifications.jsx
@@ -13,7 +13,7 @@ import {FinalEmailList} from 'indico/react/components';
 import {FieldCondition, FinalCheckbox, FinalInput, validators as v} from 'indico/react/forms';
 import {Translate} from 'indico/react/i18n';
 
-export default function RoomEditNotifications({active}) {
+export default function RoomEditNotifications({active, roomNotificationSettings}) {
   return (
     <Tab.Pane active={active}>
       <Header>
@@ -35,6 +35,7 @@ export default function RoomEditNotifications({active}) {
           <FinalInput
             fluid
             name="notification_before_days"
+            placeholder={roomNotificationSettings.notification_before_days}
             label={Translate.string('Single/Daily')}
             type="number"
             min="1"
@@ -44,6 +45,7 @@ export default function RoomEditNotifications({active}) {
           <FinalInput
             fluid
             name="notification_before_days_weekly"
+            placeholder={roomNotificationSettings.notification_before_days_weekly}
             label={Translate.string('Weekly')}
             type="number"
             min="1"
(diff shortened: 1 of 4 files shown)


indico/modules/rb/controllers/backend/misc.py
```diff
@@ -53,6 +53,16 @@ def _process(self):
...
 
 
+class RHNotificationSettings(RHRoomBookingBase):
+    def _process(self):
+        return jsonify(notification_before_days=str(rb_settings.get('notification_before_days')),
+                       notification_before_days_weekly=str(rb_settings.get('notification_before_days_weekly')),
+                       notification_before_days_monthly=str(rb_settings.get('notification_before_days_monthly')),
+                       end_notification_daily=str(rb_settings.get('end_notification_daily')),
+                       end_notification_weekly=str(rb_settings.get('end_notification_weekly')),
+                       end_notification_monthly=str(rb_settings.get('end_notification_monthly')))
```
ThiefMaster, line 63: We can use the schema here (like I mentioned on Mattermost), and I would not return this data as strings... If the placeholders need to be strings, you can just wrap them in `String(...)` in the JS

indico/modules/rb/client/js/common/rooms/edit/RoomEditNotifications.jsx
```diff
@@ -13,7 +13,7 @@ import {FinalEmailList} from 'indico/react/components';
 import {FieldCondition, FinalCheckbox, FinalInput, validators as v} from 'indico/react/forms';
 import {Translate} from 'indico/react/i18n';
 
-export default function RoomEditNotifications({active}) {
+export default function RoomEditNotifications({active, roomNotificationSettings}) {
```
ThiefMaster, line 16: I'd rename `roomNotificationSettings` to something like `defaults` or `globals` or `globalValues`

indico/modules/rb/controllers/backend/misc.py
tomasr8: Fixed ;)

ThiefMaster (approved):

PR #4939, opened 2021-06-09
Display avatars in user search

tomasr8:
When searching users, instead of the default icon, avatars are displayed.
If the user does not have an avatar, the default avatar with the first letter of
the first name is used.

The group search still displays the default result icon.

Before:
Screenshot from 2021-06-08 16-41-21
After:
Screenshot from 2021-06-08 18-07-55
Screenshot from 2021-06-09 10-47-43

### DIFF
--- a/indico/modules/core/controllers.py
+++ b/indico/modules/core/controllers.py
@@ -31,7 +31,6 @@ from indico.modules.core.settings import core_settings, social_settings
 from indico.modules.core.views import WPContact, WPSettings
 from indico.modules.legal import legal_settings
 from indico.modules.users.controllers import RHUserBase
-from indico.modules.users.util import get_avatar_from_name
 from indico.util.i18n import _, get_all_locales
 from indico.util.marshmallow import PrincipalDict, validate_with_message
 from indico.util.string import sanitize_html
@@ -200,8 +199,7 @@ class PrincipalsMixin:
                     'last_name': principal.last_name,
                     'email': principal.email,
                     'affiliation': principal.affiliation,
-                    'avatar_url': (principal.avatar_url
-                                   if principal.avatar_url else get_avatar_from_name(principal.first_name)),
+                    'avatar_url': principal.avatar_url,
                     'detail': (f'{principal.email} ({principal.affiliation})'
                                if principal.affiliation else principal.email)}
         elif principal.principal_type == PrincipalType.local_group:
--- a/indico/modules/core/controllers.py
+++ b/indico/modules/core/controllers.py
@@ -31,6 +31,7 @@ from indico.modules.core.settings import core_settings, social_settings
 from indico.modules.core.views import WPContact, WPSettings
 from indico.modules.legal import legal_settings
 from indico.modules.users.controllers import RHUserBase
+from indico.modules.users.util import get_avatar_from_name
 from indico.util.i18n import _, get_all_locales
 from indico.util.marshmallow import PrincipalDict, validate_with_message
 from indico.util.string import sanitize_html
@@ -199,6 +200,8 @@ class PrincipalsMixin:
                     'last_name': principal.last_name,
                     'email': principal.email,
                     'affiliation': principal.affiliation,
+                    'avatar_url': (principal.avatar_url
+                                   if principal.avatar_url else get_avatar_from_name(principal.first_name)),
                     'detail': (f'{principal.email} ({principal.affiliation})'
                                if principal.affiliation else principal.email)}
         elif principal.principal_type == PrincipalType.local_group:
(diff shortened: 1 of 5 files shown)


indico/modules/core/controllers.py
```diff
@@ -199,6 +200,8 @@ def _serialize_principal(self, identifier, principal):
                     'last_name': principal.last_name,
                     'email': principal.email,
                     'affiliation': principal.affiliation,
+                    'avatar_url': (principal.avatar_url
+                                   if principal.avatar_url else get_avatar_from_name(principal.first_name)),
```
ThiefMaster, line 204: I don't think the if/else is needed here; users always have a suitable `avatar_url`

indico/modules/users/util.py
```diff
@@ -441,3 +441,8 @@ def send_avatar(user):
     return send_file('avatar.png', BytesIO(user.picture), mimetype=metadata['content_type'],
                      inline=True, conditional=True, last_modified=parse_date(metadata['lastmod']),
                      cache_timeout=(86400*7))
+
+
+def get_avatar_from_name(first_name):
```
ThiefMaster, line 446:
```suggestion
def get_avatar_url_from_name(first_name):
```

that name is clearer IMO (don't forget to chance the places where it's called/imported as well ofc)

indico/modules/users/controllers.py
```diff
@@ -696,6 +696,8 @@ def _serialize_pending_user(self, entry):
         affiliation = entry.data.get('affiliation') or ''
         email = entry.data['email'].lower()
         ext_id = f'{entry.provider.name}:{entry.identifier}'
+        avatar_url = entry.data.get('avatar_url') or get_avatar_from_name(first_name)
```
ThiefMaster, line 699: AFAIL `entry` here is an `IdentityInfo` object from flask-multipass which never has an avatar URL, so you can always call the util here

indico/web/client/js/react/components/principals/Search.jsx
```diff
@@ -58,38 +59,48 @@ const searchFactory = config => {
...
-            </List.Description>
-          )}
+  const FavoriteItem = ({name, detail, added, onAdd, onRemove, avatarURL}) => {
+    const avatar = avatarURL ? (
+      <Image src={avatarURL} size="mini" avatar />
+    ) : (
+      <Icon name={resultIcon} />
+    );
+
+    const marginLeft = avatarURL ? 'margin-left' : '';
```
ThiefMaster, line 69: I don't think that's a good name for a CSS class. I'd try to find a name that describes the meaning, e.g. `has-avatar`.

indico/web/client/js/react/components/principals/Search.jsx
```diff
@@ -141,6 +152,7 @@ const searchFactory = config => {
                         onRemove={() => {
                           onRemove(x);
                         }}
+                        avatarURL={x.avatarURL}
```
ThiefMaster, line 155: I'd move it above the callbacks ;) looks cleaner IMO if the single-line props come first

tomasr8: Fixed ;)

indico/web/client/js/react/components/principals/Search.jsx
tomasr8: I used `has-avatar`

indico/modules/users/controllers.py
tomasr8: Removed ;)

indico/modules/users/util.py
tomasr8: Renamed the function

indico/modules/core/controllers.py
tomasr8: Removed the redundant check ;)

indico/web/client/js/react/components/principals/Search.jsx
```diff
@@ -58,38 +59,48 @@ const searchFactory = config => {
...
-            </List.Description>
-          )}
+  const FavoriteItem = ({name, detail, added, onAdd, onRemove, avatarURL}) => {
+    const avatar = avatarURL ? (
+      <Image src={avatarURL} size="mini" avatar />
+    ) : (
+      <Icon name={resultIcon} />
+    );
+
+    const hasAvatar = avatarURL ? 'has-avatar' : '';
```
ThiefMaster, line 69: `hasAvatarClass`? the current name sounds very bool-ish

ThiefMaster (approved):
Tested it locally and it works great! :) Just one more small naming suggestion...

indico/web/client/js/react/components/principals/Search.jsx
tomasr8: You're right, renamed ;)

PR #4940, opened 2021-06-09
Properly delete broken symlinks in `cleanup_dir`

tomasr8:
Calling `os.path.get_mtime` on a broken symlink
raises OSError because `get_mtime` calls `os.stat`
internally. Broken symlinks are now always deleted.

### DIFF
--- a/indico/util/fs.py
+++ b/indico/util/fs.py
@@ -144,15 +144,39 @@ def cleanup_dir(path, min_age, dry_run=False, exclude=None):
         has_files = False
         for filename in files:
             filepath = os.path.join(root, filename)
-            if os.path.getmtime(filepath) >= min_mtime:
+
+            try:
+                # raises an exception for invalid symlinks
+                mtime = os.path.getmtime(filepath)
+            except OSError:
+                if is_broken_symlink(filepath):
+                    if dry_run or silentremove(filepath):
+                        deleted.add(os.path.relpath(filepath, path))
+                    else:
+                        has_files = True
+                else:
+                    has_files = True
+
+                continue
+
+            if mtime >= min_mtime:
                 has_files = True
             elif dry_run or silentremove(filepath):
                 deleted.add(os.path.relpath(filepath, path))
             else:
                 has_files = True  # deletion failed
+
         if not dry_run and not has_files and not dirs and relroot:
             removedirs(path, relroot)
     return deleted
+
+
+def is_broken_symlink(path):
+    """Checks whether a file is a broken symlink
+
+    :param path: The path to the file
+    """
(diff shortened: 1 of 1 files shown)


indico/util/fs.py
```diff
@@ -144,17 +144,41 @@ def cleanup_dir(path, min_age, dry_run=False, exclude=None):
...
             else:
                 has_files = True  # deletion failed
+
         if not dry_run and not has_files and not dirs and relroot:
             removedirs(path, relroot)
     return deleted
 
 
+def is_broken_symlink(path):
+    """Checks whether a file is a broken symlink
```
ThiefMaster, line 175:
```suggestion
    """Check whether a file is a broken symlink.
```

https://www.python.org/dev/peps/pep-0257/

indico/util/fs.py
```diff
@@ -144,17 +144,41 @@ def cleanup_dir(path, min_age, dry_run=False, exclude=None):
...
+                # raises an exception for invalid symlinks
+                mtime = os.path.getmtime(filepath)
+            except OSError:
+                if is_broken_symlink(filepath):
+                    if dry_run or silentremove(filepath):
+                        deleted.add(os.path.relpath(filepath, path))
+                    else:
+                        has_files = True
+                else:
+                    has_files = True
```
ThiefMaster, line 158:
```python
from indico.core.logger import Logger
Logger.get('fs').exception('Could not stat %s', filepath)
```

indico/util/fs.py
```diff
@@ -144,17 +144,41 @@ def cleanup_dir(path, min_age, dry_run=False, exclude=None):
         has_files = False
         for filename in files:
             filepath = os.path.join(root, filename)
-            if os.path.getmtime(filepath) >= min_mtime:
+
+            try:
+                # raises an exception for invalid symlinks
+                mtime = os.path.getmtime(filepath)
+            except OSError:
```
ThiefMaster, line 151: Is it always a `FileNotFoundError` in case of broken symlinks? If yes, we could catch that one separately and in case of other OSErrors just keep the logging and `has_files = True`

tomasr8: If it's a symlink, then I think it's safe to assume that `ENOENT` means a broken symlink. But I think `ENOENT` can be raised by files which are not symlinks, so I don't think that the implication `ENOENT` => broken symlink is valid. Conversely, a broken symlink can also show up as other errors - `ELOOP` (which is detected by `is_broken_symlink`) and `ENOTDIR` which right now is ignored.

tomasr8: Based on the MM conversation, I keep it as it is now ;)

Issue #4809, opened 2021-03-05
Ellipsize long menu titles


ThiefMaster:
In some languages certain menu items are too long to fit in a single line in the side menus. It would be nice if we ellipsized those entries, and also showed a tooltip with the full text on hover (but only if it didn't fit).

![image](https://user-images.githubusercontent.com/179599/110101514-e9c3da00-7da3-11eb-9ac8-3f5766789355.png)

Maybe we could use the same opportunity to ellipsize this particular button's text in event management instead of truncating it?

![image](https://user-images.githubusercontent.com/179599/110101602-0fe97a00-7da4-11eb-97d5-fa148d2eeae1.png)

PR #4943, opened 2021-06-11
Ellipsize some side bar elements

javfg:
Ellipsizes the 'switch to display view' button and sidebar menu entries.

Closes https://github.com/indico/indico/issues/4809.


ThiefMaster (approved):
thanks, looks much better now in languages with longer strings :)

PR #4945, opened 2021-06-11
Fix `ICSCalendarLink` button overflow

tomasr8:
The SUI `Popup` element has a `max-width` of 350px which is increased to 550px while the
`max-width` of the text of the popup is set to 350px. This allows the popup to resize to fit the width of the buttons.

Screenshot from 2021-06-11 15-25-25
Screenshot from 2021-06-11 15-26-06

Edit:
Now the inputs also fill the whole width of the popup. Thanks to @javfg 
These modifications are based on https://stackoverflow.com/q/55040250.

image


indico/web/client/js/react/components/ICSCalendarLink.module.scss
```diff
@@ -48,11 +48,21 @@
...
   display: flex;
   align-items: center;
 }
+
+.popup:global(.ui.wide.popup) {
+  display: inline-flex;
+  max-width: 450px;
+}
+
+.popup-content {
```
plourenco, line 65: We could have this one as just `.popup .content` and reference it as `styleName="content"` (since parent is already `popup`) as it seems clearer.
```suggestion
.popup .content {
```

indico/web/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -56,6 +57,7 @@ const ICSExportOptions = ({options, selected, handleSetOption}) => (
               primary={key === selected}
               label={idx === 0 ? Translate.string('Export') : undefined}
               labelPosition={idx === 0 ? 'left' : undefined}
+              styleName={idx === options.length - 1 ? 'final-button' : 'mid-button'}
```
plourenco, line 60: Instead of conditionally changing the styleName, since it's only the first labeled button having an incorrect right radius, I would suggest setting `styleName="button-group"` to the parent `Button.Group` and then:

```
.button-group :global(.ui.labeled.button) button:global(.ui.button):last-child {
  border-top-right-radius: 0;
  border-bottom-right-radius: 0;
}
```

I think we could also keep this CSS in `ICSCalendarLink` and get rid of `ICSExportOptions` since they are both in the same file.

plourenco (approved):
Looks good, nice catch 👍

indico/web/client/js/react/components/ICSCalendarLink.module.scss
tomasr8: Fixed :)

indico/web/client/js/react/components/ICSCalendarLink.jsx
tomasr8: Thanks, fixed. Also deleted the file :+1:

ThiefMaster:
I think we need a min-width on the dialog. It looks quite bad now in events that do not have session blocks:

image

tomasr8:
> I think we need a min-width on the dialog. It looks quite bad now in events that do not have session blocks:
> 

You're right, I added it. I think it looks better now:

Screenshot from 2021-06-21 08-53-44

Screenshot from 2021-06-21 08-54-03

PR #4946, opened 2021-06-14
Add option to blacklist emails when registering users

### LINKED ISSUE #4644: Allow blacklisting email addresses for local account registration
There are cases of spammers signing up with very similar emails (all on the same domain). These signups are done by humans so email verification does not stop them (and neither would a CAPTCHA, if we had one).

Provide an option in the admin area to blacklist emails. This should ideally use glob-style wildcards (at least `*`) so you can blacklist e.g. `*@spammerdomain.com` but also individual email addresses.

Moderating registrations is an alternative, but puts more burden on the admins, especially if the majority of signups are legitimate.

**Additional context**
The email blacklist needs to be checked during local registration, but we could safely ignore it for SSO signups since this data is generally trusted. But maybe it's easier to enforce everywhere, ie both local and sso registration and also when managing account emails.

tomasr8:
Closes #4644 

Adds a new field to the admin user settings cog menu where one can specify a
comma-separated list of blacklisted email addresses.

It is possible to use specific addresses or wildcard patterns using the symbol *, e.g., `*@gmail.com`,
which would blacklist all `@gmail.com` emails.

image

Screenshot from 2021-06-14 14-44-55

### DIFF
--- a/indico/modules/auth/forms.py
+++ b/indico/modules/auth/forms.py
@@ -5,12 +5,14 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
+from fnmatch import fnmatch
+
 from wtforms.fields import PasswordField, SelectField, StringField, TextAreaField
 from wtforms.fields.html5 import EmailField
 from wtforms.validators import DataRequired, Email, Optional, ValidationError
 
 from indico.modules.auth import Identity
-from indico.modules.users import User
+from indico.modules.users import User, user_management_settings
 from indico.util.i18n import _
 from indico.web.forms.base import IndicoForm, SyncedInputsMixin
 from indico.web.forms.validators import ConfirmPassword, SecurePassword, used_if_not_synced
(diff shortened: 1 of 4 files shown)


indico/modules/auth/forms.py
```diff
@@ -31,6 +31,21 @@ def _check_existing_username(form, field):
...
 
 
+def _check_not_blacklisted(form, field):
+    blacklist = user_management_settings.get('email_blacklist').split(",")
+    email = field.data
+    error_msg = _('This email address is blacklisted.')
+
+    for pattern in blacklist:
+        pattern = pattern.strip()
+        if pattern.startswith('*'):
```
ThiefMaster, line 41: There's [`fnmatch`](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch) which does glob-like matching

indico/modules/users/forms.py
```diff
@@ -109,6 +109,11 @@ class AdminUserSettingsForm(IndicoForm):
     notify_account_creation = BooleanField(_('Registration notifications'), widget=SwitchWidget(),
                                            description=_('Send an email to all administrators whenever someone '
                                                          'registers a new local account.'))
+    email_blacklist = MultiStringField(_('Email blacklist'), field=('email_blacklist', _('email')),
+                                       unique=True, flat=True,
+                                       description=_('Disallow users from registering with these email addresses. '
+                                                     'Comma-separated list of addresses. '
```
ThiefMaster, line 115: this is no longer the case

ThiefMaster (approved):

PR #4950, opened 2021-06-16
Auto-expand notes from meeting note permalinks

### LINKED ISSUE #4947: Search: Link note results to fragments
We use the `event_notes.view` endpoint right now which just prints the note without any UI around it. This doesn't look amazing.

For meetings we could redirect to a new URL fragment (similar to what we use for session/contribution/subcontribution permalinks, just with a `-note` appended) which should then automatically expand that note (probably w/ JavaScript).

For conferences I would keep the current behavior as they do not have notes (unless someone created them before changing the type to conference).

Note that the `event_notes.view` endpoint itself needs to be left intact since this is also used for note permalinks. We could include some new query string arg like `from_search=1` which triggers the redirect, or just add a new endpoint that only redirects.

tomasr8:
Resolves #4947

Currently, the meeting notes permalinks lead to a plain HTML page with just the notes.

This commit changes the permalink to point to the contribution/block title fragment on the
event page (same link when clicking the direct link next to the title).

In addition, it also adds `?note=note_id` to the query string of the permalink.
This query field is used to automatically expand the relevant note when the event page is loaded.

This includes both the individual notes and the `main` note on the event itself.

To test this, go to a meeting which has minutes attached, expand the minutes and click on the permalink.
The link should lead to the event page with the block title highlighted and the minutes expanded.

Furthermore, the search results for notes now use the this new URL instead.

### DIFF
--- a/indico/modules/events/notes/templates/_note.html
+++ b/indico/modules/events/notes/templates/_note.html
@@ -1,13 +1,13 @@
-{% macro render_note(note, hidden=true, can_edit=false, for_event=none) %}
+{% macro render_note(note, hidden=true, can_edit=false, for_event=none, anchor=none) %}
     {% if note %}
         {% if for_event %}
             <div class="event-note-section">
-                {{ _render_note_box(note=note, event=for_event, hidden=hidden, can_edit=can_edit) }}
+                {{ _render_note_box(note=note, event=for_event, hidden=hidden, can_edit=can_edit, anchor=anchor) }}
             </div>
         {% else %}
             <div class="note-area-wrapper">
                 <div class="note-area {{ _togglable_classes(hidden) }}">
-                    {{ _render_note_box(note=note, can_edit=can_edit) }}
+                    {{ _render_note_box(note=note, can_edit=can_edit, anchor=anchor) }}
                 </div>
             </div>
         {% endif %}
@@ -15,7 +15,7 @@
 {% endmacro %}
 
 
-{% macro render_toggle_button(note, note_is_hidden) %}
+{% macro render_toggle_button(note, note_is_hidden, anchor=none) %}
     <div class="group i-selection note-visibility-toggle">
         {% set hash = uuid() %}
         <input type="checkbox" class="js-toggle-note-cb" id="toggle-note-{{ note.id }}-{{ hash }}" {% if not note_is_hidden %}checked{% endif %}>
@@ -25,11 +25,11 @@
 {% endmacro %}
 
 
-{% macro _render_note_box(note, event=none, hidden=false, can_edit=false) %}
+{% macro _render_note_box(note, event=none, hidden=false, can_edit=false, anchor=none) %}
     <div class="padded-box">
         <div class="padded-box-pad icon-file-text"></div>
         <div class="padded-box-content">
-            {{ _render_note_actions(note=note, event=event, hidden=hidden, can_edit=can_edit) }}
+            {{ _render_note_actions(note=note, event=event, hidden=hidden, can_edit=can_edit, anchor=anchor) }}
             <div class="note-text {% if event %}{{ _togglable_classes(hidden=hidden) }}{% endif %}">
                 {{ note.html|sanitize_html }}
             </div>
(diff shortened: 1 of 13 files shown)


indico/modules/events/templates/display/indico/meeting.html
```diff
@@ -61,8 +61,17 @@ <h1 itemprop="name">
         {% endif %}
 
         {% if event.has_note %}
-            {{ render_note(event.note, hidden=not theme_settings.show_notes, can_edit=event.can_manage(session.user),
-                           for_event=event) }}
+            {# Check if url contains ?note=note_id and if so, expand the note #}
+            {% set req = request.args.get('note')|string %}
```
ThiefMaster, line 65: it's always a string (or none if not present), so no need to convert

tomasr8: fixed ;)

indico/modules/search/result_schemas.py
```diff
@@ -199,6 +199,29 @@ class EventNoteResultSchema(ResultSchemaBase):
...
+
+        if not ev:
+            return self._get_fallback_url(data)
+
+        # notes attached to the meeting itself don't get fragments
+        if ev.note and ev.note.id == data['note_id']:
+            return url_for('events.display', event_id=data['event_id']) + f"?note={data['note_id']}"
+        else:
+            # find the parent block
+            blocks = SessionBlock.query.filter(SessionBlock.session.has(event=ev, is_deleted=False))
```
tomasr8, line 215: Can I do something like this?
`SessionBlock.query.filter(SessionBlock.session.note.has(id=3))`
instead of looping over all blocks?

ThiefMaster: something like this should work: `SessionBlock.query.filter(SessionBlock.session.has(Session.note.has(id=3)))`

tomasr8: Thanks :)

indico/modules/events/notes/templates/_note.html
```diff
@@ -54,13 +54,21 @@
...
     <div class="note-actions {% if event %}{{ _togglable_classes(hidden) }}{% endif %}">
         {% if event %}
             <a href="#" class="js-show-note-toggle">{% trans %}Hide{% endtrans %}</a>
         {% endif %}
-        <a href="{{ url_for('event_notes.view', note) }}" target="_blank"
-           class="permalink icon-link" title="{% trans %}Permanent link{% endtrans %}"></a>
+        {% if anchor %}
+            <a href="?note={{ note.id }}#{{ anchor }}" target="_blank"
+            class="permalink icon-link" title="{% trans %}Permanent link{% endtrans %}"></a>
+        {% elif anchor == '' %}
```
ThiefMaster, line 65: Also, do we really need to have separate cases for empty and none? When are we using which one right now?

tomasr8: The empty string is used to distinguish between the event note and the notes belonging to session blocks. The main event note does not have any anchor element to use, so we just omit it here. Normal notes are anchored to the block titles which have IDs. It doesn't feel very nice but I'm not sure if there is a cleaner alternative..

indico/modules/events/templates/display/indico/meeting.html
```diff
@@ -61,8 +61,17 @@ <h1 itemprop="name">
...
+            {% set req = request.args.get('note') %}
+            {% set note_id = event.note.id|string %}
+
+            {% if req == note_id %}
+                {{ render_note(event.note, hidden=false, can_edit=event.can_manage(session.user),
+                               for_event=event) }}
+            {% else %}
+                {{ render_note(event.note, hidden=not theme_settings.show_notes, can_edit=event.can_manage(session.user),
+                               for_event=event) }}
+            {% endif %}
```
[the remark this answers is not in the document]
tomasr8: The only difference from `render_notes()` is that it uses `item.can_edit_note(session.user)` instead of `event.can_manage(session.user)`. Can we unify that somehow? Perhaps it is possible to call `event.can_edit_note(session.user)`?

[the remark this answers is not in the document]
ThiefMaster: yes, it is possible :)

indico/modules/search/result_schemas.py
```diff
@@ -199,6 +199,29 @@ class EventNoteResultSchema(ResultSchemaBase):
     event_path = fields.Method('_get_event_path', dump_only=True)
 
     def _get_url(self, data):
+        note_id = data['note_id']
+        event_id = data['event_id']
+        note = EventNote.get(note_id)
+        event = Event.get(data['event_id'])
+
+        if not event:
```
ThiefMaster, line 207: `or event.type_ == EventType.conference` - the anchors don't work well for those events.

tomasr8: Thanks, fixed :)

ThiefMaster (approved):

PR #4954, opened 2021-06-18
Document Indico search functionality

### LINKED ISSUE #4922: Document Indico search functionality
- Explain what the out-of-the-box search can do (#4911)
- Explain what Citadel can do, and link to the plugin's README (on GitHub / PyPI - like we have for the Zoom plugin)
- Ideally get (the citadel developers to write) documentation on how to set up citadel in a "small" production setup (single machine, docker-composed, "fire and forget" if possible) so it can be used by Indico

plourenco:
Closes #4922


docs/source/search/_intro.rst
```diff
@@ -0,0 +1,14 @@
+Indico’s version 3.0 introduced a brand new reusable and agnostic search module
```
ThiefMaster, line 1: I think we use `'` ASCII quotes as apostrophes everywhere else

ThiefMaster: I don't think standalone "agnostic"  works; something is usually "something-agnostic", e.g. "backend-agnostic" or "database-agnostic"...

docs/source/search/_intro.rst
```diff
@@ -0,0 +1,14 @@
...
+backed up by your SQL storage by default. This module can however be decomposed
+into multiple providers, supporting any external service through a plugin.
```
ThiefMaster, line 3: you can't have more than one at the same time; not sure if "multiple providers" isn't misleading

docs/source/search/features.rst
```diff
@@ -0,0 +1,74 @@
...
+While the performance and range of features might not be the same, depending on your instance, if your volume of
+records is small and retrieval needs are simple, this module should be enough.
+
+It supports the two types of records from a total of six targets where each query is made upon the respective title:
+
+- Events
+- Categories
+
+.. note::
+    The Internal Search does not yet support: contributions, sub-contributions, event notes and attachments.
```
ThiefMaster, line 14: with indico/indico#4952 being ready, i think you could already update this (but mention that it only searches in titles/descriptions (and note content))

indico/modules/search/result_schemas.py
```diff
@@ -209,14 +216,23 @@ def _get_event_path(self, data):
...
+    #: The number of elements.
+    count: int = fields.Int(required=True)
+    #: The key that identifies the element's filter.
+    filter: str = fields.String(required=True)
 
 
 class AggregationSchema(_ResultSchemaBase):
-    label = fields.String(required=True)
-    buckets = fields.List(fields.Nested(BucketSchema), required=True)
+    """Represents an aggregation list"""
```
ThiefMaster, line 230: The title line of a docstring should end with a `.` (also in some other places like `BucketSchema`)

indico/modules/search/result_schemas.py
```diff
@@ -209,14 +216,23 @@ def _get_event_path(self, data):
...
+
+    #: The name of the aggregation.
+    label: str = fields.String(required=True)
+    #: A bucket list representing each group.
+    buckets: [BucketSchema] = fields.List(fields.Nested(BucketSchema), required=True)
```
ThiefMaster, line 235: i think this should be `t.List[BucketSchema]` (and `import typing as t`)

indico/modules/search/result_schemas.py
```diff
@@ -68,22 +69,23 @@ def _get_url(self, data):
 
 
 class LocationResultSchema(mm.Schema):
-    venue_name = fields.String(required=True)
-    room_name = fields.String(required=True)
-    address = fields.String(required=True)
+    venue_name = fields.String(required=True)  #: The venue name
+    room_name = fields.String(required=True)  #: The room name
+    address = fields.String(required=True)  #: The address
```
ThiefMaster, line 74: Isn't this the kind of docstring that's pretty redundant? Maybe better to just document non-obvious fields?

And I'd put them in the line before everywhere (we do the same in models where we use this type of doc string comment)

docs/source/search/plugin.rst
```diff
@@ -0,0 +1,48 @@
+API Reference
+*************
+
+The ``IndicoSearchProvider`` interface allows an abstract service to integrate with Indico's search module.
```
ThiefMaster, line 4: you can probably use something like this to link to the actual doc of that class:

    :class:`IndicoSearchProvider`

docs/source/search/index.rst
```diff
@@ -0,0 +1,14 @@
+Search
+======
+
+.. include:: _intro.rst
+
+.. raw:: html
+
+   <hr>
```
ThiefMaster, line 8: there's no less ugly way to do this?

docs/source/search/features.rst
```diff
@@ -0,0 +1,74 @@
...
+    Remove unused get_filters method
+
+.. autoclass:: indico.modules.search.base.IndicoSearchProvider
+    :members: get_filters
+
+Placeholders
+------------
+
+Placeholders are a special type of filters specifically designed to be part of the user-facing text based search query.
+Examples of valid placeholders would be: `affiliation:CERN` or `person:John Doe`.
```
ThiefMaster, line 64: I don't think the second example would work without quotes (i'd expect it to search for `person:John` and `Doe`

docs/source/search/features.rst
```diff
@@ -0,0 +1,74 @@
...
+------------
+
+Aggregations, `as seen in Elastic Search <https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html>`_,
+provide a way to combine information in groups according to a certain metric, such as a field value, sum or average.
+
+.. image:: ../images/search_features_aggregation.png
+
+Indico supports any bucket or metric group, composed of a key, count and filter key:
+
+.. autoclass:: indico.modules.search.result_schemas.AggregationSchema()
```
ThiefMaster, line 32: I don't think the `()` are needed here

indico/modules/search/result_schemas.py
plourenco: Ideally yes. But I just can't get class properties to show up undocumented.

docs/source/search/features.rst
plourenco: Without the `()` sphinx will assume I want to show the Model class contructor.

ThiefMaster: ah, good to know :)

indico/modules/search/result_schemas.py
ThiefMaster: `:undoc-members:` in the rst doesn't help?

docs/source/search/features.rst
plourenco: My mistake, thought about quotes and forgot to write them

indico/modules/search/result_schemas.py
plourenco: Yes, for members such as methods. Not really with class properties, unless a comment is there 😀

ThiefMaster: sigh... maybe worth reporting as a bug in sphinx?

docs/source/search/_intro.rst
```diff
@@ -0,0 +1,12 @@
+Indico's version 3.0 introduced a brand new reusable and backend-agnostic search module backed up by the
+SQL storage by default. This module can however be decomposed into a single provider,
+supporting any external service through a plugin.
+
+Indico provides multiple options for a search service, such as:
+
+- The default SQL based search.
+- A performant and feature-rich elastic search based search service, `Citadel`_ which can be integrated with
```
ThiefMaster, line 8: `Elasticsearch` is the brand name, so I'd capitalize it

```suggestion
- A performant and feature-rich Elasticsearch-based search service, `Citadel`_ which can be integrated with
```

docs/source/search/_intro.rst
```diff
@@ -0,0 +1,12 @@
...
+  Indico out-of-the-box thanks to its corresponding plugin developed in-house.
```
ThiefMaster, line 9: This feels a bit buzzwordy.. 🙈  How about this instead?

> Indico easily using the official [Indico Citadel plugin].

with [Indico Citadel plugin] linking to https://pypi.org/project/indico-plugin-citadel/

docs/source/search/features.rst
```diff
@@ -0,0 +1,77 @@
+Internal Search
+===============
+
+The Internal Search is a default SQL based engine implementation, created to support the most basic queries.
+While the performance and range of features might not be the same, depending on your instance, if your volume of
```
ThiefMaster, line 5: Too many commas ;) 

How about something like this?

> While not as fast as specialized search engines and somewhat basic (no filters or aggregations), we believe this search engine provides a decent option for smaller Indico instances which may not want to spend additional time on deploying a separate service just for search.

docs/source/search/features.rst
```diff
@@ -0,0 +1,77 @@
...
+- Attachments
+- Notes
+
+.. note::
+    The Internal Search only supports text-based search on titles, description and notes content.
+
+External Search Service
+=======================
+
+Indico supports several powerful features for aggregation, filtering and retrieval of information natively
```
ThiefMaster, line 22: > Indico provides several powerful features for aggregation and filtering when combined with an external search service supporting them. All of them are available when using the Citadel service for search.

or something like that

docs/source/search/features.rst
```diff
@@ -0,0 +1,77 @@
...
+covered by the out-of-the-box `Citadel`_ service.
+
+Aggregations
+------------
+
+Aggregations, `as seen in Elastic Search <https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html>`_,
```
ThiefMaster, line 28: `Elasticsearch`; and maybe move the link to the end of the file like you did with Citadel?

docs/source/search/plugin.rst
```diff
@@ -0,0 +1,48 @@
+API Reference
+*************
+
+The :class:`IndicoSearchProvider` interface allows an abstract service to integrate with Indico's search module.
```
ThiefMaster, line 4: s/abstract/external/? The service implementation isn't abstract anymore

docs/source/search/features.rst
plourenco: Sure, it's the only usage anyway, I had it in-line to avoid having this word associated with a specific aggregation link.

docs/source/search/plugin.rst
plourenco: Sounds good, with abstract I meant any type of external service.

docs/source/search/features.rst
ThiefMaster: Ah good point, it's really annoying that you cannot use different names for the link than the displayed text. MAybe better to leave it as-is...

docs/source/search/index.rst
plourenco: There is!

indico/modules/search/result_schemas.py
```diff
@@ -68,22 +71,23 @@ def _get_url(self, data):
...
-    description = fields.String(required=True)
-    event_type = EnumField(EventType, required=True)
-    start_dt = fields.DateTime(required=True)
-    end_dt = fields.DateTime(required=True)
-    persons = fields.List(fields.Nested(PersonSchema), required=True)
-    location = fields.Nested(LocationResultSchema, required=True)
-    highlight = fields.Nested(HighlightSchema, missing={})
+    #: The record type
+    type: SearchTarget = EnumField(SearchTarget, validate=require_search_target(SearchTarget.event))
+    event_id = fields.Int(required=True)  #: The event id
```
ThiefMaster, line 82: I'd still put all of them before the line  - otherwise it's a bit inconsistent, and every doc comment starting at a different horizontal position looks rather strange IMO

ThiefMaster (approved):

Issue #4881, opened 2021-04-19
Error exporting PDF timetable with a long abstract


bcc:
**Describe the bug**
Exporting a PDF timetable fails if the user selects the advanced options and ticks 'Print abstract content of all contributions' when an abstract contains too much text.

The user sees the following error message, which is not especially friendly:
`LayoutError: Flowable <Table@0x7FCA0B097EA8 1 rows x 3 cols(tallest row 4042)> with cell(0,0) containing ' '(526.582677165 x 4042), tallest cell 4042.0 points, too large on page 2 in frame 'normal'(526.582677165 x 773.196850394*) of template 'Later'`

**To Reproduce**
1. Create a new event with a single session.
2. Create a contribution abstract with a large quantity of text. I generated ~50 paragraphs of lorem ipsum text - this is overkill, but it demonstrates the issue nicely. 
3. Schedule that contribution for a session.
4. Go to the standard timetable page for the event.
5. Click 'PDF', and set the  'Advanced timetable' slider to 'yes'
6. Tick 'Print abstract content of all contributions', and 'Generate'

**Expected behavior**
Ideally a PDF without error, but a more friendly error message and not emailing an unexpected exception would be better than nothing. I did find https://groups.google.com/g/reportlab-users/c/wlIN3Fsg2VA which suggests it's not likely to be fixed in reportlab any time soon. 

**Additional context**
Full exception from my test instance:
```
2021-04-19 11:20:37,526  INFO     6974923bae04427b  indico.rh                 GET /event/1/timetable/ [IP=10.0.3.1] [PID=13430] [UID=1]
2021-04-19 11:20:39,782  INFO     9ca39d81354b4da6  indico.rh                 GET /event/1/timetable/pdf?_=1618831238112 [IP=10.0.3.1] [PID=13434] [UID=1]
2021-04-19 11:20:52,011  INFO     45921e3056bf4cda  indico.rh                 GET /event/1/timetable/pdf?submitted=&advanced=y&document_settings=showCoverPage&document_settings=showTableContents&document_settings=showSessionTOC&contribution_info=showContribId&contribution_info=showAbstract&session_info=useSe
ssionColorCodes&session_info=showSessionDescription&visible_entries=showContribsAtConfLevel&visible_entries=showBreaksAtConfLevel&pagesize=A4&firstPageNumber=1 [IP=10.0.3.1] [PID=13434] [UID=1]
[...]

PR #4955, opened 2021-06-18
Fix PDF timetable breaking when 'print abstracts' option is set

javfg:
Changes the layout of the timetable PDF when the advanced option 'Print abstract content of all contributions' is selected.

ReportLab is unable to split table rows when the content is bigger than the page it's placed upon, as described by [the discussion](https://groups.google.com/g/reportlab-users/c/wlIN3Fsg2VA?pli=1) linked in the bug report by @bcc.

To overcome this limitation, the timetable will not be contained in a table if the option to print the abstracts is set. The document is slightly different, but still provides the same information. If that option is unset, the previous layout persists.

Closes https://github.com/indico/indico/issues/4881.

### DIFF
--- a/indico/legacy/pdfinterface/conference.py
+++ b/indico/legacy/pdfinterface/conference.py
@@ -338,6 +338,19 @@ class TimeTablePlain(PDFWithTOC):
         subContStyle.leftIndent = 15
         self._styles['subContrib'] = subContStyle
 
+        details_style = stylesheet['Normal']
+        details_style.fontSize = modifiedFontSize(9.0, self._fontsize)
+        details_style.leftIndent = 10
+        self._styles['details_style'] = details_style
+
+        color_tablestyle = TableStyle([('FONTNAME', (0, 0), (-1, -1), 'Sans'),
+                              ('LEFTPADDING', (0, 0), (-1, -1), 3),
+                              ('RIGHTPADDING', (0, 0), (-1, -1), 0),
+                              ('TOPPADDING', (0, 0), (-1, -1), 0),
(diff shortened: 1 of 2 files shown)


indico/legacy/pdfinterface/conference.py
```diff
@@ -716,6 +661,43 @@ def _processDayEntries(self, day, story):
...
+        if self._ttPDFFormat.showContribId():
+            title_line = f'[{contribution.friendly_id}] {title_line}'
+        if sess_block.session.is_poster and contribution.board_number:
+            title_line += f' (board {contribution.board_number})'
+        if hasattr(contribution, 'start_dt') and not sess_block.session.is_poster:
+            title_line += f' ({format_time(contribution.start_dt, timezone=self._tz)}'
+            if self._ttPDFFormat.showLengthContribs():
+                title_line += f', {format_human_timedelta(contribution.timetable_entry.duration)})'
+            else:
+                title_line += ')'
```
ThiefMaster, line 675: Using a list and appending/prepending the parts would be nicer (and then joining the parts with an empty separator at the end). But since this is pretty much legacy code I'll leave it up to you whether you want to change it or not...

ThiefMaster (commented):
I only skimmed the diff (legacy code after all...) but AFAICT it looks good. So if it works fine I think it's good to go

indico/legacy/pdfinterface/conference.py
javfg: Of course! Change done.

ThiefMaster (approved):

PR #4962, opened 2021-06-24
Rename VC room occurrences to videoconference

### LINKED ISSUE #4845: Make wording in VC logic room-agnostic
We used to call it "VC room" because back with vidyo this was what it was called. But other systems often do not use it, they just call it a "meeting" (Zoom) for example.

Maybe we can just drop "room" in most of these places, ie "create videoconference" instead of "create videoconference room" etc?

plourenco:
Closes #4845

### DIFF
--- a/indico/modules/vc/__init__.py
+++ b/indico/modules/vc/__init__.py
@@ -54,7 +54,7 @@ def _extend_event_management_menu(sender, event, **kwargs):
 def _extend_event_menu(sender, **kwargs):
     def _visible(event):
         return bool(get_vc_plugins()) and VCRoomEventAssociation.find_for_event(event).has_rows()
-    return MenuEntryData(_('Videoconferences'), 'videoconference_rooms', 'vc.event_videoconference',
+    return MenuEntryData(_('Videoconference'), 'videoconference_rooms', 'vc.event_videoconference',
                          position=14, visible=_visible)
 
 
--- a/indico/modules/vc/__init__.py
+++ b/indico/modules/vc/__init__.py
@@ -54,7 +54,7 @@ def _extend_event_management_menu(sender, event, **kwargs):
 def _extend_event_menu(sender, **kwargs):
     def _visible(event):
         return bool(get_vc_plugins()) and VCRoomEventAssociation.find_for_event(event).has_rows()
-    return MenuEntryData(_('Videoconference Rooms'), 'videoconference_rooms', 'vc.event_videoconference',
+    return MenuEntryData(_('Videoconferences'), 'videoconference_rooms', 'vc.event_videoconference',
                          position=14, visible=_visible)
 
 
--- a/indico/modules/vc/controllers.py
+++ b/indico/modules/vc/controllers.py
@@ -54,7 +54,7 @@ def process_vc_room_association(plugin, event, vc_room, form, event_vc_room=None
 
     if event_vc_room.link_type != VCRoomLinkType.event and existing:
         db.session.rollback()
-        flash(_("There is already a meeting attached to '{link_object_title}'.").format(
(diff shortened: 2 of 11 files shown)


indico/modules/vc/__init__.py
```diff
@@ -54,7 +54,7 @@ def _extend_event_management_menu(sender, event, **kwargs):
 def _extend_event_menu(sender, **kwargs):
     def _visible(event):
         return bool(get_vc_plugins()) and VCRoomEventAssociation.find_for_event(event).has_rows()
-    return MenuEntryData(_('Videoconference Rooms'), 'videoconference_rooms', 'vc.event_videoconference',
+    return MenuEntryData(_('Videoconferences'), 'videoconference_rooms', 'vc.event_videoconference',
```
ThiefMaster, line 57: just 'Videoconference'? Not sure.. at least to me it sounds better

indico/modules/vc/controllers.py
```diff
@@ -54,7 +54,7 @@ def process_vc_room_association(plugin, event, vc_room, form, event_vc_room=None
 
     if event_vc_room.link_type != VCRoomLinkType.event and existing:
         db.session.rollback()
-        flash(_("There is already a VC room attached to '{link_object_title}'.").format(
+        flash(_("There is already a meeting attached to '{link_object_title}'.").format(
```
ThiefMaster, line 57: "a videoconference"? Probably not all VC systems call it "meeting"

indico/modules/vc/controllers.py
```diff
@@ -198,7 +198,7 @@ def _process(self):
             try:
                 self.plugin.update_room(self.vc_room, self.event)
             except VCRoomNotFoundError as err:
-                Logger.get('modules.vc').warning('VC room %r not found. Setting it as deleted.', self.vc_room)
+                Logger.get('modules.vc').warning('Meeting %r not found. Setting it as deleted.', self.vc_room)
```
ThiefMaster, line 201: same for the log messages here - I'd use "videoconference"

indico/modules/vc/templates/event_header.html
```diff
@@ -1,7 +1,7 @@
 {% from 'vc/list_vc_rooms.html' import render_vc_rooms_list  %}
 {% if not g.static_site and event_vc_rooms %}
     <div class="event-details-row">
-        <div class="event-details-label align-top">{% trans %}Videoconference Rooms{% endtrans %}</div>
+        <div class="event-details-label align-top">{% trans %}Videoconferences{% endtrans %}</div>
```
ThiefMaster, line 4:
```suggestion
        <div class="event-details-label align-top">{% trans %}Videoconference{% endtrans %}</div>
```

ThiefMaster: or maybe pluralization depending on whether there's one or more vcs?

indico/modules/vc/templates/vc_room_list.html
```diff
@@ -17,7 +17,7 @@
         </td>
         <td>
             <div class="title">
-                {% trans %}VC room{% endtrans %}
+                {% trans %}Meeting room{% endtrans %}
```
ThiefMaster, line 20: This sounds somewhat confusing. "VC name" maybe?

plourenco: Not sure if "videoconference name" is applicable here since more details are shown. What about just "Videoconferences"?

ThiefMaster: sounds fine as well

indico/modules/vc/__init__.py
plourenco: Sounds good to me too

PR #4967, opened 2021-06-28
Display more details when cloners are unavailable

### LINKED ISSUE #4913: Display more details when cloners are unavailable
For cloning it's kind of obvious (no data to clone), but for an import it's often due to a conflict which is not always obvious.
Maybe we could have cloners return a verbose message that we can then display there...

For example the TimetableCloner shows a conflict when the target event is shorter than the source event, and a user likely won't realize that (even I had to check the code for this)...

tomasr8:
Closes #4913

Adds a little warning icon to disabled options in the clone and import forms.
When hovered over, these icons provide a description of why the given option is disabled.

image
image
image
image

In order to report a reason when a cloner has conflicts, this commit replaces `Cloner.has_conflicts()` with
`Cloner.get_conflicts()` which returns a list of conflicts for the given cloner.

### DIFF
--- a/indico/modules/attachments/clone.py
+++ b/indico/modules/attachments/clone.py
@@ -27,8 +27,9 @@ class AttachmentCloner(EventCloner):
     def is_available(self):
         return self._has_content(self.old_event)
 
-    def has_conflicts(self, target_event):
-        return self._has_content(target_event)
+    def get_conflicts(self, target_event):
+        if self._has_content(target_event):
+            return [_('The target event already has materials')]
 
     def run(self, new_event, cloners, shared_data, event_exists=False):
         self._clone_nested_attachments = False
--- a/indico/modules/events/abstracts/clone.py
+++ b/indico/modules/events/abstracts/clone.py
@@ -25,8 +25,16 @@ class AbstractSettingsCloner(EventCloner):
     def is_visible(self):
         return self.old_event.type_ == EventType.conference
 
-    def has_conflicts(self, target_event):
-        return bool(target_event.abstract_review_questions) or bool(target_event.abstract_email_templates)
+    def get_conflicts(self, target_event):
+        conflicts = []
+
+        if target_event.abstract_review_questions:
+            conflicts.append(_('The target event already has review questions'))
+
+        if target_event.abstract_email_templates:
+            conflicts.append(_('The target event already has email templates'))
+
+        return conflicts
 
     @no_autoflush
     def run(self, new_event, cloners, shared_data, event_exists=False):
--- a/indico/modules/events/clone.py
+++ b/indico/modules/events/clone.py
@@ -26,8 +26,9 @@ class EventLocationCloner(EventCloner):
     def is_available(self):
         return self._has_content(self.old_event)
 
-    def has_conflicts(self, target_event):
-        return self._has_content(target_event)
+    def get_conflicts(self, target_event):
+        if self._has_content(target_event):
+            return [_('The target event already has a venue/room')]
 
     def run(self, new_event, cloners, shared_data, event_exists=False):
         with db.session.no_autoflush:
@@ -52,8 +53,9 @@ class EventPersonCloner(EventCloner):
     # We do not override `is_available` as we have cloners depending
     # on this internal cloner even if it won't clone anything.
 
-    def has_conflicts(self, target_event):
-        return target_event.persons.has_rows()
+    def get_conflicts(self, target_event):
+        if target_event.persons.has_rows():
+            return [_('The target event already has persons')]
 
     def run(self, new_event, cloners, shared_data, event_exists=False):
         self._person_map = {}
@@ -87,8 +89,12 @@ class EventPersonLinkCloner(EventCloner):
     def is_available(self):
         return self._has_content(self.old_event)
 
-    def has_conflicts(self, target_event):
-        return self._has_content(target_event)
+    def get_conflicts(self, target_event):
+        if self._has_content(target_event):
+            if self.old_event.type_ == EventType.lecture:
+                return [_('The target event already has speakers')]
+            else:
+                return [_('The target event already has chairpersons')]
 
     def run(self, new_event, cloners, shared_data, event_exists=False):
         self._person_map = shared_data['event_persons']['person_map']
@@ -114,11 +120,17 @@ class EventProtectionCloner(EventCloner):
     is_default = True
     uses = {'event_roles', 'registration_forms'}
 
-    def has_conflicts(self, target_event):
+    def get_conflicts(self, target_event):
+        conflicts = []
+
         if target_event.access_key != '':
-            return True
+            conflicts.append(_('The target event already has an access key'))
+
         entries = list(target_event.acl_entries)
-        return len(entries) != 1 or entries[0].user != target_event.creator
+        if len(entries) != 1 or entries[0].user != target_event.creator:
+            conflicts.append(_('The target event already has a custom ACL'))
+
+        return conflicts
 
     def run(self, new_event, cloners, shared_data, event_exists=False):
         self._event_role_map = shared_data['event_roles']['event_role_map'] if 'event_roles' in cloners else None
--- a/indico/modules/events/cloning.py
+++ b/indico/modules/events/cloning.py
@@ -10,6 +10,7 @@ from operator import attrgetter
 from indico.core import signals
 from indico.util.caching import memoize_request
 from indico.util.decorators import cached_classproperty
+from indico.util.i18n import _
 from indico.util.signals import named_objects_from_signal
 
 
@@ -75,7 +76,7 @@ class EventCloner:
         if event_exists:
             if any(cloner.new_event_only for name, cloner in all_cloners.items() if name in cloners):
                 raise Exception('A new event only cloner was selected')
-            if any(cloner.has_conflicts(new_event) for name, cloner in all_cloners.items() if name in cloners):
+            if any(cloner.get_conflicts(new_event) for name, cloner in all_cloners.items() if name in cloners):
                 raise Exception('Cloner target is not empty')
 
         # enable internal cloners that are enabled by default or required by another cloner
@@ -173,13 +174,16 @@ class EventCloner:
         """
         return True
 
-    def has_conflicts(self, target_event):
-        """Check for conflicts between source event and target event.
+    def get_conflicts(self, target_event):
+        """Return conflicts between source event and target event.
 
         Use this when cloning into an existing event to disable options
         where ``target_event`` data would conflict with cloned data.
+
+        Returns a dummy conflict it not overridden as cloners
+        do not support importing by default.
         """
-        return True
+        return [_('Importing data is not supported')]
 
     def _prepare_shared_data(self, shared_data):
         linked = self.uses | self.requires
--- a/indico/modules/events/management/forms.py
+++ b/indico/modules/events/management/forms.py
@@ -421,7 +421,8 @@ class ImportContentsForm(ImportSourceEventForm):
     def __init__(self, source_event, target_event, set_defaults=False, **kwargs):
         cloners = EventCloner.get_cloners(source_event)
         visible_options = [cloner for cloner in cloners if cloner.is_visible and not cloner.new_event_only]
-        conflicts = {cloner.name: cloner.has_conflicts(target_event) for cloner in cloners}
+        conflicts = {cloner.name: cloner.get_conflicts(target_event) for cloner in cloners}
+        cloners_with_conflicts = {name for name in conflicts if conflicts[name]}
 
         if set_defaults:
             form_params = {
@@ -434,7 +435,19 @@ class ImportContentsForm(ImportSourceEventForm):
         super().__init__(**kwargs)
         self.selected_items.choices = [(option.name, option.friendly_name) for option in visible_options]
         # We disable all cloners that are not available, handle only cloning to new events,
-        # have conflicts with target_event or any of their dependencies has conflicts with target_event
-        self.selected_items.disabled_choices = [option.name for option in visible_options if not option.is_available
-                                                or conflicts[option.name]
-                                                or any(conflicts[dep] for dep in option.requires)]
+        # have conflicts with target_event or any of their dependencies have conflicts with target_event
+        disabled_choices = []
+        reasons = {}
+        for option in visible_options:
+            if not option.is_available:
+                disabled_choices.append(option.name)
+                reasons[option.name] = _('There is nothing to import')
+            elif conflict := conflicts[option.name]:
+                disabled_choices.append(option.name)
+                reasons[option.name] = '\n'.join(conflict)
+            elif cloners_with_conflicts & option.requires_deep:
+                disabled_choices.append(option.name)
+                reasons[option.name] = _('This option depends on other options which are unavailable')
+
+        self.selected_items.disabled_choices = disabled_choices
+        self.selected_items.disabled_choices_reasons = reasons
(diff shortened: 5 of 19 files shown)


indico/modules/events/cloning.py
```diff
@@ -173,13 +173,13 @@ def is_available(self):
...
-    def has_conflicts(self, target_event):
-        """Check for conflicts between source event and target event.
+    def get_conflicts(self, target_event):
+        """Return conflicts between source event and target event.
 
         Use this when cloning into an existing event to disable options
         where ``target_event`` data would conflict with cloned data.
         """
-        return True
+        return []
```
tomasr8, line 182: The original implementation returns `True` by default. Should we add a dummy conflict message just in case? Although, I don't see any place where this could cause problems..

indico/web/templates/forms/checkbox_group_widget.html
```diff
@@ -6,9 +6,13 @@
     {% set field_disabled = input_args.pop('disabled', false) %}
     {% for subfield in field %}
         {% set disabled = field_disabled or subfield.data in field.disabled_choices|default([]) %}
-        <div class="i-checkbox {{ 'checkbox-disabled' if disabled}}">
-            {{ subfield(disabled=disabled, **input_args) }}
-            {{ subfield.label }}
+        {% set reason = field.disabled_choices_reasons.pop(subfield.data, '') %}
```
ThiefMaster, line 9: why `pop` and not just `get`?

indico/web/templates/forms/checkbox_group_widget.html
```diff
@@ -6,9 +6,13 @@
...
+        <div class="i-checkbox">
+            <span class="{{ 'checkbox-disabled' if disabled}}">
```
ThiefMaster, line 11:
```suggestion
            <span class="{{ 'checkbox-disabled' if disabled }}">
```

indico/web/templates/forms/checkbox_group_widget.html
```diff
@@ -6,9 +6,13 @@
...
+                {{ subfield(disabled=disabled, **input_args) }}
+                {{ subfield.label }}
+            </span>
+            {% if disabled %}<span class="icon-info icon-padding" title="{{ reason }}"></span>{% endif %}
```
ThiefMaster, line 15:
```suggestion
            {% if disabled %}
                <i class="icon-info icon-padding" title="{{ reason }}"></i>
            {% endif %}
```

indico/modules/events/abstracts/clone.py
```diff
@@ -25,8 +25,16 @@ class AbstractSettingsCloner(EventCloner):
     def is_visible(self):
         return self.old_event.type_ == EventType.conference
 
-    def has_conflicts(self, target_event):
-        return bool(target_event.abstract_review_questions) or bool(target_event.abstract_email_templates)
+    def get_conflicts(self, target_event):
+        conflicts = []
+
+        if bool(target_event.abstract_review_questions):
```
ThiefMaster, line 31: You do not need the `bool()` here; the old code only used it to avoid returning an actual list instead of a bool. But that's not needed when just checking for truthiness.

indico/modules/events/clone.py
```diff
@@ -114,11 +120,20 @@ class EventProtectionCloner(EventCloner):
     is_default = True
     uses = {'event_roles', 'registration_forms'}
 
-    def has_conflicts(self, target_event):
+    def get_conflicts(self, target_event):
+        conflicts = []
+
         if target_event.access_key != '':
-            return True
+            conflicts.append(_('The target event already contains an access key'))
```
ThiefMaster, line 127: I think "has" sounds better than "contains" in most of these cases.

indico/modules/events/clone.py
```diff
@@ -114,11 +120,20 @@ class EventProtectionCloner(EventCloner):
...
+
         entries = list(target_event.acl_entries)
-        return len(entries) != 1 or entries[0].user != target_event.creator
+        if len(entries) != 1:
+            conflicts.append(_('The target event has an incompatible number of ACLs'))
+
+        if entries[0].user != target_event.creator:
+            conflicts.append(_('The target event creator is not on the ACL'))
```
ThiefMaster, line 134: The logic here is different from what we had before. The case we want to cover (where we allow using this cloner) is when there there is exactly one ACL entry for the event creator (because that's the default in a new event).

So I'd do a check like this:

```python
if len(entries) != 1 or entries[0].user != target_event.creator:
    conflicts.append(_('The target event already has a custom ACL'))
```

indico/modules/events/cloning.py
```diff
@@ -173,13 +174,16 @@ def is_available(self):
...
 
-    def has_conflicts(self, target_event):
-        """Check for conflicts between source event and target event.
+    def get_conflicts(self, target_event):
+        """Return conflicts between source event and target event.
 
         Use this when cloning into an existing event to disable options
         where ``target_event`` data would conflict with cloned data.
+
+        By default returns a dummy confllict as cloners
```
ThiefMaster, line 183:
```suggestion
        By default returns a dummy conflict as cloners
```

indico/modules/events/timetable/clone.py
```diff
@@ -33,8 +33,15 @@ def is_default(self):
...
 
-    def has_conflicts(self, target_event):
-        return self._has_content(target_event) or self.old_event.duration > target_event.duration
+    def get_conflicts(self, target_event):
+        conflicts = []
+        if self._has_content(target_event):
+            conflicts.append(_('The target event already contains a timetable'))
+
+        if self.old_event.duration > target_event.duration:
+            conflicts.append(_('The event durations are imcompatible'))
```
ThiefMaster, line 42: `incompatible` but something like "The old event is too long" would be clearer for users

indico/web/templates/forms/checkbox_group_widget.html
tomasr8: Fixed :)

indico/modules/events/abstracts/clone.py
tomasr8: Right, I didn't realize why the `bool`s were there..

indico/modules/events/clone.py
tomasr8: Replaced `contains` with `has` in all conflict messages

indico/modules/events/clone.py
tomasr8: Got it, thanks :)

indico/modules/attachments/clone.py
```diff
@@ -27,8 +27,9 @@ class AttachmentCloner(EventCloner):
     def is_available(self):
         return self._has_content(self.old_event)
 
-    def has_conflicts(self, target_event):
-        return self._has_content(target_event)
+    def get_conflicts(self, target_event):
+        if self._has_content(target_event):
+            return [_('The target event already has attachments')]
```
ThiefMaster, line 32:
```suggestion
            return [_('The target event already has materials')]
```

(we use "attachments" internally, but "materials" in user-facing text)

indico/modules/events/clone.py
```diff
@@ -26,8 +26,9 @@ class EventLocationCloner(EventCloner):
     def is_available(self):
         return self._has_content(self.old_event)
 
-    def has_conflicts(self, target_event):
-        return self._has_content(target_event)
+    def get_conflicts(self, target_event):
+        if self._has_content(target_event):
+            return [_('The target event already has venues and/or rooms')]
```
ThiefMaster, line 31:
```suggestion
            return [_('The target event already has a venue/room')]
```

indico/modules/events/cloning.py
```diff
@@ -173,13 +174,16 @@ def is_available(self):
...
 
-    def has_conflicts(self, target_event):
-        """Check for conflicts between source event and target event.
+    def get_conflicts(self, target_event):
+        """Return conflicts between source event and target event.
 
         Use this when cloning into an existing event to disable options
         where ``target_event`` data would conflict with cloned data.
+
+        By default returns a dummy conflict as cloners
```
ThiefMaster, line 183:
```suggestion
        Returns a dummy conflict it not overridden as cloners
```

indico/modules/events/timetable/clone.py
```diff
@@ -33,8 +33,15 @@ def is_default(self):
...
 
-    def has_conflicts(self, target_event):
-        return self._has_content(target_event) or self.old_event.duration > target_event.duration
+    def get_conflicts(self, target_event):
+        conflicts = []
+        if self._has_content(target_event):
+            conflicts.append(_('The target event already has a timetable'))
+
+        if self.old_event.duration > target_event.duration:
+            conflicts.append(_('The source event is too long'))
```
ThiefMaster, line 42: Thinking about this again, maybe this is better?

```suggestion
            conflicts.append(_('The source event is longer than the target event'))
```

indico/web/templates/forms/checkbox_group_widget.html
```diff
@@ -6,9 +6,15 @@
...
-            {{ subfield(disabled=disabled, **input_args) }}
-            {{ subfield.label }}
+        {% set reason = field.disabled_choices_reasons.get(subfield.data) %}
+        <div class="i-checkbox">
+            <span class="{{ 'checkbox-disabled' if disabled }}">
+                {{ subfield(disabled=disabled, **input_args) }}
+                {{ subfield.label }}
+            </span>
+            {% if disabled %}
+                <span class="icon-info icon-padding" title="{{ reason }}"></span>
```
ThiefMaster, line 16: We use `<i>` for icons everywhere else. Don't forget to update the CSS selector you added in `_inputs.css` ;)

```suggestion
                <i class="icon-info icon-padding" title="{{ reason }}"></i>
```

indico/web/client/styles/partials/_forms.scss
```diff
@@ -66,7 +66,7 @@ form {
 
     .form-field {
       float: left;
-      width: 80%;
+      width: 85%;
```
ThiefMaster, line 69: This change is break forms pretty much everywhere:

image

image

image

indico/web/templates/forms/checkbox_group_widget.html
```diff
@@ -6,9 +6,15 @@
     {% set field_disabled = input_args.pop('disabled', false) %}
     {% for subfield in field %}
         {% set disabled = field_disabled or subfield.data in field.disabled_choices|default([]) %}
-        <div class="i-checkbox {{ 'checkbox-disabled' if disabled}}">
-            {{ subfield(disabled=disabled, **input_args) }}
-            {{ subfield.label }}
+        {% set reason = field.disabled_choices_reasons.get(subfield.data) %}
```
ThiefMaster, line 9: This field is used elsewhere (e.g. when managing oauth apps) where `disabled_choices_reasons` is not defined

indico/web/templates/forms/checkbox_group_widget.html
ThiefMaster: I think `icon-warning` would look better there

indico/web/client/styles/partials/_forms.scss
tomasr8: I reverted this for this PR as it is unrelated and I'll fix it in a separate commit.

indico/web/templates/forms/checkbox_group_widget.html
tomasr8: Thanks, fixed :)

indico/modules/events/management/forms.py
```diff
@@ -335,6 +335,8 @@ def __init__(self, event, set_defaults=False, **kwargs):
         super().__init__(**kwargs)
         self.selected_items.choices = [(option.name, option.friendly_name) for option in visible_options]
         self.selected_items.disabled_choices = [option.name for option in visible_options if not option.is_available]
+        self.selected_items.disabled_choices_reasons = {name: _('This module does not contain any data')
+                                                        for name in self.selected_items.disabled_choices}
```
ThiefMaster, line 339: not really needed - it's pretty clear why something is disabled when cloning

indico/modules/events/management/forms.py
```diff
@@ -434,7 +436,23 @@ def __init__(self, source_event, target_event, set_defaults=False, **kwargs):
...
-        self.selected_items.disabled_choices = [option.name for option in visible_options if not option.is_available
-                                                or conflicts[option.name]
-                                                or any(conflicts[dep] for dep in option.requires)]
+        # have conflicts with target_event or any of their dependencies have conflicts with target_event
+        disabled_choices = []
+        reasons = {}
+        for option in visible_options:
+            if not option.is_available:
+                disabled_choices.append(option.name)
+                reasons[option.name] = _('This module does not contain any data')
```
ThiefMaster, line 445:
```suggestion
                reasons[option.name] = _('There is nothing to import')
```

indico/modules/events/management/forms.py
```diff
@@ -434,7 +434,23 @@ def __init__(self, source_event, target_event, set_defaults=False, **kwargs):
...
+            if not option.is_available:
+                disabled_choices.append(option.name)
+                reasons[option.name] = _('There is nothing to import')
+            elif conflicts[option.name]:
+                disabled_choices.append(option.name)
+                # Show only the first conflict
+                reasons[option.name] = conflicts[option.name][0]
+            else:
+                for dep in option.requires:
+                    if conflicts[dep]:
```
ThiefMaster, line 450: I think `elif set(conflicts) & option.requires` does the same as these 3 lines and the break.

Please test it though, since I'm not sure what type `option.requires` is (might need an explicit `set()` as well if it's not a set)

tomasr8: I'll check it :+1:

ThiefMaster: And maybe this check should use `option.requires_deep`?

tomasr8: Yes, I think we need `option.requires_deep` unless `visible_options` is topologically sorted (it isn't). We could have a situation where
`A -> B -> C` (`->` = depends on) and `C` has a conflict. If we process `A` before `B`, we won't disable `A`.

tomasr8: Should be fixed now

indico/web/client/styles/partials/_inputs.scss
```diff
@@ -382,6 +382,14 @@ $dropdown-transition: $dropdown-transition-step ease-out;
     display: inline-block;
     margin-left: 0.5em;
   }
+
+  i.icon-padding {
+    padding-left: 5px;
+  }
+
+  span {
```
ThiefMaster, line 390: maybe safer to use a class here?

indico/modules/events/management/forms.py
```diff
@@ -434,7 +435,20 @@ def __init__(self, source_event, target_event, set_defaults=False, **kwargs):
...
+        disabled_choices = []
+        reasons = {}
+        for option in visible_options:
+            if not option.is_available:
+                disabled_choices.append(option.name)
+                reasons[option.name] = _('There is nothing to import')
+            elif conflict := conflicts[option.name]:
+                disabled_choices.append(option.name)
+                # Show only the first conflict
+                reasons[option.name] = conflict[0]
```
ThiefMaster, line 448: hm, why not show all of them, e.g. separated with linebreaks or semicolons?

indico/web/client/styles/partials/_inputs.scss
tomasr8: Added :+1:

indico/modules/events/management/forms.py
tomasr8: Added line breaks when there are multiple conflicts (check the last screenshot).

ThiefMaster (approved):

PR #4968, opened 2021-06-28
Include content in abstract listing

plourenco:
A simple proposal for the inclusion of content/description in the abstract listing, making use of another table row that simulates being part of its sibling.

This does not include collapsible rows as it didn't seem useful in a dynamic row, but suggestions are welcome.

imagem


indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -236,6 +236,16 @@
                                     {% endif %}
                                 </td>
                             </tr>
+                            {% for item in static_columns %}
+                                {% if item.id == 'description' %}
```
ThiefMaster, line 240: Since it's unlikely that we need to handle other columns here, maybe we could use this shortcut?

```
{% for item in static_columns if item.id == 'description' %}
```

ThiefMaster (commented):
This deserves a changelog entry

ThiefMaster (approved):

PR #4976, opened 2021-06-30
Add personal user tokens 🔑

### LINKED ISSUE #1934: Bring OAuth and API key system closer together
Currently we have:

- API keys: One per user, they grant full access with whatever permissions the associated users have
- OAuth tokens: Tied to an application, they only grant specific permissions

This requires you to use the full-access API key if you want to use it in a custom application (since not everyone can add new OAuth apps on Indico, and the OAuth flow is not very nice if you don't have something used by multiple users - e.g. drupal feeds on your site).

---

There is be a much cleaner way, similar to what GitHub does:

- Users can create oauth-token like keys and select the scopes they are granted
- Instead of API keys you use these tokens
- They can be revoked at any time and are usually specific to one external application so if one is compromised you don't have to reset all of them like it's the case with the current api key system

---

Migration-wise, existing API keys could be simply converted to a token with full permissions.

The only disadvantage I see is that signatures would probably go away. However, considering that OAuth doesn't use signatures either and pretty much no APIs nowadays go through all the trouble of signing requests I think that wouldn't be a problem. At least when HTTPS is being used, which should be the case anyway.

---

This would obviously be something for once the migration is finished - e.g. 2.0 or 2.1. Personally I'd go for 2.0 if we do this and have time between finishing the migration, cleaning things up, and releasing the version. That way there would be no "big" changes related to this in 2.1.

ThiefMaster:
This is basically what I suggested in #1934 six years ago, and originally wanted to have in 2.0 or 2.1, but then there was never time for it... In the end this was very straightforward to implement with all the previous oauth-related changes.

Initially I had planned on just providing the backend and CLI for it - that way people could use have used it for administrative tasks (thinking of some of the API usages on our own instance that currently all use the same API key...).

However, since the UI for it wouldn't be particularly complex, I'm now thinking about spending another half day or so to add end-user UI for it as well since that would allow us to officially deprecate the existing API key authentication in favor of using these tokens. (Edit: done!)

And with it being deprecated, we can then consider removing the old API auth in one of the 3.x versions.

I also used this opportunity to prefix our tokens - `indo_` for OAuth, `indp_` for personal tokens - this seems to be common on [various high-profile public APIs](https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/) lately.

closes #1934

---

Some screenshots of the UI - it is very similar to GitHub's personal token management ;)

image
image
image


indico/modules/oauth/__init__.py
```diff
@@ -22,3 +23,5 @@ def _extend_admin_menu(sender, **kwargs):
 @signals.menu.items.connect_via('user-profile-sidemenu')
 def _extend_profile_sidemenu(sender, user, **kwargs):
     yield SideMenuItem('applications', _('Applications'), url_for('oauth.user_apps'), 40, disabled=user.is_system)
+    if can_manage_personal_tokens() or user.personal_tokens.filter_by(revoked_dt=None).has_rows():
```
plourenco, line 26: What do you think about adding a `active_tokens` or similar column property to `User´? This would avoid repeating the `revoked_dt` filter.

ThiefMaster: yeah, I started thinking the same when writing this code...

ThiefMaster: updated; i added a util instead of a second relationship since overlapping relationships are a bit ugly (and fragile)

indico/migrations/versions/20210630_1349_5cbb0eb12eb3_add_last_used_ip_and_use_count_to_tokens.py
```diff
@@ -0,0 +1,33 @@
...
+Create Date: 2021-06-30 13:49:55.566636
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+
+# revision identifiers, used by Alembic.
+revision = '5cbb0eb12eb3'
```
javfg, line 14: Just out of curiosity, is it not better to compact all the migrations into one after you are done with the PR?

ThiefMaster: if they all touched the same tables then yes - but this one also affects the (existing) table with the oauth tokens. So we need a revision for this anyway...

Moving the changes made to the new table into the original revision creating it wouldn't be a bad idea per se. The main reason I didn't do it is that I don't plan to squash the commits of this PR as they are nicely split into different somewhat unrelated parts. Even though the only place where this would be an issue is when someone wanted to git-bisect and also follow the DB changes in the commits (which is very unlikely to happen): When you are on the commit adding these fields and want to downgrade, you'd also have to downgrade the creation of the personal tokens table, even though that revision was introduced in another commit..

javfg (approved):

docs/source/http_api/access.rst
```diff
@@ -21,23 +25,98 @@ where:
...
+
+.. exec::
+    def main():
+        from indico.core.oauth.scopes import SCOPES
+        for name, desc in sorted(SCOPES.items(), key=lambda x: x[0].split(':')[::-1]):
+            print(f'- ``{name}`` - {desc}')
+
+    main()
+
+The ``everything`` scopes are special because they can be used with *any* Indico endpoint, ie they are
```
pferreir, line 83: "i.e."?

PR #4980, opened 2021-06-30
Make it less likely to accidentally create mixed categories

### LINKED ISSUE #4971: Make it less likely to accidentally create mixed categories
Currently it's really easy to accidentally create events in the root category if creation there is not restricted or you are an admin/manager as it's pre-filled (while previously you were forced to actively pick a category).

To improve this:

- Show a warning in the event creation dialog when a category is selected which contains subcategories but no events
- Do not pre-fill the root category if it contains subcategories and no events

For the second point, check the `single_category` logic in `RHCreateEvent` - it probably just needs to be adapted

tomasr8:
Closes #4971

image

Currently, the category field in the event creation form gets prefilled with the current category. This can lead
to creating events in categories which should only contain other subcategories, if one is not careful. This could be
especially problematic in the case of root category.

This commit changes the behavior such that it does not prefill categories which only contain other categories
and no events. In the case of non-root categories, we also require that the user has create access
to the category in order to stop the prefilling.

This is to maintain the user friendliness of the category picker for non-admin users as the picker
starts at the root category if no category is prefilled.

### DIFF
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -125,8 +125,8 @@ class RHCategoryInfo(RHDisplayCategoryBase):
         children_strategy.subqueryload('acl_entries')
         children_strategy.undefer('deep_children_count')
         children_strategy.undefer('deep_events_count')
-        children_strategy.undefer('has_events'),
-        children_strategy.undefer('has_children'),
+        children_strategy.undefer('has_events')
+        children_strategy.undefer('has_children')
         return (children_strategy,
                 load_only('id', 'parent_id', 'title', 'protection_mode'),
                 subqueryload('acl_entries'),
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -125,13 +125,15 @@ class RHCategoryInfo(RHDisplayCategoryBase):
         children_strategy.subqueryload('acl_entries')
         children_strategy.undefer('deep_children_count')
         children_strategy.undefer('deep_events_count')
-        children_strategy.undefer('has_events')
+        children_strategy.undefer('has_events'),
+        children_strategy.undefer('has_children'),
         return (children_strategy,
                 load_only('id', 'parent_id', 'title', 'protection_mode'),
                 subqueryload('acl_entries'),
                 undefer('deep_children_count'),
                 undefer('deep_events_count'),
                 undefer('has_events'),
+                undefer('has_children'),
                 undefer('chain'))
 
     def _process(self):
@@ -163,7 +165,7 @@ class RHCategorySearch(RH):
         query = (Category.query
                  .filter(Category.title_matches(q))
                  .options(undefer('deep_children_count'), undefer('deep_events_count'), undefer('has_events'),
-                          joinedload('acl_entries')))
+                          undefer('has_children'), joinedload('acl_entries')))
         if session.user:
             # Prefer favorite categories
             query = query.order_by(Category.favorite_of.any(favorite_category_table.c.user_id == session.user.id)
--- a/indico/modules/categories/serialize.py
+++ b/indico/modules/categories/serialize.py
@@ -101,6 +101,7 @@ def serialize_category(category, with_favorite=False, with_path=False, parent_pa
         'title': category.title,
         'is_protected': category.is_protected,
         'has_events': category.has_events,
+        'has_children': category.has_children,
         'deep_category_count': category.deep_children_count,
         'deep_event_count': category.deep_events_count,
         'can_access': category.can_access(session.user),
--- a/indico/modules/events/controllers/creation.py
+++ b/indico/modules/events/controllers/creation.py
@@ -46,16 +46,24 @@ class RHCreateEvent(RHProtected):
     def _process_args(self):
         self.event_type = EventType[request.view_args['event_type']]
         self.root_category = Category.get_root()
-        self.single_category = not self.root_category.children
+        self.single_category = not self.root_category.has_children
+
+    def _has_only_subcategories(self, category):
+        return category.has_children and not category.has_events
 
     @cached_property
     def _default_category(self):
         try:
             category_id = int(request.args['category_id'])
         except (ValueError, KeyError):
-            return self.root_category if self.single_category else None
+            return None if self._has_only_subcategories(self.root_category) else self.root_category
+
+        category = Category.get(category_id, is_deleted=False)
+        if (category and self._has_only_subcategories(category) and
+                (category.is_root or category.can_create_events(session.user))):
+            return None
         else:
-            return Category.get(category_id, is_deleted=False)
+            return category
 
     def _get_prepared_data(self):
         try:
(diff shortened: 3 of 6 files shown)


indico/modules/events/controllers/creation.py
```diff
@@ -48,14 +48,22 @@ def _process_args(self):
...
     @cached_property
     def _default_category(self):
         try:
             category_id = int(request.args['category_id'])
         except (ValueError, KeyError):
-            return self.root_category if self.single_category else None
+            return None if self._has_only_subcategories(self.root_category) else self.root_category
+
+        category = Category.get(category_id, is_deleted=False)
+        if category and self._has_only_subcategories(category) and \
```
ThiefMaster, line 62: we don't use `\` ;) better use `()` around the whole condition - either black-style or with double indentation in the wrapped line

indico/web/client/js/jquery/widgets/jinja/category_picker_widget.js
```diff
@@ -51,6 +52,12 @@
         openInDialog: true,
         actionOn,
         onAction(category) {
+          if (category.has_subcategories && !category.has_events) {
+            $categoryWarning.removeClass('hidden');
+          } else {
+            $categoryWarning.addClass('hidden');
+          }
```
ThiefMaster, line 59: there's `toggleClass` ;)

indico/web/templates/forms/category_picker_widget.html
```diff
@@ -1,9 +1,15 @@
 {% extends 'forms/base_widget.html' %}
+{% from 'message_box.html' import message_box %}
 
 
 {% block html %}
     <div class="i-form-field-fixed-width">
         <div id="category-title-{{ field.id }}" class="text-holder-box flexrow f-a-center" data-tooltip-anchor></div>
+        <div id="category-warning-{{ field.id }}" class="hidden">
+            {% call message_box('warning') %}
+                {% trans %}The selected category does not contain any events, only subcategories{% endtrans %}
```
ThiefMaster, line 10: I'd add "Please make sure that you really want to create an event there."

indico/modules/categories/serialize.py
```diff
@@ -101,6 +101,7 @@ def serialize_category(category, with_favorite=False, with_path=False, parent_pa
         'title': category.title,
         'is_protected': category.is_protected,
         'has_events': category.has_events,
+        'has_subcategories': bool(category.children),
```
ThiefMaster, line 104: Better add `has_children` similar to `has_events` (make sure to undefer it as well when querying) - no need to query all the subcategories if we can just get the count

indico/modules/events/controllers/creation.py
```diff
@@ -48,14 +48,22 @@ def _process_args(self):
         self.root_category = Category.get_root()
         self.single_category = not self.root_category.children
 
+    def _has_only_subcategories(self, category):
+        return not bool(category.events) and bool(category.children)
```
ThiefMaster, line 52: `has_*` (see my other comment below)

indico/web/client/js/jquery/widgets/jinja/category_picker_widget.js
tomasr8: Cool, didn't know `toggleClass` takes a second argument :+1:

indico/modules/categories/controllers/display.py
```diff
@@ -163,7 +165,7 @@ def _process(self):
         query = (Category.query
                  .filter(Category.title_matches(q))
                  .options(undefer('deep_children_count'), undefer('deep_events_count'), undefer('has_events'),
-                          joinedload('acl_entries')))
+                          undefer('has_children'), joinedload('acl_entries')))
```
tomasr8, line 168: Not sure If I should add `undefer(has_children)` in other places such as: https://github.com/indico/indico/pull/4980/files#diff-791a87a96ed182d72dbd69f1d61092083942dd060d291db0b1611abe5e84e729R212

indico/modules/categories/serialize.py
tomasr8: Thanks for the help :) I think it should be good now :+1:

indico/modules/categories/controllers/display.py
ThiefMaster: basically code that results in calling `serialize_category` should undefer it when querying the category passed to it to avoid triggering the extra query

tomasr8: Thanks! That makes sense.. In that case I think I've covered everything.

indico/modules/categories/controllers/display.py
```diff
@@ -125,13 +125,15 @@ def _category_query_options(cls):
         children_strategy.subqueryload('acl_entries')
         children_strategy.undefer('deep_children_count')
         children_strategy.undefer('deep_events_count')
-        children_strategy.undefer('has_events')
+        children_strategy.undefer('has_events'),
+        children_strategy.undefer('has_children'),
```
ThiefMaster, line 129: Those are normal function calls, not a list of arguments ;)

```suggestion
        children_strategy.undefer('has_events')
        children_strategy.undefer('has_children')
```

indico/modules/events/controllers/creation.py
```diff
@@ -46,16 +46,24 @@ class RHCreateEvent(RHProtected):
     def _process_args(self):
         self.event_type = EventType[request.view_args['event_type']]
         self.root_category = Category.get_root()
-        self.single_category = not self.root_category.children
+        self.single_category = not self.root_category.has_children
```
ThiefMaster, line 49: It's only used in one place, so I'd move the `self.root_category.has_children` check down there

indico/modules/events/controllers/creation.py
```diff
@@ -46,16 +46,24 @@ class RHCreateEvent(RHProtected):
...
     def _default_category(self):
         try:
             category_id = int(request.args['category_id'])
         except (ValueError, KeyError):
-            return self.root_category if self.single_category else None
+            return None if self._has_only_subcategories(self.root_category) else self.root_category
+
+        category = Category.get(category_id, is_deleted=False)
+        if (category and self._has_only_subcategories(category) and
+                (category.is_root or category.can_create_events(session.user))):
```
ThiefMaster, line 63: `category.can_create_events(session.user)` isn't correct (it takes admin privs into account and the original idea was to not prefill if the user *cannot* create events in there).

Anyway, now that you added the message in the dialog, it indeed feels a bit confusing (so @pferreir's initial comment was right ;)) if it doesn't get prefilled..

```suggestion
        if category and category.is_root and self._has_only_subcategories(category):
```

indico/web/client/js/jquery/widgets/jinja/category_picker_widget.js
```diff
@@ -51,6 +52,8 @@
         openInDialog: true,
         actionOn,
         onAction(category) {
+          $categoryWarning.toggleClass('hidden', !category.has_children || category.has_events);
```
ThiefMaster, line 55: We also need to toggle it for the initially-prefilled category - right now we don't show it there, even though that's probably the most important time to show it.

PS: Maybe move this one after the variable definitions? ;)

ThiefMaster (approved):

PR #4981, opened 2021-06-30
Add db_query signal and signal_query method

OmeGak:
Add a `db_query` signal that allows a handler modify the query object. The `signal_query` method is expected to be used for sending the signal in order to guarantee that only one handler modifies the query.

An example of usage is the following:

```python
query = (Registration.query.with_parent(self.event)
         .filter(Registration.is_state_publishable, ~RegistrationForm.is_deleted)
         .options(subqueryload('data').joinedload('field_data'),
                  contains_eager('registration_form'))
         .signal_query('publishable-registrations-in-participant-list'))
```

On the signal handler side, the query can be intercepted like this:

```python
@signals.core.db_query.connect_via('publishable-registrations-in-participant-list')
def _intercept_publishable_registrations_in_participant_list(sender, query, **kwargs):
    return query.filter(...)
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -76,6 +76,9 @@ Internal Changes
 - Add :data:`EMAIL_BACKEND` configuration variable to support different email sending
   backends e.g. during development (:issue:`5375`, :pr:`5376`, thanks :user:`Moist-Cat`)
 - Make model attrs to clone interceptable by plugins (:pr:`5403`, thanks :user:`omegak`)
+- Add ``signal_query`` method in the ``IndicoBaseQuery`` class and the ``db_query``
+  signal, allowing to intercept and modify queries by signal handlers (:pr:`4981`,
+  thanks :user:`omegak`).
 
 
 ----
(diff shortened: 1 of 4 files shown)


indico/core/signals/__init__.py
```diff
@@ -8,6 +8,7 @@
 from indico.core.signals import (acl, agreements, attachments, category, event, event_management, menu, plugin, rb, rh,
                                  users)
 from indico.core.signals.core import *
+from indico.core.signals.lowlevel import *
```
ThiefMaster, line 11: I would keep it as a module import so it's explicit via `indico.core.signals.lowlevel.*`

indico/core/signals/lowlevel.py
```diff
@@ -0,0 +1,17 @@
...
+
+from blinker import Namespace
+
+
+_signals = Namespace()
+
+db_query = _signals.signal('db-query', '''
+The *sender* is a string identifying the context.  The *query* kwarg contains
+the query object to be customized.  The *context* kwarg contains a dictionary
+with additional values that may be required to customize the query.
```
ThiefMaster, line 16: Not sure if we need the context kwarg - I'd just pass the context directly as signal kwargs...

PR #4983, opened 2021-07-01
Implement flat category view

### LINKED ISSUE #4982: Allow displaying all events descending from a category
In the category display view, rather than listing only the events that are directly linked to the category, it could be possible to list all the events descending to the category (i.e. events linked to subcategories).

This is a feature that was implemented on the UN plugin via monkeypatching. As we are consolidating the plugin's codebase and and relying more on signals it's becoming clear: 1) that this feature requires a non-negligible number of signals to be properly implemented on the plugin side and 2) that this feature is of general usefulness to all Indico instances.

## UI proposal

The flat list of events could be triggered by:
- URL parameter (e.g. `?flat=1`, `?deep=1`)
- Link under the `View` button (eye) in category pages (e.g. `All descending events`).

|view|trigger|
|:-|:-|
|image|image|

## Performance considerations

- The implementation relies on `Category.deep_children_query` to fetch all subcategories. Is this something that may cause problems when done from the Home category, for instance?
- In large Indico instances, displaying ALL the current events descending from the Home category can produce an awfully long list. Is this something that should be limited to X number of events? Could adding some form of pagination (besides showing past and future events) relieve the strain client-side?

OmeGak:
Closes #4982.

### DIFF
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -23,9 +23,9 @@ from indico.core.db import db
 from indico.core.db.sqlalchemy.colors import ColorTuple
 from indico.core.db.sqlalchemy.util.queries import get_n_matching
 from indico.modules.categories.controllers.base import RHDisplayCategoryBase
-from indico.modules.categories.controllers.util import (get_category_view_params, group_by_month,
-                                                        make_format_event_date_func, make_happening_now_func,
-                                                        make_is_recent_func)
+from indico.modules.categories.controllers.util import (get_category_view_params, get_event_query_filter,
+                                                        group_by_month, make_format_event_date_func,
+                                                        make_happening_now_func, make_is_recent_func)
 from indico.modules.categories.models.categories import Category
 from indico.modules.categories.serialize import (serialize_categories_ical, serialize_category, serialize_category_atom,
                                                  serialize_category_chain)
@@ -254,11 +254,11 @@ class RHEventList(RHDisplayCategoryEventsBase):
         if before is None and after is None:
             raise BadRequest('"before" or "after" parameter must be specified')
         hidden_event_ids = {e.id for e in self.category.get_hidden_events(user=session.user, deep=self.is_flat)}
-        event_query_filter = (Event.category_chain_overlaps(self.category.id) if self.is_flat
-                              else (Event.category_id == self.category.id))
+        event_query_filter = get_event_query_filter(self.category, is_flat=self.is_flat,
+                                                    hidden_event_ids=hidden_event_ids)
         event_query = (Event.query
                        .options(*self._event_query_options)
-                       .filter(event_query_filter, Event.id.notin_(hidden_event_ids))
+                       .filter(event_query_filter)
                        .order_by(Event.start_dt.desc(), Event.id.desc()))
         if before:
             event_query = event_query.filter(Event.start_dt < before)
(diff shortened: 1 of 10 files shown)


indico/modules/categories/controllers/display.py
```diff
@@ -252,10 +253,12 @@ def _process_args(self):
...
             raise BadRequest('"before" or "after" parameter must be specified')
-        hidden_event_ids = {e.id for e in self.category.get_hidden_events(user=session.user)}
-        event_query = (Event.query.with_parent(self.category)
+        hidden_event_ids = {e.id for e in self.category.get_hidden_events(user=session.user, deep=self.is_flat)}
+        event_query_filter = (Event.category_chain_overlaps(self.category.id) if self.is_flat
+                              else (Event.category_id == self.category.id))
+        event_query = (Event.query
                        .options(*self._event_query_options)
-                       .filter(Event.id.notin_(hidden_event_ids))
+                       .filter(event_query_filter, Event.id.notin_(hidden_event_ids))
```
ThiefMaster, line 261: needs `~Event.is_deleted`

indico/modules/categories/controllers/util.py
```diff
@@ -64,37 +64,43 @@ def make_is_recent_func(now):
...
     if next_event_start_dt is not None and next_event_start_dt > future_threshold:
         future_threshold = next_event_start_dt + relativedelta(months=1, day=1, hour=0, minute=0)
     if previous_event_start_dt is not None and previous_event_start_dt < past_threshold:
         past_threshold = previous_event_start_dt.replace(day=1, hour=0, minute=0)
-    event_query = (Event.query.with_parent(category)
+    event_query = (Event.query
                    .options(*RHDisplayCategoryEventsBase._event_query_options)
-                   .filter(Event.id.notin_(hidden_event_ids))
+                   .filter(event_query_filter,
+                           Event.id.notin_(hidden_event_ids))
```
ThiefMaster, line 97: This needs an explicit `~Event.is_deleted` now - previously it was done in the `with_parent` as the relationship excludes deleted events.

ThiefMaster: maybe better add it to `event_query_filter` since `next_event_start_dt` etc need it as well (the fact that it wasn't there was actually a bug).

this ugly ` .first() or (None,))[0]` can probably be replaced with `.scalar()` btw

indico/modules/categories/controllers/display.py
```diff
@@ -252,10 +253,12 @@ def _process_args(self):
         after = self._parse_year_month(request.args.get('after'))
         if before is None and after is None:
             raise BadRequest('"before" or "after" parameter must be specified')
-        hidden_event_ids = {e.id for e in self.category.get_hidden_events(user=session.user)}
-        event_query = (Event.query.with_parent(self.category)
+        hidden_event_ids = {e.id for e in self.category.get_hidden_events(user=session.user, deep=self.is_flat)}
+        event_query_filter = (Event.category_chain_overlaps(self.category.id) if self.is_flat
```
ThiefMaster, line 257: This one should also use `is_visible_in`. Maybe it's time to add util for this since we use this in multiple places now...

```python
def get_event_query_filter(category, is_flat=False, hidden_event_ids=None):
    criteria = [
        Event.is_visible_in(category.id) if is_flat else (Event.category_id == category.id),
        ~Event.is_deleted,
    ]
    if hidden_event_ids:
        criteria.append(Event.id.notin_(hidden_event_ids))
    return db.and_(*criteria)
```

ThiefMaster (approved):

vasantvohra:
Great Job @OmeGak, @ThiefMaster, @openprojects 👍

PR #4986, opened 2021-07-02
Timetable: add warning in case of draft mode

### LINKED ISSUE #4977: Timetable: add warning in case of draft mode
We have a "Draft" switch in the list of contributions, which decides whether the Timetable is published or not.

image

This is at times confusing for users. One nice improvement would be to add a warning at the top of the "Timetable management" page, something like "The timetable is in draft move". (followed by a "publish it" button)

tomasr8:
Adds a warning to the manage/timetable page of
conferences in case the draft mode is enabled.
This is the same warning shown on the main settings
page.

Instead of showing the 'Manage contributions' button,
we change the warning to include a 'Publish now', which
turns off the draft mode. We show the same modal that
is shown on the contribution settings page
('manage/contributions') when the button is clicked.

image
image

closes #4977

### DIFF
--- a/indico/modules/events/contributions/client/js/PublicationButton.jsx
+++ b/indico/modules/events/contributions/client/js/PublicationButton.jsx
@@ -9,32 +9,31 @@ import publicationURL from 'indico-url:contributions.manage_publication';
 
 import PropTypes from 'prop-types';
 import React, {useState} from 'react';
-import {Checkbox} from 'semantic-ui-react';
 
-import {useTogglableValue} from 'indico/react/hooks';
+import {IButton} from 'indico/react/components';
 import {Translate} from 'indico/react/i18n';
+import {handleAxiosError, indicoAxios} from 'indico/utils/axios';
 
 import PublicationModal from './PublicationModal';
 
-export default function PublicationSwitch({eventId}) {
+export default function PublicationButton({eventId, onSuccess}) {
+  const url = publicationURL({event_id: eventId});
   const [modalOpen, setModalOpen] = useState(false);
 
-  const [published, togglePublished, loading, saving] = useTogglableValue(
-    publicationURL({event_id: eventId})
-  );
-
-  if (loading) {
-    return null;
-  }
+  const onClick = async () => {
+    try {
+      await indicoAxios.put(url);
+    } catch (error) {
+      return handleAxiosError(error);
+    }
+    onSuccess();
+    setModalOpen(false);
+  };
 
   const trigger = (
-    <Checkbox
-      label={published ? Translate.string('Published') : Translate.string('Draft')}
-      toggle
-      onClick={() => setModalOpen(true)}
-      checked={published}
-      disabled={saving}
-    />
+    <IButton onClick={() => setModalOpen(true)}>
+      <Translate>Publish now</Translate>
+    </IButton>
   );
 
   return (
@@ -42,15 +41,13 @@ export default function PublicationSwitch({eventId}) {
       open={modalOpen}
       setModalOpen={setModalOpen}
       trigger={trigger}
-      published={published}
-      onClick={() => {
-        togglePublished();
-        setModalOpen(false);
-      }}
+      published={false}
+      onClick={onClick}
     />
   );
 }
 
-PublicationSwitch.propTypes = {
+PublicationButton.propTypes = {
   eventId: PropTypes.string.isRequired,
+  onSuccess: PropTypes.func.isRequired,
 };
--- a/indico/modules/events/management/templates/settings.html
+++ b/indico/modules/events/management/templates/settings.html
@@ -1,5 +1,6 @@
 {% extends 'events/management/base.html' %}
 {% from 'events/management/_settings.html' import render_event_settings %}
+{% from 'events/contributions/management/_draft_mode_warning.html' import render_draft_mode_warning %}
 
 {% block title %}{% trans %}Settings{% endtrans %}{% endblock %}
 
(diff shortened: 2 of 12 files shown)


indico/modules/events/timetable/templates/_draft_mode_warning.html
```diff
@@ -0,0 +1,26 @@
...
+        <div class="text">
+            <div class="label">{% trans %}Contributions are in draft mode{% endtrans %}</div>
+            {% trans %}
+                While in draft mode, regular users cannot see the contributions and timetable.
+            {% endtrans %}
+        </div>
+        <div class="toolbar">
+            <div class="group">
+                <button id="draft-mode-warning-button" class="i-button" data-event-id="{{ event.id }}">
+                </button>
```
ThiefMaster, line 15: I think it'd be better to just put an empty span with the ID and data attribute here and then let the JSX render the IButton

indico/modules/events/timetable/controllers/manage.py
```diff
@@ -34,9 +35,16 @@ class RHManageTimetable(RHManageTimetableBase):
...
 
     def _process(self):
+        from indico.modules.events.contributions import contribution_settings
+
         event_info = serialize_event_info(self.event)
         timetable_data = TimetableSerializer(self.event, management=True).serialize_timetable()
+        show_draft_warning = (self.event.type_ == EventType.conference and
+                              not contribution_settings.get(self.event, 'published') and
+                              (TimetableEntry.query.with_parent(self.event).has_rows() or
+                               Contribution.query.with_parent(self.event).has_rows()))
```
ThiefMaster, line 45: We have this in two places now. Maybe add a `should_show_draft_warning` util to `indico/modules/events/util.py` and call it in both places?

indico/modules/events/management/templates/settings.html
```diff
@@ -1,5 +1,6 @@
 {% extends 'events/management/base.html' %}
 {% from 'events/management/_settings.html' import render_event_settings %}
+{% from 'events/timetable/_draft_mode_warning.html' import render_draft_mode_warning %}
```
ThiefMaster, line 3: Maybe this should be in contributions since that's where all the rest of the draft mode stuff resides

indico/modules/events/contributions/client/js/PublicationSwitch.jsx
```diff
@@ -25,87 +27,54 @@ export default function PublicationSwitch({eventId}) {
...
+  ) : (
+    <>
+      <p>
+        <Translate>Are you sure you want to publish the contribution list?</Translate>
+      </p>
+      <p>
+        <Translate>By doing so the following menu items will be accessible:</Translate>
+      </p>
+    </>
+  );
```
ThiefMaster, line 64: Why not move this into PublicationModal as well?

indico/modules/events/contributions/client/js/PublicationButton.jsx
```diff
@@ -0,0 +1,62 @@
...
+import PublicationModal from './PublicationModal';
+
+export default function PublicationButton({eventId, onSuccess}) {
+  const url = publicationURL({event_id: eventId});
+  const [modalOpen, setModalOpen] = useState(false);
+
+  const onPublish = async () => {
+    try {
+      await indicoAxios.put(url);
+      onSuccess();
```
ThiefMaster, line 25: I'd move this after the try..catch (and add a `return` in the catch) - that way we don't catch errors in that callback.

indico/modules/events/contributions/client/js/PublicationButton.jsx
```diff
@@ -0,0 +1,62 @@
...
+    } catch (error) {
+      handleAxiosError(error);
+    }
+  };
+
+  const trigger = <span onClick={() => setModalOpen(true)}>Publish now</span>;
```
ThiefMaster, line 31: missing i18n

indico/modules/events/contributions/client/js/PublicationButton.jsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <PublicationModal
+      open={modalOpen}
+      setModalOpen={setModalOpen}
+      trigger={trigger}
+      header={Translate.string('Publish now')}
+      body={body}
+      onClick={() => {
+        onPublish();
+        setModalOpen(false);
```
ThiefMaster, line 53: Maybe this should go into onPublish (and only run in case of success)?

indico/modules/events/timetable/controllers/manage.py
tomasr8: Added :+1:

indico/modules/events/management/templates/settings.html
tomasr8: Moved to `events/contributions/management`

indico/modules/events/contributions/client/js/PublicationSwitch.jsx
tomasr8: Also moved the header in as well

indico/modules/events/contributions/client/js/PublicationButton.jsx
tomasr8: I was thinking in case of error we may want to hide the `PublicationModal` anyway?

ThiefMaster: since we don't expect errors there and the user can just close it anyway, i'd rather not do that automatically

tomasr8: ok changed :+1:

PR #4989, opened 2021-07-05
Allow conveners to edit the abstract track list

plourenco:


indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -278,7 +278,7 @@ class RHDisplayAbstractsExportExcel(AbstractsExportExcel, RHDisplayAbstractsActi
 
 class RHEditReviewedForTrackList(RHAbstractBase):
     def _check_abstract_protection(self):
-        return self.abstract.can_judge(session.user, check_state=True)
+        return self.abstract.can_convene(session.user) and not self.abstract.is_in_final_state
```
plourenco, line 281: RFC: Would it be worth to have a `can_convene(session.user, check_state=True)`?

ThiefMaster: sounds like a good idea

ThiefMaster:
This deserves an "Improvement" changelog entry btw

indico/modules/events/abstracts/settings.py
```diff
@@ -129,6 +129,7 @@ class BOALinkFormat(RichEnum):
     'scale_upper': 5,
     'allow_comments': True,
     'allow_convener_judgment': False,  # whether track conveners can make a judgment (e.g. accept/reject)
+    'allow_convener_track_change': True,
```
ThiefMaster, line 132: should default to False IMO

indico/modules/events/abstracts/templates/abstract.html
```diff
@@ -32,7 +32,8 @@
     {% endcall %}
 
     <div class="i-timeline">
-        {% if abstract.can_judge(session.user, check_state=true) and abstract.event.tracks %}
+        {% if abstract.event.cfa.allow_convener_track_change and abstract.can_convene(session.user, check_state=True)
+            and abstract.event.tracks %}
```
ThiefMaster, line 36: i'd just keep it in one line (we don't have a hard line length limit in templates) or add parentheses..

indico/modules/events/abstracts/settings.py
plourenco: If I default it to false, I'm afraid current judges won't be able to edit the tracklist as before. Since now the two options (judging or changing tracks) are independent.

ThiefMaster: but judges/managers should always be able to do that regardless of this setting

ThiefMaster: so the current logic change is wrong - maybe it should be a new `abstract.can_track_change()` method which then checks for can_judge or (if the setting is enabled) can_convene?

plourenco: That is another option, even though we would have to change the convener_track_change option to be dependent on whether or not the previous (judgment) is selected.

ThiefMaster: I'd just hide it when convener judgment is enabled; `HiddenUnless` should do the job

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -278,7 +278,7 @@ class RHDisplayAbstractsExportExcel(AbstractsExportExcel, RHDisplayAbstractsActi
 
 class RHEditReviewedForTrackList(RHAbstractBase):
     def _check_abstract_protection(self):
-        return self.abstract.can_judge(session.user, check_state=True)
+        return self.event.cfa.allow_convener_judgment and self.abstract.can_convene(session.user, check_state=True)
```
ThiefMaster, line 281: Don't you want to use the new `can_change_tracks` method here? :D

indico/modules/events/abstracts/templates/abstract.html
```diff
@@ -32,7 +32,7 @@
     {% endcall %}
 
     <div class="i-timeline">
-        {% if abstract.can_judge(session.user, check_state=true) and abstract.event.tracks %}
+        {% if abstract.can_change_tracks(session.user, check_state=True) and abstract.event.tracks %}
```
ThiefMaster, line 35:
```suggestion
        {% if abstract.can_change_tracks(session.user, check_state=true) and abstract.event.tracks %}
```

indico/modules/events/abstracts/models/abstracts.py
```diff
@@ -520,18 +520,17 @@ def can_comment(self, user, check_state=False):
             return True
         return self.can_judge(user) or self.can_convene(user) or self.can_review(user)
 
-    def can_convene(self, user):
+    def can_convene(self, user, check_state=False):
```
ThiefMaster, line 523: With the suggestion above we no longer use `check_state=True` - and when having a quick look I didn't spot other places where we care about the state, so the change here could be reverted.

Thinking about this again, for judging/reviewing check_state makes sense,  but "convening" (is that even a word? :D) isn't really a "modifying" action so checking the state for it doesn't make that much sense in the end...

ThiefMaster (changes requested):

indico/modules/events/abstracts/controllers/reviewing.py
plourenco: 🙈

indico/modules/events/abstracts/models/abstracts.py
plourenco: Agree, probably not relevant anymore

ThiefMaster (approved):

PR #4996, opened 2021-07-07
Fix: Automatic notifications to consider notification sender address

### LINKED ISSUE #4992: RegistrationForm `notification_sender_address` settings
**Describe the bug**
The `notification_sender_address` field description is ambiguous in regards to the `RegistrationForm.sender_address`.

**To Reproduce**
1. Create a new event, enter the contact email in settings.
2. Create a new registration form, keeping the notification sender address empty.
3. Do some actions which trigger an automated notification email.
4. The email received has a from address from the event's contact email not the `config.no_reply` email.

**Screenshots**
## Event settings `contact email`
image

## Registration form settings `notification_sender_address`
image

## Email notification
image

**Additional context**
We fixed it in the UN plugin by just overriding the property and returning `self.notification_sender_address`, if it is `None`, the `make_email` automatically takes `config.no_reply`.
The UN doesn't want to use the contact emails for the automated notification emails.

vasantvohra:
closes #4992 

- Let me know if the commit message needs to be changed.
- If there's a need to add `changes.rst`
- As part of this minor PR,
- I upgrade my local to v3 🚀 
image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,6 +39,9 @@ Bugfixes
 - Fix errors when importing events containing abstracts or event roles from a YAML dump
   (:pr:`4995`)
 - Fix sorting abstract notification rules (:pr:`4998`)
+- No longer silently fall back to the first event contact email address when sending
+  registration emails where no explicit sender address has been configured (:issue:`4992`,
+  :pr:`4996`, thanks :user:`vasantvohra`)
 
 Version 3.0rc1
 --------------
--- a/indico/modules/events/registration/models/forms.py
+++ b/indico/modules/events/registration/models/forms.py
@@ -394,11 +394,6 @@ class RegistrationForm(db.Model):
                 .options(subqueryload('data'))
                 .all())
 
-    @property
-    def sender_address(self):
-        contact_email = self.event.contact_emails[0] if self.event.contact_emails else None
-        return self.notification_sender_address or contact_email
-
     def __repr__(self):
         return f'<RegistrationForm({self.id}, {self.event_id}, {self.title})>'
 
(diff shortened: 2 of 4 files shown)


ThiefMaster:
Maybe removing this logic from `build_default_email_template` (abstract email notifications) would be a good idea as well?

vasantvohra:
I see reply-to address there, so should I replace 
```python
event.contact_emails[0] if event.contact_emails else ''
```
with 
```python
config.NO_REPLY_EMAIL
```
?

ThiefMaster:
no, just set it to `''` all the time; that should result in indico not setting a reply-to address at all (noreply is already uses in From)

indico/modules/events/registration/models/forms.py
```diff
@@ -396,8 +396,7 @@ def active_registrations(self):
 
     @property
     def sender_address(self):
-        contact_email = self.event.contact_emails[0] if self.event.contact_emails else None
-        return self.notification_sender_address or contact_email
+        return self.notification_sender_address
```
ThiefMaster, line 399: The property is used in exactly one place - let's get rid of it and use `notification_sender_address` there directly.

ThiefMaster:
A changelog entry would be useful for this (bugfixes section i guess, that's the most suitable one IMO):

```rst
- No longer silently fall back to the first event contact email address when sending
  registration emails where no explicit sender address has been configured (:issue:`4992`,
  :pr:`4996`, thanks :user:`vasantvohra`)
```

CHANGES.rst
```diff
@@ -36,7 +36,9 @@ Improvements
 Bugfixes
 ^^^^^^^^
 
-- None so far :)
```
ThiefMaster, line 39: this needs a rebase, the changelog in master already contains a bugfix

vasantvohra: Done

ThiefMaster (approved):

PR #4999, opened 2021-07-07
Fix constrained i-dropdown height

plourenco:
Removes the fixed height on an `i-dropdown > li` since it incorrectly places most of the lists that do not fit in a single line.
I couldn't find any occurrences where this was useful, as the current dropdowns seem to play fine with it.

imagem

imagem


ThiefMaster (approved):
I can't find any places where this causes problems either

PR #5000, opened 2021-07-08
Make avatar colors depend on user ID

pferreir:
Also added 6 new colors to increase variety.

image

### DIFF
--- a/indico/modules/users/util.py
+++ b/indico/modules/users/util.py
@@ -7,6 +7,7 @@
 
 import hashlib
 import os
+import typing as t
 from datetime import datetime
 from io import BytesIO
 from operator import itemgetter
@@ -44,7 +45,7 @@ from indico.web.flask.util import send_file, url_for
 # colors for user-specific avatar bubbles
 user_colors = ['#e06055', '#ff8a65', '#e91e63', '#f06292', '#673ab7', '#ba68c8', '#7986cb', '#3f51b5', '#5e97f6',
                '#00a4e4', '#4dd0e1', '#0097a7', '#d4e157', '#aed581', '#57bb8a', '#4db6ac', '#607d8b', '#795548',
-               '#a1887f', '#fdd835', '#a3a3a3']
+               '#a1887f', '#fdd835', '#a3a3a3', '#556c60', '#605264', '#923035', '#915a30', '#55526f', '#67635a']
 
 
 def get_admin_emails():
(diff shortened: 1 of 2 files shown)


indico/modules/users/util.py
```diff
@@ -379,8 +379,15 @@ def merge_users(source, target, force=False):
     logger.info('Successfully merged %s into %s', source, target)
 
 
-def get_color_for_username(username):
-    return user_colors[crc32(username) % len(user_colors)]
+def get_color_for_user_id(user_id):
+    """Calculate a unique color for a user based on their id.
+
+        :param user_id: the user ID (int), or otherwise a string (external search results)
```
ThiefMaster, line 385: Just wondering.. wouldn't it be simpler to just use `crc32(str(user_id))` whichw ould work for all kinds of ids?

indico/modules/users/util.py
```diff
@@ -421,10 +428,17 @@ def set_user_avatar(user, avatar, filename, lastmod=None):
...
-        color = get_color_for_username(name)
+def send_default_avatar(user):
+    """Send a user's default avatar as an SVG.
+
+       :param user: A `User` object or string (external search results, registrations)
+    """
+    if isinstance(user, str):
+        text = user[0].upper()
+        color = get_color_for_user_id(user)
+    elif user.full_name:
```
ThiefMaster, line 439: That will fail when called via `/assets/avatar/blank.svg` since user is None

indico/modules/users/util.py
```diff
@@ -421,10 +428,17 @@ def set_user_avatar(user, avatar, filename, lastmod=None):
...
 
 
-def send_default_avatar(name=None):
-    if name:
-        text = name[0].upper()
-        color = get_color_for_username(name)
+def send_default_avatar(user):
+    """Send a user's default avatar as an SVG.
+
+       :param user: A `User` object or string (external search results, registrations)
```
ThiefMaster, line 434:
```suggestion
    :param user: A `User` object or string (external search results, registrations)
```

indico/modules/users/util.py
pferreir: Oh. I didn't test that edge case. When does it happen?

indico/modules/users/util.py
pferreir: Yeah, I thought of it, but kind of felt dirty doing it since `user_id` could be anything and result in unexpected behavior. But it's not a bad solution either.

indico/modules/users/util.py
ThiefMaster: anonymous users in editing timelines for example (via `_get_anonymous_user` when it's configured to not reveal the name of editors to submitters)

possibly for a pending user with no first name as well (there's a moment where you have a pending user without a user id: f90b120a135)

indico/modules/users/util.py
pferreir: Sometimes I wish Python was a strongly-typed language 🙊

ThiefMaster: you could use `def get_color_for_user_id(user_id: t.Union[int, str])` ;)
still not really strongly-typed of course, but sphinx uses this information when autodocumenting a function for example

ThiefMaster (approved):

PR #5001, opened 2021-07-08
Add signal to let plugins register additional fonts for designer templates.

vasantvohra:
Under discussion...

### DIFF
--- a/indico/legacy/pdfinterface/base.py
+++ b/indico/legacy/pdfinterface/base.py
@@ -15,6 +15,7 @@ from io import BytesIO
 
 import pkg_resources
 from PIL import Image as PILImage
+from indico.core import signals
 from reportlab import platypus
 from reportlab.lib.enums import TA_CENTER
 from reportlab.lib.fonts import addMapping
@@ -134,6 +135,7 @@ def setTTFonts():
         addMapping('LinuxLibertine', 1, 1, 'LinuxLibertine-Bold-Italic')
         pdfmetrics.registerFont(TTFont('Kochi-Mincho', os.path.join(font_dir, 'kochi-mincho-subst.ttf')))
         pdfmetrics.registerFont(TTFont('Kochi-Gothic', os.path.join(font_dir, 'kochi-gothic-subst.ttf')))
+        signals.legacy.register_additional_fonts.send()
         alreadyRegistered = True
 
 
--- a/indico/modules/designer/templates/template.html
+++ b/indico/modules/designer/templates/template.html
@@ -176,15 +176,14 @@
                     <div class="subtool">
                         <select id='font-selector' class="js-font-tool" data-attr="font">
                             <optgroup label="{% trans %}Normal Fonts{% endtrans %}">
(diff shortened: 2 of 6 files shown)


indico/modules/designer/templates/template.html
```diff
@@ -305,7 +304,7 @@
                 <div class="flexrow f-a-baseline">
                     <label for="element-list">{% trans %}Insert element:{% endtrans %}</label>
                     <select id="element-list">
-                        {% for group_id, group in placeholders.items() | sort(attribute='1.title') %}
+                        {% for group_id, group in placeholders.viewitems() | sort(attribute='1.title') %}
```
ThiefMaster, line 307: That's Python-2-only...

indico/legacy/pdfinterface/base.py
```diff
@@ -134,6 +135,7 @@ def setTTFonts():
         addMapping('LinuxLibertine', 1, 1, 'LinuxLibertine-Bold-Italic')
         pdfmetrics.registerFont(TTFont('Kochi-Mincho', os.path.join(font_dir, 'kochi-mincho-subst.ttf')))
         pdfmetrics.registerFont(TTFont('Kochi-Gothic', os.path.join(font_dir, 'kochi-gothic-subst.ttf')))
+        signals.legacy.register_additional_fonts.send()
```
ThiefMaster, line 138: I'd rather have the signal return the font specs (name and path?) and do the registration here...

indico/modules/designer/templates/template.html
vasantvohra: Yes I was in the process of reverting those lines

vasantvohra: What about two separate dictionaries? is it ok?

indico/legacy/pdfinterface/base.py
vasantvohra: Okay, but then how to go ahead with `addMapping`?

vasantvohra: @ThiefMaster,

-  I see `pdfmetrics.registerFont` calls `registerFontFamily(font.fontName)` so there's no need to explicitly do `addMapping` `from reportlab.lib.fonts import addMapping` ?

- Also, to return a name and path, do I just return a dictionary {font_name: path} from the function connected to the signal?

```python
plugin.connect(signals.legacy.register_additional_fonts, setAdditionalTTFonts)

def setAdditionalTTFonts(self):
    font_dir = os.path.join('../un/', 'indico_fonts') 
    return {'Bolt': os.path.join(font_dir, 'Bolt-Regular.ttf'), 'Bolt-Bold': os.path.join(font_dir, 'Bolt-Bold.ttf')}
```

ThiefMaster: tbh, this is legacy code, I don't think anyone in the team remembers the details why these things were done that way :)

ThiefMaster:
closing in favor of #5003 to avoid adding new signals for legacy code

PR #5002, opened 2021-07-08
Add CLI to update to the latest Python version and update the virtualenv in-place

ThiefMaster:
This is similar to the script we already use in production to avoid reinstalling all the packages after every minor Python version update (e.g. 3.9.5 to 3.9.6).

I also added a switch to force a version upgrade, so if a future Indico version requires 3.10, we can add a step to that version's upgrade asking people to run the script once to make sure they have the new Python version (even though this type of major upgrade requires a full recreation of the venv, and thus reinstalling of the indico-related packages)

### DIFF
--- a/indico/cli/python_upgrader.py
+++ b/indico/cli/python_upgrader.py
@@ -1,0 +1,140 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2021 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+# XXX: This script will be executed standalone, outside the indico virtualenv.
+# It must not use anything that's not part of the standard library of Python
+# 3.9.6 (since that's the lowest version from which it may get executed).
+
+import itertools
+import os
+import subprocess
+import sys
+from pathlib import Path
+from venv import EnvBuilder
+
+
+class PythonUpgrader:
+    def __init__(self, venv_path: Path):
+        self.venv_path = venv_path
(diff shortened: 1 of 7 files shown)


indico/cli/python_upgrader.py
```diff
@@ -0,0 +1,140 @@
...
+            if config['version'] == pyenv_local_python_version:
+                print('Venv is up to date; not doing anything')
+                sys.exit(0)
+            if tuple(map(int, config['version'].split('.')[:2])) != sys.version_info[:2]:
+                print('Venv is using a different Python 3.x version; recreating')
+                clear = True
+                upgrade = False
+            else:
+                clear = False
+                upgrade = True
```
plourenco, line 94:
```suggestion
        config_file = self.venv_path / 'pyvenv.cfg'
        upgrade, clear = False, True
        if not self.venv_path.exists() or not config_file.exists():
            print('Venv not found, creating new one')
        else:
            config = self._parse_venv_config(config_file)
            print(f'Venv found; python version {config["version"]} from {config["home"]}')
            if config['version'] == pyenv_local_python_version:
                print('Venv is up to date; not doing anything')
                sys.exit(0)
            if tuple(map(int, config['version'].split('.')[:2])) == sys.version_info[:2]:
                clear = False
                upgrade = True
            else:
                print('Venv is using a different Python 3.x version; recreating')
```

ThiefMaster: I prefer being explicit here - that way, if a branch isn't handled, it will fail rather than destroy (clear=True) the venv

PR #5004, opened 2021-07-09
Add system notices

ThiefMaster:
- option to opt-out during setup
- mention as a "major feature" in the release notes since we do not have such a prompt during an upgrade from an older version

image

### DIFF
--- a/indico/modules/admin/notices.py
+++ b/indico/modules/admin/notices.py
@@ -1,0 +1,159 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2021 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+import dataclasses
+import operator
+import platform
+import re
+import typing as t
+from enum import Enum
+
+import requests
+import yaml
+from celery.schedules import crontab
+from flask import session
+from markupsafe import Markup
+from marshmallow import EXCLUDE, ValidationError, fields, post_load
+from marshmallow_dataclass import dataclass
+from packaging.version import InvalidVersion, Version
+
+import indico
+from indico.core.cache import make_scoped_cache
+from indico.core.celery import celery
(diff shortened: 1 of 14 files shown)


indico/modules/admin/notices.py
```diff
@@ -0,0 +1,159 @@
...
+            '>': operator.gt,
+            '>=': operator.ge,
+        }
+        for key, value in dataclasses.asdict(self.when).items():
+            if value is None:
+                continue
+            if not (match := re.match(fr'^(?P<op>{"|".join(map(re.escape, ops))})\s*(?P<val>[0-9]\S*)$', value)):
+                logger.warning('Invalid criterion in %s: %s %s', self.id, key, value)
+                return False
+            current_version = versions[key]
```
tomasr8, line 80: Can it happen that there is a criteria in `SystemNoticeCriteria` which isn't in the yaml config?

indico/modules/admin/notices_test.py
```diff
@@ -0,0 +1,44 @@
...
+# LICENSE file for more details.
+
+import pytest
+from packaging.version import Version
+
+from indico.modules.admin.notices import NoticeSeverity, SystemNotice, SystemNoticeCriteria
+from indico.testing.util import extract_logs
+
+
+@pytest.mark.parametrize(('criteria', 'expected'), (
```
tomasr8, line 15: Thoughts on adding test cases which test invalid criteria, invalid versions and invalid operators?

ThiefMaster: I actually did that a few minutes ago, just not pushed ;)

indico/modules/admin/notices.py
ThiefMaster: Yes, that happens all the time, because only criteria that are actually used are in the yaml file.
But `versions` is the dict from `_get_versions` and the dataclass cannot contain keys that aren't there unless someone adds a new field to it, but not return a value for it from that function.

PR #5013, opened 2021-07-15
Add category moderation through event move requests

### LINKED ISSUE #2057: Moderated event creation
In some categories it would be useful if users could create events that need to be approved by a category manager before they are actually created.
- Test category (non-CERN users in our case) - letting everyone create events there invites spam and trolling, but there are some valid use cases, e.g. people who have a conference that's organized via our Indico system and need a test event to test certain features.
- High-profile categories (e.g. "conferences") - people end up creating test events in such categories even though they don't belong in there. Since such categories are likely to be displayed in a prominent way (upcoming events, external sites, etc) only "proper" events should show up there.

plourenco:
An embryonic version of the event moderation is described in #2057. Still working to address:

- [x] Withdraw a request if an event is forcibly moved to the target category
- [x] ACL permission to allow requesting to publish an event
- [x] Allow reasons in moderation judgments #5025 
- [x] Enhance event logs #5025 
- [ ] Allow comments while opening a new request

The current proposal flow is the following:

- User can create events in the target category? The event is immediately moved
- User has permission to propose event moves? A proposal to move is created
- Target category setting requires approval for moves? A proposal to move is created
- Otherwise normal conditions apply

imagem
imagem

closes #2057

### DIFF
--- a/indico/migrations/versions/20210715_1211_9d00917b2fa8_add_event_category_request.py
+++ b/indico/migrations/versions/20210715_1211_9d00917b2fa8_add_event_category_request.py
@@ -1,0 +1,59 @@
+"""Add event category request
+
+Revision ID: 9d00917b2fa8
+Revises: 1cec32e42f65
+Create Date: 2021-07-15 12:11:28.157824
+"""
+from enum import Enum
+
+import sqlalchemy as sa
+from alembic import op
+
+from indico.core.db.sqlalchemy import PyIntEnum, UTCDateTime
+
+
+# revision identifiers, used by Alembic.
+revision = '9d00917b2fa8'
+down_revision = '1cec32e42f65'
+branch_labels = None
+depends_on = None
+
+
+class _MoveRequestState(int, Enum):
+    pending = 0
+    accepted = 1
+    rejected = 2
+    withdrawn = 3
+
+
+def upgrade():
+    op.create_table(
+        'event_move_requests',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('event_id', sa.Integer(), nullable=False, index=True),
+        sa.Column('category_id', sa.Integer(), nullable=True, index=True),
+        sa.Column('submitter_id', sa.Integer(), nullable=True, index=True),
+        sa.Column('state', PyIntEnum(_MoveRequestState), nullable=False),
+        sa.Column('moderator_id', sa.Integer(), nullable=True),
+        sa.Column('submitted_dt', UTCDateTime, nullable=False),
+        sa.CheckConstraint('state in (1, 2) AND moderator_id IS NOT NULL OR moderator_id IS NULL',
+                           name=op.f('moderator_state')),
+        sa.ForeignKeyConstraint(['category_id'], ['categories.categories.id']),
+        sa.ForeignKeyConstraint(['event_id'], ['events.events.id']),
+        sa.ForeignKeyConstraint(['moderator_id'], ['users.users.id']),
+        sa.ForeignKeyConstraint(['submitter_id'], ['users.users.id']),
+        sa.PrimaryKeyConstraint('id'),
+        schema='categories'
+    )
+    op.create_index(op.f('ix_uq_event_move_requests_event_id'), 'event_move_requests', ['event_id'], unique=True,
+                    schema='categories', postgresql_where=sa.text('state = 0'))
+    op.add_column('categories',
+                  sa.Column('event_requires_approval', sa.Boolean(), nullable=False, server_default='false'),
+                  schema='categories')
+
+
+def downgrade():
+    op.drop_column('categories', 'event_requires_approval', schema='categories')
+    op.drop_index(op.f('ix_uq_event_move_requests_event_id'), table_name='event_move_requests', schema='categories',
+                  postgresql_where=sa.text('state = 0'))
+    op.drop_table('event_move_requests', schema='categories')
--- a/indico/migrations/versions/20210715_1211_9d00917b2fa8_add_event_category_request.py
+++ b/indico/migrations/versions/20210715_1211_9d00917b2fa8_add_event_category_request.py
@@ -34,6 +34,7 @@ def upgrade():
         sa.Column('category_id', sa.Integer(), nullable=True, index=True),
         sa.Column('submitter_id', sa.Integer(), nullable=True, index=True),
         sa.Column('state', PyIntEnum(_MoveRequestState), nullable=False),
+        sa.Column('state_reason', sa.String(), nullable=True),
         sa.Column('moderator_id', sa.Integer(), nullable=True),
         sa.Column('submitted_dt', UTCDateTime, nullable=False),
         sa.CheckConstraint('state in (1, 2) AND moderator_id IS NOT NULL OR moderator_id IS NULL',
--- a/indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
+++ b/indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
@@ -26,6 +26,12 @@ class _MoveRequestState(int, Enum):
     withdrawn = 3
 
 
+class _EventCreationMode(int, Enum):
+    restricted = 1
+    moderated = 2
+    open = 3
+
+
 def upgrade():
     op.create_table(
         'event_move_requests',
@@ -48,12 +54,20 @@ def upgrade():
     )
     op.create_index(None, 'event_move_requests', ['event_id'], unique=True,
                     schema='categories', postgresql_where=sa.text('state = 0'))
+
     op.add_column('categories',
-                  sa.Column('event_requires_approval', sa.Boolean(), nullable=False, server_default='false'),
+                  sa.Column('event_creation_mode', PyIntEnum(_EventCreationMode), nullable=False, server_default='1'),
                   schema='categories')
-    op.alter_column('categories', 'event_requires_approval', server_default=None, schema='categories')
+    op.alter_column('categories', 'event_creation_mode', server_default=None, schema='categories')
+    op.execute('UPDATE categories.categories SET event_creation_mode = 3 WHERE NOT event_creation_restricted')
+    op.drop_column('categories', 'event_creation_restricted', schema='categories')
 
 
 def downgrade():
-    op.drop_column('categories', 'event_requires_approval', schema='categories')
+    op.add_column('categories',
+                  sa.Column('event_creation_restricted', sa.Boolean(), nullable=False, server_default='true'),
+                  schema='categories')
+    op.alter_column('categories', 'event_creation_restricted', server_default=None, schema='categories')
+    op.execute('UPDATE categories.categories SET event_creation_restricted = false WHERE event_creation_mode = 3')
+    op.drop_column('categories', 'event_creation_mode', schema='categories')
     op.drop_table('event_move_requests', schema='categories')
--- a/indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
+++ b/indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
@@ -1,7 +1,7 @@
 """Add event category request
 
 Revision ID: 9d00917b2fa8
-Revises: 1cec32e42f65
+Revises: 4b097412a8d9
 Create Date: 2021-07-15 12:11:28.157824
 """
 from enum import Enum
@@ -14,7 +14,7 @@ from indico.core.db.sqlalchemy import PyIntEnum, UTCDateTime
 
 # revision identifiers, used by Alembic.
 revision = '9d00917b2fa8'
-down_revision = '1cec32e42f65'
+down_revision = '4b097412a8d9'
 branch_labels = None
 depends_on = None
 
--- a/indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
+++ b/indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
@@ -31,30 +31,29 @@ def upgrade():
         'event_move_requests',
         sa.Column('id', sa.Integer(), nullable=False),
         sa.Column('event_id', sa.Integer(), nullable=False, index=True),
-        sa.Column('category_id', sa.Integer(), nullable=True, index=True),
-        sa.Column('submitter_id', sa.Integer(), nullable=True, index=True),
+        sa.Column('category_id', sa.Integer(), nullable=False, index=True),
+        sa.Column('requestor_id', sa.Integer(), nullable=False, index=True),
         sa.Column('state', PyIntEnum(_MoveRequestState), nullable=False),
         sa.Column('moderator_comment', sa.String(), nullable=False, default=''),
         sa.Column('moderator_id', sa.Integer(), nullable=True),
-        sa.Column('submitted_dt', UTCDateTime, nullable=False),
-        sa.CheckConstraint('state in (1, 2) AND moderator_id IS NOT NULL OR moderator_id IS NULL',
-                           name=op.f('moderator_state')),
+        sa.Column('requested_dt', UTCDateTime, nullable=False),
+        sa.CheckConstraint('(state in (1, 2) AND moderator_id IS NOT NULL) OR moderator_id IS NULL',
+                           name='moderator_state'),
         sa.ForeignKeyConstraint(['category_id'], ['categories.categories.id']),
         sa.ForeignKeyConstraint(['event_id'], ['events.events.id']),
         sa.ForeignKeyConstraint(['moderator_id'], ['users.users.id']),
-        sa.ForeignKeyConstraint(['submitter_id'], ['users.users.id']),
+        sa.ForeignKeyConstraint(['requestor_id'], ['users.users.id']),
         sa.PrimaryKeyConstraint('id'),
         schema='categories'
     )
-    op.create_index(op.f('ix_uq_event_move_requests_event_id'), 'event_move_requests', ['event_id'], unique=True,
+    op.create_index(None, 'event_move_requests', ['event_id'], unique=True,
                     schema='categories', postgresql_where=sa.text('state = 0'))
     op.add_column('categories',
                   sa.Column('event_requires_approval', sa.Boolean(), nullable=False, server_default='false'),
                   schema='categories')
+    op.alter_column('categories', 'event_requires_approval', server_default=None, schema='categories')
 
 
 def downgrade():
     op.drop_column('categories', 'event_requires_approval', schema='categories')
-    op.drop_index(op.f('ix_uq_event_move_requests_event_id'), table_name='event_move_requests', schema='categories',
-                  postgresql_where=sa.text('state = 0'))
     op.drop_table('event_move_requests', schema='categories')
--- a/indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
+++ b/indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
@@ -34,7 +34,7 @@ def upgrade():
         sa.Column('category_id', sa.Integer(), nullable=True, index=True),
         sa.Column('submitter_id', sa.Integer(), nullable=True, index=True),
         sa.Column('state', PyIntEnum(_MoveRequestState), nullable=False),
-        sa.Column('state_reason', sa.String(), nullable=True),
+        sa.Column('moderator_comment', sa.String(), nullable=False, default=''),
         sa.Column('moderator_id', sa.Integer(), nullable=True),
         sa.Column('submitted_dt', UTCDateTime, nullable=False),
         sa.CheckConstraint('state in (1, 2) AND moderator_id IS NOT NULL OR moderator_id IS NULL',
--- a/indico/modules/categories/__init__.py
+++ b/indico/modules/categories/__init__.py
@@ -53,6 +53,8 @@ def _sidemenu_items(sender, category, **kwargs):
                        50, icon='users')
     yield SideMenuItem('logs', _('Logs'), url_for('logs.category', category),
                        0, icon='stack')
+    yield SideMenuItem('moderation', _('Moderation'), url_for('categories.manage_moderation', category),
+                       50, icon='users')
 
 
 @signals.menu.items.connect_via('admin-sidemenu')
--- a/indico/modules/categories/__init__.py
+++ b/indico/modules/categories/__init__.py
@@ -43,8 +43,9 @@ def _merge_users(target, source, **kwargs):
 
 
 def _is_moderation_visible(category):
-    return category.event_requires_approval and category.event_move_requests.filter(
-        EventMoveRequest.state == MoveRequestState.pending)
+    return category.event_requires_approval or category.event_move_requests.filter(
+        EventMoveRequest.state == MoveRequestState.pending).first() or any(
+        ['event_move_request' in e.permissions for e in category.acl_entries])
 
 
 @signals.menu.items.connect_via('category-management-sidemenu')
--- a/indico/modules/categories/__init__.py
+++ b/indico/modules/categories/__init__.py
@@ -12,7 +12,7 @@ from indico.core.db.sqlalchemy.protection import make_acl_log_fn
 from indico.core.logger import Logger
 from indico.core.permissions import ManagementPermission, check_permissions
 from indico.core.settings import SettingsProxy
-from indico.modules.categories.models.categories import Category
+from indico.modules.categories.models.categories import Category, EventCreationMode
 from indico.modules.categories.models.event_move_request import MoveRequestState
 from indico.modules.logs.models.entries import CategoryLogRealm
 from indico.util.i18n import _
@@ -44,7 +44,7 @@ def _merge_users(target, source, **kwargs):
 
 def _is_moderation_visible(category):
     return (
-        category.event_requires_approval or
+        category.event_creation_mode == EventCreationMode.moderated or
         category.event_move_requests.filter_by(state=MoveRequestState.pending).has_rows() or
         any('event_move_request' in entry.permissions for entry in category.acl_entries)
     )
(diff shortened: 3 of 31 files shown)


indico/modules/categories/client/js/components/CategoryModeration.jsx
```diff
@@ -0,0 +1,131 @@
...
+      return set;
+    });
+
+  if (requests.length === 0) {
+    return <Translate as="p">There are no event move requests</Translate>;
+  }
+
+  return (
+    <>
+      {/* TODO: check if module is enabled */}
```
ThiefMaster, line 46: Probably not needed; for many other things we don't do it either since accessing the page via a direct link is harmless anyway as there's nothing useful to do on it.

indico/modules/categories/client/js/components/CategoryModeration.jsx
```diff
@@ -0,0 +1,131 @@
...
+      <Table celled>
+        <Table.Header>
+          <Table.Row>
+            <Table.HeaderCell />
+            <Table.HeaderCell>User</Table.HeaderCell>
+            <Table.HeaderCell>Event</Table.HeaderCell>
+            <Table.HeaderCell>State</Table.HeaderCell>
+            <Table.HeaderCell>Submitted Date</Table.HeaderCell>
```
ThiefMaster, line 54: i18n

indico/modules/categories/controllers/management.py
```diff
@@ -117,6 +122,36 @@ def _process(self):
                                                     icon_form=icon_form, logo_form=logo_form)
 
 
+class RHAPIEventMoveRequests(RHManageCategoryBase):
+    def _process_args(self):
+        RHManageCategoryBase._process_args(self)
+        if request.method != 'GET':
+            self.move_requests = parser.parse({
+                'requests': EventRequestList(required=True, category=self.category, validate=not_empty)
+            }, unknown=EXCLUDE)['requests']
```
ThiefMaster, line 131: I'd just move it into `_process_POST`.. the main benefit of `_process_args` is to have something that runs for all methods and that can be used easily w/ inheritance. But that's nothing we need here..

indico/modules/categories/schemas.py
```diff
@@ -0,0 +1,34 @@
...
+        model = Event
+        fields = ('title', 'id')
+
+
+class CategoryUserSchema(mm.SQLAlchemyAutoSchema):
+    class Meta:
+        model = User
+        fields = ('id', 'name', 'first_name', 'last_name', 'title', 'affiliation', 'avatar_url')
+
+    name = fields.Function(lambda p: p.get_full_name(abbrev_first_name=False))
```
ThiefMaster, line 25: Let's use `full_name` here to be consistent with various other schemas

indico/modules/events/__init__.py
```diff
@@ -245,3 +245,9 @@ def _event_cloned(old_event, new_event, **kwargs):
     new_event.contact_title = old_event.contact_title
     new_event.contact_emails = old_event.contact_emails
     new_event.contact_phones = old_event.contact_phones
+
+
+@signals.event.moved.connect
+def _event_request_moved(event, old_parent):
+    if event.pending_move_request and event.pending_move_request.category == event.category:
```
ThiefMaster, line 252: This queries the request twice since it's a normal property...

Either make it a method so it's clear that it should be assigned and not just used many times, or maybe turn it into a relationship with a custom primaryjoin criterion that includes the state and has `viewonly=True`.

indico/modules/events/operations.py
```diff
@@ -383,3 +384,14 @@ def sort_reviewing_questions(questions, new_positions):
         field.position = index
     db.session.flush()
     logger.info('Reviewing questions of %r reordered by %r', questions[0].event, session.user)
+
+
+def create_event_request(event, category):
+    assert event.id != category.id
```
ThiefMaster, line 390: Did you mean to use `event.category_id` here? (I'd prefer using `event.category != category` though)

indico/modules/events/operations.py
```diff
@@ -383,3 +384,14 @@ def sort_reviewing_questions(questions, new_positions):
...
+    rq = EventMoveRequest(event=event, category=category, submitter=session.user)
+    db.session.add(rq)
```
ThiefMaster, line 392: not needed, the link to the event/category automatically adds it to the session

indico/modules/categories/controllers/management.py
```diff
@@ -117,6 +122,33 @@ def _process(self):
...
 
 
+class RHAPIEventMoveRequests(RHManageCategoryBase):
+    def _process_GET(self):
+        return EventMoveRequestSchema(many=True).jsonify(self.category.event_move_requests)
+
+    @use_kwargs({
+        'state': EnumField(MoveRequestState, required=True)
+    })
+    def _process_POST(self, state):
```
ThiefMaster, line 132: Let's reurn something - event if it's just `return '', 204` for a *no content* response

indico/modules/categories/controllers/management.py
```diff
@@ -117,6 +122,33 @@ def _process(self):
...
+        move_requests = parser.parse({
+            'requests': EventRequestList(required=True, category=self.category, validate=not_empty)
+        }, unknown=EXCLUDE)['requests']
+
+        if state not in (MoveRequestState.accepted, MoveRequestState.rejected):
```
ThiefMaster, line 137: Maybe taking an `accept` bool arg or an `action` string would be better? Especially since the logic of actually approving/rejecting the request should most likely be in an operation function which then also does stuff like sending notification emails and adding something to the event log.

indico/modules/categories/controllers/management.py
```diff
@@ -117,6 +122,33 @@ def _process(self):
...
+            raise Forbidden
```
ThiefMaster, line 138: (if you end up keeping this: `BadRequest` or `UnprocessableEntity` are probably better - or just do it via `validate=validate_with_message(...)` in the webargs field)

indico/modules/categories/models/categories.py
```diff
@@ -181,6 +181,11 @@ def __table_args__(cls):
         nullable=True,
         index=True
     )
+    event_requires_approval = db.Column(
```
ThiefMaster, line 184: `events_require_approval` makes more sense; but based on what we discussed before: Maybe a different name would be better? Because in the end this option will allow anyone to propose a move to the category (regardless of special permissions via the ACI)...

indico/modules/categories/models/categories.py
```diff
@@ -291,6 +297,11 @@ def display_tzinfo(self):
         """The tzinfo of the category or the one specified by the user."""
         return get_display_tz(self, as_timezone=True)
 
+    def can_propose_events(self, user):
+        """Check whether the user can propose move requests to the category."""
+        return self.event_requires_approval or (
+            user and self.can_manage(user, permission='event_move_request')) or self.can_create_events(user)
```
ThiefMaster, line 303: Shouldn't it return false if;

- `user` is none or
- the user can create events (i guess we'll never show the approval flow in that case anyway?)

indico/modules/categories/models/event_move_request.py
```diff
@@ -0,0 +1,114 @@
...
+                                         f'AND moderator_id IS NOT NULL OR moderator_id IS NULL',
+                                         'moderator_state'),
+                      {'schema': 'categories'})
+
+    id = db.Column(
+        db.Integer,
+        primary_key=True
+    )
+    event_id = db.Column(
+        db.Integer,
```
ThiefMaster, line 39: you can omit this arg for all FK columns since SA can infer it from the FK's target

indico/modules/categories/models/event_move_request.py
```diff
@@ -0,0 +1,114 @@
...
+        backref=db.backref(
+            'event_move_requests',
+            cascade_backrefs=False,
+            lazy='dynamic'
+        )
+    )
+    moderator = db.relationship(
+        'User',
+        foreign_keys=moderator_id,
+        lazy=False
```
ThiefMaster, line 103: this relationship is missing a backref

indico/modules/categories/models/event_move_request.py
```diff
@@ -0,0 +1,114 @@
...
+            lazy='dynamic'
+        )
+    )
+    submitter = db.relationship(
+        'User',
+        lazy=True,
+        foreign_keys=submitter_id,
+        backref=db.backref(
+            'event_move_requests',
+            cascade_backrefs=False,
```
ThiefMaster, line 96: I don't think this is needed/useful here? We only use if for a few places in EventPersons so we can create an EventPerson linked to a user but do not cascade adding it to the session unless it gets explicitly added.

indico/modules/categories/schemas.py
```diff
@@ -0,0 +1,32 @@
...
+from indico.core.marshmallow import fields, mm
+from indico.modules.categories.models.event_move_request import EventMoveRequest
+from indico.modules.events import Event
+from indico.modules.users import User
+
+
+class CategoryEventSchema(mm.SQLAlchemyAutoSchema):
+    class Meta:
+        model = Event
+        fields = ('title', 'id')
```
ThiefMaster, line 17:
```suggestion
        fields = ('id', 'title')
```

indico/modules/categories/templates/management/category_protection.html
```diff
@@ -11,7 +11,7 @@
     {{ form_row(form.permissions, skip_label=true) }}
     {{ form_rows(form, skip=(form._event_creation_fields + ('permissions',)),
                  widget_attrs={'own_no_access_contact': {'placeholder': category.no_access_contact}}) }}
-    {% call form_fieldset(_('Event creation settings')) %}
+    {% call form_fieldset(_('Event settings')) %}
```
ThiefMaster, line 14: Aren't those still creation-related settings? Sure, technically moving an event isn't creating it, but in the end it's "created" in the category that way..

indico/modules/events/management/controllers/actions.py
```diff
@@ -82,11 +82,35 @@ class RHMoveEvent(RHManageEventBase):
     def _process_args(self):
         RHManageEventBase._process_args(self)
         self.target_category = Category.get_or_404(int(request.form['target_category_id']), is_deleted=False)
-        if not self.target_category.can_create_events(session.user):
-            raise Forbidden(_('You may only move events to categories where you are allowed to create events.'))
+
+    def _check_access(self):
+        if not self.target_category.can_propose_events(session.user):
+            raise Forbidden(_('You may only move events to categories where you are allowed to propose events.'))
```
ThiefMaster, line 88: That error sounds extremely confusing...

indico/modules/events/management/controllers/actions.py
```diff
@@ -82,11 +82,35 @@ class RHMoveEvent(RHManageEventBase):
...
+        if self.event.pending_move_request:
+            raise BadRequest(_('There is already a move request pending review.'))
+
+    def _process(self):
+        if self.target_category.can_create_events(session.user):
+            self.event.move(self.target_category)
+            flash(_('Event "{}" has been moved to category "{}"')
+                  .format(self.event.title, self.target_category.title), 'success')
+        else:
+            create_event_request(self.event, self.target_category)
```
ThiefMaster, line 98: IIRC we wanted to ask for a comment (which then gets shown to the moderators) when requesting the move

indico/modules/events/management/templates/_action_menu.html
```diff
@@ -72,7 +72,7 @@
 
 
 {% macro _render_move_event(event) %}
-    <button class="i-button borderless icon-transmission hide-if-locked"
+    <button class="i-button borderless icon-transmission hide-if-locked {{ 'disabled' if event.pending_move_request is not none }}"
```
ThiefMaster, line 75:
```suggestion
    <button class="i-button borderless icon-transmission hide-if-locked {{ 'disabled' if event.pending_move_request }}"
```

None is falsy ;)

indico/modules/events/management/templates/settings.html
```diff
@@ -6,6 +6,29 @@
...
 {% block content %}
     {{ template_hook('event-actions', event=event) }}
+    {%- if event.pending_move_request -%}
+        <div class="action-box highlight">
+            <div class="section">
+                <div class="icon icon-bookmark"></div>
+                <div class="text">
+                    <div class="label">{% trans %}Pending event moderation{% endtrans %}</div>
+                    {% trans %}
+                        There is an on-going request to move this event to another category.
```
ThiefMaster, line 16: "ongoing" and I'd show which category it is (ideally with a link to it)

indico/migrations/versions/20210715_1211_9d00917b2fa8_add_event_category_request.py
```diff
@@ -0,0 +1,64 @@
...
+Revises: 1cec32e42f65
+Create Date: 2021-07-15 12:11:28.157824
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+from indico.core.db.sqlalchemy import PyIntEnum, UTCDateTime
+# revision identifiers, used by Alembic.
+from indico.modules.categories.models.event_move_request import MoveRequestState
```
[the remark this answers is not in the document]
plourenco: Yes, I saw some other cases where the Enum was manually imported so I assumed this was something to avoid from the auto-generated revision. But that's a good point, I can restore the original.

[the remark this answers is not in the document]
ThiefMaster: It's actually custom code somewhere in `indico/core/db/` that generates the embedded enums ;)

indico/modules/events/operations.py
plourenco: Yes to both. I mistook the logic here.

indico/modules/categories/controllers/management.py
plourenco: Sounds good, I agree. In the beginning, this wasn't very clear. 
I'm assuming this logging would only apply to rejects since approving already shows the default move log, unless we want to show it differently?

indico/modules/categories/models/categories.py
plourenco: Yes, `event_requires_approval` allows anyone even though the user shouldn't be none. But as for the second option, I made this method a bit more generic or hierarchical as in "if you can create events you are also allowed to propose them". But it's a fair change.

indico/modules/events/management/controllers/actions.py
plourenco: I had the impression this would suit better in a separate task. What do you think?

ThiefMaster: fine for me

indico/modules/categories/models/event_move_request.py
plourenco: Is it worth having a `backref` here? Not sure we'll ever need to check upon proposals being moderated by someone.

ThiefMaster: e.g. when handling user merges; and generally it's convenient to have it available when doing things in indico shell

indico/modules/categories/models/event_move_request.py
plourenco: Yes, it doesn't seem needed.

indico/modules/events/management/controllers/actions.py
plourenco: How about?
`You may only move events to categories you are allowed to.`
Might seem confusing to make a distinction between allowed to create/allowed to request moves.

ThiefMaster: "You may not move events to this category"? I guess you usually do not get there anyway since it's disabled in the UI...

indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
```diff
@@ -0,0 +1,60 @@
...
+        sa.CheckConstraint('state in (1, 2) AND moderator_id IS NOT NULL OR moderator_id IS NULL',
+                           name=op.f('moderator_state')),
+        sa.ForeignKeyConstraint(['category_id'], ['categories.categories.id']),
+        sa.ForeignKeyConstraint(['event_id'], ['events.events.id']),
+        sa.ForeignKeyConstraint(['moderator_id'], ['users.users.id']),
+        sa.ForeignKeyConstraint(['submitter_id'], ['users.users.id']),
+        sa.PrimaryKeyConstraint('id'),
+        schema='categories'
+    )
+    op.create_index(op.f('ix_uq_event_move_requests_event_id'), 'event_move_requests', ['event_id'], unique=True,
```
ThiefMaster, line 49:
```suggestion
    op.create_index(None, 'event_move_requests', ['event_id'], unique=True,
```

the name is autogenerated

indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
```diff
@@ -0,0 +1,60 @@
...
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('event_id', sa.Integer(), nullable=False, index=True),
+        sa.Column('category_id', sa.Integer(), nullable=True, index=True),
+        sa.Column('submitter_id', sa.Integer(), nullable=True, index=True),
+        sa.Column('state', PyIntEnum(_MoveRequestState), nullable=False),
+        sa.Column('moderator_comment', sa.String(), nullable=False, default=''),
+        sa.Column('moderator_id', sa.Integer(), nullable=True),
+        sa.Column('submitted_dt', UTCDateTime, nullable=False),
+        sa.CheckConstraint('state in (1, 2) AND moderator_id IS NOT NULL OR moderator_id IS NULL',
+                           name=op.f('moderator_state')),
```
ThiefMaster, line 41: That `op.f()` here looks wrong. It prevents name autogeneration, but the constraint name should be autogenerated based on the custom name.

indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
```diff
@@ -0,0 +1,60 @@
...
+                    schema='categories', postgresql_where=sa.text('state = 0'))
+    op.add_column('categories',
+                  sa.Column('event_requires_approval', sa.Boolean(), nullable=False, server_default='false'),
+                  schema='categories')
+
+
+def downgrade():
+    op.drop_column('categories', 'event_requires_approval', schema='categories')
+    op.drop_index(op.f('ix_uq_event_move_requests_event_id'), table_name='event_move_requests', schema='categories',
+                  postgresql_where=sa.text('state = 0'))
```
ThiefMaster, line 59: Not needed, goes away when the table is dropped

indico/migrations/versions/20210804_1211_9d00917b2fa8_add_event_category_request.py
```diff
@@ -0,0 +1,60 @@
...
+        sa.ForeignKeyConstraint(['event_id'], ['events.events.id']),
+        sa.ForeignKeyConstraint(['moderator_id'], ['users.users.id']),
+        sa.ForeignKeyConstraint(['submitter_id'], ['users.users.id']),
+        sa.PrimaryKeyConstraint('id'),
+        schema='categories'
+    )
+    op.create_index(op.f('ix_uq_event_move_requests_event_id'), 'event_move_requests', ['event_id'], unique=True,
+                    schema='categories', postgresql_where=sa.text('state = 0'))
+    op.add_column('categories',
+                  sa.Column('event_requires_approval', sa.Boolean(), nullable=False, server_default='false'),
```
ThiefMaster, line 52: The `server_default` is only needed so the column can be created even with data in the DB; it's not part of the model, so it should be removed again in an `alter_column` right after creating it (alternatively: add it in the model as well)

indico/modules/categories/models/event_move_request.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class EventMoveRequest(db.Model):
+    """A request that represents a proposed event move."""
+
+    __tablename__ = 'event_move_requests'
+    __table_args__ = (db.Index(None, 'event_id', unique=True,
+                               postgresql_where=db.text(f'state = {MoveRequestState.pending}')),
+                      db.CheckConstraint(f'state in ({MoveRequestState.accepted}, {MoveRequestState.rejected}) '
+                                         f'AND moderator_id IS NOT NULL OR moderator_id IS NULL',
```
ThiefMaster, line 29: I really do not want AND and OR on the same level without parentheses... this just makes it much harder to see the logic (and honestly I never remember which one has priority).

ThiefMaster: Also, the logic here looks wrong: It unconditionally allows a NULL moderator_id, but this should only be the case in the *other* states.

So the check condition should be `(state in (...)) = (moderator_id IS NOT NULL)`

indico/modules/categories/models/event_move_request.py
```diff
@@ -0,0 +1,119 @@
...
+        primary_key=True
+    )
+    event_id = db.Column(
+        db.ForeignKey('events.events.id'),
+        nullable=False,
+        index=True
+    )
+    category_id = db.Column(
+        db.ForeignKey('categories.categories.id'),
+        nullable=True,
```
ThiefMaster, line 44: I don't think this need to be nullable

indico/modules/categories/models/event_move_request.py
```diff
@@ -0,0 +1,119 @@
...
+        index=True
+    )
+    submitter_id = db.Column(
+        db.ForeignKey('users.users.id'),
+        nullable=True,
```
ThiefMaster, line 49: same here

indico/modules/categories/models/event_move_request.py
```diff
@@ -0,0 +1,119 @@
...
+    category = db.relationship(
+        'Category',
+        lazy=True,
+        backref=db.backref(
+            'event_move_requests',
+            cascade='all, delete-orphan',
+            lazy='dynamic'
+        )
+    )
+    submitter = db.relationship(
```
ThiefMaster, line 90: Hm, submitter vs requestor. The latter would avoid ambiguity with the other uses of "submitter" in Indico.

indico/modules/categories/operations.py
```diff
@@ -97,3 +102,19 @@ def _log_category_update(category, changes):
...
+    request.moderator_comment = reason
+    request.moderator = session.user
+    if accept:
+        request.state = MoveRequestState.accepted
+        request.event.move(request.category)
+    else:
+        category = request.category
+        sep = ' \N{RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK} '
+        category.log(CategoryLogRealm.category, LogKind.negative, 'Category', 'Move request rejected', session.user,
+                     data={'Location': sep.join(category.chain_titles), 'Reason': reason})
```
ThiefMaster, line 119: I don't think the location is useful here - it's the same category. I would rather log the event title and ID.

It should probably be logged in the event log as well.

indico/modules/categories/operations.py
```diff
@@ -97,3 +102,19 @@ def _log_category_update(category, changes):
...
+    request.state = MoveRequestState.rejected
+    request.moderator_comment = reason
+    request.moderator = session.user
+    if accept:
+        request.state = MoveRequestState.accepted
+        request.event.move(request.category)
+    else:
+        category = request.category
+        sep = ' \N{RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK} '
+        category.log(CategoryLogRealm.category, LogKind.negative, 'Category', 'Move request rejected', session.user,
```
ThiefMaster, line 118: `CategoryLogRealm.events` and "Moderation" instead of "Category" for the module name feels more appropriate here

indico/modules/events/management/templates/settings.html
```diff
@@ -6,6 +6,31 @@
...
 {% block content %}
     {{ template_hook('event-actions', event=event) }}
+    {% set move_request = event.pending_move_request %}
+    {%- if move_request -%}
+        <div class="action-box highlight">
+            <div class="section">
+                <div class="icon icon-bookmark"></div>
+                <div class="text">
+                    <div class="label">{% trans %}Pending event moderation{% endtrans %}</div>
+                    {% trans url=move_request.category, category=move_request.category.title %}
```
ThiefMaster, line 16: That does not look like a URL ;)

```suggestion
                    {% trans url=move_request.category.url, category=move_request.category.title %}
```

indico/modules/events/operations.py
```diff
@@ -384,3 +385,13 @@ def sort_reviewing_questions(questions, new_positions):
...
     db.session.flush()
     logger.info('Reviewing questions of %r reordered by %r', questions[0].event, session.user)
+
+
+def create_event_request(event, category):
+    assert event.category != category
+    rq = EventMoveRequest(event=event, category=category, submitter=session.user)
+    db.session.flush()
+    logger.info('Category move request %s to %s created by %s', rq, category, session.user)
+    event.log(EventLogRealm.event, LogKind.change, 'Event', f'Move request to {category.title} created',
```
ThiefMaster, line 395: In this log entry it would be useful to include the category chain in the detailed data (like we do in Event.move())

indico/modules/categories/__init__.py
```diff
@@ -41,6 +42,12 @@ def _merge_users(target, source, **kwargs):
     CategoryPrincipal.merge_users(target, source, 'category')
 
 
+def _is_moderation_visible(category):
+    return category.event_requires_approval or category.event_move_requests.filter(
+        EventMoveRequest.state == MoveRequestState.pending).first() or any(
+        ['event_move_request' in e.permissions for e in category.acl_entries])
```
ThiefMaster, line 48:
```suggestion
    return (
        category.event_requires_approval or
        category.event_move_requests.filter_by(state=MoveRequestState.pending).has_rows() or
        any('event_move_request' in entry.permissions for entry in category.acl_entries)
    )
```

Much more readable IMO

indico/modules/categories/models/event_move_request.py
tomasr8: Added also in the migration file.

indico/modules/categories/operations.py
tomasr8: Added event log :+1:

indico/modules/events/operations.py
tomasr8: Added the chain and ID.

indico/modules/categories/models/event_move_request.py
tomasr8: Renamed `submitter` and `submitted_dt`

ThiefMaster (approved):

PR #5016, opened 2021-07-16
Add redis server start for macOS

vaclavstepan:
When development environment is being installed on macOS via brew, both postgresql and redis should be started. This add the respective line into the installation guide.


pferreir (approved):

pferreir:
Thanks!

Issue #4687, opened 2020-10-28
Unlisted events: Backend


ThiefMaster:
To enable support in backend for unlisted events, the first steps are:

- Support events with no category / category_id in the DB (nullable column).
- Make sure that everything still works fine in those events, and features that really depend on a category (and thus break) are disabled (can be tested by manually removing the category from an event via `indico shell`). test with a non-admin account where permission checks are not aborted early!

See the [comment](https://github.com/indico/indico/issues/4294#issuecomment-603878775) on #4294 for details.

PR #5017, opened 2021-07-21
Unlisted events

javfg:
This PR includes commits related to https://github.com/indico/indico/issues/4687 so far.

### Backend changes
* Fix all problems arising after making category nullable in the event object.

### DIFF
--- a/indico/modules/events/templates/display/indico/lecture.html
+++ b/indico/modules/events/templates/display/indico/lecture.html
@@ -14,11 +14,21 @@
     {% block header %}
         <div class="event-header event-header-lecture {%- if not has_subheader %} round-bottom-corners{% endif %}">
             {% if not g.static_site %}
-                <a class="lecture-category" href="{{ url_for('categories.display', event.category) }}">
-                    {{ category }}
-                </a>
+                {% if category %}
+                    <a class="lecture-category" href="{{ url_for('categories.display', event.category) }}">
+                        {{ category }}
+                    </a>
+                {% else %}
+                    <a class="lecture-category" href="{{ url_for('event_management.settings', event) }}">
+                        {% trans %}Unlisted event{% endtrans %}
+                    </a>
+                {% endif %}
             {% else %}
-                <div>{{ category }}</div>
+                {% if category %}
+                    <div>{{ category }}</div>
+                {% else %}
+                    <div>{% trans %}Unlisted event{% endtrans %}</div>
+                {% endif %}
             {% endif %}
             {% set show_button = not event.is_locked and event.can_manage_attachments(session.user) %}
             <div class="event-manage-button">
--- a/indico/modules/events/templates/display/indico/lecture.html
+++ b/indico/modules/events/templates/display/indico/lecture.html
@@ -13,21 +13,13 @@
 <div class="event-wrapper">
     {% block header %}
         <div class="event-header event-header-lecture {%- if not has_subheader %} round-bottom-corners{% endif %}">
-            {% if not g.static_site %}
-                {% if category %}
+            {% if category %}
+                {% if not g.static_site %}
                     <a class="lecture-category" href="{{ url_for('categories.display', event.category) }}">
                         {{ category }}
                     </a>
                 {% else %}
-                    <a class="lecture-category" href="{{ url_for('event_management.settings', event) }}">
-                        {% trans %}Unlisted event{% endtrans %}
-                    </a>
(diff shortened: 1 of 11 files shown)


indico/modules/events/templates/display/indico/lecture.html
```diff
@@ -14,11 +14,21 @@
...
+                    <a class="lecture-category" href="{{ url_for('event_management.settings', event) }}">
+                        {% trans %}Unlisted event{% endtrans %}
+                    </a>
+                {% endif %}
             {% else %}
-                <div>{{ category }}</div>
+                {% if category %}
+                    <div>{{ category }}</div>
+                {% else %}
+                    <div>{% trans %}Unlisted event{% endtrans %}</div>
```
ThiefMaster, line 30: I wouldn't include that when building a static site... even during normal usage I wonder if we shouldn't just hide the category altogether if there isn't one..

indico/modules/events/templates/header.html
```diff
@@ -32,7 +34,9 @@
            title="{% trans %}Newest event{% endtrans %}"></a>
     {% endif %}
 
-    <span class="separator"></span>
+    {% if event.category or ( (rel['first'], rel['prev'], rel['next'], rel['last'])|any ) %}
```
ThiefMaster, line 37: I think `x or y or z or ...` would be more readable ;) (but what I certainly don't want is a space after an opening or before a closing parenthesis ;))

Jinja also lets you shortcut the dict item access via `rel.first` etc

indico/modules/events/views.py
```diff
@@ -214,7 +214,7 @@ def _get_body(self, params):
 
         return render_template(tpl,
                                event=self.event,
-                               category=self.event.category.title,
+                               category=self.event.category.title if self.event.category else None,
```
ThiefMaster, line 217:
```suggestion
                               category=(self.event.category.title if self.event.category else None),
```

More readable

indico/web/breadcrumbs.py
```diff
@@ -31,6 +31,11 @@ def render_breadcrumbs(*titles, category=None, event=None, management=False, cat
     assert titles or event or category
     if not category and not event:
         items = [(_('Home'), url_for_index())]
+    elif event and not category and not event.category:
```
ThiefMaster, line 34: is the `not category` check actually needed?

indico/modules/events/templates/header.html
javfg: Indeed, that was overly complicated. Also I didn't know about the `rel.first`!

indico/modules/events/templates/display/indico/lecture.html
javfg: I thought it would be a good place to link to the event management view, as usually it will be the creator who visits the display view. But I'm fine either way.

indico/web/breadcrumbs.py
javfg: I check for `category` because of the function signature:

```python
def render_breadcrumbs(*titles, category=None, event=None, management=False, category_url_factory=None):
```

I thought it might be confusing if somebody passes a category in the future and it still renders the unlisted events breadcrumb. But it might be unlikely anyone passes an Event and also a Category.

Let me know if it's best to remove the check despite that.

indico/modules/events/templates/header.html
ThiefMaster: Basically what Jinja  does is catching AttributeError/KeyError when accessing with `.`/`[]` and then trying the other way of accessing. So while it's slightly more performant to use the "right" way, that's negligible and the dot notation is more readable for sure.

One useful pitfall to keep in mind though: `foo.items` for a dict with an `items` key won't work because dicts already have an attribute with that name!

indico/web/breadcrumbs.py
ThiefMaster: If you never expect a category, maybe `assert category is None` inside the if body would be better? That way it's clear that it's not part of the condition, but if the condition is met we're never supposed to have a category.

indico/modules/events/templates/display/indico/lecture.html
ThiefMaster: Hm... there's already the button pointing to the management area anyway..

javfg: Yup, and it's pretty close too. I'll remove it.

Issue #4990, opened 2021-07-06
Unlisted events: UI Mockups


javfg:
This is a subtask for https://github.com/indico/indico/issues/4294.

# UI Mockups for unlisted events

According to https://github.com/indico/indico/issues/4294, there are 3 changes to make in the user interface to accommodate unlisted events:

1. The dashboard must include a list of unpublished events belonging to the user. Entries must contain a 'Publish' button.
2. When creating an event, the category field must be optional, and by default show 'No category', specifying that the event will not be published.
3. The event management view for an unpublished event must show a 'Publish' button. The button will take the user to a category selection dialog similar to the one in event creation.

@pferreir has built a mockup of point 3 already:

![94426977-aaf37d80-018e-11eb-973f-27adf5610f0c](https://user-images.githubusercontent.com/2699/124566988-d739da00-de43-11eb-9bd2-d3747c490e9b.png)

The task for this issue is to create mockups for points 1 and 2.

Following @ThiefMaster's advice, we are using [Balsamiq Wireframes](https://balsamiq.com/wireframes/) for this, as we have licences for the software.

ThiefMaster:
> When creating an event, the category field must be optional, and by default show 'No category', specifying that the event will not be published.

I think it would make sense to have some kind of widget where it's clear what you are doing - so not just "not selecting a category" but rather ticking a "Create as unlisted event" checkbox *or something like that*.

javfg:
### 1. Create event dialog
#### 1.1. Using a button group
| Listed | Unlisted |
| ------- | ------- |
| ![Create event dialog (listed) buttongrp](https://user-images.githubusercontent.com/6058151/124724777-93f76e00-df0c-11eb-8b0f-4fee73a2a622.png) | ![Create event dialog (unlisted) buttongrp](https://user-images.githubusercontent.com/6058151/124725008-cbfeb100-df0c-11eb-843c-90882007022c.png) |

#### 1.2. Using an on/off toggle
| Listed | Unlisted |
| ------- | ------- |
|![Create event dialog (listed) onoff](https://user-images.githubusercontent.com/6058151/124725203-fbadb900-df0c-11eb-88da-55e4744adaa5.png)| ![Create event dialog (unlisted) onoff](https://user-images.githubusercontent.com/6058151/124725218-fd777c80-df0c-11eb-9da1-da7ae3a334b0.png)|

#### 1.3. Using flowing buttons and text
| Listed | Unlisted |
| ------- | ------- |
[...]

Issue #4294, opened 2020-02-17
Unlisted events


pferreir:
Every Indico event is by definition tied to a category. While this is supposed to promote the publication of events, it has also had the not-so-nice side-effect of encouraging people to just dump their events wherever they can (e.g. category "Conferences" in `indico.cern.ch`).
One of the things which is also frustrating about events having to be in a category is that they are visible from the beginning. Even if they are still "under construction", they remain visible.

## Proposal
Indico events should by default belong into a "private" area of their owner. The creator should still be able to change the ACLs (even though they won't respect the permissions of any category). Then, whenever the event is ready, the owner will be able to move it to an actual category and thus make it visible. This could then, in the future, be extended into a further moderation step in which the category managers are supposed to vet the event before it's incorporated into the category.

ThiefMaster:
Sounds very good, and the "by default belong into a private area" could be a simply as making the category field when creating an event optional (maybe with some better UI to make it clear that you can clear the preselected category).

Since we have enumerable event IDs, we should probably also make it very clear that (unless the event is explicitly set as protected), the event cannot be considered secret/private. A header bar could be a good option for this.

ThiefMaster:
I think we have two somewhat separate parts here that can be implemented one by one

## Backend support and fixes

- support events with no category / category_id in the DB (nullable column)
- make sure that everything still works fine in those events, and features that really depend on a category (and thus break) are disabled (can be tested by manually removing the category from an event via `indico shell`)

## Updating existing UI

- support creation of such events (as mentioned in the previous comment)
- clearly indicate in the event that it is unlisted
- add a way to add an unlisted event to a category (maybe similar to how we handle moving events)

## Listing unlisted events

- create mockups for an area listing your own unlisted events (part of the dashboard? new page)
- create said area based on the mockups

pferreir:
Here's a mockup I made some time ago for an internal document:

![image](https://user-images.githubusercontent.com/2699/94426977-aaf37d80-018e-11eb-973f-27adf5610f0c.png)

(could be a basis for discussion)

ThiefMaster:
note: the livesync plugin most likely also need to be updated; we can probably handle it as if an event is in a blacklisted category if there is no category, and then use the "moved" logic which creates a `ChangeType.published` when it gets added to a category

so whoever works on this should have a `debug` livesync agent created so they notice any errors in livesync code when dealing with events with a None category

PR #5023, opened 2021-07-28
Unlisted events

javfg:
# Unlisted events ([#4294](https://github.com/indico/indico/issues/4294))

This PR includes the following tasks:

### Backend changes
- [x] Support events with no category / category_id in the DB (nullable column).
- [x] Fix all problems arising after making category nullable in the event object.

### Updating existing UI (following these [mockups](https://github.com/indico/indico/issues/4990))
- [x] Unlisted event creation.
- [x] Update event management/display view to indicate an event is unlisted.
- [x] Unlisted event "publishing" (wording might change).

### Listing unlisted events
- [x] Create a new dashboard area showing the unlisted events for the user.

---

Additional stuff we should look into:
* When cloning an event, having the new one be an unlisted event should be an option.

### DIFF
--- a/indico/core/db/sqlalchemy/protection.py
+++ b/indico/core/db/sqlalchemy/protection.py
@@ -55,6 +55,8 @@ class ProtectionMixin:
     #: Whether the object can have contact information shown in case of
     #: no access
     allow_no_access_contact = False
+    #: Whether the object can have no protection parent
+    allow_none_protection_parent = False
 
     @classmethod
     def register_protection_events(cls):
@@ -207,15 +209,20 @@ class ProtectionMixin:
             # might not need to check the parents at all
             if self.inheriting_have_acl and any(user in entry.principal for entry in iter_acl(self.acl_entries)):
                 rv = True
+            elif self.allow_none_protection_parent and self.protection_parent is None:
+                # This is the case for unlisted events, which are inheriting
+                # by default but have no parent category, so its parent
+                # protection won't be checked.
+                rv = False
             else:
                 # the parent can be either an object inheriting from this
                 # mixin or a legacy object with an AccessController
                 parent = self.protection_parent
                 if parent is None:
-                    # This is the case for unlisted events, which are inheriting
-                    # by default but have no parent category, so its parent
-                    # protection won't be checked.
-                    rv = False
+                    # This should be the case for the top-level object,
+                    # i.e. the root category, which shouldn't allow
+                    # ProtectionMode.inheriting as it makes no sense.
+                    raise TypeError(f'protection_parent of {self} is None')
                 elif hasattr(parent, 'can_access'):
                     rv = parent.can_access(user, allow_admin=allow_admin)
                 else:
--- a/indico/core/db/sqlalchemy/protection.py
+++ b/indico/core/db/sqlalchemy/protection.py
@@ -138,6 +138,9 @@ class ProtectionMixin:
         if self.disable_protection_mode:
             raise NotImplementedError
         if self.is_inheriting:
+            # Unlisted events inherit protections but have no parent, they are visible only by the owner and ACLs.
+            if not self.protection_parent:
+                return True
             return self.protection_parent.is_protected
         else:
             return self.is_self_protected
--- a/indico/core/db/sqlalchemy/protection.py
+++ b/indico/core/db/sqlalchemy/protection.py
@@ -212,10 +212,10 @@ class ProtectionMixin:
                 # mixin or a legacy object with an AccessController
                 parent = self.protection_parent
                 if parent is None:
-                    # This should be the case for the top-level object,
-                    # i.e. the root category, which shouldn't allow
-                    # ProtectionMode.inheriting as it makes no sense.
-                    raise TypeError(f'protection_parent of {self} is None')
+                    # This is the case for unlisted events, which are inheriting
+                    # by default but have no parent category, so its parent
+                    # protection won't be checked.
+                    rv = False
                 elif hasattr(parent, 'can_access'):
                     rv = parent.can_access(user, allow_admin=allow_admin)
                 else:
--- a/indico/core/db/sqlalchemy/protection.py
+++ b/indico/core/db/sqlalchemy/protection.py
@@ -317,6 +317,9 @@ class ProtectionMixin:
             self.acl_entries.remove(entry)
 
     def get_inherited_acl(self):
+        if self.protection_parent is None:
+            return []
+
         own_acl = {entry.principal for entry in self.acl_entries}
         parent_acl = self.protection_parent.get_access_list(skip_managers=True)
         return [x for x in parent_acl if x not in own_acl]
--- a/indico/migrations/versions/20210813_1633_dc53d6e8c576_make_event_category_nullable.py
+++ b/indico/migrations/versions/20210813_1633_dc53d6e8c576_make_event_category_nullable.py
@@ -1,7 +1,7 @@
 """Make event category nullable
 
 Revision ID: dc53d6e8c576
-Revises: 1cec32e42f65
+Revises: 9d00917b2fa8
 Create Date: 2021-07-07 16:48:03.719901
 """
 
@@ -10,7 +10,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = 'dc53d6e8c576'
-down_revision = '1cec32e42f65'
+down_revision = '9d00917b2fa8'
 branch_labels = None
 depends_on = None
 
--- a/indico/migrations/versions/20210813_1633_dc53d6e8c576_make_event_category_nullable.py
+++ b/indico/migrations/versions/20210813_1633_dc53d6e8c576_make_event_category_nullable.py
@@ -17,6 +17,9 @@ depends_on = None
 
 def upgrade():
     op.drop_constraint('ck_events_category_data_set', 'events', schema='events')
+    op.create_check_constraint('unlisted_events_always_inherit', 'events',
+                               'is_deleted OR category_id IS NOT NULL OR protection_mode = 1',
+                               schema='events')
 
 
 def downgrade():
@@ -31,3 +34,4 @@ def downgrade():
         ADD CONSTRAINT "ck_events_category_data_set"
         CHECK (category_id IS NOT NULL OR is_deleted)
     ''')
+    op.drop_constraint('ck_events_unlisted_events_always_inherit', 'events', schema='events')
--- a/indico/migrations/versions/20210813_1633_dc53d6e8c576_make_event_category_nullable.py
+++ b/indico/migrations/versions/20210813_1633_dc53d6e8c576_make_event_category_nullable.py
@@ -20,4 +20,14 @@ def upgrade():
 
 
 def downgrade():
-    op.create_check_constraint('category_data_set', 'events', 'category_id IS NOT NULL OR is_deleted', schema='events')
+    op.execute('SET CONSTRAINTS ALL IMMEDIATE')
+    op.execute('''
+        UPDATE events.events
+        SET is_deleted = true
+        WHERE category_id IS NULL AND NOT is_deleted
+    ''')
+    op.execute('''
+        ALTER TABLE events.events
+        ADD CONSTRAINT "ck_events_category_data_set"
+        CHECK (category_id IS NOT NULL OR is_deleted)
+    ''')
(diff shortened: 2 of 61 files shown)


indico/modules/users/templates/dashboard.html
```diff
@@ -82,6 +82,36 @@ <h3>{{ user.full_name }}</h3>
...
                 </div>
+                <div id="happeningCategories" class="dashboard-box">
+                    <h3>{% trans %}Your unlisted events{% endtrans %}</h3>
+                    <ol>
+                        {% if not unlisted_events %}
+                            <li class="no-event">
+                                <span class="event-title italic text-superfluous">
+                                    {% trans %}You have no unlisted events.{% endtrans %}
+                                </span>
+                            </li>
```
ThiefMaster, line 93: IIRC Jinja lets you put an `else` block in a `for` loop that executes when there was nothing to iterate over - so you could use this for the no-event case.

ThiefMaster: But do we want to show this box at all if there aren't unlisted events? Especially in case someone is not allowed to create them I would not want this to show up.

indico/modules/users/templates/dashboard.html
```diff
@@ -82,6 +82,36 @@ <h3>{{ user.full_name }}</h3>
                         </a>
                     </div>
                 </div>
+                <div id="happeningCategories" class="dashboard-box">
```
ThiefMaster, line 85: The `happeningCategories` ID is now duplicate, and still unused so let's just remove it in both places

indico/modules/categories/util.py
```diff
@@ -176,13 +176,16 @@ def get_visibility_options(category_or_event, allow_invisible=True):
...
+    options = []
+
+    if category is not None:
+        options = [(n + 1, ('{} \N{RIGHTWARDS ARROW} "{}"'.format(_category_above_message(n).format(n), title)))
+                   for n, title in enumerate(category.chain_titles[::-1])]
+        if event is None:
+            options[0] = (1, _('From this category only'))
+        else:
+            options[0] = (1, '{} \N{RIGHTWARDS ARROW} "{}"'.format(_('From the current category only'), category.title))
+        options[-1] = ('', _('From everywhere'))
```
ThiefMaster, line 188: I wonder if we shouldn't hide the visibility setting altogether in unlisted events. Right now you can only select "invisible", but that's *very* rarely the setting you want for an event that's in a category...

indico/modules/events/management/client/js/EventMove.jsx
```diff
@@ -80,52 +87,79 @@ function EventMove({currentCategoryId, submitMove, renderTrigger, getEventCount,
...
+
                       </p>
                       <FinalTextArea
                         name="comment"
                         label={Translate.string('Comment')}
-                        description={
+                        description={publish ?
+                          <Translate>
+                            You can provide a comment to the category managers to help them decide
+                            whether to approve your publish request.
```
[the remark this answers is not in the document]
javfg: Good idea!

indico/modules/categories/util.py
javfg: Indeed, the drop-down is empty most of the time.

ThiefMaster: yep, and by saving it you would save the "invisible" setting ;)

indico/modules/events/management/controllers/protection.py
```diff
@@ -85,7 +85,7 @@ def _process(self):
             update_event_protection(self.event, {'protection_mode': form.protection_mode.data,
                                                  'own_no_access_contact': form.own_no_access_contact.data,
                                                  'access_key': form.access_key.data,
-                                                 'visibility': form.visibility.data,
+                                                 'visibility': form.visibility.data if form.visibility else None,
```
ThiefMaster, line 88: Should we set it to None or skip this altogether?

FWIW, setting it to None doesn't seem like a problem, the only case where it would end up changing something is when we convert an event from listed to unlisted (not sure if that's possible and/or whether we need this), that had an explicit value set...

indico/modules/events/management/forms.py
```diff
@@ -270,6 +273,9 @@ def _get_event_own_visibility_horizon(self, event):
             return event.category.nth_parent(self.visibility.data - 1)
 
     def _init_visibility(self, event):
+        if event.category is None:
+            return None
```
ThiefMaster, line 277: We don't call this method, so maybe this should be an `assert event.category`?

indico/modules/events/models/events.py
```diff
@@ -923,16 +923,24 @@ def move(self, category, *, log_meta=None):
...
+        if old_category:
+            self.log(EventLogRealm.management, LogKind.change, 'Category', 'Event moved', session.user,
+                     data={'From': old_path, 'To': new_path}, meta=log_meta)
+            old_category.log(CategoryLogRealm.events, LogKind.negative, 'Content', f'Event moved out: "{self.title}"',
+                             session.user, data={'ID': self.id, 'To': new_path}, meta=log_meta)
+            category.log(CategoryLogRealm.events, LogKind.positive, 'Content', f'Event moved in: "{self.title}"',
+                         session.user, data={'From': old_path}, meta=log_meta)
+        else:
+            notify_event_creation(self)
+
```
ThiefMaster, line 939: matter of preference, but i wouldn't put an empty line here... 👾

indico/modules/users/templates/dashboard.html
```diff
@@ -82,6 +82,30 @@ <h3>{{ user.full_name }}</h3>
                         </a>
                     </div>
                 </div>
+                {% if unlisted_events %}
+                    <div class="dashboard-box">
+                        <h3>{% trans %}Your unlisted events{% endtrans %}</h3>
+                        <ol>
```
ThiefMaster, line 88: shouldn't this be `<ul>`? the list isn't really ordered/numbered, is it?

indico/web/breadcrumbs.py
```diff
@@ -5,18 +5,39 @@
...
+    """A breadcrumb to be rendered by the template.
+
+    If `url` is present, the breadcrumb will be a link to it.
+
+    The `content` is typically string or Markup object, but in some special cases
+    it may also be a different object.
+
+    Typically this class does not need to be used when registering a template hook
+    unless you need the ability to specify priority or disable markup wrapping at
+    execution time, or yield multiple snippets with different settings.
```
ThiefMaster, line 28: That block isn't correct. Copy&paste from TemplateSnippet? :P

indico/modules/events/management/controllers/protection.py
javfg: I wasn't sure either, but this can be confusing if we later enable some kind of option to 'unpublish' an event. (Which I'm not sure is a good idea). So I'll skip it if not present in the form.

indico/modules/users/templates/dashboard.html
```diff
@@ -82,6 +82,30 @@ <h3>{{ user.full_name }}</h3>
                         </a>
                     </div>
                 </div>
+                {% if unlisted_events %}
+                    <div class="dashboard-box">
+                        <h3>{% trans %}Your unlisted events{% endtrans %}</h3>
+                        <ul>
```
ThiefMaster, line 88: I just realized that we use `<ol>` for the other boxes and have CSS rules that explicitly use `ol`. So let's either keep it (even though it's weird) or update it in the other places and css as well

image

indico/modules/events/client/js/creation.js
```diff
@@ -58,20 +60,39 @@ import {camelizeKeys} from 'indico/utils/case';
...
-      elem.find('.js-category-title').text(currentCategory.title);
+      const elem = messages.filter(`.${mode}-protection-message`);
+      elem.find('.js-category-title').text(currentCategory && currentCategory.title);
       protectionMessage.html(elem);
     }
 
     options.categoryField.on('indico:categorySelected', (evt, cat) => {
-      if (!currentCategory) {
+      if (!options.listingField.prop('checked')) {
+        options.protectionModeFields.filter('[value=inheriting]').prop('disabled', true);
```
ThiefMaster, line 70: I think just disabling "inheriting" is a bad idea; this results in the event being created as a fully public, event.

Some possible options:

- keep the choice, but indicate that it's meaningless for unlisted events
- hide the choice (during creation and management), always create as inheriting, and explicit refuse access in case of an unlisted event if there's no explicit ACL entry
- hide the choice (during creation and management), always create as protected to get the same behavior from the previous bullet point. but then we should probably indicate that the event is protected and ask the user if they want to change it to inheriting when publishing the event to a category

Note that an unlisted event should never be fully public since we do not want to let someone enumerate them by just iterating over IDs and trying to access the event.

ThiefMaster:
Right now selecting a category is required when cloning an event. Should we keep it as it is or allow cloning as unlisted? (in any case this is not something for this PR, just something I realized when testing whether the clone form works ;))

indico/modules/events/client/js/creation.js
ThiefMaster: In this context, I wonder if this message is clear or misleading:

> Unlisted events are not shown in any category **and only the creator can see them.** 

I understand what you mean, but it could also be interpreted as "only the creator is able to access them" which isn't the case of course if the event can be made public.

But with the suggested changes, maybe it's OK - I guess it's obvious that someone who's added to the ACL can access the event... Alternatively, we could change it to something about "only explicitly authorized users" being able to see the event...

javfg: As discussed, creating them as inheriting and hiding the choices (option 2) seems the more sensible UX. I've implemented that one.

javfg:
> Right now selecting a category is required when cloning an event. Should we keep it as it is or allow cloning as unlisted? (in any case this is not something for this PR, just something I realized when testing whether the clone form works ;))

I think we should allow cloning as unlisted. I mentioned it in the initial post but I agree it's for another PR.

indico/core/db/sqlalchemy/protection.py
```diff
@@ -212,10 +212,10 @@ def can_access(self, user, allow_admin=True):
                 # mixin or a legacy object with an AccessController
                 parent = self.protection_parent
                 if parent is None:
-                    # This should be the case for the top-level object,
-                    # i.e. the root category, which shouldn't allow
-                    # ProtectionMode.inheriting as it makes no sense.
-                    raise TypeError(f'protection_parent of {self} is None')
```
ThiefMaster, line 218: Can we be more explicit, something like `allow_none_protection_parent`? I'd rather not disable this exception anywhere but on `Event`

indico/web/client/styles/modules/_dashboard.scss
```diff
@@ -82,20 +82,23 @@
   background: #dcebf5;
 }
 
-.dashboard-box ol {
+.dashboard-box ol,
+.dashboard-box ul {
```
ThiefMaster, line 86: I'd change the other `ol` in the dashboard template as well instead of defining the rule for both `ol` and `ul`..

javfg: While at it, I increased the width of the date field so it does not wrap on `last Wednesday` entry:

a

indico/migrations/versions/20210813_1633_dc53d6e8c576_make_event_category_nullable.py
```diff
@@ -0,0 +1,23 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.drop_constraint('ck_events_category_data_set', 'events', schema='events')
+
+
+def downgrade():
+    op.create_check_constraint('category_data_set', 'events', 'category_id IS NOT NULL OR is_deleted', schema='events')
```
ThiefMaster, line 23: Maybe we should also flag all events with no category that aren't deleted as deleted? Otherwise a downgrade fails (which is a bit annoying during development)

javfg: Ah, good idea, I was deleting them manually but this is much more convenient.

indico/web/client/js/react/components/WTFButtonsBooleanField.jsx
```diff
@@ -0,0 +1,57 @@
...
+        <strong>{falseText}</strong>
+      </IButton>
+    </>
+  );
+}
+
+WTFButtonsBooleanField.propTypes = {
+  checkboxId: PropTypes.string.isRequired,
+  trueCaption: PropTypes.array,
+  falseCaption: PropTypes.array,
```
ThiefMaster, line 49: I think `PropTypes.node` would be better here... then you can also just pass plain strings in the default value

PR #5024, opened 2021-07-28
Allow disabling public search of Indico users

OmeGak:
This PR makes it possible for Indico instances to limit searching users by email to only managers. This includes:
- Disabling registering to events as a different user than the one logged in.
- Disabling speakers to give access to specific users/groups in protected uploaded material.
- Disabling users to add other users as author/co-author in abstract submissions (details must be added manually).

### DIFF
--- a/indico/modules/attachments/controllers/management/base.py
+++ b/indico/modules/attachments/controllers/management/base.py
@@ -10,6 +10,7 @@ import mimetypes
 from flask import flash, render_template, request, session
 
 from indico.core import signals
+from indico.core.config import config
 from indico.core.db import db
 from indico.modules.attachments import logger
 from indico.modules.attachments.controllers.util import SpecificAttachmentMixin, SpecificFolderMixin
@@ -32,8 +33,9 @@ def _render_attachment_list(linked_object):
     return tpl.render_attachments(attachments=get_attached_items(linked_object), linked_object=linked_object)
 
 
-def _render_protection_message(linked_object):
-    return render_template('attachments/_protection_message.html', parent=_get_parent_info(linked_object))
+def _render_protection_message(linked_object, users_allowed=True):
+    return render_template('attachments/_protection_message.html',
+                           parent=_get_parent_info(linked_object), users_allowed=users_allowed)
 
 
 def _get_parent_info(parent):
@@ -57,6 +59,14 @@ def _get_folders_protection_info(linked_object):
     return {folder.id: folder.is_self_protected for folder in folders}
 
 
+def _is_user_search_allowed(obj=None):
+    if config.ALLOW_PUBLIC_USER_SEARCH:
+        return True
+    elif obj and obj.can_manage(session.user):
+        return True
+    return False
+
+
 class ManageAttachmentsMixin:
     """Show the attachment management page."""
     wp = None
@@ -76,7 +86,8 @@ class AddAttachmentFilesMixin:
     """Upload file attachments."""
 
     def _process(self):
-        form = AddAttachmentFilesForm(linked_object=self.object)
+        user_search_allowed = _is_user_search_allowed(self.object)
+        form = AddAttachmentFilesForm(linked_object=self.object, acl_enabled=user_search_allowed)
         if form.validate_on_submit():
             files = form.files.data
             folder = form.folder.data or AttachmentFolder.get_or_create_default(linked_object=self.object)
@@ -97,7 +108,8 @@ class AddAttachmentFilesMixin:
                   .format(count=len(files)), 'success')
             return jsonify_data(attachment_list=_render_attachment_list(self.object))
         return jsonify_template('attachments/upload.html', form=form, action=url_for('.upload', self.object),
-                                protection_message=_render_protection_message(self.object),
+                                protection_message=_render_protection_message(self.object,
+                                                                              users_allowed=user_search_allowed),
                                 folders_protection_info=_get_folders_protection_info(self.object),
                                 existing_attachment=None)
 
@@ -106,13 +118,15 @@ class AddAttachmentLinkMixin:
     """Add link attachment."""
 
     def _process(self):
-        form = AddAttachmentLinkForm(linked_object=self.object)
+        user_search_allowed = _is_user_search_allowed(self.object)
+        form = AddAttachmentLinkForm(linked_object=self.object, acl_enabled=user_search_allowed)
         if form.validate_on_submit():
             add_attachment_link(form.data, self.object)
             flash(_('The link has been added'), 'success')
             return jsonify_data(attachment_list=_render_attachment_list(self.object))
         return jsonify_template('attachments/add_link.html', form=form,
-                                protection_message=_render_protection_message(self.object),
+                                protection_message=_render_protection_message(self.object,
+                                                                              users_allowed=user_search_allowed),
                                 folders_protection_info=_get_folders_protection_info(self.object))
 
 
@@ -120,11 +134,13 @@ class EditAttachmentMixin(SpecificAttachmentMixin):
     """Edit an attachment."""
 
     def _process(self):
+        user_search_allowed = _is_user_search_allowed(self.object)
         defaults = FormDefaults(self.attachment, protected=self.attachment.is_self_protected, skip_attrs={'file'})
         if self.attachment.type == AttachmentType.file:
-            form = EditAttachmentFileForm(linked_object=self.object, obj=defaults, file=self.attachment)
+            form = EditAttachmentFileForm(linked_object=self.object, obj=defaults, file=self.attachment,
+                                          acl_enabled=user_search_allowed)
         else:
-            form = EditAttachmentLinkForm(linked_object=self.object, obj=defaults)
+            form = EditAttachmentLinkForm(linked_object=self.object, obj=defaults, acl_enabled=user_search_allowed)
 
         if form.validate_on_submit():
             folder = form.folder.data or AttachmentFolder.get_or_create_default(linked_object=self.object)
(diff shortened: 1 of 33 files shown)


indico/modules/events/abstracts/forms.py
```diff
@@ -498,7 +498,8 @@ def __init__(self, *args, **kwargs):
         if not is_invited:
             self.person_links.require_speaker_author = abstracts_settings.get(self.event, 'speakers_required')
             self.person_links.allow_speakers = abstracts_settings.get(self.event, 'allow_speakers')
-            self.person_links.disable_user_search = session.user is None
+            self.person_links.disable_user_search = (not config.ALLOW_PUBLIC_USER_SEARCH and
+                                                     not self.event.can_manage(session.user))
```
ThiefMaster, line 502: This is missing the `or session.user is None` - we never allow unauthenticated users to search.

(even though I'm not sure whether they can even get to the point where they get to the search dialog without being asked to login first...)

indico/web/client/js/jquery/widgets/jinja/person_link_widget.js
```diff
@@ -544,9 +544,11 @@ import {$T} from 'indico/utils/i18n';
       );
     };
 
-    ReactDOM.render(
-      <UserSearchWrapper />,
-      document.getElementById(`principalField-${options.fieldId}`)
-    );
+    if (!options.disableUserSearch) {
```
ThiefMaster, line 547: I don't think this is needed anymore - we already have `disabled={options.disableUserSearch}` that disables the button if user search is not allowed.

indico/web/templates/_protection_messages.html
```diff
@@ -16,7 +16,7 @@
     </div>
 {%- endmacro %}
 
-{% macro render_protected_protection_message() -%}
+{% macro render_protected_protection_message(users_allowed=True) -%}
```
ThiefMaster, line 19:
```suggestion
{% macro render_protected_protection_message(users_allowed=true) -%}
```

indico/web/templates/_protection_messages.html
```diff
@@ -25,9 +25,16 @@
                     {% trans %}Protected{% endtrans %}
                 </div>
                 <div>
-                    {% trans -%}
-                        This object is <strong>only</strong> accessible by the <strong>users specified</strong> above and the <strong>managers</strong> of <strong>parent resources</strong>.
-                    {%- endtrans %}
+                    {% if users_allowed %}
+                        {% trans -%}
+                            This object is <strong>only</strong> accessible by the <strong>users specified</strong> above and the <strong>managers</strong> of <strong>parent resources</strong>.
```
ThiefMaster, line 30: Let's wrap this? :D 

I think the overly long line is still from the time where we did not collapse whitespace inside trans blocks...

indico/modules/events/registration/util.py
```diff
@@ -105,6 +105,9 @@ def check_registration_email(regform, email, registration=None, management=False
     user = get_user_by_email(email)
     email_registration = regform.get_registration(email=email)
     user_registration = regform.get_registration(user=user) if user else None
+    email_err = validate_email_verbose(email)
+    if email_err:
+        return dict(status='error', conflict='email-invalid', email_error=email_err)
```
ThiefMaster, line 110: Hm, didn't we say something about ignoring this if there's a registration and the user/email did not change (`user and user == registration.user`)

indico/modules/events/abstracts/forms.py
OmeGak: It's required to be logged in but this is future-proof, in case it's ever allowed to submit abstracts without being logged.

indico/modules/attachments/controllers/management/base.py
```diff
@@ -57,6 +59,13 @@ def _get_folders_protection_info(linked_object):
     return {folder.id: folder.is_self_protected for folder in folders}
 
 
+def _is_user_search_allowed(event=None):
+    user_search_allowed = True
+    if not config.ALLOW_PUBLIC_USER_SEARCH and event and not event.can_manage(session.user):
+        user_search_allowed = False
+    return user_search_allowed
```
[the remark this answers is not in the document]
OmeGak: Passing `self.event` was, in fact, failing. I changed it to passing `self.object` and it works fine for both categories and events.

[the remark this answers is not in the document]
ThiefMaster: `self.object` can be a contribution, subcontribution or session as well

indico/web/client/js/jquery/widgets/jinja/person_link_widget.js
OmeGak: `disabled={options.disableUserSearch}` may be broken. It only disables the button.

ThiefMaster: Yep, that's the intended behavior - I don't think it's wrong to disable the button instead of hiding it...

indico/modules/events/forms.py
```diff
@@ -97,3 +98,7 @@ class LectureCreationForm(EventCreationFormBase):
     person_link_data = EventPersonLinkListField(_('Speakers'))
     description = TextAreaField(_('Description'), widget=CKEditorWidget())
     theme = IndicoThemeSelectField(_('Theme'), event_type=EventType.lecture, allow_default=True)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.person_link_data.disable_user_search = not config.ALLOW_PUBLIC_USER_SEARCH
```
ThiefMaster, line 104: `and not session.user.is_admin`

indico/modules/users/controllers.py
```diff
@@ -738,6 +740,24 @@ def _process_pending_users(self, results):
             if ext_id is not None:
                 cache.set(ext_id, self.externals[ext_id], timeout=86400)
 
+    @use_kwargs({
+        'category_id': fields.Integer(validate=lambda x: x >= 0),
+        'event_id': fields.Integer(validate=lambda x: x >= 0)
```
ThiefMaster, line 745: I'd skip the validate here; we don't use this for any other ID fields either (searched for `_id': fields.Int`).

*If* we wanted this level of validation I'd rather add a new custom type like `DatabaseID` which does the validation - but I think that's not really necessary since a negative ID is not invalid per se, it's just an ID that won't have a corresponding object.

indico/modules/users/controllers.py
```diff
@@ -738,6 +740,24 @@ def _process_pending_users(self, results):
...
+    }, location='query')
+    def _process_args(self, event_id=None, category_id=None):
+        print(event_id, category_id)
```
ThiefMaster, line 748: 👀

indico/modules/users/controllers.py
```diff
@@ -738,6 +740,24 @@ def _process_pending_users(self, results):
...
+        super(RHProtected, self)._process_args()
```
ThiefMaster, line 749: I'd keep the old-style `RHProtected._process_args(self)` call here; we have some places where we are explicit about which parent class methods we call (usually w/ protection checks though) and AFAIK it's not recommended to mix super and non-super parent calls.

indico/modules/users/controllers.py
```diff
@@ -738,6 +740,24 @@ def _process_pending_users(self, results):
...
+        self.event = Event.query.filter_by(id=event_id).first()
+        self.category = Category.query.filter_by(id=category_id).first()
+
+    def _check_access(self):
+        super(RHProtected, self)._check_access()
```
ThiefMaster, line 754: same here

indico/modules/users/controllers.py
```diff
@@ -738,6 +740,24 @@ def _process_pending_users(self, results):
...
 
+    @use_kwargs({
+        'category_id': fields.Integer(validate=lambda x: x >= 0),
+        'event_id': fields.Integer(validate=lambda x: x >= 0)
+    }, location='query')
+    def _process_args(self, event_id=None, category_id=None):
+        print(event_id, category_id)
+        super(RHProtected, self)._process_args()
+        self.event = Event.query.filter_by(id=event_id).first()
+        self.category = Category.query.filter_by(id=category_id).first()
```
ThiefMaster, line 751:
```suggestion
        self.event = Event.get(event_id, is_deleted=False)
        self.category = Category.get(category_id, is_deleted=False)
```

indico/modules/users/controllers.py
OmeGak: 🤭

indico/modules/events/abstracts/forms.py
```diff
@@ -498,7 +498,9 @@ def __init__(self, *args, **kwargs):
         if not is_invited:
             self.person_links.require_speaker_author = abstracts_settings.get(self.event, 'speakers_required')
             self.person_links.allow_speakers = abstracts_settings.get(self.event, 'allow_speakers')
-            self.person_links.disable_user_search = session.user is None
+            self.person_links.disable_user_search = ((not config.ALLOW_PUBLIC_USER_SEARCH) and
+                                                     (not self.event.can_manage(session.user)) or
+                                                     (session.user is None))
```
ThiefMaster, line 503: Not really what I meant in my last comment - you still have `and` and `or` on the same level.

What I want to avoid is `a and b or c` but rather being explicit using `(a and b) or c` / `a and (b or c)`

ThiefMaster (commented):
there are also a bunch of linter warnings

ThiefMaster:
The "Add Indico user" button on `/user/favorites/` should be disabled or hidden when public search is disabled. Same for the backend that adds a new favorite user as one could just bruteforce-add all the user ids to the favorites and then view them on that page.

Some widgets (mostly simple ACL widgets like the one used in the room booking settings) already support marking any user there as a favorite, but eventually we can add this everywhere or even in the user search result dialog (but not needed in this PR).

ThiefMaster:
`RHPrincipals` / `RHEventPrincipals` need the same kind of protection as `RHUserSearch`

OmeGak:
> The "Add Indico user" button on `/user/favorites/` should be disabled or hidden when public search is disabled.

Done.

kewisch:
@OmeGak Would you have some time to update this PR so we can get it merged? I'd love to make use of this in our instance.

OmeGak:
Hi @kewisch. It's been some time I didn't work on this PR after I bumped into a few non-trivial corner cases. I can find some time to rebase the code onto the current `master` but I don't think I will be able to finish it for now, as it's not a priority for my client. If you would like to have it finished urgently, maybe @ThiefMaster can guide you through whatever was missing?

PR #5025, opened 2021-07-30
Add support for a state reason in category moderation

plourenco:
Depends on #5013 

imagem

### DIFF
--- a/indico/modules/categories/models/event_move_request.py
+++ b/indico/modules/categories/models/event_move_request.py
@@ -7,7 +7,6 @@
 
 from indico.core.db import db
 from indico.core.db.sqlalchemy import PyIntEnum, UTCDateTime
-from indico.modules.events import EventLogKind, EventLogRealm
 from indico.util.date_time import now_utc
 from indico.util.enum import RichIntEnum
 from indico.util.locators import locator_property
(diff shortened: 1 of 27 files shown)


indico/modules/categories/models/event_move_request.py
```diff
@@ -54,6 +54,10 @@ class EventMoveRequest(db.Model):
         default=MoveRequestState.pending,
         nullable=False
     )
+    state_reason = db.Column(
+        db.String,
+        nullable=True
```
ThiefMaster, line 59: I'd use `nullable=False` and `default=''`; otherwise we'll almost certainly end up with a mix of empty string and NULLs at some point...

And maybe "moderator_comment" would be a better name? "state_reason" sounds a bit strange to me..

indico/modules/categories/operations.py
```diff
@@ -45,10 +45,12 @@ def update_category(category, data, skip=()):
...
 
 
-def update_event_move_request(request, accept):
+def update_event_move_request(request, accept, reason=None):
     request.state = MoveRequestState.rejected
+    request.state_reason = reason
     request.moderator = session.user
     if accept:
         request.state = MoveRequestState.accepted
+        request.state_reason = None
```
ThiefMaster, line 54: I don't think that's needed since a pending one will never hav a reason..

indico/modules/categories/models/event_move_request.py
tomasr8: Changed in #5013

PR #5027, opened 2021-08-02
Allow imagemaps in advanced RichText editors

bpedersen2:
- enable area and map elements for ckeditor
 - enable usemap on img tags in ckeditor
 - enable the coords and usemap attributes in bleach

Note: map and area where already enabled in the bleach filters.


ThiefMaster:
@bpedersen2 I updated the PR to avoid allowing those attributes everywhere but only on the elements where they are needed. Seems to work fine (as confirmed by the unit test I added ;))

ThiefMaster (approved):

PR #5031, opened 2021-08-04
Extract badge format logic into util

vasantvohra:
- We have a new module called `Accreditation` which is connected to a `Category` for badge printing purposes.
 - The RH to print badge which generates pdf has some duplicated code to avoid it may we have a Mixin upstream?
 - This would close [this code review suggestion](https://github.com/UNOG-Indico/indico-un/pull/505#discussion_r681896699)
 
 
image

### DIFF
--- a/indico/modules/designer/util.py
+++ b/indico/modules/designer/util.py
@@ -10,11 +10,25 @@ from sqlalchemy.orm import joinedload
 
 from indico.core.db import db
 from indico.modules.designer.models.templates import DesignerTemplate
+from indico.modules.designer.pdf import PIXELS_CM
 from indico.modules.designer.placeholders import GROUP_TITLES
 from indico.modules.events.models.events import Event
 from indico.modules.events.registration.models.forms import RegistrationForm
 from indico.util.date_time import now_utc
 from indico.util.placeholders import get_placeholders
+
+
+FORMAT_MAP_PORTRAIT = {
+    'A0': (84.1, 118.9),
+    'A1': (59.4, 84.1),
+    'A2': (42.0, 59.4),
+    'A3': (29.7, 42.0),
+    'A4': (21.0, 29.7),
+    'A5': (14.8, 21.0),
+    'A6': (10.5, 14.8),
+    'A7': (7.4, 10.5),
+    'A8': (5.2, 7.4),
+}
 
 
 def get_placeholder_options():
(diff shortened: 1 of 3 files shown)


ThiefMaster:
I think this should be a util, not a mixin: `format_map_portrait` and `format_map_landscape` are static and can be module-level variables. And `_get_format` doesn't use anything except these two variables so it doesn't need to be in the class either. It can be just a `get_badge_format` util in designer/util.py

vasantvohra:
`format_map_portrait` and `format_map_landscape` module-level variables in designer or registration module?

ThiefMaster:
I'd put it all in designer/util

vasantvohra:
> I'd put it all in designer/util
I made only `format_map_portrait` module-level global on `designer/util` and tested :)

indico/modules/designer/util.py
```diff
@@ -69,3 +83,11 @@ def get_default_badge_on_category(category, only_inherited=False):
     parent_chain = category.parent_chain_query.options(joinedload('default_badge_template')).all()
     return next((category.default_badge_template for
                  category in reversed(parent_chain) if category.default_badge_template), None)
+
+
+def get_badge_format(tpl):
+    format_map_landscape = {name: (h, w) for name, (w, h) in FORMAT_MAP_PORTRAIT.items()}
+    format_map = format_map_landscape if tpl.data['width'] > tpl.data['height'] else FORMAT_MAP_PORTRAIT
```
ThiefMaster, line 90: How about this instead (or use ternary if it doesn't look too bad with the rapping)? 

```python
if tpl.data['width'] > tpl.data['height']:
    format_map = {name: (h, w) for name, (w, h) in FORMAT_MAP_PORTRAIT.items()}
else:
    format_map = FORMAT_MAP_PORTRAIT
```

indico/modules/designer/util.py
```diff
@@ -10,13 +10,27 @@
...
+from indico.modules.designer.pdf import PIXELS_CM
 from indico.modules.designer.placeholders import GROUP_TITLES
 from indico.modules.events.models.events import Event
 from indico.modules.events.registration.models.forms import RegistrationForm
 from indico.util.date_time import now_utc
 from indico.util.placeholders import get_placeholders
 
 
+FORMAT_MAP_PORTRAIT = {
+        'A0': (84.1, 118.9),
```
ThiefMaster, line 22: Too much indentation

vasantvohra: Done, I didn't notice it while copy-pasting.

indico/modules/designer/util.py
vasantvohra: Done this too 👍

PR #5032, opened 2021-08-05
Preserve multi-choice field order in registration emails

### LINKED ISSUE #5020: Multi-choice option order in regform summary
> We have added a multi-choice field to the registration form, so that participants can tick the workshops they want to participate in during the conference. The ticking works well, but we still have two problems:

> In the summary the participants receive after their registration, the choices are not sorted in the same order that they were sorted in the registration form: In the form, we sorted them chronologically, but in the summary, they appear to be sorted alphabetically, which makes it confusing for participants, as it does not reflect the time schedule anymore.

I haven't tried to reproduce it, but it sounds like a bug...

tomasr8:
Closes #5020 

The problem is that `RegistrationData` is juggled around in many
places and doesn't keep the order given by the `RegistrationFormField`.
This is a problem only when asked for `friendly_data`.

Since `RegistrationData` keeps a reference to the underlying form field,
we can restore the order when we get the friendly data.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,7 @@ Bugfixes
 - Show correct placeholders in date picker fields (:pr:`5022`)
 - Correctly preselect the default currency when creating a registration form
 - Do not notify registrants when a payment transaction is created in "pending" state
+- Keep the order of multi-choice options in registration summary (:issue:`5020`, :pr:`5032`)
 
 Internal Changes
 ^^^^^^^^^^^^^^^^
--- a/indico/modules/events/registration/fields/choices.py
+++ b/indico/modules/events/registration/fields/choices.py
@@ -5,6 +5,7 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
+import sys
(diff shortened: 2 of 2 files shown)


indico/modules/events/registration/templates/emails/base_registration_details.html
```diff
@@ -43,7 +43,7 @@ <h2>{{ section.title }}</h2>
                 {{- friendly_data.choice }} ({{ friendly_data.arrival_date | format_date(locale='en_GB') }} - {{ friendly_data.departure_date | format_date(locale='en_GB') -}})
             {% endif %}
         {%- elif regdata.field_data.field.input_type == 'multi_choice' -%}
-            {{- friendly_data | join(', ') }}
+            {{- regdata.get_friendly_data(sort=False) | join(', ') }}
```
ThiefMaster, line 46:
```suggestion
            {{- regdata.get_friendly_data(sort=false) | join(', ') }}
```

in jinja the all-lowercase true/false/none is preferred

ThiefMaster (approved):

PR #5034, opened 2021-08-11
Allow filtering abstracts by 'No selection'

### LINKED ISSUE #5033: List of abstracts: filters are not well defined (missing "no value")
**Bug Description**
_It's not possible to filter the of abstracts according to some fields where no values are wanted._

**To Reproduce**
Let's say that the admin adds a custom field (not visible to submitters) called "abstract state" for a custom workflow that enhances the Indico built in. For the sake of simplicity I consider here a radio button called "**Abstract QA**" with only one value: **Yes**. This is editable only by the admin, so by default this field will have no value at submission type.

Let's say that we have 100 abstracts submitted. The admin wants to "pre-accept" them by editing them and setting the "Abstract QA" field. This admin edits 80 abstracts and sets this field to YES, then stops. 
**Please note that at this editing stage the Abstract QA field actually has two values, even if only one was set**: "_No selection_" and "_Yes_":

Selection_014

The day after, they want to QA the remaining 20 so they want to filter out all the other 80. They:

1. go to Manage Abstracts
2. click on _Customise list_
3. try to select the one that still have NOT selected a Yes in Abstract QA:
Selection_014
4. realise that it's **not possible to only list those that have NOT the "Yes" value in this field**.

**Expected behavior**
Have a "No selection" option in the filter.

**Additional context**
Please note that the logic of other multi-value fields is not correct either: as an example, consider the abstracts state.
By default **all abstracts are shown**. By entering the "Customise list" panel and opening the _State_ field, **no options are selected** which should mean "do not show any". If I select one, I only get that state. Idem for two. If I select all of them, I get back "any state", like when no state is selected. This means we have two possibilities for one filter (any) and no possibilities for another (no value for that field).

**Expected behavior**
By default that field filter should have *all options* selected. By deselecting all of them we'll get the abstracts with "no value" for that field. Or, different option, include a fictitious _value_  "no selection" like in the abstracts editing. 
The two areas should at least be consistent.

tomasr8:
Closes #5033 

Adds an extra filter option 'No selection' to custom abstract fields in the abstract list filter view.
This can be used to filter abstracts where no value for the given field has been selected.

image

### DIFF
--- a/indico/modules/events/abstracts/lists.py
+++ b/indico/modules/events/abstracts/lists.py
@@ -84,7 +84,7 @@ class AbstractListGeneratorBase(ListGeneratorBase):
         filters = super()._get_filters_from_request()
         for field in self.event.contribution_fields:
             if field.field_type == 'single_choice':
-                options = request.form.getlist(f'field_{field.id}')
+                options = [x if x != 'None' else None for x in request.form.getlist(f'field_{field.id}')]
                 if options:
                     filters['fields'][str(field.id)] = options
         return filters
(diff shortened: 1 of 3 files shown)


tomasr8:
There's a small issue when you already have some abstracts and then you
add a custom field. The old abstracts will not show up when you filter by 'No selection',
because `Abstract.field_values` does not get updated when a new custom field is added.

indico/modules/events/abstracts/lists.py
```diff
@@ -115,10 +115,23 @@ def _filter_list_entries(self, query, filters):
...
-                )))
+                )
+
+                # Support filtering by 'No selection' in single-choice abstract fields
+                if '' in field_values:
+                    field_value_criteria = db.or_(
+                        field_value_criteria,
+                        db.and_(
+                            AbstractFieldValue.contribution_field_id == contribution_type_id,
+                            AbstractFieldValue.data.op('#>>')('{}').is_(None)
```
ThiefMaster, line 130: Does this work if there is literally no data for the field? For example, a field that was created *after* submitting the abstract?

indico/modules/events/contributions/models/fields.py
```diff
@@ -131,7 +131,9 @@ def is_public(self):
 
     @property
     def filter_choices(self):
-        return {x['id']: x['option'] for x in self.field_data.get('options', {})}
+        choices = {x['id']: x['option'] for x in self.field_data.get('options', {})}
+        choices[''] = _('No selection')
```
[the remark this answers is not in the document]
ThiefMaster: Seems fine, I'd just use something like `_None` instead of the empty string to be a bit clearer ;)

indico/modules/events/abstracts/lists.py
ThiefMaster: I think this is exactly what you commented as well: I think we should handle it by making the query succeed if there is no AbstractFieldValue (ie NOT by updating existing abstracts when adding a field)

tomasr8: Yes, that was exactly my question. OK, I'll update the query :+1:

t4nec0:
> There's a small issue when you already have some abstracts and then you
> add a custom field. The old abstracts will not show up when you filter by 'No selection',
> because `Abstract.field_values` does not get updated when a new custom field is added.
> Should we handle that case?

I think this particular case **should** be handled. From the user point of view, old or new abstracts look the same, they all have no values for these fields (in particular the old ones!). Users have no clue about the data structure and how it's managed...

ThiefMaster:
He already updated it to handle this case...

PR #5037, opened 2021-08-12
Add after_registration_form_clone signal to `RegistrationFormCloner`

vasantvohra:


indico/core/signals/event/registration.py
```diff
@@ -104,3 +104,9 @@
 dictionary.  ``'status'`` is expected to be either ``'error'``, ``'warning'`` or
 ``ok``.
 ''')
+
+after_registration_form_clone = _signals.signal('after-registration-form-clone', """
+Executed after the registration form is cloned. The sender is the old ``RegistrationForm`` object being cloned.
+The new ``RegistrationForm`` object is passed in the ``new_form`` kwarg.
+The additional kwargs passed to this signal depend on the context.
```
ThiefMaster, line 111: there is no such context here, so i'd remove that line

ThiefMaster (approved):

PR #5045, opened 2021-08-18
Add "invite-options" template hook

vasantvohra:
Working on a new feature at the UN plugin side, which allows importing and sending invitations in bulk using excel. This requires the addition of buttons on the core template. Therefore, I'd like to add a template hook there.

Let me know if any changes are required on this PR.

Here are some screenshots for reference :)
Screenshot 2021-08-18 at 6 16 27 PM

Screenshot 2021-08-18 at 6 17 16 PM

### DIFF
--- a/indico/modules/events/registration/templates/management/regform_invitations.html
+++ b/indico/modules/events/registration/templates/management/regform_invitations.html
@@ -21,6 +21,7 @@
                     </div>
                 </div>
                 <div class="toolbar">
+                    {{ template_hook('registration-invite-options', event=event, regform=regform) }}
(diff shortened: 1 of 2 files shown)


indico/modules/events/registration/templates/management/regform_invitations.html
```diff
@@ -21,6 +21,7 @@
                     </div>
                 </div>
                 <div class="toolbar">
+                    {{ template_hook('invite-options', event=event, regform=regform) }}
```
ThiefMaster, line 24:
```suggestion
                    {{ template_hook('registration-invite-options', event=event, regform=regform) }}
```

Just to be clearer of the scope.

PR #5046, opened 2021-08-18
Improve event move confirmation UI + add move request reason

### LINKED ISSUE #5041: Category moderation: Improve event move UI
Right now we just show the category picker and then a confirmation prompt and that's it. The user doesn't even know if the move needs approval or not.

Let's replace that confirmation prompt with a new dialog where we show whether it's an immediate move or just a moderated move request. In case of the latter, we can also add a field where the user can provide a comment/justification.

Note: Category managers can move events in bulk from the category management content view!

ThiefMaster:
image

image

closes #5041

### DIFF
--- a/indico/modules/categories/client/js/management.js
+++ b/indico/modules/categories/client/js/management.js
@@ -9,7 +9,10 @@
           enableIfChecked:false, build_url:false, ajaxDialog:false, updateHtml:false */
 
 import _ from 'lodash';
+import React from 'react';
+import ReactDOM from 'react-dom';
 
+import {SingleEventMove, BulkEventMove} from 'indico/modules/events/management/EventMove';
 import {showUserSearch} from 'indico/react/components/principals/imperative';
 import {$T} from 'indico/utils/i18n';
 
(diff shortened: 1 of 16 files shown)


indico/modules/categories/client/js/management.js
```diff
@@ -308,6 +279,47 @@ import {$T} from 'indico/utils/i18n';
         },
       },
     }).isEverythingSelected;
+
+    _fetchSourceCategory(categoryId);
+
+    [].forEach.call(
```
javfg, line 285: Is (and line 306) this for browser compatibility? Is it worth it? (Not sure which browser versions started supporting NodeList forEach).

javfg (approved):

indico/modules/categories/client/js/management.js
ThiefMaster: oh nice, I didn't know they added forEach to NodeList. We didn't really bump the browser version requirement with the 3.0 release so I think for now I'll stay on the safe side and keep it..

https://caniuse.com/mdn-api_nodelist_foreach

PR #5051, opened 2021-08-24
Improve Category moderation UI

### LINKED ISSUE #5040: Category moderation: Improve moderation UI
- Instead of showing the date of the request in `yyyy-mm-dd`, show it in a nicer way (and including the time?). We could even go for a relative time (moment `fromNow`) and keep the absolute time in the `title`.
- Show the current category path of the event
- Show the start/end time of the event

tomasr8:
Closes #5040 

- Uses nicer dates
- Adds the original event path and date

image


indico/modules/categories/client/js/components/CategoryModeration.jsx
```diff
@@ -16,16 +17,26 @@ import {handleSubmitError} from 'indico/react/forms';
...
 
+function getEventTime({startDt, endDt}) {
+  startDt = moment(startDt);
+  endDt = moment(endDt);
+
+  if (startDt.isSame(endDt, 'day')) {
+    return `${startDt.format('D MMM YYYY, hh:mm')} → ${endDt.format('hh:mm')}`;
+  } else {
+    return `${startDt.format('D MMM YYYY, hh:mm')} → ${endDt.format('D MMM YYYY, hh:mm')}`;
+  }
```
ThiefMaster, line 30: Can we avoid hardcoding date/time formats? This is somewhat messy when it comes to l10n... not all languages use that particular date (or time) format.

indico/modules/categories/client/js/components/CategoryModeration.jsx
```diff
@@ -69,8 +85,11 @@ RequestTableRow.propTypes = {
     id: PropTypes.number,
     title: PropTypes.string,
   }).isRequired,
+  category: PropTypes.shape({
+    id: PropTypes.number,
+    chainTitles: PropTypes.arrayOf(PropTypes.string),
```
ThiefMaster, line 90:
```suggestion
    id: PropTypes.number.isRequired,
    chainTitles: PropTypes.arrayOf(PropTypes.string).isRequired,
```

indico/modules/categories/client/js/components/CategoryModeration.jsx
```diff
@@ -103,38 +122,47 @@ function RequestList({requests, onSubmit, loading}) {
...
+                requestedDt={requestedDt}
+                selected={selected.has(id)}
+                onSelect={() => select(id)}
+              />
+              {requestorComment ? (
+                <Table.Row>
+                  <Table.Cell />
+                  <Table.Cell colSpan="5">{requestorComment}</Table.Cell>
+                </Table.Row>
+              ) : null}
```
ThiefMaster, line 157: Shouldn't this be part of the `RequestTableRow` component? (ie render either one or two rows from within that component)

indico/modules/categories/schemas.py
```diff
@@ -6,15 +6,22 @@
...
 
 from indico.core.marshmallow import fields, mm
+from indico.modules.categories import Category
 from indico.modules.categories.models.event_move_request import EventMoveRequest
 from indico.modules.events import Event
 from indico.modules.users import User
 
 
-class CategoryEventSchema(mm.SQLAlchemyAutoSchema):
+class EventSchema(mm.SQLAlchemyAutoSchema):
```
ThiefMaster, line 15: Maybe `MoveRequestEventSchema` fits better since that's what we are using it for?

tomasr8: Renamed :+1:

indico/modules/categories/client/js/components/CategoryModeration.jsx
tomasr8: I changed this to `L` & `LT`

indico/modules/categories/client/js/components/CategoryModeration.jsx
```diff
@@ -37,45 +48,72 @@ function PlaceholderTableRow() {
...
+            {fullName} {affiliation && `(${affiliation})`}
+          </span>
+        </Table.Cell>
+        <Table.Cell>
+          <a href={eventURL({event_id: event.id})}>{event.title}</a>
+        </Table.Cell>
+        <Table.Cell>{category.chainTitles.join(' » ')}</Table.Cell>
+        <Table.Cell>{getEventTime(event)}</Table.Cell>
+        <Table.Cell>
+          <span title={moment(requestedDt).format('D MMM YYYY, hh:mm')}>
```
ThiefMaster, line 79: same l10n problem as we had above ;)

indico/modules/categories/client/js/components/CategoryModeration.jsx
```diff
@@ -16,16 +17,26 @@ import {handleSubmitError} from 'indico/react/forms';
 import {useIndicoAxios} from 'indico/react/hooks';
 import {Translate} from 'indico/react/i18n';
 import {indicoAxios} from 'indico/utils/axios';
-import {serializeDate} from 'indico/utils/date';
 import './CategoryModeration.module.scss';
 
+function getEventTime({startDt, endDt}) {
```
ThiefMaster, line 22: I think we need to finally figure out why the moment locale isn't always taken into account... because right now this one always uses the `en` locale, so it's using 12h format even though almost none of our current locales in Indico use that format...

ThiefMaster (approved):

PR #5057, opened 2021-08-25
Category moderation  - Move request email notifications

### LINKED ISSUE #5038: Category moderation: Requestor notification emails
When a request gets approved/rejected (or maybe withdrawn implicitly by moving the event elsewhere?), the original requestor of the move should get an email notification about this.

### LINKED ISSUE #5039: Category moderation: Manager notification emails
Category managers should be notified about new/withdrawn requests by email. Let's use the same logic as we do in `notify_event_creation` but maybe only climb up the chain until the first category that has managers or notification emails set?

An easier alternative would be to *just* take managers of the category itself into account; in that case we should show a clear warning in case the category has moderation enabled (either via ACL entries or via the creation mode setting) but no managers.

tomasr8:
Closes #5038
Closes #5039 

Sends an email notification to the requestor when a move request is accepted or rejected and
sends an email notification to category managers when a move request is created.

### DIFF
--- a/indico/modules/events/notifications.py
+++ b/indico/modules/events/notifications.py
@@ -4,6 +4,7 @@
 # Indico is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
+
 import itertools
 
 from sqlalchemy.orm import joinedload
@@ -84,4 +85,4 @@ def notify_move_request_closure(move_requests, accept, reason=''):
         events = [rq.event for rq in requests]
         template = get_template_module('events/emails/move_request_closure.txt',
                                        events=events, accept=accept, reason=reason)
-        send_email(make_email(bcc_list=requestor.email, template=template))
+        send_email(make_email(to_list=requestor.email, template=template))
--- a/indico/modules/events/notifications.py
+++ b/indico/modules/events/notifications.py
@@ -6,6 +6,7 @@
 # LICENSE file for more details.
 
 import itertools
+from operator import attrgetter
 
 from sqlalchemy.orm import joinedload
 
@@ -80,9 +81,9 @@ def notify_move_request_closure(move_requests, accept, reason=''):
     :param reason: Optional reason for rejection.
     """
 
-    move_requests = sorted(move_requests, key=lambda rq: rq.requestor.id)
-    for requestor, requests in itertools.groupby(move_requests, lambda rq: rq.requestor):
+    move_requests = sorted(move_requests, key=attrgetter('requestor.id', 'category.id'))
+    for (requestor, category), requests in itertools.groupby(move_requests, attrgetter('requestor', 'category')):
         events = [rq.event for rq in requests]
         template = get_template_module('events/emails/move_request_closure.txt',
-                                       events=events, accept=accept, reason=reason)
+                                       events=events, target_category=category, accept=accept, reason=reason)
         send_email(make_email(to_list=requestor.email, template=template))
--- a/indico/modules/events/notifications.py
+++ b/indico/modules/events/notifications.py
@@ -14,6 +14,19 @@ from indico.modules.categories.models.categories import Category
 from indico.web.flask.templating import get_template_module
 
 
+def _get_emails_from_category(category, ignore_notify_managers=False):
+    emails = set(category.event_creation_notification_emails)
+
+    if category.notify_managers or ignore_notify_managers:
+        for manager in category.get_manager_list():
+            if manager.principal_type in (PrincipalType.user, PrincipalType.email):
+                emails.add(manager.email)
+            elif manager.principal_type in (PrincipalType.local_group, PrincipalType.multipass_group):
+                emails.update(x.email for x in manager.get_members())
+
+    return emails
+
+
 def notify_event_creation(event, occurrences=None):
     """Send email notifications when a new Event is created.
 
@@ -28,20 +41,37 @@ def notify_event_creation(event, occurrences=None):
              filter(Category.notify_managers | (Category.event_creation_notification_emails != []))
              .options(joinedload('acl_entries')))
     for cat in query:
-        emails.update(cat.event_creation_notification_emails)
-        if cat.notify_managers:
-            for manager in cat.get_manager_list():
-                if manager.principal_type in (PrincipalType.user, PrincipalType.email):
-                    emails.add(manager.email)
-                elif manager.principal_type in (PrincipalType.local_group, PrincipalType.multipass_group):
-                    emails.update(x.email for x in manager.get_members())
+        emails.update(_get_emails_from_category(cat))
 
     if emails:
         template = get_template_module('events/emails/event_creation.txt', event=event, occurrences=occurrences)
         send_email(make_email(bcc_list=emails, template=template))
 
 
-def notify_move_request(move_requests, accept, reason=''):
+def notify_move_request_creation(events, target_category, comment=''):
+    """Send email notifications when a move request is created.
+
+    :param events: List of events requested to move.
+    :param target_category: The category to move the events into.
+    :param comment: Optional requestor's comment.
+    """
+    emails = set()
+    query = (target_category.chain_query
+             .filter(Category.notify_managers | (Category.event_creation_notification_emails != []))
+             .options(joinedload('acl_entries')))
+
+    # Walk up the category chain
+    for cat in reversed(query.all()):
+        emails.update(_get_emails_from_category(cat, ignore_notify_managers=True))
+
+        if emails:
+            template = get_template_module('events/emails/move_request_creation.txt',
+                                           events=events, target_category=target_category, comment=comment)
+            send_email(make_email(bcc_list=emails, template=template))
+            break
+
+
+def notify_move_request_closure(move_requests, accept, reason=''):
     """Send email notifications when a move request is accepted/rejected.
 
     :param move_requests: List of `EventMoveRequest` that were accepted/rejected.
--- a/indico/modules/events/notifications.py
+++ b/indico/modules/events/notifications.py
@@ -4,6 +4,7 @@
 # Indico is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
+import itertools
 
 from sqlalchemy.orm import joinedload
 
@@ -38,3 +39,19 @@ def notify_event_creation(event, occurrences=None):
     if emails:
         template = get_template_module('events/emails/event_creation.txt', event=event, occurrences=occurrences)
         send_email(make_email(bcc_list=emails, template=template))
+
+
+def notify_move_request(move_requests, accept, reason=''):
+    """Send email notifications when a move request is accepted/rejected.
+
+    :param move_requests: List of `EventMoveRequest` that were accepted/rejected.
+    :param accept: Whether the requests were accepted.
+    :param reason: Optional reason for rejection.
+    """
+
+    move_requests = sorted(move_requests, key=lambda rq: rq.requestor.id)
+    for requestor, requests in itertools.groupby(move_requests, lambda rq: rq.requestor):
+        events = [rq.event for rq in requests]
+        template = get_template_module('events/emails/move_request_closure.txt',
+                                       events=events, accept=accept, reason=reason)
+        send_email(make_email(bcc_list=requestor.email, template=template))
--- a/indico/modules/events/notifications_test.py
+++ b/indico/modules/events/notifications_test.py
@@ -20,7 +20,7 @@ from indico.web.flask.templating import get_template_module
 ))
 def test_move_request_creation_email_plaintext(snapshot, create_category, create_event, num_events, comment,
                                                snapshot_name):
-    cat = create_category(title='target')
+    cat = create_category(id_=314, title='target')
     events = [create_event(id_=n, title=f'test {n}') for n in range(1, num_events + 1)]
     template = get_template_module('events/emails/move_request_creation.txt',
                                    events=events, target_category=cat, comment=comment)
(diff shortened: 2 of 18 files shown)


indico/modules/events/notifications.py
```diff
@@ -38,3 +40,18 @@ def notify_event_creation(event, occurrences=None):
...
+
+def notify_move_request(move_requests, accept, reason=''):
+    """Send email notifications when a move request is accepted/rejected.
+
+    :param move_requests: List of `EventMoveRequest` that were accepted/rejected.
+    :param accept: Whether the requests were accepted.
+    :param reason: Optional reason for rejection.
+    """
+
+    for email, requests in itertools.groupby(move_requests, lambda rq: rq.requestor.email):
```
ThiefMaster, line 53: group by the user itself, not their email

ThiefMaster: also, you need to make sure the list is sorted for this to work. you can use our `group_list` util that combines sorting and grouping

indico/modules/events/templates/emails/move_request_closure.txt
```diff
@@ -0,0 +1,31 @@
...
+
+{% block subject -%}
+    Event move request update
+{%- endblock %}
+
+{% block header -%}{%- endblock %}
+
+{% block body -%}
+    {%- filter dedent -%}
+        {{ ('The following events have been moved:' if accept else 'Move requests of the following events have been rejected:' ) }}
```
ThiefMaster, line 11: The parentheses aren't needed, but I'd rather use normal if/else blocks here...

indico/modules/events/templates/emails/move_request_closure.txt
```diff
@@ -0,0 +1,31 @@
...
+
+        {% filter underline %}Events{% endfilter %}
+        {%- for evt in events %}
+            {{ evt.title }}
+            - URL: {{ evt.external_url }}{% if accept %}, New event path: {{ evt.category.chain_titles|join(' » ') }}{% endif %}
```
ThiefMaster, line 16: I'd just show the titles here, and maybe the URL behind it in parentheses. Having one line per event is going to be much more readable in a plaintext email.

ThiefMaster, line 16: just FYI, jinja supports trimming strings on any side of a tag, so you don't have to put so much in one line. e.g. `{%- if accept %}` would strip any whitespace before the `if` tag

indico/modules/events/templates/emails/move_request_closure.txt
```diff
@@ -0,0 +1,31 @@
...
+        {% endfor %}
+
+        {%- if not accept %}
+            {% filter underline %}Reason for rejection{% endfilter %}
```
ThiefMaster, line 20: Rejection reason

indico/modules/events/templates/emails/move_request_closure.txt
```diff
@@ -0,0 +1,31 @@
...
+            {{ reason }}
+        {%- endif -%}
+
```
ThiefMaster, line 23: 👾

indico/modules/events/notifications.py
```diff
@@ -27,14 +42,46 @@ def notify_event_creation(event, occurrences=None):
...
+
+def notify_move_request_creation(events, target_category, comment=''):
+    """Send email notifications when a move request is created.
+
+    :param events: List of events requested to move.
+    :param target_category: The category to move the events into.
+    :param comment: Optional requestor's comment.
+    """
+    emails = set()
+    query = (target_category.chain_query.
```
ThiefMaster, line 60: that `.` should be in the next line before `filter` (please don't fix it in `notify_event_creation` above, the unlisted events PR already does that)

indico/modules/events/notifications.py
```diff
@@ -27,14 +42,46 @@ def notify_event_creation(event, occurrences=None):
...
+             filter(Category.notify_managers | (Category.event_creation_notification_emails != []))
+             .options(joinedload('acl_entries')))
+
+    # Walk up the category chain
+    for cat in reversed([cat for cat in query]):
```
ThiefMaster, line 65: Why reversed - won't that start with the root category now?

indico/modules/events/notifications.py
```diff
@@ -13,6 +15,19 @@
 from indico.web.flask.templating import get_template_module
 
 
+def _get_emails_from_category(category):
+    emails = set(category.event_creation_notification_emails)
+
+    if category.notify_managers:
```
ThiefMaster, line 21: I think we should ignore this setting when it comes to move request notifications...

indico/modules/events/notifications.py
tomasr8: The `chain_query` docstring says it returns the root first and then the intermediate categories,
so I think we need to reverse that in order to walk up from the current category

ThiefMaster: ah good point, I was just looking at the query above but of course there the order doesn't matter so you are right...

How about this though? :)

`for cat in query.all()[::-1]`

tomasr8: How about `reversed(query.all())`? I think I like that more than `[::-1]` :)

ThiefMaster: sure

indico/modules/events/notifications.py
```diff
@@ -27,14 +41,47 @@ def notify_event_creation(event, occurrences=None):
...
+    :param accept: Whether the requests were accepted.
+    :param reason: Optional reason for rejection.
+    """
+
+    move_requests = sorted(move_requests, key=lambda rq: rq.requestor.id)
+    for requestor, requests in itertools.groupby(move_requests, lambda rq: rq.requestor):
+        events = [rq.event for rq in requests]
+        template = get_template_module('events/emails/move_request_closure.txt',
+                                       events=events, accept=accept, reason=reason)
+        send_email(make_email(bcc_list=requestor.email, template=template))
```
ThiefMaster, line 87: There's no need to use bcc here - it goes directly to a single user -> use `to_list`

indico/modules/events/templates/emails/move_request_closure.txt
```diff
@@ -0,0 +1,33 @@
+{% extends 'emails/base.txt' %}
+
+{% block subject -%}
+    Event move request update
```
ThiefMaster, line 4: Maybe we could put "accepted"/"rejected" instead of "update" here?

indico/modules/events/templates/emails/move_request_closure.txt
```diff
@@ -0,0 +1,33 @@
...
+        {% else %}
+            Move requests of the following events have been rejected:
+        {% endif %}
+
+        {% filter underline %}Events{% endfilter %}
+        {%- for evt in events %}
+            {{ evt.title }} ({{ evt.external_url }}) {# -#}
+        {% endfor %}
+
+        {% if not accept %}
```
ThiefMaster, line 22:
```suggestion
        {% if not accept and reason %}
```

the reason is optional

indico/modules/events/templates/emails/move_request_creation.txt
```diff
@@ -0,0 +1,28 @@
...
+        There are new move requests for the category:
+        {{ target_category.chain_titles|join(' » ') }}
+
+        {% filter underline %}Requested events{% endfilter %}
+        {%- for evt in events %}
+            {{ evt.title }} ({{ evt.external_url }})
+        {% endfor %}
+
+        {% filter underline %}Comment{% endfilter %}
+        {{ comment }}
```
ThiefMaster, line 20: Let's hide this if there is no comment

indico/modules/events/templates/emails/move_request_creation.txt
```diff
@@ -0,0 +1,28 @@
...
+{% block body -%}
+    {%- filter dedent -%}
+        There are new move requests for the category:
+        {{ target_category.chain_titles|join(' » ') }}
+
+        {% filter underline %}Requested events{% endfilter %}
+        {%- for evt in events %}
+            {{ evt.title }} ({{ evt.external_url }})
+        {% endfor %}
+
```
ThiefMaster, line 18: This template is missing a link to the moderation page

indico/modules/events/templates/emails/move_request_closure.txt
```diff
@@ -0,0 +1,33 @@
...
+
+{% block body -%}
+    {%- filter dedent -%}
+        {% if accept %}
+            The following events have been moved:
+        {% else %}
+            Move requests of the following events have been rejected:
+        {% endif %}
+
+        {% filter underline %}Events{% endfilter %}
```
ThiefMaster, line 17: IMHO the underlined "Events" header looks a bit strange. I'd get rid of it and put `- ` in front of each event entry.

```
The following events have been moved:

- MOAR EVENTS (https://.../event/1056189/) 
- mod test (https://.../event/1056185/) 
```

vs

```
The following events have been moved:


Events
------
MOAR EVENTS (https://.../event/1056189/) 
mod test (https://.../event/1056185/) 
```

indico/modules/events/templates/emails/move_request_closure.txt
```diff
@@ -0,0 +1,33 @@
...
+
+{% block header -%}{%- endblock %}
+
+{% block body -%}
+    {%- filter dedent -%}
+        {% if accept %}
+            The following events have been moved:
+        {% else %}
+            Move requests of the following events have been rejected:
+        {% endif %}
```
ThiefMaster, line 15: Maybe we could use different wording if there's just one event... (that's going to be the most common case after all!)

indico/modules/events/templates/emails/move_request_closure.txt
ThiefMaster, line 22: We probably need to collapse some of the space here so there's an equal amount of blank lines before the signature regardless of whether there's a reason or not

indico/modules/events/notifications_test.py
```diff
@@ -0,0 +1,46 @@
...
+    (False, 2, 'We do not want your event.\nSorry not sorry.', 'move-request-rejected-multi-reason.txt'),
+    (True, 1, '', 'move-request-accepted-single.txt'),
+    (True, 2, '', 'move-request-accepted-multi.txt'),
+))
+def test_move_request_closed_email_plaintext(snapshot, create_category, create_event, accept, num_events, reason,
+                                             snapshot_name):
+    cat = create_category(title='target')
+    events = [create_event(title=f'test {n + 1}') for n in range(num_events)]
+    template = get_template_module('events/emails/move_request_closure.txt',
+                                   events=events, target_category=cat, accept=accept, reason=reason)
```
ThiefMaster, line 44: I'm already including the target category here since I think it'd make sense to mention it in this particular email as well...

indico/modules/events/notifications.py
```diff
@@ -4,6 +4,7 @@
 # Indico is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
+import itertools
```
ThiefMaster, line 7:
```suggestion

import itertools
```

indico/modules/events/templates/emails/move_request_closure.txt
```diff
@@ -0,0 +1,48 @@
...
+        {%- else -%}
+            {% if events|length == 1 %}
+                Request to move event rejected:
+            {% else %}
+                Move requests of the following events have been rejected:
+            {% endif %}
+        {%- endif %}
+
+        {%- for evt in events %}
+            - {{ evt.title }} ({{ evt.external_url }}){# -#}
```
ThiefMaster, line 30: You can use `{%- endfor %}` in the next line to do the same thing

indico/modules/events/templates/emails/move_request_closure.txt
```diff
@@ -0,0 +1,48 @@
...
+        {% endfor %}
+
+        {% filter underline %}{{ 'New' if accept else 'Requested' }} category:{% endfilter %}
```
ThiefMaster, line 33: The `:` feels a bit weird for an underlined heading

indico/modules/events/notifications_test.py
```diff
@@ -0,0 +1,46 @@
...
+    (False, 2, 'We do not want your event.\nSorry not sorry.', 'move-request-rejected-multi-reason.txt'),
+    (True, 1, '', 'move-request-accepted-single.txt'),
+    (True, 2, '', 'move-request-accepted-multi.txt'),
+))
+def test_move_request_closed_email_plaintext(snapshot, create_category, create_event, accept, num_events, reason,
+                                             snapshot_name):
+    cat = create_category(id_=314, title='target')
+    events = [create_event(id_=n, title=f'test {n}') for n in range(1, num_events + 1)]
+    template = get_template_module('events/emails/move_request_closure.txt',
+                                   events=events, target_category=cat, accept=accept, reason=reason)
```
ThiefMaster, line 44: Let's remove the `target_category` kwarg here since we don't use it in the end, and instead create the events 2 lines above in that category.

indico/modules/events/templates/emails/move_request_creation.txt
```diff
@@ -0,0 +1,33 @@
...
+{% block subject -%}
+    Event move request created
+{%- endblock %}
+
+{% block header -%}{%- endblock %}
+
+{% block body -%}
+    {%- filter dedent -%}
+        There are new move requests for the category:
```
ThiefMaster, line 11: i'd change this to singular in case there's just a single event...

indico/modules/events/templates/emails/move_request_creation.txt
```diff
@@ -0,0 +1,33 @@
...
+            {{ evt.title }} ({{ evt.external_url }})
+        {%- endfor -%}
+
+        {% if comment %}
+
+            {% filter underline %}Comment{% endfilter %}
+            {{ comment }}
+        {%- endif %}
+
+        View all move requests here: {{ url_for('categories.manage_moderation', target_category, _external=true) }}
```
ThiefMaster, line 25: I'd move the URL to a new next line

indico/modules/events/notifications_test.py
tomasr8: We use the `target_category` in the template:

```python
{% filter underline %}{{ 'New' if accept else 'Requested' }} category{% endfilter %}
{{ target_category.chain_titles|join(' » ') -}}
```

ThiefMaster: But that's the "closure" one - you use `event.category` there, don't you?

tomasr8: If we want to always show the target category in the closure email, then we have to use `target_category` I think,
since if the move is rejected, `event.category` points to the original one.

ThiefMaster (approved):

PR #5058, opened 2021-08-25
Show pending event move requests in more places

### LINKED ISSUE #5056: Category moderation: Indicate pending requests
Maybe show a message_box on the category display view if the user is a manager and there are pending events.

We could also look into allowing to add a badge/label with a number on the "Moderation" menu entry to show the number of pending requests there.

ThiefMaster:
image

image

closes #5056


tomasr8 (approved):
Tested locally and works great :+1:

PR #5061, opened 2021-08-27
Vendor colorclass

### LINKED ISSUE #4862: Replace colorclass with something else
The last release was [in 2016](https://pypi.org/project/colorclass/#history), the upstream repo has been [archived](https://github.com/Robpol86/colorclass), and it will break in Python 3.10:

```
.../env/lib/python3.9/site-packages/colorclass/codes.py:4: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
  from collections import Mapping
```

The only reason we use it is because it provides a `__len__` on its color string objects that returns the visible length (not counting the color control codes); this is needed e.g. when using colored strings in `terminaltables`.

---

Possible alternatives:

```
<dude-x_> ThiefMaster do you know  will mcGugan's `rich` library?
<dude-x_> thats actively maintained
<SnoopJ> `colorama` and `blessings` are also well-liked
```

I don't think `colorama` is an option since it doesn't have the length magic we need (neither has `blessings`), but [`rich`](https://github.com/willmcgugan/rich) looks very fancy and feature-loaded; it could probably replace our `cformat()` util and maybe even `terminaltables` as well.

Important utils to test to make sure format codes are not included in the length: `terminaltables`, `verbose_iterator`

In case `rich` doesn't work either, we could probably vendor in just the parts of `colorclass` that we need (ie a simple `str` wrapper that implements the usual string utils in a way that works well with colored strings, or simply fork it and fix the compatibility issue.

tomasr8:
Closes #4862 

Incorporates the archived [colorclass](https://github.com/Robpol86/colorclass) library release 2.2.0 into Indico.
Unused code and compatibility with python2 is removed.

### DIFF
--- a/indico/util/benchmark.py
+++ b/indico/util/benchmark.py
@@ -8,7 +8,7 @@
 import time
 from math import isinf
 
-from indico.util.console import cformat
+import click
 
 
 class Benchmark:
@@ -58,12 +58,12 @@ class Benchmark:
     def print_result(self, slow=float('inf'), veryslow=float('inf')):
         duration = float(self)
         if duration == float('-inf'):
-            print(cformat('%{blue!}skipped'))
+            click.secho('skipped', fg='blue')
         elif duration == float('inf'):
-            print(cformat('%{red}running'))
+            click.secho('running', fg='red')
         elif duration >= veryslow:
-            print(cformat('%{red!}{}').format(self))
+            click.secho(str(self), fg='red', bold=True)
         elif duration >= slow:
-            print(cformat('%{yellow!}{}').format(self))
+            click.secho(str(self), fg='yellow', bold=True)
         else:
-            print(cformat('%{green!}{}').format(self))
+            click.secho(str(self), fg='green', bold=True)
--- a/indico/util/console.py
+++ b/indico/util/console.py
@@ -15,7 +15,6 @@ from datetime import timedelta
 
 import click
 from click.types import convert_type
-from termcolor import colored
 
 from indico.util.date_time import format_human_timedelta
 from indico.util.string import validate_email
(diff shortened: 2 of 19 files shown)


indico/util/console.py
```diff
@@ -104,17 +103,18 @@ def _print_text(text):
 
 
 def _cformat_sub(m):
-    bg = 'on_{}'.format(m.group('bg')) if m.group('bg') else None
-    attrs = ['bold'] if m.group('fg_bold') else None
-    return colored('', m.group('fg'), bg, attrs=attrs)[:-4]
+    bold = '{b}' if m.group('fg_bold') else ''
+    bg = f"{{bg{m.group('bg')}}}" if m.group('bg') else ''
+    fg = f"{{{m.group('fg')}}}"
```
ThiefMaster, line 108: I think this is one of the cases where f-strings do not improve readability (due to the need to escape the literal `{...}` in the string as `{{...}}`) - I'd consider using `%` formatting there, ie something like this:

```python
    bg = bold = ''
    if m.group('fg_bold'):
        bold = '{b}'
    if bg_color := m.group('bg'):
        bg = '{bg%s}' % bg_color
    fg = '{%s}' % m.group('fg')
    return Color(f'{bold}{bg}{fg}')
```

indico/util/benchmark.py
```diff
@@ -58,12 +58,12 @@ def __str__(self):
...
         duration = float(self)
         if duration == float('-inf'):
-            print(cformat('%{blue!}skipped'))
+            click.secho('skipped', fg='blue')
         elif duration == float('inf'):
-            print(cformat('%{red}running'))
+            click.secho('running', fg='red')
         elif duration >= veryslow:
-            print(cformat('%{red!}{}').format(self))
+            click.secho(f'{self}', fg='red', bold=True)
```
ThiefMaster, line 65: `click.secho(str(self), ...)` if you need an explicit conversion to str (same below)

indico/vendor/colorclass/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# LICENSE file for more details.
+
+# The code is taken from the colorclass library, release 2.2.0,
+# which is licensed under the MIT license and is originally
+# available on the following URL:
+# https://github.com/Robpol86/colorclass
+# Credits of the original code go to Robpol86 (https://github.com/Robpol86)
+# and other colorclass contributors.
+
+from .color import Color  # noqa
```
ThiefMaster, line 15:
```suggestion
from .color import Color  # noqa: F401
```

ThiefMaster (commented):
Let's add `indico/vendor/* linguist-vendored` to `.gitattributes` so GitHub knows this is vendored code and collapses it e.g. in PR diffs by default

indico/util/console.py
tomasr8: Just when I thought I'd never use %-formatting again.. :D

PR #5068, opened 2021-09-01
Fix duplicate imports of the moment library

### LINKED ISSUE #4538: `moment()` sometimes uses an instance with no locale configured
This happens e.g. in the event cloning dialog when previewing the dates on which the cloned events will take place when cloning with a pattern. Since it uses the default `en` locale it's using `mm/dd/yyyy` format instead of `dd/mm/yyyy` or whatever the locale uses.

`window.moment !== moment` in there, and when checking with the browser's dev tools the non-global moment is coming from another scope, so it's most likely caused by the webpack ProvidesPlugin we're using:

```js
      new webpack.ProvidePlugin({
        _: ['underscore', 'default'],
        moment: 'moment',
      }),
```

tomasr8:
Closes #4538 

For a reason yet unknown the webpack `ProvidePlugin` does not import a library as a singleton but sometimes
duplicates it several times. This causes problems when we need to set global properties on the moment object
such as moment.locale().

This PR removes the ProvidePlugin in favour of importing the library manually and
making it globally available via window.moment.

Since the library is available as a global, it should not be necessary to change any code as
the usage is identical with that of `ProvidePlugin`.

### DIFF
--- a/indico/web/client/js/jquery/index.js
+++ b/indico/web/client/js/jquery/index.js
@@ -12,6 +12,13 @@ import 'jquery-colorbox/example1/colorbox.css';
 
 import 'jquery-form';
 
+import moment from 'moment';
+// Make moment globally available.
+// Previously, we used the webpack `ProvidePlugin`, but instead of having
+// a single shared instance of moment, this resulted in duplicate instances,
+// which didn't have the correct locale as set in `indico_base.html`.
+window.moment = moment;
+
 // moment.js locales
 import 'moment/locale/tr';
 import 'moment/locale/mn';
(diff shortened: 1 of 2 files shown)


indico/web/client/js/jquery/extensions/global.js
```diff
@@ -9,6 +9,13 @@
 /* global showFormErrors:false */
 /* eslint-disable max-len */
 
+import moment from 'moment';
```
tomasr8, line 12: Not sure if this is the right file to put this..

ThiefMaster:
Does it also work if you put it in indico/web/client/js/jquery/index.js? Because in there we have all the other stuff like importing the moment locales...

indico/web/client/js/jquery/extensions/global.js
ThiefMaster: Does it also work if you put it in indico/web/client/js/jquery/index.js? Because in there we have all the other stuff like importing the moment locales...

plourenco (approved):

PR #5071, opened 2021-09-02
Localize date intervals

### LINKED ISSUE #5059: Localization: Date format in header should be locale dependent
**Describe the bug**
Currently, in event header, the date is formatted using the format_event_date macro.
This macro relies on English locale custom for formatting the event date, which may not be OK for non-English languages.

**To Reproduce**
1. Go to main page of an event.
2. Switch to a different language
3. See the date format in the event header stay the same, e.g. 1-13 listopada for Polish.

**Expected behavior**
The date format should reflect the language, e.g. probably be set by a translation template.
For Czech it would be 12.–14. září (for example).

**Additional context**
The macro in question is in modules/events/templates/display/conference/_util.html

tomasr8:
Closes #5059

Doing CTRL+F for `strftime` in the codebase, there are more instances of hardcoded
date formats, which might need some attention too. (Actually, all `format_*` functions too)

`Babel` includes `format_interval` which handles localization. After playing with the formats,
I propose two versions. Either we just call `format_interval` on both dates:

```python
{{ format_interval(start_dt, end_dt, format='dMMMMy') }}
``` 

or we replace only the day part:

```python
{{ format_interval(start_dt, end_dt, format='d') }} {{ start_dt | format_date('MMMM yyyy') }}
```
The table below shows the differences for some locales:

| Locale | `dMMMMy` | `d` + `MMMM yyyy` |
| --- | --- | --- |
| en_UK  | 15-17 Jan 2021 |  15–17 January 2021 |
| en_US | Jan 15 - 17, 2021 |  15 – 17 January 2021 |
| pl_PL | 15-17 stycznia 2021 | 15–17 stycznia 2021 |
| fr_FR | 15-17 janv. 2021 | 15–17 janvier 2021 |
| cs_CZ | 15.-17. 1. 2021 | 15.-17. ledna 2021 |

Update for localizing time formats: The default Babel formats always include a year,
which is sometimes not wanted. I figure we can use babel's `format_skeleton`, which should
try to adapt to the user's locale while being as close as possible to the format specified.

Here are some examples with a format used in indico:

| Locale | `format_skeleton(dt, format='EEEdMMM')` | `format_date(dt, format='EEE, d MMM')`
| --- | --- | --- |
| en_UK  | Fri, 3 Sep |  Fri, 3 Sep | 
| en_US | Fri, Sep 3 |  Fri, 3 Sep | 
| pl_PL | pt., 3 wrz | pt., 3 wrz |
| fr_FR | ven. 3 sept. | ven. 3 sept. |
| cs_CZ | pá 3. 9. | pá, 3 zář |


indico/web/flask/app.py
```diff
@@ -202,6 +202,7 @@ def setup_jinja(app):
     app.add_template_global(render_field, '_render_field')
     app.add_template_global(iter_form_fields, '_iter_form_fields')
     app.add_template_global(format_currency)
+    app.add_template_global(date_time_util.format_interval)
```
ThiefMaster, line 205: Why not make it a filter like the other date time utils? (btw if you do that please remove the incorrect "(indico functions returning UTF8)" comment)

ThiefMaster:
The `MMMM yyyy` part probably need to be translatable since it's not necessarily that same order in each language

indico/web/flask/app.py
tomasr8: Can I have a filter with two arguments?

ThiefMaster: yes, `foo|bar(x, ...)` basically calls `bar(foo, x, ...)`

ThiefMaster:
For this I wonder if we need to make the string itself translatable so the order becomes interchangeable...

something like `{% trans ... %}{{ day_interval }} {{ month_year }}{% endtrans %}` with the two variables being the babel-formatted dates

tomasr8:
`format_interval` is interesting.. For Polish it returns the whole month, but for Czech it just gives the number.
For me though, the Czech version is ok..

plourenco:
Aside from translating `MMMM yyyy`, I think we should keep `format_interval` as one in order to avoid having the order interchangeable and making it more complex to translate. IMO babel should handle this, because leaving it up to the translators may cause some opinionated results.

ThiefMaster:
If we are fine with the `dMMMMy` output I guess that would cover it nicely since it's locale-specific and thus removes the need to give translators the power to change the format?

plourenco:
> If we are fine with the `dMMMMy` output I guess that would cover it nicely since it's locale-specific and thus removes the need to give translators the power to change the format?

Exactly

indico/web/templates/forms/time_widget.html
```diff
@@ -4,7 +4,7 @@
     <div class="time-widget">
         {# the hidden field has autofocus on purpose to prevent our JS from focusing the time picker #}
         <input type="hidden" name="{{ field.name }}" id="{{ field.id }}-timestorage" autofocus
-               {% if field.data is not none %}value="{{ field.data.strftime('%H:%M') }}"{% endif %}
+               {% if field.data is not none %}value="{{ field.data | format_time }}"{% endif %}
```
ThiefMaster, line 7: Since this is a value that's consumed programmatically I think we should be explicit about the format, so I'd stick with strftime or use the `'code'` format of `format_time`)

tomasr8: Yep, make sense :+1:

indico/web/flask/app.py
```diff
@@ -219,14 +219,15 @@ def setup_jinja(app):
...
     # Useful constants
     app.add_template_global('^([0-9]|0[0-9]|1[0-9]|2[0-3]):[0-5][0-9]$', name='time_regex_hhmm')  # for input[type=time]
-    # Filters (indico functions returning UTF8)
+    # Filters
     app.add_template_filter(date_time_util.format_date)
     app.add_template_filter(date_time_util.format_time)
     app.add_template_filter(date_time_util.format_datetime)
     app.add_template_filter(date_time_util.format_human_date)
     app.add_template_filter(date_time_util.format_timedelta)
+    app.add_template_filter(date_time_util.format_interval)
```
ThiefMaster, line 228: I just realized that this takes two values... so probably it shouldn't be a filter since it's really weird to pass the start time and end date in completely different ways...

ThiefMaster (approved):

PR #5075, opened 2021-09-02
Increase the usage of program codes

plourenco:
The current use of program codes is very limited, these changes increase its usage up to:
- The timetable blocks
- The timetable tips
Screenshot from 2021-09-02 16-01-13
- The individual entry pages (session and contribution) (may be in the title instead?)
Screenshot from 2021-09-02 16-00-40

### DIFF
--- a/indico/modules/events/sessions/templates/display/session_display.html
+++ b/indico/modules/events/sessions/templates/display/session_display.html
@@ -16,6 +16,12 @@
 {% endblock %}
 
 {% block info %}
+    {% if sess.code %}
+        <div class="generic-data">
+            <i class="icon-tag header-data"></i>
+            {{ sess.code }}
+        </div>
+    {% endif %}
     <div class="generic-data">
         <i class="icon-time header-data"></i>
         <span class="time-info">
--- a/indico/modules/events/sessions/templates/display/session_display.html
+++ b/indico/modules/events/sessions/templates/display/session_display.html
@@ -16,7 +16,7 @@
 {% endblock %}
 
 {% block info %}
-    <div class="time-data">
+    <div class="generic-data">
         <i class="icon-time header-data"></i>
         <span class="time-info">
             {% if sess.start_dt -%}
@@ -27,7 +27,7 @@
         </span>
     </div>
     {% if sess.has_location_info %}
-        <div class="location-data">
+        <div class="generic-data">
             {{ render_location(sess, class='session-location') }}
         </div>
     {% endif %}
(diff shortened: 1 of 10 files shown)


indico/web/client/js/legacy/libs/timetable/Draw.js
```diff
@@ -147,6 +147,9 @@ type(
       if (this.eventData.slotTitle && this.eventData.slotTitle !== '') {
         title += ': ' + this.eventData.slotTitle;
       }
+      if (this.eventData.code != null) {
+        title += ` (${this.eventData.code})`
```
ThiefMaster, line 151:
```suggestion
        title += ` (${this.eventData.code})`;
```

indico/modules/events/sessions/templates/display/session_display.html
```diff
@@ -16,7 +16,13 @@
 {% endblock %}
 
 {% block info %}
-    <div class="time-data">
```
ThiefMaster, line 19:
```
>>> Event.query.filter(db.or_(Event.stylesheet.ilike(f'%{x}%') for x in ('time-data', 'location-data', 'board-data')), ~Event.is_deleted).count()
7
```

Check the custom CSS of these events in our production instance to see whether this change will break things. If yes we may want to consider leaving the old classes in addition to the new ones (or just keeping the old ones instead of replacing them with a generic one)...

plourenco: Checked those 7, they seem fine to me. Since those entries are from 2018/19, I don't think it's worth keeping.

ThiefMaster: I guess it also depends on whether we want to give people a way to target those elements in custom CSS or not..

ThiefMaster: But TBH I don't think it's needed. These classes aren't used on the main conference page, are they? Because *there* I think that'd be more important.. I'm not aware of people styling contribution/session pages...

plourenco: No, they aren't, only the individual pages.

indico/web/client/js/legacy/libs/timetable/Draw.js
```diff
@@ -147,6 +147,9 @@ type(
       if (this.eventData.slotTitle && this.eventData.slotTitle !== '') {
         title += ': ' + this.eventData.slotTitle;
       }
+      if (this.eventData.code != null) {
```
ThiefMaster, line 150:
```suggestion
      if (this.eventData.code) {
```

It's an empty string, not null

ThiefMaster (approved):

PR #5076, opened 2021-09-07
Add translations for countries

### LINKED ISSUE #5070: Localization: Country names
**Describe the bug**
When registration includes list of contry names, it uses English versions of their names.
For an event held in a different language this is unpleasant.

**To Reproduce**
1. Create a conference with a participant registration
2. In participant registration form, see the values in the Country drop down menu

**Expected behavior**
Country names should respect the event language.

tomasr8:
Closes #5070 

Extends `get_country/ies` to accept a `locale` parameter which defaults to the current locale.


ThiefMaster (approved):
thanks!

PR #5078, opened 2021-09-07
Prevent constraint violation when importing duplicate registrations

tomasr8:
This error can occur when the target event already has registrations with the same `friendly_id` and
a deleted registration form. When importing new registrations, this fix first marks all orphaned registrations as deleted.

### DIFF
--- a/indico/modules/events/registration/clone.py
+++ b/indico/modules/events/registration/clone.py
@@ -107,6 +107,8 @@ class RegistrationCloner(EventCloner):
             return [_('The target event already has registrations')]
 
     def run(self, new_event, cloners, shared_data, event_exists=False):
+        if event_exists:
+            self._delete_orphaned_registrations(new_event)
         form_map = shared_data['registration_forms']['form_map']
         field_data_map = shared_data['registration_forms']['field_data_map']
         for old_form, new_form in form_map.items():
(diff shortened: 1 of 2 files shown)


indico/modules/events/registration/clone.py
```diff
@@ -120,6 +120,8 @@ def _has_content(self, event):
                 .has_rows())
 
     def _clone_registrations(self, old_form, new_form, field_data_map):
+        self._delete_orphaned_registrations(new_form.event_id)
```
ThiefMaster, line 123: We don't need to call this when doing a normal clone, just when doing an import (`event_exists` being `True` in `run`).

indico/modules/events/registration/clone.py
```diff
@@ -150,3 +152,11 @@ def _synchronize_registration_friendly_id(self, new_event):
             .filter(Registration.event_id == new_event.id)
             .scalar() or 0
         )
+
+    def _delete_orphaned_registrations(self, event_id):
+        """Delete registrations whose registration form is deleted."""
+        registrations = db.session.query(Registration).filter(Registration.registration_form.has(is_deleted=True,
```
ThiefMaster, line 158:
```python
Registration.query.filter_by(event_id=event_id, Registration.registration_form.has(is_deleted=True)).update({Registration.is_deleted: True}, synchronize_session='fetch')
```

indico/modules/events/registration/clone.py
```diff
@@ -150,3 +152,11 @@ def _synchronize_registration_friendly_id(self, new_event):
             .filter(Registration.event_id == new_event.id)
             .scalar() or 0
         )
+
+    def _delete_orphaned_registrations(self, event_id):
+        """Delete registrations whose registration form is deleted."""
```
ThiefMaster, line 157: This comment is just explaining what the code already explains, but not the reason for it. I'd rather add something like

```python
# Avoid friendly ID conflicts with registrations in deleted regforms
```

(I'd use a comment instead of a docstring since an internal function doesn't really need *documentation*)

tomasr8:
Sentry issue: [INDICO-170](https://sentry.getindico.io/indico-team/indico/issues/2294/)

PR #5079, opened 2021-09-07
Fix reminder email ical on protected events

javfg:
Reminders in protected events are not generating the iCalendar file properly, as the access check is not passing without a user in the serializer function.

This PR adds a flag to skip the access check to the function. When creating a reminder email, the serializer is called with the flag set to `True`.

### DIFF
--- a/indico/modules/events/ical.py
+++ b/indico/modules/events/ical.py
@@ -105,22 +105,24 @@ def generate_event_component(event, user=None):
     return component
 
 
-def event_to_ical(event, user=None, scope=None):
+def event_to_ical(event, user=None, scope=None, *, skip_access_check=False):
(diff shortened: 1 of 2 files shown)


indico/modules/events/ical.py
```diff
@@ -130,7 +132,7 @@ def events_to_ical(events, user=None, scope=None):
     calendar.add('prodid', '-//CERN//INDICO//EN')
 
     for event in events:
-        if not event.can_access(user):
+        if not event.can_access(user) and not skip_access_check:
```
ThiefMaster, line 135:
```suggestion
        if not skip_access_check and not event.can_access(user):
```

no need to check if if the result is ignored

PR #5085, opened 2021-09-13
Replace translated date formats with 'format_skeleton'

tomasr8:
A few examples using `format_skeleton` for different locales:

| Locale | MMMMyyyy  | ddMMM |
| ------------- | ------------- | ---- |
| en_GB | September 2021 | 01 Sep |
| en_US | September 2021 | Sep 01 |
| fr_FR | septembre 2021 | 01 sept. |
| cs_CZ | září 2021 | 01. zář. |
| pl_PL | wrzesień 2021 | 01 wrz |
| mn_MN | 2021 оны есдүгээр сар | 9-р сарын 01 |
| ru_RU | сентябрь 2021 г. | 01 сент. |

### DIFF
--- a/indico/modules/categories/controllers/util.py
+++ b/indico/modules/categories/controllers/util.py
@@ -16,18 +16,16 @@ from indico.core import signals
 from indico.core.db import db
 from indico.modules.events.models.events import Event
 from indico.modules.events.util import serialize_event_for_json_ld
-from indico.util.date_time import format_date
-from indico.util.i18n import _, pgettext
+from indico.util.date_time import format_date, format_skeleton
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Could you add the current `MMMM yyyy` and `dd MMM` formats in the table as well to see if they differ from the ones the skeletons resolve to?

tomasr8:
Updated. The output for Polish from `format_date(..., 'MMMM yyyy')` looks incorrect to me.
I think `format_skeleton` has it right in this case. Not sure about Mongolian..

batpurev:
image

PR #5091, opened 2021-09-16
Add support for custom statuses/registration tags

### LINKED ISSUE #4948: Generic way to mark/filter registrations
Sometimes event organizers want to filter registrations based on some custom status and keep this persistent (e.g. to indicate who they already contacted about something etc)

Indico v1 had "custom statuses" which could do that. We replaced them with "management-only sections", but those are somewhat cumbersome to use: You need to edit the registration which is not only a heavy process and not something that can be done in bulk (and it also implies filling any mandatory fields added since the user registered).

Some possible solutions for this:

- bring back custom statuses; basically tags with a name and a color
- just add predefined flags with a color (like the color labels in lightroom). they wouldn't have a name/title, the interpretation would be completely up to what the event organizers decide outside indico

Regardless of the solution, assignment would be done either on the management display page of a registration or in bulk in the registration list. Of course it should also be possible to filter the registration list based on the flags

tomasr8:
Closes #4948

This is a work-in-progress PR which adds the ability to create registration tags which
can be used to mark and filter registrations. For a full description, see the linked issue.

- [x] New management page to define registration tags (view,add,edit,delete)
- [x] Assign and remove tags individually or in bulk
- [x] Add ability to filter registrations based on these tags
- [x] A better looking multiple select when assigning tags? Like the one from SUI
- [x] Find a suitable icon (swapped with badges, badges get `id-badge`)

### The management area for defining these tags:

image
image
image

### The  registration list area where one can assign and filter by these tags:

Screenshot from 2021-09-20 17-07-29
image
image
image

### The registration detail page
(don't mind the wrong icon)

tags


indico/migrations/versions/20210916_1127_b36825c7869e_add_registration_tags.py
```diff
@@ -0,0 +1,44 @@
...
+down_revision = '9d00917b2fa8'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'registration_tags',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('event_id', sa.Integer(), nullable=False, index=True),
```
ThiefMaster, line 23: Should they be linked to a particular registration form or the whole event? Not sure tbh.. maybe something to discuss on the channel. I'm fine either way; I guess event-wide has no disadvantages since it's up to the managers when to assign them anyway.

indico/modules/events/registration/controllers/management/regforms.py
```diff
@@ -46,6 +47,15 @@ def _process(self):
         return WPManageRegistration.render_template('management/regform_list.html', self.event, regforms=regforms)
 
 
+class RHManageRegistrationTags(RHManageRegFormsBase):
+    """List all registration tags for an event."""
+
+    def _process(self):
+        event_id = self.event.id
+        tags = RegistrationTag.query.filter(RegistrationTag.event_id == event_id).all()
```
ThiefMaster, line 55: `.with_parent(self.event)` - you almost never need to use ids in queries when it comes to FKs/relationships ;)

Or even better: `tags = self.event.registration_tags.all()` (or without `.all()` if you don't make the relationship dynamic; if you don't need extra filters etc you can just use a normal lazy one)

indico/modules/events/registration/controllers/management/regforms.py
```diff
@@ -292,3 +302,48 @@ def _process(self):
...
         return jsonify_form(form)
+
+
+class RHAddRegistrationTag(RHManageRegFormsBase):
+    """Add a new registration tag."""
+
+    def _process(self):
+        form = RegistrationTagForm()
+        if form.validate_on_submit():
+            tag = RegistrationTag(event_id=self.event.id)
```
ThiefMaster, line 313: similar to my comment above; don't use IDs. you can just do `event=self.event` or `self.event.registration_tags.append(tag)`. in either case, `db.session.add()` won't be needed

indico/modules/events/registration/forms.py
```diff
@@ -423,3 +423,9 @@ class RejectRegistrantsForm(IndicoForm):
 
     def is_submitted(self):
         return super().is_submitted() and 'submitted' in request.form
+
+
+class RegistrationTagForm(IndicoForm):
+    name = StringField(_('Name'), [DataRequired()])
+    color = StringField(_('Color'), [DataRequired()], widget=ColorInput())
+    fields = ['name', 'color']
```
ThiefMaster, line 431: not needed; `form_fields` works fine without specifying a list of fields and in almost all cases that's enough (no custom order, omitting some fields, etc)

indico/modules/events/registration/models/registrations.py
```diff
@@ -226,6 +226,11 @@ class Registration(db.Model):
             lazy=True
         )
     )
+    #: The registration tags assigned to this registration
+    registration_tags = db.relationship(
+        'RegistrationRegistrationTag',
+        lazy='dynamic'
```
ThiefMaster, line 232: I don't think this is a good option here

ThiefMaster: you may want to make this a proper many2many relationship (`secondary` attribute pointing to the mapping table, which would not be a full-blown `db.Model` subclass)

indico/modules/events/registration/models/registrations.py
```diff
@@ -697,3 +702,55 @@ def _set_checked_in_dt(target, value, *unused):
...
+            lazy='dynamic'
+        )
+    )
+
+    @property
+    def locator(self):
+        return dict(self.event.locator, tag_id=self.id)
+
+
+class RegistrationRegistrationTag(db.Model):
```
ThiefMaster, line 749: Search for `db.Table` to find some examples on how to define many to many mapping tables

indico/modules/events/registration/templates/management/_registration_details.html
```diff
@@ -194,6 +196,16 @@
 {% endmacro %}
 
 
+{% macro _render_registration_tags(registration, assigned_tags, all_tags) %}
+    <div class="section" id="registration-detail-registration-tags-assign"
+        data-event-id="{{registration.registration_form.event.id}}"
+        data-regform-id="{{registration.registration_form.id}}"
+        data-registration-id="{{registration.id}}"
+        data-assigned-tags='{{assigned_tags|tojson}}' data-all-tags='{{all_tags|tojson}}'>
```
ThiefMaster, line 204: indentation, and usually we put spaces after `{{` and before `}}`

indico/modules/events/registration/templates/management/_reglist.html
```diff
@@ -80,6 +80,12 @@
                                                 -
                                             {% endif %}
                                         </td>
+                                    {% elif item.id == 'registration_tags' %}
+                                        <td class="i-table">
+                                            {% for tag in registration.registration_tags|sort(attribute='name', case_sensitive=False) %}
```
ThiefMaster, line 85: lowercase `false` (same for true and none) is preferred in jinja templates

indico/modules/events/registration/blueprint.py
```diff
@@ -123,6 +123,18 @@
...
+_bp.add_url_rule('/manage/registration/tags', 'manage_registration_tags', registration_tags.RHManageRegistrationTags,
+                 methods=('GET',))
+_bp.add_url_rule('/manage/registration/tags/add', 'manage_registration_tags_add',
+                 registration_tags.RHRegistrationTagAdd, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/registration/tags/<int:tag_id>/edit', 'manage_registration_tags_edit',
+                 registration_tags.RHRegistrationTagEdit, methods=('GET', 'POST'))
+_bp.add_url_rule('/manage/registration/tags/<int:tag_id>/delete', 'manage_registration_tags_delete',
+                 registration_tags.RHRegistrationTagDelete, methods=('POST',))
+_bp.add_url_rule('/manage/registration/<int:reg_form_id>/tags/assign', 'manage_registration_tags_assign',
+                 registration_tags.RHRegistrationTagsAssign, methods=('GET', 'POST'))
```
ThiefMaster, line 136: I think we always use POST here

indico/modules/events/registration/templates/management/registration_tag_edit.html
```diff
@@ -0,0 +1,19 @@
+{% from 'forms/_form.html' import form_header, form_footer, form_rows %}
+{% from 'message_box.html' import message_box %}
+
+<div class="registration-tags">
+    {{ form_header(form, id='registration-tag-form') }}
+    {{ form_rows(form) }}
+    {% call form_footer(form) %}
+        {% if tag %}
+            <input id="save-button" class="i-button big highlight" type="submit"
+                value="{% trans %}Save{% endtrans %}" data-disabled-until-change>
```
ThiefMaster, line 10: indentation

indico/modules/events/registration/templates/management/registration_tags.html
```diff
@@ -0,0 +1,59 @@
...
+                    <a class="i-button icon-plus" data-href="{{ url_for('.manage_registration_tags_add', event) }}"
+                        data-title="{%- trans %}Add a tag{% endtrans -%}"
+                        data-ajax-dialog
+                        data-reload-after>
+                        {%- trans %}Add a tag{% endtrans -%}
+                    </a>
+                </div>
+            </div>
+            <div class="i-box-content">
+                {% if tags|length > 0 %}
```
ThiefMaster, line 51: `{% if tags %}`

indico/modules/events/registration/models/registrations.py
```diff
@@ -226,6 +249,12 @@ class Registration(db.Model):
             lazy=True
         )
     )
+    #: The registration tags assigned to this registration
+    registration_tags = db.relationship(
```
ThiefMaster, line 253: It's useful to have backrefs on all relationships, and be explicit about the laziness. In this case, you can either make it `lazy=False` or alternative make it lazy but then make sure to specify the load option e.g. in the registration list to make sure we do not send separate queries for each registration to get its tags.

Also, just `tags` would be enough for the name - `registration.tags` is more concise than `registration.registration_tags`...

indico/modules/events/registration/controllers/management/registration_tags.py
```diff
@@ -0,0 +1,124 @@
...
+class RHAPIRegistrationTagsAssign(RHManageRegFormBase):
+    """Internal API to assign and remove registration tags."""
+
+    @use_kwargs({
+        'registration_ids': fields.List(fields.Number(), required=True),
+        'add': fields.List(fields.Number(), missing=[]),
+        'remove': fields.List(fields.Number(), missing=[])
+    })
+    def _process_POST(self, registration_ids, add, remove):
+        registrations = [Registration.get(id) for id in registration_ids]
```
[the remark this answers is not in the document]
ThiefMaster: i'd just change it to use webargs in the base class - by default webargs accepts data from both json and formdata so it should cover all the cases. See my change to `RHManageCategorySelectedEventsBase` in 2984d27508 for an example.

indico/modules/events/registration/models/registrations.py
```diff
@@ -697,3 +731,48 @@ def _set_checked_in_dt(target, value, *unused):
     @listens_for(Registration.transaction, 'set')
     def _set_transaction_id(target, value, *unused):
         value.registration = target
+
+
+class RegistrationTag(db.Model):
```
ThiefMaster, line 736: Maybe this could go in a `tags.py` file?

indico/modules/events/registration/models/registrations.py
```diff
@@ -697,3 +731,48 @@ def _set_checked_in_dt(target, value, *unused):
...
+    """Registration tag used to mark/filter registrations."""
+    __tablename__ = 'registration_tags'
+    __table_args__ = {'schema': 'event_registration'}
```
ThiefMaster, line 739: I'd add a check constraint that ensures tag names are unique. See `ix_uq_tags_event_id_code_lower` for an example in Editing tags

indico/modules/events/registration/models/registrations.py
```diff
@@ -697,3 +731,48 @@ def _set_checked_in_dt(target, value, *unused):
...
+    )
+    #: The ID of the event where this tag was created
+    event_id = db.Column(
+        db.Integer,
+        db.ForeignKey('events.events.id'),
+        index=False,
+        nullable=False
+    )
+    #: Tag name
+    name = db.Column(
```
ThiefMaster, line 754: I'd use `title` to be consistent with the naming in EditingTag

indico/modules/events/registration/models/registrations.py
```diff
@@ -697,3 +731,48 @@ def _set_checked_in_dt(target, value, *unused):
...
+        db.String,
+        nullable=False
+    )
+    #: The Event where this tag was created
+    event = db.relationship(
+        'Event',
+        lazy=True,
+        backref=db.backref(
+            'registration_tags',
+            lazy='dynamic'
```
ThiefMaster, line 769: `lazy=True` should be fine here

indico/modules/events/registration/controllers/management/registration_tags.py
```diff
@@ -0,0 +1,117 @@
...
+from indico.web.flask.util import url_for
+from indico.web.util import jsonify_data, jsonify_template
+
+
+class RHManageRegistrationTags(RHManageRegFormsBase):
+    """List all registration tags for an event."""
+
+    def _process(self):
+        tags = self.event.registration_tags.all()
+        tags.sort(key=lambda tag: str.casefold(tag.name))
```
ThiefMaster, line 29: If you change the lazy=dynamic from my other comment, replace this with `tags = sorted(self.event.registration_tags, ...)` or maybe even move the sorting into the jinja template (`|sort(attribute='name')`) - then you can even use `event.registration_tags` directly in the template

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -63,7 +66,15 @@
 def _render_registration_details(registration):
     event = registration.registration_form.event
     tpl = get_template_module('events/registration/management/_registration_details.html')
-    return tpl.render_registration_details(registration=registration, payment_enabled=event.has_feature('payment'))
+
+    assigned_tags = RegistrationTagSchema(many=True).dump(registration.tags)
+    assigned_tags.sort(key=lambda tag: str.casefold(tag['name']))
```
ThiefMaster, line 71: I'd do this in a `post_dump` function of the schema

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -63,7 +66,15 @@
...
+
+    all_tags = RegistrationTagSchema(many=True).dump(RegistrationTag.query.with_parent(event).all())
```
ThiefMaster, line 73: Since you use it twice it'd be better to instantiate it just once: `schema = RegistrationTagSchema(many=True)`

ThiefMaster, line 73: You can use the relationship here instead of querying using `with_parent`

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -164,9 +175,12 @@ class RHRegistrationsActionBase(RHManageRegFormBase):
 
     registration_query_options = ()
 
-    def _process_args(self):
+    @use_kwargs({
+        'registration_id': fields.List(fields.Integer(), missing=[]),
```
ThiefMaster, line 179: I'd use the name mapping (like in the example I showed you) in order to call the attribute `registration_ids`:

    'registration_ids': fields.List(fields.Int(), data_key='registration_id', missing=[]),

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -164,9 +175,12 @@ class RHRegistrationsActionBase(RHManageRegFormBase):
...
+    })
+    def _process_args(self, registration_id):
         RHManageRegFormBase._process_args(self)
-        ids = set(request.form.getlist('registration_id'))
+        ids = set(registration_id)
```
ThiefMaster, line 183: I think we can safely skip this and just use it directly below. Duplicates in WHERE IN aren't a problem.

indico/modules/events/registration/forms.py
```diff
@@ -423,3 +423,24 @@ class RejectRegistrantsForm(IndicoForm):
...
+    name = StringField(_('Name'), [DataRequired()])
+    color = StringField(_('Color'), [DataRequired()], widget=ColorInput())
+
+
+class RegistrationTagsAssignForm(IndicoForm):
+    """
+    Form to assign registration tags to registrations.
+    """
+    add = IndicoMultipleTagSelectField(_('Add'), description=_('Selected tags to assign'))
+    remove = IndicoMultipleTagSelectField(_('Remove'), description=_('Selected tags to remove'))
```
ThiefMaster, line 441: "Select tags to ..." ("Selected" sounds weird there)

indico/modules/events/registration/forms.py
```diff
@@ -423,3 +423,24 @@ class RejectRegistrantsForm(IndicoForm):
 
     def is_submitted(self):
         return super().is_submitted() and 'submitted' in request.form
+
+
+class RegistrationTagForm(IndicoForm):
+    """
+    Form to create a new registration tag.
+    """
```
ThiefMaster, line 431:
```suggestion
    """Form to create/update a registration tag."""
```

same in the next class; these docstrings fit nicely in one line

indico/modules/events/registration/lists.py
```diff
@@ -54,11 +56,25 @@ def __init__(self, regform):
             'payment_date': {
                 'title': _('Payment date'),
             },
+            'registration_tags_present': {
```
ThiefMaster, line 59: Just `tags_present` and `tags_absent` or even `tags` and `tags_absent`?

indico/modules/events/registration/lists.py
```diff
@@ -30,6 +31,7 @@ def __init__(self, regform):
             'items': ('title', 'email', 'affiliation', 'reg_date', 'state'),
```
ThiefMaster, line 31: I'd show tags by default

indico/modules/events/registration/lists.py
```diff
@@ -54,11 +56,25 @@ def __init__(self, regform):
...
+                'filter_only': True
+            },
         }
         self.personal_items = ('title', 'first_name', 'last_name', 'email', 'position', 'affiliation', 'address',
                                'phone', 'country')
         self.list_config = self._get_config()
 
+    def _get_registration_tag_choices(self):
+        tags = RegistrationTag.query.with_parent(self.event).all()
+        tags.sort(key=lambda tag: str.casefold(tag.name))
```
ThiefMaster, line 75: In all the places where you use casefold, I'd rather use our `natural_sort_key` util. It also does a case-insensitive sort but it will work nicely when people end up having a digit at the beginning of the tag.

indico/modules/events/registration/models/registrations.py
```diff
@@ -53,6 +53,29 @@ def _get_next_friendly_id(context):
     return increment_and_get(Event._last_friendly_registration_id, Event.id == event_id)
 
 
+registrations_tags_table = db.Table(
+    'registrations_registration_tags',
+    db.metadata,
+    db.Column(
+        'registration_id',
+        db.Integer,
+        db.ForeignKey('event_registration.registrations.id', ondelete='CASCADE'),
```
ThiefMaster, line 62: When using `ondelete='CASCADE'` you need `passive_deletes=True` on the corresponding relationships (or backrefs)

indico/modules/events/registration/models/registrations.py
```diff
@@ -226,6 +249,17 @@ class Registration(db.Model):
             lazy=True
         )
     )
+    #: The registration tags assigned to this registration
+    tags = db.relationship(
+        'RegistrationTag',
+        secondary=registrations_tags_table,
+        cascade='all, delete',
```
ThiefMaster, line 256: I think this may delete a tag when hard-deleting (which we usually don't do) the corresponding registration!
https://docs.sqlalchemy.org/en/14/orm/cascades.html#delete

I don't think we need a `cascade` setting here at all

indico/modules/events/registration/templates/management/_reglist.html
```diff
@@ -24,7 +24,14 @@
                             <th class="i-table">{% trans %}Full name{% endtrans %}</th>
 
                             {% for item in static_columns %}
-                                <th class="i-table" data-sorter="text">{{ item.caption }}</th>
+                                {% set filter_only = item.get('filter_only', false) %}
+                                {% if not filter_only %}
```
ThiefMaster, line 28:
```suggestion
                                {% if not item.get('filter_only') %}
```

indico/modules/events/registration/templates/management/_reglist.html
```diff
@@ -24,7 +24,14 @@
...
+                                    {%if item.id == 'registration_tags_present' %}
```
ThiefMaster, line 29:
```suggestion
                                    {% if item.id == 'registration_tags_present' %}
```

indico/web/templates/forms/multiple_tag_select_widget.html
```diff
@@ -0,0 +1,15 @@
...
+
+{% block html %}
+    <div id="{{ field.id }}-wrapper"></div>
+{% endblock %}
+
+{% block javascript %}
+    <script>
+        setupMultipleTagSelectWidget({
+            fieldId: {{ field.id | tojson }},
+            wrapperId: "{{ field.id }}-wrapper",
```
ThiefMaster, line 11:
```suggestion
            wrapperId: '{{ field.id }}-wrapper',
```

indico/web/templates/forms/multiple_tag_select_widget.html
```diff
@@ -0,0 +1,15 @@
...
+            choices: {{ field.choices | tojson }},
+        });
+    </script>
+{% endblock %}
```
ThiefMaster, line 15: missing trailing line break (make sure to configure your editor to add those automatically)

indico/web/templates/forms/multiple_tag_select_widget.html
ThiefMaster: But I'd just derive this from `fieldId` in the code and not pass it separately...

indico/modules/events/client/js/util/list_generator.js
```diff
@@ -164,7 +164,7 @@
       applySearchFilters = setupSearchBox(filterConfig);
     }
     setupStaticURLGeneration();
-    handleRowSelection(false);
+    handleRowSelection(true);
```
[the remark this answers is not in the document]
tomasr8: When you reload the registration list view (as I do when you assign new tags with `data-reload-after`)
the action buttons ,e.g., `Remove`, 'Actions', are disabled despite the registrations staying selected. Thought it wasn't
very user friendly..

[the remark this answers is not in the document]
ThiefMaster: Oh yes, I always found this rather annoying!

indico/modules/events/registration/controllers/management/registration_tags.py
```diff
@@ -0,0 +1,115 @@
...
+
+def _assign_registration_tags(event, registrations, add, remove):
+    add_ids = [int(id) for id in add]
+    remove_ids = [int(id) for id in remove]
+    add_tags = RegistrationTag.query.with_parent(event).filter(RegistrationTag.id.in_(add_ids)).all()
+    remove_tags = RegistrationTag.query.with_parent(event).filter(RegistrationTag.id.in_(remove_ids)).all()
+
+    for reg in registrations:
+        reg.tags.update(add_tags)
+        reg.tags.difference_update(remove_tags)
```
ThiefMaster, line 84: If you put `set()` around the two queries above, you can use the normal set operators here:

```
reg.tags |= add_tags
reg.tags -= remove_tags
```

indico/modules/events/registration/forms.py
```diff
@@ -423,3 +424,33 @@ class RejectRegistrantsForm(IndicoForm):
...
+    """Form to create a new registration tag."""
+
+    title = StringField(_('Title'), [DataRequired()])
+    color = StringField(_('Color'), [DataRequired()], widget=ColorInput())
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event', None)
+        super().__init__(*args, **kwargs)
+
+    def validate_title(self, field):
```
ThiefMaster, line 439: I think you're missing the logic for editing here - changing a tag's color will most likely fail the validation in here because the current check will think the title conflicts with itself.

indico/modules/events/registration/models/tags.py
```diff
@@ -0,0 +1,60 @@
...
+    #: The ID of the object
+    id = db.Column(
+        db.Integer,
+        primary_key=True
+    )
+    #: The ID of the event where this tag was created
+    event_id = db.Column(
+        db.Integer,
+        db.ForeignKey('events.events.id'),
+        index=False,
```
ThiefMaster, line 32:
```suggestion
        index=True,
```

indico/migrations/versions/20210927_1027_b36825c7869e_add_registration_tags.py
```diff
@@ -0,0 +1,47 @@
...
+        schema='event_registration'
+    )
+
+    op.create_index('ix_uq_registration_tags_title_lower', 'registration_tags', [sa.text('lower(title)')],
+                    unique=True, schema='event_registration')
+
+    op.create_table(
+        'registrations_registration_tags',
+        sa.Column('registration_id', sa.Integer(), nullable=False, index=False),
+        sa.Column('registration_tag_id', sa.Integer(), nullable=False, index=False),
```
ThiefMaster, line 37:
```suggestion
        sa.Column('registration_id', sa.Integer(), nullable=False, index=True),
        sa.Column('registration_tag_id', sa.Integer(), nullable=False, index=True),
```

ThiefMaster (commented):
Alembic/model inconsistencies

indico/modules/events/registration/forms.py
```diff
@@ -423,3 +424,33 @@ class RejectRegistrantsForm(IndicoForm):
 
     def is_submitted(self):
         return super().is_submitted() and 'submitted' in request.form
+
+
+class RegistrationTagForm(IndicoForm):
+    """Form to create a new registration tag."""
+
+    title = StringField(_('Title'), [DataRequired()])
+    color = StringField(_('Color'), [DataRequired()], widget=ColorInput())
```
ThiefMaster, line 433: Could we use the existing color picker we have? Either `ColorPickerWidget` for free-form color choice (but that may be a bit problematic due to the text color being always white) or `IndicoSinglePalettePickerField` to select colors from a palette. We could use `ColorMixin` in the model and then `IndicoPalettePickerField` as well; that's the color palette which we also use e.g. for sessions (those entries all have fg/bg color pairs that work well together).

indico/modules/events/registration/forms.py
```diff
@@ -423,3 +424,33 @@ class RejectRegistrantsForm(IndicoForm):
...
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event', None)
```
ThiefMaster, line 436: This isn't optional (and in fact it's currently breaking edits since it's missing when instantiating the form for editing)

```suggestion
        self.event = kwargs.pop('event')
```

indico/modules/events/registration/client/js/components/RegistrationTagsEditableList.jsx
```diff
@@ -0,0 +1,151 @@
...
+                </div>
+              ))}
+            </div>
+            <Button
+              size="mini"
+              fluid
+              styleName="close-button"
+              content={Translate.string('Close')}
+              onClick={() => setOpen(false)}
+            />
```
ThiefMaster, line 127: Do we need that close button? Clicking outside the popover does the job, and I think it's not particularly common for this kind of UI to have a dedicated close button (similar dropdowns in other places in Indico don't have one either).

indico/modules/events/registration/lists.py
```diff
@@ -54,11 +57,24 @@ def __init__(self, regform):
             'payment_date': {
                 'title': _('Payment date'),
             },
+            'tags_present': {
+                'title': _('Tags present'),
```
ThiefMaster, line 61:
```suggestion
                'title': _('Tags'),
```

indico/modules/events/registration/forms.py
```diff
@@ -423,3 +424,33 @@ class RejectRegistrantsForm(IndicoForm):
...
+
+
+class RegistrationTagsAssignForm(IndicoForm):
+    """Form to assign registration tags to registrations."""
+
+    add = IndicoMultipleTagSelectField(_('Add'), description=_('Select tags to assign'))
+    remove = IndicoMultipleTagSelectField(_('Remove'), description=_('Select tags to remove'))
+    registration_id = HiddenFieldList()
+    submitted = HiddenField()
+
```
ThiefMaster, line 454: Might be nice to fail validation when selecting a tag both for adding anad removal:

```python
    def validate_remove(self, field):
        if set(self.remove.data) & set(self.add.data):
            raise ValidationError(_('You cannot add and remove the same tag'))

    validate_add = validate_remove
```

indico/modules/events/registration/client/js/components/RegistrationTagsEditableList.jsx
tomasr8: I suppose not, I mostly reused `PopoverDropdownMenu` from room booking which has a very similar behavior:
image

ThiefMaster: ah, I wasn't aware this one also had a cancel button. hm... @pferreir - any suggestions since you added the one in RB?

indico/modules/events/registration/models/tags.py
```diff
@@ -0,0 +1,60 @@
...
+            lazy=True
+        )
+    )
+
+    # relationship backrefs:
+    # - registrations (Registration.tags)
+
+    @property
+    def locator(self):
+        return dict(self.event.locator, tag_id=self.id)
```
ThiefMaster, line 60: It would be nice to have a repr in this class:

```python
    def __repr__(self):
        return format_repr(self, 'id', 'event_id', _text=self.title)
```

The default repr is `<RegistrationTag 4>` and with the custom implementation you get the much nicer `<RegistrationTag(4, event_id=614422): "foo">`

indico/modules/events/registration/models/tags.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class RegistrationTag(db.Model):
+    """Registration tag used to mark/filter registrations."""
+
+    __tablename__ = 'registration_tags'
+
+    @declared_attr
+    def __table_args__(cls):
+        return (db.Index('ix_uq_registration_tags_title_lower', db.func.lower(cls.title), unique=True),
```
ThiefMaster, line 20: This index needs to include `cls.event_id` as well (have a look at the one in `EditingTag`). Otherwise tag names would need to be unique globally across all events...

indico/modules/events/registration/models/tags.py
```diff
@@ -0,0 +1,60 @@
...
+    color = db.Column(
+        db.String,
+        nullable=False
+    )
+    #: The Event where this tag was created
+    event = db.relationship(
+        'Event',
+        lazy=True,
+        backref=db.backref(
+            'registration_tags',
```
ThiefMaster, line 50:
```suggestion
            'registration_tags',
            cascade='all, delete-orphan',
```

While we do not hard-delete events, here we need the cascade setting since otherwise the `event_id` would get nulled (which isn't possible).

indico/modules/events/registration/client/js/components/RegistrationTagsEditableList.jsx
tomasr8: Should I remove it?

What it looks like with and without:

image
image

ThiefMaster: IMHO it looks better without

indico/modules/events/registration/forms.py
```diff
@@ -423,3 +425,41 @@ class RejectRegistrantsForm(IndicoForm):
...
+    title = StringField(_('Title'), [DataRequired()])
+    color = SelectField(_('Color'), [DataRequired()], choices=list(zip(get_sui_colors(), get_sui_colors())))
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event')
+        self.tag = kwargs.pop('tag', None)
+        super().__init__(*args, **kwargs)
+
+    def validate_title(self, field):
+        if not self.tag:
```
ThiefMaster, line 442: That's not correct :/ You now don't check for collisions at all when editing... Check the logic in `EventLabelForm` for an example that handles both adding and editing.

indico/modules/events/registration/controllers/management/registration_tags.py
```diff
@@ -0,0 +1,115 @@
...
+
+        return jsonify_template('events/registration/management/registration_tag_edit.html',
+                                event=self.event, form=form, tag=None)
+
+
+class RHRegistrationTagDelete(RHManageRegFormsBase):
+    """Delete a registration tag."""
+
+    def _process(self):
+        tag = RegistrationTag.get_or_404(request.view_args['tag_id'])
```
ThiefMaster, line 53: This allows people to delete arbitrary tags even from other events (same in the editing RH). Create an base class similar to `RHManageEventLabelBase` that queries the tag using `with_parent(self.event)` in its `_process_args` and use that as the base class for editing and deleting tags.

indico/modules/events/registration/templates/management/_registration_details.html
```diff
@@ -194,6 +196,16 @@
 {% endmacro %}
 
 
+{% macro _render_registration_tags(registration, assigned_tags, all_tags) %}
+    <div class="section" id="registration-detail-registration-tags-assign"
+         data-event-id="{{ registration.registration_form.event.id }}"
+         data-regform-id="{{ registration.registration_form.id }}"
+         data-registration-id="{{ registration.id }}"
+         data-assigned-tags='{{ assigned_tags|tojson }}' data-all-tags='{{ all_tags|tojson }}'>
```
ThiefMaster, line 204: I think this may be prone to XSS... Usually we use `="{{ XXX|tojson|forceescape }}"` to ensure things are correctly escaped (tojson by default is meant for script tags, so outside you need forceescape)

indico/modules/events/registration/templates/management/_reglist.html
```diff
@@ -80,7 +86,13 @@
                                                 -
                                             {% endif %}
                                         </td>
-                                    {% else %}
+                                    {% elif item.id == 'tags_present' %}
+                                        <td class="i-table">
+                                            {% for tag in registration.tags|sort(attribute='title', case_sensitive=false) %}
+                                                <span class="registration-tag" style="background-color: {{tag.color}};">{{ tag.title }}</span>
```
ThiefMaster, line 92:
```suggestion
                                                <span class="registration-tag" style="background-color: {{ tag.color }};">{{ tag.title }}</span>
```

indico/modules/events/registration/templates/management/registration_tag_assign.html
```diff
@@ -0,0 +1,14 @@
+{% from 'forms/_form.html' import form_header, form_footer, form_rows %}
+{% from 'message_box.html' import message_box %}
+
+<div class="registration-tags">
+    {{ form_header(form, id='registration-tag-assign-form') }}
```
ThiefMaster, line 5: This ID doesn't seem to be used anywhere

```suggestion
    {{ form_header(form) }}
```

indico/modules/events/registration/templates/management/registration_tag_edit.html
```diff
@@ -0,0 +1,19 @@
+{% from 'forms/_form.html' import form_header, form_footer, form_rows %}
+{% from 'message_box.html' import message_box %}
+
+<div class="registration-tags">
+    {{ form_header(form, id='registration-tag-form') }}
```
ThiefMaster, line 5:
```suggestion
    {{ form_header(form) }}
```

unused id as well

ThiefMaster: But it looks like we can get rid fo this whole template and replace it with `jsonify_form` in the controller - this is what we do for pretty much all other simple cases where we just render a wtform in a modal dialog...

indico/web/forms/fields/simple.py
```diff
@@ -157,3 +157,10 @@ def process_data(self, data):
 
 class IndicoTagListField(HiddenFieldList):
     widget = JinjaWidget('forms/tag_list_widget.html', single_kwargs=True)
+
+
+class IndicoMultipleTagSelectField(SelectMultipleField):
+    widget = JinjaWidget('forms/multiple_tag_select_widget.html', single_kwargs=True, single_line=True)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
ThiefMaster, line 166: Let's remove this ;)

indico/modules/events/registration/forms.py
tomasr8: yup brainfart.. somehow thought I didn't have to check for collisions when editing.. :sweat_smile:

indico/modules/events/registration/templates/management/registration_tag_assign.html
tomasr8: I'd delete this template as well but it's by default rendered too wide so I need to somehow keep the `.registration-tags` class to limit the width..

ThiefMaster: IF setting it on the form itself works, you could try passing it via `form_header_kwargs` to `jsonify_form`

indico/modules/events/registration/templates/management/_registration_details.html
```diff
@@ -25,7 +25,9 @@
         {% if registration.state.name not in ('rejected', 'withdrawn') %}
             {{ _render_withdraw_actions(registration) }}
         {% endif %}
-
+        <div class="action-box">
+            {{ _render_registration_tags(registration, assigned_tags, all_tags) }}
+        </div>
```
ThiefMaster, line 30:
```suggestion
        {% if all_tags %}
            <div class="action-box">
                {{ _render_registration_tags(registration, assigned_tags, all_tags) }}
            </div>
        {% endif %}
```

No need to show the tagging UI if no tags are defined

indico/modules/events/registration/templates/management/_reglist.html
```diff
@@ -80,7 +86,13 @@
                                                 -
                                             {% endif %}
                                         </td>
-                                    {% else %}
+                                    {% elif item.id == 'tags_present' %}
+                                        <td class="i-table">
+                                            {% for tag in registration.tags|sort(attribute='title', case_sensitive=false) %}
+                                                <span class="registration-tag" style="background-color: {{ tag.color }};">{{ tag.title }}</span>
```
ThiefMaster, line 92: Now that we use the SUI colors, should we also use the SUI labels?

    <span class="ui label tiny {{ tag.color }}">{{ tag.title }}</span>

(not sure about the size; you; have to check which one fits best)

indico/modules/events/registration/templates/management/registration_tag_assign.html
```diff
@@ -0,0 +1,14 @@
...
+<div class="registration-tags">
+    {{ form_header(form) }}
+    {{ form_rows(form) }}
+    {% call form_footer(form) %}
+        <input id="save-button" class="i-button big highlight" type="submit"
+               value="{% trans %}Save{% endtrans %}" data-disabled-until-change>
+        <a href="{{ url_for('.manage_reglist', regform) }}" class="i-button big" data-button-back>
+            {%- trans %}Back{% endtrans -%}
+        </a>
+    {% endcall %}
```
ThiefMaster, line 13: I think you can use the `simple_form()` macro for this...

indico/web/client/styles/modules/_registrationform.scss
```diff
@@ -1310,3 +1310,14 @@ textarea.regFormDescriptionInputMgmt {
     color: $dark-gray;
   }
 }
+
+.registration-tag {
+  color: #fff;
+  font-weight: bold;
+  border-radius: 6px;
+  padding: 2px 7px;
+}
```
ThiefMaster, line 1319: If you go for the SUI label this is most likely no longer needed

indico/modules/events/registration/templates/management/_reglist.html
```diff
@@ -24,7 +24,13 @@
                             <th class="i-table">{% trans %}Full name{% endtrans %}</th>
 
                             {% for item in static_columns %}
-                                <th class="i-table" data-sorter="text">{{ item.caption }}</th>
+                                {% if not item.get('filter_only') %}
+                                    {% if item.id == 'tags_present' %}
+                                        <th class="i-table" data-sorter="text">{% trans %}Tags{% endtrans %}</th>
+                                    {% else %}
```
ThiefMaster, line 30: Why is this needed btw? The `.title` should be correct

indico/modules/events/registration/templates/management/_reglist.html
```diff
@@ -24,7 +24,13 @@
                             <th class="i-table">{% trans %}Full name{% endtrans %}</th>
 
                             {% for item in static_columns %}
-                                <th class="i-table" data-sorter="text">{{ item.caption }}</th>
+                                {% if not item.get('filter_only') %}
```
ThiefMaster, line 27: this one can be inlined in the for loop (`{% for item in static_columns if not item.get('filter_only') %}`)

indico/modules/events/registration/templates/management/_reglist.html
tomasr8: Thanks for spotting that, it's not needed anymore. Before, the custom filter showed `Registration tags` and the table was just `Tags`. Now they're the same so it's redundant.

indico/modules/events/registration/lists.py
```diff
@@ -54,11 +57,24 @@ def __init__(self, regform):
             'payment_date': {
                 'title': _('Payment date'),
             },
+            'tags_present': {
+                'title': _('Tags'),
+                'filter_choices': registration_tag_choices
+            },
+            'tags_absent': {
+                'title': _('Tags absent'),
```
ThiefMaster, line 65:
```suggestion
                'title': _('Without tags'),
```

I wonder if that's clearer... not 100% sure though

indico/modules/events/registration/templates/management/_reglist.html
```diff
@@ -80,7 +79,13 @@
...
                                             {% endif %}
                                         </td>
-                                    {% else %}
+                                    {% elif item.id == 'tags_present' %}
+                                        <td class="i-table">
+                                            {% for tag in registration.tags|sort(attribute='title', case_sensitive=false) %}
+                                                <span class="ui label {{ tag.color }}">{{ tag.title }}</span>
+                                            {% endfor %}
+                                        </td>
+                                    {% elif not item.get('filter_only') %}
```
ThiefMaster, line 88: What about keeping this as `else` and adding the condition to the loop itself?

indico/modules/events/registration/lists.py
tomasr8: 'Absent tags', 'Missing tags', .. dunno :confused:

Issue #5055, opened 2021-08-25
Unlisted events: Global settings / ACL


ThiefMaster:
We need a setting to globally enable/disable unlisted events and an acl setting (both are supported by `SettingsProxy`) to specify who can create such events. Since there is no "Everyone" acl entry, I would propose two settings, one ACL to allow specific users/groups to crate unlisted events, and one toggle that allows everyone to do so (in that case the ACL would not be taken into account).

Not sure what's the best place to put it in the admin area; initially we could just add it to the "General Settings"...

PR #5095, opened 2021-09-17
Unlisted events: Global settings / ACL

javfg:
This PR closes https://github.com/indico/indico/issues/5055.

Adds a settings page in the customization tree (I thought the general settings are too global to include this in there) to add permissions to create unlisted events:

image

### DIFF
--- a/indico/modules/categories/util.py
+++ b/indico/modules/categories/util.py
@@ -233,5 +233,9 @@ def serialize_category_role(role, legacy=True):
 
 
 def can_create_unlisted_events(user):
-    return (user.is_admin or unlisted_events_settings.get('enabled') and (not unlisted_events_settings.get('restricted')
-            or unlisted_events_settings.acls.contains_user('authorized_creators', user)))
+    if not user or not unlisted_events_settings.get('enabled'):
+        return False
+    elif user.is_admin or not unlisted_events_settings.get('restricted'):
+        return True
+    else:
+        return unlisted_events_settings.acls.contains_user('authorized_creators', user)
--- a/indico/modules/categories/util.py
+++ b/indico/modules/categories/util.py
@@ -22,6 +22,7 @@ from indico.modules.events import Event
 from indico.modules.events.contributions import Contribution
 from indico.modules.events.contributions.models.subcontributions import SubContribution
 from indico.modules.events.sessions import Session
+from indico.modules.events.settings import unlisted_events_settings
 from indico.modules.events.timetable.models.entries import TimetableEntry, TimetableEntryType
 from indico.util.caching import memoize_redis
 from indico.util.date_time import now_utc
@@ -229,3 +230,8 @@ def serialize_category_role(role, legacy=True):
             'name': role.name,
             'identifier': f'CategoryRole:{role.id}',
         }
+
+
+def can_create_unlisted_events(user):
+    return (user.is_admin or unlisted_events_settings.get('enabled') and (not unlisted_events_settings.get('restricted')
+            or unlisted_events_settings.acls.contains_user('authorized_creators', user)))
--- a/indico/modules/categories/util_test.py
+++ b/indico/modules/categories/util_test.py
@@ -1,0 +1,28 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2021 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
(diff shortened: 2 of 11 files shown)


indico/modules/categories/util.py
```diff
@@ -229,3 +230,8 @@ def serialize_category_role(role, legacy=True):
             'name': role.name,
             'identifier': f'CategoryRole:{role.id}',
         }
+
+
+def can_create_unlisted_events(user):
+    return (user.is_admin or unlisted_events_settings.get('enabled') and (not unlisted_events_settings.get('restricted')
+            or unlisted_events_settings.acls.contains_user('authorized_creators', user)))
```
ThiefMaster, line 237: I think this will be much more readable when splitting it into multiple statements:

```python
def can_create_unlisted_events(user):
    if not user or not unlisted_events_settings.get('enabled'):
        return False
    elif user.is_admin or not unlisted_events_settings.get('restricted'):
        return True
    else:
        return unlisted_events_settings.acls.contains_user('authorized_creators', user)
```

indico/modules/events/forms.py
```diff
@@ -102,3 +103,12 @@ class LectureCreationForm(EventCreationFormBase):
     person_link_data = EventPersonLinkListField(_('Speakers'))
     description = TextAreaField(_('Description'), widget=CKEditorWidget())
     theme = IndicoThemeSelectField(_('Theme'), event_type=EventType.lecture, allow_default=True)
+
+
+class UnlistedEventsForm(IndicoForm):
+    enabled = BooleanField(_('Enabled'), widget=SwitchWidget(), default=False)
+    restricted = BooleanField(_('Require permission'), [HiddenUnless('enabled', preserve_data=True)],
+                              widget=SwitchWidget())
+    authorized_creators = PrincipalListField(_('Permissions'), [HiddenUnless('enabled', preserve_data=True)],
```
ThiefMaster, line 112: "Authorized users" seems like a better title than "Permissions"

indico/modules/events/forms.py
```diff
@@ -102,3 +103,12 @@ class LectureCreationForm(EventCreationFormBase):
     person_link_data = EventPersonLinkListField(_('Speakers'))
     description = TextAreaField(_('Description'), widget=CKEditorWidget())
     theme = IndicoThemeSelectField(_('Theme'), event_type=EventType.lecture, allow_default=True)
+
+
+class UnlistedEventsForm(IndicoForm):
+    enabled = BooleanField(_('Enabled'), widget=SwitchWidget(), default=False)
+    restricted = BooleanField(_('Require permission'), [HiddenUnless('enabled', preserve_data=True)],
```
ThiefMaster, line 110: How about "Restricted" or "Restrict creation"? And maybe a description like "Restrict creation of unlisted events to the authorized users below"

indico/modules/events/templates/admin/unlisted_events.html
```diff
@@ -0,0 +1,14 @@
+{% extends 'layout/admin_page.html' %}
+{% from 'forms/_form.html' import simple_form %}
+
+{% block title %}
+    {% trans %}Unlisted events{% endtrans %}
+{% endblock %}
+
+{% block description %}
+    {% trans %}Unlisted events enable users to set up and customise events before publishing them into a category.{% endtrans %}
```
ThiefMaster, line 9: Source strings are en_US

```suggestion
    {% trans %}Unlisted events enable users to set up and customize events before publishing them into a category.{% endtrans %}
```

indico/modules/categories/util.py
ThiefMaster: and maybe a small unit test for that function would be nice

indico/modules/categories/util_test.py
```diff
@@ -0,0 +1,27 @@
...
+from indico.modules.categories.util import can_create_unlisted_events
+from indico.modules.events.settings import unlisted_events_settings
+
+
+@pytest.mark.usefixtures('db')
+def test_can_create_unlisted_events(dummy_user, create_user):
+    assert not can_create_unlisted_events(None)
+    unlisted_events_settings.set('enabled', False)
+    unlisted_events_settings.set('restricted', False)
+    assert not can_create_unlisted_events(dummy_user)
```
ThiefMaster, line 19: I'd check for the `admin_user` here as well (especially since the initial version of the PR did not disable it for admin users even if the feature was disabled)

ThiefMaster (approved):

PR #5100, opened 2021-09-29
possibility to specify  smtp key and certfile through config

dweinholz:
Currently it is not possible to pass a cert/key file for SMTP in the config.
The missing variables have been added in this PR


docs/source/config/settings.rst
```diff
@@ -488,6 +488,18 @@ Emails
 
     Default: ``False``
 
+.. data:: SMTP_CERTFILE
+
+    If provided, this certfile will be used for the encrypted SMTP connection
```
ThiefMaster, line 493:
```suggestion
    If provided, this certificate will be used for SMTP authentication.
```

docs/source/config/settings.rst
```diff
@@ -488,6 +488,18 @@ Emails
...
+
+    Default: ``None``
+
+.. data:: SMTP_KEYFILE
+
+    If provided, this keyfile will be used for the encrypted SMTP connection
```
ThiefMaster, line 499:
```suggestion
    If provided, this private key will be used for client certificate SMTP authentication.
```

indico/core/config.py
```diff
@@ -79,6 +79,8 @@
     'SMTP_TIMEOUT': 30,
     'SMTP_USE_CELERY': True,
     'SMTP_USE_TLS': False,
+    'SMTP_CERTFILE': None,
+    'SMTP_KEYFILE': None,
```
ThiefMaster, line 83: Please keep the alphabetic order in this list :)

indico/web/flask/app.py
```diff
@@ -153,7 +153,8 @@ def configure_emails(app, config):
     app.config['EMAIL_USE_TLS'] = config.SMTP_USE_TLS
     app.config['EMAIL_USE_SSL'] = False
     app.config['EMAIL_TIMEOUT'] = config.SMTP_TIMEOUT
-
```
ThiefMaster, line 156: Don't remove the blank line; this now violates PEP8 (2 blank lines between top-level functions)

ThiefMaster (changes requested):

ThiefMaster:
Currently we only support STARTTLS. Does client cert auth work fine with that or does it rely on TLS from the beginning (`EMAIL_USE_SSL` setting)? If not maybe it's a good opportunity to also add `SMTP_USE_SSL` for `EMAIL_USE_SSL`.

dweinholz:
> 
> 
> Currently we only support STARTTLS. Does client cert auth work fine with that or does it rely on TLS from the beginning (`EMAIL_USE_SSL` setting)? If not maybe it's a good opportunity to also add `SMTP_USE_SSL` for `EMAIL_USE_SSL`.

I only tested it with SMTP_USE_TLS - which works fine

ThiefMaster (approved):

PR #5102, opened 2021-10-01
Use nicer select field for SUI colors

ThiefMaster:
We had only a rather basic dropdown (with untranslated color names) for this.. and now with #5091 adding such a color picker in a part of Indico that is not restricted to admins, it deserves something nicer.

image


tomasr8 (approved):
Works nicely :) I think we can merge it

PR #5105, opened 2021-10-05
Modify gettext to look through plural translations

### LINKED ISSUE #5073: "Convener" tag name not localizable
**Describe the bug**
The reviewing area shows labels "Reviewer" and "Convener" where appropriate for a given user, so it's clear who can handle what in which track. For some reason, the string "Convener" is not localizable via Transifex - search for it returns nothing.
"Reviewer" is there just fine.

**To Reproduce**
1. Create a conference with tracks
2. Try to assign roles to the users using the Roles in abstract submission setup
3. Switch to some other localization, like Turkish
4. See the "Convener" string stand out in the page.

**Expected behavior**
"Convener" string should be localizable.

tomasr8:
Closes #5073 

This is kind of a hacky solution so I'm open to suggestions..

Basically, the problem is that if there is a plural translation ,e.g.,

```
# Polish
msgid "Convener"
msgid_plural "Conveners"
msgstr[0] "Lider"
```
the only way to translate `Convener` to `Lider` is to use `ngettext(..., 1)`. `gettext` ignores
the plural translations when looking for a translation.
The solution is to use our own `gettext` with a fallback to `ngettext` if a translation is not found.

I don't know if this 100% foolproof as there might be some languages where a normal translation
differs from its singular translation when used with numbers.

### DIFF
--- a/indico/util/i18n.py
+++ b/indico/util/i18n.py
@@ -50,12 +50,52 @@ def get_translation_domain(plugin_name=_use_context):
             return get_domain()
 
 
+def _gettext(tr, message):
+    """Look for a translation in both the standard and plural translations.
+
+    Same as gettext.GNUTranslations.gettext but when the translation is not found,
+    it looks for a translation in plurals to find the singular case.
+    When neither is found, the gettext fallback is used.
+
+    This is needed because by default, gettext() only looks through non-pluralized translations.
+    For example, the following cannot be translated with gettext:
+
+    # Polish
+    msgid "Convener"
+    msgid_plural "Conveners"
+    msgstr[0] "Lider"
+
+    To get a translation for 'Convener', we need ngettext('Convener', 'Conveners', 1).
+
+    See GNUTranslations.gettext & ngettext for more details.
+    """
+    translation = tr.gettext(message)
+    if message != translation:
+        return translation
+    return tr.ngettext(message, message, 1)
+
+
+def _pgettext(tr, context, message):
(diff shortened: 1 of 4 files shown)


indico/util/i18n.py
```diff
@@ -50,12 +50,48 @@ def get_translation_domain(plugin_name=_use_context):
             return get_domain()
 
 
+def __gettext(tr, message):
```
ThiefMaster, line 53: Do we really need a double-underscore name here?

indico/util/i18n.py
```diff
@@ -50,12 +50,48 @@ def get_translation_domain(plugin_name=_use_context):
...
+
 def _indico_gettext(*args, **kwargs):
     func_name = kwargs.pop('func_name', 'gettext')
     plugin_name = kwargs.pop('plugin_name', None)
 
     translations = get_translation_domain(plugin_name).get_translations()
-    return getattr(translations, func_name)(*args, **kwargs)
+
+    if func_name == 'gettext':
+        return __gettext(translations, *args, **kwargs)
```
ThiefMaster, line 92: Don't we need the special logic for `pgettext` as well (gettext w/ context)?

ThiefMaster (commented):
I think having a unit test for this would be useful.

indico/util/i18n.py
tomasr8: yup, added

ThiefMaster (approved):

PR #5113, opened 2021-10-08
Validate regform field data using schemas

### LINKED ISSUE #5112: Registration form management: Schemas for field data
Currently we can put any kind of crap in the field data of a registration form field. This it not a problem per se since any unknown data will not be used, but it's ugly nonetheless.

Let's see if we can't use marshmallow schemas to validate this stuff properly.

Related code: `_fill_form_field_with_data`, `RegistrationFormFieldBase.process_field_data`

ThiefMaster:
There are some cases where you can get validation errors due to bugs in the legacy angular code (some fields allowing both empty and 0, or simply not having a value depending on whether or not you touched them); but that's something to fix when moving the frontend to react.

closes #5112

### DIFF
--- a/indico/modules/events/registration/controllers/management/fields.py
+++ b/indico/modules/events/registration/controllers/management/fields.py
@@ -6,28 +6,75 @@
 # LICENSE file for more details.
 
 from flask import jsonify, request, session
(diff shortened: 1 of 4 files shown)


ThiefMaster:
Amended the merged commit in 2ac3e63831 (fixed `missing` deprecation)

Issue #5127, opened 2021-10-18
Privacy features (UI mockups)


tomasr8:
Mockups for privacy-related features. To be communicated to Andrei and Thomas once they're ready.

https://github.com/indico/indico/issues/1415#issuecomment-943173437

PR #5128, opened 2021-10-20
Add basic Privacy dashboard

tomasr8:
Adds basic Privacy dashboard functionality as outlined in #5127 
This PR includes:
- A new event management page (Privacy dashboard)
- A new footer item which links to the privacy policy of the current event

#### Dashboard:
screenshot

#### Privacy policy footer:
image


indico/migrations/versions/20211027_0930_82fb6c6ac6db_add_publish_registrations_with_consent.py
```diff
@@ -0,0 +1,52 @@
...
+
+class _PublishRegistrationsMode(int, Enum):
+    hide_all = 0
+    show_with_consent = 1
+    show_all = 2
+
+
+def upgrade():
+    op.add_column('forms',
+                  sa.Column('publish_registrations_mode', PyIntEnum(_PublishRegistrationsMode), server_default='1',
```
ThiefMaster, line 30:
```suggestion
                  sa.Column('publish_registrations_mode', PyIntEnum(_PublishRegistrationsMode), server_default='0',
```

indico/migrations/versions/20211027_0930_82fb6c6ac6db_add_publish_registrations_with_consent.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def downgrade():
+    op.add_column('forms',
+                  sa.Column('publish_registrations_enabled', sa.BOOLEAN(), server_default='false', nullable=False),
+                  schema='event_registration')
+    op.execute('''
+        UPDATE event_registration.forms
+        SET publish_registrations_enabled = true
+        WHERE publish_registrations_mode != 0
```
ThiefMaster, line 49:
```suggestion
        WHERE publish_registrations_mode = 2
```

I think this is safer - that way doing a downgrade+upgrade won't suddenly fully enable something that wasn't enabled before.

indico/migrations/versions/20211027_0930_82fb6c6ac6db_add_publish_registrations_with_consent.py
```diff
@@ -0,0 +1,52 @@
...
+    op.execute('UPDATE event_registration.forms SET publish_registrations_mode = 2 WHERE publish_registrations_enabled')
+    op.drop_column('forms', 'publish_registrations_enabled', schema='event_registration')
+    op.add_column('registrations',
+                  sa.Column('consented_to_publish', sa.Boolean(), nullable=False, server_default='false'),
+                  schema='event_registration')
+    op.alter_column('registrations', 'consented_to_publish', server_default=None, schema='event_registration')
+
+
+def downgrade():
+    op.add_column('forms',
```
ThiefMaster, line 43: After creating this column you also need to remove the `server_default` again:

    op.alter_column('forms', 'publish_registrations_enabled', server_default=None, schema='event_registration')

indico/migrations/versions/20211027_0930_82fb6c6ac6db_add_publish_registrations_with_consent.py
```diff
@@ -0,0 +1,52 @@
...
+                  sa.Column('publish_registrations_enabled', sa.BOOLEAN(), server_default='false', nullable=False),
```
ThiefMaster, line 44:
```suggestion
                  sa.Column('publish_registrations_enabled', sa.Boolean(), server_default='false', nullable=False),
```

ThiefMaster:
Can we find a better icon or add some other lock icon to our icon set? The alignment of the lock is REALLY weird: 
image

indico/modules/events/management/forms.py
```diff
@@ -230,6 +230,17 @@ def __init__(self, *args, **kwargs):
...
+class PrivacyDashboardForm(IndicoForm):
+    _data_controller_fields = ('data_controller_name', 'data_controller_email')
+    _privacy_policy_fields = ('privacy_policy_url', 'privacy_policy')
+    data_controller_name = StringField(_('Person/Institution'))
+    data_controller_email = EmailField(_('Contact email'), [Optional(), Email()])
+    privacy_policy_url = URLField(_('URL'), [Optional(), URL()],
+                                  description=_('The URL to an external page with the privacy policy'))
+    privacy_policy = TextAreaField(_('Text'), widget=CKEditorWidget(),
+                                   description=_('Only used if no URL is provided'))
+
```
ThiefMaster, line 242: Some validation logic that disallows providing both a url and text would be useful. In the admin UI it's not so important but I'm sure we have some event managers that don't realize the fields are mutually exclusive..

Something like this should work (untested):

```python
    def validate_privacy_policy(self, field):
        if self.privacy_policy_url.data and self.privacy_policy.data:
            raise ValidationError(_('...'))

    privacy_policy_url = privacy_policy
```

indico/modules/events/operations.py
```diff
@@ -410,3 +412,22 @@ def create_event_request(event, category, comment=''):
...
+                  'data_controller_email': 'Data controller e-mail',
+                  'privacy_policy_url': 'Privacy policy URL',
+                  'privacy_policy': 'Privacy policy'}
+    assert set(data.keys()) <= log_fields.keys()
+    changes = {}
+    for key, value in data.items():
+        old_value = privacy_settings.get(event, key)
+        if old_value != value:
+            # XXX: we copy because of https://github.com/sqlalchemy/sqlalchemy/issues/3913
+            changes[key] = (copy(old_value), copy(value))
```
ThiefMaster, line 428: Since we don't deal with any sqlalchemy collections here, copying isn't needed

indico/web/templates/_sortable_list.html
```diff
@@ -41,28 +41,29 @@
...
-                {% call(item) sortable_items(enabled_items, draggable, true, invisible_handle) %}
-                    {{ _caller(item) if _caller else '' }}
-                {% endcall %}
-            </ul>
-        </div>
-        <div>
-            <div class="titled-rule">{{ disabled_title }}</div>
-            <ul class="disabled">
-                {% call(item) sortable_items(disabled_items, draggable, true, invisible_handle) %}
+            <div class="titled-rule">{{ list.0 }}</div>
```
ThiefMaster, line 53: I'd rather use bracket notation for numeric keys. It looks really strange with dot notation ;)

indico/modules/events/registration/controllers/management/regforms.py
```diff
@@ -51,16 +52,30 @@ class RHManageRegistrationFormsDisplay(RHManageRegFormsBase):
 
     def _process(self):
         regforms = sorted(self.event.registration_forms, key=lambda f: f.title.lower())
-        form = ParticipantsDisplayForm()
+        forbidden_visibility_forms = []
```
ThiefMaster, line 55: a set would be the better choice here

indico/modules/events/registration/controllers/management/regforms.py
```diff
@@ -51,16 +52,30 @@ class RHManageRegistrationFormsDisplay(RHManageRegFormsBase):
...
+        for regform in regforms:
+            if (regform.publish_registrations_mode != PublishRegistrationsMode.show_all and
+                    regform.existing_registrations_count > 0):
+                forbidden_visibility_forms.append(regform.id)
```
ThiefMaster, line 59: And you could make the whole thing a set comprehension:

```python
        forbidden_visibility_forms = {
            rf.id for rf in regforms if (
                rf.publish_registrations_mode != PublishRegistrationsMode.show_all and
                rf.existing_registrations_count > 0
            )
        }
```

indico/modules/events/registration/templates/management/regform_display.html
```diff
@@ -145,6 +153,41 @@
...
+                    position: {
+                        my: 'left middle',
+                        at: 'right middle',
+                    },
+                    show: '',
+                    hide: '',
+                });
+                $item.addClass('forbidden-visibility');
+            });
+            var invalid_forms = 0;
```
ThiefMaster, line 176: Don't use `var` in new code (use `let` or `const` depending on the usage); and JS uses camelCase

indico/modules/events/registration/templates/management/regform_display.html
```diff
@@ -145,6 +153,41 @@
...
+            $('.js-sortable-list-widget .show-all').on('sortreceive', function(evt, ui) {
+                if (ui.item.hasClass('forbidden-visibility')) {
+                    ui.item.qtip('show');
+                    invalid_forms++;
+                    $form.find(":submit").attr('disabled', 'disabled');
```
ThiefMaster, line 181: `.prop('disabled', true)`

indico/modules/events/registration/templates/management/regform_display.html
```diff
@@ -145,6 +153,41 @@
...
+                }
+            }).on('sortremove', function(evt, ui) {
+                if (ui.item.hasClass('forbidden-visibility')) {
+                    ui.item.qtip('hide');
+                    invalid_forms--;
+                    if (invalid_forms === 0)
+                        $form.find(":submit").removeAttr('disabled');
```
ThiefMaster, line 188:
```suggestion
                    if (invalid_forms === 0) {
                        $form.find(":submit").prop('disabled', false);
                    }
```

indico/modules/events/registration/templates/management/regform_display.html
ThiefMaster: and we usually use single quotes *except* when that would cause backslash escaping within the string

indico/modules/events/registration/controllers/management/regforms.py
duartegalvao: This breaks the current way that the list is passed to the JS (`{{ forbidden_visibility_forms|tojson }}`). What Is the best way to solve this? I guess we could either iterate the set server-side in jinja, or convert to list before passing it to the template

ThiefMaster: You can use `...|list|tojson` in the template

ThiefMaster:
Some things that came up during the meeting with the ODP today:

- It's called "privacy notice" and not "privacy policy" (just a simply wording change; i think we should just change it in user-facing places though and *not* rename the `privacy_policy` settings key, since we already use that in other places and having two similar but different names isn't amazing)

- It'd probably be a good idea to allow multiple privacy notice URLs as well (e.g. for events that have different privacy notices for speakers and regular participants); otherwise people have to use the freetext field to put those links - but I think we should put this in a separate task

ThiefMaster:
I have the feeling including the registration stuff in the same PR was a bit of a mistake - can we extract the general privacy dashboard commits into a separate PR (or vice versa) so this one just adds the basic privacy dashboard with the privacy notice and data controller setting? Because that one could be merged quickly, while all the participant visibility settings are much more complex (and need some more changes after the ODP discussion)...

indico/modules/events/templates/footer.html
```diff
@@ -1,5 +1,24 @@
 {% extends 'footer.html' %}
 
+{% block footer_links_extra %}
+    {% if privacy_url %}
+        <li>
+            <a href="{{ privacy_url }}">{% trans %}Event Privacy Policy{% endtrans %}</a>
```
ThiefMaster, line 6: s/Policy/Notice/ (and let's keep the capitalization consistent between here and below)

duartegalvao: oops, missed that one. thanks!

indico/modules/events/management/forms.py
javfg: Adding to this, and keeping in mind the suggested change by Adrian might be so you can define an explicit error message. But maybe it would be worth considering using the `Exclusive` validator instead of custom logic:

```python
    privacy_policy_url = URLField(_('URL'), [Optional(), URL(), Exclusive('privacy_policy')],
                                  description=_('The URL to an external page with the privacy notice'))
    privacy_policy = TextAreaField(_('Text'), [Exclusive('privacy_policy_url')], widget=CKEditorWidget(),
                                   description=_('Only used if no URL is provided'))
```

Issue #3644, opened 2018-10-29
Make it clear on the interface that changing a "person link" to someone else is wrong


pferreir:
This is not at all clear right now. We should probably add some kind of warning or explanation here, that illustrates that the person's identity **won't change**.
![image](https://user-images.githubusercontent.com/2699/47661018-881e5000-db98-11e8-9392-689a18d7730a.png)

ThiefMaster:
Let's use the opportunity to move the dialog and the field where it's used to React. There's a ton of ultra legacy JS involved.

PR #5129, opened 2021-10-22
Refactor PersonLinkField

plourenco:
A very drafty version of the `PersonLinkField`, with shallow support of the previous Jinja-based options.

- [x] Move all serialization util classes to schemas
- [x] Handle all the previous flags as React props
- [x] Simplify the usage of `PersonListItem`
- [x] Update messages depending on the section
- [x] Add labels for icon-based roles
- [x] #3644


indico/web/templates/forms/person_link_widget.html
```diff
@@ -0,0 +1,45 @@
...
+
+{% block javascript %}
+    <script>
+        setupPersonLinkWidget({
+            fieldId: {{ field.id | tojson }},
+            eventId: {{ (field.event.id if field.event else none) | tojson }},
+            roles: {{ field.roles | default([]) | tojson }},
+            sortByLastName: {{ field.sort_by_last_name | default(false) | tojson }},
+            emptyMessage: {{ field.empty_message | default(none) | tojson }},
+            sessionUser: {{ ({
```
ThiefMaster, line 30: This is already available in the global `Indico.User`

indico/modules/events/fields.py
```diff
@@ -174,19 +169,27 @@ class EventPersonLinkListField(PersonLinkListFieldBase):
...
-    def _serialize_person_link(self, principal, extra_data=None):
-        data = (extra_data or {}) | serialize_person_link(principal)
-        data['isSubmitter'] = self.data[principal] if self.get_form().is_submitted() else principal.is_submitter
+        return {self._get_person_link(x): 'submitter' in x.get('roles', []) for x in data}
+
+    def _serialize_person_link(self, principal):
+        from indico.modules.events.persons.schemas import PersonLinkSchema
+        data = PersonLinkSchema().dump(principal)
+        data['roles'] = []
+        if self.get_form().is_submitted() and self.data[principal] or (principal.event and principal.is_submitter):
```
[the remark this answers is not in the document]
plourenco: Yes, I recall that, added on the second clause but slipped the previous one!

indico/modules/events/abstracts/forms.py
```diff
@@ -472,6 +470,11 @@ def __init__(self, *args, **kwargs):
         self.abstract = kwargs.pop('abstract', None)
         is_invited = kwargs.pop('invited', False)
         management = kwargs.pop('management', False)
+        self.person_links = AbstractPersonLinkListField(
+            _('Authors'),
+            allow_speakers=not is_invited and abstracts_settings.get(self.event, 'allow_speakers'),
+            require_primary_author=not is_invited,
+            require_speaker=not is_invited and abstracts_settings.get(self.event, 'speakers_required'))
```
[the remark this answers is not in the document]
plourenco: Yes, was wondering whether I should keep it

indico/modules/events/contributions/fields.py
```diff
@@ -19,30 +18,51 @@ class ContributionPersonLinkListField(PersonLinkListFieldBase):
...
+    widget = JinjaWidget('forms/person_link_widget.html', allow_empty_email=True)
+
+    @property
+    def roles(self):
+        roles = [{'name': 'speaker', 'label': _('Speaker'), 'icon': 'microphone', 'default': True}]
+        if self.allow_submitters:
+            roles.append({'name': 'submitter', 'label': _('Submitter'), 'icon': 'paperclip',
+                          'default': self.default_is_submitter})
+        if self.allow_authors:
+            roles.extend([
```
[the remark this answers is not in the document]
plourenco: Sure, I can do both

indico/modules/events/fields.py
```diff
@@ -127,31 +125,28 @@ def __init__(self, *args, **kwargs):
...
         if self.object and inspect(person).persistent:
             person_link = self.person_link_cls.query.filter_by(person=person, object=self.object).first()
         if not person_link:
             person_link = self.person_link_cls(person=person)
-        person_link.populate_from_dict(person_data)
+        person_link.populate_from_dict(PersonLinkSchema(
+            only=('first_name', 'last_name', 'affiliation', 'address', 'phone', 'display_order')).load(data))
         email = data.get('email', '').lower()
+        if not email:
+            raise UserValueError(_('A valid email address is required'))
```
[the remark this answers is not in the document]
plourenco: That's a good idea, but I'm afraid we might hit a lot of corner cases where an email was previously expected.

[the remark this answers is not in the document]
ThiefMaster: Yes, certainly not something for right now.

indico/web/templates/forms/person_link_widget.html
plourenco: Even though it doesn't support attributes such as `phone` or `address` but it seems perfectly reasonable to support them in a different task.

ThiefMaster: those could be included

indico/modules/events/fields.py
```diff
@@ -121,72 +77,86 @@ class PersonLinkListFieldBase(EventPersonListField):
     default_sort_alpha = True
 
     widget = None
+    create_untrusted_persons = False
 
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+        super().__init__(*args, allow_groups=False, allow_external_users=True, **kwargs)
         self.object = getattr(kwargs['_form'], self.linked_object_attr, None)
+        self.event = self.object.event
```
plourenco, line 85: This is `None` when the event type is a lecture because no event attribute is present in the object.

```
AttributeError: 'NoneType' object has no attribute 'event
```

indico/modules/events/fields.py
```diff
@@ -121,72 +77,90 @@ class PersonLinkListFieldBase(EventPersonListField):
...
+        return [{'name': 'submitter', 'label': _('Submitter'), 'icon': 'paperclip',
+                 'default': self.default_is_submitter}]
 
     def __init__(self, *args, **kwargs):
-        self.allow_submitters = True
         self.default_is_submitter = kwargs.pop('default_is_submitter', True)
+        self.empty_message = _('There are no chairpersons')
+        event_type = kwargs.pop('event_type', EventType.conference)
         super().__init__(*args, **kwargs)
+        if event_type or (self.object.event and self.object.event.type_) == EventType.lecture:
```
ThiefMaster, line 152: 🤯  

But I don't think `event_type` is ever falsy... And if I understand it correctly, you always set either an event type via kwarg or have an event... so you could probably simplify this to something like this:

```python
event_type = self.object.event.type_ if self.object and self.object.event else kwargs.pop('event_type')
```

(can we ever have `self.object` without an event?)

indico/modules/events/persons/schemas.py
```diff
@@ -5,17 +5,61 @@
...
+    address = fields.String(load_default='')
+    email = fields.String(required=True)
+    display_order = fields.Int(load_default=0, dump_default=0)
+    avatar_url = fields.Function(lambda o: o.person.user.avatar_url if o.person.user else None)
+
+    def get_title(self, obj):
+        return obj.title
+
+    def load_title(self, title):
+        return next((x.value for x in UserTitle if title == orig_string(x.title)), UserTitle.none)
```
ThiefMaster, line 38: maybe this would be a good opportunity to send the enum name instead of the title itself? that would allow you to use an `EnumField(UserTitle)`

indico/modules/events/persons/schemas.py
```diff
@@ -5,17 +5,61 @@
...
 from indico.modules.events.models.persons import EventPerson
+from indico.modules.users.models.users import UserTitle
+from indico.util.i18n import orig_string
+
+
+class PersonLinkSchema(mm.Schema):
+    class Meta:
+        unknown = EXCLUDE
+
+    _type = fields.Constant('PersonLink')
```
ThiefMaster, line 20: Does this need to interact with any legacy code that expects `_type`? If not let's maybe use just `type`

indico/modules/events/persons/schemas.py
```diff
@@ -5,17 +5,61 @@
...
+    title = fields.Method('get_title', deserialize='load_title')
+    affiliation = fields.String(load_default='')
+    phone = fields.String(load_default='')
+    address = fields.String(load_default='')
+    email = fields.String(required=True)
+    display_order = fields.Int(load_default=0, dump_default=0)
+    avatar_url = fields.Function(lambda o: o.person.user.avatar_url if o.person.user else None)
+
+    def get_title(self, obj):
+        return obj.title
```
ThiefMaster, line 35: I think this will break editing when you use a language where the titles are translated to something else (since this one will return the translated version, and that won't match any of the dropdown options client-side)

indico/web/client/js/react/components/PersonLinkField.jsx
```diff
@@ -0,0 +1,400 @@
...
+const ExternalPersonModal = ({onSubmit, onClose, person}) => (
+  <FinalModalForm
+    id="person-link-details"
+    size="tiny"
+    onClose={onClose}
+    onSubmit={onSubmit}
+    header={Translate.string('Enter Person')}
+    initialValues={person || {}}
+  >
+    {person && person.userId && (
```
ThiefMaster, line 44: indico instances migrated from 1.x have a user with ID `0`, so this needs to be an explicit `null` check

indico/web/client/js/react/components/PersonLinkField.jsx
```diff
@@ -0,0 +1,400 @@
...
+export default function PersonLinkField({
+  value: persons,
+  onChange,
+  eventId,
+  sessionUser,
+  roles,
+  emptyMessage,
+}) {
+  const [favoriteUsers] = useFavoriteUsers();
+  const [modalOpen, setModalOpen] = useState(false);
```
ThiefMaster, line 246: Not needed; `selected !== null` covers it already.

indico/web/client/js/react/components/principals/items.jsx
```diff
@@ -39,6 +35,92 @@ PendingPrincipalListItem.defaultProps = {
...
+  children: PropTypes.node,
+};
+
+PrincipalItemContent.defaultProps = {
+  detail: null,
+  children: null,
+};
+
+export const PrincipalItem = ({as, children, className, ...rest}) => {
+  const Component = as || 'div';
```
ThiefMaster, line 104: Putting `as: 'div'` in `defaultProps` doesn't work?

indico/modules/events/fields.py
plourenco: I had the feeling this line was problematic, it became confusing because we can't access `self.object` before the `super` as it doesn't exist until there, plus `event_type` needs to be "pop" before.

> (can we ever have self.object without an event?)

In the `LectureCreationForm`

indico/web/client/js/react/components/PersonLinkField.jsx
plourenco: Good point!

indico/web/client/js/react/components/PersonLinkField.jsx
plourenco: But we might not have it while manually inserting a person.

```
<Translate as={Button} type="button" onClick={() => setModalOpen(true)}>
    Enter manually
</Translate>
```

indico/web/client/js/react/components/principals/items.jsx
plourenco: It does, even though it's a bit more loose.

indico/modules/events/fields.py
ThiefMaster: what's the object there? the category?

indico/web/client/js/react/components/PersonLinkField.jsx
ThiefMaster: ah, missed that one...

indico/modules/events/fields.py
plourenco: The object is `self.object = getattr(kwargs['_form'], self.linked_object_attr, None)` where `linked_object_attr` is deemed to be an `event` or `contribution`, etc. There is none as the lecture is still being created.

plourenco: My latest changes should make it more clear.

ThiefMaster: so `self.object` is already none - which is what i was asking about in "can we ever have self.object without an event?"

plourenco: Ah, yes - I don't recall exactly, I believe not with the current changes.

ThiefMaster:
The fix from #5200 needs to be included here as well (in a different way of course) - when you are not logged in you can still submit an *invited abstract* using the UUID link from the email, and in that case the only option should be "Enter manually" (ie no search/add-myself and no failing request to load favorite users).

indico/modules/events/abstracts/fields.py
```diff
@@ -99,44 +98,56 @@ class AbstractPersonLinkListField(PersonLinkListFieldBase):
...
         super().pre_validate(form)
         for person_link in self.data:
-            if not self.allow_authors and person_link.author_type != AuthorType.none:
-                if not self.object_data or person_link not in self.object_data:
-                    person_link.author_type = AuthorType.none
             if person_link.author_type == AuthorType.none and not person_link.is_speaker:
                 raise ValidationError(_('{} has no role').format(person_link.full_name))
+        if (self.require_primary_author and
+                not any(person_link.author_type == AuthorType.primary for person_link in self.data)):
+            raise ValueError(_('You must add at least one author'))
```
ThiefMaster, line 148:
```suggestion
            raise ValidationError(_('You must add at least one author'))
```

indico/modules/events/abstracts/fields.py
```diff
@@ -99,44 +98,56 @@ class AbstractPersonLinkListField(PersonLinkListFieldBase):
...
+        if self.require_speaker and not any(person_link.is_speaker for person_link in self.data):
+            raise ValueError(_('You must add at least one speaker'))
```
ThiefMaster, line 150:
```suggestion
            raise ValidationError(_('You must add at least one speaker'))
```

indico/web/client/js/react/components/PersonLinkField.jsx
```diff
@@ -0,0 +1,401 @@
...
+  {text: Translate.string('Mx'), value: 'mx'},
+];
+
+const ExternalPersonModal = ({onSubmit, onClose, person}) => (
+  <FinalModalForm
+    id="person-link-details"
+    size="tiny"
+    onClose={onClose}
+    onSubmit={onSubmit}
+    header={Translate.string('Enter Person')}
```
ThiefMaster, line 40:
```suggestion
    header={Translate.string('Enter Person')}
    submitLabel={Translate.string('Save')}
```

the default "Submit" sounds rather weird since the dialog itself doesn't submit anything yet

ThiefMaster:
Just so we don't forget these things in the new year:

- "Add myself" still includes a ton of unnecessary data in the hidden field (e.g. `is_admin`, `favorite_users`, ...). Maybe we can use a marshmallow schema (with the default behavior of raising an exception in case of unexpected data) to process the incoming data? That way we quickly notice if anything beyond what's actually expected/needed is sent..
- Having "Title" as a placeholder looks strange; let's just keep the field empty - there's already a label indicating what it is anyway... image
- The title logic is a bit strange, the `person` in the `ExternalPersonModal` has a `'none'` string initially, but clearing the title with the x in the dropdown sets it to an empty string. Maybe we can unify this to be empty in all cases? Especially when removing the placeholder you otherwise get different behavior when opening the dialog with an empty title ("x" present) vs after clearing it ("x" not present, since there's nothing to clear)
- The widget being much wider than the other form field still feels odd...

However, this this PR got huge and is a pain to review, I'll merge it and the things above can be fixed in a separate PR...

PR #5139, opened 2021-11-03
Add forward, backward and background color to badge designer 

SegiNyn:


indico/modules/designer/client/js/index.js
```diff
@@ -424,8 +430,63 @@ import {$T} from 'indico/utils/i18n';
...
+      }
+      if (itemsOverlap(selectedItem, items[i])) {
+        if (moveToFront) {
+          zIndex = Math.max(items[i].zIndex, zIndex);
+        } else {
+          zIndex = Math.min(items[i].zIndex, zIndex);
+        }
+      }
+    }
+    const randomZIndexValue = Math.floor(Math.random() * 10) + 5;
```
ThiefMaster, line 483: Why is this needed? Using randomness seems really strange here.

indico/modules/events/registration/badges.py
```diff
@@ -83,7 +83,8 @@ def _draw_badge(self, canvas, registration, template, tpl_data, pos_x, pos_y):
 
         # Print images first
         image_placeholders = {name for name, placeholder in placeholders.items() if placeholder.is_image}
-        items = sorted(tpl_data.items, key=lambda item: item['type'] not in image_placeholders)
+        items = sorted(tpl_data.items, key=lambda item: (int(item['zIndex']) if 'zIndex' in item else 65535,
```
ThiefMaster, line 86: You can use `.get()` here

indico/modules/designer/client/js/index.js
SegiNyn: I can change this to always increase/decrease by 1. it would work also

CHANGES.rst
```diff
@@ -102,6 +102,7 @@ Improvements
 - Add an API for bulk-assigning contribution program codes programmatically (:issue:`5115`,
   :pr:`5120`)
 - Add layout setting to show videoconferences on the main conference page (:pr:`5124`)
+- Add forward, backward and background color option to badge designer(:pr:`5139`, thanks :user:`seginyn`)
```
ThiefMaster, line 105: should be in the 3.1 changelog

indico/modules/designer/client/js/index.js
```diff
@@ -424,8 +430,62 @@ import {$T} from 'indico/utils/i18n';
...
+
+    if (top1 >= bottom2 || top2 >= bottom1) {
+      return false;
+    }
+    return true;
+  }
+
+  function getZIndex(selectedItem, moveToFront) {
+    let zIndex = selectedItem.zIndex;
+    for (const i in items) {
```
ThiefMaster, line 471:
```js
Object.values(items).forEach(item => {
  ...
});
```

and then use `item` instead of `items[i]` in there.

indico/modules/designer/pdf.py
```diff
@@ -97,6 +97,8 @@ def _draw_item(self, canvas, item, tpl_data, content, margin_x, margin_y):
         style = ParagraphStyle({})
         style.alignment = ALIGNMENTS[item['text_align']]
         style.textColor = COLORS[item['color']]
+        style.backColor = item['background_color'] if item['background_color'] else None
```
ThiefMaster, line 100: I don't think we have a `background_color` for older badges, do we?

```suggestion
        style.backColor = item.get('background_color') or None
```

the ` or None` can be omitted if you do not need to normalize other falsy values such as empty strings to None

indico/modules/designer/templates/template.html
```diff
@@ -191,7 +191,8 @@
                         {% endblock %}
                     </div>
                     <div class="subtool">
-                        <select id='color-selector' class="js-font-tool" data-attr="color">
+                        <select title="{% trans %}Text color{% endtrans %}"
+                            id='color-selector' class="js-font-tool" data-attr="color">
```
ThiefMaster, line 195: align with `title` or move into the previous line

indico/modules/designer/pdf.py
SegiNyn: That right, there is no `background_color` for older badges

plourenco (approved):
Added some minor suggestions but it works great, thanks! 🙌

indico/modules/designer/client/js/index.js
```diff
@@ -424,9 +430,54 @@ import {$T} from 'indico/utils/i18n';
...
+      return false;
+    }
+    return !(top1 >= bottom2 || top2 >= bottom1);
+  }
+
+  function getZIndex(selectedItem, moveToFront) {
+    let zIndex = selectedItem.zIndex;
+    Object.values(items)
+      .filter(item => selectedItem.id !== item.id)
+      .forEach(item => {
```
pferreir, line 469: Not a very large performance penalty, but it would be more efficient to do the ID check inside the `forEach(...)` closure.

SegiNyn:
Hi, I initially used the same fixed list of colours for the background but it was suggested in an initial review that using a color picker would be preferred

pferreir:
@SegiNyn sure. I'm just wondering whether we shouldn't apply the same to the bg color.

ThiefMaster:
Consistency would clearly be useful (when making the initial suggestion I probably didn't even realized that we already have a color selector there for something)

SegiNyn:
Yes it would be better to keep it consistent. I'll change the colour picker to the fixed list of colours.

indico/modules/designer/client/js/index.js
```diff
@@ -424,9 +430,54 @@ import {$T} from 'indico/utils/i18n';
...
+        if (itemsOverlap(selectedItem, item)) {
+          if (moveToFront) {
+            zIndex = Math.max(item.zIndex, zIndex);
+          } else {
+            zIndex = Math.min(item.zIndex, zIndex);
+          }
+        }
+      });
+    return zIndex === selectedItem.zIndex ? zIndex : zIndex + (moveToFront ? 1 : -1);
```
pferreir, line 478: Untested, but maybe this would work as well?

```js
let foremostItem = _.maxBy(Object.values(items).filter(item => (selectedItem.id !== item.id) && itemsOverlap(selectedItem, item)), 'zIndex');
let zIndex = Math.min(selectedItem.zIndex, foremostItem.zIndex + 1);
```

... and I would that directly within the `back(...)` and `front(...)`, functions, using `_.maxBy(...)` and `_.minBy(...)` depending on the case. Does that make sense?

indico/modules/designer/client/js/index.js
```diff
@@ -84,10 +84,12 @@ import {$T} from 'indico/utils/i18n';
       italic: false,
       text_align: 'center',
       color: 'black',
+      background_color: null,
       font_size: '15pt',
       width: isImage(type) ? 150 : 400,
       height: isImage(type) ? 150 : null,
       text: $T('Fixed text'),
+      zIndex: itemIdCounter + 100,
```
pferreir, line 92: Why 100 and not 10 like before?

indico/modules/events/registration/badges.py
```diff
@@ -83,7 +83,8 @@ def _draw_badge(self, canvas, registration, template, tpl_data, pos_x, pos_y):
 
         # Print images first
         image_placeholders = {name for name, placeholder in placeholders.items() if placeholder.is_image}
-        items = sorted(tpl_data.items, key=lambda item: item['type'] not in image_placeholders)
+        items = sorted(tpl_data.items, key=lambda item: (int(item.get('zIndex', 65535)),
```
pferreir, line 86: Not sure about 65535 as a default value. Do we want "legacy" items to show up in front by default? Wouldn't 0 be a better choice? And maybe a comment to explain the reasoning would be of help.

pferreir:
@SegiNyn Sorry, my bad, what I meant was changing the fixed list (fg color) to a picker.

SegiNyn:
Okay, got it. Thanks!

indico/modules/designer/client/js/index.js
```diff
@@ -159,8 +163,8 @@ import {$T} from 'indico/utils/i18n';
     }
 
     newDiv.draggable({
+      // stack: templateSide + ' > div' remove this so it does not interfere with the z-index
```
pferreir, line 166: If this is to be removed, let's just remove it. We don't usually commit commented-out code.

pferreir (changes requested):
Thanks once again. Just some minor comments/suggestions.

indico/modules/designer/client/js/index.js
SegiNyn: This is because background images have a z-index of 5 and since moving items forward/backward adds +/-1 to the z-index value, it could easily become less than 5, placing the item behind the background image. Which is why I used 100 instead of 10.

indico/modules/events/registration/badges.py
SegiNyn: I don't have a reason for choosing a higher value but we can't use 0 because that would place the item behind the background image.

indico/modules/designer/client/js/index.js
SegiNyn: Yes, this works. Thank you.

indico/modules/designer/client/js/index.js
pferreir: Yeah, I get it. However, it is still possible that, after a few runs this gets again to under 5. Could we maybe have a function which "normalizes" the z-indexes after each operation? E.g. sort them in growing order, assign them z-indices according to the sequence of natural numbers (1, 2, 3, 4...) and then add 9 to all? This would keep them to a minimum. It would also hold more consistent results when you move things around.

indico/modules/events/registration/badges.py
pferreir: Yeah, well, 10, in this case.

indico/modules/designer/client/js/index.js
SegiNyn: Yeah, this would work

indico/modules/designer/client/js/index.js
```diff
@@ -424,9 +429,62 @@ import {$T} from 'indico/utils/i18n';
...
+    }
+    return !(top1 >= bottom2 || top2 >= bottom1);
+  }
+
+  function normalizeZindex() {
+    var increament = 1;
+    Object.values(items)
+      .sort((item1, item2) => item1.zIndex - item2.zIndex)
+      .forEach(item => {
+        item.zIndex = increament + 9;
```
pferreir, line 483: "increment"

pferreir: Also, this can be rewritten as (untested):

```js
_.sortBy(Object.values(items), 'zIndex').forEach((item, n) => { item.zIndex = n + 10; });
```

indico/modules/designer/client/js/index.js
```diff
@@ -424,9 +429,62 @@ import {$T} from 'indico/utils/i18n';
...
+    const right2 = left2 + item2.width;
+    const bottom2 = top2 + (item2.height || 22);
+
+    if (left1 >= right2 || left2 >= right1) {
+      return false;
+    }
+    return !(top1 >= bottom2 || top2 >= bottom1);
+  }
+
+  function normalizeZindex() {
```
pferreir, line 478: I belive the right camel-case notation would be `normalizeZIndex`.

indico/modules/designer/client/js/index.js
SegiNyn: Yes, this works

pferreir:
OK, thanks a lot. I think we're good code-wise, but I've found a few issues testing it:

e.g.

image

which actually then renders OK, but the preview is weird.

Then, there is an issue with the z-index, probably side-effect of the algorithm we didn't foresee:

Kapture 2021-12-10 at 13 34 57

Finally, it's currently impossible to figure out when an item is selected, if it's got a background. I suggest we move from changing the background of the item to adding a thick border around, or something like that.

indico/modules/designer/client/js/index.js
```diff
@@ -454,7 +454,7 @@ import {$T} from 'indico/utils/i18n';
           var zIndex = Math.max(selectedItem.zIndex, foremostItem.zIndex + 1);
           div.css('zIndex', zIndex);
           selectedItem.zIndex = zIndex;
-          normalizeZindex();
+          normalizeZIndex();
```
pferreir, line 457: Maybe related to the bug we're seeing. Shouldn't we normalize before we set the CSS property?
Something like:
```js
selectedItem.zIndex = zIndex;
normalizeZIndex();
div.css('zIndex', selectedItem.zIndex);
```

indico/modules/designer/client/js/index.js
```diff
@@ -475,14 +475,10 @@ import {$T} from 'indico/utils/i18n';
...
-    var increament = 1;
-    Object.values(items)
-      .sort((item1, item2) => item1.zIndex - item2.zIndex)
-      .forEach(item => {
-        item.zIndex = increament + 9;
-        increament++;
-      });
+  function normalizeZIndex() {
+    _.sortBy(Object.values(items), 'zIndex').forEach((item, n) => {
+      item.zIndex = n + 10;
```
pferreir, line 480: Probably related to the bug I've pointed out: `normalizeZIndex` isn't actually setting any CSS properties.

SegiNyn:
Hi, "I suggest we move from changing the background of the item to adding a thick border around, or something like that." So whenever an item is selected, we add a think border around it so it is more noticeable which item was selected while still maintaining the background colour? or the background colour is removed and remains only a thick border when the item is selected?

indico/modules/designer/client/js/index.js
SegiNyn: Yes, I think it may be because of this because moving the `normalizeZIndex` call does not fix the bug

pferreir:
> So whenever an item is selected, we add a thick border around it so it is more noticeable which item was selected while still maintaining the background colour?

Yes, precisely!

SegiNyn:
Hello @pferreir, I updated the code by adding a thicker border to selected items, make the item height use its initial value so the text is not larger than the background and updated the html zIndex values in the normalizeZIndex function

SegiNyn:
I added a 'data-id' attribute for the divs so they can be easily selected using:
`$('.ui-draggable[data-id=' + item.id + ']')` 
but I'm not sure if you would want this attribute. We could also use: 
 ` $('.ui-draggable').each(function(){
        if($(this).data('id') === item.id){
          $(this).css('zIndex', n + 10);
        }
     });`

SegiNyn:
Hello @pferreir, please could you re-review the changes (I used the github re-review button at the top but the review status was not changing). Thank you

pferreir:
> Hello @pferreir, please could you re-review the changes (I used the github re-review button at the top but the review status was not changing). Thank you

Hi there. Yeah, I'll have a look as soon as I can.

SegiNyn:
Okay, thank you!

vasantvohra:
Hello,
Can we merge this PR if it looks good? it would help us deploy this in our next release set for January 12th?

ThiefMaster:
- If I edit an existing badge (with no background color information) both colorpickers show black (which isn't correct)
- How do I clear the background color (ie revert to transparent)?

SegiNyn:
Hi @ThiefMaster, this was because the HTML colour picker defaults to black whenever there is no colour selected, and it does not have a transparent option

ThiefMaster:
I think we need to find a solution for this... at least going back to transparent background is needed for sure. Not sure what the best solution is though...  I'll ping my colleagues in case anyone has a good idea.

(it's pretty ridiculous that the native color picker doesn't provide something for this!)

SegiNyn:
Okay, thank you. We also saw that Indico has a `ColorPickerWidget` and I'm testing to see if this would work inplace of the native color picker.

ThiefMaster:
That might indeed by worth looking into. I don't think it supports clearing either but adding it there is probably easier than adding a fully new widget wrapping the native one.

SegiNyn:
Hello @ThiefMaster, we have updated the code to use the ColorPickerWidget, which has the option to clear the color, but this changes the look because of the extra input field
Screenshot 2022-01-06 at 14 59 12
Screenshot 2022-01-06 at 14 58 44

ThiefMaster:
Fine for me, I'd just apply some CSS so the whole color code fits in the text field and the color box has the same size as the field (and is maybe attached right to its left)

SegiNyn:
It now looks like this:
Screenshot 2022-01-07 at 16 58 09

ThiefMaster:
Much nicer. Could you "equalize" the heights and remove the spacing between the color display and the text field?

SegiNyn:
Screenshot 2022-01-07 at 17 36 21

ThiefMaster:
almost :) there are still rounded top-left/bottom-left corners on the text fields, but since that's no longer an outer border these corners should be regular now

SegiNyn:
Screenshot 2022-01-07 at 18 05 29

CHANGES.rst
```diff
@@ -46,6 +46,7 @@ Improvements
   settings (:issue:`4837`, :issue:`2224`, :issue:`1877`, :pr:`5179`)
 - Allow filtering the contribution list based on whether any person (speaker or author)
   has registered for the event or not (:issue:`5192`, :pr:`5193`)
+- Add forward, backward and background color option to badge designer(:pr:`5139`, thanks :user:`seginyn`)
```
ThiefMaster, line 49:
```suggestion
- Add background color option and layer order to badge/poster designer items (:pr:`5139`,
  thanks :user:`SegiNyn`)
```

indico/modules/designer/client/js/index.js
```diff
@@ -240,11 +244,28 @@ import {$T} from 'indico/utils/i18n';
...
+    if (textColorPreview.hasClass('no-value') && item.color.length > 1) {
+      textColorPreview.removeClass('no-value');
+    }
+    textColorPreview.attr('style', 'background-color: ' + item.color + '; opacity: 1;');
+    if (item.color.length < 1) {
+      textColorPreview.addClass('no-value');
+    }
+    $('#bg-color-selector').val(item.background_color);
+    var bgColorPreview = $('#bg-color-preview');
+    if (bgColorPreview.hasClass('no-value') && item.background_color.length > 1) {
```
ThiefMaster, line 258: `item.background_color` is null for existing badges. Maybe just check `item.background_color` which will cover both empty strings and null/undefined?

Also, why `> 1` and not `> 0` in this check?

indico/modules/designer/client/js/index.js
```diff
@@ -240,11 +244,28 @@ import {$T} from 'indico/utils/i18n';
     var itemStyles = _.filter([item.bold ? 'bold' : null, item.italic ? 'italic' : null]);
 
     // Change the selectors so that they match the properties of the item
+    $('#text-color-selector').val(item.color);
+    var textColorPreview = $('#text-color-preview');
+    if (textColorPreview.hasClass('no-value') && item.color.length > 1) {
```
ThiefMaster, line 249: No need for that hasClass check; removing a class that isn't there is just a no-op. And I think you can greatly simplify it like this:

```
textColorPreview.toggleClass('no-value', !item.color);
```

indico/modules/designer/client/js/index.js
```diff
@@ -240,11 +244,28 @@ import {$T} from 'indico/utils/i18n';
...
+      textColorPreview.removeClass('no-value');
+    }
+    textColorPreview.attr('style', 'background-color: ' + item.color + '; opacity: 1;');
```
ThiefMaster, line 252: You can use `.css({...})` to set the style directly instead of using a css string

indico/modules/designer/client/js/index.js
```diff
@@ -240,11 +244,28 @@ import {$T} from 'indico/utils/i18n';
     var itemStyles = _.filter([item.bold ? 'bold' : null, item.italic ? 'italic' : null]);
 
     // Change the selectors so that they match the properties of the item
+    $('#text-color-selector').val(item.color);
+    var textColorPreview = $('#text-color-preview');
```
ThiefMaster, line 248: Maybe we could use `const`/`let` in new code, even if the rest of the file is very old and still uses `var`?

indico/modules/designer/templates/template.html
```diff
@@ -191,18 +191,24 @@
...
-                            <option value="pink">{% trans %}pink{% endtrans %}</option>
-                            <option value="gray">{% trans %}gray{% endtrans %}</option>
-                            <option value="white">{% trans %}white{% endtrans %}</option>
-                        </select>
+                        <div class="i-color-field clearfix">
+                            <div class="clickable-wrapper">
+                                <span id="text-color-preview" class="color-preview"></span>
+                                <input title="{% trans %}Text color{% endtrans %}" type="text"
+                                    id="text-color-selector" value="#" class="js-font-tool" data-attr="color"
+                                    maxlength="7" size="2">
```
ThiefMaster, line 199: alignment

indico/modules/designer/templates/template.html
```diff
@@ -288,6 +294,14 @@
                                 class="i-button move-button icon-only text-color subtle icon-expand"
                                 data-direction="bottom">
                         </button>
+                        <button title="{% trans %}Send item to the back{% endtrans %}"
```
ThiefMaster, line 297: Let's be consistent on the titles, e.g.:

- Move item to the back
- Move item to the front

ThiefMaster (approved):

PR #5151, opened 2021-11-10
Update python dependencies

ThiefMaster:


requirements.txt
```diff
@@ -228,7 +228,7 @@ pynpm==0.1.2
     # via
     #   flask-webpackext
     #   pywebpack
-pyparsing==3.0.1
+pyparsing==2.4.7
```
pferreir, line 231: Is this intentional? It should be on 3.0.5

ThiefMaster: Seems to have been intentional, but apparently got reverted upstream (w/o a new release yet) in the meantime: https://github.com/pypa/packaging/pull/482

ThiefMaster: I think it's safe to merge with the downgrade and then upgrade it again once packaging released a version w/o that pin..

pferreir: OK, fair enough

pferreir (approved):

PR #5154, opened 2021-11-12
Configure participant visibility when creating a new registration form

### LINKED ISSUE #5150: Privacy dashboard: configure participant visibility when creating a new form
Related to #5128 , #5127 

When creating a new registration form or enabling participants for a meeting, 
a dialog should be shown where one can configure the visibility of the participant list.

This is important since in order to respect consent,
the visibility cannot be changed once there are existing registrations.

At the same time, when creating a registration form for a conference, 
instead of showing the whole edit form, only basic information (including the visibility) should be asked.

tomasr8:
Closes #5150 

Shows a simplified form when creating a new regform consisting of the title (for conferences)
and the participant visibility settings. All other settings are still normally accessible through the regform settings later.

Meetings:
image

Conferences:
image

New `privacy` section in the regform main settings page:
image

### DIFF
--- a/indico/modules/events/registration/controllers/management/regforms.py
+++ b/indico/modules/events/registration/controllers/management/regforms.py
@@ -18,8 +18,8 @@ from indico.modules.events.payment import payment_settings
 from indico.modules.events.registration import logger, registration_settings
 from indico.modules.events.registration.controllers.management import RHManageRegFormBase, RHManageRegFormsBase
 from indico.modules.events.registration.forms import (ParticipantsDisplayForm, ParticipantsDisplayFormColumnsForm,
-                                                      RegistrationFormForm, RegistrationFormScheduleForm,
-                                                      RegistrationManagersForm)
+                                                      RegistrationFormCreateForm, RegistrationFormEditForm,
+                                                      RegistrationFormScheduleForm, RegistrationManagersForm)
 from indico.modules.events.registration.models.forms import RegistrationForm
 from indico.modules.events.registration.models.items import PersonalDataType
 from indico.modules.events.registration.models.registrations import PublishRegistrationsMode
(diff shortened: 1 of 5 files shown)


indico/modules/events/registration/controllers/management/regforms.py
```diff
@@ -127,28 +127,31 @@ def _process(self):
...
-        return redirect(url_for('event_registration.manage_regform', regform))
-
-    def _process_GET(self):
+    def _process(self):
         regform = self.event.participation_regform
         registration_enabled = self.event.has_feature('registration')
+        form = RegistrationFormCreateForm(event=self.event, title='Participants',
+                                          currency=payment_settings.get('currency'),
+                                          visibility=[PublishRegistrationsMode.show_with_consent.name,
+                                                      PublishRegistrationsMode.hide_all.name])
```
ThiefMaster, line 136: I think here we may want different behavior for meetings and other event types... for lectures/conferences I think the default is fine, but for meetings I'd assume a "show to everyone" default - or if we can't do that, force an active choice (because most meetings do NOT use a registration, just a participant list, so there's no real opportunity where someone would be able to give that consent)

indico/modules/events/registration/controllers/management/regforms.py
```diff
@@ -127,28 +127,31 @@ def _process(self):
...
+        if form.validate_on_submit():
+            set_feature_enabled(self.event, 'registration', True)
+            if not regform:
+                regform = RegistrationForm(event=self.event, is_participation=True)
+                create_personal_data_fields(regform)
+                form.populate_obj(regform)
```
ThiefMaster, line 142: maybe explicitly skip `visibility` here?

indico/modules/events/registration/controllers/management/regforms.py
tomasr8: That's a good point, I updated the defaults to be "show to everyone" for meetings and "hide to everyone for conferences"

indico/modules/events/registration/templates/management/regform_create.html
```diff
@@ -0,0 +1,18 @@
...
+{% block content %}
+    {{ form_header(form) }}
+    {{ form_rows(form, fields=form._conference_fields) }}
+    {% call form_footer(form) %}
+        <input class="i-button big highlight" type="submit" value="{% trans %}Create{% endtrans %}"
+               data-disabled-until-change>
+        <a href="{{ url_for('.manage_regform_list', event) }}" class="i-button big" data-button-back>
+            {% trans %}Back{% endtrans %}
+        </a>
+    {% endcall %}
```
ThiefMaster, line 17: This looks like something that's covered by the `simple_form` macro..

indico/modules/events/registration/templates/management/participants.html
```diff
@@ -14,6 +21,13 @@
             {% endif %}
         </div>
     </div>
+    {% if not regform %}
+        {{ form_header(form) }}
+        {{ form_rows(form, fields=form._meeting_fields) }}
+        {% call form_footer(form) %}
+            <input class="i-button big highlight" type="submit" value="{% trans %}Create{% endtrans %}">
+        {% endcall %}
```
ThiefMaster, line 29: This looks like a `simple_form` candidate as well

indico/modules/events/registration/templates/management/participants.html
```diff
@@ -1,9 +1,16 @@
...
 {% block title %}
     {%- trans %}Participants{% endtrans -%}
 {% endblock %}
 
+{% block subtitle %}
+    {% if not regform %}
+        {%- trans %}Create form{% endtrans -%}
+    {% endif %}
+{% endblock %}
+
```
ThiefMaster, line 13: Not sure if we need this

ThiefMaster (approved):

PR #5156, opened 2021-11-12
Add option to export role members to CSV

### LINKED ISSUE #5147: Export list of Roles members
In the "roles setup" there is an "import from CSV" function. 
This is fine to quickly populate a role with a list of people/accounts.
It would be nice to also have a "export to CSV" in the same place. 
This would enable to quickly export a list of users (emails) from an event and import them to another event without having to do this one by one.

tomasr8:
Closes #5147 

Not sure which button order would be the best..

image

### DIFF
--- a/indico/modules/events/roles/blueprint.py
+++ b/indico/modules/events/roles/blueprint.py
@@ -6,8 +6,8 @@
 # LICENSE file for more details.
 
 from indico.modules.events.roles.controllers import (RHAddEventRole, RHAddEventRoleMembers, RHDeleteEventRole,
-                                                     RHEditEventRole, RHEventRoleMembersImportCSV, RHEventRoles,
-                                                     RHRemoveEventRoleMember)
+                                                     RHEditEventRole, RHEventRoleMembersExportCSV,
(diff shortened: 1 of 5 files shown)


indico/modules/events/roles/blueprint.py
```diff
@@ -26,3 +26,5 @@
 
 _bp.add_url_rule('/<int:role_id>/members/import', 'add_members_import_csv', RHEventRoleMembersImportCSV,
                  methods=('GET', 'POST'))
+_bp.add_url_rule('/<int:role_id>/members.csv', 'members_export_csv', RHEventRoleMembersExportCSV,
+                 methods=('GET',))
```
ThiefMaster, line 30: that's the default, no need to specify it

ThiefMaster:
"members" vs "users" in export/import feels strange - let's stick with one term..

indico/modules/events/roles/blueprint.py
tomasr8: thanks, I had a hunch it would be the case ;)

ThiefMaster (approved):

PR #5161, opened 2021-11-15
Add new signal 'before_registration_email_send' to RHRegistrationEmai…

vtran99:
…lRegistrants.

Request new signal before sending mail to registrants in order to attach PDF to the mail generated.
The contents of PDF is taken from the form.

### DIFF
--- a/indico/modules/events/registration/controllers/management/reglists.py
+++ b/indico/modules/events/registration/controllers/management/reglists.py
@@ -221,6 +221,8 @@ class RHRegistrationEmailRegistrants(RHRegistrationsActionBase):
             email = make_email(to_list=registration.email, cc_list=form.cc_addresses.data, bcc_list=bcc,
                                from_address=form.from_address.data, template=template, html=True,
                                attachments=attachments)
+            signals.core.before_notification_send.send('registration-custom-email', email=email,
+                                                       registration=registration, form=form)
             send_email(email, self.event, 'Registration', log_metadata={'registration_id': registration.id})
 
     def _process(self):


indico/core/signals/event/registration.py
```diff
@@ -110,3 +110,10 @@
 object being cloned. The new ``RegistrationForm`` object is passed in the ``new_form``
 kwarg.
 ''')
+
+before_registration_email_send = _signals.signal('before_registration_email_send', '''
```
ThiefMaster, line 114: Since this signal is registration-specific, the context string isn't really useful and the signal definition should be in `indico/core/signals/event/registration.py`.

How about turning this into a `before_custom_registration_email` signal in there which uses the `registration` as the sender and the `email` kwarg as-is?

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -221,6 +221,8 @@ def _send_emails(self, form):
             email = make_email(to_list=registration.email, cc_list=form.cc_addresses.data, bcc_list=bcc,
                                from_address=form.from_address.data, template=template, html=True,
                                attachments=attachments)
+            signals.event.before_registration_email_send.send('email-registrants', email=email,
```
ThiefMaster, line 224: Alternatively, we could avoid adding a new signal at all and instead re-use `before_notification_send` here?

Like this:

```python
signals.core.before_notification_send.send('registration-custom-email', email=email,
                                           registration=registration, form=form)
```

(I'd prefer this over adding a new signal!)

ThiefMaster (changes requested):

indico/modules/events/registration/controllers/management/reglists.py
vtran99: Hi Adrian,
I don't care about the type of signal, just need to capture "a signal" to change the email's contents, so YES, the solution you propose suits me as well.
If ok for you, we can go with that:
raise "before_notification_send" in RHRegistrationEmailRegistrants before send_mail.

indico/core/signals/event/registration.py
vtran99: Sure, if we go with suggestion below, no need this anymore.

indico/modules/events/registration/controllers/management/reglists.py
ThiefMaster: Let's go for that then. Please amend your commit and force-push the updated version to your branch.

While you are at it, we usually try to use [this style](https://chris.beams.io/posts/git-commit/) for commit messages (in particular points 1-6), so if you could adapt your commit message accordingly that'd be great (otherwise I can do it when merging)

CHANGES.rst
```diff
@@ -52,7 +52,8 @@ Bugfixes
 Internal Changes
 ^^^^^^^^^^^^^^^^
 
-- Nothing so far
+- Send signal `before_notification_send` before custom email to registrants (:pr:`5161`, thanks
+   :user:`vtran`)
```
ThiefMaster, line 56:
```suggestion
- Send signal `before_notification_send` before custom email to registrants (:pr:`5161`,
  thanks :user:`vtran`)
```

ThiefMaster: Actually since it's not a new signal but just sending an existing one in another place I'm not sure if we really need a changelog entry for it...

indico/modules/events/registration/controllers/management/reglists.py
vtran99: Done, tried to follow indication for commit message, hope that'll be fine.

ThiefMaster: Yep, thanks!

CHANGES.rst
vtran99: Well, I was just following indication for PR upstream. Up to you to decide if you want to keep that mention or not.

ThiefMaster (approved):

Issue #5084, opened 2021-09-13
Expose attachments checksums in the export API


avivace:
The `/export/event/` endpoint seems to not mention any kind of checksum for the attached files.

As I understood from the [AbstractFile](https://docs.getindico.io/en/stable/api/abstract/?highlight=hash#indico.modules.events.abstracts.models.files.AbstractFile.md5) model, you do already compute and save MD5 checksums. If this is consistent through all the attached files in events, would it be possible to expose this information in the API?

avivace:
This would also make possible to find duplicates in attachments, since files with the same names are allowed.
E.g. on "ilcagenda", record 1049, has files with same name in the same directory/level, so there's no way from the API to understand if it's just a duplicate or they are actually duplicate entries (in that case they are actually slightly different files).

ThiefMaster:
Yes, we store an md5 hash for all attachments, so we could indeed include this in the APIs. Would you like to send a PR for this? I think including the `attachment.file.md5` in `_build_attachment_api_data` is all that's needed.

avivace:
> Yes, we store an md5 hash for all attachments, so we could indeed include this in the APIs. Would you like to send a PR for this? I think including the `attachment.file.md5` in `_build_attachment_api_data` is all that's needed.

Sure, I can take a look

PR #5167, opened 2021-11-17
Return attachments checksum on the export endpoint

avivace:
See https://github.com/indico/indico/issues/5084

Documentation still needs to be updated

### DIFF
--- a/indico/modules/attachments/api/util.py
+++ b/indico/modules/attachments/api/util.py
@@ -118,6 +118,8 @@ def _build_attachment_api_data(attachment):
         data['filename'] = attachment.file.filename
         data['content_type'] = attachment.file.content_type
         data['size'] = attachment.file.size
+        data['checksum'] = attachment.file.md5
+
 
     elif attachment.type == AttachmentType.link:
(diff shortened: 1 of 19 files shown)


indico/modules/attachments/api/util.py
```diff
@@ -111,7 +111,8 @@ def _build_attachment_api_data(attachment):
         'description': attachment.description,
         'modified_dt': attachment.modified_dt.isoformat(),
         'type': attachment.type.name,
-        'is_protected': attachment.is_protected
+        'is_protected': attachment.is_protected,
+        'checksum': attachment.file.md5,
```
ThiefMaster, line 115: This will break for link attachments. Add it in the if type == file block below.

Also, usually you should not use your own `master` branch for PRs - always create a different branch. We often rebase+merge or force-push small changes (e.g. changing the wording in changelog entries) and it's a bit of a mess.

In fact this is a PR I will most likely rebase to 3.1.x, which means I would need to force-push 3.1.x to your master, leaving your master in a weird state afterwards - with a throwaway feature branch that's no problem since you'd delete that after the PR is merged.

PR #5178, opened 2021-11-26
Implement rendering of invites on Outlook

pferreir:
While clients such as Thunderbird can understand the ICS files we are attaching to our invites and ask the user whether they'd like to add them to their calendars:

image

... this doesn't seem to be the case with Outlook, which requires some additional metadata to do so. This PR solves that by providing the `method=REQUEST` field which is missing for that to happen.

Example:

image
image

Also added some gratuitous type information, since you can always count on me to type-annotate Python code.

### DIFF
--- a/indico/modules/events/ical.py
+++ b/indico/modules/events/ical.py
@@ -5,6 +5,11 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
+import typing as t
+from email import message
+from email.mime.base import MIMEBase
+from email.policy import compat32
+
 import icalendar
 from lxml import html
 from lxml.etree import ParserError
@@ -13,9 +18,23 @@ from werkzeug.urls import url_parse
 from indico.core import signals
 from indico.core.config import config
 from indico.core.db.sqlalchemy.protection import ProtectionMode
+from indico.modules.events.contributions.models.contributions import Contribution
+from indico.modules.events.models.events import Event
+from indico.modules.events.sessions.models.sessions import Session
+from indico.modules.users.models.users import User
 from indico.util.date_time import now_utc
 from indico.util.enum import IndicoEnum
 from indico.util.signals import values_from_signal
+
+
+class MIMECalendar(MIMEBase):
+    """MIME `text/calendar` class which adds the `method=REQUEST` to the Content-Type."""
+
+    def __init__(self, filename: str, payload: str):
+        message.Message.__init__(self, policy=compat32)
+        self.add_header('Content-Type', 'text/calendar', filename=filename, encoding='utf-8', method='REQUEST')
+        self['MIME-Version'] = '1.0'
+        self.set_payload(payload)
 
 
 class CalendarScope(IndicoEnum):
@@ -23,7 +42,14 @@ class CalendarScope(IndicoEnum):
     session = 2
 
 
-def generate_basic_component(entity, uid=None, url=None, title=None, description=None):
+def generate_basic_component(
+    entity: t.Union[Event, Session, Contribution],
+    uid: t.Optional[str] = None,
+    url: t.Optional[str] = None,
+    title: t.Optional[str] = None,
+    description: t.Optional[str] = None,
+    organizer: t.Optional[t.Tuple[str, str]] = None
+):
     """Generate an iCalendar component with basic common properties.
 
     :param entity: Event/session/contribution where properties come from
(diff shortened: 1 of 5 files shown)


indico/modules/events/ical.py
```diff
@@ -13,17 +18,36 @@
...
+from indico.modules.events.models.events import Event
+from indico.modules.events.sessions.models.sessions import Session
+from indico.modules.users.models.users import User
 from indico.util.date_time import now_utc
 from indico.util.enum import IndicoEnum
 from indico.util.signals import values_from_signal
 
 
+class MIMECalendar(MIMEBase):
+    """MIME `text/calendar` Class which adds the `method=REQUEST` to the Content-Type."""
```
ThiefMaster, line 31:
```suggestion
    """MIME ``text/calendar`` class which adds the ``method=REQUEST`` to the Content-Type."""
```

indico/modules/events/ical.py
```diff
@@ -110,24 +134,38 @@ def generate_event_component(event, user=None):
...
     :param scope: If specified, use a more detailed timetable using the given scope
     :param skip_access_check: Do not perform access checks. Defaults to False.
+    :param method: METHOD field of the iCalendar object
     """
-    return events_to_ical([event], user, scope, skip_access_check=skip_access_check)
+    return events_to_ical([event], user, scope, skip_access_check=skip_access_check, method=method)
 
 
-def events_to_ical(events, user=None, scope=None, *, skip_access_check=False):
+def events_to_ical(
```
ThiefMaster, line 155: I know it's a matter of preference, but what about using this indentation style for multiline function definitions?

```python
def events_to_ical(
    events: list[Event],
    user: t.Optional[User] = None,
    scope: t.Optional[str] = None,
    *,
    skip_access_check: bool = False,
    method: t.Optional[str] = None
):
    ...
```

I think that's actually the one we use most commonly in the codebase

indico/modules/events/reminders/models/reminders.py
```diff
@@ -210,8 +210,8 @@ def send(self):
         email_tpl = make_reminder_email(self.event, self.include_summary, self.include_description, self.message)
         attachments = []
         if self.attach_ical:
-            event_ical = event_to_ical(self.event, skip_access_check=True)
-            attachments.append(('event.ics', event_ical, 'text/calendar'))
+            event_ical = event_to_ical(self.event, skip_access_check=True, method='REQUEST')
+            attachments.append(MIMECalendar('event.ical', event_ical))
```
ThiefMaster, line 214: `.ical`? don't almost all ical attachments use `.ics`? (e.g. `vcalendar.ics` from outlook)

ThiefMaster:
I think this would be something nice for 3.1.x :) (and deserves a changelog entry IMHO)

indico/modules/events/ical.py
pferreir: Indeed.

indico/modules/events/reminders/models/reminders.py
pferreir: Yeah, it's a typo.

indico/modules/events/ical.py
```diff
@@ -13,17 +18,37 @@
...
     session = 2
 
 
-def generate_basic_component(entity, uid=None, url=None, title=None, description=None):
+def generate_basic_component(
+        entity: t.Union[Event, Session, Contribution],
+        uid: t.Optional[str] = None,
+        url: t.Optional[str] = None,
+        title: t.Optional[str] = None,
+        description: t.Optional[str] = None
```
ThiefMaster, line 50:
```suggestion
    entity: t.Union[Event, Session, Contribution],
    uid: t.Optional[str] = None,
    url: t.Optional[str] = None,
    title: t.Optional[str] = None,
    description: t.Optional[str] = None
````

File: indico/modules/events/ical.py [hunk 5]
```diff
@@ -110,24 +135,40 @@ def generate_event_component(event, user=None):
...
 
 
-def event_to_ical(event, user=None, scope=None, *, skip_access_check=False):
+def event_to_ical(
+        event: Event,
+        user: t.Optional[User] = None,
+        scope: t.Optional[str] = None,
+        *,
+        skip_access_check: bool = False,
+        method: t.Optional[str] = None
```
ThiefMaster, line 144:
```suggestion
    event: Event,
    user: t.Optional[User] = None,
    scope: t.Optional[str] = None,
    *,
    skip_access_check: bool = False,
    method: t.Optional[str] = None
```

ThiefMaster (approved):
Let's rebase to 3.1.x though

pferreir:
Rebased.

PR #5180, opened 2021-11-29
Decouple submitter permissions from person_link_data

plourenco:
Previously, a new set of person_links could only be added by providing a list composed of {PersonLink: Bool} where the boolean represented if that link should have submitter permissions. As our uses broaden, it doesn't make sense to update permissions seamlessly while adding new links.

These changes update the `person_link_data` object form to a list of person_links.
In addition, introduces a method called `update_submitter_permissions` to restore the previous behaviour.


plourenco:
I've decided to close this PR because updating the permissions becomes more inherently complex when the contribution/event is still being created, and it would most likely require updating them manually when saving each form. Since this is an action we perform in multiple forms and we may even do it in a declarative way `person_link_data = new_person_link_data`, permissions should follow to avoid neglecting them.

PR #5191, opened 2021-12-07
Add participant visibility settings

duartegalvao:
Depends on #5128

This PR includes:

- The ability for participants to consent to or revoke consent to being shown on events' participants lists.
- A new "show with consent" option for registration form's participant lists. Link to this setting on the privacy dashboard
- Ability to _unpublish_ i.e. modify visibility of registered users. This can be done in the regform list view.

Current progress:

- [x] Show with consent option on registration forms settings
- [x] Filter displayed participants based on consent options
- [x] Add consent level selector on registration forms

Privacy dashboard includes links to set participant visibility for registration forms:
image

Setting participant visibility
image

Consenting as a participant:
image

### DIFF
--- a/indico/modules/events/registration/client/js/components/ConsentToPublishEditor.jsx
+++ b/indico/modules/events/registration/client/js/components/ConsentToPublishEditor.jsx
@@ -1,0 +1,63 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2022 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import registrationChangeConsent from 'indico-url:event_registration.api_registration_change_consent';
+
+import PropTypes from 'prop-types';
+import React, {useState} from 'react';
+import {Icon} from 'semantic-ui-react';
+
+import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+
+import ConsentToPublishDropdown, {publishModePropType} from './ConsentToPublishDropdown';
+
+import './ConsentToPublishEditor.module.scss';
+
+export default function ConsentToPublishEditor({
+  locator,
+  publishToParticipants,
+  publishToPublic,
+  initialConsentToPublish,
+}) {
+  const [consentToPublish, setConsentToPublish] = useState(initialConsentToPublish);
+  const [status, setStatus] = useState('');
+
+  async function editConsentToPublish(value) {
+    setConsentToPublish(value);
+    setStatus('loading');
+    try {
+      await indicoAxios.post(registrationChangeConsent(locator), {consent_to_publish: value});
+      setStatus('success');
+    } catch (error) {
+      setStatus('error');
+      setConsentToPublish(consentToPublish);
+      return handleAxiosError(error);
+    }
+  }
+
+  return (
+    <div styleName="consent-dropdown-container">
+      <ConsentToPublishDropdown
+        publishToParticipants={publishToParticipants}
+        publishToPublic={publishToPublic}
+        onChange={(evt, {value}) => editConsentToPublish(value)}
+        value={consentToPublish}
+        loading={status === 'loading'}
+        disabled={status === 'loading'}
+      />
+      {status === 'success' && <Icon name="check" color="green" />}
+      {status === 'error' && <Icon name="x" color="red" />}
+    </div>
+  );
+}
+
+ConsentToPublishEditor.propTypes = {
+  locator: PropTypes.object.isRequired,
+  publishToParticipants: publishModePropType.isRequired,
+  publishToPublic: publishModePropType.isRequired,
+  initialConsentToPublish: PropTypes.oneOf(['nobody', 'participants', 'all']).isRequired,
+};
--- a/indico/modules/events/registration/forms.py
+++ b/indico/modules/events/registration/forms.py
@@ -70,8 +70,8 @@ class RegistrationFormForm(IndicoForm):
     modification_mode = IndicoEnumSelectField(_('Modification allowed'), enum=ModificationMode,
                                               description=_('Will users be able to modify their data? When?'))
     publish_registrations_mode = IndicoEnumSelectField(_('Publish registrations'), enum=PublishRegistrationsMode,
-                                                       description=_('Registrations from this form will be displayed '
-                                                                     'in the event page'))
+                                                       description=_('Specify under which conditions registrations '
+                                                                     'will be displayed publicly'))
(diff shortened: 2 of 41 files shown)


indico/modules/events/registration/client/js/components/ConsentToPublishEditor.jsx
```diff
@@ -0,0 +1,69 @@
...
+      {failure && <Icon name="x" color="red" />}
+    </div>
+  );
+}
+
+ConsentToPublishEditor.propTypes = {
+  locator: PropTypes.object.isRequired,
+  publishToParticipants: publishModePropType.isRequired,
+  publishToPublic: publishModePropType.isRequired,
+  initialConsentToPublish: PropTypes.oneOf(['not_given', 'participants', 'all']).isRequired,
```
ThiefMaster, line 68: I'd use 'nobody' instead of 'not_given' here. I think that fits nicely with the other ones - "show to nobody / participants / all"

indico/modules/events/registration/client/js/components/ConsentToPublishEditor.jsx
```diff
@@ -0,0 +1,69 @@
...
+export default function ConsentToPublishEditor({
+  locator,
+  publishToParticipants,
+  publishToPublic,
+  initialConsentToPublish,
+}) {
+  const [consentToPublish, setConsentToPublish] = useState(initialConsentToPublish);
+  const [loading, setLoading] = useState(false);
+  const [success, setSuccess] = useState(false);
+  const [failure, setFailure] = useState(false);
```
ThiefMaster, line 29: this looks like a good usecase for `useDispatch` since you always change multiple fields at once depending on some action. or a much more simple alternative: a single useState for a 'status', ie you'd then do `setStatus(...)` with 'loading', 'success' or 'error'

indico/modules/events/registration/client/js/components/ConsentToPublishEditor.jsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  return (
+    <div styleName="consent-dropdown-container">
+      <ConsentToPublishDropdown
+        publishToParticipants={publishToParticipants}
+        publishToPublic={publishToPublic}
+        onChange={(evt, {value}) => editConsentToPublish(value)}
+        value={consentToPublish}
+        loading={loading}
```
ThiefMaster, line 56: you may want to disable it as well while loading

indico/modules/events/registration/forms.py
```diff
@@ -524,5 +508,22 @@ class RegistrationPrivacyForm(IndicoForm):
...
+        participant_visibility, public_visibility = (PublishRegistrationsMode[v] for v in field.data)
+        if participant_visibility.value < public_visibility.value:
+            raise ValidationError(_('Registrations list visibility can not be more restrictive to participants than '
+                                    'to the public'))
+        if (self.regform and
+                self.regform.existing_registrations_count > 0 and
+                ((participant_visibility == PublishRegistrationsMode.show_all and
+                  self.regform.publish_registrations_participants != PublishRegistrationsMode.show_all) or
+                 (public_visibility == PublishRegistrationsMode.show_all and
+                  self.regform.publish_registrations_public != PublishRegistrationsMode.show_all))):
```
ThiefMaster, line 528: I don't think this block is particularly readable...

indico/modules/events/registration/client/js/components/ConsentToPublishEditor.jsx
duartegalvao: Thanks! I wanted to use an enum but I didn't find a good way to do it in JS - the strings for the states seem to be the best option

indico/modules/events/registration/forms.py
duartegalvao: I agree... Would this be more readable?
```python
participant_visibility_changed_to_show_all = (participant_visibility == PublishRegistrationsMode.show_all and
                                                self.regform.publish_registrations_participants !=
                                                PublishRegistrationsMode.show_all)
public_visibility_changed_to_show_all = (public_visibility == PublishRegistrationsMode.show_all and
                                            self.regform.publish_registrations_public !=
                                            PublishRegistrationsMode.show_all)
if (self.regform and
        self.regform.existing_registrations_count > 0 and
        (participant_visibility_changed_to_show_all or public_visibility_changed_to_show_all)):
```
Or this?

```python
old_participant_visibility = self.regform.publish_registrations_participants
old_public_visibility = self.regform.publish_registrations_public
if (self.regform and
        self.regform.existing_registrations_count > 0 and
        ((participant_visibility == PublishRegistrationsMode.show_all and
            old_participant_visibility != PublishRegistrationsMode.show_all) or
            (public_visibility == PublishRegistrationsMode.show_all and
            old_public_visibility != PublishRegistrationsMode.show_all))):
```

ThiefMaster: the second one is just as bad as the original IMHO, but how about this (based on your first one):

```python
participant_visibility_changed_to_show_all = (
    participant_visibility == PublishRegistrationsMode.show_all and
    self.regform.publish_registrations_participants != PublishRegistrationsMode.show_all
)
public_visibility_changed_to_show_all = (
    public_visibility == PublishRegistrationsMode.show_all and
    self.regform.publish_registrations_public != PublishRegistrationsMode.show_all
)

if (
    self.regform and
    self.regform.existing_registrations_count > 0 and
    (participant_visibility_changed_to_show_all or public_visibility_changed_to_show_all)
):
    ...
```

That way you avoid the ugly wrapping within a binary expression

indico/modules/events/registration/forms.py
```diff
@@ -473,3 +477,60 @@ def validate_remove(self, field):
 
     def is_submitted(self):
         return super().is_submitted() and 'submitted' in request.form
+
+
+class ChangeRegistrationVisibilityForm(IndicoForm):
```
ThiefMaster, line 482: I haven't tested it on my instance yet, but this seems like a form to bulk-manage the consent? I'm not sure whether a bulk change for this is a good idea, especially if the event manager has no way to revert hiding someone, and in particular the manager should not be able to give consent on behalf of the user..

duartegalvao: Hmm I see the problem. This form allows downgrading consent only, and was meant to fulfil one of the original requirements:
> - option for organizer to unpublish individual participant(s) regardless of consent (ie another column for this)
Now I see that this should be an additional attribute of the registration.. I will work on that

ThiefMaster: Yes exactly. But in general I think it may be enough to make this an action when viewing a specific registration. I don't see that much value in doing this in bulk (but no harm in it either)

ThiefMaster (approved):

PR #5193, opened 2021-12-07
Add a contribution filter for registered users on any role

### LINKED ISSUE #5192: Add a contribution filter for registered users on any role
We currently provide a filter on contributions for registered/unregistered speakers. It would be very useful to do the same for a user in any role (such as, for contacting contributions without any registered authors).

plourenco:
Closes #5192

### DIFF
--- a/indico/modules/events/contributions/lists.py
+++ b/indico/modules/events/contributions/lists.py
@@ -82,7 +82,7 @@ class ContributionListGenerator(ListGeneratorBase):
             if status_criteria:
                 criteria.append(db.or_(*status_criteria))
 
-        if filtered_people := filters['items'].get('people') or filters['items'].get('speakers'):
+        if 'people' in filters['items'] or 'speakers' in filters['items']:
             registration_join_criteria = [
                 Registration.event_id == Contribution.event_id,
                 Registration.is_active,
@@ -98,6 +98,7 @@ class ContributionListGenerator(ListGeneratorBase):
             registered_contribs = {id_ for id_, in contrib_query}
 
             people_criteria = []
+            filtered_people = filters['items'].get('people') or filters['items'].get('speakers')
             if 'registered' in filtered_people:
                 people_criteria.append(Contribution.id.in_(registered_contribs))
             if 'not_registered' in filtered_people:
--- a/indico/modules/events/contributions/lists.py
+++ b/indico/modules/events/contributions/lists.py
@@ -68,6 +68,20 @@ class ContributionListGenerator(ListGeneratorBase):
                          db.undefer('attachment_count'),
                          db.undefer('is_scheduled')))
 
+    def _build_registration_query(self, is_speaker=False):
+        registration_join_criteria = [
+            Registration.event_id == Contribution.event_id,
+            Registration.is_active,
+            (Registration.user_id == EventPerson.user_id) | (Registration.email == EventPerson.email)
+        ]
+        if is_speaker:
+            registration_join_criteria.append(ContributionPersonLink.is_speaker)
+        return (db.session.query(Contribution.id)
+                .filter(Contribution.event_id == self.event.id)
+                .join(ContributionPersonLink)
+                .join(EventPerson)
+                .join(Registration, db.and_(*registration_join_criteria)))
+
     def _filter_list_entries(self, query, filters):
         if not filters.get('items'):
             return query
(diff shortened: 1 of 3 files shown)


indico/modules/events/contributions/lists.py
```diff
@@ -80,28 +82,28 @@ def _filter_list_entries(self, query, filters):
             if status_criteria:
                 criteria.append(db.or_(*status_criteria))
 
-        if 'speakers' in filters['items']:
-            registration_join_criteria = db.and_(
+        if filtered_people := filters['items'].get('people') or filters['items'].get('speakers'):
```
ThiefMaster, line 85: I really don't like the idea of using the walrus operator in this way (using the assigned value somewhere much later, outside the body of that particular ``if` statement

plourenco: That was quick :D You have a point, I was trying to get rid of some lines, let me know if the new changes are simpler.

indico/modules/events/contributions/lists.py
```diff
@@ -80,28 +82,29 @@ def _filter_list_entries(self, query, filters):
...
-            filtered_speakers = filters['items']['speakers']
-            speakers_criteria = []
-            if 'registered' in filtered_speakers:
-                speakers_criteria.append(Contribution.id.in_(registered_contribs))
-            if 'not_registered' in filtered_speakers:
-                speakers_criteria.append(~Contribution.id.in_(registered_contribs))
-            if speakers_criteria:
-                criteria.append(db.or_(*speakers_criteria))
+            people_criteria = []
+            filtered_people = filters['items'].get('people') or filters['items'].get('speakers')
```
ThiefMaster, line 101: I don't think this is correct.

For example, take "Speakers: Registered" and "People: Not registered". This will end up showing contributions with unregistered speakers.

You either need to split the code here to be separate for the two filters (ie duplicate the `registered_contribs` logic with the different join criteria), or make the speakers and people filters mutually exclusive. Splitting is probably cleaner since we currently do not have any filters that are exclusive with each other either.

ThiefMaster: For testing this I recommend using an event with contributions like this that cover the combinations:

image

ThiefMaster (changes requested):

ThiefMaster:
This can probably go into 3.1.x

indico/modules/events/contributions/lists.py
plourenco: While showing unregistered speakers is indeed not correct, I'm not sure it will ever make sense to use these two filters together. In the case where we select "People: Not registered" it will mean selecting any contribution where no one is registered:

imagem

This might seem ambiguous at first, some people might even assume it means selecting any contribution where there is at least someone unregistered:

imagem

In the example above, using "Speakers: Registered" with "People: Not registered" would certainly mean no results as if we are selecting only contributions where no one is registered, we cannot expect a speaker to be registered.

indico/modules/events/contributions/lists.py
```diff
@@ -66,6 +68,20 @@ def _build_query(self):
...
+            Registration.is_active,
+            (Registration.user_id == EventPerson.user_id) | (Registration.email == EventPerson.email)
+        ]
+        if is_speaker:
+            registration_join_criteria.append(ContributionPersonLink.is_speaker)
+        return (db.session.query(Contribution.id)
+                .filter(Contribution.event_id == self.event.id)
+                .join(ContributionPersonLink)
+                .join(EventPerson)
+                .join(Registration, db.and_(*registration_join_criteria)))
```
ThiefMaster, line 83: I'd keep the `{id_ for id_, in query}` logic for the return value: there's much less obscurity if we return just plain IDs instead of a query object (I couldn't tell you w/o looking at the query log whether SA turns it into a subquery somehow or executes the query when you pass it to .`in_()`...)

indico/modules/events/contributions/lists.py
ThiefMaster: ~~I agree using them together may not make the most sense.~~

~~However, "Speakers: Registered" and "People: Not registered" would give you contributions where some speakers have registered and some people have not registered...~~

You are right, using both filters at the asme time usually makes no sense, because:

- "Registered" means *someone* from the list of persons is registered
- "Not registered" *nobody* from the list of persons is registered

FWIW I first thought "not registered" means that *someone* from the list is not registered... honestly I'm not sure what behavior makes more sense, but "nobody" registered is probably more useful for organizers since what they typically care about is that there's someone to give the presentation / explain the poster...

indico/modules/events/contributions/lists.py
ThiefMaster: OK I just checked the log and it generates subqueries within the `IN (...)`. Since it works fine and doesn't cause any problems w/ duplicate references (since there's no correlation to the outer query), not any performance problems, I think we can keep it as it is

ThiefMaster (approved):

PR #5195, opened 2021-12-08
Allow custom message on Exclusive validator

duartegalvao:
This PR adds the ability to set a custom error message on the `Exclusive` WTForms validator.

### DIFF
--- a/indico/web/forms/validators.py
+++ b/indico/web/forms/validators.py
@@ -80,13 +80,16 @@ class Exclusive:
     If any of the given fields have a value, the validated field may not have one.
     """
 
-    def __init__(self, *fields):
+    def __init__(self, *fields, message=None):
         self.fields = fields
+        self.message = message
 
     def __call__(self, form, field):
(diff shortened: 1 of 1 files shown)


indico/web/forms/validators.py
```diff
@@ -80,18 +80,22 @@ class Exclusive:
...
     def __call__(self, form, field):
         if field.data is None:
             return
         if any(form[f].data is not None for f in self.fields):
-            field_names = sorted(str(form[f].label.text) for f in self.fields)
-            msg = ngettext('This field is mutually exclusive with another field: {}',
-                           'This field is mutually exclusive with other fields: {}',
-                           len(field_names))
-            raise ValidationError(msg.format(', '.join(field_names)))
+            if self.message is None:
```
ThiefMaster, line 91: cleaner to do it like this:

```
if self.message:
    raise ValidationError(self.message)
```

that way you don't need the extra indentation level for the rest of the code

PR #5196, opened 2021-12-08
Add support for multiple privacy notice URLs

tomasr8:
Depends on #5128

Adds support for multiple URLs:

image

### DIFF
--- a/indico/modules/events/management/forms.py
+++ b/indico/modules/events/management/forms.py
@@ -43,8 +43,9 @@ from indico.web.forms.fields import (IndicoDateField, IndicoDateTimeField, Indic
                                      IndicoPasswordField, IndicoProtectionField, IndicoRadioField,
                                      IndicoSelectMultipleCheckboxField, IndicoTagListField, IndicoTimezoneSelectField,
                                      IndicoWeekDayRepetitionField, MultiStringField, RelativeDeltaField)
+from indico.web.forms.fields.itemlists import MultipleItemsField
 from indico.web.forms.fields.principals import PermissionsField
-from indico.web.forms.validators import Exclusive, HiddenUnless, LinkedDateTime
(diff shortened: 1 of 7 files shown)


indico/modules/events/templates/footer.html
```diff
@@ -1,11 +1,12 @@
 {% extends 'footer.html' %}
 
 {% block footer_links_extra %}
-    {% if privacy_url %}
+    {% for privacy_url in privacy_urls  %}
```
ThiefMaster, line 4: Maybe we could show a single link here and show a menu/dropdown-like widget if there are multiple choices?

indico/modules/events/management/forms.py
```diff
@@ -232,16 +233,22 @@ def __init__(self, *args, **kwargs):
...
     _data_controller_fields = ('data_controller_name', 'data_controller_email')
-    _privacy_policy_fields = ('privacy_policy_url', 'privacy_policy')
+    _privacy_policy_fields = ('privacy_policy_urls', 'privacy_policy')
     data_controller_name = StringField(_('Person/Institution'))
     data_controller_email = EmailField(_('Contact email'), [Optional(), Email()])
-    privacy_policy_url = URLField(_('URL'), [Optional(), URL()],
-                                  description=_('The URL to an external page with the privacy notice'))
+    privacy_policy_urls = MultipleItemsField(_('URLs'),
+                                             fields=[{'id': 'title', 'caption': _('Title'),
+                                                      'required': True, 'type': 'string'},
```
ThiefMaster, line 241: It'd be nice if the title was only required in case of multiple privacy notices (especially since we also use i18n as soon as we use a custom title)

ThiefMaster:
Could we style the dropdown trigger link the usual links (and maybe open on click instead of on hover, and ditch the arrow there?). Right now it looks a bit strange and it's not clear initially why this "link" doesn't look like the others...

PR #5201, opened 2021-12-15
Import Selectize.js globally

vasantvohra:
Importing selectize widget here to be used globally by plugin and other files.
Let me know If I have to remove any legacy local imports.

### DIFF
--- a/indico/web/client/js/jquery/index.js
+++ b/indico/web/client/js/jquery/index.js
@@ -31,6 +31,10 @@ import 'qtip2/dist/jquery.qtip.css';
 import 'vanderlee-colorpicker';
 import 'vanderlee-colorpicker/jquery.colorpicker.css';
 
+import 'selectize';
+import 'selectize/dist/css/selectize.css';
+import 'selectize/dist/css/selectize.default.css';
+
 import './widgets';
 
 // i18n (Jed)
--- a/indico/modules/events/abstracts/client/js/index.js
+++ b/indico/modules/events/abstracts/client/js/index.js
@@ -9,10 +9,6 @@
 
 import _ from 'lodash';
 
-import 'selectize';
-import 'selectize/dist/css/selectize.css';
-import 'selectize/dist/css/selectize.default.css';
-
 import {$T} from 'indico/utils/i18n';
 
 import 'indico/modules/events/reviews';
--- a/indico/modules/events/management/client/js/index.js
+++ b/indico/modules/events/management/client/js/index.js
@@ -12,10 +12,6 @@ import _ from 'lodash';
 import React from 'react';
 import ReactDOM from 'react-dom';
 
-import 'selectize';
-import 'selectize/dist/css/selectize.css';
-import 'selectize/dist/css/selectize.default.css';
-
 import 'indico/modules/events/util/list_generator';
 import 'indico/modules/events/util/static_filters';
 


ThiefMaster:
If you import it globally, remove the imports from the other JS files where they are probably no longer needed

vasantvohra:
found at two places removed from both of them.

indico/web/client/js/jquery/index.js
```diff
@@ -52,6 +52,10 @@ import './utils/forms';
 import './utils/misc';
 import './utils/routing';
 import './utils/sortablelist';
+import 'selectize';
+import 'selectize/dist/css/selectize.css';
+import 'selectize/dist/css/selectize.default.css';
```
ThiefMaster, line 57: I'd include this earlier e.g. before the `./widgets` import since that's where we also import all the other jquery stuff

vasantvohra:
Hello @ThiefMaster,
Sorry to disturb during holiday season, 

## Experimentation

I experimented with the Selectize Widget. And, I can see, this will be extensively used in our plugin, as we currently have large lists for NGOs, and we will have large lists for cities (about 88K), and we will be linking multiple locations to one city, and cities to a country via the Admin UI. 
We might allow participants to choose a city as well from the form, or during setting an event's address.

Currently I did few tests on our PMS interface,

Screenshot 2021-12-20 at 10 32 03 PM


Screenshot 2021-12-20 at 10 32 27 PM
Screenshot 2021-12-20 at 10 32 52 PM

Screenshot 2021-12-20 at 10 33 19 PM

Now considering the following screenshot and the case when values are preselected and form is being edited. And, It would not be a good idea to have preload enabled on huge lists such as events, categories, and cities. Now, the lines mentioned below reselects the previously selected value as the new options are being loaded using Ajax call.

Screenshot 2021-12-20 at 10 33 50 PM

My suggestion would be to move the the following lines to onInitialize event.
https://github.com/indico/indico/blob/87368771b4c1946feb96d60c4881971f4b7922ed/indico/web/client/js/jquery/widgets/jinja/selectize_widget.js#L67-L70

From one of the suggestions from the [stack overflow](https://stackoverflow.com/a/51546207), and looking at [docs](https://selectize.dev/docs.html)

Property | Description
-- | --
options | An object containing the entire pool of options. The object is keyed by each object's value.
items | An array of selected values.

https://github.com/indico/indico/blob/6046c4d5d98da55ccb47e81e343550e1217bd963/indico/web/forms/widgets.py#L202

I think the choices (including the pre-populated choice) could be passed to options key and items key can contain only [id],

```python
        if field.data is not None:
            choices = self._get_choices(field.data, search_field)
            value = [str(field.data.id)]
        choices, value = kwargs.pop('choices', choices), kwargs.pop('value', value)
        search_fields = [search_field]
        if self.allow_by_id:
            search_fields.append('id')
        options = {
            'options': choices,
            'valueField': self.value_field,
            'labelField': self.label_field,
            'searchField': search_fields,
            'persist': False,
            'items': value,
            'create': False,
            'maxItems': max_items,
            'closeAfterSelect': False,
            'preload': self.preload,
        }
 ```
 
 ## Option to select Multiple Items,
 there could be multiple selections on Selective widget, but set value on JS only considers the one value, if not moved to backend I'd suggest following, 
 
 ```js
 onInitialize: function() {
        if (options.selectedItem !== null) {
          if (Array.isArray(options.selectedItem)){
            var selected_items = []
            options.selectedItem.forEach(function (obj) {
              selected_items.push(obj.id);
            })
            $field[0].selectize.setValue(selected_items)
          }
          else {
            $field[0].selectize.setValue(options.selectedItem.id);
          }
        }
      },
```

ThiefMaster:
I suggest opening a separate issue/PR for what you wrote in your last comment... Making the import global works fine, so I'll merge this.

PR #5203, opened 2021-12-17
Fix category search displaying deleted categories

### LINKED ISSUE #5197: category search results list contains categories that have been previously deleted
**Bug Description**
when I search for category (i.e. I'm moving an event to another category) and I type a category name, I get in the result set all the occurrences of that category even if it has been removed (once or many times).
As shown here:
category_search

**To Reproduce**
1. Go to a conference page
2. Click on "Event action" and then "Move"
3. "Select a category" will appear
4. Type a category that has been deleted from the system, once or more times
5. In the results list you will get a row the category for every time it has been created (and deleted)  

**Expected behavior**
Category search results should not include categories that have been removed by the system

duartegalvao:
Closes #5197 

This PR includes the fix for the described issue, as well as a unit test for category searches.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -55,6 +55,7 @@ Bugfixes
 
 - Fix :data:`CUSTOM_COUNTRIES` not overriding names of existing countries (:pr:`5183`)
 - Fix error dialog when submitting an invited abstract without being logger in (:pr:`5200`)
+- Fix category picker search displaying deleted categories (:issue:`5197`, :pr:`5203`)
 
 Internal Changes
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Since this is fixing a bug a changelog entry would be nice. "Do not show deleted categories in category picker search" or something like that.

CHANGES.rst
```diff
@@ -55,6 +55,7 @@ Bugfixes
 
 - Fix :data:`CUSTOM_COUNTRIES` not overriding names of existing countries (:pr:`5183`)
 - Fix error dialog when submitting an invited abstract without being logger in (:pr:`5200`)
+- Fix category picker search displaying deleted categories (:pr:`5203`)
```
ThiefMaster, line 58: since there's an issue you can mention hat one as well via `:issue:`

PR #5207, opened 2022-01-06
Add a new React widget for privacy notice URLs

### LINKED ISSUE #5206: Privacy dashboard: new React widget for multiple URLs
Separate issue based on https://github.com/indico/indico/pull/5196#discussion_r765754870

Replace `MultipleItemsField` in privacy dashboard with a custom React widget.
Contrary to the current behavior for a single notice, the title should not be required.
For subsequent notices, both title and the URL should be required.

duartegalvao:
Closes #5206
Depends on #5196

This PR replace the MultipleItemsField in the privacy dashboard with a custom React widget.
Privacy notice titles now are only required when more than one privacy notice URL is provided.

image


ThiefMaster:
5196 has been merged now so this PR can be rebased

indico/web/client/js/react/components/WTFPrivacyPolicyURLsField.jsx
```diff
@@ -0,0 +1,108 @@
...
+// LICENSE file for more details.
+
+import PropTypes from 'prop-types';
+import React, {useState} from 'react';
+
+import {Translate} from 'indico/react/i18n';
+
+export default function WTFPrivacyPolicyURLsField({fieldId, fields, initialValues}) {
+  const [values, setValues] = useState([]);
+  const tempValues = initialValues;
```
ThiefMaster, line 15: This looks strange...

- you'll mutate initialValues when mutating tempValues i addURL
- you change local state without re-rendering

duartegalvao: Yeah, I see the problem, I'm still figuring this out. Thank's for the tips!

indico/web/forms/fields/privacy.py
```diff
@@ -0,0 +1,24 @@
...
+class IndicoPrivacyPolicyURLsField(JSONField):
+    widget = JinjaWidget('forms/privacy_policy_urls_widget.html', single_kwargs=True, single_line=True)
+
+    def pre_validate(self, form):
+        if len(self.data) > 1:
+            for item in self.data:
+                if not item.get('title', ''):
+                    raise ValidationError(_('Titles are required when more than one privacy notice is specified'))
+                if not item.get('url', ''):
+                    raise ValidationError(_('URL is required'))
```
ThiefMaster, line 24: This is also needed if there's just a single item

indico/web/forms/fields/privacy.py
```diff
@@ -0,0 +1,24 @@
...
+from indico.web.forms.widgets import JinjaWidget
+
+
+class IndicoPrivacyPolicyURLsField(JSONField):
+    widget = JinjaWidget('forms/privacy_policy_urls_widget.html', single_kwargs=True, single_line=True)
+
+    def pre_validate(self, form):
+        if len(self.data) > 1:
+            for item in self.data:
+                if not item.get('title', ''):
```
ThiefMaster, line 21: you can omit the second arg since `None` is falsy as well

indico/web/forms/fields/privacy.py
```diff
@@ -0,0 +1,23 @@
...
+from indico.util.i18n import _
+from indico.web.forms.fields.simple import JSONField
+from indico.web.forms.widgets import JinjaWidget
+
+
+class IndicoPrivacyPolicyURLsField(JSONField):
+    widget = JinjaWidget('forms/privacy_policy_urls_widget.html', single_kwargs=True, single_line=True)
+
+    def pre_validate(self, form):
+        for item in self.data:
```
ThiefMaster, line 19: Mostly a matter of preference, but I think this is a bit nicer since it only has to get the len() once (there's no meaningful difference as len is O(1) in Python anyway)

```python
if not all(x.get('url') for x in self.data):
    raise ..
if len(self.data) > 1 and not all(x.get('title') for x in self.data):
    raise ..
```

indico/web/client/js/react/components/WTFPrivacyPolicyURLsField.jsx
```diff
@@ -0,0 +1,136 @@
...
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import PropTypes from 'prop-types';
+import React, {useMemo, useState, useEffect} from 'react';
+
+import {Translate} from 'indico/react/i18n';
+
+export default function WTFPrivacyPolicyURLsField({fieldId, wrapperId, initialValues}) {
```
ThiefMaster, line 13: The naming could be much more generic, e.g. `WTFLinkListField`. Sure, we might only use it for the privacy notices, but there's no reason for not using it elsewhere if we ever want the same functionality there.

indico/web/client/js/react/components/WTFPrivacyPolicyURLsField.jsx
```diff
@@ -0,0 +1,136 @@
...
+  const parentElement = useMemo(() => document.getElementById(wrapperId), [wrapperId]);
```
ThiefMaster, line 14: Probably not needed, you can just trigger the event on the hidden input itself (you can use a React ref to access it)

indico/web/client/js/react/components/WTFPrivacyPolicyURLsField.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+  // Trigger change only after the DOM has changed
+  useEffect(() => {
+    parentElement.dispatchEvent(new Event('change', {bubbles: true}));
+  }, [values, parentElement]);
+
+  const handleChange = (id, field) => () => {
+    const idx = values.findIndex(value => value.id === id);
+    const newValues = [...values];
+    newValues[idx][field] = document.getElementById(`${id}-${field}`).value;
```
ThiefMaster, line 33: Accessing the DOM for this is a bit ugly; you could instead keep the latest values in the React state (ie making the inputs controlled components)

indico/web/client/js/react/components/WTFPrivacyPolicyURLsField.jsx
```diff
@@ -0,0 +1,136 @@
...
+    setValues(newValues);
+  };
+
+  const getProcessedValues = () => {
+    const newValues = values.filter(value => value.title !== '' || value.url !== '');
```
ThiefMaster, line 38: `value.title || value.url`

indico/web/client/js/react/components/WTFPrivacyPolicyURLsField.jsx
```diff
@@ -0,0 +1,136 @@
...
+    if (newValues.length === 1) {
+      return [{title: '', url: newValues[0].url}];
+    }
+    return newValues.map(value => ({
```
ThiefMaster, line 42: This looks like it's just copying newValues...

ThiefMaster (changes requested):

indico/web/client/js/react/components/WTFPrivacyPolicyURLsField.jsx
duartegalvao: That is for removing the `id` attribute, which is only used for the front-end.. Might not be necessary since the back-end won't store it anyway

ThiefMaster: No, we shouldn't send unnecessary data to the backend. But I think with my other suggested changes we may not need this ID at all?

indico/web/templates/forms/link_list_widget.html
```diff
@@ -2,13 +2,13 @@
 
 {% block html %}
     <div id="{{ field.id }}-wrapper"></div>
+    <input type="hidden" id="{{ field.id }}" name="{{ field.id }}" value="{{ field.data | tojson | forceescape }}" />
```
ThiefMaster, line 5: I don't remember the exact difference but usually we use `field._value()` instead of `field.data` here

indico/web/forms/fields/simple.py
```diff
@@ -177,6 +177,12 @@ class IndicoMultipleTagSelectField(SelectMultipleField):
 class IndicoLinkListField(JSONField):
     widget = JinjaWidget('forms/link_list_widget.html', single_kwargs=True, single_line=True)
 
+    def process_formdata(self, valuelist):
+        super().process_formdata(valuelist)
+        self.data = [link for link in self.data if link.get('title') or link.get('url')]
```
ThiefMaster, line 182: Why is the filtering needed? Shouldn't we always have link and url keys in the data?

duartegalvao: Because of the changes to the component (to make values persist on refresh) we can't really filter empty entries client-side anymore without potentially losing info when refreshing. So I moved the filtering to the server-side

indico/web/templates/forms/link_list_widget.html
duartegalvao: It seems that for `JSONField`s, `field._value()` returns the data in escaped json. Quite convenient, thanks!

ThiefMaster: It's just dumped, you still need `|forceescape`

indico/web/forms/fields/simple.py
ThiefMaster: Shouldn't client-side validation fail and prevent submission in that case? And when you do submit with an empty URL then the normal validation should fail.

ThiefMaster: Oh I see, the "save now" corner button works even if some fields are invalid...

duartegalvao: There is also the case of when the field is empty - there is always one element on the links list. I could change that but it may complicate the component logic a bit

I also don't see a reason to bother the user if there are empty entries on the list. I don't see what harm it would cause to "silently" delete them

ThiefMaster: When I tested it yesterday it felt all fine, so I think we can leave it as it is

ThiefMaster (approved):

PR #5209, opened 2022-01-10
Make PersonLinkField entries draggable

plourenco:


indico/modules/events/abstracts/fields.py
```diff
@@ -118,9 +118,6 @@ def __init__(self, *args, **kwargs):
         self.empty_message = _('There are no authors')
         super().__init__(*args, **kwargs)
 
-    def _convert_data(self, data):
-        return list({self._get_person_link(x) for x in data})
```
ThiefMaster, line 122: the deduplication (that's why it's converting to a set and then back to a list) is no longer needed?

indico/web/client/js/react/components/PersonLinkField.jsx
```diff
@@ -278,63 +310,67 @@ export default function PersonLinkField({
...
     <div styleName="person-link-field">
-      <Segment attached="top" styleName="segment">
-        {sections.map(({name, label, plural}) => {
-          const filterCondition = p => p.roles && p.roles.includes(name);
-          const filtered = persons.filter(filterCondition);
-          return filtered.length === 0 ? null : (
+      <DndProvider backend={HTML5Backend}>
+        <Segment attached="top" styleName="segment">
+          {sections.map(({name, label, plural}) => {
+            const filterCondition = p => p.roles && p.roles.includes(name);
```
ThiefMaster, line 316: Would `p => p.roles?.includes(name)` work here?

plourenco: Yes, sure!

indico/modules/events/abstracts/fields.py
plourenco: Yes, it is needed, I ignored the de-duplication here while trying to avoid repeating it, thanks!

ThiefMaster (approved):

PR #5213, opened 2022-01-12
Allow submitters to edit their own contributions

### LINKED ISSUE #5187: Allow a submitter to modify their own contribution details
The modification of the details of the contribution is only allowed by event managers and session coordinators.

The submitter should be able to update these details (title, description, co-authors…) but upon a configurable option in order to keep existing events unaffected.

tomasr8:
Closes #5187 

If configured in the Contribution settings -> Submitter privileges, this change allows
abstract submitters to edit their contributions once the abstracts are accepted.

image
image

### DIFF
--- a/indico/modules/events/contributions/controllers/management.py
+++ b/indico/modules/events/contributions/controllers/management.py
@@ -191,11 +191,10 @@ class RHCreateContribution(RHManageContributionsBase):
 
 
 class RHEditContribution(RHManageContributionBase):
-    """
-    Edit the contribution.
+    """Edit a contribution.
 
-    If configured in the contribution settings,
-    editing is also allowed to the submitters of the given contribution.
+    If configured in the contribution settings, editing is also
+    allowed for the submitters.
     """
 
     def _check_access(self):
@@ -355,9 +354,8 @@ class RHCreateSubContribution(RHManageContributionBase):
 class RHEditSubContribution(RHManageSubContributionBase):
     """Edit a subcontribution.
 
-    If configured in the contribution settings,
-    editing of the given subcontribution is also allowed to
-    the submitters of the parent contribution.
+    If configured in the contribution settings, editing is also
+    allowed for the submitters of the parent contribution.
     """
 
     def _check_access(self):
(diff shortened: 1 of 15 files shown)


indico/modules/events/abstracts/forms.py
```diff
@@ -125,6 +125,9 @@ class AbstractSubmissionSettingsForm(IndicoForm):
                                                     enum=SubmissionRightsType, sorted=True,
                                                     description=_('Specify who will get contribution submission rights '
                                                                   'once an abstract has been accepted'))
+    allow_submitters_edit_contributions = BooleanField(_('Allow submitters edit contributions'), widget=SwitchWidget(),
```
ThiefMaster, line 128: Should this setting really be in the CfA? It's not really related to abstracts...

indico/modules/events/contributions/models/contributions.py
```diff
@@ -547,6 +547,14 @@ def can_manage(self, user, permission=None, allow_admin=True, check_parent=True,
...
                 session_coordinator_priv_enabled(self.event, 'manage-contributions')):
             return True
+        # Submitters can manage their own contributions if configured
+        from indico.modules.events.abstracts.settings import abstracts_settings
+        submitters_can_manage = abstracts_settings.get(self.event, 'allow_submitters_edit_contributions')
+        if submitters_can_manage and any(
+            principal.has_management_permission('submit') and principal.user == user
+            for principal in self.acl_entries
+        ):
+            return True
```
ThiefMaster, line 557: The check could be simplified a lot by passing `permission='submit'` to the super `can_manage()` check at the beginning when submitters are allowed to edit.

However, they should not have full management permissions on the contribution - this would also allow things such as deleting it or even changing its schedule. They should just be able to edit basic data like title, description, speakers, etc.

plourenco (approved):
Thanks, it works well!
I just noticed that a submitter could indeed remove himself from the list or their own permissions, not sure if anyone wants to comment on this but I would say probably not worth bothering.

tomasr8:
I noticed the same thing but not sure whether this is really a problem..

ThiefMaster:
Don't we have some check for this when managing an event and removing yourself? Basically warning when you want to remove yourself (or remove your submission permissions)? The same could be nice here as well...

indico/modules/events/contributions/forms.py
```diff
@@ -34,6 +34,7 @@
 
 
 class ContributionForm(IndicoForm):
+    _submitter_editable_fields = ('title', 'description', 'duration', 'person_link_data')
```
ThiefMaster, line 37: Do we really want the duration to be editable? This is a mess once things are scheduled...

cc @plourenco

plourenco: Yes, was wondering about the same. Not needed, JACoW also confirmed.

indico/modules/events/contributions/controllers/management.py
```diff
@@ -340,7 +353,18 @@ def _process(self):
 
 
 class RHEditSubContribution(RHManageSubContributionBase):
-    """Edit the subcontribution."""
+    """
+    Edit the subcontribution.
```
ThiefMaster, line 357:
```suggestion
    """Edit a subcontribution.
```

the "header' goes into the `"""` line

ThiefMaster (commented):
changelog entry please :)

PR #5216, opened 2022-01-17
Make i-forms fields fluid

plourenco:
Depends on #5209 

imagem


plourenco:
I have the same feeling, I'm considering making them [each field] full-width, but constraint the max-width in the parent (form) component to have a better result, at least in places where it might look weird.

PR #5217, opened 2022-01-19
Add support for abstract cloning

plourenco:
Adds support for a new cloning handle, that generates a new abstract form with pre-filled information.
Also ensures a single row is selected/cloned at a time.

For the sake of simplicity, I'm re-using the existing `RHCreateAbstract`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,7 @@ Bugfixes
   timeline (:pr:`5218`)
 - Fix incorrect date in multi-day meeting date selector dropdown in certain timezones
   (:pr:`5223`)
+- Add support for abstract cloning (:pr:`5217`)
 
 
 Version 3.1
@@ -68,7 +69,6 @@ Improvements
 - Add background color option and layer order to badge/poster designer items (:pr:`5139`,
   thanks :user:`SegiNyn`)
 - Allow external users in event/category ACLs (:pr:`5146`)
-- Add support for abstract cloning (:pr:`5217`)
 
 Bugfixes
 ^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,7 @@ Improvements
 ^^^^^^^^^^^^
 
 - Prompt before leaving the event protection page without saving changes (:pr:`5222`)
+- Add the ability to clone abstracts (:pr:`5217`)
 
 Bugfixes
 ^^^^^^^^
@@ -19,7 +20,6 @@ Bugfixes
   timeline (:pr:`5218`)
 - Fix incorrect date in multi-day meeting date selector dropdown in certain timezones
   (:pr:`5223`)
-- Add support for abstract cloning (:pr:`5217`)
 
 
 Version 3.1
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -68,6 +68,7 @@ Improvements
 - Add background color option and layer order to badge/poster designer items (:pr:`5139`,
   thanks :user:`SegiNyn`)
 - Allow external users in event/category ACLs (:pr:`5146`)
+- Add support for abstract cloning (:pr:`5217`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/events/abstracts/controllers/abstract_list.py
+++ b/indico/modules/events/abstracts/controllers/abstract_list.py
@@ -10,6 +10,7 @@ from operator import attrgetter
 
 from flask import flash, jsonify, request, session
 from sqlalchemy.orm import joinedload, subqueryload
+from webargs import fields
 
 from indico.core.db import db
 from indico.core.db.sqlalchemy.util.models import get_simple_column_attrs
@@ -29,7 +30,6 @@ from indico.modules.events.contributions.models.persons import AuthorType
 from indico.modules.events.util import get_field_values
 from indico.modules.users.models.users import User
 from indico.util.i18n import _, ngettext
-from indico.util.marshmallow import ModelField
 from indico.web.args import use_kwargs
 from indico.web.util import jsonify_data, jsonify_form, jsonify_template
 
@@ -110,14 +110,11 @@ class RHAbstractListStaticURL(RHAbstractListBase):
 
 
 class RHCreateAbstract(RHAbstractListBase):
-    """Create an abstract from scratch or with existing cloned data"""
+    """Create an abstract from scratch or with existing cloned data."""
 
-    @staticmethod
-    def clone_fields(abstract):
+    def clone_fields(self, abstract):
         field_names = ['title', 'description', 'submission_comment', 'submitted_for_tracks', 'submitted_contrib_type']
-        field_data = {}
-        for f in field_names:
-            field_data[f] = getattr(abstract, f)
+        field_data = {f: getattr(abstract, f) for f in field_names}
         person_links = []
         link_attrs = get_simple_column_attrs(AbstractPersonLink)
         for old_link in abstract.person_links:
@@ -129,14 +126,20 @@ class RHCreateAbstract(RHAbstractListBase):
             field_data[f'custom_{f.contribution_field_id}'] = f.data
         return field_data
 
-    @use_kwargs({'abstract': ModelField(Abstract, data_key='abstract_id')}, location='query')
-    def _process(self, abstract=None):
+    @use_kwargs({'abstract_id': fields.Int()}, location='query')
+    def _process_args(self, abstract_id=None):
+        RHAbstractListBase._process_args(self)
+        self.abstract = None
+        if abstract_id:
+            self.abstract = Abstract.query.with_parent(self.event).filter_by(id=abstract_id).first_or_404()
+
+    def _process(self):
         is_invited = request.args.get('invited') == '1'
         abstract_form_class = make_abstract_form(self.event, session.user, notification_option=True,
                                                  management=self.management, invited=is_invited)
         cloned_fields = {}
-        if abstract:
-            cloned_fields = self.clone_fields(abstract)
+        if self.abstract:
+            cloned_fields = self.clone_fields(self.abstract)
         form = abstract_form_class(event=self.event, management=self.management, invited=is_invited, **cloned_fields)
         if is_invited:
             del form.submitted_contrib_type
@@ -170,7 +173,7 @@ class RHCreateAbstract(RHAbstractListBase):
             if tpl_components.get('hide_abstract'):
                 self.list_generator.flash_info_message(abstract)
             return jsonify_data(**tpl_components)
-        return jsonify_form(form, back=_('Cancel'), disabled_until_change=not abstract,
+        return jsonify_form(form, back=_('Cancel'), disabled_until_change=not self.abstract,
                             form_header_kwargs={'action': request.relative_url})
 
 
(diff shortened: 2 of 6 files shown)


indico/modules/events/abstracts/controllers/abstract_list.py
```diff
@@ -107,11 +110,34 @@ def _process(self):
...
+            field_data[f] = getattr(abstract, f)
+        person_links = []
+        link_attrs = get_simple_column_attrs(AbstractPersonLink)
+        for old_link in abstract.person_links:
+            link = AbstractPersonLink(person=old_link.person)
+            link.populate_from_attrs(old_link, link_attrs)
+            person_links.append(link)
+        field_data['person_links'] = person_links
+        for f in abstract.field_values:
+            field_data[f'custom_{f.contribution_field_id}'] = f.data
```
plourenco, line 129: Custom field names are generated dynamically. While this isn't the most robust approach, RFC if it's worth changing.

```python
def make_abstract_form:
...
for custom_field in event.contribution_fields:
   ...
   name = f'custom_{custom_field.id}'
```

indico/modules/events/abstracts/controllers/abstract_list.py
```diff
@@ -107,11 +110,34 @@ def _process(self):
 
 
 class RHCreateAbstract(RHAbstractListBase):
-    def _process(self):
+    """Create an abstract from scratch or with existing cloned data"""
+
+    @staticmethod
```
ThiefMaster, line 115: I don't think there's any benefit in marking it as static since we never use it in a static context..

indico/modules/events/abstracts/controllers/abstract_list.py
```diff
@@ -107,11 +110,34 @@ def _process(self):
...
+        for old_link in abstract.person_links:
+            link = AbstractPersonLink(person=old_link.person)
+            link.populate_from_attrs(old_link, link_attrs)
+            person_links.append(link)
+        field_data['person_links'] = person_links
+        for f in abstract.field_values:
+            field_data[f'custom_{f.contribution_field_id}'] = f.data
+        return field_data
+
+    @use_kwargs({'abstract': ModelField(Abstract, data_key='abstract_id')}, location='query')
```
ThiefMaster, line 132: :warning: This lets you clone abstracts from other events as well!

indico/modules/events/abstracts/controllers/abstract_list.py
ThiefMaster: looks good enough IMHO

indico/modules/events/client/js/util/list_generator.js
```diff
@@ -39,10 +39,11 @@
...
         .closest('tr')
         .toggleClass('selected', this.checked);
-      $('.js-requires-selected-row').toggleClass(
-        'disabled',
-        !$('.list input:checkbox:checked').length
-      );
+      const rows = $('.list input:checkbox:checked').length;
+      $('.js-requires-selected-row').toggleClass('disabled', !rows);
+      if (rows > 1) {
+        $('.js-requires-selected-row.single-row').toggleClass('disabled', true);
```
ThiefMaster, line 45: Wouldn't be more intuitive (w/o the if)?

```js
$('.js-requires-selected-row.single-row').toggleClass('disabled', rows != 1);
```

plourenco: Sure

indico/modules/events/abstracts/controllers/abstract_list.py
plourenco: Nice catch!

I noticed there is a `get_query` for it, do you think it would be more reasonable to go back to the `_process_args` method or have a new decorator (with `get_query=.with_parent(...)`) here given the abstract isn't always there?

ThiefMaster: It won't really work with the decorator since you don't have access to the event there - so I think using the webargs/marshmallow field here isn't that useful; you'd need to manually parse it with webargs within the method (like we sometimes to in the editing module) :/

indico/modules/events/abstracts/controllers/abstract_list.py
```diff
@@ -107,11 +110,42 @@ def _process(self):
 
 
 class RHCreateAbstract(RHAbstractListBase):
+    """Create an abstract from scratch or with existing cloned data"""
+
+    def __init__(self):
+        super().__init__()
+        self.abstract = None
```
ThiefMaster, line 117: I'd just set it to None in `_process_args` if there's no ID..

indico/modules/events/abstracts/controllers/abstract_list.py
```diff
@@ -107,11 +110,42 @@ def _process(self):
 
 
 class RHCreateAbstract(RHAbstractListBase):
+    """Create an abstract from scratch or with existing cloned data"""
```
ThiefMaster, line 113:
```suggestion
    """Create an abstract from scratch or with existing cloned data."""
```

indico/modules/events/abstracts/controllers/abstract_list.py
```diff
@@ -107,11 +110,42 @@ def _process(self):
...
+
+    def __init__(self):
+        super().__init__()
+        self.abstract = None
+
+    def clone_fields(self, abstract):
+        field_names = ['title', 'description', 'submission_comment', 'submitted_for_tracks', 'submitted_contrib_type']
+        field_data = {}
+        for f in field_names:
+            field_data[f] = getattr(abstract, f)
```
ThiefMaster, line 123:
```python
field_data = {f: getattr(abstract, f) for f in field_names}
```

indico/modules/events/abstracts/templates/management/abstract_list.html
```diff
@@ -207,6 +207,43 @@
                         </li>
                     </ul>
                 </div>
+                <div class="group">
+                    <a class="i-button arrow icon-copy js-requires-selected-row single-row
+                        hide-if-locked" data-toggle="dropdown">
```
ThiefMaster, line 212: I'd put this all in one line tbh...

indico/modules/events/abstracts/controllers/abstract_list.py
```diff
@@ -107,11 +110,37 @@ def _process(self):
...
+            self.abstract = Abstract.query.with_parent(self.event).filter_by(id=abstract_id).first_or_404()
+
     def _process(self):
         is_invited = request.args.get('invited') == '1'
         abstract_form_class = make_abstract_form(self.event, session.user, notification_option=True,
                                                  management=self.management, invited=is_invited)
-        form = abstract_form_class(event=self.event, management=self.management, invited=is_invited)
+        cloned_fields = {}
+        if self.abstract:
+            cloned_fields = self.clone_fields(self.abstract)
```
ThiefMaster, line 142:
```suggestion
        cloned_fields = self.clone_fields(self.abstract) if self.abstract else {}
```

indico/modules/events/abstracts/templates/management/abstract_list.html
```diff
@@ -207,6 +207,42 @@
...
                 </div>
+                <div class="group">
+                    <a class="i-button arrow icon-copy js-requires-selected-row single-row hide-if-locked" data-toggle="dropdown">
+                        {% trans %}Clone{% endtrans %}
+                    </a>
+                    <ul class="i-dropdown">
+                        <li>
+                            <a class="js-dialog-action"
+                               title="{% trans %}Add new abstract{% endtrans %}"
+                               data-title="{% trans %}Add new abstract{% endtrans %}"
```
ThiefMaster, line 218: Maybe these titles should mention cloning?

plourenco: Sure, I didn't initially because it could be confused with a straight clone, instead of a new form.
However, maybe something like "Clone into an abstract" or just "Clone abstract"?

plourenco: Now that I'm revising this, I don't think the titles make sense. Because after all, the UI and actions are the exact same as adding a new abstract, it's just the information that gets pre-filled, as if someone was writing it manually (+ new i18n strings). Do you think we should really change them?

imagem


imagem

ThiefMaster: hm... no strong opinion tbh - so i'll leave it up to you (or maybe get some more opinions in our MM channel)

CHANGES.rst
```diff
@@ -49,6 +49,7 @@ Improvements
 - Add background color option and layer order to badge/poster designer items (:pr:`5139`,
   thanks :user:`SegiNyn`)
 - Allow external users in event/category ACLs (:pr:`5146`)
+- Add support for abstract cloning (:pr:`5217`)
```
ThiefMaster, line 52: This is for 3.1.1, not 3.1 (already released). Once you rebase, the changelog should include the 3.1.1 section

indico/modules/events/abstracts/controllers/abstract_list.py
```diff
@@ -107,11 +110,35 @@ def _process(self):
...
+            link = AbstractPersonLink(person=old_link.person)
+            link.populate_from_attrs(old_link, link_attrs)
+            person_links.append(link)
+        field_data['person_links'] = person_links
+        for f in abstract.field_values:
+            field_data[f'custom_{f.contribution_field_id}'] = f.data
+        return field_data
+
+    @use_kwargs({'abstract_id': fields.Int()}, location='query')
+    def _process_args(self, abstract_id=None):
```
ThiefMaster, line 130:
```suggestion
    @use_kwargs({
        'abstract_id': fields.Int(load_default=None)
    }, location='query')
    def _process_args(self, abstract_id):
```

We usually use `load_default` (formerly `missing`) instead of setting a default value on the kwarg. No real difference, just for consistency

indico/modules/events/client/js/util/list_generator.js
```diff
@@ -39,10 +39,9 @@
       $(this)
         .closest('tr')
         .toggleClass('selected', this.checked);
-      $('.js-requires-selected-row').toggleClass(
-        'disabled',
-        !$('.list input:checkbox:checked').length
-      );
+      const rows = $('.list input:checkbox:checked').length;
+      $('.js-requires-selected-row').toggleClass('disabled', !rows);
+      $('.js-requires-selected-row.single-row').toggleClass('disabled', rows !== 1);
```
ThiefMaster, line 44: I think this change is no longer needed

indico/modules/events/abstracts/controllers/abstract_list.py
```diff
@@ -144,7 +171,8 @@ def _process(self):
             if tpl_components.get('hide_abstract'):
                 self.list_generator.flash_info_message(abstract)
             return jsonify_data(**tpl_components)
-        return jsonify_form(form, back=_('Cancel'), form_header_kwargs={'action': request.relative_url})
+        return jsonify_form(form, back=_('Cancel'), disabled_until_change=not self.abstract,
```
ThiefMaster, line 174: At least when cloning as an invited abstract the form will always require some input (the intended submitter).. so this could be `disabled_until_change=(not self.abstract or is_invited)`

CHANGES.rst
```diff
@@ -19,6 +19,7 @@ Bugfixes
   timeline (:pr:`5218`)
 - Fix incorrect date in multi-day meeting date selector dropdown in certain timezones
   (:pr:`5223`)
+- Add support for abstract cloning (:pr:`5217`)
```
ThiefMaster, line 22: that's not really a bugfix but rather an improvement :P

plourenco: My bad, this was auto-merged

PR #5218, opened 2022-01-20
Fix revision file access check

duartegalvao:
This PR contains a fix for a bug relating to a contribution's published revision files only being accessible to users with access to the editing timeline, instead of all users with access to the contribution.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,18 @@ Changelog
 =========
 
 
+Version 3.1.1
+-------------
+
+*Unreleased*
+
+Bugfixes
+^^^^^^^^
+
+- Fix published editable files only being visible to users with access to the editing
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -59,6 +59,8 @@ Bugfixes
 - Fix editing service API calls using the service token (:pr:`5170`)
 - Fix excessive retries for Celery tasks with a retry wait time longer
   than 1 hour (:pr:`5172`)
+- Fix published editable files only being visible to users with access to the editing
+  timeline (:pr:`5218`)
```
ThiefMaster, line 63: Add this for 3.1.1; looks like I just forgot to add the changelog stub for that after releasing 3.1 - maybe you could include a commit for this in your PR? See ef52ea02e6bf2 for the structure

PR #5225, opened 2022-01-27
Remove padding on category title

javfg:
This PR removes the padding from the category title. It was wrapping in small screens.

image

Becomes:

image


ThiefMaster:
Do we want a changelog entry for this, "Remove excessive padding around category titles" or sth like that?

PR #5231, opened 2022-01-31
Fix undefined rb notification fields

plourenco:
Fixes an undefined string value shown in room booking room creation.

imagem


indico/modules/rb/client/js/common/rooms/edit/RoomEditNotifications.jsx
```diff
@@ -35,11 +35,7 @@ export default function RoomEditNotifications({active, defaults}) {
           <FinalInput
             fluid
             name="notification_before_days"
-            placeholder={
-              defaults.notification_before_days === null
-                ? ''
-                : String(defaults.notification_before_days)
-            }
+            placeholder={defaults.notification_before_days}
```
ThiefMaster, line 38: won't this trigger some warning about passing a number to a prop expecting a string?

PR #5234, opened 2022-02-02
Badge item and Event participant list signals

SegiNyn:


indico/core/signals/event/designer.py
```diff
@@ -15,3 +15,8 @@
 The registration form is passed in the `regform` kwarg. The list of registration
 objects are passed in the `registrations` kwarg and it may be modified.
 ''')
+
+customize_badge_style = _signals.signal('customize-badge-style', '''
+Called when printing a badge.
+Sends each `item` in the badge and recieves a dictionary of styles for that `item`
```
ThiefMaster, line 21: receives

indico/modules/designer/pdf.py
```diff
@@ -92,22 +94,27 @@ def get_pdf(self):
         return data
 
     def _draw_item(self, canvas, item, tpl_data, content, margin_x, margin_y):
+        override = values_from_signal(signals.event.designer.customize_badge_style.send(item), as_list=True)
```
ThiefMaster, line 97: Wouldn't it be better to put the original data in a dict, thne pass is to the signal, and then use that dict, updated with what the signal returned, to get the data for assigning it to `style.whatever`?

ie somehing like this

```
        styles = {alignment: ..., text_color: item.get(...), ...}
        for update in values_from_signal(
            signals.event.designer.customize_badge_style.send(item, styles=styles),
            as_list=True
        ):
            styles.update(update)
```

indico/modules/events/registration/__init__.py
```diff
@@ -106,7 +107,8 @@ def _visible_registration(event):
                 .has_rows())
 
     def _visible_participant_list(event):
-        return event.has_feature('registration')
+        override = values_from_signal(signals.event.visible_participant_list.send(sender, event=event))
+        return event.has_feature('registration') if not override else override.pop()
```
SegiNyn, line 111: Hello @ThiefMaster , In this case where `override.pop()` was used, this was considering that only one plugin is subscribed to the signal. But in the case that there are multiple plugins subscribed, would it then be incorrect to only consider the first value ?

indico/modules/designer/pdf.py
```diff
@@ -92,22 +94,38 @@ def get_pdf(self):
...
-            style.fontName = FONT_STYLES[item['font_family']][0]
+            styles['fontName'] = FONT_STYLES[item['font_family']][0]
+
+        for update in values_from_signal(
+            signals.event.designer.customize_badge_style.send(item, styles=styles),
+            as_list=True
+        ):
+            styles.update(update)
+
+        style = ParagraphStyle({})
```
ThiefMaster, line 121: I didn't look more closely, but this `{}` feels like you might be able to pass `styles` directly instead of assigning one by one. Maybe you can check if that's indeed the case?

indico/modules/events/registration/__init__.py
```diff
@@ -106,7 +107,9 @@ def _visible_registration(event):
                 .has_rows())
 
     def _visible_participant_list(event):
-        return event.has_feature('registration')
+        override = values_from_signal(signals.event.visible_participant_list.send(sender, event=event))
+        override = all(override) if override else None
+        return event.has_feature('registration') if override is None else override
```
ThiefMaster, line 112: Should we even call the signal if the feature is disabled? I think we can return False early if the feature is disabled, and only then call the signal and take override into account..

SegiNyn: Okay, so the check can go something like 

```
        has_registration = event.has_feature('registration')
        if has_registration:
            override = values_from_signal(signals.event.visible_participant_list.send(sender, event=event))
            return all(override) if override else True
        return has_registration
```

ThiefMaster: Much easier:

```
if not event.has_feature('registration'):
    return False
...
```

OmeGak:
```python
```

indico/modules/designer/pdf.py
SegiNyn: It was not possible to pass the dict to the `ParagraphStyle` like this, the values were not added and the badge was printed without any style

indico/core/signals/event/designer.py
```diff
@@ -15,3 +15,9 @@
 The registration form is passed in the `regform` kwarg. The list of registration
 objects are passed in the `registrations` kwarg and it may be modified.
 ''')
+
+customize_badge_style = _signals.signal('customize-badge-style', '''
+Called when printing a badge.
+Each `item` in a badge and it's `styles` are passed in the kwarg. The `styles` is a dictionary
+which may be modified and is used to update the `item style`.
```
ThiefMaster, line 22: This is a bit misleading; while modifying the dict would in fact work, the way the signal is being called, it should rather return a dict containing just the updates...

indico/modules/designer/pdf.py
```diff
@@ -92,22 +94,38 @@ def get_pdf(self):
...
+        style.alignment = styles['alignment']
+        style.textColor = styles['textColor']
+        style.backColor = styles['backColor']
+        style.borderPadding = styles['borderPadding']
+        style.fontSize = styles['fontSize']
+        style.leading = styles['leading']
+        style.fontName = styles['fontName']
```
ThiefMaster, line 128: since they all match, you could simplify this as

```python
for key, value in styles.items():
    setattr(style, key, value)
```

indico/modules/events/registration/__init__.py
```diff
@@ -106,7 +107,9 @@ def _visible_registration(event):
                 .has_rows())
 
     def _visible_participant_list(event):
-        return event.has_feature('registration')
+        if not event.has_feature('registration'):
+            return False
+        return not any(values_from_signal(signals.event.hide_participant_list.send(sender, event=event)))
```
ThiefMaster, line 112: The sender is unused (I believe it's None), so I would not pass it on to this signal. What about using the event itself as the sender for `hide_participant_list` instead?

SegiNyn:
Hello, thank you for the reviews. We have also included a 3rd signal to this PR which is relating to badges, `draw_item_on_badge`.

indico/modules/events/registration/badges.py
```diff
@@ -99,6 +101,23 @@ def _draw_badge(self, canvas, registration, template, tpl_data, pos_x, pos_y):
             else:
                 continue
 
+            updates = {
+                'item': item,
+                'items': items,
```
ThiefMaster, line 106: Do you plan to modify these? If not I'd keep those outside and rather make the registration the sender and pass item (and items if needed) as separate kwargs. And then only keep the "simple" values like height and positions in the dict.

But if you could explain how you intend to use the signal it'll certainl be easier to see what would be the best approach here...

SegiNyn: We are modifying only the `item` and the `text`

We are using the signal to change some of the item position and color if it overlaps with other items. And also change the item color based on the registration badge attributes.

ThiefMaster: I would go for something like this then:

```python
            item_data = {'item': item, 'text': text, 'pos_x': pos_x, 'pos_y': pos_y}
            for update in values_from_signal(
                signals.event.designer.draw_item_on_badge.send(template, registration=registration, items=items, 
                                                               height=self.height, width=self.width, data=item_data),
                as_list=True
            ):
                item_data.update(update)
            self._draw_item(canvas, item_data['item'], tpl_data, item_data['text'], item_data['pos_x'],
                            item_data['pos_y'])
```

That way you keep the updateable data separate from the rest, and use a more logical sender (I don't have a strong opinion whether the registration or the template should be the sender BTW).

SegiNyn: Okay, I'll add these changes and use the registration as the sender. Thank you!

indico/modules/designer/pdf.py
```diff
@@ -92,22 +94,33 @@ def get_pdf(self):
...
-        style.fontSize = _extract_font_size(item['font_size'])
-        style.leading = style.fontSize
+        styles = {
+            'alignment': ALIGNMENTS[item['text_align']],
+            'textColor': item.get('color') or '#000000',
+            'backColor': item.get('background_color') or None,
+            'borderPadding': (0, 0, 4, 0),
+            'fontSize': _extract_font_size(item['font_size'])
+        }
+        styles['leading'] = styles['fontSize']
```
ThiefMaster, line 104: I'd just call `_extract_font_size` twice or assign `font_size = _extract_font_size(...)` and then use that in the dict literal

indico/modules/designer/pdf.py
```diff
@@ -92,22 +94,34 @@ def get_pdf(self):
...
+            styles['fontName'] = FONT_STYLES[item['font_family']][2]
         elif item['bold']:
-            style.fontName = FONT_STYLES[item['font_family']][1]
+            styles['fontName'] = FONT_STYLES[item['font_family']][1]
         else:
-            style.fontName = FONT_STYLES[item['font_family']][0]
+            styles['fontName'] = FONT_STYLES[item['font_family']][0]
+
+        for update in values_from_signal(
+            signals.event.designer.update_badge_style.send(item, styles=styles),
```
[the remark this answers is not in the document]
SegiNyn: Yes, good idea. Although neither the template nor the registration is available in this function, would it be a okay to use the `tpl_data` as the sender?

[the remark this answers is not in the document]
ThiefMaster: considering that this is not useful to match e.g. on the event or template used (check the data in there...) I don't think so. But I don't see any reason against passing this data. Just make sure posters still work since those use this method as well but obviously don't have a registration associated

PR #5235, opened 2022-02-03
Validate unknown placeholders in room_name_format & map_url_template

tomasr8:
Rejects unsupported placeholders, which prevents a 500 error
caused by directly calling .format() on the room_name_format

### DIFF
--- a/indico/modules/rb/schemas.py
+++ b/indico/modules/rb/schemas.py
@@ -400,22 +400,18 @@ class LocationArgs(mm.Schema):
         if missing:
             # validated client-side, no i18n needed
             raise ValidationError('Missing placeholders: {}'.format(', '.join(missing)))
-
-        try:
-            placeholders = get_format_placeholders(room_name_format)
-        except ValueError:
-            raise ValidationError(_('Invalid placeholder format'))
-        if invalid := set(placeholders) - {'building', 'floor', 'number'}:
-            # validated client-side, no i18n needed
-            raise ValidationError('Invalid placeholders: {}'.format(', '.join(invalid)))
+        self._validate_placeholders(room_name_format, {'building', 'floor', 'number'})
 
     @validates('map_url_template')
     def _check_map_url_template_placeholders(self, map_url_template, **kwargs):
+        self._validate_placeholders(map_url_template, {'id', 'building', 'floor', 'number', 'lat', 'lng'})
+
+    def _validate_placeholders(self, format_string, valid_placeholders):
         try:
-            placeholders = get_format_placeholders(map_url_template)
+            placeholders = get_format_placeholders(format_string)
         except ValueError:
             raise ValidationError(_('Invalid placeholder format'))
-        if invalid := set(placeholders) - {'id', 'building', 'floor', 'number', 'lat', 'lng'}:
+        if invalid := set(placeholders) - valid_placeholders:
             # validated client-side, no i18n needed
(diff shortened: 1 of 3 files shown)


indico/modules/rb/schemas.py
```diff
@@ -401,6 +401,11 @@ def _check_room_name_format_placeholders(self, room_name_format, **kwargs):
             # validated client-side, no i18n needed
             raise ValidationError('Missing placeholders: {}'.format(', '.join(missing)))
 
+        placeholders = get_format_placeholders(room_name_format)
+        for ph in placeholders:
+            if ph not in ['building', 'floor', 'number']:
+                raise ValidationError(_('Unknown placeholder "{ph}"').format(ph=ph))
```
ThiefMaster, line 407: Untested, but this can probably be simplified like this:

```python
if invalid := placeholders - {'building', 'floor', 'number'}:
    raise ValidationError(_('Invalid placeholders: {}').format(', '.join(invalid)))
```

indico/modules/rb/schemas.py
```diff
@@ -401,6 +401,10 @@ def _check_room_name_format_placeholders(self, room_name_format, **kwargs):
             # validated client-side, no i18n needed
             raise ValidationError('Missing placeholders: {}'.format(', '.join(missing)))
 
+        placeholders = get_format_placeholders(room_name_format)
+        if invalid := set(placeholders) - {'building', 'floor', 'number'}:
```
ThiefMaster, line 405: This check should be done client-side as well

indico/modules/rb/schemas.py
```diff
@@ -401,6 +401,10 @@ def _check_room_name_format_placeholders(self, room_name_format, **kwargs):
...
+            raise ValidationError(_('Invalid placeholders: {}').format(', '.join(invalid)))
```
ThiefMaster, line 406: I'd skip the i18n here; this should never trigger due to the client-side validation so no need for translators to translate it

indico/modules/rb/schemas.py
```diff
@@ -401,6 +401,10 @@ def _check_room_name_format_placeholders(self, room_name_format, **kwargs):
             # validated client-side, no i18n needed
             raise ValidationError('Missing placeholders: {}'.format(', '.join(missing)))
 
+        placeholders = get_format_placeholders(room_name_format)
```
ThiefMaster, line 404: Maybe we can catch ValueError here and turn it into something like "Invalid format string" (as a ValidationError, this one with i18n)? This happens e.g. if you put just `{` at the end of the string.

indico/modules/rb/schemas.py
```diff
@@ -401,6 +401,24 @@ def _check_room_name_format_placeholders(self, room_name_format, **kwargs):
...
+
+    @validates('map_url_template')
+    def _check_map_url_template_placeholders(self, map_url_template, **kwargs):
+        try:
+            placeholders = get_format_placeholders(map_url_template)
+        except ValueError:
+            raise ValidationError(_('Invalid placeholder format'))
+        if invalid := set(placeholders) - {'id', 'building', 'floor', 'number', 'lat', 'lng'}:
+            # validated client-side, no i18n needed
+            raise ValidationError('Invalid placeholders: {}'.format(', '.join(invalid)))
```
ThiefMaster, line 420: What about adding a `def validate_placeholders(value, placeholders)` util to avoid duplicating all this?

PR #5236, opened 2022-02-03
Update the editing colors

plourenco:


indico/modules/events/editing/client/js/editing/timeline/StateIndicator.jsx
```diff
@@ -52,8 +51,8 @@ export default function StateIndicator({label, circular, tooltip, state, monochr
         disabled={!tooltip}
       />
       {label && (
-        <div styleName="label-text" style={{color: monochrome ? Palette.gray : labelColors[state]}}>
-          {EditableStatus[state]}
+        <div styleName="label-text" style={{color: monochrome ? Palette.gray : labelColor}}>
+          {EditableStatus[state] || Translate.string('Unknown')}
```
ThiefMaster, line 55: when is the state unknown?

plourenco: Most likely never, it was a defensive strategy as I tried to manually add a `replaced` revision and ended up with an empty state. It would possibly prevent future occasions where new states were added but not updated here. 
If this seems excessive, I can get rid of it.

I also noticed there is no "replaced" state as it always ends up as "ready for review", I guess I can remove it.

ThiefMaster: IIRC replaced should show up on the original revision after the editing service replaced it with an updated one

ThiefMaster:
master or 3.1.x? I'd be OK with the latter but if it's not a priority for JACoW I'm of course fine with keeping this on master

indico/modules/events/editing/client/js/editing/timeline/StateIndicator.jsx
ThiefMaster: did you check if there's something wrong w/ the replaced state in the end if it never shows up? probably outside this PR's scope though...

plourenco: I see what you mean now, I was neglecting the "dot color" near each revision but looking towards the status list.
In any case, there were no suggestions for a "replaced" colour, so we can keep it as-is.

PR #5238, opened 2022-02-04
RB: Optional placeholders

### LINKED ISSUE #5226: [Feature] Allow to use `{site}` in `room_name_format`
The placeholder `{site}` can not be added to `room_name_format`. Checking the code, the placeholder `{site}` is not in the list of placeholders:
https://github.com/indico/indico/blob/beca2a26e649b2548aeb7fa9740e1fd24314c8b0/indico/modules/rb/models/locations.py#L50

For rather distributed organizations, adding `{site}` to the pattern would be extremely helpful.

### LINKED ISSUE #5227: [Feature] Allow to remove placeholders from `room_name_format`
It is currently not possible to remove individual placeholders from `room_name_format`. For example, for some organizations, it might be useful to drop the `{floor}` in case the `{number}` pattern already holds the floor due to a consistent room naming scheme. 

Possibility to remove individual placeholders from `room_name_format`. 

**Additional context**
Currently, the `input` field shows a `Missing placeholder(s)` message if not all placeholders are used:
https://github.com/indico/indico/blob/5c9fb38207175932a3c285796ca3b82ef0ff56bf/indico/modules/rb/client/js/modules/admin/SettingsPage.jsx#L141-L142

tomasr8:
Closes #5227
Closes #5226

- Make placeholders in `room_name_format` optional, except for the room number which is always required.
- Add {site} to placeholders that can be used

### DIFF
--- a/indico/modules/rb/client/js/common/rooms/edit/RoomEditLocation.jsx
+++ b/indico/modules/rb/client/js/common/rooms/edit/RoomEditLocation.jsx
@@ -29,7 +29,11 @@ export default function RoomEditLocation({active, roomNameFormat}) {
       />
       <Form.Group widths="four">
         <Form.Field width={8}>
-          <FinalInput name="site" label={Translate.string('Site')} />
+          <FinalInput
+            name="site"
+            label={Translate.string('Site')}
+            required={isPlaceholderRequired(roomNameFormat, 'site')}
+          />
         </Form.Field>
         <FinalInput
           name="building"
--- a/indico/modules/rb/client/js/common/rooms/edit/RoomEditLocation.jsx
+++ b/indico/modules/rb/client/js/common/rooms/edit/RoomEditLocation.jsx
@@ -12,7 +12,9 @@ import {Form, Header, Tab} from 'semantic-ui-react';
 import {FinalInput, parsers as p, validators as v} from 'indico/react/forms';
 import {Translate} from 'indico/react/i18n';
 
-export default function RoomEditLocation({active}) {
+export default function RoomEditLocation({active, roomNameFormat}) {
+  const isPlaceholderRequired = (format, name) => format.includes(`{${name}}`);
+
   return (
     <Tab.Pane active={active}>
       <Header>
@@ -29,9 +31,21 @@ export default function RoomEditLocation({active}) {
         <Form.Field width={8}>
           <FinalInput name="site" label={Translate.string('Site')} />
         </Form.Field>
-        <FinalInput name="building" label={Translate.string('Building')} required />
-        <FinalInput name="floor" label={Translate.string('Floor')} required />
-        <FinalInput name="number" label={Translate.string('Number')} required />
+        <FinalInput
+          name="building"
+          label={Translate.string('Building')}
+          required={isPlaceholderRequired(roomNameFormat, 'building')}
+        />
+        <FinalInput
+          name="floor"
+          label={Translate.string('Floor')}
+          required={isPlaceholderRequired(roomNameFormat, 'floor')}
+        />
+        <FinalInput
+          name="number"
+          label={Translate.string('Number')}
+          required={isPlaceholderRequired(roomNameFormat, 'number')}
+        />
       </Form.Group>
       <Form.Group widths="equal">
         <FinalInput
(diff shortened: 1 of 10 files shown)


indico/migrations/versions/20220203_1524_f046e456629f_make_building_optional.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'f046e456629f'
+down_revision = '3dafee32ba7d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.alter_column('rooms', 'building', server_default='', schema='roombooking')
```
ThiefMaster, line 19: I don't think this is needed; we aren't using server_default in the models either...

indico/modules/rb/client/js/common/rooms/edit/RoomEditLocation.jsx
```diff
@@ -12,7 +12,11 @@ import {Form, Header, Tab} from 'semantic-ui-react';
 import {FinalInput, parsers as p, validators as v} from 'indico/react/forms';
 import {Translate} from 'indico/react/i18n';
 
-export default function RoomEditLocation({active}) {
+function isRequired(roomNameFormat, fieldName) {
```
ThiefMaster, line 15: Maybe `isPlaceholderRequired` would be clearer?

ThiefMaster: I would *probably* define this function inline within the component..

```
const isPlaceholderRequired = (format, name) => format.includes(`{${name}}`);
```

ThiefMaster:
#5226 could probably be included in this same PR

indico/modules/rb/client/js/common/rooms/edit/RoomEditModal.jsx
```diff
@@ -102,6 +104,17 @@ function RoomEditModal({roomId, locationId, onClose, afterCreation}) {
     );
   }, [roomId]);
 
+  useEffect(() => {
```
ThiefMaster, line 107: I'd include it in this block here to avoid toggling the loading state multiple times

```js
  useEffect(() => {
    (async () => {
      if (!isNewRoom) {
        setLoading(true);
        await fetchRoomData();
        setLoading(false);
      }
    })();
  }, [isNewRoom, fetchRoomData]);
```

indico/modules/rb/client/js/common/rooms/edit/RoomEditModal.jsx
```diff
@@ -102,6 +104,17 @@ function RoomEditModal({roomId, locationId, onClose, afterCreation}) {
...
+    (async () => {
+      if (!roomNameFormat) {
```
ThiefMaster, line 109: Is this check really correct? If the locationId changed for some reason (probably never happens) we *would* need to fetch the format again

ThiefMaster:
This needs a rebase since the changes from your other PR are causing some conflicts.

indico/modules/rb/client/js/modules/admin/LocationPage.jsx
```diff
@@ -112,16 +112,10 @@ class LocationPage extends React.PureComponent {
...
-            return PluralTranslate.string(
-              'Missing placeholder: {placeholders}',
-              'Missing placeholders: {placeholders}',
-              missing.length,
-              {placeholders: missing.join(', ')}
-            );
+          if (!val.includes('{number}')) {
+            return Translate.string('Missing placeholder: number');
           }
+          const validPlaceholders = ['{building}', '{floor}', '{number}'];
```
ThiefMaster, line 118: `{site}` is missing here

indico/modules/rb/schemas.py
```diff
@@ -396,7 +396,7 @@ def _check_name_unique(self, name, **kwargs):
 
     @validates('room_name_format')
     def _check_room_name_format_placeholders(self, room_name_format, **kwargs):
-        self._validate_placeholders(room_name_format, {'building', 'floor', 'number'}, {'building', 'floor', 'number'})
+        self._validate_placeholders(room_name_format, {'building', 'floor', 'number'}, {'number'})
```
ThiefMaster, line 399: here too

ThiefMaster:
While testing this I realized that we don't really enforce uniqueness on rooms. This is fine IMHO (no harm there if admins want to do stupid things), but my original idea of making the `{number}` always doesn't seem so useful anymore. Maybe we can allow leaving the name format completely empty, but in that case make the `name` of the room mandatory?

ThiefMaster:
OK this looks like a mess to implement with the DB-side name generation and nobody asked for it so far so I think we better leave it as it is...

PR #5242, opened 2022-02-08
Add before_reminder_make_email to Reminders

vasantvohra:


CHANGES.rst
```diff
@@ -32,7 +32,8 @@ Internal Changes
 
 - Add ``category-sidebar`` template hook and blocks around category sidebar
   sections (:pr:`5237`, thanks :user:`omegak`)
-
```
ThiefMaster, line 35: Please keep the number of empty lines as they were before :)

indico/core/signals/event/reminder.py
```diff
@@ -0,0 +1,13 @@
...
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from indico.core.signals.event import _signals
+
+
+before_reminder_make_email = _signals.signal('before-reminder-make-email', '''
+Executed before reminder email is created. The parameters to create an email are passed.
```
ThiefMaster, line 12: "The parameters to create an email are passed." - in pretty much all the other signals we are clear about what params are passed, what the sender is, etc.

indico/modules/events/reminders/models/reminders.py
```diff
@@ -217,9 +218,10 @@ def send(self):
             attachments.append(MIMECalendar('event.ics', event_ical))
 
         for recipient in recipients:
-            email = make_email(
-                to_list=recipient, from_address=self.reply_to_address, template=email_tpl, attachments=attachments
-            )
+            params = dict(to_list=recipient, from_address=self.reply_to_address, template=email_tpl,
+                          attachments=attachments)
+            signals.event.reminder.before_reminder_make_email.send(params)
```
ThiefMaster, line 223: I don't think the signal is very useful in the way it's currently used. There's not even a reference to the context where this happens. Some suggestions:

- Use `self` as the `sender`
- Pass other data that you just want to provide (ie immutable data) as individual kwargs
- I guess you want some of the data to be mutable? You could either use the "updates" pattern like in another recent PR from you guys, or we could see if there's a better option.

Especially because of the last point, I recommend including details with this type of pull request on how you intend to use a new signal.

vasantvohra: Hello @ThiefMaster, I have updated the PR with your recommended changes, for now, I'd just like to pass the `html=True` in the `make_email` function.
Thanks

CHANGES.rst
ThiefMaster: Now there's an empty line between the individual changelog entries - but there should be two blank lines after all of them.

Issue #5228, opened 2022-01-29
Missing charset in iCalendar events attached to reminders


jhyeon:
**Describe the bug**
Indico 3.1 does not specify the `charset` parameter in an iCalendar event attached to reminder email.  More precisely, it seems to misname the parameter `encoding`.

 **To Reproduce**
1. Create an event
2. Add a reminder
3. Check the "Attach iCalendar file" option
4. Receive the reminder email
5. Some applications fail to handle the iCalendar event in the message

**Expected behavior**
The iCalendar event is expected to be displayed by the email application.

**Additional context**
[Section 3.1.4 of RFC5545](https://datatracker.ietf.org/doc/html/rfc5545#section-3.1.4) requires the `charset` Content-Type parameter in MIME transports.  See also [the errata](https://www.rfc-editor.org/errata/eid2038).  This rule was obeyed by previous releases of indico as the following MIME part shows:
```
--===============9109531323641687849==
Content-Type: text/calendar; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 8bit
Content-Disposition: attachment; filename="event.ics"

BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//CERN//INDICO//EN
```
This is no longer the case as of version 3.1:
```
--===============5843370041342135681==
Content-Type: text/calendar; filename="event.ics"; encoding="utf-8";
 method="REQUEST"
MIME-Version: 1.0

BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//CERN//INDICO//EN
```
Another parameter called `encoding` is set to `utf-8` instead.

PR #5250, opened 2022-02-12
Fix Content-Type param from encoding to charset per S3.1.4 RFC5545

imranyusuff:
Trivial change that addresses issue #5228.

### DIFF
--- a/indico/modules/events/ical.py
+++ b/indico/modules/events/ical.py
@@ -32,7 +32,8 @@ class MIMECalendar(MIMEBase):
 
     def __init__(self, filename: str, payload: str):
         message.Message.__init__(self, policy=compat32)
-        self.add_header('Content-Type', 'text/calendar', filename=filename, charset='utf-8', method='REQUEST')
+        self.add_header('Content-Type', 'text/calendar', charset='utf-8', method='REQUEST')
+        self.add_header('Content-Disposition', 'attachment', filename=filename)
         self['MIME-Version'] = '1.0'
         self.set_payload(payload)
 
--- a/indico/modules/events/ical.py
+++ b/indico/modules/events/ical.py
@@ -32,7 +32,7 @@ class MIMECalendar(MIMEBase):
 
     def __init__(self, filename: str, payload: str):
         message.Message.__init__(self, policy=compat32)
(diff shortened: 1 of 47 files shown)


imranyusuff:
Also I am thinking if the attachment filename can be moved to `Content-Disposition` instead.

javfg:
Thank you for the PR, @imranyusuff!

The change seems to be working properly, and you are right on the parameter being named wrong. Good catch. But I've double checked in Thunderbird and Outlook and it was also working before for me. Out of curiosity, can you tell me which client is failing for you?

> Also I am thinking if the attachment filename can be moved to `Content-Disposition` instead.

Sounds good, you can add go ahead and add that change too.

jhyeon:
> Out of curiosity, can you tell me which client is failing for you?

For me [Gnus](https://www.gnus.org/) failed.

@imranyusuff thank you!

imranyusuff:
> Thank you for the PR, @imranyusuff!
> 
> The change seems to be working properly, and you are right on the parameter being named wrong. Good catch. But I've double checked in Thunderbird and Outlook and it was also working before for me. Out of curiosity, can you tell me which client is failing for you?
> 
> > Also I am thinking if the attachment filename can be moved to `Content-Disposition` instead.
> 
> Sounds good, you can add go ahead and add that change too.

I have read carefully RFC 2183 which describe the `Content-Disposition` field. The intent is to tell the browser (or relevant client) whether to display the part as either inline or as an attachment. So need to be clear here what indico intend to present this attached vCalendar file.

Also, I also trying to find out if `filename` is actually supposed to be inside `Content-Type` field in the first place?

ThiefMaster:
Please keep merge commits out of your PR, and make sure to keep it based on 3.1.x, and not master.

imranyusuff:
Oh no. Something wrong here. I think I messed up my branch.

ThiefMaster:
`git reset --hard 3.1.x` and then `git cherry-pick ...` your actual commits.

And in case you are using some weird IDE feature to manage Git, it's best to not do that but use the normal `git` command line tool.

imranyusuff:
Created new PR #5258 and closing this one.

PR #5251, opened 2022-02-13
Fix pytest:date_time_test.py failing when timezone is east of UTC+1

imranyusuff:
`pytest` fails on `date_time_test.py` when time zone settings is east of Zurich (UTC+1).

This is because the function `datetime` will always return time at the current time zone, and converting this `datetime` time to Zurich time zone via `datetime.astimezone` will result in time to move back to previous day if the current time zone is anywhere to the east.

Fixed this by enclosing `datetime` with `as_utc` to make sure the time is in UTC before calling `datetime.astimezone`.

Futher `pytest` with this finally passed all tests.


ThiefMaster (approved):

ThiefMaster:
Thanks! FYI I rebased this to 3.1.x since it makes sense to apply this kind of bug fix to the current stable branch, not just the development branch.

PR #5252, opened 2022-02-14
Fix error when deleting time on the date input

duartegalvao:
This PR fixes a JS error that occurs when attempting to delete the time value on the new React date input. The error was caused by a missing null check.

### DIFF
--- a/indico/modules/events/registration/client/js/form/fields/DateInput.jsx
+++ b/indico/modules/events/registration/client/js/form/fields/DateInput.jsx
@@ -20,7 +20,7 @@ import {toMoment} from 'indico/utils/date';
 import '../../../styles/regform.module.scss';
 
 function DateInputComponent({value, onChange, disabled, required, dateFormat, timeFormat}) {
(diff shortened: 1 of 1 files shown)


indico/modules/events/registration/client/js/form/fields/DateInput.jsx
```diff
@@ -28,7 +28,8 @@ function DateInputComponent({value, onChange, disabled, required, dateFormat, ti
     const timeString = timeFormat ? timeValue : '00:00:00';
     onChange(`${dateString}T${timeString}`);
   };
-  const handleTimeChange = newTime => onChange(`${dateValue}T${newTime.format('HH:mm:00')}`);
+  const handleTimeChange = newTime =>
+    onChange(`${dateValue}T${newTime ? newTime.format('HH:mm:00') : ''}`);
```
ThiefMaster, line 32:
```suggestion
    onChange(newTime ? `${dateValue}T${newTime.format('HH:mm:00')}` : dateValue);
```

Untested, but I think that's more readable..

PR #5253, opened 2022-02-14
Partial regform editing

tomasr8:
This allows updating a registration with only the fields that have actually changed
as opposed to resending the whole form.

In preparation for this, the usage of `defaultValue` has to be replaced with `initialValues`.
The `defaultValue` can't be completely eliminated as that would break form setup which may
add fields after the initial load. 

Progress:
- [x] Switching from `defaultValue` to `initialValues`
- [x] Partial updates when modifying a registration


indico/modules/events/registration/client/js/form/fields/SingleChoiceInput.jsx
```diff
@@ -355,12 +356,13 @@ export default function SingleChoiceInput({
       required={isRequired}
       isRequired={isRequired}
       disabled={disabled}
-      defaultValue={defaultItem ? {[defaultItem]: 1} : {}}
+      defaultValue={setupMode ? defaultValue : undefined}
```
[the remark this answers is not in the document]
tomasr8: From my understanding yes, because the initialValues only cover the fields that are already in the regform when you load it. If you're in the setup and add a new field it won't have an initial value.. hmm actually now that I'm thinking about it I think we could just rerender the whole form with the new initialValues..

indico/modules/events/registration/fields/base.py
```diff
@@ -137,7 +137,8 @@ def unprocess_field_data(cls, versioned_data, unversioned_data):
 
     @property
     def view_data(self):
-        return self.unprocess_field_data(self.form_item.versioned_data, self.form_item.data)
+        return (self.unprocess_field_data(self.form_item.versioned_data, self.form_item.data) |
+                {'defaultValue': self.default_value})
```
ThiefMaster, line 141: I think we camelize the data later so we could still use snake case here, since that's what `unprocess_field_data` returns as well.

indico/modules/events/registration/fields/choices.py
```diff
@@ -473,6 +473,18 @@ class AccommodationField(RegistrationFormBillableItemsField):
     mm_field_class = fields.Nested
     mm_field_kwargs = {'nested': AccommodationSchema}
 
+    @property
+    def default_value(self):
+        versioned_data = self.form_item.versioned_data
+        no_accommodation_option = next(
+            (c for c in versioned_data['choices'] if c['is_no_accommodation'] and c['is_enabled']), None)
+        return {
```
ThiefMaster, line 481: same thing here, we can use snake case since it'll be converted anyway. otherwise we have a mix between snake_case and camelCase in the `view_data` dict until the conversion happens..

indico/modules/events/registration/util_test.py
```diff
@@ -418,3 +422,181 @@ def test_get_registered_event_persons(dummy_event, dummy_user, dummy_regform):
...
+    assert not reg.data_by_field[boolean_field.id].data
+    assert reg.data_by_field[multi_choice_field.id].data == {}
+    # Assert that the manager field gets properly set with management=True
+    assert reg.data_by_field[checkbox_field.id].data
+
+
+def test_modify_registration(monkeypatch, dummy_event, dummy_user, dummy_regform):
+    class _DummySesssion:
+        user = dummy_user
+    monkeypatch.setattr('indico.modules.events.registration.util.session', _DummySesssion)
```
ThiefMaster, line 512: You could also use the `request_context` fixture (using `@pytest.mark.usefixtures('request_context')`) and then call `session.set_session_user(dummy_user)`

indico/modules/events/registration/util_test.py
```diff
@@ -418,3 +422,181 @@ def test_get_registered_event_persons(dummy_event, dummy_user, dummy_regform):
...
+        'input_type': 'bool', 'default_value': False, 'title': 'Yes/No'
+    })
+    db.session.add(boolean_field)
+
+    multi_choice_field = RegistrationFormField(parent_id=section.id, registration_form=dummy_regform)
+    _fill_form_field_with_data(multi_choice_field, {
+        'input_type': 'multi_choice', 'with_extra_slots': False, 'title': 'Multi Choice',
+        'choices': [
+            {'caption': 'A', 'id': 'new:AbntJPo-cMPo8SKgcsjAr', 'is_enabled': True},
+            {'caption': 'B', 'id': 'new:H0VRkvScH-NXeL837UGNu', 'is_enabled': True},
```
ThiefMaster, line 531: how about `new:test1` and `new:test2` or something like that? ;)

indico/modules/events/registration/controllers/__init__.py
```diff
@@ -40,8 +41,18 @@ def _get_file_data(self):
                 for r in self.registration.data
                 if r.field_data.field.field_impl.is_file_field and r.storage_file_id is not None}
 
+    def _get_optional_fields(self):
+        """Get fields for which we already have a value and thus are not required."""
+
```
ThiefMaster, line 46: only classes have a blank line after their docstring

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -273,8 +273,9 @@ class RHRegistrationCreate(RHManageRegFormBase):
     def _get_user_data(self, user):
         if user is None:
             return {}
-        user_data = {t.name: getattr(user, t.name, None) for t in PersonalDataType}
-        user_data['title'] = get_title_uuid(self.regform, user_data['title'])
+        user_data = {t.name: getattr(user, t.name, t.default)
+                     if user else t.default for t in PersonalDataType}
```
ThiefMaster, line 277: I really don't like the wrapping  here :/

```python
user_data = {t.name: getattr(user, t.name, t.default) if user else t.default
                     for t in PersonalDataType}
```

or since it fits in one line:

```python
user_data = {t.name: getattr(user, t.name, t.default) if user else t.default for t in PersonalDataType}
```

indico/modules/events/registration/fields/simple.py
```diff
@@ -202,14 +206,25 @@ def view_data(self):
         return dict(super().view_data, has_time=has_time)
 
 
+class BooleanFieldSetupSchema(LimitedPlacesBillableFieldDataSchema):
+    default_value = fields.String(load_default='', validate=validate.OneOf(['', 'yes', 'no']))
+
+    @pre_load()
```
ThiefMaster, line 212:
```suggestion
    @pre_load
```

ThiefMaster (approved):

PR #5255, opened 2022-02-15
Add signals for rendering attachments

vtran99:
Add signals to intercept the attachments retrieving and the rendering of the template.


OmeGak:
Thanks for the suggestion to use `LinkType` instead, @ThiefMaster. It will require further adjustments in our implementation but it will allow for a more consistent implementation. We no longer have need for these signals, so feel free to close this PR.

PR #5257, opened 2022-02-15
Docs: mention auth for downloading attachments

avivace:
Just a quick note to mention how restricted attachments can be accessed using the API Token Authentication

### DIFF
--- a/docs/source/http-api/access.rst
+++ b/docs/source/http-api/access.rst
@@ -80,9 +80,9 @@ API tokens can have one or more of these scopes:
 
     main()
 
-The ``everything`` scopes are special because they can be used with *any* Indico endpoint, i.e. they are
-not restricted to official APIs. This has the advantage that even Indico actions which do not have a corresponding
-API can be scripted.
+The ``everything`` scopes are special because they can be used with *any* Indico endpoint (including file attachments),
+i.e. they are not restricted to official APIs. This has the advantage that even Indico actions which do not have a
+corresponding API can be scripted.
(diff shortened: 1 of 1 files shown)


ThiefMaster:
I think it's pretty logical that an "everything" token can access attachments as well. I don't mind being more explicitly here, but I disagree with putting this up there, since it'd kind of imply that any token would work...

How about adding it to this paragraph?

> The ``everything`` scopes are special because they can be used with *any* Indico endpoint, i.e. they are
not restricted to official APIs. This has the advantage that even Indico actions which do not have a corresponding
API can be scripted.

something like "They can be used to download restricted materials/attachments as well." maybe

avivace:
> I think it's pretty logical that an "everything" token can access attachments as well. I don't mind being more explicit about this, but I disagree with putting this up there, since it'd kind of imply that any token would work...
> 
> How about adding it to this paragraph?
> 
> > The `everything` scopes are special because they can be used with _any_ Indico endpoint, i.e. they are
> > not restricted to official APIs. This has the advantage that even Indico actions which do not have a corresponding
> > API can be scripted.
> 
> something like "They can be used to download restricted materials/attachments as well." maybe

Thanks for the feedback, I see your point. Something like 

> The everything scopes are special because they can be used with any Indico endpoint (including, e.g., exposed file attachments) <...>

works for you?

ThiefMaster:
Sure. I'd just drop the "exposed", that sounds a bit unnecessary there... Also maybe rebase to the 3.1.x branch and change the target branch for the PR here.

PR #5262, opened 2022-02-24
Fix a bug preventing cloning invited abstract on customize

plourenco:
Can be reproduced by customizing the list of abstracts and trying to clone them into an invited abstract. Even if an email template is set up, it will incorrectly prevent it.

Introduced in #5217


ThiefMaster:
3.1.x? ;)

PR #5268, opened 2022-02-28
Allow speaker-only menu entries

### LINKED ISSUE #5261: Restrict custom pages to speakers only
We need to provide information like zoom login details, speaker protocols, etc. to our speakers. This information evolves over the course of an event, so we want to provide it on a web page - which we can continuously update - rather than in a series of emails.
We don't want anyone else to access this information; only the speakers.

What we'd like is to be able to put this information on custom pages and restrict access to 'speakers only'.
For the menu there are two options
1. show the menu link to all users, but when a non-authorised user clicks the link then display a page with message saying something like "the content of this page is for speakers only (and maybe a suggestion to login if not already logged in). if you are a speaker and think you should have access please contact <event contact>"
2. show the menu link only to logged in users who are also confirmed speakers
(this might cause some users to get in touch asking "where has the menu gone?" if they forget to log in first)
I think there are usability pros and cons to both options (and maybe some others options I haven't considered)

We have tried series of emails but speakers end up confused if they accidentally don't check the latest one.
We have tried using custom pages, hiding them in the menu, and sending the link by email to speakers. Problem is they keep losing the link and have to ask for a reminder, often in the middle of an event when everyone's busy/stressed. Plus it's not secure - non-speakers could potentially get hold of the link and read and use the content.
What we're doing at the moment is publishing this content on our Confluence wiki instead and restricting it to the event speakers, which involves manually creating a corresponding group of speakers on Confluence and manually keeping it synchronised with the list of speakers on Indico. This also suffers from the problem described above: we have to email them the link and they can't easily find it again when they need it. Plus it means we're storing content across multiple system when it feels like it should be kept together.

duartegalvao:
Closes #5261

This PR adds the ability to create speaker-only menu entries for conferences.

image

### DIFF
--- a/indico/migrations/versions/20220225_1335_ef7a8b2e6737_add_visibility_options_to_menu_entries.py
+++ b/indico/migrations/versions/20220225_1335_ef7a8b2e6737_add_visibility_options_to_menu_entries.py
@@ -1,0 +1,44 @@
+"""Add visibility options to menu entries
+
+Revision ID: ef7a8b2e6737
+Revises: 3dafee32ba7d
+Create Date: 2022-02-25 13:35:15.089546
+"""
+
+from enum import Enum
+
+import sqlalchemy as sa
+from alembic import op
+
+from indico.core.db.sqlalchemy import PyIntEnum
+
+
+# revision identifiers, used by Alembic.
+revision = 'ef7a8b2e6737'
+down_revision = '82fb6c6ac6db'
+branch_labels = None
+depends_on = None
+
+
+class _MenuEntryAccess(int, Enum):
+    everyone = 1
+    registered_participants = 2
+    speakers = 3
+
+
+def upgrade():
+    op.add_column('menu_entries',
+                  sa.Column('access', PyIntEnum(_MenuEntryAccess), server_default='1', nullable=False),
+                  schema='events')
+    op.alter_column('menu_entries', 'access', server_default=None, schema='events')
+    op.execute('UPDATE events.menu_entries SET access = 2 WHERE registered_only')
+    op.drop_column('menu_entries', 'registered_only', schema='events')
+
+
+def downgrade():
+    op.add_column('menu_entries',
+                  sa.Column('registered_only', sa.Boolean(), server_default='false', nullable=False),
+                  schema='events')
+    op.alter_column('menu_entries', 'registered_only', server_default=None, schema='events')
+    op.execute('UPDATE events.menu_entries SET registered_only = true WHERE access > 1')
+    op.drop_column('menu_entries', 'access', schema='events')
--- a/indico/modules/events/layout/models/menu.py
+++ b/indico/modules/events/layout/models/menu.py
@@ -36,6 +36,13 @@ class MenuEntryType(RichIntEnum):
     page = 5
 
 
+class MenuEntryAccess(RichIntEnum):
+    __titles__ = [None, _('Everyone'), _('Registered participants'), _('Speakers')]
+    everyone = 1
+    registered_participants = 2
+    speakers = 3
+
+
 class MenuEntryMixin:
     def __init__(self, **kwargs):
         event = kwargs.pop('event', kwargs.get('event'))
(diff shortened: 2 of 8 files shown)


indico/modules/events/layout/models/menu.py
```diff
@@ -121,12 +129,43 @@ def default_data(self):
...
+        from indico.modules.events.contributions.models.persons import ContributionPersonLink, SubContributionPersonLink
+        from indico.modules.events.contributions.models.subcontributions import SubContribution
+        from indico.modules.events.models.persons import EventPerson
+
+        if not self.event_ref.can_manage(session.user):
+            if (self.visibility == MenuEntryVisibility.registered_users
+                    and not self.event_ref.is_user_registered(session.user)):
+                return False
+            elif (self.visibility == MenuEntryVisibility.speakers
+                    and not EventPerson.query
```
ThiefMaster, line 144: I'd really avoid putting a query in an if statement like this... better to add some util functions for this

duartegalvao: What's the best place to implement the util? Maybe a `is_speaker(self, user)` method in `EventPerson`?

ThiefMaster: It's action on a user, not a person... I'd just put it as a local `_is_speaker(event, user)` method in this file (or util.py in the module, then without the underscore)

indico/modules/events/layout/models/menu.py
```diff
@@ -121,12 +153,22 @@ def default_data(self):
...
+    @property
+    def is_accessible(self):
+        if not self.event_ref.can_manage(session.user):
+            if (self.visibility == MenuEntryVisibility.registered_users
+                    and not self.event_ref.is_user_registered(session.user)):
+                return False
+            elif (self.visibility == MenuEntryVisibility.speakers
+                    and not _is_speaker(self.event_ref, session.user)):
+                return False
+        return True
```
ThiefMaster, line 165:
```py
if can_manage:
    return True
elif visibility == registered:
    return ...is_user_registered..
elif visibility == speakers:
    return ...is_spealer..
```

more readable IMO

indico/modules/events/layout/models/menu.py
```diff
@@ -27,6 +28,30 @@ def _get_next_position(context):
     return (res[0] or 0) + 1
 
 
+def _is_speaker(event, user):
```
ThiefMaster, line 31: I just saw that we have `is_user_registered` as a method in the `Event` model... so maybe moving this there as a `is_user_speaker` method isn't a bad idea :)

indico/modules/events/layout/models/menu.py
```diff
@@ -121,12 +153,21 @@ def default_data(self):
     def is_orphaned(self):
         return bool(self.name and self.default_data is None)
 
+    @property
+    def is_accessible(self):
+        if self.event_ref.can_manage(session.user):
```
ThiefMaster, line 158: let's short-circuit this check if there's no restriction:


```py
if self.visibility == MenuEntryVisibility.everyone:
    return True
```

indico/modules/events/layout/models/menu.py
```diff
@@ -121,12 +153,21 @@ def default_data(self):
     def is_orphaned(self):
         return bool(self.name and self.default_data is None)
 
+    @property
+    def is_accessible(self):
```
ThiefMaster, line 157: I think this should be a method that takes a `user` arg - it's a bit strange to use the session user for such things. and maybe we can use the more common `can_access` name?

indico/migrations/versions/20220225_1335_ef7a8b2e6737_add_visibility_options_to_menu_entries.py
```diff
@@ -0,0 +1,44 @@
...
+
+class _MenuEntryVisibility(int, Enum):
+    everyone = 1
+    registered_users = 2
+    speakers = 3
+
+
+def upgrade():
+    op.add_column('menu_entries',
+                  sa.Column('visibility', PyIntEnum(_MenuEntryVisibility), server_default='1', nullable=False),
```
ThiefMaster, line 31: hm... `visibility` or `access`...? i don't have a strong opinion for either option though

indico/modules/events/layout/models/menu.py
duartegalvao: Should `is_visible` also take a `user` arg?

ThiefMaster: Nevermind, I didn't notice that we already have existing code using it as well... I just didn't want to introduce it in new code, but I think for the sake of ocnsistency it's fine to leave it as is then.

ThiefMaster (approved):

PR #5270, opened 2022-03-03
Add registration visibility override

duartegalvao:
This PR adds a manual override for the visibility of each registration, set by the event's manager.

This PR includes:
- A new visibility override attribute on registrations.
- Unit tests for the new attribute.
- Hide participant button on registration summary: 
screenshot

### DIFF
--- a/indico/modules/events/registration/client/js/form_submission/RegistrationFormSubmission.jsx
+++ b/indico/modules/events/registration/client/js/form_submission/RegistrationFormSubmission.jsx
@@ -53,14 +53,8 @@ function EmailNotification() {
   );
 }
 
-function ConsentToPublish({
-  publishToParticipants,
-  publishToPublic,
-  maximumConsentToPublish,
-  management,
-}) {
+function ConsentToPublish({publishToParticipants, publishToPublic}) {
   const isWarning = publishToParticipants === 'show_all' && publishToPublic === 'show_all';
-
   return (
     <Message info={!isWarning} warning={isWarning} style={{marginTop: 25}}>
       <Message.Header>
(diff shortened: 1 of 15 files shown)


indico/modules/events/registration/models/registrations.py
```diff
@@ -229,9 +229,15 @@ class Registration(db.Model):
...
     consent_to_publish = db.Column(
-        PyIntEnum(PublishConsentType),
+        PyIntEnum(RegistrationVisibility),
         nullable=False,
-        default=PublishConsentType.nobody
+        default=RegistrationVisibility.nobody
+    )
+    #: Management-set override for visibility
+    visibility_override = db.Column(
+        PyIntEnum(RegistrationVisibility),
```
ThiefMaster, line 238: I wouldn't make this an enum but rather a simple `participant_hidden` boolean. By making this an enum you're adding a huge amount of extra complexity which would otherwise be as simple as adding a check for `and not participant_hidden` to the queries etc...

The point of the feature is simply that if a participant asks the organizers to not be shown in the participant list, or the organizer wishes to hide someone (for whatever reason), they have the ability to do so, even in an event that's not consent-based.

indico/modules/events/registration/client/js/form_submission/RegistrationFormSubmission.jsx
```diff
@@ -139,13 +158,16 @@ export default function RegistrationFormSubmission() {
           ))}
           {renderPluginComponents('regformAfterSections')}
           <div>
-            {isManagement && <EmailNotification />}
+            {isManagement && (
+              <>
+                <EmailNotification />
+                <VisibilityOverride />
```
ThiefMaster, line 164: I would not make this part of the "modify registration" dialog but rather do it similar to what we have for changing consent when viewing your own registration. Combined with my other comment, this would simply be a "hide from participant list" toggle.

indico/migrations/versions/20220301_1714_57696d76f9b0_add_registration_visibility_override.py
```diff
@@ -0,0 +1,26 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '57696d76f9b0'
+down_revision = 'ef7a8b2e6737'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('registrations',
```
ThiefMaster, line 20: you need an `op.alter_column()` to remove the `server_default` since it's not part of the model (the `default=...` there is python-side)

ThiefMaster (approved):

PR #5271, opened 2022-03-03
Add plugin extension points to new react registration form

ThiefMaster:
Still need to come up with a nicer API to register plugin components; the global variable was just a dirty way to test things.


plourenco (approved):
👍

PR #5273, opened 2022-03-08
Add upload fixed image in badge designer

SegiNyn:
This adds an option to upload images specific to the badge from the badge designer. 

The image can be added from the 'insert elements' dropdown list and is called `Fixed image`

Multiple Fixed images can be added to the badge.

### DIFF
--- a/indico/modules/designer/client/js/index.js
+++ b/indico/modules/designer/client/js/index.js
@@ -38,6 +38,10 @@ import {$T} from 'indico/utils/i18n';
 
   let imageTypes = null;
 
+  let images = null;
+
+  let lastSelectedItem = null;
+
   function isImage(type) {
     return imageTypes.includes(type);
   }
@@ -113,7 +117,13 @@ import {$T} from 'indico/utils/i18n';
           zIndex: this.zIndex,
         })
         .attr('data-type', this.type)
-        .text(this.type === 'fixed' ? this.text : itemTitles[this.type]);
+        .text(
+          this.type === 'fixed'
+            ? this.text
+            : this.type === 'fixed_image'
+            ? ''
+            : itemTitles[this.type]
+        );
       return html;
     }.bind(item);
 
@@ -232,10 +242,13 @@ import {$T} from 'indico/utils/i18n';
 
     $('.element-tools').removeClass('hidden');
     $('.second-row').removeClass('disappear');
+    $('.selection-text').html(itemTitles[item.type]);
 
-    $('.selection-text').html(
-      item.type === 'fixed' ? $T.gettext('Fixed text') : itemTitles[item.type]
-    );
+    if (item.type === 'fixed_image') {
+      $('.js-remove-img').prop('disabled', !item.image_id);
+    }
+    const $imageFile = $('#imageFile');
+    $('.js-upload-img').prop('disabled', !$imageFile.val());
 
     deselectItem($('.designer-item.selected'));
     item.selected = true;
@@ -262,21 +275,28 @@ import {$T} from 'indico/utils/i18n';
     var $fixedTextField = $('#fixed-text-field');
     const $itemHeightField = $('#subtool-element-height');
     var $fontTools = $('.font-tools');
+    var $fixedImageTool = $('#fixed-image-tool');
 
     if (item.type === 'fixed') {
       $fontTools.show();
       $fixedTextField.closest('.tool').show();
       $fixedTextField.val(item.text);
       $itemHeightField.hide();
+      $fixedImageTool.hide();
     } else if (isImage(item.type)) {
       $fontTools.hide();
       $fixedTextField.closest('.tool').hide();
       $itemHeightField.show();
+      $fixedImageTool.hide();
+      if (item.type === 'fixed_image') {
+        $fixedImageTool.show();
+      }
     } else {
       $fontTools.show();
       $fixedTextField.closest('.tool').hide();
       $fixedTextField.val('');
       $itemHeightField.hide();
+      $fixedImageTool.hide();
     }
   }
 
@@ -354,6 +374,10 @@ import {$T} from 'indico/utils/i18n';
         zIndex: item.zIndex,
       });
       newDiv.append(item.toHTML());
+      if (item.type === 'fixed_image' && item.image_id) {
+        const $imageElement = createFixedImageElement(item, images[item.image_id]);
+        $imageElement.appendTo(newDiv.find('.designer-item'));
+      }
       if (item.selected && !isBackside) {
         selectItem(newDiv.find('.designer-item'));
       }
@@ -553,6 +577,11 @@ import {$T} from 'indico/utils/i18n';
     }[attribute]());
 
     div.replaceWith(selectedItem.toHTML());
+    if (selectedItem.type === 'fixed_image') {
+      $('.designer-item.selected').append(
+        createFixedImageElement(selectedItem, images[selectedItem.image_id])
+      );
+    }
   }
 
   function save(template) {
@@ -706,6 +735,47 @@ import {$T} from 'indico/utils/i18n';
     }
   }
 
+  function createFixedImageElement(item, imageUrl) {
+    const $imageElement = $('<img/>');
+    $imageElement
+      .attr({
+        src: imageUrl,
+      })
+      .css({
+        position: 'absolute',
+        left: 0,
+        top: 0,
+        height: item.height + 'px',
+        width: item.width + 'px',
+      });
+    return $imageElement;
+  }
+
+  function displayFixedImage(imageUrl) {
+    const $div = $('.ui-draggable[data-id=' + lastSelectedItem.id + '] div');
+    $div.text('');
+    $div.find('img').remove();
+    const $imageElement = createFixedImageElement(lastSelectedItem, imageUrl);
+    $imageElement.appendTo($div);
+    const $imageFile = $('#imageFile');
+    $imageFile.val('');
+    $imageFile
+      .next('label')
+      .addClass('i-button')
+      .text($T.gettext('Choose a file'));
+    $('.js-upload-img').prop('disabled', $imageFile.val());
+    $('.js-remove-img').prop('disabled', !lastSelectedItem.image_id);
+  }
+
+  function removeFixedImage(item) {
+    if (item.image_id) {
+      item.image_id = null;
+      $('.ui-draggable[data-id=' + item.id + '] div').text($T.gettext('Fixed Image'));
+      $('.ui-draggable[data-id=' + item.id + '] img').remove();
+      $('.js-remove-img').prop('disabled', !item.image_id);
+    }
+  }
+
   function unescapeHTML(str) {
     // taken from Prototype
     return str
@@ -836,6 +906,8 @@ import {$T} from 'indico/utils/i18n';
 
     imageTypes = imageTypes_;
 
+    images = {...template.images, ...backsideTemplate.images};
+
     // Item class
     $(document).ready(function() {
       var removeBackgroundQtip = $('.js-remove-bg').qtip();
@@ -849,7 +921,20 @@ import {$T} from 'indico/utils/i18n';
             .text(this.files[0].name);
         }
         $('.js-upload-bg')
-          .attr('disabled', !$this.val())
+          .prop('disabled', !$this.val())
+          .toggleClass('highlight', !!$this.val());
+      });
+
+      $('#img-form input[type="file"]').on('change', function() {
+        var $this = $(this);
+        if (this.files) {
+          $this
+            .next('label')
+            .removeClass('i-button')
+            .text(this.files[0].name);
+        }
+        $('.js-upload-img')
+          .prop('disabled', !$this.val())
           .toggleClass('highlight', !!$this.val());
       });
 
@@ -865,6 +950,14 @@ import {$T} from 'indico/utils/i18n';
       $('.js-upload-bg').click(function() {
         $('.js-toggle-side.front').click();
         $('#bg-form').submit();
+        return false;
+      });
+
+      $('.js-upload-img').click(function() {
+        lastSelectedItem = getSelectedItemData();
+        if (lastSelectedItem.type === 'fixed_image') {
+          $('#img-form').submit();
+        }
         return false;
       });
 
@@ -889,6 +982,21 @@ import {$T} from 'indico/utils/i18n';
         },
       });
 
+      $('#img-form').ajaxForm({
+        dataType: 'json',
+        iframe: false,
+        success: function(data) {
+          if (data.error) {
+            alertPopup(data.error, $T('Error'));
+            return;
+          }
+          if (lastSelectedItem.type === 'fixed_image') {
+            lastSelectedItem.image_id = data.image_id;
+            displayFixedImage(data.image_url);
+          }
+        },
+      });
+
       $('input[name=bg-position]').change(function(e) {
         e.preventDefault();
         var newPosition = $(this).val();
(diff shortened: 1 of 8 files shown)


indico/modules/designer/controllers.py
```diff
@@ -324,10 +325,11 @@ def _process(self):
             return jsonify(error='File format not accepted!')
         content_type = 'image/' + image_type
         image = DesignerImageFile(template=self.template, filename=filename, content_type=content_type)
-        self.template.background_image = image
+        if request.form.get('for_background') == 'True':
```
plourenco, line 328: Instead of checking for an input value `for_background`, I would recommend changing a bit the logic here to one of the following:
- We could rename this endpoint to `RHUploadImage` and have another dedicated one for setting the background image `RHSetBackgroundImage`
- Alternatively, we could add a query param `background=True/False` when we intend to set a background. The form action would be: 
```
action="{{ url_for('designer.upload_image', template, background=True) }}"
```

plourenco (commented):
Thanks! There were a few ESLint problems, but not worth reviewing those since we can refactor this file in a single commit before merging.

plourenco:
@ThiefMaster since there's currently more than one fixed element, what do you think about having both the `FixedText` and `FixedImage` as placeholders in the designer? 
This could prevent client-side conditionals such as [this](https://github.com/indico/indico/pull/5273/files#diff-b3ff04534fe852620feb9c66bb4e46b192e2e62a8a2d37d067893596755ddf21R118).

indico/modules/designer/controllers.py
```diff
@@ -281,7 +281,8 @@ def _process_GET(self):
     def _process_POST(self):
         data = dict({'background_position': 'stretch', 'items': []}, **request.json['template'])
```
plourenco, line 282:
```suggestion
    @use_kwargs({'background': fields.Bool(load_default=False)})
    def _process_POST(self):
        data = dict({'background_position': 'stretch', 'items': []}, **request.json['template'])
```

indico/modules/designer/controllers.py
```diff
@@ -324,10 +325,11 @@ def _process(self):
             return jsonify(error='File format not accepted!')
         content_type = 'image/' + image_type
         image = DesignerImageFile(template=self.template, filename=filename, content_type=content_type)
-        self.template.background_image = image
+        if request.args['background'] == '1':
+            self.template.background_image = image
```
plourenco, line 329:
```suggestion
        if background:
            self.template.background_image = image
```

indico/modules/designer/controllers.py
SegiNyn: Okay, thank you

indico/modules/designer/client/js/index.js
```diff
@@ -113,7 +115,13 @@ import {$T} from 'indico/utils/i18n';
...
         })
         .attr('data-type', this.type)
-        .text(this.type === 'fixed' ? this.text : itemTitles[this.type]);
+        .text(
+          this.type === 'fixed'
+            ? this.text
+            : this.type === 'fixed_image' && item.image_id
+            ? ''
+            : itemTitles[this.type]
+        );
```
plourenco, line 124: I think we can simplify this now that we are using placeholders.

```suggestion
        .text(itemTitles[this.type]);
```

plourenco (approved):

indico/modules/designer/client/js/index.js
SegiNyn: I kept this condition because the fixed text can take a user input, making the inputted text different from the item title, also when an image is uploaded for the fixed_image, it should only display the image and not still show the title. So the title should only be used when there is no image.

plourenco: Right, never mind, for some reason I assumed this was referring to the selected text.

SegiNyn:
Hello, please can we merge this PR if it looks good, so we can include it in our next release

ThiefMaster:
Since you need to backport it anyway, I guess it'd be fine for you if we merged it into master and not 3.1.x?

SegiNyn:
Yes, it'll be fine

indico/modules/designer/templates/template.html
```diff
@@ -118,7 +118,7 @@
             <div class="tool js-hide-on-flip">
                 <i class="icon-image" title="{% trans %}Template background{% endtrans %}"></i>
                 <form id="bg-form" class="content flexrow f-a-center"
-                      action="{{ url_for('designer.upload_image', template) }}"
+                      action="{{ url_for('designer.upload_image', template, background=True) }}"
```
ThiefMaster, line 121:
```suggestion
                      action="{{ url_for('designer.upload_image', template, background=true) }}"
```

indico/modules/designer/templates/template.html
```diff
@@ -271,6 +271,27 @@
                     <label for="fixed-text-field">{% trans %}Text:{% endtrans %}</label>
                     <input type="text" id="fixed-text-field">
                 </div>
+                <div id="fixed-image-tool" class="tool js-hide-on-flip">
+                    <i class="icon-image" title="{% trans %}Fixed Image{% endtrans %}"></i>
+                    <form id="img-form" class="content flexrow f-a-center"
+                          action="{{ url_for('designer.upload_image', template, background=False) }}"
```
ThiefMaster, line 277:
```suggestion
                          action="{{ url_for('designer.upload_image', template, background=false) }}"
```

ThiefMaster:
Unless I'm blind we never check that the `image_id` when saving the badge actually belongs to that particular template... so you could actually "steal" an image from another event by simply sending a different ID...

SegiNyn:
Yes you are correct, there is not check that the image_id belongs to the template. I can add a check for it.

indico/modules/designer/controllers.py
```diff
@@ -281,9 +296,15 @@ def _process_GET(self):
     def _process_POST(self):
         data = dict({'background_position': 'stretch', 'items': []}, **request.json['template'])
         self.validate_json(TEMPLATE_DATA_JSON_SCHEMA, data)
-        invalid_placeholders = {x['type'] for x in data['items']} - set(get_placeholder_options()) - {'fixed'}
+        invalid_placeholders = {x['type'] for x in data['items']} - set(get_placeholder_options())
         if invalid_placeholders:
             raise UserValueError('Invalid item types: {}'.format(', '.join(invalid_placeholders)))
+        imageItems = [item for item in data['items'] if item['type'] == 'fixed_image']
```
ThiefMaster, line 302: we do not use camelCase in python code...

indico/modules/designer/controllers.py
```diff
@@ -177,8 +179,21 @@ def _process_args(self):
...
+        for i, image_item in image_items:
+            if image_item.get('image_id'):
+                old_image = DesignerImageFile.get(image_item.get('image_id'))
+                new_image = DesignerImageFile(filename=old_image.filename, content_type=old_image.content_type,
+                                              template=new_template)
+                with old_image.open() as f:
+                    new_image.save(f)
+
+                data['items'][i]['image_id'] = new_image.id
+                data['items'][i]['image_url'] = new_image.download_url
```
ThiefMaster, line 195: Why are we saving a URL in the `data`? This should be generated on the fly...

indico/modules/designer/controllers.py
```diff
@@ -177,8 +179,21 @@ def _process_args(self):
...
         title = f'{self.template.title} (copy)'
-        new_template = DesignerTemplate(title=title, type=self.template.type, data=self.template.data,
-                                        **self.target_dict)
+        new_template = DesignerTemplate(title=title, type=self.template.type, **self.target_dict)
+
+        data = self.template.data
+        image_items = [(i, item) for i, item in enumerate(data['items']) if item['type'] == 'fixed_image']
+        for i, image_item in image_items:
+            if image_item.get('image_id'):
+                old_image = DesignerImageFile.get(image_item.get('image_id'))
```
ThiefMaster, line 188: No need for `.get()`, you know the key exists here...

indico/modules/designer/controllers.py
```diff
@@ -177,8 +179,21 @@ def _process_args(self):
...
     def _process(self):
         title = f'{self.template.title} (copy)'
-        new_template = DesignerTemplate(title=title, type=self.template.type, data=self.template.data,
-                                        **self.target_dict)
+        new_template = DesignerTemplate(title=title, type=self.template.type, **self.target_dict)
+
+        data = self.template.data
+        image_items = [(i, item) for i, item in enumerate(data['items']) if item['type'] == 'fixed_image']
+        for i, image_item in image_items:
+            if image_item.get('image_id'):
```
ThiefMaster, line 187: When would there be a `fixed_image` item without an `image_id`? If this is not expected to happen, no need to check for it - just let it fail if there's garbage data in the DB.

indico/modules/designer/controllers.py
```diff
@@ -177,8 +179,21 @@ def _process_args(self):
 
     def _process(self):
         title = f'{self.template.title} (copy)'
-        new_template = DesignerTemplate(title=title, type=self.template.type, data=self.template.data,
-                                        **self.target_dict)
+        new_template = DesignerTemplate(title=title, type=self.template.type, **self.target_dict)
+
+        data = self.template.data
+        image_items = [(i, item) for i, item in enumerate(data['items']) if item['type'] == 'fixed_image']
```
ThiefMaster, line 185: I don't think the `enumerate` is needed, there's no need to use an index. When you iterate over a list of items and mutate an item, it will change the item in the list.

indico/modules/designer/controllers.py
```diff
@@ -177,8 +179,21 @@ def _process_args(self):
 
     def _process(self):
         title = f'{self.template.title} (copy)'
-        new_template = DesignerTemplate(title=title, type=self.template.type, data=self.template.data,
-                                        **self.target_dict)
+        new_template = DesignerTemplate(title=title, type=self.template.type, **self.target_dict)
+
+        data = self.template.data
```
ThiefMaster, line 184: Careful! you end up mutating the original template's data here. I think the only reason this may not get saved is that SQLAlchemy doesn't realize it has been mutated...

indico/modules/designer/controllers.py
```diff
@@ -281,9 +296,15 @@ def _process_GET(self):
     def _process_POST(self):
         data = dict({'background_position': 'stretch', 'items': []}, **request.json['template'])
         self.validate_json(TEMPLATE_DATA_JSON_SCHEMA, data)
-        invalid_placeholders = {x['type'] for x in data['items']} - set(get_placeholder_options()) - {'fixed'}
+        invalid_placeholders = {x['type'] for x in data['items']} - set(get_placeholder_options())
         if invalid_placeholders:
             raise UserValueError('Invalid item types: {}'.format(', '.join(invalid_placeholders)))
+        image_items = [item for item in data['items'] if item['type'] == 'fixed_image']
+        for image_item in image_items:
+            if image_item.get('image_id'):
```
ThiefMaster, line 304: Again, when would there *not* be an `image_id` for a `fixed_image` item?

indico/modules/designer/controllers.py
```diff
@@ -281,9 +296,15 @@ def _process_GET(self):
...
+                image = DesignerImageFile.get(image_item['image_id'])
```
ThiefMaster, line 305: Instead of querying the images on by one, why not build a set with all the images of the template (`{img.id for img in self.template.images}`) and then simply check if every `image_id` is in that set?

indico/modules/designer/placeholders.py
```diff
@@ -319,3 +322,29 @@ class RegistrationTicketQRPlaceholder(DesignerPlaceholder):
...
+    description = _('Fixed Image')
+    is_image = True
+
+    @classmethod
+    def render(cls, item):
+        if item.get('image_id'):
+            buf = DesignerImageFile.get(item['image_id']).open()
+            content = Image.open(buf)
+        else:
+            content = cls.description
```
ThiefMaster, line 349: When would this happen? Won't a `fixed_image` *always* have an `image_id`?

indico/modules/designer/controllers.py
SegiNyn: This could happen for example when a user inserts a fixed_image element in the badge designer but they do not upload any image into it. Then the image_id would be null

indico/modules/designer/controllers.py
SegiNyn: Tbh it's because it was easier to get the url from the item in the javascript. But to change this, would it be okay to include the images url as part of the template_data?

indico/modules/designer/controllers.py
ThiefMaster: I think failing validation (so the user has to remove the item or add an image) would be better than storing broken data...

indico/modules/designer/controllers.py
```diff
@@ -177,8 +179,19 @@ def _process_args(self):
 
     def _process(self):
         title = f'{self.template.title} (copy)'
-        new_template = DesignerTemplate(title=title, type=self.template.type, data=self.template.data,
-                                        **self.target_dict)
+        new_template = DesignerTemplate(title=title, type=self.template.type, **self.target_dict)
+
+        data = dict(self.template.data)
```
ThiefMaster, line 184: you need to make a deepcopy; otherwise you still mutate the individual items..

indico/modules/designer/controllers.py
SegiNyn: Okay

indico/modules/designer/controllers.py
```diff
@@ -281,9 +297,16 @@ def _process_GET(self):
     def _process_POST(self):
         data = dict({'background_position': 'stretch', 'items': []}, **request.json['template'])
         self.validate_json(TEMPLATE_DATA_JSON_SCHEMA, data)
-        invalid_placeholders = {x['type'] for x in data['items']} - set(get_placeholder_options()) - {'fixed'}
+        invalid_placeholders = {x['type'] for x in data['items']} - set(get_placeholder_options())
         if invalid_placeholders:
             raise UserValueError('Invalid item types: {}'.format(', '.join(invalid_placeholders)))
+        image_items = [item for item in data['items'] if item['type'] == 'fixed_image']
+        template_images = ({img.id for img in self.template.images})
```
ThiefMaster, line 304: Not sure why flake8 didn't complain about this, but those parentheses are certainly not needed...

```suggestion
        template_images = {img.id for img in self.template.images}
```

indico/modules/designer/controllers.py
```diff
@@ -281,9 +297,16 @@ def _process_GET(self):
...
+        for image_item in image_items:
+            if 'image_id' not in image_item:
+                raise UserValueError('Invalid: A Fixed Image element should have an image uploaded')
+            if not image_item['image_id'] in template_images:
```
ThiefMaster, line 308:
```suggestion
            if image_item['image_id'] not in template_images:
```

indico/modules/events/registration/badges.py
```diff
@@ -94,10 +94,10 @@ def _draw_badge(self, canvas, registration, template, tpl_data, pos_x, pos_y):
             if placeholder:
                 if placeholder.group == 'registrant':
                     text = placeholder.render(registration)
-                else:
+                elif placeholder.group == 'event':
                     text = placeholder.render(registration.event)
-            elif item['text']:
-                text = item['text']
+                else:
```
ThiefMaster, line 99: I'd use `elif placeholder.group == 'fixed'` here and else fail with an error (since that'd mean there's a placeholder with an unknown/unhandled group)

indico/modules/events/registration/badges.py
```diff
@@ -96,8 +96,10 @@ def _draw_badge(self, canvas, registration, template, tpl_data, pos_x, pos_y):
                     text = placeholder.render(registration)
                 elif placeholder.group == 'event':
                     text = placeholder.render(registration.event)
-                else:
+                elif placeholder.group == 'fixed':
                     text = placeholder.render(item)
+                else:
+                    raise ValueError('Unknown placeholder group: `{}`'.format(placeholder.group))
```
ThiefMaster, line 102: Don't you have the pre-commit.githook enabled? pyupgrade would point out that this should use an f-string:

```suggestion
                    raise ValueError(f'Unknown placeholder group: {placeholder.group}')
```

ThiefMaster (approved):

indico/modules/events/registration/badges.py
SegiNyn: Okay, I will use it for subsequent PRs. Thank you for the reviews.

PR #5275, opened 2022-03-08
Add hard wrapping to map URL in event settings

javfg:
This PR fixes very long map URLs from breaking out of the settings box in event management.

Before:

image

After:

image


ThiefMaster:
I think think one is a good 3.1.x candidate as well. Generally we can put bugfixes for things that already occur in production into that branch; no need to delay them until the next release. Especially if they are very simple and unlikely to cause any n egative side effects.

PR #5276, opened 2022-03-08
Fix a bug causing table details to be counted as rows

plourenco:
#4968 added table row details (as separate `<tr/>`) but the legacy counting logic we use considers these child selectors.
These changes prevent counting rows that are aggregated.


ThiefMaster:
This sound like a good candidate for 3.1.x..

PR #5277, opened 2022-03-09
Fix evaluation of available page width & height

SegiNyn:
This is for a bug found when calculating the available height and width when printing a badge. The calculations do not take into account the `right_margin` and `bottom_margin`. 

e.g.
`available_width = self.width - (config.left_margin - config.right_margin + config.margin_columns) * cm`
but if the right_margin is taken into account, then should be:
`available_width = self.width - (config.left_margin + config.right_margin + config.margin_columns) * cm`

So for example if you have the following settings in your badge printing configurations page: 
Screenshot 2022-03-09 at 17 12 00

'top_margin': 22.42,
'bottom_margin': 0.26,
'left_margin': 6.4,
'right_margin': 6.0,
'margin_columns': 0.0,
'margin_rows': 0.0,
The right_margin is not taken into account and instead prints the badge side by side. 

Screenshot 2022-03-09 at 17 12 28

After changing the expression then it takes the right_margin into account and prints the badge one per page:
Screenshot 2022-03-09 at 17 11 36

### DIFF
--- a/indico/modules/events/registration/badges.py
+++ b/indico/modules/events/registration/badges.py
@@ -50,9 +50,9 @@ class RegistrantsListToBadgesPDF(DesignerPDFBase):
     def _build_pdf(self, canvas):
         config = self.config
 
-        available_width = self.width - (config.left_margin - config.right_margin + config.margin_columns) * cm
+        available_width = self.width - (config.left_margin + config.right_margin + config.margin_columns) * cm
         n_horizontal = int(available_width / ((self.tpl_data.width_cm + config.margin_columns) * cm))
-        available_height = self.height - (config.top_margin - config.bottom_margin + config.margin_rows) * cm
+        available_height = self.height - (config.top_margin + config.bottom_margin + config.margin_rows) * cm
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Searching that file for `available_width =` shows two places where this calculation happens, so you probably need to fix it in the second place as well

SegiNyn:
Okay, thank you. 
This would also affect the Ticket when printed with the default badge setting specified in `DEFAULT_BADGE_SETTINGS` because it would now be too big for the page. Would it be okay to also change these default values, maybe decrease the right and left margins?

ThiefMaster:
How much would you lower them?

I guess lowering the margins could be problematic when printing in some cases, but not sure..

SegiNyn:
I will test on other page sizes to be sure

SegiNyn:
I think lowering only the `right_margin` to 0.4 and keeping all other values the same would be okay. This would allow the 'default ticket' which has bigger dimensions to print in the same way as before when using the default settings

PR #5281, opened 2022-03-10
Add missing notification trigger in abstract withdrawal

tomasr8:


ThiefMaster:
So the notifications withdrawal were never triggered? I'd say that's something for 3.1.x w/ a changelog entry

tomasr8:
Yup, they were missing completely. I updated the changelog

PR #5283, opened 2022-03-10
Add retention period to regform fields

duartegalvao:
This PR enables setting a retention period for the data of each field of a registration form.

screenshot

### DIFF
--- a/indico/modules/events/registration/client/js/form/FormItem.jsx
+++ b/indico/modules/events/registration/client/js/form/FormItem.jsx
@@ -145,7 +145,7 @@ FormItem.defaultProps = {
   fieldIsPersonalData: false,
   fieldIsRequired: false,
   isRequired: false,
-  retentionPeriod: 0,
+  retentionPeriod: null,
   htmlName: null,
   sortHandle: null,
   setupMode: false,
--- a/indico/modules/events/registration/client/js/form/FormItem.jsx
+++ b/indico/modules/events/registration/client/js/form/FormItem.jsx
@@ -10,7 +10,7 @@ import React from 'react';
 import {useSelector} from 'react-redux';
 import {Form, Icon, Popup} from 'semantic-ui-react';
 
-import {Translate} from 'indico/react/i18n';
+import {PluralTranslate, Translate} from 'indico/react/i18n';
 import {Markdown, toClasses} from 'indico/react/util';
 
 import {getManagement, isPaidItemLocked} from '../form_submission/selectors';
@@ -41,6 +41,7 @@ PaidItemLocked.propTypes = {
 export default function FormItem({
   title,
   description,
+  retentionPeriod,
   inputType,
   isEnabled,
   isRequired,
@@ -77,7 +78,20 @@ export default function FormItem({
             />
           ) : (
             <Form.Field required={isRequired || meta.alwaysRequired} styleName="field">
-              <label style={disabled ? {opacity: 0.8} : null}>{title}</label>
+              <label style={{opacity: disabled ? 0.8 : 1, display: 'inline-block'}}>{title}</label>
+              {setupMode && !!retentionPeriod && (
+                <Icon
+                  name="clock outline"
+                  color="red"
+                  style={{marginLeft: '3px'}}
+                  title={PluralTranslate.string(
+                    'Field data will be purged {retentionPeriod} week after the event ended.',
+                    'Field data will be purged {retentionPeriod} weeks after the event ended.',
+                    retentionPeriod,
+                    {retentionPeriod}
+                  )}
+                />
+              )}
               <InputComponent
                 isRequired={isRequired || meta.alwaysRequired}
                 disabled={disabled}
@@ -108,6 +122,8 @@ FormItem.propTypes = {
   position: PropTypes.number.isRequired,
   /** Whether the field is required during registration */
   isRequired: PropTypes.bool,
+  /** The retention period of the field's data in weeks */
+  retentionPeriod: PropTypes.number,
   /** Whether the field is a special "personal data" field */
   fieldIsPersonalData: PropTypes.bool,
   /** Whether the field cannot be disabled */
@@ -129,6 +145,7 @@ FormItem.defaultProps = {
   fieldIsPersonalData: false,
   fieldIsRequired: false,
   isRequired: false,
+  retentionPeriod: 0,
   htmlName: null,
   sortHandle: null,
   setupMode: false,
--- a/indico/modules/events/registration/client/js/form/fields/ChoicesSetup.jsx
+++ b/indico/modules/events/registration/client/js/form/fields/ChoicesSetup.jsx
@@ -5,6 +5,7 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
+import createDecorator from 'final-form-calculate';
 import {nanoid} from 'nanoid';
 import PropTypes from 'prop-types';
 import React from 'react';
@@ -15,6 +16,7 @@ import {Translate, Param} from 'indico/react/i18n';
 import {SortableWrapper, useSortableItem} from 'indico/react/sortable';
 
 import {getCurrency} from '../../form/selectors';
+
 import './ChoicesSetup.module.scss';
 
 export const choiceShape = {
@@ -137,6 +139,11 @@ Choices.defaultProps = {
   withExtraSlots: false,
   forAccommodation: false,
 };
+
+export const choiceFieldsSettingsFormDecorator = createDecorator({
+  field: 'choices',
+  updates: choices => ({isPriceSet: choices.some(c => c.price)}),
+});
 
 function Choice({
   withExtraSlots,
--- a/indico/modules/events/registration/client/js/form/fields/registry.js
+++ b/indico/modules/events/registration/client/js/form/fields/registry.js
@@ -56,6 +56,7 @@ Available keys:
 - noRequired: optional; hide the option to make the field required
 - alwaysRequired: optional; always display the field as required
 - hasPrice: optional; show price field if the whole field can have a price attached
+- noRetentionPeriod: optional; hide the retention period setting
 */
 
 const fieldRegistry = {
@@ -63,6 +64,7 @@ const fieldRegistry = {
     title: Translate.string('Static label'),
     inputComponent: LabelInput,
     noRequired: true,
+    noRetentionPeriod: true,
     noLabel: true,
     icon: 'tag',
     customFormItem: true,
--- a/indico/modules/events/registration/client/js/form/fields/registry.js
+++ b/indico/modules/events/registration/client/js/form/fields/registry.js
@@ -15,6 +15,7 @@ import AccommodationInput, {
 } from './AccommodationInput';
 import BooleanInput, {BooleanSettings} from './BooleanInput';
 import CheckboxInput, {CheckboxSettings} from './CheckboxInput';
+import {choiceFieldsSettingsFormDecorator} from './ChoicesSetup';
 import CountryInput from './CountryInput';
 import DateInput, {
   DateSettings,
@@ -46,7 +47,7 @@ Available keys:
 - settingsComponent: optional; used if the field has custom settings
 - settingsModalSize: optional; used if the field has settings which benefit
   from a larger modal size than "tiny"
-- settingsFormDecorator: optional; a final-form decorator to apply to the
+- settingsFormDecorators: optional; the final-form decorators to apply to the
   settings form
 - settingsFormValidator: optional, a function for final-form's form-level validation
 - settingsFormInitialData: optional; initial data to use when creating a new
@@ -102,7 +103,7 @@ const fieldRegistry = {
     title: Translate.string('Date'),
     inputComponent: DateInput,
     settingsComponent: DateSettings,
-    settingsFormDecorator: dateSettingsFormDecorator,
+    settingsFormDecorators: [dateSettingsFormDecorator],
     settingsFormInitialData: dateSettingsInitialData,
     icon: 'calendar',
   },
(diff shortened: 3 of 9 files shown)


indico/modules/events/registration/client/js/form_setup/ItemSettingsModal.jsx
```diff
@@ -145,6 +145,17 @@ export default function ItemSettingsModal({id, sectionId, defaultNewItemType, on
...
+          <FinalInput
+            name="retentionPeriod"
+            type="number"
+            placeholder={Translate.string('Permanent')}
+            step="1"
+            min="0"
+            validate={v.optional(v.min(0))}
+            parse={val => (val === '' ? 0 : val)}
+            format={val => (val === 0 ? '' : val)}
+            label={Translate.string('Retention period (months)')}
```
ThiefMaster, line 157: I remember we talked about this at some point...,hadn't we decided on a *weeks* granularity for this?

indico/modules/events/registration/fields/base.py
```diff
@@ -127,6 +131,7 @@ def process_field_data(cls, data, old_data=None, old_versioned_data=None):
         data = dict(data)
         if 'places_limit' in data:
             data['places_limit'] = int(data['places_limit']) if data['places_limit'] else 0
+        data['retention_period'] = int(data['retention_period']) if data.get('retention_period') else 0
```
ThiefMaster, line 134: With the MM schemas I don't think converting to int is needed anymore

indico/modules/events/registration/client/js/form_setup/ItemSettingsModal.jsx
```diff
@@ -145,6 +145,17 @@ export default function ItemSettingsModal({id, sectionId, defaultNewItemType, on
             />
           )}
           {SettingsComponent && <SettingsComponent {...itemData} />}
+          <FinalInput
```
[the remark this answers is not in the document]
ThiefMaster: We also need to prevent setting a retention period on certain always-mandatory fields (first/last name & email)

indico/modules/events/registration/client/js/form/FormItem.jsx
```diff
@@ -76,7 +77,17 @@ export default function FormItem({
...
           ) : (
             <Form.Field required={isRequired || meta.alwaysRequired} styleName="field">
-              <label style={disabled ? {opacity: 0.8} : null}>{title}</label>
+              <label style={{opacity: disabled ? 0.8 : 1, display: 'inline-block'}}>{title}</label>
+              {!!retentionPeriod && (
+                <Icon
+                  name="clock outline"
+                  color="red"
+                  style={{marginLeft: '3px'}}
+                  title={Translate.string('Data retention period set to {retentionPeriod} weeks', {
```
ThiefMaster, line 86: needs pluralization (`PluralTranslate.string(...)`)

indico/modules/events/registration/client/js/form_setup/ItemSettingsModal.jsx
```diff
@@ -145,6 +145,20 @@ export default function ItemSettingsModal({id, sectionId, defaultNewItemType, on
...
+              name="retentionPeriod"
+              type="number"
+              placeholder={Translate.string('Permanent')}
+              step="1"
+              min="1"
+              validate={v.optional(v.min(0))}
+              parse={val => (val === '' ? 0 : val)}
+              format={val => (val === 0 ? '' : val)}
+              label={Translate.string('Retention period (weeks)')}
+              disabled={fieldIsRequired}
```
ThiefMaster, line 159: Maybe better to hide it altogether for these fields?

indico/modules/events/registration/fields/base.py
```diff
@@ -126,7 +130,8 @@ def process_field_data(cls, data, old_data=None, old_versioned_data=None):
         """
         data = dict(data)
         if 'places_limit' in data:
-            data['places_limit'] = int(data['places_limit']) if data['places_limit'] else 0
+            data['places_limit'] = data.get('places_limit', 0)
+        data['retention_period'] = data.get('retention_period', 0)
```
ThiefMaster, line 134: this seems to set a value even for fields that can't have a retention period

ThiefMaster, line 134: this seems to set a value even for fields that can't have a retention period

indico/modules/events/registration/client/js/form/FormItem.jsx
```diff
@@ -77,7 +78,20 @@ export default function FormItem({
...
             <Form.Field required={isRequired || meta.alwaysRequired} styleName="field">
-              <label style={disabled ? {opacity: 0.8} : null}>{title}</label>
+              <label style={{opacity: disabled ? 0.8 : 1, display: 'inline-block'}}>{title}</label>
+              {setupMode && !!retentionPeriod && (
+                <Icon
+                  name="clock outline"
+                  color="red"
+                  style={{marginLeft: '3px'}}
+                  title={PluralTranslate.string(
+                    'Data retention period set to 1 week',
```
ThiefMaster, line 88: Better to use the placeholder here as well. There are languages without plural forms, and IIRC it can cause some problems if the singular string doesn't contain the placeholder.

ThiefMaster: The text sounds a bit like a message after changing it. How about sth. like "Field data will be purged X weeks after the event ended." instead?

indico/modules/events/registration/client/js/form/fields/registry.js
```diff
@@ -63,6 +63,7 @@ const fieldRegistry = {
     title: Translate.string('Static label'),
     inputComponent: LabelInput,
     noRequired: true,
+    noRetentionPeriod: true,
```
ThiefMaster, line 66: please document it in the comment block above :)

```
- noRetentionPeriod: optional; hide the retention period setting
```

indico/modules/events/registration/controllers/management/fields.py
```diff
@@ -49,6 +50,12 @@ def _check_input_type(self, input_type, **kwargs):
         if input_type not in get_field_types():
             raise ValidationError('Invalid field type')
 
+    @validates('retention_period')
+    def _check_retention_period(self, retention_period, **kwargs):
+        field = self.context['field']
+        if retention_period and field.view_data.get('fieldIsRequired'):
```
ThiefMaster, line 56: `view_data` is meant only for the front-end; check if the field is a personal data field and has `field.personal_data_type.is_required` instead

indico/modules/events/registration/models/form_fields.py
```diff
@@ -86,8 +86,9 @@ def versioned_data(self, value):
     def view_data(self):
         base_dict = self.versioned_data | self.data
         base_dict.update(section_id=self.parent_id, is_enabled=self.is_enabled, title=self.title,
-                         is_required=self.is_required, input_type=self.input_type, html_name=self.html_field_name,
-                         **super().view_data)
+                         is_required=self.is_required,
+                         retention_period=self.retention_period.days // 7 if self.retention_period else None,
```
ThiefMaster, line 90:
```suggestion
                         retention_period=(self.retention_period.days // 7 if self.retention_period else None),
```

indico/modules/events/registration/client/js/form/FormItem.jsx
```diff
@@ -129,6 +145,7 @@ FormItem.defaultProps = {
   fieldIsPersonalData: false,
   fieldIsRequired: false,
   isRequired: false,
+  retentionPeriod: 0,
```
ThiefMaster, line 148: I think we use `null` for no value, don't we? So maybe using `null` here would be better..

indico/modules/events/registration/fields/base.py
```diff
@@ -128,7 +128,7 @@ def process_field_data(cls, data, old_data=None, old_versioned_data=None):
         """
         data = dict(data)
         if 'places_limit' in data:
-            data['places_limit'] = int(data['places_limit']) if data['places_limit'] else 0
+            data['places_limit'] = data.get('places_limit', 0)
```
ThiefMaster, line 131: I guess this is guaranteed to be an int now since it's coming from a marshmallow schema?
Also, no need for `.get()` here since it's inside an `if 'places_limit' in data:` block..

indico/modules/events/registration/client/js/form/fields/ChoicesSetup.jsx
```diff
@@ -138,6 +140,11 @@ Choices.defaultProps = {
   forAccommodation: false,
 };
 
+export const choiceFieldsSettingsFormDecorator = createDecorator({
+  field: 'choices',
+  updates: choices => ({isPriceSet: choices.some(c => c.price)}),
```
ThiefMaster, line 145: Doesn't this mean that we end up sending `isPriceSet` in the form data when submitting now?

indico/modules/events/registration/fields/base.py
duartegalvao: Yeah. Actually, isn't this whole `if` block pointless?

ThiefMaster: Depends... if it's used in a PATCH-type update where only modified data is present, then this would be needed to not overwrite data. But I think for field setup we always send the whole settings, so I think it's indeed pointless.

indico/modules/events/registration/client/js/form/fields/ChoicesSetup.jsx
ThiefMaster: OK this doesn't seem to happen because `getValuesForFields` in `onSubmit` only uses values for registered fields.

indico/modules/events/registration/client/js/form_setup/ItemSettingsModal.jsx
```diff
@@ -146,6 +146,38 @@ export default function ItemSettingsModal({id, sectionId, defaultNewItemType, on
...
+            <>
+              <FinalInput
+                name="retentionPeriod"
+                type="number"
+                placeholder={Translate.string('Permanent')}
+                step="1"
+                min="1"
+                validate={v.optional(v.min(0))}
+                parse={val => (val === '' ? 0 : val)}
+                format={val => (val === 0 ? '' : val)}
```
ThiefMaster, line 159: I'm not sure about this... we don't ever want `0` as the value here, it should be `null` when there's no retention period.

Also I noticed that changing this field and reverting to the previous value afterwards keeps the form "dirty" - you can easily notice this by changing the retention period to some new value and then back - the submit button becomes active, but then stays active even if you revert to the previous value.

I gave it a quick try and `parse={val => (val === '' ? null : +val)}` seems to fix the problem, but the server-side code shouldn't accept a `0` value at all...

PS: I noticed the same happens with `placesLimit`, and in the JSON we seem to send it as a string instead of a number... maybe that's why?

indico/modules/events/registration/client/js/form/fields/ChoicesSetup.jsx
duartegalvao: Good. But still, is there a cleaner way to implement this functionality? Or should I leave it as is?

ThiefMaster: It's certainly a creative solution... for now I'd leave it as it is.

indico/modules/events/registration/controllers/management/fields.py
```diff
@@ -53,10 +53,12 @@ def _check_input_type(self, input_type, **kwargs):
...
         field = self.context['field']
-        if (retention_period
-                and field.type == RegistrationFormItemType.field_pd
-                and field.personal_data_type.is_required):
-            raise ValidationError('Cannot add retention period to required field')
+        if retention_period is not None:
+            if retention_period.days < 7:
+                raise ValidationError('Retention period must be at least 1 week')
+            if (field.type == RegistrationFormItemType.field_pd
+                    and field.personal_data_type.is_required):
```
ThiefMaster, line 60: doesn't this fit in one line now?

indico/modules/events/registration/client/js/form_setup/ItemSettingsModal.jsx
duartegalvao: With the `placesLimit` we agreed to use `0` as the "no limit" value, right? If so I can't find any problem with its behavior

ThiefMaster: Yes I think so. I believe the problem is simply that we seem to keep strings there instead of converting to a number:

```patch
diff --git a/indico/modules/events/registration/client/js/form/fields/BooleanInput.jsx b/indico/modules/events/registration/client/js/form/fields/BooleanInput.jsx
index e8a22a74cd..7943aae1ef 100644
--- a/indico/modules/events/registration/client/js/form/fields/BooleanInput.jsx
+++ b/indico/modules/events/registration/client/js/form/fields/BooleanInput.jsx
@@ -147,7 +147,7 @@ export function BooleanSettings() {
         step="1"
         min="1"
         validate={v.optional(v.min(0))}
-        parse={val => (val === '' ? 0 : val)}
+        parse={val => (val === '' ? 0 : +val)}
         format={val => (val === 0 ? '' : val)}
       />
       <FinalDropdown
diff --git a/indico/modules/events/registration/client/js/form/fields/CheckboxInput.jsx b/indico/modules/events/registration/client/js/form/fields/CheckboxInput.jsx
index db957da64b..0c5543a5aa 100644
--- a/indico/modules/events/registration/client/js/form/fields/CheckboxInput.jsx
+++ b/indico/modules/events/registration/client/js/form/fields/CheckboxInput.jsx
@@ -86,7 +86,7 @@ export function CheckboxSettings() {
       step="1"
       min="1"
       validate={v.optional(v.min(0))}
-      parse={val => (val === '' ? 0 : val)}
+      parse={val => (val === '' ? 0 : +val)}
       format={val => (val === 0 ? '' : val)}
     />
   );
```

With this patch applied I correctly get a number BTW. I think whoever wrote that original code (may have been me 🙈) thought number inputs already had a numeric `value`, but it's still a string of course...

PR #5284, opened 2022-03-11
Add an option to send notifications of new abstract comments

### LINKED ISSUE #5266: Email notifications for comments on abstract
I'm always frustrated when there is no mail notifications about non-judgement comments that added to a submission. The comments can be seen only at the submission  page and both submitter and event organizers must poll their abstract pages to see new comments. This might be bearable for a submitter but it is entirely not practical for event organizers.

I would like to be able to get email notifications for each comment on a submission. Preferably, there should be an option to edit the email template for this use-case, but even simple text like "there is a new comment on your submission" with URL or the submission would be a huge improvement.

tomasr8:
Closes #5266 

Add a new setting which enables to send email notifications
to abstract judges, conveners and reviewers when a new comment is added.

image

I used the same email template we use for paper notifications:

image

### DIFF
--- a/indico/modules/events/abstracts/controllers/reviewing.py
+++ b/indico/modules/events/abstracts/controllers/reviewing.py
@@ -9,6 +9,7 @@ from flask import flash, jsonify, request, session
 from sqlalchemy.orm import joinedload, subqueryload
 from werkzeug.exceptions import Forbidden
 
+from indico.core.notifications import make_email, send_email
 from indico.modules.events.abstracts.controllers.base import RHAbstractBase, RHAbstractsBase
 from indico.modules.events.abstracts.controllers.common import (AbstractsDownloadAttachmentsMixin, AbstractsExportCSV,
                                                                 AbstractsExportExcel, AbstractsExportPDFMixin,
@@ -18,11 +19,12 @@ from indico.modules.events.abstracts.forms import (AbstractCommentForm, Abstract
 from indico.modules.events.abstracts.lists import AbstractListGeneratorDisplay
 from indico.modules.events.abstracts.models.abstracts import Abstract, AbstractState
 from indico.modules.events.abstracts.models.comments import AbstractComment
-from indico.modules.events.abstracts.models.reviews import AbstractReview
+from indico.modules.events.abstracts.models.reviews import AbstractCommentVisibility, AbstractReview
 from indico.modules.events.abstracts.operations import (create_abstract_comment, create_abstract_review,
                                                         delete_abstract_comment, judge_abstract, reset_abstract_state,
                                                         update_abstract_comment, update_abstract_review,
                                                         update_reviewed_for_tracks, withdraw_abstract)
+from indico.modules.events.abstracts.settings import abstracts_reviewing_settings
 from indico.modules.events.abstracts.util import get_track_reviewer_abstract_counts, get_user_tracks
 from indico.modules.events.abstracts.views import WPDisplayAbstractsReviewing, render_abstract_page
 from indico.modules.events.tracks.models.tracks import Track
@@ -183,10 +185,38 @@ class RHSubmitAbstractComment(RHAbstractBase):
     def _check_abstract_protection(self):
         return self.abstract.can_comment(session.user)
 
+    def _get_recipients(self, visibility):
+        recipients = set()
+        for entry in self.abstract.event.acl_entries:
+            if self.abstract.can_judge(entry.user):
+                recipients.add(entry.user)
+            elif self.abstract.can_convene(entry.user) and visibility >= AbstractCommentVisibility.conveners:
+                recipients.add(entry.user)
+            elif self.abstract.can_review(entry.user) and visibility >= AbstractCommentVisibility.reviewers:
+                recipients.add(entry.user)
+        if visibility >= AbstractCommentVisibility.contributors:
+            recipients.add(self.abstract.submitter)
+            for comment in self.abstract.comments:
+                recipients.add(comment.user)
+        return recipients
+
+    def _send_notification(self, recipients, comment):
+        for recipient in recipients:
+            tpl = get_template_module('events/abstracts/emails/comment.html', event=self.event, abstract=self.abstract,
+                                      submitter=session.user, comment=comment, recipient=recipient)
+            email = make_email(to_list=recipient.email, template=tpl, html=True)
+            send_email(email, self.event, 'Abstracts', session.user)
+
     def _process(self):
         form = AbstractCommentForm(abstract=self.abstract, user=session.user)
         if form.validate_on_submit():
             create_abstract_comment(self.abstract, form.data)
+            if abstracts_reviewing_settings.get(self.event, 'notify_on_new_comments'):
+                visibility = form.visibility.data if form.visibility else AbstractCommentVisibility.contributors
+                recipients = self._get_recipients(visibility) - {session.user}
+                comment = form.text.data
+                self._send_notification(recipients, comment)
+
             return jsonify_data(flash=False, html=render_abstract_page(self.abstract, management=self.management))
         tpl = get_template_module('events/reviews/forms.html')
         return jsonify(html=tpl.render_comment_form(form, proposal=self.abstract))
--- a/indico/modules/events/abstracts/forms.py
+++ b/indico/modules/events/abstracts/forms.py
@@ -171,6 +171,10 @@ class AbstractReviewingSettingsForm(IndicoForm):
                                                   widget=SwitchWidget(),
                                                   description=_('Enabling this allows submitters, authors, and '
                                                                 'speakers to also participate in the comments.'))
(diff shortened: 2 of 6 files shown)


indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -183,10 +185,44 @@ class RHSubmitAbstractComment(RHAbstractBase):
     def _check_abstract_protection(self):
         return self.abstract.can_comment(session.user)
 
+    def _get_recipients(self):
+        judges = set()
+        conveners = set()
+        reviewers = set()
+
+        for entry in self.abstract.event.acl_entries:
```
tomasr8, line 193: Not sure if this could be a problem performance-wise.. we loop over all the acl entries in the event here

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -183,10 +185,44 @@ class RHSubmitAbstractComment(RHAbstractBase):
...
+            elif self.abstract.can_convene(entry.user):
+                conveners.add(entry.user)
+            elif self.abstract.can_review(entry.user):
+                reviewers.add(entry.user)
+        # Prevent sending duplicate emails for different roles
+        conveners -= judges
+        reviewers -= (judges | conveners)
+        return judges, conveners, reviewers
+
+    def _send_notification(self, recipients, comment, salutation):
```
ThiefMaster, line 205: `salutation1? how about `role`? :D

indico/modules/events/abstracts/controllers/reviewing.py
ThiefMaster: This should be fine... event ACLs are usually not that huge

indico/modules/events/abstracts/templates/emails/comment.html
```diff
@@ -0,0 +1,24 @@
+{% extends 'events/abstracts/emails/base.html' %}
+
+{% block subject_message %}New abstract comment{% endblock %}
+
+{% block header_subtitle %}New abstract comment{% endblock %}
+
+{% block header_subtitle_ref %}ref #{{ abstract.friendly_id }}{% endblock %}
+
+{% block content_header %}
+    <p>Dear {{ salutation }},</p>
```
ThiefMaster, line 10: I would avoid this; it'll be a huge pain once we add i18n to emails... How about just using `Dear <firstname>`?

tomasr8: I wanted to avoid sending individual emails, but with the i18n we'll probably have to do it anyway.. :+1:

indico/modules/events/abstracts/templates/emails/comment.html
```diff
@@ -0,0 +1,24 @@
...
+
+{% block subject_message %}New abstract comment{% endblock %}
+
+{% block header_subtitle %}New abstract comment{% endblock %}
+
+{% block header_subtitle_ref %}ref #{{ abstract.friendly_id }}{% endblock %}
+
+{% block content_header %}
+    <p>Dear {{ recipient.first_name }},</p>
+    <p>{{ submitter.full_name }} commented on the paper <strong>{{ abstract.title }}</strong>:</p>
```
ThiefMaster, line 11:
```suggestion
    <p>{{ submitter.full_name }} commented on the abstract <strong>{{ abstract.title }}</strong>:</p>
```

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -183,10 +185,45 @@ class RHSubmitAbstractComment(RHAbstractBase):
...
+
     def _process(self):
         form = AbstractCommentForm(abstract=self.abstract, user=session.user)
         if form.validate_on_submit():
             create_abstract_comment(self.abstract, form.data)
+            notify_on_new_comments = abstracts_reviewing_settings.get(self.event, 'notify_on_new_comments')
+            if notify_on_new_comments:
+                judges, conveners, reviewers = self._get_recipients()
+                comment = form.text.data
+                visibility = form.visibility.data
```
ThiefMaster, line 220: This doesn't exist if you are commenting as the abstract submitter/author (if "Allow contributors in comments" is enabled):

```py
  File "/home/adrian/dev/indico/py3/env/lib/python3.9/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/adrian/dev/indico/py3/env/lib/python3.9/site-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/home/adrian/dev/indico/py3/src/indico/web/flask/util.py", line 79, in wrapper
    return obj().process()
  File "/home/adrian/dev/indico/py3/src/indico/web/rh.py", line 281, in process
    res = self._do_process()
  File "/home/adrian/dev/indico/py3/src/indico/modules/events/controllers/base.py", line 90, in _do_process
    return RHEventBase._do_process(self)
  File "/home/adrian/dev/indico/py3/src/indico/web/rh.py", line 252, in _do_process
    rv = self._process()
  File "/home/adrian/dev/indico/py3/src/indico/modules/events/abstracts/controllers/reviewing.py", line 220, in _process
    visibility = form.visibility.data
AttributeError: 'NoneType' object has no attribute 'data'
```

ThiefMaster: Also, when a comment is posted with "Visible to contributors, reviewers, conveners, and judges", I think we should also notify:

- the submitter of the abstract
- anyone who previously commented on it (that way authors etc who commented will get notified, but we don't spam all other (co-)authors)

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -183,10 +185,45 @@ class RHSubmitAbstractComment(RHAbstractBase):
...
+            email = make_email(to_list=recipient.email, template=tpl, html=True)
+            send_email(email, self.event, 'Abstracts', session.user)
+
     def _process(self):
         form = AbstractCommentForm(abstract=self.abstract, user=session.user)
         if form.validate_on_submit():
             create_abstract_comment(self.abstract, form.data)
+            notify_on_new_comments = abstracts_reviewing_settings.get(self.event, 'notify_on_new_comments')
+            if notify_on_new_comments:
+                judges, conveners, reviewers = self._get_recipients()
```
ThiefMaster, line 218: What do you think about passing the visibility of the comment to this method, and then returning a single set? I think the split return value is a leftover from when you used "Dear judge" etc, but that's no longer needed..

indico/modules/events/abstracts/forms.py
```diff
@@ -171,6 +171,11 @@ class AbstractReviewingSettingsForm(IndicoForm):
                                                   widget=SwitchWidget(),
                                                   description=_('Enabling this allows submitters, authors, and '
                                                                 'speakers to also participate in the comments.'))
+    notify_on_new_comments = BooleanField(_('Comment notifications'),
+                                          [HiddenUnless('allow_comments', preserve_data=True)],
+                                          widget=SwitchWidget(),
+                                          description=_('Send an email notification to judges, conveners and reviewers '
+                                                        'when a new comment is added to an abstract.'))
```
ThiefMaster, line 178: I'd go for something more concise like "Send email notifications when new comments are posted"

indico/modules/events/abstracts/controllers/reviewing.py
```diff
@@ -183,10 +185,39 @@ class RHSubmitAbstractComment(RHAbstractBase):
...
+                                      submitter=session.user, comment=comment, recipient=recipient)
+            email = make_email(to_list=recipient.email, template=tpl, html=True)
+            send_email(email, self.event, 'Abstracts', session.user)
+
     def _process(self):
         form = AbstractCommentForm(abstract=self.abstract, user=session.user)
         if form.validate_on_submit():
             create_abstract_comment(self.abstract, form.data)
+            notify_on_new_comments = abstracts_reviewing_settings.get(self.event, 'notify_on_new_comments')
+            if notify_on_new_comments:
```
ThiefMaster, line 215:
```suggestion
            if abstracts_reviewing_settings.get(self.event, 'notify_on_new_comments'):
```

indico/modules/events/abstracts/templates/emails/comment.html
```diff
@@ -0,0 +1,24 @@
...
+    <p>{{ submitter.full_name }} commented on the abstract <strong>{{ abstract.title }}</strong>:</p>
+    <blockquote style="border-left: 5px solid #dfdfdf; padding-left: 10px; font-style: italic; color: #999;">
+        {{ comment }}
+    </blockquote>
+{% endblock %}
+
+{% block content_body %}
+    <p>You can view this comment on the abstract page:</p>
+    <p>
+        <a href="{{ url_for('abstracts.display_abstract', abstract, _external=true) }}">
```
ThiefMaster, line 20: This link isn't correct for contributors/reviewers/conveners as they do not have management area access

PR #5290, opened 2022-03-15
Fix bug while downloading a file without a published revision

plourenco:
**How to reproduce**
Download a file from any revision within editing without published files.

For reference: #5218


plourenco:
> The fix looks fine, but as always with bugfixes like that: 3.1.x (and maybe a changelog entry as well)

This is a chronic problem 😅

ThiefMaster (approved):

PR #5291, opened 2022-03-15
Remove registration data after the retention period expires

tomasr8:
Depends on #5283 

- [x]  Add a celery task to remove `RegistrationData` where the retention period of the associated field has expired when measured from the `event.end_dt`.
- [x] Signal in the frontend that the registration data have been deleted

### DIFF
--- a/indico/modules/events/registration/client/js/form/FormItem.jsx
+++ b/indico/modules/events/registration/client/js/form/FormItem.jsx
@@ -13,7 +13,7 @@ import {Form, Icon, Popup} from 'semantic-ui-react';
 import {PluralTranslate, Translate} from 'indico/react/i18n';
 import {Markdown, toClasses} from 'indico/react/util';
 
-import {getManagement, isPaidItemLocked} from '../form_submission/selectors';
+import {getManagement, getUpdateMode, isPaidItemLocked} from '../form_submission/selectors';
 
 import {getFieldRegistry} from './fields/registry';
 
@@ -38,6 +38,22 @@ PaidItemLocked.propTypes = {
   management: PropTypes.bool.isRequired,
 };
 
+function PurgedItemLocked({isUpdateMode}) {
+  const purgedMsg = <Translate>This field is locked due to an expired retention period.</Translate>;
+  const msg = !isUpdateMode ? (
+    purgedMsg
+  ) : (
+    <>
+      {purgedMsg} <Translate>The associated registration data has been deleted.</Translate>
+    </>
+  );
+  return <Popup trigger={<Icon name="lock" />}>{msg}</Popup>;
+}
+
+PurgedItemLocked.propTypes = {
+  isUpdateMode: PropTypes.bool.isRequired,
+};
+
 export default function FormItem({
   title,
   description,
@@ -45,6 +61,7 @@ export default function FormItem({
   inputType,
   isEnabled,
   isRequired,
+  isPurged,
   sortHandle,
   setupMode,
   setupActions,
@@ -53,17 +70,19 @@ export default function FormItem({
   // TODO move outside like with setupActions etc?
   const paidItemLocked = useSelector(state => isPaidItemLocked(state, rest.id));
   const isManagement = useSelector(getManagement);
+  const isUpdateMode = useSelector(getUpdateMode);
 
   const fieldRegistry = getFieldRegistry();
   const meta = fieldRegistry[inputType] || {};
   const InputComponent = meta.inputComponent;
   const inputProps = {title, description, isRequired, isEnabled, ...rest};
-  const disabled = !isEnabled || (paidItemLocked && !isManagement);
+  const disabled = !isEnabled || isPurged || (paidItemLocked && !isManagement);
   return (
     <div
       styleName={`form-item ${toClasses({
         'disabled': !isEnabled || paidItemLocked,
-        'paid-disabled': paidItemLocked,
+        'purged-disabled': isPurged,
+        'paid-disabled': !isPurged && paidItemLocked,
         'editable': setupMode,
       })}`}
     >
@@ -109,7 +128,8 @@ export default function FormItem({
         )}
       </div>
       {setupActions && <div styleName="actions">{setupActions}</div>}
-      {paidItemLocked && <PaidItemLocked management={isManagement} />}
+      {isPurged && <PurgedItemLocked isUpdateMode={isUpdateMode} />}
+      {!isPurged && paidItemLocked && <PaidItemLocked management={isManagement} />}
     </div>
   );
 }
@@ -122,6 +142,8 @@ FormItem.propTypes = {
   position: PropTypes.number.isRequired,
   /** Whether the field is required during registration */
   isRequired: PropTypes.bool,
+  /** Whether the field's registration data have been deleted due to an expired retention period */
+  isPurged: PropTypes.bool.isRequired,
   /** The retention period of the field's data in weeks */
   retentionPeriod: PropTypes.number,
   /** Whether the field is a special "personal data" field */
--- a/indico/modules/events/registration/client/js/form/FormItem.jsx
+++ b/indico/modules/events/registration/client/js/form/FormItem.jsx
@@ -78,6 +78,24 @@ export default function FormItem({
   const inputProps = {title, description, isRequired, isEnabled, ...rest};
   const showPurged = !setupMode && isPurged;
   const disabled = !isEnabled || showPurged || (paidItemLocked && !isManagement);
+
+  let retentionPeriodIcon = null;
+  if (setupMode && retentionPeriod) {
+    retentionPeriodIcon = (
+      <Icon
+        name="clock outline"
+        color="red"
+        style={{marginLeft: '3px'}}
+        title={PluralTranslate.string(
+          'Field data will be purged {retentionPeriod} week after the event ended.',
+          'Field data will be purged {retentionPeriod} weeks after the event ended.',
+          retentionPeriod,
+          {retentionPeriod}
+        )}
+      />
+    );
+  }
+
   return (
     <div
       styleName={`form-item ${toClasses({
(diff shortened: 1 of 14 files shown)


indico/migrations/versions/20220315_1520_5123f24eb41e_add_form_field_retention_period.py
```diff
@@ -0,0 +1,30 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('form_items',
+                  sa.Column('retention_period', sa.Interval(), nullable=True),
+                  schema='event_registration')
+    op.add_column('form_items',
+                  sa.Column('is_purged', sa.Boolean(), nullable=False, server_default='false'),
```
ThiefMaster, line 24: you need to remove the `server_default` again using `alter_column` after creating the column since we don't set this in the model

indico/modules/events/registration/tasks.py
```diff
@@ -0,0 +1,37 @@
...
+                         .join(Event)
+                         .filter(RegistrationFormItem.retention_period.isnot(None),
+                                 Event.end_dt + RegistrationFormItem.retention_period <= now_utc())
+                         .all())
+
+    for data in registration_data:
+        logger.info(f'Deleting registration field data: {data}')
+        data.data = None
+        if data.field_data.field.field_impl.is_file_field:
+            data.file = None
```
ThiefMaster, line 36: You need to delete the file from storage as well...

BTW, when deleting the file make sure that you fail gracefully (e.g. if there's a `StorageReadOnlyError` just log a warning but don't fail altogether); we don't want the whole purging to fail e.g. in case the storage became read-only - this would have happened in the past with the monthly AFS volumes for example, where old ones were no longer writable. Nowdays it's not such a big deal anymore but still better to handle it nicely!

indico/modules/events/registration/tasks.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@celery.periodic_task(name='registration_fields_retention_period', run_every=crontab(hour='*/24'))
+def delete_field_data():
+    registration_data = (RegistrationData.query
+                         .join(RegistrationFormFieldData)
+                         .join(RegistrationFormField, RegistrationFormFieldData.field_id == RegistrationFormField.id)
+                         .join(RegistrationForm)
+                         .join(Event)
+                         .filter(RegistrationFormItem.retention_period.isnot(None),
```
ThiefMaster, line 28: You can also skip using `is_purged` here; that way you don't act on already-purged data again and again.

indico/modules/events/registration/client/js/form/FormItem.jsx
```diff
@@ -38,12 +38,34 @@ PaidItemLocked.propTypes = {
...
+    <Translate>
+      This field is locked due to an expired retention period. The associated registration data have
+      been deleted.
+    </Translate>
+  );
+  const msg = !isUpdateMode ? (
+    purgedMsg
+  ) : (
+    <>
+      {purgedMsg} <Translate>The associated registration data have been deleted.</Translate>
```
ThiefMaster, line 52: I think data is usually used in a singular way, ie *has been deleted*.

Also , this text seems to be repeated here since pretty much the same string is already part of `purgedMsg`

indico/modules/events/registration/controllers/management/fields.py
```diff
@@ -49,6 +51,21 @@ def _check_input_type(self, input_type, **kwargs):
...
+        retention_period = data.get('retention_period')
+        field = self.context['field']
+        required_field_types = {t for t in PersonalDataType if t.is_required}
+
+        if retention_period is not None:
+            if field.type == RegistrationFormItemType.text:
+                raise ValidationError('Cannot set retention period for static labels', 'retention_period')
+            elif (
+                field.type == RegistrationFormItemType.field_pd and
+                field.personal_data_type in required_field_types
```
ThiefMaster, line 65: `field.personal_data_type.is_required`; no need for `required_field_types` ;)

indico/modules/events/registration/controllers/management/fields.py
```diff
@@ -29,6 +30,7 @@ class Meta:
     description = fields.String(load_default='')
     is_required = fields.Bool(load_default=False)
     input_type = fields.String(required=True, validate=not_empty)
+    retention_period = fields.TimeDelta(load_default=None)
```
ThiefMaster, line 33: careful about the type here; check #5283 where this is passed as a number indicating the weeks

indico/modules/events/registration/models/items.py
```diff
@@ -177,6 +177,11 @@ class RegistrationFormItem(db.Model):
         db.CheckConstraint('current_data_id IS NULL OR type IN ({t.field}, {t.field_pd})'
                            .format(t=RegistrationFormItemType),
                            name='current_data_id_only_field'),
+        db.CheckConstraint('retention_period IS NULL OR '
+                           '((type = {t.field}) OR ((type = {t.field_pd}) AND (personal_data_type IN '
+                           '({pt.affiliation}, {pt.title}, {pt.address}, {pt.phone}, {pt.country}, {pt.position}))))'
```
ThiefMaster, line 182: I think you have some unnecessary parentheses ;) And I'd rather use `NOT IN` for the required personal data fields, ie something like this:

```
retention_period IS NULL OR
type = field OR
(type = field_pd AND personal_data_type NOT IN (<required fields>))
```

FWIW i would not hardcode the required fields but get them from the enum using `is_required`

indico/modules/events/registration/controllers/management/fields.py
```diff
@@ -49,6 +50,20 @@ def _check_input_type(self, input_type, **kwargs):
...
+        retention_period = data.get('retention_period')
+        field = self.context['field']
+
+        if retention_period is not None:
+            if field.type == RegistrationFormItemType.text:
+                raise ValidationError('Cannot set retention period for static labels', 'retention_period')
+            elif (
+                field.type == RegistrationFormItemType.field_pd and
+                field.personal_data_type.is_required
+            ):
```
ThiefMaster, line 64: I think this fits in a single line now:

```suggestion
            elif field.type == RegistrationFormItemType.field_pd and field.personal_data_type.is_required:
```

indico/modules/events/registration/tasks.py
```diff
@@ -0,0 +1,42 @@
...
+                         .join(RegistrationFormField, RegistrationFormFieldData.field_id == RegistrationFormField.id)
+                         .join(RegistrationForm)
+                         .join(Event)
+                         .filter(~RegistrationFormItem.is_purged,
+                                 RegistrationFormItem.retention_period.isnot(None),
+                                 Event.end_dt + RegistrationFormItem.retention_period <= now_utc())
+                         .all())
+
+    for data in registration_data:
+        logger.info(f'Deleting registration field data: {data}')
```
ThiefMaster, line 34:
```suggestion
        logger.info('Deleting registration field data: %r', data)
```

that way tools like sentry can properly deduplicate log messages. not really relevant for `info` but still cleaner.

and maybe this should be `debug` level? on the `info` level I'd rather log which fields have been purged for a given registration, instead of one message for each field data...

indico/modules/events/registration/tasks.py
```diff
@@ -0,0 +1,42 @@
...
+        data.data = None
+        if data.field_data.field.field_impl.is_file_field:
+            try:
+                data.delete()
+            except Exception as e:
+                logger.error(f'Failed to delete file: {e}')
```
ThiefMaster, line 40: same as above

indico/modules/events/registration/util.py
```diff
@@ -364,6 +364,11 @@ def create_registration(regform, data, invitation=None, management=False, notify
         value = data.get(form_item.html_field_name, default) if can_modify else default
         data_entry = RegistrationData()
         registration.data.append(data_entry)
+
```
ThiefMaster, line 367: you really like blank lines, don't you? :D

indico/modules/events/registration/client/js/form/FormItem.jsx
```diff
@@ -53,17 +70,38 @@ export default function FormItem({
...
   const fieldRegistry = getFieldRegistry();
   const meta = fieldRegistry[inputType] || {};
   const InputComponent = meta.inputComponent;
   const inputProps = {title, description, isRequired, isEnabled, ...rest};
-  const disabled = !isEnabled || (paidItemLocked && !isManagement);
+  const showPurged = !setupMode && isPurged;
+  const disabled = !isEnabled || showPurged || (paidItemLocked && !isManagement);
+
+  let retentionPeriodIcon = null;
+  if (setupMode && !!retentionPeriod) {
```
ThiefMaster, line 83:
```suggestion
  if (setupMode && retentionPeriod) {
```

not needed here; the `!!` is only important when the falsy value might be used in a react render (where `0` would be rendered while `false` wouldn't)

indico/modules/events/registration/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def _group_by_regform(registration_data):
+    data_by_regform = defaultdict(list)
+    for data in registration_data:
+        data_by_regform[data.field_data.field.registration_form].append(data)
+    return data_by_regform
+
+
+@celery.periodic_task(name='registration_fields_retention_period', run_every=crontab(hour='*/24'))
```
ThiefMaster, line 30: This is running every minute..  you need to pick a fixed minute/hour, e.g.:

```suggestion
@celery.periodic_task(name='registration_fields_retention_period', run_every=crontab(minute='0', hour='3'))
```

indico/modules/events/registration/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+            # expect structured data e.g. Accommodation & {Single,Multi}Choice.
+            # This makes the React code relatively simple and we can always distinguish
+            # purged fields since they have the 'is_purged' flag set to True
+            data.data = data.field_data.field.field_impl.default_value
+            if data.field_data.field.field_impl.is_file_field:
+                logger.debug('Deleting file: %s', data.filename)
+                try:
+                    data.delete()
+                except Exception as e:
+                    logger.error('Failed to delete file: %r', e)
```
ThiefMaster, line 59:
```suggestion
                    logger.error('Failed to delete file for %r: %s', data, e)
```

otherwise it'll be hard to debug if this error happens...

indico/modules/events/registration/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+                         .join(Event)
+                         .filter(~RegistrationFormItem.is_purged,
+                                 RegistrationFormItem.retention_period.isnot(None),
+                                 Event.end_dt + RegistrationFormItem.retention_period <= now_utc())
+                         .all())
+
+    data_by_regform = _group_by_regform(registration_data)
+    for regform, regform_data in data_by_regform.items():
+        fields = [data.field_data.field for data in regform_data]
+        logger.info('Purging fields from registration %r: %r', regform, fields)
```
ThiefMaster, line 45:
```suggestion
        logger.info('Purging fields from regform %r: %r', regform, fields)
```

indico/modules/events/registration/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+@celery.periodic_task(name='registration_fields_retention_period', run_every=crontab(minute='0', hour='3'))
+def delete_field_data():
+    registration_data = (RegistrationData.query
+                         .join(RegistrationFormFieldData)
+                         .join(RegistrationFormField, RegistrationFormFieldData.field_id == RegistrationFormField.id)
+                         .join(RegistrationForm)
+                         .join(Event)
+                         .filter(~RegistrationFormItem.is_purged,
+                                 RegistrationFormItem.retention_period.isnot(None),
+                                 Event.end_dt + RegistrationFormItem.retention_period <= now_utc())
```
ThiefMaster, line 39: I'd use `RegistrationFormField` here - it's a subclass of `RegistrationFormItem` anyway, but since we use `RegistrationFormField` in the join above it'll be clearer if we use it here as well.

PR #5292, opened 2022-03-17
Make get_placeholders interceptable

SegiNyn:
Hello, can we make the `get_placeholders` function interceptable so that we can change the value of the `context`

### DIFF
--- a/indico/modules/events/registration/badges.py
+++ b/indico/modules/events/registration/badges.py
@@ -81,7 +81,7 @@ class RegistrantsListToBadgesPDF(DesignerPDFBase):
             with template.background_image.open() as f:
                 self._draw_background(canvas, ImageReader(self._remove_transparency(f)), tpl_data, *badge_rect)
 
-        placeholders = get_placeholders('designer-fields')
+        placeholders = get_placeholders('designer-fields', registration=registration)
 
         # Print images first
         image_placeholders = {name for name, placeholder in placeholders.items() if placeholder.is_image}
--- a/indico/util/placeholders.py
+++ b/indico/util/placeholders.py
@@ -13,7 +13,7 @@ from markupsafe import Markup, escape
 
 from indico.core import signals
 from indico.util.decorators import classproperty
-from indico.util.signals import named_objects_from_signal
+from indico.util.signals import make_interceptable, named_objects_from_signal
 
 
 class Placeholder:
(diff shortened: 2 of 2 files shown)


ThiefMaster:
What's the need for it? Do you want to fully replace the placeholders? Otherwise just register a signal handler to return your own placeholders.

TBH I'm slightly against making this one interceptable since it feels very hacky to intercept this...

SegiNyn:
We have an object type that uses a different set of placeholders and we want to be able to call a different signal that is not 'designer-fields' to get the correct placeholders when printing badges for that type.  So we use the type of the registration to know which signal to call. This is how I was using it: 

```
registration = args.arguments['kwargs']['registration']
if isinstance(registration, Accreditation):
    args.arguments['context'] = 'accreditation-designer-fields'
```

ThiefMaster:
but `get_placeholders` is usually not called with any kwargs.. so how do you get the `registration` kwarg there to begin with?

SegiNyn:
https://github.com/indico/indico/blob/e0ae8f6be08ec14076f3f4969809bd0f6e719e35/indico/modules/events/registration/badges.py#L84

ThiefMaster:
Sorry, but I don't think this is a good idea to add upstream. `get_placeholders('designer-fields')` is called in more places, and in some of them there's no `registration` available at all, so it'd be very inconsistent.

SegiNyn:
Okay, please is there another way you could suggest for us to this?

ThiefMaster:
Unfortunately not... Do you have the link to your custom code using it? Maybe I get an idea by looking at your code..

SegiNyn:
The placeholder is being used like this: 

https://github.com/UNOG-Indico/indico-un/blob/09cb91036239b3dca1f6ac62a7641e4c01af6a10/un/modules/designer/patches.py#L57

https://github.com/UNOG-Indico/indico-un/blob/09cb91036239b3dca1f6ac62a7641e4c01af6a10/un/modules/designer/patches.py#L81

SegiNyn:
And I wanted to change it to this: 

https://github.com/UNOG-Indico/indico-un/blob/a3bdf5055bd642444df9759c1437f6f61aa8a4c1/un/modules/designer/__init__.py#L64

ThiefMaster:
patch `get_placeholders` and if it's called with the normal context, use e.g. `flask.g.rh` to check the RH used (or the list of registrations used) in order to find out if you want to use your custom context instead

SegiNyn:
Okay, I'll try that. Thank you very much.

PR #5295, opened 2022-03-21
Add a regform-wide retention period

tomasr8:
Adds a new retention period setting for the while registration form.
This retention period cannot be shorter than the retention of individual fields.

New setting item:
image

### DIFF
--- a/indico/modules/events/registration/controllers/management/fields.py
+++ b/indico/modules/events/registration/controllers/management/fields.py
@@ -30,7 +30,7 @@ class GeneralFieldDataSchema(mm.Schema):
     is_required = fields.Bool(load_default=False)
     retention_period = fields.TimeDelta(load_default=None, precision=fields.TimeDelta.WEEKS)
     input_type = fields.String(required=True, validate=not_empty)
-    retention_period = fields.TimeDelta(load_default=None)
+    retention_period = fields.TimeDelta(load_default=None, precision='weeks')
 
     @validates('input_type')
     def _check_input_type(self, input_type, **kwargs):
--- a/indico/modules/events/registration/controllers/management/fields.py
+++ b/indico/modules/events/registration/controllers/management/fields.py
@@ -30,6 +30,7 @@ class GeneralFieldDataSchema(mm.Schema):
     is_required = fields.Bool(load_default=False)
     retention_period = fields.TimeDelta(load_default=None, precision=fields.TimeDelta.WEEKS)
     input_type = fields.String(required=True, validate=not_empty)
+    retention_period = fields.TimeDelta(load_default=None)
 
     @validates('input_type')
     def _check_input_type(self, input_type, **kwargs):
--- a/indico/modules/events/registration/controllers/management/fields.py
+++ b/indico/modules/events/registration/controllers/management/fields.py
@@ -54,7 +54,7 @@ class GeneralFieldDataSchema(mm.Schema):
 
             regform_retention_period = field.registration_form.retention_period
             if regform_retention_period and retention_period > regform_retention_period:
-                raise ValidationError(_('Retention period cannot be longer than that of registration form'),
+                raise ValidationError(_('Retention period cannot be longer than that of the registration form'),
                                       'retention_period')
 
     @post_load(pass_original=True)
--- a/indico/modules/events/registration/controllers/management/fields.py
+++ b/indico/modules/events/registration/controllers/management/fields.py
@@ -17,6 +17,7 @@ from indico.modules.events.registration.fields import get_field_types
 from indico.modules.events.registration.models.form_fields import RegistrationFormField
 from indico.modules.events.registration.models.items import RegistrationFormItemType, RegistrationFormText
 from indico.modules.events.registration.util import get_flat_section_positions_setup_data, update_regform_item_positions
+from indico.util.i18n import _
 from indico.util.marshmallow import not_empty
 from indico.util.string import snakify_keys
 
@@ -30,7 +31,6 @@ class GeneralFieldDataSchema(mm.Schema):
     is_required = fields.Bool(load_default=False)
     retention_period = fields.TimeDelta(load_default=None, precision=fields.TimeDelta.WEEKS)
     input_type = fields.String(required=True, validate=not_empty)
-    retention_period = fields.TimeDelta(load_default=None, precision='weeks')
 
     @validates('input_type')
     def _check_input_type(self, input_type, **kwargs):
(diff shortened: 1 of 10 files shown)


indico/modules/events/registration/forms.py
```diff
@@ -519,3 +525,15 @@ def validate_visibility(self, field):
             (participant_visibility_changed_to_show_all or public_visibility_changed_to_show_all)
         ):
             raise ValidationError(_("'Show all participants' can only be set if there are no registered users."))
+
+    def validate_retention_period(self, field):
+        retention_period = field.data
+        fields = (RegistrationFormItem.query
+                  .with_parent(self.regform)
+                  .filter(RegistrationFormItem.retention_period.isnot(None),
```
ThiefMaster, line 533: I'd ignore exclude disabled/deleted fields here

indico/modules/events/registration/models/forms.py
```diff
@@ -243,6 +243,11 @@ class RegistrationForm(db.Model):
         nullable=True,
         index=True
     )
+    #: period for which the registration data of all fields should be kept
```
ThiefMaster, line 246:
```suggestion
    #: period for which the registration itself should be kept
```

indico/modules/events/registration/controllers/management/fields.py
```diff
@@ -51,6 +52,11 @@ def _check_retention_period(self, retention_period, **kwargs):
             if field.type == RegistrationFormItemType.field_pd and field.personal_data_type.is_required:
                 raise ValidationError('Cannot add retention period to required field')
 
+            regform_retention_period = field.registration_form.retention_period
+            if retention_period > regform_retention_period:
+                raise ValidationError('Retention period cannot be longer than that of registration form',
```
ThiefMaster, line 57: this one needs i18n since i don't think we validate it on the client side (which if fine IMO, we don't have access to the global retention period there)

indico/modules/events/registration/tasks.py
```diff
@@ -58,3 +58,19 @@ def delete_field_data():
...
+    registrations = (Registration.query
+                     .join(RegistrationForm)
+                     .join(Event)
+                     .filter(RegistrationForm.retention_period.isnot(None),
+                             Event.end_dt + RegistrationForm.retention_period <= now_utc())
+                     .all())
+
+    for reg in registrations:
+        logger.info('Purging registration: %r', reg)
+        reg.delete()
```
ThiefMaster, line 74: - delete files explicitly (just go over all field data of the registration that has a `storage_file_id`, no need to check field types etc)
- `db.session.delete(...)` to delete the registration itself

indico/modules/events/registration/controllers/management/privacy.py
```diff
@@ -27,8 +27,9 @@ def _process(self):
         form = RegistrationPrivacyForm(event=self.event, regform=self.regform, visibility=[
             self.regform.publish_registrations_participants.name,
             self.regform.publish_registrations_public.name
-        ])
+        ], retention_period=self.regform.retention_period)
         if form.validate_on_submit():
+            form.populate_obj(self.regform, skip={'visibility'})
```
ThiefMaster, line 32: hm since we just have two fields and already skip one maybe it's better to use `only=...` (or manually set `self.regform.retention_period = form.retention_period.data`)?

indico/modules/events/registration/tasks.py
```diff
@@ -58,3 +58,19 @@ def delete_field_data():
                     logger.error('Failed to delete file for %r: %s', data, e)
             data.field_data.field.is_purged = True
     db.session.commit()
+
+
+@celery.periodic_task(name='registration_retention_period', run_every=crontab(minute='0', hour='3'))
+def delete_registration():
```
ThiefMaster, line 64:
```suggestion
def delete_registrations():
```

ThiefMaster:
Maybe we should clearly indicate e.g. on the main regform management dashboard if a retention period is set? Especially since this is a permanent deletion I think it's good to clearly indicate that one is set...

I tried this and it doesn't look bad IMHO:

```jinja
    {% if regform.retention_period%}
        {%- call message_box('warning', fixed_width=true) -%}
            {% trans count=(regform.retention_period.days // 7), date=(event.end_dt + regform.retention_period)|format_date %}
                Registrations will be permanently deleted {{ count }} week after the event ended (on {{ date }}).
            {% pluralize %}
                Registrations will be permanently deleted {{ count }} weeks after the event ended (on {{ date }}).
            {% endtrans %}
        {%- endcall %}
    {% endif %}
```

image

indico/modules/events/registration/tasks.py
```diff
@@ -52,9 +59,25 @@ def delete_field_data():
...
+                     .join(Event)
+                     .filter(RegistrationForm.retention_period.isnot(None),
+                             db.cast(Event.end_dt, db.Date) + RegistrationForm.retention_period <= now_utc().date())
+                     .all())
+
+    for reg in registrations:
+        logger.info('Purging registration: %r', reg)
+        for data in reg.data:
+            if data.storage_file_id:
+                _delete_file(data)
```
ThiefMaster, line 80: I'd keep a `logger.debug('Deleting file: %s', data.filename)` here as well (or move it inside `_delete_file`)

indico/modules/events/registration/tasks.py
```diff
@@ -51,10 +59,25 @@ def delete_field_data():
...
+                     .join(RegistrationForm, RegistrationForm.id == Registration.registration_form_id)
+                     .join(Event)
+                     .filter(RegistrationForm.retention_period.isnot(None),
+                             db.cast(Event.end_dt, db.Date) + RegistrationForm.retention_period <= now_utc().date())
+                     .all())
+
+    for reg in registrations:
+        logger.info('Purging registration: %r', reg)
+        for data in reg.data:
+            if data.storage_file_id:
```
ThiefMaster, line 79: Maybe better to use `data.field_data.field.field_impl.is_file_field` like we do above to be consistent?

ThiefMaster:
Should we prevent creating new registrations once the deletion date has been reached? Just in case someone forgets to close the registration or adds a participant through the management area...

indico/modules/events/registration/controllers/management/fields.py
```diff
@@ -51,6 +52,11 @@ def _check_retention_period(self, retention_period, **kwargs):
             if field.type == RegistrationFormItemType.field_pd and field.personal_data_type.is_required:
                 raise ValidationError('Cannot add retention period to required field')
 
+            regform_retention_period = field.registration_form.retention_period
+            if regform_retention_period and retention_period > regform_retention_period:
+                raise ValidationError(_('Retention period cannot be longer than that of registration form'),
```
ThiefMaster, line 57: ...that of *the* registration form

ThiefMaster (approved):

PR #5299, opened 2022-03-25
Add validation for text input size values

duartegalvao:
This PR includes:

- Validation to ensure minimum length is smaller than maximum length in regform text inputs;
- Validation to ensure the value of the text field is within the minimum and maximum lengths set in the settings;
- A fix to an error that occurred when setting only the minimum length/value in text and number inputs.

### DIFF
--- a/indico/modules/events/registration/client/js/form/fields/TextInput.jsx
+++ b/indico/modules/events/registration/client/js/form/fields/TextInput.jsx
@@ -27,7 +27,6 @@ export default function TextInput({htmlName, disabled, title, isRequired, ...pro
       {...inputProps}
       required={isRequired}
       disabled={disabled}
-      parse={val => val || null}
     />
   );
 }
@@ -43,8 +42,8 @@ TextInput.propTypes = {
 
 TextInput.defaultProps = {
   disabled: false,
-  minLength: null,
-  maxLength: null,
+  minLength: 0,
+  maxLength: 0,
 };
 
 export function TextSettings() {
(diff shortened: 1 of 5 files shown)


indico/modules/events/registration/client/js/form/fields/TextInput.jsx
```diff
@@ -27,6 +27,7 @@ export default function TextInput({htmlName, disabled, title, isRequired, ...pro
       {...inputProps}
       required={isRequired}
       disabled={disabled}
+      parse={p.nullIfEmpty}
```
ThiefMaster, line 30: We have a prop for this in FinalInput:

```suggestion
      nullIfEmpty
```

(and no need to import `p` then)

What's the reason for sending an empty text field as `null` btw? Currently we store an empty string and never `null` if the field is empty...

duartegalvao: exactly. is there a cleaner way to go around that? otherwise fields with a min length set are always required

ThiefMaster: I don't think so... I don't like the idea that we are suddenly switching from storing an empty string to storing null.

One option could be subclassing the `Length` validator and overriding its `__call__` method to return early `if not value`

ThiefMaster: Actually, I think a minimum length of 1 makes no sense at all. If people want to require input, the field should be required. If people do not want single-char values, they can set a higher maxlength

indico/modules/events/registration/fields/simple.py
```diff
@@ -28,23 +28,41 @@
 KEEP_EXISTING_FILE_UUID = '00000000-0000-0000-0000-000000000001'
 
 
+class TextFieldDataSchema(mm.Schema):
+    min_length = fields.Integer(load_default=0, validate=validate.And(validate.Range(0), validate.NoneOf((1,))))
```
ThiefMaster, line 32: Why not simply use `validate.Range(1)`? ;)

ThiefMaster: ah nevermind, `0` is the default so that needs to be allowed as well...

ThiefMaster (approved):

PR #5302, opened 2022-03-28
Show hidden participants in participant list

### LINKED ISSUE #5301: Participant list: Show hidden participants
I think this is especially useful in the conference layout, a simple "5 more participants" row to indicate there are some unpublished participants should do the job.

If we can find a nice way of showing it in the meeting/lecture theme as well, even better!

duartegalvao:
Closes #5301

This PR adds a "X more participants" row to participants lists to indicate that there are unpublished participants, as well as a "+X" badge on meeting/lecture-type events.

screenshot

screenshot

### DIFF
--- a/indico/modules/events/registration/controllers/display.py
+++ b/indico/modules/events/registration/controllers/display.py
@@ -5,7 +5,6 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
-from operator import attrgetter
 from uuid import UUID
 
 from flask import flash, jsonify, redirect, request, session
@@ -13,6 +12,7 @@ from sqlalchemy.orm import contains_eager, joinedload, lazyload, load_only, subq
 from webargs import fields
 from werkzeug.exceptions import Forbidden, NotFound
 
+from indico.core.db import db
 from indico.modules.auth.util import redirect_to_login
 from indico.modules.events.controllers.base import RegistrationRequired, RHDisplayEventBase
 from indico.modules.events.models.events import EventType
@@ -22,8 +22,7 @@ from indico.modules.events.registration.controllers import RegistrationEditMixin
 from indico.modules.events.registration.models.forms import RegistrationForm
 from indico.modules.events.registration.models.invitations import InvitationState, RegistrationInvitation
 from indico.modules.events.registration.models.items import PersonalDataType
-from indico.modules.events.registration.models.registrations import (PublishRegistrationsMode, Registration,
-                                                                     RegistrationState)
+from indico.modules.events.registration.models.registrations import Registration, RegistrationState
 from indico.modules.events.registration.util import (check_registration_email, create_registration, generate_ticket,
                                                      get_event_regforms_registrations, get_flat_section_submission_data,
                                                      get_initial_form_values, get_user_data, make_registration_schema)
@@ -140,7 +139,7 @@ class RHParticipantList(RHRegistrationFormDisplayBase):
         headers = [PersonalDataType[column_name].get_title() for column_name in column_names]
 
         query = (Registration.query.with_parent(self.event)
-                 .filter(~RegistrationForm.is_deleted, ~Registration.is_deleted)
+                 .filter(Registration.is_state_publishable, ~RegistrationForm.is_deleted)
                  .join(Registration.registration_form)
                  .options(subqueryload('data').joinedload('field_data'),
                           contains_eager('registration_form')))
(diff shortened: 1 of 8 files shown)


indico/modules/events/registration/util.py
```diff
@@ -491,6 +491,26 @@ def get_published_registrations(event, is_participant):
...
+    :param event: the `Event` to get registrations for
+    :param is_participant: whether the user accessing the registrations is a participant of the event
+    :return: list of `Registration` objects
+    """
+    query = (Registration.query.with_parent(event)
+             .filter(~Registration.is_publishable(is_participant),
+                     ~RegistrationForm.is_deleted,
+                     ~Registration.is_deleted)
+             .join(Registration.registration_form)
+             .options(contains_eager(Registration.registration_form))
```
ThiefMaster, line 506: not needed

indico/modules/events/registration/util.py
```diff
@@ -491,6 +491,26 @@ def get_published_registrations(event, is_participant):
...
+             .order_by(db.func.lower(Registration.first_name),
+                       db.func.lower(Registration.last_name),
+                       Registration.friendly_id))
```
ThiefMaster, line 509: not needed when just couting..

indico/modules/events/registration/util.py
```diff
@@ -491,6 +491,26 @@ def get_published_registrations(event, is_participant):
...
 
+def get_hidden_registrations(event, is_participant):
+    """Get the number of hidden registrations for an event.
+
+    :param event: the `Event` to get registrations for
+    :param is_participant: whether the user accessing the registrations is a participant of the event
+    :return: list of `Registration` objects
+    """
+    query = (Registration.query.with_parent(event)
+             .filter(~Registration.is_publishable(is_participant),
```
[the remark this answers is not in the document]
ThiefMaster: I think adding these two criteria to the filter here will do the job:

```
Registration.is_active,
Registration.state.in_((RegistrationState.complete, RegistrationState.unpaid)),
```

indico/modules/events/registration/controllers/display.py
```diff
@@ -139,13 +140,13 @@ def _deduplicate_reg_data(reg_data_iter):
...
-                         ~RegistrationForm.is_deleted,
-                         ~Registration.is_deleted)
+                 .filter(~RegistrationForm.is_deleted, ~Registration.is_deleted)
                  .join(Registration.registration_form)
                  .options(subqueryload('data').joinedload('field_data'),
                           contains_eager('registration_form')))
-        registrations = sorted(_deduplicate_reg_data(_process_registration(reg, column_names) for reg in query),
+        is_participant = self.event.is_user_registered(session.user)
+        registrations = sorted(_deduplicate_reg_data(_process_registration(reg, column_names)
+                                                     for reg in query if reg.is_publishable(is_participant)),
```
ThiefMaster, line 149: needs to check state + active (see my comment on `get_hidden_registrations`)

indico/modules/events/registration/util.py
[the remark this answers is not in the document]
ThiefMaster: Since you need this check in multiple places, maybe add a `is_state_publishable` hybrid property containing these two checks; that way you can use `Registration.is_state_publishable` here and `reg.is_state_publishable` in the checks for conference events

ThiefMaster (changes requested):

duartegalvao:
> image
> 
> When there are no visible named participants, this doesn't look amazing... maybe repeating the "participants" word ("3 participants") would look less awkward?

I'm split between that or showing the participants number in a label:
image

ThiefMaster:
i think seeing screenshots of either option with some more context around it would be helpful there..

duartegalvao:
> i think seeing screenshots of either option with some more context around it would be helpful there..

Option A:
image

Option B:
image

ThiefMaster:
hm not a huge fan of either, but i think the label looks slightly better.

BTW just to make sure: we only show this if participants are *not* hidden altogether but there are simply on participants visible to *you*, right? because if it's set to 'hide all' then i don't think we should be showing this

duartegalvao:
> hm not a huge fan of either, but i think the label looks slightly better.
> 
> BTW just to make sure: we only show this if participants are _not_ hidden altogether but there are simply on participants visible to _you_, right? because if it's set to 'hide all' then i don't think we should be showing this

Yes, but thank you for reminding me, I was missing that check for meetings/lectures.

So, the "number only" should only ever be shown when:
a) the form is set to show with consent, but none of the participants have consented
b) after #5306, when the form is not set to hide all and the visibility period has expired

ThiefMaster:
consent or always, since the latter could have some manually-hidden participants too - so basically just hide it if it's set to hide_all

duartegalvao:
> consent or always, since the latter could have some manually-hidden participants too - so basically just hide it if it's set to hide_all

yes, exactly

indico/modules/events/registration/templates/display/event_header.html
```diff
@@ -76,23 +76,37 @@
...
                 </ul>
                 <a class="trigger icon-expand" title="{% trans %}See the full list{% endtrans %}"></a>
             </div>
         </div>
     </div>
+{% elif num_hidden_registrations %}
+    <div class="event-details-row">
+        <div class="event-details-label">{% trans %}Participants{% endtrans %}</div>
+        <div class="event-details-content">
+            <span class="ui circular label" title="{% trans num=num_hidden_registrations %}{{ num }} participant{% pluralize %}{{ num }} participants{% endtrans %}">
```
ThiefMaster, line 107: I think that tooltip isn't very useful here... it makes more sense above where we have normal and hidden ones, but here it adds no additional value

PR #5305, opened 2022-03-28
Update participant filter logic

plourenco:
Under "Participant Roles" there is a filter drop-down menu. By default, all roles are selected. If one wants to deselect all and then select only one or two, they have to manually go and click on all the others.

My proposal is to ease this behaviour and make the logic of none selected the same as all selected, updating the "X" for reset filters.

imagem
imagem

The images above would become the same, having by default none selected.

### DIFF
--- a/indico/modules/events/management/client/js/index.js
+++ b/indico/modules/events/management/client/js/index.js
@@ -81,8 +81,9 @@ import {SingleEventMove, EventPublish} from './EventMove';
       })
       .get();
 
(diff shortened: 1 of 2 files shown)


ThiefMaster:
What's the logic when opening the page though? I think at that point we usually want to see everything. Or is the new logic that nothing selected == no filters?

plourenco:
When the page is first opened, no filters are selected, which means everything is shown. Nothing selected == no filters == all results.

PR #5309, opened 2022-03-30
Make update_event  interceptable

vasantvohra:
We have introduced two fields arrangement_type enum (virtual, hybrid, in-person), and virtual_details fields. 
We would like to pop, save and update logs before it goes to update_events and gives an assertion error.


ThiefMaster:
Yeah that seems reasonable. Would be OK for 3.1.x as well btw (but we can stick with master if you don't need it in 3.1.x).

vasantvohra:
``` Error while rebasing to master
 ! [remote rejected]       intercept/update_events -> intercept/update_events (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/ci.yml` without `workflow` scope)
```

vasantvohra:
Fixed & done

PR #5310, opened 2022-03-31
Fix new image not added in images

SegiNyn:
This is to fix an issue with my previous PR, the newest image was not added in the images dict. So the first time an image is added, it temporarily disappears when you resize the element and then appears again after saving.  

https://user-images.githubusercontent.com/54508387/161070896-e440f51e-293e-42dc-b0b2-f3751ae60936.mov

### DIFF
--- a/indico/modules/designer/client/js/index.js
+++ b/indico/modules/designer/client/js/index.js
@@ -120,7 +120,7 @@ import {$T} from 'indico/utils/i18n';
         .text(
           this.type === 'fixed'
             ? this.text
-            : this.type === 'fixed_image'
+            : this.type === 'fixed_image' && this.image_id
             ? ''
             : itemTitles[this.type]
         );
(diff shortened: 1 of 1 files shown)


plourenco (commented):
Hi,

Thanks for following up! Resizing the image works fine now. 

However, there's a problem while saving the image. If you try to delete an image you've added (but keep the Fixed Image item) an error will come up:

imagem

This is caused by a null `fixed_image` item which does not contain an `image_id` causing the check in `RHEditDesignerTemplate`:

```python
            if image_item['image_id'] not in template_images:
                raise UserValueError(_('The image file does not belong to this template'))
```

to fail. Perhaps we could make sure it isn't None?

SegiNyn:
Hello, thank you for the review. 

Okay, so we would add the check making sure the fixed_image is not null in the JavaScript before the template is saved?

SegiNyn:
Okay, I'll add the check to prevent it from being empty before it is saved.

indico/modules/designer/client/js/index.js
```diff
@@ -594,6 +594,17 @@ import {$T} from 'indico/utils/i18n';
       );
       return;
     }
+    for (const item of Object.values(items)) {
+      if (item.type === 'fixed_image' && !item.image_id) {
```
ThiefMaster, line 598: I think you can simplify this check like this:

```js
if (Object.values(items).some(item => item.type === 'fixed_image' && !item.image_id)) {
  ...
}
```

SegiNyn: Okay, thanks

ThiefMaster (approved):

PR #5311, opened 2022-04-01
Show data controller information

### LINKED ISSUE #5307: Show data controller information
We added the setting, but don't show it anywhere. Let's figure out where we can show it and add it there....

duartegalvao:
Closes #5307

This PR adds a new privacy information menu with the data controller information to meetings and lectures, as well as a privacy information page in conferences.

screenshot

screenshot

### DIFF
--- a/indico/modules/events/controllers/display.py
+++ b/indico/modules/events/controllers/display.py
@@ -17,8 +17,7 @@ from indico.modules.events.layout.views import WPPage
 from indico.modules.events.management.settings import privacy_settings
 from indico.modules.events.models.events import EventType
 from indico.modules.events.util import get_theme
-from indico.modules.events.views import WPConferenceDisplay, WPSimpleEventDisplay
-from indico.modules.legal.views import WPDisplayPrivacyPolicy
+from indico.modules.events.views import WPConferenceDisplay, WPConferencePrivacyDisplay, WPSimpleEventDisplay
 from indico.web.args import use_kwargs
 from indico.web.flask.util import send_file, url_for
 from indico.web.rh import allow_signed_url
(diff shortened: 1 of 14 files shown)


duartegalvao:
> I wonder if there's a better way to show this in conference events... in meetings the placement is nice but in conferences it feels kind of out of place...

Maybe in conferences it would make sense to have a dedicated page with a menu entry for it?
screenshot

ThiefMaster:
I think that may be a good idea :)

indico/modules/events/templates/display/privacy.html
```diff
@@ -0,0 +1,70 @@
...
+                        <div>
+                            <a href="mailto:{{ privacy_info.data_controller_email }}">{{ privacy_info.data_controller_email }}</a>
+                        </div>
+                    {% endif %}
+                </div>
+            </div>
+        </div>
+    {% endif %}
+    {% if privacy_info.privacy_policy %}
+        {{ privacy_info.privacy_policy|sanitize_html }}
```
ThiefMaster, line 40: You're missing the "privacy notice" header here

indico/modules/events/controllers/display.py
```diff
@@ -84,8 +84,11 @@ class RHDisplayPrivacyPolicy(RHEventBase):
...
-        if urls := privacy_settings.get(self.event, 'privacy_policy_urls'):
-            # There could be multiple urls, redirect to the first one.
-            return redirect(urls[0]['url'])
+        if self.event.type_ == EventType.conference:
+            return WPConferencePrivacyDisplay.render_template(
+                'display/privacy.html',
+                self.event,
+                privacy_info=privacy_settings.get_all(self.event)
+            )
         return WPDisplayPrivacyPolicy.render_template('privacy.html',
```
ThiefMaster, line 93: I only noticed now, but this WP and template are coming from the `legal` module which is meant for the global privacy notice... so it's rendered completely unrelated to the event. Since we no longer have a footer link pointing to this page in any case, I think we can always render the conference version - so if someone appends `/privacy` to the URL in a meeting they just get it in the conference layout like it happens for most other conference-only pages as well...

ThiefMaster (approved):

PR #5313, opened 2022-04-06
Add option to delete unused 'EventPerson's

### LINKED ISSUE #5294: Let managers remove unused EventPersons
We currently hide them on the "Participant roles" page.

It'd be nice to have a a "show unused" toggle, and provide a way to remove those from the event altogether.

tomasr8:
Closes #5294 

Adds a new filter and an option to delete unused `EventPerson`s on the
_participant roles_ page

image

### DIFF
--- a/indico/modules/events/models/persons.py
+++ b/indico/modules/events/models/persons.py
@@ -350,19 +350,14 @@ class EventPerson(PersonMixin, db.Model):
     def has_links(self):
         """Whether the person is an author, speaker, etc., in the event."""
         chairpersons = {link.person for link in self.event.person_links}
-        if self in chairpersons:
-            return True
-        if (
-            self.event.has_feature('abstracts') and
-            {link for link in self.abstract_links if not link.abstract.is_deleted}
-        ):
-            return True
-        if {link for link in self.session_block_links if not link.session_block.session.is_deleted}:
-            return True
-        if {link for link in self.contribution_links if not link.contribution.is_deleted}:
-            return True
-        if {link for link in self.subcontribution_links if not link.subcontribution.contribution.is_deleted}:
-            return True
+        return (
+            self in chairpersons or
+            # Check regardless of the abstract feature being enabled
+            any(link for link in self.abstract_links if not link.abstract.is_deleted) or
+            any(link for link in self.session_block_links if not link.session_block.session.is_deleted) or
+            any(link for link in self.contribution_links if not link.contribution.is_deleted) or
+            any(link for link in self.subcontribution_links if not link.subcontribution.contribution.is_deleted)
+        )
 
 
 class PersonLinkBase(PersonMixin, db.Model):
--- a/indico/modules/events/models/persons.py
+++ b/indico/modules/events/models/persons.py
@@ -346,6 +346,24 @@ class EventPerson(PersonMixin, db.Model):
                    if ((self.user_id is not None and self.user_id == x.user_id) or
                        (self.email is not None and self.email == x.email)))
 
+    @property
+    def has_links(self):
+        """Whether the person is an author, speaker, etc., in the event."""
+        chairpersons = {link.person for link in self.event.person_links}
+        if self in chairpersons:
+            return True
+        if (
+            self.event.has_feature('abstracts') and
+            {link for link in self.abstract_links if not link.abstract.is_deleted}
+        ):
+            return True
+        if {link for link in self.session_block_links if not link.session_block.session.is_deleted}:
+            return True
+        if {link for link in self.contribution_links if not link.contribution.is_deleted}:
+            return True
+        if {link for link in self.subcontribution_links if not link.subcontribution.contribution.is_deleted}:
+            return True
+
 
 class PersonLinkBase(PersonMixin, db.Model):
     """Base class for EventPerson associations."""
--- a/indico/modules/events/persons/blueprint.py
+++ b/indico/modules/events/persons/blueprint.py
@@ -5,7 +5,8 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
-from indico.modules.events.persons.controllers import (RHEditEventPerson, RHEmailEventPersons, RHEventPersonSearch,
+from indico.modules.events.persons.controllers import (RHDeleteUnusedEventPerson, RHEditEventPerson,
+                                                       RHEmailEventPersons, RHEventPersonSearch,
                                                        RHGrantModificationRights, RHGrantSubmissionRights,
                                                        RHPersonsList, RHRevokeSubmissionRights)
 from indico.web.flask.wrappers import IndicoBlueprint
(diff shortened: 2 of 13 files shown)


indico/modules/events/persons/controllers.py
```diff
@@ -382,6 +385,23 @@ def _process(self):
...
 
+class RHDeleteUnusedEventPerson(RHPersonsBase):
+    def _process_args(self):
+        RHPersonsBase._process_args(self)
+        self.person = EventPerson.query.with_parent(self.event).filter_by(id=request.view_args['person_id']).one()
+
+    def _is_unused(self):
+        persons = self.get_persons()
+        person = persons.get(self.person.email or self.person.id)
+        return not person or person['roles'].get('no_roles', False)
```
ThiefMaster, line 396: I'm not a big fan of this logic here. It's fine to determine whether or not to show the delete option, but for the actual deletion here I'd rather add a `has_links()` method to EventPerson that checks the various relationships backrefs. I think they are all lazy=dynamic so you can use `.has_rows()` on them.

(it'll also help us spotting cases where we are missing some role on this page, because then the deletion would fail with an error and the user may report that)

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -26,7 +26,14 @@
     {% endif %}
     <td class="i-table name-column" data-searchable="{{ person.display_full_name|lower }}">
         {{ person.display_full_name }}
-        {% if person.is_untrusted %}
+        {% if 'no_roles' in person_data.roles %}
+            {% set no_roles_tooltip -%}
+                {%- trans -%}
+                    This person currently has no ties to the event.
+                {%- endtrans -%}
+            {%- endset %}
```
ThiefMaster, line 34: I think this string is short enough (and it has no linebreaks) to just put it directly in the `title="..."`

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -92,5 +99,15 @@
...
         {% endif %}
+        {% if person_data.roles['no_roles'] %}
+            <a class="i-link icon-remove"
+               title="{%- trans %}Delete person{% endtrans -%}"
+               data-href="{{ url_for('persons.delete_unused_person', person) }}"
+               data-title="{%- trans %}Delete person{% endtrans -%}"
+               data-method="DELETE"
+               data-update="#person-{{ person.id }}"
+               data-confirm="{% trans %}Do you really want to delete this person from the event?{% endtrans %}">
+               </a>
```
ThiefMaster, line 110: should be in the previous line

indico/modules/events/persons/blueprint.py
```diff
@@ -25,4 +26,6 @@
 
 # EventPerson operations
 _bp.add_url_rule('/persons/<int:person_id>/edit', 'edit_person', RHEditEventPerson, methods=('GET', 'POST'))
+_bp.add_url_rule('/persons/<int:person_id>/unused/delete', 'delete_unused_person', RHDeleteUnusedEventPerson,
```
ThiefMaster, line 29: I'd remove the `/unused/delete` part - it'd fit if this was a POST request, but for DELETE you usually point to the resource without including a verb/action in the url as well.

indico/modules/events/persons/controllers.py
```diff
@@ -382,6 +385,23 @@ def _process(self):
...
+
+    def _process(self):
+        if self._is_unused():
```
ThiefMaster, line 399: I'd rather raise an exception (e.g. Forbidden) when trying to delete a person that's still used

indico/modules/events/persons/templates/management/_person_list_row.html
```diff
@@ -26,7 +26,10 @@
     {% endif %}
     <td class="i-table name-column" data-searchable="{{ person.display_full_name|lower }}">
         {{ person.display_full_name }}
-        {% if person.is_untrusted %}
+        {% if person_data.roles['no_builtin_roles'] %}
+            <span class="icon-question"
+                  title="{%- trans %}This person currently has no ties to the event.{% endtrans -%}"></span>
+        {% elif person.is_untrusted %}
```
ThiefMaster, line 32: This also sets the `untrusted` class on the `tr` which grays it out a bit. We should probably do the same for unused persons...

indico/modules/events/models/persons.py
```diff
@@ -346,6 +346,24 @@ def has_role(self, role, obj):
...
+            self.event.has_feature('abstracts') and
+            {link for link in self.abstract_links if not link.abstract.is_deleted}
+        ):
+            return True
+        if {link for link in self.session_block_links if not link.session_block.session.is_deleted}:
+            return True
+        if {link for link in self.contribution_links if not link.contribution.is_deleted}:
+            return True
+        if {link for link in self.subcontribution_links if not link.subcontribution.contribution.is_deleted}:
+            return True
```
ThiefMaster, line 365: `any()` will make most of these checks more efficient:

```python
any(link.person == self for link in self.event.person_links)
any(link.person == self for link in self.event.abstract_links if not link.abstract.is_deleted)
...
```

indico/modules/events/models/persons.py
```diff
@@ -346,6 +346,24 @@ def has_role(self, role, obj):
...
                        (self.email is not None and self.email == x.email)))
 
+    @property
+    def has_links(self):
+        """Whether the person is an author, speaker, etc., in the event."""
+        chairpersons = {link.person for link in self.event.person_links}
+        if self in chairpersons:
+            return True
+        if (
+            self.event.has_feature('abstracts') and
```
ThiefMaster, line 356: I wouldn't check that tbh... it's very much an edge case and showing someone as unused even though they are linked to a non-deleted abstract (even though it's not visible unless the feature is re-enabled) feels wrong

indico/modules/events/models/persons.py
```diff
@@ -346,6 +346,24 @@ def has_role(self, role, obj):
...
+            {link for link in self.abstract_links if not link.abstract.is_deleted}
+        ):
+            return True
+        if {link for link in self.session_block_links if not link.session_block.session.is_deleted}:
+            return True
+        if {link for link in self.contribution_links if not link.contribution.is_deleted}:
+            return True
+        if {link for link in self.subcontribution_links if not link.subcontribution.contribution.is_deleted}:
+            return True
+
```
ThiefMaster, line 366: let's add a `return False` at the end; otherwise it returns `None` which is falsy as well but kind of ugly.

Alternative would be something like this:

```py
return (
    any(...) or
    any(...) or
    ...
)
```

Hard to say what's nicer without seeing it though, but since the any()s won't be that huge maybe it's worth a try ;)

indico/modules/events/persons/controllers.py
```diff
@@ -196,9 +198,13 @@ def get_persons(self):
         for reg in regs:
             internal_role_users[reg.user.email]['registrations'].append(reg)
 
-        # Some EventPersons will have no roles since they were connected to deleted things
-        persons = {email: data for email, data in persons.items() if any(data['roles'].values())}
         persons = persons | internal_role_users
+        # Some EventPersons will have no built-in roles since they were connected to deleted things
+        builtin_roles = set(BUILTIN_ROLES.keys())
```
ThiefMaster, line 203: `.keys()` isn't needed here

indico/modules/events/persons/templates/management/_person_list_row.html
ThiefMaster: Nevermind

PR #5320, opened 2022-04-12
Add alarm support to exported iCal.

### LINKED ISSUE #5318: Add alerts to exported calendars
I follow a couple of different sub-groups' meeting schedule by subscribing to their iCal links using my calendaring software. However, the exported links do not have reminders attached to the events, and I often forget to attend meetings.

It would be nice to have either a per-user or per-link ability to attach default notifications to events. For example, I could say "for any calendar I export, please attach an to the iCal export an alert 15 minutes before and 5 minutes before the event begins". Then my calendar software (Apple calendars in my case" will send me a reminder notification.

From my research, it appears that neither Apple calendars nor the iCal format itself support doing this from the client side, so as far as I understand it, this needs to be implemented on the server side.

**Additional context**
I am using the Indico instance at CERN, so it's possible they are using an older version without this software included.

PerilousApricot:
Many (all?) calendar clients do not support adding reminders ("alerts"
in iCal parlance) to external calendars syncrhonized via URL. Add a
per-user preference to attach reminders to any events exported to iCal.

Tested on local dev install both as a logged-in and anonymous user, both
with and without a reminder preference. This produced an iCal that Apple
Calendars properly recognized the reminders:

```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//CERN//INDICO//EN
BEGIN:VEVENT
SUMMARY:testEvent title
DTSTART;VALUE=DATE-TIME:20220425T150000Z
DTEND;VALUE=DATE-TIME:20220425T170000Z
DTSTAMP;VALUE=DATE-TIME:20220412T151800Z
UID:indico-event-8@localhost
DESCRIPTION:testEvent Description\n\nhttp://localhost:8000/event/8/
URL:http://localhost:8000/event/8/
BEGIN:VALARM
ACTION:AUDIO
DESCRIPTION:testEvent Description\n\nhttp://localhost:8000/event/8/
PRODID:-//CERN//INDICO//EN
SUMMARY:testEvent title
TRIGGER:-PT5M
END:VALARM
END:VEVENT
END:VCALENDAR
```

closes #5318

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,7 @@ Improvements
 - Add an option to send notifications of new abstract comments (:issue:`5266`, :pr:`5284`)
 - Badge/poster templates can have additional images besides the background image
   (:pr:`5273`, thanks :user:`SegiNyn`)
+- Add ability to add alerts to iCal exports (:issue:`5318`  :pr:`5320`)
 
 Bugfixes
 ^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,7 +36,7 @@ Improvements
 - Badge/poster templates can have additional images besides the background image
   (:pr:`5273`, thanks :user:`SegiNyn`)
 - Add ability to add alerts to iCal exports (:issue:`5318`, :pr:`5320`, thanks
-  :user: `PerilousApricot`)
+  :user:`PerilousApricot`)
 
 Bugfixes
 ^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,7 +35,8 @@ Improvements
 - Add an option to send notifications of new abstract comments (:issue:`5266`, :pr:`5284`)
 - Badge/poster templates can have additional images besides the background image
   (:pr:`5273`, thanks :user:`SegiNyn`)
-- Add ability to add alerts to iCal exports (:issue:`5318`  :pr:`5320`)
+- Add ability to add alerts to iCal exports (:issue:`5318`, :pr:`5320`, thanks
+  :user: `PerilousApricot`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/events/ical.py
+++ b/indico/modules/events/ical.py
@@ -5,8 +5,8 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
-import datetime
 import typing as t
+from datetime import timedelta
 from email import message
 from email.mime.base import MIMEBase
 from email.policy import compat32
@@ -147,11 +147,10 @@ def generate_event_component(
         component['description'] = data['description']
 
     # If the user exists and the add_icloud_alerts preference isn't blank
-    if user and user.settings.get('add_ical_alerts', '') != '':
+    if user and (alerts := user.settings.get('add_ical_alerts', -1)) != -1:
         alarm = icalendar.Alarm()
-        alarm.add('prodid', '-//CERN//INDICO//EN')
-        alarm.add('action', 'AUDIO')
-        alarm.add('trigger', datetime.timedelta(seconds=-60.0 * float(user.settings.get('add_ical_alerts', ''))))
+        alarm.add('action', 'DISPLAY')
+        alarm.add('trigger', timedelta(minutes=-alerts))
         alarm.add('summary', component['summary'])
         if data['description']:
             alarm.add('description', data['description'])
--- a/indico/modules/events/ical.py
+++ b/indico/modules/events/ical.py
@@ -147,10 +147,10 @@ def generate_event_component(
         component['description'] = data['description']
 
     # If the user exists and the add_icloud_alerts preference isn't blank
-    if user and (alerts := user.settings.get('add_ical_alerts', -1)) != -1:
+    if user and user.settings.get('add_ical_alerts', False):
         alarm = icalendar.Alarm()
         alarm.add('action', 'DISPLAY')
-        alarm.add('trigger', timedelta(minutes=-alerts))
+        alarm.add('trigger', timedelta(minutes=-user.settings.get('add_ical_alerts_mins')))
         alarm.add('summary', component['summary'])
         if data['description']:
             alarm.add('description', data['description'])
--- a/indico/modules/events/ical.py
+++ b/indico/modules/events/ical.py
@@ -5,6 +5,7 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
+import datetime
 import typing as t
 from email import message
 from email.mime.base import MIMEBase
@@ -144,6 +145,17 @@ def generate_event_component(
         data.update(update)
     if data['description']:
         component['description'] = data['description']
+
+    # If the user exists and the add_icloud_alerts preference isn't blank
+    if user and user.settings.get('add_ical_alerts', '') != '':
+        alarm = icalendar.Alarm()
+        alarm.add('prodid', '-//CERN//INDICO//EN')
+        alarm.add('action', 'AUDIO')
+        alarm.add('trigger', datetime.timedelta(seconds=-60.0 * float(user.settings.get('add_ical_alerts', ''))))
+        alarm.add('summary', component['summary'])
+        if data['description']:
+            alarm.add('description', data['description'])
+        component.add_component(alarm)
 
     return component
 
--- a/indico/modules/events/ical.py
+++ b/indico/modules/events/ical.py
@@ -147,7 +147,7 @@ def generate_event_component(
         component['description'] = data['description']
 
     # If the user exists and the add_icloud_alerts preference isn't blank
-    if user and user.settings.get('add_ical_alerts', False):
+    if user and user.settings.get('add_ical_alerts'):
         alarm = icalendar.Alarm()
         alarm.add('action', 'DISPLAY')
         alarm.add('trigger', timedelta(minutes=-user.settings.get('add_ical_alerts_mins')))
--- a/indico/modules/users/__init__.py
+++ b/indico/modules/users/__init__.py
@@ -33,6 +33,7 @@ user_settings = UserSettingsProxy('users', {
     'show_past_events': False,
     'name_format': NameFormat.first_last,
     'use_previewer_pdf': True,
+    'add_ical_alerts': True,
     'synced_fields': None,  # None to synchronize all fields, empty set to not synchronize
     'suggest_categories': False,  # whether the user should receive category suggestions
 }, converters={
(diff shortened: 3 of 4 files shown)


indico/modules/users/forms.py
```diff
@@ -70,6 +70,16 @@ class UserPreferencesForm(IndicoForm):
         widget=SwitchWidget(),
         description=_('The previewer is used by default for image and text files, but not for PDF files.'))
 
+    add_ical_alerts = StringField(
+        _('Add alerts to iCal'),
+        description=_('Add an alert to exported iCal files/URLs. ' +
+                      'Value should be number of minutes to notify before an event. ' +
```
ThiefMaster, line 76: Remove the `+` - strings can be concatenated without it, and i18n extraction will most likely fail with the `+` since that makes it a runtime operation (while `'foo' 'bar'` is combined at parse time and thus not affecting the AST)

indico/modules/users/__init__.py
```diff
@@ -33,6 +33,7 @@
     'show_past_events': False,
     'name_format': NameFormat.first_last,
     'use_previewer_pdf': True,
+    'add_ical_alerts': True,
```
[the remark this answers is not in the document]
ThiefMaster: Thinking about this again the empty string vs 0 is a bit of mess. Maybe better to use `0` for disabled? Or do you think people want a reminder *right* when the event starts? `-1` would work as well in that case, but otherwise I'd prefer `0`

indico/modules/users/forms.py
```diff
@@ -70,6 +70,16 @@ class UserPreferencesForm(IndicoForm):
...
+                      'A blank field disables this function.'))
+
+    def validate_add_ical_alerts(self, field):
+        if field.data != '' and not field.data.isnumeric():
+            raise ValidationError(_('Alert time must be either a number or blank (to disable).'))
```
ThiefMaster, line 81: no longer needed with the suggestion above

indico/modules/events/ical.py
```diff
@@ -145,6 +146,17 @@ def generate_event_component(
     if data['description']:
         component['description'] = data['description']
 
+    # If the user exists and the add_icloud_alerts preference isn't blank
+    if user and user.settings.get('add_ical_alerts', '') != '':
```
ThiefMaster, line 150: When going for 0 for disabled you can simplify this to:

```py
if user and (alerts := user.settings.get('add_ical_alerts')):
```

indico/modules/events/ical.py
```diff
@@ -145,6 +146,17 @@ def generate_event_component(
...
+        alarm = icalendar.Alarm()
+        alarm.add('prodid', '-//CERN//INDICO//EN')
+        alarm.add('action', 'AUDIO')
+        alarm.add('trigger', datetime.timedelta(seconds=-60.0 * float(user.settings.get('add_ical_alerts', ''))))
```
ThiefMaster, line 154: ...and this can then be simplified to `timedelta(minutes=-alerts)`

indico/modules/events/ical.py
```diff
@@ -145,6 +146,17 @@ def generate_event_component(
     if data['description']:
         component['description'] = data['description']
 
+    # If the user exists and the add_icloud_alerts preference isn't blank
+    if user and user.settings.get('add_ical_alerts', '') != '':
+        alarm = icalendar.Alarm()
+        alarm.add('prodid', '-//CERN//INDICO//EN')
```
ThiefMaster, line 152: is this needed? at least in one of our plugins we don't add it and it seems to work

CHANGES.rst
```diff
@@ -35,6 +35,7 @@ Improvements
 - Add an option to send notifications of new abstract comments (:issue:`5266`, :pr:`5284`)
 - Badge/poster templates can have additional images besides the background image
   (:pr:`5273`, thanks :user:`SegiNyn`)
+- Add ability to add alerts to iCal exports (:issue:`5318`  :pr:`5320`)
```
ThiefMaster, line 38:
```suggestion
- Add ability to add alerts to iCal exports (:issue:`5318`, :pr:`5320`, thanks
  :user:`PerilousApricot`)
```

indico/modules/events/ical.py
```diff
@@ -5,6 +5,7 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
+import datetime
```
ThiefMaster, line 8: `from datetime import timedelta` and remove the `datetime.` below when using it

indico/modules/users/__init__.py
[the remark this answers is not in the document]
PerilousApricot: I think people will want a reminder when the event starts (if this proposal would support two alarms, I'd personally set it to -15 mins and then another at 0 mins). I can change the logic to be -1 instead of blank string as you propose if you'd like.

indico/modules/users/forms.py
PerilousApricot: Ah, OK, got a pre-commit complaint about the line being too long. I didn't realize that about string concatenation

indico/modules/events/ical.py
PerilousApricot: That was just cargo-culting some other code I saw nearby. Not sure if it's needed or not

indico/modules/users/__init__.py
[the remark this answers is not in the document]
ThiefMaster: the thing is that defaulting to enabled would affect all existing users and likely annoy a good chunk of them.

if 0 min should be possible, i suggest we go for two options: a bool to enable it, and a number to set the duration in minutes (using the HiddenUnless validator to hide it unless alarms are enabled in the form)

indico/modules/events/ical.py
```diff
@@ -145,6 +146,17 @@ def generate_event_component(
     if data['description']:
         component['description'] = data['description']
 
+    # If the user exists and the add_icloud_alerts preference isn't blank
+    if user and user.settings.get('add_ical_alerts', '') != '':
+        alarm = icalendar.Alarm()
+        alarm.add('prodid', '-//CERN//INDICO//EN')
+        alarm.add('action', 'AUDIO')
```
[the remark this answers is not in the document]
PerilousApricot: At least for the Apple calendar client, it treats both the same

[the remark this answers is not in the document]
PerilousApricot: added change

indico/modules/events/ical.py
PerilousApricot: removed

indico/modules/users/__init__.py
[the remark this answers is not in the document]
PerilousApricot: It appears to work with -1 as the default in the new revision

PerilousApricot:
@ThiefMaster I've done your suggestion and split the preference into two parts, it is commit cf265cfe13. I don't know how to do the hidden field thing you talk about, though, so it's not yet complete

CHANGES.rst
```diff
@@ -35,6 +35,8 @@ Improvements
 - Add an option to send notifications of new abstract comments (:issue:`5266`, :pr:`5284`)
 - Badge/poster templates can have additional images besides the background image
   (:pr:`5273`, thanks :user:`SegiNyn`)
+- Add ability to add alerts to iCal exports (:issue:`5318`, :pr:`5320`, thanks
+  :user: `PerilousApricot`)
```
ThiefMaster, line 39:
```suggestion
  :user:`PerilousApricot`)
```

not sure how that spaces ended up there (in case it was part of my original suggestion..) 👾

indico/modules/events/ical.py
```diff
@@ -145,6 +146,16 @@ def generate_event_component(
     if data['description']:
         component['description'] = data['description']
 
+    # If the user exists and the add_icloud_alerts preference isn't blank
+    if user and user.settings.get('add_ical_alerts', False):
```
ThiefMaster, line 150:
```suggestion
    if user and user.settings.get('add_ical_alerts'):
```

the default is coming from the SettingsProxy definition if omitted

indico/modules/users/forms.py
```diff
@@ -70,6 +70,16 @@ class UserPreferencesForm(IndicoForm):
...
         description=_('The previewer is used by default for image and text files, but not for PDF files.'))
 
+    add_ical_alerts = BooleanField(
+        _('Add alerts to iCal'),
+        widget=SwitchWidget(),
+        description=_('Add an event reminder to exported iCal files/URLs.'))
+
+    add_ical_alerts_mins = IntegerField(
+        _('Timeout for iCal alerts'),
+        [Optional(), NumberRange(min=0)],
```
ThiefMaster, line 80:
```suggestion
        [HiddenUnless('add_ical_alerts'), NumberRange(min=0)],
```

indico/modules/users/forms.py
```diff
@@ -70,6 +70,16 @@ class UserPreferencesForm(IndicoForm):
         widget=SwitchWidget(),
         description=_('The previewer is used by default for image and text files, but not for PDF files.'))
 
+    add_ical_alerts = BooleanField(
+        _('Add alerts to iCal'),
+        widget=SwitchWidget(),
+        description=_('Add an event reminder to exported iCal files/URLs.'))
+
+    add_ical_alerts_mins = IntegerField(
+        _('Timeout for iCal alerts'),
```
ThiefMaster, line 79: Not sure if "Timeout" is the best word here... Can't think of a good alternative though, "delay" doesn't really fit either...

ThiefMaster:
thanks! i added the HiddenUnless in my review comment; this should be all that's needed - the rest is magic ✨ that happens automatically when this validator is used

PerilousApricot:
OK @ThiefMaster! I think I have this fully implemented in 31386b8db0

indico/modules/users/forms.py
```diff
@@ -70,6 +70,16 @@ class UserPreferencesForm(IndicoForm):
         widget=SwitchWidget(),
         description=_('The previewer is used by default for image and text files, but not for PDF files.'))
 
+    add_ical_alerts = BooleanField(
+        _('Add alerts to iCal'),
+        widget=SwitchWidget(),
+        description=_('Add an event reminder to exported iCal files/URLs.'))
+
+    add_ical_alerts_mins = IntegerField(
+        _('Calendar notification time in minutes'),
```
ThiefMaster, line 79:
```suggestion
        _('iCal notification time'),
```

That's a bit shorter (maybe even short enough to not trigger wrapping, but often hard to avoid..), and the description is a better place to put anything lengthy

ThiefMaster:
Thanks, looks good! I'll test it tomorrow - already way too late anyway :D

PerilousApricot:
@ThiefMaster thanks! have a good evening :)

ThiefMaster (approved):
Sorry, took a bit longer with the easter holidays etc. Looks all good and works fine as well. Thanks!

PerilousApricot:
Thanks a lot!--
It's dark in this basement.

Issue #5312, opened 2022-04-04
Lightweight registrations / accompanying persons


duartegalvao:
This is a new field type; the settings should be:

- max number of people to add in a registration
- the standard places limit (one person = place)

When adding a new accompanying person we should only ask for basic data (first+last name), nothing else. We don't want to add fancy field customization there (for now).

Storage would be just a JSON array containing objects in the normal field data.

ThiefMaster:
I think @javfg started working on this, wasn't he?

PR #5321, opened 2022-04-19
Lightweight registrations / accompanying persons

javfg:
This PR adds a registration field for accompanying persons. It closes https://github.com/indico/indico/issues/5312.

It includes only basic first/last name data. The people added this way can count towards the registration limit or not, and a maximum number of accompanying persons per registrant can be set too.

image

Tasks remaining:
- [x] Send remaining places in conference to the frontend.
- [ ] Properly add ids to the data coming from the backend.

### DIFF
--- a/indico/modules/events/registration/client/js/form/fields/AccompanyingPersonsInput.jsx
+++ b/indico/modules/events/registration/client/js/form/fields/AccompanyingPersonsInput.jsx
@@ -8,6 +8,7 @@
 import {nanoid} from 'nanoid';
 import PropTypes from 'prop-types';
 import React, {useState} from 'react';
+import {useFormState} from 'react-final-form';
 import {useSelector} from 'react-redux';
 import {Button, Form, Label} from 'semantic-ui-react';
 
@@ -15,7 +16,7 @@ import {FinalCheckbox, FinalField, FinalInput, validators as v} from 'indico/rea
 import {FinalModalForm} from 'indico/react/forms/final-form';
 import {Translate} from 'indico/react/i18n';
 
-import {getCurrency} from '../selectors';
+import {getCurrency, getItems} from '../selectors';
 
 import {PlacesLeft} from './PlacesLeftLabel';
 
@@ -55,10 +56,63 @@ AccompanyingPersonModal.defaultProps = {
   },
 };
 
-function AccompanyingPersonsComponent({value, disabled, onChange, price, availablePlaces}) {
+// Gather all accompanying persons field's counts.
+function countAllAccompanyingPersons(items, formState) {
+  const allAccompanyingPersonFieldNames = Object.keys(items)
+    .map(i => items[i])
+    .filter(f => f.inputType === 'accompanying_persons' && f.personsCountAgainstLimit)
+    .map(apf => apf.htmlName);
+  return allAccompanyingPersonFieldNames.reduce(
+    (count, field) => count + formState.values[field]?.length || 0,
+    0
+  );
+}
+
+function calculatePlaces(availablePlaces, maxPersons, personsInCurrentField, items, formState) {
+  if (availablePlaces === null) {
+    // Field does not count towards registration limit...
+    if (!maxPersons) {
+      // ...and has no person limit.
+      return [null, Infinity];
+    } else {
+      // ...and has a person limit.
+      return [personsInCurrentField, maxPersons];
+    }
+  } else {
+    // Field counts towards registration limit...
+    const personsInAllFieldsCount = countAllAccompanyingPersons(items, formState);
+
+    if (!maxPersons || maxPersons >= availablePlaces - personsInAllFieldsCount) {
+      // ...and has no person limit, or its person limit is greater than the registration limit.
+      return [personsInAllFieldsCount, availablePlaces];
+    } else {
+      // ...and has a person limit lower than the regitration limit.
+      return [personsInCurrentField, maxPersons];
+    }
+  }
+}
+
+function AccompanyingPersonsComponent({
+  value,
+  disabled,
+  onChange,
+  price,
+  availablePlaces,
+  maxPersons,
+}) {
   const [operation, setOperation] = useState({type: null, person: null});
   const currency = useSelector(getCurrency);
   const totalPrice = (value.length * price).toFixed(2);
+  const items = useSelector(getItems);
+  const formState = useFormState();
+
+  const [placesUsed, placesLimit] = calculatePlaces(
+    availablePlaces,
+    maxPersons,
+    value.length,
+    items,
+    formState
+  );
 
   // Add ids for persons coming from the backend (when editing a registration).
   value = value.map(p => (p.id ? p : {id: nanoid(), ...p}));
@@ -123,7 +177,7 @@ function AccompanyingPersonsComponent({value, disabled, onChange, price, availab
           size="small"
           type="button"
           onClick={handleAccompanyingPersonAdd}
-          disabled={disabled || value.length >= availablePlaces}
+          disabled={disabled || placesLimit - placesUsed === 0}
         >
           <Translate>Add accompanying person</Translate>
         </Button>
@@ -132,13 +186,9 @@ function AccompanyingPersonsComponent({value, disabled, onChange, price, availab
             {price.toFixed(2)} {currency} (Total: {totalPrice} {currency})
           </Label>
         )}
-        {availablePlaces !== null && (
+        {placesLimit !== Infinity && (
           <div styleName="places-left">
-            <PlacesLeft
-              placesLimit={availablePlaces}
-              placesUsed={value.length}
-              isEnabled={!disabled}
-            />
+            <PlacesLeft placesLimit={placesLimit} placesUsed={placesUsed} isEnabled={!disabled} />
           </div>
         )}
       </div>
@@ -172,15 +222,23 @@ AccompanyingPersonsComponent.propTypes = {
   onChange: PropTypes.func.isRequired,
   price: PropTypes.number,
   availablePlaces: PropTypes.number,
+  maxPersons: PropTypes.number,
 };
 
 AccompanyingPersonsComponent.defaultProps = {
   disabled: false,
   price: 0,
   availablePlaces: null,
+  maxPersons: null,
 };
 
-export default function AccompanyingPersonsInput({htmlName, disabled, price, availablePlaces}) {
+export default function AccompanyingPersonsInput({
+  htmlName,
+  disabled,
+  price,
+  availablePlaces,
+  maxPersons,
+}) {
   return (
     <FinalField
       name={htmlName}
@@ -188,6 +246,7 @@ export default function AccompanyingPersonsInput({htmlName, disabled, price, ava
       disabled={disabled}
       price={price}
       availablePlaces={availablePlaces}
+      maxPersons={maxPersons}
     />
   );
 }
@@ -197,12 +256,14 @@ AccompanyingPersonsInput.propTypes = {
   disabled: PropTypes.bool,
   price: PropTypes.number,
   availablePlaces: PropTypes.number,
+  maxPersons: PropTypes.number,
 };
 
 AccompanyingPersonsInput.defaultProps = {
   disabled: false,
   price: 0,
   availablePlaces: null,
+  maxPersons: null,
 };
 
 export function AccompanyingPersonsSettings() {
--- a/indico/modules/events/registration/client/js/form/fields/AccompanyingPersonsInput.jsx
+++ b/indico/modules/events/registration/client/js/form/fields/AccompanyingPersonsInput.jsx
@@ -1,0 +1,250 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2022 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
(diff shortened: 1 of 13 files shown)


indico/modules/events/registration/fields/simple.py
```diff
@@ -371,3 +371,50 @@ def _indico_email(value):
                 raise ValidationError(_('Invalid email address'))
 
         return _indico_email
+
+
+class AccompanyingPersonsField(RegistrationFormBillableField):
+    name = 'accompanying_persons'
+    mm_field_class = fields.List(fields.Nested({
+        'first_name': fields.String(required=True),
+        'last_name': fields.String(required=True),
```
ThiefMaster, line 380: This will still allow empty strings (required just means it has to be present); add `validate=not_empty`

indico/modules/events/registration/fields/simple.py
```diff
@@ -371,3 +371,50 @@ def _indico_email(value):
...
+    }))
+    setup_schema_base_cls = LimitedPlacesBillableFieldDataSchema
+    setup_schema_fields = {
+        'max_persons': fields.Integer(load_default=0, validate=validate.Range(0)),
+        'persons_count_against_limit': fields.Bool(load_default=False),
```
ThiefMaster, line 385: this doesn't seem to be checked anywhere on the server side

indico/modules/events/registration/client/js/form/fields/AccompanyingPersonsInput.jsx
```diff
@@ -0,0 +1,246 @@
...
+  };
+
+  const handleModalClose = () => {
+    setOperation({type: null, person: null});
+  };
+
+  return (
+    <Form.Group styleName="accompanyingpersons-field">
+      <ul>
+        {!value.length && <li styleName="light">No accompanying persons registered</li>}
```
ThiefMaster, line 104: i18n

indico/modules/events/registration/fields/simple.py
```diff
@@ -391,41 +391,31 @@ def default_value(self):
...
                 if not old_data or not self.has_data_changed(new_data, old_data):
                     return True
-            if new_data and (max_persons := self._get_max_persons()):
-                places_left = max_persons - self.get_places_used()
-                if not existing_registration and self.form_item.data.get('persons_count_against_limit'):
-                    # TODO remove this once get_places_used's TODO is done
-                    places_left -= 1
-                if new_data and not places_left:
-                    raise ValidationError(_('There are no places left for this option.'))
+            if (new_data
```
ThiefMaster, line 402: Maybe this logic deserves a unit test... some cases to cover:

- new registration
- updating registration, field untouched
- updating registration, field touched but same/lower count
- updating registration, field touched and count increased (this should fail if limit exceeded)

indico/modules/events/registration/models/registrations.py
```diff
@@ -825,3 +829,13 @@ def _set_checked_in_dt(target, value, *unused):
...
+
+    query = (select([db.func.sum(db.func.jsonb_array_length(RegistrationData.data)) + 1])
+             .where(db.and_((RegistrationData.registration_id == Registration.id),
+                            (RegistrationData.field_data_id == RegistrationFormFieldData.id),
+                            (RegistrationFormFieldData.field_id == RegistrationFormItem.id),
+                            (RegistrationFormItem.input_type == 'accompanying_persons'),
+                            db.cast(RegistrationFormItem.data['persons_count_against_limit'], db.Boolean)))
+             .correlate_except(RegistrationData)
+             .scalar_subquery())
+    Registration.occupied_slots = column_property(query, deferred=True)
```
ThiefMaster, line 841: Just to make sure the query is correct, use the db log util to have a look at the query that gets executed when:

- accessing the property on a Registration instance
- querying a registration (or multiple) with `.options(undefer('occupied_slots'))`

A unit test is probably a good idea as well :)

indico/modules/events/registration/templates/display/_registration_summary_blocks.html
```diff
@@ -70,7 +70,7 @@
 
 {% macro render_data(registration, field, from_management=false) %}
     {% set data = registration.data_by_field -%}
-    {% set friendly_data = data[field.id].friendly_data if field.id in data else '' -%}
+    {% set friendly_data = data[field.id].get_friendly_data(for_humans=True) if field.id in data else '' -%}
```
ThiefMaster, line 73: lowercase `true` (same for `false` and `none`) in templates (even though both work - lowercase is preferred)

indico/modules/events/registration/fields/simple.py
```diff
@@ -371,3 +371,64 @@ def _indico_email(value):
                 raise ValidationError(_('Invalid email address'))
 
         return _indico_email
+
+
+class AccompanyingPersonsField(RegistrationFormBillableField):
```
ThiefMaster, line 376: I think this isn't a "simple" field... maybe something for a separate file (e.g. `accompanying.py`)?

indico/modules/events/registration/fields/simple.py
```diff
@@ -371,3 +371,64 @@ def _indico_email(value):
...
+    name = 'accompanying_persons'
+    mm_field_class = fields.List
+    mm_field_kwargs = {'cls_or_instance': fields.Nested({
```
ThiefMaster, line 379: Not sure what to think about this... I'd treat an argument with such a name as positional-only even if it technically isn't (yet?)

A nicer alternative would be to add `mm_field_args` for positional args as well

indico/modules/events/registration/util.py
```diff
@@ -137,6 +138,9 @@ def get_flat_section_submission_data(regform, *, management=False, registration=
             continue
         if registration and isinstance(item.field_impl, (ChoiceBaseField, AccommodationField)):
             field_data = get_field_merged_options(item, registration_data)
+        elif registration and isinstance(item.field_impl, AccompanyingPersonsField):
+            field_data = item.view_data
+            field_data['availablePlaces'] = item.field_impl.get_available_places(False)
```
ThiefMaster, line 143: I *think* we camelize the dict later, so if that's indeed the case you should be able to use the `snake_cased` version here.

indico/modules/events/registration/fields/simple.py
duartegalvao: Yeah that would be useful, since this also happens in the `AccommodationField`

indico/modules/events/registration/fields/accompanying_test.py
```diff
@@ -0,0 +1,243 @@
...
+            'persons_count_against_limit': persons_count_against_limit,
+        }
+        field.versioned_data = field.field_impl.form_item.data
+        if registration:
+            RegistrationData(
+                registration_id=registration.id,
+                field_data=field.current_data,
+                data=data if data is not None else _create_accompanying_persons(no_of_persons)
+            )
+            db.session.expire(registration)
```
duartegalvao, line 49: This was a workaround to force the `registration.data` cache to be reset. Not sure if this is the best solution...

indico/modules/events/registration/client/js/form/fields/AccompanyingPersonsInput.jsx
```diff
@@ -0,0 +1,291 @@
...
+    id: null,
+    firstName: null,
+    lastName: null,
+  },
+};
+
+// Gather all accompanying persons field's counts.
+function countAllAccompanyingPersons(items, formState) {
+  const allAccompanyingPersonFieldNames = Object.keys(items)
+    .map(i => items[i])
```
ThiefMaster, line 62: `Object.values(items)` does that already..

indico/modules/events/registration/templates/display/_registration_summary_blocks.html
```diff
@@ -183,7 +183,7 @@
                                         {{ choice }} ({{ nights }} nights)
                                     {%- endtrans %}
                                 {% elif item.field_data.field.input_type == 'multi_choice' %}
-                                    {{ item.friendly_data | join(', ') }}
```
ThiefMaster, line 186: Where is this handled now?

indico/modules/events/registration/models/forms.py
```diff
@@ -490,13 +491,13 @@ def get_personal_data_field_id(self, personal_data_type):
 
 @listens_for(orm.mapper, 'after_configured', once=True)
 def _mappers_configured():
-    query = (select([db.func.count(Registration.id)])
+    query = (select([db.func.coalesce(db.func.sum(Registration.occupied_slots), 0)])
```
ThiefMaster, line 494: How can the sum be NULL? Shouldn't it at least be 1 for any registration? Or does this happen when there are no registrations yet?

indico/modules/events/registration/fields/choices.py
```diff
@@ -440,7 +440,7 @@ class AccommodationField(RegistrationFormBillableItemsField):
     versioned_data_fields = RegistrationFormBillableField.versioned_data_fields | {'choices'}
     setup_schema_base_cls = AccommodationSetupSchema
     mm_field_class = fields.Nested
-    mm_field_kwargs = {'nested': AccommodationSchema}
+    mm_field_args = (AccommodationSchema, )
```
ThiefMaster, line 443:
```suggestion
    mm_field_args = (AccommodationSchema,)
```

I think we do it without the space everywhere else

indico/modules/events/registration/fields/base.py
```diff
@@ -30,6 +30,8 @@ class RegistrationFormFieldBase:
     versioned_data_fields = frozenset({'price'})
     #: the marshmallow field class for regform submission
     mm_field_class = None
+    #: required arguments for the marshmallow field
```
ThiefMaster, line 33:
```suggestion
    #: positional arguments for the marshmallow field
```

indico/modules/events/registration/fields/accompanying_test.py
```diff
@@ -0,0 +1,242 @@
...
+from indico.modules.events.registration.models.registrations import Registration, RegistrationData, RegistrationState
+
+
+pytest_plugins = 'indico.modules.events.registration.testing.fixtures'
+
+
+@pytest.fixture
+def create_accompanying_persons_field(db, dummy_regform):
+    def _create_accompanying_persons_field(max_persons, persons_count_against_limit,
+                                           registration=None, data=None, no_of_persons=0):
```
ThiefMaster, line 24: I'd replace `no_of` with `num` everywhere :)

indico/modules/events/registration/fields/accompanying_test.py
ThiefMaster: The reason why you need this is a very common mistake with sqlalchemy relationships: Instead of using the relationships, you use the FKs, but FK columns almost never need to be assigned a value directly. If you used `registration=registration` instead of `registration_id=registration.id` a few lines above, it would most likely work as expected.

Or, a way that's even cleaner IMHO:

```py
registration.data = RegistrationData(...)
```

In that case you don't have to pass the registration as a kwarg at all, because assigning it to the other side of the relationship will do everything that's needed. And it's also very clear that you are setting the registration's data here, while with the `RegistrationData()` that's not assigned to anything it's much less explicit.

indico/modules/events/registration/models/forms.py
duartegalvao: Yep, it's when there are no registrations

indico/modules/events/registration/fields/accompanying_test.py
ThiefMaster:
```py
        if registration:
            registration.data.append(RegistrationData(
                field_data=field.current_data,
                data=(data if data is not None else _create_accompanying_persons(num_persons))
            ))
            db.session.flush()
```

indico/modules/events/registration/client/js/form/fields/AccompanyingPersonsInput.jsx
```diff
@@ -0,0 +1,288 @@
...
+    items,
+    formState
+  );
+
+  const changeReducer = action => {
+    switch (action.type) {
+      case 'ADD':
+        return [...value, {id: `new:${nanoid()}`, ...action.person}];
+      case 'EDIT':
+        return [...value.map(p => (p.id === action.person.id ? action.person : p))];
```
ThiefMaster, line 121:
```suggestion
        return value.map(p => (p.id === action.person.id ? action.person : p));
```

.map() already creates a new array. no need to clone it again

indico/modules/events/registration/client/js/form/fields/AccompanyingPersonsInput.jsx
```diff
@@ -0,0 +1,288 @@
...
+      case 'REMOVE':
+        return [...value.filter(p => p.id !== action.id)];
```
ThiefMaster, line 123:
```suggestion
        return value.filter(p => p.id !== action.id);
```

same here

indico/modules/events/registration/templates/display/_registration_summary_blocks.html
ThiefMaster: This change does not seem to be correct:

image

javfg: This is a mistake (I thought about handling the accompanying persons in this else, but I changed it back and broke multichoice. Reverting it!

PR #5322, opened 2022-04-19
Fix poster and badge printing on unlisted events

javfg:
This PR fixes a breaking category check when printing a poster or badge in an unlisted event.

### DIFF
--- a/indico/modules/designer/placeholders.py
+++ b/indico/modules/designer/placeholders.py
@@ -170,7 +170,7 @@ class CategoryTitlePlaceholder(DesignerPlaceholder):
 
     @classmethod
     def render(cls, event):
-        return event.category.title
+        return event.category.title if not event.is_unlisted else ''
 
 
 class RegistrationFullNamePlaceholder(FullNamePlaceholderBase):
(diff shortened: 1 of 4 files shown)


indico/modules/designer/placeholders.py
```diff
@@ -170,7 +170,7 @@ class CategoryTitlePlaceholder(DesignerPlaceholder):
 
     @classmethod
     def render(cls, event):
-        return event.category.title
+        return event.category.title if not event.is_unlisted else _('Unlisted event')
```
ThiefMaster, line 173: Just wondering... wouldn't an empty string be better in this case? I don't think you'd ever want "Unlisted event" printed on a badge/poster

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -415,7 +415,8 @@ def _check_access(self):
 
         # Check that template belongs to this event or a category that
         # is a parent
-        if self.template.owner != self.event and self.template.owner.id not in self.event.category_chain:
+        if (self.template.owner != self.event and
+                (not self.event.is_unlisted and self.template.owner.id not in self.event.category_chain)):
```
ThiefMaster, line 419: shouldn't this be "unlisted OR ..."? it looks like right now you can use any template if you are in an unlisted event.

or even better: allow templates from the root category; those should always be available even for an unlisted event

indico/modules/designer/placeholders.py
javfg: Makes sense!

indico/modules/events/registration/controllers/management/reglists.py
javfg: Removed the if and made it a bit clearer.

ThiefMaster (approved):

PR #5324, opened 2022-04-19
add sending signal for folder cloning

vtran99:
Request signal sent to inform about new folder creation by cloning.
It would be preferable if we have a public "clone_attachment_folder" which returns the new folder cloned.

### DIFF
--- a/indico/modules/attachments/clone.py
+++ b/indico/modules/attachments/clone.py
@@ -5,6 +5,7 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
+import flask
 from sqlalchemy.orm import joinedload, subqueryload
 
 from indico.core import signals
@@ -94,6 +95,7 @@ class AttachmentCloner(EventCloner):
         folder.populate_from_attrs(old_folder, folder_attrs)
         folder.acl_entries = clone_principals(AttachmentFolderPrincipal, old_folder.acl_entries,
                                               self._event_role_map, self._regform_map)
+        signals.attachments.folder_created.send(folder, user=flask.session.user, cloned=folder)
         for old_attachment in old_folder.attachments:
             self._attachment_map[old_attachment] = attachment = Attachment(folder=folder)
             attachment.populate_from_attrs(old_attachment, attachment_attrs)


indico/modules/attachments/clone.py
```diff
@@ -5,6 +5,7 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
+import flask
```
ThiefMaster, line 8: `from flask import session`

indico/modules/attachments/clone.py
```diff
@@ -94,6 +95,7 @@ def _clone_attachment_folder(self, old_folder, new_object):
         folder.populate_from_attrs(old_folder, folder_attrs)
         folder.acl_entries = clone_principals(AttachmentFolderPrincipal, old_folder.acl_entries,
                                               self._event_role_map, self._regform_map)
+        signals.attachments.folder_created.send(folder, user=flask.session.user, cloned=folder)
```
ThiefMaster, line 98: This `cloned=folder` arg is strange in general; and calling this signal would result e.g. in the cloning being logged to the event log (which adds noise we do not want).

ThiefMaster:
Can you elaborate on the purpose of this? I didn't check the code but I don't think we typically trigger the various creation signals when cloning an event...

indico/modules/attachments/clone.py
vtran99: or you can keep it private like it's, and adding return folder at the end...

ThiefMaster: Would it help to have a `_folder_map` (like there's `_attachment_map`)? Then you could just register your own internal cloner that depends on this cloner and access the mapping between old and new folders..

vtran99: Not really, because we need the same folder cloning as the core (cloned with attachments and attributes...).
It's just that we added a new foreign key to our folder object, and as the original cloning skipped the FK, we need to get back the new folder clones to update that FK.

ThiefMaster: If `return folder` at the end of `_clone_attachment_folder` is sufficient for you then I suggest to just do that... I think that's better than adding a (somewhat obscure) signal call here.

vtran99: That will be fine for us. Thanks.
I will update the PR.

PR #5330, opened 2022-04-22
Include affiliations in author lists and abstract exports

plourenco:
Includes the affiliation near the author list in relevant listings, in the form of `Name (Affiliation)`.
Also clones the contributions' feature of exporting affiliations in abstracts.

RFC: Shall we get rid of the excessive amount of options ("with affiliations") and just add an extra column for it by default? Seems simpler to work around it within the CSV, having "Authors" + "Authors with affiliation".

imagem

### DIFF
--- a/indico/modules/events/abstracts/templates/management/_abstract_list.html
+++ b/indico/modules/events/abstracts/templates/management/_abstract_list.html
@@ -146,7 +146,7 @@
                                         <td class="i-table" data-searchable="{{ abstract.submitter.display_full_name }}">
                                             {{- abstract.submitter.display_full_name }}
                                             {% if abstract.submitter.affiliation %}
-                                                <span class="text-muted">({{ abstract.submitter.affiliation }})</span>
+                                                <span class="text-superfluous">({{ abstract.submitter.affiliation }})</span>
                                             {% endif %}
                                         </td>
                                     {% elif item.id == 'authors' %}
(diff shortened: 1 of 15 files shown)


indico/web/client/styles/base/_typography.scss
```diff
@@ -228,6 +228,10 @@ td.groupSubTitle {
   color: $red;
 }
 
+.text-muted {
+  color: $gray;
+}
```
ThiefMaster, line 233: There's already `.text-superfluous` with the same color..

indico/modules/events/abstracts/templates/management/_abstract_list.html
```diff
@@ -144,13 +144,19 @@
...
                                         </td>
                                     {% elif item.id == 'authors' %}
                                         <td class="i-table person-row-cell"
                                             data-searchable="{{ abstract.primary_authors | join(', ', attribute='display_full_name') | lower }}">
                                             {% for author in abstract.primary_authors | sort(attribute='display_order_key_lastname') -%}
-                                                <div class="person-row icon-user">{{ author.display_full_name }}</div>
+                                                <div class="person-row inline icon-user">{{ author.display_full_name }}</div>
+                                                {% if author.affiliation %}
+                                                    <span class="text-superfluous">({{ author.affiliation }})</span>
+                                                {% endif %}
```
ThiefMaster, line 159: This should be inside the person-row, which shouldn't be inline

PR #5335, opened 2022-05-02
Add missing IP network identifier

plourenco:


ThiefMaster:
I think with the new property you could also make use of it inside `serialize_ip_network_group`. Also I'd rebase this PR to 3.1.x

indico/web/forms/fields/principals.py
```diff
@@ -152,7 +152,8 @@ class PermissionsField(JSONField):
     def __init__(self, *args, **kwargs):
         self.object_type = kwargs.pop('object_type')
         super().__init__(*args, **kwargs)
-        self.ip_networks = list(map(serialize_ip_network_group, IPNetworkGroup.query.filter_by(hidden=False)))
+        ip_schema = IPNetworkGroupSchema()
+        self.ip_networks = list(map(ip_schema.dump, IPNetworkGroup.query.filter_by(hidden=False)))
```
ThiefMaster, line 156: untested but this should do the same in a cleaner way:

```py
self.ip_networks = IPNetworkGroupSchema(many=True).dump(IPNetworkGroup.query.filter_by(hidden=False).all())
```

ThiefMaster (approved):

PR #5340, opened 2022-05-10
Fix precedence of webpack alias paths

plourenco:
Fixes an issue causing path alias such as `indico/modules/events` to take precedence over sub-paths as in `indico/modules/events/foo`, causing the latter not to be resolved.


webpack.config.babel.js
```diff
@@ -40,24 +40,28 @@ const extraResolveAliases = [
...
-  if (module.produceBundle) {
-    entryPoints[moduleName] = [dirName];
-  }
-  const modulePath = path.join('indico/modules', module.parent || '', module.name);
-  extraResolveAliases.push({name: modulePath, alias: dirName, onlyModule: false});
+    if (module.produceBundle) {
+      entryPoints[moduleName] = [dirName];
+    }
+    const modulePath = path.join('indico/modules', module.parent || '', module.name);
+    // extraResolveAliases[modulePath] = dirName;
```
ThiefMaster, line 56: 👀

Issue #5342, opened 2022-05-12
Error when user double clicks "Create my Indico Profile" when creating new account 


SegiNyn:
**Describe the bug**
A database error occurs when a user clicks on the "Create my Indico Profile" button more than once 

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to 'http://127.0.0.1:9000/register/'
2. Input an email address to get verification link
3. Click on the link sent and input user details
4. When done, double click on "Create my Indico Profile" 
5. See error 

**Expected behaviour**
The action should not be called again when the button is clicked for the second time, we could disable the button after the first click. 

**Screenshots**
<img width="1379" alt="Screenshot 2022-05-12 at 14 22 58" src="https://user-images.githubusercontent.com/54508387/168075219-a933616f-1b46-48c2-9b5b-5ee5f996a922.png">

ThiefMaster:
Maybe users shouldn't doubleclick the button. ;)

But you're right, disabling it seems like a good idea.

SegiNyn:
Okay, I'll open a PR for it.

PR #5343, opened 2022-05-12
Disable create profile button after submit

SegiNyn:
Issue: https://github.com/indico/indico/issues/5342


indico/modules/auth/templates/register.html
```diff
@@ -55,8 +55,15 @@
...
+                <input id="create-profile-button" class="i-button big highlight" type="submit"
+                       value="{% trans %}Create my Indico profile{% endtrans %}">
             {% endif %}
         {% endcall %}
     {% endif %}
+    <script>
+        $('#register-form').submit(function (){
+            $('#create-profile-button').attr('disabled', true);
+            return true;
+        });
```
ThiefMaster, line 67:
```suggestion
        $('#register-form').on('submit', () => {
            $('#create-profile-button').prop('disabled', true);
        });
```

- don't use  the legacy shorthands
- I don't think the return is needed
- disabled is a property, not an attribute

indico/modules/auth/templates/register.html
```diff
@@ -60,9 +60,11 @@
             {% endif %}
         {% endcall %}
     {% endif %}
-    <script>
-        $('#register-form').on('submit', () => {
-            $('#create-profile-button').prop('disabled', true);
-        });
-    </script>
+    {% block javascript %}
```
ThiefMaster, line 63: why?

PR #5344, opened 2022-05-16
Small improvements for the maintenance scripts

andriish:
The update is related only to the build scripts.
 - use explicit name of python executable x 2
 - assure the current directory is in sys.path
 
Best regards,

Andrii


ThiefMaster:
The changes look fine, just wondering why you needed them. With a proper virtualenv setup, the correct `python` should already be in the $PATH...

andriish:
Hi @ThiefMaster ,

I think we spoke about that once. But at some point, I had an idea to package indico in RPMs to make the installation of the whole ecosystem just with `dnf install python-indico` and configure everything with  `dnf install myconfig-indico` or so.
So I did that. But found some really minor issues in indico.

Best regards,

Andrii

andriish:
The name of python interpreter in my case is `python3` as I do not use the virtual environment

andriish:
Hi @ThiefMaster ,

would you merge this one?

Best regards,

Andrii

ThiefMaster:
The interpreter change looks fine. What's the reason for the sys.path change?

andriish:
Hi @ThiefMaster ,

Could you, please, have a look?
so far I've reverted the changes to with sys.path.
It makes sense to create a separate MR if I remember why that was needed.


Andrii

PR #5345, opened 2022-05-16
Upgrade text editor to ckeditor5

plourenco:
- [x] ~~Support markdown and/or convert all output to HTML~~
- [x] Move width/height properties to CSS
- [x] Provide a simple version
- [x] Attribute changes (e.g. font colors) do not modify `disabled_until_change`


indico/web/client/js/jquery/widgets/jinja/ckeditor_widget.js
```diff
@@ -5,37 +5,24 @@
...
+      writer.setStyle('width', `${width}px` || 'auto', editor.editing.view.document.getRoot());
+      writer.setStyle('height', `${height}px`, editor.editing.view.document.getRoot());
     });
-    $E(options.fieldId + '-editor').set(editor.draw());
-    // Re-position the dialog if we have one since the initial position is
-    // wrong due to the editor being loaded after the dialog has been opened.
-    var dialog = field.closest('.ui-dialog-content');
-    if (dialog.length) {
-      dialog.dialog('option', 'position', dialog.dialog('option', 'position'));
-    }
```
ThiefMaster, line 39: I think this is still needed - with your PR the "add minutes" dialog isn't properly centered anymore after opening it

indico/web/client/js/ckeditor.js
```diff
@@ -0,0 +1,108 @@
...
+      'bulletedList',
+      'numberedList',
+      'outdent',
+      'indent',
+      '|',
+      'link',
+      !simple && images && 'imageInsert',
+      !simple && 'insertTable',
+      '|',
+      !simple && 'blockQuote',
```
ThiefMaster, line 38: Not showing this in normal mode feels a bit strange (and in the old one we had it - test e.g. in "add minutes")

indico/web/client/js/ckeditor.js
```diff
@@ -0,0 +1,108 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2022 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+export const getConfig = ({simple = true, images = false} = {}) => ({
+  language: 'en_GB',
```
ThiefMaster, line 9: Can't we take this from the currently used locale instead?

indico/web/client/js/jquery/widgets/jinja/ckeditor_widget.js
```diff
@@ -5,37 +5,24 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
-/* global RichTextEditor:false, $E:false */
+import ClassicEditor from 'ckeditor';
```
ThiefMaster, line 8: Add this to your vscode workspace settings so you get warnings even for files that are in one of the blacklisted folders:

```json
		"eslint.options": {
			"ignore": false,
		},
```

indico/web/client/js/jquery/widgets/jinja/ckeditor_widget.js
```diff
@@ -5,37 +5,24 @@
...
-    var editor = new RichTextEditor(600, options.height, options.simple, options.images);
-    editor.set(field.val());
-    editor.onLoad(function() {
-      editor.onChange(function() {
-        field.val(editor.get()).trigger('change');
-      });
+  global.setupCKEditorWidget = async function setupCKEditorWidget(options) {
+    const {fieldId, images = false, width, height = 475, ...rest} = options;
+    const field = document.getElementById(fieldId);
+    const editor = await ClassicEditor.create(field, {...getConfig({images}), ...rest});
```
ThiefMaster, line 17: `simple` is missing in the `getConfig` call here

ThiefMaster:
Can't we get the editor to use two rows of buttons instead of the `...` menu? Not sure if users will realize some options they may have used before are now in there...

indico/web/client/js/ckeditor.js
```diff
@@ -0,0 +1,108 @@
...
+      'removeFormat',
+      '|',
+      'bulletedList',
+      'numberedList',
+      'outdent',
+      'indent',
+      '|',
+      'link',
+      !simple && images && 'imageInsert',
+      !simple && 'insertTable',
```
ThiefMaster, line 36: When adding a table I get a raw `<figure class="table"></figure>` in the displayed minutes. Probably this isn't whitelisted in our bleach config. Let's make sure that anything the editor inserts is on the HTML whitelist (for this one it seems perfectly safe to allow) to avoid ugly HTML showing up

plourenco:
> Can't we get the editor to use two rows of buttons instead of the `...` menu? Not sure if users will realize some options they may have used before are now in there...

Well, we can, but it looks a bit more cluttered.

indico/web/client/js/jquery/widgets/jinja/ckeditor_widget.js
plourenco: Yep, thought I could get away with css

ThiefMaster:
Would be nice to see screenshots of the two options for a direct comparison ;)

ThiefMaster:
OK, if there are half-filled toolbars it certainly looks weird...

indico/web/client/js/ckeditor.js
ThiefMaster: I just checked and we currently don't set a locale either (and nobody ever complained) so fine for me.

indico/web/client/js/jquery/widgets/jinja/ckeditor_widget.js
```diff
@@ -5,37 +5,38 @@
...
+    });
+    editor.editing.view.change(writer => {
+      writer.setStyle('width', `${width}px` || 'auto', editor.editing.view.document.getRoot());
+      writer.setStyle('height', `${height}px`, editor.editing.view.document.getRoot());
     });
-    $E(options.fieldId + '-editor').set(editor.draw());
     // Re-position the dialog if we have one since the initial position is
     // wrong due to the editor being loaded after the dialog has been opened.
-    var dialog = field.closest('.ui-dialog-content');
+    const dialog = $(field.closest('.ui-dialog-content'));
```
ThiefMaster, line 28: `,closest()` already returns a jquery object, no need for this `$()` around it

indico/web/client/js/jquery/widgets/jinja/ckeditor_widget.js
```diff
@@ -5,37 +5,38 @@
...
     if (dialog.length) {
       dialog.dialog('option', 'position', dialog.dialog('option', 'position'));
     }
+    // Make sure the option dropdowns are displayed above the dialog.
+    const uiDialog = field.closest('.ui-dialog-content');
+    const exPopup = field.closest('.exclusivePopup');
+    if (uiDialog) {
```
ThiefMaster, line 35: `.length`

indico/web/client/js/jquery/widgets/jinja/ckeditor_widget.js
```diff
@@ -5,37 +5,38 @@
...
+      uiDialog.style.overflow = 'inherit';
```
ThiefMaster, line 36: that won't do anything since you set it on the jquery object. you need `uiDialog[0]` to get the DOM node, or just use `.style('overflow', 'inherit')` since it's using jQuery anyway

indico/web/client/js/jquery/widgets/jinja/ckeditor_widget.js
plourenco: I'm no longer using jquery here. `document.querySelector().closest` seems to be supported in most modern browsers.

https://caniuse.com/#feat=element-closest

The reason why I'm wrapping `$()` is because we intend to call a jquery dialog afterwards, `$.dialog`. I don't think there's another way to call it.

ThiefMaster: oh, i wasn't aware `field` isn't a jquery object.. nevermind then, also on my other comments i guess

indico/web/client/js/jquery/widgets/jinja/ckeditor_widget.js
plourenco: This should work due to the reason above.

indico/web/client/js/react/components/TextEditor.jsx
```diff
@@ -0,0 +1,36 @@
...
+  );
+}
+
+TextEditor.propTypes = {
+  width: PropTypes.number,
+  height: PropTypes.number,
+};
+
+TextEditor.defaultProps = {
+  height: '400px',
```
ThiefMaster, line 34: that's not a number.... so probably it should be `PropTypes.string` or a combination of both

plourenco: Yep, I ended up fixing this in a child PR, need to cherry-pick from there.

PR #5346, opened 2022-05-17
Bugreport+ suggestion to fix identities.py

andriish:
The idea is that ipv6 adresses can contain %, e.g. fe80::a00:27ff:fe22:a069%enp0s3
This will break SQL syntax.

Best regards,
Andrii

### DIFF
--- a/indico/modules/auth/models/identities.py
+++ b/indico/modules/auth/models/identities.py
@@ -14,6 +14,7 @@ from indico.core.db import db
 from indico.core.db.sqlalchemy import UTCDateTime
 from indico.util.date_time import as_utc, now_utc
 from indico.util.passwords import PasswordProperty
+import re
 
 
 class Identity(db.Model):
@@ -96,7 +97,7 @@ class Identity(db.Model):
     def register_login(self, ip):
         """Update the last login information."""
         self.last_login_dt = now_utc()
-        self.last_login_ip = ip.replace('%','\%')
+        self.last_login_ip = re.split("%",ip)[0]
 
     def __repr__(self):
         return f'<Identity({self.id}, {self.user_id}, {self.provider}, {self.identifier})>'
--- a/indico/modules/auth/models/identities.py
+++ b/indico/modules/auth/models/identities.py
@@ -96,7 +96,7 @@ class Identity(db.Model):
     def register_login(self, ip):
         """Update the last login information."""
         self.last_login_dt = now_utc()
-        self.last_login_ip = ip
+        self.last_login_ip = ip.replace('%','\%')
 
     def __repr__(self):
         return f'<Identity({self.id}, {self.user_id}, {self.provider}, {self.identifier})>'


ThiefMaster:
Umm how does it break the SQL statement there?

- This is not a LIKE criterion where `%` would be a wildcard
- SQLAlchemy passes all the data as parameters, so nothing needs to be escaped in values
- (Aren't those IP addresses link-local? Why would a web application be accessed with such an IP address to begin with? :o)

ThiefMaster:
OK I just checked and it looks like the INET Postgres type does not support link-local IPv6 with a scope identifier:

```
indico=> select 'fe80::a00:27ff:fe22:a069'::inet;
           inet
--------------------------
 fe80::a00:27ff:fe22:a069
(1 row)

indico=> select 'fe80::a00:27ff:fe22:a069%enp0s3'::inet;
ERROR:  invalid input syntax for type inet: "fe80::a00:27ff:fe22:a069%enp0s3"
LINE 1: select 'fe80::a00:27ff:fe22:a069%enp0s3'::inet;
               ^
indico=> select 'fe80::a00:27ff:fe22:a069\%enp0s3'::inet;
ERROR:  invalid input syntax for type inet: "fe80::a00:27ff:fe22:a069\%enp0s3"
LINE 1: select 'fe80::a00:27ff:fe22:a069\%enp0s3'::inet;
```

So your change won't help; the only solution would be changing this to a text type. However, I'm not open to that unless there's a proper usecase where indico needs to be accessed using link-local IP addresses...

FWIW what I'd be open to is stripping the scope identifier from the IPv6, but even then I think there should be a reason for having to support these addresses to begin with! (It's very weird for a webapp to *not* be accessed through global-scoped IPv6 addresses only)

andriish:
Hi @ThiefMaster ,
 - I just see in the logs
 ```
[SQL: UPDATE users.identities SET last_login_dt=%(last_login_dt)s, last_login_ip=%(last_login_ip)s WHERE users.identities.id = %(users_identities_id)s]
[parameters: {'last_login_dt': datetime.datetime(2022, 5, 17, 9, 55, 56, 708329), 'last_login_ip': 'fe80::a00:27ff:fe22:a069%enp0s3', 'users_identities_id': 1}]
(Background on this error at: https://sqlalche.me/e/14/9h9h)
  File "/usr/lib64/python3.10/site-packages/sqlalchemy/engine/base.py", line 1819, in _execute_context
    self.dialect.do_execute(
  File "/usr/lib64/python3.10/site-packages/sqlalchemy/engine/default.py", line 732, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type inet: "fe80::a00:27ff:fe22:a069%enp0s3"
LINE 1: ...-05-17T09:55:56.708329'::timestamp, last_login_ip='fe80::a00...
                                                             ^
```
- I don't know that. I'm more or less literate only with sqlite in C.
- Yes, it is. I use it for debug.

Best regards,

Andrii

andriish:
> So your change won't help

Yes, I see it now. In any case this is more a bugreport than a proper PR.

> FWIW what I'd be open to is stripping the scope identifier from the IPv6, but even then I think there should be a reason for having to support these addresses to begin with! (It's very weird for a webapp to not be accessed through global-scoped IPv6 addresses only)

That is fine.  For the debug reasons I just cut off the offending part, but I think it makes sense to report such things.

PR #5358, opened 2022-05-31
Change privacy sidemenu item icon

duartegalvao:
This PR changes the privacy sidemenu item icon from a lock to a balance scale, and adds the ability to use SUI icons in sidemenu items.

screenshot

### DIFF
--- a/indico/web/menu.py
+++ b/indico/web/menu.py
@@ -73,7 +73,9 @@ class SideMenuItem:
 
     is_section = False
 
-    def __init__(self, name, title, url, weight=-1, active=False, disabled=False, section=None, icon=None, badge=None):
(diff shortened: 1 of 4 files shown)


indico/web/menu.py
```diff
@@ -82,6 +83,7 @@ def __init__(self, name, title, url, weight=-1, active=False, disabled=False, se
         self.section = section
         self.weight = weight
         self.icon = ('icon-' + icon) if icon else None
+        self.sui_icon = sui_icon
```
ThiefMaster, line 86: Might be worth adding an `assert not (self.icon and self.sui_icon)` to prevent having both on the same item?

PR #5363, opened 2022-06-10
Fix selection in abstract list options

plourenco:
Fixes an issue with selection on integer-based abstract list options.

**How to reproduce:**
Select "State: submitted" (to remove the withdrawn ones) and then "Submitted for track: <anything>" the first filter is used but visually reset.


indico/modules/events/abstracts/lists.py
```diff
@@ -88,6 +88,9 @@ def _get_filters_from_request(self):
                 options = [x if x != 'None' else None for x in request.form.getlist(f'field_{field.id}')]
                 if options:
                     filters['fields'][str(field.id)] = options
+        # Ensure enum filters remain as integers
+        for idx, value in enumerate(filters['items'].get('state', [])):
+            filters['items']['state'][idx] = int(value)
```
ThiefMaster, line 93: The fact that we always add a state filter here - even if it didn't exist before - isn't a problem?

ThiefMaster: What I meant is adding `filters['items']['state']` even if it wasn't there before. Empty list could mean "nothing", but I didn't look more closely at the code. Just wanted to bring up something to check. So if it all works fine no need to change anything

PR #5374, opened 2022-06-20
Replace selectize with SUI's remote <Dropdown>

tomasr8:
Still a work in progress but most of it already works,
just want to add some QoL improvements.

- [x] New react component based on SUI's `<Dropdown>` + new jinja widget
- [x] Highlight matching query in search results
- [x] Better `noResultsMessage`
- [x] Disable `minTriggerLength` when `preload=True`?
- [x] Remove selectize.js as a dependency

### DIFF
--- a/indico/modules/rb/event/client/js/index.js
+++ b/indico/modules/rb/event/client/js/index.js
@@ -15,7 +15,7 @@ $(document).ready(() => {
     const $fieldContainer = $target.closest('.searchable-field');
     const objectId = $target.val();
     const $bookBtn = $fieldContainer.find('.js-book-btn');
-    const linkType = $fieldContainer.data('link-type');
+    const linkType = $target.data('link-type');
     let params = {};
     if (objectId) {
       const values = $fieldContainer.data('values')[objectId];
--- a/indico/modules/rb/templates/booking_list.html
+++ b/indico/modules/rb/templates/booking_list.html
@@ -46,12 +46,13 @@
                 </div>
             </div>
             {% if has_contribs %}
-                <div class="section searchable-field" data-values="{{ contribs_data | tojson | forceescape }}" data-link-type="contribution">
+                <div class="section searchable-field" data-values="{{ contribs_data | tojson | forceescape }}">
                     <div class="text">
                         <div class="label">
                             {%- trans %}Contribution{% endtrans -%}
                         </div>
-                        {{ form_row(form.contribution, skip_label=true, hide_description=true) }}
+                        {{ form_row(form.contribution, skip_label=true, hide_description=true,
+                                    widget_attrs={'data_link_type': 'contribution'}) }}
                     </div>
                     <div class="toolbar">
                         <a href="#" target="_blank" class="i-button highlight js-book-btn disabled">
@@ -61,12 +62,13 @@
                 </div>
             {% endif %}
             {% if has_session_blocks %}
-                <div class="section searchable-field" data-values="{{ session_blocks_data | tojson | forceescape }}" data-link-type="sessionBlock">
+                <div class="section searchable-field" data-values="{{ session_blocks_data | tojson | forceescape }}">
                     <div class="text">
                         <div class="label">
                             {%- trans %}Session block{% endtrans -%}
                         </div>
-                        {{ form_row(form.session_block, skip_label=true, hide_description=true) }}
+                        {{ form_row(form.session_block, skip_label=true, hide_description=true,
+                                    widget_attrs={'data_link_type': 'sessionBlock'}) }}
                     </div>
                     <div class="toolbar">
                         <a href="#" target="_blank" class="i-button highlight js-book-btn disabled">
--- a/indico/web/forms/widgets.py
+++ b/indico/web/forms/widgets.py
@@ -179,7 +179,7 @@ class RemoteDropdownWidget(JinjaWidget):
                                 search_payload=getattr(field, 'search_payload', None),
                                 min_trigger_length=self.min_trigger_length, preload=self.preload,
                                 allow_by_id=self.allow_by_id, value_field=self.value_field,
-                                label_field=self.label_field, search_field=self.search_field, dropdown_props=kwargs)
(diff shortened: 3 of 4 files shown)


indico/web/client/js/react/components/index.js
```diff
@@ -54,3 +54,4 @@ export {default as WTFSUIColorPicker} from './WTFSUIColorPicker';
 export {default as ICSCalendarLink} from './ICSCalendarLink';
 export {default as Favorite} from './Favorite';
 export {default as PersonDetailsModal} from './PersonDetailsModal';
+export {default as WTFRemoteSearchDropdown} from './WTFRemoteSearchDropdown';
```
ThiefMaster, line 57: i'd move it next to the other WTF components? ;)

indico/web/client/js/jquery/widgets/jinja/sui_remote_search_dropdown_widget.js
```diff
@@ -0,0 +1,26 @@
...
+  ...props
+}) {
+  // Make sure the dropdown is displayed above the dialog
+  const field = $(`#${fieldId}`);
+  field.closest('.ui-dialog-content').css('overflow', 'inherit');
+  field.closest('.exclusivePopup').css('overflow', 'inherit');
+
+  ReactDOM.render(
+    <WTFRemoteSearchDropdown fieldId={fieldId} {...props} />,
+    document.getElementById(fieldId)
```
ThiefMaster, line 24: you an use `field[0]` here

indico/web/client/js/jquery/widgets/jinja/sui_remote_search_dropdown_widget.js
```diff
@@ -0,0 +1,26 @@
...
+  fieldId,
+  ...props
+}) {
+  // Make sure the dropdown is displayed above the dialog
+  const field = $(`#${fieldId}`);
+  field.closest('.ui-dialog-content').css('overflow', 'inherit');
+  field.closest('.exclusivePopup').css('overflow', 'inherit');
+
+  ReactDOM.render(
+    <WTFRemoteSearchDropdown fieldId={fieldId} {...props} />,
```
ThiefMaster, line 23: since you need all the props including field it, matbe just accept `props` in the function in line 13 and use `props.fieldId` in line 18

indico/web/client/js/react/components/WTFRemoteSearchDropdown.jsx
```diff
@@ -0,0 +1,280 @@
...
+const highlightSearch = (text, query = '') => {
+  const index = text.toLowerCase().indexOf(query);
+  if (!query || index === -1) {
+    return text;
+  }
+
+  return (
+    <>
+      {text.slice(0, index)}
+      <b>{text.slice(index, index + query.length)}</b>
```
ThiefMaster, line 26: `<mark>` is probably more appropriate for highlighting than `<b>`

indico/web/client/js/react/components/WTFRemoteSearchDropdown.jsx
```diff
@@ -0,0 +1,280 @@
...
+}) {
+  const [searchQuery, setSearchQuery] = useState('');
+  const [options, setOptions] = useState([]);
+  const [loading, setLoading] = useState(false);
+
+  const isControlled = valueFromProps !== undefined;
+  const [internalValue, setInternalValue] = useState(null);
+  const value = isControlled ? valueFromProps : internalValue;
+
+  const debounce = useMemo(() => makeAsyncDebounce(200), []);
```
ThiefMaster, line 62: I'd keep this outside the component. memoizing works, but elsewhere we just define it outside the function - less code that needs to run during renders for sure ;)

indico/web/client/js/react/components/WTFRemoteSearchDropdown.jsx
```diff
@@ -0,0 +1,280 @@
...
+
+      const opts = data.map(item => ({
+        'key': item[valueField],
+        'value': item[valueField],
+        'text': item[labelField],
+        // extra values passed to <Dropdown.Item>
+        'search': item[searchField].toLowerCase(),
+        'friendly-id': item.friendly_id,
+      }));
+      return _.sortBy(opts, 'value');
```
ThiefMaster, line 105: Is sorting by value correct? If this is for displaying the list of choices, sorting by text is likely the better choice (the value is probably an id, and thus not very meaningful for the user - it may not even be displayed anywhere)

indico/web/client/js/react/components/WTFRemoteSearchDropdown.jsx
```diff
@@ -0,0 +1,280 @@
...
+  const onSearch = (evt, {searchQuery: newSearchQuery}) => {
+    setSearchQuery(newSearchQuery);
+  };
+
+  const searchDisabledMessage =
+    minTriggerLength === 0 ? (
+      undefined
+    ) : allowById ? (
+      <PluralTranslate count={minTriggerLength}>
+        <Singular>Search requires at least one character or an id prefixed with #</Singular>
```
ThiefMaster, line 183: It's a good idea to use the count param in singular as well instead of writing "one" - languages like chinese have only one form (ie no separate plural form) and IIRC it causes some issues if the singular form doesn't contain the placeholder

indico/web/client/js/react/components/WTFRemoteSearchDropdown.jsx
```diff
@@ -0,0 +1,280 @@
...
+      defaultOpen={false}
+      openOnFocus={false}
+      value={value}
+      options={options}
+      onChange={onChange}
+      search={filterOptions}
+      searchQuery={searchQuery}
+      onSearchChange={onSearch}
+      noResultsMessage={
+        loading ? Translate.string('Loading..') : searchDisabled ? searchDisabledMessage : undefined
```
ThiefMaster, line 214: `...` or unicode ellipsis `…`? ;)

PR #5389, opened 2022-06-24
Add setting for mandatory Affiliation and Comment fields

### LINKED ISSUE #4819: Mandatory fields "Affiliation" and "Comment" in the account request form
Dear Indico team,

here is a tiny feature request for your consideration.

The problem:

In some cases one needs to enable a registration
for extrnal users. The users, however, too often ignore the fields
"Affiliation" and "Comment" even if they are explicitely asked to fill
them. This creates a situation when the only known information about
the user is the e-mail/name. It hard to decide based on that
information only if one should accept/reject the requst, especially in
the cases when the e-mail adress is from a commercial site.

Possible solution:

Mandaratory fields "Affiliation" and "Comment" in the account request
form would help to get more information for the decision making.

Alternatives:

The solution above + make it possible for indico admins to select if these fields are mandatory

example

Bet regards,

Andrii

elsbethe:
Add a setting for making the 'Affiliation' and 'Comment' field mandatory when creating a new account. When setting is active, the corresponding field is mandatory in the sign up form.

closes #4819

### DIFF
--- a/indico/modules/auth/client/js/signup.jsx
+++ b/indico/modules/auth/client/js/signup.jsx
@@ -197,6 +197,7 @@ function Signup({
               <FinalTextArea
                 required={mandatoryFields.includes('comment')}
                 name="comment"
+                initialValue=""
                 label={Translate.string('Comment')}
                 description={Translate.string(
                   'You can provide additional information or a comment for the administrators who will review your registration.'
--- a/indico/modules/auth/client/js/signup.jsx
+++ b/indico/modules/auth/client/js/signup.jsx
@@ -55,7 +55,10 @@ function Signup({
     try {
       resp = await indicoAxios.post(location.href, values);
     } catch (e) {
-      return handleSubmitError(e);
+      return handleSubmitError(
+        e,
+        hasPredefinedAffiliations ? {affiliation: 'affiliation_data'} : {}
+      );
     }
     location.href = resp.data.redirect;
     // never finish submitting to avoid fields being re-enabled
@@ -134,6 +137,7 @@ function Signup({
             {hasPredefinedAffiliations ? (
               <SyncedFinalAffiliationDropdown
                 name="affiliation_data"
+                required={moderated && mandatoryFields.includes('affiliation')}
                 syncName="affiliation"
                 syncedValues={syncedValues}
                 currentAffiliation={affiliationMeta}
@@ -232,7 +236,7 @@ Signup.propTypes = {
   emails: PropTypes.arrayOf(PropTypes.string).isRequired,
   affiliationMeta: PropTypes.object,
   hasPendingUser: PropTypes.bool,
-  mandatoryFields: PropTypes.arrayOf(PropTypes.string),
+  mandatoryFields: PropTypes.arrayOf(PropTypes.string).isRequired,
 };
 
 Signup.defaultProps = {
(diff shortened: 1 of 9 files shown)


indico/modules/auth/client/js/signup.jsx
```diff
@@ -141,6 +142,7 @@ function Signup({
               <SyncedFinalInput
                 name="affiliation"
                 label={Translate.string('Affiliation')}
+                required={ moderated && mandatoryFields.includes("Affiliation") }
```
ThiefMaster, line 145: please make sure to use eslint so it fixes formatting issues (through its *prettier* integration)

indico/modules/auth/client/js/signup.jsx
```diff
@@ -189,8 +191,9 @@ function Signup({
                 </Translate>
               </Message>
               <FinalTextArea
+                required={ mandatoryFields.includes("Comment") }
```
ThiefMaster, line 194: use all lowercase for internal values

indico/modules/users/forms.py
```diff
@@ -116,6 +116,10 @@ class AdminUserSettingsForm(IndicoForm):
                                          description=_('Whether users are allowed to generate personal API tokens. '
                                                        'If disabled, only admins can create them, but users will '
                                                        'still be able to regenerate the tokens assigned to them.'))
+    if config.data['LOCAL_MODERATION']:
+        mandatory_fields_account_request = IndicoSelectMultipleCheckboxField(_('Mandatory fields in account request'),
+                                                                         choices=['Affiliation','Comment'],
```
ThiefMaster, line 121: I'm pretty sure you can pass a list of tuples here to separate value/title (value should be lowercase, title should be translatable)

ThiefMaster (commented):
Thanks! Please also add server-side validation to make sure the mandatory fields are not empty.

elsbethe:
Thanks for your quick feedback! I added the server-side validation and processed your feedback.

indico/modules/users/forms.py
```diff
@@ -116,6 +116,10 @@ class AdminUserSettingsForm(IndicoForm):
                                          description=_('Whether users are allowed to generate personal API tokens. '
                                                        'If disabled, only admins can create them, but users will '
                                                        'still be able to regenerate the tokens assigned to them.'))
+    if config.data['LOCAL_MODERATION']:
```
ThiefMaster, line 119: - `config.LOCAL_MODERATION`
- the indentation/spacing below is incorrect

indico/modules/auth/controllers.py
```diff
@@ -435,11 +435,17 @@ class Meta:
             first_name = fields.String(required=True)
             last_name = fields.String(required=True)
             address = fields.String(load_default='')
-            affiliation = fields.String(load_default='')
+            if 'affiliation' in user_management_settings.get('mandatory_fields_account_request'):
```
ThiefMaster, line 438: Only apply this if `self.moderate_registrations` is true

indico/modules/auth/controllers.py
```diff
@@ -435,11 +435,17 @@ class Meta:
...
+                affiliation = fields.String(required=True)
```
ThiefMaster, line 439: `validate=not_empty` would be a good idea here as well (otherwise an empty string would be accepted)

ThiefMaster (approved):

PR #5398, opened 2022-06-28
Fix bug in some fields validation

duartegalvao:
This PR fixes a bug in which fields added after a registration was made skipped server-side validation.


indico/modules/events/registration/fields/accompanying.py
```diff
@@ -54,7 +54,7 @@ def get_validators(self, existing_registration):
         def _check_number_of_places(new_data):
             if existing_registration:
                 old_data = existing_registration.data_by_field.get(self.form_item.id)
-                if not old_data or len(new_data) <= len(old_data.data):
+                if old_data and len(new_data) <= len(old_data.data):
                     return True
```
ThiefMaster, line 58: I'm pretty sure we don't need a return value in any of these validation functions. they raise in case of an error, and otherwise the return value isn't meaningful

PR #5400, opened 2022-06-29
Add basic captcha support for registration submissions

tomasr8:
Adds captcha for regform submission, though the code is generic enough to be used
in any form. The built-in captcha can also be overridden with plugins.

- [x] Basic captcha working
- [x] Enable/disable in regform settings
- [x] Fix weird bug with incorrect audio length
- [x] Make it plugin friendly

image


indico/migrations/versions/20220707_1007_b60f5c45acf7_add_option_to_enable_captcha.py
```diff
@@ -0,0 +1,25 @@
+"""Add option to registration forms to enable CAPTCHA for users without an account
```
ThiefMaster, line 1: Way too verbose for a revision title ;)

"Add require_captcha to regforms" is shorter and more search-friendly since it contains the column name.

indico/migrations/versions/20220707_1007_b60f5c45acf7_add_option_to_enable_captcha.py
```diff
@@ -0,0 +1,25 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'b60f5c45acf7'
+down_revision = '0c4bb2973536'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('forms', sa.Column('require_captcha', sa.Boolean(), nullable=False, server_default='true'), schema='event_registration')
```
ThiefMaster, line 20: Default should be false, we don't want to enable this all of a sudden for existing regforms.

indico/modules/events/registration/plugins.py
```diff
@@ -0,0 +1,21 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2022 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+class CaptchaPluginMixin:
```
ThiefMaster, line 8: I would not put this (or any other generic captcha code not specific to registrations) in the registration module. Captchas may be useful for Indico account creation as well.

indico/modules/events/registration/fields/simple.py
```diff
@@ -374,3 +375,16 @@ def _indico_email(value):
...
+class CaptchaField(fields.String):
+    """Validates an answer against a CAPTCHA stored in the session."""
+
+    def __init__(self):
+        super().__init__(required=True, validate=not_empty)
+
+    def _validate(self, user_answer):
+        super()._validate(user_answer)
+        answer = session.get('captcha_answer', None)
+        if not answer or answer != user_answer:
```
ThiefMaster, line 389: This needs to be a plugin method - in case of reCaptcha or any of the other "fancy" ones (where you have to align a puzzle tile or do similar things) you typically need to call an external API to verify the result.

indico/modules/events/registration/client/js/form_submission/Captcha.jsx
```diff
@@ -0,0 +1,124 @@
...
+  const [error, setError] = useState(false);
+  const [playing, setPlaying] = useState(false);
+  const [image, setImage] = useState(null);
+  const [audio, setAudio] = useState(null);
+
+  const fetchCaptcha = () => {
+    setLoading(true);
+    indicoAxios
+      .get(generateCaptcha())
+      .then(({data}) => {
```
ThiefMaster, line 31: hmm why not use async/await?

indico/modules/events/registration/client/js/form_submission/Captcha.jsx
```diff
@@ -0,0 +1,124 @@
...
+        setError(false);
+      })
+      .catch(() => setError(true))
+      .finally(() => setLoading(false));
+  };
+  useEffect(fetchCaptcha, []);
+
+  const toggleAudio = () => {
+    if (playing) {
+      audio.pause();
```
ThiefMaster, line 45: hm... do you really want to pause here or rather stop (so it starts from the beginning when you click play again)?

indico/modules/events/registration/controllers/display.py
```diff
@@ -302,6 +302,10 @@ def _process_args(self):
         if self.invitation and self.invitation.state == InvitationState.accepted and self.invitation.registration:
             return redirect(url_for('.display_regform', self.invitation.registration.locator.registrant))
 
+    def _captcha_required(self):
```
ThiefMaster, line 305: maybe make this a `@property`?

indico/modules/events/registration/plugins.py
tomasr8: Yeah I was debating where best to put this..

indico/modules/events/registration/client/js/form_submission/RegistrationFormSubmission.jsx
```diff
@@ -85,13 +91,14 @@ ConsentToPublish.propTypes = {
...
   const sections = useSelector(getNestedSections);
-  const {submitUrl, registrationData, initialValues} = useSelector(getStaticData);
+  const {submitUrl, registrationData, initialValues, captchaRequired} = useSelector(getStaticData);
   const isUpdateMode = useSelector(getUpdateMode);
   const isModerated = useSelector(getModeration);
   const isManagement = useSelector(getManagement);
   const publishToParticipants = useSelector(getPublishToParticipants);
   const publishToPublic = useSelector(getPublishToPublic);
   const showConsentToPublish = !isManagement && publishToParticipants !== 'hide_all';
+  const hasCaptchaPlugin = getPluginObjects('regformCaptcha').length > 0;
```
ThiefMaster, line 101: You could do `const pluginCaptcha = renderPluginComponents('regformCaptcha');` and use that below. In the end it just returns a list as well so you can use that to check the length.

indico/modules/events/registration/client/js/form_submission/index.jsx
```diff
@@ -32,6 +32,7 @@ export default function setupRegformSubmission(root) {
     publishToParticipants,
     publishToPublic,
     consentToPublish = null,
+    captchaRequired = false,
```
ThiefMaster, line 35: I'm surprised this works - i'd have expected `JSON.parse(false)` to fail or at cause log a warning...

indico/modules/events/registration/plugins.py
```diff
@@ -0,0 +1,34 @@
...
+    #: the readable name of the CAPTCHA plugin
+    friendly_name = None
+
+    def generate_captcha(self):
+        """Returns a CAPTCHA and the correct answer.
```
ThiefMaster, line 13:
```suggestion
        """Return a CAPTCHA and the correct answer.
```

indico/modules/events/registration/plugins.py
```diff
@@ -0,0 +1,34 @@
...
+        CAPTCHA on the client-side.
+        As an example, the built-in CAPTCHA generator
+        returns a dictionary with a base64 encoded image and audio data.
+        For reCAPTCHA, the `answer` may be `None` as the validation is done externally and
+        the `data` will contain the Site key needed to render the reCAPTCHA widget.
+        """
+        raise NotImplementedError('CAPTCHA plugin must implement generate_captcha()')
+
+    def validate_captcha(self, answer):
+        """Validates the answer to a CAPTCHA.
```
ThiefMaster, line 27:
```suggestion
        """Validate the answer to a CAPTCHA.
```

indico/modules/core/captcha.py
```diff
@@ -0,0 +1,48 @@
...
+        def validate(user_answer):
+            if plugin := get_captcha_plugin():
+                if not plugin.validate_captcha(user_answer):
+                    raise ValidationError(_('Incorrect answer'))
+            else:
+                answer = session.get('captcha_state')
+                if not answer or answer != user_answer:
+                    raise ValidationError(_('Incorrect answer'))
+
+        super().__init__(required=True, validate=validate)
```
ThiefMaster, line 29: I think you are supposed to raise those errors from within `_deserialize`: https://marshmallow.readthedocs.io/en/stable/custom_fields.html#creating-a-field-class

ThiefMaster (approved):

Issue #5381, opened 2022-06-22
Web Content Accessibility Guidelines (WCAG) 2.1


bxo11:
These guidelines provide information and requirements for Web apps to be more accessible to people with disabilities. Indico needs some corrections/changes to fulfill these requirements.

Example changes:
- Possibility of navigating by keyboard only
  - Increased visibility of focused element (color/frame)
  - No keyboard traps/blocking elements (currently some text editors are blocking focus)
  - Custom order of interactive elements for better keyboard navigation
- Better contrast of all text elements
- Display description for small icons on mouse hover/keyboard focus
- A uniform system of informing about errors/hints/missing elements in the forms
- Confirmation for deleting/changing data
- Possibility to revert applied changes

PR #5404, opened 2022-07-04
Distinguish selected rows without colour

alyssarosenzweig:
Under the current rules, selected persons are distinguished from unselected persons only by the colour. As the pair of colours (grey / indico blue) have similar lightness, this creates a confusing situation in situations when the user cannot distinguish colours (some forms of colour blindness; some accessibility features filtering colours globally). In general, distinguishing only via colour is not recommended.

To fix, add a second indication of a row being selected, by widening the border. Before/after screenshots (with a desaturation filter applied to simulate the conditions where this change matters):

https://rosenzweig.io/Before-Desaturated.png
https://rosenzweig.io/After-Desaturated.png

Note: I have tested this change on a live Indico instance with the browser CSS inspector. I have not tested the actual SCSS, although I believe this to be correct.

---

Per WCAG section 1.4.1 ("Use of Color"):

> Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element.

this falls under #5381

### DIFF
--- a/indico/web/client/styles/widgets/_person_link.scss
+++ b/indico/web/client/styles/widgets/_person_link.scss
@@ -65,6 +65,7 @@
         &.selected {
           color: $indico-blue !important;
           border-color: $indico-blue;
+          border-width: 0.15em;
         }
 
         &:not(.other) {


ThiefMaster:
Sorry, but this widget is no longer used (we just didn't remove the CSS) in v3.2.

The new widget uses buttons from semantic-ui that look like this:

image

In case you want to update your PR, the relevant code for this is in `indico/web/client/js/react/components/PersonLinkField.jsx` (the Label elements with `color={active ? 'blue' : undefined}`). Not sure what's a good way to indicate active/inactive though without relying solely on colors and at the same time not making it less nice for people who can see colors just fine...

PR #5408, opened 2022-07-08
Add option to translate title and description of categories

### LINKED ISSUE #5353: Translations for Categories
**Problem Description**
Categories can be created using the category manager. When creating a category a title is assigned where most people use English titles and descriptions. As each category has only one title and description field, the displayed category title and description won't change if the user changes the language (not logged-in) for the web page or in his profile under "My Languages" (logged-in).

At our site, most people at least understand English, while others have severe difficulties with foreign languages and require full language support in their mother tongue in order  to take part and use events related to them and located in the category tree.

**Feature Proposed**
We propose to offer a way to create language dependent category titles and descriptions within the category manager of indico. The default title and descriptions should be assigned to the default language setting. Other languages could be added by chosing the corresponding language and adding title and description related to the current category entry.

If a translation of a category is not present, the title and description for the default language should be displayed, so backward compatibility is given when upgrading to an indico version including this feature.

elsbethe:
Add the possibility to translate the title and description of categories.

In the settings I've added buttons to create new fields to add a translation for a language

screenshot1

With the button you get a select field to select the desired language and an input field to enter the translation. You can add multiple translations if you click the button multiple times.

screenshot2

When the translations are saved, they appear in the form, so you can always edit them. When you empty the field, the translation is discarded.

screenshot3

When browsing through the categories/events the user sees the title and description of the category in the language he chose, if there is a translation available for that language. If no translation is available, the default title/description is shown.

closes #5353

### DIFF
--- a/indico/migrations/versions/20220630_1206_7c99d8f13df7_add_translation_of_title_and_.py
+++ b/indico/migrations/versions/20220630_1206_7c99d8f13df7_add_translation_of_title_and_.py
@@ -18,13 +18,13 @@ depends_on = None
 
 
 def upgrade():
-    op.add_column('categories', sa.Column('title_translations', postgresql.JSONB(astext_type=sa.Text()), nullable=True), schema='categories')
-    op.execute(r"UPDATE categories.categories SET title_translations = '{}'::jsonb")
-    op.alter_column('categories', 'title_translations', nullable=False, schema='categories')
+    op.add_column('categories', sa.Column('title_translations', postgresql.JSONB(astext_type=sa.Text()), nullable=False, server_default='{}'),
+                  schema='categories')
+    op.alter_column('categories', 'title_translations', server_default=None, schema='categories')
 
-    op.add_column('categories', sa.Column('description_translations', postgresql.JSONB(astext_type=sa.Text()), nullable=True), schema='categories')
-    op.execute(r"UPDATE categories.categories SET description_translations = '{}'::jsonb")
-    op.alter_column('categories', 'description_translations', nullable=False, schema='categories')
+    op.add_column('categories', sa.Column('description_translations', postgresql.JSONB(astext_type=sa.Text()), nullable=False, server_default='{}'),
+                  schema='categories')
+    op.alter_column('categories', 'description_translations', server_default=None, schema='categories')
 
 
 def downgrade():
--- a/indico/migrations/versions/20220630_1206_7c99d8f13df7_add_translation_of_title_and_.py
+++ b/indico/migrations/versions/20220630_1206_7c99d8f13df7_add_translation_of_title_and_.py
@@ -1,0 +1,32 @@
+"""Add translation of title and description to categories
+
+Revision ID: 7c99d8f13df7
+Revises: 0c4bb2973536
+Create Date: 2022-06-30 12:06:05.586959
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+
+# revision identifiers, used by Alembic.
+revision = '7c99d8f13df7'
+down_revision = '0c4bb2973536'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('categories', sa.Column('title_translations', postgresql.JSONB(astext_type=sa.Text()), nullable=True), schema='categories')
+    op.execute(r"UPDATE categories.categories SET title_translations = '{}'::jsonb")
+    op.alter_column('categories', 'title_translations', nullable=False, schema='categories')
+
+    op.add_column('categories', sa.Column('description_translations', postgresql.JSONB(astext_type=sa.Text()), nullable=True), schema='categories')
+    op.execute(r"UPDATE categories.categories SET description_translations = '{}'::jsonb")
+    op.alter_column('categories', 'description_translations', nullable=False, schema='categories')
+
+
+def downgrade():
+    op.drop_column('categories', 'description_translations', schema='categories')
+    op.drop_column('categories', 'title_translations', schema='categories')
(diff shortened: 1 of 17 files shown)


indico/migrations/versions/20220630_1206_7c99d8f13df7_add_translation_of_title_and_.py
```diff
@@ -0,0 +1,32 @@
...
+# revision identifiers, used by Alembic.
+revision = '7c99d8f13df7'
+down_revision = '0c4bb2973536'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('categories', sa.Column('title_translations', postgresql.JSONB(astext_type=sa.Text()), nullable=True), schema='categories')
+    op.execute(r"UPDATE categories.categories SET title_translations = '{}'::jsonb")
```
ThiefMaster, line 22: Create the column with a server_default and remove it again afterwards. See e.g. `20200623_1718_05f227f4b938_add_metadata_to_event_logs.py` for an example.

indico/modules/categories/models/categories.py
```diff
@@ -595,6 +605,12 @@ def logo_url(self):
         """Get the HTTP URL of the logo."""
         return url_for('categories.display_logo', self, slug=self.logo_metadata['hash'])
 
+    def get_title(self):
+        return self.title_translations.get(session.lang, self.title)
+
+    def get_description(self):
+        return self.description_translations.get(session.lang, self.description)
+
```
ThiefMaster, line 613: Maybe having properties like `i18n_title` and `i18n_description` would be nicer for this?

indico/modules/categories/templates/management/settings.html
```diff
@@ -27,6 +27,42 @@
...
+        {{ form_rows(form, fields=form.TITLE_FIELDS) }}
+        {% if untranslated_title_languages != '[]' %}
+            <div id="title-translation-container">
+                <div id="target-new-title-translation-fields"></div>
+            </div>
+            <div class="form-group">
+                <div class="form-label"></div>
+                <div class="form-field">
+                    <a class="i-button icon-edit"
+                    onclick="return addTranslateTitleFields('{{untranslated_title_languages}}')"
```
ThiefMaster, line 39: We don't really use inline JS anymore, it's usually not considered good practice nowadays ;)

But if you follow my advise with the "WTF react component" you won't need this stuff here anyway

indico/migrations/versions/20220630_1206_7c99d8f13df7_add_translation_of_title_and_.py
elsbethe: Thanks for the example!

indico/modules/categories/models/categories.py
elsbethe: I agree and have changed it.

indico/modules/categories/controllers/management.py
```diff
@@ -105,12 +108,69 @@ def _process(self):
...
                                 lecture_theme=self.category.default_event_themes['lecture'])
+
+        # Generate fields for existing translations
+        CategorySettingsForm.TITLE_FIELDS = ['title']
+        title_languages = get_all_locales()
+        for locale, translation in self.category.title_translations.items():
+            language = title_languages[locale]
+            language_text = f'{language[0]} ({language[1]})' if language[2] else language[0]
+            field_name = _('Title in ') + language_text
+            field_description = _('Users with ') + language_text + _(' as their language see this title.')
```
ThiefMaster, line 119: Use placeholders and formatting (`_('...').format(...)`) for all such cases - you cannot use string concatenation for i18n, since by doing so, you:

- assume it will have the same structure in all languages
- assume translators will realize that these strings belong together and form a single sentance
- assume surrounding whitespace in translatable strings is preserved, AND translators keep it

The correct way of translating the description string here is this:

` _('Users with {} as their language see this title.').format(language_text)`

indico/modules/categories/client/js/components/CategoryDescriptionTranslations.jsx
```diff
@@ -0,0 +1,60 @@
...
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import {Translate} from 'indico/react/i18n';
+
+export default function DescriptionTranslationFields({languages, counter}) {
```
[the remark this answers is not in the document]
elsbethe: I've changed it to one component. I am a bit of a React newby, so if I can do something better, please let me know.

indico/modules/categories/controllers/management.py
elsbethe: Thanks, I've changed it.

indico/modules/categories/models/categories.py
```diff
@@ -595,6 +605,14 @@ def logo_url(self):
         """Get the HTTP URL of the logo."""
         return url_for('categories.display_logo', self, slug=self.logo_metadata['hash'])
 
+    @property
+    def i18n_title(self):
+        return self.title_translations.get(session.lang, self.title)
+
+    @property
+    def i18n_description(self):
+        return self.description_translations.get(session.lang, self.description)
```
ThiefMaster, line 614: `description` has some special handling for rendering html/markdown (see `DescriptionMixin`); this logic needs to be preserved here. See RenderModeMixin `_render_getter` on what you need there.

ThiefMaster:
There's one somewhat tricky issue: In some places we use the `chain_titles` if a category which contains the titles of the parent category chain (queried in an efficient way). But of course this currently doesn't take into account the i18n you added...

I'm not sure what's the best solution for this, possibly adding an `chain_titles_translations` property similar to `chain_titles` where each entry contains the json object with the translations?

PR #5414, opened 2022-07-17
Add an optional js file to conference themes, and convert to ConferenceTheme dataclass

brittyazel:
This change is fairly straightforward, and was discussed [here](https://talk.getindico.io/t/questions-about-custom-conference-theme-plugin-development/2737/10)

In essence, I thought it would be nice to extend custom conference themes to both a single css file and a single, static, and optional javascript file. This way we can get both customized styles as well as a little bit of logic. My use-case for the javascript files is to be able do some simple html manipulation using jquerry like adding some new html elements into the banner area like what can be seen [here](https://events.gnome.org/event/77/).

I implemented this new  feature to the best of my ability, trying to keep my changes minimal and in-keep with your coding patterns. I also documented the change in the plugin.py file where the prior documentation was now out-of-date with this feature.

Likewise, in the discussion also came the idea to convert the conference theme from being declared as a series of 3 element tuples, and instead as a ConferenceTheme dataclass. The second of the two commits contains this conversion, which also remains backwards compatible with the prior method.

I did my best to test these two features as thoroughly as possible, and from my testing it all works perfectly fine and exactly how I intended. Apologies in advance if I missed something. That said, I would love for this to be merged into Indico 3.2 so my company can make use of it asap for our big Indico roll-out on September 1st.


indico/modules/events/layout/util.py
```diff
@@ -349,6 +379,43 @@ def get_css_url(event, force_theme=None, for_preview=False):
...
 
 
+def _build_js_url(theme):
+    if ':' in theme:
+        try:
+            # if the js_path is empty, we don't have a javascript file, so skip
+            if get_plugin_conference_themes()[theme].js_path:
+                path = get_plugin_conference_themes()[theme].js_path
+            else:
+                return None
```
ThiefMaster, line 389:
```suggestion
            return get_plugin_conference_themes()[theme].js_path
```

indico/modules/events/layout/util.py
```diff
@@ -349,6 +379,43 @@ def get_css_url(event, force_theme=None, for_preview=False):
         return _build_css_url(layout_settings.get(event, 'theme'))
 
 
+def _build_js_url(theme):
+    if ':' in theme:
```
ThiefMaster, line 383:
```suggestion
    if ':' not in theme:
        return None
```

and then the whole rest doesn't need to be indented in an if/else block anymore

brittyazel:
Great, thanks for the review! I agree with your changes. It looks cleaner this way

indico/modules/events/layout/util.py
brittyazel: We can't return immediately, because we're just trying to set the path variable. We use that path variable for the 'actual' return later on below.

brittyazel:
Ok, I just got home and tested your changes and there were some logic issues. I modified the code to keep the 'essence' of your changes while fixing the logic. Please take a look again.

indico/modules/events/layout/util.py
```diff
@@ -349,6 +379,41 @@ def get_css_url(event, force_theme=None, for_preview=False):
         return _build_css_url(layout_settings.get(event, 'theme'))
 
 
+def _build_js_url(theme):
+    if ':' not in theme:
+        return None
+    try:
+        path = get_plugin_conference_themes()[theme].js_path
+        if not path:
```
ThiefMaster, line 387:
```suggestion
        if not (path := get_plugin_conference_themes()[theme].js_path):
```

ThiefMaster:
Thanks, looks good on a quick glance. I'll have a closer look (and test it) when I'm in the office tomorrow.

PS: Feel free to add a changelog entry for 3.2 in the "improvements" section (search for "thanks" to find the format we usually use for external contributions)

indico/modules/events/layout/util.py
brittyazel: If we do it this way, we need to to query get_plugin_conference_themes()[theme].js_path a second time after the return None because we haven't actually set the path variable.

ThiefMaster: look closely ;) the walrus operator assigns to `path`. just a more concise way of writing it

brittyazel: Oh wow. This is the first I've heard of the walrus operator. Clever. It's what I get for going years in-between using python lol. Tested and can confirm it works.

brittyazel:
Ok your fixes are incorporated, I fixed the lint error about the imports being out of order, and I added two entries to the CHANGES file. I split the changes into two parts, being able to include a simple js file now, and the conversion to the ConferenceTheme dataclass.

indico/modules/events/layout/util.py
```diff
@@ -308,15 +308,44 @@ def is_menu_entry_enabled(entry_name, event):
...
+    js_path: str = None
+
+
 def get_plugin_conference_themes():
     data = values_from_signal(signals.plugin.get_conference_themes.send(), return_plugins=True)
-    return {':'.join((plugin.name, name)): (path, title) for plugin, (name, path, title) in data}
+    return {
+        ':'.join((plugin.name, (theme_info.name
+                                if isinstance(theme_info, ConferenceTheme)  # backwards compatibility check
+                                else ConferenceTheme(*theme_info).name))
```
bpedersen2, line 337: I would probably move the legacy handling out of  the return to avoid the double code (and double class instantiation in the legacy case)

ThiefMaster: How about this?


```py
def get_plugin_conference_themes():
    data = values_from_signal(signals.plugin.get_conference_themes.send(), return_plugins=True)
    data = [plugin, (theme_info if isinstance(theme_info, ConferenceTheme) else ConferenceTheme(*theme_info))
            for plugin, theme_info in data]
    return {f'{plugin.name}:{theme.name}': theme for plugin, theme in data}
```

indico/modules/events/layout/util.py
```diff
@@ -349,6 +378,39 @@ def get_css_url(event, force_theme=None, for_preview=False):
...
+def _build_js_url(theme):
+    if ':' not in theme:
+        return None
+    try:
+        if not (path := get_plugin_conference_themes()[theme].js_path):
+            return None
+    except KeyError:
+        return None
+    plugin = theme.split(':', 1)[0]
+    return url_for_plugin(plugin + '.static', filename=path)
```
ThiefMaster, line 390: I'd change this to an f-string here and also in `_build_css_url`:

```suggestion
    return url_for_plugin(f'{plugin}.static', filename=path)
```

CHANGES.rst
```diff
@@ -51,6 +51,13 @@ Improvements
 - Allow events to be favorited (:issue:`1662`, :pr:`5338`, thanks :user:`Leats`)
 - Include abstract content in CSV/Excel export if enabled in the abstract list
   (:issue:`5356`, :pr:`5372`, thanks :user:`rppt`)
+- Add the ability to include an optional static javascript file when defining
+  custom conference themes from within a plugin.
+  (:pr:`5414`, thanks :user:`brittyazel`)
+- Conference themes are now defined using a new ConferenceTheme dataclass as
+  opposed to using a tuple, though we still maintain backwards compatibility
+  with the old way.
```
ThiefMaster, line 59: I would skip this in the changelog tbh. It's really something internal which we typically do not mention there. However, feel free to include a short mention in the "Internal Changes" if you want (personally i'd skip it). If you want to keep it I'd reword it a bit though ("Conference themes from plugins can now be defined using a `ConferenceTheme` object instead of a tuple"). That's more concise and "can" already implies it's backwards compatible.

indico/modules/events/layout/util.py
brittyazel: @ThiefMaster unfortunately that syntax, as written is throwing errors.

ThiefMaster: Just missing some parentheses, I didn't test it.

```python
def get_plugin_conference_themes():
    data = values_from_signal(signals.plugin.get_conference_themes.send(), return_plugins=True)
    data = [(plugin, (theme_info if isinstance(theme_info, ConferenceTheme) else ConferenceTheme(*theme_info)))
            for plugin, theme_info in data]
    return {f'{plugin.name}:{theme.name}': theme for plugin, theme in data}
```

brittyazel: `   data = values_from_signal(signals.plugin.get_conference_themes.send(), return_plugins=True)
    for plugin, theme_info in data:
        if not isinstance(theme_info, ConferenceTheme):
            theme_info = ConferenceTheme(*theme_info)
        return {f'{plugin.name}:{theme_info.name}': theme_info}`

This is how my brain logically tells me to do it, but like I stated before, it only shows 1 theme if multiple are pressed for some reason

ThiefMaster: Yes, the `return` aborts the loop early... try the snippet from the last comment, I think it does exactly what's needed.

brittyazel: Sorry, you snuck in some replies while I was coding this lol. Yeah I figured out the parenthesis

ThiefMaster:
Looks good and works fine now. Thanks for the contribution!

ThiefMaster (approved):

PR #5421, opened 2022-07-21
Add signal for blocking registration modifications

duartegalvao:
Call a signal when resolving whether Indico should let a registrant modify their registration details.

### DIFF
--- a/indico/core/signals/event/registration.py
+++ b/indico/core/signals/event/registration.py
@@ -90,6 +90,14 @@ both users trying to get their own ticket and managers trying to get a
 ticket for a registrant.
 ''')
 
+is_registration_blocked = _signals.signal('is-registration-blocked', '''
+Called when resolving whether Indico should let a registrant modify their
+registration details.  The `sender` is the registrant's `Registration` object.
+
+If this signal returns ``True``, the user will not be able to modify their
+registration.
+''')
+
 filter_selectable_badges = _signals.signal('filter-selectable-badges', '''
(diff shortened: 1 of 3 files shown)


indico/core/signals/event/registration.py
```diff
@@ -90,6 +90,14 @@
 ticket for a registrant.
 ''')
 
+is_registration_blocked = _signals.signal('is-registration-blocked', '''
```
ThiefMaster, line 93: I don't like the name of it, it sounds more like it's preventing someone from register.
Maybe something like `is_registration_modification_locked` instead?

indico/modules/events/registration/models/registrations.py
```diff
@@ -408,7 +408,9 @@ def locator(self):
     @property
     def can_be_modified(self):
         regform = self.registration_form
-        return regform.is_modification_open and regform.is_modification_allowed(self)
+        return (regform.is_modification_open
+                and regform.is_modification_allowed(self)
+                and not any(values_from_signal(signals.event.is_registration_blocked.send(self), single_value=True)))
```
ThiefMaster, line 413: I don't think `single_value` should be used here.

ThiefMaster:
Just thinking.. an alternative solution could be adding this signal to block modifications on the field level (like we do for purged fields or payment-related fields in a paid registration). Like this you could restrict only changes to the accompanying persons field without having to restrict all modifications.

PR #5424, opened 2022-07-22
Generate tickets for accompanying persons

duartegalvao:
This PR adds the ability to generate tickets for each for the registrants' accompanying persons, as well as some internal changes to support the [integration of accompanying persons and the CERN Access plugin](https://github.com/indico/indico-plugins-cern/pull/113).

The ticket generation can be configured in the registration form's ticket settings dialog:

image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -573,6 +573,8 @@ Internal Changes
   thanks :user:`omegak`).
 - Update WYSIWYG editor to CKEditor 5, resulting in a slightly different look for the
   editor controls and removal of some uncommon format options (:pr:`5345`)
+- Add ``event.is_field_data_locked`` signal, allowing plugins to lock registration form
+  fields on a per-registration basis (:pr:`5424`)
 
 
 ----
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,8 @@ Improvements
 - Make editing timeline display much more straightforward (:pr:`5674`)
 - Allow event managers to delete editables from contributions (:pr:`5778`, :pr:`5892`)
 - Allow room managers to add internal notes to bookings (:issue:`5746`, :pr:`5791`)
+- Support generating tickets and badges for each of the registrant's accompanying
+  persons (:pr:`5424`)
 
 Bugfixes
 ^^^^^^^^
@@ -37,6 +39,8 @@ Internal Changes
 
 - Use (dart-)sass instead of the deprecated node-sass/libsass for CSS compilation
   (:pr:`5734`)
+- Add ``event.is_field_data_locked`` signal, allowing plugins to lock registration form
+  fields on a per-registration basis (:pr:`5424`)
 
 
 ----
@@ -431,8 +435,6 @@ Improvements
   thanks :user:`omegak`)
 - Include emails of submitters, speakers and authors in abstract/contribution
   Excel/CSV exports (:pr:`5565`)
-- Support generating tickets and badges for each of the registrant's accompanying
-  persons (:pr:`5424`)
 
 Bugfixes
 ^^^^^^^^
@@ -461,12 +463,6 @@ Internal Changes
   thanks :user:`omegak`)
 - Indicate in the ``registration_deleted`` signal whether it's a permanent deletion
   from the database or just a soft-deletion (:pr:`5559`)
-
-Internal Changes
-^^^^^^^^^^^^^^^^
-
-- Add ``event.is_field_data_locked`` signal, allowing plugins to lock registration form
-  fields on a per-registration basis (:pr:`5424`)
 
 
 Version 3.2
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -431,6 +431,8 @@ Improvements
   thanks :user:`omegak`)
 - Include emails of submitters, speakers and authors in abstract/contribution
   Excel/CSV exports (:pr:`5565`)
+- Support generating tickets and badges for each of the registrant's accompanying
+  persons (:pr:`5424`)
 
 Bugfixes
 ^^^^^^^^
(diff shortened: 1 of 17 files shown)


indico/modules/events/registration/util.py
```diff
@@ -154,6 +154,7 @@ def get_flat_section_submission_data(regform, *, management=False, registration=
             field_data['availablePlaces'] = item.field_impl.get_available_places(registration)
         else:
             field_data = item.view_data
+        field_data['isLocked'] = registration_data[item.id].is_locked
```
ThiefMaster, line 157: I'm pretty sure this will fail in case a field was added after the registration was created. I think the locked flag should be on the field, not the data model (ie `item.is_locked` in here)

indico/modules/events/registration/util.py
```diff
@@ -384,7 +385,7 @@ def modify_registration(registration, data, management=False, notify_user=True):
 
     billable_items_locked = not management and registration.is_paid
     for form_item in regform.active_fields:
-        if form_item.is_purged:
+        if form_item.is_purged or form_item.is_locked(registration):
```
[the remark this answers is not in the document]
duartegalvao: I also didn't call it there because I'm not sure if it would be better to pass `None` or the newly created `Registration` object - I guess it would depend on the potential use case?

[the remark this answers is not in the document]
ThiefMaster: `None` for sure, especially since you also need to call this when rendering the form to know whether it should be displayed as locked or not - and at that point you do not have such an object.

indico/migrations/versions/20220830_1428_0c44046dc1be_add_generate_accompanying_person_tickets.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = '0c44046dc1be'
+down_revision = 'b45847c0e62f'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('forms', sa.Column('tickets_for_accompanying_persons', sa.Boolean(), nullable=False,
+                                     server_default='true'), schema='event_registration')
```
ThiefMaster, line 21: I would default to `false` here since that way there's no change in behavior for existing events.

indico/modules/designer/placeholders.py
```diff
@@ -67,12 +79,14 @@ def render(cls, registration):
...
+    data_source = 'person'
     name_options = None
 
     @classmethod
-    def render(cls, registration):
-        name = (registration.get_personal_data().get('title', '') + ' ') if cls.with_title else ''
-        name += registration.get_full_name(**cls.name_options)
+    def render(cls, person):
+        name = ((person['registration'].get_personal_data().get('title', '') + ' ')
+                if cls.with_title and not person['is_accompanying'] else '')
```
ThiefMaster, line 88:
```py
title = ''
if cls.with_title and not person['is_accompanying']:
    title = person['registration'].get_personal_data().get('title', '') 
```

and then pass the title to `format_full_name`

indico/modules/events/registration/badges.py
```diff
@@ -107,7 +128,7 @@ def _draw_badge(self, canvas, registration, template, tpl_data, pos_x, pos_y):
 
             item_data = {'item': item, 'text': text, 'pos_x': pos_x, 'pos_y': pos_y}
             for update in values_from_signal(
-                signals.event.designer.draw_item_on_badge.send(registration, items=items, height=self.height,
+                signals.event.designer.draw_item_on_badge.send(person['registration'], items=items, height=self.height,
```
ThiefMaster, line 131: This signal should probably receive the `person` as well...

indico/modules/events/registration/models/items.py
```diff
@@ -379,6 +381,18 @@ def is_visible(cls):
                    .exists())
         return cls.is_enabled & ~cls.is_deleted & ((cls.parent_id == None) | query)  # noqa
 
+    def is_locked(self, registration):
+        """Check whether the data is locked by a plugin."""
+        values = values_from_signal(signals.event.is_field_data_locked.send(self, registration=registration))
+        return any(locked for locked, _ in values)
```
ThiefMaster, line 387: I usually use `__` since `_` is already used for i18n

indico/modules/events/registration/models/items.py
```diff
@@ -379,6 +381,18 @@ def is_visible(cls):
...
+
+    def get_locked_reason(self, registration):
+        """Get the reason for the field being locked."""
+        values = values_from_signal(signals.event.is_field_data_locked.send(self, registration=registration))
+        for locked, reason in values:
+            if locked:
+                return reason
```
ThiefMaster, line 394:
```python
return next((reason for locked, reason in values if locked), None)
```

But maybe change the signal so it just returns a lock reason or None if it's not locked? For example what would happen right now if the signal returned `True, ''` - it'd be locked, but no useful reason would be available..

CHANGES.rst
```diff
@@ -13,6 +13,8 @@ Improvements
 - Enable better image linking UI in CKEditor (:pr:`5492`)
 - Restore the "fullscreen view" option in CKEditor (:pr:`5505`)
 - Display & enforce judging deadline (:pr:`5506`)
+- Support generating tickets for each of the registrant's accompanying persons
```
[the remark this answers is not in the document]
duartegalvao: Yep, it works
Sounds like a useful improvement, I'll create an issue for it

indico/modules/events/registration/models/registrations.py
```diff
@@ -550,6 +551,16 @@ def visibility(self):
...
 
+    @property
+    def accompanying_persons(self):
+        from indico.modules.events.registration.models.form_fields import RegistrationFormFieldData
+        query = (RegistrationData.query
+                 .with_parent(self)
+                 .join(RegistrationFormFieldData)
+                 .filter(RegistrationFormFieldData.field.has(input_type='accompanying_persons')))
+        return list(itertools.chain.from_iterable(d.data for d in query.all()
+                                                  if d.field_data.field.is_active))
```
ThiefMaster: I don't think this is correct. When you disable a field it doesn't remove it from existing registrations.

ThiefMaster: I changed it to `not is_deleted`

ThiefMaster (approved):

PR #5426, opened 2022-07-25
Fix sorting of session blocks in PDF export

javfg:
This PR fixes a bug where the PDF export of a conference does not sort session blocks alphabetically when they have the same start time.

The bug was reported in [SNOW ticket INC3186005](https://cern.service-now.com/service-portal?id=ticket&n=INC3186005).


ThiefMaster:
maybe worth adding a changelog entry in the bugfixes section for this?

ThiefMaster (approved):

PR #5428, opened 2022-07-27
Add blocks to extend badge print settings

SegiNyn:


indico/modules/events/registration/templates/management/print_badges.html
```diff
@@ -2,24 +2,29 @@
...
 
 {% if registrations %}
+    {% set available_registrations = True %}
+{% endif %}
+
+{% if available_registrations %}
     {% call simple_form(settings_form, form_header_kwargs={'id': 'badge-settings-form'}, back=_('Cancel'),
                         submit=_("Download PDF"), disabled_until_change=false, disable_if_locked=false) %}
-        {{ form_row(settings_form.template) }}
+        {% block form_template %} {{ form_row(settings_form.template) }} {% endblock %}
```
ThiefMaster, line 24:
```suggestion
        {% block form_template %}
            {{ form_row(settings_form.template) }}
        {% endblock %}
```

indico/modules/events/registration/templates/management/print_badges.html
```diff
@@ -2,24 +2,31 @@
...
+                Tickets are currently not available for the selected registration.
+            {%- pluralize total -%}
+                Tickets are currently not available for {{ count }} out of the {{ total }} selected registrations.
+            {%- endtrans %}
+        {%- endcall %}
+    {% endif %}
+{% endblock %}
 
 {% if registrations %}
+    {% set available_registrations = True %}
```
ThiefMaster, line 18:
```suggestion
    {% set available_registrations = true %}
```

ThiefMaster: also this is super weird. and likely to get refactored out. why do you need this and can you find some better way of doing this?

SegiNyn: We need this because we are trying to use the content of this template to set the badge print settings at the category level without performing the print, so we do not need the registrations to be present and so we set it to always be true to get the rest of the content

ThiefMaster: maybe wrap it in a block which you then override either with or without calling `{{ super() }}`?

SegiNyn: okay, I'll try that.

SegiNyn:
I added 2 blocks, an outer block for the if condition and an inner one for the content

ThiefMaster (approved):

PR #5430, opened 2022-07-30
Add workarounds for Japanese locale

omasanori:
Like Ukrainian, Japanese is likely used officially only in Japan but the shortened name for the language (ja) and that for the country (jp) are different. Thus, a similar treatment is required.

### DIFF
--- a/indico/web/flask/session.py
+++ b/indico/web/flask/session.py
@@ -83,7 +83,7 @@ class IndicoSession(BaseSession):
         parts = self.lang.lower().split('_')  # e.g. `en_GB` or `zh_Hans_CN`
         lang = parts[0]
         territory = parts[-1]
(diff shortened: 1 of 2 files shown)


indico/web/flask/session.py
```diff
@@ -83,7 +83,7 @@ def moment_lang(self):
         parts = self.lang.lower().split('_')  # e.g. `en_GB` or `zh_Hans_CN`
         lang = parts[0]
         territory = parts[-1]
-        if lang == territory or lang == 'uk' or self.lang == 'en_US':
+        if lang == territory or lang == 'ja' or lang == 'uk' or self.lang == 'en_US':
```
ThiefMaster, line 86: `or lang in ('ja', 'uk')` please :)

omasanori: Good catch!

PR #5431, opened 2022-08-01
Remove unused regform CSS rules

tomasr8:
Mostly classes for fields which are now in React and have their styles in modules, some angular classes and tabs-related classes.

~~Maybe even all the `tabbable` classes could go, but with them I'm not sure if they're still not used somewhere.~~


ThiefMaster:
```
[adrian@eluvian:~/dev/indico/py3/src:master]> rg tabbable
indico/web/client/styles/modules/_registrationform.scss
739:.tabbable {
744:.tabbable::before,
745:.tabbable::after {
750:.tabbable::after {

indico/web/client/js/jquery/widgets/misc.js
109:        elem = $this.find(':input:tabbable');
```

sounds like they can go :)

PR #5434, opened 2022-08-03
Use `www-data` instead of `nginx` group in debian config

ClundXIII:
The wrong group is used in a few places, it should be `www-data`:

```
/etc/uwsgi-indico.ini
/etc/systemd/system/indico-uwsgi.service
/etc/systemd/system/indico-celery.service
```

Additionally if you use debian 11, no external package sources are needed! (postgres is v13 and the python stuff is installed locally for the user anyway).

### DIFF
--- a/docs/source/installation/production/debian/nginx.rst
+++ b/docs/source/installation/production/debian/nginx.rst
@@ -6,17 +6,12 @@ nginx
 1. Install Packages
 -------------------
 
-PostgreSQL and nginx are installed from their upstream repos to get
-much more recent versions.
+The minimal recommended version is debian 11 (buster).
 
 .. code-block:: shell
 
     apt install -y lsb-release wget curl gnupg
-    echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list
-    echo "deb http://nginx.org/packages/$(lsb_release -is | tr '[:upper:]' '[:lower:]')/ $(lsb_release -cs) nginx" > /etc/apt/sources.list.d/nginx.list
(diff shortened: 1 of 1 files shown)


ThiefMaster:
I think this depends on the debian version. I'm 100% sure that the guide worked the last time I tested it. So probably this changed only with Debian 11.

Maybe we should just update our docs for Debian 11 (when was it released?) and assume people with an older version have the necessary skills to adapt (w/o step by step guidance for the differences there)...

ClundXIII:
> I think this depends on the debian version. I'm 100% sure that the guide worked the last time I tested it. So probably this changed only with Debian 11.

debian 11 and 10 user `www-data`. And if I remember correctly, debian 9 as well.

ClundXIII:
> Maybe we should just update our docs for Debian 11 (when was it released?) and assume people with an older version have the necessary skills to adapt (w/o step by step guidance for the differences there)...

That sounds reasonable, yes.

ThiefMaster:
I don't remember during which version I wrote the docs, but could it be a difference between the cloud/VM images and docker images? When I tested them I created VMs on OpenStack using the latest debian cloud images

ClundXIII:
Okay, I think this should be okay. I also added the startcluster command which is needed sometimes.

axu-trex:
Wait ... so if all dependencies are already packaged for Debian, then we could just start packaging indico for Debian/sid as well? I had been thinking about this already, but I was daunted by the prospect of wading through all the dependencies and packaging them all...

ThiefMaster:
old PR, will be covered when I update our install docs for the latest OS versions before the v3.3 release

ThiefMaster:
According to my latest tests on the Ubuntu 24.04 prerelease cloud image:

- When using nginx from the nginx repos, it runs as `nginx:nginx`
- When using nginx from the debian/ubuntu repos, it runs as `www-data:www-data`

PR #5436, opened 2022-08-04
Add UI to manage series of events

### LINKED ISSUE #4048: UI to manage series of events
We are already "silently" linking together events whenever someone creates a lecture series. This is the concept of [`Series`](https://github.com/indico/indico/blob/master/indico/modules/events/models/series.py), which we introduced in 2.0.

However, we don't have a UI where organizers can manage their series, so it's impossible to e.g. add an additional lecture to a series without an admin using the CLI. The idea would be to implement a basic management UI with the following features:

 * Creating/managing an event series (just a "group of events");
 * (Un)linking an event to/from an existing series;

See also #2993. (out of the scope of this issue, much more complex and should NOT be done at the same time because of this)

---

Main tasks to do here:

- Add UI for users to create a new series
- Add UI for users to manage a series (edit its settings and add/remove events)
- Add new option to include new event(s) to the series while cloning if the current event is part of a series
- Create an event selection widget for the series UI - either pasting links to the event, or a search field that searches by title within the same category. Could also suggest events straight away (based on title/dates). To be discussed w/ the Indico team (and possibly mocked up)

Leats:
Closes #4048.
Adds the ability to manage/create series when editing an event.
Preview:
Bildschirmfoto 2022-08-04 um 10 35 33
Bildschirmfoto 2022-08-04 um 11 23 57
Bildschirmfoto 2022-08-04 um 11 22 03

### DIFF
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -42,7 +42,9 @@ from indico.util.date_time import format_date, format_number, now_utc
 from indico.util.decorators import classproperty
 from indico.util.fs import secure_filename
 from indico.util.i18n import _
+from indico.util.marshmallow import LowercaseString, not_empty
 from indico.util.signals import values_from_signal
+from indico.web.args import use_kwargs
 from indico.web.flask.templating import get_template_module
 from indico.web.flask.util import send_file, url_for
 from indico.web.rh import RH, allow_signed_url
@@ -190,14 +192,17 @@ class RHCategorySearch(RH):
 class RHCategoryManagedEventSearch(RHCategoryBase):
     """Search managed events in a category."""
 
-    def _process(self):
+    def _check_access(self):
+        if not session.user:
+            raise Forbidden
+        RHCategoryBase._check_access(self)
+
+    @use_kwargs({'q': LowercaseString(required=True, validate=not_empty)}, location='query')
+    def _process(self, q):
         from indico.modules.events.series.schemas import EventDetailsForSeriesManagementSchema
-        q = request.args['q'].lower()
         query = (
             Event.query.with_parent(self.category)
-            .filter(
-                Event.title_matches(q), Event.series == None, ~Event.is_deleted # noqa
-            )
+            .filter(Event.title_matches(q), Event.series_id.is_(None), ~Event.is_deleted)
             .options(load_only('id', 'title', 'start_dt', 'end_dt', 'category_chain'))
         )
         # Prefer favorite events
@@ -208,13 +213,8 @@ class RHCategoryManagedEventSearch(RHCategoryBase):
             db.func.lower(Event.title).startswith(q).desc(),
             db.func.lower(Event.title),
         )
-        result = [EventDetailsForSeriesManagementSchema().dump(e) for e in query if e.can_manage(session.user)]
-        return jsonify(result[:min(10, len(result))])
-
-    def _check_access(self):
-        if not session.user:
-            raise Forbidden
-        RHCategoryBase._check_access(self)
+        events = get_n_matching(query, 10, lambda event: event.can_manage(session.user))
+        return EventDetailsForSeriesManagementSchema(many=True).jsonify(events)
 
 
 class RHSubcatInfo(RHDisplayCategoryBase):
--- a/indico/modules/categories/controllers/display.py
+++ b/indico/modules/categories/controllers/display.py
@@ -18,13 +18,13 @@ from dateutil.relativedelta import relativedelta
 from flask import flash, jsonify, redirect, request, session
 from pytz import utc
 from sqlalchemy.orm import joinedload, load_only, subqueryload, undefer, undefer_group
-from werkzeug.exceptions import BadRequest, NotFound
+from werkzeug.exceptions import BadRequest, Forbidden, NotFound
 
 from indico.core import signals
 from indico.core.db import db
 from indico.core.db.sqlalchemy.colors import ColorTuple
 from indico.core.db.sqlalchemy.util.queries import get_n_matching
-from indico.modules.categories.controllers.base import RHDisplayCategoryBase
+from indico.modules.categories.controllers.base import RHCategoryBase, RHDisplayCategoryBase
 from indico.modules.categories.controllers.util import (get_category_view_params, get_event_query_filter,
                                                         group_by_month, make_format_event_date_func,
                                                         make_happening_now_func, make_is_recent_func)
@@ -37,7 +37,7 @@ from indico.modules.events.models.events import Event
 from indico.modules.events.timetable.util import get_category_timetable
 from indico.modules.news.util import get_recent_news
 from indico.modules.users import User
-from indico.modules.users.models.favorites import favorite_category_table
+from indico.modules.users.models.favorites import favorite_category_table, favorite_event_table
 from indico.util.date_time import format_date, format_number, now_utc
 from indico.util.decorators import classproperty
 from indico.util.fs import secure_filename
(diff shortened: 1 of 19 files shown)


indico/modules/events/series/__init__.py
```diff
@@ -0,0 +1,6 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2022 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
```
ThiefMaster, line 6: otherwise-empty `__init__.py` files don't get a header

I just added a case for this to my header update script in master, so after rebasing you can run `python ./bin/maintenance/update_header.py` for the file to be updated (of course you could also do it manually but using the script it ensures the editor doesn't add e.g. a single newline)

indico/modules/events/series/blueprint.py
```diff
@@ -0,0 +1,15 @@
...
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from indico.modules.events.series.controllers import RHEventSeries
+from indico.web.flask.wrappers import IndicoBlueprint
+
+
+_bp = IndicoBlueprint('event_series', __name__, url_prefix='/series/')
```
ThiefMaster, line 12:
```suggestion
_bp = IndicoBlueprint('event_series', __name__, url_prefix='/event-series')
```

I think that name in the URL is a bit better, and there shouldn't be a trailing slash the the prefix is really just prepended to the url rules below - so it'd generate `/series//`

indico/modules/events/series/schemas.py
```diff
@@ -0,0 +1,62 @@
...
+        fields = ('events', 'show_sequence_in_title', 'show_links')
+        rh_context = ('series',)
+
+    events = ModelList(Event, filter_deleted=True, data_key='event_ids', required=True, validate=not_empty)
+
+    @validates('events')
+    def _check_event_ids(self, events, **kwargs):
+        series_id = (
+            self.context['series'].id if 'series' in self.context and self.context['series'] is not None else None
+        )
```
ThiefMaster, line 37:
```suggestion
        series_id = series.id if (series := self.context.get('series')) is not None else None
```

indico/modules/events/series/schemas.py
```diff
@@ -0,0 +1,62 @@
...
+        if events is None:
+            raise ValidationError(_('No event IDs provided.'))
```
ThiefMaster, line 39: I haven't tested it but I *think* the `not_empty` validation on the `ModelList` already covers it

indico/modules/events/series/schemas.py
```diff
@@ -0,0 +1,62 @@
...
+        event_ids = [e.id for e in events]
+
+        if set(event_ids) - {e.id for e in events}:
+            raise ValidationError(_('Invalid event IDs provided.'))
```
ThiefMaster, line 43: This is no longer needed, the `ModelList` will reject any ID that doesn't exist in the DB - and in fact the check here will always return False since you basically subtract the same set from itself which is always empty and thus falsy

indico/modules/events/series/schemas.py
```diff
@@ -0,0 +1,62 @@
...
+        elif series_id is None and [e for e in events if e.series]:
```
ThiefMaster, line 44:
```suggestion
        if series_id is None and any(e.series for e in events):
```

indico/modules/events/series/schemas.py
```diff
@@ -0,0 +1,62 @@
...
+            raise ValidationError(_('At least one event is already in a series.'))
+        elif [e for e in events if e.series and e.series_id is not series_id]:
```
ThiefMaster, line 46:
```suggestion
        elif any(e.series and e.series_id != series_id for e in events):
```

indico/modules/events/series/schemas.py
```diff
@@ -0,0 +1,62 @@
...
+            raise ValidationError(_('At least one event is already in a different series.'))
+
+        for event in events:
+            if not event.can_manage(session.user):
+                raise ValidationError(_('You are not allowed to manage all of these events.'))
```
ThiefMaster, line 51:
```suggestion
        if not all(e.can_manage(session.user) for e in events):
            raise ValidationError(_('You must be a manager of all chosen events.'))
```

(btw if we have clientside validation for all of these things we can ditch the i18n markers here)

indico/modules/events/management/client/js/SeriesManagement.jsx
```diff
@@ -0,0 +1,400 @@
...
+    }
+  };
+
+  const eventOptions = results.map(evt => ({
+    key: evt.id,
+    value: evt.id,
+    meta: evt,
+    text: evt.title,
+    disabled:
+      evt.canManage !== true || evt.seriesId !== null || !!currentEvents.find(e => e.id === evt.id),
```
ThiefMaster, line 148:
```suggestion
      !evt.canManage || evt.seriesId !== null || !!currentEvents.find(e => e.id === evt.id),
```

indico/modules/events/management/client/js/SeriesManagement.jsx
```diff
@@ -0,0 +1,400 @@
...
+        <span styleName="date-span">{localMoment(evt.startDt).format('ll')}</span>
+        <span styleName="event-title">
+          {evt.title}
+          <br />
+          <TooltipIfTruncated>
+            <span>
+              {evt.categoryChain ? evt.categoryChain.join(' » ') : <Translate>Unlisted</Translate>}
+            </span>
+          </TooltipIfTruncated>
+          {evt.canManage !== true && (
```
ThiefMaster, line 160:
```suggestion
          {!evt.canManage && (
```

indico/modules/categories/controllers/display.py
```diff
@@ -187,6 +189,36 @@ def _process(self):
...
+        RHCategoryBase._check_access(self)
+
+    @use_kwargs({'q': LowercaseString(required=True, validate=not_empty)}, location='query')
+    def _process(self, q):
+        from indico.modules.events.series.schemas import EventDetailsForSeriesManagementSchema
+        query = (
+            Event.query.with_parent(self.category)
+            .filter(
+                Event.title_matches(q), Event.series_id.is_(None), ~Event.is_deleted
+            )
```
ThiefMaster, line 207: I'd either put this in one line or each criterion on a separate lane

indico/modules/categories/controllers/display.py
```diff
@@ -187,6 +189,36 @@ def _process(self):
...
+        )
+        # Prefer favorite events
+        query = query.order_by(Event.favorite_of.any(favorite_event_table.c.user_id == session.user.id).desc())
+        # Prefer exact matches and matches at the beginning, then order by event title
+        query = query.order_by(
+            (db.func.lower(Event.title) == q).desc(),
+            db.func.lower(Event.title).startswith(q).desc(),
+            db.func.lower(Event.title),
+        )
+        result = [EventDetailsForSeriesManagementSchema().dump(e) for e in query if e.can_manage(session.user)]
```
ThiefMaster, line 218: I think you can use the `get_n_matching` util here - that way you are guaranteed to get enough events if enough exist, and avoid querying and access-checking more than necessary

and at the end you can simply use `EventDetailsForSeriesManagementSchema(many=True).jsonify(events)` to dump the list

indico/modules/events/blueprint.py
```diff
@@ -58,6 +59,7 @@
 
 # Misc
 _bp.add_url_rule('/event/<int:event_id>/key-access', 'key_access', RHEventAccessKey, methods=('POST',))
+_bp.add_url_rule('/event/<int:event_id>/api/info-for-series', 'single_event_api', RHSingleEventAPI, methods=('GET',))
```
ThiefMaster, line 62:
```suggestion
_bp.add_url_rule('/event/<int:event_id>/api/info-for-series', 'single_event_api', RHSingleEventAPI)
```

GET is already the default ;)

indico/modules/events/controllers/base.py
```diff
@@ -96,6 +96,20 @@ def _do_process(self):
             return self._show_registration_form()
 
 
+class RHSingleEventAPI(RH):
```
ThiefMaster, line 99: this one is no longer used

indico/modules/events/management/client/js/index.js
```diff
@@ -35,6 +36,22 @@ import {SingleEventMove, EventPublish} from './EventMove';
...
+    if (manageSeriesContainer) {
+      const hasSeries = manageSeriesContainer.dataset.seriesId !== undefined;
+      const seriesId = hasSeries ? +manageSeriesContainer.dataset.seriesId : null;
+      ReactDOM.render(
+        React.createElement(SeriesManagement, {
+          eventId: +manageSeriesContainer.dataset.eventId,
+          categoryId: +manageSeriesContainer.dataset.categoryId,
+          timezone: manageSeriesContainer.dataset.timezone,
+          hasSeries,
+          seriesId,
```
ThiefMaster, line 49: I'd still get rid of `hasSeries` here (and move it into the component) - better to have just one source of truth (`seriesId !== null` (or undefined))

indico/modules/events/series/controllers.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class RHEventSeries(RH):
+    def _process_args(self):
+        self.series = EventSeries.get(request.view_args['series_id']) if 'series_id' in request.view_args else None
+
+    def _process_GET(self):
+        if self.series is None:
+            raise NotFound(_('A series with this ID does not exist.'))
+        return jsonify({'events': [EventDetailsSchema().dump(e) for e in self.series.events],
```
ThiefMaster, line 27:
```suggestion
        return jsonify({'events': EventDetailsSchema(many=True).dump(self.series.events),
```

But I think the whole jsonify block here can be replaced with `EventSeriesSchema().jsonify(self.series)` - you may just need to include `events = fields.Nested(EventDetailsSchema)` in that schema

indico/modules/events/management/client/js/index.js
Leats: oh lol I misunderstood what you meant. I mean, I got rid of it in the step before, but didn't got that far

ThiefMaster (approved):

PR #5439, opened 2022-08-05
Add scroll when material tooltip is too long

tomasr8:
Adds max height with a scroll to prevent the tooltip from clipping and only showing a part of the materials.

image

### DIFF
--- a/indico/web/client/styles/legacy/Default.css
+++ b/indico/web/client/styles/legacy/Default.css
@@ -792,6 +792,8 @@ body.cke_show_borders {
 
 .material_tip .qtip-content {
   background: #f8f8f8;
+  max-height: 50vh;
+  overflow-y: auto;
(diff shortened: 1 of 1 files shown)


indico/web/client/styles/legacy/Default.css
```diff
@@ -788,6 +788,8 @@ body.cke_show_borders {
 .material_tip {
   border-color: #ddd !important;
   box-shadow: 1px 1px 5px rgba(100, 100, 100, 0.2);
+  max-height: 500px;
+  overflow-y: scroll;
```
ThiefMaster, line 792: doesn't this mean there are *always* scroll bars? IIRC it's `overflow-y: auto` if you just want them to show up when needed

ThiefMaster (approved):

PR #5443, opened 2022-08-08
Fix Poster contributions start and end times

duartegalvao:
This PR fixes a bug in which Poster contributions displayed a misleading start and end time in e-mails and the HTTP API.

The bug was caused by Poster contributions not having and inherent time schedule, and thus should display their parent session's start and end times instead of its own times.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -77,6 +77,8 @@ Bugfixes
   have the same start time (:pr:`5426`)
 - Fix printing badges containing text elements with malformed HTML (:pr:`5437`,
   thanks :user:`omegak`)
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -77,6 +77,8 @@ Bugfixes
   have the same start time (:pr:`5426`)
 - Fix printing badges containing text elements with malformed HTML (:pr:`5437`,
   thanks :user:`omegak`)
+- Fix misleading start and end times for Poster contributions in the timetable HTTP API
+  and author e-mails (:pr:`5443`)
```
ThiefMaster, line 81: I think "and the contributions placeholder in emails" is clearer

ThiefMaster (approved):

PR #5447, opened 2022-08-11
Add a mapping from canonical locale names to moment locales

tomasr8:
For example, if a user's locale is `uk_UA` this will be translated as `uk` for moment.
Similarly `zh_Hans_CN` becomes `zh-cn`.

For a few moment locales I couldn't find a corresponding official locale name, but these should be fairly uncommon:
```
me (Montenegrin)
tlh (Klingon)
tzm_latn (Central Atlas Tamazight Latin)
oc-lnc (Occitan, lengadocian dialecte)
gom-deva (Konkani Devanagari script)
gom-latn (Konkani Latin script)
tzl (Talossan)
tet (Tetun Dili (East Timor))
x-pseudo (Pseudo)
```


indico/web/flask/session.py
```diff
@@ -22,6 +23,10 @@
 from indico.web.util import get_request_user
 
 
+with open('indico/moment-locales.yml') as f:
```
ThiefMaster, line 26: That looks somewhat fragile... I'd rather doing it during the first request (the function can be memoized with `@functools.cache` to avoid parsing it more than once) with an absolute path (and .yaml since we already use that for other yaml files):

```py
        with open(os.path.join(current_app.root_path, 'moment-locales.yaml')) as f:
            ...
```

(or the pathlib equivalent ;))

indico/web/flask/session.py
```diff
@@ -80,19 +85,17 @@ def lang(self, lang):
...
-            # TODO we should add some metadata that stores the canonical locale name and
-            # the name of the moment locale to avoid hacks like the one here. for example,
-            # fr_FR is handled somewhat nicely, but ukrainian (uk_UA) needs an explicit check
-            # since it's a single-country locale but locale and territory name are different..
-            # `setMomentLocale` in JS has the same problem, so any fix here should be applied
-            # over there as well.
-            return lang
-        else:
-            return f'{lang}-{territory}'
+        """Converts canonical locale identifiers to moment identifiers.
```
ThiefMaster, line 88:
```suggestion
        """Convert canonical locale identifiers to moment identifiers.
```

indico/web/templates/indico_base.html
```diff
@@ -41,7 +41,7 @@
 
     {% block late_js %}
         <script>
-            moment.locale([{{ session.moment_lang|tojson }}, 'en']);
```
ThiefMaster, line 44: Don't we want to keep this? If a locale doesn't exist in moment we want to fall back to `en` and not whatever else it picks by default...

tomasr8: `session.moment_lang` already defaults to `en` so this shouldn't be needed unless a future version of moment removes/renames some locales. But for safety we could keep it anyway

bin/maintenance/generate_moment_locales.py
```diff
@@ -0,0 +1,105 @@
...
+# LICENSE file for more details.
+
+import json
+from pathlib import Path
+
+import click
+import yaml
+
+
+HEADER = '''\
```
ThiefMaster, line 16: I'd use `.lstrip()` at the end instead of backslash escapes

ThiefMaster (approved):

PR #5449, opened 2022-08-11
Add program code to revision ZIP file name

duartegalvao:
This PR adds the contribution's program code to the revisions' "Download ZIP" file name:
image


indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -388,9 +388,12 @@ def _process(self):
 
                 with file.storage.get_local_path(file.storage_file_id) as filepath:
                     zip_handler.write(filepath, os.path.join(folder_name, filename))
+        file_name = f'revision-{self.revision.id}.zip'
```
ThiefMaster, line 391: we already have a `filename` above... how about `zip_filename` here?

PR #5455, opened 2022-08-17
Insert new categories in the right place if the categories are already sorted

tomasr8:
New categories are currently always appended to the end, even if the list has been sorted.
This change checks whether the categories are sorted (either ascending or descending) and inserts
the new category in the correct place.


indico/modules/categories/controllers/management.py
```diff
@@ -263,10 +263,29 @@ def _get_defaults(self):
 
 
 class RHCreateCategory(RHManageCategoryBase):
+    def _sort_key(self, category):
+        return (natural_sort_key(category.title), category.id)
```
ThiefMaster, line 267:
```suggestion
        return natural_sort_key(category.title), category.id
```

indico/modules/categories/controllers/management.py
```diff
@@ -263,10 +263,29 @@ def _get_defaults(self):
...
+
+    def _get_sort_order(self, categories):
+        sorted_categories = sorted(self.category.children, key=self._sort_key)
+        if categories == sorted_categories:
+            return 'asc'
+        elif categories == list(reversed(sorted_categories)):
```
ThiefMaster, line 273:
```suggestion
        elif categories == sorted_categories[::-1]:
```

untested but i think that does the same ;)

(nothing wrong with your more explicit solution though)

indico/modules/categories/controllers/management.py
```diff
@@ -263,10 +263,29 @@ def _get_defaults(self):
...
         if form.validate_on_submit():
+            sort_order = self._get_sort_order(self.category.children)
             new_category = create_category(self.category, form.data)
+            if sort_order is not None:
+                reverse = sort_order == 'desc'
+                subcategories = sorted(self.category.children, key=self._sort_key, reverse=reverse)
+                for position, category in enumerate(subcategories, start=1):
+                    category.position = position
+                self.category.log(CategoryLogRealm.category, LogKind.change,
+                                  'Content', 'Subcategories sorted', session.user)
```
ThiefMaster, line 288: I don't think we need this here - after all they were sorted before

indico/modules/categories/controllers/management.py
```diff
@@ -370,9 +389,13 @@ def _process(self):
 
 
 class RHSortSubcategories(RHManageCategoryBase):
-    def _process(self):
+
+    @use_kwargs({
+        'categories': fields.List(fields.Int())
```
ThiefMaster, line 394: `required=True`; otherwise this fails since it won't pass the kwarg if it's missing in the request data

indico/web/client/js/utils/sort.js
```diff
@@ -0,0 +1,29 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2022 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+// Same as natural_sort_key
+export function natSortCompare(a, b) {
```
ThiefMaster, line 9: This looks so much cleaner than this strnatcmp i pasted from phpjs a loooong time ago :D

PR #5466, opened 2022-08-23
Add additional event status template hook

SegiNyn:


indico/modules/categories/templates/display/event_list.html
```diff
@@ -131,6 +131,7 @@ <h4 class="{% if month.is_current %}current-month{% endif %}">
                                         {%- trans %}new{% endtrans -%}
                                     </span>
                                 {% endif %}
+                                {{ template_hook('event-status-flag', event=event) }}
```
ThiefMaster, line 134: Since the things here are labels, maybe this name is better? The "flag" part seems to be more what you end up using it for ;)

```suggestion
                                {{ template_hook('event-status-labels', event=event) }}
```

SegiNyn: Changed!

ThiefMaster (approved):

PR #5472, opened 2022-08-26
Limit atom feed to visible events

bpedersen2:
Only export events visible in a category by the atom feed to avoid
unnecessary exposure, especially in the case of exporting to public
displays.

See also[1].

[1] https://talk.getindico.io/t/atom-feeds-should-they-respect-the-visibility-settings/2776


bpedersen2:
Ping...

Would be nice to see this in 3.3

ThiefMaster:
Thanks for the reminder, I saw it recently as well when going through the PRs. I'll merge it for 3.3.

PR #5478, opened 2022-08-29
Add email checker for person links

tomasr8:
- Email is not mandatory, but if used it is checked like in registrations
 - Prevents adding two people with the same email
 - Checks for secondary emails and prevents adding the same user with a different email
 - Warns if the person already has an Indico account
 - Warns if the person is already linked to the event - in this case if there is already a person link, it will be updated with the new data and the old link will be deleted. This takes care of the issue where there are existing person links linked to deleted objects preventing users from changing emails when updating person links.

image
image
image
image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@ Improvements
 - Include speakers in abstract list columns and spreadsheet exports (:pr:`5615`)
 - Add an option to export all events in a series to ical at once (:issue:`5617`, :pr:`5620`)
 - Make it possible to load more events in series management (:pr:`5629`)
+- Check manually entered email addresses of speakers/authors/chairpersons
+  to avoid collisions and inconsistencies (:pr:`5478`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/events/blueprint.py
+++ b/indico/modules/events/blueprint.py
@@ -63,16 +63,16 @@ _bp.add_url_rule('/event/<int:event_id>/api/info-for-series', 'single_event_api'
 
 # Email checker for PersonLinkField
 # Depending on where the person link is used, the person link field generates the correct url.
-# Eeach object has its own url so that we can do proper access checks.
-event_object_url_prefixes = {
+# Each object has its own url so that we can do proper access checks.
+_event_object_url_prefixes = {
     'event': '',
-    'block': '/blocks/<int:block_id>',
+    'session_block': '/sessions/<int:session_id>/blocks/<int:session_block_id>',
     'abstract': '/abstracts/<int:abstract_id>',
     'contribution': '/contributions/<int:contrib_id>',
     'subcontribution': '/contributions/<int:contrib_id>/subcontributions/<int:subcontrib_id>',
 }
-for object_type, prefix in event_object_url_prefixes.items():
-    _bp.add_url_rule('/event/<int:event_id>/check-email' + prefix, 'check_email',
+for object_type, prefix in _event_object_url_prefixes.items():
+    _bp.add_url_rule(f'/event/<int:event_id>{prefix}/check-person-email', 'check_email',
                      RHEventCheckEmail, defaults={'object_type': object_type})
 
 
--- a/indico/modules/events/blueprint.py
+++ b/indico/modules/events/blueprint.py
@@ -60,7 +60,21 @@ _bp.add_url_rule('/event/<int:event_id>/other-view', 'display_other', redirect_v
 # Misc
 _bp.add_url_rule('/event/<int:event_id>/key-access', 'key_access', RHEventAccessKey, methods=('POST',))
 _bp.add_url_rule('/event/<int:event_id>/api/info-for-series', 'single_event_api', RHSingleEventAPI)
-_bp.add_url_rule('/event/<int:event_id>/check-email', 'check_email', RHEventCheckEmail)
+
+# Email checker for PersonLinkField
+# Depending on where the person link is used, the person link field generates the correct url.
+# Eeach object has its own url so that we can do proper access checks.
+event_object_url_prefixes = {
+    'event': '',
+    'block': '/blocks/<int:block_id>',
+    'abstract': '/abstracts/<int:abstract_id>',
+    'contribution': '/contributions/<int:contrib_id>',
+    'subcontribution': '/contributions/<int:contrib_id>/subcontributions/<int:subcontrib_id>',
+}
+for object_type, prefix in event_object_url_prefixes.items():
+    _bp.add_url_rule('/event/<int:event_id>/check-email' + prefix, 'check_email',
+                     RHEventCheckEmail, defaults={'object_type': object_type})
+
 
 # Privacy policy
 _bp.add_url_rule('/event/<int:event_id>/privacy', 'display_privacy', RHDisplayPrivacyPolicy)
--- a/indico/modules/events/blueprint.py
+++ b/indico/modules/events/blueprint.py
@@ -8,7 +8,7 @@
 from indico.modules.events.controllers.admin import (RHCreateEventLabel, RHCreateReferenceType, RHDeleteEventLabel,
                                                      RHDeleteReferenceType, RHEditEventLabel, RHEditReferenceType,
                                                      RHEventLabels, RHReferenceTypes, RHUnlistedEvents)
-from indico.modules.events.controllers.api import RHSingleEventAPI
+from indico.modules.events.controllers.api import RHEventCheckEmail, RHSingleEventAPI
 from indico.modules.events.controllers.creation import RHCreateEvent, RHPrepareEvent
 from indico.modules.events.controllers.display import RHDisplayPrivacyPolicy, RHEventAccessKey, RHExportEventICAL
 from indico.modules.events.controllers.entry import event_or_shorturl
@@ -60,6 +60,7 @@ _bp.add_url_rule('/event/<int:event_id>/other-view', 'display_other', redirect_v
 # Misc
 _bp.add_url_rule('/event/<int:event_id>/key-access', 'key_access', RHEventAccessKey, methods=('POST',))
 _bp.add_url_rule('/event/<int:event_id>/api/info-for-series', 'single_event_api', RHSingleEventAPI)
+_bp.add_url_rule('/event/<int:event_id>/check-email', 'check_email', RHEventCheckEmail)
 
 # Privacy policy
 _bp.add_url_rule('/event/<int:event_id>/privacy', 'display_privacy', RHDisplayPrivacyPolicy)
--- a/indico/modules/events/controllers/api.py
+++ b/indico/modules/events/controllers/api.py
@@ -6,13 +6,14 @@
 # LICENSE file for more details.
 
 from flask import jsonify, session
+from werkzeug.exceptions import Forbidden, NotFound
 
 from indico.modules.events import Event
-from indico.modules.events.controllers.base import RHEventBase
 from indico.modules.events.persons.util import check_person_link_email
+from indico.modules.events.util import get_object_from_args
 from indico.util.marshmallow import LowercaseString, ModelField
 from indico.web.args import use_kwargs
-from indico.web.rh import RH
+from indico.web.rh import RH, RHProtected
 
 
 class RHSingleEventAPI(RH):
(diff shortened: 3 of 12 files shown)


indico/modules/events/controllers/api.py
```diff
@@ -25,3 +27,17 @@ def _process(self, event):
...
+class RHEventCheckEmail(RHEventBase):
+    """Check a person's email when editing a person link."""
+
+    @use_kwargs({
+        'email': LowercaseString(required=True),
+    }, location='query')
+    def _process_args(self, email):
+        RHEventBase._process_args(self)
+        self.email = email
+
```
ThiefMaster, line 41: I'd just put this directly on `_process`..

indico/modules/events/fields.py
```diff
@@ -123,13 +123,23 @@ def _get_person_link(self, data):
...
                 person_link.person.user = get_user_by_email(email)
                 if inspect(person).persistent:
                     signals.event.person_updated.send(person_link.person)
             else:
-                raise UserValueError(_('There is already a person with the email {email}').format(email=email))
+                # Update the person link of the existing event person with
+                # the new data and delete the old person link
+                existing_person_link = (
+                    self.person_link_cls.query.filter_by(person=existing_person, object=self.object).first() or
+                    self.person_link_cls(person=existing_person))
```
ThiefMaster, line 137:
```suggestion
                    self.person_link_cls(person=existing_person)
                )
```

indico/modules/events/fields.py
```diff
@@ -123,13 +123,23 @@ def _get_person_link(self, data):
...
+                existing_person_link.populate_from_dict(data, keys=('first_name', 'last_name', 'affiliation',
+                                                                    'affiliation_link', 'address', 'phone',
+                                                                    '_title', 'display_order'))
+                db.session.delete(person_link)
```
ThiefMaster, line 141: I think making changes here is dangerous as form validation may fail but this would still get committed. If we have to keep this, make sure to set `self.commit = False` right after the successful validate on submit in the places where these forms are used (you can find a few places where we already do that)

indico/modules/events/persons/util.py
```diff
@@ -49,3 +51,24 @@ def get_event_person(event, data, create_untrusted_persons=False, allow_external
...
+    email = email.lower().strip()
+    person = event.persons.filter_by(email=email).first()
+    user = get_user_by_email(email)
+
+    if person is not None:
+        return dict(status='warning', conflict='person-already-exists', person_name=person.name)
+    elif user:
+        return dict(status='warning', conflict='user-already-exists')
+    email_err = validate_email_verbose(email)
+    if email_err:
```
ThiefMaster, line 71:
```suggestion
    if email_err := validate_email_verbose(email):
```

indico/web/client/js/react/components/PersonDetailsModal.jsx
```diff
@@ -141,15 +147,20 @@ export default function PersonDetailsModal({
           <Translate as="label">First Name</Translate>
           <FinalInput name="firstName" />
         </Form.Field>
-        <Form.Field>
+        <Form.Field required>
```
ThiefMaster, line 150: This isn't required on purpose. In some cases you may not have a full first and last name for a person.

indico/web/client/js/react/components/PersonDetailsModal.jsx
```diff
@@ -170,11 +181,114 @@ PersonDetailsModal.propTypes = {
...
+    if (conflict === 'person-already-exists') {
+      msg = (
+        <Translate>
+          This entry will be associated with the user{' '}
+          <Param name="name" wrapper={<strong />} value={data.person_name} />.
+        </Translate>
+      );
+    } else if (conflict === 'user-already-exists') {
+      msg = Translate.string(
+        'An Indico user with this email already exists. You can alternatively add them using the search option.'
```
ThiefMaster, line 243: Could we add a shortcut here that you can search for the user by email and select them with a single click? :)

indico/modules/events/fields.py
tomasr8: Another option which is maybe less nice but definitely safer is to return the event person id in the email checker response.
Then we'd just overwrite the old person id in the person link form.

indico/modules/events/controllers/api.py
```diff
@@ -25,3 +27,14 @@ def _process(self, event):
...
+
+class RHEventCheckEmail(RHEventBase):
+    """Check a person's email when editing a person link."""
+
+    @use_kwargs({
+        'email': LowercaseString(required=True),
+    }, location='query')
+    def _process(self, email):
+        RHEventBase._process_args(self)
```
ThiefMaster, line 39: that does not belong in `_process` - and it's called automatically anyway since `_process_args` is not being overridden here

indico/modules/users/schemas.py
```diff
@@ -39,6 +30,18 @@ def add_country_name(self, data, **kwargs):
         return data
 
 
+class UserSchema(mm.SQLAlchemyAutoSchema):
+    identifier = Function(lambda user: user.identifier)
```
ThiefMaster, line 34: Is this even needed? I'd expect it to be autogenerated correctly... In any case, `identifier = fields.String()` should be enough if it isn't

tomasr8: Not sure why it was there, seems to be generated correctly without..

indico/web/client/js/react/components/PersonLinkField.jsx
```diff
@@ -360,6 +360,12 @@ function PersonLinkField({
               onClose={onClose}
               onSubmit={onSubmit}
               person={persons[selected]}
+              eventId={eventId}
+              otherPersons={
+                selected !== null
+                  ? persons.filter(p => p.email !== persons[selected].email)
```
ThiefMaster, line 366:
```suggestion
                  ? persons.filter(p => p.id !== persons[selected].id)
```

I think using the person id is more reliable here (assuming we have it)

tomasr8: I think in this case it's better to just check the object identity since there are cases where we have neither the id nor email, like when a person is entered manually w/o specifying the email

CHANGES.rst
```diff
@@ -18,6 +18,7 @@ Improvements
   (:issue:`5560`, :pr:`5571`)
 - Allow raw HTML snippets when editing custom conference pages and event
   descriptions (:issue:`5584`, :pr:`5587`)
+- Add email checker for person links (:pr:`5478`)
```
ThiefMaster, line 21: Not sure how clear that is... How about something like "Check email addresses of speakers/authors to avoid collisions and inconsistencies"?

indico/modules/events/blueprint.py
```diff
@@ -60,7 +60,21 @@
 # Misc
 _bp.add_url_rule('/event/<int:event_id>/key-access', 'key_access', RHEventAccessKey, methods=('POST',))
 _bp.add_url_rule('/event/<int:event_id>/api/info-for-series', 'single_event_api', RHSingleEventAPI)
-_bp.add_url_rule('/event/<int:event_id>/check-email', 'check_email', RHEventCheckEmail)
+
+# Email checker for PersonLinkField
+# Depending on where the person link is used, the person link field generates the correct url.
+# Eeach object has its own url so that we can do proper access checks.
+event_object_url_prefixes = {
```
ThiefMaster, line 67: please underscore this; it's pretty specific to this one and unlike the other one shouldn't be accidentally imported

indico/modules/events/blueprint.py
```diff
@@ -60,7 +60,21 @@
 # Misc
 _bp.add_url_rule('/event/<int:event_id>/key-access', 'key_access', RHEventAccessKey, methods=('POST',))
 _bp.add_url_rule('/event/<int:event_id>/api/info-for-series', 'single_event_api', RHSingleEventAPI)
-_bp.add_url_rule('/event/<int:event_id>/check-email', 'check_email', RHEventCheckEmail)
+
+# Email checker for PersonLinkField
+# Depending on where the person link is used, the person link field generates the correct url.
+# Eeach object has its own url so that we can do proper access checks.
```
ThiefMaster, line 66: typo (Eeach)

indico/modules/events/blueprint.py
```diff
@@ -60,7 +60,21 @@
...
+event_object_url_prefixes = {
+    'event': '',
+    'block': '/blocks/<int:block_id>',
+    'abstract': '/abstracts/<int:abstract_id>',
+    'contribution': '/contributions/<int:contrib_id>',
+    'subcontribution': '/contributions/<int:contrib_id>/subcontributions/<int:subcontrib_id>',
+}
+for object_type, prefix in event_object_url_prefixes.items():
+    _bp.add_url_rule('/event/<int:event_id>/check-email' + prefix, 'check_email',
```
ThiefMaster, line 75: Don't you mean `f'/event/<int:event_id>{prefix}/check-person-email'`? ;)

indico/modules/events/controllers/api.py
```diff
@@ -29,11 +30,32 @@ def _process(self, event):
...
+        if self.object_type == 'event' and not self.object.can_manage(session.user):
+            raise Forbidden
+        if self.object_type == 'block' and not self.object.session.can_manage(session.user):
+            raise Forbidden
+        if self.object_type == 'abstract' and not self.object.can_manage(session.user):
+            raise Forbidden
+        if self.object_type == 'contribution' and not self.object.can_edit(session.user):
+            raise Forbidden
+        if self.object_type == 'subcontribution' and not self.object.can_edit(session.user):
+            raise Forbidden
```
ThiefMaster, line 55:
```python
obj = self.object.session if self.object_type == 'block' else self.object
if not obj.can_manage(session.user):
    raise Forbidden
```

indico/modules/events/controllers/api.py
```diff
@@ -29,11 +30,32 @@ def _process(self, event):
...
 
     @use_kwargs({
         'email': LowercaseString(required=True),
     }, location='query')
-    def _process(self, email):
+    def _process_GET(self, email):
```
ThiefMaster, line 60: not needed; we usually only use the specific verbs when mixing multiple methods in the same RH

indico/modules/events/util.py
```diff
@@ -83,6 +84,8 @@ def get_object_from_args(args=None):
         obj = event
     elif object_type == 'session':
         obj = Session.query.with_parent(event).filter_by(id=args['session_id']).first()
+    elif object_type == 'block':
```
ThiefMaster, line 87: maybe `session_block` would be better? Also, I'd nest the URL for session blocks inside the session like we do with subcontributions. ie `/sessions/<int:session_id>/blocks/<int:block_id>` in the blueprint

indico/modules/events/controllers/api.py
tomasr8: I also have to make the distinction between `can_edit` and `can_manage` so something like this:
```python
obj = self.object.session if self.object_type == 'session_block' else self.object
method = 'can_edit' if self.object_type in ('contribution', 'subcontribution') else 'can_manage'
if not getattr(obj, method)(session.user):
    raise Forbidden
```
At this point I kind of prefer the original dumb code :smile:

ThiefMaster: OK yeah, not great, i missed the different methods. I guess adding `can_search_persons` methods on those objects is overkill? OTOH, this could also be useful for #5024 which probably needs a similar check...

alternatively if you want to keep the individual checks, at least write them in a way so it also fails when none of the object types match - because right now a typic there would skip the access check altogether. for example, `return` in all the success cases, and if none match you raise Forbidden.

indico/modules/events/blueprint.py
```diff
@@ -61,6 +61,21 @@
...
+# Each object has its own url so that we can do proper access checks.
+_event_object_url_prefixes = {
+    'event': '',
+    'session_block': '/sessions/<int:session_id>/blocks/<int:session_block_id>',
+    'abstract': '/abstracts/<int:abstract_id>',
+    'contribution': '/contributions/<int:contrib_id>',
+    'subcontribution': '/contributions/<int:contrib_id>/subcontributions/<int:subcontrib_id>',
+}
+for object_type, prefix in _event_object_url_prefixes.items():
+    _bp.add_url_rule(f'/event/<int:event_id>/{prefix}/check-person-email', 'check_email',
```
ThiefMaster, line 75: remove the slash before `{prefix}`, right now you get two of them

indico/modules/events/util.py
```diff
@@ -83,6 +84,8 @@ def get_object_from_args(args=None):
         obj = event
     elif object_type == 'session':
         obj = Session.query.with_parent(event).filter_by(id=args['session_id']).first()
+    elif object_type == 'session_block':
+        obj = SessionBlock.get(args['session_block_id'])
```
ThiefMaster, line 88: please include the session id in the query

indico/modules/users/schemas.py
```diff
@@ -39,6 +30,17 @@ def add_country_name(self, data, **kwargs):
         return data
 
 
+class UserSchema(mm.SQLAlchemyAutoSchema):
+    class Meta:
+        model = User
+        fields = ('id', 'identifier', 'first_name', 'last_name', 'email', 'affiliation', 'affiliation_id',
+                  'address', 'title', 'affiliation_meta', 'full_name', 'phone', 'avatar_url')
```
ThiefMaster, line 37: does this reveal address/phone in places where we didn't reveal it before (ie users other than yourself)?

ThiefMaster (approved):

ThiefMaster (approved):

PR #5479, opened 2022-08-30
Fix incorrect width & height calculation in badges

tomasr8:
When calculating how many columns of badges fit on a page,
the column margin between badges is counted for each badge while it should only count (N-1)-times.

### DIFF
--- a/indico/modules/events/registration/badges.py
+++ b/indico/modules/events/registration/badges.py
@@ -51,9 +51,11 @@ class RegistrantsListToBadgesPDF(DesignerPDFBase):
         config = self.config
 
         available_width = self.width - (config.left_margin + config.right_margin + config.margin_columns) * cm
-        n_horizontal = int(available_width / ((self.tpl_data.width_cm + config.margin_columns) * cm))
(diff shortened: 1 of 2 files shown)


indico/modules/events/registration/badges.py
```diff
@@ -51,7 +51,8 @@ def _build_pdf(self, canvas):
         config = self.config
 
         available_width = self.width - (config.left_margin + config.right_margin + config.margin_columns) * cm
-        n_horizontal = int(available_width / ((self.tpl_data.width_cm + config.margin_columns) * cm))
+        n_horizontal = int((available_width + config.margin_columns * cm) /
```
ThiefMaster, line 54:
```suggestion
        n_horizontal = int((available_width + config.margin_columns*cm) /
```

indico/modules/events/registration/badges.py
```diff
@@ -147,7 +148,8 @@ def _build_pdf(self, canvas):
...
```
ThiefMaster, line 151:
```suggestion
        n_horizontal = int((available_width + config.margin_columns*cm) /
```

PR #5484, opened 2022-09-01
Add tests for plaintext mails for roombooking and users

Leats:
I accidentally forgot the tests for roombooking's and users' plaintext emails.
So in addition to #5482, some more tests following :)

### DIFF
--- a/indico/modules/rb/notifications/notifications_test.py
+++ b/indico/modules/rb/notifications/notifications_test.py
@@ -1,0 +1,137 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2022 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+import os
+from datetime import date, datetime
+from pathlib import Path
+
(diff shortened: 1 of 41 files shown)


indico/modules/rb/notifications/notifications_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@pytest.mark.parametrize('snapshot_name', (
+    'reservation_info.txt', 'reservation_key_info.txt'
+))
+def test_reservation_emails_plaintext(snapshot, snapshot_name, dummy_reservation):
+    dummy_reservation.id = 0
+    dummy_reservation.room.key_location = 'In a pocket dimension reachable via interpretative dance.'
+    dummy_reservation.start_dt = datetime(2022, 11, 11, 13, 37)
+    dummy_reservation.end_dt = datetime(2022, 11, 11, 14, 37)
```
ThiefMaster, line 129: With that many changes I'd use the `create_reservation` fixture and just pass the data to it as kwargs (same in similar places)

(I think this can/should be done especially in places where the ID gets changed manually)

PR #5485, opened 2022-09-01
Make anchor elements focusable

bxo11:
Make more elements focusable by adding a href attribute

### DIFF
--- a/indico/modules/events/abstracts/templates/display/_common.html
+++ b/indico/modules/events/abstracts/templates/display/_common.html
@@ -32,14 +32,13 @@
             {% if can_submit_abstracts %}
                 <div class="toolbar">
                     {% if session.user and not redirect_first %}
-                        <a class="i-button highlight js-show-abstract-form"
+                        <button class="i-button highlight js-show-abstract-form"
                            data-title="{% trans %}Submit new abstract{% endtrans %}"
                            data-href="{{ url_for('abstracts.submit', event) }}"
                            data-confirm-close-unsaved
-                           data-ajax-dialog
-                           href="#">
+                           data-ajax-dialog>
                             {%- trans %}Submit new abstract{% endtrans -%}
-                        </a>
+                        </button>
                     {% else %}
                         {% set submit_url = url_for('abstracts.call_for_abstracts', event, _anchor='submit-abstract') %}
                         {% set url = submit_url if session.user else url_for_login(submit_url) %}
--- a/indico/modules/events/abstracts/templates/display/_common.html
+++ b/indico/modules/events/abstracts/templates/display/_common.html
@@ -36,7 +36,8 @@
                            data-title="{% trans %}Submit new abstract{% endtrans %}"
                            data-href="{{ url_for('abstracts.submit', event) }}"
                            data-confirm-close-unsaved
-                           data-ajax-dialog>
+                           data-ajax-dialog
+                           href="#">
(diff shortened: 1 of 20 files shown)


indico/modules/events/contributions/templates/contrib_list_filter.html
```diff
@@ -40,7 +40,7 @@
             </div>
         </div>
         <div class="toolbar">
-            <a class="i-button js-reset-btn icon-close warning" type="button">
+            <a class="i-button js-reset-btn icon-close warning" type="button" href="#">
```
ThiefMaster, line 43: `<a type=button>` makes no sense; can you remove that attribute? Also, I think `href=""` is enough - that way you also immediately notice when the default event is not being prevented (we never want `#` to show up in the URL, so with an empty href it'd reload the page ie be much more noticeable ;))

bxo11: Using `href=""` causes closing popup window after clicking this element. Maybe it should be replaced with `<button>` instead of `<a>`?

ThiefMaster: If using `href=""` closes the popup window, are you sure `href="#"` doesn't append `#` to the URL? This sounds exactly like what I mentioned with the default not being prevented..

ThiefMaster: I just checked and indeed, clicking the button adds a `#` to the URL. You need to add an evt preventDefault to the `$('.js-reset-btn').on('click', function() {` handler. Once that's done, `href="#"` vs `href=""` won't make a noticeable difference.

ThiefMaster: However, the better option is to simply replace the `a` tag with a `button` tag. There's already `type="button"` on it so I think it's just an oversight.

ThiefMaster: I updated this in master since there were a few more places with this issue. So when rebasing and likely getting a conflict you can simply remove this change here.

ThiefMaster:
There are some conflicts with your other PR. Can you please rebase and fix them?

indico/modules/events/abstracts/templates/display/_common.html
```diff
@@ -36,7 +36,8 @@
                            data-title="{% trans %}Submit new abstract{% endtrans %}"
                            data-href="{{ url_for('abstracts.submit', event) }}"
                            data-confirm-close-unsaved
-                           data-ajax-dialog>
+                           data-ajax-dialog
+                           href="#">
```
ThiefMaster, line 40: I think any `a.i-button` elements may be suitable for being replaced with actual `<button>` tags.

tomasr8 (approved):

indico/modules/events/abstracts/templates/display/_common.html
ThiefMaster: @bxo11 would you have time to look into this option? I think it's cleaner (also from a semantic point of view) than adding hrefs :)

ThiefMaster (approved):

ThiefMaster:
There are many more instances of a.i-button that could most likely be replaced with actual button tags. Searching for `<a.*class.*i-button` (with regex enabled) finds over 100 such places. So if you ever feel like updating those as well... ;)

PR #5494, opened 2022-09-14
Make action icons stack horizontally

tomasr8:
In `Administration -> IP networks` & `News`

Before:
image
After:
image

### DIFF
--- a/indico/web/client/styles/modules/_admin.scss
+++ b/indico/web/client/styles/modules/_admin.scss
@@ -28,7 +28,6 @@
 }
 
 table.ip-network-group .actions {
-  width: 3em;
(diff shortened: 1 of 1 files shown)


indico/web/client/styles/modules/_admin.scss
```diff
@@ -28,7 +28,6 @@
 }
 
 table.ip-network-group .actions {
```
ThiefMaster, line 30: Why is that class even used on the "News" page? 🙈

tomasr8: no idea, but it works so I didn't want to touch it :smile:

PR #5499, opened 2022-09-15
Add a setting to disable entering persons manually

duartegalvao:
This PR adds a setting to disable the option of submitters writing in persons manually in the PersonLinkField.

The setting is located in the Participant Roles page in the management area:
image
image

Disabled:
image

Enabled:
image

Progress:
- [x] Add setting in the management area
- [x] Hide the "enter manually" button when the setting is disabled
- [x] Add server side check

### DIFF
--- a/indico/modules/events/persons/__init__.py
+++ b/indico/modules/events/persons/__init__.py
@@ -9,6 +9,7 @@ from flask import session
 
 from indico.core import signals
 from indico.core.logger import Logger
+from indico.modules.events.settings import EventSettingsProxy
 from indico.util.i18n import _
 from indico.web.flask.util import url_for
 from indico.web.menu import SideMenuItem
@@ -38,3 +39,8 @@ def _get_placeholders(sender, person, event, register_link=False, **kwargs):
     yield ContributionsPlaceholder
     if register_link:
         yield RegisterLinkPlaceholder
+
+
+persons_settings = EventSettingsProxy('persons', {
+    'allow_custom_persons': True,  # Submitters can enter persons manually on person lists
+})
--- a/indico/modules/events/persons/__init__.py
+++ b/indico/modules/events/persons/__init__.py
@@ -42,5 +42,5 @@ def _get_placeholders(sender, person, event, register_link=False, **kwargs):
 
 
 persons_settings = EventSettingsProxy('persons', {
-    'allow_custom_persons': True,  # Submitters can enter persons manually on person lists
+    'disallow_custom_persons': False,  # Disallow manually entering persons on person lists
 })
--- a/indico/modules/events/persons/controllers.py
+++ b/indico/modules/events/persons/controllers.py
@@ -31,8 +31,8 @@ from indico.modules.events.management.controllers import RHManageEventBase
 from indico.modules.events.models.persons import EventPerson
 from indico.modules.events.models.principals import EventPrincipal
 from indico.modules.events.models.roles import EventRole
-from indico.modules.events.persons import logger
-from indico.modules.events.persons.forms import EmailEventPersonsForm
+from indico.modules.events.persons import logger, persons_settings
+from indico.modules.events.persons.forms import EmailEventPersonsForm, ManagePersonListsForm
 from indico.modules.events.persons.operations import update_person
 from indico.modules.events.persons.schemas import EventPersonSchema, EventPersonUpdateSchema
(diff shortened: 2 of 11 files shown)


indico/modules/events/persons/__init__.py
```diff
@@ -38,3 +39,8 @@ def _get_placeholders(sender, person, event, register_link=False, **kwargs):
     yield ContributionsPlaceholder
     if register_link:
         yield RegisterLinkPlaceholder
+
+
+author_list_settings = EventSettingsProxy('author_lists', {
```
ThiefMaster, line 44: I'd go for `persons_settings` and `persons` here. It's not just authors.

ThiefMaster (commented):
I think right now there's no "enforcement" of this, it's just disabling the option in the UI?

duartegalvao:
> I think right now there's no "enforcement" of this, it's just disabling the option in the UI?

Right now yes, it's the next item on the progress list

indico/modules/events/persons/forms.py
```diff
@@ -38,3 +38,9 @@ def __init__(self, *args, **kwargs):
 
     def is_submitted(self):
         return super().is_submitted() and 'submitted' in request.form
+
+
+class ManagePersonListsForm(IndicoForm):
+    allow_custom_persons = BooleanField(_('Submitters can enter persons manually'), widget=SwitchWidget(),
```
ThiefMaster, line 44: Should we negate this ("Disallow manually entering...")? This is (outside JACoW) the much less common option...

indico/modules/events/persons/forms.py
```diff
@@ -38,3 +38,9 @@ def __init__(self, *args, **kwargs):
...
+                                        description=_('If disabled, submitters can only list existing Indico users on '
+                                                      'person lists.'))
```
ThiefMaster, line 46: I don't think that term is very clear. How about  "as speakers/authors"?

indico/modules/events/persons/controllers.py
```diff
@@ -460,3 +460,15 @@ def _process(self, exact, **criteria):
...
+
+
+class RHManagePersonLists(RHManageEventBase):
+    """Dialog to configure person list settings."""
+
+    def _process(self):
+        form = ManagePersonListsForm(allow_custom_persons=persons_settings.get(self.event, 'allow_custom_persons'))
+        if form.validate_on_submit():
+            persons_settings.set(self.event, 'allow_custom_persons', form.allow_custom_persons.data)
+            flash(_('Submitter edit settings changed successfully'), 'success')
```
ThiefMaster, line 472: copy&paste and not updated? ;)

ThiefMaster (commented):
IIRC we also need to disallow "edit" after adding someone via the search

indico/modules/events/persons/forms.py
```diff
@@ -38,3 +38,9 @@ def __init__(self, *args, **kwargs):
 
     def is_submitted(self):
         return super().is_submitted() and 'submitted' in request.form
+
+
+class ManagePersonListsForm(IndicoForm):
+    disallow_custom_persons = BooleanField(_('Disallow manually entering persons'), widget=SwitchWidget(),
+                                           description=_('Only allow submitters to list existing Indico users as '
```
ThiefMaster, line 45: What about "Prohibit submitters from new speakers/authors manually and only allow searching for existing users."? That way it removes the focus from "indico users" ("users with no indico account" is a thing when searching e.g. the sso directory)

indico/modules/events/persons/templates/management/person_list.html
```diff
@@ -8,6 +8,25 @@
...
+        <ul class="i-dropdown">
+            <li>
+                <a href="#"
+                   title="{% trans %}Manage person lists{% endtrans %}"
+                   data-reload-after="customData"
+                   data-title="{% trans %}Manage person lists{% endtrans %}"
+                   data-href="{{ url_for('.manage_person_lists', event) }}"
+                   data-qtip-position="right"
+                   data-ajax-dialog>
+                    {% trans %}Person lists{% endtrans %}</a>
```
ThiefMaster, line 24: Not a fan of this title. Can't think of a better one though....

duartegalvao: Maybe something like "submitters' permissions"?

ThiefMaster (approved):

PR #5491, opened 2022-09-10
Reactify taking minutes

Leats:
In preparation for #3386.

PR #5500, opened 2022-09-15
Allow taking minutes in markdown

### LINKED ISSUE #3386: Take minutes in markdown
Markdown seems to be supported in a lot of places in indico, but I can't figure out how to enter minutes in markdown format. I seem to only have the option to use the built-in HTML based editor. Could we add an option to add the minutes as markdown? 

I can get around this using browser extensions (i.e. MarkdownHere) but given the wide support that indico has for markdown it seems like it _might_ be a trivial extension.

Leats:
Depends on #5491.
Closes #3386.
Users can choose in their preferences to use a markdown editor instead of the html editor. Depending on whether or not the note is created

Uses the React Markdown Editor Lite, which was suggested by @pferreir.

Screenshot of the Markdown Editor Modal


indico/web/client/js/react/components/notes/NoteEditor.jsx
```diff
@@ -14,35 +16,91 @@ import {handleSubmitError} from 'indico/react/forms';
...
   const [loading, setLoading] = useState(false);
   const [allowWithoutChange, setAllowWithoutChange] = useState(false);
+  const [renderMode, setRenderMode] = useState(undefined);
+
+  const combineNotes = (resp, mode) => {
+    if (mode === 'markdown') {
+      setCurrentInput(
+        resp.data.notes
+          .map(note => `#- ${_.escape(note.objectTitle)}\n${note.source}`)
+          .join('\n***\n\n')
```
ThiefMaster, line 39: TIL that you can also use `***` to create a horizontal rule in markdown... I'd use the more commonly used `---` though

indico/web/client/js/react/components/notes/NoteEditor.jsx
```diff
@@ -14,35 +16,91 @@ import {handleSubmitError} from 'indico/react/forms';
 import {FinalModalForm} from 'indico/react/forms/final-form';
 import {Translate} from 'indico/react/i18n';
 import {indicoAxios, handleAxiosError} from 'indico/utils/axios';
+import {camelizeKeys} from 'indico/utils/case';
 
+import {FinalMarkdownEditor} from '../MarkdownEditor';
 import {FinalTextEditor} from '../TextEditor';
 
+// import './NoteEditor.module.scss';
```
ThiefMaster, line 24: what was the problem with this?

Leats: Oops, nothing, but that was a remnant of the time where I had the full Markdown Editor in that file. It's now `MarkdownEditor.module.scss` since it's all styling for that one!

indico/modules/users/controllers.py
```diff
@@ -354,6 +354,11 @@ def _process(self):
         return WPUser.render_template('preferences.html', 'preferences', user=self.user, form=form)
 
 
+class RHUserPreferencesMarkdownAPI(RHUserBase):
+    def _process(self):
+        return jsonify(session.user.settings.get('use_markdown_for_minutes'))
```
ThiefMaster, line 359:
```suggestion
        return jsonify(self.user.settings.get('use_markdown_for_minutes'))
```

not really relevant here but all these user endpoints are defined with `/user/<id>/...` as well in the blueprint (check the `with` block there), so they need to use the correct user in case a user id it specified

ThiefMaster (approved):

adavidzh:
Thank you all for this improvement!

PR #5507, opened 2022-09-21
Fix privacy information dropdown on Safari

duartegalvao:
This PR fixes a bug in which the CSS-based privacy information dropdown didn't open on Safari. This was caused due to safari requiring a `tabindex` attribute in order to make elements focusable.

### DIFF
--- a/indico/web/client/styles/themes/indico.scss
+++ b/indico/web/client/styles/themes/indico.scss
@@ -407,6 +407,7 @@ ul.day-list {
   .privacy-dropdown {
     display: inline-block;
     position: relative;
+    outline: none;
 
     .menu {
       display: none;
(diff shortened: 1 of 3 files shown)


indico/web/client/styles/themes/indico.scss
```diff
@@ -407,6 +407,7 @@ ul.day-list {
   .privacy-dropdown {
     display: inline-block;
     position: relative;
+    outline: none;
```
ThiefMaster, line 410: why do we need/want to remove the outline here?

duartegalvao: Safari adds this weird blue outline when you focus on elements:
image
And this doesn't affect anything on other browsers afaik

duartegalvao: Should I add a comment there specifying this is for Safari-specific behaviour?

ThiefMaster (approved):

PR #5512, opened 2022-09-28
Fix ckeditor full screen plugin when rendered in <Modal>

tomasr8:
Removes  `transform` from the `will-change` property.

Pretty hacky and I don't fully understand _why_ it works, but this should only effect performance (if at all) and not appearance so it should be safe.


ThiefMaster:
Considering the much larger modal size for the minutes ckeditor (once the markdown editor PR is merged) I think the better solution would be to add an option to disable the fullscreen plugin and do so for the minute editor

PR #5514, opened 2022-09-28
Add an option to avoid git checks. Useful when building from a tarball

andriish:
Add an option to avoid git checks. Useful when building from a tarball.

### DIFF
--- a/bin/maintenance/build-wheel.py
+++ b/bin/maintenance/build-wheel.py
@@ -243,7 +243,8 @@ def build_indico(obj, assets, add_version_suffix, ignore_unclean, no_git):
     if no_git:
         clean, output = True, None
         warn('Assuming clean non-git working tree')
-        add_version_suffix = False
+        if add_version_suffix: 
+            fail('The --no-git option cannot be used with --add-version-suffix', verbose_msg=output)
     else:
         # check for unclean git status
         clean, output = git_is_clean_indico()
@@ -254,7 +255,8 @@ def build_indico(obj, assets, add_version_suffix, ignore_unclean, no_git):
     if no_git:
         clean, output = True, None
         warn('Assuming clean non-git package')
-        add_version_suffix = False
+        if add_version_suffix: 
+            fail('The --no-git option cannot be used with --add-version-suffix', verbose_msg=output)
     else:
         # check for git-ignored files included in the package
         clean, output = package_is_clean_indico()
@@ -301,7 +303,8 @@ def build_plugin(obj, assets, plugin_dir, add_version_suffix, ignore_unclean, no
         if no_git:
             clean, output = True, None
             warn('Assuming clean non-git working tree')
-            add_version_suffix = False
+            if add_version_suffix: 
+                fail('The --no-git option cannot be used with --add-version-suffix', verbose_msg=output)
         else:
             clean, output = git_is_clean_plugin()
             if not clean and ignore_unclean:
--- a/bin/maintenance/build-wheel.py
+++ b/bin/maintenance/build-wheel.py
@@ -289,7 +289,7 @@ def _plugin_has_assets(plugin_dir):
 @click.option('--no-assets', 'assets', is_flag=True, flag_value=False, default=True, help='skip building assets')
 @click.option('--add-version-suffix', is_flag=True, help='Add a local suffix (+yyyymmdd.hhmm.commit) to the version')
 @click.option('--ignore-unclean', is_flag=True, help='Ignore unclean working tree')
-@click.option('--no-git', is_flag=True, help='Skip checks and assume clean non-git working tree')
+@click.option('--no-git', is_flag=True, help='Skip git checks and assume clean working tree')
 @click.pass_obj
 def build_plugin(obj, assets, plugin_dir, add_version_suffix, ignore_unclean, no_git):
     """Build a plugin wheel.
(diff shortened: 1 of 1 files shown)


bin/maintenance/build-wheel.py
```diff
@@ -310,8 +324,9 @@ def build_plugin(obj, assets, plugin_dir, add_version_suffix, ignore_unclean):
 @click.argument('plugins_dir', type=click.Path(exists=True, file_okay=False, resolve_path=True))
 @click.option('--add-version-suffix', is_flag=True, help='Add a local suffix (+yyyymmdd.hhmm.commit) to the version')
 @click.option('--ignore-unclean', is_flag=True, help='Ignore unclean working tree')
+@click.option('--assume-clean', is_flag=True, help='Assume clean working tree')
```
ThiefMaster, line 327: The change seems reasonable. Just wondering about this naming. What about something like `--no-git-check` or `--no-git` which seems clearer?

Using `--add_version_suffix` together with that option should probably fail as well, because getting the suffix requires git as well.

andriish: Hi @ThiefMaster ,
did that. Also, hopefully setting add_version_suffix=False with no_git makes sense.

Andrii

bin/maintenance/build-wheel.py
```diff
@@ -235,22 +235,33 @@ def cli(obj, target_dir):
...
-    # check for git-ignored files included in the package
-    clean, output = package_is_clean_indico()
-    if not clean and ignore_unclean:
-        warn('package contains unexpected files listed in git exclusions [ignored]')
-    elif not clean:
-        fail('package contains unexpected files listed in git exclusions', verbose_msg=output)
+    if no_git:
+        clean, output = True, None
+        warn('Assuming clean non-git working tree')
+        add_version_suffix = False
```
ThiefMaster, line 246: I'd rather fail if it's used in this case. Silently ignoring a flag doesn't seem such a good idea.

bin/maintenance/build-wheel.py
```diff
@@ -278,19 +289,25 @@ def _plugin_has_assets(plugin_dir):
 @click.option('--no-assets', 'assets', is_flag=True, flag_value=False, default=True, help='skip building assets')
 @click.option('--add-version-suffix', is_flag=True, help='Add a local suffix (+yyyymmdd.hhmm.commit) to the version')
 @click.option('--ignore-unclean', is_flag=True, help='Ignore unclean working tree')
+@click.option('--no-git', is_flag=True, help='Skip checks and assume clean non-git working tree')
```
ThiefMaster, line 292:
```suggestion
@click.option('--no-git', is_flag=True, help='Skip git checks and assume clean working tree')
```

bin/maintenance/build-wheel.py
andriish: Added 
```
if add_version_suffix: 
            fail('The --no-git option cannot be used with --add-version-suffix', verbose_msg=output)
```

bin/maintenance/build-wheel.py
```diff
@@ -235,22 +235,35 @@ def cli(obj, target_dir):
...
```
ThiefMaster, line 238:
```suggestion
@click.option('--no-git', is_flag=True, help='Skip git checks and assume clean working tree')
```

same as my previous suggestion, i just thought you'd update it in both places ;)

bin/maintenance/build-wheel.py
```diff
@@ -235,22 +235,35 @@ def cli(obj, target_dir):
...
-    clean, output = package_is_clean_indico()
-    if not clean and ignore_unclean:
-        warn('package contains unexpected files listed in git exclusions [ignored]')
-    elif not clean:
-        fail('package contains unexpected files listed in git exclusions', verbose_msg=output)
+    if no_git:
+        clean, output = True, None
+        warn('Assuming clean non-git working tree')
+        if add_version_suffix: 
+            fail('The --no-git option cannot be used with --add-version-suffix', verbose_msg=output)
```
ThiefMaster, line 247: I don't think `verbose_msg=output` makes sense here - `output` is always `None` anyway.

```suggestion
            fail('The --no-git option cannot be used with --add-version-suffix')
```

bin/maintenance/build-wheel.py
```diff
@@ -235,22 +235,35 @@ def cli(obj, target_dir):
...
+        clean, output = git_is_clean_indico()
+        if not clean and ignore_unclean:
+            warn('working tree is not clean [ignored]')
+        elif not clean:
+            fail('working tree is not clean', verbose_msg=output)
+    if no_git:
+        clean, output = True, None
+        warn('Assuming clean non-git package')
+        if add_version_suffix: 
+            fail('The --no-git option cannot be used with --add-version-suffix', verbose_msg=output)
```
ThiefMaster, line 259:
```suggestion
            fail('The --no-git option cannot be used with --add-version-suffix')
```

bin/maintenance/build-wheel.py
```diff
@@ -278,19 +291,26 @@ def _plugin_has_assets(plugin_dir):
...
-        clean, output = git_is_clean_plugin()
-        if not clean and ignore_unclean:
-            warn('working tree is not clean, but ignored')
-        elif not clean:
-            fail('working tree is not clean', verbose_msg=output)
+        if no_git:
+            clean, output = True, None
+            warn('Assuming clean non-git working tree')
+            if add_version_suffix: 
+                fail('The --no-git option cannot be used with --add-version-suffix', verbose_msg=output)
```
ThiefMaster, line 307:
```suggestion
                fail('The --no-git option cannot be used with --add-version-suffix')
```

andriish:
Hi @ThiefMaster ,

now it should be fine.

Andrii

PR #5529, opened 2022-10-10
Don't require app context in hybrid properties

OmeGak:


indico/modules/events/sessions/models/blocks.py
```diff
@@ -14,6 +14,7 @@
 from indico.core.db import db
 from indico.core.db.sqlalchemy.locations import LocationMixin
 from indico.core.db.sqlalchemy.util.models import auto_table_args
+from indico.modules.events.timetable import TimetableEntry
```
ThiefMaster, line 17: Everywhere else we use the full import path: `from indico.modules.events.timetable.models.entries import TimetableEntry`. It's just coincidence that we import it in `__init__.py` as well

PR #5533, opened 2022-10-11
Delete session blocks when deleting event session

OmeGak:


indico/modules/events/sessions/operations.py
```diff
@@ -97,11 +86,10 @@ def update_session_block(session_block, data):
     logger.info('Session block %s modified by %s', session_block, session.user)
 
 
-def delete_session_block(session_block):
+def delete_session_block(session_block, delete_session=True, log=True):
```
ThiefMaster, line 89:
```suggestion
def delete_session_block(session_block, *, delete_session=True, log=True):
```

ThiefMaster (approved):

PR #5545, opened 2022-10-18
Replace a.i-button with button

bxo11:
PR contains updated instances of a.i-button that can be replaced with a button tag.

### DIFF
--- a/indico/modules/attachments/templates/_attachments.html
+++ b/indico/modules/attachments/templates/_attachments.html
@@ -143,9 +143,9 @@
     {%- endif -%}
     {% if management %}
         <div class="i-box-footer text-right js-attachment-actions hide-if-locked">
-            <a href="#" class="i-button icon-folder-plus js-dialog-action"
+            <button class="i-button icon-folder-plus js-dialog-action"
                data-href="{{ url_for('attachments.create_folder', linked_object) }}"
-               data-title="{% trans %}Create a new folder{% endtrans %}">{% trans %}New folder{% endtrans -%}</a>
+               data-title="{% trans %}Create a new folder{% endtrans %}">{% trans %}New folder{% endtrans -%}</button>
         </div>
     {% endif %}
 {%- endmacro %}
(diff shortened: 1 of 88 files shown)


indico/modules/attachments/templates/_attachments.html
```diff
@@ -143,9 +143,9 @@
     {%- endif -%}
     {% if management %}
         <div class="i-box-footer text-right js-attachment-actions hide-if-locked">
-            <a href="#" class="i-button icon-folder-plus js-dialog-action"
+            <button class="i-button icon-folder-plus js-dialog-action"
                data-href="{{ url_for('attachments.create_folder', linked_object) }}"
-               data-title="{% trans %}Create a new folder{% endtrans %}">{% trans %}New folder{% endtrans -%}</a>
+               data-title="{% trans %}Create a new folder{% endtrans %}">{% trans %}New folder{% endtrans -%}</button>
```
ThiefMaster, line 148: Can you please fix the indentation here (and in other places if necessary)? We typically align the attributes below the first attribute, ie like this:

```html
<button class="..."
        foo="bar">
```

ThiefMaster:
Thanks! Did you already test all the places where you changed it to make sure nothing breaks?

indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
```diff
@@ -93,9 +93,9 @@ export default function EditableTypeDashboard() {
               }
               description={Translate.string('Toggle whether users can submit new editables')}
             >
-              <a className="i-button highlight" onClick={toggleSubmission}>
+              <button className="i-button highlight" onClick={toggleSubmission} type="button">
```
bxo11, line 96: I am receiving here a warning that buttons in `EditableTypeDashboard.jsx` file need a type. Is the `button` type right for these elements?

ThiefMaster:
Hey, sorry, but your last commit also indents the children of buttons. After rebasing your PR to the latest master, I started fixing some of them but then realized how many places you changed, so I'll have to ask you to clean up the rest of them...

bxo11:
I cleaned up the changes and indents

ThiefMaster (approved):
I made some fixes hat caused weird display issues while testing, usually due to introducing whitespace in previously-empty elements or looked weird (mixing buttons and a.i-buttons next to each other does not always look the same)

PR #5547, opened 2022-10-19
Reactify email sending dialog

duartegalvao:
This PR continues the refactoring work of #5385.

Progress:

- [x] Using the FinalModalForm
- [x] Add a cancel button (the X is a bit hidden)
- [x] Show placeholders
- [x] Address comments from #5385
- [x] Fix invitation emails
- [x] Fix role emails
- [x] Show email sent confirmation


indico/modules/events/persons/client/js/Email.jsx
```diff
@@ -0,0 +1,264 @@
...
+import {FinalCheckbox, FinalDropdown, FinalInput, handleSubmitError} from 'indico/react/forms';
+import {FinalModalForm} from 'indico/react/forms/final-form';
+import {useIndicoAxios} from 'indico/react/hooks';
+import {Param, Plural, PluralTranslate, Singular, Translate} from 'indico/react/i18n';
+import {indicoAxios} from 'indico/utils/axios';
+import {snakifyKeys} from 'indico/utils/case';
+
+import './Email.module.scss';
+
+const useIdSelector = selector =>
```
ThiefMaster, line 27: That's not a hook, please don't use `useWhatever` naming for "normal" utils.

Also, why is sizing information relevant to getting selected IDs? Is this to skip hidden checkboxes?

indico/modules/events/persons/controllers.py
```diff
@@ -257,47 +257,72 @@ def _process(self):
                                                has_predefined_affiliations=Affiliation.query.has_rows())
 
 
-class RHEmailEventPersons(RHManageEventBase):
+class RHEmailEventPersonsPreview(RHManageEventBase):
+    """Preview an email with EventPersons associated placeholders."""
+
+    @use_kwargs({
+        'body': fields.String(load_default=''),
+        'subject': fields.String(load_default=''),
```
ThiefMaster, line 265:
```suggestion
        'body': fields.String(required=True),
        'subject': fields.String(required=True),
```

Unless I'm mistaken, you always pass these arguments, so there is no need to allow calling this endpoint with them not being sent.

indico/modules/events/persons/controllers.py
```diff
@@ -257,47 +257,72 @@ def _process(self):
...
+    def _process(self, body, subject):
+        person = self.event.persons[0]
+        email_body = replace_placeholders('event-persons-email', body, event=self.event, person=person)
+        email_subject = replace_placeholders('event-persons-email', subject, event=self.event, person=person)
+        tpl = get_template_module('events/persons/emails/custom_email.html', email_subject=email_subject,
+                                  email_body=email_body)
+        return {
+            'subject': tpl.get_subject(),
+            'body': tpl.get_body()
+        }
```
ThiefMaster, line 276:
```suggestion
        return jsonify(subject=tpl.get_subject(), body=tpl.get_body())
```

I don't think we rely on the (somewhat new) implicit JSON encoding behavior of Flask when returning a dict anywhere else

indico/modules/events/persons/controllers.py
```diff
@@ -257,47 +257,72 @@ def _process(self):
...
+
+
+class RHAPIEmailEventPersons(RHManageEventBase):
     """Send emails to selected EventPersons."""
 
     def _process_args(self):
-        self.no_account = request.args.get('no_account') == '1'
         RHManageEventBase._process_args(self)
+        self.no_account = request.args.get('no_account') == '1'
```
ThiefMaster, line 284: Let's move all this stuff to marshmallow (`@use_kwargs`)

indico/modules/events/persons/controllers.py
```diff
@@ -257,47 +257,72 @@ def _process(self):
...
+        self.recipients = set(self._find_event_persons(request.args.getlist('person_id'),
```
ThiefMaster, line 285: There's a reason why this was `request.form` and not `request.args`. ;)

When you select a LOT of people, you get a really long query string with `person_id=...&person_id=...&...` - which may end up being too long (and in any case it's super ugly in access logs). So using POST data for this is much cleaner.

indico/modules/events/persons/client/js/Email.jsx
```diff
@@ -0,0 +1,264 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2022 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import emailAttributesURL from 'indico-url:persons.email_event_persons';
```
ThiefMaster, line 8: `emailAttributesURL` is kind of a strange name for this

indico/modules/events/persons/controllers.py
ThiefMaster: I'd also split this into two RHs (possibly with a base class having the `_process_args` since you need this both for sending the emails and getting the data for showing the form).

That way you can use POST for both cases (since the urls/endpoints are different) without having to use some extra arg to determine whether you want form metadata or send the emails.

indico/modules/events/persons/templates/management/person_list.html
```diff
@@ -100,6 +99,21 @@
                     </div>
                 {% endif %}
                 <div class="group">
+                    <div
+                        id="invitation-email-all"
```
ThiefMaster, line 103: that's not the wrapping style we use elsewhere in html templates ;)

indico/modules/events/persons/templates/management/person_list.html
```diff
@@ -100,6 +99,21 @@
...
+                        data-event-id="{{ event.id }}"
+                        data-person-selector="#event-participants-list tr[data-no-account] input[name=person_id][type=checkbox]"
+                        data-no-account="1">
```
ThiefMaster, line 106: No need to put a value on boolean options; just put `data-no-account` and then check for `.dataset.noAccount !== undefined`

indico/modules/events/persons/client/js/Email.jsx
```diff
@@ -0,0 +1,264 @@
...
+      </Message>
+      <Form.Field required>
+        <Translate as="label">Subject</Translate>
+        {preview.subject}
+      </Form.Field>
+      <Form.Field required>
+        <Translate as="label">Email body</Translate>
+        <TextEditor
+          value={preview.body}
+          config={{showToolbar: false}}
```
ThiefMaster, line 159: Why hide the toolbar? Isn't this an HTML email?

duartegalvao: I'm hiding the toolbar just for the preview, I think it looks cleaner...
Is there a better way to display the preview without using a `<TextEditor>`?

ThiefMaster: nevermind, I missed the fact that this is for the preview

indico/modules/events/persons/client/js/Email.jsx
duartegalvao: Yep. As far as I can tell this is only used when a filter is applied to the author list.
Maybe I should add a comment there, since this check might become pointless once author lists are converted to React (which I think it was talked about)?

indico/modules/events/persons/client/js/Email.jsx
```diff
@@ -0,0 +1,268 @@
...
+    recipients = [],
+    subject: defaultSubject,
+    body: defaultBody,
+    placeholders = [],
+  } = data || {};
+  const count = Object.values(recipientData).reduce((acc, v) => acc + v.length, 0);
+
+  const togglePreview = async ({body, subject}) => {
+    if (!preview) {
+      const {data} = await indicoAxios.post(emailPreviewURL({event_id: eventId}), {body, subject});
```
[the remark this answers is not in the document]
ThiefMaster: something like `body = body.getData ? body.getData() : body;` should fix it

indico/web/client/js/ckeditor.js
```diff
@@ -5,7 +5,12 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
-export const getConfig = ({images = true, imageUploadURL = null, fullScreen = true} = {}) => ({
+export const getConfig = ({
+  images = true,
+  imageUploadURL = null,
+  fullScreen = true,
+  showToolbar = true,
```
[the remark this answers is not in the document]
duartegalvao: It is, to hide the toolbar in the preview

indico/modules/events/persons/client/js/Email.jsx
```diff
@@ -0,0 +1,320 @@
...
+    data: snakifyKeys(recipientData),
+  });
+  const {
+    senders = [],
+    recipients = [],
+    subject: defaultSubject,
+    body: defaultBody,
+    placeholders = [],
+  } = data || {};
+  const count = Object.values(recipientData).reduce((acc, v) => acc + v.length, 0);
```
ThiefMaster, line 126:
```suggestion
  const count = recipients.length;
```

much less complicated and actually works even when there's extra stuff in recipientData ;)

PR #5550, opened 2022-10-21
Add affiliation popup in more places

tomasr8:
Noticed that session display was rendering empty parentheses for missing affiliation and
while fixing it, I decided to add the popup in other places as well.

Abstract reviewing:
image

Contribution speakers:
image

Contribution list
image

Session conveners:
image

Author/Speaker list:
image


indico/modules/events/templates/display/indico/_common.html
```diff
@@ -49,35 +51,27 @@ <h4>{{ item.venue_name }}</h4>
     {% endif %}
 {%- endmacro %}
 
-{% macro render_user_data(user_data, show_title=true, italic_affiliation=false) -%}
```
tomasr8, line 52: I don't think this parameter is used anywhere so I removed it too

indico/modules/events/contributions/templates/display/contribution_display.html
```diff
@@ -2,6 +2,7 @@
 
 {% from 'events/display/common/_conferences.html' import render_attachments %}
 {% from 'events/display/indico/_common.html' import render_location, render_users, render_speakers %}
+{% from 'events/management/_affiliation.html' import render_affiliation %}
```
ThiefMaster, line 5: Maybe we should move it outside that `management` subdir if we start using it outside management?

ThiefMaster (approved):

PR #5556, opened 2022-10-31
Include event label in email reminders

### LINKED ISSUE #5554: The cancelled flag does not appear in reminder e-mails
I added the cancelled flag to meeting https://indico.cern.ch/event/1165478/ . This flag nicely appears on the title and on the weekly summary of meetings, but not in the reminder e-mail. Hence everyone gets a reminder e-mail about a cancelled meeting. 
I reported this to indico-support, but was told it's not a bug and I should delete the reminder e-mail. That's not what I want as I do want to remind people that the meeting is cancelled.

I would like the keyword "cancelled" to appear in the e-mail title.

Alternatively I could change the title of the meeting to Cancelled: DPA WP3 (offline analysis tools) meeting, which is what many people do, and then the e-mail would have a self-explanatory title. But that defeats the point of the cancelled flag.

OmeGak:
Closes #5554.

## Screenshot

image

### DIFF
--- a/indico/modules/events/reminders/templates/emails/event_reminder.txt
+++ b/indico/modules/events/reminders/templates/emails/event_reminder.txt
@@ -4,18 +4,19 @@
 {% set address = event.address %}
 {% set location = event.venue_name %}
 {% set room = event.room_name %}
+{% set label = ' {%s}'|format(event.label.title) if event.label else '' %}
 
 
 {% block subject -%}
-    [{% trans %}Event reminder{% endtrans %}] {{ event.title }} ({{ event.start_dt | format_datetime('short', timezone=event.tzinfo) }} {{ event.timezone }})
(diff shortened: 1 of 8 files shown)


indico/modules/events/reminders/templates/emails/event_reminder.txt
```diff
@@ -7,16 +7,25 @@
 
 
 {% block subject -%}
-    [{% trans %}Event reminder{% endtrans %}] {{ event.title }} ({{ event.start_dt | format_datetime('short', timezone=event.tzinfo) }} {{ event.timezone }})
+    [{% trans %}Event reminder{% endtrans %}] {{ event.title }}
+    {%- if event.label %} {{ '{{{}}}'.format(event.label.title) }} {% endif -%}
+    ({{ event.start_dt | format_datetime('short', timezone=event.tzinfo) }} {{ event.timezone }})
```
ThiefMaster, line 12: I think you lose the space before the `(` when there's no label. Unfortunately the subject is not covered by the snapshots... might be worth at some point to refactor these snapshots so they also include the subject somehow...

ThiefMaster (approved):

PR #5571, opened 2022-11-16
Sanitize HTML pasted into ckeditor

### LINKED ISSUE #5560: Investigate how we can reduce the amount of CSS garbage when people paste from word into ckeditor
We currently allow *all css* in the editor, because using a whitelist has the annoying side-effect of REMOVING styles that are already where or that have been manually added in source mode.

However, this also has the very annoying side-effect that pasting from Outlook, Word and similar tools (basically anything that exposes HTML/Richtext to the clipboard)  has a LOT of extremely ugly CSS which we don't really want.

Maybe there's some way to apply stronger sanitization rules on pasted content in ckeditor5? Or some way to manually run code on paste where we could clean this up?

tomasr8:
closes #5560 

Probably the easiest solution is to use an external sanitizer and sanitize the pasted contents before it gets converted into ckeditor's document model. Otherwise we'd probablty have to use [writers](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_view_view-View.html#function-change) or directly manipulate the model.

This [library](https://www.npmjs.com/package/sanitize-html) seems nice because it can also filter attributes, tags, etc..
If we don't want to add another library and just care about the styles, we could roll our own sanitizer using the `DOMParser` to filter the styles (`DOMParser` is used internally by ckeditor when pasting as well):

```js
export function sanitizeStyles(
  fragment,
  allowedStyles = ['color', 'background-color', 'font-size']
) {
  allowedStyles = new Set(allowedStyles);
  const queue = [...fragment.children];
  while (queue.length) {
    const element = queue.pop();
    const style = element.style;
    const toRemove = [];
    for (let i = 0; i < element.style.length; i++) {
      const name = style.item(i);
      if (!allowedStyles.has(name)) {
        toRemove.push(name);
      }
    }
    for (const name of toRemove) {
      style.removeProperty(name);
    }
    queue.push(...element.children);
  }
  return fragment;
}
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@ Improvements
 - Warn when a user cannot create an event in the current category (:pr:`5572`)
 - Display all contributions in 'My contributions' and not just those with
   submitter privileges (:pr:`5575`)
+- Apply stronger sanitization on rich-text content pasted into CKEditor
+  (:issue:`5560`, :pr:`5571`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/web/client/js/ckeditor.js
+++ b/indico/web/client/js/ckeditor.js
@@ -184,3 +184,23 @@ export const getConfig = ({
         }
       : undefined,
 });
+
+// Sanitize HTML pasted into ckeditor.
+// Removes all but the whitelisted CSS rules.
+export function sanitizeHtml(dirty) {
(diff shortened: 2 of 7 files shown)


tomasr8:
The CSS whitelist is just an example, we probably want to allow more styles than that

ThiefMaster (approved):

PR #5572, opened 2022-11-18
Warn when a user does not have rights to create an event

### LINKED ISSUE #5469: Workflow for creating an event for unauthorized users
Using Indico 3.2.2, the permissions for a category are set to only allow creating events for authorized users. Non-administrators still see the *Create event* button, and only get informed about the missing permissions after filling out the form, and trying to create the event.

tomasr8:
Closes #5469 

When using the 'Create event' in a category where a user cannot create events, a warning is displayed in the form and submit is disabled.

This way, users know right away instead of filling out the form and trying to submit.

image

image

### DIFF
--- a/indico/modules/events/client/js/creation.js
+++ b/indico/modules/events/client/js/creation.js
@@ -58,7 +58,7 @@ import {camelizeKeys} from 'indico/utils/case';
     let multipleOccurrences = false;
 
     protectionMessage.appendTo(options.protectionModeFields.closest('.form-field'));
-    eventCreationMessage.appendTo(options.categoryField.closest('.form-field'));
+    eventCreationMessage.insertAfter($('#category-warning-event-creation-category'));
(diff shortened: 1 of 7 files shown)


indico/modules/events/client/js/creation.js
```diff
@@ -71,6 +96,15 @@ import {camelizeKeys} from 'indico/utils/case';
       protectionMessage.html(elem);
     }
 
+    // Display a warning if the user does not have the rights to create events in the selected category
+    function updateEventCreationMessage() {
+      if (canCreateInSelectedCategory()) {
+        eventCreationMessage.addClass('hidden');
+      } else {
+        eventCreationMessage.removeClass('hidden');
+      }
```
ThiefMaster, line 105:
```suggestion
      eventCreationMessage.goggleClass('hidden', canCreateInSelectedCategory());
```

ThiefMaster (approved):

PR #5575, opened 2022-11-21
Display all contributions in 'My contributions'

tomasr8:
Instead of showing only contributions for which the user is a submitter,
we show all contributions categorized by primary/secondary author, speaker or submitter.

Contributions in the `Submitter` section are only shown if the user neither an author or a speaker.

Not-so-obvious edge case: a user can have submission rights on a contribution w/o having a person link (= not added in the `People` field in contribution settings) - these contributions are included.

image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ Improvements
 
 - Display program codes in 'My contributions' (:pr:`5573`)
 - Warn when a user cannot create an event in the current category (:pr:`5572`)
+- Display all contributions in 'My contributions' and not just those with
+  submitter privileges (:pr:`5575`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/events/contributions/controllers/display.py
+++ b/indico/modules/events/contributions/controllers/display.py
@@ -5,6 +5,7 @@
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
 
+from collections import defaultdict
 from io import BytesIO
 
 from flask import jsonify, redirect, request, session
@@ -21,7 +22,7 @@ from indico.modules.events.contributions.lists import ContributionDisplayListGen
 from indico.modules.events.contributions.models.contributions import Contribution
 from indico.modules.events.contributions.models.persons import AuthorType, ContributionPersonLink
 from indico.modules.events.contributions.models.subcontributions import SubContribution
-from indico.modules.events.contributions.util import (get_contributions_with_user_as_submitter,
+from indico.modules.events.contributions.util import (get_contributions_for_user,
                                                       has_contributions_with_user_as_submitter)
 from indico.modules.events.contributions.views import WPAuthorList, WPContributions, WPMyContributions, WPSpeakerList
 from indico.modules.events.controllers.base import RHDisplayEventBase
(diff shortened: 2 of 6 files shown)


indico/modules/events/contributions/util.py
```diff
@@ -244,6 +244,15 @@ def get_contributions_for_person(event, person, only_speakers=False):
             .all())
 
 
+def get_contributions_for_user(event, user):
+    """Get all contributions for a user in the given event."""
+    return (Contribution.query.with_parent(event)
+            .join(ContributionPersonLink, ContributionPersonLink.contribution_id == Contribution.id)
+            .join(EventPerson, EventPerson.id == ContributionPersonLink.person_id)
```
ThiefMaster, line 251: I don't think you need to specify join criteria here - SA should be able to automatically derive them from the FKs/relationships.

indico/modules/events/contributions/templates/display/user_contribution_list.html
```diff
@@ -1,44 +1,17 @@
...
-                </ul>
-            </div>
-        </div>
-    </section>
+    {{ render_contribution_section(contributions['primary'], title=gettext('Primary author'),
+                                   no_contribs_msg=gettext('You are not the primary author of any contribution.')) }}
+    {{ render_contribution_section(contributions['secondary'], title=gettext('Co-author'),
+                                   no_contribs_msg=gettext('You are not the co-author of any contribution.')) }}
+    {{ render_contribution_section(contributions['speaker'], title=gettext('Speaker'),
+                                   no_contribs_msg=gettext('You are not the speaker of any contribution.')) }}
```
ThiefMaster, line 14: The `_('...')` alias exists in templates as well, and I'd use `contributions.primary` etc (more common in templates instead of using the explicit dict item notation)

indico/modules/events/contributions/controllers/display.py
```diff
@@ -97,9 +98,17 @@ def _check_access(self):
...
-        contributions = get_contributions_with_user_as_submitter(self.event, session.user)
+        contribs = get_contributions_for_user(self.event, session.user)
+        categorized = defaultdict(list)
+        for contrib in contribs:
+            if any(link.person.user == session.user for link in contrib.primary_authors):
+                categorized['primary'].append(contrib)
+            if any(link.person.user == session.user for link in contrib.secondary_authors):
+                categorized['secondary'].append(contrib)
+            if any(link.person.user == session.user for link in contrib.speakers):
+                categorized['speaker'].append(contrib)
```
ThiefMaster, line 109: You could iterate directly over `contrib.person_links` and then just check the roles (`author_type` & `is_speaker`) of that object.

indico/modules/events/contributions/util.py
```diff
@@ -244,6 +244,15 @@ def get_contributions_for_person(event, person, only_speakers=False):
             .all())
 
 
+def get_contributions_for_user(event, user):
```
ThiefMaster, line 247: I guess it's an uncommon cases, but shouldn't we still include contributions where the user is only a submitter w/o having a person link as well?

indico/modules/events/contributions/controllers/display.py
```diff
@@ -97,9 +98,21 @@ def _check_access(self):
...
+        for contrib in contribs:
+            for link in contrib.person_links:
+                if link.is_speaker:
+                    categorized['speaker'].add(contrib)
+                if link.author_type == AuthorType.primary:
+                    categorized['primary'].add(contrib)
+                if link.author_type == AuthorType.secondary:
+                    categorized['secondary'].add(contrib)
+            for principal in contrib.acl_entries:
+                if principal.user == session.user and principal.has_management_permission('submit'):
```
ThiefMaster, line 112: This could be simplified to `if contrib.can_manage(session.user, 'submit', allow_admin=False, check_parent=False)`

CHANGES.rst
```diff
@@ -11,6 +11,7 @@ Improvements
 ^^^^^^^^^^^^
 
 - Display program codes in 'My contributions' (:pr:`5573`)
+- Display all contributions in 'My contributions' categorized by author/speaker/submitter (:pr:`5575`)
```
ThiefMaster, line 14: This line is super long...

```suggestion
- Display all contributions in 'My contributions' and not just those with
  submitter privileges (:pr:`5575`)
```

How about this instead?

indico/modules/events/contributions/util.py
```diff
@@ -244,6 +244,21 @@ def get_contributions_for_person(event, person, only_speakers=False):
...
+    """Get all contributions for a user in the given event.
+
+    Gets all contributions where the user is a primary/secondary author, speaker or a submitter.
+    """
+    condition = db.or_(EventPerson.user == user,
+                       Contribution.acl_entries.any(db.and_(ContributionPrincipal.has_management_permission('submit'),
+                                                            ContributionPrincipal.user == user)))
+    return (Contribution.query.with_parent(event)
+            .join(ContributionPersonLink)
+            .join(EventPerson)
```
ThiefMaster, line 257:
```suggestion
            .join(EventPerson, outer=True)
```

otherwise the whole query won't return a user with no matching EventPerson (but please test that case just in case)

ThiefMaster:
What do you think about hiding contributions in the "submitter" list if it's already listed anywhere else?

PR #5588, opened 2022-11-29
Refactor "author list" dialog

duartegalvao:
This PR refactors all of the "author list" dialogs, now renamed to "email authors". This dialog is used in abstract, contribution, paper and session lists.

Progress:

- [X] Convert dialog to a React-SUI Modal
- [X] Fetch authors' data
- [X] Display authors' data
- [x] Implement sorting by column
- [x] Allow selecting authors
- [x] Add email button
- [x] Implement filters
- [x] Refactor authors list button implementation
- [x] Show success message in authors list modal
- [x] Use identifiers instead of IDs


indico/modules/events/management/client/js/EmailAuthors.jsx
```diff
@@ -0,0 +1,211 @@
...
+    text: Translate.string('Co-authors'),
+    isMatch: person => person.secondaryAuthor,
+  },
+  notRegistered: {
+    text: Translate.string('Not registered'),
+    isMatch: person => !person.registered,
+  },
+};
+
+const getIds = selector =>
```
duartegalvao, line 63: Since this function appears in more than one place now, should it be turned into a util? If so, where should I put it?

ThiefMaster: I'll have to look around, for now I think duplicating it isn't a big deal since it's pretty small. If I find a good place to put it (and a more descriptive name) I'll let you know

indico/modules/events/management/controllers/base.py
```diff
@@ -67,10 +66,15 @@ def _process(self):
...
-            person_roles = contribution_persons_dict[contrib_person.person]
-            person_roles['speaker'] |= contrib_person.is_speaker
-            person_roles['primary_author'] |= contrib_person.author_type == AuthorType.primary
-            person_roles['secondary_author'] |= contrib_person.author_type == AuthorType.secondary
-            person_roles['not_registered'] = contrib_person.person not in registered_persons
-
-        return jsonify_template(self.template, event_persons=contribution_persons_dict, event=self.event)
+            person_dict = contribution_persons_dict[contrib_person.person.id]
+            person_dict['id'] = contrib_person.person.id
+            person_dict['id_type'] = 'person' if contrib_person.person.user else 'user'
```
ThiefMaster, line 71: `id` is always a person id, so maybe this should be just a `type`?

indico/modules/events/management/controllers/base.py
```diff
@@ -67,10 +66,15 @@ def _process(self):
...
+            person_dict['full_name'] = contrib_person.person.full_name
+            person_dict['email'] = contrib_person.person.email
+            person_dict['affiliation'] = contrib_person.person.affiliation
+            person_dict['speaker'] |= contrib_person.is_speaker
+            person_dict['primary_author'] |= contrib_person.author_type == AuthorType.primary
+            person_dict['secondary_author'] |= contrib_person.author_type == AuthorType.secondary
+            person_dict['registered'] = contrib_person.person in registered_persons
+
+        return jsonify(event_persons=[p for _, p in sorted(contribution_persons_dict.items())])
```
ThiefMaster, line 80: I would avoid `_` here since it's already used for i18n. could be `__` instead.

indico/modules/events/management/templates/_base_person_list.html
```diff
@@ -1,3 +1,5 @@
+<!-- TODO delete this file -->
```
ThiefMaster, line 1: 👀  (just so there's an unresolved comment and we don't forget it in the end)

indico/modules/events/persons/client/js/Email.jsx
```diff
@@ -56,8 +60,13 @@ export function EmailButton({
   return (
     <>
       {!triggerSelector && (
-        <Translate as={Button} onClick={() => setOpen(true)}>
-          Send email
+        <Translate
+          as={Button}
+          onClick={() => setOpen(true)}
+          disabled={!(personIds.length || userIds.length || roleId)}
+          {...rest}
```
ThiefMaster, line 67: I don't really like this since Translate also has some properties which aren't forwarded to the element (`as`, `context`). Maybe better to be explicit here and using `<Button><Translate></Translate></Button>`?

ThiefMaster: It also breaks i18n extraction since `<Translate>` does not support expanded arguments (`context` could be provided through it, but this is not allowed because it needs to be known at extraction time w/o executing the code)

indico/modules/events/management/client/js/EmailAuthors.jsx
```diff
@@ -0,0 +1,211 @@
...
+      <Modal.Header content={Translate.string('Email authors')} />
+      <Modal.Content>
+        {successMessage}
+        <div>
+          <Translate
+            as="div"
+            style={{display: 'inline-block', fontWeight: 'bold', padding: '0.5em'}}
+          >
+            Filter by
+          </Translate>
```
ThiefMaster, line 165: SUI doesn't have something nicer to attach a label to a button group? :/

indico/modules/events/management/client/js/EmailAuthors.jsx
```diff
@@ -0,0 +1,211 @@
...
+          />
+        ) : (
+          <Message error>{CONTEXTS[context].emptyListMessage}</Message>
+        )}
+      </Modal.Content>
+      <Modal.Actions>
+        <EmailButton
+          eventId={eventId}
+          personIds={selectedPersons.filter(id =>
+            isVisible(data.eventPersons.filter(p => p.id === id)[0])
```
ThiefMaster, line 190: `data.eventPersons.find(p => p.id === id)`

but maybe using a Map would be better? if you select everything I think you get O(n²) here...

indico/modules/events/management/client/js/EmailAuthors.jsx
```diff
@@ -0,0 +1,211 @@
...
+    );
+  };
+
+  const filterButtons = Object.entries(FILTER_OPTIONS)
+    .filter(([, filter]) => !filter.contexts || filter.contexts.includes(context))
+    .map(([name, filter]) => ({
+      key: name,
+      content: filter.text,
+      onClick: () => toggleFilter(name),
+      active: activeFilters.includes(name),
```
ThiefMaster, line 151: `primary` works better to highlight the active buttons

indico/modules/events/management/client/js/EmailAuthors.jsx
duartegalvao: I don't think so... We could use a disabled button I guess but it would still need custom CSS

ThiefMaster: not a big deal - i don't think the current style is particularly bad ;)

ThiefMaster:
Sorry, but this doesn't work properly yet :/

image
image

Sorting doesn't really work either - at least it seems to be case sensitive.

On a much more positive note, the "registered for the event" indicator is now correct while the previous one was incorrect.

I'm also getting some JS warnings:

> Warning: Failed prop type: The prop `persons[2].identifier` is marked as required in `PersonList`, but its value is `undefined`.
> Warning: Each child in a list should have a unique "key" prop.

PS: I rebased your branch to the latest master and fixed the rebase conflicts (luckily just minor things in Python imports).

indico/modules/events/abstracts/controllers/abstract_list.py
```diff
@@ -226,18 +227,23 @@ def _process(self):
...
+            person_dict['affiliation'] = user_or_person.affiliation
+            person_dict['speaker'] |= abstract_person.is_speaker
+            person_dict['primary_author'] |= abstract_person.author_type == AuthorType.primary
+            person_dict['secondary_author'] |= abstract_person.author_type == AuthorType.secondary
+            person_dict['registered'] = abstract_person.person in registered_persons
         for submitter in submitters:
-            abstract_persons_dict[submitter]['submitter'] |= True
-        return jsonify_template('events/abstracts/management/abstract_person_list.html',
-                                event_persons=abstract_persons_dict, event=self.event)
+            abstract_persons_dict[submitter.id]['submitter'] |= True
```
ThiefMaster, line 245: This is causing the blank entries from my screenshot - we're adding people to the list by ID, but previously we added the user itself and thus had the name etc. So when adding someone here who's not in the dict yet, it's now necessary to populate the name etc. as well.

duartegalvao: My bad, I didn't realize there could be submitters that are not authors. Everything should be fixed now

ThiefMaster (approved):

PR #5591, opened 2022-12-01
Add title to Indico.User

### LINKED ISSUE #5590: PersonLinkField does not prefill title/phone/address when editing people
#### Missing title
To reproduce:
- Use `Add myself` to add yourself to the list of people
- Click on `Edit person` (pencil icon)

The title should be prefilled in the form, but the title field is empty. This one is because `Indico.User` does not have the title.

#### Missing phone/address
To reproduce:
- Use `Search` to add a user
- Click on `Edit person` (pencil icon)

Address and phone is missing here because the `PrincipalsMixin` doesn't serialize them.

It might be worth to reuse the `UserSchema` for these to avoid inconsistencies.

tomasr8:
Closes #5590

### DIFF
--- a/indico/web/assets/vars_js.py
+++ b/indico/web/assets/vars_js.py
@@ -69,6 +69,7 @@ def generate_user_file(user=None):
     else:
         user_vars = {
             'id': user.id,
+            'title': user._title.name,
             'firstName': user.first_name,
             'lastName': user.last_name,
             'fullName': user.full_name,
--- a/indico/web/client/js/jquery/widgets/jinja/person_link_widget.js
+++ b/indico/web/client/js/jquery/widgets/jinja/person_link_widget.js
@@ -27,6 +27,7 @@ import {camelizeKeys} from 'indico/utils/case';
     const persons = JSON.parse(field.value);
     const user = sessionUser &&
       sessionUser.id !== undefined && {
+        title: sessionUser.title,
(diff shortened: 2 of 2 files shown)


indico/web/assets/vars_js.py
```diff
@@ -69,6 +69,7 @@ def generate_user_file(user=None):
     else:
         user_vars = {
             'id': user.id,
+            'title': user._title.name,
```
ThiefMaster, line 72: Can you confirm that works well even if the user has no title?

tomasr8: It does, we have the same logic in `PrincipalsMixin` which is used when you search for people in the person link. But it's probably the best - we use `NoneValueEnumField` in other places, and in the frontend the `none` UserTitle is not in the options for the title dropdown either...

Maybe something like this would be better?
```py
'title': NoneValueEnumField(UserTitle, none_value=UserTitle.none).serialize('_title', user)
```
So we actually get a proper `null` like in user preferences.

(Or just using UserSchema directly but then we need to camelize the keys and handle avatarURL directly..)

ThiefMaster: > Or just using UserSchema directly but then we need to camelize the keys and handle avatarURL directly..

you could subclass the schema and add a post_dump to camelize and handle that one separately

PR #5595, opened 2022-12-06
Use truly-sane-lists markdown extension

### LINKED ISSUE #5592: Markdown minutes bullet list live rendiering is inconsistent with published result
**Describe the bug**
The new markdown minutes feature is awesome! I really appreciate that the team took the time to integrate a live rendering window to the right of the text entry.

Unfortunately the rules to convert bulleted lists to HTML seem inconsistent between the live rendering and the indico view. The live rendering will indent with three spaces, while the result published to the indico event requires 4. There is also an inconsistency if you start a new list immediately after a paragraph.

**To Reproduce**

1. Open the minutes editor in markdown mode.
2. Create the list below.
3. Save the minutes and look at them in the indico event.

```markdown
Wrong bullet list:
- Bullet point
   - Subbullet

The above looks fine in the live rendering, but appears incorrectly in indico. The following works fine though:

Correct bullet list:

- Bullet point
    - Subbullet

There are two changes needed to produce the correct result:

1. There must be a blank line after the paragraph for the list to start. This is not required by the live rendering.
2. There must be 4 spaces per level of indentation, whereas the live rendering only requires 3.
```

**Expected behavior**
The live rendering and published indico rendering should look the same.

**Screenshots**
This is what I see in the live rendering:
image
This is what I see in Indico:
image

**Additional context**
Again, I love being able to use markdown, and the live editor makes learning the syntax super easy. It's a bit of a problem when the live render doesn't match the result though.

ThiefMaster:
closes #5592

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,8 @@ Bugfixes
 - Do not strip custom CSS classes from HTML in CKEditor (:issue:`5584`, :pr:`5585`)
 - Use the instance's default locale instead of "no locale" (US-English) in places
   where no better information is known for email recipients (:pr:`5586`)
+- Make markdown display more consistent between editor preview and server-side
+  rendering (:issue:`5592`, :pr:`5595`)
 
 Internal Changes
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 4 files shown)


ThiefMaster:
Ah damn, that extension supports ONLY two space indentation instead of allowing both two and four space indentation. so not really usable for us.

PR #5597, opened 2022-12-07
Add setting to disable material uploads

duartegalvao:
This PR adds a setting that disallows non-managers of uploading materials to events, contributions and sessions.

The setting was added to the Materials page in the management area:
image

Progress:
- [x] Add setting in management area
- [x] Disallow submitters of uploading materials to contributions
- [x] Disallow chairpersons/speakers of lectures of uploading materials to events

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,8 @@ Improvements
   to avoid collisions and inconsistencies (:pr:`5478`)
 - Add option to use review track as accepted track when bulk-accepting abstracts
   (:pr:`5608`)
+- Add setting to only allow managers to upload attachments to events and
+  contributions (:pr:`5597`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/attachments/controllers/management/event.py
+++ b/indico/modules/attachments/controllers/management/event.py
@@ -14,11 +14,14 @@ from indico.modules.attachments.controllers.management.base import (AddAttachmen
                                                                     DeleteAttachmentMixin, DeleteFolderMixin,
                                                                     EditAttachmentMixin, EditFolderMixin,
                                                                     ManageAttachmentsMixin)
+from indico.modules.attachments.forms import SetUploadPermissionsForm
+from indico.modules.attachments.settings import attachments_settings
 from indico.modules.attachments.util import can_manage_attachments
 from indico.modules.attachments.views import WPEventAttachments, WPPackageEventAttachmentsManagement
-from indico.modules.events.controllers.base import RHEventBase
+from indico.modules.events.controllers.base import EditEventSettingsMixin, RHEventBase
 from indico.modules.events.management.controllers import RHManageEventBase
 from indico.modules.events.util import check_event_locked, get_object_from_args
+from indico.util.i18n import _
 from indico.web.flask.templating import get_template_module
 from indico.web.rh import RHProtected
 
@@ -97,3 +100,12 @@ class RHPackageEventAttachmentsManagement(AttachmentPackageMixin, RHManageEventB
     wp = WPPackageEventAttachmentsManagement
     management = True
     ALLOW_LOCKED = True
+
+
+class RHSetUploadPermissions(EditEventSettingsMixin, RHManageEventBase):
+    settings_proxy = attachments_settings
+    form_cls = SetUploadPermissionsForm
+    success_message = _('Upload permissions changed successfully')
+    log_module = 'Attachments'
+    log_message = 'Upload settings updated'
+    log_fields = {'managers_only': 'Managers only'}
--- a/indico/modules/attachments/util.py
+++ b/indico/modules/attachments/util.py
@@ -61,10 +61,14 @@ def get_attached_items(linked_object, include_empty=True, include_hidden=True, p
 
 def can_manage_attachments(obj, user, allow_admin=True):
     """Check if a user can manage attachments for the object."""
+    from indico.modules.attachments.settings import attachments_settings
     if not user:
         return False
     if obj.can_manage(user, allow_admin=allow_admin):
         return True
+    event = getattr(obj, 'event', None)
+    if event and attachments_settings.get(event, 'managers_only'):
+        return False
     if isinstance(obj, db.m.Event) and obj.can_manage(user, 'submit', allow_admin=allow_admin):
         return True
     if isinstance(obj, db.m.Contribution) and obj.can_manage(user, 'submit', allow_admin=allow_admin):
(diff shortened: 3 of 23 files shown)


indico/modules/attachments/controllers/management/event.py
```diff
@@ -97,3 +101,13 @@ class RHPackageEventAttachmentsManagement(AttachmentPackageMixin, RHManageEventB
     wp = WPPackageEventAttachmentsManagement
     management = True
     ALLOW_LOCKED = True
+
+
+class RHSetUploadPermissions(RHManageEventBase):
+    def _process(self):
+        form = SetUploadPermissionsForm(managers_only=attachments_settings.get(self.event, 'managers_only'))
+        if form.validate_on_submit():
+            attachments_settings.set(self.event, 'managers_only', form.managers_only.data)
```
ThiefMaster, line 110: let's use the get_all/set_multi version here. i don't want to have to update this once another setting gets added there ;)

indico/modules/attachments/__init__.py
```diff
@@ -63,3 +64,8 @@ def _get_attachment_cloner(sender, **kwargs):
 @signals.core.import_tasks.connect
 def _import_tasks(sender, **kwargs):
     import indico.modules.attachments.tasks  # noqa: F401
+
+
+attachments_settings = EventSettingsProxy('attachments', {
```
ThiefMaster, line 69: In most newer code we add this in a `settings.py` file instead of `__init__.py`

CHANGES.rst
```diff
@@ -22,6 +22,8 @@ Improvements
   the file (:issue:`5551`, :pr:`5593`)
 - Add option to email people with specific roles about their contributions
   or abstracts (:pr:`5598`)
+- Add a setting for only allowing managers to upload attachments to events
+  and contributions (:pr:`5597`)
```
ThiefMaster, line 26: Move to 3.2.3 changelog after rebasing

indico/modules/attachments/controllers/management/event.py
```diff
@@ -97,3 +104,20 @@ class RHPackageEventAttachmentsManagement(AttachmentPackageMixin, RHManageEventB
...
+            attachments_settings.set_multi(self.event, form.data)
+            new_settings = attachments_settings.get_all(self.event)
+            flash(_('Upload permissions settings changed successfully'), 'success')
+            changes = {k: (v, new_settings[k]) for k, v in current_settings.items() if v != new_settings[k]}
+            if changes:
+                log_fields = {'managers_only': 'Only managers can upload attachments'}
+                self.event.log(EventLogRealm.management, LogKind.change, 'Attachments', 'Settings updated',
+                               session.user, data={'Changes': make_diff_log(changes, log_fields)})
+            return jsonify_data()
+        return jsonify_form(form)
```
ThiefMaster, line 123: This is now the third place where we have this almost identical code; possibly more if we count places where we hardcoded single/few individual settings. Let's add a mixin (maybe in events/controllers/base.py) like this and make use of it in those places:

```python
class EditEventSettingsMixin:
    settings_proxy = None
    form_cls = None
    success_message = None
    log_module = None
    log_message = None
    log_fields = None
    
    def _process(self):
        current_settings = self.settings_proxy.get_all(self.event)
        form = self.form_cls(obj=FormDefaults(**current_settings))
        if form.validate_on_submit():
            self.settings_proxy.set_multi(self.event, form.data)
            new_settings = self.settings_proxy.get_all(self.event)
            flash(self.success_message, 'success')
            changes = {k: (v, new_settings[k]) for k, v in current_settings.items() if v != new_settings[k]}
            if changes:
                self.event.log(EventLogRealm.management, LogKind.change, self.log_module, self.log_message, session.user,
                               data={'Changes': make_diff_log(changes, self.log_fields)})
            return jsonify_data()
        return jsonify_form(form)
```

Note: This code is untested.

indico/modules/attachments/controllers/management/event.py
```diff
@@ -97,3 +104,20 @@ class RHPackageEventAttachmentsManagement(AttachmentPackageMixin, RHManageEventB
...
+        current_settings = attachments_settings.get_all(self.event)
+        form = SetUploadPermissionsForm(obj=FormDefaults(**current_settings))
+        if form.validate_on_submit():
+            attachments_settings.set_multi(self.event, form.data)
+            new_settings = attachments_settings.get_all(self.event)
+            flash(_('Upload permissions settings changed successfully'), 'success')
+            changes = {k: (v, new_settings[k]) for k, v in current_settings.items() if v != new_settings[k]}
+            if changes:
+                log_fields = {'managers_only': 'Only managers can upload attachments'}
+                self.event.log(EventLogRealm.management, LogKind.change, 'Attachments', 'Settings updated',
```
ThiefMaster, line 120: "Upload settings updated"? That way you can use "Managers only" in the log fields above to keep that one more concise.

indico/modules/attachments/util.py
```diff
@@ -61,16 +61,19 @@ def get_attached_items(linked_object, include_empty=True, include_hidden=True, p
...
 def can_manage_attachments(obj, user, allow_admin=True):
     """Check if a user can manage attachments for the object."""
+    from indico.modules.attachments import attachments_settings
     if not user:
         return False
     if obj.can_manage(user, allow_admin=allow_admin):
         return True
     if isinstance(obj, db.m.Event) and obj.can_manage(user, 'submit', allow_admin=allow_admin):
-        return True
+        return not attachments_settings.get(obj, 'managers_only')
```
ThiefMaster, line 70: hmmm why not just an `elif not ...: return False` after the can_manage check?

duartegalvao: I just did it this way to make use of the `isinstance` (since the settings proxy needs the event, and before that point we don't know the type of `obj`).
Otherwise we would have to do something like this to get the event:
```python
event = (obj
         if isinstance(obj, db.m.Event)
         else obj.event
         if isinstance(obj, db.m.Contribution)
         else obj.contribution.event
         if isinstance(obj, db.m.SubContribution)
         else None)
```

ThiefMaster: All those objects  including events have `.event`, but I guess categories are the one exception... 

This should work though:

```python
event = getattr(obj, 'event', None)
if event and not attachments_settings.get(event, 'managers_only'):
    return False
```

ThiefMaster:
Related issue: The `attachments.upload_ckeditor` endpoint probably gets disabled as well, but it's used by the minute editor. So we need to disable the image upload from the editor in that case. I think omitting the `data-image-upload-url` in that case will do the job (unless you prefer to allow uploading images via that endpoint even if attachment upload is disabled).

ThiefMaster:
I think you missed at least one `render_manage_button` call - for meetings it's needed to write event-level minutes, and when I tested it with a user who's just a submitter for the event, that user did not see the button and thus had no way of writing minutes for the event.

duartegalvao:
> I think you missed at least one `render_manage_button` call - for meetings it's needed to write event-level minutes, and when I tested it with a user who's just a submitter for the event, that user did not see the button and thus had no way of writing minutes for the event.

Thanks! Hadn't realized what that was for. It's fixed now

ThiefMaster (approved):

PR #5598, opened 2022-12-07
Allow emailing people about specific contribs/abstracts

ThiefMaster:
Also refactored the existing dialog to email people a bit so it can be reused more easily.


indico/modules/events/abstracts/templates/management/abstract_list.html
```diff
@@ -170,6 +170,13 @@
                             data-ajax-dialog>
                         {%- trans %}Author list{% endtrans -%}
                     </button>
+                    <div id="email-abstract-roles-container"
```
duartegalvao, line 173: The Email button isn't aligning/spacing properly with the other buttons - I think this is because the `<button>` gets rendered inside this `<div>`. It should either be at the same level with the others, or in a `<div class="group">` with the author list.

image

A quick workaround I found is to add the class `group` here. One-button groups seem to work fine.

indico/modules/events/contributions/templates/management/contributions.html
```diff
@@ -101,6 +101,13 @@
                         data-ajax-dialog>
                     {%- trans %}Author list{% endtrans -%}
                 </button>
+                <div id="email-contrib-roles-container"
```
duartegalvao, line 104: Same here

duartegalvao (changes requested):

duartegalvao (approved):

PR #5602, opened 2022-12-12
🎄 add Xmas theme 🎄

tomasr8:
I guess we can revert this change after Christmas :thinking:


ThiefMaster:
couldn't we do it via the CERN customisation?

PR #5608, opened 2023-01-06
Add option to keep track when accepting abstracts

duartegalvao:
This PR adds an option to bulk-accept abstracts in their respective "reviewed for" tracks, instead of always having to manually select the accepted track:
image

### DIFF
--- a/indico/modules/events/abstracts/operations.py
+++ b/indico/modules/events/abstracts/operations.py
@@ -200,15 +200,16 @@ def delete_abstract(abstract, delete_contrib=False):
 
 def judge_abstract(abstract, abstract_data, judgment, judge, contrib_session=None, merge_persons=False,
                    send_notifications=False):
+    if abstract_data.get('use_review_track') and len(abstract.reviewed_for_tracks) > 1:
+        return False
     abstract.judge = judge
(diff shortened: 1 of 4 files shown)


indico/modules/events/abstracts/operations.py
```diff
@@ -206,7 +206,11 @@ def judge_abstract(abstract, abstract_data, judgment, judge, contrib_session=Non
     log_data = {'Judgment': orig_string(judgment.title)}
     if judgment == AbstractAction.accept:
         abstract.state = AbstractState.accepted
-        abstract.accepted_track = abstract_data.get('accepted_track')
+        if abstract_data.get('use_review_track'):
+            tracks = abstract.reviewed_for_tracks
+            abstract.accepted_track = list(tracks)[0] if len(tracks) == 1 else None
```
duartegalvao, line 211: I'm unsure of what the behavior should be when there are multiple review tracks... Maybe we should skip the abstract altogether?

ThiefMaster: Yes, I think we should skip those and flash a warning that some abstracts have been skipped because of this.

ThiefMaster (approved):

Issue #5594, opened 2022-12-05
Export Registration Stat via HTTP API


amandahla:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
Export registration stats via http-api.
https://docs.getindico.io/en/stable/http-api/exporters/event/

**Describe alternatives you've considered**
I'm not familiar with Flask but I was thinking that would be possible to add a parameter to  "/export/event/ID.TYPE" to bring the same information generated via UI.

**Additional context**
This would be useful to create a Prometheus Exporter and monitor the event (how many registrations so far).

PR #5610, opened 2023-01-06
Add registrations number to Event exporter (#5594)

amandahla:
Add a registration field with the number of registrations for a given event to the Event exporter (HTTP-API).

### DIFF
--- a/docs/source/http-api/exporters/event.rst
+++ b/docs/source/http-api/exporters/event.rst
@@ -43,6 +43,7 @@ Result for https://indico.server/export/event/137346.json?occ=yes&pretty=yes::
                     "time": "08:00:00"
                 },
                 "_type": "Conference",
+                "registrations":1,
                 "endDate": {
                     "date": "2011-06-24",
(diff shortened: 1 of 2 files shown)


ThiefMaster:
I'm not completely opposed to adding this, but it'd certainly need to be implemented more thoroughly and take into account things like whether the registration count of a particular registration form is being published (remember: anyone who can see the event can get its data from the API as well), and it should use the same logic as the displayed registration count does. 

The `active_registration_count` and `publish_registration_count` properties of a RegistrationForm might be useful for this.

amandahla:
Thanks for bringing this up, I wasn't aware of the privacy setting. 
I did this change but now the registration number is wrong and I'm not sure why. I'll investigate more.

ThiefMaster:
I don't think this should take the privacy settings into account (nor whether the user exporting the event is registered). They are meant for revealing lists w/ names, but not for the counts.

ThiefMaster:
Closing this one considering that the PR is currently stale and it's not clear which number to expose

PR #5612, opened 2023-01-09
Allow uploading files when requesting changes

duartegalvao:
This PR adds the ability for an editor to upload a revision when requesting changes to a submitter in the Editing Timeline.

image

image

### DIFF
--- a/indico/modules/events/editing/operations.py
+++ b/indico/modules/events/editing/operations.py
@@ -220,7 +220,7 @@ def _is_request_changes_with_files(revision):
     if not revision or len(revision.editable.revisions) < 2:
         return False
     previous_revision = revision.editable.revisions[-2]
-    # when request-changes-with-files revision is made, the previous revision's final_state
+    # when a request-changes-with-files revision is made, the previous revision's final_state
     # is set to needs_submitter_changes, and a new one is created with the same reviewed_dt,
     # editor, and final_state. if any of these are different, then the two revisions are
     # actually separate request-changes, and not a request-changes-with-files
--- a/indico/modules/events/editing/operations.py
+++ b/indico/modules/events/editing/operations.py
@@ -27,7 +27,7 @@ from indico.modules.events.editing.notifications import (notify_comment, notify_
 from indico.modules.events.editing.schemas import (EditableDumpSchema, EditingConfirmationAction, EditingFileTypeSchema,
                                                    EditingReviewAction)
 from indico.modules.logs import EventLogRealm, LogKind
-from indico.util.caching import memoize
+from indico.util.caching import memoize_request
 from indico.util.date_time import now_utc
 from indico.util.fs import secure_filename
 from indico.util.i18n import _, orig_string
@@ -215,7 +215,7 @@ def create_submitter_revision(prev_revision, user, files):
     return new_revision
 
 
-@memoize
+@memoize_request
 def _is_request_changes_with_files(revision):
     if not revision or len(revision.editable.revisions) < 2:
         return False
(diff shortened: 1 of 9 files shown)


indico/modules/events/editing/operations.py
```diff
@@ -124,19 +125,25 @@ def review_editable_revision(revision, editor, action, comment, tags, files=None
...
-    elif action in (EditingReviewAction.update, EditingReviewAction.update_accept):
+    elif files and action in (EditingReviewAction.update, EditingReviewAction.update_accept,
+                              EditingReviewAction.request_update):
+        initial_state = InitialRevisionState.needs_submitter_confirmation
         final_state = FinalRevisionState.none
         editable_editor = None
         if action == EditingReviewAction.update_accept:
             final_state = FinalRevisionState.accepted
             editable_editor = editor
+        elif action == EditingReviewAction.request_update:
```
tomasr8, line 136: This might be a bit easier to read?

```python
if action == EditingReviewAction.update:
    initial_state = InitialRevisionState.needs_submitter_confirmation
    final_state = FinalRevisionState.none
    editable_editor = None
elif action == EditingReviewAction.update_accept:
    initial_state = InitialRevisionState.needs_submitter_confirmation
    final_state = FinalRevisionState.accepted
    editable_editor = editor
elif action == EditingReviewAction.request_update:
    initial_state = revision.initial_state
    final_state = FinalRevisionState.needs_submitter_changes
    editable_editor = editor
```

indico/modules/events/editing/operations.py
```diff
@@ -204,8 +211,25 @@ def create_submitter_revision(prev_revision, user, files):
     return new_revision
 
 
+@memoize
```
tomasr8, line 214: Maybe we can add some extra flag to `EditingRevision` to simplify this check? Maybe whether the files were uploaded by the user or the editor?

duartegalvao: I really wanted to avoid migrations in this PR, specially since this can be achieved without storing extra data...
Maybe it would be cleaner to have this function as a property of `EditingRevision`? (although it's not used anywhere else other than this file)

ThiefMaster: tbh i'd keep it here for now...

indico/modules/events/editing/operations.py
```diff
@@ -204,8 +214,25 @@ def create_submitter_revision(prev_revision, user, files):
...
```
ThiefMaster, line 217: I don't think memoizing this is a good idea. The data is stored globally so it's really meant for things that should be memoized forever, not just in the context of a request. We have `@memoize_request` for that.

indico/modules/events/editing/operations.py
```diff
@@ -204,8 +214,25 @@ def create_submitter_revision(prev_revision, user, files):
...
+def _is_request_changes_with_files(revision):
+    if not revision or len(revision.editable.revisions) < 2:
+        return False
+    previous_revision = revision.editable.revisions[-2]
+    return (revision != previous_revision and
```
ThiefMaster, line 222: I think the logic here deserves a comment about what in here makes it as "request changes with files" and not a normal "request changes".

ThiefMaster (approved):

indico/modules/events/editing/operations.py
```diff
@@ -220,6 +220,10 @@ def _is_request_changes_with_files(revision):
     if not revision or len(revision.editable.revisions) < 2:
         return False
     previous_revision = revision.editable.revisions[-2]
+    # when request-changes-with-files revision is made, the previous revision's final_state
```
ThiefMaster, line 223:
```suggestion
    # when a request-changes-with-files revision is made, the previous revision's final_state
```

PR #5620, opened 2023-01-11
Add option to export event series to ical

### LINKED ISSUE #5617: Could iCalendar files for all events in a series be downloaded simultaneously?
Would it be possible to add a way to download the iCalendar files for all (future) events in a series simultaneously?

The new series feature is great for regular meetings inside shared categories, but when I create a series I'd like there to be a way for invitees / participants / me to add all the events in the series to the calendar at once, without subscribing to the whole category.

Thanks for considering this!

tomasr8:
Closes #5617 

Only available for events which are a part of a series.

image

### DIFF
--- a/indico/modules/events/controllers/display.py
+++ b/indico/modules/events/controllers/display.py
@@ -28,12 +28,12 @@ class RHExportEventICAL(RHDisplayEventBase):
     @use_kwargs({
         'scope': EnumField(CalendarScope, load_default=None),
         'detail': fields.String(load_default=None),
-        'export_event_series': fields.Boolean(load_default=False)
+        'series': fields.Boolean(load_default=False)  # Export the full event series
     }, location='query')
-    def _process(self, scope, detail, export_event_series):
+    def _process(self, scope, detail, series):
         if not scope and detail == 'contributions':
             scope = CalendarScope.contribution
-        if not export_event_series:
+        if not series:
             event_ical = event_to_ical(self.event, session.user, scope)
             return send_file('event.ics', BytesIO(event_ical), 'text/calendar')
         else:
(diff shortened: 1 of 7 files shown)


indico/web/client/js/react/components/ICSCalendarLink.jsx
```diff
@@ -119,15 +127,23 @@ export default function ICSCalendarLink({endpoint, params, renderButton, popupPo
...
-        popupState.source.cancel();
-      }
-      const source = indicoAxios.CancelToken.source();
-      dispatch({type: 'OPEN', key, source});
-      const url = await fetchURL(extraParams, source);
-      dispatch({type: 'LOADED', url});
+      extraParams = {
+        ...extraParams,
+        ...(exportEventSeries && {export_event_series: exportEventSeries}),
+      };
```
ThiefMaster, line 145: Wouldn't something like this be more readable?

```js
if (exportEventSeries) {
  extraParams = {...extraParams, export_event_series: true};
}
```

(I guess you want the assignment to avoid mutating the original `extraParms`?)

ThiefMaster (approved):

indico/web/client/js/react/components/ICSCalendarLink.jsx
tomasr8: Yup, much better with the `if ..`

indico/modules/events/controllers/display.py
```diff
@@ -27,13 +27,18 @@
 class RHExportEventICAL(RHDisplayEventBase):
     @use_kwargs({
         'scope': EnumField(CalendarScope, load_default=None),
-        'detail': fields.String(load_default=None)
+        'detail': fields.String(load_default=None),
+        'export_event_series': fields.Boolean(load_default=False)
```
ThiefMaster, line 31: Hmm what about calling the URL argument just `series` or `full_series` or `include_series`?

ThiefMaster (approved):

PR #5622, opened 2023-01-12
Expose room location in room booking's UI

### LINKED ISSUE #4291: Room Booking: Expose Location in the UI
When rewriting room booking, locations pretty much became a second-class citizen in the user-facing UI (AFAIK we don't show it *anywhere*). This is fine in case there is only one location anyway, but if an instance has more than one, then the location should be visible, and having e.g. a filter to see only rooms in a given location would be useful as well.

mindouro:
Closes #4291

This PR adds the possibility to filter by location in the Room Booking, when there are more than one location.
This also adds the location value in the room's card and inside the room's details modal.

SCR-20230112-ehz
SCR-20230112-eig
SCR-20230112-ejl

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,8 @@ Improvements
 - Support Markdown when writing global announcement and apply standard HTML
   sanitization to the message (:pr:`5640`)
 - Add BCC field on contribution email dialogs (:pr:`5637`)
+- Allow filtering by location in room booking (:issue:`4291`, :pr:`5622`,
+  thanks :user:`mindouro`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/rb/client/js/common/rooms/selectors.js
+++ b/indico/modules/rb/client/js/common/rooms/selectors.js
@@ -137,3 +137,8 @@ export const getBuildings = createSelector(
     return Array.from(buildings).sort((a, b) => parseInt(a, 10) - parseInt(b, 10));
   }
 );
+
+export const getLocations = createSelector(
+  getAllRooms,
+  allRooms => new Map(Object.values(allRooms).map(room => [room.locationId, room.locationName]))
+);
--- a/indico/modules/rb/client/js/components/Room.jsx
+++ b/indico/modules/rb/client/js/components/Room.jsx
@@ -16,7 +16,6 @@ import {TooltipIfTruncated, ResponsivePopup} from 'indico/react/components';
 import {Translate} from 'indico/react/i18n';
 import {Markdown, Slot} from 'indico/react/util';
 
-import {selectors as roomsSelectors} from '../common/rooms';
 import {actions as userActions, selectors as userSelectors} from '../common/user';
 
 import DimmableImage from './DimmableImage';
@@ -34,7 +33,6 @@ class Room extends React.Component {
     isCheckingUserRoomPermissions: PropTypes.bool.isRequired,
     addFavoriteRoom: PropTypes.func.isRequired,
     delFavoriteRoom: PropTypes.func.isRequired,
-    allLocations: PropTypes.object.isRequired,
   };
 
   static defaultProps = {
@@ -151,11 +149,9 @@ class Room extends React.Component {
       addFavoriteRoom,
       delFavoriteRoom,
       isCheckingUserRoomPermissions,
-      allLocations,
       ...restProps
     } = this.props;
     const {content, actions} = Slot.split(children);
-    const showLocation = allLocations.size > 1;
 
     return (
       <Card styleName="room-card" {...restProps}>
@@ -180,11 +176,6 @@ class Room extends React.Component {
         </Card.Content>
         <Card.Content styleName="room-content" extra>
           <Icon name="user" /> {room.capacity || Translate.string('Not specified')}
-          {showLocation && (
-            <div>
-              <Icon name="map pin" /> {room.locationName || Translate.string('Not specified')}
-            </div>
-          )}
           <span styleName="room-details">
             {room.features.map(feature => (
               <RoomFeatureEntry key={feature.name} feature={feature} color="green" />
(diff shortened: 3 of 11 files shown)


ThiefMaster:
Nice! I wonder if the Location filter should be shown only if there's more than one location, since otherwise it's pretty pointless to filter by it.

mindouro:
> Nice! I wonder if the Location filter should be shown only if there's more than one location, since otherwise it's pretty pointless to filter by it.

I'm checking how many locations we have before showing the location's filter to avoid that. I also won't show the location in the room's card or details if there is only one location. 🙂

CHANGES.rst
```diff
@@ -12,6 +12,7 @@ Improvements
 
 - Include co-authors in abstract list columns and spreadsheet exports (:pr:`5605`)
 - Include speakers in abstract list columns and spreadsheet exports (:pr:`5615`)
+- Expose room location in room booking's UI (:pr:`5622`)
```
ThiefMaster, line 15: you can credit yourself here since you aren't a team member (see the `thanks :user:...` snippets used for some other external contributions)

indico/modules/rb/client/js/common/rooms/selectors.js
```diff
@@ -137,3 +137,19 @@ export const getBuildings = createSelector(
...
+export const getLocations = createSelector(
+  getAllRooms,
+  allRooms => {
+    const locations = Object.values(allRooms)
+      .map(({locationName, locationId}) => {
+        return {
+          name: locationName,
+          id: locationId,
+        };
+      })
```
ThiefMaster, line 150: Since the callback just returns something you can use the expression-style arrow function:
```suggestion
      .map(({locationName, locationId}) => ({
        name: locationName,
        id: locationId,
      }))
```

indico/modules/rb/client/js/common/rooms/selectors.js
```diff
@@ -137,3 +137,19 @@ export const getBuildings = createSelector(
...
+      .filter((value, index, self) => index === self.findIndex(t => t.id === value.id));
```
ThiefMaster, line 151: That looks like an incredibly ineffective way of removing duplicates. It'd be better to replace this whole selector function with one that simply builds a map using `new Map(Object.values(allRooms).map(room => [room.locationId, room.locationName]))` (untested, but should work) - that can then of course be converted to an array of objects if you prefer that over an ES Map object

indico/modules/rb/client/js/modules/roomList/filters/LocationForm.jsx
```diff
@@ -0,0 +1,56 @@
...
+    locations: PropTypes.array.isRequired,
+    location: PropTypes.number,
+    ...FilterFormComponent.propTypes,
+  };
+
+  static defaultProps = {
+    location: null,
+  };
+
+  setlocation = locationId => {
```
ThiefMaster, line 27: `setLocation`

CHANGES.rst
```diff
@@ -13,6 +13,7 @@ Improvements
 - Include co-authors in abstract list columns and spreadsheet exports (:pr:`5605`)
 - Include speakers in abstract list columns and spreadsheet exports (:pr:`5615`)
 - Add an option to export all events in a series to ical at once (:issue:`5617`, :pr:`5620`)
+- Expose room location in room booking's UI (:pr:`5622`, thanks :user:`mindouro`)
```
ThiefMaster, line 16:
```suggestion
- Allow filtering by location in room booking (:issue:`4291`, :pr:`5622`, thanks :user:`mindouro`)
```

indico/modules/rb/client/js/components/Room.jsx
```diff
@@ -175,9 +179,12 @@ class Room extends React.Component {
...
         <Card.Content styleName="room-content" extra>
-          <>
-            <Icon name="user" /> {room.capacity || Translate.string('Not specified')}
-          </>
+          <Icon name="user" /> {room.capacity || Translate.string('Not specified')}
+          {showLocation && (
+            <div>
+              <Icon name="map pin" /> {room.locationName || Translate.string('Not specified')}
+            </div>
+          )}
```
ThiefMaster, line 187: The room card looks rather broken with the location information added:

image

indico/modules/rb/client/js/modules/roomList/RoomFilterBar.jsx
```diff
@@ -158,11 +177,32 @@ export class RoomFilterBarBase extends React.Component {
         disabled={disabled}
       />
     );
+    // const selectedLocationName = locations.find(location => location.id === locationId)?.name;
```
ThiefMaster, line 180:
```suggestion
```

let's remove that :)

indico/modules/rb/client/js/modules/roomList/RoomFilterBar.jsx
```diff
@@ -158,11 +177,32 @@ export class RoomFilterBarBase extends React.Component {
...
+              form={({locationId: selectedLocation}, setParentField) => (
+                <LocationForm
+                  setParentField={setParentField}
+                  locations={locations}
+                  location={selectedLocation}
+                />
+              )}
+              setGlobalState={data => {
+                setFilterParameter('locationId', data.locationId);
+              }}
```
ThiefMaster, line 200:
```suggestion
              setGlobalState={data => setFilterParameter('locationId', data.locationId)}
```

ThiefMaster (changes requested):
Very nice PR! Just a few small fixes and we can merge this. :)

indico/modules/rb/client/js/components/Room.jsx
mindouro: This happens because the class `.room-card` in `Room.module.scss` has a fixed height value of 320px. My solution was removing the height of it and let Semantic UI's `Card.Group` component manage the card's heigh, keeping the same height for all cards of that row and avoiding future problems when adding any other new fields to the room card. Let me know if you prefer other approach 🙂

ThiefMaster (approved):

indico/modules/rb/client/js/components/Room.jsx
ThiefMaster: (I added a commit to not show the location there)

PR #5623, opened 2023-01-17
Replace HTML paragraphs and linebreaks with newlines in plaintext conversion

### LINKED ISSUE #5611: Line breaks in description are ignored in reminders
**Describe the bug**
Indico v3.2.1 and newer ignores line breaks and paragraph changes in the description block of reminder emails.  This problem did not occur with version 3.2 and earlier.

**To Reproduce**
1. Create an event with a multi-line/multi-paragraph description, e.g.:
```
<p>
    Title: Lecture I
</p>
<p>
    Speaker: Speaker I (Institute I)
</p>
<p>
    Abstract: First things first.
</p>
```
if viewed as a html source.
2. Add a reminder
3. Tick the __Include description__ option
4. __Preview email__
5. The description block displays the above three paragraphs in a single line:
```
Description
-----------

Title: Lecture ISpeaker: Speaker I (Institute I)Abstract: First things first.


Note
----
======== zoom connection =============================================

Link, meeting ID and passcode are available on the indico page.

======================================================================

--
Indico :: Email Notifier
```

**Expected behavior**
Indico v3.2 would format the description block like:
```
Description
-----------

Title: Lecture I

Speaker: Speaker I (Institute I)

Abstract: First things first.
```

**Additional context**
The note block is still intact.

tomasr8:
Closes #5611

### DIFF
--- a/indico/util/string.py
+++ b/indico/util/string.py
@@ -506,7 +506,7 @@ def html_to_plaintext(string):
     doc = etree.HTML(string)
     for elem in doc.xpath('//p | //br'):
         elem.tail = '\n' + elem.tail if elem.tail else '\n'
-    return doc.xpath('string()')
+    return doc.xpath('string()').strip()
 
 
 class RichMarkup(Markup):
--- a/indico/util/string.py
+++ b/indico/util/string.py
@@ -495,7 +495,18 @@ def sanitize_html(string):
 
 
 def html_to_plaintext(string):
-    return html.html5parser.fromstring(string).xpath('string()')
+    """Convert HTML to plaintext.
+
+    :param string: The HTML source string
+
+    <p> and <br> elements are converted into newline characters.
+    Any literal '\n' characters in the HTML source are removed.
+    """
+    string = string.replace('\n', '')
+    doc = etree.HTML(string)
+    for elem in doc.xpath('//p | //br'):
+        elem.tail = '\n' + elem.tail if elem.tail else '\n'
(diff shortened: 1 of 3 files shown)


indico/util/string.py
```diff
@@ -495,7 +495,18 @@ def sanitize_html(string):
...
+
+    :param string: The HTML source string
+
+    <p> and <br> elements are converted into newline characters.
+    Any literal '\n' characters in the HTML source are removed.
+    """
+    string = string.replace('\n', '')
+    doc = etree.HTML(string)
+    for elem in doc.xpath('//p | //br'):
+        elem.tail = '\n' + elem.tail if elem.tail else '\n'
```
ThiefMaster, line 508: is `elem.tail` None in some cases? Otherwise `elem.tail = f'{elem.tail}\n/` should do the job

indico/util/string.py
```diff
@@ -495,7 +495,18 @@ def sanitize_html(string):
...
+    return doc.xpath('string()')
```
ThiefMaster, line 509: Should we `.strip()` it?

indico/util/string.py
tomasr8: AFAIU if there is nothing between the elements it'll be None

indico/util/string.py
ThiefMaster: My thought is that leading/trailing whitespace will almost always be noisy, especially in the places where we use this conversion..

PR #5630, opened 2023-01-19
Keep state of undone timeline revisions

duartegalvao:
Depends on #5612

This PR turns the "Undo revision" feature of the Editing Timeline into a "soft" undo, keeping all files and comments available in the timeline:
image

Progress:
- [x] Change revision's state to Undone instead of deleting it
- [x] Keep all revision comments and display them as undone
- [x] Test all undo scenarios
- [x] Test microsservice integration


ThiefMaster:
While testing this I ended up with some cases with broken states, some not recoverable from the UI...

- "you cannot do this on this revision" error when doing "make changes and accept"
- getting stuck in a state where neither the editor nor the submitter could do anything after undo

ThiefMaster (approved):

PR #5631, opened 2023-01-22
Extend user fixtures

olifre:
This adds a `create_identity` fixture and extends the `create_group` fixture to allow specifying the group name. If no group name is passed, the group name is autogenerated as before (i.e. existing behaviour is not changed). 

I found these two fixtures helpful in writing [tests for a plugin](https://github.com/unibonn/indico-sso-group-mapping/blob/bfd2e5239d032e7e86601ec02f500812c9ad8ff0/tests/plugin_test.py) dealing with users, groups and identities and they should be generic enough to rather be part of Indico than of the plugin code :smile:.

### DIFF
--- a/indico/testing/fixtures/user.py
+++ b/indico/testing/fixtures/user.py
@@ -45,10 +45,10 @@ def dummy_user(create_user):
 @pytest.fixture
 def create_group(db):
     """Return a callable which lets you create dummy groups."""
-    def _create_group(id_):
+    def _create_group(id_, group_name=None):
         group = LocalGroup()
         group.id = id_
-        group.name = f'dummy-{id_}'
+        group.name = group_name or f'dummy-{id_}'
         db.session.add(group)
         db.session.flush()
         return group.proxy
--- a/indico/testing/fixtures/user.py
+++ b/indico/testing/fixtures/user.py
@@ -7,6 +7,7 @@
 
 import pytest
 
+from indico.modules.auth import Identity
 from indico.modules.groups.models.groups import LocalGroup
 from indico.modules.rb import rb_settings
 from indico.modules.users import User
@@ -60,3 +61,14 @@ def create_group(db):
 def dummy_group(create_group):
     """Create a mocked dummy group."""
     return create_group(1337)
+
+
+@pytest.fixture
+def create_identity(db):
+    """Return a callable which lets you create dummy identities."""
+    def _create_identity(user, provider, identifier):
+        identity = Identity(user=user, provider=provider, identifier=identifier)
+        db.session.flush()
+        return identity
+
+    return _create_identity


indico/testing/fixtures/user.py
```diff
@@ -45,10 +45,10 @@ def dummy_user(create_user):
 @pytest.fixture
 def create_group(db):
     """Return a callable which lets you create dummy groups."""
-    def _create_group(id_):
+    def _create_group(id_, group_name=None):
         group = LocalGroup()
         group.id = id_
-        group.name = f'dummy-{id_}'
+        group.name = f'dummy-{id_}' if group_name is None else group_name
```
ThiefMaster, line 51:
```suggestion
        group.name = group_name or f'dummy-{id_}'
```

explicit None checks don't matter here so this can be simplified :)

indico/testing/fixtures/user.py
```diff
@@ -60,3 +60,15 @@ def _create_group(id_):
 def dummy_group(create_group):
     """Create a mocked dummy group."""
     return create_group(1337)
+
+
+@pytest.fixture
+def create_identity(db):
+    """Return a callable which lets you create dummy identities."""
+    def _create_identity(user, provider, identifier):
+        identity = Identity(user_id=user.id, provider=provider, identifier=identifier)
```
ThiefMaster, line 69:
```suggestion
        identity = Identity(user=user, provider=provider, identifier=identifier)
```

It's always better to use the relationship attributes instead of the foreign keys. That way you most likely don't even need the `db.session.add()` call below since the user is already in the DB session and this automatically chains across relationships.

ThiefMaster (changes requested):
Just some minor mostly stylistic things to improve..

PS: Please note that git commit messages usually do not end with a period. I'll squash-and-merge the PR once you updated the two things I mentioned, so up to you whether to reword them or not.

indico/testing/fixtures/user.py
olifre: Indeed, thanks a lot — shorter code is always nicer :+1:.

indico/testing/fixtures/user.py
olifre: That's very true. The updated commits implement the suggested change. However, I tested that indeed `db.session.add()` is not needed, as you expected. 
Should I also drop this?

olifre:
> Just some minor mostly stylistic things to improve..

Thanks, these things are amongst the most useful pointers to improve upon my coding. I implemented everything, and also added the forgotten `import`, so tests should now run through correctly. 
Should I also drop the `db.session.add()`? I checked and indeed it seems it's not needed anymore.

indico/testing/fixtures/user.py
ThiefMaster: yeah I'd drop it

olifre: Done and force-pushed :smile:.

indico/testing/fixtures/user.py
```diff
@@ -60,3 +61,13 @@ def _create_group(id_):
 def dummy_group(create_group):
     """Create a mocked dummy group."""
     return create_group(1337)
+
+
+@pytest.fixture
+def create_identity(db):
+    """Return a callable which lets you create dummy identities."""
+    def _create_identity(user, provider, identifier):
+        identity = Identity(user=user, provider=provider, identifier=identifier)
```
ThiefMaster, line 70: I would keep the `db.session.flush()` here, in tests it's useful that those things are immediately sent to the DB and not only implicitly later when they are needed..

olifre: That makes sense, I added it back :+1: .

ThiefMaster (approved):

PR #5637, opened 2023-01-25
Add CC field on contribution email dialogs

duartegalvao:
This PR adds a CC field to the email dialog for contributions/author lists:

image

Note: the emails are sent as BCC, like the "send copy to myself" feature

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,7 @@ Improvements
   contributions (:pr:`5597`)
 - Support Markdown when writing global announcement and apply standard HTML
   sanitization to the message (:pr:`5640`)
+- Add BCC field on contribution email dialogs (:pr:`5637`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/events/management/controllers/emails.py
+++ b/indico/modules/events/management/controllers/emails.py
@@ -12,7 +12,7 @@ from webargs.flaskparser import abort
 from indico.core.notifications import make_email, send_email
 from indico.modules.events.contributions.models.persons import AuthorType
 from indico.util.i18n import _
-from indico.util.marshmallow import not_empty
+from indico.util.marshmallow import LowercaseString, not_empty
 from indico.util.placeholders import get_sorted_placeholders, replace_placeholders
 from indico.web.args import use_kwargs
 from indico.web.flask.templating import get_template_module
@@ -93,6 +93,7 @@ class EmailRolesSendMixin:
         'from_address': fields.String(required=True, validate=not_empty),
         'body': fields.String(required=True, validate=not_empty),
         'subject': fields.String(required=True, validate=not_empty),
+        'bcc_addresses': fields.List(LowercaseString(validate=validate.Email())),
         'copy_for_sender': fields.Bool(load_default=False),
         'recipient_roles': fields.List(
             fields.String(validate=validate.OneOf(['speaker', 'author', 'coauthor', 'submitter'])),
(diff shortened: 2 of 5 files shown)


indico/modules/events/management/controllers/emails.py
```diff
@@ -110,7 +111,7 @@ def _process(self, from_address, body, subject, copy_for_sender, recipient_roles
                                               object_context=self.object_context, **kwargs)
             email_subject = replace_placeholders('event-persons-email', subject, event=self.event,
                                                  object_context=self.object_context, **kwargs)
-            bcc = [session.user.email] if copy_for_sender else []
+            bcc = cc_addresses + [session.user.email] if copy_for_sender else cc_addresses
```
ThiefMaster, line 114:
```python
bcc = {session.user.email} if copy_for_sender else set()
bcc.update(cc_addresses)
```

I think that's clearer and also avoids duplicates

indico/modules/events/management/controllers/emails.py
```diff
@@ -93,14 +93,15 @@ def get_roles_from_person_link(self, person_link):
         'from_address': fields.String(required=True, validate=not_empty),
         'body': fields.String(required=True, validate=not_empty),
         'subject': fields.String(required=True, validate=not_empty),
+        'cc_addresses': fields.List(fields.String(validate=validate.Email())),
```
ThiefMaster, line 96: I'd use the `LowercaseString` type here

indico/web/client/js/react/components/EmailListField.jsx
```diff
@@ -47,6 +47,7 @@ const EmailListField = props => {
       placeholder={Translate.string('Please enter an email address')}
       additionLabel={Translate.string('Add email') + ' '} // eslint-disable-line prefer-template
       onChange={handleChange}
+      icon=""
```
ThiefMaster, line 50: That is this doing? Seems a bit strange to set icon to an empty string..

indico/modules/events/persons/controllers.py
```diff
@@ -365,7 +366,7 @@ def _process(self, from_address, body, subject, copy_for_sender):
             email_subject = replace_placeholders('event-persons-email', subject, person=recipient,
                                                  event=self.event, register_link=self.no_account)
             tpl = get_template_module('emails/custom.html', subject=email_subject, body=email_body)
-            bcc = [session.user.email] if copy_for_sender else []
+            bcc = cc_addresses + [session.user.email] if copy_for_sender else cc_addresses
```
ThiefMaster, line 369: same here

indico/web/client/js/react/components/EmailListField.jsx
duartegalvao: It's to hide the dropdown icon on the `EmailListField`. I don't think it makes sense to show it since it's supposed to act more like a text input...

ThiefMaster: OK makes sense

CHANGES.rst
```diff
@@ -18,6 +18,7 @@ Improvements
   to avoid collisions and inconsistencies (:pr:`5478`)
 - Add option to use review track as accepted track when bulk-accepting abstracts
   (:pr:`5608`)
+- Add CC field on contribution email dialogs (:pr:`5637`)
```
ThiefMaster, line 21: I just noticed that we use BCC for those addresses, so I think we should use BCC consistently her and in the UI as well. (I assumed it was CC when we talked about this before)

indico/web/client/js/react/components/EmailListField.jsx
```diff
@@ -22,31 +22,57 @@ const isValid = value => /^\S+@\S+\.\S+$/.test(value);
...
     onBlur();
   };
 
+  const handleChange = (e, {value: newValue}) => {
+    setValue(newValue);
+    setSearchQuery('');
+  };
+
+  const handleSearchChange = (e, {searchQuery: newSearchQuery}) => {
+    if (newSearchQuery.includes(',')) {
```
ThiefMaster, line 41: I'd handle semicolons as well

ThiefMaster (approved):

PR #5638, opened 2023-01-25
Notify registration managers on users withdrawing

### LINKED ISSUE #5633: Notify registration managers when a participant withdraws their registration
Registration managers, or those who are in the notify list for specific reg forms, get notified on various state changes. However, it appears there is no notification to managers when a registration is withdrawn by a participant.

An email is sent to registration managers, and registration form notify emails, when a user has withdrawn their registration. 

Looking at the list once a day to see who withdrew? ;-)

**Additional context**
It looks like `notify_registration_state_update` is not called in `RHRegistrationWithdraw`, while it is in `RHRegistrationManageWithdraw`. Would it be as simple as adding the call here?

kewisch:
Closes #5633

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,8 @@ Bugfixes
 - Include token in notification emails for private surveys (:pr:`5618`)
 - Fix some API calls not working with personal access tokens (:pr:`5627`)
 - Correctly handle paragraphs and linebreaks in plaintext conversion (:pr:`5623`)
+- Send manager notifications and email participant if they withdraw from an event
+  (:issue:`5633`, :pr:`5638`, thanks :user:`kewisch`)
 
 
 Version 3.2.2
--- a/indico/modules/events/registration/controllers/display.py
+++ b/indico/modules/events/registration/controllers/display.py
@@ -24,6 +24,7 @@ from indico.modules.events.registration.models.forms import RegistrationForm
 from indico.modules.events.registration.models.invitations import InvitationState, RegistrationInvitation
 from indico.modules.events.registration.models.items import PersonalDataType
 from indico.modules.events.registration.models.registrations import Registration, RegistrationState
+from indico.modules.events.registration.notifications import notify_registration_state_update
 from indico.modules.events.registration.util import (check_registration_email, create_registration, generate_ticket,
                                                      get_event_regforms_registrations, get_flat_section_submission_data,
                                                      get_initial_form_values, get_user_data, make_registration_schema)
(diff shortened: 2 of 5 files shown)


indico/modules/events/registration/controllers/display.py
```diff
@@ -409,7 +410,8 @@ def _check_access(self):
     def _process(self):
         self.registration.update_state(withdrawn=True)
         flash(_('Your registration has been withdrawn.'), 'success')
-        return redirect(url_for('.display_regform', self.registration.locator.registrant))
+        notify_registration_state_update(self.registration)
+        return redirect(url_for('event_registration.display_regform', self.registration.locator.registrant))
```
kewisch, line 414: Changing the redirect url here was necessary for the test to run. I was not able to get the test request context to descend into the event_registration blueprint. If you have ideas on how this can be done without changing the url please let me know!

CHANGES.rst
```diff
@@ -18,6 +18,8 @@ Improvements
   to avoid collisions and inconsistencies (:pr:`5478`)
 - Add option to use review track as accepted track when bulk-accepting abstracts
   (:pr:`5608`)
+- Send manager notifications and email participant if they withdraw from an event
```
ThiefMaster, line 21: I'd move this to the bugfixes section

indico/modules/events/registration/controllers/display.py
ThiefMaster, line 414: The explicit endpoint shouldn't be needed since it's in the same blueprint

indico/modules/events/registration/controllers/display_test.py
```diff
@@ -43,3 +46,50 @@ def test_RHRegistrationForm_can_register(db, dummy_regform, dummy_reg, dummy_use
     assert rh._can_register()  # withdrawn/rejected do not count against limit
     dummy_reg.state = RegistrationState.complete
     assert not rh._can_register()  # exceeding limit
+
+
+@pytest.mark.usefixtures('request_context')
+def test_RHRegistrationWithdraw(smtp, dummy_regform, dummy_reg, dummy_user):
```
ThiefMaster, line 52: not a big fan of the uppercase stuff in the name :) `test_withdraw_registration_rh`?

PS: thanks for adding a test!

indico/modules/events/registration/controllers/display.py
ThiefMaster: ah damn, I don't think there's a good solution that's not doing any awful hackery, so let's leave it as is...

ThiefMaster (approved):
Thanks for the nice PR + test. It even helped spot a typo in another template...

kewisch:
Glad I could help out!

PR #5639, opened 2023-01-26
Add before-check-access signal

OmeGak:


indico/web/rh.py
```diff
@@ -233,8 +233,10 @@ def _do_process(self):
         if rv is not None:
             return rv
 
-        self._check_access()
-        signals.rh.check_access.send(type(self), rh=self)
+        signal_rv = values_from_signal(signals.rh.before_check_access.send(type(self), rh=self))
+        if len(signal_rv) == 0:
```
ThiefMaster, line 237: IIRC we discussed the return value form the signal before and came up with something different than what's here? i don't really like that *any* return value will skip the original access check...

OmeGak: I missed the message where you proposed something different:

> True -> skip access check; False -> raise Forbidden with a generic message (ice to have for the sake of completeness, most code should probably raise an error itself though); None -> do the usual access checks. if a mix of True and False is returned, consider it False. IIRC we do the same in the can_access override signal

indico/core/signals/rh.py
```diff
@@ -18,6 +18,14 @@
 *result*.
 ''')
 
+before_check_access = _signals.signal('before-check-access', '''
+Executed right before `_check_access` of an `RH` instance has been called.
```
ThiefMaster, line 22:
```suggestion
Executed right before `_check_access` of an `RH` instance is called.
```

indico/core/signals/rh.py
```diff
@@ -18,6 +18,14 @@
...
+The *sender* is the RH class, the current instance is passed in *rh*.
+If the signal returns ``True`` or ``False``, the original `_check_access` method
+will not be executed.  If multiple subscribers to the signal return
+contradictory results, ``False`` wins and access is denied.
+''')
```
ThiefMaster, line 27:
```suggestion
In case access is denied, a generic error message will be displayed.
For custom errors, raise `Forbidden` yourself while handling the signal.
''')
```

PR #5645, opened 2023-01-31
Make overriding InternalSearch parts easier

ThiefMaster:
For our UN friends, and it's cleaner anyway to not nest those functions inside other methods...

cc @openprojects, please test it again on your side before I merge this


openprojects (approved):
LGTM

PR #5647, opened 2023-02-02
Fix Authors List filters

duartegalvao:
This PR fixes a bug in which React crashes when trying to apply a filter in an Author List with selected items. It also simplifies the code by sending the author list to the front-end as a dictionary instead of a list.

### DIFF
--- a/indico/modules/events/persons/client/js/AuthorsList.jsx
+++ b/indico/modules/events/persons/client/js/AuthorsList.jsx
@@ -10,7 +10,7 @@ import contribsPersonListURL from 'indico-url:contributions.person_list';
 import sessionsPersonListURL from 'indico-url:sessions.person_list';
 
 import PropTypes from 'prop-types';
-import React, {useMemo, useState} from 'react';
+import React, {useState} from 'react';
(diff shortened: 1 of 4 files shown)


indico/modules/events/persons/client/js/AuthorsList.jsx
```diff
@@ -145,7 +144,7 @@ export function AuthorsList({eventId, sourceIds, objectContext, onClose}) {
       <Modal.Actions>
         <EmailParticipantRolesButton
           eventId={eventId}
-          persons={selectedPersons.filter(id => isVisible(eventPersons.get(id)))}
+          persons={selectedPersons.filter(id => isVisible(data.eventPersons[id]))}
```
duartegalvao, line 147: Should we keep using a Map of eventPersons here for performance? Or is an object fine?

ThiefMaster (approved):

indico/modules/events/persons/client/js/AuthorsList.jsx
ThiefMaster: I think we just used it because we needed something mapping-like. Even with a few thousand EventPersons I doubt there will be a noticeable difference (object key lookup should be around O(1) as well)

PR #5648, opened 2023-02-02
Add setting to prevent disabling profile data sync

duartegalvao:
This PR adds a setting to `indico.conf` to prevent non-admin users of turning off their profile's personal data synchronization.
Fields listed in the `locked_fields` setting can only be de-synced by an administrator.
image

Α custom `locked_field_message` may also be set to help guide users:
image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,10 +35,8 @@ Improvements
 - Hide the top box with the latest files of an editable until it has been accepted
   and published (:issue:`5660`, :pr:`5665`)
 - Allow uploading files when requesting changes on the editing timeline (:pr:`5612`)
-- Add a :data:`locked_fields` setting to ``indico.conf`` to prevent non-admin
-  users of turning off their profile's personal data synchronization (:pr:`5648`)
 - Add ``locked_fields`` to the identity provider settings in ``indico.conf`` to
-  prevent non-admin users of turning off their profile's personal data
+  prevent non-admin users from turning off their profile's personal data
   synchronization (:pr:`5648`)
 
 Bugfixes
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,9 @@ Improvements
 - Allow uploading files when requesting changes on the editing timeline (:pr:`5612`)
 - Add a :data:`locked_fields` setting to ``indico.conf`` to prevent non-admin
   users of turning off their profile's personal data synchronization (:pr:`5648`)
+- Add ``locked_fields`` to the identity provider settings in ``indico.conf`` to
+  prevent non-admin users of turning off their profile's personal data
+  synchronization (:pr:`5648`)
 
 Bugfixes
 ^^^^^^^^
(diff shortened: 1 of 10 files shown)


CHANGES.rst
```diff
@@ -26,6 +26,8 @@ Improvements
 - Allow filtering by location in room booking (:issue:`4291`, :pr:`5622`,
   thanks :user:`mindouro`)
 - Add button to adapt column widths in paper & contribution lists (:pr:`5642`)
+- Add a :data:`locked_fields` setting to ``indico.conf`` to prevent non-admin
```
ThiefMaster, line 29: I don't think the `:data:` will help here since it's not a top-level key in the rst file. But maybe mention "to identity provider settings" there

indico/modules/auth/controllers.py
```diff
@@ -584,6 +585,9 @@ def validate_everything(self, data, **kwargs):
                     raise ValidationError(_('This field cannot be empty.'), 'first_name')
                 if 'last_name' not in data['synced_fields'] and not data['last_name']:
                     raise ValidationError(_('This field cannot be empty.'), 'last_name')
+                for field in data['synced_fields']:
+                    if field not in multipass.locked_fields:
+                        raise ValidationError('Locked fields must be synced.', field)
```
ThiefMaster, line 590: Careful, this need to take the edge case into account where there is no data for a synced field. In that case it should NOT be considered locked (otherwise such a user could not sign up).

duartegalvao: I think adding a ` and self.identity_info['data'].get(field)` will fix that but is there any way of easily testing this?

ThiefMaster: yeah, hacking your local multipass-cern or multipass-jacow code to send empty data ;)

duartegalvao: also, maybe should we only allow desyncing mandatory fields (when empty)?

ThiefMaster: not sure what you mean - as an admin you should always be able to desync (and in that case IMHO the user should be able to freely change the field), regardless if it's mandatory/empty or not

duartegalvao: im asking if we should let normal users desync empty fields if they are not mandatory
since this code is only for the signup page I don't think we even need to take admins into account...

duartegalvao: i assumed only the mandatory fields should be desynced. just pushed some code that seems to work well

indico/core/auth.py
```diff
@@ -56,6 +56,18 @@ def synced_fields(self):
...
+    def locked_fields(self):
+        """The keys that cannot be desynchronized.
+
+        This is a subset of the keys listed in ``synced_fields`` that cannot be
+        desynchronized by a non-admin user.
+        """
+        provider = self.sync_provider
+        if provider is None:
+            return set()
+        return set(provider.settings.get('locked_fields')) & self.synced_fields
```
ThiefMaster, line 69: This fails if the setting isn't present

```suggestion
        return set(provider.settings.get('locked_fields', ())) & self.synced_fields
```

PR #5649, opened 2023-02-02
Allow setting default locale w/o enforcing it

ThiefMaster:
After discussing w/ @pferreir a few days ago we thought it could be nice to let people define an event language even without enforcing it, just for the sake of having better metadata.

The question is in which metadata we want to have this. The JSON-LD event metadata we have doesn't seem to have a way to specify the event language, so the only thing that comes to my mind is API responses and maybe the HTML `lang` metadata... but that should probably only be set if the language is actually being enforced (ie not if the user overrides it with their own language preference).


pferreir:
Wouldn't https://schema.org/inLanguage work?

PR #5652, opened 2023-02-04
Allow to use existing redis for tests

### LINKED ISSUE #5651: Allow to use existing redis for tests
Testing framework uses `redis_proc` fixture (of `pytest-redis` package) for booting up a redis instance using redis executable. 
I'm using redis in container so this fixture is not suitable for me. 

`postgresql` fixture is very similar. It calls pg executables to start pg instance except environment variable `INDICO_TEST_DATABASE_URI` is set. In this case it connects to the existing db. I would like to see a similar solution. 

My suggestion is some rework on the app fixture:
```
@pytest.fixture(scope='session')
def app(request):
    """Create the flask app."""

    redis_cache_url = os.environ.get('INDICO_TEST_REDIS_CACHE_URL')
    if not redis_cache_url:
        redis_proc = request.getfixturevalue('redis_proc')
        redis_cache_url = f'redis://{redis_proc.host}:{redis_proc.port}/0'

    config_override = {
        'BASE_URL': 'http://localhost',
        'SMTP_SERVER': ('localhost', 0),  # invalid port - just in case so we NEVER send emails!
        'TEMP_DIR': request.config.indico_temp_dir.strpath,
        'CACHE_DIR': request.config.indico_temp_dir.strpath,
        'REDIS_CACHE_URL': redis_cache_url,
        'STORAGE_BACKENDS': {'default': 'mem:'},
        'PLUGINS': request.config.indico_plugins,
        'ENABLE_ROOMBOOKING': True,
        'SECRET_KEY': os.urandom(16),
        'SMTP_USE_CELERY': False,
    }
    return make_app(testing=True, config_override=config_override)
```
In fact this is simpler than we have for postgres because there's no need for manipulating redis directly from tests. At least at this moment  so I didn't want to add a separate fixture for that. It can be improved later if needed. 
Anyway I'm happy to contribute with this one or something similar if you think it's needed.

Remark: `pytest-redis` has an additional fixture `redis-nooproc` which does nothing else only compiles this redis URL I would like to put into an environment variable.

tomako:
This PR is about to consider the environment variable `INDICO_TEST_REDIS_CACHE_URL` when running pytest. This way the test suite can use an existing redis instance instead of creating a new one.

closes #5651

### DIFF
--- a/indico/testing/fixtures/app.py
+++ b/indico/testing/fixtures/app.py
@@ -15,14 +15,20 @@ from indico.web.flask.wrappers import IndicoFlask
 
 
 @pytest.fixture(scope='session')
-def app(request, redis_proc):
+def app(request):
     """Create the flask app."""
+
+    redis_cache_url = os.environ.get('INDICO_TEST_REDIS_CACHE_URL')
+    if not redis_cache_url:
+        redis_proc = request.getfixturevalue('redis_proc')
+        redis_cache_url = f'redis://{redis_proc.host}:{redis_proc.port}/0'
+
     config_override = {
         'BASE_URL': 'http://localhost',
         'SMTP_SERVER': ('localhost', 0),  # invalid port - just in case so we NEVER send emails!
         'TEMP_DIR': request.config.indico_temp_dir.strpath,
         'CACHE_DIR': request.config.indico_temp_dir.strpath,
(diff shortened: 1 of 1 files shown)


ThiefMaster:
LGTM. I assume you tested the env var support? Since the normal behavior works fine according to CI I'll go ahead and merge it in that case.

tomako:
Yes, it works in my environment using the env var.

PR #5653, opened 2023-02-06
Clarify that you cannot import events from a different Indico instance

### LINKED ISSUE #5619: Import: Make it clear that you cannot import from a different indico URL
See https://talk.getindico.io/t/import-events-from-other-instances/2915/1 for example.

Pasting an event URL is convenient, but we should check not only the ID but also the host - if it doesn't match the current instance URL, a meaningful error should be shown instead of "event does not exist".

Note: This is a very easy task which would be a nice first task for a new team member, so I recommend we leave this open for now.

GovernmentPlates:
Closes #5619

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,8 @@ Improvements
 - Add button to adapt column widths in paper & contribution lists (:pr:`5642`)
 - Add event language settings to set default and additional languages (:issue:`5606`,
   :pr:`5607`, thanks :user:`vasantvohra`)
+- Fail nicely when trying to import an event from another Indico instance (:issue:`5619`,
+  :pr:`5653`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/events/util.py
+++ b/indico/modules/events/util.py
@@ -629,12 +629,15 @@ def check_permissions(event, field, allow_networks=False):
 
 
 def get_event_from_url(url):
+    """Get an event from an Indico event URL."""
     data = url_parse(url)
     if not all([data.scheme, data.netloc, data.path]):
         raise ValueError(_('Invalid event URL'))
     event_path = re.match(r'/event/(\d+)(/|$)', data.path)
     if not event_path:
         raise ValueError(_('Invalid event URL'))
+    if not url.startswith(config.BASE_URL):
+        raise ValueError(_('Events from other Indico instances cannot be imported'))
     event_id = event_path.group(1)
     event = Event.get(event_id, is_deleted=False)
     if not event:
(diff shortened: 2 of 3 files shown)


indico/modules/events/util.py
```diff
@@ -629,12 +629,15 @@ def check_permissions(event, field, allow_networks=False):
 
 
 def get_event_from_url(url):
+    """Gets an event from an Indico event URL."""
```
ThiefMaster, line 632:
```suggestion
    """Get an event from an Indico event URL."""
```

indico/modules/events/util_test.py
```diff
@@ -0,0 +1,29 @@
...
+
+import random
+
+import pytest
+
+from indico.core.config import config
+from indico.modules.events.util import get_event_from_url
+
+
+@pytest.mark.parametrize(('dummy_event_url', 'asserted_error_message'), (
```
ThiefMaster, line 16: I'd just call it `url` here

indico/modules/events/util_test.py
```diff
@@ -0,0 +1,29 @@
...
+    ('http://example.com/foo', 'Invalid event URL'),
+    ('http://example.com/event/123', 'Events from other Indico instances cannot be imported'),
+))
+def test_get_event_from_url_raises_value_error(dummy_event_url, asserted_error_message):
+    with pytest.raises(ValueError, match=asserted_error_message):
+        get_event_from_url(dummy_event_url)
+
+
+def test_get_event_from_url_returns_event(dummy_event):
+    dummy_event.id = random.randint(1, 1000)
```
ThiefMaster, line 27: :x: not needed

indico/modules/events/util_test.py
```diff
@@ -0,0 +1,29 @@
...
+    event = get_event_from_url(f'{config.BASE_URL}/event/{dummy_event.id}')
```
ThiefMaster, line 28: you can hardcode localhost here, the config is static during tests

CHANGES.rst
```diff
@@ -28,6 +28,8 @@ Improvements
 - Add button to adapt column widths in paper & contribution lists (:pr:`5642`)
 - Add event language settings to set default and additional languages (:issue:`5606`,
   :pr:`5607`, thanks :user:`vasantvohra`)
+- Fail nicely when trying to import from external Indico instance (:issue:`5619`,
```
ThiefMaster, line 31:
```suggestion
- Fail nicely when trying to import an event from another Indico instance (:issue:`5619`,
```

sounds a bit better IMO

ThiefMaster (approved):

PR #5654, opened 2023-02-06
Allow sending reminders to pending invitees

### LINKED ISSUE #5579: Allow sending reminders to pending invites
It would be useful to be able to send reminders to people that have been invited to an event but have not yet registered.

A button could be added on the invitation page as such:

Screen Shot 2022-11-23 at 10 00 16

tomasr8:
Closes #5579 

Adds a new section on the invitations page to email invitees who have not yet registered.
Uses the new React `<EmailDialog />` component.

The button is disabled if there are no pending invitations.

image

### DIFF
--- a/indico/modules/events/persons/client/js/EmailDialog.jsx
+++ b/indico/modules/events/persons/client/js/EmailDialog.jsx
@@ -9,7 +9,7 @@ import _ from 'lodash';
 import PropTypes from 'prop-types';
 import React, {useState} from 'react';
 import {FormSpy} from 'react-final-form';
-import {Form, Button, Message} from 'semantic-ui-react';
+import {Form, Button, Message, Input, Popup, Icon} from 'semantic-ui-react';
 
 import {FinalEmailList} from 'indico/react/components';
 import PlaceholderInfo from 'indico/react/components/PlaceholderInfo';
@@ -18,6 +18,41 @@ import {FinalCheckbox, FinalDropdown, FinalInput} from 'indico/react/forms';
 import {FinalModalForm} from 'indico/react/forms/final-form';
 import {Translate, PluralTranslate, Singular, Plural, Param} from 'indico/react/i18n';
 import {indicoAxios} from 'indico/utils/axios';
+
+function RecipientsField({recipients}) {
+  return (
+    <Form.Field>
+      <Translate as="label">Recipients</Translate>
+      <Input
+        value={recipients.join(', ')}
+        readOnly
+        icon={
+          navigator.clipboard && (
+            <Popup
+              content={Translate.string('Copied!')}
+              on="click"
+              position="top center"
+              inverted
+              trigger={
+                <Icon
+                  name="copy"
+                  color="black"
+                  title={Translate.string('Copy to clipboard')}
+                  onClick={() => navigator.clipboard.writeText(recipients.join(', '))}
+                  link
+                />
+              }
+            />
+          )
+        }
+      />
+    </Form.Field>
+  );
+}
+
+RecipientsField.propTypes = {
+  recipients: PropTypes.arrayOf(PropTypes.string).isRequired,
+};
 
 export function EmailSentMessage({count}) {
   return (
@@ -45,6 +80,7 @@ export function EmailDialog({
   previewURL,
   previewContext,
   placeholders,
+  recipients,
   recipientsField,
   initialFormValues,
 }) {
@@ -124,7 +160,7 @@ export function EmailDialog({
             <PlaceholderInfo placeholders={placeholders} />
           </Form.Field>
         )}
-        {recipientsField}
+        {recipientsField || <RecipientsField recipients={recipients} />}
         <FinalEmailList
           name="bcc_addresses"
           label={Translate.string('BCC addresses')}
@@ -177,7 +213,10 @@ EmailDialog.propTypes = {
   previewURL: PropTypes.string.isRequired,
   previewContext: PropTypes.object,
   placeholders: PropTypes.array.isRequired,
-  recipientsField: PropTypes.node.isRequired,
+  /** Array of email addresses passed to <RecipientsField /> */
+  recipients: PropTypes.arrayOf(PropTypes.string),
+  /** React node to render instead of the default <RecipientsField /> component. */
+  recipientsField: PropTypes.node,
   initialFormValues: PropTypes.object,
 };
 
@@ -185,4 +224,6 @@ EmailDialog.defaultProps = {
   previewContext: {},
   initialFormValues: {},
   sentEmailsCount: 0,
+  recipients: [],
+  recipientsField: null,
 };
--- a/indico/modules/events/registration/__init__.py
+++ b/indico/modules/events/registration/__init__.py
@@ -195,6 +195,19 @@ def _get_registration_placeholders(sender, regform, registration, **kwargs):
     yield FieldPlaceholder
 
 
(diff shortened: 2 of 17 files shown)


indico/modules/events/persons/client/js/EmailDialog.jsx
```diff
@@ -177,12 +213,19 @@ EmailDialog.propTypes = {
   previewURL: PropTypes.string.isRequired,
   previewContext: PropTypes.object,
   placeholders: PropTypes.array.isRequired,
-  recipientsField: PropTypes.node.isRequired,
+  recipients: PropTypes.arrayOf(PropTypes.string),
+  /** React node to render instead of the default <RecipientsField /> component.
```
ThiefMaster, line 217: Don't we wrap after `/**` in other places? (e.g. in FinalModalForm)

indico/modules/events/persons/client/js/EmailDialog.jsx
```diff
@@ -177,12 +213,19 @@ EmailDialog.propTypes = {
...
+   * used when the 'recipients' prop is passed.
+   * 'recipients' will be ignored when using this prop.
```
ThiefMaster, line 219: These two lines sound really weird...

indico/modules/events/registration/__init__.py
```diff
@@ -195,6 +195,20 @@ def _get_registration_placeholders(sender, regform, registration, **kwargs):
     yield FieldPlaceholder
 
 
+@signals.core.get_placeholders.connect_via('registration-reminder-email')
+def _get_placeholders(sender, invitation=None, event=None, **kwargs):
+    from indico.modules.events.placeholders import EventLinkPlaceholder, EventTitlePlaceholder
+    from indico.modules.events.registration.placeholders.invitations import (EmailPlaceholder, FirstNamePlaceholder,
+                                                                             InvitationLinkPlaceholder,
+                                                                             LastNamePlaceholder)
```
ThiefMaster, line 203: module import here as well? ;)

indico/modules/events/registration/__init__.py
```diff
@@ -195,6 +195,20 @@ def _get_registration_placeholders(sender, regform, registration, **kwargs):
     yield FieldPlaceholder
 
 
+@signals.core.get_placeholders.connect_via('registration-reminder-email')
```
ThiefMaster, line 198: `registration-invitation-reminder-email`?

indico/modules/events/registration/client/js/components/EmailPendingInvitations.jsx
```diff
@@ -0,0 +1,79 @@
...
+export default function EmailPendingInvitations({
+  metadataURL,
+  previewURL,
+  sendURL,
+  onClose,
+  successTimeout,
+}) {
+  const [sentCount, setSentCount] = useState(0);
+
+  const {data, loading} = useIndicoAxios({url: metadataURL, method: 'GET'});
```
ThiefMaster, line 26:
```suggestion
  const {data, loading} = useIndicoAxios(metadataURL);
```

indico/modules/events/registration/client/js/components/EmailPendingInvitations.jsx
```diff
@@ -0,0 +1,79 @@
...
+EmailPendingInvitations.propTypes = {
+  metadataURL: PropTypes.string.isRequired,
+  previewURL: PropTypes.string.isRequired,
+  sendURL: PropTypes.string.isRequired,
+  onClose: PropTypes.func.isRequired,
+  successTimeout: PropTypes.number,
+};
+
+EmailPendingInvitations.defaultProps = {
+  successTimeout: 5000,
```
ThiefMaster, line 78: does this actually need to be a prop? I don't see why we'd ever pass a different one ;)
(copy&paste from another dialog i guess)

indico/modules/events/registration/controllers/management/invitations.py
```diff
@@ -62,10 +88,79 @@ def _process(self):
...
+    """Preview a registration reminder email."""
+
+    @use_kwargs({
+        'body': fields.String(required=True),
+        'subject': fields.String(required=True),
+    })
+    def _process(self, body, subject):
+        invitations = _query_pending_invitations(self.regform)
+        if not invitations:
+            abort(404)
```
ThiefMaster, line 154: Not a big fan of `abort()` for this - it shows 404 with a default "Not found" error which wouldn't be very clear. I think `BadRequest('No invitations')` would be clearer

indico/modules/events/registration/controllers/management/invitations.py
```diff
@@ -62,10 +88,79 @@ def _process(self):
...
+    def _process(self, from_address, body, subject, bcc_addresses, copy_for_sender):
+        if from_address not in self.event.get_allowed_sender_emails():
+            abort(422, messages={'from_address': ['Invalid sender address']})
+        invitations = _query_pending_invitations(self.regform)
+        for invitation in invitations:
+            email_body = replace_placeholders('registration-reminder-email', body, event=self.event,
+                                              invitation=invitation)
+            email_subject = replace_placeholders('registration-reminder-email', subject,
+                                                 event=self.event, invitation=invitation)
+            tpl = get_template_module('emails/custom.html', subject=email_subject, body=email_body)
```
ThiefMaster, line 114: This should be within the `force_event_locale` block too - I think we only render stuff from within the macros, but IMHO it's cleaner to keep everything inside..

indico/modules/events/registration/controllers/management/invitations.py
```diff
@@ -62,10 +88,79 @@ def _process(self):
...
+    """Get the metadata for registration reminder emails."""
+
+    def _process(self):
+        with self.event.force_event_locale():
+            tpl = get_template_module('events/registration/emails/registration_reminder_default.html', event=self.event)
+            body = tpl.get_html_body()
+            subject = tpl.get_subject()
+        placeholders = get_sorted_placeholders('registration-reminder-email')
+        invitations = _query_pending_invitations(self.regform)
+        recipients = sorted([invitation.email for invitation in invitations])
```
ThiefMaster, line 134:
```suggestion
        recipients = sorted(invitation.email for invitation in invitations)
```

indico/modules/events/registration/client/js/components/EmailPendingInvitations.jsx
tomasr8: Even if we may never never pass it, I think it's better to give it a name rather than have a numeric constant in the code.
Maybe using `const successTimeout = 5000` is better inside the component itself?

indico/modules/events/registration/controllers/management/invitations.py
tomasr8: There's two other occurrences of this in the codebase, should I change those as well (in a separate PR)?

indico/modules/events/registration/client/js/components/EmailPendingInvitations.jsx
ThiefMaster: IMHO it's still pretty clear from the context (I never liked the blanked "no magic numbers" thing popular in the java world) but if you prefer to use a constant i'm fine with it of course

indico/modules/events/registration/controllers/management/invitations.py
ThiefMaster: or just give me a commit to cherrypick... seems like the kind of thing not really worth a PR

ThiefMaster (approved):

PR #5656, opened 2023-02-06
Show paragraphs in editor fields

### LINKED ISSUE #5624: Paragraphs in editors are missing spacing
**Describe the bug**
When you write something in a ck editor, e.g. conference description, there is no visible difference between a line break and a paragraph. When you view it in the result, there is however. This is mildly confusing.

**To Reproduce**
1. Create an event with a description. Add some text, a few line breaks, a paragraph
2. Save the description
3. Compare what you see in the editor with the final result

**Expected behavior**
Paragraph spacing in the editor

**Screenshots**
image
image

**Additional context**
This apparently goes back to this change https://github.com/indico/indico/commit/1541785758f00be132caaa1bfa53a6d85a3e9045 for IE, which I guess we could revert? At the very least, removing that CSS rule will make it default to semantic-ui's 1em margin-bottom.

kewisch:
Fixes #5624

### DIFF
--- a/indico/web/client/styles/partials/_inputs.scss
+++ b/indico/web/client/styles/partials/_inputs.scss
@@ -147,29 +147,6 @@ ul {
   }
 }
 
-div[contenteditable],
-span[contenteditable] {
-  p {
-    margin: 0;
-  }
-
-  &:hover {
-    outline: 1px dashed $default-border-color;
-  }
-
-  &:focus,
-  &.focus {
-    background: white;
-    box-shadow: 1px 1px 4px rgba(20, 20, 20, 0.1) inset;
-    outline: 1px solid $blue;
-  }
-
-  &.empty {
-    color: $gray;
-    font-style: italic;
-  }
-}
-
 .i-selection {
   input[type='checkbox'],
   input[type='radio'] {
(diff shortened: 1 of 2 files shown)


indico/web/client/styles/partials/_inputs.scss
```diff
@@ -149,10 +149,6 @@ ul {
 
 div[contenteditable],
 span[contenteditable] {
```
ThiefMaster, line 151: I wonder if we need *any* of that contenteditable styling oonsidering that we no longer use it ourselves anywhere in the codebase and ckeditor should have its own internal styling...

kewisch: Yeah, those were mostly unused. There was a dashed border on hover, but I don't think this is necessary.

ThiefMaster: in fact the dashed border on hover looked weirdt here but was so subtle that probably nobody noticed it. this whole style was meant for another place where we now actually have a simple textarea for plaintext+markdown (in the past we used contenteditable for registration form sections, probably to allow people to paste formatted content from elsewhere - with all the disadvantages that come with that)

ThiefMaster (approved):

PR #5659, opened 2023-02-07
Add extra_log_fields to events and categories logging

jbtwist:
As discussed, I'm adding _extra_log_fields to logger functions in update categories and events.

### DIFF
--- a/indico/modules/categories/operations.py
+++ b/indico/modules/categories/operations.py
@@ -51,22 +51,16 @@ def move_category(category, target_category):
 
 
 @make_interceptable
-def update_category(category, data, skip=()):
-    assert set(data) <= {
-        'title', 'description', 'timezone', 'suggestions_disabled', 'is_flat_view_enabled',
-        'event_message_mode', 'event_message', 'notify_managers', 'event_creation_notification_emails',
-        *skip
-    }
+def update_category(category, data, *, skip=(), _extra_log_fields=None):
     changes = category.populate_from_dict(data, skip=skip)
     db.session.flush()
     signals.category.updated.send(category)
     logger.info('Category %s updated with %r by %s', category, data, session.user)
-    _log_category_update(category, changes)
+    _log_category_update(category, changes, _extra_log_fields)
 
 
 @make_interceptable
-def update_category_protection(category, data):
-    assert set(data) <= {'protection_mode', 'own_no_access_contact', 'event_creation_mode', 'visibility'}
+def update_category_protection(category, data, *, _extra_log_fields=None):
     changes = category.populate_from_dict(data)
     db.session.flush()
     signals.category.updated.send(category, changes=changes)
@@ -76,13 +70,13 @@ def update_category_protection(category, data):
                       'own_no_access_contact': 'No access contact',
                       'visibility': {'title': 'Visibility', 'type': 'string',
                                      'convert': lambda changes: [format_visibility(category, x) for x in changes]},
-                      'event_creation_mode': 'Event creation mode'}
+                      'event_creation_mode': 'Event creation mode',
+                      **(_extra_log_fields or {})}
         category.log(CategoryLogRealm.category, LogKind.change, 'Category', 'Protection updated', session.user,
                      data={'Changes': make_diff_log(changes, log_fields)})
 
 
-@make_interceptable
-def _log_category_update(category, changes):
+def _log_category_update(category, changes, extra_log_fields):
     log_fields = {
         'title': {'title': 'Title', 'type': 'string'},
         'description': 'Description',
@@ -92,7 +86,8 @@ def _log_category_update(category, changes):
         'event_message_mode': 'Event header message type',
         'event_message': 'Event header message',
         'notify_managers': 'Notify managers about event creation',
-        'event_creation_notification_emails': 'Event creation notification emails'
+        'event_creation_notification_emails': 'Event creation notification emails',
+        **(extra_log_fields or {})
     }
     if changes:
         what = 'Settings'
--- a/indico/modules/events/operations.py
+++ b/indico/modules/events/operations.py
@@ -140,12 +140,7 @@ def create_event(category, event_type, data, add_creator_as_manager=True, featur
 
 @make_interceptable
 @no_autoflush
-def update_event(event, update_timetable=False, **data):
-    assert set(data.keys()) <= {'title', 'description', 'url_shortcut', 'location_data', 'keywords',
-                                'person_link_data', 'start_dt', 'end_dt', 'timezone', 'keywords', 'references',
-                                'organizer_info', 'additional_info', 'contact_title', 'contact_emails',
-                                'contact_phones', 'start_dt_override', 'end_dt_override', 'label', 'label_message',
-                                'own_map_url', 'supported_locales', 'default_locale'}
+def update_event(event, *, update_timetable=False, _extra_log_fields=None, **data):
     old_person_links = event.sorted_person_links[:]
     changes = {}
     if (update_timetable or event.type == EventType.lecture) and 'start_dt' in data:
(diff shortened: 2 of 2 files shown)


indico/modules/categories/controllers/management.py
```diff
@@ -110,7 +110,7 @@ def _process(self):
         logo_form = CategoryLogoForm(obj=self.category)
 
         if form.validate_on_submit():
-            update_category(self.category, form.data, skip={'meeting_theme', 'lecture_theme'})
+            update_category(self.category, form.data, skip={'meeting_theme', 'lecture_theme'}, _extra_log_fields={})
```
ThiefMaster, line 113: No need to pass this here

indico/modules/categories/operations.py
```diff
@@ -51,12 +51,7 @@ def move_category(category, target_category):
 
 
 @make_interceptable
-def update_category(category, data, skip=()):
-    assert set(data) <= {
-        'title', 'description', 'timezone', 'suggestions_disabled', 'is_flat_view_enabled',
-        'event_message_mode', 'event_message', 'notify_managers', 'event_creation_notification_emails',
-        *skip
-    }
+def update_category(category, data, skip=(), _extra_log_fields={}):
```
ThiefMaster, line 54: Do not use mutable default values - use `None` here and `whatever or {}` when you need a dict.

indico/modules/categories/controllers/management.py
jbtwist: Without passing it there, that parameter didn't appear in the method interception.

I found a solution passing it that way, is there another better solution?

indico/modules/events/operations.py
```diff
@@ -230,7 +225,8 @@ def clone_into_event(source_event, target_event, cloners):
     return target_event
 
 
-def _log_event_update(event, changes, visible_person_link_changes=False):
+@make_interceptable
```
OmeGak, line 228: We don't really need to intercept this function.

indico/modules/categories/operations.py
```diff
@@ -76,13 +70,14 @@ def update_category_protection(category, data):
...
                       'visibility': {'title': 'Visibility', 'type': 'string',
                                      'convert': lambda changes: [format_visibility(category, x) for x in changes]},
-                      'event_creation_mode': 'Event creation mode'}
+                      'event_creation_mode': 'Event creation mode',
+                      **(_extra_log_fields or {})}
         category.log(CategoryLogRealm.category, LogKind.change, 'Category', 'Protection updated', session.user,
                      data={'Changes': make_diff_log(changes, log_fields)})
 
 
 @make_interceptable
```
OmeGak, line 79: We don't really need to intercept this function.

indico/modules/categories/operations.py
```diff
@@ -51,22 +51,16 @@ def move_category(category, target_category):
...
-def update_category(category, data, skip=()):
-    assert set(data) <= {
-        'title', 'description', 'timezone', 'suggestions_disabled', 'is_flat_view_enabled',
-        'event_message_mode', 'event_message', 'notify_managers', 'event_creation_notification_emails',
-        *skip
-    }
+def update_category(category, data, _extra_log_fields, skip=()):
```
ThiefMaster, line 54: This is now a required argument, that's not OK - we do not want to always pass this when calling the method in the core.

```suggestion
def update_category(category, data, *, skip=(), _extra_log_fields=None):
```

indico/modules/categories/operations.py
```diff
@@ -51,22 +51,16 @@ def move_category(category, target_category):
...
     signals.category.updated.send(category)
     logger.info('Category %s updated with %r by %s', category, data, session.user)
-    _log_category_update(category, changes)
+    _log_category_update(category, changes, _extra_log_fields)
 
 
 @make_interceptable
-def update_category_protection(category, data):
-    assert set(data) <= {'protection_mode', 'own_no_access_contact', 'event_creation_mode', 'visibility'}
+def update_category_protection(category, data, _extra_log_fields):
```
ThiefMaster, line 63: Same here:

```suggestion
def update_category_protection(category, data, *, _extra_log_fields=None):
```

indico/modules/events/operations.py
```diff
@@ -140,12 +140,7 @@ def create_event(category, event_type, data, add_creator_as_manager=True, featur
 
 @make_interceptable
 @no_autoflush
-def update_event(event, update_timetable=False, **data):
-    assert set(data.keys()) <= {'title', 'description', 'url_shortcut', 'location_data', 'keywords',
-                                'person_link_data', 'start_dt', 'end_dt', 'timezone', 'keywords', 'references',
-                                'organizer_info', 'additional_info', 'contact_title', 'contact_emails',
-                                'contact_phones', 'start_dt_override', 'end_dt_override', 'label', 'label_message',
-                                'own_map_url', 'supported_locales', 'default_locale'}
+def update_event(event, _extra_log_fields, update_timetable=False, **data):
```
ThiefMaster, line 143: Similar here...

```suggestion
def update_event(event, *, update_timetable=False, _extra_log_fields=None, **data):
```

indico/modules/events/operations.py
```diff
@@ -319,9 +315,8 @@ def _format_person(data):
     return f'{data.full_name} <{data.email}>' if data.email else data.full_name
 
 
-def update_event_protection(event, data):
-    assert set(data.keys()) <= {'protection_mode', 'own_no_access_contact', 'access_key',
-                                'visibility', 'public_regform_access'}
+@make_interceptable
+def update_event_protection(event, data, _extra_log_fields):
```
ThiefMaster, line 319: And here as well...

```suggestion
def update_event_protection(event, data, *, _extra_log_fields=None):
```

indico/modules/events/operations.py
```diff
@@ -332,7 +327,8 @@ def update_event_protection(event, data):
                       'access_key': {'title': 'Access key', 'type': 'string'},
                       'visibility': {'title': 'Visibility', 'type': 'string',
                                      'convert': lambda changes: [format_visibility(event, x) for x in changes]},
-                      'public_regform_access': 'Public registration form access'}
+                      'public_regform_access': 'Public registration form access',
+                      **(_extra_log_fields)}
```
ThiefMaster, line 331:
```suggestion
                      **(_extra_log_fields or {})}
```

otherwise it fails if the arg is omitted

indico/modules/events/operations.py
```diff
@@ -230,7 +225,7 @@ def clone_into_event(source_event, target_event, cloners):
     return target_event
 
 
-def _log_event_update(event, changes, visible_person_link_changes=False):
+def _log_event_update(event, changes, _extra_log_fields, visible_person_link_changes=False):
```
ThiefMaster, line 228: In here I would just call it `extra_log_fields` because it's an internal function anyway - the point of marking it as internal on the `update_*` functions is to make it clear that it usually shouldn't be used (for example a developer on the core should not use it but rather update the `log_fields` dict here)

indico/modules/categories/operations.py
```diff
@@ -76,13 +70,13 @@ def update_category_protection(category, data):
...
-                      'event_creation_mode': 'Event creation mode'}
+                      'event_creation_mode': 'Event creation mode',
+                      **(_extra_log_fields or {})}
         category.log(CategoryLogRealm.category, LogKind.change, 'Category', 'Protection updated', session.user,
                      data={'Changes': make_diff_log(changes, log_fields)})
 
 
-@make_interceptable
-def _log_category_update(category, changes):
+def _log_category_update(category, changes, _extra_log_fields):
```
ThiefMaster, line 79: See my comment on `_log_event_update` about the underscore

ThiefMaster (changes requested):

ThiefMaster (approved):

PR #5661, opened 2023-02-07
Update react search hooks

vasantvohra:


ThiefMaster:
Considering that this is originally my code I'll push it directly ;) I don't care much about commit credits but I still have it stashed anyway...

PR #5668, opened 2023-02-10
Allow services to set editable as ready for review

duartegalvao:
This PR allows microsservices to respond to the create-editable handler with a `ready_for_review` flag, which overrides the editable's state from "new" to "ready for review". This makes it possible to have microsservices that don't submit a revision when a new editable is created. Otherwise, the editable would be stuck in a "new" state, making it impossible to make any reviews.

This PR also fixes a front-end bug in which the "Judge" button appears when a revision is not yet ready to review.


indico/modules/events/editing/service.py
```diff
@@ -149,7 +150,12 @@ def service_handle_new_editable(editable, user):
         )
         resp = requests.put(_build_url(editable.event, path), headers=_get_headers(editable.event), json=data)
         resp.raise_for_status()
-    except requests.RequestException as exc:
+        resp = ServiceCreateEditableResultSchema().load(resp.json())
```
ThiefMaster, line 153: hm shouldn't we still allow returning an empty response?

CHANGES.rst
```diff
@@ -35,6 +35,8 @@ Improvements
 - Hide the top box with the latest files of an editable until it has been accepted
   and published (:issue:`5660`, :pr:`5665`)
 - Allow uploading files when requesting changes on the editing timeline (:pr:`5612`)
+- Allow microsservices to mark new editables as ready-for-review without creating
```
ThiefMaster, line 38: "Allow an editing service [check how we call it in the UI] ..."

(and i'd move it to the "internal changes" section)

indico/modules/events/editing/service.py
```diff
@@ -149,7 +150,12 @@ def service_handle_new_editable(editable, user):
         )
         resp = requests.put(_build_url(editable.event, path), headers=_get_headers(editable.event), json=data)
         resp.raise_for_status()
-    except requests.RequestException as exc:
+        resp = ServiceCreateEditableResultSchema().load(resp.json()) if resp.text else {}
+
+        if resp.get('ready_for_review', False):
```
ThiefMaster, line 155:
```suggestion
        if resp.get('ready_for_review'):
```

the default `None` is falsy as well

PR #5670, opened 2023-02-13
Introduce ACLs for menu entries

kewisch:
Hey @ThiefMaster, I'd love to get some feedback on this approach, as it is turning out to be massively more complicated than I initially expected. There are still quite a few loose ends, but I want to see if I am going into the right direction.

We sometimes have sponsored attendees at our events. We'd like to show them a page in the sidebar with the expense policies. We don't want this page to be visible to everyone else.

I'd like Menu Entries to have ACLs so I can restrict viewing the and their visibility in the menu to people who have registered using a specific form.

Right now, all we can do is send them the link to an unlisted page, or open up the expense policy page to everyone (non-sponsored attendees, random remote participants)

**Current Status**
I've hooked up UI using the permission list field and have taken a stab at the backend database. A few features are still missing:
* An explicit way to make it available to "Everyone" - I was thinking to either do so if the ACLs are empty, or add another token principal "Everyone" that needs to be explicitly added.
* Protection mode - I'll have to play around with this. I see a few ways protection modes could map to different scenarios, it could also be a way to avoid a specific "Everyone" principal.
* I've not yet managed to get saving settings to work due to some backref issues, it isn't finding the menu entry from the principal.
* Need to add a speakers button to the UI that uses the new principle.
* Probably makes sense to hook up event and category roles as well
* There are a few functions that need to be implemented for completeness, though I've not yet seen if they are used in this scenario.

Questions:
* Where I am struggling most is the database setup and fields. I've seen multiple ways to do this in the codebase (sqlalchemy schemas vs flask sqlalchemy), and there is no 1:1 comparison I can make. Am I missing anything substantial? Especially in the backrefs, I'm not sure when I need a db backref, a normal backref, and what properties I need to implement for a backref.
* I've given the speaker principal an id, which is the event id. I think we could do without and just make this a simple token principal, the event data being taken from context. Makes sense? I'd still need to have something in `SpeakerPrincipal` so that I can reference the event though.
* I'm using the`ProtectionMixin` since it provides many convenience functions. Is this the right path, or is it going overboard for a simple yes/no type of access?

Any other comments you might have are appreciated.

**Screenshots**
image

### DIFF
--- a/indico/migrations/versions/20230306_0905_e47fc6634291_add_menu_entry_acls.py
+++ b/indico/migrations/versions/20230306_0905_e47fc6634291_add_menu_entry_acls.py
@@ -1,0 +1,350 @@
+"""Add menu entry ACLs
+
+Revision ID: e47fc6634291
+Revises: 5d05eda06776
+Create Date: 2023-03-06 09:05:08.848298
+"""
+
+from enum import Enum
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.sql import expression
+
+from indico.core.db.sqlalchemy import PyIntEnum
+
+
+# revision identifiers, used by Alembic.
+revision = 'e47fc6634291'
+down_revision = '5d05eda06776'
+branch_labels = None
+depends_on = None
+
+
+class _ProtectionMode(int, Enum):
+    public = 0
+    inheriting = 1
+    protected = 2
+
+
+class _PrincipalType(int, Enum):
+    user = 1
+    local_group = 2
+    multipass_group = 3
+    email = 4
+    network = 5
+    event_role = 6
+    category_role = 7
+    registration_form = 8
+
+
+class _MenuEntryAccess(int, Enum):
+    everyone = 1
+    registered_participants = 2
+    speakers = 3
+
+
+class _MenuEntryType(int, Enum):
+    separator = 1
+    internal_link = 2
+    user_link = 3
+    plugin_link = 4
+    page = 5
+
+
+def _upgrade_access_data():
+    """
+    Upgrade the access column to protection_mode, speaker_allowed, and the principals.
+
+    Here is the mapping:
+        access                  || protection_mode | speaker_allowed | menu_entry_principals
+        ========================++=================+=================+==========================
+        everyone                ||  inheriting     |   false         | none set
+        ------------------------++-----------------+-----------------+--------------------------
+        registered_participants ||  protected      |   false         | One reg pincipal for each
+                                ||                 |                 | registration in the event
+        ------------------------++-----------------+-----------------+--------------------------
+        speakers                ||  protected      |   true          | none set
+    """
+    conn = op.get_bind()
+
+    # Update access mode to protected for registered_participants and speakers
+    conn.execute(
+        sa.text(
+            '''
+                    UPDATE events.menu_entries
+                       SET protection_mode = :protected
+                     WHERE access != :everyone
+            '''
+        ),
+        protected=_ProtectionMode.protected.value,
+        everyone=_MenuEntryAccess.everyone.value,
+    )
+
+    # Migrate "Registered participants" to principal entries for each registration form
+    conn.execute(
+        sa.text(
+            '''
+               INSERT INTO events.menu_entries_principals(menu_entry_id, type, registration_form_id)
+                    SELECT me.id AS menu_entry_id,
+                           :principal_type AS type,
+                           frm.id AS registration_form_id
+                      FROM events.menu_entries me
+          RIGHT OUTER JOIN event_registration.forms frm ON me.event_id = frm.event_id
+                     WHERE me.type NOT IN (:internal_link, :separator)
+                       AND me.access = :registered
+            '''
+        ),
+        principal_type=_PrincipalType.registration_form.value,
+        internal_link=_MenuEntryType.internal_link.value,
+        separator=_MenuEntryType.separator.value,
+        registered=_MenuEntryAccess.registered_participants.value,
+    )
+
+    # Migrate "Speaker" access to allowing speaker access, unless we've already provided broader
+    # access for registered participants
+    conn.execute(
+        sa.text(
+            '''
+                    UPDATE events.menu_entries
+                       SET speaker_allowed = TRUE
+                     WHERE access = :speakers
+                       AND id NOT IN (
+                         SELECT menu_entry_id AS id
+                           FROM events.menu_entries_principals mep
+                          WHERE mep.type = :regform
+                       )
+            '''
+        ),
+        speakers=_MenuEntryAccess.speakers.value,
+        regform=_PrincipalType.registration_form.value,
+    )
+
+
+def _downgrade_access_data():
+    """
+    Downgrade the protection_mode, speaker_allowed and principals to the access column. This isn't
+    lossless given the new feature is more fine-grained, but best effort.
+
+    Here is the mapping:
+
+         protection_mode | speaker_allowed | menu_entry_principals     || access
+        =================+=================+===========================++========================
+          protected      | true            | doesn't contain regform   ||  speakers
+                         |                 | principals                ||
+        -----------------+-----------------+---------------------------++------------------------
+          inheriting     |                 | doesn't contain regform   ||  speakers
+        parent=protected | parent=true     | principals                ||
+        -----------------+-----------------+---------------------------++------------------------
+          protected      | regardless      | at least one regform      ||  registered_participants
+                         |                 | principal                 ||
+        -----------------+-----------------+---------------------------++------------------------
+          inheriting     | regardless      | parent has at least one   ||  registered_participants
+                         |                 | regform principal         ||
+        -----------------+-----------------+---------------------------++------------------------
+          protected      | false           | doesn't contain regform   ||  everyone
+                         |                 | principals                ||  (no good mapping)
+        -----------------+-----------------+---------------------------++------------------------
+          inheriting     | false           | doesn't contain regform   ||  everyone
+                         |                 | principals                ||
+        -----------------+-----------------+---------------------------++------------------------
+          all other combinations                                       ||  everyone
+                                                                       ||  (no good mapping)
+        -----------------+-----------------+---------------------------++------------------------
+    """
+    conn = op.get_bind()
+
+    # The server_default is set to give everyone access to all menu entries. We need to restrict
+    # those appropriately now. Start with locking down for speakers, and then open up to registered
+    # participants if applicable
+
+    # Any menu entry that is accessible to speakers should only be open to speakers. If it is also
+    # open to registrants, we'll expand this later on.
+    conn.execute(
+        sa.text(
+            '''
+                    UPDATE events.menu_entries
+                       SET access = :speakers
+                     WHERE speaker_allowed = TRUE
+                       AND protection_mode = :protected
+            '''
+        ),
+        speakers=_MenuEntryAccess.speakers.value,
+        protected=_ProtectionMode.protected.value
+    )
+
+    # Inheriting menu entries with the parent accessible to speakers should be only open to
+    # speakers. Again expanding to others later on.
+    conn.execute(
+        sa.text(
+            '''
+                    UPDATE events.menu_entries
+                       SET access = :speakers
+                     WHERE menu_entries.id IN (
+                                      SELECT me.id
+                                        FROM events.menu_entries me
+                             LEFT OUTER JOIN events.menu_entries parent_me ON me.parent_id = parent_me.id
+                                       WHERE me.type NOT IN (:internal_link, :separator)
+                                         AND me.parent_id IS NOT NULL
+                                         AND me.protection_mode = :inheriting
+                                         AND parent_me.protection_mode = :protected
+                                         AND parent_me.speaker_allowed = TRUE
+                    )
+
+            '''
+        ),
+        speakers=_MenuEntryAccess.speakers.value,
+        internal_link=_MenuEntryType.internal_link.value,
+        separator=_MenuEntryType.separator.value,
+        inheriting=_ProtectionMode.inheriting.value,
+        protected=_ProtectionMode.protected.value,
+    )
+
+    # Any menu entry that has a registration attached to it should be registered participants only,
+    # provided its protection mode is set to protected
+    conn.execute(
+        sa.text(
+            '''
+                    UPDATE events.menu_entries
+                       SET access = :registered
+                     WHERE menu_entries.id IN (
+                             SELECT menu_entry_id
+                               FROM events.menu_entries_principals
+                              WHERE type = :regform
+                           )
+                       AND protection_mode = :protected
+            '''
+        ),
+        registered=_MenuEntryAccess.registered_participants.value,
+        regform=_PrincipalType.registration_form.value,
+        protected=_ProtectionMode.protected.value,
+    )
+
+    # Menu entries that are inheriting with their parent set to protected and containing
+    # registrations get the same treatment
+    conn.execute(
+        sa.text(
+            '''
+                    UPDATE events.menu_entries
+                       SET access = :registered
+                     WHERE id IN (
+                                      SELECT me.id
+                                        FROM events.menu_entries me
+                             LEFT OUTER JOIN events.menu_entries parent_me ON me.parent_id = parent_me.id
+                            RIGHT OUTER JOIN events.menu_entries_principals mep ON mep.menu_entry_id = parent_me.id
+                                       WHERE me.type NOT IN (:internal_link, :separator)
+                                         AND me.parent_id IS NOT NULL
+                                         AND me.protection_mode = :inheriting
+                                         AND parent_me.protection_mode = :protected
+                                         AND mep.type = :regform
+                           )
+            '''
+        ),
+        registered=_MenuEntryAccess.registered_participants.value,
+        internal_link=_MenuEntryType.internal_link.value,
+        separator=_MenuEntryType.separator.value,
+        inheriting=_ProtectionMode.inheriting.value,
+        protected=_ProtectionMode.protected.value,
+        regform=_PrincipalType.registration_form.value,
+    )
+
+
+def upgrade():
+    # Principals table
+    op.create_table(
+        'menu_entries_principals',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('menu_entry_id', sa.Integer(), nullable=False),
+        sa.Column('type', PyIntEnum(_PrincipalType, exclude_values={_PrincipalType.email, _PrincipalType.network}), nullable=False),
+        sa.Column('user_id', sa.Integer(), nullable=True),
+        sa.Column('local_group_id', sa.Integer(), nullable=True),
+        sa.Column('mp_group_provider', sa.String(), nullable=True),
+        sa.Column('mp_group_name', sa.String(), nullable=True),
+        sa.Column('event_role_id', sa.Integer(), nullable=True),
+        sa.Column('category_role_id', sa.Integer(), nullable=True),
+        sa.Column('registration_form_id', sa.Integer(), nullable=True),
+        sa.CheckConstraint(
+            'type != 1 OR (category_role_id IS NULL AND event_role_id IS NULL AND local_group_id IS NULL AND mp_group_name IS NULL AND mp_group_provider IS NULL AND registration_form_id IS NULL AND user_id IS NOT NULL)',
+            name=op.f('ck_menu_entries_principals_valid_user'),
+        ),
+        sa.CheckConstraint(
+            'type != 2 OR (category_role_id IS NULL AND event_role_id IS NULL AND mp_group_name IS NULL AND mp_group_provider IS NULL AND registration_form_id IS NULL AND user_id IS NULL AND local_group_id IS NOT NULL)',
+            name=op.f('ck_menu_entries_principals_valid_local_group'),
+        ),
+        sa.CheckConstraint(
+            'type != 3 OR (category_role_id IS NULL AND event_role_id IS NULL AND local_group_id IS NULL AND registration_form_id IS NULL AND user_id IS NULL AND mp_group_name IS NOT NULL AND mp_group_provider IS NOT NULL)',
+            name=op.f('ck_menu_entries_principals_valid_multipass_group'),
+        ),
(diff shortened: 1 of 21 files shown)


ThiefMaster:
ProtectionMixin (you also want the option to set it as protected or inheriting) makes perfect sense here. About the special "speaker" principal I'm not sure what's the best solution here, since you can have only one such entry for an ACL. OTOH, being able to grant access to speakers may be interesting in other places as well. I have to think about what's the best approach. What comes to my mind first is just using a column for this special kind of access (ie what we have right now, but to grant access to speakers instead of restricting it to them) and when it's protected we check that plus the ACL.

I'll think a bit about it and get back to you one of these days (I'm off today and kind of busy until at least wednesday, so likely later this week).

indico/modules/events/layout/forms.py
```diff
@@ -162,8 +163,11 @@ class MenuUserEntryFormBase(IndicoForm):
     title = StringField(_('Title'), [DataRequired()])
     is_enabled = BooleanField(_('Show'), widget=SwitchWidget())
     new_tab = BooleanField(_('Open in a new tab'), widget=SwitchWidget())
-    access = IndicoEnumRadioField(_('Visibility'), enum=MenuEntryAccess,
-                                  description=_('Specify who can see this menu item.'))
+    acl_entries = PrincipalListField(_('Visibility'), event=lambda form: form.event, allow_groups=True, allow_event_roles=True, allow_registration_forms=True, allow_speakers=True)
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event')
```
ThiefMaster, line 169: This way of getting a mandatory `event` kwargs is pretty much a Python 2 leftover. Nowadays you can do it in a much nicer way:

```suggestion
    def __init__(self, *args, event, **kwargs):
        self.event = event
```

ThiefMaster:
So after having given this some thought, I think making it an Principal type in the backend is not a good option, because all other principal types reference something, but this one doesn't have anything useful to reference.

I think using a column but integrating it more nicely with the access check logic is the best way to go. You could implement this in `ProtectionMixin`. Check how I implemented e.g. `allow_access_key`: Only when this is set on the class level will the column be added to the model, so you can add generic code for this while keeping it opt-in.

And to keep this plugin from becoming overly complex I'd limit it to:

- adding this new functionalikty
- adding ACL (and inheriting/protected protection modes) for custom pages
- adding a migration step to update the old data, ie marking custom pages as protected when they had restrictions before and converting those old restrictions accordingly: set the new flag to give speakers access if the access was set to `MenuEntryAccess.speakers`, or add all registration forms from the event to the ACL in case it was set to `MenuEntryAccess.registered_participants`

for the migration step you probably want to include this in the same one that also applies the database structure changes so you can first add the new columns and acl table, then migrate the data, and then delete the old unused column (and vice versa in case of downgrade).

kewisch:
What I am worried about is that this creates some hacks for `PrincipalListField`. The field's data is a list of principals, if we have a non-principal field here then we'd need to adjust those to convey both a list of principals, and a list of toggles such as allowing speakers. I'll investigate further if there is an elegant way to do this.

kewisch:
Looks like it won't be too much of an issue, just need to adjust the data structure sent in the form and process it correctly. I have some questions I'll ask inline in a sec.

indico/modules/events/layout/models/menu.py
```diff
@@ -292,12 +282,16 @@ class MenuEntry(MenuEntryMixin, db.Model):
...
+    #: ACLs for page visibility
+    acl_entries = db.relationship(
+        'MenuEntryPrincipal',
+        backref='menu_entry',
+        cascade='all, delete-orphan',
+        collection_class=set
     )
+    # TODO not really sure what I need this for
+    acl = association_proxy('acl_entries', 'principal', creator=lambda v: MenuEntryPrincipal(principal=v))
+
```
kewisch, line 294: I'm really struggling to do the right thing here, maybe you can help. I copied this off how `RoomPrincipal` and the likes are doing it, however when I try to save values it will tell me that the key `menu_entries` is not found on `ClassManager of <class 'indico.modules.events.registration.models.forms.RegistrationForm'>` for example when adding a registration form to the permissions.

Am I setting the backrefs incorrectly? Do I need to add something to User/Group/RegistrationForm/etc to make this work? Or is this a red herring and I'm doing something else wrong?

indico/web/forms/fields/protection.py
```diff
@@ -25,9 +25,10 @@ def __init__(self, *args, **kwargs):
         self.acl_message_url = get_acl_message_url(kwargs['_form']) if get_acl_message_url else None
         self.can_inherit_protection = self.protected_object.protection_parent is not None
         self.is_unlisted_event = isinstance(self.protected_object, db.m.Event) and self.protected_object.is_unlisted
-        if not self.can_inherit_protection and not self.is_unlisted_event:
-            kwargs['skip'] = {ProtectionMode.inheriting}
-        super().__init__(*args, enum=ProtectionMode, **kwargs)
+        # TODO is this just legacy and needs to be merged?
+        #if not self.can_inherit_protection and not self.is_unlisted_event:
+        #    kwargs['skip'] = {ProtectionMode.inheriting}
+        super().__init__(*args, enum=ProtectionMode, skip=self.protected_object.disallowed_protection_modes, **kwargs)
```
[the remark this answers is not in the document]
ThiefMaster: It's not enough because an event can typically have inheriting protection *except* if it's an unlisted event (which has no category to inherit from).

indico/modules/events/layout/models/menu.py
ThiefMaster: I cannot reproduce that error with the current code in the PR. In fact editing works when I replace `acl_entries` with `acl` in the form class.

However, that field is a bit broken for objects not yet in the DB (e.g. when creating a new page). You may want to consider using the field we also use for attachments which is a bit simpler ("protected" switch instead of the "protected/inheriting" toggle):

```patch
diff --git a/indico/modules/events/layout/controllers/menu.py b/indico/modules/events/layout/controllers/menu.py
index 2945bc093e..9240f44e74 100644
--- a/indico/modules/events/layout/controllers/menu.py
+++ b/indico/modules/events/layout/controllers/menu.py
@@ -90,7 +90,7 @@ class RHMenuEntryEditBase(RHMenuBase):

 class RHMenuEntryEdit(RHMenuEntryEditBase):
     def _process(self):
-        defaults = FormDefaults(self.entry)
+        defaults = FormDefaults(self.entry, protected=self.entry.is_self_protected)
         form_kwargs = {}
         if self.entry.is_user_link:
             form_cls = MenuLinkForm
diff --git a/indico/modules/events/layout/forms.py b/indico/modules/events/layout/forms.py
index e75f675c9e..8dfd0b2e4b 100644
--- a/indico/modules/events/layout/forms.py
+++ b/indico/modules/events/layout/forms.py
@@ -15,8 +15,9 @@ from indico.modules.events.layout import theme_settings
 from indico.modules.events.layout.util import get_css_file_data, get_logo_data, get_plugin_conference_themes
 from indico.modules.users import NameFormat
 from indico.util.i18n import _, orig_string
-from indico.web.forms.base import IndicoForm
-from indico.web.forms.fields import EditableFileField, FileField, IndicoEnumSelectField, IndicoProtectionField
+from indico.web.forms.base import IndicoForm, generated_data
+from indico.web.forms.fields import (AccessControlListField, EditableFileField, FileField, IndicoEnumSelectField,
+                                     IndicoProtectionField)
 from indico.web.forms.fields.enums import IndicoEnumRadioField
 from indico.web.forms.fields.principals import PrincipalListField
 from indico.web.forms.validators import HiddenUnless, UsedIf
@@ -163,12 +164,14 @@ class MenuUserEntryFormBase(IndicoForm):
     is_enabled = BooleanField(_('Show'), widget=SwitchWidget())
     new_tab = BooleanField(_('Open in a new tab'), widget=SwitchWidget())
-    acl_entries = PrincipalListField(_('Visibility'), event=lambda form: form.event, allow_groups=True, allow_event_roles=True, allow_category_roles=True, allow_registration_forms=True, allow_speakers=True)
-    protection_mode = IndicoProtectionField(_('Protection mode'), protected_object=lambda form: form.protected_object)
+    # acl_entries = PrincipalListField(_('ACL'), event=lambda form: form.event, allow_groups=True,
+    #                                  allow_event_roles=True, allow_category_roles=True, allow_registration_forms=True,
+    #                                  allow_speakers=True)
+    # protection_mode = IndicoProtectionField(_('Protection mode'), protected_object=lambda form: form.protected_object)

     def __init__(self, *args, event=None, **kwargs):
         self.event = event
-        self.protected_object = kwargs.get('entry', None)
+        # self.protected_object = kwargs.get('entry', None)


@@ -179,10 +182,21 @@ class MenuLinkForm(MenuUserEntryFormBase):
 class MenuPageForm(MenuUserEntryFormBase):
     html = TextAreaField(_('Content'), [DataRequired()], widget=CKEditorWidget(images=True, html_embed=True))

+    protected = BooleanField(_('Protected'), widget=SwitchWidget())
+    acl = AccessControlListField(_('Access control list'), [UsedIf(lambda form, field: form.protected.data)],
+                                 allow_groups=True, allow_external_users=True, allow_event_roles=True,
+                                 allow_category_roles=True, allow_registration_forms=True,
+                                 event=lambda form: form.event,
+                                 description=_('The list of users and groups allowed to access the page'))
+
     def __init__(self, *args, ckeditor_upload_url, **kwargs):
         self.ckeditor_upload_url = ckeditor_upload_url

+    @generated_data
+    def protection_mode(self):
+        return ProtectionMode.protected if self.protected.data else ProtectionMode.inheriting
+

 class AddImagesForm(IndicoForm):
     image = FileField('Image', multiple_files=True, accepted_file_types='image/jpeg,image/jpg,image/png,image/gif')
```


---


however, i played around a bit and this would work with the other widget. but as you can see there's much more complexity...

```patch
diff --git a/indico/modules/events/layout/forms.py b/indico/modules/events/layout/forms.py
index e75f675c9e..7cc4e5167e 100644
--- a/indico/modules/events/layout/forms.py
+++ b/indico/modules/events/layout/forms.py
@@ -12,6 +12,7 @@ from wtforms.validators import DataRequired, Optional, ValidationError
 from indico.core.config import config
 from indico.core.db.sqlalchemy.protection import ProtectionMode
 from indico.modules.events.layout import theme_settings
+from indico.modules.events.layout.models.menu import MenuEntry
 from indico.modules.events.layout.util import get_css_file_data, get_logo_data, get_plugin_conference_themes
 from indico.modules.users import NameFormat
 from indico.util.i18n import _, orig_string
@@ -163,13 +164,6 @@ class MenuUserEntryFormBase(IndicoForm):
     is_enabled = BooleanField(_('Show'), widget=SwitchWidget())
     new_tab = BooleanField(_('Open in a new tab'), widget=SwitchWidget())
-    acl_entries = PrincipalListField(_('Visibility'), event=lambda form: form.event, allow_groups=True, allow_event_roles=True, allow_category_roles=True, allow_registration_forms=True, allow_speakers=True)
-    protection_mode = IndicoProtectionField(_('Protection mode'), protected_object=lambda form: form.protected_object)
-
-    def __init__(self, *args, event=None, **kwargs):
-        self.event = event
-        self.protected_object = kwargs.get('entry', None)
-        super().__init__(*args, **kwargs)
 
 
 class MenuLinkForm(MenuUserEntryFormBase):
@@ -177,10 +171,15 @@ class MenuLinkForm(MenuUserEntryFormBase):
 
 
 class MenuPageForm(MenuUserEntryFormBase):
+    acl = PrincipalListField(_('Visibility'), event=lambda form: form.event, allow_groups=True, allow_event_roles=True,
+                             allow_category_roles=True, allow_registration_forms=True, allow_speakers=True)
+    protection_mode = IndicoProtectionField(_('Protection mode'), protected_object=lambda form: form.protected_object)
     html = TextAreaField(_('Content'), [DataRequired()], widget=CKEditorWidget(images=True, html_embed=True))
 
-    def __init__(self, *args, ckeditor_upload_url, **kwargs):
+    def __init__(self, *args, ckeditor_upload_url, event, **kwargs):
         self.ckeditor_upload_url = ckeditor_upload_url
+        self.event = event
+        self.protected_object = kwargs.get('entry', MenuEntry(event=event))
 
 
diff --git a/indico/modules/events/layout/models/menu.py b/indico/modules/events/layout/models/menu.py
index e6e05b504f..e917c4398a 100644
--- a/indico/modules/events/layout/models/menu.py
+++ b/indico/modules/events/layout/models/menu.py
@@ -43,7 +43,7 @@ class MenuEntryType(RichIntEnum):
 # TODO move protection mixin here?
 class MenuEntryMixin:
     def __init__(self, **kwargs):
-        event = kwargs.pop('event', kwargs.get('event'))
+        event = kwargs.pop('event', None)
         super().__init__(**kwargs)
         # XXX: not calling this `event` since this one should NOT use
         # the relationship to avoid mixing data from different DB sessions
@@ -317,10 +317,6 @@ class MenuEntry(MenuEntryMixin, ProtectionMixin, db.Model):
         nullable=False
     )
 
-
-    
-
-
     #: The Event containing the menu entry
     event = db.relationship(
         'Event',
@@ -418,10 +414,7 @@ class MenuEntry(MenuEntryMixin, ProtectionMixin, db.Model):
 
     @property
     def protection_parent(self):
-        if self.parent_id is None:
-            return self.event
-        else:
-            return self.parent
+        return self.parent or getattr(self, '_event_ref', self.event)
 
 
 class EventPage(db.Model):
diff --git a/indico/web/forms/fields/protection.py b/indico/web/forms/fields/protection.py
index 411f8425ba..d857f8d42d 100644
--- a/indico/web/forms/fields/protection.py
+++ b/indico/web/forms/fields/protection.py
@@ -26,9 +26,10 @@ class IndicoProtectionField(IndicoEnumRadioField):
         self.can_inherit_protection = self.protected_object.protection_parent is not None
         self.is_unlisted_event = isinstance(self.protected_object, db.m.Event) and self.protected_object.is_unlisted
         # TODO is this just legacy and needs to be merged?
-        #if not self.can_inherit_protection and not self.is_unlisted_event:
-        #    kwargs['skip'] = {ProtectionMode.inheriting}
-        super().__init__(*args, enum=ProtectionMode, skip=self.protected_object.disallowed_protection_modes, **kwargs)
+        skip = set(self.protected_object.disallowed_protection_modes)
+        if not self.can_inherit_protection and not self.is_unlisted_event:
+            skip.add(ProtectionMode.inheriting)
+        super().__init__(*args, enum=ProtectionMode, skip=skip, **kwargs)
 
     def render_protection_message(self):
         protected_object = self.get_form().protected_object
```

kewisch: Thank you, applying your patch it magically works for me now as well! What I like about the full protection field is that it has explicit information on how it is accessible when inheriting is selected, I think that is worth the extra complexity.

indico/web/forms/fields/protection.py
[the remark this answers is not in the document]
kewisch: I'd argue that this should rather be in the protection mixin, maybe with a property `effective_disallowed_protection_modes`, or renaming the property already defined on the class to `_disallowed_protection_modes` . IIUC when an event is unlisted, it has no protection parent, so the extra check on it being an unlisted even is redundant and this should work:

```python

@property
def effective_disallowed_protection_modes(self):
    if self.protection_parent:
        return self.disallowed_protection_modes
    else:
        return self.disallowed_protection_modes - {ProtectionMode.inheriting}
```

I imagine any additional use of `disallowed_protection_modes` would need to reproduce the parent check logic, but given it is the only use right now I don't need to change it in this PR. Let me know what you prefer.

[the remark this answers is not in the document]
ThiefMaster: i'd keep it simple in the PR. refactoring and features in the same PR usually make things harder to review

kewisch:
@ThiefMaster in exploring how where to add the "Speakers" option in the UI, I'm wondering if it makes sense to expose all of the built-in event roles in the `Event Roles` dropdown? This would add "Speaker", "Author", "Convener", and maybe "Chairperson".

I could do that as a special case where the built-in event roles are just added to the dropdown next to the custom roles in the UI, or I could make some (to be explored) changes to the principal to accept special values for the builtin event roles and also retrieve them from the API. 

What do you think?

ThiefMaster:
TBH I'd keep this PR simple and first of all add it with a separate form field... and then consider improvements in a separate PR.

"Event Roles" are user-defined roles (basically event-scoped groups) so I would *not* mix those.


FWIW for improvements I'd consider UI similar to what we have on the event protection page and then maybe have grayed-out lines for special things like speakers, which would become active when you tick them to enable them. But that'd be more of a bigger task to convert all the ACL related UI fields to a nice react-based widget.

kewisch:
I came to this conclusion since on the "Participant Roles" view in the event there is a mix of the custom event roles and the built-in roles like speaker and author. So there is already a mix going on there? While they are technically separate, from a product point of view they are very much the same.

If we don't want to do a major revamp now, what I'd recommend is:
* Have an allow_speaker option in the protection field
* If this field is enabled, then when building the Event Roles dropdown, add a built-in static row for speakers
* If the user selects this special event role, it gets added to the top in similar styling as the event roles ( box with SPK in it, same color as in participant roles view )
* When the data for the widget is serialized, it sees that special event role and converts it to a separate toggle, like this: `{ 'flags': { 'speakers': true }, 'principals': [...serialized principals list that was there before...] }`
* Server side separates flags and principals and updates the two database fields accordingly.

If you'd like to go with a completely separate `BooleanField` toggle in this PR that works for me, but I would really like to integrate it into the protection widget because it seems more elegant, so I would tackle that as the very next thing. I'd love to do something that isn't as elaborate as completely revamping the ACL UI, but still would integrate speakers into the widget we have.

ThiefMaster:
I think the participant roles is special since this isn't really access control related. I'm very much against mixing event roles and other things in the same dropdown.

So in this PR I'd go for the completely separate toggle.. There's a good chance I'll merge this PR only after branching off to start development for 3.3 in `master`, so merging it into master in this "kind of intermediate" (yet stable) state wouldn't be a problem at all - and it would avoid a harder-to-review PR!

kewisch:
Ok, very well. I did notice after commenting the fancy boxes and icons from the event protection widget aren't in this one, so it would indeed make them hard to differentiate. I should be able to complete this in the next few days, all I have left is the speaker toggle, testing the data migration more elaborately, and some tests.

kewisch:
@ThiefMaster this one should be ready for review now. Do we do unit tests for migrations in some way? The downgrade seems a bit more complex. Also if you could provide some guidance on the exporters, I really don't know what I am doing there and if it is correct.

indico/migrations/versions/20230306_0905_e47fc6634291_add_menu_entry_acls.py
```diff
@@ -0,0 +1,352 @@
...
+        -----------------+-----------------+---------------------------++------------------------
+          inheriting     | false           | doesn't contain regform   ||  everyone
+                         |                 | principals                ||
+        -----------------+-----------------+---------------------------++------------------------
+          all other combinations                                       ||  everyone
+                                                                       ||  (no good mapping)
+        -----------------+-----------------+---------------------------++------------------------
+    """
+    conn = op.get_bind()
+    conn.execute('SET search_path TO events')
```
ThiefMaster, line 157: I'd prefer full `schema.table` names :)

ThiefMaster:
No good way to test migrations in unit tests... :/

TBH for something this complex I usually go for the easy way and don't try to do it in pure SQL but simply query it and then update the data using Python code (all the places with a `context.is_offline_mode()` check).

Anyway, I'll giving your migration code a try! If it works in pure SQL even better ;)

indico/migrations/versions/20230306_0905_e47fc6634291_add_menu_entry_acls.py
kewisch: As do I, though there was some SQL error I couldn't get past. I tried to reproduce the error now but am no longer getting it, yay! Updated the patch.

ThiefMaster (approved):

indico/modules/events/export.yaml
```diff
@@ -221,6 +221,12 @@ export:
     fks_out:
       - event_role_id
 
+  MenuEntryPrincipal:
```
ThiefMaster, line 224: > Also if you could provide some guidance on the exporters, I really don't know what I am doing there and if it is correct.

Since you asked about how the export stuff works some time ago: I'll add some comments below to explain it

indico/modules/events/export.yaml
```diff
@@ -221,6 +221,12 @@ export:
...
+    skipif: ROW.type.name not in ('user', 'event_role', 'registration_form')
```
ThiefMaster, line 225: And table row matching this condition won't be exported. This is done to get rid of anything that is not portable between instances (category roles and groups) - only events and users are included in exports.

indico/modules/events/export.yaml
```diff
@@ -221,6 +221,12 @@ export:
...
+    fks_out:
+      - registration_form_id
+      - event_role_id
```
ThiefMaster, line 228: This ensures ordering in the export file: registration forms and event roles must be imported first, because it's impossible to create the menu entry principals without those if they are referenced (they can't be set lazily later, because an event role principal cannot have null event role id)

indico/modules/events/export.yaml
```diff
@@ -391,6 +397,8 @@ export:
   # conference menu
   MenuEntry:
     skipif: ROW.type.name == 'plugin_link'
+    fks:
+      - MenuEntryPrincipal.menu_entry_id
```
ThiefMaster, line 401: This triggers the export if MenuEntryPrincipal objects if they are referenced in a MenuEntry that has been exported

indico/modules/events/export.yaml
```diff
@@ -477,6 +485,7 @@ import:
     ContributionPrincipal.user_id: skip
     AttachmentPrincipal.user_id: skip
     AttachmentFolderPrincipal.user_id: skip
+    MenuEntryPrincipal.user_id: skip
```
ThiefMaster, line 488: When the user chooses to not import missing users during the event import, then this skips the MenuEntryPrincipal object altogether (it's pointless to import a user acl entry if that user does not exist on the target indico instance)

ThiefMaster (commented):
In case you are curious about the things I changed, I split them into many individual commits with descriptive messages.

PR #5674, opened 2023-02-14
Refactor EditingRevision

duartegalvao:
This PR is an attempt at refactoring the `EditingRevision` to make it match better with the way it's presented in the UI, as well as to simplify a lot of the processes in the backend.

This code is not functional yet.


indico/modules/events/editing/client/js/editing/timeline/TimelineItem.jsx
```diff
@@ -43,24 +43,27 @@ export default function TimelineItem({block, index}) {
   const {fileTypes} = useSelector(selectors.getStaticData);
   const isLastBlock = lastTimelineBlock.id === block.id;
   const isLastValidBlock = lastValidTimelineBlock.id === block.id;
-  const [visible, setVisible] = useState(isLastValidBlock);
-  const isUndone = block.finalState.name === FinalRevisionState.undone;
+  const hasContent = block.commentHtml || !!block.files.length;
+  const isAlwaysVisible = isLastValidBlock || lastTimelineBlockWithFiles.id === block.id;
+  const [visible, setVisible] = useState(isAlwaysVisible);
 
   useEffect(() => {
```
tomasr8, line 50: Would it be simpler if instead of the useEffect, you check `isAlwaysVisible` when setting the visibility?
Something like this:
```js
onClick={() => {
  if(!visible || !isAlwaysVisible) { setVisible(v => !v) }
}}
```

duartegalvao: found an even simpler solution:
```js
const [visibilityToggle, setVisibilityToggle] = useState(false);
const visible = visibilityToggle || isAlwaysVisible;
```

indico/modules/events/editing/models/revisions.py
```diff
@@ -18,61 +17,37 @@
 from indico.util.string import format_repr
 
 
-class InitialRevisionState(RichIntEnum):
-    __titles__ = [None, _('New'), _('Ready for Review'), _('Needs Confirmation')]
-    __css_classes__ = [None, 'highlight', 'ready', 'warning']
-    #: A revision that has been submitted by the user but isn't exposed to editors yet
+class RevisionType(RichIntEnum):
+    __titles__ = [None, _('New'), _('Ready for Review'), _('Needs Confirmation'), _('Accepted'), _('Needs Changes'),
```
tomasr8, line 21: Might be worth to add context to some of those strings e.g. `pgettext('revision', 'New')`. Otherwise this is difficult to translate in gendered languages.

indico/modules/events/editing/models/revisions.py
```diff
@@ -193,3 +152,10 @@ def get_published_files(self):
             if file.file_type.publishable:
                 files[file.file_type].append(file)
         return dict(files)
+
+    @property
+    def is_editor(self):
+        if self.type in (RevisionType.needs_submitter_confirmation, RevisionType.acceptance,
+                         RevisionType.rejection, RevisionType.reset, RevisionType.needs_submitter_changes):
+            return True
+        return False
```
tomasr8, line 161:
```suggestion
        return self.type in (RevisionType.needs_submitter_confirmation, RevisionType.acceptance, RevisionType.rejection,
                             RevisionType.reset, RevisionType.needs_submitter_changes)
```

indico/modules/events/editing/models/revisions.py
duartegalvao: As far as I can tell these titles aren't ever displayed. Maybe it would be better to remove them altogether
(although for `EditableState` it might be worth adding context, yeah)

duartegalvao: for the editable strings, would it be ok to do something like
```python
def _(s):
    return pgettext('editable', s)
```
(just asking because i dont see that anywhere else in the code and it would get super wordy otherwise)

ThiefMaster: no, that would break i18n string extraction

indico/migrations/versions/20230810_1300_0af8f63aa603_refactor_editing_revisions.py
```diff
@@ -0,0 +1,308 @@
...
+                                b.final_state = {_FinalRevisionState.needs_submitter_changes} AND
+                                a.editor_id IS NOT NULL AND
+                                a.editor_id = b.editor_id AND
+                                a.reviewed_dt IS NOT NULL AND
+                                a.reviewed_dt = b.reviewed_dt)
+    '''
+    new_revisions = []
+    for editable_id, submitter_id, editor_id, reviewed_dt, final_state, comment in conn.execute(query):
+        revision_type = {
+            _FinalRevisionState.needs_submitter_changes: _RevisionType.changes_rejection,
```
duartegalvao, line 67: I believe this is what's causing the "submitter rejected proposed changes" bug. I think this should fix it:
```suggestion
            _FinalRevisionState.needs_submitter_changes: _RevisionType.needs_submitter_changes if editor_id else _RevisionType.changes_rejection,
```

ThiefMaster: You probably want this since technically the ID can be `0`. This is incredibly unlikely since only legacy indico used 0 as a valid ID, but still ;)

```
            _FinalRevisionState.needs_submitter_changes: (_RevisionType.needs_submitter_changes
                                                          if editor_id is not None
                                                          else _RevisionType.changes_rejection),
```

ThiefMaster: @duartegalvao will you commit the fix?

duartegalvao: sure, in just a sec

indico/modules/events/editing/notifications.py
```diff
@@ -23,19 +23,20 @@ def notify_comment(comment):
     revision = comment.revision
     editable = revision.editable
     editor = editable.editor
+    submitter = editable.revisions[0].user
```
ThiefMaster, line 26: Is this logic correct? Take an editable for a contribution with two contributors, A and B (both submitters/authors on the contribution). A submitted the initial editable. At a later stage, a revision was submitted by B. Now someone comments on that revision. If I understand the logic correctly, B won't get a notification for that comment...

indico/migrations/versions/20230810_1300_0af8f63aa603_refactor_editing_revisions.py
```diff
@@ -240,7 +249,13 @@ def upgrade():
             INSERT INTO event_editing.revisions
             (editable_id, user_id, created_dt, type, is_undone, comment) VALUES
             (%s,          %s,      %s,         %s,   %s,        %s)
-        ''', rev)
+        ''', rev[:-1])
+        for comment in rev[-1]:
+            conn.execute('''
+                UPDATE event_editing.comments
+                SET revision_id = (SELECT id FROM event_editing.revisions ORDER BY id DESC LIMIT 1)
```
ThiefMaster, line 256: That looks incredibly fragile considering the subquery has no correlation with the revision/editable whatsoever...

ThiefMaster (approved):

PR #5677, opened 2023-02-16
Add an option to sync event persons with users

tomasr8:
Adds a new sync button to the Participant roles page, which syncs the event person's personal data (name, title, ..) with its parent user.

### DIFF
--- a/indico/modules/events/persons/blueprint.py
+++ b/indico/modules/events/persons/blueprint.py
@@ -9,7 +9,7 @@ from indico.modules.events.persons.controllers import (RHAPIEmailEventPersonsMet
                                                        RHDeleteUnusedEventPerson, RHEmailEventPersonsPreview,
                                                        RHEventPersonSearch, RHGrantModificationRights,
                                                        RHGrantSubmissionRights, RHManagePersonLists, RHPersonsList,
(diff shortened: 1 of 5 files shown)


indico/modules/events/persons/blueprint.py
```diff
@@ -32,6 +32,7 @@
 
 # EventPerson operations
 _bp.add_url_rule('/persons/<int:person_id>', 'update_person', RHUpdateEventPerson, methods=('PATCH',))
+_bp.add_url_rule('/persons/<int:person_id>', 'sync_with_user', RHSyncEventPerson, methods=('POST',))
```
ThiefMaster, line 35:
```suggestion
_bp.add_url_rule('/persons/<int:person_id>/sync', 'sync_with_user', RHSyncEventPerson, methods=('POST',))
```

POST on the resource url directly is not very clear (in fact such a POST is usually to replace the whole thing) - so having some custom verb in the URL is nicer

ThiefMaster (approved):

indico/modules/events/persons/blueprint.py
tomasr8: ahh yeah, missed that.. :+1:

PR #5682, opened 2023-02-18
Allow managers to ignore required field restrictions in registrations

### LINKED ISSUE #5644: Allow managers to modify a registration despite form field requirements
We have a set of internal events where the attendees are known from the start. We need to maintain a few fields internally on these attendees and would like to leverage Indico for that instead of yet another spreadsheet. Attendees are also expected to update a few fields (e.g. travel information)

So we've created a registration form with the fields, and have imported all attendees into the registrations. There are a few required fields. When emailing them for the first time, we just use {link} to let them know, and expect them to fill in all they know. 

So now we've run into an issue:
* As a registration manager, I have no way to modify the registration to flip those internal fields, or fill in partial information. 
* The form complains that the required fields are not filled out.

I'd appreciate a toggle that would allow a registration manager, event organizer, or admin, to override the required fields. They would stay in a state that they are not filled out, but I can still change other registration fields.

* Mark fields not required, make changes, mark them required again
* Mark them as optional, then nag people to fill in the missing fields

Would something like that work out? It doesn't seem too complicated, but maybe I am missing some constraints.

kewisch:
Fixes #5644.

I've decided to leave out the checkbox and have this happen implicitly. This feels more natural and requires one less thing to click on.

### DIFF
--- a/indico/modules/events/registration/client/js/form/FormItem.jsx
+++ b/indico/modules/events/registration/client/js/form/FormItem.jsx
@@ -96,8 +96,8 @@ export default function FormItem({
     );
   }
 
-  const required = meta.alwaysRequired || isRequired;
-  const managementRequired = meta.alwaysRequired || (!isManagement && isRequired);
+  const showAsRequired = meta.alwaysRequired || isRequired;
+  const inputRequired = !isManagement && showAsRequired;
   return (
     <div
       styleName={`form-item ${toClasses({
@@ -112,19 +112,19 @@ export default function FormItem({
         {InputComponent ? (
           meta.customFormItem ? (
             <InputComponent
-              showAsRequired={required}
-              isRequired={managementRequired}
+              showAsRequired={showAsRequired}
+              isRequired={inputRequired}
               disabled={disabled}
               isPurged={showPurged}
               retentionPeriodIcon={retentionPeriodIcon}
               {...inputProps}
             />
           ) : (
-            <Form.Field required={required} styleName="field">
+            <Form.Field required={showAsRequired} styleName="field">
               <label style={{opacity: disabled ? 0.8 : 1, display: 'inline-block'}}>{title}</label>
               {retentionPeriodIcon}
               <InputComponent
-                isRequired={managementRequired}
+                isRequired={inputRequired}
                 disabled={disabled}
                 isPurged={showPurged}
                 {...inputProps}
--- a/indico/modules/events/registration/client/js/form/FormItem.jsx
+++ b/indico/modules/events/registration/client/js/form/FormItem.jsx
@@ -112,6 +112,7 @@ export default function FormItem({
         {InputComponent ? (
           meta.customFormItem ? (
             <InputComponent
+              showAsRequired={required}
               isRequired={managementRequired}
               disabled={disabled}
               isPurged={showPurged}
--- a/indico/modules/events/registration/client/js/form/FormItem.jsx
+++ b/indico/modules/events/registration/client/js/form/FormItem.jsx
@@ -97,6 +97,7 @@ export default function FormItem({
   }
 
   const required = meta.alwaysRequired || isRequired;
+  const managementRequired = meta.alwaysRequired || (!isManagement && isRequired);
   return (
     <div
       styleName={`form-item ${toClasses({
@@ -111,7 +112,7 @@ export default function FormItem({
         {InputComponent ? (
           meta.customFormItem ? (
             <InputComponent
-              isRequired={required}
+              isRequired={managementRequired}
               disabled={disabled}
               isPurged={showPurged}
               retentionPeriodIcon={retentionPeriodIcon}
@@ -122,7 +123,7 @@ export default function FormItem({
               <label style={{opacity: disabled ? 0.8 : 1, display: 'inline-block'}}>{title}</label>
               {retentionPeriodIcon}
               <InputComponent
-                isRequired={required}
+                isRequired={managementRequired}
                 disabled={disabled}
                 isPurged={showPurged}
                 {...inputProps}
--- a/indico/modules/events/registration/controllers/__init__.py
+++ b/indico/modules/events/registration/controllers/__init__.py
@@ -54,7 +54,8 @@ class RegistrationEditMixin:
         schema = make_registration_schema(
             self.regform,
             management=self.management,
-            registration=self.registration
+            registration=self.registration,
+            override_required=self.management
         )(partial=optional_fields)
         form_data = parser.parse(schema)
 
--- a/indico/modules/events/registration/controllers/__init__.py
+++ b/indico/modules/events/registration/controllers/__init__.py
@@ -41,6 +41,10 @@ class RegistrationEditMixin:
 
     def _get_optional_fields(self):
         """Get fields for which we already have a value and thus are not required."""
+        if self.management:
+            # In management mode, all fields are optional
+            return True
+
         data_by_field = self.registration.data_by_field
         return [form_item.html_field_name for form_item in self.regform.active_fields
                 if data_by_field.get(form_item.id) is not None]
@@ -48,7 +52,9 @@ class RegistrationEditMixin:
     def _process_POST(self):
         optional_fields = self._get_optional_fields()
         schema = make_registration_schema(
-            self.regform, management=self.management, registration=self.registration
+            self.regform,
+            management=self.management,
+            registration=self.registration
         )(partial=optional_fields)
         form_data = parser.parse(schema)
 
--- a/indico/modules/events/registration/controllers/display_test.py
+++ b/indico/modules/events/registration/controllers/display_test.py
@@ -7,12 +7,17 @@
 
 import pytest
 from flask import request, session
-from werkzeug.exceptions import Forbidden
+from werkzeug.exceptions import Forbidden, UnprocessableEntity
 
-from indico.modules.events.registration.controllers.display import RHRegistrationForm, RHRegistrationWithdraw
+from indico.modules.events.registration.controllers.display import (RHRegistrationDisplayEdit, RHRegistrationForm,
+                                                                    RHRegistrationWithdraw)
+from indico.modules.events.registration.controllers.management.fields import _fill_form_field_with_data
+from indico.modules.events.registration.models.form_fields import RegistrationFormField
 from indico.modules.events.registration.models.forms import ModificationMode
 from indico.modules.events.registration.models.invitations import RegistrationInvitation
+from indico.modules.events.registration.models.items import RegistrationFormSection
 from indico.modules.events.registration.models.registrations import RegistrationState
+from indico.modules.events.registration.util import create_registration
 from indico.testing.util import extract_emails
 from indico.util.date_time import now_utc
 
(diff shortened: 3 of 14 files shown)


kewisch:
Please check #5644 before merging, I have a question there and I'd like to extend this PR accordingly.

kewisch:
@ThiefMaster this should be ready to review now. I've not added any tests yet though they would certainly be useful given make_registration_schema is used in a few places. I may get to that next weekend, but I wanted to see how you feel about the remaining changes.

I'm also not sure how to deal with the translation changes. I've run the i18n script and it changed a bunch of line numbers in unrelated strings, so I'm guessing this is usually run separately?

ThiefMaster:
You're right - so please don't update i18n pot files in PRs - we run the extract (and upload to transifex) from time to time; in PRs it'd just add noise and make conflicts very likely as well.

kewisch:
Apologies for the many pushes, had some issues in my local repo that made the tests act differently. Should be good now!

indico/modules/events/registration/controllers/__init__.py
```diff
@@ -41,14 +41,19 @@ def _get_file_data(self):
...
+            # In management mode, all fields are optional
+            return True
+
         data_by_field = self.registration.data_by_field
         return [form_item.html_field_name for form_item in self.regform.active_fields
                 if data_by_field.get(form_item.id) is not None]
 
     def _process_POST(self):
         optional_fields = self._get_optional_fields()
+        overrideRequired = request.json.get('override_required', False)
```
ThiefMaster, line 54: no camelcase in python code please!

indico/modules/events/registration/fields/base.py
```diff
@@ -215,5 +230,5 @@ class InvalidRegistrationFormField(RegistrationFormFieldBase):
 
     is_invalid_field = True
 
-    def create_mm_field(self, registration=None):
+    def create_mm_field(self, registration=None, overrideRequired=False):
```
ThiefMaster, line 233: no camelcase please

kewisch:
@ThiefMaster finally got around to fixing the review comments, should be good now. Maybe worth thinking about some linting plugin to enforce it, like `pep8-naming` ?

ThiefMaster:
Oh wow, I didn't know of that plugin. I'm going to look into adding that tomorrow for sure! May have to disable some checks because of too many false positives that cannot be easily ignored, but the general check for camelCase is already nice!

ThiefMaster:
#5714 - there we go. No more work for me to point out camelCase and similar issues... ;)

indico/modules/events/registration/client/js/form/FormItem.jsx
```diff
@@ -97,6 +97,7 @@ export default function FormItem({
   }
 
   const required = meta.alwaysRequired || isRequired;
+  const managementRequired = !isManagement && (meta.alwaysRequired || isRequired);
```
tomasr8, line 100:
```suggestion
  const managementRequired = !isManagement && required;
```

indico/modules/events/registration/controllers/__init__.py
```diff
@@ -41,14 +41,22 @@ def _get_file_data(self):
...
+            # In management mode, all fields are optional
+            return True
+
         data_by_field = self.registration.data_by_field
         return [form_item.html_field_name for form_item in self.regform.active_fields
                 if data_by_field.get(form_item.id) is not None]
 
     def _process_POST(self):
         optional_fields = self._get_optional_fields()
+        override_required = request.json.get('override_required', False)
```
tomasr8, line 54: Do we need this for modifications? We only send the field when creating the registration

ThiefMaster:
Do you need this very soon yourself? Otherwise I might take this PR also as something for v3.3

kewisch:
When are you expecting 3.3? We've been working around by just not making fields required, so I suspect it could go into 3.3. I'll check with the team that requested this.

indico/modules/events/registration/controllers/__init__.py
kewisch: Yes I think we can do without here.

indico/modules/events/registration/controllers/management/reglists_test.py
```diff
@@ -0,0 +1,102 @@
...
+from indico.modules.events.registration.models.items import RegistrationFormSection
+from indico.modules.events.registration.util import create_registration
+from indico.util.date_time import now_utc
+
+
+pytest_plugins = 'indico.modules.events.registration.testing.fixtures'
+
+
+def test_registration_edit_override_required(
+    db, dummy_regform, dummy_user, app_context, create_user
```
ThiefMaster, line 24:
```suggestion
    db, dummy_regform, app_context, create_user
```

This fixture is unused

indico/modules/events/registration/controllers/management/reglists_test.py
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        rh = RHRegistrationEdit()
+        rh._process_args()
+        rh._process_POST()
+
+
+def test_registration_create_override_required(
+    db, dummy_regform, dummy_user, app_context, create_user
+):
```
ThiefMaster, line 65:
```suggestion
def test_registration_create_override_required(db, dummy_regform, app_context):
```

Same here

indico/modules/events/registration/controllers/management/reglists_test.py
```diff
@@ -0,0 +1,102 @@
...
+    # Add a section and a required field
+    section = RegistrationFormSection(
+        registration_form=dummy_regform, title='dummy_section', is_manager_only=False
+    )
+    db.session.add(section)
```
ThiefMaster, line 70: not needed

indico/modules/events/registration/controllers/management/reglists_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def test_registration_edit_override_required(
+    db, dummy_regform, dummy_user, app_context, create_user
+):
+    # Add a section and a required field
+    section = RegistrationFormSection(
+        registration_form=dummy_regform, title='dummy_section', is_manager_only=False
+    )
+    db.session.add(section)
```
ThiefMaster, line 30: not needed

indico/modules/events/registration/controllers/management/reglists_test.py
```diff
@@ -0,0 +1,102 @@
...
+    db.session.flush()
+
+    boolean_field = RegistrationFormField(parent_id=section.id, registration_form=dummy_regform)
+    _fill_form_field_with_data(
+        boolean_field, {'input_type': 'bool', 'is_required': True, 'title': 'Yes/No'}
+    )
+
+    # Register a new user
+    other_user = create_user(1, first_name='first', last_name='last', email='email@example.com')
+    dummy_regform.start_dt = now_utc(False)
```
ThiefMaster, line 40: not needed

indico/modules/events/registration/controllers/management/reglists_test.py
```diff
@@ -0,0 +1,102 @@
...
+import pytest
+from flask import request
+from werkzeug.exceptions import UnprocessableEntity
+
+from indico.modules.events.registration.controllers.management.fields import _fill_form_field_with_data
+from indico.modules.events.registration.controllers.management.reglists import RHRegistrationCreate, RHRegistrationEdit
+from indico.modules.events.registration.models.form_fields import RegistrationFormField
+from indico.modules.events.registration.models.items import RegistrationFormSection
+from indico.modules.events.registration.util import create_registration
+from indico.util.date_time import now_utc
```
ThiefMaster, line 17: unused when removing the unnecessary line below ;)

indico/modules/events/registration/util.py
```diff
@@ -286,8 +286,21 @@ class Meta:
         unknown = RAISE
 
 
-def make_registration_schema(regform, management=False, registration=None, captcha_required=False):
-    """Dynamically create a Marshmallow schema based on the registration form fields."""
+def make_registration_schema(
+    regform,
```
ThiefMaster, line 290:
```suggestion
    regform,
    *,
```

Good opportunity to make those args keyword-only (we already always pass them as keywords)

indico/modules/events/registration/fields/base.py
```diff
@@ -98,12 +98,27 @@ def create_mm_field(self, registration=None):
...
+        if self.form_item.is_required and self.not_empty_if_required and not skip_required:
             validators.append(not_empty)
+
+        mm_field_kwargs = self.mm_field_kwargs
+        if skip_required:
+            mm_field_kwargs['allow_none'] = True
+
         return self.mm_field_class(*self.mm_field_args,
-                                   required=self.form_item.is_required,
+                                   required=self.form_item.is_required and not skip_required,
```
ThiefMaster, line 119:
```suggestion
                                   required=(self.form_item.is_required and not skip_required),
```

Just for readability when kwarg values contain spaces

indico/modules/events/registration/client/js/form/FormItem.jsx
```diff
@@ -97,6 +97,7 @@ export default function FormItem({
   }
 
   const required = meta.alwaysRequired || isRequired;
+  const managementRequired = !isManagement && required;
```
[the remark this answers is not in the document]
ThiefMaster: When testing I didn't notice any problems since we always preselect the "no accommodation" option, but if the accommodation field does not have that option (you can disable it if the user MUST check one), then it currently fails silently regardless of the override toggle.

ThiefMaster (commented):
Really nice PR and works well. Most of the suggestions are just stylistic :)

ThiefMaster:
> When are you expecting 3.3? We've been working around by just not making fields required, so I suspect it could go into 3.3. I'll check with the team that requested this.

Considering that the PR is really high quality (thanks for the tests!) I feel comfortable merging this into 3.2.4 before branching out for 3.3.

indico/modules/events/registration/controllers/display_test.py
```diff
@@ -93,3 +98,70 @@ def test_withdraw_registration_rh(smtp, dummy_regform, dummy_reg, dummy_user):
...
+
+
+def test_display_edit_override_required_rh(
+    db, dummy_regform, dummy_user, app_context, smtp
+):
+    # Add a section and a required field
+    section = RegistrationFormSection(
+        registration_form=dummy_regform, title='dummy_section', is_manager_only=False
+    )
+    db.session.add(section)
```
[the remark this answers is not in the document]
kewisch: Sounds good. I guess I don't need to flush then either?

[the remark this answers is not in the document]
ThiefMaster: usually flushing isn't needed (sqlalchemy has autoflush which flushes when it thing it's necessary - usually it works fine). so i'd thry removing it and see if the test still passes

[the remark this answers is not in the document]
kewisch: Sounds good, thanks! Removed the flush here, though needed to keep it in one location in reglist_test.

indico/modules/events/registration/client/js/form/FormItem.jsx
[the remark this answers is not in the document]
kewisch: For the sake of getting this PR in I'll go with the fix you suggested, but I do feel that event managers need to be able to make changes to a registration without affecting any required fields. 

If the event manager is required to fill out the accommodation field, we're missing the state "registrant has not filled this out yet", which is available for all other fields. 

If the manager is required to select "No accommodation", this will make it harder to differentiate between attendees that have deliberately decided against accommodation vs those who need more nagging to fill out the registration field.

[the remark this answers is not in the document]
ThiefMaster: True, that's a good point I didn't think about. I'll have a look before merging this if I can find a nice way to solve it.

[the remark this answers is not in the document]
ThiefMaster: (fyi i just marked this as unresolved so I don't have to dig through the comments to find these two changes tomorrow)

PR #5686, opened 2023-02-21
Use the revision's timestamp on ZIP archives

duartegalvao:
Use the revision's timestamp on ZIP archives' files instead of the datetime at the time of the download.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,7 @@ Version 3.2.4
 Improvements
 ^^^^^^^^^^^^
 
-- None so far :(
+- Use the revision's timestamp when downloading its files as a ZIP archive (:pr:`5686`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/events/editing/controllers/backend/timeline.py
+++ b/indico/modules/events/editing/controllers/backend/timeline.py
@@ -6,6 +6,7 @@
 # LICENSE file for more details.
 
 import os
+import shutil
 from io import BytesIO
 from zipfile import ZipFile, ZipInfo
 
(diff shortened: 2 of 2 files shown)


indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -388,7 +388,11 @@ def _process(self):
                 folder_name = secure_filename(file_type.name, f'file-type-{file_type.id}')
 
                 with file.storage.get_local_path(file.storage_file_id) as filepath:
-                    zip_handler.write(filepath, os.path.join(folder_name, filename))
+                    info = ZipInfo.from_file(filepath)
+                    info.filename = os.path.join(folder_name, filename)
```
ThiefMaster, line 392: Isn't this already already set in the ZipInfo constructor?

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -388,7 +388,11 @@ def _process(self):
...
+                    info.date_time = self.revision.created_dt.astimezone(self.event.tzinfo).timetuple()[:6]
+                    with open(filepath, 'rb') as fd:
+                        zip_handler.writestr(info, fd.read())
```
ThiefMaster, line 395: Isn't this code (based on the `zip_handler.write()` code better to avoid reading the whole file into memory?

            with open(filename, 'rb') as src, zip_handler.open(info, 'w') as dest:
                shutil.copyfileobj(src, dest, 1024*8)

In fact you can probably use `.open()` instaed of `.get_local_path()` to get the `src` file object

ThiefMaster:
FYI I rebased the PR and slightly reworded the changelog

indico/modules/events/editing/controllers/backend/timeline.py
duartegalvao: The problem is I'm getting the ZipInfo object from the `from_file` util, which automatically grabs all info such as permissions, but then I have to set the filename we want for the file in the zip, otherwise it will have a path like `<INDICO DIR>/indico/data/archive/event/<EVENT>/editing/...` inside the zip...

Would using the ZipInfo constructor be a better option here?

ThiefMaster: OK this makes perfect sense, there's in deed nothing else that sets the file name used inside the ZIP.

However, I now realize that passing `src.name` above probably won't always work. This may be AND kind of file object, which doesn't necessarily have a name or physical filesystem location (in case of a file stored on S3 for example).

duartegalvao: Hmm in this case it might be better to use the ZipInfo constructor then. I think there won't be any problems if we always set the permissions to be `644`, right?

ThiefMaster: that should be fine

ThiefMaster (approved):

CHANGES.rst
```diff
@@ -81,6 +81,7 @@ Improvements
   been accepted (:issue:`5666`, :pr:`5671`)
 - Prevent concurrent assignment of editors to editables (:pr:`5684`)
 - Add color labels to the filter dropdown (:issue:`5675`, :pr:`5680`)
+- Use the revision's timestamp when downloading its files as a ZIP archive (:pr:`5686`)
```
ThiefMaster, line 84: Please move this up to the 3.2.4 section

PR #5700, opened 2023-03-03
Add hook below search

openprojects:
Add hook to inject React component below the search box.

### DIFF
--- a/indico/modules/search/client/js/components/SearchApp.jsx
+++ b/indico/modules/search/client/js/components/SearchApp.jsx
@@ -298,7 +298,7 @@ export default function SearchApp({category, eventId, isAdmin}) {
           searchTerm={q || ''}
           placeholders={eventSearch ? [] : options?.placeholders || []}
         />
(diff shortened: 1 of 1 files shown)


indico/modules/search/client/js/components/SearchApp.jsx
```diff
@@ -298,6 +298,7 @@ export default function SearchApp({category, eventId, isAdmin}) {
           searchTerm={q || ''}
           placeholders={eventSearch ? [] : options?.placeholders || []}
         />
+        {renderPluginComponents('search-app-below-input', {user: Indico.User})}
```
ThiefMaster, line 301: You do not need to pass the user here, you can use it directly in your plugin.
Also, I suggest passing the props such as category, eventId and possibly isAdmin instead since especially the first two would be much harder to get from plugin code (even if you do not need those).

openprojects: Fixed.

PR #5706, opened 2023-03-16
Align status box colors with judgment dropdown

### LINKED ISSUE #5699: Match status box colors with judgement dropdown colors
As discussed, the status box colors could be tweaked to match the color dots in the judgement dropdown menu.

image

GovernmentPlates:
Closes #5699

### DIFF
--- a/indico/web/client/styles/base/_theme.scss
+++ b/indico/web/client/styles/base/_theme.scss
@@ -12,9 +12,25 @@ $semantic-color-map: (
   warning: $yellow,
   visited: $purple,
   strong: $black,
+  editing-rejected: $midnight-black,
+  editing-make-changes: $bright-yellow,
+  editing-request-changes: $dark-red,
 );
 
-$semantic-styles: (highlight, success, error, warning, visited, strong);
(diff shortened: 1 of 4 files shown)


indico/web/client/styles/base/_theme.scss
```diff
@@ -12,9 +12,22 @@ $semantic-color-map: (
   warning: $yellow,
   visited: $purple,
   strong: $black,
+  rejected: $midnight-black,
+  make-changes: $bright-yellow,
+  request-changes: $dark-red,
 );
 
-$semantic-styles: (highlight, success, error, warning, visited, strong);
+$semantic-styles: (
```
ThiefMaster, line 20: Can we avoid putting this in here? IIRC "semantic-styles" is pretty generic and meant for places where we have semantic names like "error" and "warning". But I don't think the editing-related ones belong here - they are specific to just that particular module...

If this is really needed because of the way the label is displayed, I suggest we use `editing-make-changes` etc and add a comment here telling people that it's not meant to be used elsewhere...

PR #5711, opened 2023-03-27
Allow setting primary authors as submitters

duartegalvao:
This PR adds the option of setting the abstracts' primary authors as the default submitters for the corresponding contributions. It also fixes a bug that broke the "speakers and authors" option.

image


indico/modules/events/abstracts/settings.py
```diff
@@ -89,7 +90,8 @@ class BOALinkFormat(RichEnum):
 
 SubmissionRightsType.__titles__ = {
     SubmissionRightsType.speakers: _('Speakers'),
-    SubmissionRightsType.all: _('Speakers and authors')
+    SubmissionRightsType.primary_authors: _('Primary authors'),
+    SubmissionRightsType.all: _('Speakers and all authors')
```
ThiefMaster, line 94: I think in the UI we just use "authors", don't we? So maybe "Authors" and "Speakers, authors and co-authors" would be better?

Also, do we REALLY want to exclude speakers in case of "primary authors"? It seems strange to give authors submission rights but not the actual speaker...

ThiefMaster (approved):

PR #5720, opened 2023-03-31
Fix abstract submission success message

duartegalvao:
This PR fixes a misleading abstract submission success message, which incorrectly states "You will be notified by email with the submission details." even if no notifications are sent (e.g. no notification templates are set).

### DIFF
--- a/indico/modules/events/abstracts/controllers/display.py
+++ b/indico/modules/events/abstracts/controllers/display.py
@@ -65,10 +65,15 @@ class RHSubmitAbstract(RHAbstractsBase):
         abstract_form_class = make_abstract_form(self.event, session.user, management=self.management)
         form = abstract_form_class(event=self.event)
         if form.validate_on_submit():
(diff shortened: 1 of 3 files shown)


indico/modules/events/abstracts/controllers/display.py
```diff
@@ -65,9 +65,13 @@ def _process(self):
...
         form = abstract_form_class(event=self.event)
         if form.validate_on_submit():
-            abstract = create_abstract(self.event, *get_field_values(form.data), send_notifications=True)
+            abstract, notifications_sent = create_abstract(self.event, *get_field_values(form.data),
+                                                           send_notifications=True)
             flash(_("Your abstract '{}' has been successfully submitted. It is registered with the number "
                     '#{}. You will be notified by email with the submission details.')
+                  .format(abstract.title, abstract.friendly_id)
+                  if notifications_sent
+                  else _("Your abstract '{}' has been successfully submitted. It is registered with the number #{}.")
```
ThiefMaster, line 74: That's not very readable... maybe just use two separate flash() calls?

PR #5728, opened 2023-04-06
Add support for authorized submitters in CfP

tomasr8:
We don't have a `Submission` section in CfP like we do in CfA so I added this option to the `Reviewing settings` which already has options like `Announcement` that are in submission settings in CfA.

image

### DIFF
--- a/indico/modules/events/papers/controllers/display.py
+++ b/indico/modules/events/papers/controllers/display.py
@@ -37,7 +37,7 @@ class RHSubmitPaper(RHPaperBase):
 
     def _check_access(self):
         RHPaperBase._check_access(self)
-        if not self.event.cfp.is_open:
+        if not self.event.cfp.can_submit_proceedings(session.user):
             raise NoReportError.wrap_exc(Forbidden(_('The Call for Papers is closed. '
                                                      'Please contact the event organizer for further assistance.')))
 
@@ -122,7 +122,7 @@ class RHSelectContribution(RHCallForPapers):
 
     def _check_access(self):
         RHCallForPapers._check_access(self)
-        if not self.event.cfp.is_open:
+        if not self.event.cfp.can_submit_proceedings(session.user):
             raise NoReportError.wrap_exc(Forbidden(_('The Call for Papers is closed. '
                                                      'Please contact the event organizer for further assistance.')))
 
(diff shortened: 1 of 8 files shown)


indico/modules/events/papers/templates/display/_common.html
```diff
@@ -1,14 +1,18 @@
-{% macro render_cfp_infoline(event, cfp, contributions, show_icon=true) %}
```
tomasr8, line 1: AFAICT this was copied from CfA where we show the icon where we render the infoline also on the main conference page, which we don't do with CfP. The only place where it's rendered sets `show_icon` to False so I just removed it

indico/modules/events/papers/models/call_for_papers_test.py
```diff
@@ -0,0 +1,17 @@
...
+
+import pytest
+
+from indico.modules.events.papers.settings import paper_reviewing_settings
+
+
+@pytest.mark.usefixtures('dummy_contribution')
+def test_is_authorized_submitter(dummy_user, dummy_event):
+    assert not dummy_event.cfp.is_authorized_submitter(dummy_user)
+    paper_reviewing_settings.set_multi(dummy_event, {'authorized_submitters': {dummy_user}})
```
ThiefMaster, line 16: FYI, this should work as well since it's just a single setting:

```python
paper_reviewing_settings.acls.add_principal('authorized_submitters', dummy_user)
```

But I'm fine either way

indico/modules/events/papers/models/call_for_papers_test.py
```diff
@@ -0,0 +1,17 @@
...
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+import pytest
+
+from indico.modules.events.papers.settings import paper_reviewing_settings
+
+
+@pytest.mark.usefixtures('dummy_contribution')
```
ThiefMaster, line 13: is this actually needed?

indico/modules/events/papers/controllers/display.py
```diff
@@ -37,7 +37,8 @@ def _check_paper_protection(self):
 
     def _check_access(self):
         RHPaperBase._check_access(self)
-        if not self.event.cfp.is_open:
+        cfp = self.event.cfp
+        if not cfp.is_open and not cfp.is_authorized_submitter(session.user):
```
ThiefMaster, line 41: Would it make sense to have a method in the cfp object that checks both? I see two cases here and at least one in the templates where we want to know if the user can submit papers (regardless of the reason).

tomasr8: Good point, added `can_submit_proceedings` to CfP

ThiefMaster (approved):

PR #5732, opened 2023-04-11
Add watchfiles-based reloader

ThiefMaster:
Watchman works great, but there's no proper release for Python 3.10+ and it requires installing the watchman binary.

[Watchfiles](https://watchfiles.helpmanual.io) on the other hand is much easier to install as it includes its (Rust-based) inotify backend in its wheel.

The only real drawback is that there's no magic that detects and fully groups Git changes (e.g. checkouts). But debouncing covers this just fine, and if needed the thresholds could be tuned (if some poor soul develops on a super slow machine where Git checkouts take ages).

---

I might refactor it a bit to remove the duplicate code between {watchman,watchfiles}.py but I mainly wanted to have a reloader that works well w/ Python 3.11 to make testing Python 3.11 compatibility less painful...

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,6 +44,7 @@ Internal Changes
 ^^^^^^^^^^^^^^^^
 
 - Update Python & JavaScript dependencies (:pr:`5726`)
+- Add support for the watchfiles live reloader (:pr:`5732`)
 
 
 Version 3.2.3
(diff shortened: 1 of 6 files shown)


ThiefMaster:
d69b026f24 fixes an issue during startup when not specifying the reloader but relying on the default

PR #5735, opened 2023-04-12
Include logged-in user in the custom-email logs

vasantvohra:


indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -229,7 +229,8 @@ def _send_emails(self, form):
                                    attachments=attachments)
             signals.core.before_notification_send.send('registration-custom-email', email=email,
                                                        registration=registration, form=form)
-            send_email(email, self.event, 'Registration', log_metadata={'registration_id': registration.id})
+            send_email(email, self.event, 'Registration', user=session.user,
+                       log_metadata={'registration_id': registration.id})
```
ThiefMaster, line 233: You can use `registration.log(...)` which automatically passes the `log_metadata`

vasantvohra: but this is `send_email` which uses `_log_email` that uses `event.log(...)`

ThiefMaster: oh right, nevermind

indico/modules/events/registration/controllers/management/reglists.py
```diff
@@ -229,7 +229,8 @@ def _send_emails(self, form):
                                    attachments=attachments)
             signals.core.before_notification_send.send('registration-custom-email', email=email,
                                                        registration=registration, form=form)
-            send_email(email, self.event, 'Registration', log_metadata={'registration_id': registration.id})
+            send_email(email, self.event, 'Registration', user=session.user,
```
ThiefMaster, line 232:
```suggestion
            send_email(email, self.event, 'Registration', session.user,
```

we pass it as a posarg everywhere else

vasantvohra: done

PR #5738, opened 2023-04-14
Include all authors in timetable export API

### LINKED ISSUE #5412: Include non-speaker presenters in export API
Right now only people in contribution marked as "Speaker" are in the export file. They are exported as 'presenters'.
There is a need to export also all the others with the same info, at least all grouped f.e. as 'authors' or distinguished any other way/flag.

duartegalvao:
Closes #5412

Adds an `other_authors` field to the contributions in the timetable export API, which includes all authors which are not speakers (which are already included in the `presenters` field).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,6 +42,7 @@ Improvements
   given editable type (:pr:`5755`)
 - Show only "ready to review" editables on the "get next editable" list (:pr:`5765`)
 - Disallow uploading empty files (:pr:`5767`)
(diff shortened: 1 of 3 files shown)


ThiefMaster:
This is also used when just viewing the timetable. Do we really want to include it there as well?

ThiefMaster:
ping :)

duartegalvao:
> This is also used when just viewing the timetable. Do we really want to include it there as well?

Fixed!

ThiefMaster (approved):

PR #5740, opened 2023-04-17
Include start/end times of whole booking in tooltip

### LINKED ISSUE #5730: Include start/end of whole booking in tooltip
This is for the "Bookings" timeline view. In case of recurrent bookings it would be useful if the tooltip not only showed the time for the current occurrence, but also for the whole booking (start/end date, maybe repetition schema as well).

GovernmentPlates:
Closes #5730 
Screenshot 2023-04-17 at 14 12 06

### DIFF
--- a/indico/modules/rb/client/js/common/timeline/TimelineItem.jsx
+++ b/indico/modules/rb/client/js/common/timeline/TimelineItem.jsx
@@ -165,6 +165,8 @@ class TimelineItem extends React.Component {
       }
     }
 
+    const fmt = dayBased ? 'L' : 'L LT';
+
     return dayBased && !message ? null : (
       <div styleName="popup-center">
         {!dayBased && (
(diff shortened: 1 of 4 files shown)


tomasr8:
All looks good, just the string interpolation :smile: Ideally, we'd have 3 separate messages depending on the recurrence frequency

indico/modules/rb/client/js/common/timeline/TimelineItem.jsx
```diff
@@ -138,8 +138,16 @@ class TimelineItem extends React.Component {
...
+  renderMessagePopup = (message, segmentStartDt, segmentEndDt, reservation) => {
     const {dayBased} = this.props;
+    const mapRecurrenceTypeToInfo = type =>
+      type.toLowerCase() === 'day'
+        ? Translate.string('daily')
+        : type.toLowerCase() === 'week'
+        ? Translate.string('weekly')
+        : type.toLowerCase() === 'month'
+        ? Translate.string('monthly')
+        : '';
```
ThiefMaster, line 150: A simple dict would be much easier here than a function ;) But in any case, as @tomasr8 mentioned it would be better to have 3 separate i18n strings so they can be properly translated.

ThiefMaster: I think you also need to take the repeatInterval into account - right now the tooltip wouldn't be correct for something that repeats e.g. every 2 weeks

ThiefMaster (approved):

PR #5755, opened 2023-05-03
Allow emailing users with no editable submissions

duartegalvao:
This PR adds the ability to email contribution persons that have not yet made any submissions to a given editable type:

Screen Shot 2023-05-03 at 17 17 31

### DIFF
--- a/indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
+++ b/indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
@@ -8,6 +8,7 @@
 /* global ajaxDialog:false */
 
 import anonymousTeamURL from 'indico-url:event_editing.api_anonymous_team';
+import contribsWithNoEditablesURL from 'indico-url:event_editing.api_contribs_with_no_editables';
 import enableEditingURL from 'indico-url:event_editing.api_editing_enabled';
 import emailMetadataURL from 'indico-url:event_editing.api_email_not_submitted_metadata';
 import emailPreviewURL from 'indico-url:event_editing.api_email_not_submitted_preview';
@@ -26,7 +27,7 @@ import {Checkbox, Loader} from 'semantic-ui-react';
 
 import {EmailContribAbstractRolesButton} from 'indico/modules/events/persons/EmailContribAbstractRolesButton';
 import {ManagementPageSubTitle, ManagementPageBackButton} from 'indico/react/components';
-import {useTogglableValue} from 'indico/react/hooks';
+import {useIndicoAxios, useTogglableValue} from 'indico/react/hooks';
 import {Translate} from 'indico/react/i18n';
 import {useNumericParam} from 'indico/react/util/routing';
 
@@ -78,6 +79,12 @@ export default function EditableTypeDashboard() {
     poster: Translate.string('Assign an editor to posters'),
     slides: Translate.string('Assign an editor to slides'),
   }[type];
+
+  const {data} = useIndicoAxios({
+    url: contribsWithNoEditablesURL({event_id: eventId, type}),
+    method: 'POST',
+  });
+  const numEditables = (data && data.no_contribs) || 0;
 
   return (
     <>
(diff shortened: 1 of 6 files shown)


tomasr8:
What do you think about adding the new section to the first action box?
image

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -256,3 +259,31 @@ def _process(self):
...
+
+class RHEmailNotSubmittedEditablesMetadata(EmailRolesMetadataMixin, RHEditableTypeManagementBase):
+    object_context = 'contributions'
+
+
+class RHEmailNotSubmittedEditablesPreview(EmailRolesPreviewMixin, RHEditableTypeManagementBase):
+    object_context = 'contributions'
+
+    def get_placeholder_kwargs(self):
+        contribution = Contribution.query.with_parent(self.event).first()
```
tomasr8, line 272: I think there should be some check in case there are no contribs in the event. Otherwise the email preview breaks when you use any of the contribution placeholders

duartegalvao: Fixed, and also added a count of contributions w/o editables

tomasr8: I like it!

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -256,3 +259,41 @@ def _process(self):
         editors = get_editors(self.event, self.editable_type)
         return jsonify_template('events/editing/management/editor_list.html',
                                 event_persons=editors, event=self.event)
+
+
+class RHEditablesNotSubmitted(RHEditableTypeManagementBase):
+    """Check if there are contributions without editables of this type."""
```
tomasr8, line 265: Maybe more precise would be `Return the number of contributions ...`? Since it returns the number rather than a boolean

duartegalvao: Oops, yeah initially it was just a boolean and then I decided to change

indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
```diff
@@ -9,6 +9,10 @@
 
 import anonymousTeamURL from 'indico-url:event_editing.api_anonymous_team';
 import enableEditingURL from 'indico-url:event_editing.api_editing_enabled';
+import emailMetadataURL from 'indico-url:event_editing.api_email_not_submitted_metadata';
+import emailPreviewURL from 'indico-url:event_editing.api_email_not_submitted_preview';
+import emailSendURL from 'indico-url:event_editing.api_email_not_submitted_send';
+import noEditablesNotSubmittedURL from 'indico-url:event_editing.api_no_editables_not_submitted';
```
tomasr8, line 15: This looks like a double negative and my brain fails to parse it :sweat_smile:
maybe `numberEditablesNotSubmitted` or `countEditablesNotSubmitted`? Or even something like `ContribsWithNoEditables` since it actually counts contributions and not editables? What do you think?

indico/modules/events/editing/client/js/management/editable_type/EditableTypeDashboard.jsx
```diff
@@ -75,6 +80,12 @@ export default function EditableTypeDashboard() {
     slides: Translate.string('Assign an editor to slides'),
   }[type];
 
+  const {data} = useIndicoAxios({
+    url: contribsWithNoEditablesURL({event_id: eventId, type}),
+    method: 'POST',
+  });
+  const numEditables = (data && data.no_contribs) || 0;
```
tomasr8, line 87: This could be simply `data.count` and maybe `numContribs` instead?

ThiefMaster: yes, absolutely. no need to POST here - it's neither modifying something nor has it a large payload (e.g. list of selected ids)...

tomasr8: Optional chaining could also simplify this:
```suggestion
  const numEditables = data?.no_contribs || 0;
```

ThiefMaster (approved):
works like a charm :)

PR #5756, opened 2023-05-04
Add privacy policy checkbox to registration forms

tomasr8:
Configurable in the privacy settings (each regform can be set separately):

obrazek

When enabled, users will see this when registering:

obrazek

It includes a link to the event's privacy policy.

### DIFF
--- a/indico/modules/events/registration/client/js/form_submission/RegistrationFormSubmission.jsx
+++ b/indico/modules/events/registration/client/js/form_submission/RegistrationFormSubmission.jsx
@@ -147,6 +147,9 @@ export default function RegistrationFormSubmission() {
   const showConsentToPublish = !isManagement && publishToParticipants !== 'hide_all';
 
   const onSubmit = async (data, form) => {
+    // There is no need to store whether the user accepted the privacy policy
+    // as it is not possible to submit the form without accepting it.
+    data = _.omit(data, 'agreed_to_privacy_policy');
     let resp;
     const formData = isUpdateMode ? getChangedValues(data, form, ['notify_user']) : data;
     try {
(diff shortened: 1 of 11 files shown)


indico/modules/events/registration/client/js/form_submission/RegistrationFormSubmission.jsx
```diff
@@ -37,6 +39,34 @@ import {
...
+    </label>
+  );
+
+  return (
+    <Message info style={{marginTop: 25}}>
+      <Message.Header>
+        <Translate>Privacy policy</Translate>
+      </Message.Header>
+      <Form as="div" style={{marginTop: 10}}>
+        <FinalCheckbox required label={label} name="agreed_to_privacy_policy" toggle />
```
ThiefMaster, line 60: When this checkbox is selected the submission fails because the schema on the server does not expect the `agreed_to_privacy_policy` field... so it needs to be included in the schema as well

tomasr8: It's strange that I didn't run into this issue when I was testing (though it's been a while..) The fix is to just drop the value from the form data since there's no need to store this in the backend at all - it's not possible to submit the form w/o accepting the privacy policy so any existing registration will have accepted it

ThiefMaster: Sending it to the backend and only discarding it there would have the advantage to also catch anything registering by calling the API directly. But since this is an internal API I don't think that's necessary, so your fix is fine

ThiefMaster (approved):

PR #5757, opened 2023-05-05
Export all user data

tomasr8:
Allows users to export all their user data stored in Indico. The export is a single zip file containing a yaml file `data.yml` with all the data.
Additionally, if the user has uploaded any files to Indico (attachments, abstracts, papers, registration file fields, etc..) these will also be included in the zip file in different subfolders.

### Overview of the added functionality
A new tab is added in the user profile (`/user/data-export`)

obrazek

You can select what categories you want to export. Once submitted, you will be notified via email when the
export is finished (successfully or not). You can also see the progress on the `Data export` tab.

You can download the data using the link provided in the email or by going back to your user profile and downloading it from there:

obrazek

To simplify things, only one request can exist at a time. When you request a new export, the old one is deleted (if finished, if it's still running you can't request it again).

When the export is processing:
obrazek

When it has failed (hopefully should not happen, but at least we log failures)
obrazek

When the export expires (by default the exports are cleaned after 1 month):
obrazek

When the export exceeds maximum (configurable) size:
obrazek

### Technical overview

New DB model `DataExportRequest` was added. This object describes the export request - state, requested_dt, selected options, etc..
It's in 1-to-1 relation with `User`. You can get the current export request via `user.data_export_request`

There are two new endpoints, one to render the `Data export` tab (`RHUserDataExport`) and one for the API (`RHUserDataExportAPI`) which is used to create new requests and check the state.

The data that ends up in `data.yml` is serialized using `UserDataExportSchema`. Files written to the zip archive have
unique names.

The export itself is a celery task since it is expected to run for a long time (= lots of queries to do)

I added lots of new DB fixtures which I needed to test this PR. The schema tests use snapshots from `modules/users/tests`.

### TODO:

- [x] Export editables
- [x] Export subcontributions
- [x] Do we want to export deleted items? ~~Technically we still have the data for them..~~ Yes
- [x] Tests!
- [x] File size limit
- [x] Option to include/not include files


indico/modules/users/tasks.py
```diff
@@ -37,3 +37,8 @@ def update_gravatars(user=None):
             continue
         set_user_avatar(user, gravatar, 'gravatar', lastmod)
         logger.info('Gravatar of user %s updated', user)
+
+
+# Should be similar to the Material export
```
ThiefMaster, line 42: I think you can easily ditch polling etc like we have for the material export and just send an email to the user once the export is ready...

indico/migrations/versions/20230905_1200_e2b69fe5155d_add_user_data_export_table.py
```diff
@@ -0,0 +1,48 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'data_export_requests',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('requested_dt', UTCDateTime, nullable=False),
+        sa.Column('selected_options', ARRAY(sa.Enum(DataExportOptions)), nullable=False),
```
ThiefMaster, line 28: I'd prefer to avoid the native enum in favor of our PyIntEnum type (or set `native_enum=False` in case our type doesn't work properly inside arrays which would absolutely not surprise me...)

https://www.postgresql.org/docs/12/sql-altertype.html#id-1.9.3.42.7

From 12+ you can modify enums in transactions, but not being able to use new enum members in the same transactions where they are being added means we'll never be able to do data migrations that add a new enum type, because in case of alembic upgrades all upgrades are done within the same transaction

docs/source/config/settings.rst
```diff
@@ -724,6 +724,15 @@ Storage
...
     Default: ``None``
 
+.. data:: MAX_DATA_EXPORT_SIZE
+
+    The maximum file size (in MB) for the user data export zip archive.
+    Note that this limit does not apply to the main ``data.yml`` file which is
+    always generated and typically does not exceed several megabytes. However,
+    any additional files such as attachments, papers, etc., that the user might
+    have and that do not fit within the limit are not included in the exported
+    archive. The default size is 10GB.
```
ThiefMaster, line 734:
```suggestion
    archive. The default size is 10GB.
    
    Default: ``10 * 1024``
```

indico/migrations/versions/20230905_1200_e2b69fe5155d_add_user_data_export_table.py
```diff
@@ -0,0 +1,48 @@
...
+Revises: 0c44046dc1be
+Create Date: 2023-05-08 15:37:24.134940
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import ARRAY
+
+from indico.core.db.sqlalchemy import PyIntEnum, UTCDateTime
+from indico.modules.users.models.export import DataExportOptions, DataExportRequestState
```
ThiefMaster, line 13: don't import enums, redefine them in the model (see some earlier revisions, searching in them for `class _` will probably yield good results) - otherwise the revision will break if the enum gets changed in the future

indico/testing/fixtures/storage.py
```diff
@@ -44,7 +45,39 @@ def getsize(self, file_id):
 
 
 @pytest.fixture
-def dummy_attachment(dummy_user):
-    folder = AttachmentFolder(title='dummy_folder', description='a dummy folder')
-    file_ = AttachmentFile(user=dummy_user, filename='dummy_file.txt', content_type='text/plain')
-    return Attachment(folder=folder, user=dummy_user, title='dummy_attachment', type=AttachmentType.file, file=file_)
+def create_attachment(db):
+    """Return a callable which lets you create attachments."""
+    def _create_attachment(user, object, *, title, attachment_folder_id=None, attachment_file_id=None, **kwargs):
```
ThiefMaster, line 50: I'd probably keep the title positional since it's required anyway

indico/testing/fixtures/storage.py
```diff
@@ -44,7 +45,39 @@ def getsize(self, file_id):
...
+    return create_attachment(dummy_user, dummy_event, title='dummy_attachment', description='Dummy attachment',
+                             id=42, attachment_folder_id=42, attachment_file_id=42)
+
+
+@pytest.fixture
+def create_file(db):
+    """Return a callable which lets you create (unclaimed) files."""
+    def _create_file(filename, content_type, context, data, **kwargs):
+        file = File(filename=filename, content_type=content_type, **kwargs)
+        file.save(context, data.encode('utf-8'))
```
ThiefMaster, line 73: `'utf-8'` is implied so you can omit this

indico/testing/fixtures/editing.py
```diff
@@ -0,0 +1,84 @@
...
+        db.session.add(file_type)
+        db.session.flush()
+        return file_type
+    return _create_editing_file_type
+
+
+@pytest.fixture
+def dummy_editing_file_type(create_editing_file_type):
+    """Create a dummy editing file type."""
+    return create_editing_file_type(id=42, name='TXT', extensions=['txt'], type=EditableType.paper)
```
ThiefMaster, line 67: I think all those hardcoded (but still somewhat low) IDs are a bit dangerous since if you were to create 42 objects in a test, you'd get a collision w/ the dummy one.

indico/modules/events/surveys/schemas.py
```diff
@@ -0,0 +1,30 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2023 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from marshmallow.fields import Function, List, Nested
```
ThiefMaster, line 8: I prefer `from marshmallow import fields` and using `fields.Whatever` in the schemas (it's mainly pretty old code that uses tons of individual type imports ;0)

indico/modules/events/surveys/schemas.py
```diff
@@ -0,0 +1,30 @@
...
+    question_title = Function(lambda answer: answer.question.title)
+
+
+class SurveySubmissionSchema(mm.SQLAlchemyAutoSchema):
+    class Meta:
+        model = SurveySubmission
+        fields = ('id', 'survey_id', 'survey_title', 'submitted_dt', 'is_anonymous', 'is_submitted', 'pending_answers',
+                  'answers')
+
+    survey_title = Function(lambda submission: submission.survey.title)
```
ThiefMaster, line 29: I think you can use `String(attribute='survey.title')` for this

indico/modules/users/__init__.py
```diff
@@ -72,6 +72,7 @@ def _sidemenu_items(sender, user, **kwargs):
     yield SideMenuItem('emails', _('Emails'), url_for('users.user_emails'), 70, disabled=user.is_system)
     yield SideMenuItem('preferences', _('Preferences'), url_for('users.user_preferences'), 60, disabled=user.is_system)
     yield SideMenuItem('favorites', _('Favorites'), url_for('users.user_favorites'), 50, disabled=user.is_system)
+    yield SideMenuItem('data_export', _('Data export'), url_for('users.user_data_export'), 40, disabled=user.is_system)
```
ThiefMaster, line 75: The "API Tokens" menu item is already on 40. I think this particular option can easily be made the bottom-most one you could go for position `0`.

indico/modules/users/client/js/DataExport.jsx
```diff
@@ -0,0 +1,246 @@
...
+import './DataExport.module.scss';
+
+function DataExportForm({onSubmit, exportOptions}) {
+  return (
+    <FinalForm
+      onSubmit={onSubmit}
+      initialValues={{}}
+      initialValuesEqual={_.isEqual}
+      validate={values => {
+        if (!_.values(values).some(v => v)) {
```
ThiefMaster, line 33: I don't think we really need lodash for this anymore: `Object.values(values),some(v => v)`

indico/modules/users/client/js/DataExport.jsx
```diff
@@ -0,0 +1,246 @@
...
+  }).isRequired,
+  exportOptions: PropTypes.array.isRequired,
+};
+
+function DataExport({userId, request, exportOptions}) {
+  const userIdArgs = userId !== null ? {user_id: userId} : {};
+  const [state, setState] = useState(request.state);
+
+  const onSubmit = async data => {
+    const selection = _.keys(_.pickBy(data));
```
ThiefMaster, line 111: `Object.keys(...)`

indico/modules/users/tasks.py
```diff
@@ -37,3 +60,215 @@ def update_gravatars(user=None):
...
+    fields = []
+    for opt in options:
+        fields += options_map.get(opt, (opt.name,))
+    return UserDataExportSchema(only=fields).dump(user)
+
+
+def generate_zip(export_request, temp_file, max_size_mb):
+    data = get_data(export_request)
+    with ZipFile(temp_file, 'w', allowZip64=True) as zip_file:
+        zip_file.writestr('/data.yml', yaml.dump(data))
```
ThiefMaster, line 133: Hadn't we agreed on creating separate yaml files for each top-level key?

```suggestion
        for key, subdata in data.items():
            zip_file.writestr(f'/{key}.yml', yaml.dump(subdata))
```

indico/modules/users/tasks.py
```diff
@@ -37,3 +60,215 @@ def update_gravatars(user=None):
...
+            else:
+                export_request.max_size_exceeded = True
+                logger.warning('User data export exceeds size limit, exporting only partial data: %r',
+                               export_request)
+                break
+
+
+def get_size_mb(file):
+    file = getattr(file, 'file', file)
+    return file.size / 1024  # file size is in KB
```
ThiefMaster, line 148:
```suggestion
    return file.size / 1024**2  # file size is in bytes
```

or maybe just convert the limit to bytes above and use that for the comparison ;)

indico/migrations/versions/20230905_1200_e2b69fe5155d_add_user_data_export_table.py
tomasr8: iirc I did this exactly because our enum wasn't compatible. I'll try with `native_enum=False`

indico/modules/users/tasks.py
tomasr8: Have we? I'm fine with either, just genuinely don't remember.. :sweat_smile:

ThiefMaster: at least I thought so... in any case, it makes it much easier to read the files as a human

fwiw i have the suggestion locally since i wanted to test the readability with the separate files so i can push that

tomasr8: yeah if you already have it, feel free to push it :)

indico/modules/users/models/export.py
```diff
@@ -0,0 +1,133 @@
...
+from indico.core.db import db
+from indico.core.db.sqlalchemy import PyIntEnum, UTCDateTime
+from indico.util.date_time import now_utc
+from indico.util.enum import RichIntEnum
+from indico.util.i18n import L_
+
+
+class DataExportRequestState(RichIntEnum):
+    __titles__ = [L_('Pending'), L_('Running'), L_('Success'), L_('Failed'), L_('Expired')]
+    pending = 0  # Default value when there is no request
```
ThiefMaster, line 20: I don't think `pending` is a good name for this, at least to me this implies you already submitted the request but it hasn't been picked up by the Celery worker yet. I'd say `none` would fit better.

indico/modules/users/models/export.py
```diff
@@ -0,0 +1,133 @@
...
+    success = 2
+    failed = 3
+    expired = 4  # The associated file has been deleted
+
+
+class DataExportOptions(RichIntEnum):
+    __titles__ = [L_('Personal data'), L_('Settings'), L_('Minutes & Contributions'), L_('Registrations'),
+                  L_('Room booking data'), L_('Abstracts & Papers'), L_('Survey submissions'),
+                  L_('Attachments & Materials'), L_('Editables'), L_('Miscellaneous')]
+    personal_data = 0
```
ThiefMaster, line 31: I usually try to avoid `0` for enum members that do not have some falsy meaning and start with 1 there (you can just put a `None` as the first titles item in that case)

indico/modules/users/models/export.py
```diff
@@ -0,0 +1,133 @@
...
+        backref=db.backref(
+            'data_export_request',
+            lazy=True,
+            uselist=False,
+            cascade='all, delete-orphan',
+            passive_deletes=True
+        )
+    )
+
+    file = db.relationship(
```
ThiefMaster, line 104: we're missing a backref here. it's good to always have one even when not actively using it

indico/modules/users/models/export.py
```diff
@@ -0,0 +1,133 @@
...
+        'File',
+        lazy=False,
+        cascade='all, delete-orphan',
```
ThiefMaster, line 107: I'm not sure this is a good idea. We usually do not want to simply delete File objects from the DB just because the thing referencing them got deleted - this would leave the file orphaned on the storage backend, with no metadata in indico.

You could copy the relationship config from some other place like this one and change the backref+attribute names:

```python
    custom_boa = db.relationship(
        'File',
        lazy=True,
        backref=db.backref(
            'custom_boa_of',
            lazy=True
        )
    )
```

indico/modules/users/tasks.py
```diff
@@ -37,3 +60,211 @@ def update_gravatars(user=None):
...
+
+
+def _export_user_data(export_request, buffer):
+    generate_zip(export_request, buffer, config.MAX_DATA_EXPORT_SIZE * 1024**2)
+    buffer.seek(0)
+
+    file = File(filename='data-export.zip', content_type='application/zip')
+    try:
+        file.save(('user', export_request.user.id), buffer)
+        file.claim()
```
ThiefMaster, line 97: We delete unclaimed files after 1 day, so if the file is claimed we need some custom logic to delete those archives after a certain time (1 week maybe or when requesting a new one?)

indico/modules/users/tasks.py
```diff
@@ -37,3 +60,211 @@ def update_gravatars(user=None):
...
+    for opt in options:
+        fields += options_map.get(opt, (opt.name,))
+    return UserDataExportSchema(only=fields).dump(user)
+
+
+def generate_zip(export_request, temp_file, max_size):
+    data = get_data(export_request)
+    with ZipFile(temp_file, 'w', allowZip64=True) as zip_file:
+        for key, subdata in data.items():
+            zip_file.writestr(f'{key}.yaml', yaml.dump(subdata))
```
ThiefMaster, line 134: this might be nicer for human-readability:

```suggestion
            zip_file.writestr(f'{key}.yaml', yaml.dump(subdata, allow_unicode=True))
```

```
>>> print(yaml.dump({'name': 'Adrian Mönnich'}))
name: "Adrian M\xF6nnich"


>>> print(yaml.dump({'name': 'Adrian Mönnich'}, allow_unicode=True))
name: Adrian Mönnich
```

indico/modules/users/schemas.py
```diff
@@ -83,3 +106,360 @@ class Meta:
...
+        model = Registration
+        fields = ('id', 'url', 'event_id', 'event_title', 'event_url', 'state', 'currency', 'price', 'submitted_dt',
+                  'email', 'first_name', 'last_name', 'is_deleted', 'checked_in', 'checked_in_dt', 'rejection_reason',
+                  'consent_to_publish', 'created_by_manager', 'tags', 'data')
+
+    price = fields.Decimal(as_string=True)
+    url = Function(lambda reg: url_for('event_registration.registration_details', reg, _external=True))
+    event_title = Function(lambda reg: reg.event.title)
+    event_url = Function(lambda reg: url_for('events.display', reg.event, _external=True))
+    tags = List(Nested(RegistrationTagExportSchema))
```
ThiefMaster, line 228: I don't think tags should be exported. Those are internal to the event organizers and I think they do not expect them to be disclosed to users.... We also don't export manager-only fields.

indico/modules/users/schemas.py
```diff
@@ -83,3 +106,360 @@ class Meta:
...
+    data = Method('_serialize_data')
```
ThiefMaster, line 229: This returns a somewhat useless list right now:

```yaml
  data:
  - null
  - null
  - null
  - null
  - null
  - null
  - null
```

indico/modules/users/schemas.py
```diff
@@ -83,3 +106,360 @@ class Meta:
     location = fields.String(attribute='event.location.venue_name')
     chain_titles = fields.List(fields.String(), attribute='category.chain_titles')
     label_markup = fields.Function(lambda e: e.get_label_markup('mini'))
+
+
+# Schemas for user data export
```
ThiefMaster, line 111: Maybe all those schemas should go in a separate `export_schemas.py` file? Otherwise they add a LOT of clutter to the other more generic schemas in here

indico/modules/users/schemas.py
```diff
@@ -83,3 +106,360 @@ class Meta:
...
+
+class DataExportRequestSchema(mm.SQLAlchemyAutoSchema):
+    class Meta:
+        model = DataExportRequest
+        fields = ('state', 'requested_dt', 'selected_options', 'max_size_exceeded', 'url')
+
+    selected_options = List(Enum(DataExportOptions))
+
+
+class PersonalTokenExportSchema(mm.SQLAlchemyAutoSchema):
```
ThiefMaster, line 121: Some data about legacy API keys should be exported as well

indico/modules/users/schemas.py
```diff
@@ -83,3 +106,360 @@ class Meta:
...
+        fields = ('name', 'revoked_dt', 'created_dt', 'last_used_dt', 'last_used_ip', 'use_count', 'scopes')
+
+    scopes = Function(lambda token: sorted(list(token.scopes)))
+
+
+class OAuthApplicationExportSchema(mm.SQLAlchemyAutoSchema):
+    class Meta:
+        model = OAuthApplication
+        fields = ('name', 'description', 'allowed_scopes', 'redirect_uris', 'is_enabled', 'is_trusted',
+                  'allow_pkce_flow')
```
ThiefMaster, line 133: I don't think anything besides name/description should be exported here - those are system-level apps, and things like the list of redirect URIs are none of the user's business

indico/modules/users/tasks.py
```diff
@@ -37,3 +60,211 @@ def update_gravatars(user=None):
...
+            if size <= max_size:
+                write_file(zip_file, file)
+            else:
+                export_request.max_size_exceeded = True
+                logger.warning('User data export exceeds size limit, exporting only partial data: %r',
+                               export_request)
+                break
+
+
+def collect_files(export_request):
```
ThiefMaster, line 147: I didn't check how but this yields duplicate files (which are then added to the zip file twice, with the exact same path) - I only noticed this because when unpacking the zip file with Total Commander into an empty folder it asked me whether to overwrite some files.

ThiefMaster: Looks like it's files in editables. So I think deduplicating the list returned from this function is sufficient

indico/modules/users/schemas_test.py
```diff
@@ -5,9 +5,44 @@
...
+@pytest.fixture(scope='function')
+def freeze_time():
+    dt = datetime(2022, 5, 22, 12, 0, 0)
+    with freezegun.freeze_time(dt):
+        yield
+
+
+@pytest.fixture(scope='function')
+def snapshot(snapshot):
+    def assert_json_match(data, filename):
```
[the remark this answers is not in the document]
tomasr8: Definitely, this was mostly to avoid having to specify `snapshot_dir` in every test.. I'm trying to see if there's a more elegant solution for that

indico/modules/users/schemas_test.py
```diff
@@ -16,3 +51,193 @@ def test_personal_data_schema():
...
+    link = ContributionPersonLink(is_speaker=True, person=event_person)
+    contribution.person_links = [link]
+
+    note = EventNote(object=contribution)
+    note.create_revision(RenderMode.markdown, '**Contrib Test**', user)
+    db.session.add(note)
+
+
+def test_contribution_export_schema(snapshot, db, dummy_user, dummy_contribution, dummy_event_person):
+    from indico.modules.users.schemas import ContributionExportSchema
```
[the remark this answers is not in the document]
tomasr8: nope :/ I'll try to dig into why later (it's the standard sqlalchemy error)

indico/modules/users/tasks.py
tomasr8: Probably same file referenced from multiple revisions. Including the revision id in the path should fix that

ThiefMaster: since it's the same file, simply not adding it to the archive again is actually nicer to avoid having the same data multiple times

tomasr8: was thinking the revision info could be useful - maybe if you wanted know which files are from which revision, or maybe see only files from the latest revision. but yeah, this'll duplicate the files so if there are some large files with lots of revisions the export could be needlessly large..

ThiefMaster: I'd expect people who want to do stuff with the data to process the yaml files and get the proper paths from there

tomasr8: That could work if I update the export schema to include editing revisions in the yaml file (currently I just list all the files together)

indico/modules/users/export_schemas.py
```diff
@@ -0,0 +1,398 @@
...
+    static_sites = List(Nested(StaticSiteExportSchema))
+    identities = List(Nested(IdentityExportSchema))
+    personal_tokens = List(Nested(PersonalTokenExportSchema))
+    api_key = Nested(APIKeyExportSchema)
+    old_api_keys = List(Nested(APIKeyExportSchema))
+    oauth_applications = Method('_serialize_oauth_apps')
+
+    def _serialize_oauth_apps(self, user):
+        oauth_apps = [link.application for link in user.oauth_app_links]
+        return OAuthApplicationExportSchema(many=True).dump(oauth_apps)
```
ThiefMaster, line 290: I think exporting the scopes on the link itself would be good, since that's the scope the application currently has for this particular user.

Something like this (ie using a `List(Nested(OAuthApplicationUserLinkExportSchema))`

```yaml
oauth_applications:
- scopes: ...
  application:
    name: ...
    description: ...
```

indico/modules/users/tasks.py
```diff
@@ -68,9 +68,14 @@ def export_user_data(user, options):
     if export_request and export_request.is_running:
         return
 
-    # ughh why isn't there an ORM upsert..
+    # If an export request already exists we delete it
+    # along with the actual file (if any) and create a new request
     if export_request:
+        f = export_request.file
         db.session.delete(export_request)
+        if f:
```
ThiefMaster, line 76: I'd just set `f.claimed = False` - that way the task deals with it which has much nicer error handling (see `delete_unclaimed_files`)

indico/modules/users/export_schemas.py
```diff
@@ -0,0 +1,398 @@
...
+        model = PersonalToken
+        fields = ('name', 'revoked_dt', 'created_dt', 'last_used_dt', 'last_used_ip', 'use_count', 'scopes')
+
+    scopes = Function(lambda token: sorted(list(token.scopes)))
+
+
+class APIKeyExportSchema(mm.SQLAlchemyAutoSchema):
+    class Meta:
+        model = APIKey
+        fields = ('is_active', 'is_blocked', 'is_persistent_allowed', 'created_dt', 'lat_used_dt', 'last_used_ip',
```
ThiefMaster, line 62:
```suggestion
        fields = ('is_active', 'is_blocked', 'is_persistent_allowed', 'created_dt', 'last_used_dt', 'last_used_ip',
```

ThiefMaster (commented):
Thanks for those individual commits - much nicer to review them piece by piece!

indico/modules/users/export_schemas.py
```diff
@@ -0,0 +1,432 @@
...
+class PersonalDataExportSchema(mm.SQLAlchemyAutoSchema):
+    class Meta:
+        model = User
+        fields = ('first_name', 'last_name', 'full_name', 'title', 'email', 'secondary_emails',
+                  'affiliation', 'phone', 'avatar_url', 'picture_metadata', 'picture_source',
+                  'favorite_events', 'favorite_categories')
+
+    title = NoneValueEnumField(UserTitle, none_value=UserTitle.none, attribute='_title')
+    secondary_emails = fields.Function(lambda user: list(user.secondary_emails))
+    affiliation = fields.Nested(AffiliationSchema, attribute='affiliation_link')
```
ThiefMaster, line 277: You need to include both `affiliation` and `affiliation_link` - the latter is often None (in case of the usual free-text affiliations).

Also, exclude the `meta` field the affiliation link. That's internal and not relevant for users.

indico/modules/users/export_schemas.py
```diff
@@ -0,0 +1,432 @@
...
+    favorite_events = fields.List(fields.Nested(EventExportSchema))
+    favorite_categories = fields.List(fields.Nested(CategoryExportSchema))
+
+
+class RoomBookingExportSchema(Schema):
+    from indico.modules.rb.schemas import ReservationSchema
+
+    favorite_rooms = fields.List(fields.Nested(RoomExportSchema))
+    owned_rooms = fields.List(fields.Nested(RoomExportSchema))
+    reservations = fields.List(fields.Nested(ReservationSchema))
```
ThiefMaster, line 287: `reservations_booked_for` should be included as well (`reservations` are the ones the user created)

docs/source/config/settings.rst
```diff
@@ -746,6 +746,17 @@ Storage
 
     Default: ``None``
 
+.. data:: MAX_DATA_EXPORT_SIZE
+
+    The maximum file size (in MB) for the user data export zip archive.
+    Note that this limit does not apply to the main ``data.yml`` file which is
```
ThiefMaster, line 752: `data.yml` no longer exists. Maybe reword this ;)

indico/modules/events/registration/testing/fixtures.py
```diff
@@ -15,7 +15,7 @@
 @pytest.fixture
 def dummy_regform(db, dummy_event):
     """Create a dummy registration form for the dummy event."""
-    regform = RegistrationForm(event=dummy_event, title='Registration Form', currency='USD')
+    regform = RegistrationForm(id=420, event=dummy_event, title='Registration Form', currency='USD')
```
ThiefMaster, line 18: LOL 🚬

indico/modules/users/export.py
```diff
@@ -0,0 +1,320 @@
...
+
+def get_old_requests(days):
+    """Get successful export requests older than the specified number of days."""
+    return (DataExportRequest.query
+            .filter(DataExportRequest.requested_dt < (now_utc() - timedelta(days=days)),
+                    DataExportRequest.state == DataExportRequestState.success)
+            .all())
+
+
+@email_sender
```
ThiefMaster, line 303: Ugh I really need to remove this stupid decorator altogether some day. It's one of those things that someone added after they took the 'advanced python' course and thought decorators are the hail mary :D

IMHO it's much clearer to have `send_email(make_email(...))` where you immediately see "ok that sends an email"...

indico/modules/users/models/export.py
```diff
@@ -0,0 +1,146 @@
...
+        self.file = None
+        self.state = DataExportRequestState.failed
+
+    def delete(self):
+        if self.file:
+            self.file.claimed = False
+        db.session.delete(self)
+
+    def __repr__(self):
+        return f'<DataExportRequest({self.id}, {self.user_id}, {self.state}, {self.file})>'
```
ThiefMaster, line 146: `return format_repr(self, 'id', 'user_id', 'state', 'file')`

uebmaster:
maybe here is not the place to ask, but, do you know how to see the list of users?

I see in admin---> users i can only search user by user, and i want the full list.

ThiefMaster:
yes, this is the wrong place. please use our forum, https://talk.getindico.io

Issue #5759, opened 2023-05-08
Internal search does get None sometimes


bpedersen2:
`chain_query` seems to be `None` in some cases:

I one dismisses the error dialog Contributions is empty ( as expected)

```
  File "/opt/indico/.venv/lib/python3.9/site-packages/indico/modules/search/controllers.py", line 60, in _process
    result = search_provider().search(q, session.user, page, type, **params)
  File "/opt/indico/.venv/lib/python3.9/site-packages/indico/modules/search/internal.py", line 75, in search
    pagenav, results = self.search_contribs(query, user, page, category_id, event_id,
  File "/opt/indico/.venv/lib/python3.9/site-packages/indico/modules/search/internal.py", line 267, in search_contribs
    objs, pagenav = self._paginate(query, page, Contribution.id, user, admin_override_enabled)
  File "/opt/indico/.venv/lib/python3.9/site-packages/indico/modules/search/internal.py", line 164, in _paginate
    res = get_n_matching(
  File "/opt/indico/.venv/lib/python3.9/site-packages/indico/core/db/sqlalchemy/util/queries.py", line 172, in get_n_matching
    preload_bulk(objects)
  File "/opt/indico/.venv/lib/python3.9/site-packages/indico/modules/search/internal.py", line 168, in <lambda>
    preload_bulk=lambda objs: self._preload_categories(objs, preloaded_categories)
  File "/opt/indico/.venv/lib/python3.9/site-packages/indico/modules/search/internal.py", line 115, in _preload_categories
    category_ids = set(itertools.chain.from_iterable(id for id, in chain_query))
TypeError: 'NoneType' object is not iterable
```

'https://indico.frm2.tum.de/search/?q=MLZ'

**Expected behavior**
No error

I'll try to create a patch...

ThiefMaster:
I don't see how `chain_query` can be None. I think it's more likely that one of the rows contain `None` for the ID and it fails in the outer iteration... 

It's only the contrib search that fails: https://indico.frm2.tum.de/search/api/search?q=MLZ&type=contribution&type=subcontribution
The more interesting question is why that happens though. Do you have any unlisted events in your database (you can check by querying for events with a None `category_id`)?

ThiefMaster:
Do you have both subcontributions anad normal contributions?
`Contribution.query.count()`
`SubContribution.query.count()`

bpedersen2:
Not for the unlisted event. 

The error surfaces once you have an unlisted event with published(!) contributions. 

[...]

PR #5760, opened 2023-05-08
Fix None chain_query errors

bpedersen2:
chain_query may be None, so add a guard against this.

Fixes: #5759


bpedersen2:
Abandoned in favor of #5761

PR #5767, opened 2023-05-11
Disallow uploading zero-byte files

duartegalvao:
This PR disallows uploading zero-byte files, which may help prevent user-errors.

image

### DIFF
--- a/indico/modules/files/controllers.py
+++ b/indico/modules/files/controllers.py
@@ -42,6 +42,9 @@ class UploadFileMixin:
         content_type = mimetypes.guess_type(file.filename)[0] or file.mimetype or 'application/octet-stream'
         f = File(filename=file.filename, content_type=content_type)
         f.save(context, stream)
+        if not f.size:
+            f.delete()
+            raise UnprocessableEntity
         db.session.add(f)
         db.session.flush()
         logger.info('File %r uploaded (context: %r)', f, context)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,6 +41,7 @@ Improvements
 - Allow emailing contribution persons that have not yet made any submissions to a
   given editable type (:pr:`5755`)
 - Show only "ready to review" editables on the "get next editable" list (:pr:`5765`)
+- Disallow uploading zero-byte files (:pr:`5767`)
 
 Bugfixes
 ^^^^^^^^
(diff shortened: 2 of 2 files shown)


indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -50,6 +50,9 @@ class RHEditingUploadFile(UploadFileMixin, RHContributionEditableBase):
     def get_file_context(self):
         return 'event', self.event.id, 'editing', self.contrib.id, self.editable_type.name
 
+    def validate_file(self, file):
```
ThiefMaster, line 53: I'd move this into UploadFileMixin; empty files never make any sense

duartegalvao: I agree, should it go straight on the `if` inside `_process`? Since the `validate_file` is overriden sometimes

ThiefMaster: yeah

indico/modules/files/controllers.py
```diff
@@ -29,7 +29,7 @@ class UploadFileMixin:
 
     @use_kwargs({'file': fields.Field(required=True)}, location='files')
     def _process(self, file):
-        if not self.validate_file(file):
+        if file.getbuffer().nbytes == 0 or not self.validate_file(file):
```
ThiefMaster, line 32: This breaks for large files which are buffered to disk:

```pythontb
  File "/home/adrian/dev/indico/py3/src/indico/modules/files/controllers.py", line 32, in _process
    if file.getbuffer().nbytes == 0 or not self.validate_file(file):
  File "/home/adrian/dev/indico/py3/env/lib/python3.9/site-packages/werkzeug/datastructures/file_storage.py", line 148, in __getattr__
    return getattr(self.stream._file, name)
AttributeError: '_io.BufferedRandom' object has no attribute 'getbuffer'
```

indico/modules/files/controllers.py
```diff
@@ -42,6 +42,8 @@ def _save_file(self, file, stream):
         content_type = mimetypes.guess_type(file.filename)[0] or file.mimetype or 'application/octet-stream'
         f = File(filename=file.filename, content_type=content_type)
         f.save(context, stream)
+        if f.size == 0:
+            raise UnprocessableEntity
```
duartegalvao, line 46: this seems to work fine, we don't need to delete if we raise the exception before adding to the db right @ThiefMaster ?

PR #5768, opened 2023-05-11
Add API endpoints for custom contrib fields

ThiefMaster:


indico/modules/events/contributions/controllers/management.py
```diff
@@ -805,6 +818,34 @@ def _process_POST(self):
...
+
+    @use_kwargs({'friendly': fields.Bool(load_default=False)}, location='query')
+    def _process_GET(self, friendly):
+        return jsonify({f'custom_{fv.contribution_field.id}': (fv.friendly_data if friendly else fv.data)
+                        for fv in self.contrib.field_values})
+
+    def _process_PATCH(self):
+        # it'd be nicer to use marshmallow here but converting the custom field logic to support
+        # marshmallow would be a significant refactoring. let's do hat only when moving all the
+        # places that use those fields are moved to react so we can just replace wtforms with
```
tomasr8, line 833:
```suggestion
        # marshmallow would be a significant refactoring. let's do that only when moving all the
        # places that use those fields to react so we can just replace wtforms with
```

indico/web/rh.py
```diff
@@ -398,3 +398,13 @@ def allow_signed_url(rh):
     """
     rh._ALLOW_SIGNED_URL = True
     return rh
+
+
+def json_errors(rh):
+    """Specify that the RH should always jsonify errors.
+
+    This is meant for API endpoints and similar places where errors
+    should be returned as JSON and as not a formatted HTML error page.
```
tomasr8, line 407:
```suggestion
    should be returned as JSON and not as a formatted HTML error page.
```

PR #5771, opened 2023-05-12
Add setting to force track selection on judgment

duartegalvao:
This PR adds a setting to force track selection when accepting abstracts:

Screen Shot 2023-05-12 at 13 21 34

image

### DIFF
--- a/indico/modules/events/abstracts/forms.py
+++ b/indico/modules/events/abstracts/forms.py
@@ -163,6 +163,9 @@ class AbstractReviewingSettingsForm(IndicoForm):
                                                widget=SwitchWidget(),
                                                description=_('Enabling this allows track conveners to update the track '
                                                              'an abstract is part of.'))
+    force_track_selection = BooleanField(_('Force track selection on judgment'), widget=SwitchWidget(),
+                                         description=_('Enabling this makes the track selection mandatory when '
+                                                       'accepting an abstract.'))
     allow_comments = BooleanField(_('Allow comments'), widget=SwitchWidget(),
                                   description=_('Enabling this allows judges, conveners and reviewers to leave '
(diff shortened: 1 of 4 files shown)


indico/modules/events/abstracts/forms.py
```diff
@@ -277,6 +280,8 @@ def __init__(self, *args, **kwargs):
         elif not abstract.reviews:
             kwargs.setdefault('accepted_contrib_type', abstract.submitted_contrib_type)
         super().__init__(*args, **kwargs)
+        if self.event.cfa.force_track_selection:
+            self.accepted_track.validators.append(DataRequired())
```
ThiefMaster, line 284: Not sure if this correctly works for the server-side validation or just for displaying the field as required.

Better to use our `inject_validators()` util before the super init call

ThiefMaster:
Is bulk judgment exempt from this on purpose?

duartegalvao:
> Is bulk judgment exempt from this on purpose?

Since it's a manager-only action I don't think there's a need for it, since this feature is mostly to tackle cases where there are multiple review tracks and the judge accepts without selecting the track (since in most cases, it's already preselected)

I could implement it there to make it more consistent though.. Or maybe rephrase the setting to something like "Force judges to select tracks"?

ThiefMaster:
Being more explicit sounds good. The question is whether managers doing a judgment should also be required to choose a track or not though... currently they are and I think it's fine, because in an event where you enforce it, you probably always want at rack selected, but then a wording mentioning judges explicitly won't fit anymore.

Or we could just mention "Does not affect bulk judgment by managers" in the description of the setting and leave the logic as it is right now... ;)

ThiefMaster (approved):

PR #5772, opened 2023-05-12
Add block to multiple_items_widget jinja template

jbtwist:
The intention behind this is to create a way to inject a button to this table in the UN Plugin.

### DIFF
--- a/indico/web/templates/forms/multiple_items_widget.html
+++ b/indico/web/templates/forms/multiple_items_widget.html
@@ -21,9 +21,11 @@
             </thead>
             <tbody></tbody>
         </table>
-        <button type="button" id="{{ field.id }}-add-button" class="js-add-row i-button icon-plus">
-            {%- trans %}Add{% endtrans -%}
-        </button>
+        {% block action_buttons %}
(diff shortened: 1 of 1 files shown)


indico/web/templates/forms/multiple_items_widget.html
```diff
@@ -21,9 +21,13 @@
...
         </table>
-        <button type="button" id="{{ field.id }}-add-button" class="js-add-row i-button icon-plus">
-            {%- trans %}Add{% endtrans -%}
-        </button>
+        {% block action_buttons %}
+        <div class="action_buttons">
+            <button type="button" id="{{ field.id }}-add-button" class="js-add-row i-button icon-plus">
+                {%- trans %}Add{% endtrans -%}
+            </button>
+        </div>
```
ThiefMaster, line 29: - Indent the contents of the block
- Do you really need that div? If you already use the block to add custom stuff you can just wrap the `{{ super() }}` call in the div in your custom code.

jbtwist: You're correct, I don't need that div if I summon the super inside our div, thanks

PR #5776, opened 2023-05-15
Missing line to start the postgres server. There might be other place…

umer936:


docs/source/installation/production/debian/nginx.rst
```diff
@@ -38,6 +38,7 @@ Afterwards, make sure the services you just installed are running:
 .. code-block:: shell
 
     systemctl start postgresql.service redis-server.service
+    pg_ctlcluster 13 main start
```
ThiefMaster, line 41: I don't see why this would be needed - `systemctl start postgresql.service` should start it, no?

umer936: It doesn't in my case. I'm running this in a ubuntu:22.04 Docker container

ThiefMaster:
Some general advice: Do not use the setup guide inside Docker. At least not for a real production setup. We have Docker images and a Dockerfile which you can use if you (really) prefer deploying using Docker.

umer936:
Agreed. I did it in an Ubuntu 22.04 VM and had no issues compared to trying to get the Docker container to run.

PR #5778, opened 2023-05-17
Allow deleting editables

duartegalvao:
This PR allows event managers to fully (soft-)delete editables from contributions, allowing submitters to restart the editing process from scratch.

Screen Shot 2023-05-16 at 08 30 07

Progress:

- [x] Add `Editable.is_deleted` attribute
- [x] Add delete button to the editing timeline
- [x] Restrict the button to event managers
- [x] Add a redirect after deletion is successful
- [x] Add a changelog entry


indico/modules/events/editing/operations.py
```diff
@@ -90,12 +90,21 @@ def create_new_editable(contrib, type_, submitter, files, initial_state=InitialR
...
+    print('1')
     db.session.expire(editable)
+    print('2')
     for revision in editable.revisions:
+        print('3')
+        print(revision)
+        print(revision.files)
         for ef in revision.files:
+            print('4')
             ef.file.claimed = False
```
ThiefMaster, line 102: That's a bad idea. It will result in permanent deletion of the file. Files should remain claimed even when linked to something that's soft-deleted.

indico/modules/events/editing/controllers/base.py
```diff
@@ -120,7 +120,7 @@ def _process_args(self):
         self.editable_type = EditableType[request.view_args['type']]
         self.editable = (Editable.query
                          .with_parent(self.contrib)
-                         .filter_by(type=self.editable_type)
+                         .filter_by(type=self.editable_type, is_deleted=False)
```
ThiefMaster, line 123: I suggest changing the relationship's join criterion to already exclude deleted editables - that way you never need to filter by it when using with_parent or the relationship directly.

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -191,6 +192,19 @@ def _process(self):
         return '', 201
 
 
+class RHDeleteEditable(RHContributionEditableBase):
+    """Delete a contribution's editable."""
+
+    def _check_access(self):
+        RHContributionEditableBase._check_access(self)
+        if not self.contrib.can_manage(session.user):
```
ThiefMaster, line 200: I don't think this should be possible for someone who can just manage the contribution. Should require full event management privileges IMHO

indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -62,7 +63,7 @@ def _process(self, id, paper_id):
         files = []
         if id and self.contrib.editables:
             files = [file.file for e in self.contrib.editables
-                     if e.type == self.editable.type for file in e.revisions[-1].files]
+                     if not e.is_deleted and e.type == self.editable.type for file in e.revisions[-1].files]
```
ThiefMaster, line 66: not needed if you filter the relationship (see my other comment)

indico/modules/events/editing/controllers/backend/management.py
```diff
@@ -290,7 +290,7 @@ class RHEmailNotSubmittedEditablesSend(EmailRolesSendMixin, RHEditableTypeManage
 
     def get_recipients(self, roles):
         contribs = Contribution.query.with_parent(self.event).filter(
-            ~Contribution.editables.any(Editable.type == self.editable_type)
+            ~Contribution.editables.any(Editable.type == self.editable_type, ~Editable.is_deleted)
```
ThiefMaster, line 293: probably not needed when filtering the relationship itself

indico/modules/events/contributions/models/contributions.py
```diff
@@ -586,7 +587,7 @@ def can_submit_proceedings(self, user):
 
     def get_editable(self, editable_type):
         """Get the editable of the given type."""
-        return next((e for e in self.editables if e.type == editable_type), None)
+        return next((e for e in self.editables if not e.is_deleted and e.type == editable_type), None)
```
ThiefMaster, line 590: not needed when updating the relationship

indico/modules/events/editing/client/js/editing/timeline/DeleteEditable.jsx
```diff
@@ -0,0 +1,83 @@
...
+          contrib_id: contributionId,
+          type: editableType,
+        })
+      );
+    } catch (error) {
+      return handleAxiosError(error);
+    } finally {
+      setSubmitting(false);
+    }
+    location.href = editableTypeListURL({event_id: eventId, type: editableType});
```
ThiefMaster, line 54: In case this stays within the "Editing" SPA, use `import {useHistory} from 'react-router-dom';` and  `history.push(...)` for this redirect to avoid a(n unnecessary) full page reload

indico/modules/events/editing/operations.py
```diff
@@ -89,12 +89,13 @@ def create_new_editable(contrib, type_, submitter, files, initial_state=InitialR
     return editable
 
 
-def delete_editable(editable):
+def delete_editable(editable, soft=True):
     db.session.expire(editable)
-    for revision in editable.revisions:
-        for ef in revision.files:
-            ef.file.claimed = False
-    db.session.delete(editable)
```
ThiefMaster, line 97: In case of `not soft` this needs to stay.

indico/modules/events/editing/operations.py
```diff
@@ -89,12 +89,13 @@ def create_new_editable(contrib, type_, submitter, files, initial_state=InitialR
     return editable
 
 
-def delete_editable(editable):
+def delete_editable(editable, soft=True):
```
ThiefMaster, line 92:
```suggestion
def delete_editable(editable, *, soft=True):
```

indico/modules/events/editing/operations.py
duartegalvao: but this is correct, right? (the added code is below)

duartegalvao: aah, i see what you mean

indico/migrations/versions/20230515_1145_cb46beecbb93_add_editable_is_deleted.py
```diff
@@ -0,0 +1,32 @@
...
+down_revision = '5d05eda06776'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('editables', sa.Column('is_deleted', sa.Boolean(), nullable=False, server_default='false'),
+                  schema='event_editing')
+    op.alter_column('editables', 'is_deleted', server_default=None, schema='event_editing')
+    op.drop_constraint('uq_editables_contribution_id_type', 'editables', schema='event_editing', type_='unique')
```
tomasr8, line 23: Do we need the `type_='unique'`? Docs say it's only required for MySQL

indico/migrations/versions/20230515_1145_cb46beecbb93_add_editable_is_deleted.py
```diff
@@ -0,0 +1,32 @@
...
+    op.create_index(op.f('ix_uq_editables_contribution_id_type'), 'editables', ['contribution_id', 'type'], unique=True,
+                    schema='event_editing', postgresql_where=sa.text('NOT is_deleted'))
+
+
+def downgrade():
+    op.drop_column('editables', 'is_deleted', schema='event_editing')
+    op.drop_index('ix_uq_editables_contribution_id_type', table_name='editables', schema='event_editing')
+    op.create_unique_constraint('uq_editables_contribution_id_type', 'editables', ['contribution_id', 'type'],
```
tomasr8, line 31: I think the constraint name can also be `None` here - `op.create_unique_constraint(None, ...)`

indico/migrations/versions/20230515_1145_cb46beecbb93_add_editable_is_deleted.py
duartegalvao: thanks! stuff leftover from `--autogenerate` 😅

tomasr8:
Just leaving this here, so we don't forget. We're getting this react warning when deleting editables, though it doesn't happen always.

obrazek

Issue #5746, opened 2023-04-24
Internal booking notes


GovernmentPlates:
- UI mockups to share w/ Burotel admins
- Admin setting to toggle feature
- UI to add/edit note in booking dialog (w/o wasting space since few bookings will use it)
- Backend must only return this note to authorized people (this is very important!)

PR #5791, opened 2023-05-30
Allow managers to add internal notes to room bookings

GovernmentPlates:
Closes: #5746 

- [x] UI Mockups created and sent
- [x] Create migration
- [x] Add admin setting to toggle internal notes
- [x] Create, View and Edit internal notes (WIP: edit + fix over with correct perms - not `isAdmin` -> `canManage`)
- [x] Hookup setting w/ frontend
- [x] Implement `canManage` perms (+ admin override)
- [x] Ensure backend (API) returns notes to persons with the appropriate `canManage` permissions (room owners)
- [x] Remove notes field from edit logs (API)
- [x] Cleanup/Refactor
- [x] Update changelog

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -219,6 +219,7 @@ Improvements
   (:pr:`5784`)
 - Allow admins to change the password of local accounts (:pr:`5789`, thanks
   :user:`omegak`)
+- Allow room managers to add internal notes to bookings (:issue:`5746`, :pr:`5791`)
 
 Bugfixes
 ^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,7 @@ Improvements
   restrict them to speakers or registrants (:pr:`5670`, thanks :user:`kewisch`)
 - Make editing timeline display much more straightforward (:pr:`5674`)
 - Allow event managers to delete editables from contributions (:pr:`5778`)
+- Allow room managers to add internal notes to bookings (:issue:`5746`, :pr:`5791`)
 
 Bugfixes
 ^^^^^^^^
@@ -219,7 +220,6 @@ Improvements
   (:pr:`5784`)
 - Allow admins to change the password of local accounts (:pr:`5789`, thanks
   :user:`omegak`)
-- Allow room managers to add internal notes to bookings (:issue:`5746`, :pr:`5791`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/migrations/versions/20230424_1511_cd7038a37646_add_notes_column_to_room_bookings.py
+++ b/indico/migrations/versions/20230424_1511_cd7038a37646_add_notes_column_to_room_bookings.py
@@ -1,0 +1,25 @@
+"""Add internal_note column to room bookings
+
+Revision ID: cd7038a37646
+Revises: 5d05eda06776
+Create Date: 2023-04-24 15:11:55.407209
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = 'cd7038a37646'
+down_revision = '5d05eda06776'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('reservations', sa.Column('internal_note', sa.Text(), nullable=False, server_default=''), schema='roombooking')
+    op.alter_column('reservations', 'internal_note', server_default=None, schema='roombooking')
+
+
+def downgrade():
+    op.drop_column('reservations', 'internal_note', schema='roombooking')
--- a/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
+++ b/indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
@@ -281,6 +281,23 @@ class BookingDetails extends React.Component {
     </Message>
   );
 
+  renderNotes = internalNote => {
+    if (!internalNote) {
+      return null;
+    }
+    return (
+      <Message warning icon styleName="message-icon">
+        <Icon name="clipboard list" style={{marginRight: '0.4em'}} />
+        <Message.Content>
+          <Message.Header>
+            <Translate>Internal notes</Translate>
+          </Message.Header>
+          {internalNote}
+        </Message.Content>
+      </Message>
+    );
+  };
+
   _getRowSerializer = day => {
     const {
       booking: {room},
@@ -829,6 +846,7 @@ class BookingDetails extends React.Component {
         room,
         bookedForUser,
         bookingReason,
+        internalNote,
         editLogs,
         createdDt,
         createdByUser,
@@ -896,6 +914,7 @@ class BookingDetails extends React.Component {
                 <>
                   {bookedForUser && this.renderBookedFor(bookedForUser)}
                   {this.renderReason(bookingReason)}
+                  {room.canUserViewInternalNotes && this.renderNotes(internalNote)}
                   {isLinkedToObject && (
                     <LazyBookingObjectLink type={_.camelCase(link.type)} id={link.id} />
                   )}
--- a/indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
+++ b/indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
@@ -295,6 +295,21 @@ class BookingEditForm extends React.Component {
             disabled={submitSucceeded}
           />
         </Segment>
+
+        {room.canUserViewInternalNotes && (
+          <Segment inverted styleName="booking-notes-section">
+            <p style={{marginBottom: '0.5em'}}>
+              <Translate>
+                Internal notes about the booking that will only be shown to managers
+              </Translate>
+            </p>
+            <FinalTextArea
+              name="internalNote"
+              placeholder={Translate.string('(Optional)')}
+              disabled={submitSucceeded}
+            />
+          </Segment>
+        )}
       </Form>
     );
   }
--- a/indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
+++ b/indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
@@ -297,7 +297,7 @@ class BookingEditForm extends React.Component {
         </Segment>
 
         {room.canUserViewInternalNotes && (
-          <Segment inverted styleName="booking-notes-section">
+          <Segment inverted styleName="internal-notes-segment">
             <p style={{marginBottom: '0.5em'}}>
               <Translate>
                 Internal notes about the booking are only visible to room managers.
--- a/indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
+++ b/indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
@@ -300,14 +300,10 @@ class BookingEditForm extends React.Component {
           <Segment inverted styleName="booking-notes-section">
             <p style={{marginBottom: '0.5em'}}>
               <Translate>
-                Internal notes about the booking that will only be shown to managers
+                Internal notes about the booking are only visible to room managers.
               </Translate>
             </p>
-            <FinalTextArea
-              name="internalNote"
-              placeholder={Translate.string('(Optional)')}
-              disabled={submitSucceeded}
-            />
+            <FinalTextArea name="internalNote" disabled={submitSucceeded} />
           </Segment>
         )}
       </Form>
--- a/indico/modules/rb/client/js/modules/bookRoom/BookRoomModal.jsx
+++ b/indico/modules/rb/client/js/modules/bookRoom/BookRoomModal.jsx
@@ -629,6 +629,33 @@ class BookRoomModal extends React.Component {
                     fprops.form.mutators
                   )}
               </Form>
+
+              {room.canUserViewInternalNotes && (
+                <Form
+                  id="book-notes-form"
+                  onSubmit={fprops.handleSubmit}
+                  style={{paddingTop: '1em'}}
+                >
+                  <Segment inverted styleName="booking-notes-section">
+                    <h3 style={{marginBottom: '0.5em'}}>
+                      <Icon name="clipboard list" />
+                      <Translate>Notes</Translate>
+                    </h3>
+                    <p>
+                      <Translate>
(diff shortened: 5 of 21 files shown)


indico/migrations/versions/20230424_1511_cd7038a37646_add_notes_column_to_room_bookings.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'cd7038a37646'
+down_revision = '5d05eda06776'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('reservations', sa.Column('notes', sa.Text(), nullable=True), schema='roombooking')
```
ThiefMaster, line 20: We usually store an empty string (default) and make columns not nullable. In the revision you need to create it with `serer_default=''` and then drop the server_default again right below (there are a bunch of examples if you search for `server_default=None`)

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -281,6 +281,22 @@ class BookingDetails extends React.Component {
     </Message>
   );
 
+  renderNotes = notes => {
+    if (notes) {
```
ThiefMaster, line 285:
```js
if (!notes) {
  return null;
}
```

saves you some indentation for everything below

indico/modules/rb/models/reservations.py
```diff
@@ -274,6 +278,10 @@ def is_archived(self):
     def is_repeating(self):
         return self.repeat_frequency != RepeatFrequency.NEVER
 
+    @hybrid_property
```
ThiefMaster, line 281: This should be a normal property; it wouldn't work as-is in hybrid mode anyway (`is` and `and` cannot be overridden by a class even with dunders or other magic so you'd need an explicit .expression method for it).

And FWIW, the fact that you added this method shows exactly why we generally don't use nullable string columns. ;)
There's a good chance you don't need it anymore in the end, but if you do, the whole body can be simplified to `return bool(self.notes)`

indico/modules/rb/schemas.py
```diff
@@ -356,6 +357,7 @@ class CreateBookingSchema(mm.Schema):
     link_id = fields.Int()
     link_back = fields.Bool(load_default=False)
     admin_override_enabled = fields.Bool(load_default=False)
+    notes = fields.String(data_key='notes', validate=validate.Length(min=3), required=False)
```
ThiefMaster, line 360: maybe this fits nicer above right after the `booking_reason`?

indico/modules/rb/models/reservations.py
GovernmentPlates: Ah, I see. I wrote this property for future use in the backend, but after seeing your previous comment on the migration, I'll probably just make it return the bool for now (otherwise, I'll just remove it entirely).

indico/modules/rb/controllers/backend/bookings.py
```diff
@@ -308,6 +308,7 @@ def _process(self, args):
             'end_dt': args['end_dt'],
             'repeat_frequency': args['repeat_frequency'],
             'repeat_interval': args['repeat_interval'],
+            'notes': args['notes'],
```
ThiefMaster, line 311: This should only be added if notes are enabled and the user can manage the roomt

indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -295,6 +295,21 @@ class BookingEditForm extends React.Component {
             disabled={submitSucceeded}
           />
         </Segment>
+
+        {room.canUserViewInternalNotes && (
+          <Segment inverted styleName="booking-notes-section">
+            <p style={{marginBottom: '0.5em'}}>
+              <Translate>
+                Internal notes about the booking that will only be shown to room booking managers
```
ThiefMaster, line 303:
```suggestion
                Internal notes about the booking that will only be shown to managers
```

or "to room managers", but omitting "room" makes it work more nicely in contexts like burotel where we book desks and not rooms

indico/modules/rb/client/js/common/bookings/BookingDetails.jsx
```diff
@@ -281,6 +281,23 @@ class BookingDetails extends React.Component {
...
+  renderNotes = notes => {
+    if (!notes) {
+      return null;
+    }
+    return (
+      <Message warning icon styleName="message-icon">
+        <Icon name="clipboard list" style={{marginRight: '0.4em'}} />
+        <Message.Content>
+          <Message.Header>
+            <Translate>Booking notes</Translate>
```
ThiefMaster, line 293: "Internal notes" might be a good title here as well

indico/modules/rb/api.py
```diff
@@ -185,6 +185,7 @@ def _getParams(self):
             'from': self._fromDT,
             'to': self._toDT
         }
+
```
ThiefMaster, line 188: 👾

indico/migrations/versions/20230424_1511_cd7038a37646_add_notes_column_to_room_bookings.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'cd7038a37646'
+down_revision = '5d05eda06776'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('reservations', sa.Column('notes', sa.Text(), nullable=False, server_default=''), schema='roombooking')
```
ThiefMaster, line 20: You need to `alter_column()` with `server_default=None` below since we don't want to keep that one around - we just need it during creation to populate existing rows.

CHANGES.rst
```diff
@@ -77,6 +77,7 @@ Improvements
   (:pr:`5784`)
 - Allow admins to change the password of local accounts (:pr:`5789`, thanks
   :user:`omegak`)
+- Allow administrators to add internal notes to room bookings (:issue:`5746`, :pr:`5791`)
```
ThiefMaster, line 80:
```suggestion
- Allow room managers to add internal notes to bookings (:issue:`5746`, :pr:`5791`)
```

ThiefMaster:
Tip for all those WIP commits: `git rebase -i master` and mark all but the first commit as `fixup` instead of `pick`.
For the first commit, change `pick` to `reword` and when you get the second editor, put a nice commit message like "Add internal room booking notes"

indico/migrations/versions/20230424_1511_cd7038a37646_add_notes_column_to_room_bookings.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'cd7038a37646'
+down_revision = '5d05eda06776'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('reservations', sa.Column('notes', sa.Text(), nullable=False, server_default=None), schema='roombooking')
```
tomasr8, line 20:
```suggestion
    op.add_column('reservations', sa.Column('notes', sa.Text(), nullable=False, server_default=''), schema='roombooking')
    op.alter_column('reservations', 'notes', server_default=None, schema='roombooking')
```

indico/migrations/versions/20230424_1511_cd7038a37646_add_notes_column_to_room_bookings.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def downgrade():
+    op.drop_column('reservations', 'notes', server_default=None, schema='roombooking')
```
tomasr8, line 24:
```suggestion
    op.drop_column('reservations', 'notes', schema='roombooking')
```

indico/modules/rb/schemas.py
```diff
@@ -352,6 +358,7 @@ class CreateBookingSchema(mm.Schema):
     room_id = fields.Int(required=True)
     booked_for_user = Principal(data_key='user', allow_external_users=True)
     booking_reason = fields.String(data_key='reason', validate=validate.Length(min=3), required=True)
+    notes = fields.String(data_key='notes', validate=validate.Length(min=3), required=False)
```
tomasr8, line 361: is the `data_key` needed?

indico/modules/rb/models/reservations.py
```diff
@@ -604,6 +610,10 @@ def modify(self, data, user):
                 if field == 'booked_for_user':
                     old = self.booked_for_name
                     new = self.booked_for_name = data[field].full_name
+                # If notes are being updated then we don't bother updating the edit log with this
+                if field == 'notes':
+                    setattr(self, field, data[field])
+                    continue
                 # Apply the change
                 setattr(self, field, data[field])
```
tomasr8, line 618: Might be simpler like this?
```suggestion
                # Apply the change
                setattr(self, field, data[field])
                # Don't record changes to notes
                if field == 'notes':
                    continue
```

ThiefMaster: Or further down below: 

```python
if field == 'notes':
    continue
# Record change for history entry
```

indico/modules/rb/schemas.py
GovernmentPlates: Nope, I just was following what the `booking_reason` was doing when I first implementing the `notes` system in general, as its pretty similar.

indico/modules/rb/models/reservations.py
GovernmentPlates: Nice :)

indico/modules/rb/schemas.py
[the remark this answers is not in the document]
ThiefMaster: or just remove the length validator. if someone wants to put `ok` or whatever in it, who cares ;)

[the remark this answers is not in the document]
tomasr8: even better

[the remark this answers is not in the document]
GovernmentPlates: I'm guessing I would have to remove that JS `if` validator on the `BookingEdit.jsx` component too, right?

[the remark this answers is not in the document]
tomasr8: Yup

indico/modules/rb/schemas.py
```diff
@@ -251,6 +251,12 @@ def _get_permissions(self, booking):
             admin_permissions = {x: getattr(booking, x)(session.user) for x in methods}
         return {'user': user_permissions, 'admin': admin_permissions}
 
+    @post_dump
+    def _hide_sensitive_data(self, data, **kwargs):
+        if not self.context.get('include_notes'):
```
tomasr8, line 256: You can do the check directly here w/o passing the context. Alternatively you could also use `ReservationDetailsSchema(exclude=['notes'])` to avoid the `post_dump` altogether

```suggestion
    @post_dump(pass_original=True)
    def _hide_sensitive_data(self, data, booking, **kwargs):
        if not booking.room.can_manage(session.user):
```

ThiefMaster: The context was my suggestion because I don't really like the idea of accessing the session etc in a schema. But later I noticed that we do so for some other things as well, so maybe it's not THAT ugly to do it in this particular schema...

GovernmentPlates: Yeah, this seems like a much nicer way to handle perms - added :)

indico/modules/rb/client/js/modules/bookRoom/BookRoomModal.jsx
```diff
@@ -629,6 +629,34 @@ class BookRoomModal extends React.Component {
...
                   )}
               </Form>
+
+              {room.canUserViewInternalNotes && (
+                <Form
+                  id="book-notes-form"
+                  onSubmit={fprops.handleSubmit}
+                  style={{paddingTop: '1em'}}
+                >
+                  <Segment inverted styleName="booking-notes-section">
```
tomasr8, line 639: Does this need to be in a separate form, or could it go inside `#book-room-form`?

GovernmentPlates: I would rather keep it separate.

indico/migrations/versions/20230424_1511_cd7038a37646_add_notes_column_to_room_bookings.py
```diff
@@ -0,0 +1,25 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'cd7038a37646'
+down_revision = '5d05eda06776'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('reservations', sa.Column('notes', sa.Text(), nullable=False, server_default=''), schema='roombooking')
```
ThiefMaster, line 20: Just wondering, wouldn't `internal_comment` be much clearer for this?

indico/modules/rb/util.py
```diff
@@ -178,9 +178,10 @@ def serialize_booking_details(booking):
     from indico.modules.rb.operations.bookings import (get_booking_occurrences, get_existing_room_occurrences,
                                                        group_blockings, group_nonbookable_periods)
     from indico.modules.rb.operations.misc import get_rooms_nonbookable_periods, get_rooms_unbookable_hours
-    from indico.modules.rb.schemas import reservation_details_schema, reservation_occurrences_schema_with_permissions
+    from indico.modules.rb.schemas import ReservationDetailsSchema, reservation_occurrences_schema_with_permissions
 
-    attributes = reservation_details_schema.dump(booking)
```
ThiefMaster, line 183: This change in here isn't really needed anymore now

indico/migrations/versions/20230424_1511_cd7038a37646_add_notes_column_to_room_bookings.py
GovernmentPlates: How about `internal_note`?

ThiefMaster: fine as well

ThiefMaster (approved):

PR #5794, opened 2023-05-31
Show an actual recipient on the emails' preview

duartegalvao:
This PR makes the email dialogs' preview screen use an actual recipient's data on the placeholders instead of the event's creator.

Note: roles of recipients are not taken into account on the `EmailContribAbstractRoles` due to the added complexity on the client-side (can implement if deemed necessary)

### DIFF
--- a/indico/modules/events/persons/controllers.py
+++ b/indico/modules/events/persons/controllers.py
@@ -265,25 +265,6 @@ class RHPersonsList(RHPersonsBase):
                                                has_predefined_affiliations=Affiliation.query.has_rows())
 
 
-class RHEmailEventPersonsPreview(RHManageEventBase):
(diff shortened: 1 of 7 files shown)


indico/modules/events/persons/controllers.py
```diff
@@ -313,6 +294,24 @@ def _find_role_members(self, role_ids):
         return itertools.chain.from_iterable(role.members for role in query)
 
 
+class RHEmailEventPersonsPreview(RHEmailEventPersonsBase):
```
ThiefMaster, line 297: Why move it around within the file? Makes it harder to review the actual changes... ;)

duartegalvao: because one of the changes is that it now inherits from a class that was below 😅

PR #5800, opened 2023-06-05
Add 'max_choices' option to multi choice fields

tomasr8:
Lets you configure the maximum number of options that can be selected in a multi-choice field, e.g., there are 5 options but only 2 can be selected at the same time. Sort of an extension of the single choice field which allows only one option at a time.

obrazek

### DIFF
--- a/indico/modules/events/registration/client/js/form/fields/MultiChoiceInput.jsx
+++ b/indico/modules/events/registration/client/js/form/fields/MultiChoiceInput.jsx
@@ -29,6 +29,8 @@ function MultiChoiceInputComponent({
   existingValue,
   value,
   onChange,
+  onFocus,
+  onBlur,
   disabled,
   choices,
   withExtraSlots,
@@ -38,7 +40,13 @@ function MultiChoiceInputComponent({
   const management = useSelector(getManagement);
   const currency = useSelector(getCurrency);
 
+  const markTouched = () => {
+    onFocus();
+    onBlur();
+  };
+
   const makeHandleChange = choice => () => {
+    markTouched();
     if (value[choice.id]) {
       onChange(_.omit(value, choice.id));
     } else {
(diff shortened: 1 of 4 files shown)


indico/modules/events/registration/client/js/form/fields/MultiChoiceInput.jsx
```diff
@@ -188,6 +189,16 @@ export default function MultiChoiceInput({
       placesUsed={placesUsed}
       existingValue={existingValue}
       isEqual={_.isEqual}
+      validate={value => {
+        if (maxChoices && _.keys(value).length > maxChoices) {
+          return PluralTranslate.string(
+            'At most one option can be selected',
```
ThiefMaster, line 195:
```suggestion
            'At most {n} option can be selected',
```

iirc this works better for translations in languages with no plural forms

indico/modules/events/registration/fields/choices.py
```diff
@@ -272,6 +285,19 @@ def _format_item(uuid, number_of_slots):
 
         return ', '.join(choices) if for_humans or for_search else choices
 
+    def get_validators(self, existing_registration):
+        def _check_max_choices(new_data):
+            """If `max_choices` is set, ensure that it is less than or equal to the total number of choices."""
```
ThiefMaster, line 290: I think a normal comment makes more sense here than a docstring

indico/modules/events/registration/fields/choices.py
```diff
@@ -272,6 +285,19 @@ def _format_item(uuid, number_of_slots):
...
+            if not new_data:
+                return
+
+            max_choices = self.form_item.data['max_choices']
+            if max_choices and len(new_data.keys()) > max_choices:
```
ThiefMaster, line 295: is `new_data` a dict? then `.keys()` isn't needed

PR #5802, opened 2023-06-06
Show which files were modified on each revision

duartegalvao:
This PR adds an indication to show which files were added or modified on each editing timeline revision.

image


duartegalvao:
I set it back to draft because I noticed some problems with the PR (probably shouldn't be mutating the files list as I am currently doing)

ThiefMaster (approved):

PR #5804, opened 2023-06-08
Fix error when cat not found in info-from page

SegiNyn:
When you visit ~ http://127.0.0.1:9000/category/<category_id>/info-from and the category does not exist, it raises an error


ThiefMaster:
Sorry, but what's the problem with the error? You are not supposed to call this with an invalid category ID, and the whole idea behind `.one()` is that it raises an error if no rows are found...

SegiNyn:
This endpoint can still be directly accessed by anyone and they can try different category ids that are not correct . `.one()` raises an exception when no rows are found which leads to unhandled exception. If you prefer the exception to be raised can we put it in a try block?

ThiefMaster:
There are many places that trigger exceptions when you pass crappy data. I think the exception from `.one()` isn't really an issue. 

Anyway, feel free to replace the `one()` with `first_or_404()`. That way you have a NotFound exception w/o having to write extra code.

SegiNyn:
Okay, thanks!

OmeGak:
> There are many places that trigger exceptions when you pass crappy data.

In this case, it's data that comes from `request.view_args` so a 404 seems a more reasonable response than an unhandled exception.

OmeGak:
Before merging this one, it may be worth to resolve another unhandled exception that we have been getting from the same RH:

```
2023-06-08 10:15:24,137  cbabb555594349c1  -       indico.flask - ERROR errors.py:98 -- 'str' object has no attribute 'get'

  File ".../indico/env/lib/python3.9/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File ".../indico/env/lib/python3.9/site-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File ".../indico/src/indico/indico/web/flask/util.py", line 79, in wrapper
    return obj().process()
  File ".../indico/src/indico/indico/web/rh.py", line 285, in process
    res = self._do_process()
  File ".../indico/src/indico/indico/web/rh.py", line 256, in _do_process
    rv = self._process()
  File ".../indico/src/indico/indico/modules/categories/controllers/display.py", line 159, in _process
    excluded_ids = set(request.json.get('exclude', set())) if request.json else set()
AttributeError: 'str' object has no attribute 'get'

{'data': {'get': {},
          'headers': {'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',
                      'Accept-Encoding': 'gzip,deflate,br',
                      'Connection': 'close',
                      'Content-Length': '35',
                      'Content-Type': 'application/json',
                      'Cookie': 'outdatedbrowser=hide; '
                                'indico_session=...',
                      'Host': 'indico.un.org',
                      'User-Agent': 'User-Agent: Mozilla/4.0 (compatible; MSIE '
                                    '7.0; Windows NT 5.1; 360SE)',
                      'X-Csrf-Token': '...',
                      'X-Forwarded-For': '...',
                      'X-Forwarded-Host': 'indico.un.org',
                      'X-Forwarded-Proto': 'https',
                      'X-Forwarded-Ssl': 'on',
                      'X-Real-Ip': '...'},
          'json': '+response.write(9019501*9088080)+',
          'post': {},
          'url': {'category_id': 100404}},
 'endpoint': 'categories.info_from',
 'id': 'cbabb555594349c1',
 'ip': '...',
 'method': 'POST',
 'referrer': None,
 'rh': 'RHReachableCategoriesInfo',
 'time': '2023-06-08T10:15:24.149413',
 'url': 'https://indico.un.org/category/100404/info-from',
 'user': None,
 'user_agent': 'User-Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1; '
               '360SE)'}
```

ThiefMaster:
Feel free to switch to `@use_kwargs({...})` with a schema :)

SegiNyn:
Hi @ThiefMaster. I'm not sure I understand your comment "with a schema". Can I do this? 

```
@use_kwargs({
        'excluded_ids': fields.List(fields.Int(), data_key='exclude', load_default=[]),
    })
    def _process(self, excluded_ids):
        excluded_ids = set(excluded_ids) if excluded_ids else set()
```

ThiefMaster:
Yes, that's exactly what i meant :)

SegiNyn:
Perfect! Thanks!

SegiNyn:
All done now, thank you!

indico/modules/categories/controllers/display.py
```diff
@@ -158,8 +158,11 @@ def _get_reachable_categories(self, id_, excluded_ids):
                 .options(*RHCategoryInfo._category_query_options)
                 .all())
 
-    def _process(self):
-        excluded_ids = set(request.json.get('exclude', set())) if request.json else set()
+    @use_kwargs({
+        'excluded_ids': fields.List(fields.Int(), data_key='exclude', load_default=[]),
```
ThiefMaster, line 162: I'd use `None` here to avoid passing a mutable default.

ThiefMaster (approved):

SegiNyn:
Sorry, wrong comma

PR #5808, opened 2023-06-08
Fix default value for type parameter in RHAPISearch

tomako:
The default value for a List field should be list otherwise it causes trouble later as the search providers expect list.

### DIFF
--- a/indico/modules/search/controllers.py
+++ b/indico/modules/search/controllers.py
@@ -46,7 +46,7 @@ class RHAPISearch(RH):
     @use_kwargs({
         'page': fields.Int(load_default=None),
         'q': fields.String(required=True),
-        'type': fields.List(EnumField(SearchTarget), load_default=None),
+        'type': fields.List(EnumField(SearchTarget), required=True),
         'admin_override_enabled': fields.Bool(
             load_default=False,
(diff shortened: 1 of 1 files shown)


indico/modules/search/controllers.py
```diff
@@ -46,7 +46,7 @@ class RHAPISearch(RH):
     @use_kwargs({
         'page': fields.Int(load_default=None),
         'q': fields.String(required=True),
-        'type': fields.List(EnumField(SearchTarget), load_default=None),
+        'type': fields.List(EnumField(SearchTarget), load_default=[]),
```
ThiefMaster, line 49: Passing a mutable default feels a bit dangerous here. I'd rather add `if type is None: type = []` below or simply make it required if we always pass a value for it

tomako: Since I'm not sure whether it is required in all cases I would stay on the safe side and check if it's None. 
We got a request not without the 'type' param but in a wrong format ('type[]=...') and that was the case when the default format was used. I was not able to reproduce but wanted to fix this kind of issue.

ThiefMaster: At least the core indico code always passes a type there. So making it required would be safe

tomako: Alright, then go on that way.

PR #5810, opened 2023-06-09
Avoid automatic Email for CSS upload error

openprojects:
Everytime a css with invalid chars is uploaded, the error is logged and an Email is sent.
This code avoids the automatic Email that fulfills our mailbox.


ThiefMaster:
I'm surprised this happens so often for you... don't people typically upload valid ASCII or UTF8 files?

PS: I honestly don't understand why you guys still use emails for errors and not something like Sentry...

openprojects:
> I'm surprised this happens so often for you... don't people typically upload valid ASCII or UTF8 files?
> 
> PS: I honestly don't understand why you guys still use emails for errors and not something like Sentry...

Yes, it happens quite often :(
Sentry: that's a question for @giusedb or @OmeGak

PR #5812, opened 2023-06-14
Docs: add RegistrationForm model docs

pferreir:
It only took us many years to get there, but I'm finally documenting `RegistrationForm` and friends. It's always very hard to wrap your head about this part and I think this will be a favour to future generations of developers. Also adding mermaid support to the tech docs, which could eventually replace the weird UML stuff we're using.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,21 @@
-requirements_file: docs/requirements.txt
+version: 2
+
 build:
-  image: testing
+  os: ubuntu-22.04
+  tools:
+    python: "3.10"
+    nodejs: "19"
+  jobs:
+    post_install:
+      - npm install -g @mermaid-js/mermaid-cli
+
+sphinx:
+  configuration: docs/source/conf.py
+
 python:
-  version: 3.9
-  setup_py_install: false
-  pip_install: true
+  install:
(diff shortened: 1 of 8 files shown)


.readthedocs.yml
```diff
@@ -1,7 +1,21 @@
...
 python:
-  version: 3.9
-  setup_py_install: false
-  pip_install: true
+  install:
+    - requirements: docs/requirements.txt
+    - method: pip
+      path: .
+      extra_requirements:
+        - dev
```
ThiefMaster, line 21: Is that needed? I don't think we have any extras defined there, and the indico package itself shouldn't be installed (I think)

ThiefMaster: ah hm, i thought we had skipped this, but maybe that was in the past when we still had this ZODB crap which failed to install in the RTD env. so nevermind, I think `[dev]` actually makes sense in that case.

ThiefMaster:
just spotted this in conf.py  btw: `copyright = '2021, Indico Team'` - might be nice to just use `date.today().year` there

pferreir:
> date.today().year

I think I'll send another PR to fix a few other things in the docs (e.g. warning messages)

PR #5825, opened 2023-06-27
Js updates 1

bpedersen2:
Regenerate package-lock.json to get the latest update core-js to latest supported version.


ThiefMaster:
and everything works fine? the last time i tried to touch core-js (years ago) there was a huge load of errors...

bpedersen2:
Yes, with older versions I also  encountered some problems, but the latest tests are actually working (one sympton was, that  the user search dialogs were in-operable). This  did not occur in my test now.

bpedersen2:
Seems the test suite uses a much older nodejs ( 14.x ) while I did it with node 18.x LTS.

ThiefMaster:
let's bump that to node 18 (current LTS)

Issue #5806, opened 2023-06-08
Recurring bookings with "byweekday" repetition


ThiefMaster:
- Support "byweekday" rules for recurring bookings
- Add new UI widget to select weekdays

An example for such a recurring booking would be "weekly, monday+tuesday". Right now the only way to do this would be making two separate bookings, which is not user-friendly.

Related tasks for the Burotel CERN plugin (these do NOT need to be done at the same time, but depend on this being done in the core):

- Allow weekly repetition (currently: only daily repetition)
- Verify correct ADaMS sync (should still cover the whole period)

ThiefMaster:
As discussed offline, this should be an extension to the "recurring" booking. Below the "every [n] [weeks|months]" there would be a button group (or similar) that shows the 7 week days, with the one of the start day always being selected (since it would be pointless to exclude it). You can then add additional weekdays if you want, and those would be included in the booking.

On the server side this would simply affect the rrule. You likely need a new column (postgres array of strings?) to store the weekdays. for this one I think nullable=true makes sense (and keeping it null for single/daily bookings, maybe a check constraint to enforce this).

PR #5829, opened 2023-06-28
Add support for bookings with 'by weekday' repetition

GovernmentPlates:
Closes: #5806 

- [x] Implement weekday picker component (i18n, validation etc.)
- [x] Add component to booking landing, search filter options and edit booking modal (test)
- [x] Pass selected days & data to backend
- [x] Create migration for storing selected days (postgres string array, keep nullable, check constraints)
- [x] Modify existing backend rules for booking creation
- [x] Write test cases for `recurrence_weekdays`
- [x] Fix existing unit tests
- [x] Fix issues from testing
- [x] Cleanup/Refactor
- [x] Update changelog


tomasr8:
It might be easier to use hooks rather than classes for new React components, but I think room booking already uses class components in most places so sticking w/ classes might be better here?

ThiefMaster:
I'd still use function components for anything new in order to not add even more class components...

indico/modules/rb/controllers/backend/common.py
```diff
@@ -23,6 +23,7 @@
     'end_dt': fields.DateTime(),
     'repeat_frequency': EnumField(RepeatFrequency),
     'repeat_interval': fields.Int(load_default=0),
+    'recurrence_weekdays': fields.List(fields.Str(), load_default=[]),
```
ThiefMaster, line 26: You can use `validate=validate.OneOf([...])` here to filter out invalid values early

indico/modules/rb/client/js/common/bookings/BookingEditCalendar.module.scss
```diff
@@ -51,3 +53,7 @@
     }
   }
 }
+
+.weekday {
```
GovernmentPlates, line 57: Note: I'll get around to moving these repeated `.weekday` styles into a generic stylesheet that can imported elsewhere.

indico/modules/rb/client/js/components/App.module.scss
```diff
@@ -103,3 +103,7 @@
     min-height: calc(100vh - #{$menu-height} - var(--extra-bars, 0) * 50px - 50px);
   }
 }
+
+:global(.ui.dimmer) {
+  backdrop-filter: blur(0.5rem);
+}
```
ThiefMaster, line 109: Please don't include this here, it's basically overriding SUI behavior and "re-implementing" a feature SUI already provides  if we want this behavior, the `<Modal>` components should set `dimmer="blurring"`.

indico/modules/rb/client/js/util.jsx
```diff
@@ -234,6 +237,67 @@ export function serializeRecurrenceInfo({type, number, interval}) {
...
+export function renderRecurrenceWeekdays(weekdays) {
+  const order = ['mon', 'tue', 'wed', 'thu', 'fri', 'sat', 'sun'];
+  const weekdaysMap = {
+    mon: 'Monday',
+    tue: 'Tuesday',
+    wed: 'Wednesday',
+    thu: 'Thursday',
+    fri: 'Friday',
+    sat: 'Saturday',
+    sun: 'Sunday',
```
ThiefMaster, line 254: needs i18n (use momentjs)

indico/modules/rb/client/js/util.jsx
```diff
@@ -234,6 +237,67 @@ export function serializeRecurrenceInfo({type, number, interval}) {
...
+    );
+  }
+
+  if (weekdays.length === 2) {
+    return (
+      <Translate>
+        Recurs every{' '}
+        <Param
+          name="weekdays"
+          value={`${weekdaysMap[weekdays[0]]} and ${weekdaysMap[weekdays[1]]}`}
```
ThiefMaster, line 278: `and` must be outside the param

indico/modules/rb/client/js/util.jsx
```diff
@@ -234,6 +237,67 @@ export function serializeRecurrenceInfo({type, number, interval}) {
...
+  }
+
+  // join weekdays with commas and replace last comma with 'and'
+  // (i.e. ['mon', 'wed', 'tue'] -> 'Monday, Tuesday and Wednesday')
+  const weekdaysString = weekdays
+    .map(weekday => weekdaysMap[weekday])
+    .join(', ')
+    .replace(
+      `, ${weekdaysMap[weekdays[weekdays.length - 1]]}`,
+      ` and ${weekdaysMap[weekdays[weekdays.length - 1]]}`
```
ThiefMaster, line 291: no! :D

create an array with all but the last weekdays and then join those with the `,` and then use the same logic as for 2 weekdays to show that and the last one

indico/modules/rb/client/js/util.jsx
```diff
@@ -234,6 +237,67 @@ export function serializeRecurrenceInfo({type, number, interval}) {
...
+    sat: 'Saturday',
+    sun: 'Sunday',
+  };
+
+  if (!weekdays || weekdays.length === 0) {
+    return null;
+  }
+
+  // order weekdays by their index in the order array
+  weekdays = [...new Set(weekdays)].sort((a, b) => order.indexOf(a) - order.indexOf(b));
```
ThiefMaster, line 262: probably we never get duplicates here so not sure if the dedup is needed

indico/modules/rb/client/js/util.jsx
```diff
@@ -234,6 +237,67 @@ export function serializeRecurrenceInfo({type, number, interval}) {
...
+    mon: 'Monday',
+    tue: 'Tuesday',
+    wed: 'Wednesday',
+    thu: 'Thursday',
+    fri: 'Friday',
+    sat: 'Saturday',
+    sun: 'Sunday',
+  };
+
+  if (!weekdays || weekdays.length === 0) {
```
ThiefMaster, line 257:
```suggestion
  if (weekdays.length === 0) {
```

unless you have cases where we really want to pass `null` or similar

indico/modules/rb/util.py
```diff
@@ -334,6 +334,7 @@ def get_prebooking_collisions(reservation):
 
 
 def check_impossible_repetition(data):
+    """ Check if the repetition is possible with the given data """
```
ThiefMaster, line 337: no surrounding spaces please ;)

also, this docstring is just a more verbose form (did github copilot write it? :P) of the function name...

how about something like this instead:

```
"""
Check for broken repetition data

This checks that a repetition using weekdays has a date range
containing at least one of those weekdays.
"""
```

indico/modules/rb/util.py
```diff
@@ -329,3 +331,21 @@ def get_prebooking_collisions(reservation):
     from indico.modules.rb.models.reservation_occurrences import ReservationOccurrence
     valid_occurrences = reservation.occurrences.filter(ReservationOccurrence.is_valid).all()
     return ReservationOccurrence.find_overlapping_with(reservation.room, valid_occurrences, reservation.id).all()
+
+
+def check_impossible_repetition(data):
+    """
+    Check for broken repetition data
```
ThiefMaster, line 338:
```suggestion
    """Check for broken repetition data
```

https://peps.python.org/pep-0257/

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

> The summary line may be on the same line as the opening quotes or on the next line.

We generally do "on the same line" - the only (somewhat rare) exception are docstrings where we don't really have a title but just a body; in that case we don't start in the same line.

GovernmentPlates: I see - so would this be alright?

```suggestion
    """Check for broken repetition data
    
    ...detailed explanation goes here"""
```

ThiefMaster: no, the title line usually ends with a `.` and the closing `"""` always goes on its own line (unless the whole docstring is single-line)

(my suggestion is already changing enough to match the style (nearly) all other docstrings in indico use...)

indico/modules/rb/models/reservations.py
```diff
@@ -155,6 +159,11 @@ def __table_args__(cls):
         nullable=False,
         default=0
     )  # 1, 2, 3, etc.
+    recurrence_weekdays = db.Column(
+        ARRAY(db.String),
```
tomasr8, line 163: I'm wondering if this could be an enum?

GovernmentPlates: As discussed: this would probably be a better candidate for a future PR/improvement task (one of the current downsides being that this PR is already set-up to just accept the weekday shorthands in lots of different places, whereas the upside being that we can remove the constraint for recurrence_weekdays).

indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -239,19 +272,22 @@ class BookingEditForm extends React.Component {
...
-                requireOneSelected
-                onChange={newWeekdays => {
-                  const newRecurrence = {...recurrence, weekdays: newWeekdays};
-                  onBookingPeriodChange(dates, timeSlot, newRecurrence);
-                }}
-              />
-            </>
-          )}
+          <div
+            style={{
```
ThiefMaster, line 276: `style={... ? {display: 'none'} : {}}`

indico/modules/rb/util.py
```diff
@@ -348,3 +348,26 @@ def check_impossible_repetition(data):
         return
     if not any(ReservationOccurrence.iter_start_time(start_dt, end_dt, repetition, recurrence_weekdays)):
         raise ExpectedError(_('The chosen date range does not include any of the weekdays you specified.'))
+
+
+def check_empty_candidates(candidates):
+    """Check for empty candidates
```
ThiefMaster, line 354: `"""Check for empty candidates.`

indico/modules/rb/util.py
```diff
@@ -348,3 +348,26 @@ def check_impossible_repetition(data):
...
+
+    This checks that a created time-series has at least one occurrence to filter by (similar to
+    :meth:`check_impossible_repetition` but without the need to pass all the data).
+    """
+    if not candidates:
+        raise ExpectedError(_('The chosen date range does not include any of the weekdays you specified.'))
+
+
+def handle_monthly_repetition(new_booking_data):
+    """Remove `recurrence_weekdays` for bookings with monthly repetition
```
ThiefMaster, line 364: `.` at the end

indico/modules/rb/client/js/common/bookings/BookingEditForm.jsx
```diff
@@ -101,6 +118,25 @@ class BookingEditForm extends React.Component {
...
+        values: {recurrence},
+      },
+    } = this.props;
+
+    const today = moment()
+      .format('ddd')
+      .toLowerCase();
+
+    if ((recurrence.type === 'every' && recurrence.interval === 'week') || force) {
+      if (recurrence.weekdays === null || recurrence.weekdays.length === 0) {
```
ThiefMaster, line 133: merge into a single if statement

indico/modules/rb/client/js/components/BookingBootstrapForm.jsx
```diff
@@ -322,6 +370,12 @@ class BookingBootstrapForm extends React.Component {
             />
           </Form.Group>
         )}
+        {type === 'every' && interval === 'week' && (
+          <Form.Group inline style={{marginLeft: '1em', marginRight: '1em'}}>
+            <label>{Translate.string('Recurring every')}</label>
```
ThiefMaster, line 375:
```suggestion
            <Translate as="label">Recurring every</Translate>
```

indico/modules/rb/client/js/modules/bookRoom/filters/RecurrenceForm.jsx
```diff
@@ -114,6 +147,19 @@ export default class RecurrenceForm extends FilterFormComponent {
             />
           </Form.Field>
         </Form.Group>
+        {type === 'every' && interval === 'week' && (
+          <Form.Field styleName="weekday-recurrence-section" inline>
+            <label styleName={type !== 'every' ? 'disabled' : ''}>
+              {Translate.string('Recurring every')}
```
ThiefMaster, line 153:
```suggestion
              <Translate>Recurring every</Translate>
```

ThiefMaster (approved):

PR #5833, opened 2023-07-04
Add flake8-bugbear🐻

tomasr8:


ThiefMaster:
This will be covered by #6037 since ruff integrates bugbear rules

PR #5834, opened 2023-07-05
Export regform_id in the event QR Code

tomasr8:


indico/modules/events/registration/controllers/management/tickets.py
```diff
@@ -62,6 +62,7 @@ def _process(self):
         checkin_app = OAuthApplication.query.filter_by(system_app_type=SystemAppType.checkin).one()
         qr_data = {
             'event_id': self.event.id,
+            'regform_id': self.regform.id,
             'title': self.event.title,
```
ThiefMaster, line 66: We might want to include the regform title as well, so the new app can show both. Especially useful for events with multiple regforms

PR #5835, opened 2023-07-05
Use the correct name format on person link fields

duartegalvao:
This PR makes the person link fields display the persons' name with the correct formatting, as set in the event layout settings or the user's preferences. Before, "[first name] [last name]" was always used.

image

### DIFF
--- a/indico/modules/events/fields.py
+++ b/indico/modules/events/fields.py
@@ -17,7 +17,7 @@ from indico.core import signals
 from indico.core.cache import make_scoped_cache
 from indico.core.db.sqlalchemy.util.session import no_autoflush
 from indico.core.errors import UserValueError
-from indico.modules.events.layout import theme_settings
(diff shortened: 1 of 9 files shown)


indico/modules/events/fields.py
```diff
@@ -112,6 +112,14 @@ def can_enter_manually(self):
             return True
         return self.event.can_manage(session.user) or not persons_settings.get(self.event, 'disallow_custom_persons')
 
+    @property
+    def name_format(self):
+        from indico.modules.users.models.users import NameFormat
+        name_format = layout_settings.get(self.event, 'name_format') if self.event else None
+        if name_format is None and session.user:
+            name_format = NameFormat(session.user.settings.get('name_format'))
```
duartegalvao, line 120: for some reason the `EnumConverter` in the user settings is not working... Couldn't figure out why :/

ThiefMaster (approved):

PR #5836, opened 2023-07-06
Enforce kwargs-only arguments in get_full_name

tomasr8:
All calls to `Registration.get_full_name` already use kwargs so this should be safe to change.


ThiefMaster:
`PersonMixin.get_full_name` has the same comment about only using kwargs ;)

ThiefMaster:
And `kwargs['_show_empty_names'] = True` in `User.get_full_name` can be replaced with passing it in the line below (Python 2 did not allow passing explicit kwargs and `**kwargs` in the same function call)

PR #5838, opened 2023-07-07
Add an option to anonymize users

tomasr8:
Adds a CLI command which allows Indico admins to anonymize users for the purposes of GDPR/OC11.
We only delete strictly user data, anything connected to events - data in person links, registrations, etc.. is the responsibility of event managers.

### DIFF
--- a/indico/cli/user.py
+++ b/indico/cli/user.py
@@ -225,16 +225,18 @@ def anonymize(user_id):
 
     This will remove all personal data about the given user including
     their name, affiliation and emails. Note that any data tied to events
-    such registrations will not be deleted.
+    such as registrations or speaker roles will not be deleted.
     """
     if (user := User.get(user_id)) is None:
         click.secho('This user does not exist', fg='red')
         return
     _print_user_info(user)
-    if not click.confirm(click.style('Anonymize this user?', fg='yellow')):
+    if not click.confirm(click.style('Permanently anonymize this user?', fg='yellow')):
         return
 
     anonymize_user(user)
+    db.session.flush()
+    db.session.commit()
     click.secho('Successfully anonymized user', fg='green')
 
 
@@ -256,12 +258,13 @@ def delete(user_id):
 
     try:
         db.session.delete(user)
-        db.session.commit()
+        db.session.flush()
     except IntegrityError as e:
         db.session.rollback()
-        click.secho('Failed to delete user', fg='red')
+        click.secho('Could not delete delete user', fg='red')
         click.echo(e)
     else:
+        db.session.commit()
         click.secho('Successfully deleted user', fg='green')
 
 
--- a/indico/cli/user.py
+++ b/indico/cli/user.py
@@ -257,9 +257,10 @@ def delete(user_id):
     try:
         db.session.delete(user)
         db.session.commit()
-    except IntegrityError:
+    except IntegrityError as e:
         db.session.rollback()
         click.secho('Failed to delete user', fg='red')
+        click.echo(e)
     else:
         click.secho('Successfully deleted user', fg='green')
 
--- a/indico/cli/user.py
+++ b/indico/cli/user.py
@@ -7,6 +7,7 @@
 
 import click
 from flask_multipass import IdentityInfo
+from sqlalchemy.exc import IntegrityError
 from terminaltables import AsciiTable
 
 from indico.cli.core import cli_group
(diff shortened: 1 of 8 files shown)


indico/cli/user.py
```diff
@@ -217,6 +219,42 @@ def unblock(user_id):
...
+    This will remove all personal data about the given user including
+    their name, affiliation and emails. Note that any data tied to events
+    such registrations will not be deleted.
+    """
+    user = User.get(user_id)
+    if user is None:
+        click.secho('This user does not exist', fg='red')
+        return
+    _print_user_info(user)
+    if click.confirm(click.style('Anonymize this user?', fg='yellow')):
```
ThiefMaster, line 236: `if not ...: return`

and I'd move all the logic to a util in indico/module/users

indico/cli/user.py
```diff
@@ -217,6 +219,42 @@ def unblock(user_id):
...
+        user.phone = ''
+        user.address = ''
+        user.picture = None
+        user.picture_metadata = None
+        user.picture_source = ProfilePictureSource.standard
+        user.favorite_users = set()
+        user.favorite_events = set()
+        user.favorite_categories = set()
+        user.identities = set()
+        # TODO: check what breaks when we delete all emails
```
ThiefMaster, line 251: I think the system user also has no email so you can check for places where we check if a user is the system user and see if there's anything important

indico/cli/user.py
```diff
@@ -217,6 +219,42 @@ def unblock(user_id):
...
+        UserEmail.query.filter(UserEmail.user == user).delete()
+        # TODO: More things we can delete?
```
ThiefMaster, line 253: I'd soft-delete the user itself

ThiefMaster:
We could also add an `indico user delete` cli which attempts to hard-delete the user. This would of course fail if the ussr is linked to things that do not support nulling the foreign key, but e.g. for spam users that didn't do anything in indico or users who just registered for an event this would work nicely.

indico/modules/events/export.py
```diff
@@ -262,6 +262,7 @@ def _make_idref(self, column, value, incoming=False, target_column=None):
         uuid = self.id_map[fullname].setdefault(value, self._get_uuid())
         if type_ == 'userref' and uuid not in self.users:
             user = User.get(value)
+            # TODO: Come back to this one
```
tomasr8, line 265: This is not needed anymore ;)

indico/modules/users/util.py
```diff
@@ -409,6 +411,60 @@ def merge_users(source, target, force=False):
...
+    user.address = ''
+    user.picture = None
+    user.picture_metadata = None
+    user.picture_source = ProfilePictureSource.standard
+    user.favorite_users = set()
+    user.favorite_events = set()
+    user.favorite_categories = set()
+    user.identities = set()
+
+    # TODO: check what breaks when we delete all emails
```
tomasr8, line 437: neither this

indico/modules/users/util.py
```diff
@@ -409,6 +411,60 @@ def merge_users(source, target, force=False):
...
+        EventPrincipal, CategoryPrincipal, AttachmentPrincipal, AttachmentFolderPrincipal,
+        BlockingPrincipal, TrackPrincipal, EventSettingPrincipal, SettingPrincipal, RoomPrincipal,
+        SessionPrincipal
+    ]
+
+    for principal in principals:
+        principal.query.filter(principal.user == user).delete()
+
+    user.is_deleted = True
+    # TODO: More things we can delete?
```
tomasr8, line 464: I think we already covered everything

indico/modules/users/util.py
```diff
@@ -419,6 +421,57 @@ def merge_users(source, target, force=False):
...
+
+    categories = user.category_roles.copy()
+    for category in categories:
+        category.members.remove(user)
+
+    principals = [
+        EventPrincipal, CategoryPrincipal, AttachmentPrincipal, AttachmentFolderPrincipal,
+        BlockingPrincipal, TrackPrincipal, EventSettingPrincipal, SettingPrincipal, RoomPrincipal,
+        SessionPrincipal
+    ]
```
ThiefMaster, line 467: This should not be hardcoded. You can use this to get a list of all principal classes:

```py
    classes = [sc for sc in [*PrincipalMixin.__subclasses__(), *PrincipalPermissionsMixin.__subclasses__()]
               if hasattr(sc, 'query')]
```

ThiefMaster (approved):

PR #5839, opened 2023-07-07
Enable CORS for endpoints used by the app

tomasr8:
Enables CORS for endpoints needed by the check-in app:
- /.well-known/oauth-authorization-server
- /api/events/<int:event_id>/registrants
- /api/events/<int:event_id>/registrants/<int:registrant_id>

Not sure if the decorator is a good idea.. alternatively I could just explicitly do `return data, {'Access-Control-Allow-Origin': '*'}` for each endpoint instead.

Since I was already modifying the endpoints I switched to using `@use_kwargs` instead of manual validation and made the endpoints inherit from `RHEventBase`

### DIFF
--- a/indico/web/rh.py
+++ b/indico/web/rh.py
@@ -408,3 +408,14 @@ def json_errors(rh):
     """
     rh._JSON_ERRORS = True
     return rh
+
+
+def cors_enabled(f):
+    """Enable CORS for this endpoint."""
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        response = f(*args, **kwargs)
+        response = current_app.make_response(response)
+        response.headers['Access-Control-Allow-Origin'] = '*'
+        return response
+    return wrapper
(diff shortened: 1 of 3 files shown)


indico/web/rh.py
```diff
@@ -408,3 +408,14 @@ def json_errors(rh):
...
     rh._JSON_ERRORS = True
     return rh
+
+
+def cors_enabled(f):
+    """Enable CORS for this endpoint."""
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        response = f(*args, **kwargs)
+        response = current_app.make_response(response)
```
ThiefMaster, line 418: I don't think it's a good idea to convert to a response here...  maybe we can instead set something on the RH instance and do it in the RH code after calling `response = current_app.make_response(res)` (which is already there)?

tomasr8: yeah that sounds like a better option

PR #5843, opened 2023-07-12
Align recurrence form elements

GovernmentPlates:
This PR aligns the recurrence form elements, as well as the booking filter bar icons.

Before:
image

After:
image


indico/modules/rb/client/js/modules/bookRoom/filters/RecurrenceForm.jsx
```diff
@@ -105,7 +105,7 @@ export default class RecurrenceForm extends FilterFormComponent {
               onChange={this.onNumberChange}
             />
           </Form.Field>
-          <Form.Field>
+          <Form.Field style={{paddingRight: '0em'}}>
```
ThiefMaster, line 108: I think just `0` (as a number) is enough if you want zero, isn't it? (same in the previous change above)

PR #5845, opened 2023-07-13
Add new Check-in app API

tomasr8:
Adds a more modern and comprehensive API for use by the Check-in app.

## Main changes:
- Extra endpoints for event/regform details which allows the app to fetch up-to-date data
- Registration endpoints are scoped to a given regform - no more mixing all registrations together

## New endpoints:

- `api/checkin/event/<int:event_id>` - new endpoint to fetch event details
- `api/checkin/event/<int:event_id>/registration/<int:reg_form_id>` - same but for regforms
- `api/checkin/event/<int:event_id>/registration/<int:reg_form_id>/registrations` - list of registrations (same as the old one)
- `/api/checkin/event/<int:event_id>/registration/<int:reg_form_id>/<int:registration_id>` - get/patch a single registration  (same as the old one)

## Other changes:

Enables CORS for all endpoints used by the app and for `/.well-known/oauth-authorization-server`. This is done with a new `RH` flag `CORS_ENABLED` which sets the correct header on the response.


indico/modules/events/registration/api.py
```diff
@@ -7,19 +7,106 @@
...
+@cors
+@oauth_scope('registrants')
+class RHAPICheckinBase(RHProtectedEventBase):
+    """Base class for the Check-in API."""
+
+    def _check_access(self):
+        RHProtectedEventBase._check_access(self)
+
+    def _process_args(self):
+        RHProtectedEventBase._process_args(self)
```
ThiefMaster, line 33: not needed

indico/modules/events/registration/api.py
```diff
@@ -7,19 +7,106 @@
...
+
+
+class RHAPIEvent(RHAPICheckinBase):
+    """Manage a single event."""
+
+    def _process_GET(self):
```
ThiefMaster, line 39: usually we just use `_process` when we don't need to run different logic for different methods

indico/web/errors.py
```diff
@@ -46,6 +48,19 @@ def load_error_data(uuid):
     return error_cache.get(uuid)
 
 
+def _handle_error_cors():
+    """Set CORS on error response based on RH's CORS settings."""
+    if not hasattr(g, 'rh') or (cors_config := getattr(g.rh, '_CORS', False)) is False:
```
ThiefMaster, line 53:
```suggestion
    if not hasattr(g, 'rh') or (cors_config := getattr(g.rh, '_CORS', None)) is None:
```

forgot to change this...

tomasr8:
^ updated the CORS tests

indico/modules/events/registration/api.py
```diff
@@ -7,19 +7,98 @@
...
+class RHAPIRegistration(RHAPIRegFormBase):
+    """Manage a single registration."""
+
+    def _process_args(self):
+        RHAPIRegFormBase._process_args(self)
+        registration_id = request.view_args['registration_id']
+        self.registration = (Registration.query
+                             .filter_by(id=registration_id, is_deleted=False)
+                             .join(RegistrationForm, RegistrationForm.id == Registration.registration_form_id)
+                             .filter_by(is_deleted=False)
```
ThiefMaster, line 81: Mixing `join` and `filter_by` results in pretty hard-to-understand code since the filter_by is relative to the last join.
Also, with the current code you can get any registration from any event as long as you can manage *some* event (and use that event ID in the URL)...

You can use `.with_parent(self.regform)` to query only registrations from the specified regform.

Using the normalization to make sure all url params are in sync would be a good idea as well:

```python
    normalize_url_spec = {
        'locators': {
            lambda self: self.registration
        }
    }
```

In the old code this is also missing, but `build_registrations_api_data` uses `with_parent(event)` and gets the registration via the regform, so tampering with URL params would not expose anything there.

indico/modules/events/registration/api.py
```diff
@@ -7,19 +7,98 @@
...
+    """Base class for registration forms management."""
+
+    def _check_access(self):
+        RHAPICheckinBase._check_access(self)
+        if not self.event.can_manage(session.user, permission='registration'):
+            raise Forbidden()
+
+    def _process_args(self):
+        RHAPICheckinBase._process_args(self)
+        self.regform = (RegistrationForm.query
```
ThiefMaster, line 48: you want `.with_parent(self.event)` here as well

indico/modules/events/registration/api.py
```diff
@@ -7,19 +7,98 @@
...
+                        .filter_by(id=request.view_args['reg_form_id'], is_deleted=False)
```
ThiefMaster, line 49: the `is_deleted` check is included in the `with_parent` since the relationship contains it in its join criterion

indico/modules/events/registration/schemas.py
```diff
@@ -33,3 +37,84 @@ def sort_list(self, data, many, **kwargs):
...
+
+class CheckinRegFormSchema(mm.SQLAlchemyAutoSchema):
+    class Meta:
+        model = RegistrationForm
+        fields = ('regform_id', 'title', 'introduction', 'start_dt', 'end_dt',
+                  'registration_count', 'checked_in_count')
+
+    regform_id = fields.Int(attribute='id')
+    registration_count = fields.Function(lambda regform: len(regform.registrations))
+    checked_in_count = fields.Function(lambda regform: len(regform.checked_in_registrations))
```
ThiefMaster, line 60: This is quite expensive for regforms with thousands of registrations. You query them twice...

for the total count see if `active_registration_count` or `existing_registrations_count` do what you want, and you could add another column property for the checked-in ones

indico/modules/events/registration/models/forms.py
```diff
@@ -538,3 +545,10 @@ def _mappers_configured():
              .correlate_except(Registration)
              .scalar_subquery())
     RegistrationForm.existing_registrations_count = column_property(query, deferred=True)
+
+    query = (select([db.func.coalesce(db.func.sum(Registration.occupied_slots), 0)])
+             .where((Registration.registration_form_id == RegistrationForm.id) &
+                    ~Registration.is_deleted & Registration.checked_in)
```
ThiefMaster, line 551: as soon as it doesn't fit in a single line, `and_(..., ..., ...)` is usually more readable than `... & ... & ...`

indico/modules/events/registration/api.py
```diff
@@ -7,19 +7,113 @@
...
+        from indico.modules.events.registration.schemas import CheckinEventSchema
+        return jsonify(CheckinEventSchema().dump(self.event))
+
+
+class RHAPIRegForms(RHAPIEvent):
+    """Manage registration forms."""
+
+    def _process(self):
+        from indico.modules.events.registration.schemas import CheckinRegFormSchema
+        regforms = RegistrationForm.query.with_parent(self.event).all()
```
ThiefMaster, line 48: `.options(undefer('checked_in_registrations_count'))` - otherwise this triggers one extra query for each regform

indico/modules/events/registration/schemas.py
```diff
@@ -33,3 +39,95 @@ def sort_list(self, data, many, **kwargs):
...
+                field_data['choices'] = field['choices']
+            # File field stores the uuid as data which is not helpful.
+            # We want to show the filename instead.
+            if field_id in filenames:
+                field_data['data'] = filenames[field_id]
+            section['fields'].append(field_data)
+
+        # Sort sections and fields based on 'position'
+        data = list(data.values())
+        data.sort(key=itemgetter('position'))
```
ThiefMaster, line 124:
```suggestion
        data = sorted(data.values(), key=itemgetter('position'))
```

indico/modules/events/registration/schemas.py
```diff
@@ -33,3 +39,95 @@ def sort_list(self, data, many, **kwargs):
...
+        for sections in data:
+            sections['fields'].sort(key=lambda f: f['position'])
```
ThiefMaster, line 126:
```suggestion
            sections['fields'].sort(key=itemgetter('position'))
```

indico/modules/events/registration/schemas.py
```diff
@@ -33,3 +39,95 @@ def sort_list(self, data, many, **kwargs):
...
+
+        # Remove the 'position' key
+        for section in data:
+            del section['position']
+            for field in section['fields']:
+                del field['position']
```
ThiefMaster, line 132: Instead of sorting and removing the key, I think you could also sort before generating the `data` dict since nowadays dicts are always sorted by insertion order.

indico/modules/events/registration/schemas.py
```diff
@@ -33,3 +39,86 @@ def sort_list(self, data, many, **kwargs):
...
+                for r in registration.data
+                if r.field_data.field.field_impl.is_file_field and r.storage_file_id is not None}
+
+    def _get_registration_data(self, registration):
+        regform = registration.registration_form
+        form_data = get_flat_section_submission_data(regform, registration=registration, management=True)
+        reg_data = get_form_registration_data(regform, registration, management=True)
+        filenames = self._get_filenames(registration)
+        sections = sorted(list(form_data['sections'].values()), key=itemgetter('position'))
+        fields = sorted(list(form_data['items'].values()), key=itemgetter('position'))
```
ThiefMaster, line 91: I think the `list()` is not needed since any iterable works

```suggestion
        sections = sorted(form_data['sections'].values(), key=itemgetter('position'))
        fields = sorted(form_data['items'].values(), key=itemgetter('position'))
```

indico/modules/events/registration/schemas.py
```diff
@@ -33,3 +39,86 @@ def sort_list(self, data, many, **kwargs):
...
+        fields = ('id', 'title', 'description', 'start_dt', 'end_dt')
+
+
+class CheckinRegFormSchema(mm.SQLAlchemyAutoSchema):
+    class Meta:
+        model = RegistrationForm
+        fields = ('id', 'event_id', 'title', 'introduction', 'start_dt', 'end_dt',
+                  'is_open', 'registration_count', 'checked_in_count')
+
+    is_open = fields.Bool(attribute='is_open')
```
ThiefMaster, line 58:
```suggestion
    is_open = fields.Bool()
```

indico/modules/events/registration/schemas.py
```diff
@@ -33,3 +39,86 @@ def sort_list(self, data, many, **kwargs):
         if many:
             data = sorted(data, key=lambda tag: natural_sort_key(tag['title']))
         return data
+
+
+# Schemas for the Check-in app API
```
ThiefMaster, line 44: Pretty clear from the name IMHO ;) And I'm sure sooner or later someone will add new (unrelated) classes at the bottom of the file anyway

indico/modules/events/registration/api.py
```diff
@@ -6,20 +6,122 @@
...
+    """Manage a single event."""
+
+    def _check_access(self):
+        RHAPICheckinBase._check_access(self)
+        if not self.event.can_manage(session.user, permission='registration'):
+            raise Forbidden()
+
+    def _process(self):
+        from indico.modules.events.registration.schemas import CheckinEventSchema
+        return jsonify(CheckinEventSchema().dump(self.event))
```
ThiefMaster, line 41:
```suggestion
        return CheckinEventSchema().jsonify(self.event)
```

ThiefMaster: same below

indico/modules/events/registration/api.py
```diff
@@ -6,20 +6,122 @@
...
+
+
+class RHAPIRegForms(RHAPIEvent):
+    """Manage registration forms."""
+
+    def _process(self):
+        from indico.modules.events.registration.schemas import CheckinRegFormSchema
+        regforms = (RegistrationForm.query
+                    .with_parent(self.event)
+                    .options(undefer('checked_in_registrations_count'))
```
ThiefMaster, line 51: don't you want to undefer `existing_registrations_count` as well?

PR #5854, opened 2023-07-18
Export Indico Axios for plugins

SegiNyn:


ThiefMaster:
Looking at some of our plugins importing it already seems to work:

    import {handleAxiosError, indicoAxios} from 'indico/utils/axios';

However, this may in fact duplicate all the axios stuff into the plugin bundle, I did not check that. If that's indeed the case your PR seems like a good idea.

SegiNyn:
Yes I saw that it's already being imported in for example the zoom plugin without error. But I cannot use it in our plugin without cause an import error. So I do wonder how it's being imported in the other plugins or if it's duplicate

PR #5863, opened 2023-07-24
Use explicit none check in room acl update

SegiNyn:
In the Room permission section, try removing **all** the persons assigned to a permission in the room and save the dialog and then reopen the dialog again, you will notice that the permissions do not actually get unassigned. 

Screenshot 2023-07-24 at 12 00 48


ThiefMaster (approved):
nice catch!

PR #5865, opened 2023-07-25
Make get_upcoming_events interceptable

vtran99:
Everything checked.


indico/modules/categories/util.py
```diff
@@ -129,6 +130,7 @@ def get_category_stats(category_id=None):
 
 @memoize_redis(3600)
 @materialize_iterable()
+@make_interceptable
```
ThiefMaster, line 133: are you sure you don't want it one above so you get the list instead of an iterable?

PR #5883, opened 2023-08-04
Add trailing space to signature delimiter in templates

paulmenzel:


ThiefMaster:
Oh and it looks like you just edited the test snapshot files. If you really want to add this (assuming there is a technical reason why it makes sense), you need to edit `indico/web/templates/emails/base.txt` and I suggest using something like `{{ '-- ' }}` instead of an actual trailing space in that file for the reason I mentioned in my last comment.

In the text snapshots (not meant to be edited manually anyway) the trailing whitespace isn't a problem, but in a real template that will be edited by humans we want to avoid this for sure.

PR #5887, opened 2023-08-11
Add px() facade for pixel dimensions

foxbunny:
This patch adds a `px()` utility SCSS function and changes all instances of pixel dimensions in all SCSS code to use it. For instance `300px` becomes `px(300)`. At the moment, the `px()` function simply outputs the pixel value as is.

This facade is added in order to aid the transition to text-relative units as described [here](https://talk.getindico.io/t/indico-accessibility/3179/29?u=hayavuk). 

After this patch is merged, all engineers working with CSS will be expected to follow this convention either permanently, or until the `px()` is replaced with adequate units, depending on how we decide to approach the text-relative units.

See #5888

### DIFF
--- a/indico/modules/categories/client/js/components/CategoryModeration.module.scss
+++ b/indico/modules/categories/client/js/components/CategoryModeration.module.scss
@@ -6,13 +6,14 @@
 // LICENSE file for more details.
 
 @use 'base/palette' as *;
+@use 'base/utilities' as *;
 
 :global(.ui.table).table {
   // fix to allow the table to be wider than
   // 800px set by .page-content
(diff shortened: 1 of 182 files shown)


ThiefMaster:
Would it be possible to do this in a less invasive ways such as a postcss preprocessor instead of editing every single file containing a unit using pixels?

foxbunny:
Off the top of my head I don't recall any postcss plugin that does this kind of thing, but I'll look into it.

foxbunny:
Ok, so there's a plugin that does conversion to `rem`.

https://github.com/cuth/postcss-pxtorem

I'll see if this actually solves the problem or merely kicks it under the rug. :)

PR #5892, opened 2023-08-17
Add a service handler for deleting editables

duartegalvao:
This PR adds a a service handler that is called whenever an editable is deleted.

### DIFF
--- a/indico/modules/events/editing/controllers/backend/timeline.py
+++ b/indico/modules/events/editing/controllers/backend/timeline.py
@@ -32,8 +32,8 @@ from indico.modules.events.editing.operations import (assign_editor, create_new_
 from indico.modules.events.editing.schemas import (EditableSchema, EditingConfirmationAction, EditingReviewAction,
                                                    ReviewEditableArgs)
 from indico.modules.events.editing.service import (ServiceRequestFailed, service_get_custom_actions,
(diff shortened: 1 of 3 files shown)


indico/modules/events/editing/controllers/backend/timeline.py
```diff
@@ -201,7 +201,14 @@ def _check_access(self):
             raise Forbidden
 
     def _process(self):
+        service_url = editing_settings.get(self.event, 'service_url')
```
ThiefMaster, line 204: I'd just do `if editing_settings.get(self.event, 'service_url'):` below.. getting that setting will work regardless of whether the editable exist or not

indico/modules/events/editing/service.py
```diff
@@ -195,6 +195,20 @@ def service_handle_review_editable(editable, user, action, parent_revision, revi
         raise ServiceRequestFailed(exc)
 
 
+def service_handle_delete_editable(editable):
+    try:
+        path = '/event/{}/editable/{}/{}'.format(
+            _get_event_identifier(editable.event),
+            editable.type.name,
+            editable.contribution_id
+        )
```
tomasr8, line 204: Maybe I'd put this outside the try block?

PR #5894, opened 2023-08-18
Add lang attribute that matches the currently selected language

foxbunny:
Adds the `lang` attribute to the `HTML` element in the `layout/indico_base.html` and `layout/base.html` templates.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,7 @@ Improvements
 - Allow room managers to add internal notes to bookings (:issue:`5746`, :pr:`5791`)
 - Support generating tickets and badges for each of the registrant's accompanying
   persons (:pr:`5424`)
+- Include current language in page metadata (:pr:`5894`, thanks :user:`foxbunny`)
 
 Bugfixes
 ^^^^^^^^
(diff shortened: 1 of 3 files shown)


tomasr8:
Should we use just the language or the full locale? e.g. `en` rather than `en-GB`? The region subtag is [optional](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang) but maybe it's useful in some cases?

ThiefMaster:
We already have the `moment_lang` property, maybe we could add `html_lang` in a similar way?

foxbunny:
> We already have the moment_lang property, maybe we could add html_lang in a similar way?

Sure.

CHANGES.rst
```diff
@@ -25,6 +25,7 @@ Improvements
   restrict them to speakers or registrants (:pr:`5670`, thanks :user:`kewisch`)
 - Make editing timeline display much more straightforward (:pr:`5674`)
 - Allow event managers to delete editables from contributions (:pr:`5778`)
+- Add `lang` attribute to `<html>` elements in base templates (:pr:`5894`)
```
ThiefMaster, line 28: rst uses double backticks for code highlighting (single backticks are already used for rst references and similar features)

Issue #5896, opened 2023-08-21
[A11Y] Replace the qtip2 integration in the navigation toolbar


foxbunny:
**Describe the bug**

The navigation toolbar at the top of the page currently uses the qtip2 library to show the lower-level items. This results in code that is not quite accessible.

**To Reproduce**
1. Go to main page
2. Note that `a.arrow.js-dropdown` does not have appropriate ARIA attributes to indicate its state/purpose

**Expected behavior**

Expect `a.arrow.js-dropdown` to have `aria-controls` pointing to the menu items, and `aria-expanded` that reflects its state.

**Additional context**

- https://talk.getindico.io/t/ui-state-toolbar-navigation-menu/3216/7
- https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html

ThiefMaster:
@foxbunny I guess this one can be closed? (it was referenced in the header dropdown PR I merged, but not as `closes ...`)

foxbunny:
Yes

PR #5895, opened 2023-08-18
Layout fixes

foxbunny:
Fixes layout issues resulting from user-defined stylesheets that modify text-related metrics such as letter, line, word and paragraph spacing.

See #5888

Issue #444, opened 2010-06-10
Display of material for categories


jbenito3:
``` yaml
# Ticket imported from Trac
# Originally assigned to: lsyroka
```

The display of attached material for categories is not nice and without white spaces between the elements

http://indico.cern.ch/categoryDisplay.py?categId=2628

PR #5897, opened 2023-08-21
Make header toolbar dropdown menu accessible

foxbunny:
This patch makes the following changes:

- Implements a new `<ind-menu>` custom element that is used to add necessary `aria-*` attributes to a combination of `<button>` and `<menu>` in the tollbar menu
- Adds two folders, `client/{js,styles}/custom-elements`, that will be used to house Indico custom elements
- Includes the `ind-menu` module in `client/js/index.js`
- Restyles the toolbar menu to allow buttons to be used as menu items

See #5896 

This PR requires the #5895 to be merged

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -94,6 +94,7 @@ Accessibility
   :thanks:`foxbunny`)
 - Associate form labels with form controls in the registration form (:issue:`6059`, :issue:`6073`,
   :pr:`6076`, thanks :user:`foxbunny`)
+- Make dropdown menu fully accessible (:issue:`5896`, :pr:`5897`, thanks :user:`foxbunny`)
 
 Internal Changes
 ^^^^^^^^^^^^^^^^
@@ -116,6 +117,8 @@ Internal Changes
 - Improve ``indico i18n`` CLI to support plugin-related i18n operations (:issue:`5906`, :pr:`5961`,
   thanks :user:`SegiNyn`)
 - Use `ruff <https://docs.astral.sh/ruff/>`_ for linting Python code (:pr:`6037`)
+- Add ``<ind-menu>`` custom element for managing drop-down menus (:issue:`5896`, :pr:`5897`,
+  thanks :user:`foxbunny`)
 
 
 ----
--- a/indico/web/client/js/custom_elements/TipBase.js
+++ b/indico/web/client/js/custom_elements/TipBase.js
@@ -1,0 +1,169 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2023 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import {domReady} from 'indico/utils/domstate';
+import './tips.scss';
+
+let viewportWidth = document.documentElement.clientWidth;
+let viewportHeight = document.documentElement.clientHeight;
+
+domReady.then(() => requestIdleCallback(updateClientGeometry));
+window.addEventListener('resize', updateClientGeometry);
+window.addEventListener('orientationchange', updateClientGeometry);
+
+function updateClientGeometry() {
+  viewportWidth = document.documentElement.clientWidth;
+  viewportHeight = document.documentElement.clientHeight;
+}
+
+function positionVertically(referenceRect, tooltipRect) {
+  const vw = viewportWidth;
+  const vh = viewportHeight;
+  let top = 'auto';
+  let bottom = 'auto';
+  const refCenter = `${referenceRect.left + referenceRect.width / 2}px`;
+  let arrowBorder;
+
+  // Place above or below?
+  if (tooltipRect.height < referenceRect.top) {
+    bottom = `${vh - referenceRect.top}px`;
+    arrowBorder = 'var(--tooltip-surface-color) transparent transparent';
+  } else {
+    top = `${referenceRect.bottom}px`;
+    arrowBorder = 'transparent transparent var(--tooltip-surface-color)';
+  }
+
+  // Ideal left coordinate (CSS will adjust as needed)
+  const idealLeft = referenceRect.left + (referenceRect.width - tooltipRect.width) / 2;
+  // NB: the clamp() formula assumes the tooltip content will always fit the viewport, which is ensured using CSS
+  const left = `clamp(0.5em, ${idealLeft}px, ${vw - tooltipRect.width}px - 0.5em)`;
+
+  return {
+    top,
+    bottom,
+    left,
+    'right': 'auto',
+    'ref-center': refCenter,
+    'arrow-borders': arrowBorder,
+  };
+}
+
+function positionHorizontally(referenceRect, tooltipRect) {
+  const vw = viewportWidth;
+  const vh = viewportHeight;
+  let left = 'auto';
+  let right = 'auto';
+  const refCenter = `${referenceRect.top + referenceRect.height / 2}px`;
+  let arrowBorder;
+
+  // Place right or left?
+  if (tooltipRect.width < vw - (referenceRect.x + referenceRect.width)) {
+    left = `${referenceRect.x + referenceRect.width}px`;
+    arrowBorder = 'transparent var(--tooltip-surface-color) transparent transparent';
+  } else {
+    right = `${vw - referenceRect.x}px`;
+    arrowBorder = 'transparent transparent var(--tooltip-surface-color) transparent';
+  }
+
+  // Ideal top coordinate (CSS will adjust as needed)
+  const idealTop = referenceRect.top + (referenceRect.height - tooltipRect.height) / 2;
+  // NB: the clamp() formula assumes the tooltip content will always fit the viewport, which is ensured using CSS
+  const top = `clamp(0.5em, ${idealTop}px, ${vh - tooltipRect.height}px - 0.5em)`;
+
+  return {
+    top,
+    'bottom': 'auto',
+    left,
+    right,
+    'ref-center': refCenter,
+    'arrow-borders': arrowBorder,
+  };
+}
+
+export class TipBase extends HTMLElement {
+  constructor() {
+    super();
+    this.show = this.show.bind(this);
+    this.hide = this.hide.bind(this);
+    this.dismiss = this.dismiss.bind(this);
+  }
+
+  connectedCallback() {
+    domReady.then(() => {
+      this.$tip = this.querySelector('[data-tip-content]');
+      console.assert(this.$tip, 'Must contain a *[data-tip-content] element');
+      this.setup();
+    });
+  }
+
+  disconnectedCallback() {
+    window.removeEventListener('keydown', this.dismiss);
+  }
+
+  get orientation() {
+    return this.getAttribute('orientation');
+  }
+
+  set orientation(value) {
+    this.setAttribute('orientation', value);
+  }
+
+  get shown() {
+    return this.hasAttribute('shown');
+  }
+
+  set shown(isShown) {
+    if (this.shown === isShown) {
+      return;
+    }
+    this.toggleAttribute('shown', isShown);
+    this.dispatchEvent(new Event('toggle'));
+  }
+
+  getTipCSS() {
+    const referenceRect = this.getBoundingClientRect();
+    const tooltipRect = this.$tip.getBoundingClientRect();
+
+    if (this.orientation === 'horizontal') {
+      return positionHorizontally(referenceRect, tooltipRect);
+    } else {
+      return positionVertically(referenceRect, tooltipRect);
+    }
+  }
+
+  updatePosition() {
+    this.style = '';
+    const css = this.getTipCSS();
+    for (const key in css) {
+      this.style.setProperty(`--${key}`, css[key]);
+    }
+  }
+
+  dismiss(evt) {
+    if (!this.shown || evt.code !== 'Escape') {
+      return;
+    }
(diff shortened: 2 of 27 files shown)


CHANGES.rst
```diff
@@ -25,6 +25,8 @@ Improvements
   restrict them to speakers or registrants (:pr:`5670`, thanks :user:`kewisch`)
 - Make editing timeline display much more straightforward (:pr:`5674`)
 - Allow event managers to delete editables from contributions (:pr:`5778`)
+- Support custom user-defined stylesheets with text metric overrides (:pr:`5895`)
+- Add ``<ind-menu>`` custom element and make the header toolbar dropdown accessible (:pr:`5897`)
```
ThiefMaster, line 29: I think this better fits into "Internal Changes" (and maybe we can consider introducing an a11y section later)

Also, credit yourself ;) see the "thanks :user:...` in some other places for contributions coming from outside the core indico team

foxbunny:
Rebased relevant changes on top of master, and also addressed code style issues discussed elsewhere.

foxbunny:
Rebased on master, updated folder structure to match previous PRs.

indico/web/client/styles/partials/_main.scss
```diff
@@ -11,26 +11,63 @@
...
-.toolbar.global-menu {
-  @include font-family-title-light();
+#global-menu {
+  // FIXME: This should be in the global reset
 
-  height: 3em;
-  line-height: 3em;
+  * {
+    box-sizing: border-box;
+  }
```
ThiefMaster, line 19: I think this is no longer needed since you added it globally in a previous PR?

indico/web/client/styles/partials/_main.scss
```diff
@@ -11,26 +11,63 @@
 
 /* Styles for the main menu */
 
-.toolbar.global-menu {
-  @include font-family-title-light();
```
ThiefMaster, line 15: Any reason for removing this (which changes the font? If it needs to be larger fine, but new font looks strange there.

I tried adding it back locally, and IMHO it looks better (master vs your PR + this line added back):

image
image

ThiefMaster:
I rebased it to the latest master and cleaned up some merge problems in the changelog.

The one thing I'm missing a bit in the new version is the little arrow indicating that a top menu item opens a submenu:

image
image

Also, can we avoid this little bit of white spacing between the main menu and its submenu?

indico/web/client/js/custom_elements/ind_menu.js
```diff
@@ -0,0 +1,56 @@
...
+
+import './ind_menu.scss';
+
+let lastId = 0; // Track the assigned IDs to give each element a unique ID
+
+customElements.define(
+  'ind-menu',
+  class extends HTMLElement {
+    connectedCallback() {
+      window.addEventListener('DOMContentLoaded', () => {
```
ThiefMaster, line 16: Is this still needed in a custom element? (e.g. if/because the `connectedCallback` runs before the element and its children have been fully loaded into he DOM)

indico/web/client/js/custom_elements/ind_menu.js
```diff
@@ -0,0 +1,56 @@
...
+          // Delay action as no element is focused immediately after focusout
+          requestAnimationFrame(() => {
+            if (this.contains(document.activeElement)) {
+              return;
+            }
+            trigger.removeAttribute('aria-expanded');
+          });
+        });
+
+        this.addEventListener('keydown', function(ev) {
```
ThiefMaster, line 44: `evt`

indico/web/client/js/custom_elements/ind_menu.js
```diff
@@ -0,0 +1,56 @@
...
+        trigger.setAttribute('aria-expanded', false);
+
+        list.id = list.id || `dropdown-list-${lastId++}`;
+        trigger.setAttribute('aria-controls', list.id);
+
+        trigger.addEventListener('click', function() {
+          trigger.setAttribute('aria-expanded', trigger.getAttribute('aria-expanded') !== 'true');
+        });
+
+        this.addEventListener('focusout', function() {
```
ThiefMaster, line 34: Generally I prefer arrow functions for callbacks (also the ones above/below), but *this* one is interesting because of the `this` semantics.

I think in terms of readability and maintainability (I actually had to double-check the docs how `this` is used by `addEventListener`) it might be better to just use arrows and put a more explicit `const elem = evt.target;` (or `.currentTarget`) inside the callback.

indico/web/client/styles/partials/_main.scss
foxbunny: I'm going to clean this up later once I get done with category display page. Still have some issues to address there. But yeah, I'll remove this once I get it it.

indico/web/client/styles/partials/_main.scss
foxbunny: So the idea is to first bring all styling under the DS, and then improve the design afterwards.

> but new font looks strange there.

It's not the new font. It's just inheriting it from one of the ancestors. I want to flush out SUI first before I address font-related stuff.

indico/web/client/js/custom_elements/ind_menu.js
foxbunny: "Is this still needed in a custom element?"

Unfortunately yes. We'll want to do something about deferring all <script> tags so we don't have to do this everywhere. Luckily, it's not hard to remove later either.

indico/web/client/js/custom_elements/ind_menu.js
foxbunny: > Generally I prefer arrow functions for callbacks

Sure. I've switched to arrows in newer PRs.

indico/web/client/styles/partials/_main.scss
foxbunny: Fair enough. I should note that Roboto Light is most certainly not the appropriate font for this use case. In fact, any "Light" font is not a good use case for normal size text (and below) in general. You may use it on larger headings, but that's a different use case. Not all font rendering engines are made the same, and light fonts generally tend to render poorly on anything that isn't a Mac.

All in all we'll need to address the font choices anyway at some point. My long-term plan after the migration to DS is complete is to pick a good font set that works well across the board both in terms of readability and also from the aesthetic perspective.

ThiefMaster: Sure, but IMHO those "more invasive" changes are better to keep for later since we can do all the nice improvements at first without exposing users to big "oh wth why does this look completely different now?!?!" surprises :)

foxbunny: > "oh wth why does this look completely different now?!?!"

This is called a "canoe" problem. It's a temporary distraction that does not have no serious impact on usability. We generally shouldn't upgrade canoe problems into serious failures. ;)

foxbunny:
@ThiefMaster Where does this tool tip come from? Tried searching the exact text but nothing shows up.

image

ThiefMaster:
you find it when searching for `event-type-tooltip-lecture`

foxbunny:
Ugh, more jQuery :D

indico/web/client/js/custom_elements/ind_with_tooltip.js
```diff
@@ -0,0 +1,50 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2023 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import {TipBase} from 'indico/custom_elements/TipBase';
+
+const tipDelay = 1000; // ms
```
ThiefMaster, line 10: This feels too slow. What about using something around 100-250ms? (I tested it and it feels way more reactive that way)

indico/web/client/js/custom_elements/ind_menu.js
ThiefMaster: would be nice to update it in here as well :)

ThiefMaster:
image

Can we avoid this one-pixel gap between the menu and the expanded dropdown?

ThiefMaster:
image

Would it be possible to open the tooltip to the right instead of to the top? For the first dropdown item it doesn't matter, but right now it's pretty much impossible to create a lecture after hovering meeting/conference - you have to move the mouse pointer outside the dropdown first so the tooltip closes, since it covers the dropdown entries above.

ThiefMaster:
Is the very dark gray (`#333`) required for contrast or would something slightly brighter such as `#444` or `#555` (the previous color) still have strong enough contrast (especially since the text is a bit larger now than it was before - which in this particular place actually looks better than it looked before with the smaller text IMHO ;))?

indico/web/client/js/custom_elements/ind_with_tooltip.js
foxbunny: So there's this:

image

But then again, many authors use about .5s delay instead. 

So not sure what is "typical" really. At any rate, 100~250 is way too low. This delay is used to minimize distraction while tabbing through the interface.

ThiefMaster: OK fair enough. Probably this particular tooltip is not that useful except for very novice Indico users anyway!

(and if we ever move the event creation dialog to react I could absolutely see the button to just become "Create event" and the type selection going in the dialog itself instad of being nested in a dropdown)

foxbunny:
> Is the very dark gray (#333) of the menu bar required for contrast or would something slightly brighter such as #444 or #555 (the previous color) still have strong enough contrast

As far as grays go, I think the cut-off is #737373. So yeah, we can use that. Should make for better contrast with the tooltip, too.

indico/web/client/js/custom_elements/TipBase.js
```diff
@@ -0,0 +1,123 @@
...
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import {domReady} from 'indico/utils/domstate';
+import './tips.scss';
+
+let viewportWidth = document.documentElement.clientWidth;
+let viewportHeight = document.documentElement.clientHeight;
+
+window.addEventListener('DOMContentLoaded', () => requestIdleCallback(updateClientGeometry));
```
ThiefMaster, line 14: `domReady.then(...);`?

indico/web/client/js/custom_elements/TipBase.js
```diff
@@ -0,0 +1,123 @@
...
+    const vh = viewportHeight;
+    const referenceRect = this.getBoundingClientRect();
+    const tooltipRect = this.$tip.getBoundingClientRect();
+
+    let top = 'auto';
+    let bottom = 'auto';
+    const refCenter = `${referenceRect.x + referenceRect.width / 2}px`;
+    let arrowBorder;
+
+    console.log(referenceRect);
```
ThiefMaster, line 67: remove that ;)

indico/web/client/js/custom_elements/ind_menu.js
```diff
@@ -0,0 +1,57 @@
...
+      domReady.then(() => {
+        const trigger = this.querySelector('button');
+        const list = this.querySelector('menu');
+
+        console.assert(
+          trigger.nextElementSibling === list,
+          'The <menu> element must come after <button>'
+        );
+
+        trigger.setAttribute('aria-expanded', false);
```
ThiefMaster, line 26: I guess this should be `'false'` since AFAIK DOM attributes are always strings? (even though the `false` is probably implicitly converted to a string

foxbunny: Yes, it's always converted to string, and that is negligible performance hit too, so you'd just waste two characters to be pedantic. :)

ThiefMaster: Just asking because you (correctly) check for `!== 'true'` below and I could already see a less experienced developer checking for `=== false` for whatever reason (after all, it's set to `false`!) and wondering why it doesn't work.

foxbunny: Well, what would it take for them to learn how coercion works in JavaScript?

ThiefMaster: "Explicit is better than implicit." (even though that's from the zen of python and we have javascript here ;))

foxbunny: There's another saying: 'When in Rome, do as the Romans do.' This is JavaScript. Coercion is everywhere. You're not escaping that. Gotta learn at some point.

indico/web/client/js/custom_elements/tips.scss
```diff
@@ -0,0 +1,112 @@
...
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+@use 'design_system';
+
+ind-with-tooltip,
+ind-with-toggletip {
+  --tooltip-surface-color: #000;
```
ThiefMaster, line 12: Would it cause any a11y problems if we changed this to `#333`? Feels a bit very dark right now.

image
image


(Might also be personal preference, but for anything dark-mode-related I generally prefer the "dim" version over the "full black background" version, and I guess this also applies here even if it's just a tooltip)

ThiefMaster: At least right now we don't have anything darker than the menus and the meeting page background....

(so far the preference within the team is on the `#333` version btw)

foxbunny: Pushed

PR #5903, opened 2023-08-23
Make language list more accessible

### LINKED ISSUE #5899: [A11Y] Language selector markup
**Describe the bug**

The language selector is inaccessible due to missing `href` on the anchor tags. Furthermore it is semantically incorrect as language switching is not linking to anything.

**To Reproduce**
1. Go to main page
2. Attempt to tab into the language selector
3. Observe that tab navigation skips over the language links

**Expected behavior**

The language links/buttons are focusable and operable using keyboard.

**Additional context**

- https://talk.getindico.io/t/keyboard-language-selector/3219
- https://talk.getindico.io/t/language-language-selector/3207
- https://talk.getindico.io/t/instructions-language-selector/3211
- https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions
- https://www.w3.org/WAI/WCAG21/Understanding/language-of-page.html

foxbunny:
- Fix the button markup (change A to BUTTON)
- Add lang attribute as appropriate
- Expand the ARIA label for the menu button (NOTE: i18n needed)
- Increase target surface area of the language buttons

Fixes #5899

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,7 @@ Improvements
   persons (:pr:`5424`)
 - Include current language in page metadata (:pr:`5894`, thanks :user:`foxbunny`)
 - Add keyboard shortcut (CTRL-SHIFT-A) to toggle room booking admin override (:pr:`5909`)
+- Make language list accessible (:pr:`5903`)
 
 Bugfixes
 ^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,7 +30,7 @@ Improvements
   persons (:pr:`5424`)
 - Include current language in page metadata (:pr:`5894`, thanks :user:`foxbunny`)
 - Add keyboard shortcut (CTRL-SHIFT-A) to toggle room booking admin override (:pr:`5909`)
-- Make language list accessible (:pr:`5903`)
+- Make language list accessible (:issue:`5899`, :pr:`5903`, thanks :user:`foxbunny`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/web/client/styles/widgets/_session-bar.scss
+++ b/indico/web/client/styles/widgets/_session-bar.scss
@@ -1,0 +1,35 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2023 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+@use 'base/palette' as *;
+
+#language-list {
+  padding-top: 0.5em;
+  padding-bottom: 0.5em;
+
+  button {
+    width: 100%;
+    padding: 0.5em 0.2em;
+    font-size: inherit;
+    border: none;
+    background: transparent;
+    text-align: left;
+  }
+
+  button:not(:disabled) {
+    cursor: pointer;
+    color: $dark-blue;
+  }
+
+  button:is(:hover, :focus-visible) {
+    color: inherit;
+  }
+
+  button:disabled {
+    color: $dark-gray;
+  }
+}
--- a/indico/web/templates/_session_bar.html
+++ b/indico/web/templates/_session_bar.html
@@ -171,40 +171,45 @@
         {%- if ambiguous and territory %} ({{ territory }}){% endif -%}
     {%- endset %}
     {% if lightweight %}
+        {# FIXME: [A11Y] Replace with button with appropriate state (requires new dropdown impl.) #}
         <a id="language-selector-link"
-           href=""
+           href="#language-list"
+           data-ignore-href="true"
            class="{{ 'arrow' if can_choose }} js-dropdown" {% if can_choose %}data-toggle="dropdown"{% endif %}>
             {%- trans %}My language ({{ language }}){% endtrans %}
         </a>
     {% else %}
         <button id="language-selector-link" class="icon-earth i-button {{ 'arrow js-dropdown' if can_choose }}"
-                {% if can_choose %}data-toggle="dropdown"{% else %}disabled{% endif %}>
+                {% if can_choose %}data-toggle="dropdown"{% else %}disabled{% endif %}
+                aria-label="{% trans %}The current language is {{ language }}. Use a different language.{% endtrans %}">
             {{- language -}}
         </button>
     {% endif %}
     {% if can_choose %}
-        <ul class="i-dropdown">
+        <ul id="language-list" class="i-dropdown">
             {% for name, (title, territory, __) in languages|dictsort(by='value') %}
+                {% set html_lang = name.replace('_', '-') %}
                 <li>
                     {% if force_locale and name not in force_locale_alts %}
-                        <a class="disabled" title="{% trans %}This language is not available here.{% endtrans %}">
+                        <button disabled title="{% trans %}This language is not available here.{% endtrans %}" lang="{{ html_lang }}">
                             {{- title -}}
                             {%- if territory %} ({{ territory }}){% endif -%}
-                        </a>
+                        </button>
                     {% elif name == session.lang %}
-                        <a class="disabled" title="{% trans %}This is your current language.{% endtrans %}">
+                        <button disabled title="{% trans %}This is your current language.{% endtrans %}" lang="{{ html_lang }}">
                             {{- title -}}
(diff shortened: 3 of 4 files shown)


foxbunny:
For now I'm going to leave the conversion to a `<form>` alone. I'll first convert the session bar into a pull-down menu as we've discussed, and then I'll circle back to this.

indico/web/client/styles/widgets/_session-bar.scss
```diff
@@ -0,0 +1,35 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2023 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+@use 'base/palette' as *;
+
+#language-list {
```
ThiefMaster, line 10: It's usually better to use classes for styling instead of IDs. And this may be a good case for scss nesting:

```scss
.session-bar-language-list {
  // ...

  button {
    //...
  }
}
```

indico/web/templates/_session_bar.html
```diff
@@ -171,40 +171,45 @@
...
+           href="#language-list"
+           data-ignore-href="true"
            class="{{ 'arrow' if can_choose }} js-dropdown" {% if can_choose %}data-toggle="dropdown"{% endif %}>
             {%- trans %}My language ({{ language }}){% endtrans %}
         </a>
     {% else %}
         <button id="language-selector-link" class="icon-earth i-button {{ 'arrow js-dropdown' if can_choose }}"
-                {% if can_choose %}data-toggle="dropdown"{% else %}disabled{% endif %}>
+           {% if can_choose %}data-toggle="dropdown"{% else %}disabled{% endif %}
+           aria-label="{% trans %}The current language is {{ language }}. Use a different language.{% endtrans %}">
```
ThiefMaster, line 184: we usually align attributes under the first attribute of the previous line

indico/web/templates/_session_bar.html
```diff
@@ -171,40 +171,45 @@
...
-                        <a class="disabled" title="{% trans %}This is your current language.{% endtrans %}">
+                        <button disabled data-qtip-oldtitle="{% trans %}This is your current language.{% endtrans %}" lang="{{ html_lang }}">
                             {{- title -}}
                             {%- if territory %} ({{ territory }}){% endif -%}
-                        </a>
+                        </button>
                     {% else %}
-                        <a data-href="{{ url_for('core.change_lang') }}"
+                        <button data-href="{{ url_for('core.change_lang') }}"
                            data-method="POST"
```
ThiefMaster, line 205:
```suggestion
                        <button data-href="{{ url_for('core.change_lang') }}"
                                data-method="POST"
```

(also adapt the indentation below)

ThiefMaster:
Besides those small issues it looks and works really well though! Only thing I'd look into is changing the padding/margin of the individual language entries a bit.

indico/web/client/styles/widgets/_session-bar.scss
foxbunny: > It's usually better to use classes for styling instead of IDs. And this may be a good case for scss nesting:

I use `id` when it's a single element on the page as in this case. This makes it more obvious that this element is not going to be used in multiple places, and makes editing CSS safer. The idea that "classes are better" is quite an old opinion that has to do with inefficiency of the ID-based selector in the past, and today is no longer valid. In fact, much of the problems people claim is CSS's fault comes from the overuse of classes. I only use classes in two situiations:

- when I need to mark a group of elements that are *semantically* related and need to be differentiated from other elements
- when the element needs to be styled independently of its context (portable)

In this case, the language list does not satisfy the conditions.

As far as nesting goes, I tend to use explicit selectors when the selector is not intended to be `@extend`ed later.

Let me know if you would still insist to have these changed regardless.

ThiefMaster:
> Only thing I'd look into is changing the padding/margin of the individual language entries a bit, they feel larger than before.

Just noticed that this is likely intentional ("Increase target surface area of the language buttons") so nevermind.

indico/web/client/styles/widgets/_session-bar.scss
ThiefMaster: OK, fine for me. i'd still use nesting though ;)

foxbunny:
> Besides those small issues it looks and works really well though! Only thing I'd look into is changing the padding/margin of the individual language entries a bit, they feel larger than before.

This is to increase their target size. Since we're not touching font sizes as of right now, it looks larger than it should be. Ideally I would also increase the font size.

indico/web/templates/_session_bar.html
```diff
@@ -171,40 +171,45 @@
...
     {% endif %}
     {% if can_choose %}
-        <ul class="i-dropdown">
+        <ul id="language-list" class="i-dropdown">
             {% for name, (title, territory, __) in languages|dictsort(by='value') %}
+                {% set html_lang = name.replace('_', '-') %}
                 <li>
                     {% if force_locale and name not in force_locale_alts %}
-                        <a class="disabled" title="{% trans %}This language is not available here.{% endtrans %}">
+                        <button disabled data-qtip-oldtitle="{% trans %}This language is not available here.{% endtrans %}" lang="{{ html_lang }}">
```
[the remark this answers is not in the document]
foxbunny: Indeed it does not. I'll figure it out.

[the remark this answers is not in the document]
ThiefMaster: assuming you tested with `title="..."` and it didn't work, this is probably because disabled elements do not receive events.

a possible solution could be going back to `a` elements and just adding `href="#"` - as long as the default event is prevented hat wouldn't be an issue, and they'd hopefully be accessible nonetheless.

putting the title on the `li` could also be an option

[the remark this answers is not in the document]
foxbunny: So it doesn't work with `disabled` buttons, which is, in fact, correct. I'll keep this as a `title` attribute in order to keep the translation text, and it doesn't really hurt. Since these buttons are not clickable anyway it should be fine.

The other disabled button, for the unavailable language may need additional explanation. I'm not sure the one provided in the tooltip is sufficient, though. To me it is still not clear why it is unavailable and what the user can do to rectify it. May be better to simply not render such languages to begin with?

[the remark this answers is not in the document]
ThiefMaster: "unavailable languages" are part of the 'force event languages' feature where an event can specify that it wants to allow the indico UI to only be shown in certain languages, e.g. because of event organizers who do not want mixed language content.

you can test this by creating an event and on the main event settings page setting a specific event language and toggling the 'force' switch:

image

the idea of showing the disabled languages is so a user knows why the language the may be using indico in at other times is suddenly not available.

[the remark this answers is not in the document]
foxbunny: > disabled elements do not receive events.

And they are explicitly filtered out in the tooltip code.


> the idea of showing the disabled languages is so a user knows why the language the may be using indico in at other times is suddenly not available.

What do you think about the following idea. Once we implement the new pull-down menu (with larger surface area) we could express this by saying:

"The following languages are disabled for this event:"

indico/web/client/styles/widgets/_session-bar.scss
foxbunny: > OK, fine for me. i'd still use nesting though ;)

What's your preference regarding what nesting should represent? I generally avoid nesting to represent the relative relationships of elements because fixing that in a nested block is more work than fixing it in a combinator.

ThiefMaster: I usually use it to avoid repetition. And in this case it nicely 'namespaces' the rules since they are all related to one parent component

CHANGES.rst
```diff
@@ -29,6 +29,7 @@ Improvements
 - Support generating tickets and badges for each of the registrant's accompanying
   persons (:pr:`5424`)
 - Include current language in page metadata (:pr:`5894`, thanks :user:`foxbunny`)
+- Make language list accessible (:pr:`5903`)
```
ThiefMaster, line 32:
```suggestion
- Make language list accessible (:issue:`5899`, :pr:`5903`, thanks :user:`foxbunny`)
```

(#5899 is related to this PR or even solved by it, no?)

ThiefMaster (approved):

PR #5907, opened 2023-08-25
Add hard expiration to Flask session

cbartz:
This PR adds the ability to set a hard expiration for the Flask session. This can be used in conjunction with a Flask Multipass identity provider that specifies the session expiration. See [this](https://talk.getindico.io/t/enhancing-multipass-saml-authentication-automated-refresh-for-extended-sessions/3223) post for more details on use cases and implementation considerations.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,9 @@ Improvements
 - Support weekly room bookings that take place on multiple weekdays (:pr:`5829`, :issue:`5806`)
 - Hide events marked as invisible from builtin search results unless the user is a manager
   (:pr:`5947`, thanks :user:`openprojects`)
+- Support sessions that expire at a certain date (specified by the used flask-multipass
+  provider) regardless of activity when using an external login method (:pr:`5907`, thanks
+  :user:`cbartz`)
 
 Bugfixes
 ^^^^^^^^
--- a/indico/modules/auth/__init__.py
+++ b/indico/modules/auth/__init__.py
@@ -4,6 +4,7 @@
 # Indico is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
+from datetime import datetime
 
 from flask import redirect, request, session
 from flask_multipass import MultipassException
@@ -71,6 +72,11 @@ def process_identity(identity_info):
         identity.data = identity_info.data
     if user.is_blocked:
         raise MultipassException(_('Your Indico profile has been blocked.'))
+
+    if session_expiry := identity.multipass_data.get('session_expiry'):
+        if not isinstance(session_expiry, datetime):
+            raise ValueError('Session expiry must be a datetime object')
+        session.hard_expiry = identity.multipass_data['session_expiry']
     login_user(user, identity)
 
 
--- a/indico/modules/auth/__init__.py
+++ b/indico/modules/auth/__init__.py
@@ -4,6 +4,7 @@
 # Indico is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see the
 # LICENSE file for more details.
+
 from datetime import datetime
 
 from flask import redirect, request, session
--- a/indico/modules/auth/__init__.py
+++ b/indico/modules/auth/__init__.py
@@ -32,6 +32,12 @@ logger = Logger.get('auth')
 @multipass.identity_handler
 def process_identity(identity_info):
     logger.info('Received identity info: %s', identity_info)
+
+    if session_expiry := identity_info.multipass_data.get('session_expiry'):
+        if not isinstance(session_expiry, datetime):
+            raise ValueError('Session expiry must be a datetime object')
+        session.hard_expiry = session_expiry
+
     identity = Identity.query.filter_by(provider=identity_info.provider.name,
                                         identifier=identity_info.identifier).first()
     if identity is None:
@@ -74,10 +80,6 @@ def process_identity(identity_info):
     if user.is_blocked:
         raise MultipassException(_('Your Indico profile has been blocked.'))
 
-    if session_expiry := identity.multipass_data.get('session_expiry'):
-        if not isinstance(session_expiry, datetime):
-            raise ValueError('Session expiry must be a datetime object')
-        session.hard_expiry = identity.multipass_data['session_expiry']
     login_user(user, identity)
 
 
--- a/indico/modules/auth/__init__.py
+++ b/indico/modules/auth/__init__.py
@@ -33,7 +33,7 @@ logger = Logger.get('auth')
 def process_identity(identity_info):
     logger.info('Received identity info: %s', identity_info)
 
-    if session_expiry := identity_info.multipass_data.get('session_expiry'):
+    if (multipass_data := identity_info.multipass_data) and (session_expiry := multipass_data.get('session_expiry')):
         if not isinstance(session_expiry, datetime):
             raise ValueError('Session expiry must be a datetime object')
         session.hard_expiry = session_expiry
--- a/indico/modules/auth/auth_test.py
+++ b/indico/modules/auth/auth_test.py
@@ -1,0 +1,47 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2023 CERN
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see the
+# LICENSE file for more details.
+
+from datetime import datetime, timezone
+
+import pytest
+from flask import session
+from flask_multipass import IdentityInfo, Multipass
+from flask_multipass.providers.static import StaticIdentityProvider
+
+from indico.modules.auth import process_identity
+
+
+@pytest.mark.usefixtures('database')
+def test_process_identity_sets_hard_expiry(app):
+    """Test that hard expiry is set on the session object if it is present in the multipass data."""
+    dt_now = datetime.now(timezone.utc)
+    identity_info = IdentityInfo(
+        provider=StaticIdentityProvider(
+            multipass=Multipass(), name='static', settings={}
+        ),
+        identifier='static', multipass_data={'session_expiry': dt_now}
+    )
+
+    with app.test_request_context('/ping', method='GET'):
+        process_identity(identity_info)
+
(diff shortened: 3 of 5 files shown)


indico/web/flask/session_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+    with pytest.raises(ValueError):
+        session.hard_expiry = datetime(2021, 1, 1, 0, 0, 0)
+
+
+def test_get_storage_lifetime_with_hard_expiry(app, mocker):
+    """Test that the storage lifetime is the hard expiry."""
+    dt_mock = Mock(spec_set=datetime)
+    dt_now = dt_mock.now.return_value = datetime.now(timezone.utc)
+    mocker.patch('indico.web.flask.session.datetime', dt_mock)
```
ThiefMaster, line 29: I think you may be able to avoid manually mocking datetime altgeother and instead use our `freeze_time` pytest fixture.

CHANGES.rst
```diff
@@ -29,6 +29,8 @@ Improvements
 - Support generating tickets and badges for each of the registrant's accompanying
   persons (:pr:`5424`)
 - Include current language in page metadata (:pr:`5894`, thanks :user:`foxbunny`)
+- Add the ability to use Flask sessions that don't refresh, if supported by the
+  Flask-Multipass identity provider.
```
ThiefMaster, line 33:
```suggestion
  Flask-Multipass identity provider (:pr:`5907`, thanks :user:`cbartz`)
```

indico/web/flask/session.py
```diff
@@ -153,16 +169,19 @@ def get_cookie_secure(self, app):
         return request.is_secure
 
     def get_storage_lifetime(self, app, session):
+        # Sessions with hard expiry are stored for exactly as long as they are valid.
         # Permanent sessions are stored for exactly the same duration as the session id cookie.
         # "Temporary" session are stored for a period that is not too short/long as some people
         # close their browser very rarely and thus shouldn't be logged out that often.
-        if session.permanent:
+        if session.hard_expiry:
+            return session.hard_expiry - datetime.now(timezone.utc)
```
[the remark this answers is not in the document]
cbartz: Do you want to set `app.permanent_session_lifetime` or `self.temporary_session_lifetime` to this value?

In my opinion, this value should only apply to this specific session, as the value might vary from session to session, and other users who are not using an identity provider with hard expiration should not be affected.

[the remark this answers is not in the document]
ThiefMaster: no i mean something like this:

```py
session_lifetime = ...
if session.hard_expiry:
    return max(session.hard_expiry, session_lifetime)
return session_lifetime
```

(pseudocode-ish, i didn't bother between relative/absolute times for the eaxmple)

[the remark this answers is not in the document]
cbartz: Mhh, sorry, I am not quite getting the point here. So the lifetime in storage would be the maximum of any of the lifetimes, so potentially longer than `session.hard_expiry - datetime.now(timezone.utc)`. What would be the benefit?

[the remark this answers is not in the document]
ThiefMaster: I meant `min()` in my example of course ;)

indico/modules/auth/auth_test.py
```diff
@@ -0,0 +1,47 @@
...
+
+import pytest
+from flask import session
+from flask_multipass import IdentityInfo, Multipass
+from flask_multipass.providers.static import StaticIdentityProvider
+
+from indico.modules.auth import process_identity
+
+
+@pytest.mark.usefixtures('database')
```
ThiefMaster, line 18:
```suggestion
@pytest.mark.usefixtures('db')
```

database is session-scoped (creates the actual tables etc) and rather low-level while `db` does the proper setup of the sqlalchemy sessions, prevents data from being permanently committed to the test DB, etc.

indico/modules/auth/auth_test.py
```diff
@@ -0,0 +1,47 @@
...
+        identifier='static', multipass_data={'session_expiry': dt_now}
+    )
+
+    with app.test_request_context('/ping', method='GET'):
+        process_identity(identity_info)
+
+        assert session.hard_expiry == dt_now
+
+
+@pytest.mark.usefixtures('database')
```
ThiefMaster, line 35:
```suggestion
@pytest.mark.usefixtures('db')
```

indico/web/flask/session.py
```diff
@@ -208,18 +233,17 @@ def save_session(self, app, session, response):
...
             # are saving the session anyway!
             session.permanent = True
 
         storage_ttl = self.get_storage_lifetime(app, session)
-        cookie_lifetime = self.get_expiration_time(app, session)
-        session['_expires'] = datetime.now() + storage_ttl
-
-        if refresh_sid:
-            self.storage.delete(session.sid)
-            session.sid = self.generate_sid()
```
ThiefMaster, line 222: Why did you remove this? It's a security issue ("session fixation") to not reset the session ID when logging in/logging out

cbartz: This was a mistake. Mea culpa.

ThiefMaster:
Just wondering... should there some kind of indicator that the session will expire soon when someone uses this feature? It'd be a bit disruptive if someone is e.g. writing an abstract or minutes for an event and does not realize that the session will expire in a few minutes!

ThiefMaster:
We do use flash messages in a few places, but IMHO it would be nicer to actually expose the session expiry to the client, and show a warning dialog client-side that the session will soon expire (with the option to refresh it (by going through the login again)) when getting close to the expiry.

SegiNyn:
> but IMHO it would be nicer to actually expose the session expiry to the client, and show a warning dialog client-side that the session will soon expire (with the option to refresh it (by going through the login again)) when getting close to the expiry.

Hi @ThiefMaster, we were also proposing an expiration dialog of this nature to warn users when the session is about to expire and give them the opportunity to refresh:
Screenshot 2023-10-02 at 17 10 40
Screenshot 2023-10-02 at 17 10 54

is this something we can coordinate with this pr?

cc @OmeGak

SegiNyn:
Sending a pr to @cbartz's branch sounds good to me.

indico/modules/auth/__init__.py
```diff
@@ -30,6 +32,12 @@
 @multipass.identity_handler
 def process_identity(identity_info):
     logger.info('Received identity info: %s', identity_info)
+
+    if session_expiry := identity_info.multipass_data.get('session_expiry'):
```
SegiNyn, line 36: Hi @cbartz, please does this work when logging in with a local account? I'm getting a `AttributeError: 'NoneType' object has no attribute 'get'` when logging in with a local account.

ThiefMaster: the `multipass_data` is `None` for local accounts. so this is currently a bug in the PR

SegiNyn: Okay, thanks

SegiNyn:
Hi @ThiefMaster, I have sent a PR to @cbartz's branch here: https://github.com/cbartz/indico/pull/1 There are still some problems with it like some resources being fetched when opening the dialog which refreshes the session. You won't see this if you run indico with the `-q` flag because those requests will be silenced in the logs but they still reach the app and refresh the session.

ThiefMaster:
Try this to prevent those requests from refreshing the session:

```patch
diff --git a/indico/web/flask/session.py b/indico/web/flask/session.py
index 8637f295c4..1d7fc785dd 100644
--- a/indico/web/flask/session.py
+++ b/indico/web/flask/session.py
@@ -162,6 +162,9 @@ class IndicoSessionInterface(SessionInterface):
             return self.temporary_session_lifetime

     def should_refresh_session(self, app, session):
+        if (request.blueprint == 'assets' or
+                (request.blueprint.startswith('plugin_') and request.endpoint.endswith('.static'))):
+            return False
         if session.new or '_expires' not in session:
             return False
         threshold = self.get_storage_lifetime(app, session) / 2
```

indico/web/templates/indico_new_base.html
```diff
@@ -25,4 +26,7 @@
         {{ render_announcements() }}
     {% endblock %}
     {% block body %}{% endblock %}
+    {% block session_expiration %}
```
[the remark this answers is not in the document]
SegiNyn: Yes, sorry this is a leftover that's not being used

ThiefMaster:
I forcepushed some cleanups and also getting rid of those merge commits while rebasing

ThiefMaster:
@SegiNyn  some feedback after testing this:

- I think the `RequestConfirm` component is not a good choice here. It's meant for confirmation of possibly destructive actions. For example, it has no option to only show one button (OK and Cancel doing the same is bad UX) nor to not close when clicking outside the dialog.
- Sending the user back to the login page only makes sense if there's a hard expiry date set for the session (and thus no possibility of refreshing it). Otherwise a request to an endpoint that simply refreshes its lifetime (as is the case during regular Indico usage) should be made which is much less intrusive (imagine you're writing minutes or an abstract and get this message - you do not expect a redirect, and in most cases it would not be necessary).

ThiefMaster:
Also, considering that the dialog still needs a bit of work and is not as straightforward as I originally thought, I'm thinking about merging this PR without the dialog stuff tomorrow (3.3 is under development so having no indicator of a soon-expiring session for the time being is fine). That way we avoid the indirection of PRs to a fork, and reviewing the dialog stuff properly will be easier...

SegiNyn:
hi @ThiefMaster, Yes I did notice the issue with the buttons which basically do the same thing but I had not thought of how disruptive it would be to refresh the page while the user is filling in an abstract. So I can add a custom dialog for this part. It's ok for me if you merge this pr with the hard expiry and then I create a different one for the dialog. I'm not going to be in tomorrow so I cannot work on it until Monday anyway.

ThiefMaster (approved):

ThiefMaster:
@SegiNyn I pushed a branch containing your changes, rebased on top of the latest master, to `session-expiry-with-dialog` in my indico fork. It contains some fixes I did on top of your code so it might be good for you to reset your local branch to that one.

cbartz:
Thanks for merging the PR @ThiefMaster

PR #5914, opened 2023-08-31
Improve TZ selector markup and UX

### LINKED ISSUE #5908: [A11Y] Timezone list markup improvements
**Describe the bug**

Timezone selector's behavior currently suffers from several issues related to keyboard interaction. It could also use some UX improvements as well.

**To Reproduce**

Scenario A:

1. Go to any page which has a time zone selector
2. Focus the timezone button using a keyboard
3. Open the timezone popup by pressing Enter/Return or Spacebar
4. Try to navigate into the timezone popup
5. Observe that navigation "skips" the timezone popup

Scenario B:

1. Click at the top of the timezone popup so that further keyboard navigation will start from the top of the popup
2. Focus the radios
3. Select the "Specify a timezone" radio
4. Observe that the focus is automatically moved to the select list below

**Expected behavior**

When the popup is expanded, the keyboard focus should move to the popup *when the user presses Tab key* (not automatically).

When the "Specify a timezone" radio is selected, focus should *not* automatically move to the select list.

There are also several issues with the semantics of the markup which I will address in a patch for this issue.

**Additional context**

- https://talk.getindico.io/t/keyboard-navigation-session-bar-timezone-picker/3197/3

foxbunny:
- Modify the TZ selector markup to be more keyboard accessible
- Modify the markup to convey information about states
- Modify the TZ select list behavior to automatically activate the "Custom" radio when selection is made
- Modify the TZ selector to not disable the select list when options other than "Custom" are used

This patch also adds some initial code for a design system.

Fixes #5908

### DIFF
--- a/indico/modules/core/client/js/session_bar.js
+++ b/indico/modules/core/client/js/session_bar.js
@@ -5,6 +5,7 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
+// eslint-disable-next-line import/unambiguous
 (function() {
   function setupProtection() {
     $('#protection-details-link').qtip({
@@ -40,20 +41,10 @@
 
   function setupTimezone() {
     const widget = $('#tz-selector-widget');
-    widget.on('change', 'input[name=tz_mode]', function() {
-      const customTZ = this.value === 'custom';
-      const customTZSelect = widget.find('select[name=tz]');
-      customTZSelect.prop('disabled', !customTZ);
-      if (customTZ) {
-        customTZSelect.focus();
-        scrollToCurrentTZ();
-      }
-    });
 
     $('#tz-selector-link').qtip({
       style: {
-        width: '300px',
-        classes: 'qtip-rounded qtip-shadow qtip-popup qtip-timezone',
+        classes: 'qtip-rounded qtip-shadow qtip-popup qtip-timezone qtip-wide',
         tip: {
           corner: true,
           width: 20,
@@ -80,6 +71,7 @@
       },
       events: {
         show() {
+          // eslint-disable-next-line no-undef
           _.defer(scrollToCurrentTZ);
         },
       },
--- a/indico/web/client/js/widgets/tz_selector.js
+++ b/indico/web/client/js/widgets/tz_selector.js
@@ -1,0 +1,40 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2023 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+// eslint-disable-next-line import/unambiguous
+window.addEventListener('DOMContentLoaded', () => {
+  const form = document.querySelector('#tz-selector-widget form');
+  const tzModes = document.getElementById('tz-modes');
+  const customRadio = document.getElementById('tz-mode-custom');
+  const tzCustomField = document.querySelector('#tz-custom-field select');
+
+  // No TZ selector?
+  if (!form) {
+    return;
+  }
+
+  tzCustomField.querySelector('option:checked')?.scrollIntoView();
+
+  function setCustomMode() {
+    customRadio.checked = true;
+  }
+
+  tzCustomField.addEventListener('click', setCustomMode);
+  tzCustomField.addEventListener('change', setCustomMode);
+
+  tzModes.addEventListener('change', evt => {
+    if (evt.target.value === 'user') {
+      tzCustomField.querySelector(`option[value="${evt.target.dataset.userTz}"]`)?.scrollIntoView();
+    } else if (evt.target.value === 'custom') {
+      tzCustomField.querySelector('option:checked')?.scrollIntoView();
+    }
+  });
+
+  form.addEventListener('submit', evt => {
+    evt.preventDefault();
+  });
+});
--- a/indico/web/client/styles/design_system/README.md
+++ b/indico/web/client/styles/design_system/README.md
@@ -1,0 +1,163 @@
+# Indico Design System
+
+The SCSS design system for Indico.
+
+This design system is intended for use exclusively with the `@extend`
+feature. It does not provide any mixins, SCSS variables, and similar.
+
+## Basic usage
+
+The intended usage is as follows:
+
+```scss
+@use 'design_system';
+
+some-selector {
+  @extend %generic-container;
+  --container-element-spacing: 1em;
+}
+```
+
+This results in the following output:
+
+```css
+some-selector {
+  %generic-container declarations....
+}
+
+some-selector {
+  --container-element-spacing: 1em;
+}
+```
+
+Customization is done using CSS variables. They can be defined either
+on the selected element itself, or one of its ancestors. Using CSS
+variables on ancestors allows us to target any of its descendants, so
+we are able to override the styles for a particular region of the page.
+
+```scss
+some-region {
+  --control-clickable-surface-color: #{$some-other-color};
+}
+
+some-selector {
+  @extend %generic-container;
+  --container-element-spacing: 1em;
+}
+```
+
+## Background
+
+This is decidedly different from the common way of using CSS in which the
+appearance of the element is defined using classes. Instead of adding
+classes to elements in order to specify their appearance, we instead use
+placeholder selectors and `@extend` to specify the appearance for elements
+selected by the selector. However, in a way, it is also similar to how the
+class-based CSS works, in that we use the placeholder selectors the same
+way we use classes.
+
+This approach has the following benefits:
+
+- It reduces the amount of markup being sent over the wire (all elements
+  need a single class at most).
+- It restricts all appearance-related changes to the stylesheet.
+
+
+## Why extends and not mixins
+
+The placeholder selectors are not mixins. They do not take parameters.
+Instead, we use CSS variables to customize the appearance of the
+elements. This is a deliberate trade-off to avoid duplicating CSS
+across different selectors (with `@extend`, the selector is moved to the
+placeholder's declaration block, rather than the other way around).
+
+CSS variables also allow the third party stylesheet to override various
+aspects of the visual appearance using plain CSS. Another advantage of CSS
+variables is that they are subject to CSS cascade, so they can be used by
+the parent to override appearance for its descendants, etc.
+
+
+## Media queries
+
+One trade-off of using the `@extend` method is that placeholder selectors
+cannot be used in media queries. The following code shows why:
+
+```scss
+%placeholder {
+  flex: none;
+  width: 2em;
+}
+
+@media (width < 45em) {
+  .foo {
+    @extend %placeholder;
+  }
+}
+```
+
+The above code would have to compile to this:
+
+```scss
+.foo {
+  flex: none;
+}
+
+@media (width < 45em) {
+  .foo { }
+}
+```
+
+If SCSS were to compile the code the way we instructed, it would break the
+media query. Within at-rules, we still need to use mixins. In most cases, this
+can be achieved by converting the placeholder selector into a mixin like so:
+
+```scss
+@mixin placeholder {
+  flex: none;
+  width: 2em;
+}
+
+%placeholder {
+  @include placeholder();
+}
+
+// ...
+
+@media (width < 45em) {
+  .foo {
+    @include placeholder;
+  }
+}
+```
+
+Please be careful when converting placeholder selector that extend other
+placeholder selectors. Such placeholders cannot be converted into a mixin
+and you will need to figure out a different approach. However, in most
+cases such complex placeholder selectors are not intended for use within
+media queries to begin with, so do first check that what you're trying to
+do actually makes sense.
+
+## Maintenance
+
+In order to keep this system working, please keep in mind the following
+rules of thumb:
+
+- All variable parameters that are shared among the modules in this folder
+  should be defined *as CSS variables* in the `variables.scss` module.
+- All SCSS modules in this folder should only use CSS variables except for
+  the `variables.scss` module, which can use SCSS variables as well.
+- The CSS variable name loosely follows the following convention:
+
+```scss
+--{TOPIC}-{ELEMENT/SUBTOPIC}-{CHARACTERISTIC}
+```
+
+- Ensure the CSS variable names are searchable as there are no mechanisms
+  to spot typos.
+- Placeholder selectors contain information about the structure of the
+  elements they apply to. Do not require the developer to add classes to
+  the target elements if not needed. Do not over-prescribe the layout of
+  the elements if not needed: keep it as flexible as possible (this means
+  you need to keep the nesting to the minimum, among other things, one
+  level for namespacing, and sometimes another one for pseudo-classes or
+  pseudo-elements is enough).
--- a/indico/web/client/styles/design_system/_form_controls.scss
+++ b/indico/web/client/styles/design_system/_form_controls.scss
@@ -1,0 +1,69 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2023 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+@use 'utils';
+@use 'layout';
+@use 'partials/icons';
+
+%form-label {
+  @extend %reset-font-and-color;
+  font-weight: bold;
+}
+
+%form-label-block {
+  @extend %form-label;
+  display: block;
+  margin-bottom: 0.5em;
+}
+
+%select-list {
+  border: var(--control-border);
+  border-radius: var(--control-border-radius);
+
+  option:not(:checked):hover {
+    background: var(--control-clickable-surface-color);
+  }
+
+  option:checked {
+    background: var(--control-alt-clickable-surface-color);
+    color: var(--control-alt-text-color);
+  }
+}
+
+%checkbox {
+  @extend %flex-inline-center;
+  appearance: none;
+  width: 1.5em;
+  height: 1.5em;
+  background: var(--control-indented-bg-color);
+  border: var(--control-border);
+  border-radius: var(--control-border-radius);
+
+  &:checked {
+    background: var(--control-alt-clickable-surface-color);
+    color: var(--control-alt-text-color);
+  }
+
+  &:focus-visible:not(:checked) {
+    background: var(--control-clickable-surface-focus-color);
+  }
+
+  &:focus-visible:checked {
+    background: var(--control-alt-clickable-surface-focus-color);
+  }
+
+  &:disabled {
+    background: var(--control-disabled-indented-bg-color);
+    color: var(--control-disabled-text-color);
+    border-color: var(--control-disabled-border-color);
+  }
+
+  &:checked::before {
+    @extend %icon;
+    @extend %icon-checkmark;
+  }
+}
--- a/indico/web/client/styles/design_system/_index.scss
+++ b/indico/web/client/styles/design_system/_index.scss
@@ -5,8 +5,11 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
-@forward 'widgets/occurrences';
-@forward 'widgets/permissions';
-@forward 'widgets/session-bar';
-@forward 'widgets/sortable_list';
-@forward 'widgets/tz_selector';
+@use './variables';
+@use './button';
+@use './button_group';
+@use './form_controls';
+@use './layout';
+@use './popup';
+@use './visibility';
+@use './utils';
--- a/indico/web/client/styles/design_system/_utils.scss
+++ b/indico/web/client/styles/design_system/_utils.scss
@@ -5,8 +5,12 @@
 // modify it under the terms of the MIT License; see the
 // LICENSE file for more details.
 
-@forward 'widgets/occurrences';
-@forward 'widgets/permissions';
-@forward 'widgets/session-bar';
-@forward 'widgets/sortable_list';
-@forward 'widgets/tz_selector';
+%reset-font-and-color {
+  color: inherit;
+  font: inherit;
+}
+
+%default-focus-ring {
+  outline: medium solid Highlight;
+  outline: medium solid -webkit-focus-ring-color;
+}
--- a/indico/web/client/styles/partials/_buttons.scss
+++ b/indico/web/client/styles/partials/_buttons.scss
@@ -287,21 +287,6 @@ a.i-big-button {
   }
 }
 
-.i-button.signin {
-  display: block;
-  text-align: left;
-  margin-top: 5px;
-
-  .auth-id {
-    font-weight: bold;
-  }
-
-  .login-arrow {
-    float: right;
-    @include icon-before('icon-arrow-right-sparse');
-  }
-}
-
 .i-button.next {
   @include icon-after('icon-next');
   font-weight: bold;
@@ -355,7 +340,7 @@ button::-moz-focus-inner {
 }
 
 .i-button.right {
-  margin-left: 5px;
+  margin-right: 5px;
   margin-top: 5px;
 }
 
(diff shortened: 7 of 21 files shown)


foxbunny:
@ThiefMaster Heads-up: this patch includes an experimental design system that I propose for the use on public pages. The code is contained inside the `styles/design_system` folder (there is a README) file in there that explains what the intent is and how it should work. The DS already addresses some of the color contrast issues, but it generally tries to approximately match the existing UI and integrate with it. Please check it out and let me know your thoughts.

ThiefMaster:
Not sure I'll find the time to look into it today and I'm off tomorrow, but I'll have a closer look next week, but maybe @tomasr8 can already provide some input on it in the meantime

foxbunny:
> Not sure I'll find the time to look into it today and I'm off tomorrow, but I'll have a closer look next week, but maybe @tomasr8 can already provide some input on it in the meantime

No rush

tomasr8:
Haven't gone through the DS yet, but just noticed a small error when navigating to the management area of an event (probably because the tz selector is disabled there):

```js
Uncaught TypeError: $tzCustomField is null
    js tz_selector.js:18
    js tz_selector.js:9
```

foxbunny:
> noticed a small error when navigating to the management area

Should be fixed now

foxbunny:
Not sure what this CI failure is about. Doesn't look like it's related to the changes I've made...

tomasr8:
> Not sure what this CI failure is about. Doesn't look like it's related to the changes I've made...

Maybe some GH hiccup? I tried rerunning the job and it failed again..

ThiefMaster:
Yeah, the broken python interpreter happens from time to time on github... annoying but harmless

ThiefMaster:
CI bug is fixed in #5920, rebase to the latest master and it should be fine again

indico/web/client/styles/design_system/README.txt
```diff
@@ -0,0 +1,114 @@
...
+====================
+
+
+This file is part of Indico.
+Copyright (C) 2002 - 2023 CERN
+
+Indico is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see the LICENSE file for more
+details.
+
```
ThiefMaster, line 11: not needed in a random README (also because this would not be caught by our tool to update headers)

indico/web/client/styles/design_system/_index.scss
```diff
@@ -0,0 +1,15 @@
...
+// LICENSE file for more details.
+
+@use 'variables';
+@use 'button';
+@use 'button_group';
+@use 'form_controls';
+@use 'layout';
+@use 'popup';
+@use 'visibility';
+@use 'utils';
```
ThiefMaster, line 15: I'd be clear using `./whatever` that those are relative imports and not some top-level thing (like `'base'` would be for example)

indico/web/client/styles/design_system/README.txt
```diff
@@ -0,0 +1,114 @@
...
+This design system is intended for use exclusively with the `@extend`
+feature. It does not provide any mixins, SCSS variables, and similar.
+
+
+Basic usage
+-----------
+
+The intended usage is as follows:
+
+    @import "design_system";
```
tomasr8, line 27: Any reason for `@import` instead of `@use`?

foxbunny:
The only caveat with placeholders is that each media query is essentially a separate context. That may result in counter-intuitive behavior. For example:

```
@media (width < 70rem) {
  foo {
    @extend %flex-column;
  }
}
```

This doesn't work because the placeholder itself is *outside* the media target. This is basically a trade-off of using this style of code reuse. For this, we would do something like this:

```
mixin flex-column {
  display: flex;
  flex-direction: column;
}

%flex-column {
  @include flex-column;
}
```

So we'd have both a mixin and a placeholder. The media query would result in some code duplication, but it shouldn't be too bad overall.

tomasr8:
> The only caveat with placeholders is that each media query is essentially a separate context. That may result in counter-intuitive behavior.

I think this is worth mentioning in the readme

foxbunny:
Good call. I'll fix it first thing tomorrow.

indico/web/client/styles/design_system/README.txt
```diff
@@ -0,0 +1,157 @@
+Indico Design System
```
ThiefMaster, line 1: I think this file would deserve being markdown as well :)

ThiefMaster:
image

In the old version, the timezone select dropdown is disabled when one of the first two options are checked. This is no longer the case - intentionally?

You also no longer see what your default profile timezone is - maybe the dropdown should switch to your timezone when selecting that option?

One case of unexpected behavior is if you select a custom timezone, enable 'remember these settings', and then click on "event/category" or "profile" again and save. Then it still updates your default timezone in the database (tip: run `python bin/utils/db_log.py -S` to see all non-SELECT queries being executed)

indico/web/client/styles/design_system/README.txt
ThiefMaster: And just thinking, would it make sense to move the design system in a separate PR? I think both things are pretty much ready to merge, but it might be nicer to have them in separate PRs (also because IMHO the design system deserves a changelog entry in 'internal changes')

foxbunny: > And just thinking, would it make sense to move the design system in a separate PR?

To me this type of decision is usually based on the need to revert them separately. I've no strong opinion either way.

foxbunny:
> In the old version, the timezone select dropdown is disabled when one of the first two options are checked. This is no longer the case - intentionally?

Yes. Totally intentional.

> You also no longer see what your default profile timezone is - maybe the dropdown should switch to your timezone when selecting that option?

Ok, I'll figure something out. Probably a message below the buttons or something.

> One case of unexpected behavior is if you select a custom timezone, enable 'remember these settings', and then click on "event/category" or "profile" again and save. Then it still updates your default timezone in the database (tip: run python bin/utils/db_log.py -S to see all non-SELECT queries being executed)

If I understand this correctly, "remember these settings" basically means "use custom setting as your profile timezone"?

indico/web/client/styles/design_system/README.txt
ThiefMaster: Same here (no strong opinion). If it's already a separate commit I'd split it, if not I wouldn't bother to spend/waste time with it ;)

foxbunny:
I think I see what the issue is. Since you're checking `tz and tz in common_timezones_set`, and the new implementation doesn't disable this field, it always sets the user time zone regardless of what mode you're in.

I'll fix this on Monday.

foxbunny:
Updated PR.

There is now an indication of your profile timezone:

image

The profile timezone is also highlighted in the custom timezone list, with an additional "(your profile timezone)" label:

image


This item will scroll into view when "Profile" option is selected.

When the "Custom" option is selected, any selected item in the select list will be scrolled into view.

I've fixed the behavior of the "Remember these settings" by making a change in the controller as follows:


```
        elif mode == 'custom' and tz in common_timezones_set:
            session.timezone = tz
            if update_user and session.user:  # <--- MOVED HERE
                session.user.settings.set('timezone', tz)

        if update_user and session.user:
            session.user.settings.set('force_timezone', mode != 'local')
```

This now makes the backend behavior match the UI. Please let me know if this fix is appropriate.

foxbunny:
Also, converted README to markdown.

foxbunny:
Not quite sure why these tests are failing. I tried running locally, but I get a different result. What command should I use to run the python tests?

ThiefMaster:
It's also broken when you simply access indico without being logged in (`session.user` is None in that case)

tomasr8:
`pytest --no-cov indico/web/util_test.py` to reproduce ;)

indico/web/templates/_session_bar.html
```diff
@@ -48,64 +48,76 @@
...
+                    </fieldset>
+                    <label id="tz-custom-field">
+                        <span>{% trans %}Select a custom timezone{% endtrans %}</span>
+                        <select name="tz" size="12">
+                            {% for tz in timezone_data.timezones -%}
+                                {% set is_profile_timezone = tz == profile_timezone %}
+                                <option value="{{ tz }}" {{ 'selected' if tz == session.tzinfo.zone }} {{ 'data-profile' if is_profile_timezone }}>
+                                    {{ tz }}
+                                    {% if is_profile_timezone -%}
+                                        {# Translators: example usage: "Europe/Geneva (your profile timezone)" #}
```
ThiefMaster, line 96: Nobody will read that ;x

I suggest doing this instead (untested):

```jinja
                                    {% if is_profile_timezone -%}
                                        {% trans %}{{ tz }} (your profile timezone){% endtrans %}
                                    {%- else -%}
                                        {{- tz -}}
                                    {%- endif %}                                    
```

That way the placeholder is in the translatable string so it'll be much clearer how it's used

foxbunny: Ah, great idea! Thanks

foxbunny:
> `pytest --no-cov indico/web/util_test.py` to reproduce ;)

I get a bunch of these:

```
SKIPPED [1] indico/web/util_test.py:598: Could not retrieve PostgreSQL version: [Errno 2] No such file or directory: 'initdb'
```

foxbunny:
Any idea why this test in particular would fail for me?

```
_________________________ test_proxy_to_reservation_if_last_valid_occurrence[False-True-True-True] _________________________

db = <IndicoSQLAlchemy postgresql://indico:indico@localhost:5432/indico_test>
mocker = <pytest_mock.plugin.MockerFixture object at 0x7fed88f1ea90>
create_reservation = <function create_reservation.<locals>._create_reservation at 0x7fed88f17b80>
dummy_user = <User(1337, 1337@example.test): "Guinea Pig">, not_repeating = False, only_one_valid = True, propagate = True
proxied = True, freeze_time = <function freeze_time.<locals>._freeze_time at 0x7fed88f17e50>

    @pytest.mark.parametrize(
        ('not_repeating', 'only_one_valid', 'propagate', 'proxied'),
        bool_matrix('..0', expect=False) +
        bool_matrix('..1', expect='any_dynamic')
    )
    @pytest.mark.usefixtures('smtp')
    def test_proxy_to_reservation_if_last_valid_occurrence(db, mocker, create_reservation, dummy_user,
                                                           not_repeating, only_one_valid, propagate, proxied, freeze_time):
        resv = create_reservation(start_dt=datetime.combine(date.today(), time(8)),
                                  end_dt=datetime.combine(date.today() + timedelta(days=1), time(17)),
                                  repeat_frequency=RepeatFrequency.NEVER if not_repeating else RepeatFrequency.DAY)
        freeze_time(datetime.combine(date.today(), time(8)))
        if only_one_valid:
            for occ in resv.occurrences[1:]:
                occ.state = ReservationOccurrenceState.cancelled
            db.session.flush()

        occ = resv.occurrences.first()
        mocker.patch.object(resv, 'cancel')
        occ.cancel(user=dummy_user, propagate=propagate)
>       assert resv.cancel.called == proxied
E       assert False == True
E        +  where False = <MagicMock name='cancel' id='140658181178128'>.called
E        +    where <MagicMock name='cancel' id='140658181178128'> = <Reservation(249, room_id=1133, 2023-09-25 08:00:00, 2023-09-26 17:00:00, accepted): "Testing">.cancel

indico/modules/rb/models/util_test.py:40: AssertionError
```

foxbunny:
Rebased on master

ThiefMaster:
no idea about the failing test. a wild guess would be that the failure is depending on the time of the day when it runs

foxbunny:
I currently get no failures after rebasing on master.

indico/web/client/js/widgets/tz_selector.js
```diff
@@ -0,0 +1,35 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2023 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+// eslint-disable-next-line import/unambiguous
+window.addEventListener('DOMContentLoaded', function() {
```
ThiefMaster, line 9:
```suggestion
window.addEventListener('DOMContentLoaded', () => {
```

indico/web/client/js/widgets/tz_selector.js
```diff
@@ -0,0 +1,35 @@
...
+  const tzCustomField = document.querySelector('#tz-custom-field select');
+
+  tzCustomField.querySelector('option:checked')?.scrollIntoView();
+
+  function setCustomMode() {
+    customRadio.checked = true;
+  }
+
+  tzCustomField?.addEventListener('click', setCustomMode);
+  tzCustomField?.addEventListener('change', setCustomMode);
```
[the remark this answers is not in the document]
foxbunny: Yeah, it's completely unnecessary. Just forgot to remove them.

indico/web/client/styles/widgets/_tz_selector.scss
```diff
@@ -0,0 +1,80 @@
...
+
+  fieldset {
+    @extend %button-group;
+  }
+
+  fieldset label {
+    flex: 1;
+  }
+
+  [type='radio'] {
```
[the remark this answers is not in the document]
foxbunny: 1. I think it's better to underspecify a selector whenever possible, as that keeps them more flexible. For instance, I use `[name="foo"]`, `[role="tab"]` and similar selectors where applicable.
2. If efficiency of a selector ever becomes an issue, it probably means there's something deeply wrong with the UI elsewhere :)

indico/web/client/styles/widgets/_tz_selector.scss
```diff
@@ -0,0 +1,80 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2023 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+@use 'design_system';
+
+// NB: The following rule is a temporary hack to work around QTip2 styling. It should be removed after the popup is converted into a pull-down menu.
```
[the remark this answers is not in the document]
foxbunny: Is "XXX" your convention for "todo" items?

foxbunny:
Addressed comments

indico/web/client/styles/widgets/_tz_selector.scss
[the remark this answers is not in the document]
ThiefMaster: it's more for general notes. TODO is for actual todo items

ThiefMaster:
@foxbunny I think you missed some (maybe collapsed by github?) :/

foxbunny:
Ok, will check tomorrow

indico/web/client/js/widgets/tz_selector.js
foxbunny: Btw, I think arrow callbacks can be enforced via ESLint.

foxbunny:
> I think you missed some

I went over the last batch of comments, and couldn't find what you're referring to. Can you please point me to the comments?

indico/web/client/js/widgets/tz_selector.js
ThiefMaster: I [thought so too](https://github.com/indico/eslint-config-indico/blob/909231c41a7a347b85925f720f2715a967b91696/configs/main.yaml#L162C3-L162C30)... `prefer-arrow-callback: warn`

indico/web/client/styles/design_system/README.md
```diff
@@ -0,0 +1,163 @@
...
+    --container-element-spacing: 1em;
+}
+```
+
+## Background
+
+This is decidedly different from the common way of using CSS in which the
+appearance of the element is defined using classes. Instead of adding
+classes to elements in order to specify their appearance, we instead use
+placeholder selectors and @extend to specify the appearance for elements
```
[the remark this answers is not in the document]
ThiefMaster: @foxbunny ^

indico/web/client/styles/design_system/README.md
```diff
@@ -0,0 +1,163 @@
...
+    flex: none;
+}
+
+@media (width < 45em) {
+    .foo { }
+}
+```
+
+If SCSS were to compile the code the way we instructed, it would break the
+media query. Within @rules, we still need to use mixins. In most cases, this
```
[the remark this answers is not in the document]
ThiefMaster: @foxbunny ^

indico/web/client/styles/design_system/_utils.scss
```diff
@@ -0,0 +1,16 @@
...
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+%reset-font-and-color {
+  color: inherit;
+  font: inherit;
+}
+
+%default-focus-ring {
+  outline: medium solid Highlight;
```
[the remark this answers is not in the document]
ThiefMaster: @foxbunny ^

indico/web/client/styles/partials/_icons.scss
```diff
@@ -151,7 +152,8 @@ i[class^='icon-']::before {
 .icon-checkbox-unchecked::before {
   content: '\e433';
 }
-.icon-checkmark::before {
+.icon-checkmark::before,
+%icon-checkmark {
```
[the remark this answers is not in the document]
ThiefMaster: @foxbunny ^

ThiefMaster:
I pinged you in the missed threads

foxbunny:
Gotcha, thanks!

indico/web/client/styles/partials/_icons.scss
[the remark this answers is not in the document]
foxbunny: Ok, let me see what I can do.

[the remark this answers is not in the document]
foxbunny: Btw, I was intending to duplicate all of those as placeholders. I can probably come up with a way to automate that so that it's not (as) painful to update.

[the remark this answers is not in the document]
foxbunny: Ok, there we go.

https://codepen.io/hayavuk/pen/KKbGzbW

I'll include this in the instructions.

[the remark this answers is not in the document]
foxbunny: @ThiefMaster Please try it before I push the changes, let me know if it works for you.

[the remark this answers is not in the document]
foxbunny: What it does is:

- Changes all selectors to use dashes instead of underscores
- Lower-cases all selectors
- Changes double to single quotes
- Adds the placeholders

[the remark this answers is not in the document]
foxbunny: I'm noticing that the icomoon JSON you have in the repo contains some icons that aren't in the _icons.scss. Just an FYI.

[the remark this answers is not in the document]
ThiefMaster: The extra ones are just alias names that are added by default. The unfortunately need to be removed manually. Alternatively, maybe only the first class name for any given `content` value could be kept since that's already the one we're using anyway.

Your script to convert it works fine. I'd suggest to add it in `bin/maintenance/` though, to avoid relying on some external codepen.

It could actually work directly on the _icons.scss and modify the parts between `/* Icomoon's icon definitions */` and `/* End of icomoon's icon definitions */` (so you'd just paste the icomoon css there and then run that script to fix things)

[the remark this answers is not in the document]
ThiefMaster: Actually, even better than pasting into that area would be the script taking the icomoon css file as input (or having the user paste it into the script's stdin). In either case I would change the begin/end markers to something where it's clearer that it's meant for a machine and not a human.

`/* BEGIN GENERATED ICOMOON CSS */` and `/* END GENERATED ICOMOON CSS */` for example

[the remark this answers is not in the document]
foxbunny: I think that might be a separate PR. If we take this further, we're also relying on icomoon as an external service to convert the icon. Ideally you'd just generate the icon font locally from a folder containing the icon files and do that as part of the webpack build.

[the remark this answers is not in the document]
foxbunny: Basically, I suggest we just move with the codepen for now and optimize the workflow later.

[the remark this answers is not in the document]
foxbunny: You mean like duplicate the existing declaration specifically for the placeholder?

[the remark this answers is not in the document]
foxbunny: The reason I put it there is because `::before` selectors cannot be extended, btw.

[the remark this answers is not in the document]
foxbunny: Pushed some changes. Let me know if that's what you meant.

indico/modules/core/client/js/session_bar.js
```diff
@@ -80,6 +71,7 @@
       },
       events: {
         show() {
+          // eslint-disable-next-line no-undef
```
ThiefMaster, line 74: You can avoid both `eslint-disable-next-line` statements by importing lodash properly ;) `import _ from 'lodash';`

indico/web/client/styles/design_system/_form_controls.scss
```diff
@@ -0,0 +1,69 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2023 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+@use "utils";
+@use "layout";
```
ThiefMaster, line 9:
```suggestion
@use 'utils';
@use 'layout';
```

ThiefMaster: not sure why stylelint doen't catch that...

indico/web/client/styles/partials/_icons.scss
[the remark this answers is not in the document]
ThiefMaster: yes, exactly :)

indico/modules/core/client/js/session_bar.js
foxbunny: Hm. So why wasn't it imported to begin with? 🤔

foxbunny: Anyway, I wouldn't dwell on this cause it's going away once the pull-down menu hits the PR backlog.

indico/web/client/styles/design_system/_form_controls.scss
foxbunny: Yeah, weird.

indico/modules/core/client/js/session_bar.js
ThiefMaster: bad old legacy times ;)

indico/web/client/styles/design_system/_form_controls.scss
ThiefMaster: probably related to that - maybe they already removed it from the standard ruleset: https://stylelint.io/migration-guide/to-15

foxbunny: I'll solve the problem on my end.

image

indico/web/client/styles/design_system/_utils.scss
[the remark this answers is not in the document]
foxbunny: It's from here:

https://css-tricks.com/copy-the-browsers-native-focus-styles/

image

indico/web/client/styles/partials/_buttons.scss
```diff
@@ -164,7 +164,7 @@ $i-button-spacing: 0.3rem;
 
   & ~ .i-button,
   & ~ .group {
-    margin-left: $i-button-spacing;
+    margin-right: $i-button-spacing;
```
ThiefMaster, line 167: Are you sure this change is correct and intentional?

Before:
image

After:
image


So it's removing spacing that was meant to be there to separate buttons/groups

foxbunny: Hmmmm. Probably unintentional.

foxbunny: Reverted

indico/web/client/styles/partials/_buttons.scss
```diff
@@ -289,7 +289,7 @@ a.i-big-button {
 
 .i-button.signin {
```
ThiefMaster, line 290: you can remove this whole definition since the button no longer exists. it was from the legacy login page (59afb5b37) which no longer exists

ThiefMaster (approved):

PR #5915, opened 2023-09-01
Format accompanying persons names correctly

duartegalvao:
This PR fixes the formatting of the accompanying persons names to match the event's/user's name format settings.

### DIFF
--- a/indico/modules/events/registration/fields/accompanying.py
+++ b/indico/modules/events/registration/fields/accompanying.py
@@ -12,8 +12,15 @@ from marshmallow import ValidationError, fields, post_load, pre_load, validate
 from indico.core import signals
 from indico.core.marshmallow import mm
 from indico.modules.events.registration.fields.base import RegistrationFormBillableField
+from indico.modules.users.models.users import PersonMixin
 from indico.util.i18n import _
 from indico.util.marshmallow import not_empty
+
+
+class AccompanyingPerson(PersonMixin):
+    def __init__(self, entry):
+        self.first_name = entry['firstName']
+        self.last_name = entry['lastName']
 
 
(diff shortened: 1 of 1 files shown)


indico/modules/events/registration/fields/accompanying.py
```diff
@@ -79,15 +80,15 @@ def calculate_price(self, reg_data, versioned_data):
...
 
     def get_friendly_data(self, registration_data, for_humans=False, for_search=False):
-        def _format_person(entry):
-            first_name = entry['firstName']
-            last_name = entry['lastName']
-            return f'{first_name} {last_name}'
+        class _Person(PersonMixin):
+            def __init__(self, entry):
+                self.first_name = entry['firstName']
+                self.last_name = entry['lastName']
```
ThiefMaster, line 86: how about making this a `_AccompanyingPerson` (or similar) and defining it on the module level instead of redefining it every time this method gets called?

indico/modules/events/registration/fields/accompanying.py
```diff
@@ -12,10 +12,17 @@
 from indico.core import signals
 from indico.core.marshmallow import mm
 from indico.modules.events.registration.fields.base import RegistrationFormBillableField
+from indico.modules.users.models.users import PersonMixin
 from indico.util.i18n import _
 from indico.util.marshmallow import not_empty
 
 
+class _AccompanyingPerson(PersonMixin):
```
duartegalvao, line 20: would it be ok to make it a public class so I can use it in the cern access plugin as well?

ThiefMaster: sure

PR #5922, opened 2023-09-11
Add support for picture field in registration forms

SegiNyn:
This PR adds Support for Picture Field in registration Forms. 

It works similar to the SingleFileManager component, with the addition that it accepts only images, allows to display the uploaded image, allows to capture image from Camera and Crop image. 

2 libraries were used:
 https://www.npmjs.com/package/react-webcam and
 https://www.npmjs.com/package/react-cropper

Screenshot 2023-09-11 at 16 04 08

Screenshot 2023-09-11 at 16 08 29


indico/modules/events/registration/fields/base.py
```diff
@@ -45,6 +45,8 @@ class RegistrationFormFieldBase:
     is_invalid_field = False
     #: whether this field must not be "empty" (falsy value) if required
     not_empty_if_required = True
+    #: whether this field is associated with a file and should be represented as a picture
+    is_picture_field = False
```
ThiefMaster, line 49: Based on a quick glance at the current usage of `is_file_field` you can use that one instead and do not need to create a new attribute. That way you'll also automatically get the proper behavior e.g. when purging a field due to the retention period having been reached.

indico/web/client/js/react/components/pictures/PictureWebcam.jsx
```diff
@@ -0,0 +1,67 @@
...
+            ref={cameraRef}
+            audio={false}
+            height={height}
+            screenshotFormat="image/png"
+            width={width}
+            onUserMedia={() => setUserMediaStarted(true)}
+            onUserMediaError={onUserMediaError}
+            mirrored
+          />
+          <p>click anywhere on the image to take a picture</p>
```
GovernmentPlates, line 37: No i18n?

indico/modules/events/registration/fields/base.py
SegiNyn: Since the picture field is inheriting from the FileField it also has its `is_file_field` set to True. But sure I can remove it since we can just check the fields input_type to know it is a picture.

ThiefMaster: did you even use it somewhere? at least i didn't find it when looking at the PR diff

indico/web/client/js/react/components/pictures/PictureArea.jsx
```diff
@@ -0,0 +1,181 @@
...
+  cropper: {pictureCropper: PictureCropper, isEditActive},
+  picture,
+  dragText,
+  picturePlaceholder,
+  pictureAction,
+  picturePreview,
+}) {
+  const uploadButtonIcon = picture ? 'exchange' : 'upload';
+  const uploadButtonText = picture
+    ? Translate.string('Choose new picture from your computer')
```
GovernmentPlates, line 40:
```suggestion
    ? Translate.string('Choose a new picture from your computer')
```

indico/web/client/js/react/components/pictures/PictureArea.jsx
```diff
@@ -0,0 +1,181 @@
...
+  picture: pictureDetailsShape,
+  dragText: PropTypes.string,
+  picturePlaceholder: PropTypes.string,
+  pictureAction: fileActionShape,
+  picturePreview: PropTypes.func.isRequired,
+};
+
+PictureArea.defaultProps = {
+  picture: null,
+  dragText: Translate.string('Drag picture here'),
```
GovernmentPlates, line 178:
```suggestion
  dragText: Translate.string('Drag a picture here'),
```

indico/web/client/js/react/components/pictures/PictureManager.jsx
```diff
@@ -0,0 +1,365 @@
...
+      imageSrc = reader.result;
+      dispatch({type: 'START_EDITING', imageSrc});
+    });
+    reader.readAsDataURL(file);
+  };
+
+  const dropzone = useDropzone({
+    onDropAccepted,
+    disabled: disabled || (isUploading || failed || isCapturing || isEditing),
+    accept: ['.png', '.jpg', '.gif'],
```
GovernmentPlates, line 247: As a suggestion: would it be possible to indicate which file types are accepted on the dropzone UI (i.e. 'accepted files: .jpg, .png, .gif.')?

SegiNyn: I was passing it explicitly since we wouldn't really want something else that isn't an image to be uploaded here. But in what way would you suggest to indicate this? a comment?

GovernmentPlates: In terms of indicating this, I would suggest adding an extra line of text on the dropzone UI.

GovernmentPlates: Adding onto this: [this might also be useful](https://ux.stackexchange.com/a/112618).

SegiNyn: Ah thanks, I see. Good idea, I'll add the extra text

CHANGES.rst
```diff
@@ -32,6 +32,7 @@ Improvements
 - Improve login page UI, allow overriding the logo URL (:data:`LOGIN_LOGO_URL` config option)
   and using custom logos for auth providers (``logo_url`` in the auth provider settings)
   (:pr:`5936`, thanks :user:`openprojects`)
+- Add support for Picture fields in registration forms (:pr:`5922` :user:`SegiNyn`)
```
ThiefMaster, line 35:
```suggestion
- Add support for Picture fields in registration forms (:pr:`5922`, thanks :user:`SegiNyn`)
```

indico/modules/events/registration/fields/simple.py
```diff
@@ -389,3 +390,24 @@ def _indico_email(value):
                 raise ValidationError(_('Invalid email address'))
 
         return _indico_email
+
+
+class PictureField(FileField):
```
ThiefMaster, line 395: I think you want `mm_field_class = MMFileField` (and `from indico.util.marshmallow import FileField as MMFieldField`) - that way you do not need to manually query the file below but get it nicely as the field's `value`

indico/modules/events/registration/templates/display/_registration_summary_blocks.html
```diff
@@ -99,6 +99,8 @@
         </ul>
     {% elif field.input_type == 'file' and friendly_data and from_management %}
         <a href="{{ url_for('.registration_file', data[field.id].locator.file) }}">{{ friendly_data }}</a>
+    {% elif field.input_type == 'picture' and friendly_data %}
+        <img class="picture-preview" src="{{ url_for('.registration_picture', data[field.id].locator.file) }}" alt={{ friendly_data }}>
```
ThiefMaster, line 103: does the `alt` have anything useful in it? Also, it almost certainly needs to be quoted

```suggestion
        <img class="picture-preview" src="{{ url_for('.registration_picture', data[field.id].locator.file) }}" alt="{{ friendly_data }}">
```

indico/web/client/js/react/components/pictures/PictureCropper.jsx
```diff
@@ -0,0 +1,86 @@
...
+import React from 'react';
+import Cropper from 'react-cropper';
+import {Header, Icon, Button} from 'semantic-ui-react';
+
+import 'cropperjs/dist/cropper.css';
+import './Picture.module.scss';
+
+export function PictureCropper({cropperRef, src, onCrop, backAction, minCropSize}) {
+  let size = Math.max(minCropSize, 500);
+  size += 200;
```
ThiefMaster, line 18:
```suggestion
  const size = Math.max(minCropSize, 500) + 200;
```

also, a comment on why these numbers are used might be useful...

indico/web/client/js/react/components/pictures/PictureCropper.jsx
```diff
@@ -0,0 +1,86 @@
...
+          guides
+          rotatable
+        />
+        <Icon
+          styleName="back"
+          name="arrow left"
+          color="grey"
+          size="large"
+          onClick={e => {
+            e.stopPropagation();
```
ThiefMaster, line 46:
```suggestion
          onClick={evt => {
            evt.stopPropagation();
```

more consistent with the rest of the codebase

indico/web/client/js/react/components/pictures/PictureCropper.jsx
```diff
@@ -0,0 +1,86 @@
...
+          }}
+        />
+        <Button.Group>
+          <Button
+            styleName="cropper-action-btn"
+            icon
+            float="left"
+            color="blue"
+            onClick={e => {
+              e.stopPropagation();
```
ThiefMaster, line 57:
```suggestion
            onClick={evt => {
              evt.stopPropagation();
```

same as above

indico/web/client/js/react/components/pictures/PictureCropper.jsx
```diff
@@ -0,0 +1,86 @@
...
+              if (cropperRef.current.cropper) {
+                cropperRef.current.cropper.rotate(90);
+              }
+            }}
+          >
+            Rotate
```
ThiefMaster, line 63: i18n

indico/web/client/js/react/components/pictures/PictureCropper.jsx
```diff
@@ -0,0 +1,86 @@
...
+            <Icon name="sync" />
+          </Button>
+          <Button float="left" styleName="cropper-action-btn" icon color="blue" onClick={onCrop}>
+            Done
```
ThiefMaster, line 67: i18n

indico/web/client/js/react/components/pictures/PictureManager.jsx
```diff
@@ -0,0 +1,400 @@
...
+  const cropper = {
+    pictureCropper: Cropper,
+    isEditActive: isEditing,
+  };
+  const descriptionText = minPictureSize
+    ? Translate.string(
+        'Minimum picture width or height: {minPictureSize} pixels. Allowed file types: (.jpg, .png, .gif).',
+        {minPictureSize}
+      )
+    : Translate.string('Allowed file types: .jpg, .png, .gif.');
```
ThiefMaster, line 341:
```suggestion
    : Translate.string('Allowed file types: .jpg, .png, .gif');
```

The period at the end looks really strange.

indico/web/client/js/react/components/pictures/PictureArea.jsx
```diff
@@ -0,0 +1,240 @@
...
+            </Grid.Row>
+          </Grid>
+        )}
+        {(fileRejections.length > 0 || customFileRejections) && (
+          <Message negative>
+            <Message.Header>
+              <Translate>There were some problems with the picture</Translate>
+            </Message.Header>
+            {fileRejections.map(({file, errors}) => (
+              <Message.List key={file.size}>
```
ThiefMaster, line 185: using the file size as the key is really strange. is there nothing better that's actually unique for each file?

indico/web/client/js/react/components/pictures/PictureArea.jsx
```diff
@@ -0,0 +1,240 @@
...
+                {errors.map(e => (
+                  <Message.Item key={e.code}>
+                    {file.name}:- {e.message}
+                  </Message.Item>
+                ))}
+              </Message.List>
+            ))}
+            {customFileRejections && (
+              <Message.List>
+                <Message.Item key={customFileRejections.code}>
```
ThiefMaster, line 195: no need for a key here

indico/modules/events/registration/templates/display/_registration_summary_blocks.html
SegiNyn: The information in the alt is the uploaded file name.

indico/web/client/js/react/components/pictures/PictureArea.jsx
SegiNyn: They're only passing the uploaded file and then the errors. I see examples where they use the file path as the key, is this acceptable?

indico/modules/events/registration/fields/simple.py
SegiNyn: I am getting a disparity here because the `PictureField` inherits from the `FileField` which uses `mm_field_class = UUIDString` so If I change the `mm_field_class` here I would also then need to define my own `process_form_data` function for it to not break something.

SegiNyn: @ThiefMaster, please do you have any other suggestion for this?

ThiefMaster: Nevermind in that case - I didn't realize we're doing the exact same thing for the `FileField`.... something to refactor later I guess

indico/modules/events/registration/fields/simple.py
```diff
@@ -388,3 +389,24 @@ def _indico_email(value):
...
+    name = 'picture'
+    setup_schema_fields = {
+        'min_picture_size': fields.Integer(load_default=0, validate=validate.Range(0, 1200)),
+    }
+
+    def get_validators(self, existing_registration):
+        def _picture_size_and_type(value):
+            min_picture_size = self.form_item.data.get('min_picture_size') or None
+            file = File.query.filter(File.uuid == value, ~File.claimed).first() if value else None
```
ThiefMaster, line 403:
```
if not file:
    return
```

instead of repeating the `file and` below all the time.

even better:

```python
if not value:
    return
file = ...
if not file:
    raise ValidationError('Invalid file')
```

This should not happen (hence no need for i18n on that error), but it's a bit cleaner than just silently not doing anything if the UUId does not point to a valid file.

indico/web/client/js/react/components/pictures/PictureCropper.jsx
SegiNyn: The default minsize of 500 is to allow the field fit well inside the registration form. And the 200 is to offset the min values set by the users. So the total default size is 700. For example, if a minSize of 700 is set in the field settings then the picture field is 900 so that there's enough area to crop or select the image.

indico/modules/events/registration/fields/simple.py
SegiNyn: okay

indico/modules/events/registration/fields/simple.py
```diff
@@ -388,3 +389,24 @@ def _indico_email(value):
...
+            if file and file.content_type not in ('image/png', 'image/jpg', 'image/gif'):
```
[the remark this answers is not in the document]
SegiNyn: Yes it can be included. I can include it in the accepted file types for the drop zone as well. Although because the images go through the cropper and that one produces a png image only the png image makes it to the serverside validator

indico/modules/events/registration/fields/simple.py
SegiNyn: ok thanks, changed it to not check the file multiple times

indico/modules/events/registration/fields/simple.py
```diff
@@ -388,3 +389,28 @@ def _indico_email(value):
...
+            if not value:
+                return
+            file = File.query.filter(File.uuid == value, ~File.claimed).first() if value else None
+            if not file:
+                raise ValidationError('Invalid file')
+            if file.content_type not in ('image/png', 'image/jpg', 'image/gif', 'image/webp'):
+                raise ValidationError(_('This field can accept only .jpg, .png, .gif and .webp picture formats.'))
+            min_picture_size = self.form_item.data.get('min_picture_size')
+            if min_picture_size:
+                with Image.open(file.open()) as picture:
```
ThiefMaster, line 411: with file.open() as f, Image.open(f) as img: ...

not sure if this works, if not use two separate nested with blocks. But the File should be opened using a context manager just like you'd do for a normal file on disk

CHANGES.rst
```diff
@@ -64,6 +64,8 @@ Improvements
 - Add option to allow people to register using an invitation link even if the event is
   restricted (:pr:`6094`)
 - Improve editing notifications emails (:issue:`6027`, :pr:`6042`)
+- Add support for Picture fields in registration forms (:pr:`5922` :user:`SegiNyn`)
```
ThiefMaster, line 67: remove this, looks like it got (re-)added during a rebase..

indico/web/client/js/react/components/pictures/PictureManager.jsx
```diff
@@ -0,0 +1,406 @@
...
+  }
+
+  useEffect(() => {
+    if (!setValidationError) {
+      return;
+    }
+    if (isUploading) {
+      setValidationError(Translate.string('Upload in progress'));
+    } else if (failed) {
+      setValidationError(Translate.string('Picture upload or Capture failed'));
```
ThiefMaster, line 319:
```suggestion
      setValidationError(Translate.string('Picture upload/capture failed'));
```

indico/web/client/js/react/components/pictures/PictureManager.jsx
```diff
@@ -0,0 +1,406 @@
...
+    } else if (isCapturing) {
+      setValidationError(Translate.string('Picture is being captured'));
+    } else if (isEditing) {
+      setValidationError(Translate.string('Picture editing in progress'));
```
ThiefMaster, line 323:
```suggestion
      setValidationError(Translate.string('Picture is being edited'));
```

consistency

indico/web/client/js/react/components/pictures/PictureManager.jsx
```diff
@@ -0,0 +1,406 @@
...
+  };
+
+  const cropper = {
+    pictureCropper: Cropper,
+    isEditActive: isEditing,
+  };
+
+  const descriptionText = minPictureSize
+    ? Translate.string(
+        'Minimum picture width or height: {minPictureSize} pixels. Allowed file types: (.jpg, .png, .gif, .webp)',
```
ThiefMaster, line 343:
```suggestion
        'Minimum picture width or height: {minPictureSize} pixels. Allowed file types: jpg, png, gif, webp',
```

indico/web/client/js/react/components/pictures/PictureManager.jsx
```diff
@@ -0,0 +1,406 @@
...
+        {minPictureSize}
+      )
+    : Translate.string('Allowed file types: .jpg, .png, .gif');
```
ThiefMaster, line 346:
```suggestion
    : Translate.string('Allowed file types: jpg, png, gif, webp');
```

(i guess you just forgot webp here?)

indico/modules/events/registration/fields/simple.py
```diff
@@ -388,3 +389,28 @@ def _indico_email(value):
...
+
+    def get_validators(self, existing_registration):
+        def _picture_size_and_type(value):
+            if not value:
+                return
+            file = File.query.filter(File.uuid == value, ~File.claimed).first() if value else None
+            if not file:
+                raise ValidationError('Invalid file')
+            if file.content_type not in ('image/png', 'image/jpg', 'image/gif', 'image/webp'):
+                raise ValidationError(_('This field can accept only .jpg, .png, .gif and .webp picture formats.'))
```
ThiefMaster, line 408: "This field only accepts..."

ThiefMaster, line 408: I would rather check this using the content like we do elsewhere...

```
            try:
                pic = Image.open(...)
            except OSError:
                raise ...
            if pic.format.lower() not in {'jpeg', 'png', 'gif', 'webp'}:
                raise ...
```

indico/modules/events/registration/fields/simple.py
```diff
@@ -388,3 +389,28 @@ def _indico_email(value):
...
+    name = 'picture'
+    setup_schema_fields = {
+        'min_picture_size': fields.Integer(load_default=0, validate=validate.Range(0, 1200)),
+    }
+
+    def get_validators(self, existing_registration):
+        def _picture_size_and_type(value):
+            if not value:
+                return
+            file = File.query.filter(File.uuid == value, ~File.claimed).first() if value else None
```
ThiefMaster, line 404: ` if value else None` is not needed since you check it above

SegiNyn:
Thanks for review @ThiefMaster, I would just like to make transparent these 3 changes which I added for the plugin:
[Add hook for plugins below FormItem](https://github.com/indico/indico/pull/5922/commits/1b3c01f56fa564bd746a9812104de4c18f84ffec)
[Export indico react form fields for plugins](https://github.com/indico/indico/pull/5922/commits/d205945cfa43796dc53b6b6441ee620c195bc6b5)
[Add block around registration form](https://github.com/indico/indico/pull/5922/commits/f27e4c5414634920d4b2edbd182850bea46925b3)

indico/modules/events/registration/client/js/form/fields/PictureInput.jsx
```diff
@@ -0,0 +1,85 @@
...
+import {getUpdateMode} from '../../form_submission/selectors';
+import {getStaticData} from '../selectors';
+
+import '../../../styles/regform.module.scss';
+import './FileInput.module.scss';
+
+export default function PictureInput({htmlName, disabled, isRequired, minPictureSize}) {
+  const {eventId, regformId, registrationUuid, fileData} = useSelector(getStaticData);
+  const isUpdateMode = useSelector(getUpdateMode);
+  const initialPictureDetails = isUpdateMode ? fileData[htmlName] || null : null;
```
SegiNyn, line 28: Would it be okay instead of checking the isUpdateMode to change this to:

`const initialPictureDetails = fileDate ? fileData[htmlName] || null : null;`

ThiefMaster: would be OK i guess...

SegiNyn: okay then I've changed it

indico/web/client/js/react/components/pictures/PictureArea.jsx
```diff
@@ -0,0 +1,238 @@
...
+  picturePlaceholder: PropTypes.string,
+  pictureAction: fileActionShape,
+  picturePreview: PropTypes.func.isRequired,
+  customFileRejections: PropTypes.object,
+};
+
+PictureArea.defaultProps = {
+  picture: null,
+  dragText: Translate.string('Drag a picture here'),
+  descriptionText: Translate.string('Allowed file types: .jpg, .png, .gif.'),
```
ThiefMaster, line 234: This default is not used anywhere, so having a translatable string here wastes translators' time. Maybe default to empty instead?

indico/web/client/js/react/components/pictures/PictureArea.jsx
```diff
@@ -0,0 +1,238 @@
...
+        {(fileRejections.length > 0 || customFileRejections) && (
+          <Message negative>
+            <Message.Header>
+              <Translate>There were some problems with the picture</Translate>
+            </Message.Header>
+            {fileRejections.map(({file, errors}) => (
+              <Message.List key={file.path}>
+                {errors.map(e => (
+                  <Message.Item key={e.code}>
+                    {file.name}:- {e.message}
```
ThiefMaster, line 188:
```suggestion
                    {file.name}: {e.message}
```

indico/web/client/js/react/components/pictures/PictureManager.jsx
```diff
@@ -0,0 +1,406 @@
...
+    pictureCropper: Cropper,
+    isEditActive: isEditing,
+  };
+
+  const descriptionText = minPictureSize
+    ? Translate.string(
+        'Minimum picture width or height: {minPictureSize} pixels. Allowed file types: (jpg, png, gif, webp)',
+        {minPictureSize}
+      )
+    : Translate.string('Allowed file types: jpg, png, gif, webp');
```
ThiefMaster, line 346: This does not look great and is actually kind of unnecessary since:

- when using the file upload dialog, the proper filter is already set
- when dropping an invalid file, we show an error

So in the end the whole description text could just be omitted.

The minimum size can probably be handled by a validation error as well -> no need to show it in advance

indico/web/client/js/react/components/pictures/PictureArea.jsx
```diff
@@ -0,0 +1,238 @@
...
+                        color={pictureAction.color}
+                        onClick={() => pictureAction.onClick(picture)}
+                      />
+                    )}
+                  </Card>
+                </Grid.Column>
+              )}
+              <Grid.Column verticalAlign="middle" width={!isDragActive && picture ? 6 : 16}>
+                {!picture && (
+                  <Icon name={picturePlaceholder} size="massive" styleName="picture-placeholder" />
```
ThiefMaster, line 87: I would get rid of this. It just wastes lots of space (the upload widgets are already huge enough...) and has no real benefit

ThiefMaster:
image

The alignment of those buttons is rather weird

ThiefMaster:
Can we make the 'undo' arrow look a bit nicer? Compared to the normal file upload widget it looks rather out-of-place:

image

The button alignment is weird in this view as well btw.

ThiefMaster:
The first click on the "Rotate" button doesn't do anything

indico/web/client/js/react/components/pictures/PictureArea.jsx
SegiNyn: Okay, defaulting to empty for the `descriptionText` but leaving the `dragText` since that is being used

indico/web/client/js/react/components/pictures/PictureManager.jsx
SegiNyn: Indeed the texts is quite long and I have no strong preferences about keeping it

btw there is a frontend validation for the minimum size and it appears like this below: 
 
Screenshot 2024-01-17 at 13 10 37

indico/web/client/js/react/components/pictures/PictureArea.jsx
SegiNyn: okay

indico/web/client/js/react/components/pictures/PictureManager.jsx
ThiefMaster: Nice, let's get rid of this text altogether then

ThiefMaster:
I pushed a fix since accessing a registration (and thus its picture) via a token was not handled correctly. Also used that opportunity to cleanup a bit (e.g. reusing the file download endpoint in management since it does no need special handling of pictures).

SegiNyn:
> Can we make the 'undo' arrow look a bit nicer? Compared to the normal file upload widget it looks rather out-of-place:

Is this better, moved the icon inside the picture card and added a white background color around it so it wouldn't disappear in the image beneath: 

Screenshot 2024-01-17 at 13 57 15

ThiefMaster:
much nicer IMHO!

PR #5925, opened 2023-09-12
Support for accepting terms on user registration

### LINKED ISSUE #5923: Force user to accept a privacy policy when creating a profile
We'd like to set expectations when the user signs up, and give them the chance to bail before a profile is created if they aren't happy with the terms. We need the user to confirm a privacy policy beforehand.

On https://indico.instance/register/ssoprovider I'd like there to be a checkbox that asks the user to confirm an configurable text with a checkbox, which in my case I'd use to link a privacy policy. I will likely need the flexibility to set arbitrary text and links, as it is important to write down what they are confirming.

The field should be sent to the server and verified there, so that people can't simply say they've disabled the checkbox and have not accepted the terms.

I'd like to simply do this in a plugin, since I imagine there might be a number of additional things that different instance owners might want to have on the signup page, but I haven't found a way to actually extend React forms and components from a plugin (something I really hate about React and webpack, previously coming from the browser extensions world. Hope I am just wrong!) 

**Additional context**

image

kewisch:
Fixes #5923 .


indico/modules/auth/controllers.py
```diff
@@ -453,6 +454,7 @@ class Meta:
                 affiliation = fields.String(load_default='')
             phone = fields.String(load_default='')
             affiliation_link = ModelField(Affiliation, data_key='affiliation_id', load_default=None)
+            accept_registration_terms = fields.Bool(required=True)
```
ThiefMaster, line 457: this will fail validation when there are no terms set and thus nothing is sent for the field

```suggestion
            if legal_settings.get('register_terms_prompt'):
                accept_registration_terms = fields.Bool(required=True)
```

(untested)

kewisch:
fwiw after some renaming I just noticed that it doesn't quite fit under privacy policy any more. I don't think it fits in any of the others (though maybe the Terms one), and a new section could be overkill.

I'll leave it to your judgement where it fits best.

indico/modules/auth/controllers.py
```diff
@@ -472,6 +474,11 @@ def check_email_unique(self, email, **kwargs):
                 if User.query.filter(~User.is_deleted, ~User.is_pending, User.all_emails == email).has_rows():
                     raise ValidationError('Email already in use')
 
+            @validates('accept_registration_terms')
+            def check_registration_terms(self, value):
+                if not value:
+                    raise ValidationError('Terms must be accepted')
```
ThiefMaster, line 480: You could also pass `validate=validate_with_message(bool, 'Terms must be accepted')` in the field constructor above. Or even use `validate=not_empty` since the user should never see the error message anyway ;)

indico/modules/legal/forms.py
```diff
@@ -24,3 +24,10 @@ class LegalMessagesForm(IndicoForm):
                                   description=_('The URL to an external page with the privacy policy'))
     privacy_policy = TextAreaField(_('Text'), widget=CKEditorWidget(),
                                    description=_('Only used if no URL is provided'))
+    register_terms_prompt = TextAreaField(
+        _('Registration page terms prompt'),
+        description=_(
+            'If set, user is required to confirm this message before creating a profile. You can use markdown or basic \
+             HTML formatting. Including a link to /tos or /privacy allows you to link to above policies.'
+        )
+    )
```
ThiefMaster, line 33: I'd try to keep the wrapping style consistent within a single form class

indico/modules/auth/controllers.py
kewisch: Nice, using validate=not_empty now.

indico/modules/auth/controllers.py
kewisch: Good catch!

indico/modules/users/models/users.py
```diff
@@ -277,6 +277,12 @@ class User(PersonMixin, db.Model):
         nullable=False,
         default=False
     )
+    #: the date the user has last accepted the terms of service
+    accepted_tos_date = db.Column(
+        db.DateTime,
```
ThiefMaster, line 282: if you kept the time component (didn't check that yet), use our `UTCDateTime` like we use everywhere else. if not, use the native `Date`

indico/modules/legal/__init__.py
```diff
@@ -51,3 +57,33 @@ def _inject_privacy_footer(**kwargs):
     url = legal_settings.get('privacy_policy_url')
     if url or legal_settings.get('privacy_policy'):
         return render_template('legal/privacy_footer.html', url=url)
+
+
+@signals.core.app_created.connect
+def _handle_terms_changed(app, **kwargs):
+    @app.before_request
+    def _terms_if_not_accepted():
+        terms_date_str = legal_settings.get('terms_effective_date')
```
ThiefMaster, line 66: You can add converters to a SettingsProxy. we already have `DatetimeConverter`, should be very easy to create a `DateConverter` based on that

indico/modules/users/models/users.py
kewisch: Ah nice, didn't know about that one, will do. I kept the time component for when the user accepts so there is a specific point in time, but the admin value is just a date which uses the server timezone.

indico/modules/legal/controllers.py
```diff
@@ -38,3 +63,33 @@
...
+            return redirect(url_for_index())
+
+        form = AgreementForm()
+
+        if form.validate_on_submit():
+            if not request.args.get('preview', False) and form.data['accept_terms']:
+                session.user.accepted_tos_date = now_utc()
+
+            target = request.args['return'] if request.args.get('return', '').startswith('/') else url_for_index()
+            return redirect(target)
```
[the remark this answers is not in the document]
kewisch: @ThiefMaster let me know if you are concerned about this. I specifically added a check if it starts with `/` to ensure it is a relative URL, but there might still be some cases I didn't anticipate.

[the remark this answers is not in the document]
ThiefMaster: `?return=//evil.com/phishing.html` ;)

so nope, that check is not sufficient right now

maybe just put the return URL in the session and not pass it around visibly at all

indico/modules/legal/templates/manage_messages.html
```diff
@@ -7,12 +7,25 @@
...
-        {{ form_rows(form, skip=('tos_url', 'tos', 'privacy_policy_url', 'privacy_policy')) }}
+        {{ form_rows(
+               form,
+               skip=(
+                   "tos_url",
+                   "tos",
+                   "privacy_policy_url",
+                   "privacy_policy",
+                   "terms_require_accept",
+                   "terms_effective_date",
```
ThiefMaster, line 18: single quotes please :) (they are superior! 😛)

indico/modules/users/models/users.py
```diff
@@ -277,6 +277,12 @@ class User(PersonMixin, db.Model):
         nullable=False,
         default=False
     )
+    #: the date the user has last accepted the terms of service
+    accepted_tos_date = db.Column(
```
ThiefMaster, line 281: I'd name this `accepted_tos_dt` since it's a datetime

indico/modules/legal/templates/manage_messages.html
```diff
@@ -7,12 +7,25 @@
 
 {% block content %}
     {% call simple_form(form, back_button=false) %}
-        {{ form_rows(form, skip=('tos_url', 'tos', 'privacy_policy_url', 'privacy_policy')) }}
+        {{ form_rows(
+               form,
```
ThiefMaster, line 11: not a fan of this wrapping style tbh :/

indico/modules/legal/templates/agreement.html
```diff
@@ -0,0 +1,92 @@
+{% extends 'layout/base.html' %}
+{% from 'forms/_form.html' import form_header, form_footer, form_rows, form_row_static, form_fieldset %}
+
+{% block page_class %}fixed-width-standalone-text-page{% endblock %}
+
+{% macro tos_field_reference(tos, tos_url) -%}
+  {%- if tos_url -%}
```
ThiefMaster, line 7: we usually use 4-space indentation in templates

indico/modules/legal/templates/agreement.html
```diff
@@ -0,0 +1,92 @@
...
+  }
+  .submit {
+    float: right;
+  }
+</style>
+<p>
+  {% trans %}To continue using the platform, you must agree to the updated terms. Please take a moment to read the agreements before you continue.{% endtrans %}
+</p>
+{% if tos and not tos_url %}
+  {% call form_fieldset('Terms of Service') %}
```
ThiefMaster, line 51: i18n

indico/modules/legal/templates/agreement.html
```diff
@@ -0,0 +1,92 @@
...
+{% endif %}
+{% if privacy_policy and not privacy_policy_url %}
+  {% call form_fieldset('Privacy Policy') %}
+  <div class="field policy">
+    {{ privacy_policy|sanitize_html }}
+  </div>
+  {% endcall %}
+{% endif %}
+<p class="last-updated">
+  Last updated: {{ terms_effective_date|format_date(timezone=session.tzinfo) }}
```
ThiefMaster, line 65: i18n

indico/modules/legal/templates/manage_messages.html
kewisch: Say thank you to `black` which formatted this block (and also is responsible for the double quotes :). I'll try a different approach.

indico/modules/legal/templates/agreement.html
```diff
@@ -0,0 +1,92 @@
...
+    {%- endif -%}
+{%- endmacro %}
+
+{% block content %}
+{{ form_header(form) }}
+<style class="TODO">
+    /* Leaving this here for some advice during the review. I'd expect it would be easy to create a
+     * form where the labels are skipped but the required * remains, and the submit button is on the
+     * right side. It doesn't seem so. What am I missing? It doesn't seem like I should need to use
+     * these styles to make it look pretty. */
```
kewisch, line 32: Highlighting this comment for your reviewing pleasure :)

kewisch:
@ThiefMaster I've hit a snag on this while debugging something else and could use your advice:

This approach seems to fail for urls that have `@allow_signed_url`. In the intercepting signal there is a check for `session.user`, which calls `indico.web.util`'s `get_request_user`, which will fail because `g.rh` is not yet set, so it can't actually see that the controller has _ALLOW_SIGNED_URL.

Is there a way to do this interception later, or a better way to check for the user at this stage?

indico/modules/legal/__init__.py
```diff
@@ -51,3 +56,34 @@ def _inject_privacy_footer(**kwargs):
...
+            return
+
+        if session.user.accepted_tos_dt and session.user.accepted_tos_dt >= terms_date:
+            return
+
+        if request.endpoint:
+            prefix = request.endpoint.split('.', 1)[0]
+
+            # Certain endpoints need to be passed through.
+            if prefix in ('assets', 'legal', 'auth', 'core'):
```
ThiefMaster, line 80:
```suggestion
            # Certain endpoints need to be passed through.
            if request.blueprint in ('assets', 'legal', 'auth', 'core'):
```

indico/modules/legal/__init__.py
```diff
@@ -51,3 +56,34 @@ def _inject_privacy_footer(**kwargs):
...
+                return
+
+            # Admins get a free pass on the admin interface
+            if session.user.is_admin and request.path.startswith('/admin/'):
+                return
```
ThiefMaster, line 85: TBH I'd let them completely ignore this and just show a persistent notification banner on top

ThiefMaster:
Do you mean the time when `_handle_terms_changed` runs? If yes, you could move it into the RH class right after `self._check_csrf()`

indico/modules/legal/controllers.py
```diff
@@ -5,22 +5,38 @@
...
-            legal_settings.set_multi(form.data)
+            data = form.data
+
+            has_privacy_policy = data['privacy_policy'] or data['privacy_policy_url']
+            has_tos = data['tos'] or data['tos_url']
+
+            if data['terms_require_accept'] and not has_privacy_policy and not has_tos:
+                flash(_('Requiring to accept terms requires a privacy policy '
+                        'and/or terms of service being set'), 'error')
+                return redirect(url_for('legal.manage'))
```
ThiefMaster, line 35: I would move this into the `LegalMessagesForm` and fail nicely on the validation level. Flash+redirect means any data changed/entered in the form is lost

kewisch:
> Do you mean the time when `_handle_terms_changed` runs? If yes, you could move it into the RH class right after `self._check_csrf()`

Yeah, this works much better. It feels a little unclean, but I'll leave that to the review if it works out good enough :)

@ThiefMaster also friendly reminder for this comment: https://github.com/indico/indico/pull/5925#discussion_r1359312781

indico/modules/legal/__init__.py
ThiefMaster: For banners on top? The global announcement or user impersonation status for example.

indico/modules/legal/controllers.py
```diff
@@ -38,3 +54,32 @@ def _process(self):
...
+            return redirect(url_for_index())
+
+        form = AgreementForm()
+
+        if form.validate_on_submit():
+            if not request.args.get('preview', False) and form.data['accept_terms']:
+                session.user.accepted_tos_dt = now_utc()
+
+            returnpath = session.get('legal_agreement_return_path', url_for_index())
+            del session['legal_agreement_return_path']
```
ThiefMaster, line 78: `returnpath = session.pop(...)`

indico/modules/legal/templates/agreement.html
ThiefMaster: I think if you want non-standard form formatting you're way better off doing it in react. Anyway, for a first functioning version (to test all the logic etc) I think we can keep the suboptimal styling

kewisch: Well, I would expect this is a somewhat standard case, but apparently it hasn't come up yet. So just leave out the styles now? Here is what it looks like without:

image

and with:

image

ThiefMaster: I would get rid of the `*` tbh - it's pretty much clear that the field is required, and it looks kind of out of place...

Styling wise I think it look much better if the checkbox is close to the left.

ThiefMaster: Oh, have a look at `skip_label=true` in the `form_row()` macro call. I think that should disable the `*` but still keep the normal checkbox label since it's handled elsewhere

kewisch: Yeah I did see `skip_label`, but it removes the `*` as well. If you'd like it removed, that makes things easier. I've found an elegant way under those circumstances. Thanks!

kewisch:
All current review comments taken care of :)

indico/modules/auth/client/js/signup.jsx
```diff
@@ -67,6 +74,34 @@ function Signup({
     await new Promise(() => {});
   };
 
+  const termsAcceptLabel = () => {
+    const tosElement = tosUrl ? (
+      <a href="/tos" target="_blank">
```
ThiefMaster, line 79: no hardcoded hrefs please :)

indico/modules/auth/client/js/signup.jsx
```diff
@@ -67,6 +74,34 @@ function Signup({
...
+        {Translate.string('terms of service')}
+      </a>
+    ) : (
+      Translate.string('terms of service')
+    );
+    const ppElement = privacyPolicyUrl ? (
+      <a href="/pp" target="_blank">
```
ThiefMaster, line 86: i don't think /pp exists...

indico/modules/auth/client/js/signup.jsx
```diff
@@ -223,6 +258,36 @@ function Signup({
...
+              {Translate.string('Last updated')}: {moment(termsEffectiveDate).format('LL')}
+            </div>
+          )}
+          {termsRequireAccept && (
+            <FinalCheckbox
+              name="accept_terms"
+              required
+              label={
+                <label>
+                  {Translate.string('I have read, understood, and accept the')} {termsAcceptLabel()}
```
ThiefMaster, line 284: same problem here. this does not work for i18n -> use a single Translate string for it

indico/modules/auth/client/js/signup.jsx
```diff
@@ -67,6 +74,34 @@ function Signup({
...
+      <a href="/pp" target="_blank">
+        {Translate.string('privacy policy')}
+      </a>
+    ) : (
+      Translate.string('privacy policy')
+    );
+    if ((tos || tosUrl) && (privacyPolicy || privacyPolicyUrl)) {
+      return (
+        <>
+          {tosElement} {Translate.string('and')} {ppElement}
```
ThiefMaster, line 95: please do not build translated strings like this. it doesn't work well depending on the language. it' possible to use `<Translate>` and `<Param>` to build this string without this "mess" (untested, though):

```js
import privacyPolicyURL from 'indico-url:legal.display_privacy';
import tosURL from 'indico-url:legal.display_tos';
```

```jsx
<Translate>
  I have read, understood, and accept the{' '}
  <Param name="tosLink" wrapper={<a href={tosURL()} target="_blank" />}>
    Terms of Service
  </Param>{' '}
  and{' '}
  <Param name="privacyLink" wrapper={<a href={privacyPolicyURL()} target="_blank" />}>
    Privacy Policy
  </Param>{' '}
</Translate>
```

for the other two variants, simply duplicate and adapt this block.

kewisch: Thanks, I wasn't aware of `<Param>`. I found a good way to do this in jinja as well.

kewisch:
@tomasr8 @ThiefMaster I updated the translations. Let me know if there is anything else or if this is good to go :)

indico/modules/legal/__init__.py
```diff
@@ -51,3 +56,44 @@ def _inject_privacy_footer(**kwargs):
     url = legal_settings.get('privacy_policy_url')
     if url or legal_settings.get('privacy_policy'):
         return render_template('legal/privacy_footer.html', url=url)
+
+
+@template_hook('global-announcement', markup=False)
+def _inject_announcement_header(**kwargs):
+    # Don't show if no user or not admin
+    if not session.user or not session.user.is_admin:
+        return
```
tomasr8, line 65: This _should_ be hidden for admins, right?

indico/web/rh.py
```diff
@@ -289,6 +293,10 @@ def process(self):
         try:
             init_email_queue()
             self._check_csrf()
+            terms_response = self._check_terms()
+            if terms_response:
+                return terms_response
```
tomasr8, line 298:
```suggestion
            if terms_response := self._check_terms():
                return terms_response
```

indico/modules/legal/templates/agreement.html
```diff
@@ -0,0 +1,54 @@
...
+</p>
+{% if tos and not tos_url %}
+    {% call form_fieldset(_('Terms of Service')) %}
+        <div class="field policy">
+            {{ tos|sanitize_html }}
+        </div>
+    {% endcall %}
+{% endif %}
+{% if privacy_policy and not privacy_policy_url %}
+    {% call form_fieldset('Privacy Policy') %}
```
tomasr8, line 19:
```suggestion
    {% call form_fieldset(_('Privacy Policy')) %}
```

indico/modules/legal/templates/agreement.html
```diff
@@ -0,0 +1,54 @@
...
+{% call form_row(form.accept_terms, skip_label=True) %}
+    {% set tos_link_start = ('<a href="' ~ url_for('legal.display_tos') ~ '">')|safe if tos_url else "" %}
+    {% set pp_link_start = ('<a href="' ~ url_for('legal.display_privacy') ~ '">')|safe if privacy_policy_url else "" %}
+    {% set tos_link_end = '</a>'|safe if tos_url else "" %}
+    {% set pp_link_end = '</a>'|safe if privacy_policy_url else "" %}
+
+    {% if (tos or tos_url) and (privacy_policy or privacy_policy_url) %}
+        {% trans tos_link_start=tos_link_start, pp_link_start=pp_link_start, tos_link_end=tos_link_end, pp_link_end=pp_link_end %}
+        I have read, understood and accept the {{tos_link_start}}terms of service{{tos_link_end}} and
+        and {{pp_link_start}}privacy policy{{pp_link_end}}.
```
tomasr8, line 38:
```suggestion
            I have read, understood and accept the {{tos_link_start}}terms of service{{tos_link_end}} and
            and {{pp_link_start}}privacy policy{{pp_link_end}}.
```

indico/modules/legal/templates/agreement.html
```diff
@@ -0,0 +1,54 @@
...
+        {% endtrans %}
+    {%- elif tos or tos_url %}
+        {% trans tos_link_start=tos_link_start, tos_link_end=tos_link_end %}
+        I have read, understood and accept the {{tos_link_start}}terms of service{{tos_link_end}}.
+        {% endtrans %}
+    {%- elif privacy_policy or privacy_policy_url %}
+        {% trans pp_link_start=pp_link_start, pp_link_end=pp_link_end %}
+        I have read, understood and accept the {{pp_link_start}}privacy policy{{pp_link_end}}.
```
tomasr8, line 46: We usually put a space inside the double curly brackets ;)
```suggestion
        I have read, understood and accept the {{ pp_link_start }}privacy policy{{ pp_link_end }}.
```

indico/modules/legal/__init__.py
kewisch: No, this one is visible for admins only:
* If enabled, normal users will be force redirected to a page to accept the terms before they continue. They never see a global-announcement banner.
* Admins can continue to browse normally, but will receive a global-announcement banner at the top telling them they should accept the terms.

kewisch: I've updated the checks so the banner doesn't appear when accepting terms is not forced, to align with the normal user experience.

ThiefMaster:
image

With a newly registered account and these settings I'm getting stuck on the "accept terms to continue using indico" page.

indico/modules/legal/controllers.py
```diff
@@ -38,3 +40,31 @@ def _process(self):
...
+
+
+class RHAcceptTerms(RH):
+    def _process(self):
+        terms_date = legal_settings.get('terms_effective_date')
+        user_accepted = (
+            session.user.accepted_tos_dt
+            and session.user.accepted_tos_dt >= terms_date
+        )
+        if not request.args.get('preview', False) and (
```
ThiefMaster, line 52: what's this `preview` arg? it doesn't seem to be used anywhere

indico/modules/legal/controllers.py
```diff
@@ -38,3 +40,31 @@ def _process(self):
         if url:
             return redirect(url)
         return WPDisplayPrivacyPolicy.render_template('privacy.html', content=legal_settings.get('privacy_policy'))
+
+
+class RHAcceptTerms(RH):
+    def _process(self):
+        terms_date = legal_settings.get('terms_effective_date')
```
ThiefMaster, line 47: This page breaks when `terms_data` is None

indico/modules/auth/client/js/signup.jsx
```diff
@@ -223,6 +283,34 @@ function Signup({
               />
             </Fieldset>
           )}
+          {termsRequireAccept && !!tos && !tosUrl && (
+            <>
+              <h3>{Translate.string('Terms of service')}</h3>
```
ThiefMaster, line 288: This h3 doesn't look great... maybe a fieldset for the whole terms/privacy section would be better?

kewisch:
> I think when signing up we need to set the terms accepted data not to the current date, but rather the 'terms effective date'. Same when clicking "accept" on the terms page.

The terms accepted date should be the point in time they accepted, this way we have that for legal purposes. I think we rather need to not prompt for terms of the effective date is in the future, which I believe should fix it as well. Thanks for finding this!

indico/modules/legal/controllers.py
kewisch: Ah I believe used this to preview the terms accept without having to fiddle around with dates, so more of a developer tool. Happy to remove this.

indico/modules/auth/client/js/signup.jsx
ThiefMaster: Also I just noticed that this should use `<Translate>...</Translate>` since `Translate.string()` should only be used in places where you need a plain string

indico/modules/legal/forms.py
```diff
@@ -5,14 +5,20 @@
...
 
 from indico.util.i18n import _
 from indico.web.forms.base import IndicoForm
+from indico.web.forms.fields import IndicoDateField
 from indico.web.forms.widgets import TinyMCEWidget
 
 
+class AgreementForm(IndicoForm):
+    # Label needs to be set in the html template
+    accept_terms = BooleanField(_('Accept terms'), [InputRequired()])
```
kewisch, line 19: The change you made to add the label here might make it a bit ambiguous. The checkbox says accept terms, but the descriptions says to accept the terms and privacy policy. I had left the label empty here so that the description is the label, and the checkbox specifically says accept the terms and privacy policy.

ThiefMaster: Empty label does not look great though, and more importantly you lose the ability to click something that toggles the checkbox (which also makes this an a11y issue).

I wouldn't mind setting the label dynamically, but modifying the label on the fly is a bit ugly. The cleanest way is probably to simply create the form itself dynamically (ie create the class on the fly - there are a few examples in the codebase where we use `form_cls = type('_SomeForm', (IndicoForm,), {...})`) so you can set the label dynamically as well.

Anyway, the idea was that "terms" is generic and thus covers any kind of terms you have to accept, and then the description shows the verbose details on what you're accepting.

An alternative could be adding two checkboxes (both required ofc) if there are ToS and Privacy Policy to accept. That avoids all the not-so-pretty dynamic generation of the description as well... Code-wise that's also be super easy: Have both checkboxes in the form class, and use `del self.whatever` to remove the ones not needed in the form's `__init__` after the super call.

kewisch: I think I'd like to go with setting the label dynamically and have done so with a dynamic form. I've run in to an issue where the text wraps into the next line. While I could certainly fix it for English by tweaking the text, it should probably be fixed in HTML/CSS instead. This could affect very many forms though, so I haven't taken the plunge here. If you feel more comfortable changing the styles here I'd much appreciate.

image

ThiefMaster (approved):
nice pr :)

PR #5928, opened 2023-09-13
Config option for default future months on category page

### LINKED ISSUE #2984: Improve logic to determine which events to show on the category page
> on June 27 an event of July 5th is not show automatically but the person need to click on show future events in order to see what is going on ...next week .. that is also next month …
isn’t it possible to change the time gap and show all events that are organised at least 2-3 weeks in the future ?

The current logic there is:

> Current events, which are always shown by default are events of this month and of the previous month. 
> If there are no events in this range, it will include the last and next month containing events.

We probably do not want to show partial months (this would likely require extra changes to the JS/template), but showing the full next month may also be spammy. One option could be showing the next month if there's less than one week left in the current one.

kewisch:
Fixes #2984.

### DIFF
--- a/indico/migrations/versions/20230913_2116_155cfc134f0c_future_events_threshold_setting.py
+++ b/indico/migrations/versions/20230913_2116_155cfc134f0c_future_events_threshold_setting.py
@@ -1,0 +1,25 @@
+"""Setting for future events threshold
+
+Revision ID: 155cfc134f0c
+Revises: 0c44046dc1be
+Create Date: 2023-09-13 21:16:13.839731
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '155cfc134f0c'
+down_revision = '0c44046dc1be'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('categories', sa.Column('show_future_months', sa.Integer(), nullable=False, server_default='0'), schema='categories')
+    op.alter_column('categories', 'show_future_months', server_default=None, schema='categories')
+
+
+def downgrade():
+    op.drop_column('categories', 'show_future_months', schema='categories')
--- a/indico/modules/categories/forms.py
+++ b/indico/modules/categories/forms.py
@@ -32,8 +32,8 @@ from indico.web.forms.widgets import HiddenCheckbox, SwitchWidget
 
 class CategorySettingsForm(IndicoForm):
     BASIC_FIELDS = ('title', 'description', 'timezone', 'lecture_theme', 'meeting_theme', 'visibility',
-                    'suggestions_disabled', 'is_flat_view_enabled', 'event_creation_notification_emails',
-                    'notify_managers')
+                    'suggestions_disabled', 'is_flat_view_enabled', 'show_future_months',
+                    'event_creation_notification_emails', 'notify_managers')
     EVENT_HEADER_FIELDS = ('event_message_mode', 'event_message')
 
     title = StringField(_('Title'), [DataRequired()])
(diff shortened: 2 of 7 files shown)


indico/modules/categories/forms.py
```diff
@@ -52,6 +52,10 @@ class CategorySettingsForm(IndicoForm):
                                         description=_('Allow users to view all the events descending from this '
                                                       'category in one single page. This is not recommended on large '
                                                       'categories with thousands of events.'))
+    show_future_months = IntegerField(_('Furture months threshold'), [NumberRange(min=0)],
```
ThiefMaster, line 55:
```suggestion
    show_future_months = IntegerField(_('Future months threshold'), [NumberRange(min=0)],
```

indico/modules/categories/models/categories.py
```diff
@@ -175,6 +175,11 @@ def __table_args__(cls):
         nullable=False,
         default=False
     )
+    show_future_months = db.Column(
+        db.Integer,
+        nullable=False,
+        default=1
```
ThiefMaster, line 181:
```suggestion
        default=0
```

matching the current behavior and the default set in the alembic revision

indico/migrations/versions/20230913_2116_155cfc134f0c_future_events_threshold_setting.py
```diff
@@ -0,0 +1,25 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '155cfc134f0c'
+down_revision = '0c44046dc1be'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('categories', sa.Column('show_future_months', sa.Integer(), nullable=False, server_default=str(0)), schema='categories')
```
ThiefMaster, line 20:
```suggestion
    op.add_column('categories', sa.Column('show_future_months', sa.Integer(), nullable=False, server_default='0'),
                  schema='categories')
```

indico/modules/categories/models/categories.py
kewisch: Ah yes, I was going back and forth on this and missed changing it back. Thanks!

indico/modules/categories/operations.py
```diff
@@ -83,6 +83,7 @@ def _log_category_update(category, changes, extra_log_fields):
         'timezone': {'title': 'Timezone', 'type': 'string'},
         'suggestions_disabled': 'Disable suggestions',
         'is_flat_view_enabled': 'Allow flat view',
+        'show_future_months': 'Show future months',
```
ThiefMaster, line 86:
```suggestion
        'show_future_months': 'Future months threshold',
```

indico/modules/categories/forms.py
```diff
@@ -52,6 +52,10 @@ class CategorySettingsForm(IndicoForm):
                                         description=_('Allow users to view all the events descending from this '
                                                       'category in one single page. This is not recommended on large '
                                                       'categories with thousands of events.'))
+    show_future_months = IntegerField(_('Future months threshold'), [NumberRange(min=0)],
+                                      description=_('Future months to show by default on the category page. If no '
+                                                    'events are found, the next 1 month after with events will be '
+                                                    'shown.'))
```
ThiefMaster, line 58: Not sure if this is really clear. I mean I understand what it does after looking at the code, but as a user I don't think I'd have an idea what this setting does and if/why I should change it...

kewisch: Yeah, the 1 month after also makes it a little less straightforward. How about turning it around?

> Events past the threshold will be hidden by default to avoid clutter, the user can click to expand them. If no events are found within this threshold, it is extended to show the first month with events.

ThiefMaster: sounds good!

ThiefMaster (approved):

PR #5929, opened 2023-09-15
Add block upcoming events title.

vtran99:
Request add block to macro render_upcoming_events as we need to customise rendering of events displayed.

### DIFF
--- a/indico/modules/categories/templates/display/sidebar.html
+++ b/indico/modules/categories/templates/display/sidebar.html
@@ -66,9 +66,11 @@
         <ul class="main-page-list">
             {% for event in upcoming_events %}
                 <li>
+                    {% block upcoming_title %}
                     <a href="{{ event.url }}" class="title" title="{{ event.title|striptags }}" >
                         {{ event.title|striptags }}
                     </a>
+                    {% endblock %}
                     <span class="timing">
                         {% if happening_now(event) %}
                             {% trans time=event.end_dt|format_pretty_date(tzinfo=tzinfo) -%}
--- a/indico/modules/categories/templates/display/sidebar.html
+++ b/indico/modules/categories/templates/display/sidebar.html
@@ -67,9 +67,9 @@
             {% for event in upcoming_events %}
                 <li>
                     {% block upcoming_title scoped %}
-                    <a href="{{ event.url }}" class="title" title="{{ event.title|striptags }}" >
-                        {{ event.title|striptags }}
-                    </a>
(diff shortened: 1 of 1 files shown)


indico/modules/categories/templates/display/sidebar.html
```diff
@@ -66,9 +66,11 @@ <h2 class="icon-alarm">{% trans %}Upcoming events{% endtrans %}</h2>
         <ul class="main-page-list">
             {% for event in upcoming_events %}
                 <li>
+                    {% block upcoming_title %}
```
ThiefMaster, line 69: Have you tested this, and also accessing the `event` inside the custom block? Pretty sure you need to use the `scoped` modifier on the block definition for this to work!

```suggestion
                    {% block upcoming_title scoped %}
```

Also make sure to test with multiple events, I'm not completely sure how blocks interact with loops. Probably they're fine but best to test this.

indico/modules/categories/templates/display/sidebar.html
```diff
@@ -66,9 +66,11 @@ <h2 class="icon-alarm">{% trans %}Upcoming events{% endtrans %}</h2>
...
                     <a href="{{ event.url }}" class="title" title="{{ event.title|striptags }}" >
                         {{ event.title|striptags }}
                     </a>
```
ThiefMaster, line 72:
```suggestion
                        <a href="{{ event.url }}" class="title" title="{{ event.title|striptags }}" >
                            {{ event.title|striptags }}
                        </a>
```

indico/modules/categories/templates/display/sidebar.html
vtran99: "scoped" added.
Yes, I have tested with the rendering of a list go events and it works.
For info, we added a prefix to the event title when rendering the list.

ThiefMaster: I'm very surprised it worked for you without `scoped`. When I remove the `scoped` modifier, simply accessing the indico home page (with upcoming events), it fails:

> jinja2.exceptions.UndefinedError: 'event' is undefined

ThiefMaster (approved):

indico/modules/categories/templates/display/sidebar.html
vtran99: Indeed, it works for me because in the overridden template I used "scoped" (no scoped in original block, scoped used in overridden block).
Thanks for the approval.

PR #5930, opened 2023-09-18
Fix percentage calculation for survey results

ThiefMaster:
In case of multi-choice questions, the percentace should be based on the total number of answers, not the total number of choices selected.


ThiefMaster:
(this was meant to go in 3.2.x of course)

PR #5931, opened 2023-09-19
Reimplement the category navigator using React

### LINKED ISSUE #5917: [A11Y] Category navigator accessibility
**Describe the bug**

The category navigator has multiple accessibility issues:

- Not announced as a dialog to the screen reader
- Keyboard navigation is broken
- Navigation (action) button is not visible by default

**To Reproduce**

Scenario A:

1. Go to main page
2. Activate "Navigate"
3. Observe that screen reader does not announce the presence of a dialog

Scenario B:

1. Go to main page
2. Activate "Navigate"
3. Try tabbing into one of the categories
4. Observe that navigation skips over the category list and goes to "Close" button

Scenario C:

1. Go to main page
2. Activate "Navigate"
3. Observe that "Navigate to >" is not shown for any categories

**Expected behavior**

Scenario A: the dialog is correctly announced as a dialog
Scenario B: it should be possible to use the tab navigation to access categories
Scenario C: the "Navigate to >" (or an alternative cue) should be present without hover

**Additional context**

- https://talk.getindico.io/t/keyboard-improve-keyboard-interaction-of-buttons-in-the-category-navigation/3209
- https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html

foxbunny:
- Port the category navigator to React
- Fix keyboard navigation and other a11y issues
- Make the category navigator responsive
- Add tooltip and toggletip custom elements

Fixes #5917

### DIFF
--- a/indico/modules/categories/client/js/components/NavigateTo.jsx
+++ b/indico/modules/categories/client/js/components/NavigateTo.jsx
@@ -11,7 +11,6 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 import {Param, Translate} from 'indico/react/i18n';
-import 'indico/custom_elements/ind_with_tooltip';
 
 import './NavigateTo.module.scss';
 
--- a/indico/modules/categories/client/js/components/NavigateTo.jsx
+++ b/indico/modules/categories/client/js/components/NavigateTo.jsx
@@ -13,12 +13,12 @@ import React from 'react';
 import {Param, Translate} from 'indico/react/i18n';
 import 'indico/custom_elements/ind_with_tooltip';
 
-import css from './NavigateTo.module.scss';
+import './NavigateTo.module.scss';
 
 export default function NavigateTo({category}) {
   return (
     <ind-with-tooltip>
-      <a href={categoryURL({category_id: category.id})} className={css.NavigateTo}>
+      <a href={categoryURL({category_id: category.id})} styleName="navigate-to">
         Navigate
         <Translate as="span" data-tip-content>
           Navigate to <Param name="title" value={category.title} />
--- a/indico/modules/categories/client/js/components/NavigateTo.jsx
+++ b/indico/modules/categories/client/js/components/NavigateTo.jsx
@@ -19,7 +19,9 @@ export default function NavigateTo({category}) {
   return (
     <ind-with-tooltip>
       <a href={categoryURL({category_id: category.id})} styleName="navigate-to">
-        Navigate
+        <Translate as="span" aria-hidden="true">
+          Navigate
+        </Translate>
         <Translate as="span" data-tip-content>
           Navigate to <Param name="title" value={category.title} />
         </Translate>
@@ -29,5 +31,5 @@ export default function NavigateTo({category}) {
 }
 
 NavigateTo.propTypes = {
-  category: PropTypes.object,
+  category: PropTypes.object.isRequired,
 };
--- a/indico/modules/categories/client/js/components/NavigateTo.jsx
+++ b/indico/modules/categories/client/js/components/NavigateTo.jsx
@@ -1,0 +1,33 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2024 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import categoryURL from 'indico-url:categories.display';
+
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import {Param, Translate} from 'indico/react/i18n';
+import 'indico/custom_elements/ind_with_tooltip';
+
+import css from './NavigateTo.module.scss';
+
+export default function NavigateTo({category}) {
+  return (
+    <ind-with-tooltip>
+      <a href={categoryURL({category_id: category.id})} className={css.NavigateTo}>
+        Navigate
+        <Translate as="span" data-tip-content>
+          Navigate to <Param name="title" value={category.title} />
+        </Translate>
+      </a>
+    </ind-with-tooltip>
+  );
+}
+
+NavigateTo.propTypes = {
+  category: PropTypes.object,
+};
--- a/indico/modules/categories/client/js/ind-category-navigator.js
+++ b/indico/modules/categories/client/js/ind-category-navigator.js
@@ -1,0 +1,33 @@
+// This file is part of Indico.
+// Copyright (C) 2002 - 2024 CERN
+//
+// Indico is free software; you can redistribute it and/or
+// modify it under the terms of the MIT License; see the
+// LICENSE file for more details.
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import * as categoryNavigator from 'indico/react/components/category_navigator';
+
+import NavigateTo from './components/NavigateTo';
+
+customElements.define(
+  'ind-category-navigator',
+  class extends HTMLElement {
+    connectedCallback() {
+      const options = {
+        trigger: document.getElementById(this.getAttribute('triggeredby')),
+        categoryId: this.getAttribute('category-id'),
+      };
+      ReactDOM.render(
+        <categoryNavigator.DialogViewModel
+          view={categoryNavigator.DialogView}
+          actionView={NavigateTo}
+          {...options}
+        />,
+        this
+      );
+    }
+  }
+);
--- a/indico/modules/categories/client/js/ind-category-navigator.js
+++ b/indico/modules/categories/client/js/ind-category-navigator.js
@@ -16,18 +16,31 @@ customElements.define(
   'ind-category-navigator',
   class extends HTMLElement {
     connectedCallback() {
-      const options = {
-        trigger: document.getElementById(this.getAttribute('triggeredby')),
-        categoryId: this.getAttribute('category-id'),
-      };
       ReactDOM.render(
         <categoryNavigator.DialogViewModel
+          categoryId={this.categoryId}
           view={categoryNavigator.DialogView}
-          actionView={NavigateTo}
-          {...options}
         />,
         this
       );
+      this.trigger.addEventListener('click', this.navigate);
     }
+
+    disconnectedCallback() {
+      this.trigger.removeEventListener('click', this.navigate);
+      ReactDOM.unmountComponentAtNode(this);
+    }
+
+    get categoryId() {
+      return this.getAttribute('category-id');
+    }
+
+    get trigger() {
+      return document.getElementById(this.getAttribute('triggeredby'));
+    }
+
+    navigate = () => {
+      window.dispatchEvent(new CustomEvent('open-category-navigator', {detail: NavigateTo}));
+    };
   }
 );
--- a/indico/modules/categories/templates/display/base.html
+++ b/indico/modules/categories/templates/display/base.html
@@ -107,19 +107,11 @@
         </div>
     </div>
 </div>
-<script>
-    $('#navigate-button').on('click', function(evt) {
-        $('<div>').categorynavigator({
-            openInDialog: true,
-            category: {{ category.id }},
-            actionButtonText: $T.gettext('Navigate to'),
-            dialogTitle: $T.gettext('Navigate to a category'),
-            onAction: function(category) {
-                window.location.href = build_url(Indico.Urls.Categories.display, {category_id: category.id});
-            }
-        });
-    });
-</script>
+
+<ind-category-navigator triggeredby="navigate-button"
+                        category-id="{{ category.id }}">
+</ind-category-navigator>
+
 <script type="application/ld+json">
 {{ json_ld|tojson }}
 </script>
--- a/indico/web/client/js/custom_elements/ind_with_toggletip.README.md
+++ b/indico/web/client/js/custom_elements/ind_with_toggletip.README.md
@@ -1,0 +1,36 @@
+This element is used to add toggletips to a dedicated toggletip button.
+
+Toggletips are buttons with just one purpose: open a popup message similar to a
+tooltip. They are usually styles as info buttons (circle with lower-case i or
+questionmark). Since tooltips can only supply non-critical messages on
+interactive elements, toggletips are used in cases where extra information must
+be supplied but no interactive elements are involved.
+
+Toggletips are created by placing the target button and tip content inside a
+`<ind-with-toggletip>` element. The tip content is marked using the
+`data-tip-content` attribute, and it also marked as a live region by giving
+it the `aria-live="polite"` attribute.
+
+By setting the `aria-live` attribute on the element, we cause the screen
+readers and similar accessibility technology to read the tooltip when the
+button is activated.
+
+Example:
+
+```html
+<ind-with-toggletip>
+  <button>More information</button>
+  <div data-tip-content aria-live="polite">
+    The toggle tip content...
+  </div>
+</ind-with-toggletip>
+```
+
+Note: When working with React app, you may sometimes observe that the
+toggle tip content does not change. The DOM nodes in the tree may change
+but the old text is still displayed. Alternatively, you may get a
+`DOMException`. This happens because the custom element manupulates the
+contents of the `[data-tip-content]` element. To work around this, ensure
+that React cleanly replaces the entire custom element with a new one by
+giving the `<ind-with-toggletip>` element a `key` prop. The key should be
+such that it changes whenever the tip contents change.
--- a/indico/web/client/js/custom_elements/ind_with_toggletip.README.md
+++ b/indico/web/client/js/custom_elements/ind_with_toggletip.README.md
@@ -29,7 +29,7 @@ Example:
 Note: When working with React app, you may sometimes observe that the
 toggle tip content does not change. The DOM nodes in the tree may change
 but the old text is still displayed. Alternatively, you may get a
-`DOMException`. This happens because the custom element manupulates the
+`DOMException`. This happens because the custom element manipulates the
 contents of the `[data-tip-content]` element. To work around this, ensure
 that React cleanly replaces the entire custom element with a new one by
 giving the `<ind-with-toggletip>` element a `key` prop. The key should be
--- a/indico/web/client/js/custom_elements/tips.scss
+++ b/indico/web/client/js/custom_elements/tips.scss
@@ -9,10 +9,10 @@
 
 ind-with-tooltip,
 ind-with-toggletip {
+  // XXX: The following variables are set using JavaScript:
   --tooltip-surface-color: #333;
   --tooltip-text-color: #fff;
 
-  // XXX: The following variables are set using JavaScript:
   --top: 0;
   --left: 0;
   --right: auto;
--- a/indico/web/client/js/custom_elements/tips.scss
+++ b/indico/web/client/js/custom_elements/tips.scss
@@ -9,10 +9,10 @@
 
 ind-with-tooltip,
 ind-with-toggletip {
-  // XXX: The following variables are set using JavaScript:
   --tooltip-surface-color: #333;
   --tooltip-text-color: #fff;
 
+  // XXX: The following variables are set using JavaScript:
   --top: 0;
   --left: 0;
   --right: auto;
--- a/indico/web/client/js/custom_elements/tips.scss
+++ b/indico/web/client/js/custom_elements/tips.scss
@@ -20,6 +20,8 @@ ind-with-toggletip {
   --ref-center: 0;
   --arrow-borders: transparent;
 
+  display: inline-block;
+
   [data-tip-content] {
     display: flex;
     flex-direction: column;
--- a/indico/web/client/js/react/components/category_navigator/style.scss
+++ b/indico/web/client/js/react/components/category_navigator/style.scss
@@ -119,7 +119,7 @@
 #category-nav-dialog {
   @extend %dialog-modal;
   --dialog-max-height: 80%;
-  --dialog-max-width: 80%;
+  --dialog-max-width: min(80%, 60em);
 
   // FIXME: Move to reset
   p,
--- a/indico/web/client/js/react/components/category_navigator/style.scss
+++ b/indico/web/client/js/react/components/category_navigator/style.scss
@@ -13,14 +13,14 @@
   @extend %flex-split;
 
   min-height: 3.2em;
-  padding: 0 0.5em;
+  padding: var(--content-gap-within-inline) var(--content-gap-normal);
 
   background: var(--background-base);
   border-left: 3px solid var(--container-separator-color);
   border-bottom: 1px solid var(--container-separator-color);
 
   > * {
-    padding: 1em;
+    padding: 0.5em;
   }
 
   > :first-child::before {
@@ -116,13 +116,6 @@
   border-left-color: var(--accent-color);
 }
 
-%category-title-with-breadcrumb {
-  @extend %flex-column;
-  --flex-gap: var(--content-gap-within-inline);
-  align-items: flex-start;
-  padding: 0.5em 1em;
-}
-
 #category-nav-dialog {
   @extend %dialog-modal;
 
@@ -140,7 +133,7 @@
     @extend %flex-row;
     --flex-gap: var(--content-gap-within-inline);
 
-    font-size: 87.5%;
+    font-size: var(--text-size-rel-s);
 
     button {
       @extend %link-button;
@@ -163,10 +156,6 @@
     header {
       @extend %category-list-item;
       @extend %category-list-item-highlighted;
-    }
-
-    ul {
-      padding-left: 1em;
     }
 
     li {
@@ -197,7 +186,6 @@
       @extend %flex-column;
       --flex-gap: var(--content-gap-within-inline);
       align-items: flex-start;
-      padding: 0.5em 0;
       flex: 1000;
     }
   }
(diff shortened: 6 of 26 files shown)


foxbunny:
As a side note, this is not integrated into any other parts of the Indico where the old category navigator is used. It just replaces the usage on the category pages for now.

indico/web/client/js/react/components/category_navigator/viewmodel.jsx
```diff
@@ -0,0 +1,148 @@
...
+  updateModelFromSearchError,
+  prepareForSearch,
+  clearSearch,
+  finishLoading,
+  resetCategoryList,
+} from './model';
+
+const loadingSpinnerDelay = 1500;
+
+export class DialogViewModel extends React.Component {
```
ThiefMaster, line 29: Please do not add new class-based components - we solely use functions (+ react hooks) in any new code)

ThiefMaster: In which way would it add complexity? FWIW I did not have a closer look at the rest of the PR yet, I just noticed this when skimming over it.

foxbunny: > In which way would it add complexity? 

I've fixed the PR already, but let me add the $0.02. I've done some research last night, and this whole point is kinda moot cause React team are apparently deprecating class components anyway. We can have a chat about this elsewhere if you would like to know what my stance is. :)

ThiefMaster:
Currently this does not work well with long category names

image

foxbunny:
> Currently this does not work well with long category names

Looks like it's not the category name but event-/subcategory counts.

indico/web/client/js/react/components/category_navigator/view.jsx
```diff
@@ -0,0 +1,350 @@
...
+};
+
+function BreadcrumbView({path, onChangeCategory}) {
+  if (!path?.length) {
+    return null;
+  }
+
+  return (
+    <div className="breadcrumbs">
+      <span>{Translate.string('in')}</span>
```
ThiefMaster, line 93:
```suggestion
      <Translate as="span">in</Translate>
```

indico/web/client/js/react/components/category_navigator/view.jsx
```diff
@@ -0,0 +1,350 @@
...
+      {path.map(category => (
+        <button
+          type="button"
+          key={category.id}
+          onClick={() => onChangeCategory(category.id)}
+          aria-label={Translate.string('View the {title} category', {title: category.title})}
```
ThiefMaster, line 99:
```suggestion
          aria-label={Translate.string('View the "{title}" category', {title: category.title})}
```

Depending on the category title not quoting it may look strange

indico/modules/categories/client/js/ind-category-navigator.js
```diff
@@ -0,0 +1,33 @@
...
+import ReactDOM from 'react-dom';
+
+import * as categoryNavigator from 'indico/react/components/category_navigator';
+
+import NavigateTo from './components/NavigateTo';
+
+customElements.define(
+  'ind-category-navigator',
+  class extends HTMLElement {
+    connectedCallback() {
```
ThiefMaster, line 18: Should `disconnectedCallback` be handled as well (and unmount the react component)?

indico/modules/categories/templates/display/base.html
```diff
@@ -107,19 +107,11 @@ <h1 class="category-title">
...
-            actionButtonText: $T.gettext('Navigate to'),
-            dialogTitle: $T.gettext('Navigate to a category'),
-            onAction: function(category) {
-                window.location.href = build_url(Indico.Urls.Categories.display, {category_id: category.id});
-            }
-        });
-    });
-</script>
+
+<ind-category-navigator triggeredby="navigate-button"
```
ThiefMaster, line 111: This is just one place where the category picker dialog is used.

We currently use the same dialog when creating events or moving events/categories. Since we don't want to keep two implementations of the same thing around, it would be good to adapt those parts as well in a future PR after this one is merged.

indico/web/client/js/react/components/category_navigator/view.jsx
```diff
@@ -0,0 +1,350 @@
...
+        <span aria-hidden="true">{subcategories}</span>
+      </span>
+      <ind-with-toggletip>
+        <button type="button">
+          <span>
+            <Translate>Show category stats</Translate>
+          </span>
+        </button>
+        <span data-tip-content aria-live="polite">
+          {categoryTipText}
```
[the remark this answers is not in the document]
foxbunny: I'll look into it. I don't remember what this was related to off the top of my head, so I'll provide a generic comment: anything that doesn't have an alternative way of activation should generally be always visible (e.g., you don't have hover or focus on touch devices).

indico/web/client/js/react/components/category_navigator/view.jsx
```diff
@@ -0,0 +1,350 @@
...
+        count: events,
+      }),
+      subcategories: PluralTranslate.string(
+        '{count} subcategory',
+        '{count} subcategories',
+        subcategories,
+        {count: subcategories}
+      ),
+    }
+  );
```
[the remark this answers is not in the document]
foxbunny: Hmm, right. Tricky one. I'll give it some more thought tomorrow.

indico/web/client/js/custom_elements/tips.scss
```diff
@@ -9,10 +9,10 @@
 
 ind-with-tooltip,
 ind-with-toggletip {
+  // XXX: The following variables are set using JavaScript:
```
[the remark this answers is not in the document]
foxbunny: Yeah, it's a redundant comment.

indico/web/client/js/react/components/category_navigator/view.jsx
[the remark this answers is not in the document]
ThiefMaster: At least the numbers are always visible - so right now the trigger for this tooltip is kind of adding redundant information:

image

I also just noticed a bug: This tooltip for the topmost entry always shows the number for the initial category, even when you started navigating into another one.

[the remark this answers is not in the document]
foxbunny: > At least the numbers are always visible - so right now the trigger for this tooltip is kind of adding redundant information

This redundancy is intentional. You cannot add tooltips to non-interactive elements, but those icons don't readily convey their meaning. Therefore, there was a need to add a toggletip.

> I also just noticed a bug

Thanks. I'll fix it.

indico/web/client/js/react/components/category_navigator/viewmodel.jsx
```diff
@@ -0,0 +1,131 @@
...
+import React, {useEffect, useRef, useState} from 'react';
+import ReactDOM from 'react-dom';
+
+import {getCategoryData, getCategorySearchResults} from './api';
+import * as model from './model';
+
+const loadingSpinnerDelay = 1500;
+
+export function DialogViewModel({trigger, categoryId, view: View, actionView, dialogTitle}) {
+  const [state, setState] = useState(model.init());
```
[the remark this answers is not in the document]
foxbunny: Not sure, but I've been using this pattern since before useReducer was invented. 🤔

indico/modules/categories/client/js/ind-category-navigator.js
foxbunny: Do you remove this widget from the page at any point?

indico/modules/categories/client/js/components/NavigateTo.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+import {Param, Translate} from 'indico/react/i18n';
+import 'indico/custom_elements/ind_with_tooltip';
+
+import css from './NavigateTo.module.scss';
+
+export default function NavigateTo({category}) {
+  return (
+    <ind-with-tooltip>
+      <a href={categoryURL({category_id: category.id})} className={css.NavigateTo}>
```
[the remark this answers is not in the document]
foxbunny: Changed, since I changed it to styleName anyway.

indico/modules/categories/templates/display/base.html
foxbunny: I think it would be good to have one of you do it, with me as an assistant. Over time there's probably going to be more of this type of code, so it's good to become familiar. You know, while I'm still around. 😂

indico/modules/categories/client/js/components/NavigateTo.jsx
```diff
@@ -19,7 +19,7 @@ export default function NavigateTo({category}) {
   return (
     <ind-with-tooltip>
       <a href={categoryURL({category_id: category.id})} styleName="navigate-to">
-        Navigate
+        <span aria-hidden="true">Navigate</span>
```
ThiefMaster, line 22: this also needs to be wrapped in `<Translate>`

indico/web/client/js/react/components/category_navigator/viewmodel.jsx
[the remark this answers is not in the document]
ThiefMaster: I'm fine with leaving it (even though it would probably not be much work to change), was just a bit surprised to see *you* kind of re-inventing the wheel! ;)

indico/modules/categories/client/js/ind-category-navigator.js
ThiefMaster: No, was just wondering.

FWIW when using it in other places, in particular for event creation (which is still a wtform and not using react directly but rather just to render individual widgets that happen to be implemented in react), we'd probably want to clean up (after closing the dialog the widget is no longer used, even though i guess there's no harm in leaving it mounted since the element it's mounted on no longer exists).

indico/modules/categories/templates/display/base.html
ThiefMaster: The problem is that we're quite busy atm, so I don't think "adding big functionality to this component" is a thing one of us really has the time for right now...

I'd be happy to do the integration in those places (create event, move event, move category, badge template cloning), as long as the features needed for those particular operations are already present in the new category picker:

```js
{
  actionOn: {
    // UNUSED (i think)
    categoriesWithoutEventProposalRights: {
      message: $T.gettext('Not possible for categories where you cannot propose events'),
    },
    // REQUIRED - when moving events
    categoriesWithoutEventProposalOrCreationRights: {
      message: $T.gettext('Not possible for categories where you cannot propose/create events'),
    },
    // REQUIRED - when creating events
    categoriesWithoutEventCreationRights: {
      message: $T.gettext('Not possible for categories where you cannot create events'),
    },
    // REQUIRED - when cloning badge templates or moving categories [the latter was not being checked, i just fixed that in master, commit cac224f8fa] 
    categoriesWithoutCategoryManagementRights: {
      message: $T.gettext('Not possible for categories where you are not a manager'),
    },
    // REQUIRED - when moving categories
    categoriesDescendingFrom: {
      ids: [],
      // The closest parent category title will be used if more than one parent is present in ids
      message: $T.gettext('Not possible for categories descending from category "{0}"'),
    },
    // REQUIRED - when moving categories
    categories: {
      disabled: false,
      message: $T.gettext('Not possible for this category'),
      ids: [],
      // Expects an Array of {ids: [...], message: '...'} for more specific messages
      groups: [],
    },
  },
}
```

indico/web/client/js/custom_elements/ind_with_toggletip.README.md
```diff
@@ -0,0 +1,36 @@
...
+  <div data-tip-content aria-live="polite">
+    The toggle tip content...
+  </div>
+</ind-with-toggletip>
+```
+
+Note: When working with React app, you may sometimes observe that the
+toggle tip content does not change. The DOM nodes in the tree may change
+but the old text is still displayed. Alternatively, you may get a
+`DOMException`. This happens because the custom element manupulates the
```
ThiefMaster, line 32:
```suggestion
`DOMException`. This happens because the custom element manipulates the
```

indico/web/client/js/react/components/category_navigator/view.jsx
```diff
@@ -136,40 +136,41 @@ CategoryItemLabelView.propTypes = {
...
+              <Param value={category.title} name="title" />.
+            </Singular>
+            <Plural>
+              There are <Param value={category.deepEventCount} name="count" /> events in{' '}
+              <Param value={category.title} name="title" />.
+            </Plural>
+          </PluralTranslate>{' '}
+          <PluralTranslate count={category.deepCategoryCount}>
+            <Singular>
+              Category contains <Param value={category.deepCategoryCount} name="count" />{' '}
```
ThiefMaster, line 166: > *This* category contains...

(same below)

indico/web/client/js/react/components/category_navigator/style.scss
```diff
@@ -0,0 +1,262 @@
...
+
+  min-height: 3.2em;
+  padding: var(--content-gap-within-inline) var(--content-gap-normal);
+
+  background: var(--background-base);
+  border-left: 3px solid var(--container-separator-color);
+  border-bottom: 1px solid var(--container-separator-color);
+
+  > * {
+    padding: 0.5em;
```
ThiefMaster, line 23: Still feels a big like wasting space to me on paddings... Is there any a11y problem w/ going even lower here?

ThiefMaster:
There are still a few comments unanswered - maybe github collapsed those and you missed them?

- https://github.com/indico/indico/pull/5931#discussion_r1504434819
- https://github.com/indico/indico/pull/5931#discussion_r1504449462
- https://github.com/indico/indico/pull/5931#discussion_r1504462199
- https://github.com/indico/indico/pull/5931#discussion_r1504464185
- https://github.com/indico/indico/pull/5931#discussion_r1504464794
- https://github.com/indico/indico/pull/5931#discussion_r1504475731

indico/web/client/js/react/components/category_navigator/viewmodel.jsx
```diff
@@ -0,0 +1,131 @@
...
+  const dialogRef = useRef();
+  const searchFieldRef = useRef();
+
+  useEffect(() => {
+    trigger.addEventListener('click', openDialog);
+    // eslint-disable-next-line
```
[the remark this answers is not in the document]
foxbunny: > FWIW if those functions do not change during re-renders, it would be better to properly put those in the dependency list.

If you don't change something, putting it in a dependency list just makes React perform unnecessary checks. Not sure why that would be better?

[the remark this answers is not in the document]
ThiefMaster: I think that's a question for a core react dev, but I guess there's a reason why they have their eslint rule that wants you to put anything you use in the dependency list. Unless it's just because in most cases you want stuff there, and it cannot track whether you really never change certain things...

indico/web/client/js/react/components/category_navigator/style.scss
foxbunny: It's generally not a good idea to remove vertical separation when there is ample horizontal separation between elements. But no, there's no critical a11y failure, it just makes for suboptimal readability.

indico/web/client/js/react/components/category_navigator/viewmodel.jsx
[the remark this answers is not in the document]
foxbunny: > but I guess there's a reason why they have their eslint rule that wants you to put anything you use in the dependency list

I wouldn't make that assumption myself. 😂

indico/web/client/js/react/components/category_navigator/style.scss
ThiefMaster: I think the alternative would be making the dialog bigger so you can see more at once. But I guess you agree that seeing only very few items at a time and having to scroll a lot is suboptimal?

foxbunny: Increasing the dialog size will help, sure. Additionally, we can also implement a simplified view that has less elements, and that list can be squashed tighter. That's been the usual pattern in apps I've used.

indico/web/client/js/react/components/category_navigator/view.jsx
foxbunny: Now that I'm looking at this, it'll probably be impossible to translate. I'll try to figure out a better solution.

indico/modules/categories/templates/display/base.html
foxbunny: It's not really "big functionality". All you customize is the button that triggers the action, which is supplied separately as a prop. But sure, no problem. I'll integrate it when I get around to those pages.

ThiefMaster: But your new react picker does not have the logic for all those criteria yet, does it?

As I said, I'm fine integrating it on the pages once the logic is there ;)

(I want to avoid a long period of time of having two category pickers in the codebase in parallel)

indico/web/client/js/react/components/category_navigator/view.jsx
foxbunny: It's an aria label, so it doesn't "look". :) Screen readers will not do anything special with quotes, afaik.

foxbunny: But let me double-check that just in case.

ThiefMaster: ah, missed that. nvm then :)

foxbunny: Ok, it really makes no difference. But since it makes no difference, I'll add the quote just in case there are ATs that do something with them.

indico/modules/categories/client/js/ind-category-navigator.js
foxbunny: In that case, we can add the unmount. Afaik, React does some subscriptions that won't be cleared when the node is removed.

indico/modules/categories/templates/display/base.html
foxbunny: > But your new react picker does not have the logic for all those criteria yet, does it?

What logic are we talking about? Maybe we can hop on a call during the week, so you (or someone else on your team) can give me a quick summary?

indico/web/client/js/react/components/category_navigator/viewmodel.jsx
```diff
@@ -0,0 +1,131 @@
...
+// LICENSE file for more details.
+
+import PropTypes from 'prop-types';
+import React, {useEffect, useRef, useState} from 'react';
+import ReactDOM from 'react-dom';
+
+import {getCategoryData, getCategorySearchResults} from './api';
+import * as model from './model';
+
+const loadingSpinnerDelay = 1500;
```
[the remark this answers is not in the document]
ThiefMaster: I can only speak for myself, but if there's over half a second with no activity I get the feeling that something is broken...

(the fact that browsers ditched the cursor-with-busy while pages are loading doesn't make this better)

indico/modules/categories/templates/display/base.html
ThiefMaster: When instantiating the old jquery category picker you can pass some config arguments (`actionOn: {...}`) specifying when the action button (currently that's always your `NavigateTo` component) should be disabled (right now the message why it's disabled goes into a tooltip):

image

For example when moving categories we pass this config:

```js
      actionOn: {
        categoriesWithoutCategoryManagementRights: {
          disabled: true,
        },
        categoriesDescendingFrom: {
          disabled: true,
          ids,
        },
        categories: {
          disabled: true,
          groups: [
            {
              ids: [sourceId],
              message: $T.ngettext(
                'The category is already here',
                'The selected categories are already here',
                ids.length
              ),
            },
            {
              ids,
              message: $T.ngettext(
                'This is the category you are trying to move',
                'This is one of the categories you are trying to move',
                ids.length
              ),
            },
          ],
        },
      },
```

foxbunny: Thanks, I'll look into it.

indico/web/client/js/react/components/category_navigator/view.jsx
```diff
@@ -0,0 +1,350 @@
...
+
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import Dialog from 'indico/react/components/Dialog';
+import {Translate, PluralTranslate, Param} from 'indico/react/i18n';
+import 'indico/custom_elements/ind_with_tooltip';
+import 'indico/custom_elements/ind_with_toggletip';
+
+import './style.scss';
```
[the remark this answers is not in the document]
ThiefMaster: "Famous last words" ;)

But indeed, I see you have everything under `#category-nav-dialog` which is clear enough. Otherwise classes like `.category-list-section`, `.category-search-results` or `.category-list-container` are not necessarily unique - for example such a class name could also be used on a category display page (even though people should of course check for conflicts anyway:))

indico/web/client/js/react/components/category_navigator/view.jsx
foxbunny: I'm probably overthinking this. Will leave as is for now with just the suggested change.

indico/modules/categories/templates/display/base.html
foxbunny: @ThiefMaster I've looked at this feature. 

From what I was able to gather by reading the code and trying the feature in the management section, it looks like the only thing this feature affects is:

1. Whether the button is disabled
2. The tooltip that is shown when the button is disabled

If these assumptions are correct, the good news is that the new implementation of the category navigator does not need this feature at all. It no longer deals with whether the button is disabled or not. In fact, it has no awareness of how the button is implemented or even whether it's a button. 

We pass it a component that will be rendered through the `actionView` prop. The component receives a category as its only prop, and is free to do whatever it wants. It can disable itself, change the tooltip, change the behavior, appearance, etc. Compared to the current implementation where the responsibility for every use case ever encountered in the app is addressed by the navigator component itself, this should greatly simplify the code.

If this explanation is insufficient, I think it's best that we sit down so I have a chance to walk you (or someone on your team) through the integration of at least one example, and then we can address any oversights and glitches on the spot.

ThiefMaster: Ah, that sounds somewhat easy to adapt. The only problem I have right now is calling this in a purely imperative way from JS code where no DOM element is used at all (currently we trigger this from a click handler on a button)

Ideally I'd like to use it like this (e.g. inside `_moveCategories` when selecting the target category):

```js
const target = await injectModal(resolve => (
  <CategoryNavigator categoryId={sourceId} onClose={() => resolve(null)} onAction={cat => resolve(cat)} />
));
if (target) {
  console.log('selected', target);
} else {
  console.log('dialog closed w/o clicking the action button');
}
```

However, right now it seems to rely on creating the `ind-category-nav` element and triggering an event on `window` to open it (or instantiating one of them for each item so it can bind its event on the trigger).

I guess technically we could use one `ind-category-nav` for each move button, but then we'd still need to bind a JS event when clicking the action button (since it's not just navigating to a new URL when clicking those buttons but rather sending a request in the background). To me it feels cleaner if we could use `injectModel` (which mounts a React component and provides a promise to be resolved using a callback passed to the component): That way we have one place (the click handler for the move button) that first opens the picker, and then does something with its result.

foxbunny: Actually, I forgot I changed the navigator to receive custom events to address that scenario.

So basically, it works like this now:

```
window.dispatchEvent(new CustomEvent('open-category-navigator', {detail: NavigateTo}));
```

The `detail` property is a React component that is used as a button.

You don't need to inject a modal. Render the navigator somewhere on the page, and it will set up a modal that will be open when the custom event is received.

ThiefMaster: but we may have differently configured navigators on the same page (move category + move event for example), with different actions on button click

ThiefMaster: the problem i see with the custom events is that it's very loosely coupled. while it should never happen (since it's shown in a modal after all), one could trigger the click event for two of the move buttons, and then have no way to know for which operation you have the category picker open...

ThiefMaster: > What do you mean by "configuration" in this case?

- I move an event: Any other category is a valid target
- I move a subcategory: Any category that is a child of the category I'm moving is NOT a valid target

see the double-arrow icons on this page (category settings -> content). they all open a category picker:

image

> That sounds like a bug. Why would you be able to move two categories at once?

you are right opening two modals at once would be a bug and is not going to happen unless someone plays around w/ dev tools. I just meant it as an example on why having to rely on global state is a bit ugly. because with being purely event-based I need to:

- store the relevant information, such as which category/event i'm trying to move, when clicking the "move" button next to it
- somehow provide the "config" (as explained above) to that category picker
- when receiving the event that a category has been selected, I need to use the data stored earlier in order to execute the move

at least to me this feels much more complicated than having a single event handler for the click event on the move button - in which I now have all the relevant data in scope, in there I could then simply await the picker selection and then perform the http request to execute the operation - and all the initial data is still in scope since it's all within a single (async) function.

foxbunny: > Any other category is a valid target

What should this change about the category navigator?

> store the relevant information, such as which category/event i'm trying to move, when clicking the "move" button next to it

I don't think you need to do that. You can create a component factory function that will bind such information to the button component being passed to the navigator.

```
function createMoveButton(sourceId) {
  return ({category}) => {
    ....

    const handleClick = () => {
      // Here you have access to both sourceId and category
    }

    return (
      <button>....</button>
    )
  }
}

// ....

window.dispatchEvent(new CustomEvent('open-category-navigator', {detail: createMoveButton(123)}))
```

ThiefMaster: Yeah, your example looks like it can do the job. I'll give it a try!

